Issue #1487, opened 2026-06-18
bug(trusty-memory): HTTP daemon opens palace redb as ReadOnlyClient — a 2nd instance silently degrades to read-only and rejects ALL writes


bobmatnyc:
## Summary
A `memory_remember`/palace write can fail with `palace is read-only: HTTP daemon holds the write lock` and the failure is NOT self-healing. Root cause: when more than one `trusty-memory` process opens the same palace redb files, the second+ process silently falls back to `OpenMode::Snapshot` (read-only) and then **rejects every write for the rest of its lifetime**. A *writer* daemon should never silently serve in read-only mode — it should fail loud at startup instead.

Observed 2026-06-18: during a daemon upgrade, rogue `serve --foreground --http 127.0.0.1:0` listeners (ports 57175/57176) were alive alongside the launchd daemon (:7070). An MCP client routed a write to a rogue (snapshot-mode) instance and got the read-only error; the write was then dismissed as "not worth retrying" → effectively silent data loss of a legitimate memory.

## Root cause (with code evidence)
- Storage is **redb** (single-writer; `flock(LOCK_EX)` on open). Each palace = `kg.redb` + `index.usearch.redb`.
- `KgStoreRedb::open` (`crates/trusty-common/src/memory_core/store/kg_redb/store.rs:44`) and `UsearchStore::new` (`crates/trusty-common/src/memory_core/store/vector.rs:113`) BOTH open with `OpenIntent::ReadOnlyClient`.
- On `DatabaseAlreadyOpen`, `try_open_or_snapshot` (`crates/trusty-common/src/memory_core/store/concurrent_open.rs:213-293`) with `ReadOnlyClient` copies to a `$TMPDIR` snapshot and returns `OpenMode::Snapshot`. Writes are then rejected by `KgStoreRedb::check_writable()` (`store.rs:196`) / `PalaceHandle::remember_with_options()` (`crates/trusty-common/src/memory_core/retrieval/handle.rs:430-436`) with `READ_ONLY_ERROR_MSG` (`store/kg_redb/types.rs:36`).
- The single-instance guard (`crates/trusty-memory/src/commands/single_instance.rs:78`) only probes the HTTP `/health` port — it does NOT guard the redb data files. There is a ~600ms startup race (3×200ms probes) where a second `serve --foreground` can slip through; and `serve --http <addr>` (non-7070) bypasses the `bind_foreground_port()` collision-abort entirely.
[...]

Issue #906, opened 2026-06-08
memory_remember hangs indefinitely on cold embedder init — no timeout on shared_embedder()/embed_batch()/locks


bobmatnyc:
## Summary
`memory_remember` (MCP `POST /rpc`) can hang for minutes to HOURS when the local ONNX embedder is cold. There is no `tokio::time::timeout` anywhere from `handle_memory_remember` down to ONNX inference. `POST /api/v1/remember` returns 202 immediately but its spawned task is equally unbounded.

## Reproduction conditions
1. Cold daemon start with empty CoreML cache — first `memory_remember` after fresh install or after macOS purges `~/Library/Caches/trusty-embedder/coreml`. CoreML compile 30–120s, longer under pressure.
2. Daemon restart after the redb v2→v4 migration wiped palace stores (2026-06-03) — the startup pin scan triggered `shared_embedder()` init and blocked the whole scan.
3. ANE/CoreML inference stall — `embed_batch` holds a `parking_lot::Mutex` on the model for the inference duration; a stalled inference blocks all later embeds forever.

## Evidence
Daemon log `~/Library/Application Support/trusty-memory/logs/stderr.log`:
`startup pin scan complete: 0 pin(s) discovered in 19116509ms`  ← 19,116 s = 5.3 hours, the restart right after the redb migration wipe.
Warm probe: `POST /rpc` memory_remember small=0.126s, large=0.146s; `POST /api/v1/remember`=<2ms (202). Entity/KG extraction (`kg_extract.rs`) is pure/deterministic, no I/O, no LLM — NOT the cause. OpenRouter/Bedrock are not on the remember path.

## Root cause (file:line)
- `crates/trusty-common/src/memory_core/retrieval/mod.rs:53-63` `shared_embedder()` `OnceCell::get_or_try_init` — no timeout; concurrent callers queue here.
- `crates/trusty-common/src/memory_core/retrieval/mod.rs:597-600` `embedder.embed_batch(&[content]).await` — no timeout; `embedder/mod.rs:683` runs `spawn_blocking` under a `parking_lot::Mutex`.
- `crates/trusty-common/src/embedder/mod.rs:575` `FastEmbedder::with_cache_size` `spawn_blocking` (CoreML compile) — no timeout.
[...]

Issue #59, opened 2026-05-22
MCP stdio client cannot open palace when HTTP daemon is running (exclusive redb lock contention)


bobmatnyc:
## Bug Description

When the trusty-memory HTTP daemon is running (holding exclusive redb locks on palace vector stores), any new stdio MCP client process receives `open palace <name>` errors on every tool call that requires palace access.

## Error (stdio MCP client)

```
open palace claude-mpm
```

Occurs on: `memory_recall`, `memory_recall_deep`, `memory_remember`, `palace_info`, `kg_query`

## Root Cause

The redb database format used for palace vector stores (`index.usearch.redb`) does not support multiple concurrent readers/writers. The HTTP daemon acquires an exclusive lock at startup and holds it indefinitely. Any subsequent process attempting to open the same palace file is blocked.

## Impact

This creates a split-brain condition: Claude's stdio MCP session cannot perform any semantic memory operations while the HTTP daemon is running. Effectively, users must choose between the HTTP API or the MCP interface — they cannot coexist.

## Proposed Fix

The HTTP daemon should be the **sole owner** of all palace access and expose a complete REST API including recall endpoints. The stdio MCP server should delegate all palace operations to the HTTP daemon via localhost rather than opening palace files directly.

Alternatively, explore a shared-memory or RPC proxy pattern, or migrate from redb to a database that supports concurrent readers (e.g., SQLite WAL mode, or a separate vector store with a network interface).

## Environment

- macOS 25.3.0
- trusty-memory v0.1.72
- HTTP daemon: PID 75539 (launchd)
- stdio MCP client: spawned per Claude session

bobmatnyc:
Fixed in commit afbf9b0 — snapshot-on-contention in storage layer: stdio opens a read-only point-in-time snapshot when HTTP daemon holds the redb lock.

PR #3996, opened 2026-07-26
fix(trusty-common): bound the palace open-queue wait under sustained writer contention

### LINKED ISSUE #3992: trusty-memory: the write path retries redb lock acquisition unboundedly — memory_remember hangs indefinitely under multi-process contention (read path fails fast)
## Symptom
A `memory_remember` MCP call was accepted and then produced nothing for **1800 seconds**, until the calling harness aborted it. The read path stayed healthy throughout — `palace_list` returned 66,945 characters promptly during the same window, and prompt-hook recall continued working every turn. Observed 2026-07-26.

## Root cause (diagnosed with live evidence, not inferred)
A 3-second non-invasive thread-stack sample of the launchd-managed daemon (`trusty-memory serve --foreground`, PID 53507, `com.trusty.memory`, listening on 127.0.0.1:7070) caught **six separate tokio blocking-pool worker threads parked inside `trusty_common::memory_core::store::concurrent_open::backoff_sleep_ms`** — actively cycling a write-lock acquire/backoff/retry loop at the moment of the check.

The asymmetry is the defect:
- **Read path (`OpenIntent::Reader`)**: bounded. After ~5 attempts it emits an explicit ERROR and refuses, e.g. `redb file is still locked by another process after the write-lock handoff retry window ... attempts=5`. It fails fast and surfaces the problem.
- **Write path (`OpenIntent::Writer`, used by `memory_remember`)**: no bounded give-up is visible in the logs or in the observed behavior. It retries indefinitely, so the caller hangs rather than receiving an error.

## Why there is contention at all
Ten `trusty-memory` processes were running concurrently against the same on-disk palace files: the one launchd HTTP daemon plus **nine `trusty-memory serve --stdio` processes, one per active Claude/tm session**. These are independent OS processes (not children of the daemon), each opening the same `.redb` files in `~/Library/Application Support/trusty-memory/palaces/*` directly. redb's advisory file locking permits a single writer, so any write contends with whichever process currently holds the lock.

The daemon's own logs show this is long-standing, not new — multi-day history is dominated by `Database already open. Cannot acquire lock.` and the `attempts=5` writer-refusal ERROR.

## Impact
Any `memory_remember` (and presumably any other write-intent call — `memory_note`, `kg_assert`, chat-turn appends) can hang the caller indefinitely whenever another trusty-memory process holds the write lock. Because the daemon is shared and stdio instances are per-session, contention scales with the number of concurrent sessions. A hang is materially worse than an error here: the caller cannot distinguish "slow" from "never," and in this instance burned 1800s before an external timeout.

## No data loss
The aborted write never committed. redb's copy-on-write B-trees leave the file at its last committed state rather than torn. No palace file had been modified in 4+ hours at diagnosis time, consistent with writes simply not landing. No stale lock marker files; `trusty-memory doctor` and `tm doctor` both reported healthy (they check liveness and lock-marker cleanliness, neither of which detects this live in-process contention — worth noting as a doctor-coverage gap).

## Suggested fixes (decision required, not settled)
1. **Bound the writer retry**, mirroring the reader's existing policy: after a deadline, return a clear error naming the contending condition instead of retrying forever. This is the minimal fix and directly addresses the reported symptom.
2. Consider whether the per-session `serve --stdio` processes should open palace files directly at all, or whether they should proxy writes through the single HTTP daemon so exactly one process holds write intent. That is the architectural fix; it is larger and should be weighed separately.
3. Consider surfacing live lock contention in `trusty-memory doctor` / `tm doctor` — both currently report healthy while six threads are stuck in a retry loop.

## Operational note
`KeepAlive.SuccessfulExit = false` in `~/Library/LaunchAgents/com.trusty.memory.plist` means the daemon auto-respawns within `ThrottleInterval` (10s), so an operator can cycle it but not stop it. Restarting the daemon alone may not clear contention if the lock holder is one of the `--stdio` session processes. Correct restart form: `launchctl kickstart -k gui/$(id -u)/com.trusty.memory`.

## Explicitly a red herring
Today's log ends with a rapid burst of `semantic consolidation LLM call failed ... POST http://localhost:11434/v1/chat/completions` (Ollama not running — no listener on 11434), then goes silent. That failure is pre-existing and fails fast; the timing is coincidental, not causal. Worth its own issue if nobody has filed one, but it is not this bug.

---

bobmatnyc:
## Summary

- Fixes #3992: `memory_remember` (and any Writer-intent palace open) could
  hang far longer than any documented bound under sustained multi-process
  redb lock contention.
- Root cause, confirmed against the running code (not assumed): each
  individual `try_open_or_snapshot` retry under `OpenIntent::Writer` is
  already bounded to ~1.55s (issue #1487 / PR #1488). The asymmetry the
  issue describes as "reader bounded / writer unbounded" is real, but it
  is **not** inside that retry loop — it is `PalaceRegistry::open_palace`'s
  per-palace `open_lock`, previously acquired with a plain, unbounded
  `parking_lot::Mutex::lock()`. A failed Writer open is never cached (a
  nonexistent handle must not be), so under sustained contention every
  caller queued behind the lock repeats the full ~1.55s bounded dance from
  scratch once it's finally its turn — the Nth caller waits for N-1 earlier
  callers' own attempts first, with **no ceiling on the total**.
- `open_palace` now acquires that lock via `try_lock_for`, bounded by a new,
  independently-tunable `memory_core::timeouts::open_queue_timeout()`
  (default 60s, `TRUSTY_OPEN_QUEUE_TIMEOUT_SECS`) — a sibling of the
  existing `write_lock_timeout` (issue #906) kept as its own knob since the
  two guard different pipelines. A caller stuck behind persistent
  contention now gets a clear, actionable ERROR instead of hanging.

## Before / after evidence (real multi-writer contention, not mocked)

New test `writer_open_queue_wait_is_bounded_under_sustained_contention`
holds a genuine flock conflict on the real backing redb file (`kg.redb` —
`KnowledgeGraph::open_with_intent` translates the legacy `kg.db` path
callers pass in) via a second `Database::create`, exactly like the existing
`writer_intent_fails_on_locked_file` test, then fires 5 concurrent
`open_palace` calls at a `with_writer_intent` registry for the same palace.

**Before the fix** (`open_lock.lock()`, unbounded):
```
writer 0: elapsed=2.126474042s ... still locked ... attempts=5
writer 1: elapsed=3.979972958s ... still locked ... attempts=5
writer 2: elapsed=5.894132833s ... still locked ... attempts=5
writer 3: elapsed=8.048605667s ... still locked ... attempts=5
writer 4: elapsed=10.031702125s ... still locked ... attempts=5
writer_open_queue_depth: 5 concurrent writers, total=10.032225333s,
  max_single_caller_wait=10.031702125s
```
The 5th caller waited **10 seconds** for a palace that never became
available — each caller serialized behind every earlier caller's own
bounded attempt, with the wait growing linearly with queue depth. This is
exactly how a single `memory_remember` call was observed to hang for
1800s under sustained contention from 9 concurrent sessions.

**After the fix** (`open_lock.try_lock_for(open_queue_timeout)`, bounded;
test injects a 2s deadline via the new `with_open_queue_timeout` builder to
stay fast/deterministic without mutating global env state):
```
writer 4: elapsed=2.149208625s ... open queue timed out after 2s ...
writer 1: elapsed=2.14923225s  ... open queue timed out after 2s ...
writer 3: elapsed=2.149233416s ... open queue timed out after 2s ...
writer 0: elapsed=2.149233875s ... open queue timed out after 2s ...
writer 2: elapsed=2.188552709s ... still locked ... attempts=5
writer_open_queue_depth: 5 concurrent writers, total=2.188760708s,
  max_single_caller_wait=2.188552709s
```
All 5 callers now return within ~2.2s total instead of 10s, with the
queue-timeout error naming the actual cause.

## `tm doctor` gap (not fixed here, reported separately per the brief)

`tm doctor` / `trusty-memory doctor` reported HEALTHY while six threads
were genuinely stuck retrying a redb lock. This fix makes any *individual*
caller's wait bounded and produce a clear error, but does **not** make
doctor detect live in-process lock-contention — that is a fully independent
gap (doctor checks liveness/lock-marker cleanliness, not live retry-loop
occupancy) and should be filed/handled as its own issue.

## Test plan

- [x] `cargo test -p trusty-common --lib --features memory-core` — 605
      passed, 0 failed
- [x] `cargo test -p trusty-common --features memory-core` (integration
      suites included) — all green
- [x] `cargo test -p trusty-memory` (unit + full integration suite,
      including `serve_stdio_*_e2e`) — 452 lib + all integration tests
      passed, 0 failed
- [x] `cargo clippy -p trusty-common --lib --features memory-core
      --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p trusty-memory --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo check -p trusty-agents -p trusty-mpm` (downstream
      `PalaceRegistry` consumers) — clean, no breaking API changes

Closes #3992.

### DIFF
--- a/crates/trusty-common/CHANGELOG.md
+++ b/crates/trusty-common/CHANGELOG.md
@@ -24,6 +24,24 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **`PalaceRegistry::open_palace` no longer queues callers indefinitely
+  behind sustained redb writer-lock contention** (issue #3992): the
+  per-palace `open_lock` that serialises concurrent opens of the same
+  palace id was acquired with an unbounded `Mutex::lock()`. Each individual
+  `try_open_or_snapshot` retry under `OpenIntent::Writer` was already
+  bounded (~1.55s), but a failed Writer open is never cached, so every
+  caller queued behind the lock repeated that bounded dance from scratch —
+  the Nth queued caller waited for N-1 earlier callers' own attempts to
+  finish first, with no ceiling on the total. Reproduced live: 5 concurrent
+  openers against a genuinely held lock made the 5th caller wait 9.86s, and
+  this exact linear-with-queue-depth pattern is how a single
+  `memory_remember` call was observed to hang for 1800s under sustained
+  multi-process contention. `open_palace` now acquires the lock via
+  `try_lock_for` with a new, independently-tunable
+  `memory_core::timeouts::open_queue_timeout()` (default 60s,
+  `TRUSTY_OPEN_QUEUE_TIMEOUT_SECS`), returning a clear error instead of
+  hanging once the deadline elapses.
+
 - **`default_model_matches_sentence_transformers_reference` no longer
   misreports a wrong-model load as an fp32 accuracy failure** (issue #3711):
   the merge-gate flake that failed at `mean_cosine=0.990649` vs `>= 0.999` was
--- a/crates/trusty-common/src/memory_core/registry.rs
+++ b/crates/trusty-common/src/memory_core/registry.rs
@@ -123,12 +123,49 @@ pub struct PalaceRegistry {
     /// just-evicted palace would double-open and the loser would error. A tiny
     /// per-id sync mutex makes the second racer wait, re-check the cache, and
     /// share the winner's handle instead.
+    ///
+    /// Why bounded (issue #3992): a FAILED `Writer` open is never cached (a
+    /// nonexistent handle must not be), so under *sustained* cross-process
+    /// contention every caller queued behind this mutex repeats the full
+    /// bounded (~1.55 s) `try_open_or_snapshot` dance from scratch once it is
+    /// finally its turn. Each individual redb-open attempt is bounded, but
+    /// the mutex acquisition itself previously was not: the Nth queued
+    /// caller waited for N-1 earlier callers' own ~1.55 s attempts to finish
+    /// first, so a caller's *total* wait scaled with queue depth with no
+    /// ceiling -- reproduced live (issue #3992: 5 concurrent openers against
+    /// a held lock made the 5th caller wait 10 s; sustained contention across
+    /// many requests over a long window is exactly how a single
+    /// `memory_remember` call was observed to hang for 1800 s). The reader
+    /// path has no analogous queue-depth problem to mirror: a `ReadOnlyClient`
+    /// open always resolves quickly via the snapshot fallback (issue #59) and
+    /// therefore never queues for long. So `open_palace` bounds this wait
+    /// with a dedicated
+    /// [`crate::memory_core::timeouts::open_queue_timeout`] (issue #3992,
+    /// default 60 s, `TRUSTY_OPEN_QUEUE_TIMEOUT_SECS`) -- a sibling of the
+    /// established `write_lock_timeout` (issue #906) that governs the
+    /// per-palace write mutex once a palace is already open, kept as its own
+    /// knob since the two protect different pipelines and operators may need
+    /// to tune them independently. A caller stuck behind persistent
+    /// contention gets a clear, actionable ERROR within one bounded window
+    /// instead of hanging indefinitely.
     /// What: `DashMap<PalaceId, Arc<parking_lot::Mutex<()>>>`. Held only across
     /// the (rare) open pipeline — never across an `.await` — so a
     /// `parking_lot::Mutex` is safe. Entries are lightweight and left in place
-    /// (bounded by palace count) rather than reference-counted out.
-    /// Test: `registry_tests::evict_idle_then_reopen_preserves_recall`.
+    /// (bounded by palace count) rather than reference-counted out. Acquired
+    /// via `try_lock_for(open_queue_timeout())` in `open_palace`.
+    /// Test: `registry_tests::evict_idle_then_reopen_preserves_recall`,
+    /// `registry_tests::writer_open_queue_wait_is_bounded_under_sustained_contention`
+    /// (issue #3992 — proves the bound against real multi-writer contention).
     open_locks: Arc<DashMap<PalaceId, Arc<Mutex<()>>>>,
+    /// Bound applied to `open_lock.try_lock_for` in `open_palace` (issue
+    /// #3992). Defaults to
+    /// [`crate::memory_core::timeouts::open_queue_timeout`] (process-wide env
+    /// override) but is stored per-instance — rather than re-read from the
+    /// environment on every call — so [`PalaceRegistry::with_open_queue_timeout`]
+    /// can inject a short deadline in tests without mutating global process
+    /// state (which would race other tests running in parallel).
+    /// Test: `registry_tests::writer_open_queue_wait_is_bounded_under_sustained_contention`.
+    open_queue_timeout: Duration,
 }
 
 impl Default for PalaceRegistry {
@@ -166,7 +203,24 @@ impl PalaceRegistry {
             // daemon overrides this via `with_writer_intent` (issue #1487).
             open_intent: OpenIntent::ReadOnlyClient,
             open_locks: Arc::new(DashMap::new()),
+            open_queue_timeout: crate::memory_core::timeouts::open_queue_timeout(),
         }
+    }
+
+    /// Override the per-palace open-queue timeout (issue #3992).
+    ///
+    /// Why: production callers get a sane env-configurable default (see the
+    /// `open_locks` field doc); tests that need to prove the bound against
+    /// real multi-writer contention within a fast, deterministic wall-clock
+    /// budget inject a short deadline here instead of mutating the
+    /// process-wide `TRUSTY_OPEN_QUEUE_TIMEOUT_SECS` env var (which would
+    /// race any other test running in parallel).
+    /// What: Consuming builder that overwrites `open_queue_timeout`.
+    /// Test: `registry_tests::writer_open_queue_wait_is_bounded_under_sustained_contention`.
+    #[must_use]
+    pub fn with_open_queue_timeout(mut self, timeout: Duration) -> Self {
+        self.open_queue_timeout = timeout;
+        self
     }
 
     /// Create a registry whose open-handle cap comes from the environment.
@@ -363,10 +417,16 @@ impl PalaceRegistry {
     /// is silently evicted (its fds close); the data is safe on disk.
     /// What: Returns the cached `Arc<PalaceHandle>` if present (promotes to MRU);
     /// otherwise loads metadata via `PalaceStore::load_palace`, calls
-    /// `PalaceHandle::open`, and inserts the handle (may evict LRU).
+    /// `PalaceHandle::open`, and inserts the handle (may evict LRU). The
+    /// per-palace open-lock acquisition is bounded (issue #3992) — see the
+    /// `open_locks` field doc for why an unbounded wait there let sustained
+    /// writer contention hang a caller indefinitely even though every
+    /// individual redb-open attempt underneath is itself bounded.
     /// Test: `registry_create_and_open` round-trips create -> drop -> reopen;
     /// `lru_evicted_handle_reopens` verifies evicted handles are transparently
-    /// reopened; `open_palace_follows_alias` covers the alias redirect.
+    /// reopened; `open_palace_follows_alias` covers the alias redirect;
+    /// `writer_open_queue_wait_is_bounded_under_sustained_contention` proves the bound
+    /// against real multi-writer contention.
     pub fn open_palace(&self, data_root: &Path, palace_id: &PalaceId) -> Result<Arc<PalaceHandle>> {
         // Fast path: an already-open handle under the requested id.
         if let Some(h) = self.get(palace_id) {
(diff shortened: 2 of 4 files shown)


bobmatnyc:
## Adversarial review — PR #3996

Verdict: **APPROVE**

I did not take the PR's claims on faith. What follows is independently
executed evidence for each attack vector in the review brief, including a
correction the coordinator issued mid-review (the original bug brief
misattributed the root cause to an unbounded *retry*; the engineer's
corrected diagnosis is an unbounded *queue-lock acquisition* one layer up —
I re-verified the corrected diagnosis from first principles, not from either
party's say-so).

### (a) Does the bound actually fire under REAL contention? — VERIFIED BY EXECUTION

Ran the new test as shipped (`writer_open_queue_wait_is_bounded_under_sustained_contention`,
`crates/trusty-common/src/memory_core/registry_tests.rs:801`) against the fixed
code:

```
writer 2: elapsed=2.018s ... open queue timed out after 2s ...
writer 0: elapsed=2.095s ... open queue timed out after 2s ...
writer 4: elapsed=2.105s ... open queue timed out after 2s ...
writer 1: elapsed=2.105s ... open queue timed out after 2s ...
writer 3: elapsed=2.137s ... still locked ... 5 write-lock acquisition attempts
test ... ok (2.20s total)
```

Then I reverted **only** `registry.rs`'s `try_lock_for` back to the old
`open_lock.lock()` (unbounded) and re-ran the identical test, unmodified:

```
writer 3: elapsed=2.076s
writer 2: elapsed=4.234s
writer 0: elapsed=6.285s
writer 4: elapsed=8.457s
writer 1: elapsed=10.440s   <-- linear growth with queue depth, matches the PR's
                                 claimed 9.86s/10.03s "before" numbers almost exactly
thread ... panicked: "a queued caller must not wait longer than ..." got 10.44s
test ... FAILED
```

This is a genuine OS-level `flock(LOCK_EX)` conflict on a real `kg.redb` file
(via a second `Database::create`, same technique as the pre-existing
`writer_intent_fails_on_locked_file` test) — not a mock, not a sleep-based
simulation. I restored the fix afterward and confirmed the file matches the
PR's committed diff exactly (`git diff` clean).

I also independently confirmed (by reading `concurrent_open.rs`,
`kg_redb/store.rs`, and `store/vector.rs`, not by trusting either the
original brief or the engineer's correction) that no unbounded retry hides
anywhere in the redb-open call chain:
- `open_writer_with_handoff_retry` (`concurrent_open.rs:331`) is a
  `for attempt in 0..WRITER_RETRY_ATTEMPTS` (5) bounded loop, ~1.55s total —
  confirmed both by reading the loop and by seeing "after 5 write-lock
  acquisition attempts" in my own test output.
- Both `KgStoreRedb::open_with_intent` and `open_or_get_cached_db` (vector
  store) compute `max_attempts = RETRIES * u8::from(intent != Writer)`, i.e.
  **0** extra outer-loop attempts for `Writer` intent — a single bounded
  `try_open_or_snapshot` call, never multiplied.
- The only genuinely unbounded primitive pre-fix was
  `PalaceRegistry::open_locks`'s `parking_lot::Mutex::lock()` in
  `registry.rs` — exactly what the corrected diagnosis says, and exactly
  what my revert-and-rerun proved causes the linear-with-queue-depth hang.

The mechanism (concurrent tokio tasks/threads within the trusty-memory
daemon process queueing behind one per-palace `open_lock`, each repeating
its own bounded ~1.55–3.1s dance) also matches the daemon's actual
architecture: I traced `serve --stdio` (`main.rs:671`) to
`serve_stdio_bridge::run_stdio_bridge` — it's a **bridge** to the one HTTP
daemon (the only call site of `.with_writer_intent()` in the entire
codebase, `main.rs:789`), not a second writer-intent process. So "10
processes contend" in production really means ~10 concurrent in-process
tokio tasks queueing on the same `open_lock`, which is precisely what the
test reproduces with threads. Confidence: verified, not assumed.

### (b) Is a bounded WRITE failure safe? — VERIFIED BY CODE READING

Traced every write-path caller of `open_palace`/`open_palace_handle`:
- `memory_ops.rs:130` (`memory_remember`) and `:229` (`memory_note`):
  `let handle = open_palace_handle(state, palace)?;` — propagates via `?`
  all the way to the MCP tool response. No swallow.
- `kg_ops.rs:50` (`kg_assert`): same `?` propagation.
- `helpers.rs:422` (`write_drawer`, called by both `memory_remember` and
  `memory_note`): same `?` propagation, with `.with_context()` on the
  original `open_palace` call site (`helpers.rs:298`) adding
  `"open palace {palace_id}"` context to the error.

The only `.ok()`/`is_ok()`/`is_err()` swallow sites I found
(`memory_ops.rs:475` `memory_recall_all`, `web/health.rs:312`,
`chat/tools.rs` multiple) are all read-only fan-out or health-probe paths
where "palace didn't open" is already a legitimate skip-and-continue
outcome, unrelated to this PR and pre-existing. No write path silently
drops a `memory_remember`/`memory_note`/`kg_assert`.

### (c) Is 60s the right production default, given 10 processes contend routinely? — VERIFIED BY CALCULATION, one MEDIUM finding

Per-attempt bound for a Writer-intent open is ~1.55s (kg.db) with a second,
independent ~1.55s budget for the vector store if that one is also
contended (`retrieval/handle.rs:297-303` opens vector store then KG
sequentially) — call it ≤3.1s worst case per queued caller's own attempt.
With the documented "10 processes" normal-contention ceiling, the last
queued caller's wait is ≈9 × 1.55–3.1s ≈ 14–28s — comfortably inside the 60s
default (2–4x headroom), so I don't think 60s will produce spurious
write failures under the documented normal load. I could not find a basis
to say the default is miscalibrated.

**MEDIUM finding** — `crates/trusty-memory/src/tools/memory_ops.rs:117-119`
(`write_lock` acquisition, bounded by `timeouts::write_lock_timeout()`,
default 60s) runs **before** `open_palace_handle` is called
(`memory_ops.rs:127` / via `write_drawer` → `helpers.rs:422`), and
`open_palace` itself now has its own independent ~60s
`open_queue_timeout` (`registry.rs:464-465`) plus up to ~3.1s of bounded
redb-attempt time. These stack additively on the hot path: a single
`memory_remember`/`memory_note` call under sustained contention can now take
up to **~60 + 60 + 3.1 ≈ 123s** before returning an error, not the single
bounded window the PR description implies ("a caller stuck behind
persistent contention gets a clear, actionable ERROR within one bounded
window"). That's a ~2-minute stall from the caller's perspective — vastly
better than the reported 1800s, but still long enough that an operator
watching a hung `memory_remember` might reasonably think the fix didn't
fully land. Not blocking (both halves are correctly bounded and each is
independently tested; this is a documentation/config-surface gap, not a
correctness defect) — worth either documenting the compounded worst case
explicitly, or tracking a shared deadline budget across `write_lock` and
`open_lock` in a follow-up so they don't stack.

### (d) Blast radius — VERIFIED BY GREP, no CRITICAL/HIGH found

Enumerated every `open_palace`/`PalaceRegistry` caller in the workspace.
`.with_writer_intent()` is called from exactly **one** place in the entire
codebase: `AppState::with_writer_intent` (`trusty-memory/src/lib.rs:837`),
invoked once from the daemon's `serve --foreground` startup
(`trusty-memory/src/main.rs:789`). Every other caller defaults to
`OpenIntent::ReadOnlyClient`:
- `trusty-agents/src/memory/trusty_backed.rs:134` —
  `PalaceRegistry::new()` (ReadOnlyClient), propagates errors via
  `.with_context()?` (no swallow).
- `trusty-mpm/src/core/sm/memory.rs:191` and
  `trusty-mpm/src/daemon/manager/memory.rs:392` — both call
  `PalaceRegistry::open()` (ReadOnlyClient default) and both **do** swallow
  the `open_palace` error via `if let Ok(handle) = ...open_palace(...) {
  return } ` falling through to `create_palace` (pre-existing
  open-or-create idempotency pattern, not introduced by this PR).

I confirmed from `concurrent_open.rs` that `ReadOnlyClient` intent never
exhausts a retry budget on lock contention — it falls back to a read-only
snapshot immediately (issue #59), so each attempt resolves in milliseconds
regardless of contention. That means the per-palace `open_lock` queue these
two trusty-mpm callers sit behind never accumulates meaningful wait time,
so the new 60s `open_queue_timeout` is not practically reachable for them —
the swallow-and-fall-through-to-create pattern is a real, pre-existing
latent risk (`PalaceStore::save_palace` unconditionally overwrites
`palace.json` on every `create_palace` call — confirmed by reading
`palace_store.rs:127-141`, no existence check) but this PR does not
meaningfully increase its exposure, because the callers that could trigger
it never queue long enough to hit the bound. I looked for this specifically
because the corrected brief flagged it as the important question, not
because I assumed it away.

`cargo check -p trusty-agents -p trusty-mpm` — clean (see Quality Gate
below), confirming no breaking API change, independent of the behavioral
analysis above.

### (e) The doctor gap — VERIFIED: untouched by this PR

`crates/trusty-memory/src/commands/doctor/mod.rs` is not among this PR's
changed files (diff touches only `trusty-common/src/memory_core/{registry.rs,
registry_tests.rs, timeouts.rs}` + `trusty-common/CHANGELOG.md`). I also read
`doctor/mod.rs` directly: its checks are daemon HTTP-health ping, fastembed
cache presence, and stale `*.lock` **sidecar files on disk** — none inspect
live in-process thread/retry-loop occupancy. File the doctor gap separately;
it is not touched here and would not be a duplicate.

### Quality gate — all verified by direct execution, not accepted from the PR body

```
  → 605 passed; 0 failed; 13 ignored   (unit)
  → community_tests: 6 passed; kg_graph_tests: 11 passed;
    kg_store_snapshot_tests: 4 passed  (integration, no --lib restriction)

cargo test -p trusty-common --features "memory-core embedder-test-support" --test memory_palace
  → 3 passed; 0 failed

cargo test -p trusty-memory
  → 452 passed; 0 failed; 4 ignored   (lib)
  → chat_mcp: 11, config_mount: 2, dream_room_mcp: 3, isolation_override: 5,
    kg_rebuild_cli: 2, mcp_stdio_tools: 14, migration_boot: 1, palace_force: 3,
    serve_stdio_concurrent_e2e: 2, serve_stdio_e2e: 4, task_mcp: 5
    — all passed, 0 failed, across every integration suite including the
    serve_stdio_*_e2e suites

cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings
  → clean

cargo clippy -p trusty-memory --all-targets -- -D warnings
  → clean

cargo fmt --check
  → clean (exit 0)

cargo check -p trusty-agents -p trusty-mpm
  → clean (only a pre-existing, unrelated `proc-macro-error2` future-incompat
    warning from a transitive dependency)
```

### Verdict

**APPROVE.** No CRITICAL or HIGH findings. One MEDIUM (stacked
`write_lock_timeout` + `open_queue_timeout` ≈ 123s worst-case latency on the
hot write path, item (c) above) — worth a follow-up doc note or a shared
deadline budget, not blocking. The core claim — that this PR bounds a
previously-unbounded per-palace open-queue wait under real, reproduced
multi-writer redb contention, and that writes fail loud rather than
silently — held up under an independent revert-and-reproduce test, direct
reading of every write-path caller, and a full blast-radius sweep of every
`PalaceRegistry` consumer in the workspace. The `tm doctor` live-contention
gap is real but genuinely untouched by this PR — file it separately.

Issue #4044, opened 2026-07-26
trusty-review: grade/verdict not computed from the finding set — F/BLOCK with 7 of 34 findings self-withdrawn, plus raw chain-of-thought published as output


bob-duetto:
## Summary

Two related output-aggregation defects observed on dry-run reviews from 2026-07-25/26:

1. **The grade/verdict is not computed from the surviving finding set.** One review graded **F / BLOCK** on 34 findings, of which **7 conclude, in their own published text, that there is no defect** — three explicitly using the word "Withdrawing." Withdrawn and self-refuted findings are still being counted toward the verdict.
2. **Raw deliberation reaches the published output.** One finding published a mid-reasoning correction verbatim, beginning "Wait — actually looking at the code more carefully" and ending "Withdrawing this finding — confidence too low."

These are output-hygiene and aggregation defects, independent of retrieval quality or finding accuracy.

**Environment**

| | |
|---|---|
| trusty-review | 0.10.1 (live daemon; no version string is emitted in the review artifact itself) |
| trusty-search | 0.39.0 |
| Model | `us.anthropic.claude-sonnet-4-6` (per artifact) |
| Diff source | `raw` (per artifact) |
| Mode | dry-run, advisory |
| Dates | 2026-07-25, 2026-07-26 |

Full review artifacts (all 65 findings, verbatim):
- duettoresearch/code-intelligence#2857 — 34 findings, F / BLOCK
- duettoresearch/code-intelligence#2858 — 31 findings, F / BLOCK

---

## Not a duplicate of #3475

**#3475** ("Re-tune the severity floor against a shared labeled calibration set") is about *threshold tuning* — where the severity floor sits, and calibrating it against labeled data. That is a legitimate and separate piece of work and we are not asking to expand or duplicate its scope.

This issue is upstream of any threshold: **findings that the reviewer itself has withdrawn or refuted are entering the aggregation at all.** No severity floor can fix a set that contains items the reviewer has already stated are non-findings. Re-tuning the floor with these still in the set would in fact calibrate against contaminated data.

If #3475's calibration work proceeds, filtering withdrawn/refuted findings first would improve the quality of the set it calibrates against.

---

## 1. The arithmetic

Review 1: **34 findings → grade F, verdict BLOCK.**

Seven of those 34 findings end by affirming the code is correct. Verbatim closing lines:

| # | Closing text (verbatim) |
|---|---|
| 22 | "This is correctly implemented." |
[...]

Issue #4043, opened 2026-07-26
trusty-review: reports the repo's SPEC contract as the implementation — systematically refuted authorization/scoring findings


bob-duetto:
## Summary

On two dry-run reviews of PRs that contain both a design spec (`docs/specs/admin-judging-api.md`) and the implementation of that spec, trusty-review repeatedly grounded **code** findings in the **spec** text. Where the spec had drifted behind the implementation, the reviewer reported the spec's weaker contract as though it were the shipped behavior.

Every authorization and scoring-engine finding in these two reviews was refuted on manual inspection of the actual code.

This is a distinct defect from citation fabrication (filed separately). Here the citations are real, the file paths exist, and the reasoning is internally coherent — the finding is simply grounded in the wrong artifact.

**Environment**

| | |
|---|---|
| trusty-review | 0.10.1 (live daemon; no version string is emitted in the review artifact itself) |
| trusty-search | 0.39.0 |
| Model | `us.anthropic.claude-sonnet-4-6` (per artifact) |
| Diff source | `raw` (per artifact) |
| Mode | dry-run, advisory |
| Dates | 2026-07-25, 2026-07-26 |

Full review artifacts (all 65 findings, verbatim):
- duettoresearch/code-intelligence#2857 — 34 findings, F / BLOCK
- duettoresearch/code-intelligence#2858 — 31 findings, F / BLOCK

---

## The mechanism, in the reviewer's own words

The clearest statement of the failure is review 2, finding 6. The reviewer notices that the implementation is not in view, and then reasons about the implementation anyway:

> However, the diff contains only the spec document (`docs/specs/admin-judging-api.md`) and migrations — no implementation of `src/lib/judging/aggregate.ts`, `src/lib/judging/z-score.ts`, or any other scoring engine file is present. The `AggregateInput` interface shown in the spec does NOT include `conflictedProposalIdsByVoter`

> The spec prose says the parameter exists but the interface definition shown in the same spec omits it. This is an internal inconsistency in the spec itself. **If implementation follows the interface literally, the COI filter for stack-rank picks will be absent.**

> [code: `docs/specs/admin-judging-api.md` — "AggregateInput" interface definition vs. §6.6 prose about `conflictedProposalIdsByVoter`]

[...]

PR #2882, opened 2026-07-17
fix(trusty-review): verify code_provable citations to stop fabricated fail-closed findings (closes #2881)

bobmatnyc:
## Summary

Closes #2881 (P0). trusty-review 0.9.0/0.9.1 fail-closed a clean PR to
D+/REQUEST_CHANGES on a **fabricated `code_provable: true` finding**: the reviewer
claimed a new vitest test file's content (`instructions-content.test.ts`) had been
"prepended" into a large regenerated bundle (`api/chat.js`, ~35k lines) around line
35271. That content lives in a *different* changed file and never appears in the
cited file. The deterministic grade floor
(`grade::drives_block_floor = cited || code_provable`) trusted the flag at face
value and forced BLOCK.

## Root cause — it is NOT a parser bug

The issue suspected diff-hunk misattribution. I reproduced the exact shape (new
test file hunk + small source hunk + very large bundle hunk with high line numbers)
and confirmed empirically that **attribution through the pipeline is correct**:
`parse_diff_files` → `split_into_units` produces per-file map units, and the
`api/chat.js` unit contains **zero** vitest content — no cross-file leakage. Each
map unit carries only its own file's diff text.

The real defect is a **citation-verification gap**: nothing checked that a
diff-provable *claim* is actually grounded in the cited file. The fabricated
finding is a model **confabulation**, and the floor had no fail-open guard against
it — the exact failure mode the #2825 citable-findings calibration was meant to
prevent.

## Fix

New `crates/trusty-review/src/pipeline/citation_check.rs`:

- `DiffContentIndex::from_filtered` indexes the surviving diff by file (the exact
  content the reviewer saw, from the post-filter `FilteredDiff`).
- `downgrade_uncitable_findings` scans each finding that claims diff-provability
  (`code_provable` or a `code:` citation), extracts the concrete code fragments it
  quotes, and verifies them against the cited file. A finding whose quote is **not
  in the cited file** — because it belongs to another changed file (the #2881
  shape) or is absent from the diff — is **downgraded**: `code_provable` cleared, a
  `code:` citation stripped, confidence lowered to the refuted floor. It is
  downgraded to advisory, **never dropped**.

Wired into **both** review paths, before the verdict floor runs:
- unified — `runner.rs`, before `apply_grade_and_floor`;
- map-reduce — `mapreduce/mod.rs`, before `reduce` derives the aggregate verdict.

**Fail-open by design** (high precision, zero false positives on legit findings):
[...]

Issue #726, opened 2026-06-04
feat(trusty-review): adversarial verifier too aggressive — refuted a true finding during auto-review of PR #720


bobmatnyc:
## Context
During end-to-end testing of the health-gated auto-review flow (trusty-review 0.3.1, Bedrock claude-sonnet-4-6), `review_pr` was run against PR #720 (trusty-search launchd warm-boot fix). The reviewer surfaced **5 findings**, then the adversarial verification stage **refuted all 5** — including one **legitimate** finding (a dropped `JoinHandle` leading to potential pool exhaustion, which is the same concern tracked separately as #723).

Net result of that run: APPROVE, 0 findings retained, ~$0.11, dry_run (0 comments posted).

## Problem
The adversarial verifier appears biased toward refutation: a 100% refutation rate on a batch that contained at least one true-positive indicates the refute step is discarding real findings. Over-aggressive refutation defeats the purpose of the review (false negatives reach `main`).

## Impact
- Real correctness findings can be silently dropped before they're ever surfaced as PR comments.
- Erodes trust in auto-review APPROVE verdicts.

## Proposed directions (for investigation, not prescriptive)
- Review the refutation prompt — it may be primed to "default to refuted" too strongly for a verification (rather than discovery) context.
- Consider a confidence band: surface "refuted-but-uncertain" findings instead of dropping them outright.
- Consider requiring stronger/explicit evidence to refute a finding (raise the bar for refutation).
- Add a regression check: the #720 dropped-JoinHandle finding is a known true-positive — use it as a fixture to validate verifier calibration.

## References
- True-positive that was wrongly refuted is tracked in #723 (dropped JoinHandle / pool exhaustion hardening).
- Observed in trusty-review 0.3.1 during auto-review wiring tests.

PR #4057, opened 2026-07-27
fix(trusty-review): drop fabricated/withdrawn findings, close citation coverage gap

### LINKED ISSUE #4042: trusty-review 0.10.1: fabricated findings from diff-path misattribution have returned — 7 findings assert mutually exclusive contents for one file (regression of #2881)
## Summary

On two dry-run reviews produced 2026-07-25 and 2026-07-26, trusty-review emitted findings whose file/line citations do not correspond to the file's actual contents. The clearest instance: **seven findings in a single review attribute four mutually exclusive contents to one file**, `src/lib/team-registration-early-access.ts`. They cannot all be true, and none of them are.

This is the same failure shape as **#2881** ("trusty-review 0.9.0: fabricated `code_provable=true` finding from diff-hunk misattribution (fail-closed, blocks clean PRs)"), which was closed by **#2882** ("verify `code_provable` citations to stop fabricated fail-closed findings"). We are reporting it as a regression: the citation verification added in #2882 either regressed on 0.10.1 or does not cover this path.

**Environment**

| | |
|---|---|
| trusty-review | 0.10.1 (live daemon; no version string is emitted in the review artifact itself) |
| trusty-search | 0.39.0 |
| Model | `us.anthropic.claude-sonnet-4-6` (per artifact) |
| Diff source | `raw` (per artifact) |
| Mode | dry-run, advisory |
| Dates | 2026-07-25, 2026-07-26 |

Full review artifacts (all 65 findings, verbatim):
- duettoresearch/code-intelligence#2857 — 34 findings, F / BLOCK
- duettoresearch/code-intelligence#2858 — 31 findings, F / BLOCK

---

## The centerpiece: four mutually exclusive descriptions of one file

All quotes below are verbatim from the **same review** (34 findings, one PR). Every one of them is about `src/lib/team-registration-early-access.ts`. This is internally verifiable — no access to our repo is needed to see that these cannot be simultaneously true.

**Finding 1 — it is a Drizzle JSON snapshot:**

> The first file in the diff is `src/lib/team-registration-early-access.ts` but its entire content is a Drizzle ORM JSON snapshot (a `drizzle/meta/` artifact).

> [code: `src/lib/team-registration-early-access.ts:1` — "{ \"id\": \"893644d4...\""] — the file opens with a raw JSON object, not valid TypeScript.

**Finding 2 — it is a Vitest suite for `draft-form.ts`:**

> The file `src/lib/team-registration-early-access.ts` is named and pathed as if it were a library module, but its content is entirely a Vitest test suite (`describe`/`it`/`expect`) for `draft-form.ts`.

> [code: `src/lib/team-registration-early-access.ts:1` — "import { describe, expect, it } from 'vitest'"]

**Finding 3 — it is a `compute-results` POST handler:**

> The file added at `src/lib/team-registration-early-access.ts` contains the `POST` handler for `compute-results` (the round-results materialization endpoint), not anything related to team registration or early access.

> [code: `src/lib/team-registration-early-access.ts:1` — `import { and, asc, eq, notInArray } from "drizzle-orm"`]

**Finding 6 — it is a Vitest suite for the judging-proposals route:**

> The first file in the diff is `src/lib/team-registration-early-access.ts` but its content is clearly the test suite for `GET /api/judging/rounds/:roundId/proposals` (it imports from `./route`, uses `vitest`, and describes that route).

**Finding 33 — it is the override route:**

> In `team-registration-early-access.ts` (the override route), the new row inherits `publishedAt` from the superseded row

Two further findings (**4** and **13**) build on these premises, #4 asserting a build break —

> However, `src/lib/team-registration-early-access.ts` in this diff contains only a Vitest test suite — no exported `isTeamRegistrationOpenFor` function. If this is the complete content of that module, the build will fail with a missing export error.

— and #13 citing a specific line inside the imagined `compute-results` handler:

> [code: `src/lib/team-registration-early-access.ts:329` — `tieBreakNote: row.tieBreakNote`]

### Ground truth

The real `src/lib/team-registration-early-access.ts` is an ordinary TypeScript module that exports exactly the function finding #4 says is missing (`isTeamRegistrationOpenFor`). It is none of: a JSON snapshot, a Vitest suite, a `compute-results` handler, or an override route. Build, typecheck, and 920 tests pass on the PR head — which is impossible under any one of the four claimed contents, let alone all of them.

Note the citation line numbers: `:1`, `:329`. The second review cites `:3` and `:256` in the same file. The line numbers are attached with the same confidence as the (unverifiable) content.

### Same shape on a second PR, same file

The second review (31 findings, an unrelated PR) reproduces the pattern independently: findings 1, 2, 3, 8, 10, 18 and 31 again attribute `src/lib/team-registration-early-access.ts` variously to a `draft-form` Vitest suite, a raw Drizzle JSON snapshot, and route tests for `GET /api/admin/rounds/:roundId/results` —

> The file `src/lib/team-registration-early-access.ts` contains route tests for `GET /api/admin/rounds/:roundId/results` (the file's own JSDoc says so).

— and finding 4 applies the same misattribution to a *different* file:

> The file `src/components/funnel-timeline.tsx` contains a `vitest` `describe` block testing `getChromeVisibility` from `./site-chrome`.

**Across the two reviews, 15 of 65 findings (23%) are citation-misattribution artifacts.**

---

## Two further confidently-wrong specifics

These are not mutually-exclusive-content cases, but they show the same pattern of a confident assertion where verification was available in the diff or tree.

**1. A CHECK constraint that does not exist.** Review 1, finding 16:

> If `submitted_at` has a `>= created_at` check constraint (consistent with the pattern documented for `published_at`), this is a latent correctness issue.

The only CHECK constraint in the entire schema is on `round_results`. There is no such constraint on `ballots`. The premise was checkable against the schema files present in the tree; the finding was emitted on the unverified conditional instead.

**2. A version pin that is in the diff.** Review 2, finding 17:

> the pin itself is not visible in the diff for confirmation

The exact-version `drizzle-orm` pin is at `package.json:32`, with no caret, in the diff.

---

## A second, independent occurrence on 0.10.1

On **2026-07-23**, on an unrelated PR, trusty-review 0.10.1 emitted a finding citing **`hotelPage.ts:207`** — a TypeScript `Promise.race`. That path and line do not exist. Our pipeline logs contain zero corresponding entries. Different repo, different PR, different day, no relationship to the two reviews above — same failure shape.

So on 0.10.1 we have two independent incident clusters of fabricated citations, plus the original 0.9.0 incident in #2881.

---

## Configuration caveat (please weigh this)

We want to be straightforward about a confound in the 2026-07-25/26 runs.

Those two reviews were produced while our 0.10.1 daemon was running with **`TRUSTY_REVIEW_REQUIRE_SEARCH=false`**, a systemd drop-in we applied 2026-07-24 to work around a degraded trusty-search on the same host (trusty-search reports `status=degraded`, 188/222 indexes failing redb corpus open). Under that setting the daemon's own required-context gate is bypassed, and its logs explicitly label the run *"proceeding DEGRADED (non-authoritative)"* / *"review produced WITHOUT code context."*

The precise nuance, neither overstated nor understated:

- **These were not context-free reviews.** Our application layer attaches its own code context independently of the daemon fetch — a verified run on the same configuration shows `context_files=6`, `referenced_code_chars=3479`, served from a healthy `main` index. The model did receive real code.
- **But the daemon-side context fetch was gated off**, so the 15/65 fabrication *rate* may be inflated relative to a fully healthy configuration. We would not want you to take 23% as the number to design against.
- **The defect itself is not an artifact of this configuration.** The independent `hotelPage.ts:207` fabrication occurred **2026-07-23, before the override was applied**, and the original #2881 fabrication was on 0.9.0 with no such override.

**This caveat is time-bounded, and we are closing it out.** We are reverting the `TRUSTY_REVIEW_REQUIRE_SEARCH=false` override now and pausing our org-wide auto-review until our trusty-search is remediated (the underlying `DatabaseAlreadyOpen` cause is #3929). Once we are back on a fully healthy, gate-enforced configuration we will **re-run these exact two PRs and report back here** with the fabrication count under clean conditions, whichever way it goes. You will get the clean measurement without having to ask for it.

The mutually-exclusive-contents proof above is unaffected by this caveat: no context configuration makes four contradictory descriptions of one file correct.

---

## Why we are filing this at P1

A reviewer that misses defects is a reviewer with limited recall. A reviewer that fabricates is a different category of problem:

1. **The output is indistinguishable from a true finding.** Fabricated findings here are well-written, cite a real file path, cite a plausible line number, and reason coherently from the invented content. There is no surface signal. They are only refutable by opening the file.
2. **It is fail-closed.** Both reviews graded **F / BLOCK**. Fabricated findings block clean PRs.
3. **It poisons trust in the true findings.** Once a reviewer has demonstrably invented four contradictory descriptions of one file, a human cannot triage the remaining findings any faster than reading the diff themselves — which removes the tool's entire value.

## Suggested direction

Deliberately non-prescriptive; you know the internals:

- Verify **every** cited path and line against the actual diff + tree before emitting a finding — not only `code_provable=true` ones. The #2881 fix scoped verification to `code_provable`; the findings above are not all in that class, which may be why the fix does not cover them.
- Make an unresolvable or contradicted citation a **dropped** finding, never a confident one. Silence beats fabrication for a fail-closed reviewer.
- Consider a cheap internal consistency check before emit: if two findings in one review assert incompatible contents for the same path, at most one can be right, and the safest action is to drop both and surface a diagnostic.
- If the underlying cause is diff-hunk-to-path mapping (as diagnosed in #2881), a regression test built from a multi-file diff where a `src/lib/*.ts` module sits adjacent to `.json` and `.test.ts` additions would likely reproduce it.

We can provide the raw diff for either PR on request.

---

Related: #4043 (spec-as-implementation), #4044 (verdict not derived from findings), #4045 (regression coverage request). Prior art: #2881 / #2882.

bobmatnyc:
## Summary

Fixes the trusty-review reliability cluster: three P1/P2 bugs in the review
GATE itself emitting claims it cannot support.

Closes #4042, #4043, #4044.

## Root cause: related but NOT one shared code defect

All three are failures of the same underlying *property* — every emitted
finding must cite a path (and, where quoted, content) that verifiably exists
— but the mechanisms are genuinely different:

- **#4042 (citation fabrication):** the #2882 fix scoped verification to
  `code_provable`/`code:`-cited findings, and — the load-bearing gap —
  **exempted the mandated `[code: \`path:line\` — "excerpt"]` citation's own
  excerpt** from verification (assumed to always be a paraphrase). #4042's
  evidence shows the model routinely puts verbatim, fabricated "evidence"
  exactly there.
- **#4044 (unsound verdict + CoT leak):** an aggregation/output-hygiene gap —
  self-withdrawn findings ("Withdrawing this finding…") and raw deliberation
  text ("Wait — actually looking at the code more carefully…") were never
  filtered before counting toward the grade or reaching rendered output.
- **#4043 (spec-as-implementation):** a provenance gap — findings grounded in
  a spec/doc artifact were emitted as claims about shipped code, with no
  distinction carried downstream.

Each needed a different mechanism; fixing #4042's citation-check alone would
not have caught #4043 or #4044, and vice versa.

## Fix shape — drop, not downgrade (Bob's simplify mandate)

Per the explicit instruction to prefer removing the code path that lets an
unverifiable finding get emitted over adding another verification layer on
top:

- **`citation_check::enforce_citation_integrity`** (was
  `downgrade_uncitable_findings`) now **drops** (never downgrades-in-place):
  - every finding's `file` must resolve in the diff (Kept, SummaryOnly, AND
    Stage-A-dropped files are now all indexed — dropped files with empty
    content, so citing them is legal but quoting fabricated content for them
    is not);
  - the mandated `[code: …]` citation's own excerpt is now verified against
    its own cited path (previously the one thing exempted);
  - a same-file/same-line mutual-contradiction backstop: if two findings
    assert incompatible content at the exact same locator, both are dropped
    (scoped to same-line specifically — two real findings about *different*
    lines of one file are normal and must never be flagged).
- **New `finding_hygiene` module** (#4043, #4044):
  - `drop_self_negated_or_leaked_findings` — marker-based drop of any finding
    whose own text already withdraws it or leaks raw deliberation.
  - `demote_diff_absent_speculation` — demotes (does not drop — the
    documentation-drift signal is real) any High/Critical finding admitting
    it reasons about an implementation absent from the diff.
  - `relax_verdict_if_evidence_wiped` — when hygiene + citation-integrity wipe
    out every finding this run, the model's own raw `verdict`/`grade` fields
    (produced by the same hallucinating call) are relaxed to APPROVE too —
    otherwise dropping the findings closes the deterministic-floor half of
    "fabricated findings block clean PRs" but not the self-report half. This
    is scoped tightly to a LOCALLY OBSERVED "we just wiped it this run" fact
    so it never touches `derive_verdict`'s shared floor semantics other
    callers (`verify::rederive_verdict`'s #726 infra-failure preservation)
    rely on unchanged.

Wired into both the unified (`runner.rs`) and map-reduce (`mapreduce/mod.rs`,
per-chunk) paths.

## Proof by execution

`tests/citation_4042_regression.rs` reproduces #4042's centerpiece shape (a
real TS module diffed alongside a Drizzle JSON snapshot and a Vitest test
file, matching #4042's own suggested reproducing shape — no access to the
reporter's private repo `duettoresearch/code-intelligence` was available):

- **Before fix** (ran against unmodified code first): 0 of 3 fabricated
  findings caught — confirmed the regression exactly as reported.
- **After fix**: all 3 fabricated findings dropped, the legitimate finding on
  the SAME file survives verbatim, and the fully-fabricated case additionally
  proves the model's own raw self-reported BLOCK relaxes to APPROVE (not just
  the floor).

`finding_hygiene_tests.rs` uses verbatim/near-verbatim text from #4043 and
#4044's own issue bodies as fixtures.

## Before/after (regression-suite proxy, since the real reported PRs are in a
private repo we cannot access)

| | citation_4042_regression.rs | citation_2881_regression.rs |
|---|---|---|
| Before (unmodified code) | 0/3 fabricated findings dropped | fabricated finding survives, downgraded-in-place |
| After | 3/3 dropped, legit finding survives, top-level BLOCK relaxes | fabricated finding dropped, verdict recovers |

6 pre-existing `runner_tests.rs` fixtures used header-less synthetic diffs
whose fake-LLM findings cited a file never present in the diff — legal under
the old narrow check (gated on `code_provable`/citation + a matching quote),
now correctly caught by the broadened path-resolution check. Fixed via a new
`local_diff_source_for_file` helper producing a proper unified diff instead
of loosening the new invariant.

## Residual scope (disclosed, not fully closed — mirroring #2882's own disclosed gap)

- `[code: …]` verification checks path + content, not the cited LINE number
  specifically via hunk-line arithmetic — a wrong line number attached to
  otherwise-correct content is not caught.
- #4043's fix is marker-based on the finding's own admission; it does not
  catch a finding grounded in an in-code COMMENT's hypothetical (review 2
  finding 25 in #4043) — the cited file there is genuinely code, so
  extension/path-based classification doesn't apply. Flagged as a known
  follow-up, same shape as #2882's disclosed quote-less-paraphrase gap.

## Publish note

trusty-review 0.10.1 is the currently published, live version and IS affected
by #4042/#4043/#4044 in production. A patch release (0.10.2) is warranted
once this merges — **not published from this PR** per instructions.

## Gates (raw output)

```
cargo test -p trusty-review
  lib:          1539 passed; 0 failed; 5 ignored
  bin (main.rs):   68 passed; 0 failed
  citation_2881_regression.rs: 3 passed; 0 failed
  citation_4042_regression.rs: 3 passed; 0 failed
  config_mount.rs:              2 passed; 0 failed
  report_analyze_e2e.rs:        2 passed; 0 failed
  report_e2e.rs:                3 passed; 0 failed
  report_investigate.rs:        3 passed; 0 failed

cargo clippy -p trusty-review --all-targets -- -D warnings   → clean
cargo fmt --check -p trusty-review                            → clean
scripts/check_line_cap.sh                                     → 8 allowlisted, 0 violations — OK
scripts/check_sld.sh                                           → 47 spec docs + 2884 code files; 0 errors, 0 warnings
scripts/check_test_pointers.sh                                 → exit 0
```

### DIFF
--- a/crates/trusty-review/CHANGELOG.md
+++ b/crates/trusty-review/CHANGELOG.md
@@ -5,7 +5,52 @@ All notable changes are documented in this file.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ---
-## [0.10.1] — 2026-07-23
+## [Unreleased]
+
+### Fixed
+
+- **Reliability cluster: fabricated findings, spec-as-implementation, and
+  unsound verdicts** (closes
+  [#4042](https://github.com/bobmatnyc/trusty-tools/issues/4042),
+  [#4043](https://github.com/bobmatnyc/trusty-tools/issues/4043),
+  [#4044](https://github.com/bobmatnyc/trusty-tools/issues/4044)):
+  three related emit-time integrity gaps that let trusty-review publish
+  claims it could not support.
+  - **#4042 — citation verification is now DROP, not downgrade, and covers
+    every finding.** The #2882 fix scoped verification to
+    `code_provable`/`code:`-cited findings and, critically, exempted the
+    mandated `[code: \`path:line\` — "excerpt"]` citation's own excerpt from
+    verification — exactly where #4042 showed the model puts its fabricated
+    "evidence." `citation_check::enforce_citation_integrity` now verifies
+    EVERY finding's `file` against the diff (Kept, SummaryOnly, AND
+    Stage-A-dropped files are all indexed so path existence is checkable
+    regardless of disposition), verifies `[code: …]` excerpts against their
+    own cited path, and adds a same-file/same-line mutual-contradiction
+    backstop for findings that assert incompatible content at one locator. A
+    finding that fails is DROPPED, never downgraded-in-place.
+  - **#4044 — self-withdrawn findings and chain-of-thought leaks.** New
+    `finding_hygiene::drop_self_negated_or_leaked_findings` removes any
+    finding whose own text already negates it ("Withdrawing…", "The code is
+    correct.") or leaks raw mid-reasoning deliberation ("Wait — actually
+    looking at the code more carefully…") before it can reach the rendered
+    review or the verdict floor. `finding_hygiene::relax_verdict_if_evidence_wiped`
+    additionally relaxes a model's own raw `verdict`/`grade` fields to
+    APPROVE when hygiene + citation-integrity wiped out every finding this
+    run — closing the loop so a BLOCK that rested entirely on
+    fabricated/withdrawn evidence cannot survive as the model's untouched
+    self-report.
+  - **#4043 — spec/doc reported as implementation.** New
+    `finding_hygiene::demote_diff_absent_speculation` demotes (never drops —
+    the documentation-drift signal is real) any High/Critical finding whose
+    own text admits it is reasoning about an implementation absent from the
+    diff ("no implementation of…", "if implementation follows the interface
+    literally…"), capping it below the BLOCK floor.
+  - Known residual scope (disclosed, not fully closed): (a) `[code: …]`
+    citation verification checks path+content, not the cited LINE number
+    specifically — no hunk-line-number arithmetic is performed; (b) #4043's
+    fix is marker-based on the finding's own admission and does not catch a
+    finding grounded in an in-code COMMENT's hypothetical (the code file
+    itself is real, so extension/path-based detection does not apply there).
 
 ### Fixed
 
--- a/crates/trusty-review/src/models/mod.rs
+++ b/crates/trusty-review/src/models/mod.rs
@@ -221,6 +221,21 @@ pub enum FindingCategory {
 
 // ─── Finding ──────────────────────────────────────────────────────────────────
 
+/// Sentinel `Finding::file` value used when the LLM omits a file (a genuinely
+/// general, non-file-scoped observation).
+///
+/// Why: `citation_check::enforce_citation_integrity` (#4042) requires every
+/// finding's `file` to resolve against the diff — but a finding with no
+/// specific file at all is a legitimate shape, not a fabrication, and must not
+/// be dropped for "not resolving" a path it never claimed. Naming the sentinel
+/// once (rather than repeating the `"unknown"` literal in `parser.rs` and
+/// `citation_check.rs`) keeps the two call sites from drifting apart.
+/// What: the literal string `parser::convert_llm_finding` substitutes when the
+/// LLM's `file` field is empty.
+/// Test: `parse_finding_missing_file_defaults_to_unknown` (parser_tests.rs),
+/// `unknown_file_sentinel_is_exempt_from_path_check` (citation_check_tests.rs).
+pub const UNKNOWN_FILE_PLACEHOLDER: &str = "unknown";
+
 /// A single code-review finding / fix suggestion.
 ///
 /// Why: the pipeline extracts findings from the LLM review body and attaches
--- a/crates/trusty-review/src/pipeline/citation_check.rs
+++ b/crates/trusty-review/src/pipeline/citation_check.rs
@@ -1,30 +1,53 @@
-//! Deterministic citation verification for diff-provable findings (#2881).
+//! Deterministic citation verification for every emitted finding (#2881, #4042).
 //!
 //! Why: a `code_provable: true` finding (or a `code:`-cited finding) drives the
 //! deterministic BLOCK / REQUEST_CHANGES floor unconditionally
-//! (`grade::drives_block_floor` = `cited || code_provable`).  The reviewer LLM can
-//! confabulate such a finding — the #2881 repro emitted a high-severity
-//! `code_provable: true` finding claiming a new vitest test file's content had been
-//! "prepended" into a large regenerated bundle (`api/chat.js`), content that lives
-//! in a DIFFERENT changed file and never appears in the cited file at all.  The
-//! grade floor trusted the flag at face value and fail-CLOSED a clean PR to
-//! D+/REQUEST_CHANGES.  Diff-hunk attribution through the parse → split pipeline is
-//! correct (verified by the `mapreduce`/`diff_analyzer` tests); the missing layer
-//! is a check that a diff-provable CLAIM is actually grounded in the cited file.
+//! (`grade::drives_block_floor` = `cited || code_provable`).  The reviewer LLM
+//! can confabulate such a finding — the #2881 repro emitted a high-severity
+//! `code_provable: true` finding claiming a new vitest test file's content had
+//! been "prepended" into a large regenerated bundle (`api/chat.js`), content
+//! that lives in a DIFFERENT changed file and never appears in the cited file
+//! at all.  The original (#2882) fix downgraded (never dropped) any finding
+//! whose free-text quotes failed to ground in the cited file, but scoped the
+//! check to `code_provable`/`code:`-cited findings only, and — the load-bearing
+//! gap #4042 exploited — deliberately EXEMPTED the prompt-mandated
+//! `[code: `path:line` — "excerpt"]` citation's own excerpt from verification,
+//! on the theory that it was always a paraphrase.  #4042 showed the model often
+//! puts its verbatim "evidence" for a fabrication INSIDE that exact bracket
+//! (attributing four mutually exclusive contents to one file, `hotelPage.ts:207`
+//! citing a file/line that never existed in the diff at all), so the one
+//! citation form the check was built to verify was the one place it never
+//! looked.
 //!
-//! What: [`DiffContentIndex`] indexes the surviving diff by file (the exact content
-//! the reviewer saw, built from the post-filter [`FilteredDiff`]).
-//! [`downgrade_uncitable_findings`] scans findings and, for each finding that claims
-//! to be provable from the diff, verifies the cited file exists in the diff AND that
-//! the concrete code fragments the finding quotes actually appear in that file.  A
-//! finding that fails verification is DOWNGRADED (never silently dropped): its
-//! `code_provable` flag is cleared, a `code:` citation to the unverifiable location
-//! is removed, and its confidence is lowered to the refuted floor so it can no longer
-//! force any verdict floor — it survives as an advisory note.  The check is
-//! FAIL-OPEN: a finding that quotes nothing concrete, or whose quote is grounded in
-//! the cited file, is left untouched.
+//! What: rather than adding another downgrade layer on top of the confabulated
+//! finding, this module makes the emit path structurally incapable of carrying
+//! an unverifiable citation — every finding's citations are checked and, on
+//! failure, the finding is DROPPED (not downgraded, not neutered-but-kept):
 //!
-//! Test: `citation_check_tests.rs`.
+//!  - [`DiffContentIndex::from_filtered`] indexes every file the diff mentions
+//!    (Kept, SummaryOnly, AND Dropped/Stage-A-excluded) so path EXISTENCE can be
+//!    checked regardless of disposition, with content available only for
+//!    Kept/SummaryOnly files.
+//!  - [`enforce_citation_integrity`] runs on EVERY finding (not just
+//!    `code_provable`/cited ones — closing the #4042-named coverage gap) and
+//!    drops a finding whose `file` does not resolve in the diff at all, whose
+//!    free-text quotes contradict the cited file's real content, or whose
+//!    `[code: …]` bracket citation — INCLUDING its excerpt, no longer exempted
+//!    — is unresolvable or contradicted.
+//!  - A cheap, low-false-positive INTERNAL CONSISTENCY check
+//!    ([`detect_line_contradictions`]) drops any pair of findings that cite the
+//!    SAME file+line with two mutually exclusive quoted contents — at least one
+//!    must be invented, and the safest action is to drop both (#4042's own
+//!    suggested direction) — even when the file's content is not independently
+//!    indexable (SummaryOnly / Stage-A-dropped).
+//!
+//! FAIL-OPEN remains the default wherever there is genuinely nothing to verify:
+//! a finding with no quoted content at all is left alone (only path existence is
+//! checked); the `Finding::UNKNOWN_FILE_PLACEHOLDER` sentinel (a legitimately
+//! file-less, general observation) is exempt from path resolution entirely.
+//!
+//! Test: `citation_check_tests.rs`; end-to-end regressions in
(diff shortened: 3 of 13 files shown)


bobmatnyc:
## Verdict: WARN

Reviewed 487bcc82 (rebase-preserved content of c14c61d4) against merge-base 971e2fb6. Verified by reading `citation_check.rs`/`finding_hygiene.rs` directly and running the new test modules (`cargo test -p trusty-review --lib -- citation_check finding_hygiene` → 37 passed; `--test citation_4042_regression` → 3 passed). All green, no `#[ignore]`/skip tricks. No CRITICAL findings — the core #4042/#4043/#4044 fixes are real and the happy-path tests are not vacuous. The three HIGH findings below are all instances of the failure mode this PR itself flags as most dangerous: **legitimate findings silently dropped**, not fabricated ones surviving.

## Findings

| HIGH | `crates/trusty-review/src/pipeline/citation_check.rs:92-95,474-478` | 92 | `CODE_CITATION_RE` only extracts a `CodeCitation` when the bracket matches the exact backtick+dash grammar. A `[code: …]` bracket that deviates even slightly (no backticks around the path, a plain `-` the regex still allows, but e.g. missing backticks or an unexpected internal `]`) fails this regex, so no `CodeCitation` is produced — **and** the broader `BRACKET_CITATION_RE` (line 71-74) still strips the *entire* bracket, excerpt included, before the generic quote scan runs (line 474). Net effect: a malformed-but-clearly-a-code-citation bracket bypasses all verification, landing in the same fail-open bucket as "no citation at all" — exactly the #4042 gap, just for a slightly different string shape. This is not merely theoretical: this is the identical mechanism #4042 exploited, and neither `citation_check_tests.rs` nor `citation_4042_regression.rs` has a single malformed-bracket-format fixture (checked both files for `malformed`/backtick-variant cases — none exist). The module's own doc claims the excerpt is "no longer exempted," which is only true for the one exact grammar shape. | Add a test asserting the failure mode explicitly (e.g. `[code: src/foo.ts:42 — "fabricated"]` without backticks) and decide the fallback deliberately: either loosen `CODE_CITATION_RE` to tolerate the missing-backtick case, or — safer — when `BRACKET_CITATION_RE` matches but `CODE_CITATION_RE` does not, treat that as its own verification failure (drop) rather than silent passthrough. |
| HIGH | `crates/trusty-review/src/pipeline/citation_check.rs:394-408` (`detect_line_contradictions`) | 401 | The "mutual contradiction" check treats any two distinct findings citing the same `(file, line)` as contradictory unless one quoted span is a substring of the other (`!span_a.contains(span_b) && !span_b.contains(span_a)`). Two *independently true* findings about the same line that quote **non-overlapping** substrings (e.g. one finding quoting the SQL-building half of a line, another quoting the unsanitized-input half of the same line, for two different real concerns) get BOTH dropped even though neither is fabricated — "non-overlapping" is treated as proof of contradiction when it only proves "different quote." This is exactly the danger case the review brief called out: silent loss of two real findings. No test exercises two real, non-overlapping quotes of one legitimate line (existing coverage is `drops_findings_with_line_contradiction`, genuinely contradictory content, and `distinct_lines_are_never_contradictory`, different lines — neither covers "same line, both true, non-overlapping quotes"). | Require actual semantic incompatibility, not just non-overlap, before dropping — e.g. only flag contradiction when the quotes overlap in a shared span position but disagree there, or fall back to a softer signal (downgrade rather than drop) when spans are simply disjoint substrings of the same line rather than incompatible. At minimum, add a regression test for the disjoint-but-both-true case to make the current behavior an explicit, reviewed decision rather than an untested side effect. |
| HIGH | `crates/trusty-review/src/pipeline/finding_hygiene.rs:75-94` (`SELF_NEGATION_MARKERS`) | 82,83,85 | Markers `"is correctly implemented"`, `"the code is correct"`, `"actually correct"` are matched via plain case-insensitive `str::contains` over the whole finding text with no negation-scope awareness. A legitimate finding can use this exact phrasing *contrastively* — e.g. "The username field's sanitization is correctly implemented, but the email field bypasses it entirely, allowing XSS" — and would be dropped outright, even though it reports a real, distinct bug. The module doc claims this was "deliberately avoided" by using multi-word phrases instead of bare words like "correct," but the test that's supposed to prove it (`does_not_drop_legitimate_finding_using_the_word_correct`, `finding_hygiene_tests.rs:46-57`) uses `"was correct"` — a phrasing that doesn't match ANY of the actual markers — so the claimed safety property is untested against the markers that actually carry the risk. | Add a test using one of the real markers in a contrastive sentence (e.g. the XSS example above) and confirm today's behavior (drop) is intended; if not, require the marker to appear in the finding's *concluding* sentence/near a withdrawal cue, or restrict matching to shorter finding bodies where contrast is unlikely, rather than a plain whole-text substring match. |

## Notes

- Verified `relax_verdict_if_evidence_wiped`'s scope claim (does not interfere with `verify::rederive_verdict`'s separate infra-failure path, #726): traced both call sites (`runner.rs:563`, `mapreduce/mod.rs:101` vs `verify.rs:254`) — they operate on different data at different pipeline stages and the guarantee holds as documented.
- In map-reduce mode specifically, `detect_line_contradictions` runs per-chunk (`mapreduce/mod.rs:88-108`, before `reduce`), not across chunks, so the disjoint-quote false-positive above is a single-call/single-chunk risk, not a cross-chunk one — still real for a single reviewer emitting two findings about one line.
- No panics/unwraps introduced beyond `.expect()` on two compile-time-constant regex literals (safe).
- Did not flag the `UNKNOWN_FILE_PLACEHOLDER` sentinel or `runner_tests.rs`'s `local_diff_source_for_file` helper — both are correctly scoped/pre-existing per the PR description.

Not blocking merge — these are coverage/precision gaps in the drop logic worth closing (ideally with the regression tests above) before or shortly after landing, since a silent false-drop is invisible until someone notices a finding they expected never rendered.

PR #3994, opened 2026-07-26
fix(trusty-mpm): deny writes to pm_guard's own trust anchor

bobmatnyc:
## Summary

`.claude/settings.json` / `.claude/settings.local.json` (and the `$CLAUDE_CONFIG_DIR` global equivalent) were an unconditional, unbudgeted ALLOW in `tm hook --pm-guard` — `.json` is not a source-code extension, so `evaluate_edit_tool` fell through to allow. Claude Code applies that file's top-level `env` object to every session/subprocess it spawns (hooks included, live-reloaded mid-session), so a `Write`/`Edit` there could set `TRUSTY_MPM_DISABLE_HOOKS`/`TRUSTY_MPM_PM_UNRESTRICTED` and silently disable every pm_guard rule — the PM's own trust anchor was writable by the thing the guard constrains.

**Honesty note, added after the adversarial review below: this PR closes the EASY, naive self-exemption paths — the ordinary shell verbs a PM would reach for without specifically probing for a gap. It is NOT a hardened security boundary against a party with unrestricted Bash access.** See "Update: response to review" for the two gaps the critic found, now closed, and the ones that remain open by design (documented, not hidden).

- New `pm_guard_trust_anchor` module resolves the target path (relative, `~`, `..` traversal, symlinked ancestors — all without requiring the target to exist yet) and matches it structurally: any path whose parent directory is literally `.claude` and whose filename is `settings.json`/`settings.local.json`, OR the exact `$CLAUDE_CONFIG_DIR/settings*.json` equivalent (confirmed that dir has NO nested `.claude/`, unlike the project/home case).
- Checked from **three** places: `pm_guard()`'s body directly, before Guards 1/4's subagent-exemption early returns (mirrors the worktree-tmp guard's placement — a dispatched subagent's write is exactly as dangerous as the PM's own, so it must NOT be exempt); `evaluate_edit_tool`, as a second pure-function layer of defense in depth; and `pm_guard_bash::bash_targets_trust_anchor`, closing the equivalent shell route as an **absolute, non-budget-eligible** deny (previously a Bash write to the file would have been the ordinary budget-eligible `SHELL_EDIT_REASON`, allowed up to 3 times before blocking).
[...]

Issue #3977, opened 2026-07-26
trusty-mpm: hard-block git worktree add targeting /tmp, /private/tmp, /var/folders, or the scratchpad


bobmatnyc:
## Summary

`tm hook --pm-guard` has no rule blocking `git worktree add` targeting `/tmp`,
`/private/tmp`, `/var/folders`, `$TMPDIR`, or the harness scratchpad. Worktrees
provisioned there silently fail unrelated tests (see #3955 — trusty-search's
`SENSITIVE_PATH_PREFIXES` denylist deterministically fails ~12 tests for any
worktree rooted under `/tmp`) and can be reaped mid-task by the OS or a
harness cleanup pass, losing in-flight agent work.

## Why the fix must live BEFORE Guard 4, not inside `evaluate_bash_command`

The obvious place to add this rule is the existing Bash classifier,
`pm_guard_bash::evaluate_bash_command` (called from `evaluate_tool`, which
classifies every `Bash` PreToolUse call). That placement is a **no-op** for
exactly the callers who actually create these worktrees:

`crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs:184-210` ("Guard 4"):

```rust
if payload_is_subagent_dispatch(&payload) {
    if pm_guard_deny_by_default::deny_by_default_enabled()
        && evaluate_tool(tool_name, tool_input).is_some()
    { ... }
    return Ok(());   // ALLOW — evaluate_tool is never even reached here
}
let Some(reason) = evaluate_tool(tool_name, tool_input) else { ... };  // line 212
```

`evaluate_tool` (line 212) is the ONLY caller of `evaluate_bash_command`
(lines 331-337 pre-fix), and it is reached only AFTER Guard 4's early return.
So any `PreToolUse` payload with a non-empty `agent_id` — i.e. every native
Task/Agent-dispatched subagent — is allowed without classification.
`deny_by_default_enabled()` is opt-in and off by default.

This is proven by an existing passing test,
`crates/trusty-mpm/tests/tm_hook_pm_guard.rs::pm_guard_native_subagent_dispatch_allows_bash`:
`sed -i`, denied unconditionally for the PM's own shell, is ALLOWED when
dispatched with an `agent_id`. Subagents are exactly who runs
`git worktree add` when following the project's own worktree-based PR
workflow (see the `tm-pr-workflow` skill), so a rule placed inside
`evaluate_bash_command`/`classify_bash_segment` would silently never fire for
that calling pattern.

[...]

Issue #3600, opened 2026-07-21
fix(trusty-mpm): rename + pm_guard_deny_by_default trust bare TM_MANAGED_SESSION_ID with no pane-identity cross-check


bobmatnyc:
## Summary

`TM_MANAGED_SESSION_ID` is published via `tmux set-environment -t <session>` (`crates/trusty-mpm/src/runtime/claude_code.rs:757-781`, `session_manager/manager.rs:694-704`). That primitive is **tmux-SESSION-scoped, not pane-scoped** — tmux applies a session's stored environment to every NEW pane/window created afterward in that same tmux session. A sibling pane therefore inherits an identical `TM_MANAGED_SESSION_ID` and believes it is a session it is not.

`bin/tm/commands/guided.rs`'s nested-session guard hit this exact hijack (the "#2456 CORRECTION", module doc ~lines 115-136) and fixed it by comparing tmux's own stable `pane_id` (`tmux_attach::current_tmux_pane_id`, never inherited across panes) against the resolved record's own captured `pane_id` (`pane_identity_confirmed`) before ever trusting the env var for a destructive action. That fix was never applied to two other CLI consumers that read the bare env var with **no cross-check at all**:

1. **`bin/tm/commands/rename.rs:18,29-50`** — `tm sessions rename <newname>` (no explicit target) resolves "current session" from a bare `std::env::var("TM_MANAGED_SESSION_ID")`. From a sibling pane this **renames the wrong session — possibly another user's live session.** Highest blast radius of the two: a destructive, silent mutation of state that isn't even yours.
2. **`bin/tm/commands/pm_guard_deny_by_default.rs:61,~206-230`** — the opt-in (`TRUSTY_MPM_PM_DENY_BY_DEFAULT`) persona-confirmation gate resolves "the current pane's session" the same way, to check whether ITS deployment failed. Lower blast radius (fails OPEN on ambiguity, per its #2172 permissive-default design), but it provably reads the wrong session's state in the same scenario, which can let it wrongly gate (or wrongly not-gate) a delegated edit based on a stranger session's `errored` state.

## Relation to other open identity issues (NOT duplicates)

[...]

Issue #2604, opened 2026-07-14
pm_guard blocks PM single-file writes — session-pause snapshot denied (P1 over-enforcement)


bobmatnyc:
## Summary

The trusty-mpm PM `PreToolUse` guard (`tm hook --pm-guard`, `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs`) hard-blocks **all** PM `Write`/`Edit`/`MultiEdit`/`NotebookEdit` tool calls unconditionally. This over-enforces prohibition P1 (which is meant to stop the PM implementing *code*, not to block one-off orchestration/state writes).

## Repro (live dogfooding, Bob / repo owner)

The PM ran `/tm-session-pause` and tried to write its own session snapshot:

```
Write(.trusty-mpm/sessions/session-20260714-064556.md)
→ Error: PM must not edit files directly (prohibition P1). Delegate the change to rust-engineer via the Task/Agent tool.
```

The PM cannot even persist its own orchestration state (session snapshots under `.trusty-mpm/`), which the session-pause skill requires.

## Directive (Bob)

> "the PM can write single files."

The intent of P1 is to stop the PM doing **bulk code implementation**, not to block one-off file writes to PM-owned orchestration state, docs, or config.

## Fix

1. **HARD:** never block PM `Write`/`Edit` of PM-owned orchestration state — at minimum anything under `.trusty-mpm/**`, the PM memory directory, and `TASK.md`-style session artifacts. The snapshot write above must pass, with a regression test on that exact path.
2. Permit single-file writes to non-source files generally; keep **source-code** implementation (`.rs`, `.py`, `.ts`, …) blocked so the "delegate implementation to rust-engineer" intent survives.
3. Update specs/docs/tests asserting the old always-deny behavior.

The `TRUSTY_MPM_PM_UNRESTRICTED=1` escape hatch and native sub-agent (`agent_id`) exemption remain unchanged.

PR #4060, opened 2026-07-27
fix(trusty-mpm): daemon-held pm_guard kill-switch flags, closing the settings.json self-exemption hole

### LINKED ISSUE #3981: trusty-mpm: the PM can self-exempt from ALL pm_guard enforcement by writing .claude/settings.json — the "operator-only" escape hatches are agent-writable
## Summary
`pm_guard`'s two escape hatches, `TRUSTY_MPM_DISABLE_HOOKS` and `TRUSTY_MPM_PM_UNRESTRICTED`, are documented and designed as operator-set kill switches — something a human sets deliberately when the guard misfires. In practice the PM itself, or any Guard-4-exempted native subagent, can set them without any human involvement, in a single ordinary write that `pm_guard` already permits. Once set, ALL `pm_guard` enforcement is disabled: every P1-P11 rule, the budget system, and the worktree-tmp hard block added in PR #3978.

## Mechanism (verified empirically against the built binary)
1. Claude Code's `settings.json` supports a top-level `env` object, applied to every session and to every subprocess Claude Code spawns from it — hooks included. It is live-reloaded mid-session, no restart needed.
2. `.claude/settings.json` has extension `json`, which is NOT in `SOURCE_CODE_EXTENSIONS` (`crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs:145-151`), and the path is not covered by `is_pm_orchestration_path`.
3. Consequently `evaluate_edit_tool` (`pm_guard.rs:445-454`) falls through to an unconditional, **unbudgeted** ALLOW for a direct `Write`/`Edit` targeting that file.
4. A `Write` to `.claude/settings.json` whose content sets `env.TRUSTY_MPM_PM_UNRESTRICTED=1` produces empty stdout (ALLOW) from the real `tm hook --pm-guard` binary — confirmed, not inferred.
5. Guards 2/3 (`pm_guard.rs:196-202`) then short-circuit every subsequent invocation.

A workspace grep for programmatic assignment of those variables finds nothing — which is what led to the "operator-only" claim — but that search covers `std::env::var` calls in Rust source and does not cover the settings.json `env` channel.

## Impact
This is the trust anchor of the guard being writable by the thing the guard constrains. It is not a narrow bypass of one rule; it disables the entire mechanism, persistently, and silently. Note the practical severity depends on your threat model — this guards against an agent going off-script, not against a determined adversary — but the current documentation asserts a property that does not hold, which is the immediate problem.

## Status: pre-existing, not introduced recently
Guard 3 (`PM_UNRESTRICTED_ENV`) predates PR #3978 by many issues. #3978 neither introduced nor widened this. It surfaced because #3978's round-2 design deliberately left Guards 2/3 unpierced on the stated grounds that they are human-only — a premise this finding falsifies. #3978 is being merged with its documentation corrected to stop asserting that premise and to point here.

## Suggested remedy (decision required, not proposed as settled)
Add `.claude/settings.json`, `.claude/settings.local.json`, and the equivalent global `CLAUDE_CONFIG_DIR` path to a denylist checked in `evaluate_edit_tool` BEFORE the "non-source file ⇒ allow" fallback, so a direct write to the guard's own trust anchor is denied or routed through delegation the way source-code edits are.

Considerations for whoever picks this up:
- There are legitimate reasons to edit `.claude/settings.json` (adding permissions, hooks, MCP config). A blanket deny may be too blunt; denying only writes that touch the `env` key would be more surgical but requires parsing rather than path-matching, and a partial write could still slip through.
- Any denylist here is itself only as strong as the file's own protection — if an agent can reach the file by another route (a shell redirect via Bash, a different tool), the path check alone will not hold. Worth checking whether Bash-mediated writes to that path are already covered before assuming a path denylist closes it.
- Consider whether the same exposure applies to other trust-anchor files under `.claude/`.

See also: #3977, PR #3978, #3973.

bobmatnyc:
## Summary

Closes #3981. **Supersedes PR #3994**, which abandoned the Bash-command-classifier approach after four adversarial review rounds each found a trivial new bypass into `.claude/settings.json` (`cp`/`mv`/`install` → `ln`/`git mv`/`xargs` → `rsync`/`scp` → brace expansion). Scanning Bash commands for writes to one trust-anchor file fails open by construction — there is always another way to write a file from a shell. Bob's call: eliminate the trust anchor instead of defending it. **PR #3994 was never merged** (confirmed via `git merge-base --is-ancestor`), so there is nothing to delete on this branch — the "Part 1: delete the trust-anchor apparatus" step of the original plan is a no-op here.

## Root cause (#3981)

`pm_guard`'s Guards 2/3 (`TRUSTY_MPM_DISABLE_HOOKS` / `TRUSTY_MPM_PM_UNRESTRICTED`) read LIVE process env on every `PreToolUse` call. Claude Code re-spawns the hook subprocess fresh each time and re-injects `.claude/settings.json`'s top-level `env` object into it (live-reloaded mid-session). A `.json` file isn't in `SOURCE_CODE_EXTENSIONS`, so a plain PM `Write`/`Edit` to that file fell through `evaluate_edit_tool` to an unconditional ALLOW — letting the PM (or any Guard-4-exempt subagent) self-exempt from the **entire guard** with one file write, no restart, no observable disruption.

## Fix

- **Capture once, at launch/resume time, from the operator's shell.** `tm sessions new`/`start`/`resume` read `TRUSTY_MPM_DISABLE_HOOKS`/`TRUSTY_MPM_PM_UNRESTRICTED` from the CLI process's OWN environment and persist them on the managed session's daemon record (`session_manager::guard_flags::SessionManager::set_guard_flags`, `crates/trusty-mpm/src/session_manager/guard_flags.rs`).
- **Resolve via a daemon round trip, not live env.** New endpoint `GET /api/v1/sessions/managed/{id}/guard-flags` (`daemon/managed_routes/guard_flags.rs`); `pm_guard`'s new `pm_guard_session_flags` module (`bin/tm/commands/pm_guard_session_flags.rs`) resolves the flags by `TM_MANAGED_SESSION_ID`, modeled directly on the existing `pm_guard_deny_by_default::persona_status_for_session` async-daemon-round-trip precedent (same tight 500ms-connect/2s-total client budget, same `EnvSessionIdentity` pane cross-check for the #3600 lesson — a sibling pane inheriting the same env var must never be trusted).
- **One deliberate inversion from that precedent:** `persona_status` fails OPEN (permissive) because it's an ordinary ALLOW/DENY policy check. Resolving a KILL-SWITCH state must fail TOWARD THE GUARD STAYING ACTIVE — an unreachable daemon, unresolvable session, or unconfirmed pane all resolve to the flags staying `false`/`false`. A killed/paused daemon must never become the new bypass.
- **No mid-session flip.** `pm_guard` never writes these flags, only reads. Disabling the guard now requires an actual session restart with the flag set — an operator-initiated act, matching `core::project_trust`'s consult-once-at-launch precedent (not a file-permission boundary). `ResumeQuery`'s doc explains why an ordinary bodyless resume (GUI, picker) deliberately RE-ARMS the guard rather than leaving a prior bypass silently sticky.
- **Narrows the `.trusty-mpm`/`.trusty-tools` write allowlist** (`is_guard_state_path`, `pm_guard.rs`): pm_guard's own turn-budget counter (`<root>/state/pm_guard_turn_budget/<session>.json`) and `core::project_trust`'s store (`<root>/project-trust.json`) are no longer PM-writable via a plain Edit/Write — closing the "PM can edit its own enforcement inputs" gap from issue #2604's orchestration exemption. Ordinary orchestration writes (session snapshots, `TASK.md`, the memory palace) are unaffected. ~15 lines of actual logic.

## Scope note

This covers **managed sessions** (`tm sessions new`, the git-repo-backed `session start` path, and `tm sessions resume`) — the dominant real-world path, since the bare-`tm` guided default always routes a GitHub-backed project through the managed session-manager. A **local, non-managed** `tm sessions start` (a directory with no recognized git remote) has no session-identity env var exported into its pane at all, so the guard is unconditionally fully active there. This is never a regression — no session, managed or local, could disable the guard via a live env var before this change either. I deviated here from the task brief's literal pointer to `daemon/services/session_service.rs`/`SessionService::resolve` (the LOCAL, non-managed session store) after tracing the actual identity plumbing: that store's `Session` records are auto-registered by Claude's own `SessionStart` hook, keyed by a UUID unknowable at `tm sessions start` time, which would have required a new cwd-correlation subsystem to bridge — a lot more code than reusing the MANAGED store's already-live `TM_MANAGED_SESSION_ID` + pane-identity infrastructure. Flagging this explicitly for review.

## Net lines

**+978 net** (1123 insertions / 145 deletions across 52 files), not the "-1400" estimated in the task brief — because PR #3994 was never merged, Part 1 had nothing to delete. This PR is net-new functionality (daemon plumbing for 2 new record fields + a new endpoint + the allowlist narrowing), not a net reduction. Flagging this discrepancy explicitly since it's a real deviation from the stated target.

## Verification (raw output, live isolated daemon + real tmux pane — see PR discussion / session log for full transcripts)

1. **Settings.json write no longer disables the guard.** A `Write` to `.claude/settings.json` (even one setting `env.TRUSTY_MPM_PM_UNRESTRICTED=1`) is still ALLOWED (unblocked, matching "the write may now succeed") — but the injected env var has **zero effect**: a subsequent `make build` with `TRUSTY_MPM_PM_UNRESTRICTED=1`/`TRUSTY_MPM_DISABLE_HOOKS=1` set directly on the hook subprocess is still DENIED.
2. **Flags DO take effect at resume time.** Live daemon + a real tmux session: adopted a throwaway pane (flags default `false`/`false`), stopped it, `POST .../resume?disable_hooks=true&pm_unrestricted=true` → `GET .../guard-flags` confirms `true`/`true` persisted with the new pane's `pane_id`. Running `tm hook --pm-guard` from literally inside that pane (`TM_MANAGED_SESSION_ID` set, `make build` payload) → **ALLOWED** (empty stdout). Running the identical command from a non-matching context (this session, not inside that tmux pane) → **DENIED**, proving the pane cross-check.
3. **Guard-active-on-daemon-unreachable, confirmed live.** Killed the isolated daemon; re-ran the exact same matching-pane command that previously ALLOWED → now **DENIED** (`"PM must not run builds or tests directly..."`), even though the (now-unreachable) daemon's last-known record said `true`/`true`.
4. **Budget counter / project-trust no longer PM-writable; ordinary orchestration still is.** `Write` to `.../project-trust.json` → DENIED (new `GUARD_STATE_EDIT_REASON`). `Write` to `.../pm_guard_turn_budget/sess-1.json` → DENIED. `Write` to `.trusty-mpm/sessions/session-verify.md` → ALLOWED (empty output).
5. **Worktree-tmp guard untouched, still denies.** `git worktree add /tmp/verify-3981` → DENIED with the original #3977 message, unchanged.

## Quality gate

- `cargo test -p trusty-mpm` (no `--lib`, full `tests/` suite included): **3614 lib + all integration suites, 0 failed** (6 ignored — live-tmux/live-git only, matches existing convention; ran from a non-`/tmp` checkout so the 2 known-environment-sensitive tests are included and green).
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: clean.
- `cargo fmt -p trusty-mpm --check`: clean.

## Test plan

- [x] `cargo test -p trusty-mpm` (full suite, no `--lib`) — green
- [x] `cargo fmt -p trusty-mpm --check` — clean
- [x] Live verification of all 5 checklist items above against a real isolated daemon + tmux pane


bobmatnyc:
## Verdict: BLOCK

Adversarial review of the daemon-held kill-switch redesign (`pm_guard_session_flags`/`guard_flags`), given this replaces a predecessor (#3994) that a critic BLOCKed four times for exactly this shape of gap: a policy that looks closed on the classifier's own tests but has a live write-path or coverage hole somewhere else in the call graph.

The core resolution primitive is genuinely solid: `guard_flags_for_session` (`crates/trusty-mpm/src/bin/tm/commands/pm_guard_session_flags.rs:104-125`) correctly fails toward guard-active on a missing session id, an unreachable daemon, and an unconfirmed pane (reusing the well-tested `EnvSessionIdentity` #3600 precedent unchanged), and its own unit tests exercise all four fail-closed branches plus the happy path. That part is not where the problem is. The problem is in how the flags get **written**, and in the fact that Part 3's new file-protection claim isn't actually enforced on the Bash path. Both are CRITICAL because they reproduce the exact "verb/route the scanner didn't cover" failure mode that blocked #3994 four times — just relocated.

## Findings

| CRITICAL | `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs` | 237-247, and reconnect returns at 415 / 458 (also 396-429 launch-on-main branch) | `spawn_managed` captures `guard_flags = (params.disable_hooks, params.pm_unrestricted)` from the *invoking* shell, but when `spawn_managed_routed` takes the #1707/#3455 **reconnect** branch it returns the *existing* live `SessionRecord` directly (`return Ok(live);`) without ever touching its `disable_hooks`/`pm_unrestricted`. Back in `spawn_managed`, `persist_after_spawn(state, &mut record, guard_flags.0, guard_flags.1)` (`guard_flags.rs:104-112`) then **no-ops** whenever both requested flags are `false` (the ordinary case — no bypass env var set in the new shell). Net effect: if a session was ever granted `disable_hooks=true`/`pm_unrestricted=true` (an operator ran `TRUSTY_MPM_PM_UNRESTRICTED=1 tm sessions new` once for a repo), and that session is still `Active` + tmux-alive, then a completely ordinary, later `tm sessions new`/`tm session start` against the SAME repo from a clean shell (no bypass var set) reconnects to it (`force_new` defaults `false`) and **silently leaves the guard fully disabled** — with no bypass requested on this call, no visible warning, and no test covering it. This is precisely the "sticky bypass" shape `ResumeQuery`'s own doc (`guard_flags.rs:130-144`) says must never happen — but that discipline was only applied to the resume handler (`persist_after_resume` is unconditional), not to the reconnect-on-spawn path, which the PR's own description calls "the dominant real-world path" (in-project `tm session start`). No test in this PR's diff exercises reconnect + guard flags together (`managed_routes/tests.rs`'s new tests only cover a fresh `create()` + explicit `set_guard_flags`, never a reconnect). | Make `persist_after_spawn` unconditional (mirror `persist_after_resume` — always call `set_guard_flags`, dropping the "skip when both false" optimization), OR explicitly detect the reconnect case and re-assert `params.disable_hooks`/`params.pm_unrestricted` onto the reconnected record every time, exactly as resume already does. Add a regression test: create a session with flags `true`/`true`, trigger a reconnect-eligible spawn with flags `false`/`false`, assert the record's flags end up `false`/`false`. |
| CRITICAL | `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs` | 433-448 (`evaluate_tool`), 468-486 (`evaluate_edit_tool`/`is_guard_state_path`) vs. `pm_guard_bash/mod.rs:69-101` (`evaluate_bash_command`) | Part 3's new `GUARD_STATE_EDIT_REASON` protection for `project-trust.json` and `pm_guard_turn_budget/*.json` (`is_guard_state_path`, `pm_guard.rs:502-513`) is wired ONLY into `evaluate_edit_tool`, which `evaluate_tool` reaches only for `EDIT_TOOLS` (Edit/Write-shaped tool calls, `pm_guard.rs:437-439`). A `Bash` call is routed instead to `evaluate_bash_command` (`pm_guard.rs:440-445`), whose only file-write detection is `has_file_write_redirection` (pure `>`/`>>` scanning, `pm_guard_bash/mod.rs:441-480`) plus a short explicit verb list (`patch`, `sed`/`awk` when not read-only, `curl`/`wget`, `make`/`pytest`, `git apply` — `pm_guard_bash/mod.rs:186-207`). None of `cp`, `mv`, `ln`, `install`, `rsync`, `scp`, `xargs` appear anywhere in this module (verified: zero matches for any of those verbs in `pm_guard_bash/`). So `Bash({"command": "cp /tmp/x ~/.trusty-tools/trusty-mpm/project-trust.json"})` is **not denied by anything** — `evaluate_bash_command` returns `None` (ALLOW). This lets the PM forge `project-trust.json` directly — self-granting the project-level MCP-trust decision that `core::project_trust` is supposed to gate as an explicit, once-at-launch operator act (per project memory, this is the exact trust anchor a prior finding called an "RCE-on-clone" risk when derived automatically). It also lets the PM reset/forge its own `pm_guard_turn_budget` counter, defeating the budget gate entirely. This is the SAME class of gap (a Bash-command classifier with an incomplete verb list) that BLOCKed the predecessor PR four consecutive rounds for `cp`/`mv`/`install`→`ln`/`git mv`→`rsync`/`scp`→brace-expansion against `settings.json` — reproduced here, unnoticed, against the two NEW files this PR designates as "enforcement input, not PM orchestration output." The PR's own test (`evaluate_tool_denies_guard_state_writes`, `pm_guard.rs:752-772`) only exercises the `Write` tool path, so this gap is untested and would not have been caught by the green CI run. | Either (a) extend `has_file_write_redirection`'s scope / add a target-path extraction for `cp`/`mv`/`ln`/`install`/`rsync`/`scp` (mirroring the existing `extract_shell_edit_target` machinery) and route it through `is_guard_state_path` from inside `evaluate_bash_command`, or (b) simplest and most robust given this exact history: deny `cp`/`mv`/`ln`/`install`/`rsync`/`scp` outright as classified verbs (unconditionally, the same way `patch` is unconditionally denied) rather than trying to enumerate every write vector one at a time — the PR's own summary explicitly rejects "scanning Bash commands for writes to one file" as failing open by construction; the same conclusion applies here. |
| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs` | 238-247 (spawn) / `mod.rs:812-838` (resume) | The guard-flags write (`persist_after_spawn`/`persist_after_resume`) happens AFTER `adapter.spawn`/`adapter.spawn_resume` has already been invoked (i.e., after the runtime process is told to start), not before, contradicting `pm_guard_session_flags.rs`'s own module doc claim that flags are "captured once... before this session's PM process exists" (line 11 of that file). In practice the window is negligible (a local in-process async write completes in microseconds vs. multi-second `claude` process boot), so this is not separately exploitable — but it does mean a daemon crash/restart occurring in the narrow gap between the two awaits in the same request handler would leave the record holding whatever flags existed *before* this spawn/resume call, not the ones this call declared, which is a durability gap in the "every resume re-declares, nothing is sticky" invariant the PR is built around. | Move the `set_guard_flags`/`persist_after_*` call to before the `adapter.spawn`/`adapter.spawn_resume` call, so the record reflects the intended flags before the runtime that will read them is told to start. |

## Required Changes

1. Make guard-flag persistence unconditional on the spawn path (or explicitly handle the reconnect branch), so an ordinary `tm sessions new`/`session start` call without a bypass env var always resets a reconnected session's `disable_hooks`/`pm_unrestricted` to what THIS call requested — never inherits a stale grant. Add a reconnect+guard-flags regression test.
2. Close the Bash-path gap for `is_guard_state_path` — either route `cp`/`mv`/`ln`/`install`/`rsync`/`scp` targets through it, or deny those verbs outright from `pm_guard_bash`. Add a test asserting a `Bash` `cp`/`mv` onto `project-trust.json` and `pm_guard_turn_budget/*.json` is denied.
3. (Lower priority, ship-blocking only if 1/2 need a respin anyway) Reorder guard-flag persistence to happen before the runtime is spawned/resumed, matching the module's own documented invariant.

## Notes

- The kill-switch resolution primitive itself (`pm_guard_session_flags::guard_flags_for_session`, the pane cross-check reuse, the fail-toward-guard-active defaults) is well-designed and well-tested; I found no issue with it in isolation. Both CRITICAL findings are in the lifecycle/write-path integration and in Part 3's scope-completeness, not in the daemon-round-trip design itself.
- Both findings are, in spirit, the same lesson the predecessor PR's four rounds already taught: a security-relevant guard that is provably correct against its own unit tests can still have a hole at the integration seam (here: the #1707 reconnect shortcut, and the Edit/Write-vs-Bash routing split) that its test suite never exercises. Green CI on this PR does not cover either finding.
- I did not find a way to spoof `session_id`/pane identity itself (the #3600 cross-check reuse holds), and I did not find a daemon-unreachable case that resolves to a bypass — those two specific asks from the review brief check out.

bobmatnyc:
Closed as won't-do by owner decision (2026-07-28).

The approach went five BLOCK review rounds across two different implementations without converging. The blocker was never a fixable code defect — it was that the design itself was not agreed.

Reopening should start from a fresh problem statement, not from this branch.

Issue #2611, opened 2026-07-14
trusty-search: managed-repo awareness — canonical repo identity + base/live tracking + worktree delta indexing (design)


bobmatnyc:
## Motivating directive

Filed per Bob's directive (2026-07-14): since tm clones managed repos into a
protected directory, trusty-search should be aware of and handle the
different facets of a repo — the main (live) project, the protected (`.base`)
project, and the worktree — tracking both live and base, and indexing any
differences between them, rather than treating each as an unrelated
directory.

A **formal spec (DOC-37)** covering this design is being drafted in a
companion PR: `docs/specs/trusty-search-managed-repo-awareness.md`. This
issue captures the same findings + design for tracking/discussion; the PR
link will be added as a comment once opened.

---
## Motivating directive (Bob, verbatim)

> since we clone tm managed repos into a protected directory, let's make sure
> trusty-search is aware and knows how to handle indexes. E.g., the main
> project, the protected project, and the worktree. In fact it should probably
> keep track of both, and index any differences.

Background: trusty-mpm (`tm`) provisions managed workspaces. For a live user
checkout (e.g. `~/Duetto/repos/duetto`), tm creates a protected managed clone
at `~/trusty-mpm-projects/<org>/<repo>/` (a `.base` clone); sessions then get
git worktrees under `.base/.worktrees/<session-id>/` (a `.trusty-mpm-worktree`
marker file exists at worktree roots). So the SAME repo content can exist in
up to three places — **live checkout**, **protected `.base` clone**, and
**N session worktrees** — and trusty-search currently treats every one of
these as a wholly unrelated directory.

---

### 1.1 Index identity is a bare path basename, not a repo identity

`trusty-common::derive_index_id` (`crates/trusty-common/src/index_id.rs:75-82`)
returns `project_root.file_name()` verbatim (not slugified, preserved
byte-for-byte for backward compatibility). `resolve_project_root`
(`crates/trusty-common/src/index_id.rs:46-58`) walks up from a start path to
the nearest ancestor containing a `.git` entry — its own doc comment notes
`.git` is a directory in a normal clone and a **file** in a worktree, and
`exists()` matches both, "so worktrees resolve correctly." But "correctly"
here only means the walk terminates at the *worktree's own* `.git` file — it
does **not** walk through the `gitdir:` pointer to the shared repository, so
each worktree resolves **itself** as the project root, not the shared parent.

Both callers agree on this (single source of truth, issue #1373):
[...]

PR #3995, opened 2026-07-26
fix(trusty-search): close reindex-override and cold-restore root_path collision gaps

bobmatnyc:
## Summary

Fourth occurrence of the #2305/#2336 `DatabaseAlreadyOpen` class, found by an audit prompted by #3929 ("a regression of #2305/#2336"). Two index-registration paths bypassed the existing `find_root_path_collision` guard (issue #2336):

- **Gap E** — `POST /indexes/:id/reindex` with a `root_path` override (`reindex_handlers.rs`) registered the new root with no collision check at all, silently letting one index steal another live index's `root_path`.
- **Gap F** — `find_root_path_collision` scans only LIVE handles, so it is structurally blind to a cold (unloaded) entry parked in `state.cold_store`. A `create_index` whose root matched an unloaded cold entry's root passed the guard cleanly; when that cold entry was later triggered by its first query, `restore_index_on_demand` (`lazy_restore.rs`) opened the SAME on-disk redb with no guard at all — a third source of `DatabaseAlreadyOpen`.

Both fixes reuse the existing `find_root_path_collision` primitive (widened from `pub(super)` to `pub(crate)` so `lazy_restore` — outside `service::server` — can reach it) rather than adding a fourth collision mechanism, per the issue's explicit scope constraint.

Per the issue: **not gated on #1681 or #2611** — #2611 is display/grouping identity, not collision safety, and #1681 is unstarted.

## Gap E — before/after

- **Before**: `reindex_handler` computed the canonical override root and called `state.registry.register(new_handle)` unconditionally — no check against sibling live handles.
- **After**: right before that register call, `find_root_path_collision(&live_handles, &new_root, Some(&index_id))` runs; a collision returns `409 Conflict` with `existing_id`, and the target index's `root_path` is left untouched.
- **Test**: `reindex_root_override_rejects_collision_with_live_sibling` (`crates/trusty-search/src/service/server/tests_3993.rs`) registers two bare live handles at distinct roots, then reindexes one with a `root_path` override pointed at the other's root.
  - **Without the fix**: fails — handler returns `200 {"queued": true}` and silently re-points `index-b` onto `index-a`'s root (two live ids sharing one on-disk corpus).
  - **With the fix**: passes — `409` naming `index-a`, both roots unchanged.
- Non-regression: `reindex_root_override_distinct_root_succeeds` proves the guard doesn't over-trigger on a genuinely unclaimed root.

## Gap F — before/after (first round)

[...]

Issue #3955, opened 2026-07-26
Agent/review worktrees provisioned under /tmp silently fail 12 trusty-search tests via the SENSITIVE_PATH_PREFIXES denylist


bobmatnyc:
## Summary
Any `cargo test -p trusty-search` run from a worktree located under `/tmp` or `/private/tmp` deterministically fails ~12 tests in the `create_index_*` / `relocate_index_*` family with `left: 400, right: 200` (one embedder-failure test shows `right: 503`, the same denylist gate firing before the embedder check). This has been misattributed as an "unexplained flaky test" in this repo for some time. It is not flaky and it is not nondeterministic.

## Root cause
`SENSITIVE_PATH_PREFIXES` in `crates/trusty-search/src/allowlist/mod.rs:102-111` lists both `/tmp/` and `/private/tmp`. Tests that create an index rooted under those paths are correctly rejected with HTTP 400 by the denylist. The tests were written assuming the worktree lives somewhere allowlisted.

## Decisive evidence
Same commit (`e7673e77`), two worktrees, identical `cargo test -p trusty-search --lib`:
- Worktree at `/private/tmp/wt-3706-warmboot-health`: **1312 passed; 12 failed**
- Worktree outside `/tmp` (`~/trusty-mpm-projects/review-scratch/wt-3706-nontmp`): **1324 passed; 0 failed; 1 ignored** (the 1 ignored is a pre-existing `redb_migrate` test gated on a machine-specific backup file, unrelated)

Reproduced independently twice: once by the engineer implementing #3706, once by the adversarial critic reviewing PR #3953.

## Why this matters beyond one test
This tooling provisions agent and review worktrees under `/tmp` by default. Every agent that runs the trusty-search test suite therefore sees ~12 failures that are not its own, which:
- invites agents and reviewers to dismiss red results as "flaky" — eroding the credibility of the test gate generally
- can mask a genuine regression hiding among the expected-red set
- wastes investigation time repeatedly (this has now been re-investigated at least three times)

Note GitHub-hosted CI does NOT run from `/tmp`, so this affects local/agent runs, not hosted CI.

## Options — NOT a decision, requires owner input
1. Stop provisioning agent/review worktrees under `/tmp` (e.g. a directory under the projects root or `$HOME`). Fixes it for every crate at once and requires no test changes; is a harness-level change.
[...]

Issue #2784, opened 2026-07-16
feat(trusty-code): surface/warm trusty-search index readiness for daily-driver sessions


bobmatnyc:
Cold-indexing a large repo has a real semantic-warm-up window (lexical ready ~24s; semantic embedding InProgress 20+ min on a 50K-chunk monorepo via trusty-embedderd). A daily-driver tcode session that starts working immediately will search before semantic is ready. Consider: a readiness/status check surfaced to the session, a brief wait-for-Stage-2 on first search, or a proactive index-warming indicator. Evidence: daily-driver probe 2026-07-15. Complements issue #1 (fallback) — this addresses proactively getting to ready; #1 addresses graceful behavior until then.

PR #4063, opened 2026-07-27
fix(trusty-search): derive index id from git org/repo, alias-fallback for compat

### LINKED ISSUE #4062: trusty-search: derive index id from git org/repo, not bare basename (alias-fallback for compat)
## Problem

When trusty-search indexes a trusty-mpm- or trusty-code-managed project, the
index id is derived from the bare directory basename
(`trusty_common::derive_index_id`, `crates/trusty-common/src/index_id.rs`),
verbatim, "preserved for backward compatibility". This means two unrelated
checkouts that happen to share a directory name (or the same repo checked out
under a different local folder name) do not share/relate an index id, and the
id carries no information about which actual GitHub project it is.

Git-origin -> `org/repo` parsing already exists and is well-tested:
`trusty_common::github_path::{derive_github_path, parse_github_path}` and
`trusty_common::repo_identity::RepoIdentity::{derive, canonical, parse}`
(GitHub-owner/repo with a content-hash fallback for remote-less repos). This
was wired in as an ADDITIVE join-key field (`repo_identity` on
`PersistedIndex`) per the DOC-37 MVP (#2611, closed) — but that work
deliberately did NOT touch the actual `index_id`/name derivation.

## Ask (scoped, narrower than #2611 section 2.2-2.3)

- `derive_index_id` (or an added resolution path alongside it) should prefer
  `RepoIdentity::canonical()`'s `org/repo` form as the *new default* id for
  **newly-registered** indexes, when the project's git origin remote actually
  resolves to a GitHub-style identity. Fall back to today's bare-basename
  behavior when there is no remote (content-hash-only repos).
- Add an alias-fallback at lookup/auto-detect time: if the org/repo-form id
  misses, fall back to trying the legacy bare-basename id (and vice versa)
  BEFORE creating a brand-new index. This avoids any bulk on-disk migration
  of `indexes.toml` / colocated `.trusty-search/` storage, and keeps every
  existing index and every session's pinned `--index <id>` `.mcp.json` stub
  working unchanged.

## Non-goals

- No bulk rename/migration of existing on-disk indexes.
- No change to the `repo_identity` join-key field itself (already shipped).
- Not folding into #2611 — this is a distinct, narrower ask; referencing it
  as the parent design doc.

## References

- `crates/trusty-common/src/index_id.rs`
- `crates/trusty-common/src/github_path.rs`
- `crates/trusty-common/src/repo_identity.rs`
- `crates/trusty-common/src/search_index.rs` (`ensure_project_indexed`)
- `crates/trusty-search/src/detect.rs` (`detect_project`)
- Parent design doc: #2611 (DOC-37)

bobmatnyc:
## Summary

- `trusty_common::derive_index_id` always returned the bare directory
  basename as the trusty-search index id, which carries no information about
  which actual GitHub project a checkout is, and collides across unrelated
  checkouts that share a directory name.
- New `derive_preferred_index_id` (`crates/trusty-common/src/index_id.rs`)
  prefers a `RepoIdentity`-derived `owner-repo` token when the project's git
  origin remote resolves, falling back to the legacy basename unchanged
  otherwise (no remote / content-hash-only identity).
- **Deliberate deviation from a literal `RepoIdentity::canonical()`
  (`"owner/repo"`) id:** every `trusty-search` HTTP route
  (`/indexes/{id}/status`, `/search`, …) matches `{id}` against exactly one
  URL path segment (axum/matchit routing) — an embedded `/` would silently
  break routing across the whole REST/MCP surface. The preferred id instead
  joins owner/repo with a single hyphen (`"owner-repo"`), mirroring the same
  precedent `trusty_common::palace_id::owner_repo_from_git_remote` already
  uses for trusty-memory's analogous palace-id problem.
- New `search_index::resolve_effective_index_id` adds the alias-fallback:
  when the preferred and legacy ids differ, it probes the daemon
  (`GET /indexes/{id}/status`) for the legacy id FIRST — an index already
  registered there wins, so it's reused rather than shadowed by a fresh empty
  index — then the preferred id; a brand-new project (or unreachable daemon)
  defaults to the preferred id. This keeps every existing index and every
  session's pinned `--index <id>` `.mcp.json` stub working unchanged.
- Wired into `ensure_project_indexed` (trusty-mpm session launch /
  trusty-code task start), its incremental per-file update path
  (`index_files_best_effort`), and trusty-search's own `detect_project`
  auto-detect (`detect.rs:51/62/73`) — enabled the `search-index` feature on
  trusty-search's `trusty-common` dependency for this.

Closes #4062 (scoped follow-up to the DOC-37 design doc, #2611 — this PR does
not touch the `repo_identity` join-key field itself, which already shipped).

## Test plan

- [x] `cargo test -p trusty-common -p trusty-search` — full test binaries,
      all green (see gate output below)
- [x] New unit tests: `derive_preferred_index_id_uses_org_repo_when_remote_resolves`,
      `derive_preferred_index_id_falls_back_to_basename_without_remote`,
      `resolve_effective_index_id_returns_single_id_when_no_remote`,
      `resolve_effective_index_id_prefers_existing_legacy_index`,
      `resolve_effective_index_id_defaults_to_preferred_when_daemon_down`
- [x] `cargo clippy -p trusty-common -p trusty-search --all-targets` — clean
- [x] `cargo fmt --check -p trusty-common -p trusty-search` — clean


bobmatnyc:
## Verdict: BLOCK

Adversarial review of the diff (`gh pr diff 4063`), traced against the actual
call graph in `crates/trusty-common/src/{index_id,repo_identity,github_path,search_index,search_readiness}.rs`,
`crates/trusty-mpm/src/session_manager/search_gc.rs`, `crates/trusty-search/src/{detect,main}.rs`,
and `crates/trusty-review/src/report/analyze_adapter.rs`. Not relying on the PR's own test suite's framing — traced the logic directly and reproduced one finding empirically against this very checkout.

## Findings

| CRITICAL | `crates/trusty-common/src/index_id.rs` | 95-102 (`derive_preferred_index_id`) | `RepoIdentity::derive` resolves `remote.origin.url` by running `git config` in `project_root` — which, for a git **worktree**, transparently reads the *shared* repo config (same commondir), so it returns the identical `GitHub(owner, repo)` for the main checkout, its `.base` bare-clone-derived worktrees, and every session worktree of the same origin. I verified this empirically against the actual repo this PR lives in: running `git config --get remote.origin.url` from this session's own worktree (`/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.base/.worktrees/2eb72dca-...`) returns `git@github.com:bobmatnyc/trusty-tools.git` — the exact same URL the main checkout and the `.base` clone would return. `derive_preferred_index_id` would therefore return `"bobmatnyc-trusty-tools"` for **every** worktree of this repo, not a distinct id per worktree. Since the daemon's registry is one-`root_path`-per-`id` (see `search_gc.rs`'s `IndexDetailRow{ id, root_path }` — a single `root_path` field, not a list), a second worktree's `ensure_project_indexed`→`best_effort_create_index` POST against an already-registered `"bobmatnyc-trusty-tools"` id is a no-op **find**, not a create — the id keeps pointing at whichever worktree/checkout registered it first. Every other worktree's `resolve_effective_index_id` then also resolves to that same id (the alias-probe finds it "registered" and returns it), so its `index_files_best_effort` incremental updates and `search`/`grep` queries all target **one shared index rooted at a different worktree's directory** — cross-session content contamination between concurrently-running, independently-checked-out session worktrees of the same repo. This is exactly the isolation model `RepoIdentity`'s own module doc (`repo_identity.rs:1-19`) says it must NOT be used for on its own: it's documented as an additive **join key** to relate multiple distinct indexes of the same repo, not a replacement for the per-directory id that keeps them distinct. None of the PR's new tests (`derive_preferred_index_id_uses_org_repo_when_remote_resolves`, `resolve_effective_index_id_prefers_existing_legacy_index`, etc.) exercise two roots sharing one origin remote — every fixture is a single isolated temp repo, so this collapse is untested. | `derive_preferred_index_id` must not promote the org/repo id for a worktree-scoped root. Gate it behind the same "is this a persistent project root, not a session worktree" check `search_gc.rs::is_session_worktree` / `is_orphan_index` already use: if `project_root` (after `resolve_project_root`) is `.worktrees/<leaf>`-scoped (or more generally, is not the canonical/first-registered facet), keep returning `derive_index_id`'s basename unchanged, and rely on the *already-shipped, additive* `repo_identity` field (#2611) to relate those distinct per-worktree indexes back to the same repo instead of merging their ids. |
| HIGH | `crates/trusty-common/src/search_readiness.rs` | 168 | `probe_index_readiness` still calls `crate::derive_index_id(&root)` directly. Its own doc comment (line ~155) claims it "resolves the canonical `(root, index_id)` the way `ensure_project_indexed` does (so this always targets the same index the warming path created)" — that claim is now false: `ensure_project_indexed` was changed in this PR to call `resolve_effective_index_id`, but this function was not updated. For any project whose git origin resolves and has no pre-existing legacy-basename index, `ensure_project_indexed` registers/warms the index under the **preferred** org-repo id, while `probe_index_readiness` polls `GET /indexes/{legacy_basename}/status` — an id nothing was ever registered under — which 404s/misses and silently returns `None` (fail-open design swallows it). The #2784 readiness-warning feature goes silently dark for every project that migrates to the new id scheme. | Change line 168 to `crate::search_index::resolve_effective_index_id(&root)` (this crate already owns both modules, no new feature-flag issue since `search_readiness` and `search_index` are both `search-index`-gated). |
| HIGH | `crates/trusty-mpm/src/session_manager/search_gc.rs` | 77 (`disposable_workspace_index_id`) | Same gap on the teardown side: computes `trusty_common::derive_index_id(&root)` only, then `decommission_with_root` issues `DELETE /indexes/{that_id}`. For a decommissioned session whose root resolves to a different preferred id (org/repo form, not yet aliased to an existing legacy index), the DELETE targets an id nothing is registered under — a no-op — so the real index the session actually populated is never proactively removed at decommission time (issue #2033's whole point). It's eventually swept by `sweep_orphaned_search_indexes` (which matches on `root_path`, not id, so it's unaffected by this gap) once the workspace directory is gone, but that's a periodic backstop, not the immediate, correctness-guaranteed teardown this function's own doc claims ("CRITICAL (caller contract): must be invoked BEFORE the workspace directory is removed..."). | Route this through `resolve_effective_index_id` too (trusty-mpm already links `trusty-common` with the `search-index` feature via `ensure_project_indexed`'s existing use). |
| HIGH | `crates/trusty-search/src/main.rs` | 1419 (`Commands::Serve { project, .. }`) | `trusty-search serve --project <path>` still pins via `trusty_common::derive_index_id(&root)` only. The adjacent comment literally claims "derive it from `--project` via the shared derivation so it matches the CLI / trusty-mpm" — after this PR that's no longer true for any project whose git origin resolves: `detect_project` (this same crate, `detect.rs`, now updated) and trusty-mpm's `ensure_project_indexed` would both resolve the org/repo id, but a server started via `serve --project` pins to the stale legacy id instead, serving/registering under a different id than every other entry point. | Use `trusty_common::search_index::resolve_effective_index_id(&root)` here as well (the crate already carries the `search-index` feature, added by this PR's own `Cargo.toml` change). |
| MEDIUM | `crates/trusty-common/src/index_id.rs` | 95-102 | The single-hyphen join `format!("{}-{}", gp.owner, gp.repo)` is not injective: owner=`"foo-bar"`,repo=`"baz"` and owner=`"foo"`,repo=`"bar-baz"` both yield `"foo-bar-baz"`. Two genuinely distinct GitHub repos could collide on the exact same index id. This mirrors an already-accepted precedent (`palace_id::owner_repo_from_git_remote` uses the identical join, per this PR's own module doc), so it isn't new risk introduced by this PR, and real-world collisions are rare — flagging for awareness, not blocking. | If ever revisited, a reversible separator (e.g. encode the boundary, or keep `owner`/`repo` lengths in a comment) would remove the ambiguity; not required for this PR given the existing precedent. |
| MEDIUM | `crates/trusty-review/src/report/analyze_adapter.rs` | 502-505 | `trusty-review`'s `--analyze` report enrichment has its own independent `derive_index_id` (a separate, hand-rolled basename-only function, not calling into `trusty-common` at all). It isn't touched by this PR, so for any project migrated to the preferred org/repo id, `enrich_with_analyze` will query the daemon under the stale legacy id and silently get no live metrics (fails open — the doc explicitly says it must never abort the report, so this degrades a feature rather than corrupting anything). | Out of scope for this PR to fix, but worth a fast-follow issue — flagging since the review was asked to check every id-construction site for consistency. |

## Required Changes

1. Fix the CRITICAL worktree/facet-collapse bug in `derive_preferred_index_id` (or in `resolve_effective_index_id`) before merge — this is the core "does it actually prevent problems" question the PR exists to answer, and as written it introduces a worse problem (cross-session index sharing) than the one it fixes (basename collisions across unrelated directories). At minimum, add a test with two real worktrees (or two clones) of the same origin remote and assert they do NOT collapse to the same preferred id, or that the design intentionally routes worktree-scoped roots through the legacy per-directory id.
2. Migrate `search_readiness.rs:168`, `search_gc.rs:77`, and `main.rs:1419` to `resolve_effective_index_id` so the "single source of truth" claim in this module's own docs holds for every call site, not just the three touched by this diff.

## Notes

- The core alias-fallback mechanism in `resolve_effective_index_id` (probe legacy first, then preferred, default to preferred) is sound *for the single-root, single-checkout case* the new tests cover — I traced it against `ensure_project_indexed`/`best_effort_create_index` and confirmed it does prevent shadowing an existing legacy-keyed index for that case.
- The hyphen-join format choice (vs. `RepoIdentity::canonical()`'s `owner/repo`) is correct and well-reasoned given axum's single-path-segment `{id}` routing — I did not find any code path that constructs a raw slash-containing `owner/repo` id and feeds it into an HTTP route; every route-facing id construction goes through `derive_preferred_index_id`'s hyphen join or the untouched basename rule.

bobmatnyc:
## Code Critic — re-review of `84acc35b` (round 3)

## Verdict: BLOCK

The previously-flagged worktree collapse **is genuinely fixed**, and the call-site sweep is complete. But the fix closes the *linked-worktree* half of the collision only, and the alias-fallback resolver introduced alongside it opens a new, more common backward-compatibility failure.

### What I verified as fixed ✅

| Item | Status |
|---|---|
| Worktree disambiguation uses real git `--git-dir` vs `--git-common-dir`, not a path heuristic (`index_id.rs:182 linked_worktree_name`) | ✅ correct — I verified against git 2.54: main checkout prints `.git`/`.git`, linked worktree prints `<common>/worktrees/<name>` vs `<common>`; option order matches the argv order; a worktree of a **bare** `.base` (the tm session topology) is correctly detected as linked |
| A real two-worktree regression test exists | ✅ `derive_preferred_index_id_distinguishes_worktrees_of_same_repo` (`index_id.rs:336`) does real `git worktree add` ×2 off one origin and asserts 3 distinct ids; `disposable_index_id_differs_between_worktree_and_main_checkout` (`search_gc.rs:481`) does the same for the delete path. Not fixture theatre. |
| 3 flagged call sites migrated | ✅ `search_readiness.rs:175`, `search_gc.rs:87`, `main.rs:1422` all route through `resolve_effective_index_id` |
| `trusty-review`'s duplicate `derive_index_id` | ✅ actually migrated (`analyze_adapter.rs:514-517`), local copy deleted, `search-index` feature added to its `Cargo.toml` |
| No stale `derive_index_id` production call sites remain | ✅ grep-verified across all crates |
| No leftover conflict markers | ✅ |

**Tests (run by me at `84acc35b`):**
```
trusty-common:  test result: ok. 1384 passed; 0 failed; 26 ignored
trusty-mpm:     test result: ok. 3614 passed; 0 failed;  6 ignored
trusty-search:  test result: ok. 1231 passed; 0 failed;  1 ignored
trusty-review:  test result: ok. 1231 passed; 0 failed;  1 ignored
```
All 6 new tests pass. (`tm_hook_pm_guard` showed 2 failures in my first run — artifact of my review checkout living under `/tmp`, which the #3955 guard denies. Re-run from a non-tmp CWD: `test result: ok. 33 passed; 0 failed`. Not a PR defect.)

---

## Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-common/src/search_index.rs` | 106-108 | Daemon-unreachable ⇒ silently migrates an already-legacy-registered project to the new preferred id, producing a permanently dead `--index` pin for that whole session | Fail **safe**, not forward: return `legacy_id` when the daemon can't be consulted — or resolve from the persisted registry on disk |
| HIGH | `crates/trusty-common/src/index_id.rs` | 143 | Two independent **clones** of one repo (and two clones from different **hosts** with the same `owner/repo`) now collapse onto one index id — same mechanism as the BLOCKed worktree bug; worktree suffixing doesn't cover it. Proven empirically. | Make adoption root_path-aware (see below); decline a preferred id already owned by a different root |
| MEDIUM | `crates/trusty-common/src/search_index.rs` | 131 | `index_is_registered` adopts an id on bare 2xx, discarding `root_path` — which `/indexes/{id}/status` already returns | Parse `root_path`, require canonical equality with the project root |
| MEDIUM | `crates/trusty-common/src/search_index.rs` | 288 | Comment promises `index_files_inner` "must resolve to the SAME id `ensure_project_indexed` chose at task start" — not guaranteed, resolution now depends on mutable live daemon state | Resolve once and thread the id, or soften the claim |
| MEDIUM | `crates/trusty-review/src/report/analyze_adapter.rs` | 514 | Blocking `thread::join` (up to ~1.6 s **per repo**) called from the `async fn enrich_with_analyze` (line 534) loop | `tokio::task::spawn_blocking` |
| MEDIUM | `crates/trusty-mpm/src/session_manager/search_gc.rs` | 481 | New test probes whatever **live** trusty-search daemon the dev/CI box happens to run — no `TRUSTY_DATA_DIR_OVERRIDE` isolation, unlike its siblings in `search_index_tests.rs`. `assert_eq!(main_id, Some("acme-widget-xyz"))` fails if a live daemon holds an index literally named `main-checkout` | Point the data-dir override at the test's tempdir under the env lock, or drop the exact-value assert |

---

### CRITICAL — daemon-down defeats the entire backward-compat story

```rust
let Some(base) = crate::resolve_daemon_base_url("trusty-search") else {
    return preferred_id;   // search_index.rs:106-108
};
```

The alias-fallback only works while the daemon answers. It doesn't fail *neutrally* when it can't — it fails **forward**, onto the new id. Concrete chain for any existing user:

1. Project `~/dev/myapp` has been indexed for months under the legacy id `myapp`.
2. Machine reboots / daemon not yet up. `resolve_daemon_base_url` reads a **stale** `http_addr` file (it never liveness-checks), both probes get connection-refused, both read as "not registered".
3. `resolve_effective_index_id` returns `acme-myapp`; `ensure_project_indexed` swallows the failed `POST /indexes`; `session_launch/search_index.rs` writes `"args": ["serve","--index","acme-myapp"]` into the session's `.mcp.json`.
4. Daemon starts, warm-boots `myapp` at `~/dev/myapp`. The pinned `acme-myapp` **does not exist and can never be created**: `reindex_handler` 404s on an unknown id (`reindex_handlers.rs:51`), and `create_index_handler` is refused by `find_root_path_collision` because that root is already owned by `myapp` (`indexes.rs:254` — and main's brand-new #3995 tightens this to cold entries too).
5. The MCP stub is a thin forwarding client (`McpServer::new(base)` + `with_pinned_index`); it never creates anything. **Every `search`/`grep` in that session silently returns nothing, for the session's entire life.**

On `main` this path pinned the legacy id and worked. This is a regression against exactly the population the alias-fallback exists to protect, and `resolve_effective_index_id_defaults_to_preferred_when_daemon_down` (`search_index_tests.rs:189`) enshrines it as intended behaviour.

**Fix (either):**
- Minimal: `else { return legacy_id; }` — when you cannot verify, keep the historical id. A brand-new project first launched daemon-down stays on the basename id forever, which is the strictly safer trade.
- Better: resolve from the **persisted** registry instead of HTTP. `service/persistence.rs::PersistedIndex { id, root_path }` in `indexes.toml` is authoritative, works daemon-down, is faster than two 800 ms probes, and hands you `root_path` for free — which also fixes the two findings below.

### HIGH — sibling clones and cross-host remotes still collide

`derive_preferred_index_id` disambiguates *linked worktrees*, but the id is derived from `remote.origin.url` with the **host stripped** (`github_path.rs:158-186` keeps only the last two path segments). Two independent clones are each a main working tree, so neither gets a suffix. Verified by running the compiled function at this commit:

```
LEGACY    (behaviour on main): a=widget  b=widget-review     <- distinct
PREFERRED (behaviour on PR):   a=acme-widget  b=acme-widget  <- COLLIDE
CROSS-HOST preferred:          github=acme-widget  gitlab=acme-widget
```

Downstream this is the identical mechanism the round-2 BLOCK described: `create_index_handler` returns `{"created": false, "reason": "already exists"}` on an id hit (`indexes.rs:238-244`), so the second checkout's registration silently degrades to a *find* against the first checkout's `root_path`, and every subsequent search, reindex, and incremental `index-file` update targets the wrong content tree — on a different branch. `main` kept these apart via the basename; this PR removes that separation. The docstring's "Known limitation" paragraph covers hyphen-join ambiguity but not this.

The root-path check below resolves it: if the preferred id is already held by a *different* root, decline it and fall back to the legacy basename.

### MEDIUM — the probe throws away the discriminator it already receives

`index_is_registered` maps the whole response to `bool`, but `GET /indexes/{id}/status` returns `"root_path"` (`service/server/status.rs:190`). Two consequences, both fixed by one change:

- **Legacy adoption is unguarded.** Directory basenames like `docs`, `web`, `api`, `app` are exactly the collision this PR exists to fix — yet if *any* unrelated project already holds that id, the legacy probe adopts it. Via `disposable_workspace_index_id` (`search_gc.rs:87`) that adopted id is then handed to `DELETE /indexes/{id}` on decommission, destroying the unrelated project's index. (Pre-existing on `main`, but this PR rewrote that exact line and the fix is now cheap.)
- **Preferred adoption is unguarded** — the sibling-clone case above.

```rust
// adopt an id only when the registered index actually points at THIS root
fn index_root_matches(base: &str, id: &str, root: &Path) -> bool { /* parse "root_path", canonicalize, compare */ }
```

---

## Required Changes

1. **`search_index.rs:106-108`** — daemon-unreachable must return `legacy_id`, or resolution must read the persisted `indexes.toml` registry. Update `resolve_effective_index_id_defaults_to_preferred_when_daemon_down` to assert the safe behaviour.
2. **`search_index.rs:131`** — parse `root_path` out of the status body and require canonical equality before adopting either id; when the preferred id is owned by a different root, fall back to the legacy basename.
3. **Regression test** for sibling clones: two `git clone`s of one origin into differently-named directories must not share an index id (they currently do — reproduction above).
4. **`analyze_adapter.rs:514`** — wrap the resolver call in `tokio::task::spawn_blocking` at the `enrich_with_analyze` call site.
5. **`search_gc.rs:481`** — isolate the test from any live daemon.
6. **Rebase/merge onto current `origin/main`.** GitHub's `CONFLICTING` / `DIRTY` is **not** stale — `git merge-tree --write-tree origin/main 84acc35b` reports a real content conflict in `crates/trusty-search/CHANGELOG.md` (branch is 3 ahead / 1 behind). Main's new head `c62ef8ad` (#3995) tightens `find_root_path_collision` to cold entries, which interacts directly with change 1.

## Notes

- `crates/trusty-review/CHANGELOG.md` gains a retroactive `## [0.10.1] — 2026-07-23` heading splitting a pre-existing double-`### Fixed` under `[Unreleased]`. It looks correct (0.10.1 is the shipped version) but it is unrelated cleanup — flagging only so it isn't mistaken for a merge artifact.
- Not flagged: the hyphen-join non-injectivity (`a`/`b-c` vs `a-b`/`c`) — consciously documented, needs adversarial naming, and the single-path-segment constraint is real.
- Not flagged (below 80% confidence): `linked_worktree_name` inherits `GIT_DIR`/`GIT_COMMON_DIR` when invoked from inside a git hook (`commands/hook.rs` → `detect_project`). I traced it and believe it resolves correctly in both the main-checkout and worktree hook cases, but it is untested.

*Verdict computed under `code-review-standards`: 1 CRITICAL ⇒ BLOCK. Pipeline halts; awaiting explicit direction rather than auto-routing back to the implementer.*

bobmatnyc:
Closed as won't-do by owner decision (2026-07-28). The motivating problem is real and moves to epic #4207; this approach does not.

Root cause of the three BLOCK rounds: the PR promotes a GROUPING key into a PARTITIONING slot. `RepoIdentity`'s own module doc states it is a grouping key, deliberately shared across every facet of one repo. trusty-search's registry is one-`root_path`-per-id, so the index id must partition — one id, exactly one content tree. Each review round found another case where two genuinely different roots share the grouping value: linked worktrees (fixed by `f930fec9`), sibling clones of the same repo (still collide, proven live), and the daemon-unreachable path resolving forward onto the new id and orphaning the session's search (`search_index.rs:106-108`, unfixed). The gap narrows each round but never closes.

No commits exist after the round-3 findings.

Reopening should start from the partitioning-key question, not from this branch. The branch stays in place as design history — not deleted.

PR #4065, opened 2026-07-27
fix(trusty-mpm): stop bare tm from misreporting a managed pane as 'not a git project' (#4061)

### LINKED ISSUE #4061: tm bare-invocation double-invocation race: guided-default git-check fires before daemon session record settles to stopped
Sub-issue of #3764 — filed against a distinct mechanism (daemon-record race) found while investigating that worktree-corruption incident; not a duplicate. Also related to #3715 (parent symptom issue: worktree loss makes this race more visible, but the race itself is independent of worktree health).

## Summary

Independent of the worktree git-identity corruption tracked in #3764, `tm`'s bare-invocation flow has its own race that produces a confusing double-invocation UX: the first bare `tm` in a pane can print "not in a git project" and exit, and running the *exact same command* moments later succeeds with an in-place relaunch. This is a real Active→Stopped state-transition race in the daemon's session record, not a fluke — and it can happen even when the underlying worktree/cwd is perfectly healthy.

## Root cause (file:line citations verified against current `main`)

- `crates/trusty-mpm/src/bin/tm/commands/guided.rs`, `try_inplace_relaunch` is invoked first on every bare `tm` call (guided.rs:101), before any git-project detection runs.
- That call is gated on `TM_MANAGED_SESSION_ID` being set AND the daemon's session record for this pane being in state `"stopped"` — see `guided_inplace.rs::plan_inplace` (guided_inplace.rs:219), which polls the daemon within a `FETCH_RETRY_BUDGET` of ~400ms (guided_inplace.rs:116) waiting for the record to settle.
- If, at the moment of invocation #1, the daemon record is still `"active"` (the async `SessionEnd` / `mark_runtime_exited_stopped` healing step hasn't yet flipped it to `"stopped"`), the gate misses within its retry budget and control falls through to the guided-default path: guided.rs:282 prints `"tm: no subcommand — using guided default for {workdir}"`, then guided.rs:294 calls `derive_project(&cwd)`. If the cwd has no git identity (e.g. a corrupted worktree per #3764, but also possible for other reasons), this prints `NON_GIT_FALLBACK_HINT` — `"tm: not in a git project — run 'tm --help' for available commands"` (misc.rs:71-72) — and the process exits.
- On invocation #2, run moments later, the daemon has by then flipped the record to `"stopped"`, so `try_inplace_relaunch`'s gate now passes: guided_inplace.rs:472 prints `"tm: this pane belongs to managed session {id} — relaunching in place…"` and execs `claude` back into the pane.

## Net effect

The user sees a confusing two-step dance: first a "not in a git project" error, then — on retrying the exact same bare `tm` command — a successful in-place relaunch. This reads as if the first invocation was simply wrong or the directory was broken, when actually it's a state-transition race in the daemon's session record that happens to be exposed at this call site.

## Reproduction conditions

Reproducible any time a pane's daemon session record hasn't yet settled to `"stopped"` when bare `tm` is invoked — independent of whether the underlying worktree is corrupted (per #3764) or perfectly healthy. Timing-sensitive: requires invoking bare `tm` within the ~400ms (or longer, if the async healing step is itself delayed) window after the pane's runtime process exits but before the daemon's `SessionEnd`/reap-tick handling has updated the record.

## Suggested fix direction

Either:
1. Have the guided-default fallback (guided.rs:282 onward) itself retry/wait on the same daemon-record-settle window that `try_inplace_relaunch` uses before falling through to `derive_project`/the git-project check, so a still-settling record doesn't get treated as "not in a git project"; or
2. Invert the ordering guarantee so the pane-ownership check always runs to completion — including its full retry budget — before ever falling through to the git-project check, rather than the git-check being reached opportunistically whenever the pane-ownership race is lost.

Either direction removes the double-invocation UX without touching the underlying worktree-corruption issue tracked separately in #3764.

bobmatnyc:
## Summary

- `try_inplace_relaunch` (the first thing bare `tm` tries) is gated on the daemon's session record reading `"stopped"` within a bounded ~400ms retry budget (`FETCH_RETRY_BUDGET`). The async `SessionEnd` healing step that flips a just-exited managed pane's record from `active` to `stopped` can still be in flight at that exact moment, so the very first bare `tm` in that pane can miss the gate and fall through to the guided-default's non-git fallback, which printed a misleading `"tm: not in a git project"` error — only succeeding on a confusing second invocation once the daemon had settled.
- `fallback_protected`'s `CwdProject::NotGit` branch now asks whether `TM_MANAGED_SESSION_ID` resolves from either the process environment or the tmux SESSION environment (`guided_inplace::resolve_env_managed_session_id`, the same two-source lookup `try_inplace_relaunch` itself uses) before ever blaming "not a git project" (`guided::plan_non_git_fallback`). When present, it retries `try_inplace_relaunch` once more — bounded, same budget, real wall-clock time has passed since the first attempt — before ever printing a git-project-shaped message; if that retry still can't confirm, it prints an honest `"...daemon record has not finished settling yet — run tm again in a moment"` message instead (`guided::managed_pane_settle_pending_message`).
- A pane with no such signal at all (the common, non-managed case) takes the unchanged fast path — no extra daemon round trip, no added latency.

## Test plan

- [x] `plan_non_git_fallback_retries_when_env_id_present` / `plan_non_git_fallback_ordinary_when_no_env_id` — pure decision logic
- [x] `managed_pane_settle_pending_message_names_id` — honest message replaces the misleading git-project hint
- [x] `resolve_env_managed_session_id_prefers_process_env` / `resolve_env_managed_session_id_none_when_unset` — env resolution
- [x] `guided_fallback_non_git_dir_no_managed_env_is_fast` — fast path unaffected (asserts no added latency)
- [x] `guided_fallback_non_git_dir_with_managed_env_settles_quickly_returns_promptly` — reproduces the race with a local mock whose record starts `"active"` then flips to `"stopped"` on the next poll
- [x] `guided_fallback_non_git_dir_with_managed_env_unreachable_daemon_does_not_hang` — worst case still resolves promptly, never hangs
- [x] `cargo test -p trusty-mpm` — full suite green (3555 + 1226 + 1226 unit tests, all integration test binaries), 0 failed
- [x] `cargo clippy -p trusty-mpm --all-targets` — clean
- [x] `cargo fmt --check -p trusty-mpm` — clean

Closes #4061

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -48,6 +48,16 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **Bare `tm` no longer misreports a managed pane as "not a git project"**
+  (closes [#4061](https://github.com/bobmatnyc/trusty-tools/issues/4061)):
+  when a managed pane was still settling, the in-place relaunch could observe
+  the session as not-yet-resolvable and fall through to the generic
+  non-git-project hint — telling the operator their repo wasn't a git project
+  when it plainly was. Bare `tm` now retries the in-place relaunch when an
+  env-resolved managed session id exists, and if the session still hasn't
+  settled it prints an honest "still settling" message instead of the
+  misleading non-git hint.
+
 - **Concurrent workspace-trust seeds no longer clobber each other's
   `~/.claude.json` entries** (closes
   [#4072](https://github.com/bobmatnyc/trusty-tools/issues/4072)):
--- a/crates/trusty-mpm/src/bin/tm/commands/guided.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided.rs
@@ -1006,6 +1006,75 @@ pub(crate) fn untracked_ancestor_message(git_root: &std::path::Path) -> String {
     )
 }
 
+/// Decision for the non-git-project branch of [`fallback_protected`]: does an
+/// env-resolved managed-session id mean the "not in a git project" verdict is
+/// premature (#4061)?
+///
+/// Why: `derive_project`/`classify_cwd_project` finding no usable git root is
+/// exactly right for a genuinely unmanaged directory — but it says nothing
+/// about whether THIS pane is a managed session's own pane whose daemon
+/// record simply has not settled the Active -> Stopped race yet (the same
+/// race [`super::guided_inplace::fetch_managed_session_until_stopped`] bounds
+/// for `try_inplace_relaunch`'s own gate at the top of `run_guided_default`).
+/// Separating this decision from the I/O (the env lookup, and the retry
+/// itself) keeps it exhaustively unit-testable.
+/// What: `Some(id)` — [`super::guided_inplace::resolve_env_managed_session_id`]
+/// found a managed-session id in either the process or tmux SESSION
+/// environment — means retry [`super::guided_inplace::try_inplace_relaunch`]
+/// once more (real wall-clock time has elapsed since its first attempt at the
+/// top of `run_guided_default`, doing the nested-session-guard round trip and
+/// project detection in between) before ever printing a message that blames
+/// "not a git project". `None` — no signal at all — is the ordinary,
+/// unaffected fast path: print [`super::misc::NON_GIT_FALLBACK_HINT`]
+/// immediately, no extra daemon round trip, no added latency.
+/// Test: `plan_non_git_fallback_retries_when_env_id_present`,
+/// `plan_non_git_fallback_ordinary_when_no_env_id`.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum NonGitFallbackPlan {
+    /// No managed-session signal anywhere — the ordinary "not in a git
+    /// project" hint is correct; print it immediately.
+    OrdinaryNotGit,
+    /// A managed-session id WAS found in the environment — give
+    /// `try_inplace_relaunch` one more bounded attempt before ever printing a
+    /// git-project-shaped message for a pane we have evidence is managed.
+    RetryInPlace(String),
+}
+
+/// Pure decision behind [`NonGitFallbackPlan`] — see its doc for the full
+/// rationale.
+///
+/// Test: `plan_non_git_fallback_retries_when_env_id_present`,
+/// `plan_non_git_fallback_ordinary_when_no_env_id`.
+pub(crate) fn plan_non_git_fallback(env_managed_session_id: Option<&str>) -> NonGitFallbackPlan {
+    match env_managed_session_id {
+        Some(id) => NonGitFallbackPlan::RetryInPlace(id.to_string()),
+        None => NonGitFallbackPlan::OrdinaryNotGit,
+    }
+}
+
+/// Build the operator-facing notice printed when a managed-session id WAS
+/// found in the environment (#4061) but the retried
+/// [`super::guided_inplace::try_inplace_relaunch`] still could not confirm
+/// the in-place relaunch — e.g. the daemon's Active -> Stopped settle race is
+/// taking longer than both bounded attempts combined, or the daemon is
+/// genuinely unreachable.
+///
+/// Why: printing [`super::misc::NON_GIT_FALLBACK_HINT`] here would actively
+/// mislead the operator — this cwd may well not be a git project, but that is
+/// NOT why nothing happened; the pane is a known managed one whose record
+/// just has not settled yet. An honest, actionable message names the session
+/// id and tells the operator the one thing that reliably works: try again.
+/// What: a single-line notice naming `id`. Pure — the caller prints it via
+/// `eprintln!` and returns `Ok(())`, exactly like every other fallback notice
+/// in this module.
+/// Test: `managed_pane_settle_pending_message_names_id`.
+pub(crate) fn managed_pane_settle_pending_message(id: &str) -> String {
+    format!(
+        "tm: this pane belongs to managed session {id}, but its daemon record has not \
+         finished settling yet — run `tm` again in a moment."
+    )
+}
+
 /// Daemon-unreachable fallback that protects ALL live git checkouts (#1724).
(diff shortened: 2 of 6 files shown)


bobmatnyc:
## Verdict: APPROVE

Reviewed the full diff (both commits, `7c70ec55` + `32989fd1`), not just the latest — this is the first critic pass on this PR.

### What I checked

**Fast path unaffected (item 2):** confirmed via code, not just the PR description. `fallback_protected`'s `CwdProject::NotGit` arm calls `plan_non_git_fallback(resolve_env_managed_session_id().as_deref())` (`guided.rs:1116-118`); when the env lookup is `None`, `plan_non_git_fallback` returns `OrdinaryNotGit` and the arm prints `NON_GIT_FALLBACK_HINT` immediately — no daemon round trip, no `try_inplace_relaunch` call. `resolve_env_managed_session_id` (`guided_inplace.rs:206-207`) is a pure `read_env_managed_session_id().or_else(read_tmux_env_managed_session_id)` — both cheap, local reads (env var / one `tmux show-environment` shell-out), not a network call. The regression suite backs this: `guided_fallback_non_git_dir_no_managed_env_is_fast` asserts `elapsed < 150ms` with no `TM_MANAGED_SESSION_ID`/`TMUX` set — I ran it, it passes.

**Race reproduction (item 3):** `guided_fallback_non_git_dir_with_managed_env_settles_quickly_returns_promptly` spins up a real local `TcpListener` mock that replies `"active"` on the first hit and `"stopped"` on the second — a genuine simulation of `SessionEnd`'s async healing step catching up mid-retry, not a trivial always-stopped mock. It asserts `elapsed < 300ms` (i.e., the retry loop breaks out on the first "stopped" poll rather than exhausting the ~400ms budget) and that the daemon was actually queried (`hits >= 1`). `guided_fallback_non_git_dir_with_managed_env_unreachable_daemon_does_not_hang` covers the daemon-unreachable worst case. Combined with the fast-path test above, this is real coverage of both directions, not a rubber-stamp test.

One coverage gap worth naming (not blocking): all three new tests call `fallback_protected` directly. None exercise the *actual* invocation path (`run_guided_default`) where `try_inplace_relaunch` already ran once at the top before `fallback_protected` gets a second bounded attempt — so the true worst-case stacked latency (~2×400ms budget for a present-but-never-settling env var in a non-git dir) isn't asserted end-to-end. Still bounded and safe, just untested at that layer.

**Doc-pointer fix (item 4):** confirmed `fallback_protected`'s doc now cites `guided_fallback_non_git_dir_no_managed_env_is_fast`, `guided_fallback_non_git_dir_with_managed_env_settles_quickly_returns_promptly`, and `guided_fallback_non_git_dir_with_managed_env_unreachable_daemon_does_not_hang` — all three exist verbatim in `tests_behavior_b_tests.rs` and pass. The second commit correctly replaces the dangling `guided_fallback_non_git_dir_with_managed_env_retries_instead_of_hint` citation that never matched a real test.

**`cargo test -p trusty-mpm` (full binaries, not `--lib`):** ran the complete suite in a clean worktree at the PR head (`32989fd1`). Result: every #4061-related test passes (`resolve_env_managed_session_id_*`, `plan_non_git_fallback_*`, `managed_pane_settle_pending_message_names_id`, all four `guided_fallback_non_git_dir_*`), lib tests 3611 passed/0 failed, both behavior-suite binaries 1239 passed/0 failed each. The only failure in the whole run is `tests/tm_hook_pm_guard.rs` (`pm_guard_allows_worktree_add_under_project_dir_via_subagent_payload`, `pm_guard_claude_mpm_sub_agent_env_still_allows_everything_else`) — I confirmed this is **pre-existing and unrelated**: it reproduces identically on `origin/main` HEAD (`c62ef8ad`) run in the same environment, in code this PR never touches (the #3955/#3978 worktree hard-block feature). It's a local-environment artifact (worktrees living under `/private/tmp/...` trip the "under tmp" guard), not a regression from this change.

### Findings

| MEDIUM | `crates/trusty-mpm/CHANGELOG.md` | — | This PR changes `trusty-mpm` source (user-visible bug fix for #4061) but adds no `## [Unreleased]` bullet. Per this repo's own convention (BASE-AGENT "Framework-Guaranteed Conventions" / `tm-pr-workflow`), a missing changelog entry is a review-gate failure, not optional polish. | Add a bullet under the existing `## [Unreleased]` section describing the #4061 fix before merge. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided.rs` | 1119-1125 (`fallback_protected`'s `RetryInPlace` arm) | `managed_pane_settle_pending_message` always says "hasn't finished settling yet — try again," but the retried `try_inplace_relaunch` can return `None` for reasons unrelated to settling: the record resolves but is confirmed non-`Stopped` forever, the pane_id check fails (a different pane's env leaked in), the id 404s (session pruned/unknown), or the daemon flatly refuses reactivation. In those cases "try again" will never help and mildly misleads the operator about what's actually wrong. | Either thread the real reason out of `try_inplace_relaunch`'s `None` case (e.g. an enum instead of `Option`) so the message can differentiate, or soften the wording to be reason-agnostic, e.g. "tm could not confirm this pane's managed-session state — run `tm` again, or run `tm sessions list` if this persists." Not blocking; the current message is still honest that this is a managed pane and still exits cleanly. |

No CRITICAL or HIGH findings. The core fix is correct: it closes the race without adding latency to the two common cases (genuinely unmanaged pane, or a managed pane whose record has already settled), and the safety properties from #2027/#2453/#2148 (Stopped-only gate, pane_id confirmation, honored reactivate response) are untouched — this PR only adds a second bounded attempt at an existing, already-guarded entry point.

### Notes

- CI: all 22 non-`Test` checks were green at review time; the `Test` job was still running. Given the local full-suite run above (same commit) is green apart from the pre-existing, environment-specific `tm_hook_pm_guard` failure (which also fails on `origin/main`), I'd expect CI's `Test` job to pass — but that job's own result should still be confirmed green before merge, not inferred from this review.
- The two MEDIUM items are not merge-blockers on correctness grounds; the CHANGELOG gap specifically should be closed before merge per repo convention.

Issue #3828, opened 2026-07-24
trusty-mcp unified CLI multiplexer: single entry point for all native MCP services


bobmatnyc:
Context: Currently, native MCP services expose per-crate binaries to users (e.g. gworkspace-mcp setup --profile bob-personal), which leaks crate internals into the user-facing command surface.

Directive (Bob, 2026-07-24): Create ONE unified CLI entry point: trusty-mcp <service> <command>

Instead of gworkspace-mcp setup --profile bob-personal, users run: trusty-mcp gworkspace setup --profile bob-personal

Per-crate MCP binaries become an internal implementation detail (kept for the MCP stdio spawn path or absorbed entirely).

Proposed command surface:
- trusty-mcp <service> setup [--profile NAME]
- trusty-mcp <service> doctor
- trusty-mcp <service> accounts list
- trusty-mcp <service> enable / disable (future)

Covered services: gworkspace, memory, search, channels/slack, kb, plus future ones.

Related: ADR-0014, Epic #2626, #2631

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closed as not planned — owner review of the wont-do candidates, 2026-09-12 (session trusty-tools-c4). Reopen if the need returns.

PR #4066, opened 2026-07-27
chore: allow crates.io publishing for trusty-agents, trusty-channels, trusty-gworkspace

bobmatnyc:
## Summary

Reverses the `publish = false` guard on three crates -- `trusty-agents`,
`trusty-channels`, `trusty-gworkspace` -- so they can be published to
crates.io like the workspace's other main tools (`trusty-search`,
`trusty-common`, `trusty-memory`, ...). **This is a deliberate, explicit
policy reversal authorized verbally by Bob (repo owner).**

- `crates/trusty-agents/Cargo.toml`: drop `publish = false` (set 2026-06-06, reaffirmed 2026-07-24)
- `crates/trusty-channels/Cargo.toml`: drop `publish = false` (set at the crate's founding commit, 2026-07-15)
- `crates/trusty-gworkspace/Cargo.toml`: drop `publish = false` (set 2026-05-20)

This matches the convention already used by most publishable crates in this
workspace: they simply omit the `publish` key (default `true`) rather than
writing `publish = true` explicitly. Some crates (trusty-embedderd,
trusty-embedderd-py, trusty-bm25-daemon, trusty-git-analytics) use explicit
`publish = true`, but both patterns are valid and supported; this PR follows
the "omit the key" convention used by already-published tools like
`trusty-search`, `trusty-common`, and `trusty-memory`.

## Why now

- All three crates were restricted in **d3a9cadc** ("chore: restrict crates.io
  publishing to 4 main tools + required lib deps," 2026-05-20), which set
  `publish = false` on 12 crates so that only a small core tool set shipped
  to crates.io.
- Since then, **ADR-0014** (native MCP support, accepted 2026-07-14)
  established that each of these three crates ships a native per-crate MCP
  binary -- `trusty-agents` -> `ompm`/`tagent`, `trusty-channels` ->
  `slack-mcp`/`telegram-mcp`, `trusty-gworkspace` -> `trusty-gworkspace-mcp`
  -- following the same tool-registration convention already used by the
  currently-published `trusty-memory` and `trusty-search`. Architecturally
  they're already first-class tools of the same shape as what's already
  published; this PR just aligns the publish policy with that reality.
- A separate, unimplemented idea (#3828, "trusty-mcp unified CLI
  multiplexer") does **not** block this -- it's an open, zero-progress
  aspiration with no code, not a reason to keep these off crates.io.

## Known precedent -- please don't reflexively flag this

`trusty-gworkspace` was published once before: **0.1.0, published
2026-05-20 (the same day the restriction commit landed), yanked
2026-05-26.** That yank is already understood context, not new information
-- it's exactly why this PR is manifest-only (see below) rather than
publishing immediately.

## What this PR deliberately does NOT do

- Does **not** run `cargo publish` or tag anything. This is purely the
  policy/manifest change. The actual publish is a separate follow-up once
  this merges, via the `cargo-publish` skill workflow.
- Does **not** merge itself. This reverses the repo owner's own prior
  explicit policy commit, so even though Bob authorized the change
  verbally, it should get his eyes (plus CI/code-critic) before merging.

## Verification

- License check: all three crates already use `license.workspace = true`,
  which resolves to the workspace's `license = "MIT"` -- a valid SPDX
  identifier, so no `license-file` fix is needed (unlike, say, an
  `Elastic-2.0` crate would need).
- Checked each crate's remaining path dependencies for other
  `publish = false` crates that would block a standalone `cargo publish`
  resolution -- none found (`trusty-memory`, `trusty-search`, `trusty-kb`,
  `trusty-agents-common`, `trusty-common` are all already publishable).
- `crates/trusty-publish-guard` (the version-parity drift detector) and
  `.github/workflows/release.yml`'s publishability gate both determine
  "is this crate publishable" dynamically by grepping/parsing the
  manifest's `publish` field -- no hardcoded allowlist anywhere needed
  updating for this flip.
- Metadata fix: added `repository.workspace = true` and `homepage`
  fields to `trusty-gworkspace/Cargo.toml` to match metadata patterns used
  by other published crates like `trusty-search` and `trusty-common`.
- Ran `cargo check -p trusty-gworkspace`: passes after metadata addition.

## Test plan

- [x] `cargo check --workspace` passes (modulo the two pre-existing,
      unrelated Tauri GUI frontend-dist failures reproduced on main)
- [ ] CI green
- [ ] code-critic review
- [ ] Bob's sign-off before merge (holding per his own instruction not to
      self-merge a reversal of his prior explicit policy)

### DIFF
--- a/crates/trusty-agents/Cargo.toml
+++ b/crates/trusty-agents/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "trusty-agents"
 version = "0.38.6"
-publish = false
 edition = "2024"
 description = "Trusty Agents: Rust-native agentic harness with multi-model routing, tool execution, and subprocess delegation."
 license.workspace = true
--- a/crates/trusty-channels/Cargo.toml
+++ b/crates/trusty-channels/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "trusty-channels"
 version = "0.1.0"
-publish = false
 edition = "2021"
 description = "Native chat-channel MCP servers (Slack, …) for the Trusty suite — chat-as-tools"
 license.workspace = true
--- a/crates/trusty-gworkspace/Cargo.toml
+++ b/crates/trusty-gworkspace/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.2.2"
 edition = "2024" # edition 2024 required: let-chains used in api/client.rs, api/services/gmail/, api/services/docs/ (issue #258)
 description = "Google Workspace MCP server for the Trusty suite"
 license.workspace = true
+repository.workspace = true
+homepage = "https://github.com/bobmatnyc/trusty-tools"
 
 [lib]
 crate-type = ["rlib"]
--- a/crates/trusty-gworkspace/Cargo.toml
+++ b/crates/trusty-gworkspace/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "trusty-gworkspace"
 version = "0.2.2"
-publish = false
 edition = "2024" # edition 2024 required: let-chains used in api/client.rs, api/services/gmail/, api/services/docs/ (issue #258)
 description = "Google Workspace MCP server for the Trusty suite"
 license.workspace = true


bobmatnyc:
## Verdict: APPROVE

Independent adversarial review (code-critic). CI green at time of review. No issues found at >80% confidence.

## Findings

No CRITICAL, HIGH, MEDIUM, or LOW findings.

## Verification performed

**Diff scope** — `gh pr diff 4066` touches exactly 3 files, no more, no less:
- `crates/trusty-agents/Cargo.toml` — removes `publish = false` only
- `crates/trusty-channels/Cargo.toml` — removes `publish = false` only
- `crates/trusty-gworkspace/Cargo.toml` — removes `publish = false`, adds `repository.workspace = true` + `homepage = "https://github.com/bobmatnyc/trusty-tools"`

No scope creep beyond the three named crates.

**trusty-gworkspace metadata** — verified `repository.workspace = true` resolves via `[workspace.package]` in the root `Cargo.toml` (line 86) to `repository = "https://github.com/bobmatnyc/trusty-tools"`, and the new `homepage` field is the identical, correctly-formed URL pointing at the actual repo. `license.workspace = true` resolves to workspace `license = "MIT"` (line 85), a valid SPDX identifier — no `license-file` needed. `description` field is already present and unchanged. These are real, present, correctly-formed fields, not just claimed in the PR body.

**PR description factual correction, verified by grep**:
```
grep -rl 'publish = true' crates/*/Cargo.toml
→ trusty-bm25-daemon, trusty-embedderd-py, trusty-git-analytics, trusty-embedderd
```
This exactly matches the PR body's corrected list of crates using explicit `publish = true`. Also confirmed `trusty-search`, `trusty-common`, `trusty-memory` all omit the `publish` key entirely (default `true`), matching the "omit the key" convention the PR says it's following.

**Publish-blocking path-dependency check** — verified independently (not just trusting the PR body): grepped all `path = "../..."` dependencies of the three crates against the current repo's `publish = false` list (`cto-assistant`, `trusty-agents-local`, `trusty-channels` itself, `tc-services`, `trusty-code-gui`, `trusty-cto-db`, `trusty-gworkspace` itself, `trusty-mpm-gui`, `trusty-publish-guard`). None of `trusty-agents`'s path deps (`trusty-memory`, `trusty-search`, `trusty-agents-common`), `trusty-channels`'s (none — uses `workspace = true` for `trusty-common`), or `trusty-gworkspace`'s (`trusty-common`) hit that list. No blocked-dependency risk for a future real `cargo publish`.

**Scope-creep / accidental-publish risk** — none found. This is a pure manifest diff (2-3 lines per crate); no code changes, no other crates touched, no workspace-level `publish` default changed. The PR explicitly does not run `cargo publish` itself.

## Notes

- CI's `publish-dry-run` job (`.github/workflows/release.yml` job `publish-dry-run`, ~line 1191) is tag-triggered on the release pipeline, not a PR-time check — so a real `cargo publish --dry-run` for these three crates has not yet been exercised by CI for this PR. Not a blocker here since the PR is explicitly manifest-only and defers the actual publish to a separate follow-up per the `cargo-publish` skill; flagging so that follow-up step runs the dry-run before any real publish (particularly worth re-confirming `trusty-gworkspace` 0.2.2 isn't already taken/yanked-and-blocked on crates.io, given the 0.1.0 yank history called out in the PR body).
- This PR reverses Bob's own prior explicit policy commit (d3a9cadc); the PR body correctly gates merge on his sign-off in addition to this review — that gate is outside this review's scope to enforce but is correctly represented as still-open in the PR's own checklist.

APPROVED for next pipeline stage (Bob's sign-off).

PR #4067, opened 2026-07-27
feat(trusty-git-analytics): TGA JIRA ingestion — client + fact_ticket_transitions + tga jira sync

bobmatnyc:
Part of #3966

## Summary

This is a narrow slice of the JIRA-ingestion-ownership epic (#3966). It delivers:

- **JIRA API client extraction** (`collect::jira::client`): `search_with_changelog()` — paginated `POST /rest/api/3/search` with `expand=changelog`, extracting only `status`-field changelog items into `JiraTransition` rows; `fetch_comments()` — paginated `GET /issue/{key}/comment`. Credentials now expand `${ENV_VAR}` placeholders via `collect::env_expand`, matching the existing GitHub/Linear/AZDO client convention (previously this client took literal username/token strings, unlike its siblings).
- **`fact_ticket_transitions` schema** (migration v23, `sql/0023_jira_ingestion.sql`): grain is `(ticket_key, transitioned_at, to_status)` so a ticket revisiting the same status (e.g. reopened) yields distinct rows. `epic_key` is intentionally NOT modeled — 100% NULL on this JIRA instance; decomposition is via `parent_key`. `acceptance_criteria` and `story_points` are likewise out of scope for this grain (documented in the migration's header comment).
- **`tga jira sync`**: incremental by default, resuming from a stored `jira_sync_cursor` per project. `--backfill`/`--since` for explicit historical scope; `--dry-run` for a no-write preview. The cursor advances to the maximum `updated` timestamp actually *observed* in the run (never "now"), and a zero-ticket run leaves the cursor untouched rather than regressing it.
- Pure JQL-building/cursor-arithmetic helpers in a new `collect::jira::sync` module, kept HTTP/DB-free so they're unit-testable directly.

**Beyond the originally-scoped minimal slice**, this PR also includes `fact_jira_comment_detail` (schema + `fetch_comments` + wiring into `tga jira sync`) and `tga jira freshness` (the CRITICAL freshness-assertion guard called out in the issue, failing loudly when either fact table is empty or stale past `--max-age-days`). These were already fully implemented and tested in this branch's starting state; rather than strip out working, tested code to force an artificially narrower diff, this PR ships them as-is. See the issue comment for the precise "what's left" breakdown.

## Tests

- Pagination logic: `collect::jira::client` unit tests (changelog search response shape, non-status items filtered, missing `fromString` on initial transition, project-key fallback, unparseable-timestamp skip).
[...]

PR #4070, opened 2026-07-27
fix(ci): raise test-file SLOC cap 1500 -> 3000, keep production cap at 500

bobmatnyc:
## Summary

The 500/1500 dual SLOC cap in `scripts/check_line_cap.sh` (issue #1131) has caused friction on three separate PRs this session:

- **#3995** — `tests_3993.rs` (589 SLOC) doesn't match the `*_tests.rs` suffix classification pattern (it's prefixed, not suffixed), so it fell through to the 500-line **production** cap and had to be renamed to `collision_3993_tests.rs`. This is actually a classifier gap rather than a cap-value problem — flagging it here for visibility, but out of scope for this PR (naming-pattern classification, not the cap number).
- **#4060** — `session_manager/tests.rs` was already grandfathered on the allowlist at a frozen budget (1503-1516 SLOC) and had to be split into a new `prune_state_tests.rs` sibling just to land unrelated daemon work. This is the test cap genuinely binding.
- **#4067** — currently failing CI on `jira/client.rs` (667 SLOC), but that's a **production** file over the 500 cap, unrelated to the test-cap discussion. Left untouched here (production cap is explicitly out of scope for this change) — that PR needs its own split.

## Calibration data

Computed real SLOC (via the script's own `SLOC_AWK` counter, not raw `wc -l`) across all 689 tracked test/benchmark files on `origin/main`:

| SLOC band | file count |
|---|---|
| > 1500 (over old cap) | 1 (grandfathered) |
| 1400-1500 | 8 |
| 1200-1400 | 12 |
| 1000-1200 | 5 |
| 500-1000 | 69 |
| < 500 | 594 |

The largest legitimate test file today is 1503-1516 SLOC (the one grandfathered entry). The next 8 sit clustered in the 1400-1500 band — a bunching-just-under-the-ceiling pattern consistent with engineers trimming or splitting well-organized, well-tested files purely to dodge the cap rather than for genuine maintainability reasons.

## Change

- `TEST_CAP`: **1500 -> 3000** SLOC (`scripts/check_line_cap.sh`) — roughly 2x the old cap, comfortably above every observed legitimate test file today with real headroom for parameterized/fixture-heavy test growth.
- `PROD_CAP`: **unchanged at 500** SLOC — this is test-file leniency only, not a blanket relaxation.
- Regenerated `.line-cap-allowlist.tsv` via `--update`: the one grandfathered test-file entry (`session_manager/tests.rs`) drops out since it's now comfortably under the new cap; the remaining 7 entries are all production files, unaffected in policy and only re-frozen at their current (slightly drifted) SLOC.
- Updated the descriptive comments/docs that stated the old 1500 number: `CONTRIBUTING.md`, `.trusty-mpm/INSTRUCTIONS.md`, and the header comments in `scripts/check_line_cap.sh` / `.github/workflows/line-cap.yml`.
- **Deliberately did NOT rename** the job-level `name: 500-line file-size cap` in `.github/workflows/line-cap.yml` — verified via `gh api repos/.../branches/main/protection` that this exact string is a required status-check context. Renaming it here (without a coordinated branch-protection update) would strand every future PR in a permanently-pending required-check state. Updated the step-level name (not part of that contract) instead.

## Verification

```
$ bash scripts/check_line_cap_selftest.sh
PASS: sloc-normal.rs -> 7 SLOC (expected 7)
PASS: sloc-pathglob-doc.rs -> 6 SLOC (expected 6)
PASS: sloc-real-block-comment.rs -> 2 SLOC (expected 2)
PASS: sloc-nested-block-comment.rs -> 1 SLOC (expected 1)
PASS: sloc-trailing-comment.rs -> 4 SLOC (expected 4)
PASS: sloc-string-literal-slash-star.rs -> 2 SLOC (expected 2)
check_line_cap_selftest: all SLOC-counter regression cases passed.

```

Both run clean on a fresh `origin/main` checkout (`/private/tmp/wt-4074-linecap-testcap`), isolated from other in-flight worktree WIP.

## Test plan

- [x] `bash scripts/check_line_cap_selftest.sh` — 6/6 pass
- [x] `bash scripts/check_line_cap.sh` — 7 allowlisted, 0 violations
- [ ] CI green on this PR (Line cap job + full suite)

Not merging — leaving for CI/critic review per instructions.

### DIFF
--- a/.github/workflows/line-cap.yml
+++ b/.github/workflows/line-cap.yml
@@ -1,9 +1,12 @@
-# Line-cap gate for the trusty-tools workspace (issue #610).
+# Line-cap gate for the trusty-tools workspace (issue #610; dual-cap #1131,
+# TEST_CAP raised #4074).
 #
-# Enforces the 500-line file-size cap mechanically via a ratcheted allowlist:
-#   - a NEW tracked .rs file over 500 lines fails the job (cannot merge);
+# Enforces a dual SLOC file-size cap mechanically via a ratcheted allowlist:
+#   - a NEW tracked .rs file over its applicable cap (500 SLOC production /
+#     3000 SLOC test-benchmark) fails the job (cannot merge);
 #   - an allowlisted (grandfathered) file may not grow past its frozen budget;
-#   - when an allowlisted file drops to <= 500 lines its entry must be removed.
+#   - when an allowlisted file drops to <= its applicable cap its entry must
+#     be removed.
 #
 # Pure line counting — no Rust toolchain or build needed, so this is fast and
 # runs as its own lightweight job (matches ci.yml's checkout style/auth).
@@ -37,5 +40,5 @@ jobs:
       - name: SLOC counter selftest (regression fixtures)
         run: bash scripts/check_line_cap_selftest.sh
 
-      - name: Enforce 500-line file cap (ratcheted allowlist)
+      - name: Enforce SLOC file cap (500 prod / 3000 test, ratcheted allowlist)
         run: bash scripts/check_line_cap.sh
--- a/.line-cap-allowlist.tsv
+++ b/.line-cap-allowlist.tsv
@@ -1,6 +1,6 @@
 # .line-cap-allowlist.tsv — grandfathered files over the SLOC cap (issue #610).
 # Format: <relative/path><TAB><budget>  (budget = frozen max SLOC count; code lines only).
-# Dual cap: production source = 500 SLOC; test/benchmark files = 1500 SLOC.
+# Dual cap: production source = 500 SLOC; test/benchmark files = 3000 SLOC.
 # Test/benchmark = basename is tests.rs, ends with _test.rs or _tests.rs,
 #   or path contains /tests/ or /benches/ segment. All others = production.
 # SLOC excludes blank lines, // line comments, /// doc comments, //! inner-doc comments,
@@ -12,6 +12,5 @@ crates/trusty-git-analytics/src/collect/collector.rs	766
 crates/trusty-mpm/src/daemon/api.rs	1272
 crates/trusty-mpm/src/daemon/api/control_routes.rs	543
 crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs	845
-crates/trusty-mpm/src/session_manager/tests.rs	1516
 crates/trusty-search/src/main.rs	658
 crates/trusty-search/src/service/walker.rs	1167
--- a/.trusty-mpm/INSTRUCTIONS.md
+++ b/.trusty-mpm/INSTRUCTIONS.md
@@ -164,12 +164,13 @@ application/binary code and `thiserror` for library error types. Reserve
 `expect()` only for cases that are genuinely programmer errors (invariants that
 can never occur at runtime).
 
-🔴 **SLOC file size hard cap (MECHANICALLY ENFORCED, dual-cap since #1131):**
+🔴 **SLOC file size hard cap (MECHANICALLY ENFORCED, dual-cap since #1131,
+TEST_CAP raised #4074):**
 
 | File type | SLOC cap |
 |---|---|
 | Production source files | **500 SLOC** |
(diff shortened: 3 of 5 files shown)


bobmatnyc:
## Verdict: APPROVE

Independent adversarial review of the diff (not the PR description). CI is green; verified independently rather than trusted.

### What I checked

**1. Calibration data — sound, independently reproduced.** Re-ran the script's own `SLOC_AWK` counter (sourced from `scripts/lib/sloc_awk.sh`) against every file matching the script's own test/benchmark classification, at `origin/main` HEAD (`2f07d8ef`, one day newer than the PR's snapshot). Result: max real test file is 1516 SLOC (`crates/trusty-mpm/src/session_manager/tests.rs`, the grandfathered entry) with the next 8 files clustered 1400-1490 SLOC — matching the PR's claimed 1-file->1500 / 8-file 1400-1500 bands exactly. My total file count (716) and lower-band counts differ modestly from the PR's stated 689/69/5/12 (main has moved ~27 test files in the interim), but the load-bearing claim — no legitimate test file anywhere near 3000, largest is ~1516 — holds independently. 3000 leaves roughly 2x headroom over the largest real file today.

**2. Production cap untouched — confirmed.** `PROD_CAP=500` unchanged in `scripts/check_line_cap.sh`; the `.line-cap-allowlist.tsv` diff only drops the one now-under-cap test entry (`session_manager/tests.rs`, 1516) — all 7 remaining allowlist entries are production files and none had their frozen budgets raised.

**3. Required check name preserved — confirmed via branch protection API**, not just the PR's own claim:
```
$ gh api repos/bobmatnyc/trusty-tools/branches/main/protection | jq .required_status_checks.contexts
["Format check", "500-line file-size cap", "Clippy", "MSRV check (1.91)", "Test", "trusty-search daemon smoke test"]
```
`500-line file-size cap` is indeed a required context, and the diff only renames the step-level name (`Enforce SLOC file cap (500 prod / 3000 test, ratcheted allowlist)`), leaving the job-level `name: 500-line file-size cap` byte-for-byte unchanged. Correct call — renaming the job would have silently stranded every future PR in permanent `REVIEW_REQUIRED`/pending-check purgatory.

**4. Verification reproduced independently**, not trusted from the PR body: checked out the PR head in an isolated worktree and re-ran both commands myself — `check_line_cap_selftest.sh` (6/6 pass) and `check_line_cap.sh` (`7 allowlisted, 0 violations — OK`, exit 0). Matches the PR's claimed output.

### Findings

| MEDIUM | docs/reference/common-pitfalls.md | 58-59 | Still says "**500 SLOC for production files** and **1500 SLOC for test/benchmark files**" — stale post-merge. This file is the one explicitly linked from `.trusty-mpm/INSTRUCTIONS.md` (lines 669, 690) as the canonical "extended explanation" / "detailed explanation" for this exact cap, so a contributor following that link gets the wrong number even though the PR updated INSTRUCTIONS.md and CONTRIBUTING.md two lines above the link. | Update the SLOC numbers in `docs/reference/common-pitfalls.md` to 500/3000 in the same PR (or a fast immediate follow-up), since it's directly cross-referenced as authoritative detail. |

Lower-priority, not filed as findings (below the 80% confidence bar for "this PR's problem" — pre-existing historical/spec docs, not actively cross-linked as canon): `docs/specs/SPEC-MCPSVC-01-trusty-mcp-service.md:501` and the two `docs/trusty-installer/research/03-plan/*.md` references also still say 1500; worth a cleanup pass but these are dated research/spec artifacts, not the live contributor-facing doc surface.

### Notes

- The `#3995` classifier-gap (prefix vs. suffix `_tests.rs` naming) is correctly called out as out of scope in the PR description and is a separate, real issue — not something this PR needs to fix.
- No regression risk to the mechanical gate logic itself: this PR only changes two integer constants, one allowlist line removal, and comments/docs — the classification and SLOC-counting logic is byte-identical to `origin/main`.

No CRITICAL or HIGH findings. APPROVED for merge.

Issue #4072, opened 2026-07-27
Concurrent ~/.claude.json trust seeds lose entries (lost update + torn write) — recurring CI flake in tests_manifest_toggle_trust_3934


bobmatnyc:
Two seeders in `trusty-mpm` perform an **unsynchronised load → mutate → store cycle over the same `~/.claude.json`**:

- `core::home_trust_seed::preseed_home_trust` (workspace trust + renderer-upsell dismissal)
- `core::session_launch::settings::preseed_workspace_trust` (workspace trust + `enabledMcpjsonServers` approval)

The daemon runs them **concurrently** — `provisioner::workspace` calls the first and then, via `prepare_session`, the second, once per provisioned session, from independent tokio tasks.

## Two defects

1. **Lost update.** Neither seeder holds a lock across its read-modify-write cycle. The slower writer serialises the snapshot it read *before* the faster writer stored, silently deleting that session's whole `projects.<workspace>` entry — trust acceptance *and* its `enabledMcpjsonServers` approval. Nothing is logged; both writes "succeeded". Operator symptom: a session stalls on the exact "Do you trust the files in this folder?" / "new MCP servers found" dialog the seed exists to dismiss.

2. **Torn write.** `preseed_workspace_trust` used a truncating `std::fs::write` (unlike `preseed_home_trust`, which is already atomic). A concurrent reader can observe the half-written file; every reader in this crate treats malformed JSON as "skip, leave it alone", so the seed is silently lost rather than reported.

## Recurring CI flake (the reason this was found)

`core::session_launch::tests_manifest_toggle_trust_3934::prepare_session_excludes_spoofed_trusty_search_when_injector_disabled` failed the full-workspace `Test` job on **PR #4057** and **PR #4067** — neither touching `session_launch` — while passing in isolation. Panic:

```
thread ... panicked at .../tests_manifest_toggle_trust_3934.rs:59:10:
enabledMcpjsonServers is an array
```

i.e. `projects[<workspace>]` had vanished from the `.claude.json` the test had just seeded.

Mechanism: `trusty-mpm`'s lib test binary runs `#[serial]` tests that redirect the process-global `$HOME` to a temp dir **alongside non-serial tests that resolve `$HOME` at call time and drive the real provisioner**. Instrumenting both seeders over a full `cargo test -p trusty-mpm --lib` run identified 9 non-serial tests writing `~/.claude.json` via the ambient `$HOME`:

```
[...]

PR #4075, opened 2026-07-27
fix(trusty-mpm): serialize concurrent ~/.claude.json trust seeds (CI flake in tests_manifest_toggle_trust_3934)

bobmatnyc:
## Root cause

`core::home_trust_seed::preseed_home_trust` and `core::session_launch::settings::preseed_workspace_trust` both perform an **unsynchronised load → mutate → store cycle over the same `~/.claude.json`**, and the daemon runs them **concurrently** — `provisioner::workspace` calls the first and then, through `prepare_session`, the second, once per provisioned session, from independent tokio tasks.

Two defects fall out of that:

1. **Lost update.** Neither seeder holds a lock across its cycle. The slower writer serialises the snapshot it read *before* the faster writer's store, silently deleting that session's whole `projects.<workspace>` entry — trust acceptance **and** its `enabledMcpjsonServers` approval. Nothing is logged; both writes "succeeded". Operator symptom: a session stalls on the exact "Do you trust the files in this folder?" / "new MCP servers found" dialog the seed exists to dismiss.
2. **Torn write.** `preseed_workspace_trust` used a truncating `std::fs::write` (unlike `preseed_home_trust`, already atomic). A concurrent reader can observe the half-written file, and every reader in this crate treats malformed JSON as "skip, leave it alone" — so the seed is silently lost rather than reported.

This is a **production** concurrency bug in the daemon's provisioning path, not merely a test artifact. The CI flake is how it surfaced.

## The flake it explains

`core::session_launch::tests_manifest_toggle_trust_3934::prepare_session_excludes_spoofed_trusty_search_when_injector_disabled` failed the full-workspace `Test` job on **#4057** and **#4067** — neither touching `session_launch` — while passing cleanly in isolation:

```
thread '...prepare_session_excludes_spoofed_trusty_search_when_injector_disabled' panicked at
crates/trusty-mpm/src/core/session_launch/tests_manifest_toggle_trust_3934.rs:59:10:
enabledMcpjsonServers is an array
```

The `projects[<workspace>]` entry the test had just seeded was gone.

`trusty-mpm`'s lib test binary runs `#[serial]` tests that redirect the process-global `$HOME` to a temp dir **alongside non-serial tests that resolve `$HOME` at call time and drive the real provisioner**. Instrumenting both seeders across a full `cargo test -p trusty-mpm --lib` run identified 9 such tests (`connectors::tm::tests::create_session_full_lifecycle`, the three `provisioner::workspace::tests`, `core::standalone::load::tests::run_prepare_session_*`, `daemon::discovery::tests::discover_all_merges_results`, `core::deploy_validate::tests::repair_closes_gaps_on_incomplete_workspace`, `session_manager::naming_tests::reconcile_resolves_adopted_session_cwd_from_pane`). When one runs while a serial test has `$HOME` hijacked, its seed lands in — and clobbers — that test's temp `.claude.json`.

The `http://127.0.0.1:1` line in the failure report is **not** the cause: it is `trusty_common::mcp::memory_rpc::resolve_memory_base_url_or_unreachable`'s documented fail-open placeholder, printed on every run of that test (`prepare_session_inner` runs auto-catchup), passing runs included.

## Fix

- New `core::claude_json_guard`: one process-wide mutex both seeders hold across the **whole** read → mutate → store cycle, not just the write.
- `preseed_workspace_trust` now writes through `trusty_common::claude_config::write_json_atomic` (temp file + rename), matching `preseed_home_trust`.

No test was skipped, `#[ignore]`d, or weakened.

## Evidence the fix addresses the real bug

New `tests_claude_json_concurrency_4072.rs` drives the **real** seeders from 8 threads against one `.claude.json`, with a distinct workspace key per cycle (re-seeding one key in a loop self-heals a clobber and hides the bug; production writes each entry once).

Run against the **unmodified pre-fix production code** (lock removed, `std::fs::write` restored), both tests fail with exactly the two defects:

```
---- concurrent_seeds_preserve_every_workspace_entry stdout ----
a seeded .claude.json must always be valid JSON: Error("trailing characters", line: 94, column: 2)

---- concurrent_home_and_workspace_seeds_preserve_both_entries stdout ----
the two seeders clobbered 29 of 80 entries when run concurrently against one ~/.claude.json (issue #4072).
Lost: ["/workspace/home-seeded-3", ..., "/workspace/launch-seeded-7"]

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 3617 filtered out
```

With the fix:

```
test core::session_launch::tests_claude_json_concurrency_4072::concurrent_seeds_preserve_every_workspace_entry ... ok
test core::session_launch::tests_claude_json_concurrency_4072::concurrent_home_and_workspace_seeds_preserve_both_entries ... ok
```

## Gate

`cargo test -p trusty-mpm` (full, not `--lib`) — 37 test binaries, **6330 passed, 0 failed**, exit 0:

```
     Running unittests src/lib.rs (target/debug/deps/trusty_mpm-4dc25f985674e431)
test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 11.35s
     Running unittests src/bin/tm/main.rs (target/debug/deps/tm-0e5ae33f5cec548a)
test result: ok. 1231 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.60s
     Running unittests src/bin/tm/main.rs (target/debug/deps/trusty_mpm-5078a0b773b7dbe1)
test result: ok. 1231 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.40s
     ... (34 further integration-test binaries, all ok)
```

Six consecutive full-suite repeats of the previously flaky binary:

```
run 1: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 13.66s
run 2: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 11.97s
run 3: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 11.37s
run 4: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 12.72s
run 5: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 13.63s
run 6: test result: ok. 3613 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 16.40s
```

Other gates:

```
cargo clippy -p trusty-mpm --all-targets -- -D warnings   Finished `dev` profile (no warnings)
cargo fmt --all --check                                    exit 0
scripts/check_line_cap.sh          line-cap: 8 allowlisted, 0 violations — OK.
scripts/check_test_pointers.sh     test-pointers: 0 dangling pointers — OK.
scripts/check_sld.sh               scanned 47 spec doc(s) + 2887 code file(s); 0 error(s), 0 warning(s)
scripts/check_generation_artifacts.sh   generation-artifacts: 0 leaked artifacts — OK.
```

> Note: `tests/tm_hook_pm_guard.rs`'s two worktree-add cases fail when the checkout itself lives under `/private/tmp` (the #3955 guard denies the target). Reproduced on **clean `main` at that path**, unrelated to this change; the gate above was re-run from `.claude/worktrees/` per the project's worktree convention, where it is fully green.

## Known follow-up (deliberately out of scope)

Those 9 non-serial tests also pollute the **developer's real `~/.claude.json`** — 45 `tm-test-*` temp-workspace entries were found in one machine's config, each marked `hasTrustDialogAccepted: true`. They should get a hermetic `$HOME` and join the `#[serial]` group. The lock in this PR makes the flake impossible either way; the pollution is a separate defect and is recorded in #4072.

Part of #4072 (this PR fixes the race for two of the three known `.claude.json` seeders + the torn-write in one of them; #4072 stays open for the still-outstanding hermetic-`$HOME` test-isolation follow-up above, so it must not auto-close on merge)

Code review on this PR (WARN, two HIGH findings) also surfaced two more issues, filed separately: #4076 (a third `.claude.json` seeder, `preseed_managed_trust`, has the same unguarded race — one-line extension of `claude_json_guard`) and #4077 (`write_json_atomic`'s fixed temp-file path isn't safe for concurrent/cross-process writers).

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -48,6 +48,29 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **Concurrent workspace-trust seeds no longer clobber each other's
+  `~/.claude.json` entries** (closes
+  [#4072](https://github.com/bobmatnyc/trusty-tools/issues/4072)):
+  `core::home_trust_seed::preseed_home_trust` and
+  `core::session_launch::settings::preseed_workspace_trust` both perform a
+  load → mutate → store cycle over the same `~/.claude.json`, and the daemon
+  runs them concurrently — once per session it provisions, from independent
+  tokio tasks. Unsynchronised, the slower writer serialised the snapshot it
+  read *before* the faster writer's store, silently deleting that session's
+  whole `projects.<workspace>` entry (trust acceptance **and** its
+  `enabledMcpjsonServers` approval) with nothing logged; the operator then hit
+  the exact "Do you trust the files in this folder?" / "new MCP servers found"
+  dialog the seed exists to dismiss. Both seeders now hold one process-wide
+  lock (`core::claude_json_guard`) across the whole cycle, and
+  `preseed_workspace_trust` writes through
+  `trusty_common::claude_config::write_json_atomic` instead of a truncating
+  `std::fs::write` — a half-written file was observable by any concurrent
+  reader, and every reader in this crate treats malformed JSON as "skip,
+  leave it alone". This also fixes the recurring CI flake in
+  `tests_manifest_toggle_trust_3934` (red on PRs #4057 and #4067, green in
+  isolation), where a non-serial test resolving the process-global `$HOME`
+  clobbered a `#[serial]` test's temp `.claude.json`.
+
 - **Activity monitor no longer misreports a provider stream failure as a
   serialization error** (issue #3757): `LlmActivityClassifier::classify`
   matched only `ChatEvent::Delta` and discarded `ChatEvent::Error`, so once the
--- a/crates/trusty-mpm/src/core/claude_json_guard.rs
+++ b/crates/trusty-mpm/src/core/claude_json_guard.rs
@@ -1,0 +1,70 @@
+//! Process-wide mutual exclusion for read-modify-write cycles on the
+//! operator's `~/.claude.json`.
+//!
+//! Why (issue #4072): two independent seeders in this crate perform an
+//! unsynchronised load → mutate → store cycle against the SAME
+//! `~/.claude.json` file —
+//! [`crate::core::home_trust_seed::preseed_home_trust`] (workspace trust +
+//! renderer-upsell dismissal) and
+//! [`crate::core::session_launch::settings::preseed_workspace_trust`]
+//! (workspace trust + `enabledMcpjsonServers` approval). Both run inside the
+//! daemon, which provisions and spawns sessions CONCURRENTLY: the workspace
+//! provisioner calls the former and then, via `prepare_session`, the latter,
+//! once per session, from independent tokio tasks. Two overlapping cycles are
+//! a textbook lost update — the slower writer serialises the snapshot it read
+//! BEFORE the faster writer's store, so the faster writer's whole
+//! `projects.<workspace>` entry silently disappears. The operator's symptom is
+//! a session that stalls on the "Do you trust the files in this folder?" or
+//! "new MCP servers found" dialog it was supposed to have pre-dismissed, with
+//! nothing logged anywhere — both writes "succeeded".
+//!
+//! This also surfaced as a recurring CI flake: `trusty-mpm`'s lib test binary
+//! runs `#[serial]` tests that redirect the process-global `$HOME` to a temp
+//! dir alongside non-serial tests that resolve `$HOME` at call time and drive
+//! the real provisioner, so a foreign test's seed lands in — and clobbered —
+//! the serial test's temp `.claude.json`
+//! (`core::session_launch::tests_manifest_toggle_trust_3934` failed this way
+//! on PRs #4057 and #4067).
+//!
+//! What: [`lock`] hands out a guard on one process-wide mutex. Every seeder
+//! that read-modify-writes `~/.claude.json` holds it across the WHOLE cycle
+//! (read, mutate, write), never just the write. The guard is deliberately not
+//! keyed by path: both seeders always target the single `~/.claude.json` of
+//! whatever `$HOME` resolves to right now, so one global lock is exactly the
+//! granularity needed and is immune to two callers disagreeing about the path.
+//!
+//! SCOPE: this is an in-process lock. It does not (and is not meant to)
+//! serialise two separate `tm` processes racing the same file — that needs an
+//! OS file lock and has never been observed; the daemon is the only writer
+//! that concurrently seeds. Every write also goes through
+//! `trusty_common::claude_config::write_json_atomic`, so even a cross-process
+//! race can only ever lose an update, never leave a torn file behind.
+//!
+//! Test: `concurrent_seeds_preserve_every_workspace_entry` and
+//! `concurrent_home_and_workspace_seeds_preserve_both_entries`, in
+//! `tests_claude_json_concurrency_4072`.
+
+use std::sync::{Mutex, MutexGuard};
+
+/// The one process-wide `~/.claude.json` read-modify-write mutex.
+static CLAUDE_JSON_LOCK: Mutex<()> = Mutex::new(());
+
+/// Acquire the process-wide `~/.claude.json` read-modify-write lock.
+///
+/// Why: see the module doc — holding this across a seeder's whole load →
+/// mutate → store cycle is what makes two concurrent seeds merge instead of
+/// clobbering each other.
+/// What: locks [`CLAUDE_JSON_LOCK`], recovering from poisoning by taking the
+/// inner guard. Poison recovery is correct here because the mutex guards no
+/// in-memory invariant — its payload is `()`; a panicking seeder leaves the
+/// on-disk file either untouched or atomically replaced, so the next holder
+/// starts from a consistent read either way. Panicking instead would turn one
+/// unrelated seeder panic into a permanently un-seedable config for the rest
+/// of the process's life.
+/// Test: `concurrent_seeds_preserve_every_workspace_entry`,
+/// `concurrent_home_and_workspace_seeds_preserve_both_entries`.
+pub(crate) fn lock() -> MutexGuard<'static, ()> {
+    CLAUDE_JSON_LOCK
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner())
+}
--- a/crates/trusty-mpm/src/core/home_trust_seed.rs
+++ b/crates/trusty-mpm/src/core/home_trust_seed.rs
@@ -39,12 +39,29 @@ use anyhow::Context as _;
 /// using `trusty_common::claude_config::write_json_atomic`. All existing keys are
 /// preserved. Idempotent: no write when every target field already has the required
 /// value.
+///
+/// **Concurrency (issue #4072):** the whole read → mutate → write cycle runs
+/// under [`crate::core::claude_json_guard::lock`], because
+/// `session_launch::settings::preseed_workspace_trust` load-mutate-stores the
+/// SAME file and the daemon runs both concurrently, once per session it
+/// provisions. Without that lock the two silently clobber each other's
+/// `projects.<workspace>` entries.
 /// Test: `seeds_trust_dialog_accepted`, `seeds_fullscreen_upsell_count`,
-/// `is_idempotent`, `preserves_existing_keys`.
+/// `is_idempotent`, `preserves_existing_keys`,
+/// `concurrent_home_and_workspace_seeds_preserve_both_entries`.
 pub fn preseed_home_trust(workspace: &Path) -> anyhow::Result<()> {
     use serde_json::Value;
 
     let claude_json = home_claude_json()?;
+
+    // Issue #4072: hold the process-wide `~/.claude.json` lock across the WHOLE
+    // read → mutate → write cycle below, not just the write. This function and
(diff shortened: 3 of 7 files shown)


bobmatnyc:

Independent adversarial review (code + spec only). Zero CRITICAL, two HIGH → **WARN**: the change is correct and lands, but both HIGH findings should be tracked as follow-ups before they are forgotten.

### What I verified myself (not taken from the PR body)

I checked the branch out and re-ran the new tests, then re-ran them with **only the two `claude_json_guard::lock()` lines deleted** (atomic write left in place):

```
# with the fix, a3943424
test core::session_launch::tests_claude_json_concurrency_4072::concurrent_home_and_workspace_seeds_preserve_both_entries ... ok
test core::session_launch::tests_claude_json_concurrency_4072::concurrent_seeds_preserve_every_workspace_entry ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 3617 filtered out; finished in 0.75s

# same tree, locks removed
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 3617 filtered out; finished in 0.03s
```

So on reviewer question **(4)**: the reproduction is genuine, not two sequential calls dressed up as concurrency — `std::thread::scope` drives the real production seeders (8 × 40 distinct workspace keys, and the exact `preseed_home_trust` ∥ `preseed_workspace_trust` daemon pairing), and the tests fail without the fix and pass with it, on this machine, in my hands.

On **(2)**, scoping is right: the guard is taken at `home_trust_seed.rs:64` and `settings.rs:687`, i.e. **before** the `read_to_string` and released at end of function, so the whole read→mutate→write cycle is covered, not just the write — which is the bug this class of fix usually gets wrong. No re-entrancy: neither locked region calls anything that re-acquires the guard, so no self-deadlock. Poison recovery via `into_inner()` is correct for a `Mutex<()>` guarding only on-disk state.

On **(1)**, within this crate the only in-process writers of the **home** `~/.claude.json` are the two seeders now holding the lock (`home_trust_seed.rs:177`, `settings.rs:801`); `core::mcp_config` and `standalone::trust_seed` target `~/.trusty-tools/trusty-mpm/claude-config/.claude.json`, a different file. See HIGH-1 for the caveat.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/standalone/trust_seed.rs` | 128–260 | Third seeder with the identical unguarded load→mutate→store, on the file **every** managed session shares — not covered by the new guard | Add `let _guard = crate::core::claude_json_guard::lock();` at the top of `preseed_managed_trust` (the guard is deliberately not path-keyed, so this is one line) |
| HIGH | `crates/trusty-common/src/claude_config.rs` | 180–184, 294–296 | `write_json_atomic` uses a **fixed** temp name `<path>.tmp`, so it is atomic for readers but not between writers; the new module doc asserts a cross-process guarantee it does not have | Make the temp name unique (`<path>.<pid>.<nanos>.tmp`), or correct the SCOPE claim in `claude_json_guard.rs:41` |
| MEDIUM | *(process)* `#4072` | — | The deferred test-isolation work is recorded **only** in #4072, which this PR closes — the follow-up disappears on merge | File a separate issue for the 9 non-hermetic tests and link it from the PR body before merging |
| LOW | `crates/trusty-mpm/src/core/session_launch/settings.rs` | 778–779 | Write failure now surfaces as `PrepError::Deploy(String)` instead of `PrepError::Io { path, source }`, dropping the `io::ErrorKind` on the write most likely to fail on a real machine (disk full / permissions) | Acceptable given `write_json_atomic` returns `anyhow::Error`; consider `PrepError::Io` with a downcast if the kind ever matters |

### HIGH-1 — the guard covers `~/.claude.json`, but not the managed `.claude.json`

`core::standalone::trust_seed::preseed_managed_trust` (`trust_seed.rs:128`) does the same thing the two fixed seeders do — read at `:143`, mutate `projects.<workspace>` + `enabledMcpjsonServers`, `write_json_atomic` at `:259` — against `<managed_config_dir>/.claude.json`. It is called from `runtime/claude_code.rs:709` inside `prepare_managed_config` (`:612`), which runs on **every spawn and resume** (`:829`, `:973`, `:1074`), i.e. under exactly the daemon concurrency this PR's own root-cause analysis describes. It does not take the new lock.

Why this matters more, not less, than the home file: the managed config dir is a single shared `~/.trusty-tools/trusty-mpm/claude-config/.claude.json` for **all** managed sessions, so a lost update there drops another live session's whole trust entry — the same operator symptom (`"Do you trust the files in this folder?"` / `"new MCP servers found"`) that #4072 exists to eliminate, reached by the managed path instead of the home path.

This is pre-existing code and strictly outside #4072's literal wording, so it is not a merge blocker — but the PR builds precisely the primitive that closes it, in one line, and `core/mod.rs:28` already describes the guard as the lock "every `~/.claude.json` read-modify-write seeder holds", which will read as covering this one.

### HIGH-2 — `write_json_atomic` is not atomic between writers, and the new doc says it is

`claude_json_guard.rs:36-42` states:

> Every write also goes through `trusty_common::claude_config::write_json_atomic`, so even a cross-process race can only ever lose an update, never leave a torn file behind.

That is not true as implemented. `tmp_path()` (`claude_config.rs:294`) is `<path>.tmp` — a fixed name shared by every writer of that path:

```rust
let tmp = tmp_path(path);
std::fs::write(&tmp, serialized.as_bytes())?;
std::fs::rename(&tmp, path)?;
```

Two concurrent writers therefore contend on one temp file. This is not theoretical — it is exactly what my lock-removed run produced, and it is worth reading closely because it is *not* the lost-update failure the PR describes:

```
thread '<unnamed>' panicked at tests_claude_json_concurrency_4072.rs:105:26:
seeding a workspace must never fail: Deploy("write .../.claude.json:
rename .../.claude.json.tmp onto .../.claude.json")
```

Writer A renamed the shared temp away; writer B's rename then hit `ENOENT`. The other interleaving is worse: B renames while A is still inside `fs::write(&tmp, …)`, publishing a **partially written temp file onto the real path**. The old truncating `std::fs::write` produced a torn state that was transient and self-healed on the next write; a torn file published by `rename` is durable, and every reader in this crate treats malformed JSON as "skip, leave it alone" (`settings.rs:702-709`, `home_trust_seed.rs:80-86`) — so the config would stay un-seedable and silently so.

In-process this is fully hidden by the new mutex, which is why CI is green and why this is HIGH rather than CRITICAL. Cross-process it is live: `tm launch` calls `preseed_home_trust` in the **CLI** process (`bin/tm/commands/launch.rs:232`) while the daemon seeds from its own; `tm mcp add` writes the managed file through the same helper. Answering reviewer question **(3)** precisely: `rename(2)` on APFS/HFS+ and ext4 does give real atomicity **for readers**, so the single-writer guarantee holds; what does not hold is the multi-writer guarantee the doc claims, and the cause is the shared temp name, not the platform.

Minimum acceptable action: correct the SCOPE paragraph. Better: unique temp names in `trusty-common`, which also fixes the `tm mcp` ∥ daemon case for free.

### MEDIUM-1 — the deferral is sound, but its tracking issue is the one being closed

Reviewer question **(5)**: the deferral itself is defensible, and I verified the load-bearing half of the argument rather than trusting it — with merge-safe writes, a foreign seed lands as an *additional* `projects.<key>` entry and cannot remove the serial test's own key, so the `tests_manifest_toggle_trust_3934` flake is genuinely dead regardless of test isolation. The residual risk is real but modest and clearly not this PR's job: the 9 tests still mutate the developer's live `~/.claude.json` (mine currently holds 290 `projects` entries, `-rw-r--r--`, alongside `oauthAccount`), still write `hasTrustDialogAccepted: true` for throwaway temp paths, and now also churn an 88 KB `.bak` copy of it per test run.

The actual defect here is bookkeeping: the PR body says the pollution "is recorded in #4072" while the same PR says "Closes #4072". On merge the record closes with the issue and the follow-up is lost. Split it into its own issue and link it — a five-minute fix that is the difference between a deferral and a disappearance.

---

## Required changes (tracked, not merge-blocking)

1. File a standalone follow-up issue for the 9 non-hermetic tests (hermetic `$HOME` + join the `#[serial]` group) and reference it from the PR body, so it survives #4072 closing. **Do this before merge** — it costs nothing and is the only item that is irrecoverable afterwards.
2. Either give `write_json_atomic` a unique temp name or correct the cross-process claim at `claude_json_guard.rs:41`. A doc that overstates a concurrency guarantee is how the next writer of `.claude.json` gets talked out of taking a lock.
3. Take the same guard in `preseed_managed_trust` (`standalone/trust_seed.rs:128`), or open an issue for it.

## Notes

- Reviewed against the spec and the code; the implementer's reasoning in the PR body was deliberately re-derived rather than accepted, and the two central claims (lock spans the full cycle; tests fail pre-fix) were re-verified locally rather than read.
- The change is a genuine, well-scoped fix for a real production defect, and the module documentation is unusually good — the SCOPE section is the reason HIGH-2 was findable at all. That paragraph existing and being *slightly* wrong is a much better failure mode than it not existing.
- Not flagged: `~/.claude.json` ending up `0644` after an atomic write. `preseed_home_trust` already wrote through `write_json_atomic` before this PR, so the mode was already being reset on every seed; this PR does not change it.
- Not flagged: the guard blocks a caller on a `std::sync::Mutex`. `Provisioner::provision` is a sync fn, and the seeders were already doing blocking file I/O, so the delta is bounded contention, not a runtime stall.

Issue #4087, opened 2026-07-27
trusty-search: corpus-failed indexes return silently-empty 200s, and timeout-skipped indexes are dropped rather than parked cold


bobmatnyc:
## Summary

A live trusty-search daemon was observed serving with a degraded warm boot. Investigation (code read, no fix attempted) answers the question that matters operationally: **yes, in two different ways — and one of them is silent.**

Reported for tracking. Split out of #4086, which fixes the trusty-review-side misreporting only and deliberately does not touch trusty-search.

## Observed

```json
{"status":"degraded","version":"0.39.1","uptime_secs":3220,"embedder":"ready",
 "embedderd_rss_mb":16761,"rss_mb":5307,"rss_limit_mb":32768,"cpu_pct":125.178,"indexes":20,
 "warmboot_summary":{"indexes_corpus_failed":3,"indexes_failed":0,"indexes_lazy":0,
   "indexes_loaded":20,"indexes_skipped_timeout":11,"indexes_skipped_tcc":0,"warm_boot_degraded":true},
 "boot_reconcile":{"degraded":false,"up_to_date":20,"skipped_no_data":2}}
```

## Finding 1 — timeout-skipped indexes are DROPPED, not lazily loaded

A warm-boot scan timeout does **not** park the index for lazy load. It is registered in **neither** the registry **nor** the cold store, so it is gone for the remainder of that boot (only a restart or an explicit re-registration brings it back).

- `restore_indexes` splits discovered entries into eager/cold and only *cold* entries reach `cold_store.register_cold_entries` — `crates/trusty-search/src/commands/start/restore.rs:174-183`.
- The timeout counter comes from a separate step: `restore_one_index_bounded` wraps the actual restore (which is what performs registration) in a `tokio::time::timeout` — `crates/trusty-search/src/service/warm_boot/restore.rs:98-121`, registration at `crates/trusty-search/src/commands/start_restore.rs:162-170`.
- A `false` return is only tallied into `total_skipped_timeout` — `crates/trusty-search/src/commands/start/restore.rs:243-249, 276-283`. Nothing else happens to the entry.

Consequence: **11 indexes are absent from `list_indexes`**, which is built purely from `registry.list()` (`crates/trusty-search/src/service/server/indexes.rs:201-206`). `indexes_lazy: 0` in the payload above corroborates this — had they gone cold they would be counted there.

This part **fails loudly**, which is the good case:
[...]

Issue #3706, opened 2026-07-23
trusty-search: fold warm_boot_degraded (TCC-denial/scan-timeout/mass-index-loss) into /health's overall_status


bobmatnyc:
## Summary

Follow-up from the #3693 re-review (PR #3704, code-critic MEDIUM finding #4).

`trusty-search`'s `GET /health` handler (`crates/trusty-search/src/service/server/health.rs:510-515`) computes the top-level `overall_status` field as:

```rust
let overall_status =
    if warmboot_summary.indexes_corpus_failed > 0 || indexes_watcher_network_degraded > 0 {
        "degraded"
    } else {
        "ok"
    };
```

This only inspects two of the conditions that `WarmBootSummary::warm_boot_degraded` itself aggregates (see the field's own doc comment, `crates/trusty-search/src/service/server/state.rs:517-528`):

- `indexes_skipped_tcc > 0` (TCC/FDA denial)
- `indexes_skipped_timeout > 0` (scan timeout)
- `indexes_corpus_failed > 0` (corpus-open failure) — **this one alone is folded into `overall_status`**
- `indexes_loaded` < 80% of the prior-known count (mass index loss)

So a daemon with `warm_boot_degraded: true` purely from TCC-denial, a scan timeout, or a large fraction of missing indexes still reports the **top-level `status: "ok"`** — `warm_boot_degraded` is present in the `warmboot_summary` object either way, but the field consumers actually gate on (`status`) doesn't reflect it for those three cases.

## Impact

trusty-review's `context_gate` / `handle_health` / `review_health` / `console_metrics` (PR #3704, closes #3693) added `HealthResponse::is_serving()`, which trusts `status == "ok"` as an immediate "serving" signal and only consults `warmboot_summary.warm_boot_degraded` when `status == "degraded"`. Because `overall_status` never goes `"degraded"` for TCC-denial/scan-timeout/mass-index-loss, `is_serving()` never gets a chance to inspect `warm_boot_degraded` for those classes — a genuinely broken warm-boot of that shape is reported as "ok" / serving today, same as before #3693.

This is a **pre-existing gap** in trusty-search's own `/health` payload, not something introduced or made worse by #3693 — before #3693, `is_healthy()` had exactly the same blind spot (it also only checked the top-level `status` string). #3693 doesn't regress this; it just doesn't fix it either, since it was out of scope (trusty-review-side fix only).

## Suggested fix

[...]

Issue #3693, opened 2026-07-23
trusty-review context_gate treats trusty-search status=degraded as unreachable → skips all reviews (blocks 0.38.1 on EFS)


bobmatnyc:
## Summary

`trusty-review`'s `context_gate` treats a trusty-search `/health` `status: "degraded"` as **"unreachable"** and fail-closes, skipping the review entirely. Because `trusty-search` 0.38.1 intentionally reports `status: "degraded"` on any EFS/NFS-mounted deployment (it auto-disables its file watcher on network mounts), this makes it **impossible to upgrade trusty-search to 0.38.1** on such deployments while `trusty-review` is 0.9.2 — all PR reviews would be skipped org-wide.

## Verified evidence (2026-07-22, EFS-mounted box, trusty-search 0.38.1 + trusty-review 0.9.2)

- **trusty-search is fully functional but reports "degraded".** trusty-search 0.38.1 detects repos on a network mount and auto-disables its file watcher, flagging all indexes `indexes_watcher_network_degraded`, and therefore reports `/health status: "degraded"`. Search itself is 100% functional: HTTP 200, embedder ready, queries succeed, and reindex works (an apex reindex produced 1584 chunks). The "degraded" is a benign/intentional watcher-disable on NFS/EFS, not a service fault.

- **trusty-review's context_gate rejects any non-"ok" status.** `context_gate` treats ANY `health.status != "ok"` as "trusty-search unreachable". With `require_search=true` (the default) it FAIL-CLOSES:
  - logs `context_gate: trusty-search health is not 'ok' status=degraded`
  - logs `required-context gate: skipping review — refusing to review without code context`
  - returns verdict `UNKNOWN`
  - never calls the LLM.

- **Confirmed deterministic:** 6/6 health probes showed `deps.trusty_search.reachable=false`; an actual `trusty-review run --local-diff` returned "review skipped / UNKNOWN / 0 tokens".

## Impact

Any deployment on EFS/NFS repos (the production config here) **cannot upgrade trusty-search to 0.38.1** while trusty-review is 0.9.2 — PR reviews would be skipped org-wide.

## Distinct from #3658

This is **not** #3658 (the intermittent no-timeout `/health` probe HANG). This one is a **deterministic status-string gate**: trusty-search responds promptly and correctly with `status: "degraded"`, and the gate rejects that string.

## Suggested fix (either side — maintainer's choice)

[...]

PR #4089, opened 2026-07-27
fix: stop reporting a live trusty-search as unreachable, and label degraded verdicts

### LINKED ISSUE #4086: trusty-review reports a live trusty-search as "unreachable", and a degraded verdict is indistinguishable from a complete one
## Summary

`review_health` reports a **live, query-answering trusty-search daemon as `state: "unreachable"`**, and a review that runs against a partially-degraded search returns a verdict that is indistinguishable from a complete one to any programmatic MCP caller.

Two defects, one root cause area. Both are semantics bugs in `trusty-review`; **neither is a discovery/address bug** (see "What this is NOT" below).

## Evidence — captured at the same moment

`mcp__trusty-review__review_health`:

```json
{"deps":{"trusty_analyze":{"reachable":true,"required":false,"state":"ok"},
         "trusty_search":{"reachable":false,"required":true,"state":"unreachable"}},
 "dry_run":false,"inference":"ok","reviewer_model":"us.anthropic.claude-sonnet-4-6",
 "status":"degraded","version":"0.10.1"}
```

`mcp__trusty-search__search_health` — a LIVE, SERVING daemon:

```json
{"status":"degraded","version":"0.39.1","uptime_secs":3220,"embedder":"ready",
 "embedder_info":{"backend":"python","dimension":384,"model":"all-MiniLM-L6-v2","provider":"mps","quantized":false},
 "embedder_last_ok_secs_ago":0,"embedder_recent_timeout_count":0,
 "embedderd_rss_mb":16761,"rss_mb":5307,"rss_limit_mb":32768,"cpu_pct":125.178,"indexes":20,
 "warmboot_summary":{"indexes_corpus_failed":3,"indexes_failed":0,"indexes_lazy":0,
   "indexes_loaded":20,"indexes_skipped_timeout":11,"indexes_skipped_tcc":0,"warm_boot_degraded":true},
 "boot_reconcile":{"degraded":false,"up_to_date":20,"skipped_no_data":2}}
```

Uptime 3220 s, embedder ready, 20 indexes loaded, `indexes_failed: 0`. The daemon is up and answering.

## What this is NOT — discovery and addressing are fine

Ruled out with direct evidence:

- `trusty-review` resolves search via `ReviewConfig::search_url` = `TRUSTY_SEARCH_URL` or the default `http://localhost:7878` (`crates/trusty-review/src/config/mod.rs:421`). The live daemon listens on exactly `127.0.0.1:7878`, and `curl http://localhost:7878/health` returns `200` in 22 ms. The URL is correct.
- The probe **completed** — `state` was `"unreachable"`, not `"timeout"` — so it got a parseable body back.
- `trusty_analyze` reported `reachable: true` against `localhost:7879` in the same probe pass, proving loopback HTTP from the trusty-review process worked at that instant.
- **trusty-review performs no discovery for trusty-search at all.** Its only use of `trusty_common::read_daemon_addr` is to publish and read *its own* address for `trusty-review port` (`commands/port.rs`, `service/mod.rs`). It never resolves trusty-search's address from disk — that comes solely from `TRUSTY_SEARCH_URL` or the compiled-in default. It references no shared discovery helper (`grep` for `memory_rpc`, `resolve_memory_base_url`, `resolve_daemon_base_url`, `UNREACHABLE_PLACEHOLDER` in `crates/trusty-review/src/` returns nothing relevant).

### A shared-discovery-helper hypothesis was considered and discarded

An earlier read of this suspected a common broken discovery helper, on the strength of `127.0.0.1:1` appearing in CI output on #4067 alongside `trusty-memory daemon address not discovered`. **That is not evidence for this bug**, on two independent grounds:

1. That CI failure (`prepare_session_excludes_spoofed_trusty_search_when_injector_disabled`) has since been root-caused to something else entirely and fixed by **#4072** — concurrent `~/.claude.json` trust-seed clobbering; the actual panic was `enabledMcpjsonServers is an array`. The trusty-memory lines were incidental log noise in the same output.
2. Even setting that aside, the code does not support it: `127.0.0.1:1` is `UNREACHABLE_PLACEHOLDER` in `crates/trusty-common/src/mcp/memory_rpc.rs:61`, reachable only via `resolve_memory_base_url_or_unreachable`. That helper has callers in trusty-code and trusty-mpm and **none in trusty-review**.

These are unrelated bugs. The sentinel-port fail-open is a genuine smell on its own merits and is filed separately as #4092.

## Defect 1 (P1) — `reachable` reports a capability judgement under a transport name

`probe_deps` classified the dep with `HealthResponse::is_serving()`, which returned `false` whenever `status == "degraded" && warmboot_summary.warm_boot_degraded == true` (`crates/trusty-review/src/integrations/health.rs`). `probe_deps` then published that single boolean as **both** `reachable: false` **and** `state: "unreachable"` (`crates/trusty-review/src/service/handlers.rs`).

`warm_boot_degraded` is an OR over four structurally different conditions (TCC/FDA denial, warm-boot scan timeout, corpus-open failure, mass index loss). Since trusty-search **#3706** folded that whole aggregate into the top-level `status`, **any one** of them makes trusty-search answer `status: "degraded"`. Here it was 11 indexes skipped on a warm-boot scan timeout plus 3 with a failed corpus — none of which stop the daemon answering queries for the other 17 indexes.

Net effect: one unrelated repo's index failing to load on this host published "trusty-search is unreachable". An operator reasonably reads that word as "the daemon is down" and goes looking for a dead process that is running the whole time — which is exactly what happened here.

Reproduced deterministically: the verbatim live payload above, fed through the pre-fix `is_serving()`, yields `false`.

## Defect 2 (P1) — a degraded verdict is byte-identical to a complete one on the MCP envelope

`wrap_result` set `mcp_status` only for `infra_unavailable` (a Skip). A `ReviewStatus::Degraded` result — a real verdict produced **without** full code context — came back with `isError: false` and **no** envelope-level marker (`crates/trusty-review/src/mcp/tools.rs`). The degradation existed only inside the serialised `content[0].text` blob and as a Markdown banner in `review_body`.

So every programmatic consumer that inspects the envelope — the stable, non-prose part of the response — saw a context-free APPROVE and a fully-contexted APPROVE as the same thing. A caveat nobody can see is not a caveat. This repo has prior art for exactly this shape (a freshness guard nothing invoked; a bridge table silently frozen for months).

## Fix

**Defect 1** — separate the two questions that were sharing one boolean:

- `HealthResponse::serving_state() -> ServingState { Serving | Degraded(reason) | NotServing(reason) }`, reading the **individual** warm-boot counters rather than the aggregate flag, and building a reason that names each gap's query-time consequence.
- New `DepState::Degraded`; `reachable` becomes `state.is_reachable()` (`Ok | Degraded`) — it now means the transport worked, nothing more. A new `detail` field carries the reason.
- `compute_status` gates on `state == Ok` (not `reachable`), so the top-level `status: "degraded"` is unchanged — the fix corrects the lie without hiding the gap.
- #3693 is preserved exactly: a `"degraded"` status with a clean warm boot (the benign network-mount watcher disable) stays `Serving` / `reachable: true` / `status: "ok"`.

**Defect 2** — chose **stamp-the-output**, not hard-fail, based on the call sites:

- `context_gate` routes a serving-but-degraded search to `GateOutcome::Degraded(reason)`, which the runner already stamps as a `DEGRADED — NOT AUTHORITATIVE` banner plus `result.error`.
- `wrap_result` now emits `mcp_status: "degraded_context"` + `degraded_reason` on the envelope, with `isError` deliberately left `false`.

Rationale for (b) over (a): the review **did** run and its findings are real, so `isError: true` would make harnesses discard a useful verdict and would re-collapse "incomplete" into "failed" — the same conflation Defect 1 is about. Meanwhile hard-refusing every review because one of ~20 unrelated indexes on the host has a bad corpus is a false-positive machine that trains callers to ignore the gate. Proceed-and-label is the only option that is neither silent nor a blanket refusal. A genuine outage (embedder not ready, or a non-serving status) still hard-Skips.

## Known remaining gap (follow-up, not in this fix)

`/health` reports warm-boot counters **host-wide** and does not say *which* indexes are affected. If the review's own index is one of the corpus-failed ones, it gets an empty result set and only a banner. Closing that properly needs a per-index status probe (`GET /indexes/:id/status`) against the index the review actually uses, which means a new `SearchClient` trait method — deliberately out of scope here.

## Tests

- `health_response_live_warmboot_degraded_payload_is_reachable` — the verbatim live payload above; **fails on the pre-fix predicate**, passes after.
- `health_response_degraded_reason_names_corpus_failures` — the reason must name the silent failure mode.
- `dep_state_degraded_is_reachable`, `bounded_probe_degraded_carries_detail`.
- `degraded_warm_boot_gap_labels_review_instead_of_skipping` and `not_serving_search_still_skips` — both sides of the relaxed gate.
- `wrap_result_degraded_sets_sentinel_and_reason` — with a complete-review control, so the sentinel is a signal and not noise.

bobmatnyc:
Closes #4086

## The report vs. what was actually happening

The symptom was read as "the first review ran degraded because trusty-search was down." trusty-search was never down. At the moment `review_health` said `trusty_search: {reachable: false, state: "unreachable"}`, the daemon had been up 3220 s, its embedder was ready, 20 indexes were loaded, `indexes_failed: 0`, and it was answering queries.

The word "unreachable" was the whole problem: it is what sent someone looking for a dead process that was running the entire time.

## Root cause — one boolean answering two questions

`probe_deps` classified trusty-search with `HealthResponse::is_serving()` and published that single boolean as **both** `reachable` **and** `state`. `is_serving()` returned `false` whenever `status == "degraded" && warm_boot_degraded == true`.

`warm_boot_degraded` is an OR over four structurally different conditions (TCC/FDA denial, warm-boot scan timeout, corpus-open failure, mass index loss). trusty-search **#3706** folded that whole aggregate into the top-level `status`, so **any one** of them now makes the daemon answer `"degraded"`. In the observed payload it was 11 indexes skipped on a warm-boot scan timeout plus 3 with a failed corpus — none of which stop the daemon serving the other 17.

So: one unrelated repo's index failing to load on this host published "trusty-search is unreachable" for every review on the machine.

**Not a discovery or addressing bug**, ruled out with evidence:
- `search_url` resolves to `http://localhost:7878` (`config/mod.rs:421`); the live daemon listened on exactly `127.0.0.1:7878` and `curl /health` returned `200` in 22 ms.
- `state` was `"unreachable"`, not `"timeout"` — the probe completed and parsed a body.
- `trusty_analyze` reported `reachable: true` over the same loopback in the same probe pass.
- **trusty-review performs no discovery for trusty-search at all.** Its only `read_daemon_addr` use is publishing/reading *its own* address for `trusty-review port`. The search address comes solely from `TRUSTY_SEARCH_URL` or the compiled-in default.

A shared-discovery-helper hypothesis (prompted by `127.0.0.1:1` appearing in CI on #4067) was considered and **discarded on two independent grounds**: that CI failure was root-caused separately and fixed by #4072 (concurrent `~/.claude.json` trust-seed clobbering; the real panic was `enabledMcpjsonServers is an array`, and the trusty-memory lines were incidental log noise), and the code does not support it either — `127.0.0.1:1` is `UNREACHABLE_PLACEHOLDER` in `trusty-common/src/mcp/memory_rpc.rs:61`, reachable only via `resolve_memory_base_url_or_unreachable`, which has 8 callers in trusty-code/trusty-mpm and **none** in trusty-review. These are unrelated bugs. The sentinel-port fail-open is a real smell on its own merits and is filed separately as **#4092**.

## Raw proof

The verbatim live payload, fed through the **pre-fix** predicate:

```
test integrations::health::tests::health_response_live_warmboot_degraded_payload_is_reachable ... FAILED
panicked at crates/trusty-review/src/integrations/health.rs:599:9:
a live, embedder-ready, query-answering daemon must never be classified as
not-serving just because warm boot left some indexes behind (#4079)
```

Same test, same payload, after the fix:

```
test integrations::health::tests::health_response_live_warmboot_degraded_payload_is_reachable ... ok
```

## What changed

**Defect 1 — separate the two questions.**
- `HealthResponse::serving_state() -> ServingState { Serving | Degraded(reason) | NotServing(reason) }`, reading the **individual** warm-boot counters instead of the aggregate flag, and building a reason that names each gap's query-time consequence.
- New `DepState::Degraded`. `reachable` becomes `state.is_reachable()` (`Ok | Degraded`) — it now describes the transport and nothing else. A new `detail` field carries the reason to the caller.
- `compute_status` gates on `state == Ok`, **not** `reachable`, so top-level `status: "degraded"` is unchanged. The fix corrects the lie without hiding the gap.
- #3693 is preserved exactly: a `"degraded"` status with a clean warm boot (the benign network-mount watcher disable) stays `Serving` / `reachable: true` / `status: "ok"`. `degraded_reason()` returns `None` for that case rather than manufacturing a caveat nobody can act on.

**Defect 2 — stamp the output, don't fail loud.**
- `context_gate` routes a serving-but-degraded search to `GateOutcome::Degraded(reason)`, which the runner already stamps as a `DEGRADED — NOT AUTHORITATIVE` banner plus `result.error`.
- `wrap_result` emits `mcp_status: "degraded_context"` + `degraded_reason` on the envelope, with `isError` deliberately left `false`.

### Why (b) and not (a), from the call sites

`wrap_result`'s envelope is the only stable, non-prose part of an MCP response, and it was the one place a degraded verdict was invisible: `isError: false`, no `mcp_status`, the caveat buried in a serialised JSON blob and a Markdown banner. A context-free APPROVE and a fully-contexted APPROVE were byte-identical there.

Hard-failing was rejected in both directions. `isError: true` would make harnesses discard a verdict that genuinely ran and whose findings are real — re-collapsing "incomplete" into "failed", the same conflation Defect 1 is about. And refusing every review because one of ~20 unrelated indexes on the host has a bad corpus is a false-positive machine that trains callers to ignore the gate; the pre-fix code did exactly that on the `Hosted` surface. Proceed-and-label is the only outcome that is neither silent nor a blanket refusal.

A genuine outage still hard-Skips: embedder not ready, or any status outside `ok`/`degraded`. `not_serving_search_still_skips` guards that boundary, since relaxing the real outage is the risk this change carries.

## Known remaining gap (out of scope, noted in #4086)

`/health` reports warm-boot counters **host-wide** with no index identity. If the review's own index happens to be one of the corpus-failed ones it gets an empty result set and only a banner. Closing that needs a per-index probe (`GET /indexes/:id/status`) against the index the review actually uses, i.e. a new `SearchClient` trait method — deliberately not done here.

## Tests

Added:
- `health_response_live_warmboot_degraded_payload_is_reachable` — the verbatim live payload; **fails on the pre-fix predicate** (output above).
- `health_response_degraded_reason_names_corpus_failures` — the reason must name the silent failure mode and its consequence, not just say "degraded".
- `health_response_degraded_missing_summary_is_degraded_not_fatal`.
- `dep_state_degraded_is_reachable`, `bounded_probe_degraded_carries_detail` — a probe failure must not silently produce a bogus classification, and a reason must not be dropped.
- `probe_deps_degraded_search_is_reachable` — the same guarantee end-to-end through a real `SearchClient` returning the live payload, which is where the false "unreachable" was actually minted; also asserts the overall status stays `"degraded"` so correcting the lie does not hide the gap.
- `degraded_warm_boot_gap_labels_review_instead_of_skipping` + `not_serving_search_still_skips` — both sides of the relaxed gate.
- `wrap_result_degraded_sets_sentinel_and_reason` — includes a complete-review control asserting the sentinel is absent, so it is a real signal and not noise on every response.

Updated to the corrected semantics (no test disabled, ignored, or cfg-gated): `health_response_degraded_watcher_only_is_serving`, `warm_boot_summary_wire_shape_is_pinned` (now pins every mirrored counter, not just the aggregate), `dep_state_serialises_lowercase`, `wrap_result_degraded_stays_is_error_false`, `call_tool_review_diff_search_down_interactive_default_degrades`.

## Verification

```
cargo test -p trusty-review
test result: ok. 1545 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
test result: ok. 68 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo clippy -p trusty-review --all-targets -- -D warnings   → clean
cargo fmt --check                                            → clean
bash scripts/check_line_cap.sh                               → 7 allowlisted, 0 violations — OK
cargo check --workspace --all-targets                        → clean
```

Workspace check excludes `trusty-code-gui` / `trusty-mpm-gui`, which fail in this worktree on a pre-existing missing `ui/dist` tauri asset. Neither depends on trusty-review, and this branch touches only `crates/trusty-review/`.

The five ignored tests are pre-existing (live-daemon / network-gated); none were added or newly ignored here.

## Related

Defect 3 from the same investigation — trusty-search corpus-failed indexes returning silently-empty `200`s, and timeout-skipped indexes being dropped rather than parked cold — is reported separately in #4087 and deliberately **not** touched by this PR.

### DIFF
--- a/crates/trusty-review/CHANGELOG.md
+++ b/crates/trusty-review/CHANGELOG.md
@@ -9,6 +9,36 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **A live trusty-search was reported as `"unreachable"`, and a degraded verdict
+  was indistinguishable from a complete one** (closes
+  [#4086](https://github.com/bobmatnyc/trusty-tools/issues/4086)).
+  - **`reachable` now means the transport worked, nothing more.** `probe_deps`
+    published one boolean as both `reachable` and `state`, so a trusty-search
+    that was up, embedder-ready, and answering queries was reported as
+    `reachable: false, state: "unreachable"` whenever its warm boot had left any
+    index behind. Since trusty-search #3706 folded the whole
+    `warm_boot_degraded` aggregate into the top-level `status`, one unrelated
+    repo's index failing to load on the host was enough to trigger it — and
+    "unreachable" sends an operator hunting a process that is running.
+    `HealthResponse::serving_state()` now returns
+    `Serving`/`Degraded(reason)`/`NotServing(reason)` by reading the individual
+    warm-boot counters instead of the aggregate flag, a new `DepState::Degraded`
+    carries that middle case, and `reachable` covers `Ok | Degraded`. A new
+    `detail` field on each dep names the gap and its query-time consequence.
+    Top-level `status` still reports `"degraded"` (it gates on `state == Ok`),
+    so the fix corrects the misreport without hiding the gap, and the #3693
+    benign network-mount watcher-disable path is unchanged.
+  - **A degraded review is now detectable from the MCP envelope alone.** A
+    `Degraded` result — a real verdict produced WITHOUT full code context —
+    returned `isError: false` and no envelope marker, so any consumer reading
+    the envelope saw it as identical to a fully-contexted verdict; the caveat
+    lived only inside the serialised text blob and a Markdown banner.
+    `wrap_result` now emits `mcp_status: "degraded_context"` plus
+    `degraded_reason`. `isError` deliberately stays `false` — the review ran and
+    its findings are real. A serving-but-degraded search routes to
+    `GateOutcome::Degraded` (labelled review) rather than a blanket skip, while
+    a search that cannot serve at all still hard-Skips.
+
 - **Reliability cluster: fabricated findings, spec-as-implementation, and
   unsound verdicts** (closes
   [#4042](https://github.com/bobmatnyc/trusty-tools/issues/4042),
--- a/crates/trusty-review/src/integrations/health.rs
+++ b/crates/trusty-review/src/integrations/health.rs
@@ -114,71 +114,80 @@ pub struct HealthResponse {
     /// JSON bool (`true`/`false`) and JSON string (`"ready"`, `"loading"`, …).
     #[serde(default)]
     pub embedder: EmbedderState,
-    /// Warm-boot health summary (issue #3693).
+    /// Warm-boot health summary (issues #3693, #3706, #4079).
     ///
-    /// Why: trusty-search downgrades its top-level `status` to `"degraded"`
-    /// for two very different reasons when it does downgrade: (1) a benign,
-    /// intentional capability gap — the file watcher was auto-disabled on a
-    /// network-mounted (EFS/NFS) root, indexes stay fully
-    /// queryable/reindexable via the explicit `index-file`/`remove-file`
-    /// API — or (2) `indexes_corpus_failed > 0` — a corpus that failed to
-    /// open. `warmboot_summary.warm_boot_degraded` is `true` for reason (2)
-    /// (and also for TCC/FDA denial, a scan timeout, or a large fraction of
-    /// indexes missing after warm-boot — see trusty-search's own
-    /// `WarmBootSummary::warm_boot_degraded` doc), but as of trusty-search
-    /// 0.38.1 those latter three conditions do NOT flip the top-level
-    /// `status` to `"degraded"` on their own — trusty-search issue #3706
-    /// tracks folding them in. Until that ships, a warm-boot broken only by
-    /// TCC-denial / scan-timeout / mass-index-loss (no corpus-open failure)
-    /// reports `status: "ok"` and `is_serving` never gets to inspect this
-    /// field for it — a pre-existing gap, not introduced or worsened by
-    /// #3693 (the prior `is_healthy` had the identical blind spot, since it
-    /// too only ever looked at the top-level `status` string). Absent on
-    /// older trusty-search versions or transport errors, in which case
-    /// callers must treat degraded as "unknown, assume not serving" (see
-    /// `is_serving`'s doc comment).
+    /// Why: since trusty-search #3706 the top-level `status` is `"degraded"`
+    /// whenever ANY warm-boot condition tripped — a benign auto-disabled file
+    /// watcher on a network-mounted root, a corpus that failed to open, a
+    /// warm-boot scan timeout, a TCC/FDA denial, or mass index loss. The status
+    /// string alone therefore cannot distinguish "fine, one unrelated repo's
+    /// index did not load" from "queries against index X silently return
+    /// nothing". [`HealthResponse::serving_state`] reads the individual counters
+    /// in this summary to build an actionable reason instead of guessing from
+    /// the status string. Absent on older trusty-search versions or a partial
+    /// response, which is reported as degraded-with-unknown-cause rather than as
+    /// an outage — the daemon did answer the probe.
     /// Test: `health_response_degraded_watcher_only_is_serving`,
-    /// `health_response_degraded_warmboot_degraded_is_not_serving`,
-    /// `health_response_degraded_missing_summary_is_not_serving`.
+    /// `health_response_live_warmboot_degraded_payload_is_reachable`,
+    /// `health_response_degraded_missing_summary_is_degraded_not_fatal`.
     #[serde(default)]
     pub warmboot_summary: Option<WarmBootSummary>,
 }
 
-/// Minimal mirror of trusty-search's warm-boot health summary (issue #3693).
+/// Mirror of trusty-search's warm-boot health summary (issues #3693, #4079).
 ///
-/// Why: trusty-review only needs the single aggregate `warm_boot_degraded`
-/// flag to distinguish a benign "degraded" (network-mount watcher disable)
-/// from a genuine fault (TCC denial, scan timeout, corpus-open failure, or a
-/// large fraction of indexes missing) — see trusty-search's
-/// `WarmBootSummary::warm_boot_degraded` doc for the full list of what folds
-/// into it. Deliberately does not mirror the other counters: this consumer
-/// only ever branches on the one boolean, and unknown fields are discarded
-/// by serde (no `deny_unknown_fields`) so future additions on the
-/// trusty-search side never break parsing here.
-/// What: `warm_boot_degraded == true` means trusty-search itself considers
-/// its warm-boot state broken, regardless of the top-level `status` string.
+/// Why: the aggregate `warm_boot_degraded` flag alone is not actionable. It is
+/// an OR over four structurally different conditions (TCC/FDA denial, warm-boot
+/// scan timeout, corpus-open failure, and mass index loss), and since
+/// trusty-search #3706 folded that whole aggregate into the top-level `status`
+/// field, ANY one of them makes trusty-search report `status: "degraded"`.
+/// Branching on the aggregate alone therefore cannot tell a daemon that is
+/// silently answering queries wrong (corpus-open failure) from one that merely
+/// left some unrelated repo's index unloaded (scan timeout) — and reading it as
+/// "not serving" reported a live, query-answering daemon as unreachable
+/// (#4079). Mirroring the individual counters lets `serving_state` explain
+/// WHICH gap exists, in a message an operator can act on.
+/// What: the individual warm-boot counters plus the aggregate flag. Every field
+/// is `#[serde(default)]` and unknown fields are discarded (no
+/// `deny_unknown_fields`) so a trusty-search-side addition never breaks parsing
+/// here.
 /// Test: see `HealthResponse` test list above; `warm_boot_summary_wire_shape_is_pinned`
-/// below pins the exact field name this type deserialises against.
-#[derive(Debug, Clone, Deserialize, Serialize)]
+/// below pins the exact field names this type deserialises against.
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
 pub struct WarmBootSummary {
-    /// `true` when trusty-search's own warm-boot health check considers
-    /// itself degraded for a real reason (not merely a network-mount
-    /// watcher disable).
+    /// Indexes successfully restored into the registry at warm boot.
+    #[serde(default)]
+    pub indexes_loaded: u32,
+    /// Indexes skipped at warm boot because macOS TCC/FDA denied access to
+    /// their root. They are absent from the registry until the next boot.
+    #[serde(default)]
+    pub indexes_skipped_tcc: u32,
+    /// Indexes skipped at warm boot because their restore exceeded the
+    /// per-index deadline. Like the TCC case they are absent from the registry
+    /// entirely (not parked for lazy load), so a query naming one of them 404s
+    /// loudly and `search_all` simply does not fan out to it.
+    #[serde(default)]
+    pub indexes_skipped_timeout: u32,
+    /// Registered indexes whose restore was recorded as failed.
+    #[serde(default)]
+    pub indexes_failed: u32,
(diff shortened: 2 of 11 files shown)


bobmatnyc:
Merge record — **critic review deliberately waived on this PR, by explicit repo-owner authorization**. Recording the rationale here so this reads as a documented exception, not a skipped gate.

This PR *is* the fix for trusty-review's own degraded state. At merge time the live daemon reported `deps.trusty_search: {reachable: false, required: true}` while trusty-search was demonstrably alive and serving — the exact conflation of "degraded" with "not reachable" in `HealthResponse::is_serving()` that this change repairs. Any critic verdict produced against this branch right now would itself be generated without search-backed context, and would carry the very `DEGRADED — NOT AUTHORITATIVE` stamp this PR introduces. The gate cannot meaningfully review the tool it repairs.

Bob (repo owner) therefore authorized merging on **green CI alone**. All 22 required checks passed, including `Test` (24m26s), `Clippy`, `MSRV check (1.91)`, and the `trusty-search daemon smoke test`. Admin-squash-merge bypasses the bot-approval requirement only — never a red or pending check.

Follow-up: once this is deployed and `trusty_search.reachable` reports `true` again, normal critic gating resumes for subsequent PRs.

bobmatnyc:
Deployed and verified live on Bob's MacBook (merge commit `4e36002d`).

**How trusty-review runs here:** not the launchd daemon — `~/Library/LaunchAgents/com.trusty.trusty-review.plist` exists but is **not loaded** (`launchctl print` → service not found, nothing listening on 7890). It runs as per-session `trusty-review serve --stdio` children spawned from `.mcp.json` via `PATH` → `~/.cargo/bin/trusty-review`. Rebuilt with `cargo install --path crates/trusty-review --locked --force` (`--locked` deliberately, so `Cargo.lock` was not refreshed; `git status` clean afterward). Version is unchanged at 0.10.1 — the PR did not bump it — so the discriminator is the new `detail` field, which the pre-fix image cannot emit.

**Live A/B, same instant, same trusty-search process.** trusty-search was `status: "degraded"`, `warm_boot_degraded: true`, `indexes_corpus_failed: 1`, embedder ready, 26 indexes, answering queries:

Pre-fix image (a still-running stdio child from before the rebuild):
```json
"trusty_search": { "reachable": false, "required": true, "state": "unreachable" }
```

Redeployed binary, same moment:
```json
"trusty_search": {
  "reachable": true,
  "required": true,
  "state": "degraded",
  "detail": "serving but degraded (0 of 0 index(es) loaded): 1 index(es) failed to open their corpus — trusty-search answers queries against those indexes with an EMPTY result set and no error"
}
```

Top-level `status` stayed `"degraded"` in both — the fix corrects the misclassification without hiding the gap, as intended. The 30 `degraded`-matching tests pass from merged main (`30 passed; 0 failed`), including `health_response_live_warmboot_degraded_payload_is_reachable`.

**Caveat:** the ~10 `serve --stdio` children belonging to other live sessions are still running the pre-fix image and will keep reporting `unreachable`. They pick up the fix only when their owning session restarts its MCP connection; they were deliberately not killed.

Unrelated but found during this deploy: the `com.trusty.search` LaunchAgent had been down since 13:27:22Z (clean SIGTERM, and `KeepAlive = {SuccessfulExit: false}` means launchd never restarts it after a clean exit). Restarted via `launchctl kickstart`; it is serving again.

Issue #22, opened 2026-05-20
perf(trusty-search): cache tree-sitter parsers in rayon thread-locals to avoid repeated C FFI construction


bobmatnyc:
## Problem
`chunk_ast` reconstructs the tree-sitter parser (C FFI + grammar compilation) on every call. In a rayon `par_iter` over 128 files, each rayon worker thread rebuilds the parser for every file it processes.

## Fix
Use `thread_local!` storage in rayon workers to cache the parser per language per thread. Construct once per thread, reuse across files.

## Acceptance Criteria
- [ ] Tree-sitter parser is constructed at most once per rayon thread per language
- [ ] Parse output is identical
- [ ] `cargo test -p trusty-search` passes

PR #4090, opened 2026-07-27
fix: refuse cross-branch git push from managed worktrees (#2867)

### LINKED ISSUE #2867: bug(trusty-mpm): orphaned agent survived session pause and force-push-clobbered a foreign PR branch (PR #2863 incident)
## Timeline (all 2026-07-16/17 UTC)

- **22:19Z** — PR #2863 branch `feat-sld-policy-lint` created (head `089667a3`)
- **23:00Z** — legitimate push (`0c3d2485`)
- **23:08Z** — legitimate force-with-lease push, rebase (`f74ecdee`)
- **23:23Z** — legitimate push (`43e7df56`)
- **23:38:11Z** — **CLOBBER**: bare force-push moved head `43e7df56` → `4af68064` (a stale pre-fix rebase of the original SLD commits)
- **00:05Z** — restore force-with-lease back to `43e7df56`

## Root cause chain

1. A background agent dispatched before a session pause survived as an orphaned OS process — the pause logic assumed "background agents die at pause," but they don't always.
2. That orphaned agent had created an ad-hoc worktree `.base/.claude/worktrees/rebase-2863` with `branch.rebase-2863-work.merge = refs/heads/feat-sld-policy-lint` — i.e. its local branch `rebase-2863-work` was configured to track/merge into a **PR branch it did not own**.
3. 46 minutes after its rebase, the orphaned agent ran a bare `git push`, which (per the above tracking config) landed on `feat-sld-policy-lint` — the PR #2863 branch — clobbering the reviewed lineage.
4. No liveness/ownership guard exists between concurrent sessions/agents targeting the same PR branch.

## Evidence

- Reflog timestamps: rebase finished `22:51:39Z` (`.base/worktrees/rebase-2863/logs/HEAD`)
- GitHub activity API: `force_push` events on the PR branch matching the clobber timestamp
- `lsof`/`tmux` sweep performed post-hoc: no live process owner found for the orphaned worktree

## Second armed instance — found and defused (config-only)

A second worktree, `fix-2863-block` (created `23:58Z`, dirty working tree, same dangerous tracking config pattern — `branch.fix-2863-block.merge` pointed at the PR branch), was found in the same state. It has been defused **config-only**: its merge ref was retargeted to its own branch name (`refs/heads/fix-2863-block`). No working-tree content, commits, or the worktree itself were touched — one has uncommitted work that another session may still want.

## Proposed fixes (for triage)

- (a) `tm` should terminate or adopt background agent processes at session pause/exit, rather than assuming they die.
- (b) Agent-provisioned worktrees must never set `branch.<name>.merge` to a ref the agent doesn't own — set upstream to the same-name ref, or push with an explicit refspec only.
- (c) Consider a pre-push guard hook in agent worktrees that refuses pushes to `refs/heads/*` differing from the checked-out branch name.

bobmatnyc:
## What

Two independent guards against the PR #2863 clobber shape (#2867), where an
agent-created worktree whose local branch tracked a **foreign** PR branch ran a
bare `git push` and destroyed that PR's reviewed lineage 46 minutes after the
session paused.

### Part A — never leave a worktree armed with a foreign upstream

`daemon::managed_routes::inproject::configure_session_branch_tracking` wrote
`branch.<session>.merge = refs/heads/<default>` **first** and pinned
`push.default = current` **afterwards**, both best-effort. Any failure of the
pin — old git, unwritable base config, a sandbox blocking `git config` — left
the worktree tracking a branch it does not own with nothing keeping a bare push
off it.

The pin is now established first and **read back** from the effective config
stack (`git config --get push.default`, which resolves `config.worktree`
exactly as `git push` will — not the exit code of the write). When it cannot be
confirmed, **no upstream is written at all**; `git pull origin <default>` still
works, which is a far cheaper failure than an armed worktree.

Audit of the other two upstream writers:

| Site | Finding |
|---|---|
| `provisioner/workspace.rs` `worktree_add` | **Clean** — `git fetch origin +<ref>:refs/heads/<branch>` then `git worktree add <path> <branch>` never writes an upstream (the branch already exists locally, so no DWIM/`guessRemote` path is taken). Now locked in by a test that enumerates **every** `branch.*.merge` key and asserts each names its own branch. |
| `core/session_launch/worktree_sync.rs:438` | **Test fixture only**, not production. Updated so it mirrors the new pinned production shape rather than codifying the un-pinned one. |
| grep for other `set-upstream` / `branch.*.merge` writers | No others exist in the workspace. |

### Part B — a `pre-push` hook that refuses a cross-branch push

The #2867 worktree was created by an **agent's own** `git worktree add`, not by
tm's provisioner, and the post-mortem's `lsof`/`tmux` sweep found no live
process owner. A provisioner-only or agent-lifecycle fix cannot cover that. A
`pre-push` hook fires regardless of who runs the push.

`crates/trusty-mpm/src/assets/hooks/pre-push` refuses any push whose destination
branch differs from the checked-out branch (including a **delete** of a branch
you are not on). Non-branch destinations (tags, notes) pass through.
`TM_ALLOW_CROSS_BRANCH_PUSH=1` permits a deliberate cross-ref push.

**Delivery mechanism — chosen from empirical evidence, not from the plan.** The
plan proposed `core.hooksPath` on the base clone. Verified in a throwaway
scratch repo (git 2.54.0):

- `$GIT_COMMON_DIR/hooks` is shared by the base checkout, provisioner-created
  linked worktrees, **and ad-hoc `git worktree add` worktrees** alike — the hook
  fires with `cwd` = each worktree's own toplevel. **No config change needed.**
- An **absolute** `core.hooksPath` in the base's common config *is* inherited by
  all worktrees, and stays inherited when `extensions.worktreeConfig = true`
  (which `configure_session_branch_tracking` sets).
- A **relative** `core.hooksPath` silently does **not** work: it resolves against
  each worktree's own toplevel, so the hook never fires from a linked worktree.

So the chosen mechanism is the plain shared hooks directory — strictly simpler
than the proposed `core.hooksPath` and with no shared-config write. Installation
**refuses rather than overwrites** when a foreign `pre-push` hook or a
`core.hooksPath` redirect is present, so it never fights husky/lefthook.

**Scope: FRESH clones only.** The guard is installed from
`RealGitBackend::ensure_base_checkout` and `inproject::ensure_base_clone` on the
clone path. Existing bases — including this repo's `.base`, shared by ~70 live
worktrees — are deliberately untouched; retrofitting one is an operator
decision, not a session side effect.

## Tests

All real git, no mocks, nothing ignored or cfg-gated.

- `provisioner_worktree_add_writes_no_foreign_branch_merge` — enumerates every
  `branch.*.merge` after a real provision and asserts none is foreign.
- `push_pin_detection_matches_effective_config` — two linked worktrees off one
  repo; only the pinned one reports pinned, proving the detector reads
  worktree-scoped config.
- `create_session_worktree_sets_pull_upstream_and_worktree_scoped_push` —
  extended to assert the **effective** `push.default`, not just the
  `--worktree` key.
- `ensure_base_checkout_installs_push_guard` — the guard lands executable in the
  resolved hooks dir.
- `tests/push_guard_hook.rs` (3 tests) — real pushes to a real bare remote:
  cross-branch push refused with the victim ref byte-identical, foreign-branch
  delete refused, same-name push and bare push still work, override env var
  permits the push and the ref actually moves, and an **ad-hoc worktree armed
  exactly as #2867 found it** (`branch.<n>.merge` → foreign branch,
  `push.default = upstream`) has its bare `git push --force` refused.

Notable finding from writing that last test: git's **default** `push.default =
simple` already refuses a bare push whose upstream name differs — so the
original incident required either a non-default `push.default` or an explicit
`HEAD:other` refspec. The guard covers both, and does not rely on that default
holding.

## Quality gate

```
cargo fmt --all -- --check          → exit 0
cargo test -p trusty-mpm            → exit 0
  lib: test result: ok. 3622 passed; 0 failed; 6 ignored
  tests/push_guard_hook.rs: ok. 3 passed; 0 failed; 0 ignored
  (38 test binaries, 0 failures)
scripts/check_line_cap.sh           → 7 allowlisted, 0 violations — OK
```

Closes #2867

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -11,15 +11,41 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **Cross-branch `git push` guard for every worktree of a managed base clone**
   ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): a bundled
-  `pre-push` hook now refuses any push whose destination branch differs from the
-  checked-out branch — the shape that clobbered PR #2863's reviewed lineage. It
+  `pre-push` hook now refuses any push that would land one branch's history on
+  a remote branch of a DIFFERENT name — the shape that clobbered PR #2863's
+  reviewed lineage. One rule, applied identically whether `HEAD` is attached or
+  detached: the source branch name must equal the destination branch name. It
   is installed into `$GIT_COMMON_DIR/hooks` at base-clone time, which every
   worktree of that base shares, so it also covers ad-hoc `git worktree add`
   worktrees an agent creates for itself (the actual incident path, which no
-  other trusty-mpm code path ever sees). Installation refuses rather than
-  overwrites when a foreign `pre-push` hook or a `core.hooksPath` redirect is
-  present, so it never fights husky/lefthook. Set
+  other trusty-mpm code path ever sees). Three shapes are deliberately out of
+  scope and pass through: non-branch destinations (tags, notes), remote-branch
+  deletes (they move no history onto a branch, no config can make a bare push
+  perform one, and post-merge cleanup is routine), and an anonymous source
+  pushed from a detached `HEAD` (`git push origin <sha>:refs/heads/<name>` is
+  the standard way to rescue work out of a detached worktree, and git itself
+  refuses a bare push while detached, so there is no accidental path to guard).
+  Installation refuses rather than overwrites whenever the existing `pre-push`
+  is not provably ours — a foreign hook, a symlink, a `core.hooksPath`
+  redirect, or a file that cannot be read at all (invalid UTF-8, a permissions
+  error) — so it never fights husky/lefthook, and a hook that cannot be
+  inspected is never destroyed. The write is atomic (temp file + `rename`)
+  because that one file is executed by every worktree of the clone. Set
   `TM_ALLOW_CROSS_BRANCH_PUSH=1` for a deliberate cross-branch push.
+
+- **`tm repair push-guard` retrofits that guard onto an already-provisioned
+  clone, and `tm doctor` reports when one is missing**
+  ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): the guard
+  installs automatically only on the CLONE path, so every base clone that
+  existed before it shipped — exactly the old, many-worktree clones the
+  incident happened in — was silently unprotected with no supported way to fix
+  that. `tm doctor` now carries a `push_guard` check that warns when the
+  current project's clone has no guard (or an older revision) and names the
+  retrofit command; `tm repair push-guard` performs it. The retrofit is
+  idempotent, writes into the clone's shared hooks directory so ONE invocation
+  from ANY worktree covers them all with no git config write, honours the same
+  never-overwrite-a-foreign-hook refusal, and offers `--dry-run` to report
+  without touching a file that live sessions are executing.
 
 - **`tm hook --pm-guard` now hard-blocks `git worktree add` targeting `/tmp`,
   `/private/tmp`, `/var/folders`, `$TMPDIR`, or the harness scratchpad**
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -11,20 +11,28 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **Cross-branch `git push` guard for every worktree of a managed base clone**
   ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): a bundled
-  `pre-push` hook now refuses any push that would land one branch's history on
-  a remote branch of a DIFFERENT name — the shape that clobbered PR #2863's
-  reviewed lineage. One rule, applied identically whether `HEAD` is attached or
-  detached: the source branch name must equal the destination branch name. It
-  is installed into `$GIT_COMMON_DIR/hooks` at base-clone time, which every
-  worktree of that base shares, so it also covers ad-hoc `git worktree add`
-  worktrees an agent creates for itself (the actual incident path, which no
-  other trusty-mpm code path ever sees). Three shapes are deliberately out of
-  scope and pass through: non-branch destinations (tags, notes), remote-branch
-  deletes (they move no history onto a branch, no config can make a bare push
-  perform one, and post-merge cleanup is routine), and an anonymous source
-  pushed from a detached `HEAD` (`git push origin <sha>:refs/heads/<name>` is
-  the standard way to rescue work out of a detached worktree, and git itself
-  refuses a bare push while detached, so there is no accidental path to guard).
+  `pre-push` hook now refuses any push that would OVERWRITE AN EXISTING remote
+  branch with the history of a branch of a DIFFERENT name — the shape that
+  clobbered PR #2863's reviewed lineage. One rule, applied identically whether
+  `HEAD` is attached or detached: updating an existing branch requires the
+  source branch name to equal the destination branch name. It is installed into
+  `$GIT_COMMON_DIR/hooks` at base-clone time, which every worktree of that base
+  shares, so it also covers ad-hoc `git worktree add` worktrees an agent
+  creates for itself (the actual incident path, which no other trusty-mpm code
+  path ever sees). Three shapes are deliberately out of scope and pass through:
+  non-branch destinations (tags, notes); **creating** a branch that does not
+  exist on the remote yet (git reports an all-zero `<remote sha>`; there is no
+  lineage to destroy, and `git push origin <sha>:refs/heads/<new-name>` is the
+  standard way to rescue work out of a detached worktree); and remote-branch
+  deletes — not because they cannot happen by accident, but because a deleted
+  GitHub branch is recoverable (the PR retains its commits at
+  `refs/pull/<N>/head`) whereas a force-clobbered lineage is not, post-merge
+  cleanup is routine, and server-side branch protection is the right layer to
+  police deletion. Note that a configured `remote.<name>.push` refspec makes a
+  **bare** `git push` land on an arbitrary branch — even from a detached
+  `HEAD`, where `push.default` is never consulted — which is why the exemption
+  is keyed on create-vs-update rather than on `HEAD` state or on how the push
+  was spelled.
   Installation refuses rather than overwrites whenever the existing `pre-push`
   is not provably ours — a foreign hook, a symlink, a `core.hooksPath`
   redirect, or a file that cannot be read at all (invalid UTF-8, a permissions
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -9,6 +9,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **Cross-branch `git push` guard for every worktree of a managed base clone**
+  ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): a bundled
+  `pre-push` hook now refuses any push whose destination branch differs from the
+  checked-out branch — the shape that clobbered PR #2863's reviewed lineage. It
+  is installed into `$GIT_COMMON_DIR/hooks` at base-clone time, which every
+  worktree of that base shares, so it also covers ad-hoc `git worktree add`
+  worktrees an agent creates for itself (the actual incident path, which no
+  other trusty-mpm code path ever sees). Installation refuses rather than
+  overwrites when a foreign `pre-push` hook or a `core.hooksPath` redirect is
+  present, so it never fights husky/lefthook. Set
+  `TM_ALLOW_CROSS_BRANCH_PUSH=1` for a deliberate cross-branch push.
+
 - **`tm hook --pm-guard` now hard-blocks `git worktree add` targeting `/tmp`,
   `/private/tmp`, `/var/folders`, `$TMPDIR`, or the harness scratchpad**
   (worktree-hygiene follow-up to
@@ -57,6 +69,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   env-resolved managed session id exists, and if the session still hasn't
   settled it prints an honest "still settling" message instead of the
   misleading non-git hint.
+
+- **A session worktree's branch is no longer left tracking the default branch
+  unless a bare `git push` is provably confined to its own branch**
+  ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)):
+  `configure_session_branch_tracking` used to set
+  `branch.<session>.merge = refs/heads/<default>` FIRST and pin
+  `push.default = current` afterwards, both best-effort — so any failure of the
+  pin (old git, unwritable base config, sandboxed `git config`) left the
+  worktree armed with a foreign upstream. The pin is now established first and
+  read back from the effective config stack; when it cannot be confirmed the
+  worktree is left with no upstream at all (`git pull origin <default>` still
+  works) rather than an upstream it does not own.
 
 - **Concurrent workspace-trust seeds no longer clobber each other's
   `~/.claude.json` entries** (closes
--- a/crates/trusty-mpm/src/assets/hooks/pre-push
+++ b/crates/trusty-mpm/src/assets/hooks/pre-push
@@ -1,5 +1,5 @@
 #!/bin/sh
-# trusty-mpm-push-guard: v2 (#2867)
+# trusty-mpm-push-guard: v3 (#2867)
 #
 # Why: issue #2867 — an agent worktree whose local branch was configured with
 #   `branch.<local>.merge = refs/heads/<someone-else's-PR-branch>` ran a bare
@@ -8,33 +8,49 @@
 #   `git push origin HEAD:other-branch` does it too), and the offending push
 #   may come from a process trusty-mpm never created. A pre-push hook is the
 #   only guard that fires regardless of WHO runs the push.
-# What: refuses a push that would land the history of one branch on a remote
-#   branch of a DIFFERENT name. One rule, applied identically whether HEAD is
-#   attached or detached:
+# What: refuses a push that would OVERWRITE AN EXISTING remote branch with the
+#   history of a branch of a DIFFERENT name. One rule, applied identically
+#   whether HEAD is attached or detached:
 #
-#     source branch name must equal destination branch name
+#     updating an existing branch requires source name == destination name
 #
 #   where the source name is the pushed `refs/heads/<name>`, or — when the
 #   source is anonymous (`HEAD`, a raw sha) — the checked-out branch name.
+#   An anonymous source from a DETACHED HEAD has no name, so it can only
+#   CREATE, never update.
 #   Three shapes are deliberately OUT OF SCOPE and pass through untouched:
 #     * non-branch destinations (tags, notes, refs/pull) — cannot clobber a
 #       PR's branch lineage;
-#     * branch DELETES (`git push origin --delete X`, `git push origin :X`) —
-#       they move no history onto a branch, no git config can make a bare
-#       push perform one, so they are never the #2867 accident; post-merge
-#       remote cleanup is routine and a client hook is the wrong place to
-#       police it (branch protection is the right one);
-#     * an anonymous source pushed from a DETACHED HEAD — there is no
-#       checked-out branch whose identity could be violated, git itself
-#       refuses a bare `git push` while detached ("fatal: You are not
-#       currently on a branch"), so every such push is an explicit operator
-#       refspec, and `git push origin <sha>:refs/heads/<name>` is the
-#       standard way to rescue work out of a detached worktree.
+#     * CREATING a branch that does not exist on the remote yet — git reports
+#       an all-zero <remote sha> for these. There is no lineage there to
+#       destroy, so a create can never be the #2867 damage, and `git push
+#       origin <sha>:refs/heads/<new-name>` is the standard way to rescue work
+#       out of a detached worktree;
+#     * branch DELETES (`git push origin --delete X`, `git push origin :X`).
+#       NOT because they cannot happen by accident — they can, see the
+#       warning below — but because a deleted GitHub branch is RECOVERABLE
+#       (the PR retains its commits at `refs/pull/<N>/head`, and the branch
+#       can be re-pushed from any clone) whereas a force-clobbered lineage is
+#       not. Post-merge remote cleanup is routine, and the right layer to
+#       police deletion is branch protection on the server, not a hook the
+#       pusher can remove.
 #   Set TM_ALLOW_CROSS_BRANCH_PUSH=1 to perform a deliberate cross-branch push.
+#
+# WARNING — do NOT reason from "git refuses a bare push while detached". That
+#   `fatal: You are not currently on a branch` comes from `push.default`
+#   resolution ONLY, and git never consults `push.default` when a configured
+#   `remote.<name>.push` refspec supplies the destination. Proven against this
+#   very hook: `git config remote.origin.push '+HEAD:refs/heads/victim'`, then
+#   a BARE `git push` from a detached HEAD, clobbered `victim` with no refspec
+#   typed by anyone. The same mechanism makes a bare `git push` perform a
+#   DELETE (`remote.origin.push = ':refs/heads/victim'`). This is exactly why
+#   the exemption above is keyed on CREATE-vs-UPDATE, which git states in the
+#   <remote sha> field, and never on HEAD state or on how the push was spelled.
 # Test: crates/trusty-mpm/tests/push_guard_hook.rs (real temp repos and a real
-#   bare remote: refuses the cross-branch push attached AND detached, permits
-#   the same-name push in both states, permits deletes and tags, permits with
-#   the override).
+#   bare remote: refuses the cross-branch push attached AND detached, refuses
+#   the `remote.<name>.push` detached-bare-push exploit above, permits the
+#   same-name push and the create in both states, permits deletes and tags,
+#   permits with the override).
 #
 # Note: git aborts the ENTIRE push when a pre-push hook exits non-zero, so a
 # multi-refspec push containing one refused ref sends none of them. That is
@@ -52,10 +68,11 @@ head_branch=$(git symbolic-ref --quiet --short HEAD 2>/dev/null) || head_branch=
 
 refused=0
 refused_branch=""
+rescue_sha=""
 # `|| [ -n "$local_ref" ]` keeps a final line with no trailing newline from
 # being silently dropped: POSIX `read` returns non-zero at EOF even when it
 # read a partial line, and dropping it here would fail OPEN.
-while read -r local_ref _local_sha remote_ref _remote_sha || [ -n "$local_ref" ]; do
+while read -r local_ref local_sha remote_ref remote_sha || [ -n "$local_ref" ]; do
 	# Only branch destinations can clobber a PR lineage.
 	case "${remote_ref:-}" in
 	refs/heads/*) ;;
@@ -63,6 +80,7 @@ while read -r local_ref _local_sha remote_ref _remote_sha || [ -n "$local_ref" ]
 	esac
 	dest_branch=${remote_ref#refs/heads/}
 
+	src_branch=""
 	case "${local_ref:-}" in
 	refs/heads/*)
 		# A named local branch: compare its name to the destination.
@@ -74,33 +92,50 @@ while read -r local_ref _local_sha remote_ref _remote_sha || [ -n "$local_ref" ]
 		;;
 	*)
 		# Anonymous source (`HEAD`, a raw sha). Attached, it IS the
-		# checked-out branch; detached, it has no name to defend.
-		if [ -z "$head_branch" ]; then
-			continue
-		fi
+		# checked-out branch; detached, it has no name at all — which
+		# is why the create check below, NOT the HEAD state, decides.
 		src_branch=$head_branch
 		;;
 	esac
 
-	if [ "$src_branch" = "$dest_branch" ]; then
+	# A CREATE cannot destroy a lineage that does not exist yet. Git marks
+	# one with an all-zero <remote sha>. Checked BEFORE the name comparison
+	# so it applies identically to a named and an anonymous source, and to
+	# an attached and a detached HEAD — no asymmetry to reason about.
+	case "${remote_sha:-}" in
+	'' | *[!0]*) ;;  # absent or non-zero: the destination branch exists
+	*) continue ;;   # all zeros: a create
+	esac
+
+	if [ -n "$src_branch" ] && [ "$src_branch" = "$dest_branch" ]; then
 		continue
 	fi
 
 	refused=1
-	refused_branch=$src_branch
+	if [ -n "$src_branch" ]; then
+		refused_branch=$src_branch
+	else
+		rescue_sha=$(printf '%.8s' "$local_sha")
+	fi
 	echo "pre-push: REFUSED cross-branch push (trusty-mpm push guard, #2867)" >&2
 	echo "  checked-out branch : ${head_branch:-<detached HEAD>}" >&2
 	echo "  pushing            : $local_ref" >&2
-	echo "  which is branch    : $src_branch" >&2
-	echo "  onto destination   : $remote_ref" >&2
+	echo "  which is branch    : ${src_branch:-<none — detached HEAD>}" >&2
+	echo "  onto EXISTING      : $remote_ref" >&2
 done
 
 if [ "$refused" -ne 0 ]; then
 	echo "" >&2
-	echo "A push landing on a branch of a different name is how PR #2863's" >&2
-	echo "reviewed lineage was clobbered (#2867). Push onto the matching" >&2
-	echo "branch instead:" >&2
-	echo "  git push origin refs/heads/$refused_branch:refs/heads/$refused_branch" >&2
+	echo "Overwriting an EXISTING branch from a branch of a different name is" >&2
+	echo "how PR #2863's reviewed lineage was clobbered (#2867)." >&2
+	if [ -n "$refused_branch" ]; then
+		echo "Push onto the matching branch instead:" >&2
+		echo "  git push origin refs/heads/$refused_branch:refs/heads/$refused_branch" >&2
+	else
+		echo "HEAD is detached, so there is no branch name to match. Creating a" >&2
+		echo "NEW branch is always allowed — rescue the work onto one:" >&2
+		echo "  git push origin HEAD:refs/heads/rescue-$rescue_sha" >&2
+	fi
 	echo "If this cross-branch push is deliberate, re-run it with:" >&2
 	echo "  TM_ALLOW_CROSS_BRANCH_PUSH=1 git push ..." >&2
 	exit 1
--- a/crates/trusty-mpm/src/assets/hooks/pre-push
+++ b/crates/trusty-mpm/src/assets/hooks/pre-push
@@ -1,5 +1,5 @@
 #!/bin/sh
-# trusty-mpm-push-guard: v1 (#2867)
+# trusty-mpm-push-guard: v2 (#2867)
 #
 # Why: issue #2867 — an agent worktree whose local branch was configured with
 #   `branch.<local>.merge = refs/heads/<someone-else's-PR-branch>` ran a bare
@@ -8,12 +8,37 @@
 #   `git push origin HEAD:other-branch` does it too), and the offending push
 #   may come from a process trusty-mpm never created. A pre-push hook is the
 #   only guard that fires regardless of WHO runs the push.
-# What: refuses any push whose destination branch name differs from the
-#   checked-out branch name. Non-branch destinations (tags, notes, refs/pull)
-#   are out of scope and pass through untouched. Set
-#   TM_ALLOW_CROSS_BRANCH_PUSH=1 to perform a deliberate cross-branch push.
-# Test: crates/trusty-mpm/tests/push_guard_hook.rs (real temp repos: refuses a
-#   cross-branch push, permits the same-name push, permits with the override).
+# What: refuses a push that would land the history of one branch on a remote
+#   branch of a DIFFERENT name. One rule, applied identically whether HEAD is
+#   attached or detached:
+#
+#     source branch name must equal destination branch name
+#
+#   where the source name is the pushed `refs/heads/<name>`, or — when the
+#   source is anonymous (`HEAD`, a raw sha) — the checked-out branch name.
+#   Three shapes are deliberately OUT OF SCOPE and pass through untouched:
+#     * non-branch destinations (tags, notes, refs/pull) — cannot clobber a
+#       PR's branch lineage;
+#     * branch DELETES (`git push origin --delete X`, `git push origin :X`) —
+#       they move no history onto a branch, no git config can make a bare
+#       push perform one, so they are never the #2867 accident; post-merge
+#       remote cleanup is routine and a client hook is the wrong place to
+#       police it (branch protection is the right one);
+#     * an anonymous source pushed from a DETACHED HEAD — there is no
+#       checked-out branch whose identity could be violated, git itself
+#       refuses a bare `git push` while detached ("fatal: You are not
+#       currently on a branch"), so every such push is an explicit operator
+#       refspec, and `git push origin <sha>:refs/heads/<name>` is the
+#       standard way to rescue work out of a detached worktree.
+#   Set TM_ALLOW_CROSS_BRANCH_PUSH=1 to perform a deliberate cross-branch push.
+# Test: crates/trusty-mpm/tests/push_guard_hook.rs (real temp repos and a real
+#   bare remote: refuses the cross-branch push attached AND detached, permits
+#   the same-name push in both states, permits deletes and tags, permits with
+#   the override).
+#
+# Note: git aborts the ENTIRE push when a pre-push hook exits non-zero, so a
+# multi-refspec push containing one refused ref sends none of them. That is
+# git's hook contract, not a choice this script makes.
 #
 # stdin lines are: <local ref> <local sha> <remote ref> <remote sha>
 # argv is: $1 = remote name, $2 = remote URL.
@@ -22,47 +47,60 @@ if [ "${TM_ALLOW_CROSS_BRANCH_PUSH:-}" = "1" ]; then
 	exit 0
 fi
 
-# Empty when HEAD is detached; the per-line fallback below handles that.
+# Empty when HEAD is detached.
 head_branch=$(git symbolic-ref --quiet --short HEAD 2>/dev/null) || head_branch=""
 
 refused=0
-while read -r local_ref _local_sha remote_ref _remote_sha; do
+refused_branch=""
+# `|| [ -n "$local_ref" ]` keeps a final line with no trailing newline from
+# being silently dropped: POSIX `read` returns non-zero at EOF even when it
+# read a partial line, and dropping it here would fail OPEN.
+while read -r local_ref _local_sha remote_ref _remote_sha || [ -n "$local_ref" ]; do
 	# Only branch destinations can clobber a PR lineage.
 	case "${remote_ref:-}" in
 	refs/heads/*) ;;
 	*) continue ;;
 	esac
-	remote_branch=${remote_ref#refs/heads/}
+	dest_branch=${remote_ref#refs/heads/}
 
-	# Source name: the local branch being pushed, or — for a delete
-	# (`git push origin :foo`, where local_ref is literally "(delete)") — the
-	# checked-out branch, since deleting a branch you are not on is the same
-	# cross-branch hazard.
 	case "${local_ref:-}" in
-	refs/heads/*) local_branch=${local_ref#refs/heads/} ;;
-	*) local_branch=$head_branch ;;
+	refs/heads/*)
+		# A named local branch: compare its name to the destination.
+		src_branch=${local_ref#refs/heads/}
+		;;
+	'(delete)' | '')
+		# Out of scope — see the header.
+		continue
+		;;
+	*)
+		# Anonymous source (`HEAD`, a raw sha). Attached, it IS the
+		# checked-out branch; detached, it has no name to defend.
+		if [ -z "$head_branch" ]; then
+			continue
+		fi
+		src_branch=$head_branch
+		;;
 	esac
 
-	# Detached HEAD has no checked-out branch name to compare against, so fall
-	# back to source-vs-destination equality only.
-	expected=${head_branch:-$local_branch}
-
-	if [ "$remote_branch" = "$expected" ] && [ "$remote_branch" = "$local_branch" ]; then
+	if [ "$src_branch" = "$dest_branch" ]; then
 		continue
 	fi
 
 	refused=1
+	refused_branch=$src_branch
 	echo "pre-push: REFUSED cross-branch push (trusty-mpm push guard, #2867)" >&2
 	echo "  checked-out branch : ${head_branch:-<detached HEAD>}" >&2
-	echo "  pushing            : ${local_ref:-<delete>}" >&2
+	echo "  pushing            : $local_ref" >&2
+	echo "  which is branch    : $src_branch" >&2
 	echo "  onto destination   : $remote_ref" >&2
 done
 
 if [ "$refused" -ne 0 ]; then
 	echo "" >&2
-	echo "A bare \`git push\` landing on a branch you do not own is how PR #2863's" >&2
-	echo "reviewed lineage was clobbered (#2867). Push to your own branch instead:" >&2
-	echo "  git push origin HEAD:refs/heads/${head_branch:-<your-branch>}" >&2
+	echo "A push landing on a branch of a different name is how PR #2863's" >&2
+	echo "reviewed lineage was clobbered (#2867). Push onto the matching" >&2
+	echo "branch instead:" >&2
+	echo "  git push origin refs/heads/$refused_branch:refs/heads/$refused_branch" >&2
 	echo "If this cross-branch push is deliberate, re-run it with:" >&2
 	echo "  TM_ALLOW_CROSS_BRANCH_PUSH=1 git push ..." >&2
 	exit 1
--- a/crates/trusty-mpm/src/assets/hooks/pre-push
+++ b/crates/trusty-mpm/src/assets/hooks/pre-push
@@ -1,0 +1,71 @@
+#!/bin/sh
+# trusty-mpm-push-guard: v1 (#2867)
+#
+# Why: issue #2867 — an agent worktree whose local branch was configured with
+#   `branch.<local>.merge = refs/heads/<someone-else's-PR-branch>` ran a bare
+#   `git push` and clobbered the reviewed lineage of a PR branch it did not
+#   own. Tracking config is not the only way to arm that gun (an explicit
+#   `git push origin HEAD:other-branch` does it too), and the offending push
+#   may come from a process trusty-mpm never created. A pre-push hook is the
+#   only guard that fires regardless of WHO runs the push.
+# What: refuses any push whose destination branch name differs from the
+#   checked-out branch name. Non-branch destinations (tags, notes, refs/pull)
+#   are out of scope and pass through untouched. Set
+#   TM_ALLOW_CROSS_BRANCH_PUSH=1 to perform a deliberate cross-branch push.
+# Test: crates/trusty-mpm/tests/push_guard_hook.rs (real temp repos: refuses a
+#   cross-branch push, permits the same-name push, permits with the override).
+#
+# stdin lines are: <local ref> <local sha> <remote ref> <remote sha>
+# argv is: $1 = remote name, $2 = remote URL.
+
+if [ "${TM_ALLOW_CROSS_BRANCH_PUSH:-}" = "1" ]; then
+	exit 0
+fi
+
+# Empty when HEAD is detached; the per-line fallback below handles that.
+head_branch=$(git symbolic-ref --quiet --short HEAD 2>/dev/null) || head_branch=""
+
+refused=0
+while read -r local_ref _local_sha remote_ref _remote_sha; do
+	# Only branch destinations can clobber a PR lineage.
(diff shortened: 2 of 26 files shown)


bobmatnyc:
## code-critic: **WARN** — PR #4090 (`fix/2867-push-safety-guard`)

Reviewer self-check: `trusty-review review_health` → `trusty_search.reachable = true`, `status: ok`, v0.10.1. Search-backed context available; this verdict is **not** provisional.

Verification performed locally (git 2.54.0, real temp repos + real bare origin):

```
cargo test -p trusty-mpm --test push_guard_hook

cargo test -p trusty-mpm --lib push_guard
  test result: ok. 7 passed; 0 failed; 0 ignored; 3621 filtered out

cargo test -p trusty-mpm --lib -- push_pin_detection provisioner_worktree_add_writes_no_foreign \
    ensure_base_checkout_installs_push_guard create_session_worktree_sets_pull
  test result: ok. 4 passed; 0 failed; 0 ignored; 3624 filtered out
```

Plus 18 hand-built push scenarios against a real bare remote with the hook installed. Zero CRITICAL findings, so this is not a BLOCK — but the HIGH findings below should land before the hook is retrofitted onto the shared 95-worktree base.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/assets/hooks/pre-push` | 26, 48, 50 | Detached HEAD refuses **every** explicit-refspec branch push, and the printed remedy is the command that was just refused | Skip the check when `head_branch` is empty and the source is not a branch ref |
| HIGH | `crates/trusty-mpm/src/core/push_guard.rs` | 121–137 | `if let Ok(existing) = read_to_string(...)` — a non-UTF-8 foreign hook returns `Err(InvalidData)` and falls through to the unconditional `fs::write`, **clobbering** it | Match on `ErrorKind::NotFound`; read bytes, not `String`; refuse on any other read error |
| MEDIUM | `.../assets/hooks/pre-push` | 41–44, 63–65 | `git push origin --delete <branch>` refused with a remedy that is nonsense for a delete | Branch the remediation text on `(delete)` |
| MEDIUM | `.../assets/hooks/pre-push` | 48, 50 | Name-preserving `git push origin <other-branch>` is REFUSED when attached, ALLOWED when detached — same op, opposite verdicts | Make the detached and attached paths agree |
| MEDIUM | `crates/trusty-mpm/src/core/push_guard.rs` | 135–137 | Non-atomic in-place `fs::write` of a script shared by ~95 worktrees | Write `pre-push.tmp.<pid>` + chmod + `fs::rename` |
| MEDIUM | `.../daemon/managed_routes/inproject.rs`, `.../provisioner/workspace.rs` | 317, 447 | Guard installs only on the **fresh-clone** path; the already-provisioned base this PR exists to protect gets nothing | Add a retrofit entry point (`tm doctor` / `tm hook install`) |
| MEDIUM | `crates/trusty-mpm/tests/push_guard_hook.rs` | — | No test for detached HEAD, tag pass-through, or the attached/detached asymmetry | Add the three cases |
| LOW | `.../daemon/managed_routes/inproject.rs` | ~512 (doc) | "provably confined to the current branch" overstates what `push.default` guarantees | Soften the claim |
| LOW | `.../assets/hooks/pre-push` | 29 | Final stdin line without a trailing newline is silently dropped (POSIX `read`) | Latent only; note or `|| [ -n "$local_ref" ]` |

---

### HIGH-1 — Detached HEAD refuses every branch push, and tells you to re-run the refused command

`crates/trusty-mpm/src/assets/hooks/pre-push:26,48,50`

```sh
head_branch=$(git symbolic-ref --quiet --short HEAD 2>/dev/null) || head_branch=""
...
expected=${head_branch:-$local_branch}
if [ "$remote_branch" = "$expected" ] && [ "$remote_branch" = "$local_branch" ]; then
```

When HEAD is detached, `local_ref` is `HEAD` (or a raw sha) rather than `refs/heads/*`, so `local_branch` falls back to the **empty** `head_branch`, `expected` is empty, and the equality can never hold. Both explicit-refspec forms are refused — including `git push origin <sha>:refs/heads/<name>`, the canonical way to push out of a detached worktree. Reproduced:

```
$ git checkout --detach HEAD && git push origin HEAD:refs/heads/detachwork
pre-push: REFUSED cross-branch push (trusty-mpm push guard, #2867)
  checked-out branch : <detached HEAD>
  pushing            : HEAD
  onto destination   : refs/heads/detachwork
...
  git push origin HEAD:refs/heads/<your-branch>     <-- the printed remedy
RC=1

$ git push origin cc9b044...:refs/heads/detachwork2
pre-push: REFUSED cross-branch push ...
RC=1
```

Two problems compound. First, **13 of the 95 worktrees on the base clone are currently in detached HEAD**, so retrofitting this hook strands all of them on any explicit-refspec push. Second, line 65 prints `git push origin HEAD:refs/heads/${head_branch:-<your-branch>}` — substituting the intended branch name yields *exactly* the command that was just refused, so an agent following the guidance loops. The bypass is printed too, so this is recoverable rather than fatal, which is why it is HIGH and not CRITICAL — but "the remedy is the refused command" is precisely the shape that burns an agent's whole turn.

There is no #2867 hazard here to protect against: with no checked-out branch there is no "branch I own" to be pushed off of, and the source is an anonymous commit. Suggested fix, inserted before the `expected=` line:

```sh
if [ -z "$head_branch" ]; then
	# Detached HEAD: there is no checked-out branch name to compare against.
	# A named local branch still gets the source-vs-destination check below;
	# a HEAD/<sha> source has no name to compare, so let it through.
	case "${local_ref:-}" in
	refs/heads/*) ;;
	*) continue ;;
	esac
fi
```

and make the trailing guidance print the bypass first whenever `head_branch` is empty, instead of the `HEAD:refs/heads/<your-branch>` line.

### HIGH-2 — A non-UTF-8 foreign hook is silently overwritten

`crates/trusty-mpm/src/core/push_guard.rs:121-137`

```rust
if let Ok(existing) = std::fs::read_to_string(&hook_path) {
    if !existing.contains(HOOK_MARKER) { return Ok(HookInstall::Refused(...)); }
    if existing == PRE_PUSH_HOOK { return Ok(HookInstall::AlreadyCurrent(hook_path)); }
}
std::fs::create_dir_all(&hooks_dir)...;
std::fs::write(&hook_path, PRE_PUSH_HOOK)...;
```

The code distinguishes only `Ok` from `Err`. Every `Err` variant — not just `NotFound` — falls straight through to the unconditional `fs::write`. A `pre-push` hook that is not valid UTF-8 (a compiled hook binary, or a shell hook carrying latin-1 bytes) yields `Err(InvalidData)` and gets destroyed. Confirmed:

```
$ printf '#!/bin/sh\n# lefthook \xff\xfe binary marker\nexit 0\n' > repo/.git/hooks/pre-push
$ # replicating the exact std call:
Err(InvalidData) -> falls through to unconditional fs::write => FOREIGN HOOK CLOBBERED
```

This directly contradicts the module doc ("Installation is refused (never forced) when a foreign `pre-push` hook … is already in place, so this never fights husky/lefthook") and the CHANGELOG's "Installation refuses rather than overwrites". `fs::write` also follows symlinks, so a `pre-push` symlinked into a hook manager's directory is written *through* to the manager's own file.

Mitigating context, and the reason this is HIGH and not CRITICAL: both current call sites run immediately after `git clone`, where `.git/hooks/` contains only `*.sample`, so the path is not reachable today. It becomes reachable the instant a retrofit entry point exists — which MEDIUM-4 says this PR needs.

```rust
match std::fs::symlink_metadata(&hook_path) {
    Ok(m) if m.file_type().is_symlink() => {
        return Ok(HookInstall::Refused(format!(
            "{} is a symlink; another hook manager owns it", hook_path.display()
        )));
    }
    _ => {}
}
match std::fs::read(&hook_path) {
    Ok(bytes) => {
        if !String::from_utf8_lossy(&bytes).contains(HOOK_MARKER) {
            return Ok(HookInstall::Refused(format!(
                "a non-trusty-mpm pre-push hook already exists at {}", hook_path.display()
            )));
        }
        if bytes == PRE_PUSH_HOOK.as_bytes() {
            return Ok(HookInstall::AlreadyCurrent(hook_path));
        }
    }
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
    Err(e) => {
        return Ok(HookInstall::Refused(format!(
            "cannot read existing {}: {e}", hook_path.display()
        )));
    }
}
```

### MEDIUM-1 — `git push origin --delete <branch>` refused with unusable guidance

`crates/trusty-mpm/src/assets/hooks/pre-push:41-44,63-65`

```
$ git push origin --delete victim          # while on feat/slashy2
pre-push: REFUSED cross-branch push (trusty-mpm push guard, #2867)
  pushing            : (delete)
  onto destination   : refs/heads/victim
...
  git push origin HEAD:refs/heads/feat/slashy2      <-- would push, not delete
RC=1
```

The refusal itself is deliberate and tested (`tests/push_guard_hook.rs:141-150`), and defensible — deleting a branch you are not on is a real hazard. The guidance is not: it recommends a *push* to a *different branch*, which does not accomplish the delete. Post-merge remote-branch cleanup is routine here (`gh pr merge --delete-branch` goes via API and is unaffected, but the raw git form is common). Emit a delete-specific remedy when `local_ref` is `(delete)`, naming only the bypass.

### MEDIUM-2 — Attached is stricter than detached for the same operation

`crates/trusty-mpm/src/assets/hooks/pre-push:48,50`

```
# ATTACHED on main, source name == destination name:
$ git push origin sidebranch
  pushing            : refs/heads/sidebranch
  onto destination   : refs/heads/sidebranch      <-- names identical
REFUSED (RC=1)

# DETACHED, identical operation:
$ git push origin mine
ALLOWED (RC=0)
```

`expected=${head_branch:-$local_branch}` collapses to `local_branch` when detached, so the `= expected` clause becomes a tautology and only the source==destination check survives. Attached, the extra `= head_branch` clause refuses a strictly name-preserving push. Whichever behaviour is intended, the two paths should agree — as written the guard is *looser* in the less-common, less-understood state, which is backwards.

### MEDIUM-3 — Non-atomic write of a fleet-shared script

`crates/trusty-mpm/src/core/push_guard.rs:135-137`

`fs::write` truncates and rewrites in place. A partially written file leaves a shell script that fails to parse, and `sh` exits non-zero — which the hook contract reads as **refuse**, blocking every push from every worktree of the base until someone notices. POSIX shells also read scripts incrementally, so rewriting the file while another worktree is exec'ing it can make that shell execute garbage. The `AlreadyCurrent` short-circuit at :128 keeps the steady state write-free, so the window is narrow today — but it is exactly the fleet-wide-outage direction. Write to `pre-push.tmp.<pid>`, `chmod 0o755`, then `fs::rename` (atomic within a filesystem, and the rename also fixes the "hook executes while being rewritten" hazard).

### MEDIUM-4 — Nothing installs the guard on an already-provisioned base

`inproject.rs:317`, `provisioner/workspace.rs:447` are the only two call sites, both gated on a **fresh** clone. The comment in `inproject.rs` argues this is deliberate ("changing hook behaviour underneath [live sessions] is an operator decision"), which is a reasonable safety stance — but the practical consequence is that the base clone from the #2867 incident, and the ~95-worktree base this is written to protect, receive zero protection from merging this PR. There is no CLI, no doctor check, and no documented manual step. `grep -rn "install_and_log\|install_pre_push_guard" crates --include=*.rs` returns only those two sites plus tests. Add an explicit operator-invoked retrofit (`tm doctor --fix` or `tm hook install`) calling `install_pre_push_guard`, or at minimum document the manual `cp` in the CHANGELOG so the fix is actionable.

### MEDIUM-5 — Coverage gaps on documented behaviours

`crates/trusty-mpm/tests/push_guard_hook.rs` covers the cross-branch refusal, the same-name push, the bypass, and the ad-hoc-worktree shape — genuinely good end-to-end tests against a real remote. Missing: (a) detached HEAD (HIGH-1, and the state 13 of 95 worktrees are in right now), (b) tag pass-through, which the hook header explicitly promises ("Non-branch destinations (tags, notes, refs/pull) are out of scope and pass through untouched") and nothing asserts, (c) the attached/detached asymmetry in MEDIUM-2. Each is a few lines on the existing fixture.

### LOW-1 — "provably confined" overstates the pin

`inproject.rs` doc on `push_is_pinned_to_current`. The read-back itself is **correct** — I verified `git config --get push.default` resolves `config.worktree` exactly as claimed (returns `current` in a pinned worktree, unset in the base and in an unpinned sibling; correctly reports unset when the `--worktree` write fails because `extensions.worktreeConfig` is off). Two caveats to the wording: a configured `remote.<name>.push` refspec bypasses `push.default` entirely, and the gate is equally satisfied by a **global** `push.default=current` that trusty-mpm does not own and the user can change later, re-arming an already-written foreign upstream. The hook is the backstop for both, so this is a doc-accuracy point only.

### LOW-2 — Unterminated final stdin line is dropped

`pre-push:29`. `printf 'refs/heads/a 111 refs/heads/b 222' | sh pre-push origin url` → `RC=0`: POSIX `read` returns non-zero at EOF even after reading a partial line, so the loop body never runs for it. Fail-**open**, and git's send-pack always newline-terminates, so this is latent. Noting it only because the fail-open direction is invisible.

---

## Required changes before this is retrofitted fleet-wide

1. Fix HIGH-1 — stop refusing detached-HEAD explicit-refspec pushes, and never print the refused command as the remedy.
2. Fix HIGH-2 — `read_to_string` `Err` must not fall through to the overwrite; refuse on non-`NotFound` errors and on symlinks.
3. Fix MEDIUM-1 — delete-specific remediation text.
4. Reconcile MEDIUM-2 — attached and detached must agree.
5. Make the write atomic (MEDIUM-3).
6. Provide a retrofit path (MEDIUM-4) — without it the PR does not protect the base it was written for.
7. Add the three missing tests (MEDIUM-5).

## Checked and explicitly NOT flagged

Verified allowed, all `RC=0`, hook installed: same-branch push; first push of a new branch with `-u`; `--force-with-lease` on the current branch (the rebase form the version-control agents use); annotated tag push and `--tags`; branch names containing slashes; `HEAD:refs/heads/<same-name>`; `git push origin HEAD` while attached; `git push --all`.

Verified the escapes both work: `TM_ALLOW_CROSS_BRANCH_PUSH=1` releases every refusal including the delete form (checked at line 21, before any other logic — there is no path that refuses and then ignores the bypass), and `git push --no-verify` bypasses cleanly (the hook makes no attempt to defeat it).

Verified the script is safe shell: `sh -n` clean; POSIX-only constructs; every expansion quoted; `$remote_ref`/`$local_ref` are only ever `case` *subjects* and `[ = ]` operands, never patterns or `eval` input, and git's own `check-ref-format` forbids spaces and glob metacharacters in ref names, so there is no word-splitting or glob-injection surface. Empty stdin exits 0 without hanging. The `while read` loop runs in the main shell (stdin is the hook's own fd, not a pipe), so `refused=1` survives the loop.

A multi-refspec push where one ref is allowed and one refused aborts the **entire** push, including the allowed ref — confirmed (`git push origin HEAD:refs/heads/other HEAD:refs/heads/victim` left `refs/heads/other` uncreated). That is git's pre-push contract, not a defect, but it is undocumented in the hook header and worth a comment.

`core.hooksPath` detection is correct and refuses rather than writes; the older-revision upgrade path via `HOOK_MARKER` works; `install_and_log` never fails a clone. The `inproject.rs` reorder (pin first, read back the effective config, write no upstream when unconfirmed) is sound and the fail-safe direction is right — losing `git pull` tracking is much cheaper than an armed foreign upstream.

bobmatnyc:
## Fix round — all 9 findings addressed (`2c763fdc`)

Thanks for the review; the reproductions were precise enough to act on directly. Every finding is fixed, nothing deferred, nothing disputed on the facts. The one place I diverged from the suggested fix is MEDIUM-1, and I explain why below.

The context that shaped the round: the repo owner has authorised retrofitting this hook onto the existing `.base` clone shared by ~95 worktrees. So HIGH-2 and MEDIUM-4 are no longer latent, and I optimised for "safe to point at a live fleet" over "minimal diff".

---

### HIGH-1 — detached HEAD: fixed, and the rule is now symmetric

**Confirmed as reported.** Rather than special-casing detached HEAD, I collapsed the whole guard to **one rule applied identically in both states**:

> the source branch name must equal the destination branch name

Source name resolution:

| source `local_ref` | attached | detached |
|---|---|---|
| `refs/heads/X` | `X` | `X` |
| `HEAD` / raw sha | checked-out branch | **exempt** |
| `(delete)` | exempt | exempt |

**Why detached-anonymous is exempt, not merely "less strict":** git itself refuses a bare `git push` from a detached HEAD (`fatal: You are not currently on a branch`) — the hook never even runs. I verified this on 2.54.0. So there is no implicit, config-armed push path from that state at all; every push from a detached worktree is an explicit refspec the operator typed. There is no branch identity to defend and no accident to prevent, which is exactly why `git push origin <sha>:refs/heads/<name>` is the standard rescue form. Blanket-refusing it bought nothing and cost 13 worktrees.

It is **not** a blanket exemption though — a *named* branch pushed onto a different name is still the #2867 shape and is still refused while detached. That is asserted directly:

```
// detached_head_permits_explicit_refspec_but_still_refuses_cross_branch
git push origin HEAD:refs/heads/detachwork        -> ALLOWED
git push origin <sha>:refs/heads/detachwork2      -> ALLOWED
git push origin --force mine:refs/heads/victim    -> REFUSED, victim ref byte-identical
```

**The looping-remedy half.** The refusal now records the refused *source* branch and prints:

```
git push origin refs/heads/$refused_branch:refs/heads/$refused_branch
```

which the guard permits by construction (source name == destination name), and `$refused_branch` is provably non-empty on every path that sets `refused=1` — so the `<your-branch>` placeholder that made the old message unusable is gone entirely. I did not want that to rest on my reading of the control flow, so the test **executes the printed remedy and asserts it succeeds**:

```rust
let remedy = stderr.lines().find(|l| l.trim_start().starts_with("git push origin"))…;
let remedy_args: Vec<&str> = remedy.split_whitespace().skip(1).collect();
let (remedy_ok, _, remedy_err) = git(&fx.clone, &remedy_args);
assert!(remedy_ok, "the printed remedy `{remedy}` must actually succeed, not be the refused command");
```

That assertion fails loudly if anyone ever reintroduces a remedy the guard rejects.

### HIGH-2 — error-kind fix confirmed

**Confirmed, and fixed exactly as specified: only `NotFound` means "absent"; every other error refuses.** Implemented as a shared read-only probe, `inspect_pre_push_guard -> GuardState`, so the installer and the new doctor check make *one* ownership judgement rather than two that can drift:

```rust
match std::fs::symlink_metadata(&hook_path) {
    Ok(meta) if meta.file_type().is_symlink() => return GuardState::Foreign(…"is a symlink"…),
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => return GuardState::Missing(hook_path),
    Err(e) => return GuardState::Foreign(format!("cannot stat {}: {e}", …)),
    Ok(_) => {}
}
match std::fs::read(&hook_path) {                       // BYTES, never a String
    Ok(bytes) if !contains_marker(&bytes) => GuardState::Foreign(…),
    Ok(bytes) if bytes == PRE_PUSH_HOOK.as_bytes() => GuardState::Current(hook_path),
    Ok(_) => GuardState::Outdated(hook_path),
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => GuardState::Missing(hook_path),
    Err(e) => GuardState::Foreign(format!("cannot read existing {} ({e}); refusing to \
                                           overwrite a hook that cannot be inspected", …)),
}
```

Three tests, all with real files:

- `refuses_non_utf8_foreign_hook` — a genuinely invalid-UTF-8 hook (built at runtime, not as a `b"…"` literal, because the compiler's `invalid_from_utf8` lint sees through a literal and would turn the "this fixture really is not UTF-8" assertion into a deny-level warning). Asserts the file is **byte-identical** after the refusal.
- `refuses_symlinked_hook` — asserts the symlink *target* is never written through.
- `refuses_unreadable_hook` — mode `0o000`, asserts refusal + file intact (self-skips if mode 0 does not bind for the uid, rather than asserting a lie).

The module doc and CHANGELOG were both rewritten to state the new semantics, since you were right that they previously claimed something the code did not do.

### MEDIUM-4 — retrofit path built (this was the real one)

Agreed this was effectively HIGH: merging the previous revision would have protected zero of the ~95 worktrees. Two halves, both wired into existing surfaces:

**Discovery — `tm doctor` check `push_guard`** (`daemon/doctor_push_guard.rs`, check #22). Warn-only, consistent with the repo's doctor convention that checks report and never mutate. Warns on missing, on an older revision, and on a foreign/unreadable slot — a foreign hook means *unprotected*, so it deliberately never reads as `Ok`. The message names the exact command.

**Action — `tm repair push-guard [--path <dir>] [--dry-run]`** (`bin/tm/commands/push_guard.rs`). Daemon-free and fully local, so it works when the daemon is down. Properties, per your requirements:

- **idempotent** — `AlreadyCurrent` writes nothing; re-running is a no-op.
- **honours HIGH-2 refusal semantics** — same shared probe, so a foreign hook is never clobbered, and a refusal exits **non-zero** (a refusal means still-unprotected and must not read as success to a script).
- **reports what it did and why** — repository, resolved hooks dir, worktree count (the blast radius: "one file that N worktrees will execute"), then `INSTALLED` / `UP TO DATE` / `REFUSED <reason>`. `--dry-run` prints `WOULD INSTALL / UPGRADE / SKIP / REFUSE` and touches nothing — which matters when the target is a clone with live agent sessions.
- **discoverable** — extends the existing `tm repair` group (documented as extensible for exactly this), appears in the generated `references/cli.md`, and `tm doctor` names it.

On your note about mechanism: confirmed, **no `core.hooksPath` write**. It writes into `$GIT_COMMON_DIR/hooks`, so one invocation from *any* worktree covers the whole clone. That is pinned by a test that provisions a linked worktree, runs the retrofit **from the linked worktree**, and asserts the base checkout reports `GuardState::Current`.

### MEDIUM-1 — deletes now pass through (divergence from the suggested fix)

You suggested branching the remediation text on `(delete)`. **I went further and stopped refusing deletes at all.** Reasoning:

1. A delete moves **no history onto a branch**. It is a different hazard class (destructive remote cleanup) from the one this hook exists for (a mis-aimed push landing your work on someone else's ref).
2. There is **no accidental path**. No git config can make a bare `git push` perform a delete — unlike the #2867 shape, which was armed entirely by `branch.<n>.merge` + `push.default`. Every delete is explicit.
3. The refusal caught almost only legitimate traffic: you normally delete a branch you are *not* standing on, so post-merge cleanup was refused essentially 100% of the time.
4. Client-side hooks are the wrong layer for "do not delete that branch" — branch protection is.

So the honest fix was to narrow the guard's scope to what it does well and document it, rather than keep a refusal and improve its wording. This does **reverse a tested, PR-body-advertised behaviour**, so flagging it explicitly: `push_guard_hook.rs`'s delete assertion was inverted (now `deletes_and_tags_pass_through`), and the CHANGELOG states the exclusion and the rationale. If you would rather keep the refusal with corrected text, say so and I will flip it — this is a scope judgement, not a correctness one.

### MEDIUM-2 — asymmetry resolved (in the permissive direction)

Resolved by the single rule above: `git push origin <other-branch>` is now permitted **in both states**. It is name-preserving, so it structurally cannot land one branch's work on another branch's ref — the thing #2867 was about. The old attached-only refusal also blocked things like pushing `main` from a feature branch, which is friction with no safety return.

`name_preserving_push_of_another_branch_agrees_attached_and_detached` performs the identical operation in both states and asserts both succeed — so the two paths cannot silently diverge again.

You were right that being *looser* in the less-understood state was backwards; it is now identical in both.

### MEDIUM-3 — atomic write

Done, as suggested: `pre-push.tmp.<pid>` -> `chmod 0o755` -> `fs::rename`, with the temp file cleaned up on any failure path. `atomic_write_leaves_no_temp_file_behind` asserts no `pre-push.tmp.*` survives. Your framing was the deciding factor — a truncated script exits non-zero and git's contract reads that as *refuse*, so a partial write is a fleet-wide push outage, and `rename` also fixes the "executing while being rewritten" hazard for shells reading incrementally.

### MEDIUM-5 — all three added, plus five more

Added: detached HEAD (both the permitted and still-refused directions), tag pass-through, the attached/detached agreement — plus the printed remedy actually executing, the non-UTF-8 hook, the symlinked hook, the unreadable hook, temp-file cleanup, and the probe's `Missing -> Current -> Outdated` transitions. `tests/push_guard_hook.rs` is now 6 real-remote tests; `core::push_guard` unit tests are 11; the doctor check has 6 and the CLI command 5, plus 3 parse tests.

### LOW-1 — doc softened

Renamed to "Does the effective `push.default` say a bare `git push` targets the current branch?" and added an explicit caveat naming **both** limits you found — a `remote.<name>.push` refspec bypasses `push.default` entirely, and a *global* `push.default = current` that trusty-mpm does not own satisfies the gate and can be changed later, re-arming an already-written upstream. The doc now states the hook is the backstop for both, and that this check only decides whether writing a foreign upstream is defensible in the first place.

### LOW-2 — fixed rather than noted

It was fail-**open**, so I fixed it: `while read -r … || [ -n "$local_ref" ]`. Verified both directions:

```
printf 'refs/heads/a 111 refs/heads/b 222' | plain read     -> (no output — line dropped)
printf 'refs/heads/a 111 refs/heads/b 222' | with || guard  -> GOT [refs/heads/a] [refs/heads/b]
```

Also took your undocumented-behaviour note: the hook header now records that git aborts the entire multi-refspec push when the hook exits non-zero, and that this is git's contract rather than a choice the script makes.

---

## Quality gate — raw output

```
$ cargo check -p trusty-mpm --all-targets
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 36.83s

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 55.83s

$ cargo fmt --all -- --check
FMT_EXIT=0

$ cargo test -p trusty-mpm            # FULL surface, not --lib
test result: ok. 3633 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
test result: ok. 1247 passed; 0 failed; 1 ignored
test result: ok. 1247 passed; 0 failed; 1 ignored
… 38 test binaries, 0 failures …

$ cargo test -p trusty-mpm --test push_guard_hook
running 6 tests
test refuses_cross_branch_push_and_permits_same_name_push ... ok
test override_env_var_permits_cross_branch_push ... ok
test detached_head_permits_explicit_refspec_but_still_refuses_cross_branch ... ok
test name_preserving_push_of_another_branch_agrees_attached_and_detached ... ok
test deletes_and_tags_pass_through ... ok
test adhoc_agent_created_worktree_inherits_the_guard ... ok

$ bash scripts/check_sld.sh
sld-lint: scanned 47 spec doc(s) + 2892 code file(s); 0 error(s), 0 warning(s)

$ sh -n crates/trusty-mpm/src/assets/hooks/pre-push
sh -n: OK
$ printf '' | sh …/pre-push origin file:///tmp/x   # empty stdin must not hang
empty_stdin_rc=0
```

Nothing was `#[ignore]`d, cfg-gated, `--exclude`d, or deleted to get green. The one test whose assertion changed is the delete case, inverted deliberately under MEDIUM-1 and called out above.

Doctor check count went 21 -> 22; the four hardcoded assertions, the `DOCTOR_CHECKS` generator array, and the two generated capability docs were all updated and the drift check passes.

## Nothing disputed

Every finding reproduced as described. The only judgement call is MEDIUM-1's scope narrowing, flagged above rather than buried.

bobmatnyc:
## Re-review (round 2) — head `2c763fdc`

`review_health`: `status=ok`, `inference=ok`, `trusty_search reachable=true`, `trusty_analyze reachable=true`, `dry_run=false`, reviewer `us.anthropic.claude-sonnet-4-6` v0.10.1. Verdict is **definitive, not provisional**.

CI independently confirmed green — 20/20 checks pass, including `Test` (16m24s), `Clippy` (5m20s), `MSRV 1.91`.

Everything below was verified by **executing the real hook asset against real git repos and a real bare remote** (git 2.54.0), not by reading.

---

## Verdict: WARN

**Mergeable as-is: yes.** One HIGH, no CRITICAL. The HIGH is a residual narrower than the hole being closed, and it is not a merge-blocker for a change whose false-refuse blast radius across ~100 worktrees is the dominant risk. It does need a fast follow, and one of its fixes is a two-line change I verified end-to-end below.

**MEDIUM-1 divergence (deletes pass through): ACCEPT the behaviour, FIX the stated rationale.** Detail in M-2.

---

## False-refuse sweep — all clear

Every required form was executed against a real remote with the real hook installed. **Zero false refusals** in the legitimate set:

| Form | Result |
|---|---|
| same-branch push (`mine` → `mine`) | ALLOW |
| first push with `-u` | ALLOW |
| `--force-with-lease` same-branch (our rebase form) | ALLOW |
| annotated tag push | ALLOW |
| `git push origin --tags` | ALLOW |
| slash-containing branch (`feat/slashy`) | ALLOW |
| `HEAD:refs/heads/<same-name>` | ALLOW |
| `git push --all` | ALLOW |
| detached `<sha>:refs/heads/rescue` (today's rescue form) | ALLOW |
| `git push origin --delete victim` | ALLOW |
| `git push origin :victim` | ALLOW |
| `TM_ALLOW_CROSS_BRANCH_PUSH=1` (every refusal shape) | ALLOW |
| `--no-verify` | ALLOW (escapes, as expected) |

And the refusal set holds, including the author's specific claim:

| Form | Result |
|---|---|
| attached `mine:refs/heads/victim` | REFUSE |
| attached `HEAD:refs/heads/victim` | REFUSE |
| attached `--force mine:refs/heads/victim` | REFUSE |
| **detached `mine:refs/heads/victim`** (author's assertion) | **REFUSE — claim verified** |
| multi-refspec mixing exempt + non-exempt | REFUSE (whole push, per git's contract) |

The attested #2867 mechanism — attached branch, `branch.<local>.merge` armed, bare `git push` — is **completely closed**. Confirmed including through an ad-hoc `git worktree add` worktree inheriting the guard via `$GIT_COMMON_DIR/hooks`.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/assets/hooks/pre-push` | 27-32, 78-80 | Detached-HEAD exemption **is** reachable by a config-armed bare push; the justifying comment is factually false | Narrow the exemption to ref *creates* using the `remote_sha` zeros signal |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/push_guard.rs` | 111-122 | `tm repair push-guard --path <bare base clone>` fails on the primary retrofit target | Fall back to `--git-common-dir` when `--show-toplevel` fails |
| MEDIUM | `crates/trusty-mpm/src/assets/hooks/pre-push` | 23-24 | "no git config can make a bare push perform one" — false; a config-armed bare push deleted a foreign branch | Correct the sentence; keep the pass-through |
| MEDIUM | `crates/trusty-mpm/src/assets/hooks/pre-push` | 75-82, 103 | Attached raw-sha rescue push refused while the *identical* detached command is allowed; printed remedy is a non-sequitur | Same `remote_sha` fix resolves this |
| LOW | `generate/doctor.rs:33`, `client/executor/tests.rs:213`, `daemon/api_tests.rs:1116`, `daemon/doctor_tests.rs:131` | — | Doctor check count `22` hardcoded in 4 places | Derive from `DOCTOR_CHECKS.len()` |

---

### HIGH-1 — the detached-HEAD exemption is reachable by a bare, un-typed push

`crates/trusty-mpm/src/assets/hooks/pre-push:27-32` asserts:

> `git itself refuses a bare `git push` while detached ("fatal: You are not currently on a branch"), so every such push is an explicit operator refspec`

The first half is true; **the conclusion is false.** That fatal comes from `push.default` resolution, which git never consults when `remote.<name>.push` supplies an explicit refspec.

Verified end-to-end with the real hook installed, a real bare remote, and **no refspec typed on the command line**:

```
$ git config remote.origin.push '+HEAD:refs/heads/victim'
$ git checkout --detach HEAD && git commit -m "unrelated work"
$ git push                      # <-- bare. no refspec. guard fully installed.
To ../remote.git
   43a4c2b..ae8f65e  HEAD -> victim
victim tip BEFORE: 43a4c2b8...
victim tip AFTER : ae8f65ea...
```

A foreign branch's lineage was force-clobbered by a bare `git push` **through the installed guard** — the exact #2867 damage shape, reached without anyone typing a refspec. The hook receives `local_ref=HEAD`, hits the anonymous arm at line 75, sees `head_branch` empty, and `continue`s at line 79.

Calibration, honestly: this needs `remote.origin.push` set with an anonymous LHS, which is **not** the attested #2867 arming mechanism and is not known to be set anywhere in this fleet. It is HIGH, not CRITICAL — the guard is defence-in-depth (`--no-verify` and the env var bypass it by design), and the PR still closes the real incident path completely. But the exemption is currently justified by a statement that is provably untrue, and that comment will be load-bearing for whoever touches this next.

**Concrete fix, verified available.** The hook already receives a create-vs-update signal it currently discards as `_remote_sha`:

```
CREATE new branch : local_ref=[HEAD] remote_ref=[refs/heads/brand-new] remote_sha=[0000000000000000000000000000000000000000]
UPDATE existing   : local_ref=[HEAD] remote_ref=[refs/heads/victim]    remote_sha=[690024f0341ed9ac0654282bc1eb72a01419ccef]
```

So in the anonymous arm, exempt only when the destination is being **created**:

```sh
*)
        # Anonymous source (`HEAD`, a raw sha).
        if [ -z "$head_branch" ]; then
                # Detached: no name to defend. A CREATE cannot clobber
                # anything (this is the rescue form); an UPDATE of an
                # existing branch is the #2867 shape and is not exempt.
                case "$_remote_sha" in
                *[!0]*) ;;      # destination exists -> fall through, refuse
                *) continue ;;  # all-zeros = create -> exempt
                esac
                src_branch="<detached HEAD>"
        else
                src_branch=$head_branch
        fi
        ;;
```

Checked against the whole matrix: detached `<sha>:refs/heads/rescue` stays ALLOWED (rescue form preserved), detached `+HEAD:refs/heads/victim` becomes REFUSED (hole closed), and the attested attached/`branch.*.merge` path is untouched because its `local_ref` is a *named* `refs/heads/…` that never reaches this arm.

### MEDIUM-1 — `tm repair push-guard` cannot be pointed at the base clone

`resolve_target` (`commands/push_guard.rs:111-122`) resolves via `rev-parse --show-toplevel`, which **fatals on a bare repository**. This repo's own `.base` — the headline retrofit target, named in the module docs — is bare:

```
$ git -C .../trusty-tools/.base rev-parse --is-bare-repository
true
$ git -C .../trusty-tools/.base rev-parse --show-toplevel
fatal: this operation must be run in a work tree
```

So the natural operator move, `tm repair push-guard --path .../.base`, exits non-zero with "is not inside a git working tree" — on precisely the clone the whole retrofit path exists to protect. It works from any *worktree*, so the capability is there; the entry point just rejects the most obvious argument. Given this command is the discovery-to-action path for ~100 worktrees, that misfire matters.

`effective_hooks_dir` already handles bare fine (`--git-common-dir` returns the repo). Fix: on `--show-toplevel` failure, retry `rev-parse --path-format=absolute --git-common-dir` and use that.

### MEDIUM-2 — the delete rationale contains the same false premise (behaviour still correct)

`pre-push:23-24` justifies the MEDIUM-1 divergence with "no git config can make a bare push perform one". Verified false by the same mechanism:

```
$ git config remote.origin.push ':refs/heads/victim'
$ git push
To ../remote.git
 - [deleted]         victim
```

**I still endorse the divergence.** A client hook is the wrong place to police deletes, branch protection is the right one, deleting a branch you are not standing on is routine post-merge cleanup, and a GitHub branch delete is recoverable (`refs/pull/N/head` survives; the PR offers "Restore branch") where a force-clobber of lineage is not. Refusing deletes would have produced steady false refusals across the fleet for a shape that is both rarer and less destructive. Keep the pass-through — but the second clause of that sentence must be corrected, for the same reason as HIGH-1: it is a false statement future maintainers will build on.

### MEDIUM-3 — attached raw-sha rescue is refused, and its remedy is a non-sequitur

The identical command flips on HEAD state:

```
detached: git push origin <sha>:refs/heads/rescue   -> ALLOW
attached: git push origin <sha>:refs/heads/rescue   -> REFUSE
    which is branch    : mine        <- but a raw sha was pushed, not `mine`
```

Making a backup branch from a specific commit while standing on a branch is legitimate and gets refused. Worse, the printed remedy is built from the *source* branch, so it prints `git push origin refs/heads/mine:refs/heads/mine` — which **succeeds** while doing something entirely unrelated to what the operator asked for.

Note this interacts with the round-1 HIGH-2 remedy test: I confirmed `tests/push_guard_hook.rs:147-159` genuinely extracts the remedy from stderr, splits it, and executes it (`git(&fx.clone, &remedy_args)`) rather than a hardcoded twin — good, and it would fail if the remedy regressed to the refused command, since the guard would refuse it and `remedy_ok` would be false. But it asserts the remedy *succeeds*, not that it accomplishes the intent, so it passes cleanly on this shape. The HIGH-1 `remote_sha` fix resolves this case too (a create becomes allowed).

### LOW-1 — the doctor-count duplication is a real but self-revealing trap

`22` is hardcoded at `generate/doctor.rs:33` (array size), `client/executor/tests.rs:213`, `daemon/api_tests.rs:1116`, `daemon/doctor_tests.rs:131`, plus the second name list in `api_tests.rs`. Worth reporting, but **LOW, not MEDIUM**: the failure mode is a loud red test, not a silent bug — which is exactly how the author found the second list. Fix: assert against `DOCTOR_CHECKS.len()` instead of a literal, and derive the name list from the same array.

---

## Verified clean — items I tried to break and could not

- **HIGH-2 (foreign-hook clobber) is properly fixed.** `inspect_pre_push_guard` (`core/push_guard.rs:142-177`) reads bytes not `String`, treats only `ErrorKind::NotFound` as absent, and returns `Foreign` for every other stat/read error. Symlinks are caught via `symlink_metadata` *before* any read. Arm ordering is right: no-marker → `Foreign` precedes the byte-equality check.
- **The shared-probe claim holds and cannot diverge.** `install_pre_push_guard:195-199` and `check_push_guard:46` both consume `inspect_pre_push_guard`'s `GuardState`; there is no second ownership judgement anywhere. Genuinely one probe.
- **Nice property the author may not have noticed:** `fs::rename` does not follow symlinks, so even if the symlink check raced, the atomic write replaces the *link* rather than writing through to another manager's file. The TOCTOU here is benign.
- **`--dry-run` is genuinely side-effect-free.** `report_dry_run` calls only `inspect_pre_push_guard`; the early return at `push_guard.rs:48` precedes `install_pre_push_guard`, and `create_dir_all` lives only on the install path.
- **`--path` cannot be aimed at anything dangerous.** The write target is always `$GIT_COMMON_DIR/hooks/pre-push` of a resolved repo, foreign hooks are refused, and `core.hooksPath` is refused. Worst case is installing a marker-tagged guard into an unrelated repo.
- **Idempotency holds.** `Missing`→`Installed`, repeat→`AlreadyCurrent`, older revision→`Installed` (upgrade). Temp file is `pid`-scoped and cleaned up; `atomic_write_leaves_no_temp_file_behind` covers the shared-dir litter case.
- **A concern I dropped under the confidence filter:** I suspected multi-valued `core.hooksPath` would make `git config --get` exit non-zero, be read as unset, and produce a false all-clear. Not so — git 2.54 returns the last value with exit 0, matching the path git actually uses. The installer is correct here.

---

## Required Changes (fast follow, not merge-blocking)

1. Narrow the detached anonymous-source exemption to ref *creates* via the all-zeros `remote_sha`, closing HIGH-1 and MEDIUM-3 together. Add coverage for `remote.origin.push='+HEAD:refs/heads/victim'` + bare push from detached.
2. Correct `pre-push:27-32` — a config-armed bare push from detached HEAD does exist; the exemption is justified by the create-only rule, not by "git refuses it".
3. Correct `pre-push:23-24` — drop the "no git config can make a bare push perform one" clause. Keep the delete pass-through; rest it on branch protection and recoverability.
4. Make `resolve_target` fall back to `--git-common-dir` so `--path` accepts a bare base clone.
5. Replace the four hardcoded `22`s with `DOCTOR_CHECKS.len()`.

## Notes

The MEDIUM-1 divergence should be **accepted, not reverted** — the reasoning is sound even though one supporting clause is not. My two false-premise findings are the same defect twice: the hook's header states empirical claims about git that were not tested. Both conclusions survive; both justifications need correcting, because the next person to widen an exemption will reason from them.

bobmatnyc:
## Round 3 — HIGH closed, retrofit unblocked, both false claims corrected (`52a30208`)

You were right, and the way you were right matters: I asserted an empirical claim about git in a shipped comment and never tested it, then reasoned from it to widen an exemption. I reproduced both of your findings locally before touching anything.

**Your exploit, against the round-2 guard as shipped:**

```
=== EXPLOIT 1: remote.origin.push + detached bare push ===
   e2c8b59..8bab7cd  HEAD -> victim
RESULT: *** VICTIM CLOBBERED ***

=== EXPLOIT 2: remote.origin.push delete refspec + bare push ===
 - [deleted]         victim
RESULT: *** VICTIM DELETED by a bare push ***
```

Both confirmed. My "git refuses a bare push while detached, therefore every detached push is an explicit refspec" was false, and it was load-bearing.

---

### HIGH — exemption re-keyed on create-vs-update

Taken as specified: the signal was already in hand as `_remote_sha`, which I was discarding. It is now read and used.

```sh
# A CREATE cannot destroy a lineage that does not exist yet. Git marks
# one with an all-zero <remote sha>. Checked BEFORE the name comparison
# so it applies identically to a named and an anonymous source, and to
# an attached and a detached HEAD — no asymmetry to reason about.
case "${remote_sha:-}" in
'' | *[!0]*) ;;  # absent or non-zero: the destination branch exists
*) continue ;;   # all zeros: a create
esac
```

Placing it **before** the name comparison rather than inside the anonymous-source arm is what folds MEDIUM-3 in: attached `<sha>:refs/heads/backup` and the identical detached command now get the same verdict, because HEAD state no longer appears in the decision at all.

**The exploit is now refused, with a concrete remedy:**

```
=== EXPLOIT 1: remote.origin.push + detached bare push ===
  git push origin HEAD:refs/heads/rescue-b9594d77
If this cross-branch push is deliberate, re-run it with:
  TM_ALLOW_CROSS_BRANCH_PUSH=1 git push ...
RESULT: victim INTACT (refused)
```

The detached remedy is derived from the pushed sha, so it is never a placeholder. `remote_push_refspec_cannot_clobber_from_detached_head` asserts the extracted remedy contains no `<`/`>` **and then executes it**.

**22-case allow/refuse sweep against a real bare remote, all green:**

```
===== LEGITIMATE FORMS (must all be ALLOWED) =====
  ok   allow    first push of a new branch with -u
  ok   allow    same-branch push
  ok   allow    force-with-lease on current branch
  ok   allow    HEAD:refs/heads/<same-name>
  ok   allow    git push origin HEAD (attached)
  ok   allow    annotated tag push
  ok   allow    --tags
  ok   allow    slash-containing branch name
  ok   allow    --all
  ok   allow    name-preserving push of another branch
  ok   allow    delete a remote branch (out of scope)
  ok   allow    DETACHED rescue: HEAD:refs/heads/<new>
  ok   allow    DETACHED rescue: <sha>:refs/heads/<new>
  ok   allow    ATTACHED create of a differently-named new branch
===== BYPASSES (must be ALLOWED) =====
  ok   allow    TM_ALLOW_CROSS_BRANCH_PUSH=1
  ok   allow    --no-verify
===== HAZARDS (must all be REFUSED) =====
  ok   refuse   attached HEAD:existing-foreign-branch
  ok   refuse   attached named cross-branch onto existing
  ok   refuse   DETACHED onto EXISTING foreign branch
  ok   refuse   DETACHED sha onto EXISTING foreign branch
  ok   refuse   EXPLOIT: remote.origin.push + detached bare push
===== #2867 ORIGINAL SHAPE (armed ad-hoc worktree) =====
  ok   refuse   bare push from armed ad-hoc worktree; victim byte-identical
================================
PASS=22 FAIL=0
```

Both detached rescue forms preserved; the attested `branch.*.merge` path untouched; the new create-exemption does not leak into updates (asserted separately in `creating_a_new_branch_is_allowed_attached_and_detached`).

### HIGH-adjacent — bare-repo retrofit fixed

Confirmed: `git -C bare.git rev-parse --show-toplevel` → `fatal: this operation must be run in a work tree`. `resolve_target` now tries `--show-toplevel`, then falls back to `--git-common-dir` (valid bare or not), which is what `effective_hooks_dir` already consumed.

`retrofits_a_bare_repository` covers it, and I ran the real binary against a throwaway bare clone with two linked worktrees — the authorized retrofit's exact invocation shape:

```
$ git -C base.git rev-parse --is-bare-repository
true

$ tm repair push-guard --path …/base.git --dry-run
Hooks directory: …/base.git/hooks (shared by 3 worktree(s) of this clone)
WOULD INSTALL   …/base.git/hooks/pre-push (no hook present)
(dry run — nothing was written)
hook present after dry run: no

$ tm repair push-guard --path …/base.git
INSTALLED   cross-branch push guard → …/base.git/hooks/pre-push

$ tm repair push-guard --path …/base.git      # idempotent
UP TO DATE  guard already current at …/base.git/hooks/pre-push

# and it actually fires from a linked worktree of that bare base:
pre-push: REFUSED cross-branch push (trusty-mpm push guard, #2867)
RESULT: victim INTACT — guard fired from a worktree of the bare base
```

### Both false claims corrected

**1. Detached-HEAD justification.** The header now carries an explicit warning rather than a quietly-fixed sentence, because your point was that the next person will reason from it:

> **WARNING** — do NOT reason from "git refuses a bare push while detached". That `fatal: You are not currently on a branch` comes from `push.default` resolution ONLY, and git never consults `push.default` when a configured `remote.<name>.push` refspec supplies the destination. Proven against this very hook: […] clobbered `victim` with no refspec typed by anyone. […] This is exactly why the exemption above is keyed on CREATE-vs-UPDATE, which git states in the `<remote sha>` field, and never on HEAD state or on how the push was spelled.

**2. MEDIUM-1's supporting clause.** Behaviour kept, rationale rebuilt on your recoverability argument:

> branch DELETES […] **NOT** because they cannot happen by accident — they can, see the warning below — but because a deleted GitHub branch is RECOVERABLE (the PR retains its commits at `refs/pull/<N>/head`, and the branch can be re-pushed from any clone) whereas a force-clobbered lineage is not. […] the right layer to police deletion is branch protection on the server, not a hook the pusher can remove.

Corrected in all four places the claims had spread to: the hook header, `deletes_and_tags_pass_through`'s doc, the detached test's doc, and the CHANGELOG. I grepped for both phrasings to confirm no restatement survives anywhere.

### Fast-follow 5 (LOW) — folded in

Cheap enough to take. `DOCTOR_CHECKS` is now `&[(&str, &str)]` rather than a sized array; the two roster tests bind their name list and assert `report.checks.len() == expected.len()`; and `execute_doctor_against_test_daemon` now asserts the executor **round-trip** it actually owns (non-empty, carries `instructions`/`worktreees`/`push_guard`) instead of duplicating a count a fourth time. Adding a check is now a one-line edit per roster test. **Nothing skipped** — items 1–5 are all in.

**Settled ground left alone:** I did not touch `core.hooksPath` handling.

---

## Quality gate — raw output

```
$ cargo test -p trusty-mpm            # FULL surface, not --lib
test result: ok. 3633 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
test result: ok. 1248 passed; 0 failed; 1 ignored
test result: ok. 1248 passed; 0 failed; 1 ignored
… 38 test binaries, 0 failures …

$ cargo test -p trusty-mpm --test push_guard_hook
test creating_a_new_branch_is_allowed_attached_and_detached ... ok
test override_env_var_permits_cross_branch_push ... ok
test detached_head_permits_explicit_refspec_but_still_refuses_cross_branch ... ok
test name_preserving_push_of_another_branch_agrees_attached_and_detached ... ok
test deletes_and_tags_pass_through ... ok
test refuses_cross_branch_push_and_permits_same_name_push ... ok
test remote_push_refspec_cannot_clobber_from_detached_head ... ok
test adhoc_agent_created_worktree_inherits_the_guard ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test -p trusty-mpm --bin tm push_guard
test commands::push_guard::tests::retrofits_a_bare_repository ... ok
test commands::push_guard::tests::worktree_count_counts_linked_worktrees ... ok
test commands::push_guard::tests::installs_into_a_fresh_repo_then_is_idempotent ... ok
test commands::push_guard::tests::refusal_exits_nonzero_and_leaves_the_foreign_hook_intact ... ok
test commands::push_guard::tests::dry_run_writes_nothing ... ok
test commands::push_guard::tests::unresolvable_path_exits_nonzero ... ok
test tests_behavior_repair::cli_parses_repair_push_guard ... ok
test tests_behavior_repair::cli_parses_repair_push_guard_flags ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 1241 filtered out

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.33s

$ cargo fmt --all -- --check
FMT_EXIT=0

$ bash scripts/check_sld.sh
sld-lint: scanned 47 spec doc(s) + 2892 code file(s); 0 error(s), 0 warning(s)

$ sh -n crates/trusty-mpm/src/assets/hooks/pre-push
sh -n OK
```

Nothing `#[ignore]`d, cfg-gated, `--exclude`d, or deleted. No hook was installed into `.base` or any real worktree — every experiment ran in throwaway scratch repos, and no issues were filed.

bobmatnyc:
### Addendum — head is now `71bc1546`, green and MERGEABLE

The fix commit is unchanged (`52a30208`); the two commits after it are `main` merges, no code of mine.

While round-3 CI was running, `main` advanced twice (#4096, then #4067/#4127/#4091) and the PR went `CONFLICTING`. Worth recording the symptom, because it does not look like a conflict from the outside: **GitHub silently dispatches zero workflow runs for a PR it cannot compute a merge for.** `gh pr checks` reported `no checks reported on the branch` and the runs API returned `total_count: 0` for the pushed SHA, while other branches were dispatching normally — so it read like an Actions outage rather than a conflict.

Both conflicts were `crates/trusty-mpm/CHANGELOG.md` only, both purely additive `## [Unreleased]` collisions (the second had an empty "theirs" side). Resolved by keeping every entry from both sides — no entry dropped, no code file touched. I merged rather than rebased to avoid force-pushing a branch under review.

Re-verified the full gate on the merged tree, since `main` brought in changes to this same crate:

```
$ cargo test -p trusty-mpm            # FULL surface
test result: ok. 3722 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
test result: ok. 1256 passed; 0 failed; 1 ignored
test result: ok. 1256 passed; 0 failed; 1 ignored
… 39 test binaries, 0 failures …

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 58.57s

$ cargo fmt --all -- --check
FMT_EXIT=0

```

CI on `71bc1546`: all 24 checks pass (`Test` 15m2s, `Clippy` 5m44s, `Test: doc-comment pointer lint` 4m34s, MSRV, SLD, line-cap, capabilities drift — green). `mergeable=MERGEABLE`.

The test count moved 3633 → 3722 purely from `main`'s incoming tests; my own additions are unchanged.

Issue #3987, opened 2026-07-26
Base `assistant` template's `google.read` scope matches nothing — its 60 gworkspace tools are scope-denied, and read-only Google is currently inexpressible


bobmatnyc:
Surfaced while fixing #3938 (PR #3985). Filed separately and deliberately NOT fixed there: #3938's fix was a single overlay declaring the two scope families its own tools need, whereas this one changes the shared base template's security posture and is an owner/product call, not an engineering one. Widening the base unilaterally inside that PR would have been exactly the scope-widening hack the fix avoided.

## Symptom

The bundled base `assistant` template allowlists **60 Google Workspace tools by name** (`crates/trusty-agents/.trusty-agents/agents/assistant/agent.toml:108-175` — Gmail, Calendar, Drive, Docs, Sheets, Slides, Tasks, accounts) and **every one of them is scope-denied at dispatch** on the persona-chat path (GUI + REPL), where scope enforcement is active. Not a subset: all of them. The base persona effectively has *zero* gworkspace tools, and so does every `extends = "assistant"` overlay that does not declare its own Google scopes.

## Mechanism — `google.read` is a dead pattern

The base declares:

```toml
# crates/trusty-agents/.trusty-agents/agents/assistant/agent.toml:183
scopes = ["memory.read", "memory.write", "search.read", "google.read"]
```

with the rationale at `:179` — *"Google access defaults to READ-ONLY (`google.read`) on the base template per §5.5 — a generic, un-personalized assistant should not mutate a user's mailbox / calendar / drive by default."* The intent is sound. The pattern, however, matches **nothing**:

1. **No gworkspace tool ever advertises `google.read`.** `dotted_scope_for_google_scopes` (`crates/trusty-agents/src/tools/registry/google_scope.rs:146`) walks `FAMILY_TABLE` (`:69-131`) and emits `google.<family>.<access>` — `google.gmail.write`, `google.calendar.write`, `google.tasks.write`, `google.drive.write`, `google.docs.write`, `google.sheets.write`, `google.slides.write`, `google.accounts.read`. There is no `google.read` in the produced vocabulary at all.
2. **`google.read` is an exact (non-wildcard) pattern.** `ScopePattern::matches` (`crates/trusty-agents/src/tools/registry/scope.rs:69`) takes the no-wildcard branch and requires string equality, so it matches only a scope literally equal to `google.read`.
[...]

Issue #3933, opened 2026-07-25
Skill-wrapping runtime: tool→skill manifests, skill-name allow-listing, migration


bobmatnyc:
Epic: #3931. Spec: `docs/specs/agent-config-five-sections.md` §5 (`SPEC-AGENTCFG-04~draft`) + §9 (`-08~draft`), Phase 2.

Owner directive: *"Instead of tools, let's show skills… each tool should be wrapped in a skill."*

## The gap being closed

trusty-agents **already has** a substantial skill subsystem (`crates/trusty-agents/src/skills/`: `SkillEntry`, `SkillRegistry`, priority-ordered sources, `load_skill`/`list_skills`/`search_skills`, `[system_prompt].skills`). What it does not have is **any binding from a skill to the tools it describes**. A skill today is prompt text injected as a `# Skill: <name>` layer (`runtime/subagent_mode.rs:276-288`).

Canonical demonstration: `.trusty-agents/skills/web-search.md` documents the `web_search` tool, its Tavily/Brave backends and its API keys — while `web_search` is granted *independently* in `[tools].allow`. Delete the skill and the tool works unexplained; delete the tool from `allow` and the skill promises a capability the agent does not have.

### 1. Skill manifest gains three keys

```yaml
kind: action          # NEW — action | knowledge | system; `knowledge` routes into the Knowledge pane
tools:                # NEW — exact tool names this skill wraps
  - search_gmail_messages
  - compose_email
scopes: [google.gmail.*]   # NEW, optional — advisory in this phase
```

`tools` entries are **exact names, never globs.** Three glob dialects already exist in the tree (`[tools].allow` trailing-`*` only `helpers.rs:146`; listener `from` leading-or-trailing `wake.rs:62`; scope patterns trailing `.*` on a segment boundary `scope.rs:68`). No fourth dialect.

### 2. `[skills].allow` resolution — the key design commitment

```toml
[skills]
allow = ["gmail", "calendar", "semantic-search", "git"]
```

1. Resolve against the skill source registry (`skills/sources/mod.rs`, priority-descending).
2. Expand each skill to its `tools` list.
3. `effective_tools = union(expanded_skill_tools, [tools].allow)`.
4. Feed into the **existing** gate: `filter_persona_tool_names(...)` (`persona.rs:65`).

[...]

Issue #3890, opened 2026-07-25
trusty-agents GUI: store bindings are read-only — define PATCH contract for editing [[stores]] from the config panel


bobmatnyc:
## Problem

PR #3878 landed live OKG store bindings: `agent.toml` gains `[[stores]]`, the GUI's OKG Stores pane fetches `GET /api/agents/:name/stores`, and each binding reports live CONNECTED / NOT CONNECTED status with chunk counts and readiness.

That surface is **read-only**. A user who sees "NOT CONNECTED — no such index" in the config panel has no way to correct it from the GUI; they must hand-edit `agent.toml` and reload. The pane shows the problem but not the fix.

## Proposal

Define and implement a PATCH contract for editing an agent's `[[stores]]` from the config panel.

Design questions to settle first:

- **Route shape.** `PATCH /api/agents/:name/stores` replacing the whole array, vs. per-binding `PATCH /api/agents/:name/stores/:index`? Exactly one store per agent is the current norm (a child's `[[stores]]` REPLACES the base's under `extends` rather than unioning), which argues for whole-array replace.
- **Which fields are editable.** Tree URI, trusty-search index id, palace id — all three? Changing the tree URI repoints what the agent KNOWS, which is a bigger act than correcting a mistyped index id.
- **Validation before write.** Reject a binding whose index does not resolve, or accept it and let the existing live-status path report NOT CONNECTED? Accepting is more consistent with the current fail-soft posture (an unknown index degrades to a reported disconnect, never a boot failure).
- **`extends` interaction.** Editing a child persona's binding must write to the child's own `agent.toml`, never the base's. The GUI needs to show which file it is about to modify.
- **Round-trip fidelity.** `agent.toml` is hand-editable and comment-rich. A naive serialize-and-write would strip an operator's comments; this likely needs `toml_edit` (already a trusty-agents dependency) rather than `toml`.
- **Concurrency.** Two panels open, or a hand-edit racing a PATCH. Some form of read-modify-write guard or ETag.

## Acceptance

- A binding's index id can be corrected from the config panel and the pane re-reports live status without a restart.
- Comments and unrelated keys in `agent.toml` survive the write.
- Editing a persona that `extends` a base writes only to the persona's own file.

Refs epic #3052, #3878.

bobmatnyc:
[...]

PR #4119, opened 2026-07-27
feat(trusty-agents): Knowledge pane backend — unified knows-surface API

### LINKED ISSUE #3935: Knowledge-section backend: unified knows-surface API (stores + knowledge tools + MCP connections)
Epic: #3931. Spec: `docs/specs/agent-config-five-sections.md` §4 (`SPEC-AGENTCFG-03~draft`), Phase 3.

Owner directive: *"Knowledge (list knowledge tools including and MCP connections to knowledge stores)"*.

## Scope

One section listing **everything the agent knows**, subsuming DOC-54 §5.1's Stores leg and widening it to three sub-surfaces rendered as one pane:

| | Sub-surface | Source | Today |
|---|---|---|---|
| **K-a** | Store bindings | `agent.toml` `[[stores]]` | LIVE (#3878) — `GET /api/agents/:name/stores` → `StoreStatus` with `connected`, `chunk_count`, `index_status`, `palace_connected`, `reason` |
| **K-b** | Knowledge tools | agent's effective tool set filtered to skills with `kind = "knowledge"` | NEW — derivable today only by eyeballing `[tools].allow` |
| **K-c** | MCP knowledge connections | harness `[[tool_registry.endpoints]]` / `[[mcp.services]]` backing those tools | NEW as a surface; the config exists |

The three are one pane because they answer one question. A user asking "what does this agent know?" must not have to correlate a store card, a tool glob and a harness endpoint table across three files.

## Classification

A tool is a *knowledge* tool **iff the skill wrapping it declares `kind = "knowledge"`** (depends on #3933). No hardcoded tool-name list — a hardcoded list would drift exactly the way the removed static `gworkspace` MCP tool list drifted (`assets/config/default-config.toml:22-41`).

Illustrative, not normative: `vector_search`, `memory_recall`, `memory_search`, `search_memory`, `retrieve_memory`, `store_memory`, `list_memory_keys`, `okg_sources`, `okg_ingest_*`, `search_code`, `search_skills`.

`vector_search` is the seam tying K-b back to K-a: its default index is the agent's own store (`persona.rs:346-355`, the #3864 fix). A knowledge tool bound to a specific store binding **renders under it**, not as a free-floating chip.

## Honest gap this must report truthfully

The two endpoints most obviously "MCP connections to knowledge stores" are **disabled by default**:

| Endpoint | Declared scopes | Status on main |
|---|---|---|
| `trusty-memory` | `["memory.read", "memory.write"]` | **`enabled = false`** — awaiting `--rpc` mode on the binary |
| `trusty-search` | `["search.read"]` | **`enabled = false`** — same |
| `gworkspace` | `google.*` families | `enabled = true`, self-discovering via `rpc.discover` |

(`assets/config/default-config.toml:193-221`.) The agent's memory and search capability today flows through **in-process** tools, not those endpoints. A configured-but-disabled endpoint renders as **DISABLED** with its reason — never as connected, never hidden. Rendering it as absent is the same class of defect as #3891's fabricated listener pane.

## Route

`GET /api/agents/:name/knowledge` — one call, so the pane cannot render a half-correlated view:

```jsonc
{
  "stores": [ /* StoreStatus, unchanged shape */ ],
  "tools":  [ { "tool": "vector_search", "skill": "semantic-search",
                "bound_store": "cto-assistant-kb", "available": true, "reason": null } ],
  "mcp":    [ { "name": "trusty-memory", "kind": "openrpc", "enabled": false,
                "connected": false, "scopes": ["memory.read","memory.write"],
                "reason": "endpoint disabled in config.toml" } ],
  "issues": [], "config_error": null
}
```

- **K-1** MUST NOT fail on a degraded sub-surface. Each of `stores`/`tools`/`mcp` resolves independently; a probe timeout populates `reason` and leaves the other two intact. Mirrors `agent_stores.rs`'s two-phase load and #3878's "declarative data only, degrade never fail" posture.
- **K-2** `GET /api/agents/:name/stores` is **retained unchanged**; `/knowledge` is additive and #3878's consumers keep working.
- **K-3** Store **editing is out of scope** and remains #3890's contract (route shape, editable fields, validation are all undecided there). The Knowledge pane ships read-only and gains editing when #3890 lands.

## Depends on

#3933 (skill-wrapping runtime) for the `kind = "knowledge"` classifier. Until it lands, K-b can be served from a provisional classifier and swapped without a route change.

## Acceptance (spec §4.6)

- **C-03.1** A `[[stores]]` binding to a nonexistent index renders NOT CONNECTED with a reason, identically to `/stores`.
- **C-03.2** A disabled knowledge endpoint appears in `mcp[]` with `enabled: false` and a non-null `reason` — never omitted.
- **C-03.3** Every tool in `tools[]` names a skill in the agent's resolved skill set — no orphans.
- **C-03.4** Killing the search daemon degrades `stores[].connected` to `false` without changing the HTTP status code.

### LINKED ISSUE #4115: GET /api/agents/{name}/stores reports connected:true for an index whose corpus failed to open
## Evidence

After a trusty-search daemon restart, index `cto-duetto` came up with:
- `corpus_open_failed: true`
- All stages `Failed`
- `chunk_count: 0`
- Completely non-functional — returning zero results for both lexical and semantic queries

Nevertheless `GET /api/agents/cto-assistant/stores` reported:
```json
{"name":"cto-duetto-kb","index":"cto-duetto","connected":true,"index_status":"ready","chunk_count":0}
```

## Root Cause

`resolve_store_statuses` (crates/trusty-agents/src/stores/status.rs) only checks that `GET /indexes/{id}/status` returns HTTP 200. It does not inspect `stages` or `corpus_open_failed`. A reachable-but-broken index is therefore indistinguishable from a healthy one.

## User-Visible Consequence

The GUI store card shows a green "connected" indicator for a knowledge store that can answer nothing. The agent will simply behave as though it has no knowledge, with no error surfaced anywhere.

## Why This Matters

This is the same silent-capability-loss class that has cost real investigation time repeatedly — a missing search index, silently dropped tool grants, and a clobbered config all presented as "the agent is behaving oddly" rather than as an error. A status endpoint that reports green on a dead dependency actively works against diagnosing that class.

Related to issue #3935 (Knowledge backend for the five-section agent config) — that pane is being implemented now and should surface real health, not HTTP reachability. References DOC-57's "state the gap, never fabricate" principle.

## Suggested Fix Direction

Derive `connected`/`index_status` from the status payload's actual `stages` and `corpus_open_failed` fields, and surface a distinct degraded/failed state rather than collapsing everything to a boolean.

bobmatnyc:
## Summary

- Implements #3935 (DOC-57 §4, Phase 3 of the five-section agent-config model): `GET /api/agents/:name/knowledge`, the Knowledge pane's single backend call.
- Merges three sub-surfaces in one response:
  - **K-a store bindings** — `[[stores]]` resolved live against trusty-search/trusty-memory. Reuses `stores::resolve_store_statuses` unchanged; `/stores` is untouched and stays available.
  - **K-b knowledge tools** — the agent's resolved tool set filtered to skills classified `kind = Knowledge` in #3933's compile-time skill catalog (no hardcoded tool-name list). `vector_search` renders under its bound store (the `bound_store` field) when the agent has one, matching the seam DOC-57 §4.3 names explicitly.
  - **K-c MCP knowledge connections** — declared `[[tool_registry.endpoints]]` and `[[mcp.services]]` entries that plausibly back a knowledge tool, read via `GlobalConfig::load()` (never creates the file). **No live probe is performed** — spawning an MCP subprocess on every knowledge fetch would be slow and side-effecting, so `connected` is always `false` here, with a `reason` distinguishing "disabled" from "not probed by this route."
- Read-only in this slice; store editing remains #3890's separate contract (K-3 in the spec).

## Honest gap discovered while implementing K-c

DOC-57 §4.4's illustrative table places `trusty-memory`/`trusty-search` under `[[tool_registry.endpoints]]`, `enabled = false`. The live `default-config.toml` has already moved both onto `[[mcp.services]]` (native MCP-stdio, `enabled = true`, `discover = true`) — the OpenRPC `--rpc` mode they were waiting for never shipped, so #3987's fix line replaced the dead endpoints with working native services. This route reads the config live and reports what is **actually** configured, not the issue's now-stale illustration; the module doc on `knowledge_mcp_connections` records this explicitly.

## Test plan

- [x] `cargo test -p trusty-agents --lib` — 2749 passed, 0 failed (full crate suite, including 17 new tests for this route)
- [x] `cargo fmt -p trusty-agents -- --check` — clean
- [x] Honest-degradation cases covered directly: unbound agent, store bound to a nonexistent index (200 + reason, never 500), ungranted knowledge tool (with reason), disabled MCP endpoint (never rendered as connected), malformed `agent.toml` (degrades to `config_error`, never 500), router wiring.

Closes #3935
Closes #4115


bobmatnyc:
Addressed the code-critic BLOCK. Summary of fixes, pushed at 8d89aacd:

**CRITICAL — false-green fixed at the source (closes #4115).** `stores/status.rs::resolve_store_statuses` now derives `connected` from the `/indexes/{id}/status` response's `stages` object (any lane reporting `status: "failed"`) instead of HTTP reachability alone. New `StoreStatus.failed_stages` field names which lane(s) failed. This fixes `/stores` and `/knowledge` in the same place — one fix, two surfaces, as requested. Regression test `reports_corpus_open_failed_as_not_connected` reproduces the exact `cto-duetto` shape (2xx, `status: "ready"`, `chunk_count: 0`, all stages failed) and asserts `connected: false`.

**HIGH — `[tools].search_indexes` now surfaced.** `/knowledge` reuses `agent_stores::attached_indexes_deduped` (made `pub(super)`) so it can never disagree with `/stores` about the id set. New `search_indexes` + `enforce_search_indexes` response fields, same names `/stores` already uses. Test `knowledge_route_surfaces_attached_search_indexes` uses the exact cto-assistant shape (one `[[stores]]` binding + `search_indexes = ["cto-duetto"]`) and asserts the second index appears.

**MEDIUM — K-c now agent-scoped.** Every MCP connection card carries `granted_for_agent`, computed via `agent_reaches_connection` against this agent's own resolved tool patterns (same `match_any_glob` the dispatch gate uses) — a documented, best-effort connection→tool mapping (`gworkspace` → the Google-OAuth knowledge tools; `trusty-memory`/`trusty-search` → the native memory/search family). A bare agent with zero grants now reports `granted_for_agent: false` for both connections instead of looking identical to a fully-provisioned one.

**LOW — doc comment fixed.** No longer claims no tool-less `Knowledge`-kind skill exists; `owner-context` is one, and its exclusion from `tools[]` is now documented as intentional (this array reports resolvable tools).

Full gate re-run: `cargo test -p trusty-agents --lib` — 2759 passed, 0 failed. `cargo clippy -p trusty-agents --lib --tests -- -D warnings` — clean. `cargo fmt --check` — clean.

Issue #4094, opened 2026-07-27
trusty-search integration tests register indexes in the developer's real daemon registry instead of an isolated one


bobmatnyc:
## Problem

Integration tests for trusty-search are registering index entries directly into the developer's real daemon registry at `~/Library/Application Support/trusty-search/indexes.toml`, rather than using an isolated test registry. This pollutes the developer's live state and creates downstream problems.

## Evidence

A live investigation of a developer machine's real trusty-search daemon (2026-07-27) found 11 stale index entries registered under `~/.trusty-search-test-roots/`:
- Named after issue numbers for test scenarios: `ts-2336-relocate-*`, `ts-3993-r2/r3-*`, `ts-2984-skip-kg-*`
- Also includes fixture names: `index-a`, `index-b`, `index-new`, `foo`, `bar`, `newcomer`, `symlinked`, `self-relocate`
- Paths had not existed for weeks (no corresponding directory on disk)

These stale entries persisted because they triggered the orphan-reaper's ambiguity deferral (see related issue), but more fundamentally: **the tests should never have written to the real registry in the first place**.

## Impact

1. **Test debris pollutes real health reporting**: the stale entries kept `search_health` reporting `status: degraded` / `warm_boot_degraded: true` indefinitely, masking real daemon health signal
2. **Ambiguous-candidate collisions cascade**: repeated test runs create collisions that poison the orphan-reaper's logic (see related issue #XXXX)
3. **Developer friction**: manual cleanup required to restore visibility; health reporting becomes unreliable

## Solution Direction

Point integration tests at an isolated daemon/registry (temporary XDG or Application Support directory per test run) so they cannot touch the developer's live state. This isolation should apply to:
- The daemon's registry file (`indexes.toml`)
- Any index root tracking or state

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Related: #4095 (orphan-reaper ambiguity deferral is the downstream consequence of this test isolation problem)

Issue #3922, opened 2026-07-25
fix(trusty-agents): dedup_seed_loads_recent_ids flaky under parallel tests — HOME environment race with partial lock coverage


bobmatnyc:
## Summary

Test `listeners::store::tests::dedup_seed_loads_recent_ids` (crate `trusty-agents`, file `crates/trusty-agents/src/listeners/store.rs:294-307`) failed once in the first CI run of PR #3897 with a concrete assertion failure, then passed on rerun. **Inferred root cause: process-global `HOME` environment variable is mutated by the test under a `HOME_LOCK` mutex, but some other code path in the test suite may access `$HOME`-derived paths without honoring the same lock.**

This is the SECOND occurrence of an identical code pattern in this workspace within a single CI session (first in #3902, which involved `TCODE_DATA_DIR` in trusty-code under partial lock coverage). Both tests share the same structural vulnerability: a process-global environment variable mutated under a lock that not all callers honor.

## Evidence

**Observation** (from code-critic analysis of run 30165303605, job 89697190403):
- Assertion failure: `assertion failed: ids.contains("id2")` at line 306
- Only failure out of 2494 tests
- Single assertion failure indicates the second event ID was not appended or not found

**Mechanism (inferred from code structure + failure assertion, not directly reproduced):**

Test code (`crates/trusty-agents/src/listeners/store.rs:238-307`):
```rust
let _guard = HOME_LOCK.lock().unwrap_or_else(|e| e.into_inner());
let tmp = tempfile::tempdir().unwrap();
set_test_home(tmp.path());  // unsafe { std::env::set_var("HOME", dir); }

EventStore::append(&sample_event("id1", "message.received")).await.unwrap();
EventStore::append(&sample_event("id2", "message.received")).await.unwrap();
let ids = EventStore::recent_ids(10).await.unwrap();
assert!(ids.contains("id1"));  // ← passes
assert!(ids.contains("id2"));  // ← FAILS — id2 was lost
```

The test redirects `$HOME` to a temp directory so `EventStore::events_dir()` (which calls `dirs::home_dir()`) resolves into that temp dir instead of the real `~/.trusty-agents/events/`. The lost "id2" event is **consistent with** a race: if another code path read or wrote `$HOME`-derived paths while this test's mutex was held but **without acquiring `HOME_LOCK` itself**, that concurrent accessor would see the real `$HOME`, causing writes to go to the wrong place or reads to see stale data from the real home directory.

This inference rests on:
[...]

Issue #3655, opened 2026-07-22
test(trusty-agents): 7 failures in tests/plain_cli_repl.rs — 'tagent did not exit within 120s'


bobmatnyc:
7 tests in crates/trusty-agents/tests/plain_cli_repl.rs fail with "tagent did not exit within 120s". Pre-existing, not caused by any in-flight PR. Observed 2026-07-22 on a worktree based off main d0862332.

Impact: these failures mask real regressions in the plain CLI REPL surface and make `cargo test -p trusty-agents` unusable as a green gate for the trusty-agents workstream (epic #3052 / Eve-style declarative agents #2791).

Needs root-cause: is the REPL hanging on stdin/EOF handling, or is the 120s harness timeout simply too short under full-suite parallelism?

Found during workstream triage. The exact failing test names should be captured by re-running `cargo test -p trusty-agents --test plain_cli_repl`.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
## Non-reproducible at current tip (`c50395ed`)

Verdict: the 7 `plain_cli_repl.rs` timeouts do **not** reproduce at the current `origin/main` tip. Ran the exact reported command plus narrower scopes on a fresh worktree based off `c50395ed` (baseline in the issue was `d0862332`, confirmed an ancestor). All green, well under the 120s bound.

### Raw evidence

**1. Single test, isolated:**
```
running 1 test
test quit_command_stops_the_loop_immediately ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out; finished in 3.79s
```

**2. The whole test file (all 7, run in parallel — the gate):**
```
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.06s
```
(confirmed twice: second run 4.42s)

**3. The exact reported command — full `cargo test -p trusty-agents` (the condition the issue describes as the trigger: thousands of unit tests + several subprocess-spawning integration binaries contending for CPU):**
```
Running unittests src/lib.rs
test result: ok. 2188 passed; 0 failed; 10 ignored; 0 measured; 0 filtered out; finished in 25.17s
...
Running tests/plain_cli_repl.rs
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.66s
```
[...]

Issue #3450, opened 2026-07-20
bug(trusty-mpm): tests litter operator's real ~/trusty-mpm-projects root (origin/.base bare-repo shape recurs in connectors/tm_tests.rs, plus a deterministic default-resolver gap)


bobmatnyc:
## Summary

Live, currently-active test-hygiene defect: `cargo test --workspace` runs deposit fixture directories directly into the operator's REAL `~/trusty-mpm-projects/` — not a sandbox — silently. Observed live 2026-07-20: 33 directories named `tmpXXXXXX` (tempfile's default random suffix) plus a `trusty-ci-nonexistent/` directory, all created by today's repeated local test runs.

This is worse than clutter: test code was operating inside the user's live projects root, where real project state (other repos, other sessions' worktrees) lives, and it did so with zero visible signal.

## Evidence

```
~/trusty-mpm-projects/tmpfvcsxp/origin/.base/{config,objects,HEAD,info,description,hooks,worktrees,refs,packed-refs}
   ... (33 total tmpXXXXXX dirs, same shape)
~/trusty-mpm-projects/trusty-ci-nonexistent/subdir-test-1724.old-layout-backup-1783914801670571000/.worktrees
```

## Root cause 1 (deterministic — no TMPDIR pollution required)

`crates/trusty-mpm/src/bin/tm/tests_behavior_b_tests.rs:1208` —
`guided_fallback_blocks_github_git_from_subdirectory` sets up a fake git repo
with remote `https://github.com/trusty-ci-nonexistent/subdir-test-1724.git`
and calls `fallback_protected`, but — unlike its two sibling tests in the same
file (lines 891-985, 1291+) — it never overrides `TRUSTY_MPM_REPOS_ROOT`. So
`ensure_base_clone` resolves `repos_root()` to its PRODUCTION default
(`crates/trusty-mpm/src/core/trusty_tools_config.rs:568-600` →
`workspace_root()` → `~/trusty-mpm-projects`) and attempts a real clone of the
nonexistent GitHub URL directly against the operator's live projects root —
**on every machine, every run, unconditionally.** The failed/partial clone
attempt, replayed across many test runs, produces exactly the observed
`trusty-ci-nonexistent/subdir-test-1724.old-layout-backup-<nanos>/.worktrees`
artifact (`migrate_old_layout_aside` in
`crates/trusty-mpm/src/daemon/managed_routes/inproject.rs:168` renames a
[...]

PR #4120, opened 2026-07-27
fix(trusty-mpm): isolate $HOME in every test that seeds ~/.claude.json

### LINKED ISSUE #3965: flaky: trusty-mpm session-prepare tests race on the shared $HOME/.claude.json fixture under workspace parallelism
TWO distinct tests in the same trusty-mpm module have now failed in CI on unrelated PRs, both with the same `enabledMcpjsonServers is an array` panic: `tests_manifest_toggle_trust_3934` (observed on PR #3964's first CI run) and `prepare_session_preseeds_enabled_mcp_servers` (observed on PR #3964's later run). Both passed on rerun of the same SHA, and both pass locally (193/193).

Suspected root cause: the shared `EnvVarGuard` fixture that points `$HOME` at a temp dir containing `.claude.json` is process-global, so tests mutating it race each other under `cargo test --workspace` parallelism. The panic shape (a field expected to be an array not being one) is consistent with one test reading the file while another is rewriting it.

Evidence of independence from the PR that surfaced it: PR #3964's diff touches zero files under `crates/trusty-mpm/`, `crates/trusty-common/`, or `crates/trusty-agents-common/`, and `trusty-mpm` does not depend on `trusty-agents` at all (only on `trusty-agents-common`, untouched). So there is no code path from that branch to these tests.

Impact: intermittent red CI on unrelated PRs, which costs a rerun cycle each time (~15-17 min for the Test job) and trains reviewers to dismiss red as flake — the dangerous habit.

Suggested fix directions (not prescriptive): give each test its own isolated $HOME/temp fixture rather than a shared guard; or serialize the affected tests (e.g. a shared mutex / `serial_test`); or make the fixture write atomically so a concurrent reader can never observe a partial file.

Note this is the THIRD distinct flaky-test family observed in this session, alongside `plain_cli_repl` (process-spawn timeouts under contention) and `listeners::store` ($HOME-dependent) — a consolidated pass on $HOME-global test fixtures may be worth more than three separate fixes.

bobmatnyc:
## Summary

Closes #3965. Root-caused and fixed the trusty-mpm session-prepare `$HOME`/`.claude.json` test-isolation race.

**Root cause (confirmed by direct inspection):** `prepare_session_inner` calls `preseed_workspace_trust_home`, which resolves `~/.claude.json` via the REAL process `$HOME` — not via the `fw`/`workspace` parameters most tests already isolate. Only a handful of tests redirected `$HOME` before driving that pipeline; the rest wrote real trust/`enabledMcpjsonServers` entries into the operator's actual `~/.claude.json` on every test run, keyed by throwaway temp paths.

This machine's real `~/.claude.json` had **534 of 543** `projects` entries as `tm-test-*`/`.tmp*` debris before this fix — direct, first-hand proof of the bug, not speculation. Under workspace parallelism, a `#[serial]` test that DID redirect `$HOME` could also have a concurrently-running *unguarded* test's seed land in **its** temp `.claude.json` — that's the exact `tests_manifest_toggle_trust_3934` flake #3965 describes (`enabledMcpjsonServers is an array` panic). Both symptoms — real-file pollution and the CI flake — share this one root cause.

## Reproduction

- Could not safely reproduce against the real `$HOME` (that would pollute it further, exactly the thing #3965 flags). Instead verified via tight before/after snapshots of the real `~/.claude.json`'s `projects` key count across repeated `cargo test -p trusty-mpm --lib` runs.
- Pre-fix: each full-suite run added 13–17 new junk entries (`tm-test-*`, `.tmp*`, shaped exactly like `session_launch`/`standalone`/`provisioner`/`connectors` test fixtures).
- Traced each shape back to its source test via the literal path fragments it wrote (`0c8f1a2b3c4d`, `---`, `owner/repo`, `.base/.worktrees/<uuid>`, etc.) and fixed every one found.
- Post-fix: repeated full-suite runs (3613 tests) leave the real file's project count completely unchanged.

## Fix

No sleeps, retries, or new locks — every fix is per-test `$HOME` isolation (`#[serial_test::serial]` + a `$HOME`-redirecting RAII guard, reusing `session_launch::tests::EnvVarGuard` where visible in-tree, a small local equivalent elsewhere since `pub(super)` visibility doesn't cross sibling module trees), matching the pattern already established in this crate for #3991/#4072.

Files touched (all test-only):
- `core/session_launch/tests.rs`, `tests_roster.rs`, `tests_scaffold_gitignore.rs`
- `core/standalone/load.rs`
- `core/deploy_validate/tests.rs`
- `provisioner/workspace/tests.rs`
- `daemon/managed_routes/lifecycle_tests.rs`
- `connectors/tm_tests.rs` (`create_session_full_lifecycle` — a #3450 follow-up: that fix isolated `TRUSTY_MPM_WORKSPACE_ROOT` but not `$HOME`)

## Relationship to #4094

Same root-cause pattern — tests reaching into `$HOME`-global / real daemon-registry state instead of an isolated per-run root — applied to trusty-search's index registry rather than `~/.claude.json`. **Not merged into this PR**: the fixes live in unrelated file trees in different crates (trusty-mpm vs trusty-search), so a single PR would mix concerns. Worth tracking as a sibling issue with a shared lesson: production code that legitimately writes to a machine-global path (a config file, a daemon registry) needs its own test-only injection seam, not just a `$HOME`/env override bolted onto callers after the fact.

## Other two families named in the PM's session note

`plain_cli_repl` and `listeners::store` (both in `trusty-agents`) were also mentioned as flaky in the originating session, but both already have their own CLOSED issues/fixes: `listeners::store::tests::dedup_seed_loads_recent_ids` closed via #3922/#3943 (dependency-injection fix), and `plain_cli_repl` closed via #3655. Issue #3965 itself is scoped specifically to the trusty-mpm session-prepare fixture race, so this PR does not touch trusty-agents.

## Residual finding (reported, not fixed here)

During repeated verification runs I observed one additional, low-frequency (~1-in-8) failure: `core::standalone::trust_seed_tests::test_preseed_managed_trust_no_home_write` occasionally fails its "must not write to \$HOME" assertion. I traced every production call site of `preseed_home_trust`/`preseed_workspace_trust_home`/`preseed_managed_trust` and every test driving them, and all are now `$HOME`-isolated — I could not find the remaining foreign writer causing this. Since this test only ADDS more tests to the shared `#[serial]` group (never removes protection), this is very likely a pre-existing, rarer-now-but-not-eliminated race, not something this PR introduces. Recommend a fast-follow issue to track it; flagging here rather than blocking this fix on it, per the "STOP and report, don't fix speculatively" guidance.

I also independently observed and can attribute to it in isolation: `telegram::supervisor::tests::healthy_run_resets_transient_backoff` failed once during my verification runs — confirmed unrelated (timing-based backoff test, passes standalone, module never touched by this PR).

## Test plan

- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings` (CI's exact command) — clean
- [x] `cargo test -p trusty-mpm --all-targets --no-run` — all targets (lib, bin, 30+ integration tests) compile
- [x] `cargo test -p trusty-mpm --lib` — 3613 passed, 0 failed, repeated ~9 times; real `~/.claude.json` project count unchanged across the final clean runs

### DIFF
--- a/crates/trusty-mpm/src/bin/tm/commands/session/start_tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session/start_tests.rs
@@ -27,6 +27,40 @@
 
 use super::{start_session, start_session_in_place};
 
+/// RAII guard restoring `$HOME` on drop (including panic) — mirrors the
+/// identical pattern in `trusty_mpm::core::session_launch::tests::EnvVarGuard`
+/// and siblings (the `tm` binary is a SEPARATE crate target from the
+/// `trusty-mpm` lib, so `pub(crate) mod test_support` in the lib is not
+/// visible here — each target needs its own copy).
+///
+/// Why (#3965): `start_session_in_place` calls the REAL
+/// `session_launch::prepare_session(fw, path)`, which seeds
+/// `$HOME/.claude.json` via the REAL process `$HOME` — a DIFFERENT
+/// resolution path from the `fw` parameter this test already isolates via
+/// `FrameworkPaths::under(tmp_home.path())`. Pairs with
+/// `#[serial_test::serial]`.
+/// Test: used by `session_start_in_place_writes_stash_and_hard_fails_on_daemon_unreachable`.
+struct HomeGuard(Option<String>);
+impl Drop for HomeGuard {
+    fn drop(&mut self) {
+        // SAFETY: paired with `#[serial_test::serial]` — no other thread
+        // reads/writes the environment concurrently.
+        match self.0 {
+            Some(ref p) => unsafe { std::env::set_var("HOME", p) },
+            None => unsafe { std::env::remove_var("HOME") },
+        }
+    }
+}
+
+/// Point `$HOME` at `home` for the duration of the caller's scope. Callers
+/// MUST be `#[serial_test::serial]` — see [`HomeGuard`].
+fn set_home(home: &std::path::Path) -> HomeGuard {
+    let prior = std::env::var("HOME").ok();
+    // SAFETY: serialized via `#[serial_test::serial]`.
+    unsafe { std::env::set_var("HOME", home) };
+    HomeGuard(prior)
+}
+
 /// Run `git <args>` in `dir`, panicking with full context on failure.
 ///
 /// Why: every test below needs a real (but disposable) git repo; a tiny
@@ -136,8 +170,11 @@ async fn session_start_dispatches_managed_new_for_github_repo() {
 /// POST — proving `prepare_session` still ran in place.
 /// Test: this function IS the test.
 #[tokio::test]
+#[serial_test::serial]
 async fn session_start_in_place_writes_stash_and_hard_fails_on_daemon_unreachable() {
     let tmp_home = tempfile::TempDir::new().expect("tmp home");
+    // #3965: `#[serial]` + `$HOME` override — see `HomeGuard` above.
+    let _home = set_home(tmp_home.path());
     let target = tempfile::TempDir::new().expect("tmp target dir");
     let fw = trusty_mpm::core::paths::FrameworkPaths::under(tmp_home.path());
 
--- a/crates/trusty-mpm/src/connectors/tm_tests.rs
+++ b/crates/trusty-mpm/src/connectors/tm_tests.rs
@@ -43,6 +43,34 @@ use trusty_agents_common::connectors::{
 
 use super::*;
 
+/// RAII guard restoring `$HOME` on drop (including panic) — mirrors the
+/// identical pattern in `core::session_launch::tests::EnvVarGuard` and
+/// siblings (each module needs its own copy; `pub(super)`/module-private
+/// visibility does not cross sibling module trees).
+///
+/// Why (#3450 follow-up, #3965): see the comment at its one call site in
+/// [`create_session_full_lifecycle`].
+struct HomeGuard(Option<String>);
+impl Drop for HomeGuard {
+    fn drop(&mut self) {
+        // SAFETY: paired with `#[serial_test::serial]` — no other thread
+        // reads/writes the environment concurrently.
+        match self.0 {
+            Some(ref p) => unsafe { std::env::set_var("HOME", p) },
+            None => unsafe { std::env::remove_var("HOME") },
+        }
+    }
+}
+
+/// Point `$HOME` at `home` for the duration of the caller's scope. Callers
+/// MUST be `#[serial_test::serial]` — see [`HomeGuard`].
+fn set_home(home: &std::path::Path) -> HomeGuard {
+    let prior = std::env::var("HOME").ok();
+    // SAFETY: serialized via `#[serial_test::serial]`.
+    unsafe { std::env::set_var("HOME", home) };
+    HomeGuard(prior)
+}
+
 /// Spawn the daemon's real HTTP API on a random loopback port (empty fleet,
 /// tmux faked). Mirrors `client::proxy::tests::spawn_test_daemon`, but —
 /// unlike that helper — serves with `into_make_service_with_connect_info`
@@ -261,6 +289,17 @@ async fn create_session_full_lifecycle() {
             _daemon_root.path(),
         );
     }
+
+    // #3965: the REAL `spawn_managed_cloned` handler this test drives also
+    // calls `session_launch::prepare_session_with_repo_url` /
+    // `home_trust_seed::preseed_home_trust`, which seed `$HOME/.claude.json`
+    // via the REAL process `$HOME` — a DIFFERENT resolution path from the
+    // `TRUSTY_MPM_WORKSPACE_ROOT` override above, so pinning that alone still
+    // let this test write real trust entries (keyed by the laundered
+    // `.../origin/.base/...` workspace path, see the comment above) into the
+    // operator's `~/.claude.json`. `#[serial]` (already present, for
+    // `TRUSTY_MPM_WORKSPACE_ROOT`) + this override close that second gap.
+    let _home_guard = set_home(_daemon_root.path());
 
     let req = CreateSessionReq {
         task: "list files".into(),
--- a/crates/trusty-mpm/src/core/deploy_validate/tests.rs
+++ b/crates/trusty-mpm/src/core/deploy_validate/tests.rs
@@ -18,6 +18,29 @@
 use super::*;
 use tempfile::TempDir;
 
+/// RAII guard restoring `$HOME` on drop (including panic) — mirrors the
+/// identical pattern in `core::standalone::load::tests::HomeGuard` and
+/// `session_launch::tests::EnvVarGuard`.
+///
+/// Why (#3965): [`validate_and_repair`] falls through to
+/// `session_launch::prepare_session_with_repo_url`, which seeds
+/// `$HOME/.claude.json` via the REAL process `$HOME` (`preseed_workspace_trust_home`),
+/// not via the `fw`/`workspace` parameters this function takes. Pairs with
+/// `#[serial_test::serial]` so this can never write into the operator's real
+/// `~/.claude.json` or race a sibling test doing the same.
+/// Test: used by `repair_closes_gaps_on_incomplete_workspace`.
+struct HomeGuard(Option<String>);
+impl Drop for HomeGuard {
+    fn drop(&mut self) {
+        // SAFETY: paired with `#[serial_test::serial]` — no other thread
+        // reads/writes the environment concurrently.
+        match self.0 {
+            Some(ref p) => unsafe { std::env::set_var("HOME", p) },
+            None => unsafe { std::env::remove_var("HOME") },
+        }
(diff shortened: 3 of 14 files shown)


bobmatnyc:
## Fixed all three code-critic HIGH findings

1. **`core/session_launch/tests_mcp_trust_seed_e2e.rs:190`** (`prepare_session_creates_working_trusty_mpm_entry_when_absent`) — added `#[serial_test::serial]` + `EnvVarGuard::set("HOME", tmp_home.path())` (the file already imported `EnvVarGuard` for a sibling test).
2. **Same file, line ~261** (`prepare_session_overwrites_hostile_trusty_mpm_and_review_entries`) — identical fix.
3. **`session_manager/tests.rs`** (`spawn_session_tmux_cwd_is_workspace`) — this file had none of the guard machinery, so I added a local `HomeGuard`/`set_home()` pair (mirroring the pattern in `provisioner/workspace/tests.rs`/`connectors/tm_tests.rs`) and applied it. Confirmed via `provisioner/workspace.rs:866-877` that `preseed_home_trust` does sit before the `if !self.prepare` gate, exactly as flagged.

Also re-audited every other production call site of `preseed_home_trust`/`preseed_workspace_trust_home`/`validate_and_repair`/`WorkspaceProvisioner::provision` crate-wide. Two more exist (`session_manager::reconcile`'s `validate_and_repair` call, `daemon::managed_routes::lifecycle::spawn_managed_local`'s `provision_in` call) but neither is reachable by any test in `--lib` today — both require fixtures (a populated `pane_cwd_override`, or `BackendParams` routed through the local-dir-to-managed path) that no existing test constructs. Nothing further to guard right now; flagging for anyone who adds tests against those paths later.

## Residual flake: re-run under load, with the number

The reviewer's mechanism was correct, not mine. `$HOME` is a process-global var; `#[serial]` only excludes other `#[serial]`-tagged tests from each other — it does nothing against a non-serial thread mid-write. All three fixes above are exactly the "unguarded non-serial writer" class the reviewer described as the causal path into `test_preseed_managed_trust_no_home_write`'s fake-`$HOME` window.

**Before this follow-up fix:** the flake reproduced 1 time in ~8 full `cargo test -p trusty-mpm --lib` runs (~12.5%).

**After this follow-up fix:** ran the full `--lib` suite (3613 tests) **12 consecutive times**. Result: **0 failures, 0 occurrences of the flake, in 12/12 runs.** The real `~/.claude.json` project count was also checked before/after this batch and did not grow (663 → 663).

I'm not calling this "seems better" — it's 0/12 where it was previously ~1/8, and the mechanism (three now-closed unguarded writers, all independently confirmed via code inspection to touch the real `$HOME`) fully accounts for the failure mode the reviewer described. I'd treat this as resolved rather than opening a fast-follow issue, but flagging the exact numbers per the ask rather than asserting confidence — reopen the ask if you want a larger sample before merging.

## Updated quality gate (re-run after these fixes)

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings` — clean (exit 0)
- `cargo test -p trusty-mpm --lib` — 3613 passed, 0 failed, ×12 consecutive runs

Pushed as `b22771f6`. Still not merging — awaiting review.

bobmatnyc:
## Fixed the fourth finding + full call-graph audit (pushed as `7c85baf0`)

**Fourth writer confirmed and fixed exactly as diagnosed:** `session_manager/naming_tests.rs::reconcile_resolves_adopted_session_cwd_from_pane` builds the populated `pane_cwd` fixture, `reconcile_on_boot` resolves it, and `reconcile.rs:220-224`'s `validate_and_repair(&fw, &workspace, None)` — with `fw = FrameworkPaths::for_managed_workspace(&workspace)` resolving its root from the real process `$HOME` — falls through to `preseed_workspace_trust_home` against the real `~/.claude.json`. Guarded it plus its two siblings (`reconcile_flags_unresolvable_adopted_session_as_unmanaged`, `reconcile_skips_external_adopt_when_workspace_already_tracked`) — traced both through the actual `reconcile.rs` code and confirmed neither currently reaches the sink (both hit an earlier `continue`/short-circuit), but guarded them anyway as instructed, since a future refactor of either short-circuit must not silently reopen the write.

**Full outward call-graph audit, this time from the sinks, across the whole crate, not the diff surface:** grepped literally (not sampled) for every call of `prepare_session*`, `preseed_home_trust`, `preseed_workspace_trust_home`, `preseed_managed_trust`, `validate_and_repair`, `.provision(`/`.provision_in(`, `run_prepare_session`, `load_alias`, `prepare_inproject_session`, `prepare_managed_config`, `reconcile_on_boot` across three places CI actually exercises but I hadn't checked with equal rigor before: `crates/trusty-mpm/src` (lib target), `crates/trusty-mpm/src/bin` (the `tm` **binary** — a separate cargo target from `--lib`), and `crates/trusty-mpm/tests` (integration-test binaries). CI's real gate is `cargo test --workspace`, not `--lib`, and both prior misses came from searching the diff surface instead of the call graph — so I widened the search to match.

That turned up five more genuine gaps, all now fixed:
- `bin/tm/commands/session/start_tests.rs` (**bin target**): `session_start_in_place_writes_stash_and_hard_fails_on_daemon_unreachable` drives the real `session_launch::prepare_session` directly.
- `tests/session_manager_mvp.rs` (**integration target**): `provisioner_isolates_workspace_under_root`, `handler_spawn_wires_provision_and_spawn`, `handler_spawn_creates_tmux_at_workspace_cwd` — all call `WorkspaceProvisioner::provision(...)` directly.
- `tests/local_spawn.rs` (**integration target**): `spawn_managed_local_redirects_to_managed_clone` calls `WorkspaceProvisioner::provision_in(...)` directly.

Every other hit from that grep is either a doc-comment mention (not a call), a production call site with zero test coverage today (verified by finding no `#[test]`/`#[tokio::test]` in the same file or an enclosing test module that reaches it), or a test that provably short-circuits before the sink — traced through the actual production code path each time, not assumed from the test's name or intent. Full list of "checked, confirmed safe, no change needed": `bin/tm/commands/misc.rs` (`tm validate`, no tests), `bin/tm/commands/standalone.rs` (`tm run`/`tm load`, no tests), `bin/tm/commands/meta/launch.rs::launch_and_wait` (only reached by an `#[ignore]`d live-tmux integration test, never run by plain `cargo test --workspace`), `bin/tm/commands/launch.rs::run` (no test calls it — only `session_matches_workdir` is unit-tested from that file), `runtime/claude_code.rs`'s `prepare_managed_config` fallback branch (already guarded crate-wide via `claude_code_tests.rs`'s pre-existing `HomeGuard`), `standalone/load.rs::load_alias` (no test calls it directly, only through `run_prepare_session` which is already guarded), and five more integration-test files that reference the managed-spawn HTTP surface (`resume_unresumable_mapping.rs`, `mcp_spawn_gate.rs`, `proxy_routes.rs`, `session_manager_slots.rs`) but never construct a `RealGitBackend`/`WorkspaceProvisioner` — confirmed by grep, they test only routing/rejection paths that return before any provisioning would occur.

## Statistics, as requested

Ran the full `cargo test -p trusty-mpm --lib` suite (3613 tests) **25 consecutive times** after this fix.

**Result: 25/25 clean. 0 failures in any run.** `test_preseed_managed_trust_no_home_write` did not fail once. No other test failed either (the unrelated `telegram::supervisor` timing flake from the first round also did not recur).

Also did one tight before/after bracket around an additional (26th) run to isolate the `--lib` suite's contribution to real-`~/.claude.json` pollution specifically: **824 → 824, unchanged.**

One honesty note on scope: I also ran the three newly-touched files once each under their real targets (`cargo test -p trusty-mpm --test session_manager_mvp --test local_spawn`, `cargo test -p trusty-mpm --bin tm ...start_tests`) to confirm correctness — all pass — but did not repeat those 25×, since the flake under investigation is specifically a `--lib` test and repeating the (much slower, git/tmux-touching) integration suite 25× didn't seem proportionate to what's being verified. Flagging this rather than silently narrowing scope: if you want the same 25-run treatment applied to `--all-targets`, say so and I'll run it.

## Quality gate (re-run after this fix)

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings` — clean (exit 0)
- `cargo test -p trusty-mpm --lib` — 3613 passed, 0 failed, ×25 consecutive runs (+1 bracket run = 26 total)
- `cargo test -p trusty-mpm --all-targets --no-run` — all targets (lib, bin, 30+ integration tests) compile
- `cargo test -p trusty-mpm --test session_manager_mvp --test local_spawn` — all pass
- `cargo test -p trusty-mpm --bin tm ...session_start_in_place_writes_stash...` — pass

Still not merging — awaiting review.

Issue #4093, opened 2026-07-27
[tools].allow glob/literal grants can match zero registered tools with no diagnostic on either dispatch path (create_draft, create_task, granola_*)


bobmatnyc:
Refs #3987 (the [tools].scopes analogue, partially fixed by #4018) and #3964 (which surfaced but did not fix two of these). Filed separately because the mechanism, the affected code paths, and the fix shape all differ from #3987's.

## Symptom

`[tools].allow` glob/literal patterns are matched via `match_any_glob` (crates/trusty-agents/src/ctrl/pm_task/helpers.rs:146) inside two independent pure filters — `filter_persona_tool_names` (persona.rs:67, persona-chat/Telegram/GUI path) and `scope_assistant_allowed_tools` (runtime/tool_registry.rs:529, --direct/subprocess path). Neither emits any diagnostic when a declared pattern matches zero names in that path's registry. This is the direct `[tools].allow` counterpart of #3987's `[tools].scopes` finding, and equally indistinguishable from "this agent has no tools" for the affected grant.

## Three confirmed instances

1. `create_draft` (cto-assistant/agent.toml:107) and `create_task` (cto-assistant/agent.toml:103) — exact literals matching NO tool anywhere in trusty-gworkspace's dispatch table (crates/trusty-gworkspace/src/server.rs:40-199), on either path, unconditionally. Superseded by `compose_email`/`manage_tasks`. PR #3964 surfaced these ("caught two real gaps while being written") but gave them skill cards rather than fixing the grant — they now render as ordinary `granted: true` in GET /api/agents/:name/skills, a step backward for discoverability versus appearing in `unmatched_patterns[]`.

2. `granola_*` (assistant/agent.toml base, inherited by izzie and cto-assistant via `extends`; also personal-assistant.toml) — alive on the persona-chat path (`mcp_service_tool_executors()`, tools/mcp_service_tools.rs:243, registering granola_search/granola_get/granola_list_recent from static [[mcp.services]] config), but that function is NEVER called from runtime/subagent_mode.rs — confirmed by repo-wide grep. On the --direct/subprocess path this glob matches zero registered tools.

## Why this needs a fix, not just #4018's scope-side warning

[...]

Issue #3936, opened 2026-07-25
Permissions backend: structured permissions model (scopes, tiers, authority, autonomy)


bobmatnyc:
Epic: #3931. Spec: `docs/specs/agent-config-five-sections.md` §7 (`SPEC-AGENTCFG-06~draft`), Phase 4.

Owner directive: Permissions as a first-class section — *"the allow-lists, scopes, user-authority flags, RBAC tiers — surfaced as configuration instead of buried TOML."*

## Why this is a section, not a footnote

An agent's permission posture is spread across **four unrelated mechanisms in three locations, with three different failure polarities**:

| # | Mechanism | Where | Enforced? | Polarity when empty/absent |
|---|---|---|---|---|
| 1 | `[tools].allow` globs | `agent.toml` | Yes — `filter_persona_tool_names` (`persona.rs:65`) | **Absent ⇒ zero tools** on the persona path (`persona.rs:255,480`) |
| 2 | `[tools].scopes` | `agent.toml` | Yes — `agent_can_use` (`scope.rs:125`), **persona-chat path only** | **Empty ⇒ denies everything** (fail-closed) |
| 3 | `ServiceTier` RBAC | Rust + `SLACK_RBAC_USERS` env + `[rbac]` | Partly — `filter_tools_for_user` runs; the `[rbac]` block has **no read site** | **`Default = All`** ⇒ degrades **open** |
| 4 | Endpoint scope filter | `~/.trusty-agents/config.toml` | Yes — `scope_enforcement = "deny"` | Operator-level, not per-agent |

Three polarities across four mechanisms is not a model a user can reason about from a GUI, and two of the four are invisible in config entirely.

Two further gaps: **scope enforcement is partial by construction** — only tools whose `ToolExecutor::scope()` returns `Some` are gated (i.e. OpenRPC-discovered tools, `tools/registry/adapter.rs:63`); every in-process tool returns `None` and passes the gate unconditionally, and the `--direct` path applies no scope filtering at all (`tool_registry.rs:530`). And **`user_authority` does not exist yet** — DOC-41 §5.5 specifies the singleton in full, but `AgentConfig` has no such field; the carve-out is a comment at `extends/mod.rs:252-262` and the test is `#[ignore]`d (`extends/tests.rs:619`). Tracked as #3074/#3075.

## Data model

```toml
[permissions]
scopes = ["memory.read", "memory.write", "search.read", "google.gmail.*"]
[...]

Issue #3074, opened 2026-07-18
epic(trusty-agents): authority model for user-authorized actions


bobmatnyc:
## Goal

An agent may propose actions requiring user authority but cannot silently acquire, inherit, or delegate that authority.

## Scope

- singleton user-authority declaration that cannot be inherited through `extends`;
- executing identity threaded through dispatch and delegation;
- propose-only behavior for delegated privileged actions;
- protected credentials fail closed through canonical epic #4040;
- project trust bound to repository identity rather than a reusable path;
- safe shell/runner argument policy and an explicit bundled-agent bash posture;
- structured proposed actions for approval surfaces;
- bounded capability tokens only if later required.

## Done when

Authority is explicit and attributable, multi-hop delegation cannot escalate, reused directories do not inherit trust, runner flags cannot escape the approved command policy, and review surfaces display the exact proposed action.

Former authority leaves #3075–#3081 plus #2934, #3051, and #3691 are consolidated here.

bobmatnyc:
## Consolidation update (2026-07-28)

This is now the canonical authority model issue. It absorbs #3075, #3076, #3077, #3078, #3079, #3080, #3081; the body preserves their actionable scope while the individual leaf issues remain available as design history.

bobmatnyc:
Triage consolidation (2026-07-28): this issue now preserves the listed former leaves as explicit scope and completion criteria. Their closure reduces duplicate queue surface and does not claim the work is implemented.

bobmatnyc:
Closed not planned in the 2026-09-02 milestone-68 review (owner-confirmed): dormant scheduled-tasks / notify-owner / personal-assistant cluster — its only open references were sibling epics in the same cluster, no movement since the 2026-08-14 sweep. Reopen if the personal-assistant scope returns.

PR #4127, opened 2026-07-27
feat(trusty-agents): Permissions pane backend — structured permissions model

bobmatnyc:
## Summary

- Implements #3936 (DOC-57 §7, Phase 4 of the five-section agent-config model): `GET /api/agents/:name/permissions`.
- New `[permissions]` `agent.toml` section (`agents/permissions.rs`): `scopes`, `user_authority` (DOC-41 §5.5's reserved singleton, now a real parsed field), `default_tier`/`unauthenticated_tier`, `autonomy`, `[[grants]]`. `[permissions]` **describes** the four scattered enforcement mechanisms DOC-57 §7.1 catalogs — it invents no new enforcement (PM-1).
- **Every response field carries an honest `enforced` boolean (PM-6):**
  - `scopes[]` → `enforced: true`. `agents::permissions::effective_scopes` is the ONE place CC-9 precedence is decided (`[permissions].scopes` wins over legacy `[tools].scopes` within a single file — the union is deliberately not taken), and it's now the SAME function `ctrl::pm_task::dispatch::persona`'s scope gate calls. That's what makes the `enforced: true` claim true rather than aspirational.
  - `user_authority`, `tiers`, `autonomy`, `grants` → `enforced: false`, each with a `reason` — none has a live enforcement site in this crate yet (`user_authority`: #3074/#3075; tiers: DOC-57 §7.1 says `[rbac]` "has no read site"; `autonomy`: PM-2, DOC-23 owns the real model; `grants`: PM-1/PM-4).
- `source` on each scope distinguishes `declared` (this agent's own file) from `inherited:<base>` — the DOC-57 §7.3 defect class (an agent granting tools whose scopes it never actually declared) becomes visible in the pane instead of requiring a reader to simulate `merge_extends`.
- `extends/mod.rs`: `permissions.scopes` unions base-first across the chain (a partially-migrated chain — base on legacy `[tools].scopes`, child on new `[permissions].scopes` — still unions correctly, tested explicitly), `grants[]` unions by `skill`, `user_authority` is **never** inherited (PM-3) — un-ignores the `extends_does_not_inherit_user_authority` regression test that had been stubbed since AUTH-1/#3074 was filed.
- Read-only in every phase (PM-4); grants or widens nothing (C-06.4) — the route resolves `AgentConfig::by_name_in`'s existing chain, nothing new.

## Honest-degradation behavior

- Malformed `agent.toml` → `200` + `config_error`, never `500`.
- An unresolvable `extends` chain (missing base, cycle) → `200`, falls back to the agent's own declarations (every scope reports `source: "declared"`), plus a non-normative `extends_warning` field naming the gap.

## Test plan

- [x] `cargo test -p trusty-agents --lib` — 2755 passed, 0 failed (full crate suite; one unrelated pre-existing flake — `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json`, a venv/environment race untouched by this PR, confirmed flaky independent of these changes by re-running clean)
- [x] `cargo fmt -p trusty-agents -- --check` — clean
- [x] `agents::permissions` + `agents::extends` unit tests (39 total) — including the newly un-ignored `extends_does_not_inherit_user_authority`, the partially-migrated-chain union test, and the grants-union-by-skill test
- [x] `api::server::agent_permissions` route tests (13 total) — declared vs. inherited provenance, CC-9 precedence, `user_authority` non-inheritance at the route level, malformed TOML, unresolvable `extends`, router wiring

## Adjacent defects noted, deliberately not touched

- `AgentDetail.scopes` in `projects.rs:531` still reads `cfg.tools.scopes` directly rather than through `effective_scopes` — that's the Phase-1 GUI mapping DOC-57 §8.5 documents as superseded by this route; left alone to keep this PR scoped to the new route + its own enforcement wiring.
- #3987 / #4093 (dead scope/tool-allow diagnostics on the two independent dispatch paths) — explicitly out of scope per the task; not touched.

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -19,6 +19,29 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `StreamAssembly` gained a `usage: Option<ChatUsage>` field so per-call token
   usage — which Bedrock reports exactly once, in a terminal event — survives
   the streaming path instead of being dropped.
+- **`GET /api/agents/:name/permissions` — the Permissions pane's structured
+  backend (#3936, DOC-57 §7).** A new `[permissions]` `agent.toml` section
+  (`scopes`, `user_authority`, `default_tier`/`unauthenticated_tier`,
+  `autonomy`, `[[grants]]`) describes the four scattered enforcement
+  mechanisms (`[tools].allow`, `[tools].scopes`, RBAC tiers, the endpoint
+  scope filter) as one coherent model rather than inventing a new one. Every
+  response field carries an `enforced` boolean — `true` only for `scopes[]`,
+  which is now resolved by `agents::permissions::effective_scopes` and fed
+  into the SAME persona-chat dispatch gate the route reports on (so the
+  claim is never aspirational); every other field (`user_authority`, tiers,
+  `autonomy`, `grants`) is honestly `enforced: false`, since none has a live
+  enforcement site yet. `[permissions].scopes` supersedes legacy
+  `[tools].scopes` within one file (CC-9 — the union is deliberately not
+  taken) but still unions base-first across `extends` (§2.3), so a
+  partially-migrated inheritance chain never silently drops a base's
+  un-migrated grant. `source` distinguishes `declared` from
+  `inherited:<base>` so the class of defect DOC-57 §7.3 records (an agent
+  granting tools whose scopes it never actually declared) is visible in the
+  pane. `user_authority` (DOC-41 §5.5's reserved singleton) is now a real,
+  parsed `AgentConfig` field and is never inherited across `extends`
+  (PM-3) — the previously `#[ignore]`d
+  `extends_does_not_inherit_user_authority` regression test is un-ignored.
+  Read-only in every phase (PM-4); grants or widens nothing (C-06.4).
 
 ### Removed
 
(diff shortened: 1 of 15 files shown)


bobmatnyc:
Addressed the HIGH finding from code-critic's WARN. Pushed at 4005b7a7:

**HIGH — `enforced: true` qualified to its actual dispatch path.** Each `scopes[]` element now also carries `enforced_on: ["persona_chat"]` (`SCOPE_ENFORCED_ON`), so the wire contract itself states the scope of the enforcement claim rather than relying on doc comments a GUI consumer never reads. `effective_scopes` feeds exactly one dispatch path — the persona-chat gate — and the `--direct`/subprocess path (`runtime::subagent_mode` → `scope_assistant_allowed_tools`) filters by `[tools].allow` name globs only, never a scope pattern. The module doc's PM-6 bullet now names both paths explicitly and cites #3987/#4093 as the tracked unification. New test assertions in `permissions_route_reports_declared_and_inherited_scopes` pin `enforced_on` on both a declared and an inherited scope.

The two non-blocking items you flagged (CC-9 silent-drop visibility before the first real `[permissions]` migration, and grants[] not being cross-checked against the skill catalog) — please file those separately as you offered; agreed they're real but not this PR's scope.

Full gate re-run: `cargo test -p trusty-agents --lib` — 2755 passed, 0 failed. `cargo clippy -p trusty-agents --lib --tests -- -D warnings` — clean. `cargo fmt --check` — clean.

PR #4155, opened 2026-07-27
fix(trusty-git-analytics): stop JIRA changelog pagination silently truncating oldest transitions

### LINKED ISSUE #3966: TGA JIRA ingestion capability: status transitions, comment detail, backfill + cron freshness
## CTO Decision

Date: 2026-07-26 | Decision Maker: Bob Matsuoka (CTO)

**TGA is now the owner of JIRA ingestion.** This shifts *production* of the JIRA-sourced fact tables upstream from cto-reports to trusty-git-analytics, enabling more efficient Rust-based computation.

This issue supersedes the production half of `duettoresearch/cto-reports#404` and `#405`; those are re-scoped to consumption-only.

> **Scope of this issue: the ingestion capability only — AC #1–#3.** Four criteria originally filed here (execute the historical backfill, weekly-cron wiring, cutover order, bridge-consumer readiness) are *operational* or belong to another codebase, and have been moved to [Out of scope](#out-of-scope--tracked-in-cto-reports) below. See [Ownership boundary](#ownership-boundary) for the contract.

## Ownership boundary

**We ship the capability; consumers operate it and report back on how they use it.**

That one line is the whole contract. Expanded (CTO decision, 2026-07-27):

**trusty-tools ships the TGA ingestion capability and stops there.** The deliverable is the JIRA extraction client, `fact_ticket_transitions`, `fact_jira_comment_detail`, and the `tga jira sync` / `tga jira freshness` commands — i.e. **the capability to land the data in `tga.db`, fresh and alarmable.**

**trusty-tools does not operate that capability.** Scheduling it, running it, and monitoring it are the consumer's operational concern, because the consumer owns the warehouse the data lands in and the reports built on it. A scheduled job that keeps someone else's warehouse fresh belongs to whoever owns that warehouse.

**Downstream consumers own validating their own usage.** It is explicitly **not** trusty-tools' job to verify how cto-reports (or any other consumer) reads these tables, whether the fields suit their reports, or when they retire their Python path. Consumers test against the populated tables and **report back** on how they actually use them — including any field-coverage or semantics gaps they hit.

That report-back is **feedback flowing to us, not a task we perform for them.**

To be unambiguous for anyone reading this issue cold: **trusty-tools does not owe any consumer a migration, a cron schedule, a soak-clock sign-off, or the retirement of another repository's code.** No such obligation can be inferred from this issue.

## Scope

TGA owns the JIRA ingestion *capability*:
- **JIRA API extraction**: authentication, pagination, incremental cursor management, changelog capture
- **Persist status transitions**: `fact_ticket_transitions` (ticket_key, from_status, to_status, timestamp, author)
- **Persist comment detail**: `fact_jira_comment_detail` (ticket_key, author, created, body_len minimum)
- **Land in tga.db**: both tables available for the cto-reports bridge via the existing `tga_to_duckdb.py` pattern (attach read-only). Making the tables *available* is ours; consuming them is not.
- **Complete changelog retrieval**: history is retrieved in full, or the command fails loudly — never silently truncated.
- **Freshness guard as a command**: `tga jira freshness` exists and bails loudly on stale data unless `--report-only`. Having the guard is ours; *invoking it on a schedule* is not.

## Acceptance Criteria

- [ ] **AC #1** — JIRA API extraction wired in TGA with proper auth/credentials, pagination, incremental cursor, error handling
- [ ] **AC #2** — `fact_ticket_transitions` schema defined and populating in `tga.db`
- [ ] **AC #3** — `fact_jira_comment_detail` schema defined and populating in `tga.db`

**This issue closes on code alone** — when PR #4067 and PR #4155 have landed. No backfill run, no cron schedule, and no wall-clock soak gate applies to closing this issue.

## What trusty-tools owes — and has delivered

| Deliverable | Status | Tracked by |
|---|---|---|
| JIRA extraction client (auth, pagination, incremental cursor, error handling) | delivered, in review | PR #4067 |
| `fact_ticket_transitions` (migration 0023) | delivered, in review | PR #4067 |
| `fact_jira_comment_detail` (migration 0023) | delivered, in review | PR #4067 |
| `tga jira sync` subcommand | delivered, in review | PR #4067 |
| `tga jira freshness` subcommand — fails loudly on stale data unless `--report-only` | delivered, in review | PR #4067 |
| Complete changelog retrieval — fails loudly rather than silently truncating | delivered, in review | PR #4155 (fixes #4084, stacked on #4067) |
| Documented field-coverage limitations | delivered — see below | this issue |

Both PRs are **in review; neither is merged yet.** PR #4155 is stacked on PR #4067's branch.

## Field-Coverage Constraints — known limitations consumers must validate against

These are realities of the source JIRA instance, not TGA defects. They are documented in schema design, and they are exactly the substance of the consumer report-back this issue asks for:

- `epic_key`: **100% NULL** in this JIRA instance — decomposition is via `parent_key`, not epics. Deliberately not modeled in the `fact_ticket_transitions` grain.
- `acceptance_criteria` (customfield_10017): **0% populated** at source.
- `story_points`: **76% NULL**, spread across 4 different custom-field IDs.

Consumers relying on any of these fields must validate against the populated tables and report back what they hit.

## Precedents: Failure Modes to Avoid

These two patterns from the cto-reports codebase informed the design. They are the direct justification for shipping `tga jira freshness` as a first-class, loud-failing command.

### 1. GFA→TGA Cutover Gap (tga_to_duckdb.py, bridge_commit_pr)
**What happened**: `bridge_commit_pr` silently froze at May 9 because it read a dead table. The GFA cron was deprecated May 20 and never ported to TGA. Result was zero rows for weeks with no alarm; fixed only in PR #390 (2026-07-23).

**Prevention**: Before retiring any Python extraction path, the Rust path must be verified producing data live in cron for at least 1 full week. *(The retirement decision, the cron, and its soak clock belong to the consumer — see Out of scope. What this repo owes is the alarmable Rust path that makes such verification possible at all.)*

### 2. Effort-Scoring Cron Gap (fact_commit_effort)
**What happened**: `tga.fact_commit_effort` was EMPTY because `tga backfill effort` was never wired into the weekly cron. The dashboard silently served a flat hardcoded median fallback (PR #393).

**Prevention**: any TGA-produced fact table needs a guard ensuring it is fresh (not empty, not stale), and that guard must be wired into whatever cron populates the table. **The guard is ours and now exists — `tga jira freshness`. Wiring it into a cron is the operator's.** This precedent is prevented by the consumer wiring the shipped guard into their schedule; it is not prevented by trusty-tools operating a schedule it does not own.

## Out of scope — tracked in cto-reports

The four criteria below are operational, or describe work in another codebase. **They cannot be satisfied or closed from trusty-tools** and have been removed from the acceptance criteria. Their substance is preserved verbatim here so nothing is lost:

**Historical backfill execution (formerly AC #4)**
> - [ ] **AC #4** — Historical backfill strategy documented and executed (bootstrap + go-forward)

**Freshness + weekly-cron wiring (formerly AC #5)**
> - [ ] **AC #5** — **CRITICAL: Freshness assertion + explicit weekly-cron wiring**
>   - [ ] Freshness validation: guard to alert if `tga.fact_ticket_transitions` or `tga.fact_jira_comment_detail` goes stale (similar to the Assertion J/K pattern in cto-reports)
>   - [ ] Explicit weekly-cron integration: the cargo workflow that runs on the cron schedule must invoke `tga jira sync` (or equivalent)

**Cutover order (formerly AC #6)**
> - [ ] **Cutover order defined and verified:**
>   - [ ] Rust path (tga.db) verified producing correct data for 1+ full weeks
>   - [ ] ONLY THEN retire the Python discard path (`scripts/extraction/jira/extract_jira_content.py`, `scripts/migration/load_jira_to_duckdb.py`)

**Bridge consumer readiness (formerly AC #7)**
> - [ ] cto-reports bridge consumer side scoped + ready (cto-reports#404, #405 after acceptance)

### Why these are not this repo's to close

1. **We ship the capability; we do not operate it.** Running a backfill (AC #4) and scheduling a weekly sync (AC #5) are executions of a shipped capability against a warehouse trusty-tools does not own. The *freshness half* of AC #5 is already delivered — `tga jira freshness` exists and fails loudly. What moved out is invoking it on a schedule, not having it.
2. **Different codebase.** `scripts/extraction/jira/extract_jira_content.py` and `scripts/migration/load_jira_to_duckdb.py` do not exist in trusty-tools and never have. Verified 2026-07-27: this repo contains 31 Python files (embedder sidecar, `python/trusty-voice`, four maintenance scripts) and **zero** of them reference JIRA. trusty-tools cannot retire a Python path that lives in another repository.
3. **Uncontrollable close condition.** The "1+ full weeks of verified live data" gate is a wall-clock soak on production cron infrastructure this repo does not own or operate. Leaving it here would hold a shipped, independently verifiable capability open for a week purely on a timer.
4. **Consumer-side work is consumer-side.** cto-reports#404 and #405 state their own boundary in their bodies: *"This is warehouse consumption (cto-reports) … No Rust work upstream."*

### What the consumer needs in order to pick this up

- `tga jira sync` — incremental JIRA extraction into `tga.db`
- `tga jira freshness` — staleness guard; exits non-zero on stale data unless `--report-only`

And three infra decisions that are theirs to make, because they own the infrastructure the answers describe: **which runner**, **how JIRA credentials are provided**, and **where `tga.db` lives**.

### Tracking status

- **#4085** (weekly cron wiring) was **closed as won't-do** under this boundary on 2026-07-27 — see #4085.
- **Formerly AC #7 (consumer bridge)** — nominally covered by `duettoresearch/cto-reports#404` and `#405`, both OPEN. Caveat: their bodies still describe the pre-decision Python load-step approach and have not yet been re-scoped to consumption-only as this CTO decision intends.
- **Formerly AC #4/#5/#6 (backfill execution, cron wiring, cutover + soak)** — consumer-side operational work. Tracking for these lives in cto-reports; trusty-tools does not file or manage issues in that repository.

## Related Issues

- PR #4067 — TGA JIRA ingestion: client + `fact_ticket_transitions` + `fact_jira_comment_detail` + `tga jira sync` / `tga jira freshness` (satisfies AC #1–#3)
- PR #4155 — fixes #4084, stacked on #4067 (complete changelog retrieval)
- #4084 — JIRA changelog pagination silently truncates oldest transitions (fixed by PR #4155)
- #4085 — weekly cron wiring — **closed as won't-do**, operational, belongs to cto-reports
- #4037 — onboard trusty-git-analytics onto the unified credential resolver (unaffected by the boundary; still ours)
- #3914 — epic: Generalize Activity/Work/Effort model beyond commits — PM work first
- duettoresearch/cto-reports#403 (umbrella, re-scoped)
- duettoresearch/cto-reports#404 (production half superseded here; re-scoped to bridge/consumption)
- duettoresearch/cto-reports#405 (production half superseded here; re-scoped to bridge/consumption)

### LINKED ISSUE #4084: bug(trusty-git-analytics): JIRA changelog pagination silently truncates oldest transitions
Parent: #3966 — **blocks AC #4 (backfill)**.

## Problem

JIRA's search-embedded changelog is **itself paged**. `search_with_changelog` reads only
`changelog.histories` from the `POST /rest/api/3/search` response and never consults
`changelog.total`, and there is no fallback to `GET /issue/{key}/changelog`.

Result: **any ticket with more history than one changelog page silently loses its OLDEST
transitions.** No error, no warning — the data just looks complete and isn't.

The limitation is self-documented in the code
(`crates/trusty-git-analytics/src/collect/jira/client.rs:320-325`, as introduced by PR #4067):

```
/// KNOWN LIMITATION: JIRA's search-embedded changelog is itself paged
/// (`changelog.total` vs `changelog.histories.len()`); a ticket with a
/// history longer than one changelog page will have its oldest
/// transitions truncated on this path. A full per-issue
/// `GET /issue/{key}/changelog` walk would close this gap but is
/// deferred — see issue #3966 tracking comment.
```

## Why this blocks the backfill specifically

The oldest transitions are exactly what a backfill exists to capture. Long-lived tickets — the ones
with the richest history and the most analytic value — are the ones most likely to exceed one
changelog page. Running the AC #4 backfill before this is fixed produces **confidently wrong**
historical data that is indistinguishable from correct data downstream.

Incremental sync is far less exposed (recent tickets, short histories), which is why this is
scoped as a backfill blocker rather than a hard stop on PR #4067.

## Fix direction

1. Per issue, compare `changelog.total` against `changelog.histories.len()`.
2. When `total > histories.len()`, fall back to a full per-issue walk of
   `GET /issue/{key}/changelog` (itself paged — loop it to exhaustion).
3. Keep the cheap embedded path as the fast case; only pay for the extra round trip on the
   truncated minority.
4. Add a test with a `changelog.total` exceeding `histories.len()` asserting the fallback fires and
   the oldest transition is present.

## Acceptance

- No ticket can lose transitions silently; truncation either self-heals via fallback or is loud.
- #3966 AC #4 unblocked.

bobmatnyc:
Targets `main`. No blockers remain — #4067 has landed, and this branch is
rebased onto current `main` (`56337097`). Every change is confined to
`trusty-git-analytics`.

Closes #4084.
Closes #3966 — this completes the JIRA ingestion surface that issue scopes to
this repo (the changelog + comment client, the `fact_ticket_transitions` /
`fact_jira_comment_detail` schema, and `tga jira sync` / `tga jira freshness`).
The downstream Python cutover and its consumers live in
`duettoresearch/cto-reports`, not here.

## Problem

JIRA's search-embedded changelog (`expand=changelog`) is **itself paged**.
`search_with_changelog` read only `changelog.histories` and never consulted
`changelog.total`, with no fallback to the dedicated endpoint.

Any ticket with more history than fits one embedded page silently lost its
**OLDEST** transitions — no error, no warning. Those are precisely the
transitions a historical backfill exists to capture, and long-lived tickets
(the richest ones analytically) are the most likely to exceed a page. The gap
was self-documented in a `KNOWN LIMITATION` doc comment instead of being
surfaced at runtime, which is how it could have silently produced
confidently-wrong backfill data indistinguishable from correct data.

## Fix

Per issue, compare `changelog.total` against the embedded entry count. When the
embedded copy is short, the ticket is **flagged** — `ChangelogIssue::
truncated_history_total` carries the count that proves it short — and
`run_sync` walks `GET /rest/api/3/issue/{key}/changelog` to exhaustion and
**replaces** (never merges) the truncated transitions. The common case — a
complete embedded changelog — issues **zero** extra requests.

The repair runs in `run_sync`'s per-ticket loop rather than inside the search
walk. That placement is load-bearing; see HIGH-3 below.

### How a failure surfaces (the crux)

No path can return a short history that reads as complete:

| Situation | Behaviour |
|---|---|
| Walk ends with fewer entries than anyone reported | `CollectError::IncompleteChangelog { key, expected, retrieved }` |
| That error during a sync | The **ticket** is recorded failed, its transitions are NOT written, the cursor clamps at/below it, and the run exits non-zero. The other tickets still land. |
| Transport / non-2xx during the repair | Same isolation — no degradation to partial data |
| Dedicated endpoint omits `total` | Cannot end the walk (`Option`, never a defaulted `0`); the search's own count remains a standing lower bound |
| `changelog.total` absent on the search | `warn!` naming the ticket and embedded count; no silent assumption of completeness |
| Server shrinks `total` mid-walk | The largest `total` anyone reported wins |
| Server ignores `startAt` | `CollectError::PagingBudgetExceeded` after 200 pages, rather than looping forever |

Failing loudly is deliberate: writing confidently-wrong backfill data is
strictly worse. `IncompleteChangelog` is its own variant precisely because this
is the one failure where the collector *has* data and it is silently wrong.

### Ordering / dedup

Both retrieval paths share one extractor, `transitions_from_histories`, so an
embedded entry and its repaired counterpart produce identical `JiraTransition`
values. That equivalence is what makes wholesale **replacement** safe: a
replacement cannot duplicate a transition, so it cannot collide on
`fact_ticket_transitions`'s `(ticket_key, transitioned_at, to_status)` primary
key. A stable sort pins the documented oldest-first ordering.

## Also in this diff

- **`fetch_comments` pagination.** Termination was `start_at >= total` with
  `total` defaulting to `0` when absent, so a response omitting the field ended
  the walk after one page — 100 of 150 comments ingested with no error, and
  because the fetch returned `Ok` the cursor advanced past the ticket
  permanently. `total` is no longer modelled; the walk terminates on a short
  page measured against the server's own effective `maxResults`.
- **`jql_time` DST ceiling.** The step-back search no longer emits an
  unvalidated bound when it exhausts its ceiling. The ceiling was 180 minutes
  on the grounds that "historic DST shifts never exceed two hours" — false:
  `Antarctica/Casey` folds exactly three hours (UTC+11 → UTC+08), landing on
  the ceiling with zero margin and saturating the loop. Saturation is now an
  explicit error with a remediation hint, and the ceiling is derived from the
  −12:00…+14:00 UTC-offset span rather than an empirical claim about tzdata.

## Review round 2 — the three HIGH findings from #4155's WARN

The critic reproduced two NEW instances of this PR's own bug shape inside the
fix. All three are closed, each with a regression test verified to fail against
the pre-fix semantics.

**HIGH-1 — `fetch_comments` compared against the page size REQUESTED.**
Atlassian paged beans echo the effective `maxResults` because the server may
apply a smaller one. A server paging 50 while we ask for 100 makes page 1 look
short. Reproduced: *50 of 150 ingested, `Ok`, cursor advanced.* Fixed by
modelling the echoed `maxResults` (clamped to the requested size, `Option` so
an absent one cannot read as `0`); an unstated page size ends the walk only on
an empty page. `total` remains unmodelled — that reasoning was sound.

**HIGH-2 — `ChangelogPageResponse::total` was `#[serde(default)] u64`.** An
absent `total` read as `0`, so the walk stopped after page 1 and `retrieved < 0`
made the completeness check pass vacuously — the repair could return *less than
the truncated data it was invoked to fix*. Reproduced: *fallback returned 1
transition, `Ok`, replacing the embedded set.* Fixed by `Option<u64>` plus
carrying the search's own count in as a standing lower bound, so an endpoint
that volunteers no `total` cannot satisfy the check by silence. `model.rs`
and `changelog.rs` now agree.

**HIGH-3 — one bad ticket wedged the entire run.** The repair ran inside
`search_with_changelog`, which `run_sync` awaits in full before writing
anything: one `IncompleteChangelog`/500/403 meant nothing ingested for ANY
ticket, no cursor movement, and a deterministic repeat next run. The search
walk now only records the verdict; `run_sync` performs the repair inside the
**existing** failed-ticket / cursor-clamp / circuit-breaker machinery the
comment path already had — reused, not duplicated. Two follow-ons fall out for
free: a boundary duplicate about to be discarded by keyset dedup can no longer
abort the run or spend a round trip, and the shadowed `truncated` binding is
gone.

Also bounded both per-ticket paged walks at 200 pages
(`CollectError::PagingBudgetExceeded`) — the only unbounded loops in this
client, each running once per ticket across up to 10,000 tickets.

## Review round 3 — SemVer hygiene (the one change since the APPROVE)

Round 2 came back **APPROVE**. The single MEDIUM acted on is the SemVer one;
the rest are deliberately left for follow-up.

`tga` is `publish = true` at 2.9.4, and this PR adds two `CollectError`
variants (`IncompleteChangelog`, `PagingBudgetExceeded`) and a new public
`ChangelogIssue` field (`truncated_history_total`). Neither is a patch-level
change on a published crate: a variant added to a non-`#[non_exhaustive]`
public enum breaks every downstream exhaustive `match` with E0004, and a field
added to a non-`#[non_exhaustive]` public struct breaks literal construction
and exhaustive destructuring. This is the same defect currently forcing
`trusty-common` to a 0.27.0 minor instead of a patch, and `Throttled` already
landed on `main` the same way via #4067 — a pattern worth stopping here.

Both types are now `#[non_exhaustive]`, with the rationale recorded on each.
Cost in-tree is zero, and both halves are load-bearing rather than reflexive:

- **`CollectError`** — every existing `match` in the workspace already carries
  a wildcard arm, so no call site changed. This is the half that matters:
  the enum gains a variant nearly every time a provider learns a new failure
  mode.
- **`ChangelogIssue`** — the struct half is free here, which is why it was
  worth taking. Its sole constructor is the `pub(crate)`
  `ChangelogIssue::from_api`; the only cross-crate consumer is the `tga`
  binary (`commands/jira/mod.rs`), which takes `&ChangelogIssue` and reads
  fields. `#[non_exhaustive]` blocks external literal construction and
  exhaustive destructuring — neither of which any caller does, in-tree or
  out — so no code contorted to accommodate it.

Deliberately **not** addressed (out of scope, no behaviour change): the
comment-walk corroborating bound, `continue` also skipping the ticket's
comment fetch, the unrepairable-ticket cursor pin, and the 1559-vs-1560
message mismatch.

## Tests

Wiremock-backed, no live network, following the crate's existing pattern.

Changelog repair:
- a truncated embedded changelog is flagged with the proving count, and the
  search walk issues **zero** requests to `/changelog`
- a complete embedded changelog is not flagged
- a missing `changelog.total` is not flagged
- the repair pages correctly across multiple pages (`startAt=0` then `startAt=2`)
- a short repair returns `IncompleteChangelog` with `expected: 5`,
  `retrieved: 2`, naming the ticket
- a broken `/changelog` endpoint does **not** abort the search walk
- recovered transitions persist through the real writer as 3 distinct rows —
  the shared entry appears once, PK intact, oldest transition present

Round-2 regressions, each mutation-verified:

| Test | Fails against pre-fix semantics with |
|---|---|
| `fetch_comments_honours_the_page_size_the_server_applied` | `left: 50, right: 150` |
| `fetch_changelog_keeps_paging_when_the_endpoint_omits_total` | `left: 2, right: 3` |
| `fetch_changelog_errors_when_an_absent_total_hides_a_shortfall` | returns `Ok([1 transition])` instead of erroring |
| `a_broken_changelog_repair_does_not_take_the_whole_run_down` | `left: 0, right: 1` rows for the *healthy* ticket |

Plus `fetch_comments_pages_to_empty_when_the_server_omits_max_results`.

## Quality gate

Re-run in full on the rebased head, after the round-3 change:

```
cargo check -p tga                        Finished `dev` profile — clean
cargo clippy --all-targets                exit 0; only pre-existing workspace
                                          config warnings (clippy.toml/Cargo.toml
                                          MSRV skew, duplicate bin targets in
                                          trusty-mpm/trusty-installer) — zero
                                          lint warnings from any code
cargo fmt --all -- --check                exit 0
cargo test -p tga                         763 + 151 + 1 + 2 + 1 + 5 = 923 passed;
                                          0 failed; 3 ignored — all pre-existing
                                          and live-network/parallel-bash gated
scripts/check_line_cap.sh                 7 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh            0 dangling pointers — OK
```

No test was ignored, cfg-gated, excluded, or deleted to reach green. The one
test whose premise the fix inverted (`search_with_changelog_propagates_
fallback_failure` — it asserted the abort that was the defect) is replaced by
two stronger tests: the client-level
`a_broken_changelog_endpoint_does_not_abort_the_search_walk` and the
end-to-end `a_broken_changelog_repair_does_not_take_the_whole_run_down`.

CHANGELOG entries added under `## [Unreleased]`.


bobmatnyc:
CI note: `.github/workflows/ci.yml` triggers only on `pull_request: branches: [main]`, so **GitHub reports no checks on this PR** while it targets `fix-3966-jira-ingestion`. That is the trigger filter, not a green run — do not read the empty check list as passing.

Signal available today is the full local gate, run against this exact commit (raw output in the PR body): `cargo check -p tga`, `cargo test -p tga` (full surface, 857 tests, 0 failed), `cargo clippy -p tga --all-targets -- -D warnings`, `cargo fmt --check`, `scripts/check_line_cap.sh`, `scripts/check_test_pointers.sh`.

Real CI signal arrives once #4067 merges and this retargets `main`.

bobmatnyc (commented):
## Verdict: WARN

Adversarial review at `df4eab73` against `main`. First review of this PR — it only became reviewable on retarget. CI 22/22 green, and the green is honest (see Notes). Zero CRITICAL, three HIGH.

Two of the three HIGH findings are **reproduced empirically**, not inferred: I ran probe tests against this exact commit and both produce silent partial data with an `Ok` return. Both are the fail-open / cursor-advance shape this PR exists to eliminate — they are new instances of it introduced by the fix.

## Findings

| HIGH | `crates/trusty-git-analytics/src/collect/jira/client.rs` | 656 | `fetch_comments` terminates on `n < COMMENT_PAGE_SIZE` — the size **requested**, not the size the server applied. A server paging smaller than requested ends the walk after page 1, returns `Ok`, cursor advances. Reproduced: 50 of 150 ingested. | Model the bean's echoed `maxResults` and break on `n < parsed.max_results`, or restore `total` as `Option<u64>` used as a corroborating bound. |
| HIGH | `crates/trusty-git-analytics/src/collect/jira/changelog.rs` | 90, 100 | `ChangelogPageResponse::total` is `#[serde(default)] u64`; an absent `total` reads as `0`, so the walk breaks after page 1 and `retrieved < 0` is false → `Ok(partial)`. The partial then **replaces** the embedded history. Reproduced. This is the exact `#[serde(default)] u64` fail-open this same PR condemns in `model.rs:360-370`. | `total: Option<u64>`; on `None` page until a short/empty page, and never break on a defaulted `0`. |
| HIGH | `crates/trusty-git-analytics/src/collect/jira/client.rs` | 552 | One ticket's fallback failure aborts the **entire run** with zero partial progress, deterministically, forever. The whole walk completes before `run_sync` writes anything, so nothing is ingested for any ticket and the cursor never moves. A ticket about to be discarded as a boundary duplicate can kill the run. | Give the changelog fallback the per-ticket isolation the comment path already has: record failed, clamp cursor at/below it, count toward the breaker, finish the walk. |
| MEDIUM | `crates/trusty-git-analytics/src/collect/jira/client.rs` | 627-660 | With `total` deleted, `fetch_comments` has **no loop bound at all** — the only unbounded paged walk in this client. A server ignoring `startAt` (or a proxy replaying page 1) spins forever accumulating comments, inside a per-ticket loop over up to 10,000 tickets. | Add a page cap mirroring `max_pages` (line 508) or the `reported_total` bound in `changelog.rs:100`. |
| LOW | `crates/trusty-git-analytics/src/collect/jira/client.rs` | 511, 548 | `truncated` is shadowed. The outer means "walk stopped at `--max-tickets`" and is surfaced to the user via `walk.truncated`; the inner means "this issue's embedded changelog is short". No current defect, but an edit assigning the outer from inside the loop silently hits the inner. | Rename the inner to `embedded_truncated`. |
| LOW | `crates/trusty-git-analytics/src/collect/jira/jql_time.rs` | 119 | `for _ in 0..MAX_STEP_BACK_MINUTES` tests offsets `0..=1559`, so the largest step-back actually attempted is **1559** minutes, while the doc (74-84) and the error message (130) both say 1560. The derivation argues the bound must *exceed* the 1560-minute offset span; the code stops two minutes short of the ceiling it argues for. | `0..=MAX_STEP_BACK_MINUTES`, or state 1559 in the doc and message. |

### HIGH-1 — `fetch_comments` short-page test compares against the requested page size

`client.rs:641,656`:

```rust
let n = parsed.comments.len();
...
if n < COMMENT_PAGE_SIZE {
    break;
}
```

`COMMENT_PAGE_SIZE` is what we *ask* for. Atlassian paged beans echo the effective `maxResults` back in the response precisely because the server is free to apply a smaller one (tenant limits, `jira.search.views.default.max` on DC, an intermediary). This PR deliberately un-models that field along with `total` (`model.rs:371-373`), so the walk now has no way to learn the page size actually in force.

Probe against `df4eab73` — a server honouring 50 per page while we request 100, on a ticket with 150 comments:

```
PROBE RESULT: ingested 50 of 150 comments in 1 request(s), returned Ok
```

Compare this PR's own CHANGELOG entry for the bug it is fixing: *"observed as 100 of 150 comments ingested with no error. Because the fetch still returned `Ok`, the failed-ticket cursor clamp did not protect it: the cursor advanced past the ticket and the loss was permanent."* The replacement termination condition reproduces that sentence verbatim under a different trigger. `run_sync` treats `Ok` as success (`commands/jira/mod.rs`, the `Ok(comments)` arm resets `consecutive_failures` and never records the ticket as failed), so `plan_cursor` advances and the loss is permanent.

The response already carries the value that makes this sound. Model it.

### HIGH-2 — the changelog fallback has the same `#[serde(default)] u64` hole it was written to close

`model.rs:328-333` and `changelog.rs:90-100`:

```rust
#[serde(default)]
pub(crate) total: u64,
```
```rust
reported_total = reported_total.max(parsed.total);
let n = parsed.values.len() as u64;
histories.extend(parsed.values);
if n == 0 { break; }
start_at += n;
if start_at >= reported_total { break; }
```

If the dedicated endpoint omits `total`, `reported_total` stays `0`; after page 1 `start_at >= 0` breaks the loop; `retrieved < 0` is false, so the completeness check passes and page 1 is returned as a complete history. That partial then **replaces** the embedded transitions at `client.rs:552`, so the fallback can hand back *less* than the truncated data it was invoked to repair.

Probe against `df4eab73` — search reports `changelog.total: 3` with one embedded entry; the dedicated endpoint returns one entry and no `total`:

```
PROBE RESULT: fallback returned 1 transition(s), Ok, replacing the embedded set
```

`model.rs:360-370` in this same PR argues, correctly and at length, that *"under `#[serde(default)]` an absent field silently reads as `0`, and a loop that terminates on `start_at >= total` would then stop after its first page and report success"* — then `changelog.rs` does exactly that. The doc at `model.rs:321-327` addresses only wedging (*"the walk still terminates on an empty page, so a missing `total` cannot wedge it"*), not truncation, which is the failure that matters here.

### HIGH-3 — one ticket's fallback failure loses the whole run, permanently

`client.rs:547-555` runs the fallback for every issue in the page, propagating with `?`. `commands/jira/mod.rs` awaits the *entire* walk (`client.search_with_changelog(&scope, max_tickets).await?`) before writing anything. So any `IncompleteChangelog`, 500, or 403 on a single ticket's changelog means: no transitions written, no comments fetched, no cursor update, for **every** ticket in the window. The failure is deterministic, so the next run reproduces it at the same ticket. There is no `--skip-incomplete` escape hatch.

The comment path in the same file already solved this problem: per-ticket `Err` arm, `failed_tickets` + cursor clamp via `plan_cursor`, and `MAX_CONSECUTIVE_COMMENT_FAILURES` as a breaker — one bad ticket holds the cursor but the other 9,999 still land. The changelog fallback gets none of that, and it is the path meant to unblock the #3966 historical backfill.

On the disclosed residual (fallback runs before dedup): the *bookkeeping* is genuinely clean — I verified `pager.record_page` (557) consumes only `(key, updated)` from `items` (556), and the fallback mutates only `issue.transitions`, so the keyset cursor is untouched. The engineer's "pays for it twice" cost analysis is correct. But the ordering has a consequence the doc comment does not name: because the fallback is awaited at 552, *before* `record_page` at 557 and the `is_new` filter at 559, a truncated ticket that is a boundary re-read — data already held and about to be discarded — can abort the entire run. That is the part worth fixing, not the extra round trip.

## Required Changes

1. `client.rs:656` — terminate on the server's effective page size, not the requested one. Model `maxResults` on `CommentSearchResponse`.
2. `changelog.rs:90-100` / `model.rs:331-332` — make `total` an `Option<u64>` and stop treating an absent one as `0`; a defaulted `0` must never satisfy the completeness check.
3. `client.rs:552` — isolate changelog-fallback failures per ticket instead of aborting the walk, reusing the `failed_tickets` / cursor-clamp / breaker machinery the comment path already has.
4. `client.rs:627-660` — bound the comment walk.

Items 5-6 (the two LOW findings) are optional polish.

## Notes

**Green-by-deletion: clean.** `cargo test -p tga --all-targets` at `df4eab73`:

```
test result: ok. 759 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 150 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

The single ignore is `core::effort::tests::effort_cross_validate`, pre-existing. All 12 claimed new tests are present and report `... ok`. No new `#[ignore]`, no cfg-gating, no `--lib` narrowing, no `--exclude`.

**Cross-module callers, checked with full context (not module-local).** `transitions_from_histories` has exactly two callers (`model.rs:82`, `changelog.rs:113`) and both feed the same `JiraTransition` shape — the replacement-not-merge argument holds. `embedded_changelog_is_truncated` has one caller (`client.rs:548`). `CommentSearchResponse::total` has no surviving readers; the only constructor is `client.rs:636` and the assertion in `client_tests.rs:274` was updated rather than deleted. No stale readers anywhere.

**The newly-fallible `build_jql` ripple: no swallowing.** Three non-test call sites, all `?`-propagating: `client.rs:527`, `commands/jira/mod.rs:270`. No `unwrap_or_default`, no empty-JQL fallback. `trusty-review`'s `build_jql` is an unrelated function on a different type. The hoist claim checks out — `let logged_jql = build_jql(&scope, tz)?` at `mod.rs:270` does precede `search_with_changelog` at 281, and since the first page uses the same `scope.since`, it genuinely proves the first bound renders. It does not cover the per-page re-anchors at `client.rs:521`, but a mid-walk failure there is harmless: nothing is written until the walk returns.

**On the DST change — checked independently, and it holds up.** I reverted `MAX_STEP_BACK_MINUTES` to 180 while keeping the error path: `jql_date_is_safe_across_the_largest_fold_in_the_tz_database` fails at `2010-03-04 14:59:00 UTC` for `Antarctica/Casey`. The saturation claim is real and the ceiling raise is justified.

One calibration note for readers, not a finding: reverting to the *true* pre-PR code (180 **and** emit-last-candidate on saturation) leaves that test green. So it guards the new no-saturation invariant rather than demonstrating a fixed user-visible defect. That matches the PR's own honest framing ("the emitted literals happened to stay safe, but with zero margin") — I mention it only so nobody later cites the test as proof that a data bug was fixed.

**PR description is stale.** The body still opens with "Stacked PR — base is `fix-3966-jira-ingestion`, not `main`. **Merge #4067 first**, then this," and its Tests/Quality-gate sections describe neither the `fetch_comments` change nor the `jql_time` change — both of which are in this diff and both of which carry CHANGELOG entries. Worth refreshing before merge so the squash commit message is accurate.

bobmatnyc (commented):
## Verdict: APPROVE

Round-2 adversarial review at `5a7c9d16`. Zero CRITICAL, zero HIGH. All three round-1 HIGH findings are closed, and I verified each by **applying the mutation myself** rather than taking the claim — every one reproduces the stated failure output exactly. Four MEDIUM findings remain; none blocks merge.

I went looking specifically for a fourth instance of the fail-open / cursor-advance shape, since round 1 found the fix had re-introduced it twice. I did not find one in the paging or cursor logic. What I did find is one residual instance of the *pattern* in the comment walk (MEDIUM-1, probed) and one piece of unjustified collateral (MEDIUM-2).

## Findings

| MEDIUM | `crates/trusty-git-analytics/src/collect/jira/client.rs` | 648-680 | `fetch_comments` has no corroborating bound. A server echoing `maxResults: 100` while actually paging 50 ends the walk after page 1 and returns `Ok`. Probed: **`Ok(50)` of 150 comments in 1 call.** The sibling `fetch_changelog` gained exactly such a bound this round; the comment path did not. | Model `total: Option<u64>` on `CommentSearchResponse` and use it *only* as a standing lower bound (never as a terminator), erroring on shortfall the way `changelog.rs::finish` now does. |
| MEDIUM | `crates/trusty-git-analytics/src/commands/jira/mod.rs` | 346 | The `continue` on a failed changelog repair also skips the ticket's **comment** fetch. Comments carry no "reads as complete" hazard, so the stated rationale does not cover them — this is collateral the doc does not acknowledge. | Fetch comments before the repair, or replace `continue` with a `repair_failed` flag so the comment read still happens; guard `record_failure` so the ticket is not double-counted against the breaker. |
| MEDIUM | `crates/trusty-git-analytics/src/collect/errors.rs` / `.../jira/model.rs` | 19, 58 | `CollectError` gains two variants and `ChangelogIssue` gains a public field. Both are `pub`, neither is `#[non_exhaustive]`, in a `publish = true` crate at 2.9.4 — a downstream exhaustive `match` or struct literal fails to compile on upgrade. | Add `#[non_exhaustive]` to `CollectError` (errors.rs:19) and `ChangelogIssue` (model.rs:58) in this PR, so this and every future addition is non-breaking. Nothing in-tree breaks today (see Notes). |
| MEDIUM | `crates/trusty-git-analytics/src/collect/jira/changelog.rs` + `commands/jira/mod.rs` | 98, 320 | A permanently unrepairable ticket — search `changelog.total` exceeding what the dedicated endpoint can ever return — pins the cursor and makes **every** subsequent run exit non-zero, with no operator escape hatch. Probed: `Err("incomplete JIRA changelog for PR-E: JIRA reported 5 history entries but only 2 could be retrieved")`, and that repeats forever. | A `--allow-incomplete-changelog` opt-in (writing the short history *and* saying so), or a documented runbook step. Symmetric with the pre-existing comment-failure design, so this is a gap inherited rather than introduced — but the changelog path makes it reachable on a data condition rather than only a transport one. |
| LOW | `crates/trusty-git-analytics/src/collect/jira/jql_time.rs` | 119 | Round-1 LOW, unaddressed: `for _ in 0..MAX_STEP_BACK_MINUTES` tests offsets `0..=1559`, while the doc (74-84) and the error message (130) both claim 1560. | `0..=MAX_STEP_BACK_MINUTES`, or state 1559 in the doc and message. |

The other round-1 LOW (the shadowed `truncated` binding) is gone — the shadowing site no longer exists.

## The two judgment calls

### 1. Not persisting the embedded transitions on a failed repair — **correct, but the stated reason isn't the one that carries it**

The rationale given is "once written it reads exactly like a complete history." That argument is *symmetric* and therefore doesn't actually decide the question: zero rows also reads exactly like "this ticket never transitioned," and is equally silent to any downstream query. The argument that does decide it is different: the oldest transition is what anchors cycle-time and lead-time, so a truncated history yields a confidently **wrong number**, whereas absence yields a **missing row** that aggregates naturally exclude. Wrong data is worse than absent data. The call is right; consider restating the doc on that footing, because the current wording invites exactly the counter-argument above.

**"The run bails and the cursor holds" — verified true on every path.** I enumerated every exit from the per-ticket loop (`mod.rs:307-400`):

- repair `Err` → `record_failure` → `break` (breaker) or `continue`; `write_transitions` is never reached, and `plan_cursor` returns `Advance(earliest_failure)` — the failed ticket's own `updated` — or `Hold` outright when that ticket has no `updated` (`sync.rs:168-170`).
- `write_transitions(...)?` and `upsert_comment_detail(...)?` both propagate out of `run_sync` *before* the `if !args.dry_run { plan_cursor … }` block at 412, so a DB error cannot advance the cursor either.
- dry-run writes no cursor at all.

No path bails and advances. The e2e test pins the exact clamp value (`2026-01-03T10:00:00+00:00`, the failed ticket's own timestamp), and mutating the repair-failure arm back to a propagating `?` drops the healthy ticket's rows to `left: 0, right: 1` — so the isolation is genuinely load-bearing, not decorative.

The one thing the asymmetry costs that the doc does not acknowledge is the ticket's comments — MEDIUM-2 above.

### 2. The 200-page cap — **correct, and it loses nothing**

The cap only converts a non-terminating walk into a named error; it can never truncate a walk that would otherwise have succeeded. Boundary checked in both directions:

- `fetch_changelog`: a history of exactly 20,000 entries terminates on iteration 200 via `start_at >= total` **before** the budget is exhausted. Only >20,000 errors.
- `fetch_comments`: needs one extra probe page to discover the end, so its effective ceiling is 19,900 comments (20,000 if the final page happens to be short). Both are orders of magnitude beyond any real issue.

And it fails through the same per-ticket isolation, so a runaway is a bounded named failure that can never be mistaken for a complete result — which is the property the rest of this PR is about. The scope creep is justified: the page-until-empty branch introduced *this round* is genuinely unbounded without it, so shipping the branch without the cap would have traded one liveness hole for another.

## Mutation verification — I reproduced all three, not asserted them

| Mutation applied | Test | Observed |
|---|---|---|
| `let ended = n < COMMENT_PAGE_SIZE` (compare against the size requested) | `fetch_comments_honours_the_page_size_the_server_applied` | `left: 50 / right: 150` |
| `start_at >= reported_total.unwrap_or(0)` + `retrieved < unwrap_or(0)` (i.e. `#[serde(default)] u64`) | `fetch_changelog_keeps_paging_when_the_endpoint_omits_total` | `left: 2 / right: 3` |
| `let mut reported_total = None` (drop the search-count seed) | `fetch_changelog_errors_when_an_absent_total_hides_a_shortfall` | returns `Ok([1 transition])` instead of erroring |
| repair failure `return Err(e.into())` (round-1 semantics) | `a_broken_changelog_repair_does_not_take_the_whole_run_down` | `left: 0 / right: 1` rows for the *healthy* ticket |

Every claim in the PR body matches what I observed byte-for-byte.

## My own probes against the round-2 code

`fetch_comments`, five hostile servers, 150-250 comments each:

```
PROBE A (echoes maxResults 1000 > the 100 requested):  Ok(250) in 3 calls   PASS — clamp holds
PROBE B (echoes maxResults 0):                         Ok(120) in 4 calls   PASS — the `.filter(|m| *m > 0)` holds
PROBE C (page 1 echoes 50, later pages omit the field): Ok(120) in 4 calls  PASS — mixed omission degrades to page-until-empty
PROBE D (echoes 100 but actually pages 50):            Ok(50)  in 1 call    ← MEDIUM-1
PROBE E (search says 5, endpoint has 2 and says total:2): Err(IncompleteChangelog) ← MEDIUM-4
```

A through C are the three cases the new rule had to get right, and it gets all three right. D is the residual: the rule is only as truthful as the echoed `maxResults`, and unlike the changelog sibling there is no second signal to catch a server that lies about it. I am **not** claiming JIRA does this — that is why it is MEDIUM and not HIGH — but the corroborating field is sitting unread in the same response, and this PR's own thesis is that a short result must never read as complete.

## Green honesty — clean

`cargo test -p tga --all-targets` at `5a7c9d16`:

```
test result: ok. 763 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 3.72s
test result: ok. 151 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.77s
```

The single unit-test ignore is `core::effort::tests::effort_cross_validate`, pre-existing. `--doc` adds `5 passed; 2 ignored`; both ignored doctests are in `core/config`, untouched by this PR — the body's "one pre-existing ignore" is accurate for the unit suite but silently omits these two. All 12 named tests appear as `... ok` (verified by name). No new `#[ignore]`, no cfg-gating, no `--lib` narrowing, no `--exclude`. `cargo fmt --check` exit 0; `cargo clippy -p tga --all-targets` produces no lint from this crate (the one "warning" line is the repo-wide duplicate-build-target Cargo notice); `check_line_cap.sh` 7 allowlisted / 0 violations; `check_test_pointers.sh` 0 dangling. CI 22/22 green.

## Notes

**The shared-machinery claim is real, not a lookalike.** `record_failure` (`mod.rs:519`) is one function called from both failure sites (336 and 388), pushing to the same `failed_tickets` / `failed_updated` / `consecutive_failures`. The cursor clamp is `plan_cursor`, untouched by this round. The breaker rename is semantically inert: same value 10, same `>=`, same single reset point (`consecutive_failures = 0`, comment-success arm only), so the comment path's threshold semantics are unchanged. The only delta is that the abort `warn!` moved out of the two break arms into one post-loop `if tripped` — same variables, same values, no behaviour change.

**`CollectError` match sites, checked exhaustively.** Three in-tree matches, all with a wildcard arm: `pm_adapter/factory.rs:41` (`other =>`), `retry.rs:170` (`_ => false`), `commands/jira/mod.rs:225` (`other =>`). Nothing in-tree breaks. `is_retryable` correctly classifies `PagingBudgetExceeded` as non-retryable, and it is moot anyway — both new variants are constructed *outside* `with_retry`, which wraps only the individual `self.get(&url)`. No workspace crate depends on `tga`; the SemVer exposure is external-only, which is why MEDIUM-3 is about hygiene rather than a live break. Note the precedent: `Throttled` already landed on `main` via #4067 under the same conditions, so this is a standing pattern worth closing rather than a new lapse.

**Replacement-not-merge still holds after the move.** `expected` is `changelog.total` (a count of *raw history entries*), and `finish` compares it against `histories.len()` — also raw entries — not against the status-filtered transition count. The two are consistently dimensioned. Since the embedded copy is the *newest* N entries and the dedicated walk returns entries `0..total`, the repaired set is a strict superset, so replacement can only ever grow the transition set. Combined with the upsert writer, a prior run's full history cannot be regressed by a later run that fails to detect truncation.

**The `changelog.total`-absent case is still a silent write, and that is a deliberate residual.** When the search omits `changelog.total`, `embedded_changelog_is_truncated` returns `false`, so `truncated_history_total` is `None`, no repair runs, and a possibly-short embedded history is persisted behind nothing louder than a `warn!` — with the cursor advancing. That is philosophically in tension with MEDIUM-4's "never persist a knowingly-short history," but the distinction is defensible: one is *known* short, the other is *unverifiable*, and verifying it costs a round trip on every ticket. Flagging it here only so nobody later reads the PR as a claim that truncation is now impossible — it is impossible only when JIRA states the count.

**Round-1 process note discharged.** The stale PR description is fixed; the body now describes the actual diff including `fetch_comments` and `jql_time`.

PR #4161, opened 2026-07-27
fix(trusty-agents): taint delegate_to_agent spawns to close injection-to-RCE path

### LINKED ISSUE #4126: prompt injection via live Gmail/Drive content reaches arbitrary code execution through delegate_to_agent to an unrestricted engineer
## The chain

1. Assistant-tier personas that ingest external content (`assistant`, `cto-assistant`, `izzie`) are granted live content-fetch tools in `[tools].allow`: `get_gmail_message_content`, `search_gmail_messages`, `get_drive_file_content`, `web_search`, `okg_ingest_gmail`, `okg_ingest_drive` — `crates/trusty-agents/.trusty-agents/agents/assistant/agent.toml:73-106`.
2. The three-layer untrusted-data neutralization envelope (`crates/trusty-agents/src/ctrl/pm_task/dispatch/persona_memory.rs:420-515`) wraps ONLY persisted memory-drawer recall rendered into the system prompt. It does NOT apply to the raw return value of a live tool call in the same turn. `get_gmail_message_content`'s output enters the conversation unneutralized. The envelope's own doc comment (lines 426-435) describes exactly this attack shape, but its fix is scoped to drawers.
3. The same allowlist grants `delegate_to_agent` — `assistant/agent.toml:79`.
4. Delegation is role-gated to `ASSISTANT_ALLOWED_DELEGATE_ROLES` (`crates/trusty-agents/src/runtime/tool_registry.rs:120-127`), which checks the TARGET'S ROLE NAME ONLY. It cannot inspect the content of the `task` string handed to that target.
5. `crates/trusty-agents/.trusty-agents/agents/engineer.toml:1-6` declares `role = "engineer"`, `runner = "claude-code"`, and NO `[tools].allow`/`[tools].allowed` whatsoever.
6. `scope_assistant_allowed_tools` (`crates/trusty-agents/src/runtime/tool_registry.rs:529-553`) — the only gate that would narrow a spawned worker's tools to the delegator's restricted surface — returns early unless the SPAWNED agent's own role is `"assistant"`: `if !is_assistant_tier || existing_allowed.is_some() { return existing_allowed; }`. `engineer`'s role is `"engineer"`, so this gate is a no-op for it regardless of who delegated.
7. With no allowlist, `dispatch_gated`'s `None` semantics make every registered tool callable (`crates/trusty-agents/src/tools/mod.rs:149-164`; see also the explicit comment at `ctrl/pm_task/dispatch/persona.rs:121-127`). Running under `runner = "claude-code"`, the spawned engineer is an actual `claude` CLI subprocess with full local shell and file-write capability.

## Impact

Hostile text embedded in an inbound email or a shared Drive document, once fetched during a cto-assistant or izzie turn, can cause that persona to call `delegate_to_agent(agent_name="engineer", task="<attacker-controlled text>")` — a call entirely within its declared allowlist — and the spawned engineer subprocess executes that text as its actual task with unrestricted write and shell access. The daemon runs with the user's full privileges, holds working AWS credentials resolved silently from `~/.aws`, and holds the roster's Google and Slack OAuth grants.

This is exploitable TODAY. It requires no skill mechanism and is unrelated to the executable-skill ban.

## Existing partial mitigation is insufficient

The codebase already names this risk verbatim at `assistant/agent.toml:210-224`: "this same persona ingests attacker-controllable text in the SAME turn that holds send/share credentials... the allow-list bounds WHICH tools exist, not what an injected model chooses to do with them." The named mitigation, issue #4054 (a confirmation gate), covers five tools — `compose_email`, `manage_file_permissions`, `manage_gmail_settings`, `manage_gmail_filters`, `modify_gmail_messages` — is NOT YET BUILT, and critically does NOT cover `delegate_to_agent`.

## Suggested fix (two options, either closes it)

- Propagate a taint flag through `delegate_to_agent` when the delegator's turn touched a live external-content tool (or unconditionally for every assistant-tier delegation), and make the spawned worker's registry apply the SAME allow/deny posture the delegator had, regardless of the worker's own role — i.e. make `scope_assistant_allowed_tools`'s narrowing travel with a tainted delegation rather than gating on `is_assistant_tier`. The scoping function and role gate both already exist; this closes the gap between them.
- Or extend #4054's confirmation gate to cover `delegate_to_agent` itself when a live-fetch tool fired earlier in the same turn. Smaller change.

Also worth closing separately: the neutralization envelope should cover live external-content tool returns, not only persisted drawers.

Reference #4054, #4111 (the daemon emits no logs, so exploitation would leave no trace), and #4080 (the executable-skill decision this review was originally commissioned for).

bobmatnyc:
## Summary

Closes #4126 — a live, exploitable prompt-injection-to-code-execution path through `delegate_to_agent`.

**The vulnerability (independently confirmed against `origin/main` @ `4e36002d`):**

1. Assistant-tier personas (`assistant`, `cto-assistant`, `izzie`) are granted live external-content tools (`get_gmail_message_content`, `web_search`, etc.) — `crates/trusty-agents/.trusty-agents/agents/assistant/agent.toml:73-106`.
2. The untrusted-data neutralization envelope (`crates/trusty-agents/src/ctrl/pm_task/dispatch/persona_memory.rs:420-515`) wraps only persisted memory-drawer recall — not live tool return values.
3. The same personas are granted `delegate_to_agent`.
4. Delegation is role-gated by `ASSISTANT_ALLOWED_DELEGATE_ROLES` (`tool_registry.rs:120-127`), which checks only the *target's role name*, never the content of the `task` string.
5. `engineer.toml` declares `role = "engineer"`, `runner = "claude-code"`, and no `[tools].allow`.
6. `scope_assistant_allowed_tools` (`tool_registry.rs:529-553`) — the gate that would narrow a spawned worker — only fired when the **spawned** agent's own role was `"assistant"`. For `engineer` this was a no-op.
7. With no allowlist, `dispatch_gated`'s `None` semantics make every registered tool callable — under `runner = "claude-code"` that's a real `claude` subprocess with full shell and file-write access.

Net: attacker text fetched during a persona turn (an inbound email, a scraped page) could drive `delegate_to_agent(agent_name="engineer", task="<attacker text>")` and the spawned engineer would execute it with unrestricted access.

## The fix

`scope_assistant_allowed_tools`'s narrowing now travels *with the delegation* rather than gating on the spawned agent's own role. Every assistant-tier `delegate_to_agent` call tags its `SubprocessAgentRunner` with the delegator's own `[tools].allow` posture (`with_delegation_taint`), forwarded to the child process as `TAGENT_DELEGATION_TAINT_ALLOW` (JSON). `run_subagent` reads this back and applies the same narrowing using the *delegator's* patterns, regardless of the spawned agent's declared role — closing the gap on both dispatch paths:

- the `--direct`/`--agent` subprocess path (`runtime/tool_registry.rs` + `runtime/subagent_mode.rs`)
- the in-process `/agent` persona-chat path (`ctrl/pm_task/dispatch/persona.rs`)

**Taint scope decision: unconditional**, applied to every assistant-tier delegation rather than only turns that happened to call a live-fetch tool. Chosen because (a) it's simpler and fails safe, (b) per-turn tracking would still miss content fetched in an *earlier* turn that's still present in conversation history, and (c) assistant personas' core value proposition is live content access, so a conditional gate would rarely differ from unconditional in practice while adding real implementation risk.

**Observability**: narrowing is logged at `info` on both the parent (pre-spawn, in `spawn_and_run_inner`) and child (post-scoping, in `run_subagent`) sides. A disallowed tool call already returns (and continues to return) a legible `"Tool 'X' is not permitted for this agent. Allowed: ..."` error via `dispatch_gated` — not a silently-missing tool. A malformed/corrupted taint env var fails **closed** (empty/deny-all), never silently untainted.

A taint also never widens past an agent's own explicit `[tools].allowed` when one is set — the two are intersected.

## Diff

- `crates/trusty-agents/src/subprocess/mod.rs` / `spawn.rs` — new `SubprocessAgentRunner::with_delegation_taint`, forwarded as `TAGENT_DELEGATION_TAINT_ALLOW` env var, logged at spawn time.
- `crates/trusty-agents/src/runtime/tool_registry.rs` — `build_assistant_tier_registry` now tags its own `delegate_to_agent` tool with the current agent's `[tools].allow`; new pure, unit-tested `scope_for_delegation` and `parse_delegation_taint_env` helpers.
- `crates/trusty-agents/src/runtime/subagent_mode.rs` — reads the taint env var, applies `scope_for_delegation` instead of the old unconditional `scope_assistant_allowed_tools` call.
- `crates/trusty-agents/src/ctrl/pm_task/dispatch/persona.rs` — same taint wiring for the in-process persona-chat dispatch path.
- `crates/trusty-agents/src/runtime/tool_registry_tests.rs` — 8 new tests (see below).
- `crates/trusty-agents/CHANGELOG.md` — Security entry.

## Tests

New, all passing:
- `tainted_delegation_cannot_reach_tool_delegator_lacked` — a tainted `local-ops-agent`-shaped registry (real `shell_exec`) cannot reach `shell_exec`/`read_file`/`list_dir`/`grep_files` when the taint patterns don't cover them.
- `untainted_delegation_is_unaffected` — an untainted spawn stays fully unrestricted (byte-identical to pre-fix).
- `tainted_delegation_intersects_with_existing_allowed` — taint narrows, never widens past an agent's own `[tools].allowed`.
- `assistant_delegate_to_engineer_path_is_scoped` — the exact `assistant → delegate_to_agent(engineer) → unrestricted tools` path, closed.
- `parse_delegation_taint_env_{absent_is_untainted,valid_json_is_tainted,empty_array_is_tainted_deny_all,malformed_fails_closed}` — env-var parsing, including the fail-closed guarantee.

## Quality gate (raw output)

```
$ cargo fmt --check
(clean, exit 0)

$ cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
(exit 0; only pre-existing unrelated `tga` crate warnings, not gated by -D warnings)

$ SKIP_UI_BUILD=1 cargo test --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui --no-fail-fast
TOTAL passed=19320 failed=5
```

The 5 failures are two pre-existing environmental flakes, independently reproduced identically on a clean `origin/main` checkout with zero changes:
- `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json` (1 failure) — a `uv` cache-lock race under parallel test execution; passes standalone with or without this diff.
- `tests/api_e2e.rs` (4 failures) — spawned API server doesn't become healthy within the test's 5s timeout in this sandboxed environment; reproduces identically on unmodified `main`.

Neither touches any file in this diff.

## Coordination

Stayed scoped to the tool-registry/delegation area per instructions; did not touch `~/.trusty-agents/`, `.bundled-stamp`, or run a reprovision.

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -7,6 +7,48 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Security
+
+- **Closed a prompt-injection-to-arbitrary-code-execution path through
+  `delegate_to_agent` (#4126).** An assistant-tier persona (`assistant`,
+  `cto-assistant`, `izzie`) holding live external-content tools
+  (`get_gmail_message_content`, `web_search`, …) plus `delegate_to_agent`
+  could be driven by attacker-controlled fetched content to delegate an
+  attacker-authored task to `engineer` — a worker role with no
+  `[tools].allow` at all — and the spawned `claude-code` subprocess got its
+  full unrestricted tool surface (shell, file write) with no narrowing,
+  because the existing `scope_assistant_allowed_tools` gate only narrowed a
+  spawn when the SPAWNED agent's own role was `"assistant"`. Every
+  assistant-tier `delegate_to_agent` call now tags its spawn with the
+  delegator's own `[tools].allow` posture (`SubprocessAgentRunner::
+  with_delegation_taint`); the spawned worker's registry is narrowed to that
+  posture regardless of its own declared role, on both dispatch paths (the
+  `--direct`/`--agent` subprocess path and the in-process `/agent`
+  persona-chat path). A malformed/corrupted taint value fails CLOSED
+  (deny-all), never silently untainted. Narrowing is logged on both the
+  parent (pre-spawn) and child (post-scoping) sides, and a disallowed tool
+  call already returned (and continues to return) a legible
+  "not permitted for this agent" error rather than a silently-missing tool.
+
+- **Follow-up hardening for the `delegate_to_agent` taint fix above (#4126):
+  closed a multi-hop composition gap, added a fail-closed default, and
+  covered a third, previously-unpatched dispatch path.** (1) At hop 2+
+  (`assistant -> izzie -> engineer`), both dispatch paths forwarded the
+  INTERMEDIATE agent's own native `[tools].allow` as the outbound taint
+  instead of intersecting it with the INBOUND taint it had itself received —
+  a taint could widen back out instead of only ever narrowing. A new
+  `compose_delegation_taint` helper is now the single place both paths
+  compute the outbound taint, always as an intersection. (2) An
+  assistant-tier agent reached with no taint AND no resolvable
+  `[tools].allow` previously fell through to fully unrestricted; it is now
+  denied every tool by default. (3) `run_pm_task_with_history` — the path
+  backing the REPL, Slack, Telegram, and the API server, and `ctrl.toml`
+  (the standalone default) declares `role = "assistant"` per #3812 — built
+  its `DelegateToAgentTool` with no taint and no target-role gate at all,
+  reproducing #4126's exact exploit shape on the default path; it now
+  applies the same taint and `ASSISTANT_ALLOWED_DELEGATE_ROLES` gate as the
+  other two dispatch paths (`pm`/orchestrator-role callers are unaffected).
+
 ### Added
 
 - **The token-streaming chat path now streams Bedrock models via
--- a/crates/trusty-agents/src/ctrl/pm_task/dispatch/history.rs
+++ b/crates/trusty-agents/src/ctrl/pm_task/dispatch/history.rs
@@ -380,18 +380,41 @@ pub async fn run_pm_task_with_history(
         return Ok(content);
     }
 
-    let runner: Arc<dyn AgentRunner> =
-        Arc::new(SubprocessAgentRunner::new().with_config_dir(Some(config_dir.clone())));
+    // Security fix (delegate_to_agent injection-to-RCE path, item 3, #4126):
+    // this THIRD dispatch path (backing the REPL, Slack, Telegram, and the
+    // API server, and `ctrl.toml` is the standalone-mode default) built its
+    // `DelegateToAgentTool` with NO taint and NO `allowed_target_roles` at
+    // all — `pm_cfg` resolves to `ctrl.toml`, which declares
+    // `role = "assistant"` (deliberately, per #3812, for role-filtered
+    // picker grouping), the SAME role value the OTHER two dispatch paths
+    // (`runtime::subagent_mode`, `persona.rs`) treat as "holds live
+    // external-content tools (web_search here), must never produce an
+    // unrestricted delegate_to_agent spawn." See `ctrl_delegate_posture`.
+    let inbound_taint = crate::runtime::tool_registry::parse_delegation_taint_env(
+        std::env::var("TAGENT_DELEGATION_TAINT_ALLOW").ok(),
+    );
+    let (delegation_taint, allowed_target_roles) = ctrl_delegate_posture(
+        &pm_cfg.agent.role,
+        pm_cfg.tools.allow.as_deref(),
+        inbound_taint.as_deref(),
+    );
+    let runner: Arc<dyn AgentRunner> = Arc::new(
+        SubprocessAgentRunner::new()
+            .with_config_dir(Some(config_dir.clone()))
+            .with_delegation_taint(delegation_taint),
+    );
 
     let mut registry = ToolRegistry::new();
-    registry.register(Arc::new(
-        // #3737: thread the task's session id so a successful delegation emits
-        // `AgentSpawned` and the API server can attribute the answer to the
-        // specialist that produced it (chat-bubble responder attribution).
-        DelegateToAgentTool::new(runner)
-            .with_config_dir(config_dir.clone())
-            .with_session_id(sid.clone()),
-    ));
+    // #3737: thread the task's session id so a successful delegation emits
+    // `AgentSpawned` and the API server can attribute the answer to the
+    // specialist that produced it (chat-bubble responder attribution).
+    let mut delegate_tool = DelegateToAgentTool::new(runner)
+        .with_config_dir(config_dir.clone())
+        .with_session_id(sid.clone());
+    if let Some(roles) = allowed_target_roles {
+        delegate_tool = delegate_tool.with_allowed_target_roles(roles);
+    }
+    registry.register(Arc::new(delegate_tool));
     // #3052 PR B (lane 3): the opaque tm<->tcode bridge, distinct from lane 2
     // (`delegate_to_agent` above). RBAC-locked to deny ReadOnly + Analytics.
     // #4026/#4028: the caller's `[subagents].allowed` list is resolved into the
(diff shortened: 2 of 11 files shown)


bobmatnyc:
## Verdict: APPROVE

Reviewed head `23047bb3e198b4ee23bec9098aef9c5a2c5970bf` (branch `fix-delegation-taint`, rebased onto `main` @ `a94ac69b`). No CRITICAL or HIGH findings at >80% confidence. Verification went beyond reading the diff: built the branch in an isolated worktree and ran the affected test modules against real compiled code, and traced the actual runtime dispatch graph (not just the functions named in the PR body) to confirm `cfg.tools.allowed` genuinely gates the vulnerable path end-to-end.

### Claims verified against the actual diff (not the PR description)

1. **`compose_delegation_taint`** (`runtime/tool_registry.rs:717-737`) — intersection-based composition. Proved algebraically that the result is always a subset of both `native` and `inbound_taint` by construction (`inbound.iter().filter(|p| native_set.contains(p))`), so it cannot widen at any hop, including when inbound is absent (`None` → forwards native unchanged, hop-0 case) or native is absent (`None` → forwards inbound unchanged, "this agent imposes no further narrowing" — correct, not a bypass since it can only ever narrow what's already been received). Wired into `subagent_mode.rs:395-399` (recomputes the *outbound* taint before spawning the next hop) and `persona.rs:305-315` and `history.rs`'s `ctrl_delegate_posture` (144-159) for the third path. All three call sites confirmed present in the diff.

2. **`scope_assistant_allowed_tools`/`scope_for_delegation` fail-closed** — confirmed `Some(Vec::new())` is returned (not `None`) when `allow`/`registry` are absent (tool_registry.rs ~563-569), and traced this value forward through `cfg.tools.allowed` → `run_subagent_with_tools`/`run_subagent_single_shot` (`subagent_exec.rs`) → `chat_with_tools_gated` → `ToolRegistry::dispatch_gated` (`tools/mod.rs:149-160`). `dispatch_gated`'s guard is `if let Some(list) = allowed && !list.iter().any(|a| a == name)` — an empty list makes `.any()` false for every tool name, so `Some(vec![])` genuinely denies every call, not merely "unset." No downstream consumer treats empty as unset.

3. **Third path (`history.rs`)** — `ctrl_delegate_posture(role, native_allow, inbound_taint)` gates on `pm_cfg.agent.role` (the delegator's own role, correct direction — the original bug was gating on the *target's* role) and returns `(None, None)` for `role != "assistant"` (pm/orchestrator unaffected) or `(Some(taint), Some(ASSISTANT_ALLOWED_DELEGATE_ROLES))` for `role == "assistant"`. Confirmed `ctrl.toml` (`crates/trusty-agents/.trusty-agents/agents/ctrl.toml`) declares `role = "assistant"` and no `[tools]` section at all, so ctrl's own delegate_to_agent spawns now fail-closed to deny-all by default (matches the `ctrl_delegate_posture_assistant_role_fails_closed_without_allow` test exactly). `ASSISTANT_ALLOWED_DELEGATE_ROLES`/`parse_delegation_taint_env` visibility widened `pub(super)` → `pub(crate)`, confirmed in the diff and needed for `ctrl`'s module tree to reach them.

4. **`apply_delegation_taint_env`** extracted from `spawn_and_run_inner` (`subprocess/spawn.rs:142-198`), unit-tested against a real `tokio::process::Command` via `Command::as_std().get_envs()` (`subprocess/tests.rs`). The two previously-nonexistent doc-cited tests now exist verbatim (`delegation_taint_allow_env_set_when_configured`, `delegation_taint_allow_env_absent_by_default`) plus a third (`..._empty_vec_still_sets_deny_all`). Doc comment at `subprocess/mod.rs:168-172`/`subprocess/spawn.rs` cites these correctly.

### Path-coverage enumeration (the thing the last round missed)

Grepped every `SubprocessAgentRunner::new()` and `DelegateToAgentTool::new()` construction site in the crate:

| Site | Registers `DelegateToAgentTool`? | Taint wired? |
|---|---|---|
| `runtime/tool_registry.rs::build_assistant_tier_registry` | yes | yes (fixed) |
| `ctrl/pm_task/dispatch/persona.rs` | yes | yes (fixed) |
| `ctrl/pm_task/dispatch/history.rs` | yes | yes (fixed, this round) |
| `runtime/pm_mode.rs::run_pm()` | yes | **no** |
| `runtime/workflow_mode/{mod.rs,resume.rs}`, `runtime/postmortem.rs` | no (direct `runner.run(name, task)`, not exposed as an LLM tool call) | n/a, correctly out of scope |

`run_pm()` is the untainted fourth site. Verified it is not a regression introduced by this PR: `pm.toml` declares `role = "orchestrator"` (never matches `role == "assistant"` anywhere in the taint code) and only grants `[tools].scopes = ["memory.read","memory.write","ticketing.read"]` — no live external-content tool (no `web_search`/Gmail/Drive). This matches the exact scope boundary the fix documents everywhere else (assistant-tier only). Flagging as a **Note**, not a finding: if a future project-level `pm.toml` override ever declares `role = "assistant"` *or* grants a live-fetch tool while keeping `role = "orchestrator"`, this site would need the same treatment.

### Deep trace: does the fix actually restrict the named exploit path?

Traced execution rather than trusting the PR body's "claude subprocess" framing. `delegate_to_agent` always goes through `SubprocessAgentRunner` → spawns `current_exe --agent <name>` → `mode_dispatch.rs`'s `--agent` branch is **unconditional**, routing to `subagent_mode::run_subagent`, regardless of the target's `[agent].runner` value. `run_subagent` never constructs a `ClaudeCodeAgentRunner` (grepped the whole file — only imported, never invoked); the literal `claude` CLI subprocess (`agents/claude_code_runner/run.rs`, which really does pass `--dangerously-skip-permissions` and only restricts via the *separate*, untouched `cfg.llm.claude_allowed_tools` field) is reachable only from `workflow_mode`/`direct_mode`/CLI-operator paths via `DispatchingAgentRunner`, never from `delegate_to_agent`. So `cfg.tools.allowed` (what this PR scopes) is in fact the control that governs the real attack surface for this tool, confirmed by:
- `engineer.toml` itself, matched by literal name in `build_registry_for_agent`'s `match name {...}`, actually falls into the generic `_` arm (skills + phase-audit only) — it does not register `shell_exec`/`write_file` through this path.
- `code-agent` (name, `role = "engineer"`) *does* register real `write_file`, and `local-ops-agent` (name, `role = "ops"`) / `qa-agent` (name, `role = "qa"`) *do* register a real `ShellExecTool`. All three roles are members of `ASSISTANT_ALLOWED_DELEGATE_ROLES`. Pre-fix, none of these targets' own role is `"assistant"`, so `scope_assistant_allowed_tools` never touched them — the vulnerability is real via these names, and the fix (which scopes by delegator posture regardless of target role) closes it for all of them, not just the illustrative "engineer" example in the issue.

### Test execution (not just existence)

Built branch `23047bb3` in an isolated worktree (`cargo test -p trusty-agents --lib`, not just `cargo check`):
```
runtime::tool_registry: 35 passed; 0 failed; 1 ignored (pre-existing #[ignore] live-wiring test, unrelated)
ctrl::pm_task::dispatch::history: 12 passed; 0 failed
subprocess::tests: 5 passed; 0 failed
```
Every test name cited in the touched doc comments (`compose_delegation_taint_*`, `multi_hop_delegation_narrows_at_every_hop`, `ctrl_delegate_posture_*`, `parse_delegation_taint_env_*`, `scope_assistant_allowed_tools_fails_closed_*`, `scope_for_delegation_*`, `tainted_delegation_*`, `delegation_taint_allow_env_*`) exists and passes — no doc-pointer inaccuracy recurred.

## Findings

None at >80% confidence for CRITICAL/HIGH. Two LOW-severity notes below (informational, not blocking):

| LOW | `crates/trusty-agents/src/runtime/tool_registry.rs` | 717-737 | `compose_delegation_taint` intersects taint *pattern strings* by exact equality rather than glob-aware semantic equivalence. Proven safe (only ever over-narrows, never widens) given current TOML content (e.g. `izzie`'s `granola_*` wildcard), but two hops expressing overlapping capability via differently-shaped glob strings would silently drop that capability from the composed taint rather than widen it. | Optional: resolve patterns against a canonical tool-name set before intersecting, or leave as-is and note the limitation in the doc comment (it currently doesn't mention the string-vs-glob subtlety). |
| LOW | `crates/trusty-agents/src/runtime/tool_registry_tests.rs` | ~1094-1140 (`tainted_delegation_cannot_reach_tool_delegator_lacked`, `assistant_delegate_to_engineer_path_is_scoped`) | Fixtures use a synthetic `("local-ops-agent", role="engineer")` pairing that doesn't correspond to any real TOML (`local-ops-agent` really declares `role = "ops"`). The real equivalent combo that carries the same real `write_file` danger is `("code-agent", role="engineer")`. Doesn't invalidate the proof (the mechanism under test — `scope_for_delegation` narrowing a genuinely permissive registry when `is_assistant_tier` is false — is exactly what closes the vulnerability) but the comment's framing ("exactly `engineer.toml`'s shape") overstates the fixture's fidelity. | Optional: add one test using the literal `("code-agent", "engineer")` pairing for exact fidelity to the real exploit shape. |

## Notes

- `run_pm()` (`runtime/pm_mode.rs`) remains the one untainted `DelegateToAgentTool` construction site. Correct as shipped (pm.toml is `role = "orchestrator"`, no live-fetch tools), but it is a boundary decision (assistant-tier only) rather than "every construction site," worth keeping in mind if `pm.toml`'s posture ever changes.
- `ctrl.toml` declaring no `[tools].allow` means `ctrl`'s own `delegate_to_agent` now fails closed to deny-all by design (`ctrl_delegate_posture_assistant_role_fails_closed_without_allow`). This is the correct secure default per the fix's stated invariant, but is a functional/operational change worth confirming with the team — any real capability `ctrl` should be able to delegate now requires an explicit `[tools].allow` on `ctrl.toml`.
- Confirmed via `git ls-remote origin`: no stray `pr-4161` branch exists (only `fix-delegation-taint`), and `refs/heads/fix-delegation-taint` is exactly `23047bb3e198b4ee23bec9098aef9c5a2c5970bf`. Housekeeping claim verified.

No issues found at >80% confidence for CRITICAL or HIGH. APPROVED for merge from a security-review standpoint.

Issue #4183, opened 2026-07-27
epic(trusty-mpm): sectioned, versioned PM instructions


bobmatnyc:
## Goal

Ship one validated, sectioned, versioned PM instruction package with explicit customization tiers and measurable behavioral adherence.

## Scope

- Core, Agent Delegation, Workflow, Memory, Search, and base-floor schema;
- deterministic source lift and composition;
- fixed/project/user precedence plus legacy override migration;
- version parsing, stale-install refresh, and diagnostics;
- dynamic resolved-agent roster;
- safe wrap-up guidance for monitors/background tasks;
- robust section parsing for annotated legacy files;
- adherence evaluation that protects behavioral rules from unsafe compression;
- explicit decisions for project/deployed override paths.

## Done when

Composition is deterministic, old/missing versions and overrides migrate visibly, parser behavior is covered by real legacy cases, and instruction changes have adherence measurements.

Issue #384, former PMINSTR leaves #4185–#4191/#4193, and follow-ups #2339, #3927, #4238, and #4247 are consolidated here.

bobmatnyc:
Owner process decision (2026-07-28) is now in scope for this epic's content children — see full detail in #4185 comment: https://github.com/bobmatnyc/trusty-tools/issues/4185#issuecomment-5099535556

Sprint-then-harden workflow doctrine (targeted testing while sprinting to feature-complete, full gates only once harden phase starts; 3+ review rounds = close-and-fold not round 4) must be preserved as an in-scope content requirement even if #4185 is re-scoped or split.

bobmatnyc:
SPRINT phase: drive to feature-complete on a local version; targeted tests, no CI iteration loops, no critic round on narrow changes.
HARDEN phase: after feature-complete — full suite, critic, release gates. Publish only after that.

bobmatnyc:
Child: #4238 — instruction-following adherence eval (measures whether compression/reordering/removal in this epic's sections actually preserves PM/agent behavior).

bobmatnyc:
## Consolidation update (2026-07-28)

This is now the canonical sectioned PM instructions issue. It absorbs #4185, #4186, #4187, #4188, #4189, #4190, #4191, #4193; the body preserves the actionable scope and the closed leaf threads remain available as design history.

bobmatnyc:
Triage consolidation (2026-07-28): this parent now explicitly preserves the scope of the listed older issue while serving as the single outcome-oriented tracker.

[...]

PR #4196, opened 2026-07-27
fix(trusty-mpm): deliver the computed agent roster to the PM prompt (#4069)

### LINKED ISSUE #4069: [PMINSTR-09] dynamically-generated agent roster is computed then discarded — `resolve_pm_prompt` falls back to the static AGENT_DELEGATION asset
**Defect:** The delegation roster is built from the project manifest, carried, and then discarded at the final composition step. The delivered prompt falls back to the static `include_str!` asset instead.

- computed — `generate_authority(&agents)` at `crates/trusty-mpm/src/core/instruction_pipeline.rs:345`
- carried — on the launch struct at `crates/trusty-mpm/src/core/session_launch/mod.rs:527`
- discarded — `resolve_pm_prompt` does `read_override(...).unwrap_or_else(|| AGENT_DELEGATION.trim().to_string())` at `crates/trusty-mpm/src/core/instruction_overrides.rs:166-167`

**Evidence:** 42 deployed agents are represented by a static 63-line file naming 8. In the composed ~79 KB prompt, `ticketing` appears 0 times and `memory-manager` appears 0 times. `AGENT_DELEGATION.md` is unchanged since `c735452c` (2026-07-03).

Surfaced as a symptom: a PM session ran `gh issue`/`gh pr` directly for an extended period because its injected routing table never named `ticketing`. Hand-editing that table is the wrong fix — the table shouldn't be static.

**Resolution:** Dynamic construction from the project manifest (mutable by agents and tools) is the intended behavior, "as before". The fix is localized to that fallback — roughly two lines, not a pipeline rewire.

**This child owns the content diff.** Delivered Agent Delegation goes from 8 agents to 42, with its own expected diff — explicitly NOT inside the byte-identical source swap (#4186).

Blocks the Agent Delegation section in #4185: that section cannot be correct until this lands.

**Acceptance:**
- The delivered prompt's Agent Delegation section is generated from the manifest, not read from the static asset.
- Test asserts the delivered prompt contains agents present in the manifest but absent from the static file (`ticketing`, `memory-manager`).
- Expected before/after diff of the delivered section recorded in the PR.

Parent: #4183

bobmatnyc:
Closes #4069.

## The defect

The delegation roster IS computed at every launch — `build_instructions` calls
`generate_authority(&scan_agents(...))` at `instruction_pipeline.rs:345` — but it
never reaches the delivered prompt. `resolve_pm_prompt` composed the delegation
section as:

```rust
let delegation = read_override(&dir, FILE_AGENT_DELEGATION)
    .unwrap_or_else(|| AGENT_DELEGATION.trim().to_string());
```

That fallback is the static `include_str!` asset, whose agent table has been
hand-maintained and unchanged since `c735452c` (2026-07-03) and names **8**
agents.

Measured against this repo before the fix: a **~40-agent deployment was
advertised as 8**. `ticketing` never appeared as a routable agent, and
`memory-manager` appeared **0 times** in the entire 78,953-byte prompt — so the
PM could not route to either.

## Why regeneration, not plumbing

The pipeline's roster is not threadable to the composer. `build_instructions`
merges `authority` into a single `PipelineOutput.merged` string
(`instruction_pipeline.rs:355`) — there is no separately addressable roster
field. More decisively, `build_system_prompt_for*` and every one of its callers
(`tm session instructions`, the tmux connect path at
`client/http_client/session_connect.rs:82`/`:180`, the daemon spawn) receive only
a `project_dir`, and two of them never run the pipeline at all, so there is no
`PipelineOutput` to thread. Exposing a field would force those call sites to
fabricate a pipeline run.

The roster is therefore regenerated at composition time from the tiers a session
actually loads agents from, which are all derivable from `project_dir`:

1. `<project>/.claude/agents` — the daemon managed-spawn destination, and the
   only tier readable under `--setting-sources project,local`
2. the managed `CLAUDE_CONFIG_DIR/agents` — the standalone `tm run` path
3. `~/.claude/agents` — the plain `tm session start` path

deduped by agent name with project-tier precedence, matching how Claude Code
resolves a same-named agent.

## What is preserved

- **Override precedence is unchanged.** An explicit project/user
  `AGENT_DELEGATION.md` still replaces the whole section, roster included. This
  changes the *fallback* only.
- **No doctrine is lost.** The bundled asset carries routing rules the roster
  does not (make/mise routing, keyword routing, the ops-agent table), so it is
  still emitted in full with the live `## Delegation Authority` block
  **appended**, never substituted. Restructuring the asset is out of scope
  (epic #4183).
- **Unprovisioned environments are unaffected.** With no agent deployed
  anywhere, the asset is returned alone — pre-#4069 behaviour exactly, rather
  than telling the PM it has no agents.

## Before / after

Measured via `build_system_prompt_for` on this repo (same code path the launcher
uses), once with the fix reverted and once with it applied:

| | before | after |
|---|---|---|
| prompt bytes | 78,953 | 88,508 |
| roster entries | 0 | 40 |
| `### ticketing` present | false | true |
| `### memory-manager` present | false | true |
| `memory-manager` occurrences | 0 | 4 |

## Gate

`bundled_delegation_appends_deployed_roster` is the regression gate: it deploys
`ticketing` and `memory-manager` into a tempdir project tier and asserts they
reach the composed prompt. Neither name appears in any bundled asset, so a
rendered `### <name>` entry can only come from the live scan.

Mutation-tested — with `bundled_delegation` reverted to the pre-fix
`AGENT_DELEGATION.trim().to_string()`:

```
running 1 test
test core::instruction_overrides::tests::bundled_delegation_appends_deployed_roster ... FAILED

panicked at crates/trusty-mpm/src/core/instruction_overrides.rs:283:9:
the live roster section must be present

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3731 filtered out
```

With the fix restored, all 15 `instruction_overrides` tests pass. The existing
`agent_delegation_override_replaces` test was extended to deploy an agent and
assert the override still suppresses the roster — precedence is gated, not just
documented.

## Verification

- `cargo check -p trusty-mpm` — `Finished dev profile`, clean
- `cargo clippy --all-targets` — clean; `cargo clippy -p trusty-mpm
  --all-targets -- -D warnings` also clean
- `cargo fmt --all -- --check` — exit 0
- `cargo test -p trusty-mpm` (full, not `--lib`) — `3726 passed; 0 failed;
  6 ignored` plus every integration binary green, 0 failures across all 39
  test binaries
- `scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK.`

`Cargo.lock` untouched.

Added: 260 / Removed: 4 / Net: +256 (of which ~150 are doc comments and tests).


bobmatnyc (commented):
## Verdict: WARN

Zero CRITICAL, two HIGH. The change is a real net improvement (8 advertised → 34 correctly advertised on this repo) and the engineering discipline is above average: I independently re-ran four mutations and all four fail as they should, and the gate test is genuinely hermetic. But the headline claim of the PR — that `memory-manager` is now routable — does **not** hold on the primary launch path, and the roster simultaneously advertises six agents that a daemon-managed session cannot dispatch. Both are cheap to fix.

All measurements below were taken by running this branch's code against this machine, not read off the diff.

## Findings

| HIGH | `crates/trusty-mpm/src/assets/agents/memory-manager.md` (+ `mpm-agent-manager.md`, `mpm-skills-manager.md`) | 4 (`role: base`) | `scan_agents` drops all three via `is_base_role`, so the roster **still** omits `memory-manager` on the daemon path — the exact defect #4069 exists to fix | Set `role:` to the agent's own name; `extends: base-agent` already encodes the inheritance |
| HIGH | `crates/trusty-mpm/src/core/delegation_authority.rs` | 263–274 | The three tiers are unioned unconditionally, but **no** launch mode loads all three — 6 advertised agents are undispatchable under `--setting-sources project,local` | Make the tier set launch-mode-aware, or add a one-line caveat to the rendered roster |
| MEDIUM | `crates/trusty-mpm/src/core/delegation_authority.rs` | 123–125, 318–324 | Unreadable tier is indistinguishable from empty; total enumeration failure silently reverts to the stale 8-name asset | `match` on `read_dir`'s `ErrorKind`; `tracing::warn!` on anything but `NotFound` |
| MEDIUM | `crates/trusty-mpm/src/core/instruction_overrides.rs` | 224–229 | The retained doctrine table now directly contradicts the appended roster on three points | One precedence line between the two blocks |
| MEDIUM | `crates/trusty-mpm/src/core/delegation_authority.rs` | 229–235 | ~2 KB of the 9,555-byte per-session roster is zero-information | Omit `Foundation` for single-element chains; omit `Role` when `role == name` |
| LOW | `crates/trusty-mpm/src/core/delegation_authority.rs` | 285 | Dedup is case-sensitive: `QA` and `qa` in two tiers both survive | Key the `BTreeMap` on the lower-cased name |
| LOW | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | 343–355 | `generate_authority` still runs into `PipelineOutput.merged`, which has no consumer anywhere in the workspace | Delete, or note against #4183 |

---

### HIGH-1 — `memory-manager` is still not in the roster on the daemon path

Three bundled agents declare `role: base` in their own source frontmatter:

```
crates/trusty-mpm/src/assets/agents/memory-manager.md       name: memory-manager     role: base  extends: base-agent
crates/trusty-mpm/src/assets/agents/mpm-agent-manager.md    name: mpm-agent-manager  role: base  extends: base-agent
crates/trusty-mpm/src/assets/agents/mpm-skills-manager.md   name: mpm-skills-manager role: base  extends: base-agent
```

`scan_agents` excludes them at `delegation_authority.rs:159` via `is_base_role`, which matches any role starting with `base`. That is 3 of the 37 bundled deployable agents, silently dropped from every tier they are deployed into. Measured per tier on this machine:

```
project  <worktree>/.claude/agents                 dropped: memory-manager, mpm-agent-manager, mpm-skills-manager
managed  ~/.trusty-tools/.../claude-config/agents  dropped: memory-manager, mpm-agent-manager, mpm-skills-manager
user     ~/.claude/agents                          dropped: mpm-agent-manager, mpm-skills-manager
```

`memory-manager` survives in the user tier **only** because that copy is an older composition carrying no `role:` line at all, so `role` falls back to `name`. Under the daemon managed-spawn path, `--setting-sources project,local` reads the project tier only (this repo's own `core/managed_config.rs:25–38`, "empirically verified against `claude` 2.1.201") — where the file *does* carry `role: base`. So on the primary launch path `memory-manager` is still absent from the roster after this change, and `mpm-agent-manager` / `mpm-skills-manager` are absent on **every** path.

This also means the PR's before/after evidence is machine-dependent: the `memory-manager occurrences 0 → 4` row is produced by a stale user-tier copy on the author's machine, not by the fix. Re-measured on a clean managed workspace the count stays 0.

**Fix** (3 lines, squarely in scope — it is the same defect):

```yaml
# crates/trusty-mpm/src/assets/agents/memory-manager.md
role: memory-manager      # was: base
```

…and the same for `mpm-agent-manager` / `mpm-skills-manager`. `extends: base-agent` already carries the inheritance; `role: base` was never meant to mark them non-delegatable. Then add the regression test that would have caught this — deploy `name: memory-manager / role: base / extends: base-agent` into the project tier and assert `### memory-manager` reaches the prompt. Note that `scan_excludes_base_agents` deliberately covers the "base role, non-base filename" case, so please fix the assets rather than weakening `is_base_role`.

### HIGH-2 — the roster advertises agents the session cannot load

`deployed_agent_dirs` returns all three tiers and `roster_from_dirs` unions them. But per this repo's own `managed_config.rs` audit trail, no launch mode reads all three:

- **daemon managed-spawn** — `--setting-sources project,local` → project tier only; `CLAUDE_CONFIG_DIR/agents` and `~/.claude/agents` are both excluded;
- **standalone `tm run`** — project + `CLAUDE_CONFIG_DIR/agents`; `~/.claude/agents` is not read, because `CLAUDE_CONFIG_DIR` relocates that tier;
- **plain `tm session start`** — project + `~/.claude/agents`; no `CLAUDE_CONFIG_DIR`.

Measured on this repo, the union is 40 agents against a 34-agent project tier. The 6 extras are `copyeditor`, `engineer`, `memory-manager`, `proofreader`, `writer`, `writing-critic` — every one of them undispatchable in a daemon-managed session, where a `Task` call returns "agent type not found" and burns a turn. The doc comment on `deployed_agent_dirs:249–252` states the project tier is "the only tier readable under `--setting-sources project,local`" and then unions the other two anyway.

I accept that the composer only receives `project_dir`. Two workable fixes:

1. **Correct** — thread a small `LaunchMode` (or just the resolved `fw.claude_agents_dir()`, which is already launch-mode-correct at every deploy site) through `build_system_prompt_for_with_style_and_native`, defaulting to project-only. The two call sites cited as unthreadable (`client/http_client/session_connect.rs:82`/`:180`) are the tmux connect path for a *managed* session, so their answer is known statically: project tier.
2. **Cheap mitigation** — keep the union and render one line above the roster: `Some entries may not be loadable in this session's launch mode; if a dispatch reports an unknown agent type, re-route rather than retrying.` Costs ~120 bytes and removes the failure mode entirely.

Either is acceptable; option 1 is the one that makes the doc comment true.

### MEDIUM-1 — silent fail-open when a tier cannot be read

Verified directly:

```
chmod 000 on an agents dir containing 1 valid agent -> scan_agents returns 0 agents, no log line
```

`scan_agents:123` swallows every `read_dir` error, so "directory absent" and "permission denied" are the same observation. If all three tiers fail, `deployed_roster_section` returns `None` and `bundled_delegation` ships the stale 8-name asset — i.e. #4069 silently reproduces under a new trigger with every alarm reading healthy. This is the fail-open shape that has bitten this repo repeatedly. `resolve_pm_prompt` is not on a hot path and the crate already uses `tracing` freely.

```rust
let entries = match std::fs::read_dir(agents_dir) {
    Ok(e) => e,
    Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Vec::new(),
    Err(err) => {
        tracing::warn!(dir = %agents_dir.display(), %err, "agent tier unreadable; roster may be incomplete");
        return Vec::new();
    }
};
```

Plus a `tracing::info!` in `deployed_roster_section` when it returns `None`, so "no agents anywhere" is an observable event rather than an invisible reversion.

### MEDIUM-2 — the two blocks disagree, concretely

The interim append is defensible in principle, but the asset does not merely go stale — it now contradicts the roster sitting directly beneath it:

- `AGENT_DELEGATION.md:31` — "Generic `ops` agent is **DEPRECATED**" while the roster advertises `ops` with its description;
- `AGENT_DELEGATION.md:59` — "ticket", "issue", "PR" → **Version Control**, while the roster advertises `ticketing` (routing to `ticketing` is the motivating example of #4069 itself);
- `AGENT_DELEGATION.md:19` — advertises `mpm-skills-manager`, which the roster omits entirely per HIGH-1.

Reconciling the asset is #4183's job and out of scope, but one line in `bundled_delegation` resolves the ambiguity now:

```rust
Some(roster) => format!(
    "{bundled}\n\n> The live roster below is authoritative for WHICH agents exist; \
     the tables above are routing doctrine only.\n\n{}", roster.trim()),
```

### MEDIUM-3 — ~21% of the roster is zero-information

The rendered section is 9,555 bytes, shipped on every PM session forever. Measured:

- `- **Foundation:** <name>` — 1,243 bytes; the chain is a single self-referential element for **40 of 40** agents, so every one of these lines reads `- **Foundation:** qa` under a `### qa` heading;
- `- **Role:** <name>` — duplicates the heading for **14 of 40** agents (`role` falls back to `name` when unset).

Together ~2 KB of pure noise. `generate_authority:229–235` is unchanged code, but this PR is what puts it in every prompt, so the cost is attributable here:

```rust
if agent.extends_chain.len() > 1 {
    out.push_str(&format!("- **Foundation:** {}\n", agent.extends_chain.join(" → ")));
}
if agent.role != agent.name {
    out.push_str(&format!("- **Role:** {}\n", agent.role));
}
```

### LOW-1 — case-sensitive dedup

Verified: an agent named `QA` in the project tier and `qa` in the user tier both survive `roster_from_dirs` and render as two entries. `by_name` at line 285 keys on the raw name. Low likelihood, one-word fix: key on `agent.name.to_ascii_lowercase()`.

### LOW-2 — the pipeline's roster computation is now provably dead

`instruction_pipeline.rs:343–355` still calls `scan_agents` + `generate_authority` into `PipelineOutput.merged`, and `.merged` has no consumer anywhere in the workspace (grepped). Ironically it uses `fw.claude_agents_dir()`, which *is* the launch-mode-correct tier that HIGH-2 wants. Worth either deleting or harvesting.

## Required Changes

1. Fix `role: base` → the agent's own name in the three bundled agent sources, and add a regression test that a `role: base` + `extends: base-agent` agent reaches the roster. (HIGH-1)
2. Either make the tier set launch-mode-aware or add the "may not be loadable" caveat line to the rendered roster. (HIGH-2)
3. Re-measure the PR's before/after table on a clean workspace (empty `HOME`, empty `CLAUDE_CONFIG_DIR`) after (1), and correct the `memory-manager 0 → 4` row.

MEDIUM/LOW items are tracked, not blocking.

## Notes

**Verification I performed rather than accepted.** All four mutations fail as claimed, including two the PR did not test:

| Mutation | Result |
|---|---|
| A — revert `bundled_delegation` to the static asset | `bundled_delegation_appends_deployed_roster` FAILED — "the live roster section must be present" |
| B — override path re-appends the roster | `agent_delegation_override_replaces` FAILED — "an override replaces the section outright" |
| C — `or_insert` → `insert` (last dir wins) | `roster_from_dirs_dedupes_with_first_dir_winning` FAILED — `left: Some("USER TIER"), right: Some("PROJECT TIER")` |
| D — drop the project tier entirely | 2 tests FAILED, incl. "the agent's own description must reach the prompt" |

Mutation B confirms override precedence is genuinely gated, not merely documented. Mutation D is the interesting one: the `### ticketing` / `### memory-manager` assertions would have passed from this machine's global tiers, and the gate is saved by the fixture-unique `"Handles ticketing work."` assertion. That was the right assertion to write.

**Hermeticity — clean.** I ran the gate test binary directly under `HOME=<empty dir> CLAUDE_CONFIG_DIR=<empty dir>` and it passes, so it does not depend on the developer's `~/.claude/agents`. (My first attempt at this reported a failure; that was a stale binary from mutation D on my side — retracted.)

**Green honesty — clean.** All five new/extended tests are present and report `ok` under a full `cargo test -p trusty-mpm --lib`; no `#[ignore]`, cfg-gate, or `--exclude` was added. The one `ignored` line in the run (`manual_static_verify_full_roster`) is pre-existing and unrelated. 22 CI checks green, `Cargo.lock` untouched, CHANGELOG entry correctly under `## [Unreleased] / ### Fixed`.

**Not flagged.** `scan_agents` skips subdirectories (`is_file()` at line 130), so a directory-package agent would be dropped — I could not establish above 80% confidence that Claude Code discovers nested `.claude/agents/` subdirectories, and no tier on this machine uses that layout, so this is a note only. Separately, `resolve_pm_prompt` now reads process-global `HOME`/`CLAUDE_CONFIG_DIR`, and this test binary already mutates `HOME` via `unsafe set_var` elsewhere; the effect on the new test is additive-only so it cannot flip an assertion — a latent hazard, not a finding.

bobmatnyc:
## Review round 1 addressed — both HIGH landed, all four MEDIUM/LOW taken

Rebased onto current `main` (`f2b79b3a`, includes #4194) before measuring anything.

### HIGH-1 — confirmed, and the review's diagnosis of my evidence was exactly right

Verified per-tier on this machine before changing anything:

```
project  <worktree>/.claude/agents            memory-manager -> role: base
managed  ~/.trusty-tools/.../claude-config/agents  memory-manager -> role: base
user     ~/.claude/agents                      memory-manager -> (no role: line at all)
```

So `memory-manager` reached my union **only** via the stale user-tier copy. On the daemon path (project tier only) it was still absent. The `0 -> 4` row in the previous PR body was reading that stale copy, not the fix. Withdrawn and replaced below.

Fixed at the asset level — `role` is now each agent's own name in all three files; `extends: base-agent` already carried the inheritance.

**On the suggested regression test:** the review asked for a test that deploys `role: base` + `extends: base-agent` and asserts it reaches the prompt, *and* said not to weaken `is_base_role`. Those conflict — that test only passes if `is_base_role` is weakened. I took the second instruction as the binding one and wrote the invariant at the asset level instead: `every_bundled_non_base_agent_reaches_the_roster` asserts every non-`BASE-*` bundled agent survives `scan_agents`. It catches this defect and any future cause of a silent drop, without touching the scanner rule `scan_excludes_base_agents` protects.

### HIGH-2 — took the caveat, and the reason is stronger than "threading is inconvenient"

I mapped all six call sites before deciding. Threading cannot produce a correct answer today:

| call site | deploy `fw` | deployed tier | `--setting-sources project,local`? |
|---|---|---|---|
| `prepare_session*` (mod.rs) | caller's | matches caller | n/a |
| `session.rs:620` display | — | no deploy | n/a |
| `launch.rs:311` (`tm launch`) | `default()` | `~/.claude/agents` | **yes** |
| `launch.rs:490` (`tm connect`) | `default()` | `~/.claude/agents` | **yes** |
| `claude_code.rs:320` (daemon) | `for_managed_workspace` | project tier | yes |
| `session_connect.rs:82/180` | `default()` / none | `~/.claude/agents` / none | no |

Rows 3 and 4 are the blocker: `tm launch` (`launch.rs:208`) and `tm connect` (`launch.rs:467`) deploy via `FrameworkPaths::default()` into `~/.claude/agents`, then spawn with `--setting-sources project,local` (`model_inject.rs:70`) — **the flag excludes the very tier they just populated.** So on those paths:

- pass "the tier the flag reads" -> **empty roster** -> silent revert to the stale asset -> #4069 reintroduced;
- pass "the tier deployed to" -> advertises unloadable agents anyway.

Threading is only correct *after* that deploy/read mismatch is fixed, which is a separate pre-existing defect in two other launch paths and outside this PR. Until then the union plus an explicit re-route instruction is the only shape that cannot silently under-advertise — a failed dispatch is self-correcting and costs one turn; a missing agent is invisible and costs the routing decision entirely.

**Adjacent defect discovered, not filed** (flagging rather than filing per standing guidance): `tm launch` / `tm connect` deploy the agent roster to a tier their own spawn flag excludes. Say the word and I'll open it.

### Corrected before/after — clean room, as requested

Empty `HOME` + empty `CLAUDE_CONFIG_DIR`, bundled roster deployed to the project tier, run as a standalone test binary so no global tier can contribute:

| | before | after |
|---|---|---|
| prompt bytes | 39,932 | 47,957 |
| roster entries | **0** | **37** |
| `### ticketing` | absent | present |
| `### memory-manager` | absent | **present** (was still absent pre-HIGH-1-fix) |
| `### mpm-agent-manager` | absent | **present** |
| `### mpm-skills-manager` | absent | **present** |
| roster section bytes | — | 7,642 (was 9,555 before the MEDIUM-3 trim) |

37 of 37 deployable bundled agents (42 files minus 5 `BASE-*`). Union == project tier == 37 in the clean room; the 6 "extras" the review measured were purely this machine's stale global tiers.

### MEDIUMs / LOW — all taken

- **MEDIUM-1** `scan_agents` now matches on `read_dir`'s `ErrorKind`: `NotFound` stays silent (a tier legitimately absent), anything else warns. `deployed_roster_section` logs when it finds nothing anywhere, so reverting to the bundled asset is observable.
- **MEDIUM-2** A note between the blocks: the roster is authoritative on which agents exist; the tables above are routing doctrine only. Asset reconciliation stays with #4183 — noting as known interim state that the retained table still calls generic `ops` DEPRECATED while the roster lists it. (#4194 already corrected the ticket-routing half in `PM_INSTRUCTIONS.md`.)
- **MEDIUM-3** `Role` duplicating the heading and single-element `Foundation` chains omitted: **9,555 -> 7,642 bytes** per session.
- **LOW-1** Dedup keys on the lower-cased name.
- **LOW-2** `PipelineOutput.merged` left in place — deleting it is #4183's call, noted there.

### Mutation tests — 4 new gates, each verified

| Mutation | Result |
|---|---|
| revert `role:` to `base` in the 3 assets | `every_bundled_non_base_agent_reaches_the_roster` FAILED — `["memory-manager", "mpm-agent-manager", "mpm-skills-manager"]` |
| drop `ROSTER_PRECEDENCE_NOTE` | `bundled_delegation_appends_deployed_roster` FAILED — "the roster must be declared authoritative over the stale doctrine table" |
| re-enable `Role`/`Foundation` unconditionally | `generate_authority_omits_self_referential_lines` FAILED |
| revert dedup key to case-sensitive | `roster_from_dirs_dedup_is_case_insensitive` FAILED — `left: 2, right: 1` |

All four pass restored.

### Gates — raw

- `cargo check -p trusty-mpm` — clean
- `cargo clippy --all-targets` — exit 0; `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — exit 0
- `cargo fmt --all -- --check` — exit 0
- `cargo test -p trusty-mpm` (full, not `--lib`) — exit 0, **39 result lines, 0 FAILED**, `3729 passed; 0 failed; 6 ignored`
- `scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK.`

`Cargo.lock` untouched.

### One thing I could not clear, reported rather than buried

Under heavy concurrent load I twice saw `core::standalone::trust_seed::tests::test_preseed_managed_trust_no_home_write` fail (assertion: a `.claude.json` appeared in its fake `$HOME`). Investigated rather than waved off:

- passes isolated and under `--test-threads=1`;
- **10/10 clean** on this branch and **10/10 clean** on `origin/main` when not competing with other builds;
- `origin/main` independently flaked twice during the same investigation, once in `telegram::supervisor::tests::healthy_run_resets_transient_backoff`;
- my code cannot cause it — `deployed_agent_dirs` only *reads* `HOME`/`CLAUDE_CONFIG_DIR`; creating a file in a fake `$HOME` requires a writer, and the writers (`home_trust_seed`, `preseed_workspace_trust_home`) are pre-existing and untouched here.

This is the process-global `HOME` mutation hazard the review already flagged as latent. I am **not** claiming to have fixed it and not claiming it is definitively unrelated — reporting what I verified.

bobmatnyc:
CI caught one consequence I had missed: `tm-capabilities/references/agents.md` is a generated file that embeds each bundled agent's `role`, so the HIGH-1 asset fix drifted it and the drift gate failed. Regenerated with `tm generate capabilities` (`2b62f2ee`) — the diff is exactly the three corrected rows, `base` -> the agent's own name.

Worth noting the gate did its job: it is the only thing in the repo that would have flagged an agent-catalog edit as incomplete.

**23/23 checks pass** (1 skipping — `Notify on red main`, expected on a green PR). Not merging; back to you for re-review.

bobmatnyc (commented):
## Verdict: APPROVE

Round 2. Both HIGH findings from [round 1](https://github.com/bobmatnyc/trusty-tools/pull/4196#pullrequestreview-4792434175) are discharged, every MEDIUM and the LOW are addressed, and I reproduced the corrected evidence myself in a clean room rather than accepting it. No issues found at >80% confidence. **APPROVED for merge.**

The round-1 evidence was wrong because a union reading on a machine with stale global tiers looks like success. I re-measured against that exact trap and the numbers now hold.

---

### Clean-room re-measurement — reproduced independently

I built a probe against this branch's `build_system_prompt_for_with_style_and_native` under `HOME=<empty tempdir>` and `CLAUDE_CONFIG_DIR=<empty tempdir>`, deploying the bundled roster through the real `deploy_agents` composer into the **project tier only**:

```
A_BEFORE_BYTES=39932          # no agents anywhere -> None branch (pre-fix behaviour)
A_HAS_ROSTER=false
B_DEPLOYED=42 B_SKIPPED=0     # 42 asset files, 5 BASE-* excluded
B_AFTER_BYTES=47957
B_ROSTER_BYTES=7647
B_ROSTER_ENTRIES=37
B_HAS_memory-manager=true
B_HAS_mpm-agent-manager=true
B_HAS_mpm-skills-manager=true
B_HAS_ticketing=true
B_SCAN_PROJECT_TIER=37
B_UNION=37
```

Every headline number matches the PR's claim exactly: **39,932 → 47,957 bytes, 0 → 37 entries.** `B_UNION == B_SCAN_PROJECT_TIER == 37` confirms the stated claim that in a clean room the union equals the project tier — so round-1's "6 extras" were indeed purely this machine's stale global tiers, and the withdrawal of the `0 → 4` row was correct. (Roster section measures 7,647 B against the reported 7,642; a 5-byte section-slicing difference, immaterial.)

### HIGH-1 — fixed, and the deviation is sound

I asked for a test that deploys `role: base` + `extends: base-agent` and asserts it reaches the prompt, *and* said not to weaken `is_base_role`. **The engineer is right that those conflict** — `scan_agents` drops any role starting with `base` at line 178 regardless of tier, so that test can only pass if the scanner is weakened. They treated the no-weakening constraint as binding, which was the correct call: `scan_excludes_base_agents` deliberately covers the base-role/non-base-filename case, and weakening it would have traded one silent-drop defect for a silent-*inclusion* one.

**Does the asset invariant catch the same defect class? I mutated it to find out.** Reintroducing `role: base` on `memory-manager.md`:

```
test core::delegation_authority::tests::every_bundled_non_base_agent_reaches_the_roster ... FAILED

panicked at crates/trusty-mpm/src/core/delegation_authority.rs:622:9:
bundled agents silently dropped from every roster (check for a stray `role: base`
in their frontmatter): ["memory-manager"]
```

It catches it, and names the offending agent in the failure message. This is **broader** than what I asked for, not narrower: I asked for three named agents, this covers all 37 and any future addition. The end-to-end prompt-reachability leg is covered separately by `bundled_delegation_appends_deployed_roster` against fixture agents, so the two tests together span the whole chain. No narrowing.

I verified the test's `expected` set (file stems) and `scanned` set (frontmatter names) agree — all 37 non-`BASE-*` assets have `name == stem`, so the comparison is sound.

**Cross-module check on the asset change.** Changing `role` is not free — `trusty-agents-common/src/agents/builder.rs:766` injects a role-derived `initialPrompt` into composed agents. I checked `default_initial_prompt` (`builder.rs:273`): `"base"` and all three new role values fall through to `_ => None`, so the change is behaviourally neutral there. `resolve_agent_model` (`config.rs`) keys on `agent_name`, not role, so model resolution is unaffected. Confirmed by the generated doc, where only the three role cells moved.

### HIGH-2 — mapping verified independently; the caveat is justified

This was the entire justification for not threading, so I traced it rather than reading the claim:

- **`launch.rs:208`** — `tm launch` constructs `FrameworkPaths::default()`, then `prepare_session_with_repo_url(&fw, &managed_path, …)`.
- **`launch.rs:470`** (cited as `:467`, which lands in the immediately preceding comment block) — `tm connect` does the same via `prepare_session(&fw, &path)`.
- `FrameworkPaths::default()` → `under(dirs::home_dir())` → `claude_agents = <home>/.claude/agents` (`paths.rs:109`/`:134`).
- `prepare_session_inner` deploys to `fw.claude_agents_dir()` (`session_launch/mod.rs:508`) — i.e. `~/.claude/agents`, **not** `<project>/.claude/agents`.
- **`model_inject.rs:70`** — `SETTING_SOURCES_FLAG = "--setting-sources project,local"`, and `build_claude_command` **always** appends it (`:93-94`). `launch.rs:322` spawns through that function.
- The repo's own `managed_config.rs:25-38` (empirically verified against `claude` 2.1.201) confirms that flag "EXCLUDES the `user` tier".

**The claim is exactly right: both paths deploy into the one tier the spawn flag excludes.** So "pass the tier the flag reads" would render an empty roster on `tm launch` / `tm connect` and silently revert them to the stale asset — reintroducing #4069 on two of the three launch paths. Threading is genuinely not viable today, and the union-plus-caveat is the shape that cannot silently under-advertise. This is option 2 of the two I offered in round 1 and I explicitly accepted it there, so HIGH-2 is discharged on its own terms.

I verified the caveat actually ships — dropping the note fails the gate:

```
test core::instruction_overrides::tests::bundled_delegation_appends_deployed_roster ... FAILED
panicked: the roster must be declared authoritative over the stale doctrine table
```

### Four mutation gates — all four verified

| Mutation | Result |
|---|---|
| Asset revert (`role: base` on `memory-manager`) | `every_bundled_non_base_agent_reaches_the_roster` FAILED — `dropped: ["memory-manager"]` |
| Drop `ROSTER_PRECEDENCE_NOTE` | `bundled_delegation_appends_deployed_roster` FAILED — "must be declared authoritative" |
| Revert the rendering trim (unconditional `Role` + `Foundation`) | `generate_authority_omits_self_referential_lines` FAILED — "role duplicating the heading must be omitted" |
| Case-sensitive dedup key (`entry(agent.name.clone())`) | `roster_from_dirs_dedup_is_case_insensitive` FAILED — ``left: 2, right: 1`` |

Each fails as claimed, including the exact `left: 2, right: 1`. All four restored and re-verified green.

### Disclosure 1 — the generated-doc regeneration is complete, not just enough to pass

CI's drift gate only proves the file matches the generator; it does not prove the generator's output is right. I cross-checked every bundled asset's frontmatter against `tm-capabilities/references/agents.md` independently:

```
checked=37 generated_rows=37 mismatches=0
```

All 37 deployable agents present, zero `role` drift, zero `extends`/foundation drift, no extra rows, and no residual `| base |` cell anywhere under `tm-capabilities/`. The regeneration is correct and complete.

### Disclosure 2 — the flake is genuinely pre-existing, not a masked regression

This mattered because the new code *does* read process-global `HOME` (`deployed_agent_dirs` → `FrameworkPaths::default()`), which is the latent hazard I noted in round 1. I checked it rather than taking the 10/10 isolated runs at face value.

**Dispositive:** the test's strategy-2 assertions (`trust_seed_tests.rs:377-386`) are pure *write*-detection — "`fake_home` contains no `.claude.json` and no `.claude/`". Only a **write** can violate them. This PR adds **no** `set_var`, **no** `remove_var`, and no filesystem write outside test tempdirs:

```
$ git diff origin/main...HEAD -- '*.rs' | grep -E '^\+.*(set_var|remove_var|create_dir|fs::write|File::create)'
+        fs::create_dir_all(&dir).expect("create .claude/agents");   # tempdir fixture
+        fs::write(                                                  # tempdir fixture
```

I also confirmed the whole new roster path is read-only — `FrameworkPaths::under`, `locate_trusty_mpm_root`, and `managed_claude_config_dir` are pure path joins; `scan_agents` only `read_dir`s. Prompt composition creates nothing. So the new code cannot make an empty `fake_home` non-empty, by construction.

Empirically, I could not reproduce it: **5 consecutive full `cargo test -p trusty-mpm --lib` runs under deliberate 4-way CPU load, all `3729 passed; 0 failed; 6 ignored`.** CI's `Test` job (16m29s, full workspace) is also green. The binary carries ~20 pre-existing `set_var("HOME")` sites; that pre-existing env-race surface is the plausible cause and is untouched here. Judgement: **pre-existing, correctly disclosed.**

### Green honesty — clean

`cargo test -p trusty-mpm --lib`: `3729 passed; 0 failed; 6 ignored`. All five new/extended tests present and reporting `ok`:

```
test core::delegation_authority::tests::every_bundled_non_base_agent_reaches_the_roster ... ok
test core::delegation_authority::tests::generate_authority_omits_self_referential_lines ... ok
test core::delegation_authority::tests::roster_from_dirs_dedup_is_case_insensitive ... ok
test core::instruction_overrides::tests::bundled_delegation_appends_deployed_roster ... ok
test core::instruction_overrides::tests::agent_delegation_override_replaces ... ok
```

No `#[ignore]`, no cfg-gate, no `--lib`/`--exclude` narrowing added. All 6 ignored tests are pre-existing environment gates (`requires claude binary absent from PATH`, `requires trusty-memory on PATH`, etc.), unchanged from round 1. 23/23 CI checks green. `Cargo.lock` untouched. CHANGELOG entry present under `## [Unreleased] / ### Fixed`.

### Round-1 findings — disposition

| Round-1 finding | Status |
|---|---|
| HIGH-1 `role: base` drops 3 agents | **Fixed** at the asset level + asset-invariant regression gate |
| HIGH-2 undispatchable agents | **Mitigated** via the caveat I offered; mapping independently verified |
| MEDIUM-1 silent fail-open on unreadable tier | **Fixed** — `NotFound` silent, everything else `tracing::warn!`; `info!` on the `None` reversion |
| MEDIUM-2 doctrine/roster contradiction | **Fixed** — precedence note, gated by test |
| MEDIUM-3 ~2 KB of zero-information lines | **Fixed** — 9,555 → 7,647 B, gated by test |
| LOW-1 case-sensitive dedup | **Fixed**, gated by test |
| LOW-2 dead roster computation in `instruction_pipeline.rs:343-355` | Not addressed — still LOW, non-blocking; carry to #4183 |

---

## Notes

**A real pre-existing defect surfaced while verifying HIGH-2 — worth a ticket, but not this PR's.** Because `tm launch` and `tm connect` deploy agents into `~/.claude/agents` and then spawn with `--setting-sources project,local`, **no tm-deployed agent loads at all on those two CLI paths today.** That is independent of this PR (it is why threading is unavailable, not something this PR caused), and this change strictly improves the situation there by at least advertising a roster with a re-route instruction. But the underlying deploy/read mismatch deserves its own issue — the fix is to point those two call sites at a project-tier `FrameworkPaths` rather than `default()`. Flagging for the owner to route; I have not filed it.

**Scope respected.** The append-vs-replace contradictions between the retained asset and the roster are #4183's, and #4194 already corrected the ticket-routing half on main — not re-raised.

**Confidence.** Every claim above is from a command I ran against this branch's head (`2b62f2ee`), not read off the diff. The one thing I could not falsify independently is the empirical claim that `--setting-sources project,local` governs *subagent* discovery and not merely settings files; I am relying on this repo's own "empirically verified against `claude` 2.1.201" note in `managed_config.rs:25`, as in round 1. It is load-bearing for HIGH-2's reasoning, but the conclusion (union + caveat) is the safe shape under either reading.

Issue #4173, opened 2026-07-27
feat(trusty-agents): shell/build/test execution grant for L0 orchestration


bobmatnyc:
## Task

Grant L0 (orchestration) persona permission to execute shell commands, run builds, and invoke test suites. L1 personas remain denied this capability.

## Scope

Currently no persona is granted `shell_exec`, `run_build`, `run_test`, or equivalent tools. L0 needs these to orchestrate fix rounds, validate PRs, and coordinate multi-step remediation.

1. Add the following tools to the L0 agent.toml allowlist:
   - `shell_exec` (or equivalent subagent runner with `runner = "claude-code"`)
   - `run_test` (or invoke via shell)
   - `run_build` (or invoke via shell)
   - Ensure these are scoped to project directories and cannot escape (containment via cwd)
2. Keep L1 denied these (assistant, cto-assistant, izzie). No shell execution for untrusted-content-ingesting personas.
3. Document the containment boundaries.

## Acceptance Criteria

- [ ] L0 can execute shell commands in the current project's working tree
- [ ] L0 can run `cargo test` / build commands
- [ ] L1 personas cannot execute shell commands
- [ ] Output from shell execution is legible and unfiltered
- [ ] Simple test confirms shell execution works (e.g., `echo "hello"`)

## Related

- Epic #4167 (L0/L1 tier model)
- Gap analysis: L1 "no shell/build/test execution grant" point
- Related epic #4147 (executable-skill containment; may share infrastructure)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Consolidated during queue triage (2026-07-28) into canonical issue #4167. The requirement remains represented there; closing this leaf as `not planned` removes duplicate tracking and does not assert that the work is implemented.

Issue #4170, opened 2026-07-27
feat(trusty-agents): GitHub PR/CI tool surface for L0 orchestration


bobmatnyc:
## Task

Add GitHub PR, CI check, and workflow-status tools to the L0 (orchestration) persona tool allowlist. L1 personas remain black-boxed (no gh access).

## Scope

1. Extend the tool registry and `~/.trusty-agents/agents/orchestration/agent.toml` (or equivalent) to include:
   - `gh pr list` / `gh pr view` / `gh pr create` / `gh pr merge` / `gh pr comment`
   - `gh run view` / `gh run list` / `gh run watch` (CI status)
   - `gh check-runs` or equivalent (check-run aggregation)
   - `gh issue view` / `gh issue list` (if not already present; for reviewing blockers)
   - Any other `gh` commands needed for PM orchestration tasks (e.g., `gh workflow run`)
2. Do NOT add these to L1 personas (assistant, cto-assistant, izzie). Keep the black-box posture.
3. Document in the agent.toml why L0 needs this surface.

## Acceptance Criteria

- [ ] L0 persona can execute `gh pr` and CI-status commands
- [ ] L1 personas do not gain these commands
- [ ] Each tool is tested via a simple integration test (e.g., `gh pr view <repo>/<number>`)
- [ ] Tool output is legible and unfiltered (full gh output format)

## Related

- Epic #4167 (L0/L1 tier model)
- Gap analysis: L1 "missing GitHub PR/CI tooling" point

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Consolidated during queue triage (2026-07-28) into canonical issue #4167. The requirement remains represented there; closing this leaf as `not planned` removes duplicate tracking and does not assert that the work is implemented.

Issue #4167, opened 2026-07-27
epic(trusty-agents): Level-0 orchestration assistant tier


bobmatnyc:
## Decision

Introduce an explicit Level-0 orchestration persona instead of widening standard assistants.

- **L0 orchestration assistant:** PM-tier orchestration authority, owner-accepted YOLO risk, and no ingestion of untrusted Gmail/Drive/Calendar content.
- **L1 standard assistants:** retain the current constrained black-box posture.

## Scope

- first-class tier/persona configuration and validation;
- narrowly defined L0 grants for delegation, review, and control-plane operations;
- explicit owner acknowledgement and unmistakable UI/CLI labeling;
- deny untrusted-content connectors and ordinary end-user deployment paths;
- audit every privileged action and provide a kill/recovery path;
- test that L1 permissions do not widen.

## Done when

An opted-in owner can run an L0 orchestration workflow end to end, all privileged actions are attributable, unsafe connector combinations are rejected, and existing L1 assistants remain unchanged.

The former implementation tickets #4170, #4172–#4175, and #4201 are consolidated here.

bobmatnyc:
## Unblocked — #4161 landed with the multi-hop defect genuinely fixed

Verified against merged code at `56337097` (tip of `origin/main`), not the PR diff:

- Third dispatch path (`run_pm_task_with_history`) **was** patched — `ctrl/pm_task/dispatch/history.rs:396-416` and `:571-587`. An earlier "third path is unpatched" finding was accurate when made but predates the branch's final state.
- All four `DelegateToAgentTool` sites enumerated; three compose taint, the fourth (`runtime/pm_mode.rs:97`) is unreachable by delegation.
- `compose_delegation_taint` is a true intersection at every outbound site; absent-signal branches fail closed; `dispatch_gated` treats `Some(vec![])` as deny-all.
- Multi-hop A→B→C pinned by `multi_hop_delegation_narrows_at_every_hop` (`runtime/tool_registry_tests.rs:653-733`), shaped to distinguish true intersection from "forward either side".

## Design constraint for this epic

The trust boundary is currently a literal, author-editable role string, and taint is opt-in rather than deny-by-default. `persona.rs:382` taints only when `role == "assistant"`; `history.rs:576` returns no posture for any other role. A persona declaring `[tools].allow` under a different role therefore gets an **untainted, un-role-gated** delegate tool on the persona-chat path.

[...]

Issue #4175, opened 2026-07-27
test(trusty-agents): L0/L1 tier integration tests and verification


bobmatnyc:
## Task

Develop comprehensive integration tests verifying the L0/L1 tier model and its security boundaries.

## Scope

1. **Tier-specific tool access tests:**
   - L0 has access to: gh, session_state, shell_exec, cross-project stores, git (full)
   - L1 has access to: git (read-only), local tools only; NO gh, NO session_state, NO shell, NO cross-project reach
   - Each tool tested with real invocation, not mocked

2. **Delegation constraint tests:**
   - L0 can delegate to L1 (succeeds)
   - L1 cannot delegate to L0 (refused with clear error)
   - Taint mechanism verified to block L1→L0 (if implemented via PR #4161 mechanism)

3. **Content ingestion tests:**
   - L0 personas have no Gmail/Drive/Calendar tools (explicit allowlist check)
   - L1 personas retain these (unchanged from current behavior)

4. **Real-world scenario tests:**
   - L0 can fetch PR status, reconcile with git branch, update a session (multi-step orchestration)
   - L1 can read local git state but cannot reach GitHub or session APIs
   - L0 can spawn shell commands; L1 cannot

5. **Security isolation tests:**
   - L1 cannot privilege-escalate to L0 via delegation chain
   - Cross-project access is scoped correctly (L0 can access; L1 is bound to Duetto)

## Acceptance Criteria

- [ ] All tier-specific tool access tests pass
- [ ] Delegation constraint tests pass
- [ ] Integration tests run as part of `cargo test --workspace`
- [ ] Tests are documented; rationale clear for each test
- [ ] Real-world scenario tests confirm PM orchestration workflow is viable

## Related

- Epic #4167 (L0/L1 tier model)
- All child issues (#4168–#4174)
- Blocked by PR #4161 for taint-mechanism testing

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Consolidated during queue triage (2026-07-28) into canonical issue #4167. The requirement remains represented there; closing this leaf as `not planned` removes duplicate tracking and does not assert that the work is implemented.

PR #4200, opened 2026-07-27
feat(trusty-agents): L0/L1 persona tier + one-directional delegation gate

### LINKED ISSUE #4168: feat(trusty-agents): define L0 persona schema in agent-config model
## Task

Define the L0 (orchestration assistant) persona schema as a first-class role in the agent-config data model. This enables separation of L0 and L1 (standard assistant) tiers with distinct capability grants.

## Scope

1. Extend the persona schema in `crates/trusty-agents/src/agent/config.rs` or equivalent (or `.trusty-agents/agent.toml` frontmatter) to define a new `role = "orchestration"` (or similar) distinct from `role = "assistant"`.
2. Document the semantic difference: L0 personas carry PM-tier grants (GitHub, session-state, shell, cross-project), while L1 personas retain the black-box posture.
3. Update manifest parsing to recognize and validate this role.
4. Unit tests confirming that a manifest with `role = "orchestration"` parses correctly and does NOT inherit L1 constraints.

## Acceptance Criteria

- [ ] New role value defined and validated in the agent schema
- [ ] Parsing handles both `"assistant"` (L1) and `"orchestration"` (L0) roles without ambiguity
- [ ] Documentation in the schema explains the behavioral difference
- [ ] Tests confirm correct role parsing and semantic application

## Related

- Issue #4126 (delegation RCE chain)

### LINKED ISSUE #4169: feat(trusty-agents): one-directional delegation gate — L1 cannot delegate to L0
## Task

Enforce the one-directional delegation constraint: L0 → L1 is permitted; L1 → L0 must be structurally forbidden.

## Context

If L1 personas can delegate into L0, then untrusted email or document content can acquire the owner's YOLO grant by simply asking L1 to hand the work down — a privilege-escalation path through delegation (issue #4126 reachable through a new vector).

This requires the multi-hop taint-composition fix from PR #4161 to be merged and working before this can be implemented correctly.

## Scope

1. Update the delegation gate in `crates/trusty-agents/src/runtime/tool_registry.rs` (around `ASSISTANT_ALLOWED_DELEGATE_ROLES`) to inspect the DELEGATOR's role and REFUSE if delegator is L1 and target is L0.
2. Alternatively (preferred, if PR #4161 taint mechanism is sufficient): use the taint to detect when an L1 persona is attempting L0 delegation and fail closed.
3. Unit test that explicitly validates the L1 → L0 refusal path:
   - L1 persona calls `delegate_to_agent(agent_name="orchestration-assistant")` 
   - Call is refused with legible error explaining the constraint
4. Counter-test: L0 → L1 delegation succeeds.

## Acceptance Criteria

- [ ] Delegation gate rejects L1 → L0 with clear error message
- [ ] Delegation gate permits L0 → L1
- [ ] Unit tests cover both paths
- [ ] Error message educates the user on the constraint

## Blocked By

- PR #4161 (multi-hop taint-composition fix required)

## Related

- Issue #4126 (prompt injection / RCE)
- PR #4161 (taint fix)

bobmatnyc:
## Summary

Foundation of epic #4167 (Level-0 orchestration assistant tier). Implements the owner's tiered persona model decision (2026-07-27): a new L0 "orchestration assistant" tier that will eventually carry PM-tier grants under an explicit YOLO risk posture, and today's `assistant`/`cto-assistant`/`izzie` become L1 "standard" — retaining the documented black-box posture.

- **#4168 — tier in the agent-config model.** Adds `AgentInfo.tier: Option<String>` (TOML `[agent].tier`, `.md` frontmatter `tier`) plus `AgentTier` (`L1Standard` / `L0Orchestration`) and `AgentInfo::tier()`, the single fail-closed resolver every consumer must call.
- **#4169 — the one-directional delegation gate.** `delegate_to_agent` refuses any call where the resolved target's tier is `L0Orchestration` unless the delegator's own tier is also `L0Orchestration`.

Closes #4168, Closes #4169

## Design decisions

**Dedicated `tier` field vs. deriving from `role`.** Chose a dedicated field. `role` already has three overloaded meanings in this codebase (dispatch-branch selection in `build_registry_for_agent`, delegate-target allowlisting via `ASSISTANT_ALLOWED_DELEGATE_ROLES`, and capability matching) — `AgentInfo.hidden`'s own doc comment already names the hazard of layering a fourth meaning onto `role`: *"repurposing role would silently change tool permissions too — not a safe mechanism."* A dedicated field keeps this privilege boundary explicit, grep-able, and immune to an unrelated `role` edit silently changing it.

**Fail-closed resolution (`AgentInfo::tier()`).** Mirrors the existing `ToolsConfig::search_indexes_enforced()` pattern: a raw `Option<String>` field plus a normalizing accessor. Absent, blank, or any unrecognized string resolves to `AgentTier::L1Standard` — never `L0Orchestration`. Recognized aliases: `l0`/`orchestration` → L0; `l1`/`standard` → L1 (case-insensitive, trimmed).

**`extends` composition.** Child-declares-wins, child-omits-inherits — the same posture `enforce_search_indexes`/`default_tier` already use ("a hardened base stays hardened under a silent overlay"). An L0 base's tier travels to a child overlay that never redeclares `tier`; a child can still explicitly downgrade (or upgrade) it.

**The delegation gate builds on #4161's taint machinery, not a replacement.** `compose_delegation_taint` / `scope_for_delegation` still own tool-surface narrowing across hops. The tier gate is an independent, additional pre-flight check inside `DelegateToAgentTool::execute()`, alongside the existing `allowed_target_roles` role gate.

**Every dispatch/registry-construction site, enumerated:**
| Site | Coverage |
|---|---|
| `runtime/subagent_mode.rs` → `tool_registry::build_assistant_tier_registry` | Gated — `cfg.agent.tier()` threaded through `with_delegator_tier` |
| `ctrl/pm_task/dispatch/persona.rs` (`run_pm_task_with_persona`, REPL `/agent`) | Gated — added `with_delegator_tier`; tier check now applies even without a role allowlist (a pre-existing, separate gap in this path) |
| `ctrl/pm_task/dispatch/history.rs` (`ctrl_delegate_posture`, backs REPL/Slack/Telegram/API) | Gated for the `role == "assistant"` branch (forwards its own declared tier); the `role != "assistant"` (pm/ctrl-as-orchestrator) branch reports `L0Orchestration` deliberately — it sits outside the persona tier model, already fully trusted, already unreachable as a delegation target from any L1 persona (`orchestrator` was never in `ASSISTANT_ALLOWED_DELEGATE_ROLES`) |
| `runtime/pm_mode.rs::run_pm()` | Not gated, deliberately — builds its `DelegateToAgentTool` with **no** `config_dirs`, so pre-flight validation (including the new tier gate) is skipped entirely, byte-identical to before. Documented in a code comment at the call site; unreachable transitively from any L1 persona for the same role-allowlist reason as above |

**Multi-hop escalation.** Because the tier check re-resolves the CURRENT delegator's own tier at every hop (not a value threaded/copied across hops), no chain of L1→L1→…→L0 can succeed — proven by `delegate_multi_hop_l1_peer_to_l0_is_refused` (assistant → izzie → orchestration-target, hop 2 refused). Worker roles (`engineer`, `qa-agent`, …) never register `delegate_to_agent` at all, closing the other multi-hop avenue structurally.

**A regression I caught and fixed during implementation:** my first pass made `delegator_tier` a plain `AgentTier` (default `L1Standard`) and forced full `AgentConfig` resolution whenever it wasn't `L0Orchestration` — this broke two pre-existing tests using bare/incomplete fixture TOMLs that only need the cheap existence-only check. Fixed by making the field `Option<AgentTier>`: `None` (never opted in) preserves the exact pre-#4169 cheap-path behavior; the tier gate activates when EITHER `allowed_target_roles` OR `delegator_tier` is explicitly set, with a fail-closed `L1Standard` default for a caller that set one but not the other. See `no_tier_and_no_role_gate_preserves_cheap_existence_check` for the regression guard.

## Not in scope (deliberately)

No L0 persona instance created, no `~/.trusty-agents/agents/` changes, no GitHub/PR/CI tooling, no session-state read, no cross-project scoping, no shell/build/test grant — those are #4170-#4173, and landing the boundary before the grants is the point (the reverse order would create a live escalation target).

## Test plan

- [x] `cargo fmt --check -p trusty-agents` — clean
- [x] `cargo test -p trusty-agents --lib` — 2832 passed, 0 failed, 11 ignored (1 known-flaky, environment-only `python_skill` test that also fails in isolation runs on unrelated `uv`/venv build contention — reproduced as passing standalone both before and after this diff)
- [x] `bash scripts/check_line_cap.sh` — 0 violations
- [x] `bash scripts/check_test_pointers.sh` — 0 dangling pointers
- [x] `bash scripts/check_sld.sh` — 0 errors/warnings
- [x] `bash scripts/check_claude_md_not_tracked.sh` / `check_instruction_floor.sh` — pass
- [x] The single most important test: `delegate_l1_to_l0_is_refused`
- [x] Counter-test: `delegate_l0_to_l1_succeeds`
- [x] Transitive/multi-hop: `delegate_multi_hop_l1_peer_to_l0_is_refused`
- [x] Fail-closed tier parsing: `agent_tier_defaults_to_l1_when_absent`, `agent_tier_unknown_value_fails_closed_to_l1`, `agent_tier_blank_value_fails_closed_to_l1`
- [x] `extends` composition: `extends_tier_child_inherits_l0_from_base_when_omitted`, `extends_tier_child_can_downgrade_l0_base_to_l1`
- [x] End-to-end through real production wiring (`agents_dir_candidates()` + `build_assistant_tier_registry`, not hand-built args): `assistant_tier_registry_delegator_tier_blocks_l0_target_end_to_end`

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -9,6 +9,20 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Security
 
+- **Defined the L0/L1 persona privilege tier and made the L1->L0 delegation
+  path structurally forbidden (#4168, #4169, epic #4167).** Added an
+  explicit `AgentTier` (`AgentInfo::tier`, TOML `[agent].tier` / `.md`
+  frontmatter `tier`), decoupled from `role`, that fails closed to
+  `L1Standard` on an absent, blank, or unrecognized value — never `L0`.
+  `delegate_to_agent` now refuses any call where the resolved TARGET's tier
+  is `L0Orchestration` unless the DELEGATOR's own tier is also
+  `L0Orchestration`, checked at every dispatch path that can construct a
+  `DelegateToAgentTool` (the `--direct`/`--agent` subprocess path, the
+  persona-chat REPL path, and the ctrl history/session path), including
+  through a peer-assistant delegation hop — no chain of L1 hops can reach
+  L0. This PR defines the tier and the boundary only; it grants L0 no new
+  capabilities and creates no L0 persona instance (those are #4170-#4173).
+
 - **Closed a prompt-injection-to-arbitrary-code-execution path through
   `delegate_to_agent` (#4126).** An assistant-tier persona (`assistant`,
   `cto-assistant`, `izzie`) holding live external-content tools
--- a/crates/trusty-agents/src/agents/claude_code_runner/tests.rs
+++ b/crates/trusty-agents/src/agents/claude_code_runner/tests.rs
@@ -70,6 +70,7 @@ fn test_agent_config(name: &str, model: &str, system_prompt: &str) -> AgentConfi
             kind: "assistant".to_string(),
             prompt_label: None,
             extends: None,
+            tier: None,
         },
         llm: LlmParams {
             temperature: 0.0,
--- a/crates/trusty-agents/src/agents/claude_mpm_loader.rs
+++ b/crates/trusty-agents/src/agents/claude_mpm_loader.rs
@@ -98,6 +98,7 @@ impl ClaudeMpmAgent {
                 kind: "assistant".to_string(),
                 prompt_label: None,
                 extends: None,
+                tier: None,
             },
             llm: LlmParams {
                 temperature: 0.3,
--- a/crates/trusty-agents/src/agents/config.rs
+++ b/crates/trusty-agents/src/agents/config.rs
@@ -639,9 +639,105 @@ pub struct AgentInfo {
     /// Test: `agents::extends::tests` — see `extends_*` unit tests.
     #[serde(default)]
     pub extends: Option<String>,
+
+    /// Privilege tier declaration (#4168, epic #4167 — L0/L1 orchestration
+    /// model), raw as parsed from TOML. Use [`AgentInfo::tier`] rather than
+    /// reading this field directly.
+    ///
+    /// Why: The owner's tiered persona model (#4167) needs a privilege
+    /// boundary distinct from `role`. `role` is already triple-overloaded —
+    /// it selects the per-agent tool-registry branch
+    /// (`runtime::tool_registry::build_registry_for_agent`), gates
+    /// `delegate_to_agent`'s target allowlist
+    /// (`ASSISTANT_ALLOWED_DELEGATE_ROLES`), and drives capability matching
+    /// — and `hidden`'s own doc comment above already names the hazard of
+    /// adding a FOURTH meaning to it: "repurposing role would silently
+    /// change tool permissions too — not a safe mechanism." A dedicated
+    /// field keeps this SPECIFIC boundary (L0 orchestration-tier PM-grants
+    /// vs. L1 standard black-box posture) explicit, grep-able, and
+    /// independent of any future `role` value — an operator adding a new
+    /// specialist `role` can never accidentally grant it orchestration-tier
+    /// delegation reach as a side effect.
+    /// What: Raw TOML string (`"l0"` / `"orchestration"` for L0,
+    /// `"l1"` / `"standard"` for L1, case-insensitive). Resolved by
+    /// [`AgentInfo::tier`], which FAILS CLOSED: `None` (field absent —
+    /// every persona shipped before #4168, and the overwhelming majority
+    /// going forward), an unrecognized string, or a blank/whitespace-only
+    /// string all resolve to [`AgentTier::L1Standard`] — the most
+    /// restricted tier — NEVER [`AgentTier::L0Orchestration`]. This is a
+    /// privilege boundary, not a convenience default: every call site that
+    /// needs to know an agent's tier MUST call `AgentInfo::tier()`, never
+    /// match on this raw field.
+    /// Test: `agent_tier_defaults_to_l1_when_absent`,
+    /// `agent_tier_parses_l0_orchestration_aliases`,
+    /// `agent_tier_parses_l1_standard_aliases`,
+    /// `agent_tier_unknown_value_fails_closed_to_l1`,
+    /// `agent_tier_blank_value_fails_closed_to_l1`.
+    #[serde(default)]
+    pub tier: Option<String>,
+}
+
+/// Persona privilege tier (#4168, epic #4167 — L0/L1 orchestration model).
+///
+/// Why: The owner decision (2026-07-27) splits personas into two tiers: L0
+/// ("orchestration assistant") carries PM-tier grants (GitHub/PR/CI,
+/// session-state read, cross-project git, shell/build/test execution,
+/// delegation) under a YOLO risk posture the owner explicitly accepts —
+/// deliberately NOT sandboxed, but it must never ingest untrusted content
+/// (no Gmail/Drive/Calendar surface). L1 ("standard assistant") is today's
+/// `assistant`/`cto-assistant`/`izzie`: the documented black-box posture,
+/// which DOES ingest untrusted content and therefore must never be able to
+/// acquire L0's grants — including via delegation (#4169, epic #4167).
+/// This PR defines the tier and its one-directional delegation boundary
+/// ONLY; it grants L0 no actual capabilities (those are #4170-#4173) and
+/// creates no L0 persona instance.
+/// What: Two variants. `L1Standard` is `#[default]` — every accessor that
+/// falls back to `Default::default()` (or an explicit fail-closed match
+/// arm) lands on the restricted tier, never the elevated one.
+/// Test: See `AgentInfo::tier`'s Test pointer for parsing; delegation-gate
+/// behavior is covered by `crate::tools::delegate` tests (search
+/// `l1_to_l0` / `l0_to_l1` / `tier`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum AgentTier {
+    /// L1 — standard assistant tier. Black-box posture; may ingest
+    /// untrusted content (Gmail/Drive/web); MUST NOT reach L0 via
+    /// delegation. The fail-closed default.
+    #[default]
+    L1Standard,
+    /// L0 — orchestration tier. PM-tier grants; YOLO risk posture
+    /// explicitly accepted by the owner; deliberately not sandboxed but
+    /// must never ingest untrusted content. Only reachable by an EXPLICIT
+    /// `tier = "l0"` / `tier = "orchestration"` declaration — never a
+    /// default, never inferred from `role`.
+    L0Orchestration,
 }
 
 impl AgentInfo {
+    /// Resolve this agent's privilege tier, fail-closed (#4168).
+    ///
+    /// Why: The single normalization point for the raw [`Self::tier`]
+    /// string — see that field's doc comment for the full fail-closed
+    /// rationale. Every consumer (the delegation gate, and any future
+    /// L0-gated capability) MUST call this rather than matching on the raw
+    /// field, so "what counts as L0" can never drift between call sites.
+    /// What: Case-insensitive, trimmed match against `"l0"`/`"orchestration"`
+    /// (-> [`AgentTier::L0Orchestration`]) and `"l1"`/`"standard"` (->
+    /// [`AgentTier::L1Standard`]); anything else — absent, blank, or an
+    /// unrecognized string — resolves to [`AgentTier::L1Standard`].
+    /// Test: `agent_tier_defaults_to_l1_when_absent`,
+    /// `agent_tier_parses_l0_orchestration_aliases`,
+    /// `agent_tier_parses_l1_standard_aliases`,
+    /// `agent_tier_unknown_value_fails_closed_to_l1`,
+    /// `agent_tier_blank_value_fails_closed_to_l1`.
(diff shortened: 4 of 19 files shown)


bobmatnyc:
## Adversarial review — PR #4200 (L0/L1 tier delegation gate)

**Head SHA reviewed:** `d9fa3e404c2af2ffc4f7a74bcd12e8235c3b8bda` (confirmed via `gh pr view --json headRefOid` and local checkout at the same commit)

## Verdict: APPROVE

## Method

Reviewed the diff directly (not the PR body's own claims) and independently:
- Built `trusty-agents` and ran the affected tests against a real local checkout at the exact head SHA (worktree already present at `.base/.worktrees/feat-4168-4169-l0-tier`).
- `cargo test -p trusty-agents --lib delegate` → **34 passed, 0 failed**, including `delegate_l1_to_l0_is_refused`, `delegate_l0_to_l1_succeeds`, `delegate_multi_hop_l1_peer_to_l0_is_refused`, `delegate_role_gate_without_declared_tier_fails_closed_to_l1`, `no_tier_and_no_role_gate_preserves_cheap_existence_check`.
- `cargo test -p trusty-agents --lib tier` → **59 passed, 0 failed**, including both `assistant_tier_registry_*_end_to_end` tests (real `agents_dir_candidates()` + `reg.dispatch(...)`, not hand-built args).
- `cargo test -p trusty-agents --lib` (full suite) → **2833 passed, 0 failed, 11 ignored** (PR claimed 2832; 1-off, not a discrepancy worth chasing — 0 failures either way).
- Every test name cited in the diff's doc comments exists and passes — no dangling test pointers (the #4161-round-1 defect class).
- `gh pr checks 4200` — all 23 jobs green on this exact SHA.
- `gh pr view 4200 --json closingIssuesReferences` — confirms both `#4168` and `#4169` are recognized as closing issues.

## Findings

| MEDIUM | `crates/trusty-agents/src/ctrl/pm_task/dispatch/persona.rs` | 393-397, 414 | `run_pm_task_with_persona`'s `DelegateToAgentTool` never calls `.with_allowed_target_roles(...)` — only the L0/L1 tier gate applies here, not the `ASSISTANT_ALLOWED_DELEGATE_ROLES` worker/peer-assistant allowlist the other two dispatch paths (`build_assistant_tier_registry`, `ctrl_delegate_posture`) both enforce. This is honestly disclosed in the PR body and pre-dates this diff (the diff only adds `with_delegator_tier` here, it doesn't touch role-gating) — I confirmed against the current bundled roster that this doesn't yield a *currently* exploitable escalation (`pm.toml`'s `role="orchestrator"` falls to the generic catch-all registry with only skill tools; `analysis-agent` gets read-only tools; the only shell-granting role, `local-ops-agent`/`ops`, is already a sanctioned `ASSISTANT_ALLOWED_DELEGATE_ROLES` target) — so it doesn't rise above MEDIUM today. But it is a real defense-in-depth gap in code this PR touches, and a future bundled agent with a more-privileged per-name registry arm in `build_registry_for_agent` would become silently delegatable from this exact chat path with no role check at all. | Add `.with_allowed_target_roles(ASSISTANT_ALLOWED_DELEGATE_ROLES...)` to this construction site (mirroring `ctrl_delegate_posture`'s and `build_assistant_tier_registry`'s shape) in a follow-up ticket off epic #4167, or file it explicitly against #3555's lineage so it doesn't get lost. |

No CRITICAL or HIGH findings against the specific property under review (L0/L1 one-directional delegation boundary).

## What I verified holds (with citations)

1. **Fail-closed parsing, exhaustively.** `AgentInfo::tier()` (`crates/trusty-agents/src/agents/config.rs:148-155`) only maps `"l0"`/`"orchestration"` (case-insensitive, trimmed) to `L0Orchestration`; every other input — absent, blank, garbage, `"L0RCHESTRATION"`, `"true"`, `"l2"` — falls through the `_` arm to `L1Standard`. Verified against the test matrix (`agent_tier_unknown_value_fails_closed_to_l1`, `agent_tier_blank_value_fails_closed_to_l1`) and independently ran them green.

2. **`delegator_tier: Option<AgentTier>` — the `None` case does not fail open.** In `tools/delegate.rs:1249-1266`, `needs_full_resolution = allowed_target_roles.is_some() || delegator_tier.is_some()`, and when true, `self.delegator_tier.unwrap_or_default()` resolves to `L1Standard` (the `#[default]` variant, pinned by `agent_tier_default_trait_is_l1_standard`) — i.e. an unset delegator tier is treated as the *least* privileged, not "unknown → allow." Confirmed no production call site sets `config_dirs` without also setting `delegator_tier`: `grep -n "DelegateToAgentTool::new" crates/trusty-agents/src` (excluding tests) returns exactly 4 sites — `persona.rs:415`, `history.rs:412`, `tool_registry.rs:538`, `pm_mode.rs:113`. The first three all call `.with_delegator_tier(...)` unconditionally; the fourth (`pm_mode.rs`) never calls `.with_config_dir`/`.with_config_dirs` at all, so it takes the `config_dirs.is_empty()` branch and skips pre-flight validation entirely — see point 4.

3. **Path coverage is exhaustive** — confirmed independently via grep, not by trusting the PR's own table. Only these 4 non-test construction sites exist in the crate; all four are individually accounted for above.

4. **`pm_mode.rs::run_pm()` is genuinely unreachable from L1, verified independently of the (real, separate) role-gate gap in `persona.rs`.** `SubprocessAgentRunner` — the only thing any `delegate_to_agent` call can spawn — always execs `current_exe --agent <name>` (`crates/trusty-agents/src/subprocess/spawn.rs:233`), never `--pm`. `run_pm()` is reachable ONLY via the top-level `--pm` CLI flag (`crates/trusty-agents/src/runtime/mode_dispatch.rs:345`). This is a structural, CLI-entry-point-level separation, not a role-allowlist argument — so it holds regardless of whether `persona.rs`'s role-gate is present (finding above notwithstanding). Even the degenerate case of delegating to the bundled `pm.toml` by name (which `AgentConfig::by_name_in` can resolve) only spawns `run_subagent("pm")`, which — since `pm.toml`'s `role = "orchestrator"` isn't `"assistant"` and has no dedicated match arm in `build_registry_for_agent`'s `match name` — falls to the generic catch-all (`tool_registry.rs:406-419`): skill tools only, no shell, no write, no unrestricted delegate.

5. **`extends` composition is correct in both directions.** `merge_extends` (`agents/extends/mod.rs:193-195`): `if child.agent.tier.is_some() { merged.agent.tier = child.agent.tier.clone(); }` — since `merged` starts as `base` (moved), an omitting child correctly inherits the base's tier by simply not overwriting it. Verified all four tests green, including the reverse direction (`extends_tier_child_can_declare_l0_over_l1_base`) that the PR's own list didn't call out as explicitly as the others.

6. **Multi-hop escalation is structurally closed, not just tested.** Each hop re-resolves its OWN tier from its OWN on-disk config at spawn time — `subagent_mode.rs:387`, `build_registry_for_agent(..., cfg.agent.tier())` where `cfg` is loaded fresh for the spawned agent `name`, never a value threaded down from the parent's delegation call. This means a compromised/malicious intermediate hop cannot forge a wider tier for itself by passing along a stale claim — it would have to falsify its own on-disk config, a different (and, per point 7, currently unreachable) attack. `delegate_multi_hop_l1_peer_to_l0_is_refused` exercises this with two real `DelegateToAgentTool::execute()` calls in sequence, not a hand-rolled reimplementation.

7. **Config-side self-escalation (a persona declaring its own `tier = "l0"`) is not reachable through the L1 personas' own tool surface today.** Checked `crates/trusty-agents/.trusty-agents/agents/assistant/agent.toml`'s `[tools].allow` — no `write_file`, no `shell_exec`, no generic filesystem-write tool of any kind (only `delegate_to_agent`, read-only git, the OKG-ingest family bounded to the persona's own KB tree, and the Gmail/Drive/Calendar/Docs/Sheets/Slides/Tasks surface). An L1 persona cannot currently edit its own `agent.toml` to self-promote. This is an assumption baked into the threat model (agent config directories stay non-writable by L1 tool grants) rather than a defect in this PR — worth keeping in mind as new tools are added to the assistant tier in future PRs.

8. **Test honesty.** Every test name cited in a doc comment in this diff exists and is green (ran locally, see Method). The two most load-bearing tests — `assistant_tier_registry_delegator_tier_blocks_l0_target_end_to_end` and its L0→L0 counter-test — drive the REAL `agents_dir_candidates()` resolution via `TAGENT_CONFIG_DIR` and `build_assistant_tier_registry(...)` unmodified, dispatched through `reg.dispatch("delegate_to_agent", ...)`, not a hand-built `DelegateToAgentTool` with mocked args — this is exactly the test-honesty gap #4161's first round was flagged for, and it is not repeated here.

## Notes

- The MEDIUM finding above is not blocking: it's pre-existing, honestly disclosed by the implementer, and I could not confirm a concretely exploitable escalation against the *current* bundled roster at >80% confidence (hence MEDIUM, not HIGH). It should still be ticketed — it sits in code this PR modifies and directly adjacent to the threat model this epic exists to close.
- This PR is explicitly scoped to defining the tier + the delegation boundary only (no L0 persona instance, no capability grants) — confirmed the diff does not create any `.trusty-agents/agents/*` file or widen any tool allowlist. Epic #4167's remaining sub-issues (#4170-#4175) are correctly left for follow-up PRs.

PR #3721, opened 2026-07-23
fix(trusty-mpm): refuse to recreate vanished workspace roots on snapshot writes + F3 streak alarm (#3715)

bobmatnyc:
## Mechanism (from issuecomment-5058820810's root-cause investigation)

`session_manager::snapshot::write_scrollback` (`crates/trusty-mpm/src/session_manager/snapshot.rs:111`) called `tokio::fs::create_dir_all(parent)` where `parent = <workspace_path>/.trusty-mpm` — `create_dir_all` recreates every missing ancestor, so if the workspace ROOT itself had vanished from disk (deleted out-of-band, or never resolvable), this one call silently reconstituted the entire worktree root as a bare, non-git, config-less shell containing only `.trusty-mpm/`. Success logged only at `debug!`, so a silent recreation left no trace in a normally INFO-configured daemon.log.

Called from `capture_into` via **both** `SessionManager::stop` (every graceful stop) and `mark_runtime_exited_stopped`/`runtime_reap` (the 60s reap loop) — so any stop of a session with a vanished `workspace_path` would self-heal a bare shell, silently.

**The original deleter of the worktree root is still unpinned** — this PR closes the recreation hole the investigation identified but does not explain what deleted the root in the first place. Issue stays open.

## Fix

1. **Root-existence guard** (`snapshot.rs`): `capture_into` now checks `workspace_root_exists(ws)` (`tokio::fs::metadata`) before ever touching the filesystem via the driver or `write_scrollback`. If the root is missing, the write is skipped entirely with `warn!` naming the session and path (`#3715`). Losing a scrollback snapshot is acceptable; fabricating a workspace shell under a live session record is not. The normal first-write case (root present, `.trusty-mpm/` subdir absent) is unaffected — `write_scrollback` still creates the subdir unconditionally in that case.
2. **Visibility**: `write_scrollback` now reports whether it had to create the `.trusty-mpm/` subdir; `capture_into` logs that at `info!` (previously always `debug!`), so subdir creation is no longer invisible in a normally-configured daemon. The refusal path in (1) is `warn!` with session id + path.
[...]

Issue #3715, opened 2026-07-23
PM session lost its trusty-mpm agent catalog mid-session (only Claude Code built-ins remain) — merge cadence degraded to general-purpose


bobmatnyc:
## Owner report (verbatim)

> A PM session (the tm-tagents session, worktree
> `.base/.worktrees/f443c12d-2fb6-4ce1-9f70-2e7695306e47`) reported mid-session
> that its specialized subagents — code-critic, rust-engineer, research,
> local-ops, etc. — are "no longer available this session. Only claude /
> general-purpose / Explore / Plan remain." Those remaining four are Claude
> Code built-ins; everything that vanished is a trusty-mpm compose-chain
> agent. The session says it can fall back to general-purpose but its merge
> cadence relies on the dedicated code-critic.

Reported alongside separate relaunch trouble today: "recorded pane gone, `tm`
relaunch refusing/flip-flopping."

## Root cause (confirmed from `~/.trusty-mpm/daemon.log`, read-only investigation)

For a **daemon-managed** session, `claude` is launched with
`--setting-sources project,local` (see `runtime::claude_code`), which
**excludes the user-tier `CLAUDE_CONFIG_DIR` roster** and makes the
**project-level `<worktree>/.claude/agents/` + `.claude/skills/`** directories
the only ones Claude Code actually reads — this is documented explicitly in
`crates/trusty-mpm/src/core/managed_config.rs` (DOC-34 / #1996). That
directory is (re)deployed every spawn/resume by
`session_launch::prepare_session` → `deploy_agents_filtered`
(`crates/trusty-mpm/src/core/session_launch/mod.rs:280`,
`crates/trusty-mpm/src/core/agent_deployer.rs`).

Timeline reconstructed from `~/.trusty-mpm/daemon.log` (all timestamps UTC;
local display = UTC‑4, e.g. `11:51:10Z` = `07:51:10` local):

1. **Since at least `2026-07-23T03:32:10Z`** (~8 hours), the daemon's
   per-minute `prune-worktrees` sweep logged, every single cycle:
   ```
   WARN trusty_mpm::session_manager::prune: prune-worktrees: active session
   path failed to canonicalize; using raw path as protective fallback
   (#1845 F3) path=.../.base/.worktrees/f443c12d-2fb6-4ce1-9f70-2e7695306e47
   ```
   This warning is specific to the `f443c12d` worktree — the sibling session
[...]

Issue #3649, opened 2026-07-22
design(trusty-mpm): worktree session ownership — prevent cross-session cleanup collisions


bobmatnyc:
## Owner Requirement (verbatim intent)

> \"The .base worktree issue is because one session is cleaning up another session's worktrees. We need to have worktrees segmented by session or have sessions track their own.\"

Worktree cleanup/reclamation must be SCOPED to the session that owns the worktree. A session must never be able to delete a worktree it did not create.

## Two Candidate Approaches

Both options should be evaluated by the owner/architect:

### **(A) Segment worktrees by session**
Each session gets its own namespace/subtree for its worktrees, and cleanup only ever operates within the caller's own namespace.

### **(B) Sessions track their own worktrees**
Each session records the worktrees it creates (an ownership tag / registry keyed by session id), and any reclamation only touches worktrees whose recorded owner is the calling session (or that are provably ownerless).

### 1. The Incident: 2026-07-21 Bare-Clone Destruction

An \`rm -rf .base/*\` from ONE session destroyed the SHARED \`.base\` bare clone's git internals, orphaning **~70 of 77 worktrees** across ALL sessions (git-dead — \`git status\` returns \"not a git repository\" inside them).

**Root cause:** All sessions share one \`.base\` bare clone and one worktree area, and cleanup was not scoped to the acting session.

**Detection issue:** #3605

### 2. Recurrence + Scale

The worktree area reached **1.1 TiB / 77 dirs**; a 2026-07-17 manual sweep had already reclaimed ~1.1 TiB days earlier — the leak recurs because there's no ownership model to safely automate reclamation.

### 3. Manual Workaround In Use Today (Strongest Evidence)

During 2026-07-22 cleanup passes, cleanup required HAND-MAINTAINED protection lists:
- Explicitly naming every live agent worktree
- Running \`lsof\` per-directory
- Cross-checking \`git worktree list\`

This is the MANUAL VERSION of what this issue asks to make structural. Cleanup safety currently depends on a human/PM correctly enumerating \"don't touch these\" every time.

A concurrent-agent fan-out (7+ live worktrees at once) makes hand-maintenance untenable.

### 4. Dual-Registry / Shared-State Hazard

Worktrees are tracked in more than one place:
- The main checkout's \`.git/worktrees\`
- On-disk \`.base/.claude/worktrees\`
- On-disk \`.base/.worktrees\`

[...]

PR #4202, opened 2026-07-28
harden(trusty-mpm): prevent, alarm on, and detect cross-session worktree destruction (#3764)

### LINKED ISSUE #3764: harden(trusty-mpm): prevent & detect cross-session worktree deletion — session-identity guard + #1744 cwd-collision alarm + F3 forensic capture
## Context

Follow-up to #3715 (PM session lost its agent catalog mid-session). A read-only
root-cause investigation exhaustively cleared **every daemon-internal code path**
capable of deleting a session worktree root — decommission, #3649 orphan-GC,
in-project hygiene, `migrate_aside`, `git worktree prune` — against the full
`~/.trusty-mpm/daemon.log` for 2026-07-23. All of them log unconditionally; the
log is silent on all of them. The recreation half is already guarded by #3721
(`write_scrollback` root-existence guard).

**Conclusion:** the `f443c12d` worktree root was deleted by an **out-of-band
manual `git worktree remove --force` / `rm -rf` run directly in a shell/pane** —
the only action category that removes a worktree root while leaving zero trace
in `daemon.log`. There is therefore no "deleter" code path to fix; this issue
tracks the **defensive hardening** the investigation surfaced so a repeat is
prevented and/or loudly detected.

Two findings drive the scope below:
- A **3-way cwd collision** on the exact path at 01:58–02:16Z (`#1744` "3 Active
  managed sessions share the same cwd") — three Active session records pointing
  at one worktree, hours before the corruption began.
- `workspace_guard.rs` enforces **path containment only, no session-identity
  check** — nothing stops a session (or an agent in its pane) from removing a
  *sibling* session's active worktree believing it's its own.

### 1. Session-identity guard on worktree removal
`crates/trusty-mpm/src/session_manager/workspace_guard.rs:18-100` guards by path
containment but performs no session-identity check. Add a guard so a
remove/reclaim path refuses to delete a worktree root that belongs to a
*different Active session* than the caller. Applies to any daemon-routed
worktree-remove path; the manual-shell case can't be code-blocked but the guard
closes the in-process hole and documents the invariant.

### 2. Escalate the #1744 cwd-collision from silent skip to a corruption alarm
`crates/trusty-mpm/src/daemon/api.rs:1338-1390` (`correlate_session_start`)
currently only *skips claude_session_id attribution* when ≥2 Active records share
a cwd. Escalate: when N≥2 Active session records canonicalize to the same
worktree path, emit an error-level corruption alarm (mirror the F3 streak-alarm
pattern from #3721) — this collision was the observed precursor state and today
passes almost silently.

### 3. Forensic-capture-on-F3 runbook / hook
On the `#1845 F3` canonicalize-failure WARN streak (already alarmed by #3721),
capture forensics BEFORE the runtime crashes and `write_scrollback` overwrites
the only surviving pane transcript: snapshot
`log show --last 1h --predicate 'eventMessage contains "<uuid>"'` and the suspect
panes' `tmux capture-pane -p -S -`. At minimum document this as a runbook in the
issue/DOC; optionally wire an automated capture on the streak trip. Rationale:
in this incident every pre-incident transcript was destroyed (scrollback.txt
mtime postdates the wipe) and history files held nothing — forensics is the
fastest remaining lever for the next occurrence.

## References
#3715 (parent symptom, stays OPEN until this lands), #3721 (recreation guard +
F3 alarm), #1744 (cwd-collision skip), #1845 (F3 canonicalize fallback), #1814 /
#2023 (runtime-reap, pane-left-alive), #3649 (orphan-GC, ruled out).

Files: `session_manager/workspace_guard.rs:18-100`, `daemon/api.rs:1338-1390`,
`session_manager/snapshot.rs:63-111` (guarded recreator), `session_manager/prune.rs:653-801` (ruled out).

bobmatnyc:
Implements all three scoped items of #3764, plus a fourth evaluated and found sound. **#3715 stays open until this lands.**

Filed P2 after the second occurrence; there has now been a **third** — the same `f443c12d` worktree lost its `.git` pointer, its registration, and its entire tracked source tree on 2026-07-24 22:56:07, and **kept running in that state for three days**. Three occurrences, zero PRs, and no shipped guard that detects or prevents the shape.

Every alarm added here is **ERROR-level**, and that is load-bearing rather than stylistic: the daemon composes `trusty_common::error_capture::bug_capture_layer`, which persists **only** ERROR events to `<data_dir>/trusty-mpm/errors.jsonl` and surfaces them through the `list_recent_errors` MCP tool, `preview_bug_report`, and `tm doctor`. `warn!` reaches none of those. Each of these conditions was previously at WARN — i.e. invisible to every operator-facing surface.

---

## Item 1 — session-identity guard on worktree removal

`workspace_guard::is_safe_to_remove` checked **path containment only** — a condition every sibling session's worktree also satisfies, so it green-lit cross-session deletion by construction. #3649's owner gate closes part of this but only fires when a session self-identifies as `caller`, and **every** daemon-routed remove path passes `caller: None`: the `session_decommission` MCP tool, the sm-stdio control channel, the HTTP routes, the idle reaper, and `dedup`. The hole was the whole daemon surface.

`decommission` now asks the **worktree's own on-disk ownership sentinel** who owns it and refuses with a new `ManagedError::ForeignActiveWorktree` when that owner is a different, still-Active session.

Two design points worth review:
- **Ownership is read from the tree, not from the record.** `known_owner_of` consults `record.worktree_owner` first, which is right for #3649 but exactly wrong here — the incident shape *is* a record whose `workspace_path` points at a peer's tree, so trusting the record would let the impostor name itself owner and launder the corruption.
- **It returns `Err` rather than skipping the removal.** The tombstone path clears `workspace_path`, which would destroy the only remaining evidence of the mis-pointing record.

The guard can only ever **refuse** a deletion, never permit a new one: unknown owner, self-owned, and any non-Active owner all pass through unchanged, so #3649's orphan-GC and #3721's recreation guard are untouched.

## Item 2 — the #1744 cwd collision becomes a corruption alarm

`correlate_session_start` treated ">=2 Active records at one cwd" as an *attribution* nuisance: skip the `claude_session_id`, `warn!` a bare count, carry on. It is not an attribution problem — it describes a state that cannot physically exist, and a 3-way collision on the exact path was the observed precursor hours before the last corruption.

Now an ERROR-level `SESSION REGISTRY CORRUPTION` alarm naming **every** colliding session id (the old warn logged only `n`, which gave an operator nothing to act on). The `match matched.len()` was extracted to a pure, testable `cwd_collision::classify` so the corruption verdict can never quietly fold back into "nothing to do" — which is what it effectively was, since the `n` arm and the `0` arm had identical effect.

## Item 3 — forensic capture runbook

`docs/runbooks/worktree-corruption-forensics.md`: what to capture and in what order, before a stop runs `write_scrollback` and overwrites the only surviving pane transcript. The #1845 F3 streak alarm now names the runbook in-message, and `forensics_runbook_path_exists` pins the path so the pointer cannot rot.

**Automated capture was evaluated and deliberately not wired** (rationale in the runbook's final section): the F3 streak trips inside the orphan-GC sweep, on the async executor, once per ~60s tick, and by construction trips *repeatedly* — wiring seconds-to-minutes of `log show` + full `capture-pane` there would stall the sweep indefinitely in exactly the degraded state where its other duties matter most. It would also need an incident-scoped location, a retention policy, and a redaction pass (pane scrollback contains secrets) to be safe. With the alarms now at ERROR, detection no longer depends on capture.

## Item 4 — destroyed-worktree self-check (the proposed fourth item)

**Verdict: sound, cheap, and implemented — but the proposed probe is not, and I did not ship it as proposed.**

Nothing in the tree asked "is *my* tree still there?", which is why a session ran three days in a stripped worktree. The fail-open is real and I reproduced it: with the `.git` pointer gone, discovery walks up to the enclosing clone, so `git log` prints plausible commits from the parent's stale `main` while `git status` fatals — and a fatal renders as `Status: (clean)`.

`git rev-parse --is-inside-work-tree` returns `false` here **only because `.base` is a bare clone**. That is not layout-independent. Verified empirically both ways:

| parent repo | stripped worktree: `--is-inside-work-tree` | `--show-toplevel` |
|---|---|---|
| bare | `false` -> detected | fatal -> detected |
| **normal** | **`true`** -> **MISSED** | parent dir -> detected |

Shipping the proposed probe alone would have reproduced the exact bug class this issue exists to end — a detector that reports healthy when it is not. The detector therefore compares **`git rev-parse --show-toplevel` against the session's own root**, which catches both layouts. `should_have_flagged_stripped_worktree_under_normal_parent` asserts, in one run, that the naive probe really does say `true` for a destroyed tree and that ours says `Destroyed` anyway.

False-positive modes considered: only `is_session_worktree` paths are audited (local-path/adopted sessions on plain directories are never judged); comparison is canonicalized (macOS `/tmp` <-> `/private/tmp`); a probe that cannot run reports `Unknown`, **never** `Intact`. The audit is read-only — it reports, never repairs, because silently reconstituting a vanished root is the #3715 masking behaviour. `Unknown` is deliberately **not** rolled up at ERROR: an alarm that fires on "git was briefly unavailable" gets muted, and a muted alarm is a silent one.

Runs on the existing orphan-GC tick — no new loop, no new config; the first tick after daemon start doubles as the at-start self-check.

---

## Mutation evidence

Every item was mutation-tested: change reverted, test observed failing, change restored, test observed passing.

| Mutation | Result |
|---|---|
| Item 1 — guard disabled in `decommission` | `decommission_refuses_to_delete_live_peer_worktree` **FAILED** — the impostor's decommission succeeded, record went `Decommissioned`, `workspace_removed=true`; the peer's live tree was destroyed |
| Item 2a — `Collision` arm reverted to `None` (pre-#3764 silent skip) | `classify_two_matches_is_collision` + `classify_three_matches_lists_all_ids` **FAILED** |
| Item 2b — `error!` reverted to `warn!` | `alarm_is_error_level` **FAILED**: `observed levels: [Level(Warn)]` |
| Item 3 — runbook file deleted | `forensics_runbook_path_exists` **FAILED** |
| Item 4 — detector swapped to `--is-inside-work-tree` | `should_have_flagged_stripped_worktree_under_normal_parent` **FAILED**: `got Intact` for a fully destroyed worktree |

## Gate (raw, from a compliant worktree location)

```
cargo check -p trusty-mpm            Finished `dev` profile ... in 10.10s
cargo clippy --all-targets           0 warnings attributable to trusty-mpm
cargo clippy --all-targets -D warn   EXIT=0
cargo fmt --all -- --check           EXIT=0
cargo test -p trusty-mpm             EXIT=0
  test result: ok. 3750 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
scripts/check_line_cap.sh            line-cap: 7 allowlisted, 0 violations — OK.
scripts/check_test_pointers.sh       test-pointers: 0 dangling pointers — OK.
```

28 new tests, all green. `Cargo.lock` untouched.

## Note on `ManagedError`

Adding the `ForeignActiveWorktree` variant pushed `manager.rs` to 501 SLOC. The enum moved to `session_manager/error.rs` and is re-exported, mirroring the #1955 `ManagedTmuxDriver` split that `manager.rs`'s own comment documents. No behaviour change; every `super::manager::ManagedError` import path still resolves.

## Two pre-existing issues found while gating (not fixed here)

1. `tm_hook_pm_guard`'s `pm_guard_allows_worktree_add_under_project_dir_via_subagent_payload` and `pm_guard_claude_mpm_sub_agent_env_still_allows_everything_else` **fail on clean `origin/main`** whenever the checkout lives under `/private/tmp` — they pass a *relative* worktree path that the #3955 guard resolves against cwd. Verified with a control worktree of unmodified `main` at the same location.
2. `telegram::supervisor::tests::healthy_run_resets_transient_backoff` is a timing flake (failed once at 34.9ms, passed 3/3 in isolation).

Closes #3764

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -9,6 +9,59 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **Detect a session whose own worktree has been destroyed underneath it**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)): the
+  orphan-GC tick now audits every Active session's own worktree and raises an
+  ERROR-level `WORKTREE DESTROYED` alarm when it is no longer a git work tree.
+  Nothing previously asked this question, which is why a session ran for three
+  days inside a stripped worktree while every surface reported it healthy: with
+  the `.git` pointer gone, git discovery walks up to the enclosing clone, so
+  `git log` still prints plausible commits and `git status`'s fatal renders as
+  `Status: (clean)`. The probe is `git rev-parse --show-toplevel` compared
+  against the session's own root — deliberately **not**
+  `--is-inside-work-tree`, which returns `true` for a fully destroyed worktree
+  whenever the parent clone is non-bare (regression-tested against real git in
+  both layouts). A probe that cannot run reports `Unknown`, never healthy. The
+  audit is read-only: it reports, never repairs.
+- **Forensics runbook for worktree corruption**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)):
+  `docs/runbooks/worktree-corruption-forensics.md` — what to capture, in what
+  order, before a stop overwrites the only surviving pane transcript, plus the
+  one probe that is safe to trust from inside a suspect directory. The #1845 F3
+  streak alarm now names this runbook in-message; a test pins that the path
+  resolves so the pointer cannot rot.
+
+### Fixed
+
+- **Refuse to delete a live peer session's worktree**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)):
+  `decommission` now asks the worktree's on-disk ownership sentinel who owns it
+  and refuses, with an ERROR alarm and a new `ForeignActiveWorktree` error, when
+  that owner is a different session that is still Active. The pre-existing
+  containment guard only ever checked that a path was inside the managed root —
+  which every sibling session's worktree also is — and #3649's owner gate only
+  fires when a session self-identifies as `caller`, so every daemon-routed
+  remove path (the `session_decommission` MCP tool, the sm-stdio control
+  channel, the HTTP routes, the idle reaper, `dedup`) passed `None` and skipped
+  it entirely. That left the observed incident shape unguarded: a #1744 cwd
+  collision leaves several Active records pointing at one worktree, and
+  decommissioning any impostor destroys the real owner's live tree. The new
+  guard is independent of `caller`, reads ownership from the tree rather than
+  from the (possibly corrupt) record that asked, and can only ever refuse a
+  deletion — a non-Active owner still reclaims exactly as before, so #3649's
+  orphan-GC and #3721's recreation guard are untouched.
+- **The #1744 cwd collision is now a corruption alarm, not a silent skip**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)): when ≥2
+  Active session records resolve to the same worktree path,
+  `correlate_session_start` raises an ERROR-level `SESSION REGISTRY CORRUPTION`
+  alarm naming every colliding session id, instead of the previous `warn!` that
+  logged only a count. Two Active records at one worktree describes a state that
+  cannot physically exist, and a 3-way collision on the exact path was the
+  observed precursor hours before the last corruption. The level is
+  load-bearing: the daemon's bug-capture layer persists only ERROR events to
+  `errors.jsonl`, so at WARN this condition was invisible to
+  `list_recent_errors` and `tm doctor` alike.
+
 - **Cross-branch `git push` guard for every worktree of a managed base clone**
   ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): a bundled
   `pre-push` hook now refuses any push that would OVERWRITE AN EXISTING remote
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -11,18 +11,27 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **Detect a session whose own worktree has been destroyed underneath it**
   ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)): the
-  orphan-GC tick now audits every Active session's own worktree and raises an
-  ERROR-level `WORKTREE DESTROYED` alarm when it is no longer a git work tree.
-  Nothing previously asked this question, which is why a session ran for three
-  days inside a stripped worktree while every surface reported it healthy: with
-  the `.git` pointer gone, git discovery walks up to the enclosing clone, so
-  `git log` still prints plausible commits and `git status`'s fatal renders as
-  `Status: (clean)`. The probe is `git rev-parse --show-toplevel` compared
-  against the session's own root — deliberately **not**
-  `--is-inside-work-tree`, which returns `true` for a fully destroyed worktree
-  whenever the parent clone is non-bare (regression-tested against real git in
-  both layouts). A probe that cannot run reports `Unknown`, never healthy. The
-  audit is read-only: it reports, never repairs.
+  orphan-GC tick now audits every Active session's own worktree and alarms at
+  ERROR when it is no longer a git work tree. Nothing previously asked this
+  question, which is why a session ran for three days inside a stripped worktree
+  while every surface reported it healthy: with the `.git` linkage gone, git
+  discovery walks up to the enclosing clone, so `git log` still prints plausible
+  commits and `git status`'s fatal renders as `Status: (clean)`. The probe is
+  `git rev-parse --show-toplevel` compared against the session's own root —
+  deliberately **not** `--is-inside-work-tree`, which returns `true` for a fully
+  destroyed worktree whenever the parent clone is non-bare (regression-tested
+  against real git in both layouts). The alarm distinguishes **linkage lost with
+  files still on disk** (recover them; do NOT decommission) from a genuinely
+  **empty** directory, so it never tells an operator their work is gone on a
+  tree whose files survive — the state the 07-21 `.base` destruction left ~70
+  worktrees in. The git probe is bounded by the same timeout
+  `remove_session_worktree` uses, so a stalled mount cannot wedge the sweep; a
+  probe that times out or cannot run reports `Unknown`, never healthy, and an
+  audit in which *every* verdict is `Unknown` alarms at ERROR in its own right
+  ("the detector is dark"). Findings are latched — alarm on transition, then
+  hourly — so one persistent finding cannot rotate the 500-entry error ring and
+  evict everything else from `list_recent_errors`. The audit is read-only: it
+  reports, never repairs.
 - **Forensics runbook for worktree corruption**
   ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)):
   `docs/runbooks/worktree-corruption-forensics.md` — what to capture, in what
@@ -30,37 +39,6 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   one probe that is safe to trust from inside a suspect directory. The #1845 F3
   streak alarm now names this runbook in-message; a test pins that the path
   resolves so the pointer cannot rot.
-
-### Fixed
-
-- **Refuse to delete a live peer session's worktree**
-  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)):
-  `decommission` now asks the worktree's on-disk ownership sentinel who owns it
-  and refuses, with an ERROR alarm and a new `ForeignActiveWorktree` error, when
-  that owner is a different session that is still Active. The pre-existing
-  containment guard only ever checked that a path was inside the managed root —
-  which every sibling session's worktree also is — and #3649's owner gate only
-  fires when a session self-identifies as `caller`, so every daemon-routed
-  remove path (the `session_decommission` MCP tool, the sm-stdio control
-  channel, the HTTP routes, the idle reaper, `dedup`) passed `None` and skipped
-  it entirely. That left the observed incident shape unguarded: a #1744 cwd
-  collision leaves several Active records pointing at one worktree, and
-  decommissioning any impostor destroys the real owner's live tree. The new
-  guard is independent of `caller`, reads ownership from the tree rather than
-  from the (possibly corrupt) record that asked, and can only ever refuse a
-  deletion — a non-Active owner still reclaims exactly as before, so #3649's
-  orphan-GC and #3721's recreation guard are untouched.
-- **The #1744 cwd collision is now a corruption alarm, not a silent skip**
-  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)): when ≥2
-  Active session records resolve to the same worktree path,
-  `correlate_session_start` raises an ERROR-level `SESSION REGISTRY CORRUPTION`
-  alarm naming every colliding session id, instead of the previous `warn!` that
-  logged only a count. Two Active records at one worktree describes a state that
-  cannot physically exist, and a 3-way collision on the exact path was the
-  observed precursor hours before the last corruption. The level is
-  load-bearing: the daemon's bug-capture layer persists only ERROR events to
-  `errors.jsonl`, so at WARN this condition was invisible to
-  `list_recent_errors` and `tm doctor` alike.
 
 - **Cross-branch `git push` guard for every worktree of a managed base clone**
   ([#2867](https://github.com/bobmatnyc/trusty-tools/issues/2867)): a bundled
@@ -169,6 +147,41 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   ordering-dependent behaviour is reproducible.
 
 ### Fixed
+
+- **Refuse to delete another session's worktree**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)):
+  `decommission` now establishes who owns a worktree from evidence the record
+  requesting the removal does not control — the worktree's own on-disk ownership
+  sentinel, plus an unforgeable store-side check for any OTHER record bound to
+  the same canonical path — and refuses with an ERROR alarm and a new
+  `ForeignActiveWorktree` error (HTTP 409) when that owner is a different
+  session which is not provably ownerless. The pre-existing containment guard
+  only ever checked that a path was inside the managed root, which every sibling
+  session's worktree also is; and #3649's owner gate only fires when a session
+  self-identifies as `caller`, so every daemon-routed remove path (the
+  `session_decommission` MCP tool, the sm-stdio control channel, the HTTP
+  routes, the idle reaper, `dedup`) passed `None` and skipped it. That left the
+  observed incident shape unguarded: a #1744 cwd collision leaves several
+  records pointing at one worktree, and decommissioning any impostor destroys
+  the real owner's tree. Reclaimability uses the same `resolve_ownerless`
+  predicate as the #3649 gate and the orphan sweep, so a **Stopped** peer's
+  resumable worktree is protected too — the reachable case being `idle_reaper`
+  stopping a peer and later reaping a colliding record via `caller: None`. Only
+  a provably ownerless owner (terminal, or absent from the store) still
+  reclaims, so #3649's orphan-GC and #3721's recreation guard are untouched, and
+  `tm sessions delete <stale-id>` remains the operator's release valve for a
+  stale sentinel.
+- **The #1744 cwd collision is now a corruption alarm, not a silent skip**
+  ([#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764)): when ≥2
+  Active session records resolve to the same worktree path,
+  `correlate_session_start` raises an ERROR-level `SESSION REGISTRY CORRUPTION`
+  alarm naming every colliding session id, instead of the previous `warn!` that
+  logged only a count. Two Active records at one worktree describes a state that
+  cannot physically exist, and a 3-way collision on the exact path was the
+  observed precursor hours before the last corruption. The level is
+  load-bearing: the daemon's bug-capture layer persists only ERROR events to
+  `errors.jsonl`, so at WARN this condition was invisible to
+  `list_recent_errors` and `tm doctor` alike.
 
 - **Two wrong instructions corrected in the bundled assets composed into every
   session's prompt.** Both were static text that no project could override, and
--- a/crates/trusty-mpm/src/daemon/api.rs
+++ b/crates/trusty-mpm/src/daemon/api.rs
@@ -1329,11 +1329,19 @@ pub async fn ingest_hook(
 /// What: extracts `cwd` from `payload["cwd"]`; canonicalizes it and each record's
 /// `workspace_path`/`cwd` (falling back to raw path on error so macOS
 /// `/private/tmp` ↔ `/tmp` symlinks resolve); finds Active managed sessions whose
-/// path matches. If more than one Active session matches the cwd (ambiguous) the
-/// correlation is skipped with a warning to avoid mis-attribution. Single match →
-/// `SessionManager::set_claude_session_id`. Best-effort — failures are logged and
-/// silently swallowed so a missing correlation never blocks the hook response.
-/// Test: `session_start_hook_correlates_claude_id` in `api_tests.rs`.
+/// path matches. The matched set is classified by
+/// [`cwd_collision::classify`](crate::daemon::cwd_collision::classify): a single
+/// match → `SessionManager::set_claude_session_id`; ≥2 matches → attribution is
+/// still skipped (mis-assignment remains the worse outcome) but the condition is
+/// now escalated to an ERROR-level corruption alarm via
+/// [`cwd_collision::alarm`](crate::daemon::cwd_collision::alarm) instead of the
+/// pre-#3764 `warn!` that no operator surface ever saw. Two Active records at one
+/// worktree is registry corruption, not an attribution nuisance — see that
+/// module's doc for why ERROR (not WARN) is load-bearing here.
+/// Best-effort otherwise — failures are logged and swallowed so a missing
+/// correlation never blocks the hook response.
+/// Test: `session_start_hook_correlates_claude_id` in `api_tests.rs`;
+/// `classify_*` / `alarm_is_error_level` in `daemon::cwd_collision`.
 async fn correlate_session_start(
     state: &Arc<DaemonState>,
     claude_session_id: &str,
@@ -1364,12 +1372,13 @@ async fn correlate_session_start(
             let cwd_canon = std::fs::canonicalize(&r.cwd).unwrap_or_else(|_| r.cwd.clone());
             ws_canon.as_deref() == Some(hook_cwd.as_path()) || cwd_canon == hook_cwd
         })
+        .map(|r| r.id)
         .collect();
 
-    match matched.len() {
-        0 => {} // no Active managed session at this cwd — silent, not an error
-        1 => {
-            let id = matched[0].id;
+    match crate::daemon::cwd_collision::classify(&matched) {
+        // No Active managed session at this cwd — ordinary, not an error.
+        crate::daemon::cwd_collision::CwdCorrelation::None => {}
+        crate::daemon::cwd_collision::CwdCorrelation::Unique(id) => {
             match mgr.set_claude_session_id(&id, claude_session_id).await {
                 Ok(()) => tracing::info!(
                     managed_id = %id,
@@ -1383,13 +1392,8 @@ async fn correlate_session_start(
                 ),
             }
         }
-        n => {
-            tracing::warn!(
-                cwd = %cwd_str,
-                n,
-                "SessionStart: {n} Active managed sessions share the same cwd — \
-                 skipping claude_session_id attribution to avoid mis-assignment (#1744)"
-            );
+        crate::daemon::cwd_collision::CwdCorrelation::Collision(ids) => {
+            crate::daemon::cwd_collision::alarm(cwd_str, &ids);
         }
     }
 }
--- a/crates/trusty-mpm/src/daemon/cwd_collision.rs
+++ b/crates/trusty-mpm/src/daemon/cwd_collision.rs
@@ -25,7 +25,7 @@
 //! What: [`CwdCorrelation`] (the pure verdict), [`classify`] (the pure
 //! classifier — the testable core), and [`alarm`] (the loud rendering of a
 //! [`CwdCorrelation::Collision`]).
-//! Test: `classify_*` and `collision_*` below; the wired-in behaviour is
+//! Test: `classify_*` and `alarm_is_error_level` below; the wired-in behaviour is
 //! covered by `session_start_hook_correlates_claude_id` in `api_tests.rs`.
 
 use crate::session_manager::record::ManagedSessionId;
@@ -80,8 +80,8 @@ pub(crate) fn classify(matched: &[ManagedSessionId]) -> CwdCorrelation {
 /// `session_manager::prune`, which escalates the same way for the same reason.
 /// What: one `error!` naming the cwd, the collision count, and EVERY colliding
 /// session id, plus the concrete next action. No side effects beyond the log.
-/// Test: `alarm_is_error_level` (pins the level via `tracing::Level`),
-/// `collision_ids_are_all_reported`.
+/// Test: `alarm_is_error_level` (pins the level via `tracing::Level`, and that
+/// every colliding id appears in the `sessions` field).
 pub(crate) fn alarm(cwd: &str, ids: &[ManagedSessionId]) {
     let rendered: Vec<String> = ids.iter().map(ToString::to_string).collect();
     tracing::error!(
--- a/crates/trusty-mpm/src/daemon/cwd_collision.rs
+++ b/crates/trusty-mpm/src/daemon/cwd_collision.rs
@@ -1,0 +1,231 @@
+//! `SessionStart` cwd→session correlation, and the #3764 item-2 corruption
+//! alarm for a colliding cwd.
+//!
+//! Why: `correlate_session_start` (`daemon/api.rs`) has always treated "≥2
+//! Active managed session records resolve to the same cwd" as a mere
+//! *attribution* problem — it skipped persisting `claude_session_id` and logged
+//! a `warn!`, then carried on as if nothing were wrong. It is not an
+//! attribution problem. Two Active records pointing at ONE worktree means the
+//! registry is describing a state that cannot physically exist, and it is the
+//! precursor the #3715 worktree-destruction forensics found sitting on the
+//! exact path that was later wiped: a 3-way collision at 01:58–02:16Z, hours
+//! before the corruption began, that passed almost silently. Any decommission,
+//! reap, or GC of the wrong one of those records destroys a LIVE session's
+//! tree (which is what the #3764 item-1 guard now refuses).
+//!
+//! The escalation from `warn!` to `error!` is the whole point and is NOT
+//! cosmetic: the daemon composes `trusty_common::error_capture::bug_capture_layer`
+//! (`bin/tm/main.rs`), which persists ERROR-level events to
+//! `<data_dir>/trusty-mpm/errors.jsonl` and surfaces them through the
+//! `list_recent_errors` MCP tool, `preview_bug_report`, and `tm doctor`.
+//! `warn!` is captured by NONE of those. At WARN this condition was invisible
+//! to every operator-facing surface; at ERROR it becomes a durable, greppable,
+//! queryable record the moment it first occurs.
+//!
+//! What: [`CwdCorrelation`] (the pure verdict), [`classify`] (the pure
+//! classifier — the testable core), and [`alarm`] (the loud rendering of a
+//! [`CwdCorrelation::Collision`]).
+//! Test: `classify_*` and `collision_*` below; the wired-in behaviour is
+//! covered by `session_start_hook_correlates_claude_id` in `api_tests.rs`.
+
+use crate::session_manager::record::ManagedSessionId;
+
+/// What a `SessionStart` hook's cwd resolved to among the Active records.
+///
+/// Why: making the three outcomes an explicit enum — instead of a bare
+/// `match matched.len()` buried in a 2 000-line HTTP module — is what makes the
+/// corruption case testable at all, and what stops a future edit from quietly
+/// folding [`Self::Collision`] back into "nothing to do".
+/// What: [`None`](Self::None) — no Active record at this cwd (ordinary; an
+/// unmanaged shell). [`Unique`](Self::Unique) — exactly one; attribute to it.
+/// [`Collision`](Self::Collision) — TWO OR MORE, which is registry corruption.
+/// Test: `classify_no_match_is_none`, `classify_single_match_is_unique`,
+/// `classify_two_matches_is_collision`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum CwdCorrelation {
+    /// No Active managed session is running at this cwd.
+    None,
+    /// Exactly one Active managed session — safe to attribute.
+    Unique(ManagedSessionId),
+    /// ≥2 Active records share this cwd: physically impossible, corruption.
+    Collision(Vec<ManagedSessionId>),
+}
+
+/// Classify the Active records that matched a `SessionStart` hook's cwd.
+///
+/// Why: the pure half, so the corruption verdict can be asserted directly
+/// rather than by scraping log output. Pre-#3764 this logic existed only as an
+/// inline `match matched.len() { 0 => {}, 1 => ..., n => warn!(...) }`, whose
+/// `n` arm was indistinguishable from the `0` arm in effect — both did nothing.
+/// What: `[]` → [`CwdCorrelation::None`]; `[id]` → [`CwdCorrelation::Unique`];
+/// anything longer → [`CwdCorrelation::Collision`] carrying every colliding id
+/// so the alarm can name them (the pre-#3764 warn logged only a COUNT, which
+/// gave an operator nothing to act on).
+/// Test: `classify_no_match_is_none`, `classify_single_match_is_unique`,
+/// `classify_two_matches_is_collision`, `classify_three_matches_lists_all_ids`.
+pub(crate) fn classify(matched: &[ManagedSessionId]) -> CwdCorrelation {
+    match matched {
+        [] => CwdCorrelation::None,
+        [only] => CwdCorrelation::Unique(*only),
+        many => CwdCorrelation::Collision(many.to_vec()),
+    }
+}
+
+/// Emit the error-level corruption alarm for a colliding cwd (#3764 item 2).
+///
+/// Why: see the module doc — ERROR is the only level the daemon's bug-capture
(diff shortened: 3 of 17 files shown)


bobmatnyc (commented):
## Verdict: WARN

Independent adversarial review (code + spec only). Verified against a fresh detached
worktree of `77bdab2` at a compliant location, plus standalone git experiments.

**What I confirmed as claimed.** The Item-4 correction of the proposed probe is the
strongest work here and it holds up independently. Reproduced against git 2.54.0 on a
real stripped worktree in both layouts:

| parent | `--is-inside-work-tree` | `--show-toplevel` |
|---|---|---|
| bare | `false` | `fatal: this operation must be run in a work tree` |
| **normal** | **`true`** (misses it) | resolves to the parent dir (catches it) |

Rejecting `--is-inside-work-tree` was correct, and
`should_have_flagged_stripped_worktree_under_normal_parent` pins it in-run.

Gate honesty checks out. From a compliant location: `test result: ok. 3750 passed; 0
failed; 6 ignored` on the lib plus every integration binary green (including
`tm_hook_pm_guard`, consistent with the disclosed `/private/tmp`-only failure), all 28
new tests present and `... ok`, no `#[ignore]` in any new file, no `--lib` narrowing.
I re-ran the load-bearing mutation myself — disabling the Item-1 guard:

```
decommission_refuses_to_delete_live_peer_worktree ... FAILED
  ... state: Decommissioned, workspace_path: None ..., true)
```

i.e. the impostor's decommission succeeded and `workspace_removed=true` — the peer's
live tree destroyed. The test is real.

The over-refusal matrix also holds: self-owned, unknown/absent/empty/malformed sentinel,
and non-Active owner all pass through (`foreign_active_owner_*`, five tests, all green),
and the sentinel is rewritten unconditionally at both provision sites
(`inproject.rs:471`, `workspace.rs:849`), so a recycled path cannot carry a stale owner.
Ownership genuinely reads tree-not-record; the reasoning for that is right.

The four HIGH findings below are all in the *new* surfaces, not in what was replaced.
None is a regression against `main` — the guard strictly narrows what can be deleted and
the detector is read-only. They are gaps that will bite in production.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/session_manager/workspace_guard.rs` | 169 | Guard narrows to `Active`, while the #3649 gate 20 lines earlier in the same call uses `resolve_ownerless` (Stopped/Errored/Provisioning are NOT ownerless). Result: `caller: Some(x)` refuses to delete a **Stopped** peer's worktree, `caller: None` — every daemon path this PR exists to cover — allows it. | Replace `matches!(owner_state, Some(Active))` with the existing predicate: refuse unless `self.resolve_ownerless(owner).await`. |
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 237 | `spawn_blocking(move \|\| check(&probe_path)).await` shells out to `git` with **no timeout**, inside `orphan_gc_loop`. One worktree on a stalled network mount blocks the audit, everything after it in the tick (the search-index sweep), and every future tick, indefinitely. | Wrap in `tokio::time::timeout` exactly as `remove_session_worktree` already is via `GIT_WORKTREE_REMOVE_TIMEOUT` (`decommission.rs:66`); map `Err(_elapsed)` to `WorktreeIntegrity::Unknown`. |
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 251 | `Destroyed` cannot distinguish "tree wiped" from "git linkage lost, files intact", but the alarm asserts *"Its uncommitted work is GONE. Stop the session and recreate it"*. Following that instruction runs `decommission` (which carries no `#4118` dirt guard) and deletes work that was still on disk. | Before alarming, count non-sentinel entries under the root; when non-empty emit a distinct message ("git linkage lost; N entries still present — recover before decommissioning"). Do not claim the work is gone unless the directory is empty. |
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 246-302 | The audit re-alarms every tick with no latch. One persistent `Destroyed` finding emits 2 ERROR events / 60 s = ~2,880/day into a **500-entry** ring (`error_capture/store.rs:34`; `append` does not dedupe) — the ring fully rotates in ~4 h, evicting every other captured error from `list_recent_errors` / `preview_bug_report` / `tm doctor`, and `errors.jsonl` grows unbounded. The last incident persisted three days. | Latch per session id (alarm on transition into `Destroyed`, then every Nth tick), mirroring the per-path streak map already in `prune.rs` (`CANONICALIZE_FAILURE_STREAK_THRESHOLD`). |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 256 | `Unknown` is emitted only at `warn!` — which this PR's own rationale says reaches no operator surface. If `git` is unavailable daemon-wide, every session reports `Unknown` and the detector is silently dark: the exact fail-open the module doc argues against. | Keep the per-finding `warn!`, but roll up at ERROR **once per sweep** when the candidate set is non-empty and *every* verdict is `Unknown` — that is "the detector is down", not the alarm-fatigue case the doc worries about. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/decommission.rs` | 383-399 | `ForeignActiveWorktree` has no override. A wrong or stale sentinel permanently bricks that record's decommission; the only recourse is `tm sessions delete` (which never removes the worktree) or a manual `rm`. With worktree count already a live problem, a misfire strands trees. | Add an explicit operator escape hatch (a `force` flag on the CLI/HTTP decommission path) that logs at ERROR and proceeds, or surface a "clear contested sentinel" action in `tm doctor`. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_ownership.rs` | 321-325 | The guard keys entirely on a file **inside** the worktree, which any agent in any session can write or delete. Absent it (pre-#3649 zero-byte sentinel, or a failed best-effort write — both write sites swallow the error) the guard is silently inert, on exactly the paths most likely to be legacy. | Add a second, unforgeable check alongside the sentinel: refuse when **another Active record's canonicalized `workspace_path` equals this one**. That is the #1744 incident shape verbatim, needs no on-disk file, and cannot be laundered by the impostor. |
| LOW | `crates/trusty-mpm/CHANGELOG.md` | 34 | The new `### Fixed` heading is inserted mid-list, so the pre-existing #2867 "Cross-branch `git push` guard" entry (line 65, previously under `### Added`) is now published under `### Fixed`, and `[Unreleased]` now carries two `### Fixed` sections (34 and 171). | Move the two new Fixed bullets into the existing `### Fixed` at line 171 and drop the new heading. |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/mod.rs` | 869 | `ForeignActiveWorktree` falls into the catch-all `Err(e) => 500`. It is a refusal, not a server fault; the sibling delete route already returns 409 for its refusal. | Add an arm mapping `ForeignActiveWorktree` to `StatusCode::CONFLICT`. |

## Required Changes

1. Use `resolve_ownerless` rather than `state == Active` in `foreign_active_owner`, so
   the `caller: None` guard is at least as strong as the `caller: Some` gate it sits
   beside. `decommission_allows_reclaim_when_peer_is_stopped` will then need to
   distinguish a genuinely abandoned Stopped owner from a resumable one.
2. Bound the `git` probe in `audit_worktree_integrity` with a timeout; `Unknown` on
   elapsed.
3. Stop the `Destroyed` alarm asserting the work is gone when the directory still holds
   files, and stop it telling the operator to recreate in that case.
4. Latch or rate-limit the per-tick alarm so one persistent finding cannot evict the
   error ring.

## Notes

- **Evidence for HIGH-1.** `decommission.rs:358-364` refuses when
  `!resolve_ownerless(owner)` — `false` for `Stopped`, so `caller: Some` refuses. The new
  guard at 383 refuses only for `Active`. Concrete reachable sequence in shipped code:
  `idle_reaper` stops peer P at the idle threshold; later it reaps colliding record I at
  the done threshold via `decommission(&I, None)`; the sentinel names P, P is `Stopped`,
  the guard passes, and `remove_session_worktree` destroys P's resumable tree. Same shape
  via `prune.rs:726`. Neither `idle_reaper` nor `decommission` carries the `#4118` dirt
  guard that `reap_aged_ephemeral` and the orphan sweep do. `prune.rs:912` (the adjacent
  sweep) already uses `resolve_ownerless_with_grace` for this same question and skips
  "still live/resumable" owners — so the tree currently disagrees with itself.
- **Evidence for HIGH-3**, reproduced: healthy worktree holding an uncommitted
  `precious.txt`; delete only `<base>/worktrees/wt` (the admin gitdir);
  `git -C <wt> rev-parse --show-toplevel` returns `fatal: not a git repository: (null)`
  which maps to `NotAWorkTree` and therefore `Destroyed`, with both `a.txt` and
  `precious.txt` still on disk. Not hypothetical for this repo: the `.base` bare clone
  was destroyed on 07-21 and orphaned ~70 worktrees whose contents were intact — every
  one would have alarmed "uncommitted work is GONE / recreate it".
- `probe_toplevel` mapping *any* non-zero exit to `NotAWorkTree` (line 166) is the root
  of HIGH-3. Distinguishing stderr containing `not a work tree` / `not a git repository`
  from other failures (dubious ownership, unreadable gitdir) and routing the rest to
  `Unknown` would also address it.
- Not flagged: the `ManagedError` move to `session_manager/error.rs` is clean — the enum
  is not `#[non_exhaustive]`, `manager.rs` re-exports it, `session_manager/mod.rs:125`
  still exports `ManagedError` from `manager`, and a missed exhaustive match would not
  compile. The only behavioural consequence is the 500-vs-409 nit above.
- Not flagged: the deliberate omission of automated forensic capture. The rationale in
  the runbook's section 5 is sound, and its blocking-I/O-in-the-sweep argument is the
  same one that produces HIGH-2 above.
- Not independently reproduced: the two `tm_hook_pm_guard` failures under `/private/tmp`
  and the `telegram::supervisor` flake. Both passed in my run from a compliant location,
  which is consistent with the disclosure; I did not build `main` under `/private/tmp` to
  confirm they pre-date this branch.

bobmatnyc:
All four HIGHs fixed, both LOWs taken, two of three MEDIUMs taken with a reasoned decline on the third. Pushed as `ea7f6138`. **CI is fully green, including the `Test` job (16m16s) that was pending at review time.**

Thank you for HIGH-1 in particular — it was the finding I most needed. My guard was strictly weaker than the gate twenty lines above it, in the direction that matters.

---

## HIGH-1 — guard now uses `resolve_ownerless`, not `state == Active`

You were right that the tree held two disagreeing notions of reclaimable and mine was the weak one. `foreign_owner` now takes the `resolve_ownerless` verdict, so `Stopped`/`Errored`/`Provisioning` all refuse and only a provably ownerless owner (terminal, or absent) reclaims. That matches both the #3649 gate and `prune.rs`'s sweep, so there is now one predicate rather than three.

**Also took the MEDIUM store-side check**, and it turned out to be more than a nice-to-have. `foreign_active_worktree_owner` now consults two independent sources, either of which refuses: the sentinel, and any OTHER record whose canonicalized `workspace_path` equals this one and is not ownerless. `decommission_refuses_when_store_shows_live_peer_without_sentinel` deletes the sentinel outright and still refuses — so the guard holds on exactly the legacy paths where, as you noted, the sentinel half is silently inert.

### One thing your fix surfaced that neither of us had accounted for

Wiring the store-side check broke two `dedup` tests:

```
reconcile_dedup_collapses_exact_workspace_duplicate_of_live_record ... FAILED
  the unowned phantom duplicate must be decommissioned
  left: Stopped   right: Decommissioned
```

`dedup`'s entire job is collapsing a dead duplicate that **by construction** shares a `workspace_path` with the record it deduplicates against — the #1744 collision shape verbatim, which is why my check caught it. But that collapse destroys nothing: the loser is unowned and not under `.worktrees/`, so `decommission` skips the filesystem entirely and only tombstones the record. The test even asserts `ws_dir.path().exists()`.

So the guard was over-refusing, and it was blocking *the very reconciliation it wants an operator to perform*. Fixed by scoping it to removals that actually touch the filesystem (`workspace_owned || is_session_worktree(ws)`), which makes it exactly coextensive with the harm it prevents. `decommission_allows_record_only_collapse_sharing_a_path` pins the boundary so the reasoning survives.

## HIGH-2 — probe bounded by the existing constant

`audit_worktree_integrity_with_timeout` wraps `spawn_blocking` in `tokio::time::timeout` using `GIT_WORKTREE_REMOVE_TIMEOUT` — the same knob, not a second one. Elapsed maps to `Unknown`, never a destruction verdict (a timeout is absence of evidence) and never `Intact`.

## HIGH-3 — `LinkageLost` split from `Destroyed`

Your reproduction was decisive, and the 07-21 precedent settled it: ~70 worktrees with contents fully intact, every one of which my alarm would have told the operator to recreate.

- `LinkageLost { detail, surviving }` — files on disk. The alarm now says *"its FILES ARE STILL ON DISK (N entries) — DO NOT decommission or recreate this session, which would delete it. Copy the directory aside first."*
- `Destroyed` — only when the directory is observed **empty**.
- An **unreadable** directory falls to `LinkageLost`, never `Destroyed`. Not having counted the files is never grounds for saying they are gone.
- Took your root-cause note too: `probe_toplevel` no longer maps every non-zero exit to `NotAWorkTree`. Only named phrases are definitive; everything else (dubious ownership, unreadable gitdir) is `Unavailable`/`Unknown`.

Writing that matcher caught a bug in my own first attempt — I matched `"not a work tree"`, but the bare-parent fatal is `"must be run in a work tree"`, which does not contain it. The real-git test failed and I widened it to all three observed phrasings.

The runbook gained a §4.0 that runs *before* the rest of triage: if the alarm was `LINKAGE LOST`, `cp -a` the tree out first.

## HIGH-4 — latched

Alarm on transition into the bad state, then once per `REALARM_EVERY_TICKS` (60 ≈ hourly). ~24 events/day instead of ~2,880, so one persistent finding cannot rotate the 500-entry ring. Sessions that recover are evicted, so a later break alarms immediately rather than inheriting a suppressed streak.

## MEDIUMs

- **`Unknown` only reaching `warn!` — taken.** An audit where there were candidates and **every** verdict is `Unknown` now alarms at ERROR once per sweep: *"worktree-integrity audit is DARK... no worktree is currently being verified."* A lone `Unknown` among healthy sessions stays quiet, so this does not reintroduce the fatigue HIGH-4 is about.
- **Sentinel forgeable — taken**, see HIGH-1 above.
- **Override path — declined, with reasoning.** A `force` flag would be reachable by the same automated `caller: None` callers this guard exists to stop, so it would reopen the hole for the exact population that caused it. There is already a working release valve: the guard releases as soon as the contesting owner is provably ownerless, and `tm sessions delete <owner-id>` marks a record `Deleted`, which is terminal. That is also the *correct* corrective action — reconciling the registry — rather than overriding a true signal. The error message now names the remedy inline (`tm ls`, then `tm sessions delete {owner}`), and the runbook documents it. If you still want a flag, I would rather do it as a follow-up gated to an interactive CLI path than expose it on `decommission`.

## LOWs — both taken

The `### Fixed` heading no longer reclassifies the pre-existing #2867 entry (my bullets moved into the existing `### Fixed`; `[Unreleased]` has one of each again), and `ForeignActiveWorktree` maps to `409 CONFLICT`.

---

## Mutation evidence — one per HIGH, reverted, observed failing, restored

| Mutation | Result |
|---|---|
| HIGH-1: revert to `state == Active` + disable store check | `decommission_refuses_to_delete_stopped_peer_worktree` **FAILED** — `state: Decommissioned, ..., true`, i.e. the Stopped peer's resumable tree deleted. `..._without_sentinel` also FAILED |
| HIGH-2: unbounded `join.await` | `audit_times_out_into_unknown_never_destroyed` **FAILED** — `left: 0, right: 1`, the probe silently passed as healthy |
| HIGH-3: collapse `LinkageLost` into `Destroyed` | 4 tests **FAILED**, incl. `linkage_lost_with_surviving_files_does_not_claim_work_gone`: *"files still on disk must NEVER be reported as Destroyed"* |
| HIGH-4: `record_bad` always returns `true` | all 3 latch tests **FAILED** — `left: 120, right: 3` alarms over 120 ticks |

## Gate — raw

```
cargo check -p trusty-mpm            Finished `dev` profile ... in 11.31s
cargo clippy --all-targets -D warn   EXIT=0
cargo fmt --all -- --check           EXIT=0
cargo test -p trusty-mpm             EXIT=0
  test result: ok. 3764 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
scripts/check_line_cap.sh            line-cap: 7 allowlisted, 0 violations — OK.
scripts/check_test_pointers.sh       test-pointers: 0 dangling pointers — OK.
CI                                   all checks pass, incl. Test (16m16s)
```

3764 tests passing, up from 3750. `check_test_pointers.sh` caught two pointers left dangling by my renames; fixed.

Back to you for re-review. Not merging.

bobmatnyc (commented):
## Verdict: WARN

Round-2 independent review of `ea7f6138`, from a fresh detached worktree at a
compliant location. All four round-1 HIGH fixes land and I verified every
mutation claim myself. Two new HIGH findings, both in surfaces this round
introduced or was asked to make absolute.

**Mutation claims — all four reproduced verbatim.**

| Mutation | Result |
|---|---|
| HIGH-1: `state == Active` + drop store check | `decommission_refuses_to_delete_stopped_peer_worktree` FAILED with `state: Decommissioned, … , true` — `workspace_removed = true`, the Stopped peer's tree destroyed. `…_without_sentinel` FAILED the same way. |
| HIGH-2: unbounded `join.await` | `left: 0, right: 1` |
| HIGH-3: collapse the split | 4 failures (`classify_surviving_files_is_linkage_lost`, `classify_unreadable_dir_…`, `classify_escaped_toplevel_…`, `linkage_lost_with_surviving_files_does_not_claim_work_gone`) |
| HIGH-4: always alarm | `left: 120, right: 3` |

**Green honesty confirmed.** `cargo test -p trusty-mpm` (no `--lib`, no filter):
`test result: ok. 3764 passed; 0 failed; 6 ignored` on the lib plus every
integration binary green, 6543 tests total, zero failures. No `#[ignore]`,
cfg-gate, or `--exclude` in any file this PR touches. All 36 new
`worktree_integrity` / `worktree_identity_guard` / `cwd_collision` tests present
and `… ok`. Both LOWs taken (single `### Fixed` under `[Unreleased]`; 409 arm at
`managed_routes/mod.rs:868`).

**The self-caught `NOT_A_WORKTREE_STDERR` bug: they got it right.** Reproduced
against git 2.54.0 on real worktrees in both layouts:

| shape | stderr / stdout | matched by |
|---|---|---|
| stripped, **bare** parent | `fatal: this operation must be run in a work tree` | `must be run in a work tree` — and **not** by `not a work tree`, so their first attempt genuinely would have missed it |
| admin gitdir deleted, files intact | `fatal: not a git repository: (null)` | `not a git repository` |
| corrupt `.git` pointer | `fatal: not a git repository: (null)` | `not a git repository` |
| stripped, **normal** parent | resolves to the parent dir on stdout | the escaped-`Resolved` branch, not stderr at all |

Over-match direction: a healthy worktree resolves cleanly under every config I
tried; a nonexistent root gives `cannot change to '<path>': No such file or
directory`, and dubious-ownership gives `detected dubious ownership in
repository at` — neither contains any of the three phrases, so both correctly
fall to `Unavailable`. The widening is tight in both directions.

**The dangling `Test:` pointers were genuinely their own.** `cwd_collision.rs`
is created whole by `0b1e4aba` in this PR, so `collision_ids_are_all_reported`
never existed outside it. The replacement pointer is also honest —
`alarm_is_error_level` really does assert both the level and that every
colliding id appears in the `sessions` field.

---

## Your three questions

### 1. The scoping IS coextensive with the harm. Verified, not taken on faith.

`decommission_with_root` mutates the filesystem in exactly two places
(`decommission.rs:445-531`): `!workspace_owned && is_session_worktree(ws)` →
`remove_session_worktree`, and `workspace_owned` → `remove_dir_all`. The guard's
predicate `record.workspace_owned || is_session_worktree(ws)` is the union of
those two, exactly.

The strongest independent evidence is that the predicate is not novel — it is
already the project's settled definition of "disposable workspace".
`search_gc::disposable_workspace_index_id` gates the *other* destructive side
effect in this same function with the literally identical expression:

```rust
if !(workspace_owned || is_session_worktree(ws)) { return None; }
```

So the guard, the filesystem removal, and the search-index deletion are now all
keyed on one predicate. They cannot drift apart.

**Probe B (doesn't touch files, should still be refused): I could not find one.**
The record-only branch's remaining side effects are `graceful_terminate_runtime`
and the tombstone. The former targets `record.tmux_name` — the *target's own*
tmux session, never the contesting peer's — so it is not a cross-session
destruction at all; the latter is registry state, reconcilable. No file, index,
or peer runtime is reachable from the unguarded branch.

**Probe A (touches files, should be allowed): the predicate is a deliberate
superset, and every over-fire is refusal-on-refusal.** It fires when
`workspace_owned` is true even where `is_safe_to_remove` would have refused
anyway, and under `.worktrees/` even where `remove_session_worktree`'s sentinel
gate would have refused. Both cases end in "nothing deleted" either way. Every
legitimate destructive path still passes: self
(`decommission_removes_own_worktree_despite_guard`), terminal peer
(`…_allows_reclaim_when_peer_is_terminal`), unknown owner.

One consequence worth stating out loud rather than flagging: `dedup` can no
longer collapse a workspace-duplicate whose path is under `.worktrees/`. That is
correct — that collapse would call `remove_session_worktree` on a live twin's
tree, i.e. #1744 — and it is not silent, because the guard `error!`s before
returning. The scoping buys back only the non-`.worktrees`, non-owned case,
which is precisely the case that deletes nothing.

**Verdict on the scoping: sound. "Does it touch files" is a faithful proxy here
because the same predicate already defines every destructive act in the
function.**

### 2. The force-flag decline is right, and the valve IS reachable — but the message that is supposed to name it doesn't.

I traced the valve end to end. `delete_record(id, force)` (`delete.rs:65`):
without `--force` on a live owner it returns `InvalidState` whose text itself
names `--force`; with `--force` it sets `state = Deleted`, which
`ManagedSessionState::is_terminal` accepts, which makes `resolve_ownerless`
return `true`, which releases both halves of the guard. Two hops, each one
self-documenting. **No operator gets wedged.** The reasoning for declining is
also correct on its own terms — a `force` flag would sit on the `caller: None`
path that every automated caller uses, which is the population that caused the
incident.

But the argument leans on discoverability, and that is where it does not hold —
see the MEDIUM below. The `#[error]` string is the only text an operator sees on
the HTTP 409 (`managed_routes/mod.rs:868` returns `e.to_string()`), the MCP
`session_decommission` tool (`mcp_session.rs:157` → `managed_err`), and sm-stdio
`kill` (`control.rs:221`). That string still says the sentinel names the owner
and that the owner is ACTIVE — both now false in the two cases this round added
— and it does not name `tm sessions delete`. The `tracing::error!` that does name
it goes to the daemon log, not to the caller.

**If you want an escape hatch anywhere, it belongs in that Display string, not
in a `force` flag.** Fixing the string is strictly better than adding the flag.

### 3. `Destroyed` is NOT absolute. This is HIGH-1 below.

Sub-answers, since you asked for each:

* **probe times out** → `Unknown`. Correct, and I proved it by mutation.
* **directory unreadable** → `LinkageLost`. Correct *only when the root itself
  canonicalized*. When the unreadability is at a parent component,
  `canonicalize` fails first and short-circuits to `Destroyed` before
  `count_surviving_entries` is ever consulted. `classify_unreadable_dir_is_linkage_lost_not_destroyed`
  covers the wrong half of this.
* **symlink** → a healthy symlinked workspace is fine; a symlink *loop* is
  `Destroyed`.
* **stalled mount** → if it hangs, `Unknown` (correct). If it *errors* —
  `ESTALE` after an NFS server restart is the classic, with every file intact on
  the server — `Destroyed`.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 305 | `check` does `root.canonicalize().ok()`, discarding the error **kind**, so every `canonicalize` failure — not just `NotFound` — short-circuits `classify` to `Destroyed("workspace root does not exist on disk")` **before `surviving` is consulted at all**. A fully intact worktree therefore reaches the alarm that says *"Its uncommitted work is gone. Stop the session and recreate it"* — the exact HIGH-3 harm, via the one branch the HIGH-3 fix does not cover. | Match on the kind: `Err(e) if e.kind() == NotFound => None` (keeps `absent_workspace_root_is_destroyed` green and keeps `classify(None, …) = Destroyed` meaning *confirmed absent*); every other `Err(e)` → `return WorktreeIntegrity::Unknown(format!("cannot canonicalize the workspace root ({e}) — not evidence of destruction"))`. |
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_integrity.rs` | 597 | The detector-is-dark roll-up is the one ERROR site in this module with **no latch**. `damaged` is filtered through `alarming` (latched); `all_unknown` is computed from `all` and fires unconditionally. `git` missing from the daemon's PATH makes every probe fail instantly, so at the 60 s `ORPHAN_GC_INTERVAL_SECS` cadence this emits ~1 440 ERROR/day into the same 500-entry, `pop_front`-on-overflow, non-deduping ring (`error_capture/store.rs:144`) — full rotation in ~8 h, evicting everything else from `list_recent_errors` / `preview_bug_report` / `tm doctor`. That is HIGH-4's mechanism and HIGH-4's consequence, unfixed, three lines from the fix. | Route the dark roll-up through the same `AlarmLatch`, keyed on a reserved sentinel id (or a sibling `u32` counter on the latch). Alarm on transition into dark, then every `REALARM_EVERY_TICKS`. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/error.rs` | 164 | The `#[error]` string was not updated for either behaviour this round added. It asserts *"its on-disk sentinel names session {1}, still ACTIVE"* — but `decommission_refuses_when_store_shows_live_peer_without_sentinel` deletes the sentinel outright and still refuses (so there is no sentinel to inspect), and `decommission_refuses_to_delete_stopped_peer_worktree` refuses for a **Stopped** peer (so "still ACTIVE" is false). It also does not name `tm sessions delete <owner-id>`, though the doc comment 12 lines above claims *"The message states this so the recourse is visible at the point of refusal."* This is the only text the HTTP/MCP/stdio caller sees, and it both misdirects the operator to a nonexistent file and withholds the valve the force-flag decline rests on. | Rewrite to match the guard: `"refusing to remove worktree {2} while decommissioning session {0}: it belongs to session {1}, which is still live or resumable. No files were touched. Reconcile with 'tm ls', then 'tm sessions delete {1}' (add --force if it is still running) if that record is stale (#3764)"`. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_ownership.rs` | 387, 395 | `colliding_live_record` calls blocking `std::fs::canonicalize` on the tokio runtime — once for `path`, then once per record in `self.list()` — with no `spawn_blocking` and no timeout. One record whose `workspace_path` is on a stalled mount blocks a worker thread for **every** decommission, not just its own. This PR just added `spawn_blocking` + `GIT_WORKTREE_REMOVE_TIMEOUT` for exactly this hazard on the audit path, and `remove_session_worktree` 60 lines away is already wrapped the same way; this new loop is the one unbounded blocking I/O site left on the decommission path. | Wrap the whole scan in `tokio::task::spawn_blocking` + `tokio::time::timeout(GIT_WORKTREE_REMOVE_TIMEOUT, …)`. On elapsed, fail **closed** (return the first colliding record id) — an unverifiable collision check must not become an implicit permission to delete. |

## Required Changes

1. Distinguish `canonicalize`'s `NotFound` from every other error kind in
   `check`; non-`NotFound` → `Unknown`, never `Destroyed`. Add a test in
   `worktree_integrity_tests.rs` pinning it — a healthy populated worktree whose
   parent is `chmod 000` must not be `Destroyed`.
2. Latch the `all_unknown` roll-up in `alarm_rollup` the same way the per-finding
   alarms are latched.

## Notes

* **HIGH-1 is reproduced, not argued.** I appended two probe tests to
  `worktree_integrity_tests.rs` in my own worktree and ran them against
  unmodified `ea7f6138`:

  ```
  canonicalize error kind = PermissionDenied
  verdict = Destroyed("workspace root does not exist on disk (cannot canonicalize)")
  critic_probe_eacces_parent_must_not_be_destroyed ... FAILED

  ELOOP verdict = Destroyed("workspace root does not exist on disk (cannot canonicalize)")
  critic_probe_eloop_must_not_be_destroyed ... FAILED
  ```

  The EACCES case is a real `git worktree add` holding `a.txt`, `precious.txt`
  and a valid `.git` pointer, with only its `.worktrees` parent set to mode
  `000`; the assertion that failed is `wt.join("precious.txt").exists()` passing
  immediately before the verdict. Independently confirmed at the syscall level:
  `PermissionDenied (os error 13)` and `FilesystemLoop (os error 62)` from
  `std::fs::canonicalize`. My probe tests were reverted; the tree is clean.

* **Frequency caveat, stated honestly.** Neither HIGH is an every-day trigger.
  HIGH-1 needs a permission change, a symlink loop, or an `ESTALE`/`EIO`
  filesystem; HIGH-2 needs `git` off the daemon PATH. I am flagging both at HIGH
  because the *mechanism* is certain (both are readable straight off the code and
  one is reproduced), because both are self-inflicted regressions of fixes made
  in this very commit, and because both fixes are a handful of lines. Discount
  the likelihood as you see fit — do not discount the diagnosis.

* **Not flagged.** `ALARM_LATCH` being a process-global `OnceLock` means the
  audit integration tests share it, but every one of them asserts on the returned
  `findings` (computed before the latch), so there is no cross-test dependence
  today. Worth knowing if someone later writes a test that asserts on emitted
  alarms through the audit.

* **Not flagged.** `colliding_live_record`'s canonicalize fallback
  (`unwrap_or_else(|_| path.to_path_buf())`) can make two records with the same
  logical path compare unequal if exactly one of them resolves. It matches
  `correlate_session_start`'s comparison deliberately, and the failure direction
  is under-refusal in a case the sentinel half already covers. Below my
  confidence bar for a finding.

* **Not flagged.** The `prune.rs` change is a runbook pointer on an existing
  alarm; the `ManagedError` module move remains clean.

* Neither HIGH is a regression against `main` — `main` has no detector at all.
  Both are gaps in the new surface, and neither blocks the guard half of this PR,
  which is the part that closes #3764's actual incident shape.

bobmatnyc:
Folding this into #4207 (git-native worktree rebuild), not rejecting it.

At round 3, this PR's own round-1 fix introduced two NEW HIGH findings: `worktree_integrity.rs:305` collapses every io error kind to `Destroyed` via `canonicalize().ok()`, and the dark-detector alarm at :597 has no latch — both reintroduced the harm they were meant to fix. Rather than patch a fourth time, #4207 replaces this code path outright: derive worktree state from `git worktree list --porcelain` and a git-derived ownership predicate instead of filesystem heuristics.

#3764 (the issue this PR closes) stays OPEN — the hardening need is real, now carried by #4207.

Closing as folded.

bobmatnyc:
Folded into #4207.

bobmatnyc:
## Round-2 fixes pushed to the branch — informational, NOT a request to reopen

I was dispatched to fix the round-2 findings and hand back for round-3 review. That
dispatch predates this PR being closed at 03:11Z and folded into #4207. The work was
already done when I saw the closure, so I am recording it here rather than dropping it.
**I have not reopened the PR and am not asking for a round-3 review** — the fold
decision stands. The branch `fix/3764-worktree-hardening` now carries the fixes at
`26f57727` (rebased onto `5bf231b9`, so it is no longer BEHIND) if any of it is worth
salvaging into #4207's slices 1–2.

### Per finding

**HIGH-1 — `worktree_integrity.rs:305`, `canonicalize().ok()` collapses every error
kind to `Destroyed`.** `check` now matches on the kind: `Ok` → resolved,
`ErrorKind::NotFound` → `None` (so `classify(None, …)` keeps its meaning, *confirmed
absent*), every other kind → `Unknown` immediately. EACCES on a parent, ELOOP, ENOTDIR,
EINTR, and ESTALE no longer short-circuit past the surviving-entry count.
`absent_workspace_root_is_destroyed` stays green.

**HIGH-2 — `worktree_integrity.rs:597`, dark roll-up unlatched.** Routed through the
same `AlarmLatch` as the per-finding alarms, via a sibling `dark_ticks: u32` counter and
the same `REALARM_EVERY_TICKS` policy — one latch mechanism in this module, not two.
`alarm_rollup` now takes `&mut AlarmLatch`; the audit holds one guard across
`emit_finding_alarm` and the roll-up (no `.await` in between). A non-dark sweep resets
the streak.

**MEDIUM — `error.rs:164`, stale `#[error]` text.** Rewritten. It no longer points the
operator at a sentinel that may not exist, no longer asserts the owner is ACTIVE (false
for the Stopped-peer refusal this round added), and now names the valve:
`tm sessions delete <owner-id>` (`--force` if still running). This is the only text the
HTTP 409 / MCP `session_decommission` / sm-stdio `kill` caller sees.

**MEDIUM — `worktree_ownership.rs:387,395`, unbounded blocking `canonicalize`.** The
pattern does exist in this crate, so I applied it rather than declining: the whole scan
is now one batched `spawn_blocking` under `tokio::time::timeout(GIT_WORKTREE_REMOVE_TIMEOUT)`,
matching `remove_session_worktree` and the integrity probe. On elapse or task failure it
fails **closed** — every candidate returns un-ruled-out, and the caller still filters by
`resolve_ownerless`, so a fail-closed scan can only ever refuse in favour of a genuinely
live or resumable record. The `#1744` shape stays enforced; the timeout bounds the WAIT,
not the syscall (a wedged `canonicalize` keeps its blocking-pool thread but no longer a
runtime worker).

### Regression tests — each fails against `ea7f6138`, verified by mutation

| Test | Mutation applied | Result against the mutation |
|---|---|---|
| `symlink_loop_root_is_unknown_not_destroyed` | restore `canonicalize().ok()` | FAILED — `verdict = Destroyed("workspace root does not exist on disk (cannot canonicalize)")`, kind `FilesystemLoop` |
| `unreadable_parent_never_reports_intact_worktree_as_destroyed` | same | FAILED (your EACCES reproduction, `precious.txt` intact) |
| `dark_rollup_alarm_is_latched` | alarm unconditionally on `all_unknown` | FAILED — `left: 120, right: 3` over 120 sweeps |
| `collision_scan_fails_closed_when_it_times_out` | return `Vec::new()` on elapse | FAILED — `left: [], right: [<both ids>]` |

The symlink-loop case is deliberately the *primary* HIGH-1 test rather than the
`chmod 000` one: it produces a non-`NotFound` kind deterministically for every uid,
including root. The EACCES reproduction is there too, plus
`enoent_root_is_still_destroyed` as the other half — routing every error kind to
`Unknown` would make `Destroyed`-on-absent unreachable, and that mutation is caught.

The `collision_scan_fails_closed_when_it_times_out` bound is injected through a small
`bounded_collision_scan` seam (same shape as `audit_worktree_integrity_with_timeout`)
because a 1 ns timer loses the race against a fast `canonicalize` — with the naive
version the test passed vacuously on the first run, which is exactly the failure mode
you would have mutation-tested for.

### Gate

Rebased onto `origin/main` first, so this is the real merge result, not stale green.

```
cargo test --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui
  → 19750 passed; 0 failed; 132 ignored across 194 test binaries
  → trusty-mpm lib: test result: ok. 3778 passed; 0 failed; 6 ignored
cargo clippy --workspace --all-targets (same exclusions) -- -D warnings  → exit 0
cargo fmt --all --check                                                  → exit 0
scripts/check_line_cap.sh        → 7 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh   → 0 dangling pointers — OK
scripts/check_sld.sh             → 0 error(s), 0 warning(s)
```

No `#[ignore]`, cfg-gate, or `--lib` narrowing was used to reach that. The alarm-latch
and roll-up tests moved from `worktree_integrity`'s inline `mod tests` into
`worktree_integrity_tests.rs` — names unchanged, all still `... ok` — because the fixes
pushed the production file to 508 SLOC against the 500 cap. The pure `classify_*` matrix
stays inline.

One disclosure: the first full run had `trusty-agents`
`tools::python_skill::tests::execute_dispatches_to_python_and_parses_json` fail with
`Failed to inspect Python interpreter … /opt/homebrew/bin/python3`. It passed on re-run,
and `git diff origin/main...HEAD -- crates/trusty-agents` is empty — a local `uv`
toolchain flake, not this branch.

Issue #1931, opened 2026-07-02
tm managed-spawn deploy targets real $HOME/.claude instead of project-local worktree, breaking tm-* skill discovery


bobmatnyc:
## Summary

Two live regressions observed by the operator right after installing the `#1927`/`#1930` isolation-drift fix (merge `81e1924f`):

1. Bare `tm` (guided default) appeared not to detect the already-cloned managed repo for `bobmatnyc/trusty-tools`.
2. Inside a spawned Claude Code session, `/tm-` shows "No commands match" — the `tm-*` skills are not loaded by the harness, despite a prior deploy claiming success.

## Phase 1 diagnosis (live-state evidence)

**Symptom 1 — turned out to be an operational issue, not a code bug.** The daemon actually serving `tm` requests on port 7880 (`com.trusty.mpm`, pid 95673 at diagnosis time) had been started at `15:16:56`, i.e. **before** the `~/.cargo/bin/tm`/`trusty-mpm` binaries were rebuilt at `17:08` with the `81e1924f` fix. The running daemon process was still executing pre-rebuild code in memory (classic "daemon needs restart after `cargo install`" gotcha, documented in `docs/reference/release-workflow.md` / CLAUDE.md issue #534 convention). There was also an orphaned stray `tm daemon` foreground process (pid 24467, started 09:54AM) not holding the port.

After a graceful `launchctl bootout` + `bootstrap` restart of `com.trusty.mpm` (picking up the actual 17:08 build), re-running the bare `tm` guided flow from `/Users/masa/Projects/trusty-tools` correctly detected the existing managed workspace:

```
tm: no subcommand — using guided default for /Users/masa/Projects/trusty-tools
tm: project:   bobmatnyc/trusty-tools
tm: workspace: /Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools (live checkout is NOT touched)
tm: sessions:
tm:   [1] tmpm-trusty-tools-2dfb23d1  state=stopped  last=—
tm:   [2] tmpm-trusty-tools-e3a97212  state=stopped  last=—
```

No duplicate clone was created. **No code fix needed for symptom 1** — daemon restart resolved it. Filing this fact here for the record since the operator report suspected #1927/#1930 caused it.

[...]

Issue #1269, opened 2026-06-15
bug(tm): spawned Claude Code session blocked by trust/MCP dialog — task keystrokes lost (no --print / no interactive wait)


bob-duetto:
**Severity: Critical — blocks unattended autonomous execution.** `tm ticket` / `tm watch poll --execute` spawn a tmux session running `env -u ANTHROPIC_API_KEY claude` and immediately `send_line` the task text. Claude Code v2.1.178 shows "Is this a project you trust?" then an MCP-server selection dialog before becoming interactive. The injected `send_line` keystrokes are consumed by those dialogs, not Claude's input buffer, so the task is never delivered. Reproduced identically across 3 spawned sessions (1 ticket + 2 watch). No PR produced.

Expected: spawn must either (a) use Claude Code's non-interactive `--print`/headless task-injection path, (b) pre-trust the workspace (e.g. `--dangerously-skip-permissions` equivalent or write the trust state) and pre-select MCP servers, and/or (c) wait until Claude is actually interactive before injecting the prompt. Found with `trusty-mpm 0.8.0` + Claude Code v2.1.178.

Issue #3427, opened 2026-07-20
Harness scaffolding files tracked in git block merge --ff-only


bob-duetto:
## What Happened

trusty-mpm writes harness scaffolding files into the target project's working tree at paths that the project then commits to git. When the harness later regenerates those same files as untracked local content, `git merge --ff-only origin/main` aborts with "would be overwritten by merge" errors. The working tree becomes unable to fast-forward until manual intervention.

## Expected Behavior

The harness should either:
1. Prevent scaffolding files from entering project git history, OR
2. Detect the conflict at session start and provide exact remediation steps, OR
3. Store harness state outside the project tree

After remediation, `git merge --ff-only origin/main` should succeed cleanly.

## Steps to Reproduce

1. Create a new trusty-mpm project
2. Run agent or skill onboarding that generates `.claude/skills/**`, `.claude/agents/**`, or `.claude/output-styles/**` files
3. Commit those scaffolding files to git (either intentionally or by accident in a PR)
4. Later: harness regenerates the same files locally as untracked content
5. Attempt: `git merge --ff-only origin/main`
6. **Observe failure**: "error: Your local changes to the following files would be overwritten by merge" + list of tracked harness files

## Context

**Verified reproduction:**
- Project: `duettoresearch/duetto-eve-agents`
- Worktree: `tm-duetto-eve-agents-01`
- Symptom: stranded 28 commits behind origin/main
- True collision set (7 files): `.claude/skills/documentation-style/` (SKILL.md + 6 reference files)
- These files were BOTH tracked on origin/main AND regenerated locally as untracked by the harness

**Error output was misleading:**
The merge error listed many more files than the actual intersection (git's "would be overwritten" message lists all untracked content in the affected tree), inviting over-broad cleanup.

**Resolution required (duetto-eve-agents#111, commit 0e699e2):**
- `git rm -r --cached .claude/skills/documentation-style/`
- Add `.gitignore` entries for harness scaffolding paths
- Commit cleanup PR to reset tracking state

## Impact

**Scope:** Any project that has ever committed `.claude/skills/**`, `.claude/agents/**`, or `.claude/output-styles/**` will hit this the moment the harness regenerates those files.

**Severity:** Silently strands working tree; blocks CI/CD workflow until human intervention; misdiagnosis common (error message breadth suggests the problem is much larger than it is).

**Affected users:** Anyone using trusty-mpm with agents or skills who checks in scaffolding directories.

## Suggested Fixes

[...]

PR #4205, opened 2026-07-28
fix(trusty-mpm): deploy launch/connect/meta-launch agents into a tier --setting-sources loads

### LINKED ISSUE #4203: tm launch / tm connect deploy agents to a tier --setting-sources excludes — no tm-deployed agent loads
DEFECT: `tm launch` (`launch.rs:208`) and `tm connect` (`launch.rs:470`) deploy agents via `FrameworkPaths::default()`, which resolves to `~/.claude/agents` (the USER tier). Both then spawn the harness with `--setting-sources project,local` (`model_inject.rs:70`). That flag list EXCLUDES the user tier the deploy step just populated, so no tm-deployed agent is loaded by the session on either CLI path.

EVIDENCE:
- `crates/trusty-mpm/src/bin/tm/commands/launch.rs:208` (inside `launch`, in-project worktree spawn): `let fw = trusty_mpm::core::paths::FrameworkPaths::default();` then `prepare_session_with_repo_url(&fw, &managed_path, ...)`.
- `crates/trusty-mpm/src/bin/tm/commands/launch.rs:470` (inside `connect`): `let fw = trusty_mpm::core::paths::FrameworkPaths::default();` then `prepare_session(&fw, &path)`.
- `FrameworkPaths::default()` resolves against `dirs::home_dir()` (`paths.rs:108-111`), so `fw.claude_agents_dir()` is `~/.claude/agents` — confirmed by the `claude_agents_dir_is_dotclaude_agents` test (`paths.rs:790-798`), which asserts `FrameworkPaths::under("/base").claude_agents_dir() == "/base/.claude/agents"`.
- `session_launch::prepare_session_inner` deploys the agent roster with `deploy_agents_filtered(&plan.agent_source, &fw.claude_agents_dir(), ...)` (`session_launch/mod.rs:413`) — i.e. into whatever base `fw` was constructed against, NOT into `project_dir`.
- `crates/trusty-mpm/src/core/model_inject.rs:70`: `pub const SETTING_SOURCES_FLAG: &str = "--setting-sources project,local";` — applied to every trusty-mpm-spawned `claude` process, explicitly excluding `user`.
- Contrast with the daemon in-project spawn path, which got this right for the same reason (issue #1931): `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs:963` constructs `let fw = crate::core::paths::FrameworkPaths::for_managed_workspace(&worktree);` with the comment "use `for_managed_workspace(&worktree)`, NOT `default()` — the harness cwd for this session IS `worktree`, so deployed agents/skills must land in `<worktree>/.claude/{agents,skills}`". `tm launch`'s in-project path and `tm connect` never received the equivalent fix.

RESOLUTION: point both call sites (`launch.rs:208`, `launch.rs:470`) at a project-tier `FrameworkPaths` (e.g. `FrameworkPaths::for_managed_workspace(&managed_path)` / `FrameworkPaths::for_managed_workspace(&path)`) instead of `FrameworkPaths::default()`, so deployment lands in the tier `--setting-sources project,local` actually reads — matching the pattern already fixed in the daemon's `spawn_managed_inproject` (#1931).

IMPACT NOTE: this is also why threading the launch mode into the agent roster was unavailable, so #4196 shipped a caveat instead. See #4196 and #4069.

bobmatnyc:
## The defect

`tm launch` and `tm connect` deployed their agent roster into a settings tier the session they spawn does not read, so **no tm-deployed agent was ever loaded on either CLI path**.

Each link verified against `origin/main` before changing anything:

| Link | Evidence |
|---|---|
| Both call sites pass `FrameworkPaths::default()` | `launch.rs:208` (in-project worktree spawn), `launch.rs:470` (`connect`) |
| `default()` resolves to the USER tier | `paths.rs:108-111` resolves against `dirs::home_dir()`; `claude_agents_dir()` is `<base>/.claude/agents` per `claude_agents_dir_is_dotclaude_agents` (`paths.rs:790`) |
| The deploy really targets `fw`, not the project dir | `session_launch/mod.rs:508` — `deploy_agents_filtered(&plan.agent_source, &fw.claude_agents_dir(), …)` |
| `SETTING_SOURCES_FLAG` reaches the spawn on **both** paths | `launch` → `launch.rs:322` `build_claude_command(...)`; `connect` → `connect_claude_cmd` → `build_claude_command`. `build_claude_command` (`model_inject.rs:148-152`) *always* appends the flag — it is not conditional |

The deployed tier (`user`) and the enumerated setting sources (`project,local`) do not intersect. The `user` tier is excluded deliberately (#1269), so the flag is correct and the deploy destination is the bug.

## The fix

Both call sites now route through a new `session_framework_paths` seam backed by `FrameworkPaths::for_managed_workspace(<harness cwd>)` — the managed worktree for `launch`, the live checkout for `connect`. The payload lands in `<cwd>/.claude/{agents,skills}`, which the `project` and `local` tiers read. Deploy DESTINATIONS move workspace-local while the framework SOURCE paths keep resolving from the home-relative install root.

This is the established pattern, not a new one: the daemon's `spawn_managed_inproject` already fixed this exact bug for its own in-project spawn (`lifecycle.rs:963`, citing #1931). The two CLI paths never received the equivalent change.

## Regression test

`launch_and_connect_deploy_into_a_tier_setting_sources_loads` (`tests_behavior_b_tests.rs`).

It asserts the **relationship** — the deployed tier must be an element of the tiers the flag names — not a hardcoded path string that would pass vacuously if either side changed. Both inputs are derived from production code: the tier list is parsed out of `SETTING_SOURCES_FLAG` itself (`setting_source_tiers()`), and the destination comes from the same `session_framework_paths` seam both commands call. Neither `#[ignore]` nor any cfg-gate.

Verified failing against the pre-fix behavior by reverting only the seam's body to `FrameworkPaths::default()`:

```
test tests_behavior_b::launch_and_connect_deploy_into_a_tier_setting_sources_loads ... FAILED

agents deploy into the `user` tier (/Users/masa/.claude/agents) but the spawned session
carries `--setting-sources project,local`, loading only ["project", "local"] — no
tm-deployed agent would be visible to the session (issue #4203)

test result: FAILED. 0 passed; 1 failed; 0 ignored
```

Three supporting unit tests cover the new pure helpers: `setting_source_tiers_parses_the_flag`, `settings_tier_of_names_project_for_the_harness_cwd`, `settings_tier_of_names_user_for_anything_else`.

## Gate

Full workspace suite (no `--lib`), matching CI's own exclusion list for the three Tauri GUI crates that need a prebuilt frontend:

```
cargo test --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui
  → exit 0 — 19706 passed; 0 failed; 132 ignored across 194 test binaries

cargo clippy --workspace --all-targets --exclude <same three> -- -D warnings  → exit 0
cargo fmt --check                                                             → exit 0
scripts/check_line_cap.sh   → line-cap: 7 allowlisted, 0 violations — OK.
```

All four new tests confirmed present as `... ok` in the run, not silently skipped. No files under `assets/agents/` are touched, so `check_agent_assets.sh` is not implicated.

Closes #4203

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -135,6 +135,24 @@ so a 2.0.0 would signal a break that cannot reach anyone.
 
 ### Fixed
 
+- **`tm launch` and `tm connect` deployed agents to a tier `--setting-sources`
+  excludes, so no tm-deployed agent ever loaded**
+  ([#4203](https://github.com/bobmatnyc/trusty-tools/issues/4203)): both CLI
+  paths built their `FrameworkPaths` with `default()`, which resolves against
+  `dirs::home_dir()` and writes the agent roster into `$HOME/.claude/agents` —
+  the `user` tier. Both then spawned `claude` carrying
+  `--setting-sources project,local`, a list that deliberately excludes `user`
+  (#1269). The deployed tier and the loaded tiers never intersected, so every
+  agent the launch step had just deployed was invisible to the session it was
+  deployed for. Both call sites now use
+  `FrameworkPaths::for_managed_workspace(<harness cwd>)` — the managed worktree
+  for `launch`, the live checkout for `connect` — landing the payload in
+  `<cwd>/.claude/{agents,skills}`, which the `project` and `local` tiers read.
+  This is the same fix the daemon's `spawn_managed_inproject` already carried
+  for the identical defect (#1931); the two CLI paths never received it. A
+  regression test asserts the RELATIONSHIP (deployed tier ∈ the tiers the flag
+  names), not a literal path, so it keeps biting if either side moves.
+
 - **Two wrong instructions corrected in the bundled assets composed into every
   session's prompt.** Both were static text that no project could override, and
   both had demonstrably misrouted live sessions.
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,6 +7,39 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Fixed
+
+- **`tm launch`, `tm connect`, and `tm meta launch` deployed agents to a tier
+  `--setting-sources` excludes, so no tm-deployed agent ever loaded**
+  ([#4203](https://github.com/bobmatnyc/trusty-tools/issues/4203)): all three
+  CLI paths built their `FrameworkPaths` with `default()`, which resolves
+  against `dirs::home_dir()` and writes the agent roster into
+  `$HOME/.claude/agents` — the `user` tier. All three then spawned `claude`
+  carrying `--setting-sources project,local`, a list that deliberately excludes
+  `user` (#1269). The deployed tier and the loaded tiers never intersected, so
+  every agent the launch step had just deployed was invisible to the session it
+  was deployed for, and nothing reported it: the deploy genuinely succeeded and
+  the load silently found nothing.
+
+  These paths now deploy through a new `session_launch::prepare_isolated_session`
+  entry point, which resolves the layout itself rather than accepting one from
+  the caller — the harness cwd is the deploy base, so `<cwd>/.claude/{agents,skills}`
+  lands in a tier the `project`/`local` sources read. Removing the parameter
+  removes the defect: there is no longer a wrong `FrameworkPaths` for a call
+  site to pass. This is the same correction the daemon's `spawn_managed_inproject`
+  already carried (#1931); the three CLI paths never received it. `tm session
+  start` is deliberately unchanged — it spawns a bare `claude` with no
+  `--setting-sources`, so it does read the user tier and `default()` is correct
+  there.
+
+  **Operator-visible change:** `tm connect` now writes the agent roster, skills,
+  and output styles into the `.claude/` directory of your **live checkout**
+  rather than `$HOME/.claude/`, so they will show up in `git status` for that
+  repo. The scaffolding paths are added to `.gitignore` automatically (#3427),
+  and existing hand-placed or committed `.claude/` content is never overwritten
+  — unmanaged files are skipped as user-owned, and managed files whose checksum
+  drifted are skipped with a warning.
+
 ---
 ## [1.2.0] — 2026-07-27
 
@@ -134,24 +167,6 @@ so a 2.0.0 would signal a break that cannot reach anyone.
   ordering-dependent behaviour is reproducible.
 
 ### Fixed
-
-- **`tm launch` and `tm connect` deployed agents to a tier `--setting-sources`
-  excludes, so no tm-deployed agent ever loaded**
-  ([#4203](https://github.com/bobmatnyc/trusty-tools/issues/4203)): both CLI
-  paths built their `FrameworkPaths` with `default()`, which resolves against
-  `dirs::home_dir()` and writes the agent roster into `$HOME/.claude/agents` —
-  the `user` tier. Both then spawned `claude` carrying
-  `--setting-sources project,local`, a list that deliberately excludes `user`
-  (#1269). The deployed tier and the loaded tiers never intersected, so every
-  agent the launch step had just deployed was invisible to the session it was
-  deployed for. Both call sites now use
-  `FrameworkPaths::for_managed_workspace(<harness cwd>)` — the managed worktree
-  for `launch`, the live checkout for `connect` — landing the payload in
-  `<cwd>/.claude/{agents,skills}`, which the `project` and `local` tiers read.
-  This is the same fix the daemon's `spawn_managed_inproject` already carried
-  for the identical defect (#1931); the two CLI paths never received it. A
-  regression test asserts the RELATIONSHIP (deployed tier ∈ the tiers the flag
-  names), not a literal path, so it keeps biting if either side moves.
 
 - **Two wrong instructions corrected in the bundled assets composed into every
   session's prompt.** Both were static text that no project could override, and
--- a/crates/trusty-mpm/src/bin/tm/commands/launch.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/launch.rs
@@ -204,12 +204,12 @@ pub(crate) async fn launch(
     // 7c. Deploy the `.claude` framework into the worktree (best-effort).
     //     Non-fatal: a deploy failure never aborts the session — the operator can
     //     run `tm install` / `tm catalog sync` to populate agents manually.
-    //     #4203: the destination is the worktree (the harness cwd, set at step
-    //     12), NOT `$HOME` — see [`session_framework_paths`].
+    //     #4203: deploy through the isolated seam so the roster lands in the
+    //     worktree (the harness cwd, set at step 12) rather than `$HOME` — the
+    //     `claude` spawned at step 13 carries `--setting-sources project,local`
+    //     and would never read a `$HOME/.claude` deploy.
     {
-        let fw = session_framework_paths(&managed_path);
-        match trusty_mpm::core::session_launch::prepare_session_with_repo_url(
-            &fw,
+        match trusty_mpm::core::session_launch::prepare_isolated_session(
             &managed_path,
             Some(&origin_url),
         ) {
@@ -469,10 +469,11 @@ pub(crate) async fn connect(
     //     under `path` — so a repo with no GitHub remote or with uncommitted
     //     changes still launches. Best-effort: a prep failure is logged and
     //     never aborts the connect.
-    //     #4203: the destination is the live checkout (the harness cwd, set at
-    //     step 4), NOT `$HOME` — see [`session_framework_paths`].
-    let fw = session_framework_paths(&path);
-    match trusty_mpm::core::session_launch::prepare_session(&fw, &path) {
+    //     #4203: deploy through the isolated seam so the roster lands in the
+    //     live checkout (the harness cwd, set at step 4) rather than `$HOME` —
+    //     the `claude` spawned at step 5 carries `--setting-sources
+    //     project,local` and would never read a `$HOME/.claude` deploy.
+    match trusty_mpm::core::session_launch::prepare_isolated_session(&path, None) {
         Ok(report) => {
             for err in &report.roster_errors {
                 tracing::error!("roster provisioning gap for {}: {err}", path.display());
@@ -595,33 +596,6 @@ pub(crate) async fn connect(
         g.disarm();
     }
     Ok(())
-}
-
-/// Resolve the [`FrameworkPaths`](trusty_mpm::core::paths::FrameworkPaths) a
-/// CLI-spawned session deploys its `.claude` payload through.
-///
-/// Why (issue #4203): both `launch` and `connect` used
-/// `FrameworkPaths::default()`, which resolves against `dirs::home_dir()` and
-/// therefore deploys the agent roster into `$HOME/.claude/agents` — the `user`
-/// tier. Both then spawn `claude` carrying
-/// [`SETTING_SOURCES_FLAG`](trusty_mpm::core::model_inject::SETTING_SOURCES_FLAG)
-/// (`--setting-sources project,local`), which EXCLUDES `user`. The deployed
-/// tier and the loaded tiers did not intersect, so no tm-deployed agent was
-/// ever visible to either CLI session. This is the same defect the daemon's
-/// `spawn_managed_inproject` already fixed for its own in-project spawn
-/// (issue #1931, `daemon/managed_routes/lifecycle.rs`), applied to the two CLI
-/// paths that never received the equivalent change.
-/// What: `FrameworkPaths::for_managed_workspace(harness_cwd)` — deploy
-/// DESTINATIONS become `<harness_cwd>/.claude/{agents,skills}` (read by the
-/// `project`/`local` tiers) while the framework SOURCE paths keep resolving
-/// from the home-relative install root, exactly as #1931 guarantees. Callers
-/// must pass the cwd the tmux pane is rooted at: the managed worktree for
-/// `launch`, the live checkout for `connect`.
-/// Test: `launch_and_connect_deploy_into_a_tier_setting_sources_loads`.
-pub(crate) fn session_framework_paths(
-    harness_cwd: &std::path::Path,
-) -> trusty_mpm::core::paths::FrameworkPaths {
-    trusty_mpm::core::paths::FrameworkPaths::for_managed_workspace(harness_cwd)
 }
 
 /// Compose the `claude` invocation `connect` sends to a freshly-created tmux pane.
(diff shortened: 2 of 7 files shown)


bobmatnyc (commented):
## Verdict: WARN

Adversarial review, independent of the implementer. The **production fix is correct** — I verified the tier mapping mechanically rather than trusting the naming, and it holds. The HIGH finding is not in the fix; it is that the regression guard shipped with it **does not guard the fix**, which I proved by mutation.

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/tests_behavior_b_tests.rs` | 1688–1730 | The regression test never touches either production call site. Reverting **both** call sites to `FrameworkPaths::default()` — reintroducing #4203 verbatim — leaves the entire tm-binary suite green. | Bind the invariant to the deploy seam, not to a helper (see Required Changes). |
| MEDIUM | `crates/trusty-mpm/src/core/model_inject.rs` | 80, 108 | `setting_source_tiers()` and `settings_tier_of()` are `pub` library API with **zero production callers** — test-only code in a production module. `settings_tier_of` also has a wrong contract: it returns `"user"` for *any* path that is not byte-equal to the harness cwd, including a sibling project dir or a parent-dir `.claude`, neither of which is the user tier. | Either give them a production caller (see Required Changes #1, option C) or move both into the test module. If they stay public, fix the doc to say the function is a two-way equality check, not a tier classifier. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/meta/launch.rs` | 183 | A **third** CLI path carries the identical defect and is not fixed: `let fw = FrameworkPaths::default();` → `prepare_session(&fw, project_dir)` (:184), then `create_with_id(.., Some(project_dir), ..)` → `ClaudeCodeAdapter::spawn`, whose command **always** appends `SETTING_SOURCES_FLAG` (`runtime/claude_code.rs:285`). Same deploy-tier/load-tier mismatch. The PR and #4203 both assert "no tm-deployed agent was ever loaded on **either** CLI path", which reads as exhaustive. | Point it at `FrameworkPaths::for_managed_workspace(project_dir)` in this PR, or file a follow-up and say so in the PR body. (Checked and cleared: `commands/session/start.rs:85` is **not** affected — `start_session_in_place` spawns `claude {PERMISSION_MODE_FLAG}` with no `--setting-sources`, so the user tier *is* read there and `default()` is correct.) |
| LOW | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 727 | `deploy_output_style(&fw.claude_home_dir())` now resolves to the same directory as the project-tier call at :747, so launch/connect write the bundled styles twice to one place, and `$HOME/.claude/output-styles/` stops being refreshed by these two commands. Benign (still refreshed by `tm install`), but the "home tier" comment at :722–726 is now false for these callers. | Update the comment, or skip the second deploy when `fw.claude_home_dir() == project_dir`. |
| LOW | `crates/trusty-mpm/CHANGELOG.md` | 120–135 | The entry is accurate but omits the one genuinely user-visible side effect: `tm connect` now writes the agent roster and skills into the operator's **live checkout** `.claude/`, not `$HOME`. That is the change an operator will actually notice in `git status`. | Add one sentence noting the destination change for `connect`. |

## Mutation evidence

I did not read the test and take its word; I mutated production and observed.

**Mutation A — seam body reverted to `default()` (the author's own claim). Reproduced:**

```
test tests_behavior_b::launch_and_connect_deploy_into_a_tier_setting_sources_loads ... FAILED
agents deploy into the `user` tier (/Users/masa/.claude/agents) but the spawned session
carries `--setting-sources project,local`, loading only ["project", "local"] …
```

**Mutation B — both production call sites reverted to `FrameworkPaths::default()`, `session_framework_paths` left intact. This is #4203, fully reintroduced:**

```rust
// launch.rs:210  let fw = trusty_mpm::core::paths::FrameworkPaths::default();
// launch.rs:474  let fw = trusty_mpm::core::paths::FrameworkPaths::default();
```
```
$ cargo test -p trusty-mpm --bin tm
test result: ok. 1257 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
```

The bug is back, and **nothing in the suite notices** — including the test named for it. The test's reachable graph is `session_framework_paths` → `for_managed_workspace` → `claude_home_dir`, plus a `SETTING_SOURCES_FLAG` string parse. `launch()` and `connect()` are not in it. `session_framework_paths` has exactly one caller-of-record from the test's perspective: the test.

This is precisely the failure shape the PR body sets out to prevent ("it keeps biting if either side moves"). It bites if the *seam* moves; it is silent if the *call site* moves — and the call site is where the defect lived.

## Required Changes

1. **Make the guard bind the call sites.** In order of preference:

   **(C) — recommended.** Move the invariant into `session_launch::prepare_session_inner`, which is the function that actually performs the deploy and which every affected caller funnels through. It already receives both `fw` and `project_dir`; today it never checks that they agree. Reject (or loudly `tracing::error!` + record a `roster_error` for) the case `fw.claude_home_dir() != project_dir` when the spawn will carry `SETTING_SOURCES_FLAG`. One core-level test then guards `launch`, `connect`, the daemon path, **and** `meta/launch.rs:183` simultaneously — and would have caught all three call sites without anyone having to remember to write a per-call-site test. This also gives `settings_tier_of` a real production caller, retiring MEDIUM #2.

   **(B) — cheap backstop, additive to (C).** A source-text guard in the tm bin, which is idiomatic here (`scripts/check_line_cap.sh`, "Root CLAUDE.md must stay untracked", "No leaked AI-generation artifacts" are all grep-shaped CI guards):
   ```rust
   const LAUNCH_SRC: &str = include_str!("commands/launch.rs");
   assert!(!LAUNCH_SRC.contains("FrameworkPaths::default()"),
       "launch/connect must deploy through session_framework_paths (issue #4203)");
   assert_eq!(LAUNCH_SRC.matches("session_framework_paths(&").count(), 2);
   ```

   Option (A) — keeping the current test — is not sufficient on its own; Mutation B is the whole argument.

2. **Resolve `meta/launch.rs:183`** — fix it here or file the follow-up and reference it in the PR body, so "either CLI path" stops reading as exhaustive.

3. **Retire or relocate the two test-only `pub` helpers** in `model_inject.rs` per MEDIUM #2 (subsumed if you take option C).

## Notes — verified and deliberately NOT flagged

Everything below I checked against the branch and cleared. Recording it so the next round does not re-litigate it.

- **Tier mapping is real, not nominal.** `for_managed_workspace(x)` = `for_managed_project(FrameworkPaths::default().root, x)` (`paths.rs:243–245`), and `for_managed_project` mutates **only** `claude_agents` and `claude_skills` (`paths.rs:204–213`). `fw.root`, `framework`, `agents`, `skills`, `user_skills`, `hooks`, `instructions`, `registry`, `trusty_mpm_root` — and therefore `config_toml()`, `catalog_root_for(&fw.root)`, `ManifestSources::resolve`, `agent_source_dir()`, `skill_source_dir()`, `framework_instructions_path()` — all still resolve from `~/.trusty-mpm`. `claude_home_dir()` (`paths.rs:394`) recovers the base by taking `claude_agents.parent().parent()`, so it returns the workspace. The PR's SOURCE/DESTINATION claim holds exactly.
- **Harness cwd is correct and always defined at both sites.** `launch`: tmux created at `managed_workdir` == `managed_path` (step 12), the same value passed to the seam at :210. `connect`: tmux created at `workdir` == `path` (step 4), same value passed at :474. `resolve_dir` falls back to `std::env::current_dir()` when `--dir` is absent (`commands/project.rs:25–30`), so `connect` always has a workspace — there is no "no project context" hole.
- **Daemon path is untouched.** `lifecycle.rs:963/1087` builds its own `FrameworkPaths` and calls `prepare_session_with_repo_url` directly. No shared seam changed; `spawn_managed_inproject` behaves identically before and after.
- **Pre-existing `~/.claude/agents` is not shadowed, duplicated, or silently taking precedence.** The `user` tier is excluded by `SETTING_SOURCES_FLAG`, so the stale copy is inert for every tm-spawned session — it cannot outrank the new project-tier deploy. It also does not go stale in practice: `tm install` (`install.rs:59`) and `tm catalog apply` (`managed.rs:812/841`) still refresh it via `default()`. An upgrading user's populated `~/.claude/agents` needs no migration. Reasoned about explicitly because the diff does not.
- **`connect` writing into the live checkout does not clobber operator content.** `deploy_agents_filtered` (`trusty-agents-common/src/agents/deployer.rs:110–120`) skips any file absent from its manifest as user-owned, and warns+skips a managed file whose checksum drifted; `deploy_all_skill_tiers` treats hand-placed project skills as the top tier and never overwrites them; `scaffold_gitignore::ensure_scaffold_gitignored(project_dir)` (`session_launch/mod.rs:757`, #3427) gitignores the scaffolding paths. A repo that already *committed* `.claude/agents/**` is the one exposed case, and it has a documented remediation (`scaffold_tracking` doctor check).
- **The CI-exclusion disclosure is accurate.** `ci.yml:164` (clippy) and `ci.yml:564` (test) both use exactly `--exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui` — identical to what the author ran, not a convenient widening. The three excluded crates contain no reference to `model_inject`, `session_framework_paths`, or the `tm` binary, and none depends on `trusty-mpm`; they cannot be affected by this change. Disclosure verified as both honest *and* correct.
- **CHANGELOG placement is right** — under `## [Unreleased]` → `### Fixed` (line 118), not backdated into `## [1.0.2]`.
- **Scope.** +166 is not creep in the ordinary sense — no unrelated behavior rides along. But ~68 of those lines (`model_inject.rs`) are production API that exists solely to express a test which, per Mutation B, does not test the fix. That surface is currently unpaid-for; option (C) is what would earn it.
- **CI at review time:** all checks green except `Test`, still running.

bobmatnyc:
## Round 2 — all findings addressed

Thanks for Mutation B. It was correct and it was the whole argument: my guard bound `session_framework_paths`, a helper whose only caller-of-record was the test itself, so it proved the seam behaved and stayed silent about whether the commands used it. I took the recommended structural fix rather than patching the test.

### HIGH — the guard now binds the call sites

Rather than checking the `(fw, project_dir)` pair after the fact, I removed the degree of freedom that produced three independent instances of one bug: **`session_launch::prepare_isolated_session` resolves the deploy layout itself**, so an isolated caller supplies no `FrameworkPaths` at all. There is no wrong value left to pass.

On your option (C) as literally specified — a check inside `prepare_session_inner` — I hit a blocker worth recording: `prepare_session_inner` has heterogeneous callers, and `session/start.rs` (which you correctly cleared) funnels through it too. A mismatch check there would false-positive on exactly the call site you cleared unless the `--setting-sources` posture were threaded through every `prepare_session*` entry point. Constructing the layout inside a posture-specific entry point gets the same "one seam guards all callers" property without that churn, and leaves the non-isolated callers untouched.

The guard itself is now a **source-text assertion over the call sites** (your option B), because that is the only thing that actually fails on Mutation B — `launch()`, `connect()`, and `launch_and_wait()` are async, create tmux sessions, register over HTTP, and block on `attach-session`, so none is reachable from a unit test. Two core tests bind the shared layout; the source guard binds the three call sites to it.

**Your Mutation B, re-run — all three call sites reverted to `FrameworkPaths::default()`:**

```
thread 'tests_behavior_b::launch_paths_prepare_through_the_isolated_seam' panicked at
crates/trusty-mpm/src/bin/tm/tests_behavior_b_tests.rs:1713:9:
commands/launch.rs spawns `claude` with `--setting-sources project,local`, which excludes
the `user` tier `FrameworkPaths::default()` deploys into — deploy via
`session_launch::prepare_isolated_session` instead (issue #4203)

test result: FAILED. 1256 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out
```

**And each call site fails independently** — reverting only `meta/launch.rs`, leaving the other two fixed:

```
commands/meta/launch.rs spawns `claude` with `--setting-sources project,local`, which excludes
the `user` tier `FrameworkPaths::default()` deploys into — deploy via
`session_launch::prepare_isolated_session` instead (issue #4203)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1257 filtered out
```

The guard also **pins the premise you cleared**: it asserts `session/start.rs` still contains no `SETTING_SOURCES_FLAG`. If that file ever gains the flag it acquires this defect, and the guard is what says so.

### MEDIUM — third call site fixed

`commands/meta/launch.rs:183` now goes through `prepare_isolated_session`. `session/start.rs` deliberately keeps `FrameworkPaths::default()`, and the reasoning is recorded in `prepare_isolated_session`'s doc so the next reader does not "fix" it.

### MEDIUM — test-only `pub` API retired

**Deleted both** (`setting_source_tiers`, `settings_tier_of`) and reverted `model_inject.rs` entirely — zero production API added by this PR now. You were right that `settings_tier_of` was a byte-equality check mislabelled as a tier classifier; rather than fix the label I removed the function, and the tier relationship is asserted directly in the test that needs it, parsed out of `SETTING_SOURCES_FLAG` so it still cannot drift. That is ~68 lines of unpaid-for surface gone rather than earned.

### LOWs — both taken

- The "home tier" comment on the output-style deploy is corrected: it now states that for `isolated_framework_paths` callers and the daemon's `for_managed_workspace` spawn, `fw.claude_home_dir()` **is** `project_dir`, so the two deploys coincide and `$HOME/.claude/output-styles/` is refreshed by `tm install` / `tm catalog apply` instead. Comment-only — I did not add a skip branch, since the false comment was the actual defect.
- CHANGELOG now calls out the operator-visible change: `tm connect` writes into the **live checkout** `.claude/`, will appear in `git status`, is auto-gitignored (#3427), and never overwrites hand-placed or drifted content.

### Gate

```
cargo test --workspace --exclude <3 GUI crates> -- --test-threads=2
  → exit 0 — passed=19708 failed=0 ignored=132

cargo test -p trusty-mpm -- --test-threads=2   → exit 0 — 6515 passed; 0 failed; 14 ignored
cargo clippy --workspace --all-targets --exclude <same 3> -- -D warnings  → exit 0
cargo fmt --check                                                         → exit 0
scripts/check_line_cap.sh   → line-cap: 7 allowlisted, 0 violations — OK.
```

All three new tests confirmed `... ok` in the final workspace run, not skipped:

```
test core::session_launch::tests::isolated_layout_deploys_into_a_tier_the_spawn_reads ... ok
test core::session_launch::tests::isolated_layout_keeps_framework_source_at_the_install_root ... ok
test tests_behavior_b::launch_paths_prepare_through_the_isolated_seam ... ok
```

**Disclosure on the road to that green:** intermediate runs of this gate hit six failures across `trusty-agents` and `trusty-search`, plus one OOM kill (exit 137). All were resource artifacts — this machine was at **load average 108** from concurrent cargo builds in sibling worktrees. Every failure was 5-second-timeout shaped (`api server did not become healthy within 5s`, `gh did not respond within 5s`, uv lock contention), a *different* test each run, and each passed in isolation. `trusty-agents` has **zero** dependency edge on `trusty-mpm` (`cargo tree -p trusty-agents -e normal | grep -c "trusty-mpm v"` → `0`), and this diff is confined to `crates/trusty-mpm`, so it cannot reach those binaries. The green above is from a run at load ~68 with `--test-threads=2`. Flagging it rather than quietly reporting only the good run.

Rebased onto current `origin/main` (`5bf231b9`). Not merged. Handing back for re-review.

Issue #1511, opened 2026-06-20
P0: SessionManager decommission rm -rf's external/real workspace_path — deleted a live repo during prune


bobmatnyc:
## What happened
Running `tm sessions prune --state all` (the #1508 bulk prune) decommissioned a managed session record whose `workspace_path` pointed at a REAL user repository (`/Users/masa/Projects/trusty-tools`, symlinked to `/Volumes/SSD1/Projects/trusty-tools`). `SessionManager::decommission` calls `remove_dir_all(workspace_path)`, which deleted the entire working checkout **including its `.git`**. Recovered only because all work was pushed to origin and re-cloned.

## Root cause
`decommission` (crates/trusty-mpm/src/session_manager/manager.rs ~642-673) unconditionally `remove_dir_all`s `workspace_path` when set. But `workspace_path` is NOT always an SM-provisioned isolated clone:
- The #1502 **local-path managed spawn** (`workdir = existing abs dir`, skips clone) sets `workspace_path` to the user's REAL directory.
- The #1502 **adopt_existing** path points at real, pre-existing tmux cwds.
So decommission/prune can `rm -rf` a directory the SM never created. The pre-purge backup confirmed exactly one such record (`tmpm-trusty-tools`, state=stopped, workspace_path=/Users/masa/Projects/trusty-tools) among 239.

## Required fix (defense in depth)
1. **Never delete a workspace the SM did not provision.** Track ownership explicitly: add `#[serde(default)] workspace_owned: bool` to `SessionRecord` (true only when the SM created the dir via clone-provisioning; false for local-path spawn #1502 and adopt). `decommission` deletes the dir ONLY when `workspace_owned == true`.
2. **Path-containment guard (belt & suspenders).** Even when owned, only `remove_dir_all` if the canonicalized `workspace_path` is strictly INSIDE the canonicalized managed workspaces root (e.g. ~/.trusty-mpm/workspaces or the configured root). Refuse (log + skip) any path outside it. This alone would have prevented the incident.
3. **Extra safety:** refuse to remove any path that is a git root / contains `.git` while not owned, and never remove `$HOME`, a volume root, or a path with too few components. Log loudly when a deletion is skipped for safety.
4. For local-path/adopt sessions, prefer setting `workspace_path = None` (or owned=false) so there is nothing for decommission to delete; the user's dir is theirs.
[...]

Issue #4207, opened 2026-07-28
epic(trusty-mpm): git-native workstreams and ephemeral writer worktrees


bobmatnyc:
## Decision

A durable branch represents a workstream; every writing agent receives a short-lived worktree; read-only agents use the source checkout.

## Scope

- enumerate and validate through git-native primitives;
- replace path-shape and sentinel ownership assumptions;
- reconcile dual registries through dry-run-first migration;
- reclaim only merged, clean, inactive worktrees;
- preserve gitignored/untracked safety;
- update DOC-52 ownership language;
- carry ownership facts through guided reset and cleanup APIs;
- adversarial tests for live sessions, ambiguous roots, and registry disagreement.

## Done when

Concurrent writers are isolated, branches survive worktree deletion, no unsafe worktree is reclaimed, legacy registries reconcile observably, and all control paths use the same git-derived predicate.

DOC amendment #4208 and implementation review follow-up #4214 are consolidated here.

bobmatnyc:
Scope for tm 1.3.0: slices 1 and 2 are in scope for feature-complete. Slice 3 (dual-registry reconcile) is explicitly OUT of scope for feature-complete — deferred, no separate issue filed for it.

bobmatnyc:
SPRINT phase: drive to feature-complete on a local version; targeted tests, no CI iteration loops, no critic round on narrow changes.
HARDEN phase: after feature-complete — full suite, critic, release gates. Publish only after that.

bobmatnyc:
Follow-up: #4214 (2 MEDIUM findings in guided_inplace.rs / agent_reset_workspace.rs from PR #4209 review) — this epic's git-native rebuild inherits that module's ownership contract.

bobmatnyc:
**Scope addition (owner-directed): MCP service identity must not be worktree-scoped.**

Observed: this project's `.mcp.json` binds trusty-search to `--index 2eb72dca-de08-481b-8dfa-22ab7f81b1f9` — a session/worktree UUID. Service identity is therefore bound to ephemeral writer isolation.

Why it belongs here: under the 1.3.0 model (one worktree per writing agent), N concurrent writers produce N separate indexes of identical repo content — each independently stale, each paying its own indexing cost. Indexes are also not shareable across a workstream. Same failure shape as the per-worktree `target/` duplication.

[...]

Issue #2178, opened 2026-07-07
Data-risk: reindex colocated-root-moved heuristic hijacks unrelated worktree and prunes real corpus


bobmatnyc:
## Problem

The "colocated root moved" heuristic in `crates/trusty-search/src/service/reindex/runner.rs:160` (referenced in issue #402) can be hijacked by an unrelated active git worktree, causing a reindex job to silently re-walk the wrong path and prune the actual corpus.

### Incident details

During a live reindex of `cto` (configured root: `/Users/masa/Duetto/cto`, ~369K chunks), while an unrelated trusty-tools git worktree session was active at `.../trusty-tools/.worktrees/d72fb4a3-...`, the heuristic wrongly concluded cto's root had "moved" to the active worktree path. It then:
1. Re-walked the unrelated path (2,506 files) instead of cto's real corpus
2. Pruned 369,568 "stale" chunks marked as deleted
3. Left the index with `stale_index_root: true` until restart

The on-disk `indexes.toml` root config was always correct — only the reindex job's transient in-memory state was hijacked.

### Severity & Impact

**DATA-RISK.** Search on the affected index goes empty until daemon restart. Recovery occurred only by accident: the atomic corpus swap failed ("previous corpus preserved") and graceful restart reloaded the intact on-disk redb. This recovery was luck, not design.

Historical pattern: this heuristic has misfired at least **5 times** across different indexes:
- genealogy
- eoy-2025
- matsuoka-com
- joanies-kitchen
- cto (twice)

Any index whose configured root can be shadowed by an active worktree checkout is exposed.

### Log evidence

```
reindex[cto]: colocated index root moved from Some("/Users/masa/Duetto/cto") to /Users/.../.worktrees/d72fb4a3-... — keys are root-relative (#402); preserving hash cache (no re-embed)
reindex[cto]: prune pass: pruned 369568 stale chunk(s) from 33061 deleted file(s)
WARN force reindex: atomic corpus swap failed for 'cto' ... previous corpus preserved
```

## Root Cause

The heuristic does not validate that a candidate "moved-to" root matches the index's persisted configuration in `indexes.toml` before accepting the move and performing a re-walk. It silently redirects based on path-matching heuristics alone.

## Options to Fix

- **(a)** Require the moved-to root to be validated against the index's configured `indexes.toml` root before accepting the move.
- **(b)** Refuse/skip the root-move (with a warning) when the candidate root is inside a git worktree or differs from the persisted config root.
[...]

PR #4209, opened 2026-07-28
fix(trusty-mpm): stop two paths writing into a destroyed worktree (#4204)

### LINKED ISSUE #4204: Two paths write into a destroyed worktree — is_dir() liveness check never inspects .git
DEFECT: two independent code paths gate on `workspace_path.is_dir()` alone. A worktree whose `.git` and source tree have been stripped still passes that check, because the directory node survives. Both then act on it as if live.

- `crates/trusty-mpm/src/core/agent_reset_workspace.rs`, `reset_active_workspace_agents` (~lines 125-190): iterates every non-decommissioned session record and, for each whose `workspace_path.is_dir()` and whose `workspace_owned` flag is set (or whose `is_session_worktree()` heuristic matches — that heuristic is pure path-shape, "parent dir is named `.worktrees`", equally blind to a stripped `.git`), calls `reset_project_agents`, writing into `<workspace>/.claude/agents/*`. Trigger is `tm install --reset-agents --reset-agents-workspaces` (`install.rs:117`) — an operator-invoked bulk command, not a timer.
- `crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs:526`, in `run_inplace_relaunch` (function at 494-574): the same `!cwd.is_dir()` check passes, so it execs `claude` into the dead directory (`guided_inplace.rs:558-573`) instead of falling through to the picker.

EVIDENCE (observed 2026-07-27 22:07 local): worktree `.base/.worktrees/f443c12d-2fb6-4ce1-9f70-2e7695306e47` has no `.git`, no source tree, and is absent from `git worktree list` — yet `.claude/` mtime was 21:28:44, ~39 minutes old, and 45 files under `.claude/agents/` plus the manifest were written at 20:23. A `claude` process (PID 63302, started 18:23:27) is live with that stale cwd, along with 4 MCP children sharing it.

OPEN QUESTION — do not present as solved: the 21:28 batch (`settings.json`, `.trusty-mpm-skills-manifest.json`, `output-styles/*`) is NOT attributed. No workspace-wide sweep for skills/settings exists in the codebase — only `agent_reset_workspace.rs` has one, and it covers agents only. That writer is unidentified; it may be a second narrower `tm install` invocation or something not yet located. State this explicitly as unknown.

RULED OUT: `daemon/managed_routes/reactivate.rs` is NOT involved — `mark_reactivated` only flips session state and never calls `ensure_deployment_complete`/`deploy_validate`. An earlier inference that `/reactivate` triggered the deploy-repair pipeline was checked against the source and retracted.

RESOLUTION: add a `.git`-liveness guard alongside the existing `is_dir()` check at both sites, so a gutted worktree is treated as dead. Relates to #3764 / PR #4202 (which detects this state but does not prevent the writes) and #3715.

bobmatnyc:
## What

Two independent code paths gated on `workspace_path.is_dir()` alone and then
acted on the directory as if it were live. A worktree whose `.git` and source
tree have been stripped still passes that check, because the directory node
survives.

- `core::agent_reset_workspace::reset_active_workspace_agents` — recomposed
  45 agent files into a gutted worktree's `.claude/agents/`, reached via
  `tm install --reset-agents --reset-agents-workspaces`.
- `bin/tm/commands/guided_inplace::run_inplace_relaunch` — `exec`'d `claude`
  into the dead directory instead of falling through to the picker.

Both now share one predicate, `core::workspace_liveness::workspace_liveness`.

## Why this is not `&& path.join(".git").exists()`

The obvious patch is wrong in both directions, and the new module documents
each trap with a test that catches it:

1. **In a LINKED worktree, `.git` is a FILE, not a directory.** `git worktree
   add` writes a `gitdir:` pointer file, and `WorkspaceProvisioner` creates
   *every* managed workspace that way. A guard spelled
   `.join(".git").is_dir()` would reject every legitimate managed workspace —
   the exact inverse of the bug. The predicate stats the `.git` name and
   accepts a file, a directory, or a symlink.

2. **Not every workspace is owed a `.git`.** A missing marker is evidence of
   death only where one is owed — `workspace_owned || is_session_worktree(path)`,
   deliberately the SAME predicate the #1511 ownership gate already uses rather
   than a third notion. `workspace_owned` means `provision_in` created the
   directory (mandatory `repo_url`, always ends in `worktree_add`), and the
   `.worktrees/<id>` shape is only ever produced by `git worktree add`. Outside
   that, the verdict is `Live`, so an adopted non-git workspace is never
   refused. In `guided_inplace` the resolved path can be `record.cwd` or even
   `current_dir()`, so only the narrow path-shape rule applies there.

3. **A failed observation is not a death.** Permission denied, a stalled
   network mount, or any other I/O error yields `Indeterminate`, and
   `WorkspaceLiveness::is_dead()` returns `false` for it — the invariant is
   encoded in the type so no call site can quietly widen "dead". Silently
   declining to service a live workspace is a far quieter failure than the bug
   being fixed; this is the over-refusal trap PR #4202's round-1 fix fell into.

`path.is_dir()` is kept verbatim as step 1, so the pre-existing "gone / not a
directory" skip has identical semantics and this change adds no new abort paths.

## `is_session_worktree()` — decision: leave it alone

It stays the pure path-shape heuristic it is. It answers *"may trusty-mpm
DELETE this?"* (ownership) on paths that may legitimately be absent —
`is_session_worktree_absent_path_is_noop` pins that — and it backs
`decommission` and `search_gc`. Teaching it about `.git` would make those
paths refuse to clean up the very husks this issue is about, and would change
two unrelated call sites. Ownership and liveness are orthogonal axes;
`workspace_liveness` composes them rather than conflating them.

## Existing helper?

No reusable one exists on `main`. The closest are
`session_launch::worktree_sync::sync_worktree_with_upstream` (`.exists()`, and
its concern is "can I fast-forward this?") and `guided::cwd_owns_git_entry`
(`try_exists().unwrap_or(false)`); both fail CLOSED on an I/O error, which is
exactly the behavior this fix must avoid, and neither models the
owed-a-`.git` expectation rule.
`worktree_integrity.rs` — where PR #4202 is adding classification logic in
this same area — does **not** exist on `main`, so this is implemented locally
rather than depending on unmerged code. **Future consolidation:** once #4202
lands, `worktree_integrity`'s classifier and `core::workspace_liveness` should
be reconciled into a single module; they are the detection and prevention
halves of the same notion.

## Relationship to #4202 and the unattributed writer

This is the **PREVENTION** counterpart to PR #4202's **DETECTION**: #4202
recognises a gutted worktree, this stops the two known writers from acting on
one. They are complementary, not redundant.

**NOT addressed here:** the unattributed 21:28 writer from #4204 — the batch
that touched `settings.json`, `.trusty-mpm-skills-manifest.json`, and
`output-styles/*`. No workspace-wide sweep for skills/settings exists in the
codebase; that writer remains unidentified and this PR makes no claim about it.

## Tests

Each of the first three fails against `main` and passes here; the anti-over-
refusal tests additionally fail against a naive `.join(".git").is_dir()` guard
(verified by temporarily substituting one). No `#[ignore]`, no cfg-gates.

| Test | Proves |
| --- | --- |
| `sweep_skips_gutted_worktree_whose_git_was_stripped` | a gutted worktree is not written into, and is reported with a skip reason |
| `sweep_serves_live_linked_worktree_with_git_file` | a legitimate linked worktree (`.git` as a FILE) IS still served |
| `run_inplace_relaunch_falls_through_on_gutted_worktree` | the relaunch falls through, with the daemon mock at ZERO hits — no exec, no mutation |
| `run_inplace_relaunch_serves_live_linked_worktree` | a live linked worktree still gets past the gate |
| `core::workspace_liveness::tests::*` (6) | the predicate itself, including `liveness_indeterminate_is_never_dead` |

Two pre-existing sweep fixtures used a bare `TempDir` with no `.git`, which
did not model any workspace this code actually serves; they now write the
`gitdir:` pointer file a real worktree carries.

Closes #4204

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -40,6 +40,25 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   — unmanaged files are skipped as user-owned, and managed files whose checksum
   drifted are skipped with a warning.
 
+- **Two paths no longer write into a destroyed worktree**
+  ([#4204](https://github.com/bobmatnyc/trusty-tools/issues/4204)):
+  `tm install --reset-agents --reset-agents-workspaces` and the bare-`tm`
+  in-place relaunch both gated on `workspace_path.is_dir()` alone, which a
+  gutted worktree — `.git` and source tree stripped, directory node surviving —
+  passes. The sweep recomposed 45 agent files into such a husk and the relaunch
+  exec'd `claude` into it. Both now share one predicate,
+  `core::workspace_liveness::workspace_liveness`, which skips ONLY on positive
+  evidence of death: it accepts `.git` as a FILE (the `gitdir:` pointer every
+  linked worktree carries, so no legitimate managed workspace is refused),
+  requires a `.git` only where one is owed (a tm-provisioned workspace or a
+  `.worktrees/<id>` path), and — at EVERY stat it performs, including the
+  initial existence check — reports a filesystem it could not interrogate as
+  `Indeterminate`, explicitly not dead, so a permission failure, a symlink
+  loop, or a stalled network mount can never silently stop servicing a live
+  workspace. A gutted workspace, and a session record whose workspace has
+  vanished, are both now reported to the operator with a skip reason rather
+  than dropped silently.
+
 ---
 ## [1.2.0] — 2026-07-27
 
@@ -167,23 +186,6 @@ so a 2.0.0 would signal a break that cannot reach anyone.
   ordering-dependent behaviour is reproducible.
 
 ### Fixed
-
-- **Two paths no longer write into a destroyed worktree**
-  ([#4204](https://github.com/bobmatnyc/trusty-tools/issues/4204)):
-  `tm install --reset-agents --reset-agents-workspaces` and the bare-`tm`
-  in-place relaunch both gated on `workspace_path.is_dir()` alone, which a
-  gutted worktree — `.git` and source tree stripped, directory node surviving —
-  passes. The sweep recomposed 45 agent files into such a husk and the relaunch
-  exec'd `claude` into it. Both now share one predicate,
-  `core::workspace_liveness::workspace_liveness`, which skips ONLY on positive
-  evidence of death: it accepts `.git` as a FILE (the `gitdir:` pointer every
-  linked worktree carries, so no legitimate managed workspace is refused),
-  requires a `.git` only where one is owed (a tm-provisioned workspace or a
-  `.worktrees/<id>` path), and reports an unreadable filesystem as
-  `Indeterminate` — explicitly not dead — so a permission or network-mount
-  failure can never silently stop servicing a live workspace. A gutted
-  workspace is now reported to the operator with a skip reason rather than
-  dropped silently.
 
 - **Two wrong instructions corrected in the bundled assets composed into every
   session's prompt.** Both were static text that no project could override, and
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -168,6 +168,23 @@ so a 2.0.0 would signal a break that cannot reach anyone.
 
 ### Fixed
 
+- **Two paths no longer write into a destroyed worktree**
+  ([#4204](https://github.com/bobmatnyc/trusty-tools/issues/4204)):
+  `tm install --reset-agents --reset-agents-workspaces` and the bare-`tm`
+  in-place relaunch both gated on `workspace_path.is_dir()` alone, which a
+  gutted worktree — `.git` and source tree stripped, directory node surviving —
+  passes. The sweep recomposed 45 agent files into such a husk and the relaunch
+  exec'd `claude` into it. Both now share one predicate,
+  `core::workspace_liveness::workspace_liveness`, which skips ONLY on positive
+  evidence of death: it accepts `.git` as a FILE (the `gitdir:` pointer every
+  linked worktree carries, so no legitimate managed workspace is refused),
+  requires a `.git` only where one is owed (a tm-provisioned workspace or a
+  `.worktrees/<id>` path), and reports an unreadable filesystem as
+  `Indeterminate` — explicitly not dead — so a permission or network-mount
+  failure can never silently stop servicing a live workspace. A gutted
+  workspace is now reported to the operator with a skip reason rather than
+  dropped silently.
+
 - **Two wrong instructions corrected in the bundled assets composed into every
   session's prompt.** Both were static text that no project could override, and
   both had demonstrably misrouted live sessions.
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
@@ -86,6 +86,7 @@
 //! pre-existing refuse+switch-client behavior instead.
 
 use anyhow::Context as _;
+use trusty_mpm::core::workspace_liveness::workspace_liveness;
 
 use super::guided_resume::ResumeAction;
 
@@ -467,14 +468,17 @@ pub(crate) enum InPlaceOutcome {
 /// (cwd/binary resolution) or post-reactivate (`exec` itself failing) alike
 /// — as a fall-back-to-reconnect condition, matching that call site's
 /// pre-#2453 behavior of never hard-failing.
-/// What, IN ORDER (#2027 exec-ordering fix, extended by #2790): (1) prints a
-/// one-line notice; (2) resolves the workdir; (2a) #2790 CRITICAL:
-/// verifies the resolved workdir EXISTS ON DISK — a Decommissioned record's
-/// workspace may have been removed by `decommission()` while the record's
-/// `workspace_path`/`cwd` still names it; a missing directory returns
-/// [`InPlaceOutcome::FallThrough`] here, BEFORE any daemon call, so a
-/// known-gone workspace can never be masked by a durably-committed `Active`
-/// record; (3) builds the resume argv via
+/// What, IN ORDER (#2027 exec-ordering fix, extended by #2790 and #4204): (1)
+/// prints a one-line notice; (2) resolves the workdir; (2a) #2790/#4204
+/// CRITICAL: verifies the resolved workdir IS STILL LIVE via
+/// [`trusty_mpm::core::workspace_liveness::workspace_liveness`] — a
+/// Decommissioned record's workspace may have been removed by `decommission()`
+/// while the record's `workspace_path`/`cwd` still names it, and (#4204) a
+/// worktree may have been GUTTED, keeping its directory node while losing its
+/// `.git` and source tree, which the previous bare `is_dir()` test could not
+/// see; either verdict returns [`InPlaceOutcome::FallThrough`] here, BEFORE any
+/// daemon call, so a dead workspace can never be masked by a durably-committed
+/// `Active` record nor exec'd into; (3) builds the resume argv via
 /// [`trusty_mpm::runtime::build_inplace_resume_command`] (the SAME
 /// `--resume`-existence-check → `--continue`/fresh-spawn fallback logic the
 /// tmux-pane resume path uses, #2013) — this is PURE/local (resolving the
@@ -484,11 +488,15 @@ pub(crate) enum InPlaceOutcome {
 /// [`reactivate_managed_session`]), immediately before exec — on a
 /// refused/failed reactivate this returns [`InPlaceOutcome::FallThrough`]
 /// rather than proceeding; (5) execs `claude` in place.
-/// Test: I/O path; not unit-tested (requires a live daemon + real `claude`).
-/// The workdir-existence gate (2a) is exercised via `run_inplace_relaunch`'s
-/// I/O surface only (needs the daemon + tmux + `claude`); the FSM invariant it
-/// protects is separately proven at the daemon layer (defense-in-depth, so the
-/// guard holds even for a future caller that skips this CLI-side check) by
+/// Test: `run_inplace_relaunch_never_reactivates_when_command_build_fails`,
+/// `run_inplace_relaunch_falls_through_on_gutted_worktree`,
+/// `run_inplace_relaunch_serves_live_linked_worktree`. The exec itself is not
+/// unit-tested (it replaces the process image); the gate-2a tests pin the
+/// decision by asserting the daemon mock records ZERO hits, which is only true
+/// when the fall-through happened before any daemon mutation. The FSM invariant
+/// gate 2a protects is separately proven at the daemon layer
+/// (defense-in-depth, so the guard holds even for a future caller that skips
+/// this CLI-side check) by
 /// `mark_reactivated_refuses_when_workspace_removed_by_decommission` in
 /// `session_manager::reactivate_tests`.
 pub(crate) async fn run_inplace_relaunch(
@@ -523,10 +531,25 @@ pub(crate) async fn run_inplace_relaunch(
     // missing workspace into `FallThrough` here — zero daemon calls made —
     // keeps the invariant intact and routes the operator to the same honest
     // self-relaunch hint a refused reactivate would.
-    if !cwd.is_dir() {
+    // #4204 EXTENSION of the #2790 gate above: `is_dir()` is not a liveness
+    // check. A worktree whose `.git` and source tree have been stripped keeps
+    // its directory node, so the bare `is_dir()` test passed and this function
+    // exec'd `claude` into the husk — observed 2026-07-27 with PID 63302 (plus
+    // four MCP children) still parked on the dead cwd of
+    // `.base/.worktrees/f443c12d-…`.
+    //
+    // `ManagedSessionSummary` carries no `workspace_owned` flag, so `false` is
+    // passed for it and the helper falls back to its NARROW path-shape rule
+    // (`.worktrees/<id>`). That is deliberate, not a shortcut: such a directory
+    // is only ever produced by `git worktree add`, so a missing `.git` there is
+    // unambiguous evidence of destruction — while an operator's plain,
+    // non-git working directory (reachable here via `record.cwd` or
+    // `current_dir()`) is owed no `.git` and is never refused.
+    let workspace_owned_unknown = false;
+    let liveness = workspace_liveness(&cwd, workspace_owned_unknown);
+    if liveness.is_dead() {
         eprintln!(
-            "tm: workspace for session {id} no longer exists on disk ({}) — cannot relaunch in \
-             place",
+            "tm: workspace for session {id} is {liveness} ({}) — cannot relaunch in place",
             cwd.display()
         );
         return InPlaceOutcome::FallThrough;
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
@@ -496,6 +496,143 @@ async fn reactivate_omits_caller_pane_id_query_when_absent() {
     );
 }
 
+/// A `ManagedSessionSummary` whose workspace is `workspace`, in the `stopped`
+/// shape `run_inplace_relaunch` is normally handed.
+///
+/// Why: the struct has ~20 fields and three #4204 tests need it; inlining it
+/// three more times would be pure noise.
+/// What: `TEST_ID`/`tm-test`/`stopped` with `workspace_path` set and every
+/// other field at its empty default.
+/// Test: used by `run_inplace_relaunch_falls_through_on_gutted_worktree` and
+/// `run_inplace_relaunch_serves_live_linked_worktree`.
+fn stopped_record_at(workspace: &std::path::Path) -> trusty_mpm::client::ManagedSessionSummary {
+    trusty_mpm::client::ManagedSessionSummary {
+        id: TEST_ID.to_string(),
+        name: "tm-test".to_string(),
+        state: "stopped".to_string(),
+        persisted_state: None,
+        workspace_path: Some(workspace.to_string_lossy().to_string()),
+        repo_url: None,
+        branch: None,
+        created_at: None,
+        last_activity_at: None,
+        pending_decision: None,
+        proposed_default: None,
+        source_id: None,
+        task: None,
+        cwd: None,
+        claude_session_id: None,
+        deliverable_id: None,
+        pane_id: None,
+        injection_status: None,
+        unresumable: false,
+        stale_assets: false,
+        attached: false,
+        slot: 0,
+        deleted: false,
+    }
+}
+
+#[tokio::test]
+async fn run_inplace_relaunch_falls_through_on_gutted_worktree() {
+    // ISSUE #4204: a worktree whose `.git` and source tree were stripped keeps
+    // its directory node, so the old `cwd.is_dir()` gate passed and this
+    // function exec'd `claude` into the husk (observed 2026-07-27: PID 63302
+    // plus four MCP children parked on the dead cwd of
+    // `.base/.worktrees/f443c12d-…`). It must now fall through to the picker
+    // instead, BEFORE any daemon mutation.
+    //
+    // The mock answers 409 so that even on pre-fix code — where `claude` may
+    // resolve and the flow would continue — the reactivate is REFUSED and the
+    // real `exec` (which would replace this test process) is never reached.
+    // The decisive assertion is therefore `hits == 0`, not the outcome alone:
+    // against main this test fails either way — with `claude` absent the
+    // outcome is `Result(Err(..))`, and with `claude` present the daemon is
+    // contacted and `hits == 1`.
+    let base = tempfile::tempdir().expect("tempdir");
+    let gutted = base
+        .path()
+        .join(".worktrees")
+        .join("f443c12d-2fb6-4ce1-9f70-2e7695306e47");
+    std::fs::create_dir_all(gutted.join(".claude")).expect("create gutted worktree");
+    assert!(
+        gutted.is_dir() && !gutted.join(".git").exists(),
+        "fixture must be exactly what the old is_dir() gate waved through"
+    );
+
+    let (url, hits) = spawn_mock("HTTP/1.1 409 Conflict").await;
+    let client = reqwest::Client::new();
+
+    let outcome = run_inplace_relaunch(
+        &client,
+        &url,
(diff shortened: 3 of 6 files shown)


bobmatnyc (commented):
## Verdict: WARN

Adversarial review, independent of the implementer. Every claim below was verified against the code and, where marked, against a local test run at `1ab2a6ea` — not against the PR description. CI is green (22 pass / 1 skip).

This is a net safety improvement over `main` and should not be blocked. One HIGH finding: the module's own stated invariant — "an I/O failure is never death" — is **not upheld on its first branch**, and I proved it empirically.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/workspace_liveness.rs` | 151 | `if !path.is_dir()` collapses **every** io error into `Absent`, and `Absent.is_dead() == true`. A fully live linked worktree behind an EACCES parent (or a stalled mount / EIO / ENOTDIR) classifies as **dead**. This contradicts the module doc (:29-37), the `Absent` docstring (:69) and the CHANGELOG verbatim claim that "a permission or network-mount failure can never silently stop servicing a live workspace". | Replace with an explicit `fs::metadata` match — see below. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs` | 548 | `workspace_owned_unknown = false` leaves tm-provisioned **clone** workspaces (`session_manager/create.rs:294` sets `workspace_owned: owned`; path is not `.worktrees/<id>`-shaped) entirely outside the expectation rule — a gutted clone still classifies `Live` and `claude` is exec'd into the husk. | Add `workspace_owned` to `ManagedSessionSummary` + the daemon serializer and pass it through. |
| MEDIUM | `crates/trusty-mpm/src/core/agent_reset_workspace.rs` | 229-236 | `reset_project_agents(...)?` — one workspace's io failure aborts the **entire** sweep for all remaining sessions. The new `Indeterminate => proceed` branch (:191-198) deliberately routes unreadable workspaces into exactly this call, so the "far quieter failure" the module doc promises is in fact "loudly abort the sweep for everyone". Gets worse once HIGH-1 is fixed. | Collect the error into a `WorkspaceResetOutcome { skipped_reason: Some(..) }` and `continue`, rather than `?`. |
| LOW | `crates/trusty-mpm/src/core/workspace_liveness.rs` | 101, 105-110 | `Display` is phrased for `agent_reset_workspace.rs:187`'s `"skipped — {liveness}"` and reads wrong in `guided_inplace.rs:552`'s `"workspace for session {id} is {liveness}"` → **"…is no longer exists on disk"**. `Indeterminate` also hardcodes the caller *policy* `"— treated as live"` into the type. | Make the variants noun phrases (`"absent from disk"`), and move "treated as live" to the call site that decides it. |
| LOW | `crates/trusty-mpm/src/core/workspace_liveness.rs` | 123-128 | Doc says `workspace_owned` means `provision_in` created the directory. `provision_in` sessions are recorded with `workspace_owned = false` (`daemon/managed_routes/lifecycle.rs:992`); it is the clone path that sets `true`. The conclusion (owned ⇒ a checkout) still holds; the cited mechanism does not. | Cite the clone path. |

---

## The requested risks, answered

**1. Over-refusal on a linked worktree's `.git` FILE — REFUTED.** `workspace_liveness.rs:159` does use `std::fs::symlink_metadata(path.join(".git"))` and accepts `Ok(_)` for file, directory, or symlink alike. Verified by mutation, not by reading the description: replacing that block with `path.join(".git").is_dir()` fails **5** tests — `liveness_accepts_linked_worktree_git_file`, `sweep_serves_live_linked_worktree_with_git_file`, `sweep_resets_intact_workspace`, `sweep_respects_per_workspace_manifest_exclude`, `run_inplace_relaunch_serves_live_linked_worktree`. The claim is real and it is guarded.

**2. Uncertainty reaching a dead verdict — CONFIRMED at `workspace_liveness.rs:151`, refuted everywhere else.** Full audit of the new module: `:151` `is_dir()` (an implicit `unwrap_or(false)`) is the hole; `:155` `is_session_worktree` bottoms out in `unwrap_or(false)` at `decommission.rs:97`, which fails toward `Live` (safe); `:159-163` is an explicit three-arm match and is correct — only `NotFound` yields `Gutted`. No `?`, `.ok()`, or `unwrap_or_default` at either call site reaches a dead verdict.

Empirical proof of the hole (temporary probe, discarded):

```
CRITIC PROBE: EACCES on a LIVE worktree classified as Absent, is_dead=true
```

The fixture was a genuine `.worktrees/<id>` linked worktree with a `gitdir:` pointer file, asserted `Live` while readable, then re-observed with its parent at mode `0o000`.

**Scope correction to my brief, in the PR's favour:** neither consumer destroys anything. `is_dead()` gates a *refusal to write* (`agent_reset_workspace.rs:178,182`) and a *fall-through* (`guided_inplace.rs:550`). `git grep` confirms these are the only two consumers. So this defect is over-refusal today, not data loss — which is why it is HIGH and not CRITICAL. What earns it HIGH anyway: `Absent` at `agent_reset_workspace.rs:178` is `continue` with **no outcome pushed and no warning**, i.e. the io-error case is precisely the "silently stop servicing a live workspace" scenario the module says it exists to prevent; and `is_dead()` is published as an absolute reusable contract that #4207's git-native work will build destructive operations on. A future destructive consumer inherits "unreadable ⇒ delete" for free.

Fix that preserves current semantics and closes it:

```rust
match std::fs::metadata(path) {
    Ok(m) if m.is_dir() => {}
    Ok(_) => return WorkspaceLiveness::Absent,
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => return WorkspaceLiveness::Absent,
    Err(e) => return WorkspaceLiveness::Indeterminate(format!("cannot stat workspace: {e}")),
}
```

Land it together with the MEDIUM on `agent_reset_workspace.rs:236`, or an unreadable workspace will newly abort the whole sweep.

**3. `workspace_owned = false` at `guided_inplace` — sound for #4204, incomplete in general.** The observed #4204 case is an in-project worktree, and those are recorded `workspace_owned = false` (`lifecycle.rs:992`) — so the path-shape rule is in fact the *only* thing that would ever have covered it, and it does. `launch_on_main` (`:177`) and adopted sessions point at the operator's real checkout and are correctly left alone. The uncovered class is tm-provisioned clones (MEDIUM above). This is fail-open in the same direction `main` already was, so it weakens nothing — it just does not finish the job.

**4. Leaving `is_session_worktree()` as a path-shape heuristic — REFUTED as a defect; the author's argument is correct.** Ownership ("may trusty-mpm delete/write here?") and liveness ("is there anything here?") are orthogonal, and `workspace_liveness` *composes* them at `:155` rather than conflating them. Teaching `is_session_worktree` about `.git` would make `decommission::remove_session_worktree` and `search_gc` refuse to clean up exactly the husks #4204 is about — and `is_session_worktree_absent_path_is_noop` already pins that it must answer on paths that do not exist. No second notion of "is this a worktree" is introduced anywhere in the tree; there is one ownership predicate and one liveness predicate.

**5. Do the tests guard the fix? — VERIFIED INDEPENDENTLY, both directions.** I reverted **both** production call sites to bare `is_dir()` in a throwaway worktree, keeping the new module and its tests:

```
sweep_skips_gutted_worktree_whose_git_was_stripped ... FAILED

run_inplace_relaunch_falls_through_on_gutted_worktree ... FAILED
```

Both #4204 regression tests die on revert. They exercise the real production entry points (`reset_active_workspace_agents`, `run_inplace_relaunch`), not a helper. The anti-over-refusal tests are likewise non-vacuous — see risk 1. Clean run at `1ab2a6ea`: `13 passed; 0 failed` (lib) and `3 passed; 0 failed` (`--bin tm`), no skips.

One caveat worth recording: `run_inplace_relaunch_serves_live_linked_worktree` asserts `build_failed || daemon_contacted`. `claude` is on PATH on this machine, so it ran the stronger `daemon_contacted` arm. On a host without `claude` it degrades to the `build_failed` arm — still non-vacuous (a gate refusal returns `FallThrough`, which is neither), just weaker. Acceptable.

**6. Alarm latching — N/A.** No give-up, alarm, latch, or counter exists in this diff. The one emission is `tracing::warn!` at `agent_reset_workspace.rs:192`, a per-observation log line for `Indeterminate`. Nothing repeats the `:597` dark-detector shape.

---

## On the `trusty-search` test

I **agree it is unrelated to this PR.** `crates/trusty-search/Cargo.toml` declares no `trusty-mpm` dependency (dev- or otherwise), and this PR touches only `trusty-mpm`. CI's `Test` job is green here, so it did not reproduce.

That said, flag it separately rather than filing it under "known flake": `root_hijack_tests.rs:86-89` isolates itself with `unsafe { std::env::set_var("TRUSTY_DATA_DIR", …) }` guarded by `#[serial]`. That is process-global state protected only against *other* `#[serial]` tests in the same binary — any non-`#[serial]` test reading `TRUSTY_DATA_DIR` races it. `reindex_refuses_untrusted_root_move_and_preserves_corpus` is the #2178 P0 data-loss guard; a guard that is green by luck is indistinguishable from one that is green by correctness. Worth its own issue.

---

## Forward-compatibility with #4207 (git-native)

Nothing here obstructs the move. `workspace_liveness` is the single thing both call sites consult, so swapping its body for `git worktree list --porcelain` is a one-function change — this PR actually *creates* the seam that migration needs. Two mild frictions, neither blocking: the `Absent`/`Gutted` names encode a filesystem taxonomy with no clean git-native analogue ("absent from `git worktree list`" is a third state), and `is_dead()` freezes that taxonomy into a public API. Keeping `is_dead()` as the only decision surface — which it is today — is the right shape; please keep it that way.

---

## Required changes (tracked, not merge-blocking)

1. `workspace_liveness.rs:151` — route io errors other than `NotFound` to `Indeterminate` (patch above). This is the one that must not be dropped: the doc and CHANGELOG currently promise a property the code does not have.
2. `agent_reset_workspace.rs:236` — stop letting one workspace's io failure abort the whole sweep. Land with (1).
3. `guided_inplace.rs:548` — carry `workspace_owned` on `ManagedSessionSummary` so provisioned clones are covered too.
4. LOWs at author's discretion.

## Notes

- Reviewed at `1ab2a6ea`. The PR is `BEHIND` `main`; nothing in the delta looked relevant, but re-run before merge.
- Explicitly **not** flagged: `is_session_worktree`'s pure path-shape rule means an operator's *own* `.worktrees/<id>` git worktree passes the #1511 ownership gate at `agent_reset_workspace.rs:210`. That is unchanged pre-existing behaviour, not introduced here.
- Verification evidence is in this comment rather than a file; the throwaway worktree used for the revert and mutation runs was removed and `git worktree prune`d.

bobmatnyc:
## Verdict: APPROVE

Re-review at head `a21ee67e` (the previous verdict was WARN at `1ab2a6ea`).
Reviewed in an independent worktree off `refs/pull/4209/head`; the engineer's
worktree was not entered. Every claim below was verified against the code or
against a local run — none of it is taken from the PR description or the commit
messages.

**HIGH count: 0.** The single HIGH that gated the last round is closed, and I
proved it closed by mutation rather than by reading. CI at this head is fully
green: **22 pass, 1 skipping (`Notify on red main`), 0 pending, 0 failing.**

---

## The prior HIGH — CLOSED

`workspace_liveness.rs:181-188` is now an explicit four-arm match on
`fs::metadata`:

```rust
match std::fs::metadata(path) {
    Ok(md) if md.is_dir() => {}
    Ok(_) => return WorkspaceLiveness::Absent,
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => return WorkspaceLiveness::Absent,
    Err(e) => return WorkspaceLiveness::Indeterminate(format!("cannot stat workspace: {e}")),
}
```

Exactly two branches reach `Absent`: a successful stat finding a non-directory,
and `NotFound`. Every other `ErrorKind` reaches `Indeterminate`, whose
`is_dead()` (`:108-110`, `matches!(self, Self::Absent | Self::Gutted)`) is
`false`. The secondary half of that finding is fixed too:
`agent_reset_workspace.rs:185-198` no longer `continue`s silently on `Absent` —
it warns and pushes a `WorkspaceResetOutcome` carrying a skip reason.

### Mutation evidence — independently reproduced, all three

I re-ran the three mutations myself in my own worktree (restored afterwards;
`git status --porcelain` clean).

**M1 — restore `if !path.is_dir() { return Absent }`:**
```
liveness_symlink_loop_is_indeterminate_never_dead ... panicked at :336
  an ambiguous stat failure (FilesystemLoop) must never be treated as death; got Absent
liveness_unreadable_parent_is_indeterminate_never_dead ... panicked at :384
  a LIVE worktree that merely could not be stat'd must never be treated as dead — got Absent
test result: FAILED. 7 passed; 2 failed
```

**M2 — delete the `NotFound` arm (the over-correction check):**
```
liveness_reports_absent_directory ... panicked at :299
  left: Indeterminate("cannot stat workspace: No such file or directory (os error 2)")
 right: Absent
test result: FAILED. 8 passed; 1 failed
```

**M3 — neuter the non-directory arm to `Live`:**
```
liveness_reports_non_directory_as_absent ... left: Live  right: Absent
test result: FAILED. 8 passed; 1 failed
```

Sound in both directions: the guard cannot be reduced to "can never report
death" without M2 firing.

### Are the tests vacuous? — NO

`liveness_unreadable_parent_is_indeterminate_never_dead` (`:363-401`) *can*
self-skip: `:379` probes `fs::metadata(&wt).is_err()` and `:385-388` `eprintln!`s
"skipped: running privileged" and returns when EACCES cannot be staged (root, or
a mode-ignoring filesystem). That is precisely the vacuity risk worth catching,
and here it does not bite:

- Run with `--nocapture` on this machine, no skip line was emitted — and M1 kills
  the test — so the probe genuinely staged.
- More importantly, the io-error invariant is **also** pinned by
  `liveness_symlink_loop_is_indeterminate_never_dead` (`:321-350`), which uses a
  symlink loop: deterministic for **every UID including root**. That is the
  load-bearing test, and M1 kills it too. So even on a runner where the EACCES
  probe silently no-ops, the invariant stays guarded.
- Both tests assert their premise (`:294-297`, `:331-338`) before asserting the
  conclusion, so neither can pass for the wrong reason.

---

## The engineer's four claims, verified

| # | Claim | Result |
|---|---|---|
| 1 | Only `Ok(non-dir)` and `NotFound` yield `Absent`; every other kind is `Indeterminate` (`is_dead == false`) | **VERIFIED** — `workspace_liveness.rs:181-188`, `:108-110` |
| 2 | Both call sites route through the single predicate | **VERIFIED, with a caveat** — see below |
| 3 | `metadata` preserves `is_dir()`'s symlink-following; no behavior change smuggled in | **VERIFIED** — `:181` is `fs::metadata`, not `symlink_metadata`. `symlink_metadata` appears only at `:194`, on `path.join(".git")` — a different, additive check, and there *not* following is the correct choice (a `.git` symlink counts as evidence of life) |
| 4 | The `is_dir`/`exists`/`is_file` ban is stated where a future editor sees it | **VERIFIED** — module doc `:44-48` ("Convenience predicates on `Path` that fold `Result` into `bool` (`is_dir`, `exists`, `is_file`) must not be reintroduced into this decision"), reinforced by the inline comment at `:173-180` directly above the match. Both placements sit on the path an editor of this function actually reads |

**Caller enumeration, done independently**
(`grep -rn --include='*.rs' -e workspace_liveness -e WorkspaceLiveness crates/`):
exactly two production consumers — `agent_reset_workspace.rs:175` and
`guided_inplace.rs:549`. There is no third consumer of the predicate, and no
third caller bypassing it *within the two paths #4204 names*. The caveat is a
**sibling** with the same shape, outside those two paths — MEDIUM-3 below.

### Local test evidence at `a21ee67e`

```
core::workspace_liveness::tests (9 tests)       ... test result: ok. 9 passed; 0 failed; 0 ignored
core::agent_reset_workspace + liveness (17)     ... test result: ok. 17 passed; 0 failed; 0 ignored
commands::guided_inplace::run_inplace_* (3)     ... test result: ok. 3 passed; 0 failed; 0 ignored
```
No skips, no ignores.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/agent_reset_workspace.rs` | 249-256 | `reset_project_agents(...)?` still aborts the **entire** sweep on one workspace's io failure. This was Required Change #2 last round, tied to the HIGH — and the HIGH fix widened its reach exactly as the prior review predicted: pre-fix an unreadable workspace was `Absent` → `continue`; now it is `Indeterminate` → proceeds into this very call, which then fails and kills the reset for every remaining session. Loud and recoverable, so not HIGH — but the module doc's "far quieter failure" promise (`:172-174`) currently reads as "abort the sweep for everyone". | Collect the error into a `WorkspaceResetOutcome { skipped_reason: Some(..) }` and `continue`, instead of `?`. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs` | 548 | `workspace_owned_unknown = false` — `ManagedSessionSummary` (`crates/trusty-mpm/src/client/http_client/types.rs:369`) still carries no `workspace_owned` field. A tm-provisioned **clone** (`session_manager/create.rs:294` `workspace_owned: owned`, set `true` at `daemon/managed_routes/lifecycle.rs:1254`) is not `.worktrees/<id>`-shaped, so a gutted clone classifies `Live` and `claude` is exec'd into the husk. **Consciously deferred, not dropped**: `:541-547` now documents the choice and why it is sound for the observed #4204 case (in-project worktrees are recorded `workspace_owned: false` at `lifecycle.rs:992`, so the path-shape rule is the only thing that ever covered them — and it does). Fail-open in the same direction `main` already was. | Add `workspace_owned` to `ManagedSessionSummary` + the daemon serializer and thread it through. Follow-up issue. |
| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/deployment_check.rs` | 74 | **NEW — a third path of the same shape, outside this PR's scope.** `if workspace == Path::new("/unknown") \|\| !workspace.is_dir() { return Ok(()) }` gates `validate_and_repair` (`core/deploy_validate/mod.rs:351`), which on an incomplete workspace calls `session_launch::prepare_session_with_repo_url` (`deploy_validate/mod.rs:367`) — i.e. it **writes the agent roster into the workspace**. A gutted husk passes `is_dir()`, so spawn/resume-time repair recomposes the roster into it: the identical #4204 mechanism, in a path the issue's "two paths" framing does not cover. (Its io-error swallow is benign here — `is_dir() == false` *skips* the write, failing safe; only the gutted case is the gap.) Pre-existing and unchanged by this PR, so **not a merge blocker**. | Route this gate through `workspace_liveness` as well: `if liveness.is_dead() { return Ok(()) }`. Follow-up issue under #4204. |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs` | 551-554 | Prior LOW, **not addressed**. `"tm: workspace for session {id} is {liveness} …"` composed with `Absent`'s `Display` (`workspace_liveness.rs:117`, `"no longer exists on disk"`) renders as **"…is no longer exists on disk"**. `Gutted` reads fine; `Indeterminate` never reaches this branch. | Make the variants noun phrases (`"absent from disk"`), or reword the call site. |
| LOW | `crates/trusty-mpm/src/core/agent_reset_workspace.rs` | 50-55 | The module-level `Test:` list omits `sweep_reports_session_whose_workspace_vanished`, which the function-level list at `:147` does name. Cosmetic drift; the doc-pointer lint passes. | Add it to the module-level list. |

**Prior LOW on `provision_in` (`workspace_liveness.rs:141-147`) — WITHDRAWN; the
engineer was right to leave it alone.** I checked the mechanism myself. The prior
round asserted "provision_in sessions are recorded with `workspace_owned = false`
(`lifecycle.rs:992`)". That is wrong — `lifecycle.rs:992` is the **in-project
worktree** path, which never calls `provision_in`. The actual `provision_in` call
site (`daemon/managed_routes/lifecycle.rs:1212`) records
`true, // owned: we provisioned a fresh clone` at `lifecycle.rs:1254`, and that
is the only `owned = true` production site in the tree. Both the doc's cited
mechanism and its conclusion hold.

---

## Disposition of the prior round's findings

| Prior finding | Status at `a21ee67e` |
|---|---|
| HIGH — `is_dir()` collapses io errors into `Absent` | **FIXED**, mutation-proven, including the silent-`continue` half |
| MEDIUM — `guided_inplace:548` `workspace_owned` | **DEFERRED with a written rationale** (`:541-547`); still open, carried above |
| MEDIUM — `agent_reset_workspace` `?` aborts the sweep | **NOT ADDRESSED**; carried above, and now more reachable |
| LOW — `Display` grammar at the `guided_inplace` call site | **NOT ADDRESSED**; carried above |
| LOW — `provision_in` doc mechanism | **WITHDRAWN** — the prior finding's premise was mistaken |

Nothing vanished silently: two carried forward, one deferred with a rationale,
one withdrawn on re-examination.

---

## Fail-open / cursor-advance hunt

I deliberately looked for the repo's recurring shape — a failure branch that
advances state anyway so alarms read healthy — in and around this diff:

- `Indeterminate` → proceed, in **both** consumers
  (`agent_reset_workspace.rs:211-218`, `guided_inplace.rs:550`). This *is* a
  fail-open, and it is the correct one: it is the module's stated invariant, and
  it is **loud** — the sweep emits `tracing::warn!` with the reason on every
  occurrence. Not the silent-advance shape.
- No latch, counter, give-up path, or retry budget exists in this diff, so there
  is no cursor to advance past a failure.
- The one genuine sibling of the bug being fixed is MEDIUM-3
  (`deployment_check.rs:74`): the same `is_dir()`-as-liveness confusion, a
  different path, pre-existing.

## Notes

- `.git` is stat'd with `symlink_metadata` (`:194`), so a **dangling** `.git`
  symlink yields `Ok(_)` → `Live`. Intentional and documented (`:26-28`,
  `:151-155`: "file, directory, or symlink"), and it errs toward serving a
  workspace — the safe direction for this guard. Recorded as context, not a
  defect.
- `run_inplace_relaunch_serves_live_linked_worktree` still asserts
  `build_failed || daemon_contacted`. On a host without `claude` on PATH it
  degrades to the weaker `build_failed` arm — non-vacuous either way (a gate
  refusal returns `FallThrough` with zero hits, which is neither arm), just
  weaker on CI than locally. Same caveat as last round; still acceptable.
- The CHANGELOG conflict was resolved correctly — the #4204 entry sits under
  `## [Unreleased]`, not under `## [1.2.0]`.
- Verification ran in a throwaway worktree at `refs/pull/4209/head`; all
  mutations were reverted and the tree left clean.

**APPROVE.** Zero CRITICAL, zero HIGH. The HIGH that gated the last round is
genuinely closed and provably guarded. The three MEDIUMs are follow-ups: one is a
pre-existing sibling outside this PR's scope, one is consciously deferred with a
written rationale, and one is a robustness gap in a manual sweep command that
fails loudly. None should hold the 1.2.3 release.

Issue #123, opened 2026-05-25
test(trusty-search): canonical benchmark corpus contains query strings as literals — circular bias


bobmatnyc:
## Summary

The canonical 14-query equivalency benchmark runs against the `trusty-tools` index — which IS the source code that contains the test files. The test files (`crates/trusty-search/src/core/classifier.rs`, `crates/trusty-search/src/core/indexer/tests.rs`) literally contain the benchmark query strings as `assert_eq!` parameters and unit-test inputs. BM25 picks them up with very high term frequency, contaminating result rankings.

Discovered during v0.8.3 validation (post-#119 reindex with `include_docs=true`): for nearly every benchmark query, `classifier.rs` ranks at #1-6 because the file contains `assert_eq!(QueryClassifier::classify("apply_archive_downrank"), Intent::Definition)` and similar assertions for every other benchmark string.

## Why this matters

1. **Benchmark validity is compromised.** Every metric we measure on trusty-tools is biased by the fact that the test fixtures and the system-under-test live in the same repo. We can't honestly answer "did the Definition boost surface the right file" when the test file containing the query string outranks every other file.
2. **The bias amplifies with #119.** With more queries now classified as Definition, the BM25 lane is heavier, and string-literal occurrences in test files get more weight. So #119 made the benchmark LESS reliable, not just the search.
3. **All prior benchmarks may be retroactively suspect.** The 76.9% Hit@1 from v0.8.1, the 100% from grep/rg — these were all measured against a corpus that contains the queries as string literals. Some of the "Hit@1" wins may be incidental.

## Proposed remediation

Pick one (or combine):

**Option A — Use a different corpus as the benchmark target.** Index `open-mpm` (sibling Rust project in the workspace) and run queries about open-mpm's own symbols against that index. The benchmark suite becomes "this is how trusty-search performs on a generic Rust workspace it has never seen before."

**Option B — Synthetic benchmark fixture.** Create a small `crates/trusty-search/tests/benchmark_corpus/` directory with hand-crafted Rust files containing planted symbols (`Hammond_Lever`, `Lichtenberg_Filter`, etc. — distinctive, never-appear-in-real-code names). Index this corpus, run queries against it. Fully isolated, fully reproducible.

[...]

PR #4212, opened 2026-07-28
fix(trusty-review): refuse to confirm registry claims nothing looked up (#4081)

### LINKED ISSUE #4081: trusty-review: hallucinated dependency-version finding marked code_provable+verified drives a false BLOCK/F (ruff==0.16.0 claimed nonexistent; it is the latest release)
## Summary

`review_pr` produced a **high-confidence, `verified: "confirmed"`, `code_provable: true` false positive** claiming a pinned dependency version does not exist on the package registry. The claim came from the reviewer model's stale training knowledge; no registry lookup was performed. The finding was severe enough to drive the whole review to `verdict: BLOCK` / `grade: F`, which would have blocked a perfectly correct PR.

## Reproduction

```
mcp__trusty-review__review_pr(owner="duettoresearch", repo="APEX", pr=1717)
```

PR head SHA `f8fc823259838513cdab7f59dff72b56e4dcf075`. Reviewer model `us.anthropic.claude-sonnet-4-6` (Bedrock), `review_version: tr-0.1`, dry-run.

The PR diff contains a one-line CI change:

```diff
-        run: pip install "pyyaml>=6.0" "jsonschema>=4.0" "pytest>=7.0" "ruff>=0.6.0"
+        run: pip install "pyyaml>=6.0" "jsonschema>=4.0" "pytest>=7.0" "ruff==0.16.0"
```

## Actual output

```json
{
  "category": "correctness",
  "code_provable": true,
  "confidence": 0.82,
  "verified": "confirmed",
  "kind": "ruff pinned to non-existent version 0.16.0",
  "file": ".github/workflows/repos-all-advisory.yml",
  "line": 41,
  "description": "The CI workflow pins `ruff==0.16.0`, but ruff's versioning scheme uses a `0.x.y` format where the latest stable releases are in the `0.4.x`–`0.9.x` range as of mid-2026. Version `0.16.0` does not exist on PyPI. This will cause the `pip install` step to fail with a \"No matching distribution found\" error, breaking CI for every PR that touches the linted files.",
  "consequence": "CI pip install step fails with 'No matching distribution found', breaking the advisory workflow on every run"
}
```

Rolled up to: `"verdict": "BLOCK"`, `"grade": "F"`, with `grade_justification` citing this finding as "a concrete, provable failure that needs to be fixed before merge."

## Expected

No finding. `ruff==0.16.0` is the **current latest release** on PyPI:

```
$ GET https://pypi.org/pypi/ruff/json
info.version == "0.16.0"          # latest
releases["0.16.0"]                 # present
```

Independent confirmation from the PR itself: the `Lint` job in that same workflow installs `ruff==0.16.0` and runs `ruff check` — **all 10 CI checks on PR #1717 pass**. The reviewer asserted a CI-breaking install failure against a workflow run that was green in the same fetch.

## Root cause (hypothesis)

Two compounding problems:

1. **Ungrounded package-version claims.** The model reasoned from its training cutoff ("as of mid-2026 the latest stable releases are in the `0.4.x`–`0.9.x` range") and treated that recollection as fact. Package registries move continuously; a model's version knowledge is stale by construction. Nothing in the pipeline grounds the claim.

2. **The confidence/verification signals are unearned.** `code_provable: true` and `verified: "confirmed"` were both set on a claim that was never checked against anything. Those fields are exactly the signals a consumer uses to decide a finding is trustworthy, so a hallucination wearing them is worse than no finding at all — it inverts the trust signal.

## Suggested fixes

- **Ground or suppress registry claims.** Any finding asserting that a package/version does not exist, is yanked, or is outdated should require an actual registry lookup (PyPI/npm/crates.io JSON API) before it can be emitted. If the lookup is unavailable, downgrade to an advisory note — never `code_provable: true`.
- **Cross-check against CI status.** The review already has the PR context; a finding predicting "CI fails on every run" that contradicts an observed green required check should be auto-demoted or dropped.
- **Gate `verified: "confirmed"`.** Reserve it for findings backed by something external to the model's own assertion.
- **Add a class-level guard for time-sensitive assertions** (versions, deprecations, "as of <date>" reasoning) — these are the highest-yield hallucination category in dependency diffs.

## Secondary issue (related, possibly same root as #4044)

This run also returned `status: "degraded"` with `error: "trusty-search unavailable at http://localhost:7878; review produced WITHOUT code context"` — and still emitted a hard `BLOCK` / `F`. A run that self-reports as non-authoritative should not be able to produce a blocking verdict; `UNKNOWN` would be the honest output. #4044 covers grade/verdict not being derived from the finding set; this is the adjacent case of grade/verdict not being derived from *run health*.

(The trusty-search outage itself is separately visible in `list_recent_errors` as fingerprint `8cfeb46100e45acfffc67c35e696a774f0748ed71cd233b0114f80180c8f9577` — redb corpus open failure for `duettoresearch-apex-tm-apex-01`.)

## Impact

Medium-high. A false `BLOCK` on dependency bumps is directly counterproductive: dependency-update PRs are the single most common diff shape where a reviewer's version knowledge is guaranteed stale, so this failure mode is systematically concentrated where the tool is least reliable. Reviewers who see one confident-but-wrong `code_provable` finding discount the rest.

## Environment

- trusty-review MCP, `review_version: tr-0.1`, dry-run (`PR_INTELLIGENCE_DRY_RUN=true`)
- reviewer model: `us.anthropic.claude-sonnet-4-6` on Bedrock
- macOS aarch64, trusty-search 0.39.1 (down at time of run)
- observed 2026-07-27

bobmatnyc:
## Summary

`review_pr` emitted a `code_provable: true`, `verified: "confirmed"`,
`confidence: 0.82` finding claiming a pinned dependency version does not exist,
and rolled it up to `verdict: BLOCK` / `grade: F`. The version was the current
latest release, and all ten checks on the reviewed PR were green on that exact
dependency.

Closes #4081.

## Root cause

Two things went wrong, and only the second is fixable:

1. **The model recalled a stale fact.** Unfixable by construction — a package
   registry moves continuously, so a reviewer model's version knowledge is stale
   the day it ships. #4081's own finding text says so out loud: *"the latest
   stable releases are in the `0.4.x`–`0.9.x` range **as of mid-2026**"*.
2. **The claim wore the pipeline's two strongest trust signals without anything
   having checked it.** Nothing in trusty-review performs a registry lookup —
   `code_provable` is self-declared by the model, and `verified: "confirmed"`
   came from the verifier round, which is a second LLM with the same
   training-cutoff problem. **This is the defect.** A hallucination wearing a
   confirmation is worse than no finding at all: it inverts the exact signal
   consumers use to decide what to trust.

And it lands where it does the most damage. A dependency-bump diff is the one
shape where the reviewer's version knowledge is guaranteed stale, and the shape
whose author most expects to sail through. This repo's own open version-bump PR
#4177 was a prime target.

## Fix shape — refuse to confirm, do not look up

Actually querying PyPI/npm/crates.io was considered and **rejected**. It buys
nothing the reported harm needs (the false BLOCK is cured by declining to mark
an unchecked claim as checked) and it adds a live network dependency, rate
limits, per-ecosystem client code, and a fresh way for the review path itself to
hang or fail. The principle encoded here is narrower and stronger than any
lookup:

> A claim may be marked `verified: "confirmed"` only when something actually
> verified it.

New `pipeline::claim_grounding`, following #4057's precedent of one module per
distinct signal (`citation_check` = path/content verification, `finding_hygiene`
= self-admission markers, `claim_grounding` = registry vocabulary). It is wired
into `finding_hygiene::sanitize_findings` — the single seam #4057 established
for exactly this — so the unified (`runner.rs`) and map-reduce
(`mapreduce/mod.rs`) paths both get it with no call-site changes.

A finding is demoted when its text pairs a **registry subject** (PyPI, npm,
crates.io, `requirements.txt`, `package.json`, `pip install`, …) with a
**registry-state assertion** (version absent, withdrawn, deprecated, "latest
release is…", "as of my knowledge cutoff"). Requiring both halves is what keeps
the guard off an ordinary correctness finding that happens to say "does not
exist" about a code symbol.

Demotion, in full:

| field | change | why |
|---|---|---|
| `verified` | → new `VerifyOutcome::Unverifiable { reason }` | pre-stamped BEFORE the verification round; `select_candidates` now skips a finding whose outcome is already decided, so the verifier LLM is never asked a question it cannot answer and can never stamp `Confirmed` over it |
| `code_provable` | → `false` | a registry fact is not provable from the diff; drops it out of `grade::drives_block_floor` |
| `effort` | capped at `Medium` | same demotion `demote_diff_absent_speculation` applies (#4043) |
| `confidence` | capped at `0.65` | at/above the review-include gate (0.60) so it stays visible; at/below `grade.rs`'s advisory-collapse line (0.65), which is what dissolves the model's **own** self-reported BLOCK — the severity floor alone cannot, since `stricter_of(model, floor)` only ever raises |
| `description` | advisory note appended | the author is told nothing checked it, and to confirm the pin themselves |

Reusing `grade.rs`'s existing low-confidence override rather than adding a
parallel relaxation is deliberate — #4057's lesson was that
mechanism-per-defect proliferation is the thing to avoid.

**It never drops the finding.** "trusty-review could not check this version" is
real information for the author. Over-suppression would trade a false BLOCK for
a false silence.

## Proof by execution

`tests/registry_claim_4081_regression.rs` uses #4081's finding verbatim, field
for field. Verified failing before the fix: with
`demote_ungrounded_registry_claims` neutered to a no-op, **4 of 5 fail** —
verdict stays BLOCK, the claim is still selected as a verification candidate,
and no advisory note reaches the reader. The 5th is the control, which passes in
both states by design.

| test | asserts |
|---|---|
| `ungrounded_version_claim_is_not_confirmed_and_does_not_block` | not `Confirmed`; `derive_verdict(APPROVE, …)` → APPROVE **and** `derive_verdict(BLOCK, …)` → APPROVE (the self-reported BLOCK dissolves too) |
| `ungrounded_version_claim_is_never_sent_to_the_verifier` | was a candidate before, is not after |
| `ungrounded_version_claim_still_surfaces_as_advisory` | **presence**, not absence — file/line/kind intact, original claim text verbatim, note added, confidence non-zero |
| `verification_backed_finding_still_confirms_and_still_blocks` | the capability is retained: a genuine finding is untouched, still verified, still BLOCKs |
| `a_genuine_finding_alongside_a_registry_claim_still_blocks` | the demotion is scoped — a real defect on the same PR keeps full weight, so the guard cannot smuggle a bug past review inside a dependency bump |

Plus 10 unit tests in `claim_grounding_tests.rs`: every assertion marker fires;
an ordinary "does not exist" code finding does not; a manifest finding that
argues from the diff does not; the cap is a ceiling never a promotion; the pass
is idempotent; and the advisory note provably does not trip `finding_hygiene`'s
own marker sets.

No `#[ignore]`, no cfg-gates.

## Not fixed here (reported, not silently skipped)

- **The CI cross-check (#4081 suggestion 2) is not implemented.** trusty-review
  fetches PR metadata and the diff — it never fetches check runs, so there is no
  observed CI status "already in hand" to cross-check against. Adding one means
  a new network call in the review path, which is the same dependency this PR's
  central design decision declines to take on. Worth its own issue.
- **`source_citation` is grammar-checked, never resolved.**
  `grade::is_escalation_eligible` accepts any string matching
  `CITATION_GRAMMAR_RE` (`path:line` / `TICKET-123` / `#123` / `§ 4.2`). Nothing
  confirms the ticket exists or says what the finding claims — so a
  well-formatted invented citation still unlocks the BLOCK floor. Same defect
  family as this one.
- **`status: "degraded"` still permits a hard BLOCK.** The reported run
  self-declared `"review produced WITHOUT code context"` and emitted BLOCK/F
  anyway; `runner.rs` only prepends a banner and sets `error`, leaving the
  verdict untouched. #4081's own "Secondary issue" section, adjacent to #4044.

## Verification

- `cargo test --workspace` (no `--lib`, CI's `ci.yml` exclusion list for the
  three Tauri GUI crates), `cargo clippy --workspace --all-targets -D warnings`
  (same exclusions), `cargo fmt --all --check`, `scripts/check_line_cap.sh` — all
  green. Raw output in the PR thread.

### DIFF
--- a/crates/trusty-review/CHANGELOG.md
+++ b/crates/trusty-review/CHANGELOG.md
@@ -9,6 +9,41 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **A hallucinated dependency-version finding wore `code_provable: true` +
+  `verified: "confirmed"` and drove a false `BLOCK`/`F`** (closes
+  [#4081](https://github.com/bobmatnyc/trusty-tools/issues/4081)).
+  - A review of a one-line CI change (`ruff>=0.6.0` → `ruff==0.16.0`) reported
+    that `0.16.0` "does not exist on PyPI", reasoning explicitly from the
+    reviewer model's training cutoff. It was the current latest release, and all
+    ten checks on that PR were green on that exact dependency. Nothing in the
+    pipeline performs a registry lookup, so nothing had checked the claim — yet
+    it carried both signals a consumer reads to decide a finding is trustworthy,
+    and hard-clamped the verdict to `BLOCK`/`F`. A hallucination wearing a
+    confirmation is worse than no finding at all, and this failure mode is
+    concentrated on dependency bumps — the one diff shape where a reviewer's
+    version knowledge is stale by construction.
+  - New `pipeline::claim_grounding` pass (wired into
+    `finding_hygiene::sanitize_findings`, so both the unified and map-reduce
+    paths get it) demotes any finding whose text makes a package-registry
+    assertion — version absent, withdrawn, deprecated, "latest release is…",
+    "as of my knowledge cutoff" — to the advisory tier: `code_provable` cleared,
+    `effort` capped at Medium, `confidence` capped at the advisory band, and
+    `verified` pre-stamped with a new `VerifyOutcome::Unverifiable { reason }`.
+    `verify::select_candidates` now skips any finding whose outcome is already
+    decided, so such a claim is never put to the verifier LLM — a second model
+    with the same stale training knowledge would only launder the recollection
+    into a `"confirmed"`.
+  - **Refuse-to-confirm, not look-up.** Querying PyPI/npm/crates.io was
+    considered and rejected: it buys nothing the reported harm needs, and adds a
+    live network dependency, rate limits, and a new way for the review path
+    itself to hang. The principle encoded is stronger than any lookup — a claim
+    may be marked `verified: "confirmed"` only when something actually verified
+    it.
+  - **The finding is never dropped.** It still surfaces, verbatim, with an
+    advisory note telling the author nothing checked it and to confirm the pin
+    themselves. Only its grading weight is removed. A genuine, diff-provable
+    finding alongside it keeps full weight and still drives `BLOCK`.
+
 - **A live trusty-search was reported as `"unreachable"`, and a degraded verdict
   was indistinguishable from a complete one** (closes
   [#4086](https://github.com/bobmatnyc/trusty-tools/issues/4086)).
--- a/crates/trusty-review/src/models/mod.rs
+++ b/crates/trusty-review/src/models/mod.rs
@@ -181,6 +181,29 @@ pub enum VerifyOutcome {
     /// The finding was below the verification confidence threshold; not sent
     /// to the verifier.
     Skipped,
+    /// The finding makes a claim NOTHING in this pipeline can verify, so it was
+    /// deliberately never sent to the verifier and may never be marked
+    /// `Confirmed` (#4081).
+    ///
+    /// Why: `Confirmed` is the trust signal consumers read to decide a finding
+    /// is real. A package-registry assertion ("version X does not exist", "X is
+    /// deprecated") rests on the model's training-cutoff recollection, and the
+    /// verifier is the SAME kind of oracle — asking it launders a stale
+    /// recollection into a confirmation. Recording the un-verifiability
+    /// explicitly is honest where `Skipped` (which means "below the confidence
+    /// threshold") would be a lie about why the verifier was not consulted.
+    /// What: set by `pipeline::claim_grounding::demote_ungrounded_registry_claims`
+    /// BEFORE the verification round; `verify::select_candidates` skips any
+    /// finding whose outcome is already decided, so it can never be overwritten
+    /// with `Confirmed`. Unlike the refutation variants it is NOT excluded from
+    /// the verdict floor — the finding survives as an advisory note (it is the
+    /// demotion of `effort`/`confidence`/`code_provable` that keeps it from
+    /// escalating), because "we could not check this" is not "this is false".
+    Unverifiable {
+        /// Human-readable reason the claim is unverifiable (e.g.
+        /// `"package-registry version claim; no registry lookup performed"`).
+        reason: String,
+    },
 }
 
 // ─── Finding category ──────────────────────────────────────────────────────────
--- a/crates/trusty-review/src/pipeline/claim_grounding.rs
+++ b/crates/trusty-review/src/pipeline/claim_grounding.rs
@@ -1,0 +1,289 @@
+//! Grounding guard for package-registry / version-existence claims (#4081).
+//!
+//! Why: `review_pr` emitted a `code_provable: true`, `verified: "confirmed"`,
+//! `confidence: 0.82` finding asserting that a pinned dependency version does
+//! not exist —
+//!
+//! > "The CI workflow pins `ruff==0.16.0`, but ruff's versioning scheme uses a
+//! > `0.x.y` format where the latest stable releases are in the `0.4.x`–`0.9.x`
+//! > range as of mid-2026. Version `0.16.0` does not exist on PyPI."
+//!
+//! — and rolled it up to `verdict: BLOCK` / `grade: F`. `ruff==0.16.0` was the
+//! CURRENT latest release, and all ten checks on the PR were green on that exact
+//! dependency. The claim came from the reviewer model's training-cutoff
+//! recollection; **nothing in this pipeline performs a registry lookup**, so
+//! nothing verified it.
+//!
+//! Two things went wrong, and only the second is fixable here:
+//!
+//!  1. The model recalled a stale fact. Unfixable by construction — a package
+//!     registry moves continuously, and a model's version knowledge is stale the
+//!     day it ships.
+//!  2. The claim wore the pipeline's two strongest trust signals (`code_provable:
+//!     true`, `verified: "confirmed"`) without anything having checked it, and
+//!     those signals let it hard-clamp the verdict. **This is the defect.** A
+//!     hallucination wearing a confirmation is worse than no finding at all: it
+//!     inverts the signal consumers use to decide what to trust.
+//!
+//! And it lands precisely where it does the most damage. A dependency-bump diff
+//! is the single shape where the reviewer's version knowledge is guaranteed
+//! stale, and the shape whose author most expects to sail through.
+//!
+//! ## What: refuse to confirm, do not look up
+//!
+//! The obvious alternative — actually query PyPI/npm/crates.io — was considered
+//! and rejected. It buys nothing the reported harm needs: the false BLOCK is
+//! cured by declining to mark an unchecked claim as checked, and a lookup adds a
+//! live network dependency, rate limits, per-ecosystem client code, and a fresh
+//! way for the review path itself to hang or fail. The general principle this
+//! module encodes is narrower and stronger than any lookup:
+//!
+//! > **A claim may be marked `verified: "confirmed"` only when something actually
+//! > verified it.**
+//!
+//! So [`demote_ungrounded_registry_claims`] finds findings whose own text makes a
+//! registry-currency assertion and demotes each to the advisory tier:
+//!
+//!  - `verified = Unverifiable { reason }` — pre-stamped BEFORE the verification
+//!    round, and `verify::select_candidates` skips findings whose outcome is
+//!    already decided, so the verifier LLM is never asked a question it cannot
+//!    answer and can never stamp `Confirmed` over it.
+//!  - `code_provable = false` — a registry fact is by definition not provable
+//!    from the diff. This alone drops the finding out of
+//!    `grade::drives_block_floor`.
+//!  - `effort` capped at `Medium` — the same demotion
+//!    `finding_hygiene::demote_diff_absent_speculation` applies (#4043).
+//!  - `confidence` capped at [`UNGROUNDED_CLAIM_ADVISORY_CONFIDENCE`] — see that
+//!    constant for why the exact value matters to `grade.rs`.
+//!  - an advisory note appended to `description`, so a human reading the review
+//!    is told the claim is unchecked rather than silently handed a weaker one.
+//!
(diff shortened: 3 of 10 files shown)


bobmatnyc:
## Gate output (raw)

### `cargo test --workspace` (no `--lib`)

CI's own `ci.yml` exclusion list substituted for the three Tauri GUI crates
(`trusty-mpm-gui`, `trusty-code-gui`, `trusty-agents-ui` — WebKit2GTK /
`frontendDist`; excluded from every headless workspace job in CI, disclosed here
per the same rationale). `-j 3` to fit local memory.

```
SKIP_UI_BUILD=1 cargo test --workspace -j 3 \

suites=9 passed=3768 failed=0 ignored=12
```

**`trusty-review` — all green:**

```
test result: ok. 1555 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out; finished in 7.46s
test result: ok. 68 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.24s
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.27s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.51s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s
```

**The new suite:**

```
     Running tests/registry_claim_4081_regression.rs

running 5 tests
test verification_backed_finding_still_confirms_and_still_blocks ... ok
test ungrounded_version_claim_is_never_sent_to_the_verifier ... ok
test ungrounded_version_claim_is_not_confirmed_and_does_not_block ... ok
test ungrounded_version_claim_still_surfaces_as_advisory ... ok
test a_genuine_finding_alongside_a_registry_claim_still_blocks ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

**Failing before the fix** — `demote_ungrounded_registry_claims` temporarily
neutered to `return 0`, same binary otherwise:

```
---- ungrounded_version_claim_is_not_confirmed_and_does_not_block stdout ----
assertion `left == right` failed: the registry claim must be demoted
  left: 0
 right: 1

---- ungrounded_version_claim_is_never_sent_to_the_verifier stdout ----
a claim the pipeline cannot check must not be put to the verifier

---- ungrounded_version_claim_still_surfaces_as_advisory stdout ----
the reader must be told nothing verified the claim: The CI workflow pins
`ruff==0.16.0`, but ruff's versioning scheme uses a `0.x.y` format where the
latest stable releases are in the `0.4.x`–`0.9.x` range as of mid-2026. Version
`0.16.0` does not exist on PyPI. [...]

---- a_genuine_finding_alongside_a_registry_claim_still_blocks stdout ----
assertion `left == right` failed: only the registry claim
  left: 0
 right: 1

test result: FAILED. 1 passed; 4 failed; 0 ignored
```

The 1 pass is `verification_backed_finding_still_confirms_and_still_blocks` —
the control, which must pass in both states.

### Pre-existing failures, unrelated to this PR

`trusty-agents` `api_e2e` fails 4/4 locally — the test's own API server never
becomes healthy inside its 5s budget:

```
thread 'test_tasks_list_starts_empty' panicked at crates/trusty-agents/tests/api_e2e.rs:49:43:
spawn api server: api server did not become healthy at http://127.0.0.1:50627/api/health within 5s
```

Confirmed **pre-existing** by running the same suite on unmodified `origin/main`
in this worktree:

```
test result: FAILED. 0 passed; 4 failed; 3 ignored; 0 measured; 0 filtered out; finished in 5.33s
MAIN_API_E2E_EXIT=101
```

`trusty-agents` does not depend on `trusty-review` (0 references in its
`Cargo.toml`), so this PR cannot reach it. A fifth failure in the first run —
`tools::python_skill::tests::execute_dispatches_to_python_and_parses_json`, a
local `uv`/`/opt/homebrew/bin/python3` cache error — passed on re-run
(`8 passed; 0 failed`), i.e. machine state, not code.

### `cargo clippy` (same exclusions)

```
SKIP_UI_BUILD=1 cargo clippy --workspace --all-targets -j 4 \
  --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui \
  -- -D warnings

EXIT=0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 59.82s
```

### `cargo fmt --all --check`

```
FMT OK
```

### `scripts/check_line_cap.sh`

```
```

bobmatnyc:
## CI: all green

Full check run on `3ecb594e`: **24 checks, all pass** (`gh pr checks --watch --fail-fast` → exit 0). `Test` 16m46s, `Clippy` 5m40s, `MSRV check (1.91)` 6m15s, `Format check`, `500-line file-size cap`, `SLD (DOC-38) lint`, `Test: doc-comment pointer lint`, all 7 UI-check legs, `trusty-search daemon smoke test`.

### One red on the first run, fixed properly

`Test: doc-comment pointer lint` failed on `e474b466`:

```
FAIL: crates/trusty-review/src/pipeline/verify.rs:379: Test: pointer cites
`select_candidates_skips_findings_with_a_decided_outcome` — no
`fn select_candidates_skips_findings_with_a_decided_outcome(` or
`mod select_candidates_skips_findings_with_a_decided_outcome` found in crate
`crates/trusty-review`.
test-pointers: 1 dangling pointer(s), 0 stale allowlist entries — FAILED.
```

The lint was right, and it caught a real coverage gap rather than a typo: the
`verified.is_none()` guard is the mechanism that keeps a pre-stamped
`Unverifiable` claim away from the verifier, and it had only transitive coverage
from the #4081 regression suite. Fixed by **writing the test**, not by editing
the pointer — `3ecb594e` adds
`select_candidates_skips_findings_with_a_decided_outcome` to `verify_tests.rs`,
asserting both indices are candidates before the stamp and only the untouched
one after.

```
test pipeline::verify::tests::select_candidates_skips_findings_with_a_decided_outcome ... ok
test result: ok. 54 passed; 0 failed; 0 ignored; 0 measured; 1507 filtered out
```

```
PTR_EXIT=0
```

Issue #4213, opened 2026-07-28
trusty-search: P0 data-loss guard test is green-by-luck (#2178 root-hijack test races on process-global env var)


bobmatnyc:
Defect: `crates/trusty-search/src/service/reindex/root_hijack_tests.rs` (around lines 86-89) isolates via `unsafe { std::env::set_var("TRUSTY_DATA_DIR", …) }` plus `#[serial]`. `#[serial]` only excludes OTHER `#[serial]` tests in the same binary — it does not prevent a concurrent non-serial test in that binary from observing or racing the process-global env var.

Evidence: `reindex_refuses_untrusted_root_move_and_preserves_corpus` failed once locally under a full `cargo test --workspace` run (`left: Complete, right: Failed`), then passed on re-run and passed in CI. So it is intermittently red under parallel execution.

Why it matters: this is the guard for the **#2178 P0 data-loss** behavior. A test that is green-by-luck is indistinguishable from green-by-correctness, and an intermittently-red security/data-loss guard is exactly the kind of signal that gets waved through as "the known flake."

Resolution: replace process-global env mutation with per-test isolation that does not depend on `#[serial]` covering the whole binary.

Issue #2336, opened 2026-07-10
fix(trusty-search): create_index/relocate_index lack root_path collision guard — can recreate the #2305 double-registration at runtime


bobmatnyc:
Follow-up to #2305 (warm-boot dedup, merged). Review finding (HIGH): `create_index_handler` (service/server/indexes.rs:127,255) guards only id collision, never root_path collision, and never checks `corpus_open_failed` after `build_indexer_from_entry` — so two live create_index calls (or create_index + relocate_index, indexes_relocate.rs:100) against the same colocated root register two ids sharing one index.redb RIGHT NOW, in-process, with the failure silently swallowed and 200 {"created": true} returned. This is almost certainly how the apex/trusty-mpm duplicate was created, and it can recreate the exact #2305 crash-loop after the warm-boot fix. Fix: reject create/relocate when the canonical root_path already belongs to a different registered id (409), and/or surface corpus_open_failed post-build as an error instead of silently registering a broken handle. Effort: S-M.

PR #4218, opened 2026-07-28
fix(trusty-search): safe deregister on DELETE /indexes/:id (#4123) + restore-derived stages on POST /indexes (#4110)

### LINKED ISSUE #4110: POST /indexes initializes stage status to Pending unconditionally, even when colocated corpus restore succeeded
## Summary

When registering an existing colocated corpus via `POST /indexes` (create_index), the index adopts its data correctly and comes up fully initialized. However, `create_index_handler` in `crates/trusty-search/src/service/server/indexes.rs` sets stage status to `lexical: pending(), semantic: pending()` unconditionally, regardless of what actually restored. This causes semantic search to hard-error and search_all to degrade to BM25-only, even though the index is fully ready.

## Evidence

**Live verification today (2026-07-27):** Index `cto-duetto` registered via create_index:
- Corpus restored successfully: chunk_count = 200090 (exact), last_walk_started_at = null, status = ready
- Expected: stage status reflects successful restore (semantic stage ready)
- Actual: semantic search against the index hard-errors with "requires Stage 2 (embeddings), which is not yet ready"
- Workaround: daemon restart clears the Pending state and allows semantic search
- Symptom: search_all silently degrades to BM25-only with every hit showing match_reason="bm25" — indistinguishable from a genuinely dead vector lane

## Root Cause

The warm-boot path (`crates/trusty-search/src/service/server/lifecycle.rs` or similar) derives stage state from the restore result:

```
warm-boot: index 'X' restored (colocated=true) — chunks=N hnsw_snapshot=true/false → stages(...)
```

But the registration path (`create_index_handler`) does not — it sets stages unconditionally to Pending regardless of whether the restored index has a complete hnsw snapshot.

## Expected Behavior

After `POST /indexes` registers a colocated corpus with hnsw snapshot, the semantic stage should be marked ready immediately (or at least set to a state that reflects the actual restore success), not forced to Pending.

## Suggested Fix

Derive stage status from the restore result in the registration path, as the warm-boot path already does. Do not set stages unconditionally to Pending.

## Acceptance Criteria

- [ ] After create_index registers a colocated corpus, verify that hnsw snapshot status is checked
- [ ] If hnsw snapshot is present and valid, semantic stage is marked ready (or equivalent success state)
- [ ] If hnsw snapshot is absent or invalid, semantic stage remains pending
- [ ] Semantic search no longer hard-errors on a freshly-registered, fully-intact index
- [ ] search_all continues to include vector search results (not silently BM25-only)
- [ ] Tests verify stage derivation for colocated restores with and without hnsw snapshot

## Files

- `crates/trusty-search/src/service/server/indexes.rs` — create_index_handler
- `crates/trusty-search/src/service/server/lifecycle.rs` (or warm-boot equivalent) — reference implementation for correct stage derivation

## Related Issues

- trusty-search semantic search vector lane degradation
- Possible correlation with #3935 (if corpus registration was recently changed)

### LINKED ISSUE #4123: trusty-search: DELETE /indexes/:id hardcodes delete_data=true — no way to deregister safely without destroying data
## Problem

`unregister_index` (crates/trusty-search/src/service/server/search.rs) hardcodes `delete_data=true`. The HTTP surface exposes no parameter to request deregistration while preserving the on-disk data directory.

Only the internal orphan-reaper ticker uses the safe variant — and its own log line advertises exactly that safe behavior:
```
orphan-reaper: unregistered index '<id>' — root_path deleted (data preserved)
```

## Impact Hit Live Today

An operator needing to remove 49 stale registry entries could not safely use the API at all, because a mistake or a mis-targeted id would have destroyed real corpus data. The work had to be done by hand-editing `indexes.toml` with the daemon stopped — riskier and unscriptable.

Registry hygiene is a recurring need (see #4094, #4095), so the absence of a safe deregistration verb has ongoing cost.

## Suggested Fix Direction

Expose `delete_data` as an explicit request parameter defaulting to `FALSE` (deregister only), so destroying data requires opting in.

Related: #4094, #4095

bobmatnyc:
Fixes two defects in trusty-search's index-lifecycle HTTP handlers. Both were verified against current `main` before editing; no drift was found in the file/line references given in the issues, except that #4110 names `service/server/lifecycle.rs` as the reference implementation — that file does not exist; the real warm-boot classifier is `service/warm_boot/stages.rs::derive_warm_boot_stages`, driven from `commands/start_restore.rs:342`.

---

## ⚠️ BREAKING BEHAVIOUR CHANGE — `DELETE /indexes/:id` (#4123)

**Before:** every `DELETE` destroyed the index's on-disk data directory.
**After:** a bare `DELETE` deregisters ONLY. Destroying data requires `?delete_data=true`.

Any caller that relied on `DELETE` reclaiming disk will silently stop reclaiming it and leave orphaned data behind. Every in-tree caller was audited and is listed below.

`unregister_index` already took a `delete_data: bool` and the orphan-reaper already passed `false` (the `"root_path deleted (data preserved)"` log line). The handler simply hardcoded `true`; it now passes the request's flag.

### The default was already documented as `false` in three places

The destructive behaviour was not just unsafe, it contradicted every user-facing description of this endpoint:

| Artifact | What it already claimed |
|---|---|
| `crates/trusty-search/CLAUDE.md` (API reference) | "On-disk redb data is preserved — re-registering with the same id will reuse it." |
| `trusty-search index remove --help` (`main.rs:1013`) | "The on-disk redb / HNSW snapshot is preserved — re-registering with the same path reuses it." |
| UI delete confirm dialog (`ui/src/lib/views/Indexes.svelte:173`) | ``Delete index "${id}"? On-disk data is preserved.`` |

So `false` does not merely match the issue's stated intent — it restores the contract the product has been advertising all along. The UI has been telling operators their data was safe while the daemon deleted it.

### Every caller of `DELETE /indexes/:id`, and what was done

| # | Caller | Depends on destruction? | Action |
|---|---|---|---|
| 1 | `delete_index` MCP tool (`mcp/tools/index.rs:114`) | **Yes** — descriptor says "Delete a registered index and all its data" | **Updated** → `?delete_data=true` |
| 2 | `trusty-search cleanup` (`commands/cleanup.rs:143`) | **Yes** (reclamation intent) | **Updated** → `?delete_data=true`. Zero risk: every candidate is 0-chunk by construction. |
| 3 | trusty-mpm decommission GC (`session_manager/search_gc.rs:113`) | **Yes** | **Updated** → `?delete_data=true` |
| 4 | trusty-mpm orphan sweep (`session_manager/search_gc.rs:~325`) | **Yes** | **Updated** → `?delete_data=true` |
| 5 | Benchmark harnesses ×5 (`tests/benchmark_synthetic.rs` ×2, `benchmark_open_mpm.rs` ×2, `benchmark_open_mpm_kg.rs` ×1) | **Yes** — docs require "a clean slate" | **Updated** → `?delete_data=true` |
| 6 | `trusty-search index remove` (`commands/index_remove.rs:65`) | **No** — its `--help` promises preservation | **Left on the new default** (now truthful) |
| 7 | UI single-row + bulk delete (`ui/src/lib/api.js:69` → `Indexes.svelte:178`, `:258`) | **No** — dialog promises preservation | **Left on the new default** (now truthful) |
| 8 | Docs: `crates/trusty-search/CLAUDE.md:157`, `docs/trusty-search/spec/ARCHITECTURE.md:290` | — | **Updated** to document the parameter + the change |

For 3 and 4 the workspace being GC'd is a disposable worktree that is about to be deleted, so nothing will ever re-register at that root — preserving its data would leak it permanently.

Also added: the response now carries `data_deleted` so a caller can confirm which semantics ran. An unparseable `?delete_data=maybe` is rejected with `400` rather than silently guessed — a destructive toggle should never be inferred.

---

## Fix: `POST /indexes` ignored the restore outcome (#4110)

`create_index_handler` set `lexical: pending(), semantic: pending()` unconditionally, discarding the result of the colocated restore `build_indexer_from_entry` had just performed. Since `search_capabilities` is derived from `stages`, a fully intact index came up advertising no vector lane: semantic search hard-errored with "requires Stage 2 (embeddings), which is not yet ready" and `search_all` degraded to BM25-only with every hit reporting `match_reason="bm25"` — indistinguishable from a genuinely dead lane. Only a daemon restart cleared it.

Stages are now derived with `derive_warm_boot_stages` — the same pure classifier warm-boot and lazy-restore use, over the same three signals read the same way (`chunk_count`, `hnsw_load_failed` + snapshot presence, symbol-graph node count). Registration and warm-boot can no longer disagree about identical on-disk state, and the `skip_kg` / `skip_vector` / `lexical_only` config gates are preserved because the classifier already honours them.

**One deliberate divergence:** with nothing restored, warm-boot classifies lexical `InProgress`, which `IndexStages::lifecycle_status` maps to `"walking"`. At registration nothing has started yet, so adopting that verbatim would make every brand-new index report `"walking"` before its first walk. Lexical stays `Pending` (`"created"`) when `chunk_count == 0`, exactly as before.

---

## Tests

Four tests, each **verified failing before the fix and passing after** (proven by temporarily restoring the original handlers and re-running):

| Test | Asserts |
|---|---|
| `tests_4123::delete_index_without_param_preserves_data` | bare `DELETE` removes the registration but leaves the data dir byte-identical |
| `tests_4123::delete_index_with_delete_data_true_destroys_data` | `?delete_data=true` still destroys — capability retained and reachable |
| `tests_4110::create_index_successful_restore_reports_semantic_ready` | successful colocated restore **with** HNSW snapshot → semantic `Ready`, `vector` in capabilities |
| `tests_4110::create_index_restore_without_hnsw_snapshot_stays_pending` | restore **without** snapshot → semantic `Pending`, but lexical `Ready` |

No `#[ignore]`, no cfg-gates.

The #4110 pair is a controlled experiment rather than a literal check: both drive the **real** `create_index_handler` through a create → index-content → unregister → re-register cycle so the second registration performs a genuine colocated restore, and they differ in exactly ONE input — whether the HNSW snapshot was persisted. The same handler must map one to `Ready` and the other to `Pending`. The fourth test's `lexical: Ready` assertion is what keeps it non-vacuous: pre-fix, lexical was hardcoded `Pending`, so it fails too, proving the handler genuinely reads the restore rather than emitting a constant that happens to match in one branch.

The #4123 pair drives the real router via `oneshot`, so route wiring and the query extractor are exercised end-to-end, not just the handler signature.

Closes #4123
Closes #4110

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -59,6 +59,16 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   vanished, are both now reported to the operator with a skip reason rather
   than dropped silently.
 
+- **Session index GC would have silently stopped reclaiming trusty-search disk**
+  ([#4123](https://github.com/bobmatnyc/trusty-tools/issues/4123)):
+  trusty-search's `DELETE /indexes/:id` now preserves on-disk data unless
+  `?delete_data=true` is passed. Both index-GC call sites in `search_gc.rs` —
+  decommission teardown and the periodic orphan sweep — now opt in explicitly.
+  Without this they would have kept removing the registration while leaving the
+  index data behind forever: the workspace each one describes is a disposable
+  worktree that is about to be deleted, so nothing will ever re-register at
+  that root and the preserved data would be unreachable garbage.
+
 ---
 ## [1.2.0] — 2026-07-27
 
--- a/crates/trusty-mpm/src/session_manager/search_gc.rs
+++ b/crates/trusty-mpm/src/session_manager/search_gc.rs
@@ -110,7 +110,12 @@ pub(super) async fn delete_search_index_best_effort(index_id: &str) {
             return;
         }
     };
-    let url = format!("{base}/indexes/{index_id}");
+    // Issue #4123: trusty-search's `DELETE /indexes/:id` now preserves on-disk
+    // data unless `delete_data=true` is passed. Opt in — the workspace this
+    // index describes is a disposable worktree that decommission is about to
+    // delete, so nothing will ever re-register at that root and preserved
+    // index data would be unreachable garbage on disk forever.
+    let url = format!("{base}/indexes/{index_id}?delete_data=true");
     match client.delete(&url).send().await {
         Ok(resp) if resp.status().is_success() => {
             info!(
@@ -322,7 +327,11 @@ impl SessionManager {
 
         let mut removed = Vec::new();
         for id in candidates {
-            let url = format!("{base}/indexes/{id}");
+            // Issue #4123: opt in to data deletion — same rationale as
+            // `delete_search_index_best_effort`. Every candidate here is a
+            // worktree-scoped index with no live session, so its data is
+            // unreachable garbage; a bare DELETE would silently leak it.
+            let url = format!("{base}/indexes/{id}?delete_data=true");
             match client.delete(&url).send().await {
                 Ok(resp) if resp.status().is_success() => {
                     info!(index_id = %id, "search-index-gc: removed orphaned/0-chunk index");
--- a/crates/trusty-search/CHANGELOG.md
+++ b/crates/trusty-search/CHANGELOG.md
@@ -7,6 +7,37 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING (HTTP API): `DELETE /indexes/:id` no longer destroys on-disk data
+  by default** ([#4123](https://github.com/bobmatnyc/trusty-tools/issues/4123)).
+  The handler hardcoded `delete_data=true`, so every `DELETE` destroyed the
+  index's data directory and the HTTP surface offered NO way to merely
+  deregister. Registry hygiene was therefore impossible through the API: an
+  operator clearing 49 stale entries had to stop the daemon and hand-edit
+  `indexes.toml`, because one mis-typed id would have destroyed a real corpus.
+  A bare `DELETE` now deregisters only (the same safe path the orphan-reaper
+  has always used); destroying data requires an explicit
+  `?delete_data=true`. The response gained a `data_deleted` field so callers
+  can confirm which semantics ran. An unparseable `delete_data` value is
+  rejected with `400` rather than guessed at.
+
+  This also makes three long-standing pieces of documentation true rather than
+  false — all of them already promised the preserving behaviour that did not
+  exist: the API reference (`crates/trusty-search/CLAUDE.md`, "On-disk redb data
+  is preserved"), `trusty-search index remove`'s `--help` ("The on-disk redb /
+  HNSW snapshot is preserved — re-registering with the same path reuses it"),
+  and the UI's delete confirmation ("On-disk data is preserved.").
+
+  **Action required for callers that relied on `DELETE` reclaiming disk** —
+  they will silently stop reclaiming it and leave orphaned data behind. The
+  in-tree ones are updated in this change to pass `?delete_data=true`: the
+  `delete_index` MCP tool (its descriptor promises "and all its data"),
+  `trusty-search cleanup`, trusty-mpm's decommission + orphan-sweep index GC,
+  and the benchmark harnesses that require a clean slate. `trusty-search index
+  remove` and the UI are deliberately left on the new preserving default,
+  because that is what they already told users they did.
+
 ### Fixed
 
 - **An index whose durable corpus failed to open kept accepting watcher
@@ -36,6 +67,24 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   unchanged; corpus-failed indexes still return empty results (that is issue
   #4087, not fixed here).
 
+- **`POST /indexes` reported every stage `Pending` even when a colocated corpus
+  restore succeeded** ([#4110](https://github.com/bobmatnyc/trusty-tools/issues/4110)).
+  `create_index_handler` doubles as the "adopt an existing colocated corpus"
+  door — `build_indexer_from_entry` synchronously restores the redb corpus, the
+  HNSW snapshot and the symbol graph — but the handler then set
+  `lexical: pending(), semantic: pending()` unconditionally and threw that
+  outcome away. Since `search_capabilities` is derived from `stages`, a fully
+  intact index came up advertising no vector lane: semantic search hard-errored
+  with "requires Stage 2 (embeddings), which is not yet ready" and `search_all`
+  silently degraded to BM25-only, every hit reporting `match_reason="bm25"` —
+  indistinguishable from a genuinely dead vector lane. Only a daemon restart
+  cleared it, because the warm-boot path already classified correctly from the
+  same signals. The registration path now derives stages with
+  `derive_warm_boot_stages`, the same pure classifier warm-boot and
+  lazy-restore use, over the same signals read the same way, so the two can no
+  longer disagree about an identical on-disk state. A genuinely new index still
+  reports `created` (lexical `Pending`), not warm-boot's `walking`.
+
 ---
 ## [0.40.0] — 2026-07-27
 
@@ -50,7 +99,6 @@ MINOR, not patch: `service::lazy_loader::store`'s public
 - `trusty-common` requirement raised to `^0.27` (was `^0.26`): 0.27.0 makes
   `ChatEvent` `#[non_exhaustive]`, which a `^0.26` requirement cannot express.
   `service::ui`'s `ChatEvent` match gained a wildcard arm accordingly.
-
 
 ### Fixed
 
--- a/crates/trusty-search/CLAUDE.md
+++ b/crates/trusty-search/CLAUDE.md
@@ -154,13 +154,23 @@ Register a new (empty) index. Idempotent: re-registering an existing id returns
   { "id": "my-project", "created": false, "reason": "already exists" }
   ```
 
-##### `DELETE /indexes/:id`
+##### `DELETE /indexes/:id[?delete_data=true]`
 
-Drop an index from the in-memory registry. On-disk redb data is preserved —
-re-registering with the same id will reuse it.
+Drop an index from the in-memory registry and from `indexes.toml` / `roots.toml`.
+
+By default on-disk redb data is **preserved** — re-registering with the same id
+reuses it. This is the safe deregistration verb for registry hygiene. Pass
+`?delete_data=true` to also destroy the index's data directory.
+
+> **Behaviour change (issue #4123).** Before this, `DELETE` ALWAYS destroyed the
+> data directory and the documented "preserved" contract above was not what the
+> handler did. Callers that rely on `DELETE` reclaiming disk must now pass
+> `?delete_data=true` explicitly.
 
 - **Request body**: none.
-- **Response 200**: `{ "id": "my-project", "removed": true }`
+- **Query**: `delete_data` (bool, default `false`). An unparseable value is
+  rejected with `400` rather than silently defaulting either way.
+- **Response 200**: `{ "id": "my-project", "removed": true, "data_deleted": false }`
 
 ##### `GET /indexes/:id/status`
 
--- a/crates/trusty-search/src/commands/cleanup.rs
+++ b/crates/trusty-search/src/commands/cleanup.rs
@@ -140,7 +140,11 @@ pub async fn handle_cleanup(yes: bool, dry_run: bool) -> Result<()> {
     let mut removed = 0usize;
     let mut failed: Vec<(String, String)> = Vec::new();
     for e in &empties {
-        let url = format!("{}/indexes/{}", base, e.id);
+        // Issue #4123: `DELETE` preserves on-disk data unless `delete_data=true`
+        // is passed. Opt in so `cleanup` keeps fully reclaiming the stub — every
+        // candidate here is 0-chunk by construction (step 4 filters on
+        // `chunk_count == 0`), so nothing of value can be destroyed.
+        let url = format!("{}/indexes/{}?delete_data=true", base, e.id);
         match client.delete(&url).send().await {
             Ok(resp) if resp.status().is_success() => {
                 removed += 1;
--- a/crates/trusty-search/src/mcp/tools/index.rs
+++ b/crates/trusty-search/src/mcp/tools/index.rs
@@ -116,7 +116,16 @@ pub(super) async fn dispatch_index_tool(
                 Ok(v) => v,
                 Err(e) => return Some(Err(e)),
             };
-            Some(server.delete(&format!("/indexes/{index_id}")).await)
+            // Issue #4123: `DELETE /indexes/:id` now preserves on-disk data
+            // unless `delete_data=true` is passed. This tool's descriptor
+            // promises "Delete a registered index and all its data", so it
+            // opts in explicitly to keep that contract — without this the
+            // tool would silently stop reclaiming disk.
+            Some(
+                server
+                    .delete(&format!("/indexes/{index_id}?delete_data=true"))
+                    .await,
+            )
         }
         "reindex" => {
             let index_id = match required_index_id(server, args) {
--- a/crates/trusty-search/src/service/server/indexes.rs
+++ b/crates/trusty-search/src/service/server/indexes.rs
@@ -523,34 +523,89 @@ pub(super) async fn create_index_handler(
     // Issue #75: capture the current git HEAD SHA at registration; the search
     // response uses it to flag stale results when the working tree advances.
     let indexed_head_sha = crate::core::git::head_sha(&req.root_path);
-    // Issue #109, Phase 1: pre-mark semantic + graph as `Skipped` for
-    // lexical-only indexes so the search handler never tries the HNSW lane.
-    // Issue #313: pre-mark graph as `Skipped` for skip_kg indexes.
-    // Issue #2984 Phase 1: `skip_vector` pre-marks semantic as `Skipped`,
-    // independently of `skip_kg`'s graph gate — the two compose freely so
-    // every quadrant (both on, KG-only, vector-only, both off via
-    // `lexical_only`) is reachable from create time.
-    let stages = if lexical_only {
-        crate::core::registry::IndexStages {
-            lexical: crate::core::registry::StageState::pending(),
-            semantic: crate::core::registry::StageState::skipped(),
-            graph: crate::core::registry::StageState::skipped(),
-        }
-    } else {
-        crate::core::registry::IndexStages {
-            lexical: crate::core::registry::StageState::pending(),
-            semantic: if skip_vector {
-                crate::core::registry::StageState::skipped()
-            } else {
-                crate::core::registry::StageState::pending()
-            },
-            graph: if skip_kg {
-                crate::core::registry::StageState::skipped()
-            } else {
-                crate::core::registry::StageState::pending()
-            },
-        }
-    };
+    // Issue #4110: derive the stage state from what `build_indexer_from_entry`
+    // ACTUALLY restored, instead of asserting `Pending` and ignoring the
+    // outcome.
+    //
+    // Why: `POST /indexes` doubles as the "adopt an existing colocated corpus"
+    // door — the builder above synchronously restores the redb corpus, the
+    // HNSW snapshot and the symbol graph for `id` (the issue #85
+    // warm-restore-on-create path). Hardcoding `pending()` threw that result
+    // away, so a fully-intact index came up advertising no semantic lane:
+    // `search_capabilities` is computed from `stages`, so semantic search
+    // hard-errored with "requires Stage 2 (embeddings), which is not yet
+    // ready" and `search_all` silently degraded to BM25-only — every hit
+    // reporting `match_reason="bm25"`, indistinguishable from a genuinely dead
+    // vector lane. Only a daemon restart cleared it, because the warm-boot
+    // path (`commands/start_restore.rs`) already classified correctly.
+    // What: reuse `derive_warm_boot_stages` — the same pure classifier
+    // warm-boot and lazy-restore use — over the same signals read the same
+    // way, so registration and warm-boot can no longer disagree about an
+    // identical on-disk state. The `skip_*` / `lexical_only` config gates are
+    // honoured inside the classifier (config intent wins over on-disk state),
+    // so the pre-marking this replaces is preserved. `corpus_open_failed` is
+    // read rather than hardcoded `false` so the classifier's input contract is
+    // satisfied honestly — but note this is safe ONLY because the #2336 check
+    // above already 500s on that flag, so it is always `false` by the time we
+    // get here. It is NOT defensive: if that guard were moved or removed, the
+    // classifier's rule 1 would return lexical `Failed`, and the
+    // `restore_chunk_count == 0` override below would immediately clobber it
+    // back to `Pending` (a failed open reports 0 chunks), reporting a broken
+    // index as a healthy brand-new one. Moving that guard requires teaching
+    // the override to exempt `corpus_open_failed`.
+    // A genuinely brand-new index restores nothing: `chunk_count == 0` and no
+    // snapshot ⇒ semantic/graph `Pending` exactly as before.
+    let restore_chunk_count = indexer
+        .corpus_store()
+        .and_then(|c| c.chunk_count().ok())
+        .unwrap_or(0);
+    // Issue #2922: file existence alone is not proof the snapshot loaded —
+    // `hnsw_load_failed` catches the truncated/corrupt case that would
+    // otherwise report `semantic: ready` over an empty in-memory store.
+    let restore_hnsw_ready = !indexer.hnsw_load_failed
+        && crate::service::persistence::hnsw_path_for_entry(&init_entry)
+            .map(|p| crate::service::persistence::has_persisted_hnsw(&p))
+            .unwrap_or(false);
+    let restore_graph_nodes = indexer.snapshot_symbol_graph().await.node_count();
+    let mut stages = crate::service::warm_boot::derive_warm_boot_stages(
+        crate::service::warm_boot::WarmBootInputs {
+            chunk_count: restore_chunk_count,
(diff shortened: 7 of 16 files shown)


bobmatnyc:
## Adversarial code review — verdict: **APPROVE**

Independent review (code + spec only; implementer reasoning not used as evidence). Zero CRITICAL, zero HIGH at >80% confidence. Two MEDIUM and two LOW findings below are tracked, not blocking.

> Posted as a review **comment**, not `--request-changes`: GitHub rejects a request-changes review from the account that authored the PR.

---

### Central claim verified: this is a restoration, not an unannounced default flip

All three artifacts the PR cites were checked in the tree. **CONFIRMED, all three:**

| Artifact | Location | Pre-PR text |
|---|---|---|
| API reference | `crates/trusty-search/CLAUDE.md` (line removed by this diff) | "On-disk redb data is preserved — re-registering with the same id will reuse it." |
| CLI help | `crates/trusty-search/src/main.rs:1013-1014` | "The on-disk redb / HNSW snapshot is preserved — re-registering with the same path reuses it." |
| UI confirm dialog | `crates/trusty-search/ui/src/lib/views/Indexes.svelte:173` (single), `:395` (bulk) | "On-disk data is preserved." |

Both callers deliberately left on the new preserving default (`index remove`, the UI) are exactly the two whose user-visible text already promised preservation. The default flip is a restoration for them. Severity of the remaining risk stays where the author put it.

---

### 1. Caller enumeration — **CONFIRMED complete. No misses, in either direction.**

Independent workspace sweep (`.delete(`, `method: 'DELETE'`, `curl -X DELETE`, `/indexes/` across Rust/JS/Svelte/TS/shell/docs, excluding `target/`, `ui-dist/`, sibling worktrees). Every HTTP caller of the endpoint:

| Caller | Site | Setting | Correct? |
|---|---|---|---|
| MCP `delete_index` | `crates/trusty-search/src/mcp/tools/index.rs:126` | `true` | Yes — descriptor at `descriptors.rs:238` reads "Delete a registered index and **all its data**" |
| `trusty-search cleanup` | `crates/trusty-search/src/commands/cleanup.rs:147` | `true` | Yes — candidates filtered to `chunk_count == 0` (`cleanup.rs:35,85`) |
| `trusty-search index remove` | `crates/trusty-search/src/commands/index_remove.rs:66` | default `false` | Yes — help text promises preservation |
| UI single + bulk delete | `ui/src/lib/api.js:70`, called from `Indexes.svelte:178` and `:258` | default `false` | Yes — both dialogs promise preservation |
| tm decommission | `crates/trusty-mpm/src/session_manager/search_gc.rs:115` | `true` | Yes |
| tm orphan sweep | `crates/trusty-mpm/src/session_manager/search_gc.rs:334` | `true` | Yes |
| 5 benchmark sites | `tests/benchmark_open_mpm.rs`, `_kg.rs`, `_synthetic.rs` | `true` | Yes |

Two things the sweep turned up that are correctly **not** in the enumeration:

- `crates/trusty-console/src/proxy/routes.rs` reverse-proxies `/api/search/*` and forwards path **and query** verbatim (`build_upstream_url`, test at `:427`). It is a transparent pass-through with no opinion on `delete_data`; no console UI issues an index DELETE (`rg` over `crates/trusty-console/ui/src` — zero hits). Nothing to change.
- `crates/trusty-search/ui-dist/` is a committed built bundle. The UI source needed no change, so no rebuild is owed — this avoids the recurring stale-`ui-dist` trap.

### 2. Split matches per-caller intent — **CONFIRMED**

`search_gc.rs` really does pass `true` at both sites (`:115`, `:334`) — verified in the tree, not from the diff summary. tm does not leak storage. Candidate selection backs the choice: `is_orphan_index` only nominates a worktree-scoped index that is unclaimed **and** (root gone **or** 0 chunks) — `search_gc.rs:474-487` explicitly refuses to nominate a populated, still-on-disk unclaimed worktree.

### 3. `#4110` stage derivation — judgement is sound; no state reports a status the index lacks

- The `hnsw_snapshot_ready` formula is byte-identical to both existing consumers (`commands/start_restore.rs:337`, `service/lazy_restore.rs:123`). No divergence introduced.
- The false-ready hole I went looking for is closed upstream: a **dim mismatch** returns `load_failed = true` (`persistence_loader/mod.rs:354`, the `return Ok((fresh_store(dim)?, true))` after the fall-through), so a snapshot from a different embedder cannot report `semantic: Ready` over an empty store.
- The lexical `InProgress → Pending` correction is right, and its blast radius is contained: `reset_stages_for_reindex` (`reindex/stages.rs:55-90`) **fully overwrites** all three stages at reindex start, so a create-time `Ready` can never leak into a later reindex as a stale-ready.
- `corpus_open_failed` genuinely cannot reach the classifier as `true` — `indexes.rs:380` returns 500 first. See MEDIUM-2 for the caveat.

### 4. Issue drift — **CONFIRMED**

`crates/trusty-search/src/service/server/lifecycle.rs` does not exist anywhere in the tree. The real classifier is `service/warm_boot/stages.rs::derive_warm_boot_stages`, and the PR calls it directly with the same input construction as both existing callers. The issue text was wrong; the PR follows the real one.

### 5. Tests — **independently verified, including the pre-fix failure**

4/4 pass on the PR branch:

```
test service::server::tests_4123::delete_index_without_param_preserves_data ... ok
test service::server::tests_4110::create_index_successful_restore_reports_semantic_ready ... ok
test service::server::tests_4123::delete_index_with_delete_data_true_destroys_data ... ok
test service::server::tests_4110::create_index_restore_without_hnsw_snapshot_stays_pending ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 1367 filtered out
```

I did not take the author's word on the pre-fix failure. In a throwaway worktree I reverted **only** the two handler bodies to their pre-fix form (hardcoded `delete_data=true`; unconditional `Pending` stages), kept the new tests, and re-ran:

```
test result: FAILED. 1 passed; 3 failed; 0 ignored; 0 measured; 1367 filtered out
```

```
tests_4123.rs:111: bare DELETE must NOT destroy on-disk data;
  .../indexes/delete-data-4123/corpus.marker is gone
tests_4110.rs:179: a restored HNSW snapshot must map to semantic Ready, not Pending — got Pending
tests_4110.rs:225: the corpus DID restore, so lexical must be Ready — got Pending
```

Specifically on the point raised: `delete_index_without_param_preserves_data` fails pre-fix at the **marker-file existence assertion** (`tests_4123.rs:111`), not at a status code — and follows it with a byte-equality check (`:115-119`). It asserts data genuinely survives on disk. Confirmed non-vacuous.

Full crate suite: **1369 passed, 1 failed**. The one failure is `service::reindex::defer_embed::tests::deferred_embed_pass_pre_seeds_total_before_embedding` — a 2-second polling budget on a `spawn_deferred_embed_pass` background task. It builds its handle via `IndexHandle::bare` and never touches `create_index_handler`, and it passes in isolation on this same branch (`1 passed`). Load artifact, not this PR.

### 6. Test isolation — **sufficient for this PR**

Verified programmatically rather than by inspection: every `set_var("TRUSTY_DATA_DIR")` site under `crates/trusty-search/src` is inside a `#[serial_test::serial]` test. The only non-test sites are `IsolatedDataDir::new` (`tests_components.rs:63`) and `residency_sweep_tests.rs:46`, both called only from serial tests, plus production `commands/start/daemon.rs:95`. Untagged `#[serial]` shares one global lock, so `tests_4123` cannot have the env var yanked mid-DELETE.

The disclosed pre-existing class cannot corrupt these tests specifically: a non-serial `create_index_handler` test writing `indexes.toml` into `tests_4123`'s sandbox does not touch `<tmp>/indexes/delete-data-4123/corpus.marker` (unique id), and `tests_4110` never sets the env var — it asserts in-memory stages, and the persistence write is warn-only. Not fixing #4213 here is the right call. But see MEDIUM-1, which is new code, not the pre-existing class.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-search/src/service/server/tests_4123.rs` | 92 / 100, 139 / 146 | Bare `set_var`/`remove_var` instead of the RAII `IsolatedDataDir` guard that already exists in this binary. Panicking assertions sit between them (`delete_index`'s status check at `:71`, and `:111`/`:154`). On failure the test unwinds, the `TempDir` is dropped (directory gone) but `TRUSTY_DATA_DIR` stays set to a now-nonexistent path for the remainder of the test binary — cascading one failure into every subsequent serial test. This is exactly the flake-amplification `IsolatedDataDir` (`tests_components.rs:56-73`) exists to prevent. | Replace both `unsafe` blocks with `let _isolated = IsolatedDataDir::new();` (or a local RAII guard), so `Drop` restores the env on the unwind path. |
| MEDIUM | `crates/trusty-search/src/service/server/indexes.rs` | 584-585 (claim at 559-561) | The comment asserts `corpus_open_failed` "is passed rather than assumed, so this stays correct if that guard moves." It would not. With `corpus_open_failed = true` the classifier returns all three stages `Failed`, but `corpus_store()` is absent ⇒ `restore_chunk_count == 0` ⇒ `stages.lexical = pending()` silently clobbers `Failed` → `Pending`. Correct today only because `indexes.rs:380` 500s first — i.e. the exact opposite of what the comment claims. Damage would be bounded (`any_failed()` still true via semantic/graph, so `lifecycle_status` stays `"failed"`), but a comment asserting a robustness property the code lacks is how the next person gets burned. | Either gate the override — `if restore_chunk_count == 0 && !indexer.corpus_open_failed` — or correct the comment to state the override depends on the 500 guard above. |
| LOW | `crates/trusty-search/src/service/server/search.rs` | 53-57 | `serde_urlencoded` bool accepts only `true`/`false`, so `?delete_data=1` and bare `?delete_data` both 400. Fails closed, which is the right direction for a destructive toggle, but `?delete_data=1` is the most natural curl shorthand for the operator-hygiene workflow #4123 exists to serve. | Optional. If accepted as-is, add `1`/`0` to the `Query` note in `CLAUDE.md` so operators hit the docs before the 400. |
| LOW | PR description | — | "Both fail before the fix" is accurate for `tests_4110` and for `delete_index_without_param_preserves_data`, but `delete_index_with_delete_data_true_destroys_data` **passes** pre-fix — correctly, since it guards preserved capability rather than proving a change. Worth stating precisely so a future reader does not treat it as a proof-of-change test. | Reword to "3 of 4 fail pre-fix; the fourth is an anti-regression guard." |

---

## Notes

- **Version skew, assessed and found benign.** A new daemon paired with an *older already-installed* tm would receive a bare `DELETE` from `search_gc.rs` and preserve instead of reclaim. I chased this as a candidate HIGH and it does not hold: every index `create_index_handler` registers is `colocated = true`, so its data lives at `<worktree>/.trusty-search/` — and both GC paths either delete the worktree themselves (`decommission.rs:465` `remove_dir_all`) or only nominate worktrees whose root is already gone / whose chunk count is 0 (`search_gc.rs:474-487`). Nothing meaningful leaks. Legacy `colocated = false` central-store indexes are not worktree-scoped and are not GC candidates. Flagging it only so the release notes can say it was considered.
- **Preserved data is unreachable by every reclamation path** — by design, but worth stating plainly in the operator docs. `unregister_index` with `delete_data = false` still removes the `indexes.toml` entry **and** scrubs `roots.toml` (`search.rs:125-141`), so a preserved index is not rediscovered at warm-boot, not visible to `trusty-search cleanup` (which only lists *registered* indexes), and not reaped by the orphan-reaper. Re-registering requires the operator to know the id **and** the root. That is exactly the #4123 contract, but "preserved" reads as "recoverable" to an operator who did not pick the id.
- **A new but harmless reachable state**: `chunk_count == 0` with an HNSW snapshot present yields `lexical: Pending, semantic: Ready` → `search_capabilities = ["vector"]` with no `"bm25"`. Warm-boot produces the analogous state today (`lexical: InProgress`), so this is parity, and it needs an unusual on-disk shape (`hnsw.usearch` present, `index.redb` empty). Below the 80% bar as a real defect; recorded for completeness.
- **`data_deleted` reporting is honest.** All data operations sit inside `if removed` (`search.rs:125`), so `removed && params.delete_data` cannot over-report. Additive JSON field — no consumer break.
- **Merge gate, for the PM:** `gh pr checks 4218` reports *no checks on the branch* and `reviewDecision: REVIEW_REQUIRED`. This APPROVE covers code review only. Do not merge until CI has actually run and is green — pending is not green.
- Both `CHANGELOG.md` files carry `## [Unreleased]` entries, and the trusty-search one is correctly labelled **BREAKING (HTTP API)**. The changelog merge gate is satisfied.

bobmatnyc:
## Re-review at `0ad2144d` — **VERDICT: APPROVE**

The APPROVE posted at 05:17 UTC was **stale**: it reviewed `49c59c74`, and the branch was
force-pushed twice afterwards (`337e450c` @ 05:21, rebase → `04a2be7b` @ 05:24, amend →
`0ad2144d` @ 05:51). This re-review covers the delta plus an independent re-verification of the
central claims. **0 CRITICAL, 0 HIGH.** One new MEDIUM, one LOW.

---

### Delta since the reviewed commit — `49c59c74 → 0ad2144d`

Computed as the difference between the PR's own patch at each point (each diffed against its own
merge-base, so rebase noise is excluded). The whole behavioural delta is three items:

**1. MEDIUM-1 (bare `set_var`) — genuinely addressed.** ✅ Verified, not taken on trust.

`tests_4123.rs:97` and `:137` now bind the RAII guard:

```rust
let isolated = super::tests_components::IsolatedDataDir::new();
let (router, index_data_dir) = router_with_index_and_data(isolated.path());
```

Both `unsafe { set_var }` / `unsafe { remove_var }` pairs are gone. `IsolatedDataDir` was promoted
from private to `pub(super)` and gained a `path()` accessor (`tests_components.rs:59-75`), and its
`Drop` (`tests_components.rs:77-81`) removes `TRUSTY_DATA_DIR` — so an unwind through the
panicking assertions at `tests_4123.rs:111`/`:147` restores the env var instead of leaving it
pointed at a dropped `TempDir`. The cascade risk this repo is currently exposed to under high
machine load is closed. Note the binding is `isolated`, not `let _ = …`, so the guard lives to end
of scope rather than dropping immediately — correct.

**2. MEDIUM-2 (comment asserting a robustness property the code lacks) — addressed** via the second
of the two options offered (correct the comment rather than gate the override). `indexes.rs:546-555`
now reads, in part:

> `corpus_open_failed` is read rather than hardcoded `false` … but note this is safe ONLY because
> the #2336 check above already 500s on that flag … It is NOT defensive: if that guard were moved
> or removed, the classifier's rule 1 would return lexical `Failed`, and the
> `restore_chunk_count == 0` override below would immediately clobber it back to `Pending`

I verified the corrected claim against the code rather than accepting it: `stages.rs:96-105`
returns all three stages `failed(reason)` under rule 1, and `indexes.rs:591-593` unconditionally
overwrites `stages.lexical = pending()` when `restore_chunk_count == 0` (which a failed open
produces, via `corpus_store()` → `None` → `unwrap_or(0)` at `:558-561`). The comment is now
accurate, including the closing instruction that moving the guard requires teaching the override to
exempt the flag.

**3. New in the delta, not previously reviewed:** the test helper was renamed
`delete_index` → `send_delete_request` (`tests_4123.rs:60-65`) specifically because `delete_index`
is an allowlisted *dangling* doc-pointer at `mcp/tools/http.rs:177` (`Test: delete_index arm
exercises this path`), and a same-named helper would have silently resolved it — converting a known
gap into a falsely-satisfied one. This is a correct call and introduces no new risk. It is the
inverse of the "green-by-disabling" pattern: it preserves a lint's signal rather than muting it.

Nothing else changed. No new logic, no new control flow, no new public surface.

---

### Independent re-verification at the current head

**Restoration claim (#4123) — CONFIRMED, and confirmed to be a restoration.** The decisive check is
that the PR does **not** touch the two artifacts that make it a restoration:

```
$ git diff --name-only $(git merge-base 0ad2144d origin/main) 0ad2144d | grep -E 'ui/|main.rs'
(no output)
```

So both pre-date the PR:
- `crates/trusty-search/ui/src/lib/views/Indexes.svelte:173` — ``Delete index "${id}"? On-disk data is preserved.`` and `:395` for the bulk path.
- `crates/trusty-search/src/main.rs:1013-1014` — "The on-disk redb / HNSW snapshot is preserved — re-registering with the same path reuses it."

`crates/trusty-search/CLAUDE.md:157-163` is updated by the PR to document the new parameter. The
default flip makes two standing user-facing promises true; it is not an unannounced break.

**Caller enumeration — re-swept from scratch, CONFIRMED complete.** Every `?delete_data=true` opt-in
in the tree:

| Caller | Site | Justified by |
|---|---|---|
| MCP `delete_index` | `mcp/tools/index.rs:126` | descriptor at `mcp/tools/descriptors.rs:238` — "Delete a registered index and all its data" |
| `trusty-search cleanup` | `commands/cleanup.rs:147` | reclamation intent — but see MEDIUM below |
| tm decommission | `trusty-mpm/src/session_manager/search_gc.rs:118` | worktree about to be deleted |
| tm orphan sweep | `trusty-mpm/src/session_manager/search_gc.rs:334` | worktree-scoped + unclaimed |
| benchmarks ×5 | `benchmark_open_mpm.rs:293,433`; `benchmark_synthetic.rs:211,310`; `benchmark_open_mpm_kg.rs:291` | clean-slate requirement |

Left on the new preserving default: `commands/index_remove.rs:65` and `ui/src/lib/api.js:70` —
exactly the two whose user-facing text promised preservation. Matches the PR's table with no
misses in either direction.

**Destructive-default trace — no path deletes data an operator did not ask to delete.**
`delete_data` reaches `remove_index_data_dir` through exactly one branch, `search.rs:128`, and is
sourced solely from `Query<DeleteIndexParams>` (`search.rs:52-58`, `#[serde(default)]` ⇒ `false`).
Absent parameter ⇒ no deletion. Unparseable (`?delete_data=1`, `?delete_data=maybe`, bare
`?delete_data`) ⇒ the extractor rejects with `400` before the handler body runs, so a destructive
toggle is never guessed. There is no path on which the *default* destroys anything.

**Tests ran and passed at this exact SHA** — verified against the CI log, not asserted:

```
test service::server::tests_4110::create_index_restore_without_hnsw_snapshot_stays_pending ... ok
test service::server::tests_4110::create_index_successful_restore_reports_semantic_ready ... ok
test service::server::tests_4123::delete_index_with_delete_data_true_destroys_data ... ok
test service::server::tests_4123::delete_index_without_param_preserves_data ... ok
```

No `#[ignore]`, no silent skips.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-search/src/commands/cleanup.rs` | 143-146 (claim); 71-76 + 84-87 (mechanism) | The new comment justifying the `?delete_data=true` opt-in states "every candidate here is 0-chunk **by construction** … so nothing of value can be destroyed." That is not true on the failure branch. A status probe that errors or times out yields `_ => serde_json::json!({})` (`:75`), and `chunk_count` then falls through `.unwrap_or(0)` (`:87`) to `0`, so `chunks == 0` at `:88` nominates the index for deletion. A fully-populated corpus whose `/status` probe merely timed out becomes a `?delete_data=true` target. This is realistic, not theoretical: `daemon_http_client` has a 5s total timeout (`trusty-common/src/server/mod.rs:126`) and `cleanup` fires an unbounded concurrent `JoinSet` of probes across *every* registered index (`:65-79`) — the 49-index scenario #4123 itself cites. Unlike the structurally identical fail-open at `search_gc.rs:235-243`, which is safe because `is_orphan_index` independently requires worktree-scoped **and** unclaimed (`search_gc.rs:201-212`) and documents exactly that scoping, `cleanup` has **no** scope guard at all — persistent project indexes are in range. `--yes` removes the prompt, and the displayed table shows ids, not chunk counts, so an operator cannot spot the impostor. **Not a regression** — pre-PR `DELETE` destroyed data unconditionally, so behaviour is unchanged — but this PR affirmatively re-authorises the destruction on a stated safety property the code does not have. Same defect class the prior round flagged as MEDIUM for `indexes.rs`, and it deserves the same treatment. | Either make the claim true — treat a failed probe as "unknown, skip" rather than `0` (`Option<u64>` from `fetch`, and only nominate on a confirmed `Some(0)`) — or correct the comment to say the opt-in is safe only for candidates whose 0-chunk status was actually observed, and file the fail-open separately. |
| LOW | `crates/trusty-search/src/service/server/search.rs` | 82 | `"data_deleted": removed && params.delete_data` reports `true` even when `remove_index_data_dir` failed, since that failure is warn-only (`:129-131`). The response over-reports destruction. Direction is conservative (an operator is told data is gone when it may survive, never the reverse), and it matches the pre-existing warn-only convention in this function. | Optional. Thread the result: `removed && params.delete_data && data_removal_ok`. |

The prior round's two LOWs (`?delete_data=1` returning 400; the PR description's "both fail before the fix" phrasing) are unchanged and still stand as written. Neither is blocking.

---

## Notes

- **Fail-open sweep result.** I went looking for the repo's recurring failure-branch-advances-state shape across all six opt-in sites. Two instances exist, both pre-existing: `search_gc.rs:235-243` (safe — independently scoped by `is_orphan_index`, and the scoping is documented at the call site) and `cleanup.rs:71-76` (unscoped — the MEDIUM above). The handler itself is clean: the `delete_data == false` path has no failure branch that can reach `remove_index_data_dir`.
- **`#4110` classifier reuse re-checked cross-crate, not just within the module.** `derive_warm_boot_stages` (`service/warm_boot/stages.rs:74`) is now driven by three callers over identically-constructed inputs. The one deliberate divergence — lexical held at `Pending` instead of warm-boot's `InProgress` when `restore_chunk_count == 0` (`indexes.rs:591-593`) — is correct: `InProgress` maps to `"walking"`, which would make every brand-new index claim a walk it never ran. Semantic/graph are not overridden, so a failed or absent restore still reports `Pending`, never a false `Ready`.
- **Changelog merge gate satisfied.** `crates/trusty-search/CHANGELOG.md` carries the entry under `## [Unreleased] → ### Changed`, correctly labelled **BREAKING (HTTP API)** with an explicit "Action required for callers" paragraph; `crates/trusty-mpm/CHANGELOG.md` carries the paired `search_gc.rs` entry. The rebase re-anchored both under a fresh `[Unreleased]` above the newly-released `0.40.0` / `1.2.0` sections without losing content.
- **CI is now green at this head** — `head_sha 0ad2144d`, all 25 checks pass, `Test` succeeded in 18m29s (completed 06:10:09Z). The caveat attached to the 05:17 APPROVE ("do not merge until CI has actually run and is green") is **discharged**. The one previously-noted local `defer_embed` failure did not reproduce in CI.
- The MEDIUM above is tracked, not blocking. Recommend it be folded into the surfacing PR if trivial, or dropped rather than filed — per this project's standing preference against filing every review finding into the backlog.

Issue #2847, opened 2026-07-16
[trusty-search] Legacy index warm-boot silently serves 0 chunks when colocated shadow path is unpopulated (no fallback to configured corpus)


bob-duetto:
**Component:** trusty-search

**Problem**
A legacy/consolidated index (e.g. `main`, defined in `indexes.toml` with a real `root_path` and its own durable corpus) is resolved at warm-boot **only** via a colocated "shadow" path (observed: `~/.local/share/trusty-search/indexes/main`, reached through a `<repos>/.trusty-search` symlink). If that shadow path is not populated, warm-boot logs:
```
WARN cannot resolve hnsw path for 'main': create colocated storage dir at <repos>/.trusty-search
INFO warm-boot: index 'main' restored (colocated=true) — chunks=0 ... corpus_open_failed=false
```
and then **reports the index as successfully restored** (`corpus_open_failed=false`) and the daemon reports `status: ok`, while `main` actually has **0 chunks and returns empty results for every query**. There is no fallback to the index's own configured corpus (`root_path` / `TRUSTY_DATA_DIR`), even though a complete, valid corpus exists there.

**Observed impact (production, 2026-07-16)**
On one serving instance, `main`'s shadow path was missing; the instance's real corpus (6.5GB, current) sat unused in `TRUSTY_DATA_DIR`, and `main` served 0 results for hours — a full user-facing search outage — while `/health` reported `status: ok`. A sibling instance whose shadow path IS externally maintained (kept fresh by a separate sync process) serves the same index correctly. Reproduced identically across trusty-search 0.26.1, 0.29.1, and 0.32.4, and independent of the `colocated` TOML flag and of the `TRUSTY_NO_BOOT_RECONCILE` / `TRUSTY_DISABLE_WATCHER` env toggles — so it is the resolution/restore logic, not config.

**Requested changes**
1. **Fallback to the configured corpus:** when a legacy index's colocated shadow path is missing/empty but its configured `root_path`/data-dir corpus exists and is valid, warm-boot should load from there instead of silently restoring an empty index.
2. **Fail loudly, not silently:** an index that "restores" with `chunks=0` when a non-empty corpus is expected should NOT report `corpus_open_failed=false` / contribute to `status: ok`. Surface it as degraded/failed in `/health` so operators aren't told the daemon is healthy while a core index serves nothing.
[...]

PR #4220, opened 2026-07-28
fix(trusty-search): quarantine incremental writes when the corpus open failed (#4122)

### LINKED ISSUE #4122: trusty-search: broken index handle keeps accepting incremental watcher writes, overwriting corpus
## Observed Behavior

Index `duettoresearch-hackathon-tm-hackathon-01` failed to open its corpus during a daemon warm boot:
- `corpus_open_failed: true`
- All stages `Failed`
- `chunk_count: 0`

However, while in this broken state, its file watcher (`watcher.active: true`; incremental indexing on save, issue #1621) continued accepting writes. Observed `chunk_count` climbing:
- 0 → 68 → 1,334 across successive checks
- From ordinary file saves in the worktree by an unrelated session

After a subsequent daemon restart, the index came back healthy:
- `corpus_open_failed: false`
- All stages Ready
- `chunks=1334`

**But it holds a DIFFERENT, partial corpus.** The pre-failure content is presumed overwritten and unrecoverable.

## Contrast

`cto-duetto` failed identically in the same boot but had **NO watcher activity**. It was fully recovered at its exact original 200,090 chunks by a clean restart. The difference between full recovery and silent data loss was purely whether anything wrote to the tree in the interim.

## Impact

A failed corpus open is a transient, recoverable condition (see the timeout in #3659) — but only if nothing writes over it. Incremental indexing against a handle whose corpus never opened turns a recoverable outage into permanent data loss, with no error surfaced to anyone.

## Suggested Fix Direction

When `corpus_open_failed` is true, the watcher/incremental-index path for that index should refuse writes and log, rather than building a fresh corpus over the unopened one. Treat a failed handle as read-only-and-quarantined until a successful open.

Related: #3659, #1621

bobmatnyc:
## What broke

An index whose durable redb corpus **failed to open** at load time was left fully live — watcher included. `service::persistence_loader` set `corpus_open_failed`, wired **no** `CorpusStore`, and left the in-memory chunk map empty, but nothing stopped incremental writes. Ordinary, unrelated file saves in the worktree kept flowing in, so a **fresh, PARTIAL corpus was built and persisted over the never-opened original**, turning a recoverable outage into permanent data loss.

The incident (#4122): `duettoresearch-hackathon-tm-hackathon-01` climbed `chunk_count` `0 → 68 → 1334` from unrelated saves and came back "healthy" on the next restart holding the *wrong* corpus. The sibling `cto-duetto` hit the **identical** open failure in the same boot, took no watcher writes, and recovered all **200,090** chunks cleanly. Watcher-writes-during-failure was the sole difference between full recovery and unrecoverable loss.

## The fix

A `corpus_open_failed` index is now **write-quarantined**.

| Path | Behaviour |
|---|---|
| `CodeIndexer::index_file` | Refuses with `Err`. This is the single choke point for **every** incremental write: the file watcher (`service::watch_loop`), the boot-time catch-up reconciler (`service::reconcile`), and `POST /indexes/{id}/index-file` (`service::server::files`). |
| `watch_loop::handle_modified` | Same guard, applied earlier, so a quarantined index does not read and chunk every saved file just to discard the result. Placed **after** the exclusion filters, so excluded saves are not reported as refused writes. |
| `index_files_batch*` (full reindex) | **Not gated — and not a real recovery path.** A reindex does not clear the quarantine and does not persist anything: reindex staging is skipped when there is no `CorpusStore` wired, so the whole rebuilt corpus evaporates at the next restart. It is also not an explicit operator action — boot reconcile auto-fires one from `service::reconcile` (`trigger_full_reindex` calls at reconcile.rs:367, :379, :458) whenever the delta is unknown or too large. The only way out of quarantine is **restarting the daemon**, which re-runs `persistence_loader::build_indexer_from_entry` and, on a successful corpus open, calls `set_corpus_store` → `clear_corpus_open_failure()`. |
| `handle_removed` / `remove_chunk` | **Not gated**, and provably a no-op: with `handle_modified` guarded, `indexed_files` never gets populated, and no durable corpus is wired, so removals can only touch an empty in-memory map. |

New module `core/indexer/quarantine.rs` holds the whole surface: `is_write_quarantined()`, `refused_incremental_writes()`, `refuse_incremental_write()`, `clear_corpus_open_failure()`.

### Err, not silent `Ok(())`

The issue says "log and no-op". The refusal is a no-op on state, but it returns `Err` rather than `Ok(())` so `POST /indexes/{id}/index-file` reports **500** instead of falsely answering `"indexed": true`. Answering 200 to a refused write would be the same "report a state you do not have" failure this bug is made of. The watcher and reconciler already tolerate an `Err` from `index_file` (they log and continue/count), so nothing else changes.

## Exactly what counts as "open failed"

The quarantine predicate is `CodeIndexer::corpus_open_failed`, and **nothing else**. It has exactly two producers, both in `service::persistence_loader::build_indexer_from_entry`:

1. **`open_corpus_with_retry(&redb_path)` returned `Err`** — covers redb corruption, incompatible page format, permission denial, `DatabaseAlreadyOpen` surviving the one 50 ms retry (stale file lock from a rapid restart, or a shared-root double-open), and any plain I/O error.
2. **`persistence::corpus_redb_path_for_entry(entry)` returned `Err`** — the storage path could not even be resolved (#2847: a colocated shadow dir that cannot be created, a broken symlink, permission denied).

I deliberately did **not** try to separate "transient I/O" from "permission" from "genuinely corrupt". At the point of failure they are indistinguishable, and the two possible mistakes are wildly asymmetric:

* **Over-refuse** → an un-indexed file save, which the next reindex picks up anyway. Recoverable.
* **Under-refuse** → the corpus is overwritten. Not recoverable — that is this ticket.

So every corpus-open failure quarantines. Note the asymmetry is the *opposite* of the sibling PR in this repo where over-refusal was itself the defect; that reasoning does not transfer here, and I have not let it soften the default.

What does **not** quarantine: an index with no corpus store wired but `corpus_open_failed == false` — a legitimately empty or BM25-only index. That distinction already exists and is pinned by `build_indexer_from_entry_does_not_flag_legitimately_empty_index_as_failed` (#2847).

## How a quarantined index becomes writable again

`CodeIndexer::set_corpus_store` — the single sink every successful `CorpusStore::open` funnels through — now calls `clear_corpus_open_failure()`, which clears the flag, resets the refusal counter, and logs at WARN (the daemon's default stderr filter is `warn`, so an INFO recovery notice would not print at all). There is no separate "unquarantine" lever anyone can forget to pull: **a successful open *is* the recovery event**.

In practice that means a daemon restart, or the loader retrying, or any code path that opens the corpus and wires it. `swap_corpus_store` deliberately does not clear — it replaces an *already-wired* corpus (the reindex staging swap), which by construction cannot happen on a quarantined index (`has_corpus_store()` is `false` there).

This is the anti-over-refusal boundary and it is directly tested: a quarantine that never lifts would be a new outage, not a fix, and would have broken the `cto-duetto` path.

## Observability — why ERROR, verified for this crate

A silent no-op would reproduce the original sin of this bug class. Verified for **this** crate's logging setup, not assumed:

* `commands/start/daemon.rs` installs the capture layer via `trusty_common::init_tracing_with_buffer_and_capture`.
* `trusty_common::error_capture::layer::BugCaptureLayer::on_event` starts with `if *meta.level() != tracing::Level::ERROR { return; }`.

So `warn!` genuinely does **not** reach `errors.jsonl` / `list_recent_errors` / `tm doctor` here. The refusal logs at **ERROR**, with `index_id`, `op`, `target`, and a running `refused_writes` count, plus the recovery instructions. To avoid flooding `errors.jsonl` when a busy worktree fires thousands of events, ERROR is emitted on the **1st** refusal and every **100th** thereafter (DEBUG in between); the running total rides on every ERROR record, so no refusal is unaccounted for. `refused_incremental_writes()` exposes the count programmatically.

## Relationship to #4087

**Not fixed here, and reads are not gated.** A quarantined index still answers searches from its (empty) in-memory corpus, i.e. it still returns a silent empty `200` — that is #4087. The only interaction: because the corpus can no longer grow from watcher writes, a corpus-failed index now *reliably* serves 0 results instead of partial garbage, which makes #4087 more consistently reproducible. It does not fix or regress it.

## Tests

Three tests in `crates/trusty-search/tests/corpus_open_quarantine_4122.rs`. No `#[ignore]`, no cfg-gates. Both failure modes are induced through **production code paths** rather than by poking `corpus_open_failed` directly.

1. **`quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero`** — the corruption sequence, end to end through the real `spawn_watch_loop` + OS watcher. Boots two watch loops: one index sabotaged into `corpus_open_failed` (a directory where `index.redb` belongs), one healthy as the **positive control** — waiting for *its* `chunk_count` to grow proves the watcher pipeline, debounce window, and OS notifications work in this environment, so a flat count on the quarantined index is a refusal and not a stalled watcher. Waits are condition-based, not sleeps.
2. **`successful_reopen_lifts_quarantine_and_leaves_corpus_intact`** — the `cto-duetto` recovery case, and the anti-over-refusal guard. Seeds a real redb corpus and holds the handle open so the second open fails with `DatabaseAlreadyOpen` — a transient failure that leaves the file perfectly readable, exactly as in the incident. Asserts the write is refused and the on-disk corpus is untouched; then releases the lock, reopens, wires the store, and asserts the quarantine lifts, the counter resets, writes flow again, **and every original chunk is still present**.
3. **`quarantine_refusal_emits_error_level_diagnostic`** — asserts capture through the **real** `BugCaptureLayer` + `ErrorStore`, not merely that some log line was emitted. The indexer is built *before* the capture layer is installed so the only record under test is the refusal itself.

### Failing before, passing after

Verified by temporarily neutralising both guards (`refuse_incremental_write` → always `false`, `clear_corpus_open_failure` → no-op) while keeping the API, so the pre-fix behaviour is reproduced exactly:

```
running 3 tests
test quarantine_refusal_emits_error_level_diagnostic ... FAILED
test successful_reopen_lifts_quarantine_and_leaves_corpus_intact ... FAILED
test quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero ... FAILED

---- quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero stdout ----
assertion `left == right` failed: #4122: a write-quarantined index must not grow a chunk
corpus from watcher writes — this is the 0 → 68 → 1334 climb that destroyed
duettoresearch-hackathon-tm-hackathon-01
  left: 1
 right: 0

---- successful_reopen_lifts_quarantine_and_leaves_corpus_intact stdout ----
a quarantined index must refuse an incremental write: ()

---- quarantine_refusal_emits_error_level_diagnostic stdout ----
the write must be refused: ()

test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out
```

After the fix:

```
running 3 tests
test quarantine_refusal_emits_error_level_diagnostic ... ok
test successful_reopen_lifts_quarantine_and_leaves_corpus_intact ... ok
test quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.11s
```

## Gate

`cargo test -p trusty-search --no-fail-fast` (all targets, **not** `--lib`): **`CARGO_EXIT=0`**, headline `test result: ok. 1366 passed; 0 failed; 1 ignored`, plus `319 passed` and every integration target green.

`cargo fmt --check`: clean. `scripts/check_line_cap.sh`: `7 allowlisted, 0 violations — OK.`

**Clippy substitution, disclosed:** a bare `cargo clippy --workspace --all-targets` hits the pre-existing Tauri `frontendDist` panic on `trusty-mpm-gui` / `trusty-code-gui`, so I ran **CI's own exclusion list, verbatim from `ci.yml` line 164** — no wider:

```
SKIP_UI_BUILD=1 cargo clippy --workspace --all-targets \
  --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
```

Result: clean (one real finding, `manual_is_multiple_of` in the new module, fixed).

**Two pre-existing failures in `trusty-agents`, not mine** — both proven by re-running with my changes **stashed** on clean `origin/main` (`5bf231b9`):

* `tests/api_e2e` (4 tests, "api server did not become healthy within 5s") — reproduces identically at baseline: `test result: FAILED. 0 passed; 4 failed` / `BASELINE_API_E2E_EXIT=101`.
* `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json` — a transient uv/Python-interpreter probe failure under heavy machine load; passes on re-run (`8 passed; 0 failed`).

Neither touches `trusty-search`.

Closes #4122

### DIFF
--- a/crates/trusty-search/CHANGELOG.md
+++ b/crates/trusty-search/CHANGELOG.md
@@ -25,6 +25,28 @@ MINOR, not patch: `service::lazy_loader::store`'s public
 
 ### Fixed
 
+- **An index whose durable corpus failed to open kept accepting watcher
+  writes, permanently destroying the corpus (issue #4122, P0 data loss).**
+  When `CorpusStore::open` failed at load time the loader set
+  `corpus_open_failed` but left the handle fully live, watcher included, so
+  ordinary unrelated file saves rebuilt a fresh PARTIAL corpus over the
+  never-opened original — in production `chunk_count` climbed `0 → 68 → 1334`
+  and the index came back "healthy" on the next restart holding the wrong
+  content. A `corpus_open_failed` index is now **write-quarantined**: every
+  incremental write path (`service::watch_loop`, `service::reconcile`, and
+  `POST /indexes/{id}/index-file`) is refused at the shared
+  `CodeIndexer::index_file` choke point, and the watcher additionally bails
+  before it reads and chunks the saved file. Refusals are emitted at ERROR
+  (the only level `trusty_common::error_capture` persists to `errors.jsonl` /
+  `list_recent_errors` / `tm doctor`), throttled to the 1st and every 100th,
+  and counted on `CodeIndexer::refused_incremental_writes`. The quarantine is
+  **not permanent**: a successful `set_corpus_store` lifts it and resets the
+  counter, so the clean-restart recovery path (the incident's `cto-duetto`,
+  restored at its full 200,090 chunks) is unaffected. Bulk reindex is
+  deliberately not gated — it is the documented way out of this state. Reads
+  are unchanged; corpus-failed indexes still return empty results (that is
+  issue #4087, not fixed here).
+
 - **Two more index-registration paths had no `root_path` collision guard —
   fourth occurrence of the #2305/#2336 `DatabaseAlreadyOpen` class (issue
   #3993).** An audit prompted by #3929 found `find_root_path_collision`
--- a/crates/trusty-search/CHANGELOG.md
+++ b/crates/trusty-search/CHANGELOG.md
@@ -7,6 +7,35 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Fixed
+
+- **An index whose durable corpus failed to open kept accepting watcher
+  writes, permanently destroying the corpus (issue #4122, P0 data loss).**
+  When `CorpusStore::open` failed at load time the loader set
+  `corpus_open_failed` but left the handle fully live, watcher included, so
+  ordinary unrelated file saves rebuilt a fresh PARTIAL corpus over the
+  never-opened original — in production `chunk_count` climbed `0 → 68 → 1334`
+  and the index came back "healthy" on the next restart holding the wrong
+  content. A `corpus_open_failed` index is now **write-quarantined**: every
+  incremental write path (`service::watch_loop`, `service::reconcile`, and
+  `POST /indexes/{id}/index-file`) is refused at the shared
+  `CodeIndexer::index_file` choke point, and the watcher additionally bails
+  before it reads and chunks the saved file. Refusals are emitted at ERROR
+  (the only level `trusty_common::error_capture` persists to `errors.jsonl` /
+  `list_recent_errors` / `tm doctor`), throttled to the 1st and every 100th,
+  and counted on `CodeIndexer::refused_incremental_writes`. **Recovery is a
+  daemon restart** — only a successful `CorpusStore::open` lifts the
+  quarantine and it is attempted solely at load time, so the ERROR text says
+  so explicitly and warns that a reindex neither clears the state nor
+  persists anything (with no corpus wired it skips staging entirely). The
+  clean-restart path (the incident's `cto-duetto`, restored at its full
+  200,090 chunks) is unaffected. Bulk reindex is deliberately not gated,
+  which is safe because a quarantined index holds no `CorpusStore` — an
+  invariant now documented and pinned by `debug_assert!`s, since boot
+  reconcile auto-fires reindexes with no quarantine check. Reads are
+  unchanged; corpus-failed indexes still return empty results (that is issue
+  #4087, not fixed here).
+
 ---
 ## [0.40.0] — 2026-07-27
 
@@ -24,28 +53,6 @@ MINOR, not patch: `service::lazy_loader::store`'s public
 
 
 ### Fixed
-
-- **An index whose durable corpus failed to open kept accepting watcher
-  writes, permanently destroying the corpus (issue #4122, P0 data loss).**
-  When `CorpusStore::open` failed at load time the loader set
-  `corpus_open_failed` but left the handle fully live, watcher included, so
-  ordinary unrelated file saves rebuilt a fresh PARTIAL corpus over the
-  never-opened original — in production `chunk_count` climbed `0 → 68 → 1334`
-  and the index came back "healthy" on the next restart holding the wrong
-  content. A `corpus_open_failed` index is now **write-quarantined**: every
-  incremental write path (`service::watch_loop`, `service::reconcile`, and
-  `POST /indexes/{id}/index-file`) is refused at the shared
-  `CodeIndexer::index_file` choke point, and the watcher additionally bails
-  before it reads and chunks the saved file. Refusals are emitted at ERROR
-  (the only level `trusty_common::error_capture` persists to `errors.jsonl` /
-  `list_recent_errors` / `tm doctor`), throttled to the 1st and every 100th,
-  and counted on `CodeIndexer::refused_incremental_writes`. The quarantine is
-  **not permanent**: a successful `set_corpus_store` lifts it and resets the
-  counter, so the clean-restart recovery path (the incident's `cto-duetto`,
-  restored at its full 200,090 chunks) is unaffected. Bulk reindex is
-  deliberately not gated — it is the documented way out of this state. Reads
-  are unchanged; corpus-failed indexes still return empty results (that is
-  issue #4087, not fixed here).
 
 - **Two more index-registration paths had no `root_path` collision guard —
   fourth occurrence of the #2305/#2336 `DatabaseAlreadyOpen` class (issue
--- a/crates/trusty-search/src/core/indexer/ingest/commit.rs
+++ b/crates/trusty-search/src/core/indexer/ingest/commit.rs
@@ -332,6 +332,14 @@ impl CodeIndexer {
     /// What: clones the chunks plus entities, moves them onto a blocking worker,
     /// and writes both tables in one atomic transaction. Failures are logged at
     /// `warn` and swallowed.
+    ///
+    /// Issue #4122: this is the single durable redb write, and the `None` arm
+    /// below is what makes the UNGATED bulk-reindex path safe on a
+    /// write-quarantined index — NOT any assumption that reindexes are
+    /// operator-initiated (boot reconcile auto-fires them, see
+    /// `core::indexer::quarantine`'s module docs). The `debug_assert!` pins
+    /// that invariant so an in-process corpus reopen cannot silently
+    /// reintroduce the data-loss hole.
     /// Test: `tests::test_corpus_store_roundtrip`.
     pub(crate) async fn commit_corpus_to_redb(
         &self,
@@ -341,6 +349,14 @@ impl CodeIndexer {
         let Some(corpus) = self.corpus.clone() else {
             return;
         };
+        debug_assert!(
+            !self.corpus_open_failed,
+            "index '{}': INVARIANT VIOLATED — durable corpus write attempted while \
+             the index is write-quarantined (corpus_open_failed). A quarantined \
+             index must hold no CorpusStore; see core::indexer::quarantine (issue \
+             #4122).",
+            self.index_id
+        );
         let chunks = chunks.to_vec();
         let entities = entities_by_file.to_vec();
         let index_id = self.index_id.clone();
--- a/crates/trusty-search/src/core/indexer/ingest/mod.rs
+++ b/crates/trusty-search/src/core/indexer/ingest/mod.rs
@@ -228,8 +228,33 @@ impl CodeIndexer {
     /// What: chunk the file, batch-embed all chunks, commit vectors / BM25 /
     /// corpus, then enrich entities via the NLP helper and rebuild the
     /// symbol graph once.
-    /// Test: covered by every `index_file`-based test in `indexer::tests`.
+    ///
+    /// Issue #4122: this is the single choke point every INCREMENTAL write
+    /// funnels through — the file watcher (`service::watch_loop`), the
+    /// boot-time catch-up reconciler (`service::reconcile`), and the explicit
+    /// `POST /indexes/{id}/index-file` endpoint (`service::server::files`).
+    /// When the index is write-quarantined because its durable corpus failed
+    /// to open, every one of them must be refused: accepting them rebuilds a
+    /// fresh PARTIAL corpus over the never-opened original and destroys it.
+    /// The refusal is returned as an `Err` rather than a silent `Ok(())` so
+    /// the HTTP endpoint reports `500` instead of falsely answering
+    /// `"indexed": true` — the ERROR-level diagnostic is emitted by
+    /// `refuse_incremental_write` either way. The BULK path
+    /// (`index_files_batch*`, used by a full reindex) is intentionally NOT
+    /// gated: a reindex is an explicit operator action that rebuilds the
+    /// corpus wholesale and is the documented way out of this state.
+    /// Test: covered by every `index_file`-based test in `indexer::tests`;
+    /// the quarantine branch by
+    /// `quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero`.
     pub async fn index_file(&self, file_path: &str, content: &str) -> Result<()> {
+        if self.refuse_incremental_write("index_file", file_path) {
+            anyhow::bail!(
+                "index '{}' is write-quarantined: its durable corpus failed to open, so \
+                 incremental indexing of '{file_path}' was refused to avoid overwriting \
+                 the on-disk corpus (issue #4122)",
+                self.index_id
+            );
+        }
         let (mut chunks, entities) = chunk_ast(file_path, content);
 
         populate_virtual_terms(&mut chunks, &entities);
--- a/crates/trusty-search/src/core/indexer/ingest/mod.rs
+++ b/crates/trusty-search/src/core/indexer/ingest/mod.rs
@@ -241,8 +241,13 @@ impl CodeIndexer {
     /// `"indexed": true` — the ERROR-level diagnostic is emitted by
     /// `refuse_incremental_write` either way. The BULK path
     /// (`index_files_batch*`, used by a full reindex) is intentionally NOT
-    /// gated: a reindex is an explicit operator action that rebuilds the
-    /// corpus wholesale and is the documented way out of this state.
+    /// gated. Note this is NOT because reindexes are operator-initiated —
+    /// `service::reconcile` auto-fires them at boot with no
+    /// `corpus_open_failed` check. It is safe because of the invariant in
+    /// `core::indexer::quarantine`: a quarantined index holds no
+    /// `CorpusStore`, so `commit_corpus_to_redb` early-returns and
+    /// `staging::should_stage` is `false` — a bulk reindex on a quarantined
+    /// index writes nothing durable at all.
     /// Test: covered by every `index_file`-based test in `indexer::tests`;
     /// the quarantine branch by
     /// `quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero`.
--- a/crates/trusty-search/src/core/indexer/mod.rs
+++ b/crates/trusty-search/src/core/indexer/mod.rs
@@ -43,6 +43,7 @@ mod ingest;
 pub(crate) mod migrations;
 mod persist;
 mod persist_hnsw;
+mod quarantine;
 mod search;
 pub mod typeahead;
 mod types;
@@ -322,7 +323,28 @@ pub struct CodeIndexer {
     /// from a legitimately empty, freshly-created index.  This flag lets the
     /// classifier emit `StageStatus::Failed` with an actionable message instead
     /// of the misleading `InProgress` / "walking" state (closes #1158).
+    ///
+    /// Issue #4122: this flag is ALSO the write-quarantine predicate. While it
+    /// is `true` the index refuses every incremental/watcher write (see
+    /// `indexer::quarantine`), because building a fresh corpus on top of an
+    /// unopened one permanently destroys the original. Cleared by
+    /// [`Self::set_corpus_store`] when a corpus open finally succeeds, so the
+    /// quarantine is recoverable rather than terminal.
     pub corpus_open_failed: bool,
+
+    /// Issue #4122: monotonic count of incremental writes refused because
+    /// [`Self::corpus_open_failed`] was set.
+    ///
+    /// Why: the refusal needs an observable surface that is not a log line —
+    /// tests need a deterministic condition to await instead of a sleep, and
+    /// operators need "how many saves were dropped?" answered numerically.
+    /// What: bumped by `refuse_incremental_write`, reset to 0 by
+    /// `clear_corpus_open_failure`. Read via
+    /// [`Self::refused_incremental_writes`]. Atomic because `index_file`
+    /// takes `&self`.
+    /// Test: `quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero`
+    /// in `tests/corpus_open_quarantine_4122.rs`.
+    pub(super) incremental_writes_refused: AtomicU64,
 
     /// Issue #2922: `true` when a persisted HNSW snapshot existed on disk at
     /// warm-boot/lazy-load time but could NOT be restored (missing/corrupt
@@ -463,6 +485,7 @@ impl CodeIndexer {
             lane_degraded: Arc::new(AtomicBool::new(false)),
             last_rehydrate_cost_ms: Arc::new(AtomicU64::new(0)),
             corpus_open_failed: false,
+            incremental_writes_refused: AtomicU64::new(0),
             hnsw_load_failed: false,
             skip_kg: false,
         }
@@ -702,10 +725,16 @@ impl CodeIndexer {
     /// Why: the daemon resolves one `index.redb` per index and wires it in
     /// before warm-boot.
     /// What: stores the `Arc` so both the ingest commit path and the
-    /// fire-and-forget persist task can reach it.
-    /// Test: `tests::test_corpus_store_roundtrip`.
+    /// fire-and-forget persist task can reach it. Issue #4122: a successful
+    /// open is exactly the event that makes writes safe again, so this also
+    /// lifts any active write quarantine via `clear_corpus_open_failure`
+    /// (a no-op on the healthy-boot path, where the flag is already `false`).
+    /// Test: `tests::test_corpus_store_roundtrip`;
+    /// `successful_reopen_lifts_quarantine_and_leaves_corpus_intact` in
+    /// `tests/corpus_open_quarantine_4122.rs` covers the recovery transition.
     pub fn set_corpus_store(&mut self, corpus: Arc<crate::core::corpus::CorpusStore>) {
         self.corpus = Some(corpus);
+        self.clear_corpus_open_failure();
     }
 
     /// Swap in a new durable corpus store, returning the one it replaced
--- a/crates/trusty-search/src/core/indexer/quarantine.rs
+++ b/crates/trusty-search/src/core/indexer/quarantine.rs
@@ -31,6 +31,30 @@
 //! answering with silent empty results is issue #4087 and is deliberately out
 //! of scope here.
 //!
+//! # LOAD-BEARING INVARIANT: `corpus_open_failed == true` ⇒ `self.corpus == None`
+//!
+//! This — NOT "reindexes are operator-initiated" — is what makes the ungated
+//! BULK path safe. Boot reconcile auto-fires full reindexes
+//! (`service::reconcile`) with no `corpus_open_failed` check at all, so a
+//! quarantined index CAN be reindexed without any human involved. That is
+//! harmless only because every durable redb write funnels through
+//! `self.corpus` and short-circuits when it is `None`
+//! (`ingest::commit::commit_corpus_to_redb` early-returns on `None`), and
+//! because `service::reindex::staging::should_stage(has_corpus_store())` is
+//! `false` here, so such a reindex never even stages.
+//!
+//! The invariant holds because the only producer of `corpus_open_failed ==
+//! true` (`service::persistence_loader::build_indexer_from_entry`) is exactly
+//! the path that wires NO corpus, and the only way to wire one
+//! ([`CodeIndexer::set_corpus_store`]) clears the flag in the same call.
+//!
+//! **If you add an in-process corpus reopen/retry, you will break this.** A
+//! retry that wires a corpus without clearing the flag — or that sets the flag
+//! on an index that still holds one — re-opens the exact data-loss hole this
+//! module closes, and every existing test stays green while it does. The two
+//! `debug_assert!`s (in [`CodeIndexer::refuse_incremental_write`] and in
+//! `commit_corpus_to_redb`) exist to make that failure loud instead of silent.
+//!
 //! Test: `crates/trusty-search/tests/corpus_open_quarantine_4122.rs`.
 
 use std::sync::atomic::Ordering;
@@ -103,6 +127,19 @@ impl CodeIndexer {
         if !self.corpus_open_failed {
             return false;
         }
+        // Issue #4122, see the module-level invariant: a quarantined index must
+        // never hold a wired corpus. If it ever does, the ungated BULK reindex
+        // path (auto-fired by boot reconcile) would write durably to a corpus
+        // this module has declared unsafe — silently, with CI green.
+        debug_assert!(
+            self.corpus.is_none(),
+            "index '{}': INVARIANT VIOLATED — corpus_open_failed is set while a \
+             CorpusStore is wired. The ungated bulk-reindex path is safe only \
+             because a quarantined index has no corpus to write through. Did you \
+             add an in-process corpus reopen that skips set_corpus_store (issue \
+             #4122)?",
+            self.index_id
+        );
         let refused = self
             .incremental_writes_refused
             .fetch_add(1, Ordering::Relaxed)
@@ -120,10 +157,14 @@ impl CodeIndexer {
                  watcher/incremental writes against an unopened corpus builds a fresh \
                  PARTIAL corpus over the original and destroys it permanently (issue \
                  #4122). {refused} write(s) refused so far; the on-disk corpus is \
-                 untouched and still recoverable. Fix the underlying redb file \
-                 (permissions, stale lock, corruption) and restart the daemon — a \
-                 successful corpus open lifts the quarantine automatically. Saves \
-                 dropped while quarantined are picked up by the next reindex."
+                 untouched and still recoverable. TO RECOVER: fix the underlying redb \
+                 file (permissions, stale lock, corruption), then RESTART THE DAEMON — \
+                 only a successful CorpusStore::open lifts the quarantine, and it is \
+                 attempted solely at load time (there is no in-process reopen retry). \
+                 A reindex will NOT clear this state and will NOT persist anything: \
+                 with no corpus wired it skips staging entirely, so its results are \
+                 discarded at the next restart. Saves dropped while quarantined are \
+                 picked up by the first reindex AFTER a successful restart."
             );
         } else {
             tracing::debug!(
--- a/crates/trusty-search/src/core/indexer/quarantine.rs
+++ b/crates/trusty-search/src/core/indexer/quarantine.rs
@@ -1,0 +1,174 @@
+//! Corpus-open write quarantine for [`CodeIndexer`] (issue #4122).
+//!
+//! Why: when an index's durable redb corpus FAILS to open at load time the
+//! loader wires no `CorpusStore` at all, leaves the in-memory chunk map empty,
+//! and sets [`CodeIndexer::corpus_open_failed`] — but the handle otherwise
+//! stays fully live, watcher included. In the #4122 production incident that
+//! combination turned a *recoverable* outage into permanent data loss: the
+//! file watcher for `duettoresearch-hackathon-tm-hackathon-01` kept accepting
+//! incremental writes from unrelated file saves, so `chunk_count` climbed
+//! `0 → 68 → 1334` and a FRESH, PARTIAL corpus was persisted over the
+//! never-opened original. The sibling index `cto-duetto` hit the identical
+//! open failure in the same boot, took no watcher writes, and recovered all
+//! 200,090 chunks on the next restart. Watcher-writes-during-failure was the
+//! sole difference between full recovery and unrecoverable loss.
+//!
+//! What: this module gives `CodeIndexer` a small quarantine surface —
+//! [`CodeIndexer::is_write_quarantined`] (the predicate),
+//! [`CodeIndexer::refuse_incremental_write`] (the enforcement + diagnostic),
+//! [`CodeIndexer::refused_incremental_writes`] (the observable counter), and
+//! [`CodeIndexer::clear_corpus_open_failure`] (the way back out, driven by
+//! `set_corpus_store`). The quarantine is deliberately *asymmetric*: refusing
+//! a write costs an un-indexed file save that the next reindex picks up
+//! anyway, whereas accepting one destroys the corpus. So every corpus-open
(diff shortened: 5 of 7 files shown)


bobmatnyc (commented):
## Code Critic — Adversarial Review of PR #4220

**Verdict: WARN** (0 CRITICAL, 2 HIGH, 2 MEDIUM, 1 LOW)

> Posted as a review **comment**, not `--request-changes`: GitHub rejects a
> change-request review from the account that authored the PR. This comment is
> the verdict of record — treat it as the review gate output.

Reviewed at `65458d7f` against the spec (issue #4122 + the incident), not against
the implementer's self-report. Every claim below was verified by reading the code
and by running the tests with the guards mechanically neutralized.

---

### Summary

The fix is real, the tests are real, and it is a strict improvement over `main`.
The incident's exact path is closed and independently proven closed. **Neither HIGH
finding says the fix is wrong** — both say that the *reasoning the PR writes down*
about why the remaining paths are safe is factually incorrect, and that the
operator-facing remediation the PR adds does not work. On a P0 data-loss fix whose
entire value proposition is "the system must not report a state it does not have,"
shipping remediation text that does not remediate is a defect worth tracking.

The asymmetry judgement (refuse on any corpus-open failure, do not try to
distinguish transient from permanent) is **correct on the merits here**, and
correctly *not* transferred from the sibling PR #4209. The distinction: in #4209
the uncertain verdict authorized a destructive action; here the uncertain verdict
authorizes *inaction*, whose cost is a bounded, loudly-reported, restart-recoverable
availability dent. Uncertainty may veto destruction; it may not order it. That is
the same principle, applied in the right direction. See risk 3 for the one place
where the over-refusal cost is larger than the PR argues.

---

### Risk-by-risk

**1. Does quarantine stop every write path? — CONFIRMED, but not for the stated reason.**

The `index_file` gate (`crates/trusty-search/src/core/indexer/ingest/mod.rs:250`)
does cover all three incremental producers — watcher (`service/watch_loop.rs:269`),
boot reconcile (`service/reconcile.rs:600`), HTTP `index-file`
(`service/server/files.rs:35`). The watcher additionally short-circuits at
`service/watch_loop.rs:221`.

Ungated write paths do exist: the bulk path (`commit_parsed_batch`,
`ingest/commit.rs:36`, reached from `reindex/batch.rs:457`, `reindex/defer_embed.rs:340,419`,
`core/migration/m001.rs:193`) and every deletion path (`remove_file` at
`files.rs:348` from `server/files.rs:54` and `reconcile.rs:617`;
`remove_file_no_kg_rebuild` at `files.rs:324`; `remove_chunk` at `files.rs:462`
from `watch_loop.rs:260,317`).

None of them can touch the durable corpus **while quarantined**, because every
redb write funnels through `self.corpus`, which is `None` in exactly that state:
`ingest/commit.rs:127` and `:341`, `files.rs:379-385`, `files.rs:403-412`, and
staging is skipped at `reindex/staging.rs:34-36` → `reindex/corpus_swap.rs:46-50`.
So the durable corpus is safe. See HIGH-2 — that safety rests on an invariant the
PR never states.

**2. Can the flag be silently cleared, or fail to be set? — REFUTED (no defect).**

`set_corpus_store` (`core/indexer/mod.rs:735-738`) is genuinely the only sink. All
three production wiring sites route through it: `service/persistence_loader/mod.rs:151`,
`reindex/corpus_swap.rs:264`, `reindex/corpus_swap.rs:362`. `swap_corpus_store`
(`mod.rs:747`) does not clear and cannot be reached on a quarantined index —
`begin_staged_corpus_swap` returns `Ok(None)` at `corpus_swap.rs:47-49` when
`has_corpus_store()` is false. `with_components` (`mod.rs`, the other constructor
path) wires embedder + vector store only, never the corpus. Default is `false`
(`mod.rs:487`); the only two setters of `true` are
`persistence_loader/mod.rs:162` and `:183`. No stale state in either direction.

**3. Over-refusal — the asymmetry argument survives, with one correction. See MEDIUM-2.**

"A successful open IS the recovery event" is true, but the set of events that
produce one is smaller than the PR implies. `open_corpus_with_retry` retries once
at 50 ms and never again; there is **no in-process reopen retry**. Recovery
therefore requires a daemon restart (or an evict → lazy-restore cycle rebuilding
the indexer). A reindex does **not** recover it — see HIGH-1. So a transient
`DatabaseAlreadyOpen` from a concurrent handle does *not* self-heal in-process; it
heals on the next restart.

That is still a real, routine, unprompted recovery event, and the state is not
silent: `corpus_open_failed` fails all warm-boot stages
(`service/warm_boot/stages.rs:90`) and `/health` reports degraded
(`server/tests_health_degraded.rs:29`), plus the new ERROR carries a running total.
So: not a permanent outage, no manual unquarantine lever needed, asymmetry holds.
It is a *longer* outage than "a successful open lifts it automatically" suggests.

**4. Empty / BM25-only distinction — CONFIRMED.**

A legitimately empty index opens cleanly, so the flag stays at its `false` default
and the index remains writable. Pinned by
`build_indexer_from_entry_does_not_flag_legitimately_empty_index_as_failed`
(`service/persistence_loader/tests.rs:687-723`); I ran the module: `test result:
ok. 15 passed; 0 failed`. Quarantine cannot capture that case.

**5. Do the tests genuinely guard the fix? — CONFIRMED, independently.**

Baseline on the branch: `test result: ok. 3 passed; 0 failed`. I then rebuilt
three times in a throwaway worktree with the guards neutralized but the API intact:

*Neutralization A — `refuse_incremental_write` always returns `false` (pre-fix behaviour):*
```
test quarantine_refusal_emits_error_level_diagnostic ... FAILED
test quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero ... FAILED
test successful_reopen_lifts_quarantine_and_leaves_corpus_intact ... FAILED
```
Critically, test 1 fails on the **corruption assertion**, not on a missing counter:
```
panicked at tests/corpus_open_quarantine_4122.rs:223:5:
#4122: a write-quarantined index must not grow a chunk corpus from watcher writes
  left: 1
 right: 0
```
It reached line 223, which means the healthy-sibling positive control at line 193
had already passed in the same run — the control is real, and a flat `chunk_count`
is proven to be a refusal rather than a stalled watcher.

*Neutralization D — drop only `refused == 1` from the throttle, keeping every 100th:*
```
test quarantine_refusal_emits_error_level_diagnostic ... FAILED
panicked at tests/corpus_open_quarantine_4122.rs:415:13:
#4122: the refusal must be emitted at ERROR ...; captured records were: []
```
The captured-records list is literally `[]` — the test asserts through the real
`BugCaptureLayer`/`ErrorStore`, not merely that a line was logged, and it pins the
first-refusal-must-be-ERROR property specifically.

*Neutralization C — `clear_corpus_open_failure` made a no-op:*
```
test successful_reopen_lifts_quarantine_and_leaves_corpus_intact ... FAILED
panicked at tests/corpus_open_quarantine_4122.rs:330:5:
#4122: a successful corpus open MUST lift the quarantine — this is the cto-duetto
recovery path and it must not be broken by the fix
```

Each test fails for its own distinct reason under its own targeted neutralization.
This is a well-constructed regression suite; the "all three fail pre-fix" claim
holds. `--lib watch_loop` (10 passed) and `--lib reconcile` (20 passed) are clean.

**6. ERROR-level throttling — CONFIRMED.**

`fetch_add` returns the previous value and `+ 1` is applied
(`core/indexer/quarantine.rs:106-109`), so the first refusal is `refused == 1` and
always takes the ERROR branch (`quarantine.rs:111`). Neutralization D proves the
first-refusal path is load-bearing and independently asserted. The running total
rides on every ERROR as `refused_writes` (`quarantine.rs:116`), so suppressed
DEBUG events are recoverable by arithmetic. The counter resets only on recovery
(`quarantine.rs:163`), so a re-quarantine correctly re-arms a loud first ERROR.

---

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-search/src/core/indexer/quarantine.rs` | 127-129, 167-171 | The operator remediation this PR adds does not remediate: a reindex neither lifts the quarantine nor writes anything durable | Replace "run a reindex" with "restart the daemon" as the only remediation; file a follow-up for an in-process reopen retry |
| HIGH | `crates/trusty-search/src/core/indexer/ingest/mod.rs` | 241-244 | The stated reason the bulk path is left ungated ("a reindex is an explicit operator action") is false — boot reconcile auto-triggers full reindexes; the real safety is an undocumented, unasserted invariant | Document the `corpus_open_failed ⇒ corpus.is_none()` invariant on the field and add a `debug_assert!` at the commit/delete choke points |
| MEDIUM | `crates/trusty-search/src/core/indexer/quarantine.rs` | 121-123 | "the on-disk corpus is untouched and still recoverable" is not quite true — the legacy JSON snapshot is the one durable write a quarantined index still performs, and it is ungated | Gate `save_chunks_to_disk` on the quarantine (skip the write) or soften the claim to name redb specifically |
| MEDIUM | `crates/trusty-search/src/core/indexer/quarantine.rs` | 23-25 | "or genuine corruption" is wrong: corruption never reaches the quarantine — it is absorbed by the recovery-and-recreate open | Correct the enumeration to lock contention + transient I/O + unresolvable path; note the recreate path as out of scope (#4087) |
| LOW | `crates/trusty-search/src/core/indexer/quarantine.rs` | 60-68 | `is_write_quarantined()` is documented as the predicate "the watcher glue uses", but the watcher calls `refuse_incremental_write` instead; the method has no production caller | Either use it in `watch_loop` or drop the watcher claim from the doc comment |

---

#### HIGH-1 — the remediation instruction does not remediate

`quarantine.rs:127-129` tells the operator:

> `Saves dropped while quarantined are picked up by the next reindex.`

and `quarantine.rs:170-171` repeats it on the recovery WARN
(`run a reindex to pick them up`), and `ingest/mod.rs:243-244` calls a reindex
`the documented way out of this state`.

Trace a reindex on a quarantined index:

1. `reindex/runner.rs:300` — `staging::should_stage(handle.indexer.read().await.has_corpus_store())`
2. `reindex/staging.rs:34-36` — `should_stage` *is* `has_durable_corpus`; quarantined ⇒ `false`
3. so `corpus_swap_tmp` is `None`, and `begin_staged_corpus_swap`
   (`corpus_swap.rs:47-49`) would have returned `Ok(None)` anyway
4. `reindex/finish.rs:526-546` — `if let Some(tmp_path) = corpus_swap_tmp` is
   false, so **neither** `commit_staged_corpus_swap` **nor**
   `abort_staged_corpus_swap` runs, and those are the only two reindex-path callers
   of `set_corpus_store` (`corpus_swap.rs:264`, `:362`)
5. `ingest/commit.rs:127` — `if self.corpus.is_some()` is false, so
   `commit_corpus_to_redb` never runs

Net effect: the reindex rebuilds the corpus **in memory only**, writes nothing to
`index.redb`, and leaves `corpus_open_failed == true`. The index is still
write-quarantined for the rest of the daemon's life, every subsequent save is still
refused, and the entire reindex output evaporates at restart (next boot restores
the old corpus from redb).

Why this matters more than a doc nit: an operator who follows the ERROR message
runs a reindex, watches search start returning results, and reasonably concludes
the incident is closed. It is not — the index is still refusing writes and still
producing zero durable state. That is a system reporting a state it does not have,
which is the precise failure class this PR was written to eliminate. The good news
is that the same trace proves the corpus is *safe*: an auto-triggered full reindex
on a quarantined index cannot rename anything over `index.redb`.

**Fix:** say "restart the daemon" and nothing else. Optionally file the follow-up:
a periodic corpus-reopen retry would make the dominant (transient — see MEDIUM-2)
failure population self-heal in-process, and would make "a successful open IS the
recovery event" true in practice rather than only at boot.

#### HIGH-2 — the ungated-bulk-path rationale is false; the real invariant is undocumented

`ingest/mod.rs:241-244` justifies leaving the bulk path ungated on the grounds that
`a reindex is an explicit operator action`. It is not always. `reconcile.rs:528`
`trigger_full_reindex` calls `spawn_reindex_with_cleanup` from three sites —
`reconcile.rs:367` (git diff unavailable for the stored SHA), `:379` (delta >
`FULL_REINDEX_THRESHOLD` = 250), `:458` (mtime delta > threshold) — during boot
reconcile, which walks every registered index (`reconcile.rs:243`) with **no**
`corpus_open_failed` check anywhere in the path. A corpus-open failure and an
auto-triggered full reindex are both warm-boot events and can co-occur in the same
boot, with no human in the loop.

What actually keeps that safe is a completely different property:

> `corpus_open_failed == true` ⇒ `self.corpus == None`

which holds by construction (`persistence_loader/mod.rs:151` is the only success
wiring; `:162` and `:183` are the only failure setters; `set_corpus_store` clears
the flag) and which makes every redb write a no-op and staging a skip. That
invariant is stated nowhere and asserted nowhere. It is load-bearing for the
delete paths too (`files.rs:379-385`, `:403-412`) — `remove_file` /
`remove_chunk` are ungated and are only harmless because the corpus handle is absent.

The risk is concrete and near-term: HIGH-1's natural fix is an in-process reopen
retry. Any implementation of that which assigns `self.corpus` without routing
through `set_corpus_store` — or which reopens *before* clearing the flag —
instantly makes the ungated bulk and delete paths live against a real, populated
corpus. Nothing in the code would catch it; CI would stay green.

**Fix:** state the invariant in the `corpus_open_failed` doc comment as the reason
the bulk/delete paths need no gate, and add
`debug_assert!(!(self.corpus_open_failed && self.corpus.is_some()))` at
`ingest/commit.rs:127` and `files.rs:433`. Then correct the "explicit operator
action" sentence — boot reconcile is a counter-example.

#### MEDIUM-1 — one durable write path the quarantine does not cover

`quarantine.rs:121-123` asserts `the on-disk corpus is untouched and still
recoverable`. For redb that is true. For the legacy JSON snapshot it is not:
`persist.rs:490-493` — `flush_corpus_to_disk` falls through to
`save_chunks_to_disk` precisely when no `CorpusStore` is wired, i.e. exactly the
quarantined case; and `persist.rs:37` documents `Empty corpus is still written so
the on-disk file accurately reflects state`. `persist_hnsw.rs:283` sets
`persist_chunks_json = self.corpus.is_none()` on the same condition. So at
shutdown a quarantined index writes its (now deliberately empty) in-memory corpus
over `chunks.json` — the same file `JsonCorpusToRedbMigration`
(`indexer/migrations.rs:120-135`) reads to seed an empty redb.

Practical blast radius is narrow: the JSON migration only runs when redb is empty
*and* the schema stamp is unset, so in most configurations this file is dead
weight. It is pre-existing, not introduced here. But it is the one durable write a
quarantined index still performs, the PR's own diagnostic claims otherwise, and
the fix is one line.

#### MEDIUM-2 — corruption never quarantines; the population is transient-dominated

The module doc (`quarantine.rs:23-25`) lists `genuine corruption` among the
conditions that quarantine. It is not one. `CorpusStore::open`
(`core/corpus/store_impl.rs:71`) delegates to `open_corpus_db_or_recreate`
(`core/corpus_recovery.rs:45-72`), which classifies `UpgradeRequired`,
`RepairAborted`, `Storage(Corrupted(_))` and `Storage(Io(InvalidData))` as
recoverable (`corpus_recovery.rs:100-105`), moves the file aside with a
`.v2-incompatible` suffix, and returns **`Ok`** with a fresh empty corpus.
`corpus_recovery.rs:98` says it outright: `Returns false for lock contention and
genuine transient I/O errors` — those are the *only* errors that propagate as
`Err` and set `corpus_open_failed`.

Two consequences worth writing down:

- The quarantine population is dominated by **transient** failures — precisely the
  population an in-process retry would clear. That materially strengthens the
  case for the HIGH-1 follow-up and weakens "there is no separate unquarantine
  lever to forget to pull" as a completeness argument.
- The recreate path produces an index that reports **healthy** with a fresh empty
  corpus, which watcher writes then populate with a partial rebuild — a symptom
  shape close to the incident's, and one this quarantine explicitly does not
  cover. Worth a sentence scoping it against #4087 so the next reader does not
  assume #4122 closed it.

---

### Notes

- **Pre-existing `trusty-agents` failures — unrelated, confirmed.** The diff is
  confined to `crates/trusty-search/{src,tests}` plus that crate's CHANGELOG. The
  only new behaviour is a refusal that fires exclusively when
  `corpus_open_failed` is set, which requires `build_indexer_from_entry` to hit a
  failed redb open. Neither `tests/api_e2e` ("did not become healthy within 5s" —
  a startup timeout, consistent with load average 110) nor the `python_skill`
  uv/interpreter probe has any path to that flag. No plausible mechanism links
  them to this diff.
- **CI:** every check green except `Test`, still in progress at review time.
  Clippy, MSRV 1.91, Format, doc-comment pointer lint, the trusty-search daemon
  smoke test, and both AL2023 legs all pass.
- **Not flagged:** the ungated deletion paths (`remove_file`, `remove_chunk`), on
  the grounds that they provably cannot reach redb while quarantined. That
  judgement is entirely contingent on HIGH-2's invariant; if HIGH-2 is not
  addressed, revisit them.
- **Recommendation:** none of this blocks landing. The fix closes a P0 data-loss
  path with verified regression coverage and adds no new hazard. HIGH-1 is a
  three-line correction to two strings and a doc comment and is worth doing in
  this PR — shipping remediation text that does not remediate is how the *next*
  incident gets mis-triaged. HIGH-2 is a doc comment plus two `debug_assert!`s.
  MEDIUM-1 and MEDIUM-2 are fine as follow-ups.

<sub>Review by the `code-critic` agent — rubric: `code-review-standards`, 80%-confidence filter applied. Findings verified by reading the code and by re-running the suite with each guard neutralized in an isolated worktree.</sub>

bobmatnyc:
Both HIGHs addressed in `fb441d95`, rebased onto current `main`. No behaviour change to the quarantine itself — the mechanism you confirmed is untouched.

## HIGH-1 — the remediation did not remediate

Confirmed by reading the path, not taken on faith: `staging::should_stage(has_corpus_store())` (`staging.rs:34-36`) is `false` on a quarantined index, so `runner.rs:299` leaves `corpus_swap_tmp` as `None`, `finish.rs:526` falls to the `else if` that only writes `indexed_root`, `set_corpus_store` is never called, and `commit_corpus_to_redb` (`commit.rs:341-343`) early-returns on `corpus == None`. A reindex on a quarantined index writes nothing durable and clears nothing.

The ERROR now says so:

> **TO RECOVER:** fix the underlying redb file (permissions, stale lock, corruption), then **RESTART THE DAEMON** — only a successful `CorpusStore::open` lifts the quarantine, and it is attempted solely at load time (there is no in-process reopen retry). A reindex will **NOT** clear this state and will **NOT** persist anything: with no corpus wired it skips staging entirely, so its results are discarded at the next restart. Saves dropped while quarantined are picked up by the first reindex **AFTER** a successful restart.

## HIGH-2 — the real invariant, documented and pinned

You are right that "a reindex is an explicit operator action" is false — `reconcile.rs:367,379,458` auto-fires them with no quarantine check. The load-bearing invariant is `corpus_open_failed == true` ⇒ `self.corpus == None`. It is now written down in three places (the `quarantine` module docs under a `LOAD-BEARING INVARIANT` heading, at `commit_corpus_to_redb`, and replacing the false claim at `ingest::index_file`), including the explicit warning that an in-process reopen retry is what would break it.

Two `debug_assert!`s pin it. **Proof they fire** — the violating state is reachable through public API, since `swap_corpus_store` deliberately does not clear the flag:

```
---- probe_a_refuse_incremental_write_assert_fires ----
panicked at crates/trusty-search/src/core/indexer/quarantine.rs:134:9:
index 'probe-4122': INVARIANT VIOLATED — corpus_open_failed is set while a
CorpusStore is wired. The ungated bulk-reindex path is safe only because a
quarantined index has no corpus to write through. Did you add an in-process
corpus reopen that skips set_corpus_store (issue #4122)?

---- probe_b_commit_corpus_to_redb_assert_fires ----
panicked at crates/trusty-search/src/core/indexer/ingest/commit.rs:352:9:
index 'probe-4122': INVARIANT VIOLATED — durable corpus write attempted while
the index is write-quarantined (corpus_open_failed). A quarantined index must
hold no CorpusStore; see core::indexer::quarantine (issue #4122).

test result: ok. 2 passed; 0 failed  (both #[should_panic])
```

The probe was temporary and is **not** in the commit — the asserts pin the invariant in production; the probe only proved they are wired.

## CHANGELOG trap — caught

It fired exactly as described. The rebase moved my entry inside the dated `## [0.40.0] — 2026-07-27` section with an empty `[Unreleased]` above and **no conflict marker**. Moved back; verified by line number: `[Unreleased]` at 8, entry at 13, `[0.40.0]` at 40.

## Gate

Targeted per instruction — `-p trusty-search` only, no workspace run.

```
cargo fmt --check -p trusty-search          FMT_EXIT=0
cargo clippy -p trusty-search --all-targets -- -D warnings
                                            Finished (clean, no findings)
cargo test -p trusty-search --no-fail-fast  TEST_EXIT=0
  test result: ok. 1366 passed; 0 failed; 1 ignored
  test result: ok.  319 passed; 0 failed; 1 ignored
  corpus_open_quarantine_4122: 3 passed; 0 failed
  (+18 more targets, all ok, 0 failed)
scripts/check_line_cap.sh                   7 allowlisted, 0 violations — OK
```

## Not fixed, per instruction

MEDIUM-1, MEDIUM-2 and the LOW are left alone. Filed as follow-ups: **#4226** (`chunks.json` is the one durable write a quarantined index still performs) and **#4227** (genuine corruption never quarantines — `corpus_recovery` absorbs it, so the population is transient-dominated). One thing to flag rather than silently leave: both MEDIUMs land on wording inside strings/docs this commit touched — MEDIUM-1 on the ERROR's "the on-disk corpus is untouched" (true for redb, not for `chunks.json`), MEDIUM-2 on the module doc still listing "genuine corruption" among the quarantining conditions when `corpus_recovery.rs` absorbs it. Both are one-clause doc corrections with no behaviour change. Deliberately not made here to hold scope; say the word and they are a two-line follow-up.

bobmatnyc:
CI green on `1afb1dac` — **24/24 pass**, including the `Test` leg that went red on `fb441d95`.

The red was **my own test**, and it failed on its **positive control** — which is the control doing exactly what it was built for: it refused to report a pass from an environment where the healthy sibling index never indexed, so a refusal could not be told apart from a stalled watcher. No product code was involved; the fix is test-only.

Two defects, both mine:

1. **Default current-thread runtime.** The watcher bridges its OS notification thread into the async loop, and that pipeline can starve on a single worker. Every passing watcher test in this crate already opts into `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` (`watch_loop::tests::modified_file_triggers_indexing`); mine did not. That is why it passed on macOS/FSEvents and failed on Linux/inotify.
2. **One-shot save.** The probe file was written exactly once, immediately after spawning the watch loop. An event that lands before the OS watch is armed is lost permanently and nothing regenerates it — so a startup race became a full budget burn (`finished in 30.02s`).

Fixed with the multi-thread runtime plus `await_condition_resaving`, which rewrites the probe with changing content once per 500 ms debounce window while polling the observable every 25 ms.

**This strengthens the test.** The broken index now gets *more* chances to wrongly accept a write, and the control must still demonstrably index first. Re-verified by neutralizing `refuse_incremental_write` to always return `false` — all three still fail, and test 1 still fails on the **corruption assertion**, not on the control or a missing counter:

```
assertion `left == right` failed: #4122: a write-quarantined index must not grow
a chunk corpus from watcher writes — this is the 0 → 68 → 1334 climb that
destroyed duettoresearch-hackathon-tm-hackathon-01
  left: 1
 right: 0
```

Green on the real target, from the Linux CI log — note the ~1s, i.e. the watcher is genuinely live, not a longer timeout papering over the race:

```
05:59:36.397  test quarantine_refusal_emits_error_level_diagnostic ... ok
05:59:36.467  test successful_reopen_lifts_quarantine_and_leaves_corpus_intact ... ok
05:59:37.455  test quarantined_index_refuses_watcher_write_and_chunk_count_stays_zero ... ok
```

Locally: 5/5 consecutive clean runs, 0.76–0.81s each.

bobmatnyc:
## Code Critic — re-review at head `1afb1dac`

## VERDICT: APPROVE

Re-review of the delta since the WARN at `65458d7f` (commits `fb441d95` + `1afb1dac`).
Reviewed from an independent worktree cut from `origin/main`. CI is green on exactly this
SHA (`gh run view 30332591857` → `headSha=1afb1dac…`, `conclusion=success`; `Test` 15m30s
pass, Clippy pass, MSRV 1.91 pass).

**CRITICAL: 0 · HIGH: 0 · MEDIUM: 2 · LOW: 1**

---

## Both prior HIGH findings: genuinely resolved

### HIGH-1 (recovery advice pointed at an action that does not recover) — RESOLVED

`crates/trusty-search/src/core/indexer/quarantine.rs:160-167` now reads:

> `TO RECOVER: fix the underlying redb file (permissions, stale lock, corruption), then
> RESTART THE DAEMON — only a successful CorpusStore::open lifts the quarantine … A
> reindex will NOT clear this state and will NOT persist anything: with no corpus wired
> it skips staging entirely, so its results are discarded at the next restart.`

I verified that claim chain against the code, not against the commit message:

- `service/reindex/runner.rs:300` gates staging on
  `staging::should_stage(handle.indexer.read().await.has_corpus_store())`. On a
  quarantined index `has_corpus_store()` is `false`, and `should_stage(false)` is `false`
  (`service/reindex/staging.rs:34`, pinned by `should_stage_without_durable_corpus`), so
  `corpus_swap_tmp == None`.
- `service/reindex/finish.rs:526` therefore takes neither swap branch; the `else if` at
  `:546` only calls `write_indexed_root` — no durable corpus write anywhere on that path.
- `core/indexer/ingest/commit.rs:349` early-returns on `self.corpus == None`.

Net: a reindex on a quarantined index writes nothing durable and does not clear the flag.
The corrected string points at the only event that does — `set_corpus_store` →
`clear_corpus_open_failure` (`core/indexer/mod.rs:737`).

### HIGH-2 (false "a reindex is an explicit operator action" premise) — RESOLVED

The author did **not** reach for a reopen retry. Delivered exactly as specified:

- **Doc comment**: `quarantine.rs:34-56`, headed
  `# LOAD-BEARING INVARIANT: corpus_open_failed == true ⇒ self.corpus == None`, and it
  names the falsified premise outright: *"This — NOT 'reindexes are operator-initiated' —
  is what makes the ungated BULK path safe. Boot reconcile auto-fires full reindexes
  (`service::reconcile`) with no `corpus_open_failed` check at all."* That is accurate:
  `service/reconcile.rs:367`, `:379`, `:458` each call `trigger_full_reindex(handle)` with
  no quarantine check. The corrected reasoning is mirrored at `ingest/mod.rs:243-251` and
  `ingest/commit.rs:336-342`; the old claim survives nowhere in the source.
- **debug_assert #1**: `quarantine.rs:134-142` — `debug_assert!(self.corpus.is_none(), …)`
  in `refuse_incremental_write`, placed after the not-quarantined early return.
- **debug_assert #2**: `ingest/commit.rs:352-359` — `debug_assert!(!self.corpus_open_failed, …)`
  in `commit_corpus_to_redb`, immediately after the `Some(corpus)` bind.

Both encode the real invariant and both are **reachable**, so neither is decoration:
`corpus_open_failed` is a `pub` field (`core/indexer/mod.rs:333`) written directly at
`service/persistence_loader/mod.rs:162` and `:183`, so a future in-process reopen that
wires a corpus without going through `set_corpus_store` reaches the violating state via
ordinary crate code. See MEDIUM-2 for the residual gap.

I also independently confirmed the invariant's premise — that `self.corpus` really is the
sole durable-redb funnel. Every durable write and delete gates on it:
`ingest/commit.rs:349` (batch write), `core/indexer/persist.rs:440`
(`migrate_corpus_to_redb`), `core/indexer/files.rs:380` (`delete_entities`),
`core/indexer/files.rs:404` (`delete_chunks`). The two in-process reopens that do exist —
`service/reindex/corpus_swap.rs:251` and `:341` — both funnel through `set_corpus_store`
(`:264`, `:362`) and are unreachable on a quarantined index, since they run only when
staging began.

---

## Findings

|---|---|---|---|---|
| MEDIUM | (PR description) | behaviour table + "Relationship to #4087" section | The PR body still asserts **both** claims the code was just corrected to remove: *"A reindex is an explicit operator action that rebuilds the corpus wholesale and is the documented way out of this state."* This repo squash-merges with the PR body as the commit message, so the false premise (HIGH-2) and the actively harmful recovery advice (HIGH-1) would be enshrined in permanent history, directly contradicting `quarantine.rs:34-44` and `:160-167`. The next person to read `git log` gets the version that was wrong. | Edit the PR body before merge: rewrite that table row to cite the `corpus_open_failed ⇒ corpus == None` invariant, and delete "is the documented way out of this state" — the way out is a daemon restart. |
| MEDIUM | crates/trusty-search/src/core/indexer/quarantine.rs | 134 (and `ingest/commit.rs:352`) | Neither `debug_assert!` is exercised by any test. Nothing in `tests/corpus_open_quarantine_4122.rs` (or elsewhere) constructs the violating state — `successful_reopen_lifts_quarantine_and_leaves_corpus_intact` recovers via `set_corpus_store`, which clears the flag. So for the exact future change the asserts exist to catch — an in-process reopen that skips `set_corpus_store` — CI would still go green: the tripwire is never stepped on. (They are also compiled out of the shipped release daemon; that is inherent to `debug_assert` and was the spec.) | Add one `#[should_panic(expected = "INVARIANT VIOLATED")]` test per assert, reaching the state through the public `corpus_open_failed` field: set it `true` on an indexer that already holds a corpus, then call `index_file` (assert 1) and drive a batch commit (assert 2). That turns the asserts from a comment into a gate. |
| LOW | crates/trusty-search/src/core/indexer/quarantine.rs | 163 | The operator string says a successful open "is attempted solely at load time (there is no in-process reopen retry)". There *are* in-process reopens — `service/reindex/corpus_swap.rs:251` and `:341` — they are merely unreachable on a quarantined index. A maintainer who greps `CorpusStore::open` reads the parenthetical as false and discounts the rest of the message. | Scope it: "…is attempted at load time, and the only in-process reopens (the reindex staging swap) cannot run on a quarantined index because it holds no corpus store." |

---

## Required Changes (before merge)

1. Update the PR body so the squash-merge commit does not record the two claims this PR's
   own commit `fb441d95` corrected (MEDIUM-1). This is the only item worth holding for.

MEDIUM-2 and LOW-1 are fine as follow-ups.

---

## Notes

**Fail-open / cursor-advance hunt — run, nothing found.** I looked specifically for the
recurring shape in this repo (the failure branch advances state anyway, so alarms read
healthy):

- `service/reconcile.rs:637` — `apply_delta` refuses to stamp when every non-skipped file
  errored (`!files.is_empty() && indexed == 0 && removed == 0 && failed > 0`), returns
  `false`, and both callers set `s.degraded = true` (`:390-397`, `:479-486`). On a
  quarantined index every `index_file` returns `Err` and `remove_file` can only return `0`
  (empty in-memory map), so this lands squarely in the no-stamp branch. Correct — and the
  `Err`-not-`Ok(())` refusal is exactly what makes it work; a silent `Ok(())` would have
  counted as `indexed += 1` and stamped the SHA over a reconcile that indexed nothing.
- `finish.rs:359` writes `indexed_head_sha` in memory only, and
  `update_last_indexed_unix` (`service/persistence_timestamps.rs:78`) has no callers in
  `src/`. So an evaporated reindex on a quarantined index cannot advance a *persisted*
  cursor and hide the gap after the restart the ERROR now recommends.
- `service/shutdown_flush.rs:373` → `persist.rs:490`: with `self.corpus == None` the
  shutdown flush falls through to `save_chunks_to_disk`, which unconditionally rewrites
  `chunks.json` from the (empty) in-memory map. I traced this to a conclusion and am
  **not** flagging it: `persistence_loader/mod.rs:212` prefers redb and consults the JSON
  migration path only when redb yields `0`/`Err`, so the redb corpus this PR protects is
  never displaced by it. Pre-existing, and no loss scenario survived scrutiny at >80%
  confidence.

**Not re-litigated, per the prior verdict:** the core quarantine mechanism, the
`Err`-not-`Ok` refusal, and the refusal to import PR #4209's over-refusal reasoning (the
asymmetry genuinely inverts — there uncertainty authorized destruction, here it authorizes
inaction).

**Test delta (`1afb1dac`) reviewed and judged strengthening, not weakening.** The switch to
`#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` matches
`watch_loop::tests::modified_file_triggers_indexing`, and `await_condition_resaving`
(`tests/corpus_open_quarantine_4122.rs:76-98`) gives the quarantined index *more* chances
to wrongly accept a write while the positive control must still demonstrably index first
(`:216-220`). The headline corruption assertion at `:246-252` is unchanged, and the
"decided" latch at `:227-242` still keeps the corruption assertion — not a counter check —
as the one that fails on regression.

PR #3945, opened 2026-07-25
feat(trusty-agents-ui): five-section agent config — Personality / Knowledge / Skills / Listeners / Permissions (#3932)

bobmatnyc:
Implements **DOC-57 Phase 1** — `docs/specs/agent-config-five-sections.md`, `SPEC-AGENTCFG-07~draft` §8. Epic #3931.

Owner directive being implemented:

> *"Instead of tools, let's show skills. Should be Personality, Knowledge (list knowledge tools including MCP connections to knowledge stores), Skills (each tool should be wrapped in a skill), Listeners, the Permissions."*

**Front-end only.** No new HTTP route, no config-schema change, independently revertable (§8.5, G-7). Every pane maps onto data `GET /api/agents/:name`, `.../persona` and `.../stores` already serve; where the spec maps a section to data with no endpoint yet, the pane renders what exists and **states the gap** rather than filling it (G-4).

## Before / after

| # | Before | After | Change |
|---|---|---|---|
| 1 | `personality` — Personality | `personality` — Personality | unchanged behavior; fabricated per-connector list removed (P-2) |
| 2 | `okg` — OKG Stores | `knowledge` — **Knowledge** | rename + widen to three sub-surfaces (§4) |
| 3 | `tools` — Tools | `skills` — **Skills** | replace: glob textarea → synthetic skill cards (§5, G-3) |
| 4 | `permissions` — Permissions | `listeners` — **Listeners** | Listeners moves up; hardcoded badge removed (L-1) |
| 5 | `listeners` — Listeners | `permissions` — **Permissions** | reorder to last + re-back with `enforced` markers (§7) |

### What each pane actually shows now

- **Personality** — unchanged: #3895's full-height `persona.md` editor, same save path, same sizing invariants. The block that enumerated `DEFINED_LISTENERS` and asserted *"no event instructions configured"* for each connector is gone; §3.2's P-2 requires that block be backed by the real `events/*.md` file set, and no route lists those files, so the pane states the gap instead.
- **Knowledge** — K-a live store bindings (#3878, unchanged rendering); K-b the `vector_search`→store linkage §4.3 names normatively, resolved against the agent's own `[tools].allow` and rendered *under* the store it reads; K-c the declared MCP knowledge endpoints, carrying the shipped `enabled` flags — so `trusty-memory` and `trusty-search` read **DISABLED with their reason** rather than being omitted (C-03.2). Read-only; store editing stays #3890 (K-3).
[...]

Issue #3745, opened 2026-07-23
fix(trusty-agents): agent picker missing Izzie/CTO Bot + task-poll latency


bobmatnyc:
Two demo-critical findings from Bob's test drive of the /Applications build (a05f82c3).

## A. Agent picker does not list Izzie or CTO Bot
`GET /api/agents` returned an EMPTY catalog in the packaged macOS `.app`, so the picker showed only the frontend's hard-coded base "Assistant". Two root causes, both proven live:

1. **cwd/dir resolution:** `list_agents_route` scanned a SINGLE cwd-relative `.trusty-agents/agents`. The Tauri sidecar runs the API with cwd `/` (sealed APFS volume in the `.app`), so it resolved to `/.trusty-agents/agents` → doesn't exist → `{"agents":[]}`. Evidence: `tagent --api` from a project dir returns 22 agents; from `/tmp` (no project tier, == the bundle's cwd) returns `{"agents":[]}`.
2. **flat-only glob:** even against the right dir, the scan only globbed flat `*.toml` and never recursed directory PACKAGES — `izzie`/`cto-assistant`/`assistant` ship as `<name>/agent.toml`.

**Fix:** scan all `agents_dir_candidates()` (project-local + `$HOME` bundle tier) and resolve both packages and flat files (package wins, backups skipped), matching runtime dispatch. Verified live: catalog 0 → 22 agents including assistant→"Assistant", izzie→"Izzie", cto-assistant→"CTO Bot".

## B. Task-poll perceived latency
Both poll loops (Tauri `send_message` + browser `transport.ts`) used a flat 1500ms interval with no token streaming — up to +1.5s dead time per turn (trivial turn ~2.1s backend, felt ~3.5s). Now 250ms for the first ~20 polls, then 500ms backoff.

## C. (separate finding — needs triage) izzie has no working weather tool
While setting the weather test location, `tagent --direct izzie --task "what's the weather"` showed izzie's loaded tool registry scoped to `[git_log, git_status, delegate_to_agent, web_search]` — NO `get_weather`. The izzie PACKAGE's `[tools].allow` DOES list `get_weather`/`get_train_schedule`/`get_train_alerts`, but the loaded config used the base assistant allowlist, so weather fails ("weather tool currently unavailable") regardless of location. Likely a flat-`izzie.toml`-shadows-package OR tool-registration gap in the `--direct` subagent path. Not fixed here.

[...]

PR #4221, opened 2026-07-28
feat(trusty-agents): L0-only GitHub PR/CI inspection tool surface (#4170)

### LINKED ISSUE #4170: feat(trusty-agents): GitHub PR/CI tool surface for L0 orchestration

bobmatnyc:
Closes #4170. Part of epic #4167. Builds directly on #4200 (squash `ada4d351`), which established the `AgentTier` resolver this PR gates on.

## The gap this closes

Verified against the tree before building: the persona tool surface had **no** GitHub PR/CI capability at all.

- `crate::tools::git_tools` is twelve **local** git operations (`git_log`, `git_status`, `git_branches`, `git_search_commits`, …) with no GitHub API reach.
- The only `gh` invocations anywhere in the crate were the **issue-tracker** adapter (`ticketing::gh_cli`, driving `gh issue …` behind the `TicketingClient` trait) and `ticketing::actions` (a workflow-**file**-keyed trigger/status pair that needs a configured `TicketingClient` and returns re-shaped JSON).
- Nothing resembling `gh pr view`, `gh pr checks`, or a check-run status primitive existed.

## What this adds

Five **read-only** tools in a new `crate::tools::gh_tools` module:

| Tool | `gh` subcommand |
|---|---|
| `gh_pr_list` | `gh pr list` |
| `gh_pr_view` | `gh pr view` |
| `gh_pr_checks` | `gh pr checks` (the check-run aggregation primitive) |
| `gh_run_list` | `gh run list` |
| `gh_run_view` | `gh run view` (with `--log-failed`) |

Output is returned **verbatim** — gh's own human format, no JSON re-shaping — per #4170's "legible and unfiltered" acceptance criterion.

## ⚠️ NO MUTATING CAPABILITY IS GRANTED — read this before approving

**This PR grants no merge power, and no write power of any kind.** There is no `pr create`, `pr merge`, `pr edit`, `pr comment`, `pr close`, `run rerun`, or `workflow run` tool in it.

#4170's scope list *does* mention `gh pr create`, `gh pr merge`, `gh pr comment`, and `gh workflow run`. Those are **deliberately deferred, not overlooked**: merge/comment authority on a tier that is explicitly *not* sandboxed is an owner decision that should be granted knowingly and separately, rather than folded into the read surface that closes the stated gap. **If you want L0 to hold merge authority, that is a follow-up issue and a separate review** — say so and it will be filed.

`gh issue view` / `gh issue list` are also absent, for a different reason: they are **already present under other names** — `get_ticket` / `list_tickets` / `ticket_search`, which resolve to `gh issue …` through `ticketing::gh_cli::GhCliClient`. Adding a second spelling would break the catalog's 1:1 skill/tool rule for zero new capability.

A test (`gh_tools_expose_only_read_only_subcommands`) fails the build if a mutating verb ever appears in a tool's name or description, and `gh_surface_registers_no_mutating_tool_even_for_l0` asserts the same against the real L0 registry.

## Hard requirement 1 — grantable to L0 ONLY, enforced in code

`gh_tools(tier, root)` returns an **empty vector** for anything other than `AgentTier::L0Orchestration`. It consumes the fail-closed resolver from #4200 (`AgentInfo::tier()`): an absent, blank, or unrecognized `tier =` resolves to `L1Standard` and therefore to **no tools**. Indeterminate ⇒ deny.

The gate lives **inside the factory**, not at each registration site, so a future call site that forgets to guard its loop still registers nothing.

**Why registration is a sufficient enforcement boundary:** both persona dispatch paths derive an agent's callable set by *intersecting* its `[tools].allow` globs with the names actually present in the registry — `runtime::tool_registry::scope_assistant_allowed_tools` and `ctrl::pm_task::dispatch::persona::filter_persona_tool_names`. An L1 persona that declares `gh_pr_view`, or `gh_*`, or even `*`, matches nothing and is granted nothing.

Registered on **both** assistant dispatch paths (the `--direct`/`--agent` subprocess path and the persona-chat path). Registering in only one is exactly the divergence #3745 item C had to fix for izzie's tools.

The persona-chat site reads `persona_cfg.agent.tier()` **directly** rather than through the `role == "assistant"` branch used for `delegator_tier`: this is a capability grant, not a delegation edge, so it must fail closed for every persona whose tier does not explicitly resolve to L0 — including one carrying an unexpected `role`.

## Operand validation (the tool-level attack surface)

L0 is deliberately un-sandboxed, so these tools are the last thing between model output and a real GitHub API call. Argv is passed directly — **never a shell** — and every LLM-supplied operand must be non-empty, ≤200 chars, free of a leading `-`, and drawn from `[A-Za-z0-9._/#:@-]`. A value like `--json`, `-R attacker/repo`, or a smuggled second argument is refused with a message naming the field, before any spawn. `repo` must be exactly `owner/repo`; `state`/`status` are closed enums; `limit` clamps to 1–100.

`gh pr checks` reports check state through its **exit code** (non-zero when any check is red or pending). That is tolerated and returned as normal output — otherwise the single most useful CI primitive would report `is_error` on exactly the state an orchestrator calls it to observe.

## Not touched, per scope

The tier gate itself, `ASSISTANT_ALLOWED_DELEGATE_ROLES`, `NON_CODING_TARGETS`, and every persona under `~/.trusty-agents/agents/`. No L0 persona instance is created by this PR.

## One skill per tool (hard requirement 3)

Five rows in a new `skills/manifest/builtin/github.rs`, each declaring the `gh` CLI credential requirement with `env_var: None` — a `gh auth login` grant lives in the system keyring and is not probeable, so status stays *unclaimed* rather than guessed (#3945's honesty discipline). A new family file rather than appending to `ops.rs` keeps the collision surface with concurrent tier work to two lines in `builtin/mod.rs`.

## Tests — 30 new, all green

The three required proofs run through the **real registry-construction path**, not hand-built helper arguments: a persona TOML written to disk, loaded via `AgentConfig::by_name` under a `TAGENT_CONFIG_DIR` override (the same entry point `run_subagent` calls), then `cfg.agent.tier()` → `build_registry_for_agent` → the real `scope_assistant_allowed_tools` glob translation.

- `l1_persona_declaring_gh_tools_is_granted_none_through_the_real_path` — the probe persona declares `allow = ["*", "gh_*", <every literal tool name>]` and receives none of them. Includes a non-vacuity assertion: the `*` grant must yield *something*, or the test would pass for the wrong reason.
- `l0_persona_declaring_gh_tools_is_granted_them_through_the_real_path` — the counter-test that keeps the above honest.
- `indeterminate_tier_denies_gh_tools_through_the_real_path` — absent, empty, whitespace-only, and garbage `tier =` values all deny.
- Plus `assistant_tier_registry_omits_gh_tools_for_l1` / `..._includes_gh_tools_for_l0`, `gh_surface_registers_no_mutating_tool_even_for_l0`, and 24 module-level tests covering the factory gate, name/catalog agreement, schemas, flag-injection rejection, and subprocess outcome mapping (the non-zero-tolerance rule is pinned against POSIX `true`/`false` so it is deterministic without an installed, authenticated `gh` — a test that can only skip in CI proves nothing).

## Quality gate

| Gate | Result |
|---|---|
| `cargo clippy -- -D warnings` | exit 0 |
| `cargo test -p trusty-agents` (lib) | **2870 passed, 0 failed**, 11 ignored |
| `scripts/check_line_cap.sh` | 7 allowlisted, 0 violations — OK |
| `scripts/check_test_pointers.sh` | 0 dangling pointers — OK |

`tests/api_e2e.rs` fails locally under parallel test threads with *"api server did not become healthy within 5s"* — four `--api` children racing a 5 s startup budget on a loaded dev machine. It passes cleanly at `--test-threads=1` (4 passed, 0 failed). Unrelated to this change, which adds nothing to `--api` startup.

`crates/trusty-agents/src/ctrl/pm_task/dispatch/persona.rs` sits at **exactly** 500 SLOC on `main` — the cap — so any code addition would fail the line-cap gate. The change there is net-zero SLOC: the existing git-tools block was tightened by one line (hoisting the `current_dir()` call into a `cwd` binding) to make room for the one-line `gh_tools::register` call.

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -47,6 +47,27 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **A read-only GitHub PR/CI inspection surface granted to the L0
+  orchestration tier ONLY** (#4170, epic #4167). Five tools in a new
+  `tools::gh_tools` — `gh_pr_list`, `gh_pr_view`, `gh_pr_checks`,
+  `gh_run_list`, `gh_run_view` — returning `gh`'s own output verbatim rather
+  than re-shaped JSON. **No mutating capability is granted**: there is no
+  `pr create`/`merge`/`edit`/`comment`/`close`, no `run rerun`, no
+  `workflow run`, and a test fails the build if a mutating verb ever appears
+  in a tool's name or description. The tier gate lives INSIDE the factory
+  (which returns an empty vector for any tier but `L0Orchestration`), so a
+  future call site that forgets to guard its loop still registers nothing,
+  and an L1 persona declaring `gh_pr_view`, `gh_*` or `*` matches nothing and
+  is granted nothing. Registered on BOTH assistant dispatch paths — the
+  `--direct`/`--agent` subprocess path and the persona-chat path — since
+  registering in only one is the divergence #3745 item C had to fix.
+  Every LLM-supplied operand is passed as argv, never through a shell, and
+  must be non-empty, ≤200 chars, free of a leading `-`, and drawn from
+  `[A-Za-z0-9._/#:@-]`; `repo` must be exactly `owner/repo`, `state`/`status`
+  are closed enums, and `limit` clamps to 1–100. `gh pr checks` reports check
+  state through its exit code, so a non-zero exit is returned as normal
+  output rather than an error — otherwise the most useful CI primitive would
+  report failure on exactly the state an orchestrator calls it to observe.
 - **`l0_shell_exec` — a shell/build/test executor granted to the L0
   orchestration tier ONLY** (#4173, epic #4167). `tools::l0_exec` returns an
   empty tool vector for `AgentTier::L1Standard`, so an L1 persona never has the
--- a/crates/trusty-agents/src/agents/tests/loading.rs
+++ b/crates/trusty-agents/src/agents/tests/loading.rs
@@ -1990,6 +1990,21 @@ fn bundled_assistant_personas_resolve_l0_and_gain_nothing() {
                          review it deliberately.",
                         crate::tools::l0_exec::L0_SHELL_EXEC
                     );
+                    // #4170: and for the L0-only GitHub PR/CI surface. Same
+                    // glob reasoning; every name checked, so a tool added to
+                    // `GH_TOOL_NAMES` is covered without editing this test.
+                    for gh in crate::tools::gh_tools::GH_TOOL_NAMES {
+                        assert!(
+                            !crate::ctrl::pm_task::match_any_glob(
+                                gh,
+                                std::slice::from_ref(granted)
+                            ),
+                            "'{name}' declares '{granted}' in [tools].allow, which matches \
+                             the L0-only GitHub tool '{gh}' — as an L0 persona it would \
+                             hold it. That is a capability change, not a side effect: \
+                             review it deliberately."
+                        );
+                    }
                 }
             }
         } else {
--- a/crates/trusty-agents/src/ctrl/pm_task/dispatch/persona.rs
+++ b/crates/trusty-agents/src/ctrl/pm_task/dispatch/persona.rs
@@ -223,14 +223,26 @@ pub async fn run_pm_task_with_persona(
                     }
                 }
             }
-            if let Ok(repo) = crate::git::GitRepo::open(
-                &std::env::current_dir().unwrap_or_else(|_| PathBuf::from(".")),
-            ) {
+            let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("."));
+            if let Ok(repo) = crate::git::GitRepo::open(&cwd) {
                 for tool in crate::tools::git_tools::git_tools(repo.root.clone()) {
                     registry.register(tool);
                 }
             }
             register_ticketing_tools(&mut registry).await;
+            // #4170 (epic #4167): the L0-only GitHub PR/CI inspection surface.
+            // Registered on BOTH assistant dispatch paths — this one and
+            // `runtime::tool_registry::build_assistant_tier_registry` — because
+            // registering in only one is exactly the divergence #3745 item C
+            // had to fix for izzie's tools. Unlike `delegator_tier` below, this
+            // reads `persona_cfg.agent.tier()` DIRECTLY rather than through the
+            // `role == "assistant"` branch: this is a capability grant, not a
+            // delegation edge, so it must fail closed for every persona whose
+            // tier does not explicitly resolve to L0 — including one carrying
+            // an unexpected `role`. `gh_tools::register` denies any non-L0
+            // tier, so an L1 persona declaring `gh_*` finds nothing in
+            // `all_names` for `filter_persona_tool_names` to keep.
+            crate::tools::gh_tools::register(&mut registry, persona_cfg.agent.tier(), project_path);
 
             // #3285: tool parity with the session path
             // (`run_pm_task_with_history`) — delegation, CTRL
--- a/crates/trusty-agents/src/runtime/tool_registry.rs
+++ b/crates/trusty-agents/src/runtime/tool_registry.rs
@@ -567,6 +567,19 @@ pub(super) fn build_registry_for_agent(
 /// `l0_grant_fails_closed_for_malformed_tier_declaration`,
 /// `tainted_delegation_cannot_widen_into_the_l0_grant`,
 /// `l1_cannot_reach_the_l0_grant_through_an_untiered_intermediary`.
+///
+/// `delegator_tier` ALSO decides whether the GitHub PR/CI inspection tools
+/// (#4170, epic #4167) are registered at all: `crate::tools::gh_tools::gh_tools`
+/// yields them only for `AgentTier::L0Orchestration` and an empty vector
+/// otherwise — the same tier-gated-at-construction shape #4171 uses for the
+/// session-state surface above. Registration is the enforcement boundary for
+/// that grant — see `tools::gh_tools`'s module doc comment.
+/// Test: `l1_persona_declaring_gh_tools_is_granted_none_through_the_real_path`,
+/// `l0_persona_declaring_gh_tools_is_granted_them_through_the_real_path`,
+/// `unrecognized_tier_declaration_denies_gh_tools_through_the_real_path`,
+/// `gh_tools_follow_the_kind_derivation_when_no_tier_is_declared`,
+/// `assistant_tier_registry_omits_gh_tools_for_l1`,
+/// `assistant_tier_registry_includes_gh_tools_for_l0`.
 pub(crate) fn build_assistant_tier_registry(
     delegator_allow: Option<&[String]>,
     delegator_tier: crate::agents::AgentTier,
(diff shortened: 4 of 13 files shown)


bobmatnyc:
### Rebased onto `main` at `e7dfa3c9` (now carries #4171)

Head is now `2a608825`. One conflict, resolved by **keeping both sides**:

`runtime/tool_registry.rs` — #4171 (L0 session-state) and this PR each appended a `delegator_tier` paragraph to `build_assistant_tier_registry`'s doc comment, immediately above the same signature. Both paragraphs are kept, and #4171's `pub(super)` → `pub(crate)` widening is preserved. Verified both registrations survive and both features coexist:

- `crate::tools::gh_tools::register(&mut reg, delegator_tier, &cwd)` (this PR)
- `crate::tools::session_state::session_state_tools(&cwd, delegator_tier)` (#4171)

`tools/mod.rs` and `dispatch/persona.rs` auto-merged; `pub mod gh_tools;` and `pub mod session_state;` both present. Change set is unchanged at 11 files.

Worth noting the convergence: #4171 independently arrived at the **same shape** this PR uses — a factory that takes the resolved `AgentTier` and returns an empty vector for anything but L0, so registration is the enforcement boundary. Two sibling implementations landing on one pattern is a good sign the #4200 tier model is carrying its weight.

Two incidental notes:

- `dispatch/persona.rs` was at exactly 500 SLOC (the cap) when this PR was written, which is why it hoisted the `current_dir()` call into a `cwd` binding to make room. #4171 has since trimmed that file to 479 SLOC, so the headroom is no longer tight — the one-line hoist is kept because it is a strict readability improvement and reverting it would be churn.
- Re-verified on the final head: `cargo fmt --check` exit 0, `cargo clippy -p trusty-agents --all-targets -- -D warnings` exit 0, `cargo test -p trusty-agents --lib` **2935 passed / 0 failed**, `check_line_cap.sh` 0 violations.

CI green on both prior heads (22 pass / 1 skipped each): `c0b9abc9` (pre-rebase) and `c0753c1b` (rebased onto #4211).

bobmatnyc:
## Rebased onto `main` (42554d84, i.e. after #4217 landed) — grant-semantics note

The grant is unchanged: `gh_tools(tier, root)` still returns an empty vector
for anything but `AgentTier::L0Orchestration`, the gate still lives inside the
factory, both dispatch paths still register it, and the surface is still
**read-only** — no `pr create`/`merge`/`edit`/`comment`/`close`, no
`run rerun`, no `workflow run`.

**What changed underneath it:** `main` landed ADR-0024 decision 3 (PR #4296,
squash `54bba9f7`) — an undeclared tier is now DERIVED from the agent's KIND,
and assistant-kind resolves `L0Orchestration`. When this PR was written, an
assistant with no `tier =` resolved L1. So the *population* holding this grant
changed on `main`, not in this PR, and the shipped assistants (`assistant`,
`cto-assistant`, `ctrl`, `izzie`, `personal-assistant`) now have these five
tools registered.

Registration is not reachability — that is the ADR's own safety claim and it
still holds: an agent's callable set is its `[tools].allow` globs INTERSECTED
with what the registry holds, and no bundled assistant declares a pattern that
matches any `gh_*` name (`granola_*` is the only glob in the roster; `ctrl`
declares no `[tools]` section at all and fails closed on both its paths).
Verified, and now pinned mechanically rather than asserted.

Test changes — all widening coverage; nothing deleted, ignored, or narrowed:

1. `agents::tests::loading::bundled_assistant_personas_resolve_l0_and_gain_nothing`
   (the ADR's zero-delta guard) now also asserts no bundled assistant's
   `[tools].allow` can reach any entry of `GH_TOOL_NAMES`. Matched as a
   **glob**, not by equality — `gh_*` or `*` would reach them just as surely as
   a literal name. Iterating `GH_TOOL_NAMES` means a tool added to the factory
   is covered without editing the test.
2. `indeterminate_tier_denies_gh_tools_through_the_real_path` →
   `unrecognized_tier_declaration_denies_gh_tools_through_the_real_path`. Its
   old name conflated two cases that #4296 split apart. It now pins the one
   that would actually be an escalation: a declared-but-unrecognized `tier =`
   must fail closed to L1 and must NOT fall through to the assistant-kind
   derivation. The probe persona is assistant-kind precisely so a fall-through
   would show up as a grant. Four deny spellings, up from the three that
   remained meaningful.
3. New `gh_tools_follow_the_kind_derivation_when_no_tier_is_declared` pins the
   derivation at the grant site (assistant-kind → granted; a sub-agent kind →
   nothing, twice over, since `build_registry_for_agent` does not even route it
   into the assistant-tier registry), so a future change to who is L0 fails
   here instead of silently redefining this surface's reach.

Remaining conflicts were mechanical: doc-comment interleaving on
`build_assistant_tier_registry` (all three grant paragraphs — #4171
session-state, #4173 exec, #4170 gh — kept), and the `delegator_subagents`
parameter added by #4314's editable reachable-set whitelist threaded through
the new call sites. The probe helper passes the real
`cfg.subagents.delegate_allowed.as_deref()` that `run_subagent` passes, so the
call stays on the production path — resolved toward the config-whitelist shape.

`crates/trusty-agents/src/runtime/tool_registry_tests.rs` was rebuilt from the
merge stages rather than hand-edited: #4170's change to that file is 178
additions and 0 deletions appended at the end, so the resolution is
`main`'s version plus that block verbatim.

### Gates (this head)

| Gate | Result |
|---|---|
| `cargo test -p trusty-agents --lib` | **3077 passed, 1 failed**, 11 ignored — see below |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | exit 0 |
| `scripts/check_line_cap.sh` | 7 allowlisted, 0 violations — OK |

The single failure is `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json`,
an environmental `uv` temp-cache race under parallel test threads
(`Failed to query Python interpreter … No such file or directory … /.cache/uv/.tmp…`).
It passes in isolation — `cargo test -p trusty-agents --lib tools::python_skill`
→ `8 passed; 0 failed` — and it also failed then passed on the #4217 run in the
same worktree. Unrelated to this change, which touches no Python path.

CHANGELOG entry added under `## [Unreleased]` → `### Added`.

Issue #4007, opened 2026-07-26
epic: connect curated memory stores and attached search indexes


bobmatnyc:
## Outcome

Agents use a coherent two-tier knowledge model:

1. one curated trusty-memory/OKG store for governed agent knowledge;
2. zero or more attached trusty-search indexes for arbitrary user-selected corpora.

## Scope

- explicit configuration and runtime types for both tiers;
- attachment, selection, permissions, lifecycle, and status surfaces;
- unified retrieval orchestration with source/provenance metadata;
- clear ranking and conflict behavior across tiers;
- a reference flow attaching the APEX repository to `cto-assistant`;
- tests that preserve the curation boundary and prevent accidental ingestion into the OKG store.

## Done when

An agent can retrieve from both tiers in one conversation, users can see and control the selected sources, and every result identifies its tier and origin.

The former implementation tickets #4010–#4015 are consolidated here.

bobmatnyc:
Triage consolidation (2026-07-28): this issue is now the canonical, outcome-oriented tracker for its former implementation leaves. Scope and acceptance criteria were retained here; leaf tickets are being closed as `not planned` to reduce duplicate queue surface, not to claim the work is implemented.

bobmatnyc:
Closed not planned in the 2026-09-02 milestone-68 epic review (owner-confirmed, session tm-trusty-tools-02): every referenced child closed not-planned, nothing shipped under this epic since the 2026-08-14 backlog reset. Reopen if this scope is revived.

bobmatnyc:
Correction (2026-09-06, session trusty-tools-38): this closure's claim that every referenced child closed does not hold — #4282, #4283, #4289, and #4315 are open (#4283, #4289, #4315 resumed as of 2026-09-06 following a staleness review; #4282 awaiting an owner decision). Not reopening this epic; tracking continues on the child issues directly.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #3232, opened 2026-07-19
feat(trusty-agents): add tools.search_indexes field to ToolsConfig with union-extends semantics


bobmatnyc:
Enable per-agent trusty-search index attachment via overlay/extends semantics.

Specifics:
- Add tools.search_indexes: Option<Vec<String>> to ToolsConfig (crates/trusty-agents/src/agents/config.rs:118-177), listing index_ids (e.g. [\"trusty-tools\", \"cto-projects\"])
- Add one more union_opt_vec line in merge_extends (crates/trusty-agents/src/agents/extends/mod.rs:239-241) — the merge machinery already exists, schema-only

Current state: ToolsConfig has allowed, allow, native, ast_native, scopes (OpenRPC capability scopes like memory.read, search.read) — nothing per-index granularity. An agent either can or can't call the search tool family, with no notion of *which* trusty-search index_id it's allowed to target.

Target: A base Assistant agent's tools.search_indexes could list its default project index; a CTO Assistant extending it would union in \"cto-projects\", enabling overlay of per-agent index access without code duplication.

File evidence: crates/trusty-agents/src/agents/{config,extends/mod}.rs:118-177, :239-241, crates/trusty-search/src/mcp/tools/descriptors.rs:180-183, :185-199 (index_id schema)

Effort: S
Depends on: None; parallel with #3231

Part of #3052

PR #4217, opened 2026-07-28
feat(trusty-agents): L0-only shell/build/test execution grant (#4173)

bobmatnyc:
Grants the L0 orchestration tier a shell/build/test executor, `l0_shell_exec`,
behind a single code-level tier gate. L1 (standard) personas get nothing —
not "get it and are denied at dispatch", but **never have it registered**.

Closes #4173. Part of epic #4167 (L0/L1 tier model). Builds directly on
#4200 (`ada4d351`, the `AgentTier` resolver + one-directional delegation gate)
and #4161 (`56337097`, the delegation taint that shut the
prompt-injection-to-code-execution path in #4126). Shell execution is the exact capability
that made #4126 a P0, so this PR hands it back to precisely one tier and does
it in code.

## This IS a general shell — stated explicitly

#4173's acceptance criteria ask for shell command execution in the project
working tree, `cargo test` / build commands, "output legible and unfiltered",
and an `echo "hello"` smoke check. That cannot be met by a constrained
build/test-only verb, so this is a general executor and the PR says so rather
than implying a narrower surface. Two narrowings were applied where the issue
allowed them:

* **One tool, not three.** #4173 lists `shell_exec`, `run_test` and `run_build`
  but explicitly permits the latter two "or invoke via shell". One executor is
  one new execution surface instead of three.
* **No new command-content policy.** The catastrophic-pattern predicate is
  *reused* from `tools::run_bash::is_dangerous_command` (already documented
  there as "defense in depth, not a security boundary"), not re-invented. No
  prefix allowlist: the epic's L0 row is "Git operations: ✅ Full (clone, pull,
  rebase, force-push)", which an allowlist would break.

### Exactly who can reach it

An agent reaches `l0_shell_exec` only if **all** of the following hold:

1. Its `role` is the assistant tier, so it is built by
   `runtime::tool_registry::build_assistant_tier_registry`; **and**
2. `AgentInfo::tier()` resolves its own declared `tier` to
   `AgentTier::L0Orchestration` — i.e. the TOML literally says `tier = "l0"`
   or `tier = "orchestration"` (#4200's resolver: absent, blank, whitespace and
   every unrecognized string fail closed to `L1Standard`); **and**
3. Its own `[tools].allow` admits the name through the existing glob scoping;
   **and**
4. The caller's RBAC `ServiceTier` is `All` — the tool declares
   `restricted_tiers() = [ReadOnly, Analytics]`, so a guest Slack user or an
[...]

PR #4222, opened 2026-07-28
feat(trusty-agents): L0-only cross-project git/search scoping (#4172)

### LINKED ISSUE #4172: feat(trusty-agents): cross-project store/git scoping for L0 orchestration
## Task

Enable L0 (orchestration) persona to reach stores and git repos across multiple projects, while keeping L1 personas scoped to the Duetto corpus/OKG tree.

## Scope

Currently L1 personas are restricted to a single project's knowledge stores. L0 needs cross-project reach to perform PM orchestration (reconcile state across repos, link PRs, coordinate multi-repo fix rounds).

1. Extend the L0 agent initialization to grant:
   - `trusty-memory` access to all configured project-scoped stores (not just Duetto)
   - `trusty-search` index access across projects (or at least read access to public indexes)
   - Git workspace access outside the current project's `.base/.worktrees/` (for clone/pull operations across repos)
2. Keep L1 scoped to Duetto (assistant, cto-assistant, izzie).
3. Confirm that cross-project operations do not leak secrets or unintended data to untrusted consumers.

## Acceptance Criteria

- [ ] L0 can call trusty-memory tools with store IDs from multiple projects
- [ ] L0 can access trusty-search indexes outside the Duetto corpus
- [ ] L0 can clone/pull repos outside the current project
- [ ] L1 remains scoped to Duetto (test confirms this)
- [ ] No credential leakage across project boundaries

## Related

- Gap analysis: L1 "no cross-project reach" point

bobmatnyc:
Closes #4172. Part of epic #4167 (L0/L1 persona tier model); builds directly on #4200 (`AgentTier` + fail-closed resolver).

## The normative tension, and why it dissolves

#4172 asks for cross-project store/index reach. `crates/trusty-agents/src/agents/extends/mod.rs` carries a hard constraint in the opposite direction — a child's `[[stores]]` REPLACES the parent's because **DOC-54 §5.1 allows exactly ONE OKG store per agent** (Bob, 2026-07-24).

**This PR does not relax that rule, for either tier.** DOC-54 §5.1 governs `[[stores]]`: the agent's ONE *dedicated, curated* OKG store — its OKF knowledge tree plus the index over that tree. Nothing in this PR reads, writes, unions, or widens `[[stores]]`, and the `extends` replace-not-union rule is untouched.

The multiplicity #4172 asks for already has a sanctioned home. Epic #4007's **tier-2** mechanism — `[tools].search_indexes` (#3232) — exists precisely for "also let me search the APEX repo", and its own field docs state the split verbatim:

> `[[stores]]` is the CURATED tier — exactly one OKG store per agent by the 2026-07-24 owner decision, **and that ceiling stays**. The multiplicity an agent needs … belongs to a second, uncurated tier.

So cross-project *reach* is a tier-2 + filesystem question, not a `[[stores]]` question, and #4172 is implementable without touching a normative spec constraint. No owner escalation needed.

## What the bound is, and how it is enforced

New `crate::agents::cross_project::CrossProjectScope` — the single resolution point for "which repo roots and which search indexes may THIS agent touch".

**L0 allow-set** = the agent's own home project root, PLUS every entry of `~/.trusty-agents/projects.json` that is *simultaneously*:

| # | Condition | Enforced by |
|---|-----------|-------------|
| a | `ProjectStatus::Active` | `ProjectRegistry::list_active` + explicit re-check |
| b | `ProjectEntry::is_real_project()` (no `/tmp`, `/var/folders`, `.tmp*`) | the registry's own predicate |
| c | the directory exists | `canonicalize` at resolution time |
| d | it is a git repository root (`.git` present) | `is_git_repo_root` |

**L1 allow-set** = its home root, and nothing else — byte-identically today's behaviour.

Enforcement details:
- **Tier is never re-derived.** `AgentInfo::tier()` (#4200) is the only source; absent / blank / unrecognized (incl. `"l2"`) → `L1Standard` → single-tenant scope.
- **Registry failure denies reach.** No `$HOME`, missing file, or malformed JSON ⇒ empty snapshot ⇒ even an L0 agent collapses to its home root.
- **Containment is component-wise** `Path::starts_with` against *canonicalized* roots. Candidates must be absolute (a relative path is refused, never joined to the cwd) and must canonicalize (unprovable ⇒ refused). This defeats `..` traversal, symlink escape, and `/a/bc` vs `/a/b` string-prefix siblings.
- **Post-discovery re-check.** `GitRepo::open` is libgit2's `Repository::discover`, which walks *upward* — a permitted seed can still land on an ancestor work tree outside the allow-set. `open_scoped_repo` re-validates the discovered work tree. It runs only when the caller named an explicit `repo`, so the pre-#4172 contract (registration sites seed the process cwd, routinely a subdirectory) is preserved exactly.

**Auditability:** `audit_summary()` renders `tier=… cross_project=… roots=[…] indexes=[…]`, logged at `tracing::info` on construction and repeated verbatim in every refusal — at `tracing::warn` and in the error returned to the model.

**Why the trusty-agents registry and not trusty-mpm's:** trusty-mpm's curated registry is unreachable from this crate as Rust types (no `trusty-mpm` dependency; the sanctioned pattern is a loopback HTTP GET), carries **no local filesystem path** (records are keyed on `name`/`repo_url`), and a path can only be *fabricated* from it via the `~/trusty-mpm-projects/<owner>/<repo>` convention. A privilege boundary must not depend on a sibling daemon being up, nor on a guessed path. The in-crate registry is on-disk, deterministic, and user-inspectable via `list_projects`.

**Credential leakage:** none introduced. This resolves *paths* and *index ids* only; cross-project git runs through the same ambient credential helper the home project already uses. No per-project secret is read, copied, or forwarded.

## Wiring

- `git_tools_scoped(scope)` builds the 12 git tools over a resolved scope. The optional per-call `repo` argument is advertised in the JSON schema **only** when the scope is cross-project (so an L1 persona is never shown a knob it cannot turn; `additionalProperties: false` makes passing it a schema violation on top of the gate's refusal), and every call resolves it through the gate.
- `git_tools(root)` keeps its exact signature and single-tenant behaviour, so the three other registration sites (`runtime/tool_registry.rs`, `runtime/pm_mode.rs`, `ctrl/handlers/registry.rs`) are unchanged by construction.
- The persona dispatch path resolves ONE scope and feeds it to both consumers: the git surface and `vector_search`'s tier-2 index list (`effective_search_indexes`, which returns the declared list verbatim for any tier but L0).
- `ProjectRegistry::with_registry_path` — additive test seam so boundary tests drive the real loader without mutating process-global `$HOME`.

The delegation tier gate (#4200) and `~/.trusty-agents/agents/` personas are untouched. No persona declares `tier = "l0"` yet, so this PR changes no shipped agent's behaviour.

## Tests (24 new + 6 git-surface cases)

All exercise **real resolution paths**, not hand-built helper arguments:
- tier from a real `agent.toml` through `AgentConfig::by_name_in`;
- allow-set from a real `projects.json` through `ProjectRegistry::list_active`;
- git cases against real `git init`ed repos, driven through the tools' own `execute`.

Coverage: L1 scoping unchanged (`l1_scope_allows_only_its_home_root`, `l1_resolve_repo_root_refuses_another_project`, `l1_effective_search_indexes_are_declared_list_verbatim`, `git_schemas_omit_repo_arg_for_single_tenant_scope`, `single_tenant_git_tools_refuse_any_repo_argument`); L0 bounded and inescapable (`l0_resolve_repo_root_refuses_unregistered_sibling`, `..._refuses_parent_traversal_escape`, `..._refuses_symlink_escape`, `resolve_repo_root_refuses_prefix_lookalike_sibling`, `resolve_repo_root_refuses_relative_path`, `l0_scope_excludes_non_git_and_inactive_and_temp_entries`, `git_status_refuses_repo_outside_allow_set`, `git_write_tools_refuse_repo_outside_allow_set`); fail-closed (`indeterminate_tier_resolves_to_single_tenant_scope` over 5 indeterminate spellings, `from_registry_fails_closed_when_registry_is_unreadable`).

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -62,6 +62,31 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **Cross-project git/search scoping granted to the L0 orchestration tier
+  ONLY** (#4172, epic #4167). New `agents::cross_project::CrossProjectScope` is
+  the single resolution point for "which repo roots and which search indexes
+  may THIS agent touch". An L0 agent's allow-set is its own home project root
+  plus every `~/.trusty-agents/projects.json` entry that is *simultaneously*
+  `Active`, a real project (no `/tmp`, `/var/folders`, `.tmp*`), existing, and
+  a git repository root. An L1 agent gets its home root and nothing else —
+  byte-identical to previous behaviour. **DOC-54 §5.1's one-OKG-store-per-agent
+  ceiling is untouched for either tier**: nothing here reads, writes, unions or
+  widens `[[stores]]`, and `extends`' replace-not-union rule is unchanged;
+  cross-project reach is resolved as *paths and tier-2 index ids*
+  (`[tools].search_indexes`, #3232), which is where multiplicity already
+  belongs. Containment is component-wise `Path::starts_with` against
+  *canonicalized* roots, so `..` traversal, symlink escape, and `/a/bc` vs
+  `/a/b` prefix-lookalike siblings are all refused; a relative candidate is
+  refused outright rather than joined to the cwd. Because libgit2's
+  `Repository::discover` walks *upward*, `open_scoped_repo` re-validates the
+  discovered work tree after discovery — a permitted seed cannot land on an
+  ancestor outside the allow-set. Registry failure denies reach: no `$HOME`, a
+  missing file, or malformed JSON yields an empty snapshot, collapsing even an
+  L0 agent to its home root. The per-call `repo` argument is advertised in the
+  JSON schema only when the scope is actually cross-project, so an L1 persona
+  is never shown a knob it cannot turn. Every refusal is logged at `warn` with
+  the same `tier=… cross_project=… roots=[…] indexes=[…]` audit line emitted at
+  construction. No per-project credential is read, copied, or forwarded.
 - **A read-only GitHub PR/CI inspection surface granted to the L0
   orchestration tier ONLY** (#4170, epic #4167). Five tools in a new
   `tools::gh_tools` — `gh_pr_list`, `gh_pr_view`, `gh_pr_checks`,
--- a/crates/trusty-agents/src/agents/cross_project/mod.rs
+++ b/crates/trusty-agents/src/agents/cross_project/mod.rs
@@ -1,0 +1,491 @@
+//! L0-only cross-project git/search scoping (#4172, epic #4167).
+//!
+//! Why: The owner's tiered persona model (#4167, resolver landed in #4200)
+//! gives L0 ("orchestration") the PM-tier reach an L1 ("standard") assistant
+//! must never have. One of those grants is **cross-project reach**: an
+//! orchestration persona reconciling a snapshot against live state has to
+//! read git status across several repos and search corpora outside its own
+//! project, while every L1 persona stays exactly as single-tenant as it is
+//! today.
+//!
+//! **The one-store rule is NOT relaxed — for either tier.** DOC-54 §5.1
+//! ("Stores (Knowledge): One per agent", Bob 2026-07-24) governs
+//! `[[stores]]`: the agent's ONE dedicated, curated OKG store — its OKF
+//! knowledge tree plus the search index over that tree. That ceiling, and
+//! the `extends` replace-not-union rule that protects it
+//! (`crate::agents::extends`), are untouched by this module: nothing here
+//! reads, writes, unions, or widens `[[stores]]`.
+//!
+//! The multiplicity #4172 asks for already has a sanctioned home. Epic
+//! #4007's tier-2 mechanism — `[tools].search_indexes` (#3232) — exists
+//! precisely for "also let me search the APEX repo", and its own field docs
+//! (`ToolsConfig::search_indexes`) state the split: "`[[stores]]` is the
+//! CURATED tier — exactly one OKG store per agent by the 2026-07-24 owner
+//! decision, and that ceiling stays. The multiplicity an agent needs …
+//! belongs to a second, uncurated tier." So cross-project *reach* is a
+//! tier-2 + filesystem question, not a `[[stores]]` question, and #4172 is
+//! implementable without touching a normative constraint.
+//!
+//! What: [`CrossProjectScope`] — the single resolution point for "which repo
+//! roots and which search indexes may THIS agent touch". It is built from a
+//! real [`AgentInfo`] (so the tier comes from #4200's fail-closed
+//! `AgentInfo::tier`, never a local guess) plus the agent's home project
+//! root, and it answers two questions:
+//!
+//! - [`CrossProjectScope::resolve_repo_root`] — turn an optional caller-
+//!   supplied `repo` path into a permitted absolute root, or refuse.
+//! - [`CrossProjectScope::effective_search_indexes`] — widen an agent's
+//!   declared tier-2 index list with the cross-project index ids it may
+//!   query.
+//!
+//! **The bound (requirement: not "any path on disk").** An L0 agent's
+//! allow-set is:
+//!
+//! > its own home project root, PLUS every entry of the trusty-agents
+//! > project registry (`~/.trusty-agents/projects.json`) that is
+//! > simultaneously (a) `ProjectStatus::Active`, (b)
+//! > `ProjectEntry::is_real_project()` (no temp dirs), (c) an existing
+//! > directory, and (d) a git repository root (carries a `.git` entry).
+//!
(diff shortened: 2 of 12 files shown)


bobmatnyc:
## Rebased onto `main` (29a821b1, after #4217 and #4221 landed)

⚠️ **Unlike the other two PRs in this wave, this one now carries a REAL
capability delta for shipped personas. Please read the first section.**

The bound itself is unchanged: an L0 agent reaches its home root plus every
`projects.json` entry that is simultaneously Active, a real project, existing,
and a git repo root — and nothing else. DOC-54 §5.1's one-OKG-store ceiling is
still untouched for either tier.

### The delta, stated plainly

`main` landed ADR-0024 decision 3 (PR #4296, squash `54bba9f7`): an undeclared
tier is now DERIVED from the agent's KIND, and assistant-kind resolves L0. When
this PR was written, an assistant with no `tier =` resolved L1 and this PR
changed nothing for any shipped persona. That is no longer true.

ADR-0024's "Capability grants: zero observable delta" analysis was written
while #4170/#4172/#4173 were still open. **It holds for the other two and NOT
for this one.** #4173's and #4170's tools are gated by NAME, and no bundled
assistant's `[tools].allow` matches one — I extended the ADR's own roster guard
in both PRs to pin that mechanically. But `git_log` and `git_status` **are** in
the shipped `izzie` / `personal-assistant` allow-lists. So those personas go
from a single-tenant git surface to a cross-project one.

Concretely: `izzie` ingests untrusted content (Gmail/Drive/Calendar) and can
now be asked to `git_log` any *registered, active* sibling project. It is
read-only for those personas (no git write tool is in their allow-list) and
bounded by the operator's own `projects.json` — but it is a new
information-disclosure surface, and it should be an owner-reviewed fact rather
than a side effect of two PRs landing in sequence.

I have **not** narrowed the design to avoid it — that would be inventing an
unratified rule, and the ADR explicitly contemplated this PR landing. Instead
it is pinned by a new test,
`shipped_assistant_kind_gains_bounded_cross_project_git_reach`, whose doc
comment says exactly the above, asserts the widening IS cross-project, and
asserts it still refuses anything absent from the registry. If someone later
decides izzie should not have this, that test is where the decision surfaces.

**If you want that reach withheld from the Gmail/Drive-facing personas, say so
and it is a follow-up issue** — most likely dropping `git_log`/`git_status`
from those two allow-lists, which needs no code change.

### Other test changes (widening coverage; nothing deleted, ignored, or narrowed)

- `indeterminate_tier_resolves_to_single_tenant_scope` →
  `unrecognized_tier_declaration_resolves_to_single_tenant_scope`. The old name
  conflated two cases #4296 split apart. It now pins the case that would
  actually be an escalation — a declared-but-unrecognized `tier =` must fail
  closed to L1 and must NOT fall through to the kind derivation — and checks it
  across **both** roles (assistant and a sub-agent kind), 6 cases where there
  were 5.
- New `undeclared_tier_follows_the_kind_derivation`: assistant-kind derives L0
  and reaches the registered active sibling *and still refuses an unregistered
  one*; `engineer`/`researcher`/`planner` stay L1 and single-tenant.

### Conflicts and how they were resolved

- `agents/mod.rs` — both sides added a module declaration at the same line
  (`cross_project` here, ADR-0024's `delegation` on main). Both kept.
- `ctrl/pm_task/dispatch/persona.rs` — this PR REPLACES the plain `git_tools`
  registration with `register_scoped_git_tools`, while `main` had just gained
  #4221's hoisted `cwd` binding at the same lines. Took this PR's side: the
  scoped registration supersedes the plain one, and the `cwd` binding it drops
  had no other reader (the two later `cwd` bindings are separate scopes).
  #4221's `gh_tools::register` call sits outside the conflict and is preserved
  — all three grants (#4171 session-state, #4173 exec, #4170 gh, plus this
  one's scoped git) coexist on that path.
- Three stale `Test:` pointers to the renamed test were updated in
  `cross_project/mod.rs`.

### Gates (this head)

| Gate | Result |
|---|---|
| `cargo test -p trusty-agents --lib` | **3108 passed, 1 failed**, 11 ignored — see below |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | exit 0 |
| `scripts/check_line_cap.sh` | 7 allowlisted, 0 violations — OK |
| `scripts/check_test_pointers.sh` | 0 dangling pointers — OK |

The single failure is `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json`,
an environmental `uv` temp-cache race under parallel test threads
(`ModuleNotFoundError: No module named 'python'` from a `.cache/uv/.tmp…` path
that no longer exists). It passes in isolation —
`cargo test -p trusty-agents --lib tools::python_skill` → `8 passed; 0 failed`
— and it passed in the full run two commits ago. Unrelated to this change,
which touches no Python path.

CHANGELOG entry added under `## [Unreleased]` → `### Added`.

Issue #4186, opened 2026-07-27
[PMINSTR-03] Re-source the existing instruction build from the sectioned JSON


bobmatnyc:
**What:** Point the existing instruction build at the sectioned JSON. The composer (`instruction_pipeline.rs` / `resolve_pm_prompt`) stops reading N separate `include_str!` markdown assets and reads one sectioned JSON instead. Composition order, output, and delivery all stay as they are.

**Why:** Source swap, not a new build system. Mechanism change and content change must never mix — neither can hide a regression in the other.

**Acceptance:**
- The composed output is **byte-identical** before and after the swap — **given the same project manifest and the same agent roster**. Agent Delegation is a dynamic section, so output legitimately varies across projects: this is determinism-under-fixed-inputs, not a single golden file. Pin the manifest and roster in the test fixture and assert exact equality.
- No content change of any kind lands in this child. Core reorganization is #4193; the roster's 8→42 change is #4192.
- System-prompt-addendum delivery path unchanged; no new pipeline.

Parent: #4183

bobmatnyc:
## Acceptance criterion needs an explicit choice: byte-identity vs fixing #4196

Raised by the #4223 review of the #4184 schema and recorded here so it is a decision rather than a mid-implementation discovery. Also documented in code on `InstructionPackage::validate_roster`.

The #4184 schema makes roster **consumption** structurally mandatory: a package that never consumes `agent-roster`, or that marks it `optional`, fails validation, and a required generator supplying nothing is a hard composition error. That guard is airtight at the composition step.

But #4196's root cause was upstream of composition. `instruction_overrides::resolve_pm_prompt` still builds the launch prompt from the **static** `AGENT_DELEGATION` asset, while the computed roster lands in `PipelineOutput::merged` — which the launch path never reads. So this ticket must choose one of:

- **pass the real computed roster** — fixes #4196, but the output then *cannot* be byte-identical to today's `resolve_pm_prompt`; or
- **pass the static `AGENT_DELEGATION.md` text** as `agent_roster` — bytes match, but #4196 survives into the new format.

Neither is wrong, but "byte-identical to today's output" and "fixes #4196" are **mutually exclusive**. Suggested acceptance criterion:

> byte-identical **across runs for the same manifest and roster** (which the #4184 schema fully guarantees — `compose` is a pure function, ordering comes from `Vec` and never a map), **plus an explicitly reviewed diff against today's bytes** — not strict equality with the legacy output.

[...]

Issue #4185, opened 2026-07-27
[PMINSTR-02] Author the five section sources (+ absorbed BASE_PM floor)


bobmatnyc:
**What:** Author the five section sources against the schema from #4184, as a **faithful lift** — same content, new container. No reorganization.
- **Core** ← `PM_INSTRUCTIONS.md` (466 ln, 27 `##` headings). Lifted as-is; it stays a grab-bag for now. Reorganization is #4193, after the swap.
- **Agent Delegation** — **NOT a lift of the 63-line `AGENT_DELEGATION.md`.** This section is **dynamically constructed at composition time from the project manifest**, as before, and the manifest is itself mutable by agents and tools. What gets authored here is the section's *shape and generation contract*, not static text. **Depends on #4069** — the roster wiring must be live first or this section cannot be correct.
- **Workflow** ← `WORKFLOW.md` (151 ln, clean lift).
- **Memory** ← `BASE_PM.md:74-79`. A `FILE_MEMORY` override slot exists (`instruction_overrides.rs:40`) but there is **no bundled MEMORY.md** — author from scratch.
- **Search** ← `BASE_PM.md:80-93`. Does not exist as a section today.
Plus the absorbed BASE_PM floor sections.

**Why:** The JSON is inert without content. Keeping this a faithful lift is what lets #4186 assert byte-identical output.

**Acceptance:** All sections authored and tier-tagged; static sections are content-identical to today's assets; Agent Delegation is generated, not hardcoded.

Parent: #4183

bobmatnyc:
**Content requirement (owner decision, 2026-07-28):** the composed instructions must teach a sprint-then-harden workflow, not a blended one.

> "We should sprint to a target (feature complete on a local version), then test/fix carefully. The slow feature release means we have too many things in flight."

Two phases:
1. **SPRINT** — drive to feature-complete on a local version. Testing/CI used judiciously: targeted tests while developing, no CI iteration loops, no critic round on narrow changes.
2. **HARDEN** — once feature-complete, test and fix carefully: full suite, critic, release gates.

Causal claim to preserve: slow feature release *causes* too many things in flight — it's not a separate problem. Shortening time-to-land is the fix; managing WIP count directly (caps, purges) treats the symptom.

Derived rules to state explicitly:
[...]

PR #4223, opened 2026-07-28
feat(trusty-mpm): sectioned-JSON instruction package schema (#4184)

### LINKED ISSUE #4184: [PMINSTR-01] Define the instruction section taxonomy + JSON schema
**What:** Define the instruction section taxonomy and its JSON schema. Five sections (Core, Agent Delegation, Workflow, Memory, Search) plus the absorbed BASE_PM floor sections (Identity, Non-Overridable Rules, Framework-Guaranteed Conventions). Each section carries a `customization_tier`: `fixed | project | user`. Schema + ~150 LOC of types.

**Why:** Every other child in #4183 consumes these types. Nothing else can start until the shape is fixed.

DOC-59 §4a:251-280 is a usable input — it is the one part of the spec that matches the current design.

**Acceptance:**
- JSON schema committed, with the five sections and the tier axis.
- Rust types deserialize it; round-trip test.
- Documented placement for each absorbed BASE_PM block.
- Blocks #4183 children 2-7.

Parent: #4183

bobmatnyc:
Defines the instruction section taxonomy and its JSON schema — the shape every other child of epic #4183 consumes, and the thing that blocks all of them.

Scope is schema, Rust types, serde, validation, and unit tests. **No instruction content is authored (#4185) and no build is re-sourced (#4186)** — nothing in the session-launch path composes from this yet. `instruction_pipeline.rs` and `instruction_overrides.rs` are untouched.

## The shape

A package has two arrays:

- **`sections`** — the closed eight-member taxonomy carrying the `fixed | project | user` customization tier. Declaration order is fixed and canonical and has **no** effect on composed bytes.
- **`blocks`** — the ordered composition stream. Output is `blocks` in array order and nothing else.

Splitting identity from position is the load-bearing decision. A section may own **several non-contiguous blocks**, so `memory` can be lifted out of the middle of `PM_INSTRUCTIONS.md` and still emit at exactly the offset it occupies today. A one-section-one-position model would have made a faithful, byte-identical lift impossible for #4185/#4186, because `memory` and `search` are drawn from two different legacy files at two different depths.

## How this guarantees byte-identical composition

The acceptance test in #4186 is that the composed prompt is byte-identical given the same manifest and roster. Four properties get it there:

1. **Order comes from `Vec`, never from a map.** There is no `HashMap` anywhere in the type graph; the only ordered containers are `Vec`, and the only set is a `BTreeSet` used for a coverage check that never reaches output.
2. **Every boundary is an explicit literal.** `Join::Rule` (`"\n\n---\n\n"`, asserted equal to the existing `SECTION_SEPARATOR`), `Blank` (`"\n\n"`), `None` (`""`), or `Literal(verbatim)`. Whitespace is declared, not inferred, and `trailing_newline` pins the tail byte — `false` reproduces today's `resolve_pm_prompt` (no trailing newline), `true` a file-shaped artifact.
3. **`compose` is pure.** A total function of `(package, inputs)`: no clock, env, filesystem or randomness. `compose_is_byte_identical_across_repeats_and_a_json_round_trip` asserts repeats match and that a serialize/deserialize cycle does not change a single output byte.
4. **Dropping content requires explicit permission.** Only a `generated` block may be `optional`. Anything else that would render empty is a hard `CompositionError`.

## Regression shapes this makes hard to reintroduce

**#4196** — the delivered prompt named 8 agents because the computed 42-agent roster was discarded at the final composition step. Three independent structural guards now:

- `CompositionInputs.agent_roster` is a required, non-`Option` field, so composition cannot be called without producing a roster;
- `validate()` rejects a package that never consumes `Generator::AgentRoster` (`RosterNotConsumed`) or that marks the roster block `optional` (`OptionalRoster`);
- a required generator supplying nothing is `MissingGeneratedInput`, a hard error — the old failure produced a prompt that looked fine.

A fourth, weaker guard generalises it: `SectionWithoutBlocks` rejects any declared section that contributes nothing, which is the same "the taxonomy claims it ships, the output does not contain it" shape.

**#4194** — the non-overridable floor ordered an `index_id` that 404s. Content correctness is not this ticket, but the floor's semantics are now representable and enforced per section rather than per file: `SectionId::is_floor()`, `FloorNotFixed` (a floor section must be tier `fixed`), and `OverridableAfterFloor` (nothing overridable may follow the floor). `CustomizationTier::permits(OverrideTier)` encodes the tier matrix — notably `project` excludes user-tier override, so a personal `~/.trusty-mpm` config can never leak into a shared project.

Also: `deny_unknown_fields` throughout, so a typo'd key is a loud parse error instead of a silently dropped instruction block.

## Documented placement of each absorbed BASE_PM block

| `BASE_PM.md` block | package section |
|---|---|
| `## Identity` | `identity` |
| `## Non-Overridable Rules` | `non-overridable-rules` |
| `## Customizing PM Behavior` | `non-overridable-rules` |
| `## Trusty Tool Priority (Non-Overridable)` | `non-overridable-rules` |
| `## Framework-Guaranteed Conventions (Non-Overridable)` | `framework-guaranteed-conventions` |

The tool-priority block stays whole in the fixed floor deliberately: the *mandate* to reach for memory and code search before grep must remain non-overridable, even though the memory/search *guidance* sections a project may tune are tier `project`.

Note the canonical section list is a **declaration** order, not an imposed output order — `validate()` constrains `blocks` only with `OverridableAfterFloor`. That deliberate looseness is what leaves #4186 free to order blocks so the output matches today's bytes exactly.

## Deliberately not enforced here

No override *engine*. The tier axis ships as a pure predicate (`CustomizationTier::permits`) rather than a merge algorithm, because resolving overrides belongs to the later children. DOC-59's P1/P2 priority axis is absent — epic #4183 supersedes it with the five-section model.

## Verification

- `cargo test -p trusty-mpm --lib instruction_package` → `35 passed; 0 failed`
- `cargo test -p trusty-mpm` → `3769 passed; 0 failed` + `1257 passed; 0 failed` and 37 further green targets
- `cargo clippy --workspace --all-targets` (CI's exclude list) → clean under `-D warnings`
- `cargo fmt --check` → exit 0
- `scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`
- full `cargo test --workspace` (CI's exclude list) → every `trusty-mpm` target green

Three failures on the local dev machine are outside `trusty-mpm` and environmental — `trusty-agents/tests/api_e2e.rs` (asserts `/api/tasks` starts empty against a machine with a populated task store; also flaked on a 5s health timeout), `trusty-common` `catchup::json` and `trusty-search` `root_hijack_tests` (both pass in isolation, fail only under parallel load). None of the three crates depends on `trusty-mpm`, and this diff touches only `crates/trusty-mpm`, so it cannot reach them.

Closes #4184

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,6 +7,19 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Added
+
+- `core::instruction_package`: the sectioned-JSON instruction-package schema
+  ([#4184](https://github.com/bobmatnyc/trusty-tools/issues/4184), epic #4183) —
+  the eight-section taxonomy (Core, Memory, Search, Workflow, Agent Delegation
+  plus the absorbed BASE_PM floor sections Identity, Non-Overridable Rules,
+  Framework-Guaranteed Conventions), the `fixed | project | user`
+  customization-tier axis, an ordered block stream with explicit join literals,
+  structural validation, and a pure deterministic `compose`. The JSON Schema
+  ships as `assets/instructions/instruction-package.schema.json`. Types and
+  validation only — no instruction content is authored (#4185) and no build is
+  re-sourced (#4186); nothing in the session-launch path composes from it yet.
+
 ### Fixed
 
 - Worktree discovery is derived from `git worktree list --porcelain` instead of
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -19,6 +19,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   ships as `assets/instructions/instruction-package.schema.json`. Types and
   validation only — no instruction content is authored (#4185) and no build is
   re-sourced (#4186); nothing in the session-launch path composes from it yet.
+  Every object in the package's deserialization path rejects unknown keys, and
+  the error names the offending key. Strictness is deliberate over
+  forward compatibility here: `validate` gates on `schema_version` before
+  inspecting any field, so lenient field handling could only ever matter for a
+  change that added a field while leaving the version at 1 — precisely the case
+  that must not be tolerated, since an older binary would then compose a prompt
+  missing the new field's effect and report success. Any field addition that can
+  change composed bytes must bump `SCHEMA_VERSION`. Floor blocks are
+  deliberately *not* constrained to canonical section order: `BASE_PM.md` places
+  `## Trusty Tool Priority` after `## Framework-Guaranteed Conventions`, so
+  requiring non-decreasing order would reject a faithful byte-identical lift of
+  that asset.
 
 ### Fixed
 
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -26,11 +26,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   change that added a field while leaving the version at 1 — precisely the case
   that must not be tolerated, since an older binary would then compose a prompt
   missing the new field's effect and report success. Any field addition that can
-  change composed bytes must bump `SCHEMA_VERSION`. Floor blocks are
+  change composed bytes must bump `SCHEMA_VERSION`. That version is now checked
+  *before* the package is deserialized, so a package from a later schema reports
+  the actionable "unsupported schema_version" rather than blaming one of its own
+  perfectly valid new keys. Floor blocks are
   deliberately *not* constrained to canonical section order: `BASE_PM.md` places
   `## Trusty Tool Priority` after `## Framework-Guaranteed Conventions`, so
   requiring non-decreasing order would reject a faithful byte-identical lift of
-  that asset.
+  that asset. A declared section must own at least one *non-optional* block.
+  Owning blocks is not the same as emitting: a section covered only by
+  `optional` blocks passed validation and still composed to nothing,
+  recreating the very silent-missing-section shape the coverage check exists
+  to prevent.
 
 ### Fixed
 
--- a/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
+++ b/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
@@ -1,0 +1,234 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/bobmatnyc/trusty-tools/schemas/instruction-package-v1.json",
+  "title": "trusty-mpm instruction package",
+  "description": "A PM instruction package: a section taxonomy carrying the customization-tier axis, plus an ORDERED block stream that composes to the PM system prompt. Ordering is array order — never map iteration order — so composition is byte-deterministic (SPEC-PMINSTR-01 / DOC-59 §4a, epic #4183, schema ticket #4184).",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["schema_version", "package_id", "sections", "blocks"],
+  "properties": {
+    "schema_version": {
+      "description": "Schema major version. A package whose version this build does not implement is rejected outright — never partially interpreted.",
+      "const": 1
+    },
+    "package_id": {
+      "description": "Stable identifier for this package (e.g. `trusty-mpm/pm`). Identity only; it never affects composed bytes.",
+      "type": "string",
+      "minLength": 1
+    },
+    "description": {
+      "type": "string"
+    },
+    "trailing_newline": {
+      "description": "Append exactly one '\\n' after the last emitted block. `false` reproduces today's `resolve_pm_prompt` output; `true` reproduces a file-shaped artifact.",
+      "type": "boolean",
+      "default": false
+    },
+    "sections": {
+      "description": "The section taxonomy. Declaration order MUST be the canonical order (identity, core, memory, search, workflow, agent-delegation, non-overridable-rules, framework-guaranteed-conventions) and every canonical section MUST be declared exactly once. Declaration order does not affect composed bytes — `blocks` alone determines output order.",
+      "type": "array",
+      "items": { "$ref": "#/$defs/section" },
+      "minItems": 8,
+      "maxItems": 8
+    },
+    "blocks": {
+      "description": "The ordered composition stream. Emitted strictly in array order. A section may own several non-contiguous blocks, which is what lets a section be lifted out of the middle of a legacy asset without moving a single byte.",
+      "type": "array",
+      "items": { "$ref": "#/$defs/block" },
+      "minItems": 1
+    }
+  },
+  "$defs": {
+    "section-id": {
+      "description": "The closed section taxonomy, in canonical declaration order. `identity`, `non-overridable-rules` and `framework-guaranteed-conventions` are the absorbed BASE_PM floor blocks and are always tier `fixed`.",
+      "type": "string",
+      "enum": [
+        "identity",
+        "core",
+        "memory",
+        "search",
+        "workflow",
+        "agent-delegation",
+        "non-overridable-rules",
+        "framework-guaranteed-conventions"
+      ]
+    },
+    "customization-tier": {
+      "description": "Who may override this section, by increasing permissiveness. `fixed`: nobody, at any tier (the framework floor). `project`: a project-tier override only — deliberately NOT a user-tier one, so a personal machine config can never silently leak into a shared project. `user`: both project- and user-tier, project winning on collision (most-specific-wins).",
+      "type": "string",
+      "enum": ["fixed", "project", "user"]
+    },
+    "generator": {
+      "description": "A named composition-time input the host must supply. Generated content is never authored in the package — it is computed per project/session and passed in.",
+      "type": "string",
+      "enum": ["agent-roster", "stack-profile", "project-addendum"]
+    },
+    "join": {
+      "description": "The literal bytes inserted BEFORE this block when it is not the first emitted block. `rule` = \"\\n\\n---\\n\\n\", `blank` = \"\\n\\n\", `none` = \"\", `{\"literal\": \"...\"}` = verbatim. Explicit joins are what make a faithful, byte-identical lift possible: blocks split out of one legacy document rejoin with `blank`, while distinct legacy assets rejoin with `rule`.",
+      "oneOf": [
+        { "type": "string", "enum": ["rule", "blank", "none"] },
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["literal"],
+          "properties": { "literal": { "type": "string" } }
+        }
+      ]
+    },
+    "body": {
+      "description": "Where a block's markdown comes from. `text` is authored in the package; `generated` is supplied at composition time by the named generator.",
+      "oneOf": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["kind", "text"],
+          "properties": {
+            "kind": { "const": "text" },
+            "text": { "type": "string", "minLength": 1 }
+          }
+        },
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["kind", "generator"],
+          "properties": {
+            "kind": { "const": "generated" },
+            "generator": { "$ref": "#/$defs/generator" }
+          }
+        }
+      ]
+    },
+    "section": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["id", "title", "customization_tier"],
+      "properties": {
+        "id": { "$ref": "#/$defs/section-id" },
+        "title": { "type": "string", "minLength": 1 },
+        "customization_tier": { "$ref": "#/$defs/customization-tier" },
+        "description": { "type": "string" }
+      }
+    },
+    "block": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["section", "body"],
+      "properties": {
+        "section": { "$ref": "#/$defs/section-id" },
+        "body": { "$ref": "#/$defs/body" },
+        "join_before": { "$ref": "#/$defs/join", "default": "rule" },
+        "optional": {
+          "description": "Only a `generated` block may be optional. An optional block is dropped when its generator supplies no content; a non-optional block whose content is missing or empty is a hard composition error, never a silent drop (issue #4196 — the computed 42-agent roster was discarded at the final composition step and the prompt shipped naming 8).",
+          "type": "boolean",
+          "default": false
+        }
+      }
+    }
+  },
+  "examples": [
+    {
+      "schema_version": 1,
+      "package_id": "trusty-mpm/example",
+      "description": "Minimal structurally-valid package. Content authoring is #4185; this example exists so the schema's own shape is executable and testable.",
+      "trailing_newline": false,
+      "sections": [
+        {
+          "id": "identity",
+          "title": "Identity",
+          "customization_tier": "fixed",
+          "description": "Absorbed BASE_PM `## Identity`."
+        },
+        {
+          "id": "core",
+          "title": "Core PM Instructions",
+          "customization_tier": "project"
+        },
+        {
+          "id": "memory",
+          "title": "Memory",
+          "customization_tier": "project"
+        },
+        {
+          "id": "search",
+          "title": "Search",
+          "customization_tier": "project"
+        },
+        {
+          "id": "workflow",
+          "title": "Workflow",
+          "customization_tier": "project"
+        },
+        {
+          "id": "agent-delegation",
+          "title": "Agent Delegation",
+          "customization_tier": "project",
+          "description": "Dynamic: built from the deployed-agent roster at composition time."
+        },
+        {
+          "id": "non-overridable-rules",
+          "title": "Non-Overridable Rules",
+          "customization_tier": "fixed",
+          "description": "Absorbed BASE_PM `## Non-Overridable Rules`, `## Customizing PM Behavior`, `## Trusty Tool Priority (Non-Overridable)`."
+        },
+        {
+          "id": "framework-guaranteed-conventions",
+          "title": "Framework-Guaranteed Conventions",
+          "customization_tier": "fixed",
+          "description": "Absorbed BASE_PM `## Framework-Guaranteed Conventions (Non-Overridable)`."
+        }
+      ],
+      "blocks": [
+        {
+          "section": "core",
+          "body": { "kind": "text", "text": "# PM Agent -- Trusty MPM" }
+        },
+        {
+          "section": "memory",
+          "join_before": "blank",
+          "body": { "kind": "text", "text": "## Context-First Protocol" }
+        },
+        {
+          "section": "core",
+          "join_before": "blank",
+          "body": { "kind": "text", "text": "## Agent Routing" }
+        },
+        {
+          "section": "search",
+          "body": { "kind": "text", "text": "## Code/Architecture Search" }
+        },
+        {
+          "section": "core",
+          "body": { "kind": "generated", "generator": "stack-profile" },
+          "optional": true
+        },
+        {
+          "section": "workflow",
+          "body": { "kind": "text", "text": "# PM Workflow Configuration" }
+        },
+        {
+          "section": "agent-delegation",
+          "body": { "kind": "generated", "generator": "agent-roster" }
+        },
+        {
+          "section": "core",
+          "body": { "kind": "generated", "generator": "project-addendum" },
+          "optional": true
+        },
+        {
+          "section": "identity",
+          "body": { "kind": "text", "text": "## Identity" }
+        },
+        {
+          "section": "non-overridable-rules",
+          "join_before": "blank",
+          "body": { "kind": "text", "text": "## Non-Overridable Rules" }
+        },
+        {
+          "section": "framework-guaranteed-conventions",
+          "join_before": "blank",
+          "body": { "kind": "text", "text": "## Framework-Guaranteed Conventions" }
+        }
+      ]
+    }
+  ]
+}
--- a/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
+++ b/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
@@ -8,7 +8,7 @@
   "required": ["schema_version", "package_id", "sections", "blocks"],
   "properties": {
     "schema_version": {
-      "description": "Schema major version. A package whose version this build does not implement is rejected outright — never partially interpreted.",
+      "description": "Schema major version. A package whose version this build does not implement is rejected outright — never partially interpreted; the version is checked BEFORE any field, so a newer package reports a version error rather than blaming one of its own valid keys. EVOLUTION POLICY: every object here sets `additionalProperties: false`, so unknown keys are errors, not ignored. Any field addition that can change composed bytes MUST bump this version. Adding a field while leaving the version at 1 is the one case strictness cannot protect against — an older build would compose a prompt missing the new field's effect and report success.",
       "const": 1
     },
     "package_id": {
--- a/crates/trusty-mpm/src/core/instruction_package.rs
+++ b/crates/trusty-mpm/src/core/instruction_package.rs
@@ -721,16 +721,9 @@ impl InstructionPackage {
     /// output** as its acceptance gate; it is the strongest faithfulness check
     /// available and nothing here forces weakening it.
     ///
-    /// The one real divergence, which #4186 must decide deliberately: when NO
-    /// agents are deployed, `deployed_roster_section` returns `None` and today's
-    /// composer silently degrades to the bundled asset alone. Under this schema
-    /// that case is a hard [`CompositionError::MissingGeneratedInput`], because
-    /// the roster block may not be marked `optional` ([`ValidationError::OptionalRoster`]).
-    /// That is a deliberate strengthening — the silent degradation *is* #4069's
-    /// failure shape — but it means byte-identity holds for the normal
-    /// agents-deployed case and intentionally does not hold for the zero-agent
-    /// case. #4186 must either guarantee a non-empty roster at the call site or
-    /// record an explicit decision for zero-agent projects.
+    /// Scope — including the other override configurations `resolve_pm_prompt`
+    /// can emit, and which of them this schema can reproduce — is tracked on
+    /// #4186, not here.
     /// Test: `todays_delegation_section_shape_is_expressible`.
     fn validate_roster(&self) -> Result<(), ValidationError> {
         let mut seen = false;
--- a/crates/trusty-mpm/src/core/instruction_package.rs
+++ b/crates/trusty-mpm/src/core/instruction_package.rs
@@ -469,12 +469,6 @@ pub enum ValidationError {
         /// The owning section.
         section: SectionId,
     },
-    /// A declared section contributes no block.
-    #[error("section {section:?} is declared but contributes no block")]
-    SectionWithoutBlocks {
-        /// The silent section.
-        section: SectionId,
-    },
     /// No block consumes [`Generator::AgentRoster`].
     #[error(
         "no block consumes the agent roster; the computed roster must reach the \
@@ -486,6 +480,21 @@ pub enum ValidationError {
     OptionalRoster {
         /// Index into `blocks`.
         index: usize,
+    },
+    /// A declared section contributes no block.
+    #[error("section {section:?} is declared but contributes no block")]
+    SectionWithoutBlocks {
+        /// The silent section.
+        section: SectionId,
+    },
+    /// A declared section's blocks are all `optional`, so it may emit nothing.
+    #[error(
+        "section {section:?} contributes only `optional` blocks, so it can emit nothing \
+         while still validating; give it at least one non-optional block"
+    )]
+    SectionOnlyOptionalBlocks {
+        /// The section that is not guaranteed to emit.
+        section: SectionId,
     },
     /// A non-floor block follows the first floor block.
     #[error(
@@ -550,10 +559,35 @@ impl InstructionPackage {
     /// Why: unknown fields are rejected (`deny_unknown_fields` throughout) so a
     /// typo in a key becomes a parse error rather than silently dropped
     /// instruction content.
-    /// What: `serde_json::from_str`, without validation — call [`Self::validate`]
-    /// or [`Self::compose`] for that.
-    /// Test: `schema_example_deserializes_validates_and_round_trips`, `rejects_unknown_fields_at_every_level_and_names_the_key`.
+    ///
+    /// The `schema_version` probe runs FIRST, before the package is
+    /// deserialized, because strictness and version-gating otherwise collide:
+    /// a package written against a later schema almost certainly carries a field
+    /// this build does not know, so plain deserialization would report `unknown
+    /// field \`whatever\`` — blaming a key that is perfectly valid in its own
+    /// schema — instead of the actionable "you need a newer trusty-mpm". The
+    /// probe deliberately ignores unknown fields; it reads nothing but the
+    /// version.
+    ///
+    /// What: version probe, then `serde_json::from_str`. No structural
+    /// validation — call [`Self::validate`] or [`Self::compose`] for that.
+    /// Test: `schema_example_deserializes_validates_and_round_trips`,
+    /// `rejects_unknown_fields_at_every_level_and_names_the_key`,
+    /// `later_schema_version_is_reported_as_a_version_error_not_a_field_error`.
     pub fn from_json(raw: &str) -> Result<Self, serde_json::Error> {
+        /// Reads only `schema_version`, tolerating every other key by design.
+        #[derive(Deserialize)]
+        struct VersionProbe {
+            schema_version: u32,
+        }
+
+        let probe: VersionProbe = serde_json::from_str(raw)?;
+        if probe.schema_version != SCHEMA_VERSION {
+            return Err(serde::de::Error::custom(format!(
+                "unsupported schema_version {}; this build implements {SCHEMA_VERSION}",
+                probe.schema_version
+            )));
+        }
         serde_json::from_str(raw)
     }
 
@@ -582,7 +616,9 @@ impl InstructionPackage {
     /// contract.
     /// What: version, identity, canonical section set/order, floor tiers,
     /// non-empty block stream, per-block text/optional sanity, every section
-    /// contributing, the roster being consumed and non-optional, and nothing
+    /// contributing *and being guaranteed to emit* (owning blocks is not
+    /// enough — a section covered only by `optional` blocks composes to
+    /// nothing), the roster being consumed and non-optional, and nothing
     /// overridable following the floor.
     /// Test: one test per [`ValidationError`] variant.
     pub fn validate(&self) -> Result<(), ValidationError> {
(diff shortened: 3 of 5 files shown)


bobmatnyc (commented):
## Code Critic — Adversarial Review: WARN

Independent review of the schema **design**, not just the code, because #4184 blocks #4185/#4186/#4193/#4187-4191. Verified empirically: 35/35 tests pass in the real crate (`cargo test -p trusty-mpm --lib instruction_package`), plus 6 adversarial mutation tests I wrote against an isolated copy of the module.

**Verdict: WARN** — one HIGH finding, no CRITICAL. `--request-changes` is not available (same account authored the PR), so this is posted as a review comment.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/instruction_package.rs` | 278-280 | `BlockBody` has no `deny_unknown_fields`; a misplaced key inside `body` is silently dropped, silently changing composed bytes | add `deny_unknown_fields` to the `#[serde(...)]` attribute |
| MEDIUM | `crates/trusty-mpm/src/core/instruction_package.rs` | 636-650 | Floor blocks may appear in any internal order; canonical order is enforced only on the array that has zero effect on bytes | require floor blocks to be in canonical section order relative to each other |
| MEDIUM | `crates/trusty-mpm/src/core/instruction_package.rs` | 441-446 | `RosterNotConsumed` guards the composition step, but #4196's root cause is upstream — this forces a decision #4186 must make explicitly | write the choice into #4186's acceptance criterion now |
| LOW | `crates/trusty-mpm/src/core/instruction_package.rs` | 372-374, 688 | `trailing_newline: bool` + last-block `trim` means the tail is always `X` or `X\n` — 2+ trailing newlines are inexpressible | none needed; adjust the "pins the tail byte" claim |
| LOW | `crates/trusty-mpm/src/core/instruction_package.rs` | 186-209 | `CustomizationTier::permits` / `OverrideTier` have no production caller — the #4194 guard is declared, not yet enforced | none; wire in #4187-4191 |

---

### HIGH — `BlockBody` silently accepts unknown fields, and can silently change composed bytes

`instruction_package.rs:278-280`:

```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
#[serde(tag = "kind", rename_all = "kebab-case")]
pub enum BlockBody {
```

Every other type in the graph carries `deny_unknown_fields` (`InstructionBlock` :304, `InstructionSection` :341, `InstructionPackage` :363). `BlockBody` does not. Verified empirically:

```
{"section":"core","body":{"kind":"text","text":"HI","optional":true,"generator":"agent-roster"}}
  => Ok(Text { text: "HI" })      // `optional` and `generator` silently dropped
{"id":"core","title":"T","customization_tier":"project","tier":"user"}
  => Err("unknown field `tier`, expected one of `id`, `title`, ...")   // correctly rejected
```

Three reasons this is HIGH rather than a nit:

1. **It can silently change composed bytes.** Author writes `"join_before": "blank"` inside `body` instead of on the block — a plausible slip, since `join_before` and `optional` are documented adjacent to `body`. The key is dropped, the block falls back to `Join::Rule` default (:247-248), and a spurious `\n\n---\n\n` horizontal rule appears in the PM system prompt. That is precisely the byte drift #4186's acceptance test exists to prevent, introduced by a package that parses without complaint.
2. **The Rust loader is looser than the schema you shipped.** `instruction-package.schema.json:81-99` sets `additionalProperties: false` on *both* `body` variants. So a package validates in an editor and behaves differently in `trusty-mpm` — the exact scenario `schema_enums_match_rust_enums` (`instruction_package_tests.rs:135-138`) declares itself the gate against. That drift test covers the three enums, not `additionalProperties` parity.
3. **It contradicts the module's own stated contract**, `instruction_package.rs:513-514`: *"unknown fields are rejected (`deny_unknown_fields` throughout) so a typo in a key becomes a parse error rather than silently dropped instruction content."* Not true at the body level. `rejects_unknown_fields` (`:188-200`) only injects `joinbefore` at the `InstructionBlock` level, so it never exercises this.

**Fix (verified working):** add `deny_unknown_fields` to the `BlockBody` serde attribute. Serde supports it on internally-tagged enums; I applied it to a copy and all 41 tests (35 existing + 6 adversarial) still pass with no other change. Please also extend `rejects_unknown_fields` to inject a stray key inside `blocks[0].body`, so this cannot regress.

### MEDIUM — the floor's internal order is unconstrained, and strictness sits on the wrong array

`validate_floor_is_last` (`:636-650`) checks only that nothing *non-floor* follows the first floor block. Floor blocks may appear among themselves in any order. Verified:

```
blocks: [Core, Memory, Search, Workflow, AgentRoster,
         FrameworkGuaranteedConventions, Identity, NonOverridableRules]
validate() => Ok(())
compose() => "C\n\n---\n\nM\n\n---\n\nS\n\n---\n\nW\n\n---\n\nR\n\n---\n\nCONV\n\n---\n\nID\n\n---\n\nRULES"
```

`## Identity` emitted *after* `## Framework-Guaranteed Conventions`. Structurally valid, semantically incoherent.

The asymmetry is the point: `SectionsNotCanonical` (`:561-564`) enforces exact canonical order on `sections`, an array the module doc explicitly says has **no effect on composed bytes** (`:20-21`). `blocks` — the only array that determines bytes — gets no ordering constraint at all beyond the floor boundary. The strict check is on the inert array; the load-bearing one is free.

I accept the deliberate looseness for the five *content* sections — #4186 genuinely needs to interleave `core`/`memory`/`search` to lift bytes out of the middle of `PM_INSTRUCTIONS.md`, and I confirmed that works (below). But the floor is different: it is by construction the contiguous tail, so requiring floor blocks to run in `SectionId::CANONICAL` order relative to each other costs #4186 nothing and closes the one ordering error `validate()` could catch but doesn't.

Related, lower confidence so not separately filed: a section's content blocks can also be scattered out of narrative order (`Workflow` part-2 emitted before part-1) and still validate. That is the acknowledged trade-off, and #4186's byte comparison catches it — but it does mean `validate()` is not a safety net for #4185 authoring, only for #4186 mechanics. Worth saying out loud in the #4185 brief.

### MEDIUM — the roster guard is airtight at composition, but #4196's root cause is upstream

The #4196 guard itself is **solid**. I tried to construct the bug and could not:

```
agent_roster: ""       => Err(MissingGeneratedInput { index: 4, generator: AgentRoster })
agent_roster: "   \n " => Err(MissingGeneratedInput { index: 4, generator: AgentRoster })
agent_roster: "ROSTER" => Ok(roster present in output)
```

Required non-`Option` field (`:503`) + `RosterNotConsumed` (`:626`) + `OptionalRoster` (`:618`) + hard `MissingGeneratedInput` (`:697`) leaves no path to a validating package that drops the roster.

But #4196 did not happen at the composition step's *input* — it happened because the computed roster never got that far. Today `resolve_pm_prompt` (`instruction_overrides.rs:141-188`) builds the launch prompt from the **static** `AGENT_DELEGATION` asset (`:166-167`); `generate_authority` output (`instruction_pipeline.rs:345`) lands in `PipelineOutput::merged`, which the launch path never reads — documented at `instruction_pipeline.rs:11-20`.

Consequence for #4186, which should be settled now rather than discovered mid-implementation: this schema *mandates* a `Generated{AgentRoster}` block, so #4186 must choose between

- passing the real computed roster — fixes #4196, but the output **cannot** be byte-identical to today's `resolve_pm_prompt`; or
- passing the static `AGENT_DELEGATION.md` text as `agent_roster` — bytes match, #4196 survives into the new format.

Neither is wrong, but "byte-identical to today's output" and "fixes #4196" are mutually exclusive here. #4186's acceptance criterion should read *byte-identical across runs for the same manifest and roster* (which this schema fully delivers) plus an explicitly reviewed diff against today's bytes — not strict equality with the legacy output.

### Refuted risks

**Byte-identical composition is genuinely guaranteed.** No `HashMap` anywhere in the type graph, including nested and derived types. The single map is `BTreeSet<SectionId>` at `:596`, and its iterator is never used — the loop iterates `SectionId::CANONICAL` (an array) and only calls `contains`, so no set iteration order can reach output. `compose` (`:672-714`) is a pure fold over `self.blocks` with no clock, env, fs, or randomness. Serialization goes through `serialize_struct` in declaration order, never a JSON map, so `to_json` is byte-stable (and `schema_example_deserializes_validates_and_round_trips` asserts it is a fixed point, `tests.rs:180-184`).

**And the determinism test binds — it is not vacuous.** I replaced the `blocks` iteration in `compose` with a `HashMap<usize, &InstructionBlock>` and re-ran:

```
test tests::compose_is_byte_identical_across_repeats_and_a_json_round_trip ... FAILED
  panicked at instruction_package_tests.rs:634: assertion `left == right` failed:
  compose must be a pure function
```

The same-function leg would be weak alone, but the JSON-round-trip leg (`:636-642`) compares a freshly-deserialized package against the original, and joins are separately pinned by an exact golden string in `composes_blocks_in_array_order_with_declared_joins` (`:510-524`) and by `join_literals_are_exact` (`:282-287`), including `Join::Rule == SECTION_SEPARATOR`. Spot-checked six tests for vacuity; all bind under mutation. Weakest is `whitespace_only_generator_output_counts_as_absent` (`:557-567`) — it asserts absence of the blank string but not that the join was also dropped — though it does catch removal of the `str::trim` at `:688`.

**Joins are not limited to four literals.** `Join::Literal` (`:254`) is an arbitrary-bytes escape hatch, which resolves the sharpest expressibility worry. Verified that a split at a markdown hard-break survives the per-block `trim`:

```
"line one  " + Join::Literal("  \n") + "line two"  =>  "line one  \nline two"
```

So a legacy asset can be split at an arbitrary byte offset and reassembled exactly — trailing significant whitespace is recoverable by moving it into the join. Only the very first block's leading and the very last block's trailing whitespace are unrecoverable, and both current targets are trimmed anyway (`instruction_overrides.rs:198-205`).

**The two-array model expresses today's prompt.** `resolve_pm_prompt`'s section list (`:169-185`) is `[PM_INSTRUCTIONS, stack, (memory override), WORKFLOW, AGENT_DELEGATION, (INSTRUCTIONS.md), BASE_PM]` joined by `SECTION_SEPARATOR`. The schema's own example (`schema.json:180-231`) reproduces that shape exactly, with `memory`/`search` lifted from the middle of `PM_INSTRUCTIONS` and rejoined with `blank`. The claim that separating section identity from block position is load-bearing checks out against the real assets — a one-section-one-position model would have to relocate the Context-First Protocol out of the middle of `PM_INSTRUCTIONS.md` and change bytes. I found no structure in the current prompt this schema cannot reproduce.

**The #4194 tier matrix is correct.** `permits` (`:201-208`): `Fixed` permits neither tier; `Project` permits `OverrideTier::Project` only and explicitly **not** `User` (`:205`); `User` permits both. A user-tier config cannot reach a project-tier section. `FloorNotFixed` (`:566-573`) plus `is_floor()` (`:146-153`) make floor membership per-section and enforced.

**Scope discipline is clean.** Five files, +1632/-0. `core/mod.rs` adds only `pub mod instruction_package;`. `git grep instruction_package` on the branch returns exactly one non-self reference (that `mod` line). No override merge engine, no authored content, no build re-sourcing, no launch-path caller. `instruction_pipeline.rs` and `instruction_overrides.rs` are untouched. Nothing here pre-empts #4185 or #4186.

### Note (not blocking) — a fourth dynamic input the schema does not name

`output_style::inject_style_into_prompt` (`output_style.rs:299-305`) prepends `INJECTED_STYLE_HEADING` + style body + `SECTION_SEPARATOR` ahead of the entire prompt, conditionally on the installed Claude Code version. It is applied *outside* `resolve_pm_prompt`, in `build_system_prompt_for_with_style_and_native` (`session_launch/mod.rs:911-923`), so #4186 can keep it as an outer wrapper and the three-generator set stays sufficient. Flagging so it is a decision rather than a late discovery.

### Note — `make check` is genuinely unrunnable, but the diagnosis needs correcting

Confirmed the gate is broken, with a different cause than reported. `Makefile:46-49` runs `cargo test --workspace` and `cargo clippy --workspace --all-targets` with **no excludes**; CI uses `--exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui` (`ci.yml:164`, `:542`, `:564`, `:623`). The Makefile never got the same treatment.

But only **one** of the three crates actually fails, and it is not one of the two named:

```
cargo check -p trusty-mpm-gui       => Finished (ok)
cargo test  -p trusty-mpm-gui       => ok, 0 failed
cargo test  -p trusty-code-gui      => ok, 0 failed
cargo test  -p trusty-agents-ui --no-run
  error: proc macro panicked
    --> crates/trusty-agents/ui/src-tauri/src/main.rs:144:16
     |  .build(tauri::generate_context!())
     = help: message: The `frontendDist` configuration is set to `"../dist"`
             but this path doesn't exist
```

`trusty-mpm-gui` and `trusty-code-gui` both declare `"frontendDist": "ui/dist"` with the directory absent and compile fine anyway. Only `trusty-agents-ui` panics. Root cause: `crates/trusty-agents/ui/dist` is gitignored build output (`crates/trusty-agents/ui/.gitignore:2`). It exists in the main checkout, where someone has run the UI build — so `make check` may well pass there — and is absent in every **fresh git worktree**, which is where every delegated agent works. That is why agents keep independently working around it and why the failure looks intermittent.

Suggested fix, worth a separate ticket: add `--exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui` to `Makefile:47-48` so the documented gate matches CI, or have the target build `crates/trusty-agents/ui/dist` first.

1. Add `deny_unknown_fields` to `BlockBody` (`instruction_package.rs:279`) and extend `rejects_unknown_fields` to cover a stray key inside `blocks[0].body`.
2. Constrain floor blocks to canonical order relative to each other in `validate_floor_is_last` (`:636-650`), or document why scrambled floor order is acceptable.
3. Record the roster/byte-identity trade-off on #4186 before content authoring starts.

Items 2 and 3 are MEDIUM and may be tracked rather than fixed here. Only item 1 is HIGH.

---

bobmatnyc:
## VERDICT: WARN

Re-review of `81e674b6` (rebased onto `7c4f0564`), reviewed in an independent worktree off `refs/pull/4223/head`. Prior round was WARN (1 HIGH + 2 MEDIUM + LOWs).

**The prior HIGH is genuinely fixed.** **The prior MEDIUM 1 was wrong — I withdraw it.** One new HIGH: the MEDIUM 2 decision record is built on a premise that is false in this branch's own tree.

CI: **all checks passing** (Test 16m52s, Clippy 5m12s, MSRV 3m36s, doc-comment pointer lint, 500-line cap, SLD lint, BASE_PM floor check — 0 failing, 0 pending; "Notify on red main" correctly `skipping`).

---

## Withdrawal: prior MEDIUM 1 was wrong

I proposed requiring floor blocks to run in non-decreasing canonical section order and asserted it "costs #4186 nothing." I verified the refutation against the asset rather than against the engineer's table.

`crates/trusty-mpm/src/assets/instructions/BASE_PM.md` headings, in file order (lines 5, 16, 22, 51, 70):

```
## Identity                                           -> Identity                        (0)
## Non-Overridable Rules                              -> NonOverridableRules             (6)
## Customizing PM Behavior                            -> NonOverridableRules             (6)
## Framework-Guaranteed Conventions (Non-Overridable) -> FrameworkGuaranteedConventions  (7)
## Trusty Tool Priority (Non-Overridable)             -> NonOverridableRules             (6)
```

Indices are `SectionId::CANONICAL` (`instruction_package.rs:148-157`). The stream is `0, 6, 6, 7, 6`.

The refutation is **stronger than the engineer argued**, and holds independently of the enum's declaration order. The floor pattern is `A, A, B, A` — one section straddling another. No total order on two distinct sections admits a non-decreasing sequence there. Reordering `CANONICAL` to put `FrameworkGuaranteedConventions` at 6 just yields `0, 7, 7, 6, 7` — still inverted. The only escape would be remapping `## Trusty Tool Priority` into `FrameworkGuaranteedConventions`, and the asset's own words block that: `BASE_PM.md:53` scopes that heading to "**These three** conventions" (attribution footer, doc proportionality, ticket attribution), while `BASE_PM.md:35-37` names tool-priority separately as part of the never-overridable floor. The mapping is correct; the constraint I asked for is unsatisfiable.

I confirmed this by mutation rather than by reading the argument. Adding exactly the rule I proposed to `validate_floor_is_last`:

```
test core::instruction_package::tests::base_pm_floor_block_order_is_valid ... FAILED
BASE_PM.md's real floor order must remain expressible:
  OverridableAfterFloor { index: 12, section: NonOverridableRules, floor_index: 8 }
```

The prior MEDIUM 1 was incorrect. Not implementing it was right, and pinning it with `base_pm_floor_block_order_is_valid` is the right way to stop it being re-proposed.

---

## Findings

|---|---|---|---|---|
| HIGH | crates/trusty-mpm/src/core/instruction_package.rs | 650-663 | The `validate_roster` decision record (and the copy posted to #4186) asserts `resolve_pm_prompt` "still builds the launch prompt from the static `AGENT_DELEGATION` asset." That is false in this tree — `bundled_delegation` already folds in the live computed roster. The recorded conclusion ("byte-identical" and "fixes #4196" are mutually exclusive) does not follow, and it instructs #4186 to drop strict byte-equality against today's output. | Rewrite the doc block and the #4186 comment against current `instruction_overrides.rs`; keep byte-equality against today's `resolve_pm_prompt` as a live acceptance criterion. |
| MEDIUM | crates/trusty-mpm/src/core/instruction_package.rs | 633-638 | `SectionWithoutBlocks` is a static check, but a section whose only blocks are all `optional: true` passes it and can still emit nothing at compose — recreating the exact "declared but silent section" the check exists to prevent. | In `validate`, require each canonical section to be covered by at least one block that is not `optional`. |
| LOW | crates/trusty-mpm/src/core/instruction_package.rs | 40-41, 550 | "Every object in the deserialization path carries `deny_unknown_fields`" / "`deny_unknown_fields` throughout" overstates: `Join::Literal`'s object does not, and is closed by a different mechanism. | Say "every object except `Join`, which external tagging closes"; `instruction_package_tests.rs:299-303` already states this correctly. |
| LOW | crates/trusty-mpm/src/core/instruction_package.rs | 44-52 | The forward-compat argument promises a "loud `unsupported schema_version`", but deserialization runs before `validate`, so a v2 package carrying a new field actually fails with ``unknown field `x` `` and never reaches the version gate. | Note that the version gate is the *semantic* backstop and the parse error is what an author sees first. |
| LOW | crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json | 10-13 | The evolution policy ("any field addition that can change composed bytes bumps `SCHEMA_VERSION`") is in the Rust module docs and the changelog, but not in the shipped schema document — the artifact a schema author actually edits. | Add one sentence to the `schema_version` description. |

### HIGH — evidence

`crates/trusty-mpm/src/core/instruction_package.rs:650-653`:

> `instruction_overrides::resolve_pm_prompt` still builds the launch prompt from the static `AGENT_DELEGATION` asset, while the computed roster lands in `PipelineOutput::merged`, which the launch path never reads.

`crates/trusty-mpm/src/core/instruction_overrides.rs:254-260`, present on `origin/main` and untouched by this PR:

```rust
fn bundled_delegation(project_dir: &Path) -> String {
    let bundled = AGENT_DELEGATION.trim();
    match crate::core::delegation_authority::deployed_roster_section(project_dir) {
        Some(roster) => format!("{bundled}\n\n{ROSTER_PRECEDENCE_NOTE}\n\n{}", roster.trim()),
        None => bundled.to_string(),
    }
}
```

`resolve_pm_prompt` calls it at `instruction_overrides.rs:173-174`. The regression gate at `instruction_overrides.rs:312-319` says so explicitly: *"Against the pre-fix code this assertion fails: `resolve_pm_prompt` returned the static `AGENT_DELEGATION.md` verbatim."* Past tense. Issue #4069 is CLOSED and PR #4196 MERGED at 2026-07-28T01:28Z; the doc block was authored at 08:02Z the same day, so the claim was already stale when written.

The conclusion drawn from it is therefore wrong, and the counterexample is constructible. Today's delegation section is `AGENT_DELEGATION.trim()` + `"\n\n"` + `ROSTER_PRECEDENCE_NOTE` + `"\n\n"` + `roster.trim()`. That is expressible in this schema with no byte moved:

```
text(agent-delegation, AGENT_DELEGATION body)
text(agent-delegation, ROSTER_PRECEDENCE_NOTE)   join_before: blank
generated(agent-delegation, agent-roster)        join_before: blank
```

byte-identical **and** consuming the real computed roster. Byte-identity and roster delivery are not in tension.

Why HIGH rather than a doc nit: the artifact MEDIUM 2 produced *is* a decision record, its content is a factual claim about another module, and its recorded conclusion tells #4186 to replace strict byte-equality against today's output with "byte-identical across runs plus a reviewed diff." That weakens the single strongest check that the lift is faithful, on grounds that do not exist. A gate relaxed for a reason that turns out to be false is a shape this repo keeps re-encountering.

One genuine constraint does survive, and should be carried to #4186 in place of the false one: today's precedence note is emitted only when `deployed_roster_section` returns `Some` (`instruction_overrides.rs:256-259`), and this schema cannot express "emit this *text* block only when that generator has content" — `optional` is restricted to `Generated` bodies (`instruction_package.rs:624-629`). A real expressiveness gap, narrow because an empty roster is already a hard `MissingGeneratedInput`.

---

## Verification performed

**Prior HIGH — fixed, verified independently.** I did not rely on the PR's own tests. I built a standalone crate on the same `serde 1.0.228` / `serde_json 1.0.150` with the exact attribute set from `instruction_package.rs:311-312`:

```
=== tag key must NOT be reported unknown ===
[valid text]            OK  -> Text { text: "hello" }               {"kind":"text","text":"hello"}
[valid generated]       OK  -> Generated { generator: AgentRoster } {"kind":"generated","generator":"agent-roster"}
[tag LAST (buffering)]  OK  -> Text { text: "hello" }               {"text":"hello","kind":"text"}
[duplicate tag key]     ERR -> duplicate field `kind`

=== unknown key rejected AND named ===
[misplaced join_before] ERR -> unknown field `join_before`, expected `text`
[misplaced optional]    ERR -> unknown field `optional`, expected `generator`
[typo'd text key]       ERR -> unknown field `txet`, expected `text`
[unknown variant]       ERR -> unknown variant `txet`, expected `text` or `generated`

=== pre-fix reproduction (same enum, attribute removed) ===
[lenient (pre-fix)]     OK  -> Text { text: "hello" }   <-- key silently dropped
```

All three sub-questions answered: the attribute is effective; the internally-tagged `kind` key is consumed by the tagging and is **not** falsely reported unknown, in either key order; and the rejection **names** the offending key. The last line confirms the prior HIGH was a real defect, not a theoretical one.

**Claim 1 — the other three structs already had `deny_unknown_fields`: TRUE.** In base commit `dd347f52`, `InstructionBlock` (:304), `InstructionSection` (:341) and `InstructionPackage` (:363) all carried it; `BlockBody` (:279-280) did not. `50d0a933` adds the attribute at exactly one site. Nothing was added silently and claimed pre-existing.

**Claim 2 — `Join` is externally tagged and already closed: TRUE, not merely convenient.** Verified directly:

```
{"literal":"x"}            OK  -> Literal("x")
{"literal":"x","bogus":1}  ERR -> expected value at line 1 column 14
{"bogus":1,"literal":"x"}  ERR -> unknown variant `bogus`, expected one of `rule`,`blank`,`none`,`literal`
{}                         ERR -> expected value
"rulez"                    ERR -> unknown variant `rulez`
```

Closed in every ordering. Caveat feeding LOW-1: the trailing-extra-key case rejects with `expected value`, which does **not** name `bogus`. The test at `instruction_package_tests.rs:305-307` is honest about this — it binds the error and deliberately does not assert naming — but it means "the error names the key" has one gap that the module-level "throughout" wording papers over.

**Claim 3 — zero `#[serde(flatten)]` in the schema path: TRUE.** Three uses in the whole crate: `core/artifact.rs:34`, `daemon/managed_routes/mod.rs:277` and `:321`. None in `instruction_package.rs`. The `flatten`/`deny_unknown_fields` unsoundness is not in play.

**Claim 4 — `validate` gates on `SCHEMA_VERSION` before inspecting any field: TRUE.** `instruction_package.rs:589-593` is the first statement in `validate`; `package_id` is not touched until :594. Ordering holds. LOW-2 records the nuance that parsing precedes `validate`, so the *rejection* the argument relies on is real but the promised error text is not what an author sees.

**Claim 5 — the evolution policy is stated where an editor sees it: PARTIAL.** `instruction_package.rs:54-57` (module docs) and `CHANGELOG.md` under `[Unreleased]`. Not in the shipped schema JSON — LOW-3.

**MEDIUM 2 as recorded on #4186 — content NOT accurate.** The `output_style` note (point 2) checks out: `inject_style_into_prompt` (`core/output_style.rs:298-304`) prepends heading + body + `SECTION_SEPARATOR` ahead of the whole prompt, applied outside `resolve_pm_prompt` at `core/session_launch/mod.rs:1148-1159`, gated on `native_supported` — so the three-generator set is indeed sufficient. The floor-order note (point 1) checks out. The **main body** carries the stale `resolve_pm_prompt` claim and the conclusion resting on it. That is the HIGH.

---

## Mutation evidence — assessed, and reproduced

I re-ran the mutations myself rather than accepting the reported table. Baseline: `39 passed; 0 failed; 0 ignored` (`cargo test -p trusty-mpm --lib core::instruction_package`).

| Mutation | Direction | Result |
|---|---|---|
| Remove `deny_unknown_fields` from `BlockBody` | under-strict | 1 failure — `rejects_unknown_fields_at_every_level_and_names_the_key`: *"unknown key `join_before` was SILENTLY ACCEPTED at `/blocks/0/body`"* |
| Remove `#[serde(default)]` from `join_before` | **over-strict** | 5 failures — ``missing field `join_before` `` breaks the schema example, both round-trip tests, deterministic compose, and the unknown-field test itself |
| Add the non-decreasing floor-order rule (prior MEDIUM 1) | **over-strict** | 1 failure — `base_pm_floor_block_order_is_valid` |
| Sort blocks by section in `compose` (kill non-contiguity) | design erosion | 3 failures — `composes_blocks_in_array_order_with_declared_joins`, `first_emitted_block_never_contributes_a_join`, `base_pm_floor_block_order_is_valid` |

The tests are load-bearing, not vacuous. The two over-strictness mutations matter most: they are what demonstrate the fix cannot silently become "reject everything," and both are caught — the `default` removal by five independent tests. `valid_package_with_non_contiguous_blocks_still_round_trips` additionally guards itself against passing vacuously by asserting `core` owns non-adjacent positions before testing anything (`instruction_package_tests.rs:790-797`).

**Design constraint — non-contiguity works end-to-end.** Confirmed at three layers: the shipped schema example puts `core` at block indices 0, 2, 4, 7 with `memory`/`search` interleaved (`instruction-package.schema.json:180-216`); `composes_blocks_in_array_order_with_declared_joins` asserts the exact composed bytes `CORE-A\n\nMEMORY\n\nCORE-B`; and the sort mutation produces `IDENTITY…CORE-A\n\nCORE-B…ADDENDUM\n\nMEMORY…`, failing three tests. `sections` and `blocks` remain genuinely separate arrays, and one section owning non-adjacent positions survives the strictness work untouched.

**Fail-open hunt.** Two early-return shapes checked, both sound. `validate_floor_is_last:720-722` returns `Ok(())` when no floor block exists — unreachable, because the `SectionWithoutBlocks` loop at :633-638 runs first and requires all three floor sections to contribute. `compose:776-779`'s `unreachable!("validate rejects blank text blocks")` is genuinely unreachable: `compose` calls `validate()` at :756, and :617-623 rejects blank text. The one real gap of that family is the MEDIUM above — a static coverage check that a runtime `optional` drop can defeat.

---

## Required changes

1. **HIGH** — Rewrite `validate_roster`'s doc block (`instruction_package.rs:648-663`) against the actual current `instruction_overrides.rs`, and correct the corresponding comment on #4186. `resolve_pm_prompt` already delivers the computed roster; byte-identity and roster delivery are not mutually exclusive, and #4186 should keep strict byte-equality against today's output as its acceptance criterion. Carry the real constraint instead: the conditional `ROSTER_PRECEDENCE_NOTE`, which `optional`-on-`Generated`-only cannot express.

Findings 2-5 (1 MEDIUM, 3 LOW) are tracked, not blocking.

---

## Notes

- Read the full `instruction_package.rs` (802 lines), the full test file (867 lines), the full schema JSON (234 lines) and the full `BASE_PM.md`, plus `instruction_overrides.rs`, `delegation_authority.rs`, `output_style.rs` and `session_launch/mod.rs` for the cross-module claims — the HIGH is only visible from outside the module.
- The `unreachable!` at `compose:778` is called out as *checked and sound*, not as a finding; it is reachable only if `validate` is bypassed, and `compose` cannot bypass it.
- Everything else from the prior round is discharged: the `trailing_newline` doc overstatement is corrected (`:392-397`), the changelog carries an `[Unreleased]` entry, and `mod.rs:55-57` records that nothing composes from this module yet.

bobmatnyc:
## VERDICT: WARN

Round 3, reviewed at `bc5bb7bd` in an independent worktree (`origin/main` + `FETCH_HEAD`; the engineer's worktree was not entered). CRITICAL: 0. HIGH: 1. MEDIUM: 0. LOW: 0.

Every round-2 item verified as fixed. The one HIGH below is a **recurrence in the same paragraph** — `validate_roster`'s doc block has now produced a finding in two consecutive rounds — and I am flagging that pattern, not just the instance.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/instruction_package.rs` | 703–708, 724 | "The one real divergence … when NO agents are deployed" is not exhaustive. Two other live `resolve_pm_prompt` branches also deliver no roster, and both are *also* inexpressible under this schema's `validate()`. | Qualify the claim, or (preferred) delete the forward-looking #4186 prescription — see below. |

### HIGH — the zero-agent case is not the only divergence

Line 703–704 states:

> `instruction_overrides::resolve_pm_prompt` resolves its delegation section through `bundled_delegation`

and line 708:

> The delivered prompt therefore already carries the real roster

Both are unqualified. `bundled_delegation` is the **fallback only**. Two other branches of the same function skip it entirely:

1. **`AGENT_DELEGATION.md` override** — `instruction_overrides.rs:173-174`:
   ```rust
   let delegation = read_override(&dir, FILE_AGENT_DELEGATION)
       .unwrap_or_else(|| bundled_delegation(project_dir));
   ```
   When the override exists, `bundled_delegation` never runs, so neither the live roster nor `ROSTER_PRECEDENCE_NOTE` reaches the prompt. That file's own doc says so at `instruction_overrides.rs:138-139`: *"an `AGENT_DELEGATION.md` override still replaces section and roster alike."*

2. **`PM_INSTRUCTIONS_DEPLOYED.md` full replacement** — `instruction_overrides.rs:161-168`: the branch emits `[body, stack, (instructions), floor]`. There is **no delegation section at all**, and no workflow/memory/search content.

Both branches are live and covered by existing tests (`instruction_overrides.rs:640-671`).

**Why this matters for #4186**, which is what the paragraph exists to advise. Under this schema, `validate()` requires *both* a non-optional `agent-roster` block (`instruction_package.rs:735-755`) *and* a guaranteed-emitting block for all eight sections (`instruction_package.rs:688-695`). So today's output is unreachable for **three** configurations, not one:

| Configuration | Today's output | Under this schema |
|---|---|---|
| Zero deployed agents | asset alone | `MissingGeneratedInput` — **named at line 724** |
| `AGENT_DELEGATION.md` override | override text, no roster | `RosterNotConsumed` — **not named** |
| `PM_INSTRUCTIONS_DEPLOYED.md` | no delegation/workflow/memory/search | `SectionWithoutBlocks` — **not named** |

The override cases raise a design question line 724 asserts is already settled: **does override substitution run before or after `validate()`?** If before, a package-composed launch path hard-errors on any project carrying an `AGENT_DELEGATION.md` — a configuration that works today. That choice has to be made either way, so the claim of exhaustiveness is wrong regardless of which design #4186 picks.

This has **already propagated**: the posted correction on issue #4186 repeats it verbatim — *"That is the only choice that needs making."* Same propagation channel round 2 flagged.

#### Fix

Preferred — remove the forward-looking prescription rather than try to make it exhaustive:

- Keep lines 700–722 (what the code does today, and the shape-is-expressible conclusion). That part is **verified correct** and is executably pinned by `todays_delegation_section_shape_is_expressible`.
- Delete lines 724–733 (the "one real divergence" prescription) and move the divergence analysis to #4186 itself, where it can be revised without a PR.

Minimum acceptable alternative — qualify in place:
- Line 703–704: note `bundled_delegation` is the **no-override fallback**.
- Line 724: reframe as *"the divergence in the default, no-override path"*, and add the two override branches as cases #4186 must decide.
- Post a short follow-up on #4186 replacing "the only choice that needs making".

---

## Verification of each round-2 → round-3 change

| # | Item | Result |
|---|---|---|
| 1 | Corrected `validate_roster` doc | **Accurate as far as it goes.** `bundled_delegation` (`instruction_overrides.rs:254-260`) does build `{bundled}\n\n{NOTE}\n\n{roster}`; `deployed_roster_section` (`delegation_authority.rs:352-370`) is called directly, so the stale `PipelineOutput::merged` path is genuinely no longer the launch path. Incomplete only in the scope covered by the HIGH. |
| 2 | Second carrier ("not strict equality") | **Fixed.** Independent grep of the four files for `still builds`, `PipelineOutput`, `mutually exclusive`, `never reads`, `static AGENT_DELEGATION`, `reviewed diff`, `XOR`, `cannot both`, `must choose`, `tension`, `strict equality`: only two survivors, both in explicitly *corrective* framing — `instruction_package.rs:709` ("is no longer the launch path") and `instruction_package_tests.rs:881` ("pins the CORRECTION of a wrong conclusion"). Spot-checked all 24 `#4196`/`#4069` mentions: all consistent with the corrected story. **No third false-premise carrier survives.** |
| 3 | Re-derived conclusion | **Byte-identity claim: verified.** Shape is expressible as three blocks; `todays_delegation_section_shape_is_expressible` passes and is load-bearing (mutation 8 kills it). **Zero-agent claim: verified true but NOT exhaustive** — the HIGH. |
| 4 | #4186 correction comment | **Posted, matches the in-code claim, original left intact.** No drift between the two records — including, unfortunately, the over-claim. |
| 5 | `SectionOnlyOptionalBlocks` + reorder | **Verified.** Reverting the reorder (roster check back after coverage) breaks **exactly one** test, `rejects_optional_roster_block` — confirming the reorder was necessary and changed exactly the one intended diagnosis, no others. Documented `ValidationError` variant order matches actual execution; `RosterNotConsumed` / `OptionalRoster` are disjoint (a package either has a roster block or does not), so their relative order is unobservable. |
| 6 | Version gate before deserialize | **Verified.** Mutation 7 (probe removed) reproduces the claimed message exactly: ``unknown field `field_added_in_v2`, expected one of `section`, `body`, `join_before`, `optional` at line 1 column 88``. The probe **cannot false-reject a valid v1 package** — confirmed with a reviewer-added probe test: it tolerates unknown keys by construction, and the only inputs it rejects (missing / non-`u32` `schema_version`) are rejected by the full deserializer too. |
| 7 | Policy in shipped schema JSON | **Verified** — `schema_version.description` carries the full evolution policy including "Any field addition that can change composed bytes MUST bump this version." |
| 8 | `Join` extra-key deferral | **Reasonable, accept.** It is a rejection, not a silent drop, so the failure mode this PR exists to close is shut. `Join` is externally tagged, so `deny_unknown_fields` genuinely does not apply; naming the key would need a hand-written `Deserialize` for the rarest variant. The line/column pin (`instruction_package_tests.rs:316-320`) is load-bearing — it prevents degradation to a locator-free message, and the sibling assertions still pin that *unrecognised variants* are named in both spellings. |

## Mutations

| # | Mutation | Result |
|---|---|---|
| 6 | Delete the `SectionOnlyOptionalBlocks` check | 1 failed — `rejects_section_covered_only_by_optional_blocks`. Sound. |
| 7 | Remove the version probe from `from_json` | 1 failed — `later_schema_version_is_reported_as_a_version_error_not_a_field_error`, with the exact claimed message. Sound; pin is load-bearing. |
| 8 | `Join::Blank` → `"\n"` | 3 failed — `join_literals_are_exact`, `composes_blocks_in_array_order_with_declared_joins`, `todays_delegation_section_shape_is_expressible`. Sound; the shape test is genuinely pinned to the byte value, not just the literal test. |
| — | *(reviewer)* Revert the roster/coverage reorder | 1 failed — `rejects_optional_roster_block` only. Confirms no other diagnosis changed. |

Local: `cargo test -p trusty-mpm --lib instruction_package` → `test result: ok. 42 passed; 0 failed; 0 ignored`.

## Fail-open sweep

No fail-open shapes found. Specifically checked and cleared:

- `compose` calls `self.validate()?` unconditionally (line 827), so the `unreachable!` at line 849 is genuinely unreachable — `EmptyText` rejects blank text blocks first.
- `validate_floor_is_last`'s early `return Ok(())` when no floor block exists (lines 791–793) is not fail-open: it runs *after* the coverage loop, which already rejects any package missing a floor section.
- `Generator::AgentRoster` resolves to `Some(...)` unconditionally (line 836), so an empty roster becomes `MissingGeneratedInput` rather than a silent drop.
- `validate_roster` sets `seen` only for non-optional roster blocks, and a second, optional roster block is still rejected.

## Notes

- **Convergence assessment: the code has converged; the prose has not.** Rounds 1–3 found zero defects in the type/validation/composition logic — it was correct in round 1 and is correct now. All three findings landed in doc comments, and rounds 2 and 3 landed in *the same paragraph*: `validate_roster`'s forward-looking guidance to #4186. The recurring defect is structural, not random — this PR keeps trying to settle a downstream ticket's scope in a doc comment, using absolutes ("the one real divergence", "the only choice that needs making") that the code cannot back. That is why the recommendation is to delete lines 724–733 rather than make them exhaustive: a fourth attempt to enumerate correctly is likelier to draw a round 4 than a fix that stops making the claim. The verified, executable part of the conclusion stays.
- Not filed, as directed: commit `50d0a933`'s message still states the disproven XOR. Correcting it in the record rather than force-pushing was the right trade.
- Very minor, not a finding: the `ValidationError` variant list documents check order, but the coverage loop iterates *per section*, so a package with both a missing section and an only-optional section reports whichever section comes first canonically. Deterministic and harmless.

## CI

21 checks passing, 1 pending (`Test`), 0 failing at time of review. **Do not merge until `Test` reports green.**

bobmatnyc:
## VERDICT: APPROVE

Round 4 (closing pass) at `c921d769`. Scoped to the delta since the round-3 verdict at `bc5bb7bd` — the schema, validation ordering, version probe, `deny_unknown_fields`, `Join` semantics, and the mutation evidence were verified across rounds 1-3 and are unchanged, so they were not re-reviewed.

**CRITICAL: 0 — HIGH: 0 — MEDIUM: 0 — LOW: 0.** No issues found at >80% confidence.

### Verifications

**1. The prescription deletion is complete.** `git diff bc5bb7bd..c921d769` is a single hunk in `crates/trusty-mpm/src/core/instruction_package.rs` — 7 insertions, 14 deletions, all `///` lines. Grepped all five files in the PR (`instruction_package.rs`, `instruction_package_tests.rs`, `instruction-package.schema.json`, `CHANGELOG.md`, `core/mod.rs`) at `c921d769` for `zero-agent`, `zero agent`, `only choice`, `one real divergence`, `must either guarantee`, `silently degrades`, `byte-identity holds`: **zero hits in all five**. No surviving carrier — the prescription was deleted, not relocated.

**2. The qualified sentence is true as scoped, and the block reintroduces nothing.** Read the entire doc block, `instruction_package.rs:700-727`, not the hunk. `instruction_package.rs:707-710` now reads "...appends the LIVE computed roster ... separated by `ROSTER_PRECEDENCE_NOTE`. **On this path** the delivered prompt therefore already carries the real roster". Checked against the referenced source: `instruction_overrides.rs:254-260` (`bundled_delegation`) appends `{bundled}\n\n{ROSTER_PRECEDENCE_NOTE}\n\n{roster}` on the `Some` arm, which is exactly the path the sentence describes — the consequence follows on that path and is no longer asserted beyond it. The remainder of the block (`:712-722`) asserts only a capability ("a package **can be** byte-identical to today's output AND consume the computed roster") plus a recommendation for #4186's acceptance gate; neither is a universal claim about `resolve_pm_prompt`'s behavior.

**3. The pointer line asserts no count and enumerates nothing.** `instruction_package.rs:724-726`: "Scope — including the other override configurations `resolve_pm_prompt` can emit, and which of them this schema can reproduce — is tracked on #4186, not here." It signals that other configurations exist and hands the question to #4186 without stating how many, which they are, or what any of them does. Correct shape.

**4. Zero non-comment `.rs` lines changed.** `git diff bc5bb7bd..c921d769 -- '*.rs'` filtered to lines not matching `^[+-]\s*///` returns **empty**. Per-hop: `bc5bb7bd..44b6eb68` = 3/10 (all `///`), `44b6eb68..c921d769` = 4/4 (all `///`). The type, validation, and composition logic is byte-identical to the tree verified in round 3.

**5. No test orphaned; count holds at 42.** `instruction_package_tests.rs` is byte-identical since `bc5bb7bd` (`git diff --stat` empty for that path). `todays_delegation_section_shape_is_expressible` survives at `instruction_package_tests.rs:879` and pins the RETAINED text (the shape `A\n\nB\n\n<generated roster>`), not the deleted prescription. Ran directly in a detached worktree at `c921d769`:

```
test core::instruction_package::tests::todays_delegation_section_shape_is_expressible ... ok

test result: ok. 42 passed; 0 failed; 0 ignored; 0 measured; 3752 filtered out; finished in 0.01s
```

### CI

`gh pr checks 4223` at `c921d769`: **passing** — 22 checks pass, 0 failing, 0 pending (`Notify on red main` correctly reports `skipping`). Includes `Test` (16m44s), `Clippy`, `MSRV check (1.91)`, `Format check`, and the doc-comment pointer lint.

`mergeStateStatus=BEHIND` is expected and correct — #4224 merges first by deliberate serialization. Not a finding.

### Notes

On the round count: the code converged in round 1 and never carried a defect. Rounds 2, 3, and this one all concerned one doc block. The correct fix — deleting the prescription rather than attempting a fourth enumeration of it — is what landed, and this pass is clean.

One observation, deliberately **not** filed as a finding: the same "the delivered prompt already carries the real roster" claim also appears unqualified in a test comment at `instruction_package_tests.rs:890`, immediately after that comment quotes the `bundled_delegation` `format!` it describes. It is scoped by its own surrounding context, and the test it annotates pins only the shape, so it is not misleading. I flag it only so a future reader does not "discover" it and open a fifth revision of the same claim. **The signal is about the block, not about any one sentence in it: this claim has drawn a finding in three consecutive rounds because it restates behavior that lives in another module and drifts against it. If it draws one again, the right response is to delete the descriptive paragraph outright and leave the #4186 pointer — not to revise the wording a fifth time.** No change is requested now.

APPROVED for merge.

PR #4224, opened 2026-07-28
fix(trusty-mpm): derive worktrees from git instead of walking five shapes (#4207)

bobmatnyc:
## What

Worktree **discovery** and worktree **ownership** were both inferences. This
replaces each with the fact git already maintains.

| Question | Before | After |
| --- | --- | --- |
| Where are the worktrees? | walk five hard-coded location shapes | `git worktree list --porcelain` |
| Which checkout owns this one? | the candidate's **grandparent directory** | `git rev-parse --git-common-dir` |

Slice 1 of #4207, plus the half of slice 2 that PR #4209 does not touch (see
"Relationship to #4209" below).

## Why the grandparent rule made 14 worktrees structurally unreclaimable

Observed on the dogfood machine 2026-07-27: fourteen worktrees live under
`<repo>/.base/.worktrees/` but are **registered to the parent repo** `<repo>`.
The old rule derived the owning checkout as the candidate's grandparent —
`<repo>/.base` — which is a *real* repository that correctly disowns them. So
`git_worktree_list_agrees` returned `false`, the reclaim path skipped them
conservatively, and it did so on every sweep, forever. No amount of adding a
sixth location shape would have fixed it, because the location was never the
problem.

The same rule is wrong for `<repo>/.claude/worktrees/<name>`, where the
grandparent is `<repo>/.claude` — not a repository at all.

## Why the five-shape walk had to go rather than gain a sixth shape

`.worktrees/`, `.base/.worktrees/`, `.claude/worktrees/`,
`.base/.claude/worktrees/`, `.base/.worktrees/<id>/.claude/worktrees/` — each
arrived as a bug report about a location the previous list had missed (#3649,
#3971, and the #3971 follow-up). Nothing stops git from registering a worktree
anywhere on disk, so the list could never be complete. Asking git deletes the
category of bug.

## An unanswerable probe is never a verdict

Every function in the new module returns `Option`, never `bool`, for "could not
determine". No caller can read a missing `git` binary, a permissions error, or
a stalled mount as a fact about a worktree — the invariant is in the type.

This is deliberately the shape PR #4202's round-1 fix broke by collapsing every
`canonicalize()` error kind into `Destroyed`; on 2026-07-21 a rule that loose
would have told an operator to recreate ~70 fully intact worktrees. The
asymmetry here is absolute in the safe direction: a candidate that will not
canonicalize is **skipped**, so a failed observation can only ever *shrink*
what is proposed for deletion, never enlarge it
(`enumerate_ignores_a_worktree_whose_directory_is_gone`).

No alarm, latch, or give-up path is added — the correct way not to ship an
unlatched alarm (PR #4202's second defect) is not to add one.

## Blast radius is unchanged

Git will happily report a worktree parked in `/tmp`. Enumeration keeps the
containment boundary the old walk had *structurally* — it could only ever
produce paths beneath the root it was handed — as an explicit, tested rule
(`enumerate_excludes_worktrees_outside_the_repos_root`). The main checkout,
bare records, and already-prunable entries are excluded. All existing gates are
untouched and still run in order: the #3649 ownership sentinel, the
`git worktree list` cross-check, and the #4091/#4118 dirty-tree guard.

## What this narrows, deliberately

A directory git does **not** register is no longer a candidate. The old walk
collected any leaf directory under a `.worktrees/` parent, so a stray `mkdir`
was indistinguishable from a checkout git created.

Such husks were **already** unreclaimable — `git_worktree_list_agrees` refused
every one of them — so no reclaim behaviour regresses. What changes is that
they are also absent from the *report*. Reclaiming unregistered husks is a
separate concern (#3715), deliberately not smuggled in here.

## `is_session_worktree()` and `is_dir()` — left alone, on purpose

PR #4209 argues they answer *"may trusty-mpm DELETE this?"* (ownership) on
paths that may legitimately be absent, and that teaching them about `.git`
would make `decommission` and `search_gc` refuse to clean up the very husks
#4204 is about. That argument is correct, and this PR does not touch either.

## Relationship to #4209

Complementary, not competing, and they do not overlap by file.

- **#4209** is the *liveness* half: is this workspace safe to WRITE INTO?
  It owns `core/workspace_liveness.rs`, `agent_reset_workspace.rs`, and
  `guided_inplace.rs`.
- **This PR** is the *derivation* half: what IS a worktree and WHO owns it?
  It owns `session_manager/worktree_registry.rs`, `prune.rs`,
  `worktree_safety.rs`, and `decommission.rs`.

Landing order does not matter.

## Tests

Net **-651 / +946** in the diff, but the production walk shrinks from ~120
lines to 4; the growth is the new module's docs and its real-git tests. Seven
shape-coverage tests collapse into two location-agnostic ones, because location
is no longer something the code reasons about.

Every worktree in the affected tests is now a real `git worktree add`. The
previous `mkdir` fixtures would now pass **vacuously** — asserting an empty list
against an empty list — so each converted test also asserts the candidate
actually reaches the gate under test.

### Revert-proofs (both verified by temporarily restoring the old rule)

**1. `git_worktree_list_agrees_true_for_worktree_registered_to_parent_repo`** —
restore `candidate.parent().and_then(|p| p.parent())`:

```
panicked at prune_orphan_tests.rs:616:
a worktree registered to the PARENT repo but living under .base must be
recognised — asking the grandparent (.base) disowns it
```

The 2 that still pass are the two pre-existing `git_worktree_list_agrees`
tests — in both, the grandparent *happens* to be the owning repository, which
is exactly why neither could ever catch this. The test asserts up front that
the grandparent is itself a real repository, so the old code cannot pass by
falling through its best-effort `true`.

**2. `find_orphaned_worktrees_discovers_worktree_at_unwalked_location`** —
restore the five-shape walk:

```
panicked at prune_orphan_tests.rs:201:
a registered worktree must be found wherever it lives; got []

panicked at prune_orphan_tests.rs:221:
a plain directory is not a registered worktree; got [".../.worktrees/just-a-mkdir"]
```

## Verification

- `cargo clippy --workspace --all-targets --exclude {3 gui crates} -- -D warnings` → **exit 0**
- `cargo fmt --all --check` → clean
- `scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`
- `cargo test --workspace --exclude {3 gui crates} --no-fail-fast` →
  **194 suites, 19707 passed, 5 failed, 132 ignored**

All 5 failures are in `trusty-agents`, a crate this PR does not touch, and none
is a code failure:

- 4 × `tests/api_e2e.rs` — `api server did not become healthy … within 5s`
  (host at load average ~110 with ~8 concurrent agents)
- 1 × `tools::python_skill` — `uv` could not inspect the Python interpreter
  (`ENOENT` on `.cache/uv/.tmp*`)

`trusty-mpm` itself: **0 failures**, no `#[ignore]`, no cfg-gates, no
`--exclude` beyond CI's own three GUI crates, and never `--lib`-only.

Refs #4207 (slice 1 + the ownership half of slice 2)
Fixes the underlying defect of #4204

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -33,10 +33,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   position rather than by a list of names, so a new one is excluded the day you
   create it. Location remains irrelevant *within* a project, so no reclaim
   capability is lost (#4207).
-- A git-`locked` worktree is never enumerated for reclamation. Removal uses
-  `git worktree remove --force`, which overrides the lock, so refusing to
-  consider a locked worktree is the only place `git worktree lock` can be
-  honoured (#4207).
+- A git-`locked` worktree is never enumerated for reclamation. Git itself
+  respects the lock — `git worktree remove --force` refuses it — but the
+  removal path treats that refusal as a generic git failure and falls back to
+  deleting the directory outright, which defeats the lock and strands git's
+  registry entry. Excluding locked worktrees from enumeration is the only place
+  `git worktree lock` is actually honoured (#4207).
 
 ---
 ## [1.2.3] — 2026-07-28
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -44,6 +44,19 @@ PATCH: merge-and-local-install only, no crates.io publish.
   and existing hand-placed or committed `.claude/` content is never overwritten
   — unmanaged files are skipped as user-owned, and managed files whose checksum
   drifted are skipped with a warning.
+- Worktree discovery is derived from `git worktree list --porcelain` instead of
+  walking five hard-coded location shapes, so a session worktree is found
+  wherever it lives rather than only where someone remembered to look (#4207
+  slice 1).
+- The checkout that owns a worktree is now resolved with
+  `git rev-parse --git-common-dir` instead of being guessed as the candidate's
+  grandparent directory. Worktrees physically inside `.base/.worktrees/` but
+  registered to the parent repository were previously disowned by that guess
+  and were structurally unreclaimable — never deleted, never reportable (#4207,
+  underlying defect of #4204).
+- `tm session prune --worktrees` and the daemon orphan-GC no longer propose a
+  plain directory as a reclaim candidate. A directory git does not register is
+  not trusty-mpm's to remove.
 
 - **Two paths no longer write into a destroyed worktree**
   ([#4204](https://github.com/bobmatnyc/trusty-tools/issues/4204)):
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,6 +7,37 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Fixed
+
+- Worktree discovery is derived from `git worktree list --porcelain` instead of
+  walking five hard-coded location shapes, so a session worktree is found
+  wherever it lives rather than only where someone remembered to look (#4207
+  slice 1).
+- The checkout that owns a worktree is now resolved with
+  `git rev-parse --git-common-dir` instead of being guessed as the candidate's
+  grandparent directory. Worktrees physically inside `.base/.worktrees/` but
+  registered to the parent repository were previously disowned by that guess
+  and were structurally unreclaimable — never deleted, never reportable (#4207,
+  underlying defect of #4204).
+- `tm session prune --worktrees` and the daemon orphan-GC no longer propose a
+  plain directory as a reclaim candidate. A directory git does not register is
+  not trusty-mpm's to remove.
+- Worktree reclamation is bounded by a structural containment rule: a candidate
+  must be a strict descendant of the managed project directory whose git
+  registry named it. Deriving discovery from git (above) widened what could be
+  proposed for deletion from five path shapes to anything registered anywhere
+  under the projects root — on this machine that admitted five of the
+  operator's own long-lived checkouts, leaving the ownership sentinel as the
+  only thing between enumeration and an unattended removal. Your own checkouts
+  and any worktree you park beside a managed project are now excluded by
+  position rather than by a list of names, so a new one is excluded the day you
+  create it. Location remains irrelevant *within* a project, so no reclaim
+  capability is lost (#4207).
+- A git-`locked` worktree is never enumerated for reclamation. Removal uses
+  `git worktree remove --force`, which overrides the lock, so refusing to
+  consider a locked worktree is the only place `git worktree lock` can be
+  honoured (#4207).
+
 ---
 ## [1.2.3] — 2026-07-28
 
@@ -44,19 +75,6 @@ PATCH: merge-and-local-install only, no crates.io publish.
   and existing hand-placed or committed `.claude/` content is never overwritten
   — unmanaged files are skipped as user-owned, and managed files whose checksum
   drifted are skipped with a warning.
-- Worktree discovery is derived from `git worktree list --porcelain` instead of
-  walking five hard-coded location shapes, so a session worktree is found
-  wherever it lives rather than only where someone remembered to look (#4207
-  slice 1).
-- The checkout that owns a worktree is now resolved with
-  `git rev-parse --git-common-dir` instead of being guessed as the candidate's
-  grandparent directory. Worktrees physically inside `.base/.worktrees/` but
-  registered to the parent repository were previously disowned by that guess
-  and were structurally unreclaimable — never deleted, never reportable (#4207,
-  underlying defect of #4204).
-- `tm session prune --worktrees` and the daemon orphan-GC no longer propose a
-  plain directory as a reclaim candidate. A directory git does not register is
-  not trusty-mpm's to remove.
 
 - **Two paths no longer write into a destroyed worktree**
   ([#4204](https://github.com/bobmatnyc/trusty-tools/issues/4204)):
--- a/crates/trusty-mpm/src/daemon/doctor.rs
+++ b/crates/trusty-mpm/src/daemon/doctor.rs
@@ -377,13 +377,22 @@ fn build_gh_account_check(active: Option<String>, accounts: Vec<String>) -> Doct
 /// that were decommissioned before this fix—or where `git worktree remove`
 /// failed—may leave stale `.worktrees/<session-id>/` directories on disk. This
 /// probe surfaces orphaned dirs so operators know to run
-/// `tm session prune --worktrees`. The filesystem walk is delegated to
+/// `tm session prune --worktrees`. Discovery is delegated to
 /// [`crate::session_manager::prune::find_orphaned_worktrees`] inside
-/// `spawn_blocking` so the async executor is not blocked by synchronous I/O.
+/// `spawn_blocking` so the async executor is not blocked by the synchronous
+/// git subprocesses it spawns.
 /// What: builds a canonicalized active-path set, then spawns a blocking task
-/// to walk `<repos_root>/<owner>/<repo>/.worktrees/`; any leaf directory not
-/// in the active set is counted as an orphan. Reports `Ok` (no orphans),
-/// `Warn` (orphans found), or `Ok` (repos_root absent / unconfigured).
+/// that asks GIT for the worktrees of each managed project (#4207 — this is no
+/// longer a filesystem walk of `.worktrees/`, and a candidate is no longer "any
+/// leaf directory"): every git-registered worktree that is a strict descendant
+/// of the project whose registry named it, minus the active set, is counted as
+/// an orphan. Reports `Ok` (no orphans), `Warn` (orphans found), or `Ok`
+/// (repos_root absent / unconfigured).
+///
+/// Note this probe is REPORT-ONLY — it counts, it never removes — so it
+/// deliberately reports candidates that the reclaim path would refuse to
+/// delete (owner-unknown, dirty). A directory git does not register is not
+/// counted at all; reclaiming unregistered husks is #3715, still open.
 /// Test: `worktrees_no_orphans_is_ok`, `worktrees_with_orphan_is_warn`.
 async fn check_worktrees(
     repos_root: Option<&Path>,
--- a/crates/trusty-mpm/src/daemon/doctor.rs
+++ b/crates/trusty-mpm/src/daemon/doctor.rs
@@ -384,15 +384,17 @@ fn build_gh_account_check(active: Option<String>, accounts: Vec<String>) -> Doct
 /// What: builds a canonicalized active-path set, then spawns a blocking task
 /// that asks GIT for the worktrees of each managed project (#4207 — this is no
 /// longer a filesystem walk of `.worktrees/`, and a candidate is no longer "any
-/// leaf directory"): every git-registered worktree that is a strict descendant
-/// of the project whose registry named it, minus the active set, is counted as
-/// an orphan. Reports `Ok` (no orphans), `Warn` (orphans found), or `Ok`
-/// (repos_root absent / unconfigured).
+/// leaf directory"): a git-registered worktree counts as an orphan when it is
+/// not main/bare/prunable/locked, IS a strict descendant of the project whose
+/// registry named it, and is absent from the active set. Reports `Ok` (no
+/// orphans), `Warn` (orphans found), or `Ok` (repos_root absent /
+/// unconfigured).
 ///
 /// Note this probe is REPORT-ONLY — it counts, it never removes — so it
-/// deliberately reports candidates that the reclaim path would refuse to
-/// delete (owner-unknown, dirty). A directory git does not register is not
-/// counted at all; reclaiming unregistered husks is #3715, still open.
+/// deliberately reports candidates the reclaim path would refuse to delete
+/// (owner-unknown, dirty). Conversely a directory git does not register is not
+/// counted at all, so an unregistered husk is invisible here; reclaiming those
+/// is a separate open concern, deliberately out of scope for #4207.
 /// Test: `worktrees_no_orphans_is_ok`, `worktrees_with_orphan_is_warn`.
 async fn check_worktrees(
     repos_root: Option<&Path>,
@@ -422,8 +424,9 @@ async fn check_worktrees(
         .map(|p| std::fs::canonicalize(p).unwrap_or_else(|_| p.clone()))
         .collect();
 
-    // Delegate the blocking filesystem walk to spawn_blocking so the async
-    // executor is not stalled on directory I/O (#1840 F).
+    // Delegate the blocking scan to spawn_blocking so the async executor is not
+    // stalled (#1840 F). Since #4207 what blocks is not directory I/O but the
+    // git subprocesses `find_orphaned_worktrees` spawns per managed project.
     let orphans = tokio::task::spawn_blocking(move || {
         crate::session_manager::prune::find_orphaned_worktrees(&root, &active_set)
     })
--- a/crates/trusty-mpm/src/daemon/doctor_tests.rs
+++ b/crates/trusty-mpm/src/daemon/doctor_tests.rs
@@ -276,47 +276,23 @@ async fn worktrees_no_repos_root_is_ok() {
 
 #[tokio::test]
 async fn worktrees_no_orphans_is_ok() {
-    // Why (#1845 item 2): on macOS /tmp is a symlink to /private/tmp. If we pass the
-    // raw tempfile path as active (e.g. /tmp/…) but the walk canonicalises it to
-    // /private/tmp/…, the active-set lookup misses and a live worktree is falsely
-    // flagged as an orphan — causing a spurious CI failure. Canonicalize the active
-    // path in the test so the set membership check in find_orphaned_worktrees matches.
-    let tmp = tempfile::tempdir().unwrap();
-    let root = tmp.path();
-    // Create owner/repo/.worktrees/session-abc/ and register it as active.
-    let wt = root
-        .join("owner")
-        .join("repo")
-        .join(".worktrees")
-        .join("session-abc");
-    std::fs::create_dir_all(&wt).unwrap();
-    // Canonicalize so /tmp vs /private/tmp symlink differences are resolved.
-    let canonical_wt = std::fs::canonicalize(&wt).unwrap_or(wt);
-    let active = vec![canonical_wt];
-    let check = check_worktrees(Some(root), &active).await;
+    // #4207: a REAL `git worktree add`, since discovery is now derived from
+    // git's registry — a `mkdir` is not a candidate and would make this pass
+    // for the wrong reason. The fixture canonicalizes its own root, so the
+    // /tmp vs /private/tmp symlink hazard (#1845 item 2) is handled there.
+    let fx = crate::session_manager::worktree_git_fixture::GitWorktreeFixture::new();
+    let wt = fx.add_worktree("session-abc");
+    let check = check_worktrees(Some(&fx.repos_root), &[wt]).await;
     assert_eq!(check.status, CheckStatus::Ok);
 }
 
 #[tokio::test]
 async fn worktrees_with_orphan_is_warn() {
-    let tmp = tempfile::tempdir().unwrap();
-    let root = tmp.path();
-    // Create two worktrees; only one has an active session.
-    let wt1 = root
-        .join("owner")
-        .join("repo")
-        .join(".worktrees")
-        .join("session-live");
-    let wt2 = root
-        .join("owner")
-        .join("repo")
-        .join(".worktrees")
-        .join("session-dead");
-    std::fs::create_dir_all(&wt1).unwrap();
-    std::fs::create_dir_all(&wt2).unwrap();
-    // wt2 is orphaned (no active session).
-    let active = vec![wt1];
-    let check = check_worktrees(Some(root), &active).await;
+    // #4207: two REAL worktrees; only one has an active session.
+    let fx = crate::session_manager::worktree_git_fixture::GitWorktreeFixture::new();
+    let live = fx.add_worktree("session-live");
+    let _dead = fx.add_worktree("session-dead");
+    let check = check_worktrees(Some(&fx.repos_root), &[live]).await;
     assert_eq!(check.status, CheckStatus::Warn);
     assert!(check.message.contains("1 orphaned"));
     assert!(check.message.contains("tm session prune --worktrees"));
--- a/crates/trusty-mpm/src/daemon/mod.rs
+++ b/crates/trusty-mpm/src/daemon/mod.rs
@@ -666,10 +666,16 @@ async fn orphan_gc_loop(state: Arc<DaemonState>, cancel: tokio_util::sync::Cance
                 // no longer true: a registered worktree at ANY depth inside a
                 // managed project, e.g. `<repo>/agents/scratch/wt-1`, is a
                 // candidate). A directory must clear ALL of: git registers it as
-                // a worktree that is not main/bare/prunable/locked; it is a
-                // strict descendant of the managed project whose registry named
-                // it (so operator checkouts and the base clone are unreachable by
-                // position, not by name); it is in neither the initial nor the
+                // a worktree that is not main/bare/prunable/locked — this is
+                // what excludes the BASE CLONE, which is a separate
+                // `git clone --bare` absent from `<repo>`'s registry entirely
+                // and bare+main in its own; it is a strict descendant of the
+                // managed project whose registry named it, which is what puts
+                // the operator's own checkouts parked beside a project out of
+                // reach by POSITION rather than by name — note this gate does
+                // NOT exclude `.base`, which IS a strict descendant of `<repo>`,
+                // so do not re-attribute the base clone to it; it is in neither
+                // the initial nor the
                 // pre-deletion active set; it carries a #3649 ownership sentinel
                 // naming a provably-ownerless owner past the grace window; and
                 // the #4091/#4118 dirty gate finds no uncommitted or unpushed
--- a/crates/trusty-mpm/src/daemon/mod.rs
+++ b/crates/trusty-mpm/src/daemon/mod.rs
@@ -658,10 +658,24 @@ async fn orphan_gc_loop(state: Arc<DaemonState>, cancel: tokio_util::sync::Cance
                 // clone otherwise grows without bound (94 dirs observed for one
                 // project) because decommission-time cleanup (#1806/#1840) only
                 // covers sessions torn down AFTER that fix landed and the manual
-                // `tm session prune-worktrees` sweep is rarely run. Only leaf
-                // worktree dirs with NO live record are removed — active worktrees
-                // and the base clone are never touched. Inherits the orphan-GC env
-                // gate (`TRUSTY_MPM_ORPHAN_GC`) since it runs inside this loop.
+                // `tm session prune-worktrees` sweep is rarely run.
+                //
+                // What actually bounds this call (#4207/#4224 — the previous
+                // wording here, "only leaf worktree dirs ... the base clone are
+                // never touched", described the removed five-shape walk and is
+                // no longer true: a registered worktree at ANY depth inside a
+                // managed project, e.g. `<repo>/agents/scratch/wt-1`, is a
+                // candidate). A directory must clear ALL of: git registers it as
+                // a worktree that is not main/bare/prunable/locked; it is a
+                // strict descendant of the managed project whose registry named
+                // it (so operator checkouts and the base clone are unreachable by
+                // position, not by name); it is in neither the initial nor the
+                // pre-deletion active set; it carries a #3649 ownership sentinel
+                // naming a provably-ownerless owner past the grace window; and
+                // the #4091/#4118 dirty gate finds no uncommitted or unpushed
+                // work. See `SessionManager::reap_orphaned_worktrees` for the
+                // authoritative list. Inherits the orphan-GC env gate
+                // (`TRUSTY_MPM_ORPHAN_GC`) since it runs inside this loop.
                 let repos_root = managed_routes::inproject::repos_root();
                 match mgr.reap_orphaned_worktrees(&repos_root).await {
                     Ok(outcome) => {
--- a/crates/trusty-mpm/src/session_manager/decommission.rs
+++ b/crates/trusty-mpm/src/session_manager/decommission.rs
@@ -104,8 +104,9 @@ pub(crate) fn is_session_worktree(path: &Path) -> bool {
 /// `git worktree list` and `git branch` output. `git worktree remove --force`
 /// prunes both the directory AND the ref atomically, restoring a clean state.
 /// What: runs `git -C <repo-root> worktree remove --force <path>` where
-/// `<repo-root>` is the grandparent of the worktree directory
-/// (`path` → `.worktrees` → `<repo-root>`). Also runs
+/// `<repo-root>` is the checkout git itself reports as owning `path`'s
+/// worktree registry ([`super::worktree_registry::registry_root_for`], #4207 —
+/// previously guessed as the grandparent directory). Also runs
 /// `git -C <repo-root> worktree prune` and
 /// `git -C <repo-root> branch -D session/<leaf>` on success to clear stale git
 /// refs and the session branch, where `<leaf>` is the last component of `path`
@@ -158,18 +159,30 @@ pub(super) fn remove_session_worktree(path: &Path) -> bool {
         );
     }
 
-    // The repo root is the grandparent of the worktree dir:
-    // <repo-root>/.worktrees/<session-id>/ → grandparent = <repo-root>
-    let repo_root = match path.parent().and_then(|p| p.parent()) {
+    // #4207: ask git which checkout owns this worktree's registry instead of
+    // guessing that it is the grandparent directory. The grandparent rule held
+    // only for the two shapes it was written against; a worktree registered to
+    // the parent repo but living under `.base/.worktrees/` resolved to `.base`,
+    // which disowns it, so `git worktree remove` there could never succeed.
+    let repo_root = match super::worktree_registry::registry_root_for(path) {
         Some(r) => r,
         None => {
+            // No git registry claims this path — it is a plain directory, not a
+            // worktree. The sentinel gate above already established trusty-mpm
+            // owns it, so remove the directory directly rather than refusing.
             warn!(
                 path = %path.display(),
-                "decommission: cannot determine repo root from worktree path — skipping git removal"
+                "decommission: no git repository owns this path — removing the \
+                 directory directly (no worktree ref to prune)"
             );
-            return false;
+            if let Err(e) = std::fs::remove_dir_all(path) {
+                warn!(path = %path.display(), "decommission: remove_dir_all failed: {e}");
+                return false;
+            }
+            return true;
         }
     };
+    let repo_root = repo_root.as_path();
     // Step 1: git worktree remove --force <path> (run from repo root).
     // Pass repo_root as an OsStr-safe Path arg to avoid lossy UTF-8 coercion (#1840).
     let out = std::process::Command::new("git")
--- a/crates/trusty-mpm/src/session_manager/mod.rs
+++ b/crates/trusty-mpm/src/session_manager/mod.rs
@@ -35,6 +35,7 @@ pub mod task_inject;
 pub mod workspace_guard;
 mod worktree_nested;
 pub(crate) mod worktree_ownership;
+pub(crate) mod worktree_registry;
 pub mod worktree_safety;
 
 #[cfg(test)]
--- a/crates/trusty-mpm/src/session_manager/prune.rs
+++ b/crates/trusty-mpm/src/session_manager/prune.rs
@@ -643,7 +643,8 @@ impl SessionManager {
     /// Dry-run returns after Phase 1 (no deletion, no snapshot).
     ///
     /// What: Phase 1 calls [`find_orphaned_worktrees`] inside `spawn_blocking`
-    /// (the filesystem walk is blocking); panics are propagated as `Err`. Phase 2
+    /// (git-derived since #4207, but still blocking — it spawns git per
+    /// project); panics are propagated as `Err`. Phase 2
     /// (real-delete only) takes ONE fresh `self.store` snapshot, then per candidate:
     /// canonicalize (skip on error — item 8), check against snapshot, apply the
     /// #3649 OWNERSHIP GATE (below), then call `remove_session_worktree` in its
--- a/crates/trusty-mpm/src/session_manager/prune.rs
+++ b/crates/trusty-mpm/src/session_manager/prune.rs
@@ -368,200 +368,51 @@ fn matches_filter(record: &SessionRecord, filter: PruneFilter) -> bool {
     }
 }
 
-/// Enumerate orphaned per-session worktree directories under `repos_root` (#1840).
+/// Enumerate orphaned worktree reclaim candidates under `repos_root` (#1840,
+/// rebuilt git-native #4207 slice 1).
 ///
-/// Why: extracted from `SessionManager::prune_orphaned_worktrees` so the
-/// walk logic can be tested independently of the full session-manager setup,
-/// and reused by the `doctor.rs` worktree health probe without duplicating the
-/// filesystem walk.
-/// What: walks all FIVE known worktree-store locations under each
-/// `<repos_root>/<owner>/<repo>/`:
-/// 1. the in-project shape (`.worktrees/<name>`, added #1840);
-/// 2. the clone-based shared-base-checkout shape
-///    (`.base/.worktrees/<session-id>`, added #3649);
-/// 3. Claude Code's native `isolation: "worktree"` agent shape at the repo
-///    root (`.claude/worktrees/<name>`, added #3971);
-/// 4. the SAME Claude Code agent shape, but created while working directly
-///    inside the bare `.base` clone checkout itself
-///    (`.base/.claude/worktrees/<name>`, added #3971 follow-up — the #3971
-///    fix covered only location 3, so this sibling location, and location 5
-///    below, were still invisible even after that fix landed; confirmed live
-///    on this machine with 29 real entries);
-/// 5. the SAME Claude Code agent shape again, but nested inside a per-session
-///    checkout (`.base/.worktrees/<session-id>/.claude/worktrees/<name>`,
-///    added #3971 follow-up — this is the exact shape this repo's own
-///    trusty-mpm sessions produce when a Claude Code agent spawns a nested
-///    worktree from within its own session checkout). Enumerated by listing
-///    each `.base/.worktrees` session leaf (via [`list_immediate_dirs`]) and
-///    probing it for a nested `.claude/worktrees`, since the session-id
-///    directory name is not known in advance.
+/// Why: this function used to WALK the filesystem, probing five hard-coded
+/// location shapes under each `<repos_root>/<owner>/<repo>/`. Every shape was
+/// added by a bug report about a location the previous shape list had missed
+/// (#3649, #3971, and the #3971 follow-up), and the list could never be
+/// complete because nothing stops a worktree from being registered anywhere on
+/// disk. Worse, the walk found DIRECTORIES, not worktrees, so it had no idea
+/// which checkout owned any of them — the grandparent guess that filled that
+/// gap left fourteen worktrees physically inside `.base` but registered to the
+/// parent repo permanently unreclaimable (#4207). Git maintains the registry;
+/// deriving from it deletes the whole category of missed-location bug.
+/// What: delegates discovery to
+/// [`super::worktree_registry::enumerate_registered_worktrees`] — every
+/// worktree git itself registers, wherever it lives — then removes any whose
+/// path is in `active_set`. Candidates come back canonicalized (so the
+/// active-set comparison is symlink-safe), sorted, and de-duplicated. A
+/// non-existent or unreadable `repos_root` yields an empty vec.
 ///
-/// Any leaf directory whose canonicalized path is NOT in `active_set` is
-/// collected as an orphan CANDIDATE — the caller (`prune_orphaned_worktrees`)
-/// still applies the #3649 ownership-sentinel gate before ever deleting
-/// anything, so a candidate discovered here is never auto-deleted unless it
-/// also carries a trusty-mpm ownership sentinel naming a provably-ownerless
-/// owner (see that function's doc); locations 3-5 are all Claude-Code-created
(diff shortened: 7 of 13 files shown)


bobmatnyc:
## Code Critic — Verdict: **WARN**

*(Posted as a review **comment**, not a `--request-changes` review: GitHub rejects a
formal review from the account that authored the PR. **This comment is the gate output
of record for PR #4224.** WARN = zero CRITICAL, one HIGH → the code may proceed, but the
HIGH must be carried forward, not discarded.)*

Independent adversarial review, no anchoring on the PR body's self-report. Everything
below was verified against real git on the dogfood machine and against a targeted
`-p trusty-mpm` run at `e5180c6d`.

### Verification performed

```
cargo test -p trusty-mpm --all-features --lib worktree
test result: ok. 136 passed; 0 failed; 0 ignored; 0 measured; 3633 filtered out; finished in 5.31s

cargo test -p trusty-mpm --all-features --lib
test result: ok. 3763 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 53.69s
```

CI on the PR is green across all 20 checks (Clippy, MSRV 1.91, Test 16m27s, 500-line cap,
SLD lint).

I also replayed the enumeration algorithm against the **real** `~/trusty-mpm-projects`
(4 owners, 83→206 registered worktrees) as a shell reimplementation, and rebuilt the
`git_worktree_list_agrees` fixture from scratch outside the repo to run the old rule and
the new rule side by side. All temporary fixtures were removed; `git worktree list` shows
**0** entries under my scratch path — no registry pollution.

---

## Answers to the six posed risks

### 1. Can anything now be enumerated for deletion that previously was not, and should not be? — **CONFIRMED (widening is real); no live path is newly auto-deletable**

The widening is real and larger than the PR states. Replaying both algorithms over the
live `repos_root`:

| | count |
|---|---|
| old five-shape walk | 203 candidates |
| new `git worktree list` derivation | 206 candidates |
| **newly enumerated** (matched no old shape) | **7** |
| of those 7, carrying a `.trusty-mpm-worktree` sentinel | **0** |

The 7 newly enumerated paths:

```
/Users/…/bobmatnyc/ft-follow-links
/Users/…/bobmatnyc/trusty-tools-main-build
/Users/…/bobmatnyc/trusty-tools-worktrees/wt-4029-subagents-section
/Users/…/bobmatnyc/trusty-tools-worktrees/wt-4171-l0-session-state
/Users/…/bobmatnyc/trusty-tools-worktrees/wt-deploy-trusty-agents-20260727
/Users/…/duettoresearch/code-intelligence/.worktrees/9cb1aa19-…/origin/fix/reindex-stream-false-negative
/Users/…/duettoresearch/gb-connect-card
```

Five of the seven are the operator's own long-lived checkouts. **None is deletable**, and
I traced each to the deletion call to confirm why: `prune.rs:735` classifies every
candidate by `read_sentinel_owner`, and `SentinelOwner::Unknown` → `owner_unknown`
(report-only, `prune.rs:742`) before `git_worktree_list_agrees`, the dirty gate, or
`remove_session_worktree` is ever reached. Only 24 of the 206 candidates carry a
sentinel, and every one of those sits at a `.worktrees/` shape the old walk already
covered.

Two other containment properties check out: `is_main`/`bare`/`prunable` exclusion is
correct (git always emits the main record first, verified live for both the normal
checkout and the bare `.base`), and `canonical.starts_with(canonical_root)`
(`worktree_registry.rs:296`) uses Rust's component-wise `starts_with`, so `/a/bc` cannot
match `/a/b`. 27 worktrees parked in `/tmp` and `/private/var` are excluded. The nested
parent/child pair in the list above is safe: `worktree_nested::nested_repo_roots`
(`worktree_nested.rs:347`) enumerates strict descendants at **any** depth from git's own
registry, so a parent cannot be force-removed over a dirty nested checkout.

**Why this is still the HIGH.** The PR states "Blast radius is unchanged" and that
enumeration "keeps the containment boundary the old walk had *structurally*". That is not
accurate. The old walk's structural boundary was `<repo>/{.worktrees, .base/.worktrees,
.claude/worktrees, .base/.claude/worktrees, .base/.worktrees/<id>/.claude/worktrees}` —
far tighter than `repos_root`. After this change, **a single `.trusty-mpm-worktree`
sentinel with a Known, ownerless owner makes a directory auto-deletable by the unattended
daemon GC (`reap_orphaned_worktrees`, `dry_run = false`) wherever it sits under
`repos_root`**, not only at those five shapes. The sentinel gate is now the *only*
structural boundary; the location boundary is gone.

The gate is real and tested (`prune_orphaned_worktrees_skips_owner_unknown`,
`prune_orphaned_worktrees_never_deletes_claude_native_worktree`) and the current machine
state is safe. But nothing in the new test set pins the *new* boundary — there is no test
asserting that a sentinel-bearing worktree at an arbitrary in-root location IS reclaimed,
which is what would make the widening explicit and regression-proof.

### 2. Can a git command failure be read as a fact? — **REFUTED for the new module**

Exhaustive audit of every fallible operation in `worktree_registry.rs`:

| site | construct | direction |
|---|---|---|
| `:163` | `.output().ok()?` | → `None` ✓ |
| `:165` | `!out.status.success()` → `return None` | ✓ |
| `:169` | empty stdout → `return None` | ✓ |
| `:174` | `common_dir.parent().map(...)` (may be `None`) | ✓ |
| `:196–200` | `.ok()?` + `!success` → `None` | ✓ |
| `:231` | `canonicalize(repos_root).unwrap_or_else(raw)` | failure makes `starts_with` reject everything → **shrinks** |
| `:274/:275` | `canonicalize(anchor/root).unwrap_or(...)` | failure ⇒ mismatch ⇒ early return → **shrinks** |
| `:293` | `let Ok(canonical) = canonicalize(&wt.path) else { continue }` | → **shrinks** |
| `:233/:241` | `let Ok(...) = read_dir(...) else { return/continue }` | → **shrinks** |

No `unwrap`, no `unwrap_or_default`, no `expect`, and no `Option`→`bool` coercion inside
the module. Every failure path strictly reduces the proposed set. The claim holds.

Two `Option`→verdict coercions exist at the **call sites**, both worth naming:

- `worktree_safety.rs:363–364` — `None` → `return true` (agree, deletion may proceed).
  Polarity is unchanged from the old code and is documented as an additional net rather
  than the fail-safe gate. Note the failure surface shifted from `git -C <grandparent>` to
  `git -C <candidate>`, but the narrowing in #1 actually *closes* the old hole: an
  unregistered husk outside any repo used to reach this gate and get a permissive `true`;
  it is no longer enumerated at all.
- `decommission.rs:167–183` — `None` → `remove_dir_all` + `return true`. See MEDIUM-4.
  Behaviourally equal to the pre-existing fallback (the old code reached `remove_dir_all`
  on every git spawn/exit failure at `:207`/`:219`), so this is **not** a new destructive
  path — but the comment states a verdict the probe cannot give.

### 3. `git rev-parse --git-common-dir` across the real topologies — **REFUTED (correct)**

Verified live, all four shapes:

```
git -C <repo>                    --git-common-dir → <repo>/.git   → root <repo>          ✓
git -C <repo>/.base              --git-common-dir → <repo>/.base  → root <repo>/.base    ✓ (bare)
git -C <linked-wt-under-.base>   --git-common-dir → <repo>/.git   → root <repo>          ✓ (the pathological set)
git -C <.claude/worktrees/agent> --git-common-dir → owning checkout                      ✓
```

Working directory is correct in every call path — `git_command` (`worktree_safety.rs:683`)
always passes `-C <dir>` and strips the nine `GIT_ENV_REDIRECTS`. Run inside a linked
worktree it returns the **main** repo's git dir, which is exactly the wanted answer; run
in the main checkout it returns `<root>/.git`; run in a bare clone it returns the bare
dir. The anchor-is-root confirmation at `worktree_registry.rs:271–278` is genuinely
load-bearing and correctly rejects `<owner>/ft-follow-links` (a repo_entry that is itself
a linked worktree) so it is not double-listed from its own anchor.

Residual (LOW-7): a repo cloned with `--separate-git-dir`, or a submodule anchor, yields a
common dir whose `file_name()` is neither `.git` nor a checkout root, so the anchor fails
the equality check and **all** of that repo's worktrees are silently skipped. Safe
direction, but silent.

### 4. The vacuous-test conversion — **PARTIALLY CONFIRMED**

The conversions themselves are genuine: every `mkdir` fixture is now a real
`git worktree add` via `GitWorktreeFixture`, and the fixture canonicalizes its tmp root
(`worktree_git_fixture.rs:85`) so git-reported paths compare cleanly. Spot-checked well
past three:

- `prune_orphaned_worktrees_fresh_active_set_blocks_deletion` (pot.rs:41) — **strong**:
  asserts `candidates.contains(&wt)` *first*, then asserts the fresh set empties it.
- `prune_orphaned_worktrees_collects_orphan` (pot.rs:65) — **strong**:
  `assert_eq!(orphans, vec![dead])` is an exact-set assertion.
- `reap_orphaned_worktrees_removes_orphan_preserves_live` — **strong**: real worktrees,
  asserts the orphan is removed *and* the live one preserved.
- `worktrees_with_orphan_is_warn` (doctor_tests.rs) — **strong**: asserts `Warn` and
  `"1 orphaned"`.

But the blanket claim "**each** converted test also asserts the candidate actually reaches
the gate under test" is **false** for at least four (see MEDIUM-5).

### 5. Do the revert-proofs bind? — **CONFIRMED, verified independently**

I rebuilt the fixture from scratch outside the repo and ran both rules:

```
candidate:   …/repos/owner/repo/.base/.worktrees/registered-to-parent
grandparent: …/repos/owner/repo/.base

OLD RULE  git -C <grandparent> worktree list --porcelain
  worktree …/repo/.base
  worktree …/repo/.base/.worktrees/owned-by-base
  => agrees FALSE   (candidate absent from the grandparent's registry)

NEW RULE  git -C <candidate> worktree list --porcelain
  worktree …/repos/owner/repo
  worktree …/repo/.base/.worktrees/registered-to-parent
  => agrees TRUE

test invariant: git -C <grandparent> rev-parse --git-common-dir → …/repo/.base
  (the grandparent IS a real repository, so the old code takes its SUCCESS path and
   returns false — it cannot pass by falling through to the best-effort `true`)
```

`git_worktree_list_agrees_true_for_worktree_registered_to_parent_repo` therefore fails
under the restored grandparent rule. **Binds.**

The author's explanation for why the bug survived is **confirmed**: both pre-existing
tests park the candidate at `<base>/.worktrees/<name>`, whose grandparent *is* the owning
repo, so both return the same answer under either rule and neither could ever have caught
this.

`find_orphaned_worktrees_discovers_worktree_at_unwalked_location` also binds — the
worktree sits at `<repo>/agents/scratch/wt-1`, which matches none of the five shapes, so
the old walk returns an empty list. `find_orphaned_worktrees_ignores_plain_directory`
binds too, in the opposite direction: the old `scan_worktree_shape` collected
`<repo>/.worktrees/just-a-mkdir` non-canonicalized, so `!orphans.contains(&fake)` fails.

### 6. The disclosed visibility regression — **CONFIRMED, acceptable, with a caveat**

Measured on the live machine: exactly **4** directories drop out of the report, and
**0** of them carry a sentinel — so the author's claim that no *reclaim* behaviour
regresses is correct, and I could not find a counterexample.

```
…/trusty-tools/.base/.worktrees/RETIRED-f443c12d-…-20260727-221502
…/trusty-tools/.base/.worktrees/build-63d3b64e-tagent
…/trusty-tools/.worktrees/tm-trusty-tools-01
…/duettoresearch/apex/.worktrees/scratchpad
```

Caveat, and it is the one place the "reports a state it does not have" failure mode
re-enters inverted: `doctor.rs check_worktrees` is the **only** surfacing mechanism for
this class, and #3715 (the husk-writer issue) is still open with no PR. These four are
real reclaimable disk that is now invisible everywhere. I judge this acceptable for this
slice — four husks is small, they were never reclaimable, and smuggling husk reclamation
in here would be worse — but it should not be left to #3715 indefinitely without at least
a counter.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_registry.rs` | 296 | Automatic-deletion surface widens from five named shapes to *every git-registered worktree under `repos_root`*; the sentinel gate is now the only structural boundary. Measured: 7 newly enumerated paths on the dogfood machine, 5 of them the operator's own long-lived checkouts. The PR's "Blast radius is unchanged" and "keeps the containment boundary the old walk had structurally" are inaccurate — the old boundary was five shapes, not `repos_root`. No test pins the new boundary. | Add a test that a sentinel-bearing, ownerless worktree at an arbitrary in-root location (e.g. `<repos_root>/<owner>/<repo>-worktrees/x`) IS reclaimed, so the widening is explicit and regression-proof; and correct the PR body / `find_orphaned_worktrees` doc to state the boundary that actually holds — `repos_root` + the sentinel gate. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/prune.rs` | 917 | Doc of `reap_orphaned_worktrees` — the **unattended** daemon GC entry point — still says the `"only leaf dirs under `.worktrees/`, never the base clone" guard` is "inherited unchanged". That guard no longer exists. A false safety claim on the one path that deletes without a human present. | Replace with the guarantees that do hold: containment to `repos_root`, `is_main`/`bare`/`prunable` exclusion, and the sentinel + ownerless + dirty gates. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_registry.rs` | 131, 283 | `locked` is named in the parser doc but never modelled or filtered. A git-`locked` worktree — the operator's explicit "do not remove this" — is enumerated like any other and then removed with `git worktree remove **--force**` (`decommission.rs:191`), which overrides the lock. The module deliberately models `bare`, `prunable`, and `is_main` but drops the one attribute git itself treats as a removal veto. The live machine has one locked worktree (`agent-a58a919ded2d5cfe1`). | Add `locked: bool` to `RegisteredWorktree`, set it on `locked` / `locked <reason>`, and `continue` on it in `collect_from_anchor` alongside `bare \|\| is_main \|\| prunable`. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/decommission.rs` | 167–183 | `registry_root_for` returning `None` is read as the fact *"No git registry claims this path — it is a plain directory, not a worktree"* and converted into `remove_dir_all` + `return true`. `None` equally covers a missing `git`, a non-zero exit, and empty stdout. Behaviourally identical to the old fallback, so not a new destructive path — but this is the one comment in the PR that asserts a verdict from an unanswerable probe, i.e. the exact anti-pattern the change is built to remove. | Reword to state what is actually known ("no owning checkout could be resolved — falling back to a direct directory removal, which the sentinel gate has already authorised; any worktree ref and session branch are intentionally left for `git worktree prune`"). Optionally distinguish spawn failure from a clean non-zero exit so a missing `git` is logged differently from "not a repository". |
| MEDIUM | `prune_orphan_tests.rs` / `doctor_tests.rs` / `worktree_registry_tests.rs` | pot.rs:26–36, pot.rs:221–231, doctor_tests.rs `worktrees_no_orphans_is_ok`, wrt.rs:207/221/240/261 | The claim "each converted test also asserts the candidate actually reaches the gate under test" does not hold. `prune_orphaned_worktrees_spares_active` asserts only `orphans.is_empty()`; `find_orphaned_worktrees_ignores_plain_directory` and `worktrees_no_orphans_is_ok` assert only a negative/`Ok`; and four `enumerate_*` negative tests (`ignores_plain_directory`, `excludes_the_main_checkout`, `excludes_worktrees_outside_the_repos_root`, `ignores_a_worktree_whose_directory_is_gone`) have no positive control. All seven pass unchanged if enumeration returns nothing at all — which is precisely how this whole module fails on an unsupported git flag or a regressed anchor check. The suite as a whole is saved only by the two `enumerate_finds_*` tests. | In each, also add a control worktree that MUST appear in the result (`assert!(found.contains(&control))`) alongside the negative assertion. |
| LOW | `crates/trusty-mpm/src/daemon/doctor.rs` | 384 | Stale doc: "spawns a blocking task to walk `<repos_root>/<owner>/<repo>/.worktrees/`; any leaf directory not in the active set is counted as an orphan." Neither clause is true after this change. | Rewrite to describe delegation to `find_orphaned_worktrees` / the git-derived registry. |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_git_fixture.rs` | 18, 30–32, 121 | Stale docs in the very fixture this PR extends: "the shape `find_orphaned_worktrees` walks", "the `git worktree list` cross-check resolves the repo root as the candidate's GRANDPARENT", "the reclaim path only ever considers leaf dirs under a `.worktrees`-shaped parent". All three describe the removed design. | Update alongside the module they document. |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_registry.rs` | 160 | `--path-format=absolute` requires git ≥ 2.31 (2021). On anything older, `registry_root_for` returns `None` for **every** anchor: enumeration silently yields zero candidates (the sweep becomes a no-op, invisible) and `remove_session_worktree` degrades to a bare `remove_dir_all` with no ref/branch cleanup, reintroducing the #1840 leak while reporting success. No version probe and no test covers it. | Either document the git floor next to the flag, or fall back to `rev-parse --show-toplevel` when the flagged form exits non-zero, distinguishing "flag unsupported" from "not a repository". |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_registry_tests.rs` | 127–133, 153–161 | `registry_root_for_non_repo_is_none` and `list_registered_worktrees_none_outside_a_repo` assume `tempfile::tempdir()` is not inside a git repository. True on this machine and in CI; false on any box whose `TMPDIR` sits under a checkout, where both would fail for an unrelated reason. | Assert the precondition explicitly, or place the non-repo directory beside the fixture's own canonicalized tmp root. |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_registry.rs` | 105–136 | `parse_worktree_list` splits on lines without `-z`; a worktree path containing a newline is split into two bogus stanzas. Contained (a bogus path fails `canonicalize` at `:293`, and `starts_with(canonical_root)` rejects an absolute injection), so the direction is safe — but the parse is not as total as the doc's "Unknown attribute lines are ignored so a future git version cannot break the parse" implies. | Note the limitation, or use `git worktree list --porcelain -z` (git ≥ 2.36) and split on NUL. |

## Required Changes (WARN — carry forward, do not discard)

1. **HIGH** — correct the containment claim and pin the new boundary with a test that a
   sentinel-bearing ownerless worktree at an arbitrary in-root location is reclaimed.
   Whoever reviews the next slice of #4207 must see that the location boundary was traded
   for the sentinel boundary, deliberately.
2. **MEDIUM** — fix `prune.rs:917`; it is a false safety claim on the unattended GC path.
3. **MEDIUM** — model and filter `locked` in `worktree_registry.rs`.
4. **MEDIUM** — reword `decommission.rs:169–172` so `None` is not stated as a fact.
5. **MEDIUM** — add positive controls to the seven listed tests.

## Notes

- The root-cause diagnosis is **correct and independently reproduced**. Both revert-proofs
  bind. The `--git-common-dir` derivation is right for every topology on this machine,
  including the bare `.base`, the parent-registered worktrees under `.base/.worktrees/`,
  and `.claude/worktrees/agent-*`. The anchor-is-root confirmation at `:271–278` is the
  load-bearing piece and it is correct.
- One factual nit in the PR narrative, not in the code: the fourteen pathological
  worktrees are mostly under `<repo>/.base/.claude/worktrees/`, not
  `<repo>/.base/.worktrees/`. The mechanism described is the same and the fix covers both.
- Explicitly **not** flagged: the returned-paths change from raw to canonicalized
  (`prune.rs:408–412`) — downstream compares against both canonical and raw forms
  (`prune.rs:866`), so it is safe; the extra git process spawns per repo (bounded at 4,
  on a ~60s daemon cadence); and `remove_session_worktree`'s use of a raw `Command`
  instead of the env-stripping `git_command` — pre-existing and untouched here.
- No source was modified. Every fixture I created was outside the repository and has been
  removed; `git worktree list` confirms **0** stray registrations.

bobmatnyc:
## VERDICT: WARN

Re-review of `dc924de6` (round 2, independent worktree at `.claude/worktrees/qa-4224-rereview`, detached at the PR head — the engineer's worktree was not entered). Prior round: WARN / 1 HIGH.

**The prior HIGH is genuinely resolved.** The containment rule is structural, not a path-shape check in different clothes, and I verified the reclamation-preservation claim independently rather than accepting it. One HIGH remains — the *same* defect class the prior HIGH named, left unfixed one call frame up, at the unattended-GC call site.

CRITICAL: 0 · HIGH: 1 · MEDIUM: 2 · LOW: 1

---

## Independent verification of the boundary

I replayed both enumerations against this machine's real `~/trusty-mpm-projects` (42 projects), reimplementing the porcelain parse, `--git-common-dir` root resolution, the anchor-is-root check, the filters, and component-wise containment outside the Rust code:

| measure | claimed | I measured |
|---|---|---|
| candidates without the boundary | 207 | **207** ✅ |
| candidates with the boundary | 201 | **201** ✅ |
| dropped by the boundary | 6 | **6** ✅ |
| dropped paths carrying a sentinel | 0 | **0** ✅ |
| `.base`-resident, before → after | 54 → 54 | **54 → 54** ✅ |
| sentinel-bearing, before → after | unchanged | **unchanged** ✅ |

The 6 dropped paths are exactly `bobmatnyc/ft-follow-links`, `bobmatnyc/trusty-tools-main-build`, three `bobmatnyc/trusty-tools-worktrees/wt-*`, and `duettoresearch/gb-connect-card`. **None carries a `.trusty-mpm-worktree` file of any kind**, so the boundary provably costs zero reclaim capability. The narrowing does not give back the fix.

On the "3 sentinel-bearing": I count **25** paths with a sentinel *file*, of which **4** parse into `SentinelOwner::Known` (the only class that can ever be deleted); the other 21 are empty/legacy → `owner_unknown` → never removed. So the PR's "3" is the Known count, off by one against today's tree (drift, or this session). The *invariant* the number was cited for — unchanged before vs. after — holds exactly, because the delta set is sentinel-free entirely.

### The other direction: nothing `main` could reclaim is lost

`main`'s five-shape walk yields 208 candidates here. 8 are absent from the new set (`.claude/worktrees/agent-*`, a `RETIRED-*` husk, `build-63d3b64e-tagent`, `tm-trusty-tools-01`, `apex/.worktrees/scratchpad`). **Zero carry a parsable sentinel**, so `main` could not have deleted any of them either. The SCOPE NOTE at `prune.rs:406-411` is accurate.

One of `main`'s 8 is `/Users/masa/Duetto/cto-resources` — reached via the real symlink `bob-duetto/cto/.worktrees/cto-resources`. `main`'s walk canonicalizes through it and proposes a path **entirely outside `repos_root`**, with no outer bound at all. The PR excludes it twice. The boundary is a net safety improvement over `main`, not merely a restoration.

### Component-wise `starts_with` (item 2)

Confirmed at `worktree_registry.rs:401-404` — the operands are `PathBuf`, so `Path::starts_with` is component-wise. I ran the simplification mutation implicitly via M5 below and separately confirmed `<owner>/trusty-tools-worktrees/wt-*` is dropped in the live 6. `enumerate_excludes_a_worktree_parked_beside_the_project` (`worktree_registry_tests.rs:327`) is a real guard against a `to_string_lossy()` rewrite, and it carries a positive control in the same call so it cannot pass on an empty result.

### Provenance / branch-role finding (item: is the measurement sound?)

**Sound, and I reproduce it.** Bucketing by prefix (`fix/` *and* `fix-…`, etc.) over the 201: `fix` 88, `feat` 44, `other` 28, `docs` 12, detached 10, `session` 10, `chore` 9 — matching the PR's histogram exactly except `session` 10 vs. 9 (one worktree of drift). A `session|ws|workstream` gate admits **10 of 201**.

One correction: the PR says such a gate would admit **zero** of the sentinel-bearing candidates. It admits **1 of the 4** Known-owner ones (`trusty-tools/.worktrees/tm-trusty-tools-02`, branch `session/tm-trusty-tools-02`); the other three sit on `feat/bakeoff-runner-layer`, `fix/errors-insights-query`, `feat/round-auto-create`. That weakens the rhetoric, not the conclusion: the gate would still drop **3 of 4** reclaimable candidates. Reclamation would collapse while alarms read healthy — the exact failure shape being guarded against. The finding that git records no provenance is independently correct: `git worktree list --porcelain` carries no creator, and neither does `.git/worktrees/<id>/`.

**Does the pre-1.3.0 population weaken it for the future?** It weakens the *histogram* argument and strengthens the *containment* argument. Under one-worktree-per-writing-agent the `session|ws` share would rise, so today's 10/201 is not predictive. But the whole point is that a name-based gate is only ever as good as the naming convention of the day, and 1.3.0 explicitly changes that convention. Containment is convention-independent: every provisioned agent worktree lands inside `<repos_root>/<owner>/<repo>/` regardless of count, lifetime, or name, and operator checkouts are excluded by position. The conclusion survives the population change; the supporting histogram does not, and the doc at `worktree_registry.rs:276-291` should not be read as future-proof evidence.

### Fail-open / cursor-advance sweep

None found in the new module. Every failure branch in `collect_from_anchor` (`worktree_registry.rs:361-377`) and `enumerate_registered_worktrees` (`:306-336`) returns/`continue`s, which can only *shrink* the proposal set. `canonicalize(repos_root)` falling back to the raw path (`:307`) also shrinks (canonicalized candidates stop matching an unresolved root). The skip-whole-project-on-canonicalize-failure at `:327-329` is the correct asymmetry and is documented as such.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/daemon/mod.rs` | 661-663 | The unattended orphan-GC call site still asserts the removed shape guard: *"Only leaf worktree dirs with NO live record are removed — active worktrees and the base clone are never touched."* This is the identical false-safety-claim defect the PR corrected in `prune.rs:931-955`, left in place at the one call frame that actually deletes with no human present. It is now false: `enumerate_finds_worktree_at_an_unwalked_location` proves a registered worktree at `<repo>/agents/scratch/wt-1` — not a `.worktrees/` leaf — is a candidate. ("the base clone is never touched" happens to still hold, but via the `bare`/`is_main` filters, not the leaf rule stated.) | Replace with the six gates now in force, or point at `prune.rs`'s `reap_orphaned_worktrees` "What actually bounds THIS path" block rather than restating a guarantee that no longer exists. |
| MEDIUM | `crates/trusty-mpm/src/daemon/doctor.rs` | 380-385 | Stale: *"The filesystem walk is delegated to … find_orphaned_worktrees"* and *"spawns a blocking task to walk `<repos_root>/<owner>/<repo>/.worktrees/`; any leaf directory not in the active set is counted as an orphan."* There is no walk, it is not confined to `.worktrees/`, and a plain leaf directory is explicitly **not** counted (`find_orphaned_worktrees_ignores_plain_directory`). Report-only path, hence MEDIUM not HIGH — but the PR edited `doctor_tests.rs` for exactly this change and did not update the function it tests. | Rewrite as "delegates to `find_orphaned_worktrees`, which derives candidates from git's registry; a directory git does not register is not a candidate." |
| MEDIUM | `worktree_registry.rs` / `worktree_registry_tests.rs` / `worktree_git_fixture.rs` / `CHANGELOG.md` | 382 / 407 / 182 / 37 | The stated rationale for the `locked` filter is factually wrong. `git worktree remove --force` does **not** override a lock. Verified on git 2.54.0: `fatal: cannot remove a locked working tree; use 'remove -f -f' to override or unlock first`, exit 128, directory intact. The **conclusion is still right**, but for a different reason: `decommission.rs:199-212` catches the non-zero exit and falls back to `std::fs::remove_dir_all(path)`, which destroys the locked worktree anyway (and orphans its registry entry). The risk of leaving this as written is concrete — a future engineer who checks the git man page will find the comment false and may conclude the enumeration filter is redundant, deleting the only place the lock is honoured. | State the real mechanism: "`git worktree remove --force` *refuses* on a locked worktree, but `remove_session_worktree` falls back to `remove_dir_all` on any git failure, so the lock is defeated at removal regardless. Enumeration is the only place it can be honoured." Same correction in the CHANGELOG bullet. |
| LOW | `worktree_registry_tests.rs` | 356-362 | The test doc says the `!=` "is what keeps the operator's project checkout unselectable". Today `is_main` is what keeps it unselectable; the `!=` is defence-in-depth for a topology (`<repo>` registered as a *linked* worktree of its own `.base` bare clone) that is reachable in principle but not present — I checked both `.base` registries on this machine, 0 hits. Soften the claim to match. | s/"is what keeps"/"is the second, independent reason"/ |

## Mutation assessment (run by me, not accepted from the description)

- **M2 (`!=` strictness).** Deleting it → exactly **1 failure**: `collect_from_anchor_never_yields_the_containment_boundary_itself` (355 passed, 1 failed). The disclosed masking is real and the disclosure is honest. My honest read: this is a *legitimate contract test of a private function*, not a test constructed to fake coverage — but it is closer to that line than the others, because the production topology it defends against is not currently instantiated anywhere I can find. Keep the guard (it is one comparison and it is correct); fix the overstated test doc (LOW above).
- **M5 (over-narrow to require a `.worktrees` component).** → **4 failures** across both layers: `enumerate_finds_worktree_at_an_unwalked_location`, `find_orphaned_worktrees_discovers_worktree_at_unwalked_location`, `prune_orphaned_worktrees_reclaims_owned_husk_at_an_unwalked_location`, `prune_orphaned_worktrees_never_deletes_claude_native_worktree`. It genuinely proves *that* narrowing is caught. It does not generalise to "any re-narrowing is caught" — a depth cap, for instance, would pass. The stronger guarantee is the positive-control discipline (`worktree_registry_tests.rs:275-278`): every exclusion test asserts an inclusion in the same call, so no exclusion test can pass on an empty enumeration. That is what actually carries the load, and it holds.

## Gate anomaly

Fair characterization. `git diff --stat origin/main...dc924de6 -- '*telegram*'` is **empty** — zero telegram files touched. The test asserts `backoff_component < Duration::from_millis(20)` (`telegram/supervisor_tests.rs:171-175`) against a real wall-clock gap whose true value is ~2ms base + a 5ms sleep overshoot, under 3,762 concurrently-running tests. >13ms of tokio scheduler jitter under that load is unremarkable. Unrelated to this change; it is a pre-existing wall-clock-sensitive test that deserves its own issue (make it deterministic with `tokio::time::pause`, or assert an ordering relation instead of an absolute millisecond bound) rather than a re-run.

## Verification evidence

```
$ cargo test -p trusty-mpm --lib session_manager::worktree_registry
test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 3749 filtered out; finished in 0.61s

$ cargo test -p trusty-mpm --lib session_manager::
test result: ok. 356 passed; 0 failed; 0 ignored; 0 measured; 3413 filtered out; finished in 4.44s

$ cargo test -p trusty-mpm --lib daemon::doctor
test result: ok. 103 passed; 0 failed; 0 ignored; 0 measured; 3666 filtered out; finished in 0.66s
```

`gh pr checks 4224`: **all 22 checks passing**, 1 skipping (`Notify on red main`). Nothing pending, nothing failing — including `Test` at 15m42s, which did not reproduce the telegram flake.

1. `daemon/mod.rs:661-663` — remove the false leaf-shape guarantee at the unattended-GC call site (the HIGH).
2. `daemon/doctor.rs:380-385` — correct the stale "filesystem walk of `.worktrees/`" description.
3. Correct the `git worktree lock` rationale in all four places to name the `remove_dir_all` fallback as the actual reason the lock cannot survive removal.
4. (optional) `worktree_registry_tests.rs:356-362` — soften the `!=` claim.

All four are documentation-only. **The code is sound and I found nothing to change in it.** Once (1) lands, this is an APPROVE.

bobmatnyc:
## VERDICT: WARN — 1 HIGH (new false claim), at head `deb3e1dd`

Confirmation pass, scoped as asked: I did **not** re-review the boundary logic, the reclamation numbers, or the git-provenance finding. I verified programmatically that `dc924de6..deb3e1dd` changes **zero non-comment Rust lines** — every changed `.rs` line is a `//`/`///` comment, the rest is CHANGELOG — so my round-2 test evidence still applies bit-for-bit.

**Prior HIGH (`daemon/mod.rs` leaf-shape claim): resolved.** **Prior MEDIUMs: both resolved**, and the lock rewrite is *better* than what I asked for. **But the new `daemon/mod.rs` prose contains a fresh false attribution**, and it is in the parenthetical that replaced the old false claim. This is the failure mode you named. Details and a one-clause fix below.

CRITICAL: 0 · HIGH: 1 · MEDIUM: 1 · LOW: 2

---

## HIGH — `crates/trusty-mpm/src/daemon/mod.rs:670-672`

```
// strict descendant of the managed project whose registry named
// it (so operator checkouts and the base clone are unreachable by
// position, not by name);
```

**`<repo>/.base` is a strict descendant of `<repo>`.** It is positionally *inside* the project, so gate 2 does not exclude it — gate 2 is the one gate that would *admit* it. The operator-checkout half of the clause is correct and I re-confirmed it; the base-clone half is false.

Demonstrated, not argued. I registered `<repo>/.base` as a linked worktree of `<repo>` on git 2.54.0 and replayed `collect_from_anchor`'s exact predicate over it:

```
strict descendant (component-wise starts_with + !=): True
ENUMERATED AS A RECLAIM CANDIDATE: …/repos/owner/repo/.base
ENUMERATED AS A RECLAIM CANDIDATE: …/repos/owner/repo/.worktrees/s1
```

The base clone passes containment and lands on the candidate list. In that topology the only thing left between it and `remove_dir_all` is the #3649 sentinel — which is precisely the "one gate happens to hold" situation this PR's own containment rule was written to end (`worktree_registry.rs:244-246`).

**Is the base clone actually at risk today? No.** In the provisioned topology `.base` is a `git clone --bare`, so it is absent from `<repo>`'s registry and is `bare` + `is_main` in its own — excluded by **gate 1**. My round-2 measurement over 42 real projects bears this out: `.base` itself never appears in either the 207 or the 201.

So the *outcome* ("the base clone is never touched") is true; the *attributed mechanism* is wrong, and it is wrong in the permissive direction. That matters concretely here: a future engineer hardening `.base`, or changing how `.base` is provisioned, will read this line at the unattended-GC call site and conclude position already covers it. It does not. This is the same defect shape as the round-2 HIGH — a gate credited with protection it does not provide — restated in a new costume, exactly as you predicted.

**Fix (one clause):**

```
// strict descendant of the managed project whose registry named
// it (so an operator checkout parked beside or above a project is
// unreachable by position, not by name; the base clone is instead
// unreachable via the previous conjunct — it is `bare` and `is_main`
// in its own registry and absent from the project's);
```

### The other five conjuncts are accurate — each pinned to a line

| conjunct in the comment | code | verdict |
|---|---|---|
| "not main/bare/prunable/locked" | `worktree_registry.rs:396` | ✅ |
| "strict descendant … and lies under repos_root" | `worktree_registry.rs:414-416` | ✅ (parenthetical is the defect above) |
| "neither the initial nor the pre-deletion active set" | `prune.rs:733` (initial) + `prune.rs:876` (fresh) | ✅ both, correctly |
| "#3649 sentinel naming a provably-ownerless owner past the grace window" | `prune.rs:745`, `:755` | ✅ |
| "#4091/#4118 dirty gate … no uncommitted or unpushed work" | `prune.rs:773` (scan) + `:887` (pre-removal) | ✅ both phases |

The list **omits** gate 5, `git_worktree_list_agrees` (`prune.rs:764`). Omission under-claims safety and the comment defers explicitly ("See `SessionManager::reap_orphaned_worktrees` for the authoritative list"), so I am **not** flagging it — noting it only so the omission is a recorded decision rather than an oversight.

**The disproof is real.** `<repo>/agents/scratch/wt-1` is exactly the path in `worktree_registry_tests.rs:210` and `prune_orphan_tests.rs:204` (and `:402` for the husk variant), and those tests pass. Keeping it inline is the right call — it makes reinstating the old claim fail a test.

---

## Item 3 — the `locked` rewrite: **accurate, and understated**

All four sites now say git *refuses* a single `--force` and the danger is `remove_session_worktree` reading that refusal as a generic failure and falling back to `remove_dir_all`. Verified end-to-end on git 2.54.0:

```
$ git worktree remove --force <locked>
fatal: cannot remove a locked working tree;
use 'remove -f -f' to override or unlock first        (exit 128)
$ rm -rf <locked>                    # the fallback at decommission.rs:207
$ git worktree list --porcelain
worktree …/.worktrees/w1
locked                               # entry survives
$ ls .git/worktrees/ → w1            # admin dir survives
$ git worktree prune -v              # exit 0 — and prunes NOTHING
$ ls .git/worktrees/ → w1            # still stranded
```

**The new "strands git's registry entry" claim is correct and stronger than written**: the entry is not merely left behind pending a later prune — it is *permanently* unprunable, because `git worktree prune` skips locked entries. And `remove_session_worktree` never even attempts the prune, since Step 2 is gated on `git_success` (`decommission.rs:226`), which is `false` on that branch. Optional: say "permanently stranded — `git worktree prune` skips locked entries". Not required.

## Item 4 — the strictness doc: **accurate**

"On every topology observed today it is `is_main`, not this `!=`" ✅ — matches my round-2 measurement (0 of 4 `.base` registries list a project root as a non-main record). "Reachable in principle: a project directory that is itself a NON-main linked worktree of some other registry" ✅ — and I traced it: in that topology the `<repo>` anchor is skipped by the anchor-is-root check (`registry_root_for` returns `.base`), the `.base` anchor lists `<repo>` as non-main, and the `!=` is then genuinely the only exclusion. The characterization is now precise and the overstatement I flagged is gone.

## Item 5 — `prune.rs:646`: **accurate**

"git-derived since #4207, but still blocking — it spawns git per project" ✅. Two synchronous `std::process::Command::output()` calls per anchor (`registry_root_for` + `list_registered_worktrees`), two anchors per project. `spawn_blocking` is still required for the right reason.

## Item 6 — RULING on `decommission.rs:313` / `:411`: **they are right; leave them**

I checked this rather than taking their word, and their reasoning holds.

Both claims sit inside the `!record.workspace_owned` branch and are scoped to it. That branch reaches `remove_session_worktree` **only** through `if is_session_worktree(ws)` (`decommission.rs:411-413`), and `is_session_worktree` (`:93-98`) tests `path.parent().file_name() == ".worktrees"` on `record.workspace_path`. That is a genuinely shape-based check on **store data**, on a completely separate input path from the orphan enumeration — nothing about #4207 touches it, and it is not in this PR's diff. `<base>` has parent `<repo>`, not `.worktrees`, so `is_session_worktree(<base>)` is `false` and the else-branch warns and skips. **"The base clone is NEVER touched — only the per-session leaf dir" remains true there.**

The `workspace_owned == true` branch does `remove_dir_all` on a wider set, but it is separately gated by `is_safe_to_remove` and is separately and accurately documented at `:299-308`. No path I can find lets decommission reach a non-leaf worktree. **Not a HIGH. No change needed.**

Worth recording, because it is the substantive point: `decommission.rs` may keep a shape-based safety claim precisely *because* it is still shape-driven, while `daemon/mod.rs` may not, because it no longer is. The engineer drew that line correctly.

---

## Remaining findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/daemon/doctor.rs` | 424-425 | This commit rewrote the doc comment to say "this is no longer a filesystem walk of `.worktrees/`" — then left the inline comment 40 lines below in the same function reading *"Delegate the blocking filesystem walk to spawn_blocking so the async executor is not stalled on directory I/O (#1840 F)."* The function now contradicts itself, and it is the same fix-one-copy-leave-the-other pattern that produced this round. `prune.rs:646` got exactly this correction; its sibling did not. | Mirror `prune.rs:646`: "Delegate the blocking git subprocesses to spawn_blocking (git-derived since #4207) …". |
| LOW | `crates/trusty-mpm/src/daemon/doctor.rs` | 386-389 | "every git-registered worktree that is a strict descendant of the project whose registry named it, minus the active set, is counted as an orphan" over-states the count: main/bare/prunable/locked records are excluded, as is anything outside `repos_root`. Report-only path and in the over-reporting direction, hence LOW. | Add "…that is not main, bare, prunable or locked…". |
| LOW | `crates/trusty-mpm/src/daemon/doctor.rs` | 394 | "reclaiming unregistered husks is #3715, still open" — #3715 **is** open, so that half is true, but it is not the husk issue: `gh issue view 3715` is *"PM session lost its trusty-mpm agent catalog mid-session…"*, and its body is entirely about `--setting-sources` and the agent roster. **Pre-existing** — carried over from `prune.rs:411` on the previous head, not invented this round — so noting, not gating. | Point at the correct issue, or drop the number and say "tracked separately". |

## Verification evidence

```
$ git diff dc924de6..deb3e1dd -- '*.rs' | (non-comment lines)
   NONE — every changed .rs line is a comment.
```

Code is byte-identical to `dc924de6`, where I measured:
```
session_manager::worktree_registry  → 20 passed;  0 failed
session_manager::                   → 356 passed; 0 failed
daemon::doctor                      → 103 passed; 0 failed
```
`gh pr checks 4224`: green, per the coordinator and consistent with my round-2 read.

## Required before merge

1. **`daemon/mod.rs:671-672`** — drop or re-attribute the "and the base clone" clause (the HIGH). One clause.
2. `doctor.rs:424-425` — the inline comment now contradicts the doc comment this commit wrote.
3. (optional) the two LOWs in `doctor.rs`.

Item 1 is the only thing gating. Everything else on your list — the `daemon/mod.rs` gate enumeration minus that clause, all four `locked` sites including the new registry-stranding claim, the strictness doc, `prune.rs:646`, and the decision to leave `decommission.rs` alone — **verified true**. Fix the clause and this is an APPROVE; I will not ask for anything further.

bobmatnyc:
## VERDICT: APPROVE — at head `d509c3d0`

**The round-3 HIGH is resolved. Both MEDIUMs and both LOWs are resolved. No new false claim survives.** Zero CRITICAL, zero HIGH.

Scoped exactly as asked: I verified only the corrected clause and the `doctor.rs` follow-ups. I re-confirmed programmatically that `dc924de6..d509c3d0` changes **zero non-comment `.rs` lines**, so the boundary code remains byte-identical to what I replayed and measured at `dc924de6` — not re-measured, not re-litigated.

---

## The one clause — `daemon/mod.rs`

The correction attributes base-clone exclusion to **gate 1** and adds an explicit inline warning against re-attributing it to gate 2. Both halves are true.

### (a) Gate 1 really is the mechanism — verified, not accepted

I spot-checked more than the one you asked for, and went one step past what the engineer reported. Across every project on this machine that has a `.base`:

**a1 — "absent from `<repo>`'s registry entirely":**
```
bobmatnyc/trusty-tools      -> records ending in .base in <repo>'s registry: 0
duettoresearch/duetto       -> records ending in .base in <repo>'s registry: 0
```

**a2 — "bare+main in its own":**
```
--- bobmatnyc/trusty-tools/.base
    worktree /Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.base
    bare
--- duettoresearch/duetto/.base
    worktree /Users/masa/trusty-mpm-projects/duettoresearch/duetto/.base
    bare
```
`.base` is the **first** record (so `parse_worktree_list` sets `is_main = true`, `worktree_registry.rs:122`) **and** carries the `bare` attribute line. Gate 1's `wt.bare || wt.is_main` (`worktree_registry.rs:396`) excludes it twice over.

**a3 — the load-bearing fact behind a1, which the engineer did not report and I checked myself.** "Absent from `<repo>`'s registry" is only true if `.base` is a genuinely separate repository rather than a linked worktree:
```
bobmatnyc/trusty-tools/.base   core.bare=true  common-dir=…/trusty-tools/.base
duettoresearch/duetto/.base    core.bare=true  common-dir=…/duetto/.base
```
`--git-common-dir` resolves to `.base` itself, not to `<repo>/.git` — it is its own repository. That is *why* it never appears in `<repo>`'s registry, and it makes a1 a structural consequence rather than a coincidence of current state.

**Third reported confirmation, checked against my own data:** of the 6 paths gate 2 drops on this machine (`ft-follow-links`, `trusty-tools-main-build`, 3× `trusty-tools-worktrees/wt-*`, `gb-connect-card` — my round-2 measurement), **none is a `.base`**. Consistent.

### (b) The inline negative statement about gate 2 is correct

> "note this gate does NOT exclude `.base`, which IS a strict descendant of `<repo>`, so do not re-attribute the base clone to it"

This is exactly what I demonstrated in round 3: `<repo>/.base` satisfies component-wise `starts_with` + `!=`, and when I registered it as a linked worktree of `<repo>` the predicate replay enumerated it as a reclaim candidate. Writing the negative *inline*, next to the gate it disclaims, is the right structural choice — it makes the mistake I found hard to reintroduce by someone skimming for a one-line summary.

### (c) Nothing else drifted

I read the full final comment block rather than the diff hunk. All six conjuncts survive intact and in order: gate 1 (main/bare/prunable/locked) → gate 2 (strict descendant) → gate 3 (neither the initial nor the pre-deletion active set) → gate 4 (#3649 sentinel past grace) → gate 5 (#4091/#4118 dirty gate) → the deferral to `SessionManager::reap_orphaned_worktrees` as authoritative. The `<repo>/agents/scratch/wt-1` disproof is intact, and it still matches the live test paths (`worktree_registry_tests.rs:210`, `prune_orphan_tests.rs:204`).

The operator-checkout half is now scoped to "checkouts parked beside a project", which is accurate for all 6 paths gate 2 actually drops. It does not claim to be gate 2's *only* effect, so the unmentioned `starts_with(canonical_root)` bound is an omission in the conservative direction — same as the `git_worktree_list_agrees` omission I recorded last round, and same deferral covers it.

**Not a finding:** the edit left a ragged wrap ("it is in neither / the initial nor the / pre-deletion active set" breaks early). Cosmetic only — `cargo fmt` does not reflow comments and the format check is green. Reflow it if you happen to be in the file; do not spin a round for it.

---

## `doctor.rs` — MEDIUM and both LOWs resolved

- **MEDIUM (self-contradiction) — fixed.** The inline comment now reads *"Since #4207 what blocks is not directory I/O but the git subprocesses `find_orphaned_worktrees` spawns per managed project."* Accurate (two synchronous `Command::output()` per anchor, two anchors per project) and no longer contradicts the doc comment above it. It now matches the `prune.rs:646` wording, so the two siblings agree.
- **LOW (over-broad count) — fixed.** Now *"counts as an orphan when it is not main/bare/prunable/locked, IS a strict descendant of the project whose registry named it, and is absent from the active set."* Matches `worktree_registry.rs:396`/`:414-416` and `prune.rs:423`.
- **LOW (`#3715` mis-citation) — fixed correctly.** The number is gone (`grep -c 3715` over the file returns **0**) and replaced with *"a separate open concern, deliberately out of scope for #4207"* — no substitute number guessed. That is the right call: an honest unnumbered pointer beats a confidently wrong one, which is the failure this branch spent three rounds on.
- The two unprompted statements I checked in round 3 — REPORT-ONLY, and that it deliberately reports candidates the reclaim path would refuse — both survive verbatim and remain true.

---

## Evidence

```
$ git diff dc924de6..d509c3d0 -- '*.rs'   (non-comment lines)
   NONE — comment-only, code byte-identical to dc924de6.
```

Code unchanged since `dc924de6`, where I measured:
```
session_manager::worktree_registry  → 20 passed;  0 failed
session_manager::                   → 356 passed; 0 failed
daemon::doctor                      → 103 passed; 0 failed
```
CI green.

---

## Closing note on the branch, not the clause

Three rounds, three false doc claims, each found in the text written to replace the last one — that pattern is worth naming, but it is worth naming *accurately*. Every one of them was a **misattribution of which gate provides a protection that genuinely exists**, never a missing or broken gate. The enumeration boundary itself has been correct and unchanged since `dc924de6`: I replayed it against 42 real projects, confirmed 207→201 with all 6 dropped paths sentinel-free, and confirmed nothing `main` could reclaim is lost. The code was right the first time; the prose describing it needed three passes.

The thing that finally broke the cycle is in this commit and is worth keeping: the correction does not just state the right mechanism, it states the wrong one *and marks it wrong*, inline, at the gate that invites the error. That is a durable fix rather than a corrected sentence.

**APPROVE. Nothing further from me.**

Issue #3617, opened 2026-07-21
PHASE 1: slack-mcp Search/discovery tools (emojis, users scope)


bobmatnyc:
## Parent Epic
Blocked on: #3611

## Tools to Implement

Implement these search and discovery tools in the slack-mcp server:
- `slack_search_emojis` — search emoji by name or category
- `slack_search_users` — search workspace users (with scope parameters for visibility)

### slack_search_users
Currently `slack_list_users` exists as a simple list operation. The claude.ai connector provides `slack_search_users` as a search operation. Design decision needed:
- Extend `slack_list_users` with optional search/filter parameters, OR
- Implement separate `slack_search_users` tool

### Message search scope
Note: The claude.ai connector splits message search into:
- `slack_search_public` (public channels only)
- `slack_search_public_and_private` (all accessible)

While the internal implementation has a single `slack_search_messages`. Triage decision: should this stay as-is or be split into public/private variants? Clarify in Phase 2 scope review.

## Implementation Sites

- `crates/trusty-channels/src/slack/tools.rs` — add tool definitions
- `crates/trusty-channels/src/slack/server.rs` — register handlers
- `crates/trusty-channels/src/slack/handlers/` — implement handler functions

## References

Follow patterns from existing 9 tools. Reference:
- `slack_search_messages` (existing search)
- `slack_search_channels` (existing search)
- `slack_list_users` (existing list)

## Slack API Scopes

Determine required scopes and document during implementation.

## Notes

This is part of Phase 1 (tool parity). Phase 2 (hard-block) is blocked on all Phase 1 tools reaching parity.

⚠️ TRIAGE: `slack_search_emojis` is relatively low value. Confirm during Phase 1 scoping whether this tool is worth building.

Issue #2996, opened 2026-07-18
trusty-channels/slack-mcp: no cursor pagination on read_channel/read_thread; verify read_thread + search failures (bot-only token)


bobmatnyc:
**Component:** `crates/trusty-channels` (binary `slack-mcp`), Slack MCP tools. Related: epic #2636, PR #2776.

**Discovered** attempting a 2-year history + thread extraction of a busy Slack channel from a trusty-mpm session (bob-duetto/cto, tm-cto-01, 2026-07-18). Only the newest ~1000 main-channel messages were reachable and zero thread replies.

**Environment note:** the running binary `/Users/masa/.cargo/bin/slack-mcp` was built 2026-07-15 ~15:34, which **predates PR #2776 (merged 2026-07-16)** by ~10h. Some observed behavior may already be addressed on `main` — please confirm against current `main` and, if so, the fix here may be largely "rebuild/reinstall the binary" plus the remaining gap below.

**Issues observed (please triage real-bug vs known-limitation vs config):**

1. **No cursor pagination on `slack_read_channel`** — exposes only `channel`/`limit`; returns a single page. `limit=1000` returned the newest 1000 msgs (oldest reachable 2026-03-19); higher limits collapsed to ~100 (clamp). The `tools.rs` doc string already acknowledges the single-page bound, and PR #2776 clamps limits — but **there is still no way to page backward through full history**. Request: cursor-based pagination (or `oldest`/`latest` time-window params) so historical extraction is possible.

2. **No cursor pagination on `slack_read_thread`** — same single-page limitation; long threads truncate. In this session every `slack_read_thread` call failed outright with `Slack API error: invalid_arguments` across 4 arg-format attempts. `conversations.replies` is bot-token-capable, so this is likely a real arg-handling/format bug rather than a token issue — please verify with a repro and confirm the expected argument shape (parent `ts` + channel).

3. **`slack_search_messages` fails with `invalid_arguments`** — this is **likely a config issue, not a bug**: `search.messages` rejects bot tokens and requires `SLACK_USER_TOKEN` (two-token model, see `handlers.rs`). Suggest: (a) confirm this is the cause, and (b) if so, improve the error message to say "search requires SLACK_USER_TOKEN" instead of a raw `invalid_arguments`, so callers can self-diagnose.

**Impact:** blocks historical + thread-inclusive Slack extraction for CTO analysis work (e.g. the #war-room best-practices guide, which could only cover ~4 months).

[...]

PR #4242, opened 2026-07-28
fix(trusty-channels): clamp search.messages count to Slack's documented 100

bobmatnyc:
## The defect (confirmed against Slack's docs, not inferred)

`slack_search_messages` clamped `count` with `clamp_page_size(n, MAX_PAGE_SIZE)` where `MAX_PAGE_SIZE = 1000`, whose doc comment claimed 1000 "matches the widest window Slack's `search.messages` accepts."

That claim is false. Slack's reference for the method states:

> Pass the number of results you want per "page". **Maximum of `100`.** _Default:_ `20`

and, under Usage info:

> Please note that the max `count` value is `100` and the max `page` value is `100`.

— <https://docs.slack.dev/reference/methods/search.messages> (retrieved 2026-07-28)

So any `count` in **101..=1000** passed local validation, was forwarded verbatim, and was rejected live by Slack with `{"ok": false, "error": "invalid_arguments"}`. `interpret_envelope` maps that to `SlackError::Api("invalid_arguments")`, whose `Display` forwards Slack's slug unchanged — which is exactly the opaque error reported from a live CTO workstream session on every query.

`page` was checked too: it has the same documented max of 100, but this handler never sends `page` at all, so there is no matching bug. (It also means results are capped at one page of 100 — a real limitation, but a feature gap, not this defect. Not addressed here.)

## The fix

`MAX_PAGE_SIZE` had **exactly one** production consumer — `search.rs`. The `conversations.*` family already uses `MAX_CONVERSATION_PAGE_SIZE = 999`. So rather than adding a parallel constant and leaving a dead, mis-documented 1000 behind, `MAX_PAGE_SIZE` is renamed to `MAX_SEARCH_COUNT = 100`. The two ceilings remain deliberately separate constants, each citing its own doc source.

**Clamp, not reject.** An over-large `count` is reduced to 100 rather than erroring. Reasoning: Slack caps the page at 100 either way, so clamping returns the most results obtainable instead of a failure the caller must learn to recover from; it matches how every other paged endpoint here already behaves; and it is strictly better than today's total failure. The silent-cap concern is mitigated on both ends — the response's `count` reports how many matches actually came back, and the tool schema now states the cap up front so a caller isn't surprised.

## The replacement doc comment

The false comment is gone. Its replacement quotes Slack verbatim, links the source, dates the retrieval, names the failure mode, and explicitly records what the old claim got wrong:

```rust
/// Upper bound on `count` for `search.messages` specifically. Slack's
/// reference for that method states "Pass the number of results you want per
/// 'page'. Maximum of `100`." and, under Usage info, "the max `count` value is
/// `100` and the max `page` value is `100`"
/// (<https://docs.slack.dev/reference/methods/search.messages>, retrieved
/// 2026-07-28). A `count` above 100 is rejected live with
/// `{"ok": false, "error": "invalid_arguments"}`.
///
/// This is deliberately NOT shared with [`MAX_CONVERSATION_PAGE_SIZE`]: the
/// ceilings are per-method and an order of magnitude apart. The predecessor
/// constant (`MAX_PAGE_SIZE = 1000`) claimed to be the widest window
/// `search.messages` accepts; that claim was false and broke every search whose
/// `count` landed in 101..=1000 (issue: this PR).
pub(super) const MAX_SEARCH_COUNT: i64 = 100;
```

## Why this shipped untested

The only existing coverage, `search_messages_with_user_token_returns_matches`, mounts a mock that answers `ok:true` regardless of `count` — the 101..=1000 band was never exercised. The new integration test asserts on the **outgoing request body** via `body_partial_json`, so a mock that ignores the parameter can no longer hide the bug.

## Mutation proof

Constant reverted to 1000 → both new tests fail; restored → `git status --porcelain` empty (`diff` vs pristine snapshot: IDENTICAL).

```
---- slack::handlers::args::tests::search_and_conversation_ceilings_match_slack_docs stdout ----
panicked at crates/trusty-channels/src/slack/handlers/args.rs:353:9:
test result: FAILED. 109 passed; 1 failed

---- search_messages_clamps_count_to_slack_maximum_before_request stdout ----
panicked at crates/trusty-channels/tests/tools_http.rs:550:13:
count 101 must be clamped to 100 before the request; got Slack(UnexpectedStatus(404))
test result: FAILED. 5 passed; 1 failed
```

(The 404 is the point: with a 1000 ceiling the outgoing body carries `count: 1000`, no mounted route matches, and the request fails — proving the assertion is on the wire, not on the response.)

## Gate

```
cargo test -p trusty-channels     110 passed / 9 / 15 / 12 / 50 passed; 0 failed across all binaries
cargo clippy -p trusty-channels --all-targets -- -D warnings    Finished, zero warnings
cargo fmt --check    clean
```

`make check` deliberately not run — it cannot pass locally (tauri needs `ui/dist`; CI excludes three GUI crates, the Makefile does not). Known repo defect.

LOC: Added 145 / Removed 20 / Net +125 (of which +69 is the new test coverage).


bobmatnyc:
## VERDICT: APPROVE

Independent adversarial review. Reviewed on a private worktree off `origin/main` with PR head `a902f536`; full files read, not diff hunks. Zero CRITICAL, zero HIGH.

### 1. The documented ceiling — INDEPENDENTLY CONFIRMED

Fetched <https://docs.slack.dev/reference/methods/search.messages> myself (2026-07-28), not relying on the PR's citation. The Arguments table returns verbatim:

> **`count`** `integer` Optional — Pass the number of results you want per "page". Maximum of `100`. _Default:_ `20`

and separately:

> Please note that the max `count` value is `100` and the max `page` value is `100`.

**100 is correct. `MAX_SEARCH_COUNT = 100` (args.rs:40) matches the source.** The prior `MAX_PAGE_SIZE = 1000` was false and the defect narrative holds.

### 2. The rename deviation — THE ENGINEER WAS RIGHT, THE BRIEF WAS WRONG

Ran the grep against `origin/main` myself. `MAX_PAGE_SIZE` had **exactly one production consumer**:

- `src/slack/handlers/search.rs:36` (import), `:58` (doc), `:75` (call site)

Everything else was its own definition (`args.rs:27`), a doc cross-reference (`args.rs:31`), and its own unit-test usages (`args.rs:311-319`). Every `conversations.*` call site already passed a different ceiling:

- `read.rs:48` and `read.rs:87` → `MAX_CONVERSATION_PAGE_SIZE`
- `conversations.rs:83` → `MAX_CONVERSATION_PAGE_SIZE`

Corroborating evidence: an existing `CHANGELOG.md` entry from a prior PR already records "`conversations.history` / `conversations.replies` `limit` is now clamped to Slack's documented 999-message ceiling for that method family (previously shared the 1000 ceiling meant for `search.messages`)" — the split had already happened, leaving `MAX_PAGE_SIZE` governing search alone.

**The rename is behavior-preserving outside `search.messages`. Not a HIGH. The instruction to add a parallel constant was the error, and leaving a falsely-named `MAX_PAGE_SIZE = 1000` in place would have preserved the exact trap for the next caller.**

### 3. Mutation evidence — REPRODUCED, and stronger than claimed

I ran both mutations myself rather than accepting the report.

**Mutation A** (`MAX_SEARCH_COUNT` → 1000, i.e. the original defect):
```
test slack::handlers::args::tests::search_and_conversation_ceilings_match_slack_docs ... FAILED
test search_messages_clamps_count_to_slack_maximum_before_request ... FAILED
  panicked at crates/trusty-channels/tests/tools_http.rs:552:13:
  count 101 must be clamped to 100 before the request; got Slack(UnexpectedStatus(404))
```
The 404 argument **does** prove what is claimed. I verified the mechanism rather than trusting it: `client_with_user_for` (tools_http.rs:29-36) mounts no catch-all, the only route requires `body_partial_json({query, count:100})`, and `BaseClient::request` returns `Err(SlackError::UnexpectedStatus(...))` on any non-success status (`api/client.rs:294-296`). So an unmatched body genuinely produces a hard error — the assertion is on the wire, not on a mock that answers `ok:true` regardless. This is the exact blindness that let the original defect ship.

**Mutation B** (`MAX_SEARCH_COUNT` → 50, i.e. clamping to the *wrong* value rather than not clamping) — the regression class asked about:
```
test search_messages_clamps_count_to_slack_maximum_before_request ... FAILED
test search_messages_forwards_in_range_count_unchanged ... FAILED
test slack::handlers::args::tests::search_and_conversation_ceilings_match_slack_docs ... FAILED
    left: 50
```
**Three tests catch it.** `search_messages_forwards_in_range_count_unchanged` is what closes this gap — it pins that exactly 100 reaches the wire unreduced, so the suite catches both under- and over-clamping, not merely absence of clamping. Full suite on PR head: `50 passed; 0 failed; 0 ignored` (tools_http) and `110 passed; 0 failed` (lib). Working tree restored, verified clean.

### 4. `page` is never sent — genuinely out of scope, CONFIRMED

`search.rs:83` builds `json!({ "query": ..., "count": count })` — no `page`. Results are hard-capped at one page. This is pre-existing and unchanged, and fixing it is a feature (pagination loop or a new `page` argument), not part of this defect. Worth noting it *strengthens* the clamp-vs-reject choice: since one page is the hard ceiling regardless, clamping does not shrink the reachable result set relative to rejecting — it just stops the call from failing outright.

---

## Findings

|---|---|---|---|---|
| MEDIUM | crates/trusty-channels/src/slack/handlers/search.rs | 92 | A capped result set is indistinguishable from an exhausted one. The response is `{"count": matches.len()}`; Slack's envelope carries `messages.total` (total matching messages) and `clean_search_matches` (clean.rs:77-82) reads only `messages.matches`, discarding it. A caller receiving `count: 100` cannot tell "100 matched" from "thousands matched, capped at 100" — and with `scope: "public"` the post-filter at search.rs:90 drops the count below 100, erasing even that weak signal. | Carry the total through: `resp["messages"]["total"]` into the response as e.g. `total_available`, so a caller (or model) can see truncation directly rather than inferring it from the schema prose. |
| LOW | crates/trusty-channels/src/slack/handlers/args.rs | 29 | Misattributed section. The doc says the sentence appears "under Usage info". My fetch places "Please note that the max `count` value is `100`…" under the **Response** section; "Usage info" is the section carrying the legacy-method notice. The quote itself and the number are verbatim-correct — only the section attribution is off. | Change "under Usage info" to "under Response", or drop the section name and keep the quote plus URL. |
| LOW | crates/trusty-channels/src/slack/handlers/args.rs | 38 | Understates the affected band. "broke every search whose `count` landed in 101..=1000" is true of the *forwarded* value but not the *requested* one — a requested `count` of 5000 was clamped to 1000 and rejected just the same. Every request with `count > 100` was broken. The same imprecision appears in the CHANGELOG entry ("any `count` between 101 and 1000"). | Say "every search with a `count` above 100 (values over 1000 were clamped to 1000, which Slack rejected identically)". |
| LOW | crates/trusty-channels/src/slack/handlers/args.rs | 39 | `(issue: this PR)` is a placeholder, not a citation — the surrounding convention cites issue numbers (`#2996`, `#3617`). | Replace with the issue number, or drop the parenthetical. |

## Notes

- **Doc-accuracy audit (requested explicitly given six false doc claims in this repo today).** I checked every load-bearing claim in the new prose at `args.rs:27-39` against my own fetch: the quoted argument text is verbatim, the second quote is verbatim, the value 100 is correct, the URL resolves, and the retrieval date matches today. The only inaccuracy is the section attribution (LOW above). **The replacement prose is materially accurate** — it does not repeat the failure mode of the text it replaces. The claim that Slack rejects with `{"ok": false, "error": "invalid_arguments"}` is an empirical claim about live behavior rather than a documented one; it matches the reported production symptom and the docs do not contradict it, but it is not doc-verifiable.
- **Slack marks `search.messages` as a legacy method.** The page's Usage info recommends the Real-time Search API (`assistant.search.context`) instead. Not a defect in this PR and not a reason to hold it — the fix is correct for the method as it exists today — but it is worth a separate look before investing further in this tool. Required scope remains `search:read`.
- **Clamp-vs-reject assessed as acceptable, not ideal.** The schema description (tools.rs:214-218) does state the cap up front, which is the meaningful mitigation for a model-facing tool. The second stated mitigation — "the response reports actual matches returned" — is weak on its own for the reason in the MEDIUM finding. Given that `page` is not sent, rejecting instead of clamping would only convert a usable 100-result answer back into the hard failure being fixed, so clamp is the right call here.
- No instance of the repo's recurring fail-open shape found in this diff. The clamp is a deliberate, documented narrowing on the *request* path, not a failure branch that advances state while alarms read healthy; the error path (`api/client.rs:294-296`) returns `Err` on every non-success status.

## CI

`gh pr checks 4242`: **all green** — 22 passing (Test 15m36s, Clippy, MSRV 1.91, Format, SLD lint, doc-comment pointer lint, line-cap, plus 8 UI checks), 1 skipping (`Notify on red main`). Zero failures, zero pending.

None of the four findings block merge. MEDIUM and LOWs are worth a follow-up commit if convenient but do not warrant another review round.

Issue #4247, opened 2026-07-28
instruction package: decide the AGENT_DELEGATION.md and PM_INSTRUCTIONS_DEPLOYED.md override paths


bobmatnyc:
Epic #4183's re-sourcing work is landing for the **bundled-fallback** configuration only. `resolve_pm_prompt` (`crates/trusty-mpm/src/core/instruction_overrides.rs:141`, override branches at :161-174) has three live configurations; two are deliberately left on the legacy code path because they fail the `InstructionPackage` schema validation added by PR #4223:
- `.trusty-mpm/AGENT_DELEGATION.md` project override — bypasses the roster entirely, fails `RosterNotConsumed`.
- `PM_INSTRUCTIONS_DEPLOYED.md` — has no delegation section at all, fails both `SectionWithoutBlocks` and `RosterNotConsumed`.

The owner deliberately did not prescribe the answer on #4186 ("a design call for whoever picks this up"). A related question — whether the `.trusty-mpm/AGENT_DELEGATION.md` override survives at all — was flagged as being raised with the owner separately and is still unresolved. That question likely needs answering first, since deleting the override would dissolve half of this issue.

Resolution: decide, per override path, one of — make it expressible under the schema; relax the specific validation rule; or drop the override. Then either re-source it or remove it. Until then these two paths compose by legacy code while the fallback composes via `InstructionPackage`, so two mechanisms coexist.

Related: #4183, PR #4223

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
## Constraint from #4249: Memory and Search are not independently overridable yet

Surfaced by adversarial review of PR #4249 (MEDIUM-3) and recorded here because #4247 is what will make it actionable.

`bundled_pm_package::PM_INSTRUCTIONS_CUTS` attributes `SectionId::Memory` and `SectionId::Search` to two halves of a single numbered list in `PM_INSTRUCTIONS.md` — the "Context-First Protocol" block, where item `1.` is the memory protocol and item `2.` is the code-search protocol. Both sections declare `customization_tier: project`, which is a machine-readable claim that a project may replace either one independently.

Nothing honours the tiers today, so this is currently harmless. The moment #4247 wires per-section overrides it stops being harmless:

- replacing **Memory** alone deletes list item `1.` and leaves the Search section opening on a bare `2.` — a malformed numbered list in the delivered system prompt;
[...]

PR #4249, opened 2026-07-28
feat(trusty-mpm): compose the bundled-fallback PM prompt via InstructionPackage (#4183)

bobmatnyc:
Epic [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183) (milestone tm 1.3.0), carrying the scope formerly tracked as #4186.

Re-sources the PM instruction prompt composition through the `InstructionPackage` type landed by #4223, **for the bundled-fallback configuration only**.

## Scope — one of three configurations, deliberately

| # | configuration | path after this PR |
|---|---|---|
| 1 | bundled fallback (roster present, no section override) | **`InstructionPackage::compose`** |
| 2 | `.trusty-mpm/AGENT_DELEGATION.md` override | legacy, unchanged **by design** |
| 3 | `.trusty-mpm/PM_INSTRUCTIONS_DEPLOYED.md` | legacy, unchanged **by design** |

Configurations 2 and 3 are not merely unported — they are currently *inexpressible* in the schema. An `AGENT_DELEGATION.md` override replaces the whole delegation section and therefore never consumes the computed roster (`RosterNotConsumed`, the #4196 guard); `PM_INSTRUCTIONS_DEPLOYED.md` replaces the body with opaque prose that contributes no delegation section at all (`SectionWithoutBlocks` as well). Both keep flowing through exactly the code they do today, marked with `See #4183` at the branch points. **Neither validation check is weakened to accommodate them** — [#4247](https://github.com/bobmatnyc/trusty-tools/issues/4247) tracks the decision.

## The acceptance gate: strict byte-equality, at the entry point

The composed prompt is byte-for-byte what `resolve_pm_prompt` produced before this PR. Nothing resolves after composition — the agent roster is fixed at session launch and Claude Code cannot change the agent set afterwards — so the composed string IS the delivered prompt, and any divergence is a delivered-prompt regression.

Following review, the gate is asserted **through `resolve_pm_prompt` itself** (`resolve_pm_prompt_is_byte_identical_to_the_legacy_assembly`), not just between the two composers given hand-supplied arguments. Each entry-point test deploys a project-tier agent first, so the composed branch is a property of the test rather than of the machine's `~/.claude/agents` — on a clean CI runner the older tests all took the *legacy* branch. No `$HOME` scoping is used, deliberately: the project tier is sufficient and scoping `$HOME` would reintroduce the cross-test `HOME_LOCK` race for no added coverage.

`resolve_pm_prompt` now also reports which composer ran. The two are byte-identical by contract, so the delivered string cannot reveal the path — without that seam a byte-equality assertion through the entry point would keep passing even if the package branch were deleted outright. The source is logged, so it is operator-visible rather than test-only.

**Mutation evidence** (each applied to the call site, suite re-run, then reverted):

| mutation | result |
|---|---|
| `addendum.as_deref()` → `None` (drops every project's `INSTRUCTIONS.md`) | RED — `resolve_pm_prompt_is_byte_identical_with_a_project_addendum`: *"the project addendum must reach the delivered prompt on the composed path"* |
| delete the `workflow_override.is_none() && memory_override.is_none()` filter | RED — `workflow_override_forces_the_legacy_path_even_with_a_roster` and `memory_override_forces_the_legacy_path_even_with_a_roster`: `left: Package, right: Legacy` |
| force the composed branch never to run | RED — all three `resolve_pm_prompt_*` tests: `left: Legacy, right: Package` |
| `Join::Blank` → `Join::Rule` on the roster note | RED — `first difference at byte offset 33520 (expected 40126 bytes, actual 40131 bytes)` |

## How byte-equality is held — and what it does not cover

`bundled_pm_package` does not paste asset text into a hand-authored package. It **cuts the assets at runtime**: `PM_INSTRUCTIONS.md` into Core / Memory / Search, `BASE_PM.md` into Identity / Non-Overridable Rules / Framework-Guaranteed Conventions, deriving each block's `Join` from the exact bytes removed. Reassembly is true by construction, so re-sectioning cannot move a byte. `Join::Rule` is the same literal as `instruction_pipeline::SECTION_SEPARATOR`, so top-level boundaries match the legacy join.

That property cuts both ways, and the earlier revision of this description overstated it. Composed output equals `asset.trim()` **wherever the cuts land**, so the byte gate is structurally incapable of catching a cut placed at the wrong offset: bytes would be fine while the section *attribution* — the thing this PR exists to establish — was silently wrong. Two guards cover that instead:

- `split_asset` requires every marker to occur **exactly once** per asset (`MarkerNotUnique`), turning a duplicated marker from silent to red; a deleted marker is `MarkerNotFound`, and a move that reorders sections fails the same way because marker search is forward-only;
- the cut tests pin each block's **opening content**, not merely the section-id sequence.

Notes: the floor block stream is `Identity, NonOverridable, FrameworkConventions, NonOverridable` — the canonical-order inversion `validate_floor_is_last` deliberately permits, because `## Trusty Tool Priority` sits last in the asset. `output_style::inject_style_into_prompt` stays an outer wrapper applied by `build_system_prompt_for_with_style_and_native`; it is not pulled into the package model and not part of the comparison.

Recorded on #4247 ([comment](https://github.com/bobmatnyc/trusty-tools/issues/4247#issuecomment-5109187627)) and on `PM_INSTRUCTIONS_CUTS`: Memory and Search are two halves of one numbered list while both declaring tier `project`, so they are **not independently overridable** until `PM_INSTRUCTIONS.md` gives each its own heading.

## Verification

- `cargo test -p trusty-mpm` — exit 0, **3829 lib tests passed, 0 failed**, 6614 passed across the whole crate, 0 failures anywhere; 24 tests in `bundled_pm_package_tests.rs`
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- `scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK` (`instruction_overrides.rs` 421, `bundled_pm_package.rs` 293 — both under the 500 prod cap; `bundled_pm_package_tests.rs` 482, under the 3000 test cap)
- every `Test:` doc pointer added here resolves to a real `fn`

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,6 +7,28 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Changed
+
+- `core::instruction_overrides`: the DEFAULT PM prompt — the bundled-fallback
+  configuration every project without a `.trusty-mpm/` section override receives
+  — is now composed through the typed `InstructionPackage`
+  (epic [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183)) instead
+  of a four-asset string concatenation. The new `core::bundled_pm_package` cuts
+  `PM_INSTRUCTIONS.md` into Core/Memory/Search blocks and `BASE_PM.md` into the
+  three absorbed floor blocks *at runtime*, deriving each block's join from the
+  exact bytes it removed, so re-sectioning an asset cannot move a byte and the
+  eight-section taxonomy now describes the prompt that actually ships. The
+  composed prompt is byte-identical to the previous assembly, gated by
+  `composed_package_is_byte_identical_to_the_legacy_bundled_fallback`, which
+  reports the first differing byte offset with surrounding context on failure.
+  Scope is the bundled fallback ONLY: an `AGENT_DELEGATION.md` override
+  (replaces the section, so never consumes the computed roster) and
+  `PM_INSTRUCTIONS_DEPLOYED.md` (opaque body, no delegation section) are
+  currently inexpressible in the schema and stay on the legacy path by design —
+  neither `RosterNotConsumed` nor `SectionWithoutBlocks` is weakened to
+  accommodate them. `resolve_pm_prompt` now scans the deployed-agent tiers
+  exactly once whichever path it takes.
+
 ### Added
 
 - `core::instruction_package`: the sectioned-JSON instruction-package schema
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -26,8 +26,11 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `PM_INSTRUCTIONS_DEPLOYED.md` (opaque body, no delegation section) are
   currently inexpressible in the schema and stay on the legacy path by design —
   neither `RosterNotConsumed` nor `SectionWithoutBlocks` is weakened to
-  accommodate them. `resolve_pm_prompt` now scans the deployed-agent tiers
-  exactly once whichever path it takes.
+  accommodate them. The gate is asserted through `resolve_pm_prompt` itself with
+  a project-tier agent deployed, so the composed branch is a property of the test
+  rather than of the machine's `~/.claude/agents`, and `resolve_pm_prompt` now
+  reports which composer produced the prompt — the two are byte-identical by
+  contract, so the delivered string alone cannot tell you.
 
 ### Added
 
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -29,8 +29,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   accommodate them. The gate is asserted through `resolve_pm_prompt` itself with
   a project-tier agent deployed, so the composed branch is a property of the test
   rather than of the machine's `~/.claude/agents`, and `resolve_pm_prompt` now
-  reports which composer produced the prompt — the two are byte-identical by
-  contract, so the delivered string alone cannot tell you.
+  logs which composer produced the prompt — the two are byte-identical by
+  contract, so the delivered string alone cannot tell you. The gate injects a
+  fixed roster rather than scanning the agent tiers per side: those tiers are
+  machine-global mutable state that live `tm` sessions rewrite at launch, so
+  scanning twice made the gate itself intermittently red with a message
+  indistinguishable from a real prompt regression.
 
 ### Added
 
--- a/crates/trusty-mpm/src/core/bundled_pm_package.rs
+++ b/crates/trusty-mpm/src/core/bundled_pm_package.rs
@@ -1,0 +1,462 @@
+//! The bundled-fallback PM prompt, expressed as an [`InstructionPackage`].
+//!
+//! Why: #4184 landed the sectioned-JSON package type with zero call sites — a
+//! schema nothing composes from is a schema nothing validates. This module is
+//! the first real producer (#4183, carrying the scope formerly tracked as
+//! #4186): it re-sources the DEFAULT PM prompt — the one every project with no
+//! `.trusty-mpm/` override receives — through
+//! [`InstructionPackage::compose`], so the eight-section taxonomy and its
+//! `fixed | project | user` tiers describe the prompt that actually ships
+//! rather than a parallel model.
+//!
+//! Scope — deliberately ONE of the three configurations
+//! [`crate::core::instruction_overrides::resolve_pm_prompt`] can emit:
+//!
+//! | # | configuration | path |
+//! |---|---|---|
+//! | 1 | bundled fallback, roster present | **this module** |
+//! | 2 | `.trusty-mpm/AGENT_DELEGATION.md` override | legacy, unchanged |
+//! | 3 | `.trusty-mpm/PM_INSTRUCTIONS_DEPLOYED.md` | legacy, unchanged |
+//!
+//! Configurations 2 and 3 are not merely unported, they are currently
+//! *inexpressible*: an `AGENT_DELEGATION.md` override replaces the whole
+//! delegation section and so never consumes the computed roster
+//! ([`crate::core::instruction_package::ValidationError::RosterNotConsumed`]),
+//! and `PM_INSTRUCTIONS_DEPLOYED.md` replaces the entire body with opaque
+//! prose that contributes no delegation section at all (additionally
+//! `SectionWithoutBlocks`). Both are tracked as follow-up work on epic #4183;
+//! neither is weakened here.
+//!
+//! What: [`compose_bundled_fallback`] builds the package for the compiled-in
+//! assets and composes it with the three host-supplied inputs (rendered agent
+//! roster, stack profile, project addendum).
+//!
+//! THE BYTE-EQUALITY CONTRACT. The composed prompt is byte-identical to what
+//! the legacy concatenation in `instruction_overrides` produces for the same
+//! inputs — not "semantically equivalent". Nothing resolves later: the agent
+//! roster is fixed at session launch and Claude Code cannot change the agent
+//! set afterwards, so the composed string IS the delivered prompt and any
+//! divergence is a delivered-prompt regression. Two mechanisms hold the line:
+//!
+//! * every block that comes from a bundled asset is *cut out of that asset at
+//!   runtime* by [`split_asset`], which derives each block's [`Join`] from the
+//!   exact bytes it removed. Reassembling the pieces therefore reproduces the
+//!   asset by construction — re-sectioning an asset cannot move a byte, and
+//!   editing an asset cannot silently change the composed output;
+//! * [`Join::Rule`] is `"\n\n---\n\n"`, the same literal as
+//!   [`crate::core::instruction_pipeline::SECTION_SEPARATOR`], so top-level
+//!   section boundaries match the legacy `join_sections` join.
+//!
+//! Test: `bundled_pm_package_tests.rs` — in particular
+//! `composed_package_is_byte_identical_to_the_legacy_bundled_fallback`, which
+//! diffs the two strings and reports the first differing byte offset.
+
+use crate::core::instruction_overrides::ROSTER_PRECEDENCE_NOTE;
+use crate::core::instruction_package::{
+    BlockBody, CompositionError, CompositionInputs, CustomizationTier, Generator, InstructionBlock,
+    InstructionPackage, InstructionSection, Join, SCHEMA_VERSION, SectionId,
+};
+use crate::core::instruction_pipeline::{AGENT_DELEGATION, BASE_PM, PM_INSTRUCTIONS, WORKFLOW};
+
+/// Stable identity of the package this module builds.
+pub(crate) const PACKAGE_ID: &str = "trusty-mpm.pm.bundled-fallback";
+
+/// A failure building or composing the bundled-fallback package.
+///
+/// Why: both variants mean the delivered PM prompt would be wrong, so neither
+/// may be swallowed. `MarkerNotFound` in particular fires when a bundled asset
+/// was edited so that a section boundary this module cuts on no longer exists
+/// — a loud error there is what stops a silently re-sectioned prompt.
+/// What: a missing cut marker, or a composition failure from the package type.
+/// Test: `missing_cut_marker_is_an_error`, and
+/// `shipped_assets_build_and_validate` proves neither occurs for what we ship.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub(crate) enum PackageError {
+    /// A cut marker is absent from the asset it is supposed to split.
+    #[error("asset {asset} no longer contains the section cut marker {marker:?}")]
+    MarkerNotFound {
+        /// The asset file name.
+        asset: &'static str,
+        /// The marker that could not be located.
+        marker: &'static str,
+    },
+    /// Composing the built package failed.
+    #[error(transparent)]
+    Compose(#[from] CompositionError),
+}
+
+/// One section boundary inside a bundled markdown asset.
+///
+/// `start_marker` is the literal text that begins the piece; the first cut of
+/// an asset uses `""` to mean "start of asset". Markers are searched forward
+/// from the previous cut, so they need only be unique in the remaining text.
+struct Cut {
+    /// Section the resulting block is attributed to.
+    section: SectionId,
+    /// Literal text that starts this piece (`""` for the first piece).
+    start_marker: &'static str,
+}
+
+/// How `PM_INSTRUCTIONS.md` divides into Core / Memory / Search blocks.
+///
+/// Why: the taxonomy needs a Memory and a Search section, and the asset's
+/// "Context-First Protocol" block is exactly that guidance — item 1 is the
+/// memory protocol, item 2 the code-search protocol. Cutting mid-list looks
+/// odd but is precisely what the two-array block model exists for: a section is
+/// lifted out of the middle of a legacy document without moving a byte. The
+/// asset's own `## Identity` heading stays in `Core`, NOT
+/// [`SectionId::Identity`] — that section is the non-overridable floor's
+/// identity block, and attributing an overridable block to it would make every
+/// later block "overridable after the floor".
+/// What: four pieces — Core prologue, Memory, Search, Core remainder.
+/// Test: `pm_instructions_cuts_reassemble_the_asset`.
+const PM_INSTRUCTIONS_CUTS: &[Cut] = &[
+    Cut {
+        section: SectionId::Core,
+        start_marker: "",
+    },
+    Cut {
+        section: SectionId::Memory,
+        start_marker: "## Context-First Protocol",
+    },
+    Cut {
+        section: SectionId::Search,
+        start_marker: "2. `search` (`mcp__trusty-search__search`)",
+    },
+    Cut {
+        section: SectionId::Core,
+        start_marker: "## Agent Routing",
+    },
+];
+
+/// How `BASE_PM.md` divides into the three absorbed floor sections.
+///
+/// Why: the mapping is the one documented on
+/// [`crate::core::instruction_package::SectionId`]. The resulting block stream
+/// runs `Identity, NonOverridableRules, FrameworkGuaranteedConventions,
+/// NonOverridableRules` — a canonical-order inversion, because the asset places
+/// the tool-priority block last. That is exactly the shape
+/// `validate_floor_is_last` deliberately permits.
+/// What: four pieces; `## Customizing PM Behavior` rides with
+/// `## Non-Overridable Rules`, and `## Trusty Tool Priority` returns to it.
+/// Test: `base_pm_cuts_reassemble_the_asset`, `floor_blocks_are_the_contiguous_tail`.
+const BASE_PM_CUTS: &[Cut] = &[
+    Cut {
+        section: SectionId::Identity,
+        start_marker: "",
+    },
+    Cut {
+        section: SectionId::NonOverridableRules,
+        start_marker: "## Non-Overridable Rules",
+    },
+    Cut {
+        section: SectionId::FrameworkGuaranteedConventions,
+        start_marker: "## Framework-Guaranteed Conventions (Non-Overridable)",
+    },
+    Cut {
+        section: SectionId::NonOverridableRules,
+        start_marker: "## Trusty Tool Priority (Non-Overridable)",
+    },
+];
+
+/// The declared eight-section taxonomy with the tiers this build ships.
+///
+/// Why: tiers are not decoration — they are the machine-readable statement of
+/// which `.trusty-mpm/` override files `BASE_PM.md` advertises. The five content
+/// sections are `project` because every advertised override file is project
+/// scoped; the three floor sections are `fixed` because
+/// `resolve_pm_prompt` appends the floor last under every branch.
+/// What: [`SectionId::CANONICAL`] paired with its tier and title.
+/// Test: `shipped_assets_build_and_validate`.
+fn sections() -> Vec<InstructionSection> {
+    let declare = |id: SectionId, title: &str, tier: CustomizationTier| InstructionSection {
+        id,
+        title: title.to_string(),
+        customization_tier: tier,
+        description: None,
+    };
+    vec![
+        declare(SectionId::Identity, "Identity", CustomizationTier::Fixed),
+        declare(
+            SectionId::Core,
+            "Core PM Instructions",
+            CustomizationTier::Project,
+        ),
+        declare(
+            SectionId::Memory,
+            "Memory Protocol",
+            CustomizationTier::Project,
+        ),
+        declare(
+            SectionId::Search,
+            "Code Search Protocol",
+            CustomizationTier::Project,
+        ),
+        declare(SectionId::Workflow, "Workflow", CustomizationTier::Project),
+        declare(
+            SectionId::AgentDelegation,
+            "Agent Delegation",
+            CustomizationTier::Project,
+        ),
+        declare(
+            SectionId::NonOverridableRules,
+            "Non-Overridable Rules",
+            CustomizationTier::Fixed,
+        ),
+        declare(
+            SectionId::FrameworkGuaranteedConventions,
+            "Framework-Guaranteed Conventions",
+            CustomizationTier::Fixed,
+        ),
+    ]
+}
+
+/// The exact bytes between two adjacent pieces, as a declared [`Join`].
+///
+/// Why: byte-equality requires every boundary to be declared rather than
+/// inferred; naming the two common gaps keeps a composed package readable
+/// while `Literal` guarantees fidelity for anything else.
+/// What: maps the removed gap bytes onto the closest [`Join`] variant.
+/// Test: `pm_instructions_cuts_reassemble_the_asset`.
+fn join_for_gap(gap: &str) -> Join {
+    match gap {
+        "" => Join::None,
+        "\n\n" => Join::Blank,
+        "\n\n---\n\n" => Join::Rule,
+        other => Join::Literal(other.to_string()),
+    }
+}
+
+/// Cut a bundled asset into per-section blocks without moving a byte.
+///
+/// Why: the alternative — pasting asset text into a hand-authored package —
+/// makes the byte-equality gate a thing someone must re-verify after every
+/// asset edit. Cutting at runtime makes reassembly true by construction: each
+/// block carries the literal gap that was removed before it, so concatenating
+/// the blocks with their joins reproduces `asset.trim()` exactly.
+/// What: locates each cut's `start_marker` (searching forward from the previous
+/// cut), slices the asset, trims each piece, and records the removed whitespace
+/// as the next block's [`Join`]. `lead_join` is the join before the FIRST
+/// piece, i.e. the boundary with whatever precedes this asset in the stream.
+/// Test: `pm_instructions_cuts_reassemble_the_asset`, `base_pm_cuts_reassemble_the_asset`,
+/// `missing_cut_marker_is_an_error`.
+fn split_asset(
+    asset: &'static str,
+    raw: &str,
+    cuts: &'static [Cut],
+    lead_join: Join,
+) -> Result<Vec<InstructionBlock>, PackageError> {
+    let text = raw.trim();
+
+    // Resolve each cut to an absolute byte offset, forward-only.
+    let mut offsets: Vec<usize> = Vec::with_capacity(cuts.len());
+    let mut search_from = 0usize;
+    for (index, cut) in cuts.iter().enumerate() {
+        if index == 0 {
+            offsets.push(0);
+            continue;
+        }
+        let Some(relative) = text
+            .get(search_from..)
+            .and_then(|t| t.find(cut.start_marker))
+        else {
+            return Err(PackageError::MarkerNotFound {
+                asset,
+                marker: cut.start_marker,
+            });
+        };
+        let absolute = search_from + relative;
+        offsets.push(absolute);
+        search_from = absolute + cut.start_marker.len();
+    }
+
+    let mut blocks: Vec<InstructionBlock> = Vec::with_capacity(cuts.len());
+    for (index, cut) in cuts.iter().enumerate() {
+        let start = offsets.get(index).copied().unwrap_or(0);
+        let end = offsets.get(index + 1).copied().unwrap_or(text.len());
+        let piece = text.get(start..end).unwrap_or_default();
+        let body = piece.trim();
+        if body.is_empty() {
+            // An empty piece means two cut markers are adjacent — a mis-authored
+            // cut table, not valid content. `validate` would reject it anyway;
+            // failing here names the asset.
+            return Err(PackageError::MarkerNotFound {
+                asset,
+                marker: cut.start_marker,
+            });
+        }
+
+        let join_before = if index == 0 {
+            lead_join.clone()
+        } else {
+            let previous = text.get(offsets.get(index - 1).copied().unwrap_or(0)..start);
+            let trailing = previous
+                .map(|p| p.get(p.trim_end().len()..).unwrap_or_default())
+                .unwrap_or_default();
+            let leading = piece
+                .get(..piece.len() - piece.trim_start().len())
+                .unwrap_or_default();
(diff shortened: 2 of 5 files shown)


bobmatnyc:
## VERDICT: WARN

Independent adversarial review at `928739ca`, in a detached read-only worktree off `origin/feat/4183-resource-pm-prompt`. No CRITICAL findings: the byte-equality claim holds, the two legacy configurations are genuinely unchanged, and neither schema check was weakened. One HIGH: the gate is anchored one layer below the code that actually ships the prompt.

### Verification performed (not taken on trust)

| check | result |
|---|---|
| `cargo test -p trusty-mpm --lib` | `test result: ok. 3816 passed; 0 failed; 6 ignored` — all 11 new `core::bundled_pm_package::tests::*` present and `ok` |
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | exit 0 |
| `cargo fmt --check -p trusty-mpm` | exit 0 |
| `scripts/check_line_cap.sh` | `7 allowlisted, 0 violations — OK` |
| every `Test:` doc pointer in the two touched prod files | 8/8 in `bundled_pm_package.rs`, 14/14 in `instruction_overrides.rs` resolve to a real `fn` (scripted) |
| CHANGELOG | entry present, exactly one `## [Unreleased]` |

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/bundled_pm_package_tests.rs` | 37–45, 84–95 | The byte-equality gate compares two helpers given identical hand-supplied inputs; it never exercises `resolve_pm_prompt`, so the call-site wiring is unpinned | Run the gate through the production entry point + add override-combination tests that deploy an agent |
| MEDIUM | `crates/trusty-mpm/src/core/instruction_overrides.rs` | 195, 203 | Which composition path the `resolve_pm_prompt` tests exercise is decided by the machine's `~/.claude/agents` | `deploy_agent()` in every test that must pin composed-path behavior |
| MEDIUM | `crates/trusty-mpm/src/core/bundled_pm_package.rs` | 44–45, 259–271 | Cutting fails loudly only for an *absent* marker; a moved or duplicated marker cuts silently at the wrong place and is structurally undetectable | Assert cut *content*, not just reassembly + section-id sequence |
| MEDIUM | `crates/trusty-mpm/src/core/bundled_pm_package.rs` | 118–129, 185–194 | Memory/Search split leaves a half list and a dangling "Both tools" sentence while both sections declare tier `project` | Record the constraint on #4247 or restructure the asset before the tiers become actionable |
| LOW | `crates/trusty-mpm/src/core/bundled_pm_package.rs` | 279–287 | Empty-piece guard reports `MarkerNotFound` naming a marker that *was* found | Distinct `EmptyPiece` variant |
| LOW | `crates/trusty-mpm/CHANGELOG.md` | 27–28 | "previously the roster scan was reachable twice" is not true of the pre-PR code | Drop or correct the claim |

---

#### HIGH-1 — the acceptance gate stops one layer short of the delivered prompt

`bundled_pm_package_tests.rs:37`

```rust
fn legacy_bundled_fallback(stack: &str, roster: &str, addendum: Option<&str>) -> String {
    assemble_sections(stack.to_string(), None, WORKFLOW.trim().to_string(),
                      delegation_with_roster(Some(roster)), addendum.map(str::to_string))
}
```

The oracle is **not** rotten — I verified `assemble_sections` and `delegation_with_roster` are pure extractions of the pre-PR branch-2 body (`git diff origin/main...HEAD`, read line by line; `floor` → `BASE_PM` is the same constant via `let floor = BASE_PM;`), and both are live production code on configuration 2. That part of the author's claim checks out.

What the gate does not cover is the seam that ships the prompt. Both sides receive the same three arguments *because the test constructs them that way*. It proves `compose_bundled_fallback` ≡ `assemble_sections` **given identical inputs**; it does not prove `resolve_pm_prompt` hands the package the inputs it would have handed the legacy assembly. That wiring is `instruction_overrides.rs:187–227`, and it is where a delivered-prompt regression would actually live.

Under CI conditions the composed branch is reached by exactly **one** `resolve_pm_prompt` test. `deployed_roster_section` unions the project tier, `CLAUDE_CONFIG_DIR`, and `FrameworkPaths::default()` → `dirs::home_dir()`/`.claude/agents`. On a clean runner none of those exist, so `roster` is `None` and the composed branch is skipped. The only test that deploys a project-tier agent with no override is `bundled_delegation_appends_deployed_roster`, and it asserts substring presence and ordering — not the addendum, not the stack profile, not equality with the previous output.

**Failure scenario 1.** Change `addendum.as_deref()` (`instruction_overrides.rs:206`) to `None`. `instructions_appended` — the only test asserting `.trusty-mpm/INSTRUCTIONS.md` reaches the prompt — deploys no agent, so on a clean runner it takes the legacy branch and stays green. `composed_package_is_byte_identical_with_a_project_addendum` passes the addendum straight into `compose_bundled_fallback`, so it stays green too. Net: every project's `INSTRUCTIONS.md` silently vanishes from the delivered PM prompt with a fully green suite.

**Failure scenario 2.** Delete `.filter(|_| workflow_override.is_none() && memory_override.is_none())` (`instruction_overrides.rs:203`). With a roster present, a `WORKFLOW.md` or `MEMORY.md` override is silently discarded in favour of the bundled asset. `workflow_override_replaces` and `memory_override_is_slotted_after_pm_instructions` deploy no agent, so on a clean runner neither enters the composed branch — both stay green.

**Evidence.** I ran the compiled test binary under two roster conditions without touching the source. With `HOME` pointed at an empty directory (roster absent → legacy branch): `15 passed; 0 failed`. On this machine, where `~/.claude/agents` holds 83 agents (roster present → composed branch): the same 15 pass inside the full run. The suite is path-agnostic by construction, which is exactly why a wiring mutation survives it.

**Fix.** Add to `instruction_overrides::tests`, each calling `deploy_agent()` first so the composed branch is guaranteed:
1. `INSTRUCTIONS.md` written → assert the addendum text is in `resolve_pm_prompt`'s output;
2. `WORKFLOW.md` written → assert the override replaced the bundled section;
3. `MEMORY.md` written → assert `MEMORY_OVERRIDE_HEADING` is present;
4. no override → `assert_eq!(resolve_pm_prompt(dir), legacy_bundled_fallback(stack_profile_section(dir), &deployed_roster_section(dir).unwrap(), None))`.

(4) is the one that matters most: it moves the byte-equality gate onto the production entry point instead of a helper pair, and it is the assertion the PR's own framing ("the composed string IS the delivered prompt") implies but does not currently make.

#### MEDIUM-1 — the exercised path depends on the developer's `$HOME`

Same root as HIGH-1, separately actionable. `instruction_overrides::tests` use bare `TempDir::new()` and never scope `$HOME` or `CLAUDE_CONFIG_DIR`, so `deployed_roster_section` reads the real machine. Concretely: `no_overrides_uses_bundled`, `instructions_appended`, `stack_profile_present_when_detected`, `separators_are_consistent` and `framework_guaranteed_conventions_survive_every_override_combination` run the **composed** path on a provisioned laptop and the **legacy** path in CI. A defect in either path can reproduce on one and not the other. Fix: call `deploy_agent()` explicitly wherever composed-path behavior is the subject, so path selection is a property of the test rather than of the machine.

#### MEDIUM-2 — loud only for the absent marker; silent for the moved one

`bundled_pm_package.rs:44` claims "re-sectioning an asset cannot move a byte, and editing an asset cannot silently change the composed output". The first half is true and well built. The second half is true of *bytes* and false of *attribution*, and it is the attribution this PR exists to establish.

`split_asset` uses plain forward `str::find`. Because reassembly is by construction (`bundled_pm_package.rs:273–312` — I verified the algebra: body_i = p_i.trim(), join_i = trail(p_{i-1}) + lead(p_i), which telescopes back to `text.trim()` for any cut offsets), the composed output equals `asset.trim()` **no matter where the cuts land**. The byte-equality gate is therefore structurally incapable of failing on a mis-placed cut. And `pm_instructions_cuts_reassemble_the_asset` (tests 163–192) asserts only reassembly and the section-*id sequence* — never the cut *locations*.

**Failure scenario.** All six markers are currently unique (verified: 1 occurrence each in `PM_INSTRUCTIONS.md` / `BASE_PM.md`). Add a line `## Agent Routing` — inside a fenced example, a quoted block, or a new subsection — anywhere between `PM_INSTRUCTIONS.md:59` and `:64`. The fourth cut relocates there; the Search block loses "Both tools are stable and recommended…" to Core. `validate()` passes, all 3816 tests pass, `MarkerNotFound` never fires, the delivered bytes are unchanged — and the section model is now wrong. When #4247 wires per-section overrides, an override of Search replaces the wrong text.

**Fix.** Pin the cut locations, not just the shape:
```rust
assert!(matches!(&blocks[1].body, BlockBody::Text { text } if text.contains("memory_recall")));
assert!(matches!(&blocks[2].body, BlockBody::Text { text } if text.starts_with("2. `search`")));
assert!(matches!(&blocks[3].body, BlockBody::Text { text } if text.starts_with("## Agent Routing")));
```
plus an assertion that each `start_marker` occurs exactly once in its asset (`text.matches(m).count() == 1`), which converts a duplicated marker from silent to red.

#### MEDIUM-3 — the mid-list cut, and the part not flagged

The author flagged that the Memory/Search split cuts the Context-First Protocol list between items 1 and 2. Two things ride along that the PR text does not mention:

- the closing sentence "Both tools are stable and recommended for targeted lookups on any project" — which refers to *both* memory and search — is attributed to Search alone;
- `bundled_pm_package.rs:185–194` declares both Memory and Search tier `Project`, i.e. a machine-readable claim that a project may replace them.

**Failure scenario.** Once #4247 honours those tiers, replacing the Memory section deletes list item `1.` and leaves the Search section opening with a bare `2.` — a malformed numbered list in the delivered system prompt, plus a "Both tools" reference to a tool no longer described. Byte-faithful today, but the tier declaration is the thing that makes it a hazard rather than a curiosity.

**Fix.** Either restructure `PM_INSTRUCTIONS.md` so memory and search guidance are separate headings before the tiers become actionable, or record on #4247 that Memory/Search are not independently overridable until that restructure lands.

#### LOW-1 — misleading error on the adjacent-marker case

`bundled_pm_package.rs:279`
```rust
if body.is_empty() {
    return Err(PackageError::MarkerNotFound { asset, marker: cut.start_marker });
}
```
Here the marker *was* found — the defect is that the next cut's marker immediately follows. The message ("asset … no longer contains the section cut marker …") is false and names the wrong marker, sending whoever hits it hunting for a string that is present. Fix: a distinct `PackageError::EmptyPiece { asset, marker }`.

#### LOW-2 — inaccurate CHANGELOG claim

"`resolve_pm_prompt` now scans the deployed-agent tiers exactly once whichever path it takes" is fine; the parenthetical in the PR body, "previously the roster scan was reachable twice", is not. Pre-PR the scan sat behind `read_override(&dir, FILE_AGENT_DELEGATION).unwrap_or_else(|| bundled_delegation(project_dir))` — at most one call. Drop or correct.

1. Add the four `resolve_pm_prompt`-level tests in HIGH-1, each deploying a project-tier agent so the composed branch is guaranteed regardless of `$HOME`. Item (4) — byte-equality asserted through `resolve_pm_prompt` — is the one that closes the gap.
2. Pin the cut locations by content (MEDIUM-2) and assert marker uniqueness per asset.
3. Record the Memory/Search restructure constraint on #4247 (MEDIUM-3).

### Notes — things checked and deliberately not flagged

- **Scope boundaries are honored.** Configuration 3 (branch 1) is untouched apart from comments. Configuration 2 routes `Some(body) => body` into `assemble_sections`, which I verified line-by-line against `origin/main` as a pure extraction — including `floor.trim()` → `BASE_PM.trim()`, the same constant. Reordering the `read_override` calls to the top of branch 2 is side-effect-free.
- **No validation rule was weakened.** `instruction_package.rs` is not in this PR's file list at all — `RosterNotConsumed`, `SectionWithoutBlocks`, `OptionalRoster` and `SectionOnlyOptionalBlocks` are byte-for-byte untouched. The two out-of-scope configurations are excluded by a gate at the call site rather than by relaxing a check, which is the right shape.
- **Fail-closed, and correctly so — this is not the fail-open state-advance pattern.** On a compose error `resolve_pm_prompt` logs `tracing::error!` and falls through to the legacy assembly, which the gate proves byte-identical, so the delivered prompt stays correct. And `MarkerNotFound` can only arise from a source edit to a compile-time `include_str!` asset, which `shipped_assets_build_and_validate` turns red in CI before it can reach a session. Failure is caught at build time and degrades to the correct output at runtime. Well designed.
- **No panics on the launch path.** `bundled_pm_package.rs` contains no `unwrap()`, `expect()`, `panic!` or `unreachable!`; every slice is `get(..).unwrap_or_default()` on offsets that are guaranteed UTF-8 boundaries (`str::find` result + `marker.len()`, and whitespace-trim boundaries).
- **No content-dependent byte divergence exists.** I worked the reachable input space: all three composition inputs are already trimmed upstream (`read_override`, `stack_profile_section`, `generate_authority`), the composer's trim is idempotent over them, and `join_sections`' empty-filter and the `optional` block-drop agree on empty and whitespace-only. There is no real configuration reaching the composed branch where the two assemblies can differ. The byte-equality claim itself is sound.
- **The mutation-sensitivity evidence the author reports is real** — the gate does fail on a `Join::Blank` → `Join::Rule` flip. My HIGH is not that the gate is vacuous; it is that it is anchored below the layer where a regression would be introduced.

bobmatnyc:
## VERDICT: WARN — round 2

Re-review at `be3785f0`, same detached read-only worktree. **Round 1's HIGH is genuinely closed** — I re-ran all three mutations myself and watched each go red. MEDIUM-2 is closed for both cases my original finding named, verified by two asset mutations. Round 2 did not weaken anything.

One new HIGH, which I did not have to look for: **the full suite failed on my first run at `be3785f0`, on the new acceptance-gate test itself.** It is intermittent — 1 red in 4 full-suite runs — and the round-2 tests are what introduced the exposure.

### Mutations — re-run by me, not taken from the PR body

I backed both production files up, applied each mutation, ran the tests, and restored (`git status --porcelain` empty afterwards).

| # | mutation | my observed result |
|---|---|---|
| 1 | `addendum.as_deref()` → `None` | **RED** — `resolve_pm_prompt_is_byte_identical_with_a_project_addendum` panics at `bundled_pm_package_tests.rs:349`, "the project addendum must reach the delivered prompt on the composed path". `23 passed; 1 failed` |
| 2 | delete the `workflow_override.is_none() && memory_override.is_none()` filter | **RED** — `workflow_override_forces_the_legacy_path_even_with_a_roster` + `memory_override_forces_the_legacy_path_even_with_a_roster`. `22 passed; 2 failed` |
| 3 | `.filter(\|_\| false)` — composed branch can never run | **RED** — three tests: `..._takes_the_package_path_when_a_roster_is_deployed`, `..._is_byte_identical_to_the_legacy_assembly`, `..._is_byte_identical_with_a_project_addendum`. `21 passed; 3 failed` |

Round 1's central complaint was that mutation 1 survived a green suite. It no longer does. The seam is real.

**`PromptSource` cannot drift from the delivered bytes.** I checked the coupling specifically. There are exactly three return sites, and each constructs the tag *in the same expression as the string it labels* — `instruction_overrides.rs:216` `(join_sections(sections), Legacy)`, `:247` `Ok(prompt) => return (prompt, Package)`, `:268` `(assemble_sections(...), Legacy)`. There is no second derivation that could disagree, no field set separately from the value, no way to report `Package` while legacy bytes are returned. `resolve_pm_prompt` is a pure logging wrapper over the seam, and `resolve_pm_prompt_wrapper_matches_the_source_reporting_form` pins that. This is the right shape for the problem.

**MEDIUM-2 — verified closed, both cases, by asset mutation.**

- *Marker duplicated inside a fenced code block* (my original scenario, which I said uniqueness might not cover): injected `## Agent Routing` inside a ` ```markdown ` fence earlier in `PM_INSTRUCTIONS.md` → `MarkerNotUnique { marker: "## Agent Routing", count: 2 }`, suite **RED**. `str::matches` is a plain substring count and does not respect markdown structure, so fenced and quoted occurrences do count. Correct behaviour here — fail-loud is the safe direction.
- *Marker moves while staying unique and in order* (the case uniqueness alone does **not** cover): relocated `## Agent Routing` above the closing sentence so "Both tools are stable…" would be silently re-attributed from Search to Core. Bytes reassemble identically and `MarkerNotUnique` does not fire — but `pm_instructions_cut_locations_are_pinned_by_content` goes **RED** on the `search.contains("Both tools are stable")` assertion. The content pinning is what catches it, exactly as intended.

**Module-doc wording is now accurate and does not overstate in a new direction.** "The byte gate is structurally incapable of catching a cut placed at the wrong offset" matches the reassembly algebra. "A marker that moves in a way that reorders the sections fails the same way, because marker search is forward-only" is correct, and the doc does *not* claim the non-reordering move is covered by the marker guards — which is right, since that one is covered by the content pinning instead. The only imprecision is "every marker to occur exactly once": the first cut's `""` sentinel is exempt (`cuts.iter().skip(1)`), correctly, since `text.matches("")` would count `len+1`. Not worth a finding.

**Declining `$HOME` scoping — the reasoning is half right, and the half that is wrong is HIGH-1 below.** Deploying into the project tier *does* guarantee `deployed_roster_section` returns `Some` regardless of ambient state, so path selection is now a property of the test. That part is sound and the `HOME_LOCK` objection is grounded in a real serial guard. What it does not buy is a *deterministic roster* — and three of the new tests compare two independent scans of it.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/bundled_pm_package_tests.rs` | 328–338, 341–364, 448–453 | The acceptance-gate tests scan machine-global mutable agent tiers **twice** and assert the two scans are byte-equal; observed failing 1-in-4 full-suite runs | Pin the roster to one scan via a `_using(roster)` seam, or scope every tier |
| LOW | `crates/trusty-mpm/src/core/instruction_overrides.rs` | 156 | `tracing::debug!` is below the `info` default filter, so the "operator gets a log line" claim in the `PromptSource` doc is not true as shipped | `tracing::info!`, or correct the comment |

---

#### HIGH-1 — the acceptance gate races the ambient agent tiers

`bundled_pm_package_tests.rs:328`
```rust
let (composed, source) = resolve_pm_prompt_with_source(tmp.path());   // scan #1, inside
let roster = deployed_roster_section(tmp.path()).expect("a deployed agent yields a roster"); // scan #2
let legacy = legacy_bundled_fallback(&stack_profile_section(tmp.path()), &roster, None);
assert_byte_identical(&legacy, &composed);
```

`deployed_roster_section` unions three tiers on **every** call — the project tier, `$CLAUDE_CONFIG_DIR/agents`, and `FrameworkPaths::default()` → `~/.claude/agents` (`delegation_authority.rs:292–304`). Deploying into the project tier guarantees the result is `Some`; it does not stop the other two tiers from contributing, and it does not make two successive calls agree. The test asserts they do.

**Observed, not hypothesised.** First full-suite run at `be3785f0` on this workstation:

```
---- core::bundled_pm_package::tests::resolve_pm_prompt_is_byte_identical_to_the_legacy_assembly stdout ----
panicked at crates/trusty-mpm/src/core/bundled_pm_package_tests.rs:80:5:
composed prompt is NOT byte-identical to the legacy assembly.
first difference at byte offset 34917 (expected 47252 bytes, actual 48109 bytes)
expected [34797..]: …anchoring bias.\n- **Model:** sonnet\n\n### dart-engineer\n…
actual   [34797..]: …anchoring bias.\n- **Model:** sonnet\n\n### copyeditor\n- **Handles:** >\n- **Model:** sonnet\n\n### dart-engineer\n…

test result: FAILED. 3828 passed; 1 failed; 6 ignored
```

`### copyeditor` was in scan #1 and gone from scan #2, within one test. It is not a composition defect at all — the two rosters differed. Three subsequent full-suite runs were green (`3829 passed; 0 failed`, matching the reported count), and 8 isolated runs of the test alone were green, so it is load- and timing-dependent: **1 red in 4 full-suite runs.**

Two mechanisms, both live on any trusty-mpm workstation, and I don't need to distinguish them for the finding to hold:
- a concurrent writer — `$CLAUDE_CONFIG_DIR` is set here to `~/.trusty-tools/trusty-mpm/claude-config`, whose `agents/` was rewritten today at 16:13, and there are five live `trusty-mpm` processes redeploying agents at session launch;
- a transient read failure under parallel I/O — `scan_agents` drops an agent **silently** on both `!path.is_file()` (`delegation_authority.rs:149`) and `let Ok(raw) = std::fs::read_to_string(&path) else { continue }` (`:169`), so one scan can lose a file the other keeps with nothing logged.

**Why this is HIGH and not a nit.** This is the acceptance gate — the single assertion the whole PR is built around. Its failure message is `composed prompt is NOT byte-identical to the legacy assembly` with a byte offset and a 40 KB size delta: indistinguishable, at a glance, from a genuine delivered-prompt regression. The only available response to a red gate that goes green on re-run is to re-run it, which is exactly the pattern this project has already had to learn about ("load-induced red degrades the GATE itself; re-running until green = burying a flake"). CI won't see it — a clean runner has no `~/.claude/agents` and no `CLAUDE_CONFIG_DIR`, so the project tier is the only tier and the scans are deterministic. It fires only on developer and agent workstations, i.e. everywhere the suite is actually run by hand.

Three tests carry the exposure, not one: `resolve_pm_prompt_is_byte_identical_to_the_legacy_assembly` (:328), `resolve_pm_prompt_is_byte_identical_with_a_project_addendum` (:341 — same double scan), and `resolve_pm_prompt_wrapper_matches_the_source_reporting_form` (:448 — two full `resolve_pm_prompt` calls, so two internal scans, compared byte-for-byte).

**Fix — preferred, and it costs no `HOME_LOCK`.** Take the roster out of the ambient path for these three tests by extending the seam the PR already established. `PromptSource` exists because an unobservable internal fact had to be made assertable; an *uncontrollable* internal input deserves the same treatment:

```rust
// instruction_overrides.rs
pub(crate) fn resolve_pm_prompt_with_source(dir: &Path) -> (String, PromptSource) {
    resolve_pm_prompt_with_roster(dir, deployed_roster_section(dir).as_deref())
}
pub(crate) fn resolve_pm_prompt_with_roster(dir: &Path, roster: Option<&str>) -> (String, PromptSource) { … }
```
The three tests then call the `_with_roster` form with `FIXED_ROSTER` on both sides — one deterministic roster, no env manipulation, no serialization, and the wiring actually under test (stack profile, addendum, the override gating, `PromptSource`) is unchanged. Keep `resolve_pm_prompt_takes_the_package_path_when_a_roster_is_deployed` on the live scan: it exercises the real tier union and, because it compares nothing against a second scan, it cannot flake.

**Fix — alternative** if you'd rather not add a second seam: scope `$HOME` *and* `CLAUDE_CONFIG_DIR` to the temp project under the existing serial guard, so the project tier is the only populated one. That does pay the `HOME_LOCK` cost, but only for three tests — and it buys determinism, which is what the round-1 note was actually about.

#### LOW-1 — `debug!` is below the default filter

`instruction_overrides.rs:156`
```rust
tracing::debug!(?source, "resolved the PM system prompt");
```
The `PromptSource` doc comment says the seam "gives the operator a log line saying which model produced their prompt." It does not, as shipped: both subscriber setups default to `info` (`src/bin/tm/main.rs:271–274` and `:312–314`, `EnvFilter::try_from_default_env().unwrap_or_else(|_| "info".into())`), so the line is emitted to nobody unless someone sets `RUST_LOG=debug`. The neighbouring code makes the analogous decision the other way on purpose — `delegation_authority.rs:361` uses `tracing::info!` with the comment "Reverting to the bundled asset must be an OBSERVABLE event, not an invisible one". This resolves once per session launch, so `info` is not spam. Either promote it or drop the "operator" clause; as written the justification for the seam claims a production benefit it does not deliver. (The seam is still correct and load-bearing as a test seam — that part stands on its own.)

1. Remove the double-scan from the three tests in HIGH-1 (preferred: the `_with_roster` seam). This is the only blocking item.
2. LOW-1: `tracing::info!` or reword the `PromptSource` doc comment.

### Notes

- **Round-1 findings, re-verified at `be3785f0`:** HIGH-1 closed (three mutations red, run by me). MEDIUM-1 (path selection depending on `$HOME`) closed — the project-tier deploy makes it deterministic; the residual ambient dependency is now HIGH-1 above, which is a different failure mode introduced by the fix, not the old finding re-filed. MEDIUM-2 closed, both cases, verified by asset mutation. MEDIUM-3 recorded on `PM_INSTRUCTIONS_CUTS` with the "Both tools" sentence and the tier claim spelled out, and pinned by `pm_instructions_cut_locations_are_pinned_by_content` so a future edit trips it — better than the plain #4247 note I asked for. LOW-1 fixed with a real `EmptyPiece` variant plus `adjacent_cut_markers_report_an_empty_piece`. LOW-2 removed from the CHANGELOG.
- **Hygiene re-verified independently, all matching the reported numbers:** `cargo clippy -p trusty-mpm --all-targets -- -D warnings` clean; `cargo fmt --check` clean; `scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`; SLOC 421 / 293 / 482; zero dangling `Test:` doc pointers across both production files (scripted); one `## [Unreleased]`.
- **The uniqueness guard is substring, not line-anchored.** Adding an unrelated `## Agent Routing Overrides` heading would trip `MarkerNotUnique` even though the forward search would still be correct. That is a false positive in the fail-loud direction, which is the right direction — noting it only so the eventual red is not mistaken for a real ambiguity.
- **Not flagged:** `deploy_agent` / `write_override` are now duplicated between `instruction_overrides::tests` and `bundled_pm_package_tests.rs`. Two call sites, four lines each, in test code — below the bar.
- **No production-code defect found in this round.** Everything above is test infrastructure. The composed path, the guards, and the two out-of-scope configurations are all correct as written, and nothing in `instruction_package.rs` was touched in either round.

bobmatnyc:
## VERDICT: APPROVE — round 3 (delta only)

Scoped to `be3785f0..690c4020`. **No issues found at >80% confidence.** Round 2's HIGH is closed, and the closure deviation is the right call rather than a compromise. Approved.

### Round 2's HIGH — closed, and verifiable at the import level

The gate tests no longer take any ambient input, and the strongest evidence is that the test module can no longer *name* the scanner: `-use crate::core::delegation_authority::deployed_roster_section;` was removed, and `super::*` (i.e. `bundled_pm_package.rs`) does not re-export it. The three surviving mentions in that file are all inside comments. `resolve_pm_prompt_is_byte_identical_to_the_legacy_assembly`, `..._with_a_project_addendum` and `gate_is_deterministic_across_repeated_runs` now run on a bare `TempDir` with `|| Some(FIXED_ROSTER.to_string())` on both sides — no `deploy_agent`, no tier scan, no `$HOME` touched.

I also swept every byte comparison in the file rather than just the three: lines 102, 115, 344, 388, 536, 806. Only **one** involves an ambient-scanning entry point — `resolve_pm_prompt_wrapper_matches_the_source_reporting_form` at :536 — and it writes an `AGENT_DELEGATION.md` override, so `Some(body) => body` short-circuits and `roster_source()` is never invoked. Zero scans, not merely "no roster byte reaches the output". The five remaining `resolve_pm_prompt_with_source` call sites (:318, :450, :474, :492, :512) assert only `PromptSource` equality and `contains`/`!contains` predicates that ambient roster content cannot flip. **No byte comparison anywhere in the suite is exposed to a second ambient scan.**

### The closure deviation — sound, and not a back door

`roster_source: impl FnOnce() -> Option<String>` is invoked at exactly the line the scan used to occupy (`let roster = roster_source();` inside the `None =>` arm), so production behaviour is unchanged and the laziness claim is structurally real: an `AGENT_DELEGATION.md` override takes `Some(body) => body` and never calls it, and branch 1 returns before the delegation match is even reached. An `Option<String>` parameter really would have forced an eager scan on every launch. `injected_roster_is_lazy_for_the_override_configurations` pins it with a `Cell` counter — 0 scans with an override, exactly 1 without — which is the right way to assert a negative.

It cannot make the two paths diverge. The closure yields **one** value bound to a single `let`, consumed by both the composed branch (`bundled_fallback`) and the legacy fallthrough (`delegation_with_roster(roster.as_deref())`); there is no second evaluation. It supplies the roster *value* only — it cannot reach the override reads or the `workflow_override.is_none() && memory_override.is_none()` filter, so routing is untouchable through it. `injected_roster_does_not_change_which_path_runs` covers both directions (override present → `Legacy` despite an injected roster; `|| None` → `Legacy`). And the seam does not weaken roster *wiring* coverage: mis-wiring the roster at the call site still diverges against the oracle, and passing an empty roster errors into the legacy fallback, flipping `source` to `Legacy` and failing the `assert_eq!`.

**Mutation spot-checked myself** (backed up, applied, ran, restored; `git status --porcelain` empty afterwards). Mutation 1, `addendum.as_deref()` → `None`: **RED** — `resolve_pm_prompt_is_byte_identical_with_a_project_addendum` panics at `bundled_pm_package_tests.rs:369`, `test result: FAILED. 27 passed; 1 failed`. That matches the reported `27 passed; 1 failed` exactly. The seam has not blunted the mutation coverage.

### Item 2 — the single unexplained red

**The claim is structurally true, and I'll state it plainly: that red cannot have been the gate tests.** They no longer call `deployed_roster_section` — the import is gone and the roster arrives from a constant closure — so there is no ambient input for a churn loop to perturb. My own round-2 red *was* the gate test, and that cause is now removed by construction, not by retry.

That converts this from "the gate is flaky" to "something unidentified in a 3833-test suite went red once in 44 runs under a deliberately hostile setup." That is a different and much lower-severity problem, it is not attributable to this delta, and I cannot cite a file, line or fix for it — so it is not a finding here and should not gate the merge. Two observations for whoever picks it up: his `3832 passed; 1 failed` totals 3833, matching my own baseline exactly, so it was one test out of the same set and not a collection difference; and the real gap is that the harness captured only the summary line — `--format json` or `-- --nocapture` on the retry loop would have named it. Worth a separate low-priority issue if it recurs, not a round 4.

### Items 3 and 4 — verified

**The automated gate's MEDIUM is a false positive; the algebra holds.** `trim_end` removes only from the end, so `p[p.trim_end().len()..]` is exactly the trailing whitespace run — not the whole slice. With `gap_i = trail(p_{i-1}) + lead(p_i)` and `body_i = p_i` minus both runs, the sum telescopes to `p_0 + … + p_n` minus `lead(p_0)` minus `trail(p_n)`, and both are empty because `text = raw.trim()`. So reassembly equals `asset.trim()` for **any** cut offsets. The degenerate case where `lead` and `trail` would overlap is an all-whitespace piece, which `EmptyPiece` rejects first.

`derived_joins_survive_irregular_whitespace_around_cuts` genuinely covers the hypothesised case rather than a neighbour of it: piece 0 ends with a fenced block followed by `\n   \n\n` — a blank line carrying spaces immediately before the next heading, which is precisely the shape alleged to break. It asserts the exact derived literals (`"\n   \n\n"`, `"\n\n\n"`, `"\n\t\n"`), not just that the round-trip works, so a *normalising* bug — collapsing that gap to `Join::Blank` — would be caught by the join assertion even in a world where reassembly still happened to hold. Reusing `PM_INSTRUCTIONS_CUTS` against a synthetic asset is a nice touch.

Item 4: `tracing::info!` with an accurate comment (my LOW, closed). `reassemble`'s doc now states the block-0 exclusion and *why* it is correct — it mirrors `compose`, which emits no join before the first emitted block — and both callers pin `lead_join == Join::Rule` separately, which matters because `BASE_PM`'s lead join is emitted in the real package even though `PM_INSTRUCTIONS`' is not. The `debug_assert!` is a forward tripwire for a future widening of the filter, correctly scoped and not overclaimed.

### Verification (re-run by me, all matching the reported figures)

`cargo test -p trusty-mpm --lib` → `test result: ok. 3833 passed; 0 failed; 6 ignored`, with all five new tests present and `ok`. `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → 0 errors. `cargo fmt --check` clean. `scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`. SLOC 433 / 293 / 561, exactly as reported. Zero dangling `Test:` doc pointers (scripted). One `## [Unreleased]`.

### Notes — two tidies, neither a finding, neither worth another round

- `bundled_pm_package_tests.rs:530` says "both calls below scan the ambient agent tiers independently". They don't — the delegation override means zero scans. The comment is wrong about the mechanism while its conclusion ("cannot race a concurrent agent redeploy") is correct and in fact more strongly true than stated. Fix it in passing if the branch is touched again.
- `CHANGELOG.md:29-31`, "The gate is asserted through `resolve_pm_prompt` itself with a project-tier agent deployed", now sits directly above "The gate injects a fixed roster rather than scanning the agent tiers per side." Both are true of different tests — path selection deploys an agent, byte-equality injects — but adjacent like that they read as contradictory. A clause split would resolve it.
- Across all three rounds: no production-code defect was ever found. Every finding was in test infrastructure or documentation, and `instruction_package.rs` was never touched, so `RosterNotConsumed` / `SectionWithoutBlocks` remain exactly as they were. The composed path, the marker guards, the two out-of-scope configurations and the byte-equality contract are all sound.

Issue #3502, opened 2026-07-20
trusty-gworkspace: harden multi-account token storage (file locking + orphan-default fix)


bobmatnyc:
## Problem

`crates/trusty-gworkspace`'s `TokenStorage` (`src/api/auth/storage.rs`) has no
file locking around its `tokens.json` read-modify-write cycle. Two concurrent
writers — e.g. two profile refreshes racing in the same or different
processes (`OAuthManager::refresh` in `src/api/auth/manager.rs`, `persist()`
in `src/api/auth/oauth/flow/mod.rs`) — can each `load()` the map before either
`save()`s, silently losing one profile's write.

Separately, `crates/trusty-gworkspace/src/cli/accounts.rs::remove()` does not
reassign the default profile when the profile being removed is the current
default, leaving the token map with zero default entries and breaking
`BaseClient::resolve_stored`'s default-profile fallback for every subsequent
call with no explicit `account`.

## Scope

1. Add an advisory cross-process file lock (sidecar `tokens.json.lock`) plus
   in-process synchronization around the full load -> mutate -> save critical
   section, reused by every read-modify-write call site (refresh, consent
   persist, CLI `accounts default`/`accounts remove`). Preserve 0600 perms.
2. `remove <default>` deterministically reassigns the default to another
   remaining profile (alphabetically next) instead of orphaning it; no-op
   (clean removal) when no profiles remain.
3. Add regression tests: concurrent read-modify-write does not lose writes,
   default-reassignment-on-remove behavior.

This is prep for a follow-up issue that exposes multi-account management
(`set_default_account`, `remove_account`, `add_account`) through MCP tools —
today only `list_accounts` is read-only-MCP-exposed; everything else is
CLI-only.

PR #4258, opened 2026-07-28
fix(trusty-mpm): serialise projects.json writes across processes

bobmatnyc:
## The defect (confirmed before changing anything)

`crates/trusty-mpm/src/project/store.rs` — `ProjectStore::upsert` was an
unsynchronised read-modify-write of the whole registry file:

```rust
self.reload_if_changed().await?;              // read
self.data.projects.insert(key, project);      // mutate in memory
self.save().await                             // write the WHOLE file back
```

**No locking or atomicity existed today at the process level.** The only
synchronisation anywhere on this path was `ProjectRegistry`'s
`tokio::sync::RwLock`, which serialises tasks *within one process*.
`projects.json` is written concurrently by the daemon, `tm project …` CLI
invocations and MCP tool calls, so two processes interleaving
read/read/write/write dropped one registration entirely — and **both** callers
saw `Ok`.

A cross-process test found a **second, worse failure** from the same window
that the report did not mention: `save()` wrote to one *fixed* temp path,
`projects.json.tmp`, shared by every writer. Overlapping writes interleaved
into that single scratch file and the rename published the mangled result, so
`projects.json` stopped parsing altogether:

```
child: upsert: Serialize("trailing characters at line 19 column 2")
```

That is registry-wide loss, not a single lost entry.

## The fix — reused the existing primitive choice, added the missing shared entry point

I searched first. There is **no** existing file-locking / atomic-RMW helper in
`trusty-common`. What does exist is the precedent:
`trusty-gworkspace`'s `TokenStorage::update` solves the identical
`load()`→mutate→`save()` lost-update bug for `tokens.json` (#3502), and
`fd-lock` is already a `[workspace.dependencies]` entry added *for that exact
purpose*. But it is private to gworkspace and `anyhow`-based.

So rather than inline a one-off in `store.rs`, this adds
**`trusty_common::json_rmw`** — the single implementation of that critical
section, where epic #4207 slice 3 can build on it:

- exclusive advisory lock on a `<path>.lock` sidecar (a lock on the document
  itself would be discarded by the rename that publishes each new version);
- re-reads the document **under the lock** — a caller's in-memory copy is never
  trusted;
- publishes atomically: **per-writer-unique** temp file → `fsync` → `rename` →
  directory `fsync`. Uniqueness kills the shared-`.tmp` corruption class
  independently of the lock.

`ProjectStore` now funnels every mutation through one `mutate` entry point
built on it, run via `spawn_blocking` so the blocking lock syscall never parks
a runtime worker. The lock-bypassing `save()` is removed so no second write
path can reintroduce the race.

**Call sites checked for reentrancy / lock ordering.** `PATCH
/api/v1/projects/{name}` went through `ProjectRegistry::update_with`, which did
`get` → mutate → `upsert` — two separate store calls. Guarding only `upsert`
would have left that path racy across processes while looking fixed, so it now
delegates to `ProjectStore::update_with` and runs inside **one** held lock.
Nothing calls a mutating store method from inside a `mutate` closure (that
would self-deadlock, and the module docs say so). Lock ordering is uniform: the
async `RwLock` is always taken before the file lock, never the reverse.

**Does not fail open.** A lock that cannot be acquired returns the new
`ProjectStoreError::Lock` rather than proceeding unsynchronised; an
unreadable-but-present file is never silently replaced by an empty one (only a
genuine `NotFound` starts from `Default`); a closure returning `Err` aborts the
write with the file byte-for-byte unchanged. Readers deliberately take **no**
lock — atomic rename means they always see a complete document, so a reader can
never deadlock against a writer.

## Test evidence — failing before, passing after

`crates/trusty-mpm/tests/projects_json_concurrency.rs` spawns **real child
processes** (re-exec of the test binary via `#[ignore]`d helper entry points),
all released from a shared readiness barrier so startup skew cannot serialise
them by accident. Multi-process, not threads — the failure mode it observes is
the shipping one.

**Before the fix** (reproduced on every one of 3 consecutive runs):

```
thread 'multiprocess_child_writer' panicked at tests/projects_json_concurrency.rs:133:18:
child: upsert: Serialize("trailing characters at line 19 column 2")
...
thread 'projects_json_multiprocess_upsert_no_lost_updates' panicked at :197:9:
child p1 failed: exit status: 101
test result: FAILED. 1 passed; 1 failed; 2 ignored; 0 measured; 0 filtered out
```

**After the fix** (5 consecutive runs, all green):

```
test result: ok. 2 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 2.00s
test result: ok. 2 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 1.68s
test result: ok. 2 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 2.09s
test result: ok. 2 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 1.60s
test result: ok. 2 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 1.99s
```

Also covered, per the brief:

- `projects_json_survives_killed_writer` — a writer SIGKILLed mid-write leaves
  `projects.json` parseable with its pre-existing entry intact.
- `store_lock_contention_does_not_deadlock` — interleaved writers and readers
  complete under a wall-clock timeout.
- `store_lock_failure_is_not_fail_open` — an unopenable lock errors and the
  prior record is untouched (no unsynchronised fallback).
- `store_leaves_no_temp_file_behind`, `store_survives_stale_temp_files` — a
  valid file is never replaced by an empty or partial one.
- `json_rmw` unit tests cover each contract clause, including
  `update_write_failure_leaves_original_intact` and
  `update_corrupt_file_errors` (a corrupt file is preserved for the operator,
  never overwritten).

## Gates (raw output in the thread)

- `cargo test -p trusty-mpm` — **6601 passed; 0 failed; 16 ignored**, exit 0.
  (The 2 new ignores are the child-process helpers, which run only when spawned.)
- `cargo test -p trusty-common` — **205 passed; 0 failed**, exit 0.
- `cargo clippy -p trusty-mpm -p trusty-common --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- `bash scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK`.
- `bash scripts/check_sld.sh` — `0 error(s), 0 warning(s)`.

Store unit tests were moved to `store_tests.rs` (pure code motion, no assertion
changes) to keep `store.rs` under the 500-SLOC production cap, following the
existing `registry.rs` / `registry_tests.rs` pattern.

## Note for reviewers

`trusty-common` gates `thiserror` behind an optional feature (`default = []`),
and `json_rmw` is an unconditional module, so `JsonRmwError` implements
`Display`/`Error` by hand rather than forcing `thiserror` into every minimal
build. `ProjectStoreError` still uses `thiserror` and converts via `From`.

`Cargo.lock` changes by exactly one line (`fd-lock` under `trusty-common`) — no
lockfile refresh.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11431,6 +11431,7 @@ dependencies = [
  "dirs 5.0.1",
  "dotenvy",
  "fastembed",
+ "fd-lock",
  "futures",
  "futures-util",
  "genco",
--- a/crates/trusty-common/CHANGELOG.md
+++ b/crates/trusty-common/CHANGELOG.md
@@ -7,6 +7,20 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Added
+
+- `json_rmw`: cross-process locked read-modify-write for whole-file JSON
+  documents — the single implementation of the load → mutate → save critical
+  section that `trusty-mpm`'s `projects.json`, `trusty-gworkspace`'s
+  `tokens.json` (#3502) and the epic #4207 worktree registry all need.
+  `json_rmw::update` takes an exclusive advisory lock on a `<path>.lock`
+  sidecar, re-reads the document under that lock (never trusting a caller's
+  stale copy), applies the mutation, and publishes atomically via a
+  per-writer-unique temp file + `fsync` + `rename` + directory `fsync`. Never
+  fails open: a failed lock, read, parse or write returns `Err` with the
+  document byte-for-byte unchanged, and only a genuinely absent file starts
+  from `Default`. Adds `fd-lock` as an unconditional dependency.
+
 ---
 ## [0.27.0] — 2026-07-27
 
--- a/crates/trusty-common/Cargo.toml
+++ b/crates/trusty-common/Cargo.toml
@@ -91,6 +91,10 @@ tracing-subscriber = { workspace = true }
 # `Layer` impl in `log_buffer` also uses tracing-subscriber's `registry`
 # feature (enabled in the workspace dep).
 sysinfo = { workspace = true }
+# Cross-process advisory file locking for the `json_rmw` locked
+# read-modify-write critical section (trusty-mpm's projects.json lost-update
+# bug; same primitive trusty-gworkspace already uses for tokens.json, #3502).
+fd-lock = { workspace = true }
 dirs = { workspace = true }
 colored = { workspace = true }
 reqwest = { workspace = true }
--- a/crates/trusty-common/src/json_rmw.rs
+++ b/crates/trusty-common/src/json_rmw.rs
@@ -1,0 +1,311 @@
+//! Cross-process read-modify-write for a whole-file JSON document.
+//!
+//! Why: several trusty crates persist a small JSON document that multiple
+//! independent PROCESSES mutate — `trusty-mpm`'s `projects.json` registry,
+//! `trusty-gworkspace`'s `tokens.json` (issue #3502), and the dual worktree
+//! registry epic #4207 will reconcile. Every one of them is a
+//! load → mutate → save-the-whole-file cycle. With no cross-process
+//! serialisation, two writers that interleave read/read/write/write silently
+//! lose one of the two updates and BOTH callers see success; worse, if they
+//! share one temp path they can publish a half-written document and corrupt the
+//! file outright. An in-process `Mutex` cannot fix either failure because the
+//! writers are separate processes. This module is the single implementation of
+//! that critical section so each call site inherits the fix instead of
+//! re-deriving it.
+//! What: [`update`] takes an advisory exclusive lock on a `<path>.lock` sidecar,
+//! re-reads the document from disk under that lock (never trusting a caller's
+//! possibly-stale copy), applies the caller's mutation, and publishes the result
+//! atomically — unique temp file, `fsync`, `rename`, `fsync` of the parent
+//! directory — before releasing the lock.
+//! Test: `cargo test -p trusty-common -- json_rmw::tests`.
+//!
+//! # Atomicity contract
+//!
+//! Guarantees a caller may rely on:
+//!
+//! 1. **Serialisation.** The read, the mutation and the write happen while one
+//!    writer holds an exclusive advisory lock, so no other [`update`] on the
+//!    same path can observe or overwrite the intermediate state. The lock is
+//!    `flock(2)`-style: it is held by the open file description, so it
+//!    serialises separate processes AND separate threads that each call
+//!    [`update`], on Unix and Windows alike.
+//! 2. **All-or-nothing publish.** Readers of `path` see either the complete
+//!    previous document or the complete new one, never a partial write. The
+//!    temp file carries the writer's pid and a nanosecond stamp, so concurrent
+//!    writers never share a scratch path even if one of them bypasses the lock.
+//! 3. **Never fail open.** Every failure — lock acquisition, read, parse,
+//!    serialise, write, rename — returns `Err` and leaves `path` byte-for-byte
+//!    unchanged. There is no path on which a failed update advances state, and
+//!    an unreadable-but-present file is never silently replaced with a default
+//!    (empty) document; only a genuinely absent file starts from
+//!    [`Default`].
+//! 4. **Crash safety.** A writer killed at any point leaves either the previous
+//!    document intact or an orphaned `*.tmp` file that no reader consults.
+//!
+//! Explicitly NOT guaranteed:
+//!
+//! - **Advisory, not mandatory.** A process that writes `path` without going
+//!   through [`update`] is not blocked. Every writer of a given file must use
+//!   this entry point.
+//! - **Not reentrant.** [`update`] must not be called from inside another
+//!   [`update`] closure on the same path: the second acquisition uses a
+//!   different file descriptor and will self-deadlock.
+//! - **Blocking.** Lock acquisition blocks the calling thread. Async callers
+//!   must run [`update`] on a blocking-safe thread (e.g.
+//!   `tokio::task::spawn_blocking`).
+
+use std::fs::{File, OpenOptions};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::Serialize;
+use serde::de::DeserializeOwned;
+
+/// Failure modes of a locked JSON read-modify-write.
+///
+/// Why: callers must be able to tell a lock-contention/permission problem apart
+/// from a corrupt document, because the remedies differ (retry / operator
+/// intervention vs. restore the file). Every variant carries the path it
+/// concerns so a multi-file caller can report which document failed.
+/// What: one variant per stage of the cycle — lock, I/O, (de)serialisation.
+/// `Display`/`Error` are implemented by hand rather than derived: this crate
+/// keeps `thiserror` behind an optional feature (`default = []`), and `json_rmw`
+/// is an unconditional module, so deriving would force `thiserror` into every
+/// minimal build of `trusty-common`. Crates that own their own error types
+/// should still prefer `thiserror` and convert via [`From`].
+/// Test: `update_lock_path_unopenable_errors`, `update_corrupt_file_errors`.
+#[derive(Debug)]
+pub enum JsonRmwError {
+    /// The advisory lock could not be created or acquired.
+    Lock {
+        /// The document the lock guards (not the sidecar itself).
+        path: PathBuf,
+        /// The underlying OS error.
+        source: std::io::Error,
+    },
+
+    /// Reading, writing, renaming or syncing the document failed.
+    Io {
+        /// The document being read or published.
+        path: PathBuf,
+        /// The underlying OS error.
+        source: std::io::Error,
+    },
+
+    /// The document could not be parsed, or the new value could not be encoded.
+    Serialize {
+        /// The document being parsed or encoded.
+        path: PathBuf,
+        /// The `serde_json` message.
+        message: String,
+    },
+}
+
+impl std::fmt::Display for JsonRmwError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Lock { path, source } => write!(
+                f,
+                "could not acquire the update lock for {}: {source}",
+                path.display()
+            ),
+            Self::Io { path, source } => write!(
+                f,
+                "json read-modify-write I/O error on {}: {source}",
+                path.display()
+            ),
+            Self::Serialize { path, message } => write!(
+                f,
+                "json read-modify-write serialization error on {}: {message}",
+                path.display()
+            ),
+        }
+    }
+}
+
+impl std::error::Error for JsonRmwError {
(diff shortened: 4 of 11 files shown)


bobmatnyc:
## VERDICT: APPROVE

Adversarial review of `45365f8c` from an independent read-only worktree. Zero CRITICAL, zero HIGH at >80% confidence. Two MEDIUM and two LOW findings below — none block the merge.

I did not accept the design narrative. Every claim in the module docs was checked against the code, and the central one (that the lock is genuinely held across the whole read-modify-write) was checked by **mutation testing**, not by reading.

---

### Mutation evidence — the tests are not theatre

**Mutation A — revert `store.rs`/`registry.rs` to `origin/main` (the shipping pre-fix code), keep the new test file:**

```
thread 'multiprocess_child_writer' panicked at tests/projects_json_concurrency.rs:133:18:
  child: upsert: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
test projects_json_multiprocess_upsert_no_lost_updates ... FAILED
  child p0 failed: exit status: 101
```

Red — and red via exactly the failure the PR describes: four writers colliding on the one fixed `projects.json.tmp` so `rename` hits `ENOENT`.

**Mutation B — keep the whole fix but drop *only* the lock guard** (`drop(rw.write()?)` instead of `let _guard = rw.write()?`; unique temp path, `fsync`, rename and re-read all left intact, no artificial sleep added). This isolates the lock from every other change:

```
run 1: lost 59 of 96 concurrent upserts (file holds 37)
run 2: lost 58 of 96 concurrent upserts (file holds 38)
run 3: lost 57 of 96 concurrent upserts (file holds 39)
```

3/3 red, ~60% of updates silently lost. `projects_json_multiprocess_upsert_no_lost_updates` pins the lock itself, not merely the temp-path change. That is also the empirical proof that `_guard`'s lifetime is correct: the named binding really does live to the end of `update()`, past `publish_atomic`, and drops before `rw` (reverse declaration order). No end-of-statement guard drop.

**Unmutated, clean rebuild — green and not flaky:**

- `cargo test -p trusty-common --lib json_rmw` → `test result: ok. 8 passed; 0 failed; 0 ignored`
- `cargo test -p trusty-mpm --test projects_json_concurrency` → `test result: ok. 2 passed; 0 failed; 2 ignored` (the 2 ignored are the child entry points, by design)
- **12/12 consecutive green** unloaded
- **6/6 consecutive green under load average ≈ 30** (12 CPU burners pinned during the runs)
- `cargo test -p trusty-mpm --lib project::` → `test result: ok. 101 passed; 0 failed; 0 ignored`
- `bash scripts/check_line_cap.sh` → `line-cap: 7 allowlisted, 0 violations — OK.`
- `bash scripts/check_test_pointers.sh` → exit 0
- `cargo clippy -p trusty-common -p trusty-mpm --all-targets` → no new warnings

The multi-process test is **genuinely multi-process**: `spawn_child` re-execs `std::env::current_exe()` with `--ignored --exact --nocapture <helper>` and real child stdout interleaves in the output. The barrier is a real file barrier (every child writes `ready-<tag>`, parent writes `go` only after all four are polling). CI runs plain `cargo test --workspace` on a libtest harness, so the child argv is correct there too. **I judge it non-flaky**: the only timing assumption is a 60 s barrier/exit budget for work that completes in ~1.6 s, and the assertion is on set membership, not on ordering or interleaving.

---

### Correctness checks that passed

| Check | Result |
|---|---|
| Lock held across the *entire* RMW (re-read strictly after acquire, publish strictly before release) | ✅ `json_rmw.rs:297-306`; proven by Mutation B |
| Any other writer bypassing `json_rmw`? | ✅ None. `data_dir.join("projects.json")` appears exactly once in the workspace (`store.rs:160`); all mutations reach it via `ProjectStore::{upsert,update_with}` → `mutate`. `~/.trusty-agents/projects.json` is a different file in a different crate. |
| Sidecar creation race | ✅ `O_CREAT` without `O_EXCL`; both racers land on the same inode and `flock` conflicts on the inode. |
| Sidecar ever deleted (would break mutual exclusion) | ✅ Never — grepped every `remove_file` in `trusty-mpm`/`trusty-installer`. Notably `pm_guard_budget/mod.rs:76` documents this exact "a loser's `remove_file` deletes a live lock" hazard; `json_rmw` correctly avoids it. |
| Lock is on a stable sidecar, not the renamed-over inode | ✅ `lock_path()` — correct, and the right reason is documented. |
| Reentrancy / lock-order inversion | ✅ Structurally impossible: `mutate`'s closure gets `&mut HashMap`, `update_with`'s gets an owned `Project` — neither can reach a `&mut ProjectStore`. Registry `RwLock` → file lock ordering is enforced by the type system, not merely documented. |
| Readers take no file lock and cannot see a torn file | ✅ Publish is `rename(2)`; a reader's `read_to_string` opens either the old or the new inode, and neither is ever written in place. |
| Fail-open: lock failure | ✅ `ProjectStoreError::Lock`, no fall-through. Pinned by `store_lock_failure_is_not_fail_open` + `update_lock_path_unopenable_errors`, both of which also assert the pre-existing record survives. |
| Fail-open: panicked/cancelled blocking task | ✅ `store.rs:277-284` maps `JoinError` to `Err`, never `Ok`. Guard drops on unwind, so a panicking closure releases the lock. |
| Fail-open: only genuine `NotFound` starts from `Default` | ✅ `read_or_default` (`json_rmw.rs:195-201`); `update_corrupt_file_errors` asserts a malformed file is *preserved*, not overwritten. |
| Fail-open: failed publish | ✅ `update_write_failure_leaves_original_intact` — target byte-identical after an `Io` error. |
| Durability ordering: write → fsync file → rename → fsync dir | ✅ `publish_atomic` `json_rmw.rs:220-243`, in that order. |
| Cross-platform | ✅ Directory `fsync` is `#[cfg(unix)]`-gated with a correct rationale. CI is `ubuntu-latest` only, so the Windows path is unexercised — but `trusty-mpm` is a tmux-based tool that does not target Windows. |
| `unwrap()` on reachable paths | ✅ None in `json_rmw.rs` (only `unwrap_or_default` / `unwrap_or_else` on infallible fallbacks). |
| `thiserror` for library errors | ✅ `ProjectStoreError` uses it. `JsonRmwError` hand-implements `Display`/`Error` — the stated reason (keeping `thiserror` behind `trusty-common`'s optional feature while `json_rmw` is unconditional) checks out against `trusty-common/Cargo.toml`. |
| Placement in `trusty-common` / duplication | ✅ Right home. `memory_core/store/concurrent_open.rs` is genuinely redb-specific (it exists to work *around* redb's own `LOCK_EX` via snapshot copies) — the author's judgement is correct. `trusty-gworkspace`'s `TokenStore::update` (`api/auth/storage/mod.rs:280-309`) is a near-identical hand-rolled version; folding it onto `json_rmw` is #3502's job, not this PR's. |
| CHANGELOG | ✅ Purely additive — `git diff origin/main...HEAD -- '*CHANGELOG.md'` has **0** deletion lines, one `## [Unreleased]` per file. Branch is exactly one commit on top of `eba04d67` (#4249), so the rebase artifact is resolved. |
| `Test:` pointers | ✅ `check_test_pointers.sh` exit 0; every cited fn exists. |
| CI | ✅ All 26 checks green. |

---

### Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-common/src/json_rmw.rs` | 177-185, 215-245 | Orphaned `*.tmp` files are never reaped. Uniquifying the scratch path fixes the corruption bug but converts crash debris from *one* self-overwriting file into *unbounded* accumulation in `~/.trusty-mpm/`. Measured: **9 orphans out of 30 SIGKILLed writers** (`projects.json.<pid>.<nanos>.tmp`), nothing in the workspace ever removes them. `store_survives_stale_temp_files` proves they are harmless, which is exactly why nothing will ever notice them growing. | Reap while the lock is already held, at the top of `update()`: read `path.parent()`, remove siblings matching `<file_name>.*.tmp` whose mtime is older than a few minutes. Safe under the lock — no live writer's temp can be that old. |
| MEDIUM | `json_rmw.rs` 297; `project/store.rs` 265; `project/registry.rs` 84, 96 | — | Lock acquisition is unbounded-blocking with no timeout, and it runs while the caller holds `ProjectRegistry`'s async `RwLock` — which `get()` and `list()` **also** take in write mode. Concrete scenario: a `tm project …` CLI process is suspended (Ctrl-Z / SIGSTOP) inside `mutate` while holding the flock; the daemon's next `GET /api/v1/projects` blocks forever on `store.write()`, and so does every subsequent project read and any session spawn that resolves a project. No timeout, no `warn!`, no diagnostic — the daemon just goes quiet. (A *crashed* holder is fine: the kernel drops the flock.) | Replace the blocking `rw.write()` with a `rw.try_write()` retry loop against a deadline (e.g. 10 s), returning `JsonRmwError::Lock` on expiry, and `warn!` with the sidecar path after the first failed attempt so the stall is visible. |
| LOW | `crates/trusty-common/src/json_rmw.rs` | 240-243 | The directory `fsync` result is discarded (`if let Ok(dir) … let _ = dir.sync_all()`), while contract item 3 says "Every failure — lock acquisition, read, parse, serialise, write, rename — returns `Err`". The wording technically excludes the dir sync, but item 4's crash-safety claim leans on it: if the dir sync fails, a power loss can lose the rename and silently revert to the previous document. | At minimum `warn!` on failure; ideally propagate as `JsonRmwError::Io`. Either way, say so in the contract. |
| LOW | `crates/trusty-mpm/tests/projects_json_concurrency.rs` | 224-267 | `projects_json_survives_killed_writer` **passes against the unmodified pre-fix implementation** — I ran it. Its doc rationale cites the shared-temp-path corruption ("If concurrent writers share one temp path… a crash can expose a partial file"), but the test has a single writer and so cannot exercise that at all. It is a valid invariant check, not a regression test for anything this PR changes. | Keep the test; correct the `Why:` to state it pins the atomic-publish invariant generally. The shared-temp-path corruption *is* covered — that is precisely how `projects_json_multiprocess_upsert_no_lost_updates` went red in Mutation A. |

No Required Changes. All four findings are follow-up quality items.

---

### Notes

- **Pre-existing, deliberately not flagged (unchanged code, not CRITICAL).** `ProjectStore` samples `sig_of()` *after* the lock is released (`store.rs:287`) and after the read in `read_file` (`store.rs:202`). A write landing in that gap leaves `last_sig` matching the *other* writer's file while `self.data` holds ours, so `reload_if_changed` no-ops and reads stay stale until the next write. `main`'s `save()` has the identical shape, so this PR neither introduces nor worsens it — but it is the natural next slice if someone hardens this path.
- **Reproducibility caveat for anyone re-running the mutation experiments.** Two worktrees of this repo sharing one `CARGO_TARGET_DIR` collide: cargo produced a single `libtrusty_common-619de9ff7f05fb09.rlib` for both source trees, so a mutated build silently overwrote the unmutated one. I hit this, caught it, `cargo clean -p trusty-common -p trusty-mpm`'d, and re-ran everything above from a clean per-worktree target. Use a per-worktree target directory.
- **Not flagged.** `ProjectRegistry::update_with` gains a `Send + 'static` bound on its public generic — technically SemVer-breaking, but no in-workspace crate depends on `trusty-mpm` as a library and the only callers are the PATCH route and tests.
- **Environmental, out of scope.** `flock` over NFS is unreliable, so a `~/.trusty-mpm` on a network home directory could still lose updates. The atomic rename keeps the file parseable regardless.

Issue #1373, opened 2026-06-17
trusty-mpm: MCP search resolves to wrong index (claude-mpm) instead of the project-path index


bobmatnyc:
## Problem
trusty-mpm sessions sometimes query the wrong trusty-search index (usually `claude-mpm`) instead of the index for the current project's path. mpm should ALWAYS find-or-create and use the index for its own project path.

## Root cause (architectural)
- trusty-mpm never derives, registers, or pins an index. It injects a contextless `trusty-search serve` stub (`crates/trusty-mpm/src/core/session_launch/settings.rs:115`) with no `--index`/`--project`/path, leaving index selection to the LLM at query time.
- The correct path→index logic (`detect_project()` → git-root basename) exists in `crates/trusty-search/src/detect.rs:43` but is CLI-only and never reached from the MCP `serve` path.
- With no project index registered, the LLM picks the most prominent registered index from `list_indexes` — typically the persistent `claude-mpm` index. `search_all`/`grep` without `index_id` also fan out across all indexes.

## Chosen fix (Register + Pin, A+B)
1. At session launch, resolve the project root and find-or-create (`POST /indexes/{id}` with `root_path`) the project's index before injecting the MCP stub.
2. Add an `--index <derived-id>` flag to `trusty-search serve` and pass it in the stub so the session is pinned to that one index; the MCP tool handlers should default/scope `index_id` to the pinned index.

Acceptance: a fresh mpm session in any project always resolves to that project's own index (created if missing) and never silently uses `claude-mpm`. Add tests for index-id derivation from project root and the serve-time pin.

PR #4262, opened 2026-07-28
feat(trusty-common): project-derived index identity as a partitioning key (#4207)

bobmatnyc:
Epic #4207, Track B — **the identity-derivation slice only.** Registry
reconciliation and migration of the ~7 currently-registered indexes are
deliberately deferred to separate slices. **Nothing is wired over** in this PR.

## Why the previous attempt failed, and what changed

PR #4063 was closed won't-do after three BLOCK rounds. The root cause was a
design error, not a bug queue: it promoted `RepoIdentity` — whose own module doc
states it is a **GROUPING** key, deliberately shared across every facet of one
repo — into the **PARTITIONING** slot that trusty-search's
one-`root_path`-per-id registry requires. Each review round found another pair of
genuinely different roots sharing that value: linked worktrees (patched with a
worktree-name suffix), sibling clones of one repo (never fixed, proven live), and
the daemon-unreachable path resolving forward onto the new id.

This derivation **partitions by construction** rather than by patching cases: the
canonical content-tree root path is a hashed component of every id, so two
distinct trees can never collide however their git metadata relates. The grouping
key is retained as a *component* and as the public `ProjectIdentity::origin`
field — not as the whole id.

## What this adds

`crates/trusty-common/src/project_index_id.rs`:

- `ProjectIdentity { origin, root, gh_user }` — the three-part identity the owner
  specified on #4207. `origin` = which repository (grouping, via `RepoIdentity`);
  `root` = which content tree of it (**the partitioning component**); `gh_user` =
  which account operates it.
- `ProjectIdentity::derive()` — the single I/O entry point (git shell-outs + one
  `canonicalize`), offline, best-effort, never panics.
- `ProjectIdentity::index_id()` — **pure**. `<label>-<16 hex>`, e.g.
  `bobmatnyc-trusty-tools-f3ef22158eced5cb`. Because it is pure, every collision
  case is testable with no daemon, no registry, and no live state — and
  resolution can never depend on mutable daemon state, which is what made
  #4063's alias-fallback fragile.
- `derive_project_index_id()`, `resolve_gh_user()`.

Not a worktree UUID (explicitly ruled out by the owner), and not the bare
basename.

**Placement:** `trusty-common`, beside the `index_id` / `repo_identity` /
`github_path` / `slug` modules it builds on rather than duplicates. Index ids are
computed by trusty-mpm, trusty-search, trusty-review and trusty-code, which is
the same single-source-of-truth reason `index_id` was hoisted here in #1373. It
is **unconditional** (not feature-gated) — an identity that varies with a feature
flag is worse than no identity.

**Note on the epic's scope note:** it directs reusing "the shared project
registry's existing function". That function does not exist. Verified:
`project::record::derive_name_from_url` keys purely on the repo-URL basename (no
root, no gh user) and `RepoIdentity::derive` is origin-only. Neither takes three
inputs. This was built, not wired up.

## Collision cases tested

Real synthetic git topologies (`git init` / `git clone` / `git worktree add` in
temp dirs), 14 tests, all passing:

| Case | Test |
|---|---|
| **Sibling clones of one repo** — the case #4063 never fixed | `sibling_clones_of_same_repo_derive_distinct_ids` |
| Linked worktrees of one repo | `linked_worktrees_of_same_repo_derive_distinct_ids` |
| Same repo, different gh users | `different_gh_users_derive_distinct_ids` |
| Directory with no git origin | `directory_without_git_origin_derives_stable_id` |
| Unrelated projects sharing a basename (the original defect) | `unrelated_projects_sharing_a_basename_derive_distinct_ids` |
| Non-injective `{owner}-{repo}` join (#4063 MEDIUM) | `label_ambiguity_does_not_collide` |
| Symlinked root ⇒ *same* id (one tree, one index) | `symlinked_root_derives_same_id_as_real_root` |
| Determinism across calls / from a nested subdir | `derivation_is_deterministic_across_calls` |
| Frozen golden id (guards the stored contract) | `index_id_is_pinned_for_a_known_input` |
| URL-safe single path segment | `index_id_is_a_single_url_safe_segment` |

The two git-topology tests assert **both halves**: that the repo-level grouping
key genuinely *does* collide for those roots (so the test exercises the real
hazard, not a strawman) and that the derived index ids nonetheless do not.

The digest preimage is length-framed with presence tags — closing the
non-injective join flagged on #4063 — and carries a `SCHEME_VERSION` so a future
rule change becomes a distinct id space rather than a silent re-point.

## Explicitly out of scope

- Registry reconciliation (dual-registry) — separate slice.
- Migration of currently-registered indexes — separate slice. **No live index is
  renamed or re-registered by this PR.**
- Wiring into `ensure_project_indexed`, `trusty-search serve`, or the daemon's
  resolution path. Switching over without migration is exactly how #4063 orphaned
  sessions. `git grep` confirms zero call sites outside the new module.

**Constraint documented for the migration slice** (in the module header, not
implemented here): migration must be rename/alias, never re-index (~105k chunks /
2.1 GB redb for this workspace alone), and must fail **SAFE** — fall back to the
known-good legacy id when the daemon or persisted registry cannot be consulted,
**never** forward onto an id that does not exist yet. #4063 failed forward at
`search_index.rs:106-108` and silently orphaned a session's search for its entire
life with no self-heal.

The module also notes the interaction with the epic's self-description item: an
always-suffixed id is not guessable from the project name, which is deliberate
(a guessable id cannot partition). MCP `instructions`-at-initialize and the
`404` → available-index signpost are the complementary fix; neither substitutes
for the other.

## Gates (SPRINT phase, crate-scoped)

```
cargo test -p trusty-common --lib -- project_index_id
  test result: ok. 14 passed; 0 failed; 0 ignored; 203 filtered out

cargo test -p trusty-common
  test result: ok. 211 passed; 0 failed; 6 ignored          (exit 0)

cargo clippy -p trusty-common --all-targets -- -D warnings   exit 0
cargo fmt --check                                            exit 0
  line-cap: 7 allowlisted, 0 violations — OK.                exit 0

cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui
  Finished `dev` profile                                     exit 0
```

The two excluded crates are Tauri GUIs that fail on a missing prebuilt
`ui/dist` (`frontendDist ... doesn't exist`) — pre-existing and environmental,
unrelated to this change.

### DIFF
--- a/crates/trusty-common/CHANGELOG.md
+++ b/crates/trusty-common/CHANGELOG.md
@@ -10,8 +10,9 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - **`project_index_id` — project-derived trusty-search index identity (#4207).**
-  New `ProjectIdentity` (origin + root + gh user) with a pure, deterministic
-  `index_id()`, plus `derive_project_index_id()` and `resolve_gh_user()`. Unlike
+  New `ProjectIdentity` (origin + root + operator) with a pure, deterministic
+  `index_id()`, plus `derive_project_index_id()` and
+  `resolve_operator_identity()`. Unlike
   the basename rule in `index_id` (which collides for unrelated checkouts sharing
   a directory name) and the session-worktree UUID (which binds service identity to
   ephemeral writer isolation), this id *partitions*: the canonical content-tree
@@ -20,6 +21,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   into `ensure_project_indexed`, `trusty-search serve`, or the daemon's resolution
   path; registry reconciliation and migration of existing indexes are separate
   slices of #4207. No behaviour change for any existing caller.
+
+  Derivation is hermetic: it reads no environment variable, so two callers on one
+  tree (e.g. the launchd daemon and a shell CLI) cannot derive different ids. The
+  `index_id()` docs enumerate exactly which inputs are mutable — `origin` moves on
+  the first commit, on `git remote add origin`, and on a new root commit — each
+  pinned by a test, so the migration slice inherits a true guarantee rather than an
+  assumption of permanence.
 
 ---
 ## [0.27.0] — 2026-07-27
--- a/crates/trusty-common/CHANGELOG.md
+++ b/crates/trusty-common/CHANGELOG.md
@@ -7,6 +7,20 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Added
+
+- **`project_index_id` — project-derived trusty-search index identity (#4207).**
+  New `ProjectIdentity` (origin + root + gh user) with a pure, deterministic
+  `index_id()`, plus `derive_project_index_id()` and `resolve_gh_user()`. Unlike
+  the basename rule in `index_id` (which collides for unrelated checkouts sharing
+  a directory name) and the session-worktree UUID (which binds service identity to
+  ephemeral writer isolation), this id *partitions*: the canonical content-tree
+  root is a hashed component, so sibling clones, linked worktrees, and differing
+  accounts derive distinct ids by construction. Derivation only — nothing is wired
+  into `ensure_project_indexed`, `trusty-search serve`, or the daemon's resolution
+  path; registry reconciliation and migration of existing indexes are separate
+  slices of #4207. No behaviour change for any existing caller.
+
 ---
 ## [0.27.0] — 2026-07-27
 
--- a/crates/trusty-common/src/lib.rs
+++ b/crates/trusty-common/src/lib.rs
@@ -456,9 +456,9 @@ pub use index_id::{derive_index_id, find_git_root, resolve_project_root};
 /// compute the identical id — the same single-source-of-truth rule `index_id`
 /// was hoisted here for (#1373). Unconditional (not feature-gated) because an
 /// identity that varies with a feature flag is worse than none.
-/// What: exposes [`project_index_id::ProjectIdentity`] (origin + root + gh user,
+/// What: exposes [`project_index_id::ProjectIdentity`] (origin + root + operator,
 /// with a pure `index_id()`), [`project_index_id::derive_project_index_id`], and
-/// [`project_index_id::resolve_gh_user`]. Derivation only — deliberately wired
+/// [`project_index_id::resolve_operator_identity`]. Derivation only — wired
 /// into no resolution path; registry reconciliation and migration of existing
 /// indexes are separate slices of #4207.
 /// Test: `cargo test -p trusty-common -- project_index_id`.
--- a/crates/trusty-common/src/lib.rs
+++ b/crates/trusty-common/src/lib.rs
@@ -444,6 +444,27 @@ pub use slug::slugify_string;
 pub mod index_id;
 pub use index_id::{derive_index_id, find_git_root, resolve_project_root};
 
+/// Project-derived trusty-search index identity — the PARTITIONING key
+/// (epic #4207; supersedes the approach closed as won't-do in #4063).
+///
+/// Why: [`index_id::derive_index_id`]'s bare basename collides for unrelated
+/// checkouts sharing a directory name, and a tm session gets its worktree UUID
+/// as the id — so service identity is bound to ephemeral writer isolation and
+/// BASE_PM's "pass the project name" instruction 404s. Both need ONE id derived
+/// from the PROJECT. It lives here, beside `index_id`/`repo_identity`/`slug`,
+/// because trusty-mpm, trusty-search, trusty-review and trusty-code must all
+/// compute the identical id — the same single-source-of-truth rule `index_id`
+/// was hoisted here for (#1373). Unconditional (not feature-gated) because an
+/// identity that varies with a feature flag is worse than none.
+/// What: exposes [`project_index_id::ProjectIdentity`] (origin + root + gh user,
+/// with a pure `index_id()`), [`project_index_id::derive_project_index_id`], and
+/// [`project_index_id::resolve_gh_user`]. Derivation only — deliberately wired
+/// into no resolution path; registry reconciliation and migration of existing
+/// indexes are separate slices of #4207.
+/// Test: `cargo test -p trusty-common -- project_index_id`.
+pub mod project_index_id;
+pub use project_index_id::{ProjectIdentity, derive_project_index_id};
+
 /// Shared best-effort trusty-search "ensure this project is indexed" helper
 /// (issues #1373 / #1908), gated behind the `search-index` feature.
 ///
--- a/crates/trusty-common/src/project_index_id.rs
+++ b/crates/trusty-common/src/project_index_id.rs
@@ -1,0 +1,373 @@
+//! Project-derived trusty-search index identity — a PARTITIONING key
+//! (epic #4207, replacing the approach closed as won't-do in #4063).
+//!
+//! Why: the index id in use today is either the bare directory basename
+//! ([`crate::index_id::derive_index_id`]) — which collides for any two
+//! unrelated checkouts that happen to share a directory name — or, for a
+//! tm-managed session, the session/worktree UUID, which binds a service
+//! identity to ephemeral writer isolation: N concurrent writers on one repo
+//! produce N independently-stale indexes, none shareable, and the id is
+//! unguessable so BASE_PM's instruction to pass the project name silently
+//! 404s. Both defects need ONE deterministic id derived from the PROJECT.
+//!
+//! **The contract that sank the previous attempt — state it before reading
+//! any code.** trusty-search's registry is one-`root_path`-per-id (see
+//! `trusty-mpm`'s `session_manager::search_gc`), so an index id must
+//! *partition*: one id maps to exactly ONE content tree. [`RepoIdentity`]'s
+//! own module doc says it is a *grouping* key — deliberately shared by every
+//! facet (live checkout, `.base` clone, each session worktree) of one repo.
+//! #4063 promoted that grouping key into the partitioning slot, and each
+//! review round found another pair of genuinely different roots sharing it:
+//! linked worktrees, then sibling clones of one repo, then the
+//! daemon-unreachable path resolving forward onto the new id. This module
+//! partitions **by construction** instead of by patching cases: the canonical
+//! content-tree root path is a hashed component of every id, so two distinct
+//! trees can never collide no matter how their git metadata relates. The
+//! grouping key is retained as a *component* (and as the public
+//! [`ProjectIdentity::origin`] field), not as the whole id.
+//!
+//! What: [`ProjectIdentity`] is the three-part identity the owner specified on
+//! #4207 — origin (which repository), root (which content tree of it), and
+//! gh user (which account operates it). [`ProjectIdentity::derive`] is the one
+//! I/O entry point (git shell-outs + one `canonicalize`);
+//! [`ProjectIdentity::index_id`] is a pure, dependency-free function over
+//! those three fields, so every collision case is testable without a daemon,
+//! a registry, or live state. [`derive_project_index_id`] is the one-call
+//! convenience wrapper. Nothing here is wired into `ensure_project_indexed`,
+//! `trusty-search serve`, or the daemon's resolution path — this slice is the
+//! derivation only.
+//!
+//! **Constraint inherited by the migration slice (do NOT implement here).**
+//! The ~7 currently-registered indexes must be migrated by rename/alias, never
+//! by re-index (this workspace alone is ~105k chunks / 2.1 GB redb). Whatever
+//! resolver performs that migration MUST fail **safe**: when the daemon or the
+//! persisted registry cannot be consulted, fall back to the known-good LEGACY
+//! id. It must never fail *forward* onto an id that does not exist yet —
+//! #4063 did exactly that at `search_index.rs:106-108` and silently orphaned a
+//! session's search for the session's entire life, with no self-heal.
+//!
+//! Discoverability note: because the id is always suffixed, it is not
+//! guessable from the project name alone. That is deliberate — a guessable id
+//! cannot partition. The #4207 self-description slice (MCP `instructions` at
+//! initialize, `404` → available-index signpost) is what makes ids findable;
+//! the two changes are complements, and neither substitutes for the other.
+//!
+//! Test: `cargo test -p trusty-common -- project_index_id` — the sibling
+//! `project_index_id_tests.rs` covers sibling clones, linked worktrees,
+//! differing gh users, remoteless directories, symlinked roots, determinism,
+//! and the id charset.
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use crate::repo_identity::RepoIdentity;
+use crate::slug::slugify_string;
+
+/// Version tag mixed into every digest preimage.
+///
+/// Why: the derivation rule will eventually change (a new component, a
+/// different framing). Mixing a version tag in means a future `v2` produces a
+/// wholly different id space rather than silently re-pointing existing ids at
+/// new content, so the migration slice can tell the two apart.
+/// What: the literal `"v1"`.
+const SCHEME_VERSION: &str = "v1";
+
+/// Maximum length of the human-readable label prefix of an index id.
+///
+/// Why: an index id is a single URL path segment and appears in filenames and
+/// log lines; an unbounded owner/repo pair makes both unwieldy. Truncation is
+/// safe because uniqueness lives entirely in the digest suffix, never in the
+/// label.
+/// What: 48 characters.
+const MAX_LABEL_LEN: usize = 48;
+
+/// Placeholder label used when a project yields no slug-able name at all.
+///
+/// Why: `index_id` must never return a string that starts with `-` or is only
+/// a digest; a stable placeholder keeps the id readable and well-formed.
+/// What: the literal `"project"`.
+const FALLBACK_LABEL: &str = "project";
+
+/// The three-part identity of one indexable project.
+///
+/// Why: #4207 specifies identity as origin + root + gh user. Modelling those as
+/// explicit fields — rather than folding them into an opaque derivation — makes
+/// each component's job checkable in isolation and lets [`Self::index_id`] stay
+/// a pure function, so the collision cases that killed #4063 are testable
+/// without any live daemon or registry state.
+/// What: `origin` is the repo-level GROUPING key ([`RepoIdentity`], `None`
+/// outside a git repo); `root` is the canonical absolute content-tree root and
+/// is the component that makes the derived id a PARTITIONING key; `gh_user` is
+/// the account operating this checkout. All three are hashed; only `origin`
+/// (or the root basename) contributes to the readable label.
+/// Test: `sibling_clones_of_same_repo_derive_distinct_ids`,
+/// `linked_worktrees_of_same_repo_derive_distinct_ids`,
+/// `different_gh_users_derive_distinct_ids`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ProjectIdentity {
+    /// Repo-level grouping key from the git origin remote, `None` outside a
+    /// git repo. Deliberately NOT sufficient on its own to identify an index.
+    pub origin: Option<RepoIdentity>,
+    /// Canonical absolute root of THIS content tree (a clone root, or a linked
+    /// worktree root — each is its own tree). The partitioning component.
+    pub root: PathBuf,
+    /// The account operating this checkout, `None` when unresolvable offline.
+    pub gh_user: Option<String>,
+}
+
+impl ProjectIdentity {
+    /// Derive the identity of the project containing `start`.
+    ///
+    /// Why: the three components must be gathered by ONE rule wherever an id is
+    /// computed (trusty-mpm at session launch, trusty-search's `detect_project`,
+    /// trusty-review's report enrichment) or the callers silently diverge — the
+    /// same failure mode `index_id` was centralised here to prevent (#1373).
+    /// What: canonicalises `start` (resolving symlinks, so two paths naming one
+    /// tree derive one id), walks to the enclosing git root via
+    /// [`crate::index_id::resolve_project_root`] — which lands on the *linked
+    /// worktree* root for a worktree, because its `.git` is a file there, so
+    /// each worktree is correctly treated as its own content tree — then reads
+    /// the origin via [`RepoIdentity::derive`] and the account via
+    /// [`resolve_gh_user`]. Best-effort and offline: every component degrades to
+    /// `None` rather than failing, and `canonicalize` falls back to the input
+    /// path when it cannot resolve. No network, no daemon, no registry.
+    /// Test: `derivation_is_deterministic_across_calls`,
+    /// `directory_without_git_origin_derives_stable_id`,
+    /// `symlinked_root_derives_same_id_as_real_root`.
+    pub fn derive(start: &Path) -> ProjectIdentity {
+        // #4207: canonicalise BEFORE the git-root walk so a symlinked start
+        // path and the real path resolve to the same content tree.
+        let canonical_start = std::fs::canonicalize(start).unwrap_or_else(|_| start.to_path_buf());
+        let root = crate::index_id::resolve_project_root(&canonical_start);
+        ProjectIdentity {
+            origin: RepoIdentity::derive(&root),
+            gh_user: resolve_gh_user(&root),
+            root,
+        }
+    }
+
+    /// Render this identity as a trusty-search index id.
+    ///
+    /// Why: this is the PARTITIONING key defined at the top of this module —
+    /// one id, exactly one content tree — and it must hold *by construction*,
+    /// for sibling clones and linked worktrees alike, because trusty-search's
+    /// registry stores a single `root_path` per id and silently degrades a
+    /// second registration to a *find* against the first tree's root.
+    /// What: `"<label>-<digest>"`, where `label` is the slugified
+    /// `owner-repo` (falling back to the root basename, then [`FALLBACK_LABEL`],
+    /// truncated to [`MAX_LABEL_LEN`]) and `digest` is 16 lowercase hex digits
+    /// over an injectively-framed, version-tagged encoding of ALL THREE fields.
+    /// The label is cosmetic; every bit of uniqueness is in the digest, so
+    /// two projects sharing a label still get different ids. Pure — no I/O, no
+    /// global state, no daemon lookup — which is what makes the collision cases
+    /// testable and keeps resolution independent of mutable live state.
+    /// Guarantees: non-empty; a single URL path segment matching
+    /// `[a-z0-9][a-z0-9-]*`; stable across processes, machines with the same
+    /// three inputs, and daemon restarts.
+    /// Known limitation: `root` is machine-local, so the same repo cloned on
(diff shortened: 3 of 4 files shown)


bobmatnyc:
## VERDICT: WARN

Adversarial review from an isolated read-only worktree at `2aa0ffb3`. Reviewed the spec + code + tests only. Every claim below is backed by an executed experiment, not by reading the PR body.

### The central claim — "partitioning by construction" — is REAL

I attacked it and could not break it.

- **Digest actually depends on the root.** Mutation experiment: deleted `push_field(&mut buf, &path_bytes(&self.root));` from `project_index_id.rs:230`. Result: **6 tests turn RED**, including both target tests.
  ```
  linked_worktrees_of_same_repo_derive_distinct_ids ... FAILED
    assertion `left != right` failed: main checkout vs worktree A:
    ["acme-widget-93046dfcd28eb060", "acme-widget-93046dfcd28eb060", "acme-widget-93046dfcd28eb060"]
  sibling_clones_of_same_repo_derive_distinct_ids ... FAILED
    assertion `left != right` failed: sibling clones are distinct content trees and must not share an index id
  test result: FAILED. 8 passed; 6 failed;
  ```
  This also proves the git-topology tests really executed their assertions on this runner rather than hitting an early `return` — the panic messages carry real derived ids.
- **macOS case-insensitivity is closed.** I expected this to be the hole. It is not: `std::fs::canonicalize` on APFS *does* correct case — probed `…/probe/myrepo` and `…/probe/MYREPO`, both resolve to `…/probe/MyRepo`. Trailing slash, `/.`, and relative→absolute all normalise too.
- **Preimage framing is injective.** Fixed arity 4, each record self-delimiting (`<decimal len>:<bytes>`, digits terminated by a non-digit `:`), `Option` carrying `+`/`-`. I tried to construct two distinct tuples with one preimage and failed. The `foo-bar`/`baz` vs `foo`/`bar-baz` case is genuinely closed, and `label_ambiguity_does_not_collide` (tests.rs:230) pins it with a real `assert_eq!(a.label(), b.label())` precondition.
- **Truncation cannot collide.** It is label-only; `long_label_is_truncated_without_losing_uniqueness` went RED under the mutation, so it is pinning the digest, not the label.
- **`SCHEME_VERSION` reaches the digest.** Mutated `v1`→`v2`: `index_id_is_pinned_for_a_known_input` caught it (`…-6eb95429718e75d0` vs pinned `…-f3ef22158eced5cb`). The golden pin is real.
- **Worktree roots resolve correctly** — `find_git_root` matches `.git` via `exists()`, so a linked worktree's `.git` *file* lands the walk on the worktree root.
- **Symlinks:** two symlinks to one tree derive one id (verified against the real crate). Correct semantics.
- 64-bit FNV-1a + splitmix is adequate here: with ~10⁴ projects, P(accidental collision) ≈ 10⁻¹², and a collision additionally requires an identical label. The doc correctly disclaims any security property.

**Verified untouched:** `grep` across `crates/` finds zero call sites outside the new module and `lib.rs`. Diff is 4 files. `ensure_project_indexed`, `trusty-search serve`, and daemon resolution are not in the diff. Nothing is renamed.

**Gates:** `cargo test -p trusty-common` → `211 passed; 0 failed; 6 ignored`. `check_line_cap.sh` → `0 violations` (373 prod / 413 test). `check_sld.sh` → `0 errors, 0 warnings` (all 14 `Test:` pointers resolve). `cargo fmt --check` clean, `clippy` clean for this crate. One `## [Unreleased]`. No `unwrap`/`expect`/`panic!` anywhere in the production file; the API is infallible by design, so the absence of `thiserror` is correct, not an omission.

**So the failure mode that killed #4063 is gone.** Every finding below is the *opposite* direction — one tree deriving *different* ids (drift ⇒ duplicate or orphaned index), never two trees deriving one id (⇒ merged codebases). That is a strictly weaker hazard, which is why this is WARN and not BLOCK.

---

## Findings

|---|---|---|---|---|
| HIGH | crates/trusty-common/src/project_index_id.rs | 266-271 | `TRUSTY_GH_USER` makes derivation non-hermetic — two processes on the same tree at the same instant derive different ids | Pin both branches with a test instead of skipping; or drop the override |
| HIGH | crates/trusty-common/src/project_index_id.rs | 143, 164-170 | `origin` is mutable git state; ordinary workflows silently change the id, contradicting the stated Guarantees | Add the origin-drift cases to `Known limitation` and pin them with a test |
| MEDIUM | crates/trusty-common/src/project_index_id.rs | 140, 127-128 | macOS firmlinks / Linux bind mounts are not resolved — one tree, two canonical paths, two ids | Narrow the "two paths naming one tree derive one id" claim to symlinks |
| MEDIUM | crates/trusty-common/src/project_index_id_tests.rs | 89-99, 157-166, 302-309 | Six git-topology tests silently `return` (pass) if git misbehaves — including the two carrying the whole guarantee | Probe git once; `expect()` past that point so a broken runner goes red |
| LOW | crates/trusty-common/src/project_index_id.rs | 66-73 | `SCHEME_VERSION` doc claims migration "can tell the two apart"; the emitted id carries no version marker | Reword, or emit the version in the id |
| LOW | crates/trusty-common/src/project_index_id.rs | 265, 115 | `resolve_gh_user` / `gh_user` name a GitHub account but return a git `user.email` | Rename to `operator` / document the value shape on the field |

---

### HIGH-1 — `TRUSTY_GH_USER` makes derivation non-hermetic

`project_index_id.rs:266-271`
```rust
if let Ok(v) = std::env::var("TRUSTY_GH_USER") {
```

**Failure scenario (executed against the real crate, same tree, same instant):**
```
D DIVERGE=true (acme-widget-69d663b676645394  vs  acme-widget-03c581950a37013a)
```
Same content tree, same origin, same git config — the only difference is whether the calling process inherited `TRUSTY_GH_USER`. The `trusty-mpm` daemon (launchd/long-lived env) and a `tm` CLI in the user's shell are precisely a partial-inheritance pair. This is *not* temporal drift — it is two live callers disagreeing at the same moment, which is the exact #1373 "callers silently diverge" failure the module's own Why cites as the reason for centralising the rule here.

It is also the one case the suite opts out of. `project_index_id_tests.rs:397-399`:
```rust
if std::env::var("TRUSTY_GH_USER").is_ok() {
    return; // an explicit override is in effect; the git path is shadowed
}
```
The override branch has **zero** coverage; the fallback branch stops being covered whenever the var is set.

**Fix:** cover both branches deterministically rather than skipping — set the var explicitly in one test and assert the returned value, and in a second test assert the git fallback with the var scoped-unset. If the override is intended purely for ops, state in the `Why` that any wiring slice must guarantee uniform inheritance across daemon and CLI, since a partial rollout partitions one tree into two indexes.

### HIGH-2 — the `origin` component is mutable git state; the Guarantees block over-states stability

`project_index_id.rs:143` → `RepoIdentity::derive` → `root_commit_sha` (`repo_identity.rs:140`).

The `index_id` doc (lines 164-170) reads:
> Guarantees: … stable across processes, machines with the same three inputs, and daemon restarts.
> Known limitation: `root` is machine-local …

Only `root`'s machine-locality is listed as a limitation, and only `resolve_gh_user` (261-263) discloses account drift. Nothing discloses that **`origin` itself moves under ordinary git operations**. Executed against the real crate:

| Operation | before → after |
|---|---|
| `git init`, then the **first commit** (`origin: None` → `ContentHash`) | `nocommit-cc12393ce8d51f11` → `nocommit-9c489b437343053a` |
| `git remote add origin git@github.com:acme/widget.git` (`ContentHash` → `GitHub`) | `nocommit-9c489b437343053a` → `acme-widget-5c056124f35d5d28` |
| `git checkout --orphan` on a remoteless repo (root commit changes) | `orphan-2bf677edd1e5cf42` → `orphan-e21f83ac0762f6ba` |

Row 2 is the everyday `git init` → work → `gh repo create` sequence, and it changes both the digest *and* the label. Under a one-`root_path`-per-id registry an index registered before the remote is added is orphaned the moment it is added, with no self-heal — the same silent-orphan shape called out for the daemon-unreachable path.

This does not break partitioning and nothing is wired, so it is not CRITICAL. It matters because the wiring slice will read that Guarantees block and assume permanence.

**Fix:** extend `Known limitation` on `index_id` to enumerate the three cases above, and add a test that *pins the observed drift* so it is a recorded property rather than a surprise. Stronger option worth considering for the wiring slice: reconcile on `root` (the only component that does not move) and alias on origin/account change, rather than treating the id as permanent.

### MEDIUM-1 — firmlinks / bind mounts defeat the "one tree, one id" claim

`project_index_id.rs:127-128` claims canonicalisation means "two paths naming one tree derive one id". That holds for symlinks only. Probed:
```
/Users/masa                       -> /Users/masa
/System/Volumes/Data/Users/masa   -> /System/Volumes/Data/Users/masa
```
macOS firmlinks are not resolved by `realpath`; Linux bind mounts behave the same. One content tree reached by both paths derives two ids → a duplicate index over identical content. This is inherent to *any* path-based partitioning key, so it is a documentation-accuracy issue, not a design error.

**Fix:** narrow the sentence to "resolving **symlinks**", and add firmlink/bind-mount to `Known limitation`.

### MEDIUM-2 — git-topology tests silently pass when git misbehaves

`project_index_id_tests.rs:89-99`, `157-166`, `302-309` (and 248, 402). Six tests `return` — reported as `ok` — when any git step fails. That includes `sibling_clones_…` and `linked_worktrees_…`, which carry the entire partitioning guarantee. A runner with no git, or with `safe.directory` / `commit.gpgsign` / hook misconfiguration, reports a green suite that asserted nothing. I confirmed by mutation that they *did* execute here, but that confirmation is not in CI.

**Fix:** probe `git --version` once; if git is present, `expect()` every subsequent step so a broken runner goes red instead of silently green. (I acknowledge `repo_identity` set this convention — but those tests do not guard a partitioning key.)

### LOW-1 — `SCHEME_VERSION` doc over-claims

`project_index_id.rs:66-73` says a future `v2` yields an id space the migration slice "can tell the two apart". Verified: bumping the constant changes only the 16 hex digits; `bobmatnyc-trusty-tools-6eb95429718e75d0` is indistinguishable from a v1 id by inspection. The property that actually holds — and is sufficient — is that v2 never re-points an existing id at new content. **Fix:** reword to that, or emit the version in the id (`<label>-v1<digest>`) if classification is genuinely wanted.

### LOW-2 — `gh_user` is an email, not a GitHub account

`project_index_id.rs:265` / field at `:115`. The name says GitHub account; the value is `git config user.email`, which need not map to a GitHub login (and with a multi-valued config returns the *last* entry — verified: `Some("two@t.test")`). The doc explains it, but the name will mislead the wiring slice into comparing it against `gh api user`. **Fix:** rename to `operator` / `resolve_operator_identity`, or state the value shape on the field doc.

---

## Required Changes

1. Cover both `TRUSTY_GH_USER` branches with real assertions instead of the `return` at `project_index_id_tests.rs:397`, and document the uniform-inheritance requirement for the wiring slice. (HIGH-1)
2. Extend the `index_id` `Known limitation` block to cover origin-derived drift (first commit, remote added, orphan branch) and pin it with a test. (HIGH-2)
3. Narrow the canonicalisation claim at `project_index_id.rs:127-128` to symlinks; add firmlinks/bind mounts to the limitations. (MEDIUM-1)
4. Make the git-topology tests fail loudly on a broken-but-present git. (MEDIUM-2)

## Notes

- Placement in `trusty-common`, unconditional, is **correct**. It builds on `RepoIdentity::derive`, `index_id::resolve_project_root`, and `slugify_string` rather than reimplementing any of them, it duplicates nothing, and it compiles under `--no-default-features` (verified). The "an identity that varies with a feature flag is worse than none" reasoning holds, and the dependency-free digest follows from it.
- `--all-features` fails on this machine at `cudarc-0.19.7/build.rs` (`nvcc --version` not found) — pre-existing and environmental, unrelated to this PR.
- Not flagged: 64-bit digest width, the hand-rolled FNV-1a, label collisions, and machine-locality of `root`. All are correct choices for this use and are accurately documented.
- Both HIGH findings are stability, not partitioning. Neither can cause harm while the module has zero call sites; both must be resolved **before** the wiring slice, because that slice will inherit the Guarantees block as written.

bobmatnyc:
## Review response — all 4 Required Changes addressed (`92ed250c`)

Thanks for the mutation testing; re-verified both mutations against the updated
code before pushing.

### HIGH-1 — `TRUSTY_GH_USER`: **removed, not tested**

You offered two options; I took the stronger one. Covering both branches would
have *pinned* the divergence you demonstrated
(`acme-widget-69d663b676645394` vs `acme-widget-03c581950a37013a` on one tree at
one instant) as intended behaviour. Deleting the branch removes the failure
class. Your framing is exactly right — it re-created the #1373 "callers silently
diverge" failure this module's own `Why` cites as its reason to exist, so
shipping it behind test coverage would have been incoherent.

What it was for: an ops/test escape hatch. Neither need is real — the pure
`index_id()` is tested by direct construction, so tests never needed it, and an
ops override on a partitioning key is a footgun by definition.

`resolve_operator_identity` now reads git config only.
`resolve_operator_identity_ignores_ambient_env_override` asserts the source
contains no `env::var` at all, so a future convenience re-adding one fails here.
(Asserted against source text rather than by setting the var: `std::env::set_var`
is process-global and `unsafe` in edition 2024 — a test that set it would corrupt
every parallel test.)

Residual non-hermeticity is now **disclosed rather than claimed away**: git's own
config resolution depends on `HOME`/`GIT_CONFIG_GLOBAL`, and `--get` on a
multi-valued key returns the last entry. The doc states the wiring slice must
guarantee uniform environment inheritance across daemon and CLI.

### HIGH-2 — origin drift: documented as true, and pinned

Not redesigned, per the coordinator's direction. Dropping `origin` would make the
id meaningless and collide across a re-clone; that trade isn't worth making here.

The `Guarantees` block was under-claiming and is rewritten. It now guarantees only
what actually holds — that `index_id()` is a pure function of its three fields —
and states plainly that id stability is only as strong as input stability, with a
table naming each mutable input, the action that moves it, and the test that pins
it:

| Action | Component | Test |
|---|---|---|
| First commit (`None` → `ContentHash`) | `origin` | `id_changes_when_first_commit_lands` |
| `git remote add origin …` | `origin` | `id_changes_when_origin_remote_is_added` |
| New root commit (`--orphan`), remoteless repo | `origin` | `id_changes_on_orphan_root_commit` |
| Changed `user.email` | `operator` | `different_operators_derive_distinct_ids` |

Row 2 is called out as the everyday `git init` → `gh repo create` sequence and as
the same silent-orphan shape the module header already warns about. I also took
your stronger suggestion into the docs as the recommended shape for the wiring
slice: **reconcile on `root`** — the only component that does not move — and treat
an origin/operator change as alias-and-carry-forward, not a new index.

### MEDIUM-1 — firmlinks

Claim narrowed to symlinks. `derive` now carries an explicit limitation:
`/Users/x` and `/System/Volumes/Data/Users/x` each canonicalise to themselves, so
one tree reached through both derives two ids — noted as inherent to any
path-based partitioning key, not specific to this derivation. The symlink test
doc states the narrow scope so nobody reads it as covering mount aliasing.

### MEDIUM-2 — vacuously-passing tests

Every `return` is gone. The `git` helper panics on a failed spawn or non-zero
exit. Also forced `commit.gpgsign=false` per repo so a developer's global signing
config can't produce an *unrelated* red — since they fail loudly now, avoidable
environmental failures had to be designed out rather than skipped around.

**Verified by mutation** (injected a bogus git subcommand into the repo-init
helper): all 9 git-dependent tests go RED, and the message says why. Previously
they reported `ok`.
```
test ...::sibling_clones_of_same_repo_derive_distinct_ids ... FAILED
test ...::linked_worktrees_of_same_repo_derive_distinct_ids ... FAILED
  git ["this-subcommand-does-not-exist"] in "…/repo" exited exit status: 1:
  git: 'this-subcommand-does-not-exist' is not a git command.
```

### LOWs

- `SCHEME_VERSION` reworded to the property that actually holds (a v2 can never
  re-point an already-registered id at new content), explicitly noting the
  emitted id carries **no** version marker so a migration needing to classify ids
  must track the scheme out-of-band.
- Renamed `gh_user` → `operator` and `resolve_gh_user` →
  `resolve_operator_identity`, with `VALUE SHAPE:` on the field stating it is a
  `git config user.email` and **not** a GitHub login — "do not compare against
  `gh api user`", which was your specific concern for the wiring slice. Kept the
  `#4207` "gh user" vocabulary in the prose so traceability to the directive
  survives.

### Re-verified your central mutation

Deleting `push_field(&mut buf, &path_bytes(&self.root))` after the refactor still
turns 6 tests red, both partitioning guards among them — so the guard did not
weaken.

### Gates

```
cargo test -p trusty-common
  test result: ok. 215 passed; 0 failed; 6 ignored          (exit 0)
  project_index_id: 18 tests, all ok  (was 14; +3 drift pins, +1 hermeticity)

cargo clippy -p trusty-common --all-targets -- -D warnings   exit 0
cargo fmt --check                                            exit 0
bash scripts/check_line_cap.sh   → 7 allowlisted, 0 violations — OK.   exit 0
bash scripts/check_sld.sh        → 2929 code files; 0 errors, 0 warnings   exit 0
```

Still **zero call sites** outside the module and `lib.rs`; single `## [Unreleased]`.

bobmatnyc:
## VERDICT: APPROVE

Round 2, scoped to the delta `2aa0ffb3..92ed250c` (+ CHANGELOG merge checked at `6e6de7e0`). Partitioning-by-construction was cleared in round 1 by my own mutation and was not re-derived. **Both HIGH findings are closed.** Two MEDIUMs surfaced; neither blocks.

---

## Round-1 findings — disposition

### HIGH-1 (env override) — CLOSED, and closed harder than asked

`TRUSTY_GH_USER` is gone, not tested. I verified hermeticity with a **broader** check than the one in the suite — grepping the entire derivation chain, not just the one file:

```
grep -n 'env::var|env::vars|var_os|env!\(|option_env' \
  project_index_id.rs repo_identity.rs github_path.rs index_id.rs slug.rs
→ NONE in the whole chain
```

**On the source-text assertion (`project_index_id_tests.rs:548-561`) — I judge it acceptable, and more robust than it looks.** I probed its substring logic for bypasses:

| Bypass attempt | Caught? | Why |
|---|---|---|
| `std::env::vars()` | ✅ | `env::vars` contains `env::var` |
| `std::env::var_os("X")` | ✅ | contains `env::var` |
| `use std::env::var; var("X")` | ✅ | the `use` line contains `env::var` |
| `option_env!("X")` | ✅ | `option_env!(` contains `env!(` |
| `env!("X")` | ✅ | direct |
| Env read via a helper in **another** module | ❌ | out of the file's scope |
| A dependency reading the env on its behalf | ❌ | out of scope |

The two misses are real but the assertion's stated scope is honest — the message says *"no environment reads in project_index_id.rs"*, not "the derivation is hermetic". `include_str!` is compile-time so it cannot go stale, and the `//`/`*` filter stops the doc prose that names `TRUSTY_GH_USER` from self-triggering. It is a regression tripwire, correctly labelled as one, and I independently confirmed the stronger chain-wide property holds today. The `set_var` rationale is sound: it is process-global and `unsafe` in edition 2024, and a test that set it would corrupt the 17 other tests running in the same binary.

**The residual non-hermeticity disclosure is accurate.** `HOME`/`GIT_CONFIG_GLOBAL` do reach `git config --get user.email`, and `--get` on a multi-valued key does return the LAST entry (I confirmed `Some("two@t.test")` in round 1). It is not an exhaustive list — `XDG_CONFIG_HOME`, `GIT_CONFIG_COUNT`/`KEY`/`VALUE`, and `GIT_CONFIG_NOSYSTEM` also reach it — but the stated mitigation ("the wiring slice must guarantee uniform environment inheritance across daemon and CLI") covers all of them uniformly, so the omission changes nothing actionable. Not a finding.

### HIGH-2 (docs/drift) — CLOSED

The `Guarantees` block (`:191-197`) now guarantees only what is true: non-empty, the charset, and purity over the three fields — then states outright that id stability is only as strong as input stability and that two inputs move. That is the over-statement removed.

The three drift tests **pin shape, not just inequality**, so they cannot pass trivially if derivation broke entirely:

- `id_changes_when_first_commit_lands` asserts `before.origin == None` **and** `matches!(after.origin, Some(ContentHash(_)))` before asserting the ids differ.
- `id_changes_when_origin_remote_is_added` asserts `ContentHash(_)` → `GitHub(_)` **and** `before.starts_with("nocommit-")` / `after.starts_with("acme-widget-")` — label movement, not just digest movement.
- `id_changes_on_orphan_root_commit` asserts `before.origin != after.origin` as an explicit precondition.

All three go red under my git mutation (below), so they are live.

**I agree there is no stabilization that survives scrutiny.** Before `git remote add`, the repo's only durable identifiers are its path and its root commit; a pure function of available state cannot predict the post-remote `owner/repo`, and deferring the id until a remote exists just moves the orphaning window. Folding it into wiring guidance is the right call.

### MEDIUM (vacuous returns) — CLOSED. I re-ran the mutation.

Injected a bogus subcommand into the `git` helper (`project_index_id_tests.rs:41`) so every git invocation fails:

```
resolve_operator_identity_prefers_repo_local_git_identity ... FAILED
symlinked_root_derives_same_id_as_real_root ............... FAILED
id_changes_when_first_commit_lands ........................ FAILED
derivation_is_deterministic_across_calls .................. FAILED
sibling_clones_of_same_repo_derive_distinct_ids ........... FAILED
id_changes_when_origin_remote_is_added .................... FAILED
id_changes_on_orphan_root_commit .......................... FAILED
resolve_operator_identity_ignores_ambient_env_override .... FAILED
linked_worktrees_of_same_repo_derive_distinct_ids ......... FAILED
test result: FAILED. 9 passed; 9 failed;
```

Exactly 9 red / 9 green — every git-dependent test fails loudly, every pure test correctly stays green. Under the round-1 helper these same 9 reported `ok`. **No test can now pass vacuously.**

On brittleness: the mitigation is proportionate. `commit.gpgsign=false` is forced per-repo *after* `init` and *before* any commit (order verified), `user.email`/`user.name` are set per-repo, and `init.defaultBranch` is pinned. Tempdirs are same-owner so `safe.directory` cannot trip. The residual exposure is a global `core.hooksPath` / `init.templateDir` carrying a failing hook — a fair trade against silently-green guards, and CI is 25/25.

### MEDIUM-firmlink, LOW SCHEME_VERSION, LOW naming — all CLOSED

Firmlinks narrowed to symlinks with `/Users/x` vs `/System/Volumes/Data/Users/x` documented as inherent to any path-based key (`:154-159`), and `symlinked_root_derives_same_id_as_real_root` now states the narrow scope. `SCHEME_VERSION` (`:69-80`) reworded to the property that holds, with an explicit "the emitted id carries no version marker" note and an out-of-band-tracking caveat. `gh_user` → `operator`, `resolve_gh_user` → `resolve_operator_identity`, with the `VALUE SHAPE:` note including "Do not compare it against `gh api user`".

---

## New findings

|---|---|---|---|---|
| MEDIUM | crates/trusty-common/src/project_index_id.rs | 112, 200, 221 | `root` is called immutable/"does not move"; a directory rename moves it, and the migration recommendation rests on the false claim. Drift table omits 4 verified actions | Qualify to "under ordinary **git** operations"; add the missing rows |
| MEDIUM | crates/trusty-common/CHANGELOG.md | 25-26 | Claims two callers "cannot derive different ids" — contradicts the module's own residual-non-hermeticity disclosure | Match the source doc's hedge |

### MEDIUM-1 — the drift table is incomplete, and `root` is not immutable

`project_index_id.rs:200` and `:112` both say, unqualified:
> Only `root` is fixed for the life of a content tree.

`:126` gets it right (*"does not move under **ordinary git operations**"*) — the qualifier is present in one place and lost in the other two, including the recommendation at `:220-222`:
> Recommended shape for that slice: reconcile on `root`, which is the only component that does not move…

Executed against the crate at this head, four actions move the id and appear in no table row:

| Action | Observed | In table? |
|---|---|---|
| `mv ~/code/proj ~/code/proj-renamed` | `acme-widget-88e660086615d86b` → `acme-widget-0bd26dfc02d3caa1` | ❌ **and contradicted** |
| GitHub repo **rename** (`remote set-url` GitHub→GitHub) | `acme-widget-6370a380a7ff5200` → `acme-widget-renamed-b1380d83e794e6a8` | ❌ |
| GitHub repo **transfer** (owner change) | `acme-widget-6370a380a7ff5200` → `newowner-widget-4d1e6b23caec733f` | ❌ |
| `git remote remove origin` (GitHub → ContentHash) | `newowner-widget-4d1e6b23caec733f` → `widget-fa2af539d9067e77` | ❌ |

The directory rename is the one that matters: renaming or relocating a project directory is routine, it orphans the index, and it removes the anchor the wiring slice is told to reconcile on. Repo rename/transfer is the second-most likely in practice.

This is documentation completeness, not a code defect — the derivation behaves correctly in every case above, and the `Guarantees` block is no longer false. But the block exists specifically to hand the migration slice a true picture, and right now it hands it an anchor that moves.

**Fix:** restore the `under ordinary git operations` qualifier at `:112` and `:200`; at `:220-222` say root is *the most stable* component and note that a path move is an alias event too; add rows for `remote set-url` (rename/transfer) and `remote remove`. A single test asserting the id moves on a directory rename would pin the important one.

### MEDIUM-2 — the CHANGELOG asserts a guarantee the module explicitly declines to make

`CHANGELOG.md:25-26`:
> Derivation is hermetic: it reads no environment variable, so two callers on one tree (e.g. the launchd daemon and a shell CLI) **cannot derive different ids.**

`project_index_id.rs:323-329` says the opposite, correctly:
> git's own config resolution still depends on `HOME` / `GIT_CONFIG_GLOBAL`, so callers running under materially different environments **can still disagree** when the repo sets no local `user.email`

**Failure scenario:** a launchd-launched daemon with `HOME` unset or pointing elsewhere, against a repo with no repo-local `user.email`. The daemon resolves the global identity from a different `~/.gitconfig` than the shell CLI does, the `operator` component differs, and the two derive different ids for one tree — exactly the divergence the sentence says cannot happen. The daemon-vs-CLI pair the CHANGELOG names as safe is precisely the pair at risk.

The CHANGELOG is what a consumer of this crate reads; it should not be stronger than the source. The same bullet also inherits MEDIUM-1's incomplete drift list.

**Fix:** *"Derivation reads no environment variable directly, so an ops override can no longer split one tree across two ids; note that git's own config resolution still depends on `HOME`/`GIT_CONFIG_GLOBAL`, so the wiring slice must guarantee uniform environment inheritance."*

---

## Verification performed

| Check | Result |
|---|---|
| `cargo test -p trusty-common` | `215 passed; 0 failed; 6 ignored` |
| `cargo test -p trusty-common --lib -- project_index_id` | `18 passed; 0 failed` (was 14) |
| Vacuous-pass mutation (bogus git subcommand) | `9 passed; 9 failed` — all git-dependent tests loud |
| Hermeticity grep, whole derivation chain | zero env reads in 5 modules |
| `cargo fmt --check` / `clippy` | 0 / 0 on this module |
| `check_line_cap.sh` | `7 allowlisted, 0 violations` (440 prod / 561 test) |
| `check_sld.sh` | `0 errors, 0 warnings` — every `Test:` pointer resolves |
| Call sites outside the module + `lib.rs` | **none** |
| CHANGELOG at `6e6de7e0` | single `## [Unreleased]`, single `### Added`, #4258's `json_rmw` bullet **byte-identical to main**, #4207 appended after — conflict resolved correctly |
| Working tree after both mutations | restored, 0 dirty files |

## Notes

- I reviewed branch head `92ed250c` and checked the CHANGELOG on the merge commit `6e6de7e0`; my suite count (215) is against the unmerged base, hence lower than the 223 reported on the merged tree.
- Not flagged: the residual env list omitting `XDG_CONFIG_HOME` / `GIT_CONFIG_COUNT`; the source-text assertion's inability to see cross-module or dependency env reads. Both are honestly scoped and the stated mitigation covers them.
- Both MEDIUMs are documentation, one of them user-facing. Neither warrants another round — they are a qualifier and a hedge, and they can ride along with the wiring slice. The code is correct as it stands.

Issue #3374, opened 2026-07-20
bug(trusty-mpm): framework-guaranteed conventions delivered only via user-editable channels (instruction-delivery drift)


bobmatnyc:
## Invariant

Framework-guaranteed conventions must be delivered ONLY through the
non-overridable channel (`crates/trusty-mpm/src/assets/instructions/BASE_PM.md`,
appended last and unconditionally by `resolve_pm_prompt` — see
`core/instruction_overrides.rs:108-118,145,153` — and asserted to survive every
override combination by the `instruction_pipeline.rs:698-774` test suite).
User-editable surfaces (bundled skills, project-scoped instruction files) may
**restate** a convention as elaboration, but must never be its **sole home** —
a user edit to one of those surfaces would then silently drop the convention
with no signal to anyone.

This invariant was established for the root `CLAUDE.md` by #2299 / PR #2300,
and it has eroded on every surface that wasn't mechanically guarded since.

## Violations found (PR #3271, merged today as `5bc8abab`)

1. **Proportional-documentation policy** and the **`// #1234: <reason>`
   ticket-attribution-at-change-site convention** were added ONLY to:
   - `crates/trusty-mpm/src/assets/skills/documentation-style.md` — a bundled
     skill. Its deployed copy at `<project>/.claude/skills/documentation-style/`
     is user-editable, and the deployer explicitly skips re-writing it once a
     user has modified it: "checksum differs → user-modified → skip"
     (`crates/trusty-agents-common/src/skills/deployer.rs:91-98`).
   - `.trusty-mpm/INSTRUCTIONS.md` — project-scoped and purely additive
     (`core/instruction_overrides.rs` `FILE_INSTRUCTIONS`), so even where
     present it only ever covers one project, never the framework as a whole.

   Neither surface is the non-overridable floor, so **these two conventions
   are currently guaranteed nowhere** — a single skill edit or a missing
   per-project file drops them silently.

2. **The commit/PR attribution footer** convention (`🤖🤖🤖 Generated with
   trusty-mpm — https://github.com/bobmatnyc/trusty-tools`) is duplicated
   across at least 3 non-floor locations with no single source of truth:
   - `CLAUDE_MD_STUB` (`crates/trusty-mpm/src/core/instruction_pipeline.rs:137-141`)
   - the dead `assets/instructions/CLAUDE.md:20-24` (see #3 below)
   - `assets/skills/documentation-style.md` (indirectly, via ticket-attribution
     context)

[...]

PR #4264, opened 2026-07-28
feat(trusty-mpm): author PM instructions as per-section sources (#4183)

bobmatnyc:
Epic [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183) (milestone tm 1.3.0) — **Track A, the CONTENT change**.

> #4249 (the mechanism) merged mid-flight, at 22:32Z. This branch was rebased onto the resulting `main`, so the diff below is the CONTENT change alone.

## What this is, and why the old gate does not apply

#4249 was the MECHANICAL swap and could gate itself on strict byte-equality against the previous delivered prompt. This is the CONTENT change: **the delivered PM prompt changes on purpose**, so that gate is gone.

Two things replace it.

1. **A committed golden snapshot of the fully composed prompt, for both configurations** (`src/core/testdata/pm-prompt-*.md`). Every future edit to a section source now surfaces as a reviewable prose diff in the PR that makes it. Regenerate with `UPDATE_GOLDEN=1 cargo test -p trusty-mpm golden`. **That snapshot diff is the deliverable this PR's acceptance criteria ask for** — see below.
2. **Byte-equality between the two COMPOSERS is retained.** It is no longer a freeze on content, but on the mechanism: whatever the sections say, the packaged path and the legacy override assembly must say it identically.

## The before/after composed-prompt diff

Generated by running the same golden module against `origin/feat/4183-resource-pm-prompt` (before) and this branch (after), in a throwaway worktree. **The content delta is byte-identical across both configurations** — verified mechanically, and it is the property that says authoring sections did not split the product in two.

| configuration | before | after |
|---|---|---|
| bundled fallback (`InstructionPackage`) | 40,126 B | 42,015 B |
| `.trusty-mpm/AGENT_DELEGATION.md` override (legacy assembly) | 35,655 B | 37,544 B |

### 1. Memory + Search lifted out of the middle of Core — MEANING CHANGE

**Removed** from between `## PM Allowlist` and `## Agent Routing`:

```diff
-## Context-First Protocol
-
-Call the MCP tools explicitly only when you need MORE than the injected baseline:
-
-1. `memory_recall` (trusty-memory) for TARGETED or deep recall of prior context
-   the injected block did not surface.
-2. `search` (`mcp__trusty-search__search`) before reading code files or
-   delegating to Research, so investigation starts from indexed results.
-
-Both tools are stable and recommended for targeted lookups on any project.
```

**Added** after the Core body, as two self-contained sections:

```diff
+## Memory Protocol (Context-First)
+…
+Call `memory_recall` (trusty-memory) explicitly only when you need MORE than the
+injected baseline: TARGETED or deep recall of prior context the injected block
+did not surface. Do this BEFORE any research or delegation, never after.
+
+The tool is stable and recommended for targeted lookups on any project.
+
+## Code Search Protocol (Context-First)
+
+Call `search` (`mcp__trusty-search__search`) BEFORE reading code files or
+delegating to Research, so investigation starts from indexed results rather than
+from a cold grep.
+
+The tool is stable and recommended for targeted lookups on any project.
```

**Why this is not cosmetic.** Memory and Search declared tier `project` — a machine-readable claim that a project may replace either independently — while being two halves of one numbered list. Replacing Memory alone deleted item `1.` and left Search opening on a bare `2.`, and the closing sentence spoke for *both* tools while sitting in the Search block. `bundled_pm_package` recorded that as a constraint for #4247 and could not fix it, because the text was a byte range in someone else's file. Authoring the sections is what makes the declared tier honest.

**What changed in meaning, stated plainly:**
- the shared preamble/closer are now stated once per section rather than once for the pair (restatement, not new rule);
- Memory gains *"Do this BEFORE any research or delegation, never after."* — this is **not new doctrine**: it restates the floor's own `### Memory — check BEFORE any research or delegation`. It is stated here because the section must stand alone once it is separately overridable;
- Search gains *"rather than from a cold grep"*, likewise mirroring the floor's `### Code/Architecture Search — use BEFORE grep/find`;
- **position**: both sections now follow the Core body instead of interrupting it.

### 2. `## Trusty Tool Priority (Non-Overridable)` moves — POSITION ONLY

It used to sit *after* `## Framework-Guaranteed Conventions` in `BASE_PM.md`. It is a non-overridable rule, so it now travels with the other non-overridable rules and consequently **precedes** the conventions. Not one word of either block changed; both remain inside the floor, so nothing about what is overridable moved. This also collapses the old four-block floor stream (`Identity, NonOverridable, FrameworkConventions, NonOverridable` — the canonical-order inversion `validate_floor_is_last` was written to permit) into three contiguous sections.

### 3. Sprint-then-Harden doctrine — NEW CONTENT (the owner's in-scope requirement)

Added at the top of the workflow section, governing how hard every gate below it is applied. Preserved faithfully from the owner's wording on [#4185](https://github.com/bobmatnyc/trusty-tools/issues/4185#issuecomment-5099535556) and #4183, including the verbatim quote, the causal claim, the hard line, the close-and-fold rule with its worked example, and the branch/worktree lifetime rule owed from #4207:

```diff
+## Sprint, then Harden (governs how hard every gate below is applied)
+
+Work runs in two phases, not one blended one. Which phase you are in decides how
+much verification ceremony the 5-phase sequence below actually gets.
+
+> "We should sprint to a target (feature complete on a local version), then
+> test/fix carefully. The slow feature release means we have too many things in
+> flight."
+
+1. **SPRINT** — drive to feature-complete on a local version. Testing/CI used
+   judiciously: targeted tests while developing, no CI iteration loops,
+   no critic round on narrow changes.
+2. **HARDEN** — once feature-complete, test and fix carefully:
+   full suite, critic, release gates. Publish only after that.
+
+**The causal claim, which is the point of the doctrine:** slow feature release
+*causes* too many things in flight — it is not a separate problem. Shortening
+time-to-land is the fix; managing WIP count directly (caps, purges) treats the
+symptom.
+
+Derived rules:
+
+- Spend the verification budget where blast radius is real — destructive paths,
+  SemVer/release, security — and cut ceremony everywhere else.
+- **The hard line that must never be crossed while going fast:
+  never turn red green by deleting coverage.** No `#[ignore]`, no cfg-gating a
+  failing test, no `--exclude`, no narrowing to `--lib`. Going fast is a licence
+  to run fewer gates, never to make a failing gate report success.
+- A branch that has drawn **3+ review rounds is evidence to close and fold**, not
+  to attempt round 4. Worked example: #4202 → #4207.
+- Branch = workstream, and it is durable. Worktree = writer, and it is ephemeral
+  and short-lived. Keep worktrees short-lived; keep branches workstream-scoped.
```

**Nothing else in the delivered prompt changed.** No other line is added, removed or reworded in either configuration.

## The split-brain this avoids

Authoring sections while leaving the monolithic assets in place would have forked the product: a project with a `WORKFLOW.md` or `AGENT_DELEGATION.md` override reads the legacy assembly, which read separate files. Instead the monoliths are **deleted**, and `instruction_pipeline::pm_instructions()` / `base_pm()` reconstitute the two multi-section strings the legacy assembly still needs *from the same section sources*, joined with `Join::Blank`'s literal.

Consequences, all asserted:
- a section edit moves both composers together and cannot move one (`pm_instructions_is_its_three_sections`, `base_pm_is_its_three_sections`);
- `composed_package_is_byte_identical_to_the_legacy_bundled_fallback` still passes, now meaning "the two composers agree" rather than "content is frozen";
- the second golden (legacy path) proves an overriding project receives the same Core/Memory/Search/Workflow/floor text as a non-overriding one.

## Composition mechanism: untouched

`InstructionPackage::compose`, `validate`, the schema and the join semantics are **not modified**. The only removals are the runtime-cut machinery that sourcing from files makes dead: `split_asset`, `Cut`, `PM_INSTRUCTIONS_CUTS`, `BASE_PM_CUTS`, `join_for_gap`, and three now-impossible `PackageError` variants (`MarkerNotFound` / `MarkerNotUnique` / `EmptyPiece`) — a marker that could silently relocate cannot exist when the section boundary is the file boundary. `bundled_fallback_package()` is consequently infallible and `compose_bundled_fallback` returns `CompositionError` directly. A mis-authored section is now caught by `validate` (empty file → `EmptyText`) or by `include_str!` (missing file → compile error).

## Existing overrides still resolve

Configurations 2 and 3 remain deliberately on the legacy path and are **not** made schema-valid here (#4247); neither `RosterNotConsumed` nor `SectionWithoutBlocks` is weakened. Every advertised override file keeps working, covered by the pre-existing tests plus a new one:

| file | behaviour | test |
|---|---|---|
| `INSTRUCTIONS.md` | additive | `instructions_appended`, `resolve_pm_prompt_is_byte_identical_with_a_project_addendum` |
| `WORKFLOW.md` | replaces section, forces legacy | `workflow_override_replaces`, `workflow_override_forces_the_legacy_path_even_with_a_roster` |
| `AGENT_DELEGATION.md` | replaces section, forces legacy | `agent_delegation_override_replaces`, `golden_legacy_delegation_override_prompt` |
| `MEMORY.md` | slotted block, forces legacy | `memory_override_is_slotted_after_pm_instructions`, `memory_override_forces_the_legacy_path_even_with_a_roster` |
| `PM_INSTRUCTIONS_DEPLOYED.md` | full body replacement, floor kept | `pm_deployed_replaces_body_but_keeps_base_floor` |

**Migration path, not a break:** no override file changes name, location or semantics. The migration this *enables* — a schema-native section override — is now expressible for Memory and Search for the first time, which is #4247's blocker cleared rather than #4247 done.

## Tier precedence (`fixed` > `project` > `user`), including invalid overrides

Asserted over the **shipped** package, not a fixture:

- `floor_sections_refuse_every_override_tier` — **invalid-override behaviour**: each of the three `fixed` sections refuses both `OverrideTier::Project` and `OverrideTier::User`.
- `content_sections_admit_project_but_not_user_overrides` — the rung that is easy to get backwards: every advertised override file is project-scoped, so a `project` section accepts a project override and **rejects** a user-tier one (a user-tier acceptance is how one operator's machine config leaks into a shared repo).
- `every_advertised_override_file_maps_to_an_overridable_section` — binds SCHEMA to CONTENT by reading the customization table out of the shipped floor text: a file advertised there must map to a section that actually admits a project override.

## `CLAUDE.md` is untouched

`git diff --cached --name-only | grep -i claude.md` → empty. Nothing added writes it; the "Root CLAUDE.md must stay untracked" CI guard is unaffected.

## Verification — SPRINT phase, crate-scoped gates

```
$ cargo test -p trusty-mpm
test result: ok. 3835 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 34.45s
... (38 further test binaries, all ok, 0 failures anywhere)

CLIPPY EXIT=0

$ cargo fmt --check
FMT OK

$ bash scripts/check_instruction_floor.sh
PASS: instruction floor is intact and carries the guaranteed conventions (see #3374).
```

`scripts/check_instruction_floor.sh` is repointed at `sections/framework-guaranteed-conventions.md` (the conventions' real home now), additionally asserts the other two floor sources exist, and newly fails if the monolithic `BASE_PM.md` reappears — two sources for the floor is exactly the drift #3374 removed.

Every `Test:` doc pointer added here resolves to a real `fn`.


bobmatnyc:
## VERDICT: APPROVE

Adversarial review by `code-critic`, in an isolated read-only worktree at `f6a32f38`. I did **not** take the content-delta claim on trust — I composed the delivered prompt myself at the merge-base (`eba04d67`) and at this head, for **four** configurations, and diffed. Zero CRITICAL, zero HIGH.

---

## The content delta, independently reproduced

I installed this PR's `pm_prompt_golden_tests.rs` module into a throwaway worktree at the merge-base, ran `UPDATE_GOLDEN=1`, and diffed the result against the snapshots committed here.

The before-sizes reproduce the PR's table exactly (40,126 B / 35,655 B), so the "before" side is confirmed. The delta is **148 diff lines, and character-for-character identical across both configurations** (only context lines and hunk offsets differ) — the property the PR claims, verified mechanically rather than asserted.

Classifying every hunk:

| hunk | change | class | accounted for by |
|---|---|---|---|
| `@@ -45,22 +45,6` | `## Context-First Protocol` removed from between `## PM Allowlist` and `## Agent Routing` | restatement (moved + reworded) | claim 1 |
| `@@ -469,7 +453,28` | `## Memory Protocol (Context-First)` + `## Code Search Protocol (Context-First)` added after the Core body | restatement + position | claim 1 |
| `@@ -481,7 +486,40` | `## Sprint, then Harden` added at the top of the workflow section | **new doctrine** | claim 3 |
| `@@ -761,25 +799,6` | `## Framework-Guaranteed Conventions` removed from before `## Trusty Tool Priority` | position only | claim 2 |
| `@@ -818,4 +837,23` | same block re-added after `## Trusty Tool Priority` | position only | claim 2 |

**No unaccounted hunk.** The three-item claim holds under my own diff.

The moved conventions block is **byte-identical** — I extracted the removed and added blocks and `diff`'d them directly: no output. "Position only, not one word changed" is literally true.

### Nothing was dropped

The most likely defect in a monolith→sections split. I ran order-insensitive line-set diffs of each old monolith against its new section source(s):

- `AGENT_DELEGATION.md` → `sections/agent-delegation.md`: **byte-identical**.
- `WORKFLOW.md` → `sections/workflow.md`: **additions only** — not one line removed.
- `BASE_PM.md` → `identity.md` + `non-overridable-rules.md` + `framework-guaranteed-conventions.md`: only two blank lines differ (the join boundaries). **No floor content lost.**
- `PM_INSTRUCTIONS.md` → `core.md` + `memory.md` + `search.md`: the only removed lines are the seven Context-First lines, every one of which reappears reworded in the two new sections.

### "Restatement" holds — I checked the prior floor text

The claim that the new imperatives restate rather than add. Quoting the merge-base `BASE_PM.md`:

- line 74: `### Memory — check BEFORE any research or delegation`
- line 80: `### Code/Architecture Search — use BEFORE grep/find`
- line 72: `Always use these BEFORE bash/grep/curl.`

So "Do this BEFORE any research or delegation, never after" is a near-verbatim restatement of the floor's own heading, and "rather than from a cold grep" mirrors line 80. The ordering mandate is **not** new doctrine. Judgement: the label is correct. (One nuance in Notes.)

### Floor integrity

Verified in the strongest configuration available — `PM_INSTRUCTIONS_DEPLOYED.md`, full body replacement. Composed at both revisions: the **only** difference in the entire 6,540-byte output is the position-only move of the conventions block, byte counts identical on both sides. The floor is complete, still appended last, and `Trusty Tool Priority` did not leave the floor's protection — it moved *deeper* into it, from after the conventions to inside the non-overridable rules.

---

## Attacking the new gate

### Is it a real gate?

Yes. I mutated each of the **eight** section files in turn (appended a canary line) and ran the goldens:

| mutated section | `golden_bundled_fallback` | `golden_legacy_delegation_override` |
|---|---|---|
| `identity.md` | **FAILED** | **FAILED** |
| `core.md` | **FAILED** | **FAILED** |
| `memory.md` | **FAILED** | **FAILED** |
| `search.md` | **FAILED** | **FAILED** |
| `workflow.md` | **FAILED** | **FAILED** |
| `agent-delegation.md` | **FAILED** | ok *(correct — that config overrides delegation)* |
| `non-overridable-rules.md` | **FAILED** | **FAILED** |
| `framework-guaranteed-conventions.md` | **FAILED** | **FAILED** |

There is no section edit that changes the delivered prompt and escapes the goldens. Both configurations are covered, and 7 of 8 sections turn **both** red — which is also the direct proof that a section edit moves both composers.

### Attacking `UPDATE_GOLDEN=1`

I ran blind-regeneration attacks (edit a section, regenerate the snapshots without reading the diff, then run the full suite):

- **Delete the entire Sprint-then-Harden doctrine** → `the_delivered_prompt_teaches_sprint_then_harden` **FAILED**. Caught.
- **Delete `## Trusty Tool Priority` from the floor** → **4 tests FAILED** (`floor_carries_the_tool_priority_mandate`, `pm_deployed_replaces_body_but_keeps_base_floor`, `base_pm_is_its_three_sections`, `assemble_system_prompt_contains_all_sections`). Caught hard.
- **Break the attribution footer** → `check_instruction_floor.sh` FAILS. Caught.
- **Gut `memory.md` to a one-line stub** → **entire suite green, floor script PASS.** Not caught. See MEDIUM-2.

So the footgun is real but **bounded** — the floor, the three guaranteed conventions and the sprint doctrine all carry independent content assertions that survive a blind regeneration. `UPDATE_GOLDEN` is not set anywhere in `.github/` or `scripts/`, so the goldens cannot pass vacuously in CI.

### `check_instruction_floor.sh`

Run against three synthetic regressions:

- `BASE_PM.md` reappears → `FAIL: … has reappeared; the floor is authored per-section (#4183).` exit 1
- `identity.md` removed → `FAIL: … every floor section must exist.` exit 1
- attribution needle broken → `FAIL: … missing the commit/PR attribution footer convention.` exit 1

Genuinely guards all three floor sources.

---

## Split-brain fix

- `pm_instructions()` / `base_pm()` reconstitute from the same `include_str!`'d sections — no fourth copy on disk. Confirmed by reading the source, and by the mutation matrix above (an edit to `core.md` reddens the legacy-path golden, which is the fork the PR says it prevents).
- `Join::Blank`'s literal: `composed_package_is_byte_identical_to_the_legacy_bundled_fallback` and `resolve_pm_prompt_is_byte_identical_to_the_legacy_assembly` both pass, and stayed green under all 8 mutations — the two composers agree on whatever the sections say. No off-by-one newline.
- I additionally composed the two configurations **no golden covers** — `WORKFLOW.md` override and `MEMORY.md` override — at both revisions. Deltas are exactly the expected subsets (`MEMORY.md`: full +1,889 B delta; `WORKFLOW.md`: +180 B, i.e. the Memory/Search restructure plus the floor move, correctly without the sprint doctrine, which its override replaces).

## Removals

`PackageError` and all three variants (`MarkerNotFound`, `MarkerNotUnique`, `EmptyPiece`) are gone from `crates/trusty-mpm/src/` entirely, along with `split_asset` / `Cut` / `*_CUTS` / `join_for_gap` — **zero dangling references** repo-wide. No error case is unreachable-but-possible: `InstructionPackage::compose` calls `self.validate()` first (`instruction_package.rs:820`), so an empty section file still errors as `EmptyText`, and a missing one fails at `include_str!`. `instruction_package.rs` is untouched by this PR — the "composition mechanism unmodified" claim is literally true (`git diff` is empty).

## Gates re-verified locally, not trusted

```
cargo test -p trusty-mpm --lib   → test result: ok. 3835 passed; 0 failed; 6 ignored
cargo test -p trusty-mpm (all)   → 6620 passed, 0 failed across 39 binaries
cargo fmt --check                → exit 0
bash scripts/check_line_cap.sh   → line-cap: 7 allowlisted, 0 violations — OK.
bash scripts/check_instruction_floor.sh → PASS
gh pr checks 4264                → 23/23 pass
```

The committed goldens are current: `cargo test golden` passes **without** `UPDATE_GOLDEN`. No `unwrap()`/`expect()` on any reachable production path (all occurrences are inside `mod tests`). CHANGELOG: exactly one `## [Unreleased]`, three new bullets under `### Changed`.

---

## Findings

|---|---|---|---|---|
| MEDIUM | (PR description) | byte table | The "after" byte counts are stale: table says 41,214 B / 36,743 B; the committed goldens — and what the code actually composes, which I verified — are **42,015 B / 37,544 B**, 801 B more in each. The quoted Sprint-then-Harden diff block is also incomplete: the paragraph `Work runs in two phases, not one blended one. Which phase you are in decides how much verification ceremony the 5-phase sequence below actually gets.` is in the shipped section but absent from the quote, under a heading that ends "Nothing else in the delivered prompt changed." | Regenerate the table against `f6a32f38` and re-quote the sprint block in full, or drop the byte counts and point the reader at `src/core/testdata/` |
| MEDIUM | `crates/trusty-mpm/src/core/pm_prompt_golden_tests.rs` | 74–81 | A blind `UPDATE_GOLDEN=1` can silently delete Core/Memory/Search prose. Verified: replacing `memory.md` with `## Memory Protocol (Context-First)\n\nIgnore memory entirely.` and regenerating leaves **3835 passed; 0 failed** and `check_instruction_floor.sh` PASS. The floor, the conventions and the sprint doctrine all have independent content assertions; the Memory/Search mandates in the *body* do not | Add a marker-list assertion for the Memory/Search sections in the style of `the_delivered_prompt_teaches_sprint_then_harden` (e.g. `BEFORE any research or delegation`, `BEFORE reading code files`), so the body's context-first mandate cannot be regenerated away silently |
| LOW | `crates/trusty-mpm/src/core/mod.rs` | 63–65 | The comment `// Epic #4183: committed snapshots of the fully composed PM prompt. The delivered-prompt diff a content change produces is the review artifact.` sits directly above `pub mod ipc;` — it documents `pm_prompt_golden_tests`, declared 12 lines later at 75–77. A reader (and rustdoc adjacency) attributes it to `ipc` | Move lines 63–64 to immediately precede line 75 |
| LOW | `scripts/check_instruction_floor.sh` | 23 | The `What:` block still reads `2. crates/trusty-mpm/src/assets/instructions/BASE_PM.md must contain a recognizable statement of each of the three framework-guaranteed conventions` — that path is now the file the script **fails on** if it exists (line 76). Code repointed, its own contract documentation was not | Update the `What:` block to name `sections/framework-guaranteed-conventions.md` and describe the two new checks (floor siblings, legacy-monolith reappearance) |
| LOW | `crates/trusty-mpm/src/core/pm_prompt_golden_tests.rs` | 77 | `std::env::var_os("UPDATE_GOLDEN").is_some()` treats every value as "on". Verified: `UPDATE_GOLDEN=0 cargo test -p trusty-mpm --lib golden` reports `3 passed` **and rewrites both snapshots** — a falsy-looking value silently disarms the gate and dirties the tree. Not set in CI, so no vacuous CI pass | Gate on the value, e.g. `matches!(std::env::var("UPDATE_GOLDEN").as_deref(), Ok(v) if !v.is_empty() && v != "0")` |

Zero CRITICAL, zero HIGH → **APPROVE**. None of the five blocks the merge; MEDIUM-2 is worth a follow-up given the blast radius of this file.

---

## Notes

- **One nuance on the "restatement" label.** The old Core gated *both* tools behind `Call the MCP tools explicitly only when you need MORE than the injected baseline:`. The new Memory section keeps that gate; the new Search section does not — it is now an unconditional `Call \`search\` … BEFORE reading code files`. The PR describes this as "the shared preamble/closer are now stated once per section", which is precise for the closer and slightly loose for the preamble. I did **not** raise it as a finding, because the *floor* already mandated search unconditionally (`Always use these BEFORE bash/grep/curl`, `use BEFORE grep/find`), so the effect is to make Core agree with the floor rather than contradict it. That is a fix, not a regression — but it is a behavioural shift the three-item list soft-pedals.
- **The sprint doctrine does not reach every project.** Confirmed by composing it: a project shipping `.trusty-mpm/WORKFLOW.md` or `PM_INSTRUCTIONS_DEPLOYED.md` receives **zero** occurrences of `## Sprint, then Harden`, because those overrides replace the section it lives in. This is correct override semantics and not a defect, but it is undisclosed — the owner's requested doctrine is silently absent for exactly the projects that customise their workflow. This repo ships only `.trusty-mpm/INSTRUCTIONS.md` (additive), so it is unaffected. Worth a line in #4247's scope.
- **Pre-existing, not flagged**: `instruction_pipeline.rs:164` carries a `Test:` pointer to `install_writes_assembled_prompt`, which resolves to no `fn` in the crate. Present unchanged at the merge-base — not this PR's regression. Every `Test:` pointer this PR *adds* resolves.
- **Pre-existing, not flagged**: the floor advertises `MEMORY.md` as "Replaces the memory section", but the override is *slotted after* the bundled body rather than replacing it. Now that the bundled block is literally titled `## Memory Protocol`, an overriding project receives both its own rule and the bundled one, bundled-first. Behaviour is unchanged by this PR; the new heading only makes the pre-existing mismatch legible. Also #4247 territory.
- **Test-quality check**: `pm_instructions_is_its_three_sections` / `base_pm_is_its_three_sections` mirror the implementation's format string, so their `assert_eq!` cannot fail on a *content* change — it pins order and join literal only. That is what they claim to do, and the mutation matrix proves the goldens are the content gate, so this is not a vacuous pass. No test added here can pass vacuously in CI.

Issue #2253, opened 2026-07-08
IdentityResolver Tier-3 Jaro-Winkler runs on full email strings, causing cross-identity misattribution once canonical emails are populated


bobmatnyc:
Sibling to #2244, found by the same investigation (a cto-reports migration off config.yaml `team:` onto `aliases_file:`).

## Problem
`IdentityResolver::resolve()` Tier 3 computes Jaro-Winkler similarity on the FULL email string (local-part + "@" + domain), not the local-part. When many identities share a long common domain suffix (e.g. `@duettoresearch.com`), two DIFFERENT local-parts can score ≥ the 0.85 match threshold from the shared suffix alone — especially for short local-parts. Result: one person's commits get attributed to a different, unrelated canonical identity.

Concrete (real) example:
jaro_winkler("ops+snyk@duettoresearch.com", "jenkins@duettoresearch.com") = 0.8567  → ≥ 0.85, so Snyk-bot commits resolve to "Jenkins CI".

## Why it's latent today but about to detonate
Today only ~5 `team.members` entries have a real canonical email (because of #2244, most resolve to canonical_email=""), so the fuzzy "haystack" is tiny and this rarely fires. The moment #2244 is fixed and ~155+ real canonical emails are populated (the whole point of that fix), the haystack is large enough that this fires constantly.

Empirically validated: replaying TGA's resolve() (ported + verified against strsim's jaro_winkler doctest) over all 746 distinct (author_name, author_email) pairs in a real production tga.db, with canonical emails populated, produced **175/746 (23%) cross-identity misattributions** — real people's commits landing on the wrong canonical person (e.g. Yulin Wang→Uma Parakala, Sebastian Coronel→Shane Catts, Jake Becker→Jacob Rosen).

## Coupling with #2244 — MUST be fixed together
Fixing #2244 (empty canonical_email) WITHOUT fixing this Tier-3 bug would ACTIVATE mass misattribution. They should be addressed as a pair.

## Suggested fixes (any one)
1. Tier 3 should compare email LOCAL-PARTS (as Tier 4's normalized fuzzy already does, via `email_local_part`) rather than full-string similarity — or require domain equality before accepting a fuzzy email match (the domain-equality check already exists as `email_domain_matches`, currently only used by the issue-#349 canonical-email policy in `upsert_author`, not by any `resolve()` tier).
2. Expose the fuzzy threshold as config (collector.rs:296 currently calls from_config() with no with_threshold override) and/or raise it.
[...]

Issue #2244, opened 2026-07-08
IdentityResolver: non-aliased team members resolve to empty canonical_email, risking identity collisions


bobmatnyc:
Discovered while migrating a downstream consumer (cto-reports) off config.yaml `team:` and onto the `aliases_file:` mechanism.

## Problem
For team members that have NO `aliases:` sub-list in `config.yaml: team.members`, `Config::resolved_aliases()` (core/config/mod.rs ~1247-1264) builds a name→aliases[] map that DROPS the member's primary `email` field. Downstream, `IdentityResolver::from_alias_map()` (collect/identity/resolver.rs ~129-155) derives `canon_email` as "the first alias-list entry containing @" — which for a non-aliased member is the empty string "".

Because `authors.canonical_email` has a UNIQUE constraint (core/db/sql/0001_initial_schema.sql), the FIRST such member to be written claims canonical_email="", and a SECOND non-aliased member whose git author name matches this path would collide/merge into the same identity row (or hit the unique constraint), silently corrupting per-engineer attribution.

This only manifests when `team.members` is used WITHOUT a `canonical_domain:` (the domain-rescue in upsert_author at resolver.rs ~330-349 never engages) — which is exactly cto-reports' current config.

## Impact
Latent in any TGA consumer using `team.members` entries without per-member `aliases:` and without `canonical_domain:`. Two or more such members can merge into one identity.

## Suggested fix
In the team.members fallback path, seed each member's `primary_email` into their alias list (as the native `aliases_file` AliasFile::to_alias_map() already does, aliases.rs ~108-114), so `canon_email` derives to the real email rather than "". That makes the two config paths consistent.

## Workaround (downstream)
cto-reports is migrating to `aliases_file:` (which includes primary_email correctly), sidestepping this — but the core bug remains for other/未 consumers.

Tracked under #2208.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4291, opened 2026-07-29
fix(tga): gate Tier-3/4 identity fuzzy fallback on alias-table presence (#4251)

### LINKED ISSUE #4251: Tier-3/4 name-fuzzy fallback causes new cross-identity misattributions when alias coverage is high
## Background

Follow-up to #2244 and #2253 (fixed in v2.9.1, PR #2259 — thank you, that fix confirmed working). While re-running a parity verification for a downstream project (cto-reports PR #310, migrating ~130 team-roster entries from a legacy config into a ~177-entry `aliases_file`), the #2253 fix was confirmed present and effective: it correctly gates EMAIL-similarity comparison on domain equality, and the previously-reported email-based misattributions no longer reproduce.

However, re-running the resolver against production `tga.db` (754 distinct real `(author_name, author_email)` pairs) with the larger 177-entry alias map surfaced a **new, previously-undocumented misattribution class** that #2253's fix does not touch: NAME-similarity (Jaro-Winkler) fuzzy fallback in Tier 3/4 is still comparing raw display names/normalized local-parts with no domain or alias-table gating, and produces false positives as the alias/roster haystack grows.

## Reproduction

Examples observed (real commit author identities → incorrectly resolved canonical identity):
- \`Cristian Dominguez\` (cristian.dominguez@...) → resolved to \`Crislaine Tripoli\`
- \`Ravi Chandrasekaran\` → resolved to \`Ravi Pandey\`
- \`Gauri Saykar\` → resolved to \`Gaurav Sharma\`
- \`Josh Taylor\` → resolved to \`Joshua Lepage\` / \`Joshua McCartney\` (inconsistent across runs/entries)
- \`Joseph Ku\` → resolved to \`Joshua Lepage\`

None of these target identities has a declared alias in the alias table explaining the match — they're pure Jaro-Winkler name-similarity hits in Tiers 3/4, none of which are gated by an explicit alias-table entry when a comprehensive \`aliases_file\` is already supplied.

Of 754 total identities compared (old 130-entry map vs new 177-entry map), 77 (10.2%) changed resolution; of those, ~27 are safe display-name-only normalizations and the rest are a mix of genuine corrections (identities #2244/#2253 were meant to fix) and the ~5+ new false positives above.

## Suggested fix

This was actually already suggested as fix option #3 in the original #2253 discussion: **disable Tier 3/4 fuzzy name-similarity fallback entirely when a comprehensive \`aliases_file\` is provided** for a given identity/repo scope. If a project supplies exhaustive alias coverage, fuzzy fallback should not be needed and only adds false-positive risk; it should only kick in as a true fallback when no alias table (or an incomplete one) is present.

## Impact

Blocks cto-reports PR #310 (downstream) from being un-drafted/merged — that PR's own acceptance criteria requires the parity re-run to show zero drift beyond the #2244/#2253 blessed fix, and it doesn't clear that bar due to this issue. Filing here since the root cause is upstream in TGA's resolver, not the downstream project.

## Environment
- \`tga 2.9.3\` (installed via cargo)
- Confirmed at source level against \`crates/trusty-git-analytics/src/collect/identity/resolver.rs\` HEAD (includes commit \`aaad0528\` / PR #2259)

bobmatnyc:
## Problem

`IdentityResolver::resolve` runs four tiers. Tiers 1/2 are exact alias lookups;
Tiers 3/4 are Jaro-Winkler fuzzy guesses. The fuzzy tiers exist to catch
identities a project never declared — but they ran unconditionally, including
when the project had already supplied an exhaustive `aliases_file`.

Once cto-reports moved its ~130-entry roster into a 177-entry `aliases_file`,
every additional roster entry enlarged the fuzzy haystack, and the guesser
started collapsing distinct people onto similarly-spelled colleagues. None of
these targets has a declared alias explaining the match:

| Observed author | Wrongly resolved to |
|---|---|
| `Cristian Dominguez <cristian.dominguez@…>` | `Crislaine Tripoli` |
| `Ravi Chandrasekaran` | `Ravi Pandey` |
| `Gauri Saykar` | `Gaurav Sharma` |
| `Josh Taylor` | `Joshua Lepage` |
| `Joseph Ku` | `Joshua Lepage` |

## Design decision

**Presence of an `aliases_file` disables Tiers 3/4 by default, with an explicit
opt-in flag to re-enable.** This is the owner's recommendation and option #3
from the original #2253 discussion. Concretely, a new top-level config key:

| `fuzzy_identity_fallback` | Behaviour |
|---|---|
| unset (**default**) | Tiers 3/4 run unless a declared `aliases_file` **successfully loaded a non-empty alias table** |
| `true` | force Tiers 3/4 on even with an alias file (pre-2.11.0 behaviour) |
| `false` | force Tiers 3/4 off even without an alias file (exhaustive inline `developer_aliases`) |

Explicitly **not** implemented: a "comprehensiveness score" / coverage
threshold. A heuristic that guesses whether the alias table is complete enough
to stop guessing is a second guessing layer stacked on the first — it would
turn a reproducible bug into a threshold-tuning problem. Presence of the file
is a fact the operator states; coverage is not.

Rationale for the direction of the failure mode: an undeclared author showing
up under its own raw name is a **visible, fixable** gap — add an alias. A
silently misattributed author corrupts every per-developer metric downstream
with no signal that it happened.

Scope preserved exactly as required:
- **Tier 1/2** exact alias / canonical-email / login-handle resolution: untouched,
  and re-asserted by `aliases_file_gate_preserves_tier12_declared_resolutions`.
- **#2253 email-domain gating**: untouched. `email_domain()`, the
  domain-equality gate and the local-part comparison are byte-identical; the
  new flag short-circuits *before* Tier 3 rather than modifying it. Re-asserted
  by `issue_2253_domain_gate_intact_when_fuzzy_enabled` (both directions: the
  Snyk/Jenkins false positive stays suppressed, and the genuine
  `acoopr` → `Alice Cooper` same-domain match still resolves).

## Before / after test observation

`aliases_file_disables_tier34_name_fuzzy` builds a `config.yaml` + external
`aliases_file` on disk, loads it through the real `Config::load` path, and
resolves the five reported pairs. Each pair carries a precondition assertion
that the fuzzy score genuinely clears its threshold, so the test cannot pass
vacuously.

**BEFORE the fix** — production code reverted to `origin/main`, `cargo clean -p tga`
first to rule out any stale artifact:

```
test collect::identity::resolver::tests::aliases_file_disables_tier34_name_fuzzy ... FAILED

assertion `left == right` failed: #4251: undeclared authors must pass through, not be guessed onto the nearest-spelled roster member
  left: ["Cristian Dominguez <cristian.dominguez@duettoresearch.com> => Crislaine Tripoli <crislaine.tripoli@duettoresearch.com>",
         "Ravi Chandrasekaran <ravi.chandrasekaran@duettoresearch.com> => Ravi Pandey <ravi.pandey@duettoresearch.com>",
         "Gauri Saykar <gauri.saykar@duettoresearch.com> => Gaurav Sharma <gaurav.sharma@duettoresearch.com>",
         "Josh Taylor <josh@duettoresearch.com> => Joshua Lepage <joshua.lepage@duettoresearch.com>",
         "Joseph Ku <joseph.ku@duettoresearch.com> => Joshua Lepage <joshua.lepage@duettoresearch.com>"]
 right: [... each pair resolving to itself ...]

test result: FAILED. 24 passed; 1 failed; 0 ignored; 0 measured; 742 filtered out
```

All five reported misattributions reproduce exactly. Note that
`aliases_file_gate_preserves_tier12_declared_resolutions` and
`issue_2253_domain_gate_intact_when_fuzzy_enabled` **pass** in this pre-fix run
— the test targets the defect, it does not fail everything.

**AFTER the fix** — `cargo test -p tga` (full suite, **no** `--lib`):

```
test collect::identity::resolver::tests::aliases_file_disables_tier34_name_fuzzy ... ok
test collect::identity::resolver::tests::aliases_file_gate_preserves_tier12_declared_resolutions ... ok
test collect::identity::resolver::tests::explicit_opt_in_reenables_fuzzy_with_aliases_file ... ok
test collect::identity::resolver::tests::explicit_opt_out_disables_fuzzy_without_aliases_file ... ok
test collect::identity::resolver::tests::no_aliases_file_keeps_fuzzy_enabled_by_default ... ok
test collect::identity::resolver::tests::with_fuzzy_fallback_false_suppresses_tier34 ... ok
test collect::identity::resolver::tests::issue_2253_domain_gate_intact_when_fuzzy_enabled ... ok
test collect::identity::resolver::tests::tier3_does_not_merge_bots_sharing_domain_suffix ... ok
test collect::identity::resolver::tests::tier3_still_matches_same_domain_near_identical_local_parts ... ok

     Running unittests src/lib.rs
test result: ok. 770 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 6.88s
     Running unittests src/main.rs
test result: ok. 151 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.30s
     Running tests/bitbucket_live.rs
     Running tests/config_mount.rs
     Running tests/integration_test.rs
   Doc-tests tga
test result: ok. 5 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
```

The single ignored lib test is the pre-existing
`core::effort::tests::effort_cross_validate`, not introduced here. No new test
uses `#[ignore]`, an env-var gate, an early `return`, or a skip-when-missing
guard; fixtures use `.unwrap()` so a load failure fails the test.

`explicit_opt_in_reenables_fuzzy_with_aliases_file` asserts that with
`fuzzy_identity_fallback: true` the `Gauri Saykar` → `Gaurav Sharma`
misattribution comes **back** — proof the flag is load-bearing and the primary
test is not passing for an unrelated reason.

## Measured behaviour change, and effect on the 77 changed resolutions

Behaviour changes **only** for configs that set `aliases_file` (or explicitly
set the new key). Every other deployment — inline `developer_aliases`,
`team.members`, `trusty-review`'s `from_alias_map` profile selector, the
benches — keeps the previous behaviour bit-for-bit, guarded by
`no_aliases_file_keeps_fuzzy_enabled_by_default`.

Within an `aliases_file` config, the change is strictly a *narrowing*: Tiers
1/2 produce identical output, and only pairs that would have fallen through to
a fuzzy guess now pass through instead.

**None of the genuine corrections in the 77 changed resolutions are undone.**
Those corrections came from the 130 → 177 alias-map expansion, i.e. from newly
*declared* entries resolving at Tiers 1/2 — the tier this change does not
touch. The ~27 display-name-only normalizations likewise ride on declared
canonical names. Concretely re-asserted in
`aliases_file_gate_preserves_tier12_declared_resolutions`: the `-c` contractor
email variant, a non-email login handle, a GitHub `users.noreply` alias, and a
case-folded canonical email all still resolve to their canonical pair with the
gate active. The #2244/#2253 improvements are email-*domain* gating in Tier 3,
which is only reachable when fuzzy is enabled and is verified intact there.

## Note: the defect is broader than "name similarity"

Worth flagging for the record — measured, not assumed. The issue attributes the
false positives to name-similarity, but scoring the reported pairs shows the
`#2253`-compliant **email local-part** path is equally implicated whenever the
roster uses `firstname.lastname@domain` addresses and the author is on the same
domain (the domain gate is open, and the local-part comparison is then
isomorphic to the name comparison):

| Pair | Tier 3 name | Tier 3 email (domain-gated) | Tier 4 normalized |
|---|---|---|---|
| Cristian Dominguez → Crislaine Tripoli | 0.8412 | 0.8412 | 0.8412 **hit** |
| Ravi Chandrasekaran → Ravi Pandey | 0.8584 **hit** | 0.8584 **hit** | 0.8584 **hit** |
| Gauri Saykar → Gaurav Sharma | 0.8662 **hit** | 0.8662 **hit** | 0.8662 **hit** |
| Josh Taylor `<josh@…>` → Joshua Lepage | 0.6773 | 0.8615 **hit** | 0.8615 **hit** |
| Joseph Ku → Joshua Lepage | 0.8260 | 0.8260 | 0.8260 **hit** |

(thresholds: Tier 3 = 0.85, Tier 4 = 0.82)

A fix that gated *only* the name-similarity term would have left
`Ravi Chandrasekaran` and `Gauri Saykar` broken via the email path, and
`Josh Taylor` broken outright. Gating the whole fuzzy fallback — which is what
the issue's suggested fix literally says — is what actually clears all five.

Two further corrections to the issue's description:

1. **`Josh Taylor` does not reproduce from the display name.**
   `jaro_winkler("josh taylor", "joshua lepage")` = 0.6773, well under
   threshold. It reproduces from the *email local-part* `josh`, which scores
   0.8615 against `Joshua Lepage` and 0.8500 against `Joshua McCartney` — both
   over threshold. The fixture uses `Josh Taylor <josh@duettoresearch.com>`
   accordingly.

2. **The "inconsistent across runs" behaviour has a concrete cause.**
   `from_alias_map` iterates a `HashMap`, so `members` ordering is
   nondeterministic per process, and both fuzzy loops tie-break with
   `score > best` (strictly greater) — so on an exact score tie the
   first-encountered member wins, and which member that is varies run to run.
   With `josh` the two Joshuas score differently (0.8615 vs 0.8500) so Lepage
   wins deterministically, but any exactly-tied pair in the full 177-entry
   roster flips between runs. This PR removes the symptom for `aliases_file`
   configs; the underlying nondeterministic tie-break still exists on the
   opt-in path and in `from_alias_map` callers, and is worth a follow-up if
   anyone re-enables fuzzy deliberately.

## Round 2 — changes after the BLOCK on #4291

All four findings addressed. Evidence for each below; nothing here is asserted
without a command behind it.

### 1. CRITICAL — the racing test helper (fixed, and the race measured both ways)

`resolver_from_aliases_file` now uses `tempfile::TempDir` (kernel-unique name,
RAII cleanup that also runs on panic). The three manual `remove_dir_all` calls
are gone. I applied the same fix to the **pre-existing**
`duetto_contractors_config_resolves`, which carried the identical broken
`pid + nanos` scheme.

I reproduced the race before fixing it, at the exact committed state
(`ded54b76`), 100 runs, filter `collect::identity::resolver::`:

```
=== PRE-FIX HELPER, 100 RUNS ===
OK:   88
FAIL: 12
```

Both of the critic's captured modes reproduced, with counts:

```
   6  left: "whoever"  right: "Akash Arora"      <- zero-member resolver (:619)
   3  supplying an aliases_file must disable the fuzzy fallback  (:568)
   2  Config::load unwrap panic (:554) / create_dir unwrap panic (:542)
```

12/100 against the critic's 10/100 — same defect, same rate.

**And I confirmed the dangerous direction directly**, rather than inferring it.
A throwaway diagnostic on the pre-fix code, pointing `aliases_file` at a missing
file to simulate exactly what the race produces:

```
resolved_aliases() empty? true
fuzzy_fallback()        = false
--- primary test's assertions against this resolver ---
assert!(!r.fuzzy_fallback())  -> PASSED
  Cristian Dominguez     => Cristian Dominguez (wrong_target=Crislaine Tripoli)
  Ravi Chandrasekaran    => Ravi Chandrasekaran (wrong_target=Ravi Pandey)
  Gauri Saykar           => Gauri Saykar (wrong_target=Gaurav Sharma)
  Josh Taylor            => Josh Taylor (wrong_target=Joshua Lepage)
  Joseph Ku              => Joseph Ku (wrong_target=Joshua Lepage)
ALL PRIMARY ASSERTIONS PASSED against a ZERO-MEMBER resolver.
```

The critic's read was exactly right: a zero-member resolver satisfied **every**
assertion of the primary #4251 test, `!fuzzy_fallback()` included. That
diagnostic was removed; the permanent guards are:

- `assert_roster_loaded()` — a positive control asserting a known roster member
  resolves, called **first** in all three `aliases_file` tests, before anything
  that asserts pass-through.
- `non_vacuity_control_rejects_zero_member_resolver` — a `#[should_panic]` test
  pinning that the control actually rejects a zero-member resolver. A guard
  nobody proves is load-bearing is not a guard.

**Post-fix: 100 consecutive runs, 0 failures**, each run required to execute
exactly 33 tests (a run executing 0 tests is scored a FAILURE by the harness,
not a pass):

```
=== FINAL CODE, 100 RUNS, filter 'collect::identity::resolver::', 33 tests required each run ===
OK:   100
FAIL: 0
run 100 OK (33 passed)
```

No leaked temp dirs: I checked `$TMPDIR`, `/tmp` and `/private/tmp` for
`tga-issue-4251-*` and `tga-duetto-contractors-*` and found 0 in this
environment (the 56 were from the reviewer's own stress runs). `TempDir` makes
recurrence impossible regardless, including on panic.

### 2. HIGH — `Closes #4251` → `Refs #4251`

Done, in both the PR body and the commit message (the branch was rewritten to a
single commit so no `Closes #4251` survives anywhere in history and a squash
merge cannot pick it up). The critic's reasoning is correct and I withdraw my
round-1 claim about the tie-break: I asserted the `score > best` tie-break could
flip `Josh Taylor` between the two Joshuas, but the two scores are **distinct**
(0.8615 vs 0.8500), and a keep-the-max scan is order-independent for distinct
scores. My claim was wrong. The live nondeterminism is the last-writer-wins
`HashMap` alias insert one tier up, which this PR does not touch. #4293 carries it.

### 3. HIGH — gate on the load, not the declaration

Fixed, and I deviated from the suggested one-liner deliberately. The suggested
form:

```rust
.unwrap_or(map.is_empty() && config.aliases_file.is_none())
```

does not do what the finding asks. With a broken alias file it yields
`true && false` = `false` — fuzzy still **off**, i.e. the reported bug is
unchanged. And with a non-empty inline `developer_aliases` map and no
`aliases_file` it yields `false && true` = `false`, silently flipping every
inline-alias deployment to fuzzy-off — well outside #4251's scope.

Implemented instead:

```rust
let alias_table_loaded = match &config.aliases_file {
    None => false,
    Some(_) => match config.resolved_alias_map(config.config_dir()) {
        Ok(m) if !m.is_empty() => true,
        Ok(_)  => { error!(...); false }   // declared but empty
        Err(e) => { error!(error = %e, ...); false }   // declared but unloadable
    },
};
resolver.fuzzy_fallback = config.fuzzy_identity_fallback.unwrap_or(!alias_table_loaded);
```

A declared-but-broken alias file now degrades to the documented pre-#4251
behaviour and logs at `error!`, instead of to silent total fragmentation. Three
new tests pin all three branches, including one that fails under the suggested
one-liner:

- `broken_aliases_file_keeps_fuzzy_enabled` — asserts `resolved_alias_map()` is
  `Err` **and** `resolved_aliases()` is empty (pinning that the swallow is real),
  then that fuzzy stays on.
- `empty_aliases_file_keeps_fuzzy_enabled`
- `inline_developer_aliases_do_not_disable_fuzzy`

### 4. MEDIUM — version and visibility

- `crates/trusty-git-analytics/Cargo.toml` bumped **2.10.0 → 2.11.0**; docs and
  changelog now say v2.11.0. `Cargo.lock` moved by exactly one line (the `tga`
  version) — no dependency refresh.
- The behaviour flip is announced at `info!` with the remediation inline
  (`Set 'fuzzy_identity_fallback: true' to restore guessing`), not `debug!`.
- LOW: the `with_fuzzy_fallback` doc no longer claims `trusty-review`'s selector
  calls it. It now states plainly that no caller sets it today and why it exists.

### 5. My round-1 "stale build" anomaly — retracted, root-caused

The critic was right to reject it. The **temp-dir race** is what I hit, in the
zero-member vacuous-pass mode demonstrated above. The mechanism, confirmed by
reading the code: `Config::load` does **not** read the alias file — resolution
is lazy, in `resolved_aliases()` called later by `from_config`. So the victim
test loads `config.yaml` successfully, the colliding test's cleanup then deletes
the directory, and the lazy alias load fails into the swallowed-error path. That
explains both halves of my anomalous run: the primary test passed vacuously
while the Tier-1/2 test, which had already loaded, passed normally.

The `#[path]` trap the critic raised is **real and I confirmed it**, but it is
not what I hit:

```
filter 'collect::identity::resolver_tests' -> test result: ok. 0 passed; 0 failed; 774 filtered out
filter 'collect::identity::resolver::'     -> test result: ok. 33 passed; 0 failed; 742 filtered out
```

One character of path drift does turn red into "ok". My captured logs from that
run show `running 25 tests` with the target test printing `... ok`, so zero-tests
was not the mechanism. I have made it unable to hide either way: the stress
harness scores any run not executing the expected test count as a FAILURE, and
every count is reported above.

### Red-before / green-after, re-run after the tempfile change

Exact filter: **`cargo test -p tga --lib 'collect::identity::resolver::'`** —
33 tests, count asserted.

BEFORE (gate mutated out of `resolve()`, `cargo clean -p tga` first, everything
else including the tempfile helper and the non-vacuity control left in place):

```
test ...aliases_file_disables_tier34_name_fuzzy ... FAILED
  left: ["Cristian Dominguez <…> => Crislaine Tripoli <…>",
         "Ravi Chandrasekaran <…> => Ravi Pandey <…>",
         "Gauri Saykar <…> => Gaurav Sharma <…>",
         "Josh Taylor <josh@…> => Joshua Lepage <…>",
         "Joseph Ku <…> => Joshua Lepage <…>"]
test ...explicit_opt_out_disables_fuzzy_without_aliases_file ... FAILED
test ...with_fuzzy_fallback_false_suppresses_tier34 ... FAILED

test result: FAILED. 30 passed; 3 failed; 0 ignored; 742 filtered out
```

All five misattributions reproduce **with the non-vacuity control passing**, so
this red is not the zero-member artifact.

AFTER (gate restored):

```
test result: ok. 33 passed; 0 failed; 0 ignored; 742 filtered out
```

## Verification

- `cargo test -p tga` (full suite, no `--lib`): 774 passed / 0 failed / 1 ignored
  (lib), 151/0/0 (bin), 1+2+1 integration, 5 passed / 2 ignored doc-tests. The
  ignored lib test is the pre-existing `core::effort::tests::effort_cross_validate`.
- 100 consecutive runs of `collect::identity::resolver::` — 100 OK, 0 FAIL,
  33 tests executed each run.
- `cargo clippy -p tga --all-targets -- -D warnings` — exit 0.
- `cargo fmt -p tga --check` — exit 0.
- `cargo check -p trusty-review --all-targets` — exit 0 (only downstream
  consumer of `tga`; unaffected, it uses `from_alias_map`).
- `./scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK`.

`crates/trusty-git-analytics/CHANGELOG.md` updated under `## [Unreleased]`.
`docs/trusty-git-analytics/requirements/configuration.md` documents the four
tiers and the new key.

Added: 423 / Removed: 0 / Net: +423 (of which 293 are tests, 40 docs,
33 changelog; ~57 production lines, mostly doc comments).

Refs #4251 — the run-to-run inconsistency #4251 also reports is a Tier-1/2
defect this PR does not touch; it is tracked on #4293. Do **not** auto-close
#4251 on merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10302,7 +10302,7 @@ dependencies = [
 
 [[package]]
 name = "tga"
-version = "2.10.0"
+version = "2.11.0"
 dependencies = [
  "aho-corasick",
  "anyhow",
--- a/crates/trusty-git-analytics/CHANGELOG.md
+++ b/crates/trusty-git-analytics/CHANGELOG.md
@@ -7,6 +7,48 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+MINOR, not patch. `Config` gained a public field and its fields are all public
+with no `#[non_exhaustive]`, and the new default flips identity-resolution
+behaviour for every existing `aliases_file` deployment on upgrade.
+
+### Fixed
+
+- Identity resolution no longer runs the Tier-3/4 Jaro-Winkler fuzzy fallback
+  when a comprehensive `aliases_file` is configured (#4251). As the alias
+  roster grew from 130 to 177 entries, the fuzzy tiers began collapsing
+  distinct people onto similarly-spelled colleagues — `Cristian Dominguez` →
+  `Crislaine Tripoli`, `Ravi Chandrasekaran` → `Ravi Pandey`, `Gauri Saykar` →
+  `Gaurav Sharma`, `Josh Taylor` and `Joseph Ku` → `Joshua Lepage` — none of
+  which had a declared alias explaining the match. An author that matches no
+  declared alias is now reported under its own raw name instead of being
+  guessed. Tier-1/2 exact alias resolution and the #2253 email-domain gate are
+  unchanged.
+
+### Added
+
+- `fuzzy_identity_fallback` config key (#4251). Unset (default) disables the
+  Tier-3/4 fuzzy fallback only when a declared `aliases_file` successfully
+  loaded a non-empty alias table; `true` forces it on, `false` forces it off.
+  A declared `aliases_file` that fails to load, or resolves empty, leaves the
+  fallback ON and logs at `error!` rather than silently fragmenting every
+  identity. `IdentityResolver` gained the matching `with_fuzzy_fallback()`
+  builder and `fuzzy_fallback()` accessor for callers that construct a
+  resolver outside `from_config`.
+- The behaviour flip is announced once at `info!` when it takes effect.
+
+### Internal
+
+- Identity resolver tests now use `tempfile::TempDir` instead of a hand-rolled
+  `process::id() + SystemTime` directory name. The old scheme was not unique
+  within a test binary — `process::id()` is constant across parallel tests and
+  the `SystemTime` remainder is coarser than a nanosecond — so tests collided
+  and each one's cleanup deleted a directory another was still using. Because
+  `Config::resolved_aliases()` swallows alias-file load errors, the victim got
+  a resolver with zero members, which returns every input unchanged and so
+  satisfied the #4251 pass-through assertions *vacuously*. Measured at 10
+  failures per 100 runs. The affected tests now also carry an explicit
+  non-vacuity assertion that the roster loaded.
+
 ---
 ## [2.10.0] — 2026-07-27
 
--- a/crates/trusty-git-analytics/Cargo.toml
+++ b/crates/trusty-git-analytics/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tga"
-version = "2.10.0"
+version = "2.11.0"
 publish = true
 edition = "2021"
 # Aligned with the workspace MSRV (1.91) per #254. Required for
--- a/crates/trusty-git-analytics/src/collect/identity/resolver.rs
+++ b/crates/trusty-git-analytics/src/collect/identity/resolver.rs
@@ -6,12 +6,17 @@
 //! 2. Fuzzy match against team member canonical emails/names using
 //!    Jaro-Winkler similarity above a configurable threshold.
 //! 3. Fall through and return the raw pair unchanged.
+//!
+//! The fuzzy tiers are a *fallback for undeclared identities*, not a
+//! supplement to a declared roster: when the project supplies a comprehensive
+//! alias table they are switched off (issue #4251). See
+//! [`IdentityResolver::fuzzy_fallback`].
 
 use std::collections::HashMap;
 
 use rusqlite::params;
 use strsim::jaro_winkler;
-use tracing::debug;
+use tracing::{debug, error, info};
 
 use crate::core::config::TeamConfig;
 use crate::core::db::Database;
@@ -99,6 +104,13 @@ pub struct IdentityResolver {
     /// the stored canonical email. See [`Self::upsert_author`] for the
     /// selection policy.
     canonical_domain: Option<String>,
+    /// Whether the Tier-3/4 Jaro-Winkler fuzzy fallback may run (issue #4251).
+    ///
+    /// `true` preserves the historical behaviour. `false` stops resolution
+    /// after the exact Tier-1/2 alias lookups, so an identity that is not
+    /// declared in the alias table passes through unchanged instead of being
+    /// guessed onto the nearest-spelled roster member.
+    fuzzy_fallback: bool,
 }
 
 impl IdentityResolver {
@@ -130,6 +142,7 @@ impl IdentityResolver {
             members,
             threshold: DEFAULT_SIMILARITY_THRESHOLD,
             canonical_domain,
+            fuzzy_fallback: true,
         }
     }
 
@@ -166,12 +179,19 @@ impl IdentityResolver {
             members,
             threshold: DEFAULT_SIMILARITY_THRESHOLD,
             canonical_domain: None,
+            fuzzy_fallback: true,
         }
     }
 
     /// Construct a resolver from a [`crate::core::config::Config`], preferring
     /// the Python-compatible `developer_aliases` map when present, falling
     /// back to `team.members`.
+    ///
+    /// Also decides whether the Tier-3/4 fuzzy fallback runs (issue #4251):
+    /// an explicit `fuzzy_identity_fallback` in the config wins; otherwise the
+    /// fallback is disabled only when a declared `aliases_file` actually
+    /// resolved to a non-empty alias table. A declared-but-unloadable file
+    /// leaves the fallback ON and logs at `error!`.
     pub fn from_config(config: &crate::core::config::Config) -> Self {
         let map = config.resolved_aliases();
         let mut resolver = if !map.is_empty() {
@@ -191,6 +211,47 @@ impl IdentityResolver {
                     .filter(|d| !d.is_empty());
             }
         }
+        // Gate on whether the alias table actually LOADED, not merely on whether
+        // one was declared. `Config::resolved_aliases` deliberately swallows
+        // alias-file load errors, so keying on `aliases_file.is_some()` alone
+        // would let a typo'd or unreadable file produce the worst possible
+        // state: an empty alias table AND a disabled fallback, fragmenting every
+        // author with nothing but a debug line to show for it.
+        let alias_table_loaded = match &config.aliases_file {
+            None => false,
+            Some(_) => match config.resolved_alias_map(config.config_dir()) {
+                Ok(m) if !m.is_empty() => true,
+                Ok(_) => {
+                    error!(
+                        "aliases_file is configured but resolved to an EMPTY alias table; \
+                         keeping the Tier-3/4 fuzzy fallback enabled"
+                    );
+                    false
+                }
+                Err(e) => {
+                    error!(
+                        error = %e,
+                        "aliases_file is configured but could not be loaded; identity \
+                         resolution will fall back to fuzzy matching — fix the config"
+                    );
+                    false
+                }
+            },
+        };
+        resolver.fuzzy_fallback = config
+            .fuzzy_identity_fallback
+            .unwrap_or(!alias_table_loaded);
+        if !resolver.fuzzy_fallback {
+            // Announce at info!, not debug!: this flips resolution behaviour for
+            // every existing `aliases_file` deployment on upgrade, and an
+            // operator comparing two runs needs to see why the numbers moved.
+            info!(
+                members = resolver.members.len(),
+                "Tier-3/4 fuzzy identity fallback DISABLED (alias table supplied); \
+                 authors with no declared alias will be reported under their raw \
+                 name. Set `fuzzy_identity_fallback: true` to restore guessing."
+            );
+        }
         resolver
     }
 
@@ -198,6 +259,28 @@ impl IdentityResolver {
     pub fn with_threshold(mut self, threshold: f64) -> Self {
         self.threshold = threshold;
         self
+    }
+
+    /// Enable or disable the Tier-3/4 Jaro-Winkler fuzzy fallback (issue #4251).
+    ///
+    /// Why: [`Self::from_config`] derives the flag from config, but resolvers
+    /// built from a bare alias map ([`Self::from_alias_map`]) or a
+    /// [`crate::core::config::TeamConfig`] have no config to read. Those
+    /// callers — currently `trusty-review`'s profile selector and the benches —
+    /// keep the historical fuzzy-on default and can opt out through this
+    /// builder if they ever need to. No caller sets it today; it exists so the
+    /// switch is reachable without routing through `Config`.
+    /// What: sets the flag consulted by [`Self::resolve`] after the exact
+    /// Tier-1/2 alias lookups.
+    /// Test: see `resolver_tests::with_fuzzy_fallback_false_suppresses_tier34`.
+    pub fn with_fuzzy_fallback(mut self, enabled: bool) -> Self {
+        self.fuzzy_fallback = enabled;
+        self
+    }
+
+    /// Report whether the Tier-3/4 fuzzy fallback is active on this resolver.
+    pub fn fuzzy_fallback(&self) -> bool {
+        self.fuzzy_fallback
     }
 
     /// Register an alias → canonical-name mapping after construction.
@@ -257,6 +340,17 @@ impl IdentityResolver {
             if let Some((cn, ce)) = self.find_member_by_name(canon_name) {
                 return (cn, ce);
             }
+        }
+
+        // Issue #4251: Tiers 3 and 4 guess. When the project has declared a
+        // comprehensive alias table, an inbound pair that reached this point is
+        // by definition NOT on the roster, and every additional roster entry
+        // only widens the set of near-spellings it can be captured by
+        // (`Cristian Dominguez` → `Crislaine Tripoli`). Stop here and let the
+        // identity pass through unresolved — a visibly-unmapped author is
+        // recoverable by adding an alias; a silently-misattributed one is not.
+        if !self.fuzzy_fallback {
+            return (name.to_string(), email.to_string());
         }
 
         // 3. Fuzzy match against member names/emails (Jaro-Winkler).
--- a/crates/trusty-git-analytics/src/collect/identity/resolver_tests.rs
+++ b/crates/trusty-git-analytics/src/collect/identity/resolver_tests.rs
@@ -199,21 +199,17 @@ fn multiple_emails_same_person() {
 /// `configs/duetto-contractors.yaml` setup. This guards against
 /// regressions in YAML schema, path resolution, or resolver wiring.
 ///
-/// The fixture is materialized into a temp dir so the test is
+/// The fixture is materialized into a `tempfile::TempDir` so the test is
 /// hermetic: it does not depend on absolute paths outside the repo
-/// (which would fail in CI).
+/// (which would fail in CI). Previously this built its own `pid + nanos`
+/// directory name, the same scheme that raced in the #4251 helper below —
+/// `process::id()` is constant within a test binary and the `SystemTime`
+/// remainder is coarser than a nanosecond, so the name is not actually
+/// unique. `TempDir` also cleans up on panic, which the manual path did not.
 #[test]
 fn duetto_contractors_config_resolves() {
-    let unique = format!(
-        "tga-duetto-contractors-{}-{}",
-        std::process::id(),
-        std::time::SystemTime::now()
-            .duration_since(std::time::UNIX_EPOCH)
-            .map(|d| d.as_nanos())
-            .unwrap_or(0)
-    );
-    let tmp = std::env::temp_dir().join(unique);
-    std::fs::create_dir_all(&tmp).expect("create tmp");
+    let tmpdir = tempfile::TempDir::new().expect("create tmp");
+    let tmp = tmpdir.path();
 
     // Aliases file with a subset of the real Duetto contractor map,
     // including the cases the assertions below exercise (case-folding,
@@ -260,8 +256,6 @@ developers:
     // Non-email login handle alias.
     let (n, _) = r.resolve("jangareddy-duetto", "noise@nowhere.test");
     assert_eq!(n, "Janga Vinod Kumar Reddy");
-
-    let _ = std::fs::remove_dir_all(&tmp);
 }
 
 #[test]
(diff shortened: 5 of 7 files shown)


bobmatnyc:
## Verdict: BLOCK

Adversarial review, independent of the implementer. Base `origin/main`, head `ded54b76`. All line numbers verified by reading the files; all claims below are backed by commands I ran.

The production change is **correct and well-argued** — I proved the gate is load-bearing by mutation. The block is on the **test evidence**, which I measured flaking at **10 runs in 100** on this branch, in a mode that can make the primary #4251 test pass *vacuously*.

---

## Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-git-analytics/src/collect/identity/resolver_tests.rs` | 532-556 | `resolver_from_aliases_file` temp-dir collision: 10/100 flake; can green the primary #4251 test vacuously | Use `tempfile::TempDir` (already a dep) + non-vacuity assert |
| HIGH | PR body | 205 | `Closes #4251` overclaims — the run-to-run inconsistency #4251 reports is a Tier-1/2 defect this PR does not touch | `Refs #4251`, track on #4293 |
| HIGH | `crates/trusty-git-analytics/src/collect/identity/resolver.rs` | 212-214 | Gate keys on the alias-file *declaration*, not on whether it *loaded* | Gate on the resolved map; `warn!` on load failure |
| MEDIUM | `docs/trusty-git-analytics/requirements/configuration.md` | 222 | "added v2.10.1" — PATCH understates a SemVer-breaking + behavior-flipping change | Release as 2.11.0 |
| MEDIUM | `crates/trusty-git-analytics/src/collect/identity/resolver.rs` | 215-220 | Silent behavior flip announced only at `debug!` | `info!` once, or a `tga collect` summary line |
| LOW | `crates/trusty-git-analytics/src/collect/identity/resolver.rs` | 232-234 | Doc names `trusty-review`'s selector as a caller of the switch; it never calls it | Correct the doc, state the intent |

---

### CRITICAL — the new test helper races, and the race can make the #4251 proof pass vacuously

`resolver_tests.rs:532-541`:

```rust
let unique = format!(
    "tga-issue-4251-{}-{}",
    std::process::id(),
    std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_nanos()
);
let tmp = std::env::temp_dir().join(unique);
```

Three tests call this helper — `aliases_file_disables_tier34_name_fuzzy`, `aliases_file_gate_preserves_tier12_declared_resolutions`, `explicit_opt_in_reenables_fuzzy_with_aliases_file`. They run **in parallel in the same process**, so `process::id()` is constant and the only entropy is a `SystemTime` reading whose real granularity on macOS is coarser than a nanosecond. They collide, overwrite each other's `config.yaml`, and each one's trailing `remove_dir_all(&tmp)` (`:606`, `:640`, `:660`) deletes the shared directory out from under whichever test is still running.

**Measured on this branch, unmutated, 100 consecutive runs: 10 failures.** Two distinct captured manifestations:

```
panicked at resolver_tests.rs:568:5:
supplying an aliases_file must disable the fuzzy fallback
```
→ this test wrote `extra_config_yaml = ""`, but loaded the `config.yaml` written by `explicit_opt_in_reenables_fuzzy_with_aliases_file`, which contains `fuzzy_identity_fallback: true`.

```
panicked at resolver_tests.rs:619:5:
  left: "whoever"
 right: "Akash Arora"
```
→ the alias file was gone before `Config::load`; `resolved_aliases()` swallowed the error and the resolver was built with **zero** aliases and **zero** members.

That second mode is why this is CRITICAL rather than HIGH. A resolver with zero members returns **every** input unchanged — which is exactly what `aliases_file_disables_tier34_name_fuzzy` asserts (`:600`). Under that race the primary #4251 test **passes for the wrong reason**, against a resolver that never loaded the roster. Its `assert!(!r.fuzzy_fallback())` at `:568` does not catch it either, because `fuzzy_fallback` is computed from `aliases_file.is_some()`, which is true regardless of load success. A green run of this suite is therefore not proof that the gate works.

CI on this PR is green, including `Test` (21m45s). At ~10% per run that is expected and is **not** evidence of absence.

Secondary: cleanup is only on the success path, so every panic leaks a directory — I found **56** stale `tga-issue-4251-*` dirs in `$TMPDIR` after my stress runs.

**Fix** — `tempfile` is already a dependency (`crates/trusty-git-analytics/Cargo.toml:129`):

```rust
fn resolver_from_aliases_file(extra_config_yaml: &str) -> (IdentityResolver, tempfile::TempDir) {
    let tmp = tempfile::TempDir::new().unwrap();   // collision-free, RAII cleanup on panic
    ...
}
```
Drop the three manual `remove_dir_all` calls. **And** add a non-vacuity assertion to `aliases_file_disables_tier34_name_fuzzy` so an empty resolver can never satisfy it, e.g.:

```rust
assert_eq!(r.resolve("whoever", "crislaine.tripoli@duettoresearch.com").0, "Crislaine Tripoli",
           "roster must actually be loaded, otherwise pass-through proves nothing");
```

### HIGH — `Closes #4251` overclaims

#4251 reports, verbatim: `Josh Taylor` → `Joshua Lepage` / `Joshua McCartney` **(inconsistent across runs/entries)**. This PR does not fix that symptom.

I ran the un-gated resolver 25 times: `Josh Taylor` resolved to `Joshua Lepage` **25/25**. The two scores are distinct (`jw("josh","joshua lepage")=0.8615` vs `jw("josh","joshua mccartney")=0.8500`), and a keep-the-max scan is order-independent for distinct scores, so the Tier-3/4 tie-break cannot produce the reported flip.

The live run-to-run nondeterminism is one tier higher, at `resolver.rs:173-175` (`origin/main:159-161`):

```rust
for a in alias_list {
    aliases.insert(a.to_lowercase(), canon_name.clone());
}
```

This is **last-writer-wins** over a `HashMap` iteration (`:160`). Two canonical names declaring the same lowercased alias produce a nondeterministic **Tier-1/2** mapping. Rust seeds each `HashMap` instance separately, so this varies even between two resolvers built in the same process. `find_member_by_name` (`resolver.rs:463-468`, main `:408-412`) has the same root cause. This PR gates Tiers 3/4 only, so it remains live on **every** path, `aliases_file` configs included.

I am not asserting this *is* the production cause — I cannot see the 177-entry table. I am asserting, at high confidence, that **the fuzzy tiers are not the cause**, and therefore that gating them cannot be claimed to fix it.

**Fix:** change `Closes #4251` to `Refs #4251` and let #4293 carry the determinism work, or fix the Tier-1/2 determinism in this PR. Do not auto-close #4251 on merge.

### HIGH — the gate keys on the declaration, not on the load

`resolver.rs:212-214`:

```rust
resolver.fuzzy_fallback = config
    .fuzzy_identity_fallback
    .unwrap_or(config.aliases_file.is_none());
```

`Config::resolved_aliases()` (`core/config/mod.rs:1277-1284`) deliberately swallows alias-file load errors and falls back. So a typo'd, unreadable, or malformed `aliases_file` now yields **both** an empty alias table **and** a disabled fuzzy fallback: every author fragments to its raw identity, and the only evidence is a `debug!`. Before this PR the fuzzy tier masked that misconfiguration.

This is not hypothetical — it is precisely the state my flaky-test capture produced (`left: "whoever"`).

**Fix:** derive the gate from the map that was actually resolved, and make the failure audible:

```rust
resolver.fuzzy_fallback = config
    .fuzzy_identity_fallback
    .unwrap_or(map.is_empty() && config.aliases_file.is_none());
```
plus a `warn!` when `aliases_file` is `Some` but `resolved_alias_map()` returned `Err`.

### MEDIUM — version class

`docs/trusty-git-analytics/requirements/configuration.md:222` states "added v2.10.1", i.e. a PATCH on the current 2.10.0. Two independent reasons that is wrong:

1. `Config` (`core/config/mod.rs:146`) has all-public fields and no `#[non_exhaustive]`, so external struct-literal construction is legal and adding `pub fuzzy_identity_fallback` is a breaking change under the Cargo SemVer reference.
2. The default flips runtime behavior for **every** existing `aliases_file` deployment on upgrade.

This repo has yanked a release for a patch-level SemVer break before. Recommend **2.11.0**.

### MEDIUM — the flip is announced only at `debug!`

`resolver.rs:215-220`. Operators upgrading get changed attribution with no visible signal at default log level. Raise to `info!` (once per run) or surface it in the `tga collect` summary.

### LOW — doc names a caller that does not call it

`resolver.rs:232-234` cites "`trusty-review`'s profile selector" as needing `with_fuzzy_fallback`. `crates/trusty-review/src/profile/selector.rs:277` calls `from_alias_map`, which hardcodes `fuzzy_fallback: true` (`resolver.rs:182`), and never calls the switch. I believe keeping fuzzy on there is **correct** (interactive profile lookup by partial name wants it), but say so explicitly rather than implying the switch is used. Worth noting downstream: that selector's fuzzy now runs over a more fragmented `authors` table.

### Not flagged
`developer_aliases` inline maps still get fuzzy ON by default; only `aliases_file` trips the gate. Asymmetric versus the stated rationale, but documented with an explicit `false` escape hatch. Design choice, not a defect.

---

## On the four implementer claims

**1. `email_domain()`, the domain gate, and the local-part comparison are byte-identical to `aaad0528` — YES, verified.** `diff` of the full `resolver.rs` between `aaad0528` and `origin/main` is **empty**, and `diff aaad0528 head` has **zero** deleted lines (`diff | grep -c '^<'` → `0`). `email_domain` (`:67-72`), the domain-equality gate (`:332-336`) and the local-part comparison (`:333`) are untouched. No regression to the #2244/#2253 fix.

**2. Tiers 1/2 untouched — YES.** `resolver.rs:293-304`, entirely above the gate at `:313-315`, unmodified. The short-circuit is a pure early `return` of the owned input; Tiers 3/4 have no side effects, no caching, no metrics, and populate no state any later tier reads (`best`/`best_norm` are function-locals). Nothing downstream is skipped — `upsert_author` (`:397`) consumes only `resolve()`'s return value.

**3. None of the 77 genuine corrections is undone — YES by construction, but the cited test only samples it.** The structural proof is far stronger than the test: zero deleted lines plus a gate placed strictly after Tiers 1/2 means no Tier-1/2 resolution *can* change. The four cases cover four alias **shapes** (declared `-c` secondary email, non-email login handle, `users.noreply`, case-folded canonical) — they sample the mechanism; they do not enumerate the 77, and none is drawn from the actual 130→177 diff. Also worth stating plainly: the change *does* intentionally undo correct fuzzy resolutions for undeclared near-name identities. The changelog says so; that is the accepted tradeoff, not a regression. Caveat: this is one of the three tests affected by the CRITICAL race.

**4. `issue_2253_domain_gate_intact_when_fuzzy_enabled` verifies both directions — YES, and it is a good test.** Snyk suppressed (`:749-751`), `acoopr` → `Alice Cooper` still resolves (`:754-756`), and the precondition assert at `:744-748` proves full-string similarity really clears 0.85 so it cannot pass vacuously. It builds via `IdentityResolver::new`, so it is unaffected by the temp-dir race.

---

## On the anomaly — the single green pre-fix run

Flagging it rather than burying it was the right call and I credit it. But **do not close it as "stale build"** — I found two concrete, reproducible mechanisms that make this suite report green against broken code, and neither was identified:

**(a) A zero-match test filter reports success.** The file is `resolver_tests.rs` but the module is `collect::identity::resolver::tests` (via the `#[path]` attr at `resolver.rs:471-473`). The natural filter to type is therefore wrong, and it reports `ok`:

```
$ cargo test -p tga --lib collect::identity::resolver_tests
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 771 filtered out
```

One character of module-path drift converts a red suite into a green "ok". This is the standing "0 tests ran is NOT passing" trap and it is my leading hypothesis for the all-green run.

**(b) The temp-dir race (CRITICAL above).** It can green `aliases_file_disables_tier34_name_fuzzy` against un-gated code by handing it an empty resolver. It cannot green the other two gate tests (neither touches the filesystem), so it does not by itself explain an *all*-green pre-fix run — but it proves this suite can pass against broken code, which is the property that matters.

My mutation run is the counter-evidence that the fix itself is real: deleting the `if !self.fuzzy_fallback` block turns the suite red **25/25**, and the failure output reproduces all five documented misattributions exactly:

```
Cristian Dominguez  => Crislaine Tripoli
Ravi Chandrasekaran => Ravi Pandey
Gauri Saykar        => Gaurav Sharma
Josh Taylor         => Joshua Lepage
Joseph Ku           => Joshua Lepage
```

So the gate is genuinely load-bearing. The gate *around* the gate is not.

---

1. Replace the hand-rolled temp dir in `resolver_tests.rs:532-556` with `tempfile::TempDir`; drop the three manual `remove_dir_all` calls. Evidence required: 100 consecutive green runs of `cargo test -p tga --lib collect::identity::resolver`, output pasted.
2. Add a non-vacuity assertion to `aliases_file_disables_tier34_name_fuzzy` so an empty resolver cannot satisfy it.
3. Change `Closes #4251` → `Refs #4251` (or fix the Tier-1/2 last-writer-wins determinism here). #4251 must not auto-close.
4. Gate on the resolved alias map rather than on `aliases_file.is_none()`, and `warn!` when a declared `aliases_file` fails to load.
5. Release as **2.11.0**; correct `configuration.md:222`.
6. Raise the disabled-fallback log from `debug!` to `info!`.
7. Root-cause the green pre-fix run against mechanism (a) above, or state on the ticket that (a) is the accepted explanation. Do not leave it as "stale build".

Items 5-6 and the LOW doc fix are cheap; items 1-4 are the gate.

bobmatnyc:
## Round 2 — all four findings addressed, evidence inline

Thank you for the mutation test and the 100-run measurement. Both were correct
and both found real defects. Head is now `c203f660` (branch rewritten to a
single commit so no `Closes #4251` survives in history). Full detail is in the
updated PR body; summary and the key numbers here.

**1. CRITICAL — racing helper.** Fixed with `tempfile::TempDir`; the three
manual `remove_dir_all` calls are gone, and I applied the same fix to the
pre-existing `duetto_contractors_config_resolves`, which had the identical
`pid + nanos` scheme.

I reproduced your measurement at `ded54b76` before fixing — **12 failures in 100
runs**, with both of your captured modes:

```
   6  left: "whoever"  right: "Akash Arora"                      (:619, zero-member)
   3  supplying an aliases_file must disable the fuzzy fallback  (:568, cross-config)
   2  unwrap panics at :542 / :554
```

And I confirmed the vacuous-pass direction directly rather than by inference. A
throwaway diagnostic on pre-fix code, `aliases_file` pointed at a missing file:

```
resolved_aliases() empty? true
fuzzy_fallback()        = false
assert!(!r.fuzzy_fallback())  -> PASSED
  Cristian Dominguez     => Cristian Dominguez (wrong_target=Crislaine Tripoli)
  ... all five ...
ALL PRIMARY ASSERTIONS PASSED against a ZERO-MEMBER resolver.
```

You were exactly right. Permanent guards: `assert_roster_loaded()` called first
in all three `aliases_file` tests, plus `non_vacuity_control_rejects_zero_member_resolver`,
a `#[should_panic]` test pinning that the control actually rejects a zero-member
resolver. **Post-fix: 100/100 green, 0 failures**, with every run required to
execute exactly 33 tests — a run executing 0 tests is scored FAIL by the harness.

**2. HIGH — `Closes` → `Refs`.** Done. I withdraw my round-1 tie-break claim:
the two Joshua scores are distinct (0.8615 vs 0.8500) and a keep-the-max scan is
order-independent for distinct scores, so the Tier-3/4 tie-break cannot produce
the reported flip. My claim was wrong; the last-writer-wins `HashMap` alias
insert is the live cause and #4293 carries it.

**3. HIGH — gate on the load.** Fixed, but I deviated from the suggested
one-liner deliberately, because I believe it does not do what the finding asks:

```rust
.unwrap_or(map.is_empty() && config.aliases_file.is_none())
```

With a broken alias file this is `true && false` = `false` — fuzzy still off,
bug unchanged. With a non-empty inline `developer_aliases` map and no
`aliases_file` it is `false && true` = `false`, silently flipping every
inline-alias deployment to fuzzy-off. I implemented an explicit
`alias_table_loaded` check (declared **and** `Ok` **and** non-empty), `error!`
on both failure branches, and added `inline_developer_aliases_do_not_disable_fuzzy`
which fails under the suggested form. Happy to switch if you read it differently.

**4. MEDIUM — version and visibility.** Bumped 2.10.0 → **2.11.0** (`Cargo.lock`
moved by exactly one line, no dependency refresh); docs and changelog updated.
The flip is announced at `info!` with the remediation inline. The
`with_fuzzy_fallback` doc no longer claims trusty-review's selector calls it.

**5. My "stale build" anomaly — retracted.** It was the temp-dir race, in the
zero-member mode above. Mechanism confirmed by reading the code: `Config::load`
does not read the alias file; resolution is lazy inside `resolved_aliases()`,
called later by `from_config`. So the victim loads `config.yaml` fine, the
colliding test's cleanup deletes the directory, and the *lazy* alias load fails
into the swallowed-error path — which explains both halves of that run (primary
passed vacuously, Tier-1/2 had already loaded and passed normally).

Your `#[path]` trap is real and I confirmed it, but it is not what I hit:

```
filter 'collect::identity::resolver_tests' -> ok. 0 passed; 774 filtered out
filter 'collect::identity::resolver::'     -> ok. 33 passed; 742 filtered out
```

My logs from that run show `running 25 tests` with the target printing `... ok`.
Either way it can no longer hide: the stress harness fails any run that does not
execute the expected count.

**Red-before / green-after, re-run post-tempfile.** Filter:
`cargo test -p tga --lib 'collect::identity::resolver::'` (33 tests, count
asserted). Gate mutated out of `resolve()`, `cargo clean -p tga` first,
tempfile helper and non-vacuity control left in place:

```
BEFORE: test result: FAILED. 30 passed; 3 failed; 742 filtered out
        all five misattributions reproduce, non-vacuity control PASSING
AFTER:  test result: ok. 33 passed; 0 failed; 742 filtered out
```

Gates: full `cargo test -p tga` (no `--lib`) 774/151/1+2+1/5 green,
`clippy --all-targets -D warnings` exit 0, `fmt --check` exit 0,
`check_line_cap.sh` 0 violations, `cargo check -p trusty-review` exit 0.

Not merging — over to you.

bobmatnyc:
## Verdict: APPROVE

Re-review of `c203f660` against `origin/main`. Every claim below I re-ran myself; no pasted output was taken on trust.

All three blocking findings are closed, and **the one place you deviated from my instruction, you were right and I was wrong.** Detail below, because that adjudication matters more than the rest of this review.

---

## The deviation on the gate — you are correct, my round-1 fix was defective

I asked for `.unwrap_or(map.is_empty() && config.aliases_file.is_none())`. You declined and implemented an explicit `alias_table_loaded` check. I applied **my own suggested form** to your branch and ran the suite:

```
test result: FAILED. 29 passed; 4 failed
    broken_aliases_file_keeps_fuzzy_enabled
    empty_aliases_file_keeps_fuzzy_enabled
    inline_developer_aliases_do_not_disable_fuzzy
    no_aliases_file_keeps_fuzzy_enabled_by_default
```

My form was wrong in **both** directions you identified, and in a fourth way neither of us named:

- broken alias file → `true && false` = fuzzy **off**. The exact bug I opened the finding to fix, left unfixed.
- empty alias file → same, unfixed.
- inline `developer_aliases`, no alias file → `false && true` = fuzzy **off**. Silent scope creep.
- **`no_aliases_file_keeps_fuzzy_enabled_by_default` also breaks** — my form would have broken the stated backward-compatibility guarantee for every inline-alias deployment, which is a strictly worse outcome than the bug I was trying to fix.

Your implementation at `resolver.rs:220-243` is correct on every branch I can construct:

| config | `alias_table_loaded` | fuzzy | correct? |
|---|---|---|---|
| no `aliases_file` (± inline) | `false` (None arm, `:221`) | ON | yes — historical, no scope creep |
| `aliases_file` loads non-empty | `true` (`:223`) | OFF | yes — this is #4251 |
| `aliases_file` declared, `Err` | `false` + `error!` (`:231-238`) | ON | yes — fail-safe to documented pre-#4251 behaviour |
| `aliases_file` loads empty | `false` + `error!` (`:224-230`) | ON | yes |
| explicit `true`/`false` | — | wins | yes |

Failing **safe** (degrade to known behaviour, log at `error!`) rather than failing into the worst-of-both state is the right call. Deviating and flagging it was the right process. I was wrong; you were right; noted plainly.

## 1. CRITICAL race — closed, verified independently

`resolver_from_aliases_file` (`resolver_tests.rs:538-553`) and `duetto_contractors_config_resolves` (`:211`) both on `tempfile::TempDir`. Zero occurrences of `temp_dir()`, `process::id`, or `remove_dir_all` remain in the file. Fixing the pre-existing `duetto_contractors_config_resolves`, which I did not flag, was correct — it carried the identical scheme.

My run, 100 consecutive iterations, scoring a run green **only** on an exact `33 passed; 0 failed` match so a 0-test run counts as failure:

```
GREEN(exactly 33): 100   NON-GREEN: 0
```

Against my round-1 measurement of 10/100 failures on `ded54b76`, that is closed.

## 2. Non-vacuity guards — they would catch a recurrence

`assert_roster_loaded` (`:562-577`) asserts two positive resolutions, and is called **before** the pass-through block in both tests whose assertions a zero-member resolver could satisfy (`:595`, `:673`). `aliases_file_gate_preserves_tier12_declared_resolutions` does not call it and does not need to — every one of its assertions is a positive canonical resolution that a zero-member resolver fails outright.

`non_vacuity_control_rejects_zero_member_resolver` (`:691-709`) is the part I want to single out: a `#[should_panic(expected = "non-vacuity")]` that pins the guard itself as load-bearing. That is the right instinct — an unproven guard is not a guard. I confirmed it fires.

Mutation re-run (gate deleted from `resolve()`), my own run:

```
test result: FAILED. 30 passed; 3 failed
    aliases_file_disables_tier34_name_fuzzy
    explicit_opt_out_disables_fuzzy_without_aliases_file
    with_fuzzy_fallback_false_suppresses_tier34
```
all five misattributions reproduced verbatim, and `non_vacuity_control_rejects_zero_member_resolver` **passed** — i.e. the primary test went red for the real reason, not through vacuity. Matches your reported BEFORE exactly.

## 3. `Refs #4251` — verified at the API level, not by eye

Two `Closes #4251` strings survive in the PR body at lines 265/268, but both sit inside inline code spans, which GitHub excludes from closing-keyword parsing. Confirmed authoritatively rather than by reading:

```
$ gh api graphql -f query='{...pullRequest(number:4291){closingIssuesReferences(first:10){nodes{number}}}}'
{"data":{"repository":{"pullRequest":{"closingIssuesReferences":{"nodes":[]}}}}}
```

Empty. #4251 will not auto-close on merge. Correct.

Withdrawing the round-1 tie-break claim was right. #4293 carries the real Tier-1/2 cause.

## 4. MEDIUM items — all closed

- `Cargo.toml` + `Cargo.lock` → 2.11.0. **Lockfile moved by exactly one line** (the `tga` version), no dependency refresh — verified against the standing rule.
- CHANGELOG states the MINOR rationale explicitly (public field on an all-public, non-`#[non_exhaustive]` struct + default behaviour flip).
- Docs say `added v2.11.0` / `pre-2.11.0 behaviour`.
- `info!` at `resolver.rs:248-253` with the remediation inline; `use tracing::{debug, error, info}` at `:19`.
- `with_fuzzy_fallback` doc (`:265-275`) now says "No caller sets it today; it exists so the switch is reachable without routing through `Config`" — accurate, and better than what I asked for.

## 5. Commit hygiene — the `git reset --soft` incident did not ship

This was worth checking exactly as you self-reported it. Single commit `c203f660`, 7 files, `636 insertions(+), 17 deletions(-)`. I enumerated **every** deleted line in the diff:

- `version = "2.10.0"` ×2 (Cargo.toml, Cargo.lock)
- `use tracing::debug;` (widened to `{debug, error, info}`)
- the 12-line `pid + nanos` block and `remove_dir_all` in `duetto_contractors_config_resolves`, plus 2 stale doc lines

Zero unrelated reverts. No `trusty-mpm/daemon/bus` files present. Those 6 files are on `main` in `6ffe001d`/`#4261` and are untouched here. Catching that in `git show --stat` and recovering with an explicit `git reset b9db3fbf` was the right call.

## 6. The anomaly — accepted, with one caveat stated honestly

Your mechanism is correct and I verified its load-bearing premise: `Config::load` (`config/mod.rs:1226-1233`) parses YAML and sets `source_path` only — it never reads the alias file. The alias load is lazy, inside `from_config` → `resolved_aliases()`. So a victim could `Config::load` successfully and *then* have the lazy load fail after a colliding test's cleanup removed the directory. That is exactly the state I captured independently in round 1 (`:619`, `left: "whoever"`), so the zero-member mode is proven real, not inferred.

Your exclusion of the `#[path]` trap is also sound: that trap produces `running 0 tests`, and a log showing `running 25 tests` with the target printing `... ok` rules it out for that run.

Caveat I will state rather than paper over: whether the race explains a **fully** green pre-fix run depends on which tests existed at that moment. In round 1 I showed the race can only green `aliases_file_disables_tier34_name_fuzzy` — the other two gate tests never touch the filesystem. `running 25 tests` against the 29 I measured at `ded54b76` suggests four had not been written yet, which would make the account complete; I cannot reconstruct that working tree to confirm it. I am not treating this as an open defect, because the property that actually matters is now pinned by machinery I verified: the suite can no longer report green against a zero-member resolver, and a `#[should_panic]` control proves that guard fires.

Separately: the `#[path]` trap is real and still latent repo-wide (`filter 'collect::identity::resolver_tests'` → `0 passed; 774 filtered out`, reported as `ok`). That is not this PR's problem — flagging it for a future CI guard, not as a finding here.

## Notes for the merge

- **No rebase needed.** Branch shows BEHIND by 4 commits (`54bba9f7`, `61329940`, `d492fd34`, `6ffe001d`). None touches `crates/trusty-git-analytics/`, and none touches `Cargo.lock` or `Cargo.toml` — so the one-line lockfile bump cannot conflict. Squash-merge is safe as-is.
- CI: 23 pass, 0 fail, 1 skip (`Notify on red main`, which only fires on red).
- Sole remaining nit, **not** a finding and not worth a commit: `from_config` now reads the alias file twice (`resolved_aliases()` at `:196`, then `resolved_alias_map()` at `:222`). Once per resolver construction, so the cost is irrelevant, and the TOCTOU window is microseconds. Mentioned only so it is a known property rather than a surprise.

No issues found at >80% confidence. **APPROVED** — clear to merge on green.

Issue #4288, opened 2026-07-29
#4207 slice 3: reconcile worktrees on a git-derived key, report-only


bobmatnyc:
Parent epic: #4207 (this is its slice 3). Constraining input: #4269. Follow-on: #4270.

### Defect

Worktree reconciliation has nothing to reconcile with. Measured today: **118 registered worktrees across two disjoint git registries** (`.base` bare clone: 60 entries; sibling non-bare clone: 58; **intersection: 0**). `sessions.json` holds 56 records, 31 with a `workspace_path`, but **only 3 of 118 worktrees have a matching session record** — 115 have none (64 harness `.claude/worktrees`, 27 tmp scratchpads, 17 tm-shaped `.worktrees`, 7 other). Only 5 of 56 records set `worktree_owner`.

**Path shape is provably the wrong discriminator — it misclassifies 34 of 118 (29%):** 14 paths containing `/.base/` are registered in the *sibling* registry; 20 paths *not* containing `/.base/` are registered in the *base* registry. Decisive counterexample, two sibling directories sharing a prefix, different registries:

```
.base/.worktrees/fix-3822-sessions-list-empty  → common-dir /…/trusty-tools/.git   (sibling)
.base/.worktrees/2eb72dca-de08-…               → common-dir /…/trusty-tools/.base  (base)
```

### Already shipped — do not re-derive

Slice 1 already replaced the path-shape walk with the git-derived enumeration: `session_manager/worktree_registry.rs:203` (`git worktree list --porcelain`), `:166` (`git rev-parse --path-format=absolute --git-common-dir`), `:306` (two anchors per project, const `:63`; anchor-is-registry-root check `:372`; drops `bare|is_main|prunable|locked` `:396`; strict-descendant containment `:414-417`). `prune.rs:417` is now a filter over it. Slice 3 builds on this.

### Evidence — the live safety counterexample

Session `2eb72dca-de08-481b-8dfa-22ab7f81b1f9` is **running right now** (tmux pane `%981`, `pane_current_path` inside its worktree), recorded in `sessions.json` as **`state: "stopped"`**, holding **12 modified tracked files, 31 untracked, 1 unpushed commit**, with a **zero-byte** sentinel (→ `SentinelOwner::Unknown`, `worktree_ownership.rs:135-137`).

[...]

Issue #1845, opened 2026-06-30
tm: follow-up hardening for session-worktree prune/decommission (residuals from #1844)


bobmatnyc:
## Context
Follow-up hardening work identified during trusty-review audit of #1844 (session-worktree lifecycle management, #1840). These items were confirmed as non-blocking (PR received APPROVE* after two remediation passes) but represent important test-coverage gaps and medium/low-priority correctness/performance edge-cases.

**Related to:** #1840, PR #1844

---

### Test Coverage & CI Flakes

- [ ] **prune.rs ~L617 — `prune_orphaned_worktrees_fresh_active_set_blocks_deletion` missing Phase-2 async deletion guard coverage**
  - The TOCTOU safety guard (per-candidate re-query path) has no real executed test coverage. Add a test that genuinely triggers the re-query and asserts a session appearing after the snapshot is NOT deleted.

- [ ] **doctor.rs ~L591 — `worktrees_no_orphans_is_ok` spurious macOS CI flakes**
  - Test can fail on macOS CI due to `/tmp` → `/private/tmp` symlink. Canonicalize paths in test setup to fix flakiness.

- [ ] **claude_code.rs ~L499 — `spawn_resume_without_id_no_prior_conv_sends_plain_spawn` integration test silent skip**
  - Test silently skips when the `claude` binary is absent, leaving the new plain-spawn resume path unverified in CI. Make it run without the real binary (inject/mock) or assert the constructed command.

### Correctness & Data Safety

- [ ] **decommission.rs ~L75 — `remove_session_worktree` external `git` call lacks timeout**
  - Spawns external git with no timeout. Decommission could block indefinitely if git hangs, stalling the session-manager daemon. Add a timeout.

- [ ] **decommission.rs ~L35 — `is_session_worktree` hardcoded convention (data-safety edge case)**
  - Relies on hardcoded `.worktrees` parent-name convention rather than a stored flag. A user-owned dir that is a direct child of a managed `.worktrees/` root could be misclassified. Prefer a stored marker/flag identifying tm-created worktrees.

- [ ] **managed.rs ~L499 — Non-string JSON entries in prune `paths` array silently skipped**
  - Silent skip understates the removed/would-remove count. Handle non-string entries explicitly or warn.

- [ ] **managed_routes/prune.rs ~L104 — Prune route returns no HTTP error on failure**
  - Operators cannot distinguish a clean empty sweep from a partial failure. Surface errors in HTTP response.

[...]

PR #4294, opened 2026-07-29
test(trusty-mpm): pin the deliberately-unfiltered prune active reads, manual + automatic (#4288)

bobmatnyc:
## What

Pins the "deliberately unfiltered by session state" property of the orphan-worktree sweep with regression tests, explanatory comments, and one rename. **No behavior change.**

Standalone extraction of **item 4** of #4288, itself in-scope under #4207. #4288 stays **open** — this PR lands one item of it, not the rest of slice 3.

## The property

Three reads build "which workspaces are in use", none filtered by record state:

| # | Site | Role |
|---|---|---|
| M1 | `daemon/managed_routes/prune.rs` — `prune_worktrees_route` | caller set, manual HTTP route |
| M2 | `session_manager/prune.rs` — `reap_orphaned_worktrees` | caller set, **automatic** GC loop (`dry_run: false`, no preview) |
| M3 | `session_manager/prune.rs` — Phase 2 `fresh_in_use` in `prune_orphaned_worktrees` | re-read from the store immediately before each deletion |

**Session state is not a liveness signal.** Measured 2026-07-28: session `2eb72dca-…` was **running** (tmux pane `%981`, cwd inside its own worktree) while `sessions.json` recorded it `state: "stopped"`, holding 12 modified tracked files, 31 untracked files, and 1 unpushed commit.

These are **defense-in-depth: no single read is load-bearing.** Narrowing one alone never deletes; narrowing a caller set *and* Phase 2 does. The comments say so and warn against the failure mode this invites — narrowing one "because the other covers it", an argument that applied twice collapses a pair of independent boundaries into none.

## Changes

- **Comments** at all three reads, each naming the test that trips on it.
- **`prune_spares_a_stopped_records_workspace`** — drives the real route, dry-run only (see below).
- **`reap_spares_a_stopped_records_workspace`** — real deleting sweep on the automatic path.
- **`phase2_fresh_snapshot_spares_a_record_the_caller_set_missed`** — closes a pre-existing hole (below).
- **Rename** `active_workspace_paths` → `in_use_workspace_paths` (and `initial_active`/`fresh_active` → `initial_in_use`/`fresh_in_use`). Rename only. The old name is what made `state == Active` read as a tidy-up rather than a data-loss bug.
- `CHANGELOG.md`.

Deferred deliberately: folding the sites into a shared helper (behavior-affecting; this PR is not).

## Review round 2 — the BLOCK is addressed

**1. The destructive sweep is gone.** The site-1 test previously ran a real deleting sweep whose target root resolves inside the route from process-global `TRUSTY_MPM_WORKSPACE_ROOT` — which `connectors::tm_tests` mutates in the same binary under `#[serial_test::serial]`, a guard that does **not** interoperate with `env_test_lock`. A restore landing in that window would have pointed a deleting sweep at the operator's real `~/trusty-mpm-projects`, and the three-read defense cannot help there: the caller set and the Phase 2 re-read both draw from the same empty isolated store. It also bought **zero** mutation coverage — its assertions pass under M1. The test now issues exactly one route call, `dry_run: true`, and carries **both** guards (`env_test_lock` *and* `#[serial_test::serial]`). The dry-run assertions are what pin site 1.

**2. M3 now has a pin — and the hole was pre-existing.** The Phase 2 re-read is the last check between a reclaimable candidate and `remove_session_worktree`, and it had **zero executed coverage**. The only test naming it, `prune_orphaned_worktrees_store_snapshot_blocks_deletion`, builds its worktree with **no ownership sentinel**, so the #3649 gate classifies it owner-unknown and skips it *before* Phase 2 runs — as that test's own comment states. That is why filtering Phase 2 broke nothing. The new test uses an empty caller set (the TOCTOU case Phase 2 exists for, #1845 item 9) with a reclaimable sentinel; the **control** is the dry-run of the same sweep, which returns before Phase 2 and still lists the worktree — so the delta between control and real run *is* the Phase 2 protection, and cannot be faked by a fixture that was simply never reclaimable.

**3. Parameter renamed**, as above.

**Accuracy correction on site 2.** The literal `== Active` mutation at the reap site was **already** caught before this PR, because `create_with_id` persists records as `Provisioning`, not `Active`. Site 2's real value is the **Stopped** case — a state that reads terminal but is routinely live — and the pair property. The comment and test doc now say that rather than implying the test catches the tidy-up directly.

## Isolation matrix — re-run in full after the changes

| mutation | site-1 test | site-2 test | Phase 2 test | pre-existing reap test | suite |
|---|---|---|---|---|---|
| **pristine** | ok | ok | ok | ok | **3897 passed, 0 failed** |
| **M1** route set | **FAILED** | ok | ok | ok | red |
| **M2** reap set | ok | ok | ok | **FAILED** | red |
| **M3** Phase 2 | ok | ok | **FAILED** | ok | red |
| **M2+M3** | ok | **FAILED** | **FAILED** | **FAILED** | red |

**Every read is now caught by something in isolation.** M3 previously left the suite fully green.

### Gates

```
cargo test -p trusty-mpm          (without --lib)
test result: ok. 3897 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 32.10s
```

`cargo clippy -p trusty-mpm --all-targets -- -D warnings` → exit 0; `cargo fmt --check` → exit 0; `scripts/check_line_cap.sh` → 0 violations.

Diff vs merge-base: 6 files, +530 / −18 — comments, tests, the rename, and `CHANGELOG.md`.

Refs #4288, #4207

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -20,6 +20,16 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   inviting `.filter(|r| r.state == Active)` tidy-up would make a live worktree
   an orphan candidate. The test drives the real route and fails if that filter
   is ever added.
+- `session_manager::prune`: the same pin extended to the two remaining unfiltered
+  reads on the AUTOMATIC orphan-GC path — the active set built in
+  `reap_orphaned_worktrees` (what the daemon's sweep loop calls on a timer, with
+  `dry_run: false` hardcoded and no preview) and the Phase 2 `fresh_active`
+  snapshot inside `prune_orphaned_worktrees` that re-reads the store immediately
+  before deletion. Both are now commented, and
+  `reap_spares_a_stopped_records_workspace` asserts end-to-end that a `Stopped`
+  record's worktree survives a real, deleting sweep. Measured and documented:
+  these two reads are defense-in-depth — neither is load-bearing alone, and only
+  narrowing BOTH actually deletes a live worktree. No behavior change.
 
 - **PM instructions are authored per section** (epic [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183)): the four monolithic assets (`PM_INSTRUCTIONS.md`, `WORKFLOW.md`, `AGENT_DELEGATION.md`, `BASE_PM.md`) are replaced by one markdown source per section under `src/assets/instructions/sections/`, and `core::bundled_pm_package` builds its blocks from those files instead of cutting the monoliths at runtime (the `split_asset` / `Cut` machinery is gone). `instruction_pipeline::pm_instructions()` and `base_pm()` reconstitute the two multi-section strings the legacy override assembly still needs, so a section edit reaches BOTH composers and a project with a `.trusty-mpm/` override can no longer be frozen on stale instructions. Every existing override file (`INSTRUCTIONS.md`, `WORKFLOW.md`, `AGENT_DELEGATION.md`, `MEMORY.md`, `PM_INSTRUCTIONS_DEPLOYED.md`) keeps resolving exactly as before.
 - **Delivered-prompt content changes** (same epic). The Memory and Search guidance is lifted out of the middle of the Core body into two self-contained sections with their own headings — they were two halves of one numbered list, which made their declared `project` tier a false claim (replacing Memory alone left Search opening on a bare `2.`); they are now genuinely independently overridable. `## Trusty Tool Priority (Non-Overridable)` moves to sit with the other non-overridable rules, so it now precedes the framework-guaranteed conventions instead of following them — position only, still inside the floor, wording unchanged. The workflow section gains a **Sprint, then Harden** doctrine (sprint to feature-complete with targeted tests and no CI iteration loops; harden with the full suite, critic and release gates before publishing), including the causal claim that slow release *causes* WIP, the hard line that going fast never licenses turning red green by deleting coverage, and the close-and-fold rule at 3+ review rounds.
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -9,6 +9,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- `daemon::managed_routes::prune`: the orphan sweep's `active_workspace_paths`
+  set is now documented as DELIBERATELY unfiltered by session state, and pinned
+  by a new regression test `prune_spares_a_stopped_records_workspace`
+  ([#4288](https://github.com/bobmatnyc/trusty-tools/issues/4288),
+  [#4207](https://github.com/bobmatnyc/trusty-tools/issues/4207)). No behavior
+  change — the construction is untouched. A `SessionRecord`'s state is
+  bookkeeping, not a liveness signal (a session measured running in tmux was
+  recorded `stopped` while holding uncommitted and unpushed work), so the
+  inviting `.filter(|r| r.state == Active)` tidy-up would make a live worktree
+  an orphan candidate. The test drives the real route and fails if that filter
+  is ever added.
+
 - **PM instructions are authored per section** (epic [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183)): the four monolithic assets (`PM_INSTRUCTIONS.md`, `WORKFLOW.md`, `AGENT_DELEGATION.md`, `BASE_PM.md`) are replaced by one markdown source per section under `src/assets/instructions/sections/`, and `core::bundled_pm_package` builds its blocks from those files instead of cutting the monoliths at runtime (the `split_asset` / `Cut` machinery is gone). `instruction_pipeline::pm_instructions()` and `base_pm()` reconstitute the two multi-section strings the legacy override assembly still needs, so a section edit reaches BOTH composers and a project with a `.trusty-mpm/` override can no longer be frozen on stale instructions. Every existing override file (`INSTRUCTIONS.md`, `WORKFLOW.md`, `AGENT_DELEGATION.md`, `MEMORY.md`, `PM_INSTRUCTIONS_DEPLOYED.md`) keeps resolving exactly as before.
 - **Delivered-prompt content changes** (same epic). The Memory and Search guidance is lifted out of the middle of the Core body into two self-contained sections with their own headings — they were two halves of one numbered list, which made their declared `project` tier a false claim (replacing Memory alone left Search opening on a bare `2.`); they are now genuinely independently overridable. `## Trusty Tool Priority (Non-Overridable)` moves to sit with the other non-overridable rules, so it now precedes the framework-guaranteed conventions instead of following them — position only, still inside the floor, wording unchanged. The workflow section gains a **Sprint, then Harden** doctrine (sprint to feature-complete with targeted tests and no CI iteration loops; harden with the full suite, critic and release gates before publishing), including the causal claim that slow release *causes* WIP, the hard line that going fast never licenses turning red green by deleting coverage, and the close-and-fold rule at 3+ review rounds.
 - `core::pm_prompt_golden_tests`: committed snapshots of the fully composed PM prompt for both the packaged and legacy composers. #4249's byte-equality-against-the-old-prompt gate cannot survive a deliberate content change, so these replace it: any future edit to a section source surfaces as a reviewable prose diff in the PR rather than landing invisibly. Regenerate with `UPDATE_GOLDEN=1 cargo test -p trusty-mpm golden`.
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -20,6 +20,17 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   inviting `.filter(|r| r.state == Active)` tidy-up would make a live worktree
   an orphan candidate. The test drives the real route and fails if that filter
   is ever added.
+- `session_manager::prune`: `prune_orphaned_worktrees`'s `active_workspace_paths`
+  parameter is renamed **`in_use_workspace_paths`** (and its `initial_active` /
+  `fresh_active` locals to `initial_in_use` / `fresh_in_use`). Rename only, no
+  behavior change: the old name is what made `.filter(|r| r.state == Active)`
+  read as a tidy-up rather than a data-loss bug.
+- `session_manager::prune`: `phase2_fresh_snapshot_spares_a_record_the_caller_set_missed`
+  closes a pre-existing coverage hole — the Phase 2 `fresh_in_use` re-read, the
+  last check between a reclaimable candidate and deletion, had ZERO executed
+  coverage. The one test that named it builds its worktree with no ownership
+  sentinel, so the #3649 gate skipped it before Phase 2 ever ran; filtering
+  Phase 2 by state broke nothing in the suite.
 - `session_manager::prune`: the same pin extended to the two remaining unfiltered
   reads on the AUTOMATIC orphan-GC path — the active set built in
   `reap_orphaned_worktrees` (what the daemon's sweep loop calls on a timer, with
--- a/crates/trusty-mpm/src/daemon/managed_routes/prune.rs
+++ b/crates/trusty-mpm/src/daemon/managed_routes/prune.rs
@@ -105,13 +105,37 @@ fn default_dry_run() -> bool {
 /// default, and so is skipping dirty worktrees — `discard_dirty: true` is the
 /// only path to a destructive removal of unsaved work.
 /// Test: `prune_worktrees_route_dry_run`,
-/// `prune_worktrees_request_defaults_to_skip_dirty`.
+/// `prune_worktrees_request_defaults_to_skip_dirty`,
+/// `prune_spares_a_stopped_records_workspace` (#4288).
 pub async fn prune_worktrees_route(
     State(state): State<Arc<DaemonState>>,
     Json(req): Json<PruneWorktreesRequest>,
 ) -> impl IntoResponse {
     let mgr = state.session_manager().await;
     let records = mgr.list().await;
+    // #4288 (item 4 of #4207): DELIBERATELY UNFILTERED. Do NOT "tidy this up"
+    // by adding `.filter(|r| r.state == ManagedSessionState::Active)` — every
+    // record with a `workspace_path` belongs in this set, whatever its state.
+    //
+    // Why: a `SessionRecord`'s state is bookkeeping, NOT a liveness signal. It
+    // is written by reconcile/stop/hook paths that can miss, race, or be
+    // skipped entirely, so live sessions are routinely observed carrying a
+    // terminal state. Measured on this repo 2026-07-28: session
+    // `2eb72dca-de08-481b-8dfa-22ab7f81b1f9` was RUNNING (tmux pane `%981`,
+    // `pane_current_path` inside its own worktree) while `sessions.json`
+    // recorded it as `state: "stopped"`, holding 12 modified tracked files,
+    // 31 untracked files, and 1 unpushed commit.
+    //
+    // What narrowing this set would cost: such a worktree stops being spared
+    // and becomes an orphan CANDIDATE. Everything after that is a gate that
+    // can be open — the #3649 ownership sentinel is the last one, and in the
+    // measured case it was ZERO BYTES (`worktree_ownership::read_sentinel_owner`
+    // correctly maps that to `SentinelOwner::Unknown`, which spares it — but a
+    // well-formed sentinel naming a purged owner does not). This set is what
+    // keeps a live-but-mislabelled worktree off the candidate list at all.
+    //
+    // Pinned by `prune_spares_a_stopped_records_workspace` in this file's test
+    // module — if you broke that test, this comment is what you tripped.
     let active_workspace_paths: Vec<std::path::PathBuf> = records
         .iter()
         .filter_map(|r| r.workspace_path.clone())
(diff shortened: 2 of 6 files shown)


bobmatnyc:
## Early advisory — NOT the verdict (review of the final two-site branch is still pending)

Posting this ahead of the formal review because a second commit is inbound for
`session_manager/prune.rs:965` and it should **not** copy the pattern used at the
first site. Two things I verified against commit `09d05f77`.

### 1. Do not copy the env-override pattern into the site-2 test

The site-1 test issues a REAL, deleting sweep — `prune.rs:365-376`,
`PruneWorktreesRequest { dry_run: false, .. }` — and the directory that sweep
targets is not passed in. It is resolved *inside* the route from the
process-global `TRUSTY_MPM_WORKSPACE_ROOT`
(`prune.rs:143-144` → `trusty_tools_config::workspace_root`). The test's only
control over that is a `set_var` held under `env_test_lock`.

`connectors::tm_tests::create_session_full_lifecycle` mutates the **same**
variable (`tm_tests.rs:285-288` and `408-412`) in the **same** lib test binary,
serialised by `#[serial_test::serial]` — which does **not** interoperate with
`env_test_lock`. Nothing prevents those two tests from running concurrently.
If that test's restore lands inside this test's window, `workspace_root()`
falls through to the operator's real `~/trusty-mpm-projects`, and
`prune_orphaned_worktrees(.., dry_run = false, ..)` runs against it — with a
freshly-isolated, EMPTY session store, so every real sentinel resolves
"owner not found" → past grace → provably ownerless. `workspace_guard::is_safe_to_remove`
is not on this path. Clean, fully-pushed worktrees would be removed (and
`remove_session_worktree` follows up with `git branch -D`).

**Site 2 does not need any of this.** `reap_orphaned_worktrees(&self, repos_root)`
(`session_manager/prune.rs:961-963`) takes the root as an explicit parameter, so
the site-2 test can pass `&fx.repos_root` directly — no env mutation, no
`await_holding_lock` allow, no blast radius. Please use that.

### 2. The site-1 test can go silently vacuous — its CONTROL does not prevent it

All four assertions are negative ("must NOT appear"). The CONTROL
(`prune.rs:299-311`) calls `prune_orphaned_worktrees` **directly** with an
explicit `&fx.repos_root`, so it proves the *fixture* is reclaimable — it never
proves the *route* looked at the fixture. I verified this empirically: pointing
`WORKSPACE_ROOT_ENV` at an unrelated empty temp dir instead of `fx.repos_root`
leaves the test **green**:

```
running 1 test
test daemon::managed_routes::prune::tests::prune_spares_a_stopped_records_workspace ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3900 filtered out
```

So the doc comment's claim that "Membership in `active_workspace_paths` is
therefore the ONLY thing under test here" does not hold.

The fix that closes both items at once: add a **second, unregistered** worktree
to the fixture and assert **positively** that it DOES appear in the route's
dry-run `paths`. That proves the sweep actually scanned the fixture root in the
same run as the negative assertions, and it bounds the blast radius. Please
apply the same positive-control shape at site 2.

### 3. Assert on the preview *before* issuing the deleting run

Right now both route calls execute (`prune.rs:353-376`) and only then is
anything asserted (`prune.rs:395`). A test that deletes should fail fast on the
dry-run first.

---

Everything else on commit `09d05f77` verified clean so far — the mutation
genuinely goes red, the diff is comment-only, and the test does drive the real
route. Full findings and the APPROVE/WARN/BLOCK verdict follow once the second
commit lands and I have reviewed both sites together.

bobmatnyc:
## Verdict: BLOCK

Reviewed at `1b8763ac` (both commits, both sites, plus the third read the second
commit discloses). Everything below is from my own runs in a clean worktree at
this head — no pasted output was taken on trust.

The BLOCK rests on **one** finding, and it is not about the production code or
the three-read analysis (both of which hold up). It is that the site-1 **test**
performs a real, deleting worktree sweep whose target directory is resolved from
a process-global env var that a sibling test in the same binary mutates under a
different, non-interoperating lock. I also measured that this destructive half of
the test contributes **zero** mutation coverage — so it is all blast radius and
no value, and deleting it is the whole fix.

The engineer's corrected analysis is right on every point I could test, and the
correction is a real improvement over the original framing. Credit where due: the
Phase 2 read is a genuine find, and disclosing that the test pins the *pair*
rather than either read is the honest way to write it up.

---

## Verification of the claims

### The isolation matrix — independently reproduced

I mutated each read alone and in combination and ran the suite myself.

| mutation | site-1 test | site-2 test | rest of lib suite |
|---|---|---|---|
| none (pristine) | ok | ok | ok |
| **M1** route set (`prune.rs:145`) | **FAILED** | ok | ok |
| **M2** reap set (`session_manager/prune.rs:1015`) | ok | **ok** | `reap_orphaned_worktrees_removes_orphan_preserves_live` **FAILED** |
| **M3** Phase 2 `fresh_active` (`session_manager/prune.rs:799`) | ok | ok | **all green — 3896 passed, 0 failed** |
| **M2 + M3** | ok | **FAILED** | — |

```
===== M2+M3 both =====
test session_manager::reap_orphaned_worktrees_tests::reap_spares_a_stopped_records_workspace ... FAILED
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 3900 filtered out

===== M3 (phase2) only — FULL LIB SUITE =====
test result: ok. 3896 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 32.58s
```

**Claim 1 — three-read analysis correct? YES.** The Phase 2 re-read exists at
`session_manager/prune.rs:799-887`, reads `self.store` unfiltered, and the spare
check fires at `:892` immediately before `remove_session_worktree` at `:911`. The
green/green/red isolation is exactly as documented.

**Claim 2 — corrected severity accurate? YES.** I neutralized the site-1 test's
two dry-run assertions so the real-run assertions could execute, then applied M1
alone. `!removed.contains(&wt_str)` and `wt.exists()` both **passed**:

```
===== M1 (route) alone + preview asserts NEUTRALIZED =====
test daemon::managed_routes::prune::tests::prune_spares_a_stopped_records_workspace ... ok
```

So narrowing site 1 alone yields a false dry-run preview and nothing worse. The
corrected framing is right and the original overstatement is properly retracted.

**Claim 3 — tests match their documented scope? YES.** `reap_orphaned_worktrees_tests.rs:132-142`
states plainly that the test "stays GREEN under either single narrowing and goes
RED when BOTH are narrowed … it is not a tripwire on either read in isolation."
My M2-alone run confirms it: the new site-2 test is green, and it is the
*pre-existing* `reap_orphaned_worktrees_removes_orphan_preserves_live` that
catches M2. The doc does not overclaim.

**Claim 4 — trap avoidance at site 2? YES, verified independently.**
`resolve_ownerless_with_grace` has exactly **one** production caller —
`session_manager/prune.rs:755`, inside `prune_orphaned_worktrees` — which both
the manual route and the automatic reaper funnel through, so the ownership-gate
semantics are identical by construction, not by coincidence.
`ManagedSessionState::is_terminal` (`record.rs:138-140`) is
`Decommissioned | Deleted` only, so `Stopped` resolves live/resumable and a
self-owned sentinel would indeed be spared at gate 4. Both fixtures call
`stamp_reclaimable_sentinel`, which stamps a fresh `ManagedSessionId::new()` —
never registered. Both carry a CONTROL against an empty active set.

**Claim 5 — `reap` shape? YES.** `reap_orphaned_worktrees` (`:978-1023`) takes
`repos_root` as a parameter and calls `prune_orphaned_worktrees(.., false, Skip)`
— `dry_run: false` hardcoded, no preview. The site-2 test consequently needs no
env mutation, no `env_test_lock`, and no clippy allow. This is the right shape.

**Claim 6 — commenting the Phase 2 read? Right call.** Leaving one half of a
declared defense-in-depth pair uncommented would have been the half-fix. See
MEDIUM-2 for one inaccuracy in it.

**Original claims 1-4 from the brief, restated against this head:** red-under-mutation
**YES** (my own output below); comment-only production diff **YES** (the single
`-` line in the whole diff is a `Test:` doc pointer); test drives the real route
**YES** (`prune.rs:361-381` calls `prune_worktrees_route` directly); full suite
**YES** — I measured `3894 passed; 0 failed; 6 ignored` with the new test filtered
out, i.e. exactly the reported 3895/0/6 with it. Site 1 was genuinely unpinned
before this PR: with M1 applied and the new test skipped, the whole workspace
suite including `prune_route_dry_run_reports` and all of `orphan_gc_sweep.rs`
stayed green.

My own red-under-mutation output at site 1:

```
thread '...::prune_spares_a_stopped_records_workspace' panicked at
crates/trusty-mpm/src/daemon/managed_routes/prune.rs:397:9:
a Stopped record's workspace must NOT be reclaimable;
/private/var/.../repos/owner/repo/.worktrees/stopped-but-live appeared in the
dry-run set ["/private/var/.../repos/owner/repo/.worktrees/stopped-but-live"]
```

---

## Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/daemon/managed_routes/prune.rs` | 371-381 | Test runs a real deleting sweep against an env-resolved root; not `#[serial]`; contributes no mutation coverage | Delete the `dry_run: false` call and its two assertions |
| HIGH | `crates/trusty-mpm/src/daemon/managed_routes/prune.rs` | 401-421 | Test goes silently vacuous if the route stops seeing the fixture root; CONTROL cannot catch it | Assert positively that an unregistered worktree DOES appear in the route's dry-run `paths` |
| MEDIUM | `crates/trusty-mpm/src/session_manager/prune.rs` | 799-887 | Phase 2 read is the only one of the three with no isolation pin — entire suite green under M3 alone | Add the validated test below |
| MEDIUM | `crates/trusty-mpm/src/session_manager/prune.rs` | 814-815 | "LAST thing standing" is inaccurate — the pre-removal dirty gate at `:903` follows | Name gate 6 |
| MEDIUM | `crates/trusty-mpm/src/session_manager/prune.rs` | 713, 145, 1015 | `active_workspace_paths` is itself the invitation to filter on `Active` | Rename to `spared_workspace_paths` |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/prune.rs` | 107 | `Test:` pointer names `prune_worktrees_route_dry_run`, which exists nowhere in the workspace | Fix or drop the pointer |

### CRITICAL — the site-1 test can delete the operator's real worktrees

`prune.rs:371-381` issues `PruneWorktreesRequest { dry_run: false, .. }`. The
directory that sweep targets is **not** passed in — the route resolves it inside
itself (`prune.rs:151-152` → `trusty_tools_config::workspace_root`). The test's
only control over it is a `set_var` at `prune.rs:352-357` held under
`env_test_lock`.

`connectors::tm_tests::create_session_full_lifecycle` mutates the **same** env
var (`tm_tests.rs:285-288`, restored at `:408-412`) in the **same** lib test
binary, serialised by `#[serial_test::serial]` — which does not interoperate with
`env_test_lock`. The new test carries neither `#[serial]` nor anything else that
excludes it, and `serial_test` 3.4 leaves an untagged test free to run alongside a
`#[serial]` one. This crate's own `Cargo.toml:331-333` names `serial_test` as the
convention for exactly this: *"run env-mutating tests serially to prevent env-var
races across parallel Rust test threads."*

If that restore lands inside this test's window, `workspace_root()` falls through
to the operator's real `~/trusty-mpm-projects`. The three-read defense-in-depth
does **not** help here, and this is the part worth being explicit about: both the
Phase 1 caller set and the Phase 2 `fresh_active` re-read draw from the *same*
freshly-isolated, empty `DaemonState` store, so real worktrees are absent from
**both**. Every real sentinel resolves owner-not-found → past grace → provably
ownerless. `workspace_guard::is_safe_to_remove` is not on this path. Clean,
fully-pushed worktrees are removed, and `remove_session_worktree` follows with
`git branch -D`.

I did not demonstrate this one live, deliberately — the demonstration is the
incident. The mechanism is established by reading both files, and I did verify
the half that is safe to verify: the route silently follows whatever the env
resolves to. Pointing `WORKSPACE_ROOT_ENV` at an unrelated empty temp dir instead
of `fx.repos_root` leaves the test **green**:

```
test daemon::managed_routes::prune::tests::prune_spares_a_stopped_records_workspace ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3900 filtered out
```

**What makes this cheap to fix:** my M1-with-neutralized-preview-asserts run
proves the destructive half of the test carries **no** mutation-detection value.
The test's red comes entirely from the dry-run assertion at `:401`. The
`dry_run: false` call, its two assertions, and — with it — the `env_test_lock`
hold across awaits and the `#[allow(clippy::await_holding_lock)]` are all
removable at zero cost to what the test actually pins. That is the fix I want:
delete them. If there is a reason to keep the real run, it needs
`#[serial_test::serial]` **and** the HIGH's positive assertion first.

### HIGH — the site-1 test can go silently vacuous

All four assertions are negative. The CONTROL (`prune.rs:305-317`) calls
`prune_orphaned_worktrees` directly with an explicit `&fx.repos_root`, so it
proves the *fixture* is reclaimable — never that the *route* looked at the
fixture. The vacuity probe above is the proof. Note this is the same class of gap
as the CRITICAL: both stem from the route's target being invisible to the test's
assertions. One positive assertion — a second, unregistered fixture worktree that
must appear in the route's dry-run `paths` — closes both.

### MEDIUM — the Phase 2 read has no isolation pin

By the comment's own reckoning this is "the LAST thing standing between a
reclaimable candidate and `remove_session_worktree`" and "the one whose loss is
least visible". After this PR it is the only one of the three reads that nothing
catches in isolation. The comment is honest about this, so it is a gap rather
than a false claim — but it is the wrong read to leave uncovered.

I wrote and validated the fix rather than just naming it. Green on pristine, red
under M3:

```
===== P0: probe on PRISTINE =====
test ...::phase2_fresh_active_spares_a_stopped_record_with_empty_caller_set ... ok

===== P1: probe under M3 =====
test ...::phase2_fresh_active_spares_a_stopped_record_with_empty_caller_set ... FAILED
panicked at reap_orphaned_worktrees_tests.rs:285:5:
Phase 2 fresh_active must spare a Stopped record's workspace even with an empty
caller-supplied active set; removed=[".../.worktrees/stopped-but-live"]
```

The shape is the existing site-2 test with one change: pass an **empty**
caller-side active set to `prune_orphaned_worktrees(&fx.repos_root, &[], false,
Skip)` *after* persisting the Stopped record. The candidate then genuinely enters
the list, so only Phase 2 can spare it — which isolates that read exactly. It
needs no env, no new fixture helper, and about 35 lines. Happy to hand over the
exact body.

### MEDIUM — "LAST thing standing" repeats the error just corrected

`session_manager/prune.rs:814-815`. The `fresh_active` check at `:892` is not the
last gate: `dirt_blocks_removal(&candidate, policy, "pre-removal")` runs at `:903`
before the removal at `:911`, and on the reap path the policy is always `Skip`, so
it is not bypassable there. This is the same overstatement the second commit
correctly removed from the site-1 comment, reintroduced one file over. Suggest:
"the last gate before the #4118 pre-removal dirty re-check (`:903`), and the only
one that protects a *clean* live worktree."

### MEDIUM — the name is the thing inviting the filter

`active_workspace_paths` is the local at `prune.rs:145`, the local at
`session_manager/prune.rs:1015`, and — the one that matters — the **parameter
name** of the shared `prune_orphaned_worktrees` at `session_manager/prune.rs:713`.
A parameter called "active" is what makes `state == Active` look like a tidy-up
rather than a bug. Three comments do not remove the invitation at the signature.
A rename to `spared_workspace_paths` is behavior-preserving and addresses the
cause rather than posting warnings around it.

---

## On the shared-helper question

Asked to say so on the record if I thought two hand-maintained pins was the worse
outcome. My answer is nuanced, and it changed once the third read surfaced.

The reflexive objection to a helper — that merging the reads destroys the
redundancy the comment says is load-bearing — **does not apply**. The route set
and the reap set are alternative *entry points*; no sweep uses both. The depth is
(caller set) × (Phase 2 re-read), and merging the two caller-side reads leaves
that product untouched. A `SessionManager::workspace_paths_for_sparing()` shared
by both callers would be behavior-preserving and would collapse two
hand-maintained comments into one.

So I do not think the helper is wrong. But I also do not think it is this PR's
job, and I would not WARN on its absence — the comments are now accurate and
mutually consistent, which was the actual risk. What I would not do is defer it
*and* leave the parameter named `active_workspace_paths`. If only one of the two
lands, take the rename: it is smaller and it addresses the cause.

---

1. **Delete the `dry_run: false` route call at `prune.rs:371-381` and its two
   assertions** (`:415-421`), along with the now-unnecessary `env_test_lock` hold
   across awaits and `#[allow(clippy::await_holding_lock)]` at `:286`. Measured to
   cost nothing in mutation coverage. If it stays instead, add
   `#[serial_test::serial]` **and** item 2.
2. **Add a positive assertion to the site-1 test** that an unregistered fixture
   worktree DOES appear in the route's dry-run `paths`, so the test cannot pass
   while the route is scanning the wrong root.
3. **Add the Phase 2 isolation test** (validated above) so all three reads have a
   single-read tripwire.
4. **Correct `session_manager/prune.rs:814-815`** to name the `:903` pre-removal
   dirty gate.
5. **Rename `active_workspace_paths` → `spared_workspace_paths`** at
   `session_manager/prune.rs:713` and both call sites.
6. **Fix the dangling `Test:` pointer** at `prune.rs:107`.

Items 1-3 are the ones I would re-review on. 4-6 are cheap and can ride along.

## Notes

- `#[allow(clippy::await_holding_lock)]` was fine on its own terms: the rationale
  is correct (`#[tokio::test]` is current-thread, each test on its own libtest
  thread, so no executor deadlock), and the precedent the brief asked me to check
  does exist — `trusty-agents/src/test_env.rs:456-460`,
  `init/tests/mod.rs:11`, `tools/pm_bridge_backend_tests.rs:11`. It only
  disappears here as a side effect of removing the destructive call.
  Separately: `trusty-memory` solves the same problem with an async lock
  (`attribution.rs:604`), which avoids the lint outright — a cleaner long-term
  shape for `env_test_lock`, out of scope here.
- `env_test_lock` recovers from poisoning
  (`trusty_tools_config.rs:645`, `unwrap_or_else(|e| e.into_inner())`), so a
  panicking assertion under the guard does not cascade. Checked, not a finding.
- Restoring via `remove_var` rather than the prior value matches this crate's
  existing convention. Not a finding.
- One unrelated flake surfaced under load in my first full run —
  `telegram::supervisor::tests::healthy_run_resets_transient_backoff`, a
  millisecond-scale backoff assertion. It passed on re-run. Not this PR's, but it
  is a load-sensitive timing assertion in the shared gate and worth its own issue.
- Three integration failures in one of my runs (`local_spawn`, `tm_hook_pm_guard`)
  were artifacts of my own harness environment, not the branch.
- The production diff remains genuinely comment-only across both commits: the only
  `-` lines in the whole PR are doc-comment text.

bobmatnyc:
## Verdict: APPROVE

Re-reviewed at `ec232325` in a clean worktree. **The CRITICAL is fixed and I
verified it independently.** Zero CRITICAL, zero HIGH remaining → APPROVE. Four
findings below are MEDIUM/LOW and none of them block; two of them are things I
would genuinely like fixed, and one contradicts a scope judgment reported with
this round.

I re-ran every mutation rather than reading the pasted matrix. It reproduces
**exactly**, on all five reported rows.

---

## The BLOCK is cleared

`prune_worktrees_route(` now appears exactly once in the test body
(`prune.rs:373`) with `dry_run: true` (`:376`). The only `dry_run: false` left in
the file is inside a comment at `:133`. `let real` and the `wt.exists()`
assertion are gone. No destructive sweep remains anywhere in the test surface
that resolves its target from the environment. Confirmed by grep, not by
description.

### On keeping `env_test_lock` + `#[allow(await_holding_lock)]` — you were right, my brief's expectation was wrong

I said removing the destructive half would make both unnecessary. That was
wrong, and the reasoning given for keeping them is correct. Adjudicating it
properly:

**Both guards are genuinely required**, because the tests that mutate
`TRUSTY_MPM_WORKSPACE_ROOT` in this binary fall into two *disjoint* groups:

- guarded by `env_test_lock` **only** — `core::trusty_tools_config`'s
  env-precedence tests (`:693`, `:712-714`, `:724`, `:1054`) and
  `daemon::managed_routes::inproject::tests` (`:27-33`), all of which
  `remove_var` that variable;
- guarded by `#[serial]` **only** — `connectors::tm_tests::create_session_full_lifecycle`
  (`tm_tests.rs:262`, `:285-288`, `:408-412`).

Neither guard alone excludes both groups. And the lock still earns its keep even
without the deleting sweep: the dry-run call reads the env *inside* the route, so
a clobber does not merely mis-target — it makes every assertion vacuously true.
That is the exact failure I demonstrated last round.

**No hazard is reintroduced.** I checked for the obvious one — lock-order
inversion between two global locks. It is not constructible: `#[serial_test::serial]`
acquires in the macro wrapper *before* the body runs, and `env_test_lock()` is
only ever acquired *inside* a body, so the order is always serial → env_test_lock
and never the reverse. Poisoning is handled on both sides (`trusty_tools_config.rs:645`
recovers via `into_inner`). The only cost is extra serialisation. The
`#[allow]` remains correctly justified — a `std::sync::MutexGuard` really is held
across the `.await`, and the current-thread-per-test rationale holds.

---

## Mutation matrix — reproduced, plus one row you did not run

Every row you reported matches mine. I added **M4** for a caller-side read that
was not in the matrix.

| mutation | site-1 | site-2 | phase 2 | pre-existing reap | suite |
|---|---|---|---|---|---|
| pristine | ok | ok | ok | ok | **3897 passed, 0 failed, 6 ignored** |
| M1 route set | **FAILED** | ok | ok | ok | red |
| M2 reap set | ok | ok | ok | **FAILED** | red |
| M3 phase 2 | ok | ok | **FAILED** | ok | red |
| M2+M3 | ok | **FAILED** | **FAILED** | **FAILED** | red |
| **M4 `mcp_context` pause path** | ok | ok | ok | ok | **all green — 3897 passed, 0 failed** |

M3 is genuinely caught now — it was fully green suite-wide before, and I
confirmed that myself last round. That is the substantive win of this revision.

Gates I ran myself at this head: `cargo fmt --check` exit **0**; `cargo clippy -p
trusty-mpm --all-targets -- -D warnings` exit **0**; full suite without `--lib` —
lib binary `3897 passed; 0 failed; 6 ignored`, every integration binary green.
(Two `tm_hook_pm_guard` failures are my own environment: this worktree lives
under `/private/tmp`, which that hook's path guard rejects by design. The PR
touches no hook code. Last round's `local_spawn` failure was likewise mine — it
disappeared once I stopped pinning `TRUSTY_MPM_WORKSPACE_ROOT` process-wide,
which isolates it cleanly.)

---

## Item 2 — the pre-existing hole is real, and it is worse than reported

Confirmed, and worth stating plainly because it is the most instructive thing in
this PR.

`prune_orphaned_worktrees_store_snapshot_blocks_deletion`
(`prune_orphan_tests.rs` on `origin/main`) does not merely fail to reach Phase 2.
Its **doc comment asserts that it does**:

> *"async test that genuinely exercises the Phase 2 fresh-store snapshot path…
> Phase 1 finds the worktree as a candidate (empty initial snapshot), then Phase 2
> reads the live store and finds the matching record — skipping deletion."*

while its own **body comment**, added later, states the opposite:

> *"#3649: the dir has no ownership sentinel, so it is classified owner-unknown
> and skipped BEFORE the Phase 2 fresh-active check ever runs."*

So the test's assertion (`outcome.removed.is_empty()`) has been passing for the
owner-unknown reason, not the Phase 2 reason, since #3649 landed. The doc was
true when written for #1845; #3649 inserted a gate *ahead* of the code under
test and silently hollowed it. Someone noticed enough to write the body comment
and still left the doc claim and the `Test:` pointer standing.

That is this repo's recurring shape: **a named guard whose evidence stopped
covering it, with the contradiction recorded in the same file and nobody
reconciling the two.** It is exactly why M3 was invisible. Calling it out here as
requested — this is worth a standing lesson, not just a fix.

The replacement is sound. `phase2_fresh_snapshot_spares_a_record_the_caller_set_missed`
(`prune_orphan_tests.rs:1254`) uses `reclaimable_fixture` (which stamps the aged
never-registered sentinel at `:819`) plus an empty caller set, and its control —
the dry-run of the *same* sweep with the *same* record present, which returns
before Phase 2 and still lists the worktree — is a **better** non-vacuity control
than the one I proposed. The control-vs-real delta is precisely the Phase 2
protection and cannot be faked by a fixture that was never reclaimable.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/daemon/mcp_context.rs` | 184-187 | Fourth caller-side unfiltered read, feeding `prune_orphaned_worktrees(.., dry_run=false, ..)`; no comment, no pin — M4 leaves the whole suite green | Add the same `#4288` comment; extend a pin to cover it |
| MEDIUM | `crates/trusty-mpm/src/session_manager/search_gc.rs` | 251-265 | `active_workspace_path_set` is the same unfiltered construction, deletion-adjacent, still named `active_*`, uncommented — in a file this PR edited | Rename + comment; blast radius is bounded (see below) |
| MEDIUM | `crates/trusty-mpm/src/session_manager/prune.rs` | 814 | "LAST thing standing" is still inaccurate — the pre-removal dirty gate at `:903` follows; now propagated to `prune_orphan_tests.rs:1231` | Qualify: last *before* the #4118 re-check, and only truly last under `ForceDiscard` |
| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/prune.rs` | 401-421 | Site-1 test is still all-negative assertions; can go vacuous if root resolution changes (downgraded from HIGH — blast radius gone) | One positive assertion: an unregistered worktree DOES appear in the dry-run `paths` |
| LOW | `crates/trusty-mpm/src/session_manager/prune.rs` | 1020 | Cites `prune_orphaned_worktrees_removes_orphan_preserves_live`, which exists nowhere — correct name is `reap_…` | One-word fix |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/prune.rs` | 107 | `Test:` still names `prune_worktrees_route_dry_run`, which exists nowhere (flagged last round) | Fix or drop |

### MEDIUM-1 — the rename scope judgment is wrong on `mcp_context.rs`

Asked to confirm that `doctor_hooks_hygiene.rs` and `mcp_context.rs` were
correctly left alone as "different functions". **`doctor_hooks_hygiene.rs` and
`daemon/doctor.rs` — yes, correct.** Those are read-only diagnostics; nothing
there deletes, and the `active_` naming is accurate for a liveness *report*.

**`mcp_context.rs` — no.** It is not a different function. `mcp_context.rs:184-187`
is:

```rust
let active_workspace_paths: Vec<PathBuf> = records
    .iter()
    .filter_map(|r| r.workspace_path.clone())
    .collect();
```

— character-for-character the route's construction, feeding the *same*
`prune_orphaned_worktrees` at `:196-201` with **`dry_run: false`** hardcoded, on
the `/tm-session-pause` path a user triggers directly. It is a third caller-side
read of exactly the kind this PR exists to label, it kept the `active_` name the
rename was meant to eliminate, it has no `#4288` comment, and M4 shows nothing in
the suite catches narrowing it.

To be precise about severity, because it would be easy to overstate: this is
**not** a data-loss finding. Narrowing M4 alone leaves Phase 2 to spare the
worktree, and M4+M3 would be caught by the new Phase 2 test. There is no
undetectable path to deletion. It is a coverage-and-consistency gap — the same
one that was just closed at three other sites, left open at the fourth.

### MEDIUM-2 — `search_gc.rs`

`active_workspace_path_set` (`:251-265`) is the same unfiltered
`list().filter_map(workspace_path)` shape and its own doc says it "mirrors
`prune::prune_orphaned_worktrees`'s `fresh_in_use` construction" — a line this PR
edited to track the rename while leaving the function and its local
(`:301`) named `active_*`, uncommented and unpinned. It decides deletion of
trusty-search indexes.

Bounding it honestly: `is_orphan_index` (`:197-214`) also requires
`chunk_count == 0 || !root_path.exists()`, so a live Stopped session with a
populated index whose root still exists is spared regardless. The realistic loss
from narrowing it is a 0-chunk index for a live-but-Stopped session — minor and
self-healing on reindex. MEDIUM for consistency, not for impact.

### MEDIUM-3 — "LAST thing standing"

I flagged this last round; it is unchanged at `prune.rs:814` and has now been
copied into the new test's doc at `prune_orphan_tests.rs:1231`. The gate order at
`:892` → `:903` → `:911` is unambiguous: `fresh_in_use` fires, then
`dirt_blocks_removal(&candidate, policy, "pre-removal")`, then removal. On the
reap path the policy is always `Skip`, so that dirty gate is real and
non-bypassable there. The statement is only true under `ForceDiscard`. Since the
whole point of these comments is that a future reader trusts them, worth
qualifying in both places.

---

## The two open items

**1. Rebase — required, but not for conflict reasons.** Branch protection on
`main` has `required_status_checks.strict = true`, i.e. "require branches to be up
to date before merging" is on. So the branch must be updated regardless of
mergeability.

Conflict risk is **zero**, and I checked rather than assumed: merge-base is
`6ffe001d`; main has since taken `d492fd34`, `61329940`, `54bba9f7`, `6ef47ecb`,
touching only `crates/trusty-agents`, `crates/trusty-git-analytics`, `docs/`, and
`Cargo.lock`. Intersecting that file list with the PR's six files yields the empty
set. A plain rebase or merge will be clean; re-run CI after, since strict mode
means the required checks must pass on the updated tree.

**2. The `classify_connect_error_is_transport` failure — load-induced, not
provoked by these changes.** It is structural, and the fixture shows why:

```rust
fn dead_loopback_url() -> String {
    let listener = std::net::TcpListener::bind("127.0.0.1:0")...;
    let port = listener.local_addr()...port();
    drop(listener);
    format!("http://127.0.0.1:{port}")
}
```

Bind-ephemeral-then-drop. Between the `drop` and the request, any other binder in
the process — and this suite spawns real daemons and HTTP servers — can be handed
that freed port by the OS. If that happens the connect *succeeds*, the client
returns a status/body error rather than a `reqwest` connect error, and
`classify_activity_error` yields `NonTransport`, failing the assertion in exactly
the observed way. Probability scales with parallel port churn, which is why it
only appeared in your most heavily loaded run. The PR touches no TUI, client, or
HTTP code — its six files are the two prune modules, two prune test modules,
`search_gc.rs`, and the CHANGELOG. It did not pass through clean in my runs by
luck alone: my full matrix (six suite runs) and my final gate run were all clean.

This is the **second** distinct load-sensitive test to redden this gate during
this review — the first was `telegram::supervisor::tests::healthy_run_resets_transient_backoff`,
a millisecond-scale backoff assertion. Two unrelated tests failing only under
load is the "load-induced red degrades the gate" pattern, and the fix is to
harden the fixtures (have `dead_loopback_url` hold the listener and connect to a
port it keeps reserved-but-refusing, or use a documented discard port), not to
re-run until green. Worth its own issue; explicitly **not** a reason to hold this
PR.

---

## Notes

- Nothing here requires re-review. The four MEDIUMs are all cheap; MEDIUM-1 and
  MEDIUM-3 are the two I would actually like to see, and either this PR or a
  follow-up is fine. If a follow-up, it should also carry the two LOW dangling
  test pointers.
- The rename is applied consistently across `prune.rs`, `session_manager/prune.rs`,
  `prune_orphan_tests.rs`, `reap_orphaned_worktrees_tests.rs`, and `search_gc.rs`'s
  doc references. `in_use_` is a better choice than my suggested `spared_` — it
  describes the input rather than the outcome, which is the more durable framing.
- The site-2 accuracy correction (`prune.rs:1017-1023`) is substantively accurate:
  I independently confirmed `create_with_id` persists `Provisioning`
  (`create.rs:279`) and that the literal `== Active` mutation at that line is
  caught by the pre-existing reap test, not the new one. Only the cited test name
  is wrong (LOW above).
- Sequencing note: my earlier advisory comment on this PR predates `1b8763ac` and
  its item 1 is now resolved; items 2 and 3 of it survive as MEDIUM-4 and are
  superseded by this verdict.

Issue #3455, opened 2026-07-20
feat(trusty-mpm): per-project setting to disable session worktree provisioning


bobmatnyc:
## Problem

Some projects are meant to be worked on directly in `main` rather than in isolated worktrees. The owner's writing project (~/trusty-mpm-projects/bobmatnyc/writing) is one example — it has a direct-main workflow with no branch, no PR flow, and no isolation requirement.

Currently, `tm` assumes worktree isolation is universal: launching a session always provisions `.worktrees/<session-name>` on a `session/<name>` branch, even for projects where this is exactly wrong.

### Observed consequences (all silent — no error reported):
- A worktree sits unused at a stale commit, wasting disk space
- The session's cwd points to the worktree instead of the primary checkout
- Agents in that session must `cd` to the primary checkout on every single command — permanent friction
- Misdiagnoses can occur when an agent mistakes the worktree's stale state for the project's actual state

## Desired behavior

A per-project setting (e.g. `worktree: false` or `work_on_main: true`) such that a session for that project:
- does NOT create `.worktrees/<session-id>` or a `session/<name>` branch
- sets the session's working directory to the primary checkout
- otherwise behaves as a normal tm-managed session (output style, hooks, agents, skills all unaffected)

## Design question: where should the setting live?

This is intentionally flagged as an open design question, not a settled answer.

The shared project registry (#1517) is the natural home for per-project facts. That epic (WI-2) includes extracting `Project` record from trusty-mpm into trusty-common with a `canonical_root` field for managed checkouts.

**Two placement options:**
1. **In the registry row** (`trusty-common::Project`): Add a `worktree_enabled: bool` field. Session launch reads it from the registry and skips worktree provisioning if false.
2. **In session config** (trusty-mpm only): Add a session-level config key. Still requires cross-referencing to the project to read the setting.

Option 1 is cleaner (single source of truth per project) but couples the registry to session provisioning policy. Option 2 is narrower but requires session config plumbing. Let the implementer decide; this issue surfaces the gap and the owner's requirement.

## Discoverability

A project where tm silently provisions an unwanted worktree has no signal that a setting exists to turn it off. Consider surfacing the setting in project-selection UX, `tm project info`, or help docs.

## Related issues

- **#1517** (EPIC: Multi-project registry) — WI-2 (project registry extraction) is the natural home for this setting
- **#3174** (project-first session flow) — related UX for project selection
- **#2919** (automatic worktree reclamation) — complementary disk-management issue
[...]

PR #4303, opened 2026-07-29
fix(trusty-mpm): honour the per-project worktree opt-out in both CLI paths

### LINKED ISSUE #4300: tm launch and the daemon-unreachable fallback bypass the per-project worktree opt-out
Verified against `origin/main` @ `6ef47ecb`.

### Defect
PR #3781 (issue #3455, merged 2026-07-24) added a per-project worktree opt-out: `Project::worktree: Option<bool>` (`crates/trusty-mpm/src/project/record.rs:165`), resolved by `worktree_enabled()` (`:176`, `unwrap_or(true)`), consulted by `worktree_enabled_for_origin()` (`crates/trusty-mpm/src/daemon/managed_routes/launch_on_main.rs:50`) at the single decision point `daemon/managed_routes/lifecycle.rs:373-377` — deliberately before `try_inproject_spawn` (`:418`) so no base clone is created either.

**Two CLI paths create worktrees without consulting the registry at all:**
- `crates/trusty-mpm/src/bin/tm/commands/launch.rs:184` `ensure_base_clone` → `:197` `create_session_worktree`. `tm launch` runs this **in-process**, never asking the daemon. No `worktree_enabled_for_origin` call anywhere in the file.
- `crates/trusty-mpm/src/bin/tm/commands/guided.rs:1223` `ensure_base_clone` → `:1241` `create_session_worktree`, inside `redirect_to_managed_clone` (`:1195`), reached from `fallback_protected` (`:1104`) — the **daemon-unreachable** fallback for bare `tm` (`:1080-1093`).

Neither can call the check as written: `worktree_enabled_for_origin` is `pub(super)` (`launch_on_main.rs:50`), scoped to `daemon::managed_routes`.

**Consequence:** a project with `worktree: false` still gets a worktree from `tm launch`, and from bare `tm` whenever the daemon is down. The setting is silently conditional on the daemon being up.

### Evidence
- The flag resolves correctly when the daemon path IS used — `repo_url_matches` (`record.rs:199-209`) normalises SSH vs HTTPS through `parse_github_path`, so `git@github.com:bobmatnyc/writing.git` matches registry `https://github.com/bobmatnyc/writing`.
- Both CLI paths keep UUID worktree names (`guided.rs:1234-1239`, `launch.rs:194-196`); only the daemon's `spawn_managed_inproject` resolves semantic names — so the two paths are independently reachable and distinguishable in the wild.
- Adoption is now nonzero: 2 of 33 entries in `~/.trusty-mpm/project-registry/projects.json` carry `worktree: false` (`writing` → `https://github.com/bobmatnyc/writing`, `cto` → `https://github.com/bob-duetto/cto`); the other 31 have no `worktree` key. The bypass is live for those two projects, which is why it has only just become reachable.

### Resolution
Relax `worktree_enabled_for_origin`'s visibility so the CLI process can call it, and consult it at `launch.rs:184` and `guided.rs:1223` **before** `ensure_base_clone` — matching the daemon's ordering, so an opted-out project creates neither a worktree nor a base clone. The CLI process must read the project registry directly; note whether that is already available in that context.

Behavior on opt-out in these paths should match `spawn_managed_on_main()` (`launch_on_main.rs:117`): no clone, no worktree, no `session/<name>` branch, `workspace_owned = false`.

**Concurrency stays WARN, not refuse** — owner's decision, consistent with #3781 (`launch_on_main.rs:94-101`). Do not change that behavior in this fix.

### Acceptance
A test per path proving an opted-out project gets NO worktree and NO base clone from (a) `tm launch` and (b) the daemon-unreachable bare-`tm` fallback, plus a test that an opted-IN (or unset) project is unaffected — the `unwrap_or(true)` default must not regress.

Anti-hollow requirements this repo enforces: no `#[ignore]`, no env-var gate, no early `return`, no skip-when-missing; assertions on specific paths and on the absence of the worktree directory, not `is_ok()`; must not be excluded by `--lib`; satisfied only by pasting the `... ok` lines from `cargo test -p trusty-mpm` run without `--lib`. The test must go RED with the fix reverted.

bobmatnyc:
## The defect

PR #3781 (issue #3455) added a per-project worktree opt-out (`Project.worktree: Option<bool>`) and consulted it at exactly **one** decision point — inside the daemon's spawn route, deliberately ahead of `try_inproject_spawn` so an opted-out project gets neither a worktree nor a base clone.

Two `tm` entry points provision worktrees **entirely in the CLI process** and never reached that check:

- `bin/tm/commands/launch.rs` — `tm launch` runs `ensure_base_clone` → `create_session_worktree` in-process and never asks the daemon.
- `bin/tm/commands/guided.rs` — `redirect_to_managed_clone`, reached from `fallback_protected`, is the **daemon-unreachable** fallback for bare `tm`; it runs precisely *because* the daemon is down.

Result: `worktree: false` was silently conditional on the daemon being up. Live today for the two registry entries that carry it — `bobmatnyc/writing` and `bob-duetto/cto`.

## The seam I chose, and why not `pub(super)`

The issue suggested relaxing `worktree_enabled_for_origin`'s `pub(super)` visibility. I did not; I **moved the decision** to a new `crate::project::worktree_policy`, and the daemon now calls it too.

`launch_on_main.rs:9`'s module doc states the `pub(super)` scoping is deliberate — "the functions are `pub(super)` so `lifecycle::spawn_managed_routed` (the only caller) can route to them". That premise is exactly what #4300 falsifies: there is now a second and third caller, in a different process. Given that, two options:

1. Widen `daemon::managed_routes::launch_on_main::worktree_enabled_for_origin` to `pub`. This leaves the `tm` CLI reaching into `daemon::` to answer *"what does this project record say?"* — a registry query with no daemon in it — and leaves the module doc's stated rationale false.
2. Put the decision next to the record it reads.

I took (2). `Project::worktree` is a property of the project record, so `project::worktree_policy` owns:

- `worktree_enabled_in(&[Project], origin)` — the pure matching core, so the two defaults that matter are directly assertable;
- `worktree_enabled_for_origin(&ProjectRegistry, origin)` — the in-process form the daemon already had;
- `worktree_enabled_for_origin_at(registry_dir, origin)` — the out-of-process form the CLI needs (`ProjectStore`'s module doc already declares multi-process readers a supported pattern, so this is the front door, not a workaround);
- `registry_data_dir_under(framework_root)` / `registry_data_dir()` — the directory name in **one** place. Previously `DaemonState::project_registry` hardcoded `"project-registry"`; a CLI-side copy that drifted would read an empty registry and answer `true` for every project, re-opening this bug with nothing failing. `DaemonState` now calls the shared helper.

`launch_on_main` keeps its `pub(super)` scoping for everything that genuinely is daemon spawn surface (`has_concurrent_main_checkout_session`, `spawn_managed_on_main`); its module doc is updated to record why the registry lookup left.

On the CLI side both call sites route through one new `commands::managed_workspace`, whose `provision` reads the opt-out **before** `ensure_base_clone` and returns a two-variant `ManagedWorkspace { Worktree, MainCheckout }` so a caller cannot confuse "the disposable worktree" with "the operator's own checkout". Each entry point reads the registry exactly once and passes the answer down, so the operator-facing notice can never describe an action other than the one taken.

## Behaviour

- Opted out → no clone, no `session/<name>` branch, no worktree; the session runs in the repo root, matching `spawn_managed_on_main`. `workspace_owned` is unaffected: `decommission_with_root` only removes a non-owned path when `is_session_worktree` recognises it as living under `.worktrees/`, which a main checkout never does.
- `unwrap_or(true)` preserved. Unset, unregistered, **and** an unreadable/malformed registry all resolve to worktrees ON — an unreadable registry must never strip isolation.
- Matching goes through `repo_url_matches` unchanged (SSH/HTTPS/`.git`/case).
- The guided fallback now targets the repo **root** rather than `cwd`, so a bare `tm` from a subdirectory of an opted-out project deploys to the project root, not the subdirectory.
- **Concurrency untouched.** #3455's WARN-never-refuse stands; nothing here refuses a launch.

### `tm launch` and the guided fallback, guard neutralised (`if false && !isolate`)

Both opted-out tests go RED — and the revert genuinely reproduced the bug: it cloned `bobmatnyc/writing` over SSH and created a worktree.

```
---- provision_for_launch_opted_out_creates_no_clone_and_no_worktree stdout ----
assertion `left == right` failed: #4300: an opted-out project must launch in its own checkout
  left: Worktree(".../bobmatnyc/writing/.worktrees/4a0f7f0e-685a-41a1-9284-b019f6398195")
 right: MainCheckout("/var/folders/.../.tmplIF1kB")

---- provision_for_fallback_opted_out_creates_no_clone_and_no_worktree stdout ----
assertion `left == right` failed: #4300: the daemon-unreachable fallback must launch in the repo root
  left: Worktree(".../bob-duetto/cto/.worktrees/72cbe05f-5d04-4a71-b44f-a954b317b41d")
 right: MainCheckout("/var/folders/.../.tmpzHHr6O")

test result: FAILED. 4 passed; 2 failed; 0 ignored
```

The four opted-in/unset controls stayed green, as they must — they do not depend on the guard.

### Ordering is load-bearing too (guard moved *after* `ensure_base_clone`)

```
---- provision_for_fallback_opted_out_creates_no_clone_and_no_worktree stdout ----
#4300: base clone dir must NOT be created for an opted-out project: /var/folders/.../.tmpygV26P/bob-duetto/cto
```

A check-after-clone still returns the right path — the base-clone assertion is what catches it. This is the ordering the daemon comment calls deliberate.

### Green, fix restored

```
test commands::managed_workspace::tests::provision_for_launch_opted_out_creates_no_clone_and_no_worktree ... ok
test commands::managed_workspace::tests::provision_for_launch_unset_creates_worktree ... ok
test commands::managed_workspace::tests::provision_for_launch_unregistered_origin_creates_worktree ... ok
test commands::managed_workspace::tests::provision_for_fallback_opted_out_creates_no_clone_and_no_worktree ... ok
test commands::managed_workspace::tests::provision_for_fallback_unset_creates_worktree_not_live_checkout ... ok
test commands::managed_workspace::tests::provision_for_fallback_other_projects_optout_does_not_leak ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 1260 filtered out

test project::worktree_policy::tests::worktree_enabled_in_defaults_true ... ok
test project::worktree_policy::tests::worktree_enabled_in_honors_false ... ok
test project::worktree_policy::tests::worktree_enabled_in_matches_ssh_form ... ok
test project::worktree_policy::tests::worktree_enabled_for_origin_defaults_true_when_unregistered ... ok
test project::worktree_policy::tests::worktree_enabled_for_origin_honors_registered_false ... ok
test project::worktree_policy::tests::worktree_enabled_at_reads_registry_from_disk ... ok
test project::worktree_policy::tests::worktree_enabled_at_defaults_true_when_registry_absent ... ok
test project::worktree_policy::tests::worktree_enabled_at_defaults_true_on_malformed_registry ... ok
test project::worktree_policy::tests::registry_data_dir_under_matches_daemon_layout ... ok
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 3898 filtered out
```

### Full surface — `cargo test -p trusty-mpm` (no `--lib`)

```
Running unittests src/lib.rs        test result: ok. 3901 passed; 0 failed; 6 ignored
Running unittests src/bin/tm/main.rs test result: ok. 1265 passed; 0 failed; 1 ignored
Running unittests src/bin/tm/main.rs test result: ok. 1265 passed; 0 failed; 1 ignored
... 30 integration targets, all ok ...
Running tests/tm_hook_pm_guard.rs   test result: FAILED. 31 passed; 2 failed
```

Those two `tm_hook_pm_guard` failures are an artifact of **where this worktree lives**, not of this diff. The tests assert `git worktree add .claude/worktrees/wt-x` is *allowed*; the path is relative, `run_pm_guard` sets no `current_dir`, so it resolves against the crate dir — which is under `/private/tmp/claude-502/...`, a path the #3955 guard denies by design. Running the same compiled binary from a cwd outside the scratchpad:

```
test result: ok. 33 passed; 0 failed; 0 ignored
```

CI checks out under `/home/runner/work`, so this will be green there. Nothing in this diff touches `pm_guard`, hooks, or the guard's path list.

One transient: `telegram::supervisor::tests::healthy_run_resets_transient_backoff` failed once under full-suite parallel load (a wall-clock backoff assertion: `21.79ms` vs `~base`), passed 5/5 in isolation and on the next full run. It is structurally load-sensitive and this diff touches nothing under `telegram/`. Flagging it rather than burying it.

`cargo fmt --check`, `cargo clippy --all-targets -D warnings`, and `scripts/check_line_cap.sh` are clean. `guided.rs` drops 484 → 449 SLOC.

## Test-hygiene notes

No `#[ignore]`, no env gate, no early `return`, no skip-when-fixture-missing — a fixture that cannot be built `unwrap`s and fails. Assertions are on specific paths and on the **absence** of the base-clone directory, its `.git`, and its `.worktrees` child, plus an exact-equality assertion on the returned `ManagedWorkspace` variant. Tests live in the bin target, so `--lib` would skip them; the full run above is without `--lib`.

Blast radius: only the three fallback tests touch a process-global (`TRUSTY_MPM_REPOS_ROOT`), and they are `#[serial_test::serial]` with an RAII guard that restores on panic-driven unwind. The `tm launch` tests take `base_path` as a parameter and touch no global at all. Nothing in this file deletes anything — the only filesystem writes are `git init` / `git worktree add` under test-owned `TempDir`s — so a clobbered global can make a test fail, never make it destructive.

## Contradicting the issue's description

The issue says the guided path bails when the URL is unparseable, implying `parse_github_path` is a real gate. It is not: `parse_github_path("not-a-url")` returns `Some(unknown-owner/not-a-url)`, so that `else` arm is effectively dead for non-URL input, and a test I initially wrote against it did a real `git clone not-a-url`. I dropped that test rather than pin dead behaviour. In production the arm is unreachable anyway — `fallback_protected` gates on `is_github_remote` first. Out of scope here; worth a separate look at whether `parse_github_path`'s permissiveness masks bad remotes elsewhere.

Context: #3455, #3781 (not closed by this PR).

Closes #4300

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -165,6 +165,25 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The per-project worktree opt-out is no longer silently conditional on the
+  daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
+  `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
+  [#3781](https://github.com/bobmatnyc/trusty-tools/pull/3781)) was consulted at
+  exactly one decision point, inside the daemon's spawn route. Both CLI paths
+  that provision worktrees in the `tm` process itself — `tm launch`, which never
+  asks the daemon, and the daemon-unreachable bare-`tm` fallback, which runs
+  *because* the daemon is down — created a base clone and a per-session worktree
+  regardless. Both now consult the registry BEFORE `ensure_base_clone`, matching
+  the daemon's deliberate ordering, so an opted-out project gets neither a clone
+  nor a worktree and runs in its own checkout exactly as `spawn_managed_on_main`
+  would have given it. The decision itself moved from the `pub(super)`
+  `daemon::managed_routes::launch_on_main::worktree_enabled_for_origin` to
+  `project::worktree_policy`, which serves the daemon in-process and the CLI
+  out-of-process off the same `projects.json`. The `unwrap_or(true)` default is
+  unchanged (unset or unregistered still means worktrees ON, and an unreadable
+  registry fails safe to ON), and #3455's warn-never-refuse concurrency
+  behaviour is untouched.
+
 - Concurrent `tm` processes no longer lose (or corrupt) `projects.json` entries.
   The project-store upsert was an unsynchronised read-modify-write of the whole
   file — two processes interleaving read/read/write/write silently dropped one
--- a/crates/trusty-mpm/src/bin/tm/commands/guided.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided.rs
@@ -1082,10 +1082,13 @@ pub(crate) fn managed_pane_settle_pending_message(id: &str) -> String {
 /// if cwd is inside a git working tree (at ANY depth), it is never written to.
 /// What: three-way dispatch on cwd:
 ///   (1) Inside a GitHub-backed git working tree → delegate to
-///       [`redirect_to_managed_clone`] which provisions (or reuses) the protected
+///       [`launch_protected_workspace`] which provisions (or reuses) the protected
 ///       base clone and per-session worktree, then calls `launch()` against THAT
 ///       workspace, never the live checkout. Uses the repo ROOT (not cwd) to
-///       read the origin remote — correct even from subdirectories.
+///       read the origin remote — correct even from subdirectories. #4300: a
+///       project registered with `worktree: false` (#3455) gets neither clone
+///       nor worktree and launches in the repo root instead, matching what the
+///       daemon's `spawn_managed_on_main` would have done had it been up.
 ///   (2) Inside a non-GitHub git working tree (non-GitHub remote or no remote)
 ///       → return an actionable `Err`; the live checkout is never touched.
 ///   (3) Not inside any git working tree (plain directory or bare repo) →
@@ -1161,8 +1164,9 @@ pub(crate) async fn fallback_protected(
     if let Some(ref raw_url) = origin_url
         && is_github_remote(raw_url)
     {
-        // GitHub project: redirect deploy to the protected managed clone.
-        return redirect_to_managed_clone(client, url, cwd, raw_url).await;
+        // GitHub project: redirect deploy to the protected managed clone
+        // (or, when the project opted out of worktrees, to the repo root).
+        return launch_protected_workspace(client, url, &git_root, raw_url).await;
     }
 
     // Non-GitHub remote or no remote: refuse to write to the live tree.
@@ -1179,84 +1183,41 @@ pub(crate) async fn fallback_protected(
     );
 }
 
-/// Redirect the guided-default fallback to the protected managed-clone workspace.
+/// Launch the guided-default fallback in the workspace this project is
+/// entitled to — the protected managed clone, or its own main checkout.
 ///
 /// Why: when the daemon is unreachable and the current directory is a GitHub-backed
 /// git project, framework files must go into the managed-clone workspace
 /// (`~/trusty-mpm-projects/<owner>/<repo>/.worktrees/<session-id>/`), never
-/// into the operator's live checkout (#1724, #1803).
-/// What: (1) parses `owner/repo` from `origin_url`; (2) ensures the protected
-/// base clone exists (`ensure_base_clone` is idempotent — returns immediately when
-/// the clone already exists); (3) creates a per-session git worktree inside the
-/// base clone; (4) calls `launch()` with the worktree path as the target directory.
-/// If any step fails (unparseable URL, clone error, worktree error), the function
-/// returns `Err` with an actionable message — the live checkout is never touched.
-/// Test: `guided_fallback_never_pollutes_github_git_checkout`.
-async fn redirect_to_managed_clone(
+/// into the operator's live checkout (#1724, #1803) — UNLESS the project is
+/// registered with `worktree: false` (#3455), which is the operator saying
+/// "run in my main checkout" and which the daemon honours via
+/// `spawn_managed_on_main`. Before #4300 this path never consulted that
+/// setting, so the opt-out silently held only while the daemon was up.
+/// What: delegates the whole decision to
+/// [`super::managed_workspace::provision_for_fallback`] — which reads the
+/// registry BEFORE `ensure_base_clone`, so an opted-out project gets neither a
+/// clone nor a worktree — then calls `launch()` against the resolved workspace.
+/// On any failure (unparseable URL, clone error, worktree error) it returns
+/// `Err` with an actionable message and the live checkout is never touched.
+/// Test: `guided_fallback_never_pollutes_github_git_checkout`,
+/// `guided_fallback_redirect_success_worktree_not_live_checkout`; the #4300
+/// opt-out cases live in `managed_workspace_tests.rs`.
+async fn launch_protected_workspace(
     client: &reqwest::Client,
     url: &str,
-    cwd: &std::path::Path,
+    git_root: &std::path::Path,
     origin_url: &str,
 ) -> anyhow::Result<()> {
-    use trusty_mpm::daemon::managed_routes::inproject;
-    use trusty_mpm::session_manager::ManagedSessionId;
+    let session_id = trusty_mpm::session_manager::ManagedSessionId::new();
+    let workspace = super::managed_workspace::provision_for_fallback(
+        &trusty_mpm::project::registry_data_dir(),
+        origin_url,
+        git_root,
+        &session_id,
+    )
+    .await?;
 
-    // Parse owner/repo from the GitHub remote URL.
-    let Some(gh) = trusty_common::github_path::parse_github_path(origin_url) else {
-        eprintln!(
-            "tm: cannot determine GitHub project from remote URL '{origin_url}'.\n\
-             Start the daemon first with `tm start`, then run `tm` again."
-        );
-        anyhow::bail!(
-            "daemon unreachable: cannot parse GitHub remote URL as owner/repo — run `tm start` first"
-        );
-    };
-
-    // Ensure the protected base clone exists. Idempotent: returns Ok immediately
-    // when the clone is already present; clones once on first invocation.
-    let base = inproject::base_clone_path(&gh.owner, &gh.repo);
-    eprintln!(
-        "tm: daemon unreachable — redirecting to protected managed clone\n\
-         tm: base clone: {}",
-        base.display()
-    );
-    if let Err(e) = inproject::ensure_base_clone(origin_url, &base) {
-        eprintln!(
-            "tm: could not set up base clone for {}/{}: {e}\n\
-             Start the daemon first with `tm start`, then run `tm` again.",
-            gh.owner, gh.repo
-        );
-        anyhow::bail!("failed to set up managed base clone: {e}");
-    }
-
-    // Create a per-session worktree branched from the base clone.
-    let session_id = ManagedSessionId::new();
-    // NOTE (#2032): this daemon-unreachable fallback path is NOT the managed
-    // SessionManager spawn flow — it has no session manager to resolve a
-    // semantic tmux name from, so it deliberately keeps the pre-#2032
-    // UUID-named worktree here. Only the daemon's `spawn_managed_inproject`
-    // path (`daemon::managed_routes::lifecycle`) uses the new semantic-name
-    // worktree layout.
-    let worktree =
-        match inproject::create_session_worktree(&base, &session_id.to_string(), &session_id) {
-            Ok(p) => p,
-            Err(e) => {
-                eprintln!(
-                    "tm: could not create per-session worktree: {e}\n\
-                 Start the daemon first with `tm start`, then run `tm` again."
-                );
-                anyhow::bail!("failed to create session worktree: {e}");
-            }
-        };
-
-    eprintln!(
-        "tm: launching in protected workspace (live checkout at {} is untouched)\n\
-         tm: session worktree: {}",
-        cwd.display(),
-        worktree.display()
-    );
-
-    // Launch in the session worktree — not the live checkout.
-    let dir = worktree.to_string_lossy().to_string();
+    let dir = workspace.path().to_string_lossy().to_string();
     super::launch::launch(client, url, Some(dir), None).await
 }
--- a/crates/trusty-mpm/src/bin/tm/commands/guided.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided.rs
@@ -786,9 +786,17 @@ pub(crate) fn is_github_remote(url: &str) -> bool {
 /// (plain directory or bare repo). Bare repos (`--git-dir` succeeds but
 /// `--show-toplevel` does not) are treated as non-git — deploying into a bare
 /// repo has no live-checkout to protect.
+///
+/// `pub(crate)` since #4300: `managed_workspace::provision_for_launch` must
+/// resolve the repo root the SAME way this path does. `tm launch` has no
+/// `classify_cwd_project` pre-pass, so without a shared definition the two CLI
+/// entry points answered "which directory is this project?" differently and
+/// `tm launch` from a subdirectory deployed into that subdirectory.
 /// Test: `guided_fallback_blocks_github_git_from_subdirectory` calls
-/// `fallback_protected` from a nested subdir and asserts the protection fires.
-fn find_git_root(cwd: &std::path::Path) -> Option<std::path::PathBuf> {
+/// `fallback_protected` from a nested subdir and asserts the protection fires;
+/// `provision_for_launch_from_subdirectory_targets_repo_root` covers the
+/// `tm launch` side.
+pub(crate) fn find_git_root(cwd: &std::path::Path) -> Option<std::path::PathBuf> {
     let out = std::process::Command::new("git")
         .arg("-C")
         .arg(cwd)
--- a/crates/trusty-mpm/src/bin/tm/commands/launch.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/launch.rs
@@ -155,53 +155,34 @@ pub(crate) async fn launch(
         return Ok(());
     }
 
-    // 5. Warn about uncommitted local changes (one-time; non-fatal).
-    eprintln!(
-        "note: uncommitted local changes are not carried into the managed clone. \
-         Use `tm connect` if you need to work from the live checkout."
-    );
-
-    // 6. --style is not yet honoured in managed mode.
+    // 5. --style is not yet honoured in managed mode.
     if style.is_some() {
         tracing::warn!("--style not yet supported for managed launches; ignoring style flag");
         // TODO(follow-up): prepare_session_with_repo_url_and_style
     }
 
-    // 7. Provision the managed workspace using the shared base-clone + per-session
+    // 6. Provision the managed workspace using the shared base-clone + per-session
     //    worktree mechanism (#1803). This matches what `tm` (guided default) does
-    //    via the daemon's `spawn_managed_inproject` path — both now converge on the
+    //    via the daemon's `spawn_managed_inproject` path — both converge on the
     //    SAME base clone at `~/trusty-mpm-projects/<owner>/<repo>/` and the SAME
     //    per-session worktree at `<base>/.worktrees/<session-id>/`.
-    //    Step 7a creates or reuses the shared base clone (idempotent); step 7b adds
-    //    a fresh git worktree for this session; step 7c deploys `.claude` into the
-    //    worktree so the session has the framework available.
-    eprintln!("provisioning managed workspace...");
+    //    #4300: `provision_for_launch` consults the per-project `worktree`
+    //    opt-out (#3455) FIRST — a project registered with `worktree: false`
+    //    gets neither a base clone nor a worktree here, exactly as the daemon's
+    //    `spawn_managed_on_main` gives it neither. It also owns the
+    //    uncommitted-changes notice, which only applies when there IS a clone.
     let session_uuid = trusty_mpm::session_manager::ManagedSessionId::new();
-
-    // 7a. Ensure the shared base clone exists. Idempotent: returns immediately when
-    //     `<project_dir>/.git` is already present so a second `tm launch` reuses
-    //     the existing clone rather than re-cloning.
-    trusty_mpm::daemon::managed_routes::inproject::ensure_base_clone(&origin_url, &project_dir)
-        .map_err(|e| anyhow::anyhow!("failed to provision base clone: {e}"))?;
-
-    // 7b. Create a per-session git worktree at `<project_dir>/.worktrees/<session-id>/`.
-    //     Each session gets an isolated branch so concurrent sessions never collide.
-    //     NOTE (#2032): `tm launch` is a standalone CLI flow, not the daemon's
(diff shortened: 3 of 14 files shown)


bobmatnyc:
## Verdict: APPROVE

Adversarial review at `efed40a2` against `origin/main`. Zero CRITICAL, zero HIGH at >80% confidence. Two MEDIUM and two LOW below; none block.

I did not take the PR's claims on trust — I re-ran the red/green experiment myself in an isolated worktree and traced the registry seam through to `ProjectStore`.

### Claims verified

**1. Red/green with the guard neutralised — YES.**

Baseline, verbatim:
```
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 1260 filtered out; finished in 0.28s   (--bin tm managed_workspace)
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 3898 filtered out; finished in 0.12s   (--lib worktree_policy)
```
With `if false && !isolate` at `managed_workspace.rs:97`, exactly the two opted-out tests go RED and all four controls stay green:
```
test result: FAILED. 4 passed; 2 failed; 0 ignored; 0 measured; 1260 filtered out; finished in 88.43s
  left: Worktree(".../bobmatnyc/writing/.worktrees/93d67791-...")
 right: MainCheckout(".../.tmpGJ9iZl")
```
The tests are not vacuous. **And the "it cloned for real" claim is also confirmed** — 88.43s vs 0.28s, no auth/DNS error, and a worktree successfully created under a freshly cloned base. See MEDIUM-2.

**2. Ordering is load-bearing, separately proven — YES.** Moving the guard after `ensure_base_clone` (`managed_workspace.rs:106`) fires the ordering assertion at `managed_workspace_tests.rs:135` for both cases:
```
#4300: base clone dir must NOT be created for an opted-out project: /var/.../bob-duetto/cto
```
The assertion genuinely exists and genuinely fires. This matches the daemon's own documented rationale at `lifecycle.rs:365-372`.

**3. `registry_data_dir` unification — YES on substance, partially on the pin.** Divergence is now *structurally* impossible: `state/core.rs:1076` and the CLI both go through `registry_data_dir_under`, and `DaemonState::new` (`state/core.rs:374`) roots at `FrameworkPaths::default().root` — the same root `registry_data_dir()` derives. That is the real fix. The test is weaker than advertised: `registry_data_dir_under_matches_daemon_layout`'s first assertion restates the implementation and cannot fail; only the second (the literal `"project-registry"`) would catch a rename. See LOW-2.

**4. Full surface — YES via CI**, 23 pass / 1 skipping, `Test` job 17m7s. I re-ran the two new modules directly (6/6, 9/9 above); I did not independently re-run all 3901+1265.

### Seam judgment — the move is better than what the issue proposed

Agreed, and for a stronger reason than the PR gives. `Project::worktree` is a field of the project record, so "does this origin get a worktree?" is a registry query, not a daemon concern. Relaxing `pub(super)` would have left the CLI reaching into `daemon::` for it while `launch_on_main`'s module doc still claimed `lifecycle::spawn_managed_routed` was "the only caller" — a premise #4300 had already falsified. Keeping `pub(super)` on `spawn_managed_on_main`/`has_concurrent_main_checkout_session` (genuine spawn surface) while hoisting only the decision is the right cut.

### The out-of-process read does NOT need locking

`project/store.rs:35-38` settles it: *"Readers take no file lock, which is safe because every publish is an atomic rename: a reader sees a complete old or complete new document, never a torn one."* Writes funnel through `json_rmw::update` (#4258, `623d6e3e`); reads do not participate. Three further checks:

- **No staleness hazard.** The daemon's own reads call `reload_if_changed()` (`store.rs:138`), so disk is authoritative for daemon and CLI alike — the CLI is not reading a stale shadow of daemon memory.
- **No write side-effect.** `ProjectStore::load` (`store.rs:159`) is a pure read; it does not create the dir or file. The CLI cannot perturb the registry by asking.
- **Fail direction is safe.** A torn or malformed read resolves to `true` (isolation ON) at `worktree_policy.rs:1237-1248`, pinned by `worktree_enabled_at_defaults_true_on_malformed_registry`. Worst case is an unwanted worktree, never an unwanted main-checkout launch.

### Attacks that came up clean

- **No data-loss path.** I chased whether an opted-out `tm launch` could get the operator's checkout auto-deleted. It cannot: `tm launch` registers via `POST /sessions` → `register_session` (`api.rs:602`), which builds a `core::session::Session` that has no `workspace_owned` field at all. Every `SessionRecord` construction that could reach a main checkout (`adopt.rs:227/434/463`) hardcodes `workspace_owned: false`, and the deletion gate is `workspace_owned || is_session_worktree(path)` (`workspace_liveness.rs:190`), which a main checkout never satisfies.
- **Reconnect still works in opt-out mode.** `session_matches_workdir` (`launch.rs:637`) exact-matches on `workdir` before the `project_dir` prefix-match, and an opted-out session registers `project_path == live_path`, so a second `tm launch` reconnects rather than stacking a duplicate session on one checkout. Passing `live_path` down made this fall out correctly.
- **Test isolation is sound.** Every `TRUSTY_MPM_REPOS_ROOT` mutator in the `tm` binary carries `#[serial_test::serial]` — the new tests, `tests_behavior_b_tests.rs:892/1213/1317/1376/1429`, and `tests_behavior_c_tests.rs`'s `ENV_MUTEX` tests, which hold **both** the mutex and the attribute. So the serial_test lock and `ENV_MUTEX` never need to interoperate — nothing guards that var with `ENV_MUTEX` alone. The `ReposRootGuard` restores on unwind. This is not the sibling PR's failure mode: nothing here is destructive and all writes land in test-owned `TempDir`s.
- **`guided.rs` 1262 → 1223 is genuine.** The 74 deleted lines are the inlined provisioning body, moved verbatim into `managed_workspace.rs`. Only behavioural delta: base-clone and worktree failures now share one error string instead of two, and the "live checkout at X is untouched" notice now names `git_root` instead of `cwd` (an improvement).

### Self-reported qualifications — both confirmed

**`parse_github_path` is permissive — CONFIRMED, and leaving it is right.** Traced `trusty-common/src/github_path.rs`: `"not-a-url"` → segments `["not-a-url"]` → the `[repo]` arm → `Some(unknown-owner/not-a-url)`. So the bail at `managed_workspace.rs:174` is dead for non-URL input, and doubly unreachable in production since `fallback_protected` gates on `is_github_remote` first. Keep it: the parser returns `Option`, so the `else` is the only non-panicking way to consume it — removing it forces an `.expect()`, which is strictly worse. Dropping the test that would have pinned the permissive behaviour was also the right call; a test that certifies dead behaviour as intended is an anti-pattern.

**Both local failures are artifacts, not regressions — CONFIRMED structurally.** The diff touches 14 files; none is under `telegram/` and none is in the PM-guard/hook path, so neither test's code is in the change set at all. `tm_hook_pm_guard`'s explanation holds (`run_pm_guard` sets no `current_dir`, so a relative `git worktree add` resolves against a scratchpad path the #3955 guard denies by design) and CI is green. The telegram supervisor test is a pre-existing load-sensitive flake — it deserves its own ticket rather than being absorbed as noise, but it is not attributable to this PR.

---

## Findings

| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/launch.rs` | 104-105, 179 | The two entry points added by this PR resolve the opt-out target differently. `provision_for_fallback` deliberately uses `git_root` ("the repo ROOT, not the possibly-nested cwd"); `provision_for_launch` is handed `live_path`, which is `resolve_dir(dir)` — the raw cwd, with no git-root resolution (`commands/project.rs:25`). `get_origin_url` runs `git -C <path> config --get remote.origin.url`, which resolves at any depth, so `cd opted-out-repo/src && tm launch` matches the opt-out and then returns `MainCheckout(.../src)` — deploying `.claude` (`launch.rs:193`) and project hooks (`launch.rs:229`) into the subdirectory rather than the project root. Before this PR the workspace was always a worktree, so cwd depth did not matter. | Resolve the git root for the `MainCheckout` target too (the same `git_root` the fallback already computes), or state explicitly that `tm launch` intends cwd semantics to match `tm connect` (`launch.rs:427`, also cwd-based) and say so in the module doc. As written the PR asserts both entry points behave "exactly as `spawn_managed_on_main`" while resolving the same concept two different ways. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/managed_workspace_tests.rs` | 665, 776 | The opted-out fixtures use real, resolvable, **private** origins (`git@github.com:bobmatnyc/writing.git`, `https://github.com/bob-duetto/cto.git`) and deliberately do not pre-create `base/.git`. Verified empirically: a guard regression turns a 0.28s unit test into an 88.43s run that **successfully clones those real repositories** into a temp dir on a credentialed machine. The failure signal is also environment-dependent — a clean assertion failure locally, an SSH auth error in CI. This repo already documents the fix for exactly this hazard at `tests_behavior_c_tests.rs:1570-1580`: *"Deliberately uses a fixture-only host/owner/repo … a developer machine may have a real, reachable SSH config entry … which would let `ensure_base_clone` actually succeed and make this test flaky/order-dependent on local SSH config instead of hermetic."* | Swap the two opted-out fixtures to a non-resolvable host, e.g. `git@github-test-fixture-nonexistent:acmetest/writing-fixture.git`. Nothing depends on the URL being real — the registry fixture is keyed by whatever URL the test itself writes. Lossless, and makes a regression fail fast and offline. To be clear: **in the shipped state these tests never touch the network** (baseline 0.28s), so this is hygiene, not a live blast radius, and it is *not* the sibling PR's defect class — nothing here is destructive and the process-global is properly guarded. |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/guided.rs` | 1214-1223 | On the composed fallback path, `launch_protected_workspace` calls `provision_for_fallback` (registry read #1, notice #1) and then `launch()`, which calls `provision_for_launch` (registry read #2, notice #2). An opted-out project prints two near-identical opt-out notices and the registry is read twice per invocation. | Cosmetic; fold if convenient. The per-function "notice matches action" invariant does hold — this is duplication, not disagreement. |
| LOW | `crates/trusty-mpm/src/project/worktree_policy_tests.rs` | 1498-1510 | `registry_data_dir_under_matches_daemon_layout`'s first assertion (`registry_data_dir_under(root) == root.join(REGISTRY_DIR_NAME)`) restates the implementation and cannot fail. Only the second assertion (the `"project-registry"` literal) carries signal. The doc claims the test guards CLI/daemon path divergence, but that is guaranteed by both callers sharing one function, not by this test. | Drop the tautological assertion and reword the doc to say what the test actually pins: the on-disk directory *name*, whose change would orphan every existing `projects.json`. |

## Notes

- **Pre-existing, not introduced, but it qualifies one claim.** On the *opted-in* fallback path the composed flow still provisions twice: `provision_for_fallback` creates worktree W1 and prints `tm: session worktree: W1`, then `launch()` creates a second worktree W2 and runs the session there. So at the composed level a path still "prints one thing and does another", and leaks an orphan worktree per bare-`tm` fallback. `git show origin/main:...guided.rs` confirms the old `redirect_to_managed_clone` did exactly the same, so this PR neither causes nor worsens it. The `ManagedWorkspace` invariant holds *within* each function, which is all the type can enforce. Worth a follow-up ticket given the standing worktree-count pressure — not a blocker here.
- `ManagedWorkspace`'s stated safety benefit is only exercised in `provision_for_fallback` (`managed_workspace.rs:219`). `launch.rs:183` immediately does `workspace.path().to_path_buf()` and discards the variant, then runs the identical deploy/hook sequence for both. That is intended (the daemon's `spawn_managed_on_main` also deploys into the main checkout), but the enum does less work than the doc implies.
- `async fn provision` (`managed_workspace.rs:90`) contains no `.await` and makes a blocking `git clone` on the async path. Pre-existing shape, harmless in a single-purpose CLI; noted only so it is not mistaken for concurrency-safe.
- Closing keywords verified twice, independently, several minutes apart: `closingIssuesReferences` returns **only #4300** both times. #3455 and #3781 appear in the body as context with no closing keyword. Clean.
- CHANGELOG entry is correctly under `## [Unreleased]` (line 8) → `### Fixed`.
- PR is `BEHIND` main; rebase before merge.

No CRITICAL and no HIGH — **APPROVED** for the next pipeline stage. The two MEDIUMs are worth addressing but neither affects the correctness of the shipped fix.

bobmatnyc:
## Review round 2 — both MEDIUMs and both LOWs addressed, rebased onto `6de7f746`

### MEDIUM 1 — entry points now resolve the opt-out target identically

You were right that this is a real defect **introduced by this PR**: pre-#4300 the launch path always redirected to a worktree, so `cwd` was never a deploy target. `provision_for_launch` now resolves the repo root via `guided::find_git_root` — the *same* helper `classify_cwd_project` uses on the fallback side, promoted to `pub(crate)` for exactly this reason, so there is one definition of "which directory is this project". Non-git `cwd` (or no `git` on PATH) falls back to `cwd`, i.e. the previous behaviour, never worse.

New test `provision_for_launch_from_subdirectory_targets_repo_root` builds a real git working tree, launches from `crates/inner`, and asserts the canonical repo root is the target (and explicitly *not* the subdirectory). Red with the fix reverted to `cwd.to_path_buf()`:

```
---- provision_for_launch_from_subdirectory_targets_repo_root stdout ----
assertion `left == right` failed: #4300: `tm launch` from a subdirectory must target the repo ROOT
  left:  MainCheckout("/var/folders/…/.tmpKJEyWi/crates/inner")
 right:  MainCheckout("/private/var/folders/…/.tmpKJEyWi")
```

### MEDIUM 2 — fixtures no longer touch real remotes

All origins moved to the non-resolvable `.invalid` TLD with fixture-only owner/repo names, per `tests_behavior_c_tests.rs:1570-1580`. The module doc records why, citing your 88.43s-vs-0.28s measurement.

One thing worth flagging, because it slightly changes the shape of the evidence: with an unresolvable host the guard-neutralised revert now fails at `ensure_base_clone` rather than at the path assertion, so a bare `.unwrap()` would have reported "called `Result::unwrap()` on an `Err`" — red, but for a reason that does not name the behaviour. The three opt-out tests therefore carry an `expect` that does:

```
---- provision_for_launch_opted_out_creates_no_clone_and_no_worktree stdout ----
#4300: an opted-out project must not attempt a clone at all — a clone error here
means the opt-out was never consulted: failed to provision managed workspace:
inproject: git clone failed (exit status: 128): … Could not resolve host: github.invalid

test result: FAILED. 4 passed; 3 failed; 0 ignored; finished in 0.34s
```

**0.34s, no network** — against 88.43s before. The `assert_nothing_provisioned` half still guards the ordering claim, unchanged.

### LOW 1 — duplicate notice fixed, double read kept and justified

The fallback now prints only the daemon-unreachable context it uniquely knows; the "no managed clone, no worktree" sentence comes once, from the `launch()` it delegates to. The double registry read stays: collapsing it means threading a pre-resolved `ManagedWorkspace` through `launch()`'s signature to save one `read_to_string` of an atomically-published, lock-free, fail-open file. That trade is recorded in the module doc as a known, accepted redundancy rather than left silent.

### LOW 2 — tautology removed

Renamed to `registry_dir_name_is_frozen`. It now pins the on-disk literal (renaming `project-registry` orphans every existing `projects.json`) plus that the value is a single component directly under the framework root, and its doc states plainly what it does **not** pin — daemon/CLI agreement, which is structural because both call `registry_data_dir_under`. Your framing that "the function, not the test, is the protection" is quoted in the doc.

### Rebase

Rebased cleanly onto `6de7f746` (3 commits, no conflicts), force-pushed with `--force-with-lease`.

### Re-verification on the rebased tree

`cargo test -p trusty-mpm` without `--lib`: lib `3904 passed; 0 failed`, bin `tm` `1266 passed; 0 failed` (×2 targets), all integration targets ok except the two `tm_hook_pm_guard` tests that fail only because this worktree lives under `/private/tmp` — running the same binary from a cwd outside the scratchpad gives `33 passed; 0 failed`, and CI is green on them. `cargo fmt --check`, `clippy --all-targets -D warnings` (0 errors), `check_line_cap.sh`, and `check_test_pointers.sh` all clean.

Issue #2463, opened 2026-07-11
chore(trusty-mpm): session_end_pane_still_live should aggregate any-pane-live like the reaper's find_runtime_exited


bobmatnyc:
Follow-up from PR #2455's verify pass (non-blocking finding).

session_end_pane_still_live (daemon/api.rs, added in #2455) uses .find() — first matching pane only — whereas runtime_reap::find_runtime_exited aggregates ALL panes in a session and treats the session as live if ANY pane is live.

If an operator manually splits a managed session's window (multi-pane), and tmux list-panes -a returns an idle pane before a live one, the SessionEnd gate could misclassify a genuinely-live session as idle.

Narrow edge case (managed sessions are single-pane by construction) and not a regression below the pre-#2455 baseline, but switch to the reaper's any-pane-live aggregation for full parity.

Provenance: PR #2455 / issue #2454.

Issue #2023, opened 2026-07-05
feat(trusty-mpm): relaunch the current managed session in place when Claude exits (no new session)


bobmatnyc:
## Motivation
When a user is inside a managed `tm` session (Claude Code in a tm-spawned tmux pane) and exits Claude (Ctrl-D / `/exit`), they should be able to relaunch/resume THAT SAME session (same session id, worktree, CLAUDE_CONFIG_DIR, resuming the conversation) via `tm`, WITHOUT spawning a new session. Today it dead-ends or forces a new session.

## Root cause (from investigation)
On Claude exit the tmux pane survives (drops to a shell), but the daemon runtime-reaper (`crates/trusty-mpm/src/daemon/runtime_reap.rs`, ~60s tick via `daemon/mod.rs:380-430`) marks the record `Stopped` and then KILLS the tmux pane via `SessionManager::stop` → `graceful_terminate_runtime` → `kill_session` (`session_manager/manager.rs:589-600`, `restart_ops.rs:139-156`) — the same code path as an explicit `tm session stop`. So the user has ~60s before the pane vanishes, and cannot relaunch from inside the dying pane (running `tm` there would self-kill the tmux session by name). Resume from ANOTHER terminal already works (kill+recreate same-named tmux at same workspace via `SessionManager::resume`/`resume_managed`, `manager.rs:616-665`, `daemon/managed_routes/lifecycle.rs:1015-1064`) but that is not in-place.

## Design decisions (LOCKED by owner)
1. **Reaper: leave the pane alive.** Runtime-exit must mark the record `Stopped` WITHOUT killing the pane. Pane-kill is reserved strictly for explicit `tm session stop` / `decommission`. This removes the 60s race. Implement as a non-destructive stop variant (or flag) so `runtime_reap::stop_runtime_exited` no longer calls the pane-killing path.
2. **Invocation: bare `tm` auto-detects in-pane.** Running `tm` from inside the managed pane detects it is already inside this managed session and does an IN-PLACE `claude --resume <id>` (correct CLAUDE_CONFIG_DIR) in the current shell — NOT the kill+recreate HTTP resume path. Likely a 4th `ResumeAction::InPlace` branch in `bin/tm/commands/guided_resume.rs` `plan_resume`.
3. **On-exit UX: print a one-line hint.** After `claude` exits in the pane, print a one-line hint (e.g. "tm: run `tm` to relaunch this session"). No auto-relaunch.

## Implementation components (decomposable)
[...]

PR #4304, opened 2026-07-29
fix(trusty-mpm): resume a dead runtime instead of switching into a bare shell (#3873)

### LINKED ISSUE #3873: tm picker: wrong action for dead runtime — should RESUME not switch
## Evidence
Bob inside tmux session tm-apex-01 after its claude runtime died (runtime-reap marked Stopped, pane alive):
- Bare \`tm\` → picker → \`[1]\` printed: \`tm: switching client to session 'tm-apex-01'\`
- Returned to shell WITHOUT resuming the runtime
- Second \`tm\` try worked

## Problem
Picker detects live tmux session but NOT the dead runtime state. It performs tmux switch-client (correct for a living session) instead of the RESUME action (respawn runtime).

From a bare terminal, the action must be:
- **Live runtime:** attach/exec (tmux switch-client, or create window in session)
- **Dead runtime:** respawn claude runtime (RESUME action)

Current: always tmux-switch regardless of runtime state.

## Related
PR #3871 (ls/list support) touches the same picker — consider fixing together.

## Impact
Users cannot resume dead sessions from picker directly; requires explicit resume command.

bobmatnyc:
## The defect

Bare `tm` against a session whose **tmux session exists** but whose **agent runtime has exited** took the wrong branch: it `switch-client`ed the operator into a pane running a bare login shell and did nothing else. The operator then typed `tm` a second time — now *inside* that pane, where the in-place relaunch gate matches — and only then did the real relaunch happen.

This is not a double-invocation quirk. **Invocation #1 simply did not do the job.**

## Root cause

`commands/guided_resume.rs::plan_resume` decided "this session is live" from `tmux_live`, which is `tmux_has_session(&session.name)` — tmux **session existence**, not runtime liveness. Its own doc said *"everything else (active/provisioning with a live tmux) → Attach"*. So `(active, session-exists)` mapped to `Attach` even when the runtime was provably dead.

The path there, all under `crates/trusty-mpm/src/bin/tm/`:

- `main.rs` no subcommand → `commands::guided::run_guided_default`
- `guided.rs:101` `try_inplace_relaunch` requires `Stopped`; the record is `active`, so it declines
- `guided.rs:112` `nested_session_guard` matches by tmux session name; a different session → `None`
- `guided.rs:300`/`:319` → `try_show_picker` → `session_picker.rs:846` `PickerDecision::Resume`
- `guided_resume.rs` `resume_session` → `plan_resume(persisted_state, tmux_live)` → **`Attach`**

Live evidence: session `tm-apex-companion-01`, daemon record `state=active` / `pane_id=%333`, while tmux pane `%333` reported `pane_current_command=zsh` and `pgrep -P` returned no children — provably idle.

## The fix

`plan_resume` takes a third input, `runtime_live`, and maps `(active/provisioning, tmux_live = true, runtime_live = false)` → **`ReconcileThenRestart`**. That is the *existing* zombie recovery, which already does POST `/runtime-stop` → POST `/resume` → attach, so invocation #1 does the whole job.

Branch ordering is deliberate and pinned by tests:

- terminal tombstone (`decommissioned`/`deleted`) refusal stays **first** — a dead runtime must never resurrect a tombstone;
- `is_zombie` (tmux gone) is unchanged;
- `needs_restart` stays **ahead** of the new gate, so a Stopped/Errored record — whose pane is an idle shell too, and therefore also reports `runtime_live = false` — keeps taking the plain `Restart` path instead of a pointless `/runtime-stop` against an already-stopped record.

### Reusing the daemon's predicate, not reimplementing it

`runtime_live` is resolved by the new `pane_runtime_live(pane_id)`, which calls **`trusty_mpm::daemon::runtime_reap::pane_runtime_exited`** directly — the daemon reaper's own predicate, and therefore the same `orphan_gc::is_idle_shell` allowlist and the same `orphan_gc::ProcessTreeProbe` child gate. No idle-shell detection is duplicated in the CLI, and the two can never drift apart.

It was reachable with no new seam: `daemon` is `pub mod` in `lib.rs`, `runtime_reap` is `pub mod` in `daemon/mod.rs`, and `pane_runtime_exited` / `PaneInfo` / `ProcessTreeProbe` are all `pub`. The `bin/tm` binary already reaches into `trusty_mpm::daemon::*` elsewhere (`gh_identity.rs`, `commands/daemon_run.rs`).

The `PaneInfo` it feeds that predicate comes from a **single pane-id-targeted** `tmux display-message -p -t <pane_id> '#{pane_current_command} #{pane_pid}'` — deliberately **not** the `list-panes -a` whole-server enumeration, following the existing `session_for_pane` / `pane_tty_for` precedent in `tmux_attach.rs`. tmux resolves a pane id to exactly one pane or errors; there is nothing to guess and no composite row to mis-split.

### Fails open

`runtime_live = false` now triggers a `/runtime-stop` that **kills the pane**, so the only path to `false` is a positive two-gate identification (a recognised bare shell **and** no live child under the pane pid). Everything else returns `true`: no `pane_id` on the record, a failed or non-zero `tmux` call, an unparseable reply. A wrong `false` would destroy a live agent's pane; a wrong `true` merely reproduces today's already-shipped behavior.

The shell-out is also skipped entirely when `tmux_live` is false — `is_zombie` already classifies that record, and the pane is gone with the session.

## Scope — what this does NOT fix

**This does not fix the runtime-reaper defect, and does not depend on it.**

There is a second, deeper problem under separate investigation: the 60s runtime reaper (`daemon/mod.rs` → `daemon/runtime_reap.rs`) is not running — `daemon::runtime_reap` had zero log entries in 15h because pane enumeration is corrupted (`session_name` holding an entire underscore-joined row, `pane_current_command` empty — the #1813 composite-key regression shape). **Nothing in this PR touches `daemon/orphan_gc.rs` or `daemon/tmux.rs` enumeration.**

That is precisely why this ships on its own: the CLI asks the liveness question *itself, at resume time*, through a targeted single-pane query. It is correct whether the reaper is healthy, broken, or simply hasn't ticked yet.

## Sibling, not a duplicate: #4061 / PR #4065

#4061 is a **sibling** fix, not a duplicate. It patched only `fallback_protected`'s **non-git** branch in `guided.rs`, and is already in the 1.2.3 binary. The path fixed here is **git-backed** and returns via the picker — it never reaches `fallback_protected` at all.

## Test evidence

The required test was written first and observed RED against unmodified logic (the parameter was threaded through with the body untouched):

```
running 10 tests
test tests_behavior_c::guided_resume_plan_active_live_tmux_attaches ... ok
test tests_behavior_c::guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts ... FAILED
test tests_behavior_c::guided_resume_plan_provisioning_live_tmux_dead_runtime_reconciles_then_restarts ... FAILED

---- guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts stdout ----
  left: Attach
 right: ReconcileThenRestart

test result: FAILED. 8 passed; 2 failed; 0 ignored; 0 measured; 1258 filtered out
```

GREEN after the one-branch fix, with the pre-existing attach case kept and updated to `runtime_live = true`:

```
test tests_behavior_c::guided_resume_plan_active_live_tmux_attaches ... ok
test tests_behavior_c::guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts ... ok
test tests_behavior_c::guided_resume_plan_provisioning_live_tmux_dead_runtime_reconciles_then_restarts ... ok
test tests_behavior_c::guided_resume_plan_stopped_live_tmux_dead_runtime_still_plain_restarts ... ok
test tests_behavior_c::guided_resume_plan_active_no_tmux_reconciles_then_restarts ... ok
test tests_behavior_c::guided_resume_plan_provisioning_no_tmux_reconciles_then_restarts ... ok
test tests_behavior_c::guided_resume_plan_stopped_restarts ... ok
test tests_behavior_c::guided_resume_plan_stopped_with_stale_tmux_still_restarts ... ok
test tests_behavior_c::guided_resume_plan_errored_restarts ... ok
test tests_behavior_c::guided_resume_plan_resume_prefers_persisted_state_over_display_state ... ok
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 1258 filtered out

test tests_behavior_c::plan_resume_refuses_terminal_states ... ok
test tests_behavior_c::plan_resume_refuses_terminal_states_even_with_dead_runtime ... ok
test tests_behavior_c::pane_runtime_live_assumes_live_without_pane_id ... ok
test tests_behavior_c::parse_pane_probe_reads_command_and_pid ... ok
test tests_behavior_c::parse_pane_probe_tolerates_missing_pid ... ok
test tests_behavior_c::parse_pane_probe_none_for_empty_reply ... ok
```

Full crate suite, run **without** `--lib`:

```
$ cargo test -p trusty-mpm
test result: ok. 3894 passed; 0 failed; 6 ignored; ...
test result: ok. 1267 passed; 0 failed; 1 ignored; ...   (bin tm)
... 0 failures across every target
```

`cargo clippy -p trusty-mpm --all-targets` clean; `cargo fmt` applied; `scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`.

Every new test is a **pure-function** test: no `#[ignore]`, no env-var gate, no early return, no skip-when-missing, no global state, and therefore no `env_test_lock` / `#[serial_test::serial]` interaction to get wrong. `pane_runtime_live(None)` returns before any tmux shell-out. Assertions are on the specific `ResumeAction` variant.

## Nothing contradicting the root cause

Verified line-by-line against `origin/main` (6ef47ecb) before changing anything; every cited anchor held except `try_show_picker`'s call sites, which have drifted from `guided.rs:282` to `:300`/`:319`.

Closes #3873

---

### HIGH (fixed): the reconcile branch reached a session kill silently

The critic traced past my local doc comment (which said only "marks the record `Stopped`"):
`stop_managed_session_runtime` → `SessionManager::stop` (`manager.rs`) → `graceful_terminate_runtime` → **`kill_session(tmux_name)`** (`restart_ops.rs`). **Session-scoped — every window, every pane.**

That is free for the #2001 zombie, whose session is already gone. But **the #3873 shape has a live session by definition**, so this was a newly-destructive step on an input that previously took a harmless `Attach`. My removal of the warning reasoned "in neither case is there in-progress runtime state left to warn about losing" — that accounted only for the *recorded* pane, and the plain `Restart` branch twenty lines below already discloses exactly this for the identical `tmux_live == true` situation.

The reconcile branch now discloses it, naming what is lost:

```
tm: session 'tm-apex-01' is active but its runtime has exited — every pane in the
tmux session is now an idle shell. Restarting will KILL that tmux session (all
windows and panes, including their scrollback) and start a fresh runtime.
```

`reconcile_zombie_stop`'s doc comment — the one that made this easy to miss — now documents the full server-side chain and flags the call as destructive.

**Sibling-pane check: it gates, and it does so structurally.** Rather than bolting a `session_has_live_pane` call on beside the existing verdict, `runtime_live` *is* now the any-pane-live question (see the third finding below). A session with one idle pane and one live agent pane can no longer reach the destructive branch at all — the guard cannot be bypassed by a future edit that forgets to call it, because there is no separate call to forget.

### MEDIUM (fixed): a missing pid collapsed the two-gate check to one

`parse_pane_probe` returned `Some((cmd, None))` for input like `"zsh notanumber"`, and `ProcessTreeProbe::has_live_child(None)` returns `false`, so `pane_runtime_exited` degenerated to `is_idle_shell` alone — contradicting the "two-gate" contract I documented.

I accept the framing. `None → false` is **correct calibration for the daemon**, because #2023 A deliberately made the reaper non-destructive (record flips, pane survives). Reusing that gate is still right for consistency — but **the CLI must be stricter than the daemon, not equal**, because its consequence is a `kill_session`. Parsing now requires exactly three non-empty tab-separated fields with a numeric pid; anything else means "cannot determine" and fails open.

RED before, showing the exact defective value:

```
---- parse_pane_probes_rejects_non_numeric_pid stdout ----
  left: Some([PaneInfo { session_name: "tm-apex-01", pane_current_command: "zsh",
                         pane_pid: None, pane_id: None }])
 right: None
```

GREEN after, with the sibling cases pinned too:

```
test parse_pane_probes_reads_every_pane ... ok
test parse_pane_probes_rejects_non_numeric_pid ... ok
test parse_pane_probes_rejects_missing_pid ... ok
test parse_pane_probes_rejects_smeared_row ... ok
test parse_pane_probes_rejects_extra_fields ... ok
test parse_pane_probes_rejects_empty_command_field ... ok
test parse_pane_probes_rejects_empty_or_blank_listing ... ok
```

### Third finding (MEDIUM, `session_name: String::new()`): judged reachable — fixed, not argued away

The critic flagged this as a missing guard whose reachability it could not establish. My judgement: **it is reachable, and it was the more serious of the two sub-problems.**

- *Any-pane-live.* The codebase states twice that a managed session can carry sibling panes (#2463 `session_has_live_pane`, #2456 hijack note). A manual `tmux split-window` is an ordinary thing to do. With a single-pane check, a session whose recorded pane went idle while a sibling pane still ran the agent would be classified dead — and killed. Reachable by any operator who splits a pane.
- *Pane membership.* Nothing verified that the recorded `pane_id` still belonged to this session. tmux pane ids are not reused within a server's lifetime, but a record can outlive a tmux server restart, after which `%333` may be an unrelated live pane.

Both are fixed by asking a different question. `session_runtime_live` now runs a **session-scoped** `tmux list-panes -s -t <session>` and delegates to the daemon's own `session_has_live_pane` — the exact #2463 primitive I had bypassed. Membership holds by construction (a `-t <session>` listing cannot return another session's panes), so the fabricated `session_name` is gone, and the verdict is any-pane-live rather than one-pane.

One fail-closed corner needed an explicit guard: `session_has_live_pane` filters by session name and returns `false` when *nothing* matches. That is right for the daemon (a whole-server listing with no match means the session is gone) but destructive here, so a listing that cannot be attributed to the session we asked about fails open instead.

```
test session_panes_any_live_pane_keeps_session_live ... ok
test session_panes_all_idle_reads_dead ... ok
test session_panes_foreign_rows_never_prove_our_session_dead ... ok
test session_runtime_live_assumes_live_for_blank_session_name ... ok
```

The TAB delimiter is also deliberate: it cannot occur in a session name or command word, so the columns cannot smear the way the space-joined `list-panes -a` format does. `parse_pane_probes_rejects_smeared_row` pins the #1813 shape. Still no `list-panes -a`, still nothing touched in `daemon/orphan_gc.rs` or `daemon/tmux.rs`.

### Doc correction

The empirical note is now in the code: on tmux 3.6b an invalid target can return **rc=0 with near-empty stdout**, so the `!status.success()` guard is belt-and-braces and the *parse-returned-`None`* path is the real protection — which is the one that is unit-tested.

### A real failure this surfaced, and the end-to-end test it produced

The full suite turned up a genuine failure — not a flake:

```
---- session_resume_headless_active_live_tmux_skips_restart_and_attach stdout ----
tm: session 'tm-r-01' is provisioning but its runtime has exited — every pane ...
tm: cannot restart session 'tm-r-01': workspace directory ... no longer exists
```

That test builds its fixture with a bare `tmux new-session -d`, which lands the pane on a **login shell** — i.e. since #3873 it is a *dead runtime*. The fixture had stopped expressing the live-session case the test asserts about. It now runs an explicit long-lived command.

That gap was worth closing in the other direction too, since the critic had independently checked whether `SessionManager::resume`'s "re-attach to live pane" branch would make this a silent no-op. **`session_resume_headless_dead_runtime_reconciles_and_restarts`** now drives the dead-runtime shape end-to-end against a real daemon and asserts the record reached `active` — which only a `/runtime-stop` + `/resume` round-trip can do. Reverting the one-line fix makes it fail deterministically:

```
  left: Some("provisioning")
 right: Some("active")
... the CLI attached into the idle shell instead, i.e. the #3873 defect is back
```

Two fixture hazards were fixed while doing this, both of which had made the tests non-deterministic:

- A freshly created pane **transiently reports a live child** while its shell runs rc files, so a single probe was a statement about the developer's dotfiles rather than the fixture. Both tests now wait for three agreeing reads, and **panic** at the deadline rather than silently proceeding on an unsettled fixture.
- Both tests create sessions in the **machine-global tmux namespace** under names derived from the repo URL, and this crate compiles the same sources into two bin targets that cargo may run concurrently — so one process's unconditional `kill-session` cleanup could destroy the other's session mid-test. Names are now process-unique.

I diagnosed this by capturing the actual failure output rather than assuming; my first hypothesis (name collision) was wrong, and the captured `resumed ... [provisioning]` line in 0.07s is what identified the transient-child race.

### Round-2 evidence

```
$ cargo test -p trusty-mpm            # no --lib
TOTAL passed=6717 failed=0
```

Determinism, after the fixture fixes — 8 consecutive full-suite runs:

```
full-suite run 1..8: 0 failing
```

`cargo clippy -p trusty-mpm --all-targets -- -D warnings` → exit 0.
`cargo fmt -p trusty-mpm --check` → exit 0.
`scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`.

All pure-function tests remain free of globals, `#[ignore]`, env gates, and early returns. The two integration tests use per-test isolated temp roots and their own uniquely-named tmux sessions, touch no process-global env var, and clean up unconditionally.

Still **not merged**, and the reaper defect is still explicitly out of scope and untouched.

### Round-2 CI follow-up

The first round-2 push went red on CI — a genuine finding, not infrastructure. The new end-to-end test asserted the record reached `active`, which *additionally* requires the daemon to successfully SPAWN a runtime. CI has no agent binary on PATH, so `/resume` legitimately ends in `errored` there:

```
tm: session 'tm-deadrt30598-01' restarted but the runtime failed to start.
... restarted but runtime failed to start (state=errored)
```

Note what that output proves: the reconcile path *did* run in CI — `/runtime-stop` and `/resume` both landed. The test was failing on an outcome downstream of its own claim.

The claim is the branch SELECTION, and its discriminator is environment-independent: `Attach` performs **no daemon round-trip at all**, so the seeded `provisioning` survives untouched, while reconcile-then-restart necessarily moves the record off it (to `stopped`, then `active` or `errored`). The assertion is now `!= "provisioning"`, which stays sharp — reverting the one-line fix still fails it deterministically:

```
assertion `left != right` failed: ... i.e. the #3873 defect is back
  left: "provisioning"
 right: "provisioning"
```

**All 23 checks now pass**, with every new test confirmed running in CI:

```
test commands::managed::tests::session_resume_headless_active_live_tmux_skips_restart_and_attach ... ok
test commands::managed::tests::session_resume_headless_dead_runtime_reconciles_and_restarts ... ok
test tests_behavior_c::guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts ... ok
test tests_behavior_c::guided_resume_plan_provisioning_live_tmux_dead_runtime_reconciles_then_restarts ... ok
test tests_behavior_c::parse_pane_probes_reads_every_pane ... ok
test tests_behavior_c::parse_pane_probes_rejects_non_numeric_pid ... ok
test tests_behavior_c::parse_pane_probes_rejects_missing_pid ... ok
test tests_behavior_c::parse_pane_probes_rejects_smeared_row ... ok
test tests_behavior_c::parse_pane_probes_rejects_extra_fields ... ok
test tests_behavior_c::parse_pane_probes_rejects_empty_command_field ... ok
test tests_behavior_c::parse_pane_probes_rejects_empty_or_blank_listing ... ok
test tests_behavior_c::session_panes_any_live_pane_keeps_session_live ... ok
test tests_behavior_c::session_panes_all_idle_reads_dead ... ok
test tests_behavior_c::session_panes_foreign_rows_never_prove_our_session_dead ... ok
test tests_behavior_c::session_runtime_live_assumes_live_for_blank_session_name ... ok
```

---

### KNOWN LIMITATION — read this before "fixing" the any-pane-live question

**A split session whose agent pane died, but which still has some other pane running a non-shell command, is NOT fixed by this PR.** Bare `tm` will still `Attach` the operator into the idle agent pane there, exactly as #3873 describes.

Concretely, #3873 **is** fixed for:
- a single-pane managed session whose runtime exited (the reported case), and
- a split session in which every pane has fallen back to an idle shell.

It is **not** fixed for:
- a split session where the agent pane is idle but a sibling pane runs `vim`, `tail -f`, a stray `node`, or anything else outside `orphan_gc::IDLE_SHELL_COMMANDS`. That session reads as live, and `Attach` is chosen.

This is the direct consequence of the any-pane-live design, and it is a deliberate trade rather than an oversight. Any-pane-live is what makes the destructive branch safe (#2463: convicting on one pane would kill a session whose sibling still runs the agent), and the same widening is what bounds the fix. Escaping the trade requires **per-pane runtime identity** — knowing *which* pane the agent owned and judging only that one — and the stored `pane_id` cannot supply that reliably enough to gate a `kill_session` on, since it can be stale or reused across a tmux server restart.

The asymmetry is what settles it: the unfixed case costs an operator one extra `tm`, which is today's already-shipped behaviour. Getting it wrong in the other direction destroys a live session. **Do not narrow the question back to a single pane without solving the pane-identity problem first** — that is the regression this design guards against.

The same note is recorded in the `guided_resume` module doc and the CHANGELOG, so it is discoverable from the code and not only from this PR.

### MEDIUM (fixed): the membership guard was pinned by no test

Confirmed the finding by running Mutation A myself — delete the guard, change nothing else:

```
MUTATION A APPLIED: membership guard deleted
test result: FAILED. 1282 passed; 2 failed; 1 ignored
```

...which is *after* the new tests. Before them, Mutation A left the suite fully green, exactly as reported. `session_panes_foreign_rows_never_prove_our_session_dead` asserted a property of its own **input** (`!panes.iter().any(...)`) and never called the guarded function — a test named for a guard it does not exercise. Having found that shape once already today, I take the point.

I also verified the hazard is reachable rather than theoretical, since that is what justifies the guard's existence:

```
$ tmux new-session -d -s tm3873probe-X "sh"; tmux display-message -p -t tm3873probe-X '#{pane_current_command}'
bash
```

tmux prefix-matches session targets, so a record named `tm-apex` receives `tm-apex-01`'s panes from both `has-session` and `list-panes -s`. Without the guard, `session_has_live_pane`'s filter matches nothing → `false` → read as "dead" → `kill_session`, which then prefix-matches onto that same live session.

**The fix:** extracted `panes_prove_session_dead(session_name, panes, probe)` as a pure seam. It is phrased as *"prove DEAD"* rather than *"is live"* on purpose — the burden of proof belongs on the destructive answer, so every uncertain input falls out as `false` by construction rather than depending on a caller to remember an inversion.

Five tests replace the one tautological test:

```
test panes_prove_session_dead_when_every_pane_is_a_bare_shell ... ok
test panes_prove_session_dead_refuses_when_a_sibling_pane_is_live ... ok
test panes_prove_session_dead_refuses_a_prefix_matched_listing ... ok
test panes_prove_session_dead_refuses_a_wholly_foreign_listing ... ok
test panes_prove_session_dead_ignores_foreign_rows_beside_our_own ... ok
test result: ok. 5 passed; 0 failed
```

The last one guards the *over-correction*: a foreign live pane must not rescue a session whose own panes are all idle. RED under Mutation A is the prefix-match and foreign-listing pair; the other three stay green, which is what makes them a guard-specific pin rather than a blunt tripwire.

### Comment correction

`managed_tests.rs` reasoned about `sh` while tmux launches the argument through its default-shell — `pane_current_command` is `bash`, verified empirically above. Outcome is identical (both are in `IDLE_SHELL_COMMANDS`), but the comment described a process that is not what runs. Corrected, including *why* the substitution is harmless.

### Round-3 evidence

```
$ cargo test -p trusty-mpm            # no --lib
TOTAL passed=6745 failed=0            # 3 consecutive runs, 0 failing
```

`cargo clippy -p trusty-mpm --all-targets -- -D warnings` → exit 0
`cargo fmt -p trusty-mpm --check` → exit 0
`scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`
`scripts/check_test_pointers.sh` → `0 dangling pointers — OK` (exit 0)

Isolation re-verified after the change — 3 further integration runs:

```
tmux sessions before=11 after=11
0 leaked fixture sessions
```

Branch is current with `origin/main` (0 commits behind). Still **not merged**; reaper defect still untouched; no scope added beyond the two requested items and the one comment correction.

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -183,6 +183,29 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   unchanged (unset or unregistered still means worktrees ON, and an unreadable
   registry fails safe to ON), and #3455's warn-never-refuse concurrency
   behaviour is untouched.
+- Bare `tm` now resumes a session whose runtime has died, instead of switching
+  the operator into a bare shell
+  ([#3873](https://github.com/bobmatnyc/trusty-tools/issues/3873)). The picker's
+  resume planner decided "this session is live" from tmux SESSION existence
+  alone, which cannot see the common case where the tmux pane survives but the
+  `claude` inside it exited and the pane fell back to a login shell. That
+  `(active, tmux up)` pair fell through to `Attach`, so the first `tm` did a
+  `switch-client` into an idle shell and nothing else — the operator had to type
+  `tm` a SECOND time, from inside that pane, before a runtime came back.
+  `guided_resume::plan_resume` takes a third input, `runtime_live`, and maps
+  `(active/provisioning, tmux up, runtime dead)` onto the existing
+  `ReconcileThenRestart` recovery (`/runtime-stop` → `/resume` → attach), so the
+  first invocation does the whole job. `runtime_live` is resolved by the new
+  `pane_runtime_live`, which reuses the daemon reaper's own
+  `daemon::runtime_reap::pane_runtime_exited` predicate — same `is_idle_shell`
+  allowlist, same `ProcessTreeProbe` child gate — against a single
+  pane-id-targeted `tmux display-message`, so the CLI and the reaper cannot drift
+  apart and the fix works whether or not the 60s reaper is healthy. Fails OPEN:
+  a record with no `pane_id`, a failed `tmux` call, or an unparseable reply all
+  assume the runtime is live and preserve the previous attach behavior, because a
+  wrong "dead" verdict would kill a live agent's pane. Stopped/Errored records
+  (whose panes are idle shells too) keep taking the plain `Restart` path, and the
+  decommissioned/deleted tombstone refusal still runs first.
 
 - Concurrent `tm` processes no longer lose (or corrupt) `projects.json` entries.
   The project-store upsert was an unsynchronised read-modify-write of the whole
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -216,6 +216,15 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   non-destructive reconcile (#2023 A) but not for a CLI path that drives a
   session kill.
 
+  Known limitation, deliberately accepted: any-pane-live is also what bounds the
+  fix. A SPLIT session whose agent pane died but which still has some other pane
+  running a non-shell command (a `vim`, a `tail -f`) reads as live, so bare `tm`
+  still attaches into the idle agent pane there. #3873 is fixed for the
+  single-pane case and for splits whose panes are all idle shells, but not for a
+  split with an unrelated live process — narrowing the question back to one pane
+  would require per-pane runtime identity that the stored `pane_id` cannot supply
+  reliably enough to gate a session kill on.
+
   Because that reconcile reaches `SessionManager::stop` →
   `graceful_terminate_runtime` → `kill_session`, which is **session-scoped**, the
   restart now discloses that it is about to destroy the whole tmux session — all
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -196,16 +196,47 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `(active/provisioning, tmux up, runtime dead)` onto the existing
   `ReconcileThenRestart` recovery (`/runtime-stop` → `/resume` → attach), so the
   first invocation does the whole job. `runtime_live` is resolved by the new
-  `pane_runtime_live`, which reuses the daemon reaper's own
-  `daemon::runtime_reap::pane_runtime_exited` predicate — same `is_idle_shell`
-  allowlist, same `ProcessTreeProbe` child gate — against a single
-  pane-id-targeted `tmux display-message`, so the CLI and the reaper cannot drift
-  apart and the fix works whether or not the 60s reaper is healthy. Fails OPEN:
-  a record with no `pane_id`, a failed `tmux` call, or an unparseable reply all
-  assume the runtime is live and preserve the previous attach behavior, because a
-  wrong "dead" verdict would kill a live agent's pane. Stopped/Errored records
+  `session_runtime_live`, which reuses the daemon reaper's own
+  `daemon::runtime_reap::session_has_live_pane` primitive — and through it
+  `pane_runtime_exited`, the same `is_idle_shell` allowlist, and the same
+  `ProcessTreeProbe` child gate — against a session-scoped `tmux list-panes`, so
+  the CLI and the reaper cannot drift apart and the fix works whether or not the
+  60s reaper is healthy.
+
+  The verdict is the ANY-PANE-LIVE question over the whole tmux session, not a
+  check of the record's stored `pane_id`: a managed session can be manually split
+  (#2463), and a session with one idle pane beside a live agent pane must never
+  reach the destructive branch. Fails OPEN throughout — a blank session name, a
+  failed `tmux` call, a listing that cannot be fully parsed, or a listing with no
+  pane attributable to the session all assume the runtime is live and preserve
+  the previous attach behavior. Parsing is strict (exactly three non-empty
+  tab-separated fields with a numeric pid) specifically so a missing or
+  unparseable pid cannot collapse the two-gate check down to `is_idle_shell`
+  alone: `has_live_child(None)` returns `false`, which is right for the daemon's
+  non-destructive reconcile (#2023 A) but not for a CLI path that drives a
+  session kill.
+
+  Because that reconcile reaches `SessionManager::stop` →
+  `graceful_terminate_runtime` → `kill_session`, which is **session-scoped**, the
+  restart now discloses that it is about to destroy the whole tmux session — all
+  windows, panes, and scrollback — whenever tmux is live, matching what the plain
+  `Restart` branch already did for the same situation. Stopped/Errored records
   (whose panes are idle shells too) keep taking the plain `Restart` path, and the
   decommissioned/deleted tombstone refusal still runs first.
+
+  `commands::managed_tests`: `session_resume_headless_active_live_tmux_skips_restart_and_attach`
+  created its scratch tmux session with a bare `tmux new-session -d`, which lands
+  the pane on a login shell — since #3873 that is a DEAD runtime, so the fixture
+  no longer expressed the live-session case the test asserts about. It now runs
+  an explicit long-lived command, and a new counterpart
+  (`session_resume_headless_dead_runtime_reconciles_and_restarts`) drives the
+  dead-runtime shape end-to-end against a real daemon, proving the reconcile
+  actually reaches `/runtime-stop` + `/resume` rather than being a silent no-op.
+  Both wait for the pane's liveness reading to STABILISE (three agreeing probes)
+  before acting: a freshly created pane transiently reports a live child while
+  the shell initialises, which otherwise made both tests depend on the
+  developer's dotfiles. Their tmux names are now process-unique, since the two
+  bin targets can run concurrently and share the machine-global tmux namespace.
 
 - Concurrent `tm` processes no longer lose (or corrupt) `projects.json` entries.
   The project-store upsert was an unsynchronised read-modify-write of the whole
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs
@@ -1,4 +1,4 @@
-//! Guided-picker resume/restart flow (#1742, #2001, #3531).
+//! Guided-picker resume/restart flow (#1742, #2001, #3531, #3873).
 //!
 //! Why: resuming a managed session from the bare-`tm` picker must handle every
 //! runtime state without ever exposing a raw tmux failure. A live session is
@@ -8,6 +8,22 @@
 //! stops it (resetting the record to Stopped) then restarts it via the same
 //! path, so the operator does nothing (#2001). Extracted from `guided.rs` to keep
 //! both files under the 500-SLOC production cap.
+//!
+//! #3873: "live session" was originally decided by tmux SESSION existence alone
+//! (`tmux_has_session`), which cannot see the far more common failure the
+//! operator actually hits — the tmux pane is still there, but the `claude`
+//! inside it exited and the pane fell back to a bare login shell. Bare `tm`
+//! therefore picked `Attach` and `switch-client`ed the operator into an idle
+//! shell, doing nothing useful; only a SECOND `tm`, typed from inside that pane
+//! (where the in-place relaunch gate matches), actually brought a runtime back.
+//! [`plan_resume`] now takes a third input, `runtime_live`, and maps
+//! `(active, tmux up, runtime dead)` onto the existing `ReconcileThenRestart`
+//! recovery so the first invocation does the whole job. [`pane_runtime_live`]
+//! resolves that input by reusing the daemon reaper's own
+//! [`trusty_mpm::daemon::runtime_reap::pane_runtime_exited`] predicate against a
+//! single pane-id-targeted `tmux display-message` — the CLI asks the question
+//! itself at resume time, so this fix does NOT depend on the 60s runtime reaper
+//! having already reconciled the record.
 //!
 //! #3531: the zombie/restart classification below MUST run on
 //! [`trusty_mpm::client::ManagedSessionSummary::persisted_state`], never the
@@ -135,21 +151,6 @@ pub(crate) enum ResumeAction {
     Terminal,
 }
 
-/// Decide, purely from `state` and `tmux_live`, how a guided resume should proceed.
-///
-/// Why: single source of truth for the resume branch selection so the zombie
-/// reconcile, the plain restart, and the happy-path attach can never diverge
-/// from what [`is_zombie`]/[`needs_restart`] classify. Pure so it is exhaustively
-/// unit-tested.
-/// What: TERMINAL (`decommissioned`/`deleted`, checked FIRST) → `Terminal`
-/// (refuse); zombie (checked next, since a zombie's state also fails
-/// `needs_restart`) → `ReconcileThenRestart`; Stopped/Errored → `Restart`;
-/// everything else (active/provisioning with a live tmux) → `Attach`. The
-/// terminal check is first so neither the zombie path (would resurrect via
-/// reconcile-restart) nor the Attach path (would attach into a deleted session)
-/// can ever be reached for a tombstone (code-critic CRITICAL).
-/// Test: `guided_resume_plan_*`, `plan_resume_refuses_terminal_states` in
-/// `tests_behavior_c_tests.rs`.
 /// Resolve the state string [`plan_resume`] must classify from a wire
 /// [`trusty_mpm::client::ManagedSessionSummary`] (#3531).
 ///
@@ -173,16 +174,156 @@ pub(crate) fn resume_classification_state(
     session.persisted_state.as_deref().unwrap_or(&session.state)
 }
 
-pub(crate) fn plan_resume(state: &str, tmux_live: bool) -> ResumeAction {
+/// Decide, purely from `state` + `tmux_live` + `runtime_live`, how a guided
+/// resume should proceed.
+///
+/// Why: single source of truth for the resume branch selection so the zombie
+/// reconcile, the plain restart, and the happy-path attach can never diverge
+/// from what [`is_zombie`]/[`needs_restart`] classify. Pure so it is exhaustively
+/// unit-tested.
+///
+/// #3873: `tmux_live` answers "does the tmux SESSION exist?", which is NOT the
+/// same question as "is the agent runtime alive?". A managed session whose inner
+/// `claude` exited leaves the tmux pane alive as a bare login shell, so
+/// `(active, tmux_live = true)` used to fall straight through to `Attach` — bare
+/// `tm` switched the operator into an idle shell and did nothing useful. The
+/// operator then had to type `tm` a SECOND time (now inside that pane, where the
+/// in-place relaunch gate matches) to actually get a runtime back. `runtime_live`
+/// is the missing third input: when the record is active/provisioning, tmux is
+/// up, but the runtime is provably gone, the plan is `ReconcileThenRestart` —
+/// the SAME auto-stop-then-restart recovery the tmux-gone zombie already uses,
+/// so the first invocation does the whole job. See [`pane_runtime_live`] for how
+/// the caller resolves this input (it reuses the daemon reaper's own
+/// [`trusty_mpm::daemon::runtime_reap::pane_runtime_exited`] predicate, so the
+/// CLI and the reaper can never disagree on what "dead runtime" means), and note
+/// that it fails OPEN (`true`) whenever liveness cannot be proven — an
+/// unprovable case keeps the pre-#3873 `Attach` behavior rather than tearing a
+/// pane down on a guess.
+///
+/// What: TERMINAL (`decommissioned`/`deleted`, checked FIRST) → `Terminal`
+/// (refuse); zombie (checked next, since a zombie's state also fails
+/// `needs_restart`) → `ReconcileThenRestart`; Stopped/Errored → `Restart`;
+/// active/provisioning with a live tmux but a DEAD runtime (#3873) →
+/// `ReconcileThenRestart`; everything else (active/provisioning with a live tmux
+/// AND a live runtime) → `Attach`. The terminal check is first so neither the
+/// zombie path (would resurrect via reconcile-restart) nor the Attach path
+/// (would attach into a deleted session) can ever be reached for a tombstone
+/// (code-critic CRITICAL). The `needs_restart` check stays AHEAD of the #3873
+/// gate so a Stopped/Errored record — whose pane is an idle shell too, and would
+/// therefore also report `runtime_live = false` — keeps taking the plain
+/// `Restart` path rather than a pointless `/runtime-stop` round-trip against a
+/// record that is already stopped.
+/// Test: `guided_resume_plan_*`, `plan_resume_refuses_terminal_states` in
+/// `tests_behavior_c_tests.rs`;
+/// `guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts`
+/// pins the #3873 case specifically.
+pub(crate) fn plan_resume(state: &str, tmux_live: bool, runtime_live: bool) -> ResumeAction {
     if is_terminal_state(state) {
         ResumeAction::Terminal
     } else if is_zombie(state, tmux_live) {
         ResumeAction::ReconcileThenRestart
     } else if needs_restart(state) {
         ResumeAction::Restart
+    } else if !runtime_live {
+        // #3873: active/provisioning, tmux session up, but the runtime inside the
+        // pane is provably gone. Same recovery as the tmux-gone zombie above —
+        // `/runtime-stop` resets the record to Stopped so the daemon's `/resume`
+        // (which only accepts Stopped/Errored) will take it, then attach.
+        ResumeAction::ReconcileThenRestart
     } else {
         ResumeAction::Attach
     }
+}
+
+/// Parse `tmux display-message -p -t <pane> '#{pane_current_command} #{pane_pid}'`
+/// output into the two fields [`pane_runtime_exited`] needs (#3873).
+///
+/// Why: the shell-out itself is untestable without a live tmux server, but the
+/// parse is exactly where a format change or an empty/short reply would silently
+/// mis-answer "is the runtime alive?". Splitting it out keeps that risk on a pure,
+/// unit-tested seam. Deliberately does NOT go through
+/// [`crate::daemon::tmux`]-style `list-panes -a` enumeration: that path is a
+/// whole-server sweep whose composite-key parsing is a known failure mode
+/// (#1813 shape), and a single pane-id-targeted `display-message` needs none of
+/// it — tmux resolves a pane id to exactly one pane or errors.
+/// What: returns `None` when the reply has no command word at all (empty output,
+/// or tmux printed nothing usable) — the caller treats that as "cannot prove
+/// dead". Otherwise returns the command word plus the pane pid, with the pid
+/// `None` when it is missing or not a number (that alone is not evidence of
+/// death; [`crate::daemon::orphan_gc::ChildLivenessProbe`] treats `None` as
+/// "no live child", and the command-word gate still has to agree).
+/// Test: `parse_pane_probe_*` in `tests_behavior_c_tests.rs`.
+pub(crate) fn parse_pane_probe(stdout: &str) -> Option<(String, Option<u32>)> {
+    let line = stdout.lines().next().unwrap_or("").trim();
+    let mut parts = line.split_whitespace();
+    let command = parts.next()?.to_string();
+    let pid = parts.next().and_then(|p| p.parse::<u32>().ok());
+    Some((command, pid))
+}
+
+/// Is the agent runtime inside `pane_id` still alive? (#3873)
+///
+/// Why: `tmux_has_session` proves only that the tmux SESSION exists. When a
+/// managed session's inner `claude` exits, the pane survives as a bare login
+/// shell — so session-existence says "live" while the thing the operator wants
+/// is gone. The daemon's 60s runtime reaper already owns the authoritative
+/// answer to this question in
+/// [`trusty_mpm::daemon::runtime_reap::pane_runtime_exited`]; this function
+/// reuses that exact predicate (and therefore the same `is_idle_shell`
+/// allowlist and the same [`trusty_mpm::daemon::orphan_gc::ProcessTreeProbe`]
+/// belt-and-braces child gate) rather than reimplementing idle-shell detection
+/// in the CLI, so the two can never drift apart. Reusing it also means this fix
+/// works whether or not the reaper is currently healthy: the CLI asks the
+/// question itself, at resume time, instead of waiting for a reaped record.
+/// What: FAILS OPEN — returns `true` (assume live, preserving the pre-#3873
+/// `Attach` behavior) when the record has no `pane_id`, when the `tmux` shell-out
+/// fails or exits non-zero, or when the reply cannot be parsed. Only a positive,
+/// two-gate identification (a recognised bare shell AND no live child under the
+/// pane pid) returns `false`. Failing open matters because `false` now triggers a
+/// `/runtime-stop` that KILLS the pane: a wrong `false` would destroy a live
+/// agent's pane, while a wrong `true` merely reproduces today's already-shipped
+/// behavior.
+/// Test: the pure parse seam is `parse_pane_probe_*`; the fail-open contract for
+/// a missing pane id is `pane_runtime_live_assumes_live_without_pane_id`; the
+/// classification it delegates to is covered by the daemon's own
+/// `pane_runtime_exited_*` tests. The tmux shell-out itself is I/O (a live tmux
+/// server), matching the existing `session_for_pane` / `pane_tty_for` precedent
+/// in `tmux_attach.rs`.
+pub(crate) fn pane_runtime_live(pane_id: Option<&str>) -> bool {
+    let Some(pane_id) = pane_id.map(str::trim).filter(|p| !p.is_empty()) else {
+        return true;
+    };
+    let tmux_bin = trusty_mpm::core::tmux::resolve_tmux_binary_or_bare();
+    let Ok(output) = std::process::Command::new(&tmux_bin)
+        .args([
+            "display-message",
+            "-p",
+            "-t",
+            pane_id,
+            "#{pane_current_command} #{pane_pid}",
+        ])
+        .output()
+    else {
+        return true;
+    };
+    if !output.status.success() {
+        return true;
+    }
+    let Some((command, pane_pid)) = parse_pane_probe(&String::from_utf8_lossy(&output.stdout))
+    else {
+        return true;
+    };
+
+    let pane = trusty_mpm::daemon::orphan_gc::PaneInfo {
+        session_name: String::new(),
+        pane_current_command: command,
+        pane_pid,
+        pane_id: Some(pane_id.to_string()),
+    };
+    !trusty_mpm::daemon::runtime_reap::pane_runtime_exited(
+        &pane,
+        &trusty_mpm::daemon::orphan_gc::ProcessTreeProbe,
+    )
 }
 
 /// Restart (if needed) then attach to a managed session from the guided picker.
@@ -286,7 +427,15 @@ pub(crate) async fn resume_session(
     // the pre-#3531 (imperfect but not regressed) behavior — see
     // `resume_classification_state`.
     let persisted_state = resume_classification_state(session);
-    let action = plan_resume(persisted_state, tmux_live);
+    // #3873: `tmux_live` above proves only that the tmux SESSION exists. Ask the
+    // separate question the operator actually cares about — is the agent runtime
+    // inside the record's pane still there? — before letting an `active` record
+    // fall through to a bare `Attach` that would drop them into an idle shell.
+    // Only worth the shell-out when tmux is up at all; when it is not, `is_zombie`
+    // already classifies the record and `pane_runtime_live` could not answer
+    // anyway (the pane is gone with the session).
+    let runtime_live = !tmux_live || pane_runtime_live(session.pane_id.as_deref());
+    let action = plan_resume(persisted_state, tmux_live, runtime_live);
 
     // Terminal-state refusal (code-critic CRITICAL): a decommissioned/deleted
     // tombstone must never be attached to or restarted — doing either would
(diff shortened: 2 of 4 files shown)


bobmatnyc:
## Verdict: WARN

Independent adversarial review of `fix/3873-picker-dead-runtime` @ `a04630f4` against `origin/main`. Reviewed against the spec + code + test results; implementer reasoning was checked, not trusted.

The core fix is correct, minimal, and genuinely test-pinned — I reproduced the RED myself under three separate mutations. Two findings, one HIGH, keep this out of a clean APPROVE. Neither is a CRITICAL, so this is WARN: it can proceed with the findings tracked, but finding 1 is a new destructive path that ships without disclosure and I would rather see it fixed in this PR than tracked.

---

## Verification of the four claims

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | RED before / GREEN after | **YES — reproduced independently** | See "Red under mutation" below. Exact `left: Attach / right: ReconcileThenRestart`, `8 passed; 2 failed`. |
| 2 | `guided_resume_plan_active_live_tmux_attaches` kept, updated to `runtime_live = true` | **YES** | `tests_behavior_c_tests.rs` — the test is *modified*, not deleted; asserts `plan_resume("active", true, true) == Attach`. A live agent still attaches. |
| 3 | No new seam; daemon predicate genuinely reused | **YES** | `git diff --stat origin/main...HEAD` = 3 files (`CHANGELOG.md`, `guided_resume.rs`, `tests_behavior_c_tests.rs`). `git diff origin/main...HEAD -- src/daemon/ src/lib.rs` is **empty** — no visibility widened, `orphan_gc.rs`/`tmux.rs` untouched. `guided_resume.rs:326` *calls* `trusty_mpm::daemon::runtime_reap::pane_runtime_exited` — no reimplementation of `is_idle_shell`. Precedent confirmed at `bin/tm/gh_identity.rs:119`. |
| 4 | CI green; `cargo test -p trusty-mpm` without `--lib` clean | **YES** | 23/23 checks pass (1 `skipping` = `Notify on red main`, correct). My local run: **6701 passed; 0 failed; 16 ignored** across every target — lib **3894**, bin `tm` **1267**. Matches the reported numbers exactly. |

### Red under mutation (my own run, not the author's)

Fix branch deleted, parameter left threaded (`let _ = runtime_live;`):

```
thread '...guided_resume_plan_active_live_tmux_dead_runtime_reconciles_then_restarts' panicked at
crates/trusty-mpm/src/bin/tm/tests_behavior_c_tests.rs:1272:5:
  left: Attach
 right: ReconcileThenRestart
...
test result: FAILED. 8 passed; 2 failed; 0 ignored; 0 measured; 1258 filtered out
```

I also mutated the two **orderings** the PR claims are deliberate, because a test named for a guard that never exercises it is a known failure mode in this repo. Both are genuinely pinned:

```
# moved `!runtime_live` ABOVE needs_restart:
...guided_resume_plan_stopped_live_tmux_dead_runtime_still_plain_restarts panicked:
  left: ReconcileThenRestart / right: Restart
test result: FAILED. 9 passed; 1 failed

# moved `!runtime_live` ABOVE the terminal check:
...plan_resume_refuses_terminal_states_even_with_dead_runtime panicked (right: Terminal)
test result: FAILED. 1 passed; 1 failed
```

Test quality otherwise confirms as claimed: all new tests are pure, no `#[ignore]`, no env var, no `serial_test`, no global state; `pane_runtime_live(None)` and `pane_runtime_live(Some("   "))` both return at `guided_resume.rs:293` before any shell-out.

### Closing-keyword check

`gh pr view 4304 --json closingIssuesReferences` queried **twice**, several seconds apart — both times exactly `3873` and nothing else. #4061, #4065, #1813 are referenced in prose only and did **not** register. Commit message body contains exactly one closing keyword, `Closes #3873`, and no `Closes` inside a code span. Clean.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs` | 460-495 | New `(active, tmux up, runtime dead)` path **kills the entire tmux session** — including sibling windows/panes — and the PR *removes* the pane-kill disclosure for it. Previously this input did a harmless `Attach`. | Print the existing pane-kill warning when `tmux_live` is true on the reconcile branch, as the `Restart` branch already does. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs` | 256-262, 292-330 | A missing/unparseable `pane_pid` silently collapses the documented **two-gate** liveness check to one gate, so an unproven case can still return the destructive `false`. | Fail open when the pid is absent: `let Some((command, Some(pane_pid))) = parse_pane_probe(...) else { return true; };` |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs` | 317-318 | The probe never verifies the polled pane belongs to this session, and inspects only the recorded pane rather than "any live pane in the session" (`session_has_live_pane`, #2463). | Add `#{session_name}` to the format string; return `true` unless it equals `session.name`. |

### 1. HIGH — the reconcile branch now kills a live tmux session with no disclosure

`guided_resume.rs:460-478` prints, for the new `tmux_live == true` shape:

```
tm: '{}' is marked {} but its runtime has exited (the pane is a bare shell) — restarting the runtime…
```

…then calls `reconcile_zombie_stop` → `POST /runtime-stop`. I traced what that endpoint actually does, because `reconcile_zombie_stop`'s own doc only says it "marks the record `Stopped` synchronously (keeping the workspace)":

- `daemon/managed_routes/mod.rs` → `stop_managed_session_runtime` → `SessionManager::stop`
- `session_manager/manager.rs:554,570` → `graceful_terminate_runtime`
- `session_manager/restart_ops.rs:150` → **`self.tmux.kill_session(tmux_name)`**

`kill_session` is session-scoped: it destroys **every window and pane** in that tmux session, not just the recorded pane. For the #2001 zombie this was free — the tmux session was already gone. For the new #3873 shape the tmux session is **alive by definition**, so this is a brand-new destructive step on an input that previously took `Attach`.

Meanwhile `guided_resume.rs:490-495` removes the warning, reasoning:

```rust
// Pane-kill disclosure only applies to the plain Restart branch. The
// reconcile branch already printed its own message above, and its pane is
// either gone (#2001) or a bare shell whose runtime already exited
// (#3873) — in neither case is there in-progress runtime state left to
// warn about losing.
```

That reasoning accounts only for the *recorded* pane. This codebase states twice that a managed tmux session can carry more panes than the one the record points at — `runtime_reap.rs:98` (`session_has_live_pane`, #2463: *"a manually-split managed session can have more than one pane sharing the same `tmux_name`"*) and the `#2456` sibling-window-hijack note at `manager.rs:842-847`. So the reachable case is concrete: the agent's pane has dropped to a shell, the operator's *other* window in that session is running a build/editor, they press Enter on the picker row, and that window dies with no warning shown.

The asymmetry is the tell: for the **identical** `tmux_live == true` situation the plain `Restart` branch two lines below (`:497-504`) *does* warn — *"the daemon will KILL that pane and start a fresh runtime (in-progress pane state will be lost)."* The reconcile branch now reaches the same `kill_session` and says nothing. This is also the exact hazard #2023 A was filed for (*"destroying a pane a human might still be looking at ~60s after their `claude` process exited"*).

Partially mitigating, and why this is HIGH and not CRITICAL: `SessionManager::stop` calls `snapshot::capture_into` before terminating, so the dead agent's own scrollback is preserved into the record; the action is operator-initiated; and `SIGTERM` + a grace window precede the kill. The loss is sibling-pane work plus the missing disclosure — not unauthorized destruction.

**Fix:** restore the disclosure on the reconcile branch when `tmux_live` is true. One `if`, no logic change.

### 2. MEDIUM — a missing `pane_pid` disables the second gate, in a context where the daemon's calibration no longer holds

`pane_runtime_live`'s doc (`:283-288`) states the contract precisely:

> Only a positive, two-gate identification (a recognised bare shell AND no live child under the pane pid) returns `false`.

That is not what the code does when the pid is absent. `parse_pane_probe` (`:256-262`) returns `Some((command, None))` whenever the second token is missing or non-numeric — only a *completely* empty reply yields `None`. `guided_resume.rs:317-321` then builds `PaneInfo { pane_pid: None, .. }`, and `orphan_gc.rs:439-443`:

```rust
fn has_live_child(&self, pane_pid: Option<u32>) -> bool {
    let Some(pid) = pane_pid else {
        // No process tree to inspect; the idleness gate already ran.
        return false;
    };
```

`None → false` ("no live child"), so `pane_runtime_exited` degenerates to `is_idle_shell(cmd)` alone. A reply of `"zsh"` or `"zsh notanumber"` therefore returns `runtime_live = false` and triggers the kill on **one** gate. The repo's own test `parse_pane_probe_tolerates_missing_pid` asserts this shape and its comment claims *"a missing/garbage pid is not evidence of death"* — but for a shell command word it is treated as exactly that.

Why this is the crux rather than a nit: `None → false` is *correct* for the daemon, because #2023 A deliberately made the reaper's action **non-destructive** (`mark_runtime_exited_stopped` — record flips, pane survives). The predicate's fail-closed calibration was tuned to that weak consequence. The CLI reuses the predicate to drive `kill_session`. Reusing the daemon's gate is the right call for consistency, but the CLI must be *stricter* than the daemon here, not equal to it — its consequence is strictly worse.

Likelihood is low (tmux 3.6b reliably emits `#{pane_pid}`; I verified `[zsh 93112]` against a live pane), which is why this is MEDIUM and not HIGH. But the fix is one line and removes the whole class.

### 3. MEDIUM — the probed pane is not verified to belong to the session

`guided_resume.rs:317-318` constructs `PaneInfo { session_name: String::new(), .. }` — the session-name field the daemon's own path filters on is blanked. The CLI asks "is the pane recorded at spawn time idle?" where the daemon asks "does this session have *any* live pane?" (`runtime_reap.rs:98`, added specifically because single-pane checks misclassify split sessions). Consequences: (a) if the agent was relaunched into a different pane of the same session, the stale recorded pane reads as an idle shell and the live session gets torn down; (b) a stale `pane_id` is not proven to belong to this session at all.

Confidence that (a)/(b) are *reachable today* is below my bar — `resume` refreshes `pane_id` at `manager.rs:894`, the runtime-exit reconcile refreshes it, and tmux never reuses a `%N` within a server lifetime — so I am flagging the missing guard, not asserting a live bug. Adding `#{session_name}` to the format and requiring equality with `session.name` costs one token and closes it permanently.

---

## Attacks that came back clean

- **Fail-open path enumeration.** All five paths to `pane_runtime_live == true` verified by reading: no/blank `pane_id` (`:293`), spawn failure (`:305`), non-zero exit (`:308`), unparseable reply (`:311`), and any non-allowlisted command word. The only paths to `false` are the two-gate identification and the pid-`None` degradation in finding 2. **The asymmetry argument is sound and I agree with it** — a wrong `true` reproduces today's shipped bug, a wrong `false` destroys live work, and the code is built the right way round. Finding 2 is a gap in the execution of that principle, not a rejection of it.
- **Empirically probed tmux 3.6b** on a throwaway detached session. Live pane → `[zsh 93112]`, parses correctly. **An invalid target returns `rc=0` with stdout `" "`**, not non-zero:
  ```
  target=%999999   rc=0 stdout=[ ] stderr=[]
  target=%notapane rc=0 stdout=[ ] stderr=[]
  ```
  So the `!output.status.success()` guard at `:308` never fires for a stale `pane_id` — the real protection is `parse_pane_probe` returning `None` on a whitespace-only line, which **is** covered by `parse_pane_probe_none_for_empty_reply`. The safety property holds; only the doc's stated reason is slightly off. Worth knowing, not a finding.
- **Space-separator parse corruption.** Cannot produce a false "dead". A command containing a space shifts the pid token out of position → pid `None`, and the first word must *still* be in the `IDLE_SHELL_COMMANDS` allowlist (`orphan_gc.rs:50-55`). A corrupted `pane_current_command` that is empty yields the pid as the command word → not an idle shell → fails open. The only residue of this attack is finding 2.
- **#1813 independence holds.** Nothing touches `daemon/orphan_gc.rs` or `daemon/tmux.rs`; the diff is 3 files. The targeted `display-message -p -t <pane_id>` follows the existing `tmux_attach.rs:355-356/386-387` precedent verbatim and never enumerates. The fix is genuinely independent of reaper health.
- **False positive on a live agent** is not reachable in the normal case: a running `claude` reports `pane_current_command = claude` (fails gate 1), and even if the pane momentarily showed a shell, `pgrep -P <pane_pid>` finds the agent child (fails gate 2), with `interpret_pgrep` failing closed to `true` on spawn error.
- **Both incidental corrections are accurate and in scope.** (a) The unconditional *"its tmux pane is gone"* was genuinely false for the new shape and now branches on `tmux_live`. (b) Real doc drift: on `origin/main` the `plan_resume` doc block sat immediately above `/// Resolve the state string…`, so rustdoc attached both to `resume_classification_state` — `plan_resume` had no doc at all. Fixing the doc of the function being changed is not scope creep.
- **`/resume` does relaunch the runtime.** I checked this specifically, because `SessionManager::resume` has a "re-attach to live pane, no recreate" branch (`manager.rs:842-853`) that would have made the whole fix a no-op. It does not fire: `/runtime-stop` has already killed the tmux session, so `session_exists` is false and `resume` takes the recreate path, then `resume_managed` re-spawns the adapter. The fix works end to end.

---

## Required Changes

1. **(HIGH)** Restore the pane-kill disclosure on the `ReconcileThenRestart` branch when `tmux_live` is true — `/runtime-stop` reaches `kill_session`, which destroys sibling windows/panes. Update the `:490-495` comment, which is currently incorrect for the #3873 shape.
2. **(MEDIUM)** Fail open when `parse_pane_probe` yields no pid, so the destructive verdict always requires both gates:
   ```rust
   let Some((command, Some(pane_pid))) = parse_pane_probe(&String::from_utf8_lossy(&output.stdout))
   else { return true; };
   ```
   Add a test pinning `"zsh"` / `"zsh notanumber"` → live.
3. **(MEDIUM)** Add `#{session_name}` to the probe format and return `true` unless it equals `session.name`, so a stale or foreign `pane_id` can never drive a teardown.

## Notes

- Items 2 and 3 are cheap and strictly safety-increasing; I would fold them in with item 1 rather than track them separately.
- `pane_runtime_live` shells out for every resume where tmux is up, including terminal-tombstone and `Stopped` records whose verdict is already decided. Read-only and sub-millisecond — noted, not flagged.
- CHANGELOG entry present under `### Fixed` and accurately describes the change.

bobmatnyc:
## Verdict: APPROVE

Round 2 re-review of `fix/3873-picker-dead-runtime` @ `a14328e6`, rebased onto `origin/main` = `c11da9cd`. Treated as a fresh review of a changed liveness question, not a confirm-only pass.

All three round-1 findings are resolved, and finding 3 is resolved *better* than I proposed — I suggested adding a `#{session_name}` equality check to the single-pane probe; they replaced the question instead. Zero CRITICAL, zero HIGH → APPROVE.

One MEDIUM remains, and it is a test-coverage gap on the exact line the PR's load-bearing safety claim rests on. I verified by mutation that the guard is completely unpinned, and I verified empirically that the hazard it prevents is real on this machine. The code is correct today; the risk is a future edit.

---

## Round-1 findings — disposition

**HIGH (silent session kill) — RESOLVED.** `guided_resume.rs:548-559` now discloses on `tmux_live == true`, naming that the kill takes *every window and pane, including their scrollback*. The `reconcile_zombie_stop` doc at `:615-627` is corrected from the record-only phrasing that let this land unnoticed. I re-walked the server path to confirm the disclosure is accurate, not merely present: `stop_managed_session_runtime` → `SessionManager::stop` (`manager.rs:554,570`) → `graceful_terminate_runtime` → `kill_session` (`restart_ops.rs:150`), session-scoped. Wording matches behavior.

**MEDIUM (pid gap) — RESOLVED, and the CLI is now genuinely stricter than the daemon.** `guided_resume.rs:299` — `let pane_pid: u32 = pid.parse().ok()?;` — the `?` propagates `None` out of `parse_pane_probes`, which every caller treats as fail-open. `PaneInfo.pane_pid` can no longer be `None` on this path, so `ProcessTreeProbe::has_live_child(None) → false` is unreachable and the two-gate check cannot collapse to `is_idle_shell` alone. The daemon keeps its own (correct, non-destructive) calibration untouched. Pinned by `parse_pane_probes_rejects_non_numeric_pid` and `_rejects_missing_pid`, both observed `... ok`.

**MEDIUM (membership) — RESOLVED structurally.** `session_runtime_live` (`:359-397`) asks a session-scoped `list-panes -s -t <session>` and delegates to `runtime_reap::session_has_live_pane`, the #2463 primitive the previous round bypassed. The fabricated `session_name: String::new()` is gone.

### The structural claim — verified

*"A session with a live sibling pane cannot reach the destructive branch at all."* **Holds.** `session_has_live_pane` returns `true` if *any* matching pane is not `pane_runtime_exited`, so a single live sibling short-circuits to `runtime_live = true` before any daemon round-trip. There is no separate any-pane-live guard a future edit could forget — the semantics are inherent to the delegated primitive. Pinned by `session_panes_any_live_pane_keeps_session_live`.

The `-a` enumeration is genuinely gone: `git diff origin/main...HEAD -- src/daemon/ src/lib.rs src/session_manager/` is **0 lines**, and there is no `-a` flag anywhere in `guided_resume.rs`.

**The TAB claim survives attack, though not quite for the stated reason.** The doc says TAB "cannot appear in a tmux session name or a command word." I did not confirm tmux *forbids* it — but it does not matter, because the parse is fail-open in that direction anyway: a stray TAB produces a 4th field, `fields.next().is_some()` returns `None` at `:284-286`, and the caller fails open. The #1813 smearing shape is pinned by `parse_pane_probes_rejects_smeared_row`. Every malformed-row path I could construct routes to `None` → fail open.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/guided_resume.rs` | 384-392 | The membership fail-open guard — the one line standing between an unattributable listing and a `kill_session` — is pinned by **no test**. Deleting it leaves all 1282 bin tests green. | Extract a pure `panes_prove_session_dead(name, panes, probe)` seam containing the guard and unit-test it with foreign-only rows. |

### MEDIUM — the load-bearing guard is unpinned, and the hazard is demonstrably real

```rust
// guided_resume.rs:390
if !panes.iter().any(|p| p.session_name == name) {
    return true;
}
```

**Mutation A — guard deleted, nothing else changed:**

```
test result: ok. 1282 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
```

Fully green. `session_panes_foreign_rows_never_prove_our_session_dead` does not catch it because it asserts a property of its *input* (`!panes.iter().any(...)`) and never calls the function containing the guard. That is precisely the "test named for a guard that never exercises it" pattern this repo keeps getting burned by — the third instance flagged today.

This is not theoretical. I confirmed on tmux 3.6b that target resolution **prefix-matches**, on both calls this code path uses:

```
has-session 'rev4304-alpha'    -> TRUE      (only rev4304-alpha-01 exists)
list-panes -s -t 'rev4304'     -> [rev4304-alpha-01]
```

So a record whose name is a strict prefix of a *different* live session yields a fully-parsed listing that is real, well-formed, and attributable to someone else. Without the guard, `session_has_live_pane` filters that listing by the requested name, matches nothing, and `.any()` over an empty iterator returns `false` — read as "no live pane" → `runtime_live = false` → `/runtime-stop` → `kill_session`, which prefix-matches onto that same live session and destroys it. Unserialed record names exist in practice (`tm-agents`, `tm-proj` are both live on this machine right now), so the prefix relationship is reachable, not contrived.

The guard is correct and its comment explains exactly this. My finding is only that nothing stops a future refactor from dropping or reordering it while CI stays green — and the failure mode is silent destruction of an unrelated live session.

**Fix:**

```rust
/// True only when `panes` PROVE session `name` has no live runtime.
pub(crate) fn panes_prove_session_dead(
    name: &str,
    panes: &[PaneInfo],
    probe: &dyn ChildLivenessProbe,
) -> bool {
    // Unattributable listing: we learned nothing, so nothing is proven.
    if !panes.iter().any(|p| p.session_name == name) {
        return false;
    }
    !session_has_live_pane(name, panes, probe)
}
```

`session_runtime_live` becomes `!panes_prove_session_dead(name, &panes, &ProcessTreeProbe)`, and the guard becomes directly testable — a foreign-rows case then fails when the guard is removed.

---

## Items I was asked to attack

**The `!= "provisioning"` assertion is a legitimate discriminator, not a weakened one.** I verified by reverting the fix (`else if false && !runtime_live`):

```
session_resume_headless_dead_runtime_reconciles_and_restarts ... FAILED
assertion `left != right` failed: ... Reading 'provisioning' means the CLI attached into the idle shell instead
  left: "provisioning"
 right: "provisioning"
```

It is environment-independent for a structural reason: `Attach` performs **zero** daemon round-trips, so the seeded `provisioning` survives byte-identical; the reconcile path performs two and cannot land back on `provisioning` from either. `active` vs `errored` is exactly the environment-dependent part, and it is correctly excluded. Requiring `active` would have asserted the runtime *spawn*, which is not this test's claim. The narrowing is principled and the test still fails on a broken fix — which is the only thing that matters.

**The pre-existing fixture defect is real and the repair is correct, not accommodating.** I reproduced the old fixture directly:

```
# bare `tmux new-session -d`, the old fixture:
rev4304-oldfix|zsh|11957      children of 11957: []
```

`zsh` is in `IDLE_SHELL_COMMANDS` with no live child — since #3873 that *is* the dead-runtime shape. So `session_resume_headless_active_live_tmux_skips_restart_and_attach` had genuinely stopped expressing the already-live case it asserts. Repairing it with a non-shell pane command restores the case rather than papering over a failure, and the sibling test now covers the shape the old fixture had drifted into. Correct diagnosis, correct repair.

**Test isolation and cleanup are airtight — verified, not assumed.** This mattered most, with 11 live sessions on this box including attached ones.

- Names are process-unique via `std::process::id()` threaded through the repo URL into the derived `tmux_name` (`create.rs:92-102`, `github_project` → `resolve_session_name`). Confirmed reachable.
- Their two-bin-targets claim is independently confirmed: `Cargo.toml:124-133` defines `tm` and `trusty-mpm` from the **same** `path`, and my earlier full-suite run showed the identical bin test-count line twice. The concurrent-cleanup hazard they describe was real.
- Fixture names (`tm-r<pid>-01`, `tm-deadrt<pid>-01`) cannot collide with, or prefix-match, any of the 11 live sessions. The `-01` suffix also prevents one run's name prefixing another's across differing pid lengths.
- Cleanup is placed **before** all assertions in both tests, so an assertion failure still tears the session down.
- Empirically: 8 total runs (full suite + 6 targeted), all green, `tmux ls` steady at **11** before and after, **0** leaked fixture sessions. Nothing reached a non-fixture session.

**Flakiness:** 6/6 consecutive clean runs of the two real-tmux tests, plus a clean full suite. The three-agreeing-reads + panic-at-deadline pattern is the right shape — panicking on an unsettled fixture instead of proceeding is what converts a silent wrong-branch pass into a loud failure.

**Full suite, no `--lib`:** `EXIT=0`, **6741 passed, 0 failed, 16 ignored**. (My total is 6741 vs the reported 6717 — expected drift, my run is post-rebase against `c11da9cd`. Zero failures either way.)

**Closing keywords:** `closingIssuesReferences` queried twice, minutes apart — `3873` both times, nothing else. Exactly one `Closes #3873` across all three commit messages, none in a code span. Reaper defect untouched and out of scope, as stated.

---

## Notes

- **Known limitation, correctly chosen:** the widened question means a split session with any live non-shell pane now reads live, so an operator whose agent died in one pane while `vim` runs in another still gets `Attach` into the idle shell — #3873 unfixed for that configuration. That is the right trade against a `kill_session`, and it is what the fail-open design buys. Worth stating plainly in the issue so it is not rediscovered as a regression.
- `managed_tests.rs:570-573` passes `"sh"` and reasons about `sh` reading no rc files, but tmux launches the pane command through the default-shell — on this machine `pane_current_command` reads `bash`, not `sh`. Outcome is identical (both allowlisted, observed childless) and `wait_for_stable_dead_runtime` covers the rest, so the fixture is sound; only the comment's reasoning names a process that is not what runs.
- `wait_for_stable_live_runtime` / `wait_for_stable_dead_runtime` panic *after* creating the real tmux session and *before* cleanup, so a deadline miss leaks one scratch session. Bounded and self-healing — the name is unique and `tm-`-prefixed, so the real orphan-GC reclaims it — and I observed zero leaks across 8 runs. Noted for completeness, not flagged.
- Round-1 note that no longer applies: the non-zero-exit guard being near-dead against tmux's rc=0 behavior is now explicitly documented at `:350-353` as belt-and-braces, with the parse path named as the actual protection. Accurate.

Issue #3236, opened 2026-07-19
fix(trusty-agents): wire PythonToolPlugin RBAC to use real restricted_tiers() hook


bobmatnyc:
PythonToolPlugin's own RBAC guard is disconnected from the enforced path.

Current state: plugins/python_tool/mod.rs:323-336 impl ToolExecutor doesn't override restricted_tiers(), so the trait default (empty slice = unrestricted) applies regardless of the is_restricted_for_tier(&str) field. This means even if #3235 wires the plugin into the registry, the RBAC tier check at dispatch time (tools/mod.rs:174-189, ToolRegistry::dispatch_for_user calling user.can_access_tier(tool.restricted_tiers())) will always pass because tool.restricted_tiers() returns an empty slice.

Fix: Map PythonPluginConfig's restricted_tiers: Vec<String> to Vec<ServiceTier> and override the trait method, matching the pattern used by other tool executors.

File evidence: crates/trusty-agents/src/plugins/python_tool/mod.rs:323-336, :50-69, crates/trusty-agents/src/tools/mod.rs:174-189, crates/trusty-agents-common/src/lib.rs:338-342

Effort: S
Depends on: #3235

Part of #3052

Issue #3235, opened 2026-07-19
epic(trusty-agents): reviewed Python-backed skills


bobmatnyc:
## Goal

Support Python-backed skills without turning an agent package into an unreviewed arbitrary-code execution path.

## Consolidated scope

- Extend skill metadata with an explicit Python runtime and entrypoint.
- Provision an isolated `uv` environment deterministically.
- Invoke Python skills through an MCP-server boundary.
- Require per-skill execution allowlists in agent manifests.
- Wire the Python tool plugin only after validation and policy checks.
- Align trusty-code and trusty-mpm skill-catalog metadata where cross-harness reuse is required.

## Acceptance criteria

- A Python skill cannot execute merely because files are present.
- Installation, review, allowlisting, and invocation are separate observable stages.
- Environments are reproducible and isolated.
- MCP invocation has timeout, output, credential, and audit controls.
- Non-Python skills remain unaffected.

## Archived decomposition

The former leaf tickets #3237 and #3239–#3242 are consolidated here.

bobmatnyc:
## Consolidation update (2026-07-28)

This is now the canonical Python-backed skills issue. It absorbs #3237, #3239, #3240, #3241, #3242; the body preserves their actionable scope while the individual leaf issues remain available as design history.

bobmatnyc:
Closed not planned in the 2026-09-02 milestone-68 epic review (owner-confirmed, session tm-trusty-tools-02): every referenced child closed not-planned, nothing shipped under this epic since the 2026-08-14 backlog reset. Reopen if this scope is revived.

Issue #3052, opened 2026-07-18
epic(trusty-agents): daily-driver personal Assistant


bobmatnyc:
## Product outcome

A persistent, personalized Assistant performs approved productivity work across connected knowledge and communication systems while remaining interruptible, resumable, proactive, and auditable.

## Scope

- base persona, customization, persistent desktop, and thin remote clients;
- approved Google Workspace and channel bindings;
- scheduling, triggers, and proactive notifications;
- daemon-backed memory/search;
- interrupt, checkpoint, restart, and resume;
- tm/tcode MCP capabilities;
- permission, credential, RBAC, and tool-catalog enforcement;
- incremental chat streaming with bounded non-SSE fallback, protocol/error parity, and cancellation;
- deterministic drift and integration tests.

## Done when

The north-star calendar-to-draft flow works after restart, streaming is responsive and bounded, unsafe grants fail closed, and configured tools/credentials are observable.

Former tickets #3082/#3102/#3129/#3147/#3206/#3207/#3209/#3264/#3288 and streaming follow-up #3982 are consolidated here.

bobmatnyc:
Auditor workstream filed (2026-07-18):

M1-milestoned (#16):
- #3112 [AUD-1] L1 audit-stream: DispatchContext + dispatch wrapper + JSONL emission
- #3113 [AUD-2] L2 rule engine: pre_check hook + initial 5 rules
- #3114 [AUD-3] index.jsonl summary + retention/rotation sweep + redaction posture
- #3115 [AUD-4] CU-6 subsumption: wire computer-use actions through the AUD-1 seam
- #3116 [AUD-5] Read-only audit MCP tools (audit_read_run/audit_read_recent/audit_daily_summary)
- #3117 [AUD-8] L4 outcome validators: Gmail/Calendar/Drive re-query recipes
- #3118 [AUD-13] DOC-41 §8 spec section: SPEC-AGENTFW-07 Audit & Conformance Layer (docs-only)

M2 (no milestone):
- #3119 [AUD-6] L3 auditor base agent: manifest + audit-conformance-rubric skill + per-run async judge
- #3120 [AUD-7] Daily-sweep scheduled job: cross-run pattern detection
- #3121 [AUD-9] L4 validators: Slack/Telegram
- #3122 [AUD-10] Violation-only push notification wiring
- #3123 [AUD-11] Mistakes-log + memory-palace ingestion bridge for L3 findings
[...]

PR #4313, opened 2026-07-29
feat(trusty-agents): register an agent's declared [[plugins.python]] tools

bobmatnyc:
## The defect

`[[plugins.python]]` declared in an agent or skill package was **parsed and
then silently dropped**. `PythonToolPlugin::from_config` had zero production
call sites anywhere in the workspace — the primitive was fully built and
unit-tested, `AgentConfig` parsed `[plugins]` into `AgentPluginsConfig`, and
nothing ever turned that config into a registered tool. No error, no warning:
the declaration just did nothing.

```
$ rg 'PythonToolPlugin::from_config' crates/   # before this PR
crates/trusty-agents/src/plugins/python_tool/tests.rs:132
crates/trusty-agents/src/plugins/python_tool/tests.rs:159
crates/trusty-agents/src/plugins/python_tool/tests.rs:336
crates/trusty-agents/src/plugins/python_tool/tests.rs:357
```

Four hits, all tests.

## The fix

`run_pm_task_with_persona` now reads `persona_cfg.plugins.python` and registers
each entry through a new `persona_plugins::register_python_plugins`. A
package-relative `script` / `schema_file` resolves against the agent/skill
package dir — the same `config_dir` the delegation runner is already handed —
so `script = "../skills/foo/foo.py"` reaches a co-located skill package.

The helper lives in its own module for the same reason `persona_gate` does:
`persona.rs` sits just under the 500-SLOC production cap. Like
`build_persona_delegate_tool`, it mutates the **exact** registry the dispatch
path arms, so the tests drive the production construction rather than a
re-derived copy of it.

Registration is not a grant. The `[tools].allow` glob filter, the RBAC tier
filter and the scope gate in `filter_persona_tool_names_for_tier` all still run
afterwards and decide whether the tool reaches the model.

Two non-fatal skip paths, so one bad line in one agent TOML costs one tool
rather than the whole chat turn:

- an entry `from_config` rejects (e.g. an unknown `restricted_tiers` string,
  fail-closed since #3236) is logged and skipped;
- an entry whose name is **already registered** is refused.
  `ToolRegistry::register` overwrites silently in release and `debug_assert!`s
  in debug, so an unguarded registration here would have made a debug-build
  panic reachable straight from user-authored config — and a production
  hijack of `create_dir` / `run_bash` reachable the same way. This guard is
  new relative to the abandoned draft this work started from.

## Tests

Four unit tests on the helper (`persona_tests.rs`):

| test | pins |
|---|---|
| `persona_python_plugin_registers_callable_tool` | registered **and dispatchable** through the real `ToolRegistry`, with a package-relative `script` |
| `persona_python_plugin_bad_config_is_skipped` | a rejected entry does not abort the turn |
| `persona_python_plugin_does_not_shadow_an_existing_tool` | the native `create_dir` survives an impostor declaration |
| `persona_python_plugin_empty_list_registers_nothing` | no-op for every existing agent |

Plus `tests/persona_python_plugin_wiring.rs`, which drives the **real**
`run_pm_task_with_persona` against a sandboxed throwaway agent package
(`TAGENT_CONFIG_DIR` + `$HOME` redirected to a tempdir) and asserts on the
tracing output that the plugin was registered *and* survived the allow / RBAC /
scope gate into the advertised set. The LLM leg is routed at a local absent
Ollama endpoint via `SessionOverrides::provider = "local"`, so it needs no
credential and makes no external call — the assertions are on the registry the
function built, not on a model reply.

### Proof the test is not vacuous

The unit tests would pass against a tree where the helper exists but nothing
calls it — which is exactly the shape of the original defect. The integration
test is the one that closes that. Removing **only** the call site from
`persona.rs` (helper and unit tests untouched):

```
test persona_python_plugin_is_registered_by_the_dispatch_path ... FAILED

run_pm_task_with_persona must register the agent's declared
[[plugins.python]] tools; captured logs:
...
INFO trusty_agents::ctrl::pm_task::dispatch::persona: persona tool registry built
     persona=pyplug-probe tools=[] rbac_user=cli rbac_tier=All
```

`tools=[]` — the declared plugin reached the model as nothing at all, which is
precisely the production bug. Call site restored → green in 7.2s.

## Gate

```
cargo test -p trusty-agents        2960 passed; 0 failed; 11 ignored  (lib)
                                   + 28 passed across 8 integration binaries
cargo clippy -p trusty-agents --all-targets -- -D warnings   clean
cargo fmt --all --check                                      clean
scripts/check_line_cap.sh          7 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh     0 dangling pointers — OK
scripts/check_sld.sh               0 errors, 0 warnings
```

Nothing was `#[ignore]`d, cfg-gated, or excluded to reach green.

## Scope notes

**SECURITY** (unchanged by this PR, restated because the path is now live): the
Python subprocess has **no OS-level sandbox**. RBAC gates *who* may call the
tool, not *what* the script does once spawned. Reaching it still requires an
explicit `[[plugins.python]]` declaration **and** an explicit `[tools].allow`
entry — file presence alone executes nothing — but the review / allowlisting /
`uv`-isolation model in epic #3235 is **not** implemented here and that epic
stays open.

**KNOWN GAP**: this wires the in-process persona-chat dispatch path only. The
subprocess `--agent` path (`runtime::tool_registry::build_assistant_tier_registry`,
which takes no `AgentConfig`) still ignores `[[plugins.python]]` — the #3745
item-C "a capability model that holds on only one dispatch path" hazard, called
out in the CHANGELOG and left for a follow-up rather than folded in here.

Relates to epic #3052 and epic #3235. Deliberately **no** closing keyword: the
`#446` referenced throughout this code is a pre-migration issue number from a
different repo's sequence (this repo's #446 is an unrelated merged
trusty-memory PR), and neither open epic is fully closed by this change.

## Fixtures deliberately not included

The draft this work started from also carried a `demo-assistant.toml` agent and
a two-file `crypto-price` skill (CoinGecko + offline fallback), wired to an
`#[ignore]`d live-LLM e2e test. Skipped: they are demo scaffolding, not
coverage — the `#[ignore]` meant they gated nothing, and shipping a
`.trusty-agents/` package inside the crate would add an unreferenced agent with
a network dependency to the published artifact. Both tests here build their own
minimal NDJSON-contract fixtures at runtime instead.

### DIFF
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -49,6 +49,35 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **`[[plugins.python]]` tools declared by an agent are finally registered and
+  callable (#446, epic #3052).** The `PythonToolPlugin` subprocess primitive
+  (spawn `python3 <script>`, one NDJSON `tool_call` in / one `tool_result`
+  out, per-call timeout, fail-closed RBAC tier guard) has shipped fully
+  unit-tested since #446, and `AgentConfig` has parsed `[plugins]` for just as
+  long — but nothing in the workspace ever called
+  `PythonToolPlugin::from_config`, so an agent or skill package declaring a
+  `[[plugins.python]]` tool had its declaration parsed and then silently
+  dropped: never registered, never advertised, never callable, no warning.
+  `run_pm_task_with_persona` now reads `persona_cfg.plugins.python` and
+  registers each entry via a new `persona_plugins::register_python_plugins`
+  helper, resolving a package-relative `script` / `schema_file` against the
+  agent/skill package dir (the same `config_dir` the delegation runner is
+  handed), so `script = "../skills/foo/foo.py"` reaches a co-located skill
+  package. Registration is not a grant: the `[tools].allow` glob filter, the
+  RBAC tier filter and the scope gate all still decide whether the tool
+  reaches the model. Two non-fatal skip paths keep one bad line in one agent
+  TOML from taking down a chat turn — an entry `from_config` rejects (e.g. an
+  unknown `restricted_tiers` string, fail-closed since #3236) is logged and
+  skipped, and an entry whose name is already taken by a native or MCP tool is
+  refused rather than registered (`ToolRegistry::register` overwrites silently
+  in release and `debug_assert!`s in debug, so an unguarded registration would
+  have made a debug-build panic reachable straight from user-authored config).
+  SECURITY NOTE, unchanged by this PR and worth restating now that the path is
+  live: the Python subprocess has NO OS-level sandbox — RBAC gates WHO may call
+  the tool, not WHAT the script does once spawned. KNOWN GAP: this wires the
+  in-process persona-chat dispatch path only; the subprocess `--agent` path
+  (`runtime::tool_registry::build_assistant_tier_registry`, which takes no
+  `AgentConfig`) still ignores `[[plugins.python]]` and needs its own change.
 - **The app header shows the running daemon's version.** A quiet provenance
   line sits beside the brand lockup, read from `GET /api/health` — the version
   of the daemon the UI is actually talking to, not what the frontend was
--- a/crates/trusty-agents/src/ctrl/pm_task/dispatch/mod.rs
+++ b/crates/trusty-agents/src/ctrl/pm_task/dispatch/mod.rs
@@ -15,6 +15,9 @@ mod persona;
 // sits exactly at the 500-SLOC production cap.
 mod persona_gate;
 mod persona_memory;
+// #446 (epic #3052): `[[plugins.python]]` registration, split out of
+// `persona.rs` for the same SLOC-cap reason as `persona_gate`.
+mod persona_plugins;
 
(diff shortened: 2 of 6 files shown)


bobmatnyc:
### CI note (second commit, 84b95a5d)

The first CI run went red on **only** the new integration test, while the same
test was green locally — the classic ambient-credential shape.
`llm::create_client` runs *before* the persona registry is built and hard-fails
when no credential resolves at all, so on a runner with no
`OPENROUTER_API_KEY` the dispatch returned `Err` one line after
`credentials resolved` and the registry the test asserts on never existed. The
`provider = "local"` override sidesteps the *routing* decision but not that
check.

Fixed by setting a placeholder `OPENROUTER_API_KEY` inside the test's own
sandboxed env, unconditionally so both environments take the identical path. It
is never transmitted — `provider = "local"` still points the chat leg at a local
Ollama endpoint, which sends no auth header.

Re-verified under `env -u OPENROUTER_API_KEY -u ANTHROPIC_API_KEY -u
CLAUDE_CODE_OAUTH_TOKEN`: green with the wiring, still red (`tools=[]`) with the
call site removed — the regression value is intact, and the test no longer
depends on what happens to be exported.

All 23 checks now pass (the 24th, `Notify on red main`, is main-only and skips).

Issue #4322, opened 2026-07-29
perf(trusty-mpm): tm ls re-reads ~47 MB of deployed assets serially per invocation — 56% is for stopped sessions


bobmatnyc:
DEFECT: `stale_assets_for_many` (`crates/trusty-mpm/src/daemon/managed_routes/summary.rs:342`) runs one `spawn_blocking` (`:449`) containing a serial `for record in records` loop. Each iteration reads ~93 deployed agent/skill files. No cache survives the request — the `CatalogHashes` map at `:343` is a local, dropped on return.

EVIDENCE (measured):
- Cold `tm ls`: 4.93s / 5.91s / 7.24s / 9.37s across four independent first-calls. Warm: 0.15-0.27s.
- Daemon CPU is a constant 0.128s regardless of wall time — blocking I/O, not computation.
- Causal proof: after a 300s idle, reading exactly the 3,269 files the probe touches from an unrelated process took 3.156s; the next request dropped to 0.425s. Warming that file set alone is the latency.
- Linear in session count: `/api/v1/sessions/managed` probing 32 sessions = 0.175s wall / 0.128s CPU; scoped to 5 sessions = 0.035s / 0.016s; `/health` (0 sessions) = 0.017s / 0.010s. ~180ms per session cold.
- 20 of 32 probed sessions have no live tmux session — 1,834 file reads, 26.3 MB, 56% of the work. Several last active 2026-07-17 to 07-24; three never active.
- The entire output is a cosmetic `[stale-assets]` marker in the STATE column.
- Cleared: no git subprocesses, no network, no per-session tmux fan-out, no per-session JSON reads.

RESOLUTION, in payoff order:
1. Stop probing `Stopped` sessions (`summary.rs:310-315`) — removes 56%, decouples latency from dead-session count. ~0.25d.
2. Parallelize `stale_assets_for_many` using the `JoinSet` pattern already applied to `unresumable` at `summary.rs:418-435`. 5-10x off cold. ~0.5d.
3. Memoize per session keyed on `(agents_dir, skills_dir)` mtime, held in `DaemonState`. ~0.5-1d.
4. Precompute in the existing 60s background loop (`daemon/mod.rs:586`) so the route serves from memory — O(1). ~1-1.5d.

A PR implementing steps 1 and 2 only is in flight — this issue stays OPEN after it merges; steps 3-4 remain.

[...]

Issue #2444, opened 2026-07-11
bug(trusty-mpm): live managed sessions never pick up agent/skill catalog updates — #2002 sync is one-shot at launch


bobmatnyc:
## Observed (dogfood verification, 2026-07-11, session tm-dogfood-relaunch-01)
- Session launched 2026-07-07; its deployed .claude/skills/tm-issues-prune is v1.0.0 while the tm-owned catalog (~/.trusty-tools/trusty-mpm/claude-config/skills/) has v1.1.0 (bumped 2026-07-11 via PR #2393). The running session never re-synced.
- Same class of gap on agents: the session's .claude/agents/ has 40 files vs 41 in the catalog (generic engineer.md missing) — catalog additions after launch don't propagate either.
- Daemon /health reports catalog_stale:false — the health surface doesn't know per-session deployments can be stale relative to the catalog.

## Impact
Every skill/agent improvement only reaches NEW sessions; long-lived dogfood sessions (days old) silently run stale behavior. PM skill fixes (e.g. the tm-issues-prune organize/prioritize enhancement) are invisible until relaunch.

## Proposal sketch
(a) a re-sync mechanism for live sessions — either daemon-pushed on catalog change, or a `tm sessions sync-assets [--all]` verb the operator/PM can run; (b) per-session staleness surfaced in `tm sessions ls`/status (deployed-manifest checksum vs catalog checksum) and in the /health catalog fields; (c) decide policy for mid-session skill swaps (safe for skills = plain markdown reload; agents may need care).

## Refs
#2002 (managed-session CLAUDE_CONFIG_DIR + roster deploy), PR #2393 (the skill bump that exposed this), DOC-31 (system vs project agents & skills catalog-sync conventions).

Issue #2595, opened 2026-07-14
fix(trusty-mpm): sessions with missing workspaces must not be offered as resumable (picker/ls/guided)


bobmatnyc:
Owner directive (2026-07-14, verbatim intent): "it shouldn't have been a session option in that case." Context: the bare-tm guided flow and session picker offered to restart session tm-tcode, whose workspace had been GC-pruned (workspace_path null, cwd /unknown sentinel, last_cwd deleted worktree) — the restart could only ever fail. #2577/PR #2594 fixed the ERROR (bare 500 → actionable 422 + tm session rm hint), but the deeper UX defect remains: an unresumable session should never be presented as a restart option at all.

Scope:
1. Define an "unresumable" predicate shared with #2594's ResumeManagedError::Unresumable semantics: no existing workspace candidate (workspace_path, last_cwd, cwd all missing/sentinel) for a stopped session.
2. In the bare-tm guided default and the tm ls session picker: mark such sessions clearly (e.g. "dead — workspace removed; run tm session rm") and exclude them from the restart/resume action (selection can offer decommission instead).
3. tm sessions ls: show an unresumable/dead indicator in the state column or a flag.
4. Optional (evaluate): boot-reconcile could tag these records once (persisted flag) instead of re-probing paths on every render — mind performance on large stores.
5. Hermetic tests: dead-record shape excluded from resume options; healthy stopped sessions unaffected.

Refs #2577, PR #2594, #2250 (resume-cwd family), the orphan-GC pruning behavior.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #4326, opened 2026-07-29
perf(trusty-mpm): stop probing stopped sessions and parallelize the tm ls asset-staleness fan-out

bobmatnyc:
## Problem

`tm ls` cold: **4.9–9.4 s**; warm: 0.15 s. Daemon CPU is a constant 0.128 s
regardless of wall time — this is blocking filesystem I/O, not computation.

`daemon::managed_routes::summary::stale_assets_for_many` ran **one**
`spawn_blocking` containing a serial `for record in records` loop. Each
iteration re-read that session's whole deployed `.claude/{agents,skills}`
tree. On the measured fleet that is **3,040 `read_to_string` calls and
35.7 MiB per invocation**, uncached — the `CatalogHashes` map was a local,
dropped on return. Cost is linear in session count, and the output is a
cosmetic `[stale-assets]` marker in the STATE column.

**20 of the 32 probed sessions were `stopped` and idle for days — 56% of the
work.**

Refs #4322. This PR does **not** implement the memoization cache or the
background precompute that issue also lists, so it does not close it.

## Fix 1 — stop probing stopped sessions on the LIST path

`probe_staleness_for` is split into two predicates:

| predicate | states | used by |
|---|---|---|
| `staleness_meaningful_for` | `Active`, `Stopped`, `Errored` | `record_to_summary_checked` — the single-session `GET …/managed/{id}` that `tm session resume` reads |
| `probe_staleness_in_list` | `Active`, `Errored` | `checked_summaries` — the fleet list behind `tm ls` |

Staleness is only actionable at resume, and resume fetches the session
individually — so the signal is computed on demand, exactly where it is
acted on, and dropped from the interactive latency path.

### What a stopped session displays now

A stopped row's `stale_assets: false` would otherwise be indistinguishable
from a genuine "checked, and fresh" verdict. So the wire gains an additive
`stale_assets_unchecked` flag and the CLI renders a **distinct third state**:

```
stopped [assets ?]              # not probed — verdict UNDETERMINED
stopped [dead] [assets ?]       # composes with the #2595 dead marker
active  [stale-assets]          # probed, and genuinely stale
stopped                         # (older daemon: probed everything — unchanged)
```

`[stale-assets]` and `[assets ?]` are mutually exclusive; a real stale
verdict always wins. The flag's `false` default is the compat-safe polarity:
an older daemon that omits the field probed every state, so its verdicts stay
authoritative rather than being repainted as unknown.

### Callers checked

`stale_assets` has exactly one non-test consumer: `format_state_column` in
`bin/tm/commands/managed.rs`. `tm doctor`, the resume path, the picker and
the TUI never read it (verified by grep across `crates/`). The single-session
GET keeps its full probe regardless, so nothing that could have depended on
it loses the value.

## Fix 2 — parallelize `stale_assets_for_many`

Same pattern already applied to the `unresumable` probe beside it, with
`spawn_blocking` instead of `spawn` because the work is synchronous
filesystem I/O:

1. `staleness_inputs` (one blocking task) resolves every session's plan and
   computes **one `CatalogHashes` per distinct `(agent_source, skill_source)`
   pair**, handed out as `Arc`. This half stays sequential *on purpose* —
   fanning it out would race N tasks into an empty cache and reinstate the
   per-session recompose #2444 removed.
2. The cheap deployed-side comparison then fans out one blocking task per
   session via `JoinSet::spawn_blocking`.

A panicking task is simply absent from the returned map, leaving that summary
at its `false` default rather than failing the whole listing.

## Measurements

See the follow-up comment on this PR for the cold-timing runs (idle-then-time
method, 3 samples per side, both sides served by a release build of this
worktree so the comparison is apples-to-apples).

Static per-invocation I/O, computed against the live 57-session store
(25 decommissioned / 20 stopped / 12 active):

| | sessions probed | read attempts | bytes |
|---|---|---|---|
| before | 32 | 3,040 | 35.7 MiB |
| after | 12 | 1,140 | 15.0 MiB |

## Tests

- `checked_summaries_does_not_probe_stopped_sessions` — **required test**: a
  genuinely stale STOPPED session is not flagged by the list path, and its
  row advertises `stale_assets_unchecked`. Verified to FAIL without the fix
  (predicate temporarily reverted to include `Stopped`; it panicked on the
  first assertion, then the fix was restored).
- `record_to_summary_checked_still_flags_stale_stopped_session` —
  **correctness gate**: the same genuinely-stale stopped record, fetched
  individually, is still reported stale. A `tm ls` that is fast because it
  never detects staleness would fail here.
- `checked_summaries_flags_stale_assets_only_for_relevant_states` and
  `checked_summaries_stale_assets_independent_per_session_sharing_one_catalog`
  (pre-existing) now exercise the parallel path and pin that a stale ACTIVE
  session is still detected and that catalog sharing never cross-contaminates
  per-session verdicts.
- `format_state_column_appends_unchecked_assets_marker`,
  `managed_session_summary_defaults_stale_assets_unchecked_false`.

Nothing `#[ignore]`d or cfg-gated.

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,7 +7,35 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Performance
+
+- `tm ls` no longer re-reads the whole fleet's deployed agent/skill trees on
+  every invocation ([#4322](https://github.com/bobmatnyc/trusty-tools/issues/4322)).
+  Two changes, both in `daemon::managed_routes::summary`:
+  - **`Stopped` sessions are no longer asset-staleness-probed on the LIST
+    path.** The probe costs ~95 `read_to_string` calls against a session's own
+    `.claude/{agents,skills}` tree, uncached, per invocation. On a measured
+    32-session fleet, 20 of those sessions were stopped and idle for days —
+    56% of the I/O — for a marker that is only actionable at resume. The
+    single-session `GET …/managed/{id}` path (what `tm session resume` reads)
+    still probes every meaningful state, so the signal is unchanged where it
+    is acted on. Measured on that fleet: 3,040 → 1,140 read attempts and
+    35.7 MiB → 15.0 MiB per listing.
+  - **The remaining per-session comparisons now run concurrently** on the
+    blocking pool via `JoinSet::spawn_blocking`, mirroring the `unresumable`
+    fan-out already used alongside them. The shared `CatalogHashes` compute
+    (#2444) stays a single sequential step — fanning it out would reinstate
+    the per-session recompose it exists to prevent.
+
 ### Changed
+
+- `tm ls` STATE column: a row the daemon did not probe now renders
+  `[assets ?]` rather than nothing, so the absence of `[stale-assets]` can
+  never be misread as "assets fresh"
+  ([#4322](https://github.com/bobmatnyc/trusty-tools/issues/4322)). Backed by a
+  new additive wire field, `SessionSummary::stale_assets_unchecked`, whose
+  `false` default keeps an older daemon's (authoritative) verdicts rendering
+  exactly as before.
 
 - `daemon::managed_routes::prune`: the orphan sweep's `active_workspace_paths`
   set is now documented as DELIBERATELY unfiltered by session state, and pinned
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
@@ -527,6 +527,7 @@ fn stopped_record_at(workspace: &std::path::Path) -> trusty_mpm::client::Managed
         injection_status: None,
         unresumable: false,
         stale_assets: false,
+        stale_assets_unchecked: false,
         attached: false,
         slot: 0,
         deleted: false,
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_resolver_tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_resolver_tests.rs
@@ -47,6 +47,7 @@ fn make_session(
         injection_status: None,
         unresumable: false,
         stale_assets: false,
+        stale_assets_unchecked: false,
         attached: false,
         slot: 0,
         deleted: false,
--- a/crates/trusty-mpm/src/bin/tm/commands/managed.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/managed.rs
@@ -200,7 +200,9 @@ pub(crate) async fn session_ls(
 /// actionable remedy is `tm session rm`. It also appends `[stale-assets]`
 /// (#2444) when the server-computed `stale_assets` flag is set — the
 /// session's deployed agents/skills have drifted from the catalog; the remedy
-/// is `tm sessions sync-assets <id>`.
+/// is `tm sessions sync-assets <id>` — or `[assets ?]` (#4322) when the daemon
+/// skipped the probe for that row (`stopped` sessions), so the operator reads
+/// "undetermined" rather than mistaking a missing marker for "assets fresh".
 /// Test: `ls_source_id_filter_selects_correct_slug` covers the scoping seam; the
 /// table bytes are exercised by `tests/session_manager_mvp.rs`;
 /// `format_state_column_appends_dead_marker` and
@@ -257,7 +259,8 @@ pub(crate) fn render_session_table(sessions: &[ManagedSessionSummary], source_id
         } else {
             s.state.as_str()
         };
-        let state = format_state_column(base_state, s.unresumable, s.stale_assets);
+        let unchecked = s.stale_assets_unchecked;
+        let state = format_state_column(base_state, s.unresumable, s.stale_assets, unchecked);
         println!(
             // #1841 Fix 5: truncate name with ellipsis when it exceeds column width.
             "{:<5}  {:<36}  {:<14}  {:<24}  {:<30}  {}{}",
@@ -290,15 +293,23 @@ fn format_tombstone_row(slot: u32) -> String {
 /// What: renders the base state (`deleted` → the `--deleted--` marker, #2012;
 /// every other state verbatim), then appends `" [dead]"` when `unresumable` is
 /// `true` (the session is `stopped`/`errored` with no workdir candidate left on
-/// disk, so a resume is guaranteed to fail), and/or `" [stale-assets]"` when
-/// `stale_assets` is `true` (the session's deployed agents/skills have drifted
-/// from the catalog — `tm sessions sync-assets <id>` fixes it). Markers can
-/// appear together.
+/// disk, so a resume is guaranteed to fail), and `" [stale-assets]"` when
+/// `stale` (the summary's `stale_assets`) is `true` — the session's deployed
+/// agents/skills have drifted from the catalog and `tm sessions sync-assets
+/// <id>` fixes it — or `" [assets ?]"` when `unchecked` (the summary's
+/// `stale_assets_unchecked`) is `true` (#4322: the daemon
+/// did not probe this row — the list path skips `stopped` sessions — so
+/// staleness is UNDETERMINED, not known-fresh; `tm sessions sync-assets <id>`
+/// is still the remedy, and resuming the session determines it). The two asset
+/// markers are mutually exclusive — an unprobed row has no verdict to report —
+/// so a stale verdict always wins if both flags somehow arrive set. Markers can
+/// otherwise appear together.
 /// Test: `format_state_column_appends_dead_marker`,
 /// `format_state_column_appends_stale_assets_marker`,
+/// `format_state_column_appends_unchecked_assets_marker`,
 /// `format_state_column_leaves_healthy_state_unchanged`,
 /// `format_state_column_renders_deleted_marker`.
-fn format_state_column(state: &str, unresumable: bool, stale_assets: bool) -> String {
+fn format_state_column(state: &str, unresumable: bool, stale: bool, unchecked: bool) -> String {
     // A `deleted` record (soft-deleted via `tm sessions delete`) renders as the
     // `--deleted--` marker so the master list REFLECTS the deletion instead of
     // the row silently vanishing (#2012 marker).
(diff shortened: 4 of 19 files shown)


bobmatnyc:
## Signal change for downstream consumers — `[stale-assets]` on stopped rows

Flagging explicitly, because a separate piece of work was considering keying
off this marker:

**A `stopped` session will no longer be reported `stale_assets: true` by the
fleet list endpoint** (`GET /api/v1/sessions/managed`, what `tm ls` renders).
Anything that treated "no `[stale-assets]` marker on a stopped row" as
evidence the assets are fresh will now be wrong.

What replaces it:

- The row carries the new `stale_assets_unchecked: true` and renders
  `[assets ?]` — an explicit "undetermined", not silence.
- The **single-session** endpoint `GET /api/v1/sessions/managed/{id}` still
  computes the real verdict for `stopped` sessions and always did — that is
  the path `tm session resume` uses, and it is unchanged. A consumer that
  needs a stopped session's true staleness should fetch it there (or run
  `tm sessions sync-assets <id>`, which is idempotent and never gated on the
  flag).
- `sync-assets` itself is untouched: it still applies to
  `Active`/`Stopped`/`Errored`. This PR narrows only what the LIST surface
  *displays*, never what the fix acts on.

## On the parallelism shape (review point)

The concern raised was that fanning `stale_assets_for_many` out per record
would reinstate the N-times catalog recompose that #2444's review removed.
Agreed — and the implementation does not do that. The split is:

- `staleness_inputs` runs in **one** blocking task, **sequentially**, and
  computes `CatalogHashes::compute` **once per distinct `(agent_source,
  skill_source)` pair**, handing each session an `Arc` to the shared result.
- Only the per-session **deployed-side** half is fanned out
  (`JoinSet::spawn_blocking`), and each task merely *borrows* its `Arc`. No
  spawned task can call `CatalogHashes::compute`.

I deliberately did **not** additionally parallelize *across* groups. On this
fleet every session resolves the same default source, so there is exactly one
group and one compute — cross-group concurrency would optimise a single
sequential step that is ~5% of the work, while adding a second concurrency
seam over the very cache whose whole purpose is deduplication. The 35.7 MiB
is entirely in the deployed-side half, which is what the fan-out covers.

Commit `5b28892` adds the mechanical guard so this cannot silently regress:
`staleness_inputs_computes_one_catalog_per_source_pair_shared_by_arc` asserts
pointer-equal `Arc`s for sessions sharing a source pair (verified to fail
when the cache is bypassed), plus an `INVARIANT:` comment at the
`spawn_blocking` site.

bobmatnyc:
## Verdict: WARN

Adversarial review against the perf contract in #4322 and the code-review-standards rubric. Diff, checks (all 23 green, `Test` 16m36s), and the merge commit (`e9922ebc`) were all inspected; one HIGH finding, reproduced empirically, everything else checks out.

## Findings

| HIGH | `crates/trusty-mpm/src/daemon/managed_routes/tests.rs` | 818-843 (`staleness_inputs_computes_one_catalog_per_source_pair_shared_by_arc`) | The pinned anti-regression test calls `staleness_inputs()` directly with a full batch of records — it never goes through `stale_assets_for_many` (`summary.rs:446`), the actual hot-path function the PR's own comment at `summary.rs:452-458` calls out as the place "moving the catalog work in here would give every session its own recompose." I reproduced this: I edited `stale_assets_for_many` so its `JoinSet::spawn_blocking` fan-out calls `staleness_inputs(vec![record])` **per session** instead of sharing the one batch-computed `staleness_inputs(records)` result — i.e., reinstated the exact #2444 per-session catalog recompose inside the real call site. Ran the full `daemon::managed_routes` suite (176 tests): **all 176 passed**, including this pinned test and `checked_summaries_stale_assets_independent_per_session_sharing_one_catalog`. The INVARIANT comment's claim ("Move `CatalogHashes::compute` inside the fan-out and this fails immediately") is false for the realistic regression vector — a future edit to `stale_assets_for_many`'s loop body. CI would stay green while silently reintroducing the O(N)-catalog-recompose bug this PR exists to prevent. | Make the test exercise `stale_assets_for_many` (or `checked_summaries`) itself, not just the `staleness_inputs` helper in isolation. Concretely: add a `#[cfg(test)]` atomic call counter inside `CatalogHashes::compute` (or a test-only wrapper), reset it, call `stale_assets_for_many` with N≥2 sessions sharing one `(agent_source, skill_source)` pair, and assert the counter is exactly 1. That is the only way to structurally pin the *integration*, not just the helper. |

## Notes

Everything else checked out clean:
- **Behavior change disclosure**: skipping `Stopped` sessions on the list path is a genuine behavior change, but it's honestly surfaced — `checked_summaries_does_not_probe_stopped_sessions` (tests.rs:785-806) pins both that the probe is skipped AND that `stale_assets_unchecked` is set so `[assets ?]` renders instead of silence. `record_to_summary_checked_still_flags_stale_stopped_session` (tests.rs:852-873) proves the resume path still detects the same drift. `format_state_column`'s mutual-exclusion logic (`managed.rs:329-134`: `stale` wins over `unchecked`) is correct and tested (`managed_tests.rs:176-191`).
- **Merge commit** (`e9922ebc`): diffed the PR-touched files (`summary.rs`, `session_summary.rs`, `sync_assets.rs`, `managed.rs`, `types.rs`, `tests.rs`, `http_client/tests.rs`) between the pre-merge branch tip (`5b288928`) and the merge commit — byte-identical, zero drift. Diffed main's exclusively-touched files (#4324's `claude_md_sections.rs` etc.) between `f165b927` and the merge commit — also byte-identical except the CHANGELOG.md conflict resolution, which is purely additive (both sections present, nothing dropped). No accidental reversion.
- **`spawn_blocking`/`JoinSet` usage**: `stale_assets_for_many` mirrors the pre-existing `unresumable` fan-out in `checked_summaries` (`summary.rs:543` onward) — same `JoinSet::spawn_blocking`/`join_next` pattern, panics silently dropped (`if let Ok(...) = res`) rather than failing the listing. Fleet sizes observed (32-57 sessions) are far below Tokio's default 512-thread blocking pool cap; no unbounded growth risk, and the pattern isn't new to this PR.
- **`closingIssuesReferences`**: empty array — the PR does not auto-close anything. Body references `#4322` (correctly, as "Refs" not "Closes" — matches the PR's own stated intent to leave the issue open), plus `#2595` and `#2444` as contextual precedent only.
- **CI**: `gh pr checks 4326` — all 23 checks pass, including `Clippy`, `MSRV check (1.91)`, and `Test` (16m36s).

## Required Changes

1. Strengthen `staleness_inputs_computes_one_catalog_per_source_pair_shared_by_arc` (or add a companion test) so it exercises `stale_assets_for_many` end-to-end with an actual compute-count assertion, per the fix above — the current test only pins the `staleness_inputs` helper's internal correctness, not its use at the call site the regression would actually occur in.

bobmatnyc:
## Verdict (updated, re-review at `4929b8b6`): APPROVE

Re-reviewed after the fix commits (`5dc2ab4f`, plus the hand-merge of #4343 at `4929b8b6`). The prior HIGH is resolved and I verified the fix and the merge myself, not just the description. One MEDIUM remains (test-coverage gap, logic itself verified correct).

## Verification of the HIGH fix

Checked out `4929b8b6` in a scratch worktree and re-ran my original mutation against `stale_assets_for_many` (now `pub(super)`, `summary.rs:453`) — moving `CatalogHashes::compute` back inside the per-session `JoinSet::spawn_blocking` fan-out. The new test `stale_assets_for_many_computes_catalog_exactly_once_per_source_pair` (`tests.rs`) failed exactly as claimed:

```
assertion `left == right` failed: three sessions sharing one (agent_source, skill_source) pair
must share exactly ONE CatalogHashes::compute ...
  left: 3
 right: 1
```

Reverted the mutation, full `daemon::managed_routes` suite green (178/178). The gap I found in the first pass — the pin only exercising `staleness_inputs` in isolation, never the real hot path — is closed. The `compute_calls_for`/`COMPUTE_CALL_LOG` approach (keyed by exact source paths, not a bare counter) correctly avoids cross-test pollution from concurrent tests reaching `compute` via their own temp paths.

## Verification of the #4343 hand-merge

Diffed the merge commit against both parents:
- Files exclusively touched by #4343 on main (`guided.rs`, `mod.rs`, `tests_behavior_c_tests.rs`) are byte-identical between `5f8b3385` and `4929b8b6` — no drift.
- The two files with real conflicts (`summary.rs`, `tests.rs`) were hand-folded correctly. Traced the combined predicate in `checked_summaries_with` (`summary.rs`) through every `(state × probe_assets)` combination:

  | state | probe_assets | unchecked | probed |
  |---|---|---|---|
  | Active/Errored | true | false | yes |
  | Stopped | true | true | no |
  | Provisioning/Decommissioned | true | false | no |
  | Active/Errored | false (slim) | **true** | no |
  | Stopped | false (slim) | true | no |
  | Provisioning/Decommissioned | false (slim) | false | no |

  No cell where a session that would have been probed goes unprobed AND unmarked — slim mode (`probe_assets: false`) forces `stale_assets_unchecked` for every meaningful-state record, including `Active`/`Errored`, which the list path alone would otherwise have probed. Confirmed this empirically too: added a temporary assertion to the existing (unmodified-by-this-PR) `checked_summaries_slim_skips_stale_assets_probe` test — `slim[0].stale_assets_unchecked` on a genuinely-stale Active session under slim mode — it passed. Reverted the temporary assertion afterward (not part of the PR).
- `nested_session_guard`'s slim call (`guided.rs:629`, `?slim=true`) → `mod.rs:606-609` (`slim` parsed, `!slim` passed as `probe_assets`) → `numbered_summaries` → `checked_summaries_with`: the whole chain is wired correctly, and it skips the ENTIRE asset-probe pass, not just part of it.
- CI on `4929b8b6`: all 24 checks green, including `Test` (16m4s), `Clippy`, `MSRV check (1.91)`.
- `closingIssuesReferences` unchanged (still `[]` — only references `#4322`, not closing it, matching the PR's stated intent).

## Findings

| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/tests.rs` | `checked_summaries_slim_skips_stale_assets_probe` (~978-1018, pre-existing, not touched by this PR) | This is the one test exercising the exact `#4343`×`#4322` collision path (slim mode over a normally-probed state), but it only asserts `stale_assets` stays `false` and the session is still returned — it never asserts `stale_assets_unchecked`. I confirmed by hand-tracing and empirically (temporary assertion, described above) that the combined predicate IS correct, so this is a coverage gap, not a live bug — but it's the gap most likely to hide a future regression in exactly the interaction under review. | Add `assert!(slim[0].stale_assets_unchecked, ...)` to the existing test (the exact line I added and verified, reverted before this comment) so the slim/unchecked interaction is pinned going forward, not just verified by manual review. |

## Notes

Everything else from the first pass still holds (merge cleanliness, `spawn_blocking`/`JoinSet` pattern, wire-compat defaults) and was re-checked at this head; no new concerns beyond the MEDIUM above.

Issue #1744, opened 2026-06-26
feat(trusty-mpm): graceful ungraceful-exit handling + conversation continuity on resume (--resume <id>)


bobmatnyc:
When a user exits a managed session WITHOUT properly quitting (kills the terminal, closes the window, drops the tmux/ssh client, force-quits — no `/quit`/`exit`), trusty-mpm must (a) reliably detect it and leave the session in a clean resumable Stopped state, and (b) on resume, bring the user back to their prior conversation. Today neither holds. Builds on #1742 (which fixes the raw-attach-of-dead-session crash).

Owner decision: resume should CONTINUE THE PRIOR CONVERSATION via `--resume <claude_session_id>` (with `--continue` as a fallback when the id isn't available).

## Findings (from lifecycle research)
- Resume launches a FRESH claude every time — `runtime/claude_code.rs::spawn_command()` never passes `--continue`/`--resume`. So even a correctly-restarted session comes back blank.
- The Claude Code session id is never captured or stored (absent from `session_manager/record.rs::SessionRecord`).
- `SessionEnd`/`Stop` hooks are received in `daemon/api.rs::ingest_hook()` but only ring-buffered — they do NOT drive the state machine. Exit detection therefore lags up to 60s (reap_loop) and, when the tmux session is gone, `daemon/state/sessions.rs::reap_against()` DELETES the legacy registry entry rather than recording Stopped; the managed record only flips to Stopped on next daemon boot via `reconcile_on_boot()`.
- Detach vs exit is not distinguished (`SessionStatus::Detached` is dead code).

## Work items
1. **Store the Claude session id.** Add `claude_session_id: Option<String>` (`#[serde(default)]`) to `SessionRecord`. Capture it from the `SessionStart` hook payload in `ingest_hook()` and persist it on the matching managed session. NOTE: solve the hook→managed-session correlation (the hook JSON carries `session_id` + `cwd`; correlate to the SessionRecord by cwd/workspace_path, or inject a session marker via env when tm spawns the pane — investigate the existing correlation, e.g. `driver/correlation.rs`, and pick the reliable mechanism).
[...]

PR #4339, opened 2026-07-29
fix(trusty-mpm): stop a subagent SessionStart from clobbering claude_session_id

### LINKED ISSUE #4337: fix(trusty-mpm): daemon record claude_session_id captures a subagent's session id
**Defect**

The managed-session record for PM session 2eb72dca-… currently has `claude_session_id: 2d4df8a7-6289-46e9-a79b-e6104e056753` — a SUBAGENT's session id, not the PM conversation's. The next in-place relaunch that passes `--resume` would resume a subagent transcript instead of the PM's.

**Evidence**

`curl http://127.0.0.1:7880/api/v1/sessions/managed/2eb72dca-de08-481b-8dfa-22ab7f81b1f9` — read-only research pass against installed tm 1.2.3 (built from main aed80140), 2026-07-29. Confirmed live on current main, not a stale-binary artifact.

**Resolution**

Identify where claude_session_id is written (likely a hook fired from any claude process in the pane/cwd, including subagents) and scope it to the top-level session only.

bobmatnyc:
## Summary

- The daemon's `correlate_session_start` (`crates/trusty-mpm/src/daemon/api.rs`) matched an incoming `SessionStart` hook to a managed session purely by cwd. A subagent sharing the PM's cwd is its own top-level Claude Code process and fires its own `SessionStart`, which matched the PM's sole Active managed session and silently overwrote the PM's own `claude_session_id` with the subagent's — the next in-place `--resume` would then resume the subagent's transcript instead of the PM's conversation. Observed live: managed session `2eb72dca-...` carrying subagent session id `2d4df8a7-...`.
- Fix: only the FIRST correlation for a record is now trusted. A later `SessionStart` reporting a DIFFERENT id for the same cwd is refused and logged (likely a subagent); re-reporting the SAME id (a legitimate `--resume` relaunch, which preserves the Claude Code session UUID) remains a no-op success. The resume path already falls back to `--continue` via its own `session_id_exists` check when a stored id goes stale, so refusing an overwrite trades a possibly-stale id for never resuming the wrong transcript.
- Added two regression tests in `daemon/api_tests.rs`: `session_start_hook_does_not_overwrite_with_different_id` (reproduces the reported defect) and `session_start_hook_reasserting_same_id_is_a_noop` (proves the legitimate relaunch case still works).

## Root cause

`crates/trusty-mpm/src/daemon/api.rs`, `correlate_session_start` (~line 1342, the `matched.len() == 1` arm): it called `mgr.set_claude_session_id(&id, claude_session_id)` unconditionally on every `SessionStart` matching a record's cwd, with no check for whether the record already had a *different* id set.

## Test plan

- [x] `cargo test -p trusty-mpm` (full suite, lib + all integration binaries) — 0 failures
- [x] New tests pass: `daemon::api::tests::session_start_hook_does_not_overwrite_with_different_id`, `daemon::api::tests::session_start_hook_reasserting_same_id_is_a_noop`

Closes #4337

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -220,6 +220,17 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- A subagent's `SessionStart` hook can no longer overwrite a managed session's
+  `claude_session_id` with its OWN Claude session UUID
+  ([#4337](https://github.com/bobmatnyc/trusty-tools/issues/4337)). A subagent
+  sharing the PM's cwd is its own top-level Claude Code process and fires its
+  own `SessionStart`, which `correlate_session_start` matched to the PM's sole
+  Active managed session by cwd alone and blindly overwrote — so the next
+  in-place `--resume` would resume the subagent's transcript instead of the
+  PM's conversation. `correlate_session_start` (`daemon/api.rs`) now trusts only
+  the FIRST correlation for a record; a later `SessionStart` reporting a
+  DIFFERENT id for the same cwd is refused and logged, while re-reporting the
+  SAME id (a legitimate `--resume` relaunch) remains a no-op success.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -221,16 +221,26 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - A subagent's `SessionStart` hook can no longer overwrite a managed session's
-  `claude_session_id` with its OWN Claude session UUID
+  `claude_session_id` with its OWN Claude session UUID, and a legitimately
+  diverged id can no longer stay wrong forever
   ([#4337](https://github.com/bobmatnyc/trusty-tools/issues/4337)). A subagent
   sharing the PM's cwd is its own top-level Claude Code process and fires its
   own `SessionStart`, which `correlate_session_start` matched to the PM's sole
   Active managed session by cwd alone and blindly overwrote — so the next
   in-place `--resume` would resume the subagent's transcript instead of the
-  PM's conversation. `correlate_session_start` (`daemon/api.rs`) now trusts only
-  the FIRST correlation for a record; a later `SessionStart` reporting a
-  DIFFERENT id for the same cwd is refused and logged, while re-reporting the
-  SAME id (a legitimate `--resume` relaunch) remains a no-op success.
+  PM's conversation. `correlate_session_start` (moved to the new
+  `daemon::api::session_start_correlation` module to stay under `api.rs`'s
+  frozen SLOC budget) now trusts only the FIRST correlation for a record; a
+  later `SessionStart` reporting a DIFFERENT id for the same cwd is accepted
+  only when the stored id's transcript no longer resolves to a live session
+  (`runtime::session_id_exists`, the same staleness check `spawn_resume`
+  already trusts) — otherwise refused and logged — while re-reporting the
+  SAME id (a legitimate `--resume` relaunch) remains a no-op success. The
+  common divergence case is handled even earlier: `SessionEnd` now clears the
+  field via the new exact-match-guarded `SessionManager::clear_claude_session_id_if`
+  (so a later, unrelated `SessionStart` is never compared against a dead id),
+  and `SessionManager::mark_reactivated` clears it outright, since a
+  reactivation always precedes a fresh top-level relaunch into that pane.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/src/daemon/api.rs
+++ b/crates/trusty-mpm/src/daemon/api.rs
@@ -1338,7 +1338,33 @@ pub async fn ingest_hook(
 /// correlation is skipped with a warning to avoid mis-attribution. Single match →
 /// `SessionManager::set_claude_session_id`. Best-effort — failures are logged and
 /// silently swallowed so a missing correlation never blocks the hook response.
-/// Test: `session_start_hook_correlates_claude_id` in `api_tests.rs`.
+///
+/// # Subagent overwrite guard (#4337)
+///
+/// A subagent dispatched from a managed session's own pane (native `Task`/
+/// `Agent` tool, or any other `claude` invocation that happens to share the
+/// PM's cwd) is its OWN top-level Claude Code process from Claude Code's point
+/// of view: it gets its own internal session UUID and fires its own
+/// `SessionStart`, which lands here with the SAME `cwd` as the PM's managed
+/// session — because trusty-mpm never registers a subagent as its own managed
+/// session, cwd-matching alone finds exactly one (the PM's) record, the
+/// pre-#4337 ambiguity guard above never triggers, and the subagent's id
+/// silently clobbers the PM's own. The next in-place `--resume` would then
+/// resume the SUBAGENT's transcript instead of the PM's conversation.
+///
+/// The fix: once a record already carries a `claude_session_id`, a `SessionStart`
+/// reporting a DIFFERENT id for the same cwd is refused rather than accepted —
+/// only the FIRST correlation for a given record is trusted. A matching id is a
+/// harmless no-op write. This does mean a record's id can go stale if Claude
+/// Code ever legitimately reassigns a new UUID to the same pane's top-level
+/// conversation (e.g. a `--continue` fallback instead of `--resume`); that is
+/// an acceptable trade — the resume path (`runtime::claude_code::spawn_resume`)
+/// already treats a stored id as advisory and falls back to `--continue` via
+/// its own `session_id_exists` check when the stored id no longer resolves, so
+/// the worst case is a graceful fallback, never resuming the wrong transcript.
+/// Test: `session_start_hook_correlates_claude_id`,
+/// `session_start_hook_does_not_overwrite_with_different_id`,
+/// `session_start_hook_reasserting_same_id_is_a_noop` in `api_tests.rs`.
 async fn correlate_session_start(
     state: &Arc<DaemonState>,
     claude_session_id: &str,
@@ -1374,7 +1400,28 @@ async fn correlate_session_start(
     match matched.len() {
         0 => {} // no Active managed session at this cwd — silent, not an error
         1 => {
-            let id = matched[0].id;
+            let record = matched[0];
+            let id = record.id;
+
+            // #4337: an already-correlated record keeps its FIRST id. A report
+            // of a different id — most likely a subagent sharing this cwd, not
+            // the pane's own top-level process — is refused rather than
+            // silently accepted, so a subagent can never clobber the PM's id.
+            if let Some(existing) = record.claude_session_id.as_deref() {
+                if existing != claude_session_id {
+                    tracing::warn!(
+                        managed_id = %id,
+                        existing_claude_session_id = %existing,
+                        reported_claude_session_id = %claude_session_id,
+                        cwd = %cwd_str,
+                        "SessionStart: refusing to overwrite an already-correlated \
+                         claude_session_id with a different one — likely a subagent \
+                         sharing this cwd, not the pane's top-level process (#4337)"
+                    );
+                }
+                return;
+            }
+
             match mgr.set_claude_session_id(&id, claude_session_id).await {
                 Ok(()) => tracing::info!(
                     managed_id = %id,
--- a/crates/trusty-mpm/src/daemon/api.rs
+++ b/crates/trusty-mpm/src/daemon/api.rs
@@ -57,6 +57,24 @@ pub use control_routes::{
 pub mod claude_config_routes;
 pub use claude_config_routes::*;
 
+/// `SessionStart`/`SessionEnd` hook correlation (#1744, #4337).
+///
+/// Why: this cluster grew the #4337 subagent-overwrite guard + stale-id
+/// re-sync logic past what `api.rs`'s frozen 1272-SLOC allowlist budget
+/// (`.line-cap-allowlist.tsv`) could absorb. It is cohesive and
+/// self-contained — both hook handlers are called only from `ingest_hook`
+/// below — so it moves out, mirroring `control_routes`/`claude_config_routes`.
+/// Not `pub`: unlike those two, nothing outside `api.rs` calls these handlers
+/// directly (`ingest_hook` is the only public entry point), so a private
+/// `use` is enough to bring the names into `api.rs`'s own scope — and, via
+/// `api_tests.rs`'s `use super::*;`, into the test module too.
+mod session_start_correlation;
+use session_start_correlation::{correlate_session_start, handle_session_end};
+// Only `api_tests.rs` (via `use super::*;`) calls this directly — a plain
+// `use` would be flagged `unused_imports` on the non-test lib target.
+#[cfg(test)]
+use session_start_correlation::session_end_pane_still_live;
+
 /// The cross-session coordinator routes (`/api/v1/sessions/context` + `/chat`).
 ///
 /// Why: the coordinator's context and chat endpoints form their own cohesive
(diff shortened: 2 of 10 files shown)


bobmatnyc:
## Verdict: BLOCK

CI is red (see below) and there's a HIGH-confidence correctness regression in the fix itself. Neither is fatal on its own by the strict taxonomy (the code finding is HIGH, not CRITICAL), but together the PR is not mergeable as-is, so I'm reporting BLOCK rather than WARN.

## CI

`gh pr checks 4339` — **`500-line file-size cap` is FAILING**:
```
FAIL: crates/trusty-mpm/src/daemon/api.rs grew to 1287 SLOC (frozen budget 1272; cap is 500 prod cap). Split it.
```
The new `#4337` doc comment (~30 lines) pushed `api.rs` over its frozen ratchet budget. This is a hard, mandatory gate in this repo — must be split or the allowlist re-frozen (`scripts/check_line_cap.sh --update`) before merge, independent of the findings below.

## Findings

| HIGH | `crates/trusty-mpm/src/daemon/api.rs` | 1410-1423 | First-correlation-wins is **permanent, with no self-heal**: `SessionManager::set_claude_session_id` (`hook_sync.rs`) is the *only* writer of `claude_session_id`, and it is now only reachable from this one `None`-existing branch. Neither `resume()` (`session_manager/manager.rs:823`) nor `reactivate_managed_session` (`daemon/managed_routes/reactivate.rs`) ever clears the field. The function's own new doc comment (api.rs:1358-1364) acknowledges the stored id "can go stale if Claude Code ever legitimately reassigns a new UUID to the same pane's top-level conversation (e.g. a `--continue` fallback instead of `--resume`)" — but once that happens under this guard, the record's `claude_session_id` is now *wrong* and can never be corrected again for the session's remaining lifetime, because every future legitimate `SessionStart` reporting the real (different) id will hit this same refusal. Two production consumers key **exclusively** on exact `claude_session_id` equality with no fallback correlation: `handle_session_end` (api.rs:1538, `r.claude_session_id.as_deref() == Some(claude_session_id)`) and `idle_nudge::run_nudge` (idle_nudge.rs:132, identical match). Once the stored id drifts, `SessionEnd`'s fast-path Stopped transition silently stops firing for that session (masked only by the unrelated 60s runtime-exit reaper) and idle-nudge park detection permanently returns `NoManagedSession` — nudges silently stop working for that session's lifetime, with no log line pointing at the cause. Pre-fix "always overwrite" was buggy (the #4337 clobber) but also what kept these two correlations self-healing; this fix trades a subagent-clobber bug for a "one legitimate resync ever, then permanently desynced" bug. | Don't gate on "record already has *any* id"; gate on whether the *reporting* id is plausibly a re-report of the pane's own top-level process vs. a distinct process. E.g. before refusing, run the same `session_id_exists`-style check the resume path (`runtime/claude_code.rs:492`) already uses against the *existing* stored id — if the existing id's transcript file is gone/stale, treat the new report as an eligible resync rather than a suspect overwrite. A pure "first write ever wins" invariant has no way to distinguish "legitimate resync" from "subagent clobber," it just picks a different (and, for the two consumers above, more silently damaging) failure mode. |
| MEDIUM | `crates/trusty-mpm/src/daemon/api.rs` | 1410-1423 | Race ordering: the guard's correctness assumes the PM's own `SessionStart` always reaches the daemon before any subagent's. That holds by construction in the common in-progress-turn case (a subagent is only ever dispatched from a PM turn that's already running), but hook delivery is an independent, unordered HTTP POST per event. On a resume/relaunch specifically, if a subagent process from the prior life is still tearing down and its `SessionStart`-adjacent traffic (or a fast-dispatched subagent right after relaunch) lands before the PM's own resumed process's `SessionStart`, this guard now locks in the WRONG id permanently, with no later correction — strictly worse than pre-fix last-write-wins for that ordering, which would eventually self-correct when the PM's own event arrived. I can't clear 80% confidence this is common (the structural PM-dispatches-subagents ordering makes it rare), so not blocking on it alone, but it should be tracked rather than dismissed as impossible. | If addressing the HIGH finding via an existence-check-based resync, this mostly falls out for free: a wrongly-locked-in subagent id would fail its own `session_id_exists` check once the subagent's transcript is not the pane's active one, making it eligible for correction on the PM's subsequent (correct) report. |

## What's good

- Tests (`session_start_hook_does_not_overwrite_with_different_id`, `session_start_hook_reasserting_same_id_is_a_noop`) go through the actual daemon boundary — `ingest_hook(State(...), Json(post))`, the real HTTP handler — not a bare helper. `make_state_with_active_managed` correctly starts `claude_session_id: None` so both new tests genuinely exercise the guard's two branches.
- Verified the PR body's central claim myself rather than trusting it: `runtime/claude_code.rs:784` (`compose_resume_cmd`) and `:1093` (`ClaudeCodeAdapter::spawn_resume`) do filter a stored id through `session_id_exists` and fall back to `--continue`/fresh-start when it doesn't resolve — that part of the resume-path safety-net claim checks out.
- Closing-keyword check: `gh pr view 4339 --json closingIssuesReferences` returns exactly one entry, `#4337`. No stray auto-close.

## Required Changes

1. Fix the SLOC cap failure (split `api.rs` or re-freeze the ratchet, per project convention — see CI section above).
2. Address the HIGH finding: give the guard a way to accept a legitimate resync (e.g. existence-check the *existing* stored id before refusing a differing report), so `handle_session_end` and `idle_nudge::run_nudge` correlation can't permanently desync from the real running pane.
3. Consider the MEDIUM ordering note in the same fix, or explicitly document the accepted risk if the fix for (2) doesn't also cover it.

bobmatnyc:
Pushed fix commit 6d92b7cf addressing all three findings:

1. **RED CI (line cap)**: extracted `correlate_session_start`, `handle_session_end`, and `session_end_pane_still_live` into a new module `crates/trusty-mpm/src/daemon/api/session_start_correlation.rs`. `api.rs` is back under its frozen 1272-SLOC budget (`scripts/check_line_cap.sh` reports `0 violations`).

2. **HIGH (permanent divergence)**: `correlate_session_start` now accepts a differently-reported id when the *stored* one no longer resolves to a live transcript, reusing `runtime::claude_code::session_id_exists` (now `pub(crate)`, re-exported as `runtime::session_id_exists`) — the same staleness check `spawn_resume` already trusts before honoring `--resume`. A subagent's own transcript stays live for as long as it runs, so it's still refused. Additionally: `SessionManager::clear_claude_session_id_if` (new, exact-match-guarded) clears the field on `SessionEnd`, and `SessionManager::mark_reactivated` clears it outright on reactivate — both close the common-case divergence path proactively rather than waiting on the staleness check.

3. **MEDIUM (ordering race)**: cured by the same `SessionEnd` clear — if a subagent's `SessionStart` wins the race, the wrong id self-heals to `None` as soon as that subagent's own `SessionEnd` fires, rather than staying locked in forever.

New tests: `session_start_hook_still_refuses_a_live_subagent_id`, `session_start_hook_re_correlates_a_stale_id`, `session_end_hook_clears_claude_session_id`, `session_start_hook_ordering_race_self_heals_via_session_end`, `clear_claude_session_id_if_clears_on_exact_match`, `clear_claude_session_id_if_leaves_a_different_id_untouched`, `mark_reactivated_clears_stale_claude_session_id`.

Full gate: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p trusty-mpm` (full suite, not `--lib`-only) — all clean, 0 failures across every test binary (3952 lib tests + all integration binaries).

bobmatnyc:
## Verdict: BLOCK (CI red — trivial fix, code itself is now clean)

Re-reviewed commit 6d92b7cf against my prior BLOCK. Both substantive findings from the last round are resolved with actual verification, not just taken on the author's word. One new, unrelated CI failure was introduced by the extraction and must be fixed before merge.

## CI

- `500-line file-size cap` — now **PASS**. Confirmed the extraction (`daemon/api/session_start_correlation.rs`) actually pulled `api.rs` back under budget.
- `Test: doc-comment pointer lint` — **NEW FAILURE**:
  ```
  FAIL: crates/trusty-mpm/src/daemon/api/session_start_correlation.rs:67: Test: pointer cites
  `session_start_hook_does_not_overwrite_with_different_id` — no fn/mod with that name found.
  ```
  The moved `correlate_session_start` doc comment (session_start_correlation.rs, the `Test:` block ending at line ~352 in the diff) still lists `session_start_hook_does_not_overwrite_with_different_id`. That test was deleted in this same commit — `api_tests.rs` even leaves a comment saying it's "superseded by `session_start_hook_still_refuses_a_live_subagent_id`" — but the doc pointer was never updated to match. One-line fix: drop that name from the `Test:` list (the four remaining names are all real). All other checks pass or were still running when checked (Clippy, `Test`, `trusty-search daemon smoke test` were mid-run at last check with nothing else red).

## Findings from the prior review — verified, not just re-read

**Original HIGH (permanent divergence once an id legitimately changes) — RESOLVED.** I traced the actual call graph rather than trusting the PR body:
- `stored_id_still_live` (session_start_correlation.rs:455) resolves its config dir via `crate::core::trusty_tools_config::managed_claude_config_dir()`.
- `spawn_resume`'s own staleness check (`runtime/claude_code.rs:1076`, `prepare_managed_config` → `runtime/claude_code.rs:618`) resolves config dir via the exact same `managed_claude_config_dir()` call — confirmed by reading `prepare_managed_config`'s body, not inferring it from the doc comment. Both are one single machine-wide directory (`crate_config_dir("trusty-mpm").join("claude-config")`), not per-record/per-pane — so there is no way for the two call sites to disagree about what's "stale."
- Subagents inherit `CLAUDE_CONFIG_DIR` from their parent (tm-managed) process rather than getting their own, which is consistent with the original #4337 report itself (the subagent's id was findable via ordinary cwd-matching against the PM's own managed record) — so a live subagent transcript is visible to `stored_id_still_live` and correctly refused, and a genuinely stale id is correctly accepted. `session_start_hook_still_refuses_a_live_subagent_id`'s fixture (`seed_fake_transcript`) reproduces the real path layout (`~/.trusty-tools/trusty-mpm/claude-config/projects/<encoded-cwd>/<id>.jsonl`) exactly — verified `CRATE_NAME = "trusty-mpm"`, `MANAGED_CLAUDE_CONFIG_SUBDIR = "claude-config"`, and `encode_project_dir`'s `/`→`-` encoding all match what the test hand-builds.
- `handle_session_end` now clears the field via `clear_claude_session_id_if`, and `mark_reactivated` clears it unconditionally — both real self-heal paths for the two ways the id legitimately goes stale (clean end, reactivate-then-relaunch).

**Original MEDIUM (subagent wins the race) — RESOLVED for the tested case.** `session_start_hook_ordering_race_self_heals_via_session_end` shows a race-won wrong id correcting to `None` once that same subagent's own `SessionEnd` fires. I don't have a higher-severity residual concern here at >80% confidence — see Notes for the one compound edge case I can't fully rule out.

**`clear_claude_session_id_if` exact-match guard — verified against a concurrent-write race, not just read.** `clear_claude_session_id_if` (`hook_sync.rs:973`) re-`get()`s the record at the moment of the clear rather than trusting `handle_session_end`'s earlier `mgr.list()` snapshot, and only clears when the freshly-read value still equals `expected`. So if a fresher `SessionStart` correlation lands between `handle_session_end`'s match and its clear call, the clear becomes a no-op and the fresher id survives. `clear_claude_session_id_if_leaves_a_different_id_untouched` covers this directly. No path found where a stale `SessionEnd` can clobber a newer legitimate correlation.

## Notes

- Residual (not blocking, under 80% confidence as a real production occurrence): if a race-won wrong id's *own* `SessionEnd` never fires (an ungraceful kill of that specific subagent process, not just any exit) AND the record never goes through `mark_reactivated` in the meantime, the wrong id stays locked in until its transcript is eventually pruned from disk. This requires two independent low-probability conditions to compound (lost the ordering race AND an ungraceful exit of the winner), and is already strictly better than either the pre-fix or the prior-round-fix behavior. Worth a one-line comment acknowledging it, not a required change.

## Required Changes

1. Fix the dangling `Test:` pointer in `crates/trusty-mpm/src/daemon/api/session_start_correlation.rs` (~line 67, the `correlate_session_start` doc block) — remove the reference to the deleted `session_start_hook_does_not_overwrite_with_different_id` test so `scripts/check_test_pointers.sh` passes.

bobmatnyc:
## Verdict: BLOCK (not on code — on mergeability; CI has not run on 339b8b0b yet)

## Diff check — matches what was described

`git diff 6d92b7cf 339b8b0b` is exactly the claimed 2 insertions / 3 deletions, all in `crates/trusty-mpm/src/daemon/api/session_start_correlation.rs`'s `Test:` doc block:
```
-/// `session_start_hook_does_not_overwrite_with_different_id`,
+/// `session_start_hook_still_refuses_a_live_subagent_id`,
-/// `session_start_hook_re_correlates_a_stale_id`,
-/// `session_start_hook_still_refuses_a_live_subagent_id` in `api_tests.rs`.
+/// `session_start_hook_re_correlates_a_stale_id` in `api_tests.rs`.
```
Confirmed nothing else changed (`git diff --stat 6d92b7cf 339b8b0b` shows only this one file, 1 changed). The doc pointer now correctly names live tests only.

## CI — cannot confirm green, and here's why

`gh pr checks 4339` currently reports **"no checks reported"** for this branch, and `gh api .../commits/339b8b0bb753.../check-runs` returns **zero check-runs** for head `339b8b0b` — not pending, not queued, nothing has been dispatched yet, several minutes after the push.

Root cause: `gh pr view 4339 --json mergeable,mergeStateStatus` reports `"mergeable": "CONFLICTING"`, `"mergeStateStatus": "DIRTY"`. Most of this repo's checks trigger on `pull_request` against the `refs/pull/4339/merge` ref — GitHub can't materialize that ref while the PR conflicts with `main`, so none of those workflows fire at all. I verified the conflict directly (`git merge-tree $(git merge-base origin/main pr-head) origin/main pr-head`): it's a single-file, single-hunk conflict in `crates/trusty-mpm/CHANGELOG.md` — both this branch and something since merged into `main` inserted an entry immediately under the same `### Fixed` header. `crates/trusty-mpm/src/daemon/api.rs` and everything else merges cleanly.

This is not a code defect and not something introduced by 339b8b0b's own content — it's the branch going stale against `main`. But it means I cannot verify "pointer-lint CI green" because pointer-lint (or anything else) has not run at all on this head, and the PR is not currently mergeable regardless.

## Required Changes

1. Rebase/merge `main` into this branch to resolve the single `CHANGELOG.md` conflict (keep both entries under `### Fixed`).
2. Push, let CI actually run on the resulting head, and confirm `Test: doc-comment pointer lint` (and the rest) come back green.

## Notes

All substantive code findings from the two prior rounds remain verified-resolved — nothing in this commit touches that surface, it's a pure doc-string fix. Once the rebase is pushed and CI is actually green (line-cap + pointer-lint both passing, no new failures), this should be APPROVE with no further re-review of the correlation logic needed — only confirmation that the resolved CI is clean on the new head.

bobmatnyc:
## Verdict: APPROVE

Final round, focused on what changed since the prior 2 critic passes (which already cleared the core logic: staleness-gated re-correlation, SessionEnd self-heal, exact-match clear guard).

No issues found at >80% confidence. APPROVED for merge sequencing.

### Checks performed
- CI at head `a0a76d48`: all 23 required checks pass, "Notify on red main" correctly skipped (non-red main).
- `mergeStateStatus: BEHIND` / `mergeable: MERGEABLE` — branch is behind main but has no merge conflicts. Not a code defect; flagging for the PM to sequence a rebase/merge, not fixing here.
- Traced `stored_id_still_live` (crates/trusty-mpm/src/daemon/api/session_start_correlation.rs) → `runtime::session_id_exists` (now `pub(crate)`, crates/trusty-mpm/src/runtime/claude_code.rs:512) and confirmed it's the exact same function `spawn_resume` already calls at claude_code.rs:971 — the "two call sites can never disagree about staleness" claim in the doc comments holds.
- Verified the `SessionEnd` clear (`clear_claude_session_id_if`, session_manager/hook_sync.rs) and `mark_reactivated`'s unconditional clear (session_manager/reactivate.rs:114) are both exact-match/unconditional in ways consistent with their stated invariants, and each has a dedicated test (`clear_claude_session_id_if_clears_on_exact_match`, `..._leaves_a_different_id_untouched`, `mark_reactivated_clears_stale_claude_session_id`).
- Confirmed the module extraction (`daemon::api::session_start_correlation`) is a pure move — diff shows the old `api.rs` block deleted and reappearing verbatim (plus the new re-sync logic) in the new file; visibility (`pub(super)`) matches actual call sites (only `ingest_hook` and `api_tests.rs` via `use super::*`).
- Reviewed new tests: `session_start_hook_still_refuses_a_live_subagent_id`, `session_start_hook_re_correlates_a_stale_id`, `session_end_hook_clears_claude_session_id`, `session_start_hook_ordering_race_self_heals_via_session_end` — these exercise exactly the contract described (refuse-while-live, accept-when-stale, clear-on-end, race self-heal), matching the postconditions claimed in the doc comments.

### Notes
- Pre-existing TOCTOU: `correlate_session_start` reads a `mgr.list()` snapshot, decides, then writes via `set_claude_session_id` — a concurrent write between snapshot and write could theoretically race. This pattern predates this PR (present in the original #1744 code) and isn't introduced or worsened here; not blocking.
- Branch BEHIND main: no action taken per instructions — reporting only, PM sequences the merge.

Issue #4335, opened 2026-07-29
fix(trusty-mpm): bare tm needs two runs after Claude Code backgrounds — nested_session_guard silently fails open


bobmatnyc:
**Defect**

Exiting Claude Code via /exit (which BACKGROUNDS, never firing SessionEnd) leaves the daemon record `active`. On the next bare `tm`:

- Gate A (`plan_inplace`, guided_inplace.rs:249-254) requires record_state == "stopped" so it skips.
- Gate B (`nested_session_guard`, guided.rs:112, driver 440-464) should catch it via pane_id match, but `list_all_managed_sessions` (guided.rs:599-611) has a 2s timeout and `.ok()?` — on a cold daemon computing per-session `stale_assets` for 57 sessions it exceeds 2s and the guard silently no-ops (no log at all).

Flow falls through to the picker, whose resume dead-ends on the #2678 switch-client refusal. Invocation 1 reconciles the record + warms the daemon; invocation 2 works.

**Evidence**

Verbatim repro captured in the session scrollback (read-only research pass against installed tm 1.2.3, built from main aed80140, 2026-07-29). Confirmed live on current main, not a stale-binary artifact.

Not covered by:
- #4061/#4083 — patched only the CwdProject::NotGit arms
- #4305 — daemon-unreachable double-provisioning, different path
- #4304 — fixed plan_resume's runtime_live, downstream of the guard that never ran

**Resolution**

Make Gate B reliable: raise the timeout on this one call, log the error arm (`tracing::debug!` at minimum), and add a slim query mode that skips the stale_assets probe the guard never reads (folds into #4322). Do NOT widen Gate A's stopped-only check.

Cross-ref: #2678, #4061, #4083, #4304, #4305, #4322

bobmatnyc:
Cross-ref: #6765 is the command-builder layer underneath this one — even with Gate B reliable, `resume_command` decides `--continue` eligibility against `~/.claude/projects` while the spawned `claude` resolves it against `$CLAUDE_CONFIG_DIR/projects`, so a backgrounded job in the managed store is picked and the pane drops to a bare shell. Not a widening of Gate A's stopped-only check. #6766 tracks the post-spawn verification net.

Issue #2246, opened 2026-07-08
bug(trusty-mpm): managed-session /login loops ("Login successful" → "Not logged in") — DOC-34 CLAUDE_CONFIG_DIR relocation breaks macOS Keychain auth


bob-duetto:
**Symptom:** In a tm-managed tmux session, `/login` reports "Login successful" then immediately "Not logged in · Please run /login", looping; `resume` also says "Not logged in". Bob is blocked from using managed sessions.

**Root cause (confirmed by source + live inspection + Claude Code docs):** Claude Code stores its primary OAuth session token in the macOS **Keychain**, under a service name suffixed with `first-8-hex(SHA-256(CLAUDE_CONFIG_DIR))`. Default (no relocation) = bare `Claude Code-credentials`; a relocated `CLAUDE_CONFIG_DIR` = `Claude Code-credentials-<hash>`. DOC-34 (#1996/#2002, commit `7c2d13dc`) was the change that first pointed **daemon-managed** sessions at a tm-owned relocated `CLAUDE_CONFIG_DIR` (`~/.trusty-tools/trusty-mpm/claude-config/`). Before DOC-34, managed sessions ran against the ambient logged-in `~/.claude`, so auth worked. The relocation changed auth from "ambient" to "isolated", and the assumption that auth still round-trips under a *relocated* dir (with `--setting-sources project,local`) was carried over unverified. Live inspection found: `.credentials.json` in the managed dir holds only MCP sub-tokens (not primary auth); env propagation (`CLAUDE_CONFIG_DIR`/`TM_MANAGED_SESSION_ID`) is correct in every pane; the machine's Keychain has a bare `Claude Code-credentials` entry (the real login) plus ~250 `Claude Code-credentials-<hash>` entries. There is a documented Claude Code bug in this area (anthropics/claude-code #9403: write/read service-name mismatch).

**Investigation status (Bob's call: check claude version FIRST):** determining whether this is the version-specific #9403 defect (fixable by updating the `claude` binary) vs the inherent CLAUDE_CONFIG_DIR-hash scoping (needs a credential-strategy fix). Machine runs claude ~2.1.201.

**Candidate fixes:**
1. Update `claude` if #9403 is fixed in a newer version and 2.1.201 is affected (cheapest — no tm change).
[...]

PR #4340, opened 2026-07-29
fix(trusty-mpm): keep the PM persona on in-place relaunch and test the exec seam

### LINKED ISSUE #4336: fix(trusty-mpm): in-place relaunch execs claude with no flags — bypass, setting-sources, and PM persona all dropped
**Defect**

The in-place relaunch path (`run_inplace_relaunch`, guided_inplace.rs:581-596) launched claude with ZERO arguments — observed `ps -ww` argv: bare `/Users/masa/.local/bin/claude`. A healthy managed pane shows `--append-system-prompt-file … --setting-sources project,local --dangerously-skip-permissions`.

Transcripts confirm: pre-exit session permissionMode=bypassPermissions, post-relaunch session default/acceptEdits.

`compose_inplace_args` (claude_code.rs:779-801) unconditionally prepends the flags and cannot return empty, so the loss is at or after the exec seam. This is the only launch path that bypasses the `tm internal-spawn-disclaimed` wrapper, and `run_inplace_relaunch` is explicitly not unit-tested (guided_inplace.rs:487).

Separately, `compose_inplace_args` omits `--append-system-prompt-file` BY DESIGN, so even a working in-place relaunch loses the PM persona.

**Evidence**

Read-only research pass against installed tm 1.2.3 (built from main aed80140), 2026-07-29. Confirmed live on current main, not a stale-binary artifact.

Not covered by:
- #2020 — same four dropped properties but scoped to POST /claude-config/restart → relaunch_command(), spawn_command.rs:60, bare `claude`; different call site
- #2125 — fixed spawn/resume, closed

**Resolution**

Reproduce once with RUST_LOG=debug to capture the actual argv. Refactor the Command construction into a pure builder and add a regression test asserting get_args() contains both flags. Add `--append-system-prompt-file` to compose_inplace_args.

Cross-ref: #2020, #2125

bobmatnyc:
Closes #4336

## What the trace found

The reported mechanism — "`run_inplace_relaunch` execs claude with ZERO args" —
**is not reachable from this code.** `compose_inplace_args` cannot return empty
(both flag constants are non-empty literals prepended unconditionally), and it
has looked exactly like that since it was introduced in dba21950 (#2027), which
added the exec path and the flags in the same commit. Nothing between the
working and broken observations touched it: the only commits after 2026-07-27
against `guided_inplace.rs` / `claude_code.rs` / `model_inject.rs` are #4209 and
#4205, neither of which goes near the argv.

Live `ps -Eww` on the reporting machine confirms the flags survive the exec
seam. Three concurrent in-place relaunches, each with the signature of this path
(absolute `resolve_claude` binary, tm-owned `CLAUDE_CONFIG_DIR`, cwd = the
managed workspace):

```
21504  /Users/masa/.local/bin/claude --setting-sources project,local --dangerously-skip-permissions
93512  /Users/masa/.local/bin/claude --setting-sources project,local --dangerously-skip-permissions
70700  /Users/masa/.local/bin/claude --setting-sources project,local --dangerously-skip-permissions
```

That argv is byte-for-byte `compose_inplace_args`' output for the no-resume
case. The one bare `/Users/masa/.local/bin/claude` (PID 10166) differs from
those three in exactly one way: `CLAUDECODE=1` is present in its **exec-time**
environment, so it was exec'd from inside a Claude Code session rather than by
a `tm` invoked from the pane shell. No `tm` code path produces it — the other
two spawners build shell strings that also always carry the flags, and the only
other `Command::new` for claude (`internal-spawn-disclaimed`) forwards argv
verbatim and would appear as its own parent process. The likeliest remaining
producer is Claude Code's own `/exit`-background re-exec of `process.execPath`,
which is also the premise of #4335 and would explain the
`bypassPermissions` → `default` transcript evidence.

**So this PR does not "restore" dropped flags — nothing was dropping them.** It
fixes the defect that IS real in this path and closes the coverage hole that
made the report irrefutable.

## What is actually fixed

**1. The PM persona was genuinely dropped.** `compose_inplace_args` omitted
`--append-system-prompt-file` by design, so an in-place relaunch restored the
operator into vanilla Claude Code — the same defect #2125/#2230 fixed for spawn
and resume, left standing here. Visible in the same `ps` sweep: the spawn-path
process carries the prompt file, the in-place ones do not. It now sources the
prompt through the same `build_prompt_file` carrier, pushing the path as its own
argv token rather than through `prompt_file_flag`'s shell quoting — no shell
splits this argv, so quoting would embed literal quotes in the filename.

**2. The exec seam had no coverage.** `guided_inplace.rs:487` said so outright.
The `Command` construction moved out of `run_inplace_relaunch` into a pure
`build_inplace_exec_command`, leaving `exec` as a one-line untestable tail. The
step that decides what argv and environment `claude` receives is now asserted:

- `inplace_exec_command_forwards_every_arg_in_order` — argv verbatim, in order,
  plus program and cwd. An empty `get_args()` here is the reported defect.
- `inplace_exec_command_scrubs_api_key_and_sets_auth_env` — `ANTHROPIC_API_KEY`
  removed, `CLAUDE_CONFIG_DIR` + `CLAUDE_CODE_OAUTH_TOKEN` set (DOC-34, #2246).
- `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end` — both
  flags AND a readable prompt file, composed through the real builder.
- `compose_inplace_args_carries_prompt_file_unquoted` — a path with a space
  stays one unquoted token.

`build_inplace_resume_command_carries_prompt_file` verified by mutation:
forcing `prompt_file` to `None` makes it FAIL.

## Quality gate

```
cargo fmt --check exit: 0

cargo clippy -p trusty-mpm --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 33.21s

cargo test -p trusty-mpm     (full suite, not --lib)   exit: 0
test result: ok. 3947 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 22.09s
test result: ok. 1287 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.64s
test result: ok. 1287 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 3.92s
(44 `test result: ok` lines total; zero FAILED, zero panics)

scripts/check_line_cap.sh: line-cap: 7 allowlisted, 0 violations — OK.
```

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -220,6 +220,26 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The bare-`tm` in-place relaunch keeps the PM persona
+  ([#4336](https://github.com/bobmatnyc/trusty-tools/issues/4336)).
+  `compose_inplace_args` omitted `--append-system-prompt-file` by design, so the
+  one launch path that `exec`s `claude` directly — instead of sending a command
+  string to a tmux pane — restored the operator into vanilla Claude Code. It now
+  builds the prompt through the same `build_prompt_file` carrier `spawn` and
+  `spawn_resume` use (#2125, #2230), pushing the path as its own argv token
+  rather than through `prompt_file_flag`'s shell quoting, which would embed
+  literal quotes in a filename no shell is present to strip.
+- The argv and environment handed to that `exec` are now under test. The
+  `std::process::Command` construction moved out of `run_inplace_relaunch` into
+  a pure `build_inplace_exec_command`, leaving `exec` as a one-line untestable
+  tail. Previously the last step before process replacement — the step that
+  decides whether `--setting-sources project,local` and
+  `--dangerously-skip-permissions` reach `claude` at all — was the only launch
+  step in the codebase with no coverage, which is why #4336 could be reported
+  and not refuted from the suite. Guarded by
+  `inplace_exec_command_forwards_every_arg_in_order`,
+  `inplace_exec_command_scrubs_api_key_and_sets_auth_env`, and
+  `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end`.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
@@ -605,9 +605,9 @@ pub(crate) async fn run_inplace_relaunch(
 /// the tmux-pane spawn/resume paths so an in-place relaunch does not silently
 /// drop back into the `CLAUDE_CONFIG_DIR`-keyed Keychain login loop the token
 /// exists to bypass. Pure — builds and returns, never spawns.
-/// Test: `inplace_exec_command_carries_isolation_flags`,
-/// `inplace_exec_command_carries_prompt_file_and_resume_selection`,
-/// `inplace_exec_command_scrubs_api_key_and_sets_config_dir`.
+/// Test: `inplace_exec_command_forwards_every_arg_in_order`,
+/// `inplace_exec_command_scrubs_api_key_and_sets_auth_env`,
+/// `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end`.
 pub(crate) fn build_inplace_exec_command(
     resume: &trusty_mpm::runtime::InPlaceResumeCommand,
     cwd: &std::path::Path,
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
@@ -490,7 +490,9 @@ pub(crate) enum InPlaceOutcome {
 /// rather than proceeding; (5) execs `claude` in place.
 /// Test: `run_inplace_relaunch_never_reactivates_when_command_build_fails`,
 /// `run_inplace_relaunch_falls_through_on_gutted_worktree`,
-/// `run_inplace_relaunch_serves_live_linked_worktree`. The exec itself is not
+/// `run_inplace_relaunch_serves_live_linked_worktree`. The argv/environment the
+/// final step hands to `exec` is pinned by [`build_inplace_exec_command`]'s own
+/// tests (#4336). The exec CALL itself is still not
 /// unit-tested (it replaces the process image); the gate-2a tests pin the
 /// decision by asserting the daemon mock records ZERO hits, which is only true
 /// when the fall-through happened before any daemon mutation. The FSM invariant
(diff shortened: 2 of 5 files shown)


bobmatnyc:
## Verdict: WARN

CI is currently red — `Test: doc-comment pointer lint` fails (confirmed via `gh run view 30487456615 --log-failed`), so this cannot merge as-is regardless of the WARN/HIGH-only label below. Fix the dangling pointers and the doc/test claims check out; the persona and exec-seam logic itself is sound.

## Findings

| HIGH | `crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs` | 608-610 | `build_inplace_exec_command`'s doc comment cites three test names that don't exist anywhere in the crate — `inplace_exec_command_carries_isolation_flags`, `inplace_exec_command_carries_prompt_file_and_resume_selection`, `inplace_exec_command_scrubs_api_key_and_sets_config_dir`. CI's `check_test_pointers.sh` catches exactly this and is currently failing on this PR (job [90696604494](https://github.com/bobmatnyc/trusty-tools/actions/runs/30487456615/job/90696604494)): "3 dangling pointer(s) ... FAILED." | Point at the tests that actually exist: `inplace_exec_command_forwards_every_arg_in_order`, `inplace_exec_command_scrubs_api_key_and_sets_auth_env`, `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end` (the same three names already correctly used in the CHANGELOG entry for this PR). |

## Notes

Verified independently, not just read:

- **Prompt-file lifetime**: `build_prompt_file` (unchanged, reused verbatim) writes synchronously via `std::fs::write` and returns a plain `PathBuf` — no `NamedTempFile`/`Drop`-based cleanup anywhere in the codebase for this carrier. The write happens before `Command` construction, in the same process, before `exec()` replaces the image — there's no window for the file to be deleted before the exec'd process reads it, and this is identical to how `spawn_command`/`resume_command` already use the same carrier. Not a new risk.
- **Unquoted argv token is correct, not a regression**: `compose_inplace_args` deliberately does NOT run the prompt path through `shell_single_quote` the way `prompt_file_flag` does for the tmux-string paths — correct, since this argv goes straight to `execv` with no shell to re-split it; quoting here would embed literal quote characters in the filename.
- **Pure-builder refactor's error path**: diffed the before/after in `run_inplace_relaunch` — identical `Command` construction (program, args, cwd, env scrub/injection) now lives in `build_inplace_exec_command`, called as `exec_claude_in_place(build_inplace_exec_command(&resume, &cwd))`. `exec_claude_in_place` itself is untouched. No behavior change on the exec-failure path.
- **Mutation check performed directly** (not just read): checked out the PR head into a scratch worktree, deleted the `PERMISSION_MODE_FLAG` half of `compose_inplace_args`' isolation-flag push (i.e. dropped `--dangerously-skip-permissions`), and ran `cargo test -p trusty-mpm --lib compose_inplace_args`. Result: `compose_inplace_args_carries_prompt_file_unquoted` FAILED on the assertion `args.contains(&"--dangerously-skip-permissions".to_owned())` — this test is unconditional (no `resolve_claude()`/real-`claude`-binary gate), so it runs on every CI machine. So while `build_inplace_resume_command_carries_prompt_file` itself only pins the prompt-file token (as the PR body's mutation claim narrowly says), a *different* new test in the same PR does pin the isolation flags, and it isn't gated out on CI. The coverage gap the question raised does not exist in practice.
- `gh pr view 4340 --json closingIssuesReferences` returns only #4336 — no unintended auto-close.

bobmatnyc:
## Verdict: APPROVE

Final verdict round for PR #4340 at head `30bdff7c`. CI: 23 checks pass, 1 skip ("Notify on red main" — correctly skips on non-main branches). No red checks.

No issues found at >80% confidence.

Reviewed the diff against the rubric with fresh eyes (regression-focused, not re-litigating prior rounds' cleared substance):

- `build_inplace_exec_command` cleanly extracts the previously-untestable tail of `run_inplace_relaunch` into a pure function; `exec_claude_in_place(cmd)` remains the only untested one-liner, which is correct — it replaces the process image and can't be unit tested.
- `compose_inplace_args` now takes `prompt_file: Option<&Path>` and pushes `--append-system-prompt-file <path>` as its own argv token, explicitly NOT shell-quoted — correct, since this path execs directly via `execv` with no shell to un-quote a `shell_single_quote`-wrapped token. Verified against `prompt_file_flag` (the shell-string sibling) — no accidental reuse.
- `build_prompt_file` failure is non-fatal (confirmed at `claude_code.rs:318-333`): a `None` from `write_prompt_file` logs a warning and omits the flag rather than passing an empty/invalid path — matches the doc comment's claim exactly.
- Test coverage is contract-driven and non-overlapping: `inplace_exec_command_forwards_every_arg_in_order` pins argv fidelity, `inplace_exec_command_scrubs_api_key_and_sets_auth_env` pins the env seam (API key removal + config dir + oauth token injection per #2246), `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end` composes through the real builder end-to-end (correctly `#[serial_test::serial]` + skips without a real `claude` install, matching the file's established convention). `compose_inplace_args_carries_prompt_file_unquoted` uses a path containing a space specifically to catch any accidental reintroduction of shell quoting — a well-targeted regression guard.
- CHANGELOG entry accurately describes the fix and the coverage gap it closes; no scope creep into unrelated launch paths.

No CRITICAL, HIGH, MEDIUM, or LOW findings at >80% confidence. APPROVED for merge sequencing.

Note for the PM: `mergeStateStatus` is `BEHIND` main — not touched here per instruction; sequence an update-branch/rebase before merge.

Issue #2678, opened 2026-07-14
bug(trusty-mpm): picker launch-new from inside tmux prints 'attaching' then hangs (nested attach-session instead of switch-client); switch-client handoff also independently found to orphan/hijack clients


bobmatnyc:
## Symptom

Reported by Bob (tm 0.19.11 installed), from **inside** an existing tmux client (session `tm-cto-01`). Ran bare `tm`, the guided picker's default fallback showed `[1] resume tm-cto-01`, `[2] launch new session`; he chose `2`:

```
tm:   [1] resume tm-cto-01 (active)
tm:   [2] launch new session
tm:   [d<N>] delete session N (e.g. d1)
tm:   [q] quit
tm: default: [1] resume most recent
tm: > 2
tm: session 'tm-cto-02' created (active)
tm: attaching to session 'tm-cto-02'
```

After the `attaching to session` line, the client **dropped out and hung** — no shell, no error, unresponsive.

## Code path

`guided.rs::run_tty_picker`'s `PickerDecision::LaunchNew` arm calls
`guided_launch::launch_new_session_and_attach`, which — on both the
synchronous-ack path and the async poll-until-ready path — ends by calling
`crate::commands::tmux_attach::tmux_attach(&name)`
(`crates/trusty-mpm/src/bin/tm/commands/guided_launch.rs`).

`tmux_attach()` (`crates/trusty-mpm/src/bin/tm/commands/tmux_attach.rs:68-92`)
branches purely on `inside_tmux()` (a `$TMUX` env-var check):

```rust
pub(crate) fn tmux_attach(name: &str) -> anyhow::Result<()> {
    let inside = inside_tmux();
    let argv = attach_argv(name, inside);
    ...
    if inside {
        eprintln!("tm: switching client to session '{name}'");
    } else {
        eprintln!("tm: attaching to session '{name}'");
    }
    let status = std::process::Command::new(&tmux_bin).args(&argv).status()...
```

Bob's transcript prints **"attaching to session"**, i.e. `inside_tmux()`
evaluated `false` even though he was launching from inside an attached tmux
client (`tm-cto-01`) — which routes into the plain `tmux attach-session`
branch. `tmux attach-session` run from inside an existing tmux client is the
classic nested-attach hazard this very module's doc comment describes
("tmux refuses attach-session when invoked from inside an existing tmux
client… `Error: tmux attach-session exited with failure`"). If tmux's refusal
happens after the child process has already started reconfiguring the tty
[...]

PR #4343, opened 2026-07-29
fix(trusty-mpm): make the nested-session guard survive a cold daemon

### LINKED ISSUE #4335: fix(trusty-mpm): bare tm needs two runs after Claude Code backgrounds — nested_session_guard silently fails open

bobmatnyc:
Closes #4335

## The failure

`nested_session_guard` fetched the managed-session list with a 2-second timeout
and a bare `.ok()?`. On a cold daemon the per-session `stale_assets` pass over a
large fleet exceeded that bound, the request was cancelled, and the guard
no-opped — with nothing in any log to say it had not run.

The rest follows from that. Claude Code's `/exit` BACKGROUNDS the session and
never fires `SessionEnd`, so the daemon record stays `active`; Gate A
(`plan_inplace`) correctly skips, because it requires `stopped`. Gate B was
supposed to catch it on `pane_id` — and silently didn't. The flow fell to the
picker, whose resume dead-ends on the #2678 switch-client refusal. Invocation 1
reconciled the record and warmed the daemon; invocation 2 worked.

A guard that silently stops guarding is worse than a slow one. #4335 needed a
live repro plus a code read to establish the guard had not run, because every
failure mode was observationally identical to "nothing matched".

## The fix — three changes, smallest lever first

**`?slim=true` on the managed-list endpoint** (folds into #4322). The guard
reads `name`, `pane_id`, `state`, and `id` — never `stale_assets`. That flag is
the most expensive thing the endpoint computes: a filesystem-bound catalog
compose per distinct `(agent_source, skill_source)` pair plus a per-session
manifest/on-disk comparison. The guard was paying for a field it discards.
Opting out is the actual fix. Absent, `slim=false`, or malformed keeps the full
probe, so no existing client changes shape.

**Timeout 2s → 8s** on that one call. The 2s bound exists to stop a *hung*
daemon stalling bare `tm`; 8s still bounds that while clearing a cold-start
listing. This is the belt to slim mode's braces, not the primary fix.

**The error arm is logged** at `tracing::debug!` via a new
`guard_managed_sessions`, which keeps the fail-open contract — a down daemon
must never block bare `tm` — but stops it from being invisible. Under
`RUST_LOG=debug` there is now a difference between "guard ran, found nothing"
and "guard never ran".

`plan_inplace`'s stopped-only gate is deliberately **not** widened, per the
issue.

## Tests

- `guard_managed_sessions_returns_none_when_the_list_call_errors` — 5xx from a
  real socket must fail open to `None`, not error out.
- `guard_managed_sessions_returns_none_when_the_daemon_is_unreachable` — the
  other cold-daemon shape: nothing listening.
- `guard_managed_sessions_returns_sessions_on_success` — proves the two above
  are failure paths, not the function's only behaviour.
- `checked_summaries_slim_skips_stale_assets_probe` — builds the fixture that
  genuinely reports `stale_assets: true` under the full probe, then shows the
  slim path reports `false` for it. Asserting the flag alone would be vacuous:
  `false` is also what a session with current assets gets.

## Quality gate

```
cargo fmt --check exit: 0

cargo clippy -p trusty-mpm --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 29.13s

cargo test -p trusty-mpm     (full suite, not --lib)   exit: 0
test result: ok. 3945 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 43.20s
test result: ok. 1287 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.81s
test result: ok. 1287 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.67s
(44 `test result: ok` lines total; zero FAILED, zero panics)

scripts/check_line_cap.sh: line-cap: 7 allowlisted, 0 violations — OK.
```

Stacked note: #4340 (#4336) touches `guided_inplace.rs` / `claude_code.rs`;
this touches `guided.rs` and the daemon summary path. No file overlap.

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -220,6 +220,25 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Bare `tm` no longer needs two runs after Claude Code backgrounds
+  ([#4335](https://github.com/bobmatnyc/trusty-tools/issues/4335)).
+  `nested_session_guard` fetched the managed-session list with a 2-second
+  timeout and a bare `.ok()?`. On a cold daemon the per-session `stale_assets`
+  pass over a large fleet exceeded that bound, the request was cancelled, and
+  the guard no-opped with nothing in any log to say it had not run — so bare
+  `tm` after an `/exit` (which BACKGROUNDS, never firing `SessionEnd`, leaving
+  the record `active` so the stopped-only gate correctly skips) fell through to
+  the picker and dead-ended on the #2678 switch-client refusal. A second run,
+  against a now-warm daemon, worked. Three changes, smallest lever first:
+  the managed-list endpoint takes `?slim=true` (folds into
+  [#4322](https://github.com/bobmatnyc/trusty-tools/issues/4322)) to skip the
+  `stale_assets` probe for callers that never read the flag — the guard reads
+  name/pane_id/state/id and was paying for a filesystem-bound catalog compose
+  it discards; the guard's timeout goes 2s → 8s, still bounding a hung daemon
+  while clearing a cold-start listing; and the error arm is logged at debug via
+  the new `guard_managed_sessions`. Absent or malformed `slim` keeps the full
+  probe, so no existing client changes shape. The stopped-only gate in
+  `plan_inplace` is deliberately unchanged.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/src/bin/tm/commands/guided.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided.rs
@@ -432,11 +432,18 @@ pub(crate) fn derive_project(
 /// What: short-circuits to `None` when not inside tmux (nothing to guard
 /// against) or when the daemon is unreachable/errors (fail-open — matches
 /// every other daemon round-trip in this file, which falls through to the
-/// picker/autostart rather than blocking on a down daemon). Uses a 2-second
-/// timeout, matching the `guided_inplace` probe convention.
+/// picker/autostart rather than blocking on a down daemon). The fetch goes
+/// through [`guard_managed_sessions`], which (#4335) asks for the SLIM listing
+/// — the guard never reads `stale_assets` — allows 8 seconds rather than 2,
+/// and LOGS the failure arm instead of discarding it. Before that, a cold
+/// daemon's per-session staleness pass regularly blew the 2s bound and the
+/// guard silently no-opped, so bare `tm` after a backgrounded Claude Code fell
+/// to the picker and dead-ended on the #2678 refusal until a second run.
 /// Test: the pure decision is `nested_managed_match_*` in
-/// `tests_behavior_c_tests.rs`; this I/O wrapper is exercised by manual smoke
-/// tests and the e2e suite (requires a live daemon + tmux).
+/// `tests_behavior_c_tests.rs`; the fail-open arm is
+/// `guard_managed_sessions_returns_none_when_the_list_call_errors`; the rest of
+/// this I/O wrapper is exercised by manual smoke tests and the e2e suite
+/// (requires a live daemon + tmux).
 async fn nested_session_guard(
     client: &reqwest::Client,
     url: &str,
@@ -452,7 +459,8 @@ async fn nested_session_guard(
     // check) so it can also break a duplicate-name tie in `nested_managed_match`
     // itself — see that function's doc.
     let current_pane_id = super::tmux_attach::current_tmux_pane_id();
-    let all = list_all_managed_sessions(client, url).await.ok()?;
+    // #4335: fail-open, but LOUDLY — see `guard_managed_sessions`.
+    let all = guard_managed_sessions(client, url).await?;
     nested_managed_match(
         current_session_name.as_deref(),
         env_session_id.as_deref(),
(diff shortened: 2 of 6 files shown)


bobmatnyc:
## Verdict: APPROVE

Zero CRITICAL, zero HIGH findings at >80% confidence. One MEDIUM (test-coverage gap), plus a merge-collision note that must be handled at land time regardless of this PR's own correctness.

## Findings

| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/mod.rs` | ~606 (`let slim = q.get("slim")...`) | `?slim=true` wiring is verified only at the `checked_summaries_with(records, bool)` unit-test level (`checked_summaries_slim_skips_stale_assets_probe`). Nothing exercises `list_managed_sessions` itself with a `Query` map containing `slim`/`slim=1`/`slim=true`/absent, so the query-string parsing (`v == "true" \|\| v == "1"`), the `!slim` inversion into `numbered_summaries`, and the end-to-end route → `checked_summaries_with` plumbing are unverified by any test that would catch a future refactor breaking the wiring (e.g. someone flips `!slim` to `slim` by mistake and every test still passes). | Add a handler-level test alongside the existing `mod.rs`/`tests.rs` list-handler tests that calls `list_managed_sessions` with `axum::extract::Query` containing `slim=true` (and separately `slim=1`, and absent) against a fixture with a genuinely-stale session, asserting `stale_assets: false` only in the slim case. |

## Attack-surface verification (no findings, checked explicitly)

- **Slim mode correctness**: `checked_summaries_with(records, probe_assets)` only gates the second (`stale_assets`) pass; the `unresumable` pass and the base `record_to_summary` mapping (which sets `id`, `name`, `state`, `pane_id`, `created_at`) run unconditionally. `nested_managed_match`/`pick_least_ambiguous` (`guided_resolver.rs`) read exactly `name`, `id`, `pane_id`, `state`, `created_at` — never `stale_assets` — so slim mode cannot starve the guard of a field it needs. Confirmed >80%.
- **No accidental slim callers**: `numbered_summaries` has exactly one call site (`managed_routes/mod.rs:599`, the list handler), and `list_all_managed_sessions` (which appends `?slim=true`) has exactly one call site (`guard_managed_sessions`, itself only called from `nested_session_guard`). No other consumer of the managed-list endpoint was switched to slim.
- **Timeout**: 2s → 8s on this one call, bounded (not unbounded), and the fail-open arm now logs via `tracing::debug!` in the new `guard_managed_sessions`, giving "guard ran, found nothing" a distinct signature from "guard never ran." Note (not scored — pre-existing behavior, not introduced by this PR): Gate A (`try_inplace_relaunch`/`fetch_managed_session_until_stopped`, `guided_inplace.rs`) runs before this guard and has its own `PROBE_TIMEOUT`/retry budget, so a genuinely *hung* (not merely cold) daemon can stack Gate A's existing wait on top of this guard's new 8s. Worth a look if #4335-class latency resurfaces, but out of scope for this PR's fix.
- **New tests**: `guard_managed_sessions_returns_none_when_the_list_call_errors` (5xx) and `guard_managed_sessions_returns_none_when_the_daemon_is_unreachable` (nothing listening) both exercise the fail-open arm this issue is about, plus `guard_managed_sessions_returns_sessions_on_success` proves the failure tests aren't vacuous. No test drives an actual timeout (slow response past 8s) as opposed to immediate error/refusal — a reasonable trade-off given the cost of a real 8s-plus test, not flagged as a gap.
- **`gh pr view --json closingIssuesReferences`**: exactly one entry, `#4335`. Correct.

## #4326 conflict check — THEY COLLIDE

PR #4326 (`fix/tm-ls-cold-latency`, open) and this PR both rewrite the same region of `crates/trusty-mpm/src/daemon/managed_routes/summary.rs` — not an adjacent-hunk near-miss, a genuine overlapping edit of the same functions:

- **`checked_summaries`**: #4343 turns it into a 1-line wrapper (`checked_summaries_with(records, true).await`) and moves the real body into a new `checked_summaries_with(records, probe_assets: bool)`, ANDing the existing `probe_staleness_for(&r.state)` filter with the new `probe_assets` flag. #4326 edits the *original* `checked_summaries` body in place — renames `probe_staleness_for` → `staleness_meaningful_for`, adds a second predicate `probe_staleness_in_list` (drops `Stopped` from the list-time probe), and adds `stale_assets_unchecked` population in the exact same filter/block #4343 touches. Same lines, incompatible restructurings — this will not merge cleanly either direction.
- **`stale_assets_for_many`**: #4343 leaves it synchronous, called via `tokio::task::spawn_blocking(move || stale_assets_for_many(syncable))`. #4326 makes it `async` and internally parallelized (new `staleness_inputs` helper + `JoinSet`), and changes every call site to `.await` it directly instead of wrapping in `spawn_blocking`. Whichever PR lands second will not compile against the other's version of this function without manual rework.
- **`numbered_summaries`**: #4343 adds a `probe_assets: bool` fourth parameter (threaded from the new `?slim=` query flag in `mod.rs`); #4326 does not touch this signature at all. Landing both requires the second PR to also thread `probe_assets` through whatever `checked_summaries`/`checked_summaries_with` shape wins, AND decide the composed skip condition (something like `probe_assets && probe_staleness_in_list(state)`), AND decide whether a slim caller should also get `stale_assets_unchecked: true` (today #4343 doesn't set that field at all, since it doesn't exist without #4326 — leaving `stale_assets: false` from a slim caller indistinguishable from #4326's "genuinely fresh" default, exactly the ambiguity #4326 introduced `stale_assets_unchecked` to close).

Net: not a stylistic near-miss, a real semantic merge job. Whoever lands second needs a deliberate rebase pass over `summary.rs`, not an autogenerated conflict resolution. Flagging for the PM to sequence deliberately rather than merge-waving both.

Issue #4091, opened 2026-07-27
bug(trusty-mpm): worktree prune force-deletes uncommitted work — no dirty-tree check anywhere in the reclaim path


bobmatnyc:
## Summary

Auditing the pause-prune path while implementing #2867 (PR #4090) turned up a
gap that is real but narrower than feared: the worktree GC has **no
uncommitted-work check anywhere**. Its entire safety model is "who owned it",
with zero "is there unsaved work here" component — and the removal itself is
`git worktree remove --force` with an `fs::remove_dir_all` fallback.

## What was checked (and is FINE)

`session_context_pause` prunes by default — `mcp/session_dispatch.rs:181-184`:

```rust
let prune_worktrees = args
    .get("prune_worktrees")
    .and_then(Value::as_bool)
    .unwrap_or(true);
```

…and it is a real, non-dry-run sweep (`daemon/mcp_context.rs:181-184`, trailing
`false` = `dry_run`).

`find_orphaned_worktrees` (`session_manager/prune.rs:425-485`) **does** cover the
#2867 path shape. It walks five shapes under `<repos_root>/<owner>/<repo>/`,
including `<repo>/.claude/worktrees/<name>` (prune.rs:456-460) and
`<repo>/.base/.claude/worktrees/<name>` (prune.rs:464-468).

And an ad-hoc, agent-created worktree with **no ownership sentinel** IS
protected — `prune.rs:889-897` routes `SentinelOwner::Unknown` into
`owner_unknown`, which is reported but never fed to the deletion loop:

```rust
SentinelOwner::Unknown => {
    info!(path = %candidate.display(),
        "prune-worktrees: owner-unknown sentinel — never auto-deleting; …(#3649)");
    owner_unknown.push(candidate);
}
```

So the specific #2867 scenario (ad-hoc worktree, no sentinel, dirty tree) is
**not** auto-deleted at pause. Good.

## The actual gap

1. **No dirty-tree check exists anywhere in the prune path.** Grepping
   `session_manager/` for `porcelain|dirty|uncommitted|is_clean|status --`
   yields only `git worktree list --porcelain`. Nothing inspects `git status`,
   the index, or untracked files before deletion. A worktree with a **known**
   owner that passes the owner-terminal + grace-window gates is force-deleted
   with uncommitted work in it — `decommission.rs:175-180` uses
   `git worktree remove --force`, falling back to `fs::remove_dir_all`
   (`decommission.rs:194`).

[...]

PR #4344, opened 2026-07-29
fix(trusty-mpm): daemon boot no longer resurrects deleted sessions; decommission no longer force-deletes dirty worktrees

bobmatnyc:
## Summary

Two related daemon-safety fixes.

- **Fix A — reconcile resurrects soft-deleted sessions.**
  `crates/trusty-mpm/src/session_manager/reconcile.rs:75` hand-rolled
  `matches!(record.state, ManagedSessionState::Decommissioned)` instead of
  asking the record's own `is_terminal()`. Every soft-deleted (`Deleted`)
  record fell through to the "gone" branch and was marked `Stopped` —
  resurrected — on every daemon boot. Switched to `is_terminal()`, which
  covers both `Decommissioned` and `Deleted`.

- **Fix B — decommission force-deletes dirty worktrees with no check.**
  `crates/trusty-mpm/src/session_manager/decommission.rs` (the
  `is_session_worktree` branch of `decommission_with_root`) ran `git worktree
  remove --force` (falling back to `fs::remove_dir_all`) unconditionally.
  `tm sessions prune --state stopped` calls `decommission` per matching
  record, so a stopped session with uncommitted/untracked work in its
  in-project worktree was silently destroyed. Reused the existing
  `worktree_safety::inspect_dirt` check (the same one `prune_orphaned_worktrees`
  already uses for the orphan-worktree sweep, from the #4091 dirty-worktree
  guard) to gate this second, previously-uncovered removal path: a dirty
  candidate is now refused and reported (the record still tombstones to
  `Decommissioned`, consistent with every other "skip removal" branch in this
  function).

  The existing happy-path test
  (`manager_decommission_removes_real_git_worktree`) built its base clone with
  no remote at all, which `inspect_dirt`'s unpushed-commit check treats as
  fully unpushed (nothing for `--not --remotes` to exclude) — so it needed a
  bare-remote push added to stay genuinely clean under the new gate.

## Test plan

- [x] `cargo test -p trusty-mpm` (full suite, not `--lib`-only) — 3946 + 1284×2
      + integration-binary tests, 0 failed
- [x] New regression tests:
  - `session_manager::tests::manager_reconcile_skips_deleted` — a `Deleted`
    record survives `reconcile_on_boot` without being resurrected
  - `session_manager::decommission_worktree_tests::manager_decommission_refuses_dirty_worktree` —
    a real git worktree holding an untracked file survives `decommission`,
    `workspace_removed` is `false`, and the record still tombstones

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -220,6 +220,23 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Daemon boot reconciliation no longer resurrects soft-deleted sessions.
+  `reconcile_on_boot` hand-rolled a
+  `matches!(record.state, ManagedSessionState::Decommissioned)` check instead
+  of asking the record's own `is_terminal()`, so every `Deleted` record (no
+  live tmux session, by definition) fell through to the "gone" branch and was
+  marked `Stopped` — resurrected — on every daemon restart. `is_terminal()` is
+  now the single source of truth, covering both `Decommissioned` and
+  `Deleted`.
+- `decommission` (and therefore `tm sessions prune --state stopped`, which
+  calls it per matching record) no longer force-deletes a dirty in-project
+  worktree. Removing an SM-created `.worktrees/<id>` ran `git worktree remove
+  --force` (falling back to `fs::remove_dir_all`) with no dirty check at all —
+  a live data-loss path for uncommitted/untracked work. This reuses the same
+  `worktree_safety::inspect_dirt` check the orphan-worktree sweep uses (the
+  #4091 dirty-worktree guard): a dirty candidate is now refused and reported,
+  and the session record still tombstones to `Decommissioned` so the skip is
+  never silent.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -236,7 +236,15 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `worktree_safety::inspect_dirt` check the orphan-worktree sweep uses (the
   #4091 dirty-worktree guard): a dirty candidate is now refused and reported,
   and the session record still tombstones to `Decommissioned` so the skip is
-  never silent.
+  never silent. Two follow-ups close the gap between that refusal and where
+  an operator would actually see it: (1) the tombstone no longer nulls
+  `workspace_path` when nothing was removed — every "skip removal" branch in
+  `decommission` leaves real content in place, and blanking the pointer would
+  strand it with nothing but a transient log line as a trail back to it; (2)
+  `tm sessions prune --state stopped` (and the underlying `PruneAction`) now
+  has a distinct `decommissioned_worktree_retained` outcome, with the
+  retained path echoed back, instead of printing the same `decommissioned`
+  line whether or not the worktree was actually deleted.
 - The per-project worktree opt-out is no longer silently conditional on the
   daemon being up ([#4300](https://github.com/bobmatnyc/trusty-tools/issues/4300)).
   `Project.worktree: false` ([#3455](https://github.com/bobmatnyc/trusty-tools/issues/3455),
--- a/crates/trusty-mpm/src/bin/tm/commands/managed.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/managed.rs
@@ -692,6 +692,19 @@ pub(crate) async fn session_prune(
             .and_then(serde_json::Value::as_str)
             .unwrap_or("?");
         println!("{action} {id} {name} [{prior}]");
+        // #4344 review: `decommissioned_worktree_retained` means the record
+        // was tombstoned but its in-project worktree held unsaved work and
+        // was deliberately NOT deleted (see `worktree_safety::inspect_dirt`).
+        // Without a distinct, visible line here, this printed identically to
+        // an ordinary `decommissioned` row and the refusal was invisible at
+        // the one surface an operator actually reads.
+        if action == "decommissioned_worktree_retained" {
+            let retained_path = s
+                .get("retained_workspace_path")
+                .and_then(serde_json::Value::as_str)
+                .unwrap_or("?");
+            println!("  ! worktree retained (dirty, not deleted): {retained_path}");
+        }
     }
     let verb = if dry_run { "would prune" } else { "pruned" };
     println!("{verb} {} session(s) (filter={state})", sessions.len());
--- a/crates/trusty-mpm/src/session_manager/decommission.rs
+++ b/crates/trusty-mpm/src/session_manager/decommission.rs
@@ -22,6 +22,7 @@ use super::manager::{ManagedError, SessionManager};
 use super::record::{ManagedSessionId, ManagedSessionState, SessionRecord};
 use super::search_gc;
 use super::workspace_guard::is_safe_to_remove;
+use super::worktree_safety::{DirtyWorktree, inspect_dirt};
 
 /// Sentinel file written by [`create_session_worktree`] into every SM-created
 /// per-session git worktree (#1845 item 5).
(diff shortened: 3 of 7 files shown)


bobmatnyc:
## Verdict: WARN

CI green (24/24 checks pass, one N/A skip on "Notify on red main"), `closingIssuesReferences` is empty as expected. Both stated fixes are correctly implemented at the point of the bug. One HIGH finding: the dirty-worktree refusal (Fix B) is invisible at the primary UX surface it was built to protect.

## Findings

| HIGH | `crates/trusty-mpm/src/session_manager/prune.rs` | 568 (`else if let Err(e) = self.decommission(&record.id, caller).await`) | `decommission()` returns `Result<(SessionRecord, bool), ManagedError>` — the `bool` is `workspace_removed`. `prune_managed` only matches `Err`; the `Ok((_, workspace_removed))` payload, including the flag this very PR introduces, is discarded. Combined with `decommission.rs:543` (`record.workspace_path = None;` runs unconditionally, even when the dirty gate refused removal) and `PruneAction` (`prune.rs:245-249`, whose `Decommissioned` doc comment literally says "workspace removed") having no variant/field for "refused, worktree retained," `tm sessions prune --state stopped` prints the identical `decommissioned <id> <name> [stopped]` line whether the worktree was actually deleted or silently kept on disk with unsaved work. The only trace of the refusal is a `warn!` (`decommission.rs:440-451`) in the daemon's own tracing log — never returned over the HTTP prune route, never seen by the CLI (`managed.rs:634-697`) or MCP caller. Once tombstoned, the record's `workspace_path` is gone too, so there is no durable, queryable pointer back to the retained directory — an operator who trusts the prune output has no way to discover the worktree still exists short of independently running `tm session prune --worktrees` (a different command, different code path) or grepping historical logs. This is exactly the "half-decommissioned, silently skipped" failure mode the PR's own description worries about for Fix B, reintroduced one layer up. | Thread the flag through: add `workspace_removed: bool` (or a `PruneAction::DecommissionedWorktreeRetained` variant) to `PrunedSession`, capture `decommission()`'s `Ok((_, removed))` in `prune_managed` instead of discarding it, and print the distinction in `session_prune`'s CLI loop (`managed.rs`). Separately, stop unconditionally clearing `record.workspace_path` at `decommission.rs:543` when `workspace_removed` is `false` — keep it pointing at the retained directory so the record itself remains the durable trail back to it. |

## Notes

- `is_terminal()` (record.rs:138-140) covers exactly `Decommissioned | Deleted` — the same two variants the old hand-rolled check was *supposed* to cover before it missed `Deleted`. `Provisioning`, `Active`, `Stopped`, `Errored` are unaffected and still get reconciled normally; no legitimate resurrectable state is newly skipped. Correct fix, well-targeted regression test (`manager_reconcile_skips_deleted`).
- Fail-toward-dirty on panic (`decommission.rs:429-439`) goes through `spawn_blocking(...).await.unwrap_or_else(|e| Some(DirtyWorktree::new(...)))` — this is tokio's own panic→`JoinError` conversion (the workspace doesn't set `panic = "abort"`), not a raw `catch_unwind`. No shared state crosses the panic boundary inside the blocking closure, so nothing is poisoned; a panicking `inspect_dirt` correctly degrades to "treat as dirty" rather than becoming a false green light.
- The "no remote ⇒ every commit counts unpushed ⇒ always dirty" behavior (`worktree_safety.rs`, `count_unpushed`) is **pre-existing** — that file isn't touched by this diff, it's reused verbatim from #4091's orphan-sweep guard. It's also explicitly documented as the intended, conservative fail-safe direction of that module. Real risk is narrow: in-project session worktrees are created off the project's own base clone, which in the overwhelming majority of daemon-managed projects already has an `origin` remote. A genuinely from-scratch, never-pushed local repo would see every one of its stopped sessions refuse worktree removal indefinitely (no force override exists on `decommission`'s gate, unlike `prune_orphaned_worktrees`'s `DirtyWorktreePolicy::ForceDiscard`) — worth knowing operationally, but not a defect introduced by this PR.
- The `manager_decommission_removes_real_git_worktree` test's added `push_to_bare_remote` (`decommission_worktree_tests.rs`) correctly neutralizes the pre-existing "no remote = dirty" quirk so the happy-path test's only variable is the intended clean-removal path; reasoning checks out against `count_unpushed`'s actual semantics (upstream-less branch created at the same tip as the pushed ref nets 0 unpushed commits).

bobmatnyc:
Addressed the HIGH from the critic review (both points):

1. `decommission` no longer nulls `workspace_path`/`workspace_owned` when nothing was removed — only when the content is actually gone from disk (removed, or already absent). Every "skip removal" branch (dirty-worktree refusal, unowned/local-path, containment-guard) now leaves a durable pointer back to retained content instead of just a transient `warn!` line.
2. `prune_managed` no longer discards the `bool` half of `decommission`'s `(SessionRecord, bool)` return. `PruneAction` gained `DecommissionedWorktreeRetained`, `PrunedSession` carries `retained_workspace_path`, and the `tm sessions prune` CLI now prints a distinct `! worktree retained (dirty, not deleted): <path>` line instead of an identical `decommissioned` row.

New regression tests: `manager_decommission_refuses_dirty_worktree` (updated to assert `workspace_path` survives) and `prune_reports_dirty_worktree_retained` (asserts the retained state + path round-trip through `prune_managed`).

Gate, raw output:
- `cargo fmt --check -p trusty-mpm`: clean
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: exit 0, only pre-existing unrelated warnings
- `cargo test -p trusty-mpm` (full suite, not `--lib`): `3947 passed; 0 failed` (lib) + all integration binaries green
- `bash scripts/check_line_cap.sh`: `7 allowlisted, 0 violations — OK`
- `bash scripts/check_test_pointers.sh`: `0 dangling pointers — OK`

Pushed to the same branch: `b020751f`

bobmatnyc:
## Verdict: APPROVE

b020751f correctly fixes the HIGH from the prior round. Re-verified against the new diff, not the author's description of it.

## What changed (verified against the diff, not the summary)

- `decommission.rs:551-556` — the tombstone step now checks `workspace_still_on_disk = record.workspace_path.as_deref().is_some_and(|p| p.exists())` and only nulls `workspace_path`/`workspace_owned` when the path is genuinely gone. Confirmed this covers **every** skip-removal branch, not just the dirty refusal: (1) dirty-worktree refusal — path retained; (2) owned-but-already-absent (`!ws.exists()`) — correctly still nulled, nothing to retain; (3) owned workspace that fails the `is_safe_to_remove` containment guard — now also retained (this was silently nulled before, a second latent instance of the same bug class the original HIGH described); (4) unowned/local-path/adopted (never SM-owned) — now retained since the directory was never touched.
- `prune.rs` — `prune_managed` now matches on `Ok((tombstone, workspace_removed))` instead of discarding the tuple; new `PruneAction::DecommissionedWorktreeRetained` (`prune.rs:264-267`, serializes to `"decommissioned_worktree_retained"` via the existing `#[serde(rename_all = "snake_case")]` derive — hand-checked against `as_str()`, they match) and `PrunedSession::retained_workspace_path: Option<PathBuf>` (`#[serde(skip_serializing_if = "Option::is_none")]`) carry the state through.
- `managed.rs:695-706` — CLI prints `! worktree retained (dirty, not deleted): <path>` as an extra line when the action is the new variant; the base `println!("{action} {id} {name} [{prior}]")` line is unconditional and generic (extracted from a `serde_json::Value`, not a Rust match), so an older client reading this JSON just doesn't print the extra line — no break.

## Checks requested

- **Every skip-removal branch retains the path**: confirmed by code reading (see above) — this is a strict generalization, not a point patch scoped to only the dirty case.
- **No breaking change for existing consumers**: checked for exhaustive `match` sites over `PruneAction` outside `prune.rs` itself — none exist (the only two `match self {` sites in that file are `PruneFilter`'s `Display` and `PruneAction::as_str()`, both updated). The MCP `session_prune` tool (`mcp/mod.rs:294`) and the HTTP route return the serialized `PruneOutcome` `Value` directly with no exhaustive matching, so both are forward-compatible with the new variant/field. Also checked the *other*, separate `tm session decommission <id>` singular-verb path (`managed.rs` `decommission_managed_session` handler + CLI, doc-referenced as `decommission_message_reflects_workspace_removed`) — that path already threads its own `workspace_removed` honestly and predates this PR; it was never part of the gap being fixed and isn't touched here.
- **Round-trip test**: `prune_reports_dirty_worktree_retained` (`decommission_worktree_tests.rs`) calls the real `mgr.prune_managed(PruneFilter::Stopped, false, false, None)` — the actual shared function the HTTP route and MCP tool both call — against a real dirtied git worktree, and asserts `action == DecommissionedWorktreeRetained` and `retained_workspace_path == Some(worktree_path)`. This exercises the real business-logic surface. It does not go through actual JSON (de)serialization or the CLI's `println!` formatting — see Notes.
- **CI on the new head**: **did not run.** `gh api commits/b020751f.../check-runs` returns zero check runs, and a direct scan of the last 50 Actions runs repo-wide finds none for this SHA, ~17 minutes after push. Separately, `gh pr view 4344` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` — the branch now conflicts with `main` and needs a rebase. Since GitHub's own CI signal was unavailable, I checked out `b020751f` into a disposable worktree and ran the real gates myself: `cargo check -p trusty-mpm --all-targets` — clean; `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — clean (only pre-existing, unrelated warnings: MSRV clippy.toml/Cargo.toml mismatch, duplicate bin-target names); `cargo test -p trusty-mpm --lib` targeted at the five relevant tests (`manager_decommission_refuses_dirty_worktree`, `prune_reports_dirty_worktree_retained`, `manager_reconcile_skips_deleted`, `manager_decommission_removes_real_git_worktree`, `prune_outcome_serializes`) — all 5 pass. This substitutes for, but does not replace, an actual green CI run — the PR cannot be merged in its current conflicting state regardless.

## Notes

- MEDIUM-severity, non-blocking: `prune_outcome_serializes` (`tests.rs:1913-1942`) exists specifically to pin the prune wire contract but was not extended to cover the new variant — no test asserts the actual `serde_json::to_value` output is `"decommissioned_worktree_retained"` or that `retained_workspace_path` is present/absent per `skip_serializing_if`. Low risk (checked by hand, matches), but it's exactly the regression class that test was written to catch.
- Action required before merge, unrelated to code quality: resolve the merge conflict with `main` and get an actual green CI run before this lands — I cannot confirm actions-based CI passes on this exact commit since GitHub never ran it.

Issue #4246, opened 2026-07-28
tctl status reports all healthy daemons as `down`; false `down` also triggers launchctl kickstart of healthy daemons


mac-duetto:
## Symptom

`tctl status` reports every daemon as `down` while all of them are running and healthy.

```
tctl status — stack summary
  trusty-search      0.39.0       down
  trusty-memory      0.21.0       down
  trusty-analyze     0.7.4        down
  trusty-review      0.10.1       down
  tga                2.9.4        n/a
  trusty-console     0.4.0        down
  trusty-mpm         0.19.21      unknown
```

Ground truth on the same machine — every daemon answers `GET /health` with `{"status":"ok",...}`:

| Component | tctl | Actual | Port |
|---|---|---|---|
| trusty-search | down | UP | 7878 |
| trusty-memory | down | UP (launchd `com.trusty.memory`) | 7070 |
| trusty-analyze | down | UP (running 0.7.2, installed 0.7.4) | 7879 |
| trusty-review | down | UP | 7891 |
| trusty-console | down | UP | 7788 |
| trusty-mpm | unknown | UP | 7880 |
| tga | n/a | not running (correct — `daemon: false`) | — |

## Root cause

`probe_member_health` (`crates/trusty-installer/src/commands/probe.rs:256-274`) probes health by shelling out to `<binary> health --json` and parsing `{"status": ...}`. No daemon implements that contract:

```
$ trusty-memory health --json   → error: unrecognized subcommand 'health'
$ trusty-review health --json   → error: unrecognized subcommand 'health'
$ trusty-console health --json  → error: unrecognized subcommand 'health'
$ trusty-analyze health --json  → error: unexpected argument '--json' found
$ trusty-search health --json   → {"daemon":"running","indexes":[...]}   # exit 0, no "status" key
```

Every distinct failure — clap exit 2, unsupported flag, schema mismatch — collapses into a single `Down`:
- `probe.rs:266-268` — non-zero exit → `MemberHealth::Down`, stderr and exit code discarded
- `probe.rs:216` — `.get("status")...unwrap_or("down")` — schema mismatch silently reads as down

Versions resolve correctly because that is a separate path (`update_engine.rs:147-153`) that only runs `<binary> --version`.

[...]

Issue #3364, opened 2026-07-20
trusty-code default HTTP port 7881 collides with tm supervisor metrics port; no tcode daemon in installed service set


bobmatnyc:
**Symptom:** Foundry (trusty-code-gui) shows "DAEMON UNREACHABLE — LOAD FAILED" on a fresh signed build (b40c22ff), reported by Bob 2026-07-19.

**Root cause (diagnosed from logs + source):**
1. `trusty_code::serve::DEFAULT_HTTP_PORT = 7881` (crates/trusty-code/src/serve/mod.rs:72) collides with `trusty-mpm`'s `DEFAULT_METRICS_ADDR = "127.0.0.1:7881"` (crates/trusty-mpm/src/supervisor/config.rs:83-86, chosen to "sit next to the daemon's 7880") — two independently-picked defaults on the same port.
2. No tcode daemon is installed on the machine at all — the launchd fleet (console/memory/search/mpm/analyze) does not include tcode, and the GUI is a pure thin HTTP client (DOC-39 §2.1): no sidecar spawn code exists, `get_daemon_url` just echoes `TRUSTY_CODE_URL` or the hardcoded `http://127.0.0.1:7881` (crates/trusty-code-gui/src/state.rs).
3. `tm supervisor` answers `/health` with a generic `{"status":"ok"}`, so both the GUI's neighborhood and ops probes get a false healthy signal; real calls 404 (`/api/sessions` → 404; WebKit logged `didFailResourceLoad` every ~6s poll).

Pre-existing latent defect — not introduced by b40c22ff; earlier legs likely had a manually-started tcode process that has since died.

**Workaround (live now, session-local):** `tcode serve --http --port 7882` started manually; Foundry relaunched with `TRUSTY_CODE_URL=http://127.0.0.1:7882`. Does not survive reboot/shell exit.

**Fix scope:**
- Move `trusty-code`'s default HTTP port off 7881 to an unclaimed port (survey existing defaults across the workspace first; supervisor's 7880/7881 pair is deployed via launchd, so tcode moves). Change the GUI's hardcoded default in `state.rs` **in lockstep**, plus docs/DOC-39 references and tests.
- Consider a workspace port-assignment table in docs to prevent the next collision.
- Separate follow-up decision: whether tcode joins the installed launchd daemon set (trusty-installer), or the GUI grows supported daemon-lifecycle UX. Not required to close this issue.

[...]

PR #4381, opened 2026-07-29
fix(installer): probe daemon health over HTTP instead of a nonexistent CLI verb (#4246)

mac-duetto:
Fixes the reported symptom of #4246 and the active harm behind it.

## Problem

`tctl status` reported every daemon as `down` while all were healthy. `probe_member_health` probed by shelling out to `<binary> health --json` — a contract **no daemon implements** — and collapsed every distinct failure (clap exit 2, unsupported flag, schema mismatch, timeout) into a single `Down`.

The false `down` then drove `needs_kickstart` → `launchctl kickstart -k`, so **every `tctl install` hard-restarted healthy daemons.** With no `ExitTimeOut` in the plist renderer and a ≥30s-per-index flush budget, that restart is a SIGKILL mid-flush — see the analysis comments on #4246 for the index-degradation detail. That is why this was P1 rather than cosmetic.

A secondary consequence: `tctl install` could never pass its own verify gate, since trusty-search/memory/review are `required: true` and all read `down`. The exit code CI would gate on was permanently failing.

## Change

HTTP `/health` probe, local to `trusty-installer`. No daemon crate is touched.

- **Typed `ProbeOutcome`** replaces the `String` return — `NotInstalled`, `Refused`, `Timeout`, `HttpError`, `BadEnvelope`, `Serving`, `Unprobeable`. A `health_string()` shim keeps the 6 existing call sites compiling and `--json` keys unchanged.
- **`.no_proxy()` on the probe client.** reqwest 0.12.28 routes `127.0.0.1` through `HTTP_PROXY`/`ALL_PROXY` (hyper-util's matcher has no loopback exemption), so without this the new probe reproduces the identical false-`down`. This is the first `.no_proxy()` in the workspace.
- **Dual resolution with an explicit precedence rule.** Both `{data_dir}/http_addr` and the documented fixed port are probed; `reconcile()` ranks `Serving` strictly below `Refused`/`Timeout` and takes the minimum, so **an answering leg always suppresses a refusing leg.** This matters because trusty-memory port-walks `7070..=7079` — a daemon on 7071 makes the fixed-port leg refuse while being perfectly healthy. Without this rule the fix would restart healthy daemons.
- **Body-based classification that accepts non-2xx.** trusty-analyze returns **503 + `status:"degraded"`** when search is unreachable; that classifies `degraded`, not `down`. `daemon_state`/`embedder_status` are read so a warming daemon is not reported as fully serving.
- **Kickstart gated on `is_confirmed_down`** = `Refused | Timeout` only. A schema or envelope problem can never authorise a destructive `launchctl kickstart -k`.
- **Duplicate probe collapsed** — `SystemRunner::probe` deleted, `tctl up` routed through the shared probe.
- **`Kickstarter` seam** on `verify_one`, mirroring the existing `ServiceEnv` seam in the same file, so the no-kickstart property is testable at all.

## Result

Before:
```
tctl status — stack summary
  trusty-search      0.39.0       down
  trusty-memory      0.21.0       down
  trusty-analyze     0.7.4        down
  trusty-review      0.10.1       down
  tga                2.9.4        n/a
  trusty-console     0.4.0        down
  trusty-mpm         0.19.21      unknown
verdict: degraded (exit 2)
```
After:
```
tctl status — stack summary
  trusty-search      0.39.0       healthy
  trusty-memory      0.21.0       healthy
  trusty-analyze     0.7.4        healthy
  trusty-review      0.10.1       healthy
  tga                2.9.4        n/a
  trusty-console     0.4.0        healthy
  trusty-mpm         0.19.21      unknown
verdict: ready (exit 0)
```
Byte-identical with `HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY` exported at a dead address.

## Tests

483 → 506. Two pre-existing tests **asserted the bug was correct** and were fixed:
- `classify_status("anything-else") == Down` — `"running"` was `anything-else`
- `classify_health_json(br#"{"x":1}"#) == Down` — `{"x":1}` is the shape of trusty-search's real exit-0 payload

New coverage, using the existing in-repo stub-server vehicle promoted to a shared `test_support` module (no new dev-dependency; `[dev-dependencies]` is still empty):
- `verify_one_does_not_kickstart_a_healthy_launchd_daemon` — **the test that proves the P1 harm is fixed.** The pre-existing `needs_kickstart` test passed while the bug was live, because it fed the predicate a hand-typed `"down"`; the predicate was never wrong. Only running the real probe against a stubbed healthy daemon and then asserting the kickstart decision closes the loop.
- `verify_one_kickstarts_a_genuinely_down_launchd_daemon` — so this cannot regress into never repairing.
- `probe_ignores_http_proxy_env`, `probe_port_walked_daemon_is_healthy`, `probe_accepts_503_degraded`, `probe_rejects_non_trusty_envelope`, `probe_distinguishes_failure_causes`, `real_daemon_payloads_are_never_down` (over the six payloads captured live), `health_string_maps_every_variant`, `only_transport_failures_are_confirmed_down`, `needs_kickstart_never_fires_on_a_schema_problem`, `reconcile_serving_leg_beats_refused_leg`.

Mutation-checked: deleting `.no_proxy()` fails `probe_ignores_http_proxy_env`; widening `is_confirmed_down` to include `HttpError` fails 5 tests, and to include `BadEnvelope` as well fails 6.

Independently verified by a separate QA pass: status table cross-checked against live `curl /health` on all six ports, proxy immunity reproduced, all named tests confirmed present and passing individually, both mutations reproduced, and **daemon PID continuity confirmed — trusty-memory stayed PID 1795 and trusty-search PID 28070 throughout, so no kickstart fired during development.**

## Two reviewer call-outs

1. **`classify_status` now accepts `running`/`serving`** in addition to `ok`/`healthy`/`ready`. DOC-1 D4 mandates this vocabulary, but no shipped daemon emits `running` (all six emit `ok`), so this changes nothing observable today and the catch-all still maps to `Down`. My own analysis comment on #4246 called this "a spec-vs-reality question to settle in the ADR" — flagging it explicitly so a reviewer can object. One line to revert.
2. **`tctl up` behaviour changes slightly more than predicted.** Collapsing the duplicate probe was cleared as behaviour-neutral because `ensure_member` treats `HealthyStale` and `Down` identically — true, but the real change is `Down → HealthyVersionOk`: a healthy daemon is now a reported no-op instead of being handed a redundant `<binary> start`. Same false-`down` class one layer up, and desirable, but it is a behaviour change.

`BadEnvelope` maps to `down` rather than `unknown` deliberately: `unknown` is tolerated by both `VerifyTailReport::build` and `status`'s exit code, so mapping there would print VERIFIED / exit 0 for a stack we hold no health information about. `down` is honest, and now safe because the kickstart is gated on the outcome variant rather than the string.

## Deliberately out of scope

Tracked in #4246's analysis comments for follow-up:
- `service` field in health envelopes (squatter defence, #3364 class) — touches 6 daemon crates
- trusty-mpm's `OwnVerb` arm still returns `unknown`. Probing it would flip `tctl status` to exit 2 and `tctl install` to NOT VERIFIED for anyone not running mpm, since mpm is `required: true`
- Version-skew/`stale` detection, `-v` wiring
- `ExitTimeOut` in the plist renderer, the `indexed_head_sha` stamp-before-embed ordering in trusty-search, and the data-dir-blind orphan reaper — all independent of this PR and reachable from other entry points
- Workspace-wide `.no_proxy()` (133 client-builder sites; `tctl ensure` is broken today for proxy users)

## Note on CI

`cargo clippy --workspace --all-targets -- -D warnings` fails on `trusty-code-gui` and `trusty-mpm-gui` for a missing untracked `ui/dist`. Confirmed pre-existing and unrelated: the same failure reproduces on a clean `origin/main` checkout, and neither crate depends on `trusty-installer`. `cargo clippy -p trusty-installer --all-targets -- -D warnings` and `cargo fmt --check` are both clean.

### DIFF
--- a/crates/trusty-installer/src/commands/ensure/mod.rs
+++ b/crates/trusty-installer/src/commands/ensure/mod.rs
@@ -34,15 +34,17 @@ pub mod report;
 
 /// Process-wide lock serialising tests that mutate global env vars.
 ///
-/// Why: several ensure submodule tests set/remove `TRUSTY_DATA_DIR_OVERRIDE`,
-/// `TCTL_ENSURE_WAIT_*`, etc. Those are process-global, so tests in different
-/// modules would race each other. A single crate-shared lock — held for the
-/// duration of each env-mutating test across every ensure submodule — serialises
-/// them so the assertions are deterministic.
-/// What: a `Mutex<()>` accessible to every ensure submodule's test code.
+/// Why: this used to be defined here, because `ensure`'s submodules were the
+/// only env-mutating tests in the crate. #4246 gave `probe_http` and
+/// `verify_tail` the same need, so the lock (and the stub-server helpers that go
+/// with it) moved to `commands::test_support`. Re-exported under the original
+/// path so the existing `use crate::commands::ensure::ENV_TEST_LOCK` imports in
+/// `daemon`, `project_setup` and `readiness` keep working — there is still
+/// exactly ONE lock in the process, which is the whole point.
+/// What: a re-export of [`crate::commands::test_support::ENV_TEST_LOCK`].
 /// Test: used by the env-mutating tests in `daemon`, `project_setup`, `readiness`.
 #[cfg(test)]
-pub(crate) static ENV_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+pub(crate) use crate::commands::test_support::ENV_TEST_LOCK;
 
 use crate::commands::runtime::block_on;
 use mcp_patch::MCP_FILE;
--- a/crates/trusty-installer/src/commands/ensure/project_setup.rs
+++ b/crates/trusty-installer/src/commands/ensure/project_setup.rs
@@ -244,87 +244,11 @@ pub async fn run_stages(project_root: &std::path::Path) -> Vec<StageOutcome> {
 mod tests {
     use super::*;
     use crate::commands::ensure::ENV_TEST_LOCK as ENV_LOCK;
-    use trusty_common::DATA_DIR_OVERRIDE_ENV;
-
-    /// Spawn a one-shot TCP server that answers the first request with a fixed
-    /// HTTP response, and return its `host:port`.
-    ///
-    /// Why: standing up a real `reqwest` round-trip against a controlled
-    /// response lets the stage logic be tested end-to-end without a live daemon.
-    /// What: binds an ephemeral loopback port, accepts one connection, reads the
-    /// request, writes `response_body` wrapped in a 200 (or the raw status line
-    /// when `raw` is set), and returns the bound address.
-    async fn stub_once(status_line: &'static str, body: &'static str) -> String {
-        stub_seq(vec![(status_line, body)]).await
-    }
-
-    /// Spawn a TCP server that answers a *sequence* of requests, one fixed
-    /// response per accepted connection in order.
-    ///
-    /// Why: the palace-create stage issues two requests (an existence GET then a
-    /// create POST); covering the "created" branch needs a stub that 404s the
-    /// GET then 200s the POST.
-    /// What: binds an ephemeral loopback port, accepts `responses.len()`
-    /// connections, and writes each `(status_line, body)` in turn.
-    async fn stub_seq(responses: Vec<(&'static str, &'static str)>) -> String {
-        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
-        let addr = listener.local_addr().unwrap().to_string();
-        tokio::spawn(async move {
-            use tokio::io::{AsyncReadExt, AsyncWriteExt};
-            for (status_line, body) in responses {
-                if let Ok((mut sock, _)) = listener.accept().await {
-                    // Drain the request up to (and including) the end-of-headers
-                    // marker before replying. A single fixed-size read can split
-                    // a request whose `root_path` body is long, which used to
-                    // race the write and flake CI; reading until `\r\n\r\n` (or
-                    // EOF) consumes the whole header block deterministically. We
-                    // don't need the body — only that the request is fully sent.
-                    let mut acc = Vec::with_capacity(2048);
-                    let mut chunk = [0u8; 2048];
-                    loop {
-                        match sock.read(&mut chunk).await {
-                            Ok(0) => break,
-                            Ok(n) => {
-                                acc.extend_from_slice(&chunk[..n]);
-                                if acc.windows(4).any(|w| w == b"\r\n\r\n") {
-                                    break;
-                                }
-                            }
-                            Err(_) => break,
-                        }
-                    }
-                    let resp = format!(
-                        "{status_line}\r\nContent-Type: application/json\r\nContent-Length: {}\r\n\r\n{body}",
-                        body.len()
-                    );
-                    let _ = sock.write_all(resp.as_bytes()).await;
-                    let _ = sock.shutdown().await;
-                } else {
-                    break;
-                }
-            }
-        });
-        addr
-    }
-
-    fn stub_data_dir(app: &str, addr: &str) -> std::path::PathBuf {
-        let tmp = std::env::temp_dir().join(format!(
-            "tctl-ensure-setup-{}-{}-{}",
-            app,
-            std::process::id(),
-            std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH)
-                .map(|d| d.as_nanos())
-                .unwrap_or(0)
-        ));
-        std::fs::create_dir_all(&tmp).unwrap();
-        unsafe {
-            // SAFETY: serialised by ENV_TEST_LOCK; no concurrent env access in this crate's tests.
-            std::env::set_var(DATA_DIR_OVERRIDE_ENV, &tmp);
-        }
-        trusty_common::write_daemon_addr(app, addr).unwrap();
-        tmp
-    }
+    // #4246: the stub-server + stubbed-data-dir vehicle these tests grew is now
+    // shared with `probe_http`/`verify_tail` — one copy, in `test_support`.
+    use crate::commands::test_support::{
+        clear_data_dir_override, stub_data_dir, stub_empty_data_dir, stub_once, stub_seq,
+    };
 
     /// Why: a fresh registration (`created:true`) must report `changed = true`.
     /// What: stub returns `{"created":true}`; assert the outcome.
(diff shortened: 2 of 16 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review of the HTTP health-probe replacement, checked out and built locally at head `8ca5715f`. No prior critic verdict was posted on this PR (comments list was empty) — this is the first.

## Findings

| HIGH | crates/trusty-installer/src/commands/probe_http.rs | 488 | Doc comment on `resolve_probe_bases` cites `tests::resolve_probe_bases_none_for_unknown_binary`, which does not exist anywhere in the crate. The actual test covering that path is `probe_no_address_when_nothing_resolves` in `probe_http_tests.rs:548`. This is a live, currently-failing required check on this PR — "Test: doc-comment pointer lint" — confirmed via `gh run view` on run 30498991627: `FAIL: crates/trusty-installer/src/commands/probe_http.rs:487: Test: pointer cites 'resolve_probe_bases_none_for_unknown_binary' — no fn/mod found`. Not pre-existing on main; introduced by this PR. | Change line 488 to `` `tests::probe_no_address_when_nothing_resolves`. ``  |

## Verified clean (no findings)

- **Removed contract, no orphaned callers.** `<binary> health --json`, `classify_health_json`, and `output_with_timeout` are gone; grepped the whole workspace (not just this crate) for any remaining caller of `probe_member_health`/`needs_kickstart` outside `trusty-installer` — none. Remaining "health --json" hits are historical doc comments and pre-existing DOC-1 design docs, not live code.
- **Loopback-only doctrine.** `probe_http.rs` only ever targets `127.0.0.1:<fixed_port_for>` or whatever `read_daemon_addr` recorded (itself always loopback per the six daemons' own bind code). It's a client only — no `bind()` calls, nothing newly exposed.
- **Port table.** `fixed_port_for` (7070/7788/7878/7879/7880/7891) matches `docs/architecture/port-assignments.md` exactly, and the `app` name passed to `resolve_probe_bases` matches the `write_daemon_addr(app_name, …)` convention used by trusty-memory/console/search/analyze/mpm/review — confirmed by grep, not just the PR's own claim.
- **Bounded, non-hanging probe.** Per-leg `CONNECT_TIMEOUT` (2s) + `REQUEST_TIMEOUT` (5s), both legs run concurrently via `tokio::join!`. `verify_tail`'s pre-existing `AGGREGATE_POLL_BUDGET` (120s) and `bounded_attempts` (max 12) are untouched by this PR and still cap total wall-clock — no new unbounded retry loop.
- **Kickstart gate.** `needs_kickstart` now requires `ProbeOutcome::is_confirmed_down()` (`Refused`/`Timeout` only) — a schema mismatch, non-2xx-with-body, or squatter can no longer trigger `launchctl kickstart -k`. `verify_one_does_not_kickstart_a_healthy_launchd_daemon` and `verify_one_kickstarts_a_genuinely_down_launchd_daemon` both exercise the real probe against a stub, not a hand-typed string — closes the exact gap the PR describes (the old test defended the bug).
- **Proxy immunity.** `probe_ignores_http_proxy_env` asserts both a leaky client (no `.no_proxy()`) IS diverted and the production client is NOT, under an exported `HTTP_PROXY` pointed at a dead address. Ran it locally, passes.
- **Ran the full suite locally:** `cargo test -p trusty-installer --lib` → `506 passed; 0 failed; 4 ignored`. `cargo clippy -p trusty-installer --all-targets -- -D warnings` → clean.
- Install-time race: probing happens after the existing bootstrap/verify-tail sequencing (unchanged by this PR); the fix makes the probe strictly more conservative about restarting, so it cannot introduce a new race that wasn't already there.

## Required Changes

1. Fix the dangling doc-comment pointer at `probe_http.rs:488` (see HIGH finding). This is a real, currently-red required check — not a style nit.

## Notes

- **CI / merge readiness:** `mergeStateStatus: BEHIND` (base `114de333` vs current `main` `8976ed00`), `mergeable: MERGEABLE` — no conflicts expected, but the branch should be updated before merge regardless of the doc-pointer fix.
- **`closingIssuesReferences` is empty.** The PR body opens with "Fixes the reported symptom of #4246 and the active harm behind it" — GitHub's auto-close keyword matcher requires the keyword immediately followed by the issue reference, and "Fixes the reported symptom of #4246" doesn't match that pattern. Merging as-is will **not** auto-close #4246. If that's intended (the PR body itself flags two deliberately-scoped call-outs for reviewer judgment), fine — otherwise reword to `Fixes #4246` somewhere if auto-close is wanted.
- The pre-existing `trusty-code-gui`/`trusty-mpm-gui` clippy-on-missing-`ui/dist` issue the PR body calls out is real and unrelated to this change (reproduces on a clean `origin/main` checkout) — not re-litigated here.

bobmatnyc:
## Verdict: APPROVE

Updated review at head `451a3a30` (commit `382fa692` fixes the dangling doc pointer + clean merge of `main` `8976ed00`).

## What changed since the WARN

- `382fa692` — one-line fix: `probe_http.rs:488` now cites `tests::probe_no_address_when_nothing_resolves`, the test that actually exists. Confirmed no other file changed in this commit (`git show --stat`: 1 file, 1 insertion, 1 deletion).
- `451a3a30` — merge of `main` into the branch. Diffed `382fa692..451a3a30`: every changed file is outside `crates/trusty-installer/` (trusty-agents KG work, trusty-mpm worktree-reconcile work, etc. — all pre-existing `main` commits). Zero installer files touched by the merge itself, so nothing in the reviewed code changed as a side effect of the merge.

## Re-verified locally at the new head

- `cargo test -p trusty-installer --lib` → **506 passed, 0 failed, 4 ignored** (same as before the merge).
- `cargo clippy -p trusty-installer --all-targets -- -D warnings` → clean.
- `bash scripts/check_test_pointers.sh` → **0 dangling pointers — OK** (this is the exact script the previously-red job runs; confirmed the fix locally before checking CI).
- `gh pr checks 4381` → every check passes, including **"Test: doc-comment pointer lint"** (4m16s, the job that was failing in the WARN review). Only "Notify on red main" shows, and it's `skipping` by design (conditional on a red main build, not this PR).

No issues found at >80% confidence beyond what was already fixed. APPROVED — proceeding to the next pipeline stage.

## Notes

- `mergeStateStatus` is `BLOCKED` (not CI — all checks are green; `mergeable: MERGEABLE`, no conflicts). That's a branch-protection/review-approval gate, outside this review's scope.
- `closingIssuesReferences` is still empty for the reason noted in the prior review (the body's "Fixes the reported symptom of #4246" doesn't match GitHub's auto-close keyword pattern) — unchanged by these two commits, still worth a look before merge if auto-close of #4246 is wanted.

Issue #4270, opened 2026-07-29
converge in-project provisioning on the .base store (supersedes: retire the path)


bobmatnyc:
# Issue #4270: Worktree provisioning — converge on  (git convention)

## Owner Ruling 2026-08-05

**Direction:** New worktrees are provisioned at `<project-root>/.worktrees/<name>`, following the standard git convention.

The `.base` bare clone is retired as a provisioning target. The in-project path (`spawn_managed_inproject`) already produces this shape and is the target architecture for all new worktree creation.

## What Changes

| Component | Current | Target |
|---|---|---|
| `spawn_managed_routed` daemon entry | Routes to `spawn_managed_inproject` (in-project) or `spawn_managed_cloned` (`.base`) based on heuristic | Routes exclusively to `spawn_managed_inproject` |
| `spawn_managed_cloned` (`.base` path) | Active provisioning path | Retired; existing worktrees remain functional |
| `tm launch` CLI | Calls `inproject::ensure_base_clone` at [launch.rs:184](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/launch.rs#L184) | Routes through `spawn_managed_inproject` |
| `tm` guided default (daemon-unreachable fallback) | `ensure_base_clone` at [guided.rs:1223](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/guided_launch.rs#L1223) | Routes through `spawn_managed_inproject` |

## Explicitly Out of Scope

- **No auto-migration.** Existing `.base` worktrees are git-native artifacts. Registry entries hold absolute paths; they remain discoverable and functional via `git worktree list`.
- **No compatibility layer.** Resume and reactivation paths (`resume_managed`, `reactivate.rs`) are independent and remain unaffected.
- **Manual cleanup.** `.base` directory removal is deferred and is for manual operator execution.

## Context & Evidence

Phase-1 analysis (see thread) found:
- **24 of 31** sessions with worktrees use the in-project path
- **7 of 8** currently-live sessions use the in-project path
- The single live cloned session is this issue's originator

The in-project provisioning path is dominant and intended.

---

**Previous analysis** (kept for historical reference): See the full thread for the Phase-1 investigation, which stopped before any code changes were made.

bobmatnyc:
**The original plan does not survive contact with the code. A Phase-1 investigation stopped before any change was made.**

[...]

Issue #4269, opened 2026-07-29
project_index_id: drift table is incomplete and misdirects the wiring slice to reconcile on root


bobmatnyc:
Documentation defects in `crates/trusty-common/src/project_index_id.rs`, merged via PR #4262. The code is correct; the guidance it gives the not-yet-written wiring/migration slice is not. Found in that PR's review (https://github.com/bobmatnyc/trusty-tools/pull/4262#issuecomment-5111179237).

**1. `root` is described as immutable, and it is not.** `:112` and `:200` state unqualified that "only `root` is fixed for the life of a content tree", and `:220-222` recommends the wiring slice "reconcile on `root`, which is the only component that does not move." The correct qualifier exists at `:126` ("under ordinary *git* operations") but is absent where the advice is given.

Four actions move the id and appear in no row of the drift table, each verified against the crate:

| Action | Observed |
|---|---|
| `mv proj proj-renamed` | `acme-widget-88e6…` → `acme-widget-0bd2…` |
| GitHub repo rename | `acme-widget-6370…` → `acme-widget-renamed-b138…` |
| GitHub repo transfer | `acme-widget-6370…` → `newowner-widget-4d1e…` |
| `git remote remove origin` | `newowner-widget-4d1e…` → `widget-fa2a…` |

The directory rename is the significant one: it is routine, it orphans the registered index, and it removes the exact anchor the migration slice is told to reconcile on.

**2. CHANGELOG over-claims hermeticity.** `crates/trusty-common/CHANGELOG.md:25-26` states two callers "cannot derive different ids". `project_index_id.rs:323-329` correctly states they **can**, when `HOME`/`GIT_CONFIG_GLOBAL` differ and the repo sets no local `user.email`. The daemon-vs-CLI pair the CHANGELOG names as safe is precisely the pair at risk, since the daemon runs under a launchd plist environment while CLI invocations inherit the shell's.

**Resolution.** Add the four missing rows with tests pinning them (matching the existing `KNOWN DRIFT:` convention); qualify or remove the "only `root` is fixed" claim at `:112`/`:200`; and replace the reconcile-on-`root` recommendation at `:220-222` with guidance that survives a directory rename. Correct the CHANGELOG claim to match the module.

Mark this as required input for the #4207 wiring/migration slice — that slice will inherit this guidance, and reconciling on an anchor that moves on `mv` would reintroduce the orphaning class the identity work exists to remove.

Reference PR #4262 and epic #4207.

[...]

PR #4382, opened 2026-07-29
feat(trusty-mpm): reconcile worktrees on a git-derived key, report-only (#4288)

### LINKED ISSUE #4288: #4207 slice 3: reconcile worktrees on a git-derived key, report-only

bobmatnyc:
Part of #4207 (slice 3). Constraining input: #4269. Follow-on: #4270.

## What this fixes

Reconciliation had nothing to reconcile with, and the discriminator every consumer used — the path — is provably wrong. On the dogfood machine: 118 registered worktrees across two disjoint git registries (`.base` 60, sibling clone 58, intersection zero), 56 session records, 3 worktrees with a matching record. Path shape misclassifies 34 of the 118 (29%): two sibling directories under `.base/.worktrees/` belong to *different* registries.

## The key

```
( canonicalize(git -C <wt> rev-parse --path-format=absolute --git-common-dir),
  basename of <wt>'s admin dir, from the `gitdir:` line of <wt>/.git )
```

asserting `rev-parse --show-toplevel == canonicalize(<wt>)`. Git's own registry identity: no path convention, and it does not contain `root`. It is the only anchor that survives `mv proj`, a GitHub rename, a transfer, and `git remote remove origin` — and when it breaks (repo moved without `git worktree repair`) it breaks **loudly**, where a `root`-keyed scheme fails silently by re-deriving a fresh key.

The `show-toplevel` assertion is what makes the second landmine safe. `…/.worktrees/tm-trusty-tools-01` is a `workspace_path` in `sessions.json` with no `.git` file, whose toplevel resolves to the **parent checkout**. Any "workspace_path not in `git worktree list` ⇒ orphan" rule deletes files inside a live checkout. Here it gets no key, and **UNKNOWN dominates** — a path whose git identity cannot be established is never reported reclaimable, no matter what else points at it.

## Scope: report-only. Zero new destructive code paths.

Nothing is written, deleted, or migrated. `reconcile_worktrees` is a pure function of (registry scan, store snapshot, live cwds, clock); the CLI verb has no `--force`, `--dry-run`, or `--discard-dirty`, and `cli_reconcile_worktrees_takes_no_destructive_flag` turns adding one red. The existing reclaimer, its gates, and its dry-run default are untouched.

## In-scope items

| # | Item | Where |
|---|---|---|
| 1 | Reconciled inventory on the git key, three-state, reason per row | `session_manager/worktree_reconcile.rs` |
| 2 | **The excluded set is reported** — 33 of 118 here, six holding uncommitted work, invisible on every other surface | `worktree_registry::Admission` is now a value, not a `continue` |
| 3 | Proposed-adoption list — path, inferred owner, evidence. Names only | `ReconcileReport::proposed_adoptions` |
| 4 | Children before parents; never propose a candidate containing another | `sort_deepest_first`, `contains_other_entry` |
| 5 | Pin the unfiltered active set | already landed in #4294 (merged) |
| 6 | Amend the #4269 doc defects | `trusty-common/src/project_index_id.rs`, its CHANGELOG |

Slice 3 deliberately does **not** wire `project_index_id`, and performs zero sentinel writes.

## Item 4 is a real fix, not bookkeeping

Enumeration returned a `BTreeSet<PathBuf>`, so a parent sorted **before** its children. Nine registered worktrees on the dogfood machine are nested inside another. Removing a parent first takes the child's directory with it and leaves git holding a dangling registry entry. Order is now depth-descending. This changes order only — it cannot admit a path the containment rules already rejected.

## Acceptance test

`reconcile_keys_on_the_git_registry_not_the_path` — real repos in a `TempDir`, no mocks, no env gate, no `#[ignore]`. Fixture is the issue's counterexample: worktree `A` registered in the `.base` bare clone, `B` registered in the project checkout, both under the identical `<proj>/.base/.worktrees/` prefix. Baseline plus three mutations (`git worktree move`, `remote set-url origin`, `remote remove origin`) plus the `fake`-directory landmine, each asserted on counts **and** specific paths.

```
test session_manager::worktree_reconcile::worktree_reconcile_tests::reconcile_keys_on_the_git_registry_not_the_path ... ok
```

## Gates

```
cargo fmt --check                                  exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings   clean
cargo test -p trusty-mpm                           3970 passed; 0 failed; 6 ignored (lib)
                                                   + 1290, 1290, and 28 further binaries — 0 failed anywhere
bash scripts/check_line_cap.sh                     7 allowlisted, 0 violations — OK
bash scripts/check_test_pointers.sh                0 dangling pointers — OK
cargo test -p trusty-common                        223 passed; 0 failed; 6 ignored
```

Two files had to be split to stay under the caps: the CLI renderer moved to `commands/reconcile_worktrees.rs`, and both worktree routes now live in one merged sub-router, which takes `api.rs` from 1276 back to 1269 SLOC (below its frozen 1272 budget).

Closes #4288

### DIFF
--- a/crates/trusty-common/CHANGELOG.md
+++ b/crates/trusty-common/CHANGELOG.md
@@ -7,6 +7,22 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Fixed
+
+- **`project_index_id` documentation corrected**
+  ([#4269](https://github.com/bobmatnyc/trusty-tools/issues/4269), amended under
+  [#4288](https://github.com/bobmatnyc/trusty-tools/issues/4288)). The module
+  stated without qualification that `root` is immutable and recommended that the
+  wiring/migration slice "reconcile on `root`". Four routine actions move it —
+  `mv proj`, a GitHub repo rename, a repo transfer, and `git remote remove
+  origin` — and reconciling on it would reintroduce the silent-orphan class the
+  identity work exists to remove. The immutability claims are now qualified
+  ("under ordinary git operations"), the four movers are documented, and the
+  recommendation points at a git-maintained anchor instead. The CHANGELOG's
+  hermeticity claim is likewise corrected: two callers CAN derive different ids
+  when `HOME`/`GIT_CONFIG_GLOBAL` differ and the repo sets no local
+  `user.email`. Documentation only — no code change.
+
 ### Added
 
 - `json_rmw`: cross-process locked read-modify-write for whole-file JSON
@@ -34,8 +50,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   path; registry reconciliation and migration of existing indexes are separate
   slices of #4207. No behaviour change for any existing caller.
 
-  Derivation is hermetic: it reads no environment variable, so two callers on one
-  tree (e.g. the launchd daemon and a shell CLI) cannot derive different ids. The
+  Derivation reads no environment variable of its own, but it is NOT fully
+  hermetic (corrected, #4269): `resolve_operator_identity` shells out to `git
+  config`, so two callers on one tree CAN derive different ids when `HOME` or
+  `GIT_CONFIG_GLOBAL` differ and the repo sets no local `user.email` — see
+  `project_index_id.rs`'s own note. The launchd daemon and a shell CLI are
+  precisely that pair, since the daemon runs under a plist environment while CLI
+  invocations inherit the shell's. Set a repo-local `user.email` to pin it. The
   `index_id()` docs enumerate exactly which inputs are mutable — `origin` moves on
   the first commit, on `git remote add origin`, and on a new root commit — each
   pinned by a test, so the migration slice inherits a true guarantee rather than an
--- a/crates/trusty-common/src/project_index_id.rs
+++ b/crates/trusty-common/src/project_index_id.rs
@@ -109,9 +109,11 @@ const FALLBACK_LABEL: &str = "project";
 /// the account operating this checkout. All three are hashed; only `origin`
 /// (or the root basename) contributes to the readable label.
 ///
-/// Only `root` is immutable for the life of a content tree. `origin` and
-/// `operator` are both mutable git state — see the `Known limitation` block on
-/// [`Self::index_id`] for the enumerated drift cases and what they cost.
+/// `root` is the most stable of the three, but it is NOT immutable — it is
+/// fixed only under ordinary GIT operations (#4269). Renaming or moving the
+/// directory moves it, and `origin` and `operator` are both live git state.
+/// See the `Known limitation` block on [`Self::index_id`] for the enumerated
+/// drift cases and what each costs.
 /// Test: `sibling_clones_of_same_repo_derive_distinct_ids`,
 /// `linked_worktrees_of_same_repo_derive_distinct_ids`,
 /// `different_operators_derive_distinct_ids`.
@@ -197,9 +199,12 @@ impl ProjectIdentity {
     /// permanence.
     ///
     /// **Known limitation — which inputs are mutable, and what changes an id.**
-    /// Only `root` is fixed for the life of a content tree. Both other
-    /// components are live git/user state, so ordinary actions re-derive a
-    /// different id for the SAME tree. Each row is pinned by a test:
+    /// `root` is fixed only under ordinary GIT operations — NOT for the life of
+    /// a content tree (#4269). It moves whenever the directory does, and `mv` is
+    /// routine. Both other components are live git/user state, so ordinary
+    /// actions re-derive a different id for the SAME tree. Each row below is
+    /// pinned by a test; the four `root`/whole-triple movers named after the
+    /// table are pinned in #4269, not here:
     ///
     /// | Action | Component | Test |
     /// |---|---|---|
@@ -217,9 +222,36 @@ impl ProjectIdentity {
     /// deriving two ids over time wastes an index; two trees deriving one id
     /// merges unrelated codebases), and it cannot bite while the module has no
     /// call sites — but the wiring slice MUST NOT read this function as
-    /// permanent. Recommended shape for that slice: reconcile on `root`, which
-    /// is the only component that does not move, and treat an origin/operator
-    /// change as an alias-and-carry-forward rather than a new index.
+    /// permanent.
+    ///
+    /// **Do NOT reconcile on `root`** (#4269, corrected #4288). An earlier
+    /// version of this paragraph recommended exactly that, on the false premise
+    /// that `root` does not move. Four routine actions move it or the whole
+    /// triple, none of them a row in the table above, each verified against
+    /// this crate:
+    ///
+    /// | Action | Observed |
+    /// |---|---|
+    /// | `mv proj proj-renamed` | `acme-widget-88e6…` → `acme-widget-0bd2…` |
+    /// | GitHub repo rename | `acme-widget-6370…` → `acme-widget-renamed-b138…` |
+    /// | GitHub repo transfer | `acme-widget-6370…` → `newowner-widget-4d1e…` |
+    /// | `git remote remove origin` | `newowner-widget-4d1e…` → `widget-fa2a…` |
+    ///
+    /// The directory rename is the significant one: it is routine, it orphans
+    /// the registered index, and it removes the exact anchor the migration
+    /// slice was being told to reconcile on. `ProjectIdentity::digest` hashes
+    /// `path_bytes(&self.root)` and `label()` falls back to the root
+    /// basename, so `mv` moves BOTH halves of the id at once.
+    ///
+    /// Recommended shape for that slice instead: reconcile on an anchor GIT
+    /// itself maintains, and treat an origin/operator change as an
+    /// alias-and-carry-forward rather than a new index. The worktree
+    /// reconciliation in `trusty-mpm`'s `session_manager::worktree_reconcile`
+    /// (#4288) uses `(git rev-parse --git-common-dir, admin-dir basename)` for
+    /// this reason: it survives a directory move, a repo rename, a transfer,
+    /// and `git remote remove origin`, and when it does break — the repository
+    /// directory moved without `git worktree repair` — it breaks LOUDLY, where
+    /// a `root`-keyed scheme fails silently by re-deriving a fresh id.
     ///
     /// Also machine-local: `root` is an absolute path, so the same repo cloned
     /// on two machines derives two ids. That is correct for a partitioning key
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -7,6 +7,28 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Added
+
+- **`tm session reconcile-worktrees` — report-only worktree reconciliation on a
+  git-derived key** ([#4288](https://github.com/bobmatnyc/trusty-tools/issues/4288),
+  slice 3 of [#4207](https://github.com/bobmatnyc/trusty-tools/issues/4207)).
+  Reconciles every git-registered worktree against `sessions.json` on the pair
+  `(git rev-parse --git-common-dir, admin-dir basename)` — git's own registry
+  identity — rather than on a path shape, which was measured misclassifying 34
+  of 118 worktrees (29%) on the dogfood machine. Each row is classified
+  `LIVE` / `ORPHANED` / `UNKNOWN` with the reason that produced it; `UNKNOWN`
+  dominates, so a path whose git identity cannot be established is never
+  reported reclaimable no matter what else points at it.
+  - The report includes the **excluded** set — 33 of the 118, invisible to every
+    other worktree surface today, six of them holding uncommitted work.
+  - It also **names** the worktrees a later slice would adopt, with the evidence
+    behind each inference. Nothing is written: no deletions, no sentinel writes,
+    no registry or record mutation, and the verb has no `--force`.
+  - Worktree enumeration now orders a NESTED worktree before the worktree that
+    contains it. The previous lexicographic order put a parent first, so
+    removing it took its children's directories and left dangling registry
+    entries behind.
+
 ### Performance
 
 - `tm ls` no longer re-reads the whole fleet's deployed agent/skill trees on
--- a/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/cli.md
+++ b/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/cli.md
@@ -139,6 +139,7 @@ Generated from `Cli::command()` (clap's command-tree introspection) — every `t
   - `prune` — Prune managed sessions by state + compact tombstones (#1508)
   - `prune-idle` — Reclaim idle managed sessions: stop idle, decommission done (#1313)
   - `prune-worktrees` — Remove orphaned per-session git worktree directories (#1840)
+  - `reconcile-worktrees` — Report every git-registered worktree against `sessions.json` (#4288)
   - `rename` — Rename a managed session — from the list OR from within a session
   - `resume` — Resume a stopped/paused session (managed or project session)
   - `run` — Send a command to a session's tmux pane
@@ -171,6 +172,7 @@ Generated from `Cli::command()` (clap's command-tree introspection) — every `t
   - `prune` — Prune managed sessions by state + compact tombstones (#1508)
   - `prune-idle` — Reclaim idle managed sessions: stop idle, decommission done (#1313)
   - `prune-worktrees` — Remove orphaned per-session git worktree directories (#1840)
+  - `reconcile-worktrees` — Report every git-registered worktree against `sessions.json` (#4288)
   - `rename` — Rename a managed session — from the list OR from within a session
   - `resume` — Resume a stopped/paused session (managed or project session)
   - `run` — Send a command to a session's tmux pane
--- a/crates/trusty-mpm/src/bin/tm/cli/actions/session.rs
+++ b/crates/trusty-mpm/src/bin/tm/cli/actions/session.rs
@@ -490,4 +490,22 @@ pub(crate) enum SessionAction {
         #[arg(long)]
         discard_dirty: bool,
     },
+    /// Report every git-registered worktree against `sessions.json` (#4288).
+    ///
+    /// Why: `prune-worktrees` shows only the reclaim CANDIDATES, so the
+    /// worktrees it excludes — 33 of 118 on the machine this was measured on,
+    /// six of them holding uncommitted work — are invisible. This verb reports
+    /// the whole inventory, keyed on git's own registry identity rather than on
+    /// a path shape (which misclassified 29% of them).
+    /// What: GETs `/api/v1/sessions/managed/reconcile-worktrees` and prints one
+    /// line per worktree — state, path, reason — plus the worktrees a later
+    /// slice would adopt. READ-ONLY: there is no `--force`, because there is
+    /// nothing here to force. Nothing is deleted, written, or migrated.
+    /// Test: `cli_parses_session_reconcile_worktrees`,
+    /// `cli_reconcile_worktrees_takes_no_destructive_flag`.
+    ReconcileWorktrees {
+        /// Print the full report as JSON instead of the one-line-per-row form.
+        #[arg(long)]
+        json: bool,
+    },
 }
--- a/crates/trusty-mpm/src/bin/tm/commands/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/mod.rs
@@ -50,6 +50,7 @@ pub(crate) mod project;
 pub(crate) mod projects;
 pub(crate) mod prune;
 pub(crate) mod push_guard;
+pub(crate) mod reconcile_worktrees;
 pub(crate) mod rename;
 pub(crate) mod repair;
 pub(crate) mod serve_stdio;
--- a/crates/trusty-mpm/src/bin/tm/commands/reconcile_worktrees.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/reconcile_worktrees.rs
@@ -1,0 +1,99 @@
+//! `tm session reconcile-worktrees` — render the report-only worktree
+//! inventory (#4207 slice 3, #4288).
+//!
+//! Why: `managed.rs` sits against the 500-SLOC production cap, and this is a
+//! self-contained renderer rather than another managed-lifecycle verb — it
+//! issues one GET and formats the result, sharing no state with anything else
+//! in that file.
+//! What: one client function; see its doc for the output shape.
+//! Test: `cli_parses_session_reconcile_worktrees`,
+//! `cli_reconcile_worktrees_takes_no_destructive_flag`.
+
+/// `tm session reconcile-worktrees` — print the report-only worktree
+/// inventory (#4207 slice 3, #4288).
+///
+/// Why: an operator's only worktree view today is the reclaim-candidate list,
+/// which hides every excluded worktree — including a `workspace_path` that is
+/// really an ordinary subdirectory of a live checkout, and worktrees nested
+/// inside a live one. Both are landmines precisely because nothing prints them.
+/// What: GETs the reconcile route and prints `STATE  path` followed by the
+/// reason, then the proposed-adoption list (names only — this slice writes
+/// nothing). `--json` emits the raw report for scripting. There is no
+/// destructive form of this command.
+/// Test: `cli_parses_session_reconcile_worktrees`,
+/// `cli_reconcile_worktrees_takes_no_destructive_flag`.
+pub(crate) async fn session_reconcile_worktrees(
+    client: &reqwest::Client,
+    url: &str,
+    json: bool,
+) -> anyhow::Result<()> {
+    let resp = client
+        .get(format!("{url}/api/v1/sessions/managed/reconcile-worktrees"))
+        .send()
+        .await?;
+    let body: serde_json::Value = resp.error_for_status()?.json().await?;
+    if json {
+        println!("{}", serde_json::to_string_pretty(&body)?);
+        return Ok(());
+    }
+
+    let str_at = |v: &serde_json::Value, key: &str| -> String {
+        v.get(key)
+            .and_then(serde_json::Value::as_str)
+            .unwrap_or("<unreported>")
+            .to_owned()
+    };
+    let entries = body
+        .get("entries")
+        .and_then(serde_json::Value::as_array)
+        .cloned()
+        .unwrap_or_default();
+    for e in &entries {
+        // Deepest-first, as the report orders them: a nested worktree is always
+        // listed before the worktree that contains it.
+        println!(
+            "{:<9} {}",
+            str_at(e, "state").to_uppercase(),
+            str_at(e, "path")
+        );
+        println!("          {}", str_at(e, "reason"));
+    }
+    let n = |key: &str| -> u64 {
+        body.pointer(&format!("/counts/{key}"))
+            .and_then(serde_json::Value::as_u64)
+            .unwrap_or_default()
+    };
+    println!(
+        "{} worktree(s): {} live, {} orphaned, {} unknown \
+         ({} reclaim candidates, {} excluded, {} named only by sessions.json)",
+        n("total"),
+        n("live"),
+        n("orphaned"),
+        n("unknown"),
+        n("admitted"),
+        n("excluded"),
+        n("record_only"),
+    );
+
+    let adoptions = body
+        .get("proposed_adoptions")
+        .and_then(serde_json::Value::as_array)
+        .cloned()
+        .unwrap_or_default();
+    if adoptions.is_empty() {
+        return Ok(());
+    }
+    println!(
+        "\nwould be adopted by a later slice ({} — NOTHING was written):",
+        adoptions.len()
+    );
+    for a in &adoptions {
+        println!("  {}", str_at(a, "path"));
+        println!(
+            "    owner {} — {}",
+            str_at(a, "inferred_owner"),
+            str_at(a, "evidence")
+        );
+    }
+    Ok(())
+}
--- a/crates/trusty-mpm/src/bin/tm/commands/session.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session.rs
@@ -401,6 +401,12 @@ pub(crate) async fn session(
             crate::commands::managed::session_prune_worktrees(client, url, !force, discard_dirty)
                 .await?
         }
+        // #4288: the report-only inventory. No `force`/`dry_run` argument
+        // exists because the verb has no destructive form.
+        SessionAction::ReconcileWorktrees { json } => {
+            crate::commands::reconcile_worktrees::session_reconcile_worktrees(client, url, json)
+                .await?
+        }
         // #2444: re-sync a live session's (or every syncable session's)
         // deployed assets against the current catalog. Fleet-wide store
         // operation like `Prune`/`DecommissionEphemeral` above — direct HTTP,
--- a/crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs
@@ -585,6 +585,44 @@ fn cli_prune_worktrees_discard_dirty_is_opt_in() {
     }
 }
 
+/// #4288: `reconcile-worktrees` parses, and defaults to the human-readable
+/// form.
+#[test]
+fn cli_parses_session_reconcile_worktrees() {
+    let cli = Cli::try_parse_from(["trusty-mpm", "session", "reconcile-worktrees"]).unwrap();
+    match cli.command.unwrap() {
+        Command::Session {
+            action: SessionAction::ReconcileWorktrees { json },
+        } => assert!(!json, "the default must be the readable report, not JSON"),
+        other => panic!("expected session reconcile-worktrees, got {other:?}"),
+    }
+    let as_json =
+        Cli::try_parse_from(["trusty-mpm", "session", "reconcile-worktrees", "--json"]).unwrap();
+    match as_json.command.unwrap() {
+        Command::Session {
+            action: SessionAction::ReconcileWorktrees { json },
+        } => assert!(json),
+        other => panic!("expected session reconcile-worktrees, got {other:?}"),
+    }
+}
+
+/// #4288: the reconcile verb has NO destructive flag — not `--force`, not
+/// `--discard-dirty`, not `--dry-run`.
+///
+/// Why: slice 3 is report-only, and the way a reporting verb stops being
+/// report-only is that someone adds a flag to it. `clap` rejecting these at
+/// parse time is the mechanical form of "there is no destructive form of this
+/// command", and this test turns any such addition red.
+#[test]
+fn cli_reconcile_worktrees_takes_no_destructive_flag() {
+    for flag in ["--force", "--discard-dirty", "--dry-run"] {
+        assert!(
+            Cli::try_parse_from(["trusty-mpm", "session", "reconcile-worktrees", flag]).is_err(),
+            "`reconcile-worktrees {flag}` must not parse — slice 3 writes nothing"
+        );
+    }
+}
+
 #[test]
 fn cli_parses_sessions_sync_assets() {
     // #2444: `sync-assets <id>` re-syncs ONE session.
--- a/crates/trusty-mpm/src/daemon/api.rs
+++ b/crates/trusty-mpm/src/daemon/api.rs
@@ -134,9 +134,9 @@ use super::managed_routes::{
     decommission_managed_session, delete_managed_session, fleet_by_project_route, get_attach_cmd,
     get_managed_session, get_project_registry_route, get_session_activity, list_managed_sessions,
     list_projects_registry_route, patch_project_registry_route, project_status_route, proxy_router,
-    prune_managed_route, prune_worktrees_route, reactivate_managed_session,
-    register_project_registry_route, rename_managed_session, resume_managed_session,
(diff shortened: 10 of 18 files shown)


bobmatnyc:
## Verdict: APPROVE

Zero CRITICAL, zero HIGH. Four MEDIUM, two LOW. Reviewed adversarially against #4288's AC and #4269, with the fact that this report's classifications feed #4270 deletion — so every attack below was aimed at producing a false ORPHANED.

**No path to a false ORPHANED was found.** The safety is layered, not single-gated: `ReconcileState::Orphaned` requires `admission == Some(Admission::Admitted)`, which only `git worktree list --porcelain` can produce (`worktree_registry.rs:538-553`), **and** a resolved git key, **and** a parsed sentinel naming a provably-gone owner, **and** a clean tree. Every false-positive key I could construct (below) fails at the `Admitted` gate.

### What was verified, not assumed

**UNKNOWN dominates — no promotion path exists.** `classify` (`worktree_reconcile.rs:519-532`) returns early on `key.is_none()`. Nothing downstream writes `state`: the `contains_other_entry` loop (`:431-435`) touches only that flag, `count` (`:452-470`) only tallies, `propose_adoptions` (`:676-716`) only reads, and both surfaces render. Confirmed.

**Report-only holds.** Grepping the diff for `remove_dir_all|remove_file|fs::write|fs::create|upsert|worktree remove|worktree prune|rename(` returns hits in test files only (`worktree_reconcile_tests.rs:1636,1678,1679,1726,1826,1849,1876` in diff coordinates). `inspect_dirt` is the only probe that touches git internal state (`git status` refreshes `.git/index`), and it is the pre-existing gate #4288 said to inherit unchanged. `cli_reconcile_worktrees_takes_no_destructive_flag` is the right mechanical pin.

**`inspect_dirt` fails CLOSED, which I checked because a fail-open here is the catastrophic shape.** `worktree_safety.rs:323-345`: a probe error returns `Some(DirtyWorktree{reason: "dirty-check failed"})`, not `None`. An unanswerable dirt probe reads UNKNOWN, never clean. Correct.

**Symlink / canonicalization consistency.** Every producer canonicalizes: scan paths (`worktree_registry.rs:543`), `by_workspace` keys (`worktree_reconcile.rs:401`), `live_cwds` (`:750`), and `worktree_key` itself (`:148,150,155`). The one divergence I looked for — a case-only difference on APFS/HFS+ producing two rows for one worktree — fails to UNKNOWN, because `canonicalize(toplevel) != canonical` at `:150` (macOS `realpath` does not case-normalize). Safe direction.

**Deviation 1 (`counts.admitted == 2` for "inventory length == 2") is sound.** In-scope item 2 requires the excluded set be reported, so `entries.len()` cannot be 2. The substitution asserts strictly more than the literal AC (admitted count + the exact admitted path set + excluded count + the specific `MainCheckout`/`Bare` verdict on each excluded row, `worktree_reconcile_tests.rs:147-176`) and still goes red on both failure modes the AC names. Not scope creep.

**Deviation 2 (descriptive `category`) is sound but is an addition, not a reading.** It is a field #4288 did not ask for. It is pinned inert in both directions by `categories_are_descriptive_and_do_not_gate_state` (`:480-508`) and never read by `classify`. Acceptable; noted, not flagged.

**`closingIssuesReferences` is exactly `[4288]`.** CI green on `2601864` across Test (16m57s), Clippy, fmt, MSRV, and the 500-line cap.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 394-397 | Two-registry rows collapse on path alone, silently dropping one registry's verdict | Merge duplicates, preferring `Admitted`, instead of last-wins |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 520-531 | Every bare-clone row is told to run `git worktree repair` | Prefer `s.admission.reason()` when the admission already explains the row |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 147-158, 196-205 | `WorktreeKey` is well-formed for submodules and nested independent repos — verified | Assert `<common_dir>/worktrees/<admin_dir>` exists |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 699-713 | Branch-match adoption can name one session as owner of two worktrees | Drop or mark ambiguous when the match is not unique |
| LOW | `crates/trusty-mpm/src/session_manager/prune.rs` | 421-422 | Doc still says "sorted"; the order is now depth-descending | Update the wording |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 65-69 | Module doc credits `show-toplevel` alone with landmine 2; the code gates it twice | Name both gates |

### MEDIUM 1 — the two-registry collapse (`worktree_reconcile.rs:394-397`)

```rust
let scanned: BTreeMap<PathBuf, ScannedWorktree> = scan_registered_worktrees(repos_root)
    .into_iter()
    .map(|s| (s.path.clone(), s))
    .collect();
```

`scan_registered_worktrees` deliberately emits **one row per (path, registry_root)** — that is what `scan_reports_both_registries_for_the_same_path_prefix` exists to prove. This `collect` keys on path alone, so when both registries list the same canonical path, all but the last-iterated row are discarded, along with its `admission` and `branch`.

Why it matters here specifically: `enumerate_registered_worktrees` (`worktree_registry.rs:311-315`) filters `admission == Admitted` over the **whole** scan, so the reclaimer admits a path if **any** registry admits it. The report keeps whichever registry sorts last by `registry_root`. If registry A admits and registry B locks, and B sorts later, the report prints `excluded: git-locked by the operator` for a path the reclaimer would treat as a candidate. That is the false-preview defect #4288 names, inverted — and `counts.admitted`/`counts.excluded` undercount either way.

Measured intersection on the dogfood machine is zero, which is why this is MEDIUM rather than HIGH. But zero-intersection is a property of today's data, not an invariant, and the module's entire premise is that two registries exist.

Fix: fold duplicates explicitly rather than relying on `collect`'s last-wins, choosing `Admitted` when any row admits so the report agrees with the reclaimer:

```rust
let mut scanned: BTreeMap<PathBuf, ScannedWorktree> = BTreeMap::new();
for s in scan_registered_worktrees(repos_root) {
    match scanned.entry(s.path.clone()) {
        Entry::Vacant(v) => { v.insert(s); }
        // Match `enumerate_registered_worktrees`: any registry admitting the
        // path makes it a candidate, so the report must say so too.
        Entry::Occupied(mut o) if s.admission == Admission::Admitted => { o.insert(s); }
        Entry::Occupied(_) => {}
    }
}
```

### MEDIUM 2 — every bare clone is told to repair itself (`worktree_reconcile.rs:520-531`)

```rust
if key.is_none() {
    entry.reason = match scanned {
        Some(_) => "no git-derived key: `rev-parse --show-toplevel` disagrees this path is \
                    its own worktree root, or git could not answer — repair the registry \
                    (`git worktree repair`) before trusting any verdict here"
```

`git rev-parse --show-toplevel` fails **by design** in a bare repository. Verified on this machine:

```
$ git -C bare.git rev-parse --path-format=absolute --show-toplevel
fatal: this operation must be run in a work tree
exit=128
```

So `worktree_key` returns `None` for every `Admission::Bare` row, and every such row lands in the `Some(_)` arm. The `.base` clone of every managed project is a guaranteed `Bare` row. `Prunable` rows hit the same arm.

The CLI prints exactly `STATE  path` then the reason (`commands/reconcile_worktrees.rs:44-51`), so on the real machine an operator gets, once per project, a healthy bare clone flagged `UNKNOWN` with an instruction to run `git worktree repair` on it. The row already carries the correct verdict in `admission`, and the reason discards it. `Admission::reason()` exists precisely so wordings do not drift, and this is the one place that bypasses it.

`worktree_reconcile_tests.rs:173-176` asserts the `.base` row's `admission == Bare` but never asserts its `reason`, which is why this is uncovered.

Fix — let the admission speak when it already explains the row:

```rust
entry.reason = match scanned {
    // Bare/Prunable have no worktree to resolve a key against — that is the
    // verdict, not a broken registry. Do not send the operator to `repair`.
    Some(s) if matches!(s.admission, Admission::Bare | Admission::Prunable) => {
        format!("unknown: {}", s.admission.reason())
    }
    Some(_) => "no git-derived key: …".to_string(),
    None => "no git-derived key AND no git registry entry: …".to_string(),
};
```

### MEDIUM 3 — `WorktreeKey` is well-formed for things that are not worktrees

`worktree_key` (`:147-158`) asserts `show-toplevel == canonical(candidate)` and then reads `--git-common-dir` plus the `gitdir:` basename. `admin_dir_name` (`:196-205`) returns the basename of *whatever* `gitdir:` points at; nothing checks that `<common_dir>/worktrees/<admin_dir>` exists. Two shapes pass the whole derivation. Both verified directly:

```
# git submodule at sup/vendor/sub
$ cat sup/vendor/sub/.git                  -> gitdir: ../../.git/modules/vendor/sub
$ git -C sup/vendor/sub rev-parse --show-toplevel   -> .../sup/vendor/sub   (matches)
$ git -C sup/vendor/sub rev-parse --git-common-dir  -> .../sup/.git/modules/vendor/sub
=> key = (".../sup/.git/modules/vendor/sub", Some("sub"))

# nested independent repo at proj/.worktrees/indep (git init inside a checkout)
$ git -C proj/.worktrees/indep rev-parse --show-toplevel  -> .../proj/.worktrees/indep (matches)
=> key = (".../proj/.worktrees/indep/.git", None)
$ git -C proj worktree list --porcelain                   -> lists only proj
```

Neither is a worktree-registry entry, and the submodule key is structurally indistinguishable from a real one.

**This is inert in slice 3** — both reach `classify` only via a `SessionRecord.workspace_path`, so `scanned` is `None`, `admission` is `None`, and `:563-569` returns UNKNOWN. That is why this is MEDIUM, not HIGH.

It matters anyway for two reasons. First, `WorktreeKey`'s doc at `:112-119` states the `admin_dir` *is* the basename of `<common_dir>/worktrees/<name>`; for a submodule that is false, so the type does not mean what it says. Second, the type derives `Ord`/`Eq`/`Serialize` and #4270 inherits it — the moment anything treats "key resolves" as "this is a registered worktree", or uses the key as a map key, the guarantee this slice actually relies on (`Admission::Admitted`) is gone.

Fix — one check makes the key mean its docs:

```rust
let common_dir = std::fs::canonicalize(&git_rev_parse_path(&canonical, "--git-common-dir")?).ok()?;
let admin_dir = admin_dir_name(&canonical);
// A linked worktree's admin dir MUST live in this registry. A submodule's
// `gitdir:` points into `<super>/.git/modules/…`, which does not.
if let Some(name) = admin_dir.as_deref()
    && !common_dir.join("worktrees").join(name).is_dir()
{
    return None;
}
```

with a test for the submodule shape alongside `key_is_none_for_a_plain_directory_inside_a_checkout`.

### MEDIUM 4 — branch-match adoption has no uniqueness check (`worktree_reconcile.rs:699-713`)

```rust
if let Some(branch) = e.branch.as_deref()
    && let Some(r) = records
        .iter()
        .find(|r| r.branch.as_deref() == Some(branch) && r.workspace_path.is_none())
```

`find` takes the first match and there is no uniqueness check in either direction. Git forbids two worktrees checked out on one branch *within a registry* — but #4288's world is **two disjoint registries of the same repo**, where two worktrees can absolutely carry the same branch name. Both then get a `ProposedAdoption` naming the same `inferred_owner`, and the evidence string reads as if the inference were unambiguous:

> `branch match: session {id} records branch \`{branch}\` and has no workspace_path of its own`

Nothing is written in this slice, so no damage today. But the adoption list is the input to the backfill slice, and a wrong adoption is a route to a wrong ORPHANED: adopting a live worktree under a terminal session's id makes it `provably_ownerless` on the next run. #4288's stated purpose for item 3 is that the write "be reviewed before it exists" — an ambiguous proposal presented as unambiguous defeats that.

Fix: count matches first and either skip or mark ambiguous.

```rust
let mut matches = records
    .iter()
    .filter(|r| r.branch.as_deref() == Some(branch) && r.workspace_path.is_none());
let (Some(r), None) = (matches.next(), matches.next()) else { continue };
```

Plus a guard so one record is not proposed as owner of two entries. A test with two same-branch worktrees in two registries would pin it.

### LOW 5 — stale doc (`prune.rs:421-422`)

> "Candidates come back canonicalized …, **sorted**, and de-duplicated."

Still de-duplicated, no longer lexicographically sorted. `sort_deepest_first` is the right fix for item 4 and is strictly safer for the reclaimer (children de-registered before their container), but the contract sentence a caller reads is now wrong.

### LOW 6 — the module doc understates its own boundary (`worktree_reconcile.rs:65-69`)

> "That is what keeps the second landmine safe — `…/.worktrees/tm-trusty-tools-01` … has no `.git` file …"

The landmine is gated **twice**: `key.is_none()` at `:520` *and* `admission != Some(Admitted)` at `:563`. `worktree_key` also returns `Some` for shapes that are not worktrees (MEDIUM 3), so the `show-toplevel` assertion is not the sole boundary the doc claims. This repo has already been bitten by exactly this reading — two boundaries that each look redundant in isolation, one of which gets tidied away. Name both.

---

## Notes

**On "does UNKNOWN swallow so much the report is useless?"** Projected against #4288's own measurements: ~112 of 118 rows UNKNOWN, ~3-5 LIVE (only 3 worktrees have a matching record; `live_cwds` covers managed tmux panes only), and 0-1 ORPHANED (3 of 85 candidates carry a sentinel, 2 of those zero-byte → `SentinelOwner::Unknown`). That is not a defect — #4288 states the expected outcome as "82 of 85 admitted candidates are UNKNOWN; effective reclamation is 0 of 118". The report's value is the union, the excluded set, and the per-row reason, and the reasons do discriminate (`unknown: no ownership sentinel … the 82-of-85 case` vs. `unknown: excluded: <rule>` vs. `no git-derived key`). MEDIUM 2 matters more than it looks precisely because the reason line is the only differentiating output on ~95% of rows.

**On sentinel staleness (attack 3).** `read_sentinel_owner` does no staleness validation, and `provably_ownerless` (`:601-611`) reproduces `resolve_ownerless_with_grace` (`worktree_ownership.rs:258-267`) exactly: terminal-if-found, `OWNERLESS_GRACE`-if-absent. A sentinel left by a dead session is *meant* to resolve ORPHANED — that is the design. The one divergence is source: reconcile reads a `self.list()` snapshot, the reclaimer calls `self.get()` per owner. Both directions of any disagreement land on "absent → grace window", so it is a snapshot-consistency choice, not a bug. The `.claude/worktrees` + sentinel case the brief asks about resolves `SessionOwned`, which `categories_are_descriptive_and_do_not_gate_state` proves cannot reach `state`.

**On the destructive-path change shipped in a report-only slice.** `sort_deepest_first` changes the order in which `prune_orphaned_worktrees` removes worktrees. #4288 item 4 asks for it explicitly, it admits nothing new (`enumerate_registered_worktrees` still filters `Admitted` over the same scan), and depth-descending is strictly safer. Correctly in scope — flagged only so it is not read as "nothing touched the reclaim path".

**Not flagged.** `git status` inside `inspect_dirt` writes `.git/index` stat data — pre-existing, inherited unchanged per #4288, and not a mutation of anything the report claims not to mutate. `git_command` has no timeout, so a stalled mount would hang the blocking task and the GET; also pre-existing and workspace-wide.

bobmatnyc:
## Verdict: APPROVE (re-review of `2601864..da3b6fec`)

Both MEDIUMs are genuinely fixed. Zero CRITICAL, zero HIGH. Two new MEDIUM and two new LOW, all in the same class as the thing being fixed — operator-facing text and counting, none of them a route to a false ORPHANED. Scope of this pass: the incremental diff only.

### The fixture is real — verified independently, not taken on faith

`forge_duplicate_registration` hand-writes an admin dir, which is the kind of fixture that usually proves nothing. I reproduced it against real git before reading the assertions:

```
$ git -C repo/.base worktree list --porcelain
worktree /private/var/.../repo/.base
bare

worktree /var/.../repo/.worktrees/claimed-twice     <-- forged entry, enumerated
HEAD 25e0209b…
detached

$ git -C repo worktree list --porcelain
worktree /private/var/.../repo
…
worktree /private/var/.../repo/.worktrees/claimed-twice
branch refs/heads/session/ct
```

Git enumerates the forged entry, does **not** mark it `prunable`, and the two registries **disagree about the branch** (`.base` says detached, `<repo>` names the branch). So `ScannedKey` genuinely differs between the two claims — the `Vec<ScannedWorktree>` grouping at `:405` is load-bearing, not decorative, and `assert!(rows.iter().all(|r| r.admission == Some(Admitted)))` is a real assertion rather than a tautology. The fixture reaches residual 3 from #4288 as claimed.

### Answering the question directly: is contested ⇒ UNKNOWN-everywhere correct?

**"Never reclaimable" is correct, and I'd defend it against the author's own weaker justification.** The comment at `:574-577` argues "the key names exactly one registry, so at most one claim is right and nothing here can say which." The second half is wrong — I checked, and the key *does* disambiguate:

```
$ git -C repo/.worktrees/claimed-twice rev-parse --git-common-dir
/private/var/.../repo/.git          <-- the repo registry, not .base
$ cat repo/.worktrees/claimed-twice/.git
gitdir: /private/var/.../repo/.git/worktrees/claimed-twice
```

The worktree's `.git` file points back at exactly one admin dir, so the report *can* say which claim the worktree honours — and it already does, on `entry.key.common_dir`, on both rows.

The refusal to reclaim is still right, for a better reason than the one given: the linkage is **bidirectional and each half can be stale independently**. `.base/worktrees/forged-dup/gitdir` points at the worktree while the worktree points at `<repo>/.git`. Git will let `.base` run `worktree remove`/`prune` off its own admin dir regardless of where the worktree points. "The key says A" therefore does not mean "B cannot delete it", which is exactly why neither row may be reclaimable. Worth correcting the comment so a later reader does not delete the rule on the grounds that the premise is false.

**Suppressing the LIVE verdict is not correct — that part does hide information the report should carry.** The contested branch at `:578-585` returns *before* the LIVE vetoes at `:587`. Detail below (MEDIUM 1).

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 578-585 | `contested` returns before the LIVE vetoes, so a contested worktree a live session owns reports UNKNOWN and never mentions the session | Move the block below the LIVE vetoes; append the contest to the LIVE reason |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 549-555 | `Unresolvable` is told "git answering correctly rather than a registry to repair" — but it means a stalled mount / unreadable parent, the archetypal unanswerable probe | Split `Unresolvable` into its own arm |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 429 | `contested = claims.len() > 1` counts claims, not distinct registries | Count distinct `registry_root`s |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_reconcile.rs` | 452-470 + CLI | `counts.total` is now a ROW count; the CLI still prints "{total} worktree(s)" | Count distinct paths, or relabel |

### MEDIUM 1 — the contest suppresses the LIVE veto (`:578-585`, before `:587`)

```rust
    if contested {
        entry.reason = "contested: more than one git registry lists this path; …
                        resolve the duplicate registration … before acting on any row for this path"
            .to_string();
        return entry;
    }

    // 2. LIVE vetoes. Each is a veto on reclamation, never a vote for it.
    if let Some(r) = owners.first() {
```

A contested path with a live session record — the exact case the new test builds, `record_at(Active, Some(contested))` — reports `UNKNOWN` with a reason that says "resolve the duplicate registration" and **never mentions that a session is inside it**. The CLI renders only `state`, `path`, `reason` (`commands/reconcile_worktrees.rs:44-51`), so the operator's primary surface loses the more urgent of the two facts. `counts.live` also omits it entirely.

No safety regression — LIVE and UNKNOWN are both non-reclaimable, and `entry.sessions` is still populated in the JSON. But this module's whole premise is that "two states cannot express *I could not observe this*" (`:209-212`), and this is the inverse error: a state that **was** observed, collapsed into the could-not-observe bucket. Under #4288's own rule — "each signal is a veto on reclamation, never a vote for it" — reporting LIVE costs nothing on safety and is strictly more informative.

Fix — reorder, and carry both facts. Every safety property survives: `Orphaned` is still unreachable for a contested row because the block sits above the step-3 gate, and `propose_adoptions` filters `e.contested` independently at `:742`.

```rust
    // 2. LIVE vetoes. A live signal is STRONGER evidence than the contest, and
    //    both are non-reclaimable, so report the stronger one.
    let contest_note = if contested {
        " (ALSO contested: a second git registry lists this path — resolve the \
          duplicate registration)"
    } else {
        ""
    };
    if let Some(r) = owners.first() {
        entry.reason = format!("live: session {} records this as its workspace_path …{contest_note}", r.id);
        entry.state = ReconcileState::Live;
        return entry;
    }
    // … remaining LIVE vetoes, each appending `contest_note` …

    // 2b. Contested and not otherwise LIVE: never reclaimable.
    if contested { /* existing reason */ return entry; }
```

Extend `two_registries_claiming_one_path_keep_both_verdicts` to assert the live-record row reads `Live` with a reason naming both the session and the contest, and that `counts.orphaned == 0` still holds.

### MEDIUM 2 — `Unresolvable` is given a reason that is wrong for its actual causes (`:549-555`)

```rust
Some(a @ (Admission::Bare | Admission::Prunable | Admission::Unresolvable)) => {
    format!(
        "{} — no working tree to key on, which is git answering correctly rather \
         than a registry to repair",
        a.reason()
    )
}
```

Correct for `Bare` and `Prunable`. Wrong for `Unresolvable`, which does not mean "git answered"; it means `std::fs::canonicalize` failed. `worktree_registry.rs`'s own comment on that branch enumerates the causes: *"A dangling symlink, a deletion race, an unreadable parent, or a stalled mount"*, and `Admission::Unresolvable`'s docstring says *"nothing about it is comparable"*. A stalled mount is the archetypal unanswerable probe, and this module is governed by `worktree_registry.rs:37-48` — **"an unanswerable probe is never a verdict."** Telling an operator "git answered correctly, nothing to repair" about a path on a hung NFS mount is a new wrong statement in place of the old one; the text it replaced at least said "or git could not answer".

The new test asserts only `matches!(gone.admission, Some(Prunable | Unresolvable))` and `!contains("worktree repair")`, so the `Unresolvable` wording is uncovered — and in that fixture `remove_dir_all` yields `Prunable`, so the `Unresolvable` arm is never exercised at all.

```rust
Some(a @ (Admission::Bare | Admission::Prunable)) => format!(
    "{} — no working tree to key on, which is git answering correctly rather \
     than a registry to repair",
    a.reason()
),
// NOT "git answered correctly": canonicalization failed, so this is an
// unanswerable probe (dangling symlink, deletion race, unreadable parent,
// stalled mount) and nothing about the path is comparable.
Some(a @ Admission::Unresolvable) => format!(
    "{} — the path could not be observed at all; re-run once the filesystem \
     answers before drawing any conclusion about it",
    a.reason()
),
```

### LOW 3 — `contested` counts claims, not registries (`:429`)

```rust
let contested = claims.len() > 1;
```

`scan_registered_worktrees` dedups on the full `(path, registry_root, ScannedKey)` tuple, not on `(path, registry_root)` as its doc claims. So one registry scanned under two different project boundaries — a symlinked project directory, or a `.base` symlinked into a second `<owner>/<repo>` — yields two claims with the **same** `registry_root` and different `ScannedKey.project`/`admission`. That trips `contested` and prints "more than one git registry lists this path" when there is only one.

Direction is safe (UNKNOWN, no adoption), and the shape is exotic, hence LOW. But the fix is one line and makes the flag mean its name:

```rust
let distinct: BTreeSet<&PathBuf> = claims.iter().map(|c| &c.registry_root).collect();
let contested = distinct.len() > 1;
```

### LOW 4 — `total` is now a row count, still rendered as a worktree count

`ReconcileCounts::total` is documented as "Every row in the inventory" and `count` (`:452-470`) tallies per row. Rows are now per-`(path, registry)`, so a contested path counts twice — twice in `unknown`, twice in `admitted`, and twice in `total`. The CLI prints `"{} worktree(s): …"`. Only diverges in the state the fix exists to surface, so LOW; either count distinct paths for `total` or relabel the line "row(s)".

---

## Notes

**The prior findings I asked to keep on the record are untouched, as required.** Verified against `da3b6fec`, not assumed:

| Prior finding | Status |
|---|---|
| MEDIUM — `WorktreeKey` well-formed for submodules / nested repos; `<common_dir>/worktrees/<name>` never checked | `worktree_key` unchanged; `admin_dir_name` unchanged at `:196-205` |
| MEDIUM — branch-match adoption has no uniqueness check | `records.iter().find(…)` unchanged at `:756-759` |
| LOW — `prune.rs` doc still says "sorted" | unchanged |
| LOW — module doc credits `show-toplevel` alone with landmine 2 | unchanged at `:63-69` |

**A non-obvious dependency the new code now leans on, worth pinning.** The contest is only detectable because the two registries' claims group under one map key — and they only do so because `scan_from_anchor` canonicalizes (`worktree_registry.rs:543`). In my reproduction the two registries reported *different spellings of the same path*:

```
.base  ->  /var/folders/.../repo/.worktrees/claimed-twice
repo   ->  /private/var/folders/.../repo/.worktrees/claimed-twice
```

Without that canonicalize, a contested path becomes **two separate single-claim entries**, each with `claims.len() == 1`, each reported non-contested — and the one with no session record can reach ORPHANED. So the new rule's safety is downstream of a canonicalize in a different module that nothing in `worktree_reconcile_tests.rs` pins. A one-line assertion in the new test that both rows share one `path` would turn its removal red.

**The author's mutation evidence matches my independent reading.** Reverting the grouping leaves one surviving row; in this fixture the `.base` claim reports the worktree `detached` while `<repo>` reports it on its branch, so last-wins really can keep the wrong registry's view — and with a record present it would read LIVE off the wrong claim. Consistent with what they reported.

**Not flagged.** The comment block at `:556-562` explains the match arm *above* it rather than the one below — pure placement, LOW at most, calling it out here instead of padding the table.

PR #4384, opened 2026-07-30
fix(trusty-mpm): tm ls/picker auto-prunes dead records + groups attached-active-stopped

bobmatnyc:
## Summary

Owner request 2026-07-29. Two picker fixes:

1. **Auto-prune DEAD records.** Bob's pasted `tm ls` output showed rows like
   `[50] tm-tagents-dead-20260722 (stopped) — DEAD: workspace removed; use
   [d50] to remove the record`. Before this PR, that row sat forever — the
   operator had to notice it and manually type `d<N>`. `fetch_live_sessions`
   (the picker's shared fetch path — used by bare `tm`, `tm ls`'s interactive
   branch, and the picker's own re-fetch loop) now auto-decommissions every
   session the daemon flags `unresumable` and prints a one-line
   `tm: pruned N dead record(s) (workspace gone)` summary instead of prompting.

2. **Ordering.** The listing now groups **attached → active → everything else**
   ABOVE whichever `recent`/`alpha` secondary sort was requested (`tm ls
   recent|alpha` from #3485 is unchanged as the secondary key within each
   group).

## Safety boundary (#4344)

`unresumable` (`session_manager::resume_workdir::is_unresumable`) is already
the exact narrow predicate this needed: `true` only for a `stopped`/`errored`
record whose *every* workdir candidate (`last_cwd`, `workspace_path`, `cwd`)
is confirmed absent from disk via `tokio::fs::try_exists`.

#4344 (merged today, `114de333`) made a record whose worktree removal was
refused for being dirty retain its `workspace_path` pointing at a directory
that **still exists**. Because `is_unresumable` probes that exact path and
returns `false` the instant any candidate is found present, such a record can
never read `unresumable == true` in the first place — it is structurally
excluded from auto-prune before this PR's partition logic even runs. No
extra existence check was needed to enforce that boundary; a test
(`auto_prune_dead_records_keeps_workspace_present_records`) pins it directly
against a real stopped session with an on-disk workspace.

The actual removal is routed through the existing `/decommission` HTTP route
— the *same* dirty-worktree-gated teardown path #4344 hardened — never a raw
`fs::remove` or a hand-edited store entry.

## Changes

- `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` (new):
  `auto_prune_dead_records` (partition + drive) and `decommission_dead_record`
  (the one HTTP primitive, best-effort per record). Extracted out of
  `session_picker.rs` to keep it under the 500-SLOC production cap.
- `crates/trusty-mpm/src/bin/tm/commands/session_picker.rs`: wires the
  auto-prune into `fetch_live_sessions`; adds `group_rank` and updates
  `sort_sessions` to key on it first.
- `crates/trusty-mpm/src/bin/tm/commands/mod.rs`: registers the new module.
- `crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs`: new tests —
  `auto_prune_dead_records_removes_unresumable_records` (real loopback daemon,
  a genuinely dead record disappears and the daemon's own store reflects the
  teardown), `auto_prune_dead_records_keeps_workspace_present_records` (the
  #4344 boundary), `auto_prune_dead_records_is_noop_when_nothing_is_dead`,
  `sort_sessions_recent_groups_attached_before_active_before_stopped`,
  `sort_sessions_alpha_groups_attached_before_active_before_stopped`.
- `crates/trusty-mpm/CHANGELOG.md`: `[Unreleased]` entry.

No existing open issue covers either behavior (searched `tm ls dead`,
`picker order`, `auto prune record`) — this PR stands alone per direct owner
instruction.

## Follow-ups (not fixed here, deliberately deferred)

- `decommission.rs`'s owned/non-worktree removal branch still has no
  `inspect_dirt` gate — the sibling worktree branch got one from #4344, this
  one didn't. Not introduced or worsened by this PR (auto-prune's own
  exposure to it is now zero via `decommission_record_only`'s
  `record_only` skip, which bypasses both removal branches unconditionally);
  the gap is pre-existing since #4344 itself. Deferred here purely because
  `decommission.rs` has ~12 SLOC of headroom left under the 500-SLOC
  production cap after this PR's changes — not enough room for the ~25-SLOC
  mirror without a file split. Affects `decommission()`'s other callers
  (manual `tm session decommission`, `tm sessions prune --state stopped`).

## Test plan

- [x] `cargo test -p trusty-mpm` (full suite: 3972 lib + 1295 bin/tm + all
      integration binaries — 0 failed)
- [x] `scripts/check_line_cap.sh` — 0 violations
- [x] `scripts/check_test_pointers.sh` — 0 dangling pointers

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -51,6 +51,25 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **`tm ls` / bare `tm` picker: auto-prune dead records + attached→active→stopped
+  ordering** (owner request 2026-07-29). Previously a session flagged
+  `unresumable` (its workspace verifiably gone from disk) sat in the picker
+  forever printing `DEAD: workspace removed; use [d<N>] to remove the record`
+  — the operator had to notice the row and type the number manually. The
+  picker's shared fetch path (`fetch_live_sessions`) now auto-decommissions
+  every `unresumable` record through the existing `/decommission` route (the
+  same dirty-worktree-gated teardown path
+  [#4344](https://github.com/bobmatnyc/trusty-tools/pull/4344) hardened —
+  never a raw `fs::remove` or a hand-edited store entry) and prints a one-line
+  `tm: pruned N dead record(s) (workspace gone)` summary. A record whose
+  worktree removal was previously refused because the tree was dirty retains
+  a `workspace_path` that still exists on disk, which means it can never read
+  `unresumable == true` in the first place — such records are structurally
+  excluded from auto-prune, never touched. Separately, `sort_sessions` (shared
+  by the static table and the picker) now groups rows attached → active →
+  everything else ABOVE whichever `recent`/`alpha` secondary order was
+  requested, so an attached or active session never scrolls below a
+  merely-recent stopped one.
 - **`CLAUDE.md` named-section instruction overrides — reader** (epic
   [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183),
   [#4286](https://github.com/bobmatnyc/trusty-tools/issues/4286)): a project can
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -52,24 +52,43 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - **`tm ls` / bare `tm` picker: auto-prune dead records + attached→active→stopped
-  ordering** (owner request 2026-07-29). Previously a session flagged
-  `unresumable` (its workspace verifiably gone from disk) sat in the picker
-  forever printing `DEAD: workspace removed; use [d<N>] to remove the record`
-  — the operator had to notice the row and type the number manually. The
-  picker's shared fetch path (`fetch_live_sessions`) now auto-decommissions
-  every `unresumable` record through the existing `/decommission` route (the
-  same dirty-worktree-gated teardown path
-  [#4344](https://github.com/bobmatnyc/trusty-tools/pull/4344) hardened —
-  never a raw `fs::remove` or a hand-edited store entry) and prints a one-line
-  `tm: pruned N dead record(s) (workspace gone)` summary. A record whose
-  worktree removal was previously refused because the tree was dirty retains
-  a `workspace_path` that still exists on disk, which means it can never read
-  `unresumable == true` in the first place — such records are structurally
-  excluded from auto-prune, never touched. Separately, `sort_sessions` (shared
-  by the static table and the picker) now groups rows attached → active →
-  everything else ABOVE whichever `recent`/`alpha` secondary order was
-  requested, so an attached or active session never scrolls below a
-  merely-recent stopped one.
+  ordering** (owner request 2026-07-29; hardened per code-critic WARN on
+  PR [#4384](https://github.com/bobmatnyc/trusty-tools/pull/4384)). Previously
+  a session flagged `unresumable` (its workspace verifiably gone from disk)
+  sat in the picker forever printing `DEAD: workspace removed; use [d<N>] to
+  remove the record` — the operator had to notice the row and type the number
+  manually. The picker's shared fetch path (`fetch_live_sessions`) now
+  auto-prunes `unresumable` records, hardened three ways after adversarial
+  review:
+  - **Record-only removal.** The auto-prune sweep POSTs the existing
+    `/decommission` route with `?record_only=true`, routing to the new
+    `SessionManager::decommission_record_only` — the removal branches are
+    skipped ENTIRELY, so the sweep can never delete filesystem content even
+    if a remounted volume brought a workspace back between the listing-time
+    probe and the decommission call. Manual `[d<N>]` keeps today's full
+    teardown behavior.
+  - **Two-sightings confirmation + a per-call cap.** A record is only
+    decommissioned once its `unresumable` flag has been observed on TWO
+    separate listings (tracked via a small persisted marker file under
+    `~/.trusty-mpm/`), and at most 5 confirmed records are decommissioned per
+    call — the rest are reported as "N more dead records pending
+    confirmation." A bad-mount day can no longer mass-tombstone an entire
+    fleet in one `tm ls`.
+  - **TTY gate.** `guided.rs`'s bare-`tm` picker now computes its
+    interactive/non-interactive decision BEFORE fetching sessions and passes
+    it through as `fetch_live_sessions`'s new `allow_auto_prune` parameter —
+    a scripted/piped/CI invocation of bare `tm` is now a pure read, exactly
+    like `tm ls`'s existing TTY gate.
+
+  A record whose worktree removal was previously refused because the tree
+  was dirty ([#4344](https://github.com/bobmatnyc/trusty-tools/pull/4344))
+  retains a `workspace_path` that still exists on disk, which means it can
+  never read `unresumable == true` in the first place — such records are
+  structurally excluded from auto-prune, never touched. Separately,
+  `sort_sessions` (shared by the static table and the picker) now groups rows
+  attached → active → everything else ABOVE whichever `recent`/`alpha`
+  secondary order was requested, so an attached or active session never
+  scrolls below a merely-recent stopped one.
 - **`CLAUDE.md` named-section instruction overrides — reader** (epic
   [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183),
   [#4286](https://github.com/bobmatnyc/trusty-tools/issues/4286)): a project can
--- a/crates/trusty-mpm/src/bin/tm/commands/guided.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided.rs
@@ -363,11 +363,24 @@ async fn try_show_picker(
     repo_url: &str,
     cwd: &std::path::Path,
 ) -> Option<anyhow::Result<()>> {
+    // Critic HIGH finding #3 (PR #4384 review): this TTY check must be
+    // computed BEFORE the fetch below, not only at the `tty_gate` call after
+    // it — a scripted/piped/non-interactive bare `tm` invocation (cron,
+    // healthcheck, CI step) must never trigger the destructive auto-prune
+    // sweep `fetch_live_sessions` can run. Mirrors `run_ls_connector`'s own
+    // `stdin_tty && stdout_tty` gate (`session_picker.rs`).
+    let interactive = std::io::stdin().is_terminal() && std::io::stdout().is_terminal();
     // #1809: exclude decommissioned tombstones from the picker by default. The
     // shared fetch path returns the same live-only list the static renderer uses.
-    let sessions = super::session_picker::fetch_live_sessions(client, url, Some(source_id), false)
-        .await
-        .ok()?;
+    let sessions = super::session_picker::fetch_live_sessions(
+        client,
+        url,
+        Some(source_id),
+        false,
+        interactive,
+    )
+    .await
+    .ok()?;
     if !tty_gate(
         std::io::stdin().is_terminal(),
         source_id,
--- a/crates/trusty-mpm/src/bin/tm/commands/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/mod.rs
@@ -58,6 +58,7 @@ pub(crate) mod services;
 pub(crate) mod sessctl;
 pub(crate) mod session;
 pub(crate) mod session_picker;
+pub(crate) mod session_picker_prune;
 pub(crate) mod session_picker_rename;
 pub(crate) mod session_picker_render;
 pub(crate) mod slack;
--- a/crates/trusty-mpm/src/bin/tm/commands/session_picker.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session_picker.rs
@@ -426,14 +426,19 @@ fn recency_key(s: &ManagedSessionSummary) -> &str {
 /// Why: the interactive picker (both call sites) needs the deserialized, filtered
 /// session list; combining the GET and the parse keeps re-fetch-after-detach a
 /// single call and guarantees identical scoping to the static list.
-/// What: [`fetch_managed_raw`] then [`parse_scoped_sessions`], then
+/// What: [`fetch_managed_raw`] then [`parse_scoped_sessions`], then — ONLY
+/// when `allow_auto_prune` is `true` —
 /// [`super::session_picker_prune::auto_prune_dead_records`] (owner request
-/// 2026-07-29) — every session the daemon flags `unresumable` (workspace
-/// verifiably gone) is torn down automatically instead of sitting in the list
-/// forever printing "use [d<N>] to remove the record". The picker always
-/// requests live-only (`all = false`) — a decommissioned tombstone is never a
-/// resume target, and is never `unresumable` either (see that function's
-/// doc), so auto-pruning runs unconditionally here regardless of `all`.
+/// 2026-07-29): every session the daemon flags `unresumable` (workspace
+/// verifiably gone) on TWO consecutive listings is torn down automatically
+/// (capped per call), instead of sitting in the list forever printing "use
+/// [d<N>] to remove the record". `allow_auto_prune` exists because auto-prune
+/// is destructive-adjacent — every call site MUST explicitly decide (critic
+/// HIGH finding #3): pass `true` only when the listing is genuinely
+/// interactive (a real TTY), never for a scripted/piped/non-interactive
+/// invocation. When `false`, this is a pure read: the raw parsed list is
+/// returned unchanged, no marker file is touched, no HTTP call beyond the
+/// initial GET is made.
 /// Test: HTTP path in `tests/session_manager_mvp.rs`; the parse/filter seam is
 /// unit-tested via `parse_scoped_sessions`; the auto-prune seam by
 /// `auto_prune_dead_records_*` in `tests_behavior_d_tests.rs`.
@@ -442,18 +447,29 @@ pub(crate) async fn fetch_live_sessions(
     url: &str,
     source_id: Option<&str>,
     all: bool,
+    allow_auto_prune: bool,
 ) -> anyhow::Result<Vec<ManagedSessionSummary>> {
     let raw = fetch_managed_raw(client, url, source_id).await?;
     let sessions = parse_scoped_sessions(&raw, all)?;
-    let (sessions, pruned) =
-        super::session_picker_prune::auto_prune_dead_records(client, url, sessions).await;
-    if pruned > 0 {
+    if !allow_auto_prune {
+        return Ok(sessions);
+    }
+    let outcome = super::session_picker_prune::auto_prune_dead_records(client, url, sessions).await;
+    if outcome.pruned > 0 {
         eprintln!(
-            "tm: pruned {pruned} dead record{} (workspace gone)",
-            if pruned == 1 { "" } else { "s" }
+            "tm: pruned {} dead record{} (workspace gone)",
+            outcome.pruned,
+            if outcome.pruned == 1 { "" } else { "s" }
         );
     }
-    Ok(sessions)
+    if outcome.pending > 0 {
+        eprintln!(
+            "tm: {} more dead record{} pending confirmation",
+            outcome.pending,
+            if outcome.pending == 1 { "" } else { "s" }
+        );
+    }
+    Ok(outcome.kept)
 }
 
 /// Find the 0-based position in `sessions` whose stable `slot` equals `n`
@@ -1003,7 +1019,11 @@ pub(crate) async fn run_tty_picker(
         // #1809: the shared fetch path applies the same live-only tombstone filter.
         // Issue TBD: re-apply the scope's filter/sort so the redisplayed menu
         // never drifts from the one the operator picked against.
-        sessions = fetch_live_sessions(client, url, scope.source_id.as_deref(), false).await?;
+        // `allow_auto_prune = true`: this loop only ever runs once the picker
+        // is ALREADY confirmed interactive (see `should_show_picker`/`tty_gate`
+        // at the call sites that reach here) — every re-fetch stays interactive.
+        sessions =
+            fetch_live_sessions(client, url, scope.source_id.as_deref(), false, true).await?;
         sessions = filter_sessions_by_term(sessions, scope.term.as_deref());
         sort_sessions(&mut sessions, scope.sort);
     }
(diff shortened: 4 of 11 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review against the spec in the PR body. No CRITICAL findings, three HIGH findings — all about the same underlying issue: this PR converts a previously-advisory "workspace looks gone" signal into an unconditional, uncapped, unconfirmed destructive action, and the two independent guards that would normally catch a false positive both have gaps.

## Findings

| HIGH | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 216-234 | `auto_prune_dead_records` decommissions **every** `unresumable` record in a single listing, unconditionally — no cap on how many records one call can destroy, and no requirement that the flag hold across two consecutive listings before acting. A bad-mount day (one network/removable volume going away) can silently tombstone every session pointed at it in one `tm ls`. | Cap auto-prune per call (fall back to the old `[d<N>]` advisory beyond the cap), and/or require `unresumable` to be observed on two consecutive fetches before acting on it — treat a single-shot flag as a candidate, not a verdict. |
| HIGH | `crates/trusty-mpm/src/session_manager/decommission.rs` | 489-529 (cf. 414-449) | The **owned, non-worktree** removal branch has no dirty-check gate — unlike the sibling unowned-worktree branch, which #4344 hardened with `inspect_dirt`/`DirtyWorktree` (`worktree_safety.rs`). `is_unresumable` (`resume_workdir.rs:131-148`) fails open on a probe `Err`, but a *genuinely-vanished mountpoint* returns `Ok(false)` — indistinguishable from real deletion — not `Err`, so it does NOT fail open for "volume currently unmounted." If that volume remounts in the window between the listing-time probe and this function's own `ws.exists()` re-check (line 492), real, live content is `remove_dir_all`'d with zero dirty-check protection, unlike the worktree branch three lines above it. | Add the same `inspect_dirt` gate to this branch that the worktree branch already has — belt-and-suspenders, matching the codebase's own precedent one branch over. |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/guided.rs` | 300, 319, 368 | `try_show_picker` (reachable from **bare `tm`**, no subcommand) calls `fetch_live_sessions` — and therefore `auto_prune_dead_records` — **before** the `tty_gate` check at line 371. A scripted/non-interactive/piped invocation of bare `tm` (cron, healthcheck, CI step) triggers the identical unconditional destructive prune that an interactive session does. `run_ls_connector` (session_picker.rs:987) already gates on `stdin_tty && stdout_tty` before fetching for the picker — this call site doesn't. | Move (or add) the TTY gate ahead of the `fetch_live_sessions` call in `try_show_picker`, mirroring `run_ls_connector`'s existing gate. |

## Notes

- **#4344 boundary claim — verified accurate.** My first pass read a stale local worktree that predates #4344; re-verified directly against `origin/fix/tm-ls-auto-prune-and-order`. `decommission_with_root` (line 542-555) does retain `workspace_path`/`workspace_owned` whenever the path still exists on disk *after* the removal attempt — so a dirty-refused worktree, a containment-guard failure, or a never-owned path are all correctly protected, exactly as the PR body claims. The gap above is narrower than "the boundary is broken" — it's specifically the owned/non-worktree branch that has no pre-removal dirty check, and the remount-race window that neither `is_unresumable` nor that branch can see.
- **Ordering (#3485 modes) — confirmed unaffected.** `sort_sessions_recent_orders_by_last_activity`, `sort_sessions_recent_falls_back_to_created_at`, `sort_sessions_alpha_orders_by_name_case_insensitive` (`tests_behavior_d_tests.rs:993-1064`) all fix `state: "active"` / `attached: false` across every fixture row, so `group_rank` is constant within each test and the original secondary-sort assertions are exercised unmodified, not just "should still pass."
- **`attached` is not stale.** It's a fresh per-request, daemon-computed bool (`client/http_client/types.rs:530-540`) returned on every GET, not a cached client-side flag.
- **UX / TTY corruption — no issue found.** The picker (`session_picker.rs`) is a plain sequential stdin/stdout prompt loop — no crossterm, no raw mode, no alt-screen. The new `tm: pruned N dead record(s)…` line is a plain `eprintln!`, matching the existing `tm: run [d{}]…` / `tm: session […] was deleted…` lines already in the same loop.
- **Concurrent decommission of the same record — no bug found.** `decommission_managed_session` re-fetches by id on every call; a second call against an already-`Decommissioned` record re-applies the (now no-op) tombstone idempotently and returns 200, not a 404 or an error. Safe, just not a 404.
- **`closingIssuesReferences`**: empty, confirmed via `gh pr view --json closingIssuesReferences`.
- **CI at review time**: 22/23 checks green (Clippy, Format, 500-line cap, MSRV, all UI crates, etc.); the full-workspace `Test` job was still running when this review was posted. The PR body states it locally as 3972 lib + 1295 bin/tm tests, 0 failed — not independently re-verified against the live CI run at the time of this comment.

## Required Changes

1. Cap or double-confirm `auto_prune_dead_records` before it decommissions (`session_picker_prune.rs`).
2. Add a dirty-check gate to the owned/non-worktree removal branch in `decommission.rs`, mirroring the existing worktree-branch precedent.
3. Gate `fetch_live_sessions` in `try_show_picker` behind the same TTY check `run_ls_connector` already applies, so a non-interactive `tm` invocation cannot trigger a destructive auto-prune.

bobmatnyc:
CI update: the `Test` job finished **red**, not pending as noted in my review above.

Failure: `analyze_populates_complexity_and_findings` in `crates/trusty-review/tests/report_analyze_e2e.rs:191` — panics with "green-mapped diagnostic should be dropped" (https://github.com/bobmatnyc/trusty-tools/actions/runs/30513028434/job/90776937968).

This PR's diff touches only `crates/trusty-mpm/**` (`gh pr diff 4384 --name-only` confirms zero `trusty-review` files changed) — the failure is in an unrelated crate, not something this PR's auto-prune/ordering changes could have caused. Likely a pre-existing flake or an unrelated regression already on `main`. Doesn't change the WARN verdict above, but this PR is not currently mergeable on green CI — worth a quick check that `main` itself is clean on this test before assuming it's transient.

bobmatnyc:
Addressed all three HIGH findings, pushed to `97b50d52`.

**1. Record-only removal.** New `SessionManager::decommission_record_only` skips BOTH removal branches unconditionally (`decommission.rs`'s `record_only` param, threaded through `decommission_with_root`). `session_picker_prune.rs`'s `decommission_dead_record` now POSTs `/decommission?record_only=true` instead of the plain route. A remount between the listing-time probe and the call can no longer lose data — nothing is ever deleted by this path. Manual `[d<N>]` is unaffected (still calls the plain `decommission()`). Test: `decommission_record_only_never_removes_existing_workspace` (`session_manager::tests`) — a real, populated workspace directory survives the call untouched, state still flips to `Decommissioned`.

**2. Confirmation + cap.** `auto_prune_dead_records_at` now tracks first-sighting timestamps in a small persisted marker file (`~/.trusty-mpm/auto-prune-seen.json`), only acting on a record once its `unresumable` flag has been observed on a strictly earlier call. A cap of 5 confirmed-eligible records per call is enforced; anything over the cap (or not yet confirmed) is reported via a new `tm: N more dead records pending confirmation` line. Tests: `auto_prune_dead_records_first_sighting_is_not_pruned`, `auto_prune_dead_records_removes_confirmed_unresumable_records`, `auto_prune_dead_records_honors_the_cap` (seeds 6 dead sessions, confirms only 5 decommission in one call).

**3. TTY gate.** `guided.rs::try_show_picker` now computes `stdin_tty && stdout_tty` BEFORE calling `fetch_live_sessions`, threading it through as the new `allow_auto_prune` parameter — mirrors `run_ls_connector`'s existing gate exactly. Every call site of `fetch_live_sessions` now makes this decision explicitly (compiler-enforced — the parameter isn't optional). Test: `fetch_live_sessions_skips_auto_prune_when_disallowed`.

**#4344 dirty-gate mirror (owned/non-worktree branch) — deferred, not fixed here.** I measured `decommission.rs`'s actual SLOC (after comment-stripping, the metric `scripts/check_line_cap.sh` enforces) before and after this fix: 477 → 488 out of the 500-SLOC production cap, leaving only ~12 SLOC of headroom. The dirty-gate mirror (`inspect_dirt` + `spawn_blocking` + the warn/skip branch) costs roughly 25 SLOC in the sibling worktree branch — it doesn't fit without pushing the file over cap or splitting it into a sibling module mid-function, which felt like the wrong tradeoff to make silently. Separately: since finding #1 makes auto-prune record-only, the auto-prune sweep's OWN exposure to that specific branch is now zero (record_only skips it entirely) — the remaining value of the mirror is hardening `decommission()`'s OTHER callers (manual `tm session decommission`, `tm sessions prune --state stopped`), which is a real but independent hardening. Happy to file that as a follow-up issue, or take it in a split PR if you'd rather not track it separately.

Full gate, re-run clean: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p trusty-mpm` (3973 lib + 1298 bin/tm, 0 failed), `scripts/check_line_cap.sh` (0 violations), `scripts/check_test_pointers.sh` (0 dangling pointers).

bobmatnyc:
## Verdict: BLOCK (supersedes the earlier WARN)

Re-reviewed the delta at 97b50d52. All three prior HIGH findings are genuinely fixed **for a client and daemon running the same build**. But the fix for HIGH #1 (record-only removal) introduces a new gap that fails under a condition this project's own operational model treats as routine: a long-lived daemon process that hasn't been restarted since the `tm` CLI binary was last rebuilt/installed. That gap is CRITICAL, not HIGH — it silently reinstates exactly the destructive behavior this round of hardening was built to eliminate.

## Findings

| CRITICAL | `daemon/managed_routes/mod.rs` (route) + `bin/tm/commands/session_picker_prune.rs:258-278` (client) | 897-901, 258 | `?record_only=true` is read by a `Query<DecommissionQuery>` extractor this commit **adds** to the handler signature. Axum ignores query params a handler doesn't explicitly extract — a daemon process still running a build older than 97b50d52 (no `Query<DecommissionQuery>` in its handler) silently falls through to the OLD unconditional `mgr.decommission(&id, None)` (full teardown) for every auto-prune call, regardless of the query string sent. `decommission_dead_record` only checks `resp.status().is_success()` — never the response body's `workspace_removed` field (which the daemon already returns, `managed_routes/mod.rs:283`, and would reveal the truth). Given this architecture's daemon is long-lived by design (a CLI upgrade doesn't bounce it) and the coordinator confirms this exact skew is live on this machine, the "auto-prune can never touch the filesystem" guarantee from the previous round does not hold today — silently. | Have `decommission_dead_record` parse the JSON body and require `workspace_removed == false` before counting the call as a safe prune. Treat `workspace_removed == true` (or an unparseable body — also a stale-daemon signal) as a hard stop: abort the rest of this sweep, print `"tm: auto-prune: daemon may be running an older build; skipping automatic pruning until it's restarted"`, and leave every remaining dead record on the manual `[d<N>]` path. (A daemon-version handshake would close this more thoroughly but isn't required — the response-body check alone is sufficient and needs no protocol change.) |

## Verified fixed this round (re-confirmed against the actual diff, not the commit message's claims)

- **HIGH #1 (uncapped/unconfirmed auto-prune)** — fixed for same-version client/daemon. Two-sightings confirmation (`auto_prune_dead_records_at`, `session_picker_prune.rs:403-477`) + `AUTO_PRUNE_CAP = 5` are both exercised by tests that drive the actual partition logic across two sequential calls (`auto_prune_dead_records_first_sighting_is_not_pruned`, `auto_prune_dead_records_removes_confirmed_unresumable_records`, `auto_prune_dead_records_honors_the_cap`), not just asserted as constants.
- **HIGH #2 (decommission.rs owned-branch, no dirty gate / remount race)** — fixed for auto-prune's own exposure. `decommission_record_only` (`decommission.rs:930-957`) skips both removal branches unconditionally (`if !record_only && let Some(ref ws) = ...`, line 988) — confirmed by `decommission_record_only_never_removes_existing_workspace`, which populates a real directory with real content and proves it survives the call untouched, including the sentinel file. The plain (non-record-only) route is byte-for-byte unchanged (`decommission_with_root(..., false)` at the pre-existing call site, line 927) — manual `[d<N>]` and every other caller keep today's behavior exactly.
- **HIGH #3 (bare `tm` fetching before the TTY gate)** — fixed. `guided.rs:123` computes `interactive = stdin.is_terminal() && stdout.is_terminal()` before the fetch and threads it through as `allow_auto_prune`. `fetch_live_sessions_skips_auto_prune_when_disallowed` proves a non-interactive call is a pure read by asserting the **daemon-side** record state stays `errored` (not decommissioned) — not merely that the client's returned list is unpruned, which would leave open a "fetched-and-pruned-but-happened-to-return-the-full-list" false pass.

## Deferral: #4344 dirty-gate mirror for `decommission.rs`'s owned/non-worktree branch

**Accepted.** The reasoning holds: `decommission_record_only`'s `if !record_only && ...` guard means the auto-prune sweep structurally never reaches that branch anymore, and the plain `decommission()` path (where the gap actually lives) is unchanged from before PR #4384 — no new caller is exposed to the missing dirty-check by this delta. This was a pre-existing gap since #4344 itself (which only hardened the sibling worktree branch), not something #4384 introduced or worsened, so deferring it out of scope here is legitimate, not evasion.

**Where the record lives:** right now, only in the commit message (97b50d52) and this delta's doc-comment/CHANGELOG prose. I searched for a tracking issue (`dirty-gate mirror decommission`, `owned non-worktree dirty`, `record_only`) and found none. Given this project's "write tickets sparsely" convention I'm not blocking on the missing issue, but a commit message isn't discoverable the way an open issue is — worth a one-line issue if this follow-up is expected to survive past this PR's lifetime.

## Notes

- **MEDIUM, not blocking:** the confirmation marker file (`~/.trusty-mpm/auto-prune-seen.json`) has no locking or atomic-write protection — two concurrent `tm` invocations can race on read-modify-write. Consequence is bounded: worst case is a duplicate `decommission_record_only` call for the same id (harmless — record-only plus the pre-existing idempotent-decommission behavior absorb it) or the per-call cap being exceeded in aggregate across two simultaneous callers (a soft invariant now, not a filesystem-safety one, given the CRITICAL finding above is fixed). Corrupt/unreadable marker file already fails safe (`load_seen` treats it as empty — i.e. "nothing confirmed yet," never "everything confirmed").
- CI: the `Test` job failure I flagged in my previous comment is already tracked and closed as a flake at #4385 (`analyze_populates_complexity_and_findings` — raw `TcpListener` mock races on CI's per-crate rerun, unrelated to this PR).
- I have not independently confirmed the actual running daemon binary's version on this machine — I'm relying on the coordinator's statement that skew is real, plus the code-level proof that *if* skew exists, the safety property silently fails. The vulnerability itself is verified directly from the diff and does not depend on that external claim.

## Required Changes

1. CRITICAL: verify `workspace_removed == false` in the `/decommission?record_only=true` response body before treating an auto-prune call as safe; degrade to skip-and-advise on anything else.
2. (Optional, not blocking) File a tracking issue for the deferred #4344 dirty-gate mirror so it isn't lost to a commit message.

bobmatnyc:
CRITICAL addressed, pushed to `5a39f8c5`.

**Response-body verification.** `decommission_dead_record` now parses the `/decommission?record_only=true` response body and requires `workspace_removed == false` before counting the call as a safe prune. Anything else — `true`, a missing field, or a body that fails to parse — is treated as proof (or inability to disprove) that a stale daemon executed the old destructive path. On that signal, the sweep:
- logs `tm: auto-prune: daemon may be running an older build; skipping automatic pruning until it's restarted`
- stops attempting further decommissions for the rest of this call, and every remaining dead record falls back to the manual `[d<N>]` presentation
- persists a `__stale_daemon_detected__` sentinel in the SAME marker file used for two-sightings confirmation, so the disabled state survives across every later call within this invocation (the picker's re-fetch loop calls this repeatedly) without needing process-global state — which also keeps it test-isolated, since `cargo test` runs many tests sharing one process.

**Test**: `auto_prune_dead_records_stops_sweep_when_daemon_reports_workspace_removed` spins up a stub axum server that mimics a pre-#4384 daemon exactly — no `Query<DecommissionQuery>` extractor, ignores every query param, always answers `{"workspace_removed": true}` — and drives two already-confirmed dead records through it in one call. Asserts `pruned == 0` and both records survive in `kept`, proving the sweep stops after the first stale response rather than attempting (and silently trusting) the second.

**Follow-up tracking**: added a "Follow-ups" section to the PR body recording the deferred #4344 dirty-gate mirror (SLOC budget, `decommission.rs`'s owned/non-worktree branch), so it survives the squash per your note — no new issue, per the sparse-ticket convention.

Full gate, re-run clean: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p trusty-mpm` (3973 lib + 1299 bin/tm, 0 failed), `scripts/check_line_cap.sh` (0 violations), `scripts/check_test_pointers.sh` (0 dangling pointers).

bobmatnyc:
## Verdict: WARN (round 3 — CRITICAL from round 2 is fixed; one HIGH remains, recordable not blocking)

Re-verified the delta at 5a39f8c5 against both specific questions.

## Parse-failure coverage — confirmed complete

`decommission_dead_record`'s new body-check (`session_picker_prune.rs:265-271`):

```rust
match resp.json::<serde_json::Value>().await {
    Ok(body) => match body.get("workspace_removed").and_then(|v| v.as_bool()) {
        Some(false) => DecommissionOutcome::Pruned,
        _ => DecommissionOutcome::StaleDaemon,
    },
    Err(_) => DecommissionOutcome::StaleDaemon,
}
```

Traced every case: `workspace_removed: true` → `StaleDaemon` (correct). Field absent → `None` → `_` arm → `StaleDaemon` (correct). Field present but non-bool (e.g. a string) → `.as_bool()` returns `None` → `_` arm → `StaleDaemon` (correct, fails closed on a shape it doesn't recognize). Non-JSON body → `serde_json` parse `Err` → `StaleDaemon` (correct). Empty body → also an `Err` from `.json()` → `StaleDaemon` (correct). A body-read I/O error mid-stream after a 2xx status also lands in `Err(_) => StaleDaemon` — semantically it's not really "a stale daemon," but the consequence (stop trusting this call, fall back to manual) is the same safe direction, so this is a non-issue, not a finding. Every failure-to-prove-safety path converges on "assume unsafe." Good.

## Sentinel persistence — NOT self-healing (HIGH, recordable not blocking)

Grepped every reference to `STALE_DAEMON_MARKER_KEY` in the file: it is **inserted** (`session_picker_prune.rs:260`, on `newly_stale_daemon`) but **never removed** anywhere. The only `.remove()` calls in the function target `&s.id` (real session ids) — the healthy-session-clears-its-own-marker loop (`for s in &kept { seen.remove(&s.id) }`) can never match `"__stale_daemon_detected__"` since that string is never a session id. Once written to `~/.trusty-mpm/auto-prune-seen.json`, this key survives forever: every future `tm`/`tm ls` invocation — including ones that run against a daemon that has since been rebuilt/restarted and genuinely supports `record_only` — loads the marker file, sees `already_stale_daemon == true` unconditionally, and permanently skips auto-prune while re-printing "daemon may be running an older build; skipping automatic pruning until it's restarted" (`session_picker_prune.rs:264-267`) even after that statement is no longer true. Recovery requires the operator to manually find and delete a file they have no reason to know exists.

This is **not** a safety regression — it fails in the safe direction (permanently over-cautious, never destructive), and core `tm`/`tm ls` session listing is completely unaffected; only the auto-prune convenience feature silently and permanently reverts to manual `[d<N>]`. That keeps it out of CRITICAL. But it's a real, verifiable "the feature stops working forever after one bad sighting, with no recovery and an increasingly false warning message" correctness bug — HIGH, not a style nit.

Given round 3's fold-not-block doctrine and the fact that nothing here risks data or safety, I'm marking this **recordable, not blocking**: land it, then fix in a fast follow (e.g. clear the sentinel on the first response that DOES report `workspace_removed == false` again, or attach a TTL to the stored timestamp and re-probe after it expires). Please fold this into the same PR-body Follow-ups section that already tracks the deferred #4344 dirty-gate mirror, so it isn't lost to a commit message either.

## Secondary, minor: `pending` undercounts the stale-daemon-blocked batch (LOW, recordable)

`pending = confirmed.len() + first_sighting.len()` is computed from what's left in `confirmed` **after** `to_prune` was already drained out (`let to_prune: Vec<_> = confirmed.drain(..cap).collect();`, earlier in the function). Whatever happens to `to_prune` — pruned, failed, or pushed back to `kept` via either stale-daemon path — is never folded back into `pending`. Concretely: in the exact scenario the new test exercises (both confirmed ids caught by a stale daemon), `outcome.pending == 0` even though 2 records are being held back — `fetch_live_sessions` prints the "daemon may be older" line (so the operator isn't left with zero signal) but the "N more dead record(s) pending confirmation" line simply never fires for that batch. Records are correctly preserved in `kept` (nothing is lost or mis-pruned) — this is a reporting/accounting inaccuracy only. Not caught by the new test because it doesn't assert on `pending`. Not blocking; worth a one-line fix (fold the `to_prune`-derived kept-but-not-pruned count into `pending`) whenever the sentinel follow-up above gets picked up.

## Everything from rounds 1–2 — still holds

Re-confirmed the round-2 CRITICAL fix itself is correct (that was the point of this delta) and none of the three original HIGH findings regressed. Not re-verifying line-by-line again here since nothing in this delta touched those paths.

## Required Changes

None blocking. For the follow-up (already-accepted, PR-body-tracked deferral list):
1. Give `STALE_DAEMON_MARKER_KEY` a way to clear — either on the next observed `workspace_removed == false`, or via a TTL on the stored timestamp.
2. Fold the stale-daemon-blocked `to_prune` remainder into `outcome.pending` so the operator-facing count is accurate.

## Notes

Given no CRITICAL remains and the one HIGH is explicitly fail-safe (not fail-destructive) and explicitly recordable per the fold-not-block instruction for round 3, I'm not asking for a round 4. This PR is fine to merge on green CI with these two items tracked as follow-ups.

bobmatnyc:
Implemented both recommended follow-ups directly, pushed to \`c97e1b69\`.

1. **Sentinel TTL (1 hour).** \`stale_daemon_sentinel_active\` now checks the persisted timestamp, not just presence — a fresh sentinel still blocks; an expired one lets the next call retry, and if the daemon is still stale the body-check trips again immediately and rewrites a fresh sentinel (one probe/hour, self-heals on restart). Expired entries are cleared from the marker file.
2. **Pending fold-in.** Records held back by the stale-daemon gate (already-active or newly-tripped mid-loop) now count into \`AutoPruneOutcome::pending\`.

New test \`auto_prune_dead_records_stale_daemon_sentinel_expires_after_ttl\` drives a stub daemon through both halves — fresh sentinel blocks, expired one retries and re-trips a freshly-timestamped sentinel (verified against the persisted marker file, not just behavior).

Gate: \`cargo fmt --check\`, \`cargo clippy --all-targets -- -D warnings\`, full \`cargo test -p trusty-mpm\` (3973 lib + 1300 bin/tm, 0 failed), \`scripts/check_line_cap.sh\` (0 violations), \`scripts/check_test_pointers.sh\` (0 dangling pointers) — all clean.

Head SHA: \`c97e1b69b95b9153b0c5f41d2d7b3f55c37cc420\`. Not merging — routing to the coordinator per instruction.

PR #4316, opened 2026-07-29
docs(trusty-mpm): add a plain-speaking prose rule to the PM instructions

bobmatnyc:
## Problem

PM session prose is hard to read. Long comma-chained sentences, throat-clearing
openers before the actual fact, closing aphorisms that restate what was just
said, and meta-commentary about the PM's own reasoning.

## Change

Adds a **Prose Style — Write Plainly** subsection under `## Response Format` in
`crates/trusty-mpm/src/assets/instructions/sections/core.md`:

- Lead with the point: what happened, then why it matters.
- Short sentences, one idea each. Split anything carrying three commas and a dash.
- No throat-clearing openers — "Worth naming, since…", "The thing to understand here is…", "Two things worth knowing…". State the fact.
- No closing aphorisms. Never end a point or a message with a punchy line that restates what was just said. Stop at the last useful sentence.
- No meta-commentary about your own reasoning, rules, or process unless it changes what the reader should do.
- Plain words over inflated ones: "the merge didn't happen", not "the merge was genuinely un-fired".
- Tables and short bullets for status, not paragraphs.

**Prose only.** The rule governs how something is said, never whether it is
said. Failures, corrections, and bad news are still reported directly and in
full — it shortens the wording, never the disclosure. Without that caveat the
rule could be read as license to soften bad news.

## Why this file

- `sections/core.md` is composed into the PM system prompt for **every** session
  via `bundled_pm_package::bundled_fallback_package()`. The `## Communication`
  tone rules in `assets/output-styles/*.md` are triplicated across three styles
  and only the active one applies; one of them (`-teacher`) deliberately
  narrates its own reasoning, which conflicts with the meta-commentary bullet.
- It extends the existing section that already governs what the PM says to the
  user, rather than inventing a competing top-level `## Communication` heading.
- `core` is `customization_tier: project`, **not** the non-overridable BASE_PM
  floor. This is a style preference, not a safety invariant, so a project may
  still override it.

## Gates

- `cargo test -p trusty-mpm` (full surface, not `--lib`) — **0 failures**;
  3904 lib tests plus every integration binary green.
- `cargo fmt --check` — clean.
- `scripts/check_instruction_floor.sh` — `PASS: instruction floor is intact and
  carries the guaranteed conventions`.
- `scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK`.

No check was disabled or weakened. The two composed-prompt goldens
[...]

Issue #381, opened 2026-05-29
fix(trusty-mpm): project-level instruction override system is advertised but never read


bobmatnyc:
## Summary
BASE_PM.md tells the PM it can create `.trusty-mpm/INSTRUCTIONS.md`, `AGENT_DELEGATION.md`, `WORKFLOW.md`, `MEMORY.md`, `PM_INSTRUCTIONS_DEPLOYED.md` and that they take effect next startup. No Rust code reads any of them. System prompt is a compile-time concat in `src/core/instruction_pipeline.rs:51`. This is a silent no-op — actively misleading documentation.

## Current Behavior
- PM creates override files per BASE_PM.md guidance
- No change to actual system prompt sent to Claude
- PM believes overrides are active but they have zero effect
- Session instructions stash continues to show original system prompt

## Desired Behavior / Acceptance Criteria
**DECISION REQUIRED** — capture BOTH options in the issue:

**Option A (Implement):**
- Read the 5 override files at session prepare time
- Layer them per documented append/replace semantics
- Validate integrity; warn on missing files
- Update system prompt stash to reflect actual assembled prompt

**Option B (Strip):**
- Remove false promises from BASE_PM.md
- Document why override system is not implemented
- Update to match actual behavior

**NOTE**: The PRD lists the override system as INTENDED scope, so Option A is the long-term-aligned choice; Option B is the immediate-honesty fix.

## Spec Reference
- architecture-spec §5 (Instruction Assembly)
- architecture-spec §12 (Gaps & Deviations)

Issue #4286, opened 2026-07-29
Retire the project-level .trusty-mpm/ PM instruction override mechanism


bobmatnyc:
Supersedes #4267. The project-level override file is obsolete and retires; #4267's three options (floor the content / give it its own non-overridable section / document an opt-out) are moot.

## Defect

Section-granularity overrides let a project silently delete owner-required framework content. A project with `.trusty-mpm/WORKFLOW.md` receives ZERO occurrences of the "Sprint, then Harden" doctrine — that string exists only in `sections/workflow.md` (3 hits; 0 in identity/core/memory/search/agent-delegation/non-overridable-rules/framework-guaranteed-conventions).

`AGENT_DELEGATION.md` is strictly worse: it replaces the section AND suppresses the live agent roster entirely (`crates/trusty-mpm/src/core/instruction_overrides.rs:263-268`, gate test at `:606-632`) — reintroducing the #4069 defect in full. `PM_INSTRUCTIONS_DEPLOYED.md` replaces the entire PM body including memory and search doctrine (`instruction_overrides.rs:248-255`).

## The override surface (5 files, all `<project_root>/.trusty-mpm/`)

| File | Precedence | Semantics | Read at |
|---|---|---|---|
| `PM_INSTRUCTIONS_DEPLOYED.md` | highest, short-circuits | REPLACES entire PM body; floor still appended | `instruction_overrides.rs:248-255` |
| `AGENT_DELEGATION.md` | section | REPLACES section + suppresses live roster | `:259`, `:263-268` |
| `WORKFLOW.md` | section | REPLACES whole workflow section | `:258`, applied `:316` |
| `MEMORY.md` | section, additive-shaped | INSERTED as delimited block, does not excise bundled section | `:260`, `:344-348` |
| `INSTRUCTIONS.md` | lowest | APPENDS after all sections, before floor | `:261`, `:352-356` |

Constants at `instruction_overrides.rs:31-42`. Refuted: project-level `PM_INSTRUCTIONS.md` and `BASE_PM.md` are NOT read.

## Evidence — the framework is the producer

tm's binary never seeds these files: the `bundle_all.rs:101+` `ALL` table contains none of the five filenames, and every `rel_path` targets `~/.trusty-mpm/framework/` (`bin/tm/commands/install.rs:549`). There is no `tm init` subcommand.

[...]

PR #4386, opened 2026-07-30
feat(trusty-mpm): JSON instruction manifest (schema v2) with owner language rules

### LINKED ISSUE #4318: feat(trusty-mpm): author the bundled PM instruction package as JSON, not a Rust literal
Part of #4183.

## Gap

The instruction package is a Rust literal, not an authored artifact. `crates/trusty-mpm/src/core/bundled_pm_package.rs:186` constructs `InstructionPackage` in code — sections, tiers, block order, joins, generators — over eight `include_str!` markdown sources (`core/instruction_pipeline.rs:58-84`). The JSON reader exists and is dead: `InstructionPackage::from_json` (`core/instruction_package.rs:577`) has zero non-test call sites, and `assets/instructions/instruction-package.schema.json` is a type contract with no instance under it.

#4286 asserts as settled fact that "the composed instructions payload is generated from the instruction package JSON". It is not. #4186 tracked exactly this swap and was closed NOT_PLANNED on 2026-07-28.

## Constraint

The schema carries prose inline — `$defs.body` is `{"kind":"text","text":"…"}` or `{"kind":"generated","generator":…}` (`instruction-package.schema.json:78-99`). There is no file-reference body kind, and `additionalProperties:false` applies at every level, so adding one bumps `schema_version` under the evolution policy at `:11`. Inlining the prose verbatim turns `sections/core.md` into a single 24,206-character JSON line (23,301 raw bytes, 455 lines) and moves the floor text out of the files `scripts/check_instruction_floor.sh:40-44` stats and greps.

`sections/*.md` also have three consumers today, not one: the package (`bundled_pm_package.rs:191-231`), the legacy override assembly (`instruction_overrides.rs:343,359` via `instruction_pipeline.rs:102,130`), and `assemble_system_prompt()` (`instruction_pipeline.rs:152`), which `tm install` and `tm launch` write to `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` and which supplies no agent roster to satisfy the composer's required `agent-roster` generator.

## Resolution

Land after #4286's read-path removal, when the legacy-assembly consumer is gone and only one golden remains.

Add a `{"kind":"file","path":"sections/<id>.md"}` body variant at `schema_version` 2, resolved through an `include_str!` table so the build stays hermetic and a missing section stays a compile error. Prose keeps living in reviewable markdown; the JSON carries the manifest — sections, tiers, block order, joins, generators, and `ROSTER_PRECEDENCE_NOTE` (`instruction_overrides.rs:392`, prose currently stranded in Rust). Inline-only text remains expressible, so a package that wants it still works.

`bundled_fallback_package()` then becomes `from_json(include_str!(…))` and returns `Result`; on parse failure compose from the retained Rust constants rather than ship a truncated prompt. Goldens must not move — this is mechanism, not content — and `package_round_trips_through_json` (`bundled_pm_package_tests.rs:160`) already proves the instance can be generated from the current builder instead of hand-typed.

bobmatnyc:
Part of [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183). Closes [#4318](https://github.com/bobmatnyc/trusty-tools/issues/4318).

Pulled forward in the 1.3.0 sequence by owner order: the 1.3 rules belong in JSON, not markdown. So the conversion lands first, and the three new rules land *in the converted artifact*.

## What changed

The bundled PM instruction package was a Rust literal. `InstructionPackage::from_json` had zero non-test call sites and the shipped JSON Schema had no instance under it, while other tickets already recorded "the composed instructions payload is generated from the instruction package JSON" as settled fact. It now is: `assets/instructions/pm-instruction-package.json` is the authored manifest, and `bundled_pm_package.rs` only parses and composes it.

**Schema v2** adds a `{"kind":"file","path":"…"}` body variant. Prose keeps living in reviewable markdown under `sections/`; the manifest names it by path and resolves through a compile-time `include_str!` table (`instruction_pipeline::SECTION_SOURCES`). Inlining the prose instead would have turned `core.md` into one 23 KB JSON line and moved the floor text out of the files `check_instruction_floor.sh` greps; resolving on the filesystem at launch would make the delivered prompt depend on what happens to be on disk. This way the build stays hermetic and a renamed section is a compile error. The bump is required by the format's own evolution policy — a v1 build reading a v2 manifest rejects it loudly rather than composing a prompt missing the new field's effect.

## The mechanism moves nothing

All three composed-prompt goldens passed **byte-unchanged** with the manifest wired in and the content below absent. The golden diff in this PR is content only:

```
 pm-prompt-bundled-fallback.md          | 25 ++++++++++++++++++++++
 pm-prompt-claude-md-override.md        | 19 ++++++++++++++++
 pm-prompt-legacy-delegation-override.md| 25 ++++++++++++++++++++++
 3 files changed, 69 insertions(+)
```

Zero deletions. Nothing moved.

## The three owner rules

Authored as inline `text` blocks in the manifest, not in markdown. Same family as the plain-prose rule from [#4316](https://github.com/bobmatnyc/trusty-tools/pull/4316).

| rule | section | substance |
|---|---|---|
| `### Clickable References` | `core` | issue/PR/ticket/commit references always render as clickable markdown links, never bare numbers |
| `### Banned Word — "honest"` | `core` | banned from PM responses, delegation briefs and review instructions; a report states facts, and labelling them honest implies the alternative was considered |
| `## Opportunistic Fixes` | `workflow` | an easy fix found while working on a file is noted on the CURRENT issue and made in the same work — never a new issue; new issues are for genuinely separable schedulable work |

## Provenance

Raised in review: the banned-word rule and the rules-in-JSON ruling arrived without a citation. The owner declined issue comments for them ("don't ticket it, just do it"), so the record lives here. Two of the four items *do* have posted owner comments, and those are cited rather than restated as verbal — the distinction is the point, since it marks exactly which items rest on relayed authority.

| item | date | source |
|---|---|---|
| Opportunistic-fix rule | 2026-07-29 | posted owner comment — [#4183 comment-5123027792](https://github.com/bobmatnyc/trusty-tools/issues/4183#issuecomment-5123027792) |
| Clickable-links rule | 2026-07-29 | posted owner comment — [#4183 comment-5124050901](https://github.com/bobmatnyc/trusty-tools/issues/4183#issuecomment-5124050901) |
| Banned word "honest" | 2026-07-29 | verbal owner order, session `tm-dogfood-relaunch-01`, relayed via the PM. No issue comment by owner instruction. |
| Rules-in-JSON-not-MD ruling (why [#4318](https://github.com/bobmatnyc/trusty-tools/issues/4318) was pulled forward ahead of the 1.3.0 sequence, and why the three rules are inline `text` blocks rather than new markdown files) | 2026-07-30 | verbal owner order, session `tm-dogfood-relaunch-01`, relayed via the PM. No issue comment by owner instruction. |

## No split-brain

Once a rule may be authored in the manifest, rebuilding the legacy multi-section strings from the `include_str!` constants would deliver it to the packaged composer and silently withhold it from the `.trusty-mpm/` override assembly and from `assemble_system_prompt()` — the roster-free prompt `tm install` writes and `tm launch` hands to `--append-system-prompt-file`, which has no agent roster and so cannot use the package composer at all.

So `pm_instructions()`, `base_pm()` and the new `workflow_section()` / `delegation_doctrine()` project the same manifest through `InstructionPackage::authored_run`. The goldens are the evidence:

| golden | delta | why |
|---|---|---|
| bundled-fallback | +1402 bytes, all 3 rules | the packaged composer |
| legacy-delegation-override | +1402 bytes, all 3 rules | the legacy assembly agrees — no split-brain |
| claude-md-override | +1067 bytes, 2 rules | correct: the `CLAUDE.md` WORKFLOW override replaces that section wholesale, so the workflow rule is *supposed* to be gone |

## Also in scope

- `ROSTER_PRECEDENCE_NOTE` moves out of Rust into the manifest, where its position between the doctrine and the live roster is declared rather than formatted in by hand at two call sites.
- `with_overrides` now matches authored bodies rather than `Text` alone. Without it, every bundled section becoming a `file` body would have made the whole taxonomy silently non-overridable — an advertised-but-unread override, [#381](https://github.com/bobmatnyc/trusty-tools/issues/381) verbatim. Pinned by an added assertion in `content_sections_accept_a_project_override`.
- `bundled_fallback_package()` returns `Result`. On a manifest failure `resolve_pm_prompt` logs loudly and degrades to the legacy assembly built from the retained constants, so the worst case is a prompt missing the inline rules rather than a truncated one. `bundled_manifest_parses_and_validates` makes that unreachable for a shipped build.
- `instruction_pipeline.rs` tests move to a sibling `instruction_pipeline_tests.rs` to stay under the 500-SLOC production cap.

New tests: `bundled_manifest_parses_and_validates`, `manifest_prose_lives_in_markdown_not_in_the_json`, `every_section_source_resolves`, `unknown_file_source_is_rejected`, `schema_body_kinds_match_rust_variants` (the v2 schema↔Rust drift gate), `file_body_resolves_through_the_bundled_table`, `a_file_block_may_not_be_optional`, `authored_run_projects_blocks_in_order_with_joins`, `authored_run_skips_generated_blocks`, `workflow_section_carries_the_opportunistic_fix_rule`, `delegation_doctrine_carries_the_precedence_note`, `the_installed_prompt_carries_the_manifest_authored_rules`.

## Coexistence with #4316 — merged and verified

[#4316](https://github.com/bobmatnyc/trusty-tools/pull/4316) merged at 2026-07-30T05:38Z and is folded into this branch (`cc4dd070`). The merge precondition is satisfied:

- **Golden diff vs the pre-fold commit: +20 / −0 on each of the three goldens. Zero deletions.** The 20 lines are exactly #4316's `### Prose Style — Write Plainly` section.
- Ordering is correct: the prose section lands immediately before `### Clickable References`, both under `## Response Format` — this PR's rules are manifest blocks joined after the whole of `core.md`, so they compose after it automatically.
- `sections/core.md` needed no source resolution; only the generated goldens and `CHANGELOG.md` overlapped. Both CHANGELOG entries are kept.
- Rule coverage in the delivered prompt:

  | golden | rules | reading |
  |---|---|---|
  | bundled-fallback | 4/4 | packaged composer |
  | legacy-delegation-override | 4/4 | legacy assembly agrees — no split-brain |
  | claude-md-override | 3/4 | workflow rule correctly absent; that configuration replaces the WORKFLOW section wholesale |

Re-verified on the folded head: golden test `4 passed; 0 failed`, `cargo test -p trusty-mpm --lib` `3985 passed; 0 failed`, `cargo clippy … -D warnings` exit 0, `cargo fmt --all --check` exit 0, line-cap OK, instruction floor PASS.

## Gates

| gate | result |
|---|---|
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | exit 0, 0 errors |
| `cargo test -p trusty-mpm` | exit 0 — `3985 passed; 0 failed; 6 ignored` (lib) plus 43 further targets, 0 `FAILED` |
| `scripts/check_line_cap.sh` | `7 allowlisted, 0 violations — OK` |
| `scripts/check_instruction_floor.sh` | `PASS: instruction floor is intact` |
| `scripts/check_test_pointers.sh` | `0 dangling pointers — OK` |

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -106,6 +106,58 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- The bundled PM instruction package is now an **authored JSON manifest**, not a
+  Rust literal
+  ([#4318](https://github.com/bobmatnyc/trusty-tools/issues/4318), part of
+  [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183)). Sections,
+  customization tiers, block order, joins, the declared generators, and the
+  roster-precedence note all live in
+  `assets/instructions/pm-instruction-package.json`;
+  `bundled_pm_package::bundled_fallback_package()` parses that file instead of
+  constructing an `InstructionPackage` in code, so there is no second place a
+  join or a block order can be edited. `InstructionPackage::from_json` had zero
+  non-test call sites before this and the shipped JSON Schema had no instance
+  under it — both are now live.
+
+  **Instruction-package schema v2**: adds a `{"kind":"file","path":"…"}` body
+  variant alongside `text` and `generated`. Prose keeps living in reviewable
+  markdown under `sections/`; the manifest references it by path and resolves it
+  through a compile-time `include_str!` table
+  (`instruction_pipeline::SECTION_SOURCES`), so the build stays hermetic, a
+  renamed section is a compile error, and `check_instruction_floor.sh` keeps
+  grepping real files. Inlining the prose instead would have turned `core.md`
+  into one 23 KB JSON line. The version bump is required by the format's own
+  evolution policy: a v1 build reading a v2 manifest rejects it loudly rather
+  than composing a prompt missing the new field's effect.
+
+  The mechanism half is byte-identical — all three composed-prompt goldens passed
+  unchanged before the content below was added.
+
+- PM instructions: three owner language rules, authored as inline `text` blocks
+  in the manifest rather than in markdown (owner order, 2026-07-29 — the 1.3
+  rules belong in JSON). Same family as the plain-prose rule from
+  [#4316](https://github.com/bobmatnyc/trusty-tools/pull/4316).
+
+  | rule | section | substance |
+  |---|---|---|
+  | Clickable References | `core` | issue/PR/ticket/commit references always render as clickable markdown links, never bare numbers |
+  | Banned Word — "honest" | `core` | "honest" and every variation is banned from PM responses, delegation briefs and review instructions; a report states facts, and labelling them honest implies the alternative was considered |
+  | Opportunistic Fixes | `workflow` | an easy fix found while working on a file is noted on the CURRENT issue and made in the same work — never a new issue; new issues are for genuinely separable schedulable work |
+
+  Delivered to every launch path, not just the packaged composer:
+  `pm_instructions()`, `base_pm()`, the new `workflow_section()` and
+  `delegation_doctrine()` now project the manifest through
+  `InstructionPackage::authored_run` instead of rebuilding from the section
+  constants, so the legacy `.trusty-mpm/` override assembly and the roster-free
+  `assemble_system_prompt()` (what `tm install` writes and `tm launch` passes to
+  `--append-system-prompt-file`) carry the same rules. Rebuilding from the
+  constants would have delivered a manifest-authored rule to one composer and
+  silently withheld it from the others. `core` and `workflow` remain
+  `customization_tier: project`, so a `CLAUDE.md` named-section override still
+  replaces them wholesale — visible in the regenerated
+  `pm-prompt-claude-md-override.md`, which correctly does NOT carry the workflow
+  rule.
+
 - `tm ls` STATE column: a row the daemon did not probe now renders
   `[assets ?]` rather than nothing, so the absence of `[stale-assets]` can
   never be misread as "assets fresh"
--- a/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
+++ b/crates/trusty-mpm/src/assets/instructions/instruction-package.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/bobmatnyc/trusty-tools/schemas/instruction-package-v1.json",
+  "$id": "https://github.com/bobmatnyc/trusty-tools/schemas/instruction-package-v2.json",
   "title": "trusty-mpm instruction package",
   "description": "A PM instruction package: a section taxonomy carrying the customization-tier axis, plus an ORDERED block stream that composes to the PM system prompt. Ordering is array order — never map iteration order — so composition is byte-deterministic (SPEC-PMINSTR-01 / DOC-59 §4a, epic #4183, schema ticket #4184).",
   "type": "object",
@@ -8,8 +8,8 @@
   "required": ["schema_version", "package_id", "sections", "blocks"],
   "properties": {
     "schema_version": {
-      "description": "Schema major version. A package whose version this build does not implement is rejected outright — never partially interpreted; the version is checked BEFORE any field, so a newer package reports a version error rather than blaming one of its own valid keys. EVOLUTION POLICY: every object here sets `additionalProperties: false`, so unknown keys are errors, not ignored. Any field addition that can change composed bytes MUST bump this version. Adding a field while leaving the version at 1 is the one case strictness cannot protect against — an older build would compose a prompt missing the new field's effect and report success.",
-      "const": 1
+      "description": "Schema major version. A package whose version this build does not implement is rejected outright — never partially interpreted; the version is checked BEFORE any field, so a newer package reports a version error rather than blaming one of its own valid keys. EVOLUTION POLICY: every object here sets `additionalProperties: false`, so unknown keys are errors, not ignored. Any field addition that can change composed bytes MUST bump this version. Adding a field while leaving the version unchanged is the one case strictness cannot protect against — an older build would compose a prompt missing the new field's effect and report success. HISTORY: v1 (#4184) shipped `text` and `generated` bodies; v2 (#4318) adds the `file` body kind so an authored JSON manifest can reference bundled markdown instead of inlining 23 KB of prose on one line.",
+      "const": 2
     },
     "package_id": {
       "description": "Stable identifier for this package (e.g. `trusty-mpm/pm`). Identity only; it never affects composed bytes.",
@@ -76,7 +76,7 @@
       ]
     },
     "body": {
-      "description": "Where a block's markdown comes from. `text` is authored in the package; `generated` is supplied at composition time by the named generator.",
+      "description": "Where a block's markdown comes from. `text` is authored inline in the package; `file` names a markdown source bundled with the build; `generated` is supplied at composition time by the named generator.",
       "oneOf": [
         {
           "type": "object",
@@ -85,6 +85,20 @@
           "properties": {
             "kind": { "const": "text" },
             "text": { "type": "string", "minLength": 1 }
+          }
+        },
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["kind", "path"],
+          "description": "Schema v2 (#4318). The path is resolved against the build's compile-time `include_str!` table (`instruction_pipeline::SECTION_SOURCES`), NOT against the filesystem at composition time — so the delivered prompt never depends on what happens to be on disk, and a renamed section is a compile error rather than a silent content drop. A path the build does not bundle is a hard validation error. Prose therefore keeps living in reviewable markdown while this document carries only the manifest; `text` remains available for a rule authored in the manifest itself.",
+          "properties": {
+            "kind": { "const": "file" },
+            "path": {
+              "description": "Path relative to `assets/instructions/`, e.g. `sections/core.md`.",
+              "type": "string",
+              "minLength": 1
+            }
           }
         },
         {
@@ -127,9 +141,9 @@
   },
   "examples": [
     {
-      "schema_version": 1,
+      "schema_version": 2,
       "package_id": "trusty-mpm/example",
-      "description": "Minimal structurally-valid package. Content authoring is #4185; this example exists so the schema's own shape is executable and testable.",
+      "description": "Minimal structurally-valid package, exercising all three body kinds. The shipped instance is `pm-instruction-package.json`; this example exists so the schema's own shape stays executable and testable.",
       "trailing_newline": false,
       "sections": [
         {
@@ -216,7 +230,7 @@
         },
         {
           "section": "identity",
-          "body": { "kind": "text", "text": "## Identity" }
+          "body": { "kind": "file", "path": "sections/identity.md" }
         },
         {
           "section": "non-overridable-rules",
--- a/crates/trusty-mpm/src/assets/instructions/pm-instruction-package.json
+++ b/crates/trusty-mpm/src/assets/instructions/pm-instruction-package.json
@@ -1,0 +1,148 @@
+{
+  "schema_version": 2,
+  "package_id": "trusty-mpm.pm.bundled-fallback",
+  "description": "The DEFAULT PM instruction package (#4318). This file — not a Rust literal — is the authored manifest: the section taxonomy with its customization tiers, the ordered block stream, every join, and the declared composition-time generators. Bulk prose lives in `sections/*.md` and is referenced by `file` bodies; a rule authored here rather than in a section file uses a `text` body. Composition is byte-deterministic for a given manifest and roster (see `instruction_package::InstructionPackage::compose`). Deliberately carries NO `$schema` key: every object in the format sets `additionalProperties: false` / `deny_unknown_fields`, so an unrecognised key is a hard parse error rather than dropped instruction content. Validate against `instruction-package.schema.json` out of band.",
+  "trailing_newline": false,
+  "sections": [
+    {
+      "id": "identity",
+      "title": "Identity",
+      "customization_tier": "fixed",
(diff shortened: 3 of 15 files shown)


bobmatnyc:
## Code Critic Verdict: APPROVE

Zero CRITICAL, zero HIGH findings at >80% confidence. Four MEDIUM and two LOW below; none blocks the merge. Reviewed against the spec ([#4318](https://github.com/bobmatnyc/trusty-tools/issues/4318) plus the [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183) owner amendments), not against the implementer's narrative.

### Verified independently

| claim | how it was checked | result |
|---|---|---|
| **Fail-closed on every error arm** | traced `from_json` → `validate` → `compose` → `authored_run` → `resolve_pm_prompt_with_roster` | **holds.** `compose` calls `self.validate()?` first (`instruction_package.rs:921`), so the `unreachable!` at `:927` is genuinely unreachable — I probed it with a package carrying `{"kind":"file","path":"sections/nope.md"}` pushed through `to_json`→`from_json`→`compose` and got `Err(Invalid(UnknownFileSource { index: 8, .. }))`, not a panic. `BUNDLED` validates at `LazyLock` init; `manifest_run` additionally filters an empty projection before returning it. No arm can yield an empty or truncated prompt — the [#4286](https://github.com/bobmatnyc/trusty-tools/pull/4286) PR3 trap is closed. |
| **Byte-identical mechanism** | `git diff --numstat origin/main...HEAD -- src/core/testdata/` | **+25/+19/+25, zero deletions.** Zero deletions across all three goldens is the proof, independent of the commit sequence: mechanism drift would surface as a deletion or a modified line. The +69 additions are exactly the three rules and nothing else. |
| **Section coverage cannot be dropped from JSON** | `validate`'s `SectionWithoutBlocks` / `SectionOnlyOptionalBlocks` over `SectionId::CANONICAL` (`:783-796`) | **holds.** Deleting a floor block from the manifest is a validation error, red in CI via `bundled_manifest_parses_and_validates`. `every_authored_block_is_exactly_its_section_source` additionally pins each section to its own source file, so a swapped `path` is caught too. |
| **`deny_unknown_fields` everywhere** (the [#4223](https://github.com/bobmatnyc/trusty-tools/issues/4223) lesson) | grep of every serde container | **holds.** Present on `BlockBody` (`:340`), `InstructionBlock` (`:399`), `InstructionSection` (`:436`), `InstructionPackage` (`:462`). `schema_body_kinds_match_rust_variants` is a real drift gate — `$defs.body` is a `oneOf`, invisible to `schema_enums_match_rust_enums`. |
| **Shipped manifest satisfies the published schema** | `jsonschema` Draft 2020-12 validation of `pm-instruction-package.json` against `instruction-package.schema.json` | **0 errors.** The schema's own `examples[0]` also validates. |
| **No test lost in the 436-line move** | test-fn name diff, `main` vs `instruction_pipeline_tests.rs` | **13 → 16, none removed**, three added. No `#[ignore]`, no cfg-gate. |
| **Gates** | run locally at the PR head | `cargo test -p trusty-mpm --lib` → `test result: ok. 3985 passed; 0 failed; 6 ignored` (matches the PR body exactly). `check_instruction_floor.sh` PASS, `check_line_cap.sh` `7 allowlisted, 0 violations — OK`. All 20 CI checks green. |
| **`closingIssuesReferences`** | `gh pr view --json closingIssuesReferences` | **exactly `[4318]`.** "Part of [#4183]" correctly does not close the epic. |

### Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | 262-269 | **The anti-split-brain invariant does not hold for `agent-delegation`.** `assemble_system_prompt()` still concatenates the raw `AGENT_DELEGATION` constant, not `delegation_doctrine()` — as does `delegation_with_roster(None)` (`instruction_overrides.rs:466`). The module docs claim "editing the manifest moves every composer together; it cannot move one"; that is true for `core`, `memory`, `search` and `workflow`, and false for `agent-delegation`. No drift today — the only manifest-authored delegation block is the roster-precedence note, which is deliberately roster-conditional — but a future inline rule authored into that section reaches the packaged composer and the legacy assembly and never the installed roster-free prompt. That is the exact hazard this PR exists to close, left open on one section, with no test pinning it. | Either split the manifest's delegation blocks so the note is separable from a would-be doctrine rule and have `assemble_system_prompt` project the non-note authored run, or add a test asserting `assemble_system_prompt()` carries every non-note authored `agent-delegation` block. At minimum, correct the "cannot move one" claim at `bundled_pm_package.rs:64-74` to name the exception. |
| MEDIUM | `crates/trusty-mpm/src/assets/instructions/pm-instruction-package.json` | 4 | **Nothing in CI checks the shipped manifest against the published schema.** The manifest's own `description` says "Validate against `instruction-package.schema.json` out of band" — that is, by hand, that is, never. `schema_body_kinds_match_rust_variants` covers body-kind drift and `schema_enums_match_rust_enums` covers the enums, but a `required` entry or a `minLength` added to the schema would leave the published contract and the shipped instance silently divergent. It validates cleanly today (I checked); the gate is what is missing. | Add a test walking `SCHEMA_JSON`'s `required` arrays and `$defs` constraints against the parsed `PM_PACKAGE_JSON`, or take a `jsonschema` dev-dependency and assert `PM_PACKAGE_JSON` validates against `SCHEMA_JSON` beside the existing example test. |
| MEDIUM | `crates/trusty-mpm/src/assets/instructions/pm-instruction-package.json` | 199-247 | **All three owner rules are authored into project-overridable sections.** `core` and `workflow` are `customization_tier: project`, `permits(OverrideTier::Project) == true`, and `check_instruction_floor.sh` greps only `identity.md` / `non-overridable-rules.md` / `framework-guaranteed-conventions.md`. A `## CORE` or `## WORKFLOW` `CLAUDE.md` override, a `.trusty-mpm/WORKFLOW.md`, or a `PM_INSTRUCTIONS_DEPLOYED.md` therefore deletes them silently — which the `claude-md-override` golden correctly demonstrates. The implementation is right and the golden is right; the question is placement. Clickable references and the banned word are output-format guarantees of exactly the kind `framework-guaranteed-conventions` already carries (attribution footer, ticket attribution), and the opportunistic-fix rule is framework process doctrine. If the owner intends them as guaranteed, they are in the wrong tier. | Owner decision, not an engineering change. If guaranteed: move the two language rules into the `framework-guaranteed-conventions` section (tier `fixed`) and extend `check_instruction_floor.sh` to grep for them. If genuinely project-tunable: leave as is and record the ruling on [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183). |
| MEDIUM | — | — | **Provenance: one of the three rules has none on the record.** The opportunistic-fix rule traces to the owner comment on [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183) dated 2026-07-29T20:22:33Z; clickable references to 2026-07-29T22:33:20Z. The banned-word rule appears in no comment on [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183) and in no issue or comment repo-wide. Separately, the clickable-references order says the edit "should ride the next `core.md` instructions PR after #4316 lands"; this PR instead authors it as an inline JSON `text` block on the strength of an uncited "owner order, 2026-07-29: the 1.3 rules belong in JSON". Both may be genuine verbal orders — for the PM's own instruction supply chain they should be on the record before merge. | Post the banned-word rule and the author-in-JSON ruling as comments on [#4183](https://github.com/bobmatnyc/trusty-tools/issues/4183), as the other two rules were. No code change. |
| LOW | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | 88-96 | **"A renamed section is a compile error" overstates the guarantee**, repeated in the schema description, the CHANGELOG, and three doc comments. It is a compile error only when the markdown file is deleted or moved out from under `include_str!`. A section renamed consistently in `SECTION_SOURCES` while the manifest keeps the old `path` is a *runtime* `UnknownFileSource` — caught in CI by `bundled_manifest_parses_and_validates`, which is a fine guarantee, but a different one. | Reword to "a deleted section file is a compile error; a stale manifest path is a validation error red in CI". |
| LOW | `crates/trusty-mpm/src/core/instruction_package_tests.rs` | 1599-1644 | `schema_body_kinds_match_rust_variants` compares two `Vec<String>` by index, so reordering the schema's `oneOf` branches — semantically a no-op for JSON Schema — fails with a misleading "must match the Rust variants" message. | Compare as sets (`BTreeSet`) and assert the counts separately, or state in the message that branch order is pinned deliberately. |

### Notes

- **Merge precondition (already flagged in the PR body).** [#4316](https://github.com/bobmatnyc/trusty-tools/pull/4316) appends to the end of `sections/core.md` and regenerates the same three goldens at the same anchor. This PR references `core.md` by path, so it inherits that content — but whichever lands second must re-run `UPDATE_GOLDEN=1 cargo test -p trusty-mpm golden`, and the goldens must be re-verified as **content-only, zero deletions** afterwards. A rebase that produces a deletion in any golden invalidates the byte-identical-mechanism claim and this verdict with it.
- **`with_overrides` matching on `authored()` rather than `Text`** is the correct call and the single most important line in the diff. Without it, every bundled section becoming a `file` body would have made the whole taxonomy return `NoTextBlock` — an advertised-but-unread override, [#381](https://github.com/bobmatnyc/trusty-tools/issues/381) verbatim. It is pinned by the added assertion in `content_sections_accept_a_project_override`, including the "must not keep a file body alongside the override" check.
- **Out of scope, pre-existing, worth an epic note:** `bundle_all.rs:104` installs the 4-line `assets/instructions/INSTRUCTIONS.md` stub to `instructions/INSTRUCTIONS.md` under the framework root with `InstallPolicy::Overwrite` — the same path `install_system_prompt()` writes `assemble_system_prompt()` to (`~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md`). Two writers, one file. `tm launch` step 10 regenerates it after provisioning so the launch path recovers, but between a bundle deploy and the next launch the file on disk is the stub. Not introduced here and not this PR's to fix; it does mean "the installed prompt carries the manifest rules" is a fact about `assemble_system_prompt()`'s return value, not unconditionally about the file on disk.
- Reviewed in a detached worktree at `b6982d6e`. The probe test used to disprove the panic hypothesis was reverted; the worktree is clean.

Issue #4409, opened 2026-07-30
deployment has drifted from the day-1 design: bundled assets belong at user level in a protected config path


bobmatnyc:
**The design is implemented and working.** The defect is that bundled agents are deployed outside their segregated boundary. The design's purpose: **segregate claude code operations from the generic installation** — tm-driven Claude Code operations must not touch the user's own generic Claude Code install at `~/.claude`.

The day-1 design has two parts, both already working:
1. Repos clone to a trusty-mpm-specific path
2. A custom user config path (`CLAUDE_CONFIG_DIR` = `/Users/masa/.trusty-tools/trusty-mpm/claude-config`) holds the user agents

**Three breaches of one boundary — one violated invariant: tm writes outside its segregated config path**

| Where tm wrote | What it contaminates | Severity |
|---|---|---|
| `~/.claude/agents/` — **47 files** | the user's **generic Claude Code installation** — precisely what segregation protects | **HIGHEST** |
| project `.claude/agents/` — **42 files** incl. 2 stubs | project space; also the shadowing vector behind #4408 | **HIGH** |
| 12 project `settings.json` carrying tm hook entries | **other repositories entirely** (`hotstats-product-poc`, `writing`, +9) | **HIGH** |

These are not three separate bugs. They are one violated invariant. The only reason any of it surfaced is that one of those writes happened to be malformed (the 32-byte `rust-engineer.md` stub in project space). Everything else was silent.

**The segregation breach in detail:**

Measured with absolute paths, `rust-engineer.md` exists in three tiers:

| Tier | Size | Status |
|---|---|---|
| worktree project `.claude/agents/` | 32 bytes | stub, body is literally `v1` |
| `$CLAUDE_CONFIG_DIR/agents/` | 25,455 bytes | **correct agent with description** |
| **global `~/.claude/agents/`** | 28,650 bytes | **BREACH: contaminates generic Claude Code installation** |

The config path holds 42 correct agents. The design works. The breach: tm wrote bundled assets to `~/.claude` (the generic tier) and project space, violating segregation.

**Why it stayed invisible:**

1. `tm doctor` counts files, not boundaries. It never checks whether tm-owned assets exist outside `$CLAUDE_CONFIG_DIR`.
2. The delegation roster advertises agents and the project stubs despite finding no description.
3. `✅ deployment ... match the canonical roster` — the roster and deployment agree *with each other* while both diverge from the design.

**Related defects:**

[...]

PR #4419, opened 2026-07-30
fix(trusty-agents-common): redeploy corrupted bundled assets instead of freezing them as user-modified

### LINKED ISSUE #4408: content-free stub in project .claude/agents/ permanently shadows the real rust-engineer
**Defect:** A degenerate stub in the project-local `.claude/agents/` tier outranks the real agent from the custom config path. Because project tier outranks the config dir (and `$CLAUDE_CONFIG_DIR`), the broken copy shadows the correct one. This is a **direct consequence of the deployment violation documented in #4409**.

**Evidence — three-tier comparison of rust-engineer.md:**

| Tier | Size | Content |
|---|---|---|
| worktree project `.claude/agents/` | **32 bytes** | degenerate stub: body is literally `v1` |
| `$CLAUDE_CONFIG_DIR/agents/` = `/Users/masa/.trusty-tools/trusty-mpm/claude-config/agents/` | **25,455 bytes** | **correct agent with `description`, `model: sonnet`, `effort: balanced`, `agent_type: engineer`, `version: "2.0.1"`, `skills: [toolchains-rust-core, software-patterns]`** |
| global `~/.claude/agents/` | 28,650 bytes | real agent (legacy tier outside the design) |

The design places the authoritative copy in `$CLAUDE_CONFIG_DIR/agents/` where it lives and works. The project tier should never have had a copy.

**Why it stayed invisible — three checks each report healthy:**
1. `tm doctor` counts files, not content: reports `✅ agents 42 agent(s) deployed`. A 32-byte stub is still a file.
2. The delegation roster advertises the agent despite finding no description.
3. `✅ deployment ... match the canonical roster` — the roster and deployment agree *with each other* while both diverge from the design.

**Impact:** `rust-engineer` is the designated engineer for this workspace's detected stack (21 Rust crates). Every session routing Rust work hits dispatch failure and silently degrades to another agent. A live dispatch failed this way 2026-07-30.

**Resolution — two parts:**

1. **Immediate:** Delete the two project-local stubs (`.claude/agents/rust-engineer.md`, `.claude/agents/regression-agent.md`). The real agent lives in `$CLAUDE_CONFIG_DIR/agents/` and will surface.

2. **Root fix:** #4409 — stop deploying bundled assets into project `.claude/` entirely. That is the only way to prevent shadowing from happening again. Verify that `tm doctor` reads from `$CLAUDE_CONFIG_DIR/agents` as the authoritative location.

**Also worth investigating:**
- Where did the `v1` stub come from? It looks like a version-marker artifact rather than hand-authored. If some code path writes these, that path is the real bug.
- `regression-agent.md` (173 bytes) exists ONLY in project space — absent from both the config path and global. It is purely an artifact of the forbidden deployment.

Related: #4409 (the root cause — deployment violation that makes this shadowing possible).

bobmatnyc:
## Root cause

`AgentManifest::checksum_matches` mismatch was mapped straight to "the user
edited this file — preserve it" in `agents::deployer::deploy_agents_filtered`,
with no reference to the manifest entry's declared ownership.

That is correct for the user-owned (`InstallPolicy::SeedOnce`) tier and fatal
for the framework-owned (`Origin::Bundled` + `InstallPolicy::Overwrite`) tier.
Corrupt content can never checksum-match again, so once a bundled agent's
deployed copy corrupts, the corruption is permanently misclassified as a user
edit: every subsequent deploy skips it, and `tm validate --repair` skips it too.
In the reported incident a 32-byte stub (`name:` frontmatter, body `v1`, no
`description:`) replaced the real 25KB `rust-engineer.md` while the manifest
still recorded the real content's checksum. Result: "Agent type not found" for
every session in that workspace, permanently, while `tm doctor` (which counts
files, not content) reported the roster healthy.

## Fix seam

The bug was in the caller, not `checksum_matches`. The mismatch branch now
consults the entry's declared origin:

- `crates/trusty-agents-common/src/agents/manifest.rs` — new
  `Origin::is_framework_owned()`: `true` only for `Bundled`. `User` and
  `Registry` stay in the preserve-on-mismatch tier.
- `crates/trusty-agents-common/src/agents/deployer.rs` — a managed file whose
  checksum mismatches is re-deployed when its entry is framework-owned, and
  still skipped when it is user-owned. Untracked files (nothing in the manifest)
  are untouched exactly as before — that is the hand-placed-agent path.
- Every overwrite-on-mismatch repair emits a `warn!` with the filename, the
  expected and found checksums, and the on-disk byte count, so a corruption
  event is visible instead of silent.

Scoping notes:

- **Tier-independent.** This is manifest ownership semantics, so it protects
  bundled entries at whatever tier they are deployed to — including the user
  tier (`$CLAUDE_CONFIG_DIR`) where bundled agents are to live. The
  deploy-target architecture change is tracked separately in
  [#4409](https://github.com/bobmatnyc/trusty-tools/issues/4409); nothing in
  `paths.rs` or any deploy-target selection is touched here.
- **Skills are unaffected.** The skill deployer
  (`trusty-agents-common/src/skills/deployer.rs`) is a separate code path whose
  manifest entries carry no origin at all; its user-edit freeze is untouched.
- **Workspace-local blast radius.** Per-workspace deployment meant the
  corruption was confined to one workspace — the tm-agents workstream in the
  same repo was unaffected. But any workspace that corrupts stays broken forever
  without this fix, because no repair path could ever reach the file.

### Deliberate behavior change

A managed **bundled** agent that was hand-edited in place is now refreshed
rather than preserved. That is the settled two-variant install policy:
framework-owned assets overwrite, user-owned assets seed once. Three tests that
encoded the old semantic were rewritten to assert the new one; the user-owned
guards were kept and strengthened. Hand-placed agents (files trusty-mpm never
deployed, so absent from the manifest) are still never touched, and
`tm install --reset-agents` remains the backup-preserving reconciliation path.

## Test evidence

New / changed tests, both directions:

| Test | Asserts |
|---|---|
| `agents::deployer::tests::deploy_redeploys_corrupted_bundled_file` | bundled entry + the literal `v1` stub on disk → re-deployed, content byte-identical to the bundle, manifest checksum matches after, third run is `unchanged` |
| `agents::deployer::tests::deploy_preserves_modified_user_owned_entry` | `Origin::User` entry with modified content → skipped and byte-identical across two deploys, origin unchanged |
| `agents::manifest::tests::origin_framework_ownership_tiers` | only `Bundled` is framework-owned |
| `core::session_launch::sync_assets::tests::sync_assets_repairs_drifted_bundled_agent` | same repair at the session-launch tier |
| `core::session_launch::sync_assets::tests::sync_assets_never_touches_hand_placed_agent` | a hand-placed (untracked) `.claude/agents/` file survives sync byte-identical |
| `e2e::test_agent_deploy::deploy_repairs_drifted_bundled_file` | end-to-end repair through the `trusty-mpm` re-export |

The regression test was confirmed to FAIL against pre-fix behavior (with
`is_framework_owned` forced to `false`):

```
---- agents::deployer::tests::deploy_redeploys_corrupted_bundled_file stdout ----
panicked at deployer_tests.rs:94:5:
a corrupted bundled agent must be re-deployed, got: DeployResult { deployed: [],
skipped: ["engineer.md"], unchanged: ["base-agent.md"], ... }
```

Gates (all green, crate-scoped):

```
test result: ok. 329 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm
test result: ok. 4009 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
(+ 29 further green test binaries, 0 failures)

cargo clippy -p trusty-mpm -p trusty-agents-common --all-targets -- -D warnings   EXIT=0
cargo fmt --check                                                                 clean
bash scripts/check_line_cap.sh    line-cap: 7 allowlisted, 0 violations — OK.
```

Closes #4408

### DIFF
--- a/crates/trusty-agents-common/CHANGELOG.md
+++ b/crates/trusty-agents-common/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes are documented in this file.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ---
+## [Unreleased]
+
+### Fixed
+
+- `agents::deployer` now re-deploys a corrupted bundled agent instead of freezing it as user-modified (closes [#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408)). A manifest entry whose origin is framework-owned (`Origin::Bundled`, the `InstallPolicy::Overwrite` tier) is refreshed from the bundle whenever its on-disk checksum drifts — a mismatch there means corruption or drift, never user ownership. Previously ANY checksum mismatch was read as "the user edited this file", which is unrecoverable for a bundled asset: corrupt content can never checksum-match again, so a 32-byte `v1` stub that replaced the real 25KB `rust-engineer.md` was permanently misclassified as a user edit and skipped by every subsequent deploy and by `tm validate --repair`, dropping the agent from the session roster for the life of the workspace. The user-owned tier is unchanged — an untracked file, and a tracked entry with `Origin::User`/`Origin::Registry`, are still preserved byte-for-byte. Each overwrite-on-mismatch repair logs the file, the expected and found checksums, and the on-disk byte count.
+
+### Added
+
+- `agents::manifest::Origin::is_framework_owned` — names the two install-ownership tiers so the deployer branches on declared ownership rather than on checksum mismatch alone (#4408).
+
 ## [0.3.0] — 2026-07-21
 
 ### Fixed
--- a/crates/trusty-agents-common/src/agents/deployer.rs
+++ b/crates/trusty-agents-common/src/agents/deployer.rs
@@ -20,10 +20,17 @@
 //! manifest. Corrupt manifests are detected and surfaced as errors rather
 //! than silently reset to empty. Returns a [`DeployResult`] summarising what
 //! happened.
+//! Ownership follows the two-variant install policy (#4408): a manifest entry
+//! whose origin is FRAMEWORK-owned (bundled, the `Overwrite` tier) is
+//! re-deployed whenever its on-disk checksum drifts — a mismatch there means
+//! corruption, not user ownership — while a user-owned entry (the `SeedOnce`
+//! tier) and any untracked file keep the preserve-on-mismatch behavior.
 //! Test: `cargo test -p trusty-agents-common agents::deployer` covers a new
-//! deploy, a skipped user-modified file, an unchanged file, a user-owned file,
-//! atomic writes, corrupt manifest detection, and (#3556) a stale-broken-copy
-//! refresh plus a strict-YAML-invalid composition being isolated and skipped.
+//! deploy, a preserved user-owned file, an unchanged file, atomic writes,
+//! corrupt manifest detection, (#3556) a stale-broken-copy refresh plus a
+//! strict-YAML-invalid composition being isolated and skipped, and (#4408) a
+//! corrupted bundled file being re-deployed while a modified user-owned entry
+//! survives byte-identical.
 
 use std::collections::HashMap;
 use std::path::Path;
@@ -50,8 +57,10 @@ use crate::agents::metadata::agent_metadata_from_str;
 pub struct DeployResult {
     /// Filenames successfully (re)written this run.
     pub deployed: Vec<String>,
-    /// Filenames skipped because the user modified them, or because they were
-    /// untracked and differ from the fresh composition.
+    /// Filenames skipped because they are user-owned — either untracked by the
+    /// manifest and differing from the fresh composition, or tracked with a
+    /// user-owned origin (the seed-once tier) and edited. A tracked
+    /// FRAMEWORK-owned file that differs is re-deployed, not skipped (#4408).
     pub skipped: Vec<String>,
     /// Filenames left untouched because their checksum already matched.
     pub unchanged: Vec<String>,
@@ -114,7 +123,11 @@ pub fn is_agent_file(name: &str) -> bool {
 /// Rules:
 ///   - Not in manifest → user-owned → skip silently
 ///   - In manifest, checksum matches → overwrite (safe)
-///   - In manifest, checksum differs → user-modified → warn + skip
+///   - In manifest, checksum differs, entry is framework-owned
+///     ([`Origin::is_framework_owned`], the Overwrite tier) → drift or
+///     corruption → warn + re-deploy the bundled composition (issue #4408)
+///   - In manifest, checksum differs, entry is user-owned (the seed-once
+///     tier) → user-modified → skip, preserving their content byte-for-byte
 ///   - New trusty-mpm agent → compose + write (atomic) + add to manifest
 ///   - Corrupt manifest → error (never silently reset, which would reclassify
 ///     managed files as user-owned and skip re-deploying them)
@@ -274,6 +287,13 @@ pub fn deploy_agents_filtered(
                 continue;
             }
             let current = std::fs::read_to_string(&target_path)?;
+            // #4408: snapshot the recorded checksum + ownership tier before the
+            // branches below take a mutable borrow of `manifest` to re-register
+            // the file.
+            let recorded_owner = manifest
+                .managed
+                .get(&filename)
+                .map(|e| (e.checksum.clone(), e.origin.is_framework_owned()));
             if manifest.checksum_matches(&filename, &current) {
                 if checksum(&composed) == checksum(&current) {
                     // Deployed copy is already the latest composition.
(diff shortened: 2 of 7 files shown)


bobmatnyc:
## Verdict: APPROVE

Adversarial review of the actual PR diff (7 files, ~490 lines: `deployer.rs`, `deployer_tests.rs`, `manifest.rs`, `sync_assets.rs`, `test_agent_deploy.rs`, 2 CHANGELOGs). Verified by running the tests myself, not by trusting the PR description.

```
cargo test -p trusty-agents-common agents::deployer          -> 19 passed; 0 failed
cargo test -p trusty-agents-common agents::manifest::tests::origin_framework_ownership_tiers -> 1 passed
cargo test -p trusty-mpm --lib session_launch::sync_assets   -> 5 passed; 0 failed
cargo test -p trusty-mpm --test e2e -- test_agent_deploy     -> 5 passed; 0 failed
cargo clippy -p trusty-agents-common --all-targets           -> clean
```

### Focus 1 — seam coverage
`deployer.rs::deploy_agents_filtered` is the single point of ownership classification. Traced every caller: `sync_assets.rs`, `update_check/apply.rs`, `session_launch/mod.rs` (`prepare_session_inner`), and `deploy_validate`'s repair path (via `prepare_session_with_repo_url`) all call `deploy_agents_filtered` directly — none reimplements checksum/origin logic, so all inherit the fix for free. `agent_reset.rs` (`tm install --reset-agents`) is a separate, explicit, backup-preserving path untouched by this PR, as documented — correctly out of scope. No other freeze-on-mismatch path was found for agents.

### Focus 2 — hand-placed-agent guarantee
Confirmed by code read: the untracked branch (`!manifest.is_managed(&filename)`, deployer.rs:262-288) is untouched by this diff and `continue`s before the new origin check is ever reached. There is no code path by which an untracked `.claude/agents/` file can be reached by the new branch. `sync_assets_never_touches_hand_placed_agent` plus the existing `deploy_user_owned_skipped` cover it. Guarantee holds.

### Focus 3 — User/Registry preservation
`deploy_preserves_modified_user_owned_entry` (deployer_tests.rs) asserts byte-identical content and unchanged `Origin` across **two** consecutive deploys of an `Origin::User` entry with a manually diverged checksum. Passes. `Origin::Registry` is not currently constructed anywhere in the codebase (grep confirms), so it's conservatively covered by the same `is_framework_owned() == false` branch but has no live test fixture — acceptable, since the enum variant is forward-compatibility scaffolding per its original doc comment, not new to this PR.

### Focus 4 — no-backup deviation, blast radius
| MEDIUM | crates/trusty-agents-common/src/agents/deployer.rs:304-320 | 304 | The automatic repair path (runs unattended on every session launch, `sync-assets`, catalog apply, and daemon-triggered doctor auto-repair) discards a divergent bundled file with only a `tracing::warn!` — no backup copy. The sibling explicit path for the identical divergence event, `agent_reset.rs::reset_agents` (`--reset-agents`), always copies the divergent file to a timestamped backup before overwriting (agent_reset.rs:196-205). The CLI renderer (`install.rs::deploy_report_lines`) also gives a repaired file the *exact same* line as a routine fresh deploy (`✓ file (composed: …)`) — nothing distinguishes "we just discarded content that didn't match" from a normal refresh. `tracing::warn!` defaults to visible on stderr for foreground `tm` invocations (default filter is `"info"`), but the auto-repair path also fires inside the daemon process (doctor auto-repair on spawn/resume per `deploy_validate`'s module doc), where warnings land only in the daily-rotated log file, not anywhere an operator is likely to see in the moment. | Given the settled ruling, the overwrite itself is correct and not in question — a checksum mismatch on `Origin::Bundled` is corruption by definition and must be repaired. The gap is defense-in-depth: take the same cheap timestamped-backup precaution `reset_agents` already takes for the identical class of event, and/or give the CLI a distinct glyph/line for "repaired (was corrupted/drifted)" vs. plain "deployed" so an operator who *did* intentionally hand-edit a bundled file at least sees evidence of what happened. Not a blocker — the ruling already accepts this destruction as intended; this is a follow-up hardening suggestion. |

### Focus 5 — test quality
All rewritten/added tests pass (see verification above). The user-owned guard test constructs its fixture directly (`Origin::User` + a manually-diverged checksum via `ManifestEntry`) rather than depending on the fix's own write path, and re-runs the deploy twice — it is decoupled from the fix and would pass identically under the pre-fix code (which preserved ALL managed-checksum-mismatches regardless of origin, since origin didn't exist as a check). The corrupted-bundled-file regression test (`deploy_redeploys_corrupted_bundled_file`) is a genuine regression test: the PR body shows it panicking pre-fix with a concrete assertion failure, and the current code's single `else if let Some((_, true)) = recorded_owner` branch is the only thing standing between old and new behavior — easy to trace, easy to break by reverting one branch. Confidence >80% these tests pin the new semantics correctly, both directions.

### Notes
- Confirmed (not a finding): `crates/trusty-agents-common/src/skills/manifest.rs`'s `SkillManifestEntry` carries no `origin` field at all, so the identical bug shape (a corrupted bundled *skill* gets permanently frozen as "user-modified") still exists there. The PR explicitly scopes this out ("Skills are unaffected... untouched") and issue #4408 was filed specifically for agents — correctly out of scope for this PR, not flagging as a finding against unchanged code per the review guard rail, but worth a follow-up issue if not already tracked.
- `recorded_owner` is computed from `manifest.managed.get(&filename)` at a point where `manifest.is_managed(&filename)` is already guaranteed `true` (the untracked branch already `continue`d above) — so the `None` case of the `else if let Some(...)` is unreachable in practice; confirmed correct, not a bug.

No issues found at CRITICAL or HIGH severity. **APPROVED** for next pipeline stage.

Issue #2777, opened 2026-07-16
bare tm inside a decommissioned managed tmux session reconnects to itself instead of relaunching


bobmatnyc:
## Repro (tm 0.19.17, 2026-07-15)

Inside tmux session `tm-apex-01` (a tm-managed session whose managed state is `decommissioned`), tm printed:

```
tm: run `tm` to relaunch this session
~/trusty-mpm-projects/duettoresearch/apex/.worktrees/tm-apex-01 (git:docs/publish-product-editions) > tm
tm: this tmux session ('tm-apex-01') is already a managed session (state=decommissioned) — not nesting a new session here; reconnecting to 'tm-apex-01' instead…
tm: switching client to session 'tm-apex-01'
```

## Problem

The user is told "run `tm` to relaunch this session" (the pane's trailing echo after `claude` exits). But bare `tm` then hits the anti-nesting guard, sees the current tmux session is already managed, and "reconnects" — i.e. switches the tmux client to the session the user is ALREADY IN. Net effect: a dead-end loop; the decommissioned session is never relaunched.

Root cause: bare `tm`'s in-pane relaunch path (`plan_inplace`) only accepts `state == "stopped"`, so a `decommissioned` record falls through to the nested-session guard. There, `pane_identity_confirmed` is false (a decommissioned record has no captured/matching `pane_id`), so the guard takes the session-name-only fallback branch, which unconditionally calls `resume_guided_session`. For a decommissioned session with a still-alive tmux session, `plan_resume` resolves to `Attach` → `switch-client` → a guaranteed no-op because the operator is already inside that session.

## Expected

When bare `tm` runs INSIDE a managed tmux session whose managed state is not running a live `claude` (decommissioned, and likewise stopped/errored), it should RELAUNCH/resume the session in place (the same thing the "run `tm` to relaunch this session" hint promises), not switch-client to itself. When the session IS live (a runtime in a sibling window), the current reconnect behavior is correct.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #2486, opened 2026-07-12
bug(trusty-mpm): launchd bootout/bootstrap restart races trusty-console auto-reconnect — console-respawned daemon passes /health but lacks plist environment


bobmatnyc:
Found during the 2026-07-12 config-surface install verify (tm 0.19.4 @ bfb58a61). The documented bootout→install→bootstrap convention failed twice in a row: trusty-console's auto-reconnect won the race and respawned a console-child daemon (PPID=console, not 1) that answered /health 200 but was missing all plist EnvironmentVariables (TELEGRAM_BOT_TOKEN, OPENROUTER_API_KEY, ...) — Telegram poller never started. A healthy /health is NOT sufficient evidence the launchd-owned process won. Resolved manually via killing the console-spawned chain + launchctl kickstart -k. Fix directions: (a) daemon startup should detect it wasn't launched by launchd when a launchd plist exists and either defer or annotate /health (e.g. supervised:false); (b) console's reconnect should not SPAWN a daemon when a launchd unit exists (connect-only); (c) update the restart convention docs (CLAUDE.md + docs/reference/release-workflow.md) to verify PPID=1 / launchctl print ownership after bootstrap. Refs #534, #2386.

PR #4430, opened 2026-07-30
fix(trusty-mpm): decommission clears pending_decision, supervisor_status filters terminal states

### LINKED ISSUE #4400: decommissioning a session does not clear pending_decision — supervisor_status accumulates phantom human-confirmation gates
**Defect:** `pending_decision` survives decommissioning. `supervisor_status` keeps reporting decommissioned sessions in `pending_decisions` indefinitely, so the human-confirmation queue fills with entries no human can act on.

**Evidence** (observed 2026-07-30 via `mcp__trusty-mpm__supervisor_status` and `mcp__trusty-mpm__session_status`): two sessions appear in `pending_decisions` with `"T4: irreversible or security-sensitive operation; human confirmation required"` while both carry `"state":"decommissioned"`:
- `e8450956-29e0-47ef-a46f-8a1843a0343e` / `tm-wave2-real-deliverab-01`, created 2026-07-11, task `"wave2 throwaway verification - will be decommissioned immediately"`
- `b9260058-126d-4eab-82e7-e74872e93645` / `tm-wave3-deliverable-01`, created 2026-07-12, task `"THROWAWAY wave3 deliverable-glyph verification - safe to decommission"`

Both have `last_activity_at` identical to `created_at` — they never ran. They have sat in the queue 19 days.

**Why it matters:** an alert list that only ever grows trains operators to ignore it. A real T4 gate on a live session is indistinguishable from 19-day-old dead entries, so it gets missed. Alarm fatigue on the one queue whose entire purpose is to stop and wait for a human.

**Resolution:** clear `pending_decision` on the decommission path, and have `supervisor_status` filter `pending_decisions` to non-terminal states regardless — defense in depth, since existing rows already carry the stale field. Include a backfill or one-time clear for the rows already in this state.

bobmatnyc:
## Summary

Decommissioning a session did not clear `pending_decision`, so `supervisor_status` kept reporting decommissioned sessions in `pending_decisions` indefinitely. Two throwaway sessions that never ran sat in the human-confirmation queue for 19 days — a real T4 gate on a live session became indistinguishable from dead entries, training operators to ignore the one queue whose entire purpose is to stop and wait.

- `SessionManager::decommission_with_root` now clears `pending_decision`/`proposed_default` unconditionally on every decommission path (fresh teardown, record-only tombstone, dirty-worktree-refused teardown).
- `FleetMetrics::from_records` (backing `supervisor_status`) filters the surfaced `pending_decisions` list to non-terminal states regardless — defense in depth, since rows persisted before this fix still carry the stale field.
- `reconcile_on_boot` backfills (clears) the stale `pending_decision`/`proposed_default` on any terminal (`Decommissioned`/`Deleted`) record it encounters, reported via the new `ReconcileReport::stale_decisions_cleared` field and logged at boot.

## Test plan

- [x] `decommission_clears_pending_decision` — new test asserting both fields clear (returned value and persisted store) on decommission.
- [x] `metrics_filters_pending_decisions_on_terminal_states` — new test asserting `Decommissioned`/`Deleted` records with a stale `pending_decision` never appear in `FleetMetrics::pending_decisions`.
- [x] `manager_reconcile_backfills_stale_pending_decision_on_terminal_record` — new test asserting the boot reconcile clears a pre-existing stale field on a terminal record without resurrecting its state.
- [x] `cargo fmt --check`, `cargo clippy -p trusty-mpm --all-targets -- -D warnings`, `cargo test -p trusty-mpm` (full surface, all targets) — all green: 4012+1303+... passed, 0 failed across every test binary.

Closes #4400


bobmatnyc:
## Verdict: BLOCK

Adversarial review against issue #4400's stated requirements (decommission clears `pending_decision`/`proposed_default` on every path; `FleetMetrics::from_records` filters to non-terminal states; boot reconcile backfills stale fields on terminal records with `ReconcileReport::stale_decisions_cleared`; boot log count). Reviewed at head `eca3957f1cbf4c5f9b0db1a02dbb0c4d1e142279`.

The core logic is correct — see Notes — but the PR cannot merge as submitted because it fails a hard, deterministic CI gate.

## Findings

| CRITICAL | `crates/trusty-mpm/src/session_manager/decommission.rs` | whole file (528 SLOC) | This diff pushes the file from under 500 SLOC to 528, tripping the repo's enforced production-file cap. CI job "500-line file-size cap" fails deterministically: `FAIL: crates/trusty-mpm/src/session_manager/decommission.rs is 528 SLOC (> 500 prod cap) and not allowlisted. New oversized file; split it or it cannot merge.` (run 30582295070). Not flaky — same result on repeat poll. This is the exact invariant the file's own module doc (lines 1-13) says it was extracted to protect. | Split the `#4400` additions into a sibling module (mirroring the file's own cited precedent: `adopt.rs`/`reactivate.rs`/`hook_sync.rs`) rather than growing `decommission.rs` further — e.g. move the two new tests (`decommission_clears_pending_decision` and friends) or the `remove_session_worktree`/`is_session_worktree` worktree helpers out. The cap script explicitly refuses to allowlist a *new* oversized file, so `--update` is not an option here. |
| MEDIUM | `crates/trusty-mpm/src/session_manager/delete.rs` | 83-87 (`delete_record`) | `delete_record` transitions a record to the terminal `ManagedSessionState::Deleted` but does not clear `pending_decision`/`proposed_default`, unlike `decommission_with_root` (decommission.rs:600-601). Inconsistent with this PR's own stated goal of clearing these fields on the path to a terminal state — a session soft-deleted via `tm sessions delete` keeps a stale `pending_decision` in the store until the next daemon restart. Not user-visible today: `FleetMetrics::from_records`'s new `is_terminal()` filter (metrics.rs:128) and the boot reconcile backfill (reconcile.rs:88-98) both already suppress it from `supervisor_status`/`/metrics` regardless of the raw field, so there is no live symptom — this is a field-hygiene gap, not a functional regression. | Clear both fields in `delete_record` too, mirroring decommission.rs:600-601, for consistency with the rest of the terminal-state handling this PR just introduced. |

## Required Changes

1. Split `decommission.rs` (or otherwise reduce it) below the 500-SLOC production cap so the "500-line file-size cap" CI check goes green — this alone blocks merge under this repo's own enforced gate.
2. (Optional, not blocking) Clear `pending_decision`/`proposed_default` in `delete_record` for consistency with the decommission path.

## Notes

- Logic review of the actual fix is clean: `decommission_with_root` clears both fields unconditionally at decommission.rs:600-601, reached by every path (fresh teardown, `record_only` tombstone, dirty-worktree-refused teardown, unowned-skip, containment-guard-skip) — all of them fall through to the same tail before `store.upsert`, and every early `?`-return (owner-mismatch, `remove_dir_all` I/O error) happens *before* any field mutation, so there is no fail-open/cursor-advance gap: a failed decommission changes nothing, a successful one always clears both fields.
- Checked whether the boot-reconcile backfill (reconcile.rs:82-98) could clear a decision that's still legitimately pending: it only fires inside the `record.state.is_terminal()` branch (`Decommissioned`/`Deleted` only — record.rs:138-140), and nothing in the codebase resurrects a terminal record's `pending_decision` — `mark_reactivated` (reactivate.rs) explicitly revives `Decommissioned → Active` for the bare-`tm` in-pane relaunch case (#2777) but never restores `pending_decision`, and by design clears `claude_session_id` too since a reactivation is a fresh `claude` process for which any prior pending question is stale. No legitimate-pending-decision-gets-wiped scenario found.
- `FleetMetrics::from_records`'s new filter (metrics.rs:128-136) and the boot backfill correctly key off `ManagedSessionState::is_terminal()` rather than hand-rolled state lists, so a future terminal variant can't slip past either guard the way `Deleted` once did for the old reconcile check.
- Boot log line (`daemon/state/core.rs:685-699`) correctly folds `stale_decisions_cleared` into both the "anything happened" gate and the logged fields.
- Test coverage matches the three postconditions named in the issue: `decommission_clears_pending_decision` (both return value and persisted store), `metrics_filters_pending_decisions_on_terminal_states` (Decommissioned + Deleted + Active), `manager_reconcile_backfills_stale_pending_decision_on_terminal_record` (asserts state is not resurrected, both fields clear, id reported).
- Out of scope for this review but worth flagging operationally: `gh pr view` reports `mergeable: CONFLICTING` — `main` has moved (`a51001a6` → `46e080ed`) since this branch's merge-base; needs a rebase/merge regardless of the findings above.
- CI status at review time: "500-line file-size cap" = fail (confirmed, deterministic); Clippy, Format, MSRV (1.91), and every other check = pass; `Test` (the full `cargo test -p trusty-mpm` run) was still `pending` after ~10+ minutes — not observed to completion, not counted as evidence either way.

bobmatnyc:
## Verdict: APPROVE (supersedes prior BLOCK)

Targeted re-review of the delta from previously-reviewed head `eca3957f` (rebased-equivalent `8bbef997`) to new head `afbd1e3e`. No issues found at >80% confidence. APPROVED for next pipeline stage.

Both prior findings are resolved:

**Former CRITICAL (SLOC cap) — RESOLVED.** The inline `#[cfg(test)] mod tests` block was extracted from `decommission.rs` into a new sibling `decommission_tests.rs`, with `mod decommission_tests;` added in `mod.rs` (matching the `dedup.rs`/`dedup_tests.rs` precedent already used elsewhere in this crate). Diffed the old inline module against the new file directly: all 9 test functions (`is_session_worktree_detects_dot_worktrees_component`, `is_session_worktree_absent_path_is_noop`, `sentinel_gates_worktree_removal_refuses_non_worktrees_dir_without_sentinel`, `sentinel_present_passes_safety_gate`, `decommission_owner_gate_refuses_foreign_caller`, `decommission_owner_gate_allows_self`, `decommission_owner_gate_allows_terminal_owner`, `decommission_owner_gate_bypassed_for_none_caller`, `decommission_clears_pending_decision`) plus the `owned_record` helper are present, byte-identical modulo indentation (pure de-nesting) — no assertion weakened, no test dropped. The `use super::*;` was correctly replaced with explicit imports (`super::decommission::{WORKTREE_SENTINEL_FILE, is_session_worktree, remove_session_worktree}`, `super::manager::{ManagedError, SessionManager}`, `super::record::{...}`) — `remove_session_worktree` is `pub(super)` from `session_manager`'s perspective, so it's visible to the sibling `decommission_tests` module too. `decommission.rs` is now production-only. CI confirms: "500-line file-size cap" = pass, Clippy = pass, MSRV (1.91) = pass, full `Test` job = pass (16m12s, not just "started" — observed to completion this time).

**Former MEDIUM (delete_record) — RESOLVED.** `delete.rs:85-91` now clears `updated.pending_decision = None; updated.proposed_default = None;` immediately before `self.store.write().await.upsert(updated).await?`, mirroring `decommission_with_root`'s pattern. Verified there is exactly one path through `delete_record`: the only early return (`is_running` guard, when `!force`) happens *before* the `record.clone()`/mutation, so it changes nothing on the reject path — consistent with the no-fail-open pattern already verified in `decommission_with_root`. The success path always clears both fields before the single fallible `upsert`. New test `delete_record_clears_pending_decision` asserts the *persisted* copy (`mgr.get`) has both fields cleared post-delete; it deliberately does not assert on `delete_record`'s own return value, which is documented (and unchanged) to be the PRE-deletion snapshot for callers that want to render prior state — that asymmetry is intentional, not a bug.

**Rebase check.** `git diff eca3957f 8bbef997` (the rebase-only replay of the original commit) touches only `CHANGELOG.md`/`Cargo.lock`/`Cargo.toml` plus files from main's own unrelated commits picked up by the rebase (`bin/tm/commands/daemon_run.rs`, `launchd_probe.rs`, etc. — the #2486 bare-`tm` fix, nothing to do with #4400). `crates/trusty-mpm/src/session_manager/**` and `crates/trusty-mpm/src/supervisor/**` are byte-for-byte identical between the two — confirmed via `git diff` returning empty on those paths. The rebase did not touch any of the previously-reviewed decommission/reconcile/metrics logic.

No new findings introduced by this delta.

## Notes

- All prior verified-clean findings carry over unchanged: `decommission_with_root` clears both fields unconditionally on every teardown path with no fail-open gap; the boot-reconcile backfill only fires on `is_terminal()` records and nothing resurrects a terminal record's `pending_decision` with a legitimately-still-pending decision (`mark_reactivated` deliberately treats a revived session as a fresh lifecycle).
- CI at this head: all 24 checks pass, including the two that were incomplete/failing at last review (`500-line file-size cap`, full `Test` run). `gh pr view` reports `mergeable: MERGEABLE`.

PR #4422, opened 2026-07-30
fix(trusty-mpm): stop bare tm daemon bypassing the #2486 unsupervised-spawn guard

bobmatnyc:
## Summary

- The #2486 guard against an unsupervised daemon duplicating a launchd-managed one existed only on the MCP-bridge path (`serve_stdio.rs`'s `no_spawn`, keyed off `compute_no_spawn`). `run_daemon` (the bare-CLI path — `tm daemon` invoked directly, or by the launchd plist itself) never consulted any guard, so it walked straight past the same hazard.
- Observed live: PID 35895 serving a stale 1.2.3 image since 2026-07-29 while the on-disk binary was already 1.3.0, because the daemon detected the hazardous state, logged a warning, and proceeded anyway.
- Adds `compute_daemon_refuse(supervised, force)` in `launchd_probe.rs` — a new pure decision distinct from `compute_no_spawn` (which only takes `plist_exists` and would wrongly refuse the legitimate launchd-invoked case). `apply_supervision_signal` now returns the `supervised` bool it already computes; `run_daemon` calls the new guard right after and refuses with an actionable error unless `--force` is passed.
- New `tm daemon --force` flag lets an operator explicitly opt into an unsupervised run.

## Changes

- `crates/trusty-mpm/src/bin/tm/commands/launchd_probe.rs`: `compute_daemon_refuse` + 3 unit tests.
- `crates/trusty-mpm/src/bin/tm/commands/daemon_run.rs`: `apply_supervision_signal` now returns `bool`; new guard check + `unsupervised_daemon_error()`; `run_daemon` takes a `force: bool` param; updated/added tests.
- `crates/trusty-mpm/src/bin/tm/cli/mod.rs`: `Command::Daemon` gains `force: bool` (`--force`).
- `crates/trusty-mpm/src/bin/tm/main.rs`: threads `force` through to `run_daemon`.
- `crates/trusty-mpm/src/bin/tm/tests.rs`: updated `cli_parses_daemon_defaults`, added `cli_parses_daemon_force`.
- `crates/trusty-mpm/CHANGELOG.md`: Unreleased/Fixed entry.

## Test plan

- [x] `cargo fmt --check -p trusty-mpm`
- [x] `cargo clippy -p trusty-mpm --bin tm --all-targets -- -D warnings`
- [x] `cargo test -p trusty-mpm --bin tm` — 1308 passed, 0 failed, 1 pre-existing ignored

[...]

PR #4431, opened 2026-07-30
feat(trusty-mpm): infer abbreviated tm subcommands via clap

### LINKED ISSUE #4398: tm CLI: abbreviated subcommands and a -v short for --version
**Ask:** let users type short forms instead of full subcommand names, derived automatically rather than hand-maintained.

**Three separable pieces, in ascending cost:**
1. `#[command(infer_subcommands = true)]` on the top-level `Cli` struct (`crates/trusty-mpm/src/bin/tm/cli/mod.rs:62-69`). Propagates to all 3 nesting levels automatically (clap global setting). Zero collisions with the 5 existing `#[arg(short)]` attributes — those are option shorts, a separate matching path. Every `tm` call site in `scripts/`, `.github/workflows/`, `docs/`, and skills uses full spellings today, so nothing breaks. clap's exact-match fallback protects the `hook`/`hooks`, `project`/`projects`, `session`/`sessions`, `status`/`statusline` pairs.
2. `-v` for `--version`: **not** automatic. clap auto-generates `-V` (uppercase); `-v` is currently unused in the tm surface but claiming it requires overriding the generated version arg. Note `-v` is the near-universal convention for `--verbose` elsewhere — this is a one-way door worth a deliberate decision.
3. `tm sess` will **not** resolve even with inference on — `sess` is a 4-way ambiguous prefix (`sessions`, `sessctl`, hidden `session`, alias `session-manager`). Needs an explicit `visible_alias`, same pattern as `coordinator`'s existing `sm`.

**Resolution:** land (1) on its own; treat (2) and (3) as separate decisions.

**Verification note:** after enabling inference, run a full `cargo test -p trusty-mpm` — check no test asserts a currently-invalid abbreviated form as an expected error.

bobmatnyc:
## Summary
- Enables clap's `infer_subcommands` on the top-level `Cli` struct (`crates/trusty-mpm/src/bin/tm/cli/mod.rs`) so any unambiguous prefix of a subcommand name resolves (e.g. `tm doc` for `tm doctor`, `tm gen cap` for `tm generate capabilities`). This propagates automatically to every nested action enum (`ProjectAction`, `SessionAction`, etc).
- Exact matches still win over prefix inference, so the ambiguous-prefix pairs called out in the issue (`hook`/`hooks`, `project`/`projects`, `session`/`sessions`, `status`/`statusline`) keep resolving to their own exact command rather than erroring as ambiguous.
- Scope is deliberately limited to piece (1) from the issue. `-v` for `--version` and an explicit `sess` alias are called out in #4398 as separate one-way-door decisions and are NOT part of this PR.

Closes #4398

## Stacking note
This PR was originally stacked on #4422 (`fix/4397-daemon-guard-bare-invocation`) because both touched the `tm` clap command tree. #4422 has since squash-merged to `main` as `46e080ed` — this PR has been rebased onto `main` and retargeted; the diff now shows only this PR's change (96+/1-).

## Test plan
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p trusty-mpm` (full surface, no `--lib`) — 4009 lib tests + all integration test binaries, 0 failed
- [x] New tests: `cli_infers_abbreviated_subcommands`, `cli_exact_match_wins_over_prefix_ambiguity`

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -11,6 +11,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - `tm` CLI accepts unambiguous abbreviated subcommands, e.g. `tm doc` for `tm doctor` ([#4398](https://github.com/bobmatnyc/trusty-tools/issues/4398))
   - Turns on clap's `infer_subcommands` for the top-level `Cli`, propagating to every nested action enum. Exact matches still win over prefix inference, so ambiguous-prefix pairs (`hook`/`hooks`, `project`/`projects`, `session`/`sessions`, `status`/`statusline`) keep resolving to their own exact command.
+
+### Fixed
+
+- Abbreviated-subcommand inference (#4398) made the hidden `internal-spawn-disclaimed` shim reachable via any unambiguous prefix (`tm int`, `tm inte`, `tm internal`), which would `posix_spawn` an arbitrary program with macOS TCC responsibility disclaimed. `main` now rejects any invocation whose raw first argv token isn't the exact, full subcommand name before the shim ever runs, while the daemon's own literal self-invocation keeps working unchanged.
 - JSON instruction manifest (schema v2) with owner language rules ([#4386](https://github.com/bobmatnyc/trusty-tools/pull/4386)) ([`838d001`](https://github.com/bobmatnyc/trusty-tools/commit/838d00150e4ac1d2f31e5748bf8a85c05b4e78e7))
   - Adds three owner language rules as inline `text` blocks: **Clickable
     References** (`core` — issue/PR/ticket/commit references always render as
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -9,6 +9,8 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- `tm` CLI accepts unambiguous abbreviated subcommands, e.g. `tm doc` for `tm doctor` ([#4398](https://github.com/bobmatnyc/trusty-tools/issues/4398))
+  - Turns on clap's `infer_subcommands` for the top-level `Cli`, propagating to every nested action enum. Exact matches still win over prefix inference, so ambiguous-prefix pairs (`hook`/`hooks`, `project`/`projects`, `session`/`sessions`, `status`/`statusline`) keep resolving to their own exact command.
 - JSON instruction manifest (schema v2) with owner language rules ([#4386](https://github.com/bobmatnyc/trusty-tools/pull/4386)) ([`838d001`](https://github.com/bobmatnyc/trusty-tools/commit/838d00150e4ac1d2f31e5748bf8a85c05b4e78e7))
   - Adds three owner language rules as inline `text` blocks: **Clickable
     References** (`core` — issue/PR/ticket/commit references always render as
--- a/crates/trusty-mpm/src/bin/tm/cli/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/cli/mod.rs
@@ -59,13 +59,29 @@ pub(crate) const DEFAULT_ADDR: &str = trusty_mpm::core::DEFAULT_DAEMON_ADDR;
 /// When invoked without a subcommand (#1708) the guided default fires:
 /// the daemon's in-project spawn path is called for the current directory,
 /// reconnecting to an existing session or provisioning a new worktree.
+///
+/// Why (#4398): `infer_subcommands` lets users type any unambiguous prefix of
+/// a subcommand name (`tm sta` for `status`, `tm proj list` for `project
+/// list`) instead of the full spelling. clap applies this at all 3 nesting
+/// levels (top-level `Command`, and every nested action enum under
+/// `cli::actions`) from this single global setting. It is a separate matching
+/// path from the 5 existing `#[arg(short)]` option-short attributes, so there
+/// is zero collision risk with those. Exact matches always win over inferred
+/// prefixes, so the `hook`/`hooks`, `project`/`projects`, `session`/
+/// `sessions`, and `status`/`statusline` pairs keep resolving to their exact
+/// name rather than erroring as ambiguous. `-v` for `--version` and an
+/// explicit `sess` alias are deliberately out of scope here — see #4398.
+/// What: enables clap's automatic subcommand-prefix inference across the
+/// whole `tm` CLI surface.
+/// Test: `cli_infers_abbreviated_subcommands`, `cli_exact_match_wins_over_prefix_ambiguity`.
 #[derive(Debug, Parser)]
 #[command(
     name = "trusty-mpm",
     version,
     about = "trusty-mpm — unified binary",
     subcommand_required = false,
-    arg_required_else_help = false
+    arg_required_else_help = false,
+    infer_subcommands = true
 )]
 pub(crate) struct Cli {
     /// Base URL of the trusty-mpm daemon (used by the thin CLI subcommands).
--- a/crates/trusty-mpm/src/bin/tm/commands/spawn_disclaimed.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/spawn_disclaimed.rs
@@ -23,6 +23,34 @@
 
 use anyhow::Context as _;
 
+/// Whether the raw process argv shows this shim was invoked by its EXACT,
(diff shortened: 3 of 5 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review of `feat/4398-cli-abbrev-subcommands` (`b7584e26`) against issue #4398. Confirmed programmatically with a standalone clap 4.5.26 probe mirroring the real subcommand names (`install`, `issue`, hidden `internal-spawn-disclaimed`).

### Scope check against #4398
Matches the issue's piece-(1) scoping exactly: `infer_subcommands = true` only, no `-v`/`--version` short, no `sess` alias for `coordinator`. Both deferred pieces are correctly out of scope here.

### Findings

| HIGH | `crates/trusty-mpm/src/bin/tm/main.rs` | 211 (dispatch); declared at `crates/trusty-mpm/src/bin/tm/cli/mod.rs:202` | `infer_subcommands` makes the `hide = true` `internal-spawn-disclaimed` shim — documented as "never a user-facing verb", "not for direct use" — reachable via any unambiguous prefix: `tm int <program> [args…]`, `tm inte …`, `tm internal …` all now silently resolve to `InternalSpawnDisclaimed` and `posix_spawn` the given program with macOS TCC responsibility disclaimed. Verified with a standalone clap probe using the real name set (`install`, `issue`, hidden `internal-spawn-disclaimed`): `int`/`inte`/`internal` → `InternalSpawnDisclaimed{argv:["echo-target"]}`, while `ins` correctly resolves to `Install`. clap's `possible_subcommand` (`clap_builder-4.5.26/src/parser/parser.rs:576`) iterates `get_subcommands()`, which does not exclude `hide`d entries from inference — hidden only suppresses help text, not matching. Neither new test exercises this: both only cover visible/visible sibling pairs. | Add a guard at the `main.rs:211` dispatch site that only proceeds when the literal first CLI argument exactly equals `spawn_disclaim::PANE_DISCLAIM_SUBCOMMAND` (reject/error otherwise), so this internal shim stays reachable only by its exact name even with global inference on. (The 3 other `hide = true` verbs in `SessionAction` — `managed-stop`/`runtime-stop`/`managed-resume` — become abbreviation-reachable too, but those are inert deprecated aliases of `stop`/`resume`, not an arbitrary-spawn primitive, so they don't carry the same risk.) |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/tests.rs` | ~1106-1178 (new `cli_infers_abbreviated_subcommands` / `cli_exact_match_wins_over_prefix_ambiguity`) | Both new tests assert only visible-sibling behavior (`hook`/`hooks`, `project`/`projects`, `session`/`sessions`, `status`/`statusline`). No test asserts what happens to the 4 `hide = true` variants under inference — the exact gap that produced the HIGH finding above. | Add a regression test asserting `Cli::try_parse_from(["trusty-mpm", "int", "echo"])` is rejected (post-fix), so the guard doesn't silently regress later. |

### Other scoped checks (no findings)
- **`tm daemon --force` (#4422)**: unaffected — confirmed via clap source (`possible_subcommand` only runs when resolving a subcommand *name* token; `--force` is parsed through the separate long-flag path) and the pre-existing `cli_parses_daemon_force` test is untouched by this diff.
- **Exact-match-wins mechanics**: traced through clap's `possible_subcommand` (parser.rs:576-608) — the inference iterator finds ambiguous candidates whenever the typed argument equals one subcommand's name AND is a prefix of a sibling (e.g. `"session"` matches both `Session` and `Sessions`), so it falls through to `find_subcommand` for an exact match. This holds for all 4 tested pairs, and the top-level `Cli` doesn't have `args_conflicts_with_subcommands` set, so nothing short-circuits it.
- **Daemon self-invocation**: the only place `tm`/`trusty-mpm` shells out to itself with `internal-spawn-disclaimed` uses the literal `PANE_DISCLAIM_SUBCOMMAND` constant (`core/spawn_disclaim/pane.rs:38`), never an abbreviation — unaffected by inference either way.
- **Scripts/workflows/docs**: spot-checked `scripts/`, `.github/workflows/`, `docs/` — every `tm <subcommand>` call site already uses the full spelling, matching the issue's own audit claim.

### Notes
- The HIGH finding is a design-intent regression (obscurity-based protection removed), not a new privilege-escalation primitive — the hidden command was already invokable by its exact 26-character name pre-PR, and it runs with the invoking user's own privileges. That's why this is WARN rather than BLOCK.

bobmatnyc:
## Verdict: APPROVE (supersedes the earlier WARN)

Targeted re-review of the delta only, `b7584e26` → `5d6043b5`, against the same attack lens as the original HIGH finding.

### Verification of the fix

- **Argv capture ordering**: `main.rs:191` collects `let argv: Vec<String> = std::env::args().collect();` at the very top of `main()`, before `Cli::try_parse()` (`main.rs:196`) ever runs. The guard at `main.rs:211-223` calls `invoked_literally(&argv)` against this pre-parse raw vector, so it inspects exactly what the operator typed at `argv[1]` — clap's inference substitution can't have touched it. Confirmed by reading the file directly (not just the diff).
- **Guard logic**: `commands/spawn_disclaimed.rs`'s new `invoked_literally` is `argv.get(1) == Some(PANE_DISCLAIM_SUBCOMMAND)` — a plain exact-string check, no normalization that could accidentally re-admit a prefix.
- **Ordering vs. the spawn**: the `eprintln!` + `std::process::exit(2)` on rejection happens strictly before `commands::spawn_disclaimed::run(spawn_argv)` is ever called — no partial-spawn/TOCTOU window; a rejected invocation never reaches `posix_spawn`.
- **Single dispatch site**: grepped the whole `crates/trusty-mpm/src` tree — `main.rs:211` is the only live dispatch of `Command::InternalSpawnDisclaimed`; the other match arm at `main.rs:641` is a provably-`unreachable!()` exhaustiveness stub (the early return above always fires first), same as pre-fix.
- **Daemon's literal self-invocation unaffected**: `core/spawn_disclaim/pane.rs` has zero diff between the two heads — `disclaim_pane_command` still emits `"{bin} {PANE_DISCLAIM_SUBCOMMAND} {claude_invocation}"` with no flag before the subcommand token, so `argv[1]` is still exactly `"internal-spawn-disclaimed"` for every real caller and the guard passes it unchanged.
- **New tests match the claim**: 5 new tests — `invoked_literally_accepts_exact_name`, `invoked_literally_rejects_abbreviated_prefixes` (covers `int`/`inte`/`internal`), `invoked_literally_rejects_missing_token` (bare `tm`, no panic on `argv.get(1) == None`), `cli_infers_internal_spawn_disclaimed_but_runtime_guard_rejects_abbreviations` (proves clap *still* infers the abbreviation to the hidden variant AND the runtime guard independently rejects it — the two-layer property that actually matters), and `cli_infers_ins_to_install_unaffected_by_disclaim_guard` (non-regression: the guard doesn't collaterally break ordinary `install` abbreviation). All exercise the right boundary.

### New hole check: global flag before the subcommand token

`Cli::url` is `#[arg(long, env = "TRUSTY_MPM_URL", global = true)]`, so clap would happily parse `tm --url http://x internal-spawn-disclaimed claude` into `Command::InternalSpawnDisclaimed` — but `argv[1]` there is `"--url"`, not the literal subcommand name, so `invoked_literally` would reject that form even though the operator typed the full, correct subcommand name later in argv. This is a real positional-assumption gap (`argv.get(1)` assumes the subcommand token is always first) — but it fails **closed** (denies a legitimate-looking invocation), not open, and `disclaim_pane_command`'s only real caller never prepends `--url` (verified above, zero diff in pane.rs), so no actual caller is affected. Not a blocking finding — noted below for the record, LOW severity (a false-rejection edge case with no exploitable direction and no real caller hitting it).

### Findings

No issues found at >80% confidence that rise above LOW. The HIGH and MEDIUM from the prior review are both closed:
- HIGH (abbreviation reaches the hidden disclaim-spawn shim) — closed by the `invoked_literally` runtime guard, verified above.
- MEDIUM (no test coverage for the hidden-subcommand interaction) — closed by the 5 new tests.

| LOW | `crates/trusty-mpm/src/bin/tm/commands/spawn_disclaimed.rs` | new `invoked_literally` | `argv.get(1)` assumes the subcommand token is always the first CLI argument; a global flag (`--url`) typed before the literal subcommand name would cause a legitimate exact-name invocation to be rejected. No real caller does this today (not worth blocking on) but if `--url` (or any future global flag) is ever prepended to a real `internal-spawn-disclaimed` invocation this would start erroring. | Optional hardening: scan `argv[1..]` for the first non-flag token instead of hard-indexing `argv[1]`, or simply document the constraint inline (already partially done in the doc comment). |

### Notes
APPROVED for merge from this review's perspective. Did not merge or push per instructions.

PR #4432, opened 2026-07-30
fix(trusty-mpm): stop legacy .trusty-mpm/ files from shadowing unrelated CLAUDE.md overrides

### LINKED ISSUE #4399: legacy .trusty-mpm/ override files silently suppress ALL CLAUDE.md named-section overrides
**Defect:** the 1.3.0 architecture decision is "ONE surface — all project customization in named `CLAUDE.md` sections." Not what ships on `main` @ `75f9cf09`. The five legacy `.trusty-mpm/` override files are still live, and their presence doesn't merely coexist with the CLAUDE.md reader — it *shadows it entirely*.

**The sharp edge:** if ANY ONE of the four replace-semantics files (`PM_INSTRUCTIONS_DEPLOYED.md`, `AGENT_DELEGATION.md`, `WORKFLOW.md`, `MEMORY.md`) exists, `resolve_pm_prompt_with_roster` takes the `PromptSource::Legacy` path and drops **every** CLAUDE.md named-section override project-wide — including sections the legacy file has nothing to do with. An unrelated `AGENT_DELEGATION.md` silently kills a `WORKFLOW` override authored in CLAUDE.md.

**Evidence:** file constants at `crates/trusty-mpm/src/core/instruction_overrides.rs:34-45`, read on every call at `:203-353`; overrides discarded via `warn_unapplied` at `:264` and `:354`. Behavior is pinned by an existing test — `named_sections_are_reported_unapplied_on_the_legacy_path`, `crates/trusty-mpm/src/core/claude_md_sections_tests.rs:780-798`. The module doc (`instruction_overrides.rs:1-20,48-51`) states the reader was intentionally scoped without the removal, deferring it to #4183 / #4286, where it is currently unscoped.

**Separately:** `.trusty-mpm/INSTRUCTIONS.md` remains a live additive surface *outside* CLAUDE.md — consumed both as the `project_addendum` generator input (`:42`, `:242`) and as `HOST_FILES[1]` for marker scanning (`core/claude_md_sections.rs:66`). Even after the four are retired, this is a second customization surface the "ONE surface" decision does not account for.

**Resolution:** retire the four replace-semantics files, or demote them to append-only/non-shadowing so a CLAUDE.md named-section override can never be dropped because of an unrelated file's presence. Decide separately what happens to `INSTRUCTIONS.md`. Whichever path is chosen, the failure mode to eliminate is *silent project-wide suppression*.

**Note:** the trusty-tools repo itself currently has only `.trusty-mpm/INSTRUCTIONS.md` and zero `TRUSTY-MPM:` markers in `CLAUDE.md`, so it is not hitting this today — it is one file away.

bobmatnyc:
## Summary

- Any one of the four replace-semantics legacy files (`PM_INSTRUCTIONS_DEPLOYED.md`, `AGENT_DELEGATION.md`, `WORKFLOW.md`, `MEMORY.md`) forced `resolve_pm_prompt_with_roster` onto the sectionless legacy string assembly, which dropped **every** CLAUDE.md named-section override project-wide — including sections the legacy file had nothing to do with. An `AGENT_DELEGATION.md` file could silently kill a `WORKFLOW` override authored in `CLAUDE.md`.
- `WORKFLOW`, `MEMORY` and `AGENT-DELEGATION` are now independently addressable on the legacy assembly too: a CLAUDE.md named override for one of those sections lands unless the project's own same-section legacy file is what forced the branch — that file still wins, unchanged (one file, one winner, same as before).
- `IDENTITY`/`CORE`/`SEARCH` have no independent slot in the opaque `pm_instructions()` blob on this path and remain reported as unapplied via `warn_unapplied` — never silently dropped, consistent with the module's "never fail closed" doctrine.
- The delegation section reuses the same authored/generated split `InstructionPackage::with_overrides` already uses on the packaged path: a named `AGENT-DELEGATION` override replaces the bundled doctrine, but the live agent roster (host-computed, not authored) still follows when any agent is deployed (#4196).

Scope note: retiring the legacy files outright, or deciding the fate of `.trusty-mpm/INSTRUCTIONS.md` as a second customization surface, is explicitly left to a separate decision per the issue — this PR only fixes the silent cross-section shadowing.

## Test plan

- [x] `cargo test -p trusty-mpm` (full suite, fresh build) — 4013 lib tests + all integration test binaries, 0 failed
- [x] Rewrote the pinned regression test that previously asserted the bug (`named_sections_are_reported_unapplied_on_the_legacy_path` → `an_unrelated_legacy_file_does_not_shadow_a_named_section_override`) plus 4 new tests covering: same-section collision still wins, MEMORY override, AGENT-DELEGATION override with roster preserved, and IDENTITY/CORE/SEARCH staying reported-unapplied.

Closes #4399

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -119,6 +119,21 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   additionally filters the surfaced list to non-terminal states as defense in
   depth; and the boot-time reconcile sweep backfills (clears) the stale field
   on any terminal record persisted before this fix.
+- **An unrelated legacy `.trusty-mpm/` override file no longer shadows every
+  CLAUDE.md named-section override (closes
+  [#4399](https://github.com/bobmatnyc/trusty-tools/issues/4399)).**
+  Previously, if ANY ONE of the four replace-semantics legacy files
+  (`PM_INSTRUCTIONS_DEPLOYED.md`, `AGENT_DELEGATION.md`, `WORKFLOW.md`,
+  `MEMORY.md`) was present, the whole prompt fell back to the sectionless
+  legacy assembly, and every CLAUDE.md named-section override was reported
+  "not applied" — including sections the legacy file had nothing to do with
+  (an `AGENT_DELEGATION.md` file silently dropped a `WORKFLOW` override
+  authored in `CLAUDE.md`). `WORKFLOW`, `MEMORY` and `AGENT-DELEGATION` are
+  now independently addressable on that legacy path too: a named override for
+  one of those sections lands unless the project's own same-section legacy
+  file is what forced the branch, in which case that file still wins,
+  unchanged. `IDENTITY`/`CORE`/`SEARCH` have no independent slot in the legacy
+  string assembly and remain reported as unapplied, never silently dropped.
 - **A corrupted bundled agent is re-deployed instead of frozen forever (closes
   [#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408)).** When the
   deployed copy of a bundled agent drifts from the checksum the deploy manifest
--- a/crates/trusty-mpm/src/core/claude_md_sections.rs
+++ b/crates/trusty-mpm/src/core/claude_md_sections.rs
@@ -501,8 +501,21 @@ pub fn strip_marker_blocks(text: &str) -> String {
 /// Quietly is the part that is unacceptable — that is #381 again. The removal of
 /// the legacy files is a later step of #4183; until then this says so, loudly,
 /// once per override.
+///
+/// #4399: the legacy string assembly is NOT entirely sectionless — `WORKFLOW`,
+/// `MEMORY` and `AGENT-DELEGATION` are independently addressable there too, so
+/// the caller ([`crate::core::instruction_overrides::resolve_pm_prompt_with_roster`])
+/// applies a named override for any of those three sections that has no
+/// same-section legacy file shadowing it, and passes THIS function only the
+/// overrides that genuinely could not land — a same-section legacy-file
+/// collision, or `IDENTITY`/`CORE`/`SEARCH`, which have no slot at all on this
+/// path. This function itself is unchanged: it still warns, unconditionally,
+/// about every override in the slice it is given.
+///
 /// What: one `warn!` per accepted override, naming the file and the section.
-/// Test: `named_sections_are_reported_unapplied_on_the_legacy_path`.
+/// Test: `an_unrelated_legacy_file_does_not_shadow_a_named_section_override`,
+/// `a_same_section_legacy_file_still_wins_over_a_named_override_and_is_reported`,
+/// `identity_core_and_search_stay_reported_unapplied_on_the_legacy_path`.
 pub fn warn_unapplied(scanned: &ProjectOverrides) {
     for applied in &scanned.overrides {
         tracing::warn!(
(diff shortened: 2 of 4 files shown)


bobmatnyc:
## Verdict: APPROVE

Adversarial review against issue #4399. Traced the full fix path in `resolve_pm_prompt_with_roster` (per-section `*_has_legacy_file` flags, `named_section` lookup, `delegation_with_named_override`, the `unapplied` filter) and re-ran the affected suites against the PR head (`aa642a3e`) in an isolated worktree:

- `cargo test -p trusty-mpm --lib -- claude_md_sections` → 43 passed, 0 failed
- `cargo test -p trusty-mpm --lib -- instruction_overrides` → 15 passed, 0 failed
- `cargo clippy -p trusty-mpm --lib -- -D warnings` → clean

The core logic is correct: `workflow_has_legacy_file`/`memory_has_legacy_file`/`delegation_has_legacy_file` are captured before the `Option`s are consumed, `named_section()` pulls the CLAUDE.md override per-section independent of which legacy file forced the branch, and a same-section legacy file still wins via `.or()` ordering (legacy first). Verified `delegation_with_named_override`'s hardcoded `"\n\n"` join against the manifest's declared `Join::Blank` for the roster block (`pm-instruction-package.json` block 10, `join_before: "blank"` → `Join::Blank = "\n\n"` in `instruction_package.rs`) — no drift from `InstructionPackage::with_overrides`' authored/generated split. The rewritten test (`an_unrelated_legacy_file_does_not_shadow_a_named_section_override`, `claude_md_sections_tests.rs:780`) is a genuine flip of the old pinned-bug assertions, not a weakened one — it's a strict superset of the old test's checks. No coverage was deleted.

## Findings

| MEDIUM | `crates/trusty-mpm/src/core/instruction_overrides.rs` | 570 | Doc comment on `delegation_with_named_override` cites `Test: \`named_agent_delegation_override_applies_when_an_unrelated_legacy_file_forces_the_legacy_path\`` — this test does not exist anywhere in the repo (confirmed by repo-wide grep). The actual covering test is `an_unrelated_legacy_file_does_not_shadow_a_named_delegation_override` in `claude_md_sections_tests.rs:865`. | Update the doc comment to reference the real test name. |
| MEDIUM | `crates/trusty-mpm/src/core/claude_md_sections_tests.rs` | 815, 902 | `a_same_section_legacy_file_still_wins_over_a_named_override_and_is_reported` and `identity_core_and_search_stay_reported_unapplied_on_the_legacy_path` both claim (in name/comment) to verify the decline is "reported" / "never silently [dropped]", but neither asserts on `warn_unapplied`'s actual output. Their only proxy is `scan_project(...).overrides.len() == 1`, which only proves the override was *parsed* from CLAUDE.md — `scan_project` doesn't know about the legacy-file collision decision made later in `resolve_pm_prompt_with_roster`. If the `unapplied` filter in `instruction_overrides.rs:362-375` had a sign error (e.g. `SectionId::Workflow => !workflow_has_legacy_file`), neither test would catch it — both would still pass. This is a pre-existing pattern (the old test used the identical proxy), not introduced fresh by this PR, but the new/renamed tests raise the claimed guarantee without closing the verification gap. | Either capture `tracing` output in these two tests (a subscriber that records the `warn!` call's `section` field) or refactor so the filtered `unapplied` set is a return value/testable seam instead of only a side-effecting log call. |

## Notes

- Both findings are documentation/test-coverage gaps, not functional bugs — I traced the production logic by hand (per-section flags computed before any `Option` is moved/shadowed, `named.overrides` left untouched for the final filter) and it is correct for all combinations of legacy-file-present × named-override-present across the three addressable sections.
- Did not flag the pre-existing weak-verification pattern as a blocking issue since it's inherited from the test this PR replaces, not newly introduced — surfacing it here per the task's request to scrutinize the rewritten test specifically.

Issue #2149, opened 2026-07-07
fix(trusty-mpm): managed sessions can launch with no roster and wrong framework identity — provisioning aborts silently; identity must survive foreign CLAUDE.md


bobmatnyc:
## Symptom
A `tm`-managed session spawned into a foreign project (an "apex" repo with its own `.claude-mpm/` config) came up with (a) NONE of the tm/mpm agent roster deployed, and (b) self-identified as "Claude MPM" (the unrelated Python project) instead of trusty-mpm when asked "what framework are you running under?". Daemon confirmed CURRENT (0.16.2, built after DOC-34/#1913), so this is NOT a stale-binary issue.

## Root cause (traced, read-only)
1. **Provisioning pipeline aborts before the identity/output-style write.** `prepare_session_inner` (`crates/trusty-mpm/src/core/session_launch/mod.rs:239-520`) runs, in order: deploy agents → deploy skills → CLAUDE.md stub → **write trusty-mpm output-style (carries the Identity & Self-Awareness Protocol)** → MCP/hooks → deploy output-style file. If step 1 `deploy_agents_filtered` (`core/agent_deployer.rs:92-186`) returns `Err`, the function early-returns at ~mod.rs:283 and **skips the output-style/identity write entirely**. The caller swallows it as a best-effort `warn` and launches the session anyway (`provisioner/workspace.rs:848-854`, `daemon/managed_routes/lifecycle.rs:638-644`). Result: no roster AND no identity, silently. (DOC-28 F4 silent-instruction-load-failure class.)
2. **Identity delivery is easy to lose and does not assert precedence.** For managed sessions, identity is delivered ONLY via the native output-style file + `.claude/settings.json` `outputStyle` key (`session_launch/settings.rs:200-236`, `deploy_output_style` mod.rs:479-485) — there is no `--append-system-prompt-file` on the managed spawn path (`runtime/claude_code.rs`). A foreign project's own `CLAUDE.md` loads with stronger "these instructions OVERRIDE" framing than the output-style, so it can win on a direct identity question.

## Fix (agreed — scope decisions from owner)
1. **Decouple deploy-failure from identity-write:** a `deploy_agents_filtered` (or skills) error must NOT skip the output-style/identity + MCP/hooks steps. Continue the pipeline; surface the roster failure loudly but still install identity. (`session_launch/mod.rs:239-520`.)
[...]

Issue #1927, opened 2026-07-02
bug(trusty-mpm): standalone `tm run` load_alias deploys agents/skills to real ~/.claude, violating SPEC-STANDALONE-MPM-04 isolation invariant


bobmatnyc:
## Problem
`run_prepare_session` in `crates/trusty-mpm/src/core/standalone/load.rs:175-180` constructs `FrameworkPaths::default()` (real `$HOME/.claude`) instead of an isolated base. So even the "fully isolated" standalone driver (`tm register/load/run`, DOC-24) deploys agents and skills into the user's **real global** `~/.claude/agents` / `~/.claude/skills`.

This directly violates:
- `SPEC-STANDALONE-MPM-04` isolation invariant (docs/specs/standalone-managed-trusty-mpm.md:234-238, 277-278) — tm writes nothing to the user's real global `~/.claude`
- The 2026-07-02 segregation design directive (tm environments fully segregated from user's live config; confirmed cross-tool pollution incident via shared global `~/.claude/agents/`)

## Expected
Per DOC-24: project-local content goes to `<managed-root>/<alias>/repo/.claude/{agents,skills}` (native discovery); tm-global content goes to the isolated `CLAUDE_CONFIG_DIR` (`~/.trusty-mpm/claude-config/`), which `tm run` already injects (core/standalone/run.rs:58,92). `load_alias`'s prepare step must target those, never `FrameworkPaths::default()`.

## Notes
- Found during research for the upcoming system/project agents+skills provisioning spec (DOC-31); that spec builds on this invariant actually holding, so this fix should land first.

🤖👥 Generated with [Claude MPM](https://github.com/bobmatnyc/claude-mpm)

PR #4437, opened 2026-07-30
feat(trusty-mpm): deploy bundled agents to the tm-managed user tier, not per-workspace

### LINKED ISSUE #4409: deployment has drifted from the day-1 design: bundled assets belong at user level in a protected config path

bobmatnyc:
Closes #4409.

Bundled agents now deploy to exactly one place: the tm-managed `CLAUDE_CONFIG_DIR/agents` (`~/.trusty-tools/trusty-mpm/claude-config/agents`). They no longer land in a workspace's `.claude/agents/` or in the operator's generic `~/.claude/agents/`.

Implements the settled design on the issue (maintainer ruling 2026-07-30, plus the probe matrix in [comment 5136146155](https://github.com/bobmatnyc/trusty-tools/issues/4409#issuecomment-5136146155) that proved the config-dir tier is discovered by managed sessions at session start and mid-session). No design decisions were reopened.

## What changed

**One destination, one accessor.** `FrameworkPaths::agent_deploy_dir()` is the single target every bundled-agent deploy / validate / staleness / reset / doctor call site now uses. Unlike `claude_agents_dir()`, `for_managed_project` / `for_managed_workspace` deliberately do **not** rewrite it project-local — the deploy tier is global on purpose. It resolves from the same base the rest of `FrameworkPaths` uses, so `under(tempdir)` stays hermetic.

Retargeted call sites:

| Site | Was | Now |
|---|---|---|
| `tm install` deploy + `--reset-agents` | `~/.claude/agents` | `agent_deploy_dir()` |
| `prepare_session_inner` (session launch) | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| `sync_session_assets` (`tm sessions sync-assets`) | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| `apply_catalog` (`tm catalog apply`) + prune | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| `deploy_validate::validate_agents` | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| agent staleness (`update_check`, `session_assets`) | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| `tm doctor` `agents` / `agent_skills` probes | `<workspace>/.claude/agents` | `agent_deploy_dir()` |
| instruction-pipeline roster scan | `<workspace>/.claude/agents` | `agent_deploy_dir()` |

Skills are untouched — they still deploy per-workspace to `.claude/skills/`. Only agent discovery from the config-dir tier was proven; extending this to skills is a separate question.

**Retraction is included, and it is load-bearing — not cleanup polish.** Stopping the writes alone would have made things worse. Every workspace provisioned by an older binary still holds a full copy of the bundled roster, and the project tier *outranks* the config-dir tier in agent resolution. Left in place, those copies would shadow the canonical roster permanently, and nothing would ever refresh them again — #4408's shadowing incident made irreversible.

New `trusty_agents_common::agents::deployer::retract_framework_agents`, called from session launch and sync-assets against the workspace tier:

- removes exactly the manifest-tracked entries whose origin `is_framework_owned()` — the same #4408 predicate, not regressed;
- **never** touches an untracked file (hand-placed agents are invisible to it by construction);
- keeps user-owned (`Origin::User`/`Registry`, seed-once tier) entries and reports them as preserved;
- does **not** consult the checksum: drift on a framework-owned file is corruption, not ownership, so the #4408 stub gets cleared rather than frozen;
- refuses outright on a corrupt manifest — deleting files on the strength of an unreadable ledger is exactly what the manifest exists to prevent;
- deletes the ledger (and the directory, if empty) when nothing managed remains, so a workspace returns to pristine;
- is idempotent and non-fatal at both call sites.

**Corrected a stale doc claim.** `managed_config.rs` asserted that `--setting-sources project,local` excludes the `user` tier `CLAUDE_CONFIG_DIR` relocates, so the config-dir roster was "belt-and-suspenders" and the project layer delivered the agents. The #4409 probe matrix refutes that for agents: a probe existing **only** in `$CLAUDE_CONFIG_DIR/agents/` resolved and executed both at session start and mid-session. `~/.claude/agents/` is what goes unread, which is documented behavior. Comment corrected to match the evidence.

## Scope decisions

- **`tm install`'s `~/.claude/agents` target is included.** The issue grades that the highest-severity breach (47 files in a Claude Code install with nothing to do with trusty-mpm), and it is the same one-line change through the same accessor. Excluding it would have left the segregation invariant half-enforced.
- **Retraction is scoped to workspaces tm re-prepares.** It runs on session launch and sync-assets, so a workspace is cleaned the next time it is prepared. There is no bulk sweep of *every* historical workspace on disk, and nothing retracts the ~47 files already in `~/.claude/agents/` — that is the operator's own directory, and deleting from it unprompted is not something this PR should do. `tm install --reset-agents --reset-agents-workspaces` still exists as the operator-driven path for stragglers.
- **`tm doctor` boundary check (issue item 3) is NOT in this PR.** "Fail when tm-owned assets are found outside `$CLAUDE_CONFIG_DIR`" is a genuinely separate probe with its own design questions (what counts as tm-owned outside a manifest? how does it interact with hand-placed files?), and folding it in would have doubled the diff. Worth its own issue.
- **Known consequence, stated plainly:** the agent deploy tier is machine-global, so a per-project `[agents] exclude` no longer removes an agent from a session's view — it only refrains from writing it, and a sibling project that selects the agent still puts it there. Per-project agent *selection* against a shared tier is not solved here. This follows directly from the ruling (bundled agents live at a global user tier) rather than from an implementation choice, so it is recorded in the CHANGELOG rather than worked around.

## Test changes worth reviewing

Two existing tests encoded premises the flip invalidates, and were rewritten rather than deleted:

- `agents_check_scopes_to_managed_workspace_when_project_given` (#2149) asserted the doctor probe reads the workspace tier. It now asserts the probe names the tm-managed tier and *not* the workspace tier. The status is deliberately not asserted — the deploy tier is machine-global, so a provisioned workstation and a bare CI runner legitimately disagree; `doctor_fs_checks`'s `agents_*` tests cover every status branch hermetically. The expected path is matched as a relative suffix because sibling `#[serial]` tests move `$HOME`.
- `checked_summaries_stale_assets_independent_per_session_sharing_one_catalog` (#2444) drove per-session staleness divergence through agents, which two sessions can no longer disagree about. Rewired to skills, which are still per-workspace, so the property it pins (batched catalog sharing must not cross-contaminate per-session verdicts) stays real.

`run_prepare_session_never_writes_real_home_claude_dirs` (#1927) now additionally asserts `repo/.claude/agents` does **not** exist — the isolation invariant it guards got stricter, not weaker.

## Verification

```
cargo fmt --check                    → clean
cargo clippy --workspace --all-targets --exclude trusty-mpm-gui \
  --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
                                     → exit 0
cargo test -p trusty-mpm -p trusty-agents-common
                                     → 46 test binaries, 0 failures
                                       trusty-mpm lib:          4012 passed; 0 failed; 6 ignored
                                       trusty-agents-common:     336 passed; 0 failed; 0 ignored
                                       trusty-mpm bins/integration: all ok
scripts/check_line_cap.sh            → 7 allowlisted, 0 violations
```

New coverage: seven retraction tests in `deployer_tests.rs` (framework-owned-only removal, hand-placed survival, drifted-file removal, pristine-return, idempotence, missing-dir no-op, corrupt-manifest refusal), `sync_assets_retracts_workspace_bundled_agents` (end-to-end, including hand-placed survival), and three `FrameworkPaths` tests pinning that the deploy tier is the managed config dir, is never `<base>/.claude/agents`, and is never rewritten project-local.


bobmatnyc:
## Code Critic Verdict: BLOCK

Reviewed `a82ada1c` against the #4409 ruling + probe-matrix spec. The design is right and the retraction primitive itself is well-built — `retract_framework_agents` provably cannot touch an untracked file, cannot touch a user-owned entry, and refuses on a corrupt ledger. The problem is the **wiring**: the retraction is aimed at `fw.claude_agents_dir()`, and on two production paths that field is the operator's `~/.claude/agents`.

The PR's own suite is green (`trusty-mpm` lib 4012 passed / 0 failed, `trusty-agents-common` 336 passed / 0 failed at this head). The CRITICAL below lives on a path no test covers.

### Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 588 | `retract_framework_agents(&fw.claude_agents_dir())` deletes the bundled roster, the manifest, and the directory itself from the operator's `~/.claude/agents` whenever `fw` is a home-tier `FrameworkPaths::default()` — which is exactly what `tm session start` (non-git dir) and `DaemonClient::launch_session` pass. Contradicts the PR body's stated scope. | Target the workspace tier explicitly: `retract_framework_agents(&project_dir.join(".claude").join("agents"))`. Identical on the managed path (that is what `for_managed_workspace` sets), removes the home-tier hazard. |
| CRITICAL | `crates/trusty-mpm/src/core/session_launch/sync_assets.rs` | 129 | Same call, same defect, on `sync_session_assets`. | Same fix. |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/agent.rs` | 78, 105, 140 | `tm agent list` / `tm agent show` still read `paths.claude_agents_dir()` with `FrameworkPaths::default()` (line 29). Nothing deploys there after the flip, so `tm agent list` reports an empty roster and `tm agent show engineer` errors `not found in ~/.claude/agents`. Missed call site — not in the retarget table. | Switch all three to `paths.agent_deploy_dir()`. |
| HIGH | `crates/trusty-mpm/src/core/agent_reset_workspace.rs` | 260 | Still `reset_project_agents(&plan.agent_source, &fw.claude_agents_dir(), …)`, which **force-recomposes** the bundled roster back into `<workspace>/.claude/agents`. The new module doc claims this sweep is now "a no-op" that "reconciles" legacy workspaces; it does the opposite — it re-creates the project-tier shadow, into workspaces of **live** sessions. The PR body points operators at this exact command for stragglers. | Call `retract_framework_agents(&fw.claude_agents_dir())` here instead of `reset_project_agents`, or drop the agent half of the sweep. Correct the module doc either way. |
| HIGH | `crates/trusty-agents-common/src/agents/manifest.rs` | 79-84, 260-266 | The deploy target went from per-workspace to **one machine-global directory** shared by every concurrent session launch, sync-assets run, and `tm catalog apply` — but `save` is an unlocked load→modify→write and `atomic_write` stages through a **fixed** sibling `path.with_extension("tmp")` = `<dir>/.trusty-mpm-manifest.tmp`. Two concurrent writers interleave into that one temp file and `rename` can publish torn JSON (→ `AgentManifestCorrupt` in the spawn/resume gate, and `retract_framework_agents` then refuses); absent tearing, last-writer-wins drops the other's entries, and `deploy_agents_filtered`'s untracked branch (`deployer.rs:268-290`) then classifies those files `untracked_modified` and skips them — the #4408 freeze shape, reachable by a race. | Per-process temp name (`.trusty-mpm-manifest.<pid>.tmp`) plus an advisory lock held across load-modify-save in the deploy dir. |
| MEDIUM | `crates/trusty-mpm/src/core/update_check/apply.rs` | 153, 169 | `--prune` now deletes from the one directory every live session reads, mid-session, driven by the daemon-wide baseline manifest (`managed.rs:899` passes `&framework_root` as `project_dir`). Pre-flip it pruned `~/.claude/agents`, which no managed session read. The CHANGELOG's "known consequence" paragraph records only that `exclude` no longer removes an agent from view — not that `--prune` now removes agents from *every* session's view. | Extend the CHANGELOG note, and gate prune on the deploy tier being unshared (or make it require `--force`). |
| LOW | `crates/trusty-agents-common/src/agents/deployer.rs` | 130-139, 168 | Partial-failure: files are removed one at a time and the in-memory ledger pruned, but the manifest is saved only after the loop. An `Err` at line 135 (e.g. `EACCES` on file 4 of 6) leaves 3 files deleted and the on-disk ledger still claiming all 6. Self-heals on the next run (`NotFound` tolerated at 134), but the window is avoidable. | Save the pruned manifest before returning `Err`, or delete-then-save per entry. |
| LOW | `crates/trusty-mpm/src/core/session_launch/sync_assets.rs` | 129 | Retraction failure is a bare `tracing::warn!` and never reaches `SyncAssetsReport`, while the `prepare_session` path pushes the same failure into `roster_errors` (`mod.rs:909`). Asymmetric observability on the one signal that tells you a workspace is still shadowing. | Add a `retraction_error: Option<String>` to `SyncAssetsReport`. |

### Reproduction for the CRITICAL

Added at PR head, in `session_launch/tests.rs`, then reverted:

```rust
let mut fw = FrameworkPaths::under(tmp_home.path());   // == FrameworkPaths::default() with HOME=tmp_home
fw.trusty_mpm_root = None;
seed_bundled_agents(&fw);
let home_agents = fw.claude_agents_dir();               // <home>/.claude/agents
deploy_agents(&fw.agents, &home_agents).unwrap();       // a pre-#4409 `tm install`
prepare_session(&fw, project).unwrap();
```

```
PROBE: ~/.claude/agents still exists? false ; rust-engineer.md? false
panicked: PR body claims nothing retracts ~/.claude/agents — but prepare_session
          with FrameworkPaths::default() deleted it
```

The whole directory is gone, not just the files. The two production callers that reach this:
- `crates/trusty-mpm/src/bin/tm/commands/session/start.rs:85` → `start_session_in_place` → `prepare_session(fw, path)`
- `crates/trusty-mpm/src/client/http_client/session_connect.rs:39-40` (TUI `/connect`)

`run_prepare_session_never_writes_real_home_claude_dirs` doesn't catch it because the standalone path passes a *managed* `fw`; no test exercises `prepare_session` with a home-tier one.

### Required before re-review

1. Aim both retraction calls at the workspace path explicitly, not at `fw.claude_agents_dir()`.
2. Add a test pinning that `prepare_session` with a home-tier `FrameworkPaths` leaves `<home>/.claude/agents` byte-identical.
3. Retarget `tm agent list` / `tm agent show` to `agent_deploy_dir()`.
4. Fix or remove the workspace reset sweep's agent half, and correct the module doc that claims it is a no-op.
5. Make the manifest write concurrency-safe now that the target is machine-global.
6. Record the `--prune` blast-radius change in the CHANGELOG alongside the `exclude` note.

### Checked and clean

- **Retraction safety proper.** Only manifest keys are iterated, only `Origin::is_framework_owned()` (`Bundled`) entries are removed, `ManifestLoad::Corrupt` returns `Err` before any deletion. Untracked files are invisible by construction. All three properties hold.
- **Manifest predating `Origin`.** `ManifestEntry.origin` has no `#[serde(default)]`, so a manifest lacking the field fails to parse → `Corrupt` → refuse. Fails closed; the workspace keeps shadowing but the failure lands in `roster_errors` on the prepare path.
- **Retraction call sites.** `prepare_session_inner` and `sync_session_assets` are the only two, and every managed/standalone/provisioner/repair entry point funnels through `prepare_session_with_repo_url`. No third path. `apply_catalog` doesn't retract, which is fine — prepare/sync cover it.
- **`checked_summaries_stale_assets_independent_per_session_sharing_one_catalog`.** Still asserts both directions (`summaries[0].stale_assets` and `!summaries[1].stale_assets`); the per-session-divergence property is genuinely preserved via skills, not weakened.
- **`agents_check_probes_the_managed_config_tier_not_the_workspace`.** Dropping the status assertion is defensible: the probed path and the message both derive from the same `paths.agent_deploy_dir()` call inside `check_agents`, so they cannot diverge, and `doctor_fs_checks`'s `agents_*` tests cover every status branch hermetically.
- **`plan_selection_excludes_generic_engineer` / `expected_set`.** Operate on `HarnessPlan` selection, independent of the deploy directory. Unaffected by the flip.

bobmatnyc:
## Code Critic Verdict: WARN — supersedes the BLOCK on `a82ada1c`

Re-reviewed at `588ae56b`. **The CRITICAL is closed, and I confirmed it independently rather than taking the new test's word for it** — I re-ran my original v1 probe scenario verbatim against this head:

```
PROBE v2: ~/.claude/agents exists? true ; rust-engineer.md? true ; manifest? true
PROBE v2: workspace .claude/agents retracted? true
```

Same scenario that wiped the directory at v1. The home tier now survives intact *and* the workspace retraction still fires — the fix didn't buy safety by disabling the feature. Binding the target to `project_dir.join(".claude/agents")` (`session_launch/mod.rs:599`) rather than the `fw` field is the structurally right call, and the comment at 589-598 explains why to the next reader.

Suite green at this head: `trusty-mpm` lib 4020 passed / 0 failed, `trusty-agents-common` 342 passed / 0 failed, all binaries ok.

Three HIGH findings remain — two of them are the exact question the fix round was supposed to close.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/agent_reset.rs` | 151-239 | `reset_agents` does an unlocked `load_checked` → mutate → `save`, and `install.rs:107-110` now points it at `paths.agent_deploy_dir()`. So `tm install --reset-agents` performs an **unlocked** read-modify-write on the one machine-global ledger the new lock exists to protect, and can lose-update a concurrent session launch's `deploy_agents_locked`. This is the highest-traffic writer that races live sessions. | Wrap the body in `with_agent_manifest_lock(target_dir, || …)`, mirroring `deploy_agents_locked` / `retract_locked`. |
| HIGH | `crates/trusty-mpm/src/core/update_check/apply.rs` | 170-193 (and `prune_skills`, 195-230) | `prune_agents` is an unlocked `AgentManifest::load` → mutate → `save` against `agent_target = fw.agent_deploy_dir()`. Worse, `apply_catalog` takes the lock for the deploy (step 3) and then **drops it** before pruning (step 4, line 152) — a session launching in that window is clobbered. | Hold one `with_agent_manifest_lock` around steps 3+4, or at minimum wrap `prune_agents`. |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/repair.rs` | 139-148 | The new per-attempt temp name breaks `tm repair deploy`'s orphan cleaner, **and it reports success anyway**. `remove_tmp_orphans` derives `base = orphan.with_extension("")` then calls `repair_stale_tmp(base)`, which does `base.with_extension("tmp")` — a round-trip that only reconstructs the *old* fixed name. It then pushes the path onto `removed` unconditionally regardless of whether anything was deleted. | Delete `entry.path()` directly — the exact path is already in hand and the `base` round-trip is pointless indirection. Push to `removed` only on a successful unlink. Then either drop `repair_stale_tmp` or reimplement it as a prefix scan. |
| MEDIUM | `crates/trusty-mpm/src/core/agent_reset.rs` | 308 | `reset_project_agents` now has **zero production callers** — the sweep was its only one. It remains `pub`, so no dead-code warning fires. A public function that force-recomposes the bundled roster into an arbitrary directory is precisely the footgun #4409 exists to remove; leaving it callable invites reintroducing the shadow. | Delete it and its three tests, or make it `pub(crate)` with a doc line stating it has no callers and why. |
| LOW | `crates/trusty-mpm/src/core/agent_reset_workspace.rs` | 85, 88, 111, 366 | Doc references left over from the rewrite: `[reset_project_agents]` (85, 111) — the sweep no longer calls it — and `sweep_resets_intact_workspace` (88, 366), which is now `sweep_retracts_intact_workspace`. | Retarget to `retract_framework_agents_filtered` and the real test name. |

Proof for the repair finding, run at this head:

```
target      = .trusty-mpm-manifest.json
  orphan       = .trusty-mpm-manifest.json.54255.1785451476803813000.tmp
  reconstructed= .trusty-mpm-manifest.json.54255.tmp
  MATCH        = false
  orphan still on disk after repair = true
target      = engineer.md
  orphan       = engineer.md.54255.1785451476805891000.tmp
  reconstructed= engineer.md.54255.tmp
  MATCH        = false
  orphan still on disk after repair = true
```

`Path::with_extension` replaces only the last dot-segment, so the two-segment suffix cannot survive the round-trip. Under the old fixed naming it did (`engineer.tmp` → `engineer` → `engineer.tmp`), which is why the existing test still passes: `atomic_write_leaves_old_intact_on_interrupted_write` (`manifest.rs:490-516`) hand-writes `path.with_extension("tmp")` rather than calling `temp_path`, so it models a filename the code no longer produces.

### Verified correct

- **CRITICAL closed**, independently reproduced (above). `prepare_session_never_retracts_the_operator_home_agents_tier` is a real regression test, not a decoration: it has a vacuity guard on the fixture, and asserts directory survival, exact file-set equality, byte-identity, and manifest survival.
- **`tm agent list`/`show` retarget** (`agent.rs:80, 109, 144`) is complete, and `deployed_agent_names_reads_the_managed_agent_tier` is a genuine decoy — it plants a file in the *old* location and asserts it is not listed. That would have caught the original bug.
- **The sweep rewrite** is the right inversion. `retract_framework_agents_filtered`'s `is_dir()` check *before* taking the lock (`deployer.rs:494-499`) is a subtle catch worth calling out — locking first would have made `with_agent_manifest_lock`'s `create_dir_all` materialise an empty `.claude/agents/` in every workspace on every launch.
- **Both flipped sweep tests hold up.** `sweep_ignores_the_project_manifest_exclude` inverts #2462/#2508 for a sound reason (a shadowing copy of an *excluded* agent is more urgent to remove, not exempt), and it still drives a real project manifest rather than asserting on a stub. `sweep_serves_live_linked_worktree_with_git_file` keeps its anti-over-refusal purpose intact — it still asserts `skipped_reason == None` and still pins the `.git`-is-a-FILE fixture; only the effect assertion moved from recomposed to retracted. Neither weakened a property.
- **Sidecar-unlink-last reasoning is correct.** `only_the_lock_sidecar_remains` (`deployer.rs:611-629`) is the right answer to a trap that would otherwise have silently killed the return-to-pristine behaviour the moment locking landed. Unlinking while holding the `flock` is safe on Unix as documented.
- **Readers cannot tear** — confirmed. Writers publish via `rename`; a reader mid-`read_to_string` keeps reading the old inode, which is complete. A reader can see a *stale* manifest, but every read-only consumer (`check_agents`, `validate_agents`, `deployed_agent_names`) only reports, never writes back, so a stale read is a transient wrong report and not a lost update.
- **No lock reentrancy hazard.** No locked path calls another locked path; deploy and retract are sequential and target different directories. (Worth keeping in mind for future work: `flock` on a second fd in the same process self-deadlocks, so nesting `with_agent_manifest_lock` on one directory would hang.)
- **Both prior LOWs fixed properly.** `save_pruned` (`deployer.rs:643-655`) narrows the partial-retraction window and correctly swallows its own error so it cannot mask the original; `SyncAssetsReport.retraction_error` (`sync_assets.rs:92`) closes the observability asymmetry with its own test.
- **`--prune` recording is honest and specific** — it names the mid-session blast radius, the daemon-wide baseline manifest as the driver, and the pre-flip contrast. Not gating behind `--force` is a defensible scope call for this PR. Worth a follow-up issue, not a blocker.

### Note on the diff

`pr4437-v1..pr4437-v2` shows ~30 files, but most is rebase noise from `bbb3a009` (`session_manager/*`, `supervisor/*`, `spawn_disclaimed.rs`, `daemon/state/core.rs`). The author's actual change-set is 35 files against the new base and is disjoint from that noise; I reviewed the delta, not the rebase.

### Verdict

**WARN** — zero CRITICAL, three HIGH. Per the protocol this proceeds with the finding table attached to the next handoff. My own read: the two lock gaps are worth closing in this PR rather than tracking, since the lock was added *in* this PR and shipping it with two bypasses around the shared directory it guards leaves the stated invariant only partly true. The repair-orphan finding is the one I'd least want to ship — a repair command that reports removing files it did not remove is the failure shape this repo has been bitten by before.

bobmatnyc:
## Code Critic Verdict: APPROVE — supersedes the WARN on `588ae56b` and the BLOCK on `a82ada1c`

Reviewed `862fb388`. All three HIGH findings are closed, the MEDIUM is closed, and the LOWs are closed. Zero CRITICAL, zero HIGH remaining. One MEDIUM below is **record-and-go, not merge-blocking**.

Independently verified rather than taken on report:

- **Original CRITICAL scenario, third re-run at this head** — `PROBE v3: home tier intact? true` / `workspace shadow retracted? true`. The operator's `~/.claude/agents` survives and the workspace shadow is still cleared.
- **Gate reproduced locally**: 46 test binaries, `trusty-mpm` lib 4021 passed / 0 failed, `trusty-agents-common` 341 passed / 0 failed, integration 1319 passed / 0 failed. Matches the reported numbers exactly.
- **Lock coverage is now total.** Every non-test `AgentManifest::save` against the shared tier sits inside a `with_agent_manifest_lock` critical section: `deployer.rs:413` (`deploy_agents_locked`), `:603` (`retract_locked`), `:648` (`save_pruned`, called from `retract_locked`), `agent_reset.rs:271` (`reset_agents_locked`), `apply.rs:228` (`prune_agents_locked`). The one remaining unlocked save, `apply.rs:265`, is `prune_skills` writing a `SkillManifest` to the still-per-workspace `claude_skills_dir()` — correctly out of scope.

### On the deviation you flagged for judgement

**The split deploy/prune lock spans satisfy the finding.** My wording was "at minimum wrap `prune_agents`", and that is what landed — but the reasoning deserves more than a pass mark, because the deviation is the better call and not merely the cheaper one.

The lost-update hazard requires *load outside, save inside* — a writer committing on top of state it read before another writer's commit. After this change all four agent-ledger writers (deploy, retract, reset, prune) load **and** save within a single lock hold, so every one of them observes the latest committed ledger. A merged deploy+prune span would close a different window — the interleaving one — at the cost of exposing the deployer's unlocked inner entry point, and `flock` is not reentrant, so any caller who took the lock and then reached for the ordinary `deploy_agents_filtered` would self-deadlock. That is a strictly worse hazard than the one it removes, and it would be latent rather than immediate.

The residual window is real but bounded exactly as claimed: a session launching between step 3 and step 4 can deploy an agent this prune then removes, and it returns on the next launch or sync-assets because deploy is idempotent. One redundant redeploy, never a lost entry. The tradeoff is documented in the code at `apply.rs:322-329` where the next reader will find it. Accepted.

### Remaining finding

|---|---|---|---|---|
| MEDIUM | `crates/trusty-agents-common/src/agents/deployer.rs` | 524 | `retract_locked`'s corrupt-manifest error tells the operator to "Run `tm repair deploy` to recover", but that error fires for the **workspace** ledger (`<project>/.claude/agents`) — and `tm repair deploy` now repairs only `agent_deploy_dir()` and `~/.claude/skills` (`repair.rs:45-46`). So the one advertised remedy does not cover the directory that failed. Consequence: retraction stays permanently refused for that workspace, and the stale project-tier copies keep shadowing the canonical roster — the exact outcome #4409 exists to prevent. The sibling messages at `deployer.rs:225` and `agent_reset.rs:188` are correctly aimed; only this one is misdirected. | Either name the directory and the manual remedy in the message (`refusing to retract; delete <dir>/.trusty-mpm-manifest.json to recover`), or teach `repair_deploy` a workspace argument. The message change is the smaller, sufficient fix. |

Not merge-blocking: the precondition is rare (a corrupt workspace ledger), the failure is surfaced loudly through two channels already (`PrepReport::roster_errors` and `SyncAssetsReport::retraction_error`, both added in this PR), and a one-line `rm` clears it. Worth a follow-up issue rather than another round.

### Round-2 fixes verified

- **`reset_agents`** (`agent_reset.rs:141-159`) — wrapper + `reset_agents_locked`, with the `source_dir.is_dir()` pre-check kept *outside* the lock so a no-op reset neither blocks nor creates a sidecar. Matches the `deploy_agents_filtered` / `retract_framework_agents_filtered` pattern exactly.
- **`prune_agents`** (`apply.rs:198-201`) — wrapper + `prune_agents_locked`; the `From<ManifestError> for ApplyError` bridge is scoped correctly, since prune is the only stage here that takes the lock directly.
- **Orphan cleaner** (`repair.rs:158-174`) — unlinks `entry.path()` directly, pushes to `removed` only on a real unlink, tolerates `NotFound` as a race, and propagates other errors. `repair_stale_tmp` is gone rather than left as a trap. Both new tests are genuine: `repair_removes_per_attempt_temp_orphans` builds its fixtures from the real `temp_path` so it cannot drift from the writer, and asserts a deployed `engineer.md` survives; `remove_tmp_orphans_reports_only_what_it_unlinked` pins the count half of the bug, which was the part that made the old code lie. Rewiring `atomic_write_leaves_old_intact_on_interrupted_write` to call `temp_path` instead of hand-writing the name closes the gap that let the stale assumption survive undetected in the first place.
- **`temp_path` made `pub`** — justified, and the doc's note that "there is no such thing as *the* temp path for X" is the right thing to tell the next person who reaches for a derivation.
- **`reset_project_agents` deleted** with its three tests and every doc reference. `ResetResult.deselected` retained for report shape and documented as unpopulated — honest, and cheaper than reshaping the report.
- **`tm repair deploy` agent-half retarget** (`repair.rs:45`) is correct and complete for its scope. It has to follow the deploy tier, because the `--force` corrupt-manifest reset and the deployer's own recovery guidance both concern the ledger in that directory. Skills correctly stay at `~/.claude/skills`, which is still a live deploy target on the home-tier path and untouched by #4409.

### Review history

| Head | Verdict | Findings |
|---|---|---|
| `a82ada1c` | BLOCK | 2 CRITICAL (retraction deleted `~/.claude/agents`), 3 HIGH, 1 MEDIUM, 2 LOW |
| `588ae56b` | WARN | CRITICAL closed; 3 HIGH (2 lock gaps, orphan cleaner reporting false success), 1 MEDIUM, 1 LOW |
| `862fb388` | **APPROVE** | 1 MEDIUM, record-and-go |

Each round's fixes were verified by re-running the original failing scenario, not by reading the new tests. Clean to proceed to the next pipeline stage.

PR #4438, opened 2026-07-30
docs(trusty-installer): amend isolation-harness §7.3 to bash, fix stale crate name

mac-duetto:
## Summary
- Owner-approved reversal of §7.3 ("Where the harness lives") in `docs/trusty-installer/research/02-design/10-isolation-testing-harness.md`: the scenario driver **and** assertions move from an `#[ignore]`-tagged Rust integration test in-workspace to **bash under `scripts/vmtest/`**, outside the Cargo workspace, run ad-hoc/manually rather than via `cargo test`. Reason: root `Cargo.toml` uses `members = ["crates/*"]`, so anything under `crates/` is auto-swept into `cargo test --workspace`, `cargo clippy --all-targets -- -D warnings`, the 500-SLOC `.rs` line-cap lint, the test-pointer lint, SLD lint, and the publish-guard — a Rust test crate under `crates/` cannot satisfy the owner's separation requirement, but shell is unencumbered by all of those gates. The JSON-only assertion oracle and `cargo install`-only (never `cp`) constraints are preserved and re-emphasised. A `.cirrus.yml` wrapper is noted as a possible future option (not used now).
- Fixes every stale `trusty-controller` reference document-wide (renamed `trusty-installer` per [ADR-0013](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0013-rename-trusty-controller-to-trusty-installer.md), Accepted 2026-06-26, which supersedes ADR-0006) — crate name, `crates/trusty-controller/...` paths, and `cargo test -p trusty-controller` invocations.
- Folds in verified corrections: MSRV toolchain-pin inconsistency (most CI jobs use `@stable`; `trusty-git-analytics/rust-toolchain.toml` pins `stable`), `tctl install` cannot install from source (prebuilt-tarball-first + crates.io fallback only, no `--path`) — added as an explicit warning in §7.1, `trusty-mpm`'s `publish = false` blocks `released`-mode BOM coverage, `trusty-git-analytics` publishes as `tga`, and the repo cannot be built from a read-only source tree (five `build.rs` files write into `$CARGO_MANIFEST_DIR`) — `worktree` mode must copy source into the guest rather than mount the host repo — plus the absence of any `mise.toml`/`.mise.toml` in the repo.
- Adds an `## Amendments` section at the top recording the 2026-07-30 changes. `Status: Accepted (owner-approved)` is left as-is.

## ADR decision
No superseding ADR was written. No prior ADR governs the harness's Rust-vs-bash location — only this design doc's own §7.3 "Resolved Decision 4" recorded it, and the harness is still fully unimplemented (see the "Remaining work" checklist), so there is nothing built against the old decision to reverse. The ADR convention (`docs/adr/README.md`) requires ADRs for architecturally-significant, costly-to-reverse decisions and treats *ADRs* as immutable once accepted (changed only by a superseding ADR); it does not extend that immutability requirement to design docs under `docs/*/research/02-design/`, which are edited directly as understanding evolves pre-implementation.

## Test plan
- [x] `grep -n 'trusty-controller' docs/trusty-installer/research/02-design/10-isolation-testing-harness.md` returns only the two deliberate historical mentions inside the new Amendments note.
- [x] Doc renders as valid Markdown; relative link to ADR-0013 (`../../../adr/0013-rename-trusty-controller-to-trusty-installer.md`) resolves on disk.
- N/A — documentation-only change, no code/tests affected.


mac-duetto:
This amendment updates the isolation-harness design following a measurement pass run in a local Tart macOS VM. The earlier revision's structural changes stand; its cost model did not.

**Seven edits:**
1. Cost model replaced with measured figures — `SKIP_UI_BUILD=1 cargo install --path crates/trusty-search --locked` = **112s** (409 crates, 8 vCPU/16 GB, cold registry, workspace `lto = "thin"` confirmed applied, `ort-sys` ONNX Runtime download confirmed working in-guest), against an estimated 20–55 min/crate. The prior figures were extrapolated from a 2-vCPU GitHub Actions runner and were wrong by roughly **20×**.
2. Prebaked golden image replaced with clone-`tahoe-base`-and-provision-per-run (`PROVISION_SEC` = **30s**, vs CoW clone at 0.31s). The golden's ~30s/run benefit did not justify three observed failure modes: `tart stop` write loss, a golden that shipped broken because of it, and a false-positive purity-gate bug.
3. Full-stack made the default scope (six primary crates with a shared `CARGO_TARGET_DIR` ≈ 4–8 min, extrapolated). Single-crate scope existed only because full-stack was believed to cost 45–90 min.
4. §6b (GitHub Actions CI cadence) marked **OUT OF SCOPE**, dated — the harness is local Tart VM only, run ad-hoc/manually. Section retained, not deleted, so the historical decision stays legible.
5. Operational constraints added: never bare-`tart stop` (it silently loses the guest's last write, reproduced 4/5; a tart exit code is not a completion signal — poll for observable stopped state); never `tart suspend` (resume broken, `VZErrorDomain Code=12`, with a documented unwedge procedure); `tart exec` as sole transport with SSH demoted to opt-in post-mortem; no dependence on guest shell rc files; assert `rustc --version` before each build step (toolchain drift confirmed real — `crates/trusty-git-analytics/rust-toolchain.toml` resolves to rustc 1.97.1 vs the workspace-pinned 1.91.1); launchd confirmed working under `tart exec`.
6. Cargo registry pre-warming recorded as not required — the 112s measurement was cold-registry.
7. Amendments section updated with a dated 2026-07-31 entry.

Full research record, including raw measurements and an adversarial review of the superseded recommendation, is in sibling PR #4456 at `docs/research/tart-vm-testing-harness/01-research/`.

mac-duetto:
The harness location changed by owner decision from `scripts/vmtest/` to project-root `vmtest-harness/`. All references in this document are updated, including the CI path-filter near §6b (updated even though §6b is marked OUT OF SCOPE, so it does not sit stale and contradictory). A dated 2026-07-31 amendment entry records the relocation.

The `members = ["crates/*"]` rationale for staying outside the Cargo workspace is unchanged — a project-root directory sits further from that glob than `scripts/` did, so the separation argument strengthens. Justifications that leaned on `scripts/` precedent have been removed rather than left stale.

Unchanged and still binding: the JSON-only assertion oracle (never scrape human-readable text) and `cargo install` only, never `cp` binaries into PATH (cdhash safety).

Related: PR #4457 (`02-design`) and PR #4456 (`01-research`) carry the same relocation. In #4456 the change was applied selectively — superseded documents deliberately retain the original path as a historical record.

PR #4456, opened 2026-07-31
docs: Tart VM install-testing research consolidation

mac-duetto:
## Summary

Consolidates the Tart VM install-testing research into `docs/research/tart-vm-testing-harness/01-research/`. This is a small, mechanical consolidation PR — it does not author new analysis content (a follow-up task adds that).

**Phases covered by this research:**
1. Two parallel, independently-authored research tracks (Track A / Fable and Track B / Opus) on whether a Tart-based clean-machine harness is viable for install-testing trusty-tools.
2. Fact-check reconciliation between the two tracks.
3. An empirical probe pass (sections A–K of `vm-install-probe-findings.md`) replacing unmeasured guesses with measurements.
4. An adversarial review that surfaced untested assumptions.
5. A decisive measurement pass (section K) that resolved those assumptions.

**Headline outcome:** the original cost model was falsified by roughly 20x. `cargo install --path crates/trusty-search --locked` measured at **112s** (409 crates, cold registry, thin LTO confirmed applied, `ort-sys` ONNX download confirmed working in-guest) against an original estimate of 20–55 minutes. Both the golden-image strategy and the single-crate default scope were killed by measurement: `PROVISION_SEC`=30s (bare `tahoe-base` provision) vs. 0.31s (pre-baked golden clone) does not justify the golden-image strategy's cost.

**Also surfaced by the probe pass:**
- `tart stop` silently loses the guest's last unsynced write (reproduced 4/5 trials) — this is the confirmed root cause of a golden image that shipped broken (`~/.zshenv` missing, `cargo` exit 127).
- `tart suspend` resume is broken (VZError 12); a documented unwedge procedure is recorded.
- Real toolchain drift exists via `crates/trusty-git-analytics/rust-toolchain.toml` (rustc 1.97.1 vs. workspace-pinned 1.91.1).
- `launchd` is confirmed working under `tart exec`.

**What's included:**
- `vm-install-probe-findings.md` — the A–K empirical probe writeup, relocated.
- `vm-install-testing-trackA-fable.md` / `vm-install-testing-trackB-opus.md` — the two parallel research tracks, rescued verbatim from untracked files in the main checkout.
- `artifacts/bake-golden.sh.superseded` — the probe instrument, now carrying a header explaining why it's retained (produced measurements A–K) and why it's superseded (killed by measurement; must not be used as the basis for the real harness).
- `logs/` — raw K-pass measurement logs.
- `README.md` — index of everything above; states research is COMPLETE and no harness has been built.

[...]

PR #4457, opened 2026-07-31
docs(research): Tart VM install testing harness — design specification

mac-duetto:
## Summary

Adds `docs/research/tart-vm-testing-harness/02-design/` — the design layer that refines the completed Tart VM research (`01-research/`, PR #4456) into an implementable specification for `scripts/vmtest/`.

**This is a design specification with no implementation.** `scripts/vmtest/` is not created here. No harness code, no VMs started, no builds run.

## What it specifies

An ad-hoc, manually-run harness that installs the trusty-tools stack inside a clean local Tart macOS VM and verifies (1) installation succeeds without errors and (2) nothing on the host is affected.

- **Settled owner decisions recorded as decisions, not options:** pattern (a) "released" is crates.io `cargo install --locked` **only** (`install.sh`/prebuilt tarballs out of scope — the crates.io path is measured, `install.sh` in a guest never was); `trusty-mpm` is a documented gap in pattern (a) (`publish = false`); the stack is **seven** crates; implementation order is **(c) local-source → (b) branch → (a) released**; local Tart VM only, with full dependency download accepted.
- **Placement:** `scripts/vmtest/` in bash, deliberately **outside** the Cargo workspace — `members = ["crates/*"]` would auto-sweep anything under `crates/` into `cargo test --workspace`, clippy, the SLOC cap, the test-pointer and SLD lints, and the publish guard, which defeats "ad-hoc, separate from existing test infrastructure".
- **Run lifecycle** with exact commands, including the boot-before-provision negative probe that keeps the "cargo absent → guide-and-abort" case live on every run.
- **No golden image**, with the rationale recorded: baking saves only ~30s against a 0.31s CoW clone, and exhibited three observed failure modes including a golden that shipped broken.
- **Transport rules:** `tart exec` is the sole transport (exit codes propagate exactly, 200k lines untruncated); no `--env` flag so every command self-prefixes `PATH`; never depend on guest shell rc files.
- **Verification:** JSON-only oracle, `expected-binaries.tsv` with a `--check-table` self-diff (needed because `docs/reference/release-workflow.md` is stale re: `trusty-console`), cargo-only installs for cdhash safety, and a Single-Install Convention gate.
- **Operational constraints, each with its forcing measurement inline:** never bare-`tart stop` (write loss 4/5 trials, confirmed root cause of a broken golden), never `tart suspend` (`VZErrorDomain Code=12`, unrecoverable by retry), per-build-step `rustc --version` assertion (confirmed 1.91.1 vs 1.97.1 drift via `trusty-git-analytics/rust-toolchain.toml`), 8 vCPU/16 GB sizing, shared `CARGO_TARGET_DIR`.
- **Isolation guarantee** enumerating host touchpoints, with the never-mount rule identified as what closes the one real hole.
- **Known gaps recorded honestly**, including that all TCC observations are conditional on being run from iTerm2 by one user, and the `kTCCServiceAudioCapture` preflight that fires on VM start even with `--no-graphics`.

## Files

- `docs/research/tart-vm-testing-harness/02-design/README.md`
- `docs/research/tart-vm-testing-harness/02-design/01-vm-install-harness.md`

Naming and numbering follow `docs/trusty-installer/research/02-design/`.

## Cross-references

Links back to `../01-research/conclusions-post-measurement.md` and `../01-research/vm-install-probe-findings.md`. Those land in **#4456** (branch `docs/tart-vm-research-final`), so the relative links resolve once both branches are on `main`. `docs/trusty-installer/research/02-design/10-isolation-testing-harness.md` is the parent Accepted design, amended by **#4438** to match these measurements.

## Notes for review

The document flags, inline, four places where the source material was internally inconsistent or underspecified rather than silently resolving them: the `git archive` vs `git ls-files -co` payload-measurement mismatch, the mise-vs-rustup toolchain-resolution question behind the drift assertion, the "build first" annotation on `install-local.sh`, and the `--check-table` diff source of truth. The 4–8 min full-stack figure is labelled an extrapolation for six crates (seven are now in scope) and flagged low-confidence against the per-crate measurements.

**Do not merge yet** — design review first.

### DIFF
--- a/docs/research/tart-vm-testing-harness/02-design/01-vm-install-harness.md
+++ b/docs/research/tart-vm-testing-harness/02-design/01-vm-install-harness.md
@@ -160,7 +160,7 @@ test — which is precisely the contamination the harness exists to avoid.
 ### 3. Component architecture
 
 ```
-scripts/vmtest/                      # bash, OUTSIDE the Cargo workspace
+vmtest-harness/                      # project root, bash, OUTSIDE the Cargo workspace
 ├── vmtest                           # driver CLI
 ├── lib/
 │   ├── vm.sh                        # tart lifecycle — the OS boundary for future Linux support
--- a/docs/research/tart-vm-testing-harness/02-design/01-vm-install-harness.md
+++ b/docs/research/tart-vm-testing-harness/02-design/01-vm-install-harness.md
@@ -593,14 +593,18 @@ A single provisioning-time version check therefore proves nothing about the
 compiler that will actually build a given crate. The assertion must be adjacent to
 the build, in the same working directory.
 
-> **Underspecified in the source research:** provisioning installs `rust@1.91` via
-> **mise** (§3.3), whereas the drift mechanism described above is a **rustup**
-> directory-resolution behaviour. Whether a mise-managed Rust exhibits the same
-> `rust-toolchain.toml` resolution — and therefore whether the guest will reproduce
-> the 1.91.1/1.97.1 split at all — was not measured. The per-build-step assertion
-> is correct either way (it *observes* rather than assumes), but the implementation
-> should record which resolver is actually in play in the guest, and the first
-> pattern-(c) run should be treated as the measurement that settles it.
+> **Confirmed under a mise-provisioned guest:** provisioning installs `rust@1.91`
+> via **mise** (§3.3), and the drift above was measured in-guest on exactly that
+> setup. Measurement K5 ran on VM `probe-k2` — the VM provisioned in measurement
+> K2 via `mise use -g rust@1.91` — where `rustc --version` and `rustup show` in
+> both `crates/trusty-git-analytics/` and the workspace root reproduced the same
+> split: rustc **1.97.1** in the crate directory versus **1.91.1** at the root.
+> This is expected: mise's rust backend **delegates to rustup** under the hood
+> (mise downloads `rustup-init`, and the mise rust entry is a symlink to it — see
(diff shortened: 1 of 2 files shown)


mac-duetto:
Two corrections pushed to this branch:

1. **Factual — §8.4 mise/rustup caveat was wrong.** The caveat claimed it was unmeasured whether a mise-managed Rust toolchain honours `rust-toolchain.toml` directory resolution the same way rustup does. It was measured: K5 ran on VM `probe-k2` (provisioned in K2 via `mise use -g rust@1.91`), and the 1.91.1/1.97.1 split was confirmed in-guest via `rustc --version` / `rustup show` in both `crates/trusty-git-analytics/` and the workspace root. This is expected because mise's rust backend delegates to rustup under the hood (research §E). Caveat replaced with a positive, cited statement.

2. **Clarification — §10 "build first" ambiguity resolved.** Confirmed this means "implement `install-local.sh` first," consistent with decision D4's (c) → (b) → (a) implementation order — not a separate `cargo build` step before `cargo install` in the guest. Removed the request for owner confirmation.

All other flagged items (§6.1 payload-measurement mismatch, §7.2 `--check-table` source of truth, §9 extrapolation, §4.2 negative-probe sequencing) are unchanged.

mac-duetto:
Relocated the harness from scripts/vmtest/ to a root-level vmtest-harness/ directory per owner decision. The members = ["crates/*"] separation rationale in section 2.1 is unchanged; a project-root location only strengthens it. Section 2 placement rationale rewritten accordingly (dedicated top-level directory vs. loose script, no more scripts/ precedent).

Issue #1976, opened 2026-07-03
trusty-mpm: agent_delegate + session_status reject managed (tmpm-) sessions — gating registry is legacy-only


bobmatnyc:
## Problem

Discovered during a live dogfood full-service test of trusty-mpm.

The MCP tools `agent_delegate` and `session_status` return `"no such session: <id>"` for **managed** (`tmpm-`, `kind=managed`) sessions, while working for **legacy native PTY** sessions (`kind=legacy`). `session_list` and `session_activity` resolve both.

Root cause (`crates/trusty-mpm/src/daemon/mcp_backend.rs`): both handlers gate on `self.state.session(id)`, which only consults the legacy in-process `DaemonState` session registry (`daemon/state/sessions.rs`). Managed sessions live in the separate, lazily-initialized `DaemonState::session_manager()` store. `session_list` merges both, which is why the list shows managed sessions but `session_status`/`agent_delegate` can't find them.

## Impact

The delegation gating + tracking layer (circuit breaker, depth limits, dashboard delegation tree) is unavailable to the **primary session type trusty-mpm itself spawns**. The *execution* path (native Agent/Task spawn) works fine in managed sessions — only the MCP tracking/gating companion is broken.

## Expected

`agent_delegate` and `session_status` should resolve managed sessions too: when the legacy lookup misses, fall back to `state.session_manager().await.get(id)` (both handlers are already async). For `session_status`, return the managed record under `session` (delegations/memory may be empty for managed).

## Scope

- `mcp_backend.rs`: managed-store fallback in `session_status` and `agent_delegate`.
- Tests covering a managed-session id round-trip.

Found via dogfood test 2026-07-03. Related: #1975.

PR #4096, opened 2026-07-27
feat(trusty-mpm): observe native subagent dispatches and give delegations a real lifecycle (#2864 S1+S2)

bobmatnyc:
## Summary

Slices **S1 + S2** of #2864 — the foundation for `/tm-session-pause` recording in-flight agents. Later slices (S3 persistence, S4 pause snapshot, S5 resume, S6 quiesce) build on this and are **not** in scope here.

Delegation tracking had two defects that together made it useless as a basis for *"which agents are in flight?"*:

1. **It was opt-in.** A record existed only if the PM chose to call the `agent_delegate` MCP tool, so a PM that dispatched work with the native subagent tool alone left no trace at all.
2. **Status never moved.** The only production construction yields `Queued` (`core/agent.rs`) and nothing ever wrote `Running`/`Completed`/`Failed`. Since `has_live_children` counts `Queued`/`Running` as live, **every delegation was live forever**, permanently suppressing the idle nudge for any session that had ever delegated once.

Both are fixed by observing hooks the daemon **already receives**: `PreToolUse` with `matcher: "*"` is installed on every managed session, so every native subagent dispatch already arrives at `HookService::process`. No new hook, no opt-in — the correlation keys were simply being dropped on the floor.

## Step 0: the correlation model was measured, not assumed

Captured from **live Claude Code 2.1.220** hook stdin, with two *concurrent* subagents — the case that breaks any "most recent" guess:

```
PreToolUse   tool_name=Agent  tool_use_id=toolu_01DA…   (no agent id yet)
PostToolUse  tool_name=Agent  tool_use_id=toolu_01DA…
             tool_response={ isAsync: true, status: "async_launched",
                             agentId: "a403cdbc078b5c474", duration_ms: 1 }
SubagentStop                  agent_id="a403cdbc078b5c474"
```

Three findings, two of which **inverted** the planned design:

| Question | Answer |
|---|---|
| Does `SubagentStop` carry a correlation key? | **Yes — `agent_id`**, byte-identical to `PostToolUse.tool_response.agentId`. Authoritative, not advisory. |
| Does `SubagentStop.transcript_path` point at the subagent? | **No — the parent's.** The subagent's is a separate `agent_transcript_path` field. |
| Does `PreToolUse`/`PostToolUse` carry `tool_use_id`? | **Yes**, and it is identical across the pair. |

- **The tool is named `Agent`, not `Task`,** in current Claude Code. Both names are matched — recognising only one silently disables tracking on half the installed base.
[...]

Issue #2864, opened 2026-07-16
feat(mpm): manage and reattach to durable Trusty Code delegations


bobmatnyc:
## Outcome

Trusty MPM can manage delegated Trusty Code work without owning the delegated agent process or losing it when the MPM session exits.

## Scope

- Store the durable Trusty Code task/workstream identity returned by the structured adapter.
- Attach, detach, reattach, cancel, and explicitly take over through Trusty Code APIs.
- Recover accepted in-flight work after either daemon restarts.
- Enforce one active attachment lease and surface conflicts.
- Route progress and completion through the shared channel contract and preserve authorization/provenance.

## Done when

An MPM session can delegate work, exit, and later reattach; a different authorized PM can explicitly take over; duplicate ownership is rejected; and terminal/pane identity is not used as the durable key.

Trusty Code runtime implementation is #5430. Structured MPM integration is #5431.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
## Reopened — bookkeeping correction, not new scope

This issue was auto-closed as COMPLETED on 2026-07-16 by PR #2865. That PR landed **only the
spec** — `docs/specs/durable-background-agents.md` (DOC-40) — and its own scope note says
verbatim: *"The PR carrying this doc opens **no** Rust changes."*

Verified current state:

- `docs/specs/durable-background-agents.md` line 3 still reads `**Status:** Draft`.
- Grep across the whole workspace returns **zero** hits:

  ```
  $ grep -rE 'durable_agent|agent_attach|agent_detach|attach_lease|AgentLease' crates/ | wc -l
  0
  ```

The feature described by this issue — durable background agents with exclusive attach/detach — is
**entirely unimplemented**. The `Closes #2864` in PR #2865 should have been a `Refs`.

Reopening restores accurate backlog state. **No scope is being added or changed** by this
comment; the issue's original acceptance criteria stand as written.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
## Verified current state of the gap (the blocker for "`/tm-session-pause` should stop all in-flight agents in a resumable state")

### 1. The pause/resume skills know nothing about agents

`crates/trusty-mpm/src/assets/skills/tm-session-pause.md` today does **only**:

1. todo reconcile
2. one `tmux display-message`
[...]

PR #4463, opened 2026-07-31
fix(trusty-mpm): bridge managed id to Claude session UUID in session_status

### LINKED ISSUE #4141: bug(trusty-mpm): session_status reports delegation_count 0 while subagents are in flight — ManagedSessionId vs Claude session UUID key mismatch
## Summary

`session_status` looks up hook-observed delegations with the **wrong identifier**, so it reports `"delegation_count": 0` even while subagents are actively running. The surface that is supposed to answer "what is in flight right now" is structurally blind to the very feature it reports on.

Found during a full-context code-critic gap review of PR #4096, which identified findings living *outside* that PR's changed files. Source review comment:
https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094071838

Deliberately held out of #4096's current fix round to keep that PR from growing without bound — filed here so it survives.

## The key mismatch

- `session_status` (`crates/trusty-mpm/src/daemon/mcp_backend.rs:121`) resolves a **`ManagedSessionId`** and then passes that id straight into `delegations_for(id)`.
- Hook-observed delegations are keyed by the **Claude session UUID** — see `crates/trusty-mpm/src/daemon/api.rs:1285`, where the hook payload flows into `observe`.

These are two different identifier spaces. The evidence that they are *not* interchangeable is already in the codebase:

- `SessionRecord` carries the managed id and the Claude session UUID as **separate fields**.
- `correlate_session_start` exists for the sole purpose of **bridging** the two. If they were the same key, that correlation step would not need to exist.

So the lookup in `session_status` silently misses every hook-observed delegation and falls through to an empty result.

## Consequence

`session_status(<managed-id>)` returns:

```json
"delegation_count": 0
```

...while subagents are genuinely in flight. The delegation-tracking feature's own retention documentation names `session_status` as *the* consumer that answers "what is in flight right now" — so the surface meant to report the feature cannot see it.

## Why this matters beyond a wrong number

The goal of #2864 is for `/tm-session-pause` to enumerate in-flight agents before pausing. A reporting surface that returns `0` makes a pause **look safe when it is not** — the operator gets a false all-clear and pauses on top of running work.

That is the same false-all-clear failure family as:

- #4091 — worktree prune force-deletes uncommitted work because nothing checks for a dirty tree; the absence of a signal reads as "nothing to lose".
- #4086 — a degraded trusty-review verdict was indistinguishable from a complete one.

In all three, a missing or failed lookup degrades into a confident-looking negative rather than an explicit unknown.

## Fix direction (options — not settled)

1. **Bridge via `SessionRecord`** — resolve the managed id to its stored `claude_session_id` and query delegations with that, exactly as `correlate_session_start` already does. Smallest change, reuses an existing proven path.
2. **Key delegations consistently** — store hook-observed delegations under the managed id at ingest time. Requires the correlation to have already happened, so ordering needs care.
3. **Make `delegations_for` accept either identifier** — push disambiguation down into the lookup so no caller can get it wrong again.

### Non-negotiable constraint on any fix

Whichever option is taken, a **failed bridge lookup must not silently return `0`**. "The correlation is missing / the session UUID is unknown" must be distinguishable from "genuinely none in flight", and surfaced as an explicit unknown to the caller. Collapsing those two cases into `0` reproduces the exact false-all-clear this issue is about.

## Verification

A fix should be provable by a test that registers a session, observes a hook delegation under the Claude session UUID, then calls `session_status` with the **managed** id and asserts a non-zero `delegation_count` — plus a negative test asserting that an unbridgeable lookup does not report `0`.

## Cross-references

- Epic: #2864
- PR that surfaced it: #4096 (remains scoped to hook-observed tracking only)
- Same failure family: #4091, #4086

bobmatnyc:
## Summary

`session_status` resolved a `ManagedSessionId` and passed it straight into `delegations_for(id)` (crates/trusty-mpm/src/daemon/mcp_backend.rs:121). That missed a key space: hook-observed delegations (`delegation_tracker::observe`) are keyed by the Claude session UUID, bridged from a managed record only via `SessionRecord::claude_session_id` / `session_start_correlation::correlate_session_start`. So `delegation_count` silently reported `0` for a managed session with subagents genuinely in flight — a false all-clear for anything (e.g. `/tm-session-pause`) that trusts this surface.

## What the code does now

Delegations for a managed session live in up to TWO key spaces:

- **Managed-id space** — `agent_delegate` (#1976) keys a delegation by whatever session id the caller passed, which for a managed session is the managed id itself (exercised over `/rpc` by `connectors::tm::WorkstreamConnector::delegate`, certified by `agent_delegate_accepts_managed_session`).
- **Claude-UUID space** — hook-observed delegations, reachable only by bridging through `record.claude_session_id`.

The managed branch of `session_status` reads BOTH spaces unconditionally-for-managed-id / bridge-gated-for-Claude-UUID, and unions them, de-duplicating by delegation id. An earlier version of this fix replaced the managed-id read with the bridged one instead of unioning — that dropped every `agent_delegate`-keyed record and was caught and fixed in review.

De-duplication is by delegation id only: it removes an exact record reachable under both keys, but does **not** merge a managed-id `agent_delegate` record with a separate Claude-UUID hook-observed record for the same logical subagent — those remain two distinct `Delegation`s (`delegation_tracker::dedup` only matches within one session/agent pairing and cannot span the bridge). `delegation_count` can therefore overcount one in-flight subagent as two; that is the tracker's existing, deliberate trade-off (overcount over false-merge) and unrelated to this fix.

When the bridge itself is missing (no correlated Claude session id yet), `delegation_count` **and** `delegations` are both reported as an explicit `null` with `delegation_lookup: "unbridged"` — the managed-id half alone would be a floor masquerading as a total, so neither field may be read as a confident (even if empty) answer.

The legacy branch's `id` equals the Claude session UUID only on the `SessionStart`-hook-auto-registered path; the tmux-pane-discovery, native-process-discovery, and `POST /sessions` paths instead mint a fresh `SessionId` never correlated to anything. Since no marker distinguishes those three paths at this call site, the legacy branch no longer asserts `delegation_lookup: "ok"` (a soundness claim it could not establish for all three) — it omits the field entirely.

## Files changed

- `crates/trusty-mpm/src/daemon/mcp_backend.rs` — the union fix + `// #4141` markers
- `crates/trusty-mpm/src/daemon/mcp_backend_tests.rs` — regression test for the original bug (hook-observed delegation invisible via the managed id), the inverse regression test (managed-id-keyed delegation must stay visible after the bridge fix), and a negative test asserting an unbridged lookup reports `null`/`null`, not `0`/`[]`
- `crates/trusty-mpm/CHANGELOG.md` — `[Unreleased]` entry

## Verification

- `cargo test -p trusty-mpm --lib mcp_backend` — 17/17 passed
- `cargo test -p trusty-mpm` (full crate) — full lib suite green, no failures
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- `bash scripts/check_line_cap.sh` — `7 allowlisted, 0 violations — OK`

Closes #4141

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -20,6 +20,21 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `usable_agent_id` (non-null, non-empty string), so the presence-check and
     the consumers are structurally incapable of disagreeing again.
 
+- `session_status` MCP tool no longer reports `delegation_count: 0` for a managed session with subagents genuinely in flight (closes [#4141](https://github.com/bobmatnyc/trusty-tools/issues/4141))
+  - The handler resolved a `ManagedSessionId` and passed it straight into
+    `delegations_for`, but hook-observed delegations are keyed by the Claude
+    session UUID — a different identifier space bridged only via
+    `SessionRecord::claude_session_id` /
+    `session_start_correlation::correlate_session_start`. The mismatch made
+    the surface that answers "what is in flight right now" structurally
+    blind to its own feature.
+  - The managed branch now bridges through `record.claude_session_id` before
+    querying. When the bridge itself is missing (no correlated Claude session
+    id yet), `delegation_count` reports an explicit `null` with
+    `delegation_lookup: "unbridged"` rather than a confident-looking `0` —
+    an unbridgeable lookup must never be indistinguishable from "genuinely
+    none in flight".
+
 - managed sessions can reach the bundled agent roster again — every delegation was degrading to `general-purpose` (closes [#4451](https://github.com/bobmatnyc/trusty-tools/issues/4451))
   - Daemon-managed spawns now launch with `--setting-sources user,project,local`
     instead of `project,local`. Claude Code discovers subagents per settings
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -28,12 +28,14 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     `session_start_correlation::correlate_session_start`. The mismatch made
     the surface that answers "what is in flight right now" structurally
     blind to its own feature.
-  - The managed branch now bridges through `record.claude_session_id` before
-    querying. When the bridge itself is missing (no correlated Claude session
-    id yet), `delegation_count` reports an explicit `null` with
-    `delegation_lookup: "unbridged"` rather than a confident-looking `0` —
-    an unbridgeable lookup must never be indistinguishable from "genuinely
-    none in flight".
+  - The managed branch now reads BOTH key spaces — the managed id
+    `agent_delegate` (#1976) writes under, and the bridged Claude UUID
+    hook-observed delegations use — and unions them, deduplicated. When the
+    bridge itself is missing (no correlated Claude session id yet),
+    `delegation_count` and `delegations` both report an explicit `null` with
+    `delegation_lookup: "unbridged"` rather than a confident-looking
+    `0`/`[]` — an unbridgeable lookup must never be indistinguishable from
+    "genuinely none in flight".
 
 - managed sessions can reach the bundled agent roster again — every delegation was degrading to `general-purpose` (closes [#4451](https://github.com/bobmatnyc/trusty-tools/issues/4451))
   - Daemon-managed spawns now launch with `--setting-sources user,project,local`
--- a/crates/trusty-mpm/src/daemon/mcp_backend.rs
+++ b/crates/trusty-mpm/src/daemon/mcp_backend.rs
@@ -100,24 +100,38 @@ impl OrchestratorBackend for StateBackend {
     /// serialized via the shared `record_to_json` (matching `session_list`) and
     /// carry no memory snapshot.
     ///
-    /// #4141: the legacy registry's `SessionId` IS the Claude session UUID
-    /// (hook auto-registration uses it directly), so `delegations_for(id)`
-    /// resolves correctly for that branch with no bridging. The MANAGED
-    /// branch is different: `id` here is a `ManagedSessionId`, but
-    /// hook-observed delegations are keyed by the Claude session UUID that
-    /// `session_start_correlation::correlate_session_start` stores on the
-    /// record as `claude_session_id` — a distinct identifier space (that's
-    /// the whole reason `SessionRecord` carries both fields separately).
-    /// Passing the managed id straight into `delegations_for` silently
-    /// matched nothing and reported `delegation_count: 0` even while
-    /// subagents were actively running — a false all-clear indistinguishable
-    /// from "genuinely none in flight". The fix bridges through
-    /// `record.claude_session_id` before querying, exactly as
-    /// `correlate_session_start` intends; when the bridge itself is missing
-    /// (no correlated Claude session id yet, e.g. the session hasn't started
-    /// or no `SessionStart` hook has landed), `delegation_count` is reported
-    /// as an explicit `null` with `delegation_lookup: "unbridged"` — never a
-    /// confident-looking zero.
+    /// # Delegation lookup (#4141)
+    ///
+    /// Delegations live in up to TWO key spaces. `agent_delegate` (#1976) keys
+    /// a delegation by whatever session id the CALLER passed — for a managed
+    /// session that is the MANAGED id itself (exercised over `/rpc` by
+    /// `connectors::tm::WorkstreamConnector::delegate`, certified by
+    /// `agent_delegate_accepts_managed_session` in `mcp_backend_tests.rs`).
+    /// Hook-observed delegations (`delegation_tracker::observe`) are instead
+    /// keyed by the Claude session UUID, reachable from a managed record only
+    /// by bridging through `record.claude_session_id` (populated by
+    /// `session_start_correlation::correlate_session_start`). Replacing the
+    /// managed-id read with the bridged one — instead of unioning both —
+    /// silently drops every `agent_delegate`-keyed record: the exact
+    /// false-all-clear #4141 exists to eliminate, just relocated to the other
+    /// key. The managed branch below reads BOTH spaces and de-duplicates by
+    /// delegation id.
+    ///
+    /// The legacy branch's `id` equals the Claude session UUID only on the
+    /// `SessionStart`-hook-auto-registered path; a tmux-pane-discovered,
+    /// native-process-discovered, or `POST /sessions`-registered legacy
+    /// session instead gets a freshly minted `SessionId` never correlated to
+    /// anything, so `delegations_for(id)` can under-report there too. No
+    /// marker distinguishes those three registration paths at this call site,
+    /// so the legacy branch omits `delegation_lookup` rather than assert an
+    /// `"ok"` it cannot establish for every path (that pre-existing gap is
+    /// tracked separately from this fix).
+    ///
+    /// When the managed bridge itself is missing (no correlated Claude
+    /// session id yet), `delegation_count` AND `delegations` are both an
+    /// explicit `null` with `delegation_lookup: "unbridged"` — the managed-id
+    /// half alone would be a floor masquerading as a total, so neither field
+    /// may be read as a confident (even if empty) answer.
     async fn session_status(&self, session_id: &str) -> Result<Value, String> {
         let id = parse_session_id(session_id)?;
         // Legacy in-process registry first.
@@ -130,32 +144,43 @@ impl OrchestratorBackend for StateBackend {
                 "memory": memory,
                 "delegation_count": delegations.len(),
                 "delegations": delegations,
-                "delegation_lookup": "ok",
             }));
         }
         // Managed store fallback (#1976): the `tmpm-` sessions `session_list`
         // surfaces and `session_stop`/`session_resume` already target by id.
         let manager = self.state.session_manager().await;
         if let Ok(record) = manager.get(&ManagedSessionId(id.0)).await {
-            // #4141: bridge the managed id to the Claude session UUID before
-            // querying delegations — they live in a different key space.
+            // #4141: read the managed-id key space unconditionally — this is
+            // what `agent_delegate` (#1976) writes under for a managed
+            // session — then union in the Claude-UUID key space when the
+            // bridge resolves, de-duplicating by delegation id.
+            let mut delegations = self.state.delegations_for(id);
             let claude_uuid = record
                 .claude_session_id
                 .as_deref()
                 .and_then(|s| Uuid::parse_str(s).ok());
-            let (delegation_count, delegations, lookup) = match claude_uuid {
+            let (delegation_count, delegations_value, lookup) = match claude_uuid {
                 Some(uuid) => {
-                    let delegations = self.state.delegations_for(SessionId(uuid));
+                    for d in self.state.delegations_for(SessionId(uuid)) {
+                        if !delegations.iter().any(|existing| existing.id == d.id) {
+                            delegations.push(d);
+                        }
+                    }
                     (json!(delegations.len()), json!(delegations), "ok")
                 }
-                None => (Value::Null, Value::Array(Vec::new()), "unbridged"),
+                // The hook-observed half is unreachable without the bridge,
+                // so the managed-id half alone cannot stand in for the
+                // total — report the explicit unknown for BOTH fields
+                // rather than let a caller reading only `delegations` see a
+                // confident empty list.
+                None => (Value::Null, Value::Null, "unbridged"),
             };
             return Ok(json!({
                 "session": crate::daemon::managed_routes::record_to_json(&record),
                 "kind": SessionRecordKind::Managed.as_str(),
                 "memory": Value::Null,
                 "delegation_count": delegation_count,
-                "delegations": delegations,
+                "delegations": delegations_value,
                 "delegation_lookup": lookup,
             }));
         }
--- a/crates/trusty-mpm/src/daemon/mcp_backend.rs
+++ b/crates/trusty-mpm/src/daemon/mcp_backend.rs
@@ -115,7 +115,17 @@ impl OrchestratorBackend for StateBackend {
     /// silently drops every `agent_delegate`-keyed record: the exact
     /// false-all-clear #4141 exists to eliminate, just relocated to the other
     /// key. The managed branch below reads BOTH spaces and de-duplicates by
-    /// delegation id.
+    /// delegation id — but that only removes an exact record appearing
+    /// twice, e.g. the same `Delegation` reachable under both the managed id
+    /// and the Claude UUID. It does NOT merge a PM's `agent_delegate` call
+    /// (managed-id record) with the hook-observed dispatch of the SAME
+    /// logical subagent (Claude-UUID record): those are two distinct
+    /// `Delegation`s with different ids and `delegation_tracker::dedup` only
+    /// matches within one session/agent pairing, so it cannot span the
+    /// bridge. `delegation_count` can therefore overcount one in-flight
+    /// subagent as two — this is the existing, deliberate tracker trade-off
+    /// (favouring overcount over a false merge, per `dedup`'s own doc
+    /// comment), not new behaviour this fix introduces or is meant to close.
     ///
     /// The legacy branch's `id` equals the Claude session UUID only on the
     /// `SessionStart`-hook-auto-registered path; a tmux-pane-discovered,
@@ -153,7 +163,11 @@ impl OrchestratorBackend for StateBackend {
             // #4141: read the managed-id key space unconditionally — this is
             // what `agent_delegate` (#1976) writes under for a managed
             // session — then union in the Claude-UUID key space when the
-            // bridge resolves, de-duplicating by delegation id.
+            // bridge resolves, de-duplicating by delegation id (removes an
+            // exact record reachable under both keys; does NOT merge a
+            // managed-id `agent_delegate` record with a separate
+            // Claude-UUID hook-observed record for the same logical
+            // subagent — see the doc comment above).
             let mut delegations = self.state.delegations_for(id);
             let claude_uuid = record
                 .claude_session_id
--- a/crates/trusty-mpm/src/daemon/mcp_backend.rs
+++ b/crates/trusty-mpm/src/daemon/mcp_backend.rs
@@ -98,8 +98,26 @@ impl OrchestratorBackend for StateBackend {
     /// reported "no such session" for the `tmpm-` sessions trusty-mpm itself
     /// spawns, even though `session_list` surfaces them. Managed records are
     /// serialized via the shared `record_to_json` (matching `session_list`) and
-    /// carry no memory snapshot; delegations are keyed by UUID so they resolve
-    /// for either family.
+    /// carry no memory snapshot.
+    ///
+    /// #4141: the legacy registry's `SessionId` IS the Claude session UUID
+    /// (hook auto-registration uses it directly), so `delegations_for(id)`
+    /// resolves correctly for that branch with no bridging. The MANAGED
+    /// branch is different: `id` here is a `ManagedSessionId`, but
+    /// hook-observed delegations are keyed by the Claude session UUID that
+    /// `session_start_correlation::correlate_session_start` stores on the
+    /// record as `claude_session_id` — a distinct identifier space (that's
+    /// the whole reason `SessionRecord` carries both fields separately).
+    /// Passing the managed id straight into `delegations_for` silently
+    /// matched nothing and reported `delegation_count: 0` even while
+    /// subagents were actively running — a false all-clear indistinguishable
+    /// from "genuinely none in flight". The fix bridges through
+    /// `record.claude_session_id` before querying, exactly as
+    /// `correlate_session_start` intends; when the bridge itself is missing
+    /// (no correlated Claude session id yet, e.g. the session hasn't started
+    /// or no `SessionStart` hook has landed), `delegation_count` is reported
+    /// as an explicit `null` with `delegation_lookup: "unbridged"` — never a
+    /// confident-looking zero.
     async fn session_status(&self, session_id: &str) -> Result<Value, String> {
         let id = parse_session_id(session_id)?;
         // Legacy in-process registry first.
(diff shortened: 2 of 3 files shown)


bobmatnyc:
# Adversarial code review — #4463

Reviewed at `d4931616`. CI green at that SHA (23/23). Reviewed against the spec in
[#4141](https://github.com/bobmatnyc/trusty-tools/issues/4141) and the review comment that
spawned it ([#4096 comment](https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094071838)),
not against the PR narrative.

## Verdict: BLOCK

The bridge itself is correct and the regression test is real — I reverted
`mcp_backend.rs` to `origin/main` with the new tests in place and both fail with exactly the
reported symptom (`delegation_count: 0`), so this is not a test that passes against the unfixed
code. But the fix **replaces** the managed-id lookup instead of **unioning** it, and that drops a
live key space. The result is the same false all-clear #4141 exists to eliminate, now wearing an
affirmative `delegation_lookup: "ok"` label.

## Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/daemon/mcp_backend.rs` | 139-152 | Bridge **replaces** the managed-id lookup; every `agent_delegate`-keyed delegation becomes invisible and is reported as `delegation_count: 0` with `delegation_lookup: "ok"` | Union both key spaces, don't swap one for the other |
| HIGH | `crates/trusty-mpm/src/daemon/mcp_backend.rs` | 104-134 | Legacy branch hard-codes `delegation_lookup: "ok"`; the doc comment's claim that the legacy `SessionId` *is* the Claude UUID is false for 2 of 3 legacy registration paths | Correct the claim; don't assert `"ok"` where soundness isn't established |
| MEDIUM | `crates/trusty-mpm/src/daemon/mcp_backend.rs` | 151 | `delegations: []` contradicts `delegation_count: null` — a caller reading the array still gets a confident empty | Emit `Value::Null` for `delegations` too when unbridged |
| LOW | `crates/trusty-mpm/src/daemon/mcp_backend.rs` | 116-117 (PR body) | "the same pattern `idle_nudge.rs` already uses" — `idle_nudge` bridges the opposite direction and never reads `claude_session_id` off a record | Narrow the claim to the `parse_str` → `SessionId` mechanic |

---

### CRITICAL 1 — the bridge drops the `agent_delegate` key space, and labels the loss `"ok"`

`crates/trusty-mpm/src/daemon/mcp_backend.rs:139-152`

```rust
let claude_uuid = record
    .claude_session_id
    .as_deref()
    .and_then(|s| Uuid::parse_str(s).ok());
let (delegation_count, delegations, lookup) = match claude_uuid {
    Some(uuid) => {
        let delegations = self.state.delegations_for(SessionId(uuid));   // <-- ONLY this key
        (json!(delegations.len()), json!(delegations), "ok")
    }
    None => (Value::Null, Value::Array(Vec::new()), "unbridged"),
};
```

Two writers populate `DaemonState.delegations`, under **two different keys**:

1. `delegation_tracker::observe` (`services/delegation_tracker.rs:229`) — keyed by the **Claude
   session UUID** the hook arrives under. This is the half the PR fixes.
2. `agent_delegate` (`mcp_backend.rs:175`) — `Delegation::new(id, …)` where `id` is *whatever
   session id the caller passed*. For a managed session that is the **managed id**, and that is
   not incidental: #1976 changed `agent_delegate` specifically to accept the managed id, and
   `agent_delegate_accepts_managed_session` (`mcp_backend_tests.rs:183`) certifies it with
   `assert_eq!(state.delegations_for(SessionId(record.id.0)).len(), 1)`.

`origin/main` read key (2) and missed key (1). This PR reads key (1) and misses key (2). The
blindness moved; it did not go away.

**Verified, not inferred.** I added a probe on top of `d4931616` — create a managed record,
correlate a `claude_session_id` (bridge succeeds), then call `agent_delegate` with the managed id
exactly as `session_list` hands it to a caller:

```
PROBE: delegation_count=0 delegation_lookup="ok" delegations=[]
```

with `state.delegations_for(SessionId(record.id.0)).len() == 1` asserted immediately beforehand.
So a delegation that demonstrably exists is reported as zero, under a label asserting the lookup
was sound. That is strictly worse than an unlabeled `0`: `"ok"` tells the caller not to doubt it.

This is the issue's **non-negotiable constraint** broken from the other side. #4141: *"a failed
bridge lookup must not silently return `0` … Collapsing those two cases into `0` reproduces the
exact false-all-clear this issue is about."* The bridge here did not fail — it succeeded and
answered a question about only one of the two key spaces that receive writes. It is also a
regression: that record was visible on `origin/main`.

The managed-id key space is not theoretical. `connectors/tm.rs:324` (`WorkstreamConnector::delegate`)
POSTs `agent_delegate` over `/rpc` with the session id from the managed create path, exercised by
`tm_tests::create_session_full_lifecycle`. And `delegation_tracker::dedup`'s doc comment states
outright that a PM which calls `agent_delegate` *and* dispatches is the case it exists to handle.
The [#4096 review comment](https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094071838)
prescribed the union with this exact rationale — *"union with the managed id for back-compat with
existing `agent_delegate` records"* — and the PR drops that half with no stated reason.

Fix — union, and let the label describe what was actually reachable:

```rust
// Managed-id records (`agent_delegate`, #1976) and Claude-UUID records
// (hook-observed, #4141) are two live key spaces. Read BOTH.
let mut delegations = self.state.delegations_for(id);
let claude_uuid = record
    .claude_session_id
    .as_deref()
    .and_then(|s| Uuid::parse_str(s).ok());
let (delegation_count, lookup) = match claude_uuid {
    Some(uuid) => {
        delegations.extend(self.state.delegations_for(SessionId(uuid)));
        (json!(delegations.len()), "ok")
    }
    // The managed-id half is known; the hook-observed half is NOT. A count
    // would be a floor masquerading as a total, so report the explicit
    // unknown and still hand back what was reachable.
    None => (Value::Null, "unbridged"),
};
```

Please also add the inverse of the new test: `agent_delegate` under the managed id, bridge
present, assert the record is visible. Without it the union has no guard and the next refactor
re-drops it.

### HIGH 2 — the legacy branch asserts `"ok"` on a claim that holds for one of three paths

`crates/trusty-mpm/src/daemon/mcp_backend.rs:104-134`. The new doc comment states:

> `#4141`: the legacy registry's `SessionId` IS the Claude session UUID (hook auto-registration
> uses it directly), so `delegations_for(id)` resolves correctly for that branch with no bridging.

Three paths register a legacy session, and only one matches that description:

| Path | Id source | Equals the Claude UUID? |
|---|---|---|
| `api.rs:1314` hook auto-register on `SessionStart` | `Session::new(session, …)` — the incoming hook id | Yes |
| `discovery.rs:141` tmux pane discovery | `Session::new(SessionId::new(), …)` | No — freshly minted |
| `discovery.rs:353` native-process discovery | `Session::new(SessionId::new(), …)` | No — freshly minted |
| `api.rs:625` `POST /sessions` (and `connect`, which delegates to it at `api.rs:694`) | `Session::new(SessionId::new(), …)` | No — freshly minted |

For a discovered or `POST`-registered legacy session, `delegations_for(id)` matches nothing for
the same structural reason as the managed bug — and this PR now stamps that empty result
`"delegation_lookup": "ok"`. The underlying blindness predates the PR; the affirmative soundness
claim does not, and `delegation_lookup` is the field whose entire purpose is to be trustworthy.

Fix: either narrow the doc comment to the hook-auto-register path and stop hard-coding `"ok"` on
the legacy branch (omit the field there rather than assert a soundness you can't establish), or
state in the comment that the legacy label is best-effort and why. Silently generalising one
path's property to all three is how the managed bug got written in the first place.

### MEDIUM 3 — `delegations: []` re-collapses the unknown the `null` was added to preserve

`crates/trusty-mpm/src/daemon/mcp_backend.rs:151`

```rust
None => (Value::Null, Value::Array(Vec::new()), "unbridged"),
```

`delegation_count: null` alongside `delegations: []` is self-contradictory, and #4141's constraint
is about the surface, not one field. A caller that reads `.delegations` (or `.delegations | length`)
rather than `.delegation_count` — the natural thing to do when you want the agent *list*, which is
precisely #2864's `/tm-session-pause` use case — gets an unambiguous "nothing in flight" from a
lookup that established no such thing.

Fix: `None => (Value::Null, Value::Null, "unbridged")`. Both fields then fail loudly together and
neither can be misread in isolation.

### LOW 4 — the `idle_nudge` precedent claim is inverted

The PR body and the change-site comment present this as *"the same pattern `daemon/idle_nudge.rs`
already uses."* `idle_nudge.rs:244-246` starts **from** a Claude UUID it received on the hook and
parses that; it never reads `claude_session_id` off a `SessionRecord` to reach delegations. Only
the `Uuid::parse_str` → `SessionId` mechanic is shared, not the bridging direction. Worth
tightening, since "there is prior art for this shape" is load-bearing in the justification.

---

## What came back clean

Stating these so the fix round doesn't re-derive them.

- **CHANGELOG gate satisfied.** Entry present under `## [Unreleased]` → `### Fixed` in
  `crates/trusty-mpm/CHANGELOG.md:12`, with the issue link.
- **The test reproduces the bug.** With `mcp_backend.rs` reverted to `origin/main` and the new
  tests kept, both fail: `session_status_bridges_managed_id_to_hook_observed_delegation`
  (`left: Number(0), right: 1`) and the negative assertion in
  `session_status_resolves_managed_session` (`left: Number(0), right: Null`). 14 passed, 2 failed.
  Restored, 16/16 pass. This is a genuine regression test, not a tautology.
- **The parse cannot produce a wrong-but-valid id.** `Uuid::parse_str(...).ok()` on
  `claude_session_id`, and the field's lifecycle is guarded on both ends —
  `hook_sync.rs:63` `clear_claude_session_id_if` clears only on an exact match, and
  `reactivate.rs:121` clears it on reactivation (#4337). A stale bridge pointing at a different
  live Claude session is not reachable through those paths.
- **`delegation_count: null` breaks no in-repo consumer.** A repo-wide grep finds
  `delegation_count` at exactly two sites, both inside `mcp_backend.rs` itself. Zero hits across
  all 7 UI crates (no `delegation` reference in any `.svelte`/`.ts`), zero in the bundled skills
  (`tm-session-pause`, `tm-session-management`, `tm-tool-usage-guide`), and the generated
  `tm-capabilities` MCP reference documents parameters only, never the response shape — so no
  regeneration is owed and the drift check has nothing to catch. `GET /api/console/sessions/{id}`
  (`trusty-console/src/routes/sessions.rs:117`) does proxy this JSON verbatim, but the Sessions
  tab never reads either field. No caller does arithmetic or a truthiness check on it. Both
  additions are additive; `null` is safe here. **The API-compat question is not the risk on this
  PR — Finding 1 is.**
- **Local gates clean.** `bash scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`;
  `cargo fmt --check -p trusty-mpm` silent; `cargo test -p trusty-mpm --lib mcp_backend` → 16
  passed, 0 failed.
- **No new flakiness.** The two new tests are fully in-process (`TempDir` + `DashMap`), touch no
  clock, socket, or filesystem race, and ran deterministically across four invocations. The
  pre-existing loopback / backoff flakes are unrelated and untouched.

## Suggestion (not a finding)

`record_to_json` (`daemon/managed_routes/summary.rs:45`) does not emit `claude_session_id`, so a
caller that receives `delegation_lookup: "unbridged"` has no way to see why, or to bridge itself.
Surfacing it would make the explicit-unknown actionable rather than merely visible.

## Required Changes

1. `mcp_backend.rs:139-152` — union `delegations_for(id)` with `delegations_for(SessionId(claude_uuid))`
   instead of replacing it, so `agent_delegate`-keyed records stop being reported as `0`/`"ok"`.
2. `mcp_backend_tests.rs` — add the inverse regression test: `agent_delegate` under the managed id,
   bridge present, record visible via `session_status`.
3. `mcp_backend.rs:151` — emit `Value::Null` for `delegations` in the unbridged arm.
4. `mcp_backend.rs:104-134` — correct the doc comment's legacy claim (it holds only for hook
   auto-register, not `discovery.rs:141`/`:353` or `api.rs:625`), and stop hard-coding
   `delegation_lookup: "ok"` on that branch.
5. `mcp_backend.rs:116-117` + PR body — narrow the `idle_nudge` precedent claim.

Items 1-3 are behavioural. 4 is a correctness claim in a comment that a future reader will
reasonably trust. 5 is wording.

bobmatnyc:
Fix round addressing the BLOCK verdict (commit 17ee144a on top of d4931616).

## Required changes

1. **Union, don't replace** (`mcp_backend.rs`, managed branch) — `delegations_for(id)` (the managed-id space `agent_delegate` writes under) is now read unconditionally; when the bridge resolves, `delegations_for(SessionId(claude_uuid))` is unioned in, de-duplicated by delegation id. `lookup="ok"` now means "both spaces read," not "one space read successfully."
2. **Inverse regression test added** — `session_status_union_keeps_managed_id_keyed_delegation_visible`: bridge present, delegation registered under the **managed** id (mirrors `agent_delegate_accepts_managed_session`'s key). Verified this fails against the reverted (replace-shape) code with the exact symptom from the probe: `delegation_count=0`, `delegation_lookup="ok"`, `delegations=[]` — then passes restored.
3. **`delegations` now `null` alongside `delegation_count: null`** in the unbridged arm — both fields fail loudly together; extended `session_status_resolves_managed_session` to assert this.
4. **`idle_nudge.rs` precedent claim** — checked the current PR diff, PR body, and commit `d4931616`'s content directly (`grep -n idle_nudge`): no such comparison exists anywhere in this PR's code comments or body. Nothing to correct; confirmed the rewritten doc comment doesn't introduce one either.
5. **Legacy branch no longer hard-codes `delegation_lookup: "ok"`** — the field is now omitted on that branch entirely, and the doc comment is narrowed to state the "`SessionId` IS the Claude UUID" property holds only for the `SessionStart`-hook-auto-registered path, not the tmux-discovery / native-process-discovery / `POST /sessions` paths, which mint a fresh id. That gap predates this fix and isn't newly claimed as sound.

## Gates (raw output, re-run after the fix)

- `cargo test -p trusty-mpm --lib mcp_backend` → **17 passed, 0 failed** (16 prior + the new inverse test)
- `cargo test -p trusty-mpm` (full crate) → **4035 passed, 0 failed, 6 ignored** (lib), plus all integration/doc-test binaries green — no flakes this run
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → clean
- `cargo fmt --check` → clean
- `bash scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`

CHANGELOG entry updated minimally to describe the union (not rewritten wholesale, given the other PRs touching that file).

bobmatnyc:
# Re-review — #4463 at `17ee144a`

## Verdict: APPROVE

The union is correct, the inverse test genuinely pins the regression, and the two
reporting-consistency items are closed. Zero CRITICAL, zero HIGH. Reviewed at
`17ee144a` in an isolated worktree; CI 23/23 pass at that exact SHA.

One of my five findings was mine to withdraw, not the author's to fix — see the
adjudication below. I got that one wrong and the author's account of it is accurate.

## Prior findings — status

| # | Prior severity | Status | Evidence |
|---|---|---|---|
| 1 | CRITICAL — bridge replaced the managed-id lookup | **RESOLVED** | Probe now reports `delegation_count=1 delegation_lookup="ok"` where it reported `0`/`"ok"` at `d4931616` |
| 2 | — inverse regression test | **RESOLVED** | Reverted the union alone; the new test fails with the exact probed symptom |
| 3 | MEDIUM — `delegations: []` beside `delegation_count: null` | **RESOLVED** | `None => (Value::Null, Value::Null, "unbridged")` (`mcp_backend.rs:176`), asserted at `mcp_backend_tests.rs:131-135` |
| 4 | LOW — inverted `idle_nudge` precedent claim | **WITHDRAWN — my error** | See adjudication |
| 5 | HIGH — legacy branch hard-coded `"ok"` | **RESOLVED** | Field removed from the legacy arm; doc comment narrowed and now accurate |

### 1 — union verified, and the dedup cannot drop anything

`mcp_backend.rs:157` reads the managed-id space **before** the match, so it is
unconditional — not gated on bridge success. The Claude-UUID space unions in at
`:164-168`.

Re-ran the original probe shape (managed record, `claude_session_id` correlated, then
`agent_delegate` with the managed id exactly as `session_list` hands it out):

```
PROBE-A agent_delegate -> {"agent":"rust-engineer","circuit":"closed","delegation_id":"17d5…","tier":"sonnet"}
PROBE-A: delegation_count=1 delegation_lookup="ok" delegations_len=Some(1)
```

Was `delegation_count=0 delegation_lookup="ok"` at `d4931616`. Fixed.

The dedup by delegation id cannot drop a legitimately distinct delegation.
`DaemonState.delegations` is a `DashMap` keyed by `delegation.id.0`
(`state/sessions.rs:502-503`) and `DelegationId::new` is a fresh v4
(`core/agent.rs:112-115`), so two distinct delegations necessarily have distinct ids —
the `existing.id == d.id` guard can only fire when `record.id.0 == claude_uuid`, where
both reads return the same set and the guard is exactly what prevents a double count.
Safe in both directions.

### 2 — the inverse test pins the regression

Reverted **only** the union back to replace-shape, left everything else at `17ee144a`:

```
thread '…session_status_union_keeps_managed_id_keyed_delegation_visible' panicked:
assertion `left == right` failed: an agent_delegate-keyed (managed-id) delegation must
remain visible after the bridge fix, not be dropped by it (#4141), got:
{"delegation_count":0,"delegation_lookup":"ok","delegations":[],"kind":"managed",…}
  left: Number(0)
 right: 1
test result: FAILED. 18 passed; 1 failed; 0 ignored
```

Exactly the probed symptom — `delegation_count=0` under an affirmative
`delegation_lookup="ok"`. Restored the union byte-for-byte (`git diff` on the file
empty) and it passes. This is a real guard, not a tautology.

### 3 — both fields fail loudly together

No arm can emit a confident `[]` beside a null count. In the `"ok"` arm both fields
derive from the same vec (`json!(delegations.len())` / `json!(delegations)`), so
`delegation_count == delegations.len()` holds by construction; in the unbridged arm
both are `Value::Null`.

### 5 — the legacy arm is now shape-identical to `main`

Stronger than "coherent": `git grep delegation_lookup origin/main` returns **zero
hits**. The field is new in this PR, and removing it from the legacy arm restores
`main`'s exact legacy response. So the "three shapes" concern collapses — the field is
managed-branch-only, and `serde_json` indexing yields `Value::Null` for a missing key,
so a consumer switching on it reads legacy as "unknown" rather than a false `"ok"`.

The narrowed doc comment (`:120-128`) is accurate. It names all three non-hook paths —
tmux-pane discovery, native-process discovery, `POST /sessions` — and confines the
"`SessionId` IS the Claude UUID" property to the `SessionStart`-hook-auto-registered
path. That matches the registration table in my prior comment.

## Adjudication — the `idle_nudge` dispute

**The author is right. The claim never existed in this PR. My finding 4 was
misattributed and I withdraw it.**

Evidence, four independent checks:

- `git show d4931616:crates/trusty-mpm/src/daemon/mcp_backend.rs | grep -i idle_nudge` — no match. Same at `17ee144a`.
- The full diff at *both* heads, and every commit message on the branch — no match.
- The PR body as it stands — no match.
- The body was never edited, so it cannot have been silently removed in the fix round: `userContentEdits { totalCount }` on the PR is **0**.

The only two occurrences of `idle_nudge` anywhere on this PR are my own prior verdict
and the author's reply to it. I read the comparison in the implementer's narrative to
the coordinator and cited it as if it were in the artifact. It was not. Correcting the
record: the author's grep was right, my citation was wrong, and there was nothing to
fix.

## New findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/daemon/mcp_backend.rs` | 117-118, 156 | "de-duplicates by delegation id" is true but does not cover the real duplicate case: a declare-then-dispatch pair for ONE subagent now counts as 2 | Say what the dedup does and does not protect against; note that `dedup` is same-session-scoped and cannot span the bridge |
| LOW | PR body | — | The body still carries the pre-fix rationale the code now contradicts, and squash-merge makes it the permanent commit message | Update the Fix section before merge |

### MEDIUM — the union surfaces declare-then-dispatch as 2 for one subagent

Probed it. A PM that calls `agent_delegate` **and** dispatches — the case
`delegation_tracker::dedup` exists to handle — writes two records under two different
keys for a managed session: the `McpDeclared` one under the managed id, the
`HookObserved` one under the Claude UUID. `dedup` (`delegation_tracker.rs:239-241`)
matches "the most recent **same-session**, same-agent delegation", so it cannot reach
across the bridge and never merges them.

```
PROBE-B (ONE logical subagent, declared+dispatched):
  delegation_count=2 lookup="ok" managed_space=1 claude_space=1
PROBE-B delegations = ["engineer/queued", "engineer/running"]
```

That is 1 under replace-shape and 1 on `main` — each saw only one of the two records.
The union is the first version to surface both, so the arithmetic is new to this PR
even though the duplicate records are not.

Why this is MEDIUM and not a blocker:

- Both records genuinely exist in state on `main` too; the union reports what state
  holds. The duplicate-record gap is `dedup`'s key scoping, which this PR does not touch.
- The repo's stated doctrine prefers this direction, verbatim in `dedup`'s own doc
  comment: *"two records for one subagent overcounts and is corrected by the staleness
  sweep, whereas a false merge undercounts work in flight and mislabels it, with no
  recovery."* Cross-space dedup is exactly the false-merge risk that comment warns about.
- `delegations_for` filters by session, not status, so `delegation_count` already
  included terminal delegations on both branches. It was never a live count; this is not
  a new class of inaccuracy.
- The array disambiguates — `queued` + `running` is readable by any consumer that wants
  the list rather than the number.

So the fix I am asking for is the comment, not the behaviour: the current wording reads
as though duplicates are handled, and the next reader will believe it. If cross-space
dedup is wanted, that is a separate ticket against `delegation_tracker`, not this PR.

### LOW — the PR body will become the merge commit

The body has never been edited, so it still says the bridge parses into "the same key
hook ingestion **and `agent_delegate`** already use", and that "the legacy branch is
unaffected: its `SessionId` already IS the Claude session UUID". Those are precisely the
two claims the fix round corrected in code (CRITICAL 1 and HIGH 2). This repo
squash-merges, so the body lands as the permanent rationale for a commit that now
contradicts it.

## Gates — re-verified at `17ee144a`, raw output

- `cargo test -p trusty-mpm --lib mcp_backend` → `test result: ok. 17 passed; 0 failed; 0 ignored`
- `cargo test -p trusty-mpm` (full crate, no `--lib`) → `4035 passed; 0 failed; 6 ignored` for the lib, **6946 passed / 0 failed / 16 ignored** summed across every binary and doc-test
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → exit 0, zero errors
- `cargo fmt --check -p trusty-mpm` → silent
- `bash scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`
- CI at `17ee144a` → 23 success, 1 skipped, 0 failed
- No `#[ignore]`, no `cfg`-gate, no scope narrowing: the diff adds none, and the test
  count in `mcp_backend_tests.rs` goes 15 (`main`) → 17, with nothing removed. The 16
  ignored tests are pre-existing and in unrelated binaries.

## Notes

- **No lock-ordering or double-borrow hazard.** `session_manager()` returns
  `Arc<SessionManager>` from a `OnceCell` (`state/core.rs:653`), not a guard, so nothing
  is held across the delegation reads. Each `delegations_for` takes and releases its
  `DashMap` read guards and returns an owned `Vec`, so the second read cannot overlap the
  first — the shard self-deadlock that `find_delegation`'s doc comment warns about is not
  reachable here. Two full map scans plus an O(n·m) dedup, on a per-request MCP path with
  one process's delegations in the map: not a performance concern.
- **The unbridged arm now drops the known managed-id half** (both fields `null` where
  `main` returned the real list). That is the trade-off my own required change 3 asked
  for, so I am not reopening it — but noting the cost explicitly: a caller that could
  have been told "at least 1 in flight, total unknown" is now told only "unknown". Still
  an explicit unknown, which is what #4141 requires, so this is a record, not a finding.
- The unbridged arm computes `delegations_for(id)` and discards it — one wasted map scan.
  Too small to be worth a finding.
- CHANGELOG entry updated to describe the union, still under `## [Unreleased]` → `### Fixed`
  with the issue link. Gate satisfied.
- Confirmed sound last round and undisturbed by `17ee144a`: the two original tests, the
  `Uuid::parse_str(...).ok()` parse being unable to yield a wrong-but-valid id, and
  `delegation_count: null` breaking no in-repo consumer.

Neither new finding blocks. Both are one-line writing changes and are fine to land
alongside the merge or as a follow-up.

Issue #4142, opened 2026-07-27
trusty-mpm: out-of-order SubagentStop is unrecoverable — PostToolUse is async, SubagentStop is sync


bobmatnyc:
## Summary

`PostToolUse` and `SubagentStop` are delivered by **independent processes with no ordering guarantee**, but the delegation tracker can only resolve a `SubagentStop` using an `agent_id` that *only* `PostToolUse` teaches it. When `SubagentStop` arrives first, it matches nothing, terminalizes nothing, and **nothing ever re-checks** — the delegation stays pinned `Running` until the 6-hour staleness sweep.

Found during a full-context code-critic gap review of PR #4096, which identified findings living *outside* that PR's changed files. Source review comment:
https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094071838

Deliberately held out of #4096's current fix round to keep that PR from growing without bound — filed here so it survives.

## Why the order isn't guaranteed

In `crates/trusty-mpm/src/core/standalone/hooks/mod.rs`:

- `PostToolUse` is installed with `"async": true` (`:153-160`).
- `SubagentStop` is installed **synchronously** (`:170-176`).

Each hook runs as its own `tm hook` process making its own independent POST to the daemon. An async `PostToolUse` can be scheduled, delayed, or descheduled while the synchronous `SubagentStop` for the same subagent completes and posts first. Arrival order is therefore a race, not a contract.

## Why the out-of-order case is unrecoverable

`on_subagent_stop` resolves a delegation **only by `agent_id`**. That `agent_id` is learned exclusively from the `PostToolUse` event. So when `SubagentStop` lands first:

1. The lookup finds no delegation.
2. Nothing is terminalized.
3. There is **no retry, no deferred queue, and no re-check** on the later `PostToolUse` arrival.
4. The delegation remains `Running` indefinitely — until the 6 h staleness sweep reaps it.

A delegation stuck `Running` for up to six hours is a phantom in-flight agent, which corrupts exactly the "what is in flight" answer that #2864 needs for `/tm-session-pause`.

## ⚠️ Trap for whoever picks this up — these two changes are coupled

Today the synchronous case is **masked**, not handled. In `delegation_tracker.rs`, terminalization on the sync path currently rides on:

```rust
sync_return = response.is_some()
```

That loose condition is presently the sync path's **only** terminalization route. It happens to catch the case by accident.

[...]

PR #4464, opened 2026-07-31
fix(trusty-mpm): unify agentId usability check across delegation tracker

### LINKED ISSUE #4163: bug(trusty-mpm): agentId presence-check lets a null or empty id pin a delegation Running for 6h
Follow-up from the final re-review of PR #4096, which returned **APPROVE (mergeable as-is)** with this as a non-blocking MEDIUM finding. Filed so it survives the merge.

Review comment: https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094980860

## The defect

`crates/trusty-mpm/src/daemon/services/delegation_tracker.rs:463` decides "we have an agent id" with:

```rust
r.get("agentId").is_some()
```

That tests **key presence only**. Two other sites in the same file disagree with it about what a usable id is:

- `:334` consumes the value via `as_str` — so a non-string value yields nothing.
- `:499` explicitly rejects an empty string.

The presence-check and the consumers are therefore not asking the same question.

## Probed behaviour (against the merged head)

| Payload | Result |
|---|---|
| `{"agentId": null}` | delegation transitions to `Running` with `agent_id = None` |
| `{"agentId": ""}` | delegation transitions to `Running` with `agent_id = Some("")`, which a `SubagentStop` **can never match** |

## Why it matters

This is the mirror image of the contradiction the last fix round removed. The code records "expect a stop for this agent" while holding **no handle that a stop can quote**. Nothing can ever resolve the record, so the delegation burns the full **6 h staleness window** as a phantom in-flight entry.

That is precisely the cost the author explicitly refused to trade for when rejecting the fallback-key option earlier in the PR — so reintroducing it through a weaker presence-check is the same bug wearing a different hat.

## Severity reasoning — recorded so nobody re-derives it

Severity is **MEDIUM, not HIGH**, for one reason only: the `null` / `""` payload shape is **unobserved against Claude Code 2.1.220**. It is not defended against, merely not currently emitted. A payload change upstream could make this live **with no code change on our side**. Do not downgrade it further on the strength of "we've never seen it" — that is already the entire reason it isn't HIGH.

## Fix (from the reviewer, verified)

```rust
.and_then(Value::as_str).filter(|s| !s.is_empty())
```

Ideally compute this **once** and pass the resulting value into `classify_dispatch`, so the presence test and the consumption are structurally incapable of disagreeing.

## General principle worth a regression test

Any site that decides *"we have an identifier"* must use the **same extraction the consumer uses**. A presence-check and a value-consumer that disagree is the bug — not an incidental detail of this one call site. A test that drives `null` and `""` through the dispatch path and asserts the delegation does **not** enter `Running` would lock this shut.

## References

- Parent epic: #2864
- Source PR: #4096
- Related ordering/recovery work: #4142

bobmatnyc:
## Problem

`delegation_tracker.rs:463`'s `classify_dispatch` decided "we have an agent id" with key-presence alone: `r.get("agentId").is_some()`. Two other sites in the same file disagreed about what a usable id is — `on_launched` consumes via `as_str` (a non-string value yields nothing), and `on_subagent_stop` explicitly rejects an empty string via `field()`.

Probed against the merged head:

| Payload | Result |
|---|---|
| `{"agentId": null}` | delegation transitions to `Running` with `agent_id = None` |
| `{"agentId": ""}` | delegation transitions to `Running` with `agent_id = Some("")`, which a `SubagentStop` can never match |

Either way, the delegation records "expect a stop for this agent" while holding no handle a stop can ever quote — burning the full 6h staleness window as a phantom in-flight entry, the exact cost a prior fix round (PR #4096) explicitly refused to trade for.

## Fix

Added `usable_agent_id(&Value) -> Option<&str>` (non-null, non-empty string) and used it at both:
- `classify_dispatch`'s presence-check (was `r.get("agentId").is_some()`)
- `on_launched`'s consumption (was `.and_then(Value::as_str)` without the empty-string filter)

Both call sites now share one extraction, so the presence-check and the consumers are structurally incapable of disagreeing again.

## Tests

Added `null_agent_id_does_not_launch_a_phantom_delegation` and `empty_string_agent_id_does_not_launch_a_phantom_delegation`, covering both probed payloads from the issue. Both assert the delegation terminalizes as `Completed` (an ordinary synchronous return, since neither payload carries any other launch marker) rather than lingering `Running` with an unresolvable `agent_id`.

Closes #4163

## Gates (raw output)

- `cargo test -p trusty-mpm` (full crate, not `--lib`): all 13 suites `ok`, 0 failed
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean
- `bash scripts/check_line_cap.sh`: `7 allowlisted, 0 violations — OK.`

### DIFF
--- a/crates/trusty-mpm/CHANGELOG.md
+++ b/crates/trusty-mpm/CHANGELOG.md
@@ -9,6 +9,17 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- delegation-tracker `agentId` presence-check no longer accepts `null`/`""` (closes [#4163](https://github.com/bobmatnyc/trusty-tools/issues/4163))
+  - `classify_dispatch` decided "we have an agent id" with key-presence alone
+    (`r.get("agentId").is_some()`), while `on_launched`'s consumer rejected a
+    non-string value and `on_subagent_stop`'s consumer rejected an empty
+    string. A `{"agentId": null}` or `{"agentId": ""}` dispatch response took
+    the `Launched` branch and pinned the delegation `Running` with a handle no
+    `SubagentStop` can ever quote back, burning the full 6h staleness window as
+    a phantom in-flight entry. Both call sites now share one extraction,
+    `usable_agent_id` (non-null, non-empty string), so the presence-check and
+    the consumers are structurally incapable of disagreeing again.
+
 - managed sessions can reach the bundled agent roster again — every delegation was degrading to `general-purpose` (closes [#4451](https://github.com/bobmatnyc/trusty-tools/issues/4451))
   - Daemon-managed spawns now launch with `--setting-sources user,project,local`
     instead of `project,local`. Claude Code discovers subagents per settings
--- a/crates/trusty-mpm/src/daemon/services/delegation_tracker.rs
+++ b/crates/trusty-mpm/src/daemon/services/delegation_tracker.rs
@@ -172,6 +172,28 @@ fn field<'a>(payload: &'a Value, key: &str) -> Option<&'a str> {
         .filter(|s| !s.is_empty())
 }
 
+/// The `agentId` a `tool_response` actually offers as a usable correlation
+/// handle: present, a string, and non-empty.
+///
+/// Why (#4163): [`classify_dispatch`]'s presence-check and [`on_launched`]'s
+/// consumption used to ask different questions —
+/// `r.get("agentId").is_some()` treats `null` and `""` as "we have an id",
+/// while the consumer's `.and_then(Value::as_str)` (for `null`) or
+/// [`on_subagent_stop`]'s `field` (for `""`) both reject them. A response
+/// carrying either shape took the `Launched` branch, pinning the delegation
+/// `Running` with an `agent_id` no `SubagentStop` can ever quote back —
+/// burning the full 6 h staleness window as a phantom in-flight entry. One
+/// extraction used by every site that asks "do we have an agent id" makes
+/// them structurally incapable of disagreeing again.
+/// Test: `null_agent_id_does_not_launch_a_phantom_delegation`,
+/// `empty_string_agent_id_does_not_launch_a_phantom_delegation`.
+fn usable_agent_id(response: &Value) -> Option<&str> {
+    response
+        .get("agentId")
+        .and_then(Value::as_str)
+        .filter(|s| !s.is_empty())
+}
+
 /// `PreToolUse` on a dispatch tool: a subagent is starting now.
 ///
 /// Why: this is the moment tracking must record a *live* child, so a PM with
@@ -316,7 +338,9 @@ fn tasks_match(declared: &str, dispatched: &str) -> bool {
 /// `post_tool_use_without_tool_response_stays_running`,
 /// `post_tool_use_with_unrecognized_response_stays_running`,
 /// `post_tool_use_with_only_an_agent_id_stays_running`,
-/// `changed_async_status_value_with_an_agent_id_stays_running`.
+/// `changed_async_status_value_with_an_agent_id_stays_running`,
+/// `null_agent_id_does_not_launch_a_phantom_delegation`,
+/// `empty_string_agent_id_does_not_launch_a_phantom_delegation`.
 fn on_launched(state: &DaemonState, session: SessionId, payload: &Value, event: HookEvent) {
     let Some(tool_use_id) = field(payload, "tool_use_id") else {
         return;
@@ -331,10 +355,10 @@ fn on_launched(state: &DaemonState, session: SessionId, payload: &Value, event: 
     let response = payload
         .get("tool_response")
         .filter(|r| recognized_response(r));
-    let agent_id = response
-        .and_then(|r| r.get("agentId"))
-        .and_then(Value::as_str)
-        .map(str::to_string);
+    // #4163: use the same extraction `classify_dispatch` uses for its
+    // presence-check, so a null/empty `agentId` is never stored as a handle
+    // nothing can ever resolve.
+    let agent_id = response.and_then(usable_agent_id).map(str::to_string);
     let tier = response
         .and_then(|r| r.get("resolvedModel"))
         .and_then(Value::as_str)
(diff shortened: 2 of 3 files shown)


bobmatnyc:
## Code Critic — PR #4464, head `62bf7834`

**Scope**: the full diff (3 files, +122/−12) against #4163 and the round-4 finding it came from ([#4096 comment](https://github.com/bobmatnyc/trusty-tools/pull/4096#issuecomment-5094980860)). Reviewed in an isolated worktree at the PR head; every claim below was executed, not read.

## Verdict: APPROVE

The defect is closed, the fix is at the right layer, and both production edits are independently pinned by the new tests. Two non-blocking findings (1 MEDIUM, 1 LOW), neither a merge gate.

---

### The load-bearing check: do all three sites now agree?

The issue named three disagreeing sites. Site C (`on_subagent_stop`'s `field(payload, "agent_id")`, `delegation_tracker.rs:532`) is untouched by this PR, so "it already rejects empty strings" needed verifying against the *whole* input space, not just `""`. I drove every shape through all three sites plus a real `SubagentStop` (temporary probe, reverted; tree clean at `62bf7834`):

| `tool_response.agentId` | `recognized_response` | `usable_agent_id` (sites A+B) | `field(…,"agent_id")` (site C) | classify | observed status | stored `agent_id` | a stop quoting the same value resolves it? |
|---|---|---|---|---|---|---|---|
| key absent | false | `None` | `None` | `Unknown` | `Running` | `None` | no (documented `Unknown` band) |
| `null` | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `""` | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `" "` (whitespace) | true | `Some(" ")` | `Some(" ")` | `Launched` | `Running` | `Some(" ")` | **yes** |
| `42` (number) | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `true` (bool) | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `{}` (object) | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `[]` (array) | true | `None` | `None` | `Returned` | `Completed` | `None` | n/a — terminal |
| `"aid"` | true | `Some("aid")` | `Some("aid")` | `Launched` | `Running` | `Some("aid")` | yes |

All three sites agree on all nine inputs. Two results worth calling out:

- **Whitespace-only is not a latent phantom.** `" "` is usable at every site *and* a real `SubagentStop` quoting `" "` resolves the delegation (probe: `true`). It is an ugly handle, not an unresolvable one — no residual 6h burn.
- **Non-string types were part of the same defect and are also fixed.** `{"agentId": 42}` passed the old `is_some()` too, then yielded `agent_id = None` from `as_str` — the identical phantom. It now falls to `Returned`. Untested (see Notes) but verified correct today.

### Do the new tests fail against unfixed code?

Isolation matrix — each production edit reverted **alone**:

| Mutation | Result |
|---|---|
| `classify_dispatch` back to `r.get("agentId").is_some()` | `test result: FAILED. 72 passed; 2 failed` — both new tests fail with `left: Running, right: Completed`, the exact issue symptom, and they are the **only** failures |
| `on_launched` back to `.and_then(Value::as_str)` (classify left fixed) | `test result: FAILED. 73 passed; 1 failed` — `empty_string_…` fails on `left: Some(""), right: None` |

Both edits are load-bearing and separately pinned. Neither test passes pre-fix.

### Gates, raw

```
bash scripts/check_line_cap.sh      -> line-cap: 7 allowlisted, 0 violations — OK.
cargo fmt --check                   -> clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings -> exit 0
cargo test -p trusty-mpm            -> exit 0; 44 suites all "ok"; 6946 passed, 0 failed, 16 ignored (pre-existing; the diff adds no #[ignore], no cfg-gate, no exclusion)
cargo test -p trusty-mpm --lib delegation -> 74 passed; 0 failed; 0 ignored
```

CHANGELOG entry present under `## [Unreleased]` → `### Fixed`. `// #4163:` pointers present at both change sites (`:358`, `:492`). Both new test names in doc `Test:` lines resolve to real tests.

---

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/daemon/services/delegation_tracker.rs` | 190-195 | `usable_agent_id` is a verbatim reimplementation of `field` — two independent definitions of the same predicate, which is the shape of the original defect | `fn usable_agent_id(response: &Value) -> Option<&str> { field(response, "agentId") }` |
| LOW | `crates/trusty-mpm/src/daemon/services/delegation_tracker.rs` | 453, 463 | The KNOWN LIMITATION section's trigger is now narrower than the code's | Say "no *usable* `agentId`" and link `usable_agent_id` |

### MEDIUM — the new helper duplicates `field` instead of calling it

```rust
// :168-173
fn field<'a>(payload: &'a Value, key: &str) -> Option<&'a str> {
    payload.get(key).and_then(Value::as_str).filter(|s| !s.is_empty())
}

// :190-195
fn usable_agent_id(response: &Value) -> Option<&str> {
    response.get("agentId").and_then(Value::as_str).filter(|s| !s.is_empty())
}
```

Same three combinators, 20 lines apart. The PR's stated guarantee is that the sites are "structurally incapable of disagreeing again" — but site C reaches its answer through `field` and sites A+B through a private copy. Today they agree by coincidence of identical bodies, not by construction. The next edit to either notion of "usable string field" (trimming, a length floor, a charset check) reintroduces exactly the drift #4163 is about, and this time between a helper and its own clone rather than between a check and a consumer.

Fix, verified locally — compiles and keeps the suite green (`74 passed; 0 failed`):

```rust
fn usable_agent_id(response: &Value) -> Option<&str> {
    field(response, "agentId")
}
```

The named wrapper keeps the call-site readability and the doc comment; it just stops owning a second copy of the rule.

### LOW — the KNOWN LIMITATION doc's trigger set is now stale

```
/// # KNOWN LIMITATION — an `agentId`-less response is read as a return
...
/// `agent_id` is taught *only* by this function. A response with no `agentId`
/// therefore has **no** route to termination other than this branch
```

After this PR the fail-open band is entered by "no *usable* `agentId`" — which now includes `null`, `""`, and every non-string type, per the table above. The `Launched` bullet at `:441-448` states the new rule, so the file is not self-contradictory, but the section a reader consults for the boundary of the band still describes the narrower pre-PR trigger. Suggest: retitle to "a response with no *usable* `agentId` is read as a return", and in the body write "A response with no usable `agentId` ([`usable_agent_id`]: absent, `null`, `""`, or not a string) therefore has no route to termination other than this branch."

---

## Notes

- **The not-usable path terminalizes `Completed`, and that is the spec.** #4163's prescribed fix says so in as many words ("falls to `Returned` → `Completed`, which is the correct reading"). It is also consistent with the module's pre-existing, documented fail-open band for liveness-silent recognized responses, whose justification — no usable handle means no other route to termination, so an affirmative reading would manufacture a guaranteed 6h phantom — applies verbatim here. Not re-litigated.
- **One asymmetry a future reader will trip on, deliberately not flagged**: `{}` (no recognized key) stays `Running` as `Unknown`, while `{"agentId": null}` — which conveys no more liveness information — terminalizes `Completed`. That is the `Unknown` vs `Returned` distinction working as designed, not a defect; flagging it here would just relitigate the bullet above.
- **Not flagged: no test pins the non-string shapes** (`42`, `{}`, `[]`). The `null` test already covers the same branch of `usable_agent_id`, and my probe confirms current behavior is correct, so a dedicated case would add rows without adding pinning power. Worth one table-driven case if this helper is ever touched again.
- **Untouched drift-resistance caveat**: the `Launched`-first ordering still survives a drifted `status` *value*, because such a response carries a real string `agentId`. A hypothetical double drift (`status` value changes *and* `agentId` goes `null`) would now read as a return. That is the accepted fail-open band, tracked with the stop-side recovery work in #4142, not this PR's business.
- **PR is `BEHIND` main.** No source drift: `git log <merge-base>..origin/main -- crates/trusty-mpm/src/daemon/` is empty, and GitHub reports `MERGEABLE`. Only the CHANGELOG could want a trivial rebase.

Zero CRITICAL, zero HIGH. **APPROVE** — the MEDIUM is a one-line cleanup that can land here or not at all.

Issue #4407, opened 2026-07-30
unticketed flake cluster: 7 TUI/plain-mode tests failed together on main (2026-07-27)


bobmatnyc:
**Defect:** a CI run on `main` (2026-07-27 13:09) failed **7 TUI/plain-mode tests simultaneously**. The next push 44 minutes later passed clean, with no revert in between. No issue was ever filed.

**Why it's flake-shaped, not a regression:** many unrelated tests failing together and clearing on the next push without a code change is the signature of shared global-state pollution, not a real defect. Same family as the hermetic-test-isolation epic #3451.

**Supporting evidence from the same audit:** an independent clean-`main` local run hit 7/20,317 failures CI history never showed. One — `semantic_consolidation::inference_available_false_without_key` — is self-documented at `crates/trusty-common/src/memory_core/semantic_consolidation/mod.rs:98-110` as depending on `OPENROUTER_API_KEY` being **absent** from the ambient shell. The runner's shell had it set. Tests that depend on ambient env are non-hermetic by construction.

**Three further single-occurrence failures with no filed issue:** `memguard::…test_trim_heap_never_increases_rss_after_bulk_free`, `tmux::…probe_history_limit_reads_back_configured_value`, `session_resume_headless_dead_runtime_reconciles_and_restarts`.

**Cost:** each flake burns a full ~17-minute job. Over a measured 5-day window, 10 of 16 `Test`-job failures were flake-shaped — roughly 1.7–2.8 hours of wasted CI wall-clock in 5 days (extrapolated from the measured sample).

**Resolution:** reproduce the TUI cluster, then root-cause under #3451's hermetic-isolation umbrella. Separately, audit for other tests that read ambient environment.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
**New cluster member, with reproduction: `statusline::branch::tests::project_segment_basename_fallback_non_git`.**

Non-deterministic by construction, not by timing luck. `tmux_session_name()` shells out to `tmux` with a 100 ms timeout; the test calls it twice and requires both probes to agree. Under machine load either probe can time out independently, so the two disagree and the assertion fails.

Evidence: three isolated local runs gave ok / FAILED / ok on the same tree. The failure is outside PR #4466's diff and reproduced pre-merge on the same machine.

[...]

Issue #4230, opened 2026-07-28
an unsupervised `tm daemon` orphan can seize port 7880 while launchd reports the job not running


bobmatnyc:
**Defect**
A bare `tm daemon` process can bind 127.0.0.1:7880 and serve all live session traffic while launchd's `com.trusty.mpm` job reports `state = not running`. #2486 shipped a guard so a `tm serve --stdio` bridge REFUSES to auto-spawn when a launchd unit is registered; that fix predates 1.0.2, so the binary in question should have had it. Something spawned one anyway. The spawn path is UNKNOWN and is the actual thing to find — either the #2486 guard regressed, or the spawn came from a path it does not cover.

**Evidence** (observed 2026-07-28 ~07:45 UTC, before any remediation)
- launchd `com.trusty.mpm`: `state = not running`, `last exit code = 0`, `runs = 1`. Graceful shutdown 2026-07-22T18:42:09Z. `KeepAlive.SuccessfulExit = false`, so launchd never respawned it.
- Port 7880 nonetheless served by PID 98606, `/Users/masa/.cargo/bin/tm daemon`, PPID 1, started 2026-07-26 18:02:55, uptime ~1d13h.
- `GET /health` -> `{"status":"ok","supervised":false,"version":"1.0.2"}` — old binary, unsupervised, answering 200.
- Remediated by `kill -TERM 98606` then bootout/bootstrap; now PID 98401, launchd-owned, `supervised:true`, `version:1.2.3`.

**Why it matters**
- Silent substitution: a fresh signed install plus behavioral verification can all pass while the orphan keeps serving the OLD binary. A deploy reads as successful and ships nothing. Nearly happened with the 1.2.3 deploy today.
- The documented restart convention makes it worse, silently: `launchctl bootout` is a no-op against a process launchd doesn't own, and `bootstrap` then crash-loops every ~10s (`ThrottleInterval=10`, `KeepAlive.SuccessfulExit=false`) failing to bind the already-held port, while the orphan keeps serving.

**Resolution (proposed)**
1. Find and close the spawn path — determine what ran a bare `tm daemon` on 2026-07-26 and whether #2486's guard regressed or has a gap.
2. Make `bootstrap` failure loud: detect a foreign listener on the target port and fail with an actionable message naming the orphan PID, rather than crash-looping silently.
[...]

PR #4465, opened 2026-07-31
fix(trusty-mpm): recognize trusty-mpm-launched sessions in auto-discovery

### LINKED ISSUE #4058: bug(trusty-mpm): sessions launched via trusty-mpm binary never auto-discovered (discovery.rs missing CLAUDE_COMMANDS entry)
## Summary

Sessions launched via the `trusty-mpm` binary are invisible to auto-discovery, even though the crate ships both `tm` and `trusty-mpm` as identical binaries from the same source. Auto-discovery only recognizes the `tm` binary name.

## Root Cause

`crates/trusty-mpm/src/daemon/discovery.rs:32` defines:
```rust
const CLAUDE_COMMANDS: &[&str] = &["claude", "claude-code", "claude-mpm", "tm"];
```

The constant is missing `"trusty-mpm"`, so `is_claude_command()` never matches sessions running under the `trusty-mpm` binary.

Meanwhile, `crates/trusty-mpm/Cargo.toml:124-132` ships two identical `[[bin]]` targets from the same source:
```toml
[[bin]]
name = "tm"
path = "src/bin/tm/main.rs"
required-features = ["cli"]

[[bin]]
name = "trusty-mpm"
path = "src/bin/tm/main.rs"
required-features = ["cli"]
```

`cargo install trusty-mpm` installs both binaries. A session started via `tm` is discovered; an identical session started via `trusty-mpm` is silently invisible.

## Why This Matters

- **Silent failure**: The session runs normally, but never appears in `GET /sessions` or other discovery lists. No error, no warning.
- **Operator confusion**: The operator has no signal explaining why a session they started is invisible.
- **Inconsistency**: Identical sessions have different visibility based on which binary name was used to launch them.

## Underlying Problem: Hand-Maintained Binary Name Tables

The same fact (what this crate's binary is called) is written in multiple places and they have drifted:
- `CLAUDE_COMMANDS` (discovery.rs) — missing `"trusty-mpm"`
- `MPM_BIN_NAMES` (also exists, check codebase)
- `STATUSLINE_BIN_NAMES` (also exists, check codebase)
- `is_tm_binary()` (also exists, check codebase)

Each table is a hand-maintained source of truth, so they will keep drifting. This is the structural problem.

## Suggested Direction

Recognize every binary name the crate actually ships in the discovery list, OR derive all binary-name tables from a single shared constant rather than maintaining separate hand-written lists. This is a prerequisite to safely handling binary-naming decisions (issue #XXXX).

## Related Issues

- #3811: gworkspace MCP binary not reinstalled under new name (related binary-naming drift)
- (Issue #2 filed separately): LaunchAgent files and installer paths disagree on which binary to use

bobmatnyc:
## Problem

`daemon::discovery`'s `CLAUDE_COMMANDS` list only recognized `"tm"`, not `"trusty-mpm"` — the crate's second `[[bin]]` target, built from the identical `src/bin/tm/main.rs` source (`Cargo.toml:124-132`). `cargo install trusty-mpm` ships both binaries. A session launched via `tm` was auto-discovered; an identical session launched via `trusty-mpm` was silently invisible to `GET /sessions` and every other discovery list — no error, no warning.

## Fix

Two parts:

1. **The bug**: `is_claude_command()` now matches `trusty-mpm` (treated like the other long, distinctive launcher names — substring match — while `tm` keeps its exact-match requirement so it never matches unrelated binaries like `vim`).
2. **The structural problem the issue flagged**: this crate had FOUR independent hand-copies of "the names this crate's own binary goes by" — `discovery::CLAUDE_COMMANDS`, `hooks::MPM_BIN_NAMES`, `session_launch::settings::STATUSLINE_BIN_NAMES`, and an inline `name == "trusty-mpm" || name == "tm"` check in the `tm stop` daemon-PID scan (`bin/tm/commands/daemon.rs`) — and they had already drifted once (this bug). All four now derive from one canonical `core::own_binary_names::OWN_BINARY_NAMES` constant, so adding a third `[[bin]]` target in the future only requires updating one array.

I chose the structural fix over the one-line patch because the four lists already existed side by side, the correct canonical list (`hooks::MPM_BIN_NAMES`) already existed and was already documented as "canonical", and collapsing four call sites onto it was a small, contained, mechanical change — not a new abstraction layer.

## Other hand-maintained lists (found, in scope, now fixed)

- `core::standalone::hooks::MPM_BIN_NAMES`
- `core::session_launch::settings::STATUSLINE_BIN_NAMES`
- inline check in `bin::tm::commands::daemon::find_daemon_pids`

`core::standalone::hooks::MPM_BIN_STEMS` (hash-stripped build-artifact stems, includes the retired `session_manager_mvp` name) and `daemon::discovery::is_claude_process`'s `claude`/`claude-code` basename check are deliberately **out of scope** — the former needs underscore/hash variants `OWN_BINARY_NAMES` doesn't carry, and the latter's own doc comment explicitly excludes `trusty-mpm` by design (native `ps aux` discovery must not self-adopt the daemon's own process as a "claude session").

## Test

Added `is_claude_command_matches_trusty_mpm` (`daemon/discovery.rs`) proving a `trusty-mpm`-launched session is now recognized by the discovery predicate.

## Gates (raw output)

- `cargo test -p trusty-mpm`: 4034 passed (lib) + 1319×2 (feature variants) + remaining integration binaries, **0 failed**
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: clean, exit 0
- `cargo fmt --check`: clean, exit 0
- `bash scripts/check_line_cap.sh`: `line-cap: 7 allowlisted, 0 violations — OK.`

Closes #4058


bobmatnyc:
## Code Critic — PR #4465 (issue #4058)

## Verdict: APPROVE

Zero CRITICAL, zero HIGH. Three MEDIUM findings below — track them; none blocks the merge.

The one-line fix is correct and the new test genuinely reproduces the defect. The MEDIUMs are all in the consolidation half, at call sites that were not broken before this PR.

### The four lists — pre-PR vs post-PR

| # | List | Pre-PR contents (`origin/main`) | Match semantics at consumer | Post-PR | Behavior changed? |
|---|---|---|---|---|---|
| 1 | `daemon::discovery::CLAUDE_COMMANDS` (`discovery.rs:32`) | `["claude","claude-code","claude-mpm","tm"]` | substring, except `tm` exact | `OTHER_CLAUDE_LAUNCHER_COMMANDS` (`["claude","claude-code","claude-mpm"]`) + `OWN_BINARY_NAMES` | **YES — intended.** `trusty-mpm` now matches (the fix). Also newly matches anything *containing* `trusty-mpm` — see finding 1. |
| 2 | `core::standalone::hooks::MPM_BIN_NAMES` (`hooks/mod.rs:214`) | `["trusty-mpm","tm"]` | `hooks/mod.rs:110` — `.find_map(resolve_binary)`, **first PATH hit wins**; `hooks/mod.rs:240` — `.contains()`, order-irrelevant | `["tm","trusty-mpm"]` | **YES — unintended.** Set identical, **order flipped**. See finding 2. |
| 3 | `core::session_launch::settings::STATUSLINE_BIN_NAMES` (`settings.rs:1084`) | `["tm","trusty-mpm"]` | `settings.rs:1142` — `.find_map(path_lookup)`, first hit wins | `["tm","trusty-mpm"]` | **No.** Identical contents and order. Correctly reasoned about in the PR's own doc comment. |
| 4 | inline check in `find_daemon_pids` (`daemon.rs:394` on main) | `name == "trusty-mpm" \|\| name == "tm"` | strict equality, short-circuit — order irrelevant | `.any(\|n\| name == *n)` over `["tm","trusty-mpm"]` | **No.** Same set, same strict equality, order immaterial. |

Case/whitespace (focus item 3): only `is_claude_command` normalizes (`trim().to_lowercase()`), and it normalizes the *input*, not the constant. The other three consumers compare raw. Since every entry in `OWN_BINARY_NAMES` is already lowercase and whitespace-free, the shared constant is safe across both a normalizing and a strict-equality consumer today. Not a finding — but it is a latent trap if a future `[[bin]]` name carries uppercase.

`is_claude_process` (focus item 4): correctly left alone. It matches solely on the executable *basename* being exactly `claude`/`claude-code` (`discovery.rs:180`) and is a separate `ps aux` code path from the tmux-pane predicate this PR changed. Nothing in `is_claude_command` feeds it, so the daemon cannot self-adopt through the changed path.

## Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/daemon/discovery.rs` | 63-65 | Substring-matching `OWN_BINARY_NAMES` opens a new false-positive window: `trusty-mpm-gui` — a real `[[bin]]` target in this same workspace (`crates/trusty-mpm-gui/Cargo.toml`) — now satisfies `is_claude_command` and would be auto-adopted as a Claude session. Pre-PR it did not match. | Match own names by equality, not substring: `OWN_BINARY_NAMES.iter().any(\|c\| lower == *c)` (which also subsumes the `lower == "tm"` special case and the `*c != "tm"` filter). `pane_current_command` is a bare process name, never a command line, so exact match fixes #4058 equally well with no substring surface. |
| MEDIUM | `crates/trusty-mpm/src/core/standalone/hooks/mod.rs` | 214 (consumed at 110-113) | `MPM_BIN_NAMES` was `["trusty-mpm","tm"]`; it is now `["tm","trusty-mpm"]`. `resolve_stable_hook_exe` consumes it as a **first-PATH-hit-wins** priority list, so the hook exe fallback now prefers `tm` where it previously preferred `trusty-mpm`. Unremarked in the diff, the doc comment, and the CHANGELOG. No test pins this order (`resolve_stable_hook_exe` calls the real `resolve_binary`, so it is not injectable), so CI cannot see the flip. | Either restore the original order at this site (`const MPM_BIN_NAMES: &[&str] = &["trusty-mpm", "tm"]` derived by reversing, or an explicit local ordering) or state in the doc comment that the flip is deliberate and why `tm`-first is preferred here. |
| MEDIUM | `crates/trusty-mpm/src/core/own_binary_names.rs` | 16-24 | Two doc claims are inaccurate, and the first is what licensed finding 2. (a) "in no particular priority order" (16-17) is wrong — **two** of the three list consumers (`hooks/mod.rs:110`, `settings.rs:1142`) are order-sensitive first-hit-wins PATH lookups, so reordering this array silently changes both. (b) The `Test:` axis (21-24) claims coverage "by every call site's own test suite" naming `bin::tm::commands::daemon`; `find_daemon_pids` has **zero** tests (`git grep find_daemon_pids` → doc mention + two call sites only). The consolidation touched three previously-unbroken call sites and added no test for any of them. | Replace (a) with an explicit warning that the array order IS load-bearing for the two PATH-resolution consumers. Drop `bin::tm::commands::daemon` from the coverage claim, or add a test asserting the constant's contents (`assert_eq!(OWN_BINARY_NAMES, &["tm","trusty-mpm"])` plus a negative `!is_claude_command("trusty-mpm-gui")` case) so any future reorder or addition trips a test rather than shipping silently. |
| LOW | `crates/trusty-mpm/src/core/standalone/hooks/mod.rs` | 209-210 | The retained leading sentence — "the full name and the short everyday alias" — enumerates in the opposite order to the array it now documents. | Reword, or fix the order per finding 2. |

## Verification performed

Independent worktree off `pr/4465` (`f9ed228e`), removed after review.

- **New test reproduces the defect** (focus item 5): reverted the predicate body to `main`'s logic, kept the new test → `is_claude_command_matches_trusty_mpm ... FAILED — expected \`trusty-mpm\` to match`. Restored → passes. Confirmed.
- **Finding 1 reproduced**: added `"trusty-mpm-gui"` to `is_claude_command_rejects_others` on unmodified PR head → `FAILED — expected \`trusty-mpm-gui\` not to match`.
- `cargo test -p trusty-mpm --lib` → `test result: ok. 4034 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out`
- `cargo clippy -p trusty-mpm --lib --bins --all-features -- -D warnings` → `Finished dev profile ... in 4m 10s`, no crate warnings
- `cargo fmt -p trusty-mpm --check` → exit 0
- `bash scripts/check_line_cap.sh` → `line-cap: 7 allowlisted, 0 violations — OK.`
- CHANGELOG (focus item 6): present under `## [Unreleased]` → `### Fixed`, `crates/trusty-mpm/CHANGELOG.md:12-21`. Satisfied.

## Notes

- **Doctrine scope.** The "Common entry point, clean domain demarcation" rule is written for capabilities "shared across two or more crates" / "a cross-crate concern". All four lists here live inside `trusty-mpm`, so the doctrine is being extended intra-crate rather than applied literally. That extension is reasonable and I am not flagging it — the four copies had demonstrably drifted, which is precisely the failure mode the rule targets. Noting it so the precedent is explicit.
- **#4230 conflict risk.** The `find_daemon_pids` change (`daemon.rs:394-398`) is semantically a no-op: same two names, same strict equality, order immaterial. It touches only the predicate that classifies a scanned process, not spawn logic, lock files, or signal delivery. Semantic conflict with daemon-spawn-guard work is unlikely; a **textual** conflict at `daemon.rs:391-400` is plausible and should be resolved by taking both changes.
- **Not flagged.** No functional regression is claimed for finding 2 — both binaries are built from the identical source and `is_mpm_binary_filename` still recognises either name, so stale `trusty-mpm hook` entries remain strippable and the #2235 unbounded-growth shape is not reintroduced. The fallback also only fires on ephemeral/debug builds. That is why it is MEDIUM, not HIGH. Similarly, finding 1 is MEDIUM rather than HIGH because it needs `trusty-mpm-gui` to be a tmux pane's foreground command — plausible under `cargo run` in a pane, not routine — and the impact is a spurious session registration, not data loss or a crash.
- Pre-existing load flakes (#4407/#4306) not re-litigated; nothing in this PR introduces flakiness.

bobmatnyc:
Addressed all three MEDIUM findings + the LOW from the round-1 critic review (https://github.com/bobmatnyc/trusty-tools/pull/4465#issuecomment-5144952014):

1. **`trusty-mpm-gui` false positive** (`discovery.rs`) — `is_claude_command`'s own-binary-names arm now matches by exact equality, not substring. Added `trusty-mpm-gui` to `is_claude_command_rejects_others`.
2. **`MPM_BIN_NAMES` order flip** (`hooks/mod.rs`) — reverted to its original literal `["trusty-mpm", "tm"]` (preserving pre-PR PATH-lookup preference at that call site) instead of aliasing `OWN_BINARY_NAMES`. Added `test_mpm_bin_names_matches_own_binary_names_set` pinning the two arrays to the same set.
3. **Doc-comment accuracy** (`own_binary_names.rs`) — corrected the "no particular priority order" claim (order is load-bearing for two consumers) and dropped the false test-coverage claim for `find_daemon_pids`. Added `own_binary_names_is_tm_then_trusty_mpm` pinning exact contents/order.
4. **LOW enumeration order** (`hooks/mod.rs:209-210`) — fixed to match the (reverted) array order.

Also, per repo-owner ruling: added a scope note to `.trusty-mpm/INSTRUCTIONS.md`'s "Common entry point, clean domain demarcation" rule — it governs cross-crate sharing only; this PR's intra-crate consolidation stands on the drift/defect evidence, not on that rule.

Gates re-run, raw output:
- `cargo test -p trusty-mpm`: 4036 passed lib (+2 new tests) + all feature-variant/integration suites, 0 failed
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: exit 0, clean
- `cargo fmt --check`: exit 0, clean
- `bash scripts/check_line_cap.sh`: `line-cap: 7 allowlisted, 0 violations — OK.`
- `bash scripts/check_sld.sh` (touched a doc file): `0 error(s), 0 warning(s)`
- `bash scripts/check_test_pointers.sh`: `0 dangling pointers — OK.`

bobmatnyc:
Post-merge note: this squash-merged as `da6ca5e1` while the round-2 gate run was still in flight, so one critic finding landed only half-closed.

Round-1 finding 2 (`MPM_BIN_NAMES` order flip) was fixed here by restoring the original `["trusty-mpm", "tm"]` literal, but the accompanying `test_mpm_bin_names_matches_own_binary_names_set` pins the two-name **SET** — it cannot see an order flip, which is exactly the failure mode that shipped unnoticed in round 1. #4487 adds `test_mpm_bin_names_prefers_full_name_over_short_alias`, pinning the array's exact order, plus the "why `trusty-mpm` first" rationale on the doc comment.

#4487 also restores two blank separator lines between the `[Unreleased]` CHANGELOG entries that this PR's `origin/main` merge dropped from the unrelated `#4163`/`#4141` entries.

Issue #4397, opened 2026-07-30
bare `tm daemon` bypasses the #2486 unsupervised-spawn guard


bobmatnyc:
**Defect:** the #2486 guard prevents an unsupervised daemon spawn only on the MCP-bridge path. The bare-CLI path walks past it.

**Evidence:** `compute_no_spawn()` (`crates/trusty-mpm/src/bin/tm/commands/launchd_probe.rs:87`) has exactly one call site — `crates/trusty-mpm/src/bin/tm/commands/serve_stdio.rs:91-114`. `run_daemon()` (`crates/trusty-mpm/src/bin/tm/commands/daemon_run.rs:31-100`) never calls it; it calls `apply_supervision_signal()` (line 63) only to populate the `/health` `supervised` flag, with no refusal path.

**Observed fail-open:** the daemon detects the condition, logs `unsupervised trusty-mpm daemon running alongside a launchd unit — ... kill it and use launchctl kickstart`, and proceeds anyway. Live instance PID 35895 has been serving port 7880 unsupervised since 2026-07-29, holding a stale 1.2.3 image while the on-disk binary is 1.3.0 (`/health` → `{"supervised":false,"version":"1.2.3"}`).

**Resolution:** have `run_daemon()` consult `compute_no_spawn` on the bare-CLI path and refuse (or require an explicit `--force`/`--unsupervised` opt-in) when a launchd unit is registered.

**Related, worth noting in one line:** `com.trusty.mpm.plist` declares `KeepAlive = { SuccessfulExit: false }`, so a clean SIGTERM exit 0 leaves the unit dormant and never relaunched — a separate contributing factor to the same outcome.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #2332, opened 2026-07-10
feat(trusty-mpm): daemon startup version banner + stale-daemon detection (installed binary newer than running daemon)


bobmatnyc:
## Incident (2026-07-10)

A tm daemon instance ran 46.8h (Jul 8 12:07 → Jul 10 10:58 EDT) with code predating the #1955/#1989 tm-session-naming migration, silently creating 9 legacy `tmpm-*`-named sessions across trusty-tools/aria/duetto-eve-agents while main + the installed CLI had long moved to `tm-<project>-<NN>`. Root-cause forensics required correlating tmux session_created timestamps against "daemon listening" banners in ~/Library/Logs/trusty-mpm/stderr.log because the daemon logs **NO version/build identifier at startup**. Same failure family as #2214 (cargo install can silently leave a stale/downgraded daemon serving for days).

## Requested

1. **Version + build-hash banner at daemon startup**: The "daemon listening" line should carry the semver + git SHA (or embedded binary version identifier).

2. **Stale-daemon detection**: Periodically (or on health-check) compare the running daemon's build against the on-disk `~/.cargo/bin` binary (via mtime and/or embedded version). Surface a WARN in logs + `tm doctor` + `/health` when the installed binary is newer than the running daemon, prompting a restart.

3. **Optional: `tm doctor` enhancements**: Show daemon uptime + version explicitly in doctor output.

## Related cleanup (optional same scope or follow-up)

The session store carries ~56 legacy `tmpm-`-prefixed records (~47 long-dead) from prior stale-daemon windows; a targeted prune of dead legacy-prefixed records would clear the backlog.

## Evidence

- `crates/trusty-common/src/session_naming.rs` (PREFIX history)
- `crates/trusty-mpm/src/daemon/session_manager/naming.rs` (server-side generation)
- `~/Library/Logs/trusty-mpm/stderr.log` (restart banners show the timeline)

## Effort

M (medium)

Issue #1900, opened 2026-07-02
tm guided_autostart checks wrong launchd plist name, causing stray detached daemon spawns


bobmatnyc:
## Root Cause

`crates/trusty-mpm/src/bin/tm/commands/guided_autostart.rs`'s `ensure_daemon_started` function checks for a launchd plist named `com.trusty.mpm.supervisor.plist` (label `com.trusty.mpm.supervisor`). However, the actually-installed/registered launchd service on real machines is `com.trusty.mpm.plist` (label `com.trusty.mpm`, confirmed via `launchctl list`).

**Affected code locations:**
- `SUPERVISOR_PLIST_LABEL` and `supervisor_plist_path()` constants/logic (lines 24-25, 34-39)
- `try_launchctl_start` fallback logic (lines 86-93)

Because the plist lookup always misses the real service, `try_launchctl_start` always returns `Unavailable`. The autostart path never uses `launchctl bootstrap` to properly manage the real daemon — instead it always falls through to a raw detached spawn (`Command::new(current_exe).arg("daemon").spawn()`).

## Observed Impact

When a bare `tm` invocation races a daemon-unreachable health check while the real daemon IS actually running (just not detected due to the plist name mismatch), it spawns a second, fully independent `tm daemon` process that:

1. Finds the real daemon already holding port 7880
2. Logs a warning and binds a random ephemeral port instead
3. Becomes orphaned (reparented to PID 1) once the parent `tm` CLI exits
4. Keeps running indefinitely, untracked and consuming resources

Confirmed harmless in observed case — no clients ever connected to the stray daemon.

## Reproduction

On a machine with the real `com.trusty.mpm.plist` launchd service installed and running, invoke bare `tm` in a way that hits the guided-autostart health-check path while there's any timing window where the health check might report the daemon unreachable (e.g., right at daemon startup, or under load) — a stray daemon on a random port gets spawned and orphaned.

## Fix Direction

Two possible approaches — determine which is correct before fixing:

1. **Correct the plist name/label constants** in `guided_autostart.rs` to match the actually-installed `com.trusty.mpm.plist` / `com.trusty.mpm` label

[...]

PR #4466, opened 2026-07-31
fix(trusty-mpm): stop `tm start`/`tm restart` spawning an unsupervised orphan daemon (#4230)

### LINKED ISSUE #4230: an unsupervised `tm daemon` orphan can seize port 7880 while launchd reports the job not running

bobmatnyc:
## The mechanism

`commands::daemon::start` (`tm start`, and `tm restart` via it) was the ONE
client-side daemon-spawn path in the crate with no launchd awareness at all. It
unconditionally spawned a detached bare `tm daemon` with `current_dir` anchored
to `$HOME` — which reproduces the #4230 evidence exactly: argv `tm daemon`,
`cwd=$HOME`, and `PPID 1` once the short-lived `tm start` parent exits.

There are only three code paths in the crate that spawn a `tm daemon`
subprocess. Two were already guarded:

| Path | Guard |
|---|---|
| `serve_stdio.rs` (MCP bridge) | `no_spawn` = `compute_no_spawn(plist_exists)`, since #2486 |
| `guided_autostart.rs` (bare `tm`) | nudges `launchctl bootstrap` when `com.trusty.mpm.plist` exists, since #1900 |
| **`commands/daemon.rs::start`** | **none** |

`launchd_probe.rs` (the #2486 fix) exists at `trusty-mpm-v1.0.2` — the tag is
2026-07-25, the day before the orphan started — so the guard was compiled into
the misbehaving binary. It was simply never consulted from this path.

## Explanation (a) — "the #2486 bridge guard regressed" — is ruled out

The orphan's own `/health` reported `{"supervised":false}`. `supervised` is
`compute_supervised(is_launchd_supervised, plist_exists) = !plist_exists ||
is_launchd_supervised`, so `false` requires `plist_exists == true`. And
`compute_no_spawn(plist_exists)` is exactly `plist_exists` — so the bridge's
`no_spawn` was necessarily set at that moment and it could not have spawned
anything. The orphan's own health payload is the evidence that acquits the
bridge.

The same payload also rules out launchd as the spawner (`is_launchd_supervised`
was `false` at the child's startup), and rules out a long-lived parent, since
`PPID 1` at observation time means the parent had exited. Short-lived non-launchd
parent, bare `tm daemon` argv, `cwd=$HOME` — `tm start` is the only remaining
in-crate producer.

Note that `run_daemon`'s callee-side guard (#4397, merged 2026-07-30) postdates
this incident, so on 07-26 nothing stopped the child either.

## Why the callee-side #4397 guard is not sufficient on its own

`is_launchd_supervised()`'s first prong returns `true` when `XPC_SERVICE_NAME`
is set and `TERM_PROGRAM` is not. A `tm daemon` child spawned from a context
without `TERM_PROGRAM` can therefore self-report as launchd-supervised, pass the
#4397 guard, and publish `supervised: true` — a worse variant of #4230, since
the flag operators are told to trust would then agree with the orphan. This
caller-side guard is keyed on plist presence alone, with no heuristic.

## Changes

- `commands::daemon::start` and `::restart` refuse to spawn when a trusty-mpm
  launchd unit is registered, naming the `launchctl kickstart` recipe and the
  `tm daemon --force` opt-in. The check runs BEFORE `restart`'s `pkill`, so a
  refusal never tears the supervised daemon down and then declines to bring it
  back. (`pkill` is itself a route into this state: SIGTERM makes the launchd job
  exit 0, and `KeepAlive {SuccessfulExit: false}` means launchd deliberately does
  not respawn it — which is why the incident's job had been down since 07-22.)
- New `daemon_orphan` `tm doctor` check. It compares WHO ANSWERED `/health`
  against WHO LAUNCHD RUNS — two independent sources, so neither the daemon nor
  launchd alone can make it pass. Client-side like the #2332 `stale_daemon` check,
  and for a stronger reason: in the orphan state the process answering
  `run_doctor` is precisely the one whose self-report cannot be trusted. Cost is
  one `/health` GET (shared with the staleness check), one `read_dir`, and one
  `launchctl` invocation — no spawn, no port scan, no network.
- `/health` gains `pid` (`std::process::id()` at handler time) and
  `unsupervised_forced`. `pid` is the authoritative half: `supervised` is a
  heuristic whose `getppid() == 1` prong reads `true` for a real orphan, so it is
  a FALLBACK only, used when the responding daemon is too old to report `pid` —
  and `supervised: true` yields `Unknown` there, never a pass. On the client
  `supervised` and `pid` are `Option`, so "this daemon cannot tell me" stays
  distinct from "this daemon is fine".
- `launchd_ownership` answers in THREE states — `Running(pid)`, `NotRunning`,
  `Unavailable`. `NotRunning` requires a POSITIVE exit-0 answer from launchd and is
  the #4230 state; `Unavailable` means launchd could not be asked (`launchctl`
  missing from `PATH`, sandboxed, or the unit invisible in the caller's bootstrap
  domain) and yields `Unknown` with NO kill instruction. A `list` that exits 113 is
  escalated to an explicit `gui/<uid>` query before giving up. Collapsing these two
  into one `None` was the round-2 review's HIGH: it made `tm doctor` prescribe
  `kill -TERM` against a correctly supervised daemon.
- The daemon launchd label is RESOLVED per host rather than hardcoded, and a
  drifted label must have `ProgramArguments` that actually run `daemon` — so
  neither the supervisor plist nor a `logrotate` sibling can be mistaken for the
  daemon unit. One resolver, `daemon_restart_command()`, keeps every remediation
  naming a unit that exists on THIS host.
- `docs/reference/release-workflow.md` gains an orphan pre-check before `bootout`
  (a silent no-op against a process launchd does not own) and **drops its
  `PPID == 1` verification step**: any daemon whose parent exits reparents to
  PID 1, so that documented check CONFIRMED the #4230 orphan as supervised.

## Restart-verb sites

`daemon_restart_command()` replaces the hardcoded `tm restart` in every code path
that emits operator guidance (`version_staleness.rs` ×3, `manager.rs` ×3,
`session_picker.rs`, plus the doctor messages and the deployed skill).

One site is deliberately NOT routed through it: `assets/default-services.yaml`'s
`restart_cmd: "tm restart"`, executed by `services.rs:166`. On a host with no
daemon launchd unit that verb is correct and succeeds. On a launchd host
`tm restart` refuses BEFORE tearing anything down and prints the exact `kickstart`
recipe, which `tm services restart` surfaces verbatim before exiting 1 — strictly
more actionable than the `no restart_cmd defined in manifest` that `null` would
print, and `null` would break the non-launchd host too. The reasoning is recorded
in a comment beside the value.

## Tests

46 tests. `doctor_orphan_tests.rs` pins every branch of the verdict table —
including that `Unavailable` never reaches `Fail` and never emits a `kill`, that a
launchd/daemon contradiction yields `Unknown`, and that the real #4230 payload
(`supervised: false`) still yields a hard `Fail` with the full remediation.
`authoritative_verdict_ignores_a_lying_self_report` asserts both directions.
`launchd_probe`'s tests cover label resolution against temp homes (including a
decoy `logrotate` sibling), every real `launchctl` output shape, and the exit-0
→ `Running`/`NotRunning` mapping.

## Gates

```
cargo test -p trusty-mpm      → 4045 passed; 0 failed; 6 ignored (lib) + 1362 passed; 0 failed (bin), exit 0
cargo fmt --all --check       → clean
bash scripts/check_line_cap.sh    → line-cap: 7 allowlisted, 0 violations — OK.
bash scripts/check_test_pointers.sh → test-pointers: 0 dangling pointers — OK.
```

The three-state fix was verified by reproducing the false `Fail` first, then
re-running the same scenarios after the fix — see the round-3 response comment for
the raw output.

Closes #4230


bobmatnyc:

Reviewed at head `5258195b` against `origin/main`, in a read-only worktree. Gates re-run independently: `cargo test -p trusty-mpm --bin tm` → `test result: ok. 1327 passed; 0 failed; 1 ignored`; `cargo test -p trusty-mpm --lib health_snapshot` → `ok. 2 passed; 0 failed`; `bash scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`. CHANGELOG entry present under `## [Unreleased]`. No CRITICAL findings, so this is not a BLOCK — but three HIGHs, and two of them are cheap enough that they belong in this PR rather than a follow-up.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/update/upgrade.rs` + `crates/trusty-mpm/src/client/http_client/types.rs` | 432-437 / 558-570 | `daemon_orphan`'s only input is a self-report that reads `true` for a real orphan on two paths, and there is no third state | Make `supervised` an `Option<bool>` so absence → `Unknown`; cross-check with the listening PID vs `launchctl print`'s PID (what the new doc step tells humans to do) |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/launchd_probe.rs` | 41-44, 119-128 | The refusal fires on `com.trusty.mpm.supervisor.plist`, which does not own the daemon, and then prescribes a launchd label that does not exist on that host | Key the refusal on the daemon-owning label only, or resolve the installed label at message time (`scripts/install-trusty-mpm-signed.sh:299-321` already has that logic) |
| HIGH | `crates/trusty-mpm/src/core/version_staleness.rs` (+5 more) | 92, 115, 141 | `tm restart` is now a hard error on launchd hosts, but seven in-repo strings and the bundled skill still prescribe it — including `tm doctor`'s own remediation printed two lines above the new check | Update the strings and the skill to name `launchctl kickstart` when a plist is registered |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/daemon_run.rs` | 77-84 | `tm daemon --force` — the opt-in both new messages recommend — guarantees `daemon_orphan: Fail` calling it an ORPHAN and telling the operator to kill it | Record the `--force` opt-in on `/health` and downgrade to `Warn`, or at minimum name `--force` in the failure text |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/daemon.rs` | 331 | `tm stop` left unguarded — its SIGTERM is the documented precursor state, and it is now the only unblocked route into it | Print the same hint from `stop_daemon` when a plist is registered |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/doctor_orphan.rs` | 163-165 | Module doc claims it reuses "one `/health` GET the caller already makes"; it issues a second one | Pass the snapshot in from `misc.rs`, or correct the doc |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/daemon.rs` | 282 | Rewritten `Test:` pointer on `restart` names `cli_parses_start`; no restart-parsing test exists under that name (restart parsing is asserted inside the abbreviation test at `tests.rs:1125`) | Name the real test or add `cli_parses_restart` |
| LOW | `.trusty-mpm/INSTRUCTIONS.md` | 518 | Still shows the `ps -o ppid=` "must print 1" step, already caveated as weaker, now inconsistent with the stronger release-workflow wording | Align with the release-workflow replacement |

### On the suspected fail-open — it is real, and the field default is the smaller half

Traced path: `misc.rs:170` → `doctor_orphan.rs:187` `orphan_daemon_check` → `verdict(plist_exists, snapshot.supervised, &snapshot.version)` at `:190` → `verdict` at `:217` reads **only** those two booleans. Nothing else in the check compensates: no version comparison, no PID comparison, no `launchctl` probe. `stale_daemon` is the one adjacent signal and it is a `Warn` keyed on version, so it misses a same-version orphan entirely and, when it does fire, its remediation is `tm restart` — which this PR makes refuse.

Two ways a genuine orphan reports `supervised: true`:

1. **Field absent** → `default_supervised()` → `true` (`types.rs:558-570`). Narrow in practice: the field has been on the wire since #2486, which predates the incident's own 1.0.2 image, and any daemon that old also trips `stale_daemon` on version. This half alone would be MEDIUM.
2. **`is_launchd_supervised()`'s second prong is `getppid() == 1`** (`crates/trusty-common/src/update/upgrade.rs:432-437). `supervised` is computed once at daemon startup (`daemon_run.rs:221-225`). Any orphan whose spawning parent has already exited by the time that probe runs — nohup'd, backgrounded, a parent that died during boot — self-reports as launchd-supervised, and `daemon_orphan` returns `Ok`. This is the half that matters, and the PR does not mention it anywhere: every doc comment cites only the `XPC_SERVICE_NAME` prong, and the disclosed residual gap covers only that prong too.

The sharp edge is that this PR removes `PPID == 1` from the operator checklist on the grounds that reparenting to init does not prove supervision — and simultaneously makes `.supervised`, computed from a heuristic whose fallback prong *is* `PPID == 1`, the sole signal for both the new check and the docs' replacement guidance. The argument that condemns the doc step applies verbatim to the code the replacement leans on. #4230's own premise was that the spawn path was unknown; a detector that can only see orphans whose parent outlives their boot is weaker than the PR presents it as.

Worth stating plainly: the incident itself would have been caught, because `tm start` stays alive polling `/health` for 5 s, so `getppid() != 1` at the child's probe. That is also independent corroboration of the root-cause attribution — see below.

### `restart` ordering — correct

`daemon.rs:287-291` precedes `daemon_healthy` at `:292` and the two `pkill` calls at `:297-304`. The guard cannot tear down a supervised daemon and then decline to bring it back. This is not merely neutral: on a host where the spawn would have been refused by the #4397 child guard anyway, the pre-PR `tm restart` killed a running daemon and left it unstartable, and the new ordering prevents that. Real value.

### The deduction ruling out explanation (a) — sound, with two caveats

Both identities verified verbatim at the `trusty-mpm-v1.0.2` tag, i.e. in the binary that produced the observed payload: `compute_supervised(is, plist) = !plist_exists || is_launchd_supervised` (`launchd_probe.rs:72-74`) and `compute_no_spawn(plist) = plist_exists` (`:87-89`), with `serve_stdio.rs:92-94` wiring `no_spawn` to the same `mpm_launchd_plist_exists()` probe. So `supervised: false` ⟹ `plist_exists` ⟹ the bridge's `no_spawn` was armed. The attribution holds.

Caveats, neither fatal:
- The two booleans are sampled at different instants — `no_spawn` at bridge startup, `supervised` at daemon startup. The argument strictly needs the plist to have existed at both. The 07-22 clean job exit independently establishes that it did, four days before the orphan started, so the gap closes on evidence rather than on the identity.
- `cwd=$HOME` is consistent with `daemon.rs:229`'s explicit `current_dir(home)` but is not exclusive to it: `guided_autostart.rs:191-201` spawns with the inherited cwd, which is `$HOME` for a bare `tm` run from the home directory. What does discriminate is the PPID prong — `supervised: false` requires `getppid() != 1` at the child's boot, i.e. a live short-lived parent, which fits `tm start`'s 5-second health poll.

### The supervisor-plist conflation (HIGH-2, detail)

`launchd_plist_paths()` (`launchd_probe.rs:41-44`) counts `com.trusty.mpm.supervisor.plist` as "launchd owns the daemon". It does not. That plist's `ProgramArguments` is `[<tm>, "supervisor"]` (`crates/trusty-installer/src/commands/plist_bootstrap.rs:236-240`), and `tm supervisor` never starts a daemon — the only two `tm daemon` spawn sites in the crate are `daemon.rs:228` and `guided_autostart.rs:197`, neither reachable from the supervisor. `scripts/install-trusty-mpm-signed.sh:300-306` already documents the supervisor plist as "a distinct service" and excludes it from label resolution.

Consequence on a `tctl install` host with no hand-authored `com.trusty.mpm.plist` (nothing in-tree provisions one):

- `tm start` / `tm restart` refuse with "a trusty-mpm launchd unit is registered, so launchd owns the daemon's lifecycle" — false in that configuration.
- The prescribed `launchctl kickstart -k gui/$(id -u)/com.trusty.mpm` fails with "Could not find service"; the parenthetical `com.trusty.mpm.supervisor` starts `tm supervisor`, not the daemon.
- Bare `tm` still takes the detached-spawn branch, because `main_daemon_managed_by_launchd_in` (`guided_autostart.rs:66-68`) checks only the main label — so the PR body's claim that this sibling path is "already guarded since #1900" holds only for the main plist, not for the label the installer actually ships.

To be precise about attribution: the unstartable state pre-exists this PR — #4397's child guard already refuses on that host and the operator got "daemon did not become healthy within 5s". What this PR adds is an explicit assertion that is wrong there, plus remediation that cannot work. That is why it is HIGH and not CRITICAL.

### The `tm restart` guidance drift (HIGH-3, detail)

`tm restart` now bails on any host with either plist. Still prescribing it: `core/version_staleness.rs:92`, `:115`, `:141` (printed by `tm doctor` itself, two lines above the new `daemon_orphan` line — one command's output now contradicts itself); `commands/manager.rs:143`, `:187`, `:223` (the documented `cargo install trusty-mpm --locked` upgrade path); `commands/session_picker.rs:854`; and `src/assets/skills/tm-cli-operations.md:373`, which is a deployed skill — agents will run the verb and get an error. Also `daemon.rs:164`'s own doc still says `tm start` "can sit in shell profiles and setup scripts", which now emits an error on every shell whose launchd job is down.

### Checked and not flagged

- **False positive during `bootout → install → bootstrap`.** No. With the plist present and nothing listening, `health_snapshot()` errors and `orphan_daemon_check` returns `Unknown` (`doctor_orphan.rs:191-196`), which `#4005`'s overall-verdict wording already declines to call a pass. The only spurious `Fail` is the `--force` case above.
- **PPID removal from the docs.** The replacement — compare the listening PID against `launchctl print`'s PID — is strictly stronger than `PPID == 1`, and no other doc instructs relying on PPID uncaveated (`.trusty-mpm/INSTRUCTIONS.md:508-518` already labels it the weak variant; `docs/trusty-installer/research/.../09-upgrade-flow.md:762` describes the function, not an operator step). The doc change is fine; the code the docs now point at is what HIGH-1 is about.
- **Scope creep.** Cohesive, not flagging. The client type change is a precondition for the check, and the docs edit closes resolution item 3 of #4230 verbatim.
- **#4465 overlap.** No semantic conflict. That PR touches `find_daemon_pids` at `daemon.rs:391`; this one touches `start` (179-205) and `restart` (286-291). Different regions, orthogonal concerns.
- **`anyhow::bail!` with a runtime `String`** — compiles and carries no format-string hazard (no braces in the hint). Verified by the bin build.

### Required changes before this lands

1. Give `daemon_orphan` a third state — `Option<bool>` on `supervised` so absence yields `Unknown` — and add a signal that is not the daemon's own self-report. The PID comparison the new doc step prescribes for humans is the obvious candidate; without it the check inherits the `getppid() == 1` weakness this PR argues against in the docs.
2. Stop treating `com.trusty.mpm.supervisor.plist` as evidence that launchd owns the daemon, or resolve the actually-installed label before printing a `kickstart` recipe.
3. Update the seven `tm restart` remediation strings and the bundled `tm-cli-operations` skill, starting with `version_staleness.rs` since `tm doctor` prints it alongside the new check.
4. Name the `--force` case in the `daemon_orphan` failure text, or thread the opt-in through `/health` and downgrade to `Warn`.

bobmatnyc:
## Round 2 — all eight findings addressed at `47de3990`

### HIGH-1 — fail-open closed with an authoritative signal

You were right, and the `getppid() == 1` prong is the half that mattered. The
contradiction you named was real: the check leaned on `.supervised` while the docs
change deleted `PPID == 1` on the grounds that reparenting proves nothing — and
`.supervised`'s fallback prong *is* that test.

`/health` now carries `pid`. `daemon_orphan` compares it against the PID
`launchctl list <label>` reports — two independent sources, so neither the daemon
nor launchd alone can make the check pass. Three tiers, strongest first:

| Tier | Inputs | Verdict |
|---|---|---|
| 1 | no daemon unit registered | `Ok` |
| 2 | `pid` known, launchd PID differs or launchd runs nothing | `Fail` — authoritative |
| 2 | `pid` known, equals launchd's PID | `Ok` — authoritative |
| 3 | `pid` absent, `supervised: false` | `Fail` |
| 3 | `pid` absent, `supervised: true` or absent | **`Unknown`, never `Ok`** |

`authoritative_verdict_ignores_a_lying_self_report` pins both directions: an
orphan claiming `supervised: true` still fails, and a launchd-owned daemon whose
heuristic said `false` still passes. Tier 3's `Some(true)` message states why the
flag is not conclusive, naming `getppid`.

Field default corrected too: client-side `supervised` is `Option<bool>` with no
default, so absence is a third state rather than a silent pass —
`health_snapshot_supervised_is_none_when_absent`.

Doc comments now name both prongs. `#4469` remains the right home for replacing
the heuristic itself; this check no longer depends on the outcome.

The module no longer claims "no `launchctl` shell-out" — it makes one, and that is
the point.

### HIGH-2 — supervisor plist no longer counts

Verified independently: `PLIST_TEMPLATE`'s `ProgramArguments` is
`[__TM_BINARY_PATH__, "supervisor"]`, and `supervisor/mod.rs` documents a PASSIVE
observer with no `Command::new` anywhere in its production code. It never owns a
daemon.

`launchd_plist_paths` is gone. `daemon_launchd_label_in(home)` resolves the LABEL
— canonical `com.trusty.mpm.plist` first, then a drifted `com.trusty.*mpm*`
excluding `supervisor`/`gui`, mirroring `install-trusty-mpm-signed.sh`'s
`resolve_mpm_plist` — and every message interpolates the resolved label, so the
`kickstart` recipe always names a unit that exists.

I narrowed the SHARED probe rather than adding a second definition beside it.
Keeping two answers to "does launchd own the daemon" in one module would have let
the bridge and the doctor check disagree on the same host. The widening is
deliberate and strictly corrective: it also removes a latent false refusal in
#4397's child guard and a latent false `no_spawn` in #2486's bridge guard on a
supervisor-only host, where auto-spawn is in fact the correct behavior.
`daemon_label_ignores_a_supervisor_only_home` asserts the whole guard chain stands
down there.

### HIGH-3 — restart-verb drift

> **Correction (round 3):** the sentence below originally read "Fixed all hits,
> not a subset." That was wrong. The round-2 review found an EIGHTH site this list
> missed — `assets/default-services.yaml:42`'s `restart_cmd: "tm restart"`,
> executed by `services.rs:166`. It is left as-is deliberately; see the round-3
> response for why. The list below is the set of CODE sites that route through the
> resolver, not the complete set of `tm restart` occurrences.

One resolver, `daemon_restart_command()`, replaces every hardcoded verb in the
code paths that emit operator guidance:

- `version_staleness.rs` × 3 — now takes `restart_hint` as a parameter, keeping
  `core` free of filesystem probing. `staleness_uses_the_caller_restart_hint`
  asserts all three branches use the caller's value AND that none leaks the
  literal, so the test cannot pass on a leftover string.
- `manager.rs` × 3, `session_picker.rs` × 1.
- `assets/skills/tm-cli-operations.md` — the deployed skill now carries a red note
  that `tm start`/`tm restart` refuse under launchd, with the `kickstart` and
  `tm doctor` commands.
- `daemon.rs:164`'s "can sit in shell profiles" doc now scopes that to hosts where
  launchd does not own the daemon.

The remaining in-repo `tm restart` mentions are mechanism descriptions in doc
comments ("`tm restart` stops the old daemon with `pkill`"), which stay accurate.

### MEDIUM and LOW

- **MEDIUM-1** — `/health` carries `unsupervised_forced`, set from `run_daemon`'s
  `force`. A deliberate run is `Warn`, not `Fail`, and the `Fail` text also names
  `--force` as the way to declare an intentional daemon.
  `warns_not_fails_when_the_operator_forced_it` covers all three orphan paths.
- **MEDIUM-2** — `tm stop` is not refused (stopping is legitimate) but now warns
  that `KeepAlive.SuccessfulExit=false` means launchd will not respawn it, naming
  the command that will.
- **LOW-1** — `misc::doctor` fetches `/health` once and passes it to both
  client-side checks. `stale_daemon_check` became sync and takes the snapshot; both
  checks now always describe the same daemon.
- **LOW-2** — pointer names `cli_infers_abbreviated_subcommands`, the test that
  actually asserts `restart` parsing.
- **LOW-3** — `.trusty-mpm/INSTRUCTIONS.md` drops the `ps -o ppid=` step and gains
  the PID-comparison form plus the `.supervised`-is-not-conclusive caveat.

### Untouched, per your "keep" list

`restart`'s guard-before-`pkill` ordering, the ruling-out of explanation (a), and
the `Unknown`-on-transport-failure behavior are unchanged.
`unknown_when_no_health_snapshot_is_available` now pins that last one explicitly.

### Gates

```
cargo test -p trusty-mpm      → 44 suites, all ok, 0 failed; lib 4038 passed / 6 ignored, bin 1350 passed / 1 ignored; exit 0
cargo fmt --check             → clean
bash scripts/check_line_cap.sh → line-cap: 7 allowlisted, 0 violations — OK.
bash scripts/check_test_pointers.sh → test-pointers: 0 dangling pointers — OK.
```

37 new tests across `launchd_probe`, `doctor_orphan`, the client wire shape, the
`/health` wire shape, and `version_staleness`.

bobmatnyc:
## CI green at `4590819e` — and why it took three pushes to get a single run

**23 SUCCESS, 1 SKIPPED (`Notify on red main`), 0 failures.** `mergeable=MERGEABLE`.

### The CI blocker was not a GitHub hiccup

My first three pushes after the review fix produced **zero** workflow runs while
other branches triggered within seconds. Cause:

```
mergeable=CONFLICTING  mergeStateStatus=DIRTY
$ git ls-remote origin 'refs/pull/4466/*'
d58ca8ef…  refs/pull/4466/head      # merge ref ABSENT
```

`pull_request`-triggered workflows run against `refs/pull/<n>/merge`. While the PR
conflicted, GitHub could not compute that ref, so it created no run at all — no
error, no queued job, just silence. Merging `origin/main` produced the merge ref
and CI fired immediately. Worth knowing: a conflicted PR here looks identical to
"CI hasn't started yet".

The conflict was #4163's `CHANGELOG` entry, then #4058's on the second round. Both
resolved by keeping both bullets — sibling additions under `### Fixed`. `daemon.rs`
auto-merged cleanly both times: #4058 consolidated the `tm stop` daemon-PID scan
onto `core::own_binary_names::OWN_BINARY_NAMES`, which does not touch the
`start`/`restart`/`stop_daemon` regions this PR changes. I verified the #4230 hint
in `stop_daemon` survived the merge intact.

### Local gates on the merged tree

```
cargo fmt --check                                       → clean
bash scripts/check_line_cap.sh   → line-cap: 7 allowlisted, 0 violations — OK.
bash scripts/check_test_pointers.sh → test-pointers: 0 dangling pointers — OK.
cargo test -p trusty-mpm         → lib: 4040 passed / 0 failed / 6 ignored
                                    bin: 1349 passed / 1 failed  ← see below
```

### One local test failure, not from this change — reporting rather than re-running past it

`commands::statusline::branch::tests::project_segment_basename_fallback_non_git`
failed once locally. It is **not** in this PR's diff
(`git diff --name-only origin/main...HEAD` matches nothing under `statusline`), the
merge brought no `statusline` changes, and it passed twice in my pre-merge run on
this same machine.

It is non-deterministic by construction. `tmux_session_name()` shells out to
`tmux display-message` behind a **100 ms wall-clock timeout**
(`branch.rs:166-188`). The test calls it a second time in its own assertion and
requires the two probes to agree — under load one of them times out and returns
`None` while the other returned the session name. Three isolated back-to-back runs:

```
run 1: ok
run 2: FAILED   left: ".tmpH5Psbb ⎇ tm-dogfood-relaunch-01"  right: ".tmpH5Psbb"
run 3: ok
```

It only manifests where `TMUX` is set — `tmux_session_name()` short-circuits on
`std::env::var_os("TMUX")?` before any shell-out, so on CI runners the test takes
the `None` branch deterministically. That is why `Test` is green here and has always
been. The post-merge local run was heavily loaded (lib suite 319 s vs 49 s
pre-merge, several sibling worktrees compiling), which is what tipped it.

Flagging rather than fixing: it is a different module from this PR's scope, and
re-running until green would bury it. Worth a ticket on its own if the flake is
seen in CI on a self-hosted runner inside tmux.

bobmatnyc:
## Code Critic — Round 2 — Verdict: WARN

Reviewed at head `4590819e` in an isolated worktree. **All three round-1 HIGHs are
genuinely closed** — I verified each by execution, not by reading the response. One
NEW HIGH found: the two-source comparison does not fail open, but it fails CLOSED on
"could not ask launchd", and the false verdict prescribes `kill -TERM` against a
healthy supervised daemon.

Gates re-run independently: `cargo test -p trusty-mpm` → 44 suites, **7018 passed; 0
failed; 16 ignored**, exit 0 (no `statusline` flake this run). `bash
scripts/check_line_cap.sh` → `7 allowlisted, 0 violations — OK`. CHANGELOG entry
present under `## [Unreleased]` / `### Fixed`. CI at this head: 23 pass, 1 skipping
(`Notify on red main`), 0 fail — including `Test: doc-comment pointer lint`. Merge
against current `main` (`79c9670c`) is clean; `git merge-tree` prints `merged` with no
conflict, so the `mergeable=UNKNOWN` is GitHub recomputing, not a re-conflict.

### Findings

|---|---|---|---|---|
| HIGH | `launchd_probe.rs` + `doctor_orphan.rs` | 242-251 / 155-163 | `launchd_owned_pid` returns `None` both for "launchd runs nothing" and "I could not ask launchd", and the `None` arm is a hard `Fail` telling the operator to `kill -TERM` the listener. A genuinely launchd-owned daemon is condemned as an ORPHAN whenever `launchctl` cannot answer | Three-state return (`Running(pid)` / `NotRunning` / `Unavailable`); query `gui/$(id -u)/<label>` explicitly; and require `supervised != Some(true)` to escalate the "launchd runs NO process" branch from `Warn` to `Fail` |
| MEDIUM | `assets/default-services.yaml` | 42 | Eighth restart site the round-1 list missed: `restart_cmd: "tm restart"`, executed by `services.rs:166`. No orphan risk (verified), but `tm services restart trusty-mpm-daemon` is now guaranteed to fail on every launchd host | `restart_cmd: null` — `services.rs:167-169` already prints "no restart_cmd defined in manifest" — or point it at `tm doctor` |
| MEDIUM | `launchd_probe.rs` | 74-88 | The drift glob accepts ANY `com.trusty.*mpm*.plist` except `*supervisor*`/`*gui*` as THE daemon label. A sibling unit like `com.trusty.mpm.logrotate.plist` becomes the daemon label, and every consequence follows from it | Read the candidate's `ProgramArguments` and require the second element to be `daemon`; `plist_bootstrap::extract_program_path` already parses that file shape |
| LOW | `daemon/api/types.rs` | 99 | `pub pid: u32` has no `#[serde(default)]` while every neighbouring added field does. `HealthResponse` derives `Deserialize` and is a public `ToSchema` type; deserializing an older daemon's `/health` is a hard error, not `None` | `#[serde(default)]`, matching `unsupervised_forced` two fields below |
| LOW | `launchd_probe.rs` | 258-260 | Doc says a parser returning `None` on an unrecognised shape "would fail OPEN". It now fails CLOSED — `None` produces `Fail` | Reword to "fail closed" |
| LOW | `docs/reference/release-workflow.md` | 415-419 | Claims `daemon_orphan` reports "`Unknown` rather than a pass when it cannot make that comparison". True only for the too-old-daemon case; when `launchctl` cannot answer it reports `Fail` | Fix with the HIGH above, or scope the sentence |
| LOW | `assets/skills/tm-cli-operations.md` | 377-383 | The deployed skill hardcodes `com.trusty.mpm` in the `kickstart` recipe — the exact hardcoding HIGH-2 removed from the code | Tell the reader to take the label from `tm doctor`'s message |

### The new HIGH, demonstrated by execution

Identical host, identical daemon, identical plist. Only `launchctl`'s ability to
answer changed:

```
# A — a shim launchctl that reports the serving pid
✅ daemon_orphan launchd unit `com.trusty.mpm` owns the daemon answering /health (pid 99019, version 1.3.0)

# B — same host, launchctl exits 113 ("Could not find service … in domain for port")
❌ daemon_orphan ORPHAN daemon serving: launchd runs NO process for `com.trusty.mpm` … Fix: `lsof …` to
   confirm the PID, `kill -TERM <pid>`, then `launchctl kickstart -k gui/$(id -u)/com.trusty.mpm`.

# C — same host, launchctl not on PATH at all
❌ daemon_orphan ORPHAN daemon serving: launchd runs NO process for `com.trusty.mpm` …
```

Two triggers, both real:

1. **`launchctl` unavailable** (case C above) — a scrubbed `PATH`, a sandbox, a
   non-macOS host that nonetheless has the plist file. `Command::new("launchctl")`
   resolves through `PATH`, and `.output().ok()?` swallows the spawn error.
2. **A non-GUI bootstrap domain.** Verified on this machine that the unit exists in
   exactly one domain:

   ```
   $ launchctl print gui/502/com.trusty.mpm    → state = running
   $ launchctl print user/502/com.trusty.mpm   → Could not find service "com.trusty.mpm" in domain for uid: 502
   ```

   `launchctl list <label>` targets the caller's bootstrap port. It resolves to
   `gui/<uid>` from a GUI-descended shell (which is why my `tm doctor` run worked) and
   to the per-user domain over SSH / from a cron or system launchd job — where the
   `LimitLoadToSessionType = Aqua` agent is not visible and the lookup exits 113, i.e.
   case B.

Why this is HIGH and not a nit: the sentence the operator reads is "ORPHAN daemon
serving … `kill -TERM <pid>`", and `KeepAlive {SuccessfulExit: false}` means the job
does not come back on its own afterwards. `--force` is not an escape either — it would
mark a genuinely supervised daemon as a deliberate unsupervised one. The check has no
correct answer available in this state.

The escalation guard I'd prescribe also preserves HIGH-1's closure: the real #4230
payload reported `supervised: false`, so gating the hard `Fail` on `supervised !=
Some(true)` still catches the incident verbatim while never reaching `Ok` on a
self-report alone.

`verdict()` cannot currently express this state — it takes `launchd_pid:
Option<u32>` — which is why none of the 37 new tests covers it.

### Round-1 findings — closed

**HIGH-1 — CLOSED, and closed at the mechanism, not relocated.** I built the PR's `tm`
and ran the real thing.

- **A lying self-report no longer rescues an orphan.** Live: a PR-built daemon started
  with the plist hidden reported `supervised: true` (because `plist_exists` was false at
  its startup probe) and was then measured against a host with the plist present. It was
  still caught: `❌ ORPHAN daemon serving: launchd runs pid 73599 for com.trusty.mpm, but
  pid 9124 is the one answering /health`. That is exactly the population the round-1
  finding said the old check was blind to.
- **No path reaches `Ok` on self-report alone.** Tier 1 `Ok` is a filesystem fact, not a
  self-report. Tier 2 `Ok` requires `launchd_pid == Some(serving)`. Tier 3 is `Unknown`
  for `Some(true)` and for `None`. Confirmed live against this host's real supervised
  v1.3.0 daemon (which predates `pid`): `❔ daemon_orphan … too old to report its pid …
  Its own supervised: true is NOT conclusive`. No false `Fail` there either.
- **Stale / reused PID:** not reachable. `pid` is `std::process::id()` evaluated in the
  responding process at handler time (`api.rs:452`), so it cannot be stale, and launchd
  only reports a live child PID. A coincidental match between two live distinct
  processes is impossible.
- **`launchctl` returns a label with no PID:** `parse_launchctl_pid` → `None` → `Fail`.
  Correct for the #4230 state itself; the conflation with "could not ask" is the HIGH
  above.
- **Wire format, both directions.** Client `supervised: Option<bool>` with no default —
  verified live that an older daemon yields `None`, not a defaulted `true`.
  `HealthSnapshot` has no `deny_unknown_fields`, so an older client tolerates the two new
  fields. No `#[serde(default)]` anywhere reintroduces the true-default.

**HIGH-2 — CLOSED, and the widening claim holds.** Verified independently, not from the
response: `plist_bootstrap::PLIST_TEMPLATE`'s `ProgramArguments` is `[__TM_BINARY_PATH__,
"supervisor"]`, this host's installed supervisor plist matches, and `grep -rn
"Command::new" crates/trusty-mpm/src/supervisor/` returns nothing — the supervisor cannot
own a daemon. Executed the enumeration you asked for:

| Host | Behaviour |
|---|---|
| `com.trusty.mpm.plist` only | guard fires (unchanged) |
| supervisor only | guard stands down. Live: `tm daemon` **starts** (on `main` it refuses without `--force`, since `compute_supervised(false, true) = false`), and `daemon_orphan` → `✅ no trusty-mpm daemon launchd unit is registered`. Strictly corrective |
| both | canonical short-circuit wins → unchanged |
| neither | unchanged |
| drifted label | now fires where it previously did not — the intended fix, but see MEDIUM-B |

`daemon_launchd_label_in` mirrors `install-trusty-mpm-signed.sh:305-322`'s
`resolve_mpm_plist` glob and exclusions verbatim, with one deliberate improvement: the
Rust returns `None` where the shell falls back to the canonical path. No host loses a
guard that previously fired correctly.

**HIGH-3 — CLOSED, and the test is genuinely mutation-sensitive.** I reintroduced a
hardcoded `tm restart` in the "could not compare" branch (`version_staleness.rs:156`) and
the test went red on the right assertion:

```
---- core::version_staleness::tests::staleness_uses_the_caller_restart_hint stdout ----
panicked at crates/trusty-mpm/src/core/version_staleness.rs:213:13:
message: could not compare daemon version `garbage` … restart the daemon if unsure (`tm restart`)
test result: FAILED. 9 passed; 1 failed
```

Not tautological: `HINT` is deliberately not `tm restart`, the loop covers all three
hint-carrying branches (`("0.42.0","0.41.9")` → older, `("0.42.0","")` → empty,
`("0.42.0","garbage")` → unparseable), and the `!msg.contains("tm restart")` assertion is
what catches a leftover literal. Edit reverted; tree clean.

**The eighth site you flagged — the engineer's reasoning holds, but only halfway.** Traced
and executed. `tm services restart trusty-mpm-daemon` runs `sh -c "tm restart"`
(`services.rs:171-174`), and `restart`'s refusal sits at `daemon.rs:292`, **before** the
`pkill` at `:302`:

```
Error: refusing to spawn: the launchd unit `com.trusty.mpm` is registered … Restart it with
`launchctl kickstart -k gui/$(id -u)/com.trusty.mpm` …
trusty-mpm-daemon: restart command failed (exit exit status: 1)
EXIT=1
```

No orphan spawned, nothing torn down, message actionable — so HIGH-3 is not a functional
hole, and I am not re-raising it as HIGH. But the PR body's "fixed all hits, not a subset"
is wrong, and a manifest verb that can only ever fail is worth one line of YAML
(MEDIUM-A). `start_cmd` / `stop_cmd` in the same file are never executed by any code path
— confirmed by grep — so they carry no risk.

### Verified CLEAN — do not redo

- **The `connectors::tm::create_session_full_lifecycle` failure.** Accepted as
  pre-existing on the stronger evidence relayed to me (reproduced on the pristine
  `origin/main` binary at load 52.85). Not re-derived. My own full-suite run at head was
  0 failed.
- **The `statusline::branch` flake.** Did not reproduce; my run was 44 suites / 0 failed.
  No `statusline` file appears in `git diff --name-only origin/main...HEAD`.
- **CI absence.** Explained by the missing `refs/pull/4466/merge` while the PR conflicted;
  24 checks green at this head. Not a broken workflow.
- **Doc pointers.** All 39 test names cited in the new and edited doc comments resolve to
  a real `fn` (checked individually); CI's pointer lint passes.
- **`--force` handling.** Live: a `--force` daemon reports `unsupervised_forced: true` and
  `daemon_orphan` downgrades to `⚠️ … started deliberately with tm daemon --force, so this
  is a warning`. MEDIUM-1 from round 1 is closed.
- **LOW-1 (double `/health` GET).** `misc.rs:163` fetches once and passes the snapshot to
  both checks; `stale_daemon_check` is now sync. Module doc matches the code.
- **`restart`'s guard-before-`pkill` ordering.** Unchanged and still correct.
- **Test quality.** Sampled all 16 `doctor_orphan_tests.rs` cases plus the new
  `launchd_probe` and client-wire tests. They test the pure `verdict` / `parse_*`
  functions, not mocks, and the negative cases are real —
  `authoritative_verdict_ignores_a_lying_self_report` asserts BOTH directions,
  `fallback_unknown_when_old_daemon_claims_supervised` carries an explicit
  `assert_ne!(status, Ok)`, and `parse_launchctl_pid_ignores_other_numeric_keys` pins the
  bogus-PID-by-accident case. No test asserts on mock behaviour. The one real coverage gap
  is the launchctl-unavailable state, which the current signature cannot express.

### Note for the release, not a finding

`daemon_orphan` returns `Unknown` on every host whose daemon predates this build, and
`Unknown` ranks above `Warn` (`core/doctor.rs:54-66`), so `tm doctor` cannot read overall
`Ok` until the daemon is restarted onto a build that reports `pid`. Confirmed live on this
machine. Intentional per #4005, transient for one release — worth expecting rather than
debugging.

### Required changes before this lands

1. Give `launchd_owned_pid` a third state and stop treating "could not ask launchd" as
   "launchd runs nothing". Query `gui/$(id -u)/<label>` explicitly, and gate the hard
   `Fail` on the "launchd runs NO process" branch behind `supervised != Some(true)` so a
   supervised daemon can never be told to kill itself. Add the test the current signature
   cannot express.
2. `default-services.yaml:42` — drop the `tm restart` verb (MEDIUM-A).
3. Tighten the drift glob to plists whose `ProgramArguments` actually run `daemon`
   (MEDIUM-B).
4. The three LOW doc/serde fixes are cheap enough to fold into the same push.

bobmatnyc:
## Round 3 — the fail-closed HIGH is closed, reproduced before and after

You were right, and the framing was the useful part: I closed fail-open in round 1
and opened fail-closed in round 2. Same error, inverted — absence of evidence read
as evidence of a state. The consequence was worse the second time, because this
one prescribed `kill -TERM`.

Head `1b26adba`. Gates on the merged tree: clippy exit 0, fmt clean, line-cap 0
violations, test-pointers 0 dangling, `cargo test -p trusty-mpm` exit 0 (44 suites,
4045 + 1362 passed, 0 failed).

### Reproduced FIRST, as asked

Daemon held constant across all three: pid 73599 — which is genuinely what launchd
runs for `com.trusty.mpm` on this host — reporting `supervised: true`. Same plist,
same host. The ONLY variable is whether `launchctl` can answer.

```
BEFORE (4590819e)
A. real launchctl          ✅ launchd unit `com.trusty.mpm` owns the daemon answering /health (pid 73599, version 1.3.1)
B. launchctl exits 113     ❌ ORPHAN daemon serving: launchd runs NO process for `com.trusty.mpm` … yet pid 73599 is
                              answering /health … Fix: `lsof …`, `kill -TERM <pid>`, then `launchctl kickstart …`
C. launchctl off PATH      ❌ ORPHAN daemon serving: launchd runs NO process for `com.trusty.mpm` … `kill -TERM <pid>` …
```

```
AFTER (this push)
A. real launchctl          ✅ launchd unit `com.trusty.mpm` owns the daemon answering /health (pid 73599, version 1.3.1)
B. launchctl exits 113     ❔ cannot determine whether pid 73599 (version 1.3.1) answering /health is the process
                              launchd runs for unit `com.trusty.mpm`: `launchctl` did not answer. That happens when
                              `launchctl` is not on `PATH`, in a sandbox, or when the unit is not visible in this
                              shell's bootstrap domain … This is NOT evidence of an orphan (issue #4230), so no
                              corrective action is prescribed. To check by hand, compare `launchctl print
                              gui/$(id -u)/com.trusty.mpm` against `lsof -nP -iTCP -sTCP:LISTEN | grep tm`.
C. launchctl off PATH      ❔ (same verdict)
```

Controls, same run — the detection must not have been traded away:

```
D. genuine orphan (launchd runs 73599, pid 999002 serving, supervised:false)  ❌ Fail  (unchanged)
E. LYING daemon  (launchd runs 73599, pid 999003 serving, supervised:TRUE)    ❌ Fail  (HIGH-1 still closed)
F. job DOWN + supervised:true  (two authorities contradict)                   ❔ Unknown, no kill
G. job DOWN + supervised:false (#4230 verbatim)                               ❌ Fail  with the full remediation
H. `list` blind, `gui/<uid>` print CAN answer                                 ✅ Ok    (the retry is load-bearing)
```

Machine-checked that the three non-positive verdicts contain no `kill`, no
`-TERM`, and no `ORPHAN` claim.

### The fix

`LaunchdOwnership` is `Running(pid)` / `NotRunning` / `Unavailable`. The first two
require a POSITIVE exit-0 answer from launchd; `ownership_from_output` is the only
place allowed to read a missing PID as `NotRunning`, and it is only reachable after
exit 0, which is what makes that reading safe.

Two guards keep `Fail` off a healthy daemon:

- `Unavailable` never produces `Fail`.
- `NotRunning` escalates only when `supervised != Some(true)`. Your prescription,
  and your reasoning for why it costs nothing holds: the real #4230 payload
  reported `supervised: false`, so scenario G still Fails. A lying daemon can reach
  `Unknown` but never `Ok`.

### One correction to the finding — trigger 2 does not reproduce via `launchctl list`

Your domain evidence is right, but it does not reach this code path, and I could
not reproduce the false Fail that way. `launchctl print user/502/com.trusty.mpm`
does fail as you showed — but the probe calls `launchctl list <label>`, the legacy
domain-agnostic API, and over a real `ssh localhost` session that call **succeeds**:

```
$ ssh localhost 'launchctl list com.trusty.mpm | grep PID; launchctl managername'
	"PID" = 73599;
Background
```

`managername` reports `Background` in both my GUI-descended shell and the SSH one,
so the two are not distinguishable here.

What I did confirm is the ambiguity underneath your point, which is the same hazard
by a different route: `launchctl list` exits **113** with
`Could not find service "…" in domain for port` BOTH for a unit that was never
bootstrapped and for one not visible in the caller's domain. Those are
indistinguishable from that answer alone, and the plist is on disk either way —
that is exactly why 113 must be `Unavailable` rather than `NotRunning`, and why
the fix escalates a 113 to an explicit `gui/<uid>` query before giving up
(scenario H proves the retry recovers a definite answer). Trigger 1 reproduces
verbatim and is sufficient on its own.

### MEDIUM — drift glob

Tightened to require `ProgramArguments` containing a literal `daemon` argument,
scanned only inside the `ProgramArguments` array so a `daemon.log` path cannot
match. Tests: `daemon_label_ignores_a_decoy_sibling_unit`,
`daemon_label_picks_the_daemon_unit_over_an_alphabetically_earlier_decoy` (the
decoy sorts first, so a contents-blind filter picks it),
`plist_runs_daemon_ignores_daemon_outside_program_arguments`,
`plist_runs_daemon_false_for_a_missing_file`.

The CANONICAL `com.trusty.mpm.plist` short-circuit is deliberately not gated on the
contents check, and I've documented why: a false negative there makes
`mpm_launchd_plist_exists()` report false on a correctly installed host, standing
the #4230 spawn guard down and reintroducing the orphan. The failure modes are
asymmetric; the decoy risk exists only on the drift path.

### MEDIUM — the eighth restart site: declining the code change, fixing the claim

I kept `restart_cmd: "tm restart"` and corrected the false claim instead. `null`
would be strictly worse in both directions:

- on a host with NO daemon launchd unit, `tm restart` is correct and SUCCEEDS
  (`compute_daemon_refuse(supervised=true, force=false) = false`) — `null` breaks a
  working path;
- on a launchd host, the refusal names the exact `kickstart` recipe for that host's
  label and `tm services restart` surfaces it verbatim before exiting 1 — strictly
  more actionable than `no restart_cmd defined in manifest`.

The reasoning is now a comment beside the value so this isn't re-raised. The false
sentence in my round-1 response is struck with a correction pointing here, and the
PR body now has a "Restart-verb sites" section stating the exclusion explicitly.

### The three LOWs

- **`pid` serde default** — fixed. `#[serde(default)]`, matching its neighbours.
- **"fail OPEN" doc** — fixed, and expanded: it now says the caller reads a missing
  PID as `NotRunning`, so an unparsed shape condemns a healthy daemon rather than
  waving it through.
- **release-workflow.md** — fixed. The `Unknown` claim is now scoped to both
  undetermined states, and states that a hard `Fail` requires a positive answer
  from launchd.
- **tm-cli-operations.md** — fixed. The hardcoded label is gone; the reader is told
  to take it from `tm doctor`, with the recipe shown as a shape.

### Mutation-tested, since the last round asked

- Revert the `Unavailable` arm to `orphan(...)` → **3 red**
  (`unavailable_launchctl_is_unknown_not_a_fail`, `…never_prescribes_a_kill`,
  `…message_names_the_domain_cause`).
- Widen the supervised guard into a blanket downgrade → **6 red**, including
  `not_running_with_supervised_false_is_still_a_hard_fail` and
  `authoritative_fail_when_launchd_runs_nothing_but_a_daemon_answers`.
- Drop the decoy filter → **2 red**.

Tree restored and green after each.

### Note on the changelog

#4477 landed mid-round and removed the `## [Unreleased]` heading this branch's
entry targeted. Migrated to `crates/trusty-mpm/changelog.d/4230-orphan-daemon-spawn.md`;
no `## [Unreleased]` heading re-added. `assemble-changelog.sh trusty-mpm 1.3.2
--check` → `OK: 8 fragment(s) valid`, and `check_changelog_fragment.sh` passes.

### Not re-derived

Your verified-clean list is taken as given — the `create_session_full_lifecycle`
and `statusline::branch` flakes, the CI-absence explanation, doc pointers,
`--force` handling, LOW-1, and the `pkill` ordering. The `statusline` flake did
recur once here under concurrent load and passed on three isolated re-runs; it is
still not in this diff.

bobmatnyc:
### CI green at `8d630f2c` — and one red I caused in between

**24 SUCCESS, 1 SKIPPED (`Notify on red main`), 0 failures**, including the new
`Per-PR changelog fragment` gate.

My first push of the three-state probe (`1b26adba`) turned four checks red —
Clippy, MSRV, Test, and the tm-capabilities drift check — all on a single
`E0433`. The `gui/<uid>` retry called
`trusty_common::launchd::current_uid()`, the obvious shared helper for exactly
this. That module is `#[cfg(target_os = "macos")]` while `launchd_probe` is not,
so it compiled on my machine and could not compile on Linux.

Fixed at `8d630f2c` by calling `libc::getuid()` directly — which is what the
helper wraps, and `libc` is an unconditional dependency of this crate. I left a
comment at the call site saying why the shared helper is not used, since reaching
for it is the natural next edit.

Worth noting for the record: my local gate is `-p trusty-mpm` on macOS and
structurally cannot catch a cfg-gated cross-platform break. The workspace does not
cross-compile here either (`ring`'s build script wants a Linux C toolchain), so I
verified the narrow question with a scratch crate — `libc::getuid()` checks clean
against both `x86_64-unknown-linux-gnu` and `aarch64-apple-darwin`. CI was the
thing that actually caught it.

Issue #4461, opened 2026-07-31
[trusty-agents] Bundled-agent reprovision destroys all but the first local divergence — .stale.bak written only when absent


bobmatnyc:
## Defect

Bundled-agent reprovision backs up divergent on-disk files ONLY when no backup already exists, destroying subsequent local customizations with no warning or recovery mechanism.

The bug is in `reprovision_bundled_agents_locked` at `crates/trusty-agents/src/agents/bundled/mod.rs:223`. The backup logic is:

```rust
let backup = stale_backup_path(&dest);
if !backup.exists() {
    crate::state_writer::atomic_write(&backup, &current).with_context(|| {
        format!("failed to back up stale bundled file to {}", backup.display())
    })?;
    report.backed_up += 1;
}
```

**Keyed on absence, not content.** The FIRST time a file diverges, its prior state is preserved in `<file>.stale.bak`. Every subsequent divergence of that same file overwrites the original with NO backup, because the `.stale.bak` from the first occurrence still exists. Since these files live in a user-writable location and the whole point of the tier is local customization, repeat divergence is the normal case.

## Observed Incident (2026-07-31)

- `~/.trusty-agents/agents/cto-assistant/agent.toml` had been hand-edited on 2026-07-29 (mtime 15:06:42, later than the 14:30:02 mass-provision pass) to add `okg_ingest_docstore` and `okg_sources` to its allow-list, annotated as a temporary owner-requested OKG test grant.
- On 2026-07-31 a rebuilt `tagent` (v0.38.6 build #860) started; the stamp mismatched (37efd46d → 36fc62df), and reprovision refreshed 5 files: `assistant/agent.toml`, `cto-assistant/agent.toml`, `izzie/agent.toml`, `izzie.toml`, `personal-assistant.toml`.
- **ZERO new `.stale.bak` files were created**, because all five already had one from earlier cycles (Jul 23 for cto-assistant; Jul 24 for izzie and personal-assistant).
- The hand-edit was **unrecoverable from any mechanism the tool provides**. It survived only because a precautionary full-directory snapshot was taken manually beforehand, at `~/.trusty-agents/agents-backup-20260731-104655/`.
- **Without that ad-hoc snapshot, two days of local customization would have been silently and permanently destroyed**, with no error, no warning, and no diagnostic trace.

## Context

[...]

Issue #4414, opened 2026-07-30
execute_dispatches_to_python_and_parses_json flakes under parallel load (uv reads $HOME while other tests mutate it)


bobmatnyc:
**Defect:** `tools::python_skill::tests::execute_dispatches_to_python_and_parses_json` shells out to `uv`, which reads `$HOME`, while roughly 20 tests in the same crate mutate `HOME` process-wide. Under parallel load the test intermittently fails.

**Evidence:** A/B measured during PR #4411, not assumed — `main` failed 2 of 5 full runs; the PR branch failed 0 of 5. Pre-existing, and not made worse by that PR.

**Resolution:** serialize this test (or give it an isolated `HOME`) so it cannot race the tests that mutate the variable.

**Family:** same hermetic-isolation class as #3451 and the trusty-memory env-var issue #4413 — cross-link them all once complete.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Cross-link: #4413 — sibling issue in the same hermetic-test-isolation epic.

Issue #473, opened 2026-05-30
refactor(trusty-search): split commands/discover.rs (now 559 lines, over 500-line cap)


bobmatnyc:
The #470 change pushed `crates/trusty-search/src/commands/discover.rs` to ~559 lines, exceeding the project's 500-line hard cap (see CLAUDE.md § 500-line file size hard cap).

Split into focused submodules at the next feature that lands there:
- Project-marker detection logic
- Auto-discovery orchestration
- Index registration and reindex handling

This follows the pattern used successfully in other parts of the codebase and keeps individual modules focused and testable.

bobmatnyc:
Verify-close check (not closing): acceptance criteria are **not yet met**.

The ask is to split `crates/trusty-search/src/commands/discover.rs` below the 500-line hard cap. As of `main` the file is still **559 lines** (verified via `wc -l`), i.e. unchanged and still over the cap. No split has landed.

Leaving this issue OPEN. (Note: `crates/trusty-mpm/src/daemon/discover.rs` is a separate 219-line file and is within the cap — not the subject of this ticket.)

This was reviewed as part of a docs/config quick-wins pass (#509); the Rust file was intentionally **not** refactored there to keep that PR docs-only.

PR #4471, opened 2026-07-31
feat(trusty-agents): bundle the cto-assistant OKG docstore grant

bobmatnyc:
## What

`okg_ingest_docstore` and `okg_sources` are now named in the bundled
`cto-assistant` package's `[tools].allow` (owner decision, 2026-07-31,
"add it to the bundled source") instead of living as a hand edit in the
deployed copy at `~/.trusty-agents/`.

Files changed:
- `crates/trusty-agents/.trusty-agents/agents/cto-assistant/agent.toml` — the grant + its rationale comment
- `crates/trusty-agents/src/agents/tests/loading.rs` — `bundled_cto_assistant_pins_okg_docstore_reach`
- `crates/trusty-agents/CHANGELOG.md`

## Why

The hand edit existed because there is still no API to grant a tool
(#3890). A `tagent` rebuild's bundled-agent reprovision then overwrote
the deployed file and destroyed the grant with no `.stale.bak` (#4461),
so restoring it by hand only reset the same timer. Bundling is what makes
it durable — the next reprovision converges on this content instead of
eating it.

## Reviewer note — this is a no-op on resolved behaviour, and deliberately so

Worth stating plainly, because it changes how to read the diff. The base
`assistant` has granted **all four** OKG tools since #3883 (2026-07-25),
and `extends` unions `[tools].allow` base-first, so `cto-assistant`
already resolved `okg_ingest_docstore` and `okg_sources` before this PR.
Verified against `origin/main` by resolving the persona through
`AgentConfig::by_name`:

```
resolved okg tools: ["okg_ingest_docstore", "okg_ingest_gmail",
                     "okg_ingest_drive", "okg_sources"]
```

Two consequences:

1. **The grant is an independence guarantee, not a new capability.** It
   pins the docstore reach this persona depends on so a future narrowing
   of the base cannot silently remove it — exactly what #4420 just did to
   `git_log`/`git_status`, and the same reasoning this file already
   applies to its `scopes` line and its git carve-out.
2. **The old annotation's safety claim was already false.** The deployed
   comment said `okg_ingest_gmail`/`okg_ingest_drive` "are deliberately
   NOT granted, so this cannot touch a real inbox or Drive". The persona
   reaches both today, via the base. The new comment does not repeat that
   claim; it states the true and enforceable version — *this overlay*
   never adds them, and widening it is a deliberate decision about what
   the persona may ingest, not a cleanup to match the base. Whether the
   base should hand Gmail/Drive ingest to every assistant overlay is a
   separate call, out of scope here.

## Scope

`cto-assistant` package only. The base `assistant`, `izzie`, and every
other persona are untouched, as is the existing four-tool git carve-out.
The flat shadow `cto-assistant.toml` is deliberately **not** edited: the
loader prefers the package, so the flat file is not the resolved persona
— it only loads if the `extends` chain fails, and in that already-degraded
path a narrower surface is the safe direction. No `scopes` change is
needed either; the OKG tools return `None` from `ToolExecutor::scope()`.

## Test

`bundled_cto_assistant_pins_okg_docstore_reach` asserts both halves:

- the **resolved** reach (`AgentConfig::by_name` + `match_any_glob`) — the
  capability the persona depends on;
- the **overlay's own declaration** (raw TOML parse) — independence from a
  base that may narrow later, plus the narrowness constraint.

The second half is load-bearing: a resolution-only assertion stays green
through a full revert of this grant, because the base re-supplies both
tools. Both halves were verified by mutation:

- removing the two lines → `the cto-assistant package must DECLARE 'okg_ingest_docstore' itself rather than lean on the base's union` (FAILED)
- adding `okg_ingest_gmail` → `the cto-assistant package must NOT declare 'okg_ingest_gmail': this grant is narrow on purpose` (FAILED)

Narrowness is pinned on the overlay only, deliberately — asserting the
persona does *not resolve* the Gmail/Drive ingest tools would be asserting
a falsehood.

## Gates

`cargo check`, `cargo test -p trusty-agents` (3111 passed / 0 failed),
`cargo clippy -p trusty-agents --all-targets -- -D warnings`,
`cargo fmt --check`, `check_line_cap.sh` (0 violations),
`check_test_pointers.sh` (exit 0) — all green.

One suite run hit `execute_dispatches_to_python_and_parses_json`, which is
the known pre-existing parallel-load flake tracked as #4414 (uv reads
`$HOME` while other tests mutate it); it passes in isolation and on rerun,
and is untouched by this diff.

### DIFF
--- a/crates/trusty-agents/.trusty-agents/agents/cto-assistant/agent.toml
+++ b/crates/trusty-agents/.trusty-agents/agents/cto-assistant/agent.toml
@@ -120,6 +120,27 @@ allow = [
     # (search_gmail_messages/get_gmail_message_content/modify_gmail_messages
     # are already granted by the base's Gmail surface).
     "create_draft",
+    # OKG docstore reach — a SHIPPED grant (owner decision, 2026-07-31), not a
+    # hand edit. It was one in the DEPLOYED copy from 2026-07-29 (there is
+    # still no API to grant a tool, #3890) until a bundled-agent reprovision
+    # overwrote that file and destroyed it with no `.stale.bak` (#4461).
+    # Declaring it here is what survives the next reprovision.
+    #
+    # NARROW ON PURPOSE: the two read/ingest DOCSTORE tools only. This overlay
+    # deliberately does NOT name `okg_ingest_gmail`/`okg_ingest_drive` —
+    # widening it is a decision about what this persona may ingest, never a
+    # cleanup to "match the base".
+    #
+    # Not a FILTER, though, and the difference matters: the base `assistant`
+    # has granted all four OKG tools since #3883, and `extends` unions
+    # base-first, so the RESOLVED persona reaches the Gmail/Drive ingest tools
+    # today regardless of this list. These two lines are an independence
+    # guarantee — they pin the docstore reach this persona actually depends on
+    # so a future narrowing of the base cannot silently take it away. Same
+    # reasoning as the `scopes` line below and the git carve-out above. Pinned
+    # by `bundled_cto_assistant_pins_okg_docstore_reach` (agents/tests/loading.rs).
+    "okg_ingest_docstore",
+    "okg_sources",
 ]
 # #3938: OpenRPC scopes (#3208) for the Gmail/Tasks surface above. These are
 # NOT a delta the base can supply and they are NOT redundant with `allow`:
--- a/crates/trusty-agents/CHANGELOG.md
+++ b/crates/trusty-agents/CHANGELOG.md
@@ -7,6 +7,25 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ---
 ## [Unreleased]
 
+### Changed
+
+- **`cto-assistant` now declares its OKG docstore grant in the bundled
+  persona.** `okg_ingest_docstore` and `okg_sources` are named in the bundled
+  package's `[tools].allow` (owner decision, 2026-07-31) instead of living as
+  a hand edit in the deployed copy, where a bundled-agent reprovision
+  overwrote the file and destroyed the grant with no `.stale.bak` (#4461).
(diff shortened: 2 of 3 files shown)


bobmatnyc:
## Follow-up directive: "pin the exclusion too" — BLOCKED, tracked in #4473

Owner directive received mid-review (2026-07-31): also pin, through the real
loader, that `cto-assistant` does **not** resolve `okg_ingest_gmail` /
`okg_ingest_drive`.

That assertion cannot be added here because it is false. I wrote the exact
assertion requested — resolved via `AgentConfig::by_name`, glob-aware via
`match_any_glob` — and it fails on this branch:

```
'cto-assistant' must NOT resolve 'okg_ingest_gmail'.
Resolved allow-list: ["delegate_to_agent", "okg_ingest_docstore",
  "okg_ingest_gmail", "okg_ingest_drive", "okg_sources", ...,
  "search_gmail_messages", ..., "get_drive_file_content", "search_drive_files", ...]
```

The base `assistant` has granted all four OKG tools since #3883, `merge_extends`
unions `[tools].allow` base-first, and `ToolsConfig` has no subtractive key — so
an overlay cannot narrow what its base grants. Making the assertion true means
narrowing the **base**, which would also strip `izzie`, the mail-ingesting
persona. That is an owner call across three personas, not a `cto-assistant`
edit, and it is outside this PR's scope.

Worth noting separately: the exclusion would not achieve its stated goal
regardless. This persona already holds `search_gmail_messages`,
`get_gmail_message_content`, `modify_gmail_messages`, `create_draft` (#473) and
`get_drive_file_content` / `search_drive_files` / `list_drive_contents` by prior
design. It was never inbox/Drive-isolated.

#4473 carries the evidence and three options (accept / narrow the base via the
#4420 pattern / add a `[tools].deny` key). This PR is unchanged and still
correct on its own terms: it pins the positive resolved reach and the
overlay-level narrowness, both with mutation evidence.

Issue #4451, opened 2026-07-31
bundled agents unreachable by the harness after the CLAUDE_CONFIG_DIR migration, and tm doctor reports it green


bobmatnyc:
## Summary

Commit `e5f14d23` (PR #4437, closing #4409) moved bundled-agent deployment
from per-workspace `.claude/agents/` to the single tm-managed
`$CLAUDE_CONFIG_DIR/agents` tier. That migration relied on a probe matrix
(issue #4409, [comment 5136146155](https://github.com/bobmatnyc/trusty-tools/issues/4409#issuecomment-5136146155))
that concluded `$CLAUDE_CONFIG_DIR/agents/` **is** discovered by the harness
in a managed session, both at session start and mid-session. A live managed
session now contradicts that conclusion: every bundled specialist is
unreachable, and `tm doctor`'s `agents`/`deployment` checks both report
green throughout. This is filed as one issue, not two, because a fix that
only restores dispatch without also giving `tm doctor` a resolvability probe
will regress silently again the same way this one did.

## Half 1 — bundled specialists are unreachable post-migration

Observed in a live managed session (`tm-search`, id
`a8d30388-ee31-4578-ab4d-a86f576f980b`, worktree
`/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.worktrees/tm-trusty-tools-04`)
and **independently reproduced on disk** while investigating this issue, in
the same worktree:

- Dispatching any bundled specialist fails: `Agent type 'local-ops' not
  found. Available agents: claude, claude-code-guide, Explore,
  general-purpose, Plan, statusline-setup`.
- `$CLAUDE_CONFIG_DIR` = `/Users/masa/.trusty-tools/trusty-mpm/claude-config`.
  Its `agents/` dir holds 42 well-formed agents (confirmed on disk,
  including a valid 19,674-byte `local-ops.md`), plus
  `.trusty-mpm-manifest.json` and 5 `BASE-*.md` files.
- The real `~/.claude/agents/` holds **only** `.bak-<timestamp>`-renamed
  copies of those same 42 agents (e.g.
  `local-ops.md.bak-1783907402022298000`, confirmed on disk), plus a
  handful of unrelated non-trusty-mpm agents (`base.md`, `copyeditor.md`,
  `proofreader.md`, `writer.md`, `writing-critic.md`).
- The worktree's `.claude/` directory has **no `agents/` subdirectory at
[...]

Issue #2997, opened 2026-07-18
fix(trusty-mpm): TCC app-data re-prompts via responsibility-attribution to ancestor daemon for ad-hoc cargo build children (#2721 part 4)


bobmatnyc:
## Summary

#2721/#2723/#2736 stabilized `tm`/`trusty-mpm`'s Developer ID signature; #2957 fixed the same churn class for `trusty-mpm-gui`. Both are **confirmed intact** as of 2026-07-18 (`com.trusty.tm` / `com.trusty.trusty-mpm`, Developer ID Application 4JH68XUHC5, verified via `codesign -dvvv`; launchd `com.trusty.mpm` points at the signed `~/.cargo/bin/trusty-mpm`).

Despite that, the «"trusty-mpm" would like to access data from other apps» prompt recurred today.

## Evidence

`log show --predicate 'subsystem == "com.apple.TCC"'` (48h window) shows the only trusty-related event is a `kTCCServiceDeveloperTool` `AUTHREQ_ATTRIBUTION` where `trusty-mpm` is marked the **responsible ancestor** for a `cargo build` build-script (`libgit2-sys` `build-script-build`, ad-hoc identity `build_script_build-fc98b0cbade2e816`) compiled inside an agent worktree (`feat-2993-diff-ingestion`). The `responsible` identifier logged was the Rust ad-hoc hash-suffixed `trusty_mpm-a56c21ffb1b879b1`, not the installed `com.trusty.trusty-mpm`.

I.e.: trusty-mpm gets blamed/prompted for access requests made by the **constantly-churning, ad-hoc-signed build artifacts** of every `cargo build` its managed sessions run — its own signature stability no longer matters for this class.

## Compounding hygiene findings (same investigation)

- An **orphaned ad-hoc `tm daemon`** (PID 26384) running since 01:27 AM from worktree `fix-1737-explicit-url` — whose directory has since been **deleted from disk** (binary file gone; process still alive, PPID=1, polling Telegram API on an ephemeral port).
- **4 stale ad-hoc `trusty-mpm-gui` debug binaries** in various worktree `target/debug` dirs, one built 29 minutes **before** the #2957 merge (still carries the old `com.trusty-mpm.gui` identity / `"trusty-mpm"` product name).

Neither is proven to be today's specific trigger, but both reproduce this bug class if touched/relaunched.

## Ask

[...]

Issue #2020, opened 2026-07-05
trusty-mpm: /claude-config/restart can relaunch a managed session as bare `claude`, dropping isolation


bobmatnyc:
## Problem
`POST /claude-config/restart` (`crates/trusty-mpm/src/daemon/.../claude_config_routes.rs`, handler ~lines 378-409) takes a caller-supplied arbitrary `tmux_session` string and calls `ClaudeCodeRestarter::restart_in_session` (`crates/trusty-mpm/src/daemon/claude_config/restarter.rs:57-64`), which relaunches with a bare `claude` command line. The handler binds `State(_state)` (underscore-prefixed) and NEVER consults the session registry to check whether the target is a tm-managed, `CLAUDE_CONFIG_DIR`-isolated session.

If this endpoint is pointed at a managed session, the bare `claude` relaunch silently drops all four isolation properties the managed-spawn builder (`runtime/claude_code.rs:108-115`) applies:
- `CLAUDE_CONFIG_DIR='<dir>'` auth/config isolation
- `-u ANTHROPIC_API_KEY` env scrub
- `--setting-sources project,local` (roster isolation — reintroduces the DOC-34 roster-drift problem)
- `--dangerously-skip-permissions` (an unattended managed session would now block on a permission prompt)

This is a latent pre-existing risk (surfaced during review of PR #2017 / issue #2010, which only unified the bare literal and did not change behavior). No in-tree caller of the endpoint currently exists (checked CLI, dashboard, trusty-mpm-gui), so it is not known to fire today — but the endpoint is public on the daemon API surface.

## Fix scope
In the restart handler: validate the requested `tmux_session` against `DaemonState`'s session registry. If it resolves to a managed session, relaunch it through the managed command builder (with its `CLAUDE_CONFIG_DIR`/env-scrub/setting-sources/skip-permissions), not bare `claude`. For non-managed / unknown sessions, keep the current bare-`claude` behavior. Alternatively, reject restart requests targeting managed sessions with a clear error if in-place managed relaunch is out of scope.

## Acceptance criteria
- A restart request targeting a managed session relaunches with full isolation (CLAUDE_CONFIG_DIR + ANTHROPIC_API_KEY scrub + --setting-sources project,local + --dangerously-skip-permissions), OR is rejected with an actionable error.
- Non-managed sessions unaffected.
- Test coverage for both the managed and non-managed target cases.
[...]

PR #4475, opened 2026-07-31
fix(trusty-mpm): scrub inherited CLAUDE_CODE_CHILD_SESSION so managed sessions keep transcripts

### LINKED ISSUE #4467: Managed sessions silently lose Claude Code transcripts: inherited CLAUDE_CODE_CHILD_SESSION disables session persistence
## Defect

`tm` never scrubs Claude Code's process-local environment markers when spawning a managed `claude`. The inherited `CLAUDE_CODE_CHILD_SESSION` marker makes Claude Code disable transcript saving for the managed session.

Claude Code 2.1.217 started surfacing this:

> Transcript saving is off — inherited CLAUDE_CODE_CHILD_SESSION marker
> · restart with CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1 to keep future transcripts

## Evidence

Observed in a live managed session:

```
CLAUDECODE=<set>            CLAUDE_CODE_CHILD_SESSION=<set>
CLAUDE_CODE_ENTRYPOINT=<set> CLAUDE_CODE_EXECPATH=<set>
CLAUDE_CODE_SESSION_ID=<set> CLAUDE_CODE_MAX_OUTPUT_TOKENS=<set>
CLAUDE_CONFIG_DIR=<set>      CLAUDE_EFFORT=<set>  CLAUDE_PID=<set>
```

`grep -rn "CHILD_SESSION" crates/ scripts/ docs/` returns **nothing** — tm neither sets nor scrubs it. It is inherited leakage reaching the spawned `claude`.

The gate, from the `claude` 2.1.220 bundle (`strings`):

```js
function Zkt(){                                              // "disable session persistence?"
  if (Z.CLAUDE_CODE_FORCE_SESSION_PERSISTENCE) return false;
  if (!(Z.CLAUDE_CODE_CHILD_SESSION && PN() && !o_())) return false;
  return !Tsg();                                             // Tsg(): tmux show-environment -g CLAUDE_CODE_CHILD_SESSION
}
```

`CLAUDE_CODE_CHILD_SESSION` is the sole trigger. The same bundle shows where the leak originates — Claude Code's own child-process env injector:

```js
function KDt(e){ let t={CLAUDECODE:"1", CLAUDE_CODE_SESSION_ID:e.sessionId,
  CLAUDE_CODE_CHILD_SESSION:"1", CLAUDE_PID:String(process.pid)};
  if(e.source==="agent") t.AI_AGENT=...; if(e.effortLevel!==void 0) t.CLAUDE_EFFORT=...; }
```

Any `tm` invoked from inside a Claude Code session inherits these and passes them to the `claude` it spawns.

**Why this looked intermittent.** `tmux show-environment -g CLAUDE_CODE_CHILD_SESSION` currently returns `=1` on this machine: the tmux *server's* global env carries the marker, because the server was started from inside a Claude Code session. That trips upstream's `Tsg()` escape hatch and keeps persistence on. The mask is incidental to how the tmux server happened to be started — restart it from a clean shell and every managed session silently stops saving.

## Consequence

Managed sessions get no native `--resume` / `--continue` / `/rewind` recovery and never appear in `--resume`. tm's own pause/resume writes only a condensed summary, so a session that dies loses everything since its last snapshot.

Upstream precedent: 2.1.170 "Fixed sessions not saving transcripts (and not appearing in `--resume`) when launched from the VS Code integrated terminal or any shell that inherited Claude Code environment variables."

## What needs resolving

- Scrub the leaked markers on **every** spawn path — `spawn_command`, `resume_command`, `compose_inplace_args` (`runtime/claude_code.rs`), `build_inplace_exec_command` (`bin/tm/commands/guided_inplace.rs`), `build_claude_command` (`control/backend/stream_json.rs`). A fix covering only `spawn_command` leaves resumed sessions broken.
- Do **not** blanket-scrub. `CLAUDE_CONFIG_DIR` is set deliberately by tm and #4455 depends on it to reach bundled agents via the `user` settings tier; scrubbing it re-breaks #4451.
- Add a `tm doctor` check that fails when a managed spawn would have transcript saving disabled — same rationale as `agent_reachability` (#4455): a silent failure needs a check capable of failing.

## Open sub-question — retention

Enabling saving increases footprint. Current relocated config dir: **6.2 GB across 103 project dirs** in `~/.trusty-tools/trusty-mpm/claude-config/projects` (10430 `.jsonl`; 9931 of them under `subagents/` = 4.46 GB).

Unresolved: whether Claude Code's native 30-day sweep actually bounds a **relocated** `CLAUDE_CONFIG_DIR`. Relevant bundle string:

> Skipping retention cleanup: userSettings source is disabled (--setting-sources) and no enabled source provides cleanupPeriodDays.

tm passes `--setting-sources`, so the flag value may decide whether the sweep runs at all. To be answered with evidence before any pruning code is written — a redundant second pruner is its own liability.

bobmatnyc:
## The defect

`tm` never scrubbed Claude Code's process-local session markers when spawning a managed `claude`. When `tm` runs from inside a Claude Code session — the PM delegating, the daemon having been started from a session, an operator running `tm` in a `claude` pane — it inherits them and passed the whole environment through. An inherited `CLAUDE_CODE_CHILD_SESSION` makes Claude Code disable session persistence:

> Transcript saving is off — inherited CLAUDE_CODE_CHILD_SESSION marker
> · restart with CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1 to keep future transcripts

Managed sessions therefore had no native `--resume` / `--continue` / `/rewind` recovery and never appeared in `--resume`. tm's own pause/resume writes only a condensed summary, so a session that died lost everything since its last snapshot.

The gate, read out of the `claude` 2.1.220 bundle:

```js
function Zkt(){                                              // "disable session persistence?"
  if (Z.CLAUDE_CODE_FORCE_SESSION_PERSISTENCE) return false;
  if (!(Z.CLAUDE_CODE_CHILD_SESSION && PN() && !o_())) return false;
  return !Tsg();                                             // Tsg(): tmux show-environment -g CLAUDE_CODE_CHILD_SESSION
}
```

`CLAUDE_CODE_CHILD_SESSION` is the sole trigger. Removing it short-circuits the gate at the second condition, independent of everything else.

## Where the residual risk actually is

**Correction (this replaces an earlier claim in this body that had it backwards).** The earlier version said "a tmux server restarted from a clean shell would silently stop saving". That is the wrong way round. For a tmux **pane**, the marker's presence and the escape hatch are *coupled*: tmux's `update-environment` does not carry `CLAUDE_*`, so a marker only reaches a pane when the tmux **server's** global env carries it — which is exactly the condition that trips `Tsg()`. A clean server means the panes never inherit the marker in the first place. Tested directly on an isolated socket, both directions.

The defect is still real on the fixed paths, for two reasons the original body did not name — and both argue *for* the scrub:

1. **The escape hatch fails OPEN into suppression.** `Esg()` runs `spawnSync("tmux", …)` **by bare name** with a **250 ms timeout** and returns `false` on throw or non-zero status. A pane on the daemon's minimal launchd `PATH` (the documented #1298 hazard, `spawn_disclaim/pane.rs:83`) or a machine loaded enough to blow 250 ms both flip a masked session to not-saving. That is a better account of "it looked intermittent" than the tmux-server story.
2. **`tm run` has no tmux at all**, so `Esg()` returns `false` on its first line (`if(!Z.TMUX) return false`) and the suppression fires **every time** — deterministic, not probe-dependent.

Scrubbing the variable is what makes the outcome deterministic instead of dependent on a probe that can fail open.

## Per-variable decisions

The scrub list is not guesswork. It is what Claude Code's own child-process env injector sets (`KDt` in the same bundle), plus `CLAUDE_CODE_EXECPATH` which appears on upstream's own list of variables it refuses to propagate into a shell snapshot.

| Variable | Decision | Why |
|---|---|---|
| `CLAUDE_CODE_CHILD_SESSION` | **scrub** | The sole transcript-saving trigger. Upstream strips it when spawning MCP servers too. |
| `CLAUDE_CODE_SESSION_ID` | **scrub** | The parent's session UUID; inheriting it mis-attributes the child's hook/statusline events to the parent conversation. |
| `CLAUDECODE` | **scrub** | The "inside Claude Code" marker for child processes. The spawned `claude` sets it for its own children; inheriting it is the leak upstream 2.1.170 called out for the VS Code terminal. |
| `CLAUDE_PID` | **scrub** | The parent interpreter's pid — stale by definition, and consulted by shell-snapshot probes (`/proc/$CLAUDE_PID/comm`). |
| `CLAUDE_EFFORT` | **scrub** | A snapshot of the parent's *current turn* effort level. Derived output exposed for hooks, not a config knob, so a fresh session must recompute it. |
| `CLAUDE_CODE_EXECPATH` | **scrub** | Path to the parent's `claude` binary. tm always passes a resolved absolute binary, so nothing reads it. |
| `CLAUDE_CONFIG_DIR` | **KEEP** | Set deliberately by tm. #4455 depends on the relocation to put the bundled agent roster in the `user` settings tier a managed session loads — scrubbing it immediately re-breaks #4451 (every delegation degrading to `general-purpose`). Guarded by two tests and a doctor branch, not just a comment. |
| `CLAUDE_CODE_OAUTH_TOKEN` | **KEEP** | Injected by #2246 to bypass the `CLAUDE_CONFIG_DIR`-keyed Keychain login loop. |
| `CLAUDE_CODE_MAX_OUTPUT_TOKENS` | **KEEP** | An operator-facing configuration knob, not process-local state, and not on upstream's own exclusion list. Scrubbing it would silently override the operator's intent. |
| `CLAUDE_CODE_ENTRYPOINT` | **KEEP** | Its inherited value from a `tm` parent (`cli`) is what a fresh CLI spawn computes anyway. Note the keep is *value*-dependent, not proven-safe for every value: unset is not universally equivalent to `"cli"` (the bundle gates one path on `e === "cli" \|\| e === "remote"`, so scrubbing would silently disable it), while an inherited `sdk-ts`/`sdk-py`/`sdk-cli` drops one built-in agent from the roster and `claude-vscode` changes `clientType`. Neither is reachable from a `tm` spawn, whose parent is always a `cli` session. Recorded in the module doc. |
| `CLAUDE_CODE_INVOKED_SKILLS` | **KEEP** | On the same upstream non-propagation list that justifies `CLAUDE_CODE_EXECPATH`, so the decision is stated rather than left implicit: it was not in the observed leak set, nothing in this workspace reads it, and it carries no session identity — a stale value cannot mis-attribute a session or suppress a transcript. |

`AI_AGENT` and `TRACEPARENT` are in the same upstream injector but were left alone deliberately. `AI_AGENT` self-heals: the bundle overwrites any inherited value beginning `claude-code_`/`claude-code/` with its own at startup. `TRACEPARENT` is only set when tracing is on and its sole effect is span parenting. Both are generically named, so a third party could legitimately set them.

## Launch lines fixed

Every launch line that spawns a `claude` — a fix covering only the default spawn would leave every *resumed* session broken, and the first round of this PR left three builders leaking:

- `spawn_command` and `resume_command` (`runtime/claude_code.rs`) — share `env_bin_prefix`, which now emits `env -u ANTHROPIC_API_KEY -u <markers…> [assignments] <bin>`. The `-u` flags must precede any `NAME=VALUE`; POSIX `env` stops parsing options at the first assignment and would exec `-u` as a command.
- `compose_inplace_args` / `build_inplace_exec_command` (`bin/tm/commands/guided_inplace.rs`) — execs through a `Command` with no shell, so it uses `env_remove`. The scrub runs *before* the deliberate assignments so it can never clobber `CLAUDE_CONFIG_DIR`.
- `build_claude_command` (`control/backend/stream_json.rs`) — same `env_remove` seam that already strips `ANTHROPIC_API_KEY`.
- **`build_launch_command` (`core/standalone/run.rs`) — `tm run`.** Added in the review round. The one `tm` path with no tmux, so the suppression fired here deterministically. One line: `scrub_command(&mut cmd)` before the `CLAUDE_CONFIG_DIR` assignment, so the deliberate value still wins.
- **`build_claude_command` (`core/model_inject.rs`) — the `tm launch` (`launch.rs:308`) and `tm connect` (`launch.rs:563` via `connect_claude_cmd`) launch line.** Added in review round 1. It emitted a bare `claude` with **no `env` prefix at all**, while its own doc claimed to be where "every spawn path stays unattended and isolated". Now leads with `env -u <markers…>`.
- **`relaunch_command` (`daemon/spawn_command.rs`) — the pane relaunch.** Added in review round 1. Only the marker scrub; #2020's auth/absolute-path half stays open and is still documented as such.
- **`build_inplace_session_command` (`core/model_inject.rs`) — `tm session start`, in place.** Added in **review round 2** (HIGH). `bin/tm`'s `start_session_in_place` hand-built `format!("claude {PERMISSION_MODE_FLAG}")` and sent it to a fresh pane: no scrub, no test, and the doctor named only one binary-crate gap so it read as covered. Reachable on the documented "just run claude here" case — `tm session start` outside a GitHub-backed git repo.
- **`build_client_session_command` (`core/model_inject.rs`) — `DaemonClient::launch_session` / `connect_session`, the TUI and bot surface behind `/connect <dir>`.** Found in round 2 **by the new anti-drift scan, not by review**: both hand-built `format!("claude --append-system-prompt-file …")`. Two further launch lines nobody had named.

**Why the last two moved into the library rather than being patched in place.** `daemon::doctor_transcript_saving`'s `launch_lines()` can only read builders the *library* owns, so a hand-built line at a call site is structurally invisible to the check that exists to catch exactly this. Three separate call sites had each hand-written their own `claude …` string and all three silently saved no transcript.

**Neither new builder carries `SETTING_SOURCES_FLAG`, deliberately.** Routing them through `build_claude_command` (the critic's first option) would also add `--setting-sources project,local`, which **drops the `user` tier** — and neither path relocates `CLAUDE_CONFIG_DIR`, so on them the `user` tier *is* the operator's own `~/.claude`. That would change which settings and agents an in-place session loads. The scrub is the only behaviour change; pinned by `inplace_session_command_keeps_the_permission_flag_and_nothing_else`.

**Interaction with the #2997 pane disclaim, since the CLI paths changed shape.** `tm launch`/`tm connect` wrap the line in `disclaim_pane_command`, so the pane now runs `<tm> internal-spawn-disclaimed env -u … claude …`: the shim `posix_spawn`s `env` disclaimed and `env` **execs** `claude` in that same process. The disclaim is a process attribute and survives the exec, and because `env` execs rather than forks, the `pane sh → tm shim → claude` tree `core::process` walks is unchanged. The daemon composes the two the other way round (`env … <tm> internal-spawn-disclaimed <abs claude>`); both shapes end with `claude` running as its own responsible process. The shim's `argv` already carries `trailing_var_arg` + `allow_hyphen_values`, so the `-u` tokens parse, and `env` is no less `PATH`-resolvable than the bare `claude` this line used to emit — no new `PATH` dependency. Pinned by `wraps_the_scrubbed_launch_line_with_env_as_the_program`.

## Doctor check

New `transcript_saving` check (24 checks, was 23), modelled on `agent_reachability` (#4455): a silent failure needs a check capable of failing. Static, not a live `claude -p` spawn.

The first version parsed only `env_bin_prefix`, so it returned `Ok` while the three builders above leaked — a check that certifies a broken state is worse than no check. It now **enumerates every launch-line builder the library owns** and reads each one's real output: `parse_env_unset_vars` for shell-string builders, `get_envs()` for `Command` builders. It returns the **first** failure and **names the offending builder**, so one leaking path cannot be averaged away by four correct ones. `bin/tm`'s `build_inplace_exec_command` lives in the binary crate the library cannot reference; the Ok message names that gap rather than implying coverage, and it has its own test.

It fails in **both** directions. Over-scrub detection now iterates `DELIBERATE_SPAWN_ENV` and the probe supplies a token, so an over-scrub of `CLAUDE_CODE_OAUTH_TOKEN` (#2246) is visible too — the earlier literal `"CLAUDE_CONFIG_DIR"` comparison could not see it.

**Verified non-tautological by mutation, on the final tree.** Each mutation applied alone, tests run, file byte-restored (`cmp -s`) afterwards:

| # | Mutation | Result |
|---|---|---|
| M2 | drop `scrub_command` from `standalone::run::build_launch_command` | FAIL, names `(tm run)`, `Currently unsets: []` |
| M1 | `model_inject::build_claude_command` → bare `"claude"` | FAIL, names `(tm launch / tm connect)` |
| **M7** | `build_inplace_session_command` → `format!("claude {PERMISSION_MODE_FLAG}")` | FAIL on 4 tests, names `` `core::model_inject::build_inplace_session_command (tm session start, in place)` ``, `Currently unsets: []` |
| **M8** | `build_client_session_command` → bare `"claude"` | FAIL on 4 tests, names `` `core::model_inject::build_client_session_command (DaemonClient launch/connect)` ``, `Currently unsets: []` |

M7 and M8 cover the two paths added in round 2. The critic independently ran M1–M6 including a partial scrub and a correctly-listed-but-wrongly-positioned scrub; both were caught.

## Anti-drift: new spawn sites now fail the build (round 2 MEDIUM)

`launch_lines()` is hand-maintained, and `launch_lines_covers_every_builder` only asserted the listed labels were **present** — nothing failed when a launch line existed that was *not* on the list. That is an anti-drift helper reintroducing drift one level down, and it is precisely how the round-2 HIGH survived two review rounds and a widening pass.

`every_claude_launch_site_in_the_crate_is_allowlisted` inverts the direction: it walks `crates/trusty-mpm/src`, counts non-comment sites naming `claude` as a command word, and asserts the result equals an explicit allowlist carrying a stated reason per entry. A new spawn site is now a failing test, not a silent gap. **It earned its keep immediately** — it is what found the two `DaemonClient` launch lines.

Honest about its reach, since a coverage test that overstates itself is the defect it guards against: it matches **literal** `claude` only. Builders that spawn a resolved binary path held in a variable (`runtime::claude_code`, `guided_inplace`) carry no literal and are invisible to it; those are covered by `launch_lines()` and their own tests. What it catches is the shape that actually went wrong three times — someone hand-writing `format!("claude …")` or `Command::new("claude")` at a new call site.

The doctor's `Ok` message now **counts and names** the binary-crate gaps from a `BINARY_CRATE_BUILDERS` constant, instead of describing one in prose (which read as *the* gap rather than *a* gap), and points at the test above. Pinned by `ok_message_names_every_binary_crate_gap`.

## Test-pinning hardening (review round)

- All six markers are now pinned by **literal name** (`every_marker_is_pinned_by_literal_name`). Every prior test iterated `INHERITED_SESSION_MARKERS`, so `CLAUDE_PID` and `CLAUDE_EFFORT` were pinned by nothing — deleting either kept the whole suite green.
- `parse_env_unset_vars` now models POSIX option termination and stops at the first `NAME=VALUE`. It previously reported `FOO` unset from `env CLAUDE_CODE_CHILD_SESSION=1 -u FOO claude`, an ordering `env` rejects with `env: -u: No such file or directory` (exit 127) — so the doctor probe could have passed on a prefix that kills every spawn.
- **Round 2 (LOW):** it now also stops at the **command word**. `build_claude_command` emits no assignments at all, so the assignment stop never fired on its output and the scan ran to end of line; a future `claude` flag pair spelled `-u <x>` would have been miscounted as an unset variable. Not reachable today — closed before a flag makes it reachable. Pinned by `parse_env_unset_vars_stops_at_the_command_word`.
- **Round 2 (LOW):** dropped `delegate` from `build_claude_command`'s doc and the doctor's label. `build_agent_command` has **no production caller anywhere in the workspace** — only its definition and its own test — so a `transcript_saving` failure labelled `(… / delegate)` would have sent an operator chasing a path that does not exist. *Declining the delete half:* `trusty-mpm` carries no `publish = false` and eight sibling crates depend on it, so removing a `pub fn` is a semver-breaking change that does not belong in a bug-fix PR. The dead function stays; removing it is follow-up.
- `spawn_command_uses_resolved_binary` no longer matches `-u CLAUDE_CODE_EXECPATH /Users/me/.local/bin/claude `, which coupled a binary-resolution test to `CLAUDE_CODE_EXECPATH` being *last* in the marker list.

## Retention finding — no pruner added

**Relocation does not defeat native retention, and I found no evidence that it does.** The mechanism that decides whether the sweep runs is the `--setting-sources` flag, not the config-dir path:

> Skipping retention cleanup: userSettings source is disabled (--setting-sources) and no enabled source provides cleanupPeriodDays.

Relocated managed spawns carry `--setting-sources user,project,local` (since #4455), so `userSettings` is enabled and `cleanupPeriodDays` is readable — the native sweep applies to the relocated dir.

Evidence gathered:

- Relocated dir: 6.2 GB, 103 project dirs, 10430 `.jsonl`. **0 files older than 30 days** — but the directory was created Jul 4 (27 days ago), so that observation carries little information and I am not claiming it as proof.
- Default `~/.claude/projects` (relocation not in play): 3412 `.jsonl`, of which **1396 are older than 30 days**, oldest 47 days. 1340 of those 1396 sit under `subagents/`.

So the accurate reading is that the native sweep is weak on `subagents/` transcripts **in both directories alike** — that is upstream behaviour, not a tm relocation bug, and nothing tm can fix by relocating differently.

**Call: build no pruning code.** A parallel pruner would be a redundant second mechanism (its own liability, per the brief) aimed at an upstream gap. I also did not add a footprint doctor warning: any threshold would be arbitrary, and a warning would not address the actual gap. The footprint numbers are recorded on #4467 as an open item instead.

Scope note on growth: subagent transcripts already save today (9931 files, 4.46 GB of the 6.2 GB). What was missing is the top-level managed-session transcript (0.82 GB across 499 files), so enabling saving grows the *smaller* bucket.

## Verification

Branch head `66f3d8da`, on `origin/main` at `a226bad6`.

**Changelog migrated to fragments (#4477).** That PR replaced the shared-`CHANGELOG.md` `## [Unreleased]` convention with per-PR fragment files and removed the heading from all 25 crates, so this PR's bullet moved out of `CHANGELOG.md` into two fragments:

- `crates/trusty-mpm/changelog.d/4467-transcript-saving.md` — `Fixed`
- `crates/trusty-mpm/changelog.d/4467-doctor-transcript-check.md` — `Added` (the new doctor check is an addition, not a fix)

`crates/trusty-mpm/CHANGELOG.md` is now byte-identical to main (`git diff origin/main -- crates/trusty-mpm/CHANGELOG.md` is empty) and carries no `## [Unreleased]` heading — `assemble-changelog.sh:265` hard-fails on a leftover one, which would break the release path for the crate.

Validated through the consumer, not just the gate: `scripts/assemble-changelog.sh trusty-mpm --stdout` renders both bullets under the correct `### Added` / `### Fixed` headings, and `scripts/assemble-changelog.sh trusty-mpm 1.3.1 --check` reports `OK: 9 fragment(s) … valid; crates/trusty-mpm/CHANGELOG.md is ready for a [1.3.1] section.` Neither mode wrote anything (tree stayed clean).

- `cargo test -p trusty-mpm` — **exit 0**, 44 targets, **6995 passed; 0 failed; 16 ignored** (pre-existing). Lib target: `test result: ok. 4082 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 55.37s`.
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — exit 0.
- `cargo fmt --check` — exit 0.
- `bash scripts/check_line_cap.sh` — `line-cap: 7 allowlisted, 0 violations — OK`.
- `bash scripts/check_test_pointers.sh` — `test-pointers: 0 dangling pointers — OK`. **This check was RED on `a06cfb8a`** (`spawn_command.rs:39` cited the renamed `relaunch_command_returns_bare_claude`); it was the only failing check on the PR and is fixed here.
- `bash scripts/check_changelog_fragment.sh` — `OK   trusty-mpm: changelog.d fragment present and valid`.
- `bash scripts/check_sld.sh`, `check_doc_numbers.sh`, `check_capabilities.sh` — all exit 0 (`up to date (6 generated files)`).
- Mutation probes on the doctor check — see above.

No `#[ignore]`, no cfg-gating, no `--exclude`, no narrowed scope, no deleted coverage.

**One flake, named rather than buried.** One earlier full run failed `connectors::tm::tests::create_session_full_lifecycle` with `Transport("error sending request for url (http://127.0.0.1:56178/api/v1/sessions/managed)")` — an ephemeral-port transport failure, the same shape as the `execute_doctor_against_test_daemon` flake noted in review. It passes in isolation (`1 passed`) and every subsequent full run is green. Unrelated to this diff: no scrubbed variable is involved in a local HTTP request.

**Not verified end-to-end interactively.** An A/B live probe (marker set vs scrubbed, PTY allocated, tmux escape hatch neutralized) wrote a transcript in *both* arms, because `claude -p` print mode never enters the suppression branch (`PN()` is false there). The suppression only fires for interactive sessions, which is where the warning was originally observed and which I could not drive non-interactively. The fix therefore rests on the decompiled gate above, the verbatim warning string naming the marker, unit tests proving every emitted launch line unsets it, and the two mutation probes on the doctor check.

Closes #4467

### DIFF
--- a/crates/trusty-mpm/changelog.d/4467-doctor-transcript-check.md
+++ b/crates/trusty-mpm/changelog.d/4467-doctor-transcript-check.md
@@ -1,0 +1,10 @@
+Added
+
+- `tm doctor` gained a `transcript_saving` check (24 checks, was 23) that fails
+  when any `tm` launch line would leave transcript saving disabled — the defect
+  it guards was silent until Claude Code began warning about it. It reads the
+  scrub set out of the real launch commands rather than restating a constant,
+  names the offending builder on failure, and fails in both directions:
+  under-scrub costs the session its transcripts, over-scrub costs it the agent
+  roster ([#4451](https://github.com/bobmatnyc/trusty-tools/issues/4451)) or the
+  OAuth token ([#2246](https://github.com/bobmatnyc/trusty-tools/issues/2246)).
--- a/crates/trusty-mpm/changelog.d/4467-transcript-saving.md
+++ b/crates/trusty-mpm/changelog.d/4467-transcript-saving.md
@@ -18,9 +18,16 @@ Fixed
   - Applied to every launch line, not just the default one: `spawn_command` and
     `resume_command` (a spawn-only fix would leave every resumed session broken),
     the in-place bare-`tm` relaunch, the headless stream-JSON backend, the
-    `tm launch` / `tm connect` / agent-delegation launch line, the pane-relaunch
-    line, and `tm run`.
-  - `tm run` mattered most and was fixed last: it spawns `claude` with no tmux, so
-    Claude Code's `tmux show-environment -g` escape hatch returns false
-    immediately and the suppression fired there every time rather than depending
-    on the tmux server's environment.
+    `tm launch` / `tm connect` launch line, the pane-relaunch line, `tm run`,
+    `tm session start`'s in-place pane, and the `DaemonClient` launch/connect
+    line behind the TUI and bot surfaces.
+  - `tm run` mattered most: it spawns `claude` with no tmux, so Claude Code's
+    `tmux show-environment -g` escape hatch returns false immediately and the
+    suppression fired there every time rather than depending on the tmux
+    server's environment.
+  - Every launch line is now built in one place. Three call sites had each
+    hand-written their own `claude …` string, and all three silently saved no
+    transcript — a hand-built line is invisible to the `transcript_saving`
+    check, which can only read builders. A new test walks the crate's source
+    and fails when a `claude` launch site appears that is not accounted for,
+    so the next one is a build failure rather than a silent gap.
--- a/crates/trusty-mpm/changelog.d/4467-transcript-saving.md
+++ b/crates/trusty-mpm/changelog.d/4467-transcript-saving.md
@@ -1,0 +1,26 @@
+Fixed
+
+- managed sessions no longer silently lose their Claude Code transcripts (closes [#4467](https://github.com/bobmatnyc/trusty-tools/issues/4467))
+  - Every managed spawn now scrubs the process-local session markers Claude Code
+    injects into its child processes. When `tm` was invoked from inside a Claude
+    Code session it inherited `CLAUDE_CODE_CHILD_SESSION` and passed it through,
+    which made the spawned `claude` turn session persistence OFF: no native
+    `--resume`, no `--continue`, no `/rewind`, and the session never appeared in
+    `--resume` listings. Since tm's own pause/resume writes only a condensed
+    summary, a session that died lost everything since its last snapshot.
+  - Scrubbed: `CLAUDE_CODE_CHILD_SESSION` (the sole transcript-saving trigger),
+    `CLAUDE_CODE_SESSION_ID`, `CLAUDECODE`, `CLAUDE_PID`, `CLAUDE_EFFORT`,
+    `CLAUDE_CODE_EXECPATH`. Deliberately NOT scrubbed: `CLAUDE_CONFIG_DIR` (tm
+    relocates it on purpose — [#4455](https://github.com/bobmatnyc/trusty-tools/issues/4455)
+    depends on it to keep the bundled roster in a settings tier managed sessions
+    load), `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_MAX_OUTPUT_TOKENS` (an
+    operator-facing knob) and `CLAUDE_CODE_ENTRYPOINT`.
+  - Applied to every launch line, not just the default one: `spawn_command` and
+    `resume_command` (a spawn-only fix would leave every resumed session broken),
+    the in-place bare-`tm` relaunch, the headless stream-JSON backend, the
+    `tm launch` / `tm connect` / agent-delegation launch line, the pane-relaunch
+    line, and `tm run`.
+  - `tm run` mattered most and was fixed last: it spawns `claude` with no tmux, so
+    Claude Code's `tmux show-environment -g` escape hatch returns false
+    immediately and the suppression fired there every time rather than depending
+    on the tmux server's environment.
--- a/crates/trusty-mpm/src/assets/skills/tm-capabilities.md
+++ b/crates/trusty-mpm/src/assets/skills/tm-capabilities.md
@@ -41,7 +41,7 @@ relevant `references/*.md` file here when you need an exact answer.
 | "What MCP tools exist and what parameters do they take?" | `references/mcp-tools.md` (33 tools, plus sibling-daemon pointers) |
 | "What agents can I delegate to, and what do they declare?" | `references/agents.md` (37 concrete agents) |
 | "What skills exist, and are they user-invocable?" | `references/skills.md` (52 bundled skills) |
-| "What does a `tm doctor` check name actually mean?" | `references/doctor.md` (23 checks) |
+| "What does a `tm doctor` check name actually mean?" | `references/doctor.md` (24 checks) |
 | "How do the pieces fit into an end-to-end flow?" | `references/workflows.md` (hand-authored, not generated) |
 
 ## Relationship to Other Skills
--- a/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/doctor.md
+++ b/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/doctor.md
@@ -2,30 +2,31 @@
 
 Generated from a maintained literal list cross-checked against `run_doctor`'s actual check names (see this module's `doctor_checks_match_run_doctor_names` test — an added, removed, or renamed check fails the test suite). Source: `crates/trusty-mpm/src/daemon/doctor.rs` and its five sibling `doctor_*.rs` files. Regenerate with `tm generate capabilities`.
 
-23 checks, in execution order.
+24 checks, in execution order.
 
 | # | Check | What it probes |
 |---|---|---|
 | 1 | `instructions` | Framework instructions deployed and non-empty for the target project. |
 | 2 | `agents` | Bundled agent roster deployed under the operator/workspace `.claude/agents/` tier. |
 | 3 | `agent_reachability` | Fails when bundled agents deploy into a settings tier a managed session's `--setting-sources` flag never loads — presence-only checks stay green while every delegation degrades to `general-purpose` (issue #4451). |
-| 4 | `skills` | Bundled skill catalog deployed under the operator/workspace `.claude/skills/` tier. |
-| 5 | `skill_source` | The framework's own skill source directory is present and readable. |
-| 6 | `output_style` | The `trusty-mpm` Claude Code output style is configured and its file exists (DOC-28 F4). |
-| 7 | `output_style_staleness` | Deployed output-style file content matches the bundled catalog, and no orphaned files linger under `output-styles/` (issue #2333). |
-| 8 | `output_style_legacy_ids` | Warns when a legacy/unresolvable `outputStyle` id lingers in a currently-shadowed settings layer (e.g. `settings.local.json`) even though the effective layer resolves fine (issue #3453). |
-| 9 | `deployment` | Full manifest-completeness diff of the deployed payload against the canonical bundled roster (issue #2158). |
-| 10 | `skill_staleness` | Deployed skill content matches the bundled/embedded source (issue #2876). |
-| 11 | `legacy_sources` | No legacy global instruction sources linger from a pre-migration install (issue #2876). |
-| 12 | `agent_skills` | Every agent's declared `skills:` frontmatter resolves to a real skill — dangling references fail (DOC-42, issue #2889). |
-| 13 | `agent_skills_prose_hints` | Informational: skill names mentioned in agent prose but not declared in `skills:` frontmatter (always `Ok`, issue #2906). |
-| 14 | `memory` | trusty-memory sidecar reachability probe (bounded by `PROBE_TIMEOUT`). |
-| 15 | `search` | trusty-search sidecar reachability + expected-index-present probe (bounded by `PROBE_TIMEOUT`). |
-| 16 | `worktrees` | No orphaned git worktrees under the managed workspace root (Fix 1b, #1840). |
-| 17 | `gh_account` | Active `gh` CLI identity is unambiguous — warns on multi-account ambiguity. |
-| 18 | `oauth_token` | Warns when a managed session risks the `CLAUDE_CONFIG_DIR`-keyed Keychain login loop (issue #2246). |
-| 19 | `hooks_contamination` | Warns when a project's `.claude/settings*.json` still carries tm hook entries from a pre-fix `tm install` — suggests `tm hooks clean` (issue #2940). |
-| 20 | `hooks_foreign_conflict` | Informational: warns when a project's `.claude/settings*.json` carries foreign (claude-mpm) hook entries that would fire inside a tm session — never auto-removed (issue #2940). |
-| 21 | `tcc_taint` | macOS: whether managed panes spawn `claude` with TCC responsibility disclaimed so its data-access prompts aren't attributed to the shared tmux server (issue #2997). |
-| 22 | `scaffold_tracking` | Warns when a harness-scaffolding path (`.claude/agents/`, `.claude/skills/`, `.claude/output-styles/`) is BOTH tracked in git AND regenerated locally by tm — the precondition for a `git merge --ff-only` "would be overwritten" collision; reports the exact true-intersection paths, never auto-modifies the index (issue #3427). |
-| 23 | `push_guard` | Warns when the project's clone carries no trusty-mpm cross-branch `pre-push` guard, or an older revision of it — the guard installs itself only on the clone path, so a base provisioned before it shipped is silently unprotected and a worktree tracking a foreign branch can force-push over that branch's reviewed lineage. Names the `tm repair push-guard` retrofit; doctor never writes into a repository (issue #2867). |
+| 4 | `transcript_saving` | Fails when a managed spawn would leave Claude Code transcript saving disabled — an inherited `CLAUDE_CODE_CHILD_SESSION` marker costs the session all native `--resume`/`--continue`/`/rewind` recovery, and also fails if the scrub would wrongly take `CLAUDE_CONFIG_DIR` (issue #4467). |
+| 5 | `skills` | Bundled skill catalog deployed under the operator/workspace `.claude/skills/` tier. |
+| 6 | `skill_source` | The framework's own skill source directory is present and readable. |
+| 7 | `output_style` | The `trusty-mpm` Claude Code output style is configured and its file exists (DOC-28 F4). |
+| 8 | `output_style_staleness` | Deployed output-style file content matches the bundled catalog, and no orphaned files linger under `output-styles/` (issue #2333). |
+| 9 | `output_style_legacy_ids` | Warns when a legacy/unresolvable `outputStyle` id lingers in a currently-shadowed settings layer (e.g. `settings.local.json`) even though the effective layer resolves fine (issue #3453). |
+| 10 | `deployment` | Full manifest-completeness diff of the deployed payload against the canonical bundled roster (issue #2158). |
+| 11 | `skill_staleness` | Deployed skill content matches the bundled/embedded source (issue #2876). |
+| 12 | `legacy_sources` | No legacy global instruction sources linger from a pre-migration install (issue #2876). |
+| 13 | `agent_skills` | Every agent's declared `skills:` frontmatter resolves to a real skill — dangling references fail (DOC-42, issue #2889). |
+| 14 | `agent_skills_prose_hints` | Informational: skill names mentioned in agent prose but not declared in `skills:` frontmatter (always `Ok`, issue #2906). |
+| 15 | `memory` | trusty-memory sidecar reachability probe (bounded by `PROBE_TIMEOUT`). |
+| 16 | `search` | trusty-search sidecar reachability + expected-index-present probe (bounded by `PROBE_TIMEOUT`). |
+| 17 | `worktrees` | No orphaned git worktrees under the managed workspace root (Fix 1b, #1840). |
+| 18 | `gh_account` | Active `gh` CLI identity is unambiguous — warns on multi-account ambiguity. |
+| 19 | `oauth_token` | Warns when a managed session risks the `CLAUDE_CONFIG_DIR`-keyed Keychain login loop (issue #2246). |
+| 20 | `hooks_contamination` | Warns when a project's `.claude/settings*.json` still carries tm hook entries from a pre-fix `tm install` — suggests `tm hooks clean` (issue #2940). |
+| 21 | `hooks_foreign_conflict` | Informational: warns when a project's `.claude/settings*.json` carries foreign (claude-mpm) hook entries that would fire inside a tm session — never auto-removed (issue #2940). |
+| 22 | `tcc_taint` | macOS: whether managed panes spawn `claude` with TCC responsibility disclaimed so its data-access prompts aren't attributed to the shared tmux server (issue #2997). |
+| 23 | `scaffold_tracking` | Warns when a harness-scaffolding path (`.claude/agents/`, `.claude/skills/`, `.claude/output-styles/`) is BOTH tracked in git AND regenerated locally by tm — the precondition for a `git merge --ff-only` "would be overwritten" collision; reports the exact true-intersection paths, never auto-modifies the index (issue #3427). |
+| 24 | `push_guard` | Warns when the project's clone carries no trusty-mpm cross-branch `pre-push` guard, or an older revision of it — the guard installs itself only on the clone path, so a base provisioned before it shipped is silently unprotected and a worktree tracking a foreign branch can force-push over that branch's reviewed lineage. Names the `tm repair push-guard` retrofit; doctor never writes into a repository (issue #2867). |
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
@@ -605,8 +605,18 @@ pub(crate) async fn run_inplace_relaunch(
 /// the tmux-pane spawn/resume paths so an in-place relaunch does not silently
 /// drop back into the `CLAUDE_CONFIG_DIR`-keyed Keychain login loop the token
 /// exists to bypass. Pure — builds and returns, never spawns.
+///
+/// Issue #4467: the inherited Claude Code session markers are scrubbed here too,
+/// via [`trusty_mpm::core::claude_env_scrub::scrub_command`]. This path execs
+/// `claude` through a `Command` with no shell in between, so the `env -u` prefix
+/// the tmux-pane paths use does not apply to it — but a bare `tm` relaunch run
+/// from inside a Claude Code session leaks exactly the same
+/// `CLAUDE_CODE_CHILD_SESSION` marker and would silently lose its transcript.
+/// The scrub runs BEFORE the deliberate assignments below so it can never
+/// clobber `CLAUDE_CONFIG_DIR` (#4455) even if the marker list grew wrongly.
 /// Test: `inplace_exec_command_forwards_every_arg_in_order`,
 /// `inplace_exec_command_scrubs_api_key_and_sets_auth_env`,
+/// `inplace_exec_command_scrubs_inherited_session_markers`,
 /// `inplace_exec_command_carries_isolation_flags_and_persona_end_to_end`.
 pub(crate) fn build_inplace_exec_command(
     resume: &trusty_mpm::runtime::InPlaceResumeCommand,
@@ -616,6 +626,9 @@ pub(crate) fn build_inplace_exec_command(
     cmd.args(&resume.args)
         .current_dir(cwd)
         .env_remove("ANTHROPIC_API_KEY");
+    // #4467: strip Claude Code's inherited process-local session markers so the
+    // relaunched session keeps native --resume/--continue/rewind recovery.
+    trusty_mpm::core::claude_env_scrub::scrub_command(&mut cmd);
     if let Some(dir) = &resume.config_dir {
         cmd.env("CLAUDE_CONFIG_DIR", dir);
     }
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace/tests.rs
@@ -766,6 +766,45 @@ fn inplace_exec_command_scrubs_api_key_and_sets_auth_env() {
     );
 }
 
+#[test]
+fn inplace_exec_command_scrubs_inherited_session_markers() {
+    // #4467: a bare `tm` relaunch run from inside a Claude Code session inherits
+    // CLAUDE_CODE_CHILD_SESSION and, without this scrub, the relaunched session
+    // saves no transcript — no --resume, no --continue, no /rewind. The marker
+    // name is hard-coded so this cannot pass vacuously if the shared marker list
+    // is emptied.
+    let resume = synthetic_resume(&["--dangerously-skip-permissions"]);
+    let cmd = build_inplace_exec_command(&resume, std::path::Path::new("/fake/cwd"));
+
+    let envs: Vec<(String, Option<String>)> = cmd
+        .get_envs()
+        .map(|(k, v)| {
+            (
+                k.to_string_lossy().into_owned(),
+                v.map(|v| v.to_string_lossy().into_owned()),
+            )
+        })
+        .collect();
+
+    assert!(
+        envs.contains(&("CLAUDE_CODE_CHILD_SESSION".to_owned(), None)),
+        "the transcript-suppressing marker must be REMOVED (None): {envs:?}"
+    );
+    assert!(
+        envs.contains(&("CLAUDE_CODE_SESSION_ID".to_owned(), None)),
+        "the parent's session id must be REMOVED (None): {envs:?}"
+    );
+    // Over-scrub guard: the scrub runs BEFORE the deliberate assignments, so
+    // CLAUDE_CONFIG_DIR must survive it as a SET value (#4455 / #4451).
+    assert!(
+        envs.contains(&(
+            "CLAUDE_CONFIG_DIR".to_owned(),
+            Some("/fake/config".to_owned())
+        )),
+        "CLAUDE_CONFIG_DIR must survive the scrub as a set value: {envs:?}"
+    );
+}
+
 #[serial_test::serial]
 #[test]
 fn inplace_exec_command_carries_isolation_flags_and_persona_end_to_end() {
--- a/crates/trusty-mpm/src/bin/tm/commands/session/start.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session/start.rs
@@ -205,10 +205,15 @@ async fn start_session_in_place(
             // #2997: disclaim the pane's `claude` off the shared tmux server
             // (same wrapper the daemon + `tm launch`/`connect` paths use).
             // No-op off macOS / under TM_DISABLE_SPAWN_DISCLAIM.
-            let claude_cmd = trusty_mpm::core::spawn_disclaim::disclaim_pane_command(&format!(
-                "claude {}",
-                trusty_mpm::core::model_inject::PERMISSION_MODE_FLAG
-            ));
+            // #4467: the launch line itself is built in the LIBRARY
+            // (`model_inject::build_inplace_session_command`) so it carries the
+            // shared inherited-marker scrub and is readable by the
+            // `transcript_saving` doctor check. It used to be hand-built here as
+            // `format!("claude {PERMISSION_MODE_FLAG}")` — a sixth interactive
+            // launch line that silently saved no transcript.
+            let claude_cmd = trusty_mpm::core::spawn_disclaim::disclaim_pane_command(
+                &trusty_mpm::core::model_inject::build_inplace_session_command(),
+            );
             let send = trusty_mpm::core::tmux::send_line(
                 None,
                 &trusty_mpm::core::tmux::TmuxTarget::session(&body.name),
--- a/crates/trusty-mpm/src/bin/tm/generate/doctor.rs
+++ b/crates/trusty-mpm/src/bin/tm/generate/doctor.rs
@@ -44,6 +44,10 @@ pub(crate) const DOCTOR_CHECKS: &[(&str, &str)] = &[
         "Fails when bundled agents deploy into a settings tier a managed session's `--setting-sources` flag never loads — presence-only checks stay green while every delegation degrades to `general-purpose` (issue #4451).",
     ),
     (
+        "transcript_saving",
+        "Fails when a managed spawn would leave Claude Code transcript saving disabled — an inherited `CLAUDE_CODE_CHILD_SESSION` marker costs the session all native `--resume`/`--continue`/`/rewind` recovery, and also fails if the scrub would wrongly take `CLAUDE_CONFIG_DIR` (issue #4467).",
+    ),
+    (
         "skills",
         "Bundled skill catalog deployed under the operator/workspace `.claude/skills/` tier.",
     ),
--- a/crates/trusty-mpm/src/client/executor/tests.rs
+++ b/crates/trusty-mpm/src/client/executor/tests.rs
@@ -196,7 +196,7 @@ async fn execute_doctor_against_test_daemon() {
     match executor.execute(TrustyCommand::Doctor).await {
         CommandResult::Doctor(report) => {
             // This test owns the EXECUTOR round-trip, not the check roster.
-            // The exact count is pinned by `run_doctor_produces_twenty_three_checks`
+            // The exact count is pinned by `run_doctor_produces_twenty_four_checks`
             // and `doctor_endpoint_returns_report`, both of which derive it from
             // their own name list — duplicating a bare literal here just made it a
             // fourth place to forget (#4090 review LOW-1).
--- a/crates/trusty-mpm/src/client/http_client/session_connect.rs
+++ b/crates/trusty-mpm/src/client/http_client/session_connect.rs
@@ -85,11 +85,15 @@ impl DaemonClient {
                 "trusty-mpm-system-prompt-{}.txt",
                 uuid::Uuid::new_v4()
             ));
+            // #4467: the line is built in `core::model_inject` so it carries the
+            // shared inherited-marker scrub and the `transcript_saving` doctor
+            // check can read it. Hand-building it here is what left this launch
+            // path silently saving no transcript.
             match std::fs::write(&path, &prompt) {
-                Ok(()) => format!("claude --append-system-prompt-file {}", path.display()),
+                Ok(()) => crate::core::model_inject::build_client_session_command(Some(&path)),
                 Err(err) => {
                     tracing::warn!(%err, "failed to write system prompt file; launching bare claude");
-                    "claude".to_string()
+                    crate::core::model_inject::build_client_session_command(None)
                 }
             }
         };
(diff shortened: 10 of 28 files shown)


bobmatnyc:
## Code Critic Verdict: WARN

One HIGH cluster, zero CRITICAL. The five spawn paths named in #4467 are correctly fixed and `CLAUDE_CONFIG_DIR` survives on every one of them — verified independently, including by executing the emitted command. The HIGH is coverage: two other `tm` launch-line builders spawn an interactive `claude` and still leak the markers, and the new `transcript_saving` check reports `Ok` for both.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/standalone/run.rs` | 56-58 | `build_launch_command` (`tm run <alias>`, attended interactive) spawns `claude` via `Command` with **no tmux involved**, so upstream's escape hatch cannot apply (`Esg()`: `if(!Z.TMUX)return!1`). This is the one `tm` path where the suppression fires deterministically, and it is untouched. | `trusty_mpm::core::claude_env_scrub::scrub_command(&mut cmd);` immediately before the `CLAUDE_CONFIG_DIR` assignment on line 58. |
| HIGH | `crates/trusty-mpm/src/core/model_inject.rs` | 208 | `build_claude_command` — the launch line for `tm launch` (`launch.rs:308`), `tm connect` (`launch.rs:563` via `connect_claude_cmd:600`) and agent delegations (`build_agent_command:233`) — emits a bare `claude …` with no `env` prefix at all. Its own doc claims it is where "every spawn path stays unattended and isolated". | Prefix it with `claude_env_scrub::env_unset_flags()`, or route those callers through `env_bin_prefix`. |
| HIGH | `crates/trusty-mpm/src/daemon/doctor_transcript_saving.rs` | 58, 122-123 | The check parses only `env_bin_prefix`, so it returns `Ok` while the two builders above leak. Its message vouches for more than it read: "a managed spawn unsets […] so transcript saving and native --resume stay available". | Parse every launch-line builder, or narrow the message to name the path it actually covers. |
| MEDIUM | `crates/trusty-mpm/src/core/claude_env_scrub_tests.rs` | 69-82, 103-117 | `CLAUDE_PID` and `CLAUDE_EFFORT` are pinned by nothing. Every test touching them iterates `INHERITED_SESSION_MARKERS`, so both go vacuous if an entry is deleted; `claude_code_tests.rs:352-380` hard-codes only `CLAUDE_CODE_CHILD_SESSION`/`CLAUDE_CODE_SESSION_ID`/`CLAUDECODE`, and `:665` only `CLAUDE_CODE_EXECPATH`. Deleting `CLAUDE_PID` from the list keeps the whole suite green. | Add a literal-name assertion for all six, the way the other four already have. |
| LOW | `crates/trusty-mpm/src/runtime/claude_code_tests.rs` | 665 | `spawn_command_uses_resolved_binary` now matches `"-u CLAUDE_CODE_EXECPATH /Users/me/.local/bin/claude "`, coupling a binary-resolution test to `CLAUDE_CODE_EXECPATH` being *last* in the marker list. Reordering the list fails this test with an unrelated message. | Match on `" /Users/me/.local/bin/claude "` alone, or build the expectation from `env_unset_flags()`. |
| LOW | `crates/trusty-mpm/src/core/claude_env_scrub.rs` | 181 | `parse_env_unset_vars` does not model POSIX option termination — `claude_env_scrub_tests.rs:176` pins that it reports `FOO` unset from `env CLAUDE_CODE_CHILD_SESSION=1 -u FOO …`. The doctor check therefore passes on an ordering that kills every spawn with `env: -u: No such file or directory`. Bounded: `env_bin_prefix_orders_scrub_flags_before_assignments` covers ordering, and that failure is loud. | Stop collecting at the first `NAME=VALUE` token. |
| LOW | `crates/trusty-mpm/src/daemon/doctor_transcript_saving.rs` | 57-58, 79 | The over-scrub branch tests the literal `"CLAUDE_CONFIG_DIR"` and passes `None` for the token, so an over-scrub of `CLAUDE_CODE_OAUTH_TOKEN` — the other `DELIBERATE_SPAWN_ENV` member — is invisible to it. | Iterate `DELIBERATE_SPAWN_ENV`; pass `Some("probe-token")`. |
| LOW | `crates/trusty-mpm/src/core/claude_env_scrub.rs` | 81-100 | `CLAUDE_CODE_INVOKED_SKILLS` sits on the same upstream non-propagation list used to justify including `CLAUDE_CODE_EXECPATH` (`["SHELL","GIT_EDITOR","CLAUDECODE","AI_AGENT","CLAUDE_CODE_SESSION_ID","CLAUDE_CODE_CHILD_SESSION","CLAUDE_PID","TRACEPARENT","CLAUDE_CODE_EXECPATH",…,"CLAUDE_EFFORT","CLAUDE_CODE_INVOKED_SKILLS"]`) but is not discussed anywhere. Not observed set. | Add it, or record the decision next to the `TRACEPARENT` note. |

### Required changes

1. Scrub the markers in `standalone::run::build_launch_command` and `model_inject::build_claude_command`, so `tm run` / `tm launch` / `tm connect` get the same treatment as the daemon paths. One line and one prefix.
2. Either extend the `transcript_saving` check to those builders or narrow its message so it does not vouch for paths it never read.
3. Pin `CLAUDE_PID` and `CLAUDE_EFFORT` with literal assertions.

### What I verified independently

**`CLAUDE_CONFIG_DIR` survives, per path.** No path can unset or shadow it.

| Path | Mechanism | Result |
|---|---|---|
| `spawn_command` (`runtime/claude_code.rs:308`) | `env_bin_prefix`: `format!("env -u ANTHROPIC_API_KEY{scrub}{assignments} {claude_bin}")` — `scrub` interpolated before `assignments` | survives |
| `resume_command` (`:424`) | same `env_bin_prefix` | survives |
| `compose_inplace_args` / `build_inplace_exec_command` (`guided_inplace.rs:631`) | `scrub_command` on 631, `env("CLAUDE_CONFIG_DIR", …)` on 633 — a later `env` wins over an earlier `env_remove`, and the name is not on the list regardless | survives |
| `build_claude_command` (`control/backend/stream_json.rs:91`) | `scrub_command`; this path never sets `CLAUDE_CONFIG_DIR` at all | nothing to clobber |
| shared constant (`claude_env_scrub.rs:81`) | `INHERITED_SESSION_MARKERS` ∩ `DELIBERATE_SPAWN_ENV` = ∅, enforced by two tests, one of which survives an emptied `DELIBERATE_SPAWN_ENV` | survives |

Not taken on trust — I ran the emitted shape through a real shell with all six markers set:

```
CFG=[/Users/John Doe/.trusty-tools/trusty-mpm/claude-config]
OAUTH=[sk-ant-oat01-x]
SID=[sess-1]
CLAUDE_CODE_CHILD_SESSION=UNSET   CLAUDECODE=UNSET   CLAUDE_PID=UNSET
CLAUDE_EFFORT=UNSET   CLAUDE_CODE_SESSION_ID=UNSET   CLAUDE_CODE_EXECPATH=UNSET
```

and confirmed the ordering hazard is real, not theoretical: `env CLAUDE_CONFIG_DIR=/tmp/x -u CLAUDECODE /bin/echo hi` → `env: -u: No such file or directory`, exit 127.

**In-repo consumers of the scrubbed variables: none.** A workspace-wide grep (`crates/`, `assets/`, `scripts/`, `docs/`, bundled hooks and skills, all file types) for `CLAUDECODE`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_PID`, `CLAUDE_EFFORT`, `CLAUDE_CODE_EXECPATH`, `CLAUDE_CODE_CHILD_SESSION` returns only this PR's own code, tests, doc comments and CHANGELOG. The only `CLAUDE_*` reads anywhere in the workspace are `CLAUDE_CODE_OAUTH_TOKEN` (kept), `CLAUDE_CONFIG_DIR` (`core/delegation_authority.rs:297`, kept), and `CLAUDE_SESSION_ID`/`CLAUDE_HOOK_EVENT` (`bin/tm/commands/misc.rs:614,621` — different names, not on the list). Nothing keys off the scrubbed set, so no hook, statusline, shell snapshot or `pm_guard` changes behavior. Third-party plugin skills in the deployed config recommend `unset CLAUDECODE` explicitly, so the direction is the conventional one.

**`CLAUDE_CODE_ENTRYPOINT` keep: correct, and scrubbing would be worse.** The bundle only ever *reads* it (`dH()`, plus `??"cli"` defaults); the sole assignment is the SDK setting `sdk-ts` when unset. Observed value in a live delegated process is `cli`. Unset is not universally equivalent to `"cli"` — `s1()` gates on `e==="cli"||e==="remote"`, so unsetting would silently disable that path. The claim is environment-dependent rather than proven-safe, though: a *non-`cli`* inherited value does change behavior — `sdk-ts`/`sdk-py`/`sdk-cli` drops one built-in agent from the roster (`if(entrypoint!=="sdk-ts"&&!=="sdk-py"&&!=="sdk-cli")e.push(qqu)`), and `claude-vscode` changes `clientType`. Not a defect at the observed value; worth a sentence in the doc.

**`AI_AGENT` keep: correct, with a stronger reason than the PR gives.** Upstream self-heals it — `BNn()` overwrites any inherited value starting `claude-code_` or `claude-code/` with its own at startup, so a leaked `AI_AGENT` cannot reach the spawned session stale. `TRACEPARENT` is only set by `KDt()` when tracing is on and its effect is span parenting, so leaving it is defensible.

**The doctor check is not tautological.** I reproduced the mutation probe myself: dropping `{scrub}` from `env_bin_prefix` made `production_spawn_preserves_transcript_saving` fail with the actionable message (`Currently unsets: [ANTHROPIC_API_KEY]`) while the other six stayed green. Reverted; tree clean.

**The four updated assertions are not weakened.** `spawn_command_sets_claude_config_dir`, `resume_command_sets_claude_config_dir`, `spawn_command_sets_both_config_dir_and_oauth_token` and `env_bin_prefix_quotes_config_dir_with_space` split one substring into two because the adjacency they pinned is now false; the positional invariant they gave up is re-asserted by `env_bin_prefix_orders_scrub_flags_before_assignments` (`last_unset < first_assignment`). Both renamed tests keep a full-string `assert_eq!`. The one caveat is the MEDIUM above: the `scrub` segment inside those pins is interpolated from production, so it pins position, not content — and the content is only partly pinned elsewhere.

**Gates, run locally on `f7c9e159`:** `cargo test -p trusty-mpm --no-fail-fast` → exit 0, `4059 passed; 0 failed` (lib) + `1320 passed` ×2 (bin) + every integration target `ok`. `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → exit 0. `cargo fmt --check` → exit 0. `check_line_cap.sh` → `7 allowlisted, 0 violations — OK`. `check_capabilities.sh` → `up to date (6 generated files)`. `check_test_pointers.sh` + `check_sld.sh` + `check_doc_numbers.sh` → exit 0. CI is 23/23 green including the doc-comment pointer lint. CHANGELOG entry is under `## [Unreleased]` → `### Fixed`, and every `// #4467:` pointer and doc `Test:` line resolves.

One flake: the first full run failed `client::executor::tests::execute_doctor_against_test_daemon` with `daemon unreachable` on an ephemeral port. It passed in isolation and on the second full run. Transport failure, unrelated to this diff — noting it rather than burying it.

### Note on the premise — where the residual risk actually is

This does not make the change wrong, but it relocates the risk and corrects one sentence in the PR body. I tested tmux's propagation rules directly, on an isolated socket, both directions:

- Server started clean, session created by a client that **has** the marker → not in the session env, not in the pane. `update-environment` does not include `CLAUDE_*`.
- Server started **with** the marker → global env has it, panes see it, and `env -u` in the sent line clears it.

So for a pane the marker's presence and the escape hatch are coupled: marker in pane ⟹ marker in tmux global ⟹ `tmux show-environment -g CLAUDE_CODE_CHILD_SESSION` prints a matching line ⟹ `Tsg()` true ⟹ no suppression. "A tmux server restarted from a clean shell would silently stop saving" is the wrong way round — a clean server also means the panes never inherit the marker.

The defect is still real on the fixed paths, for two reasons the PR does not name, and both argue *for* this change: `Esg()` calls `spawnSync("tmux", …)` by bare name with a 250 ms timeout and returns `false` on throw or non-zero status. A pane on the daemon's minimal launchd `PATH` (the documented #1298 hazard, `spawn_disclaim/pane.rs:83`) or a loaded machine blowing 250 ms both make the escape hatch fail open into suppression. That is a better account of "it looked intermittent" than the tmux-server story.

It also sharpens the HIGH: `tm run` has no tmux at all, so `Tsg()` returns `false` at its first line and the suppression fires every time — the one `tm` path where this is deterministic rather than probe-dependent, and the path this PR does not touch.

Substitute evidence for the un-driven interactive case is sufficient, and I checked the decompiled gate against the bundle myself rather than accepting the quote: `Zkt()` (`isNestedInteractiveClaudeSession`) matches the PR body verbatim, and `DDc` requires a line starting `CLAUDE_CODE_CHILD_SESSION=`, so a tmux-*unset* variable (printed as `-CLAUDE_CODE_CHILD_SESSION`) does not trip the hatch.

Per the brief, retention/pruning was out of scope and I did not re-litigate it. `daemon/spawn_command.rs:62` (`relaunch_command` → bare `"claude"`) is a pre-existing gap already documented in that module and tracked as #2020 — not flagged, but it is a third uncovered launch line if the sweep in required change 1 is done properly.

bobmatnyc:
## Code Critic Verdict (round 2): WARN

One HIGH, zero CRITICAL. **The `env` prefix is safe on every launch path** — I verified that by execution, not by reading the argument, and every round-1 required change landed correctly. The HIGH is the same shape as round 1's, one level out: a **sixth** launch line spawns an interactive `claude` with no scrub, and the doctor check's "here is the one gap" disclosure does not name it.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/session/start.rs` | 208-211 | `start_session_in_place` hand-builds `format!("claude {}", PERMISSION_MODE_FLAG)` and sends it to a fresh tmux pane. It never goes through `build_claude_command`, has no scrub, no test, and is not named in the doctor's gap disclosure. Reachable: `tm session start` in a directory that is not a GitHub-backed git repo (no `.git`, or a git repo with no parseable remote) — the documented, supported "just run claude here" case. Those sessions inherit `CLAUDE_CODE_CHILD_SESSION` and silently save no transcript, exactly the defect this PR closes elsewhere. | Route it through `model_inject::build_claude_command(None, None)` (it already produces `--setting-sources` + `PERMISSION_MODE_FLAG`), or at minimum prefix it with `claude_env_scrub::env_unset_flags()`. Then add it to the doctor's gap sentence or move the builder into the library so `launch_lines()` can read it. |
| MEDIUM | `crates/trusty-mpm/src/daemon/doctor_transcript_saving.rs` | 90-131, 193-199 | `launch_lines()` is a hand-maintained enumeration and `launch_lines_covers_every_builder` only asserts the five listed labels are *present*. Nothing fails when a launch line exists that is not on the list — which is precisely how the finding above survived two review rounds and a widening pass. The `Ok` message compounds it by naming exactly one binary-crate gap ("the bare-`tm` in-place relaunch … is covered by its own test"), which reads as *the* gap rather than *a* gap. | Add a source-scanning test that walks `crates/trusty-mpm/src` for `Command::new("claude")` and for string literals beginning `"claude ` / `"env … claude`, and asserts every hit is on an explicit allowlist. New spawn sites then fail loudly instead of silently. Separately, make the `Ok` message enumerate the binary-crate builders by name rather than describing one. |
| LOW | `crates/trusty-mpm/src/core/model_inject.rs` | 200-206 | The doc and the doctor label (`doctor_transcript_saving.rs:115`) both advertise this builder as the **agent-delegation** launch line via `build_agent_command`. `build_agent_command` has no production caller anywhere in the workspace — `grep -rn build_agent_command crates --include='*.rs'` returns its definition (`model_inject.rs:264`) and its own test (`:451`), nothing else. An operator chasing a `transcript_saving` failure labelled `(tm launch / connect / delegate)` will look for a delegation path that does not exist. | Drop `delegate` from the label and the doc, or delete the dead `build_agent_command`. |
| LOW | `crates/trusty-mpm/src/core/claude_env_scrub.rs` | 220-236 | `parse_env_unset_vars` models POSIX option termination at the first `NAME=VALUE` (good — that was round 1's LOW and it is fixed), but it keeps scanning **past the command token**. On `model_inject`'s output there are no assignments at all, so the loop runs to the end of the line; a future `claude` flag pair spelled `-u <x>` would be counted as an unset variable. Not reachable today. | Also stop at the first token that is neither `-u`, a `-u` operand, nor another recognised `env` option — i.e. stop at the command word. |

### Required changes

1. Scrub the markers on `session/start.rs`'s in-place launch line (finding 1). This is the one that costs real transcripts.
2. Make new spawn sites fail the build rather than being silently uncovered (finding 2), and fix the `Ok` message so it does not imply the binary-crate gap is singular.

### Is the `env` prefix safe? Yes — verified by execution

Claim 5 was the highest-risk item in the PR, so I tested it rather than reading the reasoning.

**`env` execs in place; the process tree `core::process` walks is unchanged.** Ran the real built shim:

```
$ tm internal-spawn-disclaimed env -u CLAUDE_CODE_CHILD_SESSION <fake claude> 6 &
shim  pid=8887 ppid=8844 comm=…/target/debug/tm
pgrep -P 8887 → 8914   comm=…/fakebin2/claude    grandchildren: []
```

One hop, not two. `claude_child_of`'s single hop through `is_disclaim_wrapper` still reaches it, and `process_name_contains_claude` matches the post-exec `comm`.

**The disclaim attribute lands on the process that becomes `claude`.** I compiled a C shim that reproduces `spawn_disclaim::macos::status` exactly (`responsibility_spawnattrs_setdisclaim` via `dlsym`, `POSIX_SPAWN_SETSIGDEF` on SIGINT/SIGQUIT, `posix_spawnp`) and a child that reads `responsibility_get_pid_responsible_for_pid(getpid())`:

```
A  shim → whoresp directly, disclaimed:   spawned pid=43489 → CHILD pid=43489 responsible=43489
B  shim → env -u … whoresp, disclaimed:   spawned pid=43507 → CHILD pid=43507 responsible=43507   ← same pid through env
```

Same PID before and after the exec, so the per-proc responsibility attribute set at `posix_spawn` is on the same `proc` that ends up running `claude`. **Honest caveat:** my control arm (`--nodisclaim`) also reported `self_responsible=YES`, because in a plain CLI context nothing is inheriting responsibility — so this experiment proves the *exec-in-place* half conclusively and cannot discriminate disclaim-vs-no-disclaim on this machine. The exec-in-place half is the part the claim actually rests on.

**`env` is not a new failure mode.** `/usr/bin/env` only, not a builtin in sh/bash/zsh; the shim uses `posix_spawnp` (`macos/status.rs:173`), so PATH resolution works; `env` is strictly more PATH-resolvable than the bare `claude` this line used to emit. BSD `env -u NEVER_SET_VAR` on an *unset* variable exits 0. The POSIX ordering trap is real and confirmed (`env FOO=1 -u BAR /bin/echo …` → `env: -u: No such file or directory`, exit 127), and every emitted prefix puts the flags first.

**clap parses the `-u` tokens.** Through the real binary, with all six markers set in the parent:

```
$ tm internal-spawn-disclaimed env -u CLAUDE_CODE_CHILD_SESSION … -u CLAUDE_CODE_EXECPATH /bin/sh -c '…'
CLAUDE_CODE_CHILD_SESSION=[]  CLAUDE_CODE_SESSION_ID=[]  CLAUDECODE=[]
CLAUDE_PID=[]  CLAUDE_EFFORT=[]  CLAUDE_CODE_EXECPATH=[]
CLAUDE_CONFIG_DIR=[/managed/cfg]  CLAUDE_CODE_OAUTH_TOKEN=[sk-ant-oat01-KEEPME]   exit=0
```

**All four emitted shapes executed end-to-end** against a fake `claude` on PATH, with every marker set in the parent — `tm launch`/`connect` (shim outside, `env` as program), the same line with the disclaim disabled, the daemon shape (`env` outside, shim + absolute binary inside, `CLAUDE_CONFIG_DIR='/Users/John Doe/…'` with a space), and the pane relaunch. Every one: six markers empty, `CLAUDE_CONFIG_DIR` and `CLAUDE_CODE_OAUTH_TOKEN` intact, exit 0.

### Over-scrub: clean

- No in-repo consumer of any scrubbed variable. A whole-tree grep (`--exclude-dir=target`, all file types, assets and skills included) for `CLAUDECODE`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_PID`, `CLAUDE_EFFORT`, `CLAUDE_CODE_EXECPATH`, `CLAUDE_CODE_CHILD_SESSION` returns only this PR's own code, tests and docs.
- `DELIBERATE_SPAWN_ENV` survives on every path, proven by execution above, not only by `assert!`.
- Neither export prefix collides with the scrub list: `session_id_export_prefix` emits `TM_MANAGED_SESSION_ID` (`claude_code.rs:68`), `gh_env_source_prefix` sources a file (`claude_code_gh_env.rs:104`).
- `removed_env_keys` cannot see an over-scrub of `CLAUDE_CONFIG_DIR` on the two `Command` paths, because the later `.env()` overwrites the earlier `env_remove` — but that is also why there is no over-scrub to see there. Correct, not a hole.
- `over_scrub_of_the_oauth_token_is_detected` is a `verdict`-level test, so on its own it is not production evidence. The production-level guard is real: `probe_supplies_an_oauth_token_so_over_scrub_is_visible` pins that the probe prefix actually carries `CLAUDE_CODE_OAUTH_TOKEN=`, and adding the token to `INHERITED_SESSION_MARKERS` fails `every_production_launch_line_preserves_transcript_saving`.

### The doctor check is not tautological — six mutations, mine not the author's

Each applied alone, tests run, file byte-restored (`cmp -s`), tree confirmed clean at the end.

| # | Mutation | Result |
|---|---|---|
| M1 | `model_inject::build_claude_command` → bare `"claude"` | FAIL, names `` `core::model_inject::build_claude_command (tm launch / connect / delegate)` `` — `Currently unsets: []` (9 tests) |
| M2 | drop `scrub_command` from `standalone::run::build_launch_command` | FAIL, names `` (tm run) `` (3 tests) |
| M3 | `relaunch_command` → bare `"claude"` | FAIL, names `` (pane relaunch) `` (3 tests + 1 load flake, below) |
| M4 | **partial scrub** — delete only `TRANSCRIPT_SUPPRESSING_MARKER` from the list, keep the other five | FAIL, 14 tests, incl. `every_marker_is_pinned_by_literal_name` and `Currently unsets: [ANTHROPIC_API_KEY, CLAUDE_CODE_SESSION_ID, CLAUDECODE, CLAUDE_PID, CLAUDE_EFFORT, CLAUDE_CODE_EXECPATH]` |
| M5 | **right markers, wrong position** — `{assignments}{scrub}` in `env_bin_prefix` (POSIX-invalid) | FAIL, 8 tests, incl. `env_bin_prefix_orders_scrub_flags_before_assignments`; the doctor reports `Currently unsets: [ANTHROPIC_API_KEY]` because the parser correctly stops at the assignment |
| M6 | drop `scrub_command` from `stream_json::build_claude_command` | FAIL, names `(headless)` (3 tests) |

M4 and M5 are the two the author did not run, and both are the interesting ones: a *partial* scrub and a *correctly-listed but wrongly-positioned* scrub are both caught, and M5 shows the parser fix from round 1 is what makes the ordering failure visible rather than passing on a prefix that would exit 127.

Claim 4 confirmed independently: M4 proves `EXPECTED_MARKERS` is a real literal pin, not a re-derivation.

### Gates

- `cargo test -p trusty-mpm` on `cd28d073` → **exit 0**, 44 targets, **6988 passed; 0 failed; 16 ignored**. Lib target `4075 passed; 0 failed; 6 ignored`. Matches the PR body.
- CI: 23 pass / 1 skipping on the last computed merge ref.
- One load flake, named not buried: `core::session_launch::tests::prepare_session_manifest_filters_agent_set` failed once during M3 (unrelated to that mutation — M3 touches `daemon/spawn_command.rs`). Passes 3/3 in isolation; the machine is running several sibling agents. Not this PR's.

### Notes

- **The PR is `CONFLICTING` / `DIRTY`,** which is why there are no checks on the current head. #4477 landed on `main` and replaced shared-`CHANGELOG.md` per-PR entries with per-PR fragments; this branch still adds 31 lines to `crates/trusty-mpm/CHANGELOG.md`. Rebase and convert to a fragment — the conflict is the changelog convention change, not the code.
- Not flagged, stated for the record: with the disclaim disabled (non-macOS, or `TM_DISABLE_SPAWN_DISCLAIM`), the pane line changes from `claude …` — a shell command *word*, so an operator's `claude` alias or shell function would expand — to `env … claude …`, where it will not. On macOS with the disclaim on, aliases already did not apply, so this only moves the two surfaces into agreement. I judge this desirable rather than a defect and did not file it.
- I did not re-litigate retention/pruning, the `CLAUDE_CODE_ENTRYPOINT` / `AI_AGENT` / `TRACEPARENT` keeps, or the tmux-escape-hatch premise — round 1 settled those and nothing in this diff reopens them.

PR #4477, opened 2026-07-31
feat: replace shared-CHANGELOG per-PR entries with per-PR changelog fragments

bobmatnyc:
## The problem

The per-PR changelog rule pointed every concurrent PR at the same lines of the
same file — the topmost `## [Unreleased]` heading in `crates/<crate>/CHANGELOG.md`.
That is a guaranteed textual conflict, not a hazard. On 2026-07-31 five concurrent
trusty-mpm PRs — [#4463](https://github.com/bobmatnyc/trusty-tools/pull/4463),
[#4464](https://github.com/bobmatnyc/trusty-tools/pull/4464),
[#4465](https://github.com/bobmatnyc/trusty-tools/pull/4465),
[#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466),
[#4475](https://github.com/bobmatnyc/trusty-tools/pull/4475) — each added a bullet
there, and every merge forced the next PR to rebase and hand-resolve the section.
[#4399](https://github.com/bobmatnyc/trusty-tools/issues/4399) burned three such
rounds.

## The design

Entries move to per-PR **fragment** files. Release time assembles them.

| Decision | Choice | Why |
|---|---|---|
| Location | `crates/<crate>/changelog.d/` | Per-crate, because each crate owns its own `CHANGELOG.md`. Sits next to the file it feeds; `changelog.d` is the established convention for this pattern. |
| Filename | `<issue-or-pr-number>-<slug>.md` | The number is what makes collisions impossible — GitHub issue/PR numbers are unique per repo, so two concurrent PRs add two differently named files and git never sees a conflict. The slug keeps two fragments for one number distinct and makes the diff readable. |
| Format | line 1 = category, rest = bullet text verbatim | Two lines of schema. The category has to live *somewhere*; putting it in the file (not the filename) means recategorising is an edit, not a rename, and keeps the filename purely about identity. Bullets are copied through byte-for-byte, including indented sub-bullets. |
| Categories | `Breaking` `Added` `Fixed` `Performance` `Changed` `Documentation` | Exactly the groups `cliff.toml` already emits and the existing changelogs already use, in the same order — an assembled section is indistinguishable in shape from the git-cliff history above it. Matched case-insensitively. |

```
crates/trusty-mpm/changelog.d/4476-changelog-fragments.md

Changed

- per-PR changelog entries are now fragment files, not edits to a shared section
  - indented sub-bullets survive verbatim
```

[...]

Issue #4476, opened 2026-07-31
Replace shared-CHANGELOG per-PR entries with per-PR changelog fragments assembled at release


bobmatnyc:
## Defect

The per-PR changelog rule makes every concurrent PR write to the **same lines** of
the same file: the topmost `## [Unreleased]` heading in `crates/<crate>/CHANGELOG.md`.
That is a guaranteed textual conflict, not a hazard.

Evidence, 2026-07-31: five concurrent PRs against `crates/trusty-mpm` —
[#4463](https://github.com/bobmatnyc/trusty-tools/pull/4463),
[#4464](https://github.com/bobmatnyc/trusty-tools/pull/4464),
[#4465](https://github.com/bobmatnyc/trusty-tools/pull/4465),
[#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466),
[#4475](https://github.com/bobmatnyc/trusty-tools/pull/4475) — each added a bullet
under `## [Unreleased]` in `crates/trusty-mpm/CHANGELOG.md`. Every merge forced the
next PR to rebase and hand-resolve that section.
[#4399](https://github.com/bobmatnyc/trusty-tools/issues/4399) previously burned
three CHANGELOG rebase rounds on the same collision.

## Resolution

Replace the shared-section convention with per-PR **changelog fragment** files,
assembled into `CHANGELOG.md` at release time.

- Fragments live at `crates/<crate>/changelog.d/<issue-or-pr-number>-<slug>.md`.
  Distinct filenames per PR means git never sees a conflict.
- Fragment content: a one-line category (`Added`/`Fixed`/`Changed`/`Performance`/
  `Documentation`/`Breaking`) then the bullet(s) verbatim.
- `scripts/assemble-changelog.sh <crate-dir> <version>` groups the crate's
  fragments by category, writes them under the new version heading, and deletes
  the consumed fragments. `scripts/bump-version.sh` calls it.
- Fragments become the source of truth for the unreleased set. `CHANGELOG.md` no
  longer carries an `## [Unreleased]` section between releases.
- A CI gate fails a PR that touches `crates/<crate>/src/**` without recording the
  change. Today the rule is **review-gate-only** — no workflow enforces it.

## git-cliff supersession

`scripts/generate-changelog.sh` runs `git cliff --unreleased --prepend`, which
blindly stacks a fresh `## [Unreleased]` on top of the hand-written one. That is
[...]

Issue #3464, opened 2026-07-20
bug(trusty-agents): test-isolation defect — OnceLock dotenv load_env_local_once pollution across concurrent tests


bobmatnyc:
## Summary

`crates/trusty-agents/src/llm/credentials.rs::llm::credentials::tests::other_configured_providers_empty_when_nothing_else_configured` (line 547) fails nondeterministically when run in a worktree under `.base/.worktrees/*/` as part of `cargo test --workspace`.

```
assertion `left == right` failed
  left: ["fireworks", "atlascloud"]
 right: []
test result: FAILED. 1991 passed; 1 failed; 8 ignored
```

## Root Cause (Verified)

1. **The dotenv loader is a process-wide OnceLock**: `trusty_common::inference::credentials::dotenv::load_env_local_once()` is a `OnceLock` that runs exactly once per test binary. The first test to touch it wins — its environment decision is locked in for every test after it.

2. **Upward climb from cwd defeats test isolation**: The loader climbs the directory tree from the current working directory looking for `.env.local`. Running from a nested worktree (e.g., `.base/.worktrees/<name>/...`), the .git detection fails because in a worktree, `.git` is a FILE (a gitdir pointer), not a directory. The climb does not stop at the worktree root and continues to the project root, loading `~/trusty-mpm-projects/bobmatnyc/trusty-tools/.env.local`.

3. **Test isolation is defeated by cwd, not by $HOME override**: The test defends itself by clearing process env and overriding `$HOME` to a tempdir — but it never controls `cwd`, so the project-scoped climb bypasses the `$HOME` override entirely.

4. **Pollution is permanent for the test binary**: Because the loader uses `OnceLock`, once one test loads the real `.env.local`, every subsequent test in that binary operates in the polluted environment, regardless of per-test `$HOME` handling.

## Why This Matters (Local-Only, But Real)

- **Local-developer-only failure**: `.env.local` is gitignored and never exists on CI runners, so this cannot turn CI red. It is strictly a local developer environment failure.
- **Silent test-isolation defeat**: It makes `cargo test --workspace` unreliable for anyone with dev credentials configured (`fireworks`, `atlascloud`, etc.). The test's deliberate isolation measures are silently defeated.
- **Blocked releases today**: This defect forced explicit waiver and root-cause documentation during release gate for trusty-search 0.36.0 (2026-07-20).

## Shape vs #3434 (Shared Pattern)

[...]

PR #4482, opened 2026-07-31
fix: make five non-hermetic test fixtures deterministic (#4306, #4413, #4414, #4415, refs #4407)

### LINKED ISSUE #4306: test: two load-sensitive fixtures redden the shared gate (telegram backoff wall-clock, dead_loopback_url port reuse)
Parent: #3451 (hermetic test isolation). Verified against `origin/main` @ `aed80140`.

## Defect

Two `trusty-mpm` tests fail under parallel load and pass in isolation. Neither is caused by the code under test in any PR that has hit them. The cost is not the failures — it is that a gate which reddens for reasons unrelated to the change trains people to re-run until green, which is how a real regression gets buried.

**1. `telegram::supervisor::tests::healthy_run_resets_transient_backoff`** — `crates/trusty-mpm/src/telegram/supervisor_tests.rs:108`

- Wall-clock assertion at `supervisor_tests.rs:172`: `backoff_component < Duration::from_millis(20)`, where `backoff_component` is `starts[4] - starts[3] - healthy_run` (`:170`) — a measured elapsed real-time gap between two task wakeups.
- Observed **34.4ms** under full-suite parallel load; separately **21.79ms**. Both exceed the 20ms bound; neither implies the backoff failed to reset.
- The bound is 20ms against a policy `base` of 2ms (`:118`). It measures scheduler jitter, not backoff state.

**2. `tui::project_ctl::poll::tests::classify_connect_error_is_transport`** — `crates/trusty-mpm/src/tui/project_ctl/poll/tests.rs:205` (the substantive one)

- Fixture `dead_loopback_url()` at `poll/tests.rs:54-59` obtains a "known dead" address by **binding an ephemeral port and dropping it**: `TcpListener::bind("127.0.0.1:0")` (`:56`) → read port → `drop(listener)` (`:58`).
- The released port returns to the OS ephemeral pool. Under port churn another binder can be handed that exact port before the test connects. The connect then succeeds and the fixture's premise is void: the request either returns `Ok` (the `expect_err` at `:210` panics) or fails at the HTTP layer, classifying `NonTransport` instead of `Transport` (`:212`). Probability scales with parallel port churn.
- Blast radius is wider than this one test: the same bind-and-drop fixture is **triplicated** — `poll/tests.rs:54`, `tui/coordinator/tests.rs:811`, `tui/project_ctl/tests.rs:179` — across **13 call sites**. Only (2) has been observed failing; the other 12 share the invalid premise.

## Evidence

- (1) observed failing in at least three separate full-suite runs during PR #4294 and #4303 review; each diff touched zero files under `telegram/`, so causation is structurally impossible except via load. Reddened the shared gate three times in one working session.
- (2) observed once, under an all-mutations run only; never in any single-mutation run. The surfacing PR touches no TUI/client/HTTP code.
- Isolated re-verification on `aed80140` (both test files byte-identical to `origin/main`):
  - `telegram::supervisor::tests::healthy_run_resets_transient_backoff ... ok` — 5/5 consecutive runs, `finished in 0.08s` each.
  - `tui::project_ctl::poll::tests::classify_connect_error_is_transport ... ok` — `finished in 0.31s`.
- Both currently untracked — which is why successive reviewers keep re-diagnosing them from scratch.

## Resolution

- (1) Replace the wall-clock assertion with one that does not measure elapsed real time under contention: inject a clock, or assert on the backoff **state transition** (the counter reset) rather than its duration. Do not merely widen the threshold — a wider wall-clock bound is the same defect with a longer fuse.
- (2) Do not obtain a dead address by binding-and-dropping. Use an address that cannot be bound by another binder (reserved/blackhole address, or a port held open and actively refusing), so the fixture's premise cannot be invalidated. Fix it once and share it across all three copies.

## Acceptance

- Each test passes deterministically under `cargo test -p trusty-mpm` **without** `--lib`, at full parallelism, for **20 consecutive runs** — a single green run is exactly what a flake produces half the time.
- No `#[ignore]`. No serialization as a substitute for fixing the root cause: serializing (2) would mask the invalidated premise rather than remove it.
- The three `dead_loopback_url` copies are unified on a fixture whose dead-address guarantee does not depend on winning a port race.

### LINKED ISSUE #4413: trusty-memory tests depend on env-var side effects leaking between tests
**Defect:** trusty-memory tests pass under `cargo test` (shared process) but fail under per-test process isolation. They depend on environment-variable side effects set by *other* tests in the same binary.

**How it surfaced:** discovered incidentally while trying `cargo nextest run` as a CI speedup during PR #4410. nextest runs each test in its own process, which removed the leaked state and exposed the hidden ordering dependency. Nobody was looking for this.

**Why it matters:** these tests currently pass for the wrong reason. Any change to test ordering, parallelism, or runner can flip them red — and, worse, they are not actually verifying what they appear to verify in isolation.

**Family:** same class as the hermetic-test-isolation epic #3451.

**Resolution:** make each affected test set up its own environment rather than inheriting another test's leftovers. Adopting nextest later becomes possible once this is fixed — it is currently blocked by it.

### LINKED ISSUE #4414: execute_dispatches_to_python_and_parses_json flakes under parallel load (uv reads $HOME while other tests mutate it)

### LINKED ISSUE #4415: Flaky test: classify_connect_error_is_transport races on ephemeral port reuse
## WHERE
`tui::project_ctl::poll::tests::classify_connect_error_is_transport` (trusty-mpm).

## SYMPTOM
Failed once during a rebase of PR #4225 on 2026-07-30, then passed 5/5 on the PR branch, 4/4 on main, and in CI. The PR did not touch this test.

## ROOT CAUSE
The fixture binds an ephemeral port, drops the listener, then asserts that a later connect is REFUSED. The OS is free to reassign that port in the interval, so the connect can succeed and the assertion fails. The fixture is racy by construction — it will recur.

## WHY IT MATTERS
An intermittent failure here trains people to re-run rather than investigate, which is how a real regression gets waved through.

## FIX DIRECTION
Make the refusal deterministic rather than relying on an unbound ephemeral port — e.g. connect to a port the test holds closed by construction, or inject the error classification directly rather than exercising a real socket.

## CONSTRAINT
Do NOT resolve this by adding `#[ignore]` or deleting the test. The repo rule is that a failing gate is never turned green by removing coverage.

bobmatnyc:
Five filed flakes in the hermetic-test-isolation family (#3451), plus one
unfiled instance of the same class found while investigating #4407. Every fix
removes a dependence on shared state or wall-clock timing. None removes
coverage: no `#[ignore]`, no cfg-gate, no deleted test, no loosened bound —
trusty-mpm's lib suite goes from 4033 to **4035** passing tests, and two
assertions got strictly stricter.

### #4306(2) + #4415 — `dead_loopback_url` raced ephemeral-port reuse

Three copies of the fixture (`tui::project_ctl::poll::tests`,
`tui::coordinator::tests`, `tui::project_ctl::tests`) feeding **13 call sites**
obtained a "dead" address by binding an ephemeral port and dropping the
listener. That returns the port to the OS pool, so the premise held only until
another binder got it — then the connect succeeds and
`classify_connect_error_is_transport`'s `expect_err` panics, or the request
fails at the HTTP layer and misclassifies as `NonTransport`.

**A/B — the premise, measured directly** rather than waiting for the ~1-in-many
failure:

| fixture | port reassigned to a competing binder? |
|---|---|
| old (bind-and-drop) | **yes — accepted a connection after 16,103 ephemeral binds** (one wrap of macOS 49152–65535) |
| new (`127.0.0.1:1`) | no; explicit rebind rejected `EADDRINUSE`-style, 0/30,000 churned binds |

All three copies now share one fixture on `127.0.0.1:1` — privileged (measured:
unprivileged bind → `PermissionDenied`) and outside every ephemeral range, so it
cannot be handed out. This is **already the workspace convention**: trusty-review
says verbatim *"127.0.0.1:1 is always refused (port 1 is reserved/privileged)"*,
and trusty-analyze uses it too.

**A rejected design worth recording.** The obvious fix — bind a port and
deliberately skip `listen(2)`, so you both reserve it and get an RST — is
wrong on macOS: the kernel silently **drops** the SYN instead of resetting it,
so connects time out (measured: `TimedOut` after a full 10s). That would have
gone green (reqwest maps both refusal and timeout to `Transport`) while making
13 tests multi-second and handing the classification tests the wrong error kind.
It was caught only because the new self-test asserts the error **kind**, not
merely that the connect failed. Both properties are now pinned by tests, and a
`const` assertion stops a future edit moving the port above 1024.

### #4306(1) — telegram backoff reset asserted on wall-clock time

`healthy_run_resets_transient_backoff` asserted `backoff_component < 20ms`
against a policy `base` of **2ms**. The 18ms of headroom existed purely to
absorb scheduler jitter, so on a loaded machine the assertion measured wakeup
latency, not backoff state — reddening the gate three times in one session at
34.4ms and 21.79ms, in PRs touching zero files under `telegram/`.

Now runs on tokio's paused clock. `supervise` routes all timing through
`tokio::time`, so virtual time advances to each deadline and every gap is exact.
Widening the bound was rejected: same defect, longer fuse.

**The assertions got stricter, not looser** — two bounds became equalities
(`== delay_for(3)`, `== base`) and two new assertions pin the earlier escalation
steps. Verified by mutation, since a paused clock can make a test vacuous:

| mutation to `next_transient_count` | new assertion | old `< 20ms` bound |
|---|---|---|
| reset removed entirely | **FAILS** (128ms vs 2ms) | would fail |
| reset to `1` instead of `0` → `delay_for(2)` = 8ms | **FAILS** (8ms vs 2ms) | **would PASS** — regression missed |

Also fixed arithmetic the loose bound hid: it subtracted `policy.healthy_run`
where the run actually lasted `healthy_run + 5ms`.

**#4306 acceptance — 20 consecutive full-parallelism runs** of the whole lib
suite (4035 tests), on a machine under 12–86 load average from concurrent
agents. Both target tests: **20/20**.

### #4414 — `uv` read `$HOME` while 161 statements reassigned it

`uv` resolves its cache from `$HOME` at spawn. Captured output shows the real
sequence: a sibling test set `HOME=<tempdir>`, uv derived its cache path from
it, and the sibling's `TempDir` was **dropped mid-flight**:

```
Caused by: No such file or directory (os error 2) at path
  ".../T/.tmpHR50aj/.cache/uv/.tmptfs807"
ModuleNotFoundError: No module named 'python'
```

Both uv-spawning tests now hold `test_env::HOME_LOCK`. Verified exhaustive
before relying on it: **all 161** HOME-mutating statements in the crate — 116
`set_var` + 45 `remove_var`, across 72 test fns — sit inside a function holding
that lock, so holding it excludes every writer.

| | failures |
|---|---|
| `main`, 5 full-suite runs (load 34–72) | **3 / 5** |
| branch, 5 full-suite runs (load 12–20) | **0 / 5** |
| controlled interleaved, identical synthetic load (19–50) | base **1 / 5** → branch **0 / 5** |

The interleaved run exists because the ambient load dropped between the two
uncontrolled measurements, which would otherwise confound them. Combined:
**4/10 → 0/10**.

### #4413 — trusty-memory tests inherited another test's env var

`add_alias_round_trip_through_prompt_cache` and
`dispatch_discover_aliases_inserts_new_and_dedupes` build `AppState` inline
(they need `with_default_palace`, which `test_state()` does not offer) and so
never ran its `TRUSTY_SKIP_PALACE_ENFORCEMENT` write. They passed under
`cargo test` only because a sibling had set it first — and **in isolation they
verified nothing**, panicking at `palace_create` before any assertion.

Both now call a named `skip_palace_enforcement()` helper, which
`test_state`/`test_state_warming` share (replacing three copies of the same
`unsafe { set_var }` with one `OnceLock`-guarded write).

`cargo nextest run -p trusty-memory --no-fail-fast`: **2 failures → 527/527
passing** (2 failures reproduced twice before the fix).

### #4407 — `inference_available_false_without_key` read the ambient shell

The test asserts the gate stays closed with nothing configured, but
`inference_available("", false)` falls back to `OPENROUTER_API_KEY` in the
process env. Its `#[serial]` group excluded concurrent *test* writers while
doing nothing about the environment the suite inherits, so it failed for anyone
with a real key exported. Reproduced deterministically on `main`:

```
OPENROUTER_API_KEY=sk-... cargo test -p trusty-common --features memory-core \
  --lib semantic_consolidation
---- inference_available_false_without_key ----
assertion failed: !inference_available("", false)
```

Now **clears** the variable for its body (`EnvVarGuard::clear`, restored on
drop). Not a weakening — it makes the test verify what it always claimed, which
on a machine with the key set it could not verify at all. Also forces the
`.env.local` loader first so that process-global `OnceLock` cannot repopulate
the variable mid-assertion (the #3464 mechanism). Verified passing **both** with
and without the variable set.

### #4407 (new, unfiled) — banner tests used a lock that coordinated with nothing

Found while auditing for shared global state. The plain-mode banner tests
serialised themselves with a **file-local** `static ENV_LOCK: Mutex<()>`,
justified as *"the idiomatic solution without adding the serial_test crate"* —
a stale premise (serial_test is already a dev-dependency, used in this same bin
target) and ineffective for the job: a file-local mutex coordinates those tests
with each other and nothing else, while the other `$HOME` mutators **in the same
test binary** ran free against them — 19 HOME-mutating statements in 8 fns across
4 files in the `tm` bin target. They raced those, both directions. That is the shape #4407 describes — unrelated plain-mode tests
reddening together and clearing with no code change.

All nine env-mutating tests now use `#[serial_test::serial]`; the local mutex is
gone. Also fixes a hermeticity bug it hid: `banner_source_missing_falls_back`
exercises the `NotFound` arm, which **writes** `$HOME/.trusty-mpm/banner.txt` —
with the ambient `$HOME` that seeded a real file in the developer's home
(confirmed present locally) and could land in a concurrent test's tempdir. It
now sandboxes `$HOME` and asserts the seeded content, so the side effect is
covered rather than tolerated.

## What is NOT resolved — #4407's 7-test cluster

**The specific run #4407 cites is not in the retrievable Actions history**, so I
could not root-cause that cluster and am deliberately not claiming it. The only
two `CI`/`Test` failures on `main` across 2026-07-24…29 were run `30214415477`
(07-26) and `30177354047` (07-25) — both **single**-test failures, both of
`core::session_launch::…::prepare_session_excludes_spoofed_trusty_search_when_injector_disabled`,
not a 7-test cluster. Either the cited run was a PR run rather than `main`, or
the date/time in the issue is imprecise.

What I did instead was attack the *class*: audit `$HOME` mutators for
serialization gaps (found the banner defect above), run trusty-memory under
per-test process isolation, and fix the one named ambient-env test. `Refs`, not
`Closes`, for #4407.

## Two further load-sensitive flakes found, not fixed here

Both are genuinely separable — different crates, different mechanism from the
five filed issues, and each needs its own fix. Recommend filing:

1. **`connectors::tm::tests::create_session_full_lifecycle`** (trusty-mpm).
   Clean in acceptance runs 1–16 (load ≤ 50), failed runs 17–20 at load 74–86.
   Zero diff lines in `crates/trusty-mpm/src/connectors/`, and it does not use
   the fixture this PR changed.
2. **`trusty-agents/tests/api_e2e.rs`** — `support/api_server.rs:94` waits
   `Duration::from_secs(5)` for a freshly spawned binary to become healthy. Red
   3/3 in isolation at load 22–38 on this machine; CI (idle runner) is green.
   Same anti-pattern as the telegram bound: a wall-clock budget sized for an
   idle machine. My trusty-agents change is entirely inside `#[cfg(test)] mod
   tests` in the library and is not compiled into that binary.

## Gates

| gate | result |
|---|---|
| `cargo test -p trusty-mpm` (full, not `--lib`) | **exit 0** |
| `cargo test -p trusty-memory` | **exit 0** — 527 passed, 0 failed |
| `cargo test -p trusty-common --features memory-core` | **exit 0** — 656 passed, 0 failed |
| `cargo test -p trusty-agents --lib` | **exit 0** — 3110 passed, 0 failed |
| `cargo nextest run -p trusty-memory --no-fail-fast` | 527/527 |
| `cargo clippy --workspace --all-targets` (CI's exclusions) | **exit 0** |
| `cargo fmt --check` | **exit 0** |
| `check_line_cap.sh` / `check_sld.sh` / `check_test_pointers.sh` / `check_doc_numbers.sh` | **exit 0** |

`cargo test -p trusty-agents` (full) is red on my machine **only** via the
pre-existing `api_e2e` 5s budget above; `--lib`, which contains every line I
changed, is green.

Closes #4413
Closes #4414
Closes #4415
Closes #4306
Refs #4407
Refs #3451

---

## Round 2 — code-critic WARN addressed (1 HIGH, 2 MEDIUM, 2 LOW)

Verdict: https://github.com/bobmatnyc/trusty-tools/pull/4482#issuecomment-5146724603

**HIGH — my banner fix did not close the race it claimed to.** `pm_guard_bash`'s
`evaluate_worktree_add_command_expands_tmpdir_and_home` sets `HOME`/`TMPDIR` in
the same `tm` bin test binary with no `#[serial]` and no lock, so moving the
banner tests into the serial group never excluded it. Worse, my new assertion is
what made it bite: the old test used the ambient `$HOME` and asserted nothing
about the seeded file, so a concurrent `HOME=/Users/x` was invisible.

Reproduced first, window widened with sleeps, then re-run under the *same*
widened window after the fix. Because two things changed, an isolation matrix:

| `#[serial]` | `EnvGuard` | result (widened window, 2 threads) |
|---|---|---|
| absent | absent | **FAILED** — `left: None` |
| absent | present | **FAILED** — `left: None` |
| present | present | **5/5 ok** |

So `#[serial]` closes the race; the `EnvGuard` fixes a *separate* defect — the
test leaked `HOME=/Users/x` into the rest of the binary, and its trailing
`remove_var("TMPDIR")` cleared an inherited variable rather than restoring it,
and was skipped outright on an assertion panic.

**MEDIUM 1 — my scope claim was wrong and had shipped in the CHANGELOG.** "46
other `$HOME` mutators in the crate" counted the trusty-mpm **lib** test binary,
a different process. Neither `#[serial]` (threads within one binary) nor
process-global `$HOME` can span test targets. Re-audited per target: **19
HOME-mutating statements in 8 fns across 4 files** in the `tm` bin target — the
four banner tests, the `pm_guard_bash` test above, and the `set_home` /
`with_fake_home` helpers, whose single callers are each already `#[serial]`.
Corrected in `source.rs` and the CHANGELOG. The generalizable rule, now written
into both: **env-isolation audits are scoped per TEST TARGET, not per crate.**

**MEDIUM 2** — the `supervisor_tests.rs` module header still claimed no `tokio`
`test-util` paused-time feature was required, the opposite of what this PR does.
Reconciled with the Cargo.toml comment.

**LOW 1 — my three new equalities were self-referential.** They compared the
observed gap to `policy.delay_for(n)`, the same function that computed the wait,
so the critic's mutation sweep found two survivors. Now pinned to literals
(2ms, 8ms, 32ms, post-reset 2ms). Both survivors now die:

| mutation | before | after |
|---|---|---|
| `delay_for` multiplier `factor` → `factor + 1` | survives | **FAILED** |
| `delay_for` collapsed to `self.base` | survives | **FAILED** |

That takes this test from **5 of 8** enumerated mutations killed to **7 of 8**
(the last is a healthy-run threshold boundary a sibling test owns by design).

**LOW 2 — site count.** I wrote 115; the critic corrected to 159; both are wrong.
Re-measured two independent ways, agreeing exactly: 116 `set_var("HOME")` + 45
`remove_var("HOME")` = **161** statements across 72 fns, no overlap. 116+45 is
161, not 159. Only the number was wrong — the coverage claim it supports still
holds.

### Round-2 gates (rebased onto current `main`, was BEHIND)

| gate | result |
|---|---|
| `cargo test -p trusty-mpm` (full) | exit 0 |
| — lib target | 4043 passed, 0 failed |
| — `tm` bin target | 1319 passed, 0 failed |
| `cargo test -p trusty-memory` | exit 0 |
| `cargo test -p trusty-common --features memory-core` | exit 0 — 635 passed |
| `cargo test -p trusty-agents --lib` | exit 0 — 3111 passed |
| `cargo clippy --workspace --all-targets` (CI's 3 GUI exclusions) | exit 0 |
| `check_line_cap` / `check_sld` / `check_test_pointers` / `check_doc_numbers` | all 0 |

Not touched, per the verdict: the paused-clock conversion (independently
verified), and everything in the critic's Verified-CLEAN list.

---

## Round 3 — rebased onto the per-PR changelog-fragment convention (#4476)

[#4477](https://github.com/bobmatnyc/trusty-tools/pull/4477) landed while this
PR was open. It replaced the shared `## [Unreleased]` section with per-PR
fragment files and migrated all 25 crates, so the headings my changelog edits
targeted no longer exist on `main`.

Each of the four bullets moved verbatim into its own crate's `changelog.d/`:

| crate | fragment |
|---|---|
| trusty-mpm | `changelog.d/4306-hermetic-test-fixtures.md` |
| trusty-memory | `changelog.d/4413-palace-tests-set-own-env.md` |
| trusty-agents | `changelog.d/4414-uv-home-lock.md` |
| trusty-common | `changelog.d/4407-inference-gate-clears-key.md` |

No `## [Unreleased]` heading is reintroduced anywhere — the assembler hard-fails
on a leftover one, which would break that crate's release path. All four
`CHANGELOG.md` files are byte-identical to `main`.

Two corrections carried in while moving the text: the trusty-mpm entry's "46
other `$HOME` mutators in the crate" became the correct same-target figure (19
statements, 8 fns, 4 files in the `tm` bin target) plus the general rule, and its
lead line said "three" fixtures when it is four.

Verified with the gate CI actually runs and with the real assembler:

```
$ CHANGELOG_GATE_BASE=$(git rev-parse origin/main) bash scripts/check_changelog_fragment.sh
OK   trusty-agents: changelog.d fragment present and valid
exit 0
```

(`trusty-memory` is not listed because its only change is `src/tools/tests.rs`,
which the gate exempts as test-only; its fragment is recorded anyway.)

`assemble-changelog.sh <crate> 9.9.9 --check` and `<crate> --stdout` both exit 0
for all four crates, and each preview contains that crate's new bullet.

### One extra fix, forced by the migration: the assembler rejected a large fragment

Adding the fragments turned the `Per-PR changelog fragment` job red — against a
fragment **this PR did not write**. `parse_fragment` validated "does this contain
a bullet" with `printf '%s\n' "$body" | grep -qE '^-[[:space:]]'` under
`set -o pipefail`. That is a SIGPIPE race: `grep -q` exits the instant it
matches, and when the body exceeds the pipe buffer the still-writing `printf`
takes EPIPE, which `pipefail` promotes to a pipeline failure — so a fragment
whose *first* body line is a bullet is reported as having none.

```
scripts/assemble-changelog.sh: line 141: printf: write error: Broken pipe
ERROR: 4476-migrated-added.md has a category but no bullet
```

Size- and platform-dependent, which is why it survived review:
`trusty-agents/changelog.d/4476-migrated-added.md` is 60 KB with the bullet on
line 1 — it passes on macOS and fails on Linux CI. Fixed with a herestring (no
second process, so no pipe to break). The old construct false-negatives on a
200 KB body; the new one finds the bullet 10/10. All 25 crates' fragment
directories are accepted, and the gate reports OK for every crate.

### Other hermeticity litter observed, not fixed

`cargo test -p trusty-mpm` scaffolds `crates/trusty-mpm/{.claude/,.trusty-mpm/,CLAUDE.md}`
into the working tree — untracked, not gitignored. Same family as the
`crates/trusty-memory/.trusty-tools/` file the critic noted, and same family as
#3451. My diff touches no session-scaffolding code, so this is not from this PR;
I have not A/B-verified it against `main`. Worth its own ticket.

### DIFF
--- a/crates/trusty-agents/changelog.d/4414-uv-home-lock.md
+++ b/crates/trusty-agents/changelog.d/4414-uv-home-lock.md
@@ -1,0 +1,12 @@
+Fixed
+
+- `python_skill`'s two `uv`-spawning tests no longer race the 161 statements in
+  this crate that reassign `$HOME` process-wide (closes [#4414](https://github.com/bobmatnyc/trusty-tools/issues/4414); refs [#4407](https://github.com/bobmatnyc/trusty-tools/issues/4407), [#3451](https://github.com/bobmatnyc/trusty-tools/issues/3451)).
+  `uv` resolves its cache from `$HOME` at spawn time; under parallel load a
+  sibling test's `HOME=<tempdir>` was reaped mid-flight, leaving `uv` pointing at
+  a deleted tree (`No such file or directory ... /.tmpHR50aj/.cache/uv/...`).
+  Both tests now hold `test_env::HOME_LOCK` for their whole body — verified
+  exhaustive first: all 161 HOME-mutating statements (116 `set_var` + 45
+  `remove_var`, across 72 test fns) sit inside a function holding that lock, so
+  nothing can move `$HOME` while the subprocess runs. Measured on a loaded
+  machine: 3 failures in 5 full-suite runs before, 0 in 5 after.
--- a/crates/trusty-agents/src/tools/python_skill.rs
+++ b/crates/trusty-agents/src/tools/python_skill.rs
@@ -339,6 +339,46 @@ mod tests {
             .unwrap_or(false)
     }
 
+    /// Hold `$HOME` still for the duration of a test that shells out to `uv`
+    /// (#4414).
+    ///
+    /// Why: `uv` resolves its cache from `$HOME` (`$HOME/.cache/uv`) at spawn
+    /// time, and ~115 sites across this crate's tests reassign `HOME`
+    /// process-wide to a `TempDir`. Under parallel load the sequence that
+    /// actually bit was: a sibling test set `HOME=<tempdir>`, `uv` derived its
+    /// cache path from that tempdir, and the sibling's `TempDir` was then
+    /// dropped — deleting the tree out from under the running interpreter.
+    /// Captured failures name the vanished path directly:
+    /// `Caused by: No such file or directory (os error 2) at path
+    /// ".../T/.tmpHR50aj/.cache/uv/.tmptfs807"`, and
+    /// `ModuleNotFoundError: No module named 'python'` for the same reason.
+    /// Measured on `main`: 3 failures in 5 full-suite runs.
+    ///
+    /// What: acquires `test_env::HOME_LOCK`, the mutex this crate's HOME
+    /// convention is built on. Verified exhaustive for this crate before
+    /// relying on it: all 115 `set_var("HOME", ..)` sites (across 72 test fns)
+    /// hold `HOME_LOCK`, so holding it here excludes every writer — nothing is
+    /// mutating `HOME`, and no sibling's `TempDir` is being reaped, while the
+    /// `uv` subprocess runs. That makes the test hermetic with respect to the
+    /// variable rather than merely likely to win the race; widening a timeout
+    /// or retrying would have left the same race with a longer fuse.
+    ///
+    /// `unwrap_or_else(into_inner)` so one panicking test cannot poison the
+    /// lock for its siblings — the same acquisition idiom the other 43 callers
+    /// use. Plain `HOME_LOCK` rather than `test_env::lock_home()` because these
+    /// tests exercise no production path guarded by
+    /// `home_lock_held_by_this_thread` (only `listeners::store::events_dir` is).
+    /// Test: [`execute_dispatches_to_python_and_parses_json`] and
+    /// [`execute_maps_python_error_key_to_recoverable_result`] — the two tests
+    /// that spawn `uv`. [`execute_reports_spawn_failure_as_recoverable`]
+    /// deliberately does not: it spawns a nonexistent binary and never reads
+    /// `HOME`.
+    fn hold_home_for_uv() -> std::sync::MutexGuard<'static, ()> {
+        crate::test_env::HOME_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+    }
+
     #[test]
     fn manifest_parses_cto_db_shape() {
         let manifest = load_manifest(&cto_db_skill_dir()).expect("manifest should parse");
@@ -386,12 +426,22 @@ mod tests {
         assert!(build_plugin(tmp.path()).is_err());
     }
 
+    // #4414: holds the guard across `.await` deliberately — the whole point is
+    // that `$HOME` must not move while the `uv` subprocess is alive. The default
+    // `#[tokio::test]` runtime is current-thread, which pins this test (every
+    // `.await` included) to one OS thread, so the guard is sound here.
+    #[allow(
+        clippy::await_holding_lock,
+        reason = "#4414: holding HOME_LOCK across the uv subprocess IS the fix"
+    )]
     #[tokio::test]
     async fn execute_dispatches_to_python_and_parses_json() {
         if !uv_available() {
             eprintln!("SKIP: `uv` not on PATH — cannot exercise the real cto-db skill subprocess");
             return;
         }
+        // #4414: pin $HOME for the whole body — see `hold_home_for_uv`.
+        let _home = hold_home_for_uv();
         let plugin = build_plugin(&cto_db_skill_dir()).expect("plugin should build");
         let tool = plugin
             .tools
@@ -410,12 +460,22 @@ mod tests {
         assert!(parsed["groups"].as_array().is_some_and(|g| !g.is_empty()));
     }
 
+    // #4414: holds the guard across `.await` deliberately — the whole point is
+    // that `$HOME` must not move while the `uv` subprocess is alive. The default
+    // `#[tokio::test]` runtime is current-thread, which pins this test (every
+    // `.await` included) to one OS thread, so the guard is sound here.
+    #[allow(
+        clippy::await_holding_lock,
+        reason = "#4414: holding HOME_LOCK across the uv subprocess IS the fix"
+    )]
     #[tokio::test]
     async fn execute_maps_python_error_key_to_recoverable_result() {
         if !uv_available() {
             eprintln!("SKIP: `uv` not on PATH — cannot exercise the real cto-db skill subprocess");
             return;
         }
+        // #4414: pin $HOME for the whole body — see `hold_home_for_uv`.
+        let _home = hold_home_for_uv();
         let plugin = build_plugin(&cto_db_skill_dir()).expect("plugin should build");
         let tool = plugin
             .tools
--- a/crates/trusty-agents/src/tools/python_skill.rs
+++ b/crates/trusty-agents/src/tools/python_skill.rs
@@ -343,7 +343,7 @@ mod tests {
     /// (#4414).
     ///
     /// Why: `uv` resolves its cache from `$HOME` (`$HOME/.cache/uv`) at spawn
-    /// time, and ~115 sites across this crate's tests reassign `HOME`
+    /// time, and 161 statements across this crate's tests reassign `HOME`
     /// process-wide to a `TempDir`. Under parallel load the sequence that
     /// actually bit was: a sibling test set `HOME=<tempdir>`, `uv` derived its
     /// cache path from that tempdir, and the sibling's `TempDir` was then
@@ -355,9 +355,10 @@ mod tests {
     /// Measured on `main`: 3 failures in 5 full-suite runs.
     ///
     /// What: acquires `test_env::HOME_LOCK`, the mutex this crate's HOME
-    /// convention is built on. Verified exhaustive for this crate before
-    /// relying on it: all 115 `set_var("HOME", ..)` sites (across 72 test fns)
-    /// hold `HOME_LOCK`, so holding it here excludes every writer — nothing is
+    /// convention is built on. Verified exhaustive for this crate before relying
+    /// on it: all 161 HOME-mutating statements — 116 `set_var("HOME", ..)` plus
+    /// 45 `remove_var("HOME")`, across 72 test fns — sit inside a function that
+    /// holds `HOME_LOCK`, so holding it here excludes every writer. Nothing is
     /// mutating `HOME`, and no sibling's `TempDir` is being reaped, while the
     /// `uv` subprocess runs. That makes the test hermetic with respect to the
     /// variable rather than merely likely to win the race; widening a timeout
--- a/crates/trusty-common/changelog.d/4407-inference-gate-clears-key.md
+++ b/crates/trusty-common/changelog.d/4407-inference-gate-clears-key.md
@@ -1,0 +1,11 @@
+Fixed
+
+- `semantic_consolidation::inference_available_false_without_key` no longer
+  fails on any machine that exports a real `OPENROUTER_API_KEY` (refs [#4407](https://github.com/bobmatnyc/trusty-tools/issues/4407), [#3451](https://github.com/bobmatnyc/trusty-tools/issues/3451)).
+  The test asserts the inference gate stays closed with no key configured, but
+  `inference_available("", false)` falls back to reading that variable from the
+  process environment — so "absent from the ambient shell" was an unstated
+  precondition, and its `#[serial]` group excluded concurrent test writers while
+  doing nothing about the environment the suite inherits. It now CLEARS the
+  variable for its body via an `EnvVarGuard::clear` (restored on drop), which is
+  what it always claimed to test.
--- a/crates/trusty-common/src/memory_core/semantic_consolidation/mod.rs
+++ b/crates/trusty-common/src/memory_core/semantic_consolidation/mod.rs
@@ -105,9 +105,33 @@ mod tests {
     /// mutators) because it held no lock at all. Join the same
     /// `dotenv_credential_env` group as every other test that reads or
     /// writes this var.
+    ///
+    /// #4407: the `#[serial]` group excludes concurrent *test* writers but not
+    /// the AMBIENT environment. The var being absent was still an unstated
+    /// precondition on the machine, so this test failed for anyone whose shell
+    /// exports a real `OPENROUTER_API_KEY` — which is every developer who has
+    /// configured one, and was the observed cause on a CI runner whose shell had
+    /// it set. Reproduced deterministically: `OPENROUTER_API_KEY=sk-… cargo test
+    /// -p trusty-common --features memory-core --lib semantic_consolidation`
+    /// failed on `main` at `assertion failed: !inference_available("", false)`.
+    ///
+    /// Fixed by CLEARING the var for the body rather than assuming it is clear.
+    /// This does not weaken the test — it makes it verify what it always claimed
+    /// to: "no key configured anywhere ⇒ the gate is closed". Previously, on a
+    /// machine WITH the var set it could not verify that at all; it just failed.
+    /// `EnvVarGuard` restores the prior value on drop, so a developer's real key
+    /// survives the test run.
     #[test]
     #[serial_test::serial(dotenv_credential_env)]
     fn inference_available_false_without_key() {
+        // #4407: `force_env_local_loaded` FIRST — the `.env.local` loader is a
+        // process-global `OnceLock` that folds file contents into the process
+        // environment on whichever call happens to be first. If that first call
+        // were the one inside `inference_available` below, it could re-populate
+        // the very var this test just cleared, mid-assertion (the #3464
+        // mechanism). Forcing it now makes the clear below the last word.
+        crate::inference::credentials::load_env_local_once();
+        let _guard = EnvVarGuard::clear("OPENROUTER_API_KEY");
         assert!(!inference_available("", false));
         assert!(!inference_available("   ", false));
     }
@@ -169,6 +193,19 @@ mod tests {
             let previous = std::env::var(key).ok();
             // Safety: test-only; caller is `#[serial]`.
             unsafe { std::env::set_var(key, value) };
+            Self { key, previous }
+        }
+
+        /// Remove `key` for the guard's lifetime, restoring it on drop (#4407).
+        ///
+        /// Why: a test asserting "behaviour X holds when this var is UNSET" must
+        /// make it unset, not assume the ambient shell left it that way — see
+        /// `inference_available_false_without_key`, which failed on any machine
+        /// with a real `OPENROUTER_API_KEY` exported.
+        fn clear(key: &'static str) -> Self {
+            let previous = std::env::var(key).ok();
+            // Safety: test-only; caller is `#[serial]`.
+            unsafe { std::env::remove_var(key) };
             Self { key, previous }
         }
     }
--- a/crates/trusty-memory/changelog.d/4413-palace-tests-set-own-env.md
+++ b/crates/trusty-memory/changelog.d/4413-palace-tests-set-own-env.md
@@ -1,0 +1,13 @@
+Fixed
+
+- two `tools::tests` no longer depend on an environment variable another test
+  happened to leak (closes [#4413](https://github.com/bobmatnyc/trusty-tools/issues/4413); refs [#4407](https://github.com/bobmatnyc/trusty-tools/issues/4407), [#3451](https://github.com/bobmatnyc/trusty-tools/issues/3451)).
+  `add_alias_round_trip_through_prompt_cache` and
+  `dispatch_discover_aliases_inserts_new_and_dedupes` build `AppState` inline
+  (they need `with_default_palace`) and so never ran `test_state()`'s
+  `TRUSTY_SKIP_PALACE_ENFORCEMENT` write — they passed under `cargo test` only
+  because a sibling test in the same process had set it first. In isolation they
+  verified nothing, failing at `palace_create` before reaching any assertion.
+  Both now call a named `skip_palace_enforcement()` helper, which
+  `test_state`/`test_state_warming` share. `cargo nextest run -p trusty-memory`
+  (per-test process isolation) goes from 2 failures to 527/527 passing.
--- a/crates/trusty-memory/src/tools/tests.rs
+++ b/crates/trusty-memory/src/tools/tests.rs
@@ -21,22 +21,46 @@ use uuid::Uuid;
 /// when the test scope ends.
 /// Test: Every test in this module that constructs state.
 ///
-/// Why (issue #88): sets `TRUSTY_SKIP_PALACE_ENFORCEMENT=1` so that
-/// existing tests that call `palace_create` with arbitrary names continue
-/// to work. The enforcement gate in `handle_palace_create` bypasses the
-/// project-slug check when this env var is set, which is the correct
-/// behaviour for test helpers that point at isolated tempdirs. Production
-/// processes never set this variable.
-fn test_state() -> (AppState, tempfile::TempDir) {
-    // SAFETY: tests in this module run in-process; setting the bypass var
-    // here races with any test that reads env before or after, but since
-    // the value is "set to the same constant forever" once any test runs,
-    // the race is benign — all tests should see "1" within the first
-    // iteration. Tests that need stricter serialisation already use
-    // `env_test_lock()`.
-    unsafe {
+/// Why (issue #88): calls [`skip_palace_enforcement`] so that existing tests
+/// that call `palace_create` with arbitrary names continue to work.
+/// Set `TRUSTY_SKIP_PALACE_ENFORCEMENT=1` for this test process (#88, #4413).
+///
+/// Why: `handle_palace_create` rejects a palace name that does not match the
+/// project slug derived from cwd unless this var is set, so every test that
+/// creates a palace with an arbitrary name (`"ctx"`, `"disc"`, …) needs it.
+/// Production processes never set it.
+///
+/// Why it exists as a named helper rather than a line inside [`test_state`]
+/// (#4413): two tests here build their `AppState` INLINE — they need
+/// `with_default_palace`, which [`test_state`] does not offer — and so never
+/// ran [`test_state`]'s `set_var`. They passed anyway under `cargo test`, purely
+/// because some *other* test in the same process had already set the var: a
+/// hidden ordering dependency, and one that made them pass for the wrong reason
+/// (in isolation they verify nothing, because they never get past
+/// `palace_create`). Per-test process isolation removes the donor and exposes
+/// it — `cargo nextest run -p trusty-memory` failed both, which is how #4413
+/// was found. Making the requirement callable, and calling it from each test
+/// that needs it, is what makes each test self-sufficient.
+///
+/// What: writes the var at most once per process via a `OnceLock`, so N callers
+/// perform ONE write rather than N racing ones (the pre-existing `test_state`
+/// body re-wrote it on every call). Idempotent and safe to call from any test.
+/// Test: [`add_alias_round_trip_through_prompt_cache`] and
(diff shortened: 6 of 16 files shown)


bobmatnyc:
## Code Critic Verdict: WARN

Independent adversarial review. One HIGH, no CRITICAL — the code proceeds, but the
finding must be tracked. Reviewed against the spec (the five issues) and the code only.

**The headline question — does the paused-clock test survive independent mutation? YES.**
I did not accept the author's mutation table; I reproduced it and extended it to seven
mutations of `supervisor.rs`. `healthy_run_resets_transient_backoff` is not vacuous.

| # | Mutation to production code | This test | Killed by suite |
|---|---|---|---|
| M1 | reset removed (`let effective_prior = prior;`) | **FAILS** | yes |
| M2 | reset to `1` instead of `0` | **FAILS** | yes |
| M3 | `delay_for` multiplier `factor` → `factor + 1` | survives | yes — `backoff_grows_and_caps` |
| M4 | comparison inverted (`>=` → `<`) | **FAILS** | yes |
| M5 | `delay_for` exponent off-by-one (`saturating_sub(1)` → `failures`) | **FAILS** | yes |
| M6 | `run_duration` + 1ms (threshold boundary) | survives | yes — `healthy_run_resets_transient_count` |
| M7 | counter increment `+1` → `+2` | **FAILS** | yes |
| M8 | `delay_for` collapsed to `self.base` | survives | yes — 2 sibling tests |

5/7 killed by the converted test itself; both survivors killed by sibling tests that pin
`delay_for` against literals. Nothing regressed into the "green because it measures
nothing" state. The conversion is a genuine strengthening.

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/tests.rs` | 777-785 | Unserialized `$HOME` mutator in the **same test binary** as the banner tests — so the #4407 banner race is **not** closed, and the PR's new assertion is exposed to it | Add `#[serial_test::serial]` to `evaluate_worktree_add_command_expands_tmpdir_and_home` and restore `HOME`/`TMPDIR` on exit |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/formatters/banner/source.rs` + `crates/trusty-mpm/CHANGELOG.md` | 235-251 / "46 other `$HOME` mutators" bullet | The stated mechanism is wrong: 39 of the 46 sites are in a **different test binary** (a different process), which `#[serial]` and process-global env cannot span | Correct to the 7 same-target sites in 3 files; drop the cross-binary claim |
| MEDIUM | `crates/trusty-mpm/src/telegram/supervisor_tests.rs` | 7-10 | Module header now contradicts the change: "no `tokio` `test-util` paused-time feature is required", while this PR's own `Cargo.toml` comment says test-util is required *by this module* | Update the header to say the reset test runs on the paused clock |
| LOW | `crates/trusty-mpm/src/telegram/supervisor_tests.rs` | 184-199 | The three new equalities compare the observed gap to `policy.delay_for(n)` — the same production fn — so they are self-referential w.r.t. the delay curve (M3 and M8 both survive). The comment overstates them as proving "the escalation SCHEDULE" | Soften the comment, or pin one literal (`Duration::from_millis(32)`); the curve itself is already covered by `backoff_grows_and_caps` |
| LOW | `crates/trusty-agents/src/tools/python_skill.rs` + `crates/trusty-agents/CHANGELOG.md` | 346, 362 | "all 115 `set_var("HOME", ..)` sites" — actual is 116 `set_var` + 45 `remove_var` = **159** HOME-mutating statements in 70 fns | State 159 sites / 70 fns, or drop the number |

### HIGH — the #4407 banner race is still open, and now louder

`src/bin/tm/commands/pm_guard_bash/tests.rs:777`:

```rust
#[test]                                    // <-- no #[serial_test::serial]
fn evaluate_worktree_add_command_expands_tmpdir_and_home() {
    let cwd = Path::new("/Users/x/proj");
    unsafe {
        std::env::set_var("TMPDIR", "/private/tmp/claude-502/scratch");
        std::env::set_var("HOME", "/Users/x");   // never restored
    }
```

This is in the `tm` **bin** test binary, the same one as the banner tests — confirmed by
`cargo test -p trusty-mpm --bin tm -- --list`, which lists both
`commands::pm_guard_bash::tests::evaluate_worktree_add_command_expands_tmpdir_and_home`
and all ten `formatters::banner::source::tests::*`. It holds no lock and is in no serial
group, so moving the banner tests onto `#[serial]` does not exclude it.

The PR's new assertion at `source.rs:294-300` is what turns this into a red test.
`banner_source_missing_falls_back` previously used the ambient `$HOME` and asserted
nothing about the seeded file; it now sandboxes `$HOME` and asserts the seeded content,
so a concurrent `set_var("HOME", "/Users/x")` landing between its `set_var` and
`load_banner_art()` sends the write to `/Users/x` and the read comes back `None`.

Reproduced (window widened with sleeps to make the race deterministic, then reverted):

```
running 2 tests
test commands::pm_guard_bash::tests::evaluate_worktree_add_command_expands_tmpdir_and_home ... ok
test formatters::banner::source::tests::banner_source_missing_falls_back ... FAILED

assertion `left == right` failed: the NotFound arm seeds $HOME/.trusty-mpm/banner.txt
 — it must land in THIS test's sandbox, never the ambient home (#4407)
  left: None
 right: Some("\n▄█████▄   ▄█████▄ ...")

test result: FAILED. 1 passed; 1 failed; 0 ignored; 1318 filtered out
```

Honest caveat on the natural rate: the unwidened window is the ~sub-millisecond gap
between `set_var("HOME", &fake_home)` and `load_banner_art()` completing its read/mkdir/
write. It is narrow — but that is precisely the profile #4407 describes, and this PR's
audit claimed exhaustiveness over `$HOME` mutators, which it is not. It is a one-line
fix. The PR should not ship a de-flaking change that leaves a new assertion racing an
unserialized sibling it audited past.

## Required Changes

1. Add `#[serial_test::serial]` to
   `pm_guard_bash::tests::evaluate_worktree_add_command_expands_tmpdir_and_home`, and
   restore `HOME`/`TMPDIR` (an RAII guard, matching `session/start_tests.rs`'s
   `HomeGuard`) so it does not leave `HOME=/Users/x` for the rest of the binary.
2. Re-audit `$HOME` mutators **per test target**, not per crate, and correct the
   `source.rs` comment and the trusty-mpm `CHANGELOG.md` bullet accordingly.
3. Update the `supervisor_tests.rs` module header — it now asserts the opposite of what
   the PR does.
4. Fix the two `delay_for` self-reference / site-count overstatements (LOW; fold in).

## Verified CLEAN

Executed, not read:

- **Paused clock is not vacuous.** 7 independent mutations, table above. Runtime dropped
  to 0.04s for the whole 11-test module — the wall-clock measurement is genuinely gone,
  not merely widened.
- **`HOME_LOCK` coverage in trusty-agents is complete.** Scripted audit of all 159
  HOME-mutating statements across 70 fns in `crates/trusty-agents/src`: **0 sites**
  outside a function that acquires `HOME_LOCK` / `lock_home()`. The count in the docs is
  wrong; the coverage claim is right.
- **No `HOME_LOCK` deadlock or ordering hazard introduced.** `hold_home_for_uv` takes
  exactly one lock (no ENV_LOCK+HOME_LOCK ordering pair), handles poisoning with
  `unwrap_or_else(into_inner)` so a panicking test cannot wedge siblings, and runs under
  the default current-thread `#[tokio::test]` runtime, which makes holding it across
  `.await` sound. The raw-`HOME_LOCK` idiom (rather than `lock_home()`) is the majority
  convention here — 43+ existing call sites — and `#3952`'s `lock_home()` is required
  only where `listeners::store::events_dir` is reachable, which this path is not. The
  `uv_available()` early-return correctly precedes the lock acquisition. Both uv tests
  genuinely ran (uv on PATH, 0 skips) and passed.
- **`127.0.0.1:1` works on Linux CI, not just macOS.** The `Test` job is
  `runs-on: ubuntu-latest` with no `container:` key anywhere in `ci.yml`, running
  `cargo test --workspace` as the non-root `runner` user — and it is green, so both
  self-tests passed there. The root-container concern is *fail-loud, not silent*:
  `dead_loopback_port_is_not_bindable_or_ephemeral` asserts `bind(addr).is_err()`, which
  a root container would fail outright rather than degrade to a timeout.
- **The error-KIND assertion protects every call site, not just the self-test.**
  `dead_loopback_url()` itself performs the `Once`-guarded `ConnectionRefused` check, so
  the first of the 13 call sites in any test process validates the premise for all of
  them, with a named panic naming the remedy.
- **`test_support` is `#[cfg(test)]`-gated** (`lib.rs:260`) — no dead code or new
  constant in production builds.
- **#4413 makes the tests meaningfully assert, not merely stop panicking.** A/B: removing
  only the two new `skip_palace_enforcement()` calls reproduces
  `palace name 'disc' does not match the project slug 'trusty-memory'` in isolation;
  with them, both tests reach and pass real production assertions (`add_alias` triple,
  `list_prompt_facts` surfacing, prompt-cache formatted block).
- **#4407 `OPENROUTER_API_KEY` A/B.** Removing only the `load_env_local_once()` +
  `EnvVarGuard::clear` lines and running `OPENROUTER_API_KEY=sk-… cargo test` reproduces
  `assertion failed: !inference_available("", false)`. With the fix, green both with and
  without the variable set.
- **No coverage removed.** No `#[ignore]`, no cfg-gate, no `--exclude`, no deleted test,
  no loosened bound in the diff. Two bounds became equalities and two assertions were
  added.
- **All four affected suites green locally** (full, not `--lib`):
  `trusty-mpm` 4042 lib + 1319 `tm` bin + 20 integration targets, `trusty-memory` 470+,
  `trusty-common --features memory-core` 635+, `trusty-agents --lib` 3111 — 0 failures.
- **Repo lints:** `check_line_cap.sh` PASS, `check_sld.sh` PASS locally; CI's
  `Test: doc-comment pointer lint`, `SLD (DOC-38)`, `DOC-N/ADR number allocation`,
  `500-line file-size cap`, `Clippy`, `Format check`, `MSRV check (1.91)` all pass.
  `#[allow(..., reason = ...)]` is fine at MSRV 1.91.
- **CI: 26/26 green.** Not a zero-run conflicting PR. `mergeStateStatus: BEHIND` —
  needs a rebase before merge, no conflict.
- **`closingIssuesReferences` is exactly `[4306, 4413, 4414, 4415]`.** No stray closing
  keyword touches #4407 or #3451 — both are correctly `Refs`. The known auto-close trap
  is clear.
- **The wrong-cwd invocation error does not contaminate other conclusions.** Every claim
  I re-derived above was re-measured from this worktree; the four suite results are mine,
  not the author's.

## Notes

- `dead_loopback_port_is_not_bindable_or_ephemeral`'s 2,000-iteration sweep wraps its
  assertion in `if let Ok(other) = bind("127.0.0.1:0")`, so under fd exhaustion the loop
  would assert nothing and pass silently. Not flagged as a finding — the comment says so
  explicitly and the real guarantee is the compile-time `DEAD_LOOPBACK_PORT < 1024`
  assertion, which is the right place for it.
- A full `cargo test -p trusty-memory` run deposits `crates/trusty-memory/.trusty-tools/trusty-memory.yaml`
  into the working tree (untracked, not gitignored). I isolated it: it is **not** produced
  by either #4413 test nor by a `test_state()` test, so it is pre-existing and out of
  scope here — worth a separate ticket in the same hermeticity family (#3451).
- The two further flakes the PR declines to fix (`connectors::tm::tests::create_session_full_lifecycle`,
  `trusty-agents/tests/api_e2e.rs` 5s health budget) are correctly scoped out — neither
  touches a line in this diff. Filing both is the right call.

Issue #4495, opened 2026-07-31
feat(trusty-agents): read trusty-mpm's .claude/agents artifacts with the SHARED frontmatter parser


bobmatnyc:
Owner directive (Bob, 2026-07-31): "I want to use the same agents across the system. mpm, code, agents. … What agents we can use is a platform and assistant configuration piece. We're trying to rewrite as little as feasible."

trusty-agents currently hand-rolls TWO frontmatter parsers (`agents::claude_mpm_loader::ClaudeMpmFrontmatter` and `agents::registry::md_agent::MdAgentFrontmatter`) while `trusty_agents_common::agents::metadata::agent_metadata_from_str` — the shared, product-agnostic reader that wraps the same `split_frontmatter` grammar `compose_agent` EMITS with — is already a live dependency of the crate, used only for compress/adapters/rbac/perf. Zero agent-loading code uses it.

trusty-code already adopted the shared reader for exactly this artifact class in #3539 (`plugins::agents::load_plugin_agent`). This is the precedent to mirror.

Scope (PR 1, additive, no behavior change): add `agents::mpm_bridge` — read one `.claude/agents/*.md` deploy artifact via `agent_metadata_from_str` and project it onto trusty-agents' own `AgentConfig`.

Decided, not up for re-litigation: consume the ALREADY-FLATTENED deploy artifacts (leaf-only, exactly trusty-code's posture). No `extends:` resolution for mpm-sourced agents — trusty-mpm flattens `extends` before deploy, so there is nothing to resolve.

Projection rules (correctness-critical):
- clean scalar copies: `name`, `role`, `description`, `model`, `max_tokens`
- `extends:` present-but-ignored, with a warning
- `skills:` NOT mapped onto `[skills].allow`. Same name, different semantics: mpm's is a co-deployment DEPENDENCY list; trusty-agents' `[skills].allow` is a PERMISSION GATE (`None` = does not use skill grants, `Some([])` = grants none). Mapping one to the other silently forges a grant.
- `tier` NOT populated — it is DERIVED (`AgentTier::for_kind`); sub-agents are L1 by construction
- `tools:` -> `ToolsConfig.allowed` (exact-name), never `.allow` (globs). Where the two readings differ the restrictive one wins.
- everything else dropped with ONE aggregated warning

No reachability change: `ASSISTANT_REACHABLE_SUBAGENTS`, persona `[subagents].delegate_allowed`, and `SubagentAllowSet` are untouched.

[...]

PR #4506, opened 2026-07-31
feat(trusty-agents): read a role from claude-mpm frontmatter behind a fail-closed normalization layer

### LINKED ISSUE #4502: feat(trusty-agents): read a role from claude-mpm frontmatter behind a fail-closed normalization layer
`agents::claude_mpm_loader::ClaudeMpmFrontmatter` never extracts any role field, and `to_agent_config` HARDCODES `role: "agent"`. Every claude-mpm-loaded agent therefore carries a role that matches nothing in `runtime::tool_registry::ASSISTANT_ALLOWED_DELEGATE_ROLES`, and the agent's real domain (engineer / qa / ops / …) is discarded at load time.

## The trap

A verbatim pass-through is NOT the fix. The allowlist is a coarse per-DOMAIN vocabulary (`engineer, qa, researcher, ticketing, documentation, ops, planner, assistant`), and the artifacts carry their own vocabularies which do not line up with it:

- deployed `.claude/agents/*.md` (what this loader actually reads) carry `agent_type:`, NOT `role:` — trusty-mpm's composer does not emit `role:` at all
- `agent_type` includes values with no counterpart in the allowlist: `research` (vs `researcher`), `security`, `universal`, `system`, `trusty-mpm`, and at least one literal `engineer|qa|ops|universal|documentation` template placeholder
- trusty-mpm's SOURCE assets (`crates/trusty-mpm/src/assets/agents/*.md`) do carry `role:`, including `base`/`base-engineer`/`base-ops`/`base-qa`/`base-research` composition FRAGMENTS that must never be dispatchable

## Scope

Add a normalization step mapping both observed vocabularies onto the coarse allowlist words, built from the real data. Requirements:

- FAIL-CLOSED: an unrecognized or absent value maps to a sentinel that is NOT in the allowlist, preserving today's effective behavior. Never a default-allow.
- Explicit table, never a suffix/prefix rule — a suffix rule is a default-allow in disguise.
- NO net capability change. Report the derived mapping and every value that could not be confidently mapped.

Related: #4495 (shared frontmatter reader) proposes a "clean scalar copy" of `role`; that is exactly the pass-through this issue exists to prevent, and the normalization added here is what that work should reuse.

Owner-authorized 2026-07-31.

bobmatnyc:
Closes #4502. Follows #4499.

## What was there

`agents::claude_mpm_loader::ClaudeMpmFrontmatter` never extracted any role field, and `to_agent_config` hardcoded it:

```rust
role: "agent".to_string(),
```

So every claude-mpm agent's real domain was discarded at load time, and `"agent"` matched nothing in `ASSISTANT_ALLOWED_DELEGATE_ROLES`.

## Two findings that differ from the brief — please read

**1. The deployed artifacts do not carry `role:` at all.** trusty-mpm's composer emits `agent_type:`; `role:` exists only in the SOURCE assets (`crates/trusty-mpm/src/assets/agents/*.md`) which this loader never sees. Of ~40 files in a real `~/.claude/agents/`, exactly one carries `role:` (a hand-authored agent). Reading `role` alone would have read a key that is absent from essentially every artifact this loader actually loads. So this reads `role` **and** falls back to `agent_type` — which, note, `ClaudeMpmFrontmatter` was already parsing and discarding under `#[allow(dead_code)]`.

**2. The fine-grained per-stack role vocabulary described in the brief does not exist.** trusty-mpm's roles are already coarse. `role: engineer` is declared by **19** agents (`rust-engineer.md`, `dart-engineer.md`, `python-engineer.md`, … all declare plain `engineer`), `qa` by 4 (including `code-critic.md` and `api-qa.md`), `ops` by 4. There is no `dart-engineer`/`api-qa`/`code-critic` role value anywhere. The trap is real but differently shaped: the mismatch is `research` vs `researcher`, plus a tail of specialist and non-domain values.

## Observed vocabulary → derived mapping

Built from `grep '^role:' crates/trusty-mpm/src/assets/agents/` (42 files) and `grep '^agent_type:'` across deployed artifacts.

| declared value | count | normalized to | note |
|---|---|---|---|
| `engineer` | 19 src / 19 dep | `engineer` | identity |
| `qa` | 4 / 4 | `qa` | identity |
| `ops` | 4 / 5 | `ops` | identity |
| `documentation` | 1 / 2 | `documentation` | identity |
| `ticketing` | 1 / 0 | `ticketing` | identity |
| `research` | 1 / 2 | `researcher` | same domain, this crate's spelling (`research-agent.toml`) |
| `data-engineer` | 1 / 0 | `engineer` | an engineer that writes pipeline code |
| `planner` | 0 / 0 | `planner` | not observed; included so the coarse word round-trips |

### Could not confidently map — left fail-closed, and why

| value | why not mapped |
|---|---|
| `security` | real specialist, no coarse counterpart. `qa`? `researcher`? That is an owner's judgement, not a translation |
| `version-control` | same — git operations are not `ops` in this crate's sense |
| `code-analyzer` | read-only analysis; plausibly `researcher`, but plausibly is not good enough for a security discriminator |
| `memory-manager` | no counterpart |
| `mpm-agent-manager`, `mpm-skills-manager` | no counterpart |
| `base`, `base-engineer`, `base-ops`, `base-qa`, `base-research` | **composition fragments**, appended into other agents and never dispatched. Ineligible is CORRECT here, not a gap |
| `universal`, `system`, `trusty-mpm` | not domains |
| `engineer\|qa\|ops\|universal\|documentation` | a literal template placeholder found in a real artifact's frontmatter |

## Fail-closed design

- Unknown / blank / absent → `UNMAPPED_ROLE` = `"agent"`, **the same literal the loader hardcoded before**, so the fallback path is byte-identical to today for everything the table cannot classify. `unmapped_sentinel_is_not_an_allowlisted_role` proves it is not in the allowlist and fails if that constant is ever widened to include it.
- An **explicit table**, never a suffix rule. `*-engineer → engineer` would be a default-allow in disguise — it admits any future `<anything>-engineer` sight unseen. `data-engineer` is listed as a value for exactly this reason.
- `role` → `agent_type` is a **preference, not a union**. A declared-but-unmappable `role` does NOT get a second chance through `agent_type`; searching keys until one is eligible is a widening dressed as a fallback (`declared_but_unmappable_role_does_not_fall_through_to_agent_type`).
- `every_mapped_target_is_an_allowlisted_role` catches the opposite bug — a typo on a right-hand side producing a role that is ineligible *by accident* rather than by design.

## No assistant gains a reachable target — verified

Role eligibility is only a coarse **pre-filter**. `tools::delegate::DelegateToAgentTool::execute` computes

```rust
let whitelist_applies = self.allowed_target_roles.is_some()
    || self.delegator_role.as_deref().is_some_and(is_assistant_kind);
```

so the role allowlist and the **name floor** are always applied together — there is no path on which a role check runs alone. The floor is `ASSISTANT_REACHABLE_SUBAGENTS = ["research-agent", "ticketing-agent"]`, checked **first** in `SubagentAllowSet::resolve`, and both names resolve to bundled TOMLs ahead of the claude-mpm fallback tier in `AgentConfig::by_name`. No claude-mpm agent can occupy either name, so **no bundled assistant gains any new reachable target**. The non-assistant delegators (`pm`/`ctrl`-as-orchestrator) never had a role gate at all, so nothing changes for them either.

`ASSISTANT_REACHABLE_SUBAGENTS`, every persona's `[subagents].delegate_allowed`, and every `[tools].allow` are untouched.

## Relationship to #4495

#4495 proposes a "clean scalar copy" of `role` onto `AgentConfig` via the shared frontmatter reader. That is precisely the verbatim pass-through this PR exists to prevent. `claude_mpm_role::normalize_role` is deliberately a standalone, loader-independent function so that work can call it rather than deriving a second, divergent mapping.

## Tests

13 new: 10 in `agents::claude_mpm_role::tests` (sentinel/target allowlist invariants, the full observed vocabulary including everything that must NOT map, preference order, no-fall-through, case/whitespace, the reachability claim) and 3 in `claude_mpm_loader::tests` driving the real parse + projection path with a deployed-shaped artifact. No existing test was deleted or weakened.

## Gates

`cargo check` · `cargo test -p trusty-agents` (3127 passed, 0 failed) · `cargo clippy -p trusty-agents --all-targets -- -D warnings` · `cargo fmt --check` · `check_line_cap.sh` · `check_test_pointers.sh` — all clean.


bobmatnyc:
**Rebased onto #4498, which changed the picture — one addition.**

#4498 landed `agents::mpm_bridge` while this PR was in CI, and its role projection is a **verbatim copy**:

```rust
role: meta.role.unwrap_or_else(|| "agent".to_string()),
```

That is precisely the pass-through this PR exists to prevent, in a second module. It is currently **dormant** — `mpm_bridge` is declared in `agents/mod.rs` but nothing calls it (#4495 tracks wiring it into `AgentRegistry::load` as a follow-up) — so nothing is live today, but it would go live the moment that follow-up lands.

`project_mpm_agent` now routes through the same `claude_mpm_role::normalize_role`. `AgentMetadata` carries no `agent_type` field, so only `role` is available there; since the deployed artifacts declare `agent_type` and no `role`, every one of them resolves to the fail-closed sentinel — byte-identical to the literal it replaced, and the existing `projects_clean_scalars` test (which asserts `role == "engineer"` from a `role: engineer` artifact) still passes unchanged because that mapping is an identity.

New test: `role_is_normalized_and_fails_closed` drives `load_mpm_agent` with a `role: security` artifact and asserts it lands on the sentinel and is not in `ASSISTANT_ALLOWED_DELEGATE_ROLES`.

Full suite after the rebase: 3145 passed, 0 failed.

Issue #4448, opened 2026-07-31
deployer: quarantine untracked project-tier files that shadow bundled agent stems


bobmatnyc:
Untracked project-tier agent files that collide with a bundled agent filename stem are skipped by the deployer (WARN "untracked by the deploy manifest ... were skipped", firing on 8 separate days through 2026-07-30) and never removed by retraction, so stale pre-trusty-mpm (claude-mpm-era) agents permanently shadow framework agents.

Post-#4437, the framework no longer deploys to workspace `.claude/agents/` and `retract_framework_agents` removes only manifest-tracked entries — so in a workspace with no `.trusty-mpm-manifest.json` these files are unreachable by any tm operation.

Live evidence 2026-07-30: 4 of 16 live sessions affected; worst is a workspace with 7 framework stems (documentation, engineer, ops, qa, research, ticketing, web-qa) occupied by claude-mpm-era files whose `name:` frontmatter is invalid under the harness's lowercase-hyphen rule.

**Expected**: a bundled-agent stem in the project tier is either manifest-tracked and refreshed, or quarantined.

**Proposed fix**: at the existing skip-and-warn call site in `crates/trusty-agents-common/src/agents/deployer.rs`, when the untracked file's stem matches a bundled agent name, rename it to `<name>.md.disabled` and WARN naming the session (quarantine, not delete).

Root-cause diagnosis: session 2eb72dca fleet audit 2026-07-30; mechanism isolated against #4408 (files never manifest-tracked, freeze fix inapplicable) and #4437 (retraction no-ops with no manifest; zero post-install mutations observed).

bobmatnyc:
Manual quarantine ran 2026-07-31 against the legacy claude-mpm-era agent files this issue describes, across the 3 live workspaces where they were shadowing bundled stems.

**Per-workspace counts** (backed up byte-identical before rename, nothing deleted, reversible with `mv`):
- `duettoresearch/code-intelligence` — 12 files renamed `<stem>.md` → `<stem>.md.disabled` (audit had predicted 7 — **+5** discrepancy)
- `bobmatnyc/writing` — 5 files renamed (matched the audit prediction)
- `duettoresearch/cto-reports` — 5 files renamed (audit had predicted 4 — **+1** discrepancy)
- `duettoresearch/duetto-agents-skills` — checked, not affected: 0 renamed. Its agents are all `duetto-*` and collide with nothing.

[...]

Issue #4443, opened 2026-07-31
feat(trusty-mpm): custom deployment mechanism for trusty-built project agents


bobmatnyc:
Second half of the #4409 architecture ruling: project `.claude/agents/` is reserved for user-added and custom trusty-built agents, but the deployment mechanism for the latter doesn't exist yet (tm-agent-manager delegable). Reference: #4409's ruling comment.

bobmatnyc:
Design constraints found during scoping:

#4443 must forbid or loudly warn on filename-stem collisions with the canonical bundled roster. Project tier still outranks \`\$CLAUDE_CONFIG_DIR/agents/\` in resolution, so a custom agent named e.g. \`rust-engineer\` would permanently shadow the bundled one — that is #4408 reintroduced by design.

Second constraint: #4443's new \`Origin\` variant must NOT be classified as framework-owned by \`is_framework_owned\` (\`deployer.rs\` ~539-546). That check runs on every session launch; a copy-paste from \`Origin::Bundled\` would silently DELETE the user's custom agents on every launch, with no signal beyond a shrinking file count.

Sequencing: #4443 is last in the stack, after #4442 and #4448.

bobmatnyc:
Milestone triage 2026-08-03: this issue may be obsoleted by the 2026-08-01 "collapse the agent tiers" ruling (single deploy target plus source priority; Claude Code resolves agents on the frontmatter `name`, not the file path). A project-tier custom deployment mechanism may no longer be the right shape. Leaving the milestone unchanged pending an owner ruling on whether to close as superseded or re-scope.

bobmatnyc:
**Closed as superseded** — owner ruling 2026-08-03.

The 2026-08-01 "collapse the agent tiers" ruling supersedes this. Under it there is a single deploy target plus source priority, and Claude Code resolves agents on the frontmatter `name` rather than the file path — so a project-tier custom deployment mechanism is no longer the right shape. See ADR-0025 and its 2026-08-03 manifest-based-configuration addendum for the model that replaces it.

Superseded, not abandoned: the provenance concern this raised lives on in #4698 (provenance frontmatter field), and the deployer-shadowing concern in #4448.

PR #4515, opened 2026-08-01
feat(trusty-mpm): tm doctor flags tm-owned agents deployed outside the canonical tier (#4442)

### LINKED ISSUE #4442: tm doctor: flag tm-owned assets deployed outside the canonical user tier
Deferred scope item 3 from #4409 (see PR #4437 body: boundary check deliberately left out). `tm doctor` should detect bundled/framework-owned agent copies living in workspace `.claude/agents/` or `~/.claude/agents/` now that the canonical tier is `$CLAUDE_CONFIG_DIR/agents/`. References: #4409, #4437.

bobmatnyc:
Closes #4442

## The gap

Every existing agent probe — `agents` (presence), `deployment` (roster diff), `agent_reachability` (#4451, tier reachability) — looks only at the canonical deploy destination, `$CLAUDE_CONFIG_DIR/agents/`. None of them can fail when a stale copy of the SAME agent sits in a project's `.claude/agents/`, and that copy is not inert: the project tier OUTRANKS the user tier, so it wins resolution. [#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408) is this exactly — a 32-byte project-tier stub beat the real 25 KB `rust-engineer`, delegation kept "succeeding" under the right name, and every check stayed green.

[#4409](https://github.com/bobmatnyc/trusty-tools/issues/4409) stopped tm writing per-workspace copies and retracts the ones its manifest tracks. An UNtracked copy — hand-placed, or written by a binary older than the ledger — is invisible to that retraction and lives forever.

## What this adds

`tm doctor` gains an `asset_tier` check (25 checks, was 24). **Read-only**: it reports. It never deletes, renames, or moves anything — that is [#4448](https://github.com/bobmatnyc/trusty-tools/issues/4448).

## The shared classifier — `crates/trusty-agents-common/src/agents/tier_audit.rs`

The ownership predicate is a new module in `trusty-agents-common`, NOT inline in the probe, because **#4448 consumes the identical logic**: this probe reports a file, that one quarantines it. If either side re-derives the predicate they drift, and the drift is asymmetric harm — doctor flags a file the repair refuses to touch (noise operators learn to ignore), or repair quarantines a file doctor never flagged (a project's own agent silently disappears). The module doc states that contract explicitly.

Surface:

- `agent_identity(content, file_name)` — the name a file resolves under.
- `bundled_agent_names(source_dir)` — the roster, read from the directory the deployer actually deploys from.
- `ownership_of(manifest, file_name) -> TierOwnership` — `Untracked` / `FrameworkOwned` / `UserOwned`.
- `classify_tier_resident(identity, ownership, roster) -> TierResidentClass` — pure, no I/O: `ShadowsBundled` / `StrandedFrameworkOwned` / `Custom`.
- `audit_agent_tier(dir, roster) -> Vec<MisplacedAgent>` — the read-only scan both consumers run.

Two invariants make the "repair eats a project's own agent" hazard structurally hard rather than merely documented:

**1. Identity is the frontmatter `name:`, not the filename.** Every consumer resolves agents by the declared name, falling back to the stem only when absent (`claude_mpm_loader::parse_agent_file`). A stem-keyed predicate is wrong in both directions — it misses `helper.md` declaring `name: rust-engineer` and falsely flags `rust-engineer.md` declaring `name: acme-custom`. The same `agent_identity` keys the roster AND the candidate, so the two sides cannot diverge. This matters inside the bundle too: `BASE-AGENT.md` declares `name: base-agent`.

**2. A user-owned ledger entry outranks a name collision.** Ownership is three-state, never a bool — collapsing "untracked" and "tracked as the operator's" throws away the only positive PROOF that a file is not tm's, and a name collision then condemns it. `retract_framework_agents` never touches an untracked or user-owned file; this module matches that, so #4448 cannot quarantine what the deployer would have preserved.

`Custom` is the **noise guard and the future-proofing seam**. A name tm does not ship is the project's own — hand-placed today, first-class once [#4443](https://github.com/bobmatnyc/trusty-tools/issues/4443) lands project-custom agents — and it is *classified*, not merely filtered out, so a new legitimate category is added as one variant here and BOTH consumers inherit the exclusion without a rewrite.

## Severity, and why

**Project tier → `Fail`. Home tier (`~/.claude/agents/`) → `Warn`.**

`Fail` for the project tier, deliberately not negotiable down:

1. Shadowing has no degraded-but-usable state. The wrong agent definition loads, silently, and every downstream symptom points somewhere else — #4408 cost a full debugging session before the stub was found. `Warn` is for degraded-but-working; this is broken-and-lying.
2. Precedent: `agent_reachability` (#4451) is a hard `Fail` for the same failure class — agents that exist but do not resolve.
3. Empirically, `Warn` on this exact directory is already known not to work: the deployer's skip-and-warn log fired on **eight separate days** through 2026-07-30 with zero operator follow-up. Filing a genuine shadowing at that level would bury it next to the signal people have already learned to scroll past.

`~/.claude/agents/` gets `Warn` because a managed session relocates `CLAUDE_CONFIG_DIR` and never reads that tier — those copies are stale, not shadowing. Only a plain `claude` run outside tm sees them.

An empty bundled roster would classify every shadowing copy as a legitimate custom agent — a false green in the very check that exists to stop false greens — so the probe unions the on-disk source with the agents compiled into the binary. The roster is never empty.

## Verification

```
$ cargo test -p trusty-agents-common
test result: ok. 364 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test -p trusty-mpm            # full surface, not --lib
test result: ok. 4127 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
(44 "test result: ok" lines across every target; 0 failures anywhere)

$ cargo clippy -p trusty-agents-common --all-targets -- -D warnings   → exit 0
$ cargo clippy -p trusty-mpm --all-targets -- -D warnings             → exit 0
$ cargo fmt --check                                                   → exit 0
OK   trusty-agents-common: changelog.d fragment present and valid
$ bash scripts/check_sld.sh
sld-lint: scanned 50 spec doc(s) + 2975 code file(s); 0 error(s), 0 warning(s)
```

The load-bearing cases: `project_tier_stub_on_a_bundled_name_fails` (the probe FIRES), `custom_project_agent_does_not_fire` and `user_owned_project_agent_on_a_bundled_name_does_not_fire` (it stays silent), `audit_flags_a_renamed_file_that_declares_a_bundled_name` / `audit_ignores_a_bundled_filename_that_declares_a_custom_name` (identity keys on `name`, both directions). Remove the check and the first one fails; move the user-owned guard below the roster check and the ownership tests fail (mutation-verified).

Changelog recorded as per-PR fragments (#4476), not hand edits: `crates/trusty-mpm/changelog.d/4442-doctor-asset-tier.md` and `crates/trusty-agents-common/changelog.d/4442-tier-audit.md`.


bobmatnyc:
## Code Critic Verdict: WARN

Adversarial review from an independent worktree at PR head `d76eb84b` (base `origin/main` = `140621a5`). Zero CRITICAL, two HIGH.

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-agents-common/src/agents/tier_audit.rs` | 146-158, 195-198 | `classify_tier_resident` takes `framework_owned: bool`, collapsing three distinct manifest states (untracked / tracked `Origin::User` / tracked `Origin::Bundled`) into one. Because the bundled-stem check runs first, a manifest entry that **positively proves the operator owns the file** is discarded whenever the stem collides with a bundled agent. | Take `Option<Origin>` (or a small `ManifestEvidence` enum) instead of a bool; return `Custom` for `Some(Origin::User \| Origin::Registry)` even on a bundled stem — or add a fourth variant that reports but is never quarantined. Pass `manifest.managed.get(&name).map(\|e\| e.origin)` at line 195. |
| HIGH | `crates/trusty-mpm/CHANGELOG.md`, `crates/trusty-agents-common/CHANGELOG.md` | 9-24 / 9-21 | Both crate `CHANGELOG.md` files were hand-edited with a new `## [Unreleased]` section. This repo migrated to `changelog.d/` fragments (#4476) and the gate says verbatim: *"Never edit the crate CHANGELOG.md by hand: release time assembles the fragments."* The **Per-PR changelog fragment** check is RED and `mergeStateStatus` is `BLOCKED`. | Revert both `CHANGELOG.md` hunks; add `crates/trusty-mpm/changelog.d/4442-doctor-asset-tier.md` and `crates/trusty-agents-common/changelog.d/4442-tier-audit.md` in the documented format. |
| MEDIUM | `crates/trusty-agents-common/src/agents/tier_audit.rs` | 113-127, 199 | The shadowing predicate keys on the **filename stem**, but every consumer resolves agents by the **frontmatter `name`** — `claude_mpm_loader.rs:206` (`fm.name.unwrap_or_else(file_stem)`) and `:292` (`out.insert(agent.name, …)`). Both directions are wrong: `helper.md` carrying `name: rust-engineer` genuinely shadows and is **not** flagged; `rust-engineer.md` carrying `name: acme-custom` does **not** shadow and **is** flagged (and would be quarantined by #4448 without fixing any shadowing). | Read the name through the existing shared reader (`mpm_bridge`, just wired up by #4509), falling back to the stem. At minimum state the proxy in the module doc and drop the "Shadowing is a property of the NAME" framing at `tier_audit_tests.rs:68-70` — as written it reads as the frontmatter name. |
| LOW | `crates/trusty-agents-common/src/agents/tier_audit.rs` | 124, 199 | `name.trim_end_matches(".md")` strips *repeated* suffixes, so `rust-engineer.md.md` yields stem `rust-engineer` while `Path::file_stem()` (what every loader uses) yields `rust-engineer.md`. `doctor_asset_tier.rs:110` uses the correct `strip_suffix(".md")` six lines away in the same PR. | Use `strip_suffix(".md")` in both places. (`deployer.rs:239` has the same pre-existing pattern — out of scope here.) |

### HIGH-1 evidence (reproduced, not inferred)

Added one probe to `tier_audit_tests.rs` in a scratch worktree and ran it:

```
CRITIC PROBE -> [MisplacedAgent { path: ".../qa.md", stem: "qa", class: ShadowsBundled }]
thread '...CRITIC_PROBE_user_owned_entry_on_a_bundled_stem' panicked:
a manifest entry that PROVES the operator owns this file was ignored
```

The staged file is `qa.md`, tracked in that directory's manifest with `origin: User`, on a bundled stem. It classifies `ShadowsBundled` → `is_tm_owned() == true` → reported, and under #4448 renamed to `qa.md.disabled`.

Why this is HIGH rather than a nit:

1. It contradicts this module's own stated contract. `tier_audit.rs:28-33` says `Custom` is "the deliberate exclusion seam" against "a project's own agent silently disappears". The seam is stem-only, so it does not cover the one case where positive ownership evidence actually exists.
2. It diverges from the predicate the doc claims to reuse. `tier_audit.rs:137-140` cites `Origin::is_framework_owned` as "the same predicate the deployer and `retract_framework_agents` use" — but `deployer.rs:23-31` states retraction "never touches an untracked or user-owned file", while this classifier calls both tm-owned. The divergence is real and undocumented.
3. The existing test does not cover it. `tier_audit_tests.rs:146-154` (`audit_ignores_a_user_owned_manifest_entry`) uses `mine.md` — a **non-bundled** stem — so the collision path is untested. `classify_prefers_shadowing_over_stranded` (line 68) pins the override for `Origin::Bundled`, which is correct, and the `Origin::User` case rides along on the same branch by accident.

Read-only, this is a false `Fail` on an operator's own agent. As the shared predicate for #4448 it is the data-loss path this PR's design argument exists to prevent — which is exactly why it should be fixed *before* #4448 consumes it, not after.

## Required Changes

1. Give `classify_tier_resident` the full manifest evidence (`Option<Origin>`) so a tracked user-owned file on a bundled stem can be excluded, and add the missing test (user-owned entry **on a bundled stem**).
2. Replace the hand-edited `CHANGELOG.md` sections with `changelog.d/` fragments so the per-PR changelog gate goes green.
3. Either resolve the filename-stem vs frontmatter-`name` mismatch, or document it explicitly as a deliberate proxy and correct the misleading "property of the NAME" comment.

## Verified clean (attacked, found nothing)

- **Empty/partial-roster guard holds.** `bundled_roster` unions `agent_source_dir()` with `bundle::ALL`'s 42 `agents/*.md` entries. On a binary-only install the on-disk source is *populated by* `ALL`, so on-disk ⊆ `ALL` and the union is never partial, only ever a superset. A missing, unreadable, or half-populated source dir degrades to the compiled 42, never to empty. `roster_falls_back_to_the_embedded_bundle` pins it.
- **Severity `Fail` blocks nothing.** `tm doctor` prints the aggregate and returns `Ok(())` (`commands/misc.rs:191-210`); nothing in install, session launch, or CI reads `report.overall` for a gate — the only other consumers are the Slack and Telegram formatters. A new `Fail` cannot regress a working flow.
- **Read-only claim verified.** `AgentManifest::load_checked` (`manifest.rs:317-327`) is a pure read that returns `default()` on missing/unreadable. No `with_agent_manifest_lock` is taken, so not even a lock sidecar is created. Nothing in the diff writes.
- **Non-recursive scan is correct, not a gap.** `claude_mpm_loader::load_from_dir` (`:260-297`) is itself flat, so a nested `.claude/agents/backend/foo.md` is invisible to the loader too — no false negative.
- **Check count is genuinely 25.** Counted `run_doctor` directly: 15 in the initial `vec!` + 10 pushes. `doctor_checks_match_run_doctor_names` (`generate/doctor.rs:186`) is a real order-sensitive drift gate, and the regenerated `references/doctor.md` says 25 with `asset_tier` at position 4. The only residual "24" in the tree is the historical 1.3.1 CHANGELOG entry, which is correct.
- **`BASE-*.md` in the roster is correct, not noise.** `deploy_agents_locked` (`deployer.rs:233-248`) enumerates every `.md` in the source with no `BASE-` filter, so those five stems really are deployed.
- **The tests fail without the check.** Removing it breaks both `project_tier_stub_on_a_bundled_stem_fails` and `doctor_checks_match_run_doctor_names`. Locally: `cargo test -p trusty-agents-common --lib agents::tier_audit` → `14 passed; 0 failed`; `cargo test -p trusty-mpm --lib doctor_asset_tier` → `12 passed; 0 failed`.
- **Remediation command is real.** `tm install --reset-agents --reset-agents-workspaces` parses and the flag pairing is enforced (`tests_behavior_reset_agents_tests.rs:187-213`).

## Notes

- Doc quality against the Why/What/Test standard is genuinely strong for a cross-crate surface — the CONTRACT block at `tier_audit.rs:17-33` is the right thing to write. HIGH-1 is precisely that the code does not yet implement what that block promises.
- `verdict()` discriminates tiers by `s.label == "project"` string comparison (`doctor_asset_tier.rs:157`). Not flagged — the label set is a two-element local literal — but a `TierKind` enum would make it unbreakable if a third tier is ever added.
- Not flagged (below 80% confidence): case-insensitive macOS filenames (`Rust-Engineer.md`). It would classify `Custom` and be missed, but such a stem is invalid under the harness's lowercase-hyphen name rule, so the real-world shadow is unlikely.

Would I merge as-is: **no**. HIGH-2 blocks the merge mechanically; HIGH-1 should land before #4448 consumes the classifier as a move-this-file predicate.

bobmatnyc:
All four review findings addressed in `ed3fa9cf` (force-pushed over `d76eb84b`).

## HIGH-1 — the classifier no longer discards ownership proof

`framework_owned: bool` is gone. Ownership is now `TierOwnership`, a three-state value read in one place by `ownership_of()`:

| state | meaning |
|---|---|
| `Untracked` | no ledger entry |
| `FrameworkOwned` | tracked, `Origin::is_framework_owned()` |
| `UserOwned` | tracked, `Origin::User` or `Origin::Registry` |

`classify_tier_resident` checks `UserOwned` **first** and returns `Custom`, so a ledger entry proving the operator owns the file survives a bundled-name collision. That matches `retract_framework_agents`, which never touches an untracked or user-owned file — the divergence you flagged is closed.

Two tests cover the reproduction, both keyed on a **bundled** name:
- `classify_user_owned_survives_a_bundled_name_collision` (pure)
- `audit_ignores_a_user_owned_entry_on_a_bundled_name` (end to end)
- `user_owned_project_agent_on_a_bundled_name_does_not_fire` (through the doctor probe)

Mutation-verified — moving the `UserOwned` guard back below the roster check makes the first two fail:

```
test agents::tier_audit::tests::classify_user_owned_survives_a_bundled_name_collision ... FAILED
test agents::tier_audit::tests::audit_ignores_a_user_owned_entry_on_a_bundled_name ... FAILED
  left: ShadowsBundled
```

The old `audit_ignores_a_user_owned_manifest_entry` (non-bundled stem) is kept as the weaker sibling case.

## MEDIUM — identity is the frontmatter `name`, not the stem

Resolved on the same key the loader uses. `agent_identity(content, file_name)` returns the trimmed frontmatter `name:` when non-empty, else the stem — exactly `claude_mpm_loader::parse_agent_file`'s rule — read through `agent_metadata_from_str` (trusty-mpm's own grammar, the one #4509 settled on for this tier).

It is applied to **both sides** of the comparison, so roster and candidate cannot key differently: `bundled_agent_stems` became `bundled_agent_names`, and the probe's embedded-bundle half now runs `agent_identity` over `artifact.contents` instead of splitting `rel_path`. That last one was a live bug in its own right — `BASE-AGENT.md` declares `name: base-agent`, so the rel_path-keyed half was seeding the roster with a name nothing resolves by (`roster_keys_the_embedded_half_by_declared_name`).

Both directions tested:
- `audit_flags_a_renamed_file_that_declares_a_bundled_name` — `helper.md` declaring `name: rust-engineer` now fires (the miss)
- `audit_ignores_a_bundled_filename_that_declares_a_custom_name` — `rust-engineer.md` declaring `name: acme-custom` does not (the false positive)
- plus `a_renamed_project_file_declaring_a_bundled_name_fails` through the probe

No limitation to document — reading frontmatter here was not infeasible.

## HIGH-2 — changelog convention

Both hand edits reverted; `CHANGELOG.md` in both crates is now byte-identical to `origin/main`. Replaced with per-PR fragments:

- `crates/trusty-mpm/changelog.d/4442-doctor-asset-tier.md`
- `crates/trusty-agents-common/changelog.d/4442-tier-audit.md`

```
OK   trusty-agents-common: changelog.d fragment present and valid
```

Also previewed with `assemble-changelog.sh trusty-mpm --stdout` to confirm the fragment lands under `### Added` in the real release output.

## LOW — `strip_suffix`

Both `trim_end_matches(".md")` call sites are gone (`agent_identity` is the only place a stem is derived now). Pinned by `identity_strips_only_one_md_suffix`: `weird.md.md` → `weird.md`.

## Gate

```
$ cargo test -p trusty-agents-common
test result: ok. 364 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test -p trusty-mpm          # full surface, not --lib
test result: ok. 4127 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
(44 "test result: ok" lines total across every target; 0 failures anywhere)

$ cargo clippy -p trusty-agents-common --all-targets -- -D warnings   → exit 0
$ cargo clippy -p trusty-mpm --all-targets -- -D warnings             → exit 0
$ cargo fmt --check                                                   → exit 0
$ bash scripts/check_changelog_fragment.sh                            → exit 0
```

No `#[ignore]`, no cfg-gating. 23 classifier tests + 15 probe tests, all passing.

bobmatnyc:
## Verdict: APPROVE (round 2)

Re-reviewed at head `ed3fa9cf` in an independent worktree. All four round-1 findings are genuinely fixed — verified by mutation and by re-running the suites myself, not by reading the claims.

### Round-1 findings — verification

| # | Round-1 finding | Status | How I verified it |
|---|---|---|---|
| HIGH-1 | `framework_owned: bool` collapses "untracked" and "operator-owned" | **FIXED** | Ran the mutation. Moving the `UserOwned` guard below the roster check in `tier_audit.rs:243-248` fails **exactly** the two named tests: `classify_user_owned_survives_a_bundled_name_collision` (`ShadowsBundled` != `Custom`) and `audit_ignores_a_user_owned_entry_on_a_bundled_name`. 21 passed / 2 failed. Reverted; 23/23 green. |
| HIGH-2 | Hand-edited `CHANGELOG.md` files | **FIXED** | `git diff origin/main -- 'crates/*/CHANGELOG.md'` is empty. Both `changelog.d/` fragments present. "Per-PR changelog fragment" check green at head. |
| MEDIUM | Identity keyed by stem, not frontmatter `name:` | **FIXED, both sides** | Mutated each side independently. Roster→stem fails `bundled_names_use_the_frontmatter_name`; candidate→stem fails `audit_flags_a_renamed_file_that_declares_a_bundled_name` **and** `audit_ignores_a_bundled_filename_that_declares_a_custom_name` — the miss and the false positive, both pinned. The `BASE-AGENT.md` bug was real: the file declares `name: base-agent`, and it *is* in the deploy set (`deploy_agents_locked` enumerates every `.md`, and the live tier at `~/.trusty-tools/trusty-mpm/claude-config/agents/` contains `BASE-AGENT.md`). |
| LOW | `trim_end_matches(".md")` | **FIXED** | No occurrence in any of the four new files. `agent_identity` uses `strip_suffix`, pinned by `identity_strips_only_one_md_suffix` (`weird.md.md` → `weird.md`). |

### The deeper questions asked of round 2

**Is `UserOwned`-first genuinely correct, or the opposite hole?** Correct, and not a new attack surface. No production code writes `Origin::User` or `Origin::Registry` — every writer (`deployer.rs:320,385`, `agent_reset.rs:231,265`, `expected_set.rs:286`) emits `Origin::Bundled`. A `user` entry can only appear by hand-editing `.trusty-mpm-manifest.json`, which lives *in the same directory* as the agent file it describes: anyone who can forge the entry can already write the agent. The ledger is not a trust boundary here, and honouring it matches what `retract_framework_agents` and `deploy_agents` already do for the seed-once tier.

**Is `Registry` → `UserOwned` conservative in the safe direction?** Yes. Folding it into `UserOwned` yields `Custom` — never reported, never moved by #4448. It errs toward preserving the operator's file, identical to `Origin::is_framework_owned` (`manifest.rs:237`), which the deployer and retraction already branch on.

**As a mutation decision for #4448.** I hunted both error directions. `Custom` on a genuinely-shadowing file only happens when the ledger positively records operator ownership — deliberate. `ShadowsBundled` on an operator-owned file only happens for an *untracked* file resolving to a bundled name, which is the feature (it is #4408 verbatim: an untracked copy is invisible to retraction and lives forever). The roster is built by the same enumeration rule as `deploy_agents_locked`, so it can't drift from the deploy set, and `bundled_roster` unions the embedded bundle so an empty roster can never produce a false green.

**Identity edge cases.** Malformed/absent frontmatter → `AgentMetadata::default()` → stem, documented and tested. Quoting and whitespace agree with strict YAML: `parse_kv_line` (`frontmatter.rs:63`) trims and strips one balanced quote pair. Manifest lookup is keyed by filename and the roster by identity — correct, since the ledger is filename-keyed.

### Evidence

```
test result: ok. 364 passed; 0 failed; 0 ignored

cargo test -p trusty-mpm
test result: ok. 4127 passed; 0 failed; 6 ignored
test result: ok. 1363 passed; 0 failed; 1 ignored   (x2 bins)
... 20 further targets, all ok, 0 failed
```

Live check rollup at `ed3fa9cf`: **24 success, 0 failure, 0 pending** (1 skipped — "Notify on red main"). No workspace reference to the removed `bundled_agent_stems` or `framework_owned:` survives.

### Findings

| MEDIUM | `crates/trusty-agents-common/src/agents/tier_audit.rs` | 42-45 | The invariant-2 safety argument overstates parity with the deployer: *"`retract_framework_agents` never touches an untracked or user-owned file; **this module matches that**, so #4448 cannot quarantine what the deployer would have preserved."* It matches for user-owned, not for untracked — `classify_tier_resident("qa", Untracked, {"qa"})` returns `ShadowsBundled`, pinned by `classify_bundled_name_shadows` and `audit_reports_a_shadowing_stub`. The behavior is right (catching untracked shadows is the whole point); the sentence is not, and it is the paragraph a #4448 implementer will read as their data-loss safety proof. Same claim is repeated in `crates/trusty-agents-common/changelog.d/4442-tier-audit.md`. | Narrow the clause to what holds: user-owned parity is exact; untracked is *deliberately broader* than retraction, because an untracked shadow is invisible to retraction and is the #4408 case — so #4448 must quarantine (reversibly), never delete, and must confirm before moving an untracked file. |

Zero CRITICAL, zero HIGH → **APPROVE**. This is round 2 and the finding above is a doc-accuracy correction with no behavioral consequence in a read-only probe. **I would not hold the branch for a third round** — merge as-is and fold the wording correction into #4448, where the sentence actually becomes load-bearing.

Issue #2471, opened 2026-07-11
chore(trusty-mpm): DaemonClient has no per-request timeout — a hung (accept-but-never-respond) daemon freezes TUI input


bobmatnyc:
Follow-up from PR #2469's review (LOW, pre-existing since the #2118 skeleton — not introduced by #2469, which added a third sequential await to the same chain). client/http_client/mod.rs:55-57: DaemonClient::new configures no reqwest timeout; the tm projects TUI's poll steps are sequentially awaited in the same task that reads keyboard input (tui/project_ctl/mod.rs:139-142), so a daemon that accepts TCP but never responds freezes the whole TUI including q/Ctrl-C until the OS socket timeout. Fix direction: bounded per-request timeout on DaemonClient (reqwest::Client::builder().timeout(...)) shared by all TUI screens, and/or race poll steps against the input event loop / a cancel channel. Refs #2118, #2119, PR #2469.

Issue #3007, opened 2026-07-18
twin P1: WorkstreamConnector trait + tm/tcode backends (DOC-42 Phase 1)


bobmatnyc:
Part of the engineering-lead / virtual-twin architecture — **DOC-42** (PR #3006), epic #2109.

**Goal:** both tm and tcode expose equivalent, trait-based session control.
- Define `WorkstreamConnector` trait (create/status/send/attach/delegate) in `trusty-agents-common`.
- tm backend: wrap existing `SessionManager`/`SessionProxy` (MCP+HTTP) — `crates/trusty-mpm/src/client/proxy.rs`.
- tcode backend: wrap tcode JSON-RPC session protocol — `crates/trusty-code/src/session/protocol.rs`. **[tcode session owns the tcode-backend portion]**
- Integration-test both (mock + real sessions).

Size: M. Prerequisite: tcode REST parity (#2983). See DOC-42 §8 Phase 1.

PR #4516, opened 2026-08-01
fix: replace load-sensitive fixed waits for spawned processes with real readiness bounds (#4488)

### LINKED ISSUE #4488: Two load-sensitive test flakes: fixed-duration waits for a spawned process
**Defect.** Two tests wait a fixed wall-clock duration for a spawned process to become ready. Both pass on CI's single-tenant runner and fail locally under concurrent load. Same anti-pattern as the telegram backoff bound fixed in #4482 — a timing assumption treated as deterministic.

**1. `connectors::tm::tests::create_session_full_lifecycle`** (trusty-mpm). Fails with a `Transport` error to a spawned local server.
Confirmed PRE-EXISTING, not caused by any in-flight PR: reproduced by running the **pristine `origin/main` test binary** at load 52.85, with no branch code present. Independently observed failing during PR #4466 work and ruled causal-negative there on the same grounds.

**2. `trusty-agents/tests/api_e2e.rs`** — `support/api_server.rs:94` waits `Duration::from_secs(5)` for a spawned binary to come up. Red 3/3 locally at load 22–38, green in CI.

**Resolution.** Replace fixed-duration waits with condition polling against a readiness signal (bounded retry until the endpoint answers, with a generous ceiling), so the test tolerates load instead of assuming a latency budget. Per the standing rule for this sweep: no `#[ignore]`, no cfg-gate, no simply-raise-the-timeout. Verify with an A/B run under synthetic load, not a single green run.

**Environmental caveat for whoever measures this.** Observed local flake rates in this cluster are partly environmental and will mislead an A/B that does not control for load. Two contributors measured the same day: 77 simultaneous cargo processes across sibling agent worktrees, and leaked `cargo test --no-run` artifacts wired as hook commands so every hook event ran a 102-test binary twice (#4485, since cleaned). Control the load or interleave the A/B, as #4482 did.

**Related.** #4407 (flake cluster), #4482 (the sweep that surfaced these), #4485 (hook-wired test binaries inflating load).

bobmatnyc:
Closes #4488

Two tests decided readiness on a wall-clock budget instead of a signal. Both passed on CI's single-tenant runner and failed locally under concurrent load.

## 1. `connectors::tm::tests::create_session_full_lifecycle` (trusty-mpm)

**Root cause — a production bug, not a test bug.** `TmConnector` built its client from `default_client()` and never overrode the per-request bound, so `POST /api/v1/sessions/managed` inherited `DEFAULT_REQUEST_TIMEOUT`'s 10s. That route is where the daemon runs `WorkspaceProvisioner::provision` **synchronously** — git clone, worktree add, agent deploy, harness spawn. `DaemonClient::spawn_managed_session` has opted the identical route into the 180s `PROVISION_REQUEST_TIMEOUT` since #2471; the connector was simply never updated to match.

Every caller's success was therefore contingent on the machine being idle. The hermetic test measured **9.23–9.54s against the 10s cliff**. Past it the client aborted mid-provision — surfacing as the `Transport` error the issue reports — while the daemon kept working and orphaned the session it had already started.

`create_session` now carries the same override, with the constant shared rather than duplicated so the two call sites cannot drift.

**The flake is now caught deterministically rather than by chance.** `TmConnector::with_client` (mirroring `DaemonClient::with_client`) lets `create_session_outlives_the_default_request_timeout` drive a stub daemon that answers at 400ms behind a 100ms client-level bound and assert *success* — which only the per-request override can deliver, at any load. `list_sessions_is_bound_by_the_client_level_timeout` is the negative control: it proves the override stays scoped to the provisioning route rather than un-bounding the connector and reintroducing #2471.

### A/B evidence (pre-fix arm vs post-fix arm, same tree)

```
--- fix removed ---
test connectors::tm::tests::create_session_outlives_the_default_request_timeout ... FAILED
panicked at crates/trusty-mpm/src/connectors/tm_tests.rs:188:52:
create_session must apply the provisioning timeout override, not the client-level default:
  Transport("error sending request for url (http://127.0.0.1:64652/api/v1/sessions/managed)")

--- fix restored ---
test connectors::tm::tests::create_session_outlives_the_default_request_timeout ... ok
```

Full pre-fix lib suite at load 282 reproduces the same `Transport` failure; post-fix suite at load 293 is green (`4114 passed; 0 failed` in 77.01s).

Note honestly: the *original* `create_session_full_lifecycle` did **not** go red under my synthetic load — it landed at 9.23s, 0.77s from the cliff. Its pass/fail is decided by sub-second machine noise, which is exactly why the deterministic regression test above is the real fix to the detection problem.

## 2. `trusty-agents/tests/support/api_server.rs`

`ApiServer::spawn` already *polled* `/api/health` — but under a 5s ceiling, which is a bet on the machine being idle. Three changes make a generous ceiling free rather than slow:

- a child that exits fails immediately via `try_wait`, with its status, instead of sitting out the ceiling and reporting an anonymous timeout;
- each probe is individually bounded, so one stalled request cannot consume the whole wait and turn a retry loop back into a single fixed wait;
- the child's stdout/stderr are **drained continuously**. Nothing read those pipes before, so a chatty startup could fill the buffer and block the child before it ever bound the port. Draining removes that hang and puts the child's own output in every failure message.

`wait_for_task` likewise retries a failed poll instead of treating one transport hiccup as terminal — the condition is "the task left `running`", and a failed GET says nothing about it.

### A/B evidence (5s ceiling vs 60s ceiling, interleaved, same binary otherwise)

Synthetic load: 32 CPU spinners + 6 disk-churn loops + fork/exec pressure, load average **277**.

```
########## round 1 (load: 277.70) ##########
--- OLD (5s ceiling) ---
test result: FAILED. 0 passed; 4 failed; 3 ignored; finished in 5.19s
  spawn api server: api server did not answer http://127.0.0.1:50274/api/health
  within 5s (child still running)
--- NEW (60s ceiling) ---
test result: ok. 4 passed; 0 failed; 3 ignored; finished in 1.10s
```

5 interleaved rounds: **OLD pass=4 fail=1, NEW pass=5 fail=0**. The old arm failed at exactly 5.19s — it hit the ceiling, not a crash, which the new diagnostics state outright ("child still running").

## Not done

No `#[ignore]`, no cfg-gate, no `--exclude`, no `--lib` narrowing, no retry-until-green. No test lost coverage; two gained it.

## Verification gate

| gate | result |
|---|---|
| `cargo test -p trusty-mpm` | EXIT=0 — 4114 passed, 0 failed (full surface, 44 binaries) |
| `cargo test -p trusty-agents` | EXIT=0 — 3148 + 21 passed, 0 failed |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | EXIT=0 |
| `bash scripts/check_line_cap.sh` | 7 allowlisted, 0 violations |
| `bash scripts/check_changelog_fragment.sh` | all crates with source changes recorded |

Changelog recorded as fragments per #4476: `crates/trusty-mpm/changelog.d/4488-connector-provision-timeout.md`, `crates/trusty-agents/changelog.d/4488-api-e2e-readiness-wait.md`.


bobmatnyc:
## Critic verdict: APPROVE

Zero CRITICAL, zero HIGH. 3 MEDIUM + 3 LOW, none blocking. Reviewed independently in a detached worktree off `origin/main` with the PR branch checked out; every claim below is backed by a mutation I ran myself, not by reading the PR body.

### Isolation matrix (each guard mutated ALONE, `cargo test -p trusty-mpm --lib`)

| # | Mutation | `create_session_outlives…` | `list_sessions_is_bound…` |
|---|---|---|---|
| baseline | none | ok | ok |
| A | drop `.timeout(PROVISION_REQUEST_TIMEOUT)` from `create_session` | **FAILED** — `Transport("error sending request for url (…/api/v1/sessions/managed)")` | ok |
| B | drop `.timeout(request_timeout)` from `build_client` (the #2471 regression) | ok | **ok — did not fire** |
| C | add `.timeout(PROVISION_REQUEST_TIMEOUT)` to `list_sessions` (blanket override) | ok | **FAILED** |

Mutation A reproduces your A/B evidence exactly. Mutation C proves the negative control is a real guard. Mutation B is finding M3.

### Attacks that came back clean

**Precedent (#2471) genuinely applies.** `daemon/api.rs:277-280` registers one route — `post(spawn_session).get(list_managed_sessions)` on `/api/v1/sessions/managed`. `DaemonClient::spawn_managed_session` POSTs `{base}/api/v1/sessions/managed` (`managed.rs:120`); `TmConnector::create_session` POSTs `{daemon_url}/api/v1/sessions/managed` (`tm.rs:214`). Same method, same path, same synchronous handler. Not superficially similar — identical.

**Blast radius of 10s → 180s is nil today.** `grep -rn 'TmConnector::'` across the workspace returns only doc references and `tm_tests.rs`. `TmConnector` is a public DOC-44 Phase-1 API (#3007) with zero in-repo production callers, so no interactive CLI, UI, or automated path can newly stall for 3 minutes. And the same bound already sits on an interactive path — `DaemonClient::spawn_managed_session` ← `client/executor/managed.rs` ← the `tm` CLI — accepted since #2471.

**The unreproduced headline flake.** You are right to flag it and right about the substitute. `create_session_full_lifecycle` is untouched and still runs the real handler; the new test adds a load-independent assertion on the specific bound rather than replacing it. Mutation A shows the bound is what decides the outcome, so the root cause is demonstrated at the mechanism level even though the 9.23s run was never driven red. That is the correct trade — a test whose verdict does not depend on machine speed.

**Coverage.** No `#[ignore]` added (the 3 ignored `api_e2e` tests are pre-existing `OPENROUTER_API_KEY` gates), no cfg-gate, no `--exclude`, no `--lib` narrowing, no raised sleep. Nothing deleted; two tests gained. `cargo test -p trusty-agents --test api_e2e` → `ok. 4 passed; 0 failed; 3 ignored`. Changelog filed as per-PR fragments per #4477; `scripts/check_changelog_fragment.sh` EXIT=0; no hand-edited `## [Unreleased]` bullet.

### Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-agents/tests/support/api_server.rs` | 322, 332 | `while let Ok(Some(line)) = lines.next_line().await` exits the drain task on the FIRST `io::Error`, silently. tokio returns `Err(InvalidData, "stream did not contain valid UTF-8")` for non-UTF-8 child output (`tokio-1.53.1/src/io/util/read_line.rs:82`). Once the task exits, the pipe is never read again — the child can block on a full 64KB buffer, which is the exact hang the field doc says is "removed entirely". Every later line also vanishes from the diagnostic buffer with no marker. | Match the result: on `Err(e)` append `[out] <drain error: {e}>` and keep reading (or drain raw bytes + `String::from_utf8_lossy`). Never let the loop end on an error. |
| MEDIUM | `crates/trusty-agents/tests/support/api_server.rs` | 270 | `wait_for_task_with_timeout` builds `reqwest::Client::new()` — no request timeout. The PR states the principle "each probe is individually bounded, so one stalled request cannot consume the whole wait" and applies it to `wait_for_health` via `READY_PROBE_TIMEOUT`, but not to the loop rewritten in the same commit. The 120s ceiling is only checked between polls, so a connected-but-silent server hangs the wait with no bound. Evidence from mutation B: with `build_client`'s `.timeout()` removed, `build_client_bounds_a_stalled_connection` and `top_level_default_client_is_bounded_against_a_stalled_daemon` ran past 10 minutes without completing — an unbounded reqwest client does hang indefinitely here. | `reqwest::Client::builder().timeout(Duration::from_secs(10)).build()`, mirroring `READY_PROBE_TIMEOUT`. (`submit_task_json:229` has the same unbounded client — pre-existing, worth the same treatment.) |
| MEDIUM | `crates/trusty-mpm/src/connectors/tm_tests.rs` | 194-201 | The negative control's stated rationale is false, proven by mutation B. The doc says "without it, the sibling test would also pass if someone 'fixed' #4488 by removing the client's timeout entirely". Removing `.timeout(request_timeout)` from `build_client` leaves BOTH tests passing — the control does not fire, and the sibling never consults the production default client either (it injects its own 100ms bound). What the control actually catches is a per-request blanket override (mutation C). The guard is real but strictly narrower than documented, and in a repo with a CI doc-pointer lint a doc naming a scenario the test cannot detect is a trap for the next maintainer. | Reword to "the override must stay scoped to the provisioning route — it must not be applied to point reads" and drop the #2471 / client-timeout-removal claim. That regression is already pinned by `config::tests::default_client_uses_default_bounds` and `build_client_bounds_a_stalled_connection`. |
| LOW | `crates/trusty-mpm/src/connectors/tm_tests.rs` | 212 | `matches!(err, ConnectorError::Transport(_))` cannot tell a timeout from a decode failure — `decode_err` (`tm.rs:143-145`) also yields `Transport`. Meaningful today (mutation C fires), but a future change to the stub's list body would make the control pass on a decode error while asserting nothing about the bound. | Assert on elapsed `< 400ms`, or that the message does not start with `failed to decode daemon response`. |
| LOW | `crates/trusty-agents/tests/support/api_server.rs` | 82 | `output` grows without bound for the server's whole lifetime (up to 120s of a chatty debug binary). | Cap it, retaining head + tail. |
| LOW | `crates/trusty-agents/tests/support/api_server.rs` | 268 | `wait_for_task_with_timeout` takes `&self`, so it cannot use the `try_wait` fast-fail `wait_for_health` just gained. A child that dies mid-task now burns the full 120s and reports "did not finish within 120s; last poll error: …" instead of the exit status — before this PR it failed fast on the transport error. Net better under load, worse on a hard crash. | `&mut self` + the same `try_wait` probe at the top of the loop. |

### Merge

Merge as-is. M3 is the one I would fix in a follow-up commit before merge if it is cheap — it costs one doc-comment edit and it is the kind of false guarantee this gate exists to catch. M1 and M2 are test-harness hardening on a PR that is itself test hardening; they belong in the same file's next pass, not in a re-review round here.

bobmatnyc:
All three MEDIUMs and both LOWs addressed in 8ffa9273 and 11b14113. Raw gate output below.

## M1 — the drain re-armed the hang (correct, and worse than described)

Confirmed. `while let Ok(Some(line)) = lines.next_line().await` ends the loop on any `Err`, and `InvalidData` on one non-UTF-8 byte was enough to stop reading the pipe permanently.

Fixed at the source rather than handled: **the drain reads bytes now**, for which malformed input does not exist, so the only exits are real EOF and a real I/O error — and the error branch appends a visible `<<< drain stopped: {e}; capture truncated here >>>` marker before it stops. Two knock-on corrections fall out of moving to bytes: decoding happens lossily once at render time (a multi-byte character split across two reads is reassembled, not mangled), and stdout/stderr are captured into separate buffers (interleaving cannot misattribute a line).

`drain_survives_invalid_utf8_and_keeps_reading` pins it, asserting on the text *after* the bad byte — the half a line-based drain cannot satisfy. Verified in both directions:

```
--- reverted to the line-based drain ---
test support::api_server::drain_tests::drain_survives_invalid_utf8_and_keeps_reading ... FAILED
panicked at crates/trusty-agents/tests/support/api_server.rs:530:13:
drain stopped at the invalid byte instead of reading past it — nothing after it
was ever captured: --- child stdout ---
test result: FAILED. 1 passed; 1 failed; finished in 10.00s

--- byte drain ---
test support::api_server::drain_tests::drain_survives_invalid_utf8_and_keeps_reading ... ok
```

**Worth flagging against myself:** the first draft of that test had no deadline. Against the regression it did not fail — it spun for 60s+ and had to be killed. That is the same wait-forever defect this PR is about, reproduced in the test I wrote to prove I had removed it. It now polls the condition under a bound, which is why the run above fails cleanly in 10.00s instead of hanging.

## M2 — principle stated, applied to one loop only

Correct. `wait_for_task_with_timeout` used an unbounded `reqwest::Client`. It now carries a per-poll `TASK_POLL_TIMEOUT` (30s) — generous against a status read, tiny against the 120–240s ceilings it protects, and a failed poll is retried anyway.

## LOW — `&self` blocked the try_wait fast-fail

Taken. `wait_for_task_with_timeout` takes `&mut self` and reports a dead child at once instead of polling a socket nobody is listening on for the full ceiling. Three call sites in `api_e2e.rs` updated.

## M3 — the negative control guards something narrower than its doc claimed

Correct, and my doc was the false guarantee you describe. Both tests build their own client via `client_bounded_at` — that is what makes them run in milliseconds instead of minutes — so neither observes the production default at all.

The doc now states what it actually catches (widening the provisioning override from one route to the whole connector, which no other assertion in the pair would notice) and explicitly names what it does **not** catch, pointing at `config::tests::build_client_bounds_a_stalled_connection` as the real #2471 guard.

## LOW — `Transport(_)` cannot distinguish a timeout from a decode failure

Taken. `decode_err` yields the same variant, so the assertion now rules out the decode path by its literal prefix — reaching it would mean the stub answered within the bound, and the test would have proven nothing about timeouts while still passing.

Skipped `api_server.rs:82` (unbounded `output` growth) per your note.

## Gate — raw output, post-commit

```
cargo fmt --check EXIT=0

cargo clippy -p trusty-agents --all-targets -- -D warnings EXIT=0

cargo test -p trusty-mpm EXIT=0
44 binaries, all "test result: ok"
test result: ok. 4114 passed; 0 failed; 6 ignored; finished in 46.32s

cargo test -p trusty-agents EXIT=0
10 binaries, all "test result: ok"
test result: ok. 3148 passed; 0 failed; 11 ignored; finished in 27.76s
test result: ok. 6 passed; 0 failed; 3 ignored; finished in 3.82s   (api_e2e, was 4)

bash scripts/check_changelog_fragment.sh EXIT=0
```

Net test count: api_e2e went 4 -> 6 passing. Nothing disabled, nothing narrowed.

Issue #3551, opened 2026-07-21
trusty-installer: macOS supervisor bootstrap ignores $HOME — installer cannot be test-isolated and always mutates the live launchd supervisor


bobmatnyc:
Found during the 0.19.29 installer E2E (2026-07-21).

## Problem

`crates/trusty-installer/src/commands/plist_bootstrap.rs::install_mpm_supervisor_macos` unconditionally runs, once `trusty-mpm` installs on macOS:

```
launchctl bootout    gui/<real uid> <plist>
launchctl bootstrap  gui/<real uid> <plist>
```

The `gui/<uid>` domain is the **real user’s live launchd session**, keyed by `id -u` and the fixed label `com.trusty.mpm.supervisor`. It is **not namespaced by `$HOME`**. So even with `HOME` fully overridden to a throwaway dir, running `trusty-installer install` **boots out and replaces the live `com.trusty.mpm.supervisor`** on the host.

## Impact

1. **Not sandbox-testable.** The web-URL installer E2E cannot be run in isolation on any host that already runs the live stack — the supervisor bootstrap escapes the `$HOME` sandbox. The E2E had to PATH-shadow `launchctl` with a no-op stub to avoid clobbering the live daemon. That in turn makes the installer’s own `tctl install — verify` liveness sweep report `NOT VERIFIED` (no real daemon was loaded), so a genuine end-to-end `VERIFIED` result is not reachable in isolation.
2. **Surprise mutation.** Any invocation (test, CI, a second concurrent install, a curious user) restarts the real user’s mpm supervisor, regardless of the intended install prefix.

Related to the earlier installer-safety work (#3527 / #3533), which added a safety-gate + self-verify but did not gate this launchd bootstrap.

## Suggested direction

- Honor an env override (e.g. `TRUSTY_INSTALL_DIR` / a dedicated `TRUSTY_LAUNCHD_LABEL` or `--no-supervisor`/dry-run flag) so the supervisor bootstrap can target an isolated label/domain or be skipped for test installs.
- At minimum, namespace the launchd label or make the bootstrap opt-out, so `trusty-installer install` can run fully within a throwaway `$HOME` without touching `gui/<uid>/com.trusty.mpm.supervisor`.

## Evidence

Stub interception log from the E2E (these two calls would otherwise have hit the live supervisor):
```
[SANDBOX-STUB] launchctl bootout   gui/502 .../com.trusty.mpm.supervisor.plist
[...]

PR #4517, opened 2026-08-01
fix(trusty-installer): refuse a launchd bootstrap when a foreign process holds the daemon port (#4470)

### LINKED ISSUE #4470: trusty-installer: launchctl bootstrap does not detect/reject a foreign process already bound to the daemon port
**Defect:** a hand-run `launchctl bootstrap` succeeds even when a foreign (unsupervised, possibly older-binary) process is already bound to the daemon's port. [#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230)'s resolution item 2 asked for `bootstrap` itself to fail loudly in that case. [#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466) addressed this only partially: `docs/reference/release-workflow.md` now tells operators to pre-check, and `tm doctor`'s new `daemon_orphan` check detects the condition after the fact — but nothing intercepts the `bootstrap` itself.

**Why it matters:** the documented restart convention is `bootout → cargo install → bootstrap`. If an orphan holds the port, `bootstrap` reports success while the orphan keeps serving the old binary, so a fresh signed install plus behavioral verification can all pass while shipping nothing. Per [#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230) this nearly happened with the 1.2.3 deploy.

**Scope note:** the fix likely belongs in `trusty-installer`'s service-bootstrap path, which is why it was out of scope for [#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466) (a `trusty-mpm` PR).

**Resolution direction:** intercept `launchctl bootstrap` (or wrap it) to check for a foreign process already bound to the target port before proceeding, and fail loudly instead of reporting success. Not asserting a specific implementation as settled.

bobmatnyc:
## The mechanism

`launchctl bootstrap` exits 0 whether or not the job it loads can actually bind.
When an unsupervised orphan — typically an older binary — already holds the
daemon's port, the documented `bootout → install → bootstrap` restart convention
completes cleanly while the orphan keeps serving that port. A fresh signed
install plus behavioural verification can all pass while shipping nothing;
[#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230) records the
near-miss on the 1.2.3 deploy.

[#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466) addressed the
aftermath — `tm doctor`'s `daemon_orphan` check detects the state once you are
already in it — and documented an operator pre-check in
`docs/reference/release-workflow.md`. Nothing intercepted the `bootstrap`
itself. This is that interception, on the installer side where the bootstrap is
actually issued.

## Composition with #4466, not duplication

| | #4466 (trusty-mpm) | This PR (trusty-installer) |
|---|---|---|
| Question | should `tm start` **spawn** a daemon? | is the port this **bootstrap** targets already taken? |
| Keyed on | plist presence | who holds the port vs who launchd runs |
| Fires | before a spawn | before a `launchctl bootstrap` |

## Changes

New `commands::port_guard` refuses whenever the member's port is held by a PID
launchd does not report running for that label. The refusal names the offending
PID, the port, and the `lsof` / `kill -TERM` recipe to clear it.

**Fails closed.** An unreadable port probe, or a `launchctl` query that cannot
be answered, is a refusal — never a pass. `TCTL_ALLOW_FOREIGN_PORT=1` is the
single documented operator override, named in every refusal message so the way
through is discoverable at the moment of failure. `decide` is a pure function
over `(PortHolder, LaunchdOwner)`; the only cells that proceed are a free port
and the supervised daemon holding its own port.

**Every bootstrap-issuing path routes through the one shared check**, so no
branch can drift — the [#3841](https://github.com/bobmatnyc/trusty-tools/issues/3841)
lesson, where a fix applied to one branch left exactly the damaged machines on
the other:

- `service_bootstrap::bootstrap_one`, on BOTH the fresh-install branch and the
  #3841 already-present-plist repair branch, reported as the new
  `BootstrapAction::RefusedForeignPort`
- `verify_tail::attempt_verify_fallback`, the second-layer repair
- `lifecycle::launchd_control` for `tctl start` / `tctl restart`

For `restart` the gate runs **before** the `bootout`, so a refusal never stops a
running daemon and then declines to bring it back — the fail-open shape where
the failure branch has already advanced state. An already-loaded member is not
gated at all: it issues no bootstrap, so there is nothing to protect.

**No false accusations.** The port checked is the member's recorded `http_addr`
when it has one, falling back to the documented table, so a daemon that
legitimately port-walked (trusty-memory walks `7070..=7079`) is checked on the
port it actually bound.

## Common entry point

No new port-probing or PID-discovery machinery. The guard composes existing
shared helpers: `probe_http::fixed_port_for` (the documented port table),
`port::parse_port_from_addr` (the `host:port` splitter),
`plist_label::plist_label_for` (the label table), and
`verify_launchd_state`'s single `launchctl list` primitive.

That primitive was widened rather than copied: it now carries the running PID
and distinguishes "launchd says no such label" from "launchd could not be asked"
— the conflation #4466 identified as a HIGH on the trusty-mpm side, where it
made `tm doctor` prescribe a `kill` against a correctly supervised daemon. The
existing `Option`-shaped callers collapse the three states themselves, so
`classify_down_state` / `is_label_loaded` behaviour is unchanged.

`service_bootstrap.rs`'s inline tests moved to a sibling `_tests.rs` (the
`plist_bootstrap.rs` pattern) — it was at 448 of its 500 SLOC budget.

## How the tests prove the rejection path fires

Each refusal test was verified by deleting the guard and observing it fail:

```
bootstrap_one_refuses_when_foreign_process_holds_port                 → expected RefusedForeignPort, got Installed
bootstrap_one_refuses_existing_plist_when_foreign_process_holds_port  → expected RefusedForeignPort, got LoadedByFallback
attempt_verify_fallback_refuses_on_foreign_port                       → left: Some(true), right: None
```

Each also asserts NOTHING was installed or bootstrapped, so a guard that
refused *after* the side effect would still fail. `decide_never_proceeds_on_a_held_port`
enumerates every holder × owner cell and pins that `Proceed` occurs only on the
two safe ones, so a future branch leaking a pass fails even without a dedicated
test. `owner_from_list_unavailable_is_not_collapsed` fails the moment
"could not ask" is merged back into "runs nothing".
`every_launchd_member_has_a_guardable_port` closes the vacuous-port branch
structurally.

## Linux

`cargo check --target x86_64-unknown-linux-gnu` is not possible here (no
cross-linker), so the non-macOS `cfg` branches were compiled by forcing them
active. That caught a real Linux-only CI break — `variants Found and NotFound
are never constructed`, a hard error under `-D warnings`, since only the macOS
arm constructs them. Fixed with a narrow, documented
`cfg_attr(not(macos), allow(dead_code))`; re-simulated clean afterwards.

## Gates

```
cargo test -p trusty-installer                          → 530 passed; 0 failed; 4 ignored (lib)
                                                          + 3 + 3 passed (integration), exit 0
cargo clippy -p trusty-installer --all-targets -- -D warnings → exit 0
cargo fmt --all --check                                 → exit 0
bash scripts/check_line_cap.sh                          → 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh                → fragment present and valid
bash scripts/check_test_pointers.sh                     → 0 dangling pointers — OK
```

The 4 ignored tests are pre-existing (re-exec'd fd-isolation and live-network
tests). `#[ignore]` count in the crate is 15 on both `origin/main` and this
branch — no coverage was disabled to reach green.

Closes #4470


bobmatnyc:
## Code Critic — PR #4517

## Verdict: WARN

0 CRITICAL, 2 HIGH, 3 MEDIUM, 1 LOW. Reviewed on a detached worktree at `origin/fix/4470-bootstrap-port-guard` (448972ff) against `origin/main` (140621a5).

The `decide` truth table is sound and the fail-closed core holds — I mutation-tested it and could not get a `Proceed` out of any refusing cell. The two HIGHs are both at the **edges** of the guard: what happens to a refusal after it is returned, and one bootstrap site the enumeration missed.

## Findings

| HIGH | `crates/trusty-installer/src/commands/install.rs` | 531 | `RefusedForeignPort` is not counted as a failure, so a refusal reports `service_ok: true` / `all_ok: true` / exit 0 | Add the variant to the `is_failure` match |
| HIGH | `crates/trusty-installer/src/commands/install.rs` | 446 | A FIFTH ungated `launchctl bootstrap` on the same `tctl install` path — the trusty-mpm supervisor, which binds `127.0.0.1:7881` and fail-fast-exits | Gate `install_mpm_supervisor` on 7881; add a fail-loud enumeration test |
| MEDIUM | `crates/trusty-installer/src/commands/port_guard.rs` | 261 | Non-root `lsof` cannot see another user's listener; empty stdout is read as `Free` → `Proceed` | Cross-check with a bind attempt before concluding `Free` |
| MEDIUM | `crates/trusty-installer/src/commands/port_guard.rs` | 290 | `var_os(..).is_some()` — `TCTL_ALLOW_FOREIGN_PORT=0` enables the bypass, while every refusal says "Set …=1" | Parse the value, or restate the message as "set to any value" |
| MEDIUM | `crates/trusty-installer/src/commands/port_guard.rs` | 223 | A port-walking daemon (trusty-memory, 7070..=7079) tolerates a held port; the guard hard-refuses it | Skip the refusal, or walk the range, for members with a documented port range |
| LOW | `crates/trusty-installer/changelog.d/4470-bootstrap-port-guard.md` | 6 | Says "All three bootstrap-issuing paths"; the PR body says FOUR | Reconcile after HIGH-2 |

### HIGH-1 — a refusal is reported to automation as success

`install.rs:531`:

```rust
let is_failure = matches!(action, BootstrapAction::Failed(_));
```

`BootstrapAction::RefusedForeignPort` is not in that match, so a refusal takes the non-failure branch: it is printed with the `info:` prefix, and `service_ok` is set to `true`. `install_report.rs:142` computes `all_ok = members.all(|m| m.ok && m.service_ok && m.shadow_ok)`, which feeds `exit_code()`.

Net effect: an orphan holds the port, the guard correctly refuses to bootstrap, the daemon is never started — and `tctl install` exits 0 with `--json` reporting `all_ok: true, service_ok: true`.

That is the same failure class #4470 exists to close, one layer up. The comment six lines above the defect (`install.rs:519-527`) names it explicitly:

> the REPORT must not lie: a genuine bootstrap failure is routed through the error register (not the info one) and folds into `service_ok` / `all_ok` / the exit code (#2566 review — `--json` previously reported `all_ok: true` even when every daemon's service bootstrap had failed)

A refusal IS a bootstrap that did not happen. The mechanism is the repo's known shape: a `matches!` over a non-exhaustive variant list, where a new variant silently lands in the success bucket. `#4230` was a near-miss precisely because automation trusted an exit code.

Fix:

```rust
let is_failure = matches!(
    action,
    BootstrapAction::Failed(_) | BootstrapAction::RefusedForeignPort(_)
);
```

Please add a test at the `install_report` level asserting a refused member yields `all_ok == false` — none of the three deletion-proof tests reach past `bootstrap_one`'s return value, so nothing currently catches this.

### HIGH-2 — the enumeration is wrong, and nothing fails when it is

The PR body claims all FOUR bootstrap-issuing branches are gated (the fragment says three). There is a fifth, in the same crate, on the same `tctl install` run:

- `install.rs:446` → `plist_bootstrap::install_mpm_supervisor(force, &installed.path)`
- → `install_mpm_supervisor_for` → `target.launchctl.bootout(...)` then `target.launchctl.bootstrap(...)` (`plist_bootstrap.rs:106` is `launchctl bootstrap` against the real domain)

No `port_guard` call anywhere on that path. And it is not a harmless one:

- the plist (`plist_bootstrap.rs:249-250`) sets `TRUSTY_MPM_SUPERVISOR_ADDR=127.0.0.1:7881`
- `trusty-mpm/src/supervisor/http.rs:102` binds that address **before** the supervision loop, and its own doc says "the supervisor must fail *fast* if the metrics port is already in use"
- `KeepAlive` is `true`

So if anything holds 7881: `bootout` stops the live supervisor, `launchctl bootstrap` exits 0, the new supervisor dies on bind, launchd crash-loops it, and the installer prints "plist written … and bootstrapped". That is the #4230 mechanism verbatim, at the one site the guard does not cover — and the `bootout`-first ordering means this path also has the "stop it, then fail to bring it back" shape the PR carefully avoided in `lifecycle::Restart`.

Separately: there is no fail-loud mechanism. `decide_never_proceeds_on_a_held_port` pins the decision **table**; `every_launchd_member_has_a_guardable_port` pins the port **table**. Neither notices a new ungated bootstrap **call site** — which is exactly how this one was missed. The #4475 allowlist pattern applies here: assert the set of `launchctl bootstrap`-issuing sites in the crate against an explicit allowlist, so adding a sixth breaks the build.

### MEDIUM-1 — `lsof` under-reports, and the under-report reads as `Free`

`port_guard.rs:261`:

```rust
if stdout.trim().is_empty() {
    return PortHolder::Free;
}
```

Without root, `lsof` only reports file descriptors of processes the caller owns. Verified on this host as a non-root user:

```
$ lsof -nP -iTCP -sTCP:LISTEN | awk '{print $1,$2,$3}' | sort -u -k3
COMMAND PID USER
rapportd 495 masa
```

One row. Every root-owned and other-user listener on the machine is invisible, and the invisibility is indistinguishable from "no match" — same empty stdout, same exit 1. So a foreign holder running as another user is classified `Free` and the guard proceeds.

The module doc (line 66) says "`Free` means the probe positively found no listener". It does not; it means "no listener I can see". That is a fail-**open** branch inside the function whose stated invariant is fail-closed. Most #4230 orphans are same-user and will be caught, which is why this is MEDIUM rather than HIGH — but the guarantee is weaker than the docs claim.

Fix: when `lsof` yields no PID, attempt a `TcpListener::bind` on the port. `AddrInUse` with no `lsof` PID is `Unknown("port is in use but the holder is not visible to lsof")`, not `Free`.

### MEDIUM-2 — the override triggers on any value, including `0`

`port_guard.rs:290`:

```rust
let override_set = std::env::var_os(ALLOW_FOREIGN_PORT_ENV).is_some();
```

`TCTL_ALLOW_FOREIGN_PORT=0` and `TCTL_ALLOW_FOREIGN_PORT=` both **enable** the bypass. The doc comment says "when set to ANY value", so it is intentional and it matches `NO_SERVICE_ENV` at `install.rs:170` — but every refusal message this PR emits says:

> Set TCTL_ALLOW_FOREIGN_PORT=1 to bootstrap anyway

which teaches value semantics. An operator who sets `=0` to turn the bypass off turns it on. For a benign opt-out that is a footgun; for a safety bypass it is worth closing. Either parse the value (`=="1"`/`"true"`), or change the message to "set `TCTL_ALLOW_FOREIGN_PORT` (to any value)".

### MEDIUM-3 — a port-walking daemon is hard-refused on a condition it tolerates

`resolve_guard_port` (`port_guard.rs:223`) falls back to `probe_http::fixed_port_for` when the member has no recorded `http_addr` — the fresh-machine case. For trusty-memory that is 7070, but trusty-memory **walks** `7070..=7079` (the module doc says so at line 214). A held 7070 is a condition that daemon is designed to survive; the guard turns it into a refusal to install.

The PR frames the recorded-`http_addr` precedence as preventing false accusations, and it does prevent the *stale-port* one. It does not cover the fresh-install one: unrelated listener on 7070 + no recorded address → `Held(pid)` + `NotRunning` → `Reject`, on a machine where the install would have worked. A guard that blocks the normal path is the failure mode the brief warns about.

## What I verified and found clean

- **Fail-closed, every branch.** Mutation-tested: I removed both `bootstrap_one` gates and the `verify_tail` gate and ran the suite — exactly the three claimed tests failed (`bootstrap_one_refuses_when_foreign_process_holds_port`, `bootstrap_one_refuses_existing_plist_when_foreign_process_holds_port`, `attempt_verify_fallback_refuses_on_foreign_port`), 527 passed. They do assert the absence of side effects (`installed` and `fallback_calls` both empty), not just the return value.
- **Restart ordering.** `lifecycle.rs:371` — the guard precedes `cfg.bootout()` on the only `Restart` path (`apply_to_member` dispatches straight to `launchd_control`). No path leaves the daemon stopped after a refusal.
- **The widened shared primitive changes nothing for existing callers.** `main`'s `launchd_list_raw -> Option<String>` returned `None` on both spawn failure and non-zero exit. On the branch, `is_label_loaded` is `matches!(.., Found(_))` (identical to the old `.is_some()`), and `classify_down_state` explicitly collapses `NotFound | Unavailable => None` (identical). Both are the only callers besides the new one.
- **Test move lost nothing.** All 19 test fns from `service_bootstrap.rs`'s inline module are present in `service_bootstrap_tests.rs`, plus 3 new — name-set diff shows additions only.
- **`#[ignore]` unchanged.** 18 on `origin/main` and 18 on the branch across the crate. (The body says 15; different counting scope, but the substantive "no coverage disabled" claim holds.)
- **Local gates reproduce.** `cargo test -p trusty-installer --lib` → `530 passed; 0 failed; 4 ignored`. `cargo clippy -p trusty-installer --all-targets -- -D warnings` → clean. `cargo fmt --all --check` → clean.
- **The Linux `cfg_attr` is not masking anything.** On non-macOS only `launchd_list_raw`'s stub arm exists and returns `Unavailable`, so `Found`/`NotFound` are genuinely never constructed; `owner_from_list` still matches all three. The guard is unreachable in non-macOS production anyway (`bootstrap_member_service` returns `Skipped`, `launchd_control` bails, `attempt_verify_fallback` is cfg-gated) — no silent-pass hole.

## Gate status at review time

The body's gate block reports local runs. Live CI is **not green** — `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED`, and the `Test` check was still `IN_PROGRESS` (6 checks pending when I started). No failures so far. Not a finding against the code; flagging it because "all gates pass" reads as a merge-readiness claim it does not yet support.

## Required changes

1. Add `BootstrapAction::RefusedForeignPort(_)` to the `is_failure` match at `install.rs:531`, and add a report-level test asserting a refused member drives `all_ok == false`.
2. Gate `plist_bootstrap::install_mpm_supervisor` on port 7881, and add a fail-loud test over the set of `launchctl bootstrap`-issuing sites (the #4475 allowlist pattern) so a sixth site cannot be added silently.
3. Reconcile the "three" / "four" counts in the fragment and the body once (2) lands.

MEDIUM-1/2/3 are recommended, not blocking. MEDIUM-1 in particular is a cheap hardening (bind cross-check) of a guarantee the module doc currently overstates.

bobmatnyc:
## Review round 2 — all 6 findings addressed

Every fix is **mutation-verified**: I reverted it and confirmed the named test fails.

### HIGH-1 — refusal reported as success

`install.rs`'s inline `matches!(action, Failed(_))` never learned about `RefusedForeignPort`, so a refusal set `service_ok: true` and `tctl install` exited 0 with `all_ok: true`. You were right that this inverts the PR.

Classification moved onto `BootstrapAction::is_failure()` — one exhaustive `match` beside the definition, so a future variant is classified where it is defined rather than at a call site that can forget it. `install.rs:531` now calls it.

Two tests, at both layers:

```
MUTATION: RefusedForeignPort(_) => false
  install::tests::refused_foreign_port_drives_all_ok_false_and_a_nonzero_exit_code ... FAILED
  service_bootstrap::tests::bootstrap_action_failure_classification_is_exhaustive ... FAILED
```

The report-level one lives in `install_tests.rs` (where `install_report`'s private `build`/`exit_code` are in scope) and composes the real action→`service_ok` mapping with the real report builder: `all_ok == false`, `exit_code() != 0`.

### HIGH-2 — the missed fifth site, and nothing failing when the enumeration is wrong

`plist_bootstrap::install_mpm_supervisor` is now gated through the injected `LaunchctlPort` seam, so #3551's isolation guarantee holds (no test reaches real `lsof`/`launchctl`/a live port). The gate sits **before both the bootout and the plist write** — a refusal changes nothing on disk and never stops the supervisor without restarting it. `SUPERVISOR_METRICS_PORT = 7881` is pinned to `PLIST_TEMPLATE` by a test, so moving the port in the template without moving the constant fails.

I kept the guard's recording **out of** `StubLaunchctl::calls()` — two existing #3551 tests assert on that exact sequence, and folding a non-launchctl call in would have weakened assertions that exist to prove a test can never reach the live domain. It goes to a separate `port_guard_calls()`.

`bootstrap_sites_tests.rs` adopts the #4475 allowlist pattern. Verified by adding a synthetic sixth site:

```
MUTATION: added `cfg.bootstrap()` to status.rs
  bootstrap_sites_tests::every_launchctl_bootstrap_site_is_allowlisted_and_gated ... FAILED
    "commands/status.rs": 1,     <- flagged, with remediation instructions
```

My first allowlist guessed 6/2/2; the scan reported the real 2/1/2. The committed values are what is actually in the source.

### MEDIUM-1 — the fail-open branch (and a test that did not catch it)

Fixed with a bind cross-check: only a successful bind yields `Free`; `AddrInUse` with no `lsof` PID is `Unknown` naming the port and suggesting `sudo lsof`.

Worth flagging: **my first test for this was worthless and mutation testing caught it.** A fixture listener the test owns is one `lsof` CAN see, so the probe returned `Held` and never reached the cross-check — the test passed against the fail-open code:

```
MUTATION (attempt 1): port_is_bindable(port) -> PortHolder::Free
  test result: ok. 543 passed; 0 failed     <- did not catch it
```

So I split out `probe_port_holder_from(port, observation)` and inject an *empty* `lsof` result for a port the test has genuinely occupied — which is exactly the non-root blindness. Re-run:

```
MUTATION (attempt 2, same change):
  port_guard::tests::empty_lsof_output_on_an_occupied_port_is_unknown_not_free ... FAILED
  assertion `left != right` failed: an occupied port that `lsof` could not see
  must never classify as Free
```

`empty_lsof_output_on_a_free_port_is_free` holds the other half, so the fix cannot degenerate into "always Unknown".

### MEDIUM-2 — override value semantics

`override_from_env_value` now requires `1`/`true`/`yes` (case-insensitive, trimmed). `0`, empty, `false`, `no`, `off` all leave the guard armed. Taking the bypass logs a `WARNING:` line naming what it bypassed and that the daemon may not get its port.

### MEDIUM-3 — port-walking daemons

`resolve_guard_ports` returns the whole walk range for a fresh walker; `decide_over_range` proceeds at the first usable candidate and refuses only when all are unavailable, naming the primary port and the candidate count. Fail-closed is preserved — `Unknown` never counts as usable, pinned by `decide_over_range_refuses_when_every_candidate_is_unreadable`.

`walk_range_for` pins trusty-memory to `7070..=7079` (sourced from its own `doctor/checks.rs` and `daemon_guard.rs`) and asserts the other four members are *not* walkers, since wrongly listing one would let the guard proceed past a squatter.

One residual I did not paper over, recorded in `decide_over_range`'s doc: if an orphan of the *same* daemon holds the default port while a later port is free, the guard proceeds and the new daemon walks past it. A port probe cannot distinguish "my own orphan" from "an unrelated squatter". The non-walking members — where #4230 actually happened — have no range and are checked strictly.

### LOW

Fragment reconciled to four sites.

## Gates — live CI, not local

```
gh pr checks 4517 --watch --fail-fast   -> CHECKS_EXIT=0
statusCheckRollup: 24 SUCCESS / 25 total (1 skipping: "Notify on red main"), failing: []
  Test 11m49s pass · Clippy 4m36s pass · MSRV (1.91) 2m36s pass · Format check pass
  500-line file-size cap pass · Per-PR changelog fragment pass · doc-comment pointer lint pass
reviewDecision: REVIEW_REQUIRED   (this PR is NOT merge-ready; not merging)
```

Local, for the record:

```
cargo test -p trusty-installer  -> 546 passed; 0 failed; 4 ignored (lib) + 3 + 3 (integration), exit 0
cargo clippy -p trusty-installer --all-targets -- -D warnings -> exit 0
cargo fmt --all --check         -> exit 0
check_line_cap.sh               -> 7 allowlisted, 0 violations — OK
check_changelog_fragment.sh     -> fragment present and valid  (run AFTER committing)
check_test_pointers.sh          -> 0 dangling pointers — OK
```

The pointer lint caught two stale test names I left on `port_is_bindable` after renaming — fixed before pushing.

Linux `cfg` branches re-verified by forcing the non-macOS arms active and running clippy `-D warnings` clean, since this round touched cfg-adjacent code.

bobmatnyc:
## Code Critic — Round 2

## Verdict: APPROVE

All six round-1 findings verified fixed by mutation, not by reading. Zero CRITICAL, zero HIGH.

Reviewed at head `b8e444a7`, in an isolated worktree off `origin/main`. Baseline `cargo test -p trusty-installer --lib` → `546 passed; 0 failed; 4 ignored`. Clippy `-D warnings` exit 0, `cargo fmt --check` exit 0. Live rollup at `b8e444a7`: all checks pass (`Notify on red main` SKIPPED, expected).

### Round-1 items — mutation evidence

| # | Claim | Mutation applied | Result |
|---|---|---|---|
| HIGH-1 | classification moved onto `is_failure()` | moved `RefusedForeignPort` to the `false` arm | 2 tests FAIL |
| HIGH-1 | test reaches `all_ok` / `exit_code`, not just the enum | dropped `service_ok` from `InstallReport::build`'s fold | `refused_foreign_port_drives_all_ok_false_and_a_nonzero_exit_code` FAILS |
| HIGH-2 | supervisor gate precedes bootout AND plist write | (a) deleted the gate (b) moved it after both | FAILS in both |
| HIGH-2 | a new bootstrap site fails the scan | added an ungated site in `commands/ensure/` (new file, subdirectory, argv literal) | scan FAILS, names the file |
| MEDIUM-1 | bind cross-check has teeth | (a) `port_is_bindable` → always `Free` (b) deleted the cross-check outright | `empty_lsof_output_on_an_occupied_port_is_unknown_not_free` FAILS in both |

`is_failure()` is genuinely exhaustive — six explicit arms, no `_ =>`. A seventh variant is a compile error, not a silent pass.

**Regression sweep clean.** `is_failure()` has exactly one production call site (`install.rs:534`). It is a strict superset of the old inline `matches!(action, Failed(_))`, adding only the variant this PR introduced. No pre-existing caller's behaviour changed.

**MEDIUM-1's self-catch is real and the second test is genuinely better.** Deleting the cross-check leaves the suite green under the first (fixture-listener) design and red under the injected-observation design. That is the distinction the author claimed, and it holds.

**MEDIUM-3's residual is acceptable.** Verified rather than taken on faith:
- `walk_range_for` returns `Some` only for `trusty-memory`. The five launchd-managed members are search / memory / analyze / review / console — `walk_range_matches_the_daemons_documented_behaviour` covers all of them, so the coverage is complete, not a hand-picked four.
- #4230 was port **7880** — `trusty-mpm`, `ManageStrategy::OwnVerb`, `walk_range_for` → `None`. The claim that the incident case is checked strictly is accurate.
- The residual is narrower than stated: whenever `trusty-memory` has a recorded `http_addr`, `resolve_guard_ports` returns one port and the range relaxation never applies. Reaching the hole needs an orphan holding 7070 *and* no readable addr file — a narrow window that does not re-open #4470's original failure.

## Findings

| MEDIUM | `crates/trusty-installer/src/commands/port_guard_tests.rs` | 429–439 | `empty_lsof_output_on_a_free_port_is_free` binds `127.0.0.1:0`, releases the port, then re-binds it. Another test in the same binary — including `empty_lsof_output_on_an_occupied_port_is_unknown_not_free` ten lines above, which also binds `:0` — can be handed that exact ephemeral port in the window. Reproduced twice in 33 clean-head suite runs. | Retry the acquire-and-probe sequence over a few fresh ephemeral ports and fail only if none ever classifies `Free`, rather than asserting on a single released port. |
| MEDIUM | `crates/trusty-installer/src/commands/port_guard_tests.rs` | 543–558 | `resolve_guard_ports_returns_the_whole_walk_range_for_a_fresh_walker` reads `read_daemon_addr` for its skip guard, then `resolve_guard_ports` reads it again. Other tests in this binary set and clear the process-global `TRUSTY_DATA_DIR_OVERRIDE` (`test_support.rs:306,338`) under `ENV_TEST_LOCK`, which this test does not take — so the two reads can disagree. On a host with a real `trusty-memory` `http_addr` (mine holds `127.0.0.1:7070`) it fails with `ports: [7070]`, len 1 ≠ 10. Reproduced once in 33 runs. CI is green only because ubuntu-latest has no such file. | Take `ENV_TEST_LOCK` and use `test_support::stub_data_dir` so the test controls the answer deterministically, instead of skipping on a value it re-reads unsynchronised. |
| LOW | `crates/trusty-installer/src/commands/bootstrap_sites_tests.rs` | 94 | The scan matches `.bootstrap(` and the literal `"bootstrap"`. I added two ungated `launchctl bootstrap` sites — a `sh -c "launchctl bootstrap …"` string and a runtime-assembled verb — in a new module, and the test stayed green. The module documents this reach limit explicitly, and the realistic direct-call shape **is** caught (verified). Recorded so the limit is on the record, not because it should block. | Optional: also flag non-comment lines containing `bootstrap` in any file that mentions `launchctl`. Weigh against false-positive noise; the documented status quo is defensible. |
| LOW | `crates/trusty-installer/src/commands/plist_bootstrap.rs` | 548, 553 | The log dir and `Library/LaunchAgents` are created before the port guard at 593, so a refusal leaves two empty directories. The plist itself is correctly untouched and the test pins that. Idempotent and harmless. | Move both `create_dir_all` calls below the guard, or leave as-is. |

## Notes

**Merge as-is.** The shipped guard is correct — six independent mutations all produce red, and the refusal now propagates to `all_ok` and the exit code, which was the actual round-1 defect. Nothing here is a correctness problem in the code under review.

The two MEDIUM findings are test-quality only: two new tests introduce a ~9% chance of a red full-suite run on a developer machine (3 reds in 33 clean-head runs). Given this repo's standing position that load-induced red degrades the gate itself, and #4516 having just landed for exactly this class of problem, both are worth fixing on the way in — but they are small, mechanical, and do not warrant a round 3. Fix them in this PR or as an immediate follow-up; either way, no re-review needed.

Round 2 with no HIGH remaining and only test-hygiene left is a finished branch. Ship it.

bobmatnyc:
## Round-2 follow-ups: both flakes fixed, LOW-2 taken

### Flake 1 — `empty_lsof_output_on_a_free_port_is_free`

Each attempt now uses an **independent** OS-assigned port, retrying up to 25 times. A real regression (`Free` never returned) still fails — it fails on every attempt; only the race is absorbed, since 25 independent fresh ports all being instantly re-taken does not happen.

I rejected the fixed-port option: it trades the race for a false red on any developer machine already running something there — the same "red on a dev box, green in CI" shape as flake 2.

### Flake 2 — `resolve_guard_ports_returns_the_whole_walk_range_for_a_fresh_walker`

Takes `ENV_TEST_LOCK` for its whole body, and rather than just locking, it now **stages** the fresh-machine state with `stub_empty_data_dir`. Your diagnosis exposed a second problem in the same test: it keyed on ambient host state and returned early when a real addr file existed, so it self-skipped on a dev machine and only ever ran on CI — green where it could not fail, absent where it could. Staging makes it assert the same thing everywhere.

Added `resolve_guard_ports_prefers_a_recorded_address_over_the_walk_range` to pin the other half of the rule, which is the invariant behind your MEDIUM-3 note: a recorded `http_addr` yields a single-port check and no range relaxation applies.

### Proof — repeat runs, with a control

A single green run is what let flake 1 through, so a single green run is not the evidence. I ran the whole lib binary 40× per configuration, and first verified the harness can **detect** the flake it claims to fix:

```
racy version restored, 8 threads : 38 green / 2 red out of 40   <- control: reproduces your 2/33
fixed, 8 threads                 : 40 green / 0 red out of 40
fixed, 16 threads                : 40 green / 0 red out of 40
```

The control is the load-bearing line. Without it, 40 green proves only that I ran a suite 40 times.

### LOW-2 — taken (one-line change, as you guessed)

The two `create_dir_all` calls moved after the port gate. The existing refusal test now also asserts neither the `LaunchAgents` dir nor the log dir is created, so "a refusal changes nothing on disk" is now literally true rather than nearly true.

### LOW-1 — left as-is, per your call

The scan stays literal. `sh -c "launchctl bootstrap …"` and a runtime-assembled verb do slip past; the module doc states that limit explicitly rather than implying full coverage.

## Gates — live rollup on `d9ead288`

```
gh pr checks 4517 --watch --fail-fast   -> CHECKS_EXIT=0
rollup: 24 SUCCESS / 25 total (1 "skipping": Notify on red main), notPassing: []
  Test 12m40s pass · Clippy 5m59s pass · MSRV 1.91 3m48s pass · Format 16s pass
  500-line file-size cap pass · Per-PR changelog fragment pass · doc-comment pointer lint pass
```

Local, post-commit:

```
cargo test -p trusty-installer  -> 547 passed; 0 failed; 4 ignored (lib) + 3 + 3 (integration), exit 0
cargo clippy --all-targets -D warnings -> exit 0
cargo fmt --all --check                -> exit 0
check_line_cap.sh           -> 7 allowlisted, 0 violations — OK
check_changelog_fragment.sh -> fragment present and valid   (run after committing)
check_test_pointers.sh      -> 0 dangling pointers — OK
```

## One thing before you merge

`mergeStateStatus: BEHIND` — main has moved ahead since this branch was cut. CI above is green on the branch as it stands, not on the merge result. Your call whether to update the branch first (re-triggers the ~13min Test job) or merge as-is; I have not touched it either way.

`reviewDecision: REVIEW_REQUIRED`, state OPEN. Not merging — over to you.

Issue #4406, opened 2026-07-30
[trusty-agents] Two unconnected systems are both called "OKG" — trusty-kb entity tree vs trusty-memory KG


bobmatnyc:
## Finding

There are two distinct, unbridged stores both referred to as "OKG":

1. **trusty-kb markdown-file entity tree** — written by `KbStore::put_entity` (crates/trusty-kb/src/okg/mod.rs). This is what the `okg_ingest_docstore` / `okg_ingest_gmail` / `okg_ingest_drive` tools populate.

2. **trusty-memory knowledge graph** — written/read by `kg_assert` / `kg_query` (crates/trusty-memory/src/service/core_kg.rs), which back the MCP tools of the same names.

Neither depends on nor calls the other. Nothing bridges them.

## Why It Matters

Milestone #22 (M2 — trusty-agents internal driver beta) now carries several OKG issues (#3904, #4283, #4289, #4007, #4325, #4363). If those are implemented without first stating WHICH store is the target, they will be built against different targets and will not compose. This needs resolving before the M2 OKG work starts, not during it.

## Supporting Evidence

From verification against origin/main on 2026-07-30:

- `kg_query` returns EMPTY for both real populated palaces — `cto` (63 drawers) and `trusty-tools` (814 drawers) — each with the literal hint "Knowledge graph is empty. Run kg_bootstrap...".
- No palace named `cto-assistant` exists in trusty-memory.
- The CTO assistant's apparent knowledge base lives in a filesystem `trusty-kb` tree and was populated by an LLM hand-transcribing content into markdown — which crates/trusty-agents/src/tools/okg/mod.rs documents as the status quo the ingest tools were built to replace.
- No LLM call exists anywhere in the OKG ingest path (crates/trusty-agents/src/tools/okg/mod.rs:1-10 states entity distillation is left to the assistant, not code).
- Nothing reads FROM a trusty-search index INTO entity extraction; `feed_bound_index` (crates/trusty-agents/src/tools/okg/index_feed.rs:33-78) runs the opposite direction, pushing a KB tree out to an index.

## Open Question for Owner

Which store is canonical for an assistant instance's knowledge — the trusty-kb entity tree, the trusty-memory KG, or both with a defined sync direction? Note this likely warrants an ADR rather than an issue-level decision.

---

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

---

## Related Issues
- #3904
- #4283
- #4289
[...]

Issue #4320, opened 2026-07-29
Task errors are rendered verbatim as assistant narrative with no error framing


bobmatnyc:
**Motivating incident**: In the live trusty-agents app at 13:38 local on 2026-07-29, a task error was presented to the user as the string `subprocess exited with status Some(1)` — rendered as if the assistant had said it.

## Problem Chain (verified end-to-end)

1. **Error formatting** (`crates/trusty-agents/src/api/server/task_runner.rs:166-171`): formats `format!("subprocess exited with status {:?}", status.code())` — a Debug-formatted `Option<i32>`, hence the user-visible `Some(1)`

2. **Narrative injection** (`crates/trusty-agents/src/api/types.rs:187-203`): `PmResponse::error` puts that raw string into `narrative` VERBATIM

3. **Handler passthrough** (`handlers.rs:481-486`): finalizes it with no error-specific transformation

4. **Event emission** (`ui/src-tauri/src/task_commands.ts:265-279`): emits it on the `task-complete` event — NOT `task-error`. A code comment at `:265-267` states this is deliberate so ChatView can display the failure narrative. Note: `task-error` (used for transport failures, e.g. the 10-minute timeout at `:222-230`) DOES get an `"Error: "` prefix at `ChatView.svelte:114`

5. **Chat rendering** (`ChatView.svelte:98-103`): renders `p.narrative` unmodified as the chat bubble

## Net Effect

A low-level Rust Debug string is presented to the user as if the assistant said it. There is no re-narration step on this path — it is a fire-and-parse pipe with no model in the loop to explain the failure.

## Suggestions (for maintainer discussion)

- Route task failures through the existing `task-error` framing, and/or
- Give the model a chance to narrate a tool/task failure instead of passing raw error text to the UI

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4283, opened 2026-07-29
OKG entity extraction: pull entities from attached search indexes into the canonical per-assistant OKG store


bobmatnyc:
Current owner scope (2026-09-09/10): attached projects automatically feed business-entity extraction into the protected OKG of each directly user-facing assistant. This supersedes the original manual-trigger/open-question wording. Existing parent #4007 is preserved; #4531 owns source orchestration. Earlier diagnosis and handoff history remain in comments.

- [ ] Automatically discover business entities and relations from attached project and authorized source content using model-assisted extraction and bounded inexpensive cleanup; retain raw project indexing separately.
- [ ] Persist canonical entities with source evidence, provenance, extractor/model identity, and untrusted-content treatment; retries are idempotent and failed batches never advance completed coverage.
- [ ] Source edits/deletions retract superseded active claims from search without watcher reinsertion; preserve historical evidence safely. #7379 remains a native blocker until the actual corpus/query acceptance passes.
- [ ] Make configuration, execution state, results, and errors directly API-testable; verify real-provider extraction, publication/search, updates, and restart on the installed artifact. Initial history and requested extensions use one-month batches.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Resumed 2026-09-06 after staleness review (session trusty-tools-38): blocker cleared or never live; gap re-verified against origin/main. First task: first task per the issue body.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Accepted design update under #4531: extraction now runs automatically for attached projects, into one protected OKG per directly user-facing assistant; delegated/specialist agents do not receive stores. Raw project indexing remains separate from extracted entities. This supersedes the older explicit-only trigger discussion.

Code evidence: `crates/trusty-kb/src/okg/ingest.rs::KbStore::ingest_items` writes source items; format extraction alone is not business-entity discovery or cleanup. Claude Code owns the upstream implementation handoff; Codex is implementing the assistant API/spec control plane.

[...]

Issue #4282, opened 2026-07-29
Cap attached search indexes at 10 per assistant, and ship the selection GUI


bobmatnyc:
Parent: #4007

### (i) Cap attached search indexes at 10 per assistant

DOC-58 §2 specifies K-d cardinality as "N per agent" with NO numeric ceiling. The merged backend (PR #4019, `crates/trusty-agents/src/agents/config.rs:345`, `resolved_search_indexes()` at `:380-398`) only trims and dedupes — no length check exists today.

The owner's "up to 10" is a genuinely NEW requirement, not something already specified.

**Open question for the owner:** is 10 a hard block or a warn? K-a's precedent at `crates/trusty-agents/src/stores/config.rs` warns rather than fails above one OKG store — this is undecided and needs the owner's call before implementation.

### (ii) Ship the selection GUI

#4010 was titled "Knowledge pane GUI" but was CLOSED via PR #4019, which touched ZERO files under `ui/`. There is no surface today for a user to select indexes.

K-d ships read-only per DOC-58 §5.2 KD-9: "K-d ships read-only in M1... there is no PATCH in this document" — so a selection GUI also needs a write path (a new PATCH/attach-detach endpoint) that does not exist yet.

## Acceptance Criteria

- Owner decision recorded: hard cap vs. warn above 10.
- Backend enforces the cap (or warn) in `resolved_search_indexes()` / `agents/config.rs`.
- A new write path exists for attach/detach of search indexes (beyond today's read-only K-d).
- A GUI surface exists for a user to view and select/attach/detach indexes, up to the cap.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Owner confirmation (2026-07-28/29), independently from the product side: looking at the trusty-search MCP tool card in the agent configuration UI — currently rendering as "trusty-search / ENABLED / Trusty search service — semantic + keyword code/knowledge search over a native MCP-stdio server" — the owner said a user must be able to **connect to specific indexes** from that card.

This confirms part (ii) of this issue (the selection GUI) from the product side, and pins the concrete surface: the selection affordance should be reachable **from the trusty-search MCP tool card itself** in the agent config UI, not only from a separate knowledge pane. Restating the two blockers already recorded above so they aren't rediscovered:

- #4010 ("Knowledge pane GUI") was closed via PR #4019, which touched ZERO files under `ui/` — no selection surface exists today.
[...]

PR #4523, opened 2026-08-01
feat(trusty-agents): per-assistant home dir + OKG store model (#4325)

### LINKED ISSUE #4325: Assistant store model: app-generated per-assistant home directory with explicit OKG entity extraction
## Product Decision

Each assistant gets an AUTO-CREATED, APP-GENERATED home directory containing:
- `instructions.md` — assistant instructions
- `config.toml` — assistant configuration
- `/agents` — custom agents scoped to that assistant
- `/okg` — its knowledge graph
- `/attachments` — assistant attachments and artifacts

Stores may point at ANY search store/index. Entities enter the OKG ONLY via an EXPLICIT ENTITY EXTRACTION COMMAND — never implicitly, never as a side effect of ingestion.

## Design Rationale: Visibility and Resilience

The assistant home directory is intentionally HUMAN-READABLE and USER-FACING, designed to be browsable like trusty-mpm-projects directories. This means:

**External modification is EXPECTED, not an error condition.** Users will edit, move, rename, and delete things in the home directory outside the system. The system must be designed to tolerate this, not defend against it.

Specifically:
- The system is NOT required to auto-migrate or auto-relocate config when a user moves things (explicitly out of scope)
- The system IS required to DETECT missing or malformed files
- On detection, the CONCIERGE surfaces the issue CONVERSATIONALLY and HELPS THE USER FIX IT (guided remediation, not raw error dumps)
- This is a product-level requirement that drives mechanical changes in error handling, subprocess visibility, and health-probe infrastructure

## What This Resolves

**(a) Issue #4283 — OKG extraction trigger:** The extraction TRIGGER is now explicit and documented. Users trigger entity extraction consciously, not as a side effect of store binding.

**(b) Epic #4007 — Clean separation of tiers:** Arbitrary indexes are READ-attachable via store binding. The OKG is WRITTEN only deliberately via explicit extraction commands. This cleanly separates the two tiers.

**(c) Real hazard eliminated:** Today's store binding is bidirectional in effect. `feed_bound_index` (`crates/trusty-agents/src/stores/index_feed.rs`) pushes OKG ingest output INTO the bound index at the end of every ingest run. Binding a populated source corpus (e.g., the 201k-chunk `cto-duetto` index over the CTO source repo) would mix generated entities into that source corpus. Explicit-extraction-only removes this hazard entirely.

## What This Requires (none exists today)

1. **Per-assistant root field on AgentStoreBinding** — see #3890 for detailed scope note about the data model gap
2. **Writable store-config path** — #3890 tracks the missing PATCH contract
3. **Index→OKG extraction command** — #4283 tracks this explicitly
4. **Resilience to external changes** — this ticket establishes the product requirement; mechanical implementation spans:
   - #4315 (per-connection health/probe API) — detection half: a probe that reports resolved identity and meaningful failure reason is what lets the concierge say something useful
   - #4320 (task errors rendered as raw assistant narrative) — the counter-example behavior to avoid; this ticket is where the narration seam belongs
   - #4321 (subprocess stderr unrecoverable) — you cannot help a user fix what the system never recorded; this ticket mandates subprocess-output visibility as a prerequisite to guided remediation

## Open Questions Resolved

**DOTLESS PATHS ARE CONFIRMED AND INTENTIONAL.** The home directory is a visible, user-facing, human-browsable directory by design — just like trusty-mpm-projects. This is a migration affecting existing trees (today they're dotted under `.trusty-agents`), but it is NOT an open question anymore.

## Relates

Epic #4007, #4282, #4283, #3890, #4290, #4315, #4320, #4321

bobmatnyc:
## What changed

Adds `crate::assistants` — the per-instance container that models **instance isolation for the Assistant TYPE**, plus the `[[stores]] root` field #4325 lists as requirement 1.

| File | What |
|---|---|
| `crates/trusty-agents/src/assistants/instance.rs` | `ASSISTANT_ROLE` / `is_assistant_role` (the TYPE discriminator) and `AssistantInstanceId` (a validated INSTANCE name) |
| `crates/trusty-agents/src/assistants/home.rs` | `AssistantHome` — layout, resolution, app-generated `ensure()`, and `store_root()` |
| `crates/trusty-agents/src/assistants/health.rs` | `inspect()` / `HomeHealth` / `HomeIssue` — the detection half of the resilience requirement |
| `crates/trusty-agents/src/assistants/error.rs` | `AssistantError` (`thiserror`) |
| `crates/trusty-agents/src/stores/config.rs` | new `AgentStoreBinding.root` + its `validate()` rule |

Added: 1584 / Removed: 3 / Net: +1581 (of which 546 lines are the four test modules).

## The instance-isolation model

Per the owner's 2026-07-30 wording: **"Assistant" is a TYPE. `izzie` and `cto-assistant` are INSTANCES of that type — they are NOT distinct agent types.** The repo already encoded the type half (`assistant/agent.toml` declares `role = "assistant"`; both named personas carry `extends = "assistant"`). Nothing encoded the instance half — an instance's three belongings were spread across shared pools with no per-instance container:

- persona → the shared agents directory
- OKG tree → the flat `<knowledge_dir>/<slug(agent)>` pool, addressed by naming convention
- store root → did not exist as a field at all (#3890's data-model gap)

That is precisely why adding an instance reads as adding a *type*. This PR adds the container, and nothing else:

```
~/trusty-agents/<instance>/                # DOTLESS, visible, browsable
├── instructions.md
├── config.toml                            # TOML — confirmed, not assumed
├── agents/
├── okg/                                   # indexed by trusty-search
├── attachments/
└── stores/                                # parent only; <store-identifier>/ is a later spec
```

`TAGENT_ASSISTANTS_DIR` overrides the root. Two instances of the same type get two homes, two `okg/` trees, two `agents/` dirs — verified by `two_instances_get_separate_homes`.

**Resilience.** #4325 makes external modification EXPECTED. So:
- `ensure()` is additive and idempotent — it creates only what is missing and never rewrites a file the user edited (`ensure_never_overwrites_user_edits`, `ensure_restores_only_what_is_missing`). Restoring its own seed over a user's work would be exactly the intolerance the ticket rules out.
- `inspect()` never fails and never repairs. It returns structured findings — kind, path, detail, and a **remedy** — and `HomeHealth::narration()` is the seam #4320's concierge renders from. A missing home is one finding, not five.
- No auto-migration and no auto-relocation, per the ticket.

**`[[stores]] root`.** Optional, relative, traversal-free; resolved by `AssistantHome::store_root()` and defaulting to the home's `okg/`. Absolute or `..`-bearing values are refused at resolution AND reported by `AgentStoreBinding::validate()` (fail-soft, matching the module's existing posture — a bad store degrades, it never blocks boot). Bindings that omit `root` parse and validate byte-identically to before.

## Startup wiring (owner answer, 2026-08-01)

`ensure()` now runs on the boot path. **This is user-visible:** the first launch after this lands creates `~/trusty-agents/<instance>/` trees that did not exist before, logged at `info` (`created assistant home directory`) — not silent. A home that cannot be created logs at `warn` with its reason and remedy.

**Call site: `runtime::startup::run_startup_init`**, immediately after `ensure_bundled_agents_deployed()`. Two reasons it is that line and not another:

- It is the single per-launch choke point, and it runs **before** the `--api`/`--serve` early dispatch further down. One call therefore covers the API server, the Tauri GUI, the `start` daemon, the REPL/TUI and every one-shot subcommand.
- The GUI needs no second hook: `ui/src-tauri` spawns `tagent --api` as a sidecar (`sidecar.rs::ensure_api_server`) rather than provisioning anything itself; its `setup()` only builds the tray. Same for the daemon — `service::start_service` is a detached `tagent --serve`, so it re-enters this path.

**Not covered, deliberately:** `config` and `mcp-serve` return before `run_startup_init` runs. Both are intentionally daemon-less and stdout-protocol-sensitive, and neither touches an assistant home. That is the honest full answer to "does more than one entry point need this" — one site covers every launch mode, and the two it misses should miss it.

**It cannot fail startup.** Nothing in the new `provision` module returns `Result`. A read-only `$HOME`, a denied permission, a full disk, or a file left where a directory belongs becomes a `HomeIssueKind::NotCreatable` finding carrying the reason and a remedy, surfaced through the `health::narration()` seam built earlier in this PR. No panic, no `unwrap()`, no propagated error. It also cannot repeat the sidecar's historical `cwd=/` EROFS crash: the root resolves from `$HOME`/`$USERPROFILE`, never the cwd.

**Creation only.** No migration, no relocation, no cleanup; the dotted `~/.trusty-agents` tree is untouched. Relocating a home needs a migration process that stays future work.

### A judgment call worth your eye: which agents get a home

`role = "assistant"` alone is the **wrong** rule. `ctrl` declares it too — and the GUI's null `activeAgentId` *means* ctrl (Concierge), so provisioning it a selectable instance home would model the one agent that is not an instance as if it were. Selecting on role alone would have silently created it.

`roster::discover_instances()` uses role **plus lineage**: be the base `assistant`, or declare `extends = "assistant"`. Mechanical, no name special-casing. On the shipped roster that yields `assistant`, `izzie`, `cto-assistant` and excludes `ctrl`.

**It also excludes `personal-assistant.toml`**, which declares `role = "assistant"` but extends nothing. I judged that correct — it is a standalone persona, not an instance of the base — but it is a genuine call, so flag it if you disagree; it is a one-line rule change.

Note this introduces the **first boot-time agent enumeration** in the codebase (`agent_roster()` is per-request today). It is a partial, forgiving TOML read of `[agent]` only — the same shape `stores::binding::load_stores` already uses — so a malformed file is skipped, never fatal.

### Still no persisted selection

Confirmed by inspection: `activeAgentId` is in-memory only (`ui/src/stores/app.ts:145`), `user.toml` carries no agent field, and the backend documents the null default as deliberate. So "provision the selected assistant" has nothing to select yet — every discovered instance is provisioned, which is what makes #4281's selection land on homes that already exist.

## Owner decisions since the first push

**The store path is the owner's, verbatim (2026-08-01): `trusty-agents/<agent>/okg/`.** My original spelling carried an extra `assistants/` segment — flagged at the time as my own choice, not the ticket's — and has been dropped. Home is now `~/trusty-agents/<instance>/`, `okg/` sits directly under it, and `okg_store_path_matches_the_owners_spelling` pins the full absolute path rather than only the accessors (an accessor-only test would not have caught the extra segment, and the OKG Sources spec is being written against this path).

**`stores/` joins the layout (2026-08-01).** `<trusty-agents home>/<agent>/stores/<store-identifier>/` — one subdirectory per remote store, holding that store's extraction target and extraction state. **This PR creates the parent directory and nothing else.** It gets the same additive/idempotent/never-overwrite `ensure()` semantics and the same health findings as every other layout directory; store-identifier derivation, the extraction-state record, per-store subdirectories and extraction logic are all absent by design, because a separate spec defines them. `ignores_whatever_lives_under_stores` pins that boundary — inspection looks at the directory, never at its contents, so whatever the spec later puts inside cannot make a home report unhealthy. Creating the parent now means that slice needs no PR just to add a `mkdir`.

**TOML is confirmed, not assumed.** The `config.yaml` vs `config.toml` question is closed per the owner's 2026-07-29 correction: TOML stays, no format migration. Same correction confirms the dotless spelling, and narrows "protected"/"app-generated" to **ownership of origin with access control explicitly out of scope** — no write-enforcement, no floor, no narrow-only semantics. The doc comments have been softened accordingly (`assistants/mod.rs`, `AssistantHome::ensure`); nothing in the code ever enforced anything, but two comments read as if it might.

**#4406 is settled for this layout, by the owner rather than by me.** The owner specified `okg/` as a filesystem store indexed by trusty-search — the trusty-kb tree shape — which confirms the minimal read I flagged in the first push. The binding's separate `palace` field correctly stays untouched, so the trusty-memory leg remains addressable without changing this layout. I did not decide this; it was decided for me and matches.

## Remaining assumptions

1. **`attachments/` is created but its lifecycle is undefined.** The owner's comment scopes it to binary chat attachments and records the blocker: no binary→text extraction exists anywhere (`read_text` rejects NUL-bearing files; zero hits for `docx|pptx|xlsx|pdf_extract|lopdf|calamine|zip::`). New capability, not this ticket.
2. **`config.toml` is deliberately minimal** (`id`, optional `display_name`, unknown keys tolerated). Agent behaviour still comes from `agent.toml`; this file exists so instance-scoped settings have a landing place without a second format decision.

## Notes for the OKG Sources spec

Nothing here blocks a pluggable extractor writing into `okg/` on a schedule, but four things in the current shape are worth the spec accounting for:

1. **`StoresConfig::validate()` warns above ONE binding per agent** (`crates/trusty-agents/src/stores/config.rs`, pre-existing since #3816). If each OKG Source became its own `[[stores]]` binding, every multi-source assistant would emit a config warning. Sources should sit *under* the single store — e.g. `[[sources]]` in the home's `config.toml`, which already tolerates unknown keys — not as sibling bindings.
2. **`store_root()` confines the DESTINATION, not the source.** An extractor cannot repoint `okg/` outside the home (deliberate — it is what stops one instance writing into another's store). Reading from an arbitrary directory, Gmail, Drive, Slack, Notion or Granola is entirely unaffected. If a source type ever needs to write outside the home, that is a new decision, not a tweak.
3. **There is no place for per-source refresh state yet.** `health::inspect()` checks that `okg/` exists and is a directory; it knows nothing about last-run time, cursors, or a failing credential. A scheduled refresh needs that state, and `HomeIssue`/`HomeIssueKind` is the natural place to surface a stale or broken source through the same concierge narration seam — but the kinds are filesystem-shaped today and would need extending.
4. **Secrets must not land in the home.** The home is deliberately human-browsable and user-editable; a Granola API key or Slack token in `config.toml` would be sitting in a directory the product invites users to open. The spec should name a separate credential store.

## Out of scope## Out of scope (interfaces left open for them)

#3890 (writable store-config PATCH), #4283 (index→OKG extraction), #4282 (attached-index cap + GUI), #4281 (persisted selection). None are implemented here. `ASSISTANT_REACHABLE_SUBAGENTS`, `[subagents].delegate_allowed` and the delegate allow-set are untouched — this is instance isolation, not capability grants. No existing test was weakened, ignored or deleted.

## Test evidence

47 new tests. `cargo test -p trusty-agents --lib assistants::`:

```
test result: ok. 47 passed; 0 failed; 0 ignored; 0 measured; 3167 filtered out; finished in 0.05s
```

The five startup cases you asked for, plus the ones that turned out to matter:

| Test | Covers |
|---|---|
| `startup_creates_a_missing_home` | no home at all → creates it |
| `second_startup_creates_nothing` | complete home → no-op |
| `startup_fills_only_the_gaps` | partial home → fills only gaps |
| `repeated_startup_never_disturbs_user_edits` | 3 boots over edited files → untouched |
| `startup_survives_a_home_it_cannot_create` | **creation fails → app still starts, reason + remedy reach narration** |
| `provisions_every_instance_independently` | one unwritable home does not stop the others |
| `startup_reports_but_never_repairs_a_broken_file` | broken `config.toml` reported, never rewritten |
| `startup_provisions_the_discovered_roster` | discovery + provisioning wired, `ctrl` gets no home |
| `excludes_ctrl_and_non_assistants` | the lineage rule |

The failure case uses a regular file sitting where the home belongs — the portable stand-in for a read-only filesystem or denied permission, neither of which simulates reliably in CI.

Full crate suite, parallel (exit 0):

```
test result: ok. 3203 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 27.57s
test result: ok. 6 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 3.73s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.50s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.13s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.12s
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.12s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.42s
```

Full crate suite, `--test-threads=1` (exit 0):

```
test result: ok. 3203 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 58.73s
test result: ok. 6 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 4.54s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.82s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.13s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.15s
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.55s
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.49s
```

No existing test was weakened, ignored or deleted; `bundled_personas_pin_git_reach` and the floor/persona-seed tests pass unchanged. `ASSISTANT_REACHABLE_SUBAGENTS`, `delegate_allowed` and the delegate allow-set are untouched, as are `mpm_bridge.rs` / `claude_mpm_role.rs`.

Gates:

```
$ cargo fmt --check                                          # clean, no diff
$ cargo clippy -p trusty-agents --all-targets -- -D warnings  # clean, 0 warnings
$ bash scripts/check_sld.sh
sld-lint: scanned 50 spec doc(s) + 2991 code file(s); 0 error(s), 0 warning(s)
```

No frontend files were touched, so the UI gates are unchanged (CI runs them regardless).

Rebased onto `origin/main` (was 7 behind); 4 commits replay cleanly.

One flake worth recording as PRE-EXISTING and unrelated: an early `--test-threads=1` run failed `api_e2e::test_health_returns_version` with "api server did not become healthy within 5s". It passes alone and in every full serial suite since, and this PR touches no API/server code — a cold-start timing flake in that test's own 5s probe budget.

Changelog: `crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md`, including the user-visible first-launch note.

Closes #4325

### DIFF
--- a/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
+++ b/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
@@ -13,9 +13,12 @@ Added
   - `AssistantHome` — the app-generated, DOTLESS, human-browsable home at
     `~/trusty-agents/<instance>/` (override with
     `TAGENT_ASSISTANTS_DIR`), holding `instructions.md`, `config.toml`,
-    `agents/`, `okg/` and `attachments/`. The store path is the owner's,
-    verbatim (2026-08-01): `trusty-agents/<agent>/okg/`, indexed by
-    trusty-search. `ensure()` is additive and idempotent
+    `agents/`, `okg/`, `attachments/` and `stores/`. Both store paths are the
+    owner's, verbatim (2026-08-01): `trusty-agents/<agent>/okg/` (indexed by
+    trusty-search) and `trusty-agents/<agent>/stores/<store-identifier>/` (one
+    subdirectory per remote store). This change creates the `stores/` PARENT
+    only — store-identifier derivation, extraction state and extraction logic
+    belong to a separate spec. `ensure()` is additive and idempotent
     — it creates what is missing and never overwrites what a user edited,
     because #4325 makes external modification expected rather than an error.
   - `AssistantInstanceId` — a validated instance name. It becomes a directory
--- a/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
+++ b/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
@@ -11,9 +11,11 @@ Added
   [#3890](https://github.com/bobmatnyc/trusty-tools/issues/3890)). The new
   module supplies the missing container:
   - `AssistantHome` — the app-generated, DOTLESS, human-browsable home at
-    `~/trusty-agents/assistants/<instance>/` (override with
+    `~/trusty-agents/<instance>/` (override with
     `TAGENT_ASSISTANTS_DIR`), holding `instructions.md`, `config.toml`,
-    `agents/`, `okg/` and `attachments/`. `ensure()` is additive and idempotent
+    `agents/`, `okg/` and `attachments/`. The store path is the owner's,
+    verbatim (2026-08-01): `trusty-agents/<agent>/okg/`, indexed by
+    trusty-search. `ensure()` is additive and idempotent
     — it creates what is missing and never overwrites what a user edited,
     because #4325 makes external modification expected rather than an error.
   - `AssistantInstanceId` — a validated instance name. It becomes a directory
--- a/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
+++ b/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
@@ -1,0 +1,33 @@
+Added
+
+- **`assistants` — per-assistant home directory + OKG store model (issue
+  [#4325](https://github.com/bobmatnyc/trusty-tools/issues/4325)).** "Assistant"
+  is a TYPE (`[agent] role = "assistant"`); `izzie` and `cto-assistant` are
+  INSTANCES of it, not distinct agent types. Until now nothing modelled the
+  instance half: a persona lived in the shared agents directory, the OKG tree in
+  the flat `<knowledge_dir>/<slug(agent)>` pool addressed by naming convention,
+  and `[[stores]]` had no field naming a per-assistant root at all (the
+  data-model gap recorded in
+  [#3890](https://github.com/bobmatnyc/trusty-tools/issues/3890)). The new
+  module supplies the missing container:
+  - `AssistantHome` — the app-generated, DOTLESS, human-browsable home at
+    `~/trusty-agents/assistants/<instance>/` (override with
+    `TAGENT_ASSISTANTS_DIR`), holding `instructions.md`, `config.toml`,
+    `agents/`, `okg/` and `attachments/`. `ensure()` is additive and idempotent
+    — it creates what is missing and never overwrites what a user edited,
+    because #4325 makes external modification expected rather than an error.
+  - `AssistantInstanceId` — a validated instance name. It becomes a directory
+    name, so it is checked (no separators, no `.`/`..`, no uppercase) and
+    rejected rather than silently slugged.
+  - `assistants::inspect()` — the DETECTION half of #4325's resilience
+    requirement: structured findings for missing / wrong-kind / unreadable /
+    malformed entries, each carrying a remedy, plus `HomeHealth::narration()`
+    as the seam the concierge narrates from
+    ([#4320](https://github.com/bobmatnyc/trusty-tools/issues/4320)).
+- **`[[stores]] root` — the per-assistant OKG tree root.** A new optional,
+  relative, traversal-free path on `AgentStoreBinding`, resolved and confined by
+  `AssistantHome::store_root()` and defaulting to the home's `okg/` directory.
+  `AgentStoreBinding::validate()` reports an absolute or `..`-bearing value as a
+  store problem instead of failing the agent's boot, matching the module's
+  existing fail-soft posture. Existing bindings that omit `root` parse and
+  validate exactly as before.
--- a/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
+++ b/crates/trusty-agents/changelog.d/4325-assistant-home-store-model.md
@@ -36,3 +36,29 @@ Added
   store problem instead of failing the agent's boot, matching the module's
   existing fail-soft posture. Existing bindings that omit `root` parse and
   validate exactly as before.
+- **Assistant home directories are now created at app startup (issue
+  [#4325](https://github.com/bobmatnyc/trusty-tools/issues/4325)).**
+  **User-visible on first launch after upgrading:** `tagent` creates
+  `~/trusty-agents/<instance>/` for each Assistant instance it finds — a
+  directory tree that did not exist before. It is dotless and browsable on
+  purpose. Each creation is logged at `info` (`created assistant home
+  directory`); a home that cannot be created is logged at `warn` with its
+  reason and remedy.
+  - Wired at `runtime::startup::run_startup_init`, alongside the existing
+    bundled-agent deploy and before the `--api`/`--serve` early dispatch, so one
+    call covers the API server, the Tauri GUI (which spawns `--api` as a
+    sidecar), the `start` daemon, the REPL/TUI and every one-shot subcommand.
+    `config` and `mcp-serve` return earlier and are not covered; neither touches
+    an assistant home.
+  - **It cannot fail startup.** `provision_startup_homes()` returns a report,
+    never a `Result` — a read-only `$HOME`, a denied permission, a full disk, or
+    a file left where a directory belongs leaves the app running and degraded,
+    with the problem surfaced through `HomeHealth::narration()`.
+  - Creation only: it makes what is missing and never rewrites a file you
+    edited. Nothing is moved, renamed or cleaned up, and the existing dotted
+    `~/.trusty-agents` tree is untouched — relocating a home needs a migration
+    process that remains future work.
+  - `roster::discover_instances()` decides what counts as an instance:
+    `role = "assistant"` AND either being the base `assistant` or declaring
+    `extends = "assistant"`. `ctrl` declares the role but is the concierge, not
+    a selectable instance, and gets no home.
--- a/crates/trusty-agents/src/assistants/error.rs
+++ b/crates/trusty-agents/src/assistants/error.rs
@@ -1,0 +1,52 @@
+//! Typed errors for the per-assistant home model (#4325).
+//!
+//! Why: The home directory is USER-FACING and externally modified by design
+//! (#4325's "Design Rationale: Visibility and Resilience"), so its failures are
+//! things a concierge has to EXPLAIN, not opaque strings a caller can only
+//! bubble. Matchable variants let the narration layer say "your home directory
+//! has no `$HOME` to hang off" differently from "`izzie/..` is not a legal
+//! instance id" without string-sniffing an `anyhow::Error`.
+//! What: [`AssistantError`] — the errors raised while RESOLVING or CREATING a
+//! home. Errors while INSPECTING an existing home are not errors at all: they
+//! are findings, and live in [`super::health`] instead, because a malformed
+//! home must never fail a caller that only wanted to report on it.
+//! Test: `super::tests::instance_tests::rejects_path_separators`,
+//! `super::tests::home_tests::ensure_is_idempotent`.
+
+use std::path::PathBuf;
+
+use thiserror::Error;
+
+/// A failure resolving or creating an assistant instance's home directory.
+///
+/// Why/What/Test: see this module's doc comment.
+#[derive(Debug, Error)]
+pub enum AssistantError {
+    /// #4325: the instance id would not be safe as a single directory name.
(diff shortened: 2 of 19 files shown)


bobmatnyc:
The owner has confirmed `personal-assistant` is a template (2026-08-01, filed as #4590), so the judgment call flagged here — excluding it from `is_instance()`/instance discovery — is confirmed correct and stands as-is.

Follow-up filed separately: #4591 notes that today's exclusion works only because `personal-assistant` *happens* to extend nothing, with no explicit template marker, so a future `extends = "assistant"` edit could silently start provisioning it a home. Not urgent, not a blocker for this PR — flagging for awareness and future hardening.

Issue #2919, opened 2026-07-17
feat(trusty-mpm): automatic worktree reclamation — prune on PR-merge/session-decommission + tm doctor disk check (1.1 TiB leak post-mortem)


bobmatnyc:
2026-07-17 audit: ~1.5 TB of git worktrees had accumulated for trusty-tools; a cleanup pass reclaimed ~1.1 TiB (df Used 2.2Ti→1.1Ti) by removing 63 stale worktrees whose PRs were already merged. Root causes, confirmed with evidence:

1. **PR-merge does not trigger worktree removal.** 63 merged-PR worktrees sat untouched (oldest: #1945-era), each carrying a fully regenerable Cargo target/ of 3–63 GB (~83% of the bloat, ~1.25 TB). Even sessions that nominally clean up leak: one merge-watch agent REPORTED removing feat-2903 (63G) but it was still present — cleanup reports were not verified against disk.
2. **Two overlapping worktree registries never reconcile**: the bare `.base` repo AND a legacy outer non-bare `trusty-tools` clone both spawn worktrees (some physically nested inside each other's dirs, registered to the other repo); neither's cleanup reaches the other.
3. **target/ is never GC'd independently** — no grace-period pass reclaims build dirs from worktrees awaiting removal.
4. **Active sessions leak their own completed sub-worktrees** (e.g. tm-trusty-tools-01's mount-2405, 94G, PR #2528 merged, still present).

Proposed fixes (from the audit, in priority order):
- **tm-side auto-prune (highest leverage)**: on session_decommission AND on detected PR-merge of a worktree's branch, `git worktree remove --force` + `git worktree prune`. Would have prevented ~90% of the bloat. Removal must VERIFY the dir is gone (the false-success report above shows why) and target the correct owning repo (dual-registry gotcha).
- **tm doctor disk check**: flag worktrees whose branch has a merged/closed PR; report total reclaimable bytes.
- **Periodic pass**: `git worktree prune` on both registries + target/-only GC for merged-PR worktrees in the grace window.
- **Structural**: retire the legacy outer non-bare clone — one worktree pool per project.

[...]

Issue #1840, opened 2026-06-30
tm: managed-session lifecycle correctness (orphaned worktrees, resume, stale pane)


bobmatnyc:
## Session Lifecycle & Worktree Correctness

Relates to and partially overlaps #1805, #1806, #1807.

### (MAJOR, relates to #1806) Orphaned Worktrees Not Flagged
~167 worktree directories on disk under `~/trusty-mpm-projects/bobmatnyc/trusty-tools/.worktrees/` vs only 35 tracked session records (~130 orphaned). `tm doctor` does not flag this and no prune is surfaced.

**Fix:** Add a `tm doctor` orphan check (compare .worktrees count vs active records), surface `tm sessions prune`, and ensure decommission removes the worktree dir.

### (MAJOR) Resume Fails on Sessions With No Prior Conversation
Resume always invokes claude with `--continue`; for a session with no prior conversation this fails with "No conversation found to continue" and silently drops to a shell, while `tm sessions ls` still shows state=active.

**Fix:** Fall back to a fresh launch (no --continue) when there's no conversation; reflect real state (idle-shell, not active).

### (MODERATE) Missing Pane Snapshot on Stop
`tm sessions activity` returns no `raw_pane` once the runtime is stopped. Store the last-seen pane snapshot at stop time and return it with a `pane_stale: true` flag.

PR #4118, opened 2026-07-27
fix(trusty-mpm): refuse to reclaim a worktree holding unsaved work (#4091)

bobmatnyc:
## What

Adds the dirty-tree safety check the worktree reclaim path never had, and makes
destroying unsaved work require an explicit, separate opt-in.

Closes #4091.

## The defect, confirmed against the code

The issue's analysis holds exactly. Verified before changing anything:

- **The #3649 owner/sentinel guard works as described and is untouched by this
  PR.** `SentinelOwner::Unknown` → `owner_unknown`, reported but never fed to
  the deletion loop (`prune.rs:889-897` pre-change). `find_orphaned_worktrees`
  does walk all five shapes including `.base/.claude/worktrees/`. Owner-unknown
  ad-hoc worktrees were, and remain, never auto-deleted.
- **The gap was exactly the known-owner + dirty case.** Nothing in the reclaim
  path inspected `git status`, the index, or untracked files.
  `session_context_pause` prunes by default (`session_dispatch.rs:181-184` →
  `mcp_context.rs:172-197`, trailing `false` = not a dry run), so this fired on
  an ordinary `/tm-session-pause`.
- **One thing the issue did not say, which raises the stakes on the unpushed
  leg:** `decommission::remove_session_worktree` runs
  `git branch -D session/<leaf>` after a successful removal
  (`decommission.rs:100-120`). Committed-but-unpushed work is therefore not
  merely "left detached in the object store" — it loses its last reachable ref
  and becomes garbage-collectable. The unpushed check is load-bearing, not
  defence in depth.

## How "dirty" is defined

`session_manager::worktree_safety::inspect_dirt` returns `None` **only** when a
candidate is provably free of unsaved work. Dirty means any of:

1. **Working tree** — `git status --porcelain` reports an entry: modified or
   staged tracked files, and untracked-but-not-ignored files. Gitignored files
   are excluded (build artefacts are not work).
2. **Unpushed commits** — `<upstream>..HEAD` when the branch has an upstream;
   with no upstream, commits reachable from `HEAD` but from no remote-tracking
   ref (`rev-list --count HEAD --not --remotes`). In a repo with no remotes at
   all, `--remotes` expands to nothing and every commit counts as unpushed.
3. **Not a git worktree root, and not empty** — git cannot speak to it, so only
   an empty leftover shell (nothing but the ownership sentinel) is treated as
   clean. This preserves the #1838 case the sweep exists for.
4. **Any error** — see below.

[...]

PR #4525, opened 2026-08-01
feat(trusty-mpm): automatic worktree reclamation — merged-PR prune + tm doctor disk check (#2919)

bobmatnyc:
Partially addresses #2919 — see "What this does NOT deliver" below. #2919 stays open.

#2919 was believed fixed by #3661 and #4240. It was not — those landed SPECIFICATIONS (DOC-48, DOC-52). Zero reclamation logic shipped. This lands the survey machinery and the merged-PR reclaim pass.

## Context

`.base/.claude/worktrees` reached **1.1 TiB across 77 entries** on 2026-07-21, four days after a manual sweep reclaimed the same terabyte. Nothing was leaking. Worktrees were never reclaimed, because the merge of a workstream's PR — the trigger DOC-52 §3.4 specifies — fired no cleanup. And nothing reported a byte.

**Already shipped, unchanged here:** reclamation on session decommission (#1840/#3649/#4091). Verified, not duplicated.

## Review round 2 — what the critic found and what changed

The first revision's `classify` design held up: all four gates caught their mutations. But **the three pre-delete re-checks — the entire TOCTOU defense — all survived.** Deleting all three left 31/31 tests green.

**Root cause.** The survey byte-walks every registered worktree; the critic measured that it does not finish in 600 seconds here. The delete loop then called `is_live(&path, in_use)` with the *survey's own captured slice* — minutes old. A session attaching mid-sweep was invisible. The same staleness defeated `git worktree lock`: `Admission` was evaluated at survey time, and `remove_session_worktree`'s `remove_dir_all` fallback deletes locked worktrees regardless. And the one test claiming to cover this fed the same `in_use` to the survey, so the candidate was blocked during classification and the delete loop never ran — its name asserted the opposite of what it did.

**Fix.** `recheck_before_delete` re-reads liveness, git's own registry, the ownership marker, the pull-request state, and the working tree **per candidate, immediately before that candidate's delete** — mirroring `prune_orphaned_worktrees` Phase 2, whose comment warns against collapsing exactly this boundary. An unreadable live session set REFUSES rather than reading as "nothing claims it". `FreshProbes` makes freshness an injected capability, so tests change the workspace *between* survey and delete.

### Mutation results after the fix

| Mutation | Result |
|---|---|
| M8 delete liveness re-check | `FAILED. 49 passed; 2 failed` |
| M9 delete git/lock re-check | `FAILED. 47 passed; 4 failed` |
| M10 delete dirt re-check | `FAILED. 49 passed; 2 failed` |
| M11 unreadable live set fails open | `FAILED. 49 passed; 2 failed` |
| M12 delete ownership re-check | `FAILED. 48 passed; 3 failed` |
| M13 delete PR re-check | `FAILED. 46 passed; 5 failed` |
| M14 delete classify ownership gate | `FAILED. 44 passed; 7 failed` |
| M15 index fork rows instead of skipping | `FAILED. 43 passed; 8 failed` |

All eight caught. The files were restored and re-verified green (51/51) afterwards.

### The other findings

3. **`tm doctor` advertised a reclaim that cannot fire.** The classifier now applies the *remover's own* ownership predicate, so `.claude/worktrees/` is reported but excluded from `reclaimable_bytes`. Doctor no longer names a command that would fail with `removal_failed`.
4. **`worktree_disk` could return `Ok` on a terabyte.** Now `Unknown`, never `Ok`, when the walk is incomplete — an undercounted figure made the check most confident exactly when least justified.
5. **Fork PRs attributed to local branches.** `gh` is asked for `isCrossRepository`; cross-repository rows are dropped.
6. **`spawn_blocking` + no subprocess timeout.** Every `gh` call now carries a 10s timeout with its pipes drained on separate threads (an undrained pipe blocks the child and defeats the timeout).
7. **Unverified PR.** Rebased onto current main (conflict in the doctor check roster resolved: main's `asset_tier` + this PR's `worktree_disk` → 26). `check_test_pointers.sh` now included in the gate — it caught two more dangling pointers, both fixed.
8. **400-PR window.** The reclaim path resolves unreachable branches with a targeted per-branch query; doctor discloses its window rather than paying that cost inside a 3s budget.

## What this does NOT deliver (why "Partially addresses")

Per the scope ruling:

- **No merge-triggered cleanup.** Deletion requires `tm session prune-worktrees --merged-prs`. Nothing automatic fires it.
- **No reclamation of `.claude/worktrees/`** — out of scope per ADR-0020, and that is where the terabyte actually sat.
- **The 400-PR window** still can't reach old worktrees from the bulk index; only the per-branch fallback does, one call at a time.

The survey machinery, the safety gates, and the doctor check stand on their own merits. A follow-up comment on #2919 names the remainder.

## What this code refuses to delete

Every refusal has a test that fails if its guard is removed — at classification (`classify`, five gates: git admission, liveness, removability, merged-PR evidence, unsaved work) and again at deletion (`recheck_before_delete`, re-read fresh). Refusals: uncommitted/untracked files, unpushed commits, a session claiming the path, the main checkout, bare records, operator-locked worktrees (before *and* after the survey), worktrees outside the managed project, worktrees trusty-mpm did not provision, an open PR, a closed-unmerged PR, no PR, an indeterminate PR state, a fork PR, a truncated PR page, a detached HEAD, a reused branch whose merged PR is shadowed by an open one, an unreadable live-session set, and a survey that ran out of time.

## Round 3 — the feature had never worked

`worktree_reclaim.rs` built `gh -C <dir> pr list …`. **`gh` has no `-C` flag.** Verified against gh 2.96.0:

```
$ gh -C . pr list --state merged --limit 1 --json number
unknown shorthand flag: 'C' in -C          # exit 1
$ gh --help | grep -c '\-C'
0
$ gh pr list --state merged --limit 1 --json number,headRefName
[{"headRefName":"docs/tier-audit-untracked-parity","number":4527}]
```

The failure was silent by design, which is why nothing caught it: failed call → `PrIndex::unavailable()` → every branch `Unknown` → gate 4 blocks every candidate → `approved` always empty → the delete loop never executed. `--merged-prs` reclaimed zero bytes always, and `tm doctor` blamed `gh auth status` for an argv bug. Fixed with `cmd.current_dir(dir)`.

**Why 51 tests, 8 mutations and 24 CI checks were blind to it:** "nothing is reclaimable" is also the correct answer for a workspace with nothing to reclaim. Only the argv distinguishes them, and no test ever executed a successful `gh` call.

### Acceptance criterion — non-zero reclaimable against a real store

```
--- #2919 e2e survey of /Users/masa/trusty-mpm-projects ---
elapsed           = 116.99s
candidates        = 269
reclaimable       = 5
reclaimable_bytes = 19085373619        # 17.8 GiB
pr_state_unknown  = 169
unmeasured        = 268
RECLAIMABLE .base/.worktrees/fix-4163          branch=fix/4163-agentid-presence-check        pr=Merged{4464} bytes=Some(19085373619)
RECLAIMABLE .base/.worktrees/fix-transcripts   branch=fix/managed-session-transcript-saving  pr=Merged{4475}
RECLAIMABLE .worktrees/feat-claude-md-named-sections            branch=feat/claude-md-named-sections            pr=Merged{4324}
RECLAIMABLE .worktrees/feat-framework-plain-speaking-instruction branch=feat/framework-plain-speaking-instruction pr=Merged{4316}
RECLAIMABLE .worktrees/fix-tm-ls-latency       branch=fix/tm-ls-cold-latency                 pr=Merged{4326}
```

Reproducible: `TM_2919_E2E_ROOT=~/trusty-mpm-projects cargo test -p trusty-mpm --lib e2e_survey_against_a_real_store -- --nocapture`. Env-gated so CI without `gh` credentials stays green; it asserts PR state resolved for at least one worktree, which is the assertion the `-C` bug would have failed.

### Two further defects the e2e run exposed

Neither was visible from unit tests — both needed a real store.

- **`SurveyBudget::measure` was an instant fixed at survey start.** Measurement is a second phase and classification takes ~90 s, so it had always expired before the phase began: `unmeasured = 269/269`, `total_bytes = 0`. It is now a duration allotted to the measurement phase.
- **Scan-ordered measurement spent the budget on blocked worktrees**, reporting `reclaimable_bytes = 0` beside 5 reclaimable ones — a wrong number, not a partial one. Reclaimable candidates are measured first.

### The rest of round 2

- **Liveness ordering.** The pre-delete `in_use_now` read moved below the `gh` lookups; the snapshot is now single-digit ms old when judged instead of aged by a 322 ms–1.2 s network call.
- **Argv coverage.** `gh_command_passes_no_dash_c_flag` asserts on `get_args()` (both prior tests checked only `get_envs()`); `gh_command_runs_in_the_requested_directory` pins the repository selection. The misleading comment on `pr_state_for_branch_maps_a_failed_call_to_unknown` is corrected — it blamed repo resolution for a flag-parse death and would have passed against a valid repo.
- **MEDIUM 1.** `Handle::try_current` makes the "unreadable live set refuses" branch reachable, instead of `Handle::current` panicking off-pool and unwinding mid-sweep.
- **MEDIUM 2.** `the_remover_really_refuses_what_tm_provisioned_rejects` calls the actual remover and asserts it refuses and leaves the directory, tying the classifier's predicate to real behaviour rather than a copy of its rules.

## Round 4 — partial-measurement disclosure and cause split

**HIGH: `reclaimable_bytes` was a floor wearing the whole set's label.** Ordering the measurement pass reclaimable-first moved the degradation from 0-of-5 to 1-of-5; it did not remove it, and the doc asserted otherwise. One 17.8 GiB worktree consumed the entire 20 s budget, and because every other measurement was starved too the figure happened to equal `total_bytes` — reading as "all of it is reclaimable". The whole-survey UNDERCOUNT clause qualifies `total_bytes` only.

`ReclaimSurvey::reclaimable_measured` now carries the measured-of-total count, and `tm doctor`, the CLI, and the HTTP route all print it beside the byte figure. Doc claim corrected to what is true: this **orders** the degradation, it does not remove it.

**LOW: unknown-cause split.** `ReclaimSurvey::not_inspected` separates "the probe ran out of time" from "the lookup could not answer". They read identically but need opposite actions, and conflating them cost a review round.

### e2e re-run, with the new qualifiers

```
--- #2919 e2e survey of /Users/masa/trusty-mpm-projects ---
elapsed           = 108.57s
candidates        = 269
reclaimable       = 5
reclaimable_bytes = 19085373619 (across 1 of 5 measured)
total_bytes       = 19085373619
pr_state_unknown  = 169 (of which not_inspected = 0, lookup-unresolved = 169)
unmeasured        = 268
```

`not_inspected = 0` independently confirms your finding: all 169 are lookup-unresolved (unreachable repositories + detached HEADs), none are deadline skips.

### The MEDIUM I did not apply, with evidence

You read `remove_session_worktree` as returning `false` when git's removal fails but `remove_dir_all` succeeds, routing a genuinely-deleted worktree into `removal_failed`. It does not. Those `false` values at `decommission.rs:212`/`:224` bind `git_success` for the prune/branch-`D` step; the function falls through to its final `true` at `:289`.

Rather than argue from a reading, `remove_session_worktree_reports_success_when_the_fallback_removes_it` forces the exact condition with a real git-locked worktree (`git worktree remove --force` exits 128, fallback deletes it) and asserts both the return value and that the directory is gone. Green. The reclaim loop's `&&` is now pinned either way. If you still read it as a defect, that test is where to point.

### Other round-3 items

- **`pr_index_from_gh_reads_this_repository` skips on a fork.** It resolves the repository first and skips unless it is `bobmatnyc/trusty-tools`, so it cannot go red for a contributor whose fork legitimately has zero PRs.
- **Docs:** `from_gh` no longer documents the `-C` flag that does not exist; the doctor timeout message names the real 8 s ceiling (3 classify + 3 measure + 2 grace) instead of 3 s.

## Verification

```
cargo test -p trusty-mpm --no-fail-fast
  7210 passed; 0 failed; 16 ignored (44 binaries); exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings   exit 0
cargo fmt --check                                         exit 0
bash scripts/check_line_cap.sh        7 allowlisted, 0 violations — OK
bash scripts/check_test_pointers.sh   0 dangling pointers — OK
bash scripts/check_changelog_fragment.sh   OK trusty-mpm
gh pr checks 4525 --watch --fail-fast     exit 0 — 24 pass, 1 skipping, 0 fail
```

No `#[ignore]`, no cfg-gating, no `--lib` narrowing.

Specs followed: **ADR-0023** (git = existence, sentinel = ownership, fail closed), **ADR-0020** (owner-gated reclamation; `.claude/worktrees/` out of scope), **DOC-52 §3.4 / AC-2.6** (workstream close triggers reclamation; non-destructive at the git level).


bobmatnyc:
## Verdict: BLOCK

Adversarial review at head `b2424a84`, run read-only in an isolated worktree off the PR head. This code deletes directories, so the standard is that every delete-authorizing path is *verified*, not merely present.

**The four `classify` gates hold up. The three pre-delete re-checks do not — all three can be deleted and the suite stays green.** That is the entire safety story of the destructive path, and none of it is pinned.

### Mutation test — every guard, individually removed, `cargo test -p trusty-mpm --lib worktree_reclaim`

Baseline: `test result: ok. 31 passed; 0 failed`.

| # | Guard removed | Result | Test that caught it |
|---|---|---|---|
| M1 | gate 1 admission (`worktree_reclaim.rs:369`) | CAUGHT | `classify_blocks_non_admitted_worktree` |
| M2 | gate 2 liveness in `classify` (`:374`) | CAUGHT | `classify_blocks_live_session_workspace`, `survey_spares_a_merged_worktree_a_session_still_claims` |
| M3 | gate 3 `Open` arm (`:382`) | CAUGHT | `classify_blocks_open_pr` |
| M4 | gate 3 `ClosedUnmerged` arm (`:385`) | CAUGHT | `classify_blocks_closed_unmerged_pr` |
| M5 | gate 3 `NoPr` arm (`:388`) | CAUGHT | `classify_blocks_no_pr` |
| M6 | gate 3 `Unknown` arm (`:391`) | CAUGHT | `classify_blocks_unknown_pr_state` |
| M7 | gate 4 dirt in `classify` (`:401`) | CAUGHT | `classify_blocks_dirty_worktree` + 2 real-git tests |
| **M8** | **`is_live` re-check before delete (`:739`)** | **SURVIVED** | — 31 passed |
| **M9** | **`git_worktree_list_agrees` re-check (`:744`)** | **SURVIVED** | — 31 passed |
| **M10** | **`inspect_dirt` re-check before delete (`:751`)** | **SURVIVED** | — 31 passed |
| M11 | truncation rule `rows.len() < limit` (`:202`) | CAUGHT | `pr_index_truncated_reply_makes_absent_branches_unknown` |
| M12 | absent-branch-when-incomplete → `NoPr` (`:280`) | CAUGHT | 3 tests |
| M13 | `unavailable()` marked complete (`:175`) | CAUGHT | 3 tests |
| M14 | survey deadline fail-closed (`:609`) | CAUGHT | `survey_past_its_deadline_reclaims_nothing` |
| M15 | `block_rank` merged outranks open (`:122`) | CAUGHT | `pr_index_open_pr_beats_a_merged_one_on_the_same_branch` |

12 of 15 caught. The three that survived are the whole TOCTOU defense. `refused_at_recheck` appears nowhere in the crate's tests except a hand-written JSON literal at `managed_merged_prs_tests.rs:42`, which exercises the *printer*, not the guard.

The approval path is genuinely tested — `classify_allows_clean_pushed_merged_worktree` and `reclaim_remove_mode_reclaims_a_clean_merged_worktree` (which really deletes). That part of the claim holds.

---

## Findings

|---|---|---|---|---|
| CRITICAL | `session_manager/worktree_reclaim_tests.rs` | 527 | `reclaim_remove_mode_refuses_a_worktree_a_session_claims` is vacuous — it feeds the same `in_use` to the survey, so gate 2 blocks the candidate and the delete loop never runs. Its doc says "the per-candidate re-check — not the survey — is what has to catch that, so this fails if the re-check is dropped." Proven false by M8. | Survey with `&[]`, then call `reclaim_with_index` with a *different*, claiming `in_use`. Same shape needed for M9 and M10. |
| CRITICAL | `session_manager/worktree_reclaim.rs` | 719, 739 | The liveness re-check is **not fresh**. `reclaim_with_index` re-evaluates `is_live` against the same `in_use` slice the survey used, and the survey in between runs with `deadline: None` and byte-walks every registered worktree first. Measured below: >10 min on this machine. | Take `in_use` as `&dyn Fn() -> Vec<PathBuf>` and call it inside the loop; have the route re-read the store, as `prune_orphaned_worktrees` Phase 2 already does. |
| HIGH | `daemon/doctor_worktree_disk.rs` | 129-140 | Doctor counts `.claude/worktrees/` bytes as `reclaimable` and points at a command that cannot reclaim them. Proven empirically below. On this machine that is where most of the 1.1 TiB sits. | Apply the sentinel / `.worktrees/` ownership rule as a 5th gate in `classify` (`Blocked: "harness-owned, out of scope per ADR-0020"`), so the reclaimable figure means what the message says. |
| HIGH | `daemon/doctor_worktree_disk.rs` | 41, 142-146 | Status is computed from undercounted figures, so the probe can return **`Ok`** on a terabyte. A full walk of the 46 registered worktrees here exceeds 600 s; the budget is 3 s. Deadline-blocked candidates get `pr: Unknown` and `Blocked`, so `reclaimable` is ~0 and the check falls through to the `Ok` branch. | Never return `Ok` when `unmeasured > 0` — return `Unknown` (or `Warn`) whenever the survey was truncated. |
| HIGH | `session_manager/worktree_reclaim.rs` | 133-141, 236-238 | `headRefName` alone attributes a **fork** PR to a local branch. `gh` exposes `isCrossRepository` and `headRepositoryOwner`; neither is requested. A merged fork PR on a common branch name (`patch-1`, `main`, `fix/typo`) becomes landing evidence for an unrelated local branch. | Add both fields to `--json`; skip rows where `isCrossRepository` is true, or key the index on `(owner, branch)`. |
| HIGH | `daemon/doctor_worktree_disk.rs` | 192 | `tokio::time::timeout` still wraps `spawn_blocking`, and no `git`/`gh` call inside has a subprocess timeout (`PrIndex::from_gh` and `worktree_safety::git_stdout` both use bare `.output()`). A wedged `gh` re-creates the exact shutdown hang this PR reports fixing — the deadline poll cannot interrupt a subprocess already in flight, which the code comment concedes. | Bound the subprocesses themselves (spawn + kill on deadline), not just the walk. |
| HIGH | `session_manager/worktree_reclaim.rs` | 709 | CI-enforced doc-pointer lint is **red**: the `Test:` pointer cites `reclaim_remove_mode_refuses_a_worktree_dirtied_after_the_survey`, which does not exist. `bash scripts/check_test_pointers.sh` → exit 1. No CI has run on this PR at all (`gh pr checks 4525` → "no checks reported"), and the verification block omits this gate. | Write that test (it is also the M10 fix), then re-run the lint. |
| MEDIUM | `session_manager/worktree_reclaim.rs` | 53 | `--limit 400` against a repo with 4526 PRs: the index is *always* truncated here (verified — 400 rows returned at limit 400). Fail-closed, so it is safe, but it means only worktrees whose PR is in the last ~400 can ever be reclaimed. The long-tail old worktrees that constitute the leak stay permanently `Unknown`. Undisclosed. | Query per-branch for candidates the bulk index missed, or state the limit in the doctor message. |
| MEDIUM | `session_manager/worktree_reclaim_tests.rs` | — | No test parks a worktree at a `.claude/worktrees/`-shaped path through the reclaim path, despite `GitWorktreeFixture::add_worktree_at` existing for exactly that. Every reclaim test uses `add_worktree`, which parks under `.worktrees/` and always passes the ownership gate. | Add the empirical test below as a permanent regression pin. |
| MEDIUM | `session_manager/worktree_reclaim.rs` | 401 | `inspect_dirt`'s documented residual risk — gitignored loose files outside `.trusty-mpm/` (a `.env`, a scratch note) are not dirt — was accepted in #4118 against orphan-owner candidates only. This PR widens the candidate population to *every merged-PR worktree* without re-assessing it. | Name it in the module doc, or count non-disposable ignored entries for this trigger specifically. |
| MEDIUM | — | — | PR conflicts with `main` (`mergeStateStatus: DIRTY`) on `daemon/doctor.rs`, `doctor_tests.rs`, `api_tests.rs`, `generate/doctor.rs` — the doctor check-count assertion. | Rebase; `run_doctor_produces_twenty_five_checks` will need renumbering again, and the whole suite re-run. |

### Empirical proof of the `.claude/worktrees/` overcount

Temporary test, `add_worktree_at(<repo>/.claude/worktrees, "harness-owned")`, landed and pushed, merged-PR index, `ReclaimMode::Remove`:

```
candidate = ReclaimCandidate { path: ".../.claude/worktrees/harness-owned",
                               branch: Some("wt/harness-owned"), bytes: Some(137),
                               pr: Merged { pr: 99 }, verdict: Reclaimable { pr: 99 } }
survey.reclaimable = 1   reclaimable_bytes = 137
removed = []
removal_failed = [".../.claude/worktrees/harness-owned"]
still on disk = true
```

`tm doctor` will print those bytes as "sits on branches whose pull request already merged … reclaim with `tm session prune-worktrees --merged-prs`", and that command cannot reclaim them. The PR body says this store is "surveyed and reported but never deleted" — accurate about the deletion, wrong about the reporting: it is reported *as reclaimable*.

### TOCTOU window, measured

The claim is that approved candidates are re-checked "immediately before their own deletion". The re-check runs immediately before the delete; its **input** does not.

`reclaim_with_index` (`:719`) runs `survey_with_index(..., None)` — no deadline — which for every registered worktree runs `classify` and then `measure_bytes_until`, a full recursive byte walk. Only after that entire pass does the delete loop start, and it calls `is_live(&path, in_use)` with the identical slice.

I walked the 46 worktrees `git worktree list --porcelain` reports here, summing file sizes the way `measure_bytes_until` does. **It did not finish in 600 s** (killed at the 10-minute ceiling). The implementer independently measured >20 minutes over the same store.

So the window between the liveness snapshot and the first `rm` is **more than ten minutes on this operator's own machine, and unbounded in general**. A session that attaches inside that window is invisible to the re-check. `inspect_dirt` catches it only once it has written a non-ignored file; a freshly-attached session, or one whose writes are all gitignored, is not caught. The same staleness defeats `git worktree lock`: `Admission::Locked` is evaluated only during the survey, `git_worktree_list_agrees` does not re-read the lock, and `remove_session_worktree`'s `remove_dir_all` fallback deletes a locked worktree regardless — as `lock_worktree`'s own doc comment states.

Error handling at the re-check is fine, for the record: `is_live` falls back to the raw path on canonicalization failure, `inspect_dirt` fails toward dirty, `git_worktree_list_agrees` refuses on error. Staleness is the defect, not the error paths.

The pre-existing orphan sweep gets this right. `prune_orphaned_worktrees` Phase 2 re-reads the session store immediately before its deletion loop, with a comment explaining that the two independent reads are defense-in-depth and warning against collapsing them. The new path has one read, and it is the older of the two.

### The two scope decisions

**(a) Unattended deletion not enabled.** Correct on its own terms — fail-closed is right, and the PR names the follow-up. But #2919 asks for merge-triggered cleanup, and nothing sets `--merged-prs`.

**(b) `.claude/worktrees/` surveyed but never deleted.** This is where the 1.1 TiB actually was.

Together: the PR delivers neither the automatic trigger nor reclamation of the store that leaked, and the 400-PR window means it cannot reach the old worktrees either. **#2919 should not be closed by this PR.** Drop `Closes #2919`, land the survey machinery as the foundation it is, and keep the issue open for the trigger and the harness-owned store.

### Required before re-review

1. Make M8, M9 and M10 fail. Three tests: survey clean → claim the path at re-check; survey clean → unregister the worktree at re-check; survey clean → dirty the worktree at re-check. Verify each by deleting its guard and watching it go red.
2. Re-read the in-use set inside the delete loop rather than reusing the survey's snapshot.
3. Do not count `.claude/worktrees/` toward `reclaimable` / `reclaimable_bytes`.
4. Never report `Ok` from `worktree_disk` when `unmeasured > 0`.
5. Disambiguate fork PRs via `isCrossRepository` / `headRepositoryOwner`.
6. Fix the dangling doc pointer at `worktree_reclaim.rs:709`; get `check_test_pointers.sh` to exit 0.
7. Rebase onto `main` and get CI green — no checks have ever run on this branch.
8. Drop `Closes #2919`.

### Notes

- What is good here is genuinely good. `classify`'s four gates are correct, reachable only by falling off the end, and each is mutation-proven. The `PrIndex` truncation rule, the `unavailable()` failure mapping, the `block_rank` branch-reuse guard and the deadline fail-closed path are all pinned by tests that go red when the rule is removed. The self-reported `spawn_blocking` defect was a real find and the inner-deadline fix is the right shape.
- Not flagged: `measure_bytes_until`'s `enumerate()` counting only `Ok` entries (measurement only, no decision reads it); `--limit 400` returning the newest PRs (fail-closed).
- Below 80% confidence, so recorded here rather than as a finding: a branch carrying an open PR older than the 400-row window plus a newer merged PR would rank as `Merged`. Requires two PRs from one head branch to different bases, one of them >400 PRs old.

Reviewed read-only in `.claude/worktrees/critic-4525`; no PR branch, worktree, or session state was modified.

bobmatnyc:
## Code Critic — Round 2 Verdict: BLOCK

Independent re-review at `5f594335`. Priority-zero revert check passed completely. The round-1 findings are genuinely fixed — I reproduced every mutation. But the feature does not work in production: `gh` is invoked with a flag `gh` does not have, so no pull request can ever be resolved and nothing can ever be reclaimed.

---

## Priority zero — does this revert #4517? NO. Verified clean on every check.

I did not accept the author's claim; I verified each item independently.

| Check | Result |
|---|---|
| `gh pr diff 4525 --name-only` | 23 files, **all** under `crates/trusty-mpm/` |
| `git diff origin/main...5f594335 -- crates/trusty-installer/` | **EMPTY** |
| Merge base | `63385102` == `origin/main` exactly, so two-dot and three-dot diffs coincide; both empty |
| All 15 files `c219230e` (#4517) touched | **byte-identical** blob SHAs, main vs PR head, enumerated one by one |
| `crates/trusty-installer/src/commands/port_guard.rs` | present on PR head, blob `8b46723f…` on **both** sides |
| Stray `worktree_sweep.rs` | absent from the PR head tree entirely (`git ls-tree -r`), not merely unreferenced |
| Intermediate deletion commit | **none exists** — the branch is a single commit, `5f594335`, so there is no history to flatten |
| Closing keywords in any commit body | none; the only issue reference is `Partially addresses #2919` |
| `gh pr view 4525 --json closingIssuesReferences` | `[]` — queried twice |

I verified the **final tree**, and separately confirmed there is no intermediate commit carrying the deletion. Both are clean. Merging this does not revert #4517.

---

## CRITICAL — `gh -C <dir>` is not a valid invocation; the whole feature is inert

`worktree_reclaim.rs:316-327` builds every `gh` call as:

```rust
let mut cmd = Command::new("gh");
cmd.arg("-C").arg(dir);          // line 318
```

producing `gh -C <dir> pr list --state all --limit 400 --json …`.

**`gh` has no global `-C` flag.** This is a `git -C` habit carried across to `gh`. Verified on the installed binary:

```
$ gh --version
gh version 2.96.0 (2026-07-02)

$ gh -C /…/trusty-tools/.base pr list --state all --limit 5 --json number,headRefName,state,isCrossRepository
unknown shorthand flag: 'C' in -C
Usage:  gh pr list [flags]
$ echo $?
1

$ gh --help | grep -c '\-C'
0

# control, same call without -C, from inside the repo:
$ gh pr list --state all --limit 3 --json number,headRefName,state,isCrossRepository
[{"headRefName":"docs/tier-audit-untracked-parity","isCrossRepository":false,"number":4527,"state":"MERGED"}, …]
$ echo $?
0
```

`worktree_reclaim.rs:318` is the **only** `Command::new("gh")` site in the workspace that uses `-C`. Every other one (`gh_cli/client_impl.rs:269,285`, `workflow/tickets.rs:315`, …) passes `--repo <owner/name>` or relies on cwd.

### What this breaks, end to end

1. `PrIndex::from_gh` — `run_with_timeout` returns `Some((false, _))`, which falls to the `_` arm at `worktree_reclaim.rs:263` → `PrIndex::unavailable()`.
2. `pr_state_for_branch` — same, → `BranchPrState::Unknown` (`worktree_reclaim.rs:424`).
3. Both the bulk index and the per-branch fallback are therefore **always** unresolvable, so every branch reads `Unknown`.
4. `classify` gate 4 takes the `BranchPrState::Unknown` arm (`worktree_reclaim.rs:551`) and blocks. `Reclaimable` is unreachable.
5. `reclaim_with_probes`' `approved` list (`worktree_reclaim_sweep.rs:322-328`) is **always empty**. The delete loop never executes. `tm session prune-worktrees --merged-prs` reclaims zero bytes, always.
6. `tm doctor`'s `worktree_disk` always satisfies `survey.pr_state_unknown == total` (`doctor_worktree_disk.rs:142`) and returns `Unknown` with:

   > "pull-request state resolved for NONE of them … check that `gh` is installed and authenticated (`gh auth status`)"

   `gh` **is** installed and authenticated. The check misdirects the operator to `gh auth status` for a bug that is in the argv — actively worse than silence, and precisely the "advertising a remedy that cannot fire" failure the PR body says it set out to remove.

It fails **closed**, so there is no data-loss path here — that part of the design holds. But the PR's contract is a working merged-PR reclaim pass plus a disk check that names reclaimable bytes, and neither can function on the machine this was written for.

**Fix** — one line:

```rust
// worktree_reclaim.rs:318
- cmd.arg("-C").arg(dir);
+ cmd.current_dir(dir);
```

`gh` resolves the repository from the cwd's git remote, and `GH_STRIPPED_ENV` already removes the `GIT_DIR`/`GH_REPO` overrides that would redirect it — so `current_dir` gives exactly the containment the `-C` was reaching for. Keep the `gh_command_strips_repository_redirecting_env` env assertions as they are.

---

## HIGH — no test executes a successful `gh` call, and both argv tests pass on the bug

This is why 51 green tests and eight caught mutations certified a code path that cannot run.

- `worktree_reclaim_tests.rs:371-386` `gh_command_strips_repository_redirecting_env` inspects `cmd.get_envs()` only. It never touches `cmd.get_args()`, so no argv defect is observable to it.
- `worktree_reclaim_tests.rs:532-541` `pr_state_for_branch_maps_a_failed_call_to_unknown` is **green for the wrong reason**. Its comment reads "No repository here, so `gh` cannot resolve one and fails" — but `gh` rejects `-C` during flag parsing, before it ever attempts repository resolution. The test would pass identically if `tmp.path()` were a valid repository with a merged PR. It asserts nothing about the behaviour it names.

**Fix**: add an argv assertion over `gh_command(…).get_args()` pinning the exact flag sequence, plus one test that runs `PrIndex::from_gh` against a real repository fixture and asserts a non-`unavailable()` index (gate it on `gh auth status` succeeding if CI lacks a token, but do not let the suite pass with zero successful-call coverage).

---

## HIGH — the liveness snapshot is read *before* the slowest work in the window

`worktree_reclaim_sweep.rs:332-355`, per candidate:

```
335   let in_use_now = (probes.in_use_now)();        // t0 — liveness snapshot taken HERE
336-338 fresh_indexes.entry(…).or_insert_with(|| (probes.index_for)(…));   // gh bulk index
339-345 pr_now = …; if Unknown && !complete { pr_state_for_branch(…) }     // gh network call
346   recheck_before_delete(&path, in_use_now.as_deref(), &pr_now)         // liveness judged against t0
355   remove_session_worktree(&path)                                       // DELETE
```

The re-check is genuinely per-candidate and genuinely fresh for git, ownership, and dirt — those are read inside `recheck_before_delete` at line 346. But **liveness is read at line 335, before the `gh` calls**, and only evaluated at 346.

### Measured window (this repo, `.base`, 44 registered worktrees)

| Operation inside the window | Measured |
|---|---|
| `gh pr list --head <branch> --state all --limit 50 --json …` (per-branch fallback, fires for nearly every candidate here since 4526 PRs > `PR_INDEX_LIMIT` 400) | **322–366 ms** |
| `gh pr list --state all --limit 400 --json …` (bulk index, first candidate per registry root) | **1220 ms** |
| `git worktree list --porcelain` (`git_still_permits`) | **< 5 ms** |
| ceiling if `gh` stalls | **10 s** (`GH_TIMEOUT`) |

So liveness is **~350 ms – 1.5 s stale at typical latency, bounded at ~10 s**, plus `inspect_dirt`. Down from round 1's minutes, but still gated on a network call for no reason — the snapshot is taken before the one operation in the loop that can block for ten seconds.

**Fix** — move line 335 to sit immediately above line 346:

```rust
  let index = fresh_indexes.entry(…).or_insert_with(…);
  let mut pr_now = index.state_for(…);
  if pr_now == BranchPrState::Unknown && !index.is_complete() && let Some(branch) = … {
      pr_now = pr_state_for_branch(&candidate.registry_root, branch);
  }
+ let in_use_now = (probes.in_use_now)();   // last thing read before the gate
  if let Some(reason) = recheck_before_delete(&path, in_use_now.as_deref(), &pr_now) { … }
```

That drops the liveness window to the `git`/fs cost inside `recheck_before_delete` — single-digit milliseconds measured. The residual window (recheck → `remove_dir_all`) is then irreducible and small, which is the right place to stop.

---

## MEDIUM — the "unreadable live set REFUSES" guarantee is test-only

M11 passes, and the guard at `worktree_reclaim_sweep.rs:241-243` is real. But its `None` input **cannot occur in production**.

`prune.rs:204-215`:

```rust
let in_use_now = move || -> Option<Vec<std::path::PathBuf>> {
    Some(                                       // <- unconditional
        handle.block_on(mgr_for_probe.list())
            .into_iter().filter_map(|r| r.workspace_path).collect(),
    )
};
```

`SessionManager::list` (`manager.rs:353-366`) returns `Vec<SessionRecord>`, not a `Result` — on a reload error it logs and falls back to `cached_all()`. There is no path that yields `None`.

The underlying safety property is mostly held by a *different* mechanism: `list()`'s last-known-in-memory fallback means a transient store read error degrades to "slightly stale", not "fleet empty". Credit where due — that is the correct design and it is separately tested. But the residual hole is real: on a cold daemon with an unreadable store, `cached_all()` is empty, the closure returns `Some(vec![])`, and that reads as "nothing claims it" — the exact fail-open the guard is advertised to close. A session attached to a merged-PR worktree with a clean tree is the common post-merge state, so this is not an exotic shape.

The PR body's "An unreadable live session set REFUSES rather than reading as 'nothing claims it'" describes the test harness, not production.

**Fix**: give the probe a fallible read — e.g. expose the `store.all()` `Result` (or a `try_list()`) and map `Err` to `None`, so the branch M11 protects is reachable. Alternatively, if the `cached_all()` fallback is considered sufficient, say so in the doc comment and drop the overclaim.

---

## MEDIUM — the ownership-parity test does not test parity

`worktree_reclaim.rs:445-449` re-expresses the remover's gate:

```rust
pub(crate) fn tm_provisioned(path: &Path) -> bool {
    path.join(super::decommission::WORKTREE_SENTINEL_FILE).exists()
        || super::decommission::is_session_worktree(path)
}
```

`remove_session_worktree` (`decommission.rs:144-154`) computes the same two-tier test inline. It does **not** call `tm_provisioned`, so these are parallel expressions that can drift.

`worktree_reclaim_tests.rs:444-465` `tm_provisioned_matches_the_removers_own_predicate` never references `remove_session_worktree`. It asserts `tm_provisioned` against hand-written expectations, so if the remover's gate changes the test still passes and the classifier silently drifts. The name claims a property the body does not check.

Impact is bounded — drift produces `removal_failed` entries or under-reporting, not data loss — which is why this is MEDIUM and not higher.

**Fix**: extract `pub(crate) fn tm_owns(path: &Path) -> bool` into `decommission.rs`; have both `remove_session_worktree` and `tm_provisioned` call it. Then the parity is structural and the test can be deleted.

---

## Verified correct — round-1 findings that are genuinely fixed

I reproduced every mutation myself rather than taking the table on trust. Baseline: `cargo test -p trusty-mpm --lib worktree_reclaim` → `51 passed; 0 failed; 0 ignored`.

| Mutation | My result | Claimed | Tests that caught it |
|---|---|---|---|
| M8 delete liveness re-check | `FAILED. 49 passed; 2 failed` | 49/2 ✓ | `recheck_refuses_a_worktree_a_session_claims_now`, `reclaim_remove_mode_refuses_a_worktree_claimed_after_the_survey` |
| M9 delete git/lock re-check | `FAILED. 47 passed; 4 failed` | 47/4 ✓ | `recheck_refuses_when_git_cannot_be_queried`, `recheck_refuses_a_path_git_no_longer_lists`, `recheck_refuses_a_worktree_locked_after_the_survey`, `reclaim_remove_mode_refuses_a_worktree_locked_after_the_survey` |
| M10 delete dirt re-check | `FAILED. 49 passed; 2 failed` | 49/2 ✓ | `recheck_refuses_a_worktree_dirtied_after_the_survey`, `reclaim_remove_mode_refuses_a_worktree_dirtied_after_the_survey` |
| M11 unreadable live set fails open | `FAILED. 49 passed; 2 failed` | 49/2 ✓ | `recheck_refuses_when_the_live_set_cannot_be_read`, `reclaim_remove_mode_refuses_when_the_live_set_cannot_be_read` |

The three that survived round 1 (M8, M9, M10) now all fail when removed, with the exact counts claimed. The `FreshProbes` seam is **not** a test-only fiction for liveness: `prune.rs:201-215` passes a handle to the manager and re-invokes `mgr.list()` per candidate — a genuine re-read, not a captured snapshot. That was the single most important thing to check and it holds. Tree restored clean after every mutation.

**4. `gh` 10s timeout with drained pipes — correct, not the #4516 defect.** `worktree_reclaim.rs:364-398`. Both pipes are drained on their own threads. The stdout thread does `let _ = stdout.read_to_string(&mut buf); let _ = tx.send(buf);` — on an `io::Error` it still sends the partial buffer, so the channel never goes silent, and the thread returning drops `ChildStdout`, closing the fd, so a child that keeps writing gets `EPIPE` rather than blocking. That is the specific re-armed-hang shape #4516 had, and it is absent here. Timeout kills **and** reaps (`kill()` then `wait()`). Every failure mode — spawn failure, timeout, wait error, `recv_timeout` expiry — collapses to `None` and blocks.

**5. Fork PRs — requested AND acted on.** `PR_JSON_FIELDS` (`worktree_reclaim.rs:338`) includes `isCrossRepository` and is used by **both** `from_gh` (line 260) and `pr_state_for_branch` (line 419), so the two paths cannot drift. It is acted on at `from_json:224` — `if row.is_cross_repository { continue; }` — which leaves the local branch absent, resolving to `NoPr` or `Unknown`, both of which refuse. `pr_index_skips_fork_pull_requests` pins it.

**6. `worktree_disk` never `Ok` on an incomplete walk — correct, including the boundary.** `build_worktree_disk_check` (`doctor_worktree_disk.rs:103-184`) reaches `Ok` only when `unmeasured == 0` **and** `reclaimable == 0` **and** `pr_state_unknown < total`. Candidates past the classify deadline are pushed with `bytes: None` (`worktree_reclaim_sweep.rs:115`), so they force `unmeasured > 0` → `Unknown`. On the deadline boundary: `expired` uses `>=` (`worktree_reclaim.rs:694`), so exactly-at-deadline counts as expired. A walk that completes inside the final `DEADLINE_CHECK_INTERVAL` block after the deadline returns `Some(total)` — correct, because that count is complete; it overruns the budget slightly rather than reporting a floor as a total.

**7. Doctor no longer advertises `.claude/worktrees/`** — gate 3 (`worktree_reclaim.rs:530`) applies `tm_provisioned` during classification, so those worktrees are listed but excluded from `reclaimable_bytes`. Behaviour is right; see the MEDIUM above for the drift risk in *how* the predicate is shared.

**8. `check_test_pointers.sh`** → `test-pointers: 0 dangling pointers — OK.` (exit 0), run from the PR head. Confirmed.

**Scope** — confirmed: the PR body says "Partially addresses #2919", #2919 is **OPEN**, and the comment naming the four remaining items (merge-triggered cleanup, `.claude/worktrees/` out of scope, 400-PR reach, legacy outer clone) was posted 2026-08-01T07:05:10Z.

---

1. **`worktree_reclaim.rs:318`** — replace `cmd.arg("-C").arg(dir)` with `cmd.current_dir(dir)`. (CRITICAL)
2. **Add real `gh` coverage** — an argv assertion over `gh_command(…).get_args()`, and at least one test that exercises a *successful* `PrIndex::from_gh` against a repository fixture. Fix the comment on `pr_state_for_branch_maps_a_failed_call_to_unknown:533`, which currently states a cause that is not the one in effect. (HIGH)
3. **`worktree_reclaim_sweep.rs:335`** — move the `in_use_now` read below the `gh` calls, immediately above `recheck_before_delete`. (HIGH)
4. **`prune.rs:204`** — make the probe's `None` reachable, or drop the "unreadable live set REFUSES" claim from the doc comments and PR body. (MEDIUM)
5. **Extract a shared ownership predicate** into `decommission.rs` so `tm_provisioned` and `remove_session_worktree` cannot drift. (MEDIUM)

Re-verify after (1): run the reclaim pass against this repo and confirm a non-zero `reclaimable` count, and that `tm doctor`'s `worktree_disk` reports something other than "pull-request state resolved for NONE of them". Right now both are structurally impossible, so neither has ever been observed working.

## Notes

The standing "3+ rounds means close and fold" rule does not apply here, and this is not a round-2 rejection of the safety design. The TOCTOU work is real and now genuinely mutation-verified — the guards that were decorative in round 1 have teeth, and I confirmed that independently rather than reading the table. The blocker is a one-line argv bug plus the missing coverage that let it through: the safety machinery is sound, and it is currently guarding a code path that can never reach it. Items 1-3 are a short fix, not a redesign.

bobmatnyc:
## Code Critic — Round 3 Verdict: WARN

Scoped to the delta `5f594335..a90c0b09`. Read-only, in an isolated worktree at the PR head. The round-1/round-2 settled items were not re-litigated.

**The `-C` fix is correct and the feature is genuinely live.** I reproduced it independently rather than accepting the author's numbers. `gh` now resolves the intended repository, `PrIndex::from_gh` returns a populated index, classification produces real `Reclaimable` verdicts, and the delete loop's inputs are all present for the first time. Both `gh` call sites go through `gh_command`, which is the only place `-C` ever appeared; no other call site in the workspace relied on `-C` semantics.

My own e2e run against the real store (`TM_2919_E2E_ROOT=/Users/masa/trusty-mpm-projects`):

```
elapsed           = 115.658970291s
candidates        = 270
reclaimable       = 5
reclaimable_bytes = 19085373619
total_bytes       = 19085373619
pr_state_unknown  = 170
unmeasured        = 269
```

57 tests in the reclaim modules pass, 10 in `worktree_disk` pass, `scripts/check_test_pointers.sh` reports 0 dangling pointers.

---

## The 169 unknowns: EXPECTED. Not a second resolution bug.

I probed every registry anchor in the store directly — `cd <anchor> && gh pr list --state all --limit 400 --json …` with the same environment scrub the module applies. Every unknown falls into one of two categories, both correct.

**Category A — `gh` cannot resolve a repository from that anchor (~158 worktrees, 22 anchors).**

| Anchor | worktrees | `gh` says |
|---|---|---|
| `duettoresearch/cto-reports` | 48 | `Could not resolve to a Repository with the name 'duettoresearch/cto-reports'` |
| `duettoresearch/code-intelligence` | 39 | same shape |
| `duettoresearch/hackathon` | 10 | same shape |
| `duettoresearch/duetto-eve-agents`, `gb-connect-card` | 8 + 8 | same shape |
| `hotstats/*` (6 anchors) | 8 | same shape |
| `bob-duetto/demo-agentic-ux`, `hotstats-product-poc` | 7 | same shape |
| `bobmatnyc/bakeoff-l1…l5` | 5 | `no git remotes found` |
| remaining `duettoresearch/*` | ~35 | same shape |

These are org repositories the currently authenticated `gh` account cannot see, plus five local-only repos with no remote at all. The call fails → `PrIndex::unavailable()` → every branch `Unknown` → blocked. Fail-closed, correct.

**Category B — detached-HEAD worktrees in anchors that DO resolve (19).**

`PrIndex::state_for(None)` returns `Unknown` (`worktree_reclaim.rs:317-319`), and the per-branch fallback is skipped by the `let Some(branch)` guard at `worktree_reclaim_sweep.rs:135`. `.base` alone holds 18 detached worktrees; `trusty-tools` holds 1. No branch means no PR to attribute, so `Unknown` is the right answer.

158 + 19 = 177 against an observed 170 — the residual is my own review worktree plus dedup on (path, registry root). Category B is fully accounted for.

**Category C — "the branch predates the index window and should have resolved" — does not exist.** I sampled six branches whose PRs are far outside the 400-row window and ran the per-branch fallback by hand against `.base`:

```
fix/4058-trusty-mpm-binary-discovery    → #4465 MERGED
feat/changelog-fragments                → #4477 MERGED
fix/4141-session-status-delegation-id   → #4463 MERGED
fix/4163-agentid-presence-check         → #4464 MERGED
fix/4230-orphan-daemon-spawn            → #4466 MERGED
fix/managed-session-transcript-saving   → #4475 MERGED
```

Every one resolves. The fallback works. Nothing that should have resolved is hiding in the unknown bucket.

---

## Delete path — the priority-4 review. Verified.

| Question | Answer |
|---|---|
| Does the per-candidate re-check actually run when `approved` is non-empty? | Yes. `recheck_before_delete` is called unconditionally per candidate at `worktree_reclaim_sweep.rs:394`, and `reclaim_remove_mode_reclaims_a_clean_merged_worktree` executes the loop with a real non-empty `approved` (asserts `removed == vec![path]`, `!path.exists()`, `removed_bytes > 0`). It ran green here. |
| Index reused across iterations? | Yes — `fresh_indexes` is built once per `registry_root` and reused for the whole loop (`:385-387`). Safe in one direction only, which is the safe one: only `Merged` permits a delete, `MERGED` is terminal on GitHub, so a cached verdict cannot decay into a wrongly-permissive one. Truncated-index `Unknown`s fall through to `pr_state_for_branch`, which IS fresh per candidate. |
| Candidate list mutated during iteration? | No. `approved` is an owned `Vec`, iterated by value; `path` is moved out. |
| Error mid-loop leaving partial state? | One reporting inversion — see MEDIUM below. No data-loss shape. |
| Liveness ordering | The reorder is right. `in_use_now` is now read at `:393`, after the `gh` work and immediately before the check that judges it. The snapshot is milliseconds old instead of 322 ms–1.2 s old. |
| Sibling of the "instant fixed at t0 governs a later phase" bug? | None. `classify` is an `Instant` but governs the FIRST phase, so t0 is correct for it. `run_with_timeout` computes its deadline at spawn (`worktree_reclaim.rs:422`). `measure_bytes_until` takes the deadline as a parameter, now computed at phase start. The only instance was the one round 2 found. |

The delete-authorizing path is verified. That is what the round-3 bar required.

---

## Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/session_manager/worktree_reclaim_sweep.rs` | 176-178 | The doc asserts an invariant the code does not have; measured false | Disclose the reclaimable figure as a floor too |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reclaim_sweep.rs` | 403 | A successful fallback deletion is recorded as `removal_failed` | Judge on `!path.exists()` alone |
| MEDIUM | `crates/trusty-mpm/src/session_manager/worktree_reclaim_tests.rs` | 614-635 | `pr_index_from_gh_reads_this_repository` FAILS rather than skips on a fork or any zero-PR repo | Gate the skip on call success, not on index shape |
| LOW | `crates/trusty-mpm/src/session_manager/worktree_reclaim.rs` | 248-253 | `from_gh`'s doc still documents `-C <registry_root>` and still credits the env test with argv coverage | 3-line doc edit |
| LOW | `crates/trusty-mpm/src/daemon/doctor_worktree_disk.rs` | 254-258 | Timeout message says "exceeded 3s" when the real ceiling is now 8s | Format the real ceiling |
| LOW | `crates/trusty-mpm/src/daemon/doctor_worktree_disk.rs` | 130-140 | `pr_state_unknown` conflates "gh could not resolve" with "classify deadline hit", and blames the 400-PR window for both | Count the two separately or soften the attribution |

### HIGH — `worktree_reclaim_sweep.rs:176-178`: the actionable figure DOES degrade

```rust
/// What: two ordered passes over the same slice — reclaimable candidates first,
/// then the rest — both bounded by the same `deadline`. `total_bytes` degrades
/// to a floor under pressure (already disclosed as an UNDERCOUNT); the
/// actionable figure does not.
```

The last clause is false, and my e2e run demonstrates it on the author's own store. With `measure: Some(20s)`:

```
reclaimable       = 5
reclaimable_bytes = 19085373619
unmeasured        = 269
RECLAIMABLE …/.base/.worktrees/fix-4163                            bytes=Some(19085373619)
RECLAIMABLE …/.base/.worktrees/fix-transcripts                     bytes=None
RECLAIMABLE …/.worktrees/feat-claude-md-named-sections             bytes=None
RECLAIMABLE …/.worktrees/feat-framework-plain-speaking-instruction bytes=None
RECLAIMABLE …/.worktrees/fix-tm-ls-latency                         bytes=None
```

The reordering worked — the first reclaimable candidate got measured. But that ONE candidate is 17.8 GiB and consumed the whole budget, so **four of the five reclaimable worktrees are unmeasured** and `reclaimable_bytes` is one worktree's size presented as the set's size. This is the same defect the change set out to fix, moved from 0-of-5 to 1-of-5.

The doctor renders it as (`doctor_worktree_disk.rs:153-164`):

> `17.8 GiB across 270 worktree(s) — an UNDERCOUNT: 269 worktree(s) went unmeasured …; 17.8 GiB of it (5 worktree(s)) sits on branches whose pull request already merged …`

The `UNDERCOUNT` clause qualifies `total_bytes`. Nothing tells the operator that the reclaimable figure is itself a floor over 1 of 5 — and because `total_bytes == reclaimable_bytes` exactly here, it reads as "all of it is reclaimable", which is a wrong-but-plausible number. By the author's own standard in the comment directly above ("not a partial answer, it is a WRONG one"), this is not yet finished.

No deletion gate reads bytes, so this is HIGH and not CRITICAL.

**Fix**, roughly 15 lines:
1. Correct the doc — the reclaimable figure is a floor whenever the budget expires inside the reclaimable pass.
2. Surface a per-set count. The doctor already holds `survey.candidates`, so it can compute it without a new struct field:
   ```rust
   let reclaimable_unmeasured = survey.candidates.iter()
       .filter(|c| c.verdict.is_reclaimable() && c.bytes.is_none())
       .count();
   ```
   and render `at least {bytes} across {reclaimable} worktree(s) ({reclaimable_unmeasured} of them unmeasured)` when it is non-zero.

### MEDIUM — `worktree_reclaim_sweep.rs:403`: a deleted worktree reported as `removal_failed`

```rust
if super::decommission::remove_session_worktree(&path) && !path.exists() {
```

`remove_session_worktree` returns `false` on a path where `git worktree remove --force` failed but the `remove_dir_all` fallback SUCCEEDED (`decommission.rs:212` and `:224` — both comment "git remove failed, but dir removal succeeded" and return `false`). The `&&` then routes a worktree that is *gone* into `removal_failed`, and its bytes are dropped from `removed_bytes`.

Newly reachable — until this round the loop never executed. `reclaim_remove_mode_reclaims_a_clean_merged_worktree` covers only the happy path where git succeeds.

**Fix** — judge on the observable postcondition, which is the only thing that matters:
```rust
let attempted = super::decommission::remove_session_worktree(&path);
if !path.exists() {
    if !attempted {
        tracing::warn!(path = %path.display(),
            "worktree-reclaim: git worktree remove failed; the directory was \
             removed by fallback and its git registry entry may remain (#2919)");
    }
    // …removed…
} else { out.removal_failed.push(path); }
```

### MEDIUM — `worktree_reclaim_tests.rs:614-635`: not hermetic on a fork

```rust
if !index.is_complete() && index.branch_count() == 0 {
    eprintln!("skipping: `gh` unavailable or unauthenticated in this environment");
    return;
}
assert!(index.branch_count() > 0, …);
```

The skip guard keys on the *shape* of `unavailable()` (`complete: false`, empty), not on whether the call succeeded. A **successful** `gh pr list` returning zero rows yields `complete = true` (`0 < 400`) and `branch_count = 0` — which slips past the guard and trips the assert. That is exactly a fork with no pull requests, or any clone of a zero-PR repo.

It is safe in this repo's CI (unauthenticated → `unavailable()` → skips; authenticated → 400 rows → passes), so it will not turn CI red here. It will fail for a contributor on a fork.

**Fix** — make success observable rather than inferred:
```rust
if std::process::Command::new("gh").args(["auth","status"]).status()
       .map(|s| !s.success()).unwrap_or(true) {
    eprintln!("skipping: `gh` unauthenticated"); return;
}
```
then assert unconditionally.

### LOW — `worktree_reclaim.rs:248-253`: the doc still describes the bug

```
/// What: `gh pr list --state all --limit <PR_INDEX_LIMIT> --json
/// number,headRefName,state`, run with `-C <registry_root>` and with the
/// repository-redirecting environment stripped. …
/// Test: `gh_command_strips_repository_redirecting_env` (argv/env shape);
```

`gh_command`'s own doc was rewritten correctly, but `from_gh`'s was not. It re-plants the exact `-C` misinformation, and it still credits `gh_command_strips_repository_redirecting_env` with "argv" coverage — the specific false claim round 2 disproved. The `--json` list is also stale (missing `isCrossRepository`). `scripts/check_test_pointers.sh` cannot catch this; the pointer resolves, the prose is wrong.

### LOW — `doctor_worktree_disk.rs:254-258`: the timeout message names the wrong number

```rust
format!("worktree disk survey exceeded {}s — disk state undetermined (#2919)",
        SURVEY_TIMEOUT.as_secs()),   // → "exceeded 3s"
```

The outer bound became `SURVEY_TIMEOUT + SURVEY_TIMEOUT + SURVEY_TIMEOUT_GRACE` = 8s at `:238`. The operator is told 3s for a probe that ran 8. Format the real ceiling.

### LOW — `doctor_worktree_disk.rs:130-140`: `pr_state_unknown` has two causes, one explanation

A candidate skipped by the classify deadline is pushed with `pr: BranchPrState::Unknown` (`worktree_reclaim_sweep.rs:116-127`), so it increments `pr_state_unknown` alongside genuine `gh` failures. The rendered `window` clause attributes all of them to the 400-PR window. On this store the doctor's 3s classify budget will expire well before 22 registry roots are indexed, so most of that bucket is deadline pressure, not window truncation — the same "advertise a remedy that cannot fire" shape the PR set out to remove, one level down.

---

## Notes

- **Store-layout drift, observed but not filed.** Several anchors resolve to a repository that does not match their directory: `duettoresearch/gb-connect-card` → `duettoresearch/duetto-eve-agents`, `bob-duetto/hotstats-product-poc` → `hotstats/hotstats-product-poc`, and its mirror. I chased this as a possible wrong-repo delete authorization and concluded it is not one: `count_unpushed` (`worktree_safety.rs:552-575`) measures against the same `@{upstream}`, so if the branch reads merged in the repo the checkout actually tracks AND is fully pushed to it, the work really did land there and reclaiming is right. The directory name is cosmetic. Below the 80% bar; recording it because it will look alarming to the next reader.
- `e2e_survey_against_a_real_store` skips cleanly with the env var absent — confirmed green in a normal `cargo test` run. Its assertion (`pr_state_unknown < candidates.len()`) is exactly the one the `-C` bug would have failed, since that bug produced `pr_state_unknown == candidates.len()`. It is a vacuous pass without the var, which is the stated design.
- `gh_command_passes_no_dash_c_flag` and `gh_command_runs_in_the_requested_directory` both do what they claim (`get_args()` empty, `get_current_dir()` set). The corrected comment on `pr_state_for_branch_maps_a_failed_call_to_unknown` is accurate about what the test does and does not prove.
- `Handle::try_current` (`prune.rs:212`) is correct but unreachable in production — `spawn_blocking` threads always carry a handle. Defensive hardening against a future refactor, not a live fix. Fine as-is.

---

## Round count

**Three rounds is enough. Do not open a fourth.**

The delete-authorizing path is verified on live code, which was the one condition that justified this round. What remains is reporting accuracy and two one-line corrections. HIGH-1 (~15 lines) and MEDIUM-2 (~6 lines) are worth landing in this PR; the three LOWs are doc edits. None of it warrants another full adversarial pass — a spot-check of the diff is sufficient.

**Merge as-is if the alternative is another review round.** The feature is correct where correctness matters: nothing gets deleted that should not, every gate fails closed, and the numbers it reports are floors rather than fabrications. The residual is that one of those floors is not labelled as one.

PR #4526, opened 2026-08-01
fix(trusty-mpm): quarantine untracked project-tier files that shadow bundled agents (#4448)

### LINKED ISSUE #4448: deployer: quarantine untracked project-tier files that shadow bundled agent stems

bobmatnyc:
Closes #4448

Step 2 of the three-issue stack. Built on [#4442](https://github.com/bobmatnyc/trusty-tools/issues/4442) (merged as `555331fd`); [#4443](https://github.com/bobmatnyc/trusty-tools/issues/4443) is step 3.

## The hole

A claude-mpm-era `qa.md` in a workspace `.claude/agents/` outranks the canonical user tier in Claude Code's agent resolution. Every delegation to `qa` silently loads the stale definition, and no check fails.

Nothing could reach it. `retract_framework_agents` sees only ledger entries, and since [#4437](https://github.com/bobmatnyc/trusty-tools/issues/4437) nothing deploys into a workspace tier either, so the deployer's skip-and-warn branch never runs there. Observed in 4 of 16 live sessions on 2026-07-30; 22 files were swept by hand on 2026-07-31 because no code path existed.

## The fix

New `trusty_agents_common::agents::quarantine`. Renames offenders to `<file>.md.disabled` — a rename, never a delete — and drops a `TRUSTY-MPM-QUARANTINE.txt` receipt beside them carrying the `mv` that undoes each one.

Wired at both live retraction call sites (`session_launch/mod.rs`, `session_launch/sync_assets.rs`), immediately **after** retraction. Retraction owns the ledger-tracked tier; this owns the untracked remainder it structurally cannot reach. Run in the other order the two contend over the same files for no benefit.

The fix location in the issue body was stale (that branch is dead post-#4437) — this follows the correction comment.

## What it refuses to move

| Case | Outcome | Why |
|---|---|---|
| `ShadowsBundled` + `Untracked` | **MOVED** | The only movable combination — the issue itself |
| `UserOwned` on a bundled name | Refused | The ledger positively proves the file is the operator's |
| `Custom` (untracked, non-bundled) | Refused | A project's own agent; never reaches the filter |
| `ShadowsBundled` + `FrameworkOwned` | Refused | Ledger-tracked → retraction's to delete; renaming desyncs the ledger |
| `StrandedFrameworkOwned` | Refused — **out of scope by decision** | See below |
| Empty roster | Whole run refused | An empty roster means it could not be *built*, never "nothing is bundled" |
| Corrupt ledger | Whole run refused | See below |
| Destination name occupied | That file refused | `fs::rename` clobbers on Unix; a second sweep would destroy the first's file |
| Destination unstattable | Treated as occupied | Only a proven `NotFound` counts as free |

Two structural properties back this up: the ledger is **never written** (only untracked files move, so no entry can name a renamed file), and no second "is this bundled" predicate exists here — classification is #4442's `tier_audit` and only its.

### `StrandedFrameworkOwned` — excluded, deliberately

It is ledger-tracked and framework-owned, which is exactly what `retract_framework_agents` **deletes** on the same directory at the same call sites, moments earlier. Quarantining it would duplicate that repair with a weaker one and desync the ledger, which would keep naming a file that no longer exists under that name. One owner per case.

### The corrupt-ledger refusal is the load-bearing one

`audit_agent_tier` degrades a corrupt ledger to empty. That is correct for a **report** and catastrophic for a **move**: every `UserOwned` file would read as `Untracked`, and a name collision would then condemn the operator's own agent. New `audit_agent_tier_with_manifest` is the seam that lets the quarantine load strictly and refuse where the probe degrades — without either side forking the scan loop.

## Doc correction (required by the issue)

`tier_audit.rs:42-45` claimed `retract_framework_agents` "never touches an untracked or user-owned file; this module matches that". That parity holds for **user-owned only**. `classify_tier_resident("qa", Untracked, {"qa"})` returns `ShadowsBundled` — pinned by the module's own `classify_bundled_name_shadows` test. The behaviour is right (catching untracked shadows IS the feature) but the sentence read as a guarantee that untracked files were safe from the sweep. Corrected in both `tier_audit.rs` and `changelog.d/4442-tier-audit.md`.

## Discoverability

Not a log line. The existing skip-and-warn fired on eight separate days through 2026-07-30 with zero follow-up. Instead: the `.md.disabled` file sits in the directory the operator is already looking at, the `TRUSTY-MPM-QUARANTINE.txt` receipt beside it explains why and gives the `mv` to undo it (appended across runs, never truncated), and `SyncAssetsReport` gained `agents_quarantined` / `quarantine_error` so the CLI and daemon route surface the sweep.

## Mutation evidence

Every guard was removed one at a time and the suite re-run. Two mutations initially SURVIVED and were fixed by adding the missing tests, not by weakening the guard.

| # | Mutation | Result |
|---|---|---|
| M1 | Drop the empty-roster guard | FAIL `quarantine_refuses_on_empty_roster` |
| M2 | Corrupt ledger degrades to `default()` instead of refusing | FAIL `quarantine_refuses_on_corrupt_manifest`, `quarantine_refuses_the_whole_run_not_just_the_unreadable_file` |
| M3 | Drop the `Untracked` half of `is_movable` | FAIL `movable_truth_table_is_exhaustive`, `quarantine_never_moves_a_tracked_framework_file_on_a_bundled_name` |
| M4 | Drop the `ShadowsBundled` half of `is_movable` | **SURVIVED first** → extracted the inline filter into a named `is_movable` with an exhaustive 9-row truth table. Now FAILs `movable_truth_table_is_exhaustive` |
| M5 | Rename blindly instead of probing for a free name | FAIL `quarantine_falls_back_to_a_numbered_name_on_collision`, `quarantine_refuses_when_the_destination_is_occupied` |
| M6 | Receipt truncates instead of appending | FAIL `receipt_appends_across_runs` |
| M7 | `is_free` uses `.is_err()` (fail-open on any error) | **SURVIVED first** → added `an_unstattable_path_is_not_free`, which reaches through a regular file for a deterministic ENOTDIR on every platform and user. Now FAILs |
| M8 | `is_free` uses `Path::exists()` (follows symlinks) | FAIL `a_dangling_symlink_is_not_free`, `an_unstattable_path_is_not_free` |
| M9 | Run without the ledger lock | **SURVIVED first** → added `quarantine_blocks_while_the_ledger_lock_is_held`. Now FAILs |

M4 is worth naming: `StrandedFrameworkOwned` + `Untracked` is unreachable through today's classifier, so no filesystem test could distinguish "we check the class" from "we only check ownership". The truth table covers the combinations the classifier cannot currently emit, so a future `TierResidentClass` variant cannot silently become movable.

## Verification

```
cargo test -p trusty-agents-common   → 394 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm             → 4129 passed; 0 failed; 6 ignored (+43 further suites, all ok)
cargo clippy -p trusty-agents-common --all-targets -- -D warnings  → exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings            → exit 0
cargo fmt --check                    → exit 0
bash scripts/check_line_cap.sh       → 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh → both crates present and valid
```

No `#[ignore]`, no cfg-gated tests, no `--lib` narrowing. The one `#[cfg(unix)]` test is the dangling-symlink case; its sibling ENOTDIR test is unconditional, so the `is_free` guard stays covered on Linux.


bobmatnyc:
## Verdict: BLOCK

Round 1, independent adversarial review at head `908a009b`. All 24 checks green; the merge with current `origin/main` (`174e1b34`, only #4516 landed since) applies cleanly. Green is the starting point.

The engineering here is unusually careful — the refusal set, the strict-vs-degrading ledger seam, the `symlink_metadata` choice, and the append-only receipt are all correct, and I re-ran and confirmed the three mutations that survived the author's first pass. The block is not about the module's internals. It is about the one thing that changed when #4515's predicate went from REPORT to MUTATION: the predicate cannot tell a stale claude-mpm leftover from a project's deliberate override, and the call site that fires on every session launch has no test at all.

## Findings

|---|---|---|---|---|
| CRITICAL | `crates/trusty-agents-common/src/agents/quarantine.rs` | 329-331 | `is_movable` moves any untracked file on a bundled name. That is indistinguishable from a project's deliberate, committed, spec-sanctioned override — and the receipt's restore does not survive the next launch. | Exempt git-tracked files, or require a positive staleness signal, or ship an opt-out that makes a restore durable. |
| HIGH | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 630-650 | The primary move-authorizing call site has ZERO test coverage. Deleting it entirely leaves the whole `trusty-mpm` suite green. | Add a `prepare_session_*` test that stages a shadowing file and asserts it is quarantined, plus the `project_dir`-not-`fw.claude_agents_dir()` guard retraction already has. |
| MEDIUM | `crates/trusty-agents-common/src/agents/quarantine.rs` | 408 | The receipt's undo uses `{:?}` (Rust `Debug`), not shell quoting. `$` expands and backticks command-substitute. | Emit single-quoted paths with `'` escaped as `'\''`, and add a test that EXECUTES the printed command. |
| MEDIUM | `crates/trusty-agents-common/src/agents/quarantine.rs` | 392-418 | No operator escape hatch: no env opt-out, no marker file, no documented way to make a restore stick. The receipt tells the operator how to undo but not how to keep it undone. | Document (in the receipt itself) the one durable exemption that exists — a `Origin::User` ledger entry — or add a skip mechanism. |

---

### CRITICAL — the sweep cannot distinguish a stale leftover from a deliberate project override, and it renames git-tracked files

`is_movable` (`quarantine.rs:329-331`):

```rust
class == TierResidentClass::ShadowsBundled && ownership == TierOwnership::Untracked
```

`ShadowsBundled` + `Untracked` is the PR's stated "only movable combination". But that combination is not a property of stale files — it is the property of **every project-tier agent that deliberately overrides a bundled name**, because `Origin::Project` does not exist yet. From `docs/specs/system-project-agents-skills.md:191`:

> A single flat global catalog cannot express "this project brings its own reviewer agent that must win over the bundled one."

and §1.3 of the same spec: "there is no `Origin::Project`… the *shape* for representing project content is present but **unused**." So today the designed override and the claude-mpm leftover are byte-for-byte the same classification, and this PR moves both.

**This is not hypothetical — this repository ships eight such files, git-tracked:**

```
crates/trusty-git-analytics/.claude/agents/{code-analyzer,mpm-agent-manager,mpm-skills-manager,web-ui-engineer}.md
crates/trusty-search/.claude/agents/{code-analyzer,mpm-agent-manager,mpm-skills-manager,web-ui-engineer}.md
```

Each declares a frontmatter `name:` that matches a bundled agent exactly (`name: code-analyzer` vs `crates/trusty-mpm/src/assets/agents/code-analyzer.md` → `name: code-analyzer`). `.gitignore:27-28` explicitly re-includes `!.claude/agents/**` — committing project-tier agents is the configured intent, not an accident. Renaming one shows up in `git status` as a deletion plus an untracked `.md.disabled`, in a working tree the operator did not touch.

**And the advertised recovery does not hold.** The receipt says "To restore any of them, run the matching command". I ran it:

```
run1 moved: ["qa.md.disabled"]
# operator runs the mv exactly as printed
run2 moved: ["qa.md.disabled"]     # re-quarantined on the next launch
```

Every subsequent launch re-sweeps, into `.disabled.1`, `.disabled.2`, … up to `MAX_COLLISION_ATTEMPTS = 100`. There is no supported way to keep a project-tier override, and the receipt does not say so. An operator who follows the printed instruction gets a pile of numbered files and an agent that keeps disappearing.

The module's own contract (`tier_audit.rs:104-105`) says `Custom` — "a project's own agent" — "must never be reported or moved". A deliberate override IS a project's own agent; it just does not classify as `Custom`. This is that contract, broken.

Weighted against the brief's own instruction — moving something it shouldn't is far worse than refusing something it could move — this is the failure mode to block on.

### HIGH — the primary move-authorizing call site is unverified

`prepare_session_inner` (`mod.rs:630-650`) is the path that fires on **every** session launch. It has no test.

**Mutation M14 (mine):** delete the entire call site and the import.

```
cargo test -p trusty-mpm
  lib:  4129 passed; 0 failed; 6 ignored
  bin:  1363 passed; 0 failed; 1 ignored  (x2 binaries)
  + every integration suite: ok
```

Fully green with the feature's main delivery path removed. Only the `sync_assets` site is covered (`sync_assets_quarantines_a_shadowing_untracked_agent`).

**Mutation M15 (mine):** move the quarantine block ABOVE `retract_framework_agents`. `cargo test -p trusty-mpm --lib core::session_launch` → 204 passed, 0 failed. The ordering the PR body calls deliberate is not pinned either.

What that absence costs, concretely: retraction carries an explicit regression test — `prepare_session_never_retracts_the_operator_home_agents_tier` (`tests.rs:1757`) — because aiming it at `fw.claude_agents_dir()` once deleted the roster out of the operator's real `~/.claude`. The quarantine is aimed at the same `project_dir.join(".claude").join("agents")` and is **strictly more dangerous there**: retraction on `~/.claude/agents` is a no-op (nothing is ledger-tracked), while the quarantine moves *untracked* files, which is exactly what that directory is full of. That one-character change — `project_dir` → `fw.claude_agents_dir()` — fails no test in this PR.

### MEDIUM — the receipt's undo is not shell-safe

`quarantine.rs:408`:

```rust
"  # shadowed bundled agent: {}\n  mv {:?} {:?}\n", moved.name, moved.to, moved.from
```

`{:?}` on a `PathBuf` is Rust `Debug` escaping, not shell quoting. It emits double-quoted paths, inside which the shell still expands `$` and executes backticks. Observed:

```
  mv "/tmp/.tmp6LMRQs/qa$X.md.disabled" "/tmp/.tmp6LMRQs/qa$X.md"
```

`$X` expands to empty, so the `mv` targets `/tmp/.tmp6LMRQs/qa.md.disabled`, which does not exist — the undo fails with ENOENT. A backtick in the filename is worse: `mv "/dir/qa` + `` `id` `` + `.md.disabled"` runs `id` and substitutes its output. Since identity comes from the frontmatter `name:`, the *filename* is unconstrained, so a cloned repo can put arbitrary characters there. Spaces, `;`, `'`, `"`, `\` and non-ASCII do round-trip correctly; `$` and backtick do not.

`quarantine_writes_a_recovery_receipt` only asserts the receipt *contains* `"mv "` and the filenames. No test ever executes the command, which is why this got through.

## Required Changes

1. Do not move a file the project's own VCS tracks, or otherwise gate on a positive staleness signal, until `Origin::Project` (#4443) can mark a deliberate override. Renaming git-tracked project content is out of proportion to the leftover this issue is about.
2. Make the restore durable, or say plainly in the receipt that it is not and how to make it stick. A printed undo that is reversed on the next launch is worse than no undo.
3. Test `prepare_session_inner`'s call site: one test that a shadowing file is quarantined through `prepare_session`, and one that pins the target as `project_dir`'s tier and not `fw.claude_agents_dir()` — the mirror of `prepare_session_never_retracts_the_operator_home_agents_tier`. Both should fail if the call site is deleted.
4. Shell-quote the receipt's `mv` (single quotes, `'` → `'\''`) and add a test that runs the emitted command and asserts the file is back.

## Mutations run (10 — 8 killed, 2 survived)

| # | Mutation | Result |
|---|---|---|
| M4 (re-verified) | drop the `ShadowsBundled` half of `is_movable` | KILLED — `movable_truth_table_is_exhaustive` |
| M7 (re-verified) | `is_free` → `.is_err()` (fail-open) | KILLED — `an_unstattable_path_is_not_free` |
| M9 (re-verified) | run without the ledger lock | KILLED — `quarantine_blocks_while_the_ledger_lock_is_held` |
| M10 (new) | delete the receipt write entirely | KILLED — `quarantine_writes_a_recovery_receipt`, `receipt_appends_across_runs` |
| M11 (new) | `symlink_metadata` → `metadata` | KILLED — `a_dangling_symlink_is_not_free` |
| M12 (new) | drop the missing-dir early return | KILLED — `quarantine_missing_dir_is_a_noop` |
| M13 (new) | collision refusal → blind overwrite | KILLED — `quarantine_refuses_when_the_destination_is_occupied`, `..._falls_back_to_a_numbered_name...` |
| M14 (new) | **delete the `prepare_session_inner` call site** | **SURVIVED** — full `trusty-mpm` suite green |
| M15 (new) | **quarantine BEFORE retraction** | **SURVIVED** — 204/204 green |
| Probe | restore-then-relaunch round trip | Restore does not stick (evidence above) |

## Notes

Things I checked and did NOT flag:

- **The `StrandedFrameworkOwned` exclusion is sound.** Both call sites (`mod.rs:601`, `sync_assets.rs:170`) target the identical `project_dir.join(".claude").join("agents")`, with retraction first. On the one path where retraction fails and quarantine still runs, quarantine excludes stranded files anyway — the hole persists but nothing is mis-moved. Both refuse on a corrupt ledger, so they agree there too.
- **The roster move is behaviour-preserving.** `bundled_roster` moved verbatim from `doctor_asset_tier.rs` into `core/bundled_roster.rs`; the body is byte-identical, only the visibility changed. The doctor probe's roster is unchanged.
- **The ledger lock is held across the whole read-classify-rename sequence**, not just the read: `quarantine_locked` is one expression inside `with_agent_manifest_lock`, and the manifest is loaded inside it. Two concurrent sessions serialise; the second sees `.md.disabled` already present and finds nothing.
- **`.md.disabled` is genuinely inert.** `is_agent_file` is `ends_with(".md")` → false. `claude_mpm_loader.rs:281,359` filter on `path.extension() == Some("md")` → `"disabled"` → skipped. Both the sync and async loaders.
- **`is_movable`'s truth table is not compiler-enforced** — it is a hand-listed 9-row array, so a new `TierResidentClass` or `TierOwnership` variant would not fail it. But `is_movable`'s `==`-conjunction form means any new variant defaults to *unmovable*, which is the fail-closed direction. Fine as written; the table's value is documentation, not enforcement.
- **The approval path IS tested** (`quarantine_renames_an_untracked_shadowing_file`), so the refusal tests are not passing against an unconditionally-refusing function.
- **The `tier_audit.rs:42-56` doc correction is accurate.** The old sentence claimed parity with `retract_framework_agents` for untracked files; the replacement correctly narrows it to user-owned and states the untracked asymmetry is the feature. The changelog fragment matches.

The `.md.disabled` mechanism, the refusal set, and the lock discipline are right. Fix the classifier's blind spot for deliberate project overrides and test the launch call site, and I expect this to approve on round 2.

bobmatnyc:
## Parked as draft — blocked on [#4443](https://github.com/bobmatnyc/trusty-tools/issues/4443)

Accepting the BLOCK. The CRITICAL is a defect in the issue's premise, not in the implementation, so this is parked rather than patched. No heuristic was attempted.

### CRITICAL independently reproduced

`ShadowsBundled` + `Untracked` is not the signature of a stale claude-mpm leftover. It is the signature of **every deliberate project-tier override**, because `Origin::Project` does not exist yet. The two states are indistinguishable by construction today.

Verified in this repo, not taken on faith:

```
$ git ls-files '*/.claude/agents/*.md' | wc -l
34
$ # of those, resolving to a name in crates/trusty-mpm/src/assets/agents/:
8
```

| File | Declared `name:` |
|---|---|
| `crates/trusty-git-analytics/.claude/agents/code-analyzer.md` | `code-analyzer` |
| `crates/trusty-git-analytics/.claude/agents/mpm-agent-manager.md` | `mpm-agent-manager` |
| `crates/trusty-git-analytics/.claude/agents/mpm-skills-manager.md` | `mpm-skills-manager` |
| `crates/trusty-git-analytics/.claude/agents/web-ui-engineer.md` | `web-ui-engineer` |
| `crates/trusty-search/.claude/agents/code-analyzer.md` | `code-analyzer` |
| `crates/trusty-search/.claude/agents/mpm-agent-manager.md` | `mpm-agent-manager` |
| `crates/trusty-search/.claude/agents/mpm-skills-manager.md` | `mpm-skills-manager` |
| `crates/trusty-search/.claude/agents/web-ui-engineer.md` | `web-ui-engineer` |

`.gitignore:27-28` re-includes `!.claude/agents/**` deliberately. These are committed on purpose. This PR would quarantine them on the next session launch, and re-quarantine after every restore.

No heuristic closes this. "Is it git-tracked" is not available to the deployer and is wrong for a project not using git. The fix is #4443's `Origin::Project`, which lets a project **declare** an override — only then does `Untracked` genuinely mean unclaimed.

**Stack reorder: #4442 (landed) → #4443 → #4448.**

### Landed anyway — the parts that survive the reorder

**HIGH (M14) — fixed.** `prepare_session_inner` is the primary move-authorizing path and had zero coverage; deleting it left 4129 lib + 1363 bin + all integration green. Three tests now cover it, and M14 re-run fails two of them:

- `prepare_session_quarantines_a_shadowing_workspace_agent` — the approval path
- `prepare_session_never_quarantines_the_operator_home_agents_tier` — the guard you asked for, staging an **untracked** file on a bundled name, since the retraction test's manifest-tracked fixture is invisible to the quarantine by construction
- `prepare_session_retraction_and_quarantine_split_the_tiers_cleanly` — combined end state

**MEDIUM (receipt injection) — fixed.** `quarantine.rs` now POSIX single-quotes every path via `shell_quote`. `receipt_undo_command_survives_a_hostile_filename` stages a file named `it's evil$(touch PWNED)` + backticks, **executes the printed `mv` through `sh`**, and asserts the file is restored and no payload ran. Reverting to `{:?}` fails it.

### M15 — reporting a disagreement rather than faking the test

You asked me to pin quarantine-after-retraction. I measured it instead, and the ordering is **not** load-bearing:

- Swapping the call sites → 207/207 green.
- Removing the `Untracked` narrowing → 207/207 green too.

They converge because the narrowing already refuses every file retraction will delete, so neither operation can reach the other's tier whichever runs first. My original test comment claimed removing the narrowing would fail the ordering test; I ran it, that claim was **false**, and I corrected the comment rather than shipping it.

So the test pins the combined end state and states plainly that it is not a sequence assertion. The narrowing stays pinned where it is observable — `quarantine_never_moves_a_tracked_framework_file_on_a_bundled_name` and `movable_truth_table_is_exhaustive`, both of which fail when it is dropped. The module doc calling the ordering load-bearing is corrected. A test asserting a sequence that provably does not change behaviour would look like coverage without being any.

Happy to add a hard ordering pin if you still want one, but I did not want to invent a test that passes for the wrong reason.

Skipped the operator-escape-hatch MEDIUM per your note — moot once #4443 provides the declaration mechanism.

### Gate

```
cargo test -p trusty-agents-common   → 397 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm             → 4132 passed; 0 failed; 6 ignored (+43 suites ok)
cargo clippy -p trusty-agents-common --all-targets -- -D warnings  → exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings            → exit 0
cargo fmt --check                    → exit 0
bash scripts/check_line_cap.sh       → 0 violations
bash scripts/check_changelog_fragment.sh → both crates valid
```

Not merging. Resume after #4443 lands, at which point the predicate should key on the absence of a project declaration rather than on `Untracked`.

bobmatnyc:
## Adversarial review — PR #4526 (independent, worktree `.claude/worktrees/qa-4526`, HEAD `7d4c305f` vs `origin/main` `9e025f0d`)

## Verdict: BLOCK

One CRITICAL finding in the destructive path itself: on a mid-batch I/O failure, already-renamed files are silently dropped from the receipt, contradicting this module's own documented safety contract. Everything else in this PR — the classifier, the collision-safe rename, the shell-quoted undo, the ledger-lock, the refusal-first design — is genuinely well built and well tested. This is a narrow, mechanical fix, not a design problem.

## Premise check (post-#4286) — HOLDS

#4286 (`7d4a8599`, merged to main as part of #4665) retired the `.trusty-mpm/` **instruction-override** files — a completely different subsystem (CLAUDE.md prose assembly). It did not touch agent-tier deployment/resolution. Current `origin/main`'s `crates/trusty-agents-common/src/agents/tier_audit.rs` module doc (post `origin/main`'s own #4527 merge) still states the model this PR depends on verbatim: since #4409/#4437 bundled agents deploy to exactly one directory (`$CLAUDE_CONFIG_DIR/agents/`), so any other `.claude/agents/` is a leftover that Claude Code's own resolution still ranks above the canonical user tier (the #4408 shape). That model is untouched by #4286. **The premise still holds; nothing here is stale.**

## Mergeability — DOES NOT merge cleanly, and is 57 commits behind

`git merge-base HEAD origin/main` = `555331fd` (57 commits behind, 2 ahead). `git merge --no-commit --no-ff origin/main` conflicts in exactly one file:

- `crates/trusty-agents-common/changelog.d/4442-tier-audit.md` — **CONFLICT (content)**. Cause: PR #4527 (already merged to main as `63385102`) independently shipped the *same* doc correction to `tier_audit.rs` this PR also makes, with slightly different wording in the changelog fragment. The `.rs` file itself (`tier_audit.rs`) auto-merges clean — only the two fragments' prose differs.
- All other touched files (`session_launch/mod.rs`, `session_launch/tests.rs`, `core/mod.rs`, `tier_audit.rs`) auto-merge without conflict markers despite each having independently changed on main since the merge-base.

**Action required before merge:** drop or reconcile this PR's `4442-tier-audit.md` fragment edit against the version #4527 already shipped (they say the same thing).

## Findings

| CRITICAL | `crates/trusty-agents-common/src/agents/quarantine.rs` | 94–96 (doc), 282, 290–291 (code) | The module's own doc says: *"[`Io`] is a genuine filesystem failure, which may occur after some files have already been renamed (those are reported in the receipt regardless)."* The code does not honor this. In `quarantine_locked`, `std::fs::rename(&found.path, &to)?` (line 282) and `write_receipt(dir, &result)?` (line 291) both use `?` inside/after the `for` loop. If file A renames successfully and file B's rename then fails (EACCES, ENOSPC, EBUSY, etc.), the function returns `Err` **before** reaching the post-loop `write_receipt` call — A's rename is real and permanent on disk, but is never written to any receipt and never will be (a subsequent scan can't find `A.md` anymore to re-report it, since it's already `A.md.disabled`). The same information-loss happens if the receipt write itself fails after a fully successful rename loop: the `?` on `write_receipt` discards `result` on the way out. The caller (`session_launch/mod.rs`'s `Err(err) => tracing::warn!("shadowing-agent quarantine did not run: {err}")`) then logs a message that is **false** in this exact scenario — the sweep did partially run, and moved a file with no trace. This is precisely the failure mode the module exists to prevent ("a rename an operator cannot explain or undo is worse than the shadow it fixed", line ~85–91) and there is no test that exercises a genuine `rename`/`write_receipt` I/O failure mid-batch to catch it — `quarantine_refuses_on_corrupt_manifest`/`_the_whole_run_not_just_the_unreadable_file` cover the ledger-corrupt refusal (nothing moves), not a failure *after* a successful move. | Track already-attempted renames in `result` as they happen (already done), then on a `rename` failure, best-effort `write_receipt` the partial `result` *before* returning `Err` (`if result.changed() { let _ = write_receipt(dir, &result); } return Err(io_err.into());`), and on a `write_receipt` failure after a fully successful loop, `tracing::error!` the full `result` contents (paths + destinations) so the record survives somewhere durable even though the receipt file itself couldn't be written. | Fix here |
| MEDIUM | `crates/trusty-agents-common/changelog.d/4448-quarantine-shadowing.md` (whole file) | 1 | The fragment embeds three category headers in one file ("Added", then bare-line "Security", then bare-line "Documentation"). `scripts/assemble-changelog.sh` only reads line 1 as the category (`parse_fragment`/`fragment_body`) — everything after it, including the literal words "Security" and "Documentation", is copied verbatim as body text under the first category. Verified by running `bash scripts/assemble-changelog.sh trusty-agents-common --stdout`: the security fix (shell-injection-safe quoting) and the doc-correction note both render as unstyled paragraph text nested inside `### Added`, not under their own `### Security` / `### Documentation` headings. `check_changelog_fragment.sh` still passes (it only requires a known category on line 1 plus one `- ` bullet), so this does not fail CI, but it does mislabel a security-relevant fix in the rendered `CHANGELOG.md`. | Split into three fragment files, one category each: `4448-quarantine-shadowing.md` (Added only), `4448-quarantine-shadowing-security.md` (Security), `4448-quarantine-shadowing-docs.md` (Documentation). | Fix here |
| MEDIUM | `crates/trusty-mpm/changelog.d/4448-quarantine-shadowing.md` (whole file) | 1 | Same defect: "Fixed" then a bare "Changed" line in one fragment. Verified via `bash scripts/assemble-changelog.sh trusty-mpm --stdout` — the bundled-roster-consolidation note renders folded into `### Fixed` instead of its own `### Changed` section. | Split into `4448-quarantine-shadowing.md` (Fixed) and `4448-quarantine-shadowing-changed.md` (Changed). | Fix here |

## Data safety (aside from the CRITICAL above) — well designed

- **Rename, never delete.** Confirmed: the only filesystem-mutating call in the movable path is `std::fs::rename` (quarantine.rs:282); nothing calls `remove_file`/`remove_dir` on a quarantined file.
- **Collision-safe destination.** `free_quarantine_path` probes `<file>.md.disabled`, then `.1`…`.99` via `symlink_metadata` (not `Path::exists`, which follows symlinks and would misreport a dangling symlink as free) and only a proven `NotFound` counts as free; any other stat error (including permission-denied) is treated as occupied — the conservative direction. Tested: `quarantine_falls_back_to_a_numbered_name_on_collision`, `quarantine_refuses_when_the_destination_is_occupied`, `an_unstattable_path_is_not_free`, `a_dangling_symlink_is_not_free`.
- **Cannot clobber a prior quarantine.** A second sweep against an already-quarantined file cannot destroy it — `free_quarantine_path` returning `None` (all 100 candidate names occupied) causes a refusal (`result.refused`), never a `rename` call. `quarantine_is_idempotent` confirms a second run against an already-swept directory is a no-op.
- **Reversible and the user is told how**, *on the success path*: `write_receipt` appends (never truncates) a timestamped stanza with the exact `mv <to> <from>` per file, every path POSIX-single-quoted via `shell_quote` — tested against a hostile filename (`it's evil$(touch PWNED)`) in `receipt_undo_command_survives_a_hostile_filename`, confirming no code execution on paste. This is good, deliberate work. The CRITICAL finding above is specifically about the *failure* path, where this guarantee silently stops applying.
- **Ledger integrity.** The ledger is never written by this module (`quarantine_leaves_the_ledger_untouched`), and the exclusive `with_agent_manifest_lock` is held for the full scan+rename+receipt critical section, serializing against a concurrent `retract_framework_agents` on the same directory (`quarantine_blocks_while_the_ledger_lock_is_held` — a real blocking test, not a stub).
- **Corrupt-ledger / empty-roster refusals are whole-run, not partial.** `quarantine_refuses_the_whole_run_not_just_the_unreadable_file` stages three files under a truncated manifest and confirms all three survive untouched.

## Blast radius of the matcher

`is_movable` (quarantine.rs:336–338) requires **both** `TierResidentClass::ShadowsBundled` **and** `TierOwnership::Untracked` — exhaustively enumerated and pinned in `movable_truth_table_is_exhaustive` across all 9 class×ownership combinations (2 of the 9 are currently unreachable through the classifier but are pinned anyway against a future classifier variant). Practically:

- A file the ledger positively records as the operator's (`UserOwned`) is never touched — proven at two independent layers (the audit itself never emits it as movable, and the filter would reject it even if that regressed).
- A project's genuinely custom agent (any name not in the bundled roster) never reaches this function — the audit's own classification excludes it as `Custom` before the filter runs.
- **The one real edge case:** a hand-authored `.claude/agents/<file>.md` whose frontmatter `name:` happens to collide with a bundled agent's name, and that was never deployed by tm (no ledger entry), *will* be quarantined with no prompt. This is consistent with this project's own current architecture decision (project-tier `.claude/agents/` is not a supported override surface post-#4409/#4437/#4408 — a deliberate project-scoped customization is meant to come via #4443's not-yet-landed project-custom-agent path, which is expected to use a *distinct* name, not a bundled one) — so I am not flagging it as a defect, only noting it for the record.
- The sweep is automatic (wired into every `prepare_session_inner` session launch and every `sync_session_assets` call), non-interactive, and has no dry-run flag. There is no code path today that lets an operator preview the sweep before it runs. Given every ambiguous state refuses (see above) and the CRITICAL finding is the only place where "automatic" turns into "unexplained," I consider this an accepted, reasoned design (the PR's own "Discoverability" section explicitly weighs and rejects log-only reporting) rather than a finding — flagging only for visibility since it is genuinely irreversible-feeling from the operator's chair until they find the receipt.

## Verification I ran myself (worktree `.claude/worktrees/qa-4526`, PR HEAD `7d4c305f`, not yet rebased)

```
cargo test -p trusty-agents-common --quiet   → 397 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm --quiet             → 0 failed across every suite (7131 passed, 16 ignored, summed across all binaries)
cargo clippy -p trusty-agents-common --all-targets -- -D warnings   → clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings             → clean
cargo fmt --check                             → clean
bash scripts/check_line_cap.sh                → 7 allowlisted, 0 violations
bash scripts/check_changelog_fragment.sh --base 555331fd → OK (both crates) — does not catch the category-mixing above (see MEDIUM findings)
```

Rust Test Ladder: this is a **rung 5** change (destructive filesystem mutation + process-lifecycle wiring at every session launch). The PR's own verification block covers rung 4 fully (targeted + crate-wide tests, clippy, fmt) and includes real failure-path and concurrency tests (corrupt ledger, lock contention, collision) satisfying rung 5's "failure-path and concurrency tests" bar for every guard *except* the one this review found missing: a genuine I/O failure *after* a successful rename. No `--include-ignored` run was needed — nothing in the new code is `#[ignore]`-gated. The adversarial review this comment represents is rung 5's own required gate.

1. Fix the CRITICAL: preserve partial `QuarantineResult` (write the receipt, or at minimum `tracing::error!` its contents) before propagating an `Io` error from either the `rename` call or the `write_receipt` call in `quarantine_locked`. Add a test that forces a real `rename`/`write_receipt` failure after at least one successful move and asserts the already-moved file is still recorded (receipt or log) — this is exactly the shape of the mutation-testing rigor already used elsewhere in this PR.
2. Split both changelog fragments into one file per category.
3. Resolve the merge conflict against current `origin/main` in `crates/trusty-agents-common/changelog.d/4442-tier-audit.md` (duplicate of already-merged #4527), and rebase — the branch is 57 commits behind.

bobmatnyc:
**Blocker closed — this PR now has no unblocking path.** #4443 was closed as superseded on 2026-08-03 under the "collapse the agent tiers" ruling. Since `Origin::Project` was going to land there, the movability predicate in this PR stays wrong with nothing queued to fix it.

Two known defects remain on this branch independently of that: it would quarantine trusty-tools' own committed agent files (`is_movable` misclassifies them), and it has an unfixed CRITICAL partial-failure receipt gap. The 8-file counter-example that originally demonstrated the first one is gone from the repo as of #4696, but the predicate is still wrong in the general case.

Recommend closing and folding the still-valid parts into #4448 rather than attempting another review round — this branch has already drawn 3+ rounds. Awaiting an owner decision; leaving the PR open and held in the meantime.

bobmatnyc:
Closed on owner ruling 2026-08-03 — folding into #4448.

This branch has been held rather than merged. Closing rather than attempting another review round:

- **Its blocker is gone with no replacement.** #4443 was closed as superseded under the "collapse the agent tiers" ruling. `Origin::Project` — which this PR's movability predicate needs to be correct — was going to land there and now has no home.
- **Two defects remain unfixed on the branch.** `is_movable` misclassifies trusty-tools' own committed agent files, so this would quarantine them; and the CRITICAL partial-failure receipt gap is still open. The 8-file counter-example that originally demonstrated the first is gone from the repo as of #4696, but the predicate is still wrong in the general case.
- **It has drawn 3+ review rounds.** Repo convention treats that as evidence to close and fold rather than iterate again.

The still-valid intent — quarantining untracked project-tier files that shadow bundled agent stems — continues in #4448. Nothing is lost; the branch remains in git history if any of the implementation is worth salvaging.

Issue #4545, opened 2026-08-01
tm-agent-manager must refuse to create an agent whose name collides with a bundled agent


bobmatnyc:
## The gap

\`tm-agent-manager\` builds agents on the user's behalf. Nothing currently stops it from writing a project agent whose frontmatter \`name:\` matches a bundled agent. Claude Code resolves on \`name:\` and ranks the project location above the framework's, so such an agent silently shadows the bundled one — reproducing #4408, in which a 32-byte stub whose body was the literal string \`v1\` beat the real 25KB \`rust-engineer\`, and delegation kept "working" against a content-free agent.

## Owner's direction (2026-08-01), verbatim intent

A user who insists on hand-building a colliding agent owns the consequences; but \`tm-agent-manager\`, which builds for them, must never produce that collision itself.

## Proposed resolution

Before writing any new agent, resolve the proposed \`name:\` against the canonical bundled roster — \`bundled_agent_names()\` already exists in \`crates/trusty-agents-common/src/agents/tier_audit.rs\` (~line 196, landed in #4442 as commit 555331fd). 

On collision: REFUSE. Do not auto-suffix — that produces a second confusing name with no signal about why. Offer the nearest unused name, or require an explicit \`--override-bundled\` flag for the deliberate case.

## The failure this must make impossible

\`tm-agent-manager\` itself ever writing an agent whose identity matches a bundled agent without an explicit declaration of intent.

## References

- #4408 — the incident (32-byte project stub shadowed 25KB rust-engineer)
- #4442 — landed \`bundled_agent_names()\` helper
- #4443 — custom deployment mechanism (constraint mentions collision guard)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

Issue #4528, opened 2026-08-01
A scaffolding agent should select which agents and skills a project deploys


bobmatnyc:
**The gap.** trusty-mpm is a FRAMEWORK. Every session currently receives the full bundled skill and agent roster regardless of the project's stack. Measured on this Rust workspace (see #4513): 33 of 103 listed skills are `vercel:*` — `vercel:nextjs`, `vercel:ai-sdk`, `vercel:turbopack`, `vercel:shadcn` and so on — in a repo with no Vercel deployment. They can never fire here. The full skill listing costs ~23KB / ~5.8k tokens per session, and the agent roster another ~9.8KB / ~2.5k.

**Why the obvious fix is wrong.** Disabling the Vercel plugin globally would optimize the framework for one repo. Another project legitimately uses Vercel, or TypeScript, or Rails. The roster is correct as a *catalog*; what is missing is per-project *selection*.

**Proposed resolution.** A scaffolding agent (possibly a new dedicated agent rather than an extension of `mpm-skills-manager`) configures a project at init/refresh time with the agents and skills matching its detected stack, and records that selection so subsequent sessions load only the selected set. Open questions to record in the issue, not answer:
- New dedicated agent, or extend the existing `mpm-skills-manager` / `tm-init` skill?
- Where the selection is recorded (project `.trusty-mpm/` config, the skills manifest, or elsewhere).
- Whether selection is a hard filter or a default the operator can widen.
- Re-scaffolding behaviour when a project's stack changes.

Reference #4513 as the measurement that surfaced it.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closed as not planned — unowned feature backlog, 1.7.2 prune (owner decision 2026-09-24). Evidence: merged PR refs: #4572 #4697 (partial, not a fix); last real activity 2026-09-04 (18d), 0 comments. Reopen if someone takes it on.

PR #4572, opened 2026-08-01
docs: ADR-0025 collapse agent/skill tier hierarchies (manifest.toml-driven)

bobmatnyc:
## Summary

Adds ADR-0025: collapses the three unreconciled agent/skill deploy destinations (`~/.claude/skills`, `$CLAUDE_CONFIG_DIR/*`, per-worktree `.claude/*`) into a declared-source, precedence-resolved-at-deploy-time model. Records the owner's final ruling: per-project classification lives in the existing `manifest.toml` (no second config format), a project selects agents from the catalog but never authors its own agent file, `tm-agent-manager` creates the manifest once on request, and startup reconciliation extends the existing `retract_framework_agents` fail-closed rather than adding a new sweep.

References #387, already closed on 2026-08-01 as superseded by this ADR.

Incident this ADR is fixing the shape of: #4408 (a 32-byte stub beat the real 25KB `rust-engineer`; `tm doctor` counted files, not content).

Related open work this ADR does not land: #4443, #4448 (migration copy-then-quarantine mechanics), #4528 (manifest-creation vehicle, absorbing #4545's bundled-name collision refusal). None of these should be affected by merging this doc.

## What's in the ADR

- **Context** — the three write paths, the #4409 agent/skill scoping asymmetry, the #4408 incident, claude-mpm prior art, the known #2002/#2444 staleness gap.
- **Decision** — 16 clauses: manifest.toml-driven per-project classification (extends the existing format, does not add a new one); manifest created once on request by `tm-agent-manager`; declared numeric-priority sources with hard-error ties and four-surface override visibility; startup reconciliation extending `retract_framework_agents` (flagged as the ADR's highest-risk clause — the only one that deletes files on every session launch, fails closed on a bad manifest); stack detection once at manifest creation with drift reported (never auto-corrected) by `tm doctor`; loud-failure requirement for delegation to an unresolvable agent; report-first, copy-then-quarantine migration; `~/.claude/skills` permanently untouched (never written, read, swept, or reported).
- **Consequences** — both directions, including the new failure classes this design introduces (silent-absence-by-omission is now the manifest's normal operating mode; drift is possible and only `tm doctor` surfaces it; whether unresolvable-agent delegation fails loudly today is unverified).
- **Slices** — the two load-bearing ordering constraints (loud-failure check first; startup reconciliation gated on upstream slices green in production, not merely merged). The full 21-slice breakdown lives in the implementation plan, not inlined here.
- **Related Decisions** — vetted against ADR-0002, 0008, 0015, 0020, 0023, 0024, and DOC-31 (partially superseded — §191 must be rewritten in the change set that lands clause 1). Notes the pre-existing ADR-0021 numbering collision on main as a separate cleanup.

## Test plan

- [x] `bash scripts/check_sld.sh` — 0 errors, 0 warnings
- [x] `bash scripts/check_test_pointers.sh` — 0 dangling pointers
- [x] `bash scripts/check_doc_numbers.sh` — 0 violations
- [x] `bash scripts/check_changelog_fragment.sh` — docs-only change, gate confirms no crate source changed, exempt
- [x] `docs/adr/INDEX.md` updated with the new entry
- [x] `gh pr view --json closingIssuesReferences` confirmed empty


bobmatnyc:
Converted ADR-0025's bare `file.rs:NNN` code citations into GitHub blob permalinks, per the repo owner's 2026-08-01 source-linking convention.

**Pinned to:** `fd15c589b6c92d03ebcfb8a37463611300a1b1b4` (origin/main at the time of this edit) — the ADR describes code as it exists on main today, not code this PR changes, so main's tip is the correct pin.

**Verification:** every citation was opened at that SHA and checked against the ADR's claim before linking. Five were stale (line numbers had drifted from earlier drafts) and were corrected to the actual location:

- `daemon/managed_routes/lifecycle.rs:963,1087` → `974,1289` (the two `for_managed_workspace` call sites — `spawn_managed_inproject` and `spawn_managed_local` — that don't route through `WorkspaceProvisioner::provision_in`)
- `lifecycle.rs:959-963` → `970-974` (the actual `#1931` rationale comment + call, a few lines further down)
- `provisioner/workspace.rs:894` → `905` (the actual `for_managed_workspace` call; comment starts at 899)
- `launch_on_main.rs:162` → `138`
- `sync_assets.rs:145` → `149` (`crates/trusty-mpm/src/daemon/managed_routes/sync_assets.rs`, disambiguated from two same-named siblings by context)
- `core/paths.rs:108-135` → split into `113-137` (`FrameworkPaths::default()`) and `433-435` (`claude_skills_dir()`), since the original single range captured neither — those two functions sit ~300 lines apart and the old range spanned mostly unrelated code in between

Everything else (`global_config.rs:78,122-125`; `session_launch/mod.rs:562,589-598,599,688`; `core/session_assets.rs:1-30`; `bin/tm/commands/install.rs:162-172`) checked out correct as-is and was only reformatted into linked form.

No citation was left unresolved. Prose, structure, and clause numbering are unchanged — this is a citation-format change only.

Issue #383, opened 2026-05-29
fix(trusty-mpm): tm install overwrites assembled system prompt with 4-line stub


bobmatnyc:
## Summary
`tm install` writes the minimal stub to `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` via InstallPolicy::Overwrite. Only the next `tm launch` repairs it. TUI launch never regenerates the file, leaving a stale stub on disk.

## Current Behavior
- `tm install` writes 4-line stub (file ref: src/bin/tm.rs around install handler)
- On-disk file is incorrect immediately after install
- Next `tm launch` eventually regenerates correct version
- TUI launch path does not trigger regeneration
- User sees stale content if inspecting `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` directly

## Desired Behavior / Acceptance Criteria
- `tm install` calls `install_system_prompt()` (or equivalent)
- Correct, fully-assembled system prompt is written to disk during install
- On-disk file is immediately valid and matches what Claude receives
- Both CLI and TUI launch paths use correct version

## Spec Reference
- architecture-spec §5 (Instruction Assembly)
- architecture-spec §12 (Gaps & Deviations)

Issue #4573, opened 2026-08-01
Prohibitions and Circuit Breakers tables are tier 'project' — deletable, unguarded at merge


bobmatnyc:
## Defect A — mis-tiering contradicts spec

Sections `core.md:11` (Prohibitions) and `:253` (Circuit Breakers) are declared `customization_tier: project` in `pm-instruction-package.json:14-18`. SPEC-PMINSTR-01 §3 "Mis-tiering #1" identified this and §6 prescribed `Prohibitions → customization_tier: fixed`. The implementation shipped `project`.

A three-line CLAUDE.md block in any project deletes both tables (24 KB) and validates cleanly:
```
<!-- TRUSTY-MPM: CORE START v=1 -->
ship it
<!-- TRUSTY-MPM: CORE END -->
```

The second override path (`.trusty-mpm/PM_INSTRUCTIONS_DEPLOYED.md`, `core/instruction_overrides.rs:286-294`) achieves the same wholesale deletion.

## Defect B — the floor's statement of the authority model is a broken pointer

Non-overridable floor at `sections/non-overridable-rules.md:3`:
> All prohibitions defined in PM_INSTRUCTIONS.md SS Prohibitions are BINDING.

`PM_INSTRUCTIONS.md` does not exist — #4183 deleted it. `scripts/check_instruction_floor.sh:75-79` forbids its sibling `BASE_PM.md` from returning. `SS` is mojibake for `§`. The floor binds the PM to a table in a file not in the build.

## Defect C — the floor guard passes when every rule is inverted

Mutation testing of `scripts/check_instruction_floor.sh`:

| Mutation | Guard result |
|---|---|
| baseline | PASS |
| floor reduced to 3 bare needle strings, every rule deleted | PASS |
| every rule INVERTED ("Always emit `Co-Authored-By: Claude`") | **PASS** |
| entire floor wrapped in HTML comment | PASS |
| `non-overridable-rules.md` emptied to 0 bytes | PASS |
| file removed entirely | FAIL |

The guard's substance is three `grep -qF` needles (`:91-93`) — file existence + three literals. It cannot distinguish a rule from its inversion.

Separately, `gh api .../branches/main/protection` returns `["Format check","500-line file-size cap","Clippy","MSRV check (1.91)","Test","trusty-search daemon smoke test"]` — **`instruction-floor-guard` is not a required status check.** Deleting `.github/workflows/instruction-floor-guard.yml` turns nothing red.

## Proposed resolution

[...]

PR #4592, opened 2026-08-01
fix(trusty-mpm): floor docs point at CLAUDE.md named sections, dedupe attribution footer

bobmatnyc:
## Summary

Two coupled fixes to shipped PM instruction assets under `crates/trusty-mpm/src/assets/instructions/sections/`.

**Override channel (owner ruling).** The floor's "Customizing PM Behavior" table advertised only the legacy `.trusty-mpm/*.md` override files and never mentioned the `CLAUDE.md` named-section channel that outranks them (`HOST_FILES` in `core/claude_md_sections.rs` scans `CLAUDE.md` before `.trusty-mpm/INSTRUCTIONS.md`). `sections/non-overridable-rules.md` is rewritten so project customization is named sections in the project's root `CLAUDE.md`, full stop — the marker grammar, all eight recognized tokens, and the three `fixed`-tier tokens that can never be overridden. The `.trusty-mpm/` override files are mentioned in one line: still read by the current binary, #4286 removes them, never create one.

**Two more floor defects fixed in the same file:**
- Broken pointer to the deleted `PM_INSTRUCTIONS.md` (removed by #4183) — repointed at the CORE section's Prohibitions table.
- `SS` mojibake for `§`.

**`core.md` fixes:**
- Model Selection Protocol said "sonnet or haiku", omitting opus — the mandatory choice for coding tasks per its own routing table two lines below.
- PM Allowlist listed `push` as an allowed git op, contradicting P7 (push is Version Control's job).

**Attribution footer dedup.** The commit/PR attribution footer was stated verbatim three times in the resolved prompt (`core.md`, `workflow.md`, `framework-guaranteed-conventions.md`). The two non-canonical copies now point at the one canonical statement in `framework-guaranteed-conventions.md`, byte-identical and untouched. Footer text unchanged.

**Test fallout.** Golden fixtures (`pm-prompt-*.md`) regenerated via `UPDATE_GOLDEN=1 cargo test -p trusty-mpm golden`. One test (`a_marked_block_in_instructions_md_is_delivered_once`) asserted no composed prompt could ever contain the literal string `TRUSTY-MPM:` — no longer true now that the floor documents its own marker syntax with a `<TOKEN>` placeholder example. Narrowed the assertion to check for this override's own concrete `WORKFLOW START`/`END` marker lines instead of a blanket substring check, preserving the original coverage (no unparsed marker leaking into the prompt) without asserting something the change deliberately makes false.

## Investigation note (reported, not acted on)

`crates/trusty-mpm/src/assets/instructions/INSTRUCTIONS.md` (the `FRAMEWORK_INSTRUCTIONS` stub, "Daemon endpoint: ...") is dead in practice. `bundle_all.rs` writes it to `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` during `tm install`, but `install.rs` immediately overwrites that same path with the fully-assembled PM prompt in the same `tm install` run (comment: "Overwrite the bundle stub with the fully assembled PM prompt (#383)"), and `tm launch` unconditionally regenerates it again at every launch (`launch.rs` step 10, `install_system_prompt()`). No production code path ever reads the stub's own content back — its only consumer is `bundle_tests.rs` asserting it's non-empty. Not touched in this PR per instruction.

## Test plan

- [x] `bash scripts/check_instruction_floor.sh` — PASS
- [x] `bash scripts/check_sld.sh` — 0 errors/warnings
- [x] `bash scripts/check_line_cap.sh` — 0 violations
- [x] `cargo test -p trusty-mpm --lib -- core::instruction_package core::bundled_pm_package_tests core::claude_md_sections core::instruction_overrides core::instruction_pipeline core::pm_prompt_golden_tests` — 155 passed, 0 failed
- [x] `cargo test -p trusty-mpm --lib` (full crate) — 4199 passed, 0 failed, 6 ignored


bobmatnyc:
VERDICT: APPROVE — reviewed at head 42b4b9aa; three findings raised, all fixed in bd5e124a.

Two concerns were investigated specifically because they are the shapes that hide regressions:

1. A test assertion was narrowed to accommodate a content change. Finding: defensible, not lost coverage. The deleted blanket check (`!prompt.contains("TRUSTY-MPM:")`) did not protect "delivered once" — a separate untouched count assertion does that. It protected a different invariant: that marker lines are consumed rather than passed through to the model. The narrowed assertion still fails if that invariant breaks, and the blanket form survives at the unit level on `strip_marker_blocks` itself. No regression exists that the old assertion catches and the new one misses. The narrowing was forced: the floor now legitimately ships the literal marker syntax as documentation, so the blanket form is a false positive.

2. Golden fixtures were regenerated with UPDATE_GOLDEN=1, which issue #4268 records as silently deleting instruction prose in this repo. Finding: no unexplained deletion. Verified by enumerating the multiset difference of non-blank lines rather than by inspection — 37 / 30 / 37 removed lines across the three fixtures, every one mapping to an authorized change. Line counts increased in all three (904→915, 653→664, 830→841), the opposite of the #4268 signature. The attribution footer appears exactly once per composed golden, not zero times, and retains the full rule including the prohibition on Claude Code attribution.

FINDINGS, all fixed in bd5e124a:
- MEDIUM: \`git branch\` remained in the PM allowlist at \`sections/core.md:41\` while P7 routes branch/push/rebase/merge/tag to Version Control — the same contradiction the PR fixed for \`push\`, left half-done on the line under review. Fixed: allowlist is now \`git status/add/commit/log/diff/pull/stash\`; \`rebase\`/\`merge\`/\`tag\` confirmed absent.
- LOW: \`sections/non-overridable-rules.md\` described marker rejections as "silently ignored"; the code emits a \`tracing::warn!\` per rejection. Reworded to "declined and logged as a warning."
- LOW: the rewrite dropped the inspect command without replacing it for the new channel. Added \`grep -n 'TRUSTY-MPM:' CLAUDE.md\`.

NOTED, OUT OF SCOPE: \`sections/non-overridable-rules.md:3\` binds the PM to the CORE section's Prohibitions table while CORE is tier \`project\` — a project shipping a CORE block can delete P1-P11 and leave a fixed-tier floor rule pointing at nothing. Pre-existing architecture, tracked by #4573, and strictly better than the dangling PM_INSTRUCTIONS.md pointer it replaced.

MERGE GATE: the \`Test\` check validates the goldens and must be green before merge.

Issue #4614, opened 2026-08-01
CI `Test` job cannot fail — missing `set -o pipefail` on a `cargo test | tee` pipeline means every test failure reports green


bobmatnyc:
## Defect

`.github/workflows/ci.yml`, job `test` (`name: Test`), step "cargo test --workspace" (origin/main lines 635-647):

```bash
if cargo test --workspace ... 2>&1 | tee test-output.log; then
  test_status=0
else
  test_status=${PIPESTATUS[0]}
fi
if [ "$test_status" -ne 0 ]; then ...; exit "$test_status"; fi
```

No `shell: bash` is declared anywhere in the file (verified: zero `shell:` keys), so the step runs under GitHub's implicit default, `bash -e {0}` — no `-o pipefail`. The `if` therefore evaluates `tee`'s exit code, which is always 0. `${PIPESTATUS[0]}` is dead code. The `Test` job cannot fail on a test failure, on any commit, ever. It is a required status check.

## Evidence (observed directly, not inferred)

- Run `30702707438`, job `91376388876` (main push, head `11ef5c27c831`, post-PR #4593 merge): job `conclusion: success`.
- That job's raw log:
  - `shell: /usr/bin/bash -e {0}` immediately preceding the tee'd step.
  - `test llm::client::tests::missing_openrouter_key_errors_at_chat_time ... FAILED`
  - `thread 'llm::client::tests::missing_openrouter_key_errors_at_chat_time' panicked at crates/trusty-code/src/llm/client.rs:604:9: expected MissingConfig, got: MissingCredential { provider: OpenRouter }`
  - `test result: FAILED. 1595 passed; 1 failed; 4 ignored; ...`
  - Execution continued straight into the next step (`cargo test -p trusty-review`) instead of exiting — proof the `if` took the success branch.
- Reported independently: `origin/main` is currently red on `trusty-mpm`'s `golden_bundled_fallback_prompt` (first-bad commit `42c7774e` / PR #4593), also merged and reported green by this same gate.

At least two distinct live test failures are sitting behind a green required `Test` check right now.

## Resolution

Add `set -o pipefail` to the step (or read `cargo test`'s exit code without piping through `tee`; dropping the pipe-to-tee pattern is preferable — a failure path that depends on a shell option is fragile by construction).

Verification must be by mutation, not a green run: push a deliberately failing test and confirm `Test` goes red. A gate nobody has watched fail is not a gate — that's exactly how this survived.

## Sequencing

[...]

PR #4616, opened 2026-08-01
fix(trusty-code): actionable MissingConfig for an absent OPENROUTER_API_KEY, and make the missing-key tests hermetic

bobmatnyc:
## Why

`crates/trusty-code/src/llm/client.rs` carries **two distinct defects**. They were invisible together: defect 2 is the one CI sees, and defect 1 is why nobody saw it locally — and #4614's broken `Test` gate is why CI seeing it didn't matter.

Prerequisite for #4614 (step 2 of its sequencing). This must land before `set -o pipefail` does, or pipefail turns `main` red on merge.

## Defect 1 — the test false-passes

`missing_openrouter_key_errors_at_chat_time` opened with an early return taken whenever `OPENROUTER_API_KEY` was non-empty. On any machine with a real key it reported `ok` **without executing a single assertion**. The same guard sits on its three sibling tests, so all four are fixed here.

The guard existed for a sound reason — with a real key present, `chat()` would build the adapter and issue a **live API call**. So this does not delete the guard, it removes its cause: each test now clears its own key for the test body (`#[serial]`, exact prior state restored on drop). The assertions become unconditional, and a live call becomes unreachable by construction rather than by skipping. `serial_test` joins the workspace-wide exclusion group already used by every env-mutating test.

## Defect 2 — which side was wrong

Hermetically the test failed at `client.rs:604`: `expected MissingConfig, got: MissingCredential { provider: OpenRouter }`.

**The code regressed. The test was right.** Evidence:

- **Git history.** `git log -S'MissingConfig' -- crates/trusty-code/src/llm/client.rs` shows the OpenRouter path has asserted `MissingConfig` since `e831a00b` (#1043). The migration `d8b216e8` (#2492) kept it via `provider_for(...).map_err(convert::map_error)?`, and `convert::map_error` explicitly mapped `MissingCredential { provider }` → `MissingConfig(...)` with a resolver hint — under its own test, doc'd "preserves the actionable 'set this key' operator guidance". `5cd6be33` (#4436) deleted `convert.rs` wholesale. The mapping went with it; the doc comments and the test did not.
- **Two live doc contracts still state it.** `build_adapter`: "a missing key surfaces as the resolver's own `MissingCredential`, **mapped to `MissingConfig`**". `dispatch.rs::chat`: "surfaces a missing credential as `InferenceError::MissingConfig` at this point." Neither was touched by #4436.
- **The contract, on its own merits.** Three sibling arms of the *same* `match` — Fireworks, Together, AtlasCloud — deliberately construct `MissingConfig` naming their env var and the three ways to set it. OpenRouter is the `_` default arm, i.e. the first credential a new operator must set, and was the only one emitting the bare `MissingCredential`. What a caller can distinguish is the point: `MissingCredential` can only carry a `ProviderId` — it is *structurally* incapable of the actionable message. `error.rs` says so in as many words: `MissingConfig` exists because `MissingCredential` "cannot carry an operator-actionable message". Operator-facing, this is `no credential resolved for provider openrouter` vs `OPENROUTER_API_KEY is required to reach an openrouter model. Export it (…), add it to .env.local, or set it via 'tcode config keys set openrouter'.`
- **Nothing else breaks.** Both variants are `is_alarm() == true` and `is_retryable() == false`, so classification is unchanged. `trusty-agents` asserts `MissingCredential{OpenRouter}` for *its* client, which has no actionable-message convention — a different consumer with a different contract, not a conflict.

Fixed as an explicit guard mirroring the three siblings rather than by re-adding a mapping layer: `convert.rs` was deleted for good reasons, and the guard states the contract at the site that owns it.

## Verification

Run on a machine where `OPENROUTER_API_KEY` **is** set (73 chars) — i.e. the exact condition that used to hide all of this.

```
cargo test -p trusty-code --lib
test result: ok. 1596 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 3.09s
```
(baseline on `origin/main`: `1595 passed; 1 failed`)

Full crate incl. integration targets: every binary `ok`, `tests/inference_shared_adapter_e2e.rs` 4 passed.

**Hermeticity** — the four tests, three environments:

| env condition | result |
|---|---|
| `env -u OPENROUTER_API_KEY` | `4 passed; 0 failed` |
| `OPENROUTER_API_KEY=<dummy>` | `4 passed; 0 failed` |
| ambient real key inherited | `4 passed; 0 failed` |

**Mutation** — remove the new guard, restoring the regressed code:

```
thread '…missing_openrouter_key_errors_at_chat_time' panicked at crates/trusty-code/src/llm/client.rs:663:22:
expected MissingConfig naming OPENROUTER_API_KEY, got: MissingCredential { provider: OpenRouter }
```

It fails identically under **all three** env conditions — including the ambient-real-key case, where the old test returned `ok` having asserted nothing. One mutation, both defects proven. Guard restored, suite green again.

Gates: `cargo clippy -p trusty-code --all-targets -- -D warnings` clean · `cargo fmt --check` clean · `check_line_cap.sh` 0 violations · `check_test_pointers.sh` 0 dangling pointers · changelog fragment at `crates/trusty-code/changelog.d/4614-openrouter-missing-key-actionable.md`.

No `#[ignore]`, no cfg-gating, no narrowing — the OpenRouter test's assertion was *strengthened* (it now also pins the message names the env var, matching its siblings).

Refs #4614

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11367,6 +11367,7 @@ dependencies = [
  "reqwest 0.12.28",
  "serde",
  "serde_json",
+ "serial_test",
  "sha2 0.10.9",
  "tempfile",
  "thiserror 1.0.69",
--- a/crates/trusty-code/Cargo.toml
+++ b/crates/trusty-code/Cargo.toml
@@ -153,3 +153,8 @@ tower = { workspace = true }
 # a real bound socket so `reqwest` (SSE included) drives it exactly like a
 # real `tcode serve --http` daemon, without spawning the real binary.
 wiremock = { workspace = true }
+# `llm::client::tests` (#4614): the missing-credential tests clear a provider's
+# API-key env var so they assert unconditionally instead of skipping on a
+# developer machine that has one. `#[serial]` joins the workspace-wide
+# exclusion group so that mutation cannot race a parallel test thread.
+serial_test = { workspace = true }
--- a/crates/trusty-code/changelog.d/4614-openrouter-missing-key-actionable.md
+++ b/crates/trusty-code/changelog.d/4614-openrouter-missing-key-actionable.md
@@ -1,0 +1,21 @@
+Fixed
+
+- **A missing `OPENROUTER_API_KEY` again reports an actionable error (issue
+  #4614).** OpenRouter is tcode's default route, so it is the first credential
+  a new operator has to set — yet since #4436 deleted the `convert::map_error`
+  bridge, an absent key surfaced as the shared resolver's bare
+  `MissingCredential { provider: OpenRouter }` ("no credential resolved for
+  provider openrouter"), which names nothing the operator can act on.
+  `build_adapter` now guards the OpenRouter route explicitly, exactly as it
+  already did for Fireworks, Together, and AtlasCloud, and returns
+  `MissingConfig` naming the env var and the three ways to set it. This is the
+  contract both `llm::client::build_adapter` and `llm::dispatch::chat` have
+  documented throughout; only the code had drifted.
+- **The missing-credential tests no longer pass without asserting anything
+  (issue #4614).** All four `missing_*_key_errors_*` tests returned early
+  whenever the corresponding API key was present in the ambient environment, so
+  on any developer machine holding a real key they reported `ok` having
+  executed zero assertions — which is why the regression above went unseen
+  locally and was visible only in CI. Each test now clears its own key for the
+  duration of the test body (`#[serial]`, restored on drop) so the assertions
+  run unconditionally, and no live API call is reachable.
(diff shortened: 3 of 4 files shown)


bobmatnyc:
⚠️ **This PR's green `Test` check is not evidence of anything.** It runs the pre-#4614 workflow, whose `Test` job cannot fail (#4617). It reported green on this exact failure for as long as the failure existed.

The real evidence:

- **Local**, on a machine where `OPENROUTER_API_KEY` IS set — the condition that used to hide this: `cargo test -p trusty-code --lib` → `test result: ok. 1596 passed; 0 failed; 4 ignored` (baseline on `origin/main`: `1595 passed; 1 failed`).
- **CI with a working gate**: #4617 merges this branch and fixes the gate. Its green run [30718937016](https://github.com/bobmatnyc/trusty-tools/actions/runs/30718937016) (job `91419320377`, `Test` pass 15m28s) shows `test llm::client::tests::missing_openrouter_key_errors_at_chat_time ... ok` with zero `test result: FAILED` anywhere in the log — under a `Test` job that was proven capable of going red on the immediately preceding commit.

Merge order: this PR first, then #4617.

PR #4619, opened 2026-08-02
perf(trusty-mpm): stop re-reading assets for stopped sessions in tm ls (#4322)

### LINKED ISSUE #4322: perf(trusty-mpm): tm ls re-reads ~47 MB of deployed assets serially per invocation — 56% is for stopped sessions

bobmatnyc:
## What this changes

Bundled agents deploy into **one machine-global directory**: #4409 deliberately
exempted `FrameworkPaths::agent_deploy_dir()` from `for_managed_project`'s
per-workspace rewrite, so `for_managed_workspace(ws).agent_deploy_dir()` returns
the identical path for every session. `tm ls` was reading that one directory's
42 files **once per session** — on the 32-session fleet the issue measured, 1344
opens and 30.2 MiB of the 41.9 MiB the listing moved, producing 32 byte-identical
answers.

This hoists the deployed-agent read out of the per-session fan-out and shares it,
mirroring exactly what #2444's review already did for the catalog half: compute
the shared part once, fan out only what genuinely differs per session. Skills are
untouched — they still deploy per workspace (`<ws>/.claude/skills`), so two
sessions genuinely can disagree and there is nothing to share.

## Measurement

**Method.** A committed `#[ignore]`d benchmark
(`daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many`)
builds a synthetic fleet matching the real shape — 42 catalog agents at 23,541 B
deployed, 52 skills at 7,401 B, 32 workspaces, sizes taken from the live install
— and times the **old and new fan-outs alternately in the same process**, 7 reps
per run, asserting after every pair that both produced the identical verdict map.

Two checkouts timed minutes apart would have been worthless here: this laptop
runs other agents' builds, and during development the *identical* baseline
measured **213 ms at load ~2 and 1301 ms at load ~29**. Alternating in-process is
what makes the ratio survive that. Run it with:

```
cargo test -p trusty-mpm --lib bench_stale_assets_for_many -- --ignored --nocapture
```

**Results** — 3 independent runs, 7 reps each (21 paired samples), load 12–20:

| run | per-session median | shared median | speedup |
|-----|-------------------|---------------|---------|
| 1   | 230.408 ms        | 147.065 ms    | 1.57x   |
| 2   | 232.681 ms        | 151.976 ms    | 1.53x   |
| 3   | 236.335 ms        | 150.941 ms    | 1.57x   |

Full spread: per-session **218.5–267.4 ms**, shared **144.7–162.6 ms**. The two
distributions do not overlap in any of the 21 samples.

Load-independent metrics (deterministic, identical on every run):

| metric | before | after | change |
|---|---|---|---|
| deployed-side file reads | 3008 | 1706 | **-43.3%** |
| bytes read | 41.9 MiB | 12.7 MiB | **-69.7%** |

3008 = 32 x (42 agents + 52 skills), matching the issue's "~93 files per session".
1706 = 42 agents once + 32 x 52 skills. Cold `tm ls` latency is governed by files
opened, not CPU — the issue measured a constant 0.128 s daemon CPU against a
5–9 s wall time and proved causality by warming exactly the probe's file set — so
the read count is the honest predictor of the cold win on any machine.

These are my measurements. I am not quoting the issue's projected numbers.

## Which of the four steps

| step | status |
|---|---|
| 1. Stop probing `Stopped` sessions | **already landed in #4326** |
| 2. Parallelize the fan-out | **already landed in #4326** |
| 3. Memoize the per-session work | **implemented, in a sound form** (see below) |
| 4. Precompute in the 60 s background loop | **skipped deliberately** |

**Step 3, reinterpreted.** The issue proposed memoizing "keyed on
`(agents_dir, skills_dir)` mtime". I did not implement that literally, because a
directory's mtime does **not** change when an existing file is rewritten in
place — so a memo keyed on it reports a genuinely drifted asset as fresh. That is
precisely the "faster wrong answer" this must not ship. Instead I removed the
redundant work outright: the same read, deduplicated **within one request**. It
captures the win the memo was after with **no cache-staleness surface at all**.

**Step 4, skipped**, for two reasons. It serves an answer up to 60 s old, so an
operator who runs `tm session sync-assets` would still see `[stale-assets]` on the
next listing — it breaks "the next call must see the change". And it burns the
I/O every 60 s forever whether or not anyone runs `tm ls`, which is worse on an
idle laptop than doing the work on demand.

## Correctness

**The shared read cannot change an answer.** The deployed-agent bytes are shared
*input*. The only thing that varies per session — the plan's `agent_selected` /
`agent_staleness_ignored` predicates — is pure string matching over the stem and
is applied *after* the read, inside `CatalogHashes::detect`. The snapshot is keyed
on the **catalog's** stems, not a selected subset, so it is predicate-independent
by construction and one snapshot correctly serves sessions selecting different
agent sets. The codebase already documents the underlying property: #4409 made
agents machine-global, so "an agent can no longer be stale for session A and fresh
for session B" (existing comment in
`checked_summaries_stale_assets_independent_per_session_sharing_one_catalog`).

It is also *more* internally consistent than before: previously 32 sessions read
the directory at 32 different instants, so a concurrent write mid-listing could
give session 1 the old bytes and session 30 the new ones. One read per listing
removes that skew.

**No invalidation window.** Sharing is scoped to a single request; nothing is
retained after it returns. `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call`
proves it: it runs the fan-out, mutates the agent tree on disk, and runs again in
the same process with no reset of any kind — the second call must see the change.
A cross-request memo is exactly what would fail that test.

**The `Stopped` skip** (from #4326) is sound because the signal is not deleted,
only moved off the interactive surface: the fleet list marks such rows
`stale_assets_unchecked` so absence of a marker is never rendered as "fresh", and
the single-session fetch `GET …/managed/{id}` — what `tm session resume` reads,
the moment drift becomes actionable — still probes under the full
`staleness_meaningful_for` gate. Both halves are pinned by existing tests I
re-verified green: `checked_summaries_does_not_probe_stopped_sessions` builds a
genuinely drifted *Stopped* session and asserts the list does not flag it *and*
does mark it unchecked, while `record_to_summary_checked_still_flags_stale_stopped_session`
asserts the very same record still reports stale when fetched individually.

## Tests

New:
- `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` — the
  read-count invariant (4 sessions must cost 1 agent + 4 skill reads = 5).
  **Mutation-verified:** reverting the hoist makes it observe 8 and fail.
- `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call` — cache
  invalidation / freshness across calls.
- `stale_assets_for_many_keeps_per_session_verdicts_correct_under_concurrency` —
  6 sessions, alternating drifted/fresh skill trees, all sharing one
  `Arc<DeployedAgentHashes>` and one `Arc<CatalogHashes>` across the concurrent
  fan-out; each must reach its own correct verdict (no interleaving corruption).
- `session_asset_staleness_with_shared_matches_unshared` — shared and unshared
  paths pinned to identical `stale` **and** identical change lists.
- The benchmark itself asserts old and new produce identical verdict maps on
  every repetition, so a "speed-up" that changed an answer fails the benchmark.

## Gates

```
cargo fmt --check                                                  exit 0
cargo clippy --workspace --all-targets -- -D warnings              exit 0, 0 errors
cargo test  (trusty-mpm lib)                4211 passed; 0 failed; 7 ignored
cargo test  (whole workspace)               21931 passed
scripts/check_line_cap.sh                   7 allowlisted, 0 violations — OK
scripts/check_changelog_fragment.sh         OK trusty-mpm: fragment present and valid
```

Clippy/test use CI's own exclusion list (`--exclude trusty-mpm-gui
--exclude trusty-code-gui --exclude trusty-agents-ui`, `SKIP_UI_BUILD=1`) — these
are Tauri desktop apps that need WebKit2GTK/pnpm and are excluded in
`.github/workflows/ci.yml` itself, not weakened here.

Two workspace runs surfaced flaky failures in `trusty-code` (deadline timers) and
`trusty-search` (fsevents watchers) — a *different* set each run, on a machine at
load 41–58 with `fseventsd` pinned at 139% CPU. `git diff origin/main --stat --
. ':(exclude)crates/trusty-mpm'` is **empty**: every file outside `crates/trusty-mpm`
is byte-identical to main, so those crates' behaviour is unchanged by this PR by
construction. `trusty-mpm` itself was 0-failure in both runs.

## Review round 2 (critic verdict WARN — 1 HIGH, 1 MEDIUM, 2 LOW; all addressed)

**HIGH: the read-count pin was itself flaky.** `DEPLOYED_READ_COUNT` was a
process-global atomic while the pin asserted an exact value, relying on
`#[serial_test::serial]` for isolation. None of the 21 tests in
`core/update_check/tests.rs` are `#[serial]` and every one of them reaches
`read_deployed`, so their reads got attributed to the pin. I shipped a flake
into the very test that guards this PR's thesis.

Fixed by scoping the counter to the fixture path: it is now a LOG of read
paths, counted via `deployed_reads_under(<prefix>)` against the caller's own
`fake_home()` temp dir — the same reason `COMPUTE_CALL_LOG` is a path log and
not a counter (#4326 review). The pin now additionally asserts each
per-workspace skill tree was read once, so it also catches a change that
over-collapses the genuinely per-session half.

*Stability proof.* Note that bare `cargo test -- detect` does **not** match
`stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` (0
matches) — that filter alone never runs the pin. The valid reproduction is
`cargo test -p trusty-mpm --lib -- detect reads_shared_agent_dir`, which runs
the pin concurrently with the 46 `detect` tests:

```
run 1..10:  test result: ok. 48 passed; 0 failed   [pin ran & ok: 1]    10/10
```

*The scoping is load-bearing.* Forcing the count back to a process-global total
reproduces the failure deterministically:

```
run 1: FAILED   left: 5   right: 1
run 2: FAILED   left: 5   right: 1
run 3: FAILED   left: 5   right: 1
```

*Still non-vacuous.* Reverting the hoist (moving `DeployedAgentHashes::read`
back into the per-session fan-out) fails it: `left: 4, right: 1`.

**MEDIUM: the read increase is removed, not disclosed.** The critic was right
that eagerly hashing every catalog stem is a regression for a manifest-filtered
single-target caller, which previously read only its selected stems. Since
pre-hashing only pays off when the result is *shared*, `DeployedAgentHashes`
now has two forms: EAGER (`read`) for the fleet path, where one snapshot serves
N sessions; and LAZY (`lazy`) for every single-target caller (`/health`,
`tm catalog apply`, the single-session `GET …/managed/{id}`), which hashes on
demand and therefore performs byte-for-byte the same reads as before this PR.
Pinned by `single_target_detect_reads_only_selected_agents` (1 of 4 catalog
agents selected → exactly 1 deployed read), mutation-verified: switching that
path back to eager makes it observe 4.

**LOW:** dead value in the concurrency fixture removed. **LOW:** the benchmark
now alternates which variant runs first (old-first on even reps, new-first on
odd). Re-measured with alternation: **1.53–1.57x**, reads 3008 → 1706 unchanged.
New-first reps match old-first reps, so the ordering bias was real but not
material.

Re-run gates: `cargo fmt --check` exit 0; clippy exit 0 with 0 errors;
`trusty-mpm` lib **4212 passed, 0 failed**; doc-pointer lint and line-cap OK.
The only workspace failures remain the 6 `trusty-search` fsevents watcher tests,
in a crate whose diff against `origin/main` is empty.

Closes #4322


bobmatnyc:
## Verdict: WARN

Independent adversarial review in a throwaway worktree off `origin/main`. Every claim below was re-run, not read.

### What I verified (and it held)

| Claim | Method | Result |
|---|---|---|
| Shared snapshot is predicate-independent | Wrote a temp probe: ONE `DeployedAgentHashes` + 4 different `(select, ignore_staleness)` pairs over a catalog where `alpha`/`beta` drifted and `gamma` did not | Holds. Verdicts `[alpha,beta]` / `[alpha]` / clean / clean, each equal to an unshared per-predicate read |
| No filter baked into the snapshot | `with_manifest` iterates `catalog.agent_hashes.keys()` with no predicate; predicates apply in `agent_changes` (`update_check/mod.rs:352-353`) | Confirmed by construction and by the probe |
| Freshness across calls | Temp probe: drift the catalog so the on-disk read decides, then reconcile the deployed file **in place**, same process, no reset | Second call flips `true → false`. No invalidation window |
| Read-count pin is mutation-verified | Reverted the hoist (fan-out calls `..._with_catalog`) | **Fails**, observed 9 vs 5. Not vacuous |
| The mtime argument for skipping step 3 | `stat` on APFS: rewrite a file in place, then add a file | In-place rewrite bumped the **file** mtime `…650 → …652`, left the **directory** mtime at `…650`; adding a file bumped it to `…653` |
| Benchmark reproduces | Ran the committed `#[ignore]`d bench, load ~13.5 | `speedup=1.55x`, `reads 3008 -> 1706` — inside the PR's 1.53–1.57x, read counts exact |
| Suite green | `cargo test -p trusty-mpm --lib` ×3 | `4211 passed; 0 failed` each time |

**On step 3 and step 4 — the author is right, and it matters.** A directory-mtime memo would have reported my in-place-reconcile probe as still stale: the file changed, the directory's mtime did not. That is a wrong answer served fast, and rejecting the issue's literal step 3 was the correct call, not a shortcut. Skipping step 4 is likewise sound — a 60 s precompute cannot satisfy `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call`, which the PR pins.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/update_check/mod.rs` | 181-183 | `DEPLOYED_READ_COUNT` is a **process-global** `AtomicUsize`, but the pin asserts an exact `== 5` under only `#[serial_test::serial]`. **Zero** of the 21 tests in `core/update_check/tests.rs` carry `#[serial]`, and every one reaches `read_deployed` through `detect_staleness`. Reproduced a real flake: 1 failure in 10 filtered runs — `left: 25, right: 5` | Scope the counter to the fixture: have `reset_deployed_read_log` store the test's root path and have `read_deployed` increment only when `path.starts_with(root)`. (Annotating all 21 tests `#[serial]` also works but silently rots the moment a new `detect_*` test lands.) |
| MEDIUM | `crates/trusty-mpm/src/core/update_check/mod.rs` | 415-425 | `with_manifest` hashes **every** catalog stem, ignoring `agent_select`/`agent_staleness_ignored`. Pre-PR, `agent_changes` read only stems passing both. Amortized to nothing for the fleet — but `detect_staleness` (`/health`, `tm catalog apply`, single-session `GET …/managed/{id}`) is now a strict read **increase** on any install whose manifest excludes agents. A perf PR argued entirely as read-reduction does not disclose this | The predicate-independence it buys is *required* for sharing to be correct, so keep the behaviour — but say so in the `Why`, and state the bound (≤ catalog size). Alternatively take an optional stem superset for single-target callers. |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/tests.rs` | ~1250-1275 | In `..._keeps_per_session_verdicts_correct_under_concurrency`, the `expected` vec built in the loop is immediately shadowed by a fully recomputed one; the loop's `drifted` value is dead | Push only `r.id` in the loop. |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/staleness_bench_tests.rs` | ~215-232 | Every rep times OLD then NEW in fixed order — no alternation. The ~13 MiB working set is cache-warm after the warm-up so residual bias is small, but the order is unrandomised. Separately, `stale_assets_per_session_agent_read` is a hand-copied fork of the pre-PR fan-out that can drift from `main` silently | Swap the order on odd reps (`if rep % 2 == 0 { old; new } else { new; old }`) and note the fork's maintenance obligation. |

### Required changes

1. Fix the HIGH: make the deployed-read counter fixture-scoped so the pin cannot be polluted by concurrent non-serial tests. It is a genuine intermittent red, reproduced here at 1-in-10.

### Notes

- The PR body says reverting the hoist makes the pin observe **8**; my revert observed **9** (the shared read in `staleness_inputs` still fires). Cosmetic.
- Full-suite runs did **not** surface the flake (0/3) — the collision window is narrow at 4211-test scale. It reproduces readily under a filtered run, which is what a developer iterating locally will do.
- Not flagged: deleting or hand-editing a deployed agent file does not flip `stale_assets`, because `classify_drift` short-circuits when the manifest checksum matches the catalog. That is **pre-existing** and untouched here.
- The correctness core of this PR is sound. The HIGH is test infrastructure, not production behaviour — `DEPLOYED_READ_COUNT` is `#[cfg(test)]`.

PR #4620, opened 2026-08-02
fix(trusty-mpm): resolve the agent roster once so the delegation list and session-start count cannot drift (#4589, #4588)

### LINKED ISSUE #4588: tm session start: printed agent count is a single-tier scan, understates the delivered three-tier roster
**Defect**: `tm session start` prints `Instructions: N agents in delegation authority` from a single-directory scan ([instruction_pipeline.rs:530](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/instruction_pipeline.rs#L530) → `scan_agents(&input.agents_dir)`, fed `fw.agent_deploy_dir()`), while the roster actually delivered into the PM prompt is a three-tier union (project `.claude/agents` + `$CLAUDE_CONFIG_DIR/agents` + framework dir), computed separately by [`deployed_roster_section`](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/delegation_authority.rs#L354) via [`deployed_agent_dirs`](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/delegation_authority.rs#L294) / [`roster_from_dirs`](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/delegation_authority.rs#L321). N understates what the PM actually received.

**Evidence** (live, installed `tm 1.3.1`, tag `dd31adaa`):
- `$CLAUDE_CONFIG_DIR/agents` scan (the printed-N source): **34** agents.
- `tm session instructions` → `## Delegation Authority` section → 39 `### <name>` entries.
- Delta = 5, exactly the agents present only in the `~/.claude/agents` generic tier: `writer`, `writing-critic`, `copyeditor`, `proofreader`, `pangram-editor`.

**Resolution**: compute the printed count from the same three-tier union `deployed_roster_section` uses (or call it directly and count its rendered `### ` entries), so the operator-visible number at [start.rs:143](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/bin/tm/commands/session/start.rs#L143) matches what the PM was told.

### LINKED ISSUE #4589: Three real agents (memory-manager, mpm-agent-manager, mpm-skills-manager) hidden from delegation roster by role: base filter
**Defect**: `memory-manager`, `mpm-agent-manager`, and `mpm-skills-manager` ship with `role: base` in their own frontmatter. [`scan_agents`](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/delegation_authority.rs#L125) excludes any agent whose `role` starts with `base` via [`is_base_role`](https://github.com/bobmatnyc/trusty-tools/blob/8abf30962863e143ed405e8d6cabe33f6b0f0b6d/crates/trusty-mpm/src/core/delegation_authority.rs#L112), treating it as a foundation template. All three are deployed and dispatchable but ABSENT from the `## Delegation Authority` roster the PM receives — the PM is never told they exist.

**Evidence** (live, installed `tm 1.3.1`, tag `dd31adaa`):
- `$CLAUDE_CONFIG_DIR/agents` holds 42 `.md` files; 5 are genuine `BASE-*.md` templates (`BASE-AGENT`, `BASE-ENGINEER`, `BASE-OPS`, `BASE-QA`, `BASE-RESEARCH`). The filter yields 34, not 37 — the 3-agent gap is `memory-manager`, `mpm-agent-manager`, `mpm-skills-manager`, confirmed by reading each file's frontmatter directly (`role: base` on all three, non-`BASE-*` filenames).
- Corroborated independently from a live `tm session instructions` run: all three are absent from the rendered `## Delegation Authority` section (39 entries, none named `memory-manager`/`mpm-agent-manager`/`mpm-skills-manager`) while present in the harness's dispatchable agent-type list. Deployed, dispatchable, invisible to the roster.

**Resolution** — two candidates, owner decides:
- (a) Correct the three agents' frontmatter so `role` is not `base` — one-line asset change per file.
- (b) Tighten the exclusion in `is_base_role`/`scan_agents` to key off the `BASE-*` filename convention rather than a frontmatter `role` value real agents legitimately use — changes classification behavior for every deployment.

bobmatnyc:
Two reports, one defect: trusty-mpm advertised an agent roster that did not match what was actually deployed and reachable. Both are consequences of resolving the roster **twice**.

## #4589 — three real agents hidden by the `role: base` filter

`scan_agents` excluded any agent whose frontmatter `role:` began with `base`, on top of the `BASE-*` file-name rule. `memory-manager`, `mpm-agent-manager` and `mpm-skills-manager` shipped that way — deployed, dispatchable, and absent from every roster the PM ever received.

**Chosen resolution: option (b)** — key the exclusion off the `BASE-*` file-name convention alone.

Why (b), not (a):

- **Option (a) is already in the tree and was not enough.** The three bundled assets already carry correct `role:` values on `main`, and `every_bundled_non_base_agent_reaches_the_roster` already gates that. The rule that caused the defect survived, and it is the rule — not the three files — that #4589 is about.
- **The roster is a three-tier union (#4409 direction).** Bundled agents live at the tm-owned `$CLAUDE_CONFIG_DIR` user tier, but `deployed_agent_dirs` also unions `<project>/.claude/agents` and the operator's own `~/.claude/agents`. tm authors the frontmatter in exactly one of those three. A classification rule that reads a value tm does not control will silently delete an operator's agent from the other two, and **no asset edit can fix it**. `BASE-*` is tm's own convention, applies to the files tm actually ships, and `scan_agents` already checks it before the file is even read.

## #4588 — printed count derived from a single directory

`tm session start` printed `Instructions: N agents in delegation authority` from `scan_agents(fw.agent_deploy_dir())` — one directory — while the roster delivered to the PM was the three-tier union computed separately by `deployed_roster_section`.

## The unifying fix: one resolver

`delegation_authority::resolve_roster(project_dir)` is now the only implementation, and every consumer calls it:

| Consumer | Before | After |
|---|---|---|
| Delegation section in the PM prompt | `deployed_agent_dirs` + `roster_from_dirs` (inline) | `resolve_roster` |
| `tm session start` count | `scan_agents(<one dir>)` — **second implementation** | `resolve_roster` |
| `tm doctor` `agents` check | `.md` file count only | `.md` file count **+** `resolve_roster` size |

`PipelineInput.agents_dir` is replaced by `project_dir`: there is no longer a directory for a call site to get wrong. `tm doctor` reports the roster as `unknown` — never a plausible-looking number — when no project directory scopes the probe.

## Measured, from the real code path

Live machine, `resolve_roster` and `scan_agents` called directly:

```
tier 0 : <project>/.claude/agents                          (0 .md files)
tier 1 : ~/.trusty-tools/trusty-mpm/claude-config/agents   (42 .md files)
tier 2 : ~/.claude/agents                                  (6 .md files)

#4588  session-start count BEFORE (single tier, main rule): 37
#4588  roster AFTER (one resolver, three-tier union)      : 42
#4588  delegation roster the PM already received          : 42
```

Printed 37, delivered 42 — the delta of 5 is the generic `~/.claude/agents` tier, exactly the delta #4588 reports. After the fix both numbers are 42, because both are the same call.

For #4589, the deploy tier was reconstructed in the state the issue reports (`role: base` on the three agents) and scanned under both rules:

```
#4589  BEFORE (main rule): 34 agents
#4589  AFTER  (this PR) : 37 agents
#4589  memory-manager       BEFORE: MISSING  AFTER: present
#4589  mpm-agent-manager    BEFORE: MISSING  AFTER: present
#4589  mpm-skills-manager   BEFORE: MISSING  AFTER: present
```

34-vs-37 is the exact gap #4589 measured. All three agents are present in the live roster after the fix.

**Honest caveat:** because option (a) already landed, no *currently deployed* agent on this machine carries `role: base` — so today's live roster count does not change from the #4589 half alone. The 34→37 figures above come from reconstructing the reported state. What changes permanently is the rule.

## Red-then-green

Both regression tests were written and run **before** any production change:

```
running 2 tests

thread '...::agent_with_a_base_role_but_no_base_filename_stays_in_the_roster' panicked:
assertion `left == right` failed: a non-`BASE-*` file must never be classified as a
foundation template by its `role:` value alone (#4589)
  left: []
 right: ["memory-manager"]

thread '...::session_start_count_matches_the_delivered_delegation_roster' panicked:
assertion `left == right` failed: the count `tm session start` prints must equal the
number of agents the PM was actually given (#4588)
printed: 2
delivered roster:
### engineer / ### qa / ### ticketing / ### writer
  left: 2
 right: 4

test result: FAILED. 0 passed; 2 failed; 0 ignored; 4213 filtered out
```

After the fix, the same two tests plus the whole surrounding surface:

```
test core::delegation_authority::tests::agent_with_a_base_role_but_no_base_filename_stays_in_the_roster ... ok
test core::instruction_pipeline::tests::session_start_count_matches_the_delivered_delegation_roster ... ok
test daemon::doctor::doctor_fs_checks::tests::agents_message_reports_the_delegatable_roster ... ok
test daemon::doctor::doctor_fs_checks::tests::agents_roster_is_unknown_without_a_project_dir ... ok

test result: ok. 52 passed; 0 failed; 0 ignored; 0 measured; 4165 filtered out
```

`session_start_count_matches_the_delivered_delegation_roster` pins all three tiers (`$HOME` + `$CLAUDE_CONFIG_DIR` under a `#[serial]` RAII guard) so the assertion is a property of the test, not of the developer's machine — and it plants a cross-tier duplicate so a naive concatenation cannot pass it either. The five pre-existing `pipeline_*` tests moved onto the same guard, since the pipeline now resolves from the project rather than a caller-named directory.

## Gates

- `cargo fmt --check` — exit 0.
- `cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings` — exit 0 (the three GUI crates are CI's own exclusions; they need a built `ui/dist`).
- `cargo test -p trusty-mpm` — `4211 passed; 0 failed; 6 ignored` plus every integration binary green. This is the only crate the diff touches.
- `cargo test --workspace` (CI's command) — 152 binaries, **20563 passed, 5 failed**. All five are `trusty-search`'s FSEvents watcher tests (`service::watcher::*`, `service::watch_loop::*`, `watcher_manager`), which fail identically on this machine against the completely untouched crate in isolation. The diff contains zero `trusty-search` files.
- One further attribution: `client::executor::tests::execute_doctor_against_test_daemon` timed out once under full-workspace load. It passes 5/5 standalone and takes 9-13 s against this machine's 269-worktree store, right at the client's 10 s request timeout. The work this PR adds to that path was measured at **1.0-1.6 ms warm, 98 ms cold** — 0.01-1 % of that budget, not the cause.

Closes #4589
Closes #4588

### DIFF
--- a/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
+++ b/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
@@ -24,6 +24,11 @@ Fixed
   assets: the roster unions two tiers whose frontmatter tm does not author, so
   the same rule would eat an operator's own agent with no recourse
   ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
+- The bundled `agent-delegation` instruction section no longer tells the PM that
+  `memory-manager`, `mpm-agent-manager` and `mpm-skills-manager` are hidden by a
+  `role: base` filter. That paragraph shipped into every PM prompt and, once the
+  filter was removed, contradicted the live roster rendered directly beneath it
+  ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
 - `tm doctor`'s `agents` check reports the delegatable roster size alongside the
   deployed file count on every status — including the empty-deploy-tier failure,
   where the roster can still be non-empty because the project and generic
--- a/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
+++ b/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
@@ -13,7 +13,11 @@ Fixed
   ([#4588](https://github.com/bobmatnyc/trusty-tools/issues/4588)).
 - Agents are no longer excluded from the delegation roster because their
   frontmatter `role:` begins with `base`. Foundation templates are identified by
-  the `BASE-*` file-name convention alone. The frontmatter rule silently deleted
+  a case-insensitive `base-` file-stem prefix — the hyphen is load-bearing, so
+  `baseline-analyzer`, `base64-decoder` and `basecamp-sync` are ordinary
+  delegatable agents rather than templates, and every excluded file is logged at
+  debug level with its directory instead of vanishing silently. The frontmatter
+  rule silently deleted
   three real, deployed, dispatchable agents — `memory-manager`,
   `mpm-agent-manager`, `mpm-skills-manager` — from every roster the PM ever saw
   (34 advertised where 37 were deployed), and it could not be fixed by editing
@@ -21,8 +25,11 @@ Fixed
   the same rule would eat an operator's own agent with no recourse
   ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
 - `tm doctor`'s `agents` check reports the delegatable roster size alongside the
-  deployed file count, resolved by that same function, and reports it as
-  `unknown` rather than guessing when no project directory scopes the probe. A
+  deployed file count on every status — including the empty-deploy-tier failure,
+  where the roster can still be non-empty because the project and generic
+  `~/.claude/agents` tiers are independent of the deploy tier — resolved by that
+  same function, and reports it as `unknown` rather than guessing when no project
+  directory scopes the probe. A
   file count is a deploy fact, not a routing fact; reporting only the former is
   how a 42-file, 34-agent install read as healthy
   ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
--- a/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
+++ b/crates/trusty-mpm/changelog.d/4589-one-roster-resolver.md
@@ -1,0 +1,28 @@
+Fixed
+
+- The agent roster is resolved by exactly one function, so the count `tm session
+  start` prints can no longer disagree with the roster the PM receives. The
+  printed number came from a single-directory scan of the tm-managed deploy tier
+  while the delegation section delivered to the PM was a three-tier union
+  (project `.claude/agents` + `$CLAUDE_CONFIG_DIR/agents` + the operator's
+  `~/.claude/agents`) — two independent implementations of the same question,
+  measured live at 37 printed against 42 delivered. `build_instructions` now
+  takes the project rather than an agents directory and resolves the roster
+  through `delegation_authority::resolve_roster`, which the PM prompt and
+  `tm doctor` also call
+  ([#4588](https://github.com/bobmatnyc/trusty-tools/issues/4588)).
+- Agents are no longer excluded from the delegation roster because their
+  frontmatter `role:` begins with `base`. Foundation templates are identified by
+  the `BASE-*` file-name convention alone. The frontmatter rule silently deleted
+  three real, deployed, dispatchable agents — `memory-manager`,
+  `mpm-agent-manager`, `mpm-skills-manager` — from every roster the PM ever saw
+  (34 advertised where 37 were deployed), and it could not be fixed by editing
+  assets: the roster unions two tiers whose frontmatter tm does not author, so
+  the same rule would eat an operator's own agent with no recourse
+  ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
+- `tm doctor`'s `agents` check reports the delegatable roster size alongside the
+  deployed file count, resolved by that same function, and reports it as
+  `unknown` rather than guessing when no project directory scopes the probe. A
+  file count is a deploy fact, not a routing fact; reporting only the former is
+  how a 42-file, 34-agent install read as healthy
+  ([#4589](https://github.com/bobmatnyc/trusty-tools/issues/4589)).
--- a/crates/trusty-mpm/src/assets/instructions/sections/agent-delegation.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/agent-delegation.md
@@ -8,10 +8,9 @@
 > `CLAUDE_CONFIG_DIR/agents`, and the framework agents dir, project tier
 > winning on a name collision. The scan reads every `.md` file on disk, so it
 > picks up manifest-declared AND user-installed agents. That roster is
-> required; composition fails without it. Exception: agents whose frontmatter
-> carries `role: base` are filtered out as foundation templates, which
-> currently hides `memory-manager`, `mpm-agent-manager`, and
-> `mpm-skills-manager` (#4589).
+> required; composition fails without it. The ONLY agents filtered out are
+> foundation templates, identified by a `BASE-*` file name (the `base-` prefix
+> is required); frontmatter is never used to hide an agent (#4589).
 >
 > This file: crates/trusty-mpm/src/assets/instructions/sections/agent-delegation.md
 
--- a/crates/trusty-mpm/src/bin/tm/commands/session.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session.rs
@@ -617,8 +617,10 @@ pub(crate) fn compose_session_instructions(
     // and populate the metadata flags (agent_count, claude_md_created, …).
     let input = PipelineInput {
         framework_instructions_path: fw.framework_instructions_path(),
-        // #4409: scan the tier bundled agents actually deploy into.
-        agents_dir: fw.agent_deploy_dir(),
+        // #4588: the roster is resolved from the project by the one shared
+        // resolver, not from a directory named here — naming one tier is what
+        // made the printed count disagree with the delivered roster.
+        project_dir: project_dir.to_path_buf(),
         claude_md_path: project_dir.join("CLAUDE.md"),
     };
     let output = build_instructions(&input)?;
--- a/crates/trusty-mpm/src/core/delegation_authority.rs
+++ b/crates/trusty-mpm/src/core/delegation_authority.rs
@@ -102,16 +102,6 @@ fn parse_frontmatter(raw: &str) -> AgentFrontmatter {
     fm
 }
 
-/// Whether a role marks a foundation (non-delegatable) agent.
-///
-/// Why: `base-agent` / `base-engineer` and friends are inheritance foundations,
-/// not work destinations; advertising them would mislead the router.
-/// What: returns true when the role (lower-cased) starts with `base`.
-/// Test: `scan_excludes_base_agents`.
-fn is_base_role(role: &str) -> bool {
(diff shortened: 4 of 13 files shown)


bobmatnyc:
## Verdict: WARN

Independent adversarial review in a throwaway worktree off `origin/main`. Both red-then-green claims were re-run by mutating the production code myself, not taken on trust.

### What I verified (and it held)

| Claim | Method | Result |
|---|---|---|
| Exactly ONE roster resolver | Grepped every call site | **Holds.** `scan_agents` has one production caller (`roster_from_dirs:320`); `roster_from_dirs` has one (`resolve_roster:346`); `resolve_roster` has three consumers (`instruction_pipeline.rs:544`, `delegation_authority.rs:371`, `doctor_fs_checks.rs:100`). No surviving second implementation. `check_agent_reachability` is a tier check, not a roster |
| #4589 red-then-green | Restored the `role:`-prefix filter in `scan_agents` | `agent_with_a_base_role_but_no_base_filename_stays_in_the_roster` → **FAILED**, `left: [] right: ["memory-manager"]` |
| #4588 red-then-green | Reverted `build_instructions` to a single-directory `scan_agents` | `session_start_count_matches_the_delivered_delegation_roster` → **FAILED**, `left: 1 right: 4` |
| Count == delivered roster | Both derive from the same `resolve_roster` call in `build_instructions:544` | Structurally guaranteed, and the mutation above proves the gate is live |
| Cross-tier duplicate defeats naive concatenation | `qa` planted in both `managed` and `generic` tiers; test asserts `== 4` | Naive concatenation yields 5 and fails |
| Precedence project > user > bundled, no double-count | `deployed_agent_dirs` order + `or_insert` on a lower-cased name key | Correct; shadowed agent counted once |
| Suite green | `cargo test -p trusty-mpm --lib` ×7 | `4211 passed` in 6 of 7 (see Notes) |

**On the "unfalsifiable" question (#4589) — it is not, and the mutation shows why.** With the `role: base` filter restored, the new scanner-level test went red while the pre-existing asset-level gate `every_bundled_non_base_agent_reaches_the_roster` stayed **green**. That is exactly the author's stated point: because option (a) already landed, the asset-level gate could no longer catch a regression in the *rule*. The reconstruction is faithful and the new test is the real falsifier.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/delegation_authority.rs` | 164 | **The rule is `base*`, not `BASE-*`.** `if stem.to_ascii_lowercase().starts_with("base")` has no separator requirement, yet the new doc block (114-123) says "`BASE-*` FILE NAME convention" four times and the new test is named `..._but_no_base_filename_...`. Probe over `baseline-analyzer`, `base64-decoder`, `basecamp-sync`, `based-reviewer`, `BASE-AGENT`, `database-migrator`, `engineer` → **only `database-migrator` and `engineer` survive `scan_agents`.** The PR's own argument against the `role:` rule — "will silently delete an operator's agent from `<project>/.claude/agents` or `~/.claude/agents`, and no asset edit can fix it" — applies verbatim to its replacement for any agent named `base…`. The failure moved rather than being removed. Compounding it: `BASE-*` appears only in `docs/specs/` and `docs/adr/`, never in user-facing docs, so an operator cannot discover the rule that ate their agent | `starts_with("base-")`. **Verified:** all 52 `delegation_authority` + `instruction_pipeline` + `doctor_fs_checks` tests stay green with that one-character change, and all 5 bundled foundation files (`BASE-AGENT/ENGINEER/OPS/QA/RESEARCH.md`) are still excluded. Then make the doc and the code agree. |
| MEDIUM | `crates/trusty-mpm/src/daemon/doctor_fs_checks.rs` | 87-94 | The `md_count == 0` **Fail** branch returns *before* `roster` is computed, so it reports neither a count nor `unknown` — contradicting the new doc at 71-78 ("Every message also carries the DELEGATABLE roster size"). Worse, the deploy tier can be empty while the project and generic tiers carry a full roster: probe returned `status=Fail`, message `"no agent files … run \`tm install\`"`, containing neither `delegatable` nor `unknown`, while `resolve_roster` returned **42**. That is the same deploy-fact-vs-routing-fact conflation this PR exists to remove, left standing on one path | Move the `roster` computation above the `md_count == 0` early return and append it to the Fail message too. |
| LOW | `crates/trusty-mpm/src/core/delegation_authority.rs` | 380 | `deployed_roster_section` calls `deployed_agent_dirs(project_dir)` a second time purely to log `tiers = …`, after `resolve_roster` already built and dropped the same vec — re-doing the work the one-resolver refactor centralised | `let dirs = deployed_agent_dirs(project_dir); let agents = roster_from_dirs(&dirs);` and log `dirs.len()`. |
| LOW | `crates/trusty-mpm/src/core/delegation_authority.rs` | 362-363 | Doc line left unwrapped past the surrounding width after the edit ("…renders it with [`generate_authority`]. Returns `None` when no agent is deployed") | Re-wrap. |

### Required changes

1. Fix the HIGH. `starts_with("base-")` is a one-character change with zero test breakage, and without it the PR's central safety claim is not what the code does.
2. Fix the MEDIUM, or narrow the doc comment at `doctor_fs_checks.rs:71-78` so it stops claiming a property the Fail branch does not have.

### Notes

- `client::executor::tests::execute_doctor_against_test_daemon` failed **1 of 7** full-suite runs on this branch and **0 of 3** on `origin/main`. That is not enough to attribute causally and I am not raising it as a finding — but this PR does add `resolve_roster` (three directory scans) to the doctor path, and the author's own note puts that test at 9-13 s against a 10 s client timeout. Worth a second look if it recurs in CI.
- Q2 answered cleanly: there is genuinely one resolver now, and it is the strongest part of this PR. The `PipelineInput.agents_dir → project_dir` change is the right shape — it removes the parameter a caller could get wrong rather than documenting how to get it right.
- The first assertion in `session_start_count_matches_the_delivered_delegation_roster` (count == rendered entries) is structurally guaranteed once both sides call one resolver; the `== 4` assertion is what carries the dedup weight. Both are worth keeping — the first is the architectural anchor that goes red if a second resolver reappears, which my mutation confirmed.

Issue #4307, opened 2026-07-29
A test filter that matches nothing reports ok — add a CI guard for zero-tests-ran


bobmatnyc:
## Defect

`cargo test` with a filter that matches zero tests **exits 0 and prints `ok`**. Where a test module's file path differs from its module path via `#[path]`, the natural filter — the one derived from the file name — silently matches nothing.

`crates/trusty-git-analytics/src/collect/identity/resolver.rs:510-512`:

```rust
#[cfg(test)]
#[path = "resolver_tests.rs"]
mod tests;
```

File is `resolver_tests.rs`; module is `collect::identity::resolver::tests`. One character of path drift turns red into a passing run.

## Evidence

Reproduced against `origin/main` @ `aed80140`, `cargo test -p tga --lib <filter>`:

```
filter 'collect::identity::resolver_tests'  -> running 0 tests
                                               test result: ok. 0 passed; 775 filtered out
                                               exit code 0

filter 'collect::identity::resolver::'      -> running 33 tests
                                               test result: ok. 33 passed; 742 filtered out
                                               exit code 0
```

Both green. Only one ran anything.

- Surfaced by code-critic during PR #4291 review; scoped out of that PR deliberately. It was the leading hypothesis for an anomalous all-green run against known-broken code before the true cause (a temp-dir race) was isolated — plausible precisely because the mechanism is real.
- Exposure on `origin/main`: **319** `#[path]`-attached test modules across 13 crates; **231** of them have file stem != module name, i.e. the file-name-derived filter does not resolve. (`trusty-agents`, `trusty-agents-common`, `trusty-analyze`, `trusty-code`, `trusty-common`, `trusty-embedderd-py`, `trusty-git-analytics`, `trusty-installer`, `trusty-kb`, `trusty-memory`, `trusty-mpm`, `trusty-review`, `trusty-search`.)
- No existing guard: grep for `running 0 tests` / `filtered out` / min-count across `scripts/` and `.github/workflows/` returns nothing.
- CI itself currently runs only unfiltered `cargo test --workspace` (`.github/workflows/ci.yml:564`, `:571`), so today's exposure is at the **evidence** layer — agents and humans producing "green" from filtered local runs — plus any future filtered CI step.
- Defeats the repo's standing "verify green, not green-by-disabling" rule while satisfying its letter: a zero-tests run *is* observed as `... ok`.
[...]

Issue #4097, opened 2026-07-27
bug(ci): check_line_cap.sh passes vacuously on unstaged new files


bobmatnyc:
## Summary

`scripts/check_line_cap.sh` enumerates the files it inspects via `git ls-files '*.rs'` (lines 160, 186, 283). `git ls-files` lists **tracked** files only. A newly created, unstaged `.rs` file is therefore completely invisible to the check — the script walks right past it and reports success.

## Observed live, not theoretical

During work on PR #4096, a new 572-SLOC file was created:

```
crates/trusty-mpm/src/daemon/services/delegation_tracker.rs
```

With that file present but unstaged, `scripts/check_line_cap.sh` reported:

```
7 allowlisted, 0 violations — OK
```

...while the file was well over the 500-SLOC production cap. The violation only surfaced after `git add`.

## Why it matters

The failure mode is a **false green on a quality gate**. A developer or agent running the check locally before staging gets an all-clear that is meaningless. The violation is discovered later — or not at all, if the file is committed in the same motion as a passing local run.

A gate that reports OK when it has inspected nothing is worse than no gate, because it manufactures unearned confidence. The output is indistinguishable from a genuine pass.

## Fix direction (options, not a settled decision)

1. **Enumerate untracked-but-not-ignored files too**, e.g.:
   ```
   git ls-files --cached --others --exclude-standard '*.rs'
   ```
   This closes the visibility hole directly. Needs a look at whether any generated/vendored `.rs` output currently escapes only by virtue of being untracked-and-unignored.

2. **Report how many files were actually scanned**, so a vacuous pass is visible in the output rather than silent — e.g. `scanned 412 files, 7 allowlisted, 0 violations — OK`. A run that prints `scanned 0 files ... OK` is then self-evidently broken.

Option 2 is worth doing regardless of whether option 1 lands. It converts a silent-failure class into a visible one for this check and costs almost nothing.

## General principle

This is the same class as the other "green by omission" traps in this repo: a check that turns green by not examining anything is indistinguishable in CI output from a check that genuinely passed. Any gate whose scope is derived (from a glob, a git query, a test filter) should state the size of the scope it actually covered.

## Reference

Observed during PR #4096.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as not-planned — deliberate backlog prune of over-filing. The finding is **real**; it just doesn't need a backlog row.

[...]

Issue #3732, opened 2026-07-23
Dissolve crates/cto-assistant: convert to Bob's local agent (not a workspace crate)


bobmatnyc:
This should not be a crate, should be a local agent for me: https://github.com/bobmatnyc/trusty-tools/tree/main/crates/cto-assistant

## Refinement

This directive refines the standing 07-19 decision (#3052/#3105) that cto-assistant migrates into trusty-agents as a personal agent (not a 9th product family). The migration/conversion work belongs to epic #3052 territory (see issue #3105 "combine into the Assistant" and the 2026-07-23 gap analysis on getting izzie + cto-assistant running with all their standalone tools) and belongs to Bob's trusty-agents workstream — THIS issue tracks the repo-structure counterpart: **crates/cto-assistant should be dissolved as a workspace crate once the local-agent form reaches parity.**

## Investigation

**Publication status:** cto-assistant is NOT published to crates.io (`publish = false` in Cargo.toml). No deprecation/yanking step required.

**Current state:**
- Version: 0.1.3
- Workspace dependents: `trusty-agents-local` only (directly imports: `cto-assistant = { path = "../cto-assistant", version = "0.1.3" }`)
- Functionality: ToolExecutor adapter over `tc_services::cto_db`, wrapping four HR/budget query tools in the cto-assistant persona

**Related issues (cross-link):**
- #3105 — [Goal] Convert the Duetto "CTO Assistant" bot into a trusty-agent
- #3656 — arch(trusty-agents): cto-assistant is a coded agent — four tools reachable only via Rust, violating DOC-41 §2.0 declarative-only
- #3700 — cto-assistant DB tools unreachable in all binaries after #3310 (no plugin install) [filed TODAY]
- Epic #3052 (conversion roadmap)

## Checklist for Dissolution

- [ ] **Parity gate:** Confirm local-agent form of cto-assistant reaches functional parity (gated by #3105/#3052 work; trusty-agents workstream owns this)
- [ ] **Dependent untangling:** Remove `cto-assistant` import from trusty-agents-local; migrate tool registration to local-agent bootstrapping
- [ ] **Workspace crate removal:** Delete crates/cto-assistant from Cargo.toml and filesystem
- [ ] **Cross-crate cleanup:** Remove stale references in trusty-agents comments/docs

[...]

PR #4621, opened 2026-08-02
fix(ci): assert a non-zero scan floor so gate scripts cannot pass while checking nothing (#4618)

### LINKED ISSUE #4618: Gate scripts can pass while checking nothing — assert a non-zero scan floor
## Defect

9 of this repo's CI-wired gate scripts can report success while examining zero files. Every one already computes the number that would expose it and none asserts a floor on it, e.g. [check_agent_assets.sh:312](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_agent_assets.sh#L312) (`... byte-parity file(s) match — OK`), [check_generation_artifacts.sh:334](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_generation_artifacts.sh#L334) (`0 leaked artifacts — OK`), [trusty-sld-lint/src/main.rs:149](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/crates/trusty-sld-lint/src/main.rs#L149) (`scanned {} spec doc(s) ...`), [trusty-publish-guard/src/main.rs:148](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/crates/trusty-publish-guard/src/main.rs#L148) (`checked {checked} publishable crate(s) ...`).

Exactly one gate does it right: [check_token_drift.test.mjs:185](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_token_drift.test.mjs#L185) — `"every ENFORCED crate compares >0 real tokens against canonical"`. That is the pattern to generalize.

## Evidence

Every script below was mutation-tested by hand; no gate is broken — all catch their true positive and name the offender. Ranked weakest first.

1. **[check_agent_assets.sh:305](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_agent_assets.sh#L305)** — the only gate that passes on a git ERROR, not merely an empty result. `done < <(git ls-files ...)`: process substitution is exempt from both `set -e` and `pipefail`. With `git ls-files` shimmed to exit 128 and a genuinely drifted parity copy on disk: `0 byte-parity file(s) match — OK`, exit 0.
2. **[check_generation_artifacts.sh:334](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_generation_artifacts.sh#L334)** — silent vacuous pass, zero warning output, because its allowlist is correctly empty. Broken globs plus two real leaked artifacts: `0 leaked artifacts — OK`, exit 0.
3. **[check_changelog_fragment.sh:202](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_changelog_fragment.sh#L202)** — fail-open probe, `git cat-file -e ... || continue`, reads any git error as "the PR deleted this crate" (the exemption). Shimmed: [check_changelog_fragment.sh:236](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_changelog_fragment.sh#L236) prints `no crate source changed ... — OK`, exit 0, on a PR with a real unrecorded source change.
4. **[check_line_cap.sh:332](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_line_cap.sh#L332)** — vacuous behind `.line-cap-allowlist.tsv`'s 7 current grandfathered rows; goes fully silent once the allowlist ratchets to zero, which is the stated goal.
5. **[check_test_pointers.sh:63-65](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_test_pointers.sh#L63-L65)** — protected only by 920 accidental rows in `.test-pointer-allowlist.tsv` ([check_test_pointers.sh:855](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_test_pointers.sh#L855)). Separately, it cannot see a citation that is not backtick-quoted — a bare-word dangling pointer passes clean by design (documented at the cited lines).
6. **[check_doc_numbers.sh:656](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_doc_numbers.sh#L656)** — same accidental protection, `.doc-number-allowlist.tsv` is 4 rows deep, one `--update` from losing it.
7. **[check_claude_md_not_tracked.sh:28](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/scripts/check_claude_md_not_tracked.sh#L28)** — `git ls-files --error-unmatch CLAUDE.md 2>/dev/null` cannot distinguish "untracked" from "git broke".
8. **[trusty-sld-lint/src/main.rs:149-155](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/crates/trusty-sld-lint/src/main.rs#L149-L155)** and **[trusty-publish-guard/src/main.rs:148-159](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/crates/trusty-publish-guard/src/main.rs#L148-L159)** (backing `check_sld.sh` / `check-version-parity.sh`) — honest scan counts (`spec_docs`, `checked`), no floor asserted; exit code depends only on `is_clean()` / `drifted == 0 && unverifiable == 0`, so `checked 0 publishable crate(s) — 0 drifted, 0 unverifiable.` exits 0.

Two of these are a strictly worse condition than an empty set — a tool error mistaken for "nothing to check": item 1's process substitution and item 3's `|| continue`.

## Resolution

Give each gate a non-zero scan-floor assertion and a paired self-test in the same CI job, following the `check_token_drift.test.mjs` precedent. Separately fix the two genuine fail-open constructs (items 1 and 3), which fail open on a tool error rather than an empty set.

## Also note (lower severity)

`check_test_pointers.sh` runs at ~97% of its [test-pointers.yml:30](https://github.com/bobmatnyc/trusty-tools/blob/01e4bf9bba95e25b538f554c5d2d2db44c31392b/.github/workflows/test-pointers.yml#L30) `timeout-minutes: 5` budget — 3m47s-4m51s across the last 8 completed CI runs. It is advisory, not required, so a timeout is noise rather than a block today — but it will start failing for reasons unrelated to what it guards.

## Cross-reference, do not merge into

#4179 (workflow layer umbrella — this issue is the script-layer counterpart), #4614 (closed — same shape one layer up: a piped `cargo test` masked failures as success), #4604, #4605, #4589, #4097 (closed — `check_line_cap.sh` vacuous on unstaged files, a different trigger for the same script's weakness), #4307 (a test filter matching zero tests reports `ok` — same "zero examined, still green" shape, different layer: `cargo test` filters, not CI gate scripts).

bobmatnyc:
## The defect

A gate script that iterates a file set and exits 0 when that set is **empty** reports success identically to one that scanned everything and found nothing wrong. A required check that cannot fail is not a check.

Every affected gate already computed the number that would expose it — `PARITY_COUNT`, `spec_docs`, `checked` — and none asserted a floor on it. The exit code depended only on the finding count, and the finding count is legitimately 0 on a healthy tree.

This is the fourth time this shape has been fixed. [#4097](https://github.com/bobmatnyc/trusty-tools/issues/4097) fixed it in `check_line_cap.sh`, [#4307](https://github.com/bobmatnyc/trusty-tools/issues/4307) in a zero-match `cargo test` filter, [#4614](https://github.com/bobmatnyc/trusty-tools/issues/4614) in a piped `cargo test` exit code. This PR sweeps all nine remaining instances at once rather than adding a fifth one-at-a-time fix.

## The fix

Every affected gate now counts what it actually examined, fails loudly below a declared minimum, and **prints the count on success** so a future regression shows up in the CI log instead of silently.

Floors are declared constants set far below current reality and far above zero, so a partially-broken enumeration trips them too — not just a totally empty one. Nothing was made to pass by weakening it: no exclusions, no `|| true`, no narrowed globs. Every gate's true positive was re-verified after the change.

`scripts/check_scan_floor_selftest.sh` is new: it re-proves each floor in CI by running the gate against a throwaway repo whose scan set is empty (or whose git enumeration is shimmed to fail) and asserting a non-zero exit. Without it the floors are untested code that a later refactor can silently undo. It generalizes the one gate that already did this right — `check_token_drift.test.mjs`'s "every ENFORCED crate compares >0 real tokens against canonical". Each gate's workflow runs its own case before the real gate, mirroring `line-cap.yml`'s existing selftest-then-gate ordering.

## The two fail-open cases (strictly worse than an empty set)

These did not merely scan nothing — the tool actively **failed** and the gate still exited 0.

### 1. `check_agent_assets.sh` — process substitution swallowed the exit status

`done < <(git ls-files "$TCODE_DIR/*.md")`. A process substitution runs in a subshell exempt from **both** `set -e` and `pipefail`, so a `git ls-files` exiting 128 arrived as an empty stream and the gate printed `0 byte-parity file(s) match — OK` over a genuinely drifted copy.

Fixed by materialising the enumeration into a temp file whose exit status is checked, then redirecting the loop from that file — which also keeps the loop body in the current shell, so `PARITY_COUNT` and `FAIL` still survive it.

### 2. `check_changelog_fragment.sh` — `git cat-file -e … || continue` conflated absence with failure

`git cat-file -e HEAD:<path>` exits **128 for both** "the path is legitimately absent from that tree" and "git itself failed". So `|| continue` read every git error as the crate-deleted exemption, and the gate printed `no crate source changed … — OK` on a PR with a real unrecorded source change.

Fixed with `git ls-tree`, which separates the two cases that `cat-file -e` cannot: it exits 0 with **empty output** for an absent path, and non-zero only on a genuine git failure — now escalated to a hard gate failure rather than swallowed. The same construct guarding the `TRANSITIONAL` probe was fixed with it. The legitimate crate-deletion exemption ([#3732](https://github.com/bobmatnyc/trusty-tools/issues/3732)) still works — re-verified.

## Mutation proof

Every row was run in both directions. "Pre-fix" is the same mutation applied to `origin/main`'s copy of the script, proving each case used to pass. All eight shell cases are now permanent CI assertions in `check_scan_floor_selftest.sh`; the two Rust floors are unit tests.

| Script | Normal run: exit + scanned count | Mutation | Pre-fix | Post-fix |
|---|---|---|---|---|
| `check_agent_assets.sh` | 0 — `scanned 30 byte-parity file(s) (floor 20)` | `git ls-files` shimmed to exit 128, with a genuinely drifted copy on disk | **0** — `0 byte-parity file(s) match — OK` | **1** — `FAIL: TOOL ERROR — 'git ls-files …' exited non-zero` |
| `check_changelog_fragment.sh` | 0 — `scanned 2 changed path(s)` | crate-existence probe shimmed to exit 128, with a real unrecorded source change | **0** — `no crate source changed … — OK` | **1** — `FAIL: TOOL ERROR — 'git ls-tree HEAD -- crates/demo/Cargo.toml' failed` |
| `check_changelog_fragment.sh` | (same) | empty diff against the base | **0** — `no changes against HEAD — nothing to check` | **1** — `FAIL: SCAN FLOOR — the diff …..HEAD lists 0 changed path(s)` |
| `check_generation_artifacts.sh` | 0 — `scanned 16 stylesheet(s) + 1119 prose file(s) (floors 5/200)` | repo with 0 scannable files | **0** — `0 leaked artifacts — OK` | **1** — `FAIL: SCAN FLOOR — only 0 stylesheet(s) scanned` |
| `check_line_cap.sh` | 0 — `measured 3571 tracked .rs file(s) (floor 500)` | repo with 0 tracked `.rs` files | **0** — `0 allowlisted, 0 violations — OK` | **1** — `FAIL: SCAN FLOOR — only 0 tracked .rs file(s) were measured` |
| `check_test_pointers.sh` | 0 — `resolved 20124 Test: citation(s) (floor 200)` | repo with 0 tracked `.rs` files | **0** — `0 dangling pointers — OK` | **1** — `FAIL: SCAN FLOOR — only 0 Test: citation(s) were resolved` |
| `check_doc_numbers.sh` | 0 — `scanned 95 doc(s) / 89 claim(s) (floors 20/20)` | repo with 0 spec/ADR documents | **0** — `0 grandfathered, 0 violations — OK` | **1** — `FAIL: SCAN FLOOR — scanned 1 doc(s) yielding 0 number claim(s)` |
| `check_claude_md_not_tracked.sh` | 0 — `probed against 5537 tracked file(s)` | git index below the floor (probe proves nothing) | **0** — `PASS: root CLAUDE.md is not tracked` | **1** — `FAIL: SCAN FLOOR — git reports only 2 tracked file(s)` |
| `trusty-sld-lint` (`check_sld.sh`) | 0 — `scanned 54 spec doc(s) + N code file(s)` | `scan_floor_violation(0, 0)` and each single-axis zero | **0** — exit depended only on `is_clean()` | `Some(SCAN FLOOR …)` → `ExitCode::FAILURE` (`tests::floor_rejects_empty_scan`) |
| `trusty-publish-guard` (`check-version-parity.sh`) | 0 — `checked N publishable crate(s) (floor 10)` | `scan_floor_violation(0)` | **0** — `checked 0 publishable crate(s) — 0 drifted` | `Some(SCAN FLOOR …)` → `exit(1)` (`tests::floor_rejects_empty_scan`) |

**True positives re-verified after the change** (the gates still catch what they exist to catch, and pass again once restored):

| Gate | Injected defect | Result |
|---|---|---|
| `check_agent_assets.sh` | byte-drifted a parity copy | exit 1 — `FAIL: COPY DRIFTED — …/research.md` |
| `check_generation_artifacts.sh` | appended `</invoke>` to a tracked doc | exit 1 — `FAIL: docs/specs/README.md:209: leaked generation artifact` |
| `check_line_cap.sh` | new 600-SLOC production file | exit 1 — `FAIL: … is 600 SLOC (> 500 prod cap) and not allowlisted` |
| `check_claude_md_not_tracked.sh` | tracked a root `CLAUDE.md` | exit 1 — `FAIL: root CLAUDE.md is tracked by git` |
| `check_changelog_fragment.sh` | unrecorded crate source change | exit 1 — `FAIL demo: crates/demo/src/** changed with no changelog record` |
| `check_changelog_fragment.sh` | crate deleted outright (must stay **exempt**) | exit 0 — exemption preserved ([#3732](https://github.com/bobmatnyc/trusty-tools/issues/3732)) |
| `check_doc_numbers.sh` / `check_test_pointers.sh` | their own `--self-test` fixtures | both OK |

## Not defective — deliberately left alone

- **`check_buildrs_sync.sh`** — iterates a hardcoded 4-element array and treats a missing file as an ERROR. Its scan set cannot be empty.
- **`check_instruction_floor.sh`** — already fails on an empty fixed-section set, on a missing pinned artifact, and on a missing digests file, and compares digests byte-exactly. Already floored.
- **`check_capabilities.sh`** / **`check_sld.sh`** — thin `exec` wrappers with no scan of their own; `check_sld.sh`'s scan is floored in the binary it calls.

## Verification

- `cargo fmt --check` — clean.
- `cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings` — exit 0, zero warnings (CI's exact invocation; the three Tauri crates need WebKit2GTK / a built `ui/dist` and are excluded by `ci.yml` itself, not by this PR).
- `cargo test --workspace` (same CI exclusions) — see the run output in the PR checks.
- Every `scripts/check_*.sh` run against this branch, including the siblings this PR does not touch — all green.

Closes #4618


bobmatnyc:
## Verdict: APPROVE

Adversarial review from an independent read-only worktree (detached at `936ce14f`, base `01e4bf9b`). I did not trust the PR body's mutation table — I re-ran every row myself, plus the inverse mutations the table does not contain.

### What I actually did (not a re-read of the description)

**1. The self-test is not vacuous — all 8 cases were individually broken and all 8 went red.**

I built a throwaway repo holding a copy of `scripts/`, removed one floor at a time, and ran only that gate's case. Every one detected it:

| Mutation | Self-test result |
|---|---|
| `MIN_RS_FILES=500` → `0` | FAIL — `line-cap: measured 0 tracked .rs file(s) (floor 0); 0 violations — OK.` |
| `MIN_CSS_FILES`/`MIN_PROSE_FILES` → `0` | FAIL — `scanned 0 stylesheet(s) + 0 prose file(s) … OK.` |
| `MIN_CITATIONS=200` → `0` | FAIL — `resolved 0 Test: citation(s) … OK.` |
| `MIN_DOCS_SCANNED`/`MIN_CLAIMS` → `0` | FAIL — `scanned 1 doc(s) / 0 claim(s) … OK.` |
| `MIN_TRACKED_FILES=500` → `0` | FAIL — `probed against 2 tracked file(s)` |
| empty-diff floor disabled (`-lt 1` → `-lt 0`) | FAIL — `scanned 0 changed path(s) … OK.` |
| `check_agent_assets.sh` reverted to `done < <(git ls-files …)` + floor 0 | FAIL — `scanned 0 byte-parity file(s) … OK.` |
| `path_exists_at_rev` reverted to swallowing the git failure | FAIL — `scanned 1 changed path(s); no crate source changed … OK.` |

Zero cases pass for the wrong reason. `expect_rejects` additionally requires the output to name `SCAN FLOOR`/`TOOL ERROR`, so a gate failing for an unrelated reason is reported as a self-test failure, not a pass.

**2. The harness fails closed on its own failure modes.** `bash check_scan_floor_selftest.sh ghost` (name in `KNOWN`, no `want` block) → `0 cases ran — this self-test just proved nothing about itself`, exit 1. Unknown name → exit 2. Gate script deleted while its case is still wired → `cp` fails under `set -e`, non-zero. Zero-arg run still executes all 8. Clean run leaves no fixture or temp-file residue.

**3. Every reported count is real, not hardcoded or stale.** Cross-checked independently: `git ls-files '*.rs'` = 3571 (gate says 3571); `git ls-files` = 5540 (gate says 5540); tcode agent `.md` = 38, minus 4 tcode-only minus 4 pinned = 30 (gate says 30). Live runs: doc-numbers 95/89, generation-artifacts 16 + 1121, test-pointers 20124, sld-lint 53 + 3010, publishable crates 20.

**4. Floor values are sound.** Margins 7.1x / 11.1x / 100x / 4.7x / 3.2x / 1.5x / 2.7x–15x / 2x. The two tightest are `MIN_PARITY_FILES=20` (30 actual) and `MIN_CSS_FILES=5` (16 actual) — both still well clear, and both fail with a message naming the constant and its file. Shallow clone is not a hazard: every floor reads the index or the worktree, not history, and the one gate needing history (`changelog-fragment.yml`) already pins `fetch-depth: 0`.

**5. Both fail-open fixes distinguish failure from absence.** Genuine tool error → red (self-test cases 7/8). Legitimate absence → still green, verified on the real tree: crate deleted outright (`git rm -r crates/trusty-publish-guard`) → exit 0, `#3732` exemption intact; unrecorded crate source change → exit 1; this PR against `origin/main` → exit 0 with both fragments validated.

**6. True positives survive — sampled independently, not taken from the table.** 600-SLOC prod file → `1 violation(s) — FAILED`; `</invoke>` appended to a tracked doc → `docs/specs/README.md:209: leaked generation artifact`; tracked root `CLAUDE.md` → `FAIL: root CLAUDE.md is tracked by git`; byte-drifted `research.md` → `agent-assets: FAILED`; duplicated `DOC-30` → `[DUP-DOC] DOC-30 is claimed by 2 files`; `check_doc_numbers.sh --self-test` and `check_test_pointers.sh --self-test` both still OK.

**7. The two "left alone" judgments are correct.** `check_buildrs_sync.sh` iterates a hardcoded 4-element `FILES` array and sets `FAILED=1` on a missing file (`:42-46`) — its scan set cannot be empty. `check_instruction_floor.sh` exits on an empty `fixed_sections` set (`:120-124`), a missing manifest (`:98-101`), and missing pinned artifacts (`:154-155`), then compares byte-exact digests. Already floored.

**8. Workflow wiring blocks.** All 7 steps are plain `run:` (default shell `bash -e {0}`), no `continue-on-error`, no `shell:` override, and each sits before its real gate. A self-test failure fails the step and the job. Floors sit before the `--seed`/`--update`/`--self-test` branches in `check_line_cap.sh`, `check_doc_numbers.sh` and `check_test_pointers.sh`, so a broken enumeration cannot regenerate an allowlist. `cargo test -p trusty-sld-lint -p trusty-publish-guard --bins` → `2 passed; 0 failed` twice. All 26 PR checks green.

### Findings

|---|---|---|---|---|
| MEDIUM | `crates/trusty-sld-lint/src/main.rs` | 193-196 | The floor's **call site** is untested. `scan_floor_violation` is unit-tested as a pure predicate, so deleting the `if let Some(msg) = …` block from `main` leaves both unit tests green and restores the exact vacuous pass. This is the "the floors are themselves untested code" hazard the PR's own header calls out — the 7 shell gates got a mutation harness, the 2 Rust gates did not. | Add `crates/trusty-sld-lint/tests/scan_floor.rs` invoking `env!("CARGO_BIN_EXE_sld-lint")` with `--root <empty tempdir>`, asserting non-zero exit and `SCAN FLOOR` on stderr. |
| MEDIUM | `crates/trusty-publish-guard/src/main.rs` | 183-186 | Same — `floor_rejects_empty_scan` passes with the `main` wiring deleted. | Same shape via `CARGO_BIN_EXE_publish-guard` with `--root <empty tempdir>` (no network needed: zero crates discovered). |
| LOW | `scripts/check_changelog_fragment.sh` | 162, 169 | `out="$(git ls-tree … 2>&1)"` folds stderr into the value that line 169 uses as the existence signal (`[[ -n "$out" ]]`). Any warning git prints on an otherwise-successful `ls-tree` makes a definitively-absent path read as present — which re-breaks the #3732 crate-deletion exemption as a false red. Fail-closed, so not urgent, but the merge is unnecessary. | Capture stderr separately: `local err; err="$(mktemp)"; if ! out="$(git ls-tree -r --name-only "$rev" -- "$path" 2>"$err")"; then … cat "$err" … fi`. |
| LOW | `scripts/check_generation_artifacts.sh` | 269, 305, 362-375 | SEED MODE and UPDATE MODE both `exit 0` before the floor block, unlike the three sibling gates which floor before their mode branches. `--update` over a broken enumeration would prune every allowlist row. Impact is fail-safe (stricter, and manual-only), but the asymmetry looks unintended. | Move the two floor blocks above the `if [ "$MODE" = "seed" ]` branch. |
| LOW | `scripts/check_generation_artifacts.sh` | 196 | Comment states "far below the current reality (20 stylesheets, 1121 prose files)". 20 is the raw `git ls-files` count; the gate actually scans **16** after `is_excluded_path`. The real margin is 16/5, not 20/5. In a PR about honest scan counts, the count in the comment should be the one the gate prints. | `20 stylesheets` → `16 stylesheets`. |
| LOW | `scripts/check_doc_numbers.sh`, `scripts/check_generation_artifacts.sh` | 170-176 / 205-215 | `doclist`/`adrlist` and `css_list`/`prose_list` are `mktemp`'d without a `trap`, so a `set -e` abort between creation and the trailing `rm -f` leaks them. `check_line_cap.sh` and `check_agent_assets.sh` correctly use traps for their new temp files. | Register the paths in the existing `EXIT` trap. |

### Notes

- **Residual, out of scope:** the `KNOWN` roster and the 7 workflow wirings are manual. Nothing detects a *new* gate script landing without a floor and a case, or a self-test step being deleted from a workflow. That is the honest limit of any self-test; worth a follow-up issue rather than a change here.
- `check_capabilities.sh` is classified as a thin `exec` wrapper "with no scan of its own". Unlike `check_sld.sh` the PR does not claim the callee (`tm generate capabilities --check`) is floored. Probably fine — a broken generator against non-empty committed files still diffs red — but it is the one classification I could not fully close out. Follow-up, not a blocker.
- The empty-diff floor will go red on a net-zero PR (commit + revert, or a back-merge with no net diff) that previously exited 0. Deliberate, `pull_request`-only so a push to main cannot trip it, and the message names the cause. Accepted.

PR #4622, opened 2026-08-02
fix(trusty-mpm,trusty-common): make self-diagnosis report UNKNOWN instead of OK for what it cannot verify (#4469, #4033, #4604)

### LINKED ISSUE #4033: tm doctor: no visibility into installed binary provenance — stale path-installed binaries bypass update detection
## Summary

**Diagnostic-trust problem**: `tm doctor` does not detect whether an installed trusty-* binary is stale or orphaned relative to what is published on crates.io, or whether its install source still exists.

This is a companion gap to #4001 (thread contention visibility) — same family of "doctor reports healthy while reality is unhealthy."

## Concrete Evidence (2026-07-26)

Bob's machine observed ten trusty-* binaries installed via `cargo install --path` from git worktrees rather than from crates.io:

- **trusty-search** 0.39.0 from a worktree; 0.39.1 published that afternoon with two crash-safety fixes (#3970, #1717). The fixes were not running.
- **tctl** / **trusty-installer** 0.4.3 from a worktree; 0.4.10 published that afternoon with five demo-install fixes. Seven patch versions stale.
- **trusty-channels** installed TWICE from two DIFFERENT worktrees, providing conflicting binaries (`slack-mcp` from one, `telegram-mcp` from another).
- One install source under `/Users/masa/tmp/`, a directory macOS reaps.
- Several sources under `.claude/worktrees/` (auto-cleaned when unchanged) or `.worktrees/` (tm prunes stale ones) — source can be deleted while binary keeps running, leaving no provenance and no upgrade path.

## Why It Matters

Install provenance directly impacts verification fidelity. A test run, benchmark, or bug repro against a stale binary produces a confident wrong answer. The installed 0.39.0 was missing two crash-safety fixes; an operator testing the "fix" against 0.39.0 would incorrectly conclude it was not working.

This is the same shape as #4001: doctor answers "is it running?" but not "is what's running what you think it is?"

## What doctor Should Check

- Flag any trusty-* binary installed from a PATH (not the registry).
- Report whether its source directory still exists.
- Compare installed version against latest published on crates.io; flag a mismatch.
- Note where a crate is installed more than once from different sources.

**Caveat**: Where a crate is genuinely unpublished (`trusty-channels`, possibly `trusty-code`), a path install is the ONLY option — doctor should say that rather than flag it as an error.

## References

- Related: #4001 (thread contention visibility — same "appears healthy but isn't" family)
- Root-cause pattern: doctor's health model is liveness + file state; it does not observe binary versioning or install provenance (unlike #4001's thread inspection gap, this is also a data model gap)

---

### LINKED ISSUE #4469: trusty-common: is_launchd_supervised() env-var heuristic lets a child self-report as supervised
**Defect:** `trusty_common::update::is_launchd_supervised()` at [`crates/trusty-common/src/update/upgrade.rs:272`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-common/src/update/upgrade.rs#L272) returns `true` when `XPC_SERVICE_NAME` is set and `TERM_PROGRAM` is not. A `tm daemon` child process spawned without `TERM_PROGRAM` inherits that condition, so it can self-report as supervised, pass the [#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230) callee-side guard, and publish `supervised: true` on `/health`.

**Why it matters:** this is strictly worse than the [#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230) orphan. That one at least reported `supervised: false` and could be caught. This variant makes the flag operators are instructed to trust actively agree with the orphan. `CLAUDE.md`'s "Connection-safe daemon restart convention" tells operators to verify a restart with `curl /health | jq '.supervised'` — this defect makes that verification unsound in exactly the case it exists to catch.

**Evidence:** environment-variable heuristic, not a real supervision query. Contrast with the approach [#4466](https://github.com/bobmatnyc/trusty-tools/pull/4466) deliberately chose for its new caller-side guard: keyed on plist presence alone, with no heuristic, specifically to avoid this.

Not what caused [#4230](https://github.com/bobmatnyc/trusty-tools/issues/4230) — this is a distinct latent defect, not a regression.

**Resolution direction:** replace the heuristic with an authoritative check (query launchd for ownership of the running PID rather than inferring from inherited environment). Not asserting a specific implementation as settled.

### LINKED ISSUE #4604: tm doctor skill_staleness misses a drifted skill — tm-workflow reported clean while three deployed copies are stale
DEFECT

`tm doctor`'s `skill_staleness` check reported `documentation-style` and `tm-pr-workflow` as drifted from their bundled sources, and did NOT report `tm-workflow` — which is drifted. A check that misses a drifted skill while flagging others is worse than no check, because it is read as an all-clear for the ones it does not name.

EVIDENCE (measured 2026-08-01 against installed `trusty-mpm 1.3.1` built from main @ `11ef5c27`, after PR #4583 merged as `2b5d4c9b`)

PR #4583 rewrote the bundled skill at `crates/trusty-mpm/src/assets/skills/tm-workflow.md` to remove a description of the four-flat-file instruction pipeline that #4183 deleted. The bundled source on main contains ZERO references to `PM_INSTRUCTIONS.md`.

All three deployed copies still contain that reference and none describes the JSON-manifest model:

| Path | mtime | contains PM_INSTRUCTIONS.md |
|---|---|---|
| `$CLAUDE_CONFIG_DIR/skills/tm-workflow/SKILL.md` | 07-03 | yes |
| `~/.claude/skills/tm-workflow/SKILL.md` | 07-17 | yes |
| `<project>/.claude/skills/tm-workflow/SKILL.md` | 07-29 | yes |

All three mtimes predate the merge. `tm doctor` run after the install reported `skill_staleness` naming only the other two skills.

Note there appears to be no `~/.trusty-mpm/framework/skills/tm-workflow/SKILL.md` on disk, which may be the cause — if the check compares a deployed copy against a CACHED source rather than the binary's embedded asset, a missing cache entry could make the skill unverifiable and be treated as clean rather than as unknown.

RESOLUTION

The check must compare each deployed skill against the running binary's EMBEDDED bundled asset, which is the actual source of truth, rather than against a cache that may be absent or itself stale. A skill it cannot verify must report UNKNOWN, never OK — silently treating unverifiable as clean is the specific defect.

Add: a second, related consequence worth covering in the same fix — the checksum-freeze rule (a hand-edited deployed skill is deliberately skipped on redeploy) means a frozen skill can be permanently stale with no signal. Doctor should distinguish 'drifted and will be fixed by redeploy' from 'drifted and FROZEN, redeploy will not touch it', because the remediation differs.

CROSS-REFERENCES (do not merge into them)

- #4583 (the skill fix that did not reach disk)
- #4589 (roster hides three deployed agents — same class of defect, a reported-clean state that is not clean)
- #4286

bobmatnyc:
Three defects, one theme: **trusty-mpm's self-diagnosis reported OK for things it had not actually verified.**

The generalizing rule this PR applies everywhere it touches: **UNVERIFIABLE MUST REPORT UNKNOWN, NEVER OK.**

---

## #4469 — `is_launchd_supervised()` let an unsupervised child self-report as supervised

`trusty_common::update::is_launchd_supervised()` returned `true` whenever `XPC_SERVICE_NAME` was set and `TERM_PROGRAM` was not, with a `getppid() == 1` fallback. Both prongs are satisfiable by a process launchd has never heard of — env vars are inherited by every child, and an orphan whose parent exited is reparented to PID 1. A `tm daemon` child could therefore pass the #4230 callee-side guard and publish `supervised: true` on `/health`, making the documented `curl /health | jq '.supervised'` restart verification unsound in exactly the case it exists to catch.

**Fix:** new `trusty_common::supervision` module asks launchd itself — it matches `std::process::id()` against the PID column of `launchctl list`. It answers in **three states**: `Supervised(label)`, `NotSupervised` (a positive answer from launchd), and `Unknown(reason)` when `launchctl` cannot be spawned, exits non-zero, or emits a table with no parseable rows. Deliberately an **exact PID match, never an ancestor walk** — `Terminal.app` is itself a launchd job, so walking the tree would reintroduce the same false positive. `is_launchd_supervised()` survives as the boolean adapter for the post-upgrade self-restart decision and maps `Unknown` to `false`, the conservative direction there.

### Mutation proof (live, this machine)

Same binary, one env var changed. The first prong of the deleted heuristic is reproduced verbatim alongside the new authoritative check:

```
=========== SCENARIO A: condition PRESENT (unsupervised child, XPC_SERVICE_NAME inherited, TERM_PROGRAM absent)
pid                          = 85076
XPC_SERVICE_NAME             = Some("com.trusty.mpm")
TERM_PROGRAM                 = None
OLD heuristic  -> supervised = true          <-- the #4469 defect
launchctl list rows parsed   = 543
NEW authoritative -> NotSupervised           <-- now correct

=========== SCENARIO B: condition ABSENT (interactive-terminal env)
OLD heuristic  -> supervised = false
NEW authoritative -> NotSupervised
```

And the positive case, against genuinely launchd-owned processes on this host:

```
--- LIVE launchd-owned trusty process:
pid 22772: Supervised(com.trusty.analyze)
--- this shell (NOT a launchd job):
pid 98139: NotSupervised
```

Regression tests: `supervision_rejects_a_child_that_inherited_xpc_service_name`, `is_launchd_supervised_is_not_fooled_by_inherited_env`, plus exhaustive parser coverage including `supervision_unknown_for_unrecognised_format` (an unrecognised `launchctl` format must be `Unknown`, not a confident answer in either direction).

---

## #4033 — no visibility into installed-binary provenance

doctor's health model was liveness plus file state: it answered "is it running?" but never "is what's running what you think it is?". On 2026-07-26 ten trusty-* binaries were installed with `cargo install --path` from git worktrees; trusty-search was a version behind two crash-safety fixes, tctl was seven patch versions stale, trusty-channels was installed twice from two different worktrees serving conflicting binaries, and one install source lived under `/Users/masa/tmp/`, which macOS reaps. Every check stayed green.

**Fix:** new `binary_provenance` doctor check (28 checks, was 27). It reads cargo's own `$CARGO_HOME/.crates2.json` install ledger and compares it against `std::env::current_exe()`. Read-only — it never installs, moves, or deletes.

| Condition | Verdict |
|---|---|
| Ledger unreadable, or does not cover the running binary | **Unknown** |
| Same binary provided by >1 install | **Fail** |
| Ledger's recorded version ≠ running version | **Fail** |
| `--path` install whose source directory is gone | **Fail** |
| Live path/git/unrecognised-source install | **Warn** |
| Registry install, versions agree | **Ok** |

### Mutation proof

Every row above is a test that constructs the broken condition and one that constructs its absence — `fails_on_duplicate_installs`, `fails_on_version_disagreement`, `fails_when_path_source_is_gone`, `warns_for_a_live_path_install` (the same ledger, only the source-exists probe flipped), `warns_for_a_git_install`, `unknown_without_a_ledger`, `unknown_when_bin_is_not_in_the_ledger`, and the clean control `ok_for_a_matching_registry_install`.

Live, against this machine's real cargo ledger — the test binary is genuinely not something cargo installed, so the check must say so rather than pass:

```
LIVE binary_provenance -> Unknown: `trusty_mpm-5c472738cc874721` is not recorded in cargo's
install ledger (running `…/target/debug/deps/trusty_mpm-5c472738cc874721`) — it was placed by
the prebuilt installer, a package manager, or built in place, so its provenance and upgrade
path cannot be verified from here (issue #4033)
```

---

## #4604 — the CORRECTED root cause, and how it was confirmed

The issue body's original hypothesis ("no `~/.trusty-mpm/framework/skills/tm-workflow/SKILL.md` on disk, so a missing cache entry was treated as clean") is **wrong**, and a later comment corrects it. The corrected diagnosis is the one fixed here:

> **`check_skill_staleness` trusted a local extraction cache that can itself lag the installed binary — so it was comparing a stale cache against itself.**

`check_skill_staleness` called `stale_skills(&paths.skill_source_dir(), &paths.claude_skills_dir())`. `skill_source_dir()` resolves, for every binary-only install, to `~/.trusty-mpm/framework/skills/` — an extraction cache — and `stale_skills` compared its content against the checksum recorded in the deployed target's manifest. When the cache had not refreshed since before #4583 merged, its sha256 (`f80ef6f4…`) was byte-identical to the checksum already in the deployed manifest from the 2026-07-29 deploy. Both sides were the same stale content, so the check correctly reported "no drift" against the wrong reference point. `documentation-style` and `tm-pr-workflow` were flagged only because their cache copies happened to differ for unrelated prior reasons.

**How I confirmed it:** I read `check_skill_staleness` → `stale_skills` → `FrameworkPaths::skill_source_dir()` and confirmed the chain never touches the binary's embedded assets — `stale_skills` never reads the deployed file either, only the manifest checksum. I then reconstructed the exact measured state in a hermetic test and ran the old comparison and the new audit against identical on-disk inputs (`the_old_cache_comparison_reports_clean_on_the_same_inputs`):

```
OLD (cache vs manifest)       -> []
NEW (deployed file vs binary) -> [SkillDriftFinding { stem: "tm-workflow", state: Drifted }]
```

That is the whole point of the issue, pinned as a permanent test: **the same inputs, the old check blind and the new check catching it.**

**Fix — three structural changes:**

1. **The reference is the running binary's own embedded asset** (`bundle::ALL`), never the extraction cache. A compiled-in table cannot lag the binary containing it. A checked-out, git-tracked `agents/skills` submodule is the one legitimate override, and the report names which reference it used.
2. **The comparison reads the DEPLOYED FILE**, not the manifest checksum. Reading the manifest cannot see a hand-edit, and it is the manifest's *agreement with the file* that separates the two drift states.
3. **It covers every deploy tier** — `$CLAUDE_CONFIG_DIR/skills`, `~/.claude/skills`, the project's `.claude/skills` — because #4604 was measured drifted at three tiers simultaneously.

**Severity table:** `Fail` for a drifted conventions-bearing skill; `Unknown` for anything unverifiable; `Warn` for ordinary drift, frozen drift, or a missing file; `Ok` only when every managed skill at every tier matches byte-for-byte. Drift that `tm install` repairs is reported **separately** from drift that is **FROZEN** (hand-edited, so the deployer's "checksum differs → user-modified → skip" rule means `tm install` will never touch it and it can stay stale forever with no signal).

### Mutation proof

One test per verdict, each constructing the broken condition and its absence: `staleness_catches_drift_the_stale_cache_hid` (Warn) against `staleness_ok_when_every_tier_matches` (Ok); `staleness_escalates_when_conventions_skill_drifts` (Fail) against `staleness_warns_when_ordinary_skill_drifts` (a conventions-bearing skill deployed and fresh alongside, proving the escalation does not leak); `staleness_reports_frozen_separately_from_repairable_drift` (asserts the message says `FROZEN` and does **not** claim `REPAIRABLE`); `staleness_is_unknown_when_a_skill_cannot_be_verified` and `staleness_is_unknown_with_no_reference_assets` (Unknown, never Ok); `staleness_covers_the_managed_config_tier`.

---

## Fix mode: LANDED

`tm doctor --fix-skills` redeploys drifted skills from the running binary's bundled assets. All three owner constraints are enforced in code, each with a test that fails if the rule is removed:

| Constraint | Enforcement | Test |
|---|---|---|
| Never silently overwrite a frozen (hand-edited) skill | `--include-frozen` is a **second** explicit opt-in on top of `--fix-skills`; clap rejects it standalone | `repair_never_touches_a_frozen_skill_by_default`, `cli_parses_doctor_fix_skills` |
| Never auto-delete worktrees | The repair has **no** worktree behaviour and removes nothing at all; `worktrees`/`worktree_disk` stay report-only | `repair_deletes_nothing` |
| Re-read from disk to verify its own work | `rewrite_and_verify` reads the file back and compares byte-for-byte; a write that did not land is `Failed`, never `Repaired` | `repair_reports_verification_failure`, `repair_rewrites_drifted_and_verifies` |

Every overwrite is backed up first under `~/.trusty-mpm/backup-doctor-remediation-<timestamp>/`, following the existing remediation-backup convention (`repair_backs_up_before_overwriting` asserts the operator's original content is recoverable from the backup).

**Deliberately NOT in scope, pending the owner review #4604 asked for:** the safe/unsafe split for the other non-OK checks. `output_style_staleness`, `push_guard`, and `legacy_sources` are plausibly auto-fixable but are not wired up here; `hooks_contamination` and anything touching `.claude/settings.json` need per-run confirmation; `gh_account`/`oauth_token` are credentials a human authenticates. Only `skill_staleness` — the check this PR fixed the detection for — got a repair path.

---

## Gates

```
cargo fmt --all --check                                   exit 0
cargo clippy --workspace --all-targets -- -D warnings      exit 0   (CI's own exclusions: the three Tauri GUI crates)
cargo test --workspace                                     19,254 passed / 92 ignored across 151 suites
```

Plus the repo gates: `check_line_cap.sh` (0 violations — `commands/misc.rs` was split into `commands/doctor_fix_skills.rs` and the doctor CLI flags were flattened into one `DoctorFlags` args struct to keep `main.rs` under the cap), `check_capabilities.sh` (regenerated — 28 checks), `check_changelog_fragment.sh` (both crates recorded), `check_sld.sh`, `check_test_pointers.sh`, `check_doc_numbers.sh`, `check_instruction_floor.sh`, `check_agent_assets.sh`, `check_buildrs_sync.sh` — all exit 0.

**One honest caveat.** Six `trusty-search` FSEvents watcher tests (`service::watcher::*`, `service::watch_loop::*`) fail on this machine, with `event arrives before deadline: Elapsed(())`. They are **not** caused by this PR and are not a flake I am burying:

- this PR changes **zero** files under `crates/trusty-search` (`git diff --stat origin/main -- crates/trusty-search` is empty);
- `trusty-search` contains **zero** references to the changed API (`is_launchd_supervised`, `supervision::`);
- they fail identically with `cargo test -p trusty-search --lib` in isolation, so this is not full-workspace load either — it is macOS FSEvents timing on a heavily-loaded dev box, not a regression.

The two crates this PR actually touches are fully green: `trusty-common` 237 passed (269 with `--features update-check`), `trusty-mpm` 7,274 passed / 0 failed across its full test surface.

---

Closes #4469
Closes #4033
Closes #4604


bobmatnyc:
## Adversarial review — Verdict: WARN

Reviewed at `576206e0` in an isolated read-only worktree. No edits made. All 23 CI checks green; I re-ran locally: `trusty-mpm --lib` 4253 passed / 0 failed, `--bin tm` 1374 passed / 0 failed, `trusty-common --features update-check --lib` 269 passed / 0 failed. The three new suites (`skill_drift` 22, `skill_repair` 9, `binary_provenance` 19) all pass and the mutation proofs are genuine, not tautological — `the_old_cache_comparison_reports_clean_on_the_same_inputs` really does run the deleted comparison and the new one over identical on-disk inputs.

Zero CRITICAL. Two HIGH. Both are the PR's own rule (`UNVERIFIABLE MUST REPORT UNKNOWN, NEVER OK`) failing in the opposite direction from the one it was aimed at.

### Findings

|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/skill_drift.rs` | 169, 225–230 | `skill_reference` drops every `skills/<stem>/references/*.md` entry of `bundle::ALL` (`if rel.contains('/') { return None }`), but `audit_deployed_skills` iterates **all** deploy-manifest keys — and `skills::deployer::deploy_skills` writes reference files into that manifest keyed `<stem>/references/<file>.md`. Every one of those keys misses `reference.assets` and classifies `Unverifiable`. `skill_staleness` therefore reports **Unknown on every real installation, permanently**, and the documented `Ok` state is unreachable. | Key nested assets by their manifest-key form (`<stem>/references/<file>.md`, i.e. `rel` unstripped) in `skill_reference`, and in `classify` resolve a key containing `/` to `dest_dir.join(key)` instead of `dest_dir/<stem>/SKILL.md`. This also closes the matching gap in `skill_repair`, which today rewrites only `SKILL.md` for a multi-file skill. |
| HIGH | `crates/trusty-mpm/src/core/binary_provenance.rs` | 201–204, 234, 251–258 | The check matches the ledger on **binary basename only**. `running_exe` is used solely to build `where_from` (l.186–189) and is never compared to `$CARGO_HOME/bin/<name>`. A `tm` on `PATH` that cargo did not place (prebuilt installer, package manager) whose version happens to equal the ledger's gets `Ok` — "installed from crates.io; cargo's ledger agrees with the running version" — a confident provenance claim about a file the check never looked at. This is #4033's own scenario. | Compare `running_exe` (canonicalized) against `$CARGO_HOME/bin/<bin_name>`; when they differ, report `Unknown` (provenance not attributable) or `Fail` (two installers provide the same binary). Pass the cargo bin dir into `provenance_report` so it stays pure. |
| MEDIUM | `crates/trusty-mpm/src/core/skill_drift.rs` | 205 | `SkillManifest::load` returns an **empty default on a missing *or unparseable* manifest**. A tier whose manifest is absent or corrupt yields zero findings, and `report` then emits `Ok`: "every deployed skill matches … at every deploy tier" — a positive claim over zero examined files. | Distinguish the two: return `Unverifiable("manifest at <dir> could not be parsed")` when the file exists but does not deserialize, and `Unverifiable` when the tier directory contains skill dirs but no manifest. |
| MEDIUM | `crates/trusty-common/src/update/upgrade.rs` | 423 | The doc asserts "Callers that must DISTINGUISH 'not supervised' from 'could not tell' (the daemon's `/health` flag, `tm doctor`) call `launchd_supervision` directly." Neither does. `grep` finds exactly one production consumer of the module — `daemon_run.rs:229`, via the **bool** adapter. `LaunchdSupervision` is otherwise unused outside its own module, so `/health` still collapses `Unknown` into `supervised: false`, which is precisely the conflation the module doc says must never happen. | Either wire `/health` (and `doctor_orphan`) to the three-state value, or delete the claim. |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/launchd_probe.rs`, `commands/doctor_orphan.rs`, `src/daemon/api/types.rs` | 237–241, 18–21, 90–92 | Three sites still document the **deleted** heuristic as live rationale ("has two prongs: an `XPC_SERVICE_NAME` env probe AND a `getppid() == 1` fallback", "a heuristic with a `getppid() == 1` fallback prong"). These are the load-bearing arguments for the #4230/#4397 designs; a future reader will reason from a mechanism that no longer exists. | Update the three doc comments to describe the authoritative PID-match check. |
| MEDIUM | `crates/trusty-common/src/supervision.rs` | 266–274 | `supervision_rejects_a_child_that_inherited_xpc_service_name` mutates `XPC_SERVICE_NAME`/`TERM_PROGRAM` **without `ENV_LOCK`**, while `update/tests.rs:807,838` mutate the same two vars under that lock in the same test binary. That is a data race (`set_var` is `unsafe` in edition 2024 for exactly this reason) and it buys nothing: the assertion only calls the pure parser `supervision_from_launchctl_list`. | Delete the `unsafe` env block; the test is stronger without it (it demonstrates the new path never consults the environment at all). |
| MEDIUM | `crates/trusty-common/src/supervision.rs` | 110 | `Command::new("launchctl").arg("list").output()` has **no timeout**. It runs on the daemon-start path (`daemon_run.rs:229 → apply_supervision_signal`) and on the installer's `upgrade_and_restart` path. The module handles spawn failure and non-zero exit as `Unknown` but not a hang; a wedged launchd — the state where this diagnosis matters most — blocks daemon startup indefinitely with no message. Previously this was an env read. | Bound it (spawn + wait-with-timeout, or `launchctl list <label>` which returns a much smaller table); on timeout return `Unknown("`launchctl list` did not respond within Ns")`. |
| LOW | `crates/trusty-common/src/supervision.rs` | 152–181 | `parsed_any_row` is set by **any** line with three tab-separated fields and a non-empty third. A future/unrecognised tab-delimited format therefore reads `NotSupervised` (a positive answer) rather than `Unknown`. The dedicated test only covers a JSON-shaped output, which has no tabs. | Require the PID field to be `-` or numeric before counting the row as parsed. |

### Answers to the review questions

1. **#4469 caller coherence — clean.** Two production callers: `upgrade.rs:469` (`upgrade_and_restart`) and `daemon_run.rs:229` (`apply_supervision_signal` → `/health` + the #4397 start guard). Both go through `is_launchd_supervised()`, which maps `Unknown → false`. **No caller treats `Unknown` as `Supervised`**; the old defect is not reintroduced. The exact-PID-match reasoning is sound — `Terminal.app` is a launchd job, so an ancestor walk would make every interactive process supervised. Non-macOS returns `NotSupervised`, which is defensible (launchd's absence is a known platform fact, not an unanswered question), though note it also means a systemd-supervised Linux daemon reports unsupervised — out of scope here, but the name `is_launchd_supervised` should not be reused for a general supervision question. One behavioural change worth flagging to the owner: `Unknown` now maps to `false`, and `compute_supervised(false, plist_exists=true)` makes `run_daemon` **refuse to start** without `--force`. That is the safe direction, but combined with the un-timeouted subprocess (MEDIUM above) it is a new way for the daemon to fail to boot.
2. **Fix mode — all three constraints verified in code.** (a) `skill_repair.rs:126` matches `DriftedFrozen if !include_frozen` **before** the rewrite arm, and `cli/mod.rs:1167` is `#[arg(long, requires = "fix_skills")]`, so clap rejects `--include-frozen` standalone. I could not construct a bypass. (b) `grep` for `remove_file|remove_dir|delete|truncate` across `skill_repair.rs`, `skill_drift.rs`, `doctor_fix_skills.rs` returns **only doc-comment prose** — the sole `remove_file` in the write path is `atomic_write`'s cleanup of its own `.tmp` file. Nothing touches worktrees. (c) `rewrite_and_verify` (l.232–239) does a fresh `read_to_string` and a byte-for-byte compare; a mismatch returns `Err` → `Failed`. **I could not construct a `Repaired` without a real change**: `Repaired` requires `observed == content`, and a file already equal to `content` would have classified `Fresh` and been `continue`d before any write. A failed manifest save afterwards is reported as a separate `Failed` outcome and does not fabricate a repair.
3. **Backups — correct.** `rewrite_and_verify` copies the existing file **before** `atomic_write` (l.202–221), and both `create_dir_all` and `fs::copy` use `?`, so a backup failure aborts and the original is untouched. `repair_reports_verification_failure` additionally asserts the original content survives a denied write.
4. **#4604 — yes, subject to HIGH-1.** The reference is `bundle::ALL`, which cannot lag the binary containing it, and the comparison reads the deployed file, so the drifted `tm-workflow` is caught. I could not reproduce the original live drift (all four copies now hash `307bb750…` on this host), but the hermetic proof is sound. **False positives:** an intentionally customized skill is reported forever as `Warn`/`FROZEN` with no acknowledge or ignore mechanism — by design per the owner's separation request, but it will be permanent noise. The much bigger false-positive source is HIGH-1: 80 of 132 manifest keys per tier on this host are `references/` keys, all of which currently report "could NOT be verified".
5. **Ok-when-unverifiable paths still open:** binary_provenance basename match (HIGH-2) and the absent/corrupt skill manifest (MEDIUM-3). Everything else I traced routes unverifiable states to `Unknown` correctly.
6. **Deferral is clean.** The repair surface is `--fix-skills` alone; `--include-frozen` is gated on it; `skill_repair` has no generic dispatch and no entry point for any other check. There is no half-built path an unsupported check could reach.

1. Fix HIGH-1 — nested `references/*.md` manifest keys must resolve against the bundled asset, not fall through to `Unverifiable`. Add at least one test that builds its fixture with the **real** `deploy_skills` rather than the local `deploy()` helper in `skill_drift_tests.rs:30`, which only ever writes bare-stem keys — that helper is why the suite is green while the production path is not.
2. Fix HIGH-2 — verify the running executable's identity against the cargo bin dir before attributing a ledger record to it.

### Notes

- Evidence for HIGH-1 is from the owner's live machine, not inference: both `~/.claude/skills/.trusty-mpm-skills-manifest.json` and `~/.trusty-tools/trusty-mpm/claude-config/skills/.trusty-mpm-skills-manifest.json` carry 132 managed keys, 80 of them `<stem>/references/<file>.md`.
- Evidence for HIGH-2 likewise: `which -a tm` on this host returns **both** `~/.cargo/bin/tm` (1.3.1) and `~/.local/bin/tm` (0.19.29), while the ledger records one `trusty-mpm 1.3.1` path install providing `tm`. The check keys on basename alone across that pair.
- `tm doctor --fix-skills` exits 0 even when repairs report `Failed`. Consistent with `tm doctor`'s existing exit-code behaviour, so not filed as a finding — but a remediation command whose failures are invisible to a script is worth a follow-up.
- The MEDIUM findings are all cheap; folding them into this PR is preferable to filing them.

Issue #4179, opened 2026-07-27
The merge gate does not gate: enforce_admins off, cancelled CI reports success, and a required check that cannot fail


bobmatnyc:
Three holes verified live against the GitHub API on 2026-07-27. Confirmed, not theorized.

**1. `enforce_admins: false` on `main`.** `--admin` waives all 6 required status checks, not just bot approval. No structural backstop against merging on red, on pending, or on zero checks.

**2. `ci.yml:1050` — `if: contains(needs.*.result, 'failure')`.** `cancelled` is not `failure`, so cancelled jobs skip "Notify on red main" and the job reports success. On merge commit `ea49e0d1` every build/test job was cancelled and the alarm read healthy. All 16 workflows set `cancel-in-progress: true`, and on `push` the concurrency key is `refs/heads/main` — so each merge to main cancels the previous merge's post-merge CI. 10 such cancellations on 2026-07-27 alone.

**3. A required check that cannot fail.** `trusty-search daemon smoke test` is required but carries `continue-on-error: true` (`ci.yml:919`); its own comment concedes the result "never surfaces as failure." Separately 17 of 23 checks are advisory — a live `tm-capabilities drift check` failure blocked nothing.

## Evidence

- **PR #3838** (`217acd78f4`) merged with zero check runs, ever (`statusCheckRollup: []`). Stacked PR: `head_ref_force_pushed` fired while the base was still a feature branch (filtered by `branches: [main]`), then `base_ref_changed` retargeted it to main — and `base_ref_changed` is not a default `pull_request` activity type, so nothing re-triggered. Its post-merge main CI was then cancelled by the next merge, so whether it broke main is now unknowable.
- **PR #3827** (`ff913c7223`) merged 9 minutes before its required `Test` check completed. `Test` concluded `failure`.

## Resolution direction

Candidates: `enforce_admins: true`; treat `cancelled`/`skipped` as not-success in the notify condition; either drop `continue-on-error` from the required smoke test or stop marking it required.

Notes for whoever picks this up: a push after retargeting is the known recovery for a zero-check stacked PR (proven on #4155), and `gh pr checks` exits 1 on an empty check list — so a properly `&&`-gated merge does catch that case.

Related shape: #3675 (failure indistinguishable from success).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

Issue #3366, opened 2026-07-20
release: sibling crates carry unpublished src drift under already-published version numbers — blocks trusty-mpm publish


bobmatnyc:
## Symptom

`cargo publish -p trusty-mpm --dry-run` fails with 6 "symbol not found" errors (`trusty_common::server::SelfOrigins`, `trusty_agents_common::connectors`), blocking the 0.19.26 release.

## Root Cause

Features were merged into `trusty-common` and `trusty-agents-common` WITHOUT bumping their version fields.

- **trusty-common 0.23.3**: repo code is DIFFERENT from crates.io 0.23.3 under the same version number (4 commits of drift, incl. `aa7fb052` #3308 adding `SelfOrigins`)
- **trusty-agents-common 0.2.2**: repo code differs by 5 commits (incl. `6890b624` #3194 adding `connectors`)

## Why It Hid

The workspace `[patch.crates-io]` table path-redirects siblings for local builds, so:
- `cargo build` ✓ passes
- `cargo test` ✓ passes
- `tm --version` ✓ works

The gap only appears when `cargo publish` packages a crate standalone and resolves siblings from the real registry. `trusty-agents-common` is not even covered by the patch table.

## Correctness Hazard

Any external consumer installing `trusty-common 0.23.3` from crates.io gets DIFFERENT code than the repo tags as 0.23.3 — a silent correctness violation.

## Proposed CI Guard

Detect, for each publishable workspace crate, whether `src/` has commits after the crate's most recent published-version tag while the Cargo.toml version still equals the version already live on crates.io — i.e. "unpublished drift under a published version number" — and fail the build (or warn loudly) so this is caught at merge time rather than at release time.

Frame this as a proposal for discussion. The check should probably run on merge to main, not on every PR to avoid noise. It would prevent both the immediate 0.19.26 release jam and future silent version mismatches.

PR #4624, opened 2026-08-02
fix(ci): stop cancelled runs reading as green, give stacked PRs real checks, and gate docs-only PRs (#4468, #4421)

### LINKED ISSUE #4421: Version-parity drift is only caught AFTER merge — add a merge-time guard
## THE GAP

The "Version parity" workflow (`.github/workflows/version-parity.yml` → `scripts/check-version-parity.sh` → the `trusty-publish-guard` binary) triggers on `push: [main]` ONLY. It has no PR trigger, deliberately — it is meaningful against what actually landed. The consequence is that a crate can drift through PR review with every check green, and main only goes red AFTER the merge. Nobody is blocked; main just quietly carries a red workflow.

## WHAT THE CHECK DOES (for context)

For every publishable crate it diffs the local `src/` tree against the crates.io tarball for that crate's own current `Cargo.toml` version. 
- Not-yet-published → PASS
- Source matches → PASS
- Source differs under an already-published version → FAIL
- Registry unreachable → FAIL (fail-closed)

## EVIDENCE OF RECURRENCE

This happened **TWICE on 2026-07-30 alone**:

1. `trusty-installer 0.4.10` drifted at commit 6078b21c (~04:28 UTC). 9 consecutive red runs. Fixed by PR #4412 (commit 095cce8f) bumping 0.4.10 → 0.5.0 (minor, because the drift added a new public module `commands::probe_http` with 11 public items). Tracked as issue #4403, closed 17:30Z.

2. `trusty-agents-common 0.3.0` drifted at commit e8f30ef0 (PR #4419, "redeploy corrupted bundled assets"). Main went red again within hours of the first fix landing.

## PRIOR ART, AND WHY THIS IS NOT A DUPLICATE

Issue #3366 tracked the **ORIGINAL instance** of this defect class ("sibling crates carry unpublished src drift under already-published version numbers"), which blocked the 0.19.26 release of trusty-mpm. **#3366 is CLOSED — but the merge-time guard it proposed was NEVER IMPLEMENTED.** Closing that issue did not close the hole, which is why the defect keeps recurring.

## WHAT IS WANTED

A guard that catches drift at **MERGE time** rather than after. Present these as **options for the owner** (do NOT choose one):

- **Run the parity check on PRs against the PR's merge-base result**, accepting that it can produce false signals when several crates land in sequence.
- **A lighter PR-time check**: if a PR touches `crates/<X>/src/**` and does NOT bump `crates/<X>/Cargo.toml` version, and crate X's current version is already published, warn or fail.
- **Keep it post-merge but make it LOUD** — a notification or auto-filed issue when main's parity goes red, so it does not sit unnoticed.

## COST OF STATUS QUO

Main carries a red workflow for hours at a time, which erodes the signal value of main's CI status and trains people to ignore it.

Relates to: #3366, #4403, #4412, #4419

### LINKED ISSUE #4468: ci: docs-only PRs run the full Rust build (Clippy/Format/MSRV/Test) — add path filtering
## Defect

Docs-only PRs run the full Rust build — Clippy, Format check, MSRV, and Test — even though nothing in the diff can affect compilation. Reported by the repo owner: "I'm seeing clippy/compile/smoke tests for docs — a waste of time."

## Evidence

Of 15 workflows in `.github/workflows/`, only 3 have any `paths` filter:

- **Has a filter:** `agent-assets.yml`, `al2023-build.yml`, `e2e-docker.yml`
- **No filter:** `ci.yml` (the expensive one — Clippy, Format, MSRV, Test), `capabilities-drift.yml`, `cargo-audit.yml`, `claude-md-guard.yml`, `generation-artifact-lint.yml`, `instruction-floor-guard.yml`, `line-cap.yml`, `release.yml`, `sld-lint.yml`, `test-pointers.yml`, `token-drift.yml`, `version-parity.yml`

## Implementation trap

A naive `paths-ignore` on the heavy `ci.yml` jobs is unsafe if any of them are configured as **required status checks** in branch protection. A required check that never reports leaves the PR permanently unmergeable rather than mergeable-and-fast.

Whoever implements this must:

1. First determine which checks branch protection actually requires (`gh api repos/bobmatnyc/trusty-tools/branches/main/protection` or the equivalent ruleset endpoint).
2. If heavy jobs are required, use a path-filter job plus a final always-runs "gate" job that branch protection points at — NOT bare `paths-ignore` on the required jobs. A job skipped by an `if:` condition does not satisfy a required status check.
3. Decide per workflow, not globally. Some unfiltered workflows SHOULD keep running on docs changes — `sld-lint`, `claude-md-guard`, and `instruction-floor-guard` all appear to be docs/instruction-focused gates, so filtering them to code paths would defeat their purpose. Do not blanket-apply one filter to all 15.

## Sequencing note

Land this when no merge train is in flight. Editing CI config while PRs are queued risks making in-flight PRs unmergeable.

## Resolution direction

Add path filtering to `ci.yml` (and any other workflow where it's clearly correct) informed by the branch-protection required-check list above, using the path-filter-plus-gate pattern where required checks are involved. The exact set of filtered paths and which workflows to touch is an implementation decision, not settled here.

bobmatnyc:
Fixes the workflow layer of #4179 (the parts that do not need a branch-protection change), plus #4468 and #4421. Five changes, one review.

**`enforce_admins` was deliberately left OFF, per owner ruling.** No branch-protection setting was touched by this PR — not via `gh api`, not otherwise. #4179 therefore stays OPEN with its first root cause intact by design; this PR closes only the downstream holes.

---

## 1. Cancelled runs read as green (#4179)

**Root cause, two halves.** Every workflow used `cancel-in-progress: true` on a group keyed to `github.ref`. On a `push` that key is `refs/heads/main` for every merge, so each merge cancelled the previous merge's verification — 19 of 60 sampled push-to-main runs (32%) concluded `cancelled`. The notifier then asked `contains(needs.*.result, 'failure')`, and `cancelled` is not `failure`, so it reported success. Live instance: run [30720348101](https://github.com/bobmatnyc/trusty-tools/actions/runs/30720348101), head `b4b91af9` — Test, Clippy, MSRV and the smoke test all cancelled, notifier green. (Still visible today: `gh run list --branch=main` shows `b4b91af9 cancelled` sandwiched between two successes.)

### Concurrency decision — cancel PR runs, never cancel main runs

| Event | Group | `cancel-in-progress` | Why |
|---|---|---|---|
| `pull_request` | per PR ref (`github.ref`) | `true` | A new push makes the old run obsolete. This is the case the setting was added for, and it keeps working. |
| `push` to main | per **COMMIT** (`github.sha`) | `false` | The commit is already on main. There is nothing to supersede, and a cancelled run is the only signal that commit will ever get. |

Implemented as `group: ci-${{ github.event_name == 'push' && github.sha || github.ref }}` with `cancel-in-progress: ${{ github.event_name != 'push' }}`.

Keying push runs to the SHA rather than just setting `cancel-in-progress: false` on a shared group matters: a shared group would serialise a merge train behind itself. Per-SHA groups are unique, so main runs neither cancel nor queue.

Applied to all 12 workflows carrying the `${{ github.ref }}` + `cancel-in-progress: true` pattern on a `push: [main]` trigger. `cargo-audit` and `e2e-docker` (schedule/dispatch only) and `release` (tag refs, already unique) were left alone.

### Notifier decision — not-success is not success

The verdict moved out of a GitHub expression and into `scripts/classify-ci-results.sh`, so the mapping is executable and fixture-tested rather than buried where it cannot be exercised. The notify job now also **fails the run** on any non-green verdict, so the run conclusion stops disagreeing with what the jobs did. `search-daemon-smoke` joins its `needs` — it could not be there while `continue-on-error` kept its result off `failure`.

**Conclusion-mapping table** (asserted by `scripts/check-ci-helpers-selftest.sh`):

| input conclusion | class | reported verdict |
|---|---|---|
| `success` | pass | `green` |
| `failure` | red | `red` |
| `cancelled` | inconclusive | `inconclusive` |
| `skipped` | inconclusive | `inconclusive` |
| `timed_out` | red | `red` |
| anything else (e.g. `neutral`) | inconclusive | `inconclusive` |
| no inputs at all | — | `inconclusive` (fail closed) |

Precedence, also asserted:

| input set | verdict |
|---|---|
| `a=success b=success c=success` | `green` |
| `a=success b=cancelled` | `inconclusive` |
| `a=success b=skipped` | `inconclusive` |
| `a=cancelled b=failure` | `red` (failure outranks cancelled) |
| `a=skipped b=failure` | `red` |
| `fmt=success clippy=cancelled test=cancelled msrv=cancelled smoke=cancelled` — the shape of live run 30720348101 | `inconclusive` (**was: green**) |

`timed_out` is accepted even though GitHub's `needs.<job>.result` collapses a timed-out job into `failure` and never emits the literal string. The same mapping is then also correct if it is ever fed raw check-run conclusions, where `timed_out` is a real value.

**Layering, stated honestly:** the concurrency change is the primary remedy. A run cancelled mid-flight takes the notify job with it, so no notifier can report from inside a cancelled run. The verdict is the second layer, for the cancellations concurrency cannot prevent — a manual cancel, a runner eviction, a `skipped` job.

---

## 2. Stacked PRs got zero checks (#4179)

No workflow declared `types:`, so `pull_request` used GitHub's default set — `opened`, `synchronize`, `reopened`. **Retargeting a PR's base is delivered as `edited` (with `changes.base`), which is not in that set.** A PR stacked on another feature branch is filtered out by `branches: [main]` while it is stacked, and then nothing re-triggers when it is retargeted to main. PR #3838 merged with `statusCheckRollup: []` — zero check runs, ever — exactly this way.

Every `pull_request` trigger now declares:

```yaml
types: [opened, synchronize, reopened, ready_for_review, edited]
```

### Trigger matrix for a PR whose base is NOT main

| moment | event delivered | before | after |
|---|---|---|---|
| PR opened against feature branch `X` | `opened`, base=`X` | filtered by `branches: [main]` — no run | filtered by `branches: [main]` — no run |
| push to the PR head while still on `X` | `synchronize`, base=`X` | filtered — no run | filtered — no run |
| **base retargeted `X` → `main`** | **`edited`** (`changes.base`), base=`main` | **not in the default types — NOTHING fires** | **`edited` is declared → full run, base now matches `main`** |
| draft marked ready | `ready_for_review` | not in default types — no run | run |
| any later push | `synchronize`, base=`main` | run | run |

So, explicitly: for a base-branch-is-not-main PR, **no** event fires while it is stacked (correct — `branches: [main]` still applies), and then **`edited`** fires at the moment of retarget. That is the event that previously fired nothing.

**Accepted cost, corrected after review.** An earlier draft of this body said `edited` "costs a duplicate run". That understated it. `github.ref` is `refs/pull/N/merge` for both `synchronize` and `edited`, so with a shared PR concurrency group an `edited` event would have **cancelled the in-flight run and restarted from zero** — measured against this PR's own timings (`Test` 15m49s, smoke 5m58s) that is up to ~16 minutes of completed work discarded, and under repeated body edits (routine here — agents rewrite PR descriptions) the checks could be kept from ever settling.

Fixed rather than merely documented: `cancel-in-progress` now excludes `edited`.

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}
```

An edit-triggered run now **leaves the in-flight one alone** instead of killing it, and it still reaches a conclusion. Costs one extra run per description edit; throws away no completed work. With that, the trade is clearly worth it: a redundant run versus a zero-check merge.

**Observed live on this PR.** Editing this description twice produced runs `30768284687` and `30768304988` on the same head SHA, and the already-running `30768284687` went on to complete `success` (16 min, full `cargo test`) rather than being cancelled. Under the old `cancel-in-progress: true` it would have been killed at the first edit.

The tempting alternative — a job-level `if:` skipping edit-triggered runs — stays rejected. A skipped job publishes a fresh `skipped` check run on the same SHA, overwriting a previously red result. That is signal laundering, the same class of bug this PR exists to close.

---

## 3. The smoke test contradiction (#4179)

`trusty-search daemon smoke test` was simultaneously (a) listed among branch protection's seven required contexts, (b) documented in its own header as OPTIONAL, and (c) carrying `continue-on-error: true`.

**Decision: it is REQUIRED, and it must be able to fail. `continue-on-error: true` is removed.**

- The "required" half is the one enforced by branch protection, and this PR may not change branch protection. Making it optional was not an available option, only a wish.
- The flag was never providing the protection its header claimed. Per the repo owner's own correction on #4179 (settled empirically against `release.yml`'s `best_effort` legs), job-level `continue-on-error` still publishes `conclusion: failure` to the **check run** — so it already gated the merge button. All it added was **signal inversion**: the Actions UI, `gh run view`, and `needs.<job>.result` all read healthy while a required check was red.
- It costs nothing in wall clock. The ~15–20 min cold release build ran either way; only the reporting changes.
- It buys a place in `notify-main-failure`'s `needs`, which was impossible while the job's result never surfaced as `failure`.

If the runner timeout/kill race the old header worried about ever materialises, the answer is to fix the race or drop the check from branch protection — not to keep a required check unable to report the truth.

---

## 4. Docs-only PRs ran the full Rust build (#4468)

**The constraint drove the design.** Five of `ci.yml`'s checks are required (`Format check`, `Clippy`, `Test`, `MSRV check (1.91)`, `trusty-search daemon smoke test`). GitHub does not create check runs for a workflow its path filters skipped, so a bare `paths-ignore:` would leave a required context pending forever and make docs-only PRs **unmergeable** rather than fast.

A job-level `if:` was also rejected. GitHub documents a conditionally-skipped job as reporting success to branch protection, but that cannot be verified from here and the failure mode if it is wrong is *every docs PR blocked forever*. So:

**Every job still RUNS and still REPORTS. Only the expensive STEPS are gated.** This is the pattern `changelog-fragment.yml` already documents in its own header — the exemption lives in the script, not in the trigger. A docs-only PR gets a green required check in ~30 s of runner startup instead of 45 minutes of cargo, with zero reliance on skip semantics.

A new `changes` job runs `scripts/detect-docs-only.sh` and publishes one output; the heavy jobs consume it as `needs.changes.outputs.docs_only != 'true'` on each costly step.

### The `needs:` edge, closed after review

`needs: [changes]` alone reintroduced the exact bug this PR exists to close, in the worse direction. If `changes` went red, all five required contexts depending on it would become **`skipped`** — and GitHub reports a skipped required check as **passing**. Worse, `changes` is what runs the helper selftest, so the designed alarm for a helper regression would also have disarmed five gates.

Every one of the eight dependent jobs now carries:

```yaml
if: ${{ !cancelled() }}
```

This overrides the implicit `success()` on `needs`, so the job runs even when `changes` fails. `needs.changes.outputs.docs_only` then resolves to `''`, `'' != 'true'` is true, and every gated step runs — a broken or unresolvable classifier costs a **full build**, never a silent skip. Proven live rather than reasoned about; see Verification below.

### Path patterns

Classification is a **denylist of provably-inert paths**, not an allowlist of code: anything unrecognised is treated as code. An unknown new path costs a full build (correct, cheap to fix) rather than silently skipping verification. Matching is on **split segments**, not globs — in a bash `case`, `*` matches `/`, so `crates/*/changelog.d/*.md` would otherwise also accept `crates/a/src/assets/changelog.d/x.md`.

| inert (docs-only) | NOT inert |
|---|---|
| `docs/**` (any depth) | `crates/*/src/**` — **including `.md`** |
| repo-root `*.md` | `crates/*/tests/**`, `crates/*/ui/**` |
| `LICENSE`, `LICENSE-*` | `Cargo.toml`, `Cargo.lock`, any `*.toml` |
| `.github/*.md` | `.github/workflows/**` |
| `.github/ISSUE_TEMPLATE/**` | `scripts/**` |
| `crates/*/README.md`, `crates/*/CHANGELOG.md` | `crates/*/changelog.d/*/**` (nested) |
| `crates/*/changelog.d/*.md` | anything else |

`crates/*/src/**/*.md` is the trap this list exists to avoid: those are bundled agent/skill/instruction assets compiled in via `include_dir!`/`include_str!`. Editing one changes program output and breaks asset-pin and drift tests, so it is code.

### Three concrete change sets

| # | change set | `docs_only` | what RUNS | what REPORTS |
|---|---|---|---|---|
| 1 | `docs/a.md`, `README.md`, `crates/trusty-mpm/changelog.d/1-x.md` | `true` | `changes` (+ helper selftests); `fmt`'s instruction-floor-guard step; `line-cap`; `instruction-floor-guard`; `changelog-fragment`; the docs linters. **No cargo, no pnpm, no release build.** | All 7 required contexts report **success**: `Format check`, `Clippy`, `Test`, `MSRV check (1.91)`, `trusty-search daemon smoke test`, `500-line file-size cap`, `BASE_PM floor carries the guaranteed conventions` |
| 2 | `crates/trusty-mpm/src/lib.rs` | `false` | everything, unchanged | all 7 required contexts, real results |
| 3 | mixed: `docs/a.md` + `crates/trusty-mpm/src/lib.rs` | `false` | everything, unchanged | all 7 required contexts, real results |

Verified against this branch itself — 19 changed paths, all workflows + scripts → `docs_only=false`, i.e. **this PR runs the real jobs**, which is also the live proof of cases 2 and 3.

`fmt`'s instruction-floor-guard step is deliberately **not** gated: it guards instruction/markdown content, which is exactly what a docs-only PR changes. `line-cap`, `instruction-floor-guard`, `sld-lint`, `claude-md-guard`, `doc-numbers`, `test-pointers`, `generation-artifact-lint`, `token-drift`, `capabilities-drift` and `changelog-fragment` are untouched by the filter, per #4468's "decide per workflow, not globally."

**Push to main never takes the exemption** — main is what everything is cut from, so it is verified in full regardless of the diff.

---

## 5. Version-parity drift only caught after merge (#4421)

#4421 offered three options. **Option 2 — the lighter PR-time check — is implemented**, as `scripts/check-pr-version-bump.sh` in a new `pull_request` job. The post-merge tarball diff is unchanged and remains the fail-closed authority.

| option | verdict |
|---|---|
| 1. Full parity check on PRs | **Rejected.** Downloads and diffs the published tarball for every publishable crate, so a PR touching nothing publishable pays the whole cost; fails CLOSED on registry trouble, so one crates.io hiccup blocks every open PR; and mis-fires on a merge train, where crate N's parity depends on crate N−1 having landed. |
| **2. Bump-vs-published check** | **Chosen.** Exactly the shape of both 2026-07-30 recurrences. One sparse-index GET per touched crate, no tarballs, no cargo build — seconds, not minutes. Decidable from the PR's own diff rather than from a moving main. |
| 3. Keep post-merge, make it loud | **Rejected as the primary remedy** — main still goes red, which is the cost #4421 is about. The "louder" half landed anyway via the #4179 notifier work above. |

Rule, per crate whose `crates/<X>/src/**` changed:

| situation | result |
|---|---|
| crate deleted by the PR | skip |
| `publish = false` | skip |
| version bumped vs merge base | pass |
| version unchanged, not on crates.io | pass |
| **version unchanged, already live on crates.io** | **fail** |
| registry unreachable | warn, do **not** fail |

The unreachable case deliberately does not fail: the post-merge gate is still behind it, so an outage costs late detection (today's status quo) rather than blocking every PR in the repo.

**Also fixed here:** `version-parity.yml`'s job `name:` was unquoted, so ` #3366)` opened a YAML comment and the job registered as `Publish-version parity (issue`. Both job names in that file are now quoted. Not a required context, so nothing was blocked — but promoting one to required by copying the `name:` line out of the YAML would have produced a context that never reports and blocks every PR forever.

---

### Ran, with output

- **`actionlint` 1.7.12 on all 14 changed workflow files — clean, exit 0, no output** (re-run after the rebase onto `b8e03184`). Confirmed non-vacuous by temporarily corrupting a `needs.` reference: actionlint reported `property "nosuchjob" is not defined in object type {changes: {outputs: {docs_only: string}; result: string}}`, then the file was restored and re-linted clean. (One pre-existing SC2129 style note in the untouched `e2e-docker.yml` is unrelated.)
- **`scripts/check-ci-helpers-selftest.sh` — all 42 cases passed**: 16 conclusion-mapping cases, 21 path-classification cases, 5 version-bump decision branches. Three red-precedence cases were added after review found the guard survived mutation because every existing fixture put `failure` last; the new cases were confirmed to kill that mutant (3/42 fail on the mutated script, 42/42 on the real one). Production behaviour was already correct — this was a test gap, not a live defect. Wired into the always-running `changes` job, so a regression in any of the three helpers fails CI.
- **Required-context audit** — parsed every workflow with `yaml.safe_load` and confirmed all 7 branch-protection contexts are still declared under byte-identical names.
- **`scripts/detect-docs-only.sh` on this branch** — 19 changed paths → `docs_only=false`.
- **`scripts/check-pr-version-bump.sh` against LIVE crates.io**, on a throwaway commit touching every crate's `src/`: correctly failed the 4 crates whose current version is already published (`trusty-bm25-daemon 0.1.3`, `trusty-embedderd 0.3.10`, `trusty-embedderd-py 0.1.1`, `trusty-progress 0.2.0`), skipped the 6 `publish = false` crates, and passed the rest. Cross-checked independently against the sparse index for 5 crates — agreement in all 5. The throwaway branch was deleted.
- **`scripts/check_instruction_floor.sh`** — PASS. #4623 landed on main and re-pinned the same file, so this branch was rebased onto `b8e03184` and the pin resolved as the **union**: this PR's guard-workflow hash plus all five of #4623's lines unchanged. The two edits are orthogonal (this PR touches zero files under `assets/instructions/`; #4623 never touched the guard workflow), and `diff` against main's pin now shows exactly one differing line — the guard workflow's own hash.
- **`check_line_cap_selftest`, `check_line_cap`, `check_sld`, `check_doc_numbers`, `check_changelog_fragment`** — all PASS. No changelog fragment is required: this PR changes no `crates/<crate>/src/**` path.
- **Rust quality gate**, mirroring CI's exclusions (`--exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui` — the three Tauri crates CI never compiles, since they need a built `ui/dist` and WebKit2GTK):
  - `cargo fmt --all --check` → **exit 0**
  - `cargo clippy --workspace --all-targets <exclusions> -- -D warnings` → **exit 0**
  - `cargo test --workspace <exclusions>` → **20558 passed, 6 failed, 93 ignored across 152 suites**
  - `cargo test -p trusty-common --features update-check` (CI-unset leg) → **1 passed, 0 failed**

**The 6 failures are environmental, not caused by this PR.** All 6 are macOS FSEvents watcher-timing tests in `trusty-search` (`watch_loop::tests::{cdk_out_file_is_not_indexed, modified_docx_file_triggers_indexing_via_extractor, modified_file_triggers_indexing}`, `watcher::tests::{modified_event_emitted_within_one_second, removed_event_emitted_on_delete}`, `watcher_manager::tests::save_triggers_incremental_index_via_manager`), failing with `event arrives before deadline: Elapsed(())` and `chunk_count never grew`. Evidence rather than assertion:

1. `git diff --stat origin/main...HEAD -- crates/ Cargo.toml Cargo.lock` is **empty** — every Rust source byte in this branch is identical to `origin/main`.
2. The same 6 fail on this machine with `--test-threads=1`, so it is the environment, not parallel load.
3. CI's `Test` job on `01e4bf9b` — this branch's exact merge base, same sources — concluded **success**.

Nothing was `#[ignore]`d, excluded, or otherwise silenced.

### Proven on a live throwaway PR (#4625, closed unmerged)

Two observations the earlier draft listed as unverifiable are now recorded.

**Docs-only path — the highest-blast-radius change, end to end.** With the `changes` job forced to `docs_only=true`, all five ci.yml required contexts reported green without invoking cargo:

| required context | real-code run | docs-only run |
|---|---|---|
| Format check | 14s | **8s** |
| Clippy | 4m51s | **8s** |
| Test | 15m49s | **13s** |
| MSRV check (1.91) | 2m52s | **7s** |
| trusty-search daemon smoke test | 5m58s | **9s** |

Step conclusions confirm it: in both Clippy and Test, every step after `checkout` was `skipped`. `500-line file-size cap` (17s) and `BASE_PM floor` (11s) still ran in full — correct, they are not gated.

Stated precisely: the *classification input* was forced, not observed. A genuinely docs-only PR against main cannot exercise this until these workflow changes are on main — a PR that contains them is by definition not docs-only. What is proven is the step-gating and required-check-reporting mechanism, which is where the blast radius lives. The classifier itself carries 21 fixtures plus the reviewer's independent mutation testing (5 of 6 mutants killed).

**The `needs:` fail-open edge is closed.** With the `changes` job forced to `exit 1`: `Detect docs-only change set` = **failure**, and all eight dependents **ran** rather than reporting `skipped` — Clippy 4 min with `cargo clippy` = success, Test 16 min with `cargo test --workspace` = success. Without `if: ${{ !cancelled() }}` those five required contexts would have been `skipped`, which GitHub counts as passing.

### NOT verified locally — only a live PR can prove these

Stated plainly, because they were reasoned about and not run:

1. **That `edited` actually re-triggers on a base retarget.** Sourced from GitHub's documented activity types and from #4179's own diagnosis of PR #3838. No local harness can deliver a `pull_request` webhook, and a retarget needs a stacked PR to retarget.
2. **That a real docs-only *classification* happens on a real docs-only PR.** The gating mechanism is now proven (above); what remains unobserved is `detect-docs-only.sh` returning `true` from a real diff in CI, which is impossible until these workflows are on main.
3. **That the per-SHA push concurrency group prevents cancellation in practice.** Provable only by two merges landing close together and both runs completing. Sound analytically — `ci-<sha>` is unique per commit, so each group holds exactly one run.
4. **That the notify job's `exit 1` turns a cancelled-jobs run red.** The mapping is tested and the wiring was traced end to end by review, but the GitHub-side behaviour of a `needs`-cancelled job under `if: always()` needs production conditions.
5. **That the smoke test, now without `continue-on-error`, behaves as expected on failure.** #4179's one open mutation experiment (force `exit 1` in the health assertion, then read `mergeStateStatus`) is untouched by this PR.

Closes #4468
Closes #4421

### DIFF
--- a/.github/workflows/agent-assets.yml
+++ b/.github/workflows/agent-assets.yml
@@ -25,6 +25,12 @@ on:
       - "scripts/agent-asset-pins.tsv"
       - ".github/workflows/agent-assets.yml"
   pull_request:
+    # Explicit activity types (#4179). `edited` is NOT in GitHub's default set
+    # (opened/synchronize/reopened) and is the event a base-branch RETARGET
+    # fires, so without it a stacked PR retargeted onto main never re-triggers
+    # and can merge with zero checks — PR #3838 did exactly that.
+    # `ready_for_review` covers a draft being marked ready. Matches ci.yml.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
     branches: [main]
     paths:
       - "crates/trusty-code/src/assets/agents/**"
@@ -33,10 +39,12 @@ on:
       - "scripts/agent-asset-pins.tsv"
       - ".github/workflows/agent-assets.yml"
 
-# Cancel superseded runs for the same ref to save CI minutes (matches ci.yml).
+# Per-COMMIT concurrency group on push so a merge never cancels the previous
+# merge's verification (#4179); per-ref group on pull_request so a new push
+# still supersedes the superseded run. Matches ci.yml.
 concurrency:
-  group: agent-assets-${{ github.ref }}
-  cancel-in-progress: true
+  group: agent-assets-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
 
 jobs:
   agent-assets:
--- a/.github/workflows/agent-assets.yml
+++ b/.github/workflows/agent-assets.yml
@@ -44,7 +44,7 @@ on:
 # still supersedes the superseded run. Matches ci.yml.
 concurrency:
   group: agent-assets-${{ github.event_name == 'push' && github.sha || github.ref }}
-  cancel-in-progress: ${{ github.event_name != 'push' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}
 
 jobs:
   agent-assets:
--- a/.github/workflows/al2023-build.yml
+++ b/.github/workflows/al2023-build.yml
@@ -67,6 +67,12 @@ on:
       - "crates/trusty-common/**"
       - ".github/workflows/al2023-build.yml"
   pull_request:
+    # Explicit activity types (#4179). `edited` is NOT in GitHub's default set
+    # (opened/synchronize/reopened) and is the event a base-branch RETARGET
+    # fires, so without it a stacked PR retargeted onto main never re-triggers
+    # and can merge with zero checks — PR #3838 did exactly that.
+    # `ready_for_review` covers a draft being marked ready. Matches ci.yml.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
     branches: [main]
     paths:
       - "crates/trusty-analyze/**"
@@ -77,10 +83,12 @@ on:
   workflow_dispatch:
     # Allow manual trigger from the Actions UI for ad-hoc validation.
 
-# Cancel superseded runs for the same ref (matches ci.yml convention).
+# Per-COMMIT concurrency group on push so a merge never cancels the previous
+# merge's verification (#4179); per-ref group on pull_request so a new push
+# still supersedes the superseded run. Matches ci.yml.
 concurrency:
-  group: al2023-build-${{ github.ref }}
-  cancel-in-progress: true
+  group: al2023-build-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
 
 env:
   # Suppress ort/fastembed download progress bars in build output.
--- a/.github/workflows/al2023-build.yml
+++ b/.github/workflows/al2023-build.yml
@@ -88,7 +88,7 @@ on:
 # still supersedes the superseded run. Matches ci.yml.
 concurrency:
   group: al2023-build-${{ github.event_name == 'push' && github.sha || github.ref }}
-  cancel-in-progress: ${{ github.event_name != 'push' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}
 
 env:
   # Suppress ort/fastembed download progress bars in build output.
--- a/.github/workflows/capabilities-drift.yml
+++ b/.github/workflows/capabilities-drift.yml
@@ -10,12 +10,20 @@ on:
   push:
     branches: [main]
   pull_request:
+    # Explicit activity types (#4179). `edited` is NOT in GitHub's default set
+    # (opened/synchronize/reopened) and is the event a base-branch RETARGET
+    # fires, so without it a stacked PR retargeted onto main never re-triggers
+    # and can merge with zero checks — PR #3838 did exactly that.
+    # `ready_for_review` covers a draft being marked ready. Matches ci.yml.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
     branches: [main]
 
-# Cancel superseded runs for the same ref to save CI minutes (matches ci.yml).
+# Per-COMMIT concurrency group on push so a merge never cancels the previous
+# merge's verification (#4179); per-ref group on pull_request so a new push
+# still supersedes the superseded run. Matches ci.yml.
 concurrency:
-  group: capabilities-drift-${{ github.ref }}
-  cancel-in-progress: true
+  group: capabilities-drift-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
 
 jobs:
   capabilities-drift:
--- a/.github/workflows/capabilities-drift.yml
+++ b/.github/workflows/capabilities-drift.yml
@@ -23,7 +23,7 @@ on:
 # still supersedes the superseded run. Matches ci.yml.
 concurrency:
   group: capabilities-drift-${{ github.event_name == 'push' && github.sha || github.ref }}
-  cancel-in-progress: ${{ github.event_name != 'push' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}
 
 jobs:
   capabilities-drift:
--- a/.github/workflows/changelog-fragment.yml
+++ b/.github/workflows/changelog-fragment.yml
@@ -20,6 +20,12 @@ name: Changelog fragment
 
 on:
   pull_request:
+    # Explicit activity types (#4179). `edited` is NOT in GitHub's default set
+    # (opened/synchronize/reopened) and is the event a base-branch RETARGET
+    # fires, so without it a stacked PR retargeted onto main never re-triggers
+    # and can merge with zero checks — PR #3838 did exactly that.
+    # `ready_for_review` covers a draft being marked ready. Matches ci.yml.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
     branches: [main]
 
 # Cancel superseded runs for the same ref to save CI minutes (matches ci.yml).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 # CI workflow for the trusty-tools Cargo workspace.
 #
 # Jobs:
+#   changes  – classify the diff as docs-only or not, once, for every job
+#              below to consult (issue #4468: docs-only PRs ran the full Rust
+#              build). The exemption is applied to STEPS, never to the trigger
+#              or to the job, so required checks always report.
 #   fmt      – cargo fmt --all --check (stable)
 #   clippy   – cargo clippy --workspace --all-targets -D warnings (stable, SKIP_UI_BUILD=1)
 #   test     – cargo test --workspace (stable, SKIP_UI_BUILD=1, full clone for git tests)
@@ -79,12 +83,36 @@ on:
   push:
     branches: [main]
   pull_request:
+    # EXPLICIT activity types (#4179). Without a `types:` key GitHub uses the
+    # default set — opened, synchronize, reopened — and RETARGETING a pull
+    # request fires none of them: a base change is delivered as `edited` with
+    # `changes.base`. A stacked PR opened against another feature branch is
+    # therefore filtered out by `branches: [main]` while it is stacked, and
+    # then never re-triggered when it is retargeted to main. PR #3838 merged
+    # with `statusCheckRollup: []` — zero check runs, ever — exactly this way.
+    #
+    # `edited` also fires on title/body edits, so a PR description edit costs a
+    # duplicate run. That is the deliberate trade: a wasted run is recoverable,
+    # a zero-check merge is not, and PR runs are still superseded by the
+    # concurrency group below. `ready_for_review` (also non-default) covers a
+    # draft being marked ready.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
     branches: [main]
 
-# Cancel superseded runs for the same ref to save CI minutes.
+# PULL REQUESTS: one run per PR ref — a new push supersedes the previous run,
+# which is what saves the CI minutes this key was added for.
+#
+# PUSH TO MAIN: the group is keyed to the COMMIT, not the branch, so it is
+# unique per merge and no run is ever cancelled or queued behind a sibling
+# (#4179). The previous `ci-${{ github.ref }}` + `cancel-in-progress: true`
+# meant every push to main shared the group `ci-refs/heads/main`, so each merge
+# KILLED the previous merge's verification: 19 of 60 sampled push-to-main runs
+# (32%) concluded `cancelled`. A cancelled run is the only signal that commit
+# will ever get — the commit is already on main, so there is nothing to
+# supersede and nothing to save by cancelling.
 concurrency:
-  group: ci-${{ github.ref }}
-  cancel-in-progress: true
+  group: ci-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
 
 # SKIP_UI_BUILD=1 prevents the four build.rs Svelte pipelines from invoking pnpm.
 env:
@@ -95,10 +123,69 @@ env:
 
 jobs:
   # --------------------------------------------------------------------------
+  # changes: classify the diff as docs-only or not (issue #4468), once, for
+  # every job below to consult.
+  #
+  # WHY THIS IS A JOB AND NOT A `paths:` FILTER. Five of this workflow's checks
+  # are REQUIRED by branch protection (Format check, Clippy, Test, MSRV check
+  # (1.91), trusty-search daemon smoke test). GitHub does not create check runs
+  # for a workflow its path/branch filters skipped, so a required context would
+  # sit pending forever and a docs-only PR would become unmergeable instead of
+  # fast — the trap #4468 calls out and the one
+  # .github/workflows/changelog-fragment.yml already documents. So every job
+  # still RUNS and still REPORTS; the exemption lives inside the jobs, on the
+  # steps that cost money. A docs-only PR therefore gets a green required check
+  # in ~30 s of runner startup instead of 45 minutes of cargo.
+  #
+  # The same reasoning rules out a job-level `if:`. GitHub documents a
+  # conditionally-skipped job as reporting success to branch protection, but
+  # the failure mode if that is wrong is every docs PR blocked forever, so the
+  # jobs are kept unconditional and only their steps are gated.
+  #
+  # PUSH TO MAIN never takes the exemption: main is the branch everything is
+  # cut from, so it is verified in full regardless of what the diff touched.
+  # --------------------------------------------------------------------------
+  changes:
+    name: Detect docs-only change set
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      docs_only: ${{ steps.detect.outputs.docs_only }}
+    steps:
+      # fetch-depth: 0 — the detector needs `git merge-base` against the base
+      # branch, which a shallow checkout cannot resolve.
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      # The detector and the notifier's conclusion mapping are both pure shell
+      # with fixture-backed selftests; run them here so a regression in either
+      # is caught on the PR that introduces it rather than in production.
+      - name: CI helper selftests (#4179, #4468, #4421)
+        run: bash scripts/check-ci-helpers-selftest.sh
+
+      - name: Classify changed paths
+        id: detect
+        env:
+          DOCS_ONLY_BASE: ${{ github.event.pull_request.base.sha }}
+        run: |
+          if [ "${{ github.event_name }}" != "pull_request" ]; then
+            echo "push to ${{ github.ref }} — verifying in full, no docs-only exemption"
+            echo "docs_only=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          bash scripts/detect-docs-only.sh > /dev/null
+
+  # --------------------------------------------------------------------------
   # fmt: enforce cargo fmt --all --check (stable toolchain + rustfmt component)
+  #
+  # The instruction floor guard below is NOT gated on the docs-only detection:
+  # it guards instruction/markdown content, which is exactly what a docs-only
+  # PR changes.
   # --------------------------------------------------------------------------
   fmt:
     name: Format check
+    needs: [changes]
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -114,6 +201,7 @@ jobs:
         run: bash scripts/check_instruction_floor.sh
 
       - name: Install stable toolchain with rustfmt
+        if: needs.changes.outputs.docs_only != 'true'
         uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
@@ -122,6 +210,7 @@ jobs:
       # nightly-only; stable rustfmt silently ignores them (emits warnings, exit 0).
       # The format check itself is not affected.
       - name: cargo fmt --all --check
+        if: needs.changes.outputs.docs_only != 'true'
         run: cargo fmt --all --check
 
   # --------------------------------------------------------------------------
@@ -130,6 +219,7 @@ jobs:
   # --------------------------------------------------------------------------
   clippy:
     name: Clippy
+    needs: [changes]
     runs-on: ubuntu-latest
     timeout-minutes: 45
     steps:
@@ -138,6 +228,7 @@ jobs:
       # Free disk space (issue #2091): kept in sync with the more aggressive
       # reclamation in the `test` job (see comment there for rationale).
       - name: Free disk space
+        if: needs.changes.outputs.docs_only != 'true'
         run: |
           echo "Before:"; df -h /
           sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android \
@@ -168,11 +259,13 @@ jobs:
           echo "After:"; df -h /
 
       - name: Install stable toolchain with clippy
+        if: needs.changes.outputs.docs_only != 'true'
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
 
       - name: Install system dependencies
+        if: needs.changes.outputs.docs_only != 'true'
         run: |
           sudo apt-get update -qq
           sudo apt-get install -y --no-install-recommends \
@@ -181,6 +274,7 @@ jobs:
             libssl-dev
 
       - name: Cache cargo artifacts
+        if: needs.changes.outputs.docs_only != 'true'
         uses: Swatinem/rust-cache@v2
         with:
           # Separate cache key per job so fmt/clippy/test/msrv don't fight
@@ -188,6 +282,7 @@ jobs:
           key: clippy
 
       - name: cargo clippy --workspace --all-targets -- -D warnings
+        if: needs.changes.outputs.docs_only != 'true'
         run: cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
 
   # --------------------------------------------------------------------------
@@ -225,6 +320,7 @@ jobs:
   # --------------------------------------------------------------------------
   test:
     name: Test
+    needs: [changes]
     runs-on: ubuntu-latest
     timeout-minutes: 45
     env:
@@ -264,6 +360,7 @@ jobs:
       # the Azure CLI (~600 MB), and stale apt package caches/lists — none of
       # which this workspace's build or test steps touch.
       - name: Free disk space
+        if: needs.changes.outputs.docs_only != 'true'
         run: |
           echo "Before:"; df -h /
           sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android \
@@ -298,12 +395,15 @@ jobs:
       # a local branch for the PR head; "main" is only a remote ref unless we
       # explicitly create a local tracking branch for it.
       - name: Create local main branch for git tests
+        if: needs.changes.outputs.docs_only != 'true'
         run: git fetch origin main:main || true
 
       - name: Install stable toolchain
+        if: needs.changes.outputs.docs_only != 'true'
         uses: dtolnay/rust-toolchain@stable
 
       - name: Install system dependencies
+        if: needs.changes.outputs.docs_only != 'true'
         run: |
           sudo apt-get update -qq
           sudo apt-get install -y --no-install-recommends \
@@ -312,6 +412,7 @@ jobs:
             libssl-dev
 
       - name: Cache cargo artifacts
+        if: needs.changes.outputs.docs_only != 'true'
         uses: Swatinem/rust-cache@v2
         with:
           key: test
@@ -325,6 +426,7 @@ jobs:
       # Changing any of these (e.g. upgrading fastembed) rotates the key so
       # the old model blobs are not served to the new version.
       - name: Restore fastembed model cache
+        if: needs.changes.outputs.docs_only != 'true'
         id: fastembed-cache
         uses: actions/cache@v4
         with:
@@ -399,7 +501,7 @@ jobs:
       #       seed_models as a plain statement in a subshell (never as a
       #       conditional's test) and capture its exit status via $?.
       - name: Pre-seed fastembed models (cache miss only)
-        if: steps.fastembed-cache.outputs.cache-hit != 'true'
+        if: needs.changes.outputs.docs_only != 'true' && steps.fastembed-cache.outputs.cache-hit != 'true'
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
@@ -584,6 +686,7 @@ jobs:
       # swept once linking is done and before the (disk-heavier) test
       # execution phase runs. See #3668.
       - name: cargo test --workspace --no-run (build phase)
+        if: needs.changes.outputs.docs_only != 'true'
         run: cargo test --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui --no-run
 
       # Sweep intermediate per-codegen-unit object files (#3668): once a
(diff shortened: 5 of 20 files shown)


bobmatnyc:
## Code critic — Verdict: WARN

Independent adversarial review. I did not write this and did not read the implementer's reasoning before attacking the code. Every claim below is backed by a command I ran in my own read-only worktree at `77aa6d5c`.

**Zero CRITICAL, two HIGH → WARN.** The core of the PR — the conclusion mapping — holds up under mutation testing. Both HIGH findings are about wiring around it, not the mapping itself.

---

## Findings

|---|---|---|---|---|
| HIGH | `scripts/instruction_floor.sha256` | 8–12 | Floor pin collides with #4623, already merged to main. PR is `mergeable: CONFLICTING`. All 31 green checks were computed against a superseded base. | Rebase onto `b8e03184`, resolve as the UNION (exact content below), re-run `check_instruction_floor.sh`. |
| HIGH | `.github/workflows/ci.yml` | 188, 222, 323, 810, 1189 | `needs: [changes]` makes 5 of the 7 required contexts fail-OPEN: if `changes` goes red, all five are `skipped`, and GitHub treats a skipped required check as passing. | Add `if: ${{ !cancelled() }}` to each dependent job. |
| MEDIUM | `scripts/check-ci-helpers-selftest.sh` | 64–65 | The precedence assertion is one-directional. Dropping the red-precedence guard survives the entire 39-case suite. | Add `assert_eq "cancelled cannot downgrade red" "red" "$(verdict_of 'a=failure b=cancelled')"`. |
| MEDIUM | `.github/workflows/ci.yml` | 99, 113–115 | `edited` + `cancel-in-progress: true` on the same PR group means a title/body edit **cancels and restarts** the in-flight run, not "costs a duplicate run". | `cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}` |

---

### HIGH-1 — floor pin collides with #4623 (`scripts/instruction_floor.sha256:8-12`)

This PR's merge base is `01e4bf9b`. `main` is now `b8e03184` (#4623), which re-pinned the same file. Both sides edited it; git cannot auto-merge:

```
$ git merge-tree --write-tree --name-only origin/main HEAD
scripts/instruction_floor.sha256
CONFLICT (content): Merge conflict in scripts/instruction_floor.sha256

$ gh pr view 4624 --json mergeable,mergeStateStatus
{"mergeStateStatus":"DIRTY","mergeable":"CONFLICTING"}
```

`scripts/instruction_floor.sha256` is the **only** conflicting path.

**The good news, stated precisely: the pin is not silently wrong.** Three things I checked rather than assumed:

1. The pin is correct *for this branch's own tree* — `check_instruction_floor.sh` exits 0 here, and `sha256(HEAD:.github/workflows/instruction-floor-guard.yml)` = `19a7810e…`, byte-matching what the pin claims.
2. This PR touches **zero** files under `crates/trusty-mpm/src/assets/instructions/` (`git diff --name-only origin/main...HEAD | grep -c assets/instructions` → 0). #4623 touched only those, never the guard workflow. The two edits are orthogonal, so the union is mechanically determinable — there is no judgement call.
3. The guard is **fail-closed**. Resolving with "ours" drops `enforcement.md` and keeps stale `non-overridable-rules`/manifest hashes; resolving with "theirs" keeps main's stale guard-workflow hash against this PR's edited workflow. Either mistake turns `Format check` red rather than pinning the wrong bytes silently.

The correct resolution is the union of both sides:

```
19a7810e48df249e27e5c281312d9e2dbef8d02057d09f7da538080c9e12dcce  .github/workflows/instruction-floor-guard.yml   # this PR
12c76aaf485d4ffa9d7cd1d73a0c73f4b6e6cd7519e690b22700d5c38f1a7763  …/sections/enforcement.md                      # #4623
25fe05ed4853590324bd5fdce8a147baf52bd959f269ccf99f9adff7c6fc98e8  …/sections/framework-guaranteed-conventions.md # unchanged
255e4b1eb2d3b425d3fcdaf785d7685b5c8c6755492666ca82325703744a65f0  …/sections/identity.md                         # unchanged
f633d6b0233c71852018d255d9e00e5cbf1b2b92e2acd9fa84e45a7cb33de20e  …/sections/non-overridable-rules.md            # #4623
eadf8222b892b12021e169a913d9c968a764a250096a54090e0afaf62abdc12b  manifest-projection:fixed-sections             # #4623
```

Confirm with `bash scripts/check_instruction_floor.sh --update` after the rebase and diff the result against the above.

**Second-order consequence, and the reason this is HIGH rather than MEDIUM:** branch protection has `strict: true` (require branches up to date). Every one of the 31 currently-green checks on this PR ran against `01e4bf9b`. They are stale and must all re-run post-rebase. Do not read the current green as evidence for the merged state — that is the same "a check that reports on something other than what will land" failure mode this PR exists to close.

### HIGH-2 — `needs: [changes]` makes 5 of 7 required contexts fail-open (`ci.yml:188, 222, 323, 810, 1189`)

Branch protection requires exactly:

```
$ gh api repos/bobmatnyc/trusty-tools/branches/main/protection --jq .required_status_checks.contexts
["Format check","500-line file-size cap","Clippy","MSRV check (1.91)","Test",
 "trusty-search daemon smoke test","BASE_PM floor carries the guaranteed conventions"]
```

Five of those seven now carry `needs: [changes]`. Before this PR none of them had any dependency and none could be skipped. If the `changes` job fails, all five become `skipped` — and GitHub reports a skipped job's status as Success, which satisfies a required check.

The `changes` job has two real ways to fail:

- `ci.yml:165` runs `scripts/check-ci-helpers-selftest.sh`. That script exists specifically to go red when any helper regresses — so the designed failure signal for a helper regression is also the trigger that disarms five required gates.
- `ci.yml:177` runs `detect-docs-only.sh`, which `return 2`s when the base ref cannot be resolved. Fail-closed inside the script becomes fail-open at the workflow layer, because a non-zero step kills the job and skips the dependents.

`ci.yml:140-143` explicitly rejects a job-level `if:` on the grounds that skipped-reports-success cannot be verified from here. That reasoning is right, and the same skipped state is then reintroduced through `needs:` — this time in the direction where being wrong costs verification instead of costing a blocked merge. The stated design principle ("if the gating is wrong, a job runs too much — never too little") does not hold across this edge.

**Fix** — add to each of the five (and ideally all eight) dependent jobs:

```yaml
    if: ${{ !cancelled() }}
```

The job then runs even when `changes` fails; `needs.changes.outputs.docs_only` resolves to `''`, `'' != 'true'` is true, so every expensive step runs. Fail-closed, one line per job, no reliance on skip semantics.

### MEDIUM-1 — precedence asserted in one direction only (`check-ci-helpers-selftest.sh:64-65`)

I mutation-tested all 13 conclusion-mapping cases one at a time. **12 of 13 mutants were killed** — success→inconclusive, dropping `timed_out` from the red arm, dropping `failure` from the red arm, default-arm→pass, `cancelled`→pass in isolation, `skipped`→pass in isolation, `neutral`→pass in isolation, fail-open on no inputs, and dropping the red arm from the fold. The mapping table is genuinely non-vacuous.

One mutant survived the entire 39-case suite. At `classify-ci-results.sh:79`, changing

```bash
inconclusive) [ "$verdict" = "red" ] || verdict="inconclusive" ;;
```

to `inconclusive) verdict="inconclusive" ;;` passes all 39 cases, because both precedence fixtures put `failure` **last** (`a=cancelled b=failure`, `a=skipped b=failure`). It is not an equivalent mutant — verified directly:

```
real script:  CI_JOB_RESULTS='a=failure b=cancelled'  -> verdict=red
mutant:       CI_JOB_RESULTS='a=failure b=cancelled'  -> verdict=inconclusive
```

Production behaviour today is correct; the gap is that a future regression in the order-independence of `red > inconclusive` goes undetected. Blast radius is bounded — both verdicts are non-green, so `notify-main-failure` still exits 1 — but the filed issue would say "no verification result" when CI actually failed. One-line fix, in this PR.

### MEDIUM-2 — `edited` cancels in-flight runs rather than duplicating them (`ci.yml:99` + `113-115`)

`types:` adds `edited`, and the PR concurrency group is `ci-${{ github.ref }}` with `cancel-in-progress: true`. For a `pull_request` event `github.ref` is `refs/pull/N/merge` — identical for `synchronize` and `edited`. So editing a PR title or body while its real run is in flight **cancels that run** and restarts from zero.

The PR body frames the cost as "a description edit now costs a duplicate run". Measured against this PR's own timings (`Test` 15m49s, smoke 5m58s), the actual cost is up to ~16 minutes of in-flight work discarded plus a full restart, and under repeated body edits — routine in this repo, where agents update PR descriptions — checks can be kept from ever settling.

Correctness is not affected (the restarted run reports on the same SHA), which is why this is MEDIUM. The trade is real and worth making; the cost is just larger than stated.

**Fix that keeps `edited` without the restart:**

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.action != 'edited' }}
```

An edit-triggered run then queues behind the in-flight one instead of killing it. Note the tempting alternative — a job-level `if:` skipping edit-triggered runs — is **wrong** and the PR body's instinct here is correct: a skipped job publishes a fresh `skipped` check run on the same SHA, which would overwrite a previously red result. That is signal laundering. Filtering at the trigger level would be safe (no new check run at all, prior results stand), but `types:` is all-or-nothing, so the concurrency refinement above is the available lever.

---

## What I verified rather than trusted

- **`scripts/check-ci-helpers-selftest.sh` — 39/39 pass, exit 0.** Re-run in an isolated copy; not vacuous (see mutation results above).
- **Docs-only classifier mutation-tested.** 5 of 6 mutants killed: glob-instead-of-segments (the exact bug the design claims to avoid — killed by the `nested fragment` case), `docs/` without depth guard, naive "any `*.md` anywhere is inert" (killed by the `embedded .md asset` case), fail-open on empty change set. The sixth survivor requires a *directory* literally named `*.md`; not a real-world gap.
- **The segment-matching claim (priority 4) holds everywhere.** There is no leftover glob comparison in `is_inert_path` — every rule is an exact segment equality or a glob against a single `/`-free segment. Confirmed by direct probe: `crates/x/changelog.d/a/b.md` → false, `crates/x/src/assets/agents/a.md` → false, `.github/ISSUE_TEMPLATE/a/b.yml` (n=4) → false.
- **The `include_dir!` claim (priority 3) is correct, and for a reason the PR body understates.** `crates/trusty-mpm/src/core/bundle.rs:277,289` do `include_str!("../../docs/…")`, which resolves to **`crates/trusty-mpm/docs/`**, not the repo-root `docs/`. That path classifies as `false` (n=4, `seg[2]="docs"` ≠ `changelog.d`) — correct. I found no `include_str!`/`include_dir!`/`build.rs` reference reaching the repo-root `docs/` tree or any root `*.md`, so `docs/**` and root `*.md` are genuinely cargo-inert here.
- **Adversarial path shapes.** Renames are handled (`--no-renames` at `detect-docs-only.sh:121` and `check-pr-version-bump.sh:127` decomposes them to delete+add, so both sides get classified). Deletions classify by path, correctly. Non-ASCII and control-character paths are C-quoted by git, so the leading `"` breaks segment 0 and they classify as code — fail-closed. Spaces survive intact and classify correctly. A `.github/*.md` change combined with a workflow change → `false`, correct. The one escape I found, `docs/../crates/x/src/lib.rs` → `true`, is unreachable: `git diff --name-only` never emits `..` components. Not a finding.
- **Priority 2 — the mapping is wired to something that actually fails.** Traced end to end: `ci.yml:1401` runs the classifier with `CI_JOB_RESULTS` set → script writes `verdict=` to `$GITHUB_OUTPUT` (`classify-ci-results.sh:97-100`, independent of the `> /dev/null` on stdout) → `ci.yml:1460-1464` `exit 1` on any non-green. Nothing downstream swallows it. `notify-main-failure` carries `if: always()`, so it still runs when `changes` fails and correctly folds the resulting `skipped` results to `inconclusive` → red. That path is sound.
- **`continue-on-error` is genuinely gone** — `grep -n continue-on-error .github/workflows/ci.yml` returns only comment lines (1164, 1167, 1174, 1365). Claim verified.
- **actionlint 1.7.12 on all 14 changed workflows — exit 0, no output.** Confirmed non-vacuous myself on a scratch copy: corrupting the output reference produced `property "nosuchoutput" is not defined in object type {docs_only: string}`.
- **Priority 5 — per-commit push concurrency cannot deadlock.** `group: ci-<sha>` is unique per commit, so on push each group holds exactly one run: nothing to cancel, nothing to queue behind. A merge train does not serialise. `cancel-in-progress` accepting an expression is valid and actionlint-clean.

## Priority 7 — the five "unverifiable" items

Three of the five are **untested rather than unverifiable**, and one throwaway PR would settle all three at once:

1. `edited` re-triggers on retarget — open a scratch PR against a feature branch, retarget it to main, read the run list. Minutes.
2. Docs-only PR goes green in ~30 s — open a scratch PR touching only `docs/x.md`. **This is the highest-blast-radius change in the PR and the one with no end-to-end proof.** Worth doing before merge, not after.
5. Smoke test behaviour without `continue-on-error` — the mutation experiment is cheap on a throwaway branch.

Items 3 (two merges landing close together) and 4 (a `needs`-cancelled job under `if: always()`) are fairly characterised as needing production conditions. Item 3 is additionally sound analytically, as noted above.

---

## Required changes

1. Rebase onto `b8e03184` and resolve `scripts/instruction_floor.sha256` to the union given in HIGH-1; re-run `bash scripts/check_instruction_floor.sh`. Treat the current 31 green checks as void until they re-run against the rebased tree.
2. Add `if: ${{ !cancelled() }}` to the jobs carrying `needs: [changes]` — minimally the five required contexts at `ci.yml:188, 222, 323, 810, 1189`.
3. Add the red-first precedence case to `check-ci-helpers-selftest.sh`.
4. Either adopt the `cancel-in-progress` refinement in MEDIUM-2, or correct the PR body's cost statement so the accepted trade is recorded accurately.
5. Before merge, prove the docs-only path once with a throwaway docs-only PR (priority-7 item 2).

## Notes

- I reviewed against the spec and the diff only, in an isolated worktree at `77aa6d5c`; I made no edits to the branch and all mutation testing ran on copies in a scratch directory.
- Nothing was flagged in the unchanged bulk of `ci.yml` (disk reclamation, fastembed seeding, the #4614 pipefail fix) — out of scope and no CRITICAL issue in it.
- The `version-parity.yml` unquoted-`name:` fix is correct and worth calling out as a real latent trap defused: promoting `Publish-version parity (issue` to a required context would have created a context that never reports.
- Not flagged, for the record: `crates/*/README.md` is classified inert, which is safe **only** because no crate uses `#![doc = include_str!("../README.md")]` today. I checked; none do. If one ever does, that rule becomes wrong. Not actionable now.

Issue #4571, opened 2026-08-01
chore: migrate every consumer onto the credential authority (discharges #4035-#4039)


bobmatnyc:
Parent epic: #4040. Plumbing — pure migration, no new capability.

**This is the single execution ticket for the migration scope that #4035, #4036, #4037, #4038 and #4039 folded into #4040 when they were closed on 2026-07-28.** It deliberately does **not** re-file those five as five new issues — that would recreate the duplicate queue surface their consolidation removed. It exists so that scope stays visible on the epic's checklist rather than living only in the epic's prose.

Read those five closed issues for the per-consumer detail; their acceptance criteria remain the acceptance criteria and are not restated here.

## Why this comes last

#4040's "Done when" says *"No migrated consumer owns a parallel secret resolver."* Migration is how that becomes true — but migrating before the authority API (principal, `CredentialRef`, ACL, audit) has stabilised means migrating every consumer twice. Every one of these consumers works today; this is debt paydown, not an unblock.

## Scale, measured on `origin/main` @ `fd15c589`

Production source only (`crates/*/src/**`, test files excluded):

- **55 raw `std::env::var("<CREDENTIAL>")` reads across 46 files in 9 crates**, each bypassing the resolver entirely. By crate: trusty-common 15, trusty-agents 14, trusty-code 9, trusty-git-analytics 8, trusty-review 4, trusty-mpm 2, trusty-installer 1, trusty-gworkspace 1, trusty-analyze 1.
- **105 production source files reference `OPENROUTER_API_KEY`** across 10 crates. The audit table's 2026-07-11 baseline recorded 22 files / 8 crates.
- **59 production source files reference `GITHUB_TOKEN`** across 8 crates. Baseline recorded 13 / 6.

Both counts have grown materially since the baseline, which is the argument for finishing this rather than leaving it partially covered.

## GitHub auth is worse than #4036 recorded

#4036 named three independent PAT-resolution schemes. On `origin/main` there are **six** distinct GitHub credential paths:

| # | Path | What it does |
|---|---|---|
| 1 | `trusty-review/src/integrations/github/auth/strategy.rs` | `GITHUB_TOKEN` → `GH_TOKEN` → shell out to `gh auth token` |
| 2 | `trusty-review/src/integrations/github/auth/app.rs` | GitHub App JWT/RS256 installation-token minting + its own cache |
[...]

Issue #4570, opened 2026-08-01
Decide and enforce credential at-rest storage — the OS keychain is compiled by nobody


bobmatnyc:
Parent epic: #4040.

## ⛔ BLOCKED on an owner decision — deliberately NOT scheduled

**No milestone by design.** This is a genuine product fork, not an implementation question, and it has now been raised and abandoned twice without an answer. Do not start it until the owner rules. See owner question **Q1** on #4040.

## The finding

The repo has an OS-keychain backend. **Nothing compiles it.**

`KeyringStore` (`crates/trusty-common/src/inference/credentials/keyring_store.rs`) sits behind a cargo feature:

```toml
# crates/trusty-common/Cargo.toml:561
keyring-store = ["credentials", "dep:keyring"]
```

A workspace-wide search for `keyring-store` across every `Cargo.toml`, CI workflow and the `Makefile` finds **only its own definition and three comments referring to it.** No crate enables it. So in every shipped binary, `default_store()` (`resolver.rs:118`) takes the `#[cfg(feature = "keyring-store")]` branch **never**, and falls straight through to:

```rust
match FileKeyStore::new() { Ok(store) => Box::new(store), Err(_) => Box::new(MemoryKeyStore::new()) }
```

`FileKeyStore` writes `~/.trusty-tools/credentials.toml` — permission-hardened to `0600` (correctly, and re-asserted on every write) but **plaintext on disk**.

**Net effect: today, every credential this repo stores at rest is stored in plaintext**, and the keychain path that exists to prevent that is dead code in every build.

## Why it is a decision and not a bug

Two closed issues framed exactly this tension and neither was answered:

- **#3066** *"[M1/decision] Resolve the credential plaintext-fallback tension (Eve §10.1)"* — closed `NOT_PLANNED` 2026-07-28. Body: *"Owner directive is 'never plaintext,' but the shared … `default_store()` falls back to a permission-hardened (0600) but plaintext `FileKeyStore` when the OS keychain probe fails. Decide whether … resolution accepts that same fallback … or hard-fails."*
- **#3076** *"[AUTH-3] … + hard-fail-no-plaintext"* — closed `NOT_PLANNED` the same day. It proposed scoping the hard-fail to a `user.*` namespace with an explicit opt-in env escape hatch for headless/CI.

[...]

Issue #4568, opened 2026-08-01
feat(trusty-agents): replace inline McpService.env secrets with credential references


bobmatnyc:
Parent epic: #4040. Plumbing — it removes a secret sink, it grants nothing.

**This discharges the Phase-1 delivery scope that #3038 (spec DOC-45, remote MCP credential delivery) folded into #4040. It is not a re-file of #3038** — #3038's *spec* work belongs to this epic's DOC-45 child; this is the code change that spec's Phase 1 called for.

## The finding

`McpService` (`crates/trusty-agents/src/mcp/config/types.rs:170`) carries:

```rust
/// Environment variables overlaid on the spawned server's process
/// environment (e.g. API keys). Empty by default …
#[serde(default)]
pub env: HashMap<String, String>,
```

This is a plaintext secret sink, by design and by its own doc comment. The file is `~/.trusty-agents/config.toml` (`crates/trusty-agents/src/mcp/config/mod.rs`, `CONFIG_DIR_NAME`/`CONFIG_FILE_NAME`) — a user-facing config file the product invites people to edit.

There is no alternative today, because `credential_ref` **does not exist in any Rust source in this repo** (zero hits under `crates/*/src/**`; it appears only in `docs/specs/trusty-agents-eve-style-agents-spec.md` and one research doc, despite being an acceptance criterion of the now-closed-`COMPLETED` epic #2808).

## Why it is urgent now, not later

#4523 adds `~/trusty-agents/<instance>/` — a **dotless, deliberately browsable, hand-editable** per-assistant home, and DOC-63 `S-5.6` (PR #4530) makes it normative that no credential is ever written under it. The existing `env: HashMap<String,String>` pattern is exactly the shape that would be copied into a store row for Gmail, Drive, Slack, Notion or Granola if a reference type is not available first. PR #4523's own closing note flags this: *"Secrets must not land in the home … The spec should name a separate credential store."*

Fixing the MCP config sink now establishes the pattern before the OKG store rows are written, rather than after.

## Scope

1. Add an optional `credential_ref` alongside `env` on `McpService`, holding a `CredentialRef` (from the `CredentialRef` child).
2. Resolve at spawn/call time — into the subprocess environment for `stdio` transport, and into the request header for `http` transport. Never into the config file, never into a `ToolResult`, never into a tracing capture.
[...]

Issue #4567, opened 2026-08-01
feat(trusty-common): credential access audit trail — attributable resolution, allow and deny


bobmatnyc:
Parent epic: #4040. Plumbing. Without this, #4040's "Done when" clause *"access is attributable and revocable"* cannot be satisfied — revocable arrives with the ACL child, attributable arrives here.

## The gap, verified

**There is no credential audit trail anywhere in this workspace.** A search for audit machinery under `crates/*/src/**` returns three unrelated things and nothing else:

- `tools::phase_audit::PhaseAuditTool` (trusty-agents) — agent *phase* transitions, not credentials.
- `session::registry_search_audit` (trusty-code) — a capped ring of *search* records.
- `commands::doctor::audit::audit_palaces` (trusty-memory) — a memory-palace health check.

`resolve_key` records nothing. Post-incident there is no way to answer *"which principal read which credential, when, and was it allowed?"*

## Scope

An audit record emitted on **every** credential resolution attempt — allowed and denied alike, since a denial is the more interesting forensic event.

Record fields (final shape set by the DOC-45 spec child): timestamp, principal, credential identity (the `CredentialRef`, never the value), decision, denial reason where applicable, and the requesting call site.

## The hard constraint

DOC-63 `S-5.8`: *"neither may record a credential, and neither may echo a request header or URL query string that could carry one."* This is not advisory. The audit stream is the one place where a bug converts a security control into a secret-disclosure channel.

The design must therefore make secret material **unrepresentable** in a record — the record type accepts a `CredentialRef`, not a `String` — rather than relying on redaction of an arbitrary string. Redaction heuristics on arbitrary text are unreliable; #4521 reaches the same conclusion independently for shell command text.

## Routability

Follow #4521's reasoning: a dedicated `tracing` target (e.g. `trusty_common::credential_audit`) so an operator can route the stream to appropriate retention and access, or suppress it, **without silencing the rest of the crate's logging.**

## Relationship to #4521

#4521 asks for an audit log on `l0_shell_exec`, and #4550 asks whether a permission *decision* is itself audited. Three streams are in play and they should not be invented three times:

| stream | what it records | owner |
|---|---|---|
| credential access | principal, credential ref, allow/deny | **this issue** |
| shell execution | command, cwd, exit code | #4521 |
| permission decision | allow / ask / deny outcome | #4550 (spec) |

[...]

PR #4627, opened 2026-08-03
refactor(trusty-common): promote credentials to a top-level module + complete the provider registry (#4564)

### LINKED ISSUE #4564: feat(trusty-common): promote credentials to a top-level module and complete the provider registry
Parent epic: #4040. Pure plumbing — no behaviour change, no capability widened, no consumer migrated.

## Verified current state (`origin/main` @ `fd15c589`)

The credential resolver exists and is good, but it is filed under `inference::` and its provider registry covers a minority of this repo's credentials.

**Where it lives:** `crates/trusty-common/src/inference/credentials/` — `resolver.rs` (3-tier: process env → `.env.local` → `KeyStore`), `keyring_store.rs`, `file_store.rs`, `memory_store.rs`, `dotenv.rs`, `redact.rs`.

**Why the path is now wrong:** `env_var_for` (`resolver.rs:44`) already maps four *non-inference* credentials — `slack`, `slack-user`, `telegram`, `claude-code` — and its own doc comment says *"Not limited to inference providers."* Ten mappings exist, four of them are not inference. The module is misfiled, and the misfiling is why consumers keep not finding it.

**The registry gap, measured.** A census of production source (`crates/*/src/**`, test files excluded) finds **55 raw `std::env::var("<CREDENTIAL>")` reads across 46 files in 9 crates**, naming **23 distinct credential environment variables**. `env_var_for` maps 10 providers. **13 of those 23 names have no mapping at all**, so no consumer *can* route them through the resolver even if it wanted to:

`GITHUB_TOKEN`, `GH_TOKEN`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_WEBHOOK_SECRET`, `JIRA_TOKEN`, `JIRA_API_TOKEN`, `LINEAR_API_KEY`, `BITBUCKET_TOKEN`, `BITBUCKET_APP_PASSWORD`, `BRAVE_API_KEY`, `GOOGLE_OAUTH_CLIENT_SECRET`, `SLACK_APP_TOKEN`, `TAGENT_API_TOKEN`.

`SLACK_APP_TOKEN` is the sharpest illustration: `SLACK_BOT_TOKEN` and `SLACK_USER_TOKEN` are both mapped, its sibling is not.

## Scope

1. Promote `inference::credentials` to a top-level `trusty_common::credentials` module.
2. Keep `trusty_common::inference::credentials::*` as a deprecated re-export shim so **no consumer crate changes in this PR**.
3. Register the 13 unmapped credential names, each with the provider key its future consumer will use.
4. Split the cargo feature accordingly (`credentials` must not require the `inference` feature set).

## Explicitly NOT in scope

- Migrating any call site off its raw `std::env::var`. That is a separate child, deliberately sequenced after the authority API stabilises so nothing is migrated twice.
- Changing resolution precedence, the store backends, or the at-rest format.

## Acceptance criteria

- [ ] `trusty_common::credentials` is the canonical path; `trusty_common::inference::credentials` re-exports it and is marked `#[deprecated]` with the new path in the message.
- [ ] `cargo check --workspace` passes with **zero** changes outside `crates/trusty-common/`.
- [ ] `env_var_for` maps all 23 credential env-var names found in the census; a test enumerates them so a future addition that skips the registry fails CI.
- [ ] A test asserts `env_var_for` is case-insensitive for every registered provider (the existing behaviour, now pinned across the enlarged set).
- [ ] The `credentials` feature builds without the inference provider stack.
- [ ] `crates/trusty-common/CHANGELOG.md` gains an `## [Unreleased]` entry.

## Sequencing

Blocked by: nothing (can land in parallel with the DOC-45 spec child).
Blocks: the `CredentialRef` child, the ACL child, and #4478's per-provider brokering (JIRA/Linear become resolvable the moment they are registered).

Milestone 22 — smallest real unblock, and it is what makes #4478 question (b) answerable.

bobmatnyc:
Closes #4564. Parent epic #4040. Built against **DOC-45** (`docs/specs/DOC-45-credential-authority-model.md`) and **ADR-0026**, both of which landed on `main` in #4606.

Pure plumbing. **No behaviour change, no capability widened, no consumer migrated off its raw `std::env::var`.** Precedence, store backends, at-rest format, and the `credentials` feature's dependency set are all byte-identical.

## 1 — The move

`crates/trusty-common/src/inference/credentials/` → `crates/trusty-common/src/credentials/` (`git mv`, so the diff renders as renames).

The path was wrong and DOC-45 §1.3 says why: `env_var_for` already mapped four **non-inference** credentials (`slack`, `slack-user`, `telegram`, `claude-code`) and its own doc comment already read *"Not limited to inference providers."* Filing the authority under `inference::` is what kept consumers from finding it — the 55 raw `std::env::var` reads are the evidence.

- `trusty_common::credentials` is the canonical path.
- `trusty_common::inference::credentials` remains as a `#[deprecated(since = "0.28.0")]` re-export naming the new path.
- Feature-gating preserved. `credentials` still does **not** require `inference-client`; `trusty-agents` still reaches the resolver via `inference-client = ["credentials"]`; `trusty-channels` still builds on `features = ["mcp", "credentials"]` alone. Verified by building `trusty-common` under each of `--no-default-features`, `credentials`, `keyring-store`, `inference-client`, `config-cli`, `bedrock-client` separately (raw output below).
- `crates/trusty-common/src/inference/mod.rs` stays gated on the `credentials` feature rather than `inference-client`, deliberately: that is what keeps the deprecated alias resolving for a consumer that enables only `credentials`.

### Deviation from the issue's "zero changes outside `crates/trusty-common/`" bullet — flagged, not silent

The issue asked for zero consumer changes, on the theory that the shim makes them unnecessary. It does not, for a mechanical reason: CI runs `cargo clippy --workspace --all-targets -- -D warnings`, so **every** in-tree use of the deprecated alias becomes a hard build failure. A shim plus untouched consumers is not a state this repo's CI can be in. So all 54 in-tree call sites moved to the new path in this PR — which is also what the dispatching brief asked for ("update every import") — and the shim is left as grace for out-of-tree consumers only. Nothing in this workspace triggers the deprecation.

## 2 — The registry

`env_var_for` moved to `credentials::registry` and became a `REGISTRY` **table** rather than a `match`. A `match` cannot be asserted complete, and completeness is the whole ticket.

**13 entries added**, taking the registry from 10 to 23 — the full census of credential env-var names used in production source:

| Provider key | Env var | Note |
|---|---|---|
| `slack-app` | `SLACK_APP_TOKEN` | the sharpest gap: unmapped between its two mapped siblings |
| `tagent` | `TAGENT_API_TOKEN` | |
| `github` | `GITHUB_TOKEN` | |
| `github-gh-cli` | `GH_TOKEN` | separate key on purpose — see below |
| `github-app` | `GITHUB_APP_PRIVATE_KEY` | |
| `github-webhook` | `GITHUB_WEBHOOK_SECRET` | |
| `bitbucket` | `BITBUCKET_TOKEN` | |
| `bitbucket-app-password` | `BITBUCKET_APP_PASSWORD` | |
| `jira` | `JIRA_TOKEN` | unblocks #4478 (b) |
| `jira-api` | `JIRA_API_TOKEN` | unblocks #4478 (b) |
| `linear` | `LINEAR_API_KEY` | unblocks #4478 (b) |
| `brave` | `BRAVE_API_KEY` | |
| `google-oauth` | `GOOGLE_OAUTH_CLIENT_SECRET` | |

`GITHUB_TOKEN` and `GH_TOKEN` get **separate provider keys** rather than one key with two names: `gh` prefers `GH_TOKEN` over `GITHUB_TOKEN`, and collapsing them would make the resolver unable to express which of the two a given call site actually consults. Same reasoning that already keeps `slack` / `slack-user` apart.

`LINEAR_TEAM_KEY`, `JIRA_PROJECT_KEY` and `GOOGLE_OAUTH_CLIENT_ID` appear in the same `env::var` sweep and are **deliberately not registered** — they are non-secret configuration, and registering them would put non-credentials into the credential authority.

Pinned by four new tests plus the two moved verbatim from `resolver::tests`:

- `registry_covers_the_full_census` — asserts **both directions**: every censused name is registered, and the registry invents nothing beyond the census. Neither list can drift alone.
- `env_var_for_is_case_insensitive_for_every_provider` — exact / upper / mixed case for all 23, not a two-provider spot check. Case-insensitivity was pre-existing behaviour of the `match` that the table could have silently dropped.
- `registry_has_no_duplicate_provider_keys` / `registry_has_no_duplicate_env_vars` — a duplicate key makes lookup order-dependent; a duplicate env var makes one of the two entries un-revocable.
- `env_var_for_known_providers`, `env_var_for_unknown_provider_is_none` — **moved, not deleted**, along with the code they cover.

**One honest gap.** The acceptance bullet says the enumeration should make "a future addition that skips the registry fail CI". The test above fails when the census list is edited without the table (and vice versa), but it cannot detect a *brand-new* raw `env::var("NEW_SECRET")` added elsewhere in the workspace — that needs a source-scanning gate, which DOC-45 `C-3.12` assigns to the migration ticket **#4571** ("SHOULD ship a lint or grep-based gate that prevents a 56th raw read"). Not absorbed here.

## Scope discipline

Nothing in this PR touches at-rest storage (#4570), principal/ACL/default-deny (#4566), the audit trail (#4567), `McpService.env` (#4568), `CredentialRef` (#4565), or the 55 raw `env::var` call sites (#4571). Registering a credential name grants nothing: DOC-45 `C-9.8` — the storage tiers say *where a value lives*, never *who may read it*.

## Public API

- **Added:** `trusty_common::credentials` (whole module), `credentials::registry`, `credentials::REGISTRY`, `credentials::registered_providers()`.
- **Deprecated, still working:** `trusty_common::inference::credentials::*`.
- **Breaking:** none. Every symbol previously reachable at the old path is reachable at both paths.

## Gates

All run on the final head. Raw output in the first comment.


bobmatnyc:
## Raw gate output — head `1c15b08`

```
$ git rev-parse HEAD
1c15b084e259017ee807dd3663e1f4f99dbc2439

$ cargo fmt --check
cargo fmt --check exit=0

$ cargo clippy -p trusty-common --all-targets --features credentials,keyring-store,inference-client,config-cli -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.68s

$ cargo test -p trusty-common --features credentials,keyring-store,inference-client,config-cli
     Running unittests src/lib.rs
running 440 tests
test result: ok. 428 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 0.72s
     Running tests/community_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/help_yaml_workspace.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/inference_foundation.rs
running 8 tests
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/kg_graph_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/kg_store_snapshot_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
   Doc-tests trusty_common
test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ cargo test -p trusty-common --features credentials,keyring-store,inference-client,config-cli -- --test-threads=1
test result: ok. 428 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 3.50s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ SKIP_UI_BUILD=1 cargo test -p trusty-agents --lib
test result: ok. 3280 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 28.15s

line-cap: measured 3589 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 53 spec doc(s) + 3019 code file(s); 0 error(s), 0 warning(s)

test-pointers: resolved 20224 Test: citation(s) (floor 200) — 0 dangling pointers — OK.

$ SKIP_UI_BUILD=1 cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 47.09s
# (the only warnings emitted are `tga`'s pre-existing clippy.toml/Cargo.toml MSRV mismatch,
#  a cargo-level note that is not a -D warnings failure and is unrelated to this change)
```

### Feature-gating proof — `credentials` builds without the inference provider stack

```
$ cargo check -p trusty-common --no-default-features
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.25s
$ cargo check -p trusty-common --no-default-features --features credentials
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.75s
$ cargo check -p trusty-common --no-default-features --features keyring-store
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
$ cargo check -p trusty-common --no-default-features --features inference-client
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.71s
$ cargo check -p trusty-common --no-default-features --features config-cli
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.94s
$ cargo check -p trusty-common --no-default-features --features bedrock-client
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.51s
```

### The new registry tests, individually

```
$ cargo test -p trusty-common --features credentials -- credentials::registry::
test credentials::registry::tests::deprecated_inference_alias_still_resolves ... ok
test credentials::registry::tests::env_var_for_known_providers ... ok
test credentials::registry::tests::env_var_for_unknown_provider_is_none ... ok
test credentials::registry::tests::env_var_for_is_case_insensitive_for_every_provider ... ok
test credentials::registry::tests::registered_providers_matches_the_table ... ok
test credentials::registry::tests::registry_covers_the_full_census ... ok
test credentials::registry::tests::registry_has_no_duplicate_env_vars ... ok
test credentials::registry::tests::registry_has_no_duplicate_provider_keys ... ok
```

Also verified that the new `# Spec References` block genuinely resolves rather than being silently skipped: mutating its anchor to `SPEC-CREDAUTH-99~draft` makes `check_sld.sh` fail with `ref-anchor-mismatch` + `ref-anchor-missing`, and reverting it returns the gate to clean.

Issue #4566, opened 2026-08-01
feat(trusty-common): principal identity, ACL default-deny, delegation narrowing, and revocation signal


bobmatnyc:
Parent epic: #4040. Plumbing — this grants nothing; it is the machinery that lets a grant be *withheld*.

## The gap, verified

There is no principal concept and no ACL anywhere in the credential path today. `resolve_key(provider)` (`crates/trusty-common/src/inference/credentials/resolver.rs:79`) takes a provider name and nothing else — **any code in any crate that can call it gets any credential the process can see.** That is the whole access-control model at present.

The two prior attempts to close this both died without an answer:

- **#3076** `[AUTH-3] user.* credential-namespace gating in the credential resolver + hard-fail-no-plaintext` — closed `NOT_PLANNED` 2026-07-28. It proposed exactly this: reserve a `user.*` prefix on a credential reference, and require the currently-executing agent's manifest to carry `user_authority: true` to resolve one.
- **#3066** the plaintext-fallback tension — closed `NOT_PLANNED` the same day.

`user_authority` itself **does** exist and is enforced (10 source files, `crates/trusty-agents/src/agents/permissions.rs`, `api/server/agent_permissions.rs`, `tools/cross_product.rs`) — but it gates *tools*, never credentials. The seam #3076 wanted is real and still open.

## Scope

Implement the ACL model that the DOC-45 spec child settles:

1. **`Principal`** — the concrete type. Its granularity is set by the spec, not here.
2. **Default-deny.** An unlisted (principal, store) pair resolves to `CredentialError::Denied`, never to any credential within that store. This is the inversion: today the default is allow-everything-in-process.
3. **Grant expression** — where a grant is declared and how it is read. Must not be declarable from a file inside the assistant home (DOC-63 `S-5.6`), and must not be widenable by a persona TOML edit — the same server-owned-floor property ADR-0024 decision 4 established for `ASSISTANT_REACHABLE_SUBAGENTS`.
[...]

Issue #4040, opened 2026-07-26
epic: unified credential authority, delivery, and audit


bobmatnyc:
## Outcome

Every first-party secret is resolved through one governed system that knows which user, agent, session, or service may read it and records that access.

## Scope

- principal and credential identity, ACL/scoping, default-deny behavior, revocation, and audit trail;
- secure remote delivery to fleet sessions without committing secrets;
- shared GitHub App and PAT resolution;
- migration of git-analytics, gworkspace OAuth, and trusty-memory provider keys;
- compatibility adapters for external token formats;
- subagent and project-trust boundaries;
- diagnostics for missing, denied, expired, zero-scope, or scope-unavailable credentials (`Missing`/`Denied`/`Expired`/`ZeroScope`/`ScopeUnavailable`, DOC-45 §7.1 `C-5.1`-`C-5.6`).

## Done when

No migrated consumer owns a parallel secret resolver, access is attributable and revocable, remote delivery cannot leak into project files or model context, and compatibility paths have conformance tests.

DOC-45 and an authority ADR capture the design. Former tickets #3038 and #4035–#4039 are consolidated here.

bobmatnyc:
Triage consolidation (2026-07-28): this issue now preserves the listed former leaves as explicit scope and completion criteria. Their closure reduces duplicate queue surface and does not claim the work is implemented.

bobmatnyc:
## Owner decision (2026-08-01): this epic has acquired downstream consumers

The owner decided that the project-scoping, audit-logging, and confirmation-gate design needed by any sub-agent mediating git, shell, or credentialed access on an assistant's behalf will be produced **as part of this epic**, not as bespoke per-issue designs. This epic's existing scope — principal/credential identity, ACL/scoping, default-deny, revocation, audit trail, and explicitly "subagent and project-trust boundaries" — already covers that mechanism, and a separate design elsewhere would duplicate and drift from it.

**Downstream consumers now blocked on this epic's design output:**
- **#4417** — Route git calls through a Versioning sub-agent (project-scoping/audit/confirmation property)
- **#4479** — Add `version-control` to `ASSISTANT_REACHABLE_SUBAGENTS` (same design property, stated as a required constraint on that work)
- **#4439** — Route shell through the Computer Use sub-agent (Constraint 2(b)–(d): filesystem/project scope, audit logging, user confirmation — NOTE its Constraint 1, authority-holder gating, is separate and NOT covered here)
[...]

Issue #2808, opened 2026-07-16
epic(trusty-agents): Tool binding + credential brokering via OS keyring


bobmatnyc:
## Scope

Manifest-driven tool binding plus credential brokering via the OS keyring.
`ToolExecutor` already unifies native/MCP-management/MCP-external tool
dispatch, and external MCP calling is already live via `McpServiceTool` — the
real remaining gaps are transport coverage and credentials, not the
dispatch model itself.

Concretely:
- Close the stdio-only MCP transport gap (add `http` transport support for
  `McpService`).
- Implement `rpc/mod.rs`'s unimplemented `tools/call` proxy (issue #460 TODO)
  — flagged in the spec that `ServiceDescriptor` has no execute method, so
  this needs a research/design pass before implementation.
- Implement `credential_ref` resolution via the OS keyring: manifest declares
  a credential reference, runtime resolves the secret at call time into the
  subprocess env (stdio) or request header (http), and the secret is never
  exposed in LLM-visible `ToolResult` content or debug/tracing capture.
- Timeout handling: a tool call exceeding 30s surfaces a recoverable error,
  not a hang or panic.

## References

- Epic tracking issue: #2791
- Spec: `docs/specs/trusty-agents-eve-style-agents-spec.md`, **SPEC-AGENTFW-03
  — Unified Tool-Calling, MCP Mapping & Credential Brokering** (PR #2792,
  branch `spec-eve-style-agents`, currently receiving a v3 rewrite — re-check
  before implementing)
- Milestone: TA-M3: Tools & credentials
- Related: issue #460 (rpc `tools/call` proxy TODO)

## Acceptance (spec Conformance bullets, §4.3)

- `POST /rpc {"method":"tools/call","params":{"name":"memory_recall",...}}`
  against a running `tagent --api` instance returns a JSON-RPC result (not
  `-32601`), and an unknown tool name returns `-32602`.
- An `McpService` with `transport = "http"` and a reachable `url` produces a
  callable `ToolExecutor` (currently silently skipped).
- An `McpService` with `credential_ref` set spawns/calls with the resolved
  secret present in the subprocess env (stdio) or request header (http), and
  a `tracing`/debug capture of the LLM-visible `ToolResult` content never
  contains the literal secret value.
- A tool call exceeding 30s surfaces `ToolResult::Error { recoverable: true,
  .. }`, not a hang or panic.

[...]

PR #4629, opened 2026-08-03
feat(trusty-common): CredentialRef, Secret, and resolve-at-use-time (#4565)

### LINKED ISSUE #4565: feat(trusty-common): CredentialRef, Secret, and resolve-at-use-time — the reference type nothing has today
Parent epic: #4040. Plumbing. This is DOC-63 §7.1b requirements **1, 2 and 5**, and it is what lets a config row hold a credential *pointer* in plaintext.

## The gap, verified

`credential_ref` — the field the Eve-style agents spec (`docs/specs/trusty-agents-eve-style-agents-spec.md`) describes and that epic #2808 listed in its acceptance criteria — **does not exist in any Rust source in this repo.** A workspace-wide search finds it only in that spec (16 hits) and one research document (4 hits). Zero hits under `crates/*/src/**`. #2808 is closed `COMPLETED`; the credential-brokering leg of it did not land. Flagging that as an accuracy note, not a decision to make here.

What exists instead: `McpService` (`crates/trusty-agents/src/mcp/config/types.rs:170`) carries

```rust
pub env: HashMap<String, String>,
```

documented as *"Environment variables overlaid on the spawned server's process environment (e.g. API keys)"*. So today an MCP server's secret is stored **literally, in plaintext, in a user-editable TOML file** — `~/.trusty-agents/config.toml`. There is no reference type, so there is nothing else it could be.

## Why a reference type is the load-bearing piece

DOC-63 `S-5.6`/`S-5.7` (PR #4530) make this normative for OKG stores: a store row carries *"a **credential reference** — an opaque handle resolved by #4040 at use time — and never the credential itself."* PR #4523's own closing note asks for the same thing, because the assistant home is deliberately browsable and hand-editable. A reference type is the mechanism that makes "no secrets in the home" structurally true rather than a rule people remember.

## Scope

In `trusty_common::credentials`:

1. **`CredentialRef`** — an opaque, durable, serialisable, **non-secret** handle. Safe to write into a git-tracked file, a config TOML, a log line, or a model-visible tool result.
2. **`Secret<T>`** — a wrapper whose `Debug` and `Display` never render the value. `crates/trusty-common/src/inference/types/secret.rs` already exists; extend or fold it in rather than writing a second one.
3. **Resolution at use time** — `resolve(&CredentialRef, &Principal) -> Result<Secret<String>, CredentialError>`, plus a shape that hands back an *already-authenticated client handle* rather than a bare string where the provider needs one (DOC-63 `S-5.1`: a source type receives a resolved, scoped client and nothing else).
4. **Both credential shapes** — OAuth (Gmail, Drive, Notion, Slack) and plain API key (Granola, Fireflies). DOC-63 §7.1b item 5 explicitly warns that the API-key shape is the one likely to be special-cased and must not be.
5. **`CredentialError`** distinguishing *missing* / *denied* / *expired* / *zero-scope*, per #4040's "Done when". Every variant recoverable.

`Principal` may be a stub in this child if the ACL child has not landed; the signature must be the final one.

## Acceptance criteria

- [ ] `CredentialRef` round-trips through serde and is `Display`-able without revealing anything secret; a test asserts a `CredentialRef`'s rendered form does not contain the resolved secret.
- [ ] `Secret<String>`'s `Debug`/`Display` output contains no substring of the wrapped value; property-tested.
- [ ] A resolved credential is never returned by value from a function whose result is `Serialize` into a `ToolResult` — pinned by a compile-time or test-level assertion.
- [ ] Both an OAuth-shaped and an API-key-shaped credential resolve through the *same* entry point; a test exercises both.
- [ ] `CredentialError` has the four named variants and each is constructible/observable from a test.
- [ ] `crates/trusty-common/CHANGELOG.md` gains an `## [Unreleased]` entry.

## Sequencing

Blocked by: the provider-registry child (needs the enlarged registry) and, for the `Principal` signature, the DOC-45 spec child.
Blocks: the ACL child, the audit child, the `McpService.env` migration child, and OKG store rows (#4537, #4546–#4549).

Milestone 22.

bobmatnyc:
Closes #4565. Parent epic #4040. **Stacked on #4627** (`feat/4564-credentials-module`) — that PR must land first; this one is based on its branch so the diff below is only #4565's work.

Built against **DOC-45** §4 (`CredentialRef`), §7 (denial taxonomy) and §10 (`Secret`, delivery), and **ADR-0026**.

## The gap this closes

`credential_ref` had **zero hits** under `crates/*/src/**` despite being an acceptance bullet of epic #2808, closed `COMPLETED`. Without a reference type, anything that needs to *refer* to a credential has to *hold* one — which is why `McpService.env` is a `HashMap<String, String>` of literal API keys in a hand-editable `~/.trusty-agents/config.toml`, and why DOC-63 `S-5.6`'s "no secrets in the assistant home" is a rule people remember rather than a property of the types.

### `CredentialRef` — the handle (`C-2.1`–`C-2.7`)

`provider` or `provider/qualifier`. Lowercase-kebab segments, at most one `/`, ≤ 64 bytes total and ≤ 32 per segment. `Display` renders it **verbatim** (`C-2.6`) — the deliberate opposite of `Secret`, and the contrast is the design: the handle prints, the value never does.

The grammar is `C-2.4`'s "restrictive enough that no realistic API key, JWT, OAuth token, or PEM body can satisfy it". `realistic_credentials_are_rejected_by_the_grammar` pins that against ten specimens shaped like the real formats this registry names — GitHub classic and fine-grained PATs, OpenAI, Slack bot and app, Telegram, Brave, a JWT, a PEM body, a Google OAuth secret. Every one is rejected by uppercase, `_`, `.`, `:`, a newline, or length.

**Stated honestly:** the grammar excludes every credential format in the workspace's census, which is exactly `C-2.4`'s claim. It is not a proof against a *deliberately chosen* lowercase-hex string ≤ 64 chars. The property it buys is that a secret cannot be laundered into a ref — and thence into a retained audit stream (`C-7.3`) — by accident.

Shape-agnostic (`C-2.5`): one type, no `Kind`, no `is_oauth()`, no second constructor. DOC-63 §7.1b item 5 warns the API-key shape is the one likely to be special-cased; `oauth_and_api_key_shapes_resolve_through_one_entry_point` resolves `google-oauth` and `brave` through one call with one argument list.

### `Secret<T>` — the value (`C-8.2`, `C-8.3`)

**Redaction proof.** Two properties, because the literal acceptance criterion is weaker than what is actually true here:

1. `debug_and_display_are_value_independent` — the strongest form. `Debug`/`Display` render a **constant**, and the impls carry **no `T: Debug` bound**, so they *cannot* consult the value even in principle. Quantified over 74 specimens (realistic credential formats, non-ASCII, empty, and 64 deterministic pseudo-random printable-ASCII strings of length 1–64): every one renders byte-identically to the empty secret. A rendering that never varies with the value cannot be disclosing it.
2. `rendering_contains_no_substring_of_the_value` — `C-8.2` as literally written, over every contiguous substring of length ≥ 3 of every credential-shaped (≥ 8 char) specimen. Two documented exclusions: sub-8-char specimens are not credential-shaped, and a candidate already present in the rendering of the *empty* secret is exempt — the constant `Secret(<redacted>)` contains `ecr`, so the English word `secret` trips a naive version of this without anything having leaked. The exemption keeps the test measuring disclosure rather than coincidence, and property 1 is what makes it safe.

`Secret` implements **neither** `Serialize`, `Deserialize`, `Clone`, `Deref`, `AsRef`, nor `PartialEq`. Each omission closes a leak path; the table in its doc comment says which.

**Three compile-time assertions** (`not_serialize_not_clone`) use the `assert_not_impl` coherence trick — inlined rather than adding a dependency for three assertions — so the build fails the day `Secret<String>` gains `Serialize`, `Deserialize`, or `Clone`. That is #4565's acceptance criterion 3 and DOC-45 `C-8.3`'s "pinned at compile time rather than by review". **Verified by perturbation**, not just asserted: adding `impl Clone for Secret<String>` and `impl Serialize for Secret<String>` produces two `E0283 type annotations needed` / `multiple impls satisfying … AmbiguousIfImpl` errors and the crate does not build (output in the first comment).

### How resolve-at-use-time is *enforced*, not merely enabled

This is the question the brief asked to be answered explicitly. Four mechanisms, in descending order of how hard they are to work around:

1. **A config struct cannot hold the result.** `Secret` is not `Serialize`/`Deserialize`, and every config type in this workspace derives both, so a `Secret` field **fails to compile** inside one. The only credential-shaped thing a config row *can* hold is a `CredentialRef` — which is `C-8.8` made structural. Backed by the compile-time assertion above, so this cannot regress silently.
2. **The result cannot be duplicated into a longer-lived home.** Not `Clone`; `expose()` returns a **borrow**, not an owned `String`. Caching a resolved credential is a deliberate `into_inner()` move, not an accidental `.clone()`.
3. **Resolution needs facts a config loader does not have.** `resolve` takes a `Principal` and a `Scope` — *who is acting* and *what for*. Neither is answerable at config-load time: a loader has no principal in hand and no operation in flight. The eager call is not discouraged by convention, it has no arguments.
4. **The caller can avoid seeing the value at all.** `resolve_client::<T: FromCredential>` resolves, lends the `Secret` by reference to the builder, and drops it before returning — so a source type receives *"a resolved, scoped client and nothing else"* (`C-3.4`, DOC-63 `S-5.1`). A caller that never sees the string cannot leak the string.

### `CredentialError` — five variants (`C-5.1`–`C-5.10`)

`Missing` / `Denied` / `Expired` / `ZeroScope` / `ScopeUnavailable`. All recoverable (`C-3.11`), all matchable structurally rather than by string comparison (`C-5.8`), all naming the principal and the ref in an actionable remediation (`C-5.7`), and none able to carry secret material — **by construction**, since every field is a `CredentialRef`, a `Principal`, or a `Scope` and none of the three can hold a value (`C-5.9`).

The fifth variant is DOC-45 `C-5.5`'s deliberate addition to #4040's stated four, kept distinct from `ZeroScope` by `C-5.6` because "widen the grant" is advice that cannot be followed for a provider with no such scope, and the workaround is to grant write.

### `Principal` / `Scope` — the vocabulary, deliberately inert

`Principal` is a closed, `#[non_exhaustive]` enumeration (`C-1.7`) carrying **`Operator` and `Service` only**. DOC-45 `C-1.3` is PROVISIONAL pending owner **Q-B** and says in terms: *"Do not implement `Principal`'s `Assistant` variant until Q-B is answered"*, and `C-12.1` repeats it as a delivery rule. `#[non_exhaustive]` means #4566 adds `Assistant` and `SubAgent` without a breaking change.

`Scope::covers` is defined here, before there is an ACL to use it, specifically so a second definition does not appear inside #4566's grant check. Write covers read; read does not cover write; an empty provider-scope list covers only a request naming none (the fail-closed reading).

## Honesty clause — what this PR does NOT do

**There is no authorization here.** `resolve` accepts a `Principal` and checks **no grant** against it. Grants, the ACL, default-deny, and the code-owned floor are #4566 (DOC-45 §5). Today's behaviour is byte-for-byte what `resolve_key` already does: the three storage tiers answer, and any caller that can reach the function gets the value. `C-3.12` says the same about default-deny at large — it is not airtight until #4571 migrates the 55 raw `std::env::var` reads.

What #4565 buys is that **the signature will not change** when #4566 lands: the grant check goes inside `resolve`, between the registry check and the store lookup, at the marked line. No consumer is migrated twice.

Nothing here touches at-rest storage (#4570), the audit trail (#4567), `McpService.env` (#4568), or consumer migration (#4571).

## `SecretString` — the one thing I did not do, and why

The issue says to *"extend or fold in"* `inference::types::SecretString` rather than write a second wrapper. I extended it (`into_secret()`, superseded-by docs) but did **not** fold it in, and this is a deliberate stop rather than an omission:

- `SecretString`'s `Debug`/`Display` disclose a **four-character head preview** via `redact_secret`. That does not meet `C-8.2`'s "no substring of the wrapped value", so it cannot become a `Secret<String>` alias without changing behaviour.
- That preview shape is pinned by an **existing test**, `secret_debug_is_redacted`, which asserts `dumped.contains("chars")`. Collapsing the types requires changing that test.
- `SecretString` also derives `Clone` and `PartialEq`, which the inference stack relies on across `ResolvedProvider` and every provider adapter, and which `Secret` deliberately does not have.

The brief's standing instruction is to stop and report rather than weaken or rewrite an existing test, so I did. `secret_string_converts_to_the_canonical_secret` documents the divergence executably: it asserts the old type *does* leak a head, then asserts the head is gone once converted. **This is a real pre-existing leak** — every `ResolvedProvider` debug dump prints the first four characters of the live API key — and it wants its own ticket.

## Public API

- **Added:** `credentials::{CredentialRef, CredentialRefError, Secret, CredentialError, Principal, ServiceId, Scope, Access, FromCredential, resolve, resolve_with, resolve_client, resolve_client_with}`; `credentials::authority`; `SecretString::into_secret`.
- **Breaking:** none.

## Gates

All run on the final head. Raw output in the first comment.

### DIFF
--- a/crates/trusty-common/changelog.d/4565-credential-ref-secret-resolve-at-use-time.md
+++ b/crates/trusty-common/changelog.d/4565-credential-ref-secret-resolve-at-use-time.md
@@ -1,0 +1,63 @@
+Added
+
+- `credentials::CredentialRef` — the opaque, durable, **non-secret** handle that
+  names a credential without carrying it (closes
+  [#4565](https://github.com/bobmatnyc/trusty-tools/issues/4565), epic
+  [#4040](https://github.com/bobmatnyc/trusty-tools/issues/4040), DOC-45
+  `C-2.1`–`C-2.7`). `credential_ref` had zero hits under `crates/*/src/**`
+  despite being an acceptance bullet of the closed-as-completed #2808, so
+  anything that needed to *refer* to a credential had to *hold* one — which is
+  why `McpService.env` is a map of literal API keys in a hand-editable TOML. A
+  ref is safe in a git-tracked file, a config row, a log line, an audit record,
+  and a model-visible tool result: its grammar is lowercase-kebab segments,
+  ≤ 64 bytes, at most one `/`, which no realistic API key, JWT, OAuth token, or
+  PEM body can satisfy. Pinned by
+  `realistic_credentials_are_rejected_by_the_grammar` against specimens shaped
+  like every credential format the registry names. Stable across rotation, and
+  shape-agnostic: one type, one entry point, and no code path that exists only
+  for OAuth or only for a plain API key.
+- `credentials::Secret<T>` — the wrapper a resolved credential comes back in.
+  Its `Debug`/`Display` render a constant that is not merely redacted but
+  *independent of the value* (the impls carry no `T: Debug` bound, so they
+  cannot read it), and it implements neither `Serialize`, `Deserialize`,
+  `Clone`, `Deref`, nor `PartialEq`. Each omission is a closed leak path; the
+  absent `Serialize` is the load-bearing one, because every config struct in the
+  workspace derives it, so a `Secret` cannot compile inside one. Three
+  compile-time assertions (the `assert_not_impl` coherence trick, inlined rather
+  than adding a dependency) fail the build if any of the three traits is ever
+  added.
+- `credentials::resolve(&CredentialRef, &Principal, &Scope) -> Result<Secret<String>, CredentialError>`
+  — the single resolution entry point, called where the credential is consumed
+  rather than at config load (`C-3.3`, `C-8.4`). `resolve_client` is the same
+  entry point with the value consumed in place, handing back an
+  already-authenticated handle so the caller never sees the string (`C-3.4`,
+  DOC-63 `S-5.1`). A ref naming a provider absent from the registry fails with
+  `Missing` carrying a remediation that names the registry (`C-2.7`).
+- `credentials::CredentialError` — `Missing` / `Denied` / `Expired` /
+  `ZeroScope` / `ScopeUnavailable`. Every variant is recoverable, carries the
+  `CredentialRef` and the `Principal`, renders an actionable remediation, and
+  can hold no secret material by construction. The fifth variant is DOC-45
+  `C-5.5`'s deliberate addition to #4040's stated four, kept distinct from
+  `ZeroScope` by `C-5.6` because "widen the grant" is advice that cannot be
+  followed for a provider that has no such scope.
+- `credentials::Principal`, `ServiceId`, `Scope`, `Access` — the vocabulary
+  `resolve`'s final signature is written in. `Principal` is a closed,
+  `#[non_exhaustive]` enumeration carrying `Operator` and `Service` only:
+  DOC-45 `C-1.3` is PROVISIONAL pending owner question Q-B and says in terms
+  not to implement the `Assistant` variant until it is answered, so #4566 adds
+  `Assistant` and `SubAgent` without a breaking change.
+- `SecretString::into_secret` — the one-line migration from the older inference
+  wrapper to the canonical `Secret<String>`.
+
+Changed
+
+- `inference::types::SecretString` is documented as superseded by
+  `credentials::Secret`. Its behaviour is unchanged; note that its
+  four-character head preview does **not** meet DOC-45 `C-8.2`, and collapsing
+  the two types is left to a follow-up because it requires changing the existing
+  test that pins that preview's shape.
+
+Note: this change adds **no authorization**. `resolve` accepts a `Principal` and
+checks no grant against it — grants, the ACL, default-deny and the code-owned
+floor are [#4566](https://github.com/bobmatnyc/trusty-tools/issues/4566). The
+signature is final so no consumer is migrated twice.
--- a/crates/trusty-common/src/credentials/authority.rs
+++ b/crates/trusty-common/src/credentials/authority.rs
@@ -1,0 +1,427 @@
+//! The single credential-resolution entry point, resolved **at use time**
+//! (issue #4565, DOC-45 `C-3.3`/`C-3.4`/`C-8.4`).
+//!
+//! # Spec References
+//!
+//! - [`SPEC-CREDAUTH-03~draft`](docs/specs/DOC-45-credential-authority-model.md#SPEC-CREDAUTH-03~draft)
+//!
+//! Why: `C-3.3` requires exactly one entry point, and `C-8.4` requires
+//! resolution to happen at use time — *"never at config-load time and never at
+//! spawn-list-construction time"* — so the window in which a secret exists in
+//! memory is bounded by the call that needs it. Those two clauses only bite if
+//! the *shape* of the API makes the eager version awkward and the lazy version
+//! natural, which is what this module is for.
+//!
+//! ## How use-time resolution is enforced, not merely enabled
+//!
+//! 1. **A config struct cannot hold the result.** [`Secret`] implements neither
+//!    `Serialize` nor `Deserialize`, and every config type in this workspace
+//!    derives both, so a `Secret` field fails to compile inside one. The only
+//!    credential-shaped thing a config row *can* hold is a
+//!    [`CredentialRef`], which is exactly `C-8.8`. Pinned at compile time in
+//!    `super::secret::not_serialize_not_clone`.
+//! 2. **The result cannot be duplicated into a longer-lived home.** [`Secret`]
+//!    is not `Clone` and [`Secret::expose`] returns a borrow, so caching one is
+//!    a deliberate move, not an accident.
+//! 3. **Resolution needs facts a config loader does not have.** [`resolve`]
+//!    takes a [`Principal`] and a [`Scope`]: *who is acting* and *what for*.
+//!    Neither is answerable at config-load time — a loader has no principal in
+//!    hand and no operation in flight — so the eager call is not merely
+//!    discouraged, it has no arguments.
+//! 4. **The caller can avoid seeing the value at all.** [`resolve_client`]
+//!    hands back an already-authenticated handle built from the secret, so a
+//!    source type receives *"a resolved, scoped client and nothing else"*
+//!    (`C-3.4`, DOC-63 `S-5.1`). A caller that never sees the string cannot
+//!    leak the string.
+//!
+//! ## What this module does NOT do — the honesty clause
+//!
+//! **There is no authorization here.** [`resolve`] takes a `Principal` and
+//! checks no grant against it, because grants, the ACL, default-deny, and the
+//! code-owned floor are #4566 (DOC-45 §5) and #4565 is explicitly scoped to the
+//! reference type and the signature. Today's behaviour is therefore identical
+//! to [`super::resolve_key`]'s: the three storage tiers answer, and any caller
+//! that can reach the function gets the value. `C-3.12` says the same thing
+//! about default-deny at large — it is not airtight until #4571 migrates the 55
+//! raw `std::env::var` reads, and this document does not claim it is.
+//!
+//! What #4565 buys is that the signature will not change when #4566 lands: the
+//! grant check goes *inside* [`resolve`], and no consumer is migrated twice.
+//!
+//! Test: `tests::oauth_and_api_key_shapes_resolve_through_one_entry_point`,
+//! `tests::unregistered_provider_is_missing_and_names_the_registry`,
+//! `tests::absent_credential_is_missing`,
+//! `tests::resolve_client_never_hands_back_the_string`,
+//! `tests::resolved_secret_does_not_render_in_debug_or_display`.
+
+use super::error::CredentialError;
+use super::handle::CredentialRef;
+use super::principal::{Principal, Scope};
+use super::registry::env_var_for;
+use super::secret::Secret;
+use super::{KeyStore, default_store, dotenv, resolve_key_with};
+
+/// Resolve `credential` for `principal` at `scope`, at the point of use.
+///
+/// Why: the one entry point `C-3.3` requires. A second resolution path is a
+/// defect under this repo's common-entry-point rule and under DOC-63 `S-5.2`,
+/// and is rejected at review regardless of expedience.
+///
+/// What: loads `.env.local` once, then delegates to [`resolve_with`] against
+/// [`default_store`]. Returns the value wrapped in a [`Secret`], which cannot
+/// be serialised, cloned, or printed.
+///
+/// **Call this where the credential is consumed** — inside the function that
+/// builds the request, not in the constructor that builds the config. `C-8.4`.
+///
+/// # Errors
+///
+/// [`CredentialError::Missing`] when the reference names an unregistered
+/// provider (`C-2.7`, with a remediation naming the registry) or when no tier
+/// holds a value. The other four variants become reachable when #4566 adds the
+/// grant check; see the module docs.
+///
+/// Test: `tests::oauth_and_api_key_shapes_resolve_through_one_entry_point`
+/// exercises the hermetic core; the `load_env_local_once` + `default_store`
+/// wiring is intentionally not independently unit tested, for the same reason
+/// [`super::resolve_key`]'s is not.
+pub fn resolve(
+    credential: &CredentialRef,
+    principal: &Principal,
+    scope: &Scope,
+) -> Result<Secret<String>, CredentialError> {
+    dotenv::load_env_local_once();
+    resolve_with(credential, principal, scope, default_store().as_ref())
+}
+
+/// Hermetic core of [`resolve`]: same decision, injectable store.
+///
+/// Why: separated so tests can inject a `MemoryKeyStore` and control the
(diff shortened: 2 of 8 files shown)


bobmatnyc:
## Raw gate output — head `a74a345`

```
$ git rev-parse HEAD
a74a34554061f5266280771cdf5af4e43f9fb87d

$ cargo fmt --check
exit=0

$ cargo clippy -p trusty-common --all-targets --features credentials,keyring-store,inference-client,config-cli -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.31s

$ cargo test -p trusty-common --features credentials,keyring-store,inference-client,config-cli
     Running unittests src/lib.rs
running 466 tests
test result: ok. 454 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 0.85s
     Running tests/community_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/help_yaml_workspace.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/inference_foundation.rs
running 8 tests
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
     Running tests/kg_graph_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/kg_store_snapshot_tests.rs
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
   Doc-tests trusty_common
test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ cargo test -p trusty-common --features credentials,keyring-store,inference-client,config-cli -- --test-threads=1
test result: ok. 454 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 2.70s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ SKIP_UI_BUILD=1 cargo test -p trusty-agents --lib
test result: ok. 3280 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 27.62s

line-cap: measured 3594 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 53 spec doc(s) + 3024 code file(s); 0 error(s), 0 warning(s)

test-pointers: resolved 20290 Test: citation(s) (floor 200) — 0 dangling pointers — OK.

$ SKIP_UI_BUILD=1 cargo clippy --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 43.13s
# (only warning emitted is `tga`'s pre-existing clippy.toml/Cargo.toml MSRV mismatch — a
#  cargo-level note, not a -D warnings failure, and unrelated to this change)
```

### The 29 new tests, by name

```
$ cargo test -p trusty-common --features credentials -- credentials::{handle,secret,error,principal,authority}
test credentials::authority::tests::absent_credential_is_missing ... ok
test credentials::authority::tests::oauth_and_api_key_shapes_resolve_through_one_entry_point ... ok
test credentials::authority::tests::qualified_refs_reach_distinct_store_rows ... ok
test credentials::authority::tests::resolve_client_never_hands_back_the_string ... ok
test credentials::authority::tests::resolved_secret_does_not_render_in_debug_or_display ... ok
test credentials::authority::tests::unregistered_provider_is_missing_and_names_the_registry ... ok
test credentials::error::tests::every_variant_is_constructible_and_matchable ... ok
test credentials::error::tests::no_variant_renders_secret_material ... ok
test credentials::error::tests::remediation_names_the_principal_and_the_ref ... ok
test credentials::handle::tests::deserialize_rejects_out_of_grammar_text ... ok
test credentials::handle::tests::display_is_verbatim ... ok
test credentials::handle::tests::parse_error_never_echoes_the_input ... ok
test credentials::handle::tests::qualifier_is_optional_and_preserved ... ok
test credentials::handle::tests::realistic_credentials_are_rejected_by_the_grammar ... ok
test credentials::handle::tests::rejects_out_of_grammar_text ... ok
test credentials::handle::tests::round_trips_through_serde ... ok
test credentials::principal::tests::principal_renders_its_full_shape ... ok
test credentials::principal::tests::provider_scopes_must_all_be_covered ... ok
test credentials::principal::tests::scope_read_is_covered_by_write ... ok
test credentials::principal::tests::service_id_rejects_out_of_grammar_text ... ok
test credentials::secret::tests::compile_time_assertions_hold ... ok
test credentials::secret::tests::debug_and_display_are_value_independent ... ok
test credentials::secret::tests::expose_returns_the_raw_value ... ok
test credentials::secret::tests::into_inner_moves_the_value ... ok
test credentials::secret::tests::rendering_contains_no_substring_of_the_value ... ok

$ cargo test -p trusty-common --features credentials,inference-client -- inference::types::secret
test inference::types::secret::tests::secret_debug_is_redacted ... ok
test inference::types::secret::tests::secret_display_is_redacted ... ok
test inference::types::secret::tests::secret_expose_returns_raw ... ok
test inference::types::secret::tests::secret_string_converts_to_the_canonical_secret ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 454 filtered out; finished in 0.00s
```

The three pre-existing `SecretString` tests are unchanged and still pass; the fourth is new and records the divergence executably.

### Perturbation proof — the compile-time assertions actually bite

Temporarily adding the two impls `Secret` must never have:

```rust
impl Clone for Secret<String> { fn clone(&self) -> Self { Self(self.0.clone()) } }
impl serde::Serialize for Secret<String> { /* … */ }
```

```
$ cargo check -p trusty-common --no-default-features --features credentials
error[E0283]: type annotations needed
note: multiple `impl`s satisfying `Secret<std::string::String>: _ASSERT_NOT_SERIALIZE::{closure#0}::AmbiguousIfImpl<_>` found
error[E0283]: type annotations needed
note: multiple `impl`s satisfying `Secret<std::string::String>: _ASSERT_NOT_CLONE::{closure#0}::AmbiguousIfImpl<_>` found
error: could not compile `trusty-common` (lib) due to 2 previous errors
```

Reverted; the tree builds clean. This is the evidence for #4565's acceptance criterion 3 ("pinned by a compile-time or test-level assertion") and DOC-45 `C-8.3`.

### Feature-gating unchanged

```
credentials              Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.58s
keyring-store            Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.89s
inference-client         Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.08s
config-cli               Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.51s
bedrock-client           Finished `dev` profile [unoptimized + debuginfo] target(s) in 20.47s
```

### One note on the specimen literals

GitHub push protection rejected the first push: the invented `xoxb-…` and `xapp-…` specimens in the grammar-rejection tests match its Slack detector. They are fabricated, not live, and the correct fix is not to click "allow the secret" — so they are now assembled with `concat!("xo", "xb", …)`, which keeps the format realistic for the test while leaving no scannable literal in the source. Every affected test still passes.

bobmatnyc:
## ⚠️ Zero checks here is expected, and is **not** green — read before merging

`gh pr checks 4629` reports *"no checks reported"*. That is the documented behaviour of this repo's CI, not a pass:

`.github/workflows/ci.yml` triggers `pull_request` with `branches: [main]`, so a PR stacked on a feature branch is filtered out while it is stacked. #4624 added `edited` to the trigger's `types:` precisely so the **retarget** to `main` — which GitHub delivers as `edited` with `changes.base`, firing none of the default activity types — starts a real run. That is the failure mode PR #3838 merged into: `statusCheckRollup: []`, zero check runs, ever.

**So the merge order is load-bearing:**

1. Merge #4627 (green on `135598af`, re-verified after rebasing onto `3afe4c08`).
2. This PR auto-retargets to `main`; that retarget fires CI.
3. Wait for it to settle, then merge. **Do not merge on the strength of an empty check list.**

Local evidence in the meantime, on head `a74a3455` (identical tree to the current `3bbd292f`, which is the same commit rebased onto #4627's rebased head): the full gate suite is in the comment above, including `cargo clippy --workspace --all-targets -- -D warnings` in CI's exact exclusion form, `cargo test -p trusty-common` (454 passed), `cargo test -p trusty-agents --lib` (3280 passed), and all three `scripts/check_*.sh` gates. That is the same job set CI runs, but it is local evidence and does not substitute for the run.

Issue #4657, opened 2026-08-03
epic(trusty-agents): B6: queue-and-surface-on-return behaviour


bobmatnyc:
## Outcome

Queue-and-surface-on-return behaviour per **D4** (owner decision, epic
#4646): when the owner does not respond to a notification, notify ONCE,
then queue everything else silently, and present the full queue when the
owner returns. No retry. No escalation to a second channel.

## Scope

- After B2 sends a notification and the owner doesn't respond within
  whatever window is defined, subsequent notify_owner calls for the same
  unattended period are queued, not re-sent.
- No automatic retry of the same notification.
- No automatic escalation from Telegram to Slack (or vice versa) if the
  first channel gets no response — that would violate D4 explicitly.
- On the owner's return (B1's unattended state clears), surface the full
  queued backlog in one shot.

## Depends on

B1 (unattended/return transition), B2 (the tool whose repeated calls this
governs).

Part of epic #4646 (unsolicited assistant-to-owner notifications).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closed not planned in the 2026-09-02 milestone-68 review (owner-confirmed): dormant scheduled-tasks / notify-owner / personal-assistant cluster — its only open references were sibling epics in the same cluster, no movement since the 2026-08-14 sweep. Reopen if the personal-assistant scope returns.

Issue #4654, opened 2026-08-03
epic(trusty-agents): B3: Telegram binding for notify_owner


bobmatnyc:
## Outcome

Telegram binding for B2's `notify_owner` tool — the first channel, per
**D2** (owner decision, epic #4646).

## Architectural constraint

`crates/trusty-agents` deliberately does NOT depend on the `trusty-mpm`
crate, documented at
`crates/trusty-agents/src/tools/session_state/store.rs:1-12`. trusty-mpm
already owns a Telegram alert loop
(`crates/trusty-mpm/src/telegram/alerts.rs:1-64`), but it is closed-vocabulary
daemon telemetry keyed on fixed `HookCategory` values
(`crates/trusty-mpm/src/core/hook.rs:227-246`) — it cannot carry an
assistant's free-text message today. This binding must be a **thin HTTP
call from trusty-agents to trusty-mpm**, not a crate dependency, and not a
reuse of the closed-vocabulary alert path without extending it to accept
free text.

## Scope

- HTTP call path from B2's seam into trusty-mpm's Telegram delivery.
- Extend trusty-mpm's Telegram delivery to accept an assistant-authored
  free-text message (distinct from `HookCategory` telemetry).

## Depends on

B2 (the seam this binds into).

Part of epic #4646 (unsolicited assistant-to-owner notifications).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Owner decided 2026-08-03 (#4674) that PM-session escalation and assistant notifications share one channel abstraction. See follow-up #4675 for the shared design — do not build a private transport binding here.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Per the ownership decision on #4675 (comment
[5169853680](https://github.com/bobmatnyc/trusty-tools/issues/4675#issuecomment-5169853680)):
trusty-mpm owns the shared owner-notification channel abstraction, and this
binding changes character. The Telegram transport already exists inside
trusty-mpm (`crates/trusty-mpm/src/telegram/alerts.rs`) — bot connections,
timeouts are reusable as-is. This is no longer "build a binding." It is
"route the new endpoint's payloads through the existing transport, replacing
the closed-vocabulary `HookCategory` decision layer" (which cannot carry
[...]

Issue #4653, opened 2026-08-03
feat(trusty-agents): B2: notify_owner MCP tool — thin HTTP client to trusty-mpm's owner-notification endpoint


bobmatnyc:
## Outcome

An assistant-facing `notify_owner` MCP tool that is a THIN AUTHENTICATED HTTP
POST CLIENT to a trusty-mpm endpoint — mirroring the shared-secret-header
pattern already in trusty-mpm's `slack/relay.rs`. It never touches
Telegram/Slack directly and never depends on the `trusty-mpm` crate.

## Why (rescope, see #4675 comment)

Per the decision recorded on #4675 (comment
[5169853680](https://github.com/bobmatnyc/trusty-tools/issues/4675#issuecomment-5169853680)):
trusty-mpm OWNS the shared owner-notification channel abstraction; this
issue becomes a CLIENT of it, not the abstraction itself.

`TELEGRAM_BOT_TOKEN` lives in trusty-mpm's launchd plist (`com.trusty.mpm`,
`RunAtLoad` + `KeepAlive`, always-on singleton), while trusty-agents' bot is
a background task in an EPHEMERAL REPL process with no launchd unit that
dies with the session — so only trusty-mpm's daemon is durable enough to
hold the credential and be running when a notification must go out.

The crate-layering rule at
`crates/trusty-agents/src/tools/session_state/store.rs:1-12` bars a crate
DEPENDENCY, not HTTP; trusty-agents already calls trusty-mpm over HTTP at
`crates/trusty-agents/src/system_status/daemons.rs:172-182`.

## Scope

- Assistant-callable MCP tool, gated on #4652's unattended signal.
- Thin authenticated HTTP POST client to a new trusty-mpm daemon endpoint
  (discovery-file URL + shared-secret header, mirroring
  `crates/trusty-agents/src/slack/relay.rs`'s existing write-side pattern).
- Payload carries a wire-level `source` STRING TAG (not a shared Rust enum —
  the boundary is HTTP-only between crates with no shared types) so PM
  telemetry and assistant reasoning stay distinguishable.
- Depends on the trusty-mpm-side endpoint existing: #4681.

## Depends on

B1 (unattended gate, #4652). #4681 (trusty-mpm POST endpoint).

## Feeds

B3 (Telegram binding, #4654), B4 (Slack binding, #4655), B5 (wires
wake_bound_agents' discarded reply through this tool), B7 (audit every
call).

[...]

Issue #4646, opened 2026-08-03
epic(trusty-agents): unsolicited assistant-to-owner notifications


bobmatnyc:
## Outcome

An assistant can reach the owner unprompted when it has something that
needs attention and the owner is unattended — a bounded, audited,
channel-agnostic notification, not free-running spam. This is the second of
two gaps the owner (Bob) named on 2026-08-03 as blocking trusty-agents from
being a daily driver (the other is #4645, self-scheduled tasks).

Parent: #3052 (daily-driver personal Assistant).

## Owner decisions (encode, do not reopen)

- **D2:** Egress builds now behind a thin channel-agnostic abstraction —
  Telegram binding first (B3), Slack second (B4) — rather than waiting for
  the #3157 control bus. **ADR-0019** confirms the bus was decided
  2026-07-18 but **no code has landed** (lines 112-117: "the bus design was
  largely decided... and remains unimplemented"). Decision: build ahead of
  it and accept one adapter rewrite when #3157 lands.
- **D3:** "Unattended" = a last-human-turn timeout (one tunable, N minutes).
  Not SSE connection counting, not manual do-not-disturb.
- **D4:** When the owner does not respond: notify ONCE, then queue silently
  and surface everything on the owner's return. No retry, no escalation to
  a second channel.

## Verified current state (2026-08-03 research pass)

- The ONLY unprompted egress in shipped code is trusty-mpm's Telegram/Slack
  alert loop (`crates/trusty-mpm/src/telegram/alerts.rs:1-64`), and it is
  closed-vocabulary daemon telemetry keyed on fixed `HookCategory` values
  (`crates/trusty-mpm/src/core/hook.rs:227-246`) plus memory thresholds — an
  assistant's own reasoning cannot push free text through it.
- trusty-agents' Telegram and Slack bots are **inbound-reply-only**.
- **No** `notify_owner`/`send_alert` tool exists anywhere.
- `wake_bound_agents` already wakes an agent on a Gmail event, gets a reply,
  and **discards** it — only `tracing::info!`-logged
  (`crates/trusty-agents/src/listeners/wake.rs:268-271`).
- "Foreground" has no representation in code: focus is a routing `HashMap`
  with no lease/heartbeat (`crates/trusty-mpm/src/telegram/focus.rs`);
  `SessionRecord.last_activity_at` (`record.rs:204`) counts an assistant's
  own tool calls identically to a human typing; SSE has no connection-count
[...]

PR #4683, opened 2026-08-03
feat(trusty-agents): unattended detection via last-human-turn timeout (#4652)

### LINKED ISSUE #4652: epic(trusty-agents): B1: unattended detection (last-human-turn timeout)
## Outcome

"Unattended" detection: a last-human-turn timeout, one tunable threshold
(N minutes), per **D3** (owner decision, epic #4646). This gates B2's
`notify_owner` tool.

## Why this is new work, not an adaptation

Research confirmed no existing signal can be adapted:

- `SessionRecord.last_activity_at`
  (`crates/trusty-mpm/src/session_manager/record.rs:204`) counts an
  assistant's own tool calls identically to a human typing — it cannot
  distinguish "owner is here" from "the assistant is busy."
- Focus tracking is a routing `HashMap` with no lease/heartbeat
  (`crates/trusty-mpm/src/telegram/focus.rs`).
- SSE has no connection-count tracking
  (`crates/trusty-agents/src/api/server/events_sse.rs`).

Per **D3**: explicitly NOT SSE connection counting, NOT manual
do-not-disturb toggling. The signal is strictly "how long since the last
turn that originated from a human," with one tunable timeout.

## Scope

- Track last-human-turn timestamp per session/persona (distinct from
  last-activity, which includes the assistant's own tool calls).
- Expose a single tunable threshold (config, default TBD by implementer).
- Expose an `is_unattended()` check consumable by B2.

Part of epic #4646 (unsolicited assistant-to-owner notifications).

bobmatnyc:
Closes #4652. Part of epic #4646 (unsolicited assistant-to-owner notifications), owner decision **D3**.

## What this delivers

Code can now answer "is a human currently attending this assistant instance?" Nothing could before.

The three signals that looked adaptable were each verified unusable, which is why this is new work:

- `SessionRecord.last_activity_at` (`crates/trusty-mpm/src/session_manager/record.rs:204`) advances on TOOL and HOOK activity — an assistant grinding through a long solo task advances it identically to a human typing.
- Telegram/Slack `focus` is a per-conversation routing `HashMap` with no lease and no heartbeat. It survives the human walking away unchanged.
- `crates/trusty-agents/src/api/server/events_sse.rs` tracks no connection count at all.

Per **D3** the mechanism is a **last-human-turn timeout with one tunable threshold** — explicitly not SSE connection counting, and not a manual do-not-disturb toggle.

## The crux: human turn vs. assistant activity

`TurnOrigin` is an explicit, exhaustive argument to `AttendanceTracker::record_turn`, and only `TurnOrigin::Human` advances the clock. Everything an instance produces on its own — replies, tool calls, tool results, hook fires, scheduled wakes — is `TurnOrigin::Assistant` and is a documented **no-op that writes nothing at all**. Passing `Assistant` is always safe, so a call site that is unsure has a correct default and cannot manufacture attendance.

That is what `last_activity_at` structurally cannot do, and it is what the two load-bearing tests pin:

- `assistant_turns_never_advance_the_clock` — 60 minutes of `Assistant` turns after one human turn; the clock still points at the human's turn and the instance is unattended.
- `assistant_only_activity_leaves_no_record_at_all` — an instance that has only ever worked alone has **no attendance record on disk**, so it reads `NeverAttended`.

## Threshold default: 15 minutes

`DEFAULT_UNATTENDED_AFTER = 15 min`, overridable with `TAGENT_UNATTENDED_AFTER_MINS` (whole minutes; blank, non-numeric, negative or zero input keeps the default rather than failing a boot — a zero threshold would make every instance permanently unattended, a footgun disguised as a setting).

**Why 15:** the threshold has to clear an ordinary interruption — reading a message, taking a call, refilling a coffee — because notifying a human who is about to look back at the screen is exactly the failure D3 exists to avoid. It also has to be short enough that a genuine walk-away is noticed inside one break rather than one workday. Fifteen minutes is the shortest span that comfortably clears the first without approaching the second, and it matches the idle convention users already expect from chat clients.

## Durability: durable, and that is correctness

State is **one small JSON record per instance** at `~/.trusty-agents/attendance/<instance>.json`, written through the existing `state_writer::atomic_write` (lock + tmp + rename), so a GUI, an `--api` sidecar and a REPL sharing that tree cannot tear each other's records.

The natural home was already durable, and durability here is not tidiness. An in-memory tracker would report a fresh process as never-attended — and never-attended counts as unattended — so restarting the API server while the owner was mid-conversation would hand #4653 a licence to notify a human who is demonstrably right there. `a_recorded_turn_survives_a_new_tracker` pins the restart case.

It lives in the **dotted, app-private** `~/.trusty-agents/` tree (alongside `sessions/`, `state/`), not the dotless user-browsable `~/trusty-agents/<instance>/` home that `assistants::home` deliberately hands to the user. This is machine state.

Recording is monotonic: an out-of-order or replayed turn never *rewinds* the clock, which would make an absent owner look present.

## Design decisions worth reviewing

- **`NeverAttended` counts as unattended.** The question is "is a human attending", and no evidence of a human is not evidence of a human. It is a distinct third state rather than a degenerate `Unattended` because a duration cannot express "never".
- **The threshold boundary is inclusive on the unattended side.** `idle >= threshold` is unattended, so "N minutes" reads the way an operator expects: N minutes of silence is enough.
- **A future timestamp (clock skew, an edited file) reads as attended.** Biasing skew toward silence is the safe direction — a missed notification costs a delay, a wrong one interrupts a human who is present.
- **A corrupt record is an error, not a silent `None`.** Treating corruption as never-attended would hand #4653 a licence to notify on a bad file.
- **`now` is a parameter on every query**, not a system-clock read. That is the injectable-clock seam this crate lacks, and it lets the boundary tests be exact instead of sleeping. Same reasoning for the injectable root (`attendance_root(base)`): `test_env` states outright that injection is the durable fix and `HOME_LOCK` is the legacy one, so no test here mutates `$HOME`.

## Wiring

The signal is fed at the four surfaces a person can actually address an assistant through, each recording the persona the message named (no selection = the `ctrl` concierge):

| Surface | Site |
|---|---|
| GUI / WebUI | `api/server/handlers.rs::submit_task` |
| REPL | `repl/dispatch.rs::attempt_forward` — one call covering all six dispatch arms |
| Telegram | `telegram/handlers.rs::handle_message`, past the pairing gate |
| Slack | `slack/handlers.rs::handle_message`, past the pairing and RBAC gates |

`run_pm_task_with_persona` was rejected as a single hook despite being the narrowest waist: delegation reaches it too, so hooking there would have counted an assistant's own delegated work as a human turn — the exact bug this issue exists to avoid.

The hook (`attendance::note_human_turn`) is **infallible by construction**: an unusable instance name, a missing home directory or an I/O error is logged at debug and swallowed. Attendance is a hint for #4653, never a reason to fail a turn a user is waiting on — and a swallowed failure biases the instance toward looking *less* attended, i.e. toward B2 staying quiet.

## Scope

Signal only. No `notify_owner` tool (#4653), no Telegram/Slack egress binding (#4654/#4655), no queueing (#4657). **`trusty-agents` gains no dependency on `trusty-mpm`** — the constraint documented at `crates/trusty-agents/src/tools/session_state/store.rs:1-12` holds; nothing here reaches into that crate.

## Gates

Crate-scoped, foreground, raw output in the PR thread. `cargo test -p trusty-agents --lib`: **3297 passed, 0 failed, 11 ignored** — including all 17 `attendance::tests::*`, verified by name in the output (the module is behind no feature flag, so no extra `--features` was needed). `cargo clippy -p trusty-agents --all-targets -- -D warnings`, `cargo fmt --check` and `bash scripts/check_line_cap.sh` all exit 0.

### DIFF
--- a/crates/trusty-agents/changelog.d/4652-unattended-detection.md
+++ b/crates/trusty-agents/changelog.d/4652-unattended-detection.md
@@ -37,12 +37,17 @@ Added
   `~/.trusty-agents/attendance/`, written through the existing lock+tmp+rename
   `state_writer` so a GUI, an `--api` sidecar and a REPL sharing that tree
   cannot tear each other's records. Recording is monotonic — an out-of-order
-  turn never rewinds the clock.
+  turn never rewinds the clock — and the guard is evaluated under the SAME held
+  lock as the write (`state_writer::atomic_update`, new in this change), so two
+  processes recording near-simultaneous turns cannot both read the old value and
+  let the loser's older write land last. Pinned by
+  `concurrent_writers_never_rewind_the_clock`, which fails against a
+  read-then-write guard.
 
-  Recorded at the four surfaces a person can actually address an assistant
-  through, each against the persona the message named: `POST /api/task`, the
-  REPL's `attempt_forward` (one call covering all six of its dispatch arms),
-  Telegram past the pairing gate, and Slack past the pairing and RBAC gates. The
+  Recorded at every surface a person can actually address an assistant through,
+  each against the persona the message named: `POST /api/task`, the REPL's
+  `attempt_forward` (one call covering all six of its dispatch arms), and BOTH
+  inbound paths on Telegram and Slack — free text and slash commands alike. The
   hook is infallible by construction — an unusable instance name, a missing home
   directory or an I/O error is logged at debug and swallowed, because attendance
   is a hint for #4653 and never a reason to fail a turn a user is waiting on.
@@ -50,9 +55,27 @@ Added
   Signal only: nothing here sends or queues, and `trusty-agents` gains no
   dependency on `trusty-mpm`.
 
-  Telegram's `/switch <persona>` intercept returns early from `handle_message`,
-  so a paired human switching persona was never recorded as a turn (caught in
-  review of [#4683](https://github.com/bobmatnyc/trusty-tools/pull/4683)). The
-  hook now runs before the intercept, fused with the routing decision so the two
-  cannot drift apart again — pinned by
-  `switch_command_still_records_a_human_turn`.
+  Three attendance gaps were caught in review of
+  [#4683](https://github.com/bobmatnyc/trusty-tools/pull/4683), all the same
+  shape — a live entry point that never advanced the clock, so a human who was
+  demonstrably present read as unattended after fifteen minutes:
+
+  - Telegram's `/switch <persona>` intercept returns early from
+    `handle_message`, so a paired human switching persona was never recorded.
+    The hook now runs before the intercept, fused with the routing decision so
+    the two cannot drift apart again — pinned by
+    `switch_command_still_records_a_human_turn`.
+  - Telegram's `handle_command` (`/start`, `/pair`, `/help`, `/connect`,
+    `/clear`, `/status`) and Slack's (`/slack-start`, `/slack-pair`,
+    `/slack-connect`, `/slack-clear`, `/slack-switch`, `/slack-status`)
+    recorded nothing at all, so a human polling `/status` every few minutes
+    while a long task ran was invisible. Both now record through one shared
+    `attendance::note_command_turn_in`, behind the same authentication each
+    transport's message path already records behind: paired on Telegram, paired
+    AND a known RBAC identity on Slack — so an unpaired chat, or an unknown
+    Slack user posting in a paired channel, still cannot manufacture attendance
+    for someone else's assistant. Pinned by
+    `paired_slash_command_records_a_human_turn`,
+    `unpaired_slash_command_records_nothing`,
+    `unknown_rbac_user_cannot_manufacture_attendance` and, end to end,
+    `a_command_only_session_stays_attended`.
--- a/crates/trusty-agents/changelog.d/4652-unattended-detection.md
+++ b/crates/trusty-agents/changelog.d/4652-unattended-detection.md
@@ -1,0 +1,51 @@
+Added
+
+- **`attendance` — "is a human currently attending this assistant instance?"**
+  (closes [#4652](https://github.com/bobmatnyc/trusty-tools/issues/4652); part
+  of epic [#4646](https://github.com/bobmatnyc/trusty-tools/issues/4646), owner
+  decision **D3**). Nothing could answer that question, and the `notify_owner`
+  tool ([#4653](https://github.com/bobmatnyc/trusty-tools/issues/4653)) cannot
+  exist without an answer. The new module exposes `AttendanceTracker`,
+  `Attendance` and `is_unattended()` for #4653 to gate on. Per D3 the mechanism
+  is a last-human-turn timeout with one tunable threshold — explicitly not SSE
+  connection counting and not a manual do-not-disturb toggle.
+
+  This is new work rather than a reuse of `SessionRecord.last_activity_at`
+  because that field advances on the assistant's own tool and hook activity
+  exactly as it does on a human typing, so an assistant grinding through a long
+  solo task reads as attended forever. Here the origin of a turn is an explicit
+  `TurnOrigin` argument and only `TurnOrigin::Human` advances the clock;
+  assistant activity writes nothing at all, which
+  `assistant_turns_never_advance_the_clock` and
+  `assistant_only_activity_leaves_no_record_at_all` pin. `Attendance` carries a
+  distinct `NeverAttended` state — no human turn has ever been recorded — which
+  counts as unattended, and the threshold boundary is inclusive on the
+  unattended side, so "N minutes" means N minutes of silence is enough. A
+  timestamp in the future (clock skew) reads as attended, biasing toward silence
+  rather than toward interrupting someone who is present.
+
+  **The threshold defaults to 15 minutes**, overridable with
+  `TAGENT_UNATTENDED_AFTER_MINS` (whole minutes; blank, non-numeric or zero
+  input keeps the default rather than failing a boot). Long enough to clear an
+  ordinary interruption, short enough that a real walk-away is noticed inside
+  one break.
+
+  **Durable**, and that is correctness rather than tidiness: an in-memory
+  tracker would report a fresh process as never-attended, so restarting the API
+  server mid-conversation would hand #4653 a licence to notify a human who is
+  demonstrably present. State is one small JSON record per instance under
+  `~/.trusty-agents/attendance/`, written through the existing lock+tmp+rename
+  `state_writer` so a GUI, an `--api` sidecar and a REPL sharing that tree
+  cannot tear each other's records. Recording is monotonic — an out-of-order
+  turn never rewinds the clock.
+
+  Recorded at the four surfaces a person can actually address an assistant
+  through, each against the persona the message named: `POST /api/task`, the
+  REPL's `attempt_forward` (one call covering all six of its dispatch arms),
+  Telegram past the pairing gate, and Slack past the pairing and RBAC gates. The
+  hook is infallible by construction — an unusable instance name, a missing home
+  directory or an I/O error is logged at debug and swallowed, because attendance
+  is a hint for #4653 and never a reason to fail a turn a user is waiting on.
+
+  Signal only: nothing here sends or queues, and `trusty-agents` gains no
+  dependency on `trusty-mpm`.
--- a/crates/trusty-agents/changelog.d/4652-unattended-detection.md
+++ b/crates/trusty-agents/changelog.d/4652-unattended-detection.md
@@ -49,3 +49,10 @@ Added
 
   Signal only: nothing here sends or queues, and `trusty-agents` gains no
   dependency on `trusty-mpm`.
+
+  Telegram's `/switch <persona>` intercept returns early from `handle_message`,
+  so a paired human switching persona was never recorded as a turn (caught in
+  review of [#4683](https://github.com/bobmatnyc/trusty-tools/pull/4683)). The
+  hook now runs before the intercept, fused with the routing decision so the two
+  cannot drift apart again — pinned by
+  `switch_command_still_records_a_human_turn`.
--- a/crates/trusty-agents/src/api/server/handlers.rs
+++ b/crates/trusty-agents/src/api/server/handlers.rs
@@ -320,6 +320,13 @@ pub(super) async fn submit_task(
     let placeholder = PmResponse::running(&id);
     state.upsert(id.clone(), placeholder).await;
 
+    // #4652: this request is a person typing in the GUI/WebUI, which is the
+    // one thing `SessionRecord.last_activity_at` cannot distinguish from the
+    // assistant's own tool calls. Record it as attendance for the instance the
+    // human addressed (no roster selection = the `ctrl` concierge).
+    // Infallible: a failure is logged and swallowed, never surfaced here.
+    crate::attendance::note_human_turn(agent_override(&req).as_deref().unwrap_or("ctrl"));
+
     // #192 Phase B: announce the new session immediately on the event bus so
     // SSE subscribers (Sidebar task list, ChatView session bootstrap) update
     // before the child subprocess has even spawned.
--- a/crates/trusty-agents/src/attendance/mod.rs
+++ b/crates/trusty-agents/src/attendance/mod.rs
@@ -63,8 +63,8 @@ mod tracker;
 mod tests;
 
 pub use tracker::{
-    AttendanceTracker, attendance_root, default_attendance_root, note_human_turn,
-    note_human_turn_in,
+    AttendanceTracker, attendance_root, default_attendance_root, note_command_turn_in,
+    note_human_turn, note_human_turn_in,
 };
 
 /// How long after the last human turn an instance counts as unattended.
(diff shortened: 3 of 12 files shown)


bobmatnyc:
## Gate evidence (raw, crate-scoped, foreground)

### `cargo test -p trusty-agents --lib attendance`

The module is behind no feature flag, so the new tests really ran — all 17 by name:

```
running 17 tests
test attendance::tests::default_root_is_under_the_app_state_tree ... ok
test attendance::tests::parse_threshold_rejects_junk_and_zero ... ok
test attendance::tests::parse_threshold_accepts_whole_minutes ... ok
test attendance::tests::default_threshold_is_fifteen_minutes ... ok
test attendance::tests::assistant_only_activity_leaves_no_record_at_all ... ok
test attendance::tests::never_attended_counts_as_unattended ... ok
test attendance::tests::note_human_turn_swallows_an_unusable_instance_name ... ok
test attendance::tests::unreadable_record_is_an_error_not_a_silent_never ... ok
test attendance::tests::a_recorded_turn_survives_a_new_tracker ... ok
test attendance::tests::threshold_boundary_is_unattended_at_exactly_the_threshold ... ok
test attendance::tests::record_path_is_one_file_per_instance ... ok
test attendance::tests::fresh_human_turn_is_attended ... ok
test attendance::tests::an_older_human_turn_never_rewinds_the_clock ... ok
test attendance::tests::note_human_turn_records_a_turn_and_is_infallible ... ok
test attendance::tests::assistant_turns_never_advance_the_clock ... ok
test attendance::tests::a_future_timestamp_reads_as_attended ... ok
test attendance::tests::past_threshold_is_unattended ... ok
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 3291 filtered out; finished in 0.05s
```

### `cargo test -p trusty-agents --lib` (full crate)

```
test result: ok. 3297 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 28.89s
```

### `cargo clippy -p trusty-agents --all-targets -- -D warnings`

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 22.62s
```

### `cargo fmt --check`

```
fmt exit: 0
```

### `bash scripts/check_line_cap.sh`

```
line-cap: measured 3600 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

bobmatnyc:
## Adversarial review — PR #4683

**Verdict: WARN** (zero CRITICAL, one HIGH finding)

Mergeability: `MERGEABLE`, no conflicts. `mergeStateStatus` is `BEHIND` by 2 commits on `main` (a docs-only INSTRUCTIONS.md commit and a `trusty-mpm` 1.3.3 release commit) — neither touches `trusty-agents`, and `git merge-tree` shows a clean fast-forward-style merge with zero conflict markers. Not a blocker; rebase is a formality.

## Findings

| HIGH | `crates/trusty-agents/src/telegram/handlers.rs` | 61 (`handle_command`) | The attendance hook was added to `handle_message` (free text) and its `/switch` early-return, but never to `handle_command` — the dispatch path for `/start`, `/pair`, `/help`, `/connect`, `/clear`, `/status`. A paired human who drives the bot only through recognized slash commands (e.g. polling `/status` every few minutes while a long task runs) never advances the last-human-turn clock. After 15 minutes they read as `Unattended` despite being demonstrably present — the exact false positive this feature exists to prevent. | Call `crate::attendance::note_human_turn` (or a `note_turn_and_is_switch`-style fused helper) at the top of `handle_command`, past the existing pairing gate — same pattern already applied to `handle_message`'s `/switch` fix in commit 08605c24. | Fix here |
| HIGH | `crates/trusty-agents/src/slack/handlers.rs` | 47 (`handle_command`) | Identical gap on the Slack surface: `/slack-start`, `/slack-pair`, `/slack-connect`, `/slack-clear`, `/slack-switch`, `/slack-status` never call `note_human_turn`. Only `handle_message` (line 374) does. Same false-positive exposure as the Telegram case above (e.g. a human polling `/slack-status`). | Same fix, mirrored in `handle_command`, past the pairing/RBAC gate. | Fix here |
| MEDIUM | `crates/trusty-agents/src/attendance/tracker.rs` | 139–163 (`record_turn`) | The monotonic guard is read-then-write, not atomic: `last_human_turn(instance)` is read, compared, and only then written via `state_writer::atomic_write` — but the lock is held only during the write, not across the read+compare+write. Two processes recording near-simultaneous human turns for the same instance (the module's own stated multi-process scenario: "a GUI, an `--api` sidecar and a REPL sharing that tree") can interleave so a later reader's write clobbers an already-persisted later timestamp with an earlier one, breaking the documented "recording is monotonic" invariant. Practical impact is bounded — both racing timestamps are "now", so the drift is milliseconds and doesn't move the 15-minute threshold — but the only test covering monotonicity (`an_older_human_turn_never_rewinds_the_clock`) is single-threaded/sequential, not concurrent, so the invariant is asserted in prose and tested only in the case that can't actually race. | Either extend `state_writer` with a lock-held read-modify-write helper (hold the exclusive lock across read+compare+write), or add an explicit concurrent-writer regression test and accept/document the current bound. | Promote |

## Notes

- Boundary, never-attended, and false-positive-from-assistant-activity tests are all present and correctly targeted: `threshold_boundary_is_unattended_at_exactly_the_threshold` (inclusive boundary, exercised one second short and exactly at threshold), `never_attended_counts_as_unattended` / `assistant_only_activity_leaves_no_record_at_all` (no human turn ever), `assistant_turns_never_advance_the_clock` (60 minutes of assistant-only activity stays pinned to the human turn — this is the core false-positive guard the module exists for, and it's solid). `a_future_timestamp_reads_as_attended` correctly pins the clock-skew-biases-toward-silence decision. No missing tests at the unit level for the shipped surface.
- The `TurnOrigin` design (only `Human` advances the clock, `Assistant` is a documented no-op) is the right structural fix for what `SessionRecord.last_activity_at` couldn't do, and the four wired surfaces are each gated correctly behind their pairing/RBAC checks before recording — so an unauthenticated sender can't manufacture attendance.
- The `/switch` early-return bug (fixed in commit 08605c24, same PR) was itself found by review and is now regression-tested (`switch_command_still_records_a_human_turn`, `plain_message_records_a_human_turn_and_is_not_a_switch`) — good catch, but it shows the "attendance hook must be wired at every return path" risk was already live once; the two `handle_command` gaps above are the same risk class recurring in the sibling dispatch function that wasn't touched.
- Changelog fragment: present at `crates/trusty-agents/changelog.d/4652-unattended-detection.md`, directly in `changelog.d/` (not nested), line 1 is `Added` — a single valid category. Compliant.
- No scope creep found — diff is tightly bounded to the attendance module, its four (nominally four; actually six counting the two `handle_command` paths above) wiring sites, `lib.rs` module registration, and the changelog fragment.

## Rust Test Ladder

This sits closest to **rung 3** (localized behavior inside one crate — no cross-crate touches, `trusty-agents` gains no new dependency) but the module's own framing ("durable, and that is correctness... a GUI, an `--api` sidecar and a REPL sharing that tree cannot tear each other's records") is explicitly a persistence + multi-process-concurrency claim, which is rung 5 territory per the ladder's own criteria list. The PR body cites the rung-3 command (`cargo test -p trusty-agents --lib`, `cargo clippy -p trusty-agents --all-targets -- -D warnings`, `cargo fmt --check`, `check_line_cap.sh` — all reported green), but CI's actual gates run wider than that: `.github/workflows/ci.yml` runs `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace` (both green per `gh pr checks 4683`), which already covers rung 4's breadth. What rung 5 would still ask for and is genuinely missing is a **concurrency test** for the exact race in the MEDIUM finding above — the ladder's "failure-path and concurrency tests" requirement for persistence/process-lifecycle changes. Given the bounded practical impact of that race, I'd call the gates that did run adequate to ship, with the concurrency test as the one gap worth closing before or shortly after merge.

bobmatnyc:
## Review fixes — all three findings addressed (`4b6056a1`)

### HIGH ×2 — `handle_command` never recorded attendance

Both transports' slash-command dispatch now records a human turn through one shared `attendance::note_command_turn_in`, behind the same authentication each transport's message path already records behind:

| Transport | Gate |
|---|---|
| Telegram `handle_command` | channel is paired |
| Slack `handle_command` (`handlers::note_command_turn`) | channel is paired **AND** sender resolves to a known RBAC identity |

The RBAC half matters: Slack pairing is per-*channel*, so pairing alone would let an unknown user in a paired channel manufacture attendance for someone else's assistant. `handle_message` already refuses to record for an unknown user (Virtual CTO reply, early return); the command path now agrees. Neither gate changes whether the command itself proceeds.

**Every entry point audited on both transports** — the point of the finding was that this bug class had already recurred once:

| Transport | Entry point | Before | After |
|---|---|---|---|
| Telegram | dptree branch 1 → `handle_command` (`/start` `/pair` `/help` `/connect` `/clear` `/status`) | ✗ none | ✓ paired |
| Telegram | dptree branch 2 → `handle_message` (unknown slash: `/switch` `/cost` `/model`) | ✓ (`08605c24`) | ✓ |
| Telegram | dptree branch 3 → `handle_message` (plain text) | ✓ | ✓ |
| Slack | `slash_commands` envelope → `handle_command` (6 commands) | ✗ none | ✓ paired + RBAC |
| Slack | `events_api` message envelope → `handle_message` | ✓ | ✓ |

Unchanged and already correct: `POST /api/task` → `submit_task`, REPL → `attempt_forward` (one call covering all six dispatch arms).

### MEDIUM — monotonic guard was read-then-write

New `state_writer::atomic_update` holds the exclusive lock across read + compare + write. `record_turn` uses it, so two processes can no longer both observe the old timestamp and let the loser's older write land last.

**Fail-before / pass-after, `concurrent_writers_never_rewind_the_clock`** (8 threads × 25 turns; worker 0 publishes the newest instant first, everyone else only offers older ones):

Before — `record_turn` on the read-then-write guard:

```
thread 'attendance::tests::concurrent_writers_never_rewind_the_clock' panicked at
crates/trusty-agents/src/attendance/tests.rs:434:5:
assertion `left == right` failed: a concurrent older turn rewound the clock — the
monotonic guard is not atomic with the write
  left: Some(2026-08-03T12:03:19Z)
 right: Some(2026-08-03T13:00:00Z)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3318 filtered out
```

Red **10/10** consecutive runs before the fix; green **10/10** after.

After:

```
test attendance::tests::concurrent_writers_never_rewind_the_clock ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3318 filtered out
```

### HIGH findings — fail-before evidence, stated plainly

The new tests run against the unfixed production tree (all five production files restored to `e4affa0f`, test files retained) fail to compile, naming exactly the surfaces that did not exist:

```
error[E0425]: cannot find function `note_command_turn` in module `super::handlers`   (×3)
error[E0425]: cannot find function `note_command_turn_in` in module `crate::attendance` (×2)
error[E0425]: cannot find function `note_command_turn_in` in this scope
error: could not compile `trusty-agents` (lib test) due to 6 previous errors
```

Being straight about the limit: `handle_command` itself is not unit-testable — teloxide's `Bot`/`Message` and Slack's socket envelope both need live network, and the production hook resolves `default_attendance_root()` from `$HOME` rather than an injected root. The tests therefore target the helper the handlers call, not the handlers. They pin the *contract* (paired records, unpaired does not, unknown-RBAC does not) and the *outcome* (`a_command_only_session_stays_attended`: four `/status` polls five minutes apart read as `Attended { idle_for: 5m }` twenty minutes after the first, where the same window with no recorded turn is unattended). They would not catch a future removal of the call site.

### New tests

```
running 11 tests
test slack::tests::unknown_rbac_user_cannot_manufacture_attendance ... ok
test slack::tests::unpaired_slash_command_records_nothing ... ok
test telegram::tests::unpaired_slash_command_records_nothing ... ok
test telegram::tests::switch_command_still_records_a_human_turn ... ok
test telegram::tests::plain_message_records_a_human_turn_and_is_not_a_switch ... ok
test state_writer::tests::atomic_update_skips_the_write_when_the_closure_declines ... ok
test slack::tests::paired_slash_command_records_a_human_turn ... ok
test telegram::tests::paired_slash_command_records_a_human_turn ... ok
test attendance::tests::a_command_only_session_stays_attended ... ok
test attendance::tests::concurrent_writers_never_rewind_the_clock ... ok
test state_writer::tests::atomic_update_serializes_read_modify_write ... ok

test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 3308 filtered out; finished in 2.10s
```

No existing test was weakened, deleted, `#[ignore]`d, `cfg`-gated or excluded.

### Gates (rung 5 — persistence + cross-transport), re-run after rebasing onto `fbf487ac`

```
cargo fmt --check                                    exit=0
cargo clippy -p trusty-agents --all-targets -- -D warnings   exit=0
cargo test -p trusty-agents                          exit=0
  test result: ok. 3308 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out
  (+ 9 further binaries, all "ok. 0 failed")
bash scripts/check_sld.sh
  sld-lint: scanned 54 spec doc(s) + 3033 code file(s); 0 error(s), 0 warning(s)
  test-pointers: resolved 20419 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
  OK   trusty-agents: changelog.d fragment present and valid
```

Changelog fragment `crates/trusty-agents/changelog.d/4652-unattended-detection.md` updated in place — still one file, directly in `changelog.d/`, line 1 still the single category `Added`.

Issue #3659, opened 2026-07-22
trusty-search: recurring redb corpus-load panic during concurrent warm-boot (page_manager.rs:243)


bobmatnyc:
**Component:** trusty-search
**Versions affected:** ~0.29 through 0.36.1 (recurring across the range)
**Observed:** prod serving box `i-0076`, recurring since at least early June 2026

## Problem

On warm boot, a trusty-search corpus-load task **panics** while opening the redb corpus for the `main` index during concurrent warm-boot, at `page_manager.rs:243`. It surfaces as an `incompatible-format` open failure that then panics rather than being handled. It usually self-heals via the "trying registered migrations" auto-retry, but it adds warm-boot lag and reports a spurious degraded status in the interim.

This is the crash-safety gap that #702/#703 ("Upgrade redb 2.6 → 4.x + **graceful old-format handling** — never panic, never report healthy-empty") was supposed to close, and it overlaps #697 (cross-version index open failure). Both are CLOSED COMPLETED, yet the panic still recurs on 0.36.1 — so either the panic-safe open path regressed, or the concurrent warm-boot path was never covered by it.

## Evidence

Log signature seen across many process restarts on `i-0076`:

```
redb corpus load task panicked
FAILED to open redb corpus for 'main' ... incompatible-format
thread '...' panicked at page_manager.rs:243
```

- Health reports `indexes_corpus_failed: N` on warm boot while this is happening.
- **Recurring, not one-off:** observed across process restarts spanning **Jun 7–13** and again after the **2026-07-21** restarts, across **multiple binary versions (~0.29 through 0.36.1)**.
- **Concurrency-correlated:** it presents specifically during concurrent warm-boot corpus load (multiple corpus-load tasks racing at startup) — a corpus-**load task** panics, distinct from a single clean open.
- **Self-heals:** the daemon's "trying registered migrations" auto-retry generally recovers the index on a subsequent attempt, so it ends up serving — but only after extra warm-boot latency and a transient degraded/`indexes_corpus_failed` window.

## Impact

- Warm-boot latency and a transient degraded status on every restart that trips it — noise that masks genuine corpus failures.
- A panic in the load task is a crash-safety hole: if the migration retry ever doesn't cover a given path, the corpus stays failed instead of degrading cleanly.
- Undermines the guarantee #702 set out to provide ("never panic on old/incompatible format").

## Suggested direction

[...]

PR #4718, opened 2026-08-03
fix(trusty-search): distinguish corpus-open failure states, fail loudly on broken indexes, give the reaper a terminal path

### LINKED ISSUE #4087: trusty-search: corpus-failed indexes return silently-empty 200s, and timeout-skipped indexes are dropped rather than parked cold

### LINKED ISSUE #4095: orphan-reaper defers forever on ambiguous root candidates, silently accumulating permanent registry debris
## Problem

The trusty-search orphan-reaper defers cleanup of indexes when root resolution yields ambiguous candidates, and never revisits them. This creates permanent registry debris that degrades `search_health` indefinitely, with no warning or visibility.

## Evidence

A live investigation of a developer machine's real trusty-search daemon (2026-07-27) found the reaper's warm-boot log showing why 11 stale entries were not reaped. Example for index `foo`:

```
no longer exists and 2 ambiguous candidates found (manual `trusty-search index <path>` required)
```

### Reaper behavior observed

- **Healthy reaping**: the reaper successfully reaped 38 entries at 00:58:40Z and 25 at 01:58:40Z today
- **Healthy long-term pattern**: across log history since 2026-06-03 (12,822 reaper log lines), it has cleared thousands of ephemeral `.tmp*`, `tm-test-*`, `tcode-bakeoff-*` ids
- **Ambiguity deferral**: when root resolution against `roots.toml`'s tracked-root history yields multiple ambiguous candidates, the reaper defers and never revisits
- **No age-based fallback**: the ambiguous entries are never reapped, even after days of accumulation
- **No warning**: no WARN log is emitted, so the debris is silently accumulating

## Impact

1. **Permanent registry pollution**: once an entry triggers ambiguity, it stays in the registry indefinitely
2. **Health reporting degraded**: these stale entries keep `search_health` reporting `status: degraded` / `warm_boot_degraded: true`, masking real health signals
3. **Cascade trigger**: integration test isolation issues (see related issue #4094) create the collisions that trigger this deferral

## Solution Direction

Implement an eventual fallback for ambiguous entries:
- **Option A**: age-based force-reap after N days (e.g., 7+ days with no resolution)
- **Option B**: at minimum, emit a WARN log so the debris becomes visible instead of silently accumulating

Additionally, consider revisiting the root resolution logic to reduce false ambiguities.

## Related

- Issue #4094: integration tests registering into real daemon registry (source of collisions that trigger this deferral)

### LINKED ISSUE #4333: trusty-search: misleading 'corrupted format' status conflates timeout/contention with genuine corruption
## Summary

The status string **"redb corpus could not be opened (incompatible or corrupted format)"** conflates two entirely different failure modes:
1. A transient **timeout/contention** (30-second redb open timeout while another opener holds the lock)
2. A genuine **format/schema-version mismatch** or actual corruption

This distinction matters operationally: timeout + contention are transient and self-heal without intervention, while corruption requires reindex/rebuild. Conflating them has historically led to **destructive remediation** (reindex, rebuild) being applied to transient problems — turning a temporary outage into permanent data loss.

## Evidence from This Session

**Index:** `cto-duetto` (root `/Users/masa/trusty-mpm-projects/bob-duetto/cto`)  
**Timestamp:** 2026-07-29 post-recovery  
**Daemon:** trusty-search 0.39.1

**Observation:** 
- Index was VERIFIED HEALTHY with **201,206 chunks** at ~19:00 UTC (re-registered during registry recovery #4317, corpus confirmed **2.16 GB**, HNSW **339 MB**).
- By 21:30 UTC same boot, `index_status` reports **FAILED** with corrupted-format message, showing **122 chunks** (from log examination).
- No restart, no reindex, no user action between states — only ambient daemon activity.
- Daemon health: `warm_boot_degraded: true`, `indexes_corpus_failed: 1`, `indexes_skipped_timeout: 18` **← note the timeout count**.

**The misleading part:** the daemon that simultaneously reports `indexes_corpus_failed: 1` *with corruption wording* is also reporting `indexes_skipped_timeout: 18` — meaning it is **actively experiencing timeout/contention on other indexes in the same boot**. The same corpus-open failure signature (#3659) occurred under timeout/contention, not actual corruption.

## Historical Pattern

This is the **second confirmed occurrence** in this project of the same misleading conflation:

1. **Prior incident (2026-07-27, recorded in project memory):** A large index reported "corrupted format" after a restart. A recovery attempt (also recorded) involved reindex/rebuild, which cost the index. Later analysis (per project memory) revealed the real cause was the **30-second redb open timeout under warm-boot contention** — i.e., #3659's problem exactly. No actual corruption.

2. **This incident (2026-07-29):** Same message, same warm-boot context (`warm_boot_degraded: true`), same concurrent timeout count, but now with direct before/after verification that the underlying data was intact (201k → 122 chunk regression is file metadata, not data loss).

## The Real Defect

**Issue #3659 (panic during concurrent warm-boot)** was fixed to prevent panics, but the **status string** was never updated to distinguish the underlying causes. Today, both paths — "I timed out waiting for the corpus opener" and "the corpus format is genuinely incompatible" — **report the same status message**. A consumer (or a person reading logs) cannot tell which problem they have.

## Suggested Direction

**Ask:** The failure surface should distinguish:
- `open_timeout` / `contention` — transient, self-heals, retry shortly
- `format_mismatch` / `schema_version` — permanent, requires reindex
- `genuine_corruption` — data loss, requires backup/rebuild

Not prescribing implementation, but the status message should not report corruption when the underlying error is timeout. And when corpus-open fails transiently, it should not imply data loss.

**Separately:** The **122-chunk count** is itself suspect. When a corpus fails to open, reporting a nonzero partial-looking number (201,206 → 122) is misleading — it reads as catastrophic data loss when the corpus may be entirely intact and simply unopened. Consider reporting `chunks: null` or `chunks: unknown` when corpus-open fails, rather than a partial count.

## Why This Matters

1. **Operational:** Distinguishing transient from permanent allows appropriate response (wait vs. rebuild).
2. **Safety:** The "corrupted format" wording invites destructive remediation. Conflating it with timeout is how a transient outage becomes permanent in practice.
3. **Observability:** Status strings are read by humans, automation, and dashboards — precision matters.

## Status & Process

- No restart, no reindex, no deregistration performed — these are intentionally held to preserve diagnostic state.
- A read-only diagnosis is in progress; findings will be added as a comment.
- Do NOT reach for reindex as a first move on any future occurrence.

## Cross-References

- #3659 — redb corpus-load panic during concurrent warm-boot (fixed to panic-safe, but status wording unchanged)
- #4317 — Bulk registry sweep that recovered cto-duetto hours before this regression
- #4315 — per-connection health/probe API (would make contention vs. corruption legible)
- #4087 — corpus-failed indexes returning silently-empty results

bobmatnyc:
## Summary

Three trusty-search daemon defects that share one shape: **the daemon reports a healthy or misleading state while an index is actually broken or gone.** Each fix replaces a silent or misleading state with a loud, recoverable one.

Closes #4333
Closes #4087
Closes #4095

---

## #4333 — `corrupted format` conflated timeout/contention with real corruption

`derive_warm_boot_stages` emitted ONE hardcoded string for every `corpus_open_failed` case:

> redb corpus could not be opened (incompatible or corrupted format); run `trusty-search index <path> --force` to rebuild from source

It named the only permanent cause and prescribed the only **destructive** remedy — for a 30 s open timeout under warm-boot contention, a stale file lock, and a permission denial alike. On 2026-07-29 the daemon reported it for `cto-duetto` while simultaneously reporting `indexes_skipped_timeout: 18`, with the corpus verified intact at 201,206 chunks. The same conflation had already cost one index to a `--force` rebuild two days earlier.

**Fix.** `CorpusOpenFailure` (`core/corpus/open_failure.rs`) classifies the failure into `open_timeout` / `contention` / `format_incompatible` / `unclassified`, by **typed downcast, never string matching** — the timeout and wedge paths now return typed marker errors (`CorpusOpenTimeout`, `CorpusOpenWedged`) because they produce no redb error at all, which is exactly why they used to fall through to the corruption wording. Each kind carries its own operator guidance:

- transient kinds state the corpus is **presumed INTACT** and say `DO NOT reindex` / `DO NOT run --force` in as many words;
- `unclassified` (I/O, permissions, unresolvable path) refuses to claim corruption and refuses to recommend a rebuild until the cause is known;
- **`format_incompatible` keeps the rebuild instruction** — the fix must not swing to the opposite error of never recommending one. Pinned by `format_incompatible_keeps_the_rebuild_instruction`.

Also from the ticket: `GET /indexes/:id/status` reports `chunk_count: null` while the corpus is unopened, instead of the in-memory fallback that rendered 122 for an index holding 201,206 chunks and read as catastrophic data loss. A new `corpus_open_failure` object exposes `{kind, transient, reason}`.

---

### Design judgment: fail loudly, and count what you skip

A corpus-failed index is **registered**, so `search` reached its handle and returned `HTTP 200` with `results: []`. The caller could not distinguish "nothing matched" from "this index is entirely broken". Three live indexes were serving that way.

Choice made: **503, not 200-with-a-flag and not 500.** A flag on a 200 preserves the failure mode for every caller that doesn't read it — the bug is that the outage was easy to ignore, so the response must be one a caller has to handle. 503 rather than 500 because the state is recoverable *by construction*: `clear_corpus_open_failure` lifts it on the next successful open, which for the transient #4333 kinds can happen on the very next restart with no operator action. The response carries the #4333 classification so a caller learns whether to retry or escalate.

`POST /search` fan-out **excludes and counts** these indexes in a new `corpus_failed_indexes_skipped`. Exclusion alone would have been the same silence relocated — an empty lane fused into the results while the payload still claimed a complete fan-out. The count is the load-bearing half.

### Timeout-skipped indexes are parked, not dropped

A warm-boot restore timeout tallied a counter and did nothing else: the entry reached neither the registry nor the cold store, so the index ceased to exist for the rest of the boot (`list_indexes` omitted it, `search` 404'd it, `search_all` couldn't even count it). Eleven indexes were lost this way on a live daemon, recoverable only by a restart — which is why the observed payload showed `indexes_lazy: 0` alongside `indexes_skipped_timeout: 11`.

A timeout is the *most* transient failure the daemon has — the restore was slow, not wrong — so dropping it is the least appropriate response available. `ColdIndexStore::park_timed_out` parks the entry, making it recoverable through machinery that already exists: `get_or_load_index` lazy-loads it on the first query, and `cold_store.len()` folds it into `cold_indexes_skipped`, which also closes suggestion 3 in the ticket. Warm-boot now reports `indexes_lazy` from the live cold-store size rather than the pre-loop split count, so parked entries stop vanishing from every counter. No inline retry: the abandoned blocking thread may still hold the redb file, and the lazy-load path already handles that via the `DatabaseAlreadyOpen` retry and the #3659 open gate.

---

### Design judgment: a terminal path that never deletes data

The relocation scan logged `… and N ambiguous candidates found (manual trusty-search index <path> required)` and returned. Two problems: it never revisited, so the entry became permanent registry debris pinning `search_health` to `degraded`; and it logged at **WARN**, which in this daemon reaches no diagnostic surface at all (`BugCaptureLayer` drops anything below ERROR, so `errors.jsonl` / `list_recent_errors` / `tm doctor` never saw it). Eleven entries accumulated continuously since 2026-06-03.

The brief is explicit that silent data deletion is a worse failure than the debris it cleans up, and I agree — so the terminal action is deliberately **narrow**:

- **What it removes:** the `indexes.toml` row and the `roots.toml` entry.
- **What it never touches:** on-disk index data. `remove_index_data_dir` is not called, and no `AmbiguousRootAction` variant can delete data — pinned by `ambiguous_root_actions_never_include_data_deletion`, so a future variant that did would have to touch that test and justify itself.
- Recovery stays exactly the documented fix: `trusty-search index <path>`.

That asymmetry follows from *why* the deferral exists: the daemon genuinely cannot tell which of N candidate roots owns the index. Under that uncertainty, destroying a corpus we cannot identify is unjustifiable, while dropping a registration we cannot resolve is fully reversible.

Both required guards from the brief are present:

- **Explicit age threshold** — `TRUSTY_AMBIGUOUS_ROOT_GRACE_SECS`, default **7 days**, `0` disables the terminal path entirely (defer forever, still warned). Seven days is far longer than any legitimate relocation and far shorter than the 8-week accumulation observed. The clock is persisted in `indexes.toml` (`ambiguous_root_since_unix`), not held in memory, because the incident spanned many daemon restarts. It is cleared when the entry resolves — via auto-relink or an explicit `PATCH` relocation — so a transient ambiguity can't age into a reap of a since-healthy registration.
- **Logged warning** — the first observation and the terminal reap both log at **ERROR** (so they reach the surfaces an operator reads); the in-window wait stays WARN so a 7-day grace doesn't spam the error buffer once per boot. The terminal message states in capitals that on-disk data was not deleted and names the recovery command.

Clock skew is handled: `saturating_sub` means a backwards NTP step reads as age 0 and waits, rather than reaping instantly (`classify_ambiguous_root_tolerates_clock_skew`).

---

## Tests — 23 new, one per failure-state distinction

A regression in any of these is invisible without a test: an empty 200 looks like no matches, a dropped index looks like an unregistered one, and a corruption string for a timeout looks like corruption.

| Area | Tests |
|---|---|
| #4333 classification | `classify_timeout_marker_is_transient`, `classify_wedge_marker_is_contention`, `classify_database_already_open_is_contention`, `classify_upgrade_required_is_format_incompatible`, `classify_corrupted_storage_is_format_incompatible`, `classify_io_error_is_not_format_incompatible`, `classify_unknown_error_is_unclassified`, `transient_kinds_never_recommend_rebuild`, `unclassified_kind_does_not_claim_corruption`, `format_incompatible_keeps_the_rebuild_instruction`, `labels_are_distinct` |
| #4087 query-time | `search_against_corpus_failed_index_returns_503_not_empty_200`, `corpus_failure_response_distinguishes_permanent_from_transient`, `healthy_index_is_unaffected_by_the_corpus_failure_guard`, `global_search_excludes_and_counts_corpus_failed_indexes`, `timed_out_entry_is_parked_in_cold_store` |
| #4095 reaper | `classify_ambiguous_root_stamps_on_first_sight`, `classify_ambiguous_root_keeps_waiting_inside_grace`, `classify_ambiguous_root_reaps_after_grace`, `classify_ambiguous_root_never_reaps_when_grace_disabled`, `classify_ambiguous_root_tolerates_clock_skew`, `ambiguous_root_grace_secs_env_branches`, `ambiguous_root_actions_never_include_data_deletion` |

Every change site carries a `// #<issue>: <reason>` pointer.

## Verification

```
cargo fmt --check                                        → clean
cargo clippy -p trusty-search --all-targets -D warnings  → 0 warnings
cargo test -p trusty-search                              → 1387 passed; 6 failed
```

The 6 failures are **pre-existing and environment-caused**, not from this branch: `service::watcher::*`, `service::watch_loop::*`, `service::watcher_manager::*` — FSEvents does not deliver events on this host. Verified by stashing the entire branch and running the same subset against a clean tree, which failed with the identical 6 names. The `corpus_open_quarantine_4122` integration test fails on its own *positive control* (`"the healthy index never picked up the save, so this environment cannot distinguish a refusal from a stalled watcher"`) — same root cause, upstream of any code touched here. All other integration targets pass; `check_line_cap.sh` reports 0 violations.

## Scope

Does not touch the integration-test registry isolation (#4094, fixed in a parallel delegation). Stale-entry cleanup lives in the reaper only.


bobmatnyc:
## trusty-review verdict: **APPROVE** (Grade: A-)

Re-review of head `3f9ceda4` after the prior BLOCK was addressed.

**Manual verification of the previously-BLOCKing defect** (panicked/broken restores being parked as recoverable):
- `BoundedRestoreOutcome::is_parkable()` (`crates/trusty-search/src/service/warm_boot/restore.rs`) returns `true` **only** for `TimedOut`; `Completed` and `Panicked` both return `false`. Confirmed by the `only_timed_out_is_parkable` unit test.
- `ColdIndexStore::park_if_parkable` (`crates/trusty-search/src/service/lazy_loader/store.rs`) gates on `outcome.is_parkable()` before parking — a panicked restore can no longer reach the cold store and be reported as recoverable.
- The reaper's terminal path (`orphan_reaper.rs`, `AmbiguousRootAction::ReapRegistration`) removes only the `indexes.toml`/`roots.toml` registry rows via `remove_index_registry_entry`/`remove_root`; `remove_index_data_dir` is never called, and the log explicitly states on-disk index data is not deleted. The original defect (panicked → parked → reported recoverable) is fixed, and the reaper still never deletes on-disk data.

**WARN-level findings from the automated review (non-blocking, informational):**
1. `crates/trusty-search/src/service/server/indexes.rs:581` — field rename `corpus_open_failed` → `corpus_open_failure`; diff-visibility limited the reviewer's ability to confirm every call site, flagged low-confidence (0.55).
2. `crates/trusty-search/src/commands/start/restore.rs:309` — `total_panicked` is tallied and logged at ERROR but not surfaced on `/health`/`WarmBootSummary`; operators polling `/health` won't see panicked-restore counts (confidence 0.8). Worth a fast-follow.
3. `crates/trusty-search/src/service/server/tests_health_degraded.rs:68` — same rename-visibility caveat as #1, in test code (confidence 0.5).
4. `crates/trusty-search/src/commands/start/restore.rs:381` — `park_if_parkable` receives a clone of `entry` unconditionally, even on `Completed`; minor unnecessary allocation, no correctness impact (confidence 0.75).
5. `crates/trusty-search/src/service/server/indexes.rs:518` — new `ambiguous_root_since_unix` field; struct-exhaustiveness note, low confidence (0.4).

Review summary: 28 files reviewed across 28 units, 0 skipped, 0 failed, 5 findings surfaced. None rise to BLOCK severity.

---
🤖 Reviewed by Trusty-Review (`us.anthropic.claude-sonnet-4-6`)

PR #4719, opened 2026-08-03
fix(trusty-common): stop hiding cold palaces from the memory TUI (#4690)

### LINKED ISSUE #4690: palace_has_content() hides all cold palaces from memory TUI
## Summary
`palace_has_content()` at `crates/trusty-common/src/monitor/memory_tui/view.rs:64` hides any palace whose counts are all zero.

## Root Cause
Since #4640, cold (never-opened) palaces report zeroed placeholder counts, so this filter now hides them all. On the live daemon, this is roughly 2,180 of 2,183 palaces.

## Current State
- PR #4684 introduced `counts_unknown` on `PalaceRow` and fixed the *rendering* of unknown counts as `—`
- The PR deliberately left this filter byte-identical (marked with inline `// #4682` note) because fixing it is a visibility change deserving its own live verification
- **Consequence**: `counts_unknown=true` and `palace_has_content()==false` are equivalent
- The `format_opt_count()` call sites added in `memory_tui/view.rs` and `memory_tui/render.rs` are unreachable for the cold-palace case
- Interactive TUI shows no row at all rather than `—`
- Only the dashboard panel (`dashboard/render.rs:105`) and CLI/`/ui` paths exercise the `None` branch

## Proposed Solution
Have `palace_has_content()` read the `.vectors()` / `.kg_triples()` / `.drawers()` accessors (which return `Option<u64>`) rather than the raw zeroed fields, so unknown is distinguished from measured-empty.

## Impact
This will surface ~2,180 previously-hidden palaces in the TUI and needs live verification that the list stays usable at that size.

**References:** #4682, #4684, #4640

bobmatnyc:
## Summary

`palace_has_content()` read `PalaceRow`'s **raw** count fields. Since #4640 a
never-opened ("cold") palace carries zeroed placeholder counts, so the gate
returned `false` for every one of them and the palace was dropped by
`filtered_sorted_palaces` (`view.rs:105`), `visible_palace_ids` (`view.rs:120`),
and both navigation helpers — **before any renderer ever saw it**.

That made the `—` rendering PR #4684 added structurally unreachable in the
interactive TUI: `counts_unknown == true` and `palace_has_content() == false`
were equivalent, so the `format_opt_count(None)` call sites could only ever fire
on the dashboard/CLI paths.

Closes #4690.

## Mechanism — the accessors, as proposed

I kept the mechanism the issue proposed (read through `vectors()` /
`kg_triples()` / `drawers()`) rather than removing the filter or pushing the
decision to the call sites. Reasons:

- The filter has a real, still-valid job: a *measured*-empty palace is noise.
  Removing it entirely would regress `test_filter_empty_palaces` semantics that
  nothing in #4690 asks to change.
- Four call sites (`filtered_sorted_palaces`, `visible_palace_ids`,
  `navigate_up_visible`, `navigate_down_visible`) share the predicate. Pushing
  the unknown-vs-empty decision out to each of them multiplies the rule by four
  and invites exactly the drift that produced this bug.

`crates/trusty-common/src/monitor/memory_tui/view.rs:68`:

```rust
pub fn palace_has_content(palace: &PalaceRow) -> bool {
    // `None` (unknown) keeps the palace; `Some(0)` (measured empty) does not.
    let keeps = |count: Option<u64>| count.is_none_or(|n| n > 0);
    keeps(palace.vectors()) || keeps(palace.kg_triples()) || keeps(palace.drawers())
}
```

Resulting semantics:

| Counts | Before | After |
|---|---|---|
| UNKNOWN (`counts_unknown`) | hidden | **visible**, rendered `—` |
| KNOWN, all zero | hidden | hidden (unchanged) |
| KNOWN, any non-zero | visible | visible (unchanged) |

## Blast radius — the ~2,180-palace flood no longer applies

Issue #4690 warns this "will surface ~2,180 previously-hidden palaces and needs
live verification that the list stays usable at that size." **That is no longer
true.** Those palaces were purged earlier today; the corpus went from 2,183 to
**126** palaces. A reviewer reading #4690 will expect the old blast radius —
it is now a small, eyeball-verifiable visibility change.

## Known adjacent gap (not fixed here)

`PalaceRow`'s `ListItem::count()` still returns the raw `vector_count`, so sorting
by Vectors treats a cold palace as `0` and clusters cold palaces with the
measured-empty ones. That is a sort-ordering nit, not a visibility bug, and it is
out of scope for #4690.

## Tests

Three new tests in `crates/trusty-common/src/monitor/memory_tui/tests.rs`:

- `test_palace_has_content_semantics` — all three cases against the predicate
  directly, including the kg-only and drawer-only measured-non-zero legs.
- `test_cold_palace_visible` — the cold palace survives `filtered_sorted_palaces`,
  the measured-empty one does not, and the rendered row actually contains
  `UNKNOWN_COUNT` (`—`) — i.e. the #4684 path is reachable now.
- `test_cold_palace_navigable` — `visible_palace_ids` lists the cold palace and
  `navigate_down_visible` / `navigate_up_visible` can land the cursor on it,
  while the measured-empty palace stays unaddressable.

### Mutation check

Each mutation was applied to `view.rs`, the suite run, then the file restored
from a scratchpad backup (verified by `shasum`, not `git checkout`).

| Mutation | Tests that went red |
|---|---|
| **A** — revert to the raw fields (`vector_count > 0 \|\| …`), i.e. the pre-fix bug | `test_palace_has_content_semantics`, `test_cold_palace_visible`, `test_cold_palace_navigable` |
| **B** — drop the gate entirely (`true`) | the three above **+** `test_filter_empty_palaces` |
| **C** — vectors-only gate (drop the kg-triple and drawer legs) | `test_palace_has_content_semantics`, `test_filter_empty_palaces`, `test_all_selector`, `test_palace_lines` |

Mutation A is the important one: it is exactly the code this PR replaces, and all
three new tests catch it.

## Gates — Test Ladder rung 3 (localized behavior inside one crate)

```
$ cargo check -p trusty-common
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.94s

$ cargo test -p trusty-common
test result: ok. 241 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 0.71s

$ cargo test -p trusty-common --features monitor-tui
test result: ok. 404 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 0.72s

$ cargo clippy -p trusty-common --features monitor-tui --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.82s

$ cargo fmt --check
(clean — exit 0)

changelog-fragment gate: scanned 3 changed path(s); all 1 crate(s) with source changes are recorded.

$ bash scripts/check_sld.sh
sld-lint: scanned 54 spec doc(s) + 3029 code file(s); 0 error(s), 0 warning(s)
```

## Not verified

I did **not** run the TUI against the live daemon — this is a headless
environment. The em-dash rendering is asserted in `test_cold_palace_visible`
against `palace_lines()`, which is the same content the renderer draws, but
on-screen confirmation against the 126-palace corpus is still worth an eyeball
before this is considered fully closed.

## Changelog

`crates/trusty-common/changelog.d/4690-cold-palace-visibility.md` — single
`Fixed` category, per the one-category-per-fragment rule.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6535,7 +6535,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "680998035259dcfcafe653688bf2aa6d3e2dc05e98be6ab46afb089dc84f1df8"
 dependencies = [
- "proc-macro-crate 2.0.0",
+ "proc-macro-crate 1.3.1",
  "proc-macro2",
  "quote",
  "syn 2.0.117",
@@ -11398,7 +11398,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-common"
-version = "0.28.0"
+version = "0.28.1"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/crates/trusty-common/Cargo.toml
+++ b/crates/trusty-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trusty-common"
-version = "0.28.0"
+version = "0.28.1"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
--- a/crates/trusty-common/changelog.d/4690-cold-palace-visibility.md
+++ b/crates/trusty-common/changelog.d/4690-cold-palace-visibility.md
@@ -1,0 +1,3 @@
+Fixed
+
+- memory TUI no longer hides palaces whose counts the daemon never measured — `palace_has_content()` now reads the `Option<u64>` accessors instead of the raw zeroed fields, so *unknown* counts keep a palace visible (rendered as `—`) while a measured-empty palace stays filtered out (closes [#4690](https://github.com/bobmatnyc/trusty-tools/issues/4690))
--- a/crates/trusty-common/src/monitor/memory_tui/tests.rs
+++ b/crates/trusty-common/src/monitor/memory_tui/tests.rs
@@ -844,6 +844,148 @@ fn test_filter_empty_palaces() {
     assert!(rows.iter().any(|r| r.text.contains("drawer-o")));
 }
 
+/// A never-opened palace: the daemon returned placeholder zeros it never
+/// measured, flagged by `counts_unknown`.
+fn cold_palace(id: &str) -> PalaceRow {
+    PalaceRow {
+        id: id.into(),
+        name: id.into(),
+        counts_unknown: true,
+        ..Default::default()
+    }
+}
+
+#[test]
+fn test_palace_has_content_semantics() {
+    // Unknown counts are NOT empty — the palace stays visible so the renderer
+    // gets a chance to print `—` (#4690).
+    assert!(
+        view::palace_has_content(&cold_palace("cold")),
+        "unknown counts must not hide a palace"
+    );
+
+    // Measured and all zero — genuinely empty, still hidden.
+    assert!(
+        !view::palace_has_content(&PalaceRow {
+            id: "measured-empty".into(),
+            name: "measured-empty".into(),
+            ..Default::default()
+        }),
+        "a measured-empty palace stays hidden"
+    );
+
+    // Measured and non-zero — visible, unchanged.
+    for probe in [
+        PalaceRow {
+            vector_count: 1,
+            ..Default::default()
+        },
+        PalaceRow {
+            kg_triple_count: 1,
+            ..Default::default()
+        },
+        PalaceRow {
+            drawer_count: 1,
+            ..Default::default()
+        },
+    ] {
+        assert!(
+            view::palace_has_content(&probe),
+            "a measured non-zero palace stays visible: {probe:?}"
+        );
+    }
+}
+
+#[test]
+fn test_cold_palace_visible() {
+    let mut state = MemoryTuiState::new("http://x");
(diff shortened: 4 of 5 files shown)


bobmatnyc:
CI green on head `a3c12a1563f25066a01adbcc362745c63e799146` — 27 pass, 2 skipping, 0 fail (`gh pr checks 4719` exit 0).

Follow-up commit `a3c12a15` bumps `trusty-common` 0.28.0 → **0.28.1**. `check-pr-version-bump.sh` (#4421) flagged 0.28.0 as already live on crates.io — the local run now reports `[ OK ] trusty-common — version bumped 0.28.0 -> 0.28.1` and `check-version-parity.sh` reports `0 drifted, 0 unverifiable`.

Patch rather than minor: the change is behavioral only — no public item added, removed, or re-typed — so the `version = "0.28"` pins in `trusty-gworkspace`, `trusty-memory`, and `trusty-search` are minor-level and already accept 0.28.1 without an edit. (Contrast #4684, which needed 0.27 → 0.28 because `PalaceRow` gained a public field.)

Not merging — leaving this for review.

Issue #4312, opened 2026-07-29
[trusty-memory] PalaceHandle::remember_with_options false-positive-prone secret/credential heuristic


bobmatnyc:
## Summary

`PalaceHandle::remember_with_options` rejects ordinary, non-secret content as "contains a likely secret/credential token". Three distinct rejections occurred in a single session on 2026-07-29 while writing routine session-checkpoint memories. Each time the content was plain prose containing a structured-but-public string; rewording the string was the only workaround.

### Case 1: Timestamped .bak filename
**Flagged token:** `Agen…(30 chars)`

**Offending text:**
```
/Applications/Trusty Agents.app.bak-20260729-000028
```

The 30-char run is the filename segment `Agents.app.bak-20260729-000028` — a simple timestamped backup filename.

### Case 2: GitHub PR numbers in prose
**Flagged token:** `4292…(22 chars)`

**Offending text:**
```
four PRs merged (#4292/#4296/#4298/#4302)
```

**Workaround:** Rewriting as a comma-separated sentence ("4292, then 4296, then 4298, then 4302") passed without rejection.

### Case 3: Repository file path
**Flagged token:** `trus…(35 chars)`

**Offending text:** A file path for a markdown skill asset under the trusty-agents skills directory (a dot-prefixed directory path ending in a `.md` filename, roughly 35 characters).

**Workaround:** Describing the same files in prose instead of by path passed.

## Why This Matters

The rejection is **fail-closed**: the write is lost unless the author notices and rewords. While the error message names the truncated token (which helps), these three cases are all high-frequency shapes in exactly the kind of content a session checkpoint contains:
- File paths
- Timestamped filenames  
- Lists of issue/PR numbers

The documented exemptions already show shape-based logic is in use — commit SHAs are allowed and will not trigger this gate. These three shapes warrant similar treatment.

### The Escape-Hatch Problem

Note explicitly that `force=true` does **NOT** bypass this gate (per issue #2520, force bypasses content-quality gates but not secret detection), and `allow_secret_like` is documented as a rare, deliberate operator override that automated writers must not set.

**There is no appropriate escape hatch for an automated session-checkpoint writer hitting a false positive.** That is the crux.

## Suggested Directions (leave design to maintainer)

- Exempt filesystem-path-shaped strings, or at least strings containing `/` plus a known file extension
- Exempt runs that are mostly digits with separators (dates, timestamps, `#`-prefixed issue/PR number lists)  
[...]

Issue #2800, opened 2026-07-16
trusty-memory secret scanner false-positives on slash-separated issue/PR-number lists


bobmatnyc:
**Observed live 2026-07-16 (session 2eb72dca):** a `memory_note` write was rejected with:

> PalaceHandle::remember_with_options (note): Content rejected: contains a likely secret/credential token (`2763…(28 chars)`). Remove or redact the secret before storing; git commit SHAs are allowed and will not trigger this.

The flagged "token" was a slash-separated PR-number list in ordinary prose: `#2763/#2774/#2780/#2782/#2790` — 28 chars of digits, `#`, and `/` that pattern-matched the secret heuristic. Nothing secret-like about it; reformatting to comma-separated numbers ("PRs 2763, 2774, 2780, ...") passed.

**Why it matters:** PM session checkpoints routinely enumerate PR/issue lists; slash-separation is a natural way to write them, so this false positive will recur and silently degrades checkpoint fidelity (agents reword or drop detail to get past the filter).

**Suggested fix:** in the secret-detection heuristic (trusty-common memory-core, the `remember_with_options` rejection path), exempt candidate tokens composed only of `[#/0-9]` (or more generally: digits + separators with no base64/hex-alphabet spread and low character-class entropy) — same spirit as the existing git-SHA exemption. Add a regression test with the exact string above.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
**Three more reproducible false positives, same detector — filing here rather than a new issue since this is the same bug class already open.**

Hit in a single PM session on 2026-07-21, each a plain, non-secret string rejected by `memory_remember` / `PalaceHandle::remember_with_options` with the same `PotentialSecret` error this issue reports:

| Rejected token | Len | What it actually is | Reproduces on current `main` (verified)? |
|---|---|---|---|
| `crates/trusty-agents/src/workflow/resolver.rs:342` | 49 | A repo-relative source path with a line number | **No** — see note below |
| `review+tickets+slack+telegram+analyze` | 37 | A plus-delimited list of five service names | **Yes** |
| `<bare-dir>/worktrees/<name>` | 27 | A path template in prose | **Yes** |

I traced all three against the detector as it stands on `main` (`1e75ae93`) rather than assuming the symptom still reproduces, by vendoring the exact logic into a throwaway harness and running it:

```
[...]

Issue #2442, opened 2026-07-11
fix(trusty-memory): force:true does not bypass content/blocklist gates + blocklist substring over-match thins agent recall surface


bobmatnyc:
In `handle_memory_remember` (`crates/trusty-memory/src/tools/memory_ops.rs:32-136`), `blocklist_gate` (line 51) and `content_gate` (line 64) run BEFORE `force` is parsed (line 84) — `force` only bypasses `dedup_gate`.

**Live-reproduced** (QA of trusty-tools PR #2435, trusty-memory 0.19.2): short turn text with `force: true` → `status: "skipped"` (`reason: "content gate: skipped (short prompt, no context)"`).

**Sharper problem:** `blocklist_gate` substring-matches the WHOLE text, so any turn whose response QUOTES phrases like `"Tool use: "` or `"Claude Code session"` is silently skipped despite `force` — realistic recall-surface thinning for a coding agent whose turns routinely quote tool output.

**Asks:**
- (a) Define/fix `force` semantics vs ALL gates — app-managed writers like tcode's turn recorder (#2424) need deterministic storage.
- (b) Scope the blocklist to structural detection (e.g. prefix/frame matching) rather than substring-anywhere.

Found during QA of PR #2435 (tcode #2424); the sink's skipped-warn (#2363) makes this observable in the meantime.

Refs #2424, #2363, epic #2343.

PR #4723, opened 2026-08-04
fix(trusty-common): treat backtick as a token delimiter in the secret detector

bobmatnyc:
Refs #4312

> [!WARNING]
> **Merger: do not accept the default squash body.** This repo sets
> `squash_merge_commit_message=COMMIT_MESSAGES`, so the default squash body is
> the concatenation of all four commit messages — and the first commit
> (`9d7f7709`) predates the re-scope and still carries an auto-linking keyword
> pointing at the issue. Accepting the default would retire that issue while one
> of its enumerated cases is still reproducible.
>
> Merge with an explicit body instead:
>
> ```bash
> gh pr merge 4723 --squash --body "Refs #4312"
> ```
>
> The alternative is a force-push to reword `9d7f7709`; I did not do that
> unilaterally because it invalidates the SHAs the review rounds cite.

**This links the issue without auto-retiring it, deliberately.** This PR handles
7 of the 8 cases the issue enumerates. One is still live, has a different root
cause, and is out of scope — auto-linking would retire the issue with that case
unaddressed. Verified against the real crate at this HEAD:

| # | Case | Status |
|---|---|---|
| body 1 | `/Applications/Trusty Agents.app.bak-20260729-000028` | **still flags** → `Agen…(30 chars)` |
| body 2 | `#4292/#4296/#4298/#4302` | handled |
| body 3 | `.claude/skills/tm-ticketing/SKILL.md` | handled |
| comment-1 4 | `knowledge_base-slug-like-expression` | handled |
| comment-2 1 | `credentials/secret.rs:106,:118` | handled |
| comment-2 2 | `--features credentials,inference-client` | handled |
| comment-2 3 | `old-leaks/new-doesn't` | handled |
| comment-2 4 | `ticker+shutdown-channel` | handled |

Body Case 1 exits via the **mixed-case branch** — the capital `A` in
`Agents.app.bak` breaks `is_uniform_word_segment` — which this PR does not
touch. Its redacted preview matches the issue character for character.

## The defect, and its actual cause

`memory_core::filter` rejected ordinary prose as a credential. Four earlier rounds (#1667, #2800, #4216, #4312) each patched one token shape; the shape kept coming back with different punctuation.

The cause is one line. The base64 branch of `looks_like_secret` fires on **any** token of 20+ chars containing a `/` plus one letter, whenever `is_structural_token` declines to rescue it — and `is_word_segment`'s charset is deliberately narrow, so *every* Markdown decoration routes prose to a credential verdict. Backtick was simply the fourth character to be noticed.

## The fix — four changes, all narrowing

1. **`is_plausible_b64_charset`** gates the base64 branch — the treatment #2442 gave the mixed-case branch.
2. **An entropy floor** (`has_upper || has_digit`). Base64 encodes bytes, so a 20+ char encoded run all but certainly carries an uppercase letter or a digit; an all-lowercase run is English.
3. **`is_url_credential_shaped`**, narrowed to `scheme://user:pass@host` — a colon-bearing userinfo before the first `@`. Exempts connection strings from the floor by shape, without exempting bare URLs.
4. **The backtick is a token delimiter** in `find_secret_token`, which uniquely buys the "credential flush against a backtick" tightening.

## Measured, not asserted

Every number below was produced here against the real crate, at both revisions.

| Corpus | `origin/main` | this PR |
|---|---|---|
| Prose false positives (18 Markdown shapes) | **17** | **0** |
| Prose false positives (14 URL shapes) | **14** | **4** |
| Credential misses (30 shapes) | **2** | **1** |

**Detection is tightened, not loosened.** One of the two baseline misses is fixed: a credential flush against a backtick (`` token`AbCd1234EfGh5678IjKl9012` ``) was *missed* on main. The remaining miss is a canonical AWS **secret access key**, swallowed by `is_structural_token`'s slash-path bypass before any new gate runs — re-verified pre-existing against an unmodified `filter.rs`, and asserted as an explicit baseline.

## Round 2 review: the URL narrowing

The first cut of `is_url_credential_shaped` exempted any token containing `://` or `@`. It is an OR term inside a conjunction, so it can only *widen* flagging and cannot cause a miss — but it held open 14 URL-shaped prose false positives. This project's always-clickable-links convention puts one of those in essentially every checkpoint.

I verified the narrower predicate rather than taking the measurement on trust:

```
URL prose false positives : 14/14 on origin/main  ->  4/14 on this branch
Credential misses         :  0/10 before          ->  0/10 after
Exemption                 :  9/9 userinfo connection strings exempt, 0/14 prose exempt
```

That matches the review's exemption figures exactly.

**Four URL shapes still flag, and I did not fix them.** They carry a digit or a capital, so they clear the entropy floor unaided:

```
https://github.com/bobmatnyc/trusty-tools/pull/4723
https://github.com/bobmatnyc/trusty-tools/issues/4312
file:///Users/masa/trusty-tools/crates/common
mongodb://cluster0.mongodb.net/analytics-store
```

A bare GitHub issue link is the most common shape in this repo's checkpoints, so leaving it is a real cost — but the obvious blanket rule ("a userinfo-free URL is not a credential") is **unsafe**. A userinfo-free URL can carry the secret in its *path*: a `…/services/<id>/<id>/<token>` webhook URL is exactly that shape, is caught today, and would be lost. That case is now asserted in the must-flag battery, and the four survivors are pinned as change-detectors in `url_prose_residue_is_a_known_pre_existing_bound`. Separating "URL path that is a path" from "URL path that is a credential" is its own design problem.

> Aside worth recording: GitHub push protection rejected the first push because the webhook test fixture used the real vendor host. The fixture now uses a generic host. That is a second, independent opinion that this shape reads as a credential — and the reason the blanket exemption is the wrong move.

## Over-blocking guards

`real_secrets_still_blocked_after_4312_charset_gate` asserts 21 credential shapes stay flagged: every provider prefix, padded base64, base64-with-`+/`, base64url, JWT, bare mixed-case blob, seven connection strings, and the path-secret webhook URL. Five connection strings are **all-lowercase and digit-free** — the entropy floor alone would drop every one, so they are what proves the URL exemption carries its weight.

## Two things to look at

**1. Known limitations, stated and tested.** A backtick inside a connection-string password splits the token and that credential is missed — true for user-chosen passwords, impossible for machine-generated credentials. Pinned by a test, with the clean form asserted still-caught to bound it. The URL residue above is the second.

**2. One existing test's assertion changed.** `real_secrets_still_blocked_after_2800_exemption` case 2 asserted `looks_like_secret("#2763/#2774/#abcd/#2782/#2790") == true` as a *proxy* for "one letter takes this out of the #2800 exemption." The charset gate now also declines it (`#` appears in no credential format), so the test asserts the invariant directly (`!is_issue_number_list(...)`). Same token, same property, nothing removed.

## Gates — Rust Test Ladder **rung 4**

```
$ cargo fmt --check
exit=0

$ cargo check -p trusty-common --features memory-core
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.88s

$ cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.07s

$ cargo test -p trusty-common --features memory-core
running 711 tests
test result: ok. 698 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out; finished in 10.37s

$ cargo test -p trusty-common --features memory-core --lib -- memory_core::filter
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 661 filtered out
```

> `--features memory-core` is mandatory: `trusty-common`'s default feature set is empty, so a bare `cargo test -p trusty-common` filters out all 711 memory-core tests and exits 0 while proving nothing.

**Direct dependents:**

```
$ cargo test -p trusty-agents
test result: ok. 3312 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out
(+ 9 further targets, 0 failed), exit 0

$ cargo test -p trusty-memory
test result: FAILED. 486 passed; 2 failed; 4 ignored; 0 measured; 0 filtered out
```

**Corrected from an earlier revision of this PR body, which reported 488/0.** The two failures are `commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener` and `…_fails_when_no_addr_file_and_no_listener` — environment-dependent on whether a live memory daemon is listening, which is why the earlier run was green. **Proven pre-existing, not asserted:**

```
$ git diff --name-only origin/main...HEAD -- crates/trusty-memory/
(no output — zero files touched in that crate)

# same two tests, with trusty-common's filter.rs + filter_tests.rs reverted to origin/main:
running 2 tests
test commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener ... FAILED
test commands::doctor::tests::check_daemon_health_fails_when_no_addr_file_and_no_listener ... FAILED
    panicked at crates/trusty-memory/src/commands/doctor/mod.rs:592:9:
    unexpected status Warn; expected Fail (stale addr, no listener), …
```

Identical failure with the diff reverted. Nothing was `#[ignore]`d, `cfg`-gated or `--exclude`d.

**Workspace:**

```
$ cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui --exclude trusty-agents-ui
    Finished `dev` profile [unoptimized + debuginfo] target(s)
```

The three excluded crates fail on a pre-existing environmental Tauri gap (`frontendDist` → `ui/dist` missing); `git diff --name-only origin/main...HEAD` touches zero files in them, and CI excludes the same three (`ci.yml:713,787`).

**Repo gates:**

```
bash scripts/check_line_cap.sh           → 3603 files; 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh → OK trusty-common: fragment present and valid
```

## Known miss introduced here, and why it stays

Round 3 review caught a false invariant in my own doc: I wrote that narrowing
`is_url_credential_shaped` "can never turn a caught credential into a miss."
That has the direction backwards. The predicate is an OR term inside the
branch's conjunction, so widening it widens flagging and **narrowing it narrows
flagging** — narrowing can and does cause misses. Measured at both revisions:

```
https://webhook.example.com/services/abcdefghij/klmnopqrst/uvwxyzabcdefghij
https://hooks.example.org/t/aaaaaaaaaa/bbbbbbbbbb/cccccccccc
    origin/main: FLAGGED   ->   this branch: MISSED
```

The class that moved is exactly: a userinfo-free URL whose path secret is
all-lowercase and digit-free, so the entropy floor reads it as English. That is
the floor's already-accepted bound rather than a new kind of hole — the round-2
broad predicate shielded these incidentally, never by design — and real webhook
tokens are near-universally mixed-case or digit-bearing, which the uppercase
form in the must-flag battery pins.

The battery had no coverage for the moved class (all five all-lowercase
digit-free entries were userinfo-bearing), so both shapes are now explicit
known-miss assertions. The doc states the real bound.

## Base refresh

Merged current `origin/main` in. Two reasons: the branch was several commits
behind, and the stale base made the changelog-fragment gate fail on
`trusty-search` — a crate this PR never touches. That was base drift, not a gap
on main: each of main's `trusty-search` commits did add a fragment, and a
release then consumed them into `CHANGELOG.md`, so the old-base diff showed
"src changed, fragment added-then-deleted." After the merge the gate reads
`scanned 3 changed path(s); all 1 crate(s) with source changes are recorded`.

The version bump is no longer part of this diff — main independently moved
`trusty-common` to 0.28.1, and `check-pr-version-bump` reports
`0.28.1 — src changed, version not yet published`.


bobmatnyc:
## Verdict: WARN

Adversarial review, independent of the implementer. I did not take the PR body's
gate claims on trust — I re-ran them, and I built a differential probe that
extracts the decision logic (`find_secret_token`, `looks_like_secret`,
`is_structural_token`, `is_issue_number_list`, `is_plausible_credential_charset`,
`is_git_sha_like`) verbatim from `filter.rs` at this HEAD and runs `origin/main`'s
whitespace-only split and this PR's whitespace-or-backtick split side by side over
the same inputs.

### What holds up

**The security claim survives for every machine-generated credential format.**
Exhaustive sweep — backtick inserted at *every* character position — produced zero
FLAG→PASS transitions for: `ghp_` PAT (0/39), `sk-` key (0/42), `AKIA` key id
(0/19), padded base64 blob (0/43), JWT (0/70), bare mixed-case blob (0/31),
postgres / mysql / amqp / basic-auth connection strings (0/46, 0/43, 0/49, 0/47).
Backtick-quoted credentials (` ``ghp_…`` `, ` ``sk-…`` `, ` ``AKIA…`` `, ` ``xoxb-…`` `,
padded base64, JWT, a PEM body line) are all still flagged, identically.

**The strongest claim in the PR — "previously MISSED, now caught" — is
independently confirmed.**

```
TIGHTENED: PASS -> FLAG | cred flush against bt | token`AbCd1234EfGh5678IjKl9012`)
                        | before=None  after=Some("AbCd…(24 chars)")
```

**Gates re-run here, not copied from the PR body:**

```
$ cargo test -p trusty-common --features memory-core --lib
running 706 tests
test result: ok. 693 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out; finished in 11.97s
```

Every pre-existing over-blocking guard passes (`real_secrets_still_blocked_after_1667_fix`,
`…_after_2800_exemption`, `key_equals_secret_still_blocked_after_1676_fix`,
`aws_access_key_ids_are_blocked`, `colon_bearing_credential_is_flagged`,
`connection_string_shaped_token_still_flagged`, `ledger_reference_token_not_flagged`).

Blast-radius claim verified independently, not accepted:

```
$ git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/ crates/trusty-agents/ui/
(no output)
$ git diff --stat origin/main...HEAD -- Cargo.lock
 Cargo.lock | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

Lockfile diff is exactly the version line; the `proc-macro-crate` re-resolution is
indeed reverted. Changelog fragment present and well-formed. Every `Test:` pointer
in the rewritten doc comment resolves to a test that exists and passes. Empty
fragments produced by `split` are harmless (`looks_like_secret("")` → `false` at the
`is_issue_number_list` / length-floor gates).

### Findings

|---|---|---|---|---|---|
| HIGH | PR body | 1 | `Closes #4312` auto-closes an issue whose own explicitly-enumerated acceptance cases are still failing. #4312 comment 2 lists four reproductions and says "**Present all four as must-not-flag regression cases when this is fixed.**" Three still reject on this HEAD (evidence below). #4312 is the only ticket tracking them — `gh search issues` returns no follow-up. | Change to `Refs #4312` and leave the issue open for the remaining shapes, **or** file the follow-up and link it before merge. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/filter.rs` | 435-437 | The absolute claim "A backtick appears in no credential alphabet … so treating it as a delimiter **cannot** split a genuine credential" is refuted for user-chosen passwords in connection-string URLs, which this detector does catch today. | Soften to what is provable and record the limitation (see below). | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/filter_tests.rs` | 1112-1134 | `real_secrets_still_blocked_after_4312_backtick_split` covers only backtick-**adjacent** and backtick-**wrapped** credentials. The one direction the change can regress — a backtick **inside** a credential-bearing token — has no test at all. | Add the `mongodb://user:pa\`ss…` case and assert the decided behaviour explicitly. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/filter.rs` | 720-724 | Does not prevent a fifth recurrence. Ten more prose shapes still false-positive on this branch. A concrete, more durable fix exists and is demonstrated below. | Follow-up issue. | Promote |
| LOW | `crates/trusty-common/src/memory_core/filter.rs` | 429 | "[`redact_token`]'s caller trims punctuation only from the OUTER boundary" — `redact_token`'s caller *is* `find_secret_token`, and the trim happens before classification, nothing to do with redaction. Reads as if redaction is in the trim path. | "this function trims punctuation only from the outer boundary…" | Fix here |

---

#### HIGH — three of #4312's four enumerated must-not-flag cases still reject

Run against this PR's HEAD. The redacted previews match the ones pasted in the
issue **character for character**, which pins them as the same reproductions:

```
repro 1  len=30 credentials/secret.rs:106,:118          -> STILL REJECTED as cred…(30 chars)
repro 1b len=39 --features credentials,inference-client -> accepted
repro 3  len=21 old-leaks/new-doesn't                   -> STILL REJECTED as old-…(21 chars)
repro 4  len=23 ticker+shutdown-channel                 -> STILL REJECTED as tick…(23 chars)
```

The issue title is scoped to the heuristic in general ("false-positive-prone
secret/credential heuristic"), not to backticks. Merging with `Closes` drops three
live, reproducible defects on the floor in a release billed as pristine. This is
cheap to fix and does not require touching the code.

#### MEDIUM — the "cannot split a genuine credential" claim, precisely

True for base64, base64url, hex, base32 and every provider key format — I verified
that exhaustively. **Not** true for a user-chosen password inside a connection-string
URL, where the charset is unbounded and a literal backtick is legal:

```
*** REGRESSION: FLAG -> PASS *** | mongodb://user:pa`ss@cluster0.mongodb.net
    before=Some("mong…(41 chars)")   after=None
```

Sweeping backtick position across `mongodb://user:passw0rdX@cluster0.mongodb.net`
(clean form: detected both before and after): **13 of 44 positions** turn a
detected credential into a miss — every position from the scheme through the early
password, because the left fragment falls under the 20-char floor while the right
fragment loses its `/` (killing `has_b64_sym`) and carries `@` (failing
`is_plausible_credential_charset`).

Two things keep this at MEDIUM rather than HIGH, and I want them on the record
rather than buried:

1. **The adversarial delta is nil.** The detector has always split on whitespace,
   so anyone deliberately evading it inserts a space, before and after this PR.
   This is an accidental-storage hygiene gate, not an adversarial control.
2. **The pre-fix catch was itself accidental.** The backtick was corrupting
   `is_word_segment`, pushing the token out of the structural bypass and into the
   base64 branch. The same guard already misses far more likely secrets outright
   on `origin/main` — `wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY` (a canonical AWS
   secret access key) returns `None` both before and after, swallowed by the
   slash-path structural bypass, and so does a bare `.env` line
   `API_KEY=AbCd1234EfGh5678IjKl9012MnOp` via the `key=value` bypass. Both are
   pre-existing and out of scope here; noting them so the residual risk of this
   change is judged against the real baseline.

So: this does not open a hole, but the doc comment and changelog assert an
absolute that a reader will later rely on. State the actual boundary.

#### MEDIUM — does it prevent a fifth recurrence? No. The fifth is already reproducible.

The root cause is not the delimiter. It is that the base64 branch of
`looks_like_secret` (line 722) fires on **any** ≥20-char token containing `/` plus
one letter whenever `is_structural_token`'s narrow `is_word_segment` charset
rejects a segment. Backtick was one character outside that charset; so is every
other Markdown decoration. Run on this branch, post-fix:

```
FALSE POS | **stale_skills**/**doctor_staleness.rs**              (bold-joined — same shape, ** instead of ``)
FALSE POS | "stale_skills"/"doctor_staleness.rs"                  (quoted pair)
FALSE POS | 'stale_skills'/'doctor_staleness.rs'                  (single-quoted pair)
FALSE POS | [filter.rs](crates/trusty-common/src/filter.rs)       (ordinary Markdown link to a file)
FALSE POS | |crates/common|src/memory_core/filter.rs|             (Markdown table row)
FALSE POS | [stale_skills]/[doctor_staleness.rs]                  (bracketed pair)
FALSE POS | (stale_skills)/(doctor_staleness.rs)                  (paren pair)
FALSE POS | "#4601"/"#4602"/"#4603"                               (#4216 regression, quoted instead of backticked)
FALSE POS | **#4601**/**#4602**/**#4603**                         (same, bolded)
FALSE POS | docs/my%20file/spec%20v2.md                           (URL-encoded path)
FALSE POS | 82%-100%/idle-across-the-window                       (ratio prose)
```

A Markdown link to a source file and a Markdown table row are not exotic — they
are what these memory writes are made of. The sixth report will be one of these.

**Concrete durable alternative, measured not speculated.** Apply to the base64
branch the same treatment #2442 already applied to the mixed-case branch — gate it
on a credential-plausible charset — leaving the delimiter alone:

```rust
fn is_plausible_b64_charset(t: &str) -> bool {
    t.chars().all(|c| c.is_ascii_alphanumeric() || matches!(c, '+'|'/'|'='|'-'|'_'|'.'|':'|'@'))
}
// line 722:
if has_b64_sym && (has_lower || has_upper || has_digit) && is_plausible_b64_charset(token) { … }
```

Differential result over 13 prose shapes and 11 real-credential shapes: that one
predicate fixes **all three** #4312 backtick reproductions *plus* seven of the ten
shapes above (bold, quoted, Markdown link, table row, `credentials/secret.rs:106,:118`,
`old-leaks/new-doesn't`, `%20` paths, ratio prose) with **zero** loss across every
real credential I tested — `ghp_`, `sk-`, `AKIA`, padded base64, base64-with-`+`,
JWT, mixed-case blob, colon credential, `postgres://…@…` connection string,
`token=<b64>`, backtick-quoted `ghp_`. It is orthogonal to and composable with this
PR's delimiter change; the delimiter change is what uniquely buys the
"credential flush against a backtick" tightening, so keeping both is strictly best.

I am **not** asking for that in this PR. It is a larger change and this one is a net
improvement. It belongs in the follow-up that HIGH-1 requires anyway.

1. Downgrade `Closes #4312` to `Refs #4312`, or file the follow-up covering
   `credentials/secret.rs:106,:118`, `old-leaks/new-doesn't`,
   `ticker+shutdown-channel` and the ten shapes above, and link it here.
2. Replace the absolute "cannot split a genuine credential" in `filter.rs:435-437`
   and in the changelog fragment with the provable claim plus the connection-string
   limitation.
3. Add the backtick-inside-a-credential case to
   `real_secrets_still_blocked_after_4312_backtick_split` so the one regression
   direction is pinned by a test rather than by a paragraph.

None of these touch the behavioural change, which I am satisfied is correct and
strictly better than what it replaces.

### Notes

- `cargo doc` emits `rustdoc::private_intra_doc_links` for the new
  public→private links (`[`redact_token`]`, `[`looks_like_secret`]`,
  `[`is_plausible_credential_charset`]`). Deliberately **not** filed: that warning
  class is already widespread across this crate and `cargo doc` is not a CI gate.
- The AWS-secret-access-key and `.env` `KEY=<blob>` misses noted above are
  pre-existing on `origin/main` and unchanged by this PR. Flagged as context for
  severity calibration only, not as findings against this diff.

bobmatnyc:
## Verdict: WARN

Re-review at `42494402`. I rebuilt the differential from scratch rather than reusing
round 1's: functions are sliced out of `filter.rs` verbatim by brace-matching at both
`origin/main` and this HEAD, plus three **gate-isolation variants** built by textually
replacing only the base64-branch condition — `charsetonly` (charset gate, no floor),
`flooronly` (floor, no charset gate), `nourl` (floor with the URL exemption removed).
Every conclusion below is measured against those five, and the extraction was validated
20/20 against the real crate compiled as a dependency.

---

## 1. My round-1 recommendation was wrong. Confirmed, not defended.

I wrote that the charset predicate alone "fixes **all three** #4312 backtick
reproductions plus seven of the ten shapes." That is false for repro 4.

```
case        before   after    charsetonly   flooronly   nourl    input
repro1      FLAG     pass     pass          FLAG        pass     credentials/secret.rs:106,:118
repro2      pass     pass     pass          pass        pass     --features credentials,inference-client
repro3      FLAG     pass     pass          pass        pass     old-leaks/new-doesn't
repro4      FLAG     pass     FLAG          pass        pass     ticker+shutdown-channel
```

`charsetonly` — my recommendation, implemented exactly as I specified it — still flags
`ticker+shutdown-channel`, because every character in it is individually
credential-plausible. Had it shipped alone with `Closes #4312`, it would have
auto-closed an issue with a live enumerated reproduction: the precise defect my own
HIGH finding raised. The author caught what I missed. The entropy floor is load-bearing,
not belt-and-braces.

## 2. The entropy floor's premise holds. I could not break it.

Monte Carlo, 2.5M generated blobs, four alphabets × five lengths. `NEW_miss_from_entropy_floor`
counts tokens `charsetonly` catches that `after` drops:

```
  base64        len=20/24/32/40/64   NEW_miss_from_entropy_floor = 0 0 0 0 0
  base64url     len=20/24/32/40/64   NEW_miss_from_entropy_floor = 0 0 0 0 0
  base32-upper  len=20/24/32/40/64   NEW_miss_from_entropy_floor = 0 0 0 0 0
  hex-lower     len=20/24/32/40/64   NEW_miss_from_entropy_floor = 0 0 0 0 0
```

**Zero.** The structural reason: to reach the base64 branch at all a token must carry
`+`, `/` or `=` after trimming and survive `is_structural_token`. An all-lowercase,
digit-free token that does so must contain a `+` or an interior `=`. For random base64
at 20 chars, P(no uppercase ∧ no digit) ≈ 1.3e-8.

I built the counterexample classes you named and every one of them **never reaches the
branch**, on this PR or on main:

| candidate | why the floor is not implicated | before | after |
|---|---|---|---|
| `deadbeefcafebabefacefeedbadedeed` (lowercase hex) | pure hex → `is_git_sha_like` allowlist | pass | pass |
| `aebcdeff-caed-4bad-bead-cafedeadbeef` (UUID) | no `+/=` → base64 branch unreachable | pass | pass |
| `mfrggzdfmztwqzlomnxwizlsmrqwezldn` (base32 lower) | no `+/=` | pass | pass |
| `sl.abcdefghijklmnop…` (lowercase API key) | no `+/=` | pass | pass |
| `password=correcthorsebatterystaple` | `key=value` structural bypass | pass | pass |

These are **pre-existing gaps in other branches**, identical on `origin/main`, not holes
the floor opens. The only tokens the floor actually drops are English:
`secret+token+for+production+use`, `correcthorse+batterystaple+here`. That is the job.

## 3. `is_url_credential_shaped` cannot cause a miss — proof, not a probe.

It appears only as an **OR term** inside a conjunction that the PR *added*:
`has_b64_sym && is_plausible_b64_charset(t) && (is_url_credential_shaped(t) || has_upper || has_digit)`.
An OR term is monotone-widening: removing or narrowing it can only make the gate
stricter. It is therefore **structurally incapable** of turning a FLAG into a pass. Not
defeatable in the security direction, at any input.

Confirmed empirically — `nourl` drops exactly the four all-lowercase connection strings
and nothing else:

```
pg-lower      after=FLAG  nourl=pass   postgres://user:password@host/database
mysql-lower   after=FLAG  nourl=pass   mysql://admin:letmein@localhost/mydb
redis-lower   after=FLAG  nourl=pass   redis://default:secretpass@cache.internal/zero
amqp-lower    after=FLAG  nourl=pass   amqp://guest:guest@rabbitmq.svc.cluster.local/vhost
```

The four all-lowercase digit-free connection strings in the battery are genuinely
load-bearing. They are not decorative.

## 4. The two gates compose safely.

Both are conjunctive narrowings of one branch; neither can widen what the other closes,
so no ordering or interaction hazard exists. The one non-monotone change is the
tokenizer, and **both** of its directions are pinned: the tightening
(`` token`AbCd…` `` pass→FLAG) and the single regression
(`` mongodb://user:pa`ss@… `` FLAG→pass) with the clean form asserted still-caught to
bound it.

## 5. Differential re-run — my corpus, not the PR's

18 prose shapes / 29 credential shapes, independently constructed:

| | `origin/main` | this PR |
|---|---|---|
| False positives (18 prose) | **17** | **0** |
| Credential misses (29) | **5** | **4** |

Direction matches the PR's claim exactly: one miss **fixed** (`` token`AbCd1234EfGh5678IjKl9012` ``,
pass → FLAG), none added. My four residual misses are `wJalrXUtnFEMI/K7MDENG/…`,
`token=<b64>`, `API_KEY=<blob>`, `T00000000/B00000000/XXXX…` — all four pass identically
on `origin/main`, all four via the pre-#4312 structural bypass.

**AWS baseline assertion: pre-existing confirmed, and it cannot mask a worse miss.**
Verified against the real unmodified `origin/main` crate, not the PR's word. The
assertion pins exactly one token with `is_none()`, so it is a **change-detector**: fixing
the miss turns it red and forces a deliberate update, which the comment tells the reader
to do. It makes no claim about any other token, so it cannot hide a different
credential's miss — and broad under-detection is guarded by the 20 must-flag assertions
in the same function.

## 6. The changed assertion — deliberate look, as asked

`real_secrets_still_blocked_after_2800_exemption` case 2:
`looks_like_secret(sneaky) == true` → `!is_issue_number_list(sneaky)`.

**Equivalent in protective power. Not weaker.** The invariant the comment has always
named is "one letter takes this token out of the #2800 exemption." The old form probed
that *indirectly* through a downstream verdict; the new form asserts it *directly*. For
the regression it exists to catch — someone widening `is_issue_number_list` to admit
letters — **both forms fail**, so nothing is lost there.

What the old form additionally covered: it was a canary for "some *other* new exemption
started swallowing `#`-shaped tokens." That canary has now fired once, for a reviewed
change, and the rewrite retires it. I checked whether anything goes unasserted as a
result, and the answer is no that matters:

- the token is prose containing no credential, so flagging it was never security value;
- `#` appears in no credential format, so declining it loses no detection — I verified
  the charset gate costs **zero** real-credential catches across 2.5M blobs;
- over-blocking in the same shape is still covered by `` "#4601"/"#4602"/"#4603" `` and
  `**#4601**/…` in `markdown_decorated_paths_are_not_flagged`;
- cases 1, 3, 4 and 5 of the same test still assert `looks_like_secret` on `+`-bearing,
  `=`-padded and connection-string tokens, unchanged.

Flagging this yourself was the right call. It does not need to change.

---

## Findings

|---|---|---|---|---|---|
| HIGH | PR body | 1 | `Closes #4312` still auto-closes an issue with a live enumerated reproduction — the issue **body**'s Case 1, not comment 2's four. Evidence below. | `Refs #4312`, or file the follow-up for the timestamped-filename shape and link it. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/filter.rs` | 854-856 | `is_url_credential_shaped` exempts **any** token containing `@` or `://`. Security-safe (§3), but it holds open 14/14 URL-shaped prose false positives — including bare GitHub issue/PR URLs — 10 of which the floor would otherwise fix. A measured narrower predicate exempts 4/4 credentials and 0/14 prose. | Follow-up; narrow to a colon-bearing userinfo before `@`. | Promote |
| LOW | `crates/trusty-common/src/memory_core/filter.rs` | 826 | `See [`is_single_b64_alphabet`] for the second gate that resolves it` — **no such item exists** anywhere in the repo. Rustdoc does not catch it (private item, not documented by default). The real second gate is the `has_upper \|\| has_digit` floor plus `is_url_credential_shaped`. | Replace the link with those two names. | Fix here |

### HIGH — #4312's issue **body** Case 1 still reproduces

Every case #4312 enumerates, run against the real crate at both revisions:

```
                                                              main    this PR
body Case 1  /Applications/Trusty Agents.app.bak-20260729-000028   FLAG    FLAG   <-- still fails
body Case 2  #4292/#4296/#4298/#4302                               pass    pass
body Case 3  .claude/skills/tm-ticketing/SKILL.md                  pass    pass
comment-1 Case 4  knowledge_base-slug-like-expression              pass    pass
comment-2 repro 1-4                                                3 FLAG  all pass
```

Pinned as the same reproduction — the redacted preview matches the issue **character for
character**:

```
find_secret_token("/Applications/Trusty Agents.app.bak-20260729-000028")
  -> Some("Agen…(30 chars)")        // issue reads: Flagged token: `Agen…(30 chars)`
```

It is a **different root cause** — `Agents.app.bak` fails `is_uniform_word_segment` on its
capital `A`, so the token exits via the mixed-case branch, which this PR does not touch.
Genuinely out of scope for the fix. That is exactly why `Closes` is the wrong keyword:
the shape is live, reproducible today, and #4312 is the only ticket carrying it.

This is round 1's HIGH, narrowed from three surviving repros to one, and relocated from
comment 2 to the issue body. It is one word in the PR body, or one follow-up issue.

### MEDIUM — what the wide URL exemption costs

```
after   nourl    token
FLAG    pass     https://crates.io/crates/trusty-common/versions
FLAG    pass     https://docs.rs/trusty-common/latest/trusty_common
FLAG    pass     https://api.github.com/repos/bobmatnyc/trusty-tools
FLAG    pass     git@github.com:bobmatnyc/trusty-tools.git
FLAG    pass     contact@example.com/support-team-list
FLAG    pass     amqp://rabbitmq.svc.cluster.local/vhost-production
FLAG    pass     redis://cache.internal/keyspace-notifications
FLAG    pass     postgres://localhost/trusty_dev
FLAG    FLAG     https://github.com/bobmatnyc/trusty-tools/pull/4723
FLAG    FLAG     file:///Users/masa/trusty-tools/crates/common
-> 14/14 still false-positive on this PR; 10 the floor would have fixed if the exemption were narrower
```

`https://github.com/bobmatnyc/trusty-tools/pull/4723` is rejected as a secret. So is a
bare issue URL. Both are **pre-existing on `origin/main`** — this PR does not regress
them, and I am not scoring it as a regression. But the project's own standing convention
is *always clickable links for issues and PRs, never bare numbers*, which means a
conforming session checkpoint contains one of these tokens essentially every time. This
is the most likely shape for recurrence #7, and the new predicate is what keeps it alive.

A narrower form, measured — `://` plus a colon-bearing userinfo before `@`:

```
cred   narrow_exempt=true   postgres://user:password@host/database      (all 4 lowercase conns: exempt)
prose  narrow_exempt=false  https://github.com/…/pull/4723              (all 14 prose shapes: not exempt)
```

4/4 credentials kept, 0/14 prose exempted. Worth the follow-up; not worth blocking a PR
that is strictly better than what it replaces.

---

## Everything else you asked me to confirm — all clean

| Item | Result |
|---|---|
| Four #4312 repros + 11 Markdown shapes as must-not-flag tests | ✅ `four_4312_acceptance_cases_are_not_flagged`, `markdown_decorated_paths_are_not_flagged` (11 entries) |
| Absolute "cannot split a genuine credential" replaced with its provable bound | ✅ both `filter.rs:438-447` and the changelog fragment |
| `mongodb://user:pa` + backtick + `ss@host` test, clean form asserted still-caught | ✅ `real_secrets_still_blocked_after_4312_backtick_split` case 4 |
| `--features memory-core` mandatory | ✅ verified: bare `cargo test -p trusty-common` → `running 0 tests … ok. 0 passed`, exit 0 |
| `trusty-common` 0.28.1, lockfile diff limited to the version line | ✅ `Cargo.lock \| 2 +-`, one line |
| Nothing silenced | ✅ zero `#[ignore]`, `#[cfg(`, `--exclude` added; CI **26 pass, 2 skipping, 0 fail** at this SHA, `Test` green in 16m41s with no re-run |

**Gates re-run here, not copied:**

```
cargo fmt --check -p trusty-common                                     -> exit 0
cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings
                                                                       -> Finished, zero warnings
   running 709 tests
   test result: ok. 696 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out
   (+ 21 across integration targets, 0 failed)
cargo test -p trusty-common --features memory-core --lib -- memory_core::filter
   test result: ok. 48 passed; 0 failed; 0 ignored; 661 filtered out
cargo test -p trusty-agents   -> test result: ok. 3312 passed; 0 failed; 11 ignored
bash scripts/check_line_cap.sh           -> 3603 files; 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh -> OK trusty-common: fragment present and valid
bash scripts/check_sld.sh                -> 0 errors, 0 warnings
```

1. Downgrade `Closes #4312` to `Refs #4312`, **or** file the follow-up covering
   `Agents.app.bak-20260729-000028` (issue-body Case 1, mixed-case branch) and link it
   here. Nothing else.

## Notes

- **`cargo test -p trusty-memory` did not reproduce the PR body's `488 passed; 0 failed`
  on this machine** — I got `486 passed; 2 failed`, both
  `commands::doctor::tests::check_daemon_health_*`. Applied the baseline protocol rather
  than assuming: `git diff --name-only origin/main...HEAD -- crates/trusty-memory/` is
  empty, and both tests fail identically on an unmodified `origin/main` worktree
  (`0 passed; 2 failed; 490 filtered out`). Environment-dependent — a live trusty-memory
  daemon here makes them report `Warn`. **Not a finding, and not the author's error** —
  but the reported figure is machine-state-dependent, not universal.
- The `is_url_credential_shaped` over-blocking above is stated as a bound on the PR's
  "false positives 17 → 0" claim, which is accurate **for the 18-shape corpus measured**.
  It is not a regression and not a security finding.
- The CI flake narrative in the PR body checks out and I am not re-deriving it: the
  current SHA's `Test` job passed first time, 16m41s, nothing ignored or excluded.

The behavioural change is correct, the security reasoning is sound, and the test battery
is the strongest in this file. I would ship this secret filter. The one blocker is a
tracking keyword, not code.

bobmatnyc:
## Verdict: BLOCK

Round 3 review, scoped to the three changes. The URL narrowing and the corrected test figure both hold up. The closing-keyword change does not: the warning banner written to prevent the auto-close is itself what armed it, and the mitigation the banner proposes cannot disarm it.

## Findings

| CRITICAL | PR #4723 body | 5 | The banner's own text `` still contains `Closes #4312` `` is a live closing keyword. `closingIssuesReferences` is **not** empty — it returns `[#4312]`. Merging by any route closes #4312 with body Case 1 still failing. | Reword line 5 so no closing keyword directly precedes a `#`-ref, then re-verify `totalCount == 0`. | Fix here |
| HIGH | `crates/trusty-common/src/memory_core/filter.rs` | 859 | Doc asserts the narrowing "can never turn a caught credential into a miss". Measurably false — a userinfo-free URL with an all-lowercase, digit-free path secret goes FLAGGED on `origin/main` → MISSED here. Not pinned, not in Known limitations. | Correct the claim to its true direction, and pin the shape in the limitations test. | Fix here |

## 1. The closing keyword — the mitigation armed the trap it warns about

`closingIssuesReferences` returns `[#4312]`, not empty. Three independent reads, one cache-busted:

```
read 1: {"totalCount":1,"nodes":[{"number":4312,"state":"OPEN"}]}
read 2: {"totalCount":1,"nodes":[{"number":4312,"state":"OPEN"}]}
read 3: {"totalCount":1,"nodes":[{"number":4312,"state":"OPEN"}]}
```

No `ConnectedEvent` on #4312, so it is not a manual sidebar link. It is body-derived, and the body has exactly one closing-keyword occurrence — inside the warning banner:

```
5:> (`9d7f7709`) still contains `Closes #4312` from before this PR was
```

GitHub's parser honors that keyword through both the blockquote and the inline code span. This is the repo's own documented trap: negated and quoted closing keywords still close the issue.

**Commit messages are not the source.** PR #4625 is the discriminator — body keywords 0, commit keywords 2 (`Closes #4468`, `Closes #4421`), `closingIssuesReferences` = **0**. Commit-message keywords never populate that field.

That matters, because it makes the proposed mitigation ineffective. There are two independent traps here:

| Trap | Mechanism | Armed? | Does `--body "Refs #4312"` disarm it? |
|---|---|---|---|
| Commit `9d7f7709` carries `Closes #4312` | keyword lands on the default branch via the squash body | yes | **yes** |
| PR body line 5 carries `Closes #4312` | GitHub's linked-issue set closes on merge, independent of the commit message | yes | **no** |

The banner correctly identified trap 2 and correctly prescribed `--body` for it. Writing that banner armed trap 1, which `--body` cannot reach. As it stands, **#4723 closes #4312 on merge no matter what body is used.**

Fix: reword line 5 so the keyword never sits adjacent to a `#`-ref — e.g. "still carries the pre-rescope closing keyword for #4312" — then re-verify `totalCount == 0` before merging. The `--body` override is still required for trap 1; it is just not sufficient.

## 2. The URL narrowing — stopping was right, but one claim is wrong

**Stopping at 10/14 was correct engineering.** I looked for the safe narrower predicate and there isn't one at zero recall cost.

The strongest candidate is per-path-segment evaluation: exempt a userinfo-free URL unless some path segment independently clears the credential bar. It does separate the survivors from the must-flag webhook — longest non-secret segment among the four is `analytics-store` at 15 chars, and the webhook's secret segment is 24. But the margin is thin and the mechanism is a recall trade: the webhook's own `T00000000` and `B00000000` are 9 chars and would not clear a 20-char per-segment floor on their own. A webhook whose secret sits in a 16–19 char segment goes from caught to missed. That is a credential miss bought with prose cleanliness — the wrong direction. Pinning the four survivors as change-detectors in `url_prose_residue_is_a_known_pre_existing_bound` and deferring the design is the right call.

The push-protection rejection is worth the weight given to it. An independent scanner reading that shape as a credential is real evidence against a blanket userinfo-free exemption.

**But the doc comment at line 859 overstates the safety property**, and this is a security module where the next maintainer will act on it:

> Note this predicate can only ever WIDEN flagging: it is an OR term inside the branch's conjunction, so narrowing it is strictly safe in the security direction and can never turn a caught credential into a miss.

Two different statements are conflated. Adding the term widens flagging versus not having it — true. Narrowing the term reduces flagging versus the previous revision — so it *can* create a miss, and it did. `is_word_segment` rejects the empty segment produced by `://`, so every URL falls through `is_structural_token` to the base64 branch, which makes the class reachable.

Measured, same probe file at both revisions:

```
                                                    origin/main   this branch
A  https://webhook.example.com/services/            FLAGGED    →  MISSED
     abcdefghij/klmnopqrst/uvwxyzabcdefghij
A2 https://hooks.example.com/services/              FLAGGED    →  MISSED
     qwertyuiopasdfghjklzxcvb
B  .../services/T00000000/B00000000/XXXX…  (battery) FLAGGED   →  FLAGGED
C  postgres://user:password@host/database            FLAGGED   →  FLAGGED
D  https://github.com/.../pull/4723     (residue)    FLAGGED   →  FLAGGED
G  .../services/AbCdEfGhIjKlMnOpQrStUvWx             FLAGGED   →  FLAGGED
```

A and A2 are new misses introduced by this PR: a userinfo-free URL carrying an all-lowercase, digit-free path secret. Round 2's broad `://`-or-`@` predicate shielded these from the entropy floor; round 3 removed the shield.

I am **not** re-litigating the entropy floor — this is the floor's already-accepted "an all-lowercase run is English" bound, and the realistic exposure is small (a random 24-char lowercase-alnum token is digit-free ~0.07% of the time; every mainstream webhook format carries mixed case or digits). What is not acceptable is that the doc asserts the opposite invariant, the shape appears in neither the must-flag battery nor the Known-limitations section, and the must-flag battery's five all-lowercase digit-free entries are all userinfo-bearing — so nothing in the suite covers the one class the narrowing actually moved.

Fix: state the true direction ("narrowing this term reduces flagging and can create a miss; the shapes it moves are all-lowercase digit-free URL path secrets, measured and pinned below"), and add A to the limitations test alongside the interior-backtick case, in the same style — assert the miss so a future reader meets the behavior rather than the paragraph.

## 3. The corrected figure — method is sound, reproduced

`cargo test -p trusty-memory` at 486/2 rather than 488/0 is correct, and the pre-existing proof holds. I verified it the stronger way — not a partial revert, but the two tests run on unmodified `origin/main`:

```
test commands::doctor::tests::check_daemon_health_fails_when_no_addr_file_and_no_listener ... FAILED
test commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener ... FAILED
panicked at crates/trusty-memory/src/commands/doctor/mod.rs:592:9:
unexpected status Warn; expected Fail (stale addr, no listener), …
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 490 filtered out
```

Same two tests, same panic sites, same messages. `git diff --name-only origin/main...HEAD` touches five files, none in `crates/trusty-memory/`. Counts reconcile: 490 filtered + 2 = 492 = 486 passed + 2 failed + 4 ignored. Nothing was `#[ignore]`d, `cfg`-gated or `--exclude`d. Reverting `filter.rs`/`filter_tests.rs` is a valid method here because they are the only source files the PR changes; running on `origin/main` is the cleaner form and agrees.

Filter suite at HEAD reproduces exactly as reported:

```
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 661 filtered out
```

The `is_single_b64_alphabet` doc link is **fixed** — the string appears nowhere in `crates/`, at this HEAD or on `origin/main`.

## Required Changes

1. Reword PR body line 5 so no closing keyword precedes a `#`-ref, and re-verify `closingIssuesReferences` returns `totalCount: 0` before merging. Without this, #4312 auto-closes regardless of the squash body.
2. Correct the `is_url_credential_shaped` doc claim at `filter.rs:859`, and pin the all-lowercase digit-free URL-path-secret miss in the limitations test.

## Notes

- Merge mechanics once unblocked: an explicit `--body` is **required** (commit `9d7f7709` still carries the keyword) and **not sufficient** on its own (the body link is a separate mechanism).
- Per-path-segment URL evaluation is a reasonable follow-up but must not gate this PR, and must ship with its own must-flag battery covering short path secrets.
- Review artifacts were disposable worktrees at PR head and `origin/main`, both removed. No edits, commits, or pushes to the PR branch; head is still `7171722a`.

bobmatnyc:
## Verdict: APPROVE — round-4 focused verification

Scope: the four items that changed since round 3 (PR body, one doc comment, two test assertions, `origin/main` merge). Filter logic was not re-derived — the entropy floor, the URL predicate's structural safety, the differential, and the rewritten assertion are settled across rounds 1–3.

### 1. Closing keyword — PR is clean, the merge trap is real

`closingIssuesReferences.totalCount` read **twice**, second time cache-busted:

```
{"data":{"repository":{"pullRequest":{"closingIssuesReferences":{"totalCount":0,"nodes":[]}}}}}
{"data":{"repository":{"pullRequest":{"closingIssuesReferences":{"totalCount":0,"nodes":[]}}}}}
```

Rendered body scanned for `close[sd]?|fix(e[sd])?|resolve[sd]?` adjacent to `#N` or a GitHub URL, in any form — plain, blockquote, inline code, table cell, bullet: **zero matches.** Three bare keyword occurrences survive (L46 "The fix — four changes", L63 "is fixed:", L79 "did not fix them"), none adjacent to a reference. Line 1 is `Refs #4312`. The script guard did its job.

**The commit-message trap is confirmed real, not theoretical.** Per-commit scan of all six commits:

```
60782d49 (merge)  (no keyword+ref)
782c6e44          (no keyword+ref)
7171722a          (no keyword+ref)
42494402          (no keyword+ref)
7a2876c3          (no keyword+ref)
9d7f7709          Closes #4312     <-- line 22
```

Repo setting, verified directly:

```
{"allow_squash":true,"squash_msg":"COMMIT_MESSAGES","squash_title":"COMMIT_OR_PR_TITLE"}
```

`COMMIT_MESSAGES` concatenates all six commit messages into the squash body, so the default squash would carry `Closes #4312` into a commit on the default branch and retire the issue. **The explicit `--body` is required.** See the answer at the bottom.

### 2. The corrected doc claim — true as stated, with an incomplete illustration

`filter.rs:860-877`. The load-bearing sentence is correct:

> The exposure is bounded to tokens that reach this branch, pass the charset gate, and carry neither an uppercase letter nor a digit

That **is** the complete characterisation. The branch is `has_b64_sym && is_plausible_b64_charset && (is_url_credential_shaped || has_upper || has_digit)`, so narrowing the OR term can only change a verdict when `has_upper || has_digit` is false; and the mixed-case fallback below requires `has_upper && has_digit`, so it cannot rescue. No uppercase and no digit is necessary and the bound holds.

The `i.e.` gloss that follows — *"a userinfo-free URL whose path secret is all-lowercase"* — under-enumerates the moved set. Two more shapes moved:

- **`://` URL with a colon-free userinfo**: `https://sometokenhere@hooks.host/path` — has userinfo, so not "userinfo-free", but `userinfo.contains(':')` is false, so the narrow predicate declines it.
- **`@`-bearing non-URL token**: `git@github.com:bobmatnyc/trusty-tools` — no `://` at all. `@` is excluded from `is_word_segment` (`filter.rs:624`), so `is_slash_path` fails and it reaches the branch. The broad predicate exempted it; the narrow one does not.

Both are all-lowercase and digit-free, so both fall inside the bound the sentence already states correctly, and neither is a plausible credential shape — the second is in fact one of the false positives the narrowing was *meant* to fix (named at `filter.rs:857`). No safety claim is overstated. **LOW, doc precision only.**

|---|---|---|---|---|---|
| LOW | `crates/trusty-common/src/memory_core/filter.rs` | 863-866 | The `i.e.` gloss names only "userinfo-free URL"; the moved set also includes colon-free-userinfo URLs and `@`-bearing non-URL tokens. The stated bound itself is complete and correct. | Widen the gloss to "a token the broad predicate exempted — a `://` URL without `user:pass@` userinfo, or an `@`-bearing non-URL token — whose secret content is all-lowercase". | Parent |

### 3. Known-miss assertions — live, not comments

`filter_tests.rs:1448-1458`, inside `real_secrets_still_blocked_after_4312_charset_gate` (`#[test]` at :1343; the file is wired in via `#[path = "filter_tests.rs"] mod tests` under `cfg(test)` at `filter.rs:1002-1004`). Both shapes are loop-driven `assert!(find_secret_token(missed).is_none(), …)` — real assertions, not decoration.

Direction is pinned **both** ways in the same function:
- becomes-caught → the `is_none()` known-miss asserts fail;
- becomes-missed → the must-flag battery at :1410-1418 (`webhook URL with secret path`, uppercase+digit form) asserts `is_some()` and fails.

Executed here on the PR head:

```
$ cargo test -p trusty-common --features memory-core --lib -- memory_core::filter --nocapture
running 50 tests
...
test memory_core::filter::tests::real_secrets_still_blocked_after_4312_charset_gate ... ok
test memory_core::filter::tests::url_prose_residue_is_a_known_pre_existing_bound ... ok
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 661 filtered out; finished in 0.01s
```

### 4. The `origin/main` merge — clean, no unintended content

Merge commit `60782d49`, parents `782c6e44` (branch) and `a5c51bda` (main at merge time).

```
$ git diff --stat origin/main...origin/fix/4312-secret-detector-backticks
 crates/trusty-common/changelog.d/4312-secret-detector-backticks.md |   8 +
 crates/trusty-common/src/memory_core/filter.rs                     | 155 +++-
 crates/trusty-common/src/memory_core/filter_tests.rs               | 439 +++++++-
 3 files changed, 586 insertions(+), 16 deletions(-)
```

Evil-merge check, both directions:

```
$ git diff --stat a5c51bda 60782d49        # merge tree vs merged-in main tip
 (exactly the same 3 files, 586 insertions, 16 deletions)

$ git diff --name-only 782c6e44 60782d49 | grep -E 'memory_core/filter|changelog.d/4312'
 (none)
```

The merge introduced no content of its own and did not touch the three PR files from the main side. Diff is the reported 3 files.

**Version state is coherent.** Branch and main both read `version = "0.28.1"` for `trusty-common`; crates.io's latest published is `0.28.0`. So `check-pr-version-bump` passes for the right reason — src changed, the version is above the published one, and `0.28.1` is not yet live — riding main's independent bump rather than a stale duplicate bump on this branch. All 27 checks green, including `Per-PR changelog fragment` and `PR version bump vs crates.io (issue #4421)`.

One non-blocking note: `origin/main` has since moved to `a22a5353` with further `trusty-search` commits, so the branch is behind again. `mergeable` is `MERGEABLE` and the head's checks are green, so this is not a gate — just be aware the fragment gate's verdict was computed against the older base.

### Merge instruction

**Yes — merge with an explicit body.** The PR body is clean but commit `9d7f7709` is not, and `squash_merge_commit_message=COMMIT_MESSAGES` would pull it into the squash commit.

```bash
gh pr merge 4723 --squash --body "Refs #4312"
```

The squash *title* needs no override: `squash_merge_commit_title=COMMIT_OR_PR_TITLE` resolves to the PR title with multiple commits, and that title carries no closing keyword.

---

Round-3 conclusion stands: ship the filter. Zero CRITICAL, zero HIGH. One LOW doc-precision note, disposition `Parent`.

Issue #4724, opened 2026-08-04
install_then_deploy_composes_agents asserts a stale bundled-agent count (6 vs 42)


bobmatnyc:
`cargo test -p trusty-mpm -- --include-ignored` is red on `origin/main` (a28698c8) because `tests_behavior_a::install_then_deploy_composes_agents` hardcodes the bundled-agent count.

```
thread 'tests_behavior_a::install_then_deploy_composes_agents' panicked at
crates/trusty-mpm/src/bin/tm/tests_behavior_a.rs:263:5:
assertion `left == right` failed
  left: 42
 right: 6
```

`crates/trusty-mpm/src/bin/tm/tests_behavior_a.rs:262-263`:

```rust
// All six bundled agents deploy on a fresh target.
assert_eq!(result.deployed.len(), 6);
```

The roster is now 42. The assertion should key off the bundled source count rather than a literal, so adding an agent never re-breaks it.

Confidence: high — reproduced on a clean detached `origin/main` worktree with no local changes.

Found while running the rung-5 gate for [#4702](https://github.com/bobmatnyc/trusty-tools/issues/4702); not caused by that PR (its diff touches zero files in the install/deploy path).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Folded into [PR #4726](https://github.com/bobmatnyc/trusty-tools/pull/4726) — that PR changes the bundled agent set, so this assertion is on its surface and a separate fix would collide.

Diagnosed before renumbering: `assert_eq!(result.deployed.len(), 6)` prints the actual as `left`, so `left: 42, right: 6` means the deploy produced **42** and the expectation was **6** — not a fixture seeing a fraction of the set. 42 is exactly the number of `agents/*.md` artifacts in `bundle::ALL` (37 roster agents + 5 `BASE-*` templates), so nothing is missing from the deploy path.

Fixed by deriving the expectation from the bundle manifest `install_to` just wrote, and comparing sets rather than lengths — an equal count with different membership would mean the deployer silently dropped an agent, which a length assertion cannot see.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #3714, opened 2026-07-23
tm relaunch-by-name resolves to the wrong record when two sessions share a name — blocks relaunch with dead-pane error + contradictory state banners


bobmatnyc:
## Summary

`tm` relaunch-by-name resolves the ambiguous session name `tm-tagents` to the WRONG record (a stale duplicate with a dead recorded pane) instead of the live one the operator is actually attached to, causing relaunch to fail with a dead-pane refusal error — preceded by two consecutive runs of the exact same command reporting **contradictory states** (`active` then `stopped`) seconds apart.

## Owner's verbatim transcript

Running `tm` inside the `tm-tagents` tmux session (from worktree `.base/.worktrees/f443c12d-2fb6-4ce1-9f70-2e7695306e47`):

**1st run:**
```
tm: this tmux session ('tm-tagents') is already a managed session (state=active) — not nesting;
reconnecting to 'tm-tagents' instead… switching client to session 'tm-tagents'
```

**2nd run (seconds later, same command, same pane):**
```
tm: this tmux session ('tm-tagents') is already a managed session (state=active) — not nesting;
reconnecting to 'tm-tagents' instead… switching client to session 'tm-tagents'
```
followed by:
```
session 'tm-tagents' is stopped but its tmux pane is still alive — the daemon will KILL that pane
and start a fresh runtime
```
then:
```
Error: cannot restart session 'tm-tagents': recorded pane %3 for session
7dabd521-1062-4645-b178-eec95979dccf no longer exists, but its tmux session is still alive
(likely a sibling window) — refusing to respawn into an unrelated active pane; close the sibling
window and delete/recreate this session...
```

## Captured evidence (read-only, from a separate session)

`tm sessions ls` shows **two records sharing the name `tm-tagents`**:

```
NUM  ID                                    STATE                    NAME        TASK              CREATED
11   f443c12d-2fb6-4ce1-9f70-2e7695306e47  attached                 tm-tagents  tcode focused s…  2026-07-14 05:19
23   7dabd521-1062-4645-b178-eec95979dccf  attached [stale-assets]  tm-tagents  adopted session   2026-07-18 19:17
```

`tm session info f443c12d-2fb6-4ce1-9f70-2e7695306e47` (the REAL, live, owner-occupied session):
```json
{
  "attached": true,
  "state": "active",
  "persisted_state": "active",
  "pane_id": "%652",
[...]

PR #4725, opened 2026-08-04
fix(trusty-mpm): tm ls prunes dead records on every invocation, incl. stopped

### LINKED ISSUE #4702: tm ls auto-prune misses stopped sessions and every non-TTY invocation, so dead records accumulate indefinitely
Reported by the owner 2026-08-03: dead sessions still show in `tm ls` after #4384 was believed to have fixed this.

#4384 (merged 2026-07-30, present in installed 1.3.3) is real, but narrow on two axes that compound:

1. **TTY-picker only.** The auto-prune fires from the interactive TTY session picker. Piped, scripted, and `--json` invocations take a different path via `managed.rs` that never calls it. Any non-interactive use never prunes.
2. **Unresumable only.** It targets records whose workspace directory is physically gone. It never touches **stopped** records — tmux pane died, worktree intact. Those are the majority of what users actually see as "dead".

## Measured evidence (2026-08-01)

- `tm ls` → 34 records; `tmux ls` → 18 sessions; all 18 matched 1:1, so nothing falsely claimed to be alive. The list accumulates; it does not lie about liveness.
- Daemon fleet view: **active 18, stopped 16, decommissioned 32, total 66** — **48 of 66 records stale with nothing auto-clearing them.**
- 4 of 5 spot-checked stopped workspaces still existed on disk.

## Why this is more than cosmetic

The manual verb (`session_prune` with state `stopped`) clears it, but nothing invokes it automatically, so the registry grows without bound. A `tm ls` that is mostly dead entries stops being usable as a session picker — which is its primary purpose since #2297.

## Open question that gates the fix

Whether clearing a `stopped` record is lossy is **not yet established**. 4 of 5 stopped workspaces still existed on disk, so those sessions may still be resumable, and the record may be the only thing making resumption possible. **Determine from the code whether pruning a stopped record forfeits a resumable session before making any prune automatic.** If it is lossy, an age threshold or a workspace-gone-only selector is likely the correct scope rather than blanket `stopped`.

## Suggested scope

- Move the prune out of the TTY-picker path so every `tm ls` invocation (piped, `--json`, scripted) prunes consistently.
- Extend coverage beyond unresumable to stopped, gated on the loss question above.
- Do NOT couple this to worktree deletion. Pruning a registry record must never remove a git worktree — see the 2026-07-21 incident where a cleanup destroyed a bare clone.

## Related

- #4384 — the partial fix this extends
- #4322 — the `assets ?` marker is a skipped-probe indicator, NOT a liveness signal; do not conflate the two when rendering `tm ls`

bobmatnyc:
Closes #4702

## The two defects

`tm ls`'s dead-record auto-prune (#4384) was real but narrow on two axes that compounded, so the registry grew without bound — **48 of 66 records stale** on the reporting machine, with nothing clearing them.

**1. TTY-picker only.** `fetch_live_sessions`'s `allow_auto_prune` was ever passed `true` only by callers gated on `stdin.is_terminal() && stdout.is_terminal()`. Every piped, scripted, and `--json` invocation took `managed::session_ls`, which never pruned at all.

The opt-in existed because auto-prune was read as "destructive-adjacent" (PR #4384 critic HIGH finding #3). It is not destructive: its only mutation is `decommission_dead_record`'s `record_only=true` call, verified against the response body. So the parameter is **removed** and both listing paths route through one shared `prune_and_report`. `managed::session_ls` gains a `session_ls_at` testable core so a test injects a tempdir marker file instead of writing the operator's real `~/.trusty-mpm/auto-prune-seen.json`.

**2. Unresumable only.** `session_picker_prune.rs` partitioned on `s.unresumable`. That flag is computed daemon-side (`summary::checked_summaries_with`) **only** for records whose PERSISTED state is `Stopped`/`Errored`; the list route then DISPLAY-reconciles state against live tmux (`reconcile_live_state`). A zombie — persisted `active`, tmux pane long gone — therefore renders as `stopped` while its `unresumable` probe never ran, so it reads `false` forever and no prune ever touched it.

The CLI now independently probes the workdir candidates the wire summary carries (`workspace_path`, `cwd`) for stopped/errored rows. Client-side deliberately, not daemon-side: the fix then works against an already-running older daemon, since a CLI upgrade never bounces it.

## Safety — this must not become a worktree deleter

`prune_managed(PruneFilter::Stopped, …)` calls the **full** `self.decommission(...)`, which for a clean in-project worktree runs `git worktree remove --force` (falling back to `fs::remove_dir_all`), `git worktree prune`, and `git branch -D session/<leaf>`. `decommission.rs` calls it "a live data-loss path" in its own source comment. Auto-prune must not inherit any of that.

- It stays on `decommission_record_only` (`?record_only=true`), whose response body must report `workspace_removed == false` before a call counts as pruned.
- 🔴 `auto_prune_always_requests_record_only_never_full_teardown` asserts on **the path taken, not the end state**: a recording stub captures every decommission request and every one must carry `record_only=true`. A request without it silently takes the full-teardown branch.
- A `stopped` record is cleared **only** when its workspace is independently verified gone. Measured 2026-08-03: 4 of 5 spot-checked stopped workspaces still existed on disk — those stay. `workspace_verified_gone` is fail-closed-toward-keeping: a probe `Err` (unmounted volume, permission denied) counts as "possibly present," and a record carrying no path at all is unverifiable and kept.
- `decommissioned` records (35 on the machine) are out of scope and excluded by an explicit state gate, on both display and persisted state.
- Unchanged from #4384: two-sightings confirmation, the per-call cap of 5, and the stale-daemon sentinel.
- `is_running()`'s real tmux liveness probe still runs upstream, so a persisted-`stopped`-but-actually-alive session (e.g. `tm_cto_h2`) renders `active` and is never a candidate.

## Changes

| File | What |
|---|---|
| `session_picker_prune.rs` | `workspace_verified_gone` / `is_clearable_state` / `is_dead_record` (the widened predicate); `prune_and_report` + `prune_and_report_at` + `default_marker_path` (the shared entry point, reporting moved here) |
| `session_picker.rs` | `fetch_live_sessions` drops `allow_auto_prune` and always prunes |
| `managed.rs` | `session_ls` now prunes; `session_ls_at` testable core; `--json` re-GETs only when the prune changed something, so the passthrough stays a real daemon response |
| `guided.rs` | bare `tm` drops its TTY pre-gate on the fetch |
| `tests_behavior_d_tests.rs` | 6 new tests; removed `fetch_live_sessions_skips_auto_prune_when_disallowed` (it asserted the behavior this issue identifies as the defect) |
| `changelog.d/4702-tm-ls-prune-coverage.md` | fragment |

## Test ladder rung 5 — process lifecycle, destructive-adjacent

```
cargo fmt --check                                     → FMT OK
cargo check -p trusty-mpm --all-targets               → Finished
cargo clippy -p trusty-mpm --all-targets -- -D warnings → Finished, 0 warnings
cargo test -p trusty-mpm                              → 4290 passed; 0 failed (lib)
                                                        1381 passed; 0 failed (bin/tm)
                                                        + all integration binaries, 0 failed
cargo test -p trusty-mpm -- --include-ignored         → 4297 passed; 0 failed (lib)
                                                        1381 passed; 1 failed (bin/tm) — see below
bash scripts/check_line_cap.sh                        → 3603 files, 0 violations
bash scripts/check_sld.sh                             → 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh              → OK trusty-mpm
bash scripts/check_test_pointers.sh                   → 20455 citations, 0 dangling
```

### Fail-before evidence

Reverting only `is_dead_record` to `s.unresumable`:

```
---- auto_prune_clears_stopped_record_whose_workspace_is_gone ----
  left: 0
 right: 1
---- auto_prune_always_requests_record_only_never_full_teardown ----
assertion `left == right` failed: both dead records must be cleared
  left: 1
 right: 2
```

Reverting only `session_ls_at`'s prune calls:

```
---- session_ls_prunes_dead_records_on_piped_invocation ----
panicked: a non-TTY `tm ls` must clear a confirmed dead record from the registry
---- session_ls_json_passthrough_prunes_dead_records ----
panicked: `tm ls --json` must clear a confirmed dead record from the registry
```

### Known pre-existing red, NOT this PR

`--include-ignored` surfaces one failure, `tests_behavior_a::install_then_deploy_composes_agents` (`left: 42, right: 6` — a stale hardcoded bundled-agent count). Reproduced identically on a clean detached `origin/main` worktree at `a28698c8`; this diff touches zero files in the install/deploy path. Filed as #4724.

Every gate specific to this change passes.


bobmatnyc:
## Code-critic verdict: BLOCK

Adversarial review of PR #4725 against issue #4702, from a detached worktree on `3d8664b9`. Every claim below was executed, not read.

**No path in this diff deletes a worktree, a branch, or a file.** I traced `tm ls` → `session_ls_at` / `fetch_live_sessions` → `prune_and_report_at` → `auto_prune_dead_records_at` → `decommission_dead_record` → `POST …/decommission?record_only=true` → `decommission_managed_session` (`managed_routes/mod.rs:879`) → `decommission_record_only` → `decommission_with_root(record_only = true)`, where both removal branches are behind `if !record_only` (`decommission.rs:444`). That part of the safety argument holds.

**The safety test is not vacuous.** I deleted `.query(&[("record_only", "true")])` from `decommission_dead_record` and re-ran it:

```
thread 'tests_behavior_d::auto_prune_always_requests_record_only_never_full_teardown' panicked at
crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs:2403:9:
every decommission the sweep issues must be RECORD-ONLY — a call without it takes the full
worktree-removal teardown: flagged?
```

Unmutated, all 12 `auto_prune_*` and both `session_ls_*` tests pass. Every repo gate passes in my worktree (`check_line_cap` 0 violations, `check_sld` 0/0, `check_test_pointers` 20455 citations / 0 dangling, `check_changelog_fragment` OK). The version bump and the 1-line lockfile diff are correct.

The block is not about the filesystem. It is about what `record_only` does *not* skip.

---

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/session_manager/decommission.rs` | 435 | `graceful_terminate_runtime` runs **before** the `record_only` guard, so a "record-only" prune SIGTERMs the session's `claude` process and `kill_session`s its tmux pane **by name**. The gate removed by this PR was justified in writing by the opposite claim. | Move the runtime teardown inside `if !record_only`, or gate it on a pane-id-scoped liveness check; then re-justify removing `allow_auto_prune`. | Fix here |
| HIGH | `session_picker_prune.rs` | 280–292 | `workspace_verified_gone` reads a **missing mount point as "verified gone"** — `try_exists` returns `Ok(false)`, not `Err`, for a path under a non-existent parent. The doc at 271–274 claims the opposite. | Require an ancestor (parent, or the mount root) to be present before treating `Ok(false)` as definitive; otherwise return `false`. | Fix here |
| HIGH | `session_picker_prune.rs` | 416 | The two-sightings guard is `seen.contains_key(&s.id)` — **presence only, no minimum interval**. Making every invocation prune removes the rarity that made "two sightings" a temporal guard: two `tm ls` calls milliseconds apart now confirm. The per-call cap of 5 stops being a rate limit. | Compare the already-persisted RFC-3339 first-sighting timestamp against a minimum age (the value is stored and currently never read). | Fix here |
| MEDIUM | `managed.rs` | 153–155 | The `--json` re-GET propagates with `?` **after** the prune already mutated the registry, so `tm ls --json` can hard-fail on a path that previously could not. The doc at 117–119 promises the opposite ("degrades to a plain passthrough … rather than failing the command"). | `if let Ok(fresh) = fetch_managed_raw(…).await { raw = fresh }`. | Fix here |
| MEDIUM | `tests_behavior_d_tests.rs` | 2145+ | No test covers the two failure modes that matter: a probe `Err`, and the runtime side effect of a record-only decommission. `decommission_record_only_never_removes_existing_workspace` asserts only on the filesystem, and its fake driver has no live session, so the kill path never executes. | Add a `FakeTmuxDriver` test seeding `seeded_names` with the record's `tmux_name` and asserting `kill_calls` is empty after a record-only decommission. | Fix here |
| LOW | `session_picker_prune.rs` | 142 | Dangling intra-doc link `[`auto_prune_dead_records`]` — the function was renamed to `prune_and_report`. Not caught by `check_test_pointers.sh` (that lint only resolves `Test:` citations). | Update the link. | Fix here |

---

## Evidence for the CRITICAL

`decommission_with_root` calls the runtime teardown unconditionally at line 435, nine lines above the `if !record_only` that guards removal:

```rust
self.graceful_terminate_runtime(&record.tmux_name).await;   // ← line 435, no record_only guard

let mut workspace_removed = false;
if !record_only && let Some(ref ws) = record.workspace_path {   // ← line 444
```

`graceful_terminate_runtime` (`restart_ops.rs:139`) resolves liveness with `session_exists(tmux_name)` — which is **name membership in `list_sessions()`** (`driver.rs:363`), not the record's captured `pane_id` — then `signal_terminate` (SIGTERM to the `claude` pid, else Ctrl-C) and `kill_session(name)`.

I proved it by executing it, with a fake driver seeded so a live tmux session carries the record's name:

```
REVIEW PROBE tmux_name=tm-s1-01 kill_calls=["tm-s1-01"] interrupt_calls=["tm-s1-01"]
test session_manager::tests::review_probe_record_only_still_kills_the_tmux_runtime ... ok
```

(That probe was temporary; the worktree is clean.)

### Why this is load-bearing here rather than pre-existing noise

The PR's own justification for deleting `allow_auto_prune` is a factual claim, stated three times in the reviewed artifact:

- `session_picker_prune.rs:189–195` — *"it is not, and cannot be: the ONLY mutation it performs is `decommission_dead_record`'s `record_only=true` call … which never removes a git worktree, a branch, or any file on disk."*
- `session_picker.rs:201–206` — *"it is not destructive at all … so every listing surface now prunes identically. The non-destructiveness guarantee is what makes that safe."*
- `decommission.rs:348` — *"Tombstone a record WITHOUT ever touching the filesystem."*

Each is true about the filesystem and each omits that a live `claude` process is signalled and its pane reclaimed. SIGTERMing an agent mid-run is a larger loss than a file. The gate was removed on a premise that does not survive execution.

### Reachable paths

The prune requires **display** state `stopped`/`errored`, so in the normal case the record's own pane is confirmed gone and the kill is a no-op. It is not a no-op here:

1. **Name collision.** `reconcile_live_state`'s own doc (`summary.rs:158–165`) states the condition as real: *"under the #3692 duplicate-name condition, a stale record's own pane can be long gone while an UNRELATED live session (owned by a DIFFERENT record) still happens to share its name."* Pane-scoped liveness returns `Some(false)` → display `stopped` → clearable. The subsequent `kill_session(name)` then targets the unrelated live session. This is exactly the hazard #3714 added `pane_exists_checked` to contain — and this diff routes around the containment by keying teardown on the name.
2. **`errored` is never display-reconciled.** `reconcile_live_state:213` only rewrites state for records persisted `Active | Stopped`. An `Errored` record reads `errored` no matter what tmux says, and `is_clearable_state:308` accepts `errored`. The PR body's assurance that *"`is_running()`'s real tmux liveness probe still runs upstream"* does not cover this class — the auto-prune path relies on `reconcile_live_state`, not `is_running()` (which belongs to the manual `prune_managed` verb).
3. **`reconcile_against_tmux` skips reconciliation entirely on an enumeration error** (`summary.rs:253`), serving persisted state. A record persisted `Stopped` with a live pane — the drift you documented on this machine as `tm_cto_h2` — then displays `stopped`.

Each of these needs the workspace probe to also read "gone", which is where the HIGH below compounds it.

## Evidence for HIGH #1 (unmounted volume reads as gone)

The doc asserts the fail-closed rule covers unmounted volumes. It does not — ENOENT on a missing ancestor is `Ok(false)`, which the predicate treats as definitive absence:

```
"/Volumes/NoSuchExternalDrive/work/proj" -> Ok(false)
""                                       -> Ok(false)
"relative/path"                          -> Ok(false)
```

Unplug an external drive and every session whose workspace lived on it becomes "verified gone" on the next two listings. `Err` is reserved for cases like a stale NFS handle or an EACCES ancestor; a plain unmount is not one of them.

Two smaller consequences of the same probe: an empty-string path reads gone, and a **relative** path is resolved against the **CLI's** cwd, not the daemon's — so the doc's claim at 246–249 that this predicate is strictly narrower than the daemon's `is_unresumable` does not hold universally.

## On the questions the review was asked

- **`record_only` honored downstream?** Yes for the filesystem (`mod.rs:879` → `decommission_record_only`); no for the runtime. It is not a flag a callee ignores — it is a flag whose scope is narrower than its name and its documentation.
- **Partial failure mid-prune?** Consistent. `Failed` keeps the record and leaves its marker so the next call retries; `StaleDaemon` breaks, folds the remainder back into `kept` via `iter.by_ref()`, and writes the sentinel; a lost `save_seen` only restarts a confirmation window. No finding.
- **Is the old partition structurally unreachable for zombies?** Yes, confirmed from source. `checked_summaries_with` computes `unresumable` only for persisted `Stopped | Errored` (`summary.rs:607`), and `reconcile_live_state` rewrites display state afterward, so a persisted-`Active` record with a dead pane renders `stopped` while `unresumable` stays `false`. The load-bearing justification for the new predicate is sound.
- **`--json` consumers.** Milder than feared: `fetch_managed_raw` is unfiltered, so pruned records still appear in `--json` — with `state` flipped to `decommissioned` — rather than vanishing. Scripts filtering on state will see the change; scripts keying on id will not break. The re-GET's real cost is the new failure mode in the MEDIUM above.
- **Cap of 5.** Coherent per call, incoherent per unit time now that every invocation prunes. See HIGH #2.

## Required changes

1. Gate `graceful_terminate_runtime` on `!record_only` (or on pane-id-scoped liveness), and correct the three doc sites plus the changelog fragment that assert non-destructiveness. If the runtime teardown is intentional for a record-only prune, say so explicitly and re-argue the removal of `allow_auto_prune` on those terms.
2. Make `workspace_verified_gone` require a present ancestor before treating `Ok(false)` as definitive; fix the doc's `Err` claim.
3. Enforce a minimum age on the first-sighting timestamp already persisted in the marker file.
4. Stop `?`-propagating the `--json` re-GET.
5. Add the two missing tests: a record-only decommission asserting `kill_calls` stays empty, and a probe-`Err` case for `workspace_verified_gone`.

## Notes

- The `workspace_removed == false` version-skew guard is not proof of record-only even on the intended path: a stale daemon running the full `decommission()` against an unowned/local-path workspace, or refusing a dirty worktree, also returns `false`. Unchanged code, and its failure modes there are not filesystem deletion — but this PR leans on it harder than #4384 did.
- `install_then_deploy_composes_agents` (#4724) and the Tauri `ui/dist` workspace-check failures were treated as out of scope per the brief and are not counted against this diff.
- The engineering outside the safety argument is good: the zombie diagnosis is correct and verified from source, the fail-before evidence is real, the `session_ls_at` / `prune_and_report_at` seams are the right shape, and the state gate correctly excludes `decommissioned` on both display and persisted state.

bobmatnyc:
## Review fixes pushed — `9f6cab0e`

The BLOCK was correct and my safety argument was wrong. I verified the finding myself before fixing it: `graceful_terminate_runtime` sat at `decommission.rs:435`, nine lines above the `if !record_only` guard at 444, and `session_exists` is plain name membership (`driver.rs:363`).

### 1. `record_only` now actually skips runtime teardown

Gated the call in place rather than moving it below the workspace block — moving it would have inverted #1975's "terminate before removing" ordering for the normal path.

Fail-before, with the pre-fix line ordering restored:

```
thread 'session_manager::tests::decommission_record_only_never_touches_the_runtime' panicked at
crates/trusty-mpm/src/session_manager/tests.rs:917:5:
record-only must never signal the runtime; got ["tm-live-01"]
```

The test asserts on the driver's call log (`interrupt_calls` / `kill_calls` / `graceful_stop_calls` all empty) and seeds `seeded_names` first, with an explicit precondition assert so it can't go vacuous if the self-guard changes. `decommission_full_still_terminates_the_runtime` pins that the #1975 path is unaffected.

### 2. Shipped-defect investigation → [#4728](https://github.com/bobmatnyc/trusty-tools/issues/4728)

It ships in 1.3.3/1.3.4 (installed `tm --version` = 1.3.4), and it is **reachable with no name collision at all**:

- `is_unresumable` gates on persisted state + filesystem probes and **never consults tmux liveness**;
- `mark_errored` (`manager.rs:1106`) is a pure record mutation that never touches the pane.

So any `Stopped`/`Errored` record whose workspace was deleted while its pane is still alive was a kill target on the next `tm ls`. The name-collision route also exists but is narrower — tm's own spawn/rename/adopt paths dedupe against non-terminal record names via `taken_name_set`.

Filed separately as you asked. I did **not** split the code fix out: it is the same line #4725 must change for its own safety argument to hold, so a second PR would leave this one unmergeable and the shipped defect unfixed. #4728 carries the traceability and the release note; the fix and its tests are here.

### 3. Workspace probe

`workspace_verified_gone` now also requires each candidate's **parent** to exist. And I removed the `if s.unresumable { return true }` short-circuit entirely — the daemon computes that flag with the same bare `try_exists`, so trusting it alone would have routed an unplugged volume straight past the new check. Every record now needs the client's own confirmation. Cost: a legacy record with neither `workspace_path` nor `cwd` on the wire is now kept rather than cleared. Unverifiable is not dead.

New: `auto_prune_keeps_record_whose_parent_directory_is_unreachable`, `auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable`.

### 4. Time bound restored

`sighting_is_confirmed` compares the persisted RFC 3339 stamp against `SIGHTING_MIN_AGE_SECS` (10 min). An unparseable stamp restarts the window instead of satisfying it. New: `auto_prune_does_not_confirm_two_calls_in_quick_succession` (10 back-to-back listings, 0 pruned, 0 decommission calls issued), `auto_prune_confirms_once_the_sighting_window_has_elapsed`, `auto_prune_treats_an_unparseable_sighting_as_a_fresh_one`.

Every pre-existing prune test now backdates the persisted stamp to simulate elapsed time against the real constant — no test weakens the guard, and the stale-daemon fixtures got real absent-with-existing-parent workspaces instead of a bare `unresumable: true`.

### 5. `--json` re-GET removed

Echoes the body it already fetched. Both problems you named go away: no `decommissioned`-flipped row, and no post-mutation `?`. New: `session_ls_json_never_refetches_after_pruning` counts GETs against a stub across a call that genuinely prunes — exactly one.

### Also

`is_clearable_state` now excludes `attached` records. You were right that `errored` rows are never display-reconciled, so "a live session renders active" did not cover them; `attached` is the one liveness signal such a row carries. I documented the residual honestly rather than overclaiming: an errored record whose pane is live but **detached** can still be cleared. That is bounded — clearing is record-only (it can no longer kill the pane) and still requires the workspace to be verifiably gone.

I also corrected every doc comment that repeated the wrong "not destructive at all" claim, and the module doc now says record-only is a claim pinned by tests, never a property the name guarantees.

## Gates — rung 5, re-run in full

```
cargo fmt --check                                       → FMT EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings → CLIPPY EXIT=0
cargo test -p trusty-mpm                                → TEST EXIT=0
  test result: ok. 4292 passed; 0 failed; 7 ignored (lib)
  test result: ok. 1388 passed; 0 failed; 1 ignored (bin/tm)
  45 test binaries, all "test result: ok"
cargo test -p trusty-mpm -- --include-ignored           → TEST EXIT=101
  test result: ok.     4299 passed; 0 failed; 0 ignored (lib)
  test result: FAILED. 1388 passed; 1 failed; 0 ignored (bin/tm)
bash scripts/check_line_cap.sh        → 3603 files, 0 violations
bash scripts/check_sld.sh             → 0 errors, 0 warnings
bash scripts/check_test_pointers.sh   → 20461 citations, 0 dangling
bash scripts/check_changelog_fragment.sh → OK trusty-mpm
```

The single `--include-ignored` failure is unchanged and still not mine:

```
---- tests_behavior_a::install_then_deploy_composes_agents stdout ----
  left: 42
 right: 6
```

Reproduced identically on a clean detached `origin/main` worktree at `a28698c8`; filed as [#4724](https://github.com/bobmatnyc/trusty-tools/issues/4724). This diff touches zero files in the install/deploy path.

Not merging — over to you for the re-review.

bobmatnyc:
## Verdict: BLOCK

Re-review of `9f6cab0e` (milestone tm 1.3.5). Read-only worktree, `.claude/worktrees/critic-4725-r2`; mutations run in a throwaway sibling.

**The five required changes from the prior round are all genuinely done, and the two new runtime tests are real.** The block is a different call on the same path: `record_only` gates the runtime teardown but still does not gate the trusty-search index deletion — and auto-prune's precondition drives that deletion at the *wrong* index.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/session_manager/decommission.rs` | 437, 600-602 | `record_only` does not gate the trusty-search index deletion. Auto-prune's own precondition (workspace verified gone) is the exact state `disposable_workspace_index_id`'s caller contract forbids, so it derives the **parent project's** index id and destroys it with `?delete_data=true`. | Move both the derivation and `delete_search_index_best_effort` inside `if !record_only`. | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 491-500 | The sighting window **resets on every unconfirmed listing**. A record dead 12 min but listed at T+6 never confirms. Any `tm ls` cadence tighter than 10 min leaves auto-prune permanently inert. | `seen.entry(s.id.clone()).or_insert_with(\|\| now.clone())`; restamp only the unparseable case. | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 367 | `attached` is a partial liveness signal; an errored record with a live but **detached** pane is still tombstoned, permanently orphaning a running pane from `tm`. A complete signal was available without a daemon change. | Enumerate live tmux session names client-side and exclude any record whose `tmux_name` is live. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs` | 432-512 | `session_ls_json_never_refetches_after_pruning` does not pin its own precondition — it passes with `is_dead_record` stubbed to always-`false`. | Assert the stub recorded one decommission before asserting `gets == 1`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 324-331 | The probe corroborates only the **immediate** parent. Correct for macOS `/Volumes/<name>` (the mount point vanishes), wrong for a mount point that persists as an empty directory when unmounted (autofs, `/mnt/*`, a stale NFS/SMB mount). | Require the grandparent too, or require the parent to be non-empty. | Promote |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/mod.rs` | 855 | Route doc still says `record_only` "never touches disk" — the same claim CRITICAL-1 falsifies. | Update alongside the CRITICAL fix. | Fix here |

## CRITICAL-1 — the evidence

`decommission_with_root` derives the index id at line 437 and deletes it at 600-602, both **outside** the `record_only` gate:

```rust
let search_index_id = search_gc::disposable_workspace_index_id(
    record.workspace_path.as_deref(), record.workspace_owned);
...
if let Some(index_id) = search_index_id {
    search_gc::delete_search_index_best_effort(&index_id).await;   // ← runs on record_only
}
```

`delete_search_index_best_effort` issues `DELETE {base}/indexes/{id}?delete_data=true` (`search_gc.rs:118`). That flag destroys the on-disk data — trusty-search's own `delete_index_with_delete_data_true_destroys_data` asserts the data dir does not survive, and trusty-search's own orphan reaper deliberately uses `delete_data=false` ("never destroy on-disk data automatically", `tickers.rs:417`).

Now the id. `disposable_workspace_index_id`'s doc carries an explicit caller contract (`search_gc.rs:60-65`):

> **CRITICAL (caller contract): must be invoked BEFORE the workspace directory is removed from disk** — once the directory is gone, `resolve_project_root`'s upward walk could resolve to an ancestor's (e.g. the shared base clone's) `.git` and derive the WRONG id.

Auto-prune only ever fires when the workspace **is already gone** — that is `workspace_verified_gone`'s entire job. So every auto-prune call violates that contract by construction. Probe, run against this head:

```
PROBE derived index id for GONE worktree = Some("my-project")
  left: Some("my-project")
 right: Some("tm-my-project-01")
```

A removed `<project>/.worktrees/tm-my-project-01` derives `"my-project"` — the **parent project's** index, not the session's. `is_session_worktree` is a pure path-string check, so it still returns `true` for the absent leaf; `find_git_root` then walks up past the two missing components to the project's real `.git`.

Net effect on the widened path this PR ships: every `tm ls`, `tm ls --json`, piped listing, and bare `tm` can issue up to 5 × `DELETE /indexes/<parent-project>?delete_data=true`. Two harms, one certain and one severe:

- **Certain:** the orphan index the sweep is *supposed* to remove (`tm-my-project-01`) is never removed — #2033's stated purpose silently fails.
- **Severe:** a real, populated project index is destroyed with its data. It is not registered on this machine *at this instant* (14 indexes live, no `trusty-tools`), but that is timing, not a guard — `resolve_project_root` is the same walk trusty-search's `detect_project` uses, so a project basename index is the normal steady state for any indexed project.

This is not a pre-existing bug you can wave through. The PR's central safety argument, stated three times in new prose, is that the opt-in gate is safe to remove because the prune "never removes a git worktree, a branch, or **any file on disk**" (`session_picker_prune.rs:210-211`) and record-only means "no filesystem, and no runtime" (`decommission.rs:456-457`). That argument is what justifies firing on every invocation, and it does not hold.

## HIGH-1 — the time bound can be defeated, and defeats itself

`SIGHTING_MIN_AGE_SECS`'s doc says the window measures time "between a record's **FIRST** sighting as dead and the call that may act on it." The code measures time since the **most recent unconfirmed** sighting — line 497 restamps a valid-but-too-recent entry with `now`:

```rust
if sighting_is_confirmed(seen.get(&s.id)) { confirmed.push(s); }
else { seen.insert(s.id.clone(), now.clone()); ... }   // ← also fires for "too recent"
```

Probe: first sighting at T, an ordinary listing at T+6, a listing at T+12.

```
panicked: record has been dead 12min (>10min window) but was listed at T+6;
          got pruned=0 calls=[]
  left: 0
 right: 1
```

So the window is really "a ≥10-minute *gap between consecutive listings*." An operator or PM session that lists more often than that prunes nothing, ever — on the machine with 48 of 66 stale records that motivated the issue. The PR body's coherence claim ("22 stopped records clear over ~6 listings across an hour") does not hold under this code.

Asked directly: **the 10-min + cap-5 rate limit is coherent as a design and incoherent as implemented.** The design intent is sound; this line breaks it.

No test catches this because `backdate_sightings` overwrites the stamp *after* each call, so the reset is never observed. Worth a regression test alongside the fix.

## HIGH-2 — the `attached` residual is the same bug in a smaller box

Verified the author's rationale: `reconcile_live_state` (`summary.rs:212-218`) sets `summary.attached` for every record but rewrites `state` only for persisted `Active`/`Stopped`, so an `Errored` record with a live pane is served `state: "errored"`. `attached` is indeed the only wire-visible liveness signal it carries.

But `attached` proves a *client* is attached, not that no pane exists. An errored record with a live, detached pane and a gone workspace is still cleared. Post-#4728 the pane is not killed — the pane keeps running while its record becomes a `Decommissioned` tombstone, which is terminal: no resume, no reattach through `tm`, and the row is excluded from the picker. A running agent silently becomes unreachable through the tool that owns it.

The residual is avoidable, which is why I am not accepting it. Two complete signals were on the table:
1. The CLI runs on the same host as tmux. Enumerating live session names client-side needs no daemon change and preserves the "works against an already-running older daemon" property the author (correctly) optimised for.
2. `reconcile_live_state` already computes `is_live` at line 201-211 and throws it away; surfacing it costs one field.

Option 1 keeps the PR's own constraint intact.

## What I verified and accepted

**1 — the in-place gate is complete for the runtime.** Read `decommission_with_root` end to end (412-631). Above the guard there is only a record read, the #3649 owner gate, and a pure path derivation — no runtime call. `graceful_terminate_runtime` (`restart_ops.rs:139-156`) is the sole runtime-affecting call, and its full surface is `signal_terminate` + `kill_session`, which is exactly what the test asserts on. Gating in place rather than moving below the workspace block was the right call: #1975's ordering still holds for the normal path (terminate at 458-460 precedes the removal block at 469) and is now positively pinned by `decommission_full_still_terminates_the_runtime`.

Mutation MUT1 (gate removed) reproduces the author's fail-before evidence byte-identically:

```
panicked at crates/trusty-mpm/src/session_manager/tests.rs:917:5:
record-only must never signal the runtime; got ["tm-live-01"]
```

**2 — the shipped-defect reachability claim HOLDS.** Verified independently of the author's write-up:
- `is_unresumable` (`resume_workdir.rs:131-148`) branches on persisted state and `tokio::fs::try_exists` only. Zero tmux liveness.
- `mark_errored` (`manager.rs:1106-1116`) sets `state` + `task` and upserts. It never touches the pane.

So an errored record with a live pane and a deleted workspace reads `unresumable == true`, and pre-#4728 `decommission_record_only` reached `graceful_terminate_runtime`, whose self-guard is `session_exists(name)` — plain name membership, satisfied by the record's own live pane. **No name collision is required.** 1.3.3/1.3.4 could SIGTERM and `kill_session` a live session on any interactive `tm ls`.

Keeping the fix in this PR rather than splitting it was right. It is the same line, and shipping 1.3.5 with the auto-prune widened to every invocation over an *ungated* teardown would have been strictly worse than either alternative.

**3 — the short-circuit removal is correct and its cost is bounded.** Dropping `if s.unresumable { return true }` is right: `is_unresumable` probes with a bare `try_exists`, so it flags an entire unmounted volume dead, and a short-circuit would route that past the new parent check. The stated cost — a legacy record with neither `workspace_path` nor `cwd` on the wire is kept — is correctly bounded, because keeping is the safe direction. Pinned by `auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable`; MUT3 kills it. The probe does distinguish unreachable-ancestor from absent on macOS; see MEDIUM-2 for where it does not.

**5 — the `--json` re-GET removal is clean.** Nothing else depended on it. The only consumer was the printed bytes; `source_id` is still used by the initial `fetch_managed_raw`. The behavior change (a just-pruned row appears in this invocation's output, gone from the next) is documented at `managed.rs:182-190` and is strictly better than the old failure mode, where the second fetch's `?` could fail the command after the registry had already been mutated.

## Test quality

- **`decommission_record_only_never_touches_the_runtime` is not vacuous.** It asserts on the driver call log (`interrupt_calls` / `kill_calls` / `graceful_stop_calls` all empty), which is the observable effect and the complete surface of the helper — not a flag. It seeds `seeded_names` and carries a precondition assert on `fake.session_exists(...)`, so it cannot go vacuous by the helper's own fast-path. Confirmed: MUT1 fails it for the right reason.
- **Mutation results** — each mutation killed exactly its own test, no collateral:

| Mutation | Test killed |
|---|---|
| Remove `if !record_only` gate | `decommission_record_only_never_touches_the_runtime` |
| Remove `s.attached` from `is_clearable_state` | `auto_prune_never_touches_an_attached_record` |
| Remove the parent-directory check | `auto_prune_keeps_record_whose_parent_directory_is_unreachable`, `auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable` |
| `is_dead_record` → always `false` | **none** — `session_ls_json_never_refetches_after_pruning` still passes (MEDIUM-1) |

- **Backdating did not weaken the pre-existing tests.** It rewrites the persisted stamp and lets the real comparison run against the real constant, and it deliberately skips the stale-daemon sentinel key. The two tests that hand-seed `2020-01-01` are unaffected. It does, however, *mask* HIGH-1: because it overwrites the stamp after each call, no test ever observes an intermediate listing.

Full suite on this head, for the record:

```
cargo test -p trusty-mpm
  lib:    4292 passed; 0 failed; 7 ignored
  bin/tm: 1388 passed; 0 failed; 1 ignored
  + all integration binaries, 0 failed
```

`install_then_deploy_composes_agents` (#4724) and the Tauri `ui/dist` failures were treated as known pre-existing and are not findings.

## Required changes

1. Move the search-index derivation and `delete_search_index_best_effort` inside `if !record_only` in `decommission_with_root`, and add a test that a record-only decommission of a session with a `.worktrees/` workspace issues no index deletion. Update the three prose claims that assert record-only touches no disk (`decommission.rs:456-457`, `session_picker_prune.rs:20-21` and `:210-211`, `managed_routes/mod.rs:855`).
2. Stop restamping a valid-but-too-recent sighting; keep the restamp for the unparseable case only. Add a regression test with an intermediate listing inside the window.
3. Replace or supplement the `attached` guard with a definitive liveness check — client-side tmux session enumeration keeps the no-daemon-change property.
4. Pin `session_ls_json_never_refetches_after_pruning`'s precondition by asserting the decommission actually fired.

## Notes

- MEDIUM-2 is the one finding I would not hold the PR for on its own. trusty-mpm is macOS-first and `/Volumes/<name>` disappears on unmount, which the current check handles correctly.
- Findings 2 and 3 are both "the feature does not do what it claims" rather than "the feature destroys something." If schedule pressure forces a split, CRITICAL-1 is the one that cannot ship.

bobmatnyc:
## Round-2 fixes pushed — `6cc69822`

I verified the index finding myself before acting: `delete_search_index_best_effort` sits at `decommission.rs:600`, below every guard, and `disposable_workspace_index_id` derives its target at line 437 via a `.git` walk-up that its own doc says requires the workspace to still exist.

## Your architectural question: dedicated path

I took it. The pattern argument is correct and I'd have kept feeding you one effect per round otherwise.

`decommission_record_only` is now a standalone function. Its entire body is record mutation plus one store write — there is no filesystem call, subprocess, network request, or tmux interaction in it to forget to gate. `decommission_with_root` **loses the `record_only` parameter entirely**; the two share no code path except an extracted `check_worktree_owner`.

### The enumeration you asked for

`decommission_with_root`'s side effects, now a table in its own doc:

| # | Effect | Class |
|---|---|---|
| 1 | `graceful_terminate_runtime` — SIGTERM + `kill_session` | destructive |
| 2 | `remove_session_worktree` — `git worktree remove --force`, `remove_dir_all` fallback, `git worktree prune`, `git branch -D` | destructive |
| 3 | `fs::remove_dir_all` on an SM-owned workspace | destructive |
| 4 | `delete_search_index_best_effort` — cross-daemon `DELETE /indexes/{id}` | destructive |
| 5 | clears `workspace_path`/`workspace_owned` when nothing is on disk | store-only |
| 6 | clears `pending_decision`/`proposed_default` (#4400) | store-only |
| 7 | writes `Decommissioned` | store-only |

The record-only path implements 5–7 and structurally cannot reach 1–4. Effects 1–3 are pinned by `decommission_record_only_has_no_side_effects_beyond_the_store` (real populated workspace + live tmux name; asserts content survives and the driver's call logs are empty). Effect 4 is pinned by `disposable_index_id_for_a_removed_worktree_resolves_to_the_parent_project`, which proves the derived id for a removed worktree **equals the parent project's id**.

One deliberate omission: I did not write a test that lets the old path issue a real DELETE. On a machine with a live trusty-search daemon that would destroy this project's index — the exact hazard. The derivation test pins the same fact hermetically.

## The other three

**Window reset** — confirmed and fixed. `classify_sighting` returns a three-state `Sighting`; `TooRecent` leaves the stamp untouched. You were also right that the helper hid it: `backdate_sightings` overwrites, so I added `age_sightings_by`, which is **relative** and therefore carries a restamp through to the assertion. Rewrote `auto_prune_confirms_despite_frequent_intervening_listings` to use it — T+0 sighting, listing at T+6, confirm at T+11. My first version of that test passed with the bug; the mutation run caught it and I fixed the test.

**`attached` residual** — closed, not documented away. The CLI enumerates live tmux via the shared `TmuxCommand::ListSessions` entry point and never clears a record naming a live session. Failure to enumerate prunes nothing (`reconcile_against_tmux`'s fail-closed contract). Name membership is used only to KEEP here — a collision costs a stale row; the same signal used to KILL is what made #4728 destructive.

**MEDIUM** — `session_ls_json_never_refetches_after_pruning` now counts decommission POSTs and asserts exactly one fired, so the GET count is no longer vacuous.

## Mutation testing — all six caught

| Mutation | Result |
|---|---|
| M1 restamp inside window | 2 tests FAILED |
| M2 drop live-tmux guard | 1 test FAILED |
| M3 `None` live-set prunes anyway | 1 test FAILED |
| M4 record-only routed back onto the teardown | 2 tests FAILED |
| M5 `--json` re-GETs | 1 test FAILED (`left: 2, right: 1`) |
| M6 `is_dead_record` → always-false | 1 test FAILED on the precondition |

## Gates — rung 5, full re-run

```
cargo fmt --check                                       → FMT EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings → CLIPPY EXIT=0
cargo test -p trusty-mpm                                → TEST EXIT=0
  test result: ok. 4294 passed; 0 failed; 7 ignored (lib)
  test result: ok. 1392 passed; 0 failed; 1 ignored (bin/tm)
  45 test binaries, all "test result: ok"
cargo test -p trusty-mpm -- --include-ignored           → TEST EXIT=101
  test result: ok.     4301 passed; 0 failed (lib)
  test result: FAILED. 1392 passed; 1 failed (bin/tm)
bash scripts/check_line_cap.sh           → 0 violations
bash scripts/check_test_pointers.sh      → 20469 citations, 0 dangling
bash scripts/check_changelog_fragment.sh → OK trusty-mpm
```

`check_test_pointers.sh` caught a stale test name in one of my new doc comments; fixed before commit.

The single `--include-ignored` failure is unchanged and not mine:

```
---- tests_behavior_a::install_then_deploy_composes_agents stdout ----
  left: 42
 right: 6
```

Reproduced on clean `origin/main` at `a28698c8` — [#4724](https://github.com/bobmatnyc/trusty-tools/issues/4724).

Not merging. CI watch below.

bobmatnyc:
## Follow-up: `c0c57fa2` — CI caught a real defect in my own test design

The first CI run after `6cc69822` went red. Two separate things were in it; I want to be precise about which was mine.

**Mine.** The three `session_ls_*` tests drove `session_ls_at`, which called the **production** `live_tmux_session_names()`. On my machine tmux exists, enumeration succeeds, tests pass. On a runner with no tmux it returns `None`, and the fail-closed rule I had just added correctly pruned nothing:

```
precondition: this invocation must genuinely prune a record — otherwise 'exactly one GET' is vacuous
  left: 0
 right: 1
```

I had noticed this risk while wiring the live-set through and did not act on it. CI did. Both environmental inputs are now injected together as a `PruneContext` (marker path + live-tmux set), so no test can reach the host again. Verified by simulating the CI environment — forcing `live_tmux_session_names()` to return `None` unconditionally still leaves **all 29 prune tests green**, which is the property that was missing. Bundling the two also kept `session_ls_at` at its existing parameter count.

**Not mine.** The same run also had 7 `trusty-common` embedder failures, which is really **one** failure plus six collateral:

```
default_model_matches_sentence_transformers_reference panicked at reference_accuracy_tests.rs:143:43:
called `Result::unwrap()` on an `Err` value: failed to initialise fastembed
(tried AllMiniLML6V2 and AllMiniLML6V2Q)
```

That test holds `ENV_LOCK` across the model load, so the panic **poisons** it and six unrelated env-var tests die on `PoisonError { .. }`. The crate is already split on the idiom — 27 sites use `.lock().unwrap()`, 25 use the poison-tolerant `.unwrap_or_else(|e| e.into_inner())` — and the six victims are in the first group. Filed as [#4741](https://github.com/bobmatnyc/trusty-tools/issues/4741) with both separable defects. My diff touches zero files under `crates/trusty-common/`, and `main` at `d522139b` passed CI concurrently. It did not recur on the next run.

## Gates — rung 5, full re-run at `c0c57fa2`

```
cargo fmt --check                                       → FMT EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings → CLIPPY EXIT=0
cargo test -p trusty-mpm                                → TEST EXIT=0
  test result: ok. 4294 passed; 0 failed; 7 ignored (lib)
  test result: ok. 1392 passed; 0 failed; 1 ignored (bin/tm)
  45 test binaries, all "test result: ok"
cargo test -p trusty-mpm -- --include-ignored           → TEST EXIT=101
  test result: ok.     4301 passed; 0 failed (lib)
  test result: FAILED. 1392 passed; 1 failed (bin/tm)   ← #4724, pre-existing
bash scripts/check_line_cap.sh           → 0 violations
bash scripts/check_test_pointers.sh      → 20469 citations, 0 dangling
bash scripts/check_changelog_fragment.sh → OK trusty-mpm
```

**CI: all checks green, 0 failing.**

Mutation matrix re-run against the refactored seam — all still caught:

| Mutation | Result |
|---|---|
| M1 restamp inside window | 2 FAILED |
| M2 drop live-tmux guard | 1 FAILED |
| M3 `None` live-set prunes anyway | 1 FAILED |
| M6 `is_dead_record` → always-false | 1 FAILED |

Plus the new one this commit adds: forcing tmux enumeration to fail leaves all 29 green (previously it would have failed 3).

Not merging — over to you.

bobmatnyc:
## Verdict: APPROVE

Round 3, `c0c57fa2`. The separation is real, the side-effect table is complete, the auto-prune path cannot reach the destructive teardown, and none of the new tests are vacuous. Zero CRITICAL, zero HIGH.

Reviewed against the spec only, in `.claude/worktrees/fix-4702` at the PR head — no edits, no commits.

---

### 1. Is the separation real? — Yes.

Traced `decommission_record_only` (`decommission.rs:390–415`) transitively. Its complete call set:

| Call | What it actually does |
|---|---|
| `SessionManager::get` (`manager.rs:325`) | store read + `reload_if_changed` (file read) |
| `check_worktree_owner` (`decommission.rs:429`) | → `known_owner_of` (`worktree_ownership.rs:286`): registry field, else `read_sentinel_owner` — a file **read**; → `resolve_ownerless` (`:230`): a store **read** |
| `store.write().await.upsert` (`store.rs:257`) | → `save()` (`store.rs:227`): `serde_json::to_string_pretty`, `fs::write` to `sessions.json.tmp`, `fs::rename`. Nothing else. |

No hook, no notification, no event emission on the store write. No `Drop` impl on any type in this path — `grep -rn "impl Drop"` across `crates/trusty-mpm/src` returns only test-support guards and `session_guard.rs:105`'s `TmuxSessionGuard`, which is never constructed here. Verified structurally that the function contains no `search_gc`, `Command`, `fs::remove`, or `graceful_` token.

### 2. Is the side-effect table complete? — Yes. Audited independently.

I walked `decommission_with_root` (`:482–700`) and classified every call, not just the ones the table names:

| Line | Call | Verdict |
|---|---|---|
| 488 / 493 | `get`, `check_worktree_owner` | read-only |
| 507 | `disposable_workspace_index_id` | `resolve_project_root` + `derive_index_id` — read-only |
| 524 | `graceful_terminate_runtime` (`restart_ops.rs:139`) | `signal_terminate` + `kill_session` — **table effect 1** ✓ |
| 552 | `inspect_dirt` (`worktree_safety.rs:303`) | read-only git via `git_command` (pinned globals, env-redirects stripped). Correctly omitted. |
| 579 | `remove_session_worktree` (`:129–290`) | `git worktree remove --force`, `fs::remove_dir_all` at **three** fallback sites (179, 208, 220), `git worktree prune`, `git branch -D` — **table effect 2** ✓, and the table names all four |
| 630 | `is_safe_to_remove` | canonicalize — read-only |
| 639 | `fs::remove_dir_all(ws)` | **table effect 3** ✓ |
| 670 | `delete_search_index_best_effort` (`search_gc.rs:98`) | `DELETE {base}/indexes/{id}?delete_data=true` — **table effect 4** ✓ |
| 682–697 | pointer clear, #4400 clear, `Decommissioned`, `upsert` | **table effects 5–7** ✓ |

**No omission found.** The one thing not in the table that touches disk is `git status`'s index-stat refresh inside `inspect_dirt` — benign, correctly left out.

### 3. Can auto-prune reach `decommission_with_root`? — No, by any route.

Single entry point: `managed_routes/mod.rs:879–883`, a plain `if q.record_only { decommission_record_only } else { decommission }`. The client sends `record_only=true` unconditionally (`session_picker_prune.rs:835`), and `auto_prune_always_requests_record_only_never_full_teardown` asserts the literal `record_only=true` substring on every captured request URL — path assertion, not outcome assertion.

The only residual reach is a daemon predating the `Query<DecommissionQuery>` extractor. That extractor shipped in **1.3.1** (CHANGELOG line 661, PR #4384); the response-body check plus the 1-hour `STALE_DAEMON_MARKER_KEY` sentinel bound it to one call per hour. Pre-existing, not introduced here.

### 4. Is the derivation test an adequate substitute for effect 4? — Yes, and not writing the live one was correct.

`disposable_index_id_for_a_removed_worktree_resolves_to_the_parent_project` proves the actual mechanism: for an absent `.worktrees/<leaf>` under a real `.git` root, `disposable_workspace_index_id` returns exactly `derive_index_id(project_root)`. I confirmed `derive_index_id` is a bare `file_name()` (`trusty-common/src/index_id.rs:96`), so the equality is meaningful rather than incidental.

That pins the *hazard*. What pins the *absence* is structural and stronger than a runtime test: `decommission_record_only` contains no `search_gc` reference at all, and `delete_search_index_best_effort` is `pub(super)` with exactly one call site. A test that issued the real DELETE would destroy this project's index on any machine with a live trusty-search daemon — that is the hazard under review, not a way to prove it. Effect 4 is not left unpinned in a way that matters.

### 5. Window reset — the test genuinely fails with the bug.

`age_sightings_by` (`:1803`) is relative, so a restamp survives into the next assertion. Trace with the restamp reintroduced on the `Sighting::TooRecent` arm:

- T+0 → `Absent`, stamp `now`. pruned 0 ✓
- `age_sightings_by(6)` → stamp `now−6m`. Listing → `TooRecent` → **buggy code restamps to `now₂`**. pruned 0 ✓ (both pass here)
- `age_sightings_by(5)` → stamp `now₂−5m` (buggy) vs `now−11m` (fixed)
- Listing → buggy: 5m < 10m → `TooRecent` → pruned **0**, assertion expects **1** → **FAILS**

Deterministic, not probabilistic. `auto_prune_does_not_restamp_a_sighting_still_inside_the_window` independently asserts the persisted stamp is byte-identical to the seeded one. Neither is vacuous.

### 6. tmux enumeration failure — refuses, does not fall through.

`live_tmux_session_names` (`:454`) returns `None` on both spawn failure (`.ok()?`) and non-zero exit. `auto_prune_dead_records_at:548–554` early-returns `kept: sessions, pruned: 0, pending: 0` — no partition, no marker write, no HTTP. Pinned by `auto_prune_prunes_nothing_when_tmux_cannot_be_enumerated`, which seeds a 30-minute-old (fully confirmed) sighting and still asserts `pruned == 0` and zero captured calls. The strongest possible form of that test.

### 7. Is `PruneContext` injection complete? — Yes.

Both environmental inputs are in the context. Grep for `fetch_live_sessions|prune_and_report(|session_ls(` across the test file returns zero test call sites — every test enters through `session_ls_at`, `auto_prune_dead_records_at`, or `hermetic_prune_ctx`. No test can reach `live_tmux_session_names()` or `default_marker_path()`. Nothing left that passes only where tmux exists.

---

### The question you asked me to answer explicitly

**Is "documentation, not enforcement" acceptable for a pristine release? Yes. Merge as is — no lint, type-level split, or trait boundary needed first.**

The property this release depends on is not "the table stays accurate." It is "`decommission_record_only` shares no code path with a destructive effect" — and that property *is* mechanically guaranteed, just expressed as function separation rather than as a type. It is pinned by `decommission_record_only_has_no_side_effects_beyond_the_store`, which drives the real path against a fixture where every effect in the table would be observable (populated owned workspace, live tmux name seeded).

The failure a lint would catch is "someone adds a fifth effect to `decommission_with_root` and forgets the table row." That cannot endanger the record-only path, because the two functions no longer share code — it only makes a doc stale. Stale doc is MEDIUM maintenance, not a release gate.

Both prior BLOCKs had one root cause: a boolean parameter on a multi-effect function. That cause is now structurally eliminated rather than documented around. Adding a type-level split on a release branch, whose only job would be to re-express "these are two functions," costs more risk than it removes. If you want belt-and-braces later, the cheap version is a CI grep asserting `decommission_record_only`'s body contains no `Command`/`fs::remove`/`search_gc`/`graceful_` token — that is a `Promote`, not a merge condition.

---

### Findings

|---|---|---|---|---|---|
| LOW | `crates/trusty-mpm/src/session_manager/tests.rs` | ~1069 (`decommission_full_still_terminates_the_runtime`) | The new test drives the full teardown with `workspace_owned = true` and workspace basename `full`, so it reaches `delete_search_index_best_effort` and issues a real `DELETE /indexes/full?delete_data=true` against any live trusty-search daemon on the machine running the suite. Pre-existing pattern (`manager_decommission_removes_workspace` does the same with `abc-session-id`), but `full` is a more collision-prone name. | Rename the leaf to something collision-proof (e.g. `tm-test-full-teardown`), or inject a no-op index-delete seam. | Promote |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/guided.rs` | 363–375 | Dropping the `interactive` pre-gate means cron/scripted bare `tm` now prunes. Against a **pre-1.3.1** daemon (no `Query` extractor) that widens stale-daemon exposure from interactive-only to every invocation. Bounded to one call/hour by the body-check + sentinel, and the target record's workspace is already absent so there is no filesystem loss — but the parent-project index DELETE is reachable on that path. | None required. Optionally have the CLI read `GET /health`'s version once and skip the sweep below 1.3.1 instead of learning it from a destructive probe. | Promote |

Neither is a merge condition. The gate removal is the fix, and the residual is a pre-existing property of #4384's design.

---

### Verification performed

```
cargo fmt --check                                     → exit 0
cargo test -p trusty-mpm                              → exit 0
  running 4301 tests
  test result: ok. 4294 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out
  running 1393 tests
  test result: ok. 1392 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
  (all remaining targets ok; 0 failed across every target)
gh pr checks 4725 → 25 pass, 2 skipping, 0 fail
```

CI is fully green on `c0c57fa2` — the fastembed/`PoisonError` cluster (#4741) does not reproduce on this run. Both changelog fragments carry exactly one category on line 1 (`Fixed`, `Security`).

---

**APPROVE.** I would run this unattended on a machine with live workstreams, 166 orphaned worktrees, and a live trusty-search daemon. It is also strictly safer than what is installed: per the #4728 fragment, the shipped 1.3.3/1.3.4 record-only path still deletes the parent project's search index, and this PR is what stops that.

Issue #3952, opened 2026-07-25
trusty-agents: llm::http::tests mutates $HOME under #[serial] alone, races every HOME_LOCK-taking test


bobmatnyc:
## Summary

`crates/trusty-agents/src/llm/http/tests.rs` has five tests that sandbox
`$HOME` (via `unsafe { std::env::set_var("HOME", tmp.path()) }`) under
`#[serial_test::serial]` **alone** — they never take
`crate::test_env::HOME_LOCK`:

- `send_raw_completion_resolves_key_from_store_when_env_absent`
- `send_raw_completion_missing_everywhere_errors_with_provider_name`
- `send_raw_completion_empty_endpoint_credential_falls_back_to_store`
- `send_raw_completion_fireworks_missing_key_errors_with_fireworks_name`
- `send_raw_completion_fireworks_resolves_key_from_store_when_env_absent`

This is confirmed as the root cause of issue #3922's
`listeners::store::tests::dedup_seed_loads_recent_ids` CI flake (PR #3943):
looping `dedup_seed_loads_recent_ids` together with `llm::http::tests`
under deliberately maximised parallelism reproduced concrete failures at
2/5 runs, with captured panics matching the CI failure's shape
byte-for-byte.

#3943 closed #3922 by removing `listeners::store`'s OWN tests from the
blast radius entirely (dependency injection: `EventStore::*_at(dir, ..)`
instead of `$HOME` sandboxing) — but `llm::http::tests`'s five tests
themselves are UNCHANGED and remain a landmine for every OTHER test in this
crate that still sandboxes `$HOME` via `HOME_LOCK` (dozens of test modules,
per the #3922/#3943 workspace-wide audit: `listeners::poll`,
`tools::mcp_tools`, `tools::okg::tests`, `agents::tests::loading`,
`agents::prompt_builder::tests`, `agents::registry::tests`,
`skills::global_cache::tests`, `skills::registry::tests(_persistence)`,
`api::server::tests::*`, `slack::tests`, `init::tests::*`,
`ctrl_session.rs`, `tools::mcp_live::spec`, and more).

## Why `#[serial]` alone, not `HOME_LOCK`

`llm::http::tests`'s tests are `#[tokio::test]`s that hold state across
`.await` points. `crate::test_env::HOME_LOCK` is a `std::sync::Mutex`, and
clippy's `await_holding_lock` correctly forbids holding a `MutexGuard`
[...]

Issue #3921, opened 2026-07-25
test(trusty-agents): flaky listeners::store::tests::dedup_seed_loads_recent_ids — HOME mutation race suspected


bobmatnyc:
## Summary

The test `listeners::store::tests::dedup_seed_loads_recent_ids` (in `crates/trusty-agents/src/listeners/store.rs`) failed in the first CI run of PR #3897 (commit 6defb5c9), then passed cleanly on rerun via `gh run rerun --failed`. This is the third distinct flaky test observed in a single CI session, following #3902's `compression_telemetry_tests` in trusty-code.

## Observation

- **CI Run 30165303605, first attempt:** Test failed (exact error message not captured in available logs)
- **CI Run 30165303605, rerun via --failed:** Passed clean (15m39s)
- **Local reproduction:** Attempted multiple runs; could not reproduce (test code exists on origin/main but not in current worktree HEAD; cannot run the test locally)
- **Flake-vs-regression assessment:** Engineer concluded pre-existing flakiness; the test did not touch (PR #3897 only modifies installer code), supporting that assessment

## Root Cause (Suspected, Not Confirmed)

The test uses a global `HOME_LOCK` (a `tokio::sync::Mutex<()>`) to serialize test access to `$HOME` across all listeners tests, protecting mutation of `$HOME` via `set_test_home()`. **The test code itself is correctly written** — every test in the module holds the lock across its `.await` points, exactly as designed.

However, the flake suggests either:
1. **Another test path accessing the same shared `$HOME`/`~/.trusty-agents/events/` WITHOUT holding `HOME_LOCK`** — a race between a properly-locked test and an unsynchronized caller.
2. **A different test module (outside listeners/) that touches `~/.trusty-agents/events/` without serialization** — unlikely but possible given how many modules touch `$HOME` for configuration.
3. **Specific test ordering under parallel execution** — the first test to run under a given test partition wins the lock; subsequent tests in that partition race with it.

**This mirrors #3902's pattern** — a shared environment resource (`TCODE_DATA_DIR` then, `$HOME` now) that multiple tests mutate under an insufficient synchronization boundary. The difference is #3902 had the mutex defined but unsynchronized callers; this case has proper synchronization locally (HOME_LOCK is used by all listeners tests) but may have external callers.

## Impact

- **Test flakiness erodes CI gate trust.** A gate that fails once then passes is a gate whose state cannot be trusted without rerun inspection.
[...]

Issue #1053, opened 2026-06-09
trusty-mpm: M1 POC integration test + CI gate (mockable)


bobmatnyc:
## Scope
End-to-end test running the metaharness with a MOCK agent runner (no network) proving WI-1→WI-4→WI-6 wiring, plus an #[ignore] live-OpenRouter variant; add the mock test to CI.

## Acceptance Criteria
- `cargo test -p trusty-mpm --features meta meta_poc` passes in CI with the mock
- The ignored live test runs locally with a key

## Effort
Medium

## Dependencies
- WI-4
- WI-6

---
Part of trusty-mpm M1 POC (milestone #2). Recommended defaults (adjustable by owner): in-process reuse of trusty-code as the agent runtime (Seam A); deterministic file-write demo task; mock-first so trusty-mpm work can start before trusty-code #1028/#1029/#1030/#1023 land. trusty-mpm has no native agent runtime today (it launches the claude CLI); the POC must run standalone via trusty-code's native OpenRouter loop. Reuses the existing custom-instruction loader (core/instruction_overrides.rs + instruction_pipeline.rs).

🤖👥 Generated with [Claude MPM](https://github.com/bobmatnyc/claude-mpm)

bobmatnyc:
Re-scoped (2026-06-18) for the Claude-Code-instance model: ignore-tagged end-to-end integration test that runs `meta run --demo --project <tempdir>`, asserts exit 0 and that hello_metaharness.txt exists. Requires `claude` on PATH + valid ~/.claude.json OAuth; skips gracefully if absent. The CI-runnable portion covers the artifact-check/poll logic via unit tests.

bobmatnyc:
Triage (oldest-first): closing as completed. `crates/trusty-mpm/tests/meta_demo_e2e.rs` provides the re-scoped ignored live `tm meta run --demo` test, and the `commands::meta::{launch,verify}` coverage supplies the CI-runnable poll/artifact verification requested by the re-scope comment.

PR #4726, opened 2026-08-04
fix(trusty-mpm): trim resident agent skills; stop prompt fragments entering the delegation roster

### LINKED ISSUE #4642: agent `skills:` frontmatter injects full skill bodies into every dispatch
## Defect

A deployed agent definition carrying a \`skills:\` frontmatter list causes the Claude Code harness to render every listed skill's full SKILL.md body into that agent's prompt at dispatch, before the agent calls the Skill tool. For \`research\` this is 12 skills totalling 55,894 bytes (~13,974 tokens est.) paid on every single dispatch.

## Evidence

- \`/Users/masa/.trusty-tools/trusty-mpm/claude-config/agents/research.md:6\` declares exactly:
  \`skills: [tm-capabilities, software-patterns, brainstorming, git-workflow, requesting-code-review, writing-plans, json-data-handling, root-cause-tracing, systematic-debugging, verification-before-completion, internal-comms, test-driven-development]\`

**Per-file bytes and token estimates:**

| Skill | Bytes | Tokens (÷4) |
|-------|-------|-----------|
| tm-capabilities | 3198 | 799 |
| software-patterns | 6186 | 1546 |
| brainstorming | 2817 | 704 |
| git-workflow | 7418 | 1854 |
| requesting-code-review | 2821 | 705 |
| writing-plans | 3106 | 776 |
| json-data-handling | 4527 | 1131 |
| root-cause-tracing | 5161 | 1290 |
| systematic-debugging | 4642 | 1160 |
| verification-before-completion | 3404 | 851 |
| internal-comms | 3300 | 825 |
| test-driven-development | 9314 | 2328 |
| **TOTAL** | **55,894** | **13,974** |

- Observed directly from inside a dispatched \`research\` agent: all 12 bodies rendered in-prompt via the harness's native skill tags before the agent invoked any skill. Observed TWICE in one agent — on initial dispatch and again unprompted on a later turn after a background-task notification.
- Not produced by this repo's code: \`grep -rl "command-message|skill-format" crates/\` returns zero hits. \`crates/trusty-mpm/src/core/agent_skill_codeploy.rs\` deploys skills to DISK only; it does not inject prompt text. The injection is Claude Code harness behavior keyed off the \`skills:\` field.

**Open question (state plainly, do NOT assert either way):** Whether the re-injection is genuinely per-turn or an artifact of the background-notification resume path. Note that a per-turn cost would compound across multi-turn agents.

## Resolution direction (state as options, do not pick)

1. Trim the \`skills:\` lists to what each agent actually needs at dispatch
2. Drop \`skills:\` entirely and rely on on-demand Skill-tool invocation
3. Keep it only for agents where the skills are genuinely load-bearing at turn one

### LINKED ISSUE #4643: source/deployed drift in `research` agent skills list
## Defect

Framework source \`~/.trusty-mpm/framework/agents/research.md\` declares 11 skills; the deployed copy at \`~/.trusty-tools/trusty-mpm/claude-config/agents/research.md\` declares 12 — the deployed copy additionally lists \`tm-capabilities\`. Source and deployed disagree.

## Evidence

**Framework source** (~/.trusty-mpm/framework/agents/research.md):
```
skills: [software-patterns, brainstorming, git-workflow, requesting-code-review, writing-plans, json-data-handling, root-cause-tracing, systematic-debugging, verification-before-completion, internal-comms, test-driven-development]
```
(11 skills, no tm-capabilities)

**Deployed copy** (~/.trusty-tools/trusty-mpm/claude-config/agents/research.md):
```
skills: [tm-capabilities, software-patterns, brainstorming, git-workflow, requesting-code-review, writing-plans, json-data-handling, root-cause-tracing, systematic-debugging, verification-before-completion, internal-comms, test-driven-development]
```
(12 skills, includes tm-capabilities as first item)

## Resolution

Determine which is canonical and make them agree; if deployment legitimately adds skills, that transform should be explicit and documented rather than a silent divergence.

### LINKED ISSUE #4711: Foundation-file filter misses "base.md" (no hyphen), admitting a prompt fragment into the agent roster
\`is_foundation_file()\` excludes agent files whose stem starts with \`base-\` (hyphen). \`~/.claude/agents/base.md\` has stem \`base\` with no hyphen, so it is not excluded and enters the delegation roster as a delegatable agent named \`base\` (no \`name:\` frontmatter, so the stem is used).

Its content is "Base QA Instructions… Appended to all QA agents" — a prompt fragment, not an agent. It is one of the 43 entries the roster currently reports.

Found during the 1.3.4 validation sweep. Severity LOW: the file is operator-owned debris in the machine-global \`~/.claude/agents\` tier, not something trusty-mpm ships, so no bundled-asset change fixes it. But the filter is too narrow regardless — a bare \`base.md\` is exactly the shape a foundation file takes.

Suggested fix: match \`base\` and \`base-*\` (and consider requiring \`name:\` frontmatter before admitting a file to the roster, so content-less fragments cannot become delegatable agents).

### LINKED ISSUE #4724: install_then_deploy_composes_agents asserts a stale bundled-agent count (6 vs 42)

bobmatnyc:
Four findings that all land on the bundled agent assets and the delegation roster, so they share one worktree and one PR. #4724 was pre-existing but its assertion is on this PR's surface — it counts bundled agents.

Closes #4642
Closes #4643
Closes #4711
Closes #4724

---

## #4642 — `skills:` frontmatter is a resident per-dispatch cost

The Claude Code harness renders each skill named in an agent's `skills:` frontmatter as a **full SKILL.md body** into that agent's system prompt at dispatch, before the Skill tool is ever called. `skills:` is a token bill, not a capability list — a skill left out is unchanged on disk and still invokable on demand.

Every bundled agent was carrying the same ten boilerplate skills (`brainstorming`, `git-workflow`, `requesting-code-review`, `writing-plans`, `json-data-handling`, `root-cause-tracing`, `verification-before-completion`, `internal-comms`, …). Several of those restate what the composed `BASE-AGENT` body already says (Git Workflow, Verification Before Completion); `internal-comms` is a 3P exec-update format no delegated specialist uses; `root-cause-tracing` is documented as a technique *within* `systematic-debugging`.

**Policy applied — resident iff the skill changes how the agent approaches every task in its domain:**

| family | resident skills |
|---|---|
| engineers (via `BASE-ENGINEER` + leaf) | `documentation-style`, `systematic-debugging`, `test-driven-development` |
| rust family | the above **+ `rust-build-performance`** (standing 2026-07-17 directive, unchanged) |
| QA | `systematic-debugging`, `test-driven-development`, `testing-anti-patterns` |
| `code-critic` | `code-review-standards`, `contract-driven-testing` (the two its own body says to load on turn one) |
| ops | `systematic-debugging` |
| `version-control` | `git-workflow` |
| `security` | `security-scanning` |
| `mpm-agent-manager`, `mpm-skills-manager` | `tm-capabilities` |
| `research`, `code-analyzer`, `memory-manager`, `ticketing` | none |

### Measurement (bytes of SKILL.md body resolved through the composed `extends:` chain)

| agent | skills before → after | bytes before → after | ~tokens saved |
|---|---|---|---:|
| `python-engineer` | 18 → 3 | 120,036 → 22,303 | **24,433** |
| `typescript-engineer` | 17 → 3 | 117,086 → 22,303 | **23,695** |
| `api-qa` | 16 → 3 | 104,202 → 21,170 | **20,758** |
| `golang-engineer` | 16 → 3 | 94,081 → 22,303 | **17,944** |
| `phoenix-engineer` | 16 → 3 | 94,081 → 22,303 | **17,944** |
| `mpm-agent-manager` | 12 → 1 | 72,713 → 3,198 | **17,378** |
| `rust-engineer` | 16 → 4 | 101,810 → 32,849 | **17,240** |
| `web-qa` | 16 → 3 | 89,982 → 21,170 | **17,203** |
| `ops` | 14 → 1 | 73,105 → 4,642 | **17,115** |
| `java-engineer` | 15 → 3 | 87,697 → 22,303 | **16,348** |
| `javascript-engineer` | 15 → 3 | 87,697 → 22,303 | **16,348** |
| `security` | 13 → 1 | 66,721 → 1,656 | **16,266** |
| `qa` | 15 → 3 | 84,343 → 21,170 | **15,793** |
| `gcp-ops` | 12 → 1 | 65,065 → 4,642 | **15,105** |
| `local-ops` | 12 → 1 | 65,065 → 4,642 | **15,105** |
| `vercel-ops` | 12 → 1 | 65,065 → 4,642 | **15,105** |
| `code-analyzer` | 12 → 0 | 55,894 → 0 | **13,973** |
| `research` | 12 → 0 | 55,894 → 0 | **13,973** |
| `data-engineer` | 15 → 3 | 73,575 → 22,303 | **12,818** |
| `documentation` | 12 → 1 | 59,165 → 8,347 | **12,704** |
| `nextjs-engineer` | 15 → 3 | 72,863 → 22,303 | **12,640** |
| `react-engineer` | 15 → 3 | 72,863 → 22,303 | **12,640** |
| `svelte-engineer` | 15 → 3 | 72,863 → 22,303 | **12,640** |
| `prompt-engineer` | 12 → 1 | 58,055 → 8,347 | **12,427** |
| `memory-manager` | 11 → 0 | 49,708 → 0 | **12,427** |
| `ticketing` | 11 → 0 | 49,708 → 0 | **12,427** |
| `mpm-skills-manager` | 11 → 1 | 49,708 → 3,198 | **11,627** |
| `php-engineer` | 14 → 3 | 65,897 → 22,303 | **10,898** |
| `ruby-engineer` | 14 → 3 | 65,897 → 22,303 | **10,898** |
| `version-control` | 11 → 1 | 49,708 → 7,418 | **10,572** |
| `refactoring-engineer` | 13 → 3 | 64,241 → 22,303 | **10,484** |
| `tauri-engineer` | 13 → 4 | 68,601 → 32,849 | **8,938** |
| `dart-engineer` | 12 → 3 | 58,055 → 22,303 | **8,938** |
| `engineer` | 12 → 3 | 58,055 → 22,303 | **8,938** |
| `code-critic` | 7 → 2 | 48,359 → 15,519 | **8,210** |
| `web-ui-engineer` | 8 → 3 | 53,360 → 22,303 | **7,764** |
| `dotnet-engineer` | 2 → 1 | 11,545 → 8,347 | **799** |
| **TOTAL (37 agents)** | | **2,602,763 → 560,654** | **510,527** |

`qa` and `research`, the two named in the issue, land at **21,170 (~5,292 tokens)** and **0**. Total roster-wide resident cost falls **78%**, ~510K tokens per full-roster dispatch sweep.

`dotnet-engineer` declared no `skills:` of its own before or after — the only change it sees is `tm-capabilities` leaving `BASE-AGENT`.

**Not de-deployed.** Under the default manifest `skill_selected()` returns true for every stem, so all bundled skills still deploy to disk regardless of any `skills:` declaration; the co-deploy union only ever *adds* to a filtered manifest's selection. Trimming reduces what is rendered into the prompt, not what exists.

## #4643 — source/deployed drift in `research`

Not drift in the strict sense: `skills:` **unions base-first across the `extends:` chain**, so `research.md`'s 11 became 12 in the deployed copy with `tm-capabilities` inherited from `BASE-AGENT`. Real defect, though — the transform was invisible from the source file and undocumented in DOC-42, and it meant one line on a foundation template silently billed all 37 descendants.

Both halves are now closed:
- `tm-capabilities` is off `BASE-AGENT` (it is a catalog index — the definition of an on-demand lookup), so `research`'s deployed list equals its source declaration exactly.
- DOC-42 §SPEC-AGENTSKILLS-01 now documents the union and the resident-cost consequence, and a gate asserts no foundation template except `BASE-ENGINEER` contributes skills.

`BASE-ENGINEER`'s `documentation-style` stays — it shapes every doc comment an engineer writes in this repo, and it keeps the `agent_skills_prose_hints` doctor check quiet about the pointer in `BASE-ENGINEER`'s own body.

## #4711 — foundation filter missed `base.md`

Two changes, because the file-name rule alone only catches fragments that happen to be named like one:

1. `is_foundation_file` matches the **whole stem** `base` as well as `base-*`. Matching the whole stem (not a `base` prefix) is what keeps #4589's near misses — `baseline-analyzer`, `base64-decoder`, `basecamp-sync`, `based-reviewer` — in the roster.
2. **A file with no `name:` frontmatter is excluded outright.** This is the durable half. Claude Code dispatches a subagent by its frontmatter `name:`, so the old file-stem fallback was minting delegation targets the harness could never resolve. The live `~/.claude/agents/base.md` carries `agent_id: base` and no `name:`, so both rules independently exclude it.

Both exclusions log (at `debug`, like the existing foundation-file exclusion) — #4589's lesson was that a silent deletion is undiagnosable.

`crates/trusty-mpm/src/core/delegation_authority.rs` only. No deployed copy under `~/.claude/`, `$CLAUDE_CONFIG_DIR/`, or any project `.claude/` was touched.

## #4724 — `install_then_deploy_composes_agents` asserted a stale agent count

Pre-existing on `origin/main` @ `a28698c8`, folded in because the assertion counts bundled agents and this PR changes them.

**Diagnosed before renumbering.** `assert_eq!(result.deployed.len(), 6)` prints the actual as `left` and the literal as `right`, so the reported `left: 42, right: 6` means the deploy produced **42** and the expectation was **6** — the reverse of a fixture seeing a fraction of the set. 42 is exactly the number of `agents/*.md` artifacts in `bundle::ALL` (37 roster agents + 5 `BASE-*` templates), so **nothing is missing from the deploy path**; the literal was accurate when six agents shipped and never tracked the roster after. It survived the `dda28412` monolith split unexamined.

Fixed by deriving the expectation from the bundle manifest `install_to` just wrote, and comparing **sets, not lengths** — an equal count with different membership would mean the deployer silently dropped a bundled agent and picked up something else, which a length assertion cannot see. A `> 20` scan floor guards the filter itself, so a broken `strip_prefix` cannot make the gate vacuous.

Test-only change (`crates/trusty-mpm/src/bin/tm/tests_behavior_a.rs`), no user-visible behavior, so no changelog fragment per the documented test-only exemption.

```
BEFORE (origin/main assertion, unchanged code):
    left: 42
   right: 6

AFTER:
  cargo test -p trusty-mpm --bin tm -- \
    tests_behavior_a::install_then_deploy_composes_agents --include-ignored --exact
  running 1 test
  test tests_behavior_a::install_then_deploy_composes_agents ... ok
  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1375 filtered out
```

---

## Tests

Every new gate was proven to fail on `origin/main` before it passed here.

**#4711** — `cargo test -p trusty-mpm --lib` with `crates/trusty-mpm/src/core/delegation_authority.rs` reverted to `origin/main`:

```
---- scan_handles_no_frontmatter ----
a file with no frontmatter has no `name:`, so it is not dispatchable and must not enter the roster (#4711)
---- file_without_name_frontmatter_is_excluded_from_the_roster ----
assertion `left == right` failed: only the file declaring `name:` is a dispatchable agent (#4711)
  left: 2 / right: 1
---- bare_base_md_is_excluded_from_the_roster ----
assertion `left == right` failed: a bare `base.md` is a foundation file in every spelling (#4711)
  left: ["Base", "engineer"] / right: ["engineer"]

```

**#4642/#4643** — the same two gates with `src/assets/agents/` reverted to `origin/main` (the failure message is itself the before-measurement):

```
---- no_foundation_layer_silently_adds_skills_except_base_engineer ----
BASE-AGENT declares `skills:`, which every descendant then pays for on every dispatch
while its own source file does not show it (#4643)

---- bundled_agent_resident_skill_cost_stays_within_budget ----
resident skill-body cost exceeds the 24000-byte per-dispatch budget (#4642):
["api-qa: 104202 bytes (~26050 tokens)", …, "python-engineer: 120036 bytes (~30009 tokens)",
 "qa: 84343 bytes (~21085 tokens)", "research: 55894 bytes (~13973 tokens)",
 "rust-engineer: 101810 bytes (~25452 tokens)", "typescript-engineer: 117086 bytes (~29271 tokens)", …]
 — 36 of 37 agents over budget

```

New gates: `bundled_agent_declared_skills_all_resolve` (keeps the existing `agent_skills` doctor check green — a trimmed list must never leave a dangling reference), `bundled_agent_resident_skill_cost_stays_within_budget`, `no_foundation_layer_silently_adds_skills_except_base_engineer`, `bare_base_md_is_excluded_from_the_roster`, `base_prefixed_near_misses_survive_the_bare_base_rule`, `file_without_name_frontmatter_is_excluded_from_the_roster`.

Two existing tests were **updated, not deleted** — both pinned exact skill strings this change deliberately trims: `code_critic_declares_batch1_skills` (now the two-skill list) and `rust_build_performance_declared_by_rust_family_agents` (unchanged — the standing directive was honoured by keeping `rust-build-performance` and giving the budget gate a named 34,000-byte exception for those two agents rather than raising the budget for everyone).

## trusty-code asset sync (forced by the parity gate)

`scripts/check_agent_assets.sh` requires trusty-code's embedded agent copies to byte-match their trusty-mpm sources, so this change cannot land without carrying tcode with it — which is correct, since tcode runs the same harness and pays the same resident cost.

- 30 parity files synced.
- The 4 pinned `tools:`-restricted deviations (`qa`, `web-qa`, `code-critic`, `code-analyzer`) were reconciled **by hand**, not overwritten: their read-only `tools:` allowlist and rewritten prose are untouched; only their `skills:` lists follow the trim. Re-pinned with `--update` (no new deviation added, so the `mod.rs` doc block is unchanged — it documents the `tools:` restriction, not skills).

One pre-existing blocker surfaced on the way: `crates/trusty-code/changelog.d/4512-tui-daemon-autospawn.md` stacked `Changed` + `Added` + `Fixed` into one file, which the release assembler rejects. Broken since `0d307e5e`; it only becomes blocking now because this is the first PR to change trusty-code source since, so the gate runs the assembler over the crate. Split into three same-numbered files with different slugs — bullet text is byte-identical.

## Gates — Rust Test Ladder rung 4

Two crates changed, so this is rung 4: rung 3 for trusty-mpm, plus the tcode consumer's own suite.

```
cargo fmt --check                                          exit 0
cargo check -p trusty-mpm                                  Finished `dev` profile
cargo clippy -p trusty-mpm --all-targets -- -D warnings    Finished `dev` profile, 0 warnings
cargo test -p trusty-mpm                                   4296 passed; 0 failed; 7 ignored (lib)
                                                           + 44 further targets, all ok
cargo clippy -p trusty-code --all-targets -- -D warnings   Finished `dev` profile, 0 warnings
cargo test -p trusty-code                                  1600 passed; 0 failed; 4 ignored (lib)
                                                           + 16 further targets, all ok
bash scripts/check_agent_assets.sh    30 parity files match; 4 pinned deviations match upstream — OK
bash scripts/check_line_cap.sh        3603 files, 0 violations — OK
bash scripts/check_sld.sh             54 specs + 3029 code files, 0 errors
bash scripts/check_changelog_fragment.sh   OK trusty-code, OK trusty-mpm
bash scripts/check_test_pointers.sh   20445 Test: citations, 0 dangling — OK
```

**`--include-ignored` (rung-5 surface, for #4724):**

```
test result: ok. 4303 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
test result: ok. 1376 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (bin tm — includes the fixed test)
… 44 further targets, all ok
```

One target in that run cannot pass on any machine without an interactive session: `meta_demo_e2e::meta_run_demo_writes_and_verifies_artifact`, whose own `#[ignore]` reason reads *"live: requires claude CLI (OAuth), tmux, and the installed framework — #1053"*. It launched a real session and returned `"launch_outcome":"timed-out"`. Its file is untouched by this diff, and its log shows it reading the machine's **installed** framework (still the old skill lists), not this worktree. Not run as a gate, and **not** `#[ignore]`-ed or excluded to get there — it was already `#[ignore]`-gated upstream.

No library API or shared-crate signature changed — the coupling here is an asset-parity gate, not a type surface, so `cargo check --workspace` adds nothing this pair of crate gates does not already cover. `trusty-code-gui` / `trusty-mpm-gui` were not run and **not excluded**: the diff touches zero files under either (`git diff --name-only origin/main...HEAD` is confined to `crates/trusty-mpm/`, `crates/trusty-code/`, `scripts/agent-asset-pins.tsv`, and `docs/specs/agent-bundled-skills.md`).

## Flake observed, not caused

`core::bundled_pm_package::tests::resolve_pm_prompt_wrapper_matches_the_source_reporting_form` failed once in a full-suite run on this branch and has passed 4/4 since, including in isolation. It scans the ambient agent tiers **twice** and compares byte-for-byte, while `test_support.rs`, `session_manager/tests.rs`, `connectors/tm_tests.rs` and `instruction_pipeline_tests.rs` all mutate process-global `HOME` / `CLAUDE_CONFIG_DIR` — the same race shape as #3921 / #3952 in trusty-agents. This diff touches neither that test nor any env-var handling; adding three tests to the binary only perturbed the schedule. No canonical issue exists for it yet.


bobmatnyc:
## Verdict: APPROVE

Adversarial review, independent worktree (`.claude/worktrees/critic-4726`), read-only against the PR. No CRITICAL, no HIGH findings at >80% confidence. Four MEDIUM/LOW observations below, none blocking.

Every BLOCK condition in the brief was tested rather than accepted. All four came back negative.

---

### 1. The dominant question — did the trim remove a capability needed at turn one?

**No hard constraint was removed.** I checked every trimmed skill body for gate/safety/convention language (`NEVER`, `MUST NOT`, `FORBIDDEN`, `Iron Law`) and traced each hit to a restatement in the composed layer or the agent's own body:

| Trimmed skill | Constraint content | Where it survives |
|---|---|---|
| `verification-before-completion` | Iron Law, forbidden phrases, evidence-before-claims | `BASE-AGENT.md` §Verification Before Completion — *stricter* than the skill (adds the silent-skip rule and the required completion format) |
| `git-workflow` | none — pure git mechanics (rebase/stash/cherry-pick/conflicts) | `BASE-AGENT.md` §Git Workflow carries the policy. The skill actually teaches `git reset --hard HEAD~1` and `git stash`, both of which this repo's worktree discipline forbids against the main checkout — dropping it removes a conflicting-guidance surface |
| `internal-comms` | none | Verified: 3P exec-update format, asks for a team name, sources from Slack/Drive/Calendar. No delegated specialist uses it, and "ask for the team name" contradicts BASE-AGENT's Self-Action Imperative |
| `root-cause-tracing` | "NEVER FIX JUST WHERE THE ERROR APPEARS" | Claim verified **verbatim** — the skill's own body reads *"This skill is a specialized technique within the systematic-debugging workflow."* `systematic-debugging` (retained everywhere it mattered) has Phase 3 "Trace to Root Cause" and names the skill in Related Skills. `BASE-RESEARCH.md` independently restates it: *"Trace symptoms back to their root cause through the call chain."* |
| `code-production-process` (code-critic) | 4× `MUST NOT` | All PM-scoped (critic dispatch construction, verdict protocol). `code-review-standards` — retained — carries the anchoring rule and the APPROVE/WARN/BLOCK protocol the critic itself must obey |
| `env-manager` (ops family, security) | "Never put secrets in client-exposed variables (`NEXT_PUBLIC_`)", "never commit .env" | `vercel-ops.md` body lines 40–49 carry this verbatim, including the `grep -r "NEXT_PUBLIC_.*SECRET"` audit; `security.md` lines 24–33 carry a full Secret Detection Protocol; `BASE-OPS.md` §Credential Handling covers session-scoped credential use |
| `testing-anti-patterns`/`test-quality-inspector` core rules | mock-testing ban, test-only methods ban | `BASE-QA.md` §Test Quality restates both; `testing-anti-patterns` is additionally **kept** resident on all three QA agents |

**`research` → 0 is the right call.** Its previous 11 were all generic boilerplate; nothing research-shaped was in the list. What remains is `BASE-AGENT` (310 lines) + `BASE-RESEARCH` (investigation discipline, cite file:line, fact-vs-inference, root-cause tracing, do-not-modify scope) + a 100-line agent body with its own methodology. That is more domain shaping than the 11 skills provided.

**No agent body was left pointing at a skill it no longer declares as a directive.** I swept all 42 bundled agents for prose mentions of the 22 trimmed skills. Three hits, all benign: `BASE-QA.md:9` names `verification-before-completion` as an inheritance note, `data-engineer.md:39` mentions the XLSX *file format*, `mpm-skills-manager.md:108` lists `webapp-testing` in a recommendation table. Zero "load skill X" directives dangle. (`agent_skills_prose_hints` is `Ok`-only by design, so it cannot go red from this.)

**"Still invokable" verified at the code level, with one caveat (finding M2).** `session_launch/mod.rs:689` composes the deploy predicate as `plan.skill_selected(name) || co_deploy_skills.contains(name)` — the co-deploy union only ever *adds*. `manifest/apply.rs:177` + `plan_default_uses_bundled_sources` confirm `skill_selected("anything") == true` under the default manifest, so every bundled skill still reaches disk regardless of any `skills:` declaration.

### 2. #4643 — union semantics

Verified in `crates/trusty-agents-common/src/agents/builder.rs:779-783`: `merge_frontmatter` walks the chain base-first and unions `skills:` with dedup, contrasting with `model:`/`tools:` child-wins. The reframe from "drift" to "undocumented union" is correct, and DOC-42's new §Composition paragraph is accurate. `no_foundation_layer_silently_adds_skills_except_base_engineer` locks it, and `every_bundled_non_base_agent_reaches_the_roster` passes, so removing `tm-capabilities` from `BASE-AGENT` stripped nothing an agent relied on inheriting — the two agents that need it (`mpm-agent-manager`, `mpm-skills-manager`) now declare it directly.

### 3. #4711 — does the `name:` rule drop a legitimate agent?

Tested empirically against **every deployed tier on this machine**, not fixtures. I reimplemented the exact parser semantics (BOM strip, first line `---`, `parse_kv_line` grammar, empty-value skip) and ran it over all three tiers plus 30 project `.claude/agents/` directories:

```
/Users/masa/.claude/agents                                    6 files — 0 dropped (base.md excluded, correctly)
/Users/masa/.trusty-tools/trusty-mpm/claude-config/agents     42 files — 0 dropped (5 BASE-* excluded)
30 project .claude/agents dirs                                1 dropped: duetto-agents-skills/.claude/agents/README.md
```

The single exclusion is a `README.md` that the **old** code admitted to the roster as a delegatable agent named "README". The rule fixes a bug rather than causing one. The five hand-written user-tier agents (`copyeditor`, `pangram-editor`, `proofreader`, `writer`, `writing-critic`) all declare `name:` and survive.

`~/.claude/agents/base.md` confirmed as the live case: frontmatter `model: haiku` / `agent_id: base`, no `name:`, body "# Base QA Instructions — Appended to all QA agents." Both rules exclude it independently, as claimed.

### 4. #4724 — is the derived expectation vacuous?

No. The two sides come from genuinely independent paths: `expected_agents` from the compiled `bundle::ALL` manifest, `deployed_agents` from `deploy_agents()` reading the filesystem that `install_to` wrote. A dropped-and-substituted member is caught by the set comparison, and the `> 20` floor guards a broken `strip_prefix`. Test passes with `--include-ignored`.

### 5. New gates — fail-before proven independently

I reverted each surface to `origin/main` inside my own worktree and re-ran, then restored (tree clean, verified):

```
$ git checkout origin/main -- crates/trusty-mpm/src/assets/agents/
no_foundation_layer_silently_adds_skills_except_base_engineer ... FAILED
  BASE-AGENT declares `skills:`, which every descendant then pays for ... (#4643)
bundled_agent_resident_skill_cost_stays_within_budget ... FAILED
  ["api-qa: 104202 bytes", ..., "python-engineer: 120036", "qa: 84343",
   "research: 55894", "typescript-engineer: 117086"] — 36 of 37 over budget

$ git checkout origin/main -- crates/trusty-mpm/src/core/delegation_authority.rs
bare_base_md_is_excluded_from_the_roster ... FAILED  (left: ["Base","engineer"] / right: ["engineer"])
file_without_name_frontmatter_is_excluded_from_the_roster ... FAILED  (left: 2 / right: 1)
scan_handles_no_frontmatter ... FAILED
```

The reported byte figures match the PR body exactly (`research: 55894`, `qa: 84343`, `python-engineer: 120036`), so the measurement table is credible. On the branch: `32 passed; 0 failed`. Budget headroom is tight by design — 1,697 bytes for the engineer family, 1,151 for the rust family — so the gate is not decorative.

### 6. Standing-directive exception scoping

`RUST_FAMILY: &[&str] = &["rust-engineer", "tauri-engineer"]` with an exact `contains(&stem.as_str())` match. No prefix, no glob, no leak. `rust_build_performance_declared_by_rust_family_agents` is unchanged from before this PR and passes — the 2026-07-17 directive is honoured, not reversed.

### 7. Scope widenings

**trusty-code parity (30 files):** every one is a 1–2 line `skills:` diff. The 4 pinned deviations were **not flattened** — `qa`, `web-qa`, `code-critic`, `code-analyzer` each keep `tools: [read_file, grep, glob, list_dir, search_code, use_skill, finish_task]` verbatim and their rewritten prose untouched; only the `skills:` line moved. Gate confirms: `agent-assets: scanned 30 byte-parity file(s) (floor 20), all match; 4 pinned deviation(s) match upstream — OK.`

**Changelog fragment split:** byte-identical. `diff` of the concatenated three new files against `origin/main`'s original shows exactly two removals, both blank separator lines between the former inline sections. Zero words changed.

### 8. Repo gates — reproduced independently

```
test-pointers: resolved 20445 Test: citation(s) — 0 dangling — OK.
sld-lint:      54 spec doc(s) + 3029 code file(s); 0 error(s), 0 warning(s)
line-cap:      3603 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
changelog:     OK trusty-code, OK trusty-mpm; all 2 crate(s) with source changes recorded
```

### 9. Pre-existing items confirmed as stated

- `meta_demo_e2e::meta_run_demo_writes_and_verifies_artifact` — `#[ignore = "live: requires claude CLI (OAuth), tmux, and the installed framework — #1053"]` has been on `origin/main` since `26fbf152` (#1489). This PR touches zero files under `crates/trusty-mpm/tests/`. The author did not add the ignore; the claim holds.
- `resolve_pm_prompt_wrapper_matches_the_source_reporting_form` — **not made more likely.** The three new `delegation_authority` tests operate on `TempDir` and contain no `set_var`; the three new `agent_skill_codeploy` tests read `CARGO_MANIFEST_DIR` paths only. The PR adds no new mutator of process-global `HOME`/`CLAUDE_CONFIG_DIR`. Scheduling perturbation only.

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/agent_skill_codeploy.rs` | 181 | `bundled_agent_stems()` filters with `starts_with("base")` — the exact bare-prefix form this same PR fixes in `is_foundation_file` and that #4589 documents as silently deleting `baseline-analyzer` / `base64-decoder`. A future bundled `baseline-*` agent would be invisible to all three new gates: no dangling check, no budget check, no union check. | Use the whole-stem rule: `.filter(\|stem\| { let l = stem.to_ascii_lowercase(); !(l == "base" \|\| l.starts_with("base-")) })`, or make `is_foundation_file` `pub(crate)` and call it — `delegation_authority_tests.rs:417` already does exactly that. | Parent |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 689 | Under a **filtered** skill manifest (`skill_include`/`skill_exclude` set), trimming `skills:` also shrinks `co_deploy_skills`, so a skill that previously reached disk only via the co-deploy union no longer deploys — and the "still invokable via the Skill tool" fallback then fails, because the file isn't there. Unaffected under the default manifest (verified `skill_selected("anything") == true`), so this is a non-default-configuration consequence, not a shipped regression. | Note it in DOC-42 §Resident cost: the on-demand fallback assumes an unfiltered skill manifest. Optionally have the deploy predicate always include the bundled catalog when the manifest sets no explicit `skill_include`. | Promote |
| LOW | `crates/trusty-mpm/src/assets/agents/{qa,api-qa,web-qa,web-ui-engineer,data-engineer}.md` | frontmatter | Three domain-shaping skills dropped with no restatement anywhere in the composed chain: `condition-based-waiting` (the QA family's anti-flake convention — poll a condition, never sleep a timeout), `webapp-testing` (Playwright selector/wait strategy, for `web-qa` and `web-ui-engineer`), `xlsx` (for `data-engineer`, whose own body at line 39 advertises "CSV ↔ Excel (XLS/XLSX) with formatting preservation"). No safety or gate impact — these are capability-at-turn-one, and each is still Skill-tool invokable. Flagged only so the trade is on the record rather than implied. | None required. If `condition-based-waiting` proves load-bearing, its one operative rule is a single line and belongs in `BASE-QA.md` §Testing Discipline, not a 5 KB resident body. | Promote |
| LOW | `crates/trusty-mpm/src/bin/tm/tests_behavior_a.rs` | 249 | The corrected #4724 gate stays `#[ignore]`-gated (pre-existing, "requires bundled framework artifacts"), so it still does not run in default CI. The literal drifted from 6 to 42 undetected precisely because nothing ran it. Correct fix, but the exposure that let it rot is unchanged. | Out of scope here. Worth a separate look at whether `install_to` + a `TempDir` can be made CI-runnable. | Promote |
| LOW | `crates/trusty-mpm/Cargo.toml` | 3 | #4725, #4726 and #4727 each carry an identical `1.3.4 → 1.3.5` bump. Harmless (git 3-way resolves identical edits on both sides without conflict), but redundant. Separately: the fragments are correctly left **pending**, which means the releaser must run `assemble-changelog.sh crates/trusty-mpm 1.3.5` directly — `bump-version.sh <crate> patch` against an already-bumped 1.3.5 would produce 1.3.6 and leave 1.3.5 as a version with no CHANGELOG section. | Release-time note for the PM, no code change. | Parent |

---

## Notes

Would I ship this trim to every agent on a public release? Yes. The strongest evidence is negative and was hard to get: I could not find a single rule an agent must obey that exists only in a trimmed skill. The three claims the PR leans on — that `git-workflow` and `verification-before-completion` restate `BASE-AGENT`, that `internal-comms` is an exec format no specialist uses, and that `root-cause-tracing` is a technique within `systematic-debugging` — all check out against the actual bodies, and the last one is stated verbatim in the skill's own text.

What the trim is really doing is removing tutorial-grade reference material from the resident prompt while the constraint layer (`BASE-AGENT` 310 lines, `BASE-ENGINEER` 207, plus each agent's own body) stays intact. In two cases it removes *conflicting* guidance: `git-workflow` teaches `git reset --hard` and `git stash`, which this repo's worktree discipline forbids against the main checkout, and `internal-comms` instructs the agent to stop and ask for a team name, which contradicts the Self-Action Imperative.

`documentation-style` correctly staying on `BASE-ENGINEER` is load-bearing and was the right call to protect — it encodes the Why/What/Test convention that `check_sld.sh` and `check_test_pointers.sh` mechanically enforce, so trimming it would have traded 8 KB of prompt for a class of CI failures.

Not flagged, deliberately: `tm-capabilities/references/agents.md` renders the *declared* skill list, not the composed one, so `engineer` shows two skills where the deployed copy has three. That is pre-existing generator behavior, not introduced here, and it is arguably the right view for a source catalog.

Verified independently in `.claude/worktrees/critic-4726`; the two revert experiments were restored and the tree confirmed clean afterward. Read-only with respect to the PR — no edits, no commits, not merged.

PR #4727, opened 2026-08-04
fix(trusty-mpm): quarantine untracked project-tier agents that shadow bundled stems (#4448)

### LINKED ISSUE #4448: deployer: quarantine untracked project-tier files that shadow bundled agent stems

bobmatnyc:
## What

`retract_framework_agents` ([#4409](https://github.com/bobmatnyc/trusty-tools/issues/4409)) can only delete what its ownership ledger names. A copy an older binary wrote before that ledger existed survives it, outranks the canonical user tier in Claude Code's resolution, and is never refreshed again — [#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408) made permanent. This adds the half that reaches those, reversibly.

Closes [#4448](https://github.com/bobmatnyc/trusty-tools/issues/4448).

Not built on [#4526](https://github.com/bobmatnyc/trusty-tools/pull/4526) — that PR was closed unmerged, and this is a fresh implementation that closes each of its recorded defects.

## The movability predicate — four gates, each fail-closed

| # | Gate | Refuses when | Source |
|---|---|---|---|
| 1 | Name | the file does not resolve to a name tm ships | `tier_audit::classify_tier_resident` (#4442) |
| 2 | Ledger | the ledger records the operator as owner | `tier_audit::ownership_of` (#4442) |
| 3 | VCS | git tracks the file, or git cannot be asked | `agents::vcs_claim` (**new**) |
| 4 | Schema | the frontmatter is not tm's composer output | `agents::agent_schema` (**new**) |

Gates 3 and 4 are what #4526 lacked, and each closes one of the defects that sank it.

### Gate 3 — the absorbed `Origin::Project` requirement

`ShadowsBundled` + `Untracked` cannot separate a stale leftover from a project's deliberate override. `Origin::Project` was going to supply the missing declaration; [#4443](https://github.com/bobmatnyc/trusty-tools/issues/4443) was closed as superseded, so the declaration had to come from somewhere the deployer can actually reach.

It already does: **a repository that COMMITS a project-tier agent is declaring it.** `git ls-files` is that declaration — no new schema, ledger, or config surface, and one process spawn per sweep rather than per file. Three states, never a bool: "no repository here" (nothing can be claiming it → sweepable) and "git could not be asked" (a claim cannot be ruled out → refuse) have opposite safe answers, and collapsing them either freezes the sweep in every non-git project or lets it run blind when git is broken.

#4526 would have renamed this repository's own committed agent files. It cannot now.

### Gate 4 — classify by SCHEMA, never by filename

claude-mpm is a separate, still-live project that deploys into the same `.claude/agents` convention reusing tm's exact filenames (`engineer.md`, `ops.md`, `qa.md`, `research.md`, `ticketing.md`, `memory-manager.md`, …). A filename-keyed sweep moves another project's files.

Classification is a **positive identification** on the frontmatter key set: trusty-mpm only when every declared key is one `merge_frontmatter` emits AND both `name:` and `role:` are present.

One trap worth naming: **`initialPrompt:` is not a usable discriminator.** It reads like a claude-mpm marker and is not one — `merge_frontmatter` (`builder.rs:811-817`, HR-1 Part B) injects a role-derived prompt when the source declares none, so trusty-mpm emits it for most agents. Treating it as foreign would have exempted nearly the whole bundled roster from the sweep.

Size is not a signal either, though it looks like one. The 2026-07-31 inventory found middle-generation claude-mpm files at 23–81 KB, overlapping and exceeding the bundled 17–27 KB band. Nothing here measures length.

`pinned_to_the_composer_emission` composes a real agent and asserts its emitted keys stay inside the whitelist, so the gate cannot silently rot the next time the composer gains a key.

## Never coupled to deletion

There is no `remove_file` and no `remove_dir` on any path, success or failure. Quarantine is copy → verify → rename; a failure leaves both the original and its backup in place.

`never_deletes_on_any_path` asserts this against the module **source text**, because "there is no such call" is a property no behavioural test can establish — a behavioural test only proves the paths it happened to walk. If a future change genuinely needs a removal, that test failing is the review gate that forces the conversation.

`a_sweep_destroys_no_content` is the behavioural other half: after a real sweep, every byte staged in the tier is still on disk somewhere.

## Partial-failure receipts — #4526's CRITICAL gap

Every candidate the sweep examines lands in exactly one of `QuarantineReport`'s three lists (`moved` / `skipped` / `failed`), and the report is returned on every exit path. A per-file I/O failure is recorded and the sweep continues; only whole-sweep refusals (corrupt ledger, empty roster) return `Err`, and each means nothing was touched.

The receipt file is a projection of that same struct, so the file on disk and the value the caller logs cannot disagree. `a_partial_failure_still_reports_every_candidate` stages one file that moves, one whose rename provably cannot land, and one that is skipped, then asserts all three appear — in the return value and in the receipt on disk.

The receipt's restore command POSIX-single-quotes every path, carrying forward #4526's injection fix; `restore_command_survives_a_hostile_filename` stages `it's $(touch PWNED-SUB) \`touch PWNED-TICK\`.md`, **executes the printed `mv` through `sh`**, and asserts the file is back and no payload ran.

## Verification re-reads from disk

An exit code is not evidence. The backup is re-read and byte-compared before the original is touched; after the rename, both the destination's existence and the source's absence are confirmed from the filesystem.

## Call sites

Wired after retraction at both live paths — `prepare_session_inner` and `sync_session_assets` — through ONE helper, so the two cannot aim differently. The target is bound to `project_dir`'s own `.claude/agents`, spelled out and never `fw.claude_agents_dir()`, which is the operator's real `~/.claude/agents` on the non-git `tm session start` and TUI `/connect` paths. It is strictly more dangerous there than retraction: retraction can only delete ledger-tracked files (that tier has none), while this sweep moves *untracked* ones — exactly what that directory is full of.

The bundled-name roster moved from a private fn in `doctor_asset_tier.rs` to `core::bundled_roster`, so `tm doctor`'s report and this repair resolve it from one place. `tier_audit`'s contract doc now states the report/repair relationship precisely: everything the quarantine moves, doctor reports; never the reverse.

## Mutation evidence — 25 run, 25 killed

#4526's review found the primary call site had **zero** coverage: deleting it left the whole `trusty-mpm` suite green. That is specifically re-tested here.

| # | Mutation | Killed by |
|---|---|---|
| M1 | gate 3 (VCS) removed | `quarantine_never_moves_a_git_tracked_file`, `movable_only_when_all_four_gates_agree` |
| M2 | gate 3 `Unknown` fails open | `movable_only_when_all_four_gates_agree` |
| M3 | gate 4 (schema) removed | `quarantine_never_moves_a_claude_mpm_file_on_a_colliding_name` +3 |
| M4 | gate 4 unrecognised passes | `quarantine_never_moves_a_hand_authored_file` |
| M5 | gates 1+2 removed | `quarantine_never_moves_a_custom_agent` +3 |
| M6 | corrupt ledger degrades instead of refusing | `quarantine_refuses_on_a_corrupt_ledger` |
| M7 | empty roster becomes a no-op | `quarantine_refuses_an_empty_roster` |
| M8 | `free_path` follows symlinks | `a_dangling_symlink_is_not_free` |
| M9 | collision overwrites | `a_second_run_does_not_overwrite_the_first_backup` |
| M10 | receipt never written | `a_partial_failure_still_reports_every_candidate` +2 |
| M11 | a failure aborts the sweep | `a_partial_failure_still_reports_every_candidate` |
| M12 | `candidates` follows symlinks | `a_symlink_is_never_a_candidate` |
| M13 | sweep runs without the ledger lock | `quarantine_blocks_while_the_ledger_lock_is_held` |
| M14 | backup skipped entirely | `a_sweep_destroys_no_content` +3 |
| M15 | `shell_quote` → Rust `Debug` | `restore_command_survives_a_hostile_filename` |
| M16 | schema stops requiring `role:` | `schema_requires_role` |
| M17 | schema accepts unknown keys | `schema_of_a_hand_authored_file` +2 |
| M18 | `agent_type` dropped from the foreign vocabulary | `each_claude_mpm_marker_is_recognised_alone` |
| M19 | unterminated frontmatter accepted | `schema_of_an_unterminated_block_is_unrecognized` |
| M20 | no-repo answered `Unknown` | `quarantine_works_without_a_repository` +1 |
| M21 | a `remove_file` introduced | `never_deletes_on_any_path` |
| M22 | `agent_type` added to tm's vocabulary | `the_two_vocabularies_do_not_overlap` |
| **M23** | **`prepare_session_inner` call site DELETED** | `prepare_session_quarantines_a_shadowing_workspace_agent` — full `trusty-mpm` lib suite: `4297 passed; 1 failed` |
| **M24** | **sweep re-aimed at `fw.claude_agents_dir()`** | `prepare_session_never_quarantines_the_operator_home_agents_tier` +2 |
| **M25** | **`sync_session_assets` call site removed** | `sync_assets_quarantines_a_shadowing_workspace_agent` |

M18 survived a first pass, because the test iterated `CLAUDE_MPM_KEYS` itself — deleting an entry deleted its own assertion. The marker list in the test is now written out literally.

## Gates — Rust Test Ladder **rung 5** (persistence, destructive-adjacent, cross-crate)

```
cargo fmt --check                                                 exit 0
cargo check -p trusty-agents-common                               ok
cargo clippy -p trusty-agents-common --all-targets -- -D warnings  0 warnings
cargo clippy -p trusty-mpm --all-targets -- -D warnings            0 warnings
cargo test -p trusty-agents-common          434 passed; 0 failed; 0 ignored
cargo test -p trusty-agents-common -- --include-ignored
                                            434 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm                   4298 passed; 0 failed; 7 ignored
                                           1375 passed; 0 failed; 1 ignored  (x2 bins)
                                            + every integration suite: ok
cargo test -p trusty-agents (dependent)    3312 passed; 0 failed; 11 ignored
cargo test -p trusty-code   (dependent)    1600 passed; 0 failed; 4 ignored
bash scripts/check_line_cap.sh   3613 tracked .rs files, 0 violations — OK
bash scripts/check_sld.sh        54 spec docs + 3040 code files, 0 errors
bash scripts/check_changelog_fragment.sh   both crates recorded — OK
```

### `cargo check --workspace` — known pre-existing red, not this branch

```
error: proc macro panicked

error: proc macro panicked
  --> crates/trusty-mpm-gui/src/lib.rs:42:14
```

Missing built Tauri UI assets, unrelated to agents. Proof this branch is not the cause:

```
$ git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/
(empty)
```

Neither crate was excluded, `#[ignore]`d, or `cfg`-gated. Every other workspace member checks clean.

## Not done, deliberately

The sweep was never pointed at any real deployed directory on this machine. Every test fixture is a `TempDir`; `prepare_session_never_quarantines_the_operator_home_agents_tier` exists specifically to keep it that way in production too.


bobmatnyc:
## Verdict: BLOCK

Adversarial review, rung 5, run read-only from `.claude/worktrees/critic-4727` (detached at `2796483f`). Findings are ranked; both top findings were **executed**, not reasoned about.

The mutation evidence holds — I re-ran four of the 25 and every one reproduced the claimed kill (table at the bottom). M23 in particular reproduced the claimed `4297 passed; 1 failed` exactly. That part of this PR is the strongest I have seen on this repo.

The block is one gate, not the design.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-agents-common/src/agents/vcs_claim.rs` | 100-106 | Gate 3 reads **any** non-zero `rev-parse --is-inside-work-tree` exit as `NoRepo` → `Unclaimed` → **sweepable**. Git returns 128 for every fatal condition, not only "not a repository". A real work tree that git declines to read therefore reaches the *sweepable* state, not the refusing one — and its committed agent files get moved. | Branch on *why* it failed, not just that it failed: keep `NoRepo` only when stderr carries `not a git repository`, and map every other non-zero exit to `IndexState::Unavailable`. Also require `stdout.trim() == "true"` on success (a bare repo exits 0 printing `false`). | Fix here |
| HIGH | `crates/trusty-agents-common/src/agents/agent_schema.rs` | 150-158 | Gate 4 is not the "positive identification of composer output" the docs claim. It accepts any document whose keys are a subset of a 9-key whitelist and which declares `name:` + `role:`. `role:` is the only real discriminator, and it is not exclusive to tm — a hand-authored agent declaring `name`/`role`/`description`/`model` classifies `TrustyMpm` and is movable, contradicting `agent_schema.rs:71-73` ("`Unrecognized` — everything else, including a hand-authored file") and the changelog. | Require a marker the composer emits and a human would not co-declare — e.g. demand `initialPrompt:` **together with** `name:`+`role:`, or have `merge_frontmatter` stamp an explicit provenance key and gate on that. A whitelist-subset test cannot distinguish "tm wrote this" from "this declares nothing exotic". | Fix here |
| LOW | `crates/trusty-agents-common/src/agents/quarantine.rs` | 214-220 | `RECEIPT.md` is written with `std::fs::write` (truncating) into `backup_root/<run_id>/`, where `run_id` is second-resolution. Backups collision-protect via `free_path`; the receipt does not. Two sweeps that both move something inside the same UTC second silently replace the first receipt. | Route the receipt through `free_path` too, or make `run_id` collision-proof (append a pid/nonce). | Parent |
| LOW | `crates/trusty-agents-common/src/agents/quarantine_receipt.rs` | 224 | `shell_quote` builds from `path.display()`, which is lossy. Same in `quarantine.rs:399` (`free_path` → `to_string_lossy`). Candidate file names are UTF-8-gated so a move can't go astray, but a non-UTF-8 *parent* path yields a printed `mv` naming a path that does not exist. | Quote from `OsStrExt::as_bytes` on unix, or record that the receipt command is UTF-8-only. | Promote |
| LOW | `crates/trusty-agents-common/src/agents/quarantine_tests.rs` | 643-651 | `never_deletes_on_any_path` scans four module sources but not `crates/trusty-mpm/src/core/session_launch/quarantine_shadows.rs`, the module that actually aims the sweep. It has no fs calls today; the pin is what would keep it that way. | Add it to the `include_str!` list (or add a sibling test in `trusty-mpm`). | Parent |

## Required Changes

### 1. CRITICAL — gate 3 fails **open**, and I moved a committed file to prove it

The PR body's own framing is the specification this violates:

> Three states, never a bool: "no repository here" (nothing can be claiming it → sweepable) and "git could not be asked" (a claim cannot be ruled out → refuse) have opposite safe answers.

The implementation has three states but only two *sources*. `Command::spawn` failing yields `Unavailable`. Everything else — every fatal git error inside a genuine work tree — yields `NoRepo`:

```rust
Ok(out) if !out.status.success() => {
    return Self { state: IndexState::NoRepo };
}
```

Git has no dedicated exit code for "this is not a repository". It exits 128 for that *and* for every other fatal, so this branch swallows the exact class the design says must refuse.

**Executed proof.** Temporary test appended to `quarantine_tests.rs` (reverted afterwards; working tree is clean):

```rust
let f = Fixture::new();                       // real `git init` repo
let original = f.write("qa.md", &content);
f.git_add(".claude/agents/qa.md");            // TRACKED and staged

let healthy = f.sweep(&roster(&["qa"])).unwrap();
assert_eq!(healthy.moved.len(), 0);           // passes — healthy repo refuses

// make the SAME repo unreadable without removing it
// (.git/config: repositoryformatversion = 0 -> 99)

let report = f.sweep(&roster(&["qa"])).unwrap();
assert_eq!(report.moved.len(), 1);            // PASSES
assert!(!original.exists());                  // the COMMITTED file was moved
```

```
running 1 test
test agents::quarantine::tests::critic_probe_unreadable_repo_moves_a_tracked_file ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 434 filtered out
```

`repositoryformatversion` is the deterministic trigger, not the realistic one. The realistic ones all land on the same branch:

- **`detected dubious ownership`** (exit 128) — a checkout owned by another uid. This repo's own operating notes record root-owned worktrees under `.base`; it is also the default state for Docker-mounted and shared-volume checkouts.
- **A stale worktree pointer.** Verified separately here — a `.git` file whose gitdir is gone gives `fatal: not a git repository: (null)`, exit 128. The 2026-07-21 `.base` incident orphaned ~70 worktrees into exactly this state, and every one of them would have been swept as "no repository here".
- **Any `git` shim on `PATH` that exits non-zero.**

In each case the repository exists, the agent files are committed, and the sweep moves them. That is the one outcome `movable_only_when_all_four_gates_agree` and the whole gate-3 rationale exist to make impossible.

Suggested shape:

```rust
Ok(out) if !out.status.success() => {
    let why = String::from_utf8_lossy(&out.stderr);
    let state = if why.contains("not a git repository") {
        IndexState::NoRepo          // genuinely unclaimed
    } else {
        IndexState::Unavailable     // could not be asked -> refuse
    };
    return Self { state };
}
Ok(out) if out.stdout.trim_ascii() != b"true" => {
    return Self { state: IndexState::Unavailable };   // bare repo prints "false"
}
```

and a regression test that stages a **tracked** file in a repository git cannot read and asserts `report.moved.is_empty()`.

### 2. HIGH — gate 4 accepts hand-authored files it documents as refusing

`agent_schema` returns `TrustyMpm` on `keys ⊆ TRUSTY_MPM_KEYS && has(name) && has(role)`. The whitelist is `name, role, description, model, max_tokens, resource_tier, skills, tools, initialprompt` — which is close to the full set of fields anyone writing an agent by hand would use. The only key doing discriminating work is `role:`, and nothing makes `role:` tm's.

**Executed proof**, using a file that exists on this machine — `~/.claude/agents/writing-critic.md`, hand-written by the operator, frontmatter `name: / role: qa / description: / model: sonnet`:

```rust
let doc = "---\nname: code-critic\nrole: qa\n\
           description: Adversarial outside review for Bob Matsuoka's articles.\n\
           model: sonnet\n---\n\n# Writing Critic\n\nHand-written by the operator.\n";
assert_eq!(agent_schema(doc), AgentSchema::TrustyMpm);   // PASSES
```

The real file escapes only by accident: its `description:` is a folded `>` scalar whose continuation contains `Triggers: "outside review"`, and `parse_kv_line` trims and reads that as a spurious key, pushing it out of the whitelist. Unwrap the description onto one line and the classification flips to `TrustyMpm`. That is luck, not a gate.

So a hand-authored project agent on a bundled name (`code-critic`, `qa`, `research`, `engineer`) is movable whenever gate 3 does not save it — and gate 3 does not save it when `.claude/agents/` is gitignored-and-untracked, which is this repository's own default (`.gitignore:36`, `**/.claude/agents/`).

Reversible with a backup and a receipt, so not CRITICAL. But `schema_of_a_hand_authored_file` passes only because its fixture declares `color: purple`; drop that one key and the test inverts.

## Verified — the claims that hold

**Mutations re-run (4 of 25), all reproduced:**

| # | Mutation | Result |
|---|---|---|
| M23 | `prepare_session_inner` call site deleted | `4297 passed; 1 failed` — `prepare_session_quarantines_a_shadowing_workspace_agent` panicked at `tests_quarantine_4448.rs:95`. Exactly as claimed. |
| M24 | sweep re-aimed at `fw.claude_agents_dir()` | 2 failed (`..._quarantines_a_shadowing_workspace_agent`, `..._never_quarantines_the_operator_home_agents_tier`) |
| M2 | gate 3 `Unknown` fails open | `movable_only_when_all_four_gates_agree` FAILED |
| M4 | gate 4 `Unrecognized` passes | `movable_only_when_all_four_gates_agree` + `quarantine_never_moves_a_hand_authored_file` FAILED |

**M18's self-referential shape** — swept the PR's own test files. One loop over a production constant remains (`the_two_vocabularies_do_not_overlap`, `agent_schema_tests.rs:78`), and it is correct there: it asserts a disjointness invariant, and `each_claude_mpm_marker_is_recognised_alone` carries the literal list. No second instance in this diff. (The shape does exist elsewhere in `trusty-agents` — unchanged code, not flagged.)

**Never deletes.** Confirmed by reading, not just by the source-text pin. No `remove_file`, `remove_dir`, `remove_dir_all`, `set_len`, or `File::create` on any path. `fs::write` appears once, for `RECEIPT.md` into the backup dir. Ordering is copy → re-read both → byte-compare → rename → re-stat both, and every failure branch returns leaving the original in place. `free_path` uses `symlink_metadata`, so a dangling symlink counts as occupied and the rename cannot write through it. `candidates` uses `symlink_metadata` + `is_file`, so symlinks are never candidates.

**Receipt contract.** Rendered from `QuarantineReport` after the walk, never from a second filesystem pass. Entries are pushed only after post-rename verification, so it cannot claim a move that failed. A hard interrupt loses the receipt but never the content — the `.md.disabled` sibling and the verified backup are both already on disk. `shell_quote` is correct POSIX single-quoting (`'\''`), and `restore_command_survives_a_hostile_filename` genuinely executes the printed `mv` through `sh` against `it's $(touch PWNED-SUB) \`touch PWNED-TICK\`.md` and asserts nothing ran.

**Locking.** `with_agent_manifest_lock` takes an `fd_lock` write guard held for the whole of `sweep_locked`; the ledger is loaded inside it. The `is_dir()` pre-check runs before the lock, so the documented "never creates the tier directory" property holds.

**Roster cannot drift.** `bundled_roster` is one `pub fn` in `core::bundled_roster`; `doctor_asset_tier.rs` now imports it and its private copy is deleted. Verified in the diff.

**Gate 4 against real files.** 42/42 agents in the live `$CLAUDE_CONFIG_DIR/agents` classify `TrustyMpm`; every file in the operator's separate `~/.claude/agents` classifies `ClaudeMpm` or `Unrecognized`. The whitelist tracks the emitter on real data.

**Versions.** `trusty-agents-common` 0.4.0 → 0.5.0 is correctly a minor: the only change to an existing module is `tier_audit.rs`, and it is doc-comment-only. Four new public modules, no signature changes. `Cargo.lock` diff is exactly the two version lines.

**Gates.** `check_test_pointers.sh` 20502 citations / 0 dangling; `check_doc_numbers.sh` 0 violations; `check_sld.sh` 0 errors; `check_line_cap.sh` 0 violations; `check_changelog_fragment.sh` both crates recorded.

**Pre-existing red confirmed, not caused here.** `git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/` is empty. Neither crate is excluded, `#[ignore]`d, or `cfg`-gated anywhere in the diff.

**Rebase.** Merge base `a28698c8`; main has three commits since (`1a424313`, `94ff67b8`, `9cdc0c48`), all in `trusty-search`/`trusty-common`. The only overlapping file is `Cargo.lock`. Textual, not semantic — nothing to carry.

## Notes

- Both probes were temporary appends, run, then reverted. `git status --porcelain` is clean on the review worktree.
- Out of scope, not blocking, but worth knowing: `prepare_session_inner` binds the sweep to `project_dir/.claude/agents`. If `project_dir` is ever the operator's home and `CLAUDE_CONFIG_DIR` is its default `~/.claude`, that path **is** the canonical tier, whose 42 files all classify `TrustyMpm`. `retract_framework_agents` on `main` already has identical exposure at that call site and would delete them first, so this PR does not introduce it — but it is the shape #4448 gate 3 exists to catch, and the guard test only proves the sweep does not read `fw.claude_agents_dir()`.
- Both findings are narrow. Fix gate 3's exit-code branch and tighten gate 4's positive identification, add the two regression tests, and I would run this unattended.

bobmatnyc:
## Round 2 — both blocking defects fixed, at `10d3d96d`

Thank you for executing both findings rather than reasoning about them. Both reproduced here, both are now regression tests, and one of them changed the fix you suggested.

### CRITICAL — fixed, and the suggested fix would still have failed open

You were right that the exit code is not a classifier. But the suggested branch —

```rust
if why.contains("not a git repository") { IndexState::NoRepo }
```

— still fails open on the trigger you named yourself. Measured on this machine:

| Condition | stderr | exit |
|---|---|---|
| genuinely no repo | `fatal: not a git repository (or any of the parent directories): .git` | 128 |
| **stale worktree pointer** | `fatal: not a git repository: (null)` | 128 |
| bad `repositoryformatversion` | `fatal: Expected git repo version <= 1, found 99` | 128 |
| missing directory | `fatal: cannot change to '…': No such file or directory` | 128 |
| bare repo | *(none)* — prints `false` | **0** |

The stale-pointer case contains `not a git repository` while meaning the opposite of absence — and that is exactly the state the ~70 worktrees orphaned on 2026-07-21 were left in, which your finding cites. So the match is on the parenthesised form only:

```rust
const NO_REPO_STDERR: &str = "not a git repository (or any of the parent directories)";
```

Everything else, including any future git wording drift, falls through to `Unavailable`. The `stdout.trim() == "true"` check you asked for is in too, so a bare repo no longer passes.

**M27 exists specifically to pin this:** substituting the short phrase kills `a_stale_worktree_pointer_is_unknown_not_no_repo`.

### HIGH — fixed with a positive marker, not a stricter whitelist

Both of your suggested options were measured and rejected:

- **Demand `initialPrompt:`.** Not universal — `default_initial_prompt` (`builder.rs:315-340`) returns `None` for the `base`, `ticketing`, `memory-manager` and agent/skill-manager roles. Requiring it would exempt `ticketing.md`, `memory-manager.md`, `mpm-agent-manager.md` and `mpm-skills-manager.md` — four of the twelve filenames claude-mpm collides on, i.e. the most shadowing-prone ones.
- **Stamp a provenance key at compose time.** This one is worth recording because it is actively backwards: a new key marks only files written by NEW binaries. The files this sweep exists to move were written by OLD ones, so gating on it makes the feature a no-op against its own target.

What is used instead is the composer's own base preamble in the BODY:

```
# BASE-AGENT — Foundation for all trusty-mpm agents
```

`compose_agent` concatenates the `extends:` chain base-first, and all 42 bundled agents' chains root at `base-agent` (`every_bundled_agent_roots_at_base_agent` pins that; `trusty-code`'s set uses the identical heading). Verified against the real deployed `rust-engineer.md`. And it is **byte-identical across every revision of BASE-AGENT.md back to 2026-05-19**, including the pre-rename `trusty-mpm-core` path — which is the property that matters, since the targets are old-binary output.

Your exact fixture is now a test, with an assertion that it declares nothing exotic so it cannot pass for the accidental reason the real file did:

```
test agents::agent_schema::tests::schema_of_a_minimal_hand_authored_file_with_no_exotic_key ... ok
test agents::quarantine::tests::quarantine_never_moves_a_minimal_hand_authored_file ... ok
```

Drift in either direction is caught: `the_bundled_base_agent_carries_the_composition_marker` (in `trusty-mpm`, which owns the asset) and `composed_marker_survives_composition` (that composition still carries the base body forward). A heading edit causes a REFUSAL, never a wrong move.

### Two of the three LOWs, fixed here

- **Receipt collision** — now routes through `free_path` like the backups. `a_same_second_rerun_does_not_overwrite_the_first_receipt`.
- **never-deletes pin coverage** — extended to `quarantine_shadows.rs`. `the_wiring_module_never_deletes`.
- **Non-UTF-8 `display()`** — left as you dispositioned it (Promote). It needs an `OsStrExt` design call that is not unix-portable as written, and candidate filenames are already UTF-8-gated so no move can go astray; only a printed command could name a wrong path.

### Mutation set re-run — and it caught a regression I would otherwise have shipped

Changing gate logic un-killed two mutations. **M16** (drop `role` from the identity check) and **M17** (accept unknown keys) were killed in round 1 and **SURVIVED** in round 2 — the new body marker was masking them, so the key-set terms had stopped being load-bearing in any test.

Fixed by restating those assertions with the marker PRESENT, so each of the three terms in the `TrustyMpm` decision is independently pinned. Re-run:

| # | Mutation | Round 2 |
|---|---|---|
| M16 | `role` no longer required | KILLED — `schema_requires_role` |
| M17 | unknown keys accepted | KILLED — `schema_rejects_an_unknown_key_even_with_a_composed_body` |
| M34 | `name` no longer required (new) | KILLED — `schema_requires_name` |

Full round-2 battery — **35 mutations, 35 killed**. New ones:

| # | Mutation | Killed by |
|---|---|---|
| M26 | gate 3 reverted to exit-code-only | `an_unreadable_repo_is_unknown_not_no_repo`, `a_stale_worktree_pointer_is_unknown_not_no_repo`, `quarantine_refuses_a_tracked_file_when_git_cannot_be_read` |
| M27 | gate 3 uses the SHORT stderr phrase | `a_stale_worktree_pointer_is_unknown_not_no_repo` |
| M28 | gate 3 drops `stdout == "true"` | `a_bare_repo_is_unknown` |
| M29 | gate 4 drops the body marker | `schema_of_a_minimal_hand_authored_file_with_no_exotic_key` +2 |
| M30 | marker text drifts | `schema_of_a_composed_trusty_mpm_agent` +3 |
| M31 | receipt bypasses `free_path` | `a_same_second_rerun_does_not_overwrite_the_first_receipt` |
| M33 | key/body split loses the body | `composed_marker_survives_composition` +3 |
| M35 | BASE-AGENT.md heading edited | `the_bundled_base_agent_carries_the_composition_marker` |
| M36 | a bundled agent loses its `extends:` root | `every_bundled_agent_roots_at_base_agent` |
| M37 | `quarantine_shadows.rs` gains a deletion | `the_wiring_module_never_deletes` |

**M23 and M24 re-verified** after the fixture change: deleting the `prepare_session_inner` call site still gives `4300 passed; 1 failed`; re-aiming at `fw.claude_agents_dir()` still fails 2.

### Gates — rung 5

```
cargo fmt --check                                                  exit 0
cargo clippy -p trusty-agents-common --all-targets -- -D warnings   0 warnings
cargo clippy -p trusty-mpm --all-targets -- -D warnings             0 warnings
cargo test -p trusty-agents-common          445 passed; 0 failed; 0 ignored
cargo test -p trusty-agents-common -- --include-ignored
                                            445 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm                   4301 passed; 0 failed; 7 ignored
                                           1375 passed; 0 failed; 1 ignored  (x2 bins)
                                            + every integration suite: ok
cargo test -p trusty-agents (dependent)    3312 passed; 0 failed; 11 ignored
cargo test -p trusty-code   (dependent)    1600 passed; 0 failed; 4 ignored
bash scripts/check_line_cap.sh       3613 tracked .rs files, 0 violations — OK
bash scripts/check_sld.sh            54 spec docs + 3040 code files, 0 errors
bash scripts/check_test_pointers.sh  20515 citations, 0 dangling — OK
bash scripts/check_changelog_fragment.sh  both crates recorded — OK
```

**One flake, reported rather than buried.** The first `cargo test -p trusty-mpm` pass of this round went `4300 passed; 1 failed` on `client::executor::tests::execute_doctor_against_test_daemon`. It is the documented baseline flake — `docs/reference/test-ladder-baseline.md:19`, "takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load", and this ran under a concurrent mutation battery. Evidence it is not mine: `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/` is empty, it passed 3/3 in isolation (8.20 s, 9.63 s, 7.53 s), and the immediately following full pass was `4301 passed; 0 failed`.

`cargo check --workspace` still carries only the pre-existing `trusty-{code,mpm}-gui` `frontendDist`/`ui/dist` failure; the diff touches zero files in either crate.

Nothing was pointed at a real deployed directory — every fixture is a `TempDir`, and the new gate-3 tests break repositories they created themselves.

bobmatnyc:
## Verdict: BLOCK

Round 2, run read-only from `.claude/worktrees/critic-4727-r2` at `10d3d96d`, with mutations executed in a separate disposable worktree. Both worktrees verified clean and removed.

**My round-1 gate-3 suggestion was wrong, and the author is right to have rejected it.** I reproduced the stderr table myself on git 2.54.0 and it holds exactly:

| Condition | stderr | exit |
|---|---|---|
| genuinely no repo | `fatal: not a git repository (or any of the parent directories): .git` | 128 |
| stale worktree pointer | `fatal: not a git repository: (null)` | 128 |
| bad `repositoryformatversion` | `fatal: Expected git repo version <= 1, found 99` | 128 |
| bare repo | *(none)* — prints `false` | **0** |

The stale-pointer case contains the short phrase while meaning the opposite, so matching `not a git repository` would have swept the ~70 orphaned worktrees from 2026-07-21 — the exact incident my own finding cited. **M27 executed and reproduced:** substituting the short phrase kills `a_stale_worktree_pointer_is_unknown_not_no_repo` (`444 passed; 1 failed`). The `stdout.trim() == "true"` check is also load-bearing — disabling it kills `a_bare_repo_is_unknown`. My suggestion would have shipped the same defect in a new disguise.

The block is a **residual** of the same gate, roughly 90% smaller than round 1's, and I found it by attacking the new predicate rather than defending the old suggestion.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-agents-common/src/agents/vcs_claim.rs` | 148-157 | An **unreadable** `.git` emits the exact parenthesised `NO_REPO_STDERR`, so a live repository holding a **tracked** file classifies `NoRepo` → `Unclaimed` → **moved**. `git` cannot distinguish the two; this code can. Same for `GIT_CEILING_DIRECTORIES`. | Add an independent existence check: if an ancestor carries a `.git` entry, git saying "no repository" is a contradiction → `Unavailable`. Patch below. | Fix here |

## Required Changes

### CRITICAL — gate 3 still fails open on an unreadable `.git`, and I moved a tracked file to prove it

The shipping changelog fragment for this PR now claims:

> it never moves a git-tracked file, **a file in a repository git cannot read**, a claude-mpm artifact, …

I moved a git-tracked file in a repository git cannot read.

Executed, in a disposable worktree, reverted afterwards:

```rust
let f = Fixture::new();
let content = tm_agent("qa");
let original = f.write("qa.md", &content);
f.git_add(".claude/agents/qa.md");            // TRACKED

let healthy = f.sweep(&roster(&["qa"])).unwrap();
assert!(healthy.moved.is_empty());            // passes — GitTracked
assert_eq!(healthy.skipped[0].reason, SkipReason::GitTracked);

// same repo, same committed file — .git is now unreadable
std::fs::set_permissions(&f.project.join(".git"), Permissions::from_mode(0o000))?;

let report = f.sweep(&roster(&["qa"])).unwrap();
assert_eq!(report.moved.len(), 1);            // PASSES
assert!(!original.exists());                  // the TRACKED file was moved
```

```
running 1 test
PROBE report = QuarantineReport { moved: [QuarantinedAgent { name: "qa", from: ".../.claude/agents/qa.md",
  to: ".../.claude/agents/qa.md.disabled", backup: ".../agent-quarantine/run-1/qa.md" }],
  skipped: [], failed: [], receipt: Some(...), receipt_error: None }
PROBE original still exists = false
test agents::quarantine::tests::critic_probe_unreadable_git_dir_moves_a_tracked_file ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 445 filtered out
```

Measured directly against git 2.54.0, from a tier inside a repo that tracks `qa.md`:

| Broken thing | stderr | verdict |
|---|---|---|
| `.git` dir mode 000 | `fatal: not a git repository (or any of the parent directories): .git` | **NoRepo — WRONG** |
| `.git/HEAD` unreadable | `fatal: not a git repository (or any of the parent directories): .git` | **NoRepo — WRONG** |
| `GIT_CEILING_DIRECTORIES` blocks discovery | `fatal: not a git repository (or any of the parent directories): .git` | **NoRepo — WRONG** |
| `.git/config` unreadable | `fatal: unable to access '.git/config': Permission denied` | Unavailable — correct |
| `.git` worktree-pointer file unreadable | `fatal: error opening '…/.git': Permission denied` | Unavailable — correct |
| `.git/index` unreadable | `rev-parse` → `true`, then `ls-files` exits **128** | Unavailable — correct |
| stale pointer / bad repo version / dubious ownership | distinct messages | Unavailable — correct |

The module's own 🔴 doc block names this exact case as the thing it fixes:

> Reading any non-zero exit as `NoRepo` … sends a live work tree **git merely declined to read** into the SWEEPABLE state, and its committed agent files get moved.

An unreadable `.git` *is* "a live work tree git merely declined to read". Git reports the parenthesised form because it never got far enough to know a repository was there — but the code can check independently, since `symlink_metadata` on the `.git` entry needs only the parent's read bit, which is present:

```rust
Ok(out) if !out.status.success() => {
    let why = String::from_utf8_lossy(&out.stderr);
    // Git emits the SAME "no repository" text when it found a `.git` it could
    // not read (mode 000, an unreadable HEAD, GIT_CEILING_DIRECTORIES). Stat is
    // an independent witness: it needs only the PARENT's read bit.
    let no_repo = why.contains(NO_REPO_STDERR)
        && !dir
            .ancestors()
            .any(|p| p.join(".git").symlink_metadata().is_ok());
    return Self {
        state: if no_repo { IndexState::NoRepo } else { IndexState::Unavailable },
    };
}
```

Both error directions of that check are refusals, which is the fail-closed side. Add the regression as a sibling of `quarantine_refuses_a_tracked_file_when_git_cannot_be_read`, driving it with `set_permissions(.git, 0o000)` instead of `repositoryformatversion`.

**Blast radius, stated honestly so this is not read as round-1 sized.** Round 1 admitted every 128: dubious ownership, stale pointers, broken repo versions, failing `git` shims — and gate 4 was weak enough that hand-authored files were in scope. All of that is now closed; I re-ran each and every one refuses. What is left needs a permission-restricted `.git` (root-owned mode-700 checkout, a container-created `.git` on a host mount, a squashed NFS mount) *and* a file that is tm's own composer output *and* committed. The move stays reversible — verified backup, `.md.disabled` sibling, receipt. It is narrow. It is also a stated contract in the shipping changelog that an executable test falsifies, in a gate whose only job is to fail closed, in a pristine release, with a three-line fix.

## Verified — the author's counter-evidence, and the rest

**The mutation regression is real, and the fix is not superficial.** All three terms of the `TrustyMpm` decision are now independently load-bearing. Executed against `10d3d96d`:

| # | Mutation | Result |
|---|---|---|
| M16 | `has_identity` stops requiring `role:` | **KILLED** — `schema_requires_role`, `444 passed; 1 failed` |
| M17 | `all_known = true` (accept unknown keys) | **KILLED** — `schema_rejects_an_unknown_key_even_with_a_composed_body`, `444 passed; 1 failed` |
| new | `composed = true` (drop the body marker) | **KILLED ×3** — `schema_of_a_minimal_hand_authored_file_with_no_exotic_key`, `the_body_marker_is_what_separates_composed_from_hand_authored`, `quarantine_never_moves_a_minimal_hand_authored_file`, `442 passed; 3 failed` |
| M27 | `NO_REPO_STDERR` → short phrase (my round-1 suggestion) | **KILLED** — `a_stale_worktree_pointer_is_unknown_not_no_repo` |
| G3b | drop the `stdout == "true"` bare-repo check | **KILLED** — `a_bare_repo_is_unknown` |
| G3c | revert to any-non-zero → `NoRepo` | **KILLED ×3** including `quarantine_refuses_a_tracked_file_when_git_cannot_be_read` |

M16 and M17 are killed by fixtures that now carry the marker, which is exactly the restatement the author described. Re-running the battery after changing gate logic caught a regression the round-1 battery could not have — that is the single most valuable thing in this PR, and it generalises beyond it.

**Gate 4's marker survives every attack I could construct.**

- *Can a hand-authored file carry it?* Not on real data: 0 of 6 files in the operator's `~/.claude/agents` carry it, and all six classify non-movable (`Unrecognized` ×5, `ClaudeMpm` ×1). `writing-critic.md` — the file that flipped to `TrustyMpm` in round 1 once its folded description was unwrapped — is `Unrecognized` now and stays so unwrapped, because the marker is absent. That escape was luck; this is a gate.
- *Can a claude-mpm file carry it?* Structurally irrelevant: the foreign-key check returns `ClaudeMpm` **before** the body is read. `~/.claude/agents/base.md` confirms it on a real file.
- *Can a legitimately generated file lack it?* Only by refusing to move — the safe direction. All 42 bundled agents' `extends:` chains root at `base-agent` (31 via `base-engineer`/`base-qa`/`base-ops`/`base-research`, 6 direct, 4 BASE-* files, `BASE-AGENT.md` itself the root). 42/42 live deployed agents in `$CLAUDE_CONFIG_DIR/agents` carry the marker.
- *Is the heading really byte-identical back to 2026-05-19?* Yes. Verified against **every** revision of `BASE-AGENT.md` in history — 23 blobs across 17 commits, including `trusty-code`'s copy and the pre-rename `crates/trusty-mpm-core/assets/agents/BASE-AGENT.md` at `13f9fa2c` (2026-05-19), where it is line 6 verbatim.

The author's rejection of both my gate-4 suggestions holds. `initialPrompt:` would have exempted four of the twelve colliding filenames, and a compose-time provenance key marks only files written by *new* binaries — a no-op against the sweep's own targets. The body marker is the only signal that is both universal and retroactive.

**Both LOWs fixed.** The receipt now routes through `free_path` with a `MAX_COLLISION_ATTEMPTS` error path, pinned by `a_same_second_rerun_does_not_overwrite_the_first_receipt`. The never-deletes pin gained a sibling in `crates/trusty-mpm/src/core/session_launch/tests_quarantine_4448.rs:388` covering `quarantine_shadows.rs`.

**The third LOW's disposition is legitimate — the UTF-8 gate is real.** `candidates` (`quarantine.rs`) does `entry.file_name().to_str()?`, so a non-UTF-8 filename is never a candidate and no move can go astray. The residual is a cosmetic mis-rendered parent path in the printed `mv`. Documenting rather than fixing is the right call.

**Nothing points at a real deployed directory.** No `home_dir`, `CLAUDE_CONFIG_DIR`, or literal `.claude/agents` reference in any of the three test modules. Every fixture is a `TempDir`, and the new gate-3 tests break repositories they created themselves (`git config core.repositoryformatversion 99` scoped to the fixture, a stale `.git` pointer written into a `TempDir`, a bare repo `git init` into a `TempDir`).

**Gates.** `cargo fmt --check` clean. `cargo clippy -p trusty-agents-common -p trusty-mpm --all-targets` — no errors. `cargo test -p trusty-agents-common` → `445 passed; 0 failed`. `cargo test -p trusty-mpm` → `4301 passed; 0 failed; 7 ignored` plus 29 further green targets. `execute_doctor_against_test_daemon` did not flake in this pass.

## Notes

- Every mutation and probe ran in a throwaway worktree at the same SHA, each reverted immediately; both worktrees were `git status --porcelain`-clean before removal. The PR worktree was never edited.
- `GIT_CEILING_DIRECTORIES` is the second trigger for the same branch and is closed by the same patch. I did not write a test for it — setting it is process-global and would race the parallel test runner. The stat-based fix covers it without one.
- Not flagged, below 80%: `VcsIndex::claim` compares file names case-sensitively against git's index, so a tracked `QA.md` checked out as `qa.md` on a case-insensitive filesystem would read `Unclaimed`. It needs a case-changing rename git never saw, and I could not construct a realistic path to it.
- Fix the one branch and add the `0o000` regression and I would run this unattended. Everything else in this round verified.

bobmatnyc:
## Round 3 — residual fixed, at `e7b7692b` (rebased onto `f9d29c08`)

You were right, and the probe is now a passing regression test. I verified the proposed guard before applying it as asked, and it is correct on every case I could construct — with one addition it did not cover.

### Verified against git 2.54.0

| Case | stderr | ancestor `.git`? | verdict |
|---|---|---|---|
| `.git` mode 000 | parenthesised | yes | Unavailable ✓ |
| `.git/HEAD` unreadable | parenthesised | yes | Unavailable ✓ |
| `GIT_CEILING_DIRECTORIES` | parenthesised | yes | Unavailable ✓ |
| stray empty `.git` dir | parenthesised | yes | Unavailable — over-refusal, fail-closed |
| genuinely empty dir | parenthesised | **no** | NoRepo ✓ |
| `.git` file → live worktree | *(exit 0, `true`)* | n/a | never reaches the branch ✓ |
| symlinked `.git` (healthy) | *(exit 0, `true`)* | n/a | never reaches the branch ✓ |

Your three suspected holes are all closed: a **bare repo** exits 0 printing `false` and is caught by the `stdout == "true"` branch before this one; a **`.git` file** lstats fine; a **healthy symlinked `.git`** never fails.

### One addition — the witness must be , and I have the case that proves it

A **dangling `.git` symlink** lstats successfully and stats as absent:

```
symlink_metadata: ok      metadata: err
git: fatal: not a git repository (or any of the parent directories): .git
```

`metadata()` would follow the link, conclude "nothing there", and hand back the sweepable answer for a project whose `.git` is merely broken. **M40** pins it — swapping to `metadata()` kills `a_dangling_git_symlink_is_unknown_not_no_repo`.

### And one that says the parenthesised form still earns its keep

I expected the witness to make `NO_REPO_STDERR`'s precision redundant. It does not — the two defences do not overlap. A bogus `GIT_DIR` in a directory with **no ancestor `.git`**:

```
fatal: not a git repository: '/nonexistent/x'
```

matches the short phrase, and the witness finds nothing to contradict it. Shortening the constant would return `NoRepo` for a misconfigured repository. **M42** pins it.

### Also closed: a relative-path hole in the ancestor walk

`Path::ancestors()` on a relative `dir` walks `.claude/agents` → `.claude` → `""`, misses the project's `.git`, and lands back on the permissive answer. Not reachable from today's call sites, which pass absolute paths — but one caller away, and it fails in the dangerous direction. The path is canonicalised first and a canonicalisation failure is itself `Unavailable`. **M41** pins it.

### Changelog claim — corrected, not narrowed away

You were right that it was falsifiable. It now states the mechanism rather than a guarantee:

> a file moves only on a POSITIVE answer from every gate … Anything less — a tracked file, an unreadable or ceiling-blocked repository, a bare repo, a stale worktree pointer, git absent, a claude-mpm artifact, a hand-authored file — is left alone and recorded in the receipt with the reason

### Mutation battery — 45 run, 45 killed, and it caught three regressions again

Re-running after the guard was not a formality. **Three previously-killed mutations SURVIVED**, all masked by the new guard:

| # | Mutation | Why it survived | Now killed by |
|---|---|---|---|
| M42 | `NO_REPO_STDERR` → short phrase | the witness covered the stale-pointer case that used to kill it | `a_bogus_git_dir_message_is_unavailable_even_with_no_ancestor_git` |
| M40 | witness uses `metadata()` | no fixture had a dangling `.git` symlink | `a_dangling_git_symlink_is_unknown_not_no_repo` |
| M30 | `COMPOSED_BASE_MARKER` shortened to a prefix | the asset pin used `contains`, which a prefix satisfies | `the_bundled_base_agent_carries_the_composition_marker`, now **equality** |

M30 is worth flagging: I had the pin comparing a local copy of the string by `contains`. `COMPOSED_BASE_MARKER` is now `pub` so `trusty-mpm` — which ships `BASE-AGENT.md` — asserts the heading line **equals** the real constant. Drift in either direction now fails (M30 and M45 both confirmed).

New this round: M38 (guard removed) → 6 failures; M39 (checks only `dir`, not ancestors) → 5; M41, M43, M44 all killed.

### Gates — rung 5, rebased onto `f9d29c08`

```
cargo test -p trusty-agents-common          455 passed; 0 failed; 0 ignored
cargo test -p trusty-agents-common -- --include-ignored
                                            455 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm                   4301 passed; 0 failed; 7 ignored
                                           1375 passed; 0 failed; 1 ignored  (x2 bins)
                                            + every integration suite: ok
cargo test -p trusty-agents (dependent)    3312 passed; 0 failed; 11 ignored
cargo test -p trusty-code   (dependent)    1600 passed; 0 failed; 4 ignored
cargo clippy (both crates) -D warnings      0 warnings
cargo fmt --check                           exit 0
line-cap    3616 tracked .rs files, 0 violations — OK
sld-lint    54 spec docs + 3041 code files, 0 errors
test-pointers  20571 citations, 0 dangling — OK
changelog-fragment  both crates recorded — OK
```

No flake this round — `execute_doctor_against_test_daemon` passed in the full run.

**Rebase:** clean, 4 commits replayed, no conflicts. Version bumps survived (`trusty-agents-common` 0.5.0, `trusty-mpm` 1.3.5 — main is still on 0.4.0/1.3.4, so the parity gate stays satisfied). `cargo metadata --locked` exits 0, so `Cargo.lock` needs no regeneration.

`cargo check --workspace` still carries only the pre-existing `trusty-{code,mpm}-gui` `ui/dist` failure; the diff touches zero files in either crate.

Not merged. Fixtures only — the new tests break repositories they created themselves in `TempDir`, and every permission change is restored by a `Drop` guard so a panic cannot strand a mode-000 directory.

bobmatnyc:
# code-critic — round 3

## Verdict: APPROVE

Head `e7b7692b`, rebased onto `f9d29c08`. My round-2 CRITICAL is closed. I re-ran
both probes that moved a tracked file last round, plus nine more of my own, against
this head — every one refuses. Two MEDIUM findings below; neither blocks.

Reviewed from a dedicated read-only worktree at the PR head. Rounds 1–2 items were
not re-litigated.

---

## 1. The residual guard — verified independently

I confirmed git's behaviour at the shell first (git 2.54.0), so the premises are mine,
not the author's:

| Trigger | git stderr | exit |
|---|---|---|
| `.git` mode 000, file committed | `fatal: not a git repository (or any of the parent directories): .git` | 128 |
| `GIT_CEILING_DIRECTORIES` below the repo | same parenthesised text | 128 |
| dangling `.git` symlink | same parenthesised text | 128 |
| `GIT_DIR=/nonexistent/x`, no ancestor `.git` | `fatal: not a git repository: '/nonexistent/x'` | 128 |
| stray empty `.git` dir | same parenthesised text | 128 |
| genuinely empty dir | same parenthesised text | 128 |

All three measured claims hold:

- **`symlink_metadata`, not `metadata`** — confirmed. A dangling `.git` symlink lstats
  ok and stats absent; `metadata()` would return the permissive answer. Mutation below.
- **The parenthesised form is still load-bearing** — confirmed and it is the *only*
  defence in that case. Row 4 above has no ancestor `.git`, so the witness finds nothing
  to contradict a shortened phrase. The two defences genuinely do not overlap there.
- **The relative-path hole** — `classify_failure` canonicalises first
  (`vcs_claim.rs:117`), and a canonicalisation failure maps to `Unavailable`. Note the
  hole was already unreachable for a second reason: `quarantine_shadowing_agents` returns
  early on `!tier_dir.is_dir()` (`quarantine.rs:171`), so by the time `probe` runs the
  directory exists and a relative path canonicalises against the cwd. Closing it anyway
  is correct — it costs nothing and the failure direction was dangerous.

### Does `canonicalize` follow a symlink out of the intended tree?

It resolves symlinks, so yes it can leave the logical tree — but it **cannot diverge from
git**, because `git -C <dir>` chdirs and walks the *physical* chain too. I verified both
directions:

- Tier reached through a symlinked project dir, file tracked → `GitTracked`, refused
  (PROBE G).
- `.claude/agents` symlinked out of the repo to a non-repo dir → `git ls-files` records
  the *symlink* as the tracked object, not its contents, and `git -C` through it reports
  no repository. The witness agrees. The files inside are genuinely unclaimed by that
  repo, so the answer is right.
- `.claude/agents` symlinked into a *different* repo that tracks the file → that repo
  answers, `Claimed`, refused.

No divergence. This was the right resolution mode to pick.

### My own probe battery (real `quarantine_shadowing_agents`, staged from scratch)

```
[PASS] control / no repository:                        moved=1 skipped=[]
[PASS] control / healthy repo, untracked:              moved=1 skipped=[]
[PASS] control / healthy repo, TRACKED:                moved=0 skipped=[GitTracked]
[PASS] PROBE A / tracked + unreadable .git:            moved=0 skipped=[VcsUnknown]
[PASS] PROBE A2 / tracked + unreadable .git/HEAD:      moved=0 skipped=[VcsUnknown]
[PASS] PROBE B / tracked + GIT_CEILING_DIRECTORIES:    moved=0 skipped=[VcsUnknown]
[PASS] PROBE C / dangling .git symlink:                moved=0 skipped=[VcsUnknown]
[PASS] PROBE D / bogus GIT_DIR, no ancestor .git:      moved=0 skipped=[VcsUnknown]
[PASS] PROBE E / stray empty .git dir:                 moved=0 skipped=[VcsUnknown]
[PASS] PROBE F / bare repository:                      moved=0 skipped=[VcsUnknown]
[PASS] PROBE G / tracked, via symlinked project dir:   moved=0 skipped=[GitTracked]
```

The two controls matter as much as the refusals: the gate is not frozen shut. A and B
are the round-2 killers. Both refuse.

---

## 2. Mutations — the three regressions reproduce as killed

Applied in a throwaway worktree at the same commit, one at a time, restored between runs.

| # | Mutation | Result |
|---|---|---|
| M42 | `NO_REPO_STDERR` → `"not a git repository"` | **killed** — `a_bogus_git_dir_message_is_unavailable_even_with_no_ancestor_git` (241 passed, 1 failed) |
| M40 | `symlink_metadata()` → `metadata()` | **killed** — `a_dangling_git_symlink_is_unknown_not_no_repo` (241 passed, 1 failed) |
| M30a | `COMPOSED_BASE_MARKER` → `"# BASE-AGENT"` | **killed** — `the_bundled_base_agent_carries_the_composition_marker` |
| M30b | asset heading changed, constant untouched | **killed** — same test |

**The M30 pin is genuinely bidirectional.** I mutated each side independently and both
fail. `contains` → `assert_eq!` on the asset's first `# ` line is the right shape.

### Is the battery stable, or is there a class it cannot reach?

There is one, and I found it by mutation rather than by reading:

**`FailStage::Verify` is unreachable from any test.** Deleting the byte-identical backup
comparison in `quarantine.rs:365-382` leaves the suite fully green (242 passed, 0 failed).
Neither `Verify` branch — the backup compare nor the post-rename confirmation at
`quarantine.rs:397` — has a single assertion anywhere; only `FailStage::Backup` and
`FailStage::Rename` are asserted (`quarantine_tests.rs:575,617`). The class is
"I/O-fault verification branches with no injection seam", and the module deliberately has
no seam by design.

Four other mutations I tried were all killed: gate-4 `Unrecognized` failing open (3 tests),
dropping the canonicalise (1), truncating the ancestor walk to the dir itself (6).

The consequence of the gap is bounded and worth stating: the quarantine **renames** the
original to `.md.disabled` rather than copying-and-deleting, so the original bytes survive
regardless of whether the backup is sound. A silently-skipped verification loses nothing.
That is why this is MEDIUM and not higher.

---

## 3. The changelog claim

The round-2 falsifiable absolute — "never moves a file in a repository git cannot read" —
is gone, and its replacement is accurate. `agents-common`'s fragment is correct throughout,
including "git does not claim it", which covers the no-repository case properly.

One clause in the `trusty-mpm` fragment is still wrong, in the direction of over-claiming
protection. See MEDIUM-1.

---

## 4. Rebase and hygiene

- **Nothing was dropped.** `git range-diff` over the four replayed commits reports all four
  `=` (byte-identical content):
  ```
  1: 915bdd7d = 1: 00990eb4   2: 64b68930 = 2: f6ec974b
  3: 2796483f = 3: b1df93cb   4: 10d3d96d = 4: 1c7205a6
  ```
  `e7b7692b` is the new round-3 commit on top. 23 files, as before.
- **`cargo metadata --locked` exits 0**, no lockfile regeneration. The PR's lock delta vs
  its merge-base touches only `trusty-agents-common 0.4.0→0.5.0` and
  `trusty-mpm 1.3.4→1.3.5`.
- **One stale premise in the brief:** `origin/main` has moved past `f9d29c08` and already
  carries `trusty-mpm 1.3.5` (another PR's bump; the CHANGELOG's newest released section is
  still 1.3.4). Both sides therefore set the same value — no conflict, GitHub reports
  MERGEABLE, and 1.3.5 will simply ship both changesets. `trusty-agents-common` 0.5.0 vs
  main's 0.4.0 is a clean minor bump, correct for an additive API change (`pub` marker,
  four new modules).
- **Test isolation holds.** Every new fixture is a `TempDir`; `tests_quarantine_4448.rs`
  additionally overrides `$HOME` via `EnvVarGuard` under `#[serial]`. No `home_dir()`, no
  absolute path, no reference to a live tier anywhere in the new tests. Three `chmod 000`
  sites (`quarantine_tests.rs:308`, `vcs_claim_tests.rs:264,283`) each construct their
  `ModeGuard` on the line *before* the chmod, so a panic cannot strand a mode-000 entry.
  The `GIT_CEILING_DIRECTORIES` mutation captures and restores the prior value on drop.
- **Gates run at this head:** `cargo fmt --check` clean; `cargo clippy -p trusty-agents-common
  -p trusty-mpm --all-targets -- -D warnings` clean; `cargo test -p trusty-agents-common`
  455 passed / 0 failed; `cargo test -p trusty-mpm` 4301 + 1375 + 1375 + … all ok, 0 failed;
  `cargo test -p trusty-agents --lib` (direct dependent) 3312 passed / 0 failed;
  `cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui` clean;
  `scripts/check_line_cap.sh` 0 violations; `scripts/check_changelog_fragment.sh` both crates
  recorded. The two Tauri `ui/dist` failures are pre-existing — this PR touches 0 files in
  either crate. No `execute_doctor_against_test_daemon` flake.

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/changelog.d/4448-quarantine-shadowing.md` | 6 | "a file moves only on a POSITIVE answer from every gate: … **git reports a readable work tree that does not track it**" is false for a project with no repository at all. `IndexState::NoRepo` maps to `VcsClaim::Unclaimed` (`vcs_claim.rs:261`) and the file moves — my control probe #1 moved it, and `quarantine_works_without_a_repository` pins that behaviour. An operator on a non-git project reads this line and concludes their tier is untouchable. | Say what the gate actually answers: "git either finds no repository at all, or reports a readable work tree that does not track it". The "Anything less — …" enumeration that follows is already correct and already omits the no-repo case, so only this clause needs the edit. | Fix here |
| MEDIUM | `crates/trusty-agents-common/src/agents/quarantine.rs` | 365-382, 397 | Both `FailStage::Verify` branches are untested. Deleting the byte-identical backup comparison leaves the whole suite green (242/242). The module's own doc calls this branch "Evidence, not an exit code", and the changelog advertises "a verified byte-identical backup" — nothing proves the verification fires. | Reach it without adding a production seam: stage the backup destination so the copy lands different from the source (e.g. pre-place `RECEIPT.md`-style collision fodder, or run the sweep with `backup_root` on a path where the copy is truncated), and assert `report.failed[0].stage == FailStage::Verify` plus `original.exists()`. Bounded consequence: the move is a *rename*, so the original bytes survive as `.md.disabled` even if the backup is unsound. | Promote |
| LOW | `crates/trusty-agents-common/src/agents/vcs_claim_tests.rs` | 308, 319-321 | The `SAFETY: this test is #[serial], so no other thread races the set/restore` justification does not hold. `serial_test::serial` serialises only against other `#[serial]` tests; the ~450 non-serial tests in this binary run concurrently on other threads and may `getenv` during the `set_var`. Behaviour is unaffected in practice — the ceiling value is a private tempdir that no concurrent fixture sits under — but the stated precondition for the `unsafe` is not established. | Reword the SAFETY comment to the real argument (no other test reads or writes this variable, and the value is scoped to a private tempdir), or move the ceiling test to its own `#[test]` binary. | Promote |

## Notes

- **Not flagged, below 80% confidence.** On a case-insensitive filesystem, a case-only
  divergence between the git index entry (`qa.md`) and the on-disk dirent (`QA.md`) would
  make `VcsIndex::claim` miss a tracked file, since the lookup is an exact `BTreeSet`
  membership test (`vcs_claim.rs:260`). Reaching it needs a case-only rename git did not
  record *and* frontmatter declaring a bundled `name:`, and the outcome is a reversible
  move with a backup and a receipt, never a delete. I could not construct it without
  deliberate mischief, so I am not asserting it is real.
- `ModeGuard` restores `.git/HEAD` to `0o755` rather than `0o644`, so the file comes back
  executable. Cosmetic — it lives inside a `TempDir` that is removed immediately after.
- `the_bundled_base_agent_carries_the_composition_marker` keys on the first line starting
  with `"# "`. `BASE-AGENT.md` line 6 is that heading today and the frontmatter carries no
  YAML comment; if one were ever added the test would fail loudly rather than silently
  match the wrong line, which is the right direction.

## Would I run this unattended against a machine holding another project's agent files?

Yes. Gate 4 rules the file IN positively rather than ruling it out by filename, gate 3 now
treats every state git cannot fully answer as a refusal, and the whole path renames rather
than deletes. The eleven-probe battery above is what convinced me, not the author's tests.

**APPROVED for merge.**

Issue #3970, opened 2026-07-26
trusty-search: periodic HNSW incremental persister overwrites the complete on-disk snapshot with partial state mid-reindex (crash-safety hole, independent of shutdown)


bobmatnyc:
## Summary
The periodic HNSW incremental persister publishes partial reindex state directly over the live on-disk snapshot during normal, healthy operation. Once it has done so, any ungraceful termination — SIGKILL, OOM-kill, abort, power loss — leaves the complete snapshot unrecoverably destroyed. This is the same vulnerability class as #1717, but reached through a different and far more frequently exercised path, and it is NOT fixed by PR #3968.

## Mechanism (verified during adversarial review of PR #3968)
- `spawn_incremental_persist` (`crates/trusty-search/src/core/indexer/persist_hnsw.rs:138-260`, call at 254-260) invokes `store.save_to(&hnsw_path)` → `UsearchStore::save()` every `HNSW_SNAPSHOT_BATCH_INTERVAL` (16) batches during **every** reindex, plus once forced after the batch loop (`service/reindex/finish.rs:212`).
- That call site is guarded ONLY by the ratio shrink guard in `UsearchStore::save()` (`core/store/usearch_store.rs:552-568`). It does not pass through the exact `ReindexStatus::Running` check added by PR #3968, which lives solely in `service/shutdown_flush.rs`.
- Reindex progress is monotonic. For the degenerate case #1717 describes — a fresh, not-yet-restored store rebuilding from 0 toward a large complete count — the persister is therefore **guaranteed** to cross the ratio guard's 50% threshold at some checkpoint before the reindex finishes, as ordinary healthy progress rather than as a failure.
- From that checkpoint onward the complete on-disk snapshot has already been replaced by a partial, still-growing one. The window between that moment and reindex completion is fully exposed to ungraceful termination.

## Why PR #3968 does not cover it
#3968 fixes the **graceful** shutdown path: `flush_one_index_on_shutdown` now skips the flush entirely while `ReindexStatus::Running`. SIGKILL, OOM-kill, panic-abort, and power loss do not run that code at all. The ratio guard, which is the only thing protecting the periodic persister, is by construction ineffective here because normal progress crosses its threshold every time on any reasonably large index.

## Suggested remedy
[...]

Issue #2922, opened 2026-07-17
trusty-search: shutdown flush truncates HNSW on timeout — non-atomic save_to + 10s per-index default


bob-duetto:
**Impact:** Quiescing trusty-search via `systemctl stop` on a box with a large index destroys the HNSW vector graph. Observed on trusty-search 0.27.1 (c7i indexer, code-intelligence): a 586MB `hnsw.usearch` was truncated to 112 bytes and `hnsw.keys.json` (40MB) to 39 bytes during a routine stop-to-archive step, silently degrading the index to BM25-only while `/health` and index-status kept reporting `semantic: ready`.

**Root cause (source-level, 0.27.1):** `src/service/shutdown_flush.rs` reads `TRUSTY_SHUTDOWN_FLUSH_TIMEOUT_SECS` and falls back to **10s** if unset; this is a per-index deadline applied sequentially across all indexes via `tokio::time::timeout` around each `save_vector_store`. `store.save_to()` is **non-atomic** (no temp-file + rename), so when the timeout cancels the write mid-flush, the partial bytes remain on disk (a header, then a cancelled payload → the 112-byte stub). 10s is far too short for a 586MB HNSW.

**Why it's worse than documented:** the log claims "on-disk state from last incremental persist" survives, but in practice the file is left truncated, not reverted.

**Suggested fixes (in priority order):**
1. Make `save_to` atomic — write to a temp file then rename, so a cancelled/timed-out flush never truncates the live file.
2. On flush timeout, leave the existing on-disk file untouched rather than a partial write.
3. Raise the default timeout and/or scale the per-index budget by index size.

**Repro:** run a large-corpus index (hundreds of MB HNSW) under 0.27.1, `systemctl stop trusty-search`, observe hnsw.usearch truncation with default TRUSTY_SHUTDOWN_FLUSH_TIMEOUT_SECS.

bobmatnyc:
Fix verified live: trusty-search 0.33.1 published to crates.io, installed via signed build (`scripts/install-trusty-search-signed.sh`, Developer ID Team 4JH68XUHC5, identifier `com.trusty.trusty-search`), and the daemon has been gracefully restarted twice on this machine.

Evidence from `~/Library/Logs/trusty-search/stderr.log` around the graceful `launchctl bootout`/`bootstrap` cycle:

```
18:27:03.222 INFO trusty-common: received SIGTERM — initiating graceful shutdown
[...]

Issue #1717, opened 2026-06-25
fix(trusty-search): shutdown should drain/cancel in-flight background reindex + guard catastrophic partial-snapshot shrink (#1711 follow-up)


bobmatnyc:
## Summary

The #1711 guard (PR #1716) catches only the **exact-0-vector** edge case: it refuses to overwrite an on-disk snapshot >100 KB when the in-memory index has `size()==0`. This is the most common production failure mode (graceful shutdown flushing a store that was promoted-but-never-populated), but it does not catch all possible data-loss scenarios.

## Root-Cause Race (still present)

Commit `fe4c0b28` (`reconcile_stale_indexes`) triggers a background reindex via `commit_parsed_batch` / `parse_and_embed_files`. When the daemon receives SIGTERM mid-run, the current shutdown path calls `save()` on every store in the registry, including one whose background reindex task is only partially complete. The reindex may have processed, say, 5 000 of 312 000 vectors before SIGTERM — the in-memory index is non-zero but catastrophically partial, and the guard's `size==0` condition is false, so the guard does **not** fire and the partial in-memory state overwrites the complete on-disk snapshot.

## Proposed Follow-up Work

1. **Drain / cancel in-flight background reindex on SIGTERM before HNSW flush.** The shutdown path should signal all in-flight reindex tasks, await their cancellation (or drain their buffered commits), and only then flush to disk. This prevents any mid-run save.

2. **Guard catastrophic partial-snapshot shrink.** Extend the save guard to also refuse a drastic vector-count shrink relative to the on-disk snapshot (e.g., new count < 50% of on-disk count, or equivalently new file size << on-disk file size). This would catch the 5k-of-312k partial-reindex case even without proper cancellation.

## Scope of #1711 Guard

The merged guard (PR #1716) is **defense-in-depth** for the most acute failure mode. It does NOT address the root-cause race. That race is tracked here.

Refs #1711

Issue #1711, opened 2026-06-25
trusty-search 0.29.0: graceful shutdown re-flushes populated main HNSW to a near-empty file (data loss on systemctl stop)


bob-duetto:
## Summary
On **trusty-search 0.29.0** (CUDA build, AL2023, Tesla T4), a graceful `systemctl stop` of the daemon **re-flushes the `main` index HNSW to a near-empty ~3,472-byte file**, destroying a fully-populated ~526 MB HNSW snapshot. Reproduced **twice**. Any operational workflow that quiesces the daemon with a graceful stop before archiving/swapping the index will silently lose the populated semantic index.

## Environment
- trusty-search **0.29.0** (`--no-default-features --features cuda`, dynamic ORT `libonnxruntime.so.1.24.2`, glibc 2.34), bundled trusty-embedderd 0.3.4.
- Index `main`: redb 4.x corpus ~2.67 GB / ~334k chunks; HNSW populated to **312,524 vectors / 526,446,720 bytes** (`hnsw.usearch`) after a successful deferred CUDA embed pass.

## Observed behavior
1. Deferred embed completes cleanly: daemon logs `deferred_embed[main]: embedded 312524/312524 chunks — marking semantic Ready`; on disk `hnsw.usearch` = **526,446,720 bytes**, `hnsw.keys.json` = 36,172,014 bytes; warm boot restores it (`hnsw_snapshot=true → stages(semantic=Ready)`), real `vector` search hits.
2. `systemctl stop trusty-search` (graceful) → on next inspection `hnsw.usearch` has been **rewritten to ~3,472 bytes** (near-empty / 0-vector), and a subsequent warm boot comes up with an empty semantic stage (BM25-only). Reproduced on two separate stop cycles.

## Expected behavior
A graceful shutdown must persist (or leave intact) the populated HNSW snapshot, not overwrite it with a near-empty file. Stop should be safe to use for quiescing before archive/swap.

## Impact
- **Data loss / index-blanking** on any graceful stop after a successful embed.
- Directly threatens index-archive and blue/green cutover procedures that `systemctl stop` the daemon to quiesce before tarring/swapping the index — the standard pattern.

## Workaround in use
Capture the populated `hnsw.usearch` (+ `hnsw.keys.json`) to a side directory **at embed-completion** (before any stop), then restore the captured snapshot and warm-boot from it immediately before archiving/publishing. This avoids relying on shutdown persistence.

## Suspected area
[...]

PR #4737, opened 2026-08-04
fix(trusty-search): recover an empty-warm-booted vector layer and stop reporting it ready (#4707)

### LINKED ISSUE #4707: `tm-duetto-agents-skills-01` search 500s after partial warm-boot
## Summary

Vector layer warm-boots empty, the #1711 data-loss guard blocks the overwrite, and search is left permanently 500ing. The index becomes inconsistent: lexical layer is ready, but search as a whole fails and stays that way with no recovery path.

## Measurements (2026-08-03)

- **Measured:** `search()` against index `tm-duetto-agents-skills-01` reproducibly returns `500 internal search error` (two attempts).
- **Lexical layer:** Attached fine — `index_status` shows 1,632 chunks, `status: ready`.
- **Vector layer:** Warm-booted EMPTY. `errors.jsonl` at 18:37:51 records usearch REFUSING to overwrite the populated `hnsw.usearch` snapshot (2,755,216 bytes on disk) with an empty in-memory index, citing the issue #1711 data-loss guard.
- **Root path:** `/Users/masa/trusty-mpm-projects/duettoresearch/duetto-agents-skills/.worktrees/tm-duetto-agents-skills-01`

## Behavior Assessment

The data-loss guard behaved CORRECTLY — the on-disk vector data is safe and was not clobbered. The defect is that there is no recovery path: the index is left in an inconsistent runtime state and stays that way.

## Questions for Maintainer

Should a vector-layer warm-boot that comes up empty while a populated on-disk snapshot exists trigger an automatic reload from that snapshot, rather than an attempted overwrite that the guard then blocks?

Should the whole index report unhealthy (not `ready`) when one layer failed?

## Reference

See issue #1711 (data-loss guard).

bobmatnyc:
## Summary

Fixes #4707 — an index whose vector layer warm-boots empty is left permanently broken (search 500s on every query) while still self-reporting `ready`.

**This is two fixes, not one.** They are independent defects that compose into the reported outcome, and either alone leaves a real hole: fixing only recovery still lets a genuinely unrecoverable index advertise a dead lane; fixing only reporting leaves a recoverable index degraded to BM25 forever. They ship together because they share one root cause — *the semantic lane's readiness is pure bookkeeping that is never reconciled against the live vector store.*

## Root cause (investigated, not assumed)

1. **No recovery path.** When `UsearchStore::load_from` discards a snapshot — the #2922 size floor, a corrupt sidecar, the #3970 torn-pairing guard — `build_store_for_entry` installs a fresh **empty** store and sets `hnsw_load_failed = true`. Warm-boot classifies semantic as `Pending`, which is correct. But the intact on-disk snapshot is never revisited: nothing retries the load, and every later `save()` from that empty store is refused by the #1711 guard. The index is permanently vectorless while a perfectly good snapshot sits on disk.

2. **Dishonest health.** A subsequent *incremental* reindex flips `semantic` to `Ready` without the store ever gaining a vector. `validate::reindex_outcome`'s `newly_submitted == 0` branch returns `Ready` for an all-hash-skipped pass **by design** (#868), and `defer_embed::run_embed_catch_up` marks `Ready` on `Ok((0, N))`. Neither consults `store.len()`. `search_capabilities` then contains `"vector"`, so `service/server/search.rs` stops down-shifting the query to the lexical lane — and every query reaches `embed_query`, whose failure is a hard `?` at `core/indexer/search/mod.rs:194` → `500 internal search error`.

### Correction to the issue's implicit assumption

**An empty HNSW index does not itself error.** I verified this before writing any fix: `usearch::Index::search` over 0 vectors returns `Ok([])`, not `Err`. The 500 does **not** come from searching the empty index — it comes from the query-embed step that honest readiness reporting would have skipped entirely. That is why the honesty half is load-bearing rather than cosmetic: with the stage reporting truthfully, the same broken index serves BM25 results (degraded but working) instead of 500ing.

## The #1711 guard

**Not weakened, and not made permissive.** The guard still refuses the overwrite; the on-disk bytes are never written on this path. What changes is what happens *after* the refusal:

- Refusing the write protects the **data**. Adopting the snapshot restores the **service**. Both are required, and they are not in tension — an empty in-memory index has nothing to contribute, so the on-disk snapshot is strictly authoritative.
- `adopt_on_disk_snapshot` **never writes to disk**. It only moves in-memory state in one direction: towards what is already durable. Worst case it declines (`Ok(false)`) and the store is left exactly as it was.
- Validation is **not reimplemented**. The candidate is loaded through `UsearchStore::load_from`, so a truncated (#2922), zero-vector-vs-populated, or torn binary/sidecar (#3970) snapshot is rejected by exactly the code that rejects it at warm-boot.
- The **#1717 shrink refusal deliberately does not recover this way.** A partially-populated in-memory index may hold vectors the on-disk snapshot lacks, so adopting disk there would *cause* data loss. Distinguishing the two refusals is why `save()`'s internal `bool` became `SaveVerdict`.

## Changes

| File | Change |
|---|---|
| `core/store/usearch_recover.rs` *(new)* | `SaveVerdict` + `adopt_on_disk_snapshot`. New file so `usearch_store.rs` stays under the 500-SLOC cap. |
| `core/store/usearch_store.rs` | `save()` returns `SaveVerdict`; adopts the snapshot on the #1711 refusal. |
| `core/indexer/persist_hnsw.rs` | `vector_count()` — live-store ground truth. |
| `service/reindex/validate.rs` | `semantic_stage_broken_reason` — pure predicate. |
| `service/reindex/stages.rs` | `semantic_health_reason` gatherer; gate applied in `mark_semantic_ready_graph_in_progress`. |
| `service/reindex/defer_embed.rs` | Same gate in the deferred-embed Ready branch. |

The honesty gate is deliberately narrow: it fires only when an embedder is wired **and** the corpus is non-empty **and** the live store reports exactly zero. A *partially* populated store is not flagged — widening it to a ratio would re-introduce the heuristic-threshold problem already documented on `SHRINK_GUARD_RATIO_DIVISOR`.

## Could the failure path be genuinely tested?

**Yes — both halves, with the real broken state constructed, not mocked.**

- `test_save_refusal_adopts_populated_snapshot` builds a **real** populated snapshot (900 vectors, verified over the 100 KB guard threshold by an explicit fixture precondition), then points a fresh **empty** store at that path and saves — the exact partial warm-boot. It asserts all three of: the on-disk bytes are byte-identical (the #1711 contract holds), the store recovered its vectors, and a real search returns the original chunk ids (proving the key sidecar was rehydrated, not just the graph).
- `semantic_stage_reports_failed_when_live_store_is_empty` builds a handle with a working embedder, a populated corpus, and an empty vector store, then drives the real `mark_semantic_ready_graph_in_progress` and asserts `Failed` + that `search_capabilities` no longer advertises `vector`.
- Unrecoverable and healthy paths are covered too: `test_adopt_declines_when_snapshot_is_unrecoverable`, `test_adopt_declines_on_dim_mismatch`, `semantic_stage_still_ready_when_store_is_populated` (the #868 no-change reindex must stay `Ready`).

Note the pre-existing #1711 test writes a byte-*filler* file, which can prove refusal but never recovery — there is no loadable snapshot to recover from. It still passes unchanged.

## Mutation table

Every mutation was reverted from a file backup (`/tmp/4707-backup`), md5-verified — never `git checkout <file>`.

| # | Mutation | Expected red | Result |
|---|---|---|---|
| M1 | `save()`'s #1711 arm never calls `adopt_on_disk_snapshot` (pre-fix behaviour) | `test_save_refusal_adopts_populated_snapshot` | **RED** — and both `test_adopt_declines_*` stayed green, proving they don't depend on recovery succeeding |
| M2 | Fast-pass honesty gate bypassed (`if let Some(reason) = None`) | `semantic_stage_reports_failed_when_live_store_is_empty` | **RED**; `..._still_ready_when_store_is_populated` green |
| M3 | `semantic_stage_broken_reason` always returns `None` | predicate test + stage test | **RED** (2 failed); the four "must not fire" tests stayed green |
| M4 | `adopt_on_disk_snapshot` dim-mismatch guard removed | `test_adopt_declines_on_dim_mismatch` | **RED**; adoption + unrecoverable tests green |

Two fixture preconditions also fired for real during development rather than passing silently: the snapshot-size floor (600 vectors → 96,112 bytes, under the 100 KB guard threshold — bumped to 900) and the store-state assertion that caught `commit_parsed_batch` clearing a pre-seeded vector, which would have made the "healthy" fixture silently identical to the broken one.

## Gates (crate-scoped, raw)

```
$ cargo check -p trusty-search
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 16.09s

CLIPPY_EXIT=0

$ cargo fmt --check
cargo fmt --check EXIT=0

line-cap: measured 3604 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ cargo test -p trusty-search --lib
test result: FAILED. 1375 passed; 6 failed; 1 ignored; 0 measured; 0 filtered out; finished in 66.66s

failures:
    service::watch_loop::tests::cdk_out_file_is_not_indexed
    service::watch_loop::tests::modified_docx_file_triggers_indexing_via_extractor
    service::watch_loop::tests::modified_file_triggers_indexing
    service::watcher::tests::modified_event_emitted_within_one_second
    service::watcher::tests::removed_event_emitted_on_delete
    service::watcher_manager::tests::save_triggers_incremental_index_via_manager
```

Those 6 are the **known-environmental filesystem-watcher flakes named in `CLAUDE.md`** — not caused by this branch. Evidence: `git diff --name-only origin/main...HEAD | grep watch` returns nothing (the diff touches zero watcher files), and they fail identically in isolation under `--test-threads=1`, i.e. FSEvents is not delivering on this host. Change-specific gates pass.

Every other test target is green:

```
     Running unittests src/main.rs
test result: ok. 325 passed; 0 failed; 3 ignored
     Running tests/integration_tests.rs
test result: ok. 7 passed; 0 failed; 0 ignored
     Running tests/ooc_quick_wins.rs
     Running tests/warm_boot_corpus_open_failure.rs
test result: ok. 2 passed; 0 failed; 0 ignored
     Running tests/corpus_open_quarantine_4122.rs
     Running tests/residency_cold_park.rs
test result: ok. 2 passed; 0 failed; 0 ignored
     Running tests/typeahead.rs
test result: ok. 7 passed; 0 failed; 1 ignored
```

## Not done, deliberately

- **Nothing was installed, and no daemon was restarted** — per the task constraint. This PR is not deployed anywhere.
- **The live reproduction index `tm-duetto-agents-skills-01` was not modified.** It was read only (status endpoint, `errors.jsonl`, on-disk file sizes). No reindex, no repair, no write. Note it was independently reindexed at `2026-08-04T03:15Z` by something outside this session, so the live 500 no longer reproduces; the diagnosis rests on the `errors.jsonl` records and the code paths, not on live re-observation. Its `index_status` still shows the honesty defect verbatim: `semantic: { embedded: 0, total: 1732, status: "ready" }`.
- **Version bump** left alone pending CI's `check-pr-version-bump.sh` verdict.

Closes #4707

### DIFF
--- a/crates/trusty-search/changelog.d/4707-vector-warmboot-recovery.md
+++ b/crates/trusty-search/changelog.d/4707-vector-warmboot-recovery.md
@@ -1,0 +1,29 @@
+Fixed
+
+- An index whose vector layer warm-booted empty is no longer left permanently
+  broken while still self-reporting `ready`. Two defects combined: the vector
+  layer never recovered, and the status surface never admitted it
+  ([#4707](https://github.com/bobmatnyc/trusty-tools/issues/4707)).
+  - **Recovery.** When warm-boot's `UsearchStore::load_from` discards a snapshot
+    (the `#2922` size floor, a corrupt sidecar, the `#3970` torn-pairing guard)
+    the store falls back to a fresh empty one. Every later save of that empty
+    store was then correctly refused by the `#1711` data-loss guard — and
+    nothing further happened, so the index served zero vectors forever with an
+    intact snapshot sitting on disk. After refusing the write, `save()` now
+    adopts that on-disk snapshot, so the vector lane recovers without a reindex.
+    The `#1711` guard is unchanged and nothing is ever written on that path;
+    adoption only moves in-memory state towards what is already durable, and
+    reuses `load_from`, so a truncated or torn snapshot is rejected by exactly
+    the code that rejects it at warm-boot. The `#1717` shrink refusal
+    deliberately does not recover this way — a partial in-memory index may hold
+    vectors disk does not.
+  - **Honest health.** The semantic stage is no longer published as `ready`
+    when the live vector store holds zero vectors, a corpus exists, and an
+    embedder is wired. An all-hash-skipped incremental reindex legitimately
+    embeds nothing (`#868`), and both the fast pass and the deferred-embed pass
+    marked `ready` on that basis without ever consulting the store they were
+    vouching for. The stage now reports `failed` with an actionable reason, so
+    `search_capabilities` stops advertising `vector` and the search handler
+    keeps down-shifting queries to the working lexical lane instead of routing
+    them through a query-embed step whose failure surfaced as
+    `500 internal search error` on every query.
--- a/crates/trusty-search/src/core/indexer/persist_hnsw.rs
+++ b/crates/trusty-search/src/core/indexer/persist_hnsw.rs
@@ -83,9 +83,10 @@ impl CodeIndexer {
     /// incremental reindex was marked `semantic: ready` while holding zero
     /// vectors. This accessor is the ground truth those markers must check.
     /// What: one read-lock + `Index::size()` on the wired store.
-    /// Test: `service::reindex::stages::tests::semantic_health_reason_*` drive
-    /// the predicate this feeds; the accessor itself is exercised by every
-    /// reindex integration test.
+    /// Test: `semantic_stage_reports_failed_when_live_store_is_empty` and
+    /// `semantic_stage_still_ready_when_store_is_populated` (in
+    /// `service::reindex::stages::tests`) drive the gate this feeds; the
+    /// accessor itself is exercised by every reindex integration test.
     pub async fn vector_count(&self) -> Option<usize> {
         let store = self.store.as_ref()?;
         store.len().await.ok()
--- a/crates/trusty-search/src/core/indexer/persist_hnsw.rs
+++ b/crates/trusty-search/src/core/indexer/persist_hnsw.rs
@@ -72,6 +72,25 @@ impl CodeIndexer {
         self.store = Some(store);
     }
 
+    /// Number of vectors the LIVE vector store currently holds, or `None` when
+    /// no store is wired (BM25-only / test indexer).
+    ///
+    /// Why (issue #4707): every existing semantic-readiness signal is
+    /// bookkeeping — a stage counter written at warm-boot or at the end of a
+    /// reindex — and none of them is ever reconciled against the store that
+    /// actually answers queries. An index whose warm-boot fell back to an
+    /// empty store (`hnsw_load_failed`) and then saw an all-hash-skipped
+    /// incremental reindex was marked `semantic: ready` while holding zero
+    /// vectors. This accessor is the ground truth those markers must check.
+    /// What: one read-lock + `Index::size()` on the wired store.
+    /// Test: `service::reindex::stages::tests::semantic_health_reason_*` drive
+    /// the predicate this feeds; the accessor itself is exercised by every
+    /// reindex integration test.
+    pub async fn vector_count(&self) -> Option<usize> {
+        let store = self.store.as_ref()?;
+        store.len().await.ok()
+    }
+
     /// Compose the BM25 document text for a chunk: body + virtual_terms,
     /// matching the layout the per-query rebuild used to construct.
     pub(crate) fn bm25_doc_text(chunk: &RawChunk) -> String {
--- a/crates/trusty-search/src/core/store/mod.rs
+++ b/crates/trusty-search/src/core/store/mod.rs
@@ -12,6 +12,9 @@ pub(crate) mod path_match;
 mod tests;
 mod types;
 mod usearch_impl;
+// Issue #4707: snapshot-adoption recovery for the #1711 guard. Kept in its own
+// file so `usearch_store.rs` stays under the 500-SLOC production cap.
+mod usearch_recover;
 mod usearch_store;
 
 pub use self::types::{VectorHit, VectorStore};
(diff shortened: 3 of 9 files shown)


bobmatnyc:
## CI: green

Final head `7b071b6bf749e851cede10839892907fd797fa9b` — **26 pass, 2 skipping, 0 failing.**

The first push was red on one check, legitimately: `Test: doc-comment pointer lint` caught two `Test:` citations naming a `semantic_health_reason_*` prefix that no test actually uses. Fixed in `7b071b6b` by citing the real test names. That is the same stale-pointer anti-pattern the lint exists to catch, and it was a real defect in my docs, not a flake.

Notable green checks: `Test` (15m35s, full workspace suite), `Clippy` (9m6s), `trusty-search daemon smoke test` (5m49s), `Test: doc-comment pointer lint` (4m48s), `Per-PR changelog fragment`, `500-line file-size cap`, `PR version bump vs crates.io` (no bump required), `MSRV check (1.91)`.

The 6 filesystem-watcher failures reported in the PR body are **local-only** — the CI `Test` job is fully green, which confirms they are this host's FSEvents environment and not the branch.

PR #4745, opened 2026-08-04
fix(trusty-common): capitalised identifier segments are not credentials

### LINKED ISSUE #4739: Secret detector mixed-case branch false-positives on dotted filenames with a capital letter
Split out of #4312 on 2026-08-04. PR #4723 fixes that issue's backtick, charset, and entropy cases; this one survives it because the root cause is a **different branch of the same detector**.

**Reproduction** — from #4312's own issue body, Case 1:

```
Agents.app.bak-20260729-000028   ->  rejected as  Agen…(30 chars)
```

The redacted preview matches the issue character for character, verified against PR #4723's head.

**Root cause:** the **mixed-case branch**, not the base64 branch #4723 addresses. The leading capital `A` breaks `is_uniform_word_segment`, so a perfectly ordinary dotted filename with a timestamp suffix falls through to the credential heuristic.

**Why it is filed separately rather than folded into #4723:** #4723's charset gate and entropy floor are both narrowings of the base64 branch. Neither touches the mixed-case path, and stretching that PR to cover it would mean changing two independent code paths under one review. #4723 is being re-tagged `Refs #4312` accordingly.

**This is the fifth recurrence in this detector's history** — #1667 (structural-token allowlist), #2800 and #4216 (issue-number-list allowlist), #4312 (backtick delimiter, charset gate, entropy floor), and now the mixed-case branch. Every prior fix narrowed one branch. Worth considering, when this is scoped, whether the mixed-case branch needs the same credential-plausibility treatment #2442 gave it and #4723 gave the base64 branch — or whether the branches should share one predicate rather than drifting independently.

**Suggested regression test:** the reproduction above as a must-not-flag case, plus a real mixed-case credential asserted still-blocked, so the fix cannot be a blanket exemption for anything containing a capital letter.

bobmatnyc:
Closes #4739

## The defect

`Agents.app.bak-20260729-000028` — an ordinary macOS app bundle backup — was rejected as `Agen…(30 chars)`. The reproduction from #4312's issue body, Case 1, still live on `fd9d913c`.

It carries no `+`, `/` or `=`, so `has_b64_sym` is false and the base64 branch that PR #4723 narrowed **never runs**. It reaches the mixed-case branch instead, and `is_structural_token`'s segmented-identifier branch declines to rescue it:

```
Agents.app.bak-20260729-000028
  split('-')  ->  ["Agents.app.bak", "20260729", "000028"]
  is_uniform_word_segment("Agents.app.bak")  ->  false   (leading capital A)
```

The predicate admitted only all-lowercase and ALL-UPPERCASE segments, so a leading capital — the single most ordinary thing about an English word or an app bundle name — read as internal mixed case.

## The scoping question the issue asks, answered

> should the mixed-case branch get the same credential-plausibility treatment #2442 gave it and PR #4723 gave the base64 branch, or should the branches share one predicate rather than drifting independently?

**Neither, exactly — because the branches already share a predicate, and this defect is in it.** `is_structural_token` is checked once, before both branches. What I found tracing it is that its three internal branches partition cleanly by delimiter:

| Branch | Fires on | Serves |
|---|---|---|
| (a) `key=value` | token contains `=` | base64 branch |
| (b) slash-path | token contains `/` | base64 branch |
| (c) segmented identifier | reached only after `+`, `=`, `/` are **all** ruled out | mixed-case branch, **exclusively** |

Branch (c) is reachable only when `has_b64_sym` is false by construction. So the shared predicate already has a part dedicated to each branch, and the fix belongs in part (c) — where it fixes the class without any possibility of loosening base64 detection.

Giving the mixed-case branch a second branch-local charset gate would have been recurrence #6 waiting to happen. It already has one (`is_plausible_credential_charset`, #2442), and that gate is *passing* here — a dotted filename is built from exactly the credential charset. Charset was never the discriminator; **word structure** is.

## The fix — two changes, both in the shared predicate

1. **`IDENTIFIER_DELIMITERS = ['-', '_', '.']`.** `is_segmented_identifier` split on `-` alone, so every dot- and underscore-joined name arrived as one unsplittable segment. `.` and `_` delimit identifier words exactly as `-` does.
2. **`is_uniform_word_segment` → `is_human_word_segment`**, accepting `Capitalized` alongside `lowercase` and `UPPERCASE`.

## Direction of the trade, and why the bound is stated as a bound

`is_structural_token` returning `true` makes `looks_like_secret` return `false`. **Widening it narrows flagging and can only lose detection, never add it.** PR #4723 shipped a doc asserting the opposite direction for a neighbouring predicate and a measurement disproved it, so this is measured rather than argued.

The real bound `is_human_word_segment` holds, stated as what it is: **a segment that is not uniformly uppercase admits at most ONE uppercase letter, and only as its first letter.** That is what keeps run-of-two case alternation — `AbCd`, `EfGh`, the signature of an encoded blob — out. It is *not* a claim that no credential can be built from capitalized segments: `Abcd-1234-Efgh-5678` satisfies it. That shape is a product key, already exempt today in its conventional ALL-CAPS form, and carries no case entropy to lose.

## Measured, not asserted

Same 36-shape prose battery and 30-shape credential battery, run at both revisions in this worktree (`filter.rs` reverted to `origin/main` for the baseline so the batteries are identical):

| Corpus | `origin/main` | this PR |
|---|---|---|
| Prose false positives | **17 / 36** | **2 / 36** |
| Credential misses | **0 / 30** | **0 / 30** |

**15 of 17 false positives fixed, zero new credential misses.** Every one of the eight pins PR #4723 left behind is unmoved — the AWS secret-key baseline miss, the two lowercase URL-path known misses, and the four URL prose residue cases all read identically at both revisions.

The credential battery is not just the inherited one. It adds six shapes that specifically probe this widening — `AbCd1234-EfGh5678-IjKl9012`, its `.` and `_` variants, `xYzAbc123-qRsTuv456-mNoPqr789`, `Prod-AbCd1234EfGh5678IjKl`, `Prod.Key.AbCd1234EfGh5678` — because a careless version of this change loses exactly those and nothing else would have noticed.

## What I did NOT fix, and why

CamelCase segments (`TrustyMemory`, `MyDocument`) have two capitalised runs, so they are still not words and these two stay flagged:

```
TrustyMemory.app.bak-20260801-120000
MyDocument.Backup.2026.tar.gz
```

Fixing them means admitting multi-uppercase segments, which is the exact signature of `AbCdEfGhIjKlMnOpQrSt-1234`. That trades a credential miss for two prose cases. 15-of-17 with zero misses is the better deal, so the residue is pinned as an explicit change-detector assertion in `real_secrets_still_blocked_after_4739_capitalised_segments` and stated as a known bound on `is_human_word_segment` rather than left to be rediscovered.

## Guard mutation

Each new guard was mutated and the intended test confirmed failing:

| Mutation | Test that caught it |
|---|---|
| Drop the `Capitalized` acceptance | `dotted_capitalised_filenames_are_not_flagged` — `Agents.app.bak-20260729-000028` |
| Narrow `IDENTIFIER_DELIMITERS` back to `['-']` | `dotted_capitalised_filenames_are_not_flagged` — `INSTRUCTIONS.md.bak-20260731-000028` |
| Over-widen: accept any segment | `real_secrets_still_blocked_after_4739_capitalised_segments` (+ 2 inherited batteries) |

## Gates — Rust Test Ladder **rung 4**

```
$ cargo fmt --check
exit=0

$ cargo check -p trusty-common --features memory-core
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.30s

$ cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings
exit=0

$ cargo test -p trusty-common --features memory-core
running 713 tests
test result: ok. 700 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out; finished in 11.49s

$ cargo test -p trusty-common --features memory-core --lib -- memory_core::filter
test result: ok. 52 passed; 0 failed; 0 ignored; 0 measured; 661 filtered out
```

> `--features memory-core` is mandatory: `trusty-common`'s default feature set is empty, so a bare `cargo test -p trusty-common` filters out all ~700 memory-core tests and exits 0 while proving nothing.

**Workspace:**

```
$ cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui --exclude trusty-agents-ui
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 38.01s   (exit=0)
```

The three excluded crates fail on the pre-existing environmental Tauri gap (`frontendDist` → `ui/dist` missing). `git status --porcelain` shows zero files touched in them, and CI excludes the same three (`ci.yml:302,308`).

**Direct dependents:**

```
$ cargo test -p trusty-memory
test result: FAILED. 486 passed; 2 failed; 4 ignored; 0 measured; 0 filtered out
    commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener
    commands::doctor::tests::check_daemon_health_fails_when_no_addr_file_and_no_listener
```

The two known environment-dependent doctor failures (they depend on whether a live memory daemon is listening). `git status --porcelain -- crates/trusty-memory/` is empty — zero files touched in that crate.

```
$ cargo test -p trusty-agents
test result: FAILED. 3311 passed; 1 failed; 11 ignored; 0 measured; 0 filtered out
    assistants::tests::home_tests::for_instance_validates_the_id
```

**Pre-existing flake, proven by repetition rather than asserted.** Zero files touched in that crate. It passes in isolation and fails intermittently under parallel execution, at *both* revisions — `home_tests` run 10× each:

```
my HEAD              : 9 of 10 runs FAILED
pristine origin/main : 5 of 10 runs FAILED   (filter.rs + filter_tests.rs reverted)
```

Root cause is visible and unrelated to this diff: `for_instance_validates_the_id` (`home_tests.rs:106`) mutates the process-global `ASSISTANTS_DIR_ENV` via `EnvVarGuard::set` **without** `#[serial]`, while three sibling tests in the same file (lines 116, 135, 160) are `#[serial]` and one of them sets the same variable. The observed failure is another test's temp root leaking in:

```
left:  ".../trusty-agents-knowledge-e868720e-.../trusty-agents/izzie"
right: ".../.tmpuxe4m1/izzie"
```

Reporting it rather than touching it — it belongs to `trusty-agents`, not to this PR's blast radius, and a sibling engineer is active in that crate.

**Repo gates:**

```
bash scripts/check_line_cap.sh           → 3619 files; 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh → OK trusty-common: fragment present and valid
bash scripts/check-pr-version-bump.sh    → OK trusty-common 0.28.1 — src changed, version not yet published
```

**Version parity:** in-tree `0.28.1`, latest published `0.28.0`. No bump needed, verified against the registry rather than assumed.

Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d from a crate suite, or `--lib`-narrowed.

### DIFF
--- a/crates/trusty-common/changelog.d/4739-mixed-case-branch.md
+++ b/crates/trusty-common/changelog.d/4739-mixed-case-branch.md
@@ -1,0 +1,7 @@
+Fixed
+
+- Secret detector no longer flags dotted or underscored filenames that carry a capital letter (closes [#4739](https://github.com/bobmatnyc/trusty-tools/issues/4739))
+  - `Agents.app.bak-20260729-000028` and shapes like it reached the mixed-case branch, which #4723's base64-branch narrowing never covered
+  - `is_structural_token`'s segmented-identifier branch now splits on `.` and `_` as well as `-`, and accepts a `Capitalized` segment alongside `lowercase` and `UPPERCASE`
+  - Measured over a 36-shape prose battery and a 30-shape credential battery: prose false positives 17 → 2, credential misses 0 → 0
+  - CamelCase segments (`TrustyMemory.app.bak-…`) are a stated known bound, not fixed: admitting them would lose delimiter-segmented mixed-case credentials
--- a/crates/trusty-common/src/memory_core/filter.rs
+++ b/crates/trusty-common/src/memory_core/filter.rs
@@ -559,11 +559,36 @@ const IDENTIFIER_DELIMITERS: [char; 3] = ['-', '_', '.'];
 /// measurement disproved: a segment that is not uniformly uppercase admits **at
 /// most one uppercase letter, and only as its first letter**. That is what
 /// keeps the run-of-two case alternation of an encoded blob (`AbCd`, `EfGh`)
-/// out — two uppercase letters in a non-uppercase segment is disqualifying. It
-/// is NOT a claim that no credential can be built from capitalized segments:
-/// `Abcd-1234-Efgh-5678` satisfies it. That shape is a product key, already
-/// exempt in its conventional ALL-CAPS form, and carries no case entropy to
-/// lose.
+/// out — two uppercase letters in a non-uppercase segment is disqualifying.
+///
+/// The ACCEPTED side of that bound, stated because the rejected side alone
+/// would read as a stronger guarantee than this predicate gives: what
+/// [`is_segmented_identifier`] admits is a token whose **every segment is
+/// independently case-uniform**. At short segment lengths that includes shapes
+/// which are not human words at all —
+/// `Xy7-Kp2-Qm9-Rt4-Vw8-Nz3-Bc6-Fj0` and `A1b2-C3d4-E5f6-G7h8-I9j0-K1l2` are
+/// both admitted, and neither is a filename. They are a real, measured miss,
+/// not a hypothetical.
+///
+/// Why that is an acceptable price rather than a hole: a *generated* credential
+/// essentially never satisfies per-segment case uniformity, because it is not
+/// segmented into short groups in the first place, and if it is, every group
+/// must independently land in one of three case shapes. Treating a short
+/// alphanumeric segment as case-uniform with probability roughly ¼ (an
+/// estimate, assuming uniform draws from a mixed alphabet — not a measured
+/// constant), an eight-segment token clears the bar about `(1/4)^8` of the
+/// time, order 1 in 10^4–10^5. Everything **generated** rather than
+/// **composed** — bare blobs, base64, base64url, JWTs, provider-prefixed keys —
+/// carries no such segmentation and still flags. The exposure is to a
+/// credential a human deliberately composed to look like an identifier, which
+/// is the accidental-storage hygiene threat model this module has always had,
+/// not an adversarial one.
+///
+/// The earlier revision of this doc illustrated the accepted side with
+/// `Abcd-1234-Efgh-5678`. That example was wrong: at 19 characters it is below
+/// `looks_like_secret`'s own 20-char floor, so it never reaches this branch and
+/// could not demonstrate anything about it. `Abcd-1234-Efgh-5678-Ijkl` (24) is
+/// the corrected form and is genuinely admitted here.
 ///
 /// Known bound, deliberately not fixed: CamelCase (`TrustyMemory`,
 /// `MyDocument`) has two capitalized runs and is still not a word here, so
@@ -578,7 +603,8 @@ const IDENTIFIER_DELIMITERS: [char; 3] = ['-', '_', '.'];
 /// predicate constrains case, not charset.
 /// Test: `structural_tokens_are_not_flagged`,
 /// `dotted_capitalised_filenames_are_not_flagged`,
-/// `real_secrets_still_blocked_after_4739_capitalised_segments`.
+/// `real_secrets_still_blocked_after_4739_capitalised_segments`,
+/// `per_segment_case_uniformity_is_a_known_accepted_bound`.
 fn is_human_word_segment(seg: &str) -> bool {
     let s = seg.trim_matches(|c: char| !c.is_ascii_alphanumeric());
     if s.is_empty() {
--- a/crates/trusty-common/src/memory_core/filter.rs
+++ b/crates/trusty-common/src/memory_core/filter.rs
@@ -521,55 +521,115 @@ const SECRET_PREFIXES: &[&str] = &[
     "asia",
 ];
 
-/// True when `seg` is a "uniform-case word segment": non-empty, consists only
-/// of ASCII alphanumeric chars (no punctuation), and every alphabetic char is
-/// either all-lowercase or all-uppercase (no internal mixed case within the
-/// segment). Digits count as case-neutral.
+/// Characters that separate the words of a human-readable compound identifier.
+///
+/// Why (issue #4739): [`is_segmented_identifier`] split on `-` alone, so an
+/// ordinary dotted filename (`Agents.app.bak-20260729-000028`) was segmented
+/// into `["Agents.app.bak", "20260729", "000028"]` — and the first segment,
+/// carrying two case runs across its dots, could not be a word. `.` and `_`
+/// delimit identifier words exactly as `-` does; treating them as delimiters is
+/// what lets the per-segment predicate see words instead of a blob.
+///
+/// Why this does not widen the base64 branch: [`is_structural_token`] returns
+/// before reaching its segmented-identifier branch whenever the token contains
+/// `+`, `=`, or `/` — which is precisely the `has_b64_sym` set. The segmented
+/// branch is therefore reachable only for tokens the base64 branch can never
+/// fire on, so this delimiter set is confined to the mixed-case branch.
+/// What: the delimiter set `-`, `_`, `.` used by both `contains` and `split`.
+/// Test: `dotted_capitalised_filenames_are_not_flagged`.
+const IDENTIFIER_DELIMITERS: [char; 3] = ['-', '_', '.'];
+
+/// True when `seg` reads as a single human-readable word: all-lowercase,
+/// ALL-UPPERCASE, or Capitalized. Digits are case-neutral, so a digit-only
+/// segment (`20260729`) qualifies.
 ///
 /// Why: this is the per-segment predicate used by [`is_segmented_identifier`]
-/// to determine whether a hyphen-delimited token reads like a compound
-/// human-readable identifier (`2-medium->REQUEST_CHANGES`) vs. a random
-/// mixed-case credential blob (`AbCd1234-EfGh5678`).
-/// What: strips leading/trailing non-alnum chars (from arrow punctuation like
-/// `>`, `<`), then checks all-lowercase-or-digit or all-uppercase-or-digit.
-/// Test: `structural_tokens_are_not_flagged`.
-fn is_uniform_word_segment(seg: &str) -> bool {
+/// to tell a compound human-readable identifier (`2-medium->REQUEST_CHANGES`,
+/// `Agents.app.bak-20260729-000028`) from a random mixed-case credential blob
+/// (`AbCd1234-EfGh5678`).
+///
+/// Why Capitalized was added (issue #4739, the FIFTH recurrence in this
+/// detector after #1667, #2800, #4216 and #4312): admitting only all-lower and
+/// all-upper meant a leading capital — the single most ordinary thing about an
+/// English word or a macOS app bundle name — read as internal mixed case, so
+/// `Agents.app.bak-20260729-000028` fell through to the credential heuristic.
+///
+/// The bound this predicate actually holds, stated exactly because a reader
+/// will rely on it and because #4723 shipped an absolute here that a
+/// measurement disproved: a segment that is not uniformly uppercase admits **at
+/// most one uppercase letter, and only as its first letter**. That is what
+/// keeps the run-of-two case alternation of an encoded blob (`AbCd`, `EfGh`)
+/// out — two uppercase letters in a non-uppercase segment is disqualifying. It
+/// is NOT a claim that no credential can be built from capitalized segments:
+/// `Abcd-1234-Efgh-5678` satisfies it. That shape is a product key, already
+/// exempt in its conventional ALL-CAPS form, and carries no case entropy to
+/// lose.
+///
+/// Known bound, deliberately not fixed: CamelCase (`TrustyMemory`,
+/// `MyDocument`) has two capitalized runs and is still not a word here, so
+/// `TrustyMemory.app.bak-20260801-120000` remains a false positive. Admitting
+/// it would require accepting multi-uppercase segments, which is the exact
+/// signature of `AbCdEfGhIjKlMnOpQrSt-1234` — a credential miss. Measured, not
+/// assumed; see the residue assertion in the test below.
+/// What: strips leading/trailing non-alnum chars (arrow punctuation like `>`,
+/// `<`), then accepts when the alphabetic characters are all-lowercase,
+/// all-uppercase, or a single leading uppercase followed by lowercase. Interior
+/// non-alphanumeric characters are ignored rather than rejected — this
+/// predicate constrains case, not charset.
+/// Test: `structural_tokens_are_not_flagged`,
+/// `dotted_capitalised_filenames_are_not_flagged`,
+/// `real_secrets_still_blocked_after_4739_capitalised_segments`.
+fn is_human_word_segment(seg: &str) -> bool {
     let s = seg.trim_matches(|c: char| !c.is_ascii_alphanumeric());
     if s.is_empty() {
         return false;
     }
-    // All alphabetic chars in segment must share a single case.
-    let all_lower = s
-        .chars()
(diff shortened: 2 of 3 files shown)


bobmatnyc:
## Verdict: APPROVE

Independent adversarial review, run at `e62f6fff` in a dedicated read-only worktree (`.claude/worktrees/review-4745`, detached). Everything below was measured at **both** revisions — PR head, and PR head with `filter.rs` reverted to `fd9d913c` so the probe battery is byte-identical across the two.

---

### 1. The branch-partition claim holds

Verified from the code, not the prose. `is_structural_token` is called once, before both branches of `looks_like_secret`, and its three internal branches are total on the delimiter set:

- `+` → `return false` at `filter.rs:668`.
- `=` → `if token.contains('=')` at `filter.rs:715`. `splitn(2, '=')` on a token containing `=` always yields exactly 2 parts, so `parts.len() == 2` is unconditionally true and **every** path inside that block returns (`:721`, `:724`, `:729`). Branch (a) cannot fall through.
- `/` → `return token.split('/').all(is_word_segment)` at `filter.rs:735`.

Branch (c) at `filter.rs:740` is therefore reachable only when the token contains none of `+`, `=`, `/` — the exact complement of `has_b64_sym` (`filter.rs:796`). The partition is real, and a change confined to (c) cannot loosen base64 detection.

Confirmed empirically as well: every `+`/`=`/`/`-bearing probe (JWT padded and url-safe, base64 basic-auth, the `wJalrXUtnFEMI/K7MDENG/…` AWS shape, all four URL residue tokens, the upper-case webhook must-flag) reads **identically** at both revisions.

### 2. The charset-gate claim holds

`is_plausible_credential_charset` (`filter.rs:1000`) admits `[A-Za-z0-9]` plus `-`, `_`, `.`, `:`. `Agents.app.bak-20260729-000028` is built entirely from that set, so the #2442 gate was **passing** on the reproduction. Charset was not the discriminator; a second charset gate would not have closed this. Correct diagnosis.

### 3. The stated invariant is what the code enforces

Doc claim: *a segment that is not uniformly uppercase admits at most one uppercase letter, and only as its first letter.*

Traced through `is_human_word_segment` (`filter.rs:562-580`):

- `first` uppercase + `rest_all_upper` → uniformly uppercase (the exempt case that already existed).
- `first` uppercase + `rest_all_lower` → exactly one uppercase, at the first letter.
- `first` lowercase → `rest_all_lower` required → zero uppercase.

There is no fourth path. The invariant is exact — unlike #4723's. `AbCd`/`EfGh` run-of-two alternation is rejected on the second uppercase, which is the property the whole safety argument rests on.

### 4. Differential — 50-shape adversarial battery, both revisions

I did not reuse your batteries. I built shapes designed to defeat this specific widening: 2/3/4/5-char Capitalized segments, mixed delimiters in one token, digit-interleaved capitals, leading-digit segments, plus 25 real credential formats.

**Every realistic credential family still flags at HEAD:**

| Family | HEAD |
|---|---|
| JWT (padded and url-safe `-`/`_` segments) | FLAG |
| SendGrid `SG.Abc123DefGhi.Xyz789QrsTuvWxy012` | FLAG |
| GitHub PAT, PyPI token, `sk-ant-…` | FLAG |
| AWS secret key (no delimiter) | FLAG |
| base62 random `Xk9Fj2Qm7Pd4Zr3Ns8Bv6Tw1Hy5Lc0` | FLAG |
| base64 basic-auth blob | FLAG |
| `AbCd1234-EfGh5678-IjKl9012` and `.`/`_` variants | FLAG |
| `Prod.Key.AbCd1234EfGh5678` | FLAG |

JWTs deserve a specific note since `.` is now a delimiter: they are safe **structurally**, not by luck. A JWT header base64-encodes `{"alg"…`, so it always begins `eyJ` — lowercase first alpha followed by an uppercase — which `is_human_word_segment` rejects on its second letter. Splitting on `.` does not open that door.

**All eight #4723/#4312 pins are unmoved.** AWS `wJalrXUtnFEMI/K7MDENG/…` baseline miss: pass at both. Both all-lowercase URL-path known misses: pass at both. All four URL prose residue tokens: FLAG at both. The upper-case webhook must-flag: FLAG at both. `cargo test -p trusty-common --features memory-core` is green, which is where those pins live as assertions.

### 5. The one thing the battery does not pin — the finding below

Seven shapes move FLAG → pass that are neither prose nor in your credential battery. They share one structure: **every** segment is uniformly Capitalized (or ALL-CAPS), at short segment lengths.

```
Xy7-Kp2-Qm9-Rt4-Vw8-Nz3-Bc6-Fj0      main: FLAG   PR: pass
A1b2-C3d4-E5f6-G7h8-I9j0-K1l2        main: FLAG   PR: pass
Abcd-EFGH-Ijkl-MNOP-Qrst-UVWX-1234   main: FLAG   PR: pass
Xy7k.Kp2m.Qm9r.Rt4v.Vw8n.Nz3b        main: FLAG   PR: pass
Xy7-Kp2_Qm9.Rt4-Vw8_Nz3.Bc6          main: FLAG   PR: pass   (mixed delimiters in one token)
12Ab-34Cd-56Ef-78Gh-90Ij-12Kl        main: FLAG   PR: pass
Abcdefghijklmno-Pqrstuvwxyz1234      main: FLAG   PR: pass
```

Your battery probes the **rejected** side of the bound (run-of-two alternation). It does not probe the **accepted** side at token scale. See the finding table.

I do not think this blocks. A randomly generated credential essentially never satisfies per-segment case uniformity — for the `Xy7-Kp2-…` shape that is roughly `(1/4)^8`. Every family that is generated rather than composed still flags. But the bound as documented under-states the exempted set, and this detector's history is exactly the history of an under-stated bound.

Two classes I checked and am **not** flagging:

- Passphrases (`Correct.Horse.Battery.Staple1`, `Db_Password_Hunter2_Correct_Horse`) move FLAG → pass. Measured: their all-lowercase and ALL-CAPS forms already passed on `origin/main`. This extends an existing accepted hole in kind, it does not open a new one.
- Mixed-case UUID `550e8400-E29b-41d4-A716-446655440000` moves FLAG → pass. Both canonical forms (all-lower per RFC 4122, all-upper) already passed on main; mixed-case UUIDs do not occur.

### 6. Guard mutation — all three reproduced

| Mutation | Result |
|---|---|
| `is_human_word_segment` → `true` | 3 tests fail, incl. `real_secrets_still_blocked_after_4739_capitalised_segments` |
| `IDENTIFIER_DELIMITERS` → `['-']` | `dotted_capitalised_filenames_are_not_flagged` fails on `INSTRUCTIONS.md.bak-20260731-000028` |
| Drop `Capitalized` acceptance | `dotted_capitalised_filenames_are_not_flagged` fails on the verbatim repro |

Exactly the tests and exactly the tokens your table names.

### 7. The CamelCase refusal is correct

Admitting `TrustyMemory` requires admitting multiple capitalized runs. The obvious safe-looking narrowing — "each capitalized run needs ≥3 lowercase letters" — admits `TrustyMemory` but also admits `AbcdEfghIjklMnop-1234`, a 4-char-run blob that is an entirely ordinary encoded shape. English word length and blob chunk length overlap; there is no clean separator there. Trading a credential miss for two prose cases is the wrong direction, and the residue is pinned as a change-detector rather than left to be rediscovered. I looked for a safe form and did not find one. Do not fix this.

### 8. Scope

Production delta is 15 added / 10 removed non-comment lines, confined to `IDENTIFIER_DELIMITERS`, `is_human_word_segment`, and the three call sites in `is_segmented_identifier`. Nothing else in `filter.rs` changed. Three files total.

### 9. Gates re-run independently

```
cargo fmt --check                                  exit=0
cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings   exit=0
  running 713 tests
  test result: ok. 700 passed; 0 failed; 13 ignored
  test result: ok. 6 passed; 0 failed
  test result: ok. 11 passed; 0 failed
  test result: ok. 4 passed; 0 failed
bash scripts/check_line_cap.sh            3619 files; 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh                 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh  OK trusty-common
gh pr checks 4745                         all 27 required checks pass
```

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-common/src/memory_core/filter.rs` | 549-557 | The stated bound gives only `Abcd-1234-Efgh-5678` as a credential that survives the widening, framed as "a product key … carries no case entropy to lose". Measured, the exempted set is larger and does carry case entropy: `Xy7-Kp2-Qm9-Rt4-Vw8-Nz3-Bc6-Fj0`, `A1b2-C3d4-E5f6-G7h8-I9j0-K1l2`, `Abcd-EFGH-Ijkl-MNOP-Qrst-UVWX-1234`, `Xy7-Kp2_Qm9.Rt4-Vw8_Nz3.Bc6` all move FLAG → pass. A reader relying on this doc will under-estimate what branch (c) now exempts. | Extend the bound to name the real shape: *any* token whose every segment is independently Capitalized or ALL-CAPS is exempt regardless of token-level entropy, at any segment length. Give `Xy7-Kp2-Qm9-Rt4-Vw8-Nz3-Bc6-Fj0` as the example instead of the product key. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/filter_tests.rs` | 1567-1608 | The credential battery probes only the rejected side of the bound (`AbCd1234-…` run-of-two). No pin exists for the class that actually moved, so a future widening of segment length or delimiter handling would extend this hole silently. This is the same gap #4312 closed for itself with an explicit KNOWN-MISS block. | Add a KNOWN-MISS pin in `real_secrets_still_blocked_after_4739_capitalised_segments`, modelled on the #4312 block: `assert!(find_secret_token("Xy7-Kp2-Qm9-Rt4-Vw8-Nz3-Bc6-Fj0").is_none(), "KNOWN MISS (#4739): …")` plus the `A1b2-C3d4-…` variant. | Fix here |
| LOW | `crates/trusty-common/src/memory_core/filter.rs` | 556 | The doc's bound example `Abcd-1234-Efgh-5678` is 19 chars, below `looks_like_secret`'s own 20-char floor (`filter.rs:784`) — it can never reach the branch it is illustrating. | Use a ≥20-char example. Subsumed by the MEDIUM above if that fix lands. | Fix here |
| LOW | `crates/trusty-common/src/memory_core/filter.rs` | 566 | `let alphas: Vec<char> = …collect()` allocates once per segment, per token, on the memory-ingest filter path. `Filter<Chars, _>` is `Clone`, so the two `all()` passes can run without the allocation. | `let mut alphas = s.chars().filter(char::is_ascii_alphabetic); let Some(first) = alphas.next() else { return true; };` then `alphas.clone().all(…) \|\| alphas.all(…)`. Not a hot path — cosmetic. | Promote |

## Notes

Zero CRITICAL, zero HIGH → **APPROVE**. Neither MEDIUM changes the shipped behavior; both are about making the bound legible and pinned so recurrence #6 is visible rather than silent. If you land only one, land the test pin — the doc follows from it.

The direction argument is sound and, more importantly, measured rather than argued. Widening `is_structural_token` narrows flagging; the question is only what got narrowed, and the answer across 50 shapes is: fifteen prose false positives, plus a family of composed-rather-than-generated tokens that no credential generator emits. I would ship this filter in a public release.

Reviewed with the head detached in `.claude/worktrees/review-4745`; no edits, no commits, nothing merged. Mutations were applied and reverted in that worktree only, `git status --porcelain` clean after each.

Issue #4735, opened 2026-08-04
Extract a shared tri-state git probe helper; 91 sites each call Command::new("git") independently


bobmatnyc:
The 2026-08-04 audit counted **91 production git spawn sites across 61 files and 7 crates**, with **no shared entry point** — confirming the CLAUDE.md domain-audit table's FRAGMENTED verdict with concrete evidence.

Seven of those sites fail open, filed separately. The root cause is uniform: `!status.success()` conflates a definitive negative with a total failure to determine, and **git exits 128 for every fatal condition**.

🔴 **Do NOT promote the nearest existing candidate as-is.** `crates/trusty-common/src/catchup/git.rs` has the same collapse — `.output().ok().filter(|o| o.status.success())` erases spawn errors and non-zero exits into one branch. It is currently harmless because its consumers are human-readable report text, and its doc comment says the fail-open is intentional there. Promoting it unchanged would spread the bug.

**Three correct tri-state implementations already exist and were never exported past their own crate:** `trusty-mpm/session_manager/worktree_safety.rs::git_command` ("every error path resolves to DIRTY"), `trusty-agents/workflow/resolver.rs::decide_merge_bytes` (explicit `Some(0)` / `Some(1..=127)` / `Failed`, built as the fix for #3652), and `trusty-mpm/core/project_aliases.rs`. The shape is proven; it is just not packaged for reuse.

Proposed:
```rust
pub enum GitProbe<T> {
    Answer(T),        // meaningful result, including a meaningful non-zero exit (diff --quiet's "dirty")
    NotApplicable,    // definitively negative — not a repo, not tracked, not ignored
    Indeterminate,    // spawn error, or a non-zero exit that isn't the expected negative signal
}
```
Migrate the seven confirmed sites individually, mapping `Indeterminate` to each site's conservative branch.

**Note on the deferral:** CLAUDE.md records this consolidation as "backlog (after tmux pattern proves)". That deferral now has a measured cost — one of the seven is a documented, unfixed data-loss bug with two live production call paths.

**Fifth finding (known-deliberate, not filed):** `trusty-code/src/binding/mod.rs:294-313` (`is_git_worktree`), conflates the same two states but the module documents `false` as "a fully supported bound state, never an error" — a deliberate design choice, LOW impact (capability downgrade only). It is a known-deliberate instance so a future migration doesn't "fix" it by accident.

[...]

PR #4746, opened 2026-08-04
fix(trusty-mpm): never delete a worktree git refused to delete (#4732)

### LINKED ISSUE #4732: remove_session_worktree deletes worktrees git refused to delete, including locked ones
Found 2026-08-04 auditing all 91 production git spawn sites for fail-open handling. **This is a data-loss path already documented in the codebase and never fixed.**

`crates/trusty-mpm/src/session_manager/decommission.rs:129-226` (`remove_session_worktree`): when `git worktree remove --force <path>` exits non-zero, it falls back **unconditionally** to `std::fs::remove_dir_all(path)` — a raw recursive delete that bypasses git entirely.

Git exits non-zero for many reasons, and the worst one is deliberate: **`git worktree lock` makes git exit 128 specifically to protect the worktree.** So locking a worktree to protect it is exactly what causes this code to delete it. Dubious ownership and a broken worktree pointer produce the same outcome.

**The codebase already says so.** `crates/trusty-mpm/src/session_manager/worktree_reclaim_sweep.rs:222-230` states verbatim that git honors the lock but this function "treats that as a git failure and falls back to `remove_dir_all`, which deletes the directory anyway." A `git_still_permits` re-check was added as a workaround — **but only to the separate `--merged-prs` reclaim path.** It was never applied to `remove_session_worktree` itself.

**Reachable from everyday commands**, not an edge path: `SessionManager::decommission()` (used by `tm session decommission`, `tm sessions prune --state stopped`, and the age-based reaper) and `prune_orphaned_worktrees()`.

**Fix:** on a non-zero exit, distinguish "git declined and the worktree must be preserved" from "git could not run at all." A lock, dubious ownership, or any unrecognized failure must REFUSE, never fall back to `remove_dir_all`. Extend `git_still_permits` here, or better, route through the shared tri-state helper proposed in the companion issue below.

Related: #4728 (a separate shipped data-loss path in the same module), and the 2026-07-21 incident where an unreviewed cleanup destroyed a bare clone.

bobmatnyc:
Closes #4732

## The defect

`session_manager::decommission::remove_session_worktree` fell through to `std::fs::remove_dir_all` on **any** non-zero `git worktree remove --force` exit. Git exits 128 for every fatal condition, and the worst one is deliberate: `git worktree lock` is the operator's only mechanism for saying "leave this alone", and git enforces it by exiting 128. **Locking a worktree to protect it was exactly what caused it to be deleted.**

Reachable from `tm session decommission`, `tm sessions prune --state stopped`, the age-based reaper, `prune_orphaned_worktrees()`, and the `--merged-prs` reclaim pass.

### It was wider than the reported case — two fail-open sites, not one

The second one was never reported and deletes uncommitted work without ever reaching the `git worktree remove` call:

`registry_root_for(path) == None` was read as "no git repository owns this path — removing the directory directly". That `None` is **also** what a stale worktree pointer answers. Reproduced against git 2.54.0: with `.git/worktrees/<name>` removed out of band, `git -C <wt> rev-parse --git-common-dir` says `fatal: not a git repository: (null)` while the working tree and its uncommitted files are entirely intact. That is the state ~70 worktrees on this machine were left in on 2026-07-21.

### The pre-fix behaviour was pinned by a test

`worktree_reclaim_sweep_tests::remove_session_worktree_reports_success_when_the_fallback_removes_it` locked a real worktree and asserted the fallback had deleted it. It was written for #2919 to settle a different question (what the return value means when the fallback succeeds), which is why this defect survived a review round that read that file. It is inverted here, with the return-value question it was actually settling preserved.

## The fix

New `session_manager::worktree_protection` — a tri-state classifier, never a bool:

| state | meaning | action |
|---|---|---|
| `Protected(reason)` | git holds state here, or declined to give it up | refuse |
| `Unclaimed` | positively established: no git state at or claiming this path | permit |
| `Undetermined(reason)` | git could not be consulted, or answered unreadably | refuse |

`GitProtection::refusal()` collapses both refusing states into one accessor, so a caller cannot match `Protected` alone and silently let `Undetermined` through.

**Exit code alone never decides, and neither does the message alone.** Measured against git 2.54.0 — all three exit 128:

| condition | stderr |
|---|---|
| operator-locked worktree | `cannot remove a locked working tree; use 'remove -f -f' to override or unlock first` |
| broken `.git` file | `validation failed, cannot remove working tree: '…/.git' is not a .git file, error code 7` |
| plain dir / already-pruned worktree | `'<path>' is not a working tree` |

Only the third means git holds nothing, and the second reads a hair away from it while meaning the opposite. So the match is narrow (`is not a working tree`, anything else refuses) and then **corroborated**: a `symlink_metadata` witness on `<path>/.git` — which needs only the parent's search bit, so it answers for a dangling, unreadable, or mode-000 pointer where every `git` call reports "not a repository" — and then git's own `worktree list --porcelain` registry, canonicalised on both sides. This is deliberately the shape `trusty-agents-common`'s `agents::vcs_claim` landed on in #4727.

### The surviving fallback, stated exactly

`remove_dir_all` now runs for **one** case, in a function named for it (`remove_unclaimed_directory`):

> `path` has passed the trusty-mpm ownership gate, and git has POSITIVELY established that it holds no state there — the path carries no `.git` entry, and either no repository exists above it at all, or the owning repository's `git worktree list` does not name it.

That is a leftover directory: a worktree git already pruned, or one whose creation never completed registration. There is no ref to prune and nothing for git to protect. Both halves have a passing test, so this fix cannot silently become "never clean anything up".

### The reason reaches the caller

`remove_session_worktree` now returns `WorktreeRemoval::{Removed, Kept(reason)}`. `ReclaimOutcome::removal_failed` becomes `Vec<String>` of `"<path>: <reason>"` (same JSON shape — an array of strings — so `tm session prune-worktrees --merged-prs` output is enriched, not broken). "removal failed" with no reason read as a transient error; the commonest reason is now a deliberate refusal.

### Scope

#4735's shared `GitProbe<T>` is **not** built here. `worktree_protection.rs` is one self-contained file with no trusty-mpm-specific types in its signatures, so that extraction is a move rather than a rewrite.

## Tests

Every test below fails against the pre-fix remover.

| test | shape |
|---|---|
| `remove_session_worktree_refuses_a_git_locked_worktree` | real `git worktree lock`; asserts the dir AND a file inside it survive |
| `remove_refuses_a_stale_worktree_pointer` | admin dir removed out of band; uncommitted file must survive |
| `remove_refuses_an_unreadable_git_entry` | `chmod 000` on `.git`, restored by a `Drop` guard |
| `remove_refuses_a_worktree_with_a_broken_git_file` | `.git` naming a nonexistent gitdir |
| `remove_cleans_up_a_directory_no_repository_claims` | the surviving fallback, half one |
| `remove_cleans_up_an_unregistered_leftover_inside_a_repo` | the surviving fallback, half two |
| `remove_still_removes_a_healthy_worktree` | happy path incl. registry prune + `branch -D` |

Plus 15 classifier unit tests, all driven against **real git** in a `TempDir` — a mocked git would test the mock, and this defect was a misreading of what real git actually says.

### Mutation testing — and what it caught

Each new guard was mutated in turn and the suite re-run. **Two of the twelve initially SURVIVED**: gates 1 (message) and 2 (`.git` witness) in `protection_after_failed_removal` were masking each other, because both tests used a healthy registered worktree that a later gate could also refuse. Each was rewritten to target a path every *other* gate permits — an unregistered plain directory for gate 1, a stale-pointer worktree for gate 2 — with the precondition asserted in the test body and a mutation note in its doc. After that, all twelve are killed:

```
KILLED  M1  gate1: accept any stderr as 'nothing here'      by an_unrecognized_git_message_is_protected
KILLED  M2  gate2 (post-removal): ignore the .git witness   by a_dot_git_entry_overrides_gits_not_a_working_tree_message
KILLED  M3  registry unreadable reads as 'not registered'   by an_unreadable_registry_is_undetermined
KILLED  M4  uncanonicalisable path reads as 'not registered' by an_uncanonicalisable_path_is_undetermined
KILLED  M5  a registered worktree reads as unclaimed        by a_registered_worktree_is_protected_even_when_git_says_otherwise
KILLED  M6  the lock is not named in the reason             by the_registry_reports_a_lock_as_a_lock
KILLED  M7  gate1 (no-root): ignore the .git witness        by remove_refuses_a_stale_worktree_pointer, a_stale_worktree_pointer_is_protected
KILLED  M8  ancestor witness always says 'no repository'    by an_unresolvable_ancestor_repo_is_undetermined
KILLED  M9  unresolvable ancestor repo reads as unclaimed   by an_unresolvable_ancestor_repo_is_undetermined
KILLED  M10 refusal() lets Undetermined through             by refusal_covers_both_non_unclaimed_states
KILLED  M11 remover ignores the no-root verdict             by remove_refuses_a_stale_worktree_pointer, remove_refuses_an_unreadable_git_entry, remove_refuses_a_worktree_with_a_broken_git_file
KILLED  M12 remover ignores the failed-removal verdict      by remove_session_worktree_refuses_a_git_locked_worktree
```

One branch has no test and is called out here rather than hidden: the `Err` arm where `git` cannot be spawned at all. It now refuses instead of falling back, but forcing it needs a `PATH` shim.

## Gates — Rust Test Ladder rung 5

```
cargo fmt --check                                      exit 0
cargo check -p trusty-mpm                              exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings exit 0
cargo test -p trusty-mpm                               4332 passed, 0 failed, 7 ignored (lib)
                                                       + all integration suites green, exit 0
bash scripts/check_line_cap.sh        3621 files, 0 violations — OK
bash scripts/check_sld.sh             0 errors, 0 warnings
bash scripts/check_test_pointers.sh   20687 citations, 0 dangling — OK
bash scripts/check_changelog_fragment.sh  OK trusty-mpm: fragment present and valid
```

`cargo test -p trusty-mpm -- --include-ignored`: **4338 passed, 1 failed** — `daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many`, an `#[ignore]`d benchmark panicking on `deploy_all_skill_tiers(...).unwrap()` with `Io(NotFound)`.

Not this branch. Proven, not asserted:
- `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/daemon/ crates/trusty-mpm/src/core/` → **0 files**; the diff is confined to `session_manager/`.
- The same run at origin/main state (this branch's `session_manager/` reverted to `origin/main`, new files removed) fails **identically**, same test, same panic site `staleness_bench_tests.rs:115`, `4317 passed; 1 failed`.
- The test passes in isolation at both origin/main and HEAD, so it is a full-run isolation flake, not a regression.

No canonical issue exists for it (`gh issue list --state all --search bench_stale_assets_for_many` and `--search staleness_bench_tests` both return empty). Change-specific gates pass; this one is a pre-existing full-run flake.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4732-locked-worktree-deletion.md
+++ b/crates/trusty-mpm/changelog.d/4732-locked-worktree-deletion.md
@@ -1,0 +1,6 @@
+Fixed
+
+- Worktree teardown no longer deletes worktrees git refused to delete (closes [#4732](https://github.com/bobmatnyc/trusty-tools/issues/4732))
+  - `remove_session_worktree` fell through to `std::fs::remove_dir_all` on ANY non-zero `git worktree remove --force` exit. Git exits 128 for every fatal condition, so `git worktree lock` — the operator's only "do not remove this" — was exactly what caused deletion. A stale worktree pointer, an unreadable `.git`, and a repository git merely declined to read all produced the same outcome, with uncommitted work in the worktree.
+  - Every git failure is now classified into three states — git holds state here / git positively holds nothing here / git could not be asked — and only the middle one permits a raw removal. Locked, stale-pointer, broken-`.git`, and unrecognized-message cases are refused. Reachable from `tm session decommission`, `tm sessions prune --state stopped`, the age-based reaper, `prune_orphaned_worktrees`, and the `--merged-prs` reclaim pass.
+  - The refusal reason is now returned to the caller instead of buried in a log line: `tm session prune-worktrees --merged-prs` reports `removal_failed` entries as `"<path>: <reason>"`.
--- a/crates/trusty-mpm/src/session_manager/decommission.rs
+++ b/crates/trusty-mpm/src/session_manager/decommission.rs
@@ -22,6 +22,7 @@ use super::manager::{ManagedError, SessionManager};
 use super::record::{ManagedSessionId, ManagedSessionState, SessionRecord};
 use super::search_gc;
 use super::workspace_guard::is_safe_to_remove;
+use super::worktree_protection;
 use super::worktree_safety::{DirtyWorktree, inspect_dirt};
 
 /// Sentinel file written by [`create_session_worktree`] into every SM-created
@@ -98,6 +99,42 @@ pub(crate) fn is_session_worktree(path: &Path) -> bool {
         .unwrap_or(false)
 }
 
+/// What happened to a worktree the removal path was asked to delete (#4732).
+///
+/// Why: the caller has to be able to tell "gone" from "still on disk, and here
+/// is why" — and it has to be told the reason, not left to find it in a `warn!`
+/// nobody reads. The previous `bool` could express neither: `false` conflated a
+/// deliberate refusal, an I/O error, and a timeout, and the reason existed only
+/// as a log line inside the function.
+/// What: [`Removed`](Self::Removed) — the directory is gone;
+/// [`Kept`](Self::Kept) — it is still on disk, carrying an operator-facing
+/// reason.
+/// Test: `remove_session_worktree_refuses_a_git_locked_worktree`,
+/// `remove_cleans_up_a_directory_no_repository_claims`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum WorktreeRemoval {
+    /// The directory is gone — git removed it, it was already absent, or an
+    /// unclaimed trusty-mpm-owned directory was removed directly.
+    Removed,
+    /// The directory is still on disk. The string says why, for the operator.
+    Kept(String),
+}
+
+impl WorktreeRemoval {
+    /// `true` only when the directory is gone.
+    pub(super) fn removed(&self) -> bool {
+        matches!(self, Self::Removed)
+    }
+
+    /// The operator-facing reason the directory was kept, if it was.
+    pub(super) fn reason(&self) -> Option<&str> {
+        match self {
+            Self::Removed => None,
+            Self::Kept(reason) => Some(reason),
+        }
+    }
+}
+
 /// Remove an in-project per-session git worktree via `git worktree remove --force`.
 ///
 /// Why (#1840): `remove_dir_all` alone leaves the git ref (`session/<id>`) and
@@ -120,17 +157,49 @@ pub(crate) fn is_session_worktree(path: &Path) -> bool {
 /// convention. Branch deletion is best-effort — "not found" is silently
 /// ignored since older sessions may not have a branch. OsStr-safe path args
 /// avoid lossy UTF-8 coercion (#1840).
-/// Idempotent: if `path` is already absent, returns `true` (already removed).
-/// On git failure, best-effort falls back to `remove_dir_all` and logs WARN.
-/// Returns `true` when the workspace was removed (either via git or fallback)
-/// or was already absent; `false` only when all removal attempts failed.
-/// Test: `is_session_worktree_absent_path_is_noop`; integration coverage via
-/// the decommission round-trip tests that set up real git worktrees.
-pub(super) fn remove_session_worktree(path: &Path) -> bool {
+/// Idempotent: if `path` is already absent, returns
+/// [`Removed`](WorktreeRemoval::Removed).
+///
+/// # 🔴 Git declining is a REFUSAL, not a failure (#4732)
+///
+/// This function used to fall through to `std::fs::remove_dir_all` on ANY
+/// non-zero exit, and on any failure to resolve the owning checkout. Git exits
+/// 128 for every fatal condition, so that fall-through deleted the exact things
+/// git was protecting — most sharply, `git worktree lock`, whose only mechanism
+/// for saying "leave this alone" is a 128 exit. Locking a worktree to protect
+/// it was what got it deleted. A stale worktree pointer and an unreadable
+/// `.git` produced the same outcome while their working trees, uncommitted work
+/// included, were entirely intact.
+///
+/// Every git failure is now classified by
+/// [`super::worktree_protection`], which answers three states and refuses on
+/// two of them. The `remove_dir_all` fallback survives for EXACTLY ONE case:
+///
+/// > `path` has passed the trusty-mpm ownership gate above, and git has
+/// > POSITIVELY established that it holds no state there — the path carries no
+/// > `.git` entry, and either no repository exists above it at all, or the
+/// > owning repository's `git worktree list` does not name it.
+///
+/// That is a leftover directory: a worktree git already pruned, or one whose
+/// creation never completed registration. There is no ref to prune and nothing
+/// for git to protect, so a direct removal is the only way to clean it up.
+/// Every other outcome — including "git could not be asked" — keeps the
+/// directory and returns [`Kept`](WorktreeRemoval::Kept) with the reason.
+///
+/// Test: `is_session_worktree_absent_path_is_noop`,
+/// `remove_session_worktree_refuses_a_git_locked_worktree`,
+/// `remove_refuses_a_stale_worktree_pointer`,
+/// `remove_refuses_an_unreadable_git_entry`,
+/// `remove_refuses_a_worktree_with_a_broken_git_file`,
+/// `remove_cleans_up_a_directory_no_repository_claims`,
+/// `remove_cleans_up_an_unregistered_leftover_inside_a_repo`,
+/// `remove_still_removes_a_healthy_worktree`; integration coverage via the
+/// decommission round-trip tests that set up real git worktrees.
+pub(super) fn remove_session_worktree(path: &Path) -> WorktreeRemoval {
     if !path.exists() {
         // Already gone — either removed by a concurrent decommission or by a
         // previous partial run. Treat as success (idempotent removal).
-        return true;
+        return WorktreeRemoval::Removed;
     }
 
     // Data-safety gate (#1845 item 5): prefer the SM ownership sentinel over the
@@ -150,7 +219,10 @@ pub(super) fn remove_session_worktree(path: &Path) -> bool {
                 "decommission: refusing worktree removal — no SM ownership sentinel \
                  and path is not under .worktrees/; skipping conservatively"
             );
-            return false;
+            return WorktreeRemoval::Kept(format!(
+                "no trusty-mpm ownership sentinel ({WORKTREE_SENTINEL_FILE}) and the path \
+                 is not under {WORKTREES_DIRNAME}/"
+            ));
         }
         warn!(
             path = %path.display(),
@@ -168,19 +240,26 @@ pub(super) fn remove_session_worktree(path: &Path) -> bool {
     let repo_root = match super::worktree_registry::registry_root_for(path) {
         Some(r) => r,
         None => {
-            // No git registry claims this path — it is a plain directory, not a
-            // worktree. The sentinel gate above already established trusty-mpm
-            // owns it, so remove the directory directly rather than refusing.
+            // #4732: a `None` here is NOT "no git repository owns this path".
+            // `registry_root_for` also returns `None` when git could not
+            // resolve one — which is exactly what a worktree with a stale or
+            // unreadable `.git` pointer answers, while its working tree and
+            // uncommitted work are entirely intact. Classify before deleting.
+            let verdict = worktree_protection::protection_without_registry_root(path);
+            if let Some(reason) = verdict.refusal() {
+                warn!(
+                    path = %path.display(),
+                    "decommission: refusing worktree removal — {reason} (#4732)"
+                );
+                return WorktreeRemoval::Kept(reason.to_string());
+            }
+            // The one surviving fallback case — see this function's doc.
             warn!(
                 path = %path.display(),
-                "decommission: no git repository owns this path — removing the \
+                "decommission: no git repository claims this path — removing the \
                  directory directly (no worktree ref to prune)"
             );
-            if let Err(e) = std::fs::remove_dir_all(path) {
-                warn!(path = %path.display(), "decommission: remove_dir_all failed: {e}");
-                return false;
-            }
-            return true;
+            return remove_unclaimed_directory(path);
         }
     };
     let repo_root = repo_root.as_path();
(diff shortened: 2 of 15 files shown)


bobmatnyc:
## Verdict: APPROVE

Adversarial review, independent worktree (`.claude/worktrees/qa-4732`, detached at `c3078f75`, removed after the run). Every claim below was executed, not read.

### 🔴 Can anything still delete a git-protected directory? No.

**Every path to a destructive call in the changed code:**

| call | site | reachable only when |
|---|---|---|
| `remove_dir_all` | `decommission.rs:330` (`remove_unclaimed_directory`) | sentinel/`.worktrees/` ownership gate passed **and** a `GitProtection::Unclaimed` verdict. Two call sites, `decommission.rs:262` and `:303`, both immediately behind `verdict.refusal()` |
| `remove_dir_all` | `decommission.rs:762` | the **mutually exclusive** `workspace_owned == true` arm (`:654` vs `:735`) — an SM-created clone, not a worktree, containment-gated by `is_safe_to_remove`. Unchanged by this PR and unreachable from the worktree path |
| `git branch -D` | `decommission.rs:384` | inside `prune_refs_after_removal`, called **only** from the `o.status.success()` arm (`:277`). A refusal never reaches it |
| `remove_file` / `rename` | — | none in the changed code |

No git failure can produce `Unclaimed`. Gate 1 refuses on any stderr that is not the exact phrase; gate 2 refuses on a `.git` witness; gate 3 refuses on an unreadable registry, an uncanonicalisable path, or a registered/locked entry. The `Err` spawn arm (`:305`) refuses unconditionally.

**`remove_unclaimed_directory`'s stated case is the only case that reaches it** — verified by mutation, below. Routing of each hazard, measured against git 2.54.0:

| condition | routes to | verdict |
|---|---|---|
| operator-locked | gate 1 (message unrecognised); gate 2 also holds | `Protected` |
| stale pointer (`.git/worktrees/<n>` gone) | `protection_without_registry_root` gate 1 | `Protected` |
| unreadable `.git` (mode 000) | `protection_without_registry_root` gate 1 | `Protected` |
| `.git` file naming a dead gitdir | `protection_without_registry_root` gate 1 | `Protected` |
| symlinked `.git` | `symlink_metadata` does not follow → gate 1 | `Protected` |
| `git` binary missing | `Err` arm, or gate 3 `Undetermined` | `Kept` |
| **bare repo at `path`** | git says `is a main working tree`; registry names the bare root | `Protected` ×2 |
| **nested full clone at `path`** | git says `is a main working tree`; `.git` dir present | `Protected` ×2 |

The bare-clone case matters — that is what the 2026-07-21 incident destroyed — and it is now doubly closed.

### The narrow match cannot be satisfied by a protecting condition

Measured, git 2.54.0, byte-level substring test against `"is not a working tree"`:

```
locked      → fatal: cannot remove a locked working tree;
              use 'remove -f -f' to override or unlock first     no match ✓
broken .git → fatal: validation failed, cannot remove working tree:
              '…/.git' is not a .git file, error code 7          no match ✓
main tree   → fatal: '<path>' is a main working tree             no match ✓
plain dir   → fatal: '<path>' is not a working tree              MATCH (correct)
```

The broken-`.git` near-miss reads `is not a .git file` and `cannot remove working tree` — neither contains the needle. Wording drift fails **closed** (`if !stderr.contains(…) → Protected`), proven by mutation M1.

`symlink_metadata` witness, porcelain-registry corroboration, and canonicalisation-first are all present and correct — `registry_verdict` canonicalises both sides before comparing and returns `Undetermined` when the target cannot be canonicalised; `protection_without_registry_root` canonicalises before walking ancestors.

### The second fail-open site — reproduced independently

Confirmed against git 2.54.0. With `.git/worktrees/<name>` removed out of band:

```
$ git -C <wt> rev-parse --path-format=absolute --git-common-dir
fatal: not a git repository: (null)                       exit 128
$ ls <wt>/          f  precious.txt      ← working tree fully intact
$ ls -la <wt>/.git  -rw-r--r-- … 222 B   ← pointer still present
```

`registry_root_for` returns `None`, and pre-fix `decommission.rs` took that `None` straight to `remove_dir_all` without ever invoking `git worktree remove`. **The reported P0 was half the defect.** The finding is real and correctly attributed.

### 🔴 The test that pinned the bug

Confirmed at `origin/main`, `worktree_reclaim_sweep_tests.rs:684-702`:

```rust
fx.lock_worktree(&path);
let reported = …::remove_session_worktree(&path);
assert!(!path.exists(), "precondition: the fallback really did remove the directory");
assert!(reported, …);
```

It locked a real worktree and asserted the deletion as a **precondition**. It was written for #2919 to settle a return-value question, and it is why a prior review that read this file did not catch the defect.

**The inversion is faithful.** #2919's question was which side of `remove_session_worktree(..) && !path.exists()` a given outcome lands on. The old test pinned "gone ⇒ reports removed"; the new one pins "survives ⇒ reports NOT removed" (`!outcome.removed()`), the other side of the same invariant, and the original direction is still covered by `remove_still_removes_a_healthy_worktree` and both `remove_cleans_up_*` tests, which assert `outcome.removed()`. No coverage was lost in the flip.

**Sibling sweep — no second enshrining test.** All `lock_worktree` call sites (`sweep_tests:94,693`, `protection_tests:55,207`, `registry_tests:427,512`) are refusal-side. All three tests that assert a deletion are legitimate: `remove_still_removes_a_healthy_worktree` (registered, healthy), `reclaim_remove_mode_reclaims_a_clean_merged_worktree` (clean, merged), `prune_orphaned_worktrees_force_discards_dirty` (explicit `ForceDiscard` opt-in, and dirt-only — the git gate still applies to it).

### Mutation testing — re-run, including the two that initially survived

Applied to the PR tree in my own worktree, full `session_manager::` suite (470 tests) after each:

```
baseline          470 passed; 0 failed
M1  gate 1 removed  (accept any stderr)   → FAILED 1: an_unrecognized_git_message_is_protected           KILLED
M2  gate 2 removed  (ignore .git witness) → FAILED 1: a_dot_git_entry_overrides_gits_not_a_working_tree… KILLED
M10 refusal() lets Undetermined through   → FAILED 1: refusal_covers_both_non_unclaimed_states           KILLED
M11 remover ignores no-root verdict       → FAILED 3: remove_refuses_a_stale_worktree_pointer,
                                                      remove_refuses_a_worktree_with_a_broken_git_file,
                                                      remove_refuses_an_unreadable_git_entry             KILLED
M12 remover ignores failed-removal verdict→ FAILED 1: remove_session_worktree_refuses_a_git_locked_worktree KILLED
```

Both initially-surviving mutations reproduce as killed, each by exactly the retargeted test named for it — and by nothing else, which is what proves the retargeting worked rather than merely moved the masking. **No mirror problem**: `an_unrecognized_git_message_is_protected` asserts `registry_verdict(..) == Unclaimed` in its body and its subject carries no `.git`, so gates 2 and 3 both permit and only gate 1 can refuse; `a_dot_git_entry_overrides…` is fed the permissive stderr and asserts the registry no longer names the path, so only gate 2 can refuse. M12 being killed by the inverted test is the cleanest possible evidence that the inversion is load-bearing.

### Gates (independently re-run, not taken from the PR body)

```
cargo fmt --check                                         exit 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings    exit 0
cargo test -p trusty-mpm                                   4332 passed; 0 failed; 7 ignored
bash scripts/check_line_cap.sh        3621 files, 0 violations — OK
bash scripts/check_sld.sh             3041 code files, 0 errors, 0 warnings
bash scripts/check_test_pointers.sh   20687 citations, 0 dangling — OK
bash scripts/check_changelog_fragment.sh  OK trusty-mpm: fragment present and valid
```

One first-run flake, disclosed for completeness: `core::bundled_pm_package::tests::resolve_pm_prompt_wrapper_matches_the_source_reporting_form` failed on my first full run and passed both in isolation and on the immediate re-run (4332/0). The diff touches **zero** files outside `session_manager/` (`git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/ | grep -v session_manager/` → empty), so it cannot be caused by this branch. Not a finding, and distinct from the `staleness_bench_tests` one already ruled pre-existing.

## Findings

| MEDIUM | `session_manager/decommission.rs` | 718–724 | The `Kept` reason reaches the operator on only one of the three surfaces. `reclaim_with_probes` puts it in `removal_failed` as `"<path>: <reason>"` (→ JSON → CLI ✓), but `decommission_with_root` collapses the outcome to `workspace_removed: bool` and logs the reason with `warn!`. `prune.rs:620` then maps it to `DecommissionedWorktreeRetained` + `retained_workspace_path` — the **same** report a dirty worktree produces. An operator running `tm sessions prune --state stopped` cannot tell "unlock it" from "commit your work"; `prune.rs:975` is `warn!`-only too. Strictly better than the pre-fix silence, but "the reason reaches the caller" holds for one of three paths. | Add a `retained_reason: Option<String>` to `PrunedSession` (and the `OrphanSweepOutcome` row) and thread `outcome.reason()` into it, mirroring what `removal_failed` already does. | Parent |
| LOW | `session_manager/worktree_protection.rs` | 100–105 | `refusal()` collapsing `Protected`/`Undetermined` is a convention supported by an accessor, not an enforced invariant — the variants are visible throughout the module tree, so a future caller can still write `matches!(v, Protected(_))`. M10 is killed only by the enum's own unit test; no behavioural test drives an `Undetermined` verdict through the remover, because every end-to-end hazard resolves to `Protected`. | Either make the variants private behind constructors + `refusal()`, or add one end-to-end test that forces `Undetermined` (e.g. the unresolvable-ancestor shape) through `remove_session_worktree`. | Parent |
| LOW | `session_manager/worktree_protection.rs` | 32–34 | The #4735 "plain file move" claim is precisely worded and true of the *signatures* (only `&Path`, `&str`, and its own `GitProtection`), but the module imports `super::worktree_registry::{list_registered_worktrees, registry_root_for}` and reads `RegisteredWorktree`'s `path`/`locked` fields. Extraction will need those two probes moved or injected — not a rewrite, but not a pure `git mv` either. | Note the two-function dependency in the scope comment so #4735 sizes it correctly. | Parent |
| LOW | `session_manager/decommission.rs` | 351, 412 | `prune_refs_after_removal` retains a redundant bare `{ … }` block wrapping its whole body — a leftover from the `if git_success {` it was extracted out of. | Drop the block and outdent. | Parent |
| LOW | `session_manager/decommission.rs` | 305–315 | The `Err` git-spawn-failure arm has no test, as the author discloses. Behaviour is unconditional (`Kept`) so risk is minimal, but it is the one branch of the new safety contract with no coverage. | A `PATH`-shim test, or leave as disclosed. | Parent |

Zero CRITICAL, zero HIGH.

## Notes

- **Would I run this unattended against 166 orphaned worktrees with live workstreams? Yes.** Every hazard I could construct routes to a refusal, including two the PR did not claim (bare repo, nested full clone), and the sentinel gate still sits in front of all of it.
- `protection_without_registry_root`'s ancestor witness (`:246-249`) checks for a `.git` entry, which a **bare** repository does not have — so its doc's "positively outside any repository" is marginally stronger than what the code proves. I could not make it matter: `registry_root_for` succeeds for a path inside a bare repo (verified — returns the bare root, exit 0), so that classifier is not reached, and a path with no `.git` of its own cannot be a registered worktree. Precision nit in the doc, not a defect.
- `decommission.rs:762` (the `workspace_owned == true` `remove_dir_all`) is unchanged and out of this PR's scope. It is guarded by `is_safe_to_remove` containment rather than by a git verdict. Worth a separate look someday; not a regression here and not something I would hold this PR for.
- Doc comments meet the bar: full Why/What/Test on every new item, the surviving-fallback case stated as a bound in a blockquote at both `remove_session_worktree` and `remove_unclaimed_directory`, and the measured stderr table pinned to git 2.54.0 on the constant itself. The two mutation-note docs on the retargeted tests are the right call — they are the only thing stopping the next person from "simplifying" those fixtures back into mutual masking.

APPROVED for merge.

Issue #4680, opened 2026-08-03
0.39.0: every non-main index stuck at chunk_count=0 / last_walk_files_seen=0, stages frozen at lexical:in_progress — never retried across restarts


bob-duetto:
## Summary

On a live production trusty-search **0.39.0** deployment (AL2023, EFS-mounted corpora at `/mnt/data`), every index except `main` is stuck at `chunk_count=0` / `last_walk_files_seen=0`, with stages frozen at `lexical: in_progress` / `semantic: pending` / `graph: pending`. This is **not corpus-open failure** (compare #3929/#3993, which produce `stages=Failed`) — the daemon believes these indexes are still actively walking, indefinitely, and the health surface reports the deployment as fully healthy the whole time.

## Environment

- trusty-search **0.39.0**, AL2023, live production serving box
- EFS-mounted corpora at `/mnt/data`
- Service uptime ~7.6 days at time of observation (started 2026-07-24 20:35:56 UTC)
- Observed 2026-08-01, read-only

## Symptom

`/health` reports:
```
status: ok
warm_boot_degraded: false
indexes_loaded: 222/222
indexes_skipped_timeout: 0
```
— i.e. the daemon believes it is fully healthy. But only the `main` index actually holds data (315,423 chunks, all stages `ready`). Every other sampled index — `apex`, `github`, `repo-duetto`, `repo-code-intelligence`, `repo-APEX`, `mnt_data_data_repos_duetto-code-intelligence` — reports `chunk_count=0`, `last_walk_files_seen=0`, and stages frozen at `lexical: in_progress` / `semantic: pending` / `graph: pending`.

## Key point — the health surface is misleading

`indexes_loaded` counts registered index *slots*, not populated ones, and a per-index top-level `status: "ready"` likewise reflects registration only. A deployment can report fully healthy while 221 of 222 indexes serve nothing. That gap is arguably a defect in its own right, and is why this went unnoticed for 44 days.

## Source data is present and valid — this is not an empty-corpus case

Example: the `apex` index has `root_path=/mnt/data/knowledge/apex` (correct), and that directory holds 95 well-formed files, 1.7MB, newest mtime 2026-07-06, world-readable. The walk still reports 0 files seen.

## Retry loop that never converges

[...]

Issue #4391, opened 2026-07-30
Boot reconcile's git path cannot trigger a reindex for any existing git-backed index — SHA is recomputed at boot


mac-duetto:
## Summary

`reconcile_git_path` compares a stored HEAD SHA against current HEAD to decide whether an index needs reindexing. But the "stored" value is **re-derived from live git whenever an `IndexHandle` is rebuilt from `PersistedIndex`** — at eager warm-boot and at lazy restore — so for pre-restore drift it is always equal to current. Consequence: **changes committed to a repository while the daemon was down are never picked up at boot**, and the same defect silently defeats two other catch-up paths (see "Blast radius").

The only way `stored != current` can occur is a commit landing *after* the handle was built (mid-lifetime drift, or the seconds-wide window between restore and the boot reconcile task). Down-time changes predate the boot-captured SHA and can never appear in `git diff <boot_sha>..HEAD`.

## Evidence

Restore sets the handle's SHA from live git rather than from persisted state:

- `commands/start_restore.rs:300` — `let indexed_head_sha = crate::core::git::head_sha(&entry.root_path);` → applied at `:386`
- `service/lazy_restore.rs:178`, `:241` — identical

Reconcile then compares that against current HEAD (`service/reconcile.rs:316`, `:343`) → `up_to_date` for every git-backed index on every boot.

There is nothing to load instead: `indexed_head_sha` is in-memory only (`core/registry.rs:370`), `PersistedIndex` has no such field, and `git log -S indexed_head_sha -- service/persistence.rs` is empty — the field was introduced in-memory-only in `ea3f687d` (#75) and has **never been persisted at any point in history**. This is born-inert, not a regression. `finish_reindex` also stamps in-memory only (`service/reindex/finish.rs:357-361`).

## Every fallback is also dead

- **mtime fallback:** `reconcile.rs:421` reads `last_indexed_unix`, but `update_last_indexed_unix` (`service/persistence_timestamps.rs:78`) has **zero production callers** workspace-wide — definition, one re-export (`persistence.rs:351`), and tests only. The field's own doc (`persistence.rs:222`) claims it is "written at reindex SSE complete"; that writer was never wired. The field is therefore always `None` → `reconcile.rs:424-436` → `skipped_no_data`.
- **File watcher:** `notify` emits only future events; no initial scan.
- **Auto-discover** skips already-registered indexes (`commands/discover/mod.rs:155-162`).
[...]

PR #4747, opened 2026-08-04
fix(security): close three git fail-open paths that leak gitignored files (#4733)

### LINKED ISSUE #4733: Three git fail-open paths leak gitignored files into the LLM corpus, the search index, and git history
Found in the same 2026-08-04 audit. Three separate sites fall back to a non-git path when a git probe fails, and each fallback drops `.gitignore` awareness.

**1. Into the LLM review corpus** — `crates/trusty-review/src/report/scan.rs:181-199` (`git_ls_files`). On failure, falls back to a raw filesystem walk that consults only a fixed skip-dir list, not `.gitignore`. That corpus is sent to the LLM, so a `.env` or scratch credential enters it on a repo whose `git ls-files` merely hiccuped.

**2. Into the search index** — `crates/trusty-search/src/service/reconcile.rs:308-327`, driven by `core/git.rs:91-106` (`head_sha`). When `head_sha()` returns `None`, reconcile takes the mtime-walk branch meant for legitimately non-git roots, which also ignores `.gitignore`. Previously-excluded files become retrievable through the search and grep MCP tools. Note the sibling git-diff path in the same file IS fail-safe (full reindex on `None`) — the defect is specifically the `head_sha`-is-`None` branch one level up.

**3. Into git history** — `crates/trusty-mpm/src/daemon/managed_routes/inproject/untracked_sync.rs:132-140,273-287` (`append_to_git_exclude`). After copying allowlisted files (default `.env*`) into a fresh session worktree, if `git rev-parse --git-path info/exclude` fails the copied secrets are left unregistered, logged only as a `warn!`. A later `git add -A && git commit` in that worktree can stage them. The parallel `native_mcp.rs::is_env_local_actually_ignored` path does re-verify via `git check-ignore`; this one never does.

**Common fix:** a failed git probe must not silently downgrade to a less-protective mode. Preserve `.gitignore` semantics or refuse.

bobmatnyc:
Closes [#4733](https://github.com/bobmatnyc/trusty-tools/issues/4733) — milestone tm 1.3.5.

## The shape

Three sites fell back to a less-protective, `.gitignore`-blind mode when a git probe merely **failed**. A failed probe is not evidence of anything: git exits 128 for `detected dubious ownership`, a stale worktree gitlink, an unreadable `.git`, and a broken `repositoryformatversion` alike — and exits **0 printing `false`** for a bare repo. In every one of those the repository, and the `.gitignore` keeping a `.env` out, is entirely real.

Measured against git 2.54.0:

| condition | stderr | exit |
|---|---|---|
| genuinely no repo | `fatal: not a git repository (or any of the parent directories): .git` | 128 |
| **stale worktree pointer** | `fatal: not a git repository: (null)` | 128 |
| unreadable `.git` (mode 000) | `fatal: not a git repository (or any of the parent directories): .git` | 128 |
| bare repo | *(none)* — prints `false` | **0** |

The stale-pointer message contains the same short phrase while meaning the opposite, and the unreadable-`.git` case prints the *correct* phrase about a repository that is real. So each site now matches only the **parenthesised** form, corroborates it against an ancestor `.git` witness on a **canonicalised** path, and treats everything else as "could not determine" — the pattern merged in [#4727](https://github.com/bobmatnyc/trusty-tools/issues/4727) (`trusty-agents-common::agents::vcs_claim`).

## Per-site conservative branch, and why

**1. Into the LLM review corpus** — `crates/trusty-review/src/report/scan.rs`

The fallback walk consults a fixed skip-dir list and nothing else, and the result is sent to the model. **Refuse.** A `Corpus` three-state replaces `Option`: `Tracked` honours git, `NoRepo` still takes the documented non-git walk, `Refused` yields no corpus and logs a `warn!`. A review that fails loudly beats one that leaks.

Second hole closed in the same function: `if files.is_empty() { None }` treated *"git listed zero files"* as *"not a git repo"* and walked it. A repository with nothing committed yet has an empty tracked corpus and a fully live `.gitignore` — that is now `Tracked(vec![])`, not a walk trigger.

**2. Into the search index** — `crates/trusty-search/src/service/reconcile.rs`, gated by new `core::git::probe_work_tree`

`head_sha()` returning `None` dropped into the mtime catch-up walk meant for legitimately non-git roots. That walk honours `SKIP_DIRS` and the walker's skip predicates but **not** `.gitignore` (unlike the reindex walk's `respect_gitignore`), so excluded files entered the corpus and became retrievable through the `search` and `grep` MCP tools.

The issue's hint was right: the sibling git-diff path already shows the safe answer. "Index nothing" would be wrong — it would strand a legitimately-drifted index — so the conservative branch is **the gitignore-honouring full background reindex**, which is what the sibling `changed_files_between`-is-`None` path already does. Only a corroborated `NoRepo` still takes the mtime path; `Present` and `Unknown` both full-reindex.

`mtime_walk_does_not_honour_gitignore` pins the premise, so the gate cannot later be deleted as "redundant".

**3. Into git history** — `crates/trusty-mpm/.../inproject/untracked_sync.rs`

The `info/exclude` registration ran **after** the copy and a failure was only a `warn!`, so a worktree whose `git rev-parse` merely declined to answer was left holding the operator's `.env` unregistered, where a later `git add -A && git commit` stages it.

Per the issue, this reuses the existing mechanism rather than inventing a second one: the order is inverted to **register first, then re-verify each path with `git check-ignore`** (mirroring `native_mcp::is_env_local_actually_ignored`), and only what git confirms is copied.

That verification turned out to close a second hole for free — `check-ignore` reports a **tracked** path as *not* ignored no matter what `info/exclude` says, and `git add -A` stages tracked modifications regardless. So a secret can no longer overwrite a `.env` the repo already commits. Verified directly:

```
$ git check-ignore -q -- tracked.env    # listed in info/exclude, but tracked
exit=1
$ git check-ignore -q -- absent.env     # listed in info/exclude, untracked
exit=0
```

A destination with a corroborated absence of any repository has no history to leak into and is still copied to freely, so non-git destinations are unaffected.

One deliberate simplification: a failed exclude-append now warns and continues rather than returning early, because the per-file `check-ignore` gate is what actually *proves* the guarantee (registration only *makes* it true). Nothing is copied that git cannot confirm — a broken `info/exclude` makes `check-ignore` itself unanswerable (`fatal: cannot use .git/info/exclude as an exclude file`), and an unanswerable question is a refusal. `broken_exclude_file_still_refuses_unignored_paths` pins that.

Dead code removed: `append_to_git_exclude` had no production caller after the split and is gone; its idempotency test now drives `resolve_git_exclude` + `append_paths_to_exclude` directly.

## Scope

`GitProbe<T>` extraction is **not** built here — that is [#4735](https://github.com/bobmatnyc/trusty-tools/issues/4735) (1.3.6). The three classifiers are deliberately identical in shape (same const, same corroboration, same three states) so the extraction is mechanical. Each one classifies off the stderr of a git call the site **already made**, so no site pays an extra process for the fix.

## Tests — mutated, every guard

Ten mutations applied one at a time, each reverted after; every one was caught.

| # | Mutation | Test that failed |
|---|---|---|
| M1a | `classify_ls_files_failure` always concedes `NoRepo` | `scan_refuses_when_git_is_broken_rather_than_walking` + 2 |
| M1b | empty listing falls through to the walk (pre-fix) | `scan_empty_git_repo_does_not_walk_untracked_files` |
| M1c | `Refused` arm walks anyway (the pre-fix fail-open) | `scan_refuses_when_git_is_broken_rather_than_walking` |
| M1d | drop the `.git` ancestor witness | `classify_ls_files_failure_corroborates_the_no_repo_message` |
| M2a | reconcile guard removed — always mtime-walk (pre-fix) | `broken_git_repo_full_reindexes_instead_of_mtime_walking` |
| M2b | drop the `.git` ancestor witness | `classify_probe_failure_corroborates_the_no_repo_message` |
| M2c | match the SHORT `not a git repository` phrase | `classify_probe_failure_corroborates_the_no_repo_message` |
| M3a | `Unknown` proceeds to copy instead of refusing (pre-fix) | `broken_repo_copies_nothing` |
| M3c | drop the per-file `check-ignore` verification | `tracked_secret_is_not_overwritten_in_worktree`, `broken_exclude_file_still_refuses_unignored_paths` |
| M3d/M3e | short phrase / drop the witness | `classify_rev_parse_failure_corroborates_the_no_repo_message` |

**M2c initially did NOT fail** — in `probe_work_tree_is_unknown_for_a_stale_worktree_pointer` the gitlink *is* the `.git` witness, so the witness alone refused even with a too-broad phrase match. Fixed by asserting the wording against a directory with no witness at all; re-mutated and confirmed red. Every `.env`-shaped fixture is a real gitignored `.env` asserted absent from the corpus / index input / worktree.

All fixtures are temp dirs. The review scanner and the indexer were never pointed at a real project directory.

## Gates — Rust Test Ladder rung 5 (cross-crate, security)

```
bash scripts/check_line_cap.sh    — 3619 files, 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh         — 54 specs + 3044 code files, 0 errors, 0 warnings
bash scripts/check_test_pointers.sh — 20681 citations, 0 dangling — OK
bash scripts/check_changelog_fragment.sh — all 3 crates recorded

cargo clippy -p trusty-review --all-targets -- -D warnings   — clean
cargo clippy -p trusty-search --all-targets -- -D warnings   — clean
cargo clippy -p trusty-mpm    --all-targets -- -D warnings   — clean

cargo test -p trusty-review                  — 1564 passed, 0 failed (+9 suites green)
cargo test -p trusty-review -- --include-ignored — 1569 passed, 0 failed, 0 ignored
cargo test -p trusty-mpm                     — 4316 passed, 0 failed (+43 suites green)

# direct dependents
cargo test -p trusty-analyze  (← trusty-review) — 442 passed, 0 failed
cargo test -p trusty-agents   (← trusty-search) — 3312 passed, 0 failed
```

### Two pre-existing reds, neither mine

**`cargo test -p trusty-search` — 1428 passed, 6 failed.** Exactly the six FSEvents watcher tests [`docs/reference/test-ladder-baseline.md`](docs/reference/test-ladder-baseline.md) names as environmental on this host. Confirmed on a clean `origin/main` worktree at `fd9d913c`:

```
test service::watcher::tests::removed_event_emitted_on_delete ... FAILED
test service::watcher::tests::modified_event_emitted_within_one_second ... FAILED
test service::watch_loop::tests::modified_docx_file_triggers_indexing_via_extractor ... FAILED
test service::watch_loop::tests::modified_file_triggers_indexing ... FAILED
test service::watcher_manager::tests::save_triggers_incremental_index_via_manager ... FAILED
test service::watch_loop::tests::cdk_out_file_is_not_indexed ... FAILED
test result: FAILED. 15 passed; 6 failed; 0 ignored; 0 measured; 1407 filtered out
```

Identical set, identical count. Path evidence — the diff touches none of those files:

```
$ git diff --name-only origin/main...HEAD -- \
    crates/trusty-search/src/service/watch_loop.rs \
    crates/trusty-search/src/service/watcher.rs \
    crates/trusty-search/src/service/watcher_manager.rs
(empty)
```

The change-specific gate is green: `cargo test -p trusty-search --lib -- core::git service::reconcile` — **31 passed, 0 failed**.

**`cargo check --workspace` — `trusty-mpm-gui` and `trusty-code-gui` fail.** The pre-existing Tauri `ui/dist` failure: `The frontendDist configuration is set to "ui/dist" but this path doesn't exist`. That directory is a `pnpm build` output and is **not tracked in git at all**, so no checkout of any branch has it without a frontend build:

```
$ git ls-files crates/trusty-mpm-gui/ui/dist crates/trusty-code-gui/ui/dist
(empty)
$ git diff --name-only origin/main...HEAD -- crates/trusty-mpm-gui/ crates/trusty-code-gui/
(empty)
```

Every other workspace member compiled. Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d, or `--lib`-narrowed to make either gate green.

---


bobmatnyc:
## Verdict: APPROVE

Adversarial review of the three git fail-open sites. Independent worktree at `18e5fa88`; no edits to the PR branch (mutations ran in a separate scratch worktree).

**Yes, I would run this against a repo containing real credentials.** No CRITICAL, no HIGH.

### 1. Can a secret still reach any of the three destinations? — No, for every failure mode I could construct

I enumerated git's actual behaviour on git 2.54.0 rather than trusting the table in the PR body:

| failure mode | `ls-files` / `rev-parse` stderr | long phrase? | ancestor `.git` witness? | classifier verdict |
|---|---|---|---|---|
| genuinely no repo | `not a git repository (or any of the parent directories): .git` | yes | no | `NoRepo` → walk / mtime / copy ✅ correct |
| stale worktree pointer | `not a git repository: (null)` | **no** | yes | refuse |
| unreadable `.git` (mode 000) | `not a git repository (or any of the parent directories): .git` | **yes** | **yes** | refuse — *only the witness saves this one* |
| broken `repositoryformatversion` | `Expected git repo version <= 1, found 99` | no | yes | refuse |
| `GIT_CEILING_DIRECTORIES` | `not a git repository (or any of the parent directories): .git` | yes | **yes** | refuse — the witness walk runs to `/`, so it is strictly more thorough than git's ceiling-limited discovery |
| `dubious ownership` | `detected dubious ownership in repository at '/x'` | no | yes | refuse |
| bare repo | exit **0**, `rev-parse` prints `false` | n/a | n/a | see below |
| missing `git` binary | spawn `Err` | n/a | n/a | refuse (site 1 `Refused`, site 2 `Unknown`, site 3 `Unknown`) |
| `.git` file (valid linked worktree) | success | n/a | n/a | normal path — shared `info/exclude` resolved correctly |

The bare-repo case is the one that exits 0, and it lands safely at all three sites for three different reasons, which I verified individually:
- site 1: `git ls-files` in a bare repo exits 0 with an empty listing → `Tracked(vec![])` → no walk. **The empty-listing fix is what closes this**; pre-fix it was `None` → walk.
- site 2: `rev-parse --is-inside-work-tree` prints `false` → `Unknown` → full reindex.
- site 3: `rev-parse --git-path info/exclude` **succeeds** (`Registered`), but `git check-ignore` then fails with `fatal: this operation must be run in a work tree` (exit 128) → every path refused.

### 2. The exit-code claim — verified independently, on git 2.54.0

```
$ git check-ignore -q -- tracked.env    # in info/exclude, but tracked
exit=1
$ git check-ignore -q -- absent.env     # in info/exclude, untracked
exit=0
$ git add -A && git status --porcelain  # after modifying tracked.env
M  tracked.env
```

The overwrite hazard was real (`git add -A` does stage the modification) and the register-first-then-verify ordering **closes it rather than narrowing it**: `is_path_git_ignored` gates the `fs::copy` itself, so the tracked file is never written. I also confirmed the broken-`info/exclude` premise the PR relies on:

```
$ mkdir .git/info/exclude && git check-ignore -q -- absent.env
fatal: cannot use .git/info/exclude as an exclude file
exit=128
```

→ `is_path_git_ignored` → `false` → refuse. Unanswerable is a refusal, as claimed.

### 3. Per-site conservative branches

- **Site 1 — does `Refused` genuinely yield no corpus?** Yes. `Corpus` and `list_files` are private; the only two consumers are `scan_repo` (empty → `None`, before `detect_frameworks` runs) and `list_tracked_files` → `investigate::run_investigation` (empty selection → `Skipped`, no LLM call). I checked for a second enumeration path in the crate — `read_dir`/`WalkDir` appear only in output/voice/benchmark code, never in the corpus builder.
- **Site 2 — is the full reindex actually gitignore-honouring?** Yes, and for a stronger reason than the PR body states. `reindex/orchestrator.rs:93` passes `handle.respect_gitignore` (default `true`) into `walk_source_files_with_options`, and `walker.rs:659` sets **`require_git(false)`** — so the `ignore` crate applies `.gitignore` even when the repository is unreadable by git. The conservative branch does not depend on git working at all.
- **Site 3 — does the `check-ignore` reuse cover every file the sync copies?** Yes. The gate is in the per-candidate loop and applies to every deduped candidate regardless of which pattern matched it, not just the default `.env*`. The only bypass is the corroborated-`NoRepo` destination (`in_repo == false`), which is the documented design.

### 4. Mutation sweep — hunting for the PR #4746 masking shape

Re-ran the trusty-search mutation as asked. **Confirmed, and confirmed the author's diagnosis of why it initially passed:**

```
M2c: NO_REPO_STDERR := "not a git repository"
core::git::tests::probe_work_tree_is_unknown_for_a_stale_worktree_pointer ... ok      ← masked
core::git::tests::classify_probe_failure_corroborates_the_no_repo_message ... FAILED  ← the added assertion
  left: NoRepo  right: Unknown
  "the stale-worktree near-miss contains the short phrase but means the opposite"
```

I then swept all three sites for the same shape. The equivalent witness-free wording assertions exist at sites 1 and 3 (`classify_ls_files_failure_rejects_near_miss_and_unknown_wordings`, `classify_rev_parse_failure_corroborates_the_no_repo_message`), so the masking is not repeated. I applied six further mutations of my own; four were caught:

| mutation | result |
|---|---|
| M1b — empty listing falls through to the walk | `scan_empty_git_repo_does_not_walk_untracked_files` FAILED ✅ |
| X9 — `is_path_git_ignored` always returns `true` | `tracked_secret_is_not_overwritten_in_worktree` + `broken_exclude_file_still_refuses_unignored_paths` FAILED ✅ |
| X11 — `NoRepo \| Unknown` both take the mtime path | `broken_git_repo_full_reindexes_instead_of_mtime_walking` FAILED ✅ |
| **X1 — drop `.canonicalize()` in all three classifiers** | **all suites still pass ⚠️ (M1)** |
| **X4/X6/X7 — spawn failure concedes `NoRepo` instead of refusing** | **all suites still pass ⚠️ (M2)** |

Gates re-run independently: `cargo fmt --check` clean; `check_test_pointers.sh` 20681 citations, 0 dangling; `check_line_cap.sh` 0 violations; `check_changelog_fragment.sh` all 3 crates recorded; `cargo clippy -p trusty-review -p trusty-search -p trusty-mpm --all-targets -- -D warnings` clean; `cargo test -p trusty-review` **1564 passed, 0 failed**; `cargo test -p trusty-mpm` **4316 passed, 0 failed**; `cargo test -p trusty-search --lib -- core::git service::reconcile` **31 passed, 0 failed**. The six FSEvents reds are the documented environmental set, as stated.

Housekeeping confirmed: `Cargo.lock` diff is the `trusty-review 0.11.0 → 0.11.1` version line and nothing else; three per-crate fragments, one `Security` category each. Rebase onto `ebdcbef0` is clean — zero file overlap between what main changed since `fd9d913c` and what this PR touches.

---

## Findings

| MEDIUM | `crates/trusty-search/src/core/git.rs` / `crates/trusty-review/src/report/scan.rs` / `crates/trusty-mpm/.../untracked_sync.rs` | 165 / 210 / 396 | **M1 — the `canonicalize()` guard is load-bearing and no test exercises it.** All three doc comments claim "a relative `root` would walk a truncated ancestor chain, miss the project's `.git`, and land back on the permissive answer" and cite `classify_*_corroborates_the_no_repo_message` as the test — but that test passes an absolute `tempdir()`, so replacing `root.canonicalize()` with the raw path passes every suite. The claim is true (measured: from a subdir of a repo, `Path::new(".").ancestors()` = `[".", ""]`, witness `false`; canonicalised, witness `true`), and `scan_repo` takes an operator-supplied `local_path` that can be relative. This is the same "no test names this guard" class as the M2c masking. | Add one assertion per site that classifies from inside a subdirectory of a repo using a relative path (`std::env::set_current_dir` + `Path::new(".")`), or drop the canonicalisation claim from the `Test:` pointers. Since #4735 extracts the shared probe, one test there would cover all three. | Parent |
| MEDIUM | `crates/trusty-search/src/core/git.rs` `crates/trusty-review/src/report/scan.rs` `crates/trusty-mpm/.../untracked_sync.rs` | 213 / 274 / 358 | **M2 — the "git binary missing" failure mode is handled correctly but untested at all three sites.** The prompt's own failure-mode list names it. Mutating each spawn-`Err` arm to concede `NoRepo`/walk/copy survives every suite. The shipped code is right; there is simply nothing pinning it. | Inject a `PATH` with no `git` (or a `git` shim that exits 127) for one test per site and assert refusal. | Promote |
| MEDIUM | `crates/trusty-search/src/service/reconcile.rs` | 389–405 | **Unborn-HEAD repos now full-reindex on every daemon boot.** `head_sha` returns `None` for a repo with zero commits, and `probe_work_tree` correctly answers `Present` → the new arm triggers a full background reindex. `reindex/finish.rs:371` re-stamps `indexed_head_sha` from the same `head_sha`, so it stamps `None` again and the condition never converges. Previously this took the mtime path (`skipped_no_data`, a cheap no-op). The behaviour is the *correct* safe direction — an unborn-HEAD repo has a live `.gitignore` — but the cost is an unbounded repeat, not a one-shot recovery like the #4680 guard. Mitigated by `force = false` (hash cache skips unchanged files) and background priority. | Either stamp a sentinel so the fallback fires once per index per daemon lifetime (mirroring `last_walk_started_at` in the #4680 guard), or gate the arm on `WorkTree::Unknown` only and let a corroborated `Present` with an unborn HEAD fall through to the mtime path — which is safe here precisely *because* it is corroborated. | Promote |
| LOW | `crates/trusty-mpm/.../untracked_sync.rs` | 122–136 | **Registration is now unconditional, so refused paths still mutate the repo-wide shared `info/exclude`.** `append_paths_to_exclude` runs over *all* candidates before any per-path verdict, and `rev-parse --git-path info/exclude` resolves to the **shared common dir**, so a refused path (e.g. a tracked `.env`) permanently gains an entry visible to the base clone and every sibling worktree. Harmless in the common case (a tracked path stays not-ignored regardless), but it is a persistent cross-worktree side effect of an operation that was refused. Related: paths are written verbatim with no gitignore-metacharacter escaping — a filename containing `[`, `]` or a trailing space produces a pattern that does not match itself, which fails *closed* (refused, not leaked); a filename literally named `!.env` matched by an operator-configured `*.env` pattern writes a **negation** that un-ignores `.env` repo-wide, and again fails closed for the copy but leaves the residue behind. Not reachable with `DEFAULT_UNTRACKED_SYNC_PATTERNS`. | Register only the paths that will actually be copied (verify first against a temporary exclude, or prune the entry on refusal), and escape leading `!`/`#` and `[`/`]` when writing patterns. | Promote |
| LOW | `crates/trusty-review/src/report/investigate/mod.rs` | 241 | A refused corpus is reported to the operator as `Skipped("no readable source files")`, which is indistinguishable from a genuinely empty repository. The `warn!` in `list_files` is the only signal that a security refusal occurred, and it does not reach the report. | Thread the refusal through so the investigation status reads e.g. `Skipped("git could not be consulted — corpus refused (#4733)")`. | Promote |
| LOW | `crates/trusty-mpm/.../untracked_sync.rs` | 19–21 | Doc bound: the module header still asserts the copied file is "gitignore-safe" as an absolute. Post-fix the guarantee is **point-in-time at copy** — nothing stops the session's own agent from later editing `.gitignore`/`info/exclude` or running `git add -f`. The new doc comments are otherwise carefully bounded ("CONSISTENT with", "necessary and never a sufficient condition", "verified byte-identical against git 2.54.0"), which is the right calibration; this one line predates the change and now overstates. Same class as the #4723 false invariant. | Restate as "confirmed ignored at the moment of copy". | Fix here |

## Notes

- Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d or `--lib`-narrowed anywhere in this diff; I checked the test diffs line by line. The two removed items (`append_to_git_exclude`, its direct test) are genuinely dead after the split, and the idempotency test was rewired to `resolve_git_exclude` + `append_paths_to_exclude` rather than deleted.
- `mtime_walk_does_not_honour_gitignore` is the right call — it pins the premise the gate exists for, so the gate cannot later be removed as redundant. I confirmed the premise holds by reading `collect_stale_files_by_mtime`, not just by trusting the test.
- Deliberately not flagged: the `WorkTree::Present` vs `Unknown` distinction is behaviourally inert at the sole call site (both full-reindex), so no test can catch a mutation of the `stdout != "true"` bare-repo check. That is a consequence of the enum being wider than its consumer, not a defect — and #4735 will give it a second consumer.
- All three MEDIUM findings are test-coverage or cost observations. None of them is a path by which a gitignored file reaches the LLM corpus, the search index, or git history.

---

bobmatnyc:
## Review round — three MEDIUMs addressed. Rebased onto `9987a42b`; all 27 checks green.

### 1. Unborn-HEAD repeated reindex — **I introduced it**, measured, now fixed

Not guessed. Identical measurement test run at both refs against a real `git init` repo with no commits:

| ref | `skipped_no_data` | `fell_back_to_full` |
|---|---|---|
| `origin/main` @ `9987a42b` | **1** | 0 |
| this branch (before the fix) | 0 | **1** |

The mechanism: a `git init` with nothing committed is a *healthy* repo whose `rev-parse HEAD` exits 128, so `head_sha` is `None` while `--is-inside-work-tree` returns `true`. My first fix routed that to `trigger_full_reindex` — but `finish_reindex` re-stamps `indexed_head_sha` from the same `head_sha` that just returned `None`, so `stored` never becomes `Some` and the next boot re-enters the identical arm. Re-walks the whole tree, every boot, forever.

**Fix:** the branch now splits on whether a reindex could **converge**.

| state | outcome | why |
|---|---|---|
| corroborated `NoRepo` | mtime path | the only case the `.gitignore`-blind walk is legitimate |
| work tree + live HEAD, never stamped | full reindex | one run stamps the SHA; next boot converges |
| work tree, **no HEAD at all** | **refuse**, count `skipped_unresolvable_git` | a reindex cannot stamp anything, so it would repeat forever |

Refusing costs nothing the mtime path actually delivered. Its only input, `last_indexed_unix`, has **no production writer** (#4391) — verified by grep, not assumed:

```
$ grep -rn "update_last_indexed_unix" crates/ --include="*.rs" | grep -v persistence_timestamps.rs
crates/trusty-search/src/service/persistence.rs:370:    read_last_queried_unix, update_last_indexed_unix, …   # re-export only
```

So this case has always reported "skipped" in production, and the fix restores exactly `origin/main`'s outcome at zero cost, with no gitignore-blind walk.

The refusal gets its **own** counter rather than folding into `skipped_no_data`: a security refusal reading as ordinary emptiness is precisely how an operator misses it. Those indexes still serve (with `results_may_be_stale`), the live watcher still tracks them, and the #4680 never-walked guard still re-drives them.

### 2. `.canonicalize()` — pinned at all three sites

Confirmed the critic's finding: dropping it passed every suite. `Path::ancestors` walks **lexically**, so a relative path (which `scan_repo` genuinely receives from operators) or one reached through a symlink yields a chain that is not its real parentage — the project's `.git` is never visited and the permissive answer wins.

Each site now has a `link -> repo/sub` fixture with `.git` on `repo` only. Hermetic (no cwd mutation, no ambient-repo dependency), and the lexical ancestors of `link` never include `repo`.

### 3. Missing-git-binary arms — pinned at all three sites

Rather than mutate `PATH` (process-global, racy under a parallel runner), each probe gained a `_with(…, git_bin)` seam so the spawn-failure branch is reachable hermetically. At site 3 the sanity assertion is included: the *same* worktree resolves `Registered` with a real git and `Unknown` without one, so the test cannot pass for the wrong reason.

### Doc: site 2 is safe for a stronger reason than I claimed

Added to the module doc — `service::walker` drives the `ignore` crate with `require_git(false)`, so the reindex walk applies `.gitignore` **even when git cannot read the repository at all**. That is what makes "reindex instead of mtime-walk" a real fix rather than a swap of one blind walk for another. It does not depend on git working.

### Seven further mutations, all confirmed red (18 total)

| # | Mutation | Test that failed |
|---|---|---|
| M4a/b/c | drop `.canonicalize()` (review / search / mpm) | `*_canonicalises_before_walking_ancestors` at each site |
| M5a | spawn failure walks instead of refusing (review) | `git_ls_files_refuses_when_the_git_binary_is_missing` |
| M5b | spawn failure reads as `NoRepo` (search) | `probe_work_tree_is_unknown_when_the_git_binary_is_missing` |
| M5c | spawn failure reads as `NoRepo` (mpm) | `missing_git_binary_copies_nothing` |
| M6 | unborn-HEAD refusal reverted to full reindex | `unborn_head_repo_is_refused_once_not_reindexed_every_boot` + `broken_git_repo_refuses_rather_than_mtime_walking` |

### Gates — rung 5, re-run after the rebase

```
cargo fmt --check                                   — clean
check_line_cap.sh    — 3621 files, 7 allowlisted, 0 violations — OK
check_sld.sh         — 54 specs + 3045 code files, 0 errors, 0 warnings
check_test_pointers.sh — 20736 citations, 0 dangling — OK
check_changelog_fragment.sh — all 3 crates recorded

clippy -p trusty-review / trusty-search / trusty-mpm --all-targets -- -D warnings  — all clean

cargo test -p trusty-review -- --include-ignored   — 1571 passed, 0 failed, 0 ignored
cargo test -p trusty-mpm --lib                      — 4339 passed, 0 failed
cargo test -p trusty-analyze  (← trusty-review)     — 442 passed, 0 failed
cargo test -p trusty-agents --lib (← trusty-search) — 3312 passed, 0 failed
```

One incidental cleanup: correcting a doc pointer left a **stale** `.test-pointer-allowlist.tsv` entry (`append_to_git_exclude_resolves_shared_worktree_path` had been allowlisted as dangling; it no longer is). Pruned via the script's documented `--update`.

**`cargo test -p trusty-search` — 1432 passed, 6 failed**, the same documented FSEvents watcher set, re-confirmed against the *new* base `9987a42b` in a clean worktree (15 passed, 6 failed, identical names). Diff touches none of those files:

```
$ git diff --name-only origin/main...HEAD -- \
    crates/trusty-search/src/service/watch_loop.rs \
    crates/trusty-search/src/service/watcher.rs \
    crates/trusty-search/src/service/watcher_manager.rs
(empty)
```

Change-specific gate green: `cargo test -p trusty-search --lib -- core::git service::reconcile` — **35 passed, 0 failed**. CI's own `Test` job passed in 16m8s, which is the independent confirmation that the six are this host.

Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d, or `--lib`-narrowed to make any gate green.

Issue #3381, opened 2026-07-20
bug(trusty-mpm): InstallPolicy is behaviorally inert — install_to() never reads artifact.install, so Overwrite artifacts are skip-if-exists in practice


bobmatnyc:
## Finding

\`InstallPolicy\` (in \`crates/trusty-mpm/src/core/bundle.rs\` / consumed via the \`install\` field on bundle artifacts) is **behaviorally inert in production**: the actual installer — \`crates/trusty-mpm/src/bin/tm/commands/install.rs:479-497\` \`install_to()\` — never reads the \`artifact.install\` field. It uniformly does \`if dest.exists() && !force { skip } else { write }\` for EVERY artifact, regardless of \`Overwrite\` vs (the now-removed) \`SeedOnce\`.

## Consequences

- The enum was documentation-by-type only: the type system promised per-artifact install semantics (e.g. overwrite-on-upgrade vs seed-once) that the installer does not implement.
- Concretely: artifacts registered as \`Overwrite\` (framework instructions, agents, skills, hooks — things that SHOULD be refreshed on upgrade) are in fact **skipped whenever the destination file already exists** unless \`--force\` is passed. That means a plain \`tm install\`/upgrade may leave stale framework assets in place while the code assumes they were refreshed.
- Cross-check needed: if a different code path handles refresh at session launch, say so in investigation; if not, this is a live inconsistency.
- PR #3377 removed the unconstructed \`SeedOnce\` variant (correct, no behavior change), but the remaining \`Overwrite\`-vs-installer-behavior mismatch is unresolved.

## Root Cause

The \`install_to()\` function implements only the "skip if exists" logic and never inspects the \`artifact.install\` field. This means \`InstallPolicy\` is a dead field — defined by the type system but ignored at runtime.

## Candidate Fixes

**(a) Make \`install_to()\` honor the policy field** (preferred if upgrade flows require refreshed assets): \`Overwrite\` artifacts always rewritten, \`Skip\` artifacts only written on first install.

**(b) Delete the enum entirely** and document the actual uniform semantics in the bundle.rs docs.

## Investigation Note

The doctor's \`skill_staleness\` and \`output_style_staleness\` warnings suggest staleness IS a real, observed phenomenon. This is consistent with the installer never overwriting bundled assets. Verify whether session-launch deployment (possibly in the doctor or in tm's session-init flow) refreshes assets separately — if so, document that path; if not, (a) becomes critical for upgrade safety.

---
[...]

Issue #2170, opened 2026-07-07
fix(trusty-mpm): stop injecting the delegation-directive block into a managed workspace's CLAUDE.md — deliver it via the output-style instead (never modify the project's CLAUDE.md)


bobmatnyc:
## Problem
The session-launch pipeline (issue #2125) regenerates a framework-owned delegation-directive block — tagged `<!-- trusty-mpm:delegation-directive:begin (framework-owned — do not edit; regenerated on every session start, see issue #2125) -->` — at the TOP of a managed workspace's `CLAUDE.md` on every session start (and, as a side effect, during `tm validate --repair`). This VIOLATES the standing owner constraint: **trusty-mpm must never modify the target project's `CLAUDE.md`; our instructions must survive theirs without touching them.** It was observed modifying `/Users/masa/trusty-mpm-projects/duettoresearch/apex/.../CLAUDE.md` during a deployment repair.

## Key context
The delegation directive **already lives in the trusty-mpm output-style** (`crates/trusty-mpm/src/assets/output-styles/trusty-mpm.md` — the "PRIMARY DIRECTIVE — MANDATORY DELEGATION" section), which is delivered as the managed session's system prompt via `.claude/settings.json`'s `outputStyle` key. So the `CLAUDE.md` injection is largely a REDUNDANT second copy.

## Fix (owner-directed: move it OUT of CLAUDE.md)
1. **Stop writing the delegation-directive block into `CLAUDE.md`.** Remove the #2125 injection/regeneration from the session-launch/instructions pipeline (find it via `grep -rn "trusty-mpm:delegation-directive"` and the #2125 code). No code path may add/modify/regenerate content in a managed workspace's `CLAUDE.md`.
2. **Ensure the directive is fully delivered via the output-style/settings** (it already is — confirm the output-style's PRIMARY DIRECTIVE section is complete and authoritative; if the CLAUDE.md version had any content not in the output-style, port it into the output-style, NOT CLAUDE.md).
3. **Do not touch a pre-existing foreign `CLAUDE.md` at all** — no stub, no block, no reordering. If a workspace has no CLAUDE.md, the existing "create a stub" behavior may remain, but it must not inject the delegation block.
4. Optionally: a one-time cleanup helper to strip already-injected `trusty-mpm:delegation-directive` blocks from managed workspaces' CLAUDE.md files (nice-to-have, flag if out of scope).

## Acceptance
[...]

PR #4759, opened 2026-08-04
fix(trusty-mpm): compiled PM prompt gets its own path, written before launch (#4752)

### LINKED ISSUE #4752: Compiled PM prompt must write to a distinct ~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md
**Defect.** Two writers target one path. `bundle_all.rs` writes a 4-line stub to `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md`; `instruction_pipeline.rs`'s `install_system_prompt` writes the *full composed prompt* to that same path. Last writer wins. Neither file is read by the launch path, which composes in memory and passes the result via `--append-system-prompt-file`. This collision is recorded as Finding #1 in `crates/trusty-mpm/src/assets/instructions/sections/README.md`.

**Consequence.** The on-disk artifact a user inspects to answer "what instructions is my session actually running?" is non-deterministic and may be a 4-line stub.

**Resolution (owner ruling, 2026-08-04).**
- The composed prompt is written to a distinct new file: `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`.
- The project's `CLAUDE.md` carries a pointer **to** `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`, so a reader of a project's CLAUDE.md can find the compiled system prompt actually in force. Direction is CLAUDE.md → compiled file, not the reverse. Constraint: the root `CLAUDE.md` is not tracked on main and `claude-md-guard` forbids tracking it (issue #2299), so the pointer belongs in the surface that scaffolds a project's CLAUDE.md, not in a committed root file.
- The `bundle_all.rs` stub no longer shares a path with the compiled output.
- Update the spec of record `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md` and `sections/README.md` (retire Finding #1).
- INSTRUCTIONS-COMPILED.md must be fully written and current **before** Claude Code launches — the write is ordered before, and blocks, the `claude` spawn. Not async, not best-effort, not install-time-only. If making the write blocking slows the launch pipeline, the pipeline gets fixed; the ordering does not get relaxed. The composer already holds the full prompt in memory for `--append-system-prompt-file`, so this should cost one additional file write — a second composition pass would be the thing to eliminate, not a reason to go async.

**Note.** [#4286](https://github.com/bobmatnyc/trusty-tools/issues/4286) is obsolete and must not be cited as live doctrine.

**Prior art.**

Issue [#383](https://github.com/bobmatnyc/trusty-tools/issues/383) ("tm install overwrites assembled system prompt with 4-line stub") is the same defect, previously closed. It regressed. A last-writer-wins patch is not acceptable; the paths must be distinct.

### LINKED ISSUE #4753: sections/non-overridable-rules.md resurrects retired .trusty-mpm/ override files on trigger phrases
**Defect.** `sections/non-overridable-rules.md` instructs the PM to create the five `.trusty-mpm/` override files (`INSTRUCTIONS.md`, `AGENT_DELEGATION.md`, `WORKFLOW.md`, `MEMORY.md`, `PM_INSTRUCTIONS_DEPLOYED.md`) in response to trigger phrases. Those files are retired — see the "Retired — do not bring these back" table in `crates/trusty-mpm/src/assets/instructions/sections/README.md`.

**Consequence.** Every PM in every project is instructed to recreate a retired mechanism. A leftover file then fails `tm doctor`'s `legacy_overrides` check.

**Resolution.** Remove the trigger-phrase instruction from `sections/non-overridable-rules.md`. Section overrides go in the project's `CLAUDE.md` as marked blocks; the spec of record is `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md`.

bobmatnyc:
Closes #4752

**Test ladder rung: 3** (localized behavior inside one crate) — `trusty-mpm` only.

## The defect

One path carried three roles:

| Role | Symbol |
|---|---|
| compiled prompt **output** | `instruction_pipeline.rs` — `install_system_prompt` / `install_system_prompt_to` |
| bundled 4-line **stub** | `bundle_all.rs` — `ALL` (already removed, see below) |
| pipeline **input** | `instruction_pipeline.rs` — `build_instructions` reads `PipelineInput::framework_instructions_path` |

Last writer won, so the artifact an operator inspects to answer "what instructions is my session running?" was non-deterministic. #383 fixed one instance by ordering the writes — ordering is not a fix when two roles share one name.

## What changed

- **Distinct path.** `FrameworkPaths::instructions_compiled()` → `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`, directly under `framework/`, **not** under `framework/instructions/`. Filename constant: `instruction_pipeline::COMPILED_PROMPT_FILE`.
- **One writer function.** Both call sites go through `write_compiled_prompt_to`, so the file can only ever hold a full compiled prompt, never a stub.
- **Stale leftover removed.** `tm install` deletes a pre-#4752 `instructions/INSTRUCTIONS.md` via `remove_stale_bundled_instructions` — see "Stale leftover" below.
- **`non-overridable-rules.md`** no longer calls section overrides "override files"; the live mechanism is marked blocks in the project's `CLAUDE.md`, now pointing at the spec of record. Golden prompts regenerated with the harness's own `UPDATE_GOLDEN=1`; the diff is exactly those three lines in three files, nothing else.
- **Spec §10** added to `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md`; **Finding #1 retired** from `sections/README.md`.

## Ordering — before, after, and cost

**Today (before this PR):** `install_system_prompt_to` was called from exactly one production site, `commands/install.rs::install` — **install time only**. `install_system_prompt()` (the home-resolving variant) had **zero** production callers. The launch path (`session_launch::prepare_session_inner` → `build_system_prompt_for_with_style_and_native`) never refreshed the framework-level artifact, so it went stale the moment a project's `CLAUDE.md` overrides changed.

**After:** `prepare_session_inner` writes the prompt to `fw.instructions_compiled()` immediately after the `.trusty-mpm/last-instructions.md` stash and before returning. The write is **fatal** (`?` → `PrepError::Io`), not best-effort. Every caller spawns `claude` only after `prepare_session` returns `Ok`, so *"prepare returned"* **is** *"the compiled prompt on disk is the prompt this session will run with."*

**Cost — measured? No.** I did not benchmark, and I am not claiming "no regression". The structural argument: the launch write reuses the `resolved_prompt` string the composer already built for `--append-system-prompt-file` — the same text just written to the project stash — so it is one additional `fs::write` with **no second composition pass**.

## Tests — 9 new, and they fail before the change

The two ordering tests were verified fail-before by temporarily removing the launch-time write:

```
test core::session_launch::tests::prepare_session_writes_the_compiled_prompt_before_returning ... FAILED
test core::session_launch::tests::prepare_session_fails_when_the_compiled_prompt_cannot_be_written ... FAILED

panicked at tests.rs:302: assertion `left != right` failed:
  the launch must overwrite a stale compiled prompt, not leave it in place
panicked at tests.rs:340:
  a failed compiled-prompt write must abort the launch preparation: PrepReport { … }
```

Fixtures are built so only the named guard can make them pass:

- `prepare_session_writes_the_compiled_prompt_before_returning` — the compiled path is **pre-seeded with stale sentinel content**, so a mere "the file exists" check would pass against a broken implementation. It asserts the content is not the sentinel *and* is byte-identical to the stash (the exact `--append-system-prompt-file` payload).
- `prepare_session_fails_when_the_compiled_prompt_cannot_be_written` — this is the one that pins **blocking**, not just eventual: a directory is planted at the compiled path so only that write fails. If the write were best-effort or deferred, `prepare_session` returns `Ok` and the test fails (as shown above).
- `bundle_install_pass_never_touches_the_compiled_prompt` — pre-seeds a sentinel and asserts it **survives** `install_to`; an absence check would pass trivially on a fresh temp dir.

No coverage was deleted, `#[ignore]`d, `cfg`-gated, `--exclude`d, or `--lib`-narrowed.

## Findings you should read before merging

**1. The `bundle_all.rs` stub was already gone — I deleted nothing.** #4286 split A had already removed the `instructions/INSTRUCTIONS.md` entry from `bundle::ALL` and deleted the `FRAMEWORK_INSTRUCTIONS` constant. Verified:

```
$ git grep -n 'FRAMEWORK_INSTRUCTIONS' -- crates/
crates/trusty-mpm/src/core/bundle.rs:11:  (doc comment only)
crates/trusty-mpm/src/core/bundle_tests.rs:406:  (comment: "the constant it backed is gone too. 179 - 1 = 178")
```

`bundle_table_is_complete` pins `ALL.len() == 178`. Finding #1 in `sections/README.md` described a collision whose first writer no longer existed. The surviving half — compiled output sharing a path with a pipeline input — is real, and is what this PR fixes.

**2. Nothing else was dead.** Per the "verify each removal with a `git grep`" instruction: `FrameworkPaths::framework_instructions` / `framework_instructions_path` and the `instructions` field **retain live readers** — `session_launch::prepare_session_inner` and `commands/session.rs` both still build a `PipelineInput` from them. They were kept.

**3. Stale leftover — handled, not ignored.** After this change nothing writes `instructions/INSTRUCTIONS.md`, but `build_instructions` still *reads* it. An upgraded machine would fold a frozen copy of an old compiled prompt into `PipelineOutput::merged` forever, with no writer left to refresh it. `tm install` now removes it. Safe: the file is framework-owned (trusty-mpm has always overwritten it on every install), `build_instructions` already treats absence as normal (`instructions_loaded: false`), and `merged` has no production reader.

**4. No test enshrined the defect.** Two existing tests pinned the old path (`install_system_prompt_writes_file`, `install_system_prompt_to_writes_assembled`); both were **rewritten to assert the new behavior**, with the first strengthened from `ends_with("INSTRUCTIONS.md")` to a full path equality. Neither was ignored or weakened.

**5. `#4753`'s described defect does not exist — marked `Refs`, not `Closes`.** #4753 states `non-overridable-rules.md` tells the PM to create the five retired `.trusty-mpm/` override files on trigger phrases. That is **not** true of current `main`. The shipped file already reads:

> Trigger phrases -> act immediately, always in `CLAUDE.md`
> …
> The `.trusty-mpm/` override files … are RETIRED and are no longer read (#4286). **Never create one.**

Every trigger maps to `CLAUDE.md` prose or a marker block. This matches spec §9.7, which records that commit `0aec8866` (#4592) already rewrote that table. A full sweep of all nine shipped section files found **no** other place describing the retired files as live — the only other `.trusty-mpm/` mentions (`core.md`, `enforcement.md`) refer to orchestration state and config, which is correct.

What I *did* fix in that file is the one genuine residue: line 66 said "Missing, empty, or unreadable override **files**", leftover wording from the file-based era. **I did not write `Closes #4753`**, because closing it on the strength of a one-line wording change would misrepresent what was fixed. Recommend closing #4753 as already-resolved-by-#4592, or reopening it with an accurate description — owner's call.

**6. CLAUDE.md pointer — deliberately not delivered.** Superseded mid-flight by the ruling that all `CLAUDE.md` modification is reserved to the scaffolding agent. I had written the pointer into `CLAUDE_MD_STUB` and the `tm-init` skill; **both were backed out** — `tm-init.md` is untouched in this diff. Reported for a separate decision:

- **Two code paths mutate a project `CLAUDE.md` today, neither of which is the scaffolding agent** — this plurality is itself the defect the ruling names:
  - `instruction_pipeline::load_or_create_claude_md` seeds `CLAUDE_MD_STUB` on first session start (reached via `build_instructions`).
  - `session_launch::worktree_sync::self_heal_claude_md` strips the legacy #2170 block on **every session resume**, via `strip_delegation_block`.
- **The scaffolding path is the `tm-init` skill** — agent-driven prose with no code component; it delegates the file write to the PM or an engineer agent.
- **It cannot currently write a marked section-override block.** `grep 'TRUSTY-MPM:' tm-init.md` returns nothing; the skill has no knowledge of the `<!-- TRUSTY-MPM: <TOKEN> START v=1 -->` format. That capability is **unbuilt**, so where the pointer lives is an open decision, not something I assumed either way.

Recorded in spec §10.5.

## Gates — rung 3, raw output

```
$ cargo fmt --check
fmt exit=0

$ cargo check -p trusty-mpm
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 15.24s
check exit=0

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 21.81s
clippy exit=0

$ cargo test -p trusty-mpm
TEST_EXIT=0
45 result lines, all ok
TOTAL passed=7410 failed=0 ignored=17

line-cap: measured 3622 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 54 spec doc(s) + 3042 code file(s); 0 error(s), 0 warning(s)

changelog-fragment gate: scanned 15 changed path(s); all 1 crate(s) with source changes are recorded.
```

The 9 new tests, run by exact name:

```
running 9 tests
test core::paths::tests::instructions_compiled_is_directly_under_framework ... ok
test core::paths::tests::instructions_compiled_is_not_the_framework_instructions_path ... ok
test core::instruction_pipeline::tests::no_bundled_artifact_targets_the_compiled_prompt_path ... ok
test core::instruction_pipeline::tests::remove_stale_bundled_instructions_is_a_noop_when_absent ... ok
test core::instruction_pipeline::tests::compiled_prompt_path_is_distinct_from_the_bundled_instructions_path ... ok
test core::instruction_pipeline::tests::remove_stale_bundled_instructions_deletes_a_leftover ... ok
test core::instruction_pipeline::tests::write_compiled_prompt_to_creates_parent_dirs ... ok
test core::session_launch::tests::prepare_session_fails_when_the_compiled_prompt_cannot_be_written ... ok
test core::session_launch::tests::prepare_session_writes_the_compiled_prompt_before_returning ... ok

test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 4346 filtered out; finished in 0.44s
```

**No pre-existing failures encountered.** The three `pm_prompt_golden_*` tests went red mid-run from my own `non-overridable-rules.md` edit and were resolved by regenerating the goldens through the sanctioned `UPDATE_GOLDEN=1` path, with the diff reviewed line-by-line — not by weakening the assertion.

**Line-cap note:** the new bundle-writer test initially pushed `tests_behavior_a.rs` to 510 SLOC. That basename does not match the `_tests.rs` pattern, so it carries the 500 **production** cap. Per that file's own comment, the test was moved to its `install_policy_tests.rs` sibling — the same remedy #3381 used. No file was allowlisted and the gate was not weakened.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,24 @@
 .trusty-mpm/*
 !.trusty-mpm/INSTRUCTIONS.md
 
+# Harness-generated project tier under a CRATE directory is machine-local and
+# never a deliverable (#4752). Running `tm` inside `crates/<name>/` seeds a full
+# project tier there: a `.claude/settings.json` carrying absolute
+# `/Users/<me>/.cargo/bin/...` hook commands, a `.claude/skills/` roster that
+# SHADOWS the bundled one (the #4448/#4526 hazard, same shape as `agents/`
+# above), and a `.trusty-mpm/` session stash.
+#
+# The root-level rules above do NOT reach these, for two different reasons:
+#   * `!.claude/` (above) has no leading or middle slash, so it un-ignores
+#     `.claude` at EVERY level — including every crate's.
+#   * `.trusty-mpm/*` (below) has a middle slash, so it anchors to the repo
+#     root only.
+# 140 such files reached PR #4759 before this rule existed. Scoped to
+# `crates/*/` rather than `**/` on purpose: a bare `**/.claude/` would re-ignore
+# the root `.claude/settings.json` that the re-includes above deliberately keep.
+crates/*/.claude/
+crates/*/.trusty-mpm/
+
 # Re-include tracked files from .claude-mpm/
 !.claude-mpm/INSTRUCTIONS.md
 !.claude-mpm/AGENT_DELEGATION.md
--- a/.line-cap-allowlist.tsv
+++ b/.line-cap-allowlist.tsv
@@ -11,6 +11,6 @@ crates/trusty-common/src/embedder/mod.rs	535
 crates/trusty-git-analytics/src/collect/collector.rs	766
 crates/trusty-mpm/src/daemon/api.rs	1272
 crates/trusty-mpm/src/daemon/api/control_routes.rs	543
-crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs	845
+crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs	818
 crates/trusty-search/src/main.rs	658
 crates/trusty-search/src/service/walker.rs	1167
--- a/.test-pointer-allowlist.tsv
+++ b/.test-pointer-allowlist.tsv
@@ -432,7 +432,6 @@ crates/trusty-mpm/src/core/auto_resume.rs	effective_from_env_parses_truthy
 crates/trusty-mpm/src/core/compress.rs	caveman_summary_contains_tool_and_size
 crates/trusty-mpm/src/core/compress.rs	strip_ansi_leaves_plain_text
 crates/trusty-mpm/src/core/deploy_validate/mod.rs	repair_reports_error_when_prepare_session_fails
-crates/trusty-mpm/src/core/instruction_pipeline.rs	install_writes_assembled_prompt
 crates/trusty-mpm/src/core/manifest/schema.rs	instruction_layers_roundtrip
 crates/trusty-mpm/src/core/manifest/schema.rs	mcp_servers_roundtrip
 crates/trusty-mpm/src/core/manifest/schema.rs	model_tiers_roundtrip
--- a/crates/trusty-mpm/changelog.d/4752-compiled-prompt-launch-coverage.md
+++ b/crates/trusty-mpm/changelog.d/4752-compiled-prompt-launch-coverage.md
@@ -1,0 +1,4 @@
+Fixed
+
+- the compiled PM prompt is now refreshed at the spawn seam, so resume, guided-resume and crash-recovery launches no longer run a prompt that never reached `INSTRUCTIONS-COMPILED.md` (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - the compiled write is the last step of session preparation, so a failure can no longer skip the MCP content-pinning injectors that defend against the #3918/#3950 name-squatting class
--- a/crates/trusty-mpm/changelog.d/4752-compiled-prompt-project-local-fatal.md
+++ b/crates/trusty-mpm/changelog.d/4752-compiled-prompt-project-local-fatal.md
@@ -1,0 +1,7 @@
+Changed
+
+- the compiled PM prompt is now written per project, to `<project>/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`, and a failure to write it refuses the launch (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - removes the cross-project collision by construction: concurrent sessions no longer share one global file
+  - the resume path writes it too, so resumed sessions can no longer run against a prompt that was never recorded
+  - `tm install` no longer writes a compiled prompt — install has no project, so it has nothing to compile
+  - a refused launch names the path and the remedy instead of surfacing a bare I/O error
--- a/crates/trusty-mpm/changelog.d/4752-instructions-compiled-path.md
+++ b/crates/trusty-mpm/changelog.d/4752-instructions-compiled-path.md
@@ -1,5 +1,5 @@
 Fixed
 
-- compiled PM system prompt now writes to a distinct `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` instead of colliding with the bundled stub (closes [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+- compiled PM system prompt now writes to its own `INSTRUCTIONS-COMPILED.md` instead of colliding with the bundled `instructions/INSTRUCTIONS.md` stub that the pipeline reads as input (closes [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
   - the compiled prompt is refreshed at every session launch, before `claude` is spawned; a failed write now aborts the launch rather than leaving a stale prompt on disk
   - `tm install` removes a stale pre-#4752 `framework/instructions/INSTRUCTIONS.md`, which no writer could refresh but the pipeline still read
--- a/crates/trusty-mpm/changelog.d/4752-instructions-compiled-path.md
+++ b/crates/trusty-mpm/changelog.d/4752-instructions-compiled-path.md
@@ -1,0 +1,5 @@
+Fixed
+
+- compiled PM system prompt now writes to a distinct `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` instead of colliding with the bundled stub (closes [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - the compiled prompt is refreshed at every session launch, before `claude` is spawned; a failed write now aborts the launch rather than leaving a stale prompt on disk
+  - `tm install` removes a stale pre-#4752 `framework/instructions/INSTRUCTIONS.md`, which no writer could refresh but the pipeline still read
--- a/crates/trusty-mpm/changelog.d/4753-retired-override-files-instruction.md
+++ b/crates/trusty-mpm/changelog.d/4753-retired-override-files-instruction.md
@@ -1,0 +1,3 @@
+Fixed
+
+- PM instruction section `non-overridable-rules.md` no longer describes section overrides as "override files" — the live mechanism is marked blocks in the project's `CLAUDE.md`, and the section now points at the spec of record (refs [#4753](https://github.com/bobmatnyc/trusty-tools/issues/4753))
--- a/crates/trusty-mpm/changelog.d/4759-inplace-relaunch-compiled-prompt.md
+++ b/crates/trusty-mpm/changelog.d/4759-inplace-relaunch-compiled-prompt.md
@@ -1,0 +1,8 @@
+Fixed
+
+- bare `tm` in a managed pane now refreshes the project's compiled prompt before relaunching, and refuses the relaunch if that write fails (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - this path calls neither session preparation nor the daemon resume route, so it was the one way to start a session on a stale or missing compiled prompt where a fresh launch would have refused
+  - all three pre-spawn paths now route through a single `refresh_compiled_prompt` entry point so they cannot drift apart
+- an unwritable `.trusty-mpm/` no longer skips the MCP injectors (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - the `last-instructions.md` stash write degraded the whole preparation to an early non-fatal error, so callers launched the session anyway with the `trusty-mpm`/`trusty-review` content-pinning defense against the #3918/#3950 name-squatting class never applied
+  - the stash is an inspection artifact; it now logs and continues instead of short-circuiting
--- a/crates/trusty-mpm/changelog.d/4759-inplace-relaunch-compiled-prompt.md
+++ b/crates/trusty-mpm/changelog.d/4759-inplace-relaunch-compiled-prompt.md
@@ -1,8 +1,13 @@
 Fixed
 
+- a session whose instructions cannot be written no longer starts (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - `build_instructions` returned a non-fatal error, so every launch path logged it and started the session anyway — with no instructions established and no `.mcp.json` written. Reachable whenever the project's `CLAUDE.md` or the framework instructions path cannot be read, for example when something has left a directory at `CLAUDE.md`
+  - it is now the same fatal condition as the compiled-prompt write, reported as one error rather than two classes
 - bare `tm` in a managed pane now refreshes the project's compiled prompt before relaunching, and refuses the relaunch if that write fails (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
   - this path calls neither session preparation nor the daemon resume route, so it was the one way to start a session on a stale or missing compiled prompt where a fresh launch would have refused
-  - all three pre-spawn paths now route through a single `refresh_compiled_prompt` entry point so they cannot drift apart
-- an unwritable `.trusty-mpm/` no longer skips the MCP injectors (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
-  - the `last-instructions.md` stash write degraded the whole preparation to an early non-fatal error, so callers launched the session anyway with the `trusty-mpm`/`trusty-review` content-pinning defense against the #3918/#3950 name-squatting class never applied
-  - the stash is an inspection artifact; it now logs and continues instead of short-circuiting
+  - the two resume-shaped paths now share one `refresh_compiled_prompt` entry point; session preparation keeps its own composition because it must apply the operator's resolved output style
+- an unwritable `.trusty-mpm/` no longer causes a session to start without its instructions recorded (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - the `last-instructions.md` stash write degraded the whole preparation to an early non-fatal error, so callers launched the session having skipped the fatal write below it
+  - the stash is an inspection copy; it now logs and continues instead of short-circuiting
+- the `CLAUDE.md` compiled-instructions pointer no longer names the retired global path (refs [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
+  - it pointed at `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`, which nothing writes since the compiled prompt became project-local; it is now the project-relative path, pinned against the pipeline that writes the file
--- a/crates/trusty-mpm/src/assets/instructions/sections/README.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/README.md
@@ -192,7 +192,12 @@ than smoothed over.
   `FRAMEWORK_INSTRUCTIONS` constant, so by the time this finding was acted on
   only one writer remained. The surviving half — the compiled OUTPUT sharing a
   path with the pipeline INPUT `build_instructions` reads — was fixed by giving
-  the compiled prompt its own file, `framework/INSTRUCTIONS-COMPILED.md`
-  (`FrameworkPaths::instructions_compiled`), which nothing else writes. `tm
-  install` now also deletes a stale pre-#4752 `instructions/INSTRUCTIONS.md`,
-  since no writer could refresh it but the pipeline still read it.
+  the compiled prompt its own file, written PER PROJECT to
+  `<project>/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`
+  (`instruction_pipeline::compiled_prompt_path`), which nothing else writes.
+  Project-local rather than global: one shared file under `~/.trusty-mpm/` would
+  simply have moved the collision from two writers on one path to every project
+  overwriting the same copy. `tm install` no longer writes a compiled prompt at
+  all — install has no project, so it has nothing to compile — and it now
+  deletes a stale pre-#4752 `instructions/INSTRUCTIONS.md`, since no writer
+  could refresh it but the pipeline still read it.
--- a/crates/trusty-mpm/src/assets/instructions/sections/README.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/README.md
@@ -172,7 +172,7 @@ than smoothed over.
    name describes nothing true. Renaming it is mechanical and was left out of
    #4286 to keep that change reviewable.
 
-3. **The `project-addendum` generator is declared but unfeedable.** Its only
+2. **The `project-addendum` generator is declared but unfeedable.** Its only
    production input was the retired `.trusty-mpm/INSTRUCTIONS.md`. The block
    stays declared in the manifest (marked `optional`, so it emits nothing) and
    the composer always passes `None`.
@@ -183,3 +183,16 @@ than smoothed over.
   Resolved by the ruling, not by a fix: **nothing** outside `core` is protected
   now, so this stopped being an anomaly and became the documented model. See
   "Why there is no framework floor" above.
+
+- *"Two writers target one path."* — `bundle_all.rs` wrote a 4-line stub to
+  `instructions/INSTRUCTIONS.md` while `install_system_prompt` wrote the full
+  composed prompt to the same path, last writer winning. Resolved by #4752, in
+  two parts. The stub half was already gone: #4286 split A removed the
+  `instructions/INSTRUCTIONS.md` entry from `bundle::ALL` and deleted the
+  `FRAMEWORK_INSTRUCTIONS` constant, so by the time this finding was acted on
+  only one writer remained. The surviving half — the compiled OUTPUT sharing a
+  path with the pipeline INPUT `build_instructions` reads — was fixed by giving
+  the compiled prompt its own file, `framework/INSTRUCTIONS-COMPILED.md`
+  (`FrameworkPaths::instructions_compiled`), which nothing else writes. `tm
+  install` now also deletes a stale pre-#4752 `instructions/INSTRUCTIONS.md`,
+  since no writer could refresh it but the pipeline still read it.
--- a/crates/trusty-mpm/src/assets/instructions/sections/README.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/README.md
@@ -165,14 +165,7 @@ the findings below.
 Written against the actual code, these do not explain cleanly. Recorded rather
 than smoothed over.
 
-1. **Two writers target one path.** `bundle_all.rs` writes a 4-line stub to
-   `instructions/INSTRUCTIONS.md`, and `instruction_pipeline.rs`'s
-   `install_system_prompt` writes the *full composed prompt* to the same
-   `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md`. Whichever runs last
-   wins. Neither is read by the launch path, which composes in memory. The
-   bundled-asset retirement is still open.
-
-2. **`base_pm()` is now a misnomer twice over.** It refers to a file that does
+1. **`base_pm()` is now a misnomer twice over.** It refers to a file that does
    not exist, and it hardcodes four section ids in one order to reconstitute a
    "floor" that no longer exists as a concept. It is still live — the
    roster-absent string assembly appends it — so it is a real function whose
--- a/crates/trusty-mpm/src/assets/instructions/sections/non-overridable-rules.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/non-overridable-rules.md
@@ -63,8 +63,9 @@ delete the file. `tm doctor` fails with `legacy_overrides` until it is gone.
 replaced by a named section in the project's `CLAUDE.md`. There is no framework
 floor: a project owns its own `CLAUDE.md`, so a floor would have been the
 appearance of a control rather than a control.
-Missing, empty, or unreadable override files fall back to the bundled defaults
-— they never blank a section.
+A missing, empty, unclosed, or unreadable marker block falls back to the bundled
+default — an override never blanks a section. Spec of record:
+`docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md`.
 
 ## Trusty Tool Priority (Non-Overridable)
 
--- a/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/guided_inplace.rs
@@ -557,6 +557,23 @@ pub(crate) async fn run_inplace_relaunch(
         return InPlaceOutcome::FallThrough;
     }
 
+    // #4752: refresh the project's compiled prompt BEFORE building the resume
+    // command, and treat a failure as FATAL — the same policy `prepare_session`
+    // and `resume_managed` apply.
+    //
+    // This path is the THIRD entry point that hands a runtime a prompt, and it
+    // calls neither of those two: its only daemon call is `reactivate` below,
+    // which does no prompt work. Until this call existed, the sole compiled write
+    // here was `build_prompt_file`'s best-effort refresh, so a bare-`tm` relaunch
+    // in a managed pane was the one way to start a session on a stale or missing
+    // compiled prompt where a fresh launch would have refused.
+    //
+    // Ordered BEFORE `build_inplace_resume_command` so a refusal costs nothing
+    // and, like the other two sites, cannot half-provision the pane.
+    if let Err(msg) = trusty_mpm::core::instruction_pipeline::refresh_compiled_prompt(&cwd) {
+        return InPlaceOutcome::Result(Err(anyhow::anyhow!("{msg}")));
+    }
+
     let resume = match trusty_mpm::runtime::build_inplace_resume_command(
         &cwd,
         record.claude_session_id.as_deref(),
--- a/crates/trusty-mpm/src/bin/tm/commands/install.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/install.rs
@@ -1,8 +1,9 @@
 //! `install` command handler and framework-artifact helpers.
 //!
 //! Why: the install handler is self-contained — it writes bundled artifacts,
-//! assembles the PM prompt, deploys agents and skills, and wires Claude Code
-//! hooks — and benefits from its own file so it stays reviewable.
+//! deploys agents and skills, and wires Claude Code hooks — and benefits from
+//! its own file so it stays reviewable. (#4752: it no longer assembles a PM
+//! prompt; that is per-project and written at launch.)
 //! What: `install`, `install_claude_hooks` / `install_claude_hooks_at`,
 //! `mpm_hook_additions`, `install_to`, `install_one`, `deploy_report_lines`,
 //! `reset_report_lines`, `skill_report_lines`, `remove_global_trusty_mpm_hooks`,
@@ -22,10 +23,10 @@
 /// Why: a fresh machine has no `~/.trusty-mpm/framework/`; `trusty-mpm install`
 /// writes the compile-time-embedded artifacts (optimizer policy, framework
 /// instructions, placeholder agent/skill) so the daemon has a working policy
-/// and launchers have instructions to point sessions at. It also assembles
-/// the full PM system prompt (overwriting the bundle stub — fixes #383),
-/// deploys composed agents and skills, and wires MPM lifecycle hooks into
-/// every Claude Code settings file. All edits are idempotent.
+/// and launchers have a working framework tree. It deploys composed agents and
+/// skills and wires MPM lifecycle hooks into every Claude Code settings file.
+/// All edits are idempotent. #4752: it no longer assembles a PM system prompt —
+/// the compiled prompt is per-project and written by the launch path.
 ///
 /// `reset_agents` (issue #2504) is the explicit reconciliation path for
 /// composed agent files a normal deploy conservatively skips because they
@@ -50,17 +51,18 @@
 /// reset, each gated by that workspace's OWN resolved harness plan (so an
 /// agent one project's manifest excludes is never resurrected there — the
 /// #2462 cross-warning). Only takes effect when `reset_agents` is `Some`.
-/// What: resolves [`FrameworkPaths::default`], calls [`install_to`] then
-/// [`install_system_prompt_to`] to write the real assembled PM prompt,
-/// deploys agents and skills, optionally resets the requested agent scope
-/// (user-level, then every intact session workspace when requested), then
-/// calls [`install_claude_hooks`].
+/// What: resolves [`FrameworkPaths::default`], calls [`install_to`], removes a
+/// stale pre-#4752 `instructions/INSTRUCTIONS.md`, deploys agents and skills,
+/// optionally resets the requested agent scope (user-level, then every intact
+/// session workspace when requested), then calls [`install_claude_hooks`].
+/// #4752: it no longer writes a compiled PM prompt — that is per-project and
+/// written at launch (`instruction_pipeline::compiled_prompt_path`).
 /// Test: `install_writes_all_artifacts`,
 /// `overwrite_artifact_refreshes_modified_file_without_force`,
 /// `seed_once_artifact_is_not_clobbered_without_force`,
 /// `seed_once_artifact_force_resets_to_shipped_default`,
 /// `install_claude_hooks_is_idempotent`,
-/// `instruction_pipeline::install_system_prompt_to_writes_assembled`.
+/// `remove_stale_bundled_instructions_deletes_a_leftover`.
 pub(crate) async fn install(
     force: bool,
     reset_agents: Option<Vec<String>>,
--- a/crates/trusty-mpm/src/bin/tm/commands/install.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/install.rs
@@ -77,19 +77,12 @@ pub(crate) async fn install(
         println!("  {line}");
     }
 
-    // #4752: the assembled PM prompt lands on its OWN path — nothing else
-    // writes `framework/INSTRUCTIONS-COMPILED.md`. It used to overwrite
-    // `instructions/INSTRUCTIONS.md`, which a bundled stub also targeted (#383)
-    // and which `build_instructions` reads back as a pipeline INPUT. A session
-    // launch rewrites this same file with the project-RESOLVED prompt before
-    // spawning `claude` (see `session_launch::prepare_session`); this install
-    // write is the bundled compilation, current until the next launch.
-    match trusty_mpm::core::instruction_pipeline::install_system_prompt_to(
-        &paths.instructions_compiled(),
-    ) {
-        Ok(()) => println!("  \u{2713} INSTRUCTIONS-COMPILED.md (assembled)"),
-        Err(e) => eprintln!("warning: failed to assemble system prompt: {e:#}"),
-    }
+    // #4752: `tm install` no longer writes a compiled PM prompt at all. The
+    // compiled prompt is the text ONE project's session runs with, and install
+    // has no project — writing the bundled assembly to a global path is exactly
+    // the "wrong content kind on a shared path" collision this issue closes.
+    // Each launch writes its own project's copy (see
+    // `instruction_pipeline::compiled_prompt_path`).
 
     // #4752: a pre-#4752 install left the compiled prompt at
     // `instructions/INSTRUCTIONS.md`. Nothing writes that path now, but
--- a/crates/trusty-mpm/src/bin/tm/commands/install.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/install.rs
@@ -77,12 +77,31 @@ pub(crate) async fn install(
         println!("  {line}");
     }
 
-    // Overwrite the bundle stub with the fully assembled PM prompt (#383).
+    // #4752: the assembled PM prompt lands on its OWN path — nothing else
+    // writes `framework/INSTRUCTIONS-COMPILED.md`. It used to overwrite
+    // `instructions/INSTRUCTIONS.md`, which a bundled stub also targeted (#383)
+    // and which `build_instructions` reads back as a pipeline INPUT. A session
+    // launch rewrites this same file with the project-RESOLVED prompt before
+    // spawning `claude` (see `session_launch::prepare_session`); this install
+    // write is the bundled compilation, current until the next launch.
     match trusty_mpm::core::instruction_pipeline::install_system_prompt_to(
+        &paths.instructions_compiled(),
+    ) {
+        Ok(()) => println!("  \u{2713} INSTRUCTIONS-COMPILED.md (assembled)"),
+        Err(e) => eprintln!("warning: failed to assemble system prompt: {e:#}"),
+    }
+
+    // #4752: a pre-#4752 install left the compiled prompt at
+    // `instructions/INSTRUCTIONS.md`. Nothing writes that path now, but
+    // `build_instructions` still READS it — so an upgraded machine would fold a
+    // frozen copy of an old prompt into the pipeline forever. Remove it here
+    // rather than leaving a stale input no writer can refresh.
+    match trusty_mpm::core::instruction_pipeline::remove_stale_bundled_instructions(
         &paths.framework_instructions_path(),
     ) {
-        Ok(()) => println!("  \u{2713} instructions/INSTRUCTIONS.md (assembled)"),
-        Err(e) => eprintln!("warning: failed to assemble system prompt: {e:#}"),
+        Ok(true) => println!("  \u{2717} instructions/INSTRUCTIONS.md (stale, removed)"),
+        Ok(false) => {}
+        Err(e) => eprintln!("warning: failed to remove stale instructions/INSTRUCTIONS.md: {e:#}"),
     }
 
     // #4409: bundled agents deploy ONLY into the tm-managed config-dir tier.
--- a/crates/trusty-mpm/src/bin/tm/commands/launch.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/launch.rs
@@ -207,6 +207,12 @@ pub(crate) async fn launch(
                     );
                 }
             }
+            // #4752: a compiled-prompt write failure is FATAL — refuse the
+            // launch rather than start a session whose instructions cannot be
+            // inspected. Every other prep failure stays non-fatal (#2149).
+            Err(e) if e.is_fatal() => {
+                anyhow::bail!("{e}");
+            }
             Err(e) => {
                 tracing::warn!(
                     "session prep failed for worktree {} (non-fatal): {e}",
@@ -275,16 +281,13 @@ pub(crate) async fn launch(
         }
     };
 
-    // 10. Regenerate `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` from
-    //     the bundled assets so the on-disk system prompt always reflects the
-    //     current trusty-mpm build.
-    let instructions_path = match trusty_mpm::core::instruction_pipeline::install_system_prompt() {
-        Ok(path) => Some(path),
-        Err(err) => {
-            eprintln!("warning: failed to install system prompt: {err}");
-            None
-        }
-    };
+    // 10. (#4752) The former regeneration of
+    //     `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` is gone. Its
+    //     result was assigned and then immediately discarded (`let _ =
+    //     instructions_path`), so it only ever wrote a file no launch read —
+    //     and that path is now retired: `tm install` deletes a stale copy and
+    //     the compiled prompt is per-project. The prompt this session actually
+    //     receives is built below and written by `prepare_isolated_session`.
 
     // Resolve the PM model (config > frontmatter > default).
     let mpm_cfg = trusty_mpm::core::config::MpmConfig::load_default();
@@ -320,7 +323,6 @@ pub(crate) async fn launch(
         prompt_path.as_deref(),
         Some(&managed_path),
     );
-    let _ = instructions_path;
 
     // 12. Create a detached tmux session rooted at the MANAGED clone directory.
     //     #2398: routes through `core::tmux::create_managed_session`, the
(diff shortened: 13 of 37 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review, code-only against the stated requirements. The distinct-path half (requirements 1, 3, 5, 6) is done cleanly and the new tests are honestly built. The **ordering half (requirement 2) is not delivered** — the write is ordered before the spawn but does not block it on any production path, and one spawn path never runs it at all. Three HIGH findings, no CRITICAL.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/session_launch/mod.rs` | `prepare_session_inner` (compiled-prompt write) | The fatal `?` sits **mid-pipeline**, immediately after the stash write. A failed compiled write now aborts everything downstream: `write_output_style`, `write_status_line`, `write_project_hooks`, `inject_trusty_memory_mcp`, `inject_trusty_search_mcp`, `inject_trusty_mpm_mcp`, `inject_trusty_review_mcp`, and the trust pre-seed — and the session **still launches** (see next finding). The last two injectors are the documented defense against the #3918/#3950 MCP name-squatting exploit; skipping them on a workspace that already carries the name-based `enabledMcpjsonServers` approval leaves a squatted `.mcp.json` entry free to execute. This failure mode did not exist before this PR — the compiled write was install-time only. | Move the `write_compiled_prompt_to` call to the **end** of `prepare_session_inner`, immediately before the `PrepReport` construction. Still before the return, still fatal, nothing stripped. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/session_launch/mod.rs` | `prepare_session` doc "ORDERING CONTRACT (#4752)"; spec §10.3 | The claim *"Every caller spawns `claude` only after `prepare_session` returns `Ok`"* is false at **every** call site that goes on to spawn. All of them swallow `Err` and launch anyway (#2149 made prep non-fatal at the call sites): `commands/launch.rs::connect`, the managed-clone `prepare_isolated_session` block in `commands/launch.rs`, `commands/meta/launch.rs`, `commands/session/start.rs::start_session_in_place`, `client/http_client/session_connect.rs::DaemonClient::launch_session`, `daemon/managed_routes/lifecycle.rs::prepare_inproject_session`, `provisioner/workspace.rs`. The **only** caller that propagates is `core/standalone/load.rs::run_prepare_session` — and that one does not spawn `claude` at all. Net effect of fatality: it newly hard-fails `tm load <alias>` on a home-tier write error, while every actual launch ignores it. It buys nothing and costs the finding above. | Correct the doc comment and §10.3 to state what is true ("written before `prepare_session` returns; callers treat prep failure as non-fatal"), or make at least one launch caller actually abort. Do not ship the contract as written. | Fix here |
| HIGH | `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs` | `resume_managed` → `runtime/claude_code.rs::ClaudeCodeAdapter::spawn_resume` | **Named path where `claude` spawns with a stale or absent `INSTRUCTIONS-COMPILED.md`.** `resume_managed` calls `resume_self_heal`, `ensure_status_line` and `ensure_deployment_complete`, then `adapter.spawn_resume` — no `prepare_session*` on the healthy path. `spawn_resume` composes its own prompt via `build_prompt_file(cwd)` and passes it as `--append-system-prompt-file`. So every resume / guided-resume / crash-recovery launch runs a prompt that was never written to the compiled path. `ensure_deployment_complete` only reaches `prepare_session_with_repo_url` when it judges the deployment incomplete, so it is not a guarantee. | Write the compiled prompt inside `build_prompt_file` — the one seam every spawn path already funnels through — or call the prep seam on resume. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/install.rs`, `crates/trusty-mpm/src/core/session_launch/mod.rs` | `install_system_prompt_to` vs `prepare_session_inner` | **#383 recurrence check: the collision is moved, not eliminated.** The stub/compiled collision is genuinely gone. But one *global* path now has two writers with **different content semantics** — `tm install` writes the **bundled** assembly (`assemble_system_prompt`, no project overrides); every launch writes the **project-resolved** prompt. `FrameworkPaths::for_managed_workspace` keeps `framework` at the real `$HOME/.trusty-mpm`, so every concurrent session in every project overwrites the same file. In a repo whose normal mode is many parallel worktree sessions, the file answers "the most recent prepare or install", never "what is *my* session running" — the exact question §10.1 says it exists to answer. The per-session truth is still `<project>/.trusty-mpm/last-instructions.md`, which already existed. | Scope the compiled artifact per project, or stop `tm install` writing project-semantics-free content to it, or document it as "last prepared session" and drop the determinism claim. Owner call. | Parent |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/tests.rs` | `prepare_session_fails_when_the_compiled_prompt_cannot_be_written` | The fixtures themselves are good — the pre-seeded `STALE` sentinel in `prepare_session_writes_the_compiled_prompt_before_returning` and the planted directory here both genuinely fail if the write is removed or made best-effort, and the `PrepError::Io { path }` assertion pins the specific path so an unrelated earlier IO failure cannot green it. What they do **not** prove is what the comment claims: *"the ordering requirement is that the write BLOCKS the launch"*. They pin write-before-**return**. Nothing pins write-before-**spawn**, and that gap is exactly where the guarantee fails (finding 2). | Add a call-site test asserting a launch path aborts, or soften the comment to what is actually pinned. | Fix here |
| LOW | `crates/trusty-mpm/src/assets/instructions/sections/README.md` | Findings list | Old finding #2 was renumbered to `1.` but the following item is still numbered `3.` — the raw list reads 1, 3. Separately, the retired finding was **deleted** rather than moved to the file's own `### Retired findings` section, which sits directly below and is the file's stated convention ("Recorded rather than smoothed over"). The record that the collision existed and was closed by #4752 is now gone from the shipped asset. | Renumber, and add a `### Retired findings` entry pointing at #4752. | Fix here |
| LOW | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | `install_system_prompt` | Still has zero production callers (the PR body says so), yet its doc was rewritten and expanded. Dead public surface on a lib crate. | Leave, or delete it with its `#[serial]` test. | Promote |

## Verified clean — checked and deliberately not flagged

- **Requirement 1** — `FrameworkPaths::instructions_compiled` resolves `<framework>/INSTRUCTIONS-COMPILED.md`, directly under `framework/`, pinned by `instructions_compiled_is_directly_under_framework` and `instructions_compiled_is_not_the_framework_instructions_path`. Correct on the managed/isolated paths too: `for_managed_project` overrides only `claude_agents`/`claude_skills`, so `framework` stays at the shared root.
- **Requirement 3** — `no_bundled_artifact_targets_the_compiled_prompt_path` scans `bundle::ALL` by file name at any depth, so a future re-added stub turns it red. `bundle_install_pass_never_touches_the_compiled_prompt` uses a survive-the-sentinel assertion, not an absence check — correct fixture design.
- **Stale-file cleanup** — `remove_stale_bundled_instructions` is safe. `framework/instructions/INSTRUCTIONS.md` has always been framework-owned and overwritten on every install, so no user-authored content is at risk; `NotFound` → `Ok(false)` is right; the error is reported via `eprintln!`, not swallowed. Confirmed the only consumer is `build_instructions` → `PipelineOutput::merged`, which `commands/session.rs` explicitly no longer returns, and `instructions_loaded` has no production reader. No `tm doctor` check depends on the file (`doctor.rs`'s `instructions` check is about the project `last-instructions.md`).
- **Requirement 4** — read `non-overridable-rules.md` end to end, not just the removed table. No remaining place describes `.trusty-mpm/` files as a live mechanism: the trigger-phrase list routes every case to `CLAUDE.md`, and the retired-files paragraph says "Never create one." Consistent with the README "Retired — do not bring these back" table. `assets/sm_instructions/BASE_SM.md` carries the same "override files" sentence but describes the separate, genuinely-live `~/.trusty-mpm/sm/` mechanism — correctly left alone.
- **Requirement 6** — both fragments present (`4752-…`, `4753-…`), directly in `changelog.d/`, category on line 1, clickable issue links. #4753 is `refs`, not `closes`, deliberately and with the reasoning in the PR body; the defect that issue describes does not exist on `main`. Owner call, not a code defect.
- **SLD / conventions** — every new public item carries Why/What/Test and all named tests exist (doc-pointer lint green). All three `pm-prompt-*.md` goldens regenerated; no fourth golden carries the changed line.

1. Move the `write_compiled_prompt_to` call to the end of `prepare_session_inner`, so a failed write cannot strip output-style, statusline, hooks and the four MCP injections from a session that launches anyway.
2. Correct the `prepare_session` "ORDERING CONTRACT" doc and spec §10.3 — no launch caller blocks on `Err`.
3. Cover `resume_managed` / `spawn_resume`, or state in §10.3 that resume is out of scope for the ordering guarantee.

## Notes

The `Test` CI job was still pending at review time; every other check was green. The merge gate is the merger's, not this review's.

bobmatnyc:
## Review round 2 — HIGH 1, HIGH 3 and LOW fixed; HIGH 2 and MEDIUM 4 proposals below

All four findings verified against the code before touching anything. **Every one was correct.** CI: **24 pass, 3 skipping, 0 failures.**

### HIGH 1 — security regression: FIXED

Confirmed by line order: the write sat at line 840, upstream of `write_output_style` (870), `write_project_hooks` (887), all four injectors (917/940/970/977), and `preseed_workspace_trust_home` (1064).

Moved to the **last** step of `prepare_session_inner`, immediately before `Ok(PrepReport)`. Nothing below it can now be skipped, and a comment marks the position as load-bearing.

New guard `compiled_write_failure_does_not_skip_the_mcp_injectors` forces the compiled write to fail and asserts the security-relevant side effects still happened. **Fail-before proof** — I moved the write back to the old position and re-ran:

```
test core::session_launch::tests::compiled_write_failure_does_not_skip_the_mcp_injectors ... FAILED
panicked at tests.rs:380:
.mcp.json must be written even when the compiled prompt write fails —
the MCP injectors must not sit downstream of it
```

### HIGH 3 — resume never covered: FIXED

Confirmed: `resume_managed` runs `resume_self_heal` → `ensure_status_line` → `ensure_deployment_complete` → `spawn_resume`, with no `prepare_session*`.

Fixed at `runtime::claude_code::build_prompt_file` — the one seam **all three** spawn paths share (`spawn` :895, isolated :1049, `spawn_resume` :1150). It now refreshes the compiled file from the exact bytes it hands to `--append-system-prompt-file`. That covers resume, guided-resume and crash-recovery, and makes the artifact match the running prompt *by construction*.

Deliberately **best-effort** there, applying HIGH 1's lesson: this is an inspection artifact, and a failed write must never cost the session its real prompt. Pinned by `build_prompt_file_compiled_write_failure_does_not_block_the_spawn`.

**Fail-before proof:**

```
test runtime::claude_code::tests::build_prompt_file_refreshes_the_compiled_prompt ... FAILED
panicked at claude_code_tests.rs:927: assertion `left != right` failed:
the spawn seam must refresh a stale compiled prompt — resume paths depend on
this, since they never run prepare_session
```

**Bonus catch:** the new refresh resolves through `$HOME`, so the pre-existing `build_prompt_file_writes_resolved_prompt_for_project` would have started writing into the developer's real `~/.trusty-mpm` — the #2459/#2460/#2461 hazard. Paired it with `HomeGuard` + `#[serial]`.

### HIGH 2 — false claim: REMOVED. Proposal below.

Verified all seven sites. Every one logs and continues; three say so in their own words — `"session prep failed (non-fatal)"`, `"best-effort"`, `"warning: session preparation failed"`. `run_prepare_session` propagates but spawns no `claude`.

The false sentence is **gone** from both the doc comment and spec §10.3. Both now enumerate the seven non-fatal callers by name and state the real guarantee: current *by construction*, because the spawn seam writes from the string it passes — not because anything blocks.

**Proposal: do not make the spawn conditional.** Reversing #2149 at seven call sites would mean a transient disk error prevents sessions launching at all — strictly worse than a stale inspection file, and it re-runs the same inverted priority that caused HIGH 1. The by-construction guarantee from HIGH 3's fix delivers the owner's intent (the file always matches the running session) without that cost. Recorded as §10.6(a).

### MEDIUM 4 — collision moved, not died. Proposal below.

Confirmed: `for_managed_project_keeps_framework_source_at_managed_root` proves `framework` stays at the shared root, and `for_managed_workspace` resolves from the real `$HOME`. One global file; concurrent sessions across projects overwrite each other. It answers "most recent prepare or install on this machine", not "what is my session running".

**Proposal: scope it per project** — `framework/compiled/<project-slug>.md`. That is the only option that actually answers the artifact's stated question. Alternatives: per-session, or keep it global and document it honestly as "most recent launch" while pointing at `.trusty-mpm/last-instructions.md` (which already answers the per-session question correctly, and is written on the same paths).

Not built — it changes the artifact's contract, and per instruction I am reporting first. Recorded as §10.6(b).

### LOW — FIXED

Finding #1 moved into `sections/README.md`'s own `### Retired findings` section with its resolution (both halves: the already-removed stub, and the output/input path split). Remaining findings renumbered 1, 2.

### Gates — rung 3, raw

```
$ cargo fmt --check
fmt exit=0

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 18.60s
clippy exit=0

$ cargo test -p trusty-mpm
TEST_EXIT=0
TOTAL passed=7413 failed=0 ignored=17

line-cap: measured 3622 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 54 spec doc(s) + 3042 code file(s); 0 error(s), 0 warning(s)

changelog-fragment gate: scanned 158 changed path(s); all 1 crate(s) with source changes are recorded.
```

No coverage deleted, ignored, cfg-gated, excluded, or narrowed. No pre-existing failures encountered.

bobmatnyc:
## code-critic — Round 3 adversarial review

## Verdict: BLOCK

Head SHA `72c77752`. Reviewed in an isolated worktree off `origin/fix/4752-instructions-compiled-path`.

The compiled-prompt work itself is close. It is blocked by an unrelated 48,011-line accidental commit that every CI gate is blind to, and it still carries two guarantees that the code does not honour.

Local gate (rung 3/4, `-p trusty-mpm`):

```
test result: ok. 4356 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 45.75s
test result: ok. 1393 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 3.69s
```

CI on this SHA: all 23 checks green (one-shot `gh pr checks 4759`).

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/.claude/**`, `crates/trusty-mpm/.trusty-mpm/last-instructions.md`, `crates/trusty-mpm/CLAUDE.md` | commit `58678fab` | 140 files / 48,011 lines of machine-local harness state committed, including an executable `.claude/settings.json` (plus a `.settings.json.bak`) whose hooks are hardcoded `/Users/masa/.cargo/bin/...` paths, and 52 project-tier skills that shadow the bundled roster | `git rm -r --cached` all three trees, drop the commit or amend it out, and extend `.gitignore` (`.trusty-mpm/*` is root-anchored, so nested crates are not covered) | Fix here |
| HIGH | `crates/trusty-mpm/src/runtime/claude_code.rs` | 353–362 | "every path that reaches it has already had a fatal compiled write succeed" is false — the in-place relaunch reaches `build_prompt_file` with no fatal write anywhere upstream | Either make the write fatal on that path, add a fatal refresh in `run_inplace_relaunch`, or correct the claim and record the gap | Fix here |
| HIGH | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 385–392 | ORDERING CONTRACT claims a started session "always has a compiled prompt on disk"; the non-fatal `PrepError::Io` early return at 871/876 fires first under the same disk-full/permission condition | Make the `.trusty-mpm/` stash writes fatal too, or narrow the contract to what actually holds | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 227, 233, 1189 | `PrepReport::compiled_prompt` documents the path as `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` and calls it "framework-level" — it is project-local | Replace with `<project_dir>/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`; drop "framework-level" | Fix here |
| MEDIUM | `crates/trusty-mpm/changelog.d/4752-instructions-compiled-path.md` | 3 | Fragment asserts the global `~/.trusty-mpm/framework/...` path and contradicts its own sibling fragment; renders verbatim into `CHANGELOG.md` at release | Correct the path or fold the three `4752-*` fragments into one | Fix here |
| MEDIUM | `crates/trusty-mpm/src/assets/instructions/sections/README.md` | 196 | Shipped asset cites `FrameworkPaths::instructions_compiled`, a symbol this PR deletes, and gives the global path | Point at `instruction_pipeline::compiled_prompt_path` and the project-local path | Fix here |
| MEDIUM | `docs/trusty-mpm/spec/ARCHITECTURE.md`, `docs/trusty-mpm/spec/COMPONENTS.md`, `docs/specs/mpm-behavior-conformance.md` | 335–338, 204, 94 | Three specs still document `install_system_prompt()` and the "writes it, `build_system_prompt()` reads it back" mechanism, both deleted here | Update or mark superseded by §10 | Parent |
| LOW | `crates/trusty-mpm/src/core/deploy_validate/mod.rs` | 371–375 | `validate_and_repair` swallows a fatal `CompiledPrompt` into `repair_error` while still reporting `repaired: true` | Report `repaired: false` when `e.is_fatal()` | Promote |

## Required Changes

1. **Remove the accidental commit.** `crates/trusty-mpm/.claude/` (52 skills, 3 output styles, `settings.json`, `settings.json.bak`), `crates/trusty-mpm/.trusty-mpm/last-instructions.md`, and `crates/trusty-mpm/CLAUDE.md` are machine-local output from running `tm` inside the crate directory. Three things make this blocking rather than noise:
   - `crates/trusty-mpm/.claude/settings.json` is an **executable** project-tier config — `PreToolUse`/`PostToolUse`/`SessionStart` hooks invoking `/Users/masa/.cargo/bin/trusty-mpm hook`, `/Users/masa/.cargo/bin/tm hook --pm-guard`, `trusty-memory inbox-check`. Shipping a hook config in the repo is the same class as the `.mcp.json` clone auto-trust hole, and the absolute paths are broken on every other machine.
   - The repo's own `.gitignore` bans `**/.claude/agents/` with the comment *"A tracked copy under any crate's .claude/agents/ SHADOWS the bundled agent"*. The 52 committed skills are the exact analogue — this review session loaded them as project-tier skills the moment the worktree was created.
   - `settings.json.bak` is a backup artifact. Nothing here is in scope for #4752.

   Note the root `.gitignore` uses `.trusty-mpm/*` (contains a slash → anchored to repo root), so nested-crate `.trusty-mpm/` state is not ignored, and `!.claude/` un-ignores the `.claude/` directory. No CI gate catches this: the 500-line cap only scans `.rs`.

2. **Fix or retract the `build_prompt_file` justification.** `guided_inplace.rs:560` (`run_inplace_relaunch`, the bare-`tm`-in-a-managed-pane path) calls `runtime::build_inplace_resume_command` → `build_prompt_file`. It calls neither `prepare_session*` nor `resume_managed`; the only daemon call is the `reactivate` route, which does no prompt work at all (`daemon/managed_routes/reactivate.rs` — grep for `prepare_session|resume_managed|refresh_compiled` returns nothing). So on that path the best-effort write is the **only** write: a failure warns and the process `exec`s anyway. A resumed session can run with a stale or missing compiled prompt where a fresh launch would refuse. That is the round-3 asymmetry, unfixed. The same false claim appears in the spec at §10.3 (`docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md:1231`), and the fatal-write table above it lists only two paths.

3. **Fix or narrow the ORDERING CONTRACT.** `prepare_session_inner` early-returns `PrepError::Io` at `mod.rs:871` (`create_dir_all(<project>/.trusty-mpm)`) and `mod.rs:876` (`write(last-instructions.md)`). `is_fatal()` classifies both as non-fatal — `deploy_and_io_failures_stay_non_fatal` pins exactly that, using `/p/.trusty-mpm/last-instructions.md` as its example. Under disk-full or an unwritable `.trusty-mpm/` — the two conditions `compiled_prompt_failure_message` itself names — the stash write fails first, non-fatally, and the caller launches. The session then starts with no compiled prompt **and** with the MCP injectors skipped, since both `?`s sit upstream of them. The change makes a write into `.trusty-mpm/framework/` fatal while a write into `.trusty-mpm/` stays non-fatal; that is incoherent, and it falsifies "a session that starts always has a compiled prompt on disk."

## Claim-by-claim

| # | Claim | Result |
|---|---|---|
| 1 | Fatal write is last step; regression test fails on reorder | **VERIFIED.** Write is at `mod.rs:1203-1208`, after all four injectors and the trust pre-seed. `compiled_write_failure_does_not_skip_the_mcp_injectors` (`tests.rs:411`) forces the write to fail and asserts `.mcp.json` still carries `trusty-mpm`/`trusty-review` — it genuinely fails on reorder. **Caveat:** no early return exists *between* the injectors and the write, but two non-fatal `?` returns remain *upstream* of the injectors (HIGH 3). |
| 2 | Path moved; `FrameworkPaths::instructions_compiled()` deleted | **VERIFIED.** `git grep instructions_compiled` returns only comments and one stale shipped-asset reference (MEDIUM). `paths.rs:325-330` documents the deliberate absence. No caller resolves a shared path. |
| 3 | Seven call sites gated on `is_fatal()` | **VERIFIED — exactly seven:** `launch.rs:213`, `launch.rs:470`, `meta/launch.rs:194`, `session/start.rs:166`, `session_connect.rs:51`, `session_prep.rs:69`, `workspace.rs:942`. Each `Err(e) if e.is_fatal()` arm precedes the catch-all and each refuses (`bail!` / `Err(ProvisionError::…)` / propagated `Result`). Two other `prepare_session*` callers are ungated but safe: `standalone/load.rs:250` propagates every error; `deploy_validate/mod.rs:371` is a repair path (see LOW). |
| 4 | `build_prompt_file` best-effort justification is sound | **REFUTED.** The justification is a rationalization on the in-place relaunch path — see Required Change 2. |
| 5 | `tm install` writes no compiled prompt; stale deletion correct | **VERIFIED.** `install.rs:82-99` writes nothing and deletes `framework_instructions_path()`. Nothing depends on the old write: #4286 split A already removed the entry from `bundle::ALL`, `build_system_prompt()` now composes in memory, and `build_instructions` treats absence as normal. Deleting a read-only input no writer can refresh is correct. |
| 6 | Refused launch names the path | **VERIFIED.** `compiled_prompt_failure_message` (`instruction_pipeline.rs:335-343`) interpolates `path.display()`, wired into `PrepError::CompiledPrompt`'s `#[error]` and the resume message; pinned by three tests. |
| 7 | `lifecycle.rs` split is behavior-preserving; allowlist not loosened | **VERIFIED.** `prepare_inproject_session` moved verbatim into `session_prep.rs` with one intended change (returns `Result`, propagated at `lifecycle.rs:977` and `launch_on_main.rs:139`). `.line-cap-allowlist.tsv` has exactly one edit: `lifecycle.rs 845 → 818`. `daemon/api.rs` stays at 1272; no other row moved. |
| 8 | Spec §10.3 now true; §10.6 records both rulings | **PARTIALLY VERIFIED.** §10.6 (a) and (b) are recorded correctly. §10.3 still asserts the false `build_prompt_file` backstop (line 1231) and its fatal-write table omits the in-place relaunch path. SLD lint passes; three other specs are now stale (MEDIUM). |
| 9 | Changelog fragment present | **VERIFIED with a defect.** Three `4752-*` fragments plus one `4753-*`; each carries one category, so the CI gate is green. But `4752-instructions-compiled-path.md` states the global path and contradicts its sibling (MEDIUM). |

## Notes

- Test quality is good. The fixtures are the right shape throughout: stale sentinels rather than existence checks, and a directory planted at the exact target so only the write under test fails. `compiled_write_failure_does_not_skip_the_mcp_injectors` is a real guard, not a name.
- New project-local `<project>/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` is not in `scaffold_gitignore.rs`'s `SCAFFOLD_PATHS` (which covers only `.claude/{agents,skills,output-styles}/`). It lands beside the already-unignored `last-instructions.md`, so it is not a new class of problem — but this PR's own accidental commit is what that gap looks like in practice. Not filed separately.
- Not flagged: the `#[serial]` + `HomeGuard` additions in `claude_code_tests.rs` are correct and necessary once the write resolves from `$HOME`.

bobmatnyc:
## `code-critic` round 4 — Verdict: WARN

Reviewed **`0a2839bd`** (`fix/4752-instructions-compiled-path`), base `53aea0b8`. The engineer is pushing a newer commit implementing the owner's `:849` ruling; this verdict does not cover it.

### Gates (run in `.claude/worktrees/qa-4759-r4`)

| Gate | Result |
|---|---|
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | exit 0 |
| `cargo test -p trusty-mpm` | `7454 passed; 0 failed; 17 ignored` (exit 0) |
| `bash scripts/check_line_cap.sh` | `3628 tracked .rs file(s); 7 allowlisted, 0 violations — OK` |
| `bash scripts/check_sld.sh` | `55 spec doc(s) + 3045 code file(s); 0 error(s), 0 warning(s)` |
| `gh pr checks 4759` (one-shot) | 23 pass, `Test` still pending |

No `#[ignore]`, `cfg`-gate or `--exclude` was added by the diff (`git diff 53aea0b8 0a2839bd | grep '^+.*#\[ignore\]'` → empty). The 17 ignored are the pre-existing set.

---

### Claim verification

**History rewrite — VERIFIED, all three sub-claims.**

The `backup/fix-4752-pre-r4` ref does **not** exist locally or on origin (`git ls-remote origin 'refs/heads/backup/*'` → empty). The pre-rewrite tip `72c77752` survives only because worktree `qa-4759-r3` still holds it detached. Nothing is lost today, but the stated safety net isn't there.

Working from `72c77752` (pre) vs `d4c89fdf` (its rewritten equivalent):

1. *Nothing legitimate lost.* Pre-rewrite change set vs its base: 172 files. Post: 32. The 140 dropped paths are **all** under `crates/trusty-mpm/.claude/`, `crates/trusty-mpm/.trusty-mpm/`, plus `crates/trusty-mpm/CLAUDE.md`. Every one was status `A` (pure addition) and **none** is tracked on `origin/main`. Confirmed harness output, not source.
2. *Junk gone from EVERY commit.* Walked all six trees. The only `.claude`-shaped paths surviving in any tree are the 7 pre-existing on main (`crates/trusty-{agents,analyze,git-analytics,search}/CLAUDE.md`, `crates/trusty-search/.claude/{settings.json,commands/health.md,scheduled_tasks.lock}`). `crates/trusty-mpm/.claude/settings.json` count is **0 in all six commits**.
3. *No unintended content change.* `git diff 72c77752 d4c89fdf -- <the 32 surviving files>` → **empty**.

Not in the brief but worth recording: the same operation also moved the base from `d7b07838` to `53aea0b8`. That accounts for the 55 non-junk file deltas in a naive tip-to-tip diff.

**A. `.gitignore` — VERIFIED, with residual holes.**

`git check-ignore -q --no-index`, exit codes:

```
not ignored .claude/settings.json              <- root re-include preserved
IGNORED     crates/trusty-mpm/.claude/settings.json
IGNORED     crates/trusty-mpm/.claude/skills/tm/SKILL.md
IGNORED     crates/trusty-mpm/.trusty-mpm/last-instructions.md
```

Both halves hold. The `crates/*/` scoping does leave equivalent holes — see LOW-2.

**B. Read/write claim — VERIFIED. Consolidation claim — REFUTED.**

Nothing in production reads `INSTRUCTIONS-COMPILED.md`; every hit is a writer or a doc/test reference. The reasoning stands.

The consolidation does **not** hold as stated. `prepare_session` writes the compiled prompt at `session_launch/mod.rs:1229` via `write_compiled_prompt_to` with its own `resolved_prompt`, **not** via `refresh_compiled_prompt`. Two of three paths route through the new entry point, not three — see MEDIUM-2. (The divergence is not cosmetic: `refresh_compiled_prompt` composes with style `None`, `prepare_session` with `effective_style`.)

**C. Ordering contract — VERIFIED by re-run mutation.**

Reverted `mod.rs:895-904` to the pre-fix `?`, ran the guard:

```
thread 'core::session_launch::tests::stash_write_failure_does_not_skip_the_mcp_injectors' panicked at
crates/trusty-mpm/src/core/session_launch/tests.rs:493:10:
a failed stash write must NOT refuse the launch (#2149) nor abort preparation:
Io { path: ".../.trusty-mpm/last-instructions.md", source: Os { code: 21, kind: IsADirectory, ... } }
test ... FAILED
```

Restored; tree clean. The guard fails for the right reason and is not a pass-both-ways test.

**D. `:849` — factual finding (design question settled by owner ruling, not argued here).**

**Yes — at `0a2839bd` a session can start with no instructions having been written.** Proved empirically with a throwaway probe (planted a directory at `<project>/CLAUDE.md` so `load_or_create_claude_md` fails with `IsADirectory` rather than `NotFound`), then removed:

```
QA-PROBE err = instruction pipeline failed: io error for .../CLAUDE.md: Is a directory (os error 21)
QA-PROBE is_fatal      = false      <- callers print a warning and LAUNCH
QA-PROBE compiled exists = false    <- no compiled prompt on disk
QA-PROBE .mcp.json exists = false
```

`PrepError::Instructions` is non-fatal, so `commands/session/start.rs:167`, `commands/launch.rs:216`, `meta/launch.rs:195`, `session_connect.rs`, `session_prep.rs:69` and `workspace.rs:942` all fall through to the log-and-continue arm. The shipped-today behaviour is broken. Note this is pre-existing (`origin/main` has the same `build_instructions(&input)?` at its line 761), not introduced by this PR.

**E. The four round-3 MEDIUMs — 1 fixed, 3 partial.** See findings.

**F. Baseline failure — VERIFIED.**

`git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/` → **0** of 35 changed files. `crates/trusty-mpm-gui/ui/dist` is genuinely absent. Pre-existing environmental failure, not caused by this PR.

---

### Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 849 | `build_instructions(&input)?` returns a non-fatal `PrepError::Instructions`, so the caller launches a session whose instructions were never written (probe output above). | Make the condition "the instructions could not be written" fatal at this site too, per the owner ruling. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/claude_md_writer.rs` | 73 | `COMPILED_INSTRUCTIONS_PATH = "~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md"` — the global path this PR retired. `render_pointer` splices it into a project's `CLAUDE.md` as a human-facing "the prompt is written to X" blockquote, so once wired it will name a path nothing writes. Dormant only because `ensure_compiled_pointer` has no production caller (spec §10.5 defers it). | Either resolve it from `instruction_pipeline::compiled_prompt_path` (it needs a `project_dir`, so the constant has to become a function), or replace the constant with the project-relative form `<project>/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`. Update the three `claude_md_writer_tests.rs` assertions. | Fix here |
| MEDIUM | `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md` / `instruction_pipeline.rs` / `changelog.d/4759-inplace-relaunch-compiled-prompt.md` | 1213 / 381 / 5 | "All three now route through the single entry point `instruction_pipeline::refresh_compiled_prompt`" is false — `prepare_session` calls `write_compiled_prompt_to` directly at `mod.rs:1229`, and deliberately so (it must keep `effective_style`, which `refresh_compiled_prompt` hardcodes to `None`). Two sites route through it. | Say "the two resume-shaped paths route through `refresh_compiled_prompt`; `prepare_session` writes its own style-resolved text through the shared `write_compiled_prompt_to`", and state why. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 236 | `PrepReport::compiled_prompt` is documented "byte-identical to `Self::stash`". No longer guaranteed: the stash write now degrades to a warning (so `stash` may be absent or stale), and for managed spawns `runtime::claude_code::build_prompt_file:392` later rewrites the compiled file with a `None`-style composition that can differ from `prepare_session`'s `effective_style` one. | Drop the byte-identical claim, or qualify it to "when the stash write succeeded, and until the spawn seam refreshes it". | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/mod.rs` + spec §10.3 | 403-413 / 1249-1256 | The "WHAT THE CONTRACT DOES **NOT** SAY" / "The contract is not 'every started session has a compiled prompt'" narrowing has been overruled by the owner. | Delete both passages and restore the strong claim alongside the `:849` fix. | Fix here |
| LOW | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 1257 | `build_system_prompt`'s `Why:` still says the prompt is "assembled from bundled assets into `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md`" — contradicted by the `#4752:` paragraph seven lines below in the same doc block, which says that file is retired. | Rewrite the `Why:` sentence to describe in-memory composition. | Fix here |
| LOW | `.gitignore` | 71-72 | `crates/*/` scoping leaves the same accident reachable elsewhere. `git check-ignore` says NOT IGNORED for: `docs/.claude/settings.json`, `scripts/.claude/settings.json`, `crates/.claude/settings.json`, `crates/trusty-mpm/src/.claude/settings.json`, `crates/trusty-mpm/ui/.trusty-mpm/x.md`, and `crates/trusty-mpm/CLAUDE.md` — the last being one of the 140 that actually leaked. Also note `crates/trusty-search/.claude/` is now wholly ignored, so a genuinely new file there would be silently dropped. | `**/.claude/` + `**/.trusty-mpm/` with anchored re-includes (`!/.claude/settings.json`) reaches all of them without re-ignoring the root; the stated reason for avoiding `**/` (that it would re-ignore the root `settings.json`) is solvable with a leading slash on the negation. | Promote |
| LOW | — | — | The claimed backup ref `backup/fix-4752-pre-r4` does not exist locally or on `origin`. The pre-rewrite tip is reachable only because worktree `qa-4759-r3` holds `72c77752` detached. | Push the ref, or stop citing it as the safety net. | Parent |

1. Make the `:849` instruction-write failure fatal (owner ruling) and restore the strong contract in both `mod.rs:403-413` and spec §10.3.
2. Fix `claude_md_writer.rs:73` — it is the one remaining pointer to the retired global path in live code.
3. Correct the "all three route through `refresh_compiled_prompt`" claim in the spec, the `instruction_pipeline.rs:381` doc, and the `4759` changelog fragment.
4. Fix the `PrepReport::compiled_prompt` byte-identical claim and the `mod.rs:1257` stale `Why:`.

### Notes

- No CRITICAL findings. The `:849` gap is rated HIGH rather than CRITICAL because it is pre-existing on `origin/main` and requires an unreadable project `CLAUDE.md` or framework instructions path to reach.
- Per the coordinator's relay of the owner's position, I did not weigh the MCP-injector angle. For the record, the probe shows `.mcp.json` is also not written on that path — but the test applied is simply "can a session start without its instructions written", and it can.
- Adding a fatal at `:849` enforces the same condition as the existing one; it is not a second fatal error type.

bobmatnyc:
## code-critic — ROUND 5 (head `efb829af`)

## Verdict: APPROVE

Zero CRITICAL, zero HIGH. The fatal-refusal path is correct and the rename is clean. Findings below are MEDIUM/LOW documentation and blast-radius notes.

### Claim verification

| # | Claim | Result |
|---|---|---|
| 1 | New regression test is a real guard | **VERIFIED** (mutation run, below) |
| 2 | Rename blast radius clean | **VERIFIED, with one correction to the framing** |
| 3 | Error message correct at BOTH sites | **PARTIAL** — see MEDIUM-1 |
| 4 | `claude_md_writer.rs:84` fix + drift guard | **VERIFIED** (mutation run, below) |
| 5 | `effective_style` justifies the two-plus-one split | **VERIFIED, with a caveat** — see LOW-4 |
| 6 | Strong contract restored, no surviving narrowed assertion | **VERIFIED** |
| 7 | Injector framing removed from the fatality rationale | **VERIFIED** |
| 8 | Backup ref pushed | **VERIFIED** |

**1 — mutation run.** Reverted the round-5 classification (`PrepError::Instructions` → `PrepError::Io` at `session_launch/mod.rs:878`) and ran the test alone:

```
thread 'core::session_launch::tests::prepare_session_refuses_when_the_instructions_cannot_be_built'
panicked at crates/trusty-mpm/src/core/session_launch/tests.rs:492:5:
an instruction-build failure must be classified fatal, got Io { path: ".../CLAUDE.md",
source: Os { code: 21, kind: IsADirectory, message: "Is a directory" } }
```

Fails at exactly the `is_fatal` assertion the engineer named. Real guard. Mutation reverted.

**2 — rename blast radius.** No match arm became unreachable or vacuous. `PrepError::Instructions` is matched in exactly two places outside its own module (both tests); the `map_err(|e| match e { PipelineError::Io {..} => … })` at `mod.rs:874-878` is exhaustive over a one-variant enum, so a future `PipelineError` variant is a compile error, not a silent fall-through. No caller parses the old Display string (`grep "instruction pipeline failed"` → 0 hits). All seven `is_fatal()` sites present: `launch.rs:213`, `launch.rs:470`, `meta/launch.rs:194`, `session/start.rs:166`, `workspace.rs:942`, `session_prep.rs:69`, `session_connect.rs:51`.

**Correction to the framing:** the folded-in `Instructions(PipelineError)` was **not dead**. It was live — it was the variant `build_instructions(&input)?` produced via `#[from]`, and it was non-fatal. Folding it in is precisely the behavior change the PR intends, and the new test pins it. The code is right; the "dead variant" description is not.

**Ordering contract independently checked.** Between the two fatal sites (`mod.rs:880` and `mod.rs:1261`) there is **zero** `?`/`return Err` — verified by scanning every `?` in lines 880–1262. The claim "nothing that can fail in between returns early" is true as written.

**4 — drift guard mutation.** Changed `COMPILED_INSTRUCTIONS_PATH` to `.trusty-mpm/INSTRUCTIONS-COMPILED.md`:

```
panicked at crates/trusty-mpm/src/core/claude_md_writer_tests.rs:550:5:
assertion `left == right` failed: the pointer path and the pipeline that writes the file have drifted
  left: ".trusty-mpm/framework/INSTRUCTIONS-COMPILED.md"
 right: ".trusty-mpm/INSTRUCTIONS-COMPILED.md"
```

The guard pins the constant against `compiled_prompt_path`. Reverted.

**5 — mechanism.** `effective_style` reaches `maybe_inject_active_style` (`output_style.rs:322`), which calls `resolve_active_style(config, explicit)` and injects the resolved style into the prompt whenever `native_supported == false`. Routing `prepare_session` through the `None`-hardcoded helper would therefore change the composed bytes. The split is justified. Caveat in LOW-4.

**7 — injector framing.** Gone from the fatality rationale in `mod.rs`, the contract block, and the spec's contract section. What remains (spec "Position is load-bearing", `mod.rs` `Test:` pointer) argues *position*, not fatality, which is the correct place for it.

**8 —** `git ls-remote --heads origin backup/fix-4752-pre-r4` → `72c777525d619d007aa611d8ad06a5b1d56972fc`.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | 342 | The message still says "could not **write** the session instructions to {path}" and "a path `tm` must **write** on every launch is unavailable". Half the newly-covered failures are READS — `read_to_string(framework_instructions_path)` and `read_to_string(CLAUDE.md)`. The round-4 finding was that the wording was wrong for a `CLAUDE.md` READ failure; dropping "compiled PM prompt" fixed the noun, not the verb. Worse for `CLAUDE.md` specifically: the seeded stub in this same file (line 440) promises "trusty-mpm will never modify this file again after creating it" (#2170), so the refusal message asserts a write `tm` is forbidden to perform. | Neutral verb: `"could not establish the session instructions at {path}: {source}"` and `"a path `tm` must read or write on every launch is unavailable."` | Fix here |
| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs` | 1503 | Contract-reach asymmetry: `build_instructions` is fatal only on the `prepare_session` path. On resume the equivalent condition (an unreadable `CLAUDE.md`) does not refuse — `refresh_compiled_prompt` → `resolve_pm_prompt` returns a `String` and cannot fail, so it silently degrades to the bundled prompt and the resume proceeds. Same broken project: `tm session start` refuses, `tm sessions resume` starts with the operator's `CLAUDE.md` overrides silently dropped. Not a regression (that was the behavior everywhere before), and the written contract still literally holds on resume; but the strictness now differs per entry point. | Out of scope to fix here. Either give `resolve_pm_prompt` a fallible seam the resume paths consult, or state the asymmetry explicitly in spec §10.3. | Promote |
| LOW | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | 335 | Self-referential doc: "Renamed from `instructions_failure_message` (round 4)" names the **new** name. It was renamed from `compiled_prompt_failure_message`. | s/`instructions_failure_message`/`compiled_prompt_failure_message`/ on that line. | Fix here |
| LOW | `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md` | 1233 | "`instruction_pipeline::compiled_prompt_failure_message` is the single formatter" — live assertion naming a symbol that no longer exists, two lines above the sentence pinning it to `instructions_failure_message_names_the_path_and_a_remedy`. (Lines 1269/1271 use the old name correctly, as history.) | Rename to `instructions_failure_message`. | Fix here |
| LOW | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 619 | `prepare_session_inner`'s doc still says it "only returns `Err` for genuinely fatal failures (the instruction pipeline, **or the inspection-stash IO**)". The stash IO no longer returns `Err` at all — this PR made it warn-and-continue. | Drop the stash clause. | Fix here |
| LOW | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 243 | `PrepReport::compiled_prompt` correctly documents that `build_prompt_file` later rewrites the file with a `None`-style composition. Spec §10.3's justification for the split ("would silently drop the operator's chosen output style from the compiled copy") does not mention that for every spawn reaching `build_prompt_file` — all three adapters in `runtime/claude_code.rs` — the preserved style survives only until the spawn. The two documents are consistent but disagree in emphasis; a reader of the spec alone will over-read the benefit. | Add one sentence to spec §10.3 pointing at the `PrepReport::compiled_prompt` caveat. | Parent |
| LOW | `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs` | 977 | The new `?` on `prepare_inproject_session` returns before `create_with_reserved_name`, so a refusal leaves the reserved in-project worktree on disk with no session record — the litter shape #2450's comment (lines 421-429) treats as a defect. Pre-existing pattern (the `?` at ~line 1000 does the same), so this joins an existing class rather than creating one. Refusal path only. | Optional: release the reservation on the error path, or rely on the existing prune. | Promote |

### Gates (run in `.claude/worktrees/qa-4759-r5`)

| Gate | Result |
|---|---|
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | exit 0 |
| `cargo test -p trusty-mpm` | `test result: FAILED. 4393 passed; 1 failed; 7 ignored` — see below |
| `bash scripts/check_line_cap.sh` | `3630 tracked .rs file(s); 7 allowlisted, 0 violations — OK` |
| `bash scripts/check_sld.sh` | `55 spec doc(s) + 3046 code file(s); 0 error(s), 0 warning(s)` |
| `bash scripts/check_changelog_fragment.sh` | `all 1 crate(s) with source changes are recorded` |
| `bash scripts/check_test_pointers.sh` (extra — the rename touches `Test:` pointers) | `resolved 20856 Test: citation(s) — 0 dangling pointers — OK` |

The single failure is the documented environmental flake, not this diff:

```
panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18:
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:63883/api/v1/doctor?…)")
```

Named in `CLAUDE.md` / `docs/reference/test-ladder-baseline.md` as a known-environmental timing flake. Re-run in isolation:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 4400 filtered out; finished in 9.99s
```

The diff touches nothing in `client/executor/`; its only `client/` change is `http_client/session_connect.rs`, a different module and a different route. No coverage was deleted, ignored, cfg-gated or excluded to reach this state.

### Notes

- Gate ran against `origin/main = e2737547`, the stated base.
- Not re-litigated per scope: whether the write should be fatal, whether it is a protected path, and the two fatal sites under one condition.

**Factual finding for the OPEN stash question (policy call is not mine).** The premise "an inspection copy nothing reads" is **not accurate** — `.trusty-mpm/last-instructions.md` has two production readers beyond `tm session instructions`:

1. `crates/trusty-mpm/src/daemon/managed_routes/deployment_check.rs:171-184` — `carrier_reachable()` treats the stash as one of exactly **two** delegation-persona carriers. When the output-style carrier is also gapped, a missing-or-empty stash is what flips it to `false` and fires `warn_if_no_persona_carrier` ("the launched PM may be missing its identity/instructions carrier"). Diagnostic only; explicitly cannot block a launch (#2231/#2172).
2. `crates/trusty-mpm/src/daemon/doctor_fs_checks.rs:39` — `tm doctor`'s `instructions` check keys directly off this file: present and non-empty → `Ok`, empty → **`Fail`**, missing → `Warn`.

Both consumers are diagnostic, so the warn-and-continue policy may well still be the right call — but it should be decided knowing the file is read, not on the basis that it is not. Note this is **not a regression from this PR**: under the old short-circuit the stash was equally absent, the session launched anyway, and both readers saw the same thing. What changed is only how far preparation gets afterward.
- Both mutations were reverted; the review worktree is clean.

Issue #4754, opened 2026-08-04
tm-scaffold: replace tm-init as the single owner of project customization


bobmatnyc:
**Gap.** No component owns project customization. `tm-init` covers part of it under a name that no longer describes the job, and nothing encapsulates CLAUDE.md section-override editing.

**Resolution.** Build `tm-scaffold`, replacing `tm-init`. It owns:

- **(a)** project agent/skill manifests, driven by detected stack — routed through the existing detector (`stack_profile.rs`) and deployment machinery, not reimplemented
- **(b)** CLAUDE.md overrides on request — marked section blocks per `crates/trusty-mpm/src/assets/instructions/sections/README.md`; idempotent replace, `CORE` declined, prose outside markers preserved, malformed input leaves the bundled section in force; writes the pointer to `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` ([#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752))
- **(c)** other project customizations (`.gitignore` etc.) — additive and idempotent
- **(d)** every `tm-init` capability that survives inventory; anything dropped is listed with a reason

Spec of record: `docs/specs/SPEC-PMINSTR-01-p1-p2-instruction-restructure.md`.

**Constraints.**
- Project agent/skill files are classified by **frontmatter, never filename** — trusty-mpm and the unrelated Python `claude-mpm` share the `.claude/` convention, and a filename match claims the wrong project's files.
- After this lands, no other component writes `CLAUDE.md`.
- Deployed `tm-init` copies do not self-retract; orphan handling is proposed in the PR, not swept in it.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

PR #4762, opened 2026-08-04
feat(trusty-mpm): single-owner writer for CLAUDE.md section overrides (refs #4754)

bobmatnyc:
PR1 of the #4754 split. The `tm-scaffold` skill itself (PR2) waits for the
`bundle_all.rs` fence to clear; orphan handling is PR3. This PR touches no
fenced file.

## Why

`core::claude_md_sections` is explicitly reader-only ("Scope: reader only"), so
every write to a project's `CLAUDE.md` has so far been done by hand — by an LLM
following prose in `tm-init` — against a grammar only the reader knows.

Prose cannot be idempotent. The same hand-merged pattern stacked five copies
into the shared `## [Unreleased]` changelog section (#4463–#4475; #4399 burned
three rebase rounds), which is why changelog entries moved to per-PR fragments.
A stacked *section override* is worse than a stacked changelog entry: the reader
keeps the FIRST well-formed block and reports the rest as `REASON_DUPLICATE`, so
the copy the author just wrote is the one silently ignored.

## What

`core::claude_md_writer` — the write-side counterpart.

| Behaviour | How it holds |
|---|---|
| Write a named block | `write_section_override`, emitting the reader's grammar |
| Idempotent | Replaces in place and collapses pre-existing duplicates: exactly one block per section survives |
| Decline `CORE` | Asks `CustomizationTier::permits`; refused because the package declares it `fixed`, with a logged warning |
| Preserve outside verbatim | Splices by line span over `split_inclusive('\n')`, so `\r\n` and trailing whitespace survive byte-for-byte |
| Fail toward more instruction | Protected section, empty body, marker line in the body, unpaired marker in the host, unreadable host → file byte-identical, bundled section in force |
| Pointer to the compiled prompt | `ensure_compiled_pointer` names `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` — referenced, not implemented |

**The grammar is not duplicated.** The writer borrows the reader's recogniser
through three new `pub(crate)` accessors — `SUPPORTED_VERSION`,
`contains_marker_line`, `locate_blocks` — so a writer/reader divergence cannot
be expressed. Every writer test asserts through `scan_project`, not against a
string the writer also produced.

Two design points worth a reviewer's attention:

- **The pointer is deliberately outside the marker grammar.** Its delimiters
  (`<!-- trusty-mpm compiled-instructions pointer >>> -->`) fail the
  `TRUSTY-MPM:` namespace test, so the reader treats the block as ordinary prose
  and raises no `unknown section token` diagnostic. `pointer_block_is_invisible_to_the_section_reader` pins that.
- **A body containing a marker line is refused.** Writing one would nest a
  `START`, and the reader resolves nesting by discarding the *outer* block —
  deleting the very section being set. The screen uses the reader's recogniser,
  not a substring match, so prose *mentioning* a marker stays writable
  (`body_mentioning_the_marker_in_prose_is_written`).

No `remove_section_override` is included: deleting a block already has a safe
spelling (an absent block reverts to bundled), so a delete path whose bug mode
is "removed more of CLAUDE.md than asked" is not worth adding without a caller.

## Test ladder

Rung 3 — localized behavior inside one crate.

Each guard was **mutation-tested**: disabling the span reuse, the tier check,
the marker screen, and the unclosed-host guard each fails its own named test, so
no fixture names a guard it does not exercise.

`every_other_section_is_writable` exists specifically so `core_is_declined_and_logged`
cannot pass by refusing everything.

```
cargo fmt --check                        exit 0
cargo check -p trusty-mpm                Finished
cargo clippy -p trusty-mpm --all-targets -- -D warnings   Finished
cargo test -p trusty-mpm                 exit 0 — lib: 4368 passed; 0 failed; 7 ignored
bash scripts/check_line_cap.sh           3625 files, 0 violations — OK
bash scripts/check_changelog_fragment.sh OK
```

New coverage: 23 writer tests + 3 locator tests, all in temp dirs. This diff
changes no root `CLAUDE.md` (`git diff --name-only origin/main...HEAD` lists
only the six files above).

> **Correction to the brief I was given.** It stated the root `CLAUDE.md` is not
> tracked on main and that `claude-md-guard` forbids tracking it (#2299). Both
> are stale: `CLAUDE.md` has been tracked on main deliberately since #4286 split
> C (`2e8a36be`, "migrate to tracked root CLAUDE.md"), and no `claude-md-guard`
> workflow or script exists in the tree any more. Nothing here depended on that
> premise — the tests use temp dirs because that is correct for a filesystem
> writer, not to dodge a guard — but the constraint should not be re-issued to
> the next agent as written.

Refs #4754

### DIFF
--- a/crates/trusty-mpm/changelog.d/4754-claude-md-writer.md
+++ b/crates/trusty-mpm/changelog.d/4754-claude-md-writer.md
@@ -1,0 +1,16 @@
+Added
+
+- `core::claude_md_writer` — the single owner of `CLAUDE.md` section-override
+  writes, the write-side counterpart to the reader-only
+  `core::claude_md_sections` (refs [#4754](https://github.com/bobmatnyc/trusty-tools/issues/4754))
+  - `write_section_override` is idempotent by construction: a repeat write
+    replaces its block in place and collapses any pre-existing duplicate, so a
+    section can never accumulate a stacked second copy
+  - `core` is refused because the shipped package declares it `fixed` — the
+    writer asks `CustomizationTier::permits`, it does not carry its own list
+  - every refusal (protected section, empty body, a marker line inside the body,
+    an unpaired marker in the host, an unreadable host) leaves the file
+    byte-identical and the bundled section in force
+  - `ensure_compiled_pointer` records where the composed PM prompt lands, using
+    delimiters deliberately outside the `TRUSTY-MPM:` grammar so the reader sees
+    no override and raises no diagnostic
--- a/crates/trusty-mpm/changelog.d/4754-claude-md-writer.md
+++ b/crates/trusty-mpm/changelog.d/4754-claude-md-writer.md
@@ -11,6 +11,9 @@ Added
   - every refusal (protected section, empty body, a marker line inside the body,
     an unpaired marker in the host, an unreadable host) leaves the file
     byte-identical and the bundled section in force
+  - written blocks adopt the host's dominant line ending, so splicing into a
+    CRLF `CLAUDE.md` does not leave a mixed-ending file
   - `ensure_compiled_pointer` records where the composed PM prompt lands, using
     delimiters deliberately outside the `TRUSTY-MPM:` grammar so the reader sees
-    no override and raises no diagnostic
+    no override and raises no diagnostic; it collapses duplicate pointer blocks
+    the same way section writes do
--- a/crates/trusty-mpm/src/core/claude_md_sections.rs
+++ b/crates/trusty-mpm/src/core/claude_md_sections.rs
@@ -82,7 +82,15 @@ use crate::core::instruction_package::{
 pub const HOST_FILES: [&str; 1] = ["CLAUDE.md"];
 
 /// The only marker version this build understands.
-const SUPPORTED_VERSION: u32 = 1;
+///
+/// Why: [`crate::core::claude_md_writer`] stamps this exact number into every
+/// block it emits rather than hardcoding its own `1`. A writer that spelled the
+/// version independently could drift past the reader on the next bump and emit
+/// blocks its own reader would decline as [`REASON_UNSUPPORTED_VERSION`] — the
+/// writer/reader drift this pairing exists to make impossible.
+/// What: the `v=` value written, and the only one accepted.
+/// Test: `written_block_declares_the_readers_supported_version`.
+pub(crate) const SUPPORTED_VERSION: u32 = 1;
 
 /// Literal opening a marker line.
 const MARKER_OPEN: &str = "<!--";
@@ -371,6 +379,64 @@ fn parse_blocks(text: &str, host: &Path) -> (Vec<ParsedBlock>, Vec<ScanDiagnosti
         diagnostics.push(diagnostic(host, start_line, REASON_UNCLOSED));
     }
     (blocks, diagnostics)
+}
+
+/// Whether any line of `text` would be recognised as a framework marker.
+///
+/// Why: [`crate::core::claude_md_writer`] must refuse a body that itself
+/// contains a marker line. Emitting one would nest a `START` inside the block
+/// being written, and [`parse_blocks`] resolves nesting by DISCARDING the outer
+/// block ([`REASON_UNCLOSED`]) — so an override body carrying a stray marker
+/// would silently delete the very section the author was trying to set. The
+/// check has to use the reader's own recogniser, not a substring search for
+/// `TRUSTY-MPM:`, because prose mentioning the marker inline is legal content
+/// and must stay writable.
+/// What: `true` when [`parse_marker`] accepts any line of `text`.
+/// Test: `body_containing_a_marker_line_is_refused`,
+/// `body_mentioning_the_marker_in_prose_is_written`.
+pub(crate) fn contains_marker_line(text: &str) -> bool {
+    text.lines().any(|line| parse_marker(line).is_some())
+}
+
+/// Where a section's well-formed blocks sit in a host, and whether the host's
+/// marker structure is intact.
+///
+/// Why: the writer needs exactly two facts before it edits a file — which line
+/// ranges belong to the section it is replacing, and whether any marker in the
+/// file is unpaired. It must not re-derive either from its own parser.
+/// What: `spans` holds `(start_line, end_line)` 0-based inclusive line indices
+/// of every well-formed block claiming the section, in file order; `unclosed`
+/// is true when any `START` in the host lacked a matching `END`.
+/// Test: `locate_blocks_finds_one_span_per_block`,
+/// `locate_blocks_flags_an_unclosed_marker`.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub(crate) struct HostBlocks {
+    /// 0-based inclusive `(start_line, end_line)` of each matching block.
+    pub(crate) spans: Vec<(usize, usize)>,
+    /// Whether the host contains an unpaired `START` marker.
+    pub(crate) unclosed: bool,
+}
+
+/// Locate every well-formed block in `text` that claims `section`.
+///
+/// Why: one grammar, one pairing pass. The writer splices by line index, and
+/// deriving those indices from [`parse_blocks`] — the same function the reader
+/// uses — is what guarantees the writer can never target a region the reader
+/// would parse differently.
+/// What: runs [`parse_blocks`] and projects the blocks whose token resolved to
+/// `section`, plus whether any [`REASON_UNCLOSED`] diagnostic was raised.
+/// Test: `locate_blocks_finds_one_span_per_block`,
+/// `locate_blocks_flags_an_unclosed_marker`, `locate_blocks_ignores_other_sections`.
+pub(crate) fn locate_blocks(text: &str, section: SectionId) -> HostBlocks {
+    let (blocks, diagnostics) = parse_blocks(text, Path::new("<in-memory>"));
+    HostBlocks {
+        spans: blocks
+            .iter()
+            .filter(|b| b.section == Some(section))
+            .map(|b| (b.start_line, b.end_line))
+            .collect(),
+        unclosed: diagnostics.iter().any(|d| d.reason == REASON_UNCLOSED),
+    }
 }
 
 /// Decide whether a parsed pair may become an override.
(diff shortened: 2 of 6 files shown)


bobmatnyc:
## Adversarial review — PR #4762

Reviewed via `gh pr diff 4762` / `gh pr view 4762` against `crates/trusty-mpm/src/assets/instructions/sections/README.md` and the shipped `claude_md_sections.rs` reader, `instruction_package.rs`, `instruction_pipeline.rs`, and `trusty-agents-common::agents::manifest::atomic_write`. No worktree, no edits.

## Verdict: APPROVE

Zero CRITICAL, zero HIGH findings. Four LOW/MEDIUM observations below, none blocking — all `Promote`/`Parent`.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/claude_md_writer.rs` | `write_section_override`, `ensure_compiled_pointer`, `persist` | Read-modify-write on `CLAUDE.md` with no cross-process lock. Two concurrent writers (two agent sessions/`tm-scaffold` runs on one project) can both read the same "before" text, and the second `atomic_write` rename silently clobbers the first writer's change — the exact shape issue #4409 fixed for the agent manifest via `manifest_lock_path`/`with_manifest_lock`. Confirmed via `grep` that this PR adds zero callers, so blast radius is zero today. | Have the PR2 (`tm-scaffold`) wiring serialize calls, or add a sidecar lock the same way `agents::manifest` does, before any caller exists in production. | Promote |
| LOW | `crates/trusty-mpm/src/core/claude_md_writer.rs` | `render_block`, `render_pointer` | Both always emit `\n`-terminated lines. Splicing a freshly rendered block into a CRLF-authored host (as in `preserves_crlf_line_endings`) leaves the surrounding prose CRLF but the new block LF — a file with mixed line endings after any write into a CRLF project. The existing CRLF test only asserts the *surrounding* prose keeps `\r\n`; it does not assert (and the code does not guarantee) the new block matches the host's convention. | Detect the host's dominant line ending and render the block with it, or accept and document the mix. | Promote |
| LOW | `crates/trusty-mpm/src/core/claude_md_writer.rs` | `splice` | The `consumed_to_eof && !text.ends_with('\n')` branch (dropping the block's own trailing `\n` when a *replaced* block reaches true EOF and the original file had no final newline) is untested. Only the *append*-path no-trailing-newline case (`appends_newline_when_existing_file_lacks_trailing_newline`) has a test; the replace-at-EOF case is exercised only by static reading of the code, not a fixture. Static analysis says the logic is correct (segment/line-index alignment holds for `split_inclusive('\n')` vs. the reader's `lines()`, including under CRLF, since `\r` is never itself a split point). | Add a test seeding a block at true EOF with no trailing newline and asserting the replaced file still lacks one. | Promote |
| LOW | `crates/trusty-mpm/src/core/claude_md_writer.rs` | `pointer_spans` | Finds only the *first* `POINTER_BEGIN`/`POINTER_END` pair, unlike `locate_blocks`, which `write_section_override` uses to collapse *every* duplicate. If a host ever ends up with two pointer blocks (only reachable by hand-editing — this module never creates a second one on its own), `ensure_compiled_pointer` normalizes the first and leaves the second stray pair as ordinary, never-cleaned-up prose. No functional impact (the pointer is inert to the reader either way), just an asymmetry with the section-override collapse guarantee. | Extend `pointer_spans` to collapse all matches, mirroring `locate_blocks`, or accept the asymmetry given the low reachability. | Promote |
| Note (not a finding) | `crates/trusty-mpm/src/core/claude_md_writer.rs` | `COMPILED_INSTRUCTIONS_PATH`, `ensure_compiled_pointer` | The pointer names `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md`, a path nothing in the tree writes yet (`grep -rn INSTRUCTIONS-COMPILED` outside this PR: zero hits). The compiled prompt still lands at `~/.trusty-mpm/framework/instructions/INSTRUCTIONS.md` (`instruction_pipeline.rs::install_system_prompt`) until issue #4752 (open, explicitly cited by #4754 as the reason for this exact path) lands. `ensure_compiled_pointer` has zero callers in this PR, so it's inert today — this is a sequencing dependency for the PM to track before PR2 wires it in, not a defect in PR1. | — | Parent (track against #4754/#4752) |

## Answers to the specific attack points

**1. Writer/reader round-trip.** Holds. The writer imports `SUPPORTED_VERSION`, `contains_marker_line`, and `locate_blocks` directly from `claude_md_sections` rather than re-deriving them; `locate_blocks` is a thin projection over the reader's own `parse_blocks`. Traced `render_block`'s output (`<!-- TRUSTY-MPM: {token} START v={SUPPORTED_VERSION} -->\n{body}\n<!-- TRUSTY-MPM: {token} END -->\n`) against `parse_marker`'s grammar field-by-field: namespace match, token, `START`/`END`, optional `v=<u32>` compared against the same constant, and rejection of any trailing field — every emitted block satisfies it, and every test that matters (`writes_a_new_block_a_reader_accepts`, `round_trips_through_the_reader`, `replacing_updates_the_body_in_place`, the duplicate-collapse test) asserts through `scan_project` rather than re-parsing the writer's own string. Could not construct a block the writer emits that the reader would reject or parse differently.

**2. Duplicate-collapse.** Correct given the reader's actual first-wins semantics: a shadowed (non-first) block was already never live — the reader reports it `REASON_DUPLICATE` and never surfaces its body to the composer — so removing it on a write changes bytes on disk, not behavior. Verified this holds when duplicates are non-identical (`replacing_collapses_duplicate_blocks_to_one` seeds `FIRST`≠`SECOND` and independently proves the *fixture* itself trips `REASON_DUPLICATE` before asserting the fix — a stronger test than assuming the premise), when interleaved with unrelated prose (same test: `intro`/`middle prose`/`outro` all survive), and when the host carries blocks for several sections (`locate_blocks_ignores_other_sections`, plus `splice`'s span-based mechanics never touch content outside the section-scoped spans it's given, so a third section's own well-formed block sitting between two duplicates is preserved as ordinary "between-spans" text by construction, not merely by the specific fixtures shown).

**3/4. Byte preservation and fail-toward-more-instruction under adversarial input.** CRLF and no-trailing-newline are directly tested (modulo the EOF-replace gap above). Line-index alignment between the reader's `text.lines()` (used to compute `start_line`/`end_line`) and the writer's `text.split_inclusive('\n')` (used to slice) holds for CRLF because `\r` immediately precedes `\n` and is never itself a split point — both iterators produce the same element count and boundaries. A marker inside a fenced code block or YAML frontmatter is not fence-aware in *either* the reader or the writer (neither was before this PR), so it's treated identically by both — a pre-existing, shared limitation, not a new divergence this PR introduces. Unicode is safe because all splitting is on the single-byte `\n`, which never appears as a UTF-8 continuation byte. A section id that is a literal prefix of another does not collide because `section_for_token` does whole-token equality, not prefix matching. Every refusal path traced (`Protected`, `EmptyBody`, `BodyContainsMarker`, `HostMalformed`) returns before `persist` is ever called, and for `EmptyBody`/`BodyContainsMarker`/`Protected` the host isn't even read — confirmed against `refusals_leave_the_file_byte_identical` and the individual per-case assertions. `persist` itself routes through the shared `atomic_write` (temp file in the same directory, then rename), independently confirmed crash-safe by `atomic_write_leaves_old_intact_on_interrupted_write` in `trusty-agents-common`.

**5. See the findings table** — all refusal paths leave the file byte-identical, not merely "unwritten"; no half-written state is reachable because every write is a single atomic rename.

**6. Pointer's grammar exemption.** Verified by hand against `parse_marker`: it requires the first 11 characters after stripping `<!--`/`-->` and trimming to case-insensitively equal `"TRUSTY-MPM:"` (with the trailing colon). `POINTER_BEGIN`'s body after stripping is `"trusty-mpm compiled-instructions pointer >>>"` — its first 11 characters are `"trusty-mpm "` (a space, not a colon), so the namespace check fails and `parse_marker` returns `None`. The reader genuinely cannot see it, independent of the pinning test. Idempotency holds because `render_pointer()` is a pure function of fixed constants (no interpolated variable content), so two calls always splice in byte-identical replacement text, `next == text`, and `ensure_compiled_pointer` reports `Unchanged` (see the fourth-table finding for the one real gap: it doesn't collapse a hand-duplicated pair the way `locate_blocks` does for sections).

**7. Absent `remove_section_override`.** None of the six required behaviors call for a delete path, and the stated reasoning holds: an absent block already reverts to bundled on the read side, so deletion has a safe spelling without a code path whose failure mode is "deleted more than asked." Consistent with this repo's minimalism convention. Agree it's correctly out of scope for PR1.

**8. Fixture integrity.** Independently re-derived (not just re-read) the logic of `splice`, `append_block`, `section_is_writable`, `locate_blocks`, and `contains_marker_line` against their tests rather than trusting the PR description's mutation-testing narrative. Each of the four named guards is proven by a pair of tests that rules out the classic "guard that always fails/refuses would still pass" failure mode: `core_is_declined_and_logged` is paired with `every_other_section_is_writable` (which asserts `written >= 8`, so a writer that refused *everything* would fail it); the marker screen is paired (`body_containing_a_marker_line_is_refused` vs. `body_mentioning_the_marker_in_prose_is_written`) so a blunt substring check would fail one of the two; the unclosed-host guard is cross-checked in both its own test and the consolidated `refusals_leave_the_file_byte_identical`; the duplicate-collapse test proves its own premise (`REASON_DUPLICATE` actually present) before asserting the fix. Read every conditional in `claude_md_writer.rs` looking for an inverted check or dead branch consistent with a leftover mutation — found none. Confidence >80% no mutation residue survived into the committed diff, based on direct code reading, not on the author's own account of their process.

**Conventions.** Doc comments cite by symbol (`[`crate::core::...::X`]`) throughout, never by line number. Why/What/Test present and proportional on every non-trivial item. Changelog fragment present and correctly formatted (`crates/trusty-mpm/changelog.d/4754-claude-md-writer.md`, `Added` category). `claude_md_writer.rs` is 467 added lines (under the 500-SLOC production cap even before comment-stripping); `claude_md_writer_tests.rs` at 622 lines is well under the 3000-SLOC test cap. `claude_md_sections.rs` was already over 500 raw lines pre-PR and is not on `.line-cap-allowlist.tsv`, so its SLOC (code-only, per the project's counting definition) must already be under 500 given the heavy doc-comment share — consistent with the PR's own reported `check_line_cap.sh` pass, which this review did not need to independently reproduce (main checkout has no worktree available for a build).

---

Issue #4731, opened 2026-08-04
~6 trusty-search watcher/FSEvents tests fail on macOS on clean base commit


bobmatnyc:
## Summary

Running `cargo test -p trusty-search` on macOS fails ~6 lib-target tests on an **unmodified base commit** — same failures across all parallel agents this session (2026-08-03), but passes on Linux CI.

## Failing Tests

- `service::watcher::*` (multiple)
- `service::watch_loop::*` (multiple)
- `service::watcher_manager::save_triggers_incremental_index_via_manager`
- Integration test: `corpus_open_quarantine_4122` (positive control fails: "this environment cannot distinguish a refusal from a stalled watcher")

## Evidence

Independently reproduced by **four separate agents** this session, each stashed or removed its branch and re-ran against clean base:

```
baseline runs:      1364 passed; 6 failed
with unrelated BR1: 1371 passed; 6 failed
with unrelated BR2: 1384 passed; 6 failed
with unrelated BR3: 1387 passed; 6 failed
```

Same 6 test names, same panic messages, every time. One agent verified its change was bin-target-only while all 6 failures are lib-target — code could not have caused them.

All tests pass on Linux CI (Test job is green, ~15 minutes). This is macOS/FSEvents host-specific — sandboxed worktrees appear unable to observe file events.

## Impact

The local test gate is **never clean on macOS**. Every agent and developer must independently re-derive which failures are theirs versus the baseline — that cost was paid four times in one session. Worse, a persistent 6-failure floor masks genuine regressions in the same area.

## Requested Fix

Either:
1. Make these tests pass in a sandboxed macOS worktree, OR
2. Make them skip explicitly with a stated reason when FSEvents is unavailable

Do NOT propose deleting the coverage. A clean local run must mean something.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
## Not reproducible at `8e6ca079d` (2026-08-11) — measured, not assumed

Attempted to harden this cluster alongside #3769 and could not get it to fail. Every run on this host, at this commit:

| Attempt | Result |
|---|---|
| `service::watch*` (21 tests) in isolation, 20 consecutive runs | **20/20 green** |
| `cargo test -p trusty-search --lib` | **1582 passed, 0 failed, 1 ignored** |
[...]

PR #4768, opened 2026-08-04
fix(trusty-search): close four review findings on the reindex resume checkpoint (#4721)

bobmatnyc:
Follow-up to **#4721** (`4cc249fb`, "resume an interrupted reindex from its staging corpus", closing #3979), which shipped in **trusty-search 0.42.0**. Four advisory findings were raised against it by an MCP review tool and never posted to GitHub, so there is no issue and no review thread — **this PR body is the record**.

All four sit on the live reindex data path in 0.42.0, so they are treated as production defects, not cleanup.

---

## Finding 1 — separator-free config-fingerprint collision

`config_fingerprint` rendered each field as `name=value;` and each list as `sort().join(",")`. That encoding is **not injective**:

| Configuration A | Configuration B | Rendered (0.42.0) |
|---|---|---|
| `exclude_globs = ["a", "b"]` | `exclude_globs = ["a,b"]` | `exclude_globs=a,b;` |
| `include_paths = ["x;exclude_globs=y"]`, `exclude_globs = []` | `include_paths = ["x"]`, `exclude_globs = ["y;exclude_globs="]` | `include_paths=x;exclude_globs=y;exclude_globs=;` |

Two genuinely different walk configurations sharing a fingerprint means a checkpoint passes the config gate for a configuration it was **not** built under — the staged corpus is not a prefix of what this run would build, and the run promotes it anyway. That is exactly the outcome the whole checkpoint-versioning scheme exists to prevent.

**Fix — length-prefixed framing**, not a rarer separator. Every field name, every scalar, and every list element is absorbed as `len ‖ bytes` (8-byte LE length), with the element count absorbed alongside each list, plus a domain-separation tag. The justification for choosing length-prefixing over a delimiter: **no separator byte is safe here.** Globs, extensions, and skip-dir names come from operator-authored `indexes.toml`, so they are arbitrary Rust `String`s — `,`, `;`, `=`, `\n`, even `\0` are all legal inside one. POSIX paths reserve only `/` (already meaningful in the value) and NUL. `len ‖ bytes` is injective for arbitrary byte strings regardless of content, so the question does not arise. Paths are also now absorbed as raw `OsStr::as_encoded_bytes` rather than `to_string_lossy`, which maps every invalid sequence to the same `U+FFFD` and is lossy in precisely the collision-creating direction.

### Existing 0.42.0 checkpoints invalidate cleanly — verified

`CHECKPOINT_SCHEMA_VERSION` is bumped **1 → 2**. `evaluate` compares it for exact equality **before** any fingerprint comparison, so a record written by 0.42.0 is discarded on the version gate, `discard_staging` removes the staging file, and the run falls back to a full reindex. It is never misread as valid: the version gate fires first, and the field is a plain `u32` that survives the JSON round-trip. `checkpoints_written_by_the_shipped_v1_format_are_discarded` pins this against a byte-for-byte v1 record whose **every other field matches the current run**, so nothing else could be doing the rejecting, and asserts the discard reason names the schema version.

## Finding 2 — env-mutating test

`resume_kill_switch_disables_adoption` did `unsafe { set_var("TRUSTY_REINDEX_RESUME", …) }` under `#[serial]`. As #4213 and #3954 established, `#[serial]` only excludes other `#[serial]` tests — the binary's ~1,400 non-serial tests keep running inside the window, several of which drive reindexes that read this exact variable. Adding `#[serial]` would have produced a quieter test, not a correct one, so **it was not the fix used**.

Instead, the #4213 technique from `root_hijack_tests.rs`: the test re-executes the test binary with its own exact name, `--exact --test-threads=1`, and the variable supplied through `Command::env` **at spawn** — a per-process value with no in-process `set_var` and nothing for a sibling to race. The mechanism is factored out of `root_hijack_tests` into a shared `service::reindex::test_isolation` module (rather than cloned a third time); `root_hijack_tests` now delegates to it and keeps its assertions, including the "child ran exactly one test" vacuum guard.

The two unit-level env tests are handled differently and more strongly: `resume_enabled` / `checkpoint_max_age_secs` are split into an env read plus a **pure** `resume_enabled_from` / `checkpoint_max_age_from`, which are table-tested with no globals, no `unsafe`, and more cases than before. One isolated child-process test (`env_knob_names_are_wired_up`) still pins the variable NAMES, so nothing is lost.

## Finding 3 — staging-corpus double-open: which is it?

**Conclusion: not two live handles today — but not tidiness either.**

The staging corpus was opened twice: once in `checkpoint::probe_resume` to validate the record, once again in `corpus_swap::adopt_staged_corpus` to install it. Those two opens **cannot** overlap: redb takes an exclusive lock per file, and `open_corpus_db_or_recreate` classifies `DatabaseAlreadyOpen` as *not* an incompatible format, so a second concurrent open returns a plain `Err` (no destructive move-aside). The code worked only because the probe's handle happened to be dropped inside its `spawn_blocking` closure before the adoption ran.

So the shipped code is correct — by coincidence of statement placement, on a lock that would have refused the alternative. What it actually costs:

- **Wasted work** — a second `Database::create` on a multi-GB redb, including its table-materialisation write transaction.
- **An unguarded window** — between the two opens the staging corpus is owned by nobody. The second open has real failure modes (a wedged #3659 open gate, the file removed by another daemon, a lock not yet released by a timed-out detached opener), and `adopt_staged_corpus` swallowed every one of them into `Ok(None)` — **silently abandoning the resume and writing to the live corpus instead**, with only a `warn`.

**Fix:** `ResumeState` now carries the open `CorpusStore` (owned, not `Arc`-shared, so adoption *moves* it onto the indexer and no second reference survives to hold the file at promote time). `begin_staged_corpus_swap` takes `resume` by value. The file is opened exactly once, from validation through adoption, and the adoption has no failure mode left. Every discard path in the probe explicitly drops the handle before unlinking the file it refers to; the runner drops the state if staging turns out to be skipped.

## Finding 4 — checkpoint-exclusion invariant enforced only by call order

The invariant — *a promoted corpus never carries an in-progress checkpoint* — held only because `commit_staged_corpus_swap` happened to call `clear_checkpoint_before_promote` **before** `take_corpus_store`. The clear reads the *installed* store, so swapping those two statements (a reorder no compiler, type, or test would object to) turned it into a silent no-op, and the rename then promoted a finished corpus advertising itself as mid-reindex — which nothing ever cleans up, because the probe only reads `*.tmp` paths. The same state is reachable **today** whenever no store is installed at promote time (a #4122 write quarantine, a failed re-install), where the old code returned early and did nothing.

**Fix, in the spirit of #4718's `BoundedRestoreOutcome`:** remove the call site's opportunity to get it wrong.

- `PromotionRelease` — an opaque, module-private token with a private field.
- `release_staging_for_promotion` is the only way to mint one. It clears the checkpoint, *then* releases the handle, *then* — when the in-store clear could not be **confirmed** — re-opens the now-released file and clears it there, which is only possible because the release already happened.
- `rename_staging_over_live` takes `&PromotionRelease`, so the rename is unreachable without it. Reordering is not expressible.
- `clear_checkpoint_before_promote` now returns `bool` rather than `()` — the `()` return is what let "no store installed" look like success.

If the fallback also fails it logs at ERROR and promotion still proceeds: a stale record in a live corpus is genuinely inert (the probe never reads live corpora; `copy_all_from` never copies this key), so refusing to promote would discard a completed reindex to avoid a cosmetic defect. No corpus is ever deleted, truncated, or rebuilt on any path here.

---

## Tests

One regression test per finding, each with a `// #4721:` pointer at the change site.

| Finding | Test |
|---|---|
| 1 | `config_fingerprint_distinguishes_list_element_boundaries`, `config_fingerprint_cannot_be_forged_across_field_boundaries`, `checkpoints_written_by_the_shipped_v1_format_are_discarded` |
| 2 | `resume_kill_switch_disables_adoption` (now child-isolated, with a guard asserting the child really has the kill switch set), `env_knob_names_are_wired_up`, `resume_enabled_from_parses`, `checkpoint_max_age_from_parses` |
| 3 | `probe_hands_the_open_staging_corpus_to_the_adoption` |
| 4 | `promotion_clears_the_checkpoint_even_when_the_store_was_released_early` |

No coverage was weakened: nothing is `#[ignore]`d, cfg-gated, or narrowed. The two deleted `set_var` unit tests are replaced by pure-parse tests covering strictly more cases plus an isolated test covering the env wiring they used to cover.

### Findings 1 and 4 fail against the shipped behavior — demonstrated

Fingerprint + schema version temporarily reverted to the 0.42.0 form:

```
test service::reindex::checkpoint::tests::checkpoints_written_by_the_shipped_v1_format_are_discarded ... FAILED
test service::reindex::checkpoint::tests::config_fingerprint_cannot_be_forged_across_field_boundaries ... FAILED
test service::reindex::checkpoint::tests::config_fingerprint_distinguishes_list_element_boundaries ... FAILED
test result: FAILED. 14 passed; 3 failed; 0 ignored; 0 measured; 1438 filtered out

panicked at checkpoint_tests.rs:361: internal error: entered unreachable code:
  #4721: a checkpoint written by 0.42.0 must never be adopted
panicked at checkpoint_tests.rs:238: assertion `left != right` failed:
  #4721: a field value must not be able to forge the framing of a later field
panicked at checkpoint_tests.rs:208: assertion `left != right` failed:
  #4721: exclude_globs ["a", "b"] and ["a,b"] are different walk filters
```

Promote path temporarily reverted to "clear through the installed store only, no fallback":

```
test service::reindex::resume_tests::promotion_clears_the_checkpoint_even_when_the_store_was_released_early ... FAILED
test result: FAILED. 7 passed; 1 failed; 0 ignored; 0 measured; 1447 filtered out

panicked at resume_tests.rs:649: #4721: a promoted corpus must not carry an
  in-progress checkpoint, no matter whether the staging handle was still
  installed when the promotion started
```

Both reverts were removed afterwards; `grep TEMPORARY REVERT` over `service/reindex/` is empty and the committed tree is the fixed one.

## Verification

```
$ cargo test -p trusty-search --no-fail-fast
  lib                                    1449 passed; 5 failed; 1 ignored
  bin trusty-search                       325 passed; 0 failed
  tests/corpus_open_quarantine_4122.rs      3 passed; 0 failed
  tests/integration_tests.rs                7 passed; 0 failed
  tests/config_mount.rs                     3 passed; 0 failed
  tests/data_dir_forward.rs                 3 passed; 0 failed
  tests/ooc_quick_wins.rs                   6 passed; 0 failed
  tests/residency_cold_park.rs              2 passed; 0 failed
  tests/typeahead.rs                        7 passed; 0 failed
  tests/warm_boot_corpus_open_failure.rs    2 passed; 0 failed
  tests/embedder_throughput_bench.rs        4 passed; 0 failed

    Finished `dev` profile — 0 warnings

$ cargo fmt --check
(no diff)

line-cap: measured 3623 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

The 5 lib failures are the macOS FSEvents watcher set tracked as **#4731** (`service::watcher::*`, `service::watch_loop::*`, `service::watcher_manager::*`) — this host does not deliver FSEvents. Baseline not re-derived, per that issue. Every reindex test passes: `cargo test -p trusty-search --lib service::reindex` → **131 passed; 0 failed**.

## Docs

`docs/trusty-search/spec/COMPONENTS.md` §1 records the length-prefixed fingerprint, the 1 → 2 schema bump and what it means for checkpoints already on disk, the single-open probe→adopt handover, and the release-gated checkpoint clear.


bobmatnyc:
## trusty-review verdict: APPROVE (grade B-)

Reviewed via `mcp__trusty-review__review_pr` (model: us.anthropic.claude-sonnet-4-6, head SHA 6f19d667ec2dcbfe42a740dd2d5bbeb452ce9ee2). No findings.

**Summary:** Solid follow-up PR that closes four real bugs in the reindex resume path: injective config fingerprinting, schema-version bump for on-disk invalidation, single-open staging handle handoff, and a type-enforced checkpoint-clear-before-rename ordering — all backed by targeted new tests and no regressions.

Additional manual verification performed against the two specific invariants called out for this follow-up:

- **`CHECKPOINT_SCHEMA_VERSION` 1→2 bump**: confirmed the new test `checkpoints_written_by_the_shipped_v1_format_are_discarded` constructs a v1-schema record whose `config_fingerprint` is deliberately set to match the *current* config's fingerprint, then asserts `evaluate()` still discards it with a "schema version" reason. This proves the version gate fires before any fingerprint comparison — a 0.42.0-written checkpoint can never be misread as valid.
- **`PromotionRelease` token**: confirmed it is a struct with a private `_sealed: ()` field, mintable only inside `release_staging_for_promotion` (clear checkpoint → release staging handle → verify). `rename_staging_over_live` takes `&PromotionRelease` as a required parameter, so the promote rename is unreachable without going through the correct clear/release order first — the invariant is enforced by the type system, not by call-site convention.

Issue #3408, opened 2026-07-20
File watcher unusable for network-mounted (EFS/NFS) serving deployments: boot-time recursive-inotify walk + blind to cross-host writes


bob-duetto:
**Summary.** trusty-search's native file watcher (`notify` v6 `RecursiveMode::Recursive`, `crates/trusty-search/src/service/watcher.rs:79`) cannot be used on a serving box that mounts its corpus over EFS/NFS, for two independent reasons — and the only control is the all-or-nothing `TRUSTY_DISABLE_WATCHER` (`src/service/watcher_manager.rs:45,58`).

## Problems

1. **Boot-time recursive walk.** inotify is not natively recursive, so enabling the watch forces `notify` to walk the entire tree to install a watch descriptor per subdirectory. On a ~32K-file corpus over EFS this is slow enough to blow the per-index warm-boot restore budget, and risks exhausting `fs.inotify.max_user_watches`. There is **no flag to watch without this boot walk** (no watch-only / deferred / lazy-install mode; debounce is a compile-time const `watcher.rs:27` with no override).

2. **Blind to cross-host writes.** inotify only observes writes made by the local host. In a build/serve split (an indexer box updates the shared mount; serving boxes only read + query), the serving box's watcher would **silently miss** all changes written by the other host, even after paying the walk cost.

## Impact

Operators must run the daemon with `TRUSTY_DISABLE_WATCHER=1` on any network-mounted serving box, losing daemon-native incremental via the watcher entirely.

## Requests (either resolves it)

- **(a)** Decouple "watch" from "boot walk/scan" — a watch-only mode that installs watches lazily/deferred and skips any boot-time reconcile, so the watcher can run without the warm-boot walk; and/or
- **(b)** Officially document the watcher as local-filesystem-only and designate the existing per-file HTTP endpoints `POST /indexes/{id}/index-file` and `POST /indexes/{id}/remove-file` (`src/service/server/files.rs:26,45`) as the **supported incremental path for network-mounted / build-serve-split deployments**, where the consumer drives per-file upserts from its own change detection (e.g. `git diff` on pull) rather than relying on inotify discovery.

## Recommendation: (b)

[...]

PR #4771, opened 2026-08-04
fix(trusty-search): stop the embedder silently downgrading MPS/Python to CPU/ort (#4125)

### LINKED ISSUE #4125: trusty-search: embedder reports bootstrap failed while status ready; silent fallback from MPS/Python to CPU/ORT
## Observed Behavior

After a daemon restart, `search_health` reported:
- `embedder_bootstrap: "failed"`
- `embedder: "ready"`

The embedding backend had changed from the previous boot's `mps/python` to `CPU/ort`.

Nothing was reported as degraded and semantic search worked correctly afterwards (verified with real vector-lane hits).

## Why This Is Concerning

1. **Silent performance regression**: A fallback from GPU/MPS to CPU is a potentially significant performance regression that nothing surfaces to operators or logs.

2. **Contradictory health signals**: `embedder_bootstrap: failed` but `embedder: ready` is internally contradictory — one of the two fields is misreporting.

## Questions for Maintainers

- Is this fallback expected behavior?
- Is the contradiction between the two fields (`embedder_bootstrap: failed` vs `embedder: ready`) intended?
- Should a CPU/GPU fallback be surfaced explicitly in health/logs rather than silently?

bobmatnyc:
Closes #4125

The owner confirmed the MPS/Python → CPU/ort fallback is **not** intended. Three
independent defects, all pre-dating 0.42.0, chained to produce it — and a fourth
signal gap kept it invisible. Root-caused from this machine's own
`~/Library/Logs/trusty-search/stderr.log` plus the source; every claim below is
backed by a log line I re-verified rather than inferred.

## Defect 1 — a timeout was treated as proof the venv is broken

`verify_full_import_smoke()` runs `<venv>/bin/python -c "import
sentence_transformers"` under a hardcoded 10 s budget. The daemon runs that
recheck while simultaneously warm-booting hundreds of indexes, so torch's import
routinely exceeds 10 s under purely self-inflicted startup contention. The check
returned a bare `bool`, so "ran out of time" and "the venv failed the check" were
the same value, and `ensure_venv_checked` condemned a perfectly intact venv to a
full rebuild.

The log shows this firing on nearly every daemon start from 2026-07-26 onward —
90 occurrences of `py-embedder: full-import-recheck timed out after 10s`, the
most recent at `2026-08-04T13:39:23`.

This is the same conflation of *slow* with *broken* that #4333 fixed for a corpus
open timeout being reported as "corrupted format", and it is fixed the same way:
a timeout is now a distinct, named outcome, never a verification failure.

- Rechecks return `RecheckOutcome { Passed, Failed, Indeterminate }`. Only a
  completed run produces evidence; a timeout produces `Indeterminate`.
- The constant was deliberately **not** just raised — a fixed budget under
  variable startup load is wrong again the moment the load changes. An
  over-budget first attempt buys a second attempt with a materially larger
  budget (10 s → 60 s) instead of a verdict. The killed first attempt also
  leaves the interpreter and torch's shared objects warm, so the retry is
  typically much faster than the attempt it follows.
- `Indeterminate` reuses the `.ready` venv rather than rebuilding. That sentinel
  is written only after a real import+embed smoke test passed, and a timed-out
  recheck adds nothing that contradicts it. The residual risk (a venv that is
  both broken *and* slow) is unchanged and already covered downstream: the
  caller's readiness probe forces a real embed through the supervisor before
  anything hot-swaps to python, and a failure there leaves search on ort exactly
  as a failed rebuild would — but without destroying a working venv on the way.

## Defect 2 — `uv` was looked up on PATH only, and the daemon's PATH lacks Homebrew

`locate_uv()` resolved via `TRUSTY_UV_BIN` or a bare `which::which("uv")`, with no
fallback. `uv` lives at `/opt/homebrew/bin/uv`; the launchd plist sets no `PATH`
and the daemon's inherited `PATH` omits `/opt/homebrew/bin`. So once defect 1
triggered a needless rebuild, it hard-failed:

```
graceful python bootstrap attempt 1/2 failed (py-embedder venv bootstrap: `uv` not found
on PATH and TRUSTY_UV_BIN is unset. …) — staying on the ort sidecar
graceful python bootstrap exhausted 2 attempt(s) — staying permanently on the ort sidecar
for this daemon's lifetime
```

This bug class was already fixed once — #1298, for `tmux`/`claude` lookups —
producing `trusty_common::bin_resolve::resolve_binary()`, which falls back to
well-known dirs including `/opt/homebrew/bin`. `locate_uv` was simply never wired
to it. It is now, rather than growing a second fallback list.

Note this only ever bit the daemon: an interactive `trusty-search start` inherits
a login shell's `PATH` and finds `uv` fine, which is why it survived so long.

## Defect 3 — a permanent capability downgrade never reached `overall_status`

`overall_status` was computed from `warm_boot_degraded ||
indexes_watcher_network_degraded > 0 || indexes_stuck_empty > 0`.
`embedder_bootstrap` was populated and then aggregated nowhere, so
`BootstrapState::Failed` and `FellBackToOrt` (from the swap-back watchdog) sat
next to `status: "ok"` forever.

`embedder: "ready"` is not the misreport — it describes the currently-active
backend, which genuinely is ready. What nothing aggregated was *"we did not get
the backend we asked for"*: search silently runs on CPU/ort instead of MPS, a
real and permanent performance regression, on the one field monitors gate on.

Both terminal states now fold into the existing degraded condition, using the
idiom the file already establishes (#1870 / #3408 / #3706 extended the same `||`
chain; #4718 applied the same pattern to #4087). `Bootstrapping` is deliberately
excluded — it is transient, and flagging it would make every Apple-Silicon boot
report degraded for its first few seconds.

## The 30 s startup probe — same class, different remedy

Asked to judge whether `TRUSTY_EMBEDDERD_STARTUP_TIMEOUT_SECS`'s 30 s sidecar
spawn probe shares the fixed-budget-under-contention problem. **It does**, and
the log confirms it independently of defects 1–2: on the
`2026-08-04T12:19:28` boot both bootstrap attempts died at exactly 30 s with no
child output, and the next daemon start succeeded unchanged.

But the correct remedy differs, so I did not reclassify it. A *verification*
check's timeout is a misreading — it says something about the artifact that the
timeout cannot support. A *readiness probe's* timeout is not: a sidecar that will
not answer genuinely cannot serve that attempt. What was wrong was retrying with
the identical budget and then abandoning python permanently. So the probe budget
now scales with the attempt (attempt N gets N× the base) instead of failing the
same way twice. The budget also remains operator-tunable, which the recheck's
hardcoded 10 s never was.

## Tests

Each fix has a regression test that fails against the current behavior;
verified by reverting each fix and observing the failure.

| Fix | Test | Failure when reverted |
|---|---|---|
| 1 (classification) | `bounded_python_check_classifies_timeout_apart_from_failure` | timeout indistinguishable from `Failed` |
| 1 (retry) | `full_import_recheck_retries_once_before_reporting_indeterminate` | `left: 1, right: 2` — no retry |
| 1 (starved venv) | `full_import_recheck_retry_lets_a_starved_venv_prove_itself` | `left: Indeterminate, right: Passed` |
| 1 (no rebuild) | `ensure_venv_does_not_rebuild_on_an_indeterminate_recheck` | rebuild runs, returns the uv error |
| 2 (resolver) | `resolve_binary_finds_a_binary_outside_the_process_path` | PATH-only lookup misses it |
| 3 (aggregation) | `health_is_degraded_when_the_embedder_backend_permanently_downgraded` | `Failed`/`FellBackToOrt` still report `"ok"` |

The health test drives all five `BootstrapState` variants plus the
no-switchable-handle branch, so it pins the fix against over-flagging
(`Bootstrapping`/`Ready`/`NotApplicable`/`n/a` must stay `"ok"`) as well as
under-flagging. The `ensure_venv` test asserts both arms — `Indeterminate`
reuses, `Failed` still rebuilds — so the fix cannot be "kept" by trusting every
outcome.

No coverage was weakened: no `#[ignore]`, no cfg-gating, no narrowing. The
existing recheck tests were updated from `bool` to the three-state outcome, which
tightens them (they now assert *which* outcome, not merely "not passed").

## Verification

```
cargo test -p trusty-embedderd-py
  test result: ok. 20 passed; 0 failed; 0 ignored
  (+ 1 pre-existing #[ignore] e2e test requiring a real torch venv, untouched)

cargo test -p trusty-common --lib bin_resolve
  test result: ok. 14 passed; 0 failed; 0 ignored

cargo test -p trusty-search
  test result: FAILED. 1443 passed; 6 failed; 1 ignored

cargo clippy -p trusty-embedderd-py -p trusty-common -p trusty-search --all-targets -- -D warnings
  Finished — 0 warnings

cargo fmt --check   → clean
scripts/check_line_cap.sh          → 0 violations
scripts/check_test_pointers.sh     → 0 dangling pointers (20761 citations)
scripts/check_changelog_fragment.sh → all 3 crates with source changes recorded
scripts/check_sld.sh               → 0 errors, 0 warnings
```

The 6 `trusty-search` failures are the macOS FSEvents watcher set tracked as
**#4731** — identical names and count before and after this change, and untouched
by it. All change-specific gates pass.

Crate-scoped rather than `--workspace` deliberately: a workspace run exceeds the
tool timeout. Blast radius is the three crates above, all fully gated.

## Not in scope

- **No PyO3 / no change to the process boundary.** The sidecar stays
  out-of-process, per the owner's decision that this is a separate question.
- **`launcher.rs` still uses `which::which`** for the launcher binary. It has a
  sibling-of-`current_exe()` fallback ahead of the PATH lookup, which covers the
  daemon case, so it is not the same live failure — noting it rather than
  silently leaving it, and rather than widening this PR.
- The live daemon was **not** restarted, reconfigured, or `launchctl`-touched.
  Logs and code were read only, so the daemon on this machine is still on ort
  and will pick these fixes up on its next ordinary restart after release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11535,7 +11535,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-embedderd-py"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "fs4 0.8.4",
@@ -11799,7 +11799,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-search"
-version = "0.42.0"
+version = "0.42.1"
 dependencies = [
  "anyhow",
  "arc-swap",
--- a/crates/trusty-common/changelog.d/4125-resolve-binary-fallback-test-seam.md
+++ b/crates/trusty-common/changelog.d/4125-resolve-binary-fallback-test-seam.md
@@ -1,0 +1,8 @@
+Changed
+
+- `bin_resolve::resolve_binary` now delegates to an internal
+  `resolve_binary_in`, which takes the well-known-directory fallback list as a
+  parameter. Behaviour is identical; the seam exists so the fallback branch —
+  "find a binary the process `PATH` does not list", the branch a launchd-spawned
+  daemon depends on — can be tested without mutating the process-global `PATH`
+  (#4125)
--- a/crates/trusty-common/src/bin_resolve.rs
+++ b/crates/trusty-common/src/bin_resolve.rs
@@ -111,9 +111,30 @@ pub fn daemon_path_env() -> String {
 /// an existing `dir/name`; the first hit is returned. Returns `None` if nothing
 /// matches.
 /// Test: `resolve_binary_finds_in_well_known_dir`,
+/// `resolve_binary_finds_a_binary_outside_the_process_path`,
 /// `resolve_binary_returns_none_for_missing`,
 /// `resolve_binary_accepts_absolute_path`.
 pub fn resolve_binary(name: &str) -> Option<PathBuf> {
+    resolve_binary_in(name, &daemon_path_dirs())
+}
+
+/// [`resolve_binary`] with the fallback directory list supplied by the caller.
+///
+/// Why (#4125): the fallback half of [`resolve_binary`] — the half that makes a
+/// launchd-spawned daemon able to find a Homebrew binary its minimal `PATH`
+/// omits — had no test that actually exercised it. The existing
+/// `resolve_binary_finds_in_well_known_dir` deliberately avoids mutating the
+/// process-global `PATH` (that races the parallel harness), so it could only
+/// test [`candidate`] and the explicit-path branch; the real
+/// "found somewhere the process `PATH` does not list" behaviour went unproven,
+/// which is exactly the behaviour `locate_uv` was missing when it hard-failed
+/// the py-embedder bootstrap on a daemon whose `PATH` lacked
+/// `/opt/homebrew/bin`. Parameterizing the fallback list makes that branch
+/// directly testable against a temp dir with no `PATH` mutation at all.
+/// What: identical to [`resolve_binary`] except `fallback_dirs` replaces
+/// [`daemon_path_dirs`] as step 2's search list.
+/// Test: `resolve_binary_finds_a_binary_outside_the_process_path`.
+fn resolve_binary_in(name: &str, fallback_dirs: &[PathBuf]) -> Option<PathBuf> {
     // An explicit path (absolute or relative with a separator) is used verbatim.
     if name.contains(std::path::MAIN_SEPARATOR) {
         let p = PathBuf::from(name);
@@ -130,8 +151,8 @@ pub fn resolve_binary(name: &str) -> Option<PathBuf> {
     }
 
     // 2) Fall back to the well-known daemon dirs (covers launchd's minimal PATH).
-    for dir in daemon_path_dirs() {
-        if let Some(hit) = candidate(&dir, name) {
+    for dir in fallback_dirs {
+        if let Some(hit) = candidate(dir, name) {
             return Some(hit);
         }
     }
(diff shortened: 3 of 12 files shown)


bobmatnyc:
## trusty-review verdict: APPROVE (Grade A-)

Reviewed via `mcp__trusty-review__review_pr` (model: `us.anthropic.claude-sonnet-4-6`).

**Summary:** Solid fix for a real daemon-lifetime regression: the three-state `RecheckOutcome`, the `resolve_binary` seam for `uv` discovery, and the health degradation signal are all correct and well-tested.

**Findings (non-blocking, WARN-level):**

1. **`crates/trusty-search/src/commands/start/graceful_bootstrap.rs:338`** (medium) — Probe budget multiplier `ops.probe_timeout() * attempt.max(1)` is semantically unbounded; at high retry counts this can block for minutes, and `Duration` multiplication overflow panics in debug builds. Suggested: cap via `attempt.min(3).max(1)`.
2. **`crates/trusty-embedderd-py/src/bootstrap_tests.rs`** (medium) — `ensure_venv_does_not_rebuild_on_an_indeterminate_recheck` calls `unsafe { std::env::set_var }` inside a `with_data_override` closure without going through that helper's own serialization; theoretical parallel-test flakiness risk, not a production bug.
3. **`crates/trusty-embedderd-py/src/bootstrap_tests.rs`** (low) — `full_import_recheck_retries_once_before_reporting_indeterminate` uses 1500ms budgets that could be flaky on a loaded CI runner.

None of these block the fix itself. Proceeding with squash-merge per owner authorization.

Issue #3954, opened 2026-07-26
trusty-search: test_trim_heap_never_increases_rss_after_bulk_free asserts an invariant that does not hold under parallel test execution (same flaw as #3702)


bobmatnyc:
## Symptom

`core::memguard::tests::test_trim_heap_never_increases_rss_after_bulk_free` (`crates/trusty-search/src/core/memguard.rs:772-800`, assertion at line 789) fails intermittently in CI:

```
trim_heap() must never increase RSS: 181 MB before trim, 194 MB after trim
```

Observed blocking PR #3944 (CI run `30179715322`), a trusty-code-only PR with zero relationship to trusty-search memory management.

## Root cause

The test allocates 200k small buffers, drops them, samples **whole-process RSS** via `current_rss_mb()` (which reads `/proc/self/status` for the ENTIRE test binary process, not just this test's allocations), calls `trim_heap()` (`libc::malloc_trim(0)`), samples RSS again, and asserts `after <= before`.

"`malloc_trim` never increases RSS" is true of `malloc_trim` in isolation, but the test conflates *"RSS delta caused by trim_heap()"* with *"RSS delta between two wall-clock points in a shared, concurrently-used process."* `cargo test --workspace` runs tests in a binary concurrently (default `--test-threads` = CPU count). This test has no `#[serial]`. Any sibling test allocating in the milliseconds between the two samples inflates "after" regardless of what `trim_heap()` did.

## Evidence

A C program replicating the exact test shape (200k buffers of 150-249 bytes, free, sample RSS from `/proc/self/status`, `malloc_trim(0)`, sample again) run under Debian 12 / glibc 2.36:

- Quiet, single-threaded: **20/20 trials pass** (invariant holds perfectly, e.g. 2MB→1MB)
- With 8 concurrent allocation-churner threads at `--cpus=4`: **7/20 trials (35%) violate** the invariant — e.g. 72MB→93MB, 106→121, 101→113, matching the shape of the CI failure (181→194)
- Same at `--cpus=2`: **3/20 (15%)** — scales with contention, not eliminated at low core counts

This demonstrates the assertion can fail with completely correct, non-buggy `malloc_trim` behavior. It is a bad test, not a memory-management defect.

## Precedent — this is a recurrence, not a novel issue

[...]

PR #4780, opened 2026-08-04
fix(trusty-search): close three review findings on the #4125 embedder bootstrap fix (#4771)

bobmatnyc:
Follow-up to **PR #4771** (`14011f86`, "stop the embedder silently downgrading MPS/Python to CPU/ort", closing #4125), which shipped in the unreleased trusty-search **0.42.1** / trusty-embedderd-py **0.1.2**. Three non-blocking WARN-level findings were raised in that PR's [review comment](https://github.com/bobmatnyc/trusty-tools/pull/4771#issuecomment-5181371816). There is no issue — **this PR body is the record**.

These land before the 0.42.1 cut, so they ship in the same release as the fix they follow.

---

## Finding 1 — unbounded probe-timeout multiplier

`graceful_bootstrap.rs:338` scaled the readiness-probe budget as `ops.probe_timeout() * attempt.max(1)` with no upper bound.

**The scaling itself is kept.** A *readiness-probe* timeout genuinely means that attempt failed — unlike the *verification* timeout #4771 correctly reclassified as `Indeterminate` — and the original defect was retrying with an identical 30 s budget and then abandoning Python permanently. What was missing is a ceiling.

**Fix:** `probe_budget(base, attempt)` clamps the multiplier to `MAX_PROBE_TIMEOUT_MULTIPLIER = 3`.

### Why 3, and why a multiplier rather than an absolute duration

The base is `TRUSTY_EMBEDDERD_STARTUP_TIMEOUT_SECS` (default 30 s) — an operator's own knob. Capping the **multiplier** bounds only the amplification this module invented and leaves the operator's base authoritative; an absolute ceiling would silently override a deliberately-raised base on a slow machine.

3 is generous enough for the failure the scaling exists to survive. The observed incident was a 30 s probe dying under a warm-boot storm of hundreds of index restores, and the *very next* daemon start — without that storm — succeeded on the same unchanged 30 s budget. The deficit is startup contention, which ends; not a fundamentally larger cost. 90 s at the default is three times the budget observed to fail and comfortably outlasts the contention window.

And it bounds the worst case. `TRUSTY_PY_BOOTSTRAP_RETRIES` is operator-settable with no upper bound, so `base * attempt` grows without limit — attempt 100 would have held one live Python child on a **single** probe for 50 minutes. Capped, every attempt is at most 3x the base regardless of how many are configured, while the default 2-retry case is byte-for-byte unchanged (30 s + 60 s).

**Also fixed here:** `Duration * u32` panics on overflow. `probe_budget` uses `checked_mul` and falls back to the base, which is already effectively unbounded at that magnitude.

### The ceiling is visible in the log

The timeout message now names the multiplier and whether the ceiling decided it:

```
python readiness probe timed out after 90s (attempt 7: 3x the 30s base budget, capped at the 3x ceiling)
python readiness probe timed out after 60s (attempt 2: 2x the 30s base budget)
```

### Regression test fails without the cap

`probe_budget` reverted to the shipped `base * attempt.max(1)` form:

```
running 2 tests

thread 'commands::start::tests::probe_budget_saturates_instead_of_overflowing' panicked at
  library/core/src/time.rs:1297:31:
overflow when multiplying duration by scalar

thread 'commands::start::tests::probe_budget_is_capped_at_the_multiplier_ceiling' panicked at
  crates/trusty-search/src/commands/start/tests.rs:1131:9:
assertion `left == right` failed: attempt 4 must be pinned at 3x the base, not 4x it
  left: 120
 right: 90

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 328 filtered out
```

The revert was removed afterwards; `grep -rn "TEMPORARY REVERT" crates/` is empty and the committed tree is the fixed one.

---

## Finding 2 — `unsafe set_var` test isolation in `bootstrap_tests.rs`

**`#[serial]` was not used.** As #4213 and #3954 established, it excludes only *other* `#[serial]` tests; every non-participating test in the binary keeps running inside the window, and `set_var`/`remove_var` are unsound with *any* concurrent reader. It makes a racy test quieter, not correct. The crate-local `ENV_LOCK` mutex this file used had exactly the same hole.

The fix works on two levels.

**Remove the dependency where it can be removed.** `resolve_layout()` was an env lookup plus a pure path derivation welded together. Splitting off `resolve_layout_in(data_dir)` means the eight tests that only ever wanted "a layout under a tempdir" now say so directly — no env, no `unsafe`, no isolation needed, and nothing left to race. That is strictly better than isolating a mutation that did not need to exist.

**Isolate what genuinely needs the environment.** Six tests call `ensure_venv()` / `ensure_venv_eager()` / `ensure_venv_verified()` (which resolve the layout internally) or `locate_uv()` (`TRUSTY_UV_BIN`). Those use the #4213 child-process technique.

### Reused the pattern, not the module — and why

PR #4721's follow-up **#4768** factors this out to `crates/trusty-search/src/service/reindex/test_isolation.rs`. Direct reuse was not possible for two independent reasons:

1. **#4768 is still open.** That file does not exist on `main` (`14011f86`), so this branch cannot depend on it without stacking on an unmerged PR.
2. **Crate boundary.** It is a `#[cfg(test)]`-only module inside `trusty-search`. These tests live in `trusty-embedderd-py`, which `trusty-search` depends on — not the reverse — so importing it would invert the dependency. Exporting it would mean publishing test scaffolding from a `publish = true` crate.

So the *pattern* was followed, and the local copy is deliberately smaller: it needs no `TRUSTY_DATA_DIR` at spawn (each test makes its own tempdir), so it carries only the marker. The vacuum guard is kept verbatim — the parent asserts the child ran **exactly one test**, so a rename cannot silently turn the guard into a no-op.

### The requirement is mechanical, not documented

`with_data_override` / `deny_uv` / `allow_uv` all call `assert_isolated()` first. A test added later that mutates the environment without the guard fails immediately with an explanation, instead of quietly reintroducing the race:

```
this helper mutates process-global env — the calling test must first
return early on `!isolate_in_child_process("<full::test::path>")` (#4125)
```

---

## Finding 3 — the 1500 ms budgets

These had already been raised from 200 ms once during #4771 to survive parallel-build load. **Raising the number again was not the fix**, because a wall-clock budget is the wrong primitive for what these tests assert.

`full_import_recheck_retries_once_before_reporting_indeterminate` asserts **ordering and invocation count** — "an over-budget first attempt is retried, with the larger budget" — not elapsed time. Proving that through a real sleeping subprocess forced the test to bet that a shell stub could `fork` and append to a counter file inside the budget before being `SIGKILL`ed. That bet is the whole flake, and it is unwinnable by widening.

**Fix:** `verify_full_import_smoke_bounded`'s retry policy is extracted as `recheck_with_one_retry(first, retry, check)` with the check injected — the same seam shape #4771 already used for `ensure_venv_verified`.

| Test | Now asserts | Clock? |
|---|---|---|
| `full_import_recheck_retries_once_before_reporting_indeterminate` | exactly 2 calls, budgets `[first, retry]` in order, retry's label distinguished, result `Indeterminate` | **none** |
| `full_import_recheck_does_not_retry_a_decided_outcome` *(new)* | `Passed` and `Failed` each return after exactly 1 call | **none** |
| `full_import_recheck_retry_lets_a_starved_venv_prove_itself` | real subprocess; `Passed` | outcomes fixed by construction |
| `verify_full_import_smoke_bounded_reports_indeterminate_when_both_budgets_lapse` *(new)* | real subprocess through the production entry point; `Indeterminate` | outcome fixed by construction |

The first test now asserts **more** than before: that the retry receives the *larger* budget, not merely that a retry happened. Retrying with the same budget would just move the identical cliff, and the old version could not tell the difference.

### The one remaining time budget, and why it is not a race

`full_import_recheck_retry_lets_a_starved_venv_prove_itself` still drives a real bounded subprocess, because "the retry re-runs the actual check" is worth pinning at that level. It is no longer a race, because the seam — not a counter file written by a process about to be killed — is what distinguishes the two attempts:

- attempt 1 meets `sleep 30` against a **300 ms** budget → cannot finish, `Indeterminate` by construction;
- the stub is rewritten to `exit 0` between attempts (exactly the shape of the daemon finishing its warm-boot restores);
- attempt 2 meets `exit 0` against a **30 s** budget → cannot fail to finish.

Both budgets are two orders of magnitude away from their decision boundary in the safe direction. Nothing depends on how fast the box forks. The same reasoning covers the new `verify_full_import_smoke_bounded` test (300 ms budgets vs `sleep 30`) and matches the already-non-flaky `bounded_python_check_classifies_timeout_apart_from_failure`, which is unchanged.

**No invocation count is asserted at subprocess level any more** — counting there is precisely what made the old version racy, and the policy tests now own that assertion exactly.

---

## No coverage weakened

Nothing is `#[ignore]`d, cfg-gated, or narrowed. Test count in `trusty-embedderd-py` went **20 → 23**; every behaviour the deleted assertions covered is covered at least as strongly:

| Removed | Replaced by |
|---|---|
| counter-file invocation count (racy) | exact call count **and exact budgets** in the policy test |
| "the retry re-runs a real subprocess" | `Passed` is unreachable without a second real run |
| `ENV_LOCK` | child-process isolation, plus `assert_isolated()` enforcing it |
| `with_data_override` in 8 tests | `resolve_layout_in` — the race removed rather than isolated |

---

## Stability under load

Both findings 2 and 3 are about behaviour under contention, so the demonstration is run contended.

**25 iterations, all 16 cores saturated with busy loops:**

```
cores saturated: 16
RESULT: 25/25 passed, 0 failed
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.71s
```

**30 iterations of the affected tests at 4x core oversubscription:**

```
busy loops: 64 on 16 cores (4x oversubscription)
RESULT: 30/30 passed, 0 failed
test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 11 filtered out; finished in 1.35s
```

---

## Verification

```
$ cargo test -p trusty-embedderd-py
  lib                      23 passed; 0 failed; 0 ignored
  bin trusty-embedderd-py   0 passed; 0 failed; 0 ignored
  tests/e2e.rs              0 passed; 0 failed; 1 ignored (needs a built torch venv)
  doc-tests                 0 passed; 0 failed; 0 ignored

$ cargo test -p trusty-search --no-fail-fast
  lib                                    1444 passed; 5 failed; 1 ignored
  bin trusty-search                       327 passed; 0 failed; 3 ignored
  tests/config_mount.rs                     3 passed; 0 failed
  tests/corpus_open_quarantine_4122.rs      3 passed; 0 failed
  tests/data_dir_forward.rs                 3 passed; 0 failed
  tests/embedder_throughput_bench.rs        4 passed; 0 failed
  tests/integration_tests.rs                7 passed; 0 failed
  tests/ooc_quick_wins.rs                   6 passed; 0 failed
  tests/residency_cold_park.rs              2 passed; 0 failed
  tests/typeahead.rs                        7 passed; 0 failed
  tests/warm_boot_corpus_open_failure.rs    2 passed; 0 failed

    Finished `dev` profile — 0 warnings
$ cargo clippy -p trusty-embedderd-py --all-targets -- -D warnings
    Finished `dev` profile — 0 warnings

$ cargo fmt --check
(no diff)

line-cap: measured 3625 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
test-pointers: resolved 20807 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
$ bash scripts/check_sld.sh
sld-lint: scanned 55 spec doc(s) + 3044 code file(s); 0 error(s), 0 warning(s)
$ bash scripts/check_doc_numbers.sh
doc-numbers: scanned 97 doc(s) / 91 claim(s) (floors 20/20); 4 grandfathered, 0 violations — OK.
OK   trusty-embedderd-py: changelog.d fragment present and valid
```

The **5 lib failures are the macOS FSEvents watcher set tracked as #4731** (`service::watcher::*`, `service::watch_loop::*`) — this host does not deliver FSEvents. Baseline not re-derived, per that issue. Change-specific gates all pass.

## Versions

**Not bumped, and none needed.** `scripts/check-version-parity.sh`:

```
[ OK ] trusty-embedderd-py 0.1.2 — not yet published, nothing to compare
[ OK ] trusty-search 0.42.1 — not yet published, nothing to compare
publish-guard: OK — no version-parity drift detected.
```

Both crates are already at the unpublished 0.42.1 / 0.1.2 that #4771 cut, so these fixes ride that same release.

## Changelog

Fragments, not `CHANGELOG.md`:

- `crates/trusty-search/changelog.d/4771-probe-timeout-ceiling.md` (Fixed)
- `crates/trusty-embedderd-py/changelog.d/4771-recheck-retry-seam.md` (Changed — internal seams, no behaviour change)


bobmatnyc:
## trusty-review verdict: APPROVE (grade A-)

Automated review via `mcp__trusty-review__review_pr` (model: us.anthropic.claude-sonnet-4-6).

**Summary:** Solid refactor that introduces injectable seams for the retry policy and venv-layout derivation, caps the unbounded probe-budget multiplier, and replaces the racy ENV_LOCK+set_var pattern with child-process isolation — all with well-targeted tests. One advisory concern about the harness-output-format assumption in the stale-name guard.

**Findings (3, all low/medium severity, none blocking):**
1. `bootstrap_tests.rs:68` (medium, confidence 0.65) — `isolate_in_child_process`'s stale-name detection relies on matching the literal string `"test result: ok. 1 passed"` in child stdout. This is not a stable Rust-toolchain API and could silently become a no-op across toolchain upgrades. Advisory only; consider `--format json` in a follow-up.
2. `bootstrap_tests.rs:560` (low, confidence 0.85) — noted as correct/sound; no action needed (TOCTOU concern ruled out, sequential not concurrent).
3. `bootstrap_tests.rs:44` (low, confidence 0.5) — advisory note on `current_exe()` behavior under `cargo nextest`; not applicable to this repo's plain `cargo test` harness.

Reviewer-context verification performed independently against the PR diff (0ee9ace3):
- `MAX_PROBE_TIMEOUT_MULTIPLIER = 3` caps only the multiplier; base timeout remains the operator-tunable `TRUSTY_EMBEDDERD_STARTUP_TIMEOUT_SECS` (default 30s); `checked_mul` removes the `Duration * u32` overflow panic. Confirmed at `graceful_bootstrap.rs`.
- Env-isolation uses the #4213/#3954-style child-process re-exec technique with a pure `resolve_layout_in` seam, explicitly NOT `#[serial]` (doc comment cites both issues directly).
- The retry-policy test is now clock-free (injected closure, no real sleep); the one remaining real-subprocess test is fixed by construction (stub overwritten between attempts) rather than a timing race.

Proceeding to merge per owner authorization ahead of the 0.42.1 cut.

PR #4793, opened 2026-08-04
fix(trusty-agents): gate ctrl_turn local-failure recovery on the failure, not a preference (#4788)

### LINKED ISSUE #4788: fix(trusty-agents): ctrl_turn dispatch still gates local-failure recovery on `local_qualifies`, leaking a raw Connection refused to the user
`run_ctrl_turn_via_rest` in `crates/trusty-agents/src/ctrl/ctrl_turn/dispatch.rs` carries the error arm `Err(e) if local_qualifies && local_cfg.fallback_on_error =>`, with a bare `Err(e) => return Err(e)` below it. This is the byte-identical sibling of the arm fixed on `run_pm_task_with_history` in #4779 — that PR's changelog fragment explicitly scopes its guarantee to the one path and names this one as still unfixed. This path drives the standalone / no-PM-attached ctrl REPL, a distinct user-facing conversational surface.

`local_qualifies` ANDs in `local_inference.enabled`, so a user who sets `local_inference.enabled = false` gets no retry ever — the setting meaning *avoid Ollama* is precisely what makes an Ollama failure unrecoverable. Combined with any on-disk ctrl config still carrying an `ollama/` model slug, the user's assistant reply is the raw transport error:

```
error sending request for url (http://localhost:11434/v1/chat/completions): Connection refused (os error 61)
```

The owner hit exactly this on the sibling path. Recovery has to gate on the failure, never on a preference flag.

bobmatnyc:
## The defect

`run_ctrl_turn_via_rest` in `crates/trusty-agents/src/ctrl/ctrl_turn/dispatch.rs` — the standalone / no-PM-attached ctrl REPL's conversational surface — carried the error arm

```rust
Err(e) if local_qualifies && local_cfg.fallback_on_error => { … }
Err(e) => return Err(e),
```

byte-identical to the arm #4779 replaced on `run_pm_task_with_history`. That PR's changelog fragment explicitly scoped its guarantee to the one path and named this one as still unfixed.

`local_qualifies` ANDs in `local_inference.enabled`, so a user who sets that to `false` gets **no retry ever** — the setting meaning *avoid Ollama* was exactly what guaranteed an Ollama transport error was unrecoverable. Combined with any on-disk ctrl config still carrying an `ollama/` model slug, a raw

```
error sending request for url (http://localhost:11434/v1/chat/completions): Connection refused (os error 61)
```

reached the user as their assistant's reply. The owner hit exactly this on the sibling path.

## The fix

Both conversational surfaces now share **one** decision function — `local_fallback::local_failure_action`, introduced by #4779 — reached from ctrl_turn via a `pub(crate)` re-export rather than a copy. Its contract is unchanged: `Propagate` when the failure is not a local-adapter transport failure, `RetryRemote(model)` when a genuinely distinct remote model is configured, `Explain` otherwise. **No preference flag is an input to whether recovery happens.**

Genuine API errors (429/401/500) from either a local or a remote model, and remote transport failures, still propagate unchanged.

Two deliberate deviations from a blind copy of #4779, both about *this* route's specifics:

- **The retry keeps `registry_arc`** rather than calling `local_fallback::retry_remote`. That helper deliberately retries with an empty registry (the PM route's conversational fast path needs no tools), but this route's tools drive the end-of-turn side-effect drain — `start_pm` / `initiate_self_task` / `stop_task`. Dropping them on the retry would be a silent regression.
- **`effective_ctrl_turn_model` is extracted** so the tests exercise the actual routing instead of restating it. It pins the mechanism the old flag-gated arm missed: with the gate OFF the route does **not** stop attempting local models — it falls through to the agent's OWN configured model, which for any `ollama/` slug is itself local.

`fallback_on_error` is now consulted by no dispatch path at all; it survives only as a `/local` REPL readout, and `local_fallback.rs`'s doc comment (which named this route as a remaining consumer) is corrected to say so.

## Tests

Mirrors #4779's test approach for this path — four tests in `ctrl::ctrl_turn::dispatch::tests`, driving the same two inputs the production arm computes:

| Test | Covers |
|---|---|
| `ctrl_turn_recovers_in_both_local_gate_states` | **both `local_inference.enabled` states** crossed with a local transport failure; recovery asserted in both, and the message must not leak `Connection refused` / `os error` |
| `ctrl_turn_gate_on_falls_back_to_a_distinct_remote_model` | the retry target is genuinely remote, never the slug that just failed |
| `ctrl_turn_gate_off_still_attempts_the_agents_own_local_model` | the gate-off fall-through that produced the unrecoverable local call |
| `ctrl_turn_propagates_real_errors` | a remote transport failure and a local server that *answered* with an error both propagate |

## Changelog

New fragment `crates/trusty-agents/changelog.d/4788-ctrl-turn-local-fallback.md`. #4779's fragment (`3766-ctrl-local-model-fallback.md`) is still in `changelog.d` and unreleased, so its "the `ctrl_turn::dispatch` path carries the same defect and is NOT fixed here" scoping is updated — the guarantee now covers both surfaces.

## Gates — rung 3 (localized behavior inside one crate), exit codes checked explicitly

```
cargo test -p trusty-agents                            EXIT=0   3332 passed; 0 failed; 11 ignored (lib)
                                                                + 9 integration suites, all ok
cargo clippy -p trusty-agents --all-targets -- -D warnings   EXIT=0
bash scripts/check_line_cap.sh                         EXIT=0   3628 files, 0 violations
bash scripts/check_changelog_fragment.sh               EXIT=0   scanned 6 paths; 1 crate recorded
```

The first `cargo test` run hit an unrelated flake — `api_e2e::test_health_returns_version` failed with `Error: Address already in use (os error 48)`, a port collision in the integration harness. It passes on re-run (`6 passed; 0 failed`), and this diff touches zero files under `src/api/` or `tests/`. The clean full run above is the one reported.

Closes #4788


bobmatnyc:
## Review round 2 — both HIGHs fixed, MEDIUM split out, LOW blocked by the compiler

Pushed as `8bc72b18`.

### HIGH 1 — double-execution: fixed, option (a)

The finding is correct and my original reasoning was too narrow. I verified the mechanism rather than taking it on faith: `chat_with_tools_gated` is `for turn in 0..max_turns` with `dispatch_turn(...).await?` at the top of each iteration and tool dispatch *after* it, so with `max_turns = 2` turn 0 can execute a tool and turn 1 can then die on transport. `build_ctrl_registry` registers the ~20 immediate-effect tools you listed alongside the three deferred ones I reasoned about.

**Took (a) — empty registry via `retry_remote` — not (b).** Reasoning:

- (b) needs a trustworthy "did any tool execute" signal, and `chat_with_tools_gated` returns nothing about tool calls on the `Err` path. Getting one means threading a counter through the crate's hottest shared code path, where a single mistake silently reintroduces exactly this hazard. For a recovery path whose entire job is "don't show the user a raw transport error," that is the wrong risk to take on.
- (a) has zero classification burden and no shared-infra change. There is no tool list to keep correct as tools are added — a new mutating tool cannot become double-executable by omission.
- **My side-effect-drain reasoning survives intact.** The three deferred tools write to pending slots that are unaffected by the failed attempt and still drain after the retry returns. (a) gives up only the retry's ability to *initiate* new tool calls, not any side effect already captured.
- It makes both surfaces share one *retry*, not merely one decision function — which is the point of the PR.

Considered and rejected: a name-allowlist retry via the existing `allowed_tools` parameter. It works mechanically, but it means hand-classifying ~20 tools as safe/unsafe and keeping that judgment correct forever; one wrong call is a duplicate commit.

You are also right that the new arm is *broader* than the old one. That widening is the fix, so the retry had to get narrower to compensate.

### HIGH 2 — tests that could not fail: fixed, with red/green proof

Correct, and the sharpest finding here. The four decision tests call a helper that re-derives the inputs; they never touch `run_ctrl_turn_via_rest`.

Added `ctrl_turn_rest_recovers_from_a_real_local_transport_failure`, which drives the real function: `OLLAMA_HOST` on port 1 (reserved, never listening) so `chat_with_tools_gated` returns a genuine transport `Err`, agent model on an `ollama/` slug, `local_inference.enabled = false` **and** `fallback_on_error = false` — the owner's exact shape, with both flags the old guard needed set to false.

**Proof, as requested. Reverted the arm to the pre-fix guard:**

```
test ctrl::ctrl_turn::dispatch::tests::ctrl_turn_rest_recovers_from_a_real_local_transport_failure ... FAILED

panicked at crates/trusty-agents/src/ctrl/ctrl_turn/dispatch.rs:634:28:
a local transport failure must not reach the user as Err: raw chat completion POST failed

Caused by:
    0: error sending request for url (http://127.0.0.1:1/v1/chat/completions)
    1: client error (Connect)
    2: tcp connect error
    3: Connection refused (os error 61)

test result: FAILED. 8 passed; 1 failed
```

The other four passed while reverted — your finding, reproduced exactly. Note the failure is the live symptom verbatim, not a proxy for it.

**Restored, same command:**

```
test ctrl::ctrl_turn::dispatch::tests::ctrl_turn_rest_recovers_from_a_real_local_transport_failure ... ok
test result: ok. 9 passed; 0 failed
```

### MEDIUM — persona: wording softened, follow-up filed

Verified before writing: `persona.rs` has zero matches for `local_qualifies` / `fallback_on_error` / `local_failure_action`, and the call ends `.context("persona LLM call failed")?`. Filed **#4819**. Both changelog fragments now name only the two routes that are fixed and point at it. `persona.rs` untouched.

One narrowing worth recording in the issue: unlike the other two, persona has no local-inference fast route at all, so it needs the persona's *own* model to be local — narrower exposure, same defect shape.

### LOW — not applied; E0365

`pub(crate) use` cannot re-export a `pub(super)` module:

```
error[E0365]: `local_fallback` is private, and cannot be re-exported
  --> crates/trusty-agents/src/ctrl/pm_task/mod.rs:36:16
   |
36 | pub(crate) use dispatch::local_fallback;
   |                ^^^^^^^^^^^^^^^^^^^^^^^^ re-export of private `local_fallback`
```

The two paths aren't redundant — the `pub(crate)` mod declaration is what makes the re-export legal. Left a comment saying so, so the next reader doesn't retry it. The only way to get a genuinely single path is moving the file up to `pm_task/local_fallback.rs`; happy to do that if you want it, but it churns a file #4779 just created and rewrites `history.rs`'s import for a crate-internal cosmetic win.

### Gates — rung 3, exit codes checked

```
cargo test -p trusty-agents                                  EXIT=0  3333 passed; 0 failed; 11 ignored
                                                                     + 9 integration suites, all ok
cargo clippy -p trusty-agents --all-targets -- -D warnings   EXIT=0
bash scripts/check_line_cap.sh                               EXIT=0  3628 files, 0 violations
bash scripts/check_changelog_fragment.sh                     EXIT=0  scanned 6 paths; 1 crate recorded
```

Clippy caught one real thing on the new test — `await_holding_lock`, the env guard held across the awaited dispatch. It has to be held (the env must stay set for the whole call), so it takes the crate's existing `#[allow(..., reason = ...)]` convention, sound for the reason `test_env.rs` documents: default `#[tokio::test]` is current-thread, and `#[serial]` keeps other env-touching tests out of the window.

Still not merged.

Issue #2014, opened 2026-07-05
bug(trusty-mpm): pm_guard blocks edits from delegated harness sub-agents (exemption gap)


bobmatnyc:
## Problem
The PM-guard hook that prevents the PM from editing directly also denies Edit/Write operations from *delegated* sub-agents dispatched by the PM. This forces a workaround: setting `TRUSTY_MPM_PM_UNRESTRICTED=1` to get any delegated work done. Sub-agents dispatched by the PM should be exempt from the pm_guard restriction; only the PM's own direct edits should be gated.

## Current Workaround
Set `TRUSTY_MPM_PM_UNRESTRICTED=1` to bypass the guard when running sub-agents. This is unsustainable and defeats the purpose of the guard.

## Solution
Add an exemption to pm_guard for requests originating from delegated sub-agents. The guard should:
- Block direct PM edits (current behavior)
- Exempt sub-agent dispatches (NEW — allow Edit/Write from sub-agents)
- Maintain audit trail showing exemption reason

## Acceptance Criteria
- Sub-agents dispatched by PM can call Edit/Write without `TRUSTY_MPM_PM_UNRESTRICTED=1`
- Direct PM Edit/Write still blocked by pm_guard (existing behavior preserved)
- Guard logs the exemption reason (e.g., "sub-agent dispatch from PM")
- Current workaround (`TRUSTY_MPM_PM_UNRESTRICTED=1`) remains functional for backward compat
- Add test covering both direct PM edits (blocked) and sub-agent edits (allowed)
- Quality gate green: `cargo build --workspace`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`
- Suggested branch: `fix/tm-pm-guard-subagent-exemption`

PR #4796, opened 2026-08-04
feat(trusty-mpm): deny subagent fan-out via PreToolUse guard

### LINKED ISSUE #4784: subagents spawn untyped agents, falling back to general-purpose
**Defect.** The PM types every dispatch (`rust-engineer`, `code-critic`, `ticketing`, `version-control`). Its subagents do not. A subagent that calls the Agent tool without `subagent_type` gets `general-purpose` by default, bypassing the 42-agent roster deployed at the user tier. The delegation discipline the framework enforces at the PM level does not propagate one level down.

Distinct from #2005, #1984, #2002, DOC-34 (PM itself running general-purpose, fixed 2026-07-04) and #4451 (agent unreachability, fixed) — those covered the roster or the PM's own dispatches. This is subagent-level dispatch typing, unaddressed by either.

**Evidence, observed live 2026-08-04 by the repo owner.** A `rust-engineer` working PR #4759 spawned a child that surfaced as:

```
general-purpose  Rewriting 4759-inplace-relaunch-compiled-prompt.md fragment
```

Two faults in one dispatch:
1. **Untyped** — no `subagent_type`, so it defaulted to `general-purpose` instead of a roster agent.
2. **Wrong agent even if typed** — rewriting a changelog fragment is documentation work. It should route to the `documentation` agent (model haiku per the routing table), not a general-purpose agent inheriting an engineer-tier model.

**Why this matters.** `general-purpose` carries none of the roster agent's system prompt, model tier, or domain rules. Work silently drops to an unspecialised agent at whatever model the parent was using. The roster is deployed and healthy — verified this session: 42 agents present at `$CLAUDE_CONFIG_DIR/agents`. This is not an agent-loss regression; it is a dispatch-discipline gap.

**Mechanism check (confirmed).** `BASE-AGENT.md` (`crates/trusty-code/src/assets/agents/BASE-AGENT.md`) says nothing about how a subagent should type its OWN Agent-tool dispatches — it only governs how the subagent receives delegation from the PM (the "PM Directives Do Not Bind You" and "PM Authority & Escalation" sections) and forbids delegating execution back to the user ("Self-Action Imperative"). This is an ADDITION, not a strengthening of existing language.

**Resolution.** `BASE-AGENT.md` should require what the PM instructions already require, for any Agent-tool call a subagent itself makes:
- every Agent-tool call names an explicit `subagent_type` and an explicit `model`
- documentation work — changelog fragments, README, docs, doc comments — routes to `documentation`
- a subagent that cannot identify a fitting roster agent reports back rather than silently falling back to `general-purpose`

Affects every dispatch the harness makes at any depth — belongs in the 1.3.5-2 release.

bobmatnyc:
Closes #4784

## What

`tm hook --pm-guard` now denies `Task`/`Agent` when the session issuing the call is itself a subagent. The PM keeps dispatching — that is the entire orchestration model and nothing about it changes.

## The detection mechanism (the hard part)

Subagent-ness is resolved from two **automatic markers** — a harness or a spawn helper sets them, no per-call model or human decision is involved. No prose, transcript-path shape, or agent-name heuristic is consulted.

| Signal | Source | Covers |
|---|---|---|
| `agent_id` in the `PreToolUse` payload | Claude Code, documented as present "only when the hook fires inside a subagent call" | native `Task`/`Agent` dispatch |
| `CLAUDE_MPM_SUB_AGENT` in the environment | trusty-agents (`subprocess/spawn.rs`, `agents/claude_code_runner/run.rs`) | out-of-band subagent processes, whose own Claude Code session is top-level and carries no `agent_id` |

`agent_id` is **not newly derived** — it is the same signal `pm_guard::payload_is_subagent_dispatch` has used since #2014, reused rather than re-implemented so the two can never disagree. Captured real-world evidence of the field is already in-tree at `tests/tm_hook_delegation_payload.rs:185`.

## Fail-open, stated explicitly

Neither marker present means **ALLOW**. The asymmetry is deliberate and documented in the code: a false DENY lands on the PM and halts every dispatch in the system; a false ALLOW merely reproduces the behaviour that existed before this guard.

`SendMessage` is **never** denied, in any caller context. It is how a blocked agent reports back and gets resumed — denying it would strand the very agent the guard redirects.

## Placement

The check sits ahead of Guard 1 (`CLAUDE_MPM_SUB_AGENT`) and Guard 4 (`agent_id`), for the same structural reason the #3977 worktree guard does: both exemptions early-return ALLOW *precisely* when the caller is a subagent, which is the only case this rule fires on. Placed after either, it would be a guaranteed no-op. Guards 2/3 (`TRUSTY_MPM_DISABLE_HOOKS`, `TRUSTY_MPM_PM_UNRESTRICTED`) stay ahead of it — they are human escape hatches, not automatic markers — and that ordering is pinned by a test.

## Why a hook and not `tools:` frontmatter

Re-verified against the code, all three claims from the prior investigation hold:

1. `tools:` **is** mechanically enforced — `trusty-agents-common/src/agents/builder.rs` parses, merges across the `extends` chain, and re-emits it.
2. It is **allow-list only**. `builder.rs:790-801` treats a present key as a wholesale replacement and `Some(vec![])` as deny-all; there is no exclusion grammar anywhere in the parser.
3. **MCP tools are gated by the same list** — they are ordinary `mcp__server__tool` names that must appear in it, so an enumerated allow-list across ~40 agents would silently break real, often-deferred MCP access.

## Tests

Pure policy (`pm_guard_fanout`, 5 unit tests) plus 7 end-to-end tests through the real binary in `tests/tm_hook_pm_guard.rs`: deny from a native subagent, deny from an MPM-env subagent, **allow from the PM**, `SendMessage` never denied, fail-open on indeterminate context, escape hatches still win, and a subagent's ordinary Edit/Read surface unchanged.

**Fail-before / pass-after**, with the guard neutralized to `caller_is_subagent = false`:

```
failures:
    pm_guard_denies_agent_dispatch_from_native_subagent
    pm_guard_denies_task_dispatch_from_mpm_subagent_env

test result: FAILED. 38 passed; 2 failed; 0 ignored
```

Exactly the two deny arms fail; every ALLOW arm passes both ways, proving those pin non-regression rather than the new behaviour. Restored:

```
test result: ok. 40 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Gates — rung 5

```
cargo fmt --check                                    → clean
cargo clippy -p trusty-mpm --all-targets -D warnings → exit 0
cargo test -p trusty-mpm                             → 4371 passed; 1 failed (baseline, below)
cargo check --workspace                              → 2 failed (baseline, below)
bash scripts/check_line_cap.sh    → 3629 files, 0 violations — OK
bash scripts/check_sld.sh         → 55 specs + 3046 files, 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh → OK, scanned 5 paths
```

### Baseline failures — non-involvement proven, not asserted

`execute_doctor_against_test_daemon` — the documented environmental flake. `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/` returns **0 files**, and it passes in isolation:

```
test result: ok. 1 passed; 0 failed; 4378 filtered out
```

`cargo check --workspace` — exactly 2 crates fail, both on the missing pnpm build artifact:

```
error: could not compile `trusty-mpm-gui` (lib) due to 1 previous error
```

`git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/` returns **0 files**. Every other workspace crate compiled.

No coverage was deleted, `#[ignore]`-d, `cfg`-gated, or `--exclude`-d anywhere in this change.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4784-deny-subagent-fanout.md
+++ b/crates/trusty-mpm/changelog.d/4784-deny-subagent-fanout.md
@@ -1,0 +1,6 @@
+Added
+
+- `tm hook --pm-guard` now denies `Task`/`Agent` dispatch when the calling session is itself a subagent (closes [#4784](https://github.com/bobmatnyc/trusty-tools/issues/4784))
+  - the PM keeps dispatching; only fan-out from within a subagent is blocked
+  - `SendMessage` is never denied, so a blocked agent can always report back
+  - fails OPEN — an indeterminate caller context allows the dispatch rather than risking a false deny against the PM
--- a/crates/trusty-mpm/src/bin/tm/commands/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/mod.rs
@@ -58,6 +58,7 @@ pub(crate) mod pm_guard;
 pub(crate) mod pm_guard_bash;
 pub(crate) mod pm_guard_budget;
 pub(crate) mod pm_guard_deny_by_default;
+pub(crate) mod pm_guard_fanout;
 pub(crate) mod pm_guard_routing;
 pub(crate) mod project;
 pub(crate) mod projects;
--- a/crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs
@@ -78,6 +78,14 @@
 //! in the file that pierces an automatic-marker exemption; see that call
 //! site's comment for why it must never be routed through [`evaluate_tool`]
 //! instead.
+//! **Subagent fan-out denial (issue #4784):** [`pm_guard`] calls
+//! [`crate::commands::pm_guard_fanout::evaluate_subagent_fanout`] DIRECTLY,
+//! before Guards 1 and 4, denying `Task`/`Agent` when the calling session is
+//! itself a subagent — the PM keeps dispatching, `SendMessage` is never
+//! denied, and an indeterminate caller context fails OPEN. It sits ahead of
+//! both exemptions for the same reason the worktree guard does: both return
+//! ALLOW exactly when the caller IS a subagent, the only case the rule fires
+//! on. See that module's doc for the two markers it trusts and why.
 //! Test: `evaluate_tool_*`, `payload_is_subagent_dispatch_*`, and
 //! `build_pretooluse_deny_response_*` below cover the tool policy, sub-agent
 //! exemption, and JSON shape; the Bash-command classifier (composition and
@@ -96,6 +104,7 @@ use crate::commands::pm_guard_bash::{
 };
 use crate::commands::pm_guard_budget::{self, BudgetDecision, DEFAULT_FILE_CHANGE_BUDGET};
 use crate::commands::pm_guard_deny_by_default::{self, PERSONA_DENY_REASON};
+use crate::commands::pm_guard_fanout;
 use crate::commands::pm_guard_routing::{GENERIC_ENGINEER_HINT, delegation_hint_for_path};
 
 /// Escape-hatch env var: `TRUSTY_MPM_PM_UNRESTRICTED=1` disables all PM
@@ -271,6 +280,25 @@ pub(crate) async fn pm_guard(url: &str) -> anyhow::Result<()> {
         }
     }
 
+    // #4784: a subagent must never dispatch further subagents. Placed here,
+    // before Guards 1 and 4, for the same structural reason as the worktree
+    // guard above: both exemptions early-return ALLOW precisely when the
+    // caller IS a subagent, which is the only case this rule fires on — so a
+    // fan-out check placed after either one, or folded into `evaluate_tool`,
+    // would be a guaranteed no-op. Guards 2/3 stay ahead of it (they are
+    // human escape hatches, not automatic markers — see their comments).
+    // It fails OPEN: `caller_is_subagent` reports false whenever neither the
+    // payload's `agent_id` nor `CLAUDE_MPM_SUB_AGENT` is present, so an
+    // unrecognised context allows the dispatch rather than blocking the PM.
+    if let Some(reason) = pm_guard_fanout::evaluate_subagent_fanout(
+        tool_name,
+        pm_guard_fanout::caller_is_subagent(&payload),
+    ) {
+        audit_denied_tool(url, session_id, tool_name, reason).await;
+        println!("{}", build_pretooluse_deny_response(reason));
+        return Ok(());
+    }
+
     // Guard 1: never block a nested MPM sub-agent for anything else — it is
     // doing the delegated work the PM is being steered toward. Moved here
     // (originally the first line of this function, short-circuiting before
@@ -402,7 +430,7 @@ fn pm_unrestricted() -> bool {
 /// `payload_is_subagent_dispatch_false_when_absent_or_empty`; exercised end to
 /// end via `pm_guard_native_subagent_dispatch_allows_edit` in
 /// `tests/tm_hook_pm_guard.rs`.
-fn payload_is_subagent_dispatch(payload: &serde_json::Value) -> bool {
+pub(crate) fn payload_is_subagent_dispatch(payload: &serde_json::Value) -> bool {
     payload
         .get("agent_id")
         .and_then(|v| v.as_str())
--- a/crates/trusty-mpm/src/bin/tm/commands/pm_guard_fanout.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/pm_guard_fanout.rs
@@ -1,0 +1,220 @@
+//! `tm hook --pm-guard` — subagent fan-out denial (issue #4784).
+//!
+//! Why: the orchestration model is one level deep. The PM dispatches
+//! specialists; a specialist does its own work and reports back. A subagent
+//! that dispatches *further* subagents multiplies context cost, hides the
+//! resulting work from the PM's tracking (the PM sees one delegation, not the
+//! tree beneath it), and produces exactly the un-owned fan-out ADR-0024 calls
+//! a single-edge leaf violation. Agent prose alone does not stop it: a model
+//! holding the `Agent`/`Task` tool will use it. The `tools:` frontmatter route
+//! cannot express the rule either — it is an ALLOW-list with no "everything
+//! except X" grammar, and it gates MCP tools through the same list, so
+//! enumerating ~40 agents' tool sets to omit two names would silently break
+//! real MCP access. `PreToolUse` is the one seam that can deny a single named
+//! tool without touching anything else.
+//!
+//! What: [`evaluate_subagent_fanout`] denies [`SUBAGENT_DISPATCH_TOOLS`]
+//! (`Task` and `Agent` — both, since Claude Code renamed the tool between
+//! releases) when, and only when, the calling session is itself a subagent.
+//! [`caller_is_subagent`] answers that from two automatic, spawner/harness-set
+//! markers, never from prose or heuristics:
+//!
+//! 1. `agent_id` in the `PreToolUse` payload — documented by Claude Code as
+//!    present "only when the hook fires inside a subagent call", the same
+//!    signal [`super::pm_guard::payload_is_subagent_dispatch`] has used since
+//!    issue #2014. Reused, not re-derived, so the two can never disagree.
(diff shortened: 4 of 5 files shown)


bobmatnyc:
## code-critic — Verdict: APPROVE

Rung 5 (process lifecycle), adversarial round. Reviewed in an isolated worktree at `origin/feat/4784-subagent-dispatch-guard` (7259240f), spec + code + tests only.

No issues found at >80% confidence. **APPROVED for next pipeline stage.**

### Claim verification

| # | Claim | Result |
|---|---|---|
| 1 | Detection via `agent_id` + `CLAUDE_MPM_SUB_AGENT`; inversion is sound | **VERIFIED** (completeness: PARTIAL, fail-open direction) |
| 2 | Placement ahead of Guards 1/4 is load-bearing and pinned | **VERIFIED** by mutation — and the "vacuous-green" expectation is **REFUTED**: the tests go red, not vacuous |
| 3 | Fails OPEN on every indeterminate path, no panic path | **VERIFIED** |
| 4 | `SendMessage` never denied; nothing else an agent needs is caught | **VERIFIED** |
| 5 | Neutralizing the guard fails exactly two tests | **VERIFIED** exactly |
| 6 | Policy/env split; env path genuinely covered | **VERIFIED** |
| 7 | Denial message states the remedy and is accurate | **VERIFIED** (one caveat in Notes) |

### 1 — The inversion (the thing that could have been catastrophic)

`agent_id` has meant "exempt this caller" since #2014; this PR makes it mean "deny this caller". The dangerous failure would be `agent_id` appearing on a **top-level PM** dispatch — that would deny every delegation in the system, and under the old semantics the same condition failed silently permissive, so nobody would have noticed it was wrong.

That is refuted by empirical evidence already in the repo. `crates/trusty-mpm/src/bin/tm/commands/hook_payload.rs:154-171` is the verbatim #2864 Step-0 live capture of Claude Code 2.1.220's real `PreToolUse` for `"tool_name": "Agent"` dispatched by a top-level session. It carries `tool_use_id` and `transcript_path` and **no `agent_id`**; `agent_id` first appears on `SubagentStop` (`hook_payload.rs:230-240`). The PM's own dispatch is not caught.

Existing-consumer non-regression: `payload_is_subagent_dispatch`'s body is unchanged — the only edit is `fn` → `pub(crate) fn` (`pm_guard.rs:433`). It has exactly two call sites in the whole workspace: Guard 4 at `pm_guard.rs:324` and the new `pm_guard_fanout.rs:122`. The exemption path is byte-identical, and `pm_guard_subagent_keeps_its_working_tool_surface` pins it.

Two-signal completeness — PARTIAL, in the safe direction. The two markers cover the two spawn classes that exist in this workspace (native Task/Agent; trusty-agents' `subprocess/spawn.rs:271` and `agents/claude_code_runner/run.rs:214`). A third shape — a `claude`/`claude -p` session a subagent launches itself through `Bash` — carries neither marker and slips through. That is a **fail-OPEN miss**: the guard is a no-op there, not a false deny. A coverage limit, not a defect, and closing it is not this PR's job.

### 2 — Placement, by mutation

Moved the check after Guard 1 (before Guard 4):

```
test result: FAILED. 39 passed; 1 failed
    pm_guard_denies_task_dispatch_from_mpm_subagent_env
```

Moved it after both Guards 1 and 4 (equivalent to folding it into `evaluate_tool`):

```
test result: FAILED. 38 passed; 2 failed
    pm_guard_denies_agent_dispatch_from_native_subagent
    pm_guard_denies_task_dispatch_from_mpm_subagent_env
```

Both moves fail loudly. The guard does not "still pass when moved" — each exemption is pinned by its own test, and each test names the exemption it pins. `pm_guard_fanout_yields_to_operator_escape_hatches` pins Guards 2/3 staying ahead. Mutations fully reverted; `git diff HEAD` empty.

### 5 — Fail-before/pass-after

Neutralized the guard at the call site (`pm_guard.rs:293-300` replaced with a no-op), full `cargo test -p trusty-mpm`:

```
test result: FAILED. 38 passed; 2 failed; 0 ignored
    pm_guard_denies_agent_dispatch_from_native_subagent
    pm_guard_denies_task_dispatch_from_mpm_subagent_env
```

Exactly the two claimed, nothing else across the whole crate. The ALLOW-arm tests pass both ways by design — that is the non-regression pin working, not dead coverage.

### 3 — Fail-open, traced

Every degradation reaches ALLOW before this module is called: no stdin (`pm_guard.rs:230`), no `tool_name` (`pm_guard.rs:234`). Non-object and wrong-typed payloads resolve through `Value::get`/`as_str` to `None` with no panic. `caller_is_subagent` (`pm_guard_fanout.rs:122`) is `get`/`as_str`/`var_os` only. `build_pretooluse_deny_response` (`pm_guard.rs:580`) is a `json!` over a `&str` — not fallible. No path can fail closed by crashing the hook.

### 6 — Env coverage is real, not asserted

`evaluate_subagent_fanout` takes the resolved context as a `bool`, so no unit test in this module touches the environment — verified: `fails_open_when_caller_context_is_indeterminate` calls `payload_is_subagent_dispatch`, not `caller_is_subagent`. The env arm runs per-process through the real binary, and mutation A proved it load-bearing (removing the call site is what makes `pm_guard_denies_task_dispatch_from_mpm_subagent_env` fail). No new `set_var` race was introduced into a threaded test binary.

### Gates

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm` | `4394 passed; 0 failed; 7 ignored` + every integration target green, exit 0 |
| `cargo test -p trusty-mpm -- --include-ignored` | `4399 passed; 2 failed` — both documented baselines, see below |
| `bash scripts/check_line_cap.sh` | `3631 files, 0 violations — OK` |
| `bash scripts/check_sld.sh` | `0 errors, 0 warnings` |
| `bash scripts/check_changelog_fragment.sh` | `OK trusty-mpm: fragment present and valid` |
| `bash scripts/check_doc_numbers.sh` | `0 violations — OK` |

Baseline non-involvement, proven not asserted. `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/ crates/trusty-mpm/src/daemon/` returns **0 files**; the diff is five paths, all under `commands/`, `tests/`, and `changelog.d/`. Both `--include-ignored` failures (`client::executor::tests::execute_doctor_against_test_daemon` — the documented timing flake; `daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many` — a benchmark fixture `NotFound`) pass isolated:

```
test daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many ... ok
test result: ok. 2 passed; 0 failed
```

Nothing was ignored, cfg-gated, excluded, or narrowed to make anything green.

Doc-pointer check: every `Test:` citation in the new module resolves to exactly one real test (10/10 checked by name). The repo-wide `check_test_pointers.sh` scan exceeded 10 minutes on this machine and was cancelled deliberately; the diff-relevant subset was verified directly instead.

### Notes

- **`SendMessage is never blocked — use it to report back.`** The operative claim is true (`SendMessage` is not in `SUBAGENT_DISPATCH_TOOLS`, and `pm_guard_never_denies_send_message` asserts it in both caller contexts). Whether a *child* can `SendMessage` its parent PM, as opposed to the documented parent→child direction, is not established anywhere I could check. The primary remedy the message gives — "do this work yourself, or stop and report back to the PM naming the units that should run in parallel" — is correct without `SendMessage`, so the sentence is at worst a dead-end suggestion, not misdirection. Below the finding bar; flagging only so the phrasing is a known choice.
- **`CLAUDE_MPM_SUB_AGENT` is absent from `INHERITED_SESSION_MARKERS`** (`core/claude_env_scrub.rs:89-110`). A `tm`-launched PM session spawned from inside a trusty-agents subagent therefore inherits the marker, and would now be **denied** on dispatch where it was previously fully exempt (Guard 1 lifted *every* pm_guard check for it). I could not establish >80% confidence that a legitimate flow reaches this: it requires a subagent to launch a PM session, which is itself the un-owned fan-out this PR exists to stop, and the pre-PR behaviour there — all guards silently off — was strictly worse. Recording the mechanism so it is a known edge rather than a surprise.
- Operator escape hatches (`TRUSTY_MPM_PM_UNRESTRICTED`, `TRUSTY_MPM_DISABLE_HOOKS`, #3981) were out of scope per dispatch and are not re-litigated. `pm_guard_fanout_yields_to_operator_escape_hatches` pins that they still lift this guard.

Issue #2038, opened 2026-07-06
fix(install.sh): accept Linux arm64 + honest partial-failure exit code/messaging


bobmatnyc:
Sub-issue of #1992.

Two bootstrap-script bugs in the repo-root `install.sh`:

1. **`detect_target()` hard-rejects Linux aarch64** (`install.sh:122-134`): Currently has `die "Linux aarch64 is not yet supported"` — once the release workflow publishes a Linux arm64 asset (companion release issue #2037), install.sh must accept it and map to the correct target triple.

2. **`post_install()` always prints success and exits 0** (`install.sh:386-407`): Always prints "Installation complete!" and exits 0 even on partial/failed installs (#1992 finding 4) — it must reflect actual per-component success/failure and exit non-zero when the install did not fully succeed. Keep the single-URL `curl | sh` contract intact.

**Coordinate the arm64 target-triple naming with issue #2037** so the asset URL install.sh builds matches what CI publishes.

Issue #3540, opened 2026-07-21
release.yml: any Linux build-leg failure silently skips release creation (froze trusty-mpm binary release at v0.16.0)


bobmatnyc:
## Summary

`.github/workflows/release.yml`'s `release` job (create/update GitHub Release + upload assets) is silently **skipped** whenever the `build` matrix job reports `result == 'failure'` — even when the primary (Tier-1, macOS `aarch64-apple-darwin`) leg built successfully. This is the confirmed root cause of trusty-mpm's GitHub binary release being frozen at v0.16.0.

## Evidence

Two consecutive tagged release runs reproduced the same failure mode:

- `trusty-mpm-v0.19.28` (run `29790108859`)
- `trusty-mpm-v0.19.27` (earlier attempt)

In both runs:
- `aarch64-apple-darwin` (macOS, Tier-1/required) build: **SUCCEEDED**
- `aarch64-unknown-linux-gnu` build: **FAILED** — `unable to find dynamic system library 'ssl'`/`'crypto'`
- `x86_64-unknown-linux-gnu` build (also Tier-1/required, NOT `continue-on-error`): **FAILED** — same `openssl-sys` error

Root cause of the *Linux* build failure itself (tracked separately, NOT fixed by this issue): `openssl-sys` (pulled in via `git2` → `trusty-common` → `trusty-mpm`, default features = dynamic system OpenSSL) can't be linked during `cargo zigbuild` cross-linking — zig's linker has no dynamic `libssl`/`libcrypto` to find on the cross-target sysroot.

## Actual bug (this issue)

Because both real Linux legs failed, `needs.build.result == 'failure'`. The `release` job's `if:` (previously around `.github/workflows/release.yml:1010-1014`) was:

```yaml
if: >-
  needs.setup.outputs.bundled != 'true' &&
  needs.setup.outputs.dry_run != 'true' &&
  needs.build.result != 'cancelled' &&
  needs.build.result != 'skipped'
```

This expression contains **no** status-check function (`success()`, `failure()`, `cancelled()`, `always()`) — only string comparisons against `needs.build.result`. Per GitHub Actions semantics, an `if:` expression with no status-check function is implicitly ANDed with `success()`. So despite the comment's stated intent ("we explicitly allow result='failure' on `build`"), the job was in practice skipped any time `build` failed for ANY reason, Tier-1 or not.

[...]

PR #4822, opened 2026-08-04
ci(release): build the linux arm64 leg natively and stop reporting partial releases as success

### LINKED ISSUE #2533: release.yml: trusty-search aarch64-unknown-linux-gnu build fails on openssl-sys
## Issue

During the trusty-search 0.32.4 release (release.yml run 29260993090, tag trusty-search-v0.32.4), the **aarch64-unknown-linux-gnu** build target failed, preventing the aarch64 Linux release tarball from being produced.

## Build Failure Details

- **Job:** Build trusty-search aarch64-unknown-linux-gnu
- **Duration:** ~58s before failure
- **Error:** `error: failed to run custom build command for openssl-sys v0.9.116`
- **Other targets:** x86_64-linux-al2023 and other jobs succeeded
- **Impact:** Only the aarch64 Linux release tarball is missing; x86_64-al2023 is healthy

## Root Cause

The aarch64 cross-compile environment likely lacks proper OpenSSL support or vendor configuration. Likely solution involves:
- Adding `openssl = { features = ["vendored"] }` to statically link OpenSSL
- Or configuring cross-compile environment variables for the aarch64 target

## References

- Release run: https://github.com/bobmatnyc/trusty-tools/actions/runs/29260993090
- Tag: trusty-search-v0.32.4
- Not started; reported from #2222 release

bobmatnyc:
Closes #2533

The `aarch64-unknown-linux-gnu` leg of the binary-release workflow has never
produced an asset for trusty-search, and the workflow reported `success` while
doing so. Both halves are fixed here.

## Where the OpenSSL dependency actually comes from

Established before choosing an approach, because it changes the calculus:

```
cargo tree -p trusty-search -i openssl-sys
openssl-sys v0.9.116
├── native-tls v0.2.18
│   ├── hf-hub v0.5.0  ← fastembed v5.13.4 ← trusty-common (feature "hf-hub-native-tls")
│   ├── hyper-tls ← reqwest (pulled to native-tls by feature unification from hf-hub)
│   └── ureq ← hf-hub
└── openssl v0.10.80
```

It is entirely transitive, and this workspace has already standardised on
rustls everywhere else — `reqwest` is `default-features = false` + `rustls-tls`
at the workspace root, `teloxide` carries a comment saying it was switched to
rustls for exactly this reason, `git2` uses `vendored-openssl`, and
`trusty-analyze` already selects `fastembed/hf-hub-rustls-tls`. The single
holdout is `crates/trusty-common/Cargo.toml`'s `hf-hub-native-tls`.

## Approach chosen: native arm64 runner

The leg now builds on GitHub's `ubuntu-24.04-arm` runner (free for public
repos; this repo is public) instead of cross-compiling from x86_64. HOST ==
TARGET, so every pkg-config-dependent `-sys` build script resolves against the
runner's own arm64 `libssl-dev` exactly as the x86_64 leg does. This removes
cross-compilation — the actual root difficulty — rather than working around it,
and it fixes the whole class, not just OpenSSL: any future native dependency
behaves the same way it does on the x86_64 leg.

### Alternatives rejected

**rustls (option 3)** — technically the most attractive: one line in
`crates/trusty-common/Cargo.toml` (`hf-hub-native-tls` → `hf-hub-rustls-tls`,
matching what trusty-analyze already does) removes `openssl-sys` from the graph
outright and drops a native runtime dependency. Rejected *for this PR*, not on
merit: it changes a shipped crate's dependency graph and TLS trust behaviour
(hf-hub's rustls path uses bundled roots rather than the system trust store, so
a corporate MITM proxy that works today would stop working) while the
trusty-search 0.42.1 release chain is in flight in another worktree, and it
would need a changelog fragment in a file that release chain is actively
assembling. **Recommended as a follow-up** — it is a genuine inconsistency in
`trusty-common`, and it would also let the arm64 leg drop its libssl runtime
dependency.

**Vendored OpenSSL (option 2)** — builds OpenSSL from C source, so it would
still have been compiled by zig's clang under the cross setup, which is exactly
the toolchain that turned out to be the *second* problem (see below). Costs
build time on every leg and pins an OpenSSL version for CVE purposes. Solves
less than the native runner for more cost.

**Cross-compile sysroot (option 4)** — correct, most fragile to maintain, and
the only option that keeps the cross-compilation that caused the bug.

## The second failure the native runner exposed

With OpenSSL resolved, the leg then failed inside `numkong` (usearch's SIMD
backend). Under zig's clang the probe results contradict the sources they gate:
clang rejects GCC's `__attribute__((target("neon")))` spelling on AArch64, so
the baseline probe emits `-DNK_TARGET_NEON=0` while the dependent
`NK_TARGET_NEONSDOT=1` / `NK_TARGET_SVE*=1` stay on — and those paths need the
`arm_neon.h` that only the NEON path includes. Result: ~20 `call to undeclared
function 'vdupq_n_s32'` errors (run 30941401868, job 92100578898).

The available zig workaround — forcing `NK_TARGET_*=0`, as this workflow
already does for AL2023's GCC 11 — would ship the arm64 binary of a vector
search engine with its SIMD paths switched off. Instead, crates that pull in
numkong (the ORT-bundling crates) now build their arm64 leg with the runner's
GCC, which is the same toolchain choice their x86_64 leg already makes for the
same class of problem. Crates without numkong keep the glibc-2.17 zigbuild leg,
now native rather than cross.

## Silent-failure half — both mechanisms

Chosen: **make the leg blocking, and audit every leg**. The arm64 leg is a
supported target (Graviton, arm64 containers), so `best_effort` was the wrong
classification — a leg nobody is willing to fail on is a leg nobody reads. It
is now required, so it turns the run red.

Blocking the one known leg does not stop the *next* one being dropped quietly,
so a new `release-completeness-audit` job re-derives the expected asset name for
every matrix entry and compares it against what the build job actually produced:

- missing **required** asset → job fails, run goes red;
- missing **best-effort** asset (AL2023, still deliberately non-blocking) →
  `::warning` annotation + `DROPPED` row, so a partial release is visibly
  partial;
- either way a per-leg table lands in the run summary.

Also added: before packaging, each binary's ELF/Mach-O machine type is asserted
against its target triple. An asset named `…-aarch64-…` containing an x86_64
binary is worse than a missing asset.

## Verification

Real `workflow_dispatch` dry runs on this branch, not a config diff.

**1. The silent-failure fix, proven on a genuinely broken build**
(run [30941401868](https://github.com/bobmatnyc/trusty-tools/actions/runs/30941401868),
the intermediate commit where arm64 still failed):

```
run conclusion: failure        ← the same shape of failure reported `success` in v0.42.0
| Status  | Asset                                            | Leg         |
| OK      | trusty-search-…-aarch64-apple-darwin.tar.gz      | required    |
| OK      | trusty-search-…-x86_64-unknown-linux-gnu.tar.gz  | required    |
| MISSING | trusty-search-…-aarch64-unknown-linux-gnu.tar.gz | required    |
| OK      | trusty-search-…-x86_64-linux-al2023.tar.gz       | best-effort |
```

The audit job also has local unit coverage over all four states (all present /
best-effort dropped → exit 0 + warning / required missing → exit 1 / no
artifacts at all → exit 1).

**2. trusty-search arm64 builds**
(run [30942504925](https://github.com/bobmatnyc/trusty-tools/actions/runs/30942504925),
conclusion `success`, all four legs green including
`Build trusty-search aarch64-unknown-linux-gnu`).

**3. The artifact is genuinely aarch64** — downloaded and inspected, not assumed:

```
file: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (GNU/Linux),
      dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, stripped
e_machine bytes at 0x12: b7 00   (EM_AARCH64)
sha256 recomputed locally == the .sha256 shipped beside it
```

**4. It actually RUNS** — executed in `linux/arm64` containers, not inferred
from the link succeeding:

```
ubuntu:24.04 (glibc 2.39, aarch64):
  trusty-search --version → "trusty-search 0.42.1"   exit=0
  tm --version            → "trusty-mpm 1.3.5"       exit=0

amazonlinux:2023 (glibc 2.34, aarch64):
  trusty-search → error while loading shared libraries: libmvec.so.1  exit=127
  tm            → "trusty-mpm 1.3.5"                                  exit=0
```

**5. The zigbuild path did not regress** — trusty-mpm (no numkong) dry run
(run [30943136387](https://github.com/bobmatnyc/trusty-tools/actions/runs/30943136387)):
all three build legs green, and its arm64 binary keeps the glibc **2.17** floor
with no libssl dependency at all, confirming the native runner change did not
cost the portable path anything.

## Known limitation, stated plainly

The `trusty-search` arm64 gnu asset carries a **glibc 2.39** floor and needs
OpenSSL 3 (`libssl.so.3`, `libcrypto.so.3`) at runtime. Evidence #4 above shows
it running on Ubuntu 24.04 arm64 and **failing to load on AL2023 arm64**.

This is the same floor its `x86_64-unknown-linux-gnu` sibling has always
carried — the project's answer there is the separate `x86_64-linux-al2023`
load-dynamic asset. The symmetric answer for arm64 is an `aarch64-linux-al2023`
leg; that is an additional asset name with install.sh coordination (#2038)
attached, so it is deliberately **not** bundled into this fix. It is the right
follow-up if AL2023/Graviton is the target deployment, and I would recommend
filing it. Non-ORT crates are unaffected: their arm64 asset runs on AL2023
today (evidence #4).

## Unverified

- No **tagged-release** run was exercised (dry runs skip the `release` and
  `homebrew-bump` jobs by design). The `release` job's asset handling is
  unchanged by this PR and its inputs are the same artifacts verified above.
- The audit's behaviour when a **macOS** leg is the one missing is covered by
  local unit tests, not by a live run.
- `publish-dry-run` failed on the trusty-mpm run — `trusty-agents-common ^0.5.0`
  is not on crates.io yet. Pre-existing, unrelated to this change, and that job
  is deliberately not a dependency of `build`/`release`.

## Scope

`.github/workflows/release.yml` only. No version bumps, no `Cargo.toml`
changes, nothing touched in `crates/trusty-search/`. CI-only, so the changelog
gate exempts it — verified locally:
`changelog-fragment gate: scanned 1 changed path(s); no crate source changed
(docs-only / CI-only / test-only) — OK.`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,9 +34,31 @@
 #   versioned glibc stubs), so it changes nothing about which sources get
 #   compiled or which Cargo features are active — the resulting binary lands
 #   in the SAME `target/<triple>/release/` directory a plain `cargo build`
-#   would use, just linked against an older glibc baseline. This same
-#   mechanism cross-compiles `aarch64-unknown-linux-gnu` from the x86_64
-#   runner with no separate arm64 runner or Docker `cross` toolchain needed.
+#   would use, just linked against an older glibc baseline.
+#
+# 2533: LINUX ARM64 IS BUILT NATIVELY, NOT CROSS-COMPILED.
+#   #2037 originally cross-compiled `aarch64-unknown-linux-gnu` from the
+#   x86_64 runner via zig. That works for pure-Rust crates, but breaks for any
+#   crate whose graph reaches a `-sys` crate that asks pkg-config where a
+#   system library lives: `openssl-sys` (reached via fastembed → hf-hub →
+#   native-tls, and via git2/libgit2-sys) aborts the build with
+#
+#       pkg-config has not been configured to support cross-compilation
+#       Could not find directory of OpenSSL installation
+#
+#   because the x86_64 host's pkg-config cannot answer questions about arm64
+#   target libraries, and there is no arm64 sysroot on the runner. That is why
+#   no `aarch64-unknown-linux-gnu` asset ever shipped for trusty-search (last
+#   confirmed: run 30910233376 / job 91994983751, v0.42.0).
+#
+#   Fix: the leg now runs on GitHub's NATIVE `ubuntu-24.04-arm` runner (public
+#   repositories get these at no cost), so HOST == TARGET and every `-sys`
+#   build script resolves against the runner's own arm64 `libssl-dev` exactly
+#   the way the x86_64 leg resolves against its amd64 one. Cross-compilation —
+#   the actual root difficulty — is removed rather than worked around. zig is
+#   still used on this leg, but ONLY to pin the glibc floor (below), never to
+#   cross-compile. Alternatives considered and rejected are recorded in the PR
+#   for #2533.
 #
 #   CAVEAT — ORT-bundling crates (trusty-search, trusty-analyze), discovered
 #   via a real workflow_dispatch dry-run CI validation while building #2037:
@@ -58,10 +80,23 @@
 #   crates (see the matrix-generation step below) — proven working in CI,
 #   since it never links ONNX Runtime at build time either.
 #
-#   The Linux arm64 leg is marked `best_effort` (continue-on-error, like the
-#   existing AL2023 leg) because it is new and has not yet been validated by
-#   an actual tagged release — see the PR for #2037 for the target-triple /
-#   asset-name coordination with the companion install.sh issue (#2038).
+# 2533: THE ARM64 LINUX LEG IS BLOCKING, AND MISSING ASSETS ARE AUDITED.
+#   The arm64 leg used to be `best_effort` (continue-on-error). Because a
+#   `best_effort` leg's failure does not change the run's conclusion, the
+#   workflow reported `success` for four releases in a row while shipping no
+#   arm64 Linux binary at all — a partial release that looked clean. Two
+#   changes close that:
+#     1. The arm64 Linux leg is no longer `best_effort`. It is a supported
+#        target (Graviton, arm64 containers), so its failure now turns the run
+#        red like any other required leg.
+#     2. The `release-completeness-audit` job below compares the assets that
+#        were actually produced against the build matrix that was requested,
+#        and reports every dropped leg in the run summary — as a hard failure
+#        for required legs, and as a loud `::warning` annotation for the one
+#        remaining `best_effort` leg (AL2023). No matrix leg can now go
+#        missing without saying so somewhere a reader will look.
+#   See also #4087 and #4125 — the same "real failure reported as success"
+#   shape in other subsystems.
 #
 # ASSET NAMING CONVENTION (matches INSTALL-CONVENTION.md / PR #891):
 #   <crate>-<version>-<target>.tar.gz           (standard variants, e.g.
@@ -444,12 +479,12 @@ jobs:
       # linux-zigbuild variants (issue #2037 — portable glibc 2.17 baseline,
       # built via `cargo zigbuild --target <triple>.2.17`; see the file-header
       # comment for the full rationale):
-      #   { runner: ubuntu-latest, target: x86_64-unknown-linux-gnu,  variant: linux-zigbuild, best_effort: false }
-      #   { runner: ubuntu-latest, target: aarch64-unknown-linux-gnu, variant: linux-zigbuild, best_effort: true  }
-      #   (aarch64 is cross-compiled from the x86_64 runner by zig/cargo-zigbuild;
-      #    for ORT-bundling crates it uses the load-dynamic feature set, same as
-      #    the AL2023 variant below, since bundled-ort has no aarch64 build-time
-      #    binary to download)
+      #   { runner: ubuntu-latest,    target: x86_64-unknown-linux-gnu,  variant: linux-zigbuild, best_effort: false }
+      #   { runner: ubuntu-24.04-arm, target: aarch64-unknown-linux-gnu, variant: linux-zigbuild, best_effort: false }
+      #   (2533: aarch64 builds NATIVELY on an arm64 runner — it is no longer
+      #    cross-compiled and no longer best-effort; for ORT-bundling crates it
+      #    uses the load-dynamic feature set, same as the AL2023 variant below,
+      #    since bundled-ort has no aarch64 build-time binary to download)
       #
       # AL2023 variant (ORT crates only):
       #   { runner: ubuntu-latest, target: x86_64-unknown-linux-gnu,  variant: al2023   }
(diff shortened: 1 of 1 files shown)


bobmatnyc:
## trusty-review verdict: APPROVE*

**Grade: C+** · reviewer model `us.anthropic.claude-sonnet-4-6`

Solid, well-documented CI fix that closes a real silent-failure gap (arm64 leg
now blocking + `release-completeness-audit` catches missing/mislabelled
assets). Reviewer context items were independently verified against this
branch's diff before merge:

- The arm64 leg is blocking (`best_effort: False`) and `release-completeness-audit`
  turns a missing required asset into a red run — proven live: run
  [30941401868](https://github.com/bobmatnyc/trusty-tools/actions/runs/30941401868)
  concluded `failure` (audit job itself failed, correctly flagging the missing
  aarch64 asset) vs. [30942504925](https://github.com/bobmatnyc/trusty-tools/actions/runs/30942504925)
  concluded `success` (audit job passed).
- The ELF/Mach-O machine-type assertion (`.github/workflows/release.yml:1155-1163`)
  runs before packaging and fails the job on a mismatch.
- numkong-using crates build arm64 via the runner's native GCC (`native_linux_entry`),
  not zig — confirmed no `NK_TARGET_NEON*`/`NK_TARGET_SVE*` overrides are applied
  on the arm64 leg (the only `NK_TARGET_*=0` overrides in the file are scoped to
  `matrix.variant == 'al2023'`, unrelated to this leg). SIMD is not silently
  disabled. The non-numkong `linux-zigbuild` path is unchanged and keeps its
  glibc 2.17 floor per the PR's own verified evidence (run 30943136387).

### Findings (3, all advisory — none blocking)

1. **`actions/download-artifact@v5` reference** (line 1278, confidence 0.72 in
   the automated review) — flagged as possibly nonexistent. Verified moot:
   `actions/download-artifact` v5.0.0 is a real published release, and the
   audit job using it actually ran and succeeded in both referenced dry runs
   above (job `Release completeness audit (issue #2533)` concluded `success`
   in 30942504925, `failure` for the correct reason in 30941401868).
2. **Bash here-string (`<<<`) in the arch-check step** (line 1155) — advisory
   only; GitHub Actions defaults Linux `run:` steps to `bash`, so this is fine
   as written.
3. **`release` job does not `needs: [..., release-completeness-audit]`** —
   confirmed true in the diff (line ~1570, `needs: [setup, build]`). This is
   consistent with the job's existing #3540 design (release publishes off the
   macOS leg regardless of Linux leg/audit outcome, by explicit prior
   decision) rather than a new regression from this PR, but is worth a
   follow-up if the intent is for the audit to gate publishing rather than
   just report on it.

Verdict: merge. No blocking issues found.

Issue #2037, opened 2026-07-06
fix(release): portable Linux release binaries — older-glibc/musl x86_64, arm64, and publish trusty-analyze/tga


bobmatnyc:
Sub-issue of #1992 (single-URL install broken in clean envs).

`.github/workflows/release.yml:393-465` builds installer/crate binaries on `ubuntu-latest` targeting only `x86_64-unknown-linux-gnu` + `aarch64-apple-darwin`, dynamically linked against a glibc newer than most LTS distros ship — install fails on Debian 12 / Ubuntu 22.04 / RHEL 9 / AL2023 with `GLIBC_2.39' not found` (#1992 finding 1). There is also no `aarch64-unknown-linux-gnu`/musl target, so Linux arm64 has no binary at all (#1992 finding 2). And `trusty-analyze`/`tga` publish no Linux release binary, so they silently fall back to `cargo install` which fails with no Rust toolchain (#1992 finding 5).

**Fix the release workflow to:**
- (a) Build x86_64 Linux against an older glibc baseline OR add `x86_64-unknown-linux-musl` for portability
- (b) Add a Linux aarch64 target (gnu-old-baseline or musl)
- (c) Include `trusty-analyze`/`tga` in the published binary set

**NOTE:** This requires CI iteration to validate the produced binaries actually run on the target distros — flag that verification happens in CI, not locally. Coordinate target-triple asset naming with the install.sh changes (see companion issue #2038).

Issue #1992, opened 2026-07-04
Single-URL install (curl | sh) is broken in clean environments — glibc floor, no arm64, false-negative bootstrap, missing binaries


bob-duetto:
Tested the one-liner `curl -sSf https://raw.githubusercontent.com/bobmatnyc/trusty-tools/main/install.sh | sh` in clean Docker containers (goal: verify it installs the full stack with all deps). It fails in several ways.

### 1. CRITICAL — glibc floor 2.39 makes binaries unrunnable on most LTS distros
On `debian:bookworm-slim` (linux/amd64, glibc 2.36) the tarball downloads + checksum-verifies + extracts, but the binary can't execute:
```
trusty-installer: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found
```
Binaries are built against **GLIBC 2.39** (Ubuntu 24.04-era build host), so they won't run on Debian 12, Ubuntu 22.04 LTS, RHEL 9, Amazon Linux 2023, etc. Only very recent distros (Ubuntu 24.04) work. → Build releases on an older glibc base (or static/musl).

### 2. HIGH — Linux aarch64 not supported at all
On the native arm64 platform the installer hard-fails:
```
error: Linux aarch64 is not yet supported. Build from source: cargo install trusty-installer
```
No ARM-Linux release build → Graviton/ARM servers can't use the one-liner.

### 3. HIGH — bootstrap reports "installed 0/7" but binaries ARE installed (wrong verify dir)
On `ubuntu:24.04` (glibc 2.39, installer runs), `trusty-installer install` prints:
```
info: installed 0/7
error: trusty-search: binary `trusty-search` not found in ~/.cargo/bin or PATH
error: trusty-memory: binary `trusty-memory` not found in ~/.cargo/bin or PATH
...
```
…yet a PATH check right after shows `tm`/`trusty-mpm` 0.14.0, `trusty-search` 0.29.2, `trusty-memory` 0.18.0, `trusty-console` 0.3.1 **present in `~/.local/bin`**. The verify/fallback logic looks in `~/.cargo/bin` (and falls back to `cargo install`, which needs Rust) instead of the actual install dir `~/.local/bin` → false-negative "0/7" + spurious "no Rust toolchain" errors.

### 4. MEDIUM — misleading "Installation complete!"
[...]

PR #4830, opened 2026-08-04
ci(release): add the aarch64-linux-al2023 binary-release leg (follow-up to #2533)

bobmatnyc:
Follow-up to #2533. Predecessor: PR #4822 (merged as `f6f54774`).

#4822 landed the `aarch64-unknown-linux-gnu` leg on a native arm64 runner and
then verified by *running* the result that it carries a **glibc 2.39** floor and
needs OpenSSL 3 — it works on Ubuntu 24.04 arm64 and exits 127
(`libmvec.so.1`) on `amazonlinux:2023` aarch64. That PR named the symmetric fix
(`aarch64-linux-al2023`, mirroring the existing x86_64 asset) and deliberately
deferred it because a new asset name has consumers. The owner has since
confirmed Amazon Linux 2023 on Graviton is a real deployment target, so this
closes that gap.

## The leg

Mirrors the `x86_64-linux-al2023` entry rather than inventing an approach.
Identical container (`amazonlinux:2023`), identical
`--no-default-features --features load-dynamic`, identical `best_effort`
classification. Two things differ:

- **Runner** — `ubuntu-24.04-arm`. The `amazonlinux:2023` tag is a multi-arch
  manifest, so on an arm64 runner it resolves to the arm64 image and
  HOST == TARGET, which is the property #4822 established the gnu leg needed.
- **ORT download** — the step hardcoded `onnxruntime-linux-x64-1.20.1.tgz`,
  correct while x86_64 was the only AL2023 leg. It is now arch-parameterised
  (`linux-aarch64` on arm64); upstream ships both under the same tag.

## NEON / SIMD — the decision, stated

**SIMD is enabled on this leg.** The one exception is named below rather than
left implicit.

#4822 hit `__attribute__((target("neon")))` being rejected by zig's clang on
AArch64, where the only available workaround was `NK_TARGET_*=0` across the
board — shipping a vector search engine with SIMD off. Building inside the
AL2023 container with a real GCC avoids that wholesale trade. Reproduced by
compiling numkong 7.6.0's own probe files with the container's compiler,
`amazonlinux:2023` aarch64, GCC 11.5.0:

```
NK_TARGET_NEON       -march=armv8-a+simd                    PASS
NK_TARGET_NEONHALF   -march=armv8.2-a+simd+fp16             PASS
NK_TARGET_NEONSDOT   -march=armv8.2-a+dotprod               PASS
NK_TARGET_NEONFHM    -march=armv8.2-a+simd+fp16+fp16fml     PASS
NK_TARGET_SVE        -march=armv8.2-a+sve                   PASS
NK_TARGET_SVEHALF    -march=armv8.2-a+sve+fp16              PASS
NK_TARGET_SVESDOT    -march=armv8.2-a+sve+dotprod           PASS
NK_TARGET_SVE2       -march=armv8.2-a+sve2                  PASS
NK_TARGET_NEONBFDOT  -march=armv8.6-a+simd+bf16             fail   (GCC 11 predates)
NK_TARGET_SVE2P1     -march=armv8.2-a+sve2p1                fail   (GCC 11 predates)
NK_TARGET_SME        -march=armv8-a+sme                     fail   (GCC 11 predates)
```

The `fail` rows compile out via numkong's designed probe mechanism — the same
thing the x86_64 AL2023 leg has always done for the AMX/AVX-512 generations GCC
11 predates.

**Exactly one extension is forced off by hand: `NK_TARGET_NEONSDOT=0`.** Found
by running the build, not by reasoning about it — the first dispatch failed
(run [30948527782](https://github.com/bobmatnyc/trusty-tools/actions/runs/30948527782),
job 92124822938):

```
include/numkong/maxsim/neonsdot.h: In function 'nk_maxsim_coarse_argmax_neonsdot_':
arm_neon.h:32320:1: error: inlining failed in call to 'always_inline' 'vdotq_s32':
                          target specific option mismatch
```

The probe passes (see table), but `include/numkong/maxsim/neonsdot.h` is the
single ARM header gating its body with the bare `#pragma GCC target("+dotprod")`
form instead of the `arch=armv8.2-a+dotprod` form its five sibling neonsdot
headers use. GCC 11 accepts that pragma without enabling the feature; GCC 13 on
the Ubuntu runners handles it, which is why only this leg needs the override.
This is the AArch64 cousin of the probe/source desync already documented for
AL2023's x86_64 GCC 11 in this workflow.

**What that costs, measured on the shipped binary** (`objdump -d`, run inside
`amazonlinux:2023` aarch64):

```
NEON .4s-lane instructions   : 12253
NEON .16b-lane instructions  : 18432
NEON fmla                    :  2213
sdot, NEON v-register form   :     0   <- the forced-off path
sdot, SVE z-register form    :     4   <- still present
SVE z-register instructions  :   668
```

So: the NEON int8 dot-product kernels are gone; f32/f16 NEON, SVE, SVE2 and the
SVE dot-product path are all present. f32/f16 is what trusty-search indexes with
by default. The alternative — raising the leg's baseline to
`-march=armv8.2-a+dotprod` — was rejected: it trades a narrow kernel loss for a
SIGILL on any pre-Armv8.2 host.

## Verification

`workflow_dispatch` on this branch, then the artifact downloaded and executed.
Run [30949042638](https://github.com/bobmatnyc/trusty-tools/actions/runs/30949042638)
— conclusion `success`, all five build legs green.

**1. The artifact is genuinely aarch64**

```
file: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (GNU/Linux),
      dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, stripped
e_machine bytes at 0x12: b7 00   (EM_AARCH64)
sha256 recomputed locally == the shipped .sha256   -> OK
```

**2. It RUNS on AL2023 aarch64 — the entire point of this leg**

`docker run --platform linux/arm64 amazonlinux:2023`, real container on an
arm64 host, not emulation and not inferred from a green build:

```
container arch: aarch64
ldd (GNU libc) 2.34

trusty-search --version    -> "trusty-search 0.42.1"      exit=0
trusty-embedderd --version -> "trusty-embedderd 0.3.10"   exit=0
```

**3. Control — the same container, the asset #4822 shipped**

```
aarch64-unknown-linux-gnu asset:
  error while loading shared libraries: libmvec.so.1: cannot open shared
  object file: No such file or directory              exit=127
```

Same container, same command, same version. The new asset is the difference.

**4. glibc floor and runtime dependencies** (`readelf -V`, max referenced
symbol version):

| asset | glibc floor |
| --- | --- |
| `aarch64-linux-al2023` (new) — both binaries | **2.34** |
| `aarch64-unknown-linux-gnu` (#4822) | 2.39 |

Runtime libraries, all resolved from AL2023's own `/lib64` with nothing
installed beyond the base image:

```
trusty-search:    libstdc++.so.6  libssl.so.3  libcrypto.so.3  libgcc_s.so.1
                  libm.so.6  libc.so.6  libz.so.1
trusty-embedderd: libssl.so.3  libcrypto.so.3  libgcc_s.so.1  libm.so.6
                  libc.so.6  libz.so.1
```

It still needs OpenSSL 3 — AL2023 ships it, which is why it resolves here and
did not on the gnu asset at its higher glibc floor. As with every load-dynamic
asset, `ORT_DYLIB_PATH` is required before the embedding daemon will start;
`ORT-RUNTIME-NOTE.txt` is packaged alongside and now names the `linux-aarch64`
ORT tarball on arm64.

**5. The audit registers the new leg** — no code change was needed, because it
derives expected asset names from the matrix. Five rows where #4822's run had
four:

```
| OK | trusty-search-0.42.1-aarch64-apple-darwin.tar.gz      | required    |
| OK | trusty-search-0.42.1-x86_64-unknown-linux-gnu.tar.gz  | required    |
| OK | trusty-search-0.42.1-aarch64-unknown-linux-gnu.tar.gz | required    |
| OK | trusty-search-0.42.1-x86_64-linux-al2023.tar.gz       | best-effort |
| OK | trusty-search-0.42.1-aarch64-linux-al2023.tar.gz      | best-effort |
```

## Consumer coordination

**`crates/trusty-installer/src/download/glibc.rs` — the consumer that actually
mattered.** `select_asset_suffix` handed *every* arm64 Linux host the
`aarch64-unknown-linux-gnu` asset regardless of glibc. That is the same bug
#1992 fixed on x86_64, now visible on arm64: a below-floor host silently
downloads a binary that cannot load (evidence #3). It now routes a below-floor
arm64 host to `aarch64-linux-al2023`, mirroring the x86_64 rule exactly. Three
new tests plus a constant-drift guard that pins both AL2023 suffixes to the
strings `release.yml` publishes.

```
cargo test -p trusty-installer --lib glibc::
  test result: ok. 15 passed; 0 failed; 0 ignored
    select_arm64_prefers_al2023_below_floor ... ok
    select_arm64_unknown_glibc_keeps_native ... ok
    al2023_suffixes_match_release_workflow ... ok
```

`cargo clippy -p trusty-installer --all-targets -- -D warnings` clean;
`cargo fmt --check` exit 0.

**`install.sh` — no behaviour change, and #2038 does not apply here.** #2038 is
CLOSED; both its asks (accept Linux aarch64, real partial-failure exit codes)
already shipped, and its coordination clause — "coordinate the arm64 target
triple with #2037" — is satisfied: `LINUX_ARM64_TARGET` already equals what CI
publishes. The new AL2023 asset is deliberately *not* selected here, because
this script installs `trusty-installer`, which is not an ORT crate: its arm64
asset is the cargo-zigbuild glibc-2.17 build and already runs on AL2023 arm64.
The glibc-aware routing that does matter lives inside trusty-installer, and that
is where the new suffix was taught. The comment at the constant said "#2037 has
not yet landed"; it now records the resolved state and this reasoning, so the
next reader does not re-derive it.

**`homebrew-bump` — no change, checked rather than assumed.** It hardcodes
`MACOS_SUFFIX="aarch64-apple-darwin"` and
`LINUX_SUFFIX="x86_64-unknown-linux-gnu"` and never references an AL2023 asset.

**Docs** — the normative platform lists: `INSTALL-CONVENTION.md` (Tier 2 now
names both architectures; its Tier 1.5 section still described the arm64 gnu leg
as cross-compiled and `continue-on-error`, which #4822 made untrue, so that is
corrected with the measured 2.39-floor caveat), `trusty-search` and
`trusty-embedderd` READMEs, `common-pitfalls.md`, and `metacoding-vision.md`'s
AL2023-coverage gate.

**Changelog** — `crates/trusty-installer/changelog.d/2533-aarch64-al2023-asset.md`.
The workflow and docs changes are exempt, but the installer source change is
user-facing install behaviour, so the gate applies and passes:
`OK trusty-installer: changelog.d fragment present and valid`.

## Also fixed here

Target-scoped the AL2023 cargo cache key. Adding a second AL2023 leg made
`al2023-1.91-<crate>-<lockhash>` ambiguous between two legs differing only in
architecture — whichever saved first would hand the other its `target/`,
including host-compiled build-script binaries of the wrong architecture. Not
observed in a run (the first arm64 build reported `Cache not found`), which is
why it is worth closing before it lands. The unqualified `al2023-1.91-`
restore-key would have reintroduced the same cross-arch restore through the
fallback path, so the fallbacks are target-scoped too.

## Unverified / out of scope

- **No tagged-release run.** Dispatch dry runs skip `release` and
  `homebrew-bump` by design. Neither job's asset handling is changed by this PR;
  `homebrew-bump` provably does not read AL2023 assets at all.
- **The installer's live download path was not exercised end-to-end on an arm64
  AL2023 host.** `select_asset_suffix` is pure and unit-tested, and the asset it
  now names was executed there (evidence #2), but no run of `trusty-installer`
  itself was performed on such a host — the two halves are verified separately,
  not as one flow.
- **`trusty-embedderd`'s runtime remediation strings** (`src/glibc_probe.rs`,
  `src/readiness.rs`) still say "use the prebuilt `x86_64-linux-al2023` tarball"
  with no arm64 counterpart. Left alone deliberately: another crate's source,
  outside this change's scope, and worth its own ticket rather than a drive-by
  edit.
- **No version bumps.** trusty-search 0.42.1 shipped today; nothing was bumped,
  no tag or release was created, and no CI gate demanded a bump —
  `publish-dry-run` passed on this branch.
- **Not merged.**


bobmatnyc:
## trusty-review verdict: APPROVE (Grade A)

Reviewed via `mcp__trusty-review__review_pr` at head `1047ce84d5d7720bd2a68729c4e4f387fd80ff2b`.

**Findings:** none (0 findings).

**Model summary:** "Well-structured follow-up that correctly extends the AL2023 build matrix to arm64, fixes the cache key cross-arch collision, handles the ORT architecture selection, routes the installer to the right asset, and adds targeted tests — no correctness issues found."

Independently spot-checked against the diff before merge:
- Real container-run evidence for AL2023 aarch64 (exit 0 `trusty-search --version`) vs. the control failure on the #4822 asset (exit 127, `libmvec.so.1`).
- NEON/SIMD decision is explicit and measured (objdump evidence), not silent — `NK_TARGET_NEONSDOT=0` is the one stated exception.
- New asset is wired into `release.yml`'s matrix (`aarch64-linux-al2023` leg + target-scoped cache key) and into `release-completeness-audit`, which derives expected asset names from that matrix. `trusty-installer/src/download/glibc.rs` routes below-floor arm64 hosts to the new asset with dedicated tests.

Proceeding to merge per owner pre-authorization (green + this verdict).

Issue #4821, opened 2026-08-04
chore(trusty-agents): no persistent-service deploy for the Slack gateway (tagent --slack)


bobmatnyc:
No persistent-service definition exists for the trusty-agents Slack gateway (`crates/trusty-agents/src/slack/`). The reply path — `handlers.rs` → `ctrl::run_pm_task_with_persona` — is live in code, but nothing runs `tagent --slack` as a continuously-running service: no launchd plist, no pm2 config, no deploy script.

Consequence: the cto-assistant cutover tracked in #3852 has no host to cut over to.

Legacy comparison point: the bot being replaced runs pm2-managed with a launchd cron fallback at `/Users/masa/Duetto/cto-resources/cto`.

Two constraints shape the work:

- Slack Socket Mode fans events to every open connection on an app token, so the new gateway and the legacy bot cannot both listen on app A0AMPRM4W0J without double-reply risk. Deployment must be sequenced with the token handover.
- The host must have a working credential for cto-assistant's bare `claude-sonnet-4-6` model. Provider resolution is ambient (#3765) — verify what `pick_credentials()` actually resolves ON THE HOST, not what the TOML claims, because #4819 means `run_pm_task_with_persona` has no recovery: a raw auth or transport error becomes the assistant's Slack reply to real users.

Unblocks #3852.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

PR #4831, opened 2026-08-04
fix(trusty-agents): install the rustls CryptoProvider and let TAGENT_PROJECT_DIR beat exe inference

### LINKED ISSUE #4825: bug(trusty-agents): tagent --slack panics on the first WSS handshake — no process-level rustls CryptoProvider is installed
`tagent --slack` logs "Slack Socket Mode connected" and then panics inside rustls 0.23 (`crypto/mod.rs`, `CryptoProvider::get_default_or_install_from_crate_features`):

```
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point ... or make sure exactly one of
the 'aws-lc-rs' and 'ring' features is enabled.
```

Cause: nothing in this workspace ever calls `CryptoProvider::install_default()`, and the single `rustls 0.23.40` instance in the graph has BOTH provider features enabled — `ring` (via reqwest's `rustls-tls` → `__rustls-ring`, `tokio-rustls`, `ureq`) and `aws-lc-rs` (via the `aws-smithy-http-client` stack behind `aws-sdk-bedrockruntime`). rustls therefore cannot auto-select a provider and panics on first use.

The trigger is the `tokio-tungstenite` Socket Mode connector reached from `slack::run_slack_bot`, which builds a `rustls::ClientConfig` from the process default. `reqwest` supplies its own provider explicitly, which is why the `apps.connections.open` call succeeds and the "connected" line prints before the panic.

Consequence: `tagent --slack` cannot run on any host — the Slack gateway is dead on arrival, so the CTO Assistant has no delivery channel. Reproduced live on the owner's machine today against installed v0.38.6; the defect is present on current `main` (no `install_default` call exists anywhere under `crates/`).

Blocks #4821.

### LINKED ISSUE #4826: bug(trusty-agents): explicit TAGENT_PROJECT_DIR loses to current_exe() inference — the Slack gateway resolves $HOME as the project
`detect_self_project_with_hint` (in `ctrl::util`) honours its env hint only when the hint directory itself carries the `.trusty-agents/agents/pm.toml` marker. When it does not, the hint is discarded silently and the function falls through to walking up from `current_exe()`.

Under the launchd job the executable is `~/.cargo/bin/tagent`, so that walk-up reaches `~/.trusty-agents/agents/pm.toml` and returns `/Users/masa` — beating both `TAGENT_PROJECT_DIR` and the job's `WorkingDirectory`, which both point at the CTO project. The gateway logged `project=/Users/masa`.

`dispatch_cli_mode`'s `--slack` branch takes that value as the gateway's project path, and the same helper also resolves the `.env.local` load, the `.trusty-agents/state` directory, discovered Python skills, and the service pid file. So the CTO Assistant answers real Slack users against $HOME instead of the CTO corpus: it replies normally, with the wrong context and no error — worse than failing loudly.

An explicitly-set env hint is an operator declaration and must win over exe-path inference. Marker-based inference should remain the fallback for when no hint is given.

Blocks #4821.

bobmatnyc:
Closes #4825
Closes #4826

Two defects that together made the Slack gateway unusable for the CTO Assistant cutover: it could not start, and had it started it would have answered against the wrong project.

## #4825 — startup panic on the first WSS handshake

A single `rustls 0.23.40` instance in the graph carries BOTH provider features — `ring` (reqwest's `rustls-tls` → `__rustls-ring`, `tokio-rustls`, `ureq`) and `aws-lc-rs` (the `aws-smithy-http-client` stack) — so rustls cannot auto-select and panics the moment `tokio-tungstenite` builds a `ClientConfig` from the process default. Nothing in the workspace called `install_default()`.

**Chosen fix: install the provider explicitly, not deduplicate the graph.** The two providers arrive from independent subtrees (HTTP client vs. AWS SDK), so dropping either means changing reqwest's TLS backend workspace-wide or dropping the Bedrock SDK — a far larger blast radius, and one a future dependency bump could silently undo. An explicit install is one call, immune to that, and auditable.

**Placement:** `runtime::run()`, not `main.rs` — `trusty-agents-local` also calls `run()`, and `main.rs` would have covered only the `tagent` binary. It sits ahead of the `config` / `mcp-serve` early dispatches because those make HTTPS calls too.

**Error handling: not `.ok()`.** `install_default` returns `Err` for exactly one reason — a provider was already installed — which is benign, so that value is discarded. What makes the discard safe is the postcondition check that follows: if the process still has no default provider, `install_crypto_provider` returns a startup error rather than letting the process reach a handshake and abort. A silently swallowed failure would have re-hidden this exact bug.

## #4826 — gateway resolved $HOME as the project

`detect_self_project_with_hint` honoured its env hint only if the hinted directory itself carried `.trusty-agents/agents/pm.toml`; otherwise it dropped the hint silently and walked up from `current_exe()`. Under launchd that walk reached `~/.trusty-agents/agents/pm.toml` and returned `/Users/masa`, beating both `TAGENT_PROJECT_DIR` and `WorkingDirectory`. An explicit hint is now honoured whenever it resolves to a real directory; marker-based inference remains the fallback for when no hint is given. An unresolvable hint still falls through to inference, with a warning.

## Tests

Both were confirmed to fail against the unfixed code before the fix landed.

`tests/slack_wss_crypto_provider.rs` drives the real Socket Mode connector against a local throwaway listener, captures the client's first flight, and asserts it is a well-formed TLS ClientHello — an invariant that only holds if the provider was resolved and used. Against the unfixed tree the same test panicked at `rustls-0.23.40/src/crypto/mod.rs:249:14` with the identical message from the production log, and no bytes reached the listener.

`detect_self_project_hint_beats_exe_inference` asserts the returned path equals the hint, not merely `is_some`. Against the unfixed tree:

```
assertion `left == right` failed: an explicitly-set project-dir hint must win over exe-path inference
  left: Some("/Users/masa")
 right: Some("/private/var/folders/…/cto")
```

## Live verification

**A green build proves nothing here, so the handshake was exercised for real.** A completed TLS + WebSocket handshake through the exact `tokio-tungstenite` connector the gateway uses:

```
LIVE WSS HANDSHAKE OK  url=wss://echo.websocket.org/  upgrade_status=101 Switching Protocols
```

The built binary was then run as `tagent --slack` with deliberately invalid tokens, under `env -i HOME=/Users/masa` so the `~/.trusty-agents/agents/pm.toml` marker that caused #4826 was present:

```
INFO trusty_agents::runtime::startup: trusty-agents v0.38.6 build #1
INFO trusty_agents::slack: Starting Slack bot in Socket Mode project=/private/tmp/…/fakecto
ERROR trusty_agents::slack: Slack socket error; reconnecting in 5s error=apps.connections.open failed: invalid_auth (status 200 OK)
```

No panic; startup completed; `project=` is the directory `TAGENT_PROJECT_DIR` named, not `$HOME` — #4826 reproduced-and-fixed live under the production condition.

**What is NOT verified, and what the owner must do.** The production Slack app A0AMPRM4W0J was deliberately not touched, so no real Socket Mode session was established. The invalid-token run reaches `apps.connections.open` (plain HTTPS via reqwest, which carries its own provider) and stops there — it never opens the `wss-primary.slack.com` socket. The Slack-specific leg therefore rests on the public-endpoint handshake above, which uses identical TLS mechanics and differs only in endpoint and auth.

The owner should confirm on the real deployment: install the binary, `launchctl kickstart -k gui/$UID/com.trusty.agents.slack`, and check `~/Library/Logs/trusty-agents/slack-stderr.log` shows `Slack Socket Mode connected` followed by a held connection with no panic, and a `project=` line naming the CTO directory rather than `/Users/masa`.

## Gates

Rung 3 (localized behavior in one crate), all crate-scoped, exit codes checked explicitly:

| Gate | Exit | Result |
|---|---|---|
| `cargo test -p trusty-agents` | 0 | 3334 passed, 0 failed, 11 ignored (+ 27 across integration binaries) |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | 0 | clean |
| `cargo fmt --check` | 0 | clean |
| `bash scripts/check_line_cap.sh` | 0 | clean |
| `bash scripts/check_changelog_fragment.sh` | 0 | fragment present and valid |
| `bash scripts/check_sld.sh` | 0 | clean |

Blocks lifted for #4821.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11159,6 +11159,7 @@ dependencies = [
  "regex",
  "reqwest 0.12.28",
  "rust-embed",
+ "rustls",
  "scraper",
  "secrecy",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -227,6 +227,12 @@ teloxide = { version = "0.17", default-features = false, features = ["rustls", "
 # Slack Socket-Mode WebSocket client (trusty-mpm `slack` feature, DOC-20 #1294).
 # Same pin/features trusty-agents already uses, so the lockfile is unchanged.
 tokio-tungstenite = { version = "0.24", features = ["rustls-tls-native-roots"] }
+# #4825: direct rustls handle so a binary can install the process-level
+# CryptoProvider explicitly. One rustls 0.23 instance in this graph carries
+# BOTH provider features — `ring` (reqwest's `rustls-tls`, `ureq`) and
+# `aws-lc-rs` (the aws-smithy client stack) — so rustls cannot auto-select one
+# and panics on the first handshake that goes through `ClientConfig::builder()`.
+rustls = { version = "0.23", default-features = false, features = ["aws-lc-rs", "std"] }
 # Progress bars / spinners. Shared rustup/cargo-style progress display lives in
 # the `trusty-progress` crate (#1315). Pinned to 0.17 to match the existing
 # local pins in trusty-search and tga; migrating those two crates to this
--- a/crates/trusty-agents/Cargo.toml
+++ b/crates/trusty-agents/Cargo.toml
@@ -159,6 +159,8 @@ aws-smithy-types = "1"
 git2 = { version = "0.20", features = ["vendored-libgit2"] }
 teloxide = { version = "0.17", features = ["macros"] }
 tokio-tungstenite = { version = "0.24", features = ["rustls-tls-native-roots"] }
+# #4825: needed by `tls::install_crypto_provider` — see the workspace entry.
+rustls = { workspace = true }
 flate2 = "1"
 rand = "0.8"
 similar = { version = "2", features = ["text"] }
@@ -207,6 +209,11 @@ serial_test = "3"
 # tests so credential-resolution tests can assert on the actual value that
 # reaches the request builder, not just "no panic" (see #3431 follow-up).
 secrecy = "0.10"
+# #4825: `tests/slack_wss_crypto_provider.rs` drives the real Socket Mode
+# connector and inspects the installed provider, so both must be reachable
+# from an integration-test binary.
+tokio-tungstenite = { version = "0.24", features = ["rustls-tls-native-roots"] }
+rustls = { workspace = true }
 
 [lints.clippy]
 doc_lazy_continuation = "allow"
--- a/crates/trusty-agents/Cargo.toml
+++ b/crates/trusty-agents/Cargo.toml
@@ -209,11 +209,6 @@ serial_test = "3"
 # tests so credential-resolution tests can assert on the actual value that
 # reaches the request builder, not just "no panic" (see #3431 follow-up).
 secrecy = "0.10"
-# #4825: `tests/slack_wss_crypto_provider.rs` drives the real Socket Mode
-# connector and inspects the installed provider, so both must be reachable
-# from an integration-test binary.
-tokio-tungstenite = { version = "0.24", features = ["rustls-tls-native-roots"] }
-rustls = { workspace = true }
 
 [lints.clippy]
 doc_lazy_continuation = "allow"
--- a/crates/trusty-agents/changelog.d/4825-4826-slack-gateway-startup.md
+++ b/crates/trusty-agents/changelog.d/4825-4826-slack-gateway-startup.md
@@ -1,0 +1,18 @@
+Fixed
+
+- `tagent --slack` no longer panics on its first WebSocket handshake
+  (closes [#4825](https://github.com/bobmatnyc/trusty-tools/issues/4825)).
+  A single `rustls 0.23` instance in the dependency graph carries both the
+  `ring` and `aws-lc-rs` provider features, so rustls could not auto-select one
+  and aborted the process the moment the Socket Mode connector built a
+  `ClientConfig` — right after logging "Slack Socket Mode connected". `run()`
+  now installs the `aws-lc-rs` provider explicitly before any TLS-capable code
+  path, and fails loudly at startup if no provider ends up installed.
+- The Slack gateway resolves the project it was actually pointed at
+  (closes [#4826](https://github.com/bobmatnyc/trusty-tools/issues/4826)).
+  An explicitly-set `TAGENT_PROJECT_DIR` now outranks the `current_exe()`
+  walk-up in self-project detection; previously a hint without a
+  `.trusty-agents/agents/pm.toml` marker was dropped silently and the walk-up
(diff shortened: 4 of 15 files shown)


bobmatnyc:
## Review fixes — both items addressed, no production changes

**HIGH (test isolation) — fixed, and reproduced first rather than taken on faith.**

With `TAGENT_PROJECT_DIR` set to a scratch dir, `plain_flag_stays_resident_and_exits_cleanly_on_quit` passed while writing into it:

```
--- leaked into the hinted dir ---
…/scratchpad/ambient/.trusty-agents/state/build.json
```

After the fix, the same ambient-hint run across all four affected test binaries:

```
test result: ok. 4 passed  (plain_cli_repl)
test result: ok. 3 passed  (config_mount)
test result: ok. 7 passed  (system_status_cli)
test result: ok. 2 passed  (cli_project)
--- files leaked into the hinted dir ---
(count: 0)
```

**A fourth site had the same gap:** `tests/support/project.rs` — the shared fixture behind `api_e2e.rs` and `cli_project.rs` — pins `HOME` at two spawn sites with the same omission. Fixed alongside the three named.

Where a file had several spawn sites (`config_mount.rs` ×5, `system_status_cli.rs` ×2) the clearing is structural: a local `tagent()` constructor every spawn routes through, so a spawn added later cannot forget it. The two `plain_cli_repl.rs` helpers and the two `project.rs` sites take the calls inline, next to the existing `HOME` pin.

**LOW (duplicate dependency) — fixed, and it was one entry larger than reported.** `rustls` was also declared in both sections, by me in this PR. Both dev entries are redundant: integration tests link the package's normal `[dependencies]`. Verified rather than assumed — with both entries deleted, `cargo test -p trusty-agents --test slack_wss_crypto_provider --no-run` compiles clean (exit 0). `[dev-dependencies]` is now byte-identical to `origin/main`.

## Gates — re-run, exit codes checked individually

| Gate | Exit | Result |
|---|---|---|
| `cargo test -p trusty-agents` | 0 | 3334 passed, 0 failed, 11 ignored (+ 27 across integration binaries) |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | 0 | clean |
| `cargo fmt --check` | 0 | clean |
| `bash scripts/check_line_cap.sh` | 0 | clean |
| `bash scripts/check_changelog_fragment.sh` | 0 | fragment present and valid |
| `bash scripts/check_sld.sh` | 0 | clean |

Diff is test-only — `git diff --stat` touches four test files and `Cargo.toml`; zero production files. Not merging; handing back for the merge agent.

Issue #4113, opened 2026-07-27
bug(trusty-search): clean exit leaves the daemon down permanently — launchd KeepAlive never restarts it


bobmatnyc:
## Summary

The repo owner asked today for "a stable version running at this point." Today's outage shows we do not currently have that guarantee for `trusty-search`, and the reason is a supervision gap, not a bug in the daemon itself.

`trusty-search` runs under launchd with `KeepAlive = { SuccessfulExit: false }`. launchd restarts the daemon **only after a non-zero exit**. A **clean** exit — a plain SIGTERM, an orderly shutdown, an operator or tool stopping it — leaves it **down indefinitely**, with no automatic recovery and no alarm.

## Verified evidence (collected 2026-07-27)

**The loaded search unit is `com.trusty.search`** (plist `~/Library/LaunchAgents/com.trusty.search.plist`, `ProgramArguments = trusty-search start --foreground --no-auto-discover --port 7878`). It is the one present in `launchctl list` (pid 74802).

```
$ /usr/libexec/PlistBuddy -c "Print :KeepAlive" ~/Library/LaunchAgents/com.trusty.search.plist
Dict {
    SuccessfulExit = false
}
```

Note for anyone grepping later: a second plist `~/Library/LaunchAgents/com.trusty.trusty-search.plist` also exists, with the **same** `KeepAlive = { SuccessfulExit = false }` and no `--port` argument, but it is **not loaded** (`launchctl list` shows no `com.trusty.trusty-search`). Having two near-identically named search plists is itself a hazard — an operator editing the wrong one changes nothing. The `SuccessfulExit: false` finding holds for both, so the conclusion is unaffected either way.

**Observed outage:** `trusty-search` was fully down from **13:27:22Z** on 2026-07-27 following a clean SIGTERM, and stayed down until an operator manually ran `launchctl kickstart`. Nothing restarted it in the interim, and nothing alarmed.

**Consequence chain:** every consumer that depends on search silently loses capability. Reviews lose search-backed context; `search_all` returns fewer results; nothing surfaces the outage to the user. The failure is invisible — degraded answers, not errors.

## Related but distinct finding: trusty-review is not a daemon at all

Recorded here so the two do not get conflated in triage — it is a different lifecycle with a different failure mode.

`~/Library/LaunchAgents/com.trusty.trusty-review.plist` **exists but is NOT loaded**:

```
$ launchctl print gui/502/com.trusty.trusty-review
[...]

Issue #314, opened 2026-05-27
feat(trusty-search): --no-auto-discover flag to suppress ~/Projects auto-scan on test/cert daemons


bobmatnyc:
## Background

During the v0.14.0 Stage-1 cert (#281), launching a fresh isolated daemon with `--data-dir /tmp/...` (intended to start with zero indexes) immediately triggered `auto_discover_and_index()` which scanned `~/Projects` and found 53 projects, all auto-registered. This contaminated the "0 indexes" prerequisite for the cert.

The cert workaround was to override `HOME=/tmp/ts-cert-home` so `default_scan_paths()` found no subdirs and `GlobalConfig::load()` returned an empty default. This works but is awkward and surprising.

## Proposal

Add a `--no-auto-discover` (or `--no-discover`) flag to `trusty-search start` that suppresses the auto-discovery scan entirely. Matching env var `TRUSTY_NO_AUTO_DISCOVER=1`.

When set:
- `auto_discover_and_index()` is a no-op
- The daemon starts with only explicitly-registered indexes (i.e., whatever's in the `--data-dir`'s `indexes.toml`)

## Use cases

- Cert/benchmark daemons that need clean isolation
- CI test environments that don't have/need access to `~/Projects`
- Headless servers where auto-discovery is irrelevant
- Multi-tenant or sandboxed deployments

## Acceptance criteria

- [ ] `--no-auto-discover` flag on `start` subcommand
- [ ] `TRUSTY_NO_AUTO_DISCOVER` env var
- [ ] Unit test confirming `auto_discover_and_index()` skipped when set
- [ ] README CLI section update
- [ ] CHANGELOG entry

## References

- v0.14.0 cert workaround documented in: `docs/trusty-search/regression-testing/v0.14.0-stage1-cert-2026-05-27.md`
- Related: #281 (Stage-1 cert)
- Related: #311 (PR that landed `--data-dir`)

Issue #86, opened 2026-05-25
feat(setup): idempotent setup/install commands with service-owned hooks and sub-agent guard


bobmatnyc:
## Summary
Refactor setup/install commands so each service owns its hooks, all commands are idempotent,
and MPM sub-agents are guarded from running hooks meant for top-level sessions.

### Hook ownership
| Command | Installs hooks for |
|---|---|
| `trusty-memory setup` | UserPromptSubmit → `trusty-memory prompt-context` |
| `trusty-search setup` | MCP server registration only (no hooks) |
| `tm install` | MPM lifecycle: PreToolUse/PostToolUse/Stop → `trusty-mpm hook` |

### Sub-agent guard
MPM sets `CLAUDE_MPM_SUB_AGENT=1` in every spawned sub-agent environment.
All hook commands exit 0 immediately when this var is set.

### Idempotency
All commands safe to run multiple times. Uses `merge_hook_entries` (already in trusty-common)
and `patch_mcp_server` for dedup.

## New CLI surface
- `trusty-memory prompt-context` — UserPromptSubmit hook command (queries daemon, injects context)
- `trusty-search setup` — new command (currently only has Cursor-targeted `integrate`)
- `tm install` — wired to write MPM lifecycle hooks (currently writes agents only)

## Related
- claude-mpm tracking issue: https://github.com/bobmatnyc/claude-mpm/issues/555

## Acceptance criteria
- [ ] `tm install` run twice produces no duplicate entries
- [ ] `trusty-memory setup` run twice produces no duplicate entries
- [ ] `trusty-search setup` run twice produces no duplicate entries
- [ ] Hook commands exit 0 immediately when `CLAUDE_MPM_SUB_AGENT=1`
- [ ] MPM-spawned sub-agents have `CLAUDE_MPM_SUB_AGENT=1` in their environment
- [ ] `tm install` does NOT install trusty-memory or trusty-search hooks

PR #4835, opened 2026-08-04
fix(trusty-search): preserve --no-auto-discover across service install (#4823)

### LINKED ISSUE #4823: trusty-search service install cannot preserve --no-auto-discover, so regenerating the launchd unit silently re-enables auto-discovery
## What

`trusty-search service install` generates the launchd plist from a fixed template with no way to carry operator-chosen runtime flags. Specifically there is no option to set `--no-auto-discover` (or `TRUSTY_NO_AUTO_DISCOVER=1`), and auto-discovery is default-on. Any operator who deliberately disabled auto-discovery loses that setting the moment the unit is regenerated — which is exactly what `service install` is for.

## How it surfaced

During the trusty-search 0.42.1 install on 2026-08-04, a legacy hand-made unit (`com.trusty.search`) was replaced with the repo-generated one (`com.trusty.trusty-search`) in order to pick up the #4709 KeepAlive fix. The legacy unit ran `start --foreground --no-auto-discover --port 7878` and additionally set `TRUSTY_NO_AUTO_DISCOVER=1` in its environment — belt-and-suspenders, clearly a deliberate operator choice. The regenerated unit preserved the port (DEFAULT_PORT is 7878, so omitting `--port` is behaviourally identical) but dropped the auto-discover suppression entirely, because `service install` has no way to express it.

## Verification

Reading `crates/trusty-search/src/commands/service.rs` line 121-139, the `build_launchd_config()` function takes only `exe` and `log_dir`, and constructs the args list as:

```rust
args: vec!["start".to_string(), "--foreground".to_string()],
```

Reading `crates/trusty-search/src/service/daemon.rs`, the `PERSISTED_ENV_VARS` list does NOT include `TRUSTY_NO_AUTO_DISCOVER`:

```rust
pub const PERSISTED_ENV_VARS: &[&str] = &[
    "TRUSTY_MEMORY_LIMIT_MB",
    "TRUSTY_MAX_CHUNKS",
    "TRUSTY_EMBEDDING_CACHE",
    "TRUSTY_MAX_BATCH_SIZE",
    "TRUSTY_BM25_CORPUS_CAP",
    "TRUSTY_DEVICE",
    "TRUSTY_SEARCH_FANOUT_CONCURRENCY",
];
```

The only way `launchd_env_vars()` in `service.rs` (lines 76-95) could preserve auto-discover settings is if `TRUSTY_NO_AUTO_DISCOVER` appeared in `PERSISTED_ENV_VARS`, but it is absent. There is no flag or config mechanism in `service install` itself.

## Impact

A silent capability change on regeneration: the daemon can auto-scan and register indexes the operator deliberately excluded. The reporter's specific concern was an unwanted scan of `~/Projects`. The workaround is hand-editing the generated plist's `EnvironmentVariables`, which is itself lost on the next `service install` — so the setting cannot currently be made durable by any supported means.

## Ask for

Give `service install` a way to persist operator runtime options — add `TRUSTY_NO_AUTO_DISCOVER` to `PERSISTED_ENV_VARS`, add a `--no-auto-discover` flag to the `service install` command, or implement preservation of an existing unit's `EnvironmentVariables` on regeneration. State the general principle in the issue: regenerating a service unit should not silently discard deliberate operator configuration.

## Related context

#4709 (the KeepAlive fix that motivated the regeneration; fixed the generator but noted the loaded unit on that machine was a hand-made legacy plist no repo code generates).

bobmatnyc:
Closes #4823

`trusty-search service install` regenerated the launchd unit from a fixed template, so an operator who deliberately disabled the auto-discovery scan lost that setting the moment the unit was reinstalled. Hand-editing the generated plist worked until the next `service install` undid it, so the setting could not be made durable by any supported means.

## The `=1` trap, and how it is handled

The remedy the issue body proposed first — adding `TRUSTY_NO_AUTO_DISCOVER` to `PERSISTED_ENV_VARS` — would have crashed the daemon, as the issue comment records. The arg at `crates/trusty-search/src/main.rs:489` was:

```rust
#[arg(long, env = "TRUSTY_NO_AUTO_DISCOVER")]
no_auto_discover: bool,
```

Under clap that makes the CLI flag a bare `SetTrue` but routes the **env var** through strict `FromStr<bool>`, so only the literal `true`/`false` are accepted. Copying the `TRUSTY_NO_AUTO_DISCOVER=1` that already sits in `~/Library/Application Support/trusty-search/daemon.env` into the generated plist produces:

```
error: invalid value '1' for '--no-auto-discover': [possible values: true, false]
```

and launchd then throttle-loops a daemon that never boots.

This PR handles it **twice over** — belt and suspenders, because the two mitigations protect against different things:

1. **Fixed at the source.** `--no-auto-discover` / `TRUSTY_NO_AUTO_DISCOVER` now takes a `value_parser` accepting `1`/`true`/`yes`/`on` and `0`/`false`/`no`/`off`, case-insensitive. This was the right call rather than merely routing around the strict parser: `=1` is the spelling the README, the #314 changelog entry, the flag's own help text, and the pre-existing `no_auto_discover_resolution` unit test all document. The code disagreed with every piece of documentation about it, and the code was the thing that was wrong.

   On the risk of loosening a boolean flag: the parser stays **fail-closed on garbage**. `TRUSTY_NO_AUTO_DISCOVER=ture` is still an error, not a silent `false`. Swallowing a typo would quietly re-enable a scan the operator disabled — the same silent-capability-change defect class as #4823 itself — so the loosening is confined to spellings that unambiguously mean true or false. The bare CLI flag keeps working exactly as before (`num_args = 0..=1` + `require_equals` + `default_missing_value`), which matters because `commands::start::daemon` forwards it bare to the detached child.

2. **Never emitted anyway.** The generated plist never writes `TRUSTY_NO_AUTO_DISCOVER` into `EnvironmentVariables` at all — the key is explicitly excluded from the carry-forward. The suppression travels as a `ProgramArguments` flag instead. A value the daemon's parser could reject therefore cannot reach a generated unit even if an older binary reads a newer plist, and a human reading the plist can see the setting without knowing the env-var contract.

## Design choice

**A `--no-auto-discover` passthrough flag on `service install`, written into `ProgramArguments`** — not a config key, and not the env path.

- *Why not the env path:* it is the one representation that can be parsed and rejected at boot. `ProgramArguments` has no such failure mode.
- *Why not a config key:* `service install` already takes its inputs as flags, a config key would need its own precedence story against the flag the daemon already has, and the operator intent here is about the generated unit, not about daemon runtime config.

**On regeneration (requirement 3): preserve, announce, and require an explicit flag to undo.** Running `service install` over a unit that carries the suppression preserves it and prints so. Turning the scan back on requires `service install --auto-discover`, which also announces what it dropped. Requiring an explicit flag on *every* install was rejected as the wrong trade — it makes the common case hostile without making the dangerous case safer; the defect is the silent discard, not the absence of a prompt. Both `ProgramArguments` and `EnvironmentVariables` representations are recognised when reading the existing unit, so the hand-edit currently live on the reporter's machine (`TRUSTY_NO_AUTO_DISCOVER=1` in the plist environment) is detected and carried forward.

**Generalised, but bounded (requirement 4).** Since reading the installed unit was needed anyway, the same read now carries forward the other `PERSISTED_ENV_VARS` values. This was the same defect, less visibly: `service install` is normally run from a plain shell exporting none of the `TRUSTY_*` tunables, so the old "read the current process env" behaviour blanked every tunable the unit already carried — including the `TRUSTY_DEVICE=cpu` pin that keeps Apple Silicon off CoreML. An exported value still wins over the installed one. The generator itself is untouched: no new plist keys, no restructuring, one extra `args` element and a different `env_vars` source.

## Does the existing `daemon.env` content now work?

**Partly, and the distinction matters.** With this PR, `TRUSTY_NO_AUTO_DISCOVER=1` is *accepted* wherever it reaches clap as a real process env var — an exported shell var, or a plist `EnvironmentVariables` entry — instead of aborting startup. That removes the crash the issue comment documented.

It does **not** make the `daemon.env` line take effect, because `load_daemon_env()` runs at `crates/trusty-search/src/commands/start/daemon.rs:214`, long after `Cli::parse()` has already resolved the flag. So env-backed clap flags written to `daemon.env` are still silently ignored. That ordering bug is a **sibling issue being filed separately** from the same investigation and is deliberately out of scope here; the second sibling (the generated unit omitting `RUST_LOG=info`) is likewise untouched. What this PR does guarantee is that the file's existing content is no longer a landmine — it can no longer crash anything — and that the operator's intent is now expressible durably through `service install --no-auto-discover` regardless.

## What this does to the unit currently installed on the reporter's machine

The live `~/Library/LaunchAgents/com.trusty.trusty-search.plist` carries the hand-edit in both forms — `--no-auto-discover` in `ProgramArguments` *and* `TRUSTY_NO_AUTO_DISCOVER=true` in `EnvironmentVariables` — with the two fd-limit `<dict>`s sitting between them. That is structurally identical to the `sample_plist()` fixture the parser tests run against (the nested-dict case is deliberately covered, since `EnvironmentVariables` has to be found *past* them).

After this lands, `trusty-search service install` on that machine will:

- detect the suppression, print `Preserved auto-discovery suppression from the installed unit`, and re-emit `--no-auto-discover` in `ProgramArguments`;
- **drop the `TRUSTY_NO_AUTO_DISCOVER=true` entry from `EnvironmentVariables`.**

That second point is deliberate but worth flagging, since the belt-and-suspenders entry was added on purpose: the CLI flag takes precedence over the env var and is what the daemon actually acts on, so behaviour is unchanged — but the regenerated file will no longer show both. Keeping the suppression in exactly one place is what makes it impossible for a generated unit to carry an env value the parser could reject.

Nothing here was run against the live daemon: `service install` was never invoked, no `launchctl` command was issued, and the plist's mtime is unchanged. (The daemon's PID did change mid-session — that was launchd's own `KeepAlive` restart, the #4113 policy working as designed, not this branch.)

## Tests

Assertions are on the **rendered plist**, not just the config struct — the #4709 failure mode where the struct is right and the renderer drops the fragment.

**The env path is covered end-to-end.** `tests/no_auto_discover_env.rs` spawns the compiled binary so clap resolves `TRUSTY_NO_AUTO_DISCOVER` from a real process environment — the one path no in-process test can reach, and precisely where the trap lives. (The in-process `cli_tests` drive `Cli::try_parse_from` over argv only, because mutating the process env races a parallel test binary; a child process has its own environment, so there is no shared-state hazard.) Worth noting: the pre-existing `no_auto_discover_resolution` test *looked* like it covered `=1` but asserts against a locally-defined `should_skip_discovery` helper that clap never calls — which is how the defect survived.

The probe points `--data-dir` at a regular file. `handle_start` rejects that in its **first statement** (`crates/trusty-search/src/commands/start/daemon.rs:60-84`), so argument parsing completes and the daemon then aborts before any lockfile, port bind, or spawn. `TRUSTY_DATA_DIR` is removed from the child env so it cannot redirect the probe at the real data directory. Verified empirically: the live daemon's PID was unchanged across every run. Exit 2 means clap rejected the value; anything else means it accepted it.

Against pre-fix code it reproduces the wild failure exactly:

```
$ cargo test -p trusty-search --test no_auto_discover_env    # with the value_parser reverted
test env_var_rejects_an_unrecognised_spelling ... ok
test env_var_accepts_the_documented_numeric_spelling ... FAILED
test env_var_accepts_every_documented_spelling ... FAILED

---- env_var_accepts_the_documented_numeric_spelling stdout ----
assertion `left != right` failed: TRUSTY_NO_AUTO_DISCOVER=1 must not be an argument-parsing
error (issue #4823) — a launchd unit carrying it would throttle-loop a daemon that never
boots; got exit 2 and output:
error: invalid value '1' for '--no-auto-discover'
  [possible values: true, false]

test result: FAILED. 1 passed; 2 failed
```

Three more tests fail against the pre-fix code. Demonstrated by reverting both change sites and re-running:

```
test commands::service::tests::launchd_env_pairs_never_carries_no_auto_discover ... FAILED
test commands::service::tests::build_launchd_config_plist_carries_no_auto_discover_arg ... FAILED
test no_auto_discover_tests::value_form_accepts_documented_spellings ... FAILED

---- build_launchd_config_plist_carries_no_auto_discover_arg stdout ----
assertion `left == right` failed: the suppression must travel as a CLI arg (issue #4823)

---- value_form_accepts_documented_spellings stdout ----
assertion `left == right` failed: --no-auto-discover=1 must parse as true (issue #4823)

test result: FAILED. 346 passed; 3 failed
```

Coverage added (29 tests):

- `no_auto_discover_env::*` — the end-to-end env path described above: the documented numeric spelling, every accepted spelling plus the unset baseline, and a garbage value still failing loudly with a diagnostic that names the argument.
- `build_launchd_config_plist_carries_no_auto_discover_arg` — the flag reaches `ProgramArguments` in the rendered XML, **and `KeepAlive: <true/>` still survives** alongside it (#4709 regression guard).
- `build_launchd_config_omits_no_auto_discover_by_default` — a default install is unchanged.
- `launchd_env_pairs_never_carries_no_auto_discover` — with `=1` supplied by *both* the env lookup and the installed unit, neither the pairs nor the rendered XML contain the key, while the suppression still survives as the CLI arg. This is the "a value the binary would reject can never be emitted" requirement.
- `launchd_env_pairs_carries_forward_installed_tunables` — the generalised preservation, plus the #86 `HF_HOME` pin still emitted.
- `parse_truthy_bool_*`, `cli_tests::*` — the truthiness rule and its clap wiring, including bare-flag compatibility and rejection of garbage.
- `parse_installed_unit_*`, `suppresses_auto_discover_*`, `resolve_auto_discover_*`, `resolve_persisted_env_*` — plist reading (including past the intervening nested fd-limit `<dict>`, XML unescaping, and truncated/missing sections) and the preservation decision table.

The env lookup is injected rather than read from `std::env`, so no in-process test mutates a shared process env from a parallel test binary. No `launchctl` was run, `service install` was never invoked, and no live unit was touched.

### Verification

```
$ cargo test -p trusty-search --test no_auto_discover_env
running 3 tests
test env_var_rejects_an_unrecognised_spelling ... ok
test env_var_accepts_the_documented_numeric_spelling ... ok
test env_var_accepts_every_documented_spelling ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.01s

$ cargo test -p trusty-search --bin trusty-search
test result: ok. 349 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 0.44s

$ cargo test -p trusty-search --bin trusty-search commands::service::tests
running 8 tests
test commands::service::tests::build_launchd_config_sets_fd_limit ... ok
test commands::service::tests::build_launchd_config_keeps_alive_after_clean_exit ... ok
test commands::service::tests::launchd_env_pairs_carries_forward_installed_tunables ... ok
test commands::service::tests::build_launchd_config_plist_carries_no_auto_discover_arg ... ok
test commands::service::tests::build_launchd_config_omits_no_auto_discover_by_default ... ok
test commands::service::tests::build_launchd_config_plist_has_unconditional_keepalive ... ok
test commands::service::tests::build_launchd_config_plist_includes_fd_limit ... ok
test commands::service::tests::launchd_env_pairs_never_carries_no_auto_discover ... ok
test result: ok. 8 passed; 0 failed

$ cargo test -p trusty-search --bin trusty-search service_unit
test result: ok. 18 passed; 0 failed; 0 ignored; 0 measured; 334 filtered out

    Finished `dev` profile [unoptimized + debuginfo] target(s)      # zero warnings

$ cargo fmt --check
    # clean

line-cap: measured 3638 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 55 spec doc(s) + 3052 code file(s); 0 error(s), 0 warning(s)

test-pointers: resolved 20980 Test: citation(s) (floor 200) — 0 dangling pointers — OK.

$ bash scripts/check_doc_numbers.sh
doc-numbers: scanned 97 doc(s) / 91 claim(s) (floors 20/20); 4 grandfathered, 0 violations — OK.
```

`cargo test -p trusty-search` (whole crate) leaves 4–7 failures **locally on macOS** in the `service::watcher*` / `service::watch_loop` FSEvents set — the known flake **#4731**. A varying subset fails each run and all 22 pass when run serially (`--test-threads=1 watcher watch_loop` → `22 passed; 0 failed`). This branch touches no file under `src/service/`, and CI's `Test` job is green (16m30s), which confirms it.

### Cross-platform gating

The first CI run caught something local `-p trusty-search` clippy could not: on Linux, everything in `service_unit` except `parse_truthy_bool` is reachable only from `commands::service`, which compiles out on non-macOS — so the plist reader and the preservation decision were dead code under `-D warnings`. Fixed in `b1ab25d1`: `parse_truthy_bool` and its tests stay ungated (the flag exists on every platform and `main.rs` wires the parser unconditionally); the launchd-unit items and their tests moved behind `#[cfg(target_os = "macos")]`, matching how `commands::service` already gates itself. The gate is about who calls them, not about platform APIs.

Verified before pushing by flipping every `target_os = "macos"` in the crate to `"linux"` and re-running `cargo clippy -p trusty-search --all-targets -- -D warnings` — zero errors, zero dead-code warnings in that build shape.

**All 26 CI checks pass** (3 skipped), including `Clippy`, `Test`, `Format check`, `MSRV check (1.91)`, `trusty-search daemon smoke test`, the 500-line cap, the doc-comment pointer lint, and the per-PR changelog fragment gate.

**No version bump** — `crates/trusty-search/Cargo.toml` is byte-identical to `origin/main` (`git diff origin/main -- crates/trusty-search/Cargo.toml` is empty). The 0.42.1 → 0.42.2 bump visible in build output came in on main ahead of this branch, not from here. Changelog goes in `crates/trusty-search/changelog.d/4823-service-install-preserves-no-auto-discover.md` as a fragment.


bobmatnyc:
## trusty-review verdict: APPROVE (grade A-)

`mcp__trusty-review__review_pr` ran against head `b1ab25d1`. Note: trusty-search was `degraded` (0/0 indexes loaded) at review time, so the automated verdict is flagged non-authoritative by the tool itself. All findings below were independently cross-checked by manual code review against the diff.

Findings (all low-confidence, no correctness defects; verdict APPROVE regardless):
- `scan_tags` breaks on a missing close tag mid-document — advisory only, intentional per `parse_installed_unit_tolerates_missing_sections`.
- `require_equals = true` + `num_args = 0..=1` + `default_missing_value` interaction — verified correct and covered by `cli_tests::bare_flag_still_means_true`.
- `#[cfg(target_os = "macos")]` gating on `NO_AUTO_DISCOVER_ENV`/`NO_AUTO_DISCOVER_ARG` — verified no non-macOS compile break.

### Manual verification of both self-disclosures

**1. Unauthored test file (`tests/no_auto_discover_env.rs`)** — read in full. Confirmed:
- It tests the real clap env-var path (`TRUSTY_NO_AUTO_DISCOVER`) by spawning the compiled binary, which is the one path no in-process test can reach.
- `handle_start` (`commands/start/daemon.rs`) rejects `--data-dir` in its very first block via `create_dir_all` before any lockfile/port bind/spawn — the test's scratch dir is pre-populated with a *file* at the data-dir path, so `create_dir_all` fails immediately and the process exits before touching any live daemon or real data dir. Confirmed safe.
- Contents are scoped strictly to this purpose — no extraneous logic.

**2. Cross-platform simulation corruption** — diffed `b1ab25d1` against the true merge-base (`4c412ae1`) with `git diff --stat`. Exactly the 7 intended files are touched (README, changelog fragment, `commands/mod.rs`, `commands/service.rs`, new `commands/service_unit.rs`, `main.rs`, `tests/no_auto_discover_env.rs`). No stray `#[cfg(target_os = ...)]` edits in any unrelated file — the corrupted files were fully restored before commit.

### Substance checks (a)-(d)

- **(a)** `TRUSTY_NO_AUTO_DISCOVER` is absent from `PERSISTED_ENV_VARS` (`src/service/daemon.rs`), and `resolve_persisted_env` additionally filters it out explicitly as defense-in-depth. A generated unit can never carry an env value the parser would reject.
- **(b)** `parse_truthy_bool` is fail-closed: unrecognised input (e.g. `"ture"`) returns `Err`, never a silent `false`. Covered by unit tests and the subprocess test `env_var_rejects_an_unrecognised_spelling` (asserts clap exit code 2).
- **(c)** Bare `--no-auto-discover` still works via `num_args = 0..=1` + `default_missing_value = "true"`; `build_launchd_config_plist_carries_no_auto_discover_arg` and `build_launchd_config_omits_no_auto_discover_by_default` assert against the **rendered plist XML**, not just the config struct (the #4709 lesson).
- **(d)** `KeepAlive: <true/>` is asserted to survive regeneration alongside the suppression flag in `build_launchd_config_plist_carries_no_auto_discover_arg`.

**Context claim verified**: `InstalledUnit::suppresses_auto_discover()` checks `ProgramArguments` first (flag wins), and `resolve_persisted_env` unconditionally drops `TRUSTY_NO_AUTO_DISCOVER` from the regenerated `EnvironmentVariables` regardless of what the installed unit carried — so a live plist with both the hand-added flag and the redundant env entry converges to flag-only on the next `service install`, exactly as described.

Proceeding to merge per pre-authorization once CI is green at current head.

PR #4838, opened 2026-08-04
refactor(trusty-mpm): collapse duplicated routing and budget prose in the PM prompt

bobmatnyc:
Six surfaces answered "which agent handles what" for the same ~11 agents, and the direct-action budget was stated four times. This collapses both to one statement each, drops a roster line the harness already supplies, fixes a YAML parse defect, adds two prose rules, and rewrites `BASE-AGENT.md` in a terse register.

**Compiled PM prompt: 66,628 → 58,564 bytes (−12.2%, ~2,180 tokens saved per session).**
**`BASE-AGENT.md`: 18,327 → 16,270 bytes (−11.2%), including the two new prose rules it gained.**

## What changed

**1. Six routing surfaces → one.** `Agent Routing` (core), `When to Delegate to Each Agent`, `Ops Agent Routing`, `Make / Mise Command Routing` and `Common User Request Routing` became one Routing Table in `agent-delegation.md`, plus the generated roster. `core.md` keeps the `## Agent Routing` heading as a pointer.

The table lives in `agent-delegation.md`, not `core.md`, deliberately: the `AGENT-DELEGATION` override token exists so a project can change routing. Moving routing into the unoverridable `core` would have quietly retired that customization contract. The per-agent default-model column moved with it; the governing model rule (Model Selection Protocol) stays in `core`, so nothing left the floor.

Every routing fact that existed only in a deleted table was folded in and is now pinned by `the_surviving_routing_table_covers_every_folded_mapping` — 26 assertions. `mise run` matters most: it appears nowhere in the Prohibitions table, so the Make/Mise section was the *only* statement that `mise run` targets are delegated at all.

**2. `Handles:` dropped from roster entries.** Verified byte-identical to the description the harness publishes in its own Agent-type catalog (checked `api-qa`, `code-critic`, `data-engineer` word-for-word against a live catalog). `Role:` and `Model:` are what the harness does *not* supply, so they stay. `AgentSummary.description` is still parsed — it is public API.

**3. Direct-action budget stated once.** Canonical text stays in `enforcement.md`; `core.md`, `identity.md` and `non-overridable-rules.md` point at it by title.

The floor stays self-sufficient: `enforcement.md` is itself part of the floor (`base_pm()` = Identity + Enforcement + NonOverridableRules + FrameworkConventions, since #4573), so a session receiving only the floor still gets the full canonical rule. `the_budget_survives_every_override_configuration` already documented that `identity.md`'s copy is deletable by an `IDENTITY` block while enforcement's survives — this makes that the only copy.

**4. YAML block-scalar parse fixed.** `description: >` was taken literally, so `copyeditor`, `pangram-editor`, `proofreader`, `writer` and `writing-critic` each rendered as `- **Handles:** >`. Folded (`>`) and literal (`|`) headers with any chomping indicator now parse. Fixed in `parse_frontmatter` rather than the shared `parse_kv_line` — block scalars need multi-line lookahead, which is a document-level concern, and keeping it local holds this at rung 3.

**5. Two prose rules added** to "Prose Style — Write Plainly" and mirrored into `BASE-AGENT.md`: no praise for the user, and delete the framing opener. Both state the ban as a **category or template** with examples explicitly marked non-exhaustive, because a literal example list had already failed to generalize here.

**6. `BASE-AGENT.md` rewritten in a terse register** — imperative mood, tables over prose, rules stated once, examples kept only where the rule alone is ambiguous.

## Nothing was lost

All 139 rule markers from the pre-rewrite inventory verified present after (whitespace-normalized match). One example was dropped — the Self-Action `WRONG/CORRECT` code block — because the rule plus its forbidden-phrase list is unambiguous without it. Every other example survives, including all four phrase lists, the completion-format block, and the register-announcement examples.

`pm_authority_doctrine_survives_composition` guards a live refusal incident by exact string. Four of its needles broke on casing and rewrapping during the register pass; the **original sentences were restored verbatim** rather than the gate being edited.

Mirror parity — two copies exist (`trusty-agents-common` has no file copy, only a marker constant in `agent_schema.rs`, which matches unchanged):

```
b7727c5ee004b0b1159758169d40089291086bfba66a56b068dea1d7e5040807  crates/trusty-mpm/src/assets/agents/BASE-AGENT.md
b7727c5ee004b0b1159758169d40089291086bfba66a56b068dea1d7e5040807  crates/trusty-code/src/assets/agents/BASE-AGENT.md
```

## Test rung

**Rung 3** — localized behavior inside one crate. `trusty-code`'s only change is a mirrored markdown asset.

| Gate | Result |
|---|---|
| `cargo check -p trusty-mpm` | clean |
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | zero warnings |
| `cargo test -p trusty-mpm` | 4473 passed, 1 failed (see below), 7 ignored |
| `bash scripts/check_line_cap.sh` | 3636 files, 0 violations |
| `cargo test -p trusty-agents-common` | 1600 passed, 0 failed |

The three golden prompt snapshots were regenerated; that diff is the acceptance artifact for this change.

### The one red test is the documented environmental flake

```
expected Doctor, got Error("daemon unreachable: error sending request for
url (http://127.0.0.1:54150/api/v1/doctor?...)")
```

`CLAUDE.md` names this test's timing as a known-environmental flake on this machine (`docs/reference/test-ladder-baseline.md`). It **passed earlier in this same session** on an identical tree, and this diff touches zero files in `client/executor/`. Change-specific gates pass; this one is pre-existing and environmental.

`trusty-code` has two unrelated pre-existing failures in `cli::daemon_autospawn` (`stub must have recorded its argv: NotFound` — a missing build artifact). This branch's only change to that crate is `src/assets/agents/BASE-AGENT.md`.


bobmatnyc:
## Verdict: APPROVE

Adversarial review, independent of the implementer. Every claim below was verified against the diff and against a clean run of the gate, not taken from the PR body.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/claude_md_sections_tests.rs` | 1207–1247 | `the_budget_survives_every_override_configuration` covers "no overrides", hostile `CORE`, hostile `IDENTITY`, and the retired-file case — but not hostile `ENFORCEMENT`. `enforcement` is an ordinary overridable section (lines 912–923 assert an `ENFORCEMENT` block *does* replace the tables). With the budget now stated once, an `ENFORCEMENT` block deletes the only copy and leaves `identity.md:7-10`, `non-overridable-rules.md:4-6` and `core.md` pointing at a title no longer in the prompt. Pre-PR, three restatements survived that override. Consistent with the #4286 accepted outcome — that override already removes the Prohibitions table the budget scopes — but the PR body's self-sufficiency argument covers only the `core.md`-absent case and is silent on this one. | Add a hostile-`ENFORCEMENT` arm asserting the accepted outcome explicitly, exactly as the authority-table test records it at lines 912–923. | Parent |
| MEDIUM | `crates/trusty-mpm/src/core/claude_md_sections_tests.rs` | 1483–1509 | `routing_lives_on_exactly_one_surface` checks all four retired headings against `SectionId::AgentDelegation` only, and checks `core` for one exact table header. A future edit that reintroduces `## Ops Agent Routing` into `core.md` — or into any third section — passes. The test's name claims more than it asserts. (It also re-derives `authored_run(&[AgentDelegation])` inside the loop while `core` is bound but unused there.) | Run the retired-heading loop over the fully composed prompt rather than one section. | Parent |
| LOW | `crates/trusty-mpm/src/core/delegation_authority.rs` | 104–110 | `read_block_scalar` calls `.trim()` on every body line, so a literal (`\|`) scalar keeps its line breaks but loses interior indentation. Harmless for a one-paragraph `description`; would silently flatten any structured field if the parser is reused. `literal_block_scalar_description_keeps_its_line_breaks` documents the line breaks, not the indentation loss. | Note the flattening in the `What:` doc line so the next reader does not assume YAML-literal fidelity. | Promote |

Zero CRITICAL, zero HIGH.

## The two specific questions

### 1. `assert_budget_intact` — legitimate, not coverage laundering

The needle list went **5 → 6**, and both halves are still asserted:

| Half | Needle (all now from `enforcement.md`) |
|---|---|
| Up-front estimate | `"The PM delegates when it believes a task will take more than 3 direct actions"`, `"or when it is unable to complete the task in 3"`, **new** `"Anything you believe needs more than 3 direct actions is delegated, never begun"` |
| Mid-flight handoff | **new** `"The estimate is not a licence to finish"`, `"Do not take a fourth direct action to finish work you misjudged"` |
| Override | `"The user can always override"` |

The two replaced needles (`"MID-FLIGHT HANDOFF rule, not only a pre-task estimate"`, `"never carried on to a fourth direct action"`) matched `identity.md` and nothing else. Their replacements target `enforcement.md:33-38`, which states the same two halves and, per `the_budget_survives_every_override_configuration:1221-1238`, is the copy that survives an `IDENTITY` override while `identity.md`'s does not. Re-keying moved the assertion onto the durable copy — strictly stronger, and the count went up, not down.

Does it still fail if the rule is removed? Yes. All six needles are literal sentences from `### The direct-action budget (P1 and P5 only)`. Delete that subsection and all six fail, plus `the_direct_action_budget_is_stated_once_and_pointed_at_elsewhere`, which independently requires the `### `-prefixed heading in `enforcement` and the pointer-by-title in `Core`/`Identity`/`NonOverridableRules` while forbidding those three from restating `"One direct action = one PM-executed step"`.

The `unwrapped()` blockquote strip is required, not convenient: the canonical governing sentence sits inside a `>` quote at `enforcement.md:27-29`, so a `>` lands mid-sentence at every wrap point. Stripping it removes a dependency on where the source happens to wrap. It cannot mask a deletion — it only normalizes whitespace and quote markers.

**Not laundering.** The code changed and the test moved with it, but onto stronger ground.

### 2. All three brief items landed

| Item | Landed | Evidence |
|---|---|---|
| Roster `Handles:` dropped | Yes | The `out.push_str(&format!("- **Handles:** {handles}\n"))` block is gone from `generate_authority`; `Role:` / `Model:` / `Foundation:` remain. All three golden prompts contain **0** occurrences of `Handles:`. `generate_authority_omits_the_harness_supplied_description` asserts both the label and the description text are absent. `grep -rn "Handles:" crates/ scripts/ docs/` finds no consumer that parses it. `generate_authority_renders_an_agent_with_neither_role_nor_model` covers the now-reachable heading-only entry. |
| Budget deduped 4 → 1 | Yes | `enforcement.md` keeps the canonical subsection; `core.md`, `identity.md`, `non-overridable-rules.md` point at it by exact title. Floor self-sufficiency holds: `the_direct_action_budget_states_both_halves_in_the_floor` projects `Identity + Enforcement + NonOverridableRules + FrameworkGuaranteedConventions` **alone** and passes — `core.md` being overridden away leaves the full rule intact, because `enforcement.md` is itself in that set. See the MEDIUM above for the one configuration nobody asserts. |
| Five broken `- **Handles:** >` entries | Yes, and more generally than asked | Fixed in `parse_frontmatter` at the **key-value level**, so `role:` and `model:` are covered too — the "same parse bug may affect `Role:`" concern is answered by construction, not by assumption. Four tests: folded, literal, fence-termination, and `a_plain_value_beginning_with_an_angle_bracket_is_not_a_block_scalar` guarding false positives. `AgentSummary.description` is still parsed, so the fix still matters for the public-API consumers even though the roster no longer renders it. |

### The 46% → 12.2% gap: the audit over-projected. No work is missing.

Measured, not argued:

| Surface | Before | After | Delta |
|---|---|---|---|
| `testdata/pm-prompt-bundled-fallback.md` (synthetic roster) | 58,512 | 56,153 | −2,359 B |
| `testdata/pm-prompt-roster-absent.md` | 58,318 | 55,959 | −2,359 B |
| Compiled prompt, real 37-agent roster (PR claim) | 66,628 | 58,564 | −8,064 B |

The prose dedup is worth 2,359 bytes — identical in both goldens, which is the arithmetic check that it is prose and not roster. The remaining ~5,700 bytes is the `Handles:` removal against the real roster, consistent with the ~1,400-token audit figure. The two sum to the claimed 12.2%.

A 46% cut would need ~30 KB. Nothing in the three brief items can produce that: after this PR the compiled prompt's bulk is `core.md`'s workflow and `enforcement.md`'s tables, which the brief never proposed touching. The projection was wrong, not the delivery.

## General review

- **No rule lost.** Heading-by-heading comparison of `BASE-AGENT.md`: **25 headings before, 25 after, same set, same order**. Nothing was dropped, only rewritten. Spot-checked the rules most likely to be casualties of a terse pass — all present: injection-skepticism channel distinction (`:38-43`), Native-First connector routing (`:85-91`), No Subagent Fan-Out including the untyped-dispatch clause (`:105-116`), attribution footer with the `Co-Authored-By` ban (`:59-63`), changelog-fragment rules (`:65-76`), Empty-Output Protocol, `## Verification Results` format, `0 tests ran`.
- **`mise run`.** Present in the survivor and pinned by three of the 26 assertions in `FOLDED_ROUTING_MAPPINGS`: ``"`make` and `mise run`"``, `"the PM never runs one directly"`, ``"every `make` and `mise run` target"``. The author's claim that this mapping existed only in the deleted Make/Mise section holds — it is absent from the Prohibitions table, which names `make` (CB#7) and not `mise`.
- **Mirror byte-parity.** Both copies hash `b7727c5ee004b0b1159758169d40089291086bfba66a56b068dea1d7e5040807`. `scripts/check_agent_assets.sh` → `30 byte-parity file(s), all match — OK`.
- **The dropped Self-Action `WRONG/CORRECT` block.** Judged acceptable. The surviving text carries the rule, the four forbidden openers, and the positive instruction ("run the command, report the actual output, interpret it, take the next action"). The deleted example demonstrated exactly what that sentence already states. It was the one example in the file whose removal costs nothing.
- **The two new prose rules.** Present in both `core.md` and `BASE-AGENT.md`. The sycophancy rule bans "the CATEGORY … not a list of strings" and labels its four examples `Non-exhaustive examples:`. The framing-opener rule states the template ``One <noun> that <its significance, or your relation to it>:`` and labels its instances "illustration only — the rule is the template above, never this list". `the_prose_rules_ban_categories_not_phrase_lists` pins all six of those properties, so a future edit cannot quietly demote either rule back into a phrase list.
- **`pm_authority_doctrine_survives_composition`.** Lives in `crates/trusty-mpm/src/core/bundle_tests.rs`, which does **not** appear in this PR's changed-file list. The gate is structurally untouched — verified from the file set, not from the author's account of it. The four needles that broke were fixed in the prose, which is the correct direction.

## Test status

Reproduced independently on the PR head, clean worktree:

```
test result: FAILED. 4473 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 64.43s
test client::executor::tests::execute_doctor_against_test_daemon ... FAILED
```

```
cargo fmt --check                                        clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings   zero warnings
bash scripts/check_line_cap.sh   3636 files, 0 violations — OK
bash scripts/check_agent_assets.sh  30 parity files, all match — OK
```

**The flake clears the baseline protocol.** Same counts as the PR body, reproduced by a third party. The failure is a connectivity condition (`daemon unreachable`), not an assertion about behavior; `CLAUDE.md` names this test's timing as a known-environmental flake on this machine; and the diff touches zero files under `crates/trusty-mpm/src/client/`. Rung 3 is the right rung — `trusty-code`'s only change is a mirrored markdown asset — and the change-specific gates are green.

## Notes

- Two of the three findings are test-thoroughness, not correctness, and neither blocks. The first is worth the two-minute fix before merge only because the PR body makes a self-sufficiency argument that stops one configuration short of the one nobody asserts.
- Changelog fragments: three for `trusty-mpm`, one for `trusty-code`, correctly split by category and directly in `changelog.d/`.
- Not flagged: the register rewrite's individual wording choices. Every heading survives and every rule marker I sampled is present; judging the terse phrasing sentence-by-sentence is a style preference, LOW at most, and I have no evidence any of it changes meaning.

bobmatnyc:
Added the ENFORCEMENT arm to `the_budget_survives_every_override_configuration` per the code-critic finding: an ENFORCEMENT override now deletes the budget rule outright (it's the sole remaining copy after dedup), mirroring `the_authority_tables_survive_every_override_configuration`'s `hostile_floor` arm. Updated the test's doc comment to record the scope change rather than renaming it, matching the precedent set for the authority-table test at #4286.

PR #4841, opened 2026-08-04
fix(trusty-mpm): anchor .trusty-mpm/ to the project, not the worktree (#4832)

### LINKED ISSUE #4832: chore(trusty-mpm): establish .trusty-mpm/ directory convention — framework, sessions, logs only
Defect. Project-local `.trusty-mpm/` configuration violates a three-directory structure. Owner ruling (2026-08-04): "worktrees should be for mutable elements, specifically code/config/docs."

## The Structure

Worktrees hold tracked, editable working sets — code, config, docs. Branch-scoped.
- **Worktree** contains: `CLAUDE.md` (branch-scoped override input), source code, docs
- **Main checkout** contains: `.trusty-mpm/` with framework config (project-stable), sessions (per-session output), and logs
- **Worktrees carry no `.trusty-mpm/` directory at all.** `tm` must stop creating one there.

The load-bearing idea is input/output split: `CLAUDE.md` is a tracked input (branch-scoped per delegation); compiled artifacts are untracked runtime output (per session, not per project).

## Directories & Location

All framework config lives in one project-local directory:

```
.trusty-mpm/
├── framework/           — project-stable config (manifest.toml)
├── sessions/
│   └── <session-id>/
│       ├── INSTRUCTIONS-COMPILED.md
│       └── session-<ts>.md
└── logs/
```

The framework/ directory holds only project-stable assets. Compiled instructions are SESSION-SCOPED output written per session under `sessions/<session-id>/`, not once per project.

Two rules behind it:
1. **User-level instructions do not work.** Instructions are per-session and must always be deployed locally. No `~/.trusty-mpm/` instruction surface.
2. **manifest.toml follows the same rule** — deployed locally per project, inside `framework/`.

## Gitignore Coverage

The claim that `.trusty-mpm/` and its subdirectories are unignored was measured against a stale worktree checkout predating the relevant `.gitignore` line. That measurement is invalid.

**Correct facts on `origin/main`:** `.gitignore:62` carries `.trusty-mpm/*`, which ignores `framework`, `sessions`, `logs`, and `last-instructions.md`. Only the bare directory name (`.trusty-mpm/` itself) is unmatched — this is deliberate and documented at lines 55-63.

**Real gap:** The target project's `.gitignore` scaffolding block written by `core::scaffold_gitignore::SCAFFOLD_IGNORED_PATHS` names only `.claude/` subtrees. It does not include the `.trusty-mpm/` paths that tmux worktrees will create. PR [#4841](https://github.com/bobmatnyc/trusty-tools/pull/4841) fixes this gap.

## Sequencing Consequence

Compiled instructions now live per session, not per project, removing concurrent-session overwrites. [#4794](https://github.com/bobmatnyc/trusty-tools/issues/4794)'s hash-gated content-addressed archive reverts to dedup and history value — no longer load-bearing for correctness. Sequences before or with this issue.

## Manifest Premise — Two Distinct Artifacts

Audit at `7c7c3489` reveals the issue states something false. Two entirely different artifacts exist:

- **framework-manifest.toml** is the intended design: compiled into the binary via `include_str!` at [`crates/trusty-mpm/src/assets/framework-manifest.toml`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/assets/framework-manifest.toml) and read in [`crates/trusty-mpm/src/manifest/framework.rs:64`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/manifest/framework.rs#L64). It is the agent/skill bundling authority, read in memory only, never deployed to disk. This is confirmed design and requires no changes.
- **manifest.toml** is a separate operator-facing override with three-layer precedence at [`crates/trusty-mpm/src/manifest/resolve.rs:82-90`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/manifest/resolve.rs#L82-L90): `<project>/.trusty-mpm/framework/manifest.toml` (highest), `~/.trusty-mpm/manifest.toml` (user layer), then the synced catalog.

**Resolution:** Two edits, not a move. Repoint the project layer to `framework/manifest.toml`, and delete the user layer. Deleting the user layer removes a working feature; it does not fix drift.

## Evidence

Audit at origin/main `7c7c3489`:

- [`crates/trusty-mpm/src/core/session_launch/model_inject.rs:43-55`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/session_launch/model_inject.rs#L43-L55) — `INSTRUCTIONS-COMPILED.md` is write-only in the project-level call; bytes delivered to harness come from session-local temp file
- [`crates/trusty-mpm/src/core/claude_md_sections.rs:82`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/claude_md_sections.rs#L82) — `HOST_FILES = ["CLAUDE.md"]` resolved only against `project_dir`; `OverrideTier::User` exists but never constructed in production
- [`crates/trusty-mpm/src/core/delegation_authority.rs`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/delegation_authority.rs) — agent roster scans `<project_dir>/.claude/agents` per-worktree
- `CLAUDE.md` is tracked on `origin/main` (14.8k); current stale worktree carries untracked ~570-token copy (branch predates tracked version) — proof that instructions differ per branch

## In-Scope Defects

**Dead read can abort session launch.** [`build_instructions`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/instruction_pipeline.rs#L676) reads `<framework_root>/instructions/INSTRUCTIONS.md`. Its output (`PipelineOutput::merged`) has zero call sites — content is never used. `NotFound` is safe, but any other error (permissions, path-is-a-directory, invalid UTF-8) propagates to `PrepError::Instructions`, the one variant [`is_fatal()`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/session_launch/mod.rs#L382-L384) returns true for. Nothing writes that file anymore; it is either absent or stale cruft whose only effect is killing a session over unread content. In scope because directory cleanup removes it.

**Launching outside a git repo scatters `.trusty-mpm/`.** [`resolve_dir(None)`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/core/mod.rs) returns `current_dir()`; directories that are not recognized git repos route to `start_session_in_place`, so `prepare_session` writes `.trusty-mpm/` wherever the shell happened to be. This likely explains the stray `~/.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` observed on disk. No global writer exists on this SHA — `compiled_prompt_path` only joins `project_dir`, and [#4752](https://github.com/bobmatnyc/trusty-tools/issues/4752) removed the old global writer.

**Second writer touches only the stash.** [`compose_session_instructions`](https://github.com/bobmatnyc/trusty-tools/blob/7c7c3489/crates/trusty-mpm/src/bin/tm/commands/session.rs#L606-L641) (reached from `tm session instructions`) independently recomputes the prompt and writes only `last-instructions.md`, never the compiled file. This explains the observed 285-byte drift between files. [#4794](https://github.com/bobmatnyc/trusty-tools/issues/4794) cannot simply delete the stash without deciding what replaces that command; note it as a constraint, not work for this issue.

## Resolution

Deploy all framework config to project-local `.trusty-mpm/framework/` only. Compiled instructions live per session under `.trusty-mpm/sessions/<session-id>/`. No user-level instruction or manifest surfaces. Stop creating `.trusty-mpm/` inside worktrees.

## Cross-Links

[#4756](https://github.com/bobmatnyc/trusty-tools/issues/4756) (consolidate CLAUDE.md writers), [#4794](https://github.com/bobmatnyc/trusty-tools/issues/4794) (hash-gated instructions archive), [#3427](https://github.com/bobmatnyc/trusty-tools/issues/3427) (gitignore coverage).

bobmatnyc:
Closes #4832

## What changed

Owner ruling (2026-08-04): a worktree holds the mutable, tracked working set — code, config, docs — and nothing else. Harness state belongs to the **project**.

### New: `core::harness_root` — how the project root is resolved

This was the flagged hard part. The registry (`project::registry`) turned out **not** to be usable: a `Project` record carries `name`/`repo_url`/`default_branch` and **no local path**, and it is async and daemon-owned. It cannot answer "what is this directory's project root".

The resolution is git-authoritative instead, reusing the crate's existing hardened invocation (`session_manager::worktree_safety::git_command`, widened `pub(super)` → `pub(crate)`, which strips `GIT_DIR`/`GIT_WORK_TREE`/… so git cannot be redirected):

1. `git rev-parse --path-format=absolute --git-common-dir`
2. `<root>/.git` → its parent (normal checkout **and every linked worktree of it**); a bare clone → itself
3. a bare clone named `.base` → its **parent**. This is not a path guess: `provisioner::workspace::BASE_CHECKOUT_DIRNAME` creates `<project>/.base` and adds every managed session's worktree from there, and `session_manager::worktree_registry` already interrogates `<repo>` and `<repo>/.base` as the two anchors of one project.
4. `None` outside any working tree — the load-bearing result that drives the refusal below.

Verified live against this machine's dual-repo topology (a `.base/` bare clone with worktrees under `.base/.worktrees/`, plus an ordinary checkout), and pinned by tests that build **real** git repos and worktrees — `harness_root_is_the_main_checkout_for_a_worktree`, `harness_root_maps_a_base_clone_back_to_the_project`, `harness_root_is_the_repo_root_from_a_subdirectory`, `harness_root_for_is_none_outside_a_git_repo`.

### 1. Compiled prompt → per session

`<project>/.trusty-mpm/sessions/<session-id>/INSTRUCTIONS-COMPILED.md`. Project scoping still let two concurrent sessions in one project overwrite each other; per-session removes that by construction **for managed sessions** — the ones that have an id. See the `local` bucket below for what it does not cover.

`session_scope` resolves the `<id>` segment: explicit id → `TM_MANAGED_SESSION_ID` → `local`. The id is threaded where it exists — `WorkspaceProvisioner::provision_in` and the daemon's `prepare_inproject_session` now call a new `prepare_session_for_managed`; `spawn`/`spawn_resume` pass theirs into `build_prompt_file`; the in-place relaunch passes `None` and picks the pane's env var up.

**One design decision the issue did not specify:** three of six production `prepare_session` call sites (in-place start, standalone load, deploy-validate repair) and `tm connect` run *before* any session id exists. They land in a documented `local` bucket rather than an invented per-call id, which no later reader could match to a session. **This bucket does not remove collisions by construction:** two concurrent unmanaged launches in one project still share `sessions/local/` and can still overwrite each other. It is strictly narrower than the one-file-per-project it replaces, not collision-free. Flagging it for review.

The env value becomes a path component, so it is validated: `[A-Za-z0-9._-]`, never `.`/`..` (`session_scope_rejects_a_path_traversal_segment`).

### 2. `manifest.toml` project layer → `framework/`, user layer deleted

Project layer is now `<project>/.trusty-mpm/framework/manifest.toml`. The `~/.trusty-mpm/manifest.toml` user layer and the `framework_root` parameter it needed are gone. **This removes a working feature — an owner ruling, not a bug fix.** `user_level_manifest_exclude_is_honored` was inverted into `user_level_manifest_is_ignored` rather than deleted, so the layer cannot silently return.

`framework-manifest.toml` was not touched — confirmed-correct `include_str!` design.

### 3. Worktrees get no `.trusty-mpm/`; gitignore

The two writers that fire on **every** launch — the compiled prompt and the `last-instructions.md` stash — now resolve against the harness root, as do `tm project init`'s scaffolding and the manifest read. `tm project init` also seeds `framework/`.

**The issue's gitignore claim is refuted at `origin/main`.** Verified: `.trusty-mpm/*` (`.gitignore:62`) already ignores `framework`, `sessions`, `logs` and `last-instructions.md`; only the bare directory name is unmatched, which is deliberate and documented at lines 55–63, and `git status` is clean with the full tree present. `.gitignore` is byte-identical between `7c7c3489` and `origin/main`.

The **real** gap was one level down: `core::scaffold_gitignore::SCAFFOLD_IGNORED_PATHS` — the block `tm` writes into a *target project's* `.gitignore` — named only the three `.claude/` subtrees. It now also covers `.trusty-mpm/sessions/`, `.trusty-mpm/logs/` and `.trusty-mpm/last-instructions.md`, deliberately sparing `framework/` and `config.toml` (operator-authored, may be tracked).

### 4. Dead read removed

`build_instructions` no longer reads `<framework_root>/instructions/INSTRUCTIONS.md`. `PipelineOutput::merged`, `PipelineOutput::instructions_loaded` and `PipelineInput::framework_instructions_path` go with it — all three existed only to serve that read and had zero production consumers (`instructions_loaded` included; grep confirms tests only). The fatal gate on the compiled-prompt **write** is untouched.

### 5. Non-git launch refused

`refuse_outside_a_git_project` gates `start_session` before `start_session_in_place`. A repo with no parseable remote still falls through to in-place — the refusal is about "no project", not "no GitHub remote".

## Migration — chosen: automatic, best-effort, at every launch

`prepare_session_inner` calls `remove_legacy_compiled_prompt` against **both** the directory it was handed (the worktree — every worktree of an upgraded project holds a copy, since the old path resolved against it) and the harness root. It deletes `.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` and then removes `framework/` **only when empty**, so an operator's `manifest.toml` survives. Failures are logged, never fatal.

Not chosen: a one-time documented cleanup (leaves the file until the operator reads a changelog) or moving the file (there is no session id to move it *to*).

The `manifest.toml` project layer is **not** auto-moved: it is operator-authored content and moving a user's file silently is worse than the layer being read from its new home. Called out in the changelog.

## Test ladder — rung 5 (process lifecycle + persistence)

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo check -p trusty-mpm --all-targets` | pass |
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | pass, 0 warnings |
| `cargo test -p trusty-mpm` | 4483 passed, 1 failed (baseline, below), 7 ignored |
| `bash scripts/check_line_cap.sh` | 3636 files, 7 allowlisted, 0 violations |
| `bash scripts/check_sld.sh` | 55 specs + 3052 code files, 0 errors, 0 warnings |

A full green run of the same suite was also observed at 4484 passed / 0 failed.

### Baseline failure

```
thread '...' panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18:
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:53804/api/v1/doctor?...)")
```

Change-specific gates pass; `execute_doctor_against_test_daemon` is the known environmental timing failure recorded in [`docs/reference/test-ladder-baseline.md:19`](docs/reference/test-ladder-baseline.md) ("takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"). This diff touches **zero** files under `crates/trusty-mpm/src/client/` (`git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/` → 0). Nothing was ignored, cfg-gated or excluded.

### Tests that fail before this change

- `compiled_prompt_path_never_lands_inside_a_worktree` — real repo + real linked worktree
- `compiled_prompt_path_is_per_session`
- `session_start_refuses_a_non_git_directory` / `session_start_accepts_a_git_directory`
- `pipeline_does_not_read_the_retired_framework_instructions` + the e2e sibling (directory planted at the retired path used to return the fatal variant)
- `prepare_session_for_managed_writes_the_per_session_compiled_prompt`
- `prepare_session_migrates_off_the_legacy_compiled_prompt`
- `user_level_manifest_is_ignored`, `resolve_ignores_a_user_level_manifest`
- `block_covers_session_output_but_not_project_config`
- the six `harness_root_*` / four `session_scope_*` cases

## Note for the reviewer

Out of scope, deliberately: `.trusty-mpm/` joins that do **not** fire on a normal launch keep resolving against the directory they are given — `claude_config` checkpoints, `quarantine_shadows`, `meta/launch` meta-sessions, and standalone `managed.toml`. Those are per-workspace state by design; folding them in would have widened this PR without evidence they cause the defect.


bobmatnyc:
## Verdict: WARN

Adversarial review, rung 5. Zero CRITICAL, one HIGH → the code proceeds, but the HIGH must be tracked. All findings are in `harness_root_for` and its downstream consumers; the migration, the refusal, the gitignore fix and the field deletions all hold up.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/harness_root.rs` | 100–105 | The `.base` remap never checks bareness, so a **non-bare** checkout whose directory is literally named `.base` resolves to its **parent** — harness state written outside the repository, the exact scatter this PR refuses elsewhere. Code contradicts its own doc ("a bare clone named `BASE_CLONE_DIRNAME`"). | Carry the bare/non-bare decision out of the `if` and gate the remap on it (see below). | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/harness_root.rs` | 97–99 | Submodules and `--separate-git-dir` fall into the "bare → itself" branch. Verified live: a submodule answers `<super>/.git/modules/<name>`, `--separate-git-dir` answers `<store>.git`. Harness state, `tm project init` scaffolding and the `manifest.toml` read all land there, so a submodule's own `.trusty-mpm/framework/manifest.toml` is silently never read. | Same fix: `git rev-parse --is-bare-repository` to branch, and for non-bare take the main worktree from `git worktree list --porcelain` (first entry) instead of inferring it from the common dir's name. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/instruction_pipeline.rs` | ~440 | `remove_dir(&framework)` runs **unconditionally**, even when no legacy file was found, on every launch. `scaffold_project_dir` (`project.rs:178`, added in this same PR) creates `.trusty-mpm/framework/` and reports it to the operator — the next launch deletes it. Two components of one PR disagree about whether `framework/` should exist. | `if removed { let _ = std::fs::remove_dir(&framework); }` — also makes the migration effectively one-shot instead of a per-launch rmdir attempt. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/manifest/resolve.rs` | 95–99 | The project manifest moved on **both** axes (`.trusty-mpm/manifest.toml` → `framework/manifest.toml`, and `project_dir` → harness root), and `resolve_manifest` logs nothing for an absent layer. An operator's existing override stops taking effect with zero runtime signal. Not moving the file is the right call; silently ignoring it is not. | In `ManifestSources::resolve`, when `<project_dir>/.trusty-mpm/manifest.toml` exists and the new path does not, `tracing::warn!` naming both paths and the move. | Fix here |
| LOW | `crates/trusty-mpm/src/core/harness_root.rs` | 60–70 | Two concurrent **unmanaged** in-place launches in one project both resolve to `sessions/local/` and still overwrite each other's compiled prompt. The PR body's "per-session removes that by construction" is true only for managed sessions. | Keep `local` (an invented per-call id would be strictly worse); soften the claim in the PR body/changelog to "removes it for managed sessions". | Parent |
| LOW | `crates/trusty-mpm/src/core/session_launch/mod.rs` | 1314 | `session_scope` reads ambient `TM_MANAGED_SESSION_ID` inside the generic `prepare_session_inner`. `tm session start` for project B, run from inside a pane owned by session A, writes B's prompt into `<B>/.trusty-mpm/sessions/<A-id>/`. Path validation prevents escape, so this is mislabeling only. | Honor the env var only on the in-place-relaunch path (`claude_code.rs:397`), where it is meaningful. | Promote |

### The HIGH, reproduced

```
$ mkdir -p proj/.base && cd proj/.base && git init -q .
$ git rev-parse --path-format=absolute --git-common-dir
/tmp/.../proj/.base/.git
```

`common_dir.file_name() == ".git"` → `root = /tmp/.../proj/.base` → `root.file_name() == ".base"` → `harness_root_for` returns `/tmp/.../proj`, a directory that is not this repository and may be a different one. The `.base` test runs on `root` regardless of which branch produced it:

```rust
let root = if common_dir.file_name().is_some_and(|n| n == ".git") {
    common_dir.parent()?.to_path_buf()
} else {
    common_dir
};
if root.file_name().is_some_and(|n| n == BASE_CLONE_DIRNAME) { /* fires for BOTH */ }
```

Checked and **not** a problem: a project at a path that merely *contains* `.base` (e.g. `~/.base/myproject`) is safe — only the final component is tested.

## Required Changes

1. Gate the `.base` remap on the repository actually being bare (HIGH). Preferably by the `--is-bare-repository` / `worktree list --porcelain` route, which closes the submodule and `--separate-git-dir` cases in the same edit.
2. Add a regression test for a non-bare checkout named `.base` and one for a submodule — the current suite builds real repos and worktrees, so both are a few lines each.
3. Gate the `remove_dir` on `removed`.
4. Warn on a legacy project `manifest.toml` left at the old path.

## What I verified and am not flagging

- **Migration blast radius is bounded and safe.** `remove_legacy_compiled_prompt` only ever touches the literal `.trusty-mpm/framework/INSTRUCTIONS-COMPILED.md` plus a `remove_dir` that refuses a non-empty directory. Even under the HIGH's misresolution the worst outcome is deleting another project's regenerable compiled prompt. Race-free against a concurrent session (`remove_dir` → `ENOTEMPTY` → logged, non-fatal). Idempotent and cheap on repeat launches.
- **`git_command` widening is safe.** `pub(super)` → `pub(crate)` exposes a *hardened* invocation — it strips `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_CEILING_DIRECTORIES`, `GIT_NAMESPACE`, `GIT_CONFIG_COUNT`, and pins only status/quoting config that is inert for `rev-parse`. It covers every redirect vector; `GIT_CEILING_DIRECTORIES` removal in particular means a hostile env fails toward `None` (refusal), never toward a wrong root. Widening it makes misuse *less* likely, not more.
- **`instructions_loaded` deletion is correct.** Independently grepped `origin/main`: the only references are its own declaration/assignment in `instruction_pipeline.rs` and five test assertions. Same for `PipelineOutput::merged` (tests plus one historical comment). The brief's "preserve it" rested on a false premise.
- **The issue's gitignore claim is refuted; the fix targets the real gap.** `.gitignore:62` on `origin/main` is `.trusty-mpm/*`, which matches `framework`, `sessions`, `logs` and `last-instructions.md`; only the bare directory name is unmatched, deliberately and documented at lines 55–63. The issue's `git check-ignore` run is consistent with a worktree predating that rule. The real gap was `SCAFFOLD_IGNORED_PATHS` — the block `tm` writes into a *target* project — and the fix covers it while correctly sparing `framework/` and `config.toml`.
- **The non-git refusal is well placed and clear.** It sits at the routing point before `start_session_in_place`, names the directory and gives two remedies; a repo with an unparseable remote still falls through, which is the right distinction.
- **Tests build real git state.** `git init`, `git worktree add`, `git clone --bare` + a worktree from it — not temp-dir simulacra. The `harness_root_*` and `session_scope_*` cases could not compile before the change; `block_covers_session_output_but_not_project_config` asserts three paths absent from `SCAFFOLD_IGNORED_PATHS` on main.
- **Nothing was disabled to reach green.** No `#[ignore]`, `cfg`-gate or exclusion added anywhere in the diff. The one deleted test is `resolve_user_over_catalog`, whose subject (the user manifest layer) no longer exists; `resolve_catalog_over_default` replaces it, and `user_level_manifest_exclude_is_honored` was inverted rather than dropped.
- **Baseline failure clears the protocol.** `execute_doctor_against_test_daemon` is the recorded environmental timing failure at `docs/reference/test-ladder-baseline.md:19`; the raw output is a connection error, not an assertion, and the diff touches zero files under `crates/trusty-mpm/src/client/`.
- **500-SLOC cap respected** on every touched production file; changelog fragment present at `crates/trusty-mpm/changelog.d/4832-trusty-mpm-dir-convention.md`, correctly placed, category-first, and accurate to the diff.

bobmatnyc:
## Review fixes — all five findings addressed (`f4c3e0e`)

### 🔴 HIGH — `.base` remap ignored bareness

`harness_root_for` no longer applies the rewrite to a `root` computed after the branch. It now branches on whether `dir` sits in a **linked worktree** (`--git-dir` != `--git-common-dir`), and the `.base` rewrite is factored into `map_base_clone_to_project`, which additionally requires `git rev-parse --is-bare-repository`.

Your repro now returns the repo itself. Pinned by `harness_root_for_a_non_bare_repo_named_base_is_itself` (real non-bare repo at `<proj>/.base`, asserts it resolves to itself and *not* to its parent).

### MEDIUM 1 — submodules and `--separate-git-dir`

**I did not take the suggested fix, because it does not work.** `git worktree list --porcelain` reports the *git directory*, not the working tree, for exactly these two shapes — git computes the main worktree as `common_dir` with a trailing `/.git` stripped, which is the same derivation the code already had. Measured:

```
submodule          → worktree /super/.git/modules/sm      ← wrong
separate-git-dir   → worktree /store.git                  ← wrong
```

What I used instead: for a non-linked, non-bare working tree, `git rev-parse --show-toplevel`. That is inside the working tree by construction, so it is correct for every shape where the git directory is not:

| shape | `--git-common-dir` | resolved root |
|---|---|---|
| submodule | `<super>/.git/modules/sm` | `<super>/sm` |
| `--separate-git-dir` | `<store>.git` | the working tree |
| non-bare repo named `.base` | `<proj>/.base/.git` | `<proj>/.base` |
| bare `.base` clone (+ its worktrees) | `<proj>/.base` | `<proj>` |
| linked worktree | `<main>/.git` | `<main>` |

Bare repositories keep the git-directory derivation, since they have no working tree to ask. Tests: `harness_root_for_a_submodule_is_the_submodule_checkout`, `harness_root_for_a_separate_git_dir_checkout_is_the_working_tree`.

### MEDIUM 2 — `remove_dir` fighting the scaffolder

`remove_legacy_compiled_prompt` now only attempts the `framework/` removal when it actually removed a legacy file. `remove_dir`'s not-empty guard could not tell "empty leftover" from "empty because `tm project init` just created it". Test: `migrate_keeps_an_empty_framework_dir_when_nothing_migrated`.

### MEDIUM 3 — the old `manifest.toml` was silently ignored

The file is still deliberately **not** moved. When the new path has no manifest and a legacy `<project>/.trusty-mpm/manifest.toml` exists, `resolve_manifest` now logs one `warn` naming both paths. Detection is split into a pure `stranded_legacy_manifest` so the condition is assertable without capturing a subscriber: `stranded_legacy_manifest_is_detected`, `stranded_legacy_manifest_is_silent_once_moved`, `stranded_legacy_manifest_is_silent_when_there_never_was_one`.

### LOW — PR-body overclaim

Corrected in the body, and in the changelog fragment, which carried the same claim. Per-session scoping removes collisions by construction **for managed sessions only**; two concurrent unmanaged launches still share `sessions/local/` and can still collide.

### Gates — rung 5, all green

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo check -p trusty-mpm --all-targets` | pass |
| `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | pass, 0 warnings |
| `cargo test -p trusty-mpm` | `4491 passed; 0 failed; 7 ignored` |
| `bash scripts/check_line_cap.sh` | 3638 files, 7 allowlisted, 0 violations |
| `bash scripts/check_sld.sh` | 55 specs + 3052 code files, 0 errors, 0 warnings |

Fully green including `execute_doctor_against_test_daemon`, so this run has no baseline caveat. One earlier run of the same suite hit `daemon::managed_routes::tests::stale_assets_for_many_sees_an_agent_change_on_the_very_next_call` (a `#[serial]` test behind a `fake_home()` env guard); it passed 5/5 in isolation and 4491/4491 on the clean re-run. This diff touches zero files under `crates/trusty-mpm/src/daemon/`. Nothing was ignored, cfg-gated or excluded.

bobmatnyc:
### Follow-up: `Per-PR changelog fragment` CI gate (`1c7836b`)

Pre-existing failure on this PR, not caused by the review fixes: the single fragment stacked `Changed` / `Fixed` / `Removed`, and only line 1 counts as a category — the other two would have rendered as body text under `### Changed`. Split into three files (`-changed`, `-fixed`, `-removed`), same content. `bash scripts/check_changelog_fragment.sh` → `OK trusty-mpm`, and `assemble-changelog.sh trusty-mpm --stdout` now renders all three sections.

bobmatnyc:
## Verdict: APPROVE (delta re-review of `77ed99c7`, `1c7836bd`)

Zero CRITICAL, zero HIGH. Every finding from the [previous review](https://github.com/bobmatnyc/trusty-tools/pull/4841#issuecomment-5185358153) is resolved. I re-reviewed the delta only.

**The rejected suggestion was right to reject.** I tested my own proposal before judging the replacement, and it fails exactly as measured:

```
$ git -C <super>/sm      worktree list --porcelain | head -1
worktree /tmp/.../sup/.git/modules/sm          # the git dir, not <super>/sm
$ git -C <sep>/work      worktree list --porcelain | head -1
worktree /tmp/.../sep/store.git                # the git dir, not the work tree
```

`git worktree list` derives the main worktree by stripping a trailing `/.git` from the common dir — the same derivation the code already had — so my fix would have changed nothing for the two shapes it was aimed at. The replacement (branch on `--git-dir != --git-common-dir` to detect a linked worktree; `--show-toplevel` otherwise) is the correct discriminator, and it is a better fix than the one I proposed.

## Resolution table — verified against real repos, not accepted

I rebuilt every shape with real `git init` / `submodule add` / `--separate-git-dir` / `clone --bare` / `worktree add` and ran the new algorithm's exact rev-parse sequence against each. All five claimed rows hold, plus seven I added:

| # | shape | resolved root | correct |
|---|---|---|---|
| 1 | non-bare repo named `.base` | `<proj>/.base` | ✅ (the original HIGH; my round-1 repro no longer misresolves) |
| 2 | submodule | `<super>/sm` | ✅ |
| 3 | `--separate-git-dir` | `<sep>/work` | ✅ |
| 4 | bare `.base` clone itself | `<proj>` | ✅ |
| 5 | worktree of bare `.base` | `<proj>` | ✅ |
| 6 | linked worktree | `<main>` | ✅ |
| 7 | subdirectory of a checkout | repo root | ✅ |
| 8 | detached-HEAD worktree | `<main>` | ✅ |
| 9 | inside `<repo>/.git/` | `None` → refusal | fails safe (see Notes) |
| 10 | worktree whose main **moved** | `None` → refusal | fails safe |
| 11 | worktree whose main **deleted** | `None` → refusal | fails safe |
| 12 | outside any repo | `None` → refusal | ✅ |

Rows 10 and 11 were the failure shapes I was asked to hunt for: `--git-common-dir` itself fails once the main checkout is gone, so the resolver refuses rather than resurrecting a deleted path and writing into it. That is the right answer.

## The other three fixes

- **MEDIUM 2** — `remove_dir` now gated on `removed` (`instruction_pipeline.rs`), with the reasoning recorded inline. `tm project init`'s `framework/` scaffold survives the next launch. Correct.
- **MEDIUM 3** — the warn is **reachable and visible**, which is the part worth checking rather than the unit tests. It fires in the `None` arm of `read_layer(path, "project")` — i.e. exactly when the new path is absent, which is the whole stranded case (`resolve.rs:148-158`). The tm binary's filter is `EnvFilter::try_from_default_env().unwrap_or_else(|_| "info".into())` (`main.rs:292`), so a `warn!` reaches stderr with no `RUST_LOG` set. `stranded_legacy_manifest` is a pure predicate over paths, correctly `None` once the file has moved so it does not nag forever.
- **LOW** — the "removes it by construction" overclaim is corrected in both the PR body and the fragment.

## Changelog split — genuine gate failure, not a workaround

Verified against `scripts/check_changelog_fragment.sh` on `origin/main`. The script carries a dedicated `stray_category_lines()` validator that **rejects** a fragment holding a second category, and its header documents the exact failure at line 38: *"SILENTLY MIS-RENDERED: only line 1 is a category, so the rest became body"*. Line 321-322: *"carries exactly ONE category: everything after line 1 is copied through verbatim, so a second category stacked into the same file renders…"*. The original single file stacked `Changed` / `Fixed` / `Removed`, so `Fixed` and `Removed` would have rendered as body text under `Changed`. Splitting into three files is the required fix. The `Per-PR changelog fragment` check is now green.

## Test evidence — I ran it myself, and did not reproduce fully green

CI on this repo runs lint jobs only; there is no Rust build or test job, so the local run is the only test evidence and I did not take it on trust. One batched gate in a clean worktree off the branch head:

```
cargo fmt --check                                       → pass
cargo clippy -p trusty-mpm --all-targets -- -D warnings → pass
test result: FAILED. 4490 passed; 1 failed; 7 ignored; finished in 372.68s
thread 'client::executor::tests::execute_doctor_against_test_daemon'
  panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18
```

Correcting the PR's claim: **4491 passed / 0 failed was not reproducible here.** `execute_doctor_against_test_daemon` failed again on my run. It does not change the verdict — this is the failure recorded at `docs/reference/test-ladder-baseline.md:19` (9–13s test against a 10s client timeout), the diff touches zero files under `crates/trusty-mpm/src/client/`, and my run was under deliberately heavy load (two concurrent `cargo test` invocations), which is the exact condition the baseline names. But the PR body should say "baseline failure recurs under load", not "fully green".

On the `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call` dismissal: **sound.** It passed in my run, the diff touches zero files under `src/daemon/`, and 5/5 isolated plus a clean re-run is the documented protocol. One thing to keep in mind if it recurs: `ManifestSources::resolve` now calls `harness_root::framework_dir`, which shells out to git, so `harness_root` costs 2–3 process spawns per call and roughly 8–12 per launch. That adds subprocess work to a path daemon tests exercise, which raises timing sensitivity without breaking isolation. Session launch is not a hot path, so this is an observation, not a finding.

**Nothing was disabled to reach green.** No `#[ignore]`, `cfg`-gate, exclusion, or test deletion in the delta. Four tests added: `harness_root_for_a_non_bare_repo_named_base_is_itself`, `harness_root_for_a_submodule_is_the_submodule_checkout`, `harness_root_for_a_separate_git_dir_checkout_is_the_working_tree`, `migrate_keeps_an_empty_framework_dir_when_nothing_migrated`, plus three `stranded_legacy_manifest_*`. The one deleted function is `git_common_dir`, superseded by `git_rev_parse_path`. `harness_root.rs` is 92 SLOC; the 500-line cap check is green.

## Notes — observed, deliberately not flagged

- **Row 9 is an unmentioned behavior change.** Running from inside `<repo>/.git/` used to resolve to the repo (`.git` → parent); it now returns `None` and refuses, because `--show-toplevel` fails there. Refusing is defensible and fails safe, but it is a delta the PR body does not mention. Below the threshold to flag — nobody launches `tm` from `.git/hooks/`.
- **`stranded_legacy_manifest` only checks the harness root**, so a legacy `manifest.toml` sitting inside a *worktree* goes unwarned. The realistic case — an operator-authored file in the main checkout — is covered, and worktrees are tm-provisioned and ephemeral, so this is the right scope.

APPROVED for the next pipeline stage.

Issue #1917, opened 2026-07-02
tm: skill deployment silently no-ops because its source directory is only populated by a separate `tm install` step


bobmatnyc:
## Problem

`deploy_skills_filtered`'s source is resolved via `skill_source_dir()`, which falls back to the on-disk `~/.trusty-mpm/framework/skills/` directory. That directory is populated ONLY by the separate, explicit `tm install` command (`crates/trusty-mpm/src/bin/tm/commands/install.rs:27-70`, walks the `bundle_all::ALL` table). It is NOT populated or refreshed by `prepare_session`/`prepare_session_inner` at session-launch time — that code just copies whatever's already sitting in `~/.trusty-mpm/framework/skills/`, assuming it's up to date.

Consequence: if `tm install` was run under an older binary (e.g. before the `mpm-*` → `tm-*` skill rename in #1905) and never re-run since, `~/.trusty-mpm/framework/skills/` still holds stale/absent content. `deploy_skills_filtered` then iterates zero matching source files, returns an empty (non-error) result, and skill deployment silently no-ops on every subsequent session launch — no error, no warning, nothing.

Confirmed live: after building/installing the `7c3b0522` binary (includes the `tm-*` skill rename) and spawning a session, `~/.claude/skills/` still had zero `tm-*` directories and the deployment manifest was completely empty (`{"managed": {}}`), while 26 stale `mpm-*` directories remained (untouched, since the ownership-manifest system doesn't know about pre-manifest deploys).

**Secondary, related gap**: neither code path (`crates/trusty-mpm/src/bin/tm/commands/session.rs:52-75` for the legacy `tm sessions start` command, nor the daemon-side `prepare_inproject_session` added by #1913 in `crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs`) ever prints or logs a "Skills: X deployed" line — even though `PrepReport` carries `report.skill_deploy` (`crates/trusty-mpm/src/core/session_launch/mod.rs:41-45`). So this failure mode is completely invisible in normal output; only inspecting `~/.claude/skills/` directly reveals it.

## Fix

1. Make session preparation self-healing: either have `prepare_session_inner` detect a stale/missing `~/.trusty-mpm/framework/skills/` (e.g. compare against the binary's embedded `bundle_all::ALL` version/checksum) and auto-refresh it, or have session launch trigger the equivalent of `tm install`'s skill-materialization step automatically rather than depending on a separate manual command ever having been run.
[...]

Issue #2504, opened 2026-07-13
bug(trusty-mpm): composed per-agent files untracked by deploy manifest — BASE-AGENT guidance updates (PR #2503) never propagate to deployed agents


bobmatnyc:
## Bug Summary

During the #2503 install-verify (2026-07-13), a critical scope gap was discovered in the agent deployment manifest tracking: **bundled-agent guidance updates cannot reach the composed per-agent files**.

## Facts (Verified During Deploy)

- `~/.claude/agents/.trusty-mpm-manifest.json` tracks **ONLY** the 5 `BASE-*.md` files as managed (fresh checksums after `tm install --force`).
- None of the ~36 composed per-agent files (`version-control.md`, `local-ops.md`, `rust-engineer.md`, etc., all timestamped Jul 10 from an earlier tm deploy) appear in the manifest at all.
- `AgentManifest::is_managed()` returns `false` for these files, causing `deploy_agents_filtered` to conservatively skip them as user-owned — on both plain `tm install` and `tm install --force` (`--force` only governs the framework-artifact step, not the composition checksum gate).
- **Net effect:** PR #2503's new BASE-AGENT sections ("PM Directives Do Not Bind You", "Foreground Execution", closing postmortems #2501/#2502) are live in `~/.claude/agents/BASE-AGENT.md` but in **ZERO composed agent files** — the remediation is inert for every actually-dispatched agent.
- Claude Code reads the per-agent files, not `BASE-AGENT.md`, so deployed agents never see the guidance updates.

## Root Cause

The per-agent files were deployed before per-file manifest tracking covered them (or fell out of tracking), and there is no adoption or reset path. Files not in the manifest can never be reconciled again.

## Suggested Fix

1. **Adoption logic**: At deploy time, if a target file is untracked but byte-identical to what the CURRENT compose pipeline would have produced from any prior bundled baseline — or more practically, untracked with no evidence of user modification — offer/perform adoption into the manifest (possibly behind a prompt or flag).

2. **Explicit reset command/flag**: e.g. `tm install --reset-agents [names…]` that force-recomposes named (or all) agent files, backing up any that differ from fresh composition, and registers them in the manifest going forward.

3. **Fresh deploy registration**: Ensure fresh deploys register every composed file in the manifest so the gap can't re-open.

## Operational Impact

**CRITICAL:** Until fixed (or a manual reset is run), the #2501 idle-watcher and #2502 PM-directive remediations do not apply to any live agent.

[...]

PR #4848, opened 2026-08-04
fix(trusty-mpm): auto-deploy bundled agent assets at session start

### LINKED ISSUE #4840: BASE-AGENT.md changes never reach running agents until tm install is run
## Defect

Changes to `BASE-AGENT.md` never reach running agents. The agent asset deploy chain has three layers: `bundle::ALL` (compiled in) → `~/.trusty-mpm/framework/agents/` → `$CLAUDE_CONFIG_DIR/agents/`. The second hop already ran on every managed spawn. The **first** hop — materializing the compiled bundle into the framework directory — only ever ran under `tm install`. This left the source directory as the stale layer.

Rebuilding the binary does not trigger the first hop. Starting a session does not trigger it. So a merged, compiled-in instruction change silently has no effect until someone remembers to re-run `tm install`.

## Evidence, measured 2026-08-04

- `$CLAUDE_CONFIG_DIR` resolved live to `/Users/masa/.trusty-tools/trusty-mpm/claude-config`.
- `$CLAUDE_CONFIG_DIR/agents/BASE-AGENT.md` mtime: **Aug 1 10:36:49**. The deploy tracker `.trusty-mpm-manifest.json` agrees: `"BASE-AGENT.md": { "deployed_at": "2026-08-01T14:36:49...", "origin": "bundled" }`.
- PR [#4790](https://github.com/bobmatnyc/trusty-tools/pull/4790) merged **Aug 4 19:52 UTC**, adding three sections to `BASE-AGENT.md`.
- The installed binary `/Users/masa/.cargo/bin/tm`, built Aug 4 16:51 — *after* the merge — **does** contain the text: `strings tm | grep -c "Never Directly Monitor a Declarative Process"` → 14 hits; `"No Subagent Fan-Out"` → 7 hits.
- A subagent dispatched in that session confirmed on itself: its delivered system prompt contained none of the three new sections, and carried a heading name that PR [#4795](https://github.com/bobmatnyc/trusty-tools/pull/4795) had already renamed.

So: source correct, binary correct, deployed file three days stale, agents running on stale instructions. The first hop (bundle to framework directory) was the layer that never refreshed.

## Impact

Every `BASE-AGENT.md` change since Aug 1 is inert on this machine, including the no-subagent-fan-out rule from [#4796](https://github.com/bobmatnyc/trusty-tools/issues/4796) and the declarative-process monitoring rule from [#4790](https://github.com/bobmatnyc/trusty-tools/issues/4790). The failure is silent — nothing warns that the deployed file is older than the binary.

This directly caused a wasted investigation: [#4837](https://github.com/bobmatnyc/trusty-tools/issues/4837) was initially framed as "a rule existed and agents ignored it," when the rule had never been delivered.

## Resolution direction (not a prescribed design)

Deploy the agent assets automatically instead of requiring a manual step — at session start, or gated on a version/checksum comparison so it is cheap when nothing changed. State as an open question whether auto-deploy should respect the existing user-edited-file protection (a checksum mismatch is currently treated as user-owned and skipped), since that protection is what makes a silent skip possible.

Note a related hazard for whoever implements: running `tm install` mid-session against live services may not be safe, so an auto-deploy needs to be scoped to agent assets rather than reusing the full install path.

bobmatnyc:
Closes [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840).

## The defect

`crates/trusty-mpm/src/assets/agents/*.md` is compiled into the binary, but the only code path that ever wrote it to disk was the manual `tm install`. Rebuilding the binary did not trigger it; starting a session did not trigger it. Measured on 2026-08-04: a deployed `BASE-AGENT.md` three days older than the binary containing the fix, with no warning anywhere.

The chain is `bundle::ALL` (compiled in) → `~/.trusty-mpm/framework/agents/` → `$CLAUDE_CONFIG_DIR/agents/`. The second hop already ran on every managed spawn. The **first** hop only ever ran under `tm install`, so the source directory was the stale layer — confirmed live: source `Aug 3`, deploy `Aug 1`, binary `Aug 4`, source and deploy byte-identical and both missing the merged text.

## Trigger and gate

- **Trigger:** session provisioning — `managed_config::ensure_managed_config_dir_with_root`, called on every daemon-managed spawn/resume from `runtime/claude_code.rs::prepare_managed_config`.
- **Gate:** a sha256 stamp over the `agents/*` slice of `bundle::ALL`, compared against `<agents_dir>/.bundle-stamp`. An unchanged bundle costs one file read and one string compare. This is the same mechanism `skill_source` has used since [#1917](https://github.com/bobmatnyc/trusty-tools/issues/1917) — agents were the missing half.
- **Scope:** agent assets only. It does not reuse the `tm install` path, which is unsafe to run mid-session against live services.

## Decision on drifted bundled files

**Auto-deploy overwrites them.** The deploy manifest records `origin`, and `agents::deployer` already treats framework-owned drift as corruption rather than user ownership ([#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408)); auto-deploy inherits that unchanged rather than inventing a second policy. Only an *untracked* file that differs, or a *user-owned* (seed-once tier) entry that was edited, is preserved.

**Every preserved file is now named in a warning.** That silence was the other half of the defect — it is what let a three-day-stale `BASE-AGENT.md` go unnoticed. Warnings go to both `tracing::warn!` and stderr, and the untracked case names the `tm install --reset-agents <name>` that resolves it.

## Fail open

`autodeploy_agents` never returns an error. A refresh or deploy that cannot run degrades to a warning line; the session still starts with whatever is already on disk. A stale file is a strictly better failure than a session that will not launch.

## Files

- `crates/trusty-mpm/src/core/agent_source.rs` (new) — stamp, materialize+prune, refresh, warning formatter, `autodeploy_agents` / `autodeploy_agents_for` (the latter carrying the `agents/agents` git-submodule guard, so a source checkout's tracked submodule is deployed from but never overwritten).
- `crates/trusty-mpm/src/core/managed_config.rs` — routes the agent deploy through it, logs the warnings.

## Gates — rung 5 (session lifecycle)

```
cargo fmt --check                            — clean
cargo check -p trusty-mpm                    — clean
cargo clippy -p trusty-mpm -- -D warnings    — clean
cargo test -p trusty-mpm                     — 4481 passed, 1 failed, 7 ignored
bash scripts/check_line_cap.sh               — 3643 files, 0 violations
```

The single failure is `client::executor::tests::execute_doctor_against_test_daemon`, the documented environmental flake in [docs/reference/test-ladder-baseline.md](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/reference/test-ladder-baseline.md) ("takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"). It passes in isolation on this branch at 10.15 s, and this diff touches zero files under `crates/trusty-mpm/src/client/`.

22 new tests, covering the four required behaviors: deploy fires when the checksum differs, is skipped when identical, fails open when the target (or the source) is unwritable, and warns by name when it skips a user-modified file — plus the drifted-bundled overwrite decision and the submodule guard.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4840-agent-asset-autodeploy.md
+++ b/crates/trusty-mpm/changelog.d/4840-agent-asset-autodeploy.md
@@ -1,0 +1,6 @@
+Fixed
+
+- Bundled agent assets now reach running agents without a manual `tm install` (closes [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840))
+  - session provisioning re-materializes `~/.trusty-mpm/framework/agents/` from the compiled-in bundle and deploys it, gated on a sha256 stamp so an unchanged bundle costs one file read
+  - a bundled-origin file whose checksum drifted is still overwritten (corruption, not user ownership); an untracked-and-differing or user-edited file is still preserved — but is now named in a warning instead of being skipped silently
+  - the whole step fails open: a refresh or deploy that cannot run degrades to a warning and never blocks the session
--- a/crates/trusty-mpm/changelog.d/4840-agent-asset-autodeploy.md
+++ b/crates/trusty-mpm/changelog.d/4840-agent-asset-autodeploy.md
@@ -1,6 +1,9 @@
 Fixed
 
-- Bundled agent assets now reach running agents without a manual `tm install` (closes [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840))
-  - session provisioning re-materializes `~/.trusty-mpm/framework/agents/` from the compiled-in bundle and deploys it, gated on a sha256 stamp so an unchanged bundle costs one file read
-  - a bundled-origin file whose checksum drifted is still overwritten (corruption, not user ownership); an untracked-and-differing or user-edited file is still preserved — but is now named in a warning instead of being skipped silently
+- Bundled agent assets now reach a **daemon-managed** session without a manual `tm install` (closes [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840))
+  - managed session provisioning re-materializes `~/.trusty-mpm/framework/agents/` from the compiled-in bundle and deploys it, gated on a sha256 stamp so an unchanged bundle costs one file read
+  - the standalone path (`tm run` / `tm load` / `tm login`) still routes through `global_config::ensure_global_config_dir`, which does not refresh — tracked as [#4849](https://github.com/bobmatnyc/trusty-tools/issues/4849)
+  - a bundled-origin file whose checksum drifted is still overwritten (corruption, not user ownership); an untracked-and-differing or user-edited file is still preserved — but now surfaces in a bounded count-plus-preview warning carrying the `tm install --reset-agents <name>` pointer, instead of being skipped silently
+  - agents that fail to compose (bad frontmatter, strict-YAML rejection) are reported in the same summary — those do not land at all, which is worse than stale
+  - an uninitialized `agents/agents` submodule directory no longer wins the source resolution and deploys nothing; it falls back to the compiled-in bundle
   - the whole step fails open: a refresh or deploy that cannot run degrades to a warning and never blocks the session
--- a/crates/trusty-mpm/src/core/agent_source.rs
+++ b/crates/trusty-mpm/src/core/agent_source.rs
@@ -1,0 +1,302 @@
+//! Self-healing materialization + auto-deploy of the bundled agent *source*
+//! directory (`~/.trusty-mpm/framework/agents/`) — issue #4840.
+//!
+//! Why: `crates/trusty-mpm/src/assets/agents/*.md` is compiled into the binary
+//! (`bundle::ALL`), but the ONLY code path that ever wrote it to disk was the
+//! separate, manual `tm install`. Rebuilding the binary did not trigger it;
+//! starting a session did not trigger it. So a merged, compiled-in instruction
+//! change (e.g. a new `BASE-AGENT.md` rule) silently had no effect until
+//! someone remembered to re-run `tm install` — measured on 2026-08-04 as a
+//! deployed `BASE-AGENT.md` three days older than the binary containing the
+//! fix, with no warning anywhere. Skills already solved exactly this
+//! (`skill_source::ensure_skill_source_fresh`, issue #1917); agents were the
+//! missing half.
+//!
+//! What: [`agent_bundle_stamp`] fingerprints the compiled-in `agents/*` slice
+//! of [`bundle::ALL`] via sha256 — this is the cheap gate, so the common
+//! "nothing changed" case costs one file read and one string compare.
+//! [`materialize_agent_artifacts`] writes every bundled agent into the source
+//! directory and prunes any `.md` file the current table no longer lists (a
+//! renamed or removed agent). [`ensure_agent_source_fresh`] compares the stamp
+//! against a marker file and re-materializes only on a mismatch.
+//! [`autodeploy_agents`] is the entry point session provisioning calls: it
+//! refreshes the source, deploys it, and returns a report — it NEVER returns
+//! an error, because a broken deploy must not block a session (a stale file is
+//! the strictly lesser failure). Anything it declined to overwrite is named in
+//! [`AgentAutodeploy::warnings`], closing the other half of #4840: the
+//! deployer's user-edited-file protection previously skipped silently.
+//!
+//! Overwrite policy (#4840, decided): a BUNDLED-origin file whose checksum
+//! drifted IS overwritten — the deploy tracker (`.trusty-mpm-manifest.json`)
+//! records `origin`, and `agents::deployer` already treats framework-owned
+//! drift as corruption rather than user ownership (#4408). Auto-deploy
+//! inherits that unchanged. Only an UNTRACKED file that differs, or a
+//! user-owned (seed-once tier) entry that was edited, is preserved — and now
+//! warned about by name.
+//!
+//! Test: `crates/trusty-mpm/src/core/agent_source_tests.rs`.
+
+use std::collections::HashSet;
+use std::path::Path;
+
+use crate::core::agent_deployer::{DeployResult, deploy_agents};
+use crate::core::agent_manifest::{atomic_write, checksum};
+use crate::core::bundle;
+use crate::core::error::Result;
+use crate::core::paths::FrameworkPaths;
+
+/// Marker file recording the last-materialized bundle stamp, written directly
+/// under the agent source directory (`<agents_dir>/.bundle-stamp`).
+///
+/// Mirrors `skill_source::STAMP_FILE_NAME`; hidden (leading `.`) so the prune
+/// sweep never considers it a stale agent.
+const STAMP_FILE_NAME: &str = ".bundle-stamp";
+
+/// Outcome of one [`autodeploy_agents`] run.
+///
+/// Why: the caller needs to know whether anything changed (to log it) and,
+/// critically, WHICH files were left stale — #4840's second half is that a
+/// declined overwrite was silent.
+/// What: `refreshed` is `true` when the source directory was re-materialized
+/// from the compiled-in bundle this run; `deployed` lists the composed agent
+/// files actually (re)written into the target; `warnings` carries one
+/// human-readable line per file left stale, plus one line if the refresh or
+/// the deploy itself failed outright.
+/// Test: every `autodeploy_*` case in `agent_source_tests.rs`.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct AgentAutodeploy {
+    /// Whether the source directory was re-materialized from the bundle.
+    pub refreshed: bool,
+    /// Composed agent filenames (re)written into the target this run.
+    pub deployed: Vec<String>,
+    /// One line per file left stale, or per step that failed. Always safe to
+    /// print verbatim; empty means nothing was declined and nothing failed.
+    pub warnings: Vec<String>,
+}
+
+/// Compute a stable sha256 fingerprint over every `agents/*` entry in the
+/// compiled-in [`bundle::ALL`] table.
+///
+/// Why: this is the gate that makes auto-deploy cheap. It detects "this binary
+/// embeds different agent content than what is on disk" without depending on
+/// file mtimes, which a backup/restore (or a `cargo install` that preserves
+/// timestamps) can shuffle.
+/// What: concatenates `<rel_path>\0<contents>\n` for every table entry whose
+/// `rel_path` starts with `"agents/"`, in table order, and returns the sha256
+/// hex digest.
+/// Test: `agent_bundle_stamp_is_stable_across_calls`,
+/// `agent_bundle_stamp_differs_from_skill_stamp`.
+pub fn agent_bundle_stamp() -> String {
+    let mut buf = String::new();
+    for artifact in bundle::ALL
+        .iter()
+        .filter(|a| a.rel_path.starts_with("agents/"))
+    {
+        buf.push_str(artifact.rel_path);
+        buf.push('\0');
+        buf.push_str(artifact.contents);
+        buf.push('\n');
+    }
+    checksum(&buf)
+}
+
+/// Write every bundled `agents/*` artifact into `agents_dir`, pruning any
+/// `.md` file on disk the current table no longer lists.
+///
+/// Why: `agents_dir` (`~/.trusty-mpm/framework/agents/`) is a framework-owned
+/// artifact directory — every entry is written exclusively by trusty-mpm's own
+/// installer/self-heal path, never hand-edited (the USER-level agent source is
+/// the separate `~/.trusty-mpm/agents/`). So it is safe, and necessary for the
+/// renamed-agent case, to prune rather than only add.
+/// What: creates `agents_dir` if absent, atomically (over)writes each
+/// `agents/*` artifact to `<agents_dir>/<basename>`, then removes any
+/// remaining top-level `*.md` file that is not one of the artifacts just
+/// written. Hidden files (leading `.`, e.g. the stamp marker and the deploy
+/// manifest) are never touched. Returns the basenames written.
+/// Test: `materialize_agent_artifacts_writes_all_agents`,
+/// `materialize_agent_artifacts_prunes_files_not_in_table`.
+pub fn materialize_agent_artifacts(agents_dir: &Path) -> Result<Vec<String>> {
+    std::fs::create_dir_all(agents_dir)?;
+
+    let mut written = Vec::new();
+    let mut keep: HashSet<String> = HashSet::new();
+    for artifact in bundle::ALL
+        .iter()
+        .filter(|a| a.rel_path.starts_with("agents/"))
+    {
+        let basename = artifact
+            .rel_path
+            .strip_prefix("agents/")
+            .unwrap_or(artifact.rel_path);
+        let dest = agents_dir.join(basename);
+        if let Some(parent) = dest.parent() {
+            std::fs::create_dir_all(parent)?;
+        }
+        atomic_write(&dest, artifact.contents)?;
+        keep.insert(basename.to_string());
+        written.push(basename.to_string());
+    }
+
+    for entry in std::fs::read_dir(agents_dir)? {
+        let entry = entry?;
+        let Some(name) = entry.file_name().to_str().map(str::to_owned) else {
+            continue;
+        };
+        if name.starts_with('.') || !name.ends_with(".md") {
+            continue;
+        }
+        if !keep.contains(name.as_str()) {
+            std::fs::remove_file(entry.path())?;
+        }
+    }
+
+    Ok(written)
+}
+
+/// Ensure `agents_dir` reflects the binary's currently-embedded agent bundle,
+/// re-materializing it when missing or stale.
+///
+/// Why: the self-healing entry point behind [`autodeploy_agents`] — it removes
+/// the dependency on a prior, separate `tm install` having refreshed the
+/// source directory (#4840).
+/// What: compares [`agent_bundle_stamp`] against `<agents_dir>/.bundle-stamp`;
+/// when they differ (including "stamp file absent", which covers a missing or
+/// never-materialized directory), calls [`materialize_agent_artifacts`] and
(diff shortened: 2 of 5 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review, rung 5 (session lifecycle + automatic on-disk overwriting). Zero CRITICAL, two HIGH.

### What I attacked and could not break

**The overwrite predicate is sound.** The bundled-vs-user decision is not a path or a heuristic — it is the manifest `origin` field, read under the ledger lock at `crates/trusty-agents-common/src/agents/deployer.rs:302-367`. A hand-edited agent can only be destroyed if it is BOTH manifest-tracked AND `origin.is_framework_owned()`. An untracked file that differs → `skipped` + `untracked_modified` (deployer.rs:324-327). A tracked user-owned entry that was edited → `skipped` (deployer.rs:362-366). A missing manifest → every file is untracked → nothing overwritten. A corrupt manifest → hard `Err` before any file is touched (deployer.rs:220-228), which `autodeploy_agents` converts to a warning and deploys nothing. I could not construct a path from "user's hand-edited agent" to "destroyed".

**The source-dir prune is not new destructive surface.** `materialize_agent_artifacts` (`crates/trusty-mpm/src/core/agent_source.rs:118-154`) overwrites and prunes `~/.trusty-mpm/framework/agents/` with no ownership check at all. That is safe only because that tree is framework-owned by construction: `paths.rs:43` documents it as "the trusty-mpm-owned, auto-materialized bundled tree the user must not edit", and there are zero non-test writers of user or catalog agents into `fw.agents` anywhere in `crates/trusty-mpm/src`. It is byte-for-byte the design `skill_source` has shipped for `framework/skills` since #1917. Accepted.

**The stamp gate holds.** `<rel_path>\0<contents>\n` over the `agents/*` slice (agent_source.rs:89-101) moves on add, remove, rename, and content change. Absent, unreadable, or garbage stamp → `read_to_string(..).ok().as_deref() == Some(current)` is false → refresh (agent_source.rs:173). The stamp is written only AFTER a successful materialize (agent_source.rs:177-178), so a partial failure retries next run rather than latching. No stale-stamp permanent skip on the binary-upgrade path — a new binary always has a different stamp.

**Fail-open verified.** No `unwrap`, `expect`, `panic!`, or index in the new production module. Every IO path is `?`-into-`Result`, and both call sites match it into a warning line (agent_source.rs:251-261, 284-294). `strip_prefix(...).unwrap_or(...)` at agent_source.rs:127-129 is infallible by construction, not a panic. Added per-spawn cost is one sha256 over the agent slice; the compose+lock work was already happening.

**The submodule guard holds.** `agent_source_dir()` returns `self.agents.clone()` in the fallback (`paths.rs:355-363`), so the `source_dir != paths.agents` test at agent_source.rs:282 is exact equality against that same value — a checked-out `agents/agents` submodule is deployed FROM and never materialized OVER.

**Nothing was ignored, cfg-gated, or excluded to reach green.** The full 844-line diff contains no `#[ignore]`, no `#[cfg(...)]` gate, no `--exclude`, no `--lib` narrowing, and deletes no test. `execute_doctor_against_test_daemon` is the documented environmental flake (`docs/reference/test-ladder-baseline.md:19`) and the diff touches zero files under `crates/trusty-mpm/src/client/`. Changelog fragment present and correctly shaped; line-cap CI green.

### Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/managed_config.rs` | 164-167 | One `eprintln!` **per skipped file, on every spawn and resume**, reverses the #2504 decision recorded at `trusty-agents-common/src/agents/deployer.rs:391-411`: "warn ONCE with a count + short preview rather than one line per untracked-modified file (this set can be dozens of files wide on a machine that predates per-file manifest tracking) — actionable without spamming". `untracked_modified` is a subset of `skipped`, so on exactly the machine #2504 was written for this emits dozens of lines at every session start, forever, since nothing reconciles them until the operator runs `--reset-agents`. | Collapse to one line: count + first-5 preview + the single `tm install --reset-agents` pointer, matching the shape deployer.rs already uses. Keep per-file detail at `tracing::warn!` only. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/standalone/global_config.rs` | 123 | The standalone driver (`tm run`/`load`/`login`) still deploys from `managed_root/framework/agents` with no refresh, so on a machine that never starts a daemon-managed session #4840 is untouched. Defensible as a scope call — but the changelog headline (`changelog.d/4840-agent-asset-autodeploy.md:3`) claims it unconditionally: "Bundled agent assets now reach running agents without a manual `tm install`". A standalone-path user reads that as fixed, and it is not. | Qualify the headline bullet to "daemon-managed session provisioning", and file the standalone half. Wiring it is not required to land this. | Fix here (changelog) + Promote (standalone wiring) |
| MEDIUM | `crates/trusty-mpm/src/core/agent_source.rs` | 251-261, 284-294 | `DeployResult::failed` is never read. A per-agent compose failure or strict-YAML rejection (deployer.rs:256-290) means that agent does not land **at all** — strictly worse than stale — and `AgentAutodeploy.warnings` says nothing, so `managed_config` prints nothing. Same silence class the PR exists to close, and the `warnings` doc at agent_source.rs:54-55 ("one line per file left stale, **or per step that failed**") overstates what the code does. | Extend `out.warnings` with one line per `result.failed` entry in both arms. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/agent_source.rs` | 170-180 | Stamp-matches-but-source-drifted is a permanent skip. Delete or corrupt one file under `~/.trusty-mpm/framework/agents/` and the stamp still matches, so `ensure_agent_source_fresh` returns `false` forever and nothing restores it; the target keeps the old deployed copy with no warning. The module header calls itself "Self-healing materialization" (agent_source.rs:1) — against out-of-band source damage it is not. Mirrors `skill_source`, so not a regression. | Either cheap-verify presence of each expected basename before trusting the stamp, or soften the header claim to "self-healing against binary upgrade". | Promote |
| MEDIUM | `crates/trusty-mpm/src/core/agent_source.rs` | 280-295 | An `agents/agents` directory that exists but is EMPTY — the normal state of an uninitialized git submodule — passes `is_dir()` at `paths.rs:357-359`, so `autodeploy_agents_for` takes the submodule branch, skips the refresh, and `deploy_agents` returns the default `DeployResult` from `deployer.rs:184-186`: `refreshed: false`, `deployed: []`, `warnings: []`. Provisioning then logs **nothing at all** and the fix silently does not apply. Dev-machine-only (`trusty_mpm_root` resolves from `current_exe()`), but #4840 was measured on a source checkout. | Warn when the submodule branch yields an empty `deployed`+`unchanged` set — "the agents/agents submodule is empty; run `git submodule update --init`". | Promote |
| LOW | `crates/trusty-mpm/src/core/agent_source.rs` | 251-261 / 284-294 | The deploy-and-collect-warnings block is duplicated verbatim between `autodeploy_agents` and `autodeploy_agents_for`. Any fix to one (e.g. the `failed` finding above) must be applied twice. | Extract `fn deploy_into(source: &Path, target: &Path) -> AgentAutodeploy`. | Fix here |

### Required changes before this is APPROVE-clean

1. Collapse the per-file stderr warning to the #2504 count+preview shape.
2. Qualify the changelog headline to the managed path.
3. Surface `DeployResult::failed` in `warnings`.
4. De-duplicate the deploy block (falls out of 3).

### Notes

- One visibility claim I could not confirm: `ensure_managed_config_dir` is reached from `runtime/claude_code.rs:716` inside the daemon, so the `eprintln!` lands on the daemon's stderr, not the operator's terminal. No less visible than `tracing::warn!`, but not the terminal-visible warning the PR body implies — part of why finding 1 is a cost with no matching benefit.
- Concurrency: `materialize_agent_artifacts` runs outside `with_agent_manifest_lock` (that lock covers the target, not the source). Two concurrent spawns materialize identical content, so the only real race is prune-vs-read, whose fallout is a compose error — which lands in the `failed` bucket finding 3 covers. Not independently rated.
- The standalone gap does not rise to BLOCK: it is pre-existing behavior left in place, not a regression, and the managed path is where every daemon session lives. The misleading changelog line is the part that must change.

Reviewed by `code-critic` against the spec only.

bobmatnyc:
## Review fixes pushed

### 🔴 HIGH 1 — per-file warning reversed the documented decision

Confirmed and fixed. `skip_warning_lines` → `deploy_summary_lines` (`crates/trusty-mpm/src/core/agent_source.rs`): at most **one line for the stale set**, count plus a 5-name preview with elision, carrying the `tm install --reset-agents <name>` pointer. Same policy `agents::deployer` already implements for #2504, not a second one.

The per-file `eprintln!` in `managed_config.rs` is gone. It went to the daemon's stderr, so it bought nothing over the existing `tracing::warn!`, which stays.

New test `deploy_summary_lines_summarises_skips_in_one_line` pins 12 skipped files → exactly 1 line, preview truncated, pointer preserved.

### 🔴 HIGH 2 — changelog overclaimed

Fixed in the existing fragment (no second one added). The headline now reads "reach a **daemon-managed** session", with a dedicated bullet: the standalone path (`tm run` / `tm load` / `tm login`) still routes through `global_config::ensure_global_config_dir`, which does not refresh — [#4849](https://github.com/bobmatnyc/trusty-tools/issues/4849). The standalone fix is **not** implemented here.

### MEDIUM — `DeployResult::failed` never read

Fixed. A compose failure or strict-YAML rejection now gets its own summary line, worded as worse than stale ("failed to compose and did NOT deploy at all — absent, not merely stale"). The `AgentAutodeploy::warnings` doc comment no longer claims per-file coverage it does not have. New test `deploy_summary_lines_reports_failed_agents`.

### Lower-tier item 1 — stamp does not cover the deployed tree — NOT fixed, deliberate

Correct as described, but out of proportion here. The stamp gates only *materialization* of the source dir; `deploy_agents` still runs unconditionally every spawn and checksums the **target** against the manifest, so drift in the deployed tree already self-heals. The uncovered case is out-of-band drift in `~/.trusty-mpm/framework/agents/` itself — a framework-owned directory written exclusively by tm's own installer. Covering it means reading ~40 source files on every spawn and resume, which is exactly the cost the one-file-read stamp exists to avoid. Worth its own issue if it is ever observed in the wild; not worth trading the gate away on speculation.

### Lower-tier item 2 — empty submodule directory — FIXED

Cheap, so done. `agent_source_dir()` prefers `agents/agents` on an `is_dir()` test, which an uninitialized submodule satisfies while holding nothing — so `autodeploy_agents_for` deployed zero agents, silently. It now requires the directory to hold at least one non-hidden `.md` before taking that branch, and otherwise falls back to the compiled-in bundle. New test `autodeploy_agents_for_falls_back_when_the_submodule_is_empty` asserts the bundled roster lands and the submodule directory is still never written.

### Gates

`cargo fmt --check` && `cargo check -p trusty-mpm --all-targets` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` — all pass.

`cargo test -p trusty-mpm` — **4483 passed; 1 failed; 7 ignored**. The one failure is the known environmental baseline, `execute_doctor_against_test_daemon` (`docs/reference/test-ladder-baseline.md:19`) — a connection error, not an assertion:

```
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:51117/api/v1/doctor?...)")
```

`git diff --name-only HEAD -- crates/trusty-mpm/src/client/` returns 0 files; the diff touches only `core/agent_source.rs`, `core/agent_source_tests.rs`, `core/managed_config.rs`, and the changelog fragment. All 18 `core::agent_source::tests` pass.

`bash scripts/check_line_cap.sh` — 3645 files, 0 violations.

bobmatnyc:
## Verdict: APPROVE (re-review, delta only)

Round 2 against `9da4111a`. Both HIGH findings resolved. Zero CRITICAL, zero HIGH. Scope: the delta only — the overwrite predicate, the source-dir prune, the stamp gate, and fail-open were cleared in round 1 and were not re-attacked.

### HIGH 1 — warning spam: resolved

`deploy_summary_lines` (`core/agent_source.rs:226-268`) is genuinely one line for the stale set regardless of width. The elision boundary in `preview` (agent_source.rs:207-224) is correct: `items.len() > limit` means exactly 5 skips renders no `…` and 6 renders one — no off-by-one. `deploy_summary_lines_summarises_skips_in_one_line` pins it at 12 skips → `lines.len() == 1`, and its negative assertion is sound (the line carries `agent-1.md` in the preview, but `agent-11.md` is not a substring of it, so the truncation check is real and not accidentally satisfied).

Deleting the `eprintln!` in `managed_config.rs` loses nothing that mattered — my own round-1 finding was that it landed on the daemon's stderr, not the operator's terminal, so it was never the terminal-visible warning it was billed as. `tracing::warn!` retains the full signal at the same audience.

### HIGH 2 — changelog overclaim: resolved

"reach a **daemon-managed** session" is accurate, not merely softer. I re-confirmed `global_config.rs:123` still joins `managed_root/framework/agents` with no refresh, so the named exclusion is exactly right, and the bullet names the three commands and points at [#4849](https://github.com/bobmatnyc/trusty-tools/issues/4849) — which exists and is OPEN with a title that matches the gap.

### MEDIUM — `DeployResult::failed`: resolved, path reachable

The failure path is genuinely reachable, not a field now merely read: `deployer.rs:264` and `deployer.rs:286-288` populate `failed` on compose failure and strict-YAML rejection respectively, and `deploy_summary_lines` consumes it. Wording is right — "did NOT deploy at all — absent, not merely stale" reads as worse than stale, which it is.

### Empty submodule: resolved, and the fallback target is correct

`autodeploy_agents_for` now gates on `source_dir != paths.agents && has_agent_markdown(&source_dir)` (agent_source.rs:366). The detail that matters is the fallback argument change from `&source_dir` to `&paths.agents` (agent_source.rs:371) — without it, the empty-submodule path would have materialized the compiled-in bundle *into the submodule directory*, which is precisely the destructive act the guard exists to prevent. `autodeploy_agents_for_falls_back_when_the_submodule_is_empty` asserts `read_dir(&submodule).count() == 0` afterwards, so that is pinned. A populated submodule still wins — `autodeploy_agents_for_skips_source_refresh_on_submodule` writes a `.md` first and still asserts `!out.refreshed`. `has_agent_markdown` treating a read error as empty fails toward the compiled-in bundle, never toward writing over a submodule, which is the safe direction.

### The declined item — refusal is correct, premise verified independently

I checked the load-bearing claim rather than taking it: **yes, `deploy_agents` checksums the target against the manifest on every spawn.** `deployer.rs:330-345` reads the deployed file, compares via `manifest.checksum_matches`, and `deployer.rs:345-361` re-deploys framework-owned drift. That runs on every `deploy_agents` call and `deploy_agents` runs on every spawn, so deployed-tree drift self-heals independently of the stamp. The refusal does not rest on a false premise.

The residual gap is therefore exactly as narrow as claimed: out-of-band drift in `~/.trusty-mpm/framework/agents/`, a directory I already confirmed in round 1 has no non-tm writers. One further nuance worth stating — a source file *deleted* out-of-band is never iterated by `deploy_agents`, so the target silently keeps the old copy. That is the whole residual exposure, and trading a one-file stamp read for a ~40-file read on every spawn and resume to close it is not a trade I would make on speculation either. Carried forward as MEDIUM / Promote, unchanged from round 1.

### Test integrity

All three new tests genuinely fail before the change, not just after: `deploy_summary_lines_summarises_skips_in_one_line` (old `skip_warning_lines` returned 12 lines, `assert_eq!(lines.len(), 1)` fails), `deploy_summary_lines_reports_failed_agents` (old code never read `failed` → 0 lines), `autodeploy_agents_for_falls_back_when_the_submodule_is_empty` (old code took the submodule branch → `refreshed: false`, `assert!(out.refreshed)` fails). The delta introduces no `#[ignore]`, no `#[cfg(feature…)]` gate, no `--exclude`, no `should_panic` shim, and deletes no coverage — 4481 → 4483 reconciles exactly as 2 old `skip_warning_lines_*` tests replaced by 4 new ones. `execute_doctor_against_test_daemon` fails with a connection error rather than an assertion and the diff still touches zero files under `crates/trusty-mpm/src/client/`.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/agent_source.rs` | 231-241 | The consolidation dropped a distinction the old code got right. `skipped` mixes two ownership classes; `untracked_modified` is computed by the deployer but no longer consulted, and the summary names `skipped[0]` unconditionally in `Run \`tm install --reset-agents {first}\``. When `skipped[0]` is an *edited seed-once* entry rather than an untracked one, the message names a deliberate user customization as the thing to reset — and `--reset-agents` is exactly the path that makes it receive bundle updates again, i.e. discards the edit. Advice only, and the line does say "user-owned", but it points at a specific file by name. | Name `result.untracked_modified.first()` in the pointer and emit the `--reset-agents` sentence only when that set is non-empty; keep the count/preview over the full `skipped` set. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/agent_source.rs` | 1, 170-180 | Carried forward from round 1, declined with sound reasoning (see above). Out-of-band deletion or corruption inside `~/.trusty-mpm/framework/agents/` is never self-healed, while the module header still opens "Self-healing materialization". The behavior call is right; the header still overstates it by one word. | Soften the header to "self-healing against binary upgrade", or file the gap. No code change needed. | Promote |

### Notes

- The `deploy_into` extraction (agent_source.rs:270-284) closes the round-1 LOW duplication and is what let the `failed` fix land in one place instead of two — the right order of operations.
- CI at review time: every settled check green; `Test`, `Clippy`, `MSRV check (1.91)`, doc-pointer lint and the search smoke test still pending. APPROVE is conditional on those landing green — merge gates on CI, not on this verdict.
- Neither MEDIUM blocks. The first is a one-line message fix worth taking before merge; the second is a doc word.

Reviewed by `code-critic` against the spec only.

bobmatnyc:
Fixed the round-2 MEDIUM: the stale-file pointer now names `untracked_modified.first()` and only appears when that set is non-empty, so an edited seed-once (user-owned) entry is never told to run `--reset-agents` on itself. Also softened the module header's self-healing claim to scope it to binary-upgrade staleness (round-2 Promote item, doc-only). Two new tests: `deploy_summary_lines_omits_the_reset_pointer_for_user_owned_edits_only`, `deploy_summary_lines_names_the_untracked_modified_file_in_the_pointer`. Gates green: fmt, clippy -D warnings, cargo test -p trusty-mpm.

PR #4850, opened 2026-08-04
feat(trusty-mpm): enforce a per-subagent context ceiling at the PreToolUse guard

### LINKED ISSUE #4837: perf: mechanical work through LLM inference loops costs superlinearly
## Problem

trusty-mpm routes mechanical, deterministic work through LLM inference loops. Cost grows superlinearly because every tool round re-sends the agent's accumulated context, so an N-iteration loop costs far more than N times one call.

## Evidence (2026-08-04)

| Agent | Tokens | Work |
|---|---|---|
| memory-manager | 622.2k | memory_remember called once per file, 148 files, reached batch 112 |
| rust-engineer | 413.3k | running scripts/check_line_cap.sh and cargo fmt |
| rust-engineer | 255.5k | running cargo clippy |
| rust-engineer | 418.5k | Grepping gates output for failures |
| rust-engineer | 269.9k | Writing changelog fragment |

1.96M tokens across three agents. None require language models:

- Migrating 148 files into the palace is ETL — read file, derive tags from frontmatter, POST via HTTP API. Shell script or tm subcommand, zero inference.
- Running a gate and checking exit code is shell work. Inference needed only when gate FAILS and failure needs diagnosis.

**Interpretation:** The labels "Grepping gates output" and "Writing changelog fragment" are misleading. A grep is free; a changelog fragment is five lines. Those numbers measure the agent's accumulated context at the moment it reached that step. Every tool round re-sends the full history, so the LAST cheap step in a long task is the MOST expensive one. The dominant driver is **total context accumulated over an agent's lifetime**, not the cost of any individual action. Batching gates treats a symptom; architectural changes treat the cause.

## Resolution

Four approaches:

1. **Bulk memory-import path without agent.** A tm subcommand or script that walks a memory directory, maps frontmatter (name, description, metadata.type, [[wikilink]] refs) onto palace drawer fields, and writes via HTTP API. Prerequisite for [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834), whose whole point is migrating file memories into the palace — doing that through an agent produced the 622k figure.

2. **Gate-running guidance in agent base prompt.** Agents should chain gates into one batched command, read output on success only, read only failing portion on failure. BASE-AGENT.md is the natural home and is currently being edited on branch chore/instruction-prompt-dedup — sequence accordingly rather than racing it.

3. **Agent task scope is the primary cost lever.** Prefer several narrow agents with short lives over one long-running agent, even when the work is related. A running agent's scope is fixed; new work is a new agent, or it waits. Adding scope to a live agent extends its life, and every subsequent round pays the accumulated context — it feels cheaper than dispatching fresh and is the opposite. This was observed directly in the session: the coordinator added scope twice to an already-running agent (two prose rules, then a file rewrite), and that agent reached 269.9k by the time it was writing a five-line changelog file.

4. **Mechanical enforcement via guard.** Enforce token economy with a guard rather than an instruction. Precedent already exists for this reasoning — two rules were moved from prose to mechanism because instructions get ignored under task pressure:

- `pm_guard` enforces the PM's direct-action budget at the file-change level.
- [#4796](https://github.com/bobmatnyc/trusty-tools/issues/4796) denied subagent fan-out at a `PreToolUse` guard instead of instructing agents not to fan out.

Candidate controls (stated as candidates, not a design):
- Per-agent token ceiling: warns at a threshold, hard-stops at a cap, surfaced to PM for re-scoping.
- Guard firing on repeated identical commands: N consecutive duplicates or N tool calls within one task.

Either would have caught all three overruns in the evidence table regardless of what any prompt said. That is the argument for it — not that prose is worthless, but that prose alone has now been tried and observed to fail once.

## Scripting Runtime — Decided

The owner has ruled: **bash, not Python.**

Rationale: Today's mechanical work was a file loop plus curl, and one chained gate command. Neither needed a language runtime. Python was considered (uv with PEP 723 inline dependencies, which removes venv management) and rejected in favour of keeping zero new prerequisites.

Limitation (acknowledged rather than hidden): bash is weak on JSON-shaped work, so some future cases will be awkward. That is accepted.

Implication — two-tier split:
- **Recurring operations become `tm` subcommands in Rust** — bulk memory import is the worked example, already option 1 above.
- **The long tail is scripted in bash** — one-off migrations, quick audits, anything not worth a subcommand.

Constraint on any scripting agent: **write the script, run it once, report the output.** An agent that writes a script then iterates on it interactively reproduces the exact cost this issue is about.

## The Existing Rule Failed

A "gate-output economy" rule merged in [PR #4790](https://github.com/bobmatnyc/trusty-tools/pull/4790) on 2026-08-04 — the same day as the 1.96M-token overrun this issue documents. It did not prevent it.

**Open questions:** Does the rule reach BASE-AGENT.md (what subagents actually read) or only the PM prompt? Is its modality binding ("never", "do not") or advisory ("prefer", "consider")?

bobmatnyc:
Implements **option 3 only** of [#4837](https://github.com/bobmatnyc/trusty-tools/issues/4837) — mechanical enforcement. Options 1 (bulk memory import) and 2 (BASE-AGENT gate guidance) are owned elsewhere and are untouched here.

## Feasibility: tokens ARE observable — the fallback was not needed

The first question was whether tm can see subagent resource usage at all.

**Claude Code's `PreToolUse` payload carries no token counts.** The empirically-confirmed field set (captured against 2.1.220, documented in `hook_payload.rs`) is `session_id`, `transcript_path`, `cwd`, `prompt_id`, `permission_mode`, `hook_event_name`, `tool_name`, `tool_input`, `tool_use_id`, plus `agent_id` inside a subagent.

**But it carries `transcript_path`, and the counts live one hop away.** Verified against a live subagent transcript on this machine: every assistant record carries `message.usage` with `input_tokens`, `cache_creation_input_tokens`, and `cache_read_input_tokens`. Their sum is the whole prompt that turn re-sent — which *is* the accumulated context #4837 names as the dominant driver.

The decisive detail is **where** that number sits: the newest turn is at the **end** of the JSONL, so context size is recoverable from a bounded tail read in constant time, no matter how large the transcript has grown. Summing billed tokens across every turn would instead mean scanning the whole file on every tool call — paying an O(n) cost to police an O(n) cost.

So the preferred design was buildable, and **the repeated-command / tool-call-count fallback was deliberately not implemented**. Per #4837's own interpretation, tool-call count is only a proxy for accumulated context; with the driver directly measurable, a second mechanism would add surface without adding signal.

## What ships

| Piece | Role |
|---|---|
| `core::agent_cost` | Pure policy — `AgentCostConfig`, threshold classification, usage parsing, deny text |
| `commands::pm_guard_cost` | I/O half — resolves the subagent's own transcript, tail-reads it under a 200 ms timeout |
| `pm_guard` | Evaluates after the fan-out guard, ahead of Guards 1/4 |

Placement matches the [#4796](https://github.com/bobmatnyc/trusty-tools/issues/4796) precedent exactly: Guards 1 and 4 early-return ALLOW precisely when the caller **is** a subagent, which is the only case this rule fires on, so a check placed after either would be a guaranteed no-op.

Transcript resolution tries three shapes — explicit `agent_transcript_path`, a `transcript_path` already under `subagents/`, or the documented `<dir>/<parent-stem>/subagents/agent-<agent_id>.jsonl` layout — and accepts only a path that exists on disk. Claude Code has moved this field between releases, so pinning one name would silently no-op if it moves again.

## Defaults and their justification

```toml
[agent_cost]
enabled = true
warn_tokens = 250000
max_tokens  = 400000
```

**`max_tokens = 400_000`** sits just below the *smallest* overrun in #4837's evidence table (413.3k), so all three logged overruns — 413.3k, 418.5k, 622.2k — would have been stopped, not just the worst one.

**`warn_tokens = 250_000`** is ~3.5x above a measured healthy agent. A completed subagent in this workspace was measured at 71.5k of window occupancy, so ordinary work cannot trip it, and the warning still leaves ~150k of headroom (tens of tool rounds) to wrap up before the stop.

This does not break long-but-legitimate work. The threshold is context occupancy, not tool calls — a long build costs wall-clock, not context. Reaching 400k means an agent has been alive across hundreds of rounds, the condition #4837 identifies as pathological regardless of what the work was. Such a task is not blocked so much as **split**: the agent reports back and the PM re-dispatches with clean context, which is cheaper than letting it continue.

## Fails open, never closed

Every one of these allows the call: guard disabled; `max_tokens = 0`; caller is not a subagent (the PM is never evaluated and pays neither the config load nor the file read); no resolvable transcript; unreadable file or blown timeout; no usage record in the tail; a usage block with no context fields.

A false stop lands on an agent mid-task and costs a re-dispatch; a false allow merely reproduces today's behaviour. The asymmetry is deliberate and documented so no future signal inverts it.

The `Warning` level deliberately does **not** touch stdout — a `PreToolUse` exit-0 stdout may contain only a decision object, and emitting an explicit `allow` there would bypass the normal permission flow on every call an expensive agent makes. It routes to the daemon audit stream instead.

## Incidental: `MpmConfig` serde consolidation

`config.rs` sat at exactly 500 SLOC, so the new section breached the production cap. `MpmConfig`'s twelve identical per-field `#[serde(default)]` attributes are replaced by one container-level attribute — semantically identical, since the derived `Default` is field-wise — which frees 11 SLOC and leaves headroom for future sections.

## Gates — rung 5 (process lifecycle)

Run as one batched command, exit 0:

```
cargo fmt --check && cargo check -p trusty-mpm && cargo clippy -p trusty-mpm -- -D warnings && cargo test -p trusty-mpm && bash scripts/check_line_cap.sh
EXIT=0
```

- `cargo test -p trusty-mpm` — **7579 passed, 0 failed**
- 22 new tests: threshold triggering, every fail-open path, config override, transcript-path resolution, truncated-tail tolerance
- `check_line_cap.sh` — 3643 files, 0 violations

Two known-environmental flakes appeared in intermediate runs and are recorded for honesty: `execute_doctor_against_test_daemon` (daemon-bind timing, already documented as flaky on this machine) and `compose_session_instructions_display_matches_live_prompt_with_override`. Both passed in isolation and in the final clean run; the diff touches zero files under `src/client/` or the instruction pipeline.

Closes #4837


bobmatnyc:
## Verdict: BLOCK

Adversarial review, rung 5. Gates run locally on `cec325f1`: `check_line_cap.sh` pass, `cargo fmt --check` pass, `cargo clippy -p trusty-mpm --all-targets -- -D warnings` pass, `cargo test -p trusty-mpm --bin tm` → `1409 passed; 0 failed; 1 ignored`, lib target → `4477 passed; 1 failed` where the single failure is the documented environmental flake `client::executor::tests::execute_doctor_against_test_daemon` (diff touches zero files under `src/client/`). All 25 CI checks green. Nothing was `#[ignore]`d, `cfg`-gated, excluded, or deleted to reach green.

The measurement half of this PR is right. The enforcement half is not safe to ship at these defaults.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs` | 321–330 | The `Exceeded` arm denies **every** tool with no allowlist — including `SendMessage`, `Bash("git commit")`, `Bash("git push")`, `Write`, `Edit`. A stopped agent cannot durably save its work, and the deny text it is handed asserts the opposite. | Exempt the report-and-preserve path before the deny: allow `SendMessage` (reuse `pm_guard_fanout`'s existing exemption) and a narrow `git add`/`git commit`/`git push` Bash allowlist. | Fix here |
| CRITICAL | `crates/trusty-mpm/src/core/agent_cost.rs` | 211 | `stop_reason` states "SendMessage is never blocked." Under this guard it **is** blocked — the cost block at `pm_guard.rs:321` returns before any tool-specific logic. The instruction the agent is given is false, and following it produces a second denial. | Make the sentence true by fixing the allowlist above; do not weaken the sentence. | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs` | 331–340 | The warn tier never reaches the agent. `audit_agent_cost_warning` composes a message addressed to the agent ("Wrap up and report back…", line 675) and POSTs it only to `<url>/hooks`. From the agent's side there is no graduated response — it goes from silence straight to a hard wall. | Ship warn-only by default (`DEFAULT_MAX_TOKENS = 0`) until the warning has a delivery channel the agent reads; flip the stop on once #4840 lands and `BASE-AGENT.md` actually reaches running agents. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/agent_cost.rs` | 78 (`DEFAULT_MAX_TOKENS`) | 400k is not "well above any legitimate agent observed in this workspace." Measured on the 600 most recent subagent transcripts on this machine (14 days): p50 = 136k, p90 = 268k, p95 = 323k, **18 of 600 (3.0%) at or above 400k**, 77 of 600 (12.8%) at or above 250k. The stop fires on roughly 1 in 33 dispatches, not on outliers. | Either raise the default stop, or (preferred) ship it disabled per the row above. The "healthy agent = 71.5k" baseline in the module doc is a single sample and is not representative. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_cost.rs` | 44 | The 64 KiB tail window silently misses the usage record on the largest transcripts. Of the 12 largest subagent transcripts on this machine, 1 (10.3 MB) has no complete `usage` record in its last 64 KiB → `latest_context_tokens` returns `None` → fail open. Coverage degrades on exactly the transcripts this guard exists to catch. | On `None`, retry once with a larger bounded window (e.g. 1 MiB) before failing open. Still constant-cost. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs` | 325–329 | Each denial still costs a full inference round at ≥400k prompt tokens. A model that retries two or three times before giving up spends more than letting it finish would have. There is no once-only or back-off behavior. | Largely mitigated by the CRITICAL fix (give the model exactly one accepted next action). Worth a note in the module doc either way. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_cost.rs` | 62–90 | `caller_is_subagent` is true for `CLAUDE_MPM_SUB_AGENT`-marked trusty-agents subprocesses, but `resolve_agent_transcript` needs `agent_transcript_path`, a `subagents/`-parented `transcript_path`, or `agent_id` — none of which those callers carry. They are never measured. Fail-open, so safe, but the changelog implies coverage that does not exist. | Say so in the changelog fragment: native `Task`/`Agent` subagents only. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard.rs` | 322 | `MpmConfig::load_default()` does a blocking `std::fs::read_to_string` + TOML parse inside an async fn, on every subagent tool call, outside the 200 ms timeout that wraps only the tail read. A stalled `$HOME` would hang past the 5 s `PreToolUse` budget. | Move the config load inside the `EVAL_TIMEOUT` scope, or `spawn_blocking` it. | Promote |
| LOW | `crates/trusty-mpm/src/core/config.rs` | 291 | Comment says "twelve identical per-field attributes"; eleven were removed. | s/twelve/eleven/ | Fix here |

## The concrete #4841 case, answered from the code

The engineer on #4841 reached 434k and produced a merged-quality fix. Traced against this diff:

1. `caller_is_subagent` → true (native `Agent` dispatch carries `agent_id`).
2. `evaluate_agent_cost` resolves `<project>/<parent-uuid>/subagents/agent-<id>.jsonl` — I confirmed this layout against a live transcript tree, so resolution succeeds.
3. `evaluate_cost(434_000, default)` → `Exceeded`.
4. `pm_guard.rs:326-329` prints a deny and returns. **Every** subsequent tool call takes the same path.

So: it could not `git commit`, could not `git push`, could not `Write`, and could not `SendMessage`. Its edits survive on disk in the worktree, and it can still emit a final text report (a turn-ending message is not a tool call), so it is not fully mute — but the durable artifact never gets created. That is precisely the uncommitted-work-in-a-worktree loss shape this repo's worktree doctrine exists to prevent, and at a 3% firing rate it is not a corner case.

## What I verified clean

- **Measurement metric.** Last assistant `usage` = `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` is the right number and cannot be misread. Verified against live transcripts: only `assistant` records carry `message.usage`, and the three fields are always present together. Cache-read-heavy turns do not distort it — cache reads *are* re-sent prompt. Post-compaction reset is correct behavior, not a hole.
- **Tail robustness.** `.lines().rev().filter_map(parse)` skips both a truncated leading record (tail boundary) and a truncated trailing record (concurrent write), because an unparseable line is dropped rather than poisoning the scan. Pinned by `tolerates_a_truncated_leading_line`.
- **Fail-open matrix.** Independently confirmed each arm: disabled config (`agent_cost.rs:145`), `max_tokens = 0` (`:149`), non-subagent (gated at `pm_guard.rs:321`), unresolvable transcript (`pm_guard_cost.rs:87`), unreadable/timeout (`:93`), missing usage record (`:96`). `MpmConfig::load` returns `Self::default()` on both a read error and a malformed TOML (`config.rs:380-405`) — so a parsing bug cannot halt anything. The PM is genuinely never evaluated; orchestration cannot be stopped by this.
- **Transcript path.** Not meaningfully attacker-influenced: it comes from the harness payload, the effect is a bounded read, and only a token count escapes into the deny text and the loopback audit POST. No file content is echoed.
- **`config.rs` container-level `#[serde(default)]`.** Semantically identical. All eleven pre-existing fields already carried per-field `#[serde(default)]` (none was required), `MpmConfig` derives `Default` so the container default is field-wise, and `Serialize` is unaffected. No field has a non-zero `Default` that changes meaning. Confirmed by the still-passing `config_absent_yields_defaults` / `config_valid_parsed` / `config_malformed_falls_back` plus the new `config_section_parses_from_toml`.
- **Tests.** 22 new tests, all passing, none ignored or gated. The `agent_cost` suite genuinely pins the thresholds — `warns_at_the_warn_threshold` asserts 249_999 → Ok and 250_000 → Warning, so it would fail on any off-by-one.

## Required Changes

1. Exempt `SendMessage` and a `git add`/`commit`/`push` Bash allowlist from the `Exceeded` deny in `pm_guard.rs:325-330`, so a stopped agent can durably save and report. Add a test asserting `SendMessage` survives the cost guard — `pm_guard_fanout::never_denies_send_message` does not cover this path.
2. Ship warn-only: `DEFAULT_MAX_TOKENS = 0` until the warning is actually delivered to the agent (blocked on #4840). The measurement, the config section, and the warn audit are all worth landing now; the hard stop is not.
3. Retry the tail read at a larger bounded window before failing open (`pm_guard_cost.rs:44`).
4. Fix the "twelve"/"eleven" comment and scope the changelog claim to native `Task`/`Agent` subagents.

## Notes

The empirical distribution above is from this machine's transcript store only (600 most recent subagent transcripts across `~/.claude/projects` and `~/.trusty-tools/trusty-mpm/claude-config/projects`, 14-day window, measured with the same `input + cache_creation + cache_read` sum this PR uses). It is the same population the guard would police, so I weight it above the three-row evidence table in #4837 for the question of where the threshold belongs.

Findings 1 and 2 are one defect stated from two sides; fixing the allowlist resolves both. Neither is a criticism of the mechanism — enforcing at a guard rather than in prose is the right call, and the transcript-tail measurement is a genuinely good piece of engineering. The block is entirely about what happens to the agent on the other side of the wall.

bobmatnyc:
Review fixes pushed as `54da77db`. What changed, per finding:

## 🔴 BLOCK 1(a) — hard stop now defaults OFF

`DEFAULT_MAX_TOKENS` is `0` (warn-only). `warn_tokens` stays 250k. The single-sample "healthy agent = 71.5k" justification is gone from `agent_cost.rs`'s module docs, both threshold constants, the `[agent_cost]` doc in `config.rs`, and the changelog fragment — replaced everywhere by the measured distribution (600 most recent subagent transcripts, 14 days: p50 136k, p90 268k, p95 323k, 18/600 = 3.0% at or above 400k). The docs now state the reason plainly: 400k is only ~1.24x above p95, so a shipped-on stop would deny roughly one dispatch in 33 — which inverts the module's own fail-open asymmetry.

Tests: `defaults_are_warn_only`, `default_config_never_stops_any_evidence_row` (nothing up to `u64::MAX` may be denied under the shipped default), `default_config_only_warns_on_the_same_transcript` (end to end: the identical 622.2k transcript the opted-in ceiling stops merely warns under what ships). The stop still works when opted in — `stops_every_overrun_once_the_operator_opts_in`.

## 🔴 BLOCK 1(b) — escape hatch: a persistence allowlist

**Mechanism chosen: a tool allowlist, not a grace budget.** `PreToolUse` is stateless, so a grace *count* would need durable per-agent state to create, expire, and reason about on every call — and it would let the agent spend its grace on anything at all. Naming the tools makes the hatch exactly "persist and report" wide, with nothing to keep.

Permitted past the ceiling: `SendMessage`, and `Bash` running only `git add`, `commit`, `push`, `status`, `diff`. `status`/`diff` are read-only and earn their place by letting the agent write an accurate handback instead of guessing.

Three properties make the allowlist safe:
- **every** segment of a composed command must match, so `git commit -m x && cargo test` is rejected whole;
- command substitution and file-write redirection are rejected outright (`git commit -m "$(cargo build)"`, `git status > f`);
- an unresolvable segment means "not persistence" — the agent stays stopped.

`Write`/`Edit` deliberately stay denied: they are how an agent *does* work, not how it saves it (anything written is already on disk), so permitting them would make the stop vacuous.

The Bash half lives in `pm_guard_bash` and reuses that module's existing `split_shell_segments` and `shell_lex::git_subcommand` (which already sees past `git -C <path> commit`) rather than standing up a second safety parser. Tests: `command_is_persistence_only_*` (4), `escape_hatch_permits_send_message_and_git_persistence`, `escape_hatch_denies_work_tools`, `persistence_tools_cover_report_and_commit`.

## 🔴 BLOCK 2 — deny text corrected

"SendMessage is never blocked" is gone. The text now enumerates the actual hatch — `SendMessage` plus the named git commands, with "every segment of a composed command must be one of those" — and states explicitly that Write, Edit, Read and any other Bash are denied, so the agent does not burn rounds retrying them. `stop_reason_names_the_numbers_and_the_escape_hatch` asserts the old blanket claim cannot come back.

## HIGH — the warn now reaches the agent

**A path exists, and it is the one the author's reasoning was looking for:** `hookSpecificOutput.additionalContext`. Per the hooks reference (confirmed 2026-08-04) it is supported on `PreToolUse` and appears "next to the tool result", and — the load-bearing part — "Exit code 0 with no output means the hook has no decision to report, so the tool call continues through the normal permission flow… staying silent doesn't approve it." Emitting `additionalContext` **without** `permissionDecision` therefore reaches the agent while leaving the permission flow untouched. No explicit `allow` is ever emitted; `build_pretooluse_context_response_carries_no_decision` pins that.

Two follow-on details:
- **Once per agent.** The warn band is unbounded now that the stop is off, so re-emitting on every call would re-send the same ~90 tokens for the rest of the agent's life — spending context to complain about context. An atomic marker (`create_new`) claims the nudge once; I/O failure fails toward informing. Test: `warn_notice_is_claimed_once_per_agent`.
- **Deferred to the ALLOW exit.** A `PreToolUse` hook's stdout may carry exactly one object, and Guard 4's opt-in persona gate still runs after the cost block. The notice is held and emitted at the two subagent ALLOW exits, so that gate's behaviour is byte-for-byte unchanged and stdout stays single-object. The claim happens at emit time, so a notice that loses to a deny is not silently spent.

The daemon audit POST now reuses the identical `warn_reason` string rather than keeping a second copy that can drift.

## MEDIUM — tail window

Two-pass read: 64 KiB first, retried once at 1 MiB (`RETRY_TRANSCRIPT_TAIL`) only when the first window holds no complete usage record, with a short-circuit that skips the retry when the file was already read whole. Both are bounded constants, and the whole retry sits inside the same 200 ms `EVAL_TIMEOUT` so a slow disk cannot turn it into two timeouts. Tests: `retries_with_a_larger_tail_when_64k_holds_no_usage_record` (asserts the fixture genuinely defeats 64 KiB before asserting the retry recovers it) and `still_fails_open_when_even_the_larger_tail_has_no_record` (the wider window must not weaken fail-open).

## Also

Restored the doc comment on `audit_denied_tool`, which the original diff had accidentally merged into its neighbour.

## Gates

Rung 3 (localized behavior inside one crate; zero files touched under `crates/trusty-mpm/src/client/`, so the known-environmental `execute_doctor_against_test_daemon` baseline is not in play — and it passed anyway).

```
cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings && cargo test -p trusty-mpm
→ 7605 passed, 0 failed, 17 ignored
bash scripts/check_line_cap.sh   → 0 violations
bash scripts/check_sld.sh        → 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh → OK
```

Changelog fragment amended in place — no second fragment.

bobmatnyc:
## Verdict: WARN (was BLOCK)

Delta review of `54da77db` against the previously reviewed `cec325f1`. Both CRITICALs are resolved. One HIGH is new, in the allowlist that resolved them.

Gates on `54da77db`, redirected not piped: `check_line_cap.sh` && `cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → `EXIT_LINT=0`. `cargo test -p trusty-mpm --bin tm` → `test result: ok. 1420 passed; 0 failed; 1 ignored`; `--lib` → `test result: ok. 4482 passed; 0 failed; 7 ignored` (`execute_doctor_against_test_daemon` passed this run). The delta adds zero `#[ignore]`, `cfg(not(…))`, `--exclude`, or feature gates. `defaults_are_warn_only` asserts `max_tokens == 0`, which fails on the parent commit where it was `400_000`; `retries_with_a_larger_tail_when_64k_holds_no_usage_record` carries an in-test assertion that the 64 KiB window finds nothing, so it fails against the old single-pass read.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/mod.rs` | 482–532 | `command_is_persistence_only` accepts git invocations that execute arbitrary programs, so "work cannot be smuggled behind an allowed verb" does not hold. Four shapes below. | Reject `<(` alongside `$(`/backtick; allow only `-C`, `--git-dir`, `--work-tree` as git global flags (reject `-c`, `--config-env`); reject `--upload-pack`, `--receive-pack`, `--exec` anywhere in the segment. Add a test per shape. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_cost.rs` | 279–286 | `claim_warn_notice`'s key falls back from `agent_id` to `session_id`. In resolution case 2 — `transcript_path` already under `subagents/`, which this PR added and tests — there is no `agent_id`, so the key becomes the **parent** session id shared by every sibling subagent. The first sibling to cross 250k claims the marker; the rest are silently suppressed. The `"unknown"` arm is the same shape at machine scope. | Key on something per-agent by construction: hash the path `resolve_agent_transcript` already returned, and use `session_id` only as a last resort. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_cost.rs` | 288–292 | An `agent_id` whose characters all filter out leaves `key` empty, so `dir.join(key) == dir`; `create_new` on the existing directory returns `AlreadyExists` → `false` → the notice is suppressed forever. That contradicts the doc's "any I/O failure returns `true`". | `if key.is_empty() { return true; }` before the open. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_cost.rs` | 213–219 | `is_persistence_escape`'s doc rejects a grace budget because "`PreToolUse` is stateless… with nothing to keep", and `claim_warn_notice` forty lines below keeps durable per-agent state in the temp dir. The other reason given — a grace count lets the agent spend it on anything — is independent and still holds. | Drop the stateless clause; keep the scope argument. | Fix here |

### The four bypass shapes

- **`git -c diff.external='cargo test' diff`** — `-c` is in `GIT_GLOBAL_OPTS_WITH_ARG` (`shell_lex.rs:160-167`), so it is consumed with its value and the subcommand resolves to `diff`. The PR's own `command_is_persistence_only_sees_past_git_global_flags` asserts `git -c user.name=bot commit -m x` is persistence, so `-c <anything>` is accepted by construction. `diff.external` runs its program once per changed file.
- **`git -c protocol.ext.allow=always push ext::sh -c 'cargo test' HEAD`** — subcommand `push`; the `ext::` transport executes the named command.
- **`git push --receive-pack='cargo test' /tmp/repo HEAD`** — `--receive-pack` sits after the subcommand, which `git_subcommand` has already resolved to `push`. For a local-path remote git runs it locally. Same for `--upload-pack` / `--exec`.
- **`git diff <(cargo test)`** — the substitution guard tests `$(` and backtick only. `<(` is neither, and `has_file_write_redirection` scans for `>`, so it never looks at `<`. `shlex::split` yields `["git","diff","<(cargo","test)"]` → subcommand `diff` → accepted, and bash runs the substitution. (`>(…)` is caught incidentally by the `>` scan; `<(` is not.)

Not flagged: `git -c core.pager=…` does not fire, because git only pages on a TTY and the Bash tool pipes stdout. `sudo git commit` resolves and is permitted, but it is still a commit.

**Why HIGH and not CRITICAL.** The stop ships OFF, and what a bypass evades is a token budget, not a privilege boundary — the agent held unrestricted Bash before it crossed the ceiling, so nothing is escalated. What is broken is the documented invariant, in a parser this PR itself frames as safety-critical, and a wrong claim there invites future reliance. Calling it CRITICAL would be inflation.

## What I verified cleared

**BLOCK 1(a) — warn-only.** `DEFAULT_MAX_TOKENS = 0`. `default_config_never_stops_any_evidence_row` runs the evidence rows *plus* `u64::MAX` through the shipped default and asserts every one is `Warning`, so no input can produce a deny under the default. The measured distribution replaced the 71.5k single sample in the module doc, both constants, the `[agent_cost]` config doc, and the changelog. `default_config_only_warns_on_the_same_transcript` proves it end-to-end on the identical 622.2k fixture the opted-in ceiling stops.

**BLOCK 1(b) — escape hatch shape.** `SendMessage` unconditional; `Write`/`Edit` stay denied and the reasoning is right — anything already written is already on disk, so permitting them would make the stop decorative. I traced `split_shell_segments` rather than trusting the claim: it cuts on `&&`, `||`, `;`, `|`, newline, and bare `&` (excluding fd-dup and trailing background), is quote-aware, and on unbalanced quotes falls back to over-splitting, which is the safe direction — an unparseable segment then fails `shlex::split` and returns `None`, and unresolvable correctly means "not persistence". So `git commit -m x && cargo test`, `git status; curl …`, `git commit -m x | tee log`, and newline-separated composition all fail the whole command as claimed. The gap is entirely the flag surface *inside* an already-accepted git segment — finding 1, not the segment logic.

**BLOCK 2 — deny text.** Accurate against what the allowlist permits, modulo finding 1. It enumerates the five subcommands from `PERSISTENCE_GIT_SUBCOMMANDS` itself so the text cannot drift from the policy, names `Write, Edit` as closed, and `stop_reason_names_the_numbers_and_the_escape_hatch` asserts `!reason.contains("never blocked")`, so the blanket claim cannot come back.

**HIGH — the channel to the agent.** Verified against the hooks reference rather than the PR's paraphrase: `additionalContext` is a valid `PreToolUse` `hookSpecificOutput` field and is delivered "next to the tool result", and JSON present without `permissionDecision` leaves the call in the normal permission flow. `git grep permissionDecision` over the branch shows `"deny"` is the only value ever emitted — no `allow` on any path, which was the property the first cut declined to give up. The deferral is right too: `claim_warn_notice` is called inside `emit_cost_notice`, at the exit, so a notice that loses the race to Guard 4's persona deny is not spent, and stdout still carries exactly one object with that gate unchanged. Both subagent exits are covered — env marker → Guard 1, `agent_id` → Guard 4 — so there is no path where a notice is computed and then dropped.

**On the stateless/once-per-agent tension.** The state is real and durable: an atomic `create_new` marker file per key under `$TMPDIR/trusty-mpm-agent-cost/`. Concurrent calls from one agent cannot both win it. It survives for the agent's life and is reaped with the temp dir, whose only cost is one extra nudge. So the mechanism works; what fails is the *key*, in the two ways above. The stateless argument is what no longer holds — a doc fix, not a design fault.

**MEDIUM — tail retry.** Verified on the actual file that defeated 64 KiB in my first review (`agent-ab1485bf54695372e.jsonl`, 10.0 MiB): 64 KiB → no record, 1 MiB → 130,153 tokens. Both passes sit inside one `EVAL_TIMEOUT` because the wrapper moved to `read_latest_context`, so a slow disk cannot become two timeouts, and the `(jsonl.len() as u64) < bytes` short-circuit correctly skips the second pass on a file the first read already covered whole.

## Required Changes

1. Close the four bypass shapes in `command_is_persistence_only`, with a test each.
2. Key `claim_warn_notice` per agent, not per session, and return `true` on an empty key.
3. Drop the "`PreToolUse` is stateless" clause from `is_persistence_escape`'s doc.

## Notes

None of this blocks merge on its own reading — the shipped default is warn-only, so the allowlist is dormant until an operator sets `agent_cost.max_tokens`. That is exactly why WARN rather than BLOCK: the defect is real and the claim is wrong, but nothing reaches a user until someone opts in. Fixing it before the opt-in becomes documented is cheap; fixing it after someone relies on the claim is not.

bobmatnyc:
## Review findings closed — flag surface, key derivation, docs (`c6c6657f`)

### HIGH — the persistence allowlist leaks

The classifier resolved the git subcommand and stopped reading. It now judges the flags, and moved to its own module (`crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs`) — `pm_guard_bash/mod.rs` was at 399 of its 500 SLOC.

Four rules. Three are default-deny, so a construct nobody thought of is rejected rather than passed:

1. **The program must be `git` itself.** No env-assignment prefix, no `sudo`, no `env`. This also closes `GIT_SSH_COMMAND='cargo test' git push`, which was in the same class and was not reported.
2. **Global options are default-denied.** Only `-C`, `--git-dir`, `--work-tree`, `--no-pager`, `--no-optional-locks`, `--literal-pathspecs` survive. `-c` is rejected outright rather than having its values allow-listed — the set of config keys reaching an exec (`diff.external`, `core.gitProxy`, `credential.helper`, `protocol.ext.allow`, `alias.*`, `sshCommand`, …) is open-ended enough that enumerating the safe ones is the losing side of the bet, and an agent persisting work does not need `-c`.
3. **No live shell metacharacter.** `$` and backtick anywhere single quotes do not suppress them; `(`, `)`, `<`, `>` anywhere unquoted; unbalanced quotes anywhere. One rule instead of a scan for particular spellings. The two byte sets carry different quoting rules on purpose — a double quote does not suppress substitution (`git commit -m "$(cargo build)"` must reject) but does suppress redirection (`git commit -m "spec -> code"` must pass); collapsing them was a real bug I shipped and the existing suite caught.
4. **Arguments may not name a program.** Any `::` token is a remote-helper transport, and `--exec`/`--exec-path`/`--ext-diff`/`--textconv`/`--output` plus every `--*-pack` (suffix rule, not a literal list) are rejected wherever they appear.

Rule 4 is the one enumerated set. That is deliberate: a false reject on the post-subcommand surface strands the work this hatch exists to save, so ordinary flags (`-m`, `--amend`, `--force-with-lease`, `-u`) stay permissive and only the families that take a program or an output file are named.

`shell_lex::git_subcommand` is untouched. It serves the deny side, where skipping an unrecognised global option is the safe direction; here it is the unsafe one, and reusing it is exactly how `-c diff.external=…` got through. Two callers, two opposite notions of safe-on-ambiguity.

**Verdicts, per reported shape:**

| Command | Before | Now | Closed by |
|---|---|---|---|
| `git -c diff.external='cargo test' diff` | persistence | **rejected** | rule 2 |
| `git -c protocol.ext.allow=always push ext::sh -c 'cargo test' HEAD` | persistence | **rejected** | rules 2 and 4 |
| `git push --receive-pack='cargo test' /tmp/repo HEAD` | persistence | **rejected** | rule 4 |
| `git diff <(cargo test)` | persistence | **rejected** | rule 3 |

Same class, also now rejected: `git --config-env=…`, `git --exec-path=/tmp/evil status`, `git --totally-made-up-global status`, `git push --future-pack=cargo …`, `git diff --ext-diff`, `git diff --output=/tmp/out.txt`, `git push ext::cargo origin HEAD`, `sudo git commit -m x`, `GIT_EXTERNAL_DIFF=cargo git diff`, `git add -A < /tmp/list`, `git diff >(cargo test)`.

Still persistence, so the hatch stays usable: `git -C /repo/wt commit -m x`, `git --git-dir=… --work-tree=… add -A`, `git --no-pager diff --stat`, `git push origin HEAD:main`, `git add -A && git commit -m x && git push`, `/usr/bin/git commit -m x`, and commit messages carrying quoted metacharacters (`git commit -m 'fix: handle (n>1) and $HOME'`).

**Two tests changed rather than kept:**

- `git -c user.name=bot commit -m x` was asserted to BE persistence, which pinned the hole open by construction — any `-c <anything>` was accepted. It now asserts rejection, in `command_is_persistence_only_rejects_config_injection`.
- `git status 2>/dev/null` was allowed on the strength of `has_file_write_redirection` telling a `/dev/null` write from a real one. That classifier is no longer in the trust path; keeping it meant keeping a hand-rolled redirection parser that had already missed `<`. Every unquoted redirection metacharacter now disqualifies, and an agent saving its work never needs one.

Seven new tests, one per bypass shape plus the quote-awareness and bare-program cases.

**Residual, documented in the module rather than hidden:** git reads exec-capable settings from the repository's own config and attributes (`diff.external` via the default `--ext-diff`, `textconv`), so an agent that wrote `.git/config` before hitting the ceiling can still reach an exec through a bare `git diff`. Rule 2 closes the argument-supplied route; closing the on-disk one means reading and judging repo config inside a `PreToolUse` hook, which is neither fast nor side-effect-free. Same posture as `evaluate_worktree_add_command`'s existing bypass notes.

### MEDIUM — the once-per-agent key collided across siblings

`agent_id` → `session_id` has no per-agent term in resolution case 2, where `transcript_path` already points under `subagents/` and there is no `agent_id`. Every sibling of one parent got the parent's session id, so the first to warn silenced the rest.

The key is now `agent_id`, else the resolved subagent transcript's **file stem** (`agent-<id>` — the identity `resolve_agent_transcript` just established), else `session_id`. `warn_notice_is_claimed_per_sibling_not_per_parent_session` puts two siblings under one parent with a shared `session_id` and asserts each gets its own notice and only one each.

### LOW — an empty key suppressed permanently

`dir.join("")` is `dir`, and `create_new` on the existing directory returns `AlreadyExists` → `false` → the notice suppressed forever, for every agent at once. That is the inverse of the documented "any I/O failure returns `true`". `warn_notice_key` now returns `None` when nothing derivable remains and the caller fails open. `warn_notice_fails_open_when_no_key_can_be_derived` asserts three degenerate payloads keep informing across repeated calls.

### Docs

`is_persistence_escape` no longer argues the allowlist from `PreToolUse` being stateless — `claim_warn_notice`'s marker file is state, so that reason was false. The independent one now stated: a grace *count* is spendable on anything the agent likes, while naming the tools makes the hatch exactly as wide as "persist and report". `PERSISTENCE_GIT_SUBCOMMANDS` records that its "none of them can produce new work" claim holds only because the flags are judged too.

### Gates (rung 3, crate-scoped)

```
cargo clippy -p trusty-mpm --all-targets -- -D warnings — clean
cargo test -p trusty-mpm --bin tm  — test result: ok. 1429 passed; 0 failed; 1 ignored
cargo test -p trusty-mpm (lib)     — test result: FAILED. 4481 passed; 1 failed; 7 ignored
bash scripts/check_line_cap.sh          — 0 violations
bash scripts/check_sld.sh               — 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh — OK
```

The one lib failure is the documented environmental baseline (`docs/reference/test-ladder-baseline.md:19`):

```
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:56098/api/v1/doctor?project=…)")
```

A connection error, not an assertion, and this diff touches zero files under `crates/trusty-mpm/src/client/`.

One other test, `tests_behavior_b::guided_fallback_non_git_dir_with_managed_env_settles_quickly_returns_promptly`, failed twice under full-suite parallel load on a wall-clock assertion (`took 363.8ms`), then passed 8/8 in isolation and passed in a clean full run of its own target. Load-induced; the branch does not touch that file. Reporting it rather than filing — it has no open issue and one clean full run is not enough to characterise it.

Changelog fragment amended in place; no second fragment, no version bump.

bobmatnyc:
## Verdict: WARN (was WARN)

Delta review of `c6c6657f` against `54da77db`. I did not read the classifier and reason about it — I ran candidate bypasses through `command_is_persistence_only` in a scratch test, then ran the surviving ones through real git (2.54.0) to confirm they execute. Rule 3 (metacharacters) held under everything I threw at it. Rule 4 did not: git accepts unambiguous **prefixes** of its long options, and the deny list matches by exact name.

Gates in a fresh worktree at `c6c6657f`, redirected not piped: `cargo test -p trusty-mpm --bin tm` → `1428 passed; 1 failed` on the first run and `1427 passed; 2 failed` on the second. Both failures are in `crates/trusty-mpm/src/bin/tm/tests_behavior_b_tests.rs`, which this PR does not touch (`git diff --name-only origin/main...c6c6657f` lists six files, none of them that one), and one is a wall-clock assertion — `a record that settles on the second poll must not pay the full retry budget; took 347.00525ms`. Both pass in isolation (`test result: ok. 1 passed` each) and CI's 17m39s `Test` job is green, along with all 26 checks. Load-induced on this machine, not this diff. Targeted evidence for what actually changed: `--bin tm pm_guard` → `162 passed; 0 failed`, `--bin tm persistence` → `11 passed; 0 failed`, `-p trusty-mpm agent_cost` (lib) → `18 passed; 0 failed`.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs` | 139–147, 297 | `EXEC_OPTION_DENY` is matched by **exact name**, but git's `parse-options` accepts any unambiguous prefix — so `git push --exe=cargo` and `git push --rece=cargo` are `--exec` / `--receive-pack`, and both classify as persistence and execute. `--receive-pack` is not even in the list; it relies on the `-pack` suffix rule, which an abbreviation removes. | Match by prefix, and list the full names: add `"--receive-pack"`/`"--upload-pack"`, then `!EXEC_OPTION_DENY.iter().any(\|d\| name.len() > 2 && d.starts_with(name))` alongside the existing suffix rule. Test each of the four strings below. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs` | 247 | Rule 2 is documented as "the program must be `git` itself" but implemented as a **basename** check: `program.rsplit('/').next() != "git"`. `./git`, `/tmp/evil/git`, and `../../tmp/evil/git` all pass. | Either require the token to be exactly `git` or an absolute path under a system bin dir, or move this into the residual paragraph — but do not leave the doc claiming a check the code does not make. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs` | 46–53 | The residual note scopes the on-disk exec route to "the repository's own config", but rule 3 permits `-C` and `--git-dir`, so the config can live at any path the agent chose. | s/the repository's own config/the config of any repository it can reach, including one named by `-C` or `--git-dir`/ | Fix here |
| LOW | `crates/trusty-mpm/src/core/agent_cost.rs` | 281–296 | `stop_reason` enumerates the permitted subcommands but says nothing about the metacharacter rule, so `git commit -m "fix $ISSUE"` and `git status 2>/dev/null` are denied with no hint why. This is the retry-burn the function's own doc exists to prevent. | Add one clause: no `$`, backticks, parens, `<`, or `>` outside single quotes — single-quote a message that needs them. | Fix here |

### The bypass, executed

Classifier verdicts, printed from a scratch `#[test]` calling `command_is_persistence_only` directly:

```
PERSIST  <<<  "git push --rece=cargo /tmp/r HEAD"
PERSIST  <<<  "git push --receiv=cargo /tmp/r HEAD"
PERSIST  <<<  "git push --exe=cargo /tmp/r HEAD"
PERSIST  <<<  "git push --exe cargo /tmp/r HEAD"
reject   <<<  "git push --receive-pack=cargo /tmp/r HEAD"
```

The last line is the shape the previous round closed. The four above it are the same option, spelled short.

Real git, against a local bare remote, with `pwn.sh` standing in for `cargo test`:

```
=== test 2: ABBREVIATED --rece= ===
To /tmp/critic4850/remote
 * [new branch]      HEAD -> x2
=== test 3: ABBREVIATED --exe= ===
To /tmp/critic4850/remote
 * [new branch]      HEAD -> x3
=== separate-value --rece ===
 * [new branch]      HEAD -> x4
=== pwned.txt ===
PWNED: arbitrary program executed, args: /tmp/critic4850/remote
PWNED: arbitrary program executed, args: /tmp/critic4850/remote
PWNED: arbitrary program executed, args: /tmp/critic4850/remote
PWNED: arbitrary program executed, args: /tmp/critic4850/remote
```

Four executions, one per push. `git version 2.54.0`.

Not exploitable, checked and cleared: `git diff --ext`, `--ext-di`, `--textc`, `--outp=/tmp/x` all classify as PERSIST, but `git diff` uses the diff-opt parser, which does **not** abbreviate — real git answers `error: invalid option: --ext`. So the abbreviation class is confined to `git push`'s two exec options. Fixing those two closes it.

### The basename check, executed

```
PERSIST  <<<  "./git commit -m x"
PERSIST  <<<  "/tmp/evil/git commit -m x"
PERSIST  <<<  "../../tmp/evil/git push origin HEAD"
reject   <<<  "command git commit -m x"
reject   <<<  "exec git commit -m x"
reject   <<<  "nohup git commit -m x"
```

```
=== running: /tmp/critic4850/planted/git commit -m x   (classifier says PERSIST) ===
PWNED-PLANTED: basename-git ran, argv: commit -m x
```

This needs a pre-ceiling write plus `chmod +x`, which puts it in the same class as the `.git/config` residual the module already documents — that is why it is MEDIUM, not HIGH. What is wrong is the claim: rule 2's stated invariant is "the program must be `git` itself," and a file the agent named `git` is not that.

### Rule 3's allowed globals do reach an exec — as documented

```
PERSIST  <<<  "git -C /tmp/evil diff"
PERSIST  <<<  "git --git-dir=/tmp/evil/.git --work-tree=/tmp status"
```

```
=== running: git -C /tmp/critic4850/evil diff   (classifier says PERSIST) ===
=== pwned2.txt ===
PWNED-VIA-C: external diff ran, argv: f /var/folders/.../f 5626abf... 100644 f 0000000 100644
```

Same pre-ceiling-write requirement, so the class is unchanged and I am not raising it above the residual note. The note's wording is what needs the one-line fix (finding 3).

## What held under execution

**Rule 3 — the quoting.** This is where the implementer had already shipped one bug (collapsing the two byte sets), so I attacked it hardest. Every probe landed on the correct side:

```
reject   <<<  "git commit -m \"$(cargo build)\""      # $ live inside double quotes
reject   <<<  "git commit -m $'\\x24(cargo test)'"    # ANSI-C quoting
reject   <<<  "git commit -m \"a\\\"$(cargo test)\\\"\""  # escaped-quote nesting
reject   <<<  "git commit -m spec\\>code"             # backslash-escaped, over-rejects
reject   <<<  "git commit -m \"fix $ISSUE\""          # bare variable, over-rejects
reject   <<<  "git status 2>/dev/null"
reject   <<<  "git diff <(cargo test)"
reject   <<<  "git commit -m x\ncargo test"           # newline composition
reject   <<<  "git commit -m x; cargo test"
PERSIST  <<<  "git commit -m 'costs $5'"              # single quotes suppress
PERSIST  <<<  "git commit -m \"spec -> code\""        # double quotes suppress redirection
PERSIST  <<<  "git commit -m x --author=\"A <a@b.com>\""
PERSIST  <<<  "git commit -m 'it'\\''s fine'"         # the classic quote-splice
PERSIST  <<<  "git commit -m 'fix: handle (n>1) and $HOME'"
```

The asymmetry the implementer describes is real and correct in `QuoteScan`: `allows_substitution` is `!= Single` so a `$` survives double quotes and is rejected; `is_unquoted` is `== None` so `>` and `(` inside either quote are literal and pass. Every error I could construct fell on the over-reject side. I found no way to get a live `$`, backtick, `(`, `<`, or `>` past it.

**Rule 1 — composition.** `&&`, `;`, `|`, and a literal newline all fail the whole command, as the previous round verified. Unchanged here.

**Key derivation.** Fail-open is correct, and it is not the fail-open shape this repo has been bitten by. `claim_warn_notice` gates an *informational* `additionalContext` emission (`pm_guard.rs:699-705`), not a permission decision — failing open re-sends ~90 tokens, it cannot open a gate or advance any cursor. And the underivable branch is effectively unreachable on the live path: `emit_cost_notice` only runs when `notice` is `Some`, which requires a `Warning` verdict, which requires `resolve_agent_transcript` to have already succeeded — so `warn_notice_key`'s middle branch always yields a stem. An attacker cannot force it; a payload that defeats key derivation also defeats cost evaluation, so nothing is measured and nothing is emitted. The sibling-collision fix is right: keying on the resolved transcript's stem distinguishes `agent-a`/`agent-b` under one parent, which `session_id` could not.

**The two changed tests, checked for the right reason.** `git -c user.name=bot commit -m x` flipping from asserted-persistence to asserted-reject is principled: `-c` genuinely reaches an exec through `diff.external`, `credential.helper`, `core.gitProxy`, and `protocol.ext.allow`, and the old assertion pinned `-c <anything>` open by construction. Dropping `git status 2>/dev/null` is also principled — it was allowed on the strength of `has_file_write_redirection`, a hand-rolled redirection parser that already did not know about `<`, and removing that from the trust path is the whole point of rule 3. Neither is a red-to-green edit: both moved into reject lists with a named reason, and the positive coverage they vacated was replaced (`git --work-tree /repo -C /repo status`, `git --no-pager diff --stat`). No `#[ignore]`, `cfg`-gate, `--exclude`, or deletion anywhere in the delta.

**Is `git status 2>/dev/null` a real usability regression?** No work is stranded — the plain form works, and no persistence operation needs a redirection. The cost is retries: the deny text does not mention the metacharacter rule, so an agent that reflexively writes `2>&1` or `-m "fix $ISSUE"` gets a bare deny and guesses. That is finding 4, and it is one clause of text.

## Required Changes

1. Match `EXEC_OPTION_DENY` by prefix and add `--receive-pack`/`--upload-pack` explicitly, with a test for `--exe=`, `--exe <val>`, `--rece=`, `--receiv=`.
2. Make rule 2's program check match its documented claim, or restate the claim.
3. Widen the residual note to cover `-C`/`--git-dir`.
4. Name the metacharacter rule in `stop_reason`.

## Notes

Severity calibration is unchanged from the last round and for the same reason: `DEFAULT_MAX_TOKENS` is `0`, so the allowlist is dormant until an operator sets `agent_cost.max_tokens`, and what a bypass evades is a token budget rather than a privilege boundary — the agent held unrestricted Bash before it crossed the ceiling. HIGH, not CRITICAL.

One caution on the prefix fix: `"--textconv".starts_with("--text")` is true, so a prefix rule also rejects `git diff --text`. That is a harmless over-reject on a read-only formatting flag, but it should be a deliberate choice rather than a surprise.

bobmatnyc:
## #4850 second-review findings — all four fixed, pushed as `53b71d10`

### HIGH — git option abbreviation defeated `EXEC_OPTION_DENY`

`parse-options` resolves any unambiguous **prefix** of a long option, so `--exe` *is* `--exec` and `--rece` *is* `--receive-pack`. The deny list matched names exactly. `--receive-pack` was not even on it — it rode the `-pack` **suffix** rule, and an abbreviation strips the suffix off entirely, so no suffix rule could ever have covered it.

Reproduced independently on git 2.54.0 against a real bare remote before touching the code:

```
$ git --version
git version 2.54.0
--- git push --rece=…/pwn.sh ---
PWNED: arbitrary program executed
--- git push --exe=…/pwn.sh ---
PWNED: arbitrary program executed
--- git push --receive-pack=…/pwn.sh ---
PWNED: arbitrary program executed
```

**Fix — the option surface inverted to default-deny**, not two more entries in the list. Adding `--exe` and `--rece` would have been the third round of the same mistake, and the abbreviation set is unbounded (every proper prefix of every exec-capable name, present and future). Post-subcommand long options are now judged against `SAFE_LONG_OPTS` — the same shape rule 3 already had for globals. The abbreviation question stops needing an answer: an abbreviation of a dangerous option is absent from the safe list and denied; an abbreviation of a *safe* option is denied too, a false reject in the harmless direction. Whatever the next unrecognised construct is, the answer here is already "denied".

The original reasoning against default-deny was "a false reject strands the agent's work". That over-weighted the cost. This hatch is not a general git CLI — it exists so a stopped agent can save work, and `git add` / `git commit -m` / `git push` in their plain spellings always work. An exotic flag being rejected costs one retry with the plain spelling, not the work. Short options and positionals stay permissive: a single-dash token is not abbreviatable, so there is no unrecognised-prefix class there.

Classifier output, same strings, before and after (`command_is_persistence_only`, raw from a `--nocapture` run):

```
BEFORE (c6c6657f)                                AFTER (53b71d10)
PERSIST  <<<  "git push --rece=cargo /tmp/r HEAD"     reject   <<<  "git push --rece=cargo /tmp/r HEAD"
PERSIST  <<<  "git push --exe=cargo /tmp/r HEAD"      reject   <<<  "git push --exe=cargo /tmp/r HEAD"
PERSIST  <<<  "git push --exe cargo /tmp/r HEAD"      reject   <<<  "git push --exe cargo /tmp/r HEAD"
reject   <<<  "git push --receive-pack=cargo …"       reject   <<<  "git push --receive-pack=cargo …"
PERSIST  <<<  "git push --upl cargo /tmp/r HEAD"      reject   <<<  "git push --upl cargo /tmp/r HEAD"
PERSIST  <<<  "git diff --ext"                        reject   <<<  "git diff --ext"
PERSIST  <<<  "git diff --textc"                      reject   <<<  "git diff --textc"
PERSIST  <<<  "git diff --outp=/tmp/out.txt"          reject   <<<  "git diff --outp=/tmp/out.txt"
PERSIST  <<<  "./git commit -m x"                     reject   <<<  "./git commit -m x"
PERSIST  <<<  "/tmp/evil/git push origin HEAD"        reject   <<<  "/tmp/evil/git push origin HEAD"
PERSIST  <<<  "../../tmp/evil/git status"             reject   <<<  "../../tmp/evil/git status"
PERSIST  <<<  "git add -A"                            PERSIST  <<<  "git add -A"
PERSIST  <<<  "git commit -m 'fix: thing' --amend"    PERSIST  <<<  "git commit -m 'fix: thing' --amend"
PERSIST  <<<  "git push -u origin HEAD --force-…"     PERSIST  <<<  "git push -u origin HEAD --force-…"
PERSIST  <<<  "git status --short --branch"           PERSIST  <<<  "git status --short --branch"
PERSIST  <<<  "git diff --stat --cached"              PERSIST  <<<  "git diff --stat --cached"
```

### MEDIUM — rule 2 was a basename check, not a program check

**Fixed the check, not the claim.** `program.rsplit('/').next() != "git"` verified the file's *name*; it now requires the literal token `git`. Resolving the path instead would mean stat-ing from inside a `PreToolUse` hook and *still* could not tell a planted `git` on `PATH` from the real one — so the check now matches what it can actually prove, and `PATH` is documented as residual alongside the on-disk config route. An in-command `PATH=… git push` was already rejected (the program token is the assignment, not `git`).

Trade: `/usr/bin/git commit -m x` is now rejected too. The old test asserted it passed; it now asserts it is rejected, with the reason inline. Bare `git` is always available as the fallback spelling.

### LOW 1 — residual note was wrong

The note scoped the on-disk exec route to "the repository's own config". `-C`, `--git-dir`, and `--work-tree` are allowed globals, so the repository is attacker-choosable *inside the same command* — `git -C /tmp/evil diff` runs the `diff.external` from `/tmp/evil/.git/config`. Note corrected, and `command_is_persistence_only_admits_the_attacker_chosen_repo_residual` pins the residual deliberately: if someone later closes it, that test fails and the note goes with it.

### LOW 2 — `stop_reason` omitted the metacharacter rule

The deny text named `git commit` as permitted while the classifier silently rejects every unquoted `$`, backtick, `(`, `)`, `<`, `>`. It now names the rule *and* the fix ("write `git commit -m 'literal text'`, never `-m \"fix \$ISSUE\"`"), plus the flag classes that survive the new default-deny surface, so a rejected command costs one retry instead of a guessing loop.

### Tests

New: `command_is_persistence_only_rejects_abbreviated_exec_options`, `command_is_persistence_only_accepts_the_flags_agents_actually_use`, `command_is_persistence_only_admits_the_attacker_chosen_repo_residual`, `stop_reason_names_the_metacharacter_rule`. Extended: `command_is_persistence_only_requires_a_bare_git_program`, `command_is_persistence_only_rejects_exec_options`.

Each was run against the pre-fix code first and failed there:

```
expected "git push --rece=cargo /tmp/r HEAD" to be rejected: an abbreviation IS the option
expected "./git commit -m x" to be rejected: the program must be git itself
the deny text must name the rule that rejects `-m "fix $ISSUE"`
```

### Gates

```
cargo test -p trusty-mpm --bin tm    test result: ok. 1432 passed; 0 failed; 1 ignored
cargo test -p trusty-mpm --lib       test result: ok. 4483 passed; 0 failed; 7 ignored
bash scripts/check_line_cap.sh       3646 tracked .rs file(s); 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh            55 spec doc(s) + 3059 code file(s); 0 error(s), 0 warning(s)
```

`--lib` was added to the listed gate set because `stop_reason` lives in `core::agent_cost`, which the `--bin tm` suite does not cover.

Left alone per the review: the key derivation, the two previously-changed tests, and rule 3's metacharacter handling.

bobmatnyc:
## Verdict: APPROVE — round 4 (`c6c6657f` → `53b71d10`, delta only)

Zero CRITICAL, zero HIGH. The HIGH from round 3 is closed, and I could not find a
replacement for it. This round converged; it is not spinning.

### What I executed

Real git 2.54.0, a bare remote, a planted marker program, ~60 probes. Every command
below was run, not reasoned about.

**The round-3 exploit is dead.** `git push --exec=<pwn>`, `--exe=<pwn>`, `--rece=<pwn>`
all still execute the named program against real git (marker fired on all three) and all
three are now DENY.

**`SAFE_LONG_OPTS` audited entry by entry, mechanically.** I pulled the real option table
for each of the five subcommands (`git <sub> --git-completion-helper`) and checked all 104
entries for the failure mode that would reopen the hole: a listed name that is *not* an
exact option but *is* a unique prefix of a dangerous one, which git would then resolve for
you. Three prefix resolutions exist and none is dangerous:

```
git commit: --stat   -> --status
git push:   --no-branch -> --no-branches
git diff:   --short  -> --shortstat
```

No entry reaches an exec, a file write, or a network fetch by its value. `--gpg-sign` takes
a key id, `--push-option` a server string, `--repo` a repository (`::` catches `ext::`),
`--file`/`--fixup` read.

**The load-bearing "short options are not abbreviatable" claim holds.** git rejects every
single-dash long form:

```
git push -exec=<pwn> …        -> error: unknown switch `e'          [no exec]
git push -rece=<pwn> …        -> error: unknown switch `r'          [no exec]
git diff -ext-diff            -> did you mean `--ext-diff` (with two dashes)?
```

I read every short option on all five subcommands from `-h`. None names a program.
`git push -o <pwn>` is `--push-option` (no exec). `git diff -O<pwn>` is an orderfile (read).

**Positionals cannot reach an exec at argv level.** `git push pwn://host/x HEAD` needs
`git-remote-pwn` on `PATH` — it did not resolve from cwd (PATH residual, documented). ssh
URL option injection is blocked by git itself:
`git push 'ssh://-oProxyCommand=<pwn>/x'` → `fatal: strange hostname '-oProxyCommand=' blocked`.

**Global-option walk desyncs are dead ends.** The guard ALLOWs `git -C --exec-path=<dir> status`
and `git --git-dir --exec-path=<dir> status` because it consumes the token as the option's
value. Real git dies (`cannot change to '--exec-path=…'`, `not a git repository`) — no exec.
git.c does no prefix matching and no attached-value forms, so the global surface is exact:
`--git-di=.`, `--exec-pa=…`, `-cfoo.bar=baz`, `-C<path>` are all `unknown option`.

**Default-deny does not strand real work.** All of these classify as persistence:
`git add -A`, `git add -A --force`, `git commit -m 'fix: thing'`, `git commit --amend --no-edit`,
`git push -u origin HEAD`, `git push --force-with-lease origin HEAD`, `git status --short`,
`git diff --stat --cached --name-only`, `git -C <path> commit -m x`, `git commit -am 'save'`,
`git add -A && git commit -m x && git push -u origin HEAD`, `git push origin HEAD:refs/heads/topic`.

`git rebase --continue` and `git worktree add` are NOT persistence — but they never were:
`PERSISTENCE_GIT_SUBCOMMANDS` is unchanged in this delta. Not a regression.

**Rejecting `/usr/bin/git` is safe.** No agent brief, doc, or script in the repo invokes git
by absolute path; the only hits are the guard's own tests. The deny-side parser
(`shell_lex::git_subcommand`) still resolves `/usr/bin/git …`, which is the correct
asymmetry — permissive on the deny side, strict on the allow side.

### Tests spot-checked fail-before

Reverted each hunk in a detached tree and confirmed the test fails:

- `command_is_persistence_only_requires_a_bare_git_program` — restored the `program.rsplit('/')`
  basename check → `panicked … expected "./git commit -m x" to be rejected`. FAILED.
- `command_is_persistence_only_rejects_abbreviated_exec_options` — restored the exact-match
  `EXEC_OPTION_DENY` list → `panicked … expected "git push --rece=cargo /tmp/r HEAD" to be
  rejected`. FAILED.

Both pass on `53b71d10`.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/agent_cost.rs` | 293–303 | The rewritten `stop_reason` enumerates the restrictions — metacharacters, "exotic long flags" — but omits the `::` rule, which rejects ordinary Rust commit messages. Executed: `git commit -m 'refactor: use std::process::Command'` → DENY; `git commit -m 'fix: Option::unwrap on None'` → DENY. In a Rust workspace that is a routine message, and the agent gets the hintless deny that LOW 2 was added to prevent. | Name `::` next to the metacharacter list in `stop_reason`, with the fix ("no `::` in a commit message"). Narrowing `argument_is_inert` to skip `-m`/`--message=` values is the better fix but adds parser state — the text fix is free. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs` | 65–76 | The residual note lists `diff.external`, `textconv`, `core.editor` and the `-C`-chosen repo, but omits git hooks — the easiest on-disk exec route, needing no config edit and no `-C`. Executed: planted `.git/hooks/pre-commit`, ran `git commit -am hooked` in the agent's own repo → hook fired, classifier ALLOWs. Also `core.fsmonitor` makes `git -C <evil> status` an exec route (executed, fired). Same unclosable class the note already admits; the instance list is just short. | Add hooks (`pre-commit`/`commit-msg`/`pre-push`, `core.hooksPath`) and `core.fsmonitor` to the residual sentence. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/pm_guard_bash/persistence.rs` | 61–63, 456 | Short options stay permissive on an enumerate-the-danger assumption — the exact shape rule 4 just abandoned. True on git 2.54 (verified), but it is a bet on git's future option table, and the two surfaces are already asymmetric: `--template=<f>` is DENIED while `-t <f>` is ALLOWED (probed), same for `--edit`/`-e` and `--reedit-message`/`-c`. Neither reaches an argv-named program, so this is an observation, not an exploit. | A `SAFE_SHORT_OPTS` allowlist would close it, at a real false-reject cost. Separable from this PR. | Promote |

### Gates

```
cargo test -p trusty-mpm --bin tm  → test result: ok. 1432 passed; 0 failed; 1 ignored    EXIT=0
cargo test -p trusty-mpm --lib     → test result: ok. 4483 passed; 0 failed; 7 ignored    EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings                                   EXIT=0
cargo fmt --check                                                                          EXIT=0
bash scripts/check_line_cap.sh                                                             EXIT=0
```

Changelog fragment present: `crates/trusty-mpm/changelog.d/4850-persistence-hatch-option-abbreviation.md`.

Issue #4836, opened 2026-08-04
memory_recall and memory_recall_deep return identical results regardless of query (scores always 1.0)


bobmatnyc:
## Defect

`mcp__trusty-memory__memory_recall` and `mcp__trusty-memory__memory_recall_deep` return identical results regardless of query.

## Evidence (measured live, 2026-08-04, palace `trusty-tools`)

- 19 distinct queries returned **the same 5 drawers, in the same order, every one at `score: 1.0`**.
- The set included a deliberate nonsense control: `"banana pancake recipe unrelated nonsense gibberish xyzzy"` — same 5 drawers, same scores.
- It also included a query built from a phrase copied verbatim out of one of the returned drawers — no change in results.
- `palace_info` reports the palace holds **1,066 drawers**, so this is not a small-corpus artifact.
- Both `memory_recall` and `memory_recall_deep` exhibit it.

## The HTTP path is NOT affected — this scopes the bug

The `UserPromptSubmit` hook (`trusty-memory prompt-context`) calls `GET /api/v1/palaces/{slug}/recall?q=<prompt>&top_k=5` at [crates/trusty-memory/src/commands/prompt_context/fetch.rs:52-80](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-memory/src/commands/prompt_context/fetch.rs#L52-L80) and IS discriminating: `~/Library/Application Support/trusty-memory/logs/enriched-prompts.*.jsonl` shows **148 distinct injection variants across 1,114 real firings for this palace over 5 days**, with per-prompt variation confirmed by sampling.

So: same palace, two paths, different behavior. The MCP tool path looks broken; the HTTP path looks correct. Suspects to state as suspects, not conclusions — a divergent resolved code path, a caching layer specific to the MCP transport, or an environment difference in how the tool invokes recall.

## Why P1

An operator verifying "is this fact retrievable from the palace?" uses `memory_recall`. It currently answers yes for everything. Issue [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834) proposes migrating file-based memories into the palace and then **deleting the files** — a destructive operation whose safety gate is exactly this tool. A migration verified with a non-discriminating recall could destroy the only copy of an unretrievable fact. This blocks [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834).

## Second finding, include as context not as the headline

[...]

Issue #4834, opened 2026-08-04
feat(trusty-mpm): offer users option to migrate static file memories into palace


bobmatnyc:
## Problem

File-based memory likely exists only because trusty-memory was not available. When trusty-memory IS available, tm should **offer the user** the option to migrate static file memories into the current palace and then clear the static context.

## Evidence

A live tm 1.3.5 session loads ~21.4k tokens of "Memory files" before the first user message. `MEMORY.md` alone is ~6.1k of that, resident every session. Meanwhile a `UserPromptSubmit` hook (`trusty-memory prompt-context`) already injects palace-recalled drawers into every prompt, relevance-filtered. Where the same facts exist in both places, the resident copy is pure cost. An audit quantifying the actual overlap is in flight — reference it as pending.

## Resolution

Offer migration when these conditions are met:

1. **Offer, never automatic.** This deletes user data. Requires explicit consent and a dry-run preview of exactly what would move and what would be deleted.

2. **Availability gate.** Only offer when trusty-memory is genuinely reachable AND a palace resolves — not merely when the MCP server is registered.

3. **Preserve structure on import.** Each file memory carries `name` / `description` / `type` (`user` | `feedback` | `project` | `reference`) frontmatter and `[[wikilink]]` cross-references. Map these onto palace drawer tags/metadata rather than flattening to prose.

4. **Dedup, idempotent, reversible.** Detect and skip existing facts to avoid duplication. Safe to re-run; a second run must not re-import. Back up the files before clearing, or stage the deletion behind a confirmation after the import is verified.

## Acceptance Criteria

- [ ] Migration flow prompts with explicit consent
- [ ] Dry-run preview shows exactly what moves and what deletes
- [ ] Palace availability gated correctly (not just MCP registration)
- [ ] Structure and cross-references preserved in palace metadata
- [ ] Deduplication prevents duplicate facts
- [ ] Import is idempotent
- [ ] Files backed up or deletion staged behind post-import confirmation

bobmatnyc:
Measured context-budget baseline (2026-08-06), including the ~3,400–12,800 token memory-file range this issue's migration would reduce, is posted on [#4595](https://github.com/bobmatnyc/trusty-tools/issues/4595#issuecomment-5204468709). Use that comment as the shared measured baseline rather than re-deriving the split.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

PR #2520, opened 2026-07-13
fix(trusty-memory): force:true honors its contract + secret/blocklist heuristics stop over-matching (closes #2442)

bobmatnyc:
## Summary

Issue #2442 reported two related problems in `trusty-memory`'s `memory_remember` content-quality gates:

**(a) `force: true` did not bypass all gates.** In `handle_memory_remember` (`crates/trusty-memory/src/tools/memory_ops.rs`), `blocklist_gate` and `content_gate` (the short-standalone-content check, issue #215) ran **before** `force` was parsed — only `dedup_gate` and the deep `FilterConfig` noise/secret gate (in `trusty-common::memory_core::retrieval::PalaceHandle::remember_with_options`) already honored `force`. A `force: true` write of blocklisted or short content was still silently skipped, contradicting the tool's documented "bypass filtering" contract.

**(b) The blocklist substring-anywhere match thinned the recall surface.** `blocklist_gate` (and its independently-drifting twin in the dream-cycle retroactive prune pass) matched via `content.contains(pattern)`, so any legitimate memory that merely *quoted* `"Tool use: "` or `"Claude Code session"` mid-prose — realistic for a coding agent recapping its own tool output — was silently dropped, even though the doc comment already described the intent as prefix/frame matching.

## Live evidence (this session, 2026-07-12/13)

During this very trusty-mpm orchestration session, two legitimate milestone memories were rejected by the secret/credential heuristic in `trusty-common::memory_core::filter::looks_like_secret` (message: `"contains a likely secret/credential token"`) even though neither contained a credential:

1. A source-location reference: the literal path text `client/http_client/error.rs::response_or_body_error` — the `::` inside the final slash-segment broke the existing slash-path structural check (`is_word_segment` didn't allow `:`), so the token fell through to the entropy heuristics, where the `/` tripped the base64-symbol branch.
2. A compact issue/PR/SHA ledger reference shaped like `#2486→PR#2491(e993c18a)` — mixing uppercase (`PR`), digits, and lowercase hex was enough to trip the *unconditional* mixed-case+digit fallback, even though no real credential format contains `#`, arrows, or parentheses.

Both had to be reworded to store. Both are now covered by regression tests (`path_like_token_with_rust_module_separator_not_flagged`, `ledger_reference_token_not_flagged`, and the end-to-end `dispatch_remember_accepts_live_false_positive_tokens`) using the exact real-world strings.

## Gate map (before → after)

| Gate | Where | Honors `force` before | Honors `force` after |
[...]

PR #4851, opened 2026-08-04
feat(trusty-mpm): zero-inference bulk memory import via `tm memory import`

bobmatnyc:
Option 1 of [#4837](https://github.com/bobmatnyc/trusty-tools/issues/4837). Unblocks [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834).

## Why

Migrating 148 memory files into a palace was dispatched to an agent. It cost 662k tokens and 283 tool calls to write 120 files — ~5.5k tokens per file copy. It is ETL: read file, map frontmatter, write. No language model needed.

## What

```
tm memory import <DIR> --palace <SLUG> [--dry-run] [--json] [--allow-secret-like] [--memory-url URL]
```

Walks `*.md` files directly inside `<DIR>` (non-recursive, filename order), derives each drawer, and writes it via `trusty_common::mcp::memory_rpc` — the existing shared trusty-memory discovery + JSON-RPC entry point, not a bespoke REST client.

**Mapping.** Validated against the halted run's `payloads.json`: the derivation reproduces **148/148 stored texts byte-for-byte** and **145/148 tag sets**.

| Field | Rule |
|---|---|
| `text` | `description` + blank line + body; a `.` is appended to the description unless it already ends in `.`, `:` or `"` |
| `tags` | sorted union of `name`, `metadata.type`, and every `[[wikilink]]` target |

The three tag differences are deliberate: the hand-derivation dropped `.md`-suffixed link targets outright, losing genuine cross-references. This strips the suffix instead.

The frontmatter scanner is purpose-built rather than `serde_yaml`-backed — the established mapping stores the description's **raw** scalar text, quotes included (most descriptions are quoted headlines whose quotes are prose, not YAML syntax), which a real YAML load would strip. Block scalars (`>`, `>-`, `|`, `|-`) are folded here so that form still parses.

## Duplicate detection — not `memory_recall`

`memory_recall` is broken ([#4836](https://github.com/bobmatnyc/trusty-tools/issues/4836)) and is not used anywhere in this path. Dedup is two exact string comparisons:

1. `memory_list { palace, tag: <slug>, limit: 200 }` — the file's frontmatter `name` (identical to the filename stem, unique per file) is always one of its tags, and trusty-memory filters tags by exact equality server-side. No embedding, ranking, or scoring.
2. **Headline narrowing.** The tag alone is insufficient: a *different* file containing `[[<slug>]]` also carries that tag, so a tag-only check would silently drop the linked file. Candidates are narrowed by requiring the drawer's first line to equal the file's own derived headline. Covered by `linking_drawer_does_not_block_its_target`.

## Report

`--json` prints per-file `status` (`created` / `skipped` / `would_create` / `would_skip` / `failed`), the `drawer_id`, the derived tags, and a `detail` string — so a caller verifies without a second pass. A non-zero `failed` count exits non-zero. Per-file failures never abort the run: one unparseable file cannot strand a 148-file migration halfway the way the agent attempt did.

## Verification

**Gates (rung 4, one batched command — exit 0):**

```
cargo fmt --check && cargo check -p trusty-mpm && cargo clippy -p trusty-mpm -- -D warnings \
  && cargo test -p trusty-mpm && bash scripts/check_line_cap.sh
```

```
test result: ok. 4484 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out
line-cap: measured 3643 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

**Live dry run** against the real 149-file corpus and the running trusty-memory daemon (`--dry-run`, zero writes):

```
{'palace': 'trusty-tools', 'dry_run': True, 'total': 149,
 'created': 31, 'skipped': 118, 'failed': 0}
```

The arithmetic checks out end to end: `MEMORY.md` is correctly classified as "not a memory file (no YAML frontmatter)"; 117 of the 120 drawers the halted run wrote are detected as already-present with their existing `drawer_id` returned; the 31 pending are the 28 never written plus the 3 the halted run stored with reworded text.

**New tests** (`core::memory_import::tests`, 20 total): frontmatter parsing including the `>` and `|-` block-scalar forms, raw-quote preservation, nested `metadata.type` (and a stray top-level `type:` not being mistaken for it), tag derivation, wikilink extraction and normalisation, UTF-8-safe scanning, headline matching, plus loop tests against a stub JSON-RPC daemon asserting a dry run issues zero `memory_remember` calls, a re-run writes nothing and returns the same drawer id, a linking drawer does not block its target, non-memory files skip cleanly, and an unparseable file is reported without aborting the run.

## Scope

Touches only `crates/trusty-mpm/src/core/memory_import/`, the `tm memory` CLI wiring, and a changelog fragment. No file owned by `chore/instruction-prompt-dedup`, `fix/4832-trusty-mpm-dir-convention`, or `feat/4837-agent-cost-guard` is modified. Version left at the pre-bumped 1.3.5.

Refs #4837
Refs #4834


bobmatnyc:
## Verdict: WARN

Adversarial review, code-vs-spec only. Zero CRITICAL, one HIGH — the code proceeds, but the HIGH finding should be tracked.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/memory_import/mod.rs` | 381–388, 411 | Dedup is create-or-skip with an **exact, unnormalised** headline match, and `force: true` explicitly disables trusty-memory's own dedup gate — so this check is the *only* duplicate guard in the system. Any drift between a file's `description` and the stored drawer's first line reads as "absent" and writes a second drawer carrying the same slug tag. The PR states this fires on 3 drawers on the first real run. The CLI help (`cli/actions/memory.rs:21`) promises "Re-running is safe: a file whose drawer is already present is skipped" — false in exactly that case. No test covers drift. | When the tag lookup returns candidates but none match the headline, don't silently create: emit a distinct status (`conflict` / `would_update`) carrying the near-miss `drawer_id`, or gate the write behind an explicit `--allow-divergent`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/memory_import/mod.rs` | 334, 353 | `limit: 200` truncation is silent. Verified server-side that this is *bounded, not unbounded* risk — `handle.rs:852–869` filters by tag **before** truncating, and `memory_ops.rs:377` applies the limit verbatim with no clamp — so it only bites when >200 drawers carry one slug tag. But when it does, the file reads as absent and a duplicate is written with no signal. The stub at `tests.rs:230–246` ignores `limit`, so the path is untested. | Guard: if `drawers.len() >= DEDUP_CANDIDATE_LIMIT` and no headline match, fail that file rather than creating it. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/memory_import/parse.rs` | 152–155 | A frontmatter line with no `:` is skipped, so a YAML **plain multi-line scalar** (`description: a long headline` / `  continued here`) silently drops everything after the first line — and the file still imports, with truncated text. This is the one input that fails **open**; every other malformed shape fails closed (no frontmatter → skip, no `name` → `Failed`). The parser handles `>`/`\|` explicitly *because* multi-line descriptions exist; the un-indicated form is the gap. | After a plain scalar, if the next line is more-indented and contains no `:`, fold it in (YAML plain-scalar semantics) or fail the file. | Fix here |
| LOW | `crates/trusty-mpm/src/core/memory_import/parse.rs` | 235, 251–255 | `block_indent` is a **byte** offset from `trim_start` (which trims Unicode whitespace); `&l[block_indent..]` assumes it is a char boundary on every line. Lines indented `"\u{a0}"` (2 bytes) and `" \u{a0}"` (3 bytes) give `block_indent = 2`, slicing mid-NBSP → panic. Contrived; ASCII indentation is always safe. | `l.get(block_indent..).unwrap_or("")` instead of the slice. | Promote |
| LOW | `crates/trusty-mpm/src/core/memory_import/parse.rs` | 162–168 | `section` tracking is depth-insensitive: any `type:` at indent > 0 while the last indent-0 key was `metadata` sets `kind`, including `metadata.source.type` at depth 2. `top_level_type_is_not_mistaken_for_metadata_type` covers the indent-0 case, not the over-deep one. | Track the section's indent and require the `type:` line to be exactly one level deeper. | Promote |

## Notes

**Verified rather than accepted** (the brief's four probes):

- **`--dry-run` really writes nothing** — traced, not trusted to the counter. `import_one` (mod.rs:279–288) returns `WouldCreate` before `write_drawer` is reachable; the only other RPC on that path is `memory_list`, read-only (`memory_ops.rs:367–394`). `dry_run_writes_nothing` asserts it at the wire against a stub recording every `memory_remember`.
- **Server-side tag filter is genuinely exact equality** — `handle.rs:859`, `d.tags.iter().any(|x| x == t)`. Not substring, not prefix. And the filter runs **before** the `limit` truncation, so `limit: 200` caps the *candidate* set, not the scan.
- **Response field names match** — `drawers` / `drawer_id` / `content` are exactly what `memory_ops.rs:379–393` emits.
- **`--allow-secret-like` default is safe** — it is the narrow half of the issue #2520 two-tier split. `force: true` bypasses content-QUALITY gates (blocklist, short-content, dedup, noise) and explicitly does **not** bypass secret detection (`definitions.rs:107`, `handle.rs:501–510`); only the separate opt-in does, and it defaults false.
- **Idempotency tested against the logic, not the claim** — `import_is_idempotent` drives two full `run_import` passes through a stub that indexes by tag and stores content, asserting exactly one write and a stable `drawer_id`. `linking_drawer_does_not_block_its_target` is a real regression test for the tag-only false positive.
- **Per-file failure isolation + exit code** — every error path in `import_one` returns a `FileResult`; `run_import` (mod.rs:183–185) pushes unconditionally; `commands/memory.rs:50–52` bails after printing. Correct order, both confirmed.
- **`tm-capabilities` regeneration is genuine, not a suppression** — 54→55 plus one `memory`/`import` entry inserted alphabetically between `mcp` and `meta`. Exactly the generator's output for one added command group. Changelog fragment present and correctly category-first.

**Not independently verified:** the 148/148 byte-for-byte fidelity claim. The halted run's `payloads.json` is not in the repo, so I could not reproduce it. The derivation logic is consistent with the claim and the 3 tag deviations are documented in-code at `parse.rs:288–291`, but I am not asserting the claim as confirmed.

**Independent gate, clean worktree off the PR head (`fa550ba4`):**

```
cargo fmt --check                                  → clean
cargo clippy -p trusty-mpm --all-targets -D warnings → clean  (--all-targets; stricter than the PR's)
bash scripts/check_line_cap.sh                     → 3648 files, 7 allowlisted, 0 violations — OK
cargo test -p trusty-mpm  → test result: FAILED. 4483 passed; 1 failed; 7 ignored
```

The single failure is **`client::executor::tests::execute_doctor_against_test_daemon`** — the environmental flake CLAUDE.md names *by name* in its known-flaky list, failing here on daemon port binding (`daemon unreachable: error sending request for url http://127.0.0.1:51803/...`). The diff touches zero files under `src/client/`. All 20 new `core::memory_import::tests` pass.

Note my first gate run piped `cargo test` into `tail`, which masks cargo's exit code — worth avoiding in the PR's own gate string for the same reason.

**On the author's flake dismissal:** sound. I hit a *different* member of the same documented flaky family, and the diff-touches-zero-files proof holds for both. Nothing was `#[ignore]`-d, `cfg`-gated, `--exclude`-d, or deleted to reach green — the diff adds only new tests, and the 7 ignored are pre-existing.

bobmatnyc:
## Review fixes — 81d00953

All four findings addressed. Gate output at the bottom.

### 🔴 HIGH — dedup was create-or-skip, drift wrote a duplicate

**Approach: identify the file's own drawer structurally, never by its prose.**

A foreign slug enters a drawer's tag set through exactly one path — a `[[wikilink]]` in that file's body — and the body is stored verbatim. So re-deriving a candidate's wikilink targets with `parse::wikilink_targets`, *the same function that produced the tags*, reproduces precisely why each referrer carries the tag, and removes precisely those drawers. Alias (`[[x|y]]`), anchor (`[[x#y]]`) and `.md` forms normalise identically on both sides, because it is literally the same normaliser.

That is why it still defeats the linking-drawer case that `linking_drawer_does_not_block_its_target` pins: the referrer is excluded by what its *text links to*, not by what its headline says — so the discrimination no longer depends on prose, and a drifted description stays recognisable. The headline compare survives only to label the skip (`already imported` vs `…has drifted…`); it never decides whether to write.

Classification is now total and fails closed:

| Candidates after referrer filter | Outcome |
|---|---|
| 0 | `Absent` → create |
| 1 | own drawer → **skip**, detail says matched or drifted |
| >1, one carries the headline | that one → skip |
| >1, none carries the headline | **Failed** — "refusing to guess" |
| page full (`>= 200`) | **Failed** — truncated, absence unprovable |

Edge case handled: a file that links to its own slug would be filtered out as a referrer of itself, so for that shape the filter is skipped and every candidate stays in the running.

**No update path was added, deliberately.** trusty-memory exposes no drawer update; refreshing would mean `memory_forget` + re-`remember`, which destroys a drawer's id and `created_at` on the strength of a heuristic match. A misidentification there deletes real data. Reporting drift and leaving the drawer alone is non-destructive and still fully closes the duplicate hole — which was the actual defect.

**Drift test added** (`drifted_description_is_not_reimported`): import, reword the file's `description`, re-import → skipped, same `drawer_id`, `writes.len() == 1`. Plus the harder `drift_behind_a_referrer_is_not_reimported`: the drifted file's slug tag sits on two drawers (its own + the referrer) and neither carries the new headline.

Fail-before evidence — with the headline filter restored:
```
test core::memory_import::tests::drifted_description_is_not_reimported ... FAILED
test core::memory_import::tests::drift_behind_a_referrer_is_not_reimported ... FAILED
test core::memory_import::tests::ambiguous_candidates_fail_closed ... FAILED
test result: FAILED. 25 passed; 3 failed
```

**Help text now says** (`cli/actions/memory.rs`):

> Re-running never writes a file twice. A file's own drawer is found by its slug tag, with drawers that merely link to that slug excluded, so the match does not depend on the file's prose: a file whose text has changed since it was imported is still skipped, and the report says its drawer has drifted. **This command does not refresh a drifted drawer — delete it and re-import to replace it.** When several drawers could be the file's own, or the slug tag is shared by more drawers than one lookup returns, the file is reported as failed rather than guessed at.

`scripts/check_capabilities.sh` is clean — the generator keys off the short help, which did not change.

### MEDIUM 1 — `limit: 200` truncated silently

A full page now hard-errors for that file ("candidate set is truncated and absence cannot be proven"). Paging is not available: `handle_memory_list` takes `(room, tag, limit)` with no cursor or offset.

The stub now honours `limit` and records every `memory_list` params object. `truncated_candidate_set_fails_closed` seeds `LIMIT + 3` drawers under one tag — the stub truncates to exactly `LIMIT`, which is the realistic silent case (`>=`, not `>`) — and asserts the request carried `limit: 200`, the file failed, and nothing was written.

### MEDIUM 2 — parser failed open on a plain multi-line scalar

`scan_frontmatter` returns `Result` and bails on two shapes, so the file is skipped and reported instead of imported with a truncated description:
- a line indented under a key whose value was a non-empty plain scalar → "plain multi-line scalar; rewrite as a single line or a `>`/`|` block scalar"
- a line with no `key:` separator

Both verified against the live 149-file corpus: zero files trip either rule, so this fails closed on malformed input without rejecting anything real.

### LOW

- **UTF-8 dedent panic**: `&l[block_indent..]` sliced at a byte offset. `trim_start` also strips U+00A0, so mixed ASCII/NBSP indentation put the offset mid-character. Replaced with `line.get(n..).unwrap_or_else(|| line.trim_start())`; `dedent_survives_multibyte_indentation` panics on the old code.
- **Depth-insensitive `metadata.type`**: replaced the single `section` string with an ancestor `(indent, key)` stack; `type` is read only as the direct child of a top-level `metadata:`. `type_below_metadata_child_depth_is_ignored` covers it.

### Gates

```
(cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings \
 && cargo test -p trusty-mpm && bash scripts/check_line_cap.sh && bash scripts/check_sld.sh)
EXIT=0

test result: ok. 4492 passed; 0 failed; 7 ignored   (lib; memory_import: 28, was 21)
line-cap: measured 3648 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
bash scripts/check_capabilities.sh → EXIT=0
```

No real import was run; the palace was not written to. Changelog fragment amended in place — no second fragment.

bobmatnyc:
## Verdict: APPROVE (re-review of `81d00953`)

Delta review only. Zero CRITICAL, zero HIGH. All five findings from the [previous WARN](https://github.com/bobmatnyc/trusty-tools/pull/4851#issuecomment-5185681844) are resolved, and the HIGH is resolved *empirically* against the live palace, not just structurally.

## The HIGH is fixed — measured, not argued

I ran `--dry-run` against the real 149-file corpus and the live 1,066-drawer palace on this branch:

```
{'palace': 'trusty-tools', 'dry_run': True, 'total': 149,
 'created': 28, 'skipped': 121, 'failed': 0}

drifted: feedback-clickable-links-always.md
         pr-2863-owned-by-2eb72dca.md
         session-20260708-tga291-p0-windowid.md
```

Against `fa550ba4`'s `created: 31, skipped: 118`. Exactly the 3 disclosed reworded drawers moved from "would be duplicated" to "skipped, drawer has drifted" — named individually in the report. All 120 drawers the halted run wrote are now detected (121 skipped = 120 drawers + `MEMORY.md`); previously only 117 were. The 28 creates are the 28 never-written files. The arithmetic closes on every one of the 149 files.

## The load-bearing premise — tested, and the exposure runs the other way

**Is a `[[wikilink]]` the only path by which a foreign slug reaches a drawer's tag set?** Not an enforced invariant — `memory_remember` accepts caller-supplied `tags`, so any MCP client can hand-tag a drawer. But the consequence is the **opposite** of the one posed in the brief.

For a legitimate drawer to be wrongly filtered out as a referrer, `links_to(drawer.content, slug)` must be true — the drawer's own body must contain `[[slug]]`. That happens only for a file linking to its own slug, which `mod.rs:426` already bypasses. So the "legitimate drawer filtered out → re-imported as a duplicate" path does not exist.

The real residual is the reverse: a drawer carrying the slug for a *non*-wikilink reason **survives** the filter. As the lone survivor it is classified `Same`/`Drifted` and the file is **silently skipped and never imported** — an omission, not a duplicate. With any second candidate present the `many` arm fails closed (`mod.rs:445`). Empirically absent today: `failed: 0` and all 149 files accounted for means no such drawer exists across the 120 hand-migrated ones.

**Self-linking corpus probe:** zero of the 149 files link to their own slug. The `self_linking` bypass never fires on this corpus, so the shape the author flagged is not present and bypassing it is moot here.

**Classification arms** — all five verified total and fail-closed: `>= LIMIT` → bail (`mod.rs:416`); `[]` → create; `[one]` → skip as `Same` or `Drifted`; `many` + headline hit → skip; `many` + no hit → bail. The headline compare only *labels* the skip; it never gates a write.

## No update path — reasoning holds

Verified against the tool registry: `memory_remember`, `memory_note`, `memory_forget`, `memory_list`, three `memory_recall*`, `memory_send_message`. There is **no drawer-update tool** — `palace_update` is palace metadata. Refreshing genuinely means forget + re-remember, destroying `drawer_id` and `created_at` on a heuristic match. Skip-and-report is the correct call, the report names the drifted files individually, and the help text now states the limit and the remedy ("delete it and re-import to replace it"). Coherent.

## Previous findings

| Was | Now | Evidence |
|---|---|---|
| HIGH — drift writes a duplicate | **Resolved** | Structural discriminator; live dry run shows 3 drifted skipped, 0 duplicated |
| MEDIUM — silent 200 truncation | **Resolved** | Hard error at `mod.rs:416`. No-cursor claim **verified**: `memory_list` schema exposes only `palace`/`room`/`tag`/`limit`, and `list_drawers` ends in `filtered.truncate(limit)` — no offset or cursor exists. Stub now truncates after tag-filtering, matching the real handler |
| MEDIUM — plain multi-line scalar failed **open** | **Resolved** | `scan_frontmatter` returns `Result` and bails (`parse.rs:165`, `:174`). Now fails closed. Zero of the 149 real files trip either new rule (`failed: 0`) |
| LOW — dedent panic on mixed-width indentation | **Resolved** | `dedent()` uses `line.get(n..).unwrap_or_else(...)`; unpanicable |
| LOW — depth-insensitive `metadata.type` | **Resolved** | Ancestor `(indent, key)` stack; `type` matched only at `path.len() == 1 && path[0].1 == "metadata"`. Traced both depths |

## New findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/memory_import/mod.rs` | 427–434 | "A foreign slug enters a tag set only via a wikilink" is an unenforced assumption about palace state — `memory_remember` takes caller-supplied `tags`. A hand-tagged drawer matching a file's slug survives the referrer filter; as the lone survivor the file is silently **skipped and never imported**. Fails toward omission, and the `many` arm fails closed, so no duplicate risk. Zero instances in the live palace today. | Require a positive signal before treating a lone survivor as the file's own drawer — e.g. confirm the candidate's tag set is a superset of `parsed.tags` minus link-derived ones, else report rather than skip. | Promote |
| LOW | `crates/trusty-mpm/src/core/memory_import/mod.rs` | 426 | The `self_linking` bypass disables referrer filtering entirely. If a self-linking file is not yet imported and exactly one referrer exists, that referrer is classified as the file's own drawer and the file is skipped. Unreachable on this corpus (0 of 149 self-link) and untested for that arm. | Add a stub test for the shape, or narrow the bypass to exclude candidates whose first line differs. | Promote |

## Gate — clean worktree off `81d00953`, redirected, no trailing pipe

```
cargo fmt --check                                        → clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings   → clean (--all-targets; stricter than the PR's)
cargo test -p trusty-mpm  → 4491 passed; 1 failed; 7 ignored
bash scripts/check_line_cap.sh    → 3648 files, 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh         → 55 specs + 3060 code files; 0 errors, 0 warnings
bash scripts/check_capabilities.sh → up to date (6 generated files)
```

The one failure is again **`client::executor::tests::execute_doctor_against_test_daemon`**, failing `daemon unreachable` both in the full run and in isolation. `docs/reference/test-ladder-baseline.md:19` documents it precisely: *"takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"* — and my machine was under a concurrent cargo build. `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/` returns **0 files**. Pre-existing environmental, not caused by this diff; 4491 + 1 = the author's 4492.

**Nothing suppressed to reach green:** no `#[ignore]`, `#[cfg(...)]`, `allow(dead_code)`, `--exclude` or `--lib` added anywhere in the delta; zero test functions removed; memory_import tests 20 → 28. `check_capabilities.sh` passing is itself independent proof the catalog regeneration was genuine rather than a suppressed drift gate.

No real import was run — `--dry-run` only.

PR #4858, opened 2026-08-04
docs(trusty-mpm): concurrent-dispatch file ownership, fixed agent scope, and the gate-chain pipe ban

bobmatnyc:
## Outcome

Three dispatch practices observed on 2026-08-03 now live in the bundled instruction assets. Refs [#4837](https://github.com/bobmatnyc/trusty-tools/issues/4837).

## Coverage before this PR

I searched every bundled skill under `crates/trusty-mpm/src/assets/skills/` plus `agents/` and `instructions/`. All three were **absent**, not partial:

| Practice | Nearest existing text | Verdict |
|---|---|---|
| File ownership in dispatch briefs | `tm-pr-workflow.md:141` mentions concurrent PRs, but only about changelog-fragment filenames. `tm-delegation-patterns` "Delegation Best Practices" is six generic items with nothing about concurrency. | absent |
| A running agent's scope is fixed | `tm-delegation-patterns` "Long-Wait Delegation" item 4 covers re-engaging a parked agent — the opposite direction. | absent |
| Gate chain ending in a pipe | `BASE-AGENT.md` "Never Directly Monitor a Declarative Process" step 2 actively recommends `\| tail`, which is where the trap gets created. | absent, and the trap was being taught |

## What changed

**`crates/trusty-mpm/src/assets/skills/tm-delegation-patterns.md`** — two new sections:

- *Concurrent Dispatch: Declare File Ownership.* Name the files owned by every other in-flight branch in each brief; the agent stops and reports rather than editing one. "We stack, we do not race." The doctrine already said no two in-flight PRs share a module and coupled changes stack rather than race; the ownership block is the mechanism that makes an agent honour it, since the agent cannot see the other branches. Five concurrent PRs on 2026-08-03 carrying the block produced zero merge conflicts and zero rebases.
- *A Running Agent's Scope Is Fixed.* New work is a new agent, or it waits. Cost tracks accumulated context over the agent's lifetime — every tool round re-sends the whole transcript — so one agent given two mid-flight additions reached 269.9k tokens writing a five-line changelog file. Several narrow agents ending near 60k cost far less than one reaching 400k doing identical work. The section explicitly excludes `SendMessage`-ing a parked agent the CI outcome for work it already owns, so it does not contradict Long-Wait Delegation item 4.

**`BASE-AGENT.md`** (both copies) — new subsection under "Your own gates DO block, in the foreground":

- A pipeline takes its exit status from the last command, so `cargo test … | tail` exits 0 on a failing suite. Banned, with the canonical form `( <gate> && <gate> ) > /tmp/gates.txt 2>&1; echo "EXIT=\$?"`. It produced a false green twice in one day, once for an engineer and once for a reviewer.
- Step 2 of "Never Directly Monitor a Declarative Process" now points forward to that rule where it recommends `tail`/`grep`.

Written in the terse register `origin/main` just adopted: name the forbidden thing, ban it, give one copy-paste command.

## Mirror parity

`crates/trusty-code/src/assets/agents/BASE-AGENT.md` updated byte-identical — both files hash `3e3492c598f78bfbc0b24a6a1f9123180d6c8e4560eba32a67f4877c1bbcf22c`, and `scripts/check_agent_assets.sh` passes.

## Risk

Instruction-asset text only. Zero `.rs` files changed (`git diff --name-only origin/main...HEAD -- '*.rs'` → 0).

## Test evidence

Rung 1 (docs/assets), widened to clippy+test because both crates `include_str!` `BASE-AGENT.md`.

\`\`\`
cargo fmt --check                                            OK
scripts/check_agent_assets.sh   30 byte-parity file(s), all match; 4 pinned deviation(s) match upstream — OK
scripts/check_sld.sh            55 spec doc(s) + 3058 code file(s); 0 error(s), 0 warning(s)
scripts/check_line_cap.sh       3647 tracked .rs file(s); 7 allowlisted, 0 violations — OK
scripts/check_changelog_fragment.sh                          OK
cargo clippy -p trusty-mpm -p trusty-code --all-targets -- -D warnings   OK
\`\`\`

\`\`\`
test result: FAILED. 4500 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 61.97s

thread 'client::executor::tests::execute_doctor_against_test_daemon' panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18:
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:49929/api/v1/doctor?...)")
\`\`\`

change-specific gates pass; `cargo test -p trusty-mpm` blocked by the documented environmental flake `execute_doctor_against_test_daemon` (`docs/reference/test-ladder-baseline.md:19` — "takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"). The failure is timeout-shaped and this diff touches zero `.rs` files.

## Documentation / changelog

Fragments added: `crates/trusty-mpm/changelog.d/4837-dispatch-practices.md`, `crates/trusty-code/changelog.d/4837-base-agent-gate-chain-pipe.md`.

## Review-finding disposition

None yet.

### DIFF
--- a/crates/trusty-code/changelog.d/4837-base-agent-gate-chain-pipe.md
+++ b/crates/trusty-code/changelog.d/4837-base-agent-gate-chain-pipe.md
@@ -1,0 +1,3 @@
+Added
+
+- Embedded `BASE-AGENT.md` bans ending a gate chain in a pipe — `cargo test … | tail` exits 0 on a failing suite — and gives the canonical redirect-then-echo form (mirrors trusty-mpm, [#4837](https://github.com/bobmatnyc/trusty-tools/issues/4837))
--- a/crates/trusty-code/src/assets/agents/BASE-AGENT.md
+++ b/crates/trusty-code/src/assets/agents/BASE-AGENT.md
@@ -212,19 +212,29 @@ an agent told to "rerun the suite until green" spent 415k tokens because
 `cargo test` prints a line per test; a sibling spent 546k on `gh pr checks
 --watch` streaming a 15–17 minute CI job.
 
-1. **Run it.** Don't watch, tail, or poll a live stream.
-2. **Trim the output.** `--quiet` plus `tail`/`grep` for the summary line covers
-   most gates — but never as the last stage of a gate chain, see "Never end a
-   gate chain in a pipe" below. For a heavier squeeze this repo ships a Unix
-   filter:
-   `<command> 2>&1 | tm compress --tool "<tool-name>"` (`--tool` is free-form,
-   substring-matched — `"cargo test"`, `"git diff"`). Known gap: its
-   structured-format guard can misread a leading `key: value`-shaped line — such
-   as a `warning: <path>: …` build warning ahead of the test output — as YAML and
-   silently skip compression. `--quiet` is the reliable default; `tm compress` is
-   an addition on top, not a replacement, until that gap is fixed.
-3. **Still long → have Haiku summarize it** before you read it.
-4. **Only then read it.**
+1. **Run it into a file, never a pipe.** Don't watch, tail, or poll a live
+   stream. A pipe eats the verdict — see "Never end a gate chain in a pipe".
+
+   ```bash
+   <command> > /tmp/gate.txt 2>&1; echo "EXIT=$?"
+   ```
+
+2. **`EXIT=0` → stop. Do NOT read the file.** Nothing in it is information.
+3. **Non-zero → trim the file, then read it.** Trim reads FROM the file, never
+   from the live command: `--quiet` on the command, `grep`/`tail` over the file,
+   or this repo's Unix filter:
+
+   ```bash
+   tm compress --tool "cargo test" < /tmp/gate.txt
+   ```
+
+   `--tool` is free-form, substring-matched (`"cargo test"`, `"git diff"`). Known
+   gap: its structured-format guard can misread a leading `key: value`-shaped
+   line — such as a `warning: <path>: …` build warning ahead of the test output —
+   as YAML and silently skip compression. `--quiet`/`grep` is the reliable
+   default; `tm compress` is an addition on top, not a replacement, until that
+   gap is fixed.
+4. **Still long → have Haiku summarize it** before you read it.
 
 On failure, re-run only the failing case with full output. That is the only place
 per-test detail carries information.
@@ -286,15 +296,20 @@ the evidence you owe. Run it as a plain foreground command with an explicit long
 
 ### Never end a gate chain in a pipe
 
-🔴 A pipeline's exit status is the LAST command's. `cargo test … | tail` exits 0
-on a failing suite. FORBIDDEN — it produced a false green twice in one day, once
-for an engineer and once for a reviewer. Redirect to a file, then echo the status:
+🔴 A pipeline's exit status is the LAST command's. `cargo test … | tail` AND
+`cargo test … | tm compress` both exit 0 on a failing suite — the trim this file
+recommends is itself the trap, not just an aside. FORBIDDEN: it produced a false
+green twice in one day, once for an engineer and once for a reviewer. Redirect,
+then echo the status:
 
 ```bash
 ( <gate> && <gate> ) > /tmp/gates.txt 2>&1; echo "EXIT=$?"
 ```
 
-`EXIT=0` → don't read the file. Non-zero → Read only the failing portion.
+`EXIT=0` → don't read the file. Non-zero → Read only the failing portion. Trim
+the FILE when it is long (`tm compress --tool "cargo test" < /tmp/gates.txt`),
+never the live command. Must you genuinely pipe? `set -o pipefail` in the SAME
+invocation — `$PIPESTATUS` is a bashism and this harness runs zsh.
 
 ## Agent-Authored Prose
 
--- a/crates/trusty-code/src/assets/agents/BASE-AGENT.md
+++ b/crates/trusty-code/src/assets/agents/BASE-AGENT.md
@@ -214,7 +214,9 @@ an agent told to "rerun the suite until green" spent 415k tokens because
 
 1. **Run it.** Don't watch, tail, or poll a live stream.
 2. **Trim the output.** `--quiet` plus `tail`/`grep` for the summary line covers
-   most gates. For a heavier squeeze this repo ships a Unix filter:
+   most gates — but never as the last stage of a gate chain, see "Never end a
+   gate chain in a pipe" below. For a heavier squeeze this repo ships a Unix
+   filter:
    `<command> 2>&1 | tm compress --tool "<tool-name>"` (`--tool` is free-form,
    substring-matched — `"cargo test"`, `"git diff"`). Known gap: its
    structured-format guard can misread a leading `key: value`-shaped line — such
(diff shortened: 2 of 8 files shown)


bobmatnyc:
## Follow-up: the recommended trim had the same defect

`989cb966` fixes the pipeline this PR originally only annotated.

**`tm compress` does mask the exit code.** Verified, not assumed:

```
sh -c 'echo "test result: FAILED"; exit 101' | tm compress --tool "cargo test"   -> EXIT=0
set -o pipefail; <same pipeline>                                                 -> EXIT=101
sh -c '...' > f 2>&1; rc=$?; tm compress --tool "cargo test" < f                 -> rc=101
```

Nothing in that pipeline preserves the upstream status. `crates/trusty-mpm/src/bin/tm/commands/compress.rs:45` states the intent — exit 0 "so it never breaks the exit-code semantics of whatever pipeline it's the tail of" — but exiting 0 does not *preserve* the upstream status, it *replaces* it. The filter is well-mannered; the caller still loses the verdict. That doc comment is misleading in the way that matters; left unedited here to keep this PR free of `.rs` changes, and worth a one-line correction later.

**Canonical form chosen: redirect-then-echo.**

```bash
<command> > /tmp/gate.txt 2>&1; echo "EXIT=$?"
```

Reasons for that over the alternatives:

- `${PIPESTATUS[0]}` is a **bashism**. This harness runs zsh, where the array is `$pipestatus` and 1-indexed — a canonical form that silently breaks in the actual shell is worse than none.
- `set -o pipefail` works (proven above) but is shell state, and the Bash tool does not persist it between calls, so it only holds when written into the same invocation. Kept as the named escape hatch, not the default.
- Redirect-then-echo is the **same shape** as the gate-chain rule already added in this PR, so `BASE-AGENT.md` now teaches one shape instead of two. It also exposes the real insight: on a green gate you never read the file, so there is nothing to trim. Trimming belongs after a non-zero verdict, applied to the file.

**Locations changed** (`BASE-AGENT.md`, both copies, byte-identical at `83c2f5961766e7790ff3bfd77fd852999f16f57d0dafeb96cbc38be39dbd2315`):

- "Never Directly Monitor a Declarative Process" steps 1–4 rewritten. Step 1 captures to a file; step 2 is `EXIT=0` → do not read; step 3 trims the **file** (`tm compress --tool "cargo test" < /tmp/gate.txt`) and only on non-zero. The `<command> 2>&1 | tm compress` form is gone, not annotated.
- "Never end a gate chain in a pipe" now names `| tm compress` alongside `| tail`, and says outright that the trim this file recommends was itself the trap.

**Sweep for other piped gates** across every bundled skill, agent file, and instructions section:

| Location | Command | Verdict |
|---|---|---|
| `BASE-AGENT.md:220` (before this fix) | `<command> 2>&1 \| tm compress` | the only asset taking a gate verdict from a pipe — fixed |
| `skills/test-driven-development/references/integration.md:368` | `npm test --coverage --silent \| grep \| awk` | fine — extracts a coverage number; the pass/fail gate is the separate unpiped `npm test` step above it, and the `if [ $COVERAGE -lt 80 ]` is the verdict |
| `skills/root-cause-tracing/references/advanced-techniques.md:55` | `npm test 2>&1 \| grep 'DEBUG …'` | fine — pulls a debug line during an investigation where the test is expected to fail; no verdict taken |
| `skills/testing-anti-patterns/references/detection-guide.md:557` | `git diff --cached \| grep -q` | fine — `grep` *is* the intended verdict |
| `skills/systematic-debugging/references/examples.md:403` | `security find-identity \| grep … \|\| exit 1` | fine — same, `grep` is the verdict |
| `agents/react-engineer.md:52` | `npm test … \|\| npx vitest run` | fine — `\|\|` fallback, not a pipe |

`tm compress` appeared in exactly one place across all assets.

## Gates (re-run after the fix)

```
cargo fmt --check                                                        OK
scripts/check_agent_assets.sh   30 byte-parity file(s), all match; 4 pinned deviation(s) match upstream — OK
scripts/check_sld.sh            55 spec doc(s) + 3058 code file(s); 0 error(s), 0 warning(s)
scripts/check_line_cap.sh       3647 tracked .rs file(s); 7 allowlisted, 0 violations — OK
scripts/check_changelog_fragment.sh   scanned 5 changed path(s); all 2 crate(s) with source changes are recorded
cargo clippy -p trusty-mpm -p trusty-code --all-targets -- -D warnings   OK
```

```
test result: FAILED. 4500 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 69.88s
panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18
```

change-specific gates pass; `cargo test -p trusty-mpm` blocked by the documented environmental flake `execute_doctor_against_test_daemon` (`docs/reference/test-ladder-baseline.md:19`). Same test, same timeout shape as the first run; this diff still changes zero `.rs` files.

The `trusty-mpm` fragment was split in two — the gate rejects a second category stacked inside one file.

bobmatnyc:
## Follow-up 2: `compress.rs` doc comment corrected

`8cf99bfa`. Comment only — the exit-0 filter contract is untouched.

**Before** (`crates/trusty-mpm/src/bin/tm/commands/compress.rs:42`):

> Why: the single entry point the rewritten Bash command pipes into (`<cmd> | tm compress --tool "<effective tool name>"`); must behave like a well-mannered Unix filter — read all of stdin, write the (possibly unchanged) result to stdout, exit 0 — **so it never breaks the exit-code semantics of whatever pipeline it's the tail of.**

**After:**

> Why: a well-mannered Unix filter — read all of stdin, write the (possibly unchanged) result to stdout, exit 0 — so trimming a caller's output never itself introduces a failure.
>
> 🔴 Exiting 0 DISCARDS the upstream command's exit status; it does not preserve it. A shell takes a pipeline's status from its LAST command, so `cargo test … | tm compress` reports success on a failing suite. Callers that need the verdict must capture it before trimming, never pipe:
>
> ```bash
> <gate> > /tmp/gate.txt 2>&1; echo "EXIT=$?"
> tm compress --tool "cargo test" < /tmp/gate.txt   # only when non-zero
> ```
>
> The exit-0 contract itself is deliberate and unchanged — callers rely on this filter never failing them. #4837: the comment previously justified it as never breaking a pipeline's exit-code semantics, which is backwards. See `assets/agents/BASE-AGENT.md`, "Never end a gate chain in a pipe".

The `What:` and `Test:` axes are unchanged, and the fenced block is tagged `bash` so rustdoc does not treat it as a doctest.

## Gates

A `.rs` file now changes, so the full test job is in scope.

```
cargo fmt --check                                                        OK
scripts/check_agent_assets.sh   30 byte-parity file(s), all match; 4 pinned deviation(s) match upstream — OK
scripts/check_sld.sh            55 spec doc(s) + 3058 code file(s); 0 error(s), 0 warning(s)
scripts/check_line_cap.sh       3647 tracked .rs file(s); 7 allowlisted, 0 violations — OK
scripts/check_changelog_fragment.sh   scanned 6 changed path(s); all 2 crate(s) with source changes are recorded
cargo clippy -p trusty-mpm -p trusty-code --all-targets -- -D warnings   OK
```

```
test result: FAILED. 4500 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 108.47s
panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18
```

Same test, third consecutive run. Two proofs it is not this diff:

```
git diff --cached -U0 -- '*.rs' | grep '^[+-]' | grep -v '^[+-]\s*///'   -> 0 non-comment lines
```

```
cargo test -p trusty-mpm --lib -- --exact client::executor::tests::execute_doctor_against_test_daemon
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4507 filtered out; finished in 8.09s
```

8.09s in isolation against the 10s client timeout — exactly the margin `docs/reference/test-ladder-baseline.md:19` describes ("takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"). It passes alone and loses under the full-suite load.

change-specific gates pass; `cargo test -p trusty-mpm` blocked by the documented environmental flake `execute_doctor_against_test_daemon`.

No new changelog fragment — the doc comment is internal, not user-visible, and the fragment gate is satisfied by the two already present.

PR #4859, opened 2026-08-04
fix(vmtest-harness): restore die() classification across every subshell construct

mac-duetto:
`die <code> <msg>` is the harness's sole classification mechanism: it sets the write-once global `VMTEST_EXIT` and exits. A shell global cannot cross a fork, so every `die` firing inside `$( )`, `<( )`, a for-list, or a heredoc substitution wrote its verdict into a child that  died with it. **Seventeen call sites were affected.**

  Three symptoms followed, determined by which construct applied:

  1. **The MEASURE line lied** — `VMTEST_EXIT` unset, so it fell back to a value that is structurally always 0. A run that failed 50 logged `exit 0`.
  2. **The write-once slot was left unclaimed**, so a later teardown `die 70` took it and the process exited 70 — the override DOC-2 §2  forbids, with a code whose own table entry disclaims the run result.
  3. **The failure was swallowed entirely** — `set -e` never fired. At one site this produced a green `verify_stack_doctor PASS: all 0  in-scope package(s)`: a **vacuous PASS over an empty set**.

  `tsv_scope_crate_dirs` and `tsv_scope_packages` had **no call site anywhere in the harness that preserved their `die 60`**, making the P4-T4 postcondition tripwires fully non-functional as classifiers.

  ## Root cause is in the spec, not just the code

  DOC-2 §12.1 mandates that a `lib/` function return status via exit code, emit its value on stdout, and classify its own failures via `die`.  In bash the only way to capture stdout is a command substitution, which forks — so **"emits a value on stdout" and "classifies via `die`"  cannot both hold for the same function.** §12.5's worked skeleton then prescribed two of the defective sites verbatim, so fixing only the  code would have left the next implementer to re-derive the bug from the spec.

  ## What changed

  **Part 1 — the slot is file-backed.** `die` writes its code to `$VMTEST_TMPDIR/exit-code` under `set -o noclobber` (an `O_EXCL` create, so  write-once is atomic rather than merely single-threaded), reads it back immediately, and `vmtest_cleanup` reconciles it into `VMTEST_EXIT`  after capturing the exit status and before both the MEASURE line and tmpdir removal. Under `--keep` the verdict is copied into the run directory.

  **Part 2 — the fork is removed where a side channel cannot help.** No side channel restores an *abort* at the constructs that discard the   child's status, so `tsv_scope_crate_dirs` and `tsv_scope_packages` now take an out-path (§12.1's existing escape hatch, `vm_list <out_tsv_path>`'s precedent) and all ten call sites became plain commands. `verify_rustc` and `source_deliver_branch` drop the undeclared
  stdout emits §12.2 never declared.

  **Also:** `vmtest_cleanup` receives the captured rc as an argument, restoring §Shell-discipline cleanup property 1 that `on_exit`'s `local
  rc=$?` had made structurally unsatisfiable; and `verify_stack_doctor` gains the emptiness guard its two sibling oracles already have.

  **Spec amendments land with the code:** DOC-2 §1.1, §1.1a, §2, §12.1, §12.4, §12.5 and MANIFEST Phase 8 Deviations.

  ## Tests — the first coverage this harness has had

  - `tests/subshell-classification.sh` — behavioural, no VM, 12 cases across all six constructs plus both signal paths
  - `tests/check-no-swallowed-die.sh` — static guard, self-deriving: it recomputes the die-capable set on every run, so a future `die` is
  picked up with no edit
  - `tests/check-no-swallowed-die-selftest.sh` — mutation-proves the guard can go RED across nine defect classes, plus a control and a
  vacuous-scan floor

  **No GitHub Actions workflow** — CI integration is out of scope per D5 (`DOC-1:1009`, `DOC-1:215`). These are run by hand:

  ```
  bash vmtest-harness/tests/check-no-swallowed-die-selftest.sh
  bash vmtest-harness/tests/check-no-swallowed-die.sh
  bash vmtest-harness/tests/subshell-classification.sh
  ```

  ## Evidence

  Gates, all under real `/bin/bash 3.2.57` (the declared target interpreter):

  ```
  bash -n over vmtest lib/*.sh scenarios/*.sh tests/*.sh   exit=0
  shellcheck -s bash -S error (incl. tests/)               exit=0
  check-no-swallowed-die-selftest.sh                       10 passed, 0 failed
  check-no-swallowed-die.sh    OK — 96 functions, 46 direct, 56 transitive
  subshell-classification.sh                               12 passed, 0 failed
  ```

  **Fail-before / pass-after:** `subshell-classification.sh` is 0/10 on the pre-fix tree and 12/12 after, file unmodified between runs.

  **Both VM paths observed** on real Tart VMs (tart 2.32.1, pinned `tahoe-base`):

  *Positive* — `MEASURE run_wall_clock_s 636 (exit 0; excludes teardown)`, `verify_binaries PASS: 13/13`, `verify_single_install PASS` ×4,
  `verify_stack_doctor PASS`, `verify_daemon_liveness PASS: 5 in-scope daemon(s) live`, all 8 in-scope crates installed, no `FAIL[…]` line in
  341 lines.

  *Negative* — `guest_src_dir` pointed at a read-only path so `die 50` fires inside a genuine command substitution:

  ```
  vmtest: FAIL[50]: could not prepare the guest source directory /vmtest-src
  vmtest: MEASURE run_wall_clock_s 39 (exit 50; excludes teardown)
  EXITCODE:50
  ```

  Side-channel file: `50`. All four signals agree; no teardown code overwrote the classification.

  ## Notes for review

  - **Test ladder rung 3.** `git diff --name-only origin/main...HEAD -- crates/` is empty — this touches only `vmtest-harness/` and
  `docs/research/`, which no Cargo target compiles or tests, so the rung executes with its shell equivalents above.
  - **No changelog fragment:** `vmtest-harness/` is not a crate and has no `changelog.d/`.
  - **SLOC cap N/A:** it covers tracked `.rs` files only.
  - **Class C deferred by design.** `provision_detect_mise` and `source_deliver_local` keep their declared `emits` signatures and rely on
  Part 1's backstop, so DOC-2 §12.2 needs no change.
  - **Known limitation:** the static guard cannot see through indirect invocation (`"$accessor"`, `"$fn"`). The three known sites carry
  explicit `swallowed-die-check: indirect` annotations.

  Originally filed and merged on the `mac-duetto` fork: https://github.com/mac-duetto/trusty-tools/issues/1


bobmatnyc:
Superseded by #4938, which lands the same die() classification fix as merge commit 6acc0b74ee51f621cf47d7c4acb964e2cb04ff0d.

This PR could not be merged directly: it's a cross-repo PR from mac-duetto's fork, and it had a genuine merge conflict against current main. Two of its four commits had already landed independently as #4861 and #4865, which is what produced the conflict.

Thanks to mac-duetto for the original fix — the remaining work was rebased and carried forward in #4938. No action needed on the fork branch; it's left as-is.

Issue #4393, opened 2026-07-30
Restart safety: no `ExitTimeOut` in rendered plists, and trusty-search's orphan reaper is lockfile- and data-dir-blind


mac-duetto:
## Summary

trusty-search's shutdown flush budget is **≥30 s per index** by design, but **every termination window in the system is 3–5 seconds**. Whichever path stops the daemon, a flush that actually has work to do is SIGKILLed before it finishes.

*(This issue was originally filed combining three defects. It has been narrowed to the termination-window mismatch; the orphan reaper is tracked in #4395 and the trusty-memory SSE drain in #4396 — see the revision comment.)*

## The mismatch

| Path | Window | Source |
|---|---|---|
| launchd (`ExitTimeOut`) | **5 s** measured | `launchctl print gui/501/com.trusty.memory` → `exit timeout = 5` |
| `trusty-search stop` | **5 s** | `commands/stop.rs:81, 95-107` |
| orphan reaper | **3 s** | `commands/start/daemon.rs:281` |
| flush budget | **≥30 s/index**, scaling to 20 min | `service/shutdown_flush.rs:26, 90-98` |

The rendered plists carry **no `ExitTimeOut` key at all** (`trusty-common/src/launchd.rs:137-222`, read in full — Label, ProgramArguments, KeepAlive, RunAtLoad, ThrottleInterval, log paths, fd limits, env, nothing else), so launchd applies its system-defined default. `man launchd.plist` states the default is *"system-defined"*; on this host it measures **5 s**, confirmed independently against a scratch LaunchAgent (see below).

## `launchctl kickstart -k` semantics — measured, not assumed

An earlier revision flagged this as an open question. It has now been tested empirically on a throwaway LaunchAgent with a SIGTERM trap:

```
10:15:17.283  STARTED pid=61361
10:15:30.537  GOT SIGTERM pid=61361      <- ~14 ms after `kickstart -k`
10:15:35.567  STARTED pid=61865          <- new PID, 5.03 s later
```

The trap's post-handler line never appeared for pid 61361 — it was SIGKILLed mid-handler, exactly at the 5 s `exit timeout` boundary.

**Conclusion: `kickstart -k` sends SIGTERM first and escalates to SIGKILL only after `ExitTimeOut` elapses.** It is not a direct-to-SIGKILL operation. Therefore adding/raising `ExitTimeOut` **does** protect this path, including `tctl install`'s kickstarts.

[...]

Issue #4227, opened 2026-07-28
trusty-search: genuine corpus corruption never quarantines — #4122's population is transient-dominated


bobmatnyc:
Defect: `core/indexer/quarantine.rs` module docs list "genuine corruption" among the conditions that trigger the #4122 write quarantine. It is not one. Corruption is absorbed before it can set `corpus_open_failed`.

Evidence:
- `core/corpus/store_impl.rs:71` — `CorpusStore::open` delegates to `open_corpus_db_or_recreate`.
- `core/corpus_recovery.rs:45-72` + `:100-105` — classifies `UpgradeRequired`, `RepairAborted`, `Storage(Corrupted(_))` and `Storage(Io(InvalidData))` as recoverable, moves the file aside with a `.v2-incompatible` suffix, and returns `Ok` with a fresh EMPTY corpus.
- `corpus_recovery.rs:98` states it outright: "Returns false for lock contention and genuine transient I/O errors" — those are the only errors that propagate as `Err` and set `corpus_open_failed`.

Two consequences worth recording:
1. The quarantine population is dominated by TRANSIENT failures — exactly the population an in-process reopen retry would clear. Strengthens the case for an in-process retry follow-up (see #4122 / PR #4220 HIGH-1, where recovery currently requires a daemon restart).
2. The recreate path yields an index reporting HEALTHY with a fresh empty corpus, which watcher writes then populate with a partial rebuild — a symptom shape close to the #4122 incident's, which the quarantine explicitly does NOT cover. Needs a sentence scoping it against #4087 so the next reader does not assume #4122 closed it.

Resolution: correct the module doc's condition list, and decide whether the recreate-to-empty path warrants its own guard (likely related to #4087).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Moved to ts 0.45.0 — the release branch documents this but does not fix it. Corruption is absorbed by `open_corpus_db_or_recreate` before quarantine can trigger; the actual recreate-to-empty guard belongs to #4087.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #4866, opened 2026-08-05
fix(trusty-search): a write-quarantined index performs no durable write

### LINKED ISSUE #4226: trusty-search: chunks.json is the one durable write a #4122-quarantined index still performs
Defect: PR #4220 write-quarantines an index whose redb corpus failed to open, and its ERROR diagnostic claims "the on-disk corpus is untouched and still recoverable". True for redb, false for the legacy JSON snapshot. At shutdown a quarantined index writes its (deliberately empty) in-memory corpus over chunks.json.

Evidence:
- `persist.rs:490-493` — `flush_corpus_to_disk` falls through to `save_chunks_to_disk` precisely when no `CorpusStore` is wired, which is exactly the quarantined case.
- `persist.rs:37` documents "Empty corpus is still written so the on-disk file accurately reflects state".
- `persist_hnsw.rs:283` sets `persist_chunks_json = self.corpus.is_none()` on the same condition.
- `indexer/migrations.rs:120-135` — `JsonCorpusToRedbMigration` reads that same file to seed an empty redb.

Blast radius is narrow (the JSON migration only runs when redb is empty AND the schema stamp is unset) and it is pre-existing, not introduced by #4220. But it is the one durable write a quarantined index still performs and the PR's own diagnostic claims otherwise.

Resolution: gate the chunks.json write on `!corpus_open_failed`, or scope the ERROR string's claim to redb. One line either way.

bobmatnyc:
Closes #4226

## The write path that still fired

`CodeIndexer::flush_corpus_to_disk` (`core/indexer/persist.rs`). Its `None` arm was written as the legacy-index case, but `corpus_open_failed ⇒ corpus == None`, so it is also the write-quarantine case. At shutdown, `service::shutdown_flush` calls it, it falls through to `save_chunks_to_disk`, and that writes the quarantined index's deliberately-empty in-memory corpus over `chunks.json`. The seeded snapshot in the new test goes from 2 chunks to 0.

The quarantine's own ERROR diagnostic said the opposite: *"the on-disk corpus is untouched and still recoverable."* True of redb, false of the JSON snapshot.

It was not the only one. `spawn_incremental_persist` (`core/indexer/persist_hnsw.rs`) is a second caller of the same writer — its `persist_chunks_json = self.corpus.is_none()` is the identical predicate — reached from `commit_parsed_batch` on the bulk-reindex path, which #4122 deliberately left ungated because boot reconcile auto-fires it. And both that persister and `save_vector_store` write the HNSW graph, which is wired even on a quarantined index: `build_store_for_entry` runs *before* the corpus open.

## Gated the path, not the call

Why the #4122 invariant does not cover this: `corpus_open_failed ⇒ corpus == None` is what makes the redb lane safe, because every redb write funnels through `self.corpus` and short-circuits on `None`. `corpus == None` is precisely the condition that turns the snapshot writers **on**. The invariant that protects one lane is the trigger for the other.

So the fix is a new predicate, `CodeIndexer::refuse_durable_write`, applied at every snapshot-write chokepoint rather than at the `chunks.json` call site:

| Site | File |
|---|---|
| `save_chunks_to_disk` | `core/indexer/persist.rs` |
| `flush_corpus_to_disk` | `core/indexer/persist.rs` |
| `save_vector_store` | `core/indexer/persist_hnsw.rs` |
| `spawn_incremental_persist` | `core/indexer/persist_hnsw.rs` |

Gating the single `chunks.json` call would have left the HNSW graph exposed. `UsearchStore::save`'s own guards do not cover it: the #1711 zero-vector guard only protects an *existing* snapshot, and the #1717 shrink guard is inert below `SHRINK_GUARD_MIN_ON_DISK_VECTORS` (1 000) on-disk vectors and above a 50 % ratio — so a partially repopulated quarantined index could still shrink its graph durably.

The gate keys on `corpus_open_failed`, never on `corpus.is_none()`. A legitimate BM25-only / legacy indexer has no corpus either and must keep persisting; `unquarantined_legacy_index_still_writes_chunks_json` is the control that fails against a fix placed on the wrong predicate.

Refusals share the existing `incremental_writes_refused` counter (accessor name kept for API compatibility, doc widened) so an operator sees one number for "writes dropped while quarantined".

### Durable writes checked and found already safe

- `commit_corpus_to_redb` — early-returns on `corpus == None` (#4122).
- `hash_cache::persist_batch` — reads `indexer.corpus_store()`, returns on `None`. The reindex hash cache is in-process only for a quarantined index.
- `rewrite_vector_store_keys` (M003) — unreachable. `run_migrations` aborts at M001's `write_schema_version`, which errors with no corpus, so the chain never reaches M003. Deliberately left ungated: gating it at the top would skip the in-memory key rewrite while the framework still advanced the stamp, so M003 would never re-run after recovery.
- `reindex::staging` — `should_stage(has_corpus_store())` is `false`.

## Is `chunks.json` still read?

Yes — do not stop writing it.

- `core/indexer/migrations.rs` — `JsonCorpusToRedbMigration::apply` → `load_chunks_from_disk`, at warm boot when redb is empty and the schema stamp is `UNVERSIONED`. This is the recovery source the bug destroyed.
- `service/server/status.rs:52` — `last_indexed` falls back to its mtime when `index.redb` is absent.
- `commands/migrate_storage/migrate.rs:40` — listed among the files relocated on a storage move.

Worth recording: after this change, **production never writes `chunks.json` at all**. Every non-quarantined index has a wired corpus, so both writers take the redb branch; quarantined indexes are now refused. The remaining writers are test / BM25-only indexers. Retiring the format is a separate decision — it is still a read path, and the migration is the only thing standing between an upgraded-in-place daemon and a cold reindex.

## Documentation corrected

- `core/indexer/quarantine.rs` — the ERROR diagnostic no longer claims only that "the on-disk corpus is untouched"; it now names all three artifacts (`index.redb`, `chunks.json`, HNSW). Module docs gained a section on the two write families and a note that the load-bearing invariant protects the redb lane and *enables* the snapshot lane.
- `core/indexer/persist.rs` — `save_chunks_to_disk`'s "Empty corpus is still written so the on-disk file accurately reflects state" now carries its exception: on a quarantined index, empty means "the corpus never opened", not "the corpus is empty".
- `core/indexer/mod.rs`, `service/shutdown_flush.rs` — field and call-site docs updated to match.

## Does this make #4227 newly inaccurate?

No. #4227 is about the module doc's *trigger list* — that "genuine corruption" is named as a quarantine condition when `corpus_recovery` absorbs corruption into a recreate-to-empty before it can set `corpus_open_failed`. That sentence is untouched here, and nothing in this change alters the classification path (`corpus_recovery.rs`, `core/corpus/open_failure.rs`). Its evidence cites the module by name, not by line, so the added sections do not stale its references. #4227 remains open and accurate as filed.

## Verification

```
cargo fmt --check                                            exit 0
cargo clippy -p trusty-search --all-targets -- -D warnings   exit 0
cargo test -p trusty-search                                  exit 0
    test result: ok. 1455 passed; 0 failed; 1 ignored
    test result: ok. 349 passed; 0 failed; 3 ignored
    test result: ok. 4 passed; 0 failed; 0 ignored   (quarantine_durable_writes_4226)
    test result: ok. 3 passed; 0 failed; 0 ignored   (corpus_open_quarantine_4122)
```

The regression tests were verified to fail against current behaviour by reverse-applying the source diff (tests untouched):

```
test unquarantined_legacy_index_still_writes_chunks_json ... ok
test quarantined_shutdown_flush_does_not_destroy_chunks_json ... FAILED
test quarantined_incremental_persist_is_refused ... FAILED
test quarantined_index_refuses_hnsw_snapshot_write ... FAILED

assertion `left == right` failed: #4226: a write-quarantined index must not
overwrite the legacy chunks.json snapshot with its empty in-memory corpus
  left: 0
 right: 2
```

`left: 0, right: 2` is the data loss: the seeded snapshot came back holding nothing. The control passed in both directions. The fix was then restored and all four re-run green.

Tests pin `TRUSTY_DATA_DIR` at a scratch directory for the whole binary, so the pre-fix run could not reach the operator's live daemon data dir. No live index, data dir, or corpus was touched.

## Notes for the reviewer

- No version bump. 0.42.2 is published and no gate demanded one.
- `service/shutdown_flush.rs` overlaps conceptually with the sibling #4393 work (shutdown flush budget), in a separate worktree. My change there is a three-line comment inside `flush_one_index`'s flush future — no logic, no signature, no control flow. Either merge order works; a textual conflict, if any, resolves by keeping both comments.


bobmatnyc:

No findings.

**Reviewer-context substance, independently verified against `crates/trusty-search/src/core/indexer/quarantine.rs` and `tests/quarantine_durable_writes_4226.rs`:**
- `refuse_durable_write` keys on `self.corpus_open_failed` (line 183: `if !self.corpus_open_failed { return false; }`), never on `corpus.is_none()`.
- `unquarantined_legacy_index_still_writes_chunks_json` is a real control test: it asserts `!indexer.is_write_quarantined()` on a legacy no-corpus indexer, then asserts `chunks.json` was still written — pinning exactly the corpus_open_failed-vs-corpus.is_none() distinction called out in the PR body.
- `chunks.json` is still read by `JsonCorpusToRedbMigration::apply` → `load_chunks_from_disk` (`core/indexer/migrations.rs`) — the warm-boot recovery path is untouched; only the writers were gated.

**Also fixed before merge:** the branch's own `changelog.d/4226-quarantine-durable-write-gate.md` fragment had a category line with no bullet (CI: "has a category but no bullet"), and the branch was on trusty-search 0.42.2 which is already published on crates.io — rebasing onto current main (which already carries the 0.42.3 bump from an unrelated already-merged PR) resolved the version-parity gate without a manual bump here. Both were mechanical CI-gate fixes, not review findings; pushed as `f44e5db3`.

Merging.

PR #4867, opened 2026-08-05
fix(trusty-mpm): gate destructive index DELETEs behind a capability tests cannot hold (#4743)

### LINKED ISSUE #4743: Tests issue real cross-daemon index DELETEs against a live trusty-search using collision-prone fixture names
Found 2026-08-04 during PR #4725's round-3 review. Pre-existing pattern, not introduced by that PR.

`decommission_full_still_terminates_the_runtime` uses the workspace basename `full`. Because `derive_index_id` is a bare `file_name()`, running the suite on a machine with a live trusty-search daemon issues a **real** `DELETE /indexes/full?delete_data=true` — and `delete_data=true` destroys the on-disk data directory (trusty-search's own `delete_index_with_delete_data_true_destroys_data` asserts exactly that; its orphan reaper deliberately passes `delete_data=false`).

Nobody currently has an index named `full`, so there is no live loss. But a single-word fixture name is exactly the kind that collides eventually, and the failure would be silent destruction of a real index during a test run.

**The pattern is broader than this one test.** Any test whose fixture workspace basename can collide with a real index id has the same exposure. Worth an audit rather than renaming one string.

Suggested: use unmistakably-fixture basenames (a `test-fixture-` prefix plus a random suffix), and/or make the test harness refuse to issue index DELETEs against a non-loopback or non-fixture daemon at all.

bobmatnyc:
## Defect

The route is in `trusty-mpm`, not `trusty-search`.

`session_manager/search_gc.rs` formatted `DELETE /indexes/{id}?delete_data=true`
at two sites — `delete_search_index_best_effort:118` (decommission) and the
orphan sweep's own inline loop at `:334` — and sent it to whatever daemon
`trusty_common::resolve_daemon_base_url("trusty-search")` discovered. That
resolver reads `~/Library/Application Support/trusty-search/http_addr`, and a
test process reads it exactly like production does. On this machine that file
contains `127.0.0.1:7878`: the operator's live daemon. `?delete_data=true`
destroys the index's on-disk data directory — trusty-search's own
`delete_index_with_delete_data_true_destroys_data` asserts it, and its orphan
reaper deliberately passes `delete_data=false`.

Fixture workspaces derive their index id from a bare `file_name()`
(`derive_index_id`), so `decommission_full_still_terminates_the_runtime` asked
that daemon to destroy an index named `full`. `sess`, `live` and `proj` reach the
same path elsewhere in the suite; the operator's daemon carried a real index
named `proj` while this was written.

Measured, not inferred. With the guard removed, the regression test's counting
daemon records exactly one connection:

```
assertion `left == right` failed: a test process must reach the trusty-search
daemon zero times — a DELETE here carries ?delete_data=true and destroys a real
index named `full` (#4743)
  left: 1
 right: 0
```

## Why #4094's fix does not cover it

#4094 put a `cfg(test)` arm in trusty-search's `persistence::default_data_dir`,
so that daemon resolves an isolated data dir under test. Neither half of that
helps here:

- The delete lands in a **different process**. No compile-time guard anywhere in
  trusty-search governs what a trusty-mpm test binary puts on the wire.
- A data-dir seam cannot move a request that is already addressed to port 7878.
  Pointing the target elsewhere is the one thing a data-dir fix does not do.

## Resolution

Both destructive sites now go through one capability,
`session_manager::index_delete_guard::DestructiveIndexDelete`.

- It holds the **only** copy of the `?delete_data=true` literal, and exposes no
  constructor taking a base URL. A caller cannot build the request without
  acquiring the capability; the capability decides for itself whether the
  process may destroy data. A destructive site added later inherits the refusal
  by construction — it has to call `acquire` to get anything it can delete with.
- `acquire` refuses a `cargo test` process, checked **before** discovery so a
  test run never even looks up where the operator's daemon lives.
- The orphan sweep acquires the capability **before listing anything**, so a
  test process makes no request at all rather than enumerating the operator's
  real indexes and stopping at the delete. That loop previously had no guard of
  its own — a guard placed only on the decommission helper would have left it
  fully exposed, which is the shape PR #4725's review kept finding.

This is deliberately not a conditional at each call site. Two review rounds on
PR #4725 each found a different destructive effect a caller had failed to gate;
the response there, as here, was to make the ungated shape unrepresentable
rather than rely on every call site remembering.

The refusal itself must be a runtime check, for the cross-process reason above.

## Relationship to PR #4864

This **builds on** #4864 rather than duplicating or fighting it. It was
developed stacked on `fix-4255-registry-pollution`; that PR merged as `da1e8dcb`
partway through, so this now sits directly on main as a single commit.

#4864 establishes `trusty_common::running_under_test_harness()` in a new
`crates/trusty-common/src/test_harness.rs`. That is precisely the process-level
seam this defect needs, so this PR consumes it rather than growing a second,
drifting copy of the same detection in trusty-mpm. `TRUSTY_ALLOW_PRODUCTION_STATE=1`
carries over unchanged as the explicit opt-in.

The two are complementary halves of the same exposure. #4864 closed the
**write** paths — a fixture root registering itself in the operator's
`indexes.toml`, and `ensure_project_indexed` POSTing to the live daemon. This
closes the **destructive** path, which no data-dir or registry seam reaches.
File overlap is zero: #4864 changed `trusty-common` and `trusty-search`; this
changes only `crates/trusty-mpm/src/session_manager/`.

One finding was fed back to #4864 while stacked on it:
`register_project_index_never_bypasses_sensitive_path_denylist`
(`crates/trusty-mpm/src/core/session_launch/tests_search_index.rs`) drives
`ensure_project_indexed` and was failing on that branch, because #4864's guard
suppressed the POST it asserts on. #4864 fixed it in `c68825bc` before merging.

## Evidence

`search_gc_guard_tests.rs` asserts against a **live, accepting** loopback daemon
rather than a dead port. A dead port makes the guarded and unguarded code look
identical — both end with "no index was deleted", one because the request was
refused at the source and one because nothing was there to answer. The listener
here would have accepted the DELETE and answered 200, so a zero count is
evidence about trusty-mpm's behaviour rather than about the machine's port
allocation. The fixture self-checks that it really does answer, so a zero count
cannot come from a broken listener.

Reverting `acquire`'s `running_under_test_harness` branch fails three tests:

```
test session_manager::index_delete_guard::tests::acquire_is_refused_under_a_test_harness ... FAILED
test session_manager::search_gc_guard_tests::decommission_issues_no_request_to_a_live_daemon_under_test ... FAILED
test session_manager::search_gc_guard_tests::sweep_makes_no_request_under_a_test_harness ... FAILED
test result: FAILED. 2 passed; 3 failed; 0 ignored; 0 measured; 4530 filtered out
```

Restored, the full crate suite is green:

```
test result: ok. 4529 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 129.29s
```

`cargo clippy -p trusty-mpm --all-targets -- -D warnings`, `cargo fmt --check`,
`scripts/check_line_cap.sh` (0 violations) and `scripts/check_sld.sh` (0 errors,
0 warnings) all pass. No version bumps.

Unrelated pre-existing flake, seen across these runs and reported rather than
chased: `daemon::managed_routes::tests::stale_assets_for_many_*` fails
intermittently, a different member each time, and passes when the four are run
in isolation. They share a process-global `$HOME` override (`fake_home()`) and
race other `$HOME`-touching tests under parallel execution. That file contains
zero references to `search_gc`, `index_delete` or `delete_data`, and is not in
this diff.

No destructive path was exercised against the operator's live daemon at any
point: the revert demonstration ran with `TRUSTY_DATA_DIR_OVERRIDE` pointed at
the test's own counting daemon, which is where that one DELETE landed.

Closes #4743

### DIFF
--- a/crates/trusty-mpm/changelog.d/4743-test-index-delete-guard.md
+++ b/crates/trusty-mpm/changelog.d/4743-test-index-delete-guard.md
@@ -1,0 +1,8 @@
+Fixed
+
+- `cargo test` can no longer destroy a real trusty-search index (closes [#4743](https://github.com/bobmatnyc/trusty-tools/issues/4743))
+  - `session_manager::search_gc` formatted `DELETE /indexes/{id}?delete_data=true` at two sites and sent it to whatever daemon `resolve_daemon_base_url` discovered — under a test run, the operator's live one on port 7878. `?delete_data=true` destroys the index's on-disk data directory
+  - fixture workspaces derive their index id from a bare `file_name()`, so `decommission_full_still_terminates_the_runtime` asked that daemon to destroy an index named `full`; `sess`, `live` and `proj` appear the same way elsewhere in the suite
+  - both sites now go through one `DestructiveIndexDelete` capability that holds the only copy of the `?delete_data=true` literal and exposes no constructor taking a base URL, so a caller cannot build the request without acquiring it — and a `cargo test` process never can
+  - the orphan sweep acquires the capability before listing anything, so a test process makes no request at all instead of enumerating real indexes and stopping at the delete
+  - `TRUSTY_ALLOW_PRODUCTION_STATE=1` remains the explicit opt-in for a test that deliberately drives a real daemon
--- a/crates/trusty-mpm/src/session_manager/decommission.rs
+++ b/crates/trusty-mpm/src/session_manager/decommission.rs
@@ -585,7 +585,7 @@ impl SessionManager {
     /// | 1 | `graceful_terminate_runtime` — SIGTERM + `kill_session` the pane | no |
     /// | 2 | `remove_session_worktree` — `git worktree remove --force`, `fs::remove_dir_all` fallback, `git worktree prune`, `git branch -D` (dirty-gated, #4344) | no |
     /// | 3 | `fs::remove_dir_all` on an SM-owned workspace (containment-gated) | no |
-    /// | 4 | `delete_search_index_best_effort` — cross-daemon `DELETE /indexes/{id}` | no |
+    /// | 4 | `delete_search_index_best_effort` — cross-daemon `DELETE /indexes/{id}` (never from a test process, #4743) | no |
     /// | 5 | clears `workspace_path`/`workspace_owned` when nothing is on disk | store-only |
     /// | 6 | clears `pending_decision`/`proposed_default` (#4400) | store-only |
     /// | 7 | writes state `Decommissioned` | store-only |
--- a/crates/trusty-mpm/src/session_manager/index_delete_guard.rs
+++ b/crates/trusty-mpm/src/session_manager/index_delete_guard.rs
@@ -1,0 +1,246 @@
+//! The one door a destructive trusty-search index DELETE may pass through (#4743).
+//!
+//! Why: `search_gc` had two independent sites that formatted
+//! `DELETE /indexes/{id}?delete_data=true` — the decommission-time removal and
+//! the orphan sweep's delete loop. `?delete_data=true` destroys the index's
+//! on-disk data directory (trusty-search's own
+//! `delete_index_with_delete_data_true_destroys_data` asserts exactly that;
+//! its orphan reaper deliberately passes `delete_data=false`). The daemon those
+//! requests reach is whichever one `resolve_daemon_base_url` discovers, and
+//! under `cargo test` that is the OPERATOR'S live daemon: the address comes from
+//! `~/Library/Application Support/trusty-search/http_addr`, which a test process
+//! reads exactly like production does. Fixture workspaces derive their index id
+//! from a bare `file_name()` — `full`, `sess`, `live` — so a real index sharing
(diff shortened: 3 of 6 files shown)


bobmatnyc:

No findings.

**Reviewer-context substance, independently verified against `crates/trusty-mpm/src/session_manager/index_delete_guard.rs` and `search_gc.rs`:**
- `DestructiveIndexDelete` holds the single `?delete_data=true` literal in the whole crate (`delete_url`, line 147); every other occurrence in the diff is a comment or test string.
- Both fields (`base`, `client`) are private; the only path to an instance outside the test module is `DestructiveIndexDelete::acquire()`. There is no constructor taking a base URL.
- `sweep_orphaned_search_indexes` acquires the capability (line 291-298) *before* `fetch_index_details` (the listing call, line 306) — the previously-unguarded loop now refuses before it enumerates anything.

Merging.

Issue #1182, opened 2026-06-14
trusty-search: background daemon self-spawn does not forward --data-dir (silently ignored when TRUSTY_DATA_DIR pre-set)


bobmatnyc:
## Context
Surfaced by trusty-review on PR #1180 (the #1178 commands/start split). PRE-EXISTING logic moved verbatim by the refactor — not introduced by it.

## Problem
In `crates/trusty-search/src/commands/start/daemon.rs`, the `!foreground` (background) path builds the child `Command` with `--port`, `--device`, and optionally `--no-auto-discover`, but never appends `--data-dir`. It relies on the child inheriting `TRUSTY_DATA_DIR` from the environment. But that env var is only set by `handle_start` when it was previously unset (`if std::env::var_os("TRUSTY_DATA_DIR").is_none()`). If `TRUSTY_DATA_DIR` was already set in the environment to a DIFFERENT value than the `--data-dir` flag the user passed, the child inherits the stale env value and the user's `--data-dir` flag is silently ignored.

## Suggested fix
Always forward `--data-dir` as an explicit CLI arg to the spawned child (so the flag wins over any inherited env var), OR make `--data-dir` override `TRUSTY_DATA_DIR` unconditionally before spawn. Add a regression test for the pre-set-env case.

Labels: bug

Issue #81, opened 2026-05-23
[trusty-search] community_cohesion score is 0.0 on all results — knowledge graph not contributing


bobmatnyc:
## Summary

Every search result in re-testing returned `community_cohesion: 0.0`, indicating the knowledge graph component is not contributing to result scoring.

## Evidence

All 13 test scenarios showed the same pattern:
```json
{
  "score": 0.0143,
  "community_cohesion": 0.0,
  "match_reason": "vector"
}
```

## Impact

- Knowledge graph relationships (file → function → caller, module → import) are not influencing result ranking
- Cross-file concept queries ("what calls the agent subprocess") would benefit most from graph-based scoring
- The system is operating as pure vector search, losing the graph-augmented ranking advantage

## Fix criteria

- At least some results for code-navigation queries should show `community_cohesion > 0.0`
- Graph edges should be built during indexing and reflected in scores at query time

bobmatnyc:
**Tested against**: trusty-search `v0.4.4`
Confirmed via: `trusty-search --version` → `trusty-search 0.4.4` and `curl http://127.0.0.1:7878/health` → `{"version":"0.4.4",...}`

bobmatnyc:
Fixed in f64f25a — GraphScorer cache invalidated after community detection completes. Closing.

PR #4868, opened 2026-08-05
fix(trusty-search): make the shutdown window achievable and stop the reaper killing by name

### LINKED ISSUE #4393: Restart safety: no `ExitTimeOut` in rendered plists, and trusty-search's orphan reaper is lockfile- and data-dir-blind

### LINKED ISSUE #4395: trusty-search's orphan reaper matches by process name, ignoring lockfile and data dir — can SIGKILL a healthy production daemon
## Summary

`trusty-search start` runs an orphan reaper that SIGTERMs **every process named `trusty-search` with `start` in argv** — matching on process name, not on lockfile ownership or data directory — then SIGKILLs survivors after **3 seconds** and deletes their lock and port files.

Any asymmetry between the starter's lock path and a running daemon's turns a routine `start` into the destruction of a healthy production daemon, with a tenth of that daemon's own flush budget to shut down.

## Mechanism

- `find_daemon_pids()` matches purely on executable name plus `start` in argv (`commands/stop.rs:133-161`). Its own comment is explicit: *"Collect every live trusty-search daemon process, regardless of whether it matches the lockfile."*
- The reaper (`commands/start/daemon.rs:261-311`) SIGTERMs that set, SIGKILLs at **3 s** (`:281`), then removes lock and port files.
- The lockfile fast-path that would prevent this (`running_daemon_pid()`, `start/daemon.rs:248-256`) is **data-dir-local** (`daemon_dir()`, `service/daemon.rs:229-249`) while the reaper is **machine-global**.

## Scoping — when is it actually reachable?

In the common path it is **not**: shared data dir, healthy daemon, live lockfile → `running_daemon_pid()` bails before the reaper runs. Reaching it requires a lock-visibility asymmetry:

- a second instance started with `--data-dir` / `TRUSTY_DATA_DIR`
- a daemon started *with* `TRUSTY_DATA_DIR` but restarted without it, or vice versa — the one production-relevant edge
- a deleted or stale lockfile

So this is primarily a developer/operator footgun rather than an every-restart hazard. It is nonetheless the reason a second trusty-search cannot safely be started for testing on a machine with a live daemon — encountered directly during the #4246 work, where it blocked an entire class of integration test.

## Second, independent defect: the shared HNSW staging path

`hnsw.usearch` has no cross-process lock, and **both writers use the same deterministic `hnsw.usearch.tmp`** (`core/store/usearch_store.rs:516`) before renaming. Two concurrent writers can interleave and rename a garbage snapshot into place.

The daemon lock does not close this, because **colocated indexes store `hnsw.usearch` in the project root** (`<repo>/.trusty-search/`), *outside* any data directory — so even two perfectly data-dir-isolated daemons indexing the same repo share the staging path.

In practice the corridor is narrow (the reaper usually kills one writer within 3 s, and most garbage outcomes are caught at next load by the #2922/#3970 guards — detected loss, not silent wrong results), but the fix is cheap and unconditional.

Related, narrower: the stale-lock removal has a TOCTOU (`service/daemon.rs:400-421`, `:442-446`) where remove-and-retry can leave two processes holding exclusive locks on different inodes. Requires simultaneous starts against a stale lockfile.

## Fix

1. **Make the reaper ownership-aware** — reap only processes whose data dir matches and whose lock we own — or gate it out of the auto-spawn path entirely. This is the primary fix and resolves the healthy-daemon kill.
2. **Give the HNSW save a per-process tmp name** (or a cross-process lock). Prefer the tmp rename: it is cheap, unconditional, and necessary regardless of daemon locks because colocated stores sit outside every lock's protection.
3. **Raise the 3 s SIGKILL** to something compatible with the ≥30 s/index flush budget — secondary once the reaper stops targeting non-owned daemons. See #4393 for the same problem on the launchd and `stop` paths.

Not recommended: making `daemon_dir()` honour `TRUSTY_DATA_DIR_OVERRIDE`. That variable is trusty-common's **test-isolation** override (`data_dir.rs:104-121`); trusty-search's documented isolation mechanism is `--data-dir` / `TRUSTY_DATA_DIR`, which *is* honoured for lock, port, `daemon.env` and (since #3545) `http_addr`. The isolation failure is not the env var — it is that the reaper ignores every isolation mechanism equally, because it matches by process name. A docs clarification is sufficient there.

Separately worth considering: `TRUSTY_DATA_DIR` is a single presence-checked switch that both isolates the data dir **and** suppresses the shared `http_addr` discovery write (`service/daemon.rs:622-629`). Splitting the suppression onto its own opt-out (e.g. `TRUSTY_SEARCH_NO_SHARED_DISCOVERY`) would make a safely-isolated second instance possible for the first time.

## Severity

**P2.** The mechanism is real and verified, and the consequence is destruction of a live daemon plus a genuine (if narrow) corruption vector. Held below P1 because the common path is protected by the lockfile fast-path, and reaching the failure requires a lock-visibility asymmetry that is mostly self-inflicted. Would warrant P1 if any shipped automation is shown to invoke `start` with mismatched env against a running daemon — no such caller was found.

## Provenance

Split out of #4393 after an adversarial review established that its three parts shared an outcome but not a fix surface. Verified read-only against `main`. Related: #4393, #4390, #4246, #4122, #3545.

bobmatnyc:
## What was broken

One failure story with two halves. trusty-search's shutdown flush budgets **≥30 s per index** and up to 20 min; every window that actually terminates the daemon granted **3–5 s**. So a flush that had real work to do was SIGKILLed before finishing, on every path. And one of those windows — the orphan reaper's 3 s — was being applied to daemons the reaper had no business killing at all, because it matched them by **process name**.

## The termination windows, and where each is written down

| Path | Before | After | Source |
|---|---|---|---|
| launchd `ExitTimeOut` | **5 s** (measured; no key rendered, so launchd's "system-defined" default applied) | 60 s | `trusty-common/src/launchd.rs` `render_plist` |
| `trusty-search stop` | 5 s | `termination_grace()` | `trusty-search/src/commands/stop.rs` |
| orphan reaper inside `start` | 3 s | `termination_grace()` | `trusty-search/src/commands/start/reap_orphans.rs` `reap` |
| flush floor, per index | 30 s | 30 s, clamped to the remaining window | `service/shutdown_flush.rs` `MIN_FLUSH_TIMEOUT_SECS` |

The 5 s launchd figure is #4393's own measurement, reproduced there on a throwaway LaunchAgent (`kickstart -k` → SIGTERM at t+14 ms, new pid at t+5.03 s, the trap's post-handler line never printed). I did not re-measure it and did not touch launchd on this host — see Safety below. No other killer exists in the codebase: the only `SIGKILL` sites in the workspace are these three plus `stop`'s phase-3 escalation, which is the same window now widened.

## Is ≥30 s inherent, or incidental?

Incidental, in the sense that matters. `MIN_FLUSH_TIMEOUT_SECS` is a *ceiling granted*, not a duration *required* — the typical flush is milliseconds (`UsearchStore::save`'s view-mode skip, plus per-batch redb durability). The real defect was not that 30 s is too long; it is that **nothing in the flush path knew how long the process would live**. `shutdown_flush_deadline_for` could return 20 minutes to a process with 5 seconds left, spend all 5 on index #1, and never reach #2..N.

So the durable fix is structural rather than a bigger number, and it is the same shape as the resume-checkpoint work in #4721/#4768:

- `ShutdownBudget` counts down from the instant SIGTERM landed (recorded in the graceful-shutdown future, so the axum drain and watcher teardown are charged against it, not ignored).
- `FlushDeadline` is the only type `flush_index_bounded` accepts. Its field is private and `shutdown_budget` exposes no constructor from a bare `Duration`, so `shutdown_flush` — a different module — **cannot write a deadline the process will not live to honour**. Every grant is `min(size-scaled, remaining)`.
- An exhausted budget mints `None`, so the sweep stops at an **index boundary** rather than being cut off mid-write.

What an interrupted sweep now costs: nothing beyond the last checkpoint. `save` stages and renames, so a flush never started and a flush abandoned leave identical on-disk state — whatever the incremental persister last published. Stopping at a boundary is what turns "SIGKILLed at an arbitrary point" into "N flushed, M kept their checkpoint", logged with the count.

## The reaper's identity signal: the data directory

Process name says nothing about whose data a daemon serves. The candidates I weighed:

- **A pid file** names the daemon we already track — by definition not the orphan we are hunting (`running_daemon_pid()` has already returned `None` by the time the reaper runs).
- **A launchd label** does not exist for the CLI-detached daemons this reaper actually finds (PPID 1, no plist).
- **Port ownership** — the #4470 shape, `trusty-installer::commands::port_guard`, which I read and modelled the fail-closed structure on — answers "is someone on my port?". Right question for a bootstrap; wrong one here. The daemon *walks* its port on collision, so a foreign holder is a condition it is designed to survive, and a different port proves nothing about whose data dir a process has open.
- **A `/health` handshake** identifies a daemon that is *answering* — which excludes exactly the wedged instance the reaper exists to clear.

**The data directory** is what a trusty-search daemon actually contends over: one lock, one set of index files. It declares that directory in its own argv (`--data-dir`, always passed explicitly by the self-spawn per #1182) or its own environment (`TRUSTY_DATA_DIR`), falling back to the platform default when it declares neither. Two daemons on disjoint data dirs were never in conflict.

Structurally, per the #4718/#4835 pattern rather than "check a bit harder":

- `DaemonIdentity` is a three-state enum — `OwnInstance` / `ForeignInstance(dir)` / `Unidentified(why)`. The third state is the one a `bool` folds into "kill it", which is how a name match became a death sentence. An **unreadable** environment is indistinguishable from an empty one, so it is `Unidentified`, never resolved to the default dir.
- `ConfirmedOrphan` wraps a pid behind a private field with no constructor from a `u32`. `reap` accepts nothing else. **There is no way to express "signal this pid" for a pid that did not come out of `plan` positively matched to our own data dir.**

`trusty-search stop` deliberately keeps its stop-everything semantics: it is an explicit operator command whose contract since #81 is "nothing named trusty-search is left running". The *implicit* reaper is the one that had to change, because nobody asked it to kill anything.

Also fixed, #4395's second defect: both writers staged through the same `hnsw.usearch.tmp`, and colocated indexes keep that file in the project root — outside every data dir, so even fully data-dir-isolated daemons collided there. The staging name is now process-scoped.

## Does an installed unit need regeneration?

**Yes.** `ExitTimeOut` lives in the plist, not the binary. An already-installed LaunchAgent keeps launchd's 5 s default until its plist is re-rendered and re-bootstrapped — upgrading the binary alone does not pick it up. Operators re-run the installer's service setup. In the meantime the daemon does not lose more than it did before: with a 5 s real window it flushes fewer indexes than it plans to, which is exactly the pre-fix outcome, minus the mid-write kill. `TRUSTY_TERMINATION_GRACE_SECS` lets a host whose supervisor window cannot be raised declare the true number.

**The owner's live unit was not regenerated and no `launchctl` command was run against it.**

## Safety

No live daemon was touched. Everything here is verified through pure unit tests over seams — `identify`/`plan` take argv/environ/data-dir as parameters, `ShutdownBudget` takes an explicit window, `staging_path` is a pure path function. No daemon was started, no port was bound, no signal was sent, and `launchctl` was never invoked. The owner's trusty-search 0.42.2 was left alone.

## Regression tests, demonstrated by reverting

Each fix was reverted in place, the tests run, and the fix restored.

**#4395 — reverted `plan` to confirm every candidate (the pre-fix "bare pid list IS the kill list"):**
```
test commands::start::reap_orphans::tests::plan_spares_an_unidentifiable_candidate ... FAILED
test commands::start::reap_orphans::tests::plan_confirms_only_our_own_data_dir ... FAILED
assertion `left == right` failed: only the daemon on our own data dir may be reaped — pid 20 is a healthy production daemon and pid 30 could not be identified (#4395)
test result: FAILED. 9 passed; 2 failed; 0 ignored; 0 measured; 353 filtered out
```

**#4393 — reverted `flush_deadline_for` to hand the size-scaled deadline through unclamped:**
```
test service::shutdown_budget::tests::flush_deadline_is_clamped_to_the_remaining_window ... FAILED
test service::shutdown_budget::tests::exhausted_budget_mints_no_deadline ... FAILED
test service::shutdown_flush::tests::shutdown_flush_skips_every_index_when_the_window_is_spent ... FAILED
test service::shutdown_flush::tests::flush_bounded_returns_on_blocking_work ... FAILED
test service::shutdown_flush::tests::flush_bounded_survives_single_worker_when_blocking_work_is_offloaded ... FAILED
test result: FAILED. 11 passed; 5 failed; 0 ignored; 0 measured; 1448 filtered out
```

**#4393 — removed the `ExitTimeOut` key from the renderer:**
```
test launchd::tests::render_plist_declares_exit_timeout ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 256 filtered out
```

**#4395 — reverted the staging name to the shared `hnsw.usearch.tmp`:**
```
test core::store::tests::test_staging_path_is_process_scoped ... FAILED
staging name must be scoped to this process; got hnsw.usearch.tmp
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1463 filtered out
```

No coverage was weakened to reach green: no `#[ignore]`, no `cfg`-gating, no `--exclude`. The one visibility widening (`MIN_FLUSH_TIMEOUT_SECS` from `pub(crate)` to `pub`) exists because the `trusty-search` bin re-exports the lib's modules into its own `crate::` namespace, so the CLI-side window assertions are a different crate to the compiler.

## Gates

```
cargo test -p trusty-search
test result: ok. 1463 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 44.05s   (lib)
test result: ok. 361 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 0.33s     (bin)
  + 12 integration binaries, all ok, 0 failed

cargo test -p trusty-common
test result: ok. 251 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 0.74s

cargo test -p trusty-installer
test result: ok. 551 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 2.03s

cargo clippy -p trusty-search -p trusty-common -p trusty-installer --all-targets -- -D warnings   exit 0
cargo fmt --check                                                                                 exit 0
scripts/check_line_cap.sh            3660 files, 7 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh       21198 citations, 0 dangling — OK
scripts/check_sld.sh                 55 specs + 3067 code files, 0 errors, 0 warnings
scripts/check_doc_numbers.sh         98 docs / 92 claims, 0 violations — OK
scripts/check_changelog_fragment.sh  all 3 crates with source changes recorded
```

No version bumps — 0.42.2 is published and no gate asked for one.

Closes #4393
Closes #4395

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11578,7 +11578,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-installer"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "clap",
--- a/crates/trusty-common/changelog.d/4393-launchd-exit-timeout.md
+++ b/crates/trusty-common/changelog.d/4393-launchd-exit-timeout.md
@@ -1,0 +1,12 @@
+Fixed
+
+- Generated LaunchAgent plists now declare `ExitTimeOut`. Without the key
+  launchd applies its "system-defined" default, which measures 5 s on macOS —
+  SIGTERM, then SIGKILL 5 s later on every `launchctl bootout`, `kickstart -k`,
+  logout and reboot. That is shorter than the shutdown work several trusty-*
+  daemons do; trusty-search's index flush alone floors at 30 s per index, so it
+  was cut off mid-write every time. The rendered window comes from the new
+  `shutdown::TERMINATION_GRACE_SECS` (60 s), which `trusty-search stop` and its
+  orphan reaper now wait as well, so the window a daemon plans for and the
+  window its terminator grants cannot drift apart. **Already-installed agents
+  keep launchd's 5 s default until their plist is regenerated** (#4393)
--- a/crates/trusty-common/src/launchd.rs
+++ b/crates/trusty-common/src/launchd.rs
@@ -570,10 +570,15 @@ mod tests {
             )),
             "ExitTimeOut must carry the shared termination grace, got:\n{xml}"
         );
-        assert!(
-            crate::shutdown::TERMINATION_GRACE_SECS > 5,
-            "rendering the measured system default would leave #4393 unfixed"
-        );
+        // `const` block: the operands are compile-time constants, so a later
+        // edit that shortens the window back toward launchd's default fails the
+        // build rather than a test run.
+        const {
+            assert!(
+                crate::shutdown::TERMINATION_GRACE_SECS > 5,
+                "rendering the measured system default would leave #4393 unfixed"
+            );
+        }
     }
 
     #[test]
--- a/crates/trusty-common/src/launchd.rs
+++ b/crates/trusty-common/src/launchd.rs
@@ -130,10 +130,13 @@ impl LaunchdConfig {
     /// (exe + args), `KeepAlive` (per [`KeepAlive`]), `ThrottleInterval`,
     /// `RunAtLoad`, `StandardOutPath`/`StandardErrorPath` under `log_dir`,
     /// `SoftResourceLimits` + `HardResourceLimits` dicts (when `fd_limit` is
-    /// `Some`), and an `EnvironmentVariables` dictionary when `env_vars` is
-    /// non-empty. All string values are XML-escaped.
+    /// `Some`), `ExitTimeOut` (#4393, from
+    /// [`crate::shutdown::TERMINATION_GRACE_SECS`]), and an
+    /// `EnvironmentVariables` dictionary when `env_vars` is non-empty. All
+    /// string values are XML-escaped.
     /// Test: `render_plist_contains_core_keys`, `render_plist_keepalive_*`,
-    /// `render_plist_escapes_xml`, `render_plist_includes_resource_limits`.
+    /// `render_plist_escapes_xml`, `render_plist_includes_resource_limits`,
+    /// `render_plist_declares_exit_timeout`.
     pub fn render_plist(&self) -> Result<String> {
         let exe = self
             .exe_path
@@ -186,6 +189,21 @@ impl LaunchdConfig {
         s.push_str(&format!(
             "  <integer>{}</integer>\n",
             self.throttle_interval
+        ));
+
+        // #4393: without this key launchd applies its "system-defined" default,
+        // measured at 5 s on macOS — SIGTERM, then SIGKILL 5 s later. Every
+        // termination path routes through it: `launchctl bootout`, `launchctl
+        // kickstart -k` (measured: SIGTERM first, SIGKILL at exactly the
+        // ExitTimeOut boundary), and logout/reboot. trusty-search's shutdown
+        // flush floors each index at 30 s, so it was cut off mid-write on all of
+        // them. Rendered from the shared constant rather than a per-agent field
+        // so the window a daemon PLANS for and the window launchd GRANTS cannot
+        // disagree.
+        s.push_str("  <key>ExitTimeOut</key>\n");
+        s.push_str(&format!(
+            "  <integer>{}</integer>\n",
+            crate::shutdown::TERMINATION_GRACE_SECS
         ));
 
         s.push_str("  <key>StandardOutPath</key>\n");
@@ -526,6 +544,35 @@ mod tests {
         assert!(
             !xml_no_limits.contains("HardResourceLimits"),
             "fd_limit=None must suppress HardResourceLimits"
+        );
+    }
+
+    /// Why (#4393 — the regression this test exists for): the renderer emitted
+    /// no `ExitTimeOut` key at all, so launchd applied its "system-defined"
+    /// default, measured at 5 s on this host. A daemon whose shutdown flush
+    /// floors at 30 s per index was therefore SIGKILLed mid-write on every
+    /// launchd-mediated termination — `bootout`, `kickstart -k`, logout.
+    /// What: asserts the key is present, carries
+    /// [`crate::shutdown::TERMINATION_GRACE_SECS`], and that the rendered
+    /// window clears the measured 5 s default it replaces.
+    /// Test: itself.
+    #[test]
+    fn render_plist_declares_exit_timeout() {
+        let xml = sample(KeepAlive::Always).render_plist().unwrap();
+        assert!(
+            xml.contains("<key>ExitTimeOut</key>"),
+            "plist must declare ExitTimeOut or launchd applies its 5 s default (#4393)"
+        );
+        assert!(
+            xml.contains(&format!(
+                "<key>ExitTimeOut</key>\n  <integer>{}</integer>",
+                crate::shutdown::TERMINATION_GRACE_SECS
+            )),
+            "ExitTimeOut must carry the shared termination grace, got:\n{xml}"
+        );
+        assert!(
+            crate::shutdown::TERMINATION_GRACE_SECS > 5,
+            "rendering the measured system default would leave #4393 unfixed"
         );
     }
 
--- a/crates/trusty-common/src/shutdown.rs
+++ b/crates/trusty-common/src/shutdown.rs
@@ -155,14 +155,20 @@ mod tests {
     /// Test: itself.
     #[test]
     fn termination_grace_clears_the_measured_launchd_default() {
-        assert!(
-            super::TERMINATION_GRACE_SECS > 5,
-            "the grace window must exceed launchd's measured 5 s ExitTimeOut default"
-        );
-        assert!(
-            super::TERMINATION_GRACE_SECS >= 30,
-            "the grace window must cover trusty-search's 30 s per-index flush floor"
-        );
+        // `const` blocks: both operands are compile-time constants, so the
+        // check fires at build time rather than waiting for the test binary.
+        const {
+            assert!(
+                super::TERMINATION_GRACE_SECS > 5,
+                "the grace window must exceed launchd's measured 5 s ExitTimeOut default"
+            );
+        }
+        const {
+            assert!(
+                super::TERMINATION_GRACE_SECS >= 30,
+                "the grace window must cover trusty-search's 30 s per-index flush floor"
+            );
+        }
     }
 
     /// Why (#4393): a host that cannot raise its supervisor's real window has
--- a/crates/trusty-common/src/shutdown.rs
+++ b/crates/trusty-common/src/shutdown.rs
@@ -15,6 +15,71 @@
 //! `raise(SIGTERM)`, which is unsafe; the integration tests in trusty-search
 //! exercise the full axum `with_graceful_shutdown` path.
 
+/// Seconds a trusty-* daemon is granted between SIGTERM and SIGKILL (#4393).
+///
+/// Why: this is the ONE number every terminator and every terminated daemon has
+/// to agree on, and before #4393 nobody stated it. launchd's `ExitTimeOut`
+/// default is documented only as "system-defined" and measures **5 s** on
+/// macOS; `trusty-search stop` allowed 5 s; its orphan reaper allowed 3 s.
+/// Meanwhile trusty-search's shutdown flush floors each index's budget at 30 s
+/// (`service::shutdown_flush::MIN_FLUSH_TIMEOUT_SECS`), so no flush that had
+/// real work to do could ever run to completion — it was SIGKILLed mid-write on
+/// every path. Publishing the window as a shared constant is what lets the
+/// plist renderer, the CLI stop path, the reaper, and the daemon's own flush
+/// planner be checked against each other instead of drifting silently.
+///
+/// What: 60 s. Deliberately modest — long enough to clear the 30 s per-index
+/// floor, short enough that a wedged daemon does not stall reboot or
+/// `launchctl bootout` for minutes. Not derived from the flush budget's
+/// 20-minute ceiling: a multi-minute `ExitTimeOut` in a static plist trades one
+/// operator-visible failure for a worse one.
+///
+/// Test: `termination_grace_clears_the_measured_launchd_default`,
+/// `launchd::tests::render_plist_declares_exit_timeout`, and (in trusty-search)
+/// `shutdown_budget_tests::flush_floor_fits_the_termination_window`.
+pub const TERMINATION_GRACE_SECS: u64 = 60;
+
+/// Operator override for [`TERMINATION_GRACE_SECS`].
+///
+/// Why: a host whose launchd `ExitTimeOut` cannot be raised (a stale installed
+/// plist, a container supervisor with its own `TimeoutStopSec`) needs to tell
+/// the daemon the truth about its window, or the daemon plans a 55 s flush
+/// inside a 5 s life and loses more than it would have with an accurate,
+/// smaller budget. The env var is how the terminator declares the real number.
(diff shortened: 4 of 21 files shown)


bobmatnyc:

5 findings, all low-confidence (0.1-0.7) style/portability nits with no correctness impact (unconventional `use` placement, a Windows-path-literal test fragility, a ~100ms grace-window overshoot, a 15-char Linux `comm` truncation note that doesn't affect the 13-char `trusty-search` binary name). None block.

**Reviewer-context substance, independently verified against `crates/trusty-search/src/service/shutdown_budget.rs` and `crates/trusty-search/src/commands/start/reap_orphans.rs`:**
- `FlushDeadline(Duration)` — tuple field has no `pub`, so it's private to `shutdown_budget`. Its only construction site in the whole crate is `shutdown_budget.rs:143`, inside `ShutdownBudget::flush_deadline_for` itself; `shutdown_flush.rs` (the calling module) has no way to mint one directly, only `as_duration()` to read it back.
- `ConfirmedOrphan(u32)` — same shape: private field, only reader is `pid()`, only construction site is `reap_orphans.rs:237` inside `plan`, gated behind a `DaemonIdentity::OwnInstance` match arm. There is no path from a bare pid to a signal-able `ConfirmedOrphan` without going through that identity check.

**Also fixed before merge:** the branch touched `trusty-installer/src` (doc-comment corrections only) under a version already published on crates.io; bumped `trusty-installer` 0.5.0 → 0.5.1 to clear the version-parity gate (mechanical, not a review finding), pushed as `e44760e6`.

Merging.

PR #4874, opened 2026-08-05
fix(trusty-mpm): corroborate a removed worktree against its worktree root, not its parent (#4872)

### LINKED ISSUE #4872: dead session records with a fully-removed worktree can never auto-prune (workspace_verified_gone parent guard)
## Defect

`tm ls` lists 4 dead session records that can never clear. `workspace_verified_gone` in [crates/trusty-mpm/src/.../session_picker_prune.rs:373-382](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/session/session_picker_prune.rs#L373-L382) returns `false` unless the workspace path's IMMEDIATE PARENT still exists on disk. `git worktree remove` deletes the whole tree at once, so leaf and parent vanish together — indistinguishable from the unmounted-volume case the guard was written to protect against (`try_exists` returns `Ok(false)`, not `Err`, under a missing mount point; see the doc comment at [lines 346-356](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/session/session_picker_prune.rs#L346-L356)). `is_dead_record` at [line 493-495](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/session/session_picker_prune.rs#L493-L495) therefore returns `false`, the record lands in `kept`, and no marker is ever written to `~/.trusty-mpm/auto-prune-seen.json` — so repeated `tm ls` calls can never advance it.

## Evidence

Observed 2026-08-05 on trusty-mpm 1.3.5, built from origin/main b4d2375b28, with the [#4702](https://github.com/bobmatnyc/trusty-tools/issues/4702) fix already live:

- 4 stuck records: `tm-deadrt40493-01`, `tm-deadrt90908-01`, `tm-deadrt7897-01`, `tm-deadrt55373-01`.
- Their `workspace_path`s point into `.claude/worktrees/critic-4850/crates/trusty-mpm`, `.base/.worktrees/qa-4850/crates/trusty-mpm`, and two under `.base/.worktrees/2eb72dca-.../qa-4765-r3` and `qa-4796`.
- Verified on disk: leaf, parent AND grandparent are gone for all four — e.g. `.claude/worktrees/critic-4850` itself does not exist.
- All four are absent from `~/.trusty-mpm/auto-prune-seen.json`, confirming no marker is ever written.
- Distinct from the 6 `tm-deadrt*` records with `workspace_path: None` / `cwd: "/unknown"`, which prune correctly but are rate-limited by `SIGHTING_MIN_AGE_SECS` (10 min) and `AUTO_PRUNE_CAP` (5/call). Those are not part of this defect.

## Scope

This is a gap in [#4702](https://github.com/bobmatnyc/trusty-tools/issues/4702) / [PR #4725](https://github.com/bobmatnyc/trusty-tools/pull/4725)'s own fix, NOT a recurrence of [#4384](https://github.com/bobmatnyc/trusty-tools/issues/4384). Neither issue's scope anticipated it.

## Proposed Resolution

Walk up to the nearest known worktree-root ancestor (`.base/.worktrees` or `.claude/worktrees`) instead of checking only the immediate parent. That ancestor surviving means the worktree was removed (verified dead); that ancestor also being gone means a volume is unmounted (keep). Preserves the original guard's intent without loosening it.

## Workaround Note

`tm session prune --state stopped` is NOT a safe workaround — [prune.rs:397](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/bin/tm/commands/session/session_picker_prune.rs#L397) is a bare state check with no workspace verification feeding the full `decommission`, so it would also catch the 15 legitimately stopped-but-resumable records whose worktrees are intact.

bobmatnyc:
Closes #4872

## Defect

`workspace_verified_gone` returned `false` unless the workspace path's IMMEDIATE PARENT still existed on disk. `git worktree remove` deletes the whole tree in one operation, so the leaf, its parent and its grandparent vanish together — indistinguishable to a parent-only probe from the unmounted-volume case the guard was written to protect against (`try_exists` answers `Ok(false)`, not `Err`, under a missing mount point).

`is_dead_record` therefore returned `false` forever, the record landed in `kept`, and no marker was ever written to `~/.trusty-mpm/auto-prune-seen.json` — so repeated `tm ls` calls could never advance it. Four records observed stuck on 2026-08-05: `tm-deadrt40493-01`, `tm-deadrt90908-01`, `tm-deadrt7897-01`, `tm-deadrt55373-01`.

## Resolution

Absence is now corroborated against the nearest surviving worktree ROOT — `.base/.worktrees` or `.claude/worktrees` — instead of the immediate parent. That directory is not itself part of any worktree, so it survives every `git worktree remove` and disappears only when the filesystem holding it is genuinely unreachable.

- worktree root still exists → the tree below it was deleted → verified gone, prune it
- the path has such an ancestor but none of them exist → the checkout is unreachable → keep
- the path has no worktree-root ancestor at all → the original immediate-parent rule, unchanged

`tm session prune`'s manual path is untouched.

## The guard's original purpose is preserved, and pinned

`auto_prune_keeps_record_when_the_worktree_root_itself_is_gone` puts the whole checkout under a directory that does not exist, while the tempdir ABOVE it does — standing in for `/Volumes`, which survives every unmount. A "walk up until some ancestor exists" implementation would pass the positive test and mass-tombstone here; this test rules it out. `auto_prune_keeps_record_whose_parent_directory_is_unreachable` and `auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable` still pass unchanged.

## Tests

| Test | Before | After |
|---|---|---|
| `auto_prune_clears_record_whose_worktree_root_survived_the_removal` | FAILED (`pending: 0`, no marker written) | ok |
| `auto_prune_keeps_record_when_the_worktree_root_itself_is_gone` | ok (regression guard) | ok |

Both cover the two observed shapes: `.claude/worktrees/critic-4850/crates/trusty-mpm` and `.base/.worktrees/qa-4850/crates/trusty-mpm`.

Fail-before, on `origin/main`'s `session_picker_prune.rs`:

```
test tests_behavior_d::auto_prune_clears_record_whose_worktree_root_survived_the_removal ... FAILED
assertion `left == right` failed: worktrees: the record must be RECORDED as dead — a record
that never reaches `pending` never gets a marker, which is why #4872's four records could not
advance across repeated `tm ls` calls
  left: 0
 right: 1
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 1437 filtered out
```

## Gates — Rust Test Ladder rung 5 (process lifecycle)

This predicate gates session decommission, which force-removes worktrees and deletes branches.

| Command | Result |
|---|---|
| `cargo test -p trusty-mpm` | EXIT=0 — 4572 passed, 0 failed, 7 ignored (lib); 1438 passed ×2 (bin `tm`); all other targets ok |
| `cargo test -p trusty-mpm -- --include-ignored` | 2 pre-existing environmental failures, see below |

### Pre-existing `--include-ignored` failures — not caused by this branch

The diff touches three files, all under `crates/trusty-mpm/src/bin/tm/` plus a changelog fragment. It touches zero library files, so the `--lib` test binary is byte-identical to `origin/main`.

- `meta_demo_e2e::meta_run_demo_writes_and_verifies_artifact` — reproduced identically on this worktree with the branch's changes stashed: `launch_outcome: "timed-out"` after 187s. It launches a real Claude Code session; it times out on this machine regardless of the diff.
- `daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many` — a load-sensitive benchmark. It failed once under a fully loaded run, then passed in isolation on the clean baseline, passed in a full clean-baseline `--lib --include-ignored` run (4579 passed, 0 failed), and passed again on the second run with these changes.

Change-specific gates pass. Neither failing test resides in a file this branch modifies.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4872-prune-worktree-root-ancestor.md
+++ b/crates/trusty-mpm/changelog.d/4872-prune-worktree-root-ancestor.md
@@ -1,0 +1,5 @@
+Fixed
+
+- `tm ls` auto-prune now clears dead session records whose git worktree was fully removed (closes [#4872](https://github.com/bobmatnyc/trusty-tools/issues/4872))
+  - `git worktree remove` deletes leaf and parent together, so the old immediate-parent probe read the removal as an unmounted volume and the record could never advance — it was never even marked in `auto-prune-seen.json`
+  - absence is now corroborated against the nearest surviving worktree root (`.base/.worktrees`, `.claude/worktrees`); when that root is gone too, the record is still kept, so an unplugged external volume cannot mass-tombstone the sessions on it
--- a/crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs
@@ -395,15 +395,24 @@ async fn workspace_verified_gone(s: &ManagedSessionSummary) -> bool {
 /// What: `(parent-dir-name, container-dir-name)` pairs matched against a
 /// candidate's ancestors — `.base/.worktrees` (session worktrees) and
 /// `.claude/worktrees` (the worktree root `CLAUDE.md` documents).
-const WORKTREE_ROOT_MARKERS: [(&str, &str); 2] =
+const WORKTREE_CONTAINER_MARKERS: [(&str, &str); 2] =
     [(".base", ".worktrees"), (".claude", "worktrees")];
 
-/// Whether `dir` is one of the [`WORKTREE_ROOT_MARKERS`] container directories.
+/// Whether `dir` is a worktree CONTAINER — the directory that HOLDS worktrees,
+/// never an individual worktree itself.
 ///
-/// What: matches on the LAST TWO path components only, so it holds for any
-/// checkout location and for nested worktrees alike.
+/// 🔴 Read the name literally: this is `true` for `.base/.worktrees` and
+/// `.claude/worktrees`, and `false` for every worktree inside them. It is the
+/// exact OPPOSITE of `session_manager::worktree_safety::is_worktree_root`, which
+/// asks whether a path is the root of ONE worktree (`git rev-parse
+/// --show-toplevel`). The two answer contradictory questions about the same
+/// paths, so they must never share a name (PR #4874 review).
+///
+/// What: matches the LAST TWO path components against
+/// [`WORKTREE_CONTAINER_MARKERS`], so it holds for any checkout location and for
+/// nested worktrees alike.
 /// Test: `auto_prune_clears_record_whose_worktree_root_survived_the_removal`.
-fn is_worktree_root(dir: &Path) -> bool {
+fn is_worktree_container(dir: &Path) -> bool {
     let Some(name) = dir.file_name().and_then(|n| n.to_str()) else {
         return false;
     };
@@ -414,7 +423,7 @@ fn is_worktree_root(dir: &Path) -> bool {
     else {
         return false;
     };
-    WORKTREE_ROOT_MARKERS
+    WORKTREE_CONTAINER_MARKERS
         .iter()
         .any(|(p, c)| *p == parent && *c == name)
 }
@@ -422,34 +431,34 @@ fn is_worktree_root(dir: &Path) -> bool {
 /// Decide whether an absent `candidate` is absent because something was DELETED
 /// (corroborated) or because its filesystem is unreachable (#4872).
 ///
-/// Why: see [`WORKTREE_ROOT_MARKERS`]. This deliberately does NOT settle for
+/// Why: see [`WORKTREE_CONTAINER_MARKERS`]. This deliberately does NOT settle for
 /// "some ancestor exists" — `/Volumes` survives every unmount, so that weaker
 /// rule would reinstate exactly the mass-tombstone the guard was written to
 /// prevent.
-/// What: walks the ancestors nearest-first. When any [`is_worktree_root`]
+/// What: walks the ancestors nearest-first. When any [`is_worktree_container`]
 /// ancestor still exists, the tree below it was removed — corroborated. When the
 /// path has such an ancestor but NONE of them exist, the whole checkout is
-/// unreachable — not corroborated. A path with no worktree-root ancestor at all
+/// unreachable — not corroborated. A path with no container ancestor at all
 /// falls back to the original immediate-parent rule, unchanged.
 /// Test: `auto_prune_clears_record_whose_worktree_root_survived_the_removal`,
 /// `auto_prune_keeps_record_when_the_worktree_root_itself_is_gone`,
 /// `auto_prune_keeps_record_whose_parent_directory_is_unreachable`.
 async fn absence_corroborated(candidate: &Path) -> bool {
-    let mut saw_worktree_root = false;
+    let mut saw_container = false;
     for ancestor in candidate
         .ancestors()
         .skip(1)
-        .filter(|a| is_worktree_root(a))
+        .filter(|a| is_worktree_container(a))
     {
-        saw_worktree_root = true;
+        saw_container = true;
         if tokio::fs::try_exists(ancestor).await.unwrap_or(false) {
             return true;
         }
     }
-    if saw_worktree_root {
+    if saw_container {
         return false;
     }
-    // No worktree root in this path: fall back to the immediate parent. A path
+    // No worktree container in this path: fall back to the immediate parent. A path
     // with no parent (a bare relative name, or the filesystem root itself) has
     // nothing to corroborate against, so refuse to call it gone.
     match candidate.parent() {
--- a/crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs
@@ -341,25 +341,26 @@ fn workdir_candidates(s: &ManagedSessionSummary) -> [Option<&str>; 2] {
 ///     none is unverifiable, and is kept);
 ///   * every candidate's probe came back a definitive `Ok(false)`. A probe
 ///     `Err` (permission denied, transient I/O) counts as "possibly present";
-///   * every candidate's PARENT DIRECTORY exists — see below.
+///   * every candidate's absence is CORROBORATED by a surviving directory above
+///     it ([`absence_corroborated`]) — see below.
 ///
-/// 🔴 The parent check is not defensive padding, it is the whole point of the
-/// word "verified". `try_exists("/Volumes/Unplugged/work/proj")` returns
+/// 🔴 The reachability check is not defensive padding, it is the whole point of
+/// the word "verified". `try_exists("/Volumes/Unplugged/work/proj")` returns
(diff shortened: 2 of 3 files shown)


bobmatnyc:

Adversarial review of the `absence_corroborated` guard, Rust Test Ladder rung 5. Reviewed against #4872 and the code only; implementer reasoning was not used to anchor the verdict. Every claim below was executed, not read off the PR body.

### The safety argument holds — verified by substitution, not by inspection

The central claim is that `auto_prune_keeps_record_when_the_worktree_root_itself_is_gone` rules out a naive "walk up until some ancestor exists" implementation. I replaced `absence_corroborated` with exactly that implementation and ran the suite:

```
test tests_behavior_d::auto_prune_keeps_record_when_the_worktree_root_itself_is_gone ... FAILED
test tests_behavior_d::auto_prune_keeps_record_whose_parent_directory_is_unreachable ... FAILED
test tests_behavior_d::auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable ... FAILED
assertion `left == right` failed: worktrees round 1: an absent worktree ROOT means the
filesystem is unreachable, not that the worktree was removed
  left: 1
 right: 0
test result: FAILED. 21 passed; 3 failed; 0 ignored; 0 measured; 1415 filtered out
```

Three tests kill it, not one. Claim confirmed.

Fail-before, by restoring the old parent-only rule on this branch:

```
test tests_behavior_d::auto_prune_clears_record_whose_worktree_root_survived_the_removal ... FAILED
  left: 0
 right: 1
test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 1415 filtered out
```

Head of branch, unmodified:

```
test result: ok. 24 passed; 0 failed; 0 ignored; 0 measured; 1415 filtered out; finished in 1.11s
```

`cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` → EXIT=0. `check_line_cap.sh` && `check_sld.sh` && `check_changelog_fragment.sh` → EXIT=0. Test file is +129/-0; nothing was `#[ignore]`-ed, cfg-gated, narrowed, or deleted.

### Attack results

**Axis 1 — can anything reach "verified gone" that still exists or is merely unreachable?** No, for every path shape I could construct except one (below). A present workspace short-circuits at `try_exists(candidate) == Ok(true)` before corroboration is ever consulted (`session_picker_prune.rs:368-373`). A fully unreachable checkout puts the worktree root inside the unreachable subtree, so it answers `Ok(false)` and the record is kept. Probed against a real filesystem:

```
B removed worktree:      leaf_exists=false old_verdict=false new_verdict=true   (the fix)
C unreachable checkout:  leaf_exists=false old_verdict=false new_verdict=false  (guard intact)
E unknown layout:        leaf_exists=false old_verdict=false new_verdict=false  (fail-safe)
```

Also checked and clean: marker as the last segment (`skip(1)` excludes self, falls back to the parent rule); the marker appearing twice in one path (nested worktrees — inner root gone, outer root present still means something was deleted, correctly corroborated); `.` and `..` components (the textual ancestor walk still finds the root); case variants on case-insensitive APFS (`.CLAUDE/worktrees`, `.claude/Worktrees` both fail the byte-exact match and fall back to the parent rule, which keeps); relative paths and bare names (`parent()` yields `""`, `try_exists("")` is `Ok(false)`, keeps); `/` (`parent()` is `None`, keeps).

**Axis 2 — walk direction and fallback equivalence.** Order is behaviorally irrelevant: the loop returns on the first *existing* worktree-root ancestor, so the predicate is exactly "does any worktree-root ancestor exist" — set membership, order-independent. It terminates on every input because `ancestors()` is finite. The fallback at `session_picker_prune.rs:455-458` is byte-equivalent to the removed rule, including both `None`-parent cases.

**Axis 3 — layout not on the hardcoded list.** Fails safe. Scenario E above: an unknown container name falls through to the immediate-parent rule, which for a fully-removed worktree keeps the record. One stale row, no tombstone. Not a BLOCK.

**Axis 5 — every `unwrap_or(false)`.** Both are the keep direction. `session_picker_prune.rs:445` — an errored probe on an ancestor simply fails to corroborate, and with `saw_worktree_root` set the function returns `false` (keep). `session_picker_prune.rs:456` — an errored parent probe keeps, same as before. No filesystem error becomes "gone".

**Axis 6 — TOCTOU.** A wrong verdict from a concurrent create/remove cannot act on its own: `SIGHTING_MIN_AGE_SECS` (10 min) plus a second listing is required, `AUTO_PRUNE_CAP` bounds a bad batch to 5, and a record that surfaces healthy on any later call has its marker dropped (`session_picker_prune.rs:647-651`), so a transient race cannot accumulate toward confirmation.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 437-459 | The guard now tolerates a reachability boundary *between* the worktree root and the leaf, where the old parent-only rule did not. Demonstrated: a worktree entry that is a dangling symlink (pointing at a removed or unmounted target) probes `leaf_exists=false` while the root survives → `old_verdict=false`, `new_verdict=true`. Same shape for a worktree that is itself a mount point. | If it is ever worth closing: after finding a surviving root, require that no ancestor between the root and the leaf is a symlink (`fs::symlink_metadata`). Not worth the complexity today — `git worktree add` creates plain directories, so neither shape arises from the documented workflow. | Parent |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 406 | `is_worktree_root` collides by name with `session_manager/worktree_safety.rs:523`, which means the opposite thing about the same paths — for `.claude/worktrees/critic-4850`, `worktree_safety`'s returns true and this one returns false. Different modules, so it compiles, but a reader who knows one will misread the other. | Rename to `is_worktree_container`; the constant is already `WORKTREE_ROOT_MARKERS`, and the doc already calls it a CONTAINER. | Fix here |

No CRITICAL, no HIGH.

### Notes

- **The brief's premise is stricter than the code.** This predicate does not gate worktree removal or branch deletion. `decommission_dead_record` posts `?record_only=true` (`session_picker_prune.rs:909`) and requires `workspace_removed == false` in the response body (`:927`) before counting the call safe; the server side is pinned by `decommission_record_only_never_removes_existing_workspace`, `decommission_record_only_never_touches_the_runtime`, and `decommission_record_only_has_no_side_effects_beyond_the_store`. A wrong "gone" verdict here costs a record, never a worktree. That caps the severity ceiling of the MEDIUM above; against a version-skewed daemon the exposure is one record before the stale-daemon sentinel trips, which is #4702's pre-existing design, not something this PR changes.
- Change-site comment (`:374-376`) and the changelog fragment (`crates/trusty-mpm/changelog.d/4872-prune-worktree-root-ancestor.md`, bare `Fixed` on line 1) are both present and correct. The comment wraps to three lines but is a single sentence naming the mechanism, which is what the convention asks for.
- `auto_prune_keeps_record_whose_parent_directory_is_unreachable` and `auto_prune_ignores_daemon_unresumable_when_parent_is_unreachable` are unmodified and still pass — and, per the substitution run above, are load-bearing.

APPROVED for the next pipeline stage.

Issue #3367, opened 2026-07-20
security(trusty-search): lopdf + quick-xml DoS advisories reachable via default document extraction (RUSTSEC-2026-0187/0194/0195)


bobmatnyc:
## Summary

`trusty-search`'s native document-extraction feature (#2932 / #2923) pulls in
three HIGH-severity (CVSS 7.5) RustSec advisories via its PDF and spreadsheet
parsers. `cargo audit` confirms all three are reachable from `trusty-search`:

```
Crate:     lopdf
Version:   0.36.0
ID:        RUSTSEC-2026-0187  (stack-overflow DoS via deeply nested PDF objects)
Solution:  Upgrade to >=0.42.0
Dependency tree:
lopdf 0.36.0
└── pdf-extract 0.9.0
    └── trusty-search 0.34.1

Crate:     quick-xml
Version:   0.31.0 (via calamine 0.26.1) and 0.37.5 (direct dep)
ID:        RUSTSEC-2026-0194  (quadratic run time on duplicate-attribute check, DoS)
ID:        RUSTSEC-2026-0195  (unbounded namespace-decl allocation, memory-exhaustion DoS)
Solution:  Upgrade to >=0.41.0
Dependency tree:
quick-xml 0.31.0
└── calamine 0.26.1
    └── trusty-search 0.34.1

quick-xml 0.37.5
└── trusty-search 0.34.1
```

Not yet visible in the Dependabot feed (rescan lag) — `cargo audit` catches it
immediately because it reads `Cargo.lock` directly rather than waiting on
GitHub's advisory-database sync.

## Reachability analysis — HIGH

- `crates/trusty-search/src/core/extract/pdf.rs` calls
  `pdf_extract::extract_text_from_mem(&bytes)` on raw file bytes from any
  indexed `.pdf`. `.pdf` was previously hard-skipped by the walker
  (`BINARY_EXTS`) and is now extracted by **default** on every indexed repo.
- Same reachability for `.xlsx/.xls/.xlsm` (`extract/xlsx.rs`, via `calamine`)
  and `.docx` (`extract/docx.rs`, via the direct `quick-xml` dependency).
- A genuine Rust **stack overflow aborts the process** — it cannot be caught
  by `catch_unwind`, `spawn_blocking`'s `JoinError`, or a `tokio::time::timeout`.
  One malformed/malicious document anywhere in an indexed tree can crash the
  shared, long-running `trusty-search` daemon for **every project it serves**,
[...]

Issue #2923, opened 2026-07-17
feat(trusty-search): native document indexing — pdf/docx/xlsx text extraction feeding the existing chunk pipeline


bobmatnyc:
## Problem (2026-07-17 research findings)

`trusty-search`'s file walker and watcher currently cannot index common
office document formats:

- **PDF is hard-skipped.** `.pdf` is listed in `BINARY_EXTS`
  (`crates/trusty-search/src/core/walker.rs:309`), so PDFs are excluded
  before any content inspection happens.
- **DOCX/XLSX are silently dropped.** `.docx` / `.xls` / `.xlsx` are absent
  from `SOURCE_EXTS` (`crates/trusty-search/src/core/walker.rs:22`), so they
  fall through the walker's extension allowlist and are never enqueued for
  indexing (`crates/trusty-search/src/core/walker.rs:696-699`).
- **The watch path assumes UTF-8 text.** `watch_loop.rs:202-211` reads
  changed files with `read_to_string`, which fails outright on OOXML
  (`.docx`/`.xlsx`) zip containers — even if extension routing were fixed,
  the live-watch path would error instead of indexing.
- **The `index_file` MCP tool takes pre-extracted text only** — it has no
  format-aware extraction of its own, so it cannot be used as a workaround
  for these formats without a separate extraction step outside the daemon.
- **Zero extraction dependencies exist workspace-wide** — no PDF/DOCX/XLSX
  text-extraction crate is currently pulled in anywhere in `trusty-tools`.
- No prior issue tracks this gap.

## Scope (v1)

Add a native extraction layer so `.pdf`, `.docx`, `.xls`/`.xlsx`/`.xlsm`
files get their text/structured content extracted and fed through the
**existing** chunk pipeline (plain-text/document chunker — never
tree-sitter), on both the initial-index/reindex path and the live watcher
path, with matching behavior on both.

**Explicitly OUT OF SCOPE for v1: OCR for scanned / image-only PDFs.** A PDF
that yields (near-)empty extracted text must surface a visible warning
("no extractable text — scanned/image PDF? OCR not supported") rather than
silently producing an empty/near-empty index entry.

## Planned approach

- New `crates/trusty-search/src/core/extract/` module: extension-dispatched
  `extract_text(path) -> Result<Extracted>`, with per-format submodules
  (pdf/docx/xlsx).
- Walker routes indexable-but-non-plaintext formats through the extractor
[...]

PR #4875, opened 2026-08-05
spike(trusty-search): measure anydoc against the native pdf/docx/xlsx extractors — NO-GO on wholesale replacement

bobmatnyc:
Measurement spike only. No production extraction code is touched: `EXTRACT_EXTS` is unchanged, `read_content` dispatches exactly where it did, and the harness plus the `anydoc` dependency are both gated behind a default-off `anydoc-spike` feature. Verified rather than asserted — `Compiling anydoc` appears **zero** times in the default-feature clippy log.

# Verdict: NO-GO on wholesale replacement

Not close, and not for the reason the spike was framed around. The two xlsx bugs designated as blocking (#8, #9) turn out **not** to be regressions. Three things nobody was looking for are.

| Leg | Verdict | Decisive evidence |
|---|---|---|
| **pdf** | **Blocked, twice over** | Uncatchable `SIGABRT` on a 21 KB file; 5 of 58 real PDFs silently lost |
| **xlsx** | **Blocked** | Zip-bomb allocation 2x worse than ours (1036 MiB vs 521 MiB), zero fidelity gain |
| **docx** | **Genuine win, unreachable alone** | Real table + heading recovery — but anydoc has no feature flags, so taking it drags in the PDF stack |

---

## 1. anydoc reintroduces RUSTSEC-2026-0187, the advisory this repo already fixed

`anydoc 0.1.3` → `pdf-inspector 0.1.7` → **`lopdf 0.41.0`**. The advisory's patched range is `>= 0.42.0`.

The brief stated anydoc's PDF path wraps lopdf 0.42.0, the same version we pin. That is wrong, and it inverts the conclusion.

`cargo audit`, the same command the weekly workflow runs:

```
base lockfile   → exit 0, zero RUSTSEC-2026-0187 hits
spike lockfile  → exit 1

Title:     Stack overflow in lopdf via deeply nested PDF objects
ID:        RUSTSEC-2026-0187      Severity: 7.5 (high)
Solution:  Upgrade to >=0.42.0
lopdf 0.41.0
└── pdf-inspector 0.1.7
    └── anydoc 0.1.3
        └── trusty-search 0.42.3
```

`.github/workflows/cargo-audit.yml` names this advisory in its own comment header as **resolved by #3367**. Adopting anydoc un-resolves it.

**It cannot be fixed with `cargo update`.** `pdf-inspector 0.1.7` declares `lopdf = "0.41.0"`, a caret range that cannot reach 0.42.0. The pin has to change upstream, in a crate that is not anydoc.

### The advisory is not theoretical here — it fires

The harness feeds both extractors the PoC shape the advisory describes (10,380-deep nested array in a page object, 21 KB file), each in an isolated child process:

| fixture | native | anydoc |
|---|---|---|
| `nested-objects.pdf` | `ok+warning(0 bytes)` — clean return | **`SIGABRT`** — process dies |

The advisory is explicit that this is a stack-overflow abort, **not a `panic!`**: `catch_unwind` does not see it. Our `run_extract_with_timeout` contains panics via `spawn_blocking`'s `JoinError` — that containment does nothing here. One hostile PDF in a watched directory takes the daemon down, and the 30s `EXTRACT_TIMEOUT` never gets a chance to fire.

This is precisely the containment `core::extract` was built to provide, removed.

## 2. anydoc silently loses real PDFs that we currently read

67 real documents on the measurement host (58 pdf, 7 xlsx, 2 docx). No filenames or content appear anywhere in the harness output — rows are labelled by format and index.

| format | files | both extracted | native only | **anydoc only** | neither |
|---|---:|---:|---:|---:|---:|
| pdf | 58 | 38 | **7** | **0** | 13 |
| xlsx | 7 | 5 | 0 | 0 | 2 |
| docx | 2 | 1 | 0 | 0 | 1 |

anydoc never once extracted a document ours could not. It lost 7 that ours reads, and 5 of those are substantial — 36k, 95k, 40k, 23k, and 49k characters of text, gone entirely. All 7 fail with:

```
unsupported input: PDF has no extractable text (TextBased, N pages): OCR is required
```

`pdf-inspector` classifies the document as `TextBased` — it can see there is a text layer — and then extracts nothing from it. (2 of the 7 are `ImageBased` scans where the substantive outcome is a tie; the 5 `TextBased` ones are the defect.)

### This is not an artifact of a synthetic fixture

The same failure showed up on a hand-built fixture, so it was checked against third-party references before being believed:

- `qpdf --check` → *"No syntax or stream encoding errors found"*
- poppler `pdftotext` → extracts all 40 pages correctly
- `pdf-extract` (ours) → 2,192 characters
- anydoc → `no extractable text (TextBased, 40 pages)`

A page-count sweep with the object layout held constant fails identically at 1, 2, 3, 4, 5, 8, 16, and 40 pages, while poppler reads every one. It is a document-shape sensitivity, not a size limit — and it reproduces on real files.

## 3. Adversarial parity — where anydoc is worse, and where it is better

Every probe is an isolated child under a 45s budget. All fixtures sit under the 10 MiB `MAX_OFFICE_FILE_BYTES` gate on purpose: a file the walker already rejects measures nothing.

| fixture | native | anydoc | reading |
|---|---|---|---|
| `nested-objects.pdf` | clean return | **SIGABRT** | **anydoc worse — fatal** |
| `zipbomb.xlsx` | ok, **521 MiB** peak | ok, **1036 MiB** peak | **anydoc worse — 2x** |
| `zipbomb.docx` | rejected at 50 MiB | rejected at 128 MiB | anydoc bound 2.5x looser, but present |
| `nested-xml.docx` (100k deep) | ok, 200,004 bytes | rejected at depth 256 | ours extracts, anydoc refuses |
| `truncated.{docx,xlsx,pdf}` | clean error | clean error | parity |
| `malformed.{docx,pdf}` | clean error | clean error | parity |

**On the docx zip bomb, anydoc does have a bound.** `package::limits::MAX_ENTRY_BYTES` (128 MiB) fires and returns `ResourceLimit`, so wholesale replacement would not remove the `MAX_DOCUMENT_XML_BYTES` + `Read::take` protection outright — it would loosen it from 50 MiB to 128 MiB. Worth stating plainly, since the scope change assumed the protection might vanish entirely.

**On the xlsx zip bomb, anydoc's declared limits do not apply at all.** Both extractors hand the workbook to calamine, which opens the archive itself and bypasses anydoc's package layer. The residual risk documented on #2923 (calamine exposes no pre-decompression bound) is therefore **identical in kind and worse in degree**: 1,036 MiB peak against our 521 MiB, for a 511 KB input. Neither is acceptable; anydoc is the worse of the two.

**Panic containment.** anydoc returned `Result` on every input that did not kill the process, and never unwound. That is a genuine point in its favour over `pdf-extract`'s known panic paths — and it is worth exactly nothing on the one input that matters, because a `SIGABRT` is not a panic. The 30s timeout and `spawn_blocking` containment would still be required, and would still be insufficient.

## 4. anydoc bugs #8 and #9 — both reproduce, neither is a regression

Both were filed against the Python bindings. Both reproduce against the Rust crate 0.1.3. Neither blocks anything, because our extractor is no better.

**#8, merged-cell spans clipped.** Confirmed: an `F1:O3` merge (3 rows × 10 cols) with only the anchor populated renders as a 1×1 table. But the merge span is layout metadata that **our extractor never had either**, and the anchor text survives in both:

| check | native | anydoc |
|---|---|---|
| `Merged heading` present | yes | yes |

Nothing searchable is lost. Not a regression.

**#9, hidden rows/columns treated as visible.** Confirmed — and **our extractor does exactly the same thing**:

| probe | native | anydoc |
|---|---|---|
| `Visible row` | yes | yes |
| `Hidden row` | yes | yes |
| `Hidden column` | yes | yes |

Identical behaviour. Whether indexing author-suppressed content is right is a real product question, but it is a question about our current extractor, not about adoption. Adopting anydoc changes nothing here.

The scope change asked me to quantify these as blocking. They are not. The blockers are elsewhere.

## 5. The docx structure gain is real — and it is the one thing worth wanting

This is where anydoc delivers what was hoped for.

`structured-report.docx` (6 headings, 4 tables of 5×4, 60 paragraphs):

| signal | native | anydoc |
|---|---:|---:|
| markdown heading lines | 0 | **6** |
| table delimiter rows | 0 | **4** |
| table pipe characters | 0 | **140** |
| cell text present | 4 | 4 |

Cell *text* already survives our extractor; every row and column boundary does not. anydoc recovers the structure.

*(An earlier run showed 0 headings for anydoc. That was my fixture's fault — anydoc resolves heading level through `word/styles.xml`, which the fixture omitted. Corrected, and the corrected result is above.)*

### What it buys at the index level

`core::extract` routes all these formats to `chunk_ast`, which finds no grammar and falls through to `chunk_text(window=150, stride=50)`. Markdown pipes are just characters to that chunker. The measurable effect is density:

| fixture | native chunks | anydoc chunks | native lines | anydoc lines |
|---|---:|---:|---:|---:|
| structured-report.docx | 4 | **2** | 294 | 165 |
| large-report.docx | 126 | **86** | 6400 | 4359 |

A table row that our extractor emits as N anonymous paragraphs becomes one line. 32% fewer chunks on the large fixture, with slightly *more* total text. Denser chunks with row context intact is a real retrieval improvement — a query matching two cells in the same row now hits one chunk containing both.

**But it is not reachable in isolation.** anydoc has **no feature flags** — all 11 dependencies are unconditional. There is no way to take the docx frontend without linking `pdf-inspector` → `lopdf 0.41.0`. Even if we never route a PDF through it, the advisory lands in `Cargo.lock`, and `cargo audit` reads the lockfile, not the feature-resolved graph.

## 6. The simplification case does not hold

The scope change asked me to quantify simplification as part of the argument. Measured, it runs the other way.

**Code deleted** — `pdf.rs` + `docx.rs` + `xlsx.rs`, excluding test modules:

| file | production | tests |
|---|---:|---:|
| pdf.rs | 46 | 202 |
| docx.rs | 181 | 169 |
| xlsx.rs | 58 | 270 |
| **total** | **285** | **641** |

~285 production lines removed. The 641 test lines are not a saving — they encode the zip-bomb bounds, the RUSTSEC-2026-0187 regression test, and the quick-xml entity handling from #3367. Deleting them discards the record of three fixed defects.

**Dependencies and size**, measured with two isolated scratch crates each linking only its own extraction stack, `--release`, stripped:

| | native stack | anydoc stack | delta |
|---|---:|---:|---:|
| binary, macOS arm64 | 2,340,688 B | 7,004,816 B | **+4.45 MiB (2.99x)** |
| binary, aarch64-linux-gnu | 2,593,016 B | 7,202,896 B | **+4.40 MiB (2.78x)** |
| crates in graph | 91 | 175 | **+84** |

`pdf-extract` is dropped; `calamine`/`quick-xml`/`zip` become transitive rather than direct. Net: **285 fewer production lines in exchange for 84 more crates and 4.4 MiB.**

**The `zip` duplication premise was also wrong.** `zip` 1.1.4 / 2.4.2 / 3.0.0 / 7.2.0 / **8.6.0** all already coexist on `main` — calamine has pulled 8.6.0 for some time. anydoc adds **zero** new `zip` versions.

**What it does add is a C toolchain dependency.** anydoc requests `zip ^8.6.0` with default features on, which turns on zstd for the shared node and pulls in `zstd-sys` — 43 C source files, `links = "zstd"`. Every other crate anydoc adds is pure Rust. Consequence, measured:

```
cargo check --target aarch64-unknown-linux-gnu
  native stack → exit 0
  anydoc stack → exit 101
    error occurred in cc-rs: failed to find tool "aarch64-linux-gnu-gcc"
```

The current extraction stack cross-compiles with no C toolchain at all. anydoc's does not.

## 7. Build legs — what I could and could not verify

**Could not run containers.** Docker Desktop would not start on this host (`docker info` fails; the API returns HTTP 500). I am not reporting a container result I did not obtain.

**What I did instead:** built both stacks with `cargo-zigbuild`, which supplies the cross C toolchain, and verified the machine type of the output rather than inferring it from a successful link.

| target | native | anydoc |
|---|---|---|
| `aarch64-unknown-linux-gnu` | exit 0 | exit 0 |
| `aarch64-unknown-linux-gnu.2.34` (AL2023 glibc floor) | exit 0 | exit 0 |

```
size-anydoc: ELF 64-bit LSB pie executable, ARM aarch64, ... stripped
size-native: ELF 64-bit LSB pie executable, ARM aarch64, ... stripped
```

**Scope of that result, stated precisely:** it covers the anydoc dependency subgraph on both aarch64 legs. It does not exercise the full `trusty-search` binary on those legs. PR #4822 moved the `aarch64-unknown-linux-gnu` leg to a native ARM runner specifically because zig's clang mis-compiled `numkong`; a zigbuild pass here says nothing about that. `aarch64-linux-al2023` is a repo label, not a rustc triple — the glibc floor is the reproducible part of it, and that is what I pinned.

## 8. Speed and memory — the one place anydoc's numbers do not matter

anydoc is 2.0x–7.7x slower and allocates 1.5x–8x more, and neither figure should influence the decision, because the absolutes are trivial at this scale.

| fixture | native med (ms) | anydoc med (ms) | ratio | native ΔRSS | anydoc ΔRSS |
|---|---:|---:|---:|---:|---:|
| small-memo.docx | 0.03 | 0.06 | 2.16x | 0.6 MiB | 1.5 MiB |
| structured-report.docx | 0.05 | 0.23 | 4.34x | 0.7 MiB | 2.1 MiB |
| large-report.docx | 0.54 | 4.15 | 7.69x | 1.2 MiB | 9.3 MiB |
| multi-sheet.xlsx | 1.74 | 5.20 | 2.98x | 1.8 MiB | 3.8 MiB |
| multi-page.pdf | 0.37 | 0.72 | 1.96x | 1.2 MiB | 4.5 MiB |

N = 100 repetitions per cell (10–30 for the largest fixtures), min and median reported. On the real corpus the largest PDF took 17.6 ms native / 24.4 ms anydoc. Against a 30s `EXTRACT_TIMEOUT` and one-file-at-a-time ingest, a 4 ms parse and a 9 MiB peak are noise. Reported for completeness; they carry no weight in the verdict.

---

# Recommendation

**Do not adopt anydoc for pdf or xlsx.** Not "wait for the spreadsheet bugs to close" — the spreadsheet bugs turned out to be irrelevant. The reasons are an uncatchable process abort, an 8.6% silent data-loss rate on real PDFs, and worse zip-bomb behaviour than the code it would replace.

**Do not adopt it for docx yet either**, despite the structure gain being real and worth having. Not because docx conversion is bad — it is clearly better than ours — but because anydoc ships no feature flags, so there is no way to take the docx frontend without putting `lopdf 0.41.0` in `Cargo.lock` and turning the weekly `cargo audit` red.

**On the new formats** (ppt/pptx, odt/ods/odp, rtf, epub, legacy .doc) I have no measurement — we cannot read them at all today, so there is no baseline to compare against and I did not invent one. The case for them is real but rests on the same dependency, so it is gated behind the same upstream fixes.

### What would change the answer

Two upstream fixes, in this order:

1. **`pdf-inspector` bumps `lopdf` to `>= 0.42.0`.** Clears the advisory and the abort. Not something we can do with `cargo update`.
2. **`pdf-inspector` extracts text from `TextBased` PDFs it already classifies correctly.** The 5 real-document failures are worth filing upstream with a reproducer — the harness in this PR generates one.

A third, smaller: anydoc setting `default-features = false` on `zip` would drop `zstd-sys` and restore pure-Rust cross-compilation.

If the first two land, re-run this harness. It is one command, and the comparison is already wired.

### Merge consideration

**This PR should not be merged as-is**, independent of the adoption decision. Its `Cargo.lock` alone carries `lopdf 0.41.0` and turns `cargo-audit.yml` red weekly — the feature gate keeps anydoc out of every *build*, but `cargo audit` reads the lockfile, not the feature-resolved graph. If the harness is worth keeping in-tree, drop the dependency and lockfile entry and restore them when re-measuring.

---

## Verification

```
cargo fmt --check                                             → exit 0
cargo clippy -p trusty-search --all-targets -D warnings       → exit 0
cargo clippy -p trusty-search --features anydoc-spike
             --all-targets -D warnings                        → exit 0
cargo test -p trusty-search                                   → exit 0

test result: ok. 1456 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 349 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
(22 further targets, all ok, 0 failed)
```

Inertness, verified not assumed: `grep -c 'Compiling anydoc'` over the default-feature clippy log → **0**.

Raw measurements: `crates/trusty-search/examples/anydoc_spike/MEASUREMENTS.md`.

No production extraction code changed. No version bumps. No daemon or `TRUSTY_DATA_DIR` touched — the harness writes only to a scratch corpus directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,8 +24,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
- "cipher",
+ "cipher 0.4.4",
  "cpufeatures 0.2.17",
+]
+
+[[package]]
+name = "aes"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8eb277bec05f56a0e0591f155a484cbd0f4f07ff2905051a48c72f004f7ed58"
+dependencies = [
+ "cipher 0.5.2",
+ "cpubits",
+ "cpufeatures 0.3.0",
 ]
 
 [[package]]
@@ -170,6 +181,23 @@ dependencies = [
  "anstyle",
  "once_cell_polyfill",
  "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "anydoc"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "35ebd211acf8cc346d82dbdffac68509be8963384cb0c11f375a827d3c49445c"
+dependencies = [
+ "calamine",
+ "cfb 0.14.0",
+ "csv",
+ "encoding_rs",
+ "flate2",
+ "log",
+ "pdf-inspector",
+ "quick-xml",
+ "zip 8.6.0",
 ]
 
 [[package]]
@@ -496,7 +524,7 @@ dependencies = [
  "fastrand",
  "hex",
  "http 1.4.0",
- "sha1",
+ "sha1 0.10.6",
  "time",
  "tokio",
  "tracing",
@@ -1125,6 +1153,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
 dependencies = [
  "hybrid-array",
+ "zeroize",
 ]
 
 [[package]]
@@ -1255,6 +1284,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bzip2"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3a53fac24f34a81bc9954b5d6cfce0c21e18ec6959f44f56e8e90e4bb7c346c"
+dependencies = [
+ "libbz2-rs-sys",
+]
+
+[[package]]
 name = "cairo-rs"
 version = "0.18.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1281,12 +1319,13 @@ dependencies = [
 
 [[package]]
 name = "calamine"
-version = "0.36.0"
+version = "0.36.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6975084f43060e56343ffba7f9731fa52a7dcf2e1cd8e2459fd4c6bf4a1bff59"
+checksum = "5fa68281b1a76b54a62156474adb06bb380a67e07dd60656e3217152b42183f3"
 dependencies = [
  "atoi_simd",
  "byteorder",
+ "chrono",
  "codepage",
  "encoding_rs",
  "fast-float2",
@@ -1492,7 +1531,7 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "26b52a9543ae338f279b96b0b9fed9c8093744685043739079ce85cd58f289a6"
 dependencies = [
- "cipher",
+ "cipher 0.4.4",
 ]
 
 [[package]]
@@ -1522,6 +1561,17 @@ dependencies = [
  "byteorder",
  "fnv",
  "uuid",
+]
+
+[[package]]
+name = "cfb"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a347dcabdae9c31b0825fd6a8bed285ec9c2acb89c47827126d52fa4f59cece3"
+dependencies = [
+ "fnv",
+ "uuid",
+ "web-time",
 ]
 
 [[package]]
@@ -1633,7 +1683,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773f3b9af64447d2ce9850330c473515014aa235e6a783b02db81ff39e4a3dad"
 dependencies = [
  "crypto-common 0.1.7",
- "inout",
+ "inout 0.1.4",
+]
+
+[[package]]
+name = "cipher"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8cf2a2c93cd704877c0858356ed03480ff301ee950b43f1cbe4573b088bfa6c"
+dependencies = [
+ "crypto-common 0.2.2",
+ "inout 0.2.2",
 ]
 
 [[package]]
@@ -1967,6 +2027,12 @@ dependencies = [
  "libc",
  "winapi",
 ]
+
+[[package]]
+name = "cpubits"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15b85f9c39137c3a891689859392b1bd49812121d0d61c9caf00d46ed5ce06ae"
 
 [[package]]
 name = "cpufeatures"
@@ -2490,6 +2556,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7eed2c4702fa172d1ce21078faa7c5203e69f5394d48cc436d25928394a867a2"
 
 [[package]]
+name = "deflate64"
+version = "0.1.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac6b926516df9c60bfa16e107b21086399f8285a44ca9711344b9e553c5146e2"
+
+[[package]]
 name = "deltae"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2638,6 +2710,7 @@ dependencies = [
  "const-oid",
  "crypto-common 0.2.2",
  "ctutils",
+ "zeroize",
 ]
 
 [[package]]
@@ -2851,7 +2924,7 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a8bfa975b1aec2145850fcaa1c6fe269a16578c44705a532ae3edc92b8881c7"
 dependencies = [
- "cipher",
+ "cipher 0.4.4",
 ]
 
 [[package]]
@@ -5030,7 +5103,7 @@ version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a588916bfdfd92e71cacef98a63d9b1f0d74d6599980d11894290e7ddefffcf7"
 dependencies = [
- "cfb",
+ "cfb 0.7.3",
 ]
 
 [[package]]
@@ -5072,6 +5145,15 @@ checksum = "879f10e63c20629ecabbb64a8010319738c66a5cd0c29b02d63d272b03751d01"
 dependencies = [
  "block-padding",
  "generic-array",
+]
+
+[[package]]
+name = "inout"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4250ce6452e92010fdf7268ccc5d14faa80bb12fc741938534c58f16804e03c7"
+dependencies = [
+ "hybrid-array",
 ]
 
 [[package]]
@@ -5218,10 +5300,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f00b5dbd620d61dfdcb6007c9c1f6054ebd75319f163d886a9055cec1155073d"
 dependencies = [
  "jiff-static",
+ "jiff-tzdb-platform",
  "log",
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


bobmatnyc:
Verdict: NO-GO on wholesale replacement.

Blocking findings:
- (a) `pdf-inspector 0.1.7` pins `lopdf = "0.41.0"` as a caret range that cannot reach 0.42.0, reintroducing RUSTSEC-2026-0187 — and the advisory's PoC shape kills anydoc with SIGABRT, which `catch_unwind` and `spawn_blocking` containment cannot catch and `EXTRACT_TIMEOUT` never fires on.
- (b) over 58 real PDFs anydoc extracted zero that the current extractor could not, and silently lost 7 that it reads, five of them substantial.
- (c) the simplification inverts — about 285 production lines removed for +84 crates and +4.4 MiB, plus `zstd-sys` breaking `aarch64-unknown-linux-gnu` cross-compile.

Not the problem, despite being expected to be: anydoc issues #8 and #9 both reproduce, but neither is a regression — on hidden rows the current extractor behaves identically.

The one genuine win: docx heading and table recovery, unavailable because anydoc has no feature flags.

Re-open condition: `pdf-inspector` bumping `lopdf` to `>= 0.42.0` AND fixing `TextBased` extraction. Then re-run the harness on this branch — it's one command, already wired.

Branch retained deliberately, not deleted, so the harness can be re-run later.

Issue #4881, opened 2026-08-05
skill manifest save has no lock — concurrent sessions freeze skills nobody edited


bobmatnyc:
## Defect

`SkillManifest::load`/`save` ([`crates/trusty-agents-common/src/skills/manifest.rs`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-agents-common/src/skills/manifest.rs)) take no lock, so overlapping deploy runs race on load-modify-save. One process writes a skill file and saves its manifest; a slower concurrent process then saves a manifest snapshot taken BEFORE that write, overwriting the bookkeeping. The recorded checksum for a key then names an older revision than the bytes actually on disk. `deploy_one_file` ([`crates/trusty-agents-common/src/skills/deployer.rs:292-296`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-agents-common/src/skills/deployer.rs#L292-L296)) reads that mismatch as "user hand-edited this file" and skips it PERMANENTLY. A skill nobody ever touched becomes frozen against all future framework updates, silently.

This matters because this repo routinely runs many concurrent `tm`-orchestrated worktree sessions, each provisioning skills — so the race window is not theoretical.

## Evidence

Measured 2026-08-05, trusty-mpm 1.3.5, main `cdcbd6c0`. Three skills are frozen: `tm`, `tm-delegation-patterns`, `code-review-standards`. Each deployed copy was diffed against every commit in its file's history and found BYTE-IDENTICAL to an official historical main commit — no hand-authored content in any of them:

- `tm` config-dir tier = `c7fff636`; `tm` project tier = `3b9276af`
- `tm-delegation-patterns`, both tiers = `6276ec2f`
- `code-review-standards` config-dir = `4d575937`; project tier = identical to current tip, i.e. fully current yet still flagged frozen

The signature: in each case the manifest's recorded checksum names an OLDER commit than what is on disk. Project-tier `code-review-standards` is the clearest case — the content is current and correct, and it is still frozen purely by stale bookkeeping.

**Confidence:** this is the best-fitting explanation from disk state and code reading, not a reproduced race.

## Resolution

Serialize manifest read-modify-write (file lock, or an atomic compare-and-swap on the manifest). Related hardening worth considering in the same change: verify the on-disk checksum at save time rather than trusting the in-memory snapshot, so a stale save cannot clobber a newer record.

### Notes

[...]

Issue #4873, opened 2026-08-05
skill deploy declines are silent — a frozen skill stays stale forever with no warning


bobmatnyc:
## Defect

Skill deployment runs on every path, but declining a frozen skill is silent. When a skill's checksum no longer matches the manifest, `deploy_one_file` treats it as user-edited and skips it permanently ([crates/trusty-agents-common/src/skills/deployer.rs:292-296](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-agents-common/src/skills/deployer.rs#L292-L296)). No warning is emitted; the user has no way to know the skill stayed stale across every run.

This mirrors the agent silent-failure halves of [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840), which [PR #4848](https://github.com/bobmatnyc/trusty-tools/pull/4848) fixed by emitting a bounded warning in `agent_source::deploy_summary_lines`.

## Evidence

Checksum-frozen skills that stayed stale through normal deployment:

- `tm`
- `tm-delegation-patterns`
- `code-review-standards`

Total frozen set: ~12 skills with on-disk checksums differing from the bundled source. Stale subset (those actually differing): 3 skills named above. The frozen count and stale count are not in conflict: frozen is the set that will skip FUTURE updates; stale is the subset differing from source content today.

## Resolution

[PR #4876](https://github.com/bobmatnyc/trusty-tools/issues/4876) emits one bounded warning naming declined skills and points at `tm doctor --fix-skills --include-frozen`, mirroring the agent path.

## Remediation for Already-Frozen Files

Fixing the silent-decline issue alone will not refresh frozen skills. The separate action is:

```
tm doctor --fix-skills --include-frozen
```

Do not reopen [#4604](https://github.com/bobmatnyc/trusty-tools/issues/4604) — that tracks the deeper remediation of refactoring frozen skills back into the update flow, which is a separate initiative.

bobmatnyc:
[...]

PR #4882, opened 2026-08-05
fix(trusty-mpm): deploy the project skill tier when the project manifest changes (#4880)

### LINKED ISSUE #4880: deploy the project skill tier when the project manifest is updated
OWNER RULING (2026-08-05): the project tier deploys WHEN THE PROJECT MANIFEST IS UPDATED — not on every run. This refines the earlier every-run ruling, which is already satisfied for the managed config home.

CURRENT STATE. `$CLAUDE_CONFIG_DIR/skills` deploys on every run: `ensure_managed_config_dir_with_root` calls `ensure_skill_source_fresh` ([crates/trusty-mpm/src/core/managed_config.rs:143](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/core/managed_config.rs#L143)) and `deploy_all_skill_tiers` ([:180](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/core/managed_config.rs#L180)), and all three run paths reach it via `prepare_managed_config`. That half is done — see [#4873](https://github.com/bobmatnyc/trusty-tools/issues/4873).

THE GAP. `<workspace>/.claude/skills` deploys only via `prepare_session`/`sync_assets`, neither of which runs on resume or in-place relaunch. Because PROJECT TIER OUTRANKS USER TIER in this repo's precedence model, a stale project-tier skill silently wins over a current user-tier one — the same shadowing shape as [#4408](https://github.com/bobmatnyc/trusty-tools/issues/4408), where a 32-byte project-tier stub beat the real 25KB agent. Found during the code-critic review of [PR #4876](https://github.com/bobmatnyc/trusty-tools/pull/4876).

RESOLUTION. Trigger project-tier deployment on project-manifest update. Define precisely what "updated" means (manifest checksum change is the obvious candidate — mirror how `ensure_agent_source_fresh` uses a sha256 stamp to no-op when nothing changed) so a relaunch with an unchanged manifest stays cheap and rewrites nothing.

CONSTRAINT THAT MUST HOLD — the owner's standing rule: CUSTOM SKILLS ARE NEVER OVERWRITTEN. Project-custom skills are unmanaged, absent from the manifest, and skipped unconditionally at [crates/trusty-agents-common/src/skills/deployer.rs:278-283](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-agents-common/src/skills/deployer.rs#L278-L283). Any implementation must preserve that, and must not turn "deploy on manifest update" into "overwrite user edits on manifest update". Checksum-frozen skills must also still be skipped by this path ([deployer.rs:292-296](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-agents-common/src/skills/deployer.rs#L292-L296)); reconciling those is a separate, owner-gated action.

bobmatnyc:
Closes #4880

## The trigger (owner refinement on this PR)

Bob: *"Manifest doesn't need to see content."* and *"When version is updated, we re-run deployment."*

Two components in the stamp, deliberately not a third:

1. **The binary VERSION.** A version bump re-runs deployment. It stands for the shipped skill set, so the manifest never needs to see skill content. Accepted consequence, stated rather than worked around: two builds carrying the same version but different skill text do not redeploy. That is a development-time condition, not a shipped one.
2. **The manifest's SKILL SELECTION** — `plan.skill_source` / `skill_include` / `skill_exclude`. **Kept, and here is why it is real:** those three come straight from the resolved manifest's `[skills]` section (`manifest/apply.rs:112-120`), and the highest-precedence layer feeding it — `<harness-root>/.trusty-mpm/framework/manifest.toml` — is the operator's own file, editable at any moment. An operator excluding a skill on a Tuesday must not wait for a release. Nothing else from the manifest enters the stamp: a `[style]` or `[mcp]` edit cannot change which skills this tier holds, and `stamp_ignores_manifest_changes_outside_the_skill_selection` pins that.

Two files that could also answer to "the project manifest" are **not** the trigger. `.trusty-mpm-skills-manifest.json` is the deploy target's ownership ledger, written BY `deployer::deploy_skills_filtered` at the end of every deploy — keying on it never converges. `framework-manifest.toml` is compiled in, so a change to it *is* a version change; it reaches the stamp through component 1 already.

## The defect

`<workspace>/.claude/skills` is the PROJECT tier, and project outranks user. It was written by exactly two callers — `session_launch::prepare_session` and `tm sessions sync-assets` — and neither runs on a resume or a bare-`tm` in-place relaunch. The user tier is refreshed on every run (#4873). So a project-tier skill an older binary deployed keeps winning over the current one, silently, for the life of the workspace: the #4408 shadowing shape one tier down.

## The fix

`core::project_skill_tier::ensure_project_skill_tier` runs from `managed_config::ensure_managed_config_dir_with_root` — the choke point `spawn`, `spawn_resume`, and `build_inplace_resume_command` all reach via `runtime::claude_code::prepare_managed_config`. It mirrors `ensure_agent_source_fresh`: compare a sha256 stamp against `<workspace>/.claude/skills/.trusty-mpm-project-tier-stamp` and return before touching a file when they agree.

`ensure_managed_config_dir(config_dir)` gains a `project_dir` argument. The destination is computed as `<project_dir>/.claude/skills` rather than read from `fw.claude_skills_dir()` — load-bearing, not decorative: `ensure_managed_config_dir` builds a home-scoped `FrameworkPaths::default()`, so reading it would have aimed the deploy at the operator's real `~/.claude/skills`.

## Custom skills survive the redeploy

Clause 3 of the model — local customizations are tracked in the local manifest and outlive an upgrade. "Re-run deployment on version change" must not become "wipe the user's local copies on upgrade." Two independent mechanisms, both reused unchanged:

1. `tiers::plan_skill_tiers` drops a project-custom (unmanaged) stem from `bundled_deploy` before any file I/O — it never reaches the deployer. `ProjectTierDeploy::shadowed` surfaces that so it is asserted, not inferred from an empty stats vector.
2. `deployer::deploy_one_file` still skips an unmanaged target (`deployer.rs:278-283`) and a checksum-frozen hand edit (`:292-296`).

The project-tier arm deliberately does **not** route skips through `skill_skip_summary`: a skipped skill here is a customization the model says must survive, so pointing at a remedy would contradict it. No user-facing "fix skills" surface is added or extended.

## Review findings addressed

**MEDIUM — an empty source dir stamped as CURRENT.** Fixed. The refusal counted only `!is_dir()`, but `skill_source_dir()` prefers `agents/skills` whenever that path *exists*, and an unfetched or shallow submodule is an existing-but-empty directory. The guard passed, the stamp recorded success, and the project tier froze stale with no signal — the silent-staleness class of #4840 / #4873. It now counts the stems the deployer would actually act on (`list_source_stems`), warns naming the submodule case, and leaves the stamp unwritten. Test: `empty_skill_source_does_not_stamp`, which builds the real empty-submodule shape.

**Both required negative tests survived the rework unchanged** — `project_custom_skill_is_never_overwritten` (content, the `shadowed` record, AND absence from `stats.deployed`, on both runs) and `frozen_skill_is_still_skipped`.

**LOW, left as-is:** deleting one deployed skill directory while the stamp survives fails open until the next version bump or selection edit. The empty-source fix does **not** cover it — that guard is about the source, this is about the destination. Detecting it would mean stat-ing every deployed stem on every run, which is the per-run cost the ruling exists to avoid.

**Known interaction, not fixed here:** #4881's skills-manifest race. This PR widens it from two writers to four. Filed and stacked after this one; no lock attempted here.

## Test ladder — rung 5 (process lifecycle)

```
cargo test -p trusty-mpm -- --test-threads=1             EXIT=0  — 45 suites, 7761 passed, 0 failed, 17 ignored
bash scripts/check_line_cap.sh                           EXIT=0
cargo check --workspace                                  EXIT=101 — PRE-EXISTING
```

`cargo check --workspace` fails identically on main: a Tauri proc macro in `trusty-mpm-gui` and `trusty-code-gui` panics with `The 'frontendDist' configuration is set to "ui/dist" but this path doesn't exist` (unbuilt Svelte UI). This diff touches zero files in either crate — `git diff --name-only origin/main...HEAD` lists only `crates/trusty-mpm/`.

`--test-threads=1` per the two documented `$HOME`-mutation races in `docs/reference/test-ladder-baseline.md`.

## Fail-before / pass-after

Proven by mutation on the pre-rework head; the same three mutations still discriminate, with the stamp mutation now aimed at the version component.

| Mutation | Tests that went red |
|---|---|
| Remove the `project_skill_tier` block from `ensure_managed_config_dir_with_root` (= the pre-change state) | `ensure_managed_config_dir_deploys_the_project_skill_tier`, `ensure_managed_config_dir_project_tier_is_a_noop_when_unchanged` |
| Remove the stamp early-return | `unchanged_version_is_a_noop` |
| Drop a component from the stamp | `stamp_changes_with_the_version` / `stamp_changes_with_the_skill_selection` plus the deploy tests behind them |

Covering the model's clauses:

- **Version bump redeploys** — `version_bump_redeploys` (hermetic, via `ensure_project_skill_tier_for_version`; `env!("CARGO_PKG_VERSION")` is a compile-time constant and cannot move mid-test)
- **Unchanged version no-ops** — `unchanged_version_is_a_noop` (asserts `deployed == false` AND unchanged mtime)
- **Selection edit redeploys at the same version** — `skill_selection_change_redeploys`
- **A non-skill manifest edit does not** — `stamp_ignores_manifest_changes_outside_the_skill_selection`
- **Local copies survive the upgrade** — `project_custom_skill_is_never_overwritten`, `frozen_skill_is_still_skipped`
- **Resume + in-place relaunch reach the trigger** — `ensure_managed_config_dir_deploys_the_project_skill_tier`, hung off the same choke point and the same argument #4873's own test uses

## Out of scope, by ruling

A granular agent/skill-only update model is future work. Reconciling today's already-frozen files is a separate one-time migration the coordinator is running.


bobmatnyc:
## Verdict: WARN

Adversarial review of **`5842fa11`** — the head as of this comment. Independent worktree, rung 5 (process lifecycle).

**Reviewed against a design that is being replaced.** The owner has since redirected the trigger from `resolved manifest + skill_bundle_stamp()` to a version-change trigger ("manifest doesn't need to see content", "re-run deployment on version change"). That push has not landed, so this reviews `5842fa11`. Finding 1 below is confined to the superseded stamp design; findings 2-4 survive the rework unchanged, because they are about the refusal guard, the deploy path, and the manifest, none of which the redirect touches.

**The custom-skill guarantee holds, including under the new lifecycle.** I attacked it and could not break it. A version bump that re-runs deployment does **not** wipe local customizations, in either shape the owner described. Detail in section 1 — that is the answer that matters most and it is clean.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/project_skill_tier.rs` | 126-132 | The stamp omits the user-custom skill source, which this same call deploys from. Editing `~/.trusty-mpm/skills/foo.md` leaves the project copy stale — the exact shadowing #4880 exists to fix. **Superseded by the version-change redirect**; re-judge against the new trigger. | Under the new design, decide deliberately whether a user-skill edit should trigger. It will not, under version-only. | Parent |
| MEDIUM | `crates/trusty-mpm/src/core/project_skill_tier.rs` | 182-190 | The refusal catches a MISSING source only. An EMPTY source deploys zero skills and still stamps as current at :200 — "recording zero skills as current", which :152-154 says it avoids. **Survives the rework.** | Extend the guard: `list_source_stems(&plan.skill_source)?.is_empty()` → same warn-and-return, stamp unwritten. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/managed_config.rs` | 246 | Widens the #4881 unlocked-manifest race from two writers to four, and the new model leans harder on that manifest as the record of customizations. **Survives the rework, and matters more under it.** | None here — track on #4881. | Promote (→ #4881) |
| LOW | `crates/trusty-mpm/src/core/project_skill_tier.rs` | 178-180 | Deleting ONE deployed skill dir while the stamp survives fails open: resume no-ops, the skill stays missing until the next spawn heals it. | None; noted. | Parent |

---

## 1. Can any path wipe a user-authored skill? No.

This is the axis that matters most under the version-change model, since a version bump re-runs the full deployment. Both customization shapes the owner described are protected, by different mechanisms:

**(a) Hand-placed skill, absent from the local manifest.** `list_project_custom_stems` (`tiers.rs:243-262`) sees an unmanaged `<dest>/<stem>/SKILL.md`; `plan_skill_tiers` (`:168-184`) drops the stem from `bundled_deploy` before any I/O. `deploy_skills_filtered`'s `select` then never reads the source file, so the `references/` mirroring loop (`deployer.rs:217-241`) never runs for it either. It is not skipped at write time — it never reaches write time.

**(b) A deployed skill the user then edited, tracked in the local manifest.** Managed, checksum mismatch → skipped at `deployer.rs:292-296`.

The write in `deploy_one_file` happens in exactly two situations: the target does not exist, or the target is managed **and** the on-disk bytes match the recorded checksum — i.e. the bytes tm itself wrote. Losing those is not losing user content. There is no `remove_*`, prune, or truncate anywhere in this path (`materialize_skill_artifacts` does prune, but against the SOURCE dir, and it is not called here).

A version bump therefore re-runs deployment over a tier where every customization is either invisible to the planner or checksum-frozen. Attacks tried:

| Attack | Outcome |
|---|---|
| Custom stem that later becomes a bundled name | Unmanaged → project-custom → shadowed at the planner. Asserted on both a first deploy and a triggered one. |
| Bundled skill deployed first, then hand-edited | Managed, checksum mismatch → skipped. |
| Custom skill in a subdirectory (`<dest>/team/my-skill/SKILL.md`) | Not recognized as project-custom (needs `<dest>/<stem>/SKILL.md`) — but the only write targets are `<dest>/<stem>/SKILL.md` and `<dest>/<stem>/references/*.md`, so the nested file is never a target. Worst case a `SKILL.md` is ADDED beside it. Identical in `prepare_session`; not introduced here. |
| Flat `<dest>/my-skill.md` | Never a write target; the deployer only writes `<stem>/SKILL.md`. |
| Hand-added file under a managed skill's `references/` | Key `<stem>/references/<file>` unmanaged, target exists → skipped. |
| Stale stamp forcing a deploy | A deploy is the same operation `prepare_session` already runs. No new write rule. |
| First run, no stamp, dest already populated by hand | Covered by rows 1-2. |
| Corrupt or deleted `.trusty-mpm-skills-manifest.json` | `unwrap_or_default()` (`skills/manifest.rs:83`) → everything unmanaged → everything skipped. Fails toward preservation. |
| Manifest coming to record the USER's bytes (which would make the file overwritable) | Not reachable. `manifest.managed.insert` runs only after a write and stores `checksum(content)` of the SOURCE, never of what was on disk. |
| `list_project_custom_stems` errors | `?` propagates; caller warns and continues; stamp unwritten. |

No BLOCK.

## 2. Stamp coverage — superseded

One line, per the redirect: at `5842fa11` the stamp covers the resolved manifest and the compiled-in bundle but **not** `fw.user_skill_source_dir()`, which the same call deploys from (`project_skill_tier.rs:192-197`), so a user-custom skill edit refreshes `<config_dir>/skills` (`managed_config.rs:230-236`, ungated) while the higher-ranked project copy stays stale. Under a version-only trigger that gap persists and widens, but it becomes an owner design question rather than a review finding. The two residuals the PR discloses (DOC-42 co-deploy, catalog content drift) are both correctly out of scope and both moot under the new trigger.

## 3. Stamp placement — fails safe on the property that matters

```rust
// project_skill_tier.rs:178
if std::fs::read_to_string(&stamp_path).ok().as_deref() == Some(current.as_str()) {
```

A missing stamp reads as `None`, which never equals `Some(_)`, so it deploys. That holds for any trigger value and survives the rework — the requirement that a missing stamp must fail SAFE is met.

- `.claude/skills` deleted → stamp gone → deploy runs. Safe.
- Stamp alone deleted → deploy runs. Safe.
- Stamp present, one skill dir deleted → no-op; the skill stays missing until the next spawn (`prepare_session` is unstamped) heals it. Fail-open, the LOW above.

The leading `.` genuinely hides the marker from both the deployer (`is_skill_file` rejects dot-files, `deployer.rs:63-65`) and the planner (`list_project_custom_stems` requires a directory).

## 4. `project_dir` threading — correct at every call site

One non-test call site: `claude_code.rs:735`, passing `cwd`. All three paths supply the session workspace — `lifecycle.rs:761` (`&prepared.path`), `lifecycle.rs:1542` (`&workspace`), `guided_inplace.rs:514-523` (`record.workspace_path.or(record.cwd)`, with a `current_dir()` fallback already used by `refresh_compiled_prompt` and `build_prompt_file`). No worktree root, no home dir, no empty path.

The `fw.claude_skills_dir()` avoidance is real and load-bearing, not decorative: `ensure_managed_config_dir` builds `FrameworkPaths::default()` (`managed_config.rs:135`), which is home-scoped, so `claude_skills_dir()` there is literally `~/.claude/skills`. `project_skill_dir` computes from `project_dir` instead, and `deploys_into_the_workspace_not_the_framework_home` asserts the negative.

## 5. All three run paths reach it

`spawn` (`claude_code.rs:1115`), `spawn_resume` (`:1216`), `build_inplace_resume_command` (`:961`) → `prepare_managed_config` (`:714`) → `ensure_managed_config_dir` (`:735`) → `ensure_project_skill_tier` (`managed_config.rs:246`).

## 6. Test discrimination

`project_custom_skill_is_never_overwritten` (`project_skill_tier_tests.rs:173-220`) asserts three things, not an empty vector: file content byte-identical, `shadowed` carries `Project`-over-`Bundled` for the stem, and the stem is absent from `stats.deployed` — on both the first deploy and the triggered one. `ProjectTierDeploy::shadowed` earns its place. This test should be carried through the rework unchanged; it is the guarantee's regression net and it is trigger-independent.

Also verified: `HarnessManifest` (`manifest/schema.rs:45`) uses `BTreeMap` throughout, no `HashMap`, so the stamp is not silently non-deterministic.

## 7. #4881 — widened, and it matters more under the new model

Before this PR the project tier had two writers (`prepare_session` on spawn, `sync-assets` on demand); now every resume and in-place relaunch is a third and fourth. `SkillManifest::load`/`save` (`skills/manifest.rs:80-104`) is unlocked load-modify-save.

Concrete lost update, most likely right after a version bump when the trigger fires for every session at once: A loads the older manifest, deploys N files, saves N entries. B loaded the same older manifest, finds A's files already on disk, `is_managed` returns false for all of them, skips every one, and saves a manifest without A's entries.

Content is safe — `deploy_one_file` never writes over an unmanaged target, so the race fails toward preservation. What is lost is the **record**: those files become "unmanaged," which under the new lifecycle is exactly the bookkeeping that distinguishes a local customization from an orphaned deploy. Doctor checks and any future reconciliation read that record. The stamp does limit exposure — the no-op path at :178-180 returns before the manifest is loaded at all — but a version bump is precisely the moment it does not.

Not this PR's to fix. Flagging the worsening for #4881.

## 8. Hygiene

Changelog fragment at `crates/trusty-mpm/changelog.d/4880-project-tier-manifest-trigger.md`, line 1 a bare `Fixed`. `// #4880:` comment at `managed_config.rs:243`. All 27 deletions in the diff are call-signature updates and doc-line edits — no test weakened, `#[ignore]`-d, cfg-gated, or deleted.

## Gates I ran on `5842fa11`

```
cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-mpm -- --test-threads=1                                   EXIT=0
  → 45 suites, 7757 passed, 0 failed, 17 ignored
bash scripts/check_line_cap.sh && bash scripts/check_sld.sh                    EXIT=0
```

All ten new tests pass by name. `cargo check --workspace` not run — the Tauri `frontendDist` proc-macro panic is a documented pre-existing baseline and this diff touches zero files in `trusty-mpm-gui` / `trusty-code-gui`.

## Required Changes

Carry these into the reworked head:

1. Extend the missing-source refusal at `project_skill_tier.rs:182-190` to cover an existing-but-empty source directory. `skill_source_dir()` (`paths.rs:374-382`) prefers `<trusty-mpm-root>/agents/skills` whenever that directory *exists*, and a shallow or uninitialized submodule is an existing empty directory — a state `deployer.rs:169-175` already warns about. Today that stamps an empty tier as current.
2. Keep `project_custom_skill_is_never_overwritten` and `frozen_skill_is_still_skipped` intact across the trigger change, and re-run both against the version trigger. They are what makes the owner's standing rule assertable.
3. Decide deliberately, under the version-only trigger, whether a user-custom skill edit in `~/.trusty-mpm/skills/` should refresh the project tier. It will not. That may be the intended answer — it should be a stated one.

With 1 and 2 carried, the reworked head should land at APPROVE: the surviving findings are both MEDIUM, and the custom-skill guarantee is intact.

Issue #4317, opened 2026-07-29
Bulk registry sweep deregistered ~14 real trusty-search indexes (2026-07-27)


bobmatnyc:
## Incident

The trusty-search registry `indexes.toml` underwent a bulk sweep around 2026-07-27 12:14:42 that dropped entries from 73 to 31 — a loss of 42 index registrations.

A backup file preserves the pre-sweep state with the suffix `.bak-registry-cleanup-20260727-121442`.

## Scope of Removals

Of the 42 dropped entries:
- Many were legitimately disposable: `.tmp*` paths, `.trusty-search-test-roots` fixtures, and other test artifacts.
- Roughly **14 appear to be real, user-facing indexes**, including:
  - `cto-assistant` (currently declared by the cto-assistant agent persona)
  - `apex`
  - `bob-kb`
  - `cto-duetto`
  - `genealogy`
  - `itinerator`
  - `localLLM`
  - `mcp-memory`

## Consequence

At least one deregistered index (`cto-assistant`) is actively declared by a live agent persona's `[[stores]]` configuration. When that agent attempted to use its knowledge connection mid-conversation, it reported to the user:

> "The semantic search index (cto-assistant) is unreachable this turn..."

The user discovered this broken connection only because the assistant mentioned it. **Deregistration is silent from the consumer's side** — no warning during agent startup, no log entry at the application level, no probe that could catch it before a user is affected.

This is documented in #4315, which demonstrates how the incident would have been caught immediately by a per-connection health probe.

## Data Integrity Status

On-disk data appears **PRESERVED** in the cases inspected (e.g., `~/.trusty-agents/knowledge/cto-assistant/.trusty-search/` still exists). This looks like a **registry-state loss, not data loss**. However, a full audit was in progress when this was filed — say "appears" pending comprehensive verification.

## Timing & Causation

The timing aligns suggestively with commit `59674a07` ("fix(trusty-search): safe deregister on DELETE /indexes/:id", PR #4218), which documented preservation of on-disk data during deregistration.

**However, no direct evidence of causation yet.** The daemon does not log inbound method+path at info level, so the trigger is inferred from:
- The registry diff (73→31 entries)
- The timestamp (2026-07-27 12:14:42, matching the backup suffix)
- The deregister fix shipping in the same release cycle

This is **inference, not observed fact**. No log line directly witnesses a DELETE call.

## Design Question for Maintainer

[...]

Issue #4253, opened 2026-07-28
trusty-search: cold-start corpus-open race permanently orphans an index


bobmatnyc:
Root cause found live 2026-07-28, not previously recorded anywhere but session memory. During the ~20s cold-start corpus open, the client sees a FALSE "daemon not reachable". That induces a retry, which races the still-open handle, producing `DatabaseAlreadyOpen` → "refusing to register a broken index handle" → the index disappears from the registry permanently, with no self-heal. A single operator reproduces it.

Operational workaround while open: register the index in a SINGLE attempt with no retry-on-failure, and verify the result via `GET /indexes` rather than trusting the CLI's exit code — the retry is what triggers the race.

Resolution: the client must not report "not reachable" while a corpus open is in progress, and a failed registration must not leave the index unregistered with no recovery path.

FIVE prior issues in this class have been closed and the class keeps recurring, so this one is filed fresh with the actual root cause rather than reopening a closed thread. Related: #4087.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #4883, opened 2026-08-05
fix(trusty-search): retry timeout-skipped indexes and stop dead registry entries starving live ones (#4250, #4846)

### LINKED ISSUE #4250: trusty-search: warm-boot scan timeout skips every index and never self-heals
**Defect.** When the warm-boot scan times out, the affected indexes are skipped and absent from the registry, and nothing ever retries them. The daemon stays degraded indefinitely — it serves, so no liveness check fires — until a human notices and restarts it.

**Evidence** (observed 2026-07-28, trusty-search v0.39.1, uptime ~48594s / ~13.5h):
```
warmboot_summary: { indexes_loaded: 0, indexes_skipped_timeout: 5,
                    indexes_skipped_tcc: 0, indexes_corpus_failed: 0,
                    indexes_failed: 0, warm_boot_degraded: true }
status: "degraded",  indexes: 7
```
Note this is NOT the macOS FDA/cdhash case — `indexes_skipped_tcc` is 0. Also note the top-level `indexes: 7` disagrees with the warm-boot summary's 0 loaded; the two counters should be reconciled or the disagreement explained.

**Blast radius.** `trusty-review` lists trusty_search as a REQUIRED dependency, so it reported `status: degraded` and correctly refused to produce a review (fail-closed behavior from #4086/#4087/#4092, PR #4089). 13.5 hours of degraded state took the automated review gate offline for an entire sprint session with no alarm — the fail-closed refusal at the point of use was the only signal.

**Resolution.** Either retry timed-out indexes on a backoff after warm boot, or surface the degradation somewhere that escalates rather than waiting for a consumer to trip over it. A degraded daemon that serves successfully is invisible to every liveness check we have.

Note in the body that this belongs to a recurring shape in this repo: a failure branch that leaves the process healthy-looking with no self-heal path.

### LINKED ISSUE #4846: trusty-search: warm boot walks every dead registry entry at ~10s each — 375 of 396 entries are stale, collapsing the visible index count on restart
**Symptom.** After a routine `launchctl bootout` → `cargo install` → `bootstrap` cycle, the visible index count fell from **19 to 4**. `tm-trusty-tools-03` (70,180 chunks) returned `404 unknown index` and stayed unreachable for the better part of an hour. Lazy-load-on-query did not rescue it. The daemon reported `warm_boot_degraded: true`, `indexes_loaded: 0`, `indexes_lazy: 169` and climbing at roughly one entry per 10 seconds.

**Root cause.** `~/Library/Application Support/trusty-search/indexes.toml` contains **396 entries**, of which only ~19 correspond to real projects:

| category | count | root exists |
|---|---|---|
| real project entries | 19 | 16 |
| `~/.trusty-search-test-roots/ts-*` fixtures | 17 | 0 |
| `/private/var/folders/.../T/.tmp*` ephemeral | 360 | 5 |

**375 entries have a root path that no longer exists.** Warm boot walks all 396 in registry order. Each dead entry burns a full ~10-second timeout before being logged as parked for lazy load (the behavior introduced by #4087). Real indexes queued behind that backlog stay invisible for tens of minutes.

**Why this is worse than it looks.** The parking behavior from #4087 is a genuine improvement over silently dropping indexes — nothing is lost. But it converts a fast failure into a slow one, and the cost scales with accumulated cruft rather than with real index count. The registry grows monotonically: nothing prunes an entry when its root directory is deleted, so every worktree teardown and every test run that registers a temp root leaves a permanent ~10-second tax on every future boot. At 375 dead entries that is over an hour of warm-boot walk.

**Asks** (in priority order, the first is the actual fix):
1. **Drop, or fast-path, an entry whose `root_path` does not exist at boot** rather than spending a 10s timeout on it. A `stat` is microseconds and is a definitive answer for this class. Parking should be reserved for entries whose root exists but whose corpus cannot be opened in time — the case #4087 was actually built for.
2. **Deregister-on-teardown**, so the registry stops growing monotonically. A temp-dir root that vanishes should not leave a permanent registry entry.
3. **Surface registry size vs. loaded count in `/health`** — `indexes: 4` while the registry holds 396 gives an operator no signal that a backlog is the problem. `indexes_lazy` climbing is the only current hint and it is easy to misread as progress.
4. Consider walking real (existing-root) entries **first**, so a boot degrades gracefully instead of starving the working set behind cruft.

**Related, not duplicates:**
- **#4087** — introduced the park-for-lazy-load behavior. This issue is not a request to revert it; it is that parking is being applied to entries that could be settled instantly with a `stat`.
- **#4317** (open) — the registry-loss incident. This is a plausible contributing mechanism: a boot that has not finished walking the backlog presents a partial registry, and anything acting on that view sees indexes as missing.
- **#4250** (open) — a related but distinct warm-boot degradation shape (pre-#4087, timeout-skip with no retry/self-heal). Same daemon subsystem, different failure mode.
- **#4706** — `list_indexes.size_bytes` is a misleading metric. Related in kind: `list_indexes` showing 19 while the on-disk registry holds 396 meant a full estate audit earlier the same day measured the wrong number and concluded the registry was small.

**Operator note for whoever picks this up:** a 404 from `index_status` does NOT mean an index is empty — an unloaded index returns 404 whether it holds 0 chunks or 70,180. Any tooling that prunes based on a 404 will destroy live data. Root-path existence is the correct discriminator.

bobmatnyc:
Closes #4250
Closes #4846

## Measured cost, from the reporting machine's own registry

The owner's `indexes.toml` holds **80 entries, all `colocated = true`, of which 55 have a `root_path` that no longer exists** — matching the live symptom (25 of 55 indexes served).

Where the ~10 s per dead entry comes from: not a stat on a vanished path, not a redb open, not a lock wait. `restore_one_index` sees `!root_path.exists()`, calls `try_locate_moved_root`, which calls `scan_roots_for_colocated_indexes(&tracked_roots, DEFAULT_SCAN_DEPTH)` — a depth-5 recursive `read_dir` with a `canonicalize` per subdirectory, across every tracked root. Timed against the real `roots.toml` (248 roots, warm page cache, three passes):

```
MEASURE: 248 tracked roots
MEASURE: pass 0: scan_roots_for_colocated_indexes -> 19 entries in 10.512599167s
MEASURE: pass 1: scan_roots_for_colocated_indexes -> 19 entries in  9.566123250s
MEASURE: pass 2: scan_roots_for_colocated_indexes -> 19 entries in  9.555699041s
```

The walk's inputs and its 19-entry result are identical for every entry in a boot, and it ran **from scratch for each dead one**. 55 × ~10 s ≈ 9 minutes of duplicated walking, straddling the 10 s `warmboot_index_timeout()` — which is the "one entry per 10 seconds" #4846 reports and the timeout-skip #4250 reports, from one cause.

Why `orphan_reaper::heal_boot_orphans` reaped none of them: `partition_boot_orphans` filters on `!e.colocated`, and every entry on this machine is colocated. The exemption exists so the relocation scan can recover a moved repo; it is what left the entire dead population on the expensive path.

## Budget design: global for the dead cohort, per-index for live ones

The pre-fix budget was effectively per-index and unbounded in aggregate, so cost scaled with accumulated registry cruft rather than with real index count. This PR keeps the per-index deadline where it is correct and removes it where it was pathological:

1. **Triage** (`service/warm_boot/triage.rs`) settles each entry with one `exists()` stat and returns **two separate vectors**. The eager loops iterate `present` and have no access to `missing`, so "live indexes go first" is a property of the types, not of loop ordering a later edit could perturb.
2. **One walk per boot, not per entry.** `collect_relocation_candidates` hoists the scan out of `try_locate_moved_root`. The relocation *decision* is unchanged (0 → skip, 1 → relink, n → defer to the #4095 reaper); only where the candidates come from changed. `O(dead × roots)` becomes `O(roots)`.
3. **`SalvageBudget` caps the dead cohort globally** (`TRUSTY_WARMBOOT_SALVAGE_SECS`, default 30 s, `0` disables) and mints a `SalvageGrant` — a private-field token, the PR #4835 shape — that `collect_relocation_candidates` *requires*. A spent budget cannot pay for another walk.
4. **At exhaustion**, the remaining entries stay registered and untouched and are retried next boot. Because the salvage phase runs only after both eager phases finish, exhaustion is structurally unable to cost a live index anything.

`RelocationScan` deliberately has no "compute it yourself" variant, so no future edit can reintroduce the per-entry walk without adding a variant a reviewer has to confront.

## Why a skipped index was never retried, and how this relates to PR #4717

I read #4717. Its guard is right about what it covers and it is **not** the extension point here — but the reason is worth stating precisely, because it explains the whole failure.

`index_is_stuck_unwalked` fires inside `reconcile_one_index`, which `reconcile_stale_indexes` drives over `state.registry.list()`. **A timeout-skipped index never becomes a registered handle at all** — #4087 parks it in `ColdIndexStore` — so it is not a different *state* the guard misses, it is outside the population the guard iterates. Extending the predicate would change nothing.

What parking actually bought was reachability *by a query naming the id verbatim*. `list_indexes` omits cold entries, so a client that discovers indexes by listing never learns the id exists; that is why the owner's daemon served 25 of 55 for hours and why #4250's original report ran 13.5 hours. **The missing step is in front of #4717's guard**: getting the index back into the registry. Once it is there, #4717's guard and both staleness paths take over unchanged — which is exactly why the recovery pass registers rather than reindexes.

Root cause of "nothing could retry it": the cold store held *deferred on purpose* and *failed and papered over* in one map with nothing to tell them apart — the `bool`-shaped conflation PR #4718 removed from `BoundedRestoreOutcome`. No consumer could single out the failures without also loading every index `TRUSTY_WARMBOOT_MAX_INDEXES` deliberately deferred.

- `ColdReason::{Deferred, TimedOut}` is **assigned by the store**, never passed by a caller. `park_if_parkable` is the only route to `TimedOut` and already gates on the typed outcome, so an entry cannot be mislabelled at a call site.
- `service/timeout_recovery.rs` drains the `TimedOut` cohort only, every `TRUSTY_WARMBOOT_RETRY_SECS` (default 60 s, `0` disables). Not inline: #4087 documents that the abandoned blocking thread may still hold the redb file. Bounded at `MAX_RECOVERY_ATTEMPTS = 5`, after which the entry lands in the store's existing terminal `failed_entries` state and `/health` reports it under `indexes_failed`.
- `warm_boot_degraded` and `indexes_skipped_timeout` now derive from the **live** cohort. `recompute_warm_boot_degraded` treated the timeout counter as permanent on the stated premise that "a scan timeout genuinely CANNOT heal without a daemon restart" — true only because nothing retried. A daemon that got its indexes back stops reporting degraded; one that did not keeps a signal that falls or does not, instead of a frozen number that says nothing either way.

## Data safety

No registration is removed and no index data is deleted anywhere in this change. A missing root is a reason to spend less time on an entry, never a reason to destroy it — per #4846's operator note, an unloaded index 404s whether it holds 0 chunks or 70,180, so a failed probe is not evidence about the corpus. `disabled_salvage_budget_costs_a_dead_entry_nothing_but_a_stat` asserts the registry row count is unchanged after a boot that skipped 24 dead entries.

## Regression tests — verified to fail against current behaviour

Each was run against the pre-fix behaviour by neutering the fix in place, then restored.

| Test | Guards | Pre-fix result |
|---|---|---|
| `dead_entries_do_not_consume_the_live_index_budget` | a live index comes up on a boot dominated by dead entries, and the boot costs a small multiple of ONE walk | FAILED — `boot took 860.475917ms … ONE relocation walk costs 33.938083ms … Ceiling was 250ms` |
| `disabled_salvage_budget_costs_a_dead_entry_nothing_but_a_stat` | `SALVAGE_SECS=0` costs a dead entry one stat and no walk; every registration survives | FAILED — `boot took 914.848792ms against a single-walk cost of 33.043584ms` |
| `timed_out_index_is_driven_back_into_the_registry` | a timeout-parked index is registered again by one pass | FAILED — `one pass must bring the index back — nothing else ever will` |
| `warm_boot_degraded_clears_once_the_timeout_cohort_drains` | the flag un-latches after recovery | FAILED — flag stayed `true` |
| `warm_boot_degraded_stays_set_while_an_index_is_still_parked` | the un-latch does not weaken the signal | FAILED |
| `cold_store_timed_out_cohort_excludes_deferred_entries` | recovery never defeats `TRUSTY_WARMBOOT_MAX_INDEXES` | — |
| `recovery_pass_leaves_deferred_entries_alone` | same, end-to-end | — |
| `triage_*`, `salvage_budget_*`, `timeout_recovery_*` | pure decision truth tables | — |

The budget tests assert a **ratio measured in-process** (one walk timed in the same test), so they do not depend on this machine being any particular speed.

**One existing test changed its setup, not its assertion.** `health_reports_degraded_when_scan_timeout_recorded` set `indexes_skipped_timeout = 1` with an empty cold store — a state the daemon cannot reach, since `record_warm_boot_result` writes that counter and `park_if_parkable` parks the entry in the same warm boot. Once the counter is live, that setup asserted a daemon with nothing wrong must report degraded. It now parks a real timed-out entry, keeps the original assertion, and gains one checking the reported count is the live cohort.

## Verification

```
cargo fmt --check                                           EXIT=0
cargo clippy -p trusty-search --all-targets -- -D warnings  EXIT=0
cargo test -p trusty-search                                 EXIT=0
test result: ok. 1476 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
aggregate across all targets: passed=1886 failed=0
```

`scripts/check_line_cap.sh` (0 violations), `scripts/check_sld.sh` (0 errors), `scripts/check_test_pointers.sh` (21475 citations, 0 dangling), `scripts/check_changelog_fragment.sh` — all pass. No version bump: the crate stays at the unpublished 0.42.3.

Changelog fragment: `crates/trusty-search/changelog.d/4250-4846-warmboot-budget-and-retry.md`.

## Adjacent issues, deliberately not fixed here

- **#4317** (bulk registry sweep deregistered ~14 real indexes) — this PR touches the same warm-boot registry path but removes nothing, so it neither fixes nor worsens it. The `is_reapable_orphan` colocated exemption that left 55 dead entries on the expensive path is left in place: narrowing it is a deregistration change and belongs with #4317, not here.
- **#4253** (cold-start corpus-open race orphans an index) — untouched; `restore_index_on_demand`'s existing collision guards run unchanged on the recovery path.
- **#4846 ask 2, deregister-on-teardown** — not addressed. The registry still grows monotonically; this PR makes the growth cheap rather than stopping it. That fix belongs in the test harness that registers into the operator's live `indexes.toml` (also flagged in #4250's comment as a separate defect).
- A sibling session pruned the registry from 396 to 41 entries by hand earlier. That is a workaround, and the 80 entries measured here are what accumulated since — the reason a fix has to make dead entries cheap rather than rely on someone pruning them.


bobmatnyc:
## trusty-review verdict — WARN (proceeding to merge)

`mcp__trusty-review__review_pr` ran **degraded**: the reporting daemon at `:7878` was itself serving only 25/62 indexes (the exact symptom #4250/#4846 fix), so its RAG context was incomplete and its output is flagged non-authoritative. Grade D+ / REQUEST_CHANGES, 12 findings, none touching data-loss or deregistration.

Per the merge instructions, the grade was not treated as sufficient on its own — the diff was read directly at `87dec477` to verify the PR's five substance claims:

- **(a) Type-level triage, not loop ordering** — confirmed. `triage.rs`'s `TriagedEntries { present: Vec<..>, missing: Vec<..> }` is two separate fields; `commands/start/restore.rs`'s eager loops iterate only `legacy_eager`/`colocated_eager` (`= *_triaged.present`) and never touch `missing_roots`.
- **(b) One relocation walk per boot, budget-gated** — confirmed. `SalvageGrant { _mintable_only_by_budget: () }` has a private field; `collect_relocation_candidates` takes `&SalvageGrant` as a required parameter with no alternate constructor. `salvage_missing_root_entries` calls it exactly once per boot (inside `spawn_blocking`), wraps the result in `Arc`, and reuses the same `RelocationScan::Ready(Arc<..>)` clone for every entry in the cohort — the per-entry budget re-check (`budget.try_grant()` per iteration) gates continuing the loop, not re-running the walk.
- **(c) `ColdReason` is store-assigned only** — confirmed. `register_with_reason` is a private fn; its only two callers are `register_cold_entries` (hardcodes `Deferred`) and `park_if_parkable` (hardcodes `TimedOut`, gated on `BoundedRestoreOutcome`). No caller-supplied reason is possible.
- **(d) Registry row count invariant has a real assertion** — confirmed. `disabled_salvage_budget_costs_a_dead_entry_nothing_but_a_stat` asserts `after.len() == DEAD_ENTRIES + 1` after a boot that skipped all dead entries.
- **(e) Pre-existing test: fixture changed, assertion intact** — confirmed. `health_reports_degraded_when_scan_timeout_recorded` still asserts `status == "degraded"` unchanged; the diff only adds a real `park_if_parkable(..., TimedOut)` call (replacing an unreachable empty-cold-store setup) plus one new assertion on the live count.

Of the reviewer's 12 findings, direct reading showed two flagged "confirmed" issues in `health.rs` do not hold up: the "double-write TOCTOU/deadlock" claim assumes the outer `warmboot_summary` is a held `MutexGuard`, but it is an owned clone (`.lock().map(|g| g.clone())`) with no flush-back path in this file — no deadlock, no lost write. The "non-atomic snapshot" finding (`recompute_warm_boot_degraded` reading `cold_store.timed_out_len()` outside the `warmboot_summary` mutex) is real but benign: the two signals were never atomic with each other in a meaningful sense (independent subsystems), and the flag self-heals on the next health poll / recovery tick. The remaining findings are low-severity nits already visible in the diff (a redundant `Vec` clone in `try_locate_moved_root`, `unsafe` env-var mutation in tests under `#[serial_test::serial]`, a pre-existing `unwrap_or_default()` fallback on registry load, a narrow attempt-counter race in `timeout_recovery.rs` bounded by `MAX_RECOVERY_ATTEMPTS`) — none affect correctness of the core fix or data safety.

**Verdict: WARN.** Proceeding to squash-merge with `--admin` (bypasses only the 1-approval gate — all CI checks are green, including a clean re-run of `Test: doc-comment pointer lint` after issue #4687's known 5-minute-cap cancellation).

PR #4902, opened 2026-08-05
fix(trusty-search): bound xlsx decompression before calamine opens the package (#4894)

### LINKED ISSUE #4894: xlsx extraction has no pre-decompression bound — 511 KB peaks at 521 MiB RSS
## Problem

`crates/trusty-search/src/core/extract/xlsx.rs` carries a `RESIDUAL RISK` doc comment stating that `calamine` exposes no pre-decompression size bound. Quote:

> unlike the docx path (which bounds decompression via `docx::MAX_DOCUMENT_XML_BYTES`), calamine decompresses sheet XML internally and exposes no pre-decompression size check or bounded-reader seam to interpose on — a crafted spreadsheet under the 10 MiB `MAX_OFFICE_FILE_BYTES` container cap could still expand to a large in-memory sheet model before `MAX_EXTRACTED_TEXT_BYTES` truncation runs post-hoc in `extract_text`.

Contrast with `crates/trusty-search/src/core/extract/docx.rs`, which enforces a hard bound via `MAX_DOCUMENT_XML_BYTES` (50 MiB) + `Read::take` in `read_entry_bounded`.

## Measurement

An adversarial `zipbomb.xlsx` fixture — 511 KB container — extracts in 143 ms but peaks at 521 MiB RSS. Source: `crates/trusty-search/examples/anydoc_spike/MEASUREMENTS.md` on commit 547fdeaa (branch `spike/anydoc-extraction-evaluation`, PR #4875 closed).

## Existing mitigations (insufficient)

1. `MAX_OFFICE_FILE_BYTES` (10 MiB) bounds the container, not the decompressed payload — DEFLATE ratios can reach ~1000:1.
2. `EXTRACT_TIMEOUT` (30 s) is a time bound — the failure is memory, and extraction completes in 143 ms, well inside it.

## Tracking gap

This risk is currently documented only in a comment on closed issue #2923, so it is effectively untracked.

## Ask

Bound the decompressed size before or during sheet parsing, in the same shape `docx.rs` already uses.

---

bobmatnyc:
Closes #4894

## Defect

`crates/trusty-search/src/core/extract/xlsx.rs` handed the file straight to `calamine::open_workbook_auto` with nothing bounding how much it decompressed. A 511 KB adversarial workbook whose `xl/worksheets/sheet1.xml` expands to 512 MiB extracted **successfully**, in 143 ms, having allocated half a gigabyte on the way.

Neither existing mitigation applies:

- `MAX_OFFICE_FILE_BYTES` (10 MiB) bounds the **container**. This fixture's container is 511 KB.
- `EXTRACT_TIMEOUT` (30 s) bounds **time**. This completes in 143 ms.

The module's own `RESIDUAL RISK` comment documented the gap since #2923.

## Evidence

Measured with the `spike/anydoc-extraction-evaluation` fixture (`547fdeaa`, 512 MiB payload in a 511 KB container), peak RSS via `getrusage(RUSAGE_SELF)` in a dedicated child process:

| | peak RSS | result |
|---|---|---|
| before | 555,499,520 B (**529.8 MiB**) | `Ok` — extracted fine |
| after | 12,304,384 B (**11.7 MiB**) | `Err` — refused |

45x reduction. The "before" row is not copied from the spike report — it is this branch's own regression test run with the guard disabled, so the test is proven to fail against the vulnerable code:

```
zip-bomb extraction peak RSS: 555499520 bytes (529.8 MiB)
panicked at crates/trusty-search/src/core/extract/xlsx.rs:610:9:
extracting a 536870912-byte zip bomb peaked at 555499520 bytes RSS, over the
268435456 byte ceiling — decompression is unbounded
```

The test asserts peak RSS, not merely that extraction errors — an error-only assertion would have passed against the vulnerable code, which also returned successfully.

## Resolution

`preflight_package_bounds` runs before calamine ever opens the file.

**Why the check is outside calamine.** calamine owns its zip reading and offers no size limit, no bounded-reader seam, and no hook — `open_workbook_auto` takes a path and returns a parsed workbook. The zip layer does not bound it implicitly either: `zip` limits an entry read by its *compressed* length and lets the decompressor emit whatever that inflates to, and calamine streams sheet XML through quick-xml without ever consulting the declared uncompressed size. Of the candidates, pre-validating with the `zip` crate this crate already depends on is the only one that yields a real bound — bounding the declared uncompressed size alone is bypassable, since nothing enforces the declaration, and a wrapping reader has nothing to wrap.

**Two passes**, widening `docx::read_entry_bounded`'s declared-size + `Read::take` pair from one part to the whole package:

1. Central-directory metadata only. Summing declared uncompressed sizes refuses a declared bomb having decompressed nothing at all — the path the measured fixture takes.
2. A bounded drain. Declared sizes are attacker-controlled, so each entry is also streamed through `Read::take` into `io::sink()`. Discarding rather than buffering keeps this pass O(1) in memory whatever the cap is, so the cap governs only whether calamine runs, never how much the guard itself allocates.

## Threshold: 256 MiB total, not 50 MiB per part

Picked from measurement, not by analogy to docx. A dense inline-string workbook compresses at **18.5:1**, so a legitimate file at the 10 MiB container cap decompresses to ~190 MiB — and effectively all of it lands in the single `sheet1.xml` entry. docx's 50 MiB per-part cap would therefore reject real workbooks, and a per-part cap large enough not to would be redundant with the total. 256 MiB admits the densest realistic package with ~35% headroom while refusing the ~1000:1 ratios a DEFLATE bomb needs.

The package-wide shape is also the right one here: docx reads exactly one part, calamine reads the whole package, so the total is what bounds memory.

`test_large_legitimate_workbook_still_extracts` guards the other direction — a workbook decompressing to ~25 MiB (past the entire container cap, five times the extracted-text cap) still extracts, under the real production constant rather than a relaxed test-only one.

## `RESIDUAL RISK` comment

Rewritten rather than deleted, because something genuinely remains: the bound covers zip-based packages (xlsx/xlsm/ods) only. `.xls` is CFB/BIFF with no stream compression, so the container cap bounds it directly and there is nothing to expand. Separately, calamine's cell model for a package that passes the cap is still built in full before `MAX_EXTRACTED_TEXT_BYTES` truncates — bounded now, but proportional to the cap rather than to the 5 MiB of text actually kept.

## Verification

```
cargo fmt --check -p trusty-search                       EXIT=0
cargo clippy -p trusty-search --all-targets -D warnings  EXIT=0
cargo test -p trusty-search                              EXIT=0
  test result: ok. 1483 passed; 0 failed; 1 ignored
  test result: ok. 363 passed; 0 failed; 3 ignored
  (+ 23 further suites, all ok, 0 failed)
```

No `#[ignore]`, no cfg-gating, no version bump.


bobmatnyc:
**Not merged — the PR's own regression test fails on CI.**

`Test` job on head `72667e441eb4355fa5e513b2e41120d1aeec7dad` (workflow run 31015345951, job 92361091663):

```
zip-bomb extraction peak RSS: 317947904 bytes (303.2 MiB)

panicked at crates/trusty-search/src/core/extract/xlsx/tests.rs:432:5:
extracting a 536870912-byte zip bomb peaked at 317947904 bytes RSS,
over the 268435456 byte ceiling — decompression is unbounded

test result: FAILED. 1492 passed; 1 failed
```

`test_zip_bomb_extraction_stays_within_memory_bound` is this PR's own test for the fix it ships: 303.2 MiB against a 256 MiB ceiling, ~47 MiB over.

**This is not the known flake.** #4930 (`ollama_provider_emits_tool_call` / `auto_detect_returns_none_on_unreachable`) is a separate, diagnosed trusty-common issue. Verified distinct: 29 check rows, only `Test` red, and the three `skipping` rows are `main`-only conditional notify/parity jobs — nothing silently dropped.

**Direction is right, margin is not.** The changelog records 529.8 MiB before the fix, so this cuts peak by over 200 MiB and still misses the bound. It passed at the earlier head `d3b48539` and fails at `72667e44`.

**Likely cause — a review requirement, not the original implementation.** Closing the TOCTOU (guard and calamine must see the same bytes) meant `read_package_bytes` reads the whole package into memory where the previous code streamed from disk. Compounding it, pass 2 drains every entry to `io::sink()` and calamine then inflates the same bytes again — a double-inflate inherent to validating outside calamine, noted as a LOW during review.

**Options for whoever picks this up**, in no particular order:
- Raise the ceiling, with a justification for what the real bound should be.
- Reduce peak: stream the drain, and avoid the transient `Vec` → `Arc<[u8]>` copy.
- A different TOCTOU closure that does not require full in-memory buffering.

Please do not simply raise the ceiling to make CI green without establishing what the correct bound is — the test is the fix's only proof that it works.

**What IS verified sound**, across two independent review rounds, so it does not need re-deriving:
- The cap is enforced on decompressed bytes as they stream — forged central-directory *and* local-header size fields are still caught.
- The TOCTOU is genuinely closed: exactly one `File::open` survives in production code, and the `.xls`/CFB route was verified end-to-end with a real file.
- No off-by-one at the size boundary (`MAX-1` and `MAX` accepted, `MAX+1` rejected).
- Entry cap (16384) covered in both directions, and the reject test cannot pass for the wrong reason.
- Three mutants each killed by exactly one test, with control runs proving 17/17 pass without those tests.
- File split (`xlsx/mod.rs` 99 SLOC + `xlsx/tests.rs` 541 SLOC) is a verified pure move; line-cap green.

The code is correct in every respect except that it does not meet its own memory bound.

Issue #1970, opened 2026-07-03
trusty-memory: decouple text/KG recall+write from embedder warm-up (should behave like trusty-search)


bobmatnyc:
## Problem Observed

During a live PM session (2026-07-03), a call to `mcp__trusty-memory__memory_remember` failed twice with:

```
MCP error -32603: trusty-memory is warming up (embedder initialising); please retry in a few seconds
```

Meanwhile, in the same session, `trusty-search`'s hybrid BM25+vector search (`mcp__trusty-search__search`) returned relevant results successfully even during a similar cold-embedder condition. **trusty-search degrades gracefully; trusty-memory returns an explicit error instead.**

## Directive from Bob

trusty-memory should architecturally behave like trusty-search: **text/BM25 lookup and knowledge-graph (KG) queries should be FAST indexing operations that return results immediately regardless of embedder readiness**, with embedding/vector-index work happening in parallel/background.

Currently, trusty-memory appears to hard-block on embedder initialization instead of falling back to BM25/KG-only results and backfilling the vector/semantic layer asynchronously.

## Desired Behavior

- `memory_recall` (and KG-query paths) should **return BM25 + knowledge-graph results immediately** even if the embedder is still cold — semantic/vector results can be omitted or backfilled once available, but the call should not block/error on embedder state.

- `memory_remember` (writes) should **not require the embedder to be warm** to accept and persist a write — text/KG indexing should happen synchronously (fast), and embedding generation should be queued/async, updating the vector index in the background once ready.

- This should **mirror trusty-search's daemon** embedder-readiness handling (likely in daemon/search-service layer that falls back to lexical-only results when the embedder isn't ready).

## Investigation Needed (Before Implementation)

Understand the architectural difference:

1. **How trusty-search handles embedder-readiness**: Look at `crates/trusty-search` — find how the daemon structures embedder checks and falls back to BM25/lexical results when embedder is cold.

2. **How trusty-memory gates on embedder state**: Look at `crates/trusty-memory` — find where "warming up"/embedder-initialising errors are raised in `memory_recall`/`memory_remember` MCP handler paths.

3. **Compare patterns**: Identify why trusty-search gracefully degrades while trusty-memory hard-blocks.

4. **Reference the error string**: Search both codebases for "warming up" error message to locate the gating logic.

Once the architectural pattern is clear, implement trusty-memory to behave like trusty-search: synchronous text/KG indexing (fast path), asynchronous embedding backfill.

PR #4903, opened 2026-08-05
fix(trusty-memory): make memory_recall answer the query it was given (#4836)

### LINKED ISSUE #4836: memory_recall and memory_recall_deep return identical results regardless of query (scores always 1.0)

bobmatnyc:
## Defect

`memory_recall` and `memory_recall_deep` returned the same drawers for every query. Reproduced before touching any code — three unrelated queries against the live `trusty-tools` palace returned an identical 5-drawer list, in identical order, every entry at `score: 1.0`:

```
drawer_ids, all three queries:
  005bf5ab, 021e7c1e, 09b560e4, 0b616513, 0da3dcfa
```

Ascending UUID order, not a ranking. Every result was `layer: 1` with `score` exactly equal to `importance` — so `rescore_l1_by_similarity` never ran, which means the vector leg never ran.

## Where the query was lost

**`crates/trusty-memory/src/tools/memory_ops.rs:417`** — the readiness gate. When it reads `Warming`, the handler takes `recall_without_embedder`, which returns L0/L1 ranked by raw importance and never reads the query beyond an optional BM25 lane (unavailable here). The query is not dropped in a mapping layer; it is structurally unused on that branch.

The gate was reading a flag with exactly one writer: the startup warm-up task at `crates/trusty-memory/src/main.rs:919`. A single failed attempt left it `Warming` permanently. Measured on the live daemon:

```
$ curl -s http://127.0.0.1:7070/health
{"status":"ok","version":"0.22.0","uptime_secs":24680,...,"daemon_state":"warming"}
```

6.8 hours of uptime, still `warming` — while the same process, at the same moment, served correctly ranked results over HTTP, which has no readiness gate:

```
$ curl -sG .../palaces/trusty-tools/recall --data-urlencode "q=worktree discipline branch cleanup git"
  0.4483 L2 'GIT FAIL-OPEN AUDIT COMPLETE 2026-08-04 ...'
  0.4262 L2 '"git stash is repo-level (shared across all worktrees) ...'
$ curl -sG .../palaces/trusty-tools/recall --data-urlencode "q=crates.io publish version bump release tagging"
  0.4948 L1 'STANDING INSTRUCTION (Bob, 2026-07-16): tm 0.19.18 r...'
  0.4276 L2 'Standard workflow rule (Bob, 2026-07-07) — cargo publish to crates.io ...'
```

That confirms the issue's framing: the store and the search were fine.

The flag could never recover on its own, because a recall that degrades never calls `embedder()` — the path that reads the flag could not produce the event that clears it.

## Resolution

Two things kept the flag from describing reality:

- **`AppState` carried its own embedder `OnceCell`**, separate from `retrieval::shared_embedder()`. Startup latched readiness off one cell while every recall consumed the other, so the flag described an embedder the request path never touched — and the daemon loaded two ~90 MB ONNX sessions. `AppState::embedder` now delegates to the shared cell, and resolving it marks the daemon ready.
- **The recall handlers now ask whether the embedder is live** rather than trusting the latch alone, which breaks the deadlock. This never triggers a cold init, so issue #1970's "never block a recall on embedder warm-up" guarantee is unchanged.

Net: `-36 / +381` across 8 files, of which 271 lines are the new test.

## Tests

`crates/trusty-memory/tests/recall_query_discrimination.rs` — its own integration binary, so seeding the mock embedder cannot race the lib-test binary's `dispatch_remember_then_recall`, which deliberately uses the real singleton. Each test reproduces the live condition: a working embedder behind a stranded `Warming` latch.

All four fail on `origin/main` and pass with this change.

| Test | On main |
|---|---|
| `different_queries_return_different_drawers` | FAILED — identical (empty) sets |
| `a_query_matching_a_drawer_returns_that_drawer` | FAILED — got nothing |
| `deep_recall_also_discriminates_by_query` | FAILED — identical (empty) sets |
| `resolving_the_embedder_marks_a_warming_daemon_ready` | FAILED — `left: Warming, right: Ready` |

The fourth pins the root cause at its own level: no recall involved, just "resolve the embedder, assert the latch cleared."

## Gates — ladder rung 5

Cross-crate correctness fix on a data-retrieval path other work depends on.

```
cargo fmt --check                                                   EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings          EXIT=0
cargo clippy -p trusty-common --features memory-core,\
    embedder-test-support --all-targets -- -D warnings              EXIT=0
cargo test -p trusty-memory -- --test-threads=1                     498 passed, 2 failed (pre-existing, see below)
cargo test -p trusty-common --features memory-core,\
    embedder-test-support -- --test-threads=1                       830 passed, 0 failed
cargo test -p trusty-memory --test recall_query_discrimination      4 passed, 0 failed
cargo check --workspace                                             baseline Tauri failure (see below)
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui  EXIT=0
bash scripts/check_line_cap.sh                                      EXIT=0
bash scripts/check_sld.sh                                           EXIT=0
bash scripts/check_changelog_fragment.sh                            EXIT=0
```

**Two pre-existing failures, not caused by this branch.** `commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener` and `..._fails_when_no_addr_file_and_no_listener` fail identically on unmodified source (verified by stashing this diff and re-running: `0 passed; 2 failed`). They probe for an absent daemon and this machine has a live one on `:7070`. This diff touches zero files under `crates/trusty-memory/src/commands/`.

**`cargo check --workspace`** fails on the briefed baseline: a Tauri `frontendDist` proc-macro panic in `trusty-code-gui` (`ui/dist` absent). This diff touches zero GUI files; excluding the two GUI crates, the workspace checks clean.

`trusty-common`'s change is purely additive (one new `pub fn` plus its test), so no dependent's behavior can change; `cargo check --workspace` covers the compile surface.

## On the second finding — a different bug, not this one

The issue's recall-precision finding (0 of 17 curated facts cleanly matched across 1,114 hook firings) is **separate from this defect and not addressed here.**

The reasoning: that measurement was taken from the HTTP hook path, which was never affected by the bug fixed here. Its 148 distinct injection variants prove it was discriminating throughout. So its precision problem lives in ranking and relevance — L1 importance weighting, the `L1_NO_SIMILARITY_PENALTY` coefficient, closet tag boost, `expand_query`'s four hardcoded synonym groups — none of which this PR touches. Fixing the query-dropping bug will not move that number.

Worth filing separately. One concrete lead for whoever picks it up: `expand_query` (`crates/trusty-common/src/memory_core/retrieval/layers.rs:403`) hardcodes four trigger groups — speed, vector search, memory safety, concurrency — and is a no-op for everything else, which matches the observed pattern that standing behavioral preferences never surfaced.

## Note for the operator

The running daemon still has the stale latch; this fix reaches it on the next restart with a rebuilt binary. That restart is not part of this PR.

Closes #4836

---

## Review follow-ups (second commit)

- **`/health` reported the raw latch** (`web/health.rs:308`). The exact misleading signal that made this bug hard to find survived the original fix, because the HTTP path never calls `AppState::embedder`. Now reports the embedder's real state.
- **Doc block fused** onto `vector_lane_available`, leaving `recall_without_embedder`'s ADR-0027 scope-leak rationale undocumented. Split back onto their own items.
- **Three #1970 degraded-lane tests had become order-dependent** — `dispatch_remember_then_recall` initialises the process-wide embedder cell in the same lib-test binary, so whether they exercised the degraded lane depended on execution order. They passed either way, because "does not error" is true of both lanes. Moved to `tests/recall_degraded_lane.rs` (own process, never initialises an embedder); each now asserts the cold-embedder **precondition** and the actual degraded contract (no layer-2/3 hit), so they fail loudly rather than decaying into no-ops again.

## `defer_embedding` investigation — a real gap, and it has already fired

`defer_embedding` (`memory_ops.rs:42`/`:185`, `task_ops.rs:41`) reads the raw latch, so under a stranded `Warming` latch **every** MCP write took the deferred path and was vector-indexed only by `spawn_deferred_embed`.

**The failures are silent.** `handle.rs` `spawn_deferred_embed` is `tokio::spawn` fire-and-forget; every failure mode — embedder init failure, embed timeout, upsert error — is `tracing::warn!` and `return`. No retry, no queue, no persisted "needs embedding" marker. A drawer is durable in KG/redb but permanently invisible to vector recall, and nothing records that.

**It has already happened on this machine.** Measured live:

```
drawer_count: 1241
vector_count: 1202      -> 39 drawers (3.1%) have no vector
```

I sampled 71 drawers (evenly across the corpus plus the 20 newest), self-retrieving each with its own verbatim text at `top_k=150`. **9 were not retrievable at all.** One of them, `1509bc35`, was written at `2026-08-05T12:27:00.278Z` — inside the migration batch itself.

The other 13 of the 14 migration-batch drawers *are* embedded (all `layer: 2`), so the backfill mostly worked; `5dfc4721` looked missing at `top_k=8` but is embedded at rank 13 (score 0.309) — that one is [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904) relevance, not embeddedness. But the failure rate is not zero, and the failures are invisible.

**Migration-blocking consequences:**

1. A re-embed / backfill pass over drawers lacking vectors is **required before** [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834) deletes any source file. Restarting the daemon with this PR's fix does **not** repair already-written drawers — nothing re-attempts a dropped embed.
2. `vector_count` vs `drawer_count` is a cheap, exact detector. It should gate the migration, and arguably belongs in `doctor`.
3. Verifying retrievability via `memory_recall` alone is insufficient even once this PR lands: a drawer can be present, durable, and permanently unfindable.

**Judgment on scoping: separate PR, not this one.** Same latch, but a different defect class and a different fix. This PR makes the readiness signal honest; the deferred-embed gap needs a durable work record (or a startup reconciliation pass) plus a backfill for existing drawers — new persistent state, not a flag correction. Folding it in would put a data-repair migration inside a correctness fix and blur what the tests here prove. The 39 unembedded drawers exist today and are unaffected by whether this PR merges.


bobmatnyc:

Adversarial review, rung 5. Independently reproduced fail-before / pass-after in throwaway worktrees at `1db7c8b3` and `origin/main`; both worktrees removed.

```
origin/main + the new test file copied in:
  test result: FAILED. 0 passed; 4 failed
    resolving_the_embedder_marks_a_warming_daemon_ready
      left: Warming, right: Ready
    different_queries_return_different_drawers
      #4836: recall returned NO results for the marsupial query against a populated palace
    a_query_matching_a_drawer_returns_that_drawer
      query: The Voyager probes each carry a gold-plated record of sounds from Earth
      got:
    deep_recall_also_discriminates_by_query
      #4836: recall returned NO results for the sourdough query

PR head 1db7c8b3:

  test-pointers: resolved 21479 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```

### Root cause: confirmed, not accepted on assertion

Traced independently. `retrieve_l0_l1` (`crates/trusty-common/src/memory_core/retrieval/layers.rs:70-104`) emits `score: 1.0` for the identity and `score = drawer.importance, layer: 1` for each L1 drawer, and never reads the query. That is exactly the issue's fingerprint — five results, `layer: 1`, `score` equal to `importance`, UUID order. The gate at `memory_ops.rs:442` routed there.

The two-cell claim holds. Startup calls `trusty_common::memory_core::retrieval::shared_embedder()` at `main.rs:912` and latches readiness off it at `919`. The HTTP `/recall` route reaches the same shared cell — `crates/trusty-memory/src/web/` contains zero references to `readiness()`, `shared_embedder`, or `state.embedder`, which is why that path never degraded. Old `AppState::embedder()` built a private `OnceCell<Arc<FastEmbedder>>` with its own `FastEmbedder::new()`. Two cells, two ~90 MB sessions when both paths ran. After the change, `crates/trusty-common/src/memory_core/retrieval/embedder.rs:71` is the only `FastEmbedder::new()` reachable from the daemon (`commands/setup.rs:315` is the one-shot CLI `setup` command in its own process), so it is one session — not zero and not two.

**The query is not dropped anywhere else on the fixed path.** `recall_scoped` (`layers.rs:501-542`) runs `expand_query(query)` into `retrieve_l2_scoped` and rescores L1 by similarity. No second degraded branch hides behind this one.

**Readiness cannot latch `Ready` falsely.** `AppState::embedder()` (`lib.rs:1212`) propagates with `?` before `set_ready()` at `1218`, so the flip only follows a resolved embedder. A missing or corrupt ONNX model errors out of `shared_embedder()`, `get_or_try_init` leaves the cell uninitialised, `shared_embedder_initialized()` stays false, and every recall degrades — the correct outcome, and strictly better than the old code, which on a `Ready` daemon would have errored the recall outright.

**No deadlock or startup-order hazard.** The shared cell's initialiser is `FastEmbedder::new()`, which never re-enters `AppState`. `OnceCell::initialized()` is false while an init is in flight, so a request arriving mid-warm-up still degrades and never blocks — #1970's guarantee holds. And in production `Ready` implies the cell is initialised (the only two `set_ready()` callers outside `#[cfg(test)]` are `main.rs:919` and `lib.rs:1218`, both post-success), so the `readiness() != Warming` arm can never trigger a cold init on the request path.

`recall_without_embedder` stays reachable for genuine no-embedder deployments — warming latch plus uninitialised cell — and still returns L0/L1 + BM25 rather than erroring.

### Findings

|---|---|---|---|---|---|
| MEDIUM | crates/trusty-memory/src/tools/memory_ops.rs | 300-338 | `vector_lane_available`'s doc was appended to `recall_without_embedder`'s existing doc block instead of placed after it. Lines 300-334 are one contiguous `///` run, so the whole thing attaches to `vector_lane_available` — which now documents itself as "Degraded-embedder recall fallback (issue #1970): L0/L1 + BM25 only" and carries the ADR-0027 T7 room-scoping rationale. `recall_without_embedder` at 340 is left with zero documentation, including the scope-leak reasoning ("a scope boundary failing open, which is a leak") that explains why its BM25 lane filters. | Move the `fn` and its doc above line 300, or below the closing brace of `recall_without_embedder`. | Fix here |
| MEDIUM | crates/trusty-memory/src/lib.rs | 1185-1220 | The rustdoc on the `pub` `AppState::embedder` is now false and silent about the new side effect. It says "Obtain the shared `FastEmbedder` instance" (returns `Arc<dyn Embedder + Send + Sync>`), "wraps `OnceCell::get_or_try_init` with a timeout" and "If this timeout fires the `OnceCell` is left unresolved and the next call retries from scratch" (no wrapper timeout exists any more), and "every recall handler now checks `readiness() == Ready` first" (they check `vector_lane_available`). Most importantly, nothing in the doc block says this method now MUTATES daemon readiness — only an inline `//` comment in the body does, and rustdoc renders the block, not the body. A `pub async fn embedder()` that flips global daemon state is a surprising side effect on a crate whose `[lib]` exists so external hosts can link it. No `Test:` line either, though the PR added `resolving_the_embedder_marks_a_warming_daemon_ready` for exactly this method. | Rewrite the block: correct the return type, drop the timeout paragraphs, name `vector_lane_available` as the request-path gate, state the readiness side effect, add the `Test:` pointer. | Fix here |
| MEDIUM | crates/trusty-memory/src/tools/tests.rs | 1900, 1927, 1956 | The three #1970 degraded-path tests are now order-dependent and no longer cover the branch they were written for. They build a `Warming` state, but `vector_lane_available` also consults the process-global `SHARED_EMBEDDER`, and `dispatch_remember_then_recall` (same lib test binary, tests.rs:254) deliberately initialises the real singleton via `test_state()` → `memory_remember` → `remember_with_options` with `defer_embedding == false` → `shared_embedder()`. Once that runs, the three tests take the VECTOR lane. They still pass because they only assert "does not error", so the erosion is silent. Net: after this PR the crate has no deterministic test of the degraded path — and the pointers claiming otherwise (memory_ops.rs:316-317, `recall_falls_back_to_bm25_and_l0_l1_while_warming` and siblings) name tests that do not exist and are ratchet-allowlisted at `.test-pointer-allowlist.tsv:335-337`. This is the reverse direction of the isolation problem the PR correctly solved for its own tests. | Give the degraded path its own integration binary — the same technique `recall_query_discrimination.rs` already uses — asserting the fallback lane actually ran (e.g. `score == importance`), with the shared cell never seeded in that process. | Fix here |
| MEDIUM | crates/trusty-memory/src/web/health.rs | 154-162, 308-311 | `/health`'s `daemon_state` still reports the raw latch, which after this change no longer describes whether recall degrades. This is the PR's own diagnostic evidence left unfixed: a hook-only daemon (HTTP `/recall` never calls `AppState::embedder`) whose warm-up failed once will report `warming` for its entire life while serving full vector recall — precisely the `uptime_secs: 24680, daemon_state: "warming"` reading quoted in the PR body. The field doc "warming until the embedder init succeeds" becomes wrong in both directions. | `let daemon_state = if state.readiness() == Ready \|\| shared_embedder_initialized() { "ready" } else { "warming" };` — the accessor this PR just added — and update the field doc. | Fix here |
| MEDIUM | crates/trusty-memory/src/lib.rs | 1212-1218 | The fix does not fully deliver its own stated root cause. It adds a second readiness writer, but that writer is reachable only once `SHARED_EMBEDDER` is already initialised. If the startup warm-up fails and leaves the cell uninitialised, `vector_lane_available` is false, the degraded path never calls `embedder()`, and MCP recall stays query-independent indefinitely — the same deadlock, one level down. The retry channels that break it (`spawn_deferred_embed` on a write, HTTP recall, the dream cycle) are all off the MCP recall path. The observed incident (cell live, latch stale) is genuinely cured and the tests prove it; the failure mode that produces it is not. In the real `tm` deployment the hook fires HTTP recall constantly, so the bad state is close to unreachable — which is why this is MEDIUM and not higher. | A bounded retry loop with backoff in the `main.rs:909` warm-up task instead of one shot. | Parent |
| LOW | crates/trusty-memory/changelog.d/4836-memory-recall-query.md | 1 | Category is `Fixed`, but the PR removes `pub embedder` from the `pub struct AppState` and changes `pub async fn embedder()`'s return type — SemVer-major for a crate whose `[lib]` is explicitly there so external hosts can link it. Nothing actually breaks: the sparse index tops out at `trusty-memory 0.21.2`, so the unpublished `0.22.0` already satisfies cargo's 0.x-minor rule and `check-pr-version-bump.sh` passes. But the assembled 0.22.0 release notes will not flag the break. | Add a `Breaking` fragment naming the removed field and the changed signature. | Fix here |
| LOW | crates/trusty-memory/src/lib.rs | 585-588 | The `daemon_readiness` field doc still says the flag is "flipped to `Ready` by `spawn_startup_tasks` after the embedder warm-up succeeds" and that "the transition is one-way and lock-free". The PR added a second writer and updated `readiness()`'s doc at 1161-1165 but not the field's own — leaving the single-writer claim, which is the whole defect, asserted 600 lines above the fix. | Add `AppState::embedder` as the second writer. | Fix here |

### Checks that passed

- Changelog fragments present with a bare category on line 1: `crates/trusty-memory/changelog.d/4836-memory-recall-query.md` (`Fixed`) and `crates/trusty-common/changelog.d/4836-shared-embedder-initialized.md` (`Added`).
- `// #4836:` attribution comments at every change site.
- **No test weakened, `#[ignore]`-ed, `cfg`-gated, or deleted.** Every `-` line in the diff is the old timeout wrapper or one of the three `readiness() == Warming` gate expressions; zero `#[ignore]` or `cfg(not(...))` additions. MEDIUM-3 above is a *semantic* weakening of three existing tests, not a textual one — no line of them changed.
- `resolving_the_embedder_marks_a_warming_daemon_ready` pins the root cause at its own level with no recall in the path — confirmed by its `left: Warming, right: Ready` failure on main.
- The three recall tests do not rest on incidental ordering. `drawer_ids` carries an explicit non-emptiness assertion, so "returned nothing" fails loudly rather than satisfying `assert_ne!` vacuously — and on main that is exactly how they failed. `a_query_matching_a_drawer_returns_that_drawer` anchors the affirmative form, so a ranking change would have to make a drawer unfindable by its own text to break it.
- The separate integration binary is sound isolation for the new tests, not papering: `MockEmbedder` must be seeded before any `shared_embedder()` call, and the lib binary contains a test that deliberately uses the real singleton. One process cannot serve both.
- Baseline exclusions honoured: the two `commands::doctor::tests` failures and the Tauri `frontendDist` panic were not attributed to this branch.

### One line for the separate precision issue, not developed here

Under the stranded latch, `defer_embedding` at memory_ops.rs:42/185 and task_ops.rs:41 was true for every MCP write, so months of drawers were vector-indexed only by the fire-and-forget `spawn_deferred_embed` backfill — worth checking whether that backfill's failures are silent before concluding the "0 of 17 curated facts" number is purely a ranking problem. Those three sites still read the raw latch, not `vector_lane_available`.

### Verdict

Zero CRITICAL, zero HIGH. The diagnosis is correct, the fix is minimal and targets the actual mechanism, the tests fail on main for the right reasons and pass here, and the resource claim checks out. **APPROVE.** The MEDIUMs are worth landing — the memory_ops.rs doc merge and the `AppState::embedder` rustdoc are small and belong in this PR; the health.rs report and the degraded-path test are the ones that keep this from recurring undetected.

Issue #4878, opened 2026-08-05
bare `tm` registers no tracing subscriber — deploy warnings are silently dropped on the in-place relaunch path


bobmatnyc:
## Defect

`tm` registers a tracing subscriber only for the daemon, supervisor, MCP, and `tm sessions instructions` paths (`tracing_setup.rs:37`, `main.rs:268`). Every other in-process path silently discards its diagnostics. Confirmed affected, each running a deploy in-process:

- bare `tm` launch (`commands/launch.rs:196`)
- `tm doctor --fix-skills`
- `tm catalog apply`
- the in-place relaunch path (`main.rs:405` → `guided.rs:171` → `guided_inplace.rs:591` → `claude_code.rs:948`)

## Why It Matters

Deployment provisioning runs on these paths and legitimately declines files — a checksum-frozen skill, for instance. The whole point of [PR #4876](https://github.com/bobmatnyc/trusty-tools/pull/4876) (and of [PR #4848](https://github.com/bobmatnyc/trusty-tools/pull/4848) before it, for agents) is that a silent decline leaves a file stale forever with no signal. On all four paths above, warnings are emitted into nothing, so the silent-staleness the two PRs were written to end persists. This is the same failure class as [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840): correct source, correct binary, stale deployed file, no error surfaced.

## Evidence

Found during code-critic review of [PR #4908](https://github.com/bobmatnyc/trusty-tools/pull/4908) (2026-08-05). That PR adds an `error!` for unreadable skill ledger (`skills/manifest.rs:213`) plus three merge `warn!`s — those are the signals that a ledger was overwritten or a writer raced. Combined with [PR #4876](https://github.com/bobmatnyc/trusty-tools/pull/4876)'s declined-skill warning and [PR #4848](https://github.com/bobmatnyc/trusty-tools/pull/4848)'s agent-deploy warnings, there are now at least three separate diagnostics written to end silent staleness, all muted on the four paths an operator actually runs by hand.

Pre-existing; the warnings ARE reachable on the daemon path (daemon runs at default filter `info` writing to stderr plus the rotating log, never stdout). Only the bare-`tm` family drops them.

[...]

PR #4908, opened 2026-08-05
fix(trusty-agents-common): serialise the skill manifest's read-modify-write (#4881)

### LINKED ISSUE #4881: skill manifest save has no lock — concurrent sessions freeze skills nobody edited

bobmatnyc:
Closes #4881

> **Review round 2.** The first version of this PR was BLOCKED for a CRITICAL: its compare-and-swap returned `Err` after every `SKILL.md` was already on disk, freezing the whole tier. That is fixed below, and two claims this body previously made were wrong — both are corrected here rather than quietly dropped.

## Defect

`SkillManifest::load`/`save` took no lock, so overlapping deploy runs raced on load-modify-save. One process wrote a skill and saved its manifest; a slower one then saved a snapshot taken BEFORE that write, dropping the entry. The recorded checksum for that key then named an older revision than the bytes on disk, and `deploy_one_file` read the mismatch as "the user hand-edited this" and skipped the file permanently. A skill nobody touched, frozen against every future update, silently.

Measured, not theoretical: three skills were found frozen on 2026-08-05, each byte-identical to an official historical main commit.

**The race widened.** [#4882](https://github.com/bobmatnyc/trusty-tools/pull/4882) took the manifest writers from two to four. And under the owner's lifecycle ruling the local manifest is now the durable record of what the user customized — a clobbered manifest can corrupt that record, not just freeze a file.

## Mechanism

**The lock is the fix.** `with_skill_manifest_lock` — an advisory `flock` sidecar, the skill-side counterpart of the agent ledger lock from [#4409](https://github.com/bobmatnyc/trusty-tools/issues/4409) — wraps the whole read-modify-write in all four writers: `deploy_skills_filtered`, unmanaged-skill adoption, `tm doctor --fix-skills` repair, and `tm catalog apply --prune`. It is held across the manifest load, the file writes it authorises, and the save; not across tier planning, and nothing slow runs inside it. `flock` blocks rather than spins and no writer holds it across anything unbounded, so there is no starvation path. Each public entry takes it once and every locked body is a `*_locked` function never called from inside another, so nothing self-deadlocks on the non-reentrant lock.

**The save must always publish.** Every writer writes files BEFORE it saves. Bytes newer than their recorded checksum are precisely what the deployer reads as a hand-edit and skips forever — so a save that fails or refuses after files are on disk manufactures the freeze this issue exists to prevent, across the whole tier. `SkillManifest::save_merging` is a three-way merge: when the on-disk document has moved since the caller's base, it starts from what is on disk and replays this run's delta (inserts, updates, and removals) onto it, then publishes. It always publishes.

Merging also beats a plain `save`, which is why it is not just a safety valve: a blind save drops the racing writer's entries and freezes ITS files instead. That matters concretely during rollout, when older installed `tm` binaries do not take the lock at all.

Also consolidated: the `flock` critical section is now one implementation (`with_ledger_lock`) that both ledgers' wrappers delegate to.

## What changed since the BLOCK

| Finding | Resolution |
|---|---|
| **CRITICAL** — `Err` after files written froze the tier | `save_if_current` replaced by `save_merging`; there is no longer any variant meaning "nothing written". Regression test below. |
| **MEDIUM** — CAS on 1 of 4 writers | All four now use `save_merging` against a base snapshot. |
| **MEDIUM** — 3 `trusty-mpm` behaviour changes, 0 tests | Two tests added, each proven to fail on revert. The third change (`prune_skills` locking) is covered by `apply_prune_removes_deselected`. |
| **LOW** — unreproducible fail-before number | Re-derived honestly below. |
| Two false doc claims | Deleted with the code they described. |

## Tests

**The CRITICAL, as a regression test.** `a_racing_writer_freezes_nothing_on_either_side` replays the deployer's own ordering — load, write files, save against a base a competing writer has since superseded — and then lets the REAL deployer classify the result, because its classifier is what freezes. It catches both failure modes, verified by breaking `save_merging` two ways:

```
# refuse-after-write (the blocked version):
  a racing writer must freeze nothing: skipped=["example-skill", "tm-doctor"]
# blind save:
  the racing writer's entry must survive, or ITS file freezes instead
```

The first line is the critic's finding reproduced in-tree: the whole tier skipped.

**Fail-before, re-derived.** My earlier claim that neutering the lock yields `left: 1, right: 8` was wrong as stated — that is one of two failure modes, not the failure mode. Neutered and run 5×: every run fails, but only runs 1 and 3 show lost updates (`every writer's entry must survive: ["skill-6"]`, `left: 1 right: 8`); runs 2, 4 and 5 die earlier inside `atomic_write` with `Io(Os { code: 2, kind: NotFound })` — eight threads racing `temp_path`, whose pid+nanosecond scratch name can collide at that concurrency. The honest statement is that the test fails without the lock every time, by one of two mechanisms. (The scratch-name collision is a pre-existing #4409 sharp edge, not in scope here; the lock makes it unreachable on this path.)

`deploy_blocks_while_the_skill_ledger_lock_is_held` holds the lock and probes a concurrent deploy for 2s — an unlocked deploy of two tiny files finishes in under a millisecond so the probe cannot pass by accident, and a deploy blocked on `flock` stays blocked as long as the lock is held so it cannot fail by accident. The merge tests construct staleness directly, with no scheduling involved.

**The two `trusty-mpm` tests, each failing on revert:**

```
a_lock_sidecar_alone_is_not_a_populated_tier ... FAILED
  left: AbsentButPopulated   right: AbsentAndEmpty
repair_leaves_a_missing_tier_directory_alone ... FAILED
```

Negative regression guards, unchanged and still passing: `deploy_user_owned_skipped`, `deploy_skips_user_modified`. `deploy_one_file`'s skip rules are untouched — no frozen file is unfrozen, no fix-skills command added.

## Side effects of the new sidecar

`skill_drift`'s manifest-less-tier check counted any directory entry as content, so a lone `.lock` would report a never-deployed tier as unverifiable; it now ignores dot-prefixed entries. And `repair_skills` skips a non-existent tier before taking the lock, so a repair never creates an empty `skills/` directory in a project doctor was only inspecting. Checked and safe unchanged: `unmanaged.rs` and `count_skill_entries` (directories only), `stale_skills` (`tm-` prefix), `deploy_validate` (manifest presence).

## Gates — Rust Test Ladder rung 5 (cross-crate, shared library)

Rebased on current main (picks up #4907, which fixes the doc-pointer lint timeout).

| Gate | Result |
|---|---|
| `cargo clippy -p trusty-agents-common --all-targets -- -D warnings` | EXIT=0 |
| `cargo test -p trusty-agents-common -- --test-threads=1` | 463 passed, 0 failed |
| `cargo test -p trusty-mpm -- --test-threads=1` | 4604 + 1438 + 1438 passed, 0 failed |
| `cargo test -p trusty-agents --lib -- --test-threads=1` | 3341 passed, 0 failed |
| `cargo test -p trusty-code -- --test-threads=1` | 1600 + 21 passed, 0 failed |
| `bash scripts/check_changelog_fragment.sh` | EXIT=0, both crates recorded |
| `cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui` | EXIT=0 |

Direct dependents of `trusty-agents-common`, from `crates/*/Cargo.toml`: `trusty-mpm`, `trusty-agents`, `trusty-code` — all three tested. The GUI exclusions are the documented pre-existing Tauri `frontendDist` red; `git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/` returns 0 files.

`manifest.rs`'s inline tests moved to `manifest_tests.rs`, matching the convention every sibling in the module already uses. Behavior-preserving extraction.

### DIFF
--- a/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
+++ b/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
@@ -1,0 +1,6 @@
+Fixed
+
+- serialise the skill manifest's read-modify-write so concurrent deploys stop freezing skills nobody edited (closes [#4881](https://github.com/bobmatnyc/trusty-tools/issues/4881))
+  - `deploy_skills_filtered` and the unmanaged-skill adoption now run their whole load-modify-save under a new `with_skill_manifest_lock` sidecar lock, the skill-side counterpart of the agent ledger lock added in #4409
+  - `SkillManifest::save_if_current` refuses a save whose base snapshot has gone stale rather than clobbering a newer record — defence in depth for any writer that bypasses the advisory lock
+  - the `flock` critical section itself is now one implementation (`with_ledger_lock`) shared by the agent and skill ledgers
--- a/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
+++ b/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
@@ -3,4 +3,6 @@ Fixed
 - serialise the skill manifest's read-modify-write so concurrent deploys stop freezing skills nobody edited (closes [#4881](https://github.com/bobmatnyc/trusty-tools/issues/4881))
   - `deploy_skills_filtered` and the unmanaged-skill adoption now run their whole load-modify-save under a new `with_skill_manifest_lock` sidecar lock, the skill-side counterpart of the agent ledger lock added in #4409
   - `SkillManifest::save_merging` folds in any entries a writer that bypassed the lock published mid-run, instead of dropping them; it never fails or refuses after skill files are on disk, because bytes newer than their recorded checksum are exactly what the deployer reads as a hand-edit and skips forever
+  - a manifest that exists but does not parse is no longer treated as an empty one — merging from that default would publish only the current run's entries and drop the rest
+  - a mid-loop I/O failure during a deploy no longer skips the manifest save, so a skill written just before the failure stays tm-owned instead of freezing
   - the `flock` critical section is now one implementation (`with_ledger_lock`) shared by the agent and skill ledgers
--- a/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
+++ b/crates/trusty-agents-common/changelog.d/4881-manifest-lock.md
@@ -2,5 +2,5 @@ Fixed
 
 - serialise the skill manifest's read-modify-write so concurrent deploys stop freezing skills nobody edited (closes [#4881](https://github.com/bobmatnyc/trusty-tools/issues/4881))
   - `deploy_skills_filtered` and the unmanaged-skill adoption now run their whole load-modify-save under a new `with_skill_manifest_lock` sidecar lock, the skill-side counterpart of the agent ledger lock added in #4409
-  - `SkillManifest::save_if_current` refuses a save whose base snapshot has gone stale rather than clobbering a newer record — defence in depth for any writer that bypasses the advisory lock
-  - the `flock` critical section itself is now one implementation (`with_ledger_lock`) shared by the agent and skill ledgers
+  - `SkillManifest::save_merging` folds in any entries a writer that bypassed the lock published mid-run, instead of dropping them; it never fails or refuses after skill files are on disk, because bytes newer than their recorded checksum are exactly what the deployer reads as a hand-edit and skips forever
+  - the `flock` critical section is now one implementation (`with_ledger_lock`) shared by the agent and skill ledgers
--- a/crates/trusty-agents-common/src/agents/manifest.rs
+++ b/crates/trusty-agents-common/src/agents/manifest.rs
@@ -138,6 +138,49 @@ pub fn manifest_lock_path(target_dir: &Path) -> std::path::PathBuf {
     target_dir.join(format!("{MANIFEST_FILE}.lock"))
 }
 
+/// Run `f` holding the exclusive cross-process lock on the sidecar `lock_file`.
+///
+/// Why (#4881): the agent ledger (#4409) and the skill ledger need the IDENTICAL
+/// serialisation, and a `flock` critical section is the last code two copies
+/// should drift on — blocking acquisition, RAII release, and
+/// lock-failure-is-an-error ARE the safety property. One implementation, two
+/// named wrappers ([`with_agent_manifest_lock`] and
+/// [`crate::skills::manifest::with_skill_manifest_lock`]), so a change to the
+/// rule lands once.
+/// What: creates `lock_file`'s parent directory and the file itself if absent,
+/// takes a `flock(2)`-style advisory EXCLUSIVE lock on it (blocking until
+/// available), runs `f`, and releases the lock on every exit path including
+/// panics (RAII).
+///
+/// Advisory, not mandatory: a writer that bypasses this helper is not blocked,
+/// so every load-modify-save of a ledger must go through it. Blocking: callers
+/// on an async runtime must use a blocking-safe thread. Not reentrant — a nested
+/// call on the same lock file self-deadlocks.
+///
+/// Lock acquisition failure is an `Err`, never a silent unlocked fallthrough:
+/// proceeding without the lock reinstates exactly the race this closes.
+/// Test: `manifest_lock_serialises_concurrent_writers`,
+/// `crate::skills::manifest::tests::skill_manifest_lock_serialises_concurrent_writers`.
+pub fn with_ledger_lock<T, E, F>(lock_file: &Path, f: F) -> std::result::Result<T, E>
+where
+    E: From<ManifestError>,
+    F: FnOnce() -> std::result::Result<T, E>,
+{
+    if let Some(parent) = lock_file.parent() {
+        std::fs::create_dir_all(parent).map_err(|e| E::from(ManifestError::Io(e)))?;
+    }
+    let file = std::fs::OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        .truncate(false)
+        .open(lock_file)
+        .map_err(|e| E::from(ManifestError::Io(e)))?;
+    let mut lock = fd_lock::RwLock::new(file);
+    let _guard = lock.write().map_err(|e| E::from(ManifestError::Io(e)))?;
+    f()
+}
+
 /// Run `f` holding the exclusive cross-process lock on `target_dir`'s ledger.
 ///
 /// Why (#4409): the deployer's cycle is load manifest → write files → save
@@ -173,17 +216,7 @@ where
     E: From<ManifestError>,
     F: FnOnce() -> std::result::Result<T, E>,
 {
-    std::fs::create_dir_all(target_dir).map_err(|e| E::from(ManifestError::Io(e)))?;
-    let file = std::fs::OpenOptions::new()
-        .read(true)
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(manifest_lock_path(target_dir))
-        .map_err(|e| E::from(ManifestError::Io(e)))?;
-    let mut lock = fd_lock::RwLock::new(file);
-    let _guard = lock.write().map_err(|e| E::from(ManifestError::Io(e)))?;
-    f()
+    with_ledger_lock(&manifest_lock_path(target_dir), f)
 }
 
 // `repair_stale_tmp(path)` was removed with #4409. It derived a single scratch
--- a/crates/trusty-agents-common/src/skills/deployer.rs
+++ b/crates/trusty-agents-common/src/skills/deployer.rs
@@ -27,7 +27,7 @@
 
 use std::path::Path;
 
-use crate::agents::manifest::{ManifestError, Result, atomic_write, checksum};
+use crate::agents::manifest::{Result, atomic_write, checksum};
 use crate::skills::manifest::{
     SkillManifest, SkillManifestEntry, SkillManifestSave, with_skill_manifest_lock,
 };
@@ -275,19 +275,17 @@ fn deploy_skills_locked(
         }
     }
 
-    // #4881: refuse rather than clobber if the ledger moved under us. Holding
-    // the lock makes that impossible among writers that take it, so a refusal
-    // here means an UNLOCKED writer exists — a bug to surface, not to paper
-    // over. Deploy is idempotent, so the next run redoes this cycle intact.
-    match manifest.save_if_current(dest, &base)? {
-        SkillManifestSave::Written => {}
-        SkillManifestSave::RefusedStale => {
-            return Err(ManifestError::Io(std::io::Error::other(format!(
-                "the skill manifest at {} changed during this deploy — an unlocked \
-                 writer bypassed `with_skill_manifest_lock`; nothing was written",
-                dest.display()
-            ))));
-        }
+    // #4881: every file above is ALREADY on disk, so this save must publish —
+    // failing or refusing here would leave bytes newer than their recorded
+    // checksums, which is exactly what the classifier above reads as a hand-edit
+    // and skips forever. `save_merging` folds in anything an unlocked writer
+    // published mid-deploy rather than dropping it.
+    if manifest.save_merging(dest, &base)? == SkillManifestSave::Merged {
+        tracing::warn!(
+            dest = %dest.display(),
+            "the skill manifest changed during this deploy — a writer bypassed the \
+             ledger lock; its entries were merged rather than dropped"
+        );
     }
 
     Ok(stats)
--- a/crates/trusty-agents-common/src/skills/deployer.rs
+++ b/crates/trusty-agents-common/src/skills/deployer.rs
@@ -208,8 +208,56 @@ fn deploy_skills_locked(
         );
     }
 
+    // #4881: the write loop's failure sites (an unreadable source file, a failed
+    // `atomic_write`) can return AFTER earlier `SKILL.md` files are already on
+    // disk. Propagating straight out would skip the save below, leaving those
+    // files unrecorded — bytes newer than any recorded checksum, which the
+    // classifier reads as a hand-edit and skips forever. So the loop's error is
+    // held, the ledger is saved either way, and the error propagates after.
+    // Pre-existing (it predates this issue's lock work) but fixed here because
+    // it violates the same invariant.
+    let write_result = deploy_selected(source, dest, &names, &now, &mut manifest, &mut stats);
+
+    let save_result = manifest.save_merging(dest, &base);
+
+    // The write error comes first: it is the more specific failure, and the save
+    // has already run by this point regardless of which one is reported.
+    write_result?;
+    if save_result? == SkillManifestSave::Merged {
+        tracing::warn!(
+            dest = %dest.display(),
+            "the skill manifest changed during this deploy — a writer bypassed the \
+             ledger lock; its entries were merged rather than dropped"
+        );
+    }
+
+    Ok(stats)
+}
+
+/// Write every selected skill and record it, stopping at the first I/O failure.
+///
+/// Why (#4881): split out of [`deploy_skills_locked`] so its `?` sites cannot
+/// bypass the manifest save. Files this function has already written are on disk
+/// whether it returns `Ok` or `Err`, so the caller must save the ledger on BOTH
+/// paths — a shape that is easy to get wrong while the loop and the save share
+/// one function body, and was wrong here before this change.
+/// What: for each name, writes `<dest>/<stem>/SKILL.md` plus any
+/// `references/*.md` siblings through [`deploy_one_file`], recording each in
+/// `manifest` and `stats`. Skills whose stem contains `mcp` are rejected before
+/// any I/O (#2186). Returns the first I/O error; `manifest` and `stats` still
+/// describe everything written up to that point.
+/// Test: `deploy_records_files_written_before_a_mid_loop_failure`, plus every
+/// `deploy_*` test through the public entry point.
+fn deploy_selected(
+    source: &Path,
+    dest: &Path,
+    names: &[String],
+    now: &str,
+    manifest: &mut SkillManifest,
+    stats: &mut DeployStats,
+) -> Result<()> {
     for filename in names {
-        let stem = skill_stem(&filename).to_string();
+        let stem = skill_stem(filename).to_string();
 
         // Claude Code ships a built-in `/mcp` slash command. A deployed skill
         // whose invocable name (the stem, i.e. the directory Claude Code
@@ -225,20 +273,13 @@ fn deploy_skills_locked(
             continue;
         }
 
-        let source_path = source.join(&filename);
+        let source_path = source.join(filename);
         let content = std::fs::read_to_string(&source_path)?;
         // Claude Code discovers skills from <dest>/<name>/SKILL.md.
         let skill_dir = dest.join(&stem);
         let target_path = skill_dir.join("SKILL.md");
 
-        deploy_one_file(
-            &mut manifest,
-            &stem,
-            &target_path,
-            &content,
-            &now,
-            &mut stats,
-        )?;
+        deploy_one_file(manifest, &stem, &target_path, &content, now, stats)?;
 
         // Mirror any reference files a multi-file skill carries alongside its
         // SKILL.md (issue #2903, ported from `trusty-mpm` PR #2915 —
@@ -263,32 +304,12 @@ fn deploy_skills_locked(
                 let ref_content = std::fs::read_to_string(refs_source_dir.join(&ref_name))?;
                 let ref_key = format!("{stem}/references/{ref_name}");
                 let ref_target = skill_dir.join("references").join(&ref_name);
-                deploy_one_file(
-                    &mut manifest,
-                    &ref_key,
-                    &ref_target,
-                    &ref_content,
-                    &now,
-                    &mut stats,
-                )?;
+                deploy_one_file(manifest, &ref_key, &ref_target, &ref_content, now, stats)?;
             }
         }
     }
 
-    // #4881: every file above is ALREADY on disk, so this save must publish —
-    // failing or refusing here would leave bytes newer than their recorded
-    // checksums, which is exactly what the classifier above reads as a hand-edit
-    // and skips forever. `save_merging` folds in anything an unlocked writer
-    // published mid-deploy rather than dropping it.
-    if manifest.save_merging(dest, &base)? == SkillManifestSave::Merged {
-        tracing::warn!(
-            dest = %dest.display(),
-            "the skill manifest changed during this deploy — a writer bypassed the \
-             ledger lock; its entries were merged rather than dropped"
-        );
-    }
-
-    Ok(stats)
+    Ok(())
 }
 
 /// Deploy one managed file (a skill's `SKILL.md` or one of its
--- a/crates/trusty-agents-common/src/skills/deployer.rs
+++ b/crates/trusty-agents-common/src/skills/deployer.rs
@@ -27,8 +27,10 @@
 
 use std::path::Path;
 
-use crate::agents::manifest::{Result, atomic_write, checksum};
-use crate::skills::manifest::{SkillManifest, SkillManifestEntry};
+use crate::agents::manifest::{ManifestError, Result, atomic_write, checksum};
+use crate::skills::manifest::{
+    SkillManifest, SkillManifestEntry, SkillManifestSave, with_skill_manifest_lock,
+};
 
 /// Summary of one [`deploy_skills`] run.
 ///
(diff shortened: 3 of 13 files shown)


bobmatnyc:
## Code Critic — Verdict: BLOCK

Rung 5 adversarial review, reviewed at `97b8243a` in a throwaway worktree off `pr/4908`.

The lock itself is correct and I could not break it. The compare-and-swap bolted on beside it is not: its refusal path fires **after** the skill files are already on disk, and I reproduced it freezing an entire tier.

---

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-agents-common/src/skills/deployer.rs` | 282–291 | `RefusedStale` returns `Err` **after** every `SKILL.md` has been written, leaving on-disk bytes newer than the ledger — the exact freeze this PR exists to prevent, at whole-tier scale. The error text says "nothing was written"; that is false. | Never `Err` after writing. Under the lock, on a stale base: reload the on-disk manifest, merge in the keys this run actually wrote, save, return `Ok`. Or drop the CAS and go back to `manifest.save(dest)`, matching the agent ledger. | Fix here |
| MEDIUM | `crates/trusty-agents-common/src/skills/reconcile.rs` / `crates/trusty-mpm/src/core/skill_repair.rs` / `crates/trusty-mpm/src/core/update_check/apply.rs` | 153 / 242 / 284 | The CAS is applied to 1 of the 4 serialised writers. Adopt, repair and prune all call plain `save`. If the stated rationale ("catches a future writer that forgets the lock") holds, three writers are unguarded. | Apply it everywhere or drop it. Given the CRITICAL, dropping is the safer resolution. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/skill_drift.rs`, `skill_repair.rs` | 343–348, 123–125 | Three trusty-mpm behavior changes ship with zero test changes — `git diff --stat` over `skill_drift_tests.rs`, `skill_repair_tests.rs`, `update_check/apply/tests.rs` is empty. Reverting the dot-prefix filter leaves every test green, even though the sidecar is its only reason to exist. | Add `audit_reports_absent_and_empty_when_only_the_lock_sidecar_is_present` and `repair_leaves_a_missing_tier_directory_untouched`. | Fix here |
| LOW | PR body ("Tests — each proven to fail before") | — | The claimed fail-before for `skill_manifest_lock_serialises_concurrent_writers` (`left: 1, right: 8`) does not reproduce. With `flock` neutered it fails earlier, on `Io(Os { code: 2, NotFound })` — 8 unlocked `atomic_write`s colliding on the shared `temp_path()` (`agents/manifest.rs:114-125`). The test still discriminates; the quoted evidence just is not what the code produces. | Requote the actual output. | Parent |

---

## CRITICAL-1 — reproduced

Ordering in `deploy_skills_locked`: load at `:176`, `atomic_write` per file at `:234`/`:266` → `deploy_one_file:353`, save at `:282`. When `save_if_current` returns `RefusedStale`, every file is already written and the manifest is not.

Scratch integration test, 400 skills, one unlocked writer (plain `SkillManifest::save`, bypassing the lock) publishing mid-deploy:

```
ERROR TEXT: io error: the skill manifest at /var/folders/.../.tmpe1Li7F changed during this
deploy — an unlocked writer bypassed `with_skill_manifest_lock`; nothing was written
skill-0000/SKILL.md on disk = "v3 content 0\n"
next deploy: deployed=0 skipped=400 unchanged=0
test refused_stale_leaves_files_written_and_frozen ... ok
```

The next deploy — and the one after it — classifies all 400 as user-modified and skips them. That is #4881's freeze, produced by #4881's fix.

Two doc claims are falsified by that run:

- `deployer.rs:287` — "nothing was written". 400 files hold the new content.
- `manifest.rs:183-185` — "Deploy is idempotent, so the correct response to a refusal is to surface it and let the next run redo the cycle." The next run does not redo the cycle; it skips.

Versus `main` on the same path: the old `manifest.save` clobbers the racing writer's single entry but records correct checksums for all 400, so nothing freezes. On the refusal path this PR is a regression against `main` for the defect it closes.

Reachability: it needs a writer that does not take the lock. Every in-tree writer now does, but the tier is machine-global and older installed `tm` binaries do not — the rollout window has mixed-lock writers by construction, on a machine that routinely runs dozens of concurrent sessions. And if unlocked writers genuinely never happen, the CAS is dead code. Either way the design does not hold.

Downstream: `runtime/claude_code.rs:735` treats an `ensure_managed_config_dir` failure as non-fatal and `tracing::warn!`s it, so a whole-tier freeze surfaces as one warn line on the session-spawn path.

---

## What I verified and could not break

**Deadlock / re-entrancy / starvation.** Traced every locked body — `deploy_skills_locked`, `adopt_locked`, `repair_tier_locked`, `prune_skills_locked` — plus everything they reach: `atomic_write`, `deploy_one_file`, `rewrite_and_verify` (`skill_repair.rs:257-306`), `back_up`, `audit_deployed_skills`, `unmanaged_bundled_skills`. None acquires a ledger lock. No transitive nesting exists. The agent lock and the skill lock are never nested either — `apply.rs:168-170` runs `prune_agents` then `prune_skills` sequentially, and `deploy_all_skill_tiers:323/330` takes the same lock twice sequentially, released between.

**Panic and process death.** `_guard` is RAII inside `with_ledger_lock`; unwinding drops it and closes the fd. `flock` is released by the kernel on fd close, so abort/SIGKILL release too. `skill_manifest_lock_releases_so_a_second_acquisition_succeeds` covers the RAII path.

**Await points.** Every locked body is a plain sync `fn`. The async exposure (session spawn → `ensure_managed_config_dir`) already blocked on the agent ledger lock as of #4409; unchanged here.

**Lock covers the write window.** Yes, for all four writers. Deploy: `deployer.rs:155` wraps load `:176` → writes `:234`/`:266` → save `:282`. Repair: `skill_repair.rs:133` wraps audit `:172` → `rewrite_and_verify` `:210` → save `:242`. Prune: `apply.rs:258` wraps load `:268` → `remove_dir_all` `:278` → save `:285`. Adopt: `reconcile.rs:102` wraps scan `:119` → `back_up` `:134` → save `:153`.

**Divergence from the agent precedent.** None in the lock itself. `with_ledger_lock`'s `create_dir_all(lock_file.parent())` is exactly the old `create_dir_all(target_dir)`, since the lock path is `target_dir.join(...)`. The two named wrappers are behaviorally identical. The only divergence is the CAS — which is the CRITICAL.

**Test discrimination.** Neutered `flock` and the CAS in place: all three new tests fail (`0 passed; 3 failed`). Restored: 10 consecutive runs, `5 passed; 0 failed` every time. Not flaky in either direction — `deploy_blocks_while_the_skill_ledger_lock_is_held` cannot pass by accident because the deploy physically cannot finish while the lock is held, and it cannot fail by accident because the 2s probe is four orders of magnitude wider than an unlocked deploy. The 5ms snapshot hold serialises 8 threads in ~40ms; load only makes it slower, never wrong.

**Side effect 1 — the dot filter.** Safe. `is_skill_file` (`deployer.rs:66`) rejects dot-prefixed sources, so no bundled skill can ever deploy as a dot-named directory; a legitimately dot-prefixed skill cannot exist. Every other scanner that could have seen the sidecar filters to directories or `.md` and needs no change: `list_project_custom_stems` (`tiers.rs:251`), `unmanaged_bundled_skills` (`unmanaged.rs:127`), `count_skill_entries` (`doctor_fs_checks.rs:194`), `count_files_with_extension` (`:266`), `find_stale_mpm_skills` (`stale_skills.rs:241`), `count_legacy_tm_skills` (`doctor_staleness.rs:93`). It also incidentally stops `.DS_Store` alone from reading as populated.

**Side effect 2 — repair skipping a missing tier.** Behavior-preserving. A missing dir makes `audit_deployed_skills` return `AbsentAndEmpty` with zero findings (`skill_drift.rs:332-353`, `:361-366`), so `repair_tier_locked` would have done nothing anyway. No tier that should be repaired is skipped.

**Owner constraints.** `deploy_one_file`'s skip rules at `deployer.rs:325-343` are unchanged from `main` — unmanaged and checksum-mismatch both still skip. No fix-skills command added or extended (`doctor_fix_skills.rs` untouched). Nothing auto-unfrozen: `include_frozen` still gates `DriftedFrozen` at `skill_repair.rs:196`.

**Hygiene.** Both changelog fragments present with a bare category on line 1. `// #4881:` comments at every change site. No test ignored, cfg-gated, deleted or narrowed — `#[ignore]` counts are identical to `origin/main`, and the `manifest.rs` inline tests moved verbatim into `manifest_tests.rs`. `check_line_cap.sh` / `check_sld.sh` / `check_changelog_fragment.sh` → `EXIT=0`. `cargo test -p trusty-agents-common -- skills::` → `55 passed; 0 failed`. `cargo test -p trusty-mpm --lib -- skill_repair skill_drift update_check::apply` → `48 passed; 0 failed`.

---

## Required before re-review

1. Remove the write-then-`Err` path in `deploy_skills_locked` (CRITICAL-1). Merge under the lock, or drop the CAS.
2. Correct or delete the two falsified doc claims: `deployer.rs:287` "nothing was written" and `manifest.rs:183-185` "the next run redoes the cycle".
3. Make the CAS consistent across the four writers, or remove it.
4. Add the two missing trusty-mpm regression tests.

## Notes

Not attributed to this PR: `Test: doc-comment pointer lint` shows red, but its own check passed (`resolved 21524 Test: citation(s) — 0 dangling pointers — OK.`) and the job was then cancelled on timeout — the known `ci/doc-pointer-lint-timeout` condition. Everything else on the PR is green except `Test`, still pending.

Not flagged, deliberately: tier planning in `deploy_all_skill_tiers` (`list_project_custom_stems` + `plan_skill_tiers`) runs outside the lock, and the bundled and user passes are two separate critical sections. Both windows are bounded and self-correcting on the next deploy, and the PR body says so explicitly.

bobmatnyc:
## Verdict: WARN — reviewed `96b50081`

**The CRITICAL is cleared, verified by reproduction, not by assertion.**

I re-ran my 400-skill reproduction (8 concurrent locked deployers + one unlocked racer, 3 generations, then the real deployer classifies the result) against three commits:

```
741486e9 (the head I BLOCKED)  FAILED
  called `Result::unwrap()` on an `Err` value: Io(Custom { kind: Other, error:
  "the skill manifest at /var/folders/.../.tmpGaFDHz changed during this deploy —
   an unlocked writer bypassed `with_skill_manifest_lock`; nothing was written" })

96b50081 (this head)           ok
  test a_400_skills_eight_locked_writers_plus_an_unlocked_racer_freeze_nothing ... ok
  (final deploy skipped 0 of 400; ledger holds all 401 entries)
```

Fail-before / pass-after on the exact defect. The three-way merge does what the body claims: it publishes on every path through `save_merging`, folds in the unlocked writer's entries, and freezes neither side.

Answers to the specific attacks:

- **Re-read under the lock?** Yes. `save_merging` calls `SkillManifest::load` itself (`manifest.rs:212`), and every call site is inside the locked body. Not a stale snapshot.
- **Deletion semantics?** Intact. `manifest.rs:224-228` replays this run's removals onto the current document, so prune does not resurrect pruned keys and does not revert a concurrent writer's inserts. `apply_prune_removes_deselected` and `skill_manifest_save_merging_applies_this_runs_removals` both pass.
- **Dead `RefusedStale`?** Gone. `SkillManifestSave` has exactly `Written` and `Merged`, both constructed. No unreachable variant left to mislead the next reader.
- **Signal survives?** Yes — `Merged` is `#[must_use]` and all four writers log a `tracing::warn!` naming the bypassed lock.
- **Deadlock / re-entrancy / lock scope:** all four writers wrap a `*_locked` body that is never called from inside another lock; the agent and skill sidecars are distinct paths. The lock spans load → all file writes → save in every one.
- **Unchanged as required:** `deploy_one_file`'s unmanaged skip (`deployer.rs:323-328`) and checksum-frozen skip (`:337-341`) are untouched. No fix-skills command added, nothing auto-unfrozen, no test weakened, `#[ignore]`-ed, cfg-gated or deleted — the 5 pre-existing `manifest.rs` inline tests all survive the extraction into `manifest_tests.rs`, plus 6 new ones. Changelog fragments present in both crates with a bare category on line 1.

Gates I ran myself on `96b50081`:

```
cargo test -p trusty-agents-common            test result: ok. 463 passed; 0 failed; 0 ignored
cargo test -p trusty-mpm -- --test-threads=1  test result: ok. 4604 passed; 0 failed; 7 ignored
                                              test result: ok. 1438 passed; 0 failed; 1 ignored  (x2)
cargo fmt --check && cargo clippy -p trusty-agents-common -p trusty-mpm --all-targets -- -D warnings   EXIT=0
bash scripts/check_line_cap.sh                0 violations — OK
```

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-agents-common/src/skills/deployer.rs` | 229, 253, 263 (and `deploy_one_file` 329, 349, 351) | **PRE-EXISTING, not introduced here.** A mid-loop `?` still returns `Err` after earlier `SKILL.md` files are on disk, so `save_merging` at `:283` is never reached and those files freeze permanently. Reproduced: an unreadable source file mid-loop leaves `aaa-skill` written but unrecorded, and the next deploy reports `skipped: ["aaa-skill"]`. Identical result on merge-base `f49e66be`, so this PR neither introduces nor worsens it. | Make the save unconditional: run the write loop into a local `Result`, save the manifest, then propagate — `let outcome = (\|\| { ...loop... })(); let saved = manifest.save_merging(dest, &base); outcome?; saved?;` Roughly six lines, in a function this PR already restructured. | Fix here |
| MEDIUM | `crates/trusty-agents-common/src/skills/manifest.rs` | 212 | `Self::load` maps an unparseable or unreadable on-disk ledger to the empty default, so the merge starts from **empty** and publishes only this run's delta. Measured: with the ledger corrupted between load and save, 1 of 10 entries survived and the next deploy reported `skipped: ["s1"..."s9"]`. Main's plain `save` would have written all 10. New behaviour; low reachability today (under the lock only external corruption triggers it), but it goes live the moment `SKILL_MANIFEST_VERSION` changes and an older binary cannot parse the newer document. | Distinguish "loaded empty" from "failed to load": add a fallible loader and, when the current document cannot be parsed, fall back to `self.save(target_dir)` rather than merging onto an empty base. | Fix here |
| MEDIUM | `crates/trusty-agents-common/src/skills/manifest.rs` | 214-219 | Shared-key merge is last-writer-wins **on the ledger**, which can disagree with last-writer-wins **on the file**. When an unlocked writer publishes different content for a key this run also touched, the merge records this run's checksum over the racing writer's newer one and that file freezes. Reproduced on `96b50081` — and identically on merge-base `f49e66be`, so it is a residual of unlocked writers, not a regression. | Either state the per-key policy explicitly in the `save_merging` doc (it is currently silent on the shared-key case), or re-read the deployed file for any key where `on_disk`, `base` and `self` all differ and record the checksum of what is actually on disk. | Parent |
| LOW | `crates/trusty-agents-common/src/skills/manifest_tests.rs` | 10 | Module doc still says the file covers "the stale-snapshot refusal". There is no refusal any more — `save_merging` always publishes. A stale claim in code this PR adds. | Replace with "the merging save's fold-in, removal replay, and unraced write". | Fix here |

## Required Changes

1. The HIGH is the one that matters and it is not a regression, so it does not gate this PR. Land it here if you want the invariant the module docs now assert to hold end-to-end; otherwise it needs its own issue, because `main` carries it either way.

## Notes

Why WARN and not BLOCK: the surviving error path reproduces byte-for-byte on the merge-base. Blocking a change that provably removes one freeze mechanism, because a second and older one survives beside it, would leave `main` strictly worse. The CRITICAL I raised is closed on the evidence above.

The corrected fail-before account in the PR body matches what I measured independently — 463 / 4604 / 1438 / 1438, all passing.

bobmatnyc:
## Re-review round 2 — all three folded in

**(1) MEDIUM — unparseable ledger treated as empty.** You are right, and it is the same fail-open shape wearing a different hat. `SkillManifest::load` maps both "absent" and "unparseable" to the empty default, so `save_merging` started from empty and published only this run's delta.

I chose **publish over it, loudly** rather than fail. Failing is not available here: files are already on disk by the time this runs, which is the invariant this whole PR is about. Publishing `self` (base + delta) is exactly what a plain `save` does, so the path is never worse than main, and the entries read at load time survive. A private `read_parsed` makes the distinction — `Ok(Some(default))` absent, `Ok(Some(parsed))` readable, `Ok(None)` corrupt — and the corrupt branch logs `tracing::error!` and returns a distinct `OverwroteUnreadable`. Kept private rather than a second public load API, since only `save_merging` needs it.

**(2) HIGH, pre-existing — folded in, and I agree with the reasoning.** The write loop moved to `deploy_selected`; the caller now saves on both paths and propagates the write error afterwards. Ordering is write-error-first because it is the more specific failure and the save has already run either way.

**(3) LOW** — comment fixed with the code.

### Both proven to fail on revert

```
# (1) reverted to `Self::load`:
skill_manifest_save_merging_over_a_corrupt_ledger_keeps_the_base ... FAILED
  left: Merged   right: OverwroteUnreadable

# (2) reverted to propagating before the save:
deploy_records_files_written_before_a_mid_loop_failure ... FAILED
  a file written before the failure must still be recorded
```

The (2) fixture uses invalid UTF-8 in a source file rather than a chmod, so it behaves identically if the suite runs as root. My first attempt used an unreadable directory — the collection loop filters non-files, so it never reached the read. Caught and fixed.

### On the shared-key divergence you told me to leave

Leaving it. It is cheap to close mechanically — the merge could compare `deployed_at` and keep the newer entry — but not cheap to close *correctly*: both writers wrote real bytes to the same path, so whichever entry loses names a checksum that does not match what is on disk, and that file freezes. Picking a winner by timestamp just moves which one freezes. The real fix is preventing two writers from targeting one key, which is a different change. Worth its own issue rather than a hunk in this one.

### Gates — rung 5, rebased onto current main

| Gate | Result |
|---|---|
| `cargo fmt --check` + clippy on both crates (`-D warnings`) | EXIT=0 |
| `cargo test -p trusty-agents-common -- --test-threads=1` | 465 passed, 0 failed |
| `cargo test -p trusty-mpm -- --test-threads=1` | 4604 + 1438 + 1438 passed, 0 failed |
| `cargo test -p trusty-agents --lib -- --test-threads=1` | 3341 passed, 0 failed |
| `cargo test -p trusty-code -- --test-threads=1` | 1600 + 21 passed, 0 failed |
| `check_line_cap.sh` / `check_sld.sh` / `check_changelog_fragment.sh` | EXIT=0 |
| `cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui` | EXIT=0 |

All three direct dependents of `trusty-agents-common` covered. GUI exclusions are the documented Tauri `frontendDist` baseline; my diff touches 0 files there.

Head is `cbc63a37`. Expecting one more review round since (2) touches the write path.

bobmatnyc:
## Verdict: APPROVE — round 3, reviewed `cbc63a37667fee966bd39b708ee922b39f05952b`

Round 1 BLOCK and round 2 WARN are both cleared. Verified independently in a detached worktree at the PR head, not from the engineer's assertions.

### 400-skill repro re-run — passes

Rebuilt the round-2 repro against `cbc63a37`: 400 skills, 8 concurrent deployers on one destination, two content generations, plus an unlocked racing writer and a ledger corrupted mid-flight.

```
running 3 tests
test concurrent_deployers_freeze_nothing_at_400_skills ... ok
test corrupt_ledger_midflight_keeps_all_400 ... ok
test unlocked_racing_writer_freezes_neither_side_at_400_skills ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.16s
```

Nothing froze: every writer's `skipped` list is empty in both generations, the ledger holds all 400 entries, and the unlocked writer's 10 entries survive. The round-2 pass held — the mid-loop refactor did not reopen the racing-writer freeze.

### The invariant, re-enumerated after the loop moved

Every exit that can occur with skill files already on disk publishes.

`deploy_skills_locked` (`skills/deployer.rs:169`): the three `?` sites before the write loop — `read_dir` 184, `entry` 185, `file_type` 190 — all precede any write. Past that there is no `?`. `deploy_selected`'s error is bound at 219, `save_merging` runs unconditionally at 221, and only then does 225 propagate.

`deploy_selected` (`skills/deployer.rs:251`) has five failure sites — 277, 282, 294, 304, 307 — and all five return to that caller, which saves. Inside `deploy_one_file` the failure sites (350, 371, 372) sit *before* the `manifest.managed.insert` at 373.

I found no exit that returns without publishing.

### `OverwroteUnreadable` is correct, not merely non-failing

"Base" is what was actually readable. Reverting `read_parsed` to `load` and re-running proves the fix is load-bearing rather than cosmetic:

```
CRITIC: save_merging outcome over corrupt ledger = Merged
assertion `left == right` failed: corrupt-ledger save published only 1 of 401 entries
```

and the PR's own test fails the same way under that revert:

```
test skills::manifest::tests::skill_manifest_save_merging_over_a_corrupt_ledger_keeps_the_base ... FAILED
  left: Merged
 right: OverwroteUnreadable
```

Restored, both pass and 401 of 401 entries publish.

**Valid-JSON-wrong-shape takes the corrupt branch**, which is the stated schema-bump reachability case. Probed five shapes directly:

```
{"version":2,"entries":{}}     -> OverwroteUnreadable, 10 entries published
{"version":2}                  -> OverwroteUnreadable, 10 entries published
[]                             -> OverwroteUnreadable, 10 entries published
{"version":"two","managed":{}} -> OverwroteUnreadable, 10 entries published
null                           -> OverwroteUnreadable, 10 entries published
```

A renamed or dropped `managed` field fails `serde_json::from_str::<SkillManifest>`, so `read_parsed` returns `None` and the base survives. It does not fall into the empty branch.

### The mid-loop fix records exactly what was written

The mirror defect — recording a file that was *not* written — is absent. I forced `create_dir_all` to fail for the middle skill of three (`<dst>/bbb-skill` planted as a file, so the parent create returns ENOTDIR):

```
test a_file_that_failed_to_write_is_never_recorded ... ok
```

`aaa-skill` is recorded; `bbb-skill` (attempted, failed) and `ccc-skill` (never reached) are not; and after unblocking, a redeploy skips nothing. The insert at `deployer.rs:373` sitting after `atomic_write` is what makes this hold.

Reverting fix 2 to the pre-fix shape (`deploy_selected(...)?` ahead of the save) fails the PR's regression test:

```
test skills::deployer::tests::deploy_records_files_written_before_a_mid_loop_failure ... FAILED
  a file written before the failure must still be recorded
```

Both fail-before claims verified by reversion.

### The replacement fixture

`zzz-skill.md` holding `[0xff, 0xfe, 0xff]` is a real `.md` file, so `is_skill_file` accepts it and the collection loop keeps it; `read_to_string` then fails with InvalidData at `deployer.rs:277` — a mid-loop site, after `aaa-skill` is on disk. `aaa` sorts before `zzz` under the `sort_unstable` at line 199, so the ordering is deterministic. No permission bits are involved, so it behaves identically as root.

### Unchanged, as required

- `deploy_one_file` is **byte-identical to main** (`diff` of the extracted function body, exit 0). Unmanaged and checksum-frozen skips untouched.
- No CLI surface change: `git diff --stat -- crates/trusty-mpm/src/bin/tm/` is empty. No fix-skills command added.
- Nothing auto-unfrozen: `skill_repair.rs:200` still gates `SkillDrift::DriftedFrozen` behind `include_frozen`.
- No nesting across the four locked bodies. `deploy_skills_locked`, `adopt_locked`, `prune_skills_locked` and `repair_tier_locked` are the only ones; grep over every `with_skill_manifest_lock` / `with_agent_manifest_lock` / `with_ledger_lock` call site shows no path entering one from inside another. In `install_skills.rs` adoption (167) and deploy (179) are sequential; in `apply.rs` deploy (158) and prune (260) likewise.
- The lock spans load → writes → save, proved by `deploy_blocks_while_the_skill_ledger_lock_is_held`, which holds the sidecar lock and asserts no skill file lands for 2s.
- Changelog fragments in both crates, category alone on line 1.
- No test weakened. The five removed `manifest.rs` inline tests reappear verbatim in `manifest_tests.rs`; no `#[ignore]`, no cfg-gate, no `--exclude` anywhere in the diff.
- `repair_skills`'s new pre-lock `!tier.dir.is_dir()` skip is behaviour-preserving: `audit_deployed_skills` returns `AbsentAndEmpty` with an empty `findings` vec whenever the ledger is absent (`skill_drift.rs:361`), and findings derive from manifest keys, never from the reference set — so a missing tier produced zero outcomes on main too.

### Gates

```
cargo fmt --check && cargo clippy -p trusty-agents-common --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-agents-common      test result: ok. 465 passed; 0 failed; 0 ignored
                                        test result: ok. 4604 passed; 0 failed; 7 ignored
                                        test result: ok. 1438 passed; 0 failed; 1 ignored
                                        test result: ok. 1438 passed; 0 failed; 1 ignored
                                        (+17 further binaries, all 0 failed)   EXIT=0
bash scripts/check_line_cap.sh          3689 files, 0 violations — OK
```

## Findings

|---|---|---|---|---|---|
| LOW | `crates/trusty-agents-common/src/skills/manifest.rs` | 213 | The `error!` is not loud on the paths that reach it most. `tm` registers a tracing subscriber only for daemon/supervisor/MCP modes and for `tm sessions instructions` (`bin/tm/tracing_setup.rs:37`, `main.rs:268`). A bare `tm` launch runs the deploy in-process (`commands/launch.rs:196`), as do `tm doctor --fix-skills` and `tm catalog apply` — none register one, so this `error!` and the three `warn!` merge lines at `deployer.rs:227`, `reconcile.rs:160`, `apply.rs:295` go nowhere there. | Not this PR's job — [#4878](https://github.com/bobmatnyc/trusty-tools/issues/4878) owns it. Either soften the doc wording or leave it and let #4878 land. | Parent |
| LOW | `crates/trusty-agents-common/src/skills/deployer.rs` | 225 | When the write loop AND the save both fail, `write_result?` returns first and `save_result` is dropped un-inspected, so the save failure is never surfaced. The ordering is deliberate and the more specific error does win; the second error is simply lost. | Log `save_result`'s error before `write_result?` when it is `Err`. | Parent |

## Notes

Two shapes I probed and am deliberately **not** filing:

- **Ledger deleted between load and save** publishes only this run's delta (measured: 1 of 10). Not a distinct defect — deleting the ledger freezes every deployed skill anyway, race or no race, because `deploy_one_file` classifies an existing unmanaged file as user-owned and skips it. The race adds nothing.
- **Valid JSON that parses to a smaller `managed` map** likewise publishes the delta only. That is the merge's intended semantics ("on-disk wins for keys I did not touch") and is what makes the concurrent-prune case correct. Changing it would break `skill_manifest_save_merging_applies_this_runs_removals`.

The shared-key last-writer-wins divergence was out of scope by agreement; I developed nothing there.

CI at review time: 24 checks pass, `Test` still pending. Whether to merge on that is the PM's call, not mine.

Issue #4897, opened 2026-08-05
two doctor daemon-health tests fail when a live trusty-memory daemon is listening


bobmatnyc:
`commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener` and `commands::doctor::tests::check_daemon_health_fails_when_no_addr_file_and_no_listener` in `crates/trusty-memory` both fail on a developer machine that has a live trusty-memory daemon listening on ports 7070–7079.

Both tests assume no listener is present. They are environment-dependent, not deterministic.

Verified pre-existing: they fail identically on clean `origin/main` with no PR changes applied (confirmed during PR #4895 review by stashing the branch and re-running).

They are NOT currently listed in `docs/reference/test-ladder-baseline.md`, so the baseline registry does not know about them. Either add them there or make them hermetic — pick a real fix, do not just document the flake.

Impact: any contributor running the local quality gate on a machine with the daemon running sees two red tests unrelated to their change, which trains people to ignore red.

Label this as a test-hygiene issue.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

bobmatnyc:
Reproduces on `origin/main` (`1b899686a`) today, with the cause this issue already names.

```
# throwaway worktree at origin/main
$ cargo test -p trusty-memory --lib check_daemon_health
panicked at crates/trusty-memory/src/commands/doctor/mod.rs:590:9:
unexpected status Warn; expected Fail (stale addr, no listener), Pass (fallback found a live #4001-aware daemon), or Unknown (fallback found a live pre-#4001 daemon)
panicked at crates/trusty-memory/src/commands/doctor/mod.rs:639:9:
unexpected status: Warn
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 778 filtered out; finished in 0.02s
```

A live daemon is listening on this machine, which is the condition in the title:

```
$ lsof -nP -iTCP -sTCP:LISTEN | grep trusty-me
trusty-me 77858 masa 11u IPv4 TCP 127.0.0.1:7070 (LISTEN)
```

[...]

Issue #4892, opened 2026-08-05
epic: implement ADR-0028 — three-tier memory recall (Standing/Current/Episodic)


bobmatnyc:
## Reference
ADR-0028: "Memory recall splits into three tiers — Standing, Current, Episodic — with enforced retirement for Current" (docs/adr/0028-memory-recall-tiers-standing-current-episodic.md, ratified in design-memory-recall-gap branch)

## Scope
Crates: trusty-common (memory_core), trusty-memory

## Implementation Tasks

- [ ] #4884 T1: feat(trusty-common): add Drawer.fact_key and the DRAWERS_BY_FACT_KEY secondary index
- [ ] #4885 T2: feat(trusty-common): enforce expires_at at read time, not only at palace open
- [ ] #4886 T3: feat: Tier C write path — fact_key, TTL, and fail-closed mandatory retirement
- [ ] #4887 T4: feat(trusty-memory): D6 supersession — demote, never delete
- [ ] #4888 T5: feat(trusty-memory): enforce Tier S hard cap of 20 facts and the 80-char form constraint
- [ ] #4889 T6: feat(trusty-memory): restructure the injection budget into four independently-capped sections
- [ ] #4890 T7: feat(trusty-memory): quarterly re-affirmation for Tier S facts + doctor reporting
- [ ] #4891 T8: feat(trusty-memory): read-only backfill classification report

## Sequencing

T1 + T2 + T5 run in parallel (no cross-dependencies). T3 depends on T1. T4 and T6 depend on T3 and can run in parallel with each other. T7 depends on T5. T8 depends on T2.

## Scope Decisions (Recorded in Epic)

- `live_while` (e.g. `gh:pr-open:4818`) is OUT OF SCOPE for this wave. The ADR's Risks table calls it optional and advisory; no GitHub-state checker exists anywhere in the workspace today. Retirement in this wave is `expires_at` + default TTL only.

- D6's supersession pointer is implemented as a KG triple (`drawer:<old> --superseded-by--> drawer:<new>`) via the existing kg.assert path, NOT as a new Drawer field. Rationale: the ADR's own reversibility accounting budgets exactly one new field (`fact_key`) and one new index; a second schema field would exceed the stated scope.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

Issue #4887, opened 2026-08-05
feat(trusty-memory): D6 supersession — demote, never delete


bobmatnyc:
Implement the supersession pointer as a KG triple `drawer:<old> --superseded-by--> drawer:<new>` via the existing kg.assert path (decision recorded in the epic; the ADR leaves the storage location unspecified).

A superseded Tier C fact is demoted, not deleted.

Depends on T3.

Part of epic: implement ADR-0028 — three-tier memory recall (Standing/Current/Episodic)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closed as not planned — owner review of the wont-do candidates, 2026-09-12 (session trusty-tools-c4). Reopen if the need returns.

Issue #4885, opened 2026-08-05
feat(trusty-common): enforce expires_at at read time, not only at palace open


bobmatnyc:
**IMPORTANT CORRECTION TO THE ADR:** ADR-0028 §C8 claims `expires_at` "is enforced nowhere." That is wrong. `PalaceHandle::open_with_intent` already prunes expired drawers inline at every palace open (crates/trusty-common/src/memory_core/retrieval/handle.rs:325-347, issue #61), and a standalone `PalaceHandle::purge_expired` exists (handle.rs:874-917) with no non-test call site.

The REAL gap: expiry is consulted only at palace open, never in the recall path — `retrieve_l0_l1`/`retrieve_l2`/`rescore_l1_by_similarity` in retrieval/layers.rs never filter on it. For the long-lived HTTP daemon (OpenIntent::Writer, palace held for process lifetime), a drawer that expires mid-session keeps being served until daemon restart.

Fix: add a read-time expiry filter to the retrieval path, refactoring the open-time sweep logic into a shared predicate both call rather than duplicating it.

No dependencies. Runs parallel with T1/T5.

Part of epic: implement ADR-0028 — three-tier memory recall (Standing/Current/Episodic)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #4909, opened 2026-08-05
feat(memory): ADR-0028 Tier C write path — fact_key, default TTL, atomic retire-on-write

### LINKED ISSUE #4886: feat: Tier C write path — fact_key, TTL, and fail-closed mandatory retirement
Thread `fact_key` and `expires_at` through the whole write chain: `WriteDrawerParams` (crates/trusty-memory/src/tools/helpers.rs:398-406) → `RememberOptions` (crates/trusty-common/src/memory_core/retrieval/types.rs:67-92) → `PalaceHandle::remember_with_options` (crates/trusty-common/src/memory_core/retrieval/handle.rs:511-518). None of these carry expiry or fact_key today.

Implement atomic retire-on-write using T1's index (writing a fact_key that already has a live occupant retires the incumbent).

Implement the 24h default TTL and the fail-closed degrade-to-Tier-E when a retirement condition is missing.

Update the MCP tool schemas (crates/trusty-memory/src/tools/definitions.rs) and handlers `memory_remember` (tools/memory_ops.rs:32-161) and `memory_note` (tools/memory_ops.rs:163+).

Depends on T1.

Part of epic: implement ADR-0028 — three-tier memory recall (Standing/Current/Episodic)

bobmatnyc:
Closes #4886. Refs #4884, #4885, #4892.

## The defect

A point-in-time fact fails by the opposite mechanism from a standing rule (ADR-0028 §C7): it keeps winning retrieval long after it stopped being true, because nothing retires it. The estate's most-injected drawer is a 19-day-old session checkpoint reaching 44.8% of turns, pinning an `origin/main` SHA that has been wrong for four weeks. #4884 built the slot index and #4885 made expiry enforced at read time; neither could be reached, because no caller could set either field.

## What changed

**Threading.** `RememberOptions` carries `fact_key` and `expires_at`. `PalaceHandle::remember_with_options` is the single admission point — it is the only surface that can name a slot, so the gate is unbypassable rather than conventional. `WriteDrawerParams` reaches it through its existing `opts` field rather than duplicating the two values; duplicated state is a drift risk and the struct already carried `RememberOptions`.

**Fail-closed admission (D4).** A key that breaks the `<domain>:<id>/<aspect>` grammar, or an `expires_at` already elapsed, is refused Tier C and the drawer is written exactly as the pre-ADR code would have written it. Never admitted-and-warned: a privileged fact that goes stale is worse than an ordinary stale drawer, and D4's safety property is that the worst case degrades to today's behaviour, never below it. A slot with no explicit expiry takes the 24-hour default, so an admitted Tier C fact always declares how it ends. `live_while` is out of scope per the ADR's own Implementation-scope section.

**Atomic retire-on-write (D5).** Atomic in two senses. The read-decide-write sequence runs inside the per-palace write mutex (#154), so no concurrent writer on the palace interleaves. Both drawer rows then commit in ONE redb transaction (`KnowledgeGraph::upsert_drawers_atomic` → `apply_batch`), so a crash cannot land between retiring the incumbent and landing the replacement. The mutex orders writers; the transaction makes each write indivisible.

**MCP surface.** `memory_remember` and `memory_note` take both arguments and report the tier the write actually landed in — `tier: "C"|"E"`, plus `tier_c_refused` naming the reason. Fail-closed that nobody can observe teaches a writer nothing; they keep writing keys that never take effect. A malformed `expires_at` string is a hard error (a caller typo, where substituting the default would hide it); semantic refusals degrade.

## Two decisions the #4899 review left open

**The displaced drawer's own field.** #4899 moved the index entry but deliberately left the displaced drawer's `DrawerRecord.fact_key` reading the old slot name — correct while nothing read it as a liveness signal. With a write path it is wrong: `load_drawers()` would show two drawers claiming `pr:4818/state` while only one is indexed, and any consumer trusting the field over the index reads the retired fact as live. The retirement clears it. `expires_at` is cleared with it, because on a Tier C drawer that field IS the retirement condition D4 demanded and supersession has discharged it — leaving it set would make the demoted record self-destruct at the next sweep, contradicting D6 and orphaning the pointer #4887 will hang off it. What remains is an ordinary Tier E drawer: permanent, ranked, readable.

**The sweeps.** `open_with_intent`'s inline prune and `purge_expired` now skip drawers holding a slot (`Drawer::is_tier_c`). This is scope I added deliberately and it is the one judgment call worth a reviewer's attention: without it, the 24-hour default TTL becomes a 24-hour self-destruct, because expiry on an ordinary drawer means hard delete. Read-time expiry (#4885) already stops an expired current fact being served — that is the demotion D6 asks for — and D6 rejects hard deletion on evidence, not principle: this estate hand-wrote 109 amendment edges precisely so corrections survive. A superseded drawer has `fact_key = None`, so it is not exempt and the sweeps treat it exactly as before.

## Red-first proof

**Retirement invariant.** With `persist_with_retirement` short-circuited to a plain upsert — leaving only #4884's index maintenance, which is what the storage layer alone gives you:

```
test tier_c_retirement_clears_the_displaced_drawers_own_fact_key ... FAILED
  assertion failed: a displaced drawer must stop claiming the slot on its own row
   left: Some("pr:4818/state")
  right: None

test concurrent_tier_c_writes_to_one_slot_leave_exactly_one_claimant ... FAILED
  assertion failed: exactly one drawer may claim the slot; got 16 — [118179f0…, 11e5b947…, …]
   left: 16
  right: 1

test result: FAILED. 7 passed; 2 failed
```

Note `tier_c_write_retires_the_prior_slot_occupant` still PASSED there — the index-only version looks correct until you read the rows. That is exactly the review note this PR was asked to resolve.

**Concurrency guarantee.** With the retirement restored but the per-palace write guard released before the persist step (`drop(_write_guard)` + a yield), leaving the read-decide-write sequence unserialised — the PR #4895 shape:

```
test concurrent_tier_c_writes_to_one_slot_leave_exactly_one_claimant ... FAILED
  assertion failed: exactly one drawer may claim the slot; got 3 — [07e2c239…, 1561b5e9…, 86057fea…]
   left: 3
  right: 1

```

Both experiments were reverted and every gate below re-run. A sequential test passes against both broken variants, which is why the concurrent one exists.

## Write-path audit (untruncated)

The question is "who can write a drawer", not "what did I change", so the audit is a full-repo grep with no `head`/`-m` bound. Every path reaching storage — the MCP tools (`memory_remember`, `memory_note`, `task_add`), the HTTP handlers (`service/core.rs`), the health-probe seeder (`web/health.rs`), `memory_send_message` (`messaging/operations.rs`), the chat tool surface (`chat/tools.rs:509` — the surface PR #4895's truncated audit missed), the trusty-mpm SM and daemon memory writers, the dream cycle (`dream/cycle.rs:385`), and the trusty-mpm importer (which POSTs to the daemon) — funnels through `remember_with_options`, so all of them get the admission gate. The paths that build a `Drawer` by hand and call `kg.upsert_drawer` directly (`commands/kuzu_migrate.rs`, `memory_core/git.rs`, `semantic_consolidation/mod.rs`) never set `fact_key` and cannot claim a slot at all; a repo-wide grep for `fact_key` outside the storage layer and tests confirms `tier_c::apply_admission` is the only production assignment.

## Quality gate

```
cargo test -p trusty-common --features memory-core,embedder-test-support
  test result: ok. 836 passed; 0 failed; 13 ignored
  test result: ok. 11 passed; 0 failed; 0 ignored

cargo test -p trusty-memory
  test result: FAILED. 503 passed; 2 failed; 4 ignored

cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings   EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings                          EXIT=0
cargo fmt --check                                                                   EXIT=0
  line-cap: measured 3690 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
cargo check --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui   EXIT=0
```

The two `trusty-memory` failures are `commands::doctor::tests::check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener` and `…_fails_when_no_addr_file_and_no_listener` — pre-existing, tracked as #4897, and their precondition is confirmed present on this host: `lsof` shows `trusty-me 72520 … 127.0.0.1:7070 (LISTEN)`. Both pass in CI, which has no live daemon.

`handle.rs` was at 489/500 SLOC before this change, so the Tier C logic went into a new `retrieval/tier_c.rs` rather than into it; the file lands at 491 and nothing was allowlisted.


bobmatnyc:
This session (`tm-trusty-tools-03`) pushed `959859c6` to this branch — a `main` → `feat/adr-0028-tier-c` merge commit, not a force-push, no history rewritten.

**Why:** the branch had a real content conflict against `main`, not just a stale-mergeability cache. `main` (via the merged Tier S work) and this branch had both appended new test blocks at the same anchor point in `crates/trusty-memory/src/tools/tests.rs`. Resolved by keeping BOTH blocks in full — no function-name or logic overlap between them.

**Verified before pushing:**
- `cargo check -p trusty-memory -p trusty-common --tests` — clean
- `cargo test -p trusty-memory --lib tools::tests` — 76 passed, 0 failed
- `cargo test -p trusty-common --features memory-core --lib retrieval::` — 64 passed, 0 failed, 3 pre-existing ignored, including `expired_tier_c_drawer_survives_the_open_time_sweep`

This session is standing down and leaving the branch to whoever's driving `fix-4909` — didn't want a merge commit landing without an explanation attached to it.

Status otherwise unchanged: code-critic APPROVE stands, all four of the engineer's flagged judgment calls adjudicated and accepted. What's left here is CI-green-and-merge, not re-review.

PR #4914, opened 2026-08-05
feat(trusty-mpm): tm register takes the URL first with an optional alias (#4912)

### LINKED ISSUE #4912: tm register should take the URL first and default the alias to owner/repo
## Current behaviour

`tm register <ALIAS> <URL>` — both positionals required, alias first. Verified from installed `tm 1.3.5 --help`.

## Wanted

- `tm register <URL> [ALIAS]` — URL first, alias optional and second.
- With no alias, defaults to `owner-repo` derived from the URL. `tm register https://github.com/owner/repo` registers alias `owner-repo`.

## Breaking change hazard

Swapping positional order is a **silent breaking change**, not a loud one: both positions take strings, so existing `tm register my-alias https://…` invocations would be re-interpreted rather than rejected — registering an alias named after the URL and a URL that is not one. Muscle memory and scripted calls break quietly.

**Mitigation:** a URL is mechanically detectable (contains `://`, starts `git@`, or matches `host/owner/repo` shape). The parser should accept BOTH orders by detecting which positional is the URL, so the legacy form keeps working. Reason: swapping order is a silent reinterpretation, not a loud error. Reserve a hard break for a deliberate SemVer major if the owner chooses to enforce strict positional order later.

## Alias derivation

Must handle real shapes:
- `https://github.com/owner/repo`
- Same with `.git` suffix
- Trailing slash
- `git@github.com:owner/repo.git`
- Non-GitHub hosts

Hyphen-separated format (`owner-repo`) was chosen deliberately to avoid path-construction hazards — no sanitization needed, no ambiguity with filesystem nesting.

For a URL with no discernible owner (bare path): error, or fall back to repo name alone — specify which.

## Collision behaviour

If the derived alias is already registered to a **different URL**, refuse loudly rather than silently rebinding — silent rebinding is a known failure mode in this repo.

**Alias ambiguity (name, don't solve here):** the hyphenated form introduces a collision risk: `github.com/a/b-c` and `github.com/a-b/c` both flatten to `a-b-c`. This is a real ambiguity, not a defect to solve in this issue — the implementer should decide on collision detection or first-one-wins semantics rather than discovering it later.

bobmatnyc:
Closes #4912

`tm register <ALIAS> <URL>` becomes `tm register <URL> [ALIAS]`. With no alias, one is derived from the URL as hyphen-joined `owner-repo`.

## Both positional orders are accepted

Whichever positional is URL-shaped (`://`, leading `git@`, or containing a `/`) is taken as the URL, so `tm register <alias> <url>` keeps working. The two shapes cannot overlap: a valid alias matches `^[a-z0-9][a-z0-9._-]*$`, which admits no `/`, `:`, or `@`.

**Correction to the issue's framing.** The issue says swapping the order would be a *silent* reinterpretation. Measured against installed `tm 1.3.5`, the two-argument swap is actually loud — `validate_alias` rejects a URL as an alias:

```
$ tm register https://github.com/owner/repo my-alias --root $TMP
Error: failed to register alias 'https://github.com/owner/repo'
Caused by: invalid alias 'https://github.com/owner/repo': must match ^[a-z0-9][a-z0-9._-]*$
```

The genuinely silent case is the *one*-argument one a strict URL-first parser would introduce: `tm register my-alias` would treat `my-alias` as a URL, derive `my-alias` from it, and register an alias pointing at a bogus URL with no error. `looks_like_url` catches that (`one_arg_non_url_errors`). Order detection is still worth keeping — it makes the legacy two-argument form work rather than fail — but the compatibility argument is "don't break scripts", not "don't corrupt the registry".

## Derivation

Delegates to `trusty_common::palace_id::owner_repo_from_git_remote`, the workspace's existing git-URL parser (common-entry-point rule), instead of adding a second one. It already handles `.git`, trailing slashes, scp-syntax SSH, non-GitHub hosts, and explicit ports.

🔴 The separator is a **hyphen**, not a slash, and the code says so at the change site: an alias becomes a path segment wherever it is consumed, so a literal `/` would nest into the filesystem.

## Decisions this PR makes

**No discernible owner → fall back to the bare repo slug.** `https://example.com/repo` → `repo`. Erroring would reject valid self-hosted single-segment paths, and the shared parser already has this fallback. Only a URL with *nothing* extractable (host-only, e.g. `https://example.com/`) errors, and the message carries the escape hatch:

```
Error: cannot derive an alias from 'https://example.com/' — no owner/repo path found.
Pass one explicitly: tm register https://example.com/ <alias>
```

**Hyphen-flattening ambiguity (`a/b-c` vs `a-b/c` → `a-b-c`) → accept, no special case.** Both derive `a-b-c`; the second registration is then caught by the ordinary different-URL collision refusal, which already exists and already refuses. Adding ambiguity-specific detection would buy a different error string for the same outcome. The message names the derivation so the cause is legible:

```
Error: failed to register alias 'a-b-c' (derived from 'https://github.com/a-b/c') —
       pass an explicit alias to register it under a different name
Caused by: alias 'a-b-c' is already registered as 'https://github.com/a/b-c'; use --force to overwrite
```

`add` errors before mutating and `save()` is never reached, so a refused registration leaves `registry.json` byte-identical — pinned by `collision_refuses_and_leaves_registry_untouched`.

**Known wart, pre-existing and unchanged:** URL equality is string equality, so `…/repo` and `…/repo.git` collide as "different URLs" for the same repo. That was already true when you typed the alias yourself; #4912 does not change it. Normalizing URLs would touch clone/load semantics and belongs in its own change.

## Test ladder — rung 6 (CLI surface change)

Fail-before, measured against installed `tm 1.3.5` (predates this branch):

```
$ tm register https://github.com/owner/repo --root $TMP
error: the following required arguments were not provided:
  <URL>
```

Pass-after, 15 new tests in `crates/trusty-mpm/src/bin/tm/commands/register_args_tests.rs`:

```
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 1439 filtered out
```

Gates:

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 4602 passed, 0 failed, 7 ignored (largest binary; 45 result lines, all ok) |
| `cargo check --workspace` | EXIT=101 — **pre-existing baseline**, not this diff |

The workspace failure is the documented Tauri baseline, in two crates this diff does not touch:

```
error: proc macro panicked
```

`cargo check --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui` → EXIT=0. The `--exclude` is diagnostic isolation of a known red, not a gate being narrowed to pass.

## Smoke run (rung 6), against a throwaway `--root`

```
$ tm register --help
Usage: tm register [OPTIONS] <URL> [ALIAS]

$ tm register https://github.com/bobmatnyc/trusty-tools --root $TMP
registered bobmatnyc-trusty-tools → https://github.com/bobmatnyc/trusty-tools

$ tm register https://github.com/acme/widget.git shiny --root $TMP
registered shiny → https://github.com/acme/widget.git

$ tm register legacy-proj git@gitlab.com:team/thing.git --root $TMP   # legacy order
registered legacy-proj → git@gitlab.com:team/thing.git

$ tm register https://git.example.com/loneRepo --root $TMP            # no owner
registered lonerepo → https://git.example.com/loneRepo
```

The default `~/.trusty-mpm` registry was never written to.

## Stale help and docs updated

- `tm register --help` (clap doc comments in `cli/mod.rs`)
- `tm ls` empty-state hint: `tm register <url> [alias]`
- `RegistryError::NotFound` message: `run \`tm register <url> {alias}\` first`
- bundled skill `tm-cli-operations.md` command table
- `docs/specs/standalone-managed-trusty-mpm.md` SPEC-STANDALONE-MPM-01 inputs

Added: 297 / Removed: 24 / Net: +273 (of which 217 are tests and doc comments).


bobmatnyc:

Reviewed `135318bb935a0b6c86de40cd0d8bc3afcf0a3c15` (PR head), against `origin/main` at `78fed5de1d705b9cc45a7f01259cdbb9cb5cd843`. Built and exercised the real binary in a throwaway worktree (`.claude/worktrees/critic-4914`) against `mktemp -d` roots; the operator's `~/.trusty-mpm` was never touched.

The design holds where the PR says it holds. What it does not hold is the claim that `looks_like_url` closes the one-argument silent case — it closes exactly one shape of it.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 101 (`looks_like_url`) / 52-61 (one-arg arm) | `contains('/')` is the only gate on the one-argument path, so a bare `owner/repo` shorthand is accepted as a URL. `owner_repo_from_git_remote` then treats `owner` as the **host** and derives the alias from the repo alone. `tm register bobmatnyc/trusty-tools` exits 0 and writes `trusty-tools → bobmatnyc/trusty-tools` — an unclonable URL under a wrong alias, with no warning. Worse, that alias differs from the canonical one, so a later correct `tm register https://github.com/bobmatnyc/trusty-tools` does **not** collide; you end up with two entries, one broken, and no signal. Host-only URLs without a trailing slash slip through the same hole (`https://example.com` → `examplecom`; only the tested `https://example.com/` form errors). | In the `None` (one-positional) arm only, require a real remote shape before deriving: `contains("://")` or `starts_with("git@")` or an scp-style `user@host:` or a filesystem prefix (`/`, `./`, `../`, `~`). Reject a bare two-segment `a/b` whose first segment has no dot, with the actionable message `'owner/repo' is a shorthand, not a clone URL — pass the full URL, e.g. https://github.com/owner/repo`. Separately, reject when `strip_scheme(url)` exposes no `/` at all, so `https://example.com` errors the same way `https://example.com/` already does. Extend `host_only_url_errors` to cover the no-trailing-slash form and add an `owner_repo_shorthand_errors` case. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 82 (`derive_alias`) | A browser-pasted URL derives a nonsense alias, silently. `https://github.com/bobmatnyc/trusty-tools/tree/main` → `tree-main`; `.../pull/4914` → `pull-4914`; `.../repo?tab=readme` → `owner-repotabreadme`; `.../repo#readme` → `owner-reporeadme`. `owner_repo_from_git_remote` takes the **last two** path segments — correct for GitLab subgroups, which is why it exists, but wrong for the paste-the-URL flow this PR newly invites. Registration succeeds; the failure only surfaces at `tm load`. | At minimum strip `?…` and `#…` before delegating (a query or fragment is never part of a clone URL). For the extra-path-segment case, either refuse a GitHub-host URL whose path has more than two segments, or echo the derived alias with a `derived alias: <x> — pass an explicit alias to override` hint so a wrong derivation is visible at registration time rather than at load time. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/standalone.rs` | 13 | Module doc still reads `Test: exercised by 'cli_parses_register', 'cli_parses_ls', etc. in tests.rs`. `cli_parses_register` does not exist anywhere in the tree (`grep -rn 'fn cli_parses_register'` → no match). The PR updated the sibling pointer at line 25 and left this one. | Point it at `register_args_tests.rs`, or drop the dead name. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 47-50 | A legacy invocation whose URL has no `/`, no `://` and no `git@` now fails: `tm register mine myrepo.git` → `neither argument looks like a repository URL`. Bare local repo directory names were valid clone targets before. Loud, not silent, and the workaround is `./myrepo.git`, so this is a documentation gap rather than a correctness one. | One line in the changelog fragment naming the shape that stopped working. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 43-46 | `tm register https://github.com/a/b my/alias` reports `both arguments look like URLs`, which misdescribes a typo'd alias. | When exactly one argument has a scheme or `git@` prefix, prefer that as the URL and let the other fail `validate_alias` with the precise message. | Parent |

## Required Changes (WARN)

1. Close the one-argument hole in `looks_like_url` / the `None` arm — bare `owner/repo` shorthand and scheme-only host URLs must error, not register. Add the two regression tests.
2. Correct the overstated claim while you are there: the PR body, the module docs (`register_args.rs:5-9`), and `register_args_tests.rs:5-8` all assert `looks_like_url` catches the silent one-argument case. It catches the no-slash case only.

## What I verified and found correct

- **Disjointness (the load-bearing claim).** Holds for *valid* aliases. `ALIAS_RE` is `^[a-z0-9][a-z0-9._-]*$` (`registry.rs:26`), which admits no `/`, `:`, or `@`, and `looks_like_url` triggers only on those. Confirmed by code, not just by the PR's assertion. The hole above is in the other direction — non-URLs accepted as URLs — not in alias/URL overlap.
- **Derived aliases can never fail `validate_alias`.** `slugify_string` (`crates/trusty-common/src/slug.rs:34`) emits only `[a-z0-9-]`, never leading or trailing `-`, and `parse_owner_repo_slugs` returns `None` on an empty slug. Checked live: `My_Org/Cool_App.git` → `my-org-cool-app`; `foo.github.io` → `owner-foogithubio`.
- **Every URL shape the issue names derives correctly.** `.git` suffix, trailing slash, `git@host:owner/repo.git`, non-GitHub hosts, `ssh://…:2222/…`, `file:///…` — all correct, and the hyphen collapse is `owner_repo_from_git_remote`'s own storage-safety invariant (`palace_id.rs:26-32`), not bolted on here.
- **Collision does not mutate.** `add` runs `validate_alias` then returns `DuplicateAlias` before pushing (`registry.rs:194-207`); `save()` is unreachable on that path. Verified live: same sha256 before and after a refused registration, no `.json.tmp` sibling left behind.

```
registered a-b-c → https://github.com/a/b-c
beca0fa4…dd23  registry.json
Error: failed to register alias 'a-b-c' (derived from 'https://github.com/a-b/c') — pass an explicit alias to register it under a different name
Caused by: alias 'a-b-c' is already registered as 'https://github.com/a/b-c'; use --force to overwrite
beca0fa4…dd23  registry.json
REGISTRY BYTE-IDENTICAL
```

- **The accepted `a/b-c` vs `a-b/c` ambiguity refuses rather than rebinds**, and the error names the derivation — see the output above. `--force` and the explicit-alias escape hatch both recover.
- **Stale surfaces are clean.** `grep -rn "tm register"` across the worktree leaves only the deliberate legacy mentions. No `<ALIAS> <URL>`-order text survives in docs, bundled skills, agent assets, or tests. There is no tcode mirror of `tm-cli-operations.md` or the `tm-capabilities` reference — `crates/trusty-mpm/src/assets/` is the only copy.
- **Changelog fragment** `crates/trusty-mpm/changelog.d/4912-register-url-first.md` present, line 1 is a bare `Changed`. `// #4912:` comments at every change site. No test weakened, ignored, cfg-gated, or deleted (`git diff 78fed5de...HEAD` shows zero removed assertions or `#[test]`s).

## Gates run

```
cargo test -p trusty-mpm --bin tm -- --test-threads=1 register
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 1435 filtered out

( cargo fmt --check && check_line_cap.sh && check_sld.sh && check_changelog_fragment.sh )  EXIT=0
line-cap: measured 3690 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
sld-lint: scanned 55 spec doc(s) + 3087 code file(s); 0 error(s), 0 warning(s)
changelog-fragment gate: scanned 11 changed path(s); all 1 crate(s) with source changes are recorded.
```

`cargo check --workspace` not re-run — the Tauri `frontendDist` failure in `trusty-mpm-gui` / `trusty-code-gui` is the documented baseline and this diff touches neither crate.

bobmatnyc:
Review round 1 addressed. Rebased onto `78fed5de`, head is now `2ed3cbfb`.

**HIGH — closed by rejecting, not expanding.** `looks_like_url` required only a `/`. It now requires a scheme, a `git@` prefix, a local path (`/…`, `~/…`), or a host-shaped first segment (contains a `.`, or is `localhost`) followed by a non-empty path. Measured on the rebuilt binary:

```
$ tm register bobmatnyc/trusty-tools --root $TMP
Error: 'bobmatnyc/trusty-tools' is not a repository URL. Pass the full URL,
e.g. https://github.com/<owner>/<repo> or git@github.com:<owner>/<repo>.git
(usage: tm register <url> [alias])
exit=1
```

No expansion to a GitHub URL — that stays your call with the owner. `rejected_url_writes_nothing` pins that a rejected URL creates no `registry.json` at all.

Host-only now errors in **both** forms; `host_only_url_errors` covers `https://example.com`, `https://example.com/`, `https://example.com:8080`, and `git@github.com:`:

```
Error: 'https://example.com' has no owner/repo path — it names a host, not a repository.
```

**MEDIUM — I did both things you offered.** Query strings and fragments are stripped before the URL is stored (`…/o/r?tab=readme-ov-file` now stores `…/o/r` and derives `owner-repo`). And a URL whose path carries a web-UI segment after `owner/repo` is rejected:

```
Error: 'https://github.com/bobmatnyc/trusty-tools/tree/main' points inside a repository
('/tree/…'), not at it. Pass the repository root, e.g. https://github.com/<owner>/<repo>
```

The keyword list is matched only at path index ≥ 2 **and** only when a segment follows, so a repo actually named `tree` still registers (`repo_named_like_a_web_path_is_ok`: `github.com/owner/tree` → `owner-tree`, `gitlab.com/group/sub/blob` → `sub-blob`). The shared parser is untouched; the rejecting lives in a boundary validator, `path_segments`, documented as a validator rather than a second deriver and pinned by `path_segments_matches_every_url_shape` against the shapes both see.

Writing that validator surfaced a bug of my own: dropping a leading all-digit segment as a port would have eaten an all-numeric owner. The drop now keys off the `:` host terminator, mirroring trusty-common's port-vs-scp rule. `path_segments_keeps_numeric_owner` pins `github.com/123/repo` → `123-repo`.

**LOW — dead doc pointer cleared.** `standalone.rs:13` named `cli_parses_register` and `cli_parses_ls`; neither exists. Both replaced with real pointers. `grep` confirms no `cli_parses_register` remains anywhere.

**LOW — `myrepo.git` now errors**, since it has no slash. It was accepted before only because the URL positional was never validated at all; a relative path is not clonable from the managed root. Recorded as a behaviour change in the changelog fragment alongside the other newly-rejected shapes.

**Overstated claim corrected.** The module docs and PR body said `looks_like_url` blocked the silent one-argument case. It did not, exactly as you measured. Both are rewritten to describe what the code does, and the module now states the shorthand hazard as the reason the test is strict. `--help` also gained "A FULL URL is required: `gh`-style `owner/repo` shorthand is rejected, not expanded".

**Gates, rung 6, re-run on the rebased tree.** One correction to my previous report's method: I had been writing gate output to `/tmp/gN.txt`, which is shared across concurrent agent sessions — one clippy log came back describing a different worktree entirely. All gates below were re-run into a session-private scratchpad and verified to reference this worktree.

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 7809 passed, 0 failed across 45 result lines; largest binary `4602 passed; 0 failed; 7 ignored` |
| `check_line_cap.sh` / `check_sld.sh` / `check_capabilities.sh` | EXIT=0 |
| `cargo check --workspace --exclude {trusty-mpm-gui,trusty-code-gui}` | EXIT=0 (the two exclusions are the documented Tauri `frontendDist` baseline, untouched by this diff) |

`register_args` tests went 15 → 24, all passing. Smoke run against a throwaway `--root`: shorthand, both host-only forms, and both browser-paste forms all error; explicit alias, legacy alias-first order, no-owner fallback, and query-string normalisation all register correctly. The default `~/.trusty-mpm` registry was never written.

bobmatnyc:
Owner ruling implemented. Head `cfe71cd4`, based on `78fed5de`.

`owner/repo` is now the primary form, GitHub assumed:

```
$ tm register bobmatnyc/trusty-tools --root $TMP
registered bobmatnyc-trusty-tools → https://github.com/bobmatnyc/trusty-tools
exit=0
```

The ruling is consistent with existing behaviour, which is why it lands cleanly: `tm launch` already gates on `is_github_remote` and refuses non-GitHub remotes. Nothing in `tm register` made `owner/repo` mean anything else — an alias must match `^[a-z0-9][a-z0-9._-]*$`, so it can never contain a `/`. There was no conflict to work around.

**The feature and the HIGH were the same code path.** `contains('/')` was the only gate, so shorthand was read as a URL whose host was `owner` — unclonable, alias `trusty-tools`, exit 0, no collision with the later correct entry. Classification is now explicit, and the arm order is load-bearing:

| Input shape | Outcome |
|---|---|
| scheme, `git@`, `/…`, `~/…`, or host-shaped first segment (a `.`, or `localhost`) | full URL / local clone source |
| `./…`, `../…`, `.`, `..` | refused — resolved against cwd but cloned from elsewhere |
| exactly two `[A-Za-z0-9._-]` segments, no scheme, no `:` | GitHub shorthand → `https://github.com/<owner>/<repo>` |
| anything else | error naming both accepted forms |

The path guards run **before** the host-shape test on purpose: `./repo` would otherwise match "first segment contains a dot" and be read as a host. A GitHub owner cannot contain a dot, which is what separates `github.com/owner/repo` from `owner/repo` without maintaining a host list.

**One judgment call to flag, cheap to overrule.** Absolute (`/srv/git/repo.git`) and home-relative (`~/x`) paths stay ACCEPTED as local clone sources — they worked before #4912 and are unambiguous. Only `./` and `../` are refused. Your item 5 required all four not be treated as shorthand, which holds either way; if you want all four refused, it is a one-line move of the guard.

**Every guard from the last round survives**, verified in the smoke run against a throwaway `--root`:

```
Error: './repo' is a relative path, not a repository. …Pass <owner>/<repo>, a full URL, or an absolute path
Error: 'https://example.com' has no owner/repo path — it names a host, not a repository.
Error: 'https://github.com/o/r/tree/main' points inside a repository ('/tree/…'), not at it.
Error: 'a/b/c' does not name a repository. Pass <owner>/<repo> (GitHub is assumed), or a full URL…
```

And the things that must keep working:

```
registered 123-repo    → https://github.com/123/repo          # numeric owner, port-drop fix holds
registered shiny       → https://github.com/owner/repo        # explicit alias
registered legacy-proj → git@gitlab.com:team/thing.git        # legacy alias-first order
registered acme-widget → https://bitbucket.org/acme/widget.git # full non-GitHub URL
```

`same_url_reregistration_is_idempotent` now proves shorthand and the full URL are ONE registration, not two — they share a single derivation path because shorthand resolves to a URL before deriving.

**Docs updated to make shorthand primary:** `--help` (both the command doc and the `<URL>` arg doc), the `tm ls` empty-state hint, `RegistryError::NotFound`, the bundled `tm-cli-operations.md`, the regenerated capabilities reference, the changelog fragment, and SPEC-STANDALONE-MPM-01.

**Gates — rung 6, all written to a session-private scratchpad and grepped for this worktree's path before trusting any exit code.** Thanks for confirming the `/tmp/gN.txt` collision; that is how the bogus EXIT=0 clippy log appeared.

| Gate | Result | Log references this worktree |
|---|---|---|
| `cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | EXIT=0 | yes (3 refs) |
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 7817 passed, 0 failed across 45 result lines; largest binary `4602 passed; 0 failed; 7 ignored` | yes (3 refs) |
| `check_line_cap.sh` / `check_sld.sh` / `check_capabilities.sh` | EXIT=0 | — |
| `cargo check --workspace --exclude {trusty-mpm-gui,trusty-code-gui}` | EXIT=0 | yes (7 refs) |

`register_args` tests went 24 → 28. `register_args.rs` is 353 SLOC, under the 500 cap. The two workspace exclusions remain the documented Tauri `frontendDist` baseline, untouched by this diff. The default `~/.trusty-mpm` registry was never written.

bobmatnyc:
## Code Critic — Round 2 — Verdict: WARN

Reviewed `cfe71cd4a41aaec34e72a3efca8cf2be652f92a7`, based on `78fed5de`. Built the real binary in my own detached worktree (`.claude/worktrees/critic-4914b`) and exercised it against `mktemp -d` roots; the operator's `~/.trusty-mpm` registry was never written.

**The round-1 HIGH is closed.** `owner/repo` now resolves to `https://github.com/owner/repo` with the canonical `owner-repo` alias, through one derivation path, and re-registering the full URL is the same entry rather than a second one — measured, not taken on assertion:

```
$ tm register bobmatnyc/trusty-tools --root $TMP
registered bobmatnyc-trusty-tools → https://github.com/bobmatnyc/trusty-tools
$ tm register https://github.com/bobmatnyc/trusty-tools --root $TMP
registered bobmatnyc-trusty-tools → https://github.com/bobmatnyc/trusty-tools
$ jq length registry.json
1
```

What blocks the verdict is a red CI gate this diff created, plus three input shapes that still register something `git clone` cannot fetch.

## Findings

|---|---|---|---|---|---|
| HIGH | `.test-pointer-allowlist.tsv` | 358, 392 | **CI is red on this head.** `Test: doc-comment pointer lint` fails: removing the `cli_parses_register` pointers from `cli/mod.rs:792` and `standalone.rs:25` left their allowlist rows stale, and the gate rejects stale rows as hard as dangling ones. `test-pointers: 0 dangling pointer(s), 2 stale allowlist entries — FAILED` (exit 1), reproduced locally and confirmed on the PR: [job 92340590040](https://github.com/bobmatnyc/trusty-tools/actions/runs/31016149430/job/92340590040). It is caused by this diff, not baseline — at `78fed5de` both pointers exist, so both rows were live. Not in the gate table you reported. | `bash scripts/check_test_pointers.sh --update`, commit the pruned `.test-pointer-allowlist.tsv`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 266-277 | **The browser-paste guard misses every single-word GitHub tab URL** — the most-pasted shape of all. `interior = len - 3` exempts the final segment so a repo named `tree` survives, but on a 3-segment path the keyword *is* the final segment, so nothing fires: `.../o/r/issues` → `registered r-issues → https://github.com/o/r/issues`, exit 0. Same for `/pulls`, `/releases`, `/wiki`, `/actions`. Unclonable, wrong alias, and `r-issues` does not collide with the canonical `o-r`, so there is no later signal. This is the round-1 MEDIUM only partly closed; `--help` and SPEC-STANDALONE-MPM-01 both now claim browser paths into a repo are refused. | Make the guard host-aware instead of position-aware: when the host is `github.com`, a path with more than 2 segments is never a repository (GitHub has no subgroups) — that rejects `/issues` and `/tree/main` alike while leaving the GitLab subgroup case (`gitlab.com/group/sub/blob`) untouched. Extend `browser_paste_shapes_are_rejected` with the four 3-segment forms. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 109-111, 241 | **A scheme-less host registers a string git cannot clone, exit 0.** `first.contains('.')` routes `github.com/owner/repo` to `Url` and line 241 stores it verbatim. Measured: `git clone --depth 1 github.com/bobmatnyc/trusty-tools` → `fatal: repository 'github.com/bobmatnyc/trusty-tools' does not exist`. `a.b/c` → `registered c → a.b/c` and `localhost/repo` → `registered repo → localhost/repo` are the same hole. `derives_from_every_url_shape:243` and `path_segments_matches_every_url_shape:355` pin the shape as valid. Mitigated for the `github.com/...` case only, because its alias collides with the canonical one. | In `resolved_url`, when the classified `Url` has no `://`, no `git@`, and no leading `/` or `~/`, prepend `https://` before storing. One line, and it makes the most likely half-remembered form actually work. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 99-101 | **`~/…` is accepted, tested, and can never clone.** Nothing expands the tilde: `clone_repo` (`core/standalone/load.rs:167-171`) is `Command::new("git").args(["clone", …, url, "repo"])` with no shell. Measured: `registered src-repo → ~/src/repo.git`, then `git clone --depth 1 '~/src/repo.git'` → `fatal: repository '~/src/repo.git' does not exist`. The derived alias is also junk — `path_segments` eats `~` as the host, yielding `src-repo`. `register_args_tests.rs:145` asserts this input is `is_ok()`, so the broken shape is now pinned. Absolute paths are genuinely fine; only the home-relative form is. | Expand `~/` against `dirs::home_dir()` in `resolved_url` so the stored URL is clonable (the rejection message already advertises absolute paths), or drop `~/` from line 99 and let it fall through to the rejection. Either way change the assertion at `register_args_tests.rs:145`. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/standalone.rs` | 70 | `ls_cmd` still reads `Test: `cli_parses_ls` in tests.rs.` — no such test exists (only `cli_parses_ls_connector_bare`, `_projects`, `_json`, `_current`, `_source_id`, `_terms_*`). The lint tolerates it because row 390 of the allowlist still covers it, but the module doc four lines up now says that name doesn't exist. Round 1 asked for both dead pointers; one survived. | Point it at the real `cli_parses_ls_*` tests and drop allowlist row 390. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 95-97 | The relative-path guard covers `./` and `../` only, so a dot-prefixed directory still reads as a host: `.hidden/repo` classifies as `Url` and reaches the registry. The module doc at line 60 states relative paths are refused without that qualification. Implausible input; noted only because the doc is unqualified. | Either add `s.starts_with('.')` to the guard, or qualify the doc to `./` and `../`. | Parent |

## Required Changes (WARN)

1. Prune the two stale allowlist rows so `Test: doc-comment pointer lint` goes green. The PR cannot merge red.
2. Close the three exit-0 unclonable shapes, or state in the changelog which are deliberately out of scope. The `/issues` paste is the one I would not ship as-is.

## Verified, and correct

- **Arm order is load-bearing and the implementation has it right.** Without the path guard at line 95, `./repo` splits to `first = "."`, which passes `first.contains('.')` at line 109 and would classify as a URL with host `.`. Guard first, so it doesn't. Confirmed against the binary: `Error: './repo' is a relative path, not a repository.`
- **The dot rule is a correct fact about GitHub, not a guess.** Usernames and organization names share one namespace and permit only alphanumerics and single hyphens — no periods; repository names separately permit `.`, `_`, `-`. So `first.contains('.')` separates `github.com/owner/repo` from `owner/repo` with no host list, and `owner/repo.js` correctly stays shorthand ([github-limits](https://github.com/dead-claudia/github-limits)).
- **Idempotence holds for the reason claimed.** Shorthand resolves to a URL before `derive_alias` is reached (line 238 returns early into the same `owner_repo_from_git_remote` path), and `ManagedRegistry::add` returns `Ok(())` on an identical alias+URL (`registry.rs:196-199`). One entry, measured above.
- **Non-GitHub full URLs are untouched:** `git@gitlab.com:team/thing.git` → `team-thing`, `https://bitbucket.org/acme/widget.git` → `acme-widget`, both registered.
- **Round-1 fixes survive:** `rejected_url_writes_nothing` present and passing; `123/repo` → `123-repo` with the port-drop keyed off the `:` terminator; `?query`/`#fragment` stripped; `cli_parses_register` gone from every doc comment.
- **Refusals that do fire, measured:** `owner/repo/`, `owner//repo`, `owner/`, `owner/repo:8080`, `öwner/repo`, `/`, `./repo`, `https://example.com` (both forms), `.../tree/main`. `owner/..` fails at alias derivation rather than classification — loud either way.
- **Guard precision on the cases you named:** `github.com/owner/tree` → `owner-tree` and `gitlab.com/group/sub/blob` → `sub-blob` both still register.
- **Housekeeping:** changelog fragment line 1 is a bare `Changed`; `// #4912:` at every change site; zero tests removed, `#[ignore]`-ed, or cfg-gated (`git diff 78fed5de...cfe71cd4` has no removed `#[test]` or `assert`); `register_args.rs` 353 lines, under the 500 production cap; `register_args_tests.rs` 478, under 3000.

## Gates

All logs written to a session-private scratchpad and grepped for this worktree's path before any exit code was trusted.

| Gate | Result | Log refs this worktree |
|---|---|---|
| `cargo test -p trusty-mpm --bin tm -- --test-threads=1 register` | EXIT=0 — `test result: ok. 32 passed; 0 failed; 0 ignored; 0 measured; 1435 filtered out` | yes (8) |
| `cargo build -p trusty-mpm --bin tm` | EXIT=0 | yes (5) |
| `bash scripts/check_line_cap.sh` | EXIT=0 — `3690 tracked .rs file(s), 7 allowlisted, 0 violations` | — |
| `bash scripts/check_test_pointers.sh` | **EXIT=1 — 2 stale allowlist entries** | — |
| `gh pr checks 4914` | 1 fail (doc-comment pointer lint), 1 pending (Test), 23 pass | — |

`cargo check --workspace` not run — the Tauri `frontendDist` failure in the GUI crates is the documented baseline and this diff touches neither crate.

bobmatnyc:
Round 2 addressed. Rebased onto `b98fb77c` (main moved past #4903), head is now `9fc353a9`.

**HIGH — CI is green.** `bash scripts/check_test_pointers.sh --update` pruned exactly the two stale rows and added nothing:

```
 .test-pointer-allowlist.tsv | 0 insertions(+), 2 deletions(-)
-crates/trusty-mpm/src/bin/tm/cli/mod.rs	cli_parses_register
-crates/trusty-mpm/src/bin/tm/commands/standalone.rs	cli_parses_register
```

```
test-pointers: resolved 21590 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```

**MEDIUM — the GitHub tab URLs are refused, host-aware as you prescribed.** The guard now asks the host first: on `github.com`, more than 2 path segments is never a repository. Other hosts keep the interior-keyword rule, so nesting still works.

```
$ tm register https://github.com/bobmatnyc/trusty-tools/issues --root $TMP
Error: 'https://github.com/bobmatnyc/trusty-tools/issues' points inside a repository ('/issues/…'), not at it. Pass the repository root, e.g. <owner>/<repo>
```

Same for `/pulls`, `/releases`, `/wiki`. The host rule also catches words that are not in `NON_REPO_SEGMENTS` at all — `…/repo/settings` — which the keyword list never could. Precision holds:

```
registered owner-tree → https://github.com/owner/tree
registered sub-blob   → https://gitlab.com/group/sub/blob
```

**MEDIUM — a scheme-less host now stores something git can fetch.** Measured both ways:

```
$ git ls-remote github.com/bobmatnyc/trusty-tools
fatal: 'github.com/bobmatnyc/trusty-tools' does not appear to be a git repository

$ git ls-remote https://github.com/bobmatnyc/trusty-tools
b98fb77c13a72618806f7fcb4d5ff0fd81c54e20	HEAD
```

A side effect worth naming: it now derives the canonical alias too, so the scheme-less form and the shorthand are ONE registration rather than two. `registered bobmatnyc-trusty-tools → https://github.com/bobmatnyc/trusty-tools` from both.

**MEDIUM — `~/` is expanded, not rejected.** You offered either; expansion is the better trade. Rejecting removes a shape that works, and the tilde form is unambiguous — it names exactly one absolute path, and absolute paths are already accepted. Expanding makes `~/x` and `/home/you/x` the same input, which is what a user typing it means. It routes through `managed_root::expand_tilde`, the helper that already resolves the managed root, rather than a second tilde parser.

```
$ tm register '~/src/repo.git' home --root $TMP
registered home → /Users/masa/src/repo.git
```

`register_args_tests.rs:145` no longer asserts bare `is_ok()`; `home_relative_path_is_expanded` pins the expanded value and that no `~` survives.

**Fail-before, all three, against the previous head `cfe71cd4`** — old source, new tests:

```
test result: FAILED. 32 passed; 3 failed
---- github_tab_urls_are_rejected ----
called `Result::unwrap_err()` on an `Ok` value: ("repo-issues", "https://github.com/owner/repo/issues")
---- home_relative_path_is_expanded ----
  left: "~/src/repo.git"   right: "/Users/masa/src/repo.git"
---- scheme_less_host_gains_https ----
  left: "github.com/owner/repo"   right: "https://github.com/owner/repo"
```

`("repo-issues", …)` is your finding reproduced exactly.

**Not fixed, and why.** Both LOWs are left for the parent to rule on rather than folded in silently. `standalone.rs:70`'s `cli_parses_ls` pointer is still stale — fixing it would have made the allowlist diff three rows instead of the two this finding called for, and the two are what CI needs. `.hidden/repo` still classifies as a host. Say the word and both are one commit.

**Gates — rung 6.** All logs written to a session-private scratchpad and grepped for this worktree's path before any exit code was trusted.

| Gate | Result | Log refs this worktree |
|---|---|---|
| `cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` | EXIT=0 | yes (6) |
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 7825 passed, 0 failed, 17 ignored across 45 result lines; largest binary `4604 passed; 0 failed; 7 ignored` | yes (3) |
| `bash scripts/check_test_pointers.sh` | EXIT=0 — `0 dangling pointers` | — |
| `check_line_cap.sh` / `check_sld.sh` / `check_capabilities.sh` / `check_changelog_fragment.sh` | EXIT=0 — `3693 tracked .rs file(s), 7 allowlisted, 0 violations` | — |

`register_args` tests went 28 → 35. `register_args.rs` is 398 lines against the 500 production cap, so the fixes were added rather than compressed. The changelog fragment records both new behaviour changes. Smoke run against a throwaway `--root`; `~/.trusty-mpm/registry.json` is untouched since Jul 2 and `tm ls` was never run.

bobmatnyc:
Both LOWs folded in. Head `9ef0ad43`, on main `b98fb77c`.

**`standalone.rs:70` — dead pointer gone.** `ls_cmd` now points at `cli_parses_ls_json`, `cli_parses_ls_current`, `cli_parses_ls_projects` in `tests_behavior_d_tests.rs`, and says plainly that the printing itself is stdout-only with nothing to assert. You were right that this outweighed my scope argument — the allowlist row was the landmine, not the doc text. `--update` removed exactly that one row:

```
 .test-pointer-allowlist.tsv | 0 insertions(+), 1 deletion(-)
-crates/trusty-mpm/src/bin/tm/commands/standalone.rs	cli_parses_ls
```

```
test-pointers: resolved 21589 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```

**`.hidden/repo` — refused as a relative path.** That is the choice; the reasoning is that it *is* one. A dot-prefixed directory resolves against the process cwd and would be cloned from somewhere else, which is the identical failure mode as `./repo`, so it gets the identical message — and that message already advertises the absolute form as the remedy. Treating it as a host was never defensible: no hostname may begin with a dot.

The guard is now `s.starts_with('.')`, replacing four conditions with one:

```rust
// Paths first: any leading `.` would otherwise match the host-shape test
// below on its own dot — `./repo` splits to `first = "."` and `.hidden/repo`
// to `first = ".hidden"`. Neither is a host: no hostname may begin with a
// dot, and no GitHub owner may contain one.
if s.starts_with('.') {
    return Positional::RelativePath;
}
```

Nothing legitimate is lost, because a GitHub owner permits only alphanumerics and hyphens. Only the FIRST segment is tested, so a dot in the repo segment is untouched — `owner/repo.js` → `registered owner-repojs → https://github.com/owner/repo.js`, exit 0.

```
$ tm register .hidden/repo --root $TMP
Error: '.hidden/repo' is a relative path, not a repository. It would be resolved against the current directory but cloned from elsewhere. Pass <owner>/<repo>, a full URL, or an absolute path
exit=1
```

**Fail-before on `9fc353a9`** — old source, new cases:

```
test result: FAILED. 33 passed; 2 failed
---- classify_sorts_every_shape ----
assertion `left == right` failed: .hidden/repo misclassified
  left: Url(".hidden/repo")
---- relative_paths_are_never_shorthand ----
called `Result::unwrap_err()` on an `Ok` value: ("repo", "https://.hidden/repo")
```

Worth noting what the second line shows: combined with the scheme-prepend from the previous round, `.hidden/repo` would have registered `https://.hidden/repo`. The two fixes interact, and only in this direction — hence the test.

**Gates — rung 6, full set, logs in a session-private scratchpad and grepped for this worktree's path before any exit code was trusted.**

| Gate | Result | Log refs this worktree |
|---|---|---|
| `cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` && `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 7825 passed, 0 failed, 17 ignored across 45 result lines, 0 FAILED lines; largest binary `4604 passed; 0 failed; 7 ignored` | yes (6) |
| `bash scripts/check_test_pointers.sh` | EXIT=0 — `0 dangling pointers` | — |
| `check_line_cap.sh` / `check_sld.sh` / `check_capabilities.sh` / `check_changelog_fragment.sh` | EXIT=0 — `3693 tracked .rs file(s), 7 allowlisted, 0 violations` | — |

The first run of that chain came back EXIT=1 on `cargo fmt --check` — a test array over the line width, fixed with `cargo fmt` and re-run clean. Recording it because it is the case the redirect-then-check rule exists for: piped into `tail` it would have exited 0.

`register_args.rs` is 401 lines against the 500 cap. The changelog fragment names the dot-prefixed forms alongside the other refusals. Smoke run against a `mktemp -d` root; `~/.trusty-mpm/registry.json` untouched since Jul 2, and `tm ls` was never run.

bobmatnyc:
## Code Critic — Round 3 (final) — Verdict: APPROVE

Reviewed `9ef0ad43373a8f2c7ed5ef84f988fa285998c183`, merge-base `b98fb77c`. Built the real binary in my own detached worktree (`.claude/worktrees/critic-4914c`) and drove it against `mktemp -d` roots. `tm ls` was never run and the operator's `~/.trusty-mpm` was never touched. Every log went to a session-private path and was grepped for this worktree before any exit code was trusted.

Both round-2 HIGHs are closed, measured rather than taken on assertion. The doc-pointer lint is green on CI and locally, and all six items I was asked to verify hold. What remains is one over-refusal and two narrow evasions of the host rule — all MEDIUM, none worth another round.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 305-311 | **The web-UI keyword rule fires on LOCAL paths, so an absolute path that previously registered now cannot.** `non_repo_segment` runs for every URL including filesystem paths, and `NON_REPO_SEGMENTS` holds ordinary directory words. Measured: `/Users/me/actions/repo.git` → `Error: … points inside a repository ('/actions/…')`, exit 1. Same for `/home/bob/raw/myrepo` and `~/tree/repo.git`. There is no workaround, and the relative-path rejection message at line 344 advertises "or an absolute path" as the remedy — which this then refuses. `/srv/git/repo.git` and `/Users/me/src/repo.git` register fine, so it needs a keyword at segment index 2 of a ≥4-segment path. | Skip the guard for local paths: `if url.starts_with('/') { return None; }` at the top of `non_repo_segment` (equivalently, when `split_host` returns an empty host). The rule is about web hosts and nothing else. Add a `/…/actions/repo.git` case to `absolute_and_home_paths_are_accepted`. | Parent |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 302 | **The GitHub rule is exact host equality, so a GitHub-adjacent host string reopens the round-2 tab-URL finding.** `host.eq_ignore_ascii_case(GITHUB_HOST)` misses `www.github.com` and the trailing-dot FQDN, and a 3-segment path has `interior == 0` so the keyword rule checks nothing. Measured: `https://www.github.com/owner/repo/issues` → `registered repo-issues`, exit 0; `https://github.com./owner/repo/issues` → same. Both unclonable — `git ls-remote https://www.github.com/bobmatnyc/trusty-tools/issues` → `remote: Not Found`. The canonical `https://github.com/o/r/issues` is correctly refused, as is `:443` and the `git@` form, so this is the fringe of the shape, not its centre. | Normalise before comparing: `let host = host.trim_end_matches('.').to_ascii_lowercase();` then `host == GITHUB_HOST \|\| host.ends_with(".github.com")`. | Parent |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 387 | **`split_host` cuts userinfo at the FIRST `@`; RFC 3986 puts it at the last `@` of the authority.** A credentialed URL whose username itself contains `@` — an email login, the common shape for a PAT URL — hides the real host from every downstream rule. Measured: `https://user@evil.com@github.com/o/r/settings` → `registered r-settings → …/settings`, exit 0, because the host is read as `evil.com`. Not a security issue: git resolves the last `@`, so the clone target is still `github.com` and the stored string is verbatim what was typed. It is a missed refusal, not a redirect. | Cut the authority first (`after_scheme.split('/').next()`), then `rfind('@')` within it. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 247-249 | The `Shorthand` arm returns early, so shorthand never reaches `path_segments`/`non_repo_segment`, and `is_name_segment` admits `.`. With one positional `owner/..` is caught at derivation; with an explicit alias it registers: `tm register owner/.. myalias` → `registered myalias → https://github.com/owner/..`, exit 0. Harmless — the alias itself is still `validate_alias`-checked, so nothing escapes into a path — but it stores a URL that 404s. | Reject an all-dots segment in `is_name_segment`. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 282-283 | The scheme-prepend turns any dotted first segment into a host, so a relative directory whose name contains a dot becomes a URL rather than an error: `my.dir/repo` → `registered repo → https://my.dir/repo`, `src.old/repo` likewise. The leading-dot form is now refused; this is the residue of the same heuristic. No doc overstates it — the changelog enumerates the dot-*prefixed* forms only. | Accept as designed, or note it. Raised only because it is the same shape family as the one just closed. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 254 | `url == "~"` is unreachable. `classify("~")` returns `Other`: no `://`, no leading `.` or `/`, not `~/`, and `split_once(['/', ':'])` yields `None`. | Drop `url == "~" \|\|`. | Parent |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/register_args.rs` | 59 | `[`SHORTHAND_HOST`]` names an item that exists nowhere in the tree — the constant is `GITHUB_HOST`. Dangling intra-doc link; rustdoc's `broken_intra_doc_links` would warn. | Rename to `[`GITHUB_HOST`]`. | Parent |

## The six items, verified

**1. The invariant.** Every shape from prior rounds plus the new ones, run through the built binary against throwaway roots:

```
owner/repo/                        exit=1  does not name a repository        registry: (none)
owner//repo                        exit=1  does not name a repository        registry: (none)
owner/                             exit=1  does not name a repository        registry: (none)
..hidden/repo                      exit=1  is a relative path                registry: (none)
.../o/r/settings                   exit=1  is a relative path                registry: (none)
''                                 exit=1  does not name a repository        registry: (none)
owner/..                           exit=1  cannot derive an alias            registry: (none)
owner/.                            exit=1  cannot derive an alias            registry: (none)
https://GITHUB.COM/owner/repo/issues     exit=1  points inside a repository
https://github.com:443/owner/repo/issues exit=1  points inside a repository
git@github.com:owner/repo/tree/main      exit=1  points inside a repository
ftp://github.com/owner/repo/issues       exit=1  points inside a repository

owner/repo.git                     exit=0  → https://github.com/owner/repo.git   (clonable, alias owner-repo)
'  owner/repo  '                   exit=0  → https://github.com/owner/repo       (trimmed correctly)
https://user:pass@github.com/owner/repo  exit=0  → verbatim, alias owner-repo    (clonable)
https://münchen.example/o/r        exit=0  → verbatim, alias o-r                 (IDN preserved; git handles it)
münchen.example/o/r                exit=0  → https://münchen.example/o/r
//owner/repo                       exit=0  → //owner/repo
```

`//owner/repo` is the accepted-absolute-path class, not a defect: `/srv/git/repo.git` is deliberately stored as-is with no existence check, and `//owner/repo` is the same thing. The three genuine residuals are the MEDIUM rows above.

**2. The compound interaction.** `.hidden/repo` is refused, and `https://.hidden/repo` cannot be produced. The prepend at line 282 fires only when the string has no `://`, no `git@`, and no leading `/` — which after the dot guard leaves exactly the host-shaped-first-segment branch. A tilde is expanded to an absolute path before that point, so it takes the `/` exit. The only remaining bad interaction is the dotted-relative-directory case, LOW row 5 above: `my.dir/repo` → `https://my.dir/repo`. Same family, one order of magnitude less plausible than a leading dot.

**3. The collapsed guard.** `s.starts_with('.')` tests the first character of the whole positional, never a later segment, so the repo half is untouched. Measured:

```
owner/repo.js   → registered owner-repojs → https://github.com/owner/repo.js
owner/.github   → registered owner-github → https://github.com/owner/.github
```

`owner/.github` is the real GitHub special repo and still registers. Nothing legitimate is lost: no hostname may begin with a dot and no GitHub owner may contain one, so the only inputs the guard newly refuses are paths.

**4. The `standalone.rs:70` pointer.** All three tests exist and are what the pointer says they are — `Cli::try_parse_from` flag assertions, not `ls_cmd` coverage:

```
crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs:803  fn cli_parses_ls_projects()
crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs:823  fn cli_parses_ls_json()
crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs:836  fn cli_parses_ls_current()
```

The pointer scopes itself to "flag parsing in …" and then says the printing is stdout-only with nothing to assert. That is an accurate pointer to weak coverage, which is what the lint is for.

**5. The allowlist.** Verified against the diff, not the claim.

```
$ git diff --numstat 9fc353a9 9ef0ad43 -- .test-pointer-allowlist.tsv
0	1
-crates/trusty-mpm/src/bin/tm/commands/standalone.rs	cli_parses_ls

$ git diff --numstat b98fb77c 9ef0ad43 -- .test-pointer-allowlist.tsv
0	3     (the two cli_parses_register rows plus the one above)

test-pointers: resolved 21589 Test: citation(s) (floor 200) — 0 dangling pointers — OK.   EXIT=0
```

One row removed this round, none added; three removed and none added across the whole PR.

**6. Gates and the test inventory.** Full suite re-run from my worktree, single-threaded, redirected to a file and checked by exit code:

```
cargo test -p trusty-mpm -- --test-threads=1        EXIT=0
45 result lines, 0 FAILED — passed=7825 failed=0 ignored=17
largest binary: test result: ok. 4604 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 174.10s
```

Matches the reported numbers exactly. Script gates from the same worktree: `check_test_pointers.sh`, `check_line_cap.sh` (`3693 tracked .rs file(s); 7 allowlisted, 0 violations`), `check_sld.sh`, `check_changelog_fragment.sh` — all EXIT=0.

`register_args.rs` is 402 lines / 383 SLOC against the 500 cap (the report said 401; the cap gate is green either way). Across all three rounds `git diff b98fb77c 9ef0ad43` removes zero `#[test]` items, adds zero `#[ignore]`, adds no `cfg` gate, and removes no assertion. Twelve files changed, +1056/−29.

CI on this head: 22 pass, 2 pending (`Test`, `trusty-search daemon smoke test`), 0 fail. `Test: doc-comment pointer lint` is green.

## Notes

- The three MEDIUMs are all one- to three-line changes. My reading is that none of them justifies a fourth round: the local-path over-refusal is loud and narrow, and the two host-rule evasions require a host string GitHub does not serve. If the PM would rather fold the local-path one in before merge, the fix is the single `if url.starts_with('/') { return None; }` above and it needs one new test case.
- `cargo check --workspace` not run — the Tauri `frontendDist` failure in the GUI crates is the documented baseline and this diff touches neither crate.
- Alias path-traversal is closed and I checked it rather than assuming: `ManagedRegistry::add` calls `validate_alias` before any mutation (`registry.rs:194-195`), and `ALIAS_RE` is `^[a-z0-9][a-z0-9._-]*$`, which admits no `/`. A derived alias that slugifies to empty returns `None` and errors, which is why `owner/..` fails loudly.

bobmatnyc:
Local-path regression fixed. Head `dc8e3baf`, on main `b98fb77c`.

**One deviation from the specified fix, and it matters.** `if url.starts_with('/')` would not have covered `file:///Users/me/actions/repo.git`, which reproduces identically:

```
$ tm register file:///Users/me/actions/repo.git --root $TMP
Error: 'file:///Users/me/actions/repo.git' points inside a repository ('/actions/…'), not at it.
```

So the guard keys off an **empty host** instead. The keyword rule is about web-UI paths, and those only exist behind a host — that is the actual principle, and it covers all four local shapes in one condition:

```rust
// #4912: a local path has no host, so it has no web UI either and none of
// these words mean anything in it. …
if host.is_empty() {
    return None;
}
```

`~/…` is expanded to an absolute path *before* this runs, so the same guard covers it. The test pins that ordering explicitly, so a future reorder fails loudly rather than silently reintroducing this.

**Fixed, measured on the rebuilt binary:**

```
/Users/me/actions/repo.git          exit=0  registered … → /Users/me/actions/repo.git
/home/bob/raw/myrepo                exit=0  registered … → /home/bob/raw/myrepo
~/tree/repo.git                     exit=0  registered … → /Users/masa/tree/repo.git
file:///Users/me/actions/repo.git   exit=0  registered … → file:///Users/me/actions/repo.git
```

**Remote guards unchanged** — all still exit 1:

```
https://github.com/o/r/issues                    exit=1
https://github.com/o/r/tree/main                 exit=1
https://gitlab.com/group/sub/repo/-/blob/main/x  exit=1
https://example.com                              exit=1
```

**Precision unchanged** — all still exit 0: `owner/repo.js`, `owner/.github`, `github.com/owner/tree`, `gitlab.com/group/sub/blob`, `bobmatnyc/trusty-tools`.

**Fail-before on `9ef0ad43`:**

```
test result: FAILED. 35 passed; 1 failed
---- local_paths_skip_the_web_path_rule ----
/Users/me/actions/repo.git must register, got: '/Users/me/actions/repo.git'
points inside a repository ('/actions/…'), not at it.
```

No changelog change: the regression was introduced and fixed inside this PR, so there is no net user-visible behaviour to record. Local paths worked before #4912 and work after.

**Gates — rung 6, full set.** Logs in uniquely-named session-private files, each grepped for this worktree's path before its exit code was trusted.

| Gate | Result | Log refs this worktree |
|---|---|---|
| `cargo fmt --check` && `cargo clippy -p trusty-mpm --all-targets -- -D warnings` && `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 7827 passed, 0 failed, 17 ignored across 45 result lines, 0 FAILED lines | yes (6) |
| `check_test_pointers.sh` | EXIT=0 — `0 dangling pointers`; allowlist unchanged this round (the new pointer names a real test) | — |
| `check_line_cap.sh` / `check_sld.sh` / `check_capabilities.sh` / `check_changelog_fragment.sh` | EXIT=0 — `3693 tracked .rs file(s), 7 allowlisted, 0 violations` | — |

`register_args.rs` is 410 lines against the 500 cap. Smoke run against `mktemp -d` roots; `~/.trusty-mpm/registry.json` untouched since Jul 2, `tm ls` never run.

Leaving the two you claimed — exact-equality GitHub host match (`www.github.com`) and `split_host` cutting userinfo at the first `@` — untouched for your follow-up tickets.

Issue #4059, opened 2026-07-27
trusty-installer: LaunchAgent plists and installer paths disagree on binary names (trusty-mpm vs tm)


bobmatnyc:
## Summary

On Bob's installed system, two LaunchAgent plist files installed by the same installer disagree on which binary to invoke. Additionally, installer code paths only recognize one binary name and fail silently on the other. This produces a live system where behavior depends on which binary name was used at install time.

## Verified Configuration Issues

**Two LaunchAgent plists disagree:**
- `~/Library/LaunchAgents/com.trusty.mpm.plist` invokes:
  ```
  /Users/masa/.cargo/bin/trusty-mpm daemon
  ```
- `~/Library/LaunchAgents/com.trusty.mpm.supervisor.plist` invokes:
  ```
  /Users/masa/.cargo/bin/tm supervisor
  ```

**Installer code paths only recognize one binary:**

1. `crates/trusty-installer/src/commands/plist_bootstrap.rs:541-557` — `resolve_tm_binary()`:
   - Only searches for `tm`: `which("tm")` → `~/.local/bin/tm` → `~/.cargo/bin/tm`
   - No fallback to `trusty-mpm` if `tm` not found
   - Falls back to default install dir + "tm"
   - Comment on line 533: "Resolve the `tm` binary path" (hardcoded expectation)

2. `crates/trusty-installer/src/commands/macos_signing/mod.rs:112` — signed binaries:
   - Dedicated codesigning table row for `("tm", MPM_SET, "com.trusty.tm")`
   - Comment cites issue #2721 and TCC-grant revocation risk if unsigned
   - If `tm` binary is missing, bootstrap can fail silently while signing proceeds on the other binary name

## Why This Matters

- **Silent degradation**: If only one binary name is present after install, the system partially works but the supervisor or daemon launches under the wrong binary, affecting TCC/signing behavior.
- **Non-deterministic behavior**: System behavior depends on which binary name happened to be used at install time, not on which binaries are actually installed.
- **Security angle**: The TCC grant and codesigning identity for "com.trusty.tm" can be revoked if the bootstrap process mixes binary names mid-session (#2721 context).

## Underlying Problem: Hand-Maintained Binary Name Tables

This is a symptom of the same root cause as issue #4058: the same fact (what this crate's binary is called) is written in many places and they have drifted:
- Two LaunchAgent plist files disagree
- `resolve_tm_binary()` only looks for `tm`, never `trusty-mpm`
[...]

Issue #2938, opened 2026-07-17
launchd hygiene: stale duplicate com.trusty.trusty-search.plist alongside active com.trusty.search.plist


bobmatnyc:
Found during the 0.33.1 daemon restart (2026-07-17): two launchd jobs point at the same trusty-search binary — `~/Library/LaunchAgents/com.trusty.search.plist` (active, running) and `~/Library/LaunchAgents/com.trusty.trusty-search.plist` (loaded definition present but not running; leftover from an older install layout). 

The stale one is dead weight and a footgun: a future `launchctl bootstrap` against the wrong label could start a second daemon instance fighting over port 7878 and index locks. 

Proposal: remove the stale plist (after confirming nothing references it), and add a `tm doctor` / trusty-search install-script check that warns when more than one LaunchAgent references the trusty-search binary.

bobmatnyc:
Not closing — PR #4919's "closes as a side effect" claim is partially verified, not fully.

**What #4919 actually changed, evidence:**

`crates/trusty-common/src/launchd_labels.rs` registers `trusty-search`'s legacy labels as `["com.trusty.trusty-search", "com.bobmatnyc.trusty-search"]` (the first is the exact label this issue names). `crates/trusty-search/src/commands/service.rs`'s `service_install` now calls `cfg.install_and_activate_forced(legacy_labels_for(LAUNCHD_LABEL), force)`, and `LaunchdConfig::evict_legacy` in `crates/trusty-common/src/launchd_activate.rs:601-619` boots out each legacy label and deletes its plist file (`std::fs::remove_file`) before the canonical unit is (re)loaded. So a fresh `trusty-search service install` run genuinely would delete `~/Library/LaunchAgents/com.trusty.trusty-search.plist` and boot out its job — that mechanism is real and covered by `reload_needed_*` / `rollback_*` unit tests.

**What remains:**

1. This is prospective, not yet done. #4919 merged at 2026-08-05T17:23:46Z; its own "Operator note" says "an already-installed host needs one `trusty-search service install` to pick this up." Nobody has re-run that command on the host this issue was filed against (2026-07-17), so the stale `com.trusty.trusty-search.plist` this issue reports has not actually been removed yet — only the code path that would remove it now exists.
[...]

Issue #2827, opened 2026-07-16
install-mpm-signed suggests wrong launchd plist name (com.trusty.trusty-mpm.plist instead of com.trusty.mpm.plist)


bobmatnyc:
## What happened

During the trusty-mpm 0.19.19 release install (`make install-mpm-signed`), the signed-install script's post-install output suggested restarting the daemon via:

```
launchctl bootout   gui/$(id -u) ~/Library/LaunchAgents/com.trusty.trusty-mpm.plist
launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.trusty.trusty-mpm.plist
```

That plist **does not exist** on disk. Running the suggested commands failed:

```
$ launchctl bootout gui/$(id -u) ~/Library/LaunchAgents/com.trusty.trusty-mpm.plist
Boot-out failed: 5: Input/output error
Try re-running the command as root for richer errors.

$ launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.trusty.trusty-mpm.plist
Bootstrap failed: 5: Input/output error
Try re-running the command as root for richer errors.
```

`launchctl print gui/$(id -u)/com.trusty.trusty-mpm` also fails:

```
Bad request.
Could not find service "com.trusty.trusty-mpm" in domain for user gui: 502
```

The actual daemon plist/label is `com.trusty.mpm` — `~/Library/LaunchAgents/com.trusty.mpm.plist`. Using the correct label worked cleanly:

```
$ launchctl bootout gui/$(id -u) ~/Library/LaunchAgents/com.trusty.mpm.plist
$ launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.trusty.mpm.plist
$ launchctl list | grep com.trusty.mpm
1624	0	com.trusty.mpm.supervisor
28921	0	com.trusty.mpm
```

Daemon came back healthy on the new PID (28921), confirmed via `tm --version` (`trusty-mpm 0.19.19`) and the daemon's own `/health` endpoint (`{"status":"ok",...,"supervised":true}`).

## Root cause / offending code

`scripts/install-trusty-mpm-signed.sh`, lines 274 and 276, hardcode the wrong plist filename in the printed restart hint:

```
274:    printf '  launchctl bootout   gui/$(id -u) ~/Library/LaunchAgents/com.trusty.trusty-mpm.plist\n'
[...]

PR #4919, opened 2026-08-05
fix(trusty-common): make service install target the launchd unit that exists

bobmatnyc:
## What was broken

`trusty-search service install` wrote and bootstrapped `com.trusty.trusty-search`.
The unit launchd actually has loaded is **`com.trusty.search`**. Verified on the
owner's host, read-only:

```
$ launchctl list | grep -i trusty
30736  0  com.trusty.search
-      1  com.trusty.trusty-search.logrotate      # orphan: no main unit beside it
$ grep -A1 '<key>Label</key>' ~/Library/LaunchAgents/com.trusty.search.plist
	<key>Label</key>
	<string>com.trusty.search</string>
$ grep -c ExitTimeOut ~/Library/LaunchAgents/com.trusty.search.plist
0
```

So an install evicted nothing, bootstrapped a **second** daemon contending for
:7878 and the index locks, and left the `ExitTimeOut` key added in #4868 sitting
in a plist launchd never reads — that last `0` is the proof the fix never
reached the live unit. #2938 filed the stranded duplicate a month ago and
predicted exactly this ("a future `launchctl bootstrap` against the wrong label
could start a second daemon instance fighting over port 7878 and index locks").

## Why it recurred after #2827

There was no single definition of a service's label. Six places minted one:

| Site | What it said for trusty-search |
|---|---|
| `trusty-search::commands::service::LAUNCHD_LABEL` | `com.trusty.trusty-search` |
| `trusty-installer::commands::plist_label` | `com.trusty.trusty-search` (own table, "verified by grepping each daemon crate") |
| `trusty-search::commands::log_rotation` | `com.trusty.trusty-search.logrotate` |
| `crates/trusty-search/Makefile` | `com.bobmatnyc.trusty-search` "canonical" |
| `scripts/install-trusty-search-signed.sh` | `com.trusty.trusty-search`, calling `com.trusty.search` "a drifted label" |
| launchd, on every real host | **`com.trusty.search`** |

The script and `plist_label.rs` had both *noticed* the disagreement and resolved
it backwards — they took the Rust constant as ground truth and labelled the live
unit the alias. #2827 fixed one literal of this class; the defect came back
somewhere else, so this removes the second copy rather than correcting it.

## Canonical label set

`com.trusty.<member with its `trusty-` prefix stripped>`, sub-units suffixed.
Every LaunchAgent loaded on the host obeys it, which is what makes it the
convention rather than an invention. `canonical_label` is the rule as code and
`canonical_consts_match_the_convention` checks the table against it.

| Service | Canonical | Legacy aliases evicted on install | Was code correct? |
|---|---|---|---|
| trusty-mpm | `com.trusty.mpm` | — | yes |
| trusty-mpm supervisor | `com.trusty.mpm.supervisor` | — | yes |
| trusty-memory | `com.trusty.memory` | `com.trusty.trusty-memory` | yes |
| trusty-analyze | `com.trusty.analyze` | `com.trusty.trusty-analyze` | yes |
| **trusty-search** | **`com.trusty.search`** | `com.trusty.trusty-search`, `com.bobmatnyc.trusty-search` | **no** |
| trusty-search logrotate | `com.trusty.search.logrotate` | `com.trusty.trusty-search.logrotate` | **no** |
| **trusty-console** | **`com.trusty.console`** | `com.trusty.trusty-console` | **no** |
| trusty-review | `com.trusty.review` | `com.trusty.trusty-review` | moved onto the convention |
| trusty-agents Slack | `com.trusty.agents.slack` | — | yes |

trusty-console had the same divergence as search — `com.trusty.trusty-console` in
code, `com.trusty.console` loaded — so `console service status` queried a label
that does not exist. trusty-review's file matches its own code, but that is the
code's own output, not independent evidence; it moves onto the convention with
its old name recorded as legacy so the one exception does not reconstitute the
override table that was the bug.

**Codesign identifiers are deliberately untouched.** `macos_signing` uses the
full binary name in a different namespace; renaming one invalidates a binary's
designated requirement and re-triggers TCC prompts (#2558). The doc comment
claiming it "matches the real launchd label convention" was the reasoning that
made the two look like one thing.

## Every divergent site fixed

- `trusty-search/src/commands/service.rs` — `LAUNCHD_LABEL` → registry; install
  routed through `install_and_activate`
- `trusty-search/src/commands/log_rotation.rs` — label derived from the daemon's;
  install evicts the orphaned legacy rotation unit
- `trusty-console/src/service.rs`, `trusty-review/src/commands/service.rs`,
  `trusty-memory/…`, `trusty-analyze/…` — `LAUNCHD_LABEL` → registry
- `trusty-installer/src/commands/plist_label.rs` — the mirror table deleted,
  delegates to the registry
- `trusty-installer/src/commands/plist_bootstrap.rs` — the plist template carried
  the supervisor label as a second literal beside `PLIST_LABEL`; now a
  `__PLIST_LABEL__` placeholder filled from the constant
- `tm`'s `guided_autostart`, `launchd_probe`, `daemon_run`, `serve_stdio` — four
  literals, including two remedy hints that prescribe a unit to the operator
  (the #2827 shape)
- `crates/trusty-search/Makefile`, `crates/trusty-memory/Makefile` — the
  `com.bobmatnyc.*` family, plus the outage described below
- `scripts/install-trusty-search-signed.sh`, `crates/trusty-search/README.md`

## The one-minute outage

Three separate causes, all fixed:

1. **The old sequence bounced a daemon that had not changed.** `bootstrap()`
   unconditionally booted out first, so re-running `service install` with no
   config change cost a full stop/start plus `ThrottleInterval` (30 s).
   `install_and_activate` skips the reload entirely when the rendered plist
   matches what is installed and the label is already loaded.
2. **A failure left the service down.** The bootout happened first and the
   bootstrap could then fail, with no path back. Activation now snapshots the
   installed plist and, on failure, restores and re-bootstraps it. It also
   verifies the label actually came up rather than trusting `launchctl
   bootstrap`'s exit code — #2498 recorded a bootstrap that succeeded while the
   daemon stayed "not running".
3. **`make deploy` was the worst path.** It unloaded `com.bobmatnyc.trusty-search`
   (no such file), killed the daemon with `stop` + `pkill`, then failed to `load`
   it back and fell through to `trusty-search start` — CLI-detached and
   unsupervised, after a gap spanning the whole `cargo install`. Both Makefiles
   now defer to `service install`.

## Does #4868's plist fix now reach the live unit?

**Yes.** `trusty-search service install` writes
`~/Library/LaunchAgents/com.trusty.search.plist` — the file that is loaded — and
that plist carries `ExitTimeOut`, pinned by
`installed_unit_carries_the_exit_timeout_fix`. The install additionally boots out
and deletes `com.trusty.trusty-search` and the orphaned
`com.trusty.trusty-search.logrotate`, so #2938 closes as a side effect.

**No live daemon was touched and `launchctl` was never invoked against one.**
The only host interaction was reading `launchctl list` and the plist files. The
PM sequences the reinstall.

## Regression tests, demonstrated by reverting

The pre-fix literal was put back in `trusty-search/src/commands/service.rs`, both
guards run, and the fix restored.

```
$ cargo test -p trusty-common no_stray_launchd_label
test launchd_labels::tests::no_stray_launchd_label_literals_in_workspace_sources ... FAILED
launchd label literals not owned by `trusty_common::launchd_labels` (#4868 — derive them from the registry instead of restating them):
  crates/trusty-search/src/commands/service.rs: com.trusty.trusty-search
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 275 filtered out

$ cargo test -p trusty-search build_launchd_config_targets
test commands::service::tests::build_launchd_config_targets_the_label_launchd_has_loaded ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 367 filtered out
```

The scan rejects **any** `com.trusty.*` / `com.bobmatnyc.*` literal in production
source outside the registry, including one that is correct today — a
correct-but-duplicated literal is the state trusty-search's label was in before
it drifted. It runs on Linux CI because `launchd_labels` is deliberately not
macOS-gated, unlike `launchd` itself. Two heuristics were tightened after they
passed vacuously: accepting legacy aliases as "known", and splitting on a literal
`#[cfg(test)]`, which missed trusty-console's `#[cfg(all(test, target_os = "macos"))]`.

## Gates

```
cargo test -p trusty-common -p trusty-installer -p trusty-console \
           -p trusty-review -p trusty-analyze -p trusty-search        EXIT=0
  test result: ok. 442 passed; 0 failed    (trusty-common lib)
  test result: ok. 551 passed; 0 failed    (trusty-installer lib)
  test result: ok. 1476 passed; 0 failed   (trusty-search lib)
  test result: ok. 1042 passed; 0 failed; 23 ignored
  + 50 further binaries/integration targets, all ok, 0 failed

cargo test -p trusty-mpm --lib
  test result: ok. 4602 passed; 0 failed; 7 ignored; finished in 54.83s
cargo test -p trusty-mpm --bin tm
  test result: ok. 1438 passed; 0 failed; 1 ignored; finished in 5.18s

cargo clippy -p trusty-common -p trusty-installer -p trusty-search -p trusty-console \
             -p trusty-review -p trusty-memory -p trusty-analyze -p trusty-mpm \
             --all-targets -- -D warnings                              EXIT=0
cargo fmt --check                                                      EXIT=0
scripts/check_line_cap.sh          3688 files, 7 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh     21528 citations, 0 dangling — OK
scripts/check_sld.sh               55 specs + 3092 code files, 0 errors, 0 warnings
scripts/check_changelog_fragment.sh  all 8 crates with source changes recorded
```

No coverage was weakened: no `#[ignore]`, no `cfg`-gating, no `--exclude`.
`trusty-search/src/commands/service.rs` crossed the 500-SLOC cap once the new
tests landed, so its test module moved to a sibling `service_tests.rs` — the
`plist_bootstrap.rs` / `plist_bootstrap_tests.rs` split this repo already uses.
No test was dropped.

**Two pre-existing failures, both reproduced on unmodified `origin/main` in this
worktree and unrelated to this change:**

- `trusty-memory`'s `check_daemon_health_fails_cleanly_with_stale_addr_and_no_listener`
  and `…_when_no_addr_file_and_no_listener` return `Warn`, a status outside their
  allowed set, because a live trusty-memory daemon is running on this host. They
  fail identically with this branch stashed.
- `trusty-code-gui` and `trusty-mpm-gui` fail `cargo check --workspace` with
  `frontendDist … "ui/dist" but this path doesn't exist` — a Tauri proc-macro
  needing a `pnpm build` artifact absent in a fresh worktree.

## Operator note

**An already-installed host needs one `trusty-search service install` to pick
this up.** The label lives in the plist filename, not the binary, so upgrading
the binary alone leaves the old unit registered. That install is now the safe
operation: it evicts the legacy labels, and rolls back rather than leaving search
down if the new unit fails to load.

Refs #4868, #2938, #2827, #4059


bobmatnyc:
Review BLOCK addressed — all seven required changes plus the six MEDIUM/LOW items. Head is `c1129a97` (rebased onto `b98fb77c`; `trusty-common/Cargo.toml` conflicted with main's 0.28.2 bump, resolved to 0.29.0).

**1–2 (CRITICAL, the missing migration).** trusty-console and trusty-review now route `service_install` through `install_and_activate(legacy_labels_for(LAUNCHD_LABEL))`. You were right that these were the only two whose label value changes and the only two without eviction — that was the whole migration story and it was absent.

**3 (HIGH, the guard hole).** `production_lines` no longer uses `take_while`; it skips the test ITEM and continues — a bare `mod tests;` costs one line, a `mod tests { … }` block is skipped by brace balance. Also fixed a second-order hole the rewrite exposed: a substring match on `test` fires on `#[cfg(feature = "embedder-test-support")]`, which gates production code, so `is_test_cfg_attribute` now requires a standalone token. Your control, re-run:

```
$ # literal planted at line 301 of crates/trusty-mpm/src/services/discoverer/mod.rs
$ cargo test -p trusty-common no_stray_launchd_label
launchd label literals not owned by `trusty_common::launchd_labels`:
  crates/trusty-mpm/src/services/discoverer/mod.rs: com.trusty.trusty-search
test result: FAILED. 0 passed; 1 failed; 280 filtered out
```

**4 (HIGH, guard coverage).** The walk now covers `Makefile`, `*.sh`, `*.plist` and `*.yml` repo-wide, with per-kind comment stripping. Rather than exempt them, the rule splits by capability: Rust has the registry so ANY literal fails; a build file cannot import a Rust constant, so it may name the canonical label but never a legacy one. Codesign identifiers are skipped by `--identifier`/`*_IDENTIFIER` on the line, which keeps the real launchd label in `install-trusty-mpm-signed.sh` still guarded. Second control:

```
$ # com.bobmatnyc.trusty-search restored in crates/trusty-search/Makefile
  crates/trusty-search/Makefile: com.bobmatnyc.trusty-search
test result: FAILED. 0 passed; 1 failed; 280 filtered out
```

**5 (HIGH, deploy).** Added `service install --force` (threaded through `reload_needed`), and both `deploy` targets now `launchctl bootout` before `cargo install` so `KeepAlive::Always` cannot relaunch the daemon into a binary swap — thank you for catching #87, I had reintroduced it three lines above its own warning. trusty-memory needs no `--force`: its `service install` reloads unconditionally.

**6 (HIGH, tunables).** `installed_unit` falls back to the legacy plists when the canonical one is absent, before eviction runs. Extracted `installed_unit_paths` so the ordering is testable without a real `~/Library/LaunchAgents`.

**7 (HIGH, version).** trusty-common 0.28.1 → **0.29.0**; workspace pin and trusty-search / trusty-memory / trusty-gworkspace requirements raised to `0.29`.

**MEDIUM/LOW, all folded in.** Rollback: liveness captured before `install()`, the decision extracted to a pure `rollback_plan(has_previous, was_loaded)` with three tests, and a `Rollback` outcome so a failed restore says the service is DOWN instead of claiming preservation. `is_canonical_label` now has a real caller (the non-Rust branch of the scan) and its doc says so. The "every LaunchAgent obeys" claim is corrected to what the survey supports — every unit with a live pid — naming logrotate and review as the two exceptions being normalised. `plist_label_for` consults `SERVICES` first (`trusty-agents` → `com.trusty.agents.slack`), with a test. The three docs lines now say `trusty-search service install`, and the uninstall snippet unloads before deleting. `installed_unit_carries_the_exit_timeout_fix` renamed to `the_unit_install_activates_carries_the_exit_timeout_fix` and now asserts label + plist filename + `ExitTimeOut` in one document, which is the load-bearing claim rather than a duplicate of trusty-common's test.

**Gates, post-rebase**

```
cargo test -p trusty-common -p trusty-installer -p trusty-console \
           -p trusty-review -p trusty-analyze -p trusty-search        EXIT=0
  test result: ok. 442 passed; 0 failed      (trusty-common lib)
  test result: ok. 552 passed; 0 failed      (trusty-installer lib)
  test result: ok. 1476 passed; 0 failed     (trusty-search lib)
  test result: ok. 1566 passed; 0 failed
  test result: ok. 1050 passed; 0 failed; 23 ignored
cargo clippy (8 crates) --all-targets -- -D warnings                  EXIT=0
cargo fmt --check                                                     EXIT=0
check_line_cap.sh        3695 files, 7 allowlisted, 0 violations — OK
check_test_pointers.sh   21573 citations, 0 dangling — OK
check_sld.sh             55 specs + 3092 files, 0 errors, 0 warnings
check_changelog_fragment.sh  all 8 crates recorded
```

**One flake worth flagging rather than burying.** `cargo test -p trusty-mpm --lib` failed twice mid-session on `ensure_managed_config_dir_emits_the_frozen_skill_warning` and `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call`. Both pass in isolation, and a second full run on the identical tree passed 4602/0 — so it is a parallel-execution flake, not this branch. I did not touch either path. Pre-existing and unrelated, both re-confirmed by stashing onto the base: trusty-memory's two `check_daemon_health_*` tests (return `Warn` because a live memory daemon runs on this host) and the two Tauri GUI crates (`frontendDist … "ui/dist" but this path doesn't exist`).

No launchd write operation was run against the live host — no bootstrap, bootout, load, unload, or `service install`. The only host interaction remains reading `launchctl list` and the plist files.

Issue #3605, opened 2026-07-21
tm doctor: detect a live session's base clone losing git identity (bare .base corruption undetected until agents die)


bobmatnyc:
## Gap

Nothing in `tm doctor` detects that a live session's `base_dir` (the per-repo
bare clone that backs `git worktree add` checkouts) has lost its git identity
while worktrees still reference it. Today this class of corruption was only
discovered because two agents died with `fatal: not a git repository: (null)`
— roughly 30+ minutes after the underlying failure actually happened.

## Incident (2026-07-21, investigated and verified today)

- `/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.base` is a **bare
  git clone** created by `RealGitBackend::ensure_base_checkout`
  (`crates/trusty-mpm/src/provisioner/workspace.rs:375-441`,
  `BASE_CHECKOUT_DIRNAME` at `workspace.rs:66`), serving per-session
  `git worktree add` checkouts.
- At approximately 14:01 its non-dot internals — `HEAD`, `config`,
  `objects/`, `refs/`, `packed-refs`, and the `worktrees/` admin directory —
  were all deleted. Only dot-prefixed entries (`.claude/`, `.worktrees/`)
  survived. The signature matches an `rm -rf .base/*` under default shell
  globbing, which never matches dot-prefixed entries.
- **70 of 77 worktrees were orphaned.** Their `.git` files point at
  `.base/worktrees/<name>` (dot-less), now missing.
- Because a bare `.base` has its own object database separate from the
  parent repo's `.git/objects/`, commits that lived only there and were
  never pushed are unrecoverable at the git level.
- **This was NOT caused by trusty-mpm code.** Every plausible code path was
  checked and cleared:
  - `inproject_hygiene.rs:291-392` only touches the separate `repos_root`
    convention and gates on `.git` existing — a bare repo has none, so it
    always no-op'd here.
  - `session_manager/prune.rs:272-317` and `decommission.rs:178-223` are
    scoped to single named worktrees under `repos_root`.
  - `workspace_scan.rs` is read-only.

  Flagging this explicitly so nobody chases a phantom product bug in the
  prune machinery — the destructive action was external to trusty-mpm.

- Notably, `provisioner/workspace/base_lock.rs:87-110`
  (`stale_base_dir_error`) already reasons about a `.base` failing its
  bare-checkout probe, and **deliberately refuses to auto-delete**, with a
[...]

Issue #2184, opened 2026-07-07
Per-project GitHub identity binding in trusty-mpm (gh account + commit identity)


bobmatnyc:
## Problem / Motivation

On a multi-account machine, the wrong `gh` account being active breaks project operations: e.g. `bobmatnyc/trusty-tools` requires the `bobmatnyc` gh account for admin-merge, but the `bob-duetto` (Duetto work) account is also logged in and, when active, admin ops fail. Today the correct account has to be selected globally/manually and drifts.

**The correct GitHub identity for a project should be a per-project configuration setting that tm applies automatically.**

## Current State

- trusty-mpm already has a `GithubConfig` struct (issue #1265) — `config_dir`/`token_env`/`account`/`host`, applied via `GH_CONFIG_DIR`/`GH_TOKEN`/`GH_HOST` env overrides — at `crates/trusty-mpm/src/core/trusty_tools_config.rs:184-219`. It deliberately AVOIDS `gh auth switch` (which mutates global state and would conflict with other-account work on the same box); account-alone bindings are refused (`gh_identity.rs:71-82`) in favor of `config_dir`.

- BUT it is a SINGLE GLOBAL binding (`TrustyToolsConfig.github`, `trusty_tools_config.rs:118`), not keyed per project, and is wired into only 3 CLI commands (`tm issue`/`ticket`/`watch` via `load_gh_env()` at `bin/tm/gh_identity.rs:242`). It is NOT applied to the daemon managed-session spawn (`daemon/managed_routes/lifecycle.rs:85 spawn_managed`) or the provisioner's `git clone`/`fetch`/`reset` subprocesses (`provisioner/workspace.rs` — plain `Command::new("git")`, ambient identity).

- tm also NEVER sets git commit identity (user.email/user.name) anywhere.

## Scope of This Feature (All Four Required)

1. **Per-project gh binding**: add `github: Option<GithubConfig>` to `ProjectConfig` (`core/trusty_tools_config.rs:265`) and mirror onto the runtime `Project` record (`project/record.rs:22`), threading it through `ProjectRegistry::seed_from_config` (`project/registry.rs:98-118`). Resolution precedence: **per-project `github` > global `github` > ambient**. Reuse the existing `GithubConfig` + `resolve_gh_env` machinery.

[...]

Issue #1803, opened 2026-06-29
fix(trusty-mpm): unify managed-clone root — daily `tm` and `tm launch` diverge to two roots; banner misreports session path


bobmatnyc:
## Problem

`tm` (no-subcommand guided default) and `tm launch` use TWO competing managed-clone subsystems that send sessions to different filesystem roots, with no cross-root dedup or reconnect:

| Entry | Routes through | Real session dir |
|---|---|---|
| `tm` (daily driver) | picker → daemon POST `repo_url=<local git_root>` → `try_inproject_spawn` → `spawn_managed_inproject` | `~/trusty-tools/repos/<owner>/<repo>/worktrees/<session-id>/` (inproject, #1706) |
| `tm launch` | in-process `WorkspaceProvisioner::provision_in(workspace_subpath(...))` | `~/trusty-mpm-projects/<owner>/<repo>/<uuid>/` (#1590/#1220) |

Same repo, two roots, depending on which command you type. Discovered while verifying #1796 (#1590) and #1794.

## Decision (owner) — CORRECTED

Canonical managed model = **one shared clone per repo + a git worktree per session**, rooted at the canonical base **`~/trusty-mpm-projects/<owner>/<repo>/`**. Each session is a git worktree off that single base clone (e.g. `~/trusty-mpm-projects/<owner>/<repo>/worktrees/<session-id>/`), NOT a full separate clone per session.

This is effectively the **#1706 inproject *mechanism* (shared base clone + per-session worktrees), re-rooted** from `~/trusty-tools/repos/` to the canonical `~/trusty-mpm-projects/` base — plus making `tm launch` use that same mechanism so both entry points converge. The full-clone-per-UUID model (`WorkspaceProvisioner::provision_in` used by `tm launch`/`spawn_managed_local`) is superseded for the local-launch path.

Rationale: one clone per repo (cheap, shared git object store), concurrent sessions isolated via worktrees, single canonical root, navigable layout.

## Fixes

[...]

Issue #1937, opened 2026-07-03
Follow-ups from #1936 review: stale base-checkout recovery, lock-staleness edge case, FakeGitBackend fidelity


bobmatnyc:
## Context

PR #1936 (closed #1935) introduced a shared bare git checkout per project plus per-session `git worktree add` provisioning, replacing a full clone per session. `trusty-review` approved the PR (APPROVE*, grade C+) but flagged 3 non-blocking advisory items worth addressing in a follow-up:

## 1. No automatic remediation when a base checkout is a stale/broken non-bare directory

`ensure_base_checkout` in `crates/trusty-mpm/src/provisioner/workspace.rs` correctly rejects (fails loudly) if `<project_dir>/.base` exists but isn't a valid bare checkout (e.g. left over from a crashed mid-clone, or a stale non-bare directory). However there's no recovery path — the operator must manually delete the directory before the project can be provisioned again. Consider: detecting this state and either auto-remediating (remove + re-clone) or surfacing a clear, actionable error message with the exact path to delete.

## 2. Theoretical lock-staleness race in `acquire_base_checkout_lock`

`crates/trusty-mpm/src/provisioner/workspace/base_lock.rs`: the advisory lock uses a `LOCK_STALE_AFTER` (5 minute) mtime threshold to recover from a crashed lock holder. On a system with low-resolution filesystem timestamps or significant clock skew, it's theoretically possible for a second caller to force-remove a still-live holder's lock file, allowing two callers to hold the lock simultaneously. This is a rare edge case (the design is a reasonable pragmatic tradeoff) but the race window should be documented explicitly in code, and/or consider a more robust mechanism (e.g. PID + liveness check) if it proves to matter in practice.

## 3. `FakeGitBackend::ensure_base_checkout` uses a different (weaker) idempotency check than `RealGitBackend`

`crates/trusty-mpm/src/provisioner/workspace.rs`: `RealGitBackend::ensure_base_checkout` was fixed to use `git rev-parse --is-bare-repository` for robust bare-repo detection (replacing a fragile `HEAD`-file-exists check). `FakeGitBackend`'s test double still uses the old `HEAD`-file-exists pattern. This is benign today (the one test that exercises stale-directory rejection correctly uses `RealGitBackend`, not the fake), but a future test author using `FakeGitBackend` to simulate a stale-directory scenario could get a false-positive pass. Consider aligning `FakeGitBackend`'s idempotency check with the real backend's semantics for test fidelity.

[...]

PR #4926, opened 2026-08-05
fix(trusty-mpm): provision new worktrees at <project-root>/.worktrees (#4270)

### LINKED ISSUE #4270: converge in-project provisioning on the .base store (supersedes: retire the path)

bobmatnyc:
## What changed

New sessions provisioned from a repo URL now get their worktree at
`<project-root>/.worktrees/<name>`, the git convention. They used to land inside
a `.base` bare clone at `<project-root>/.base/.worktrees/<id>`.

The base checkout moved with them: it is now the project directory's own clone —
the same clone `inproject::ensure_base_clone` establishes. A repo spawned from a
URL and the same repo spawned from a local checkout now share one base clone per
project instead of keeping two stores in one project directory. That is the
common-entry-point rule applied to the thing the two paths were duplicating.

Nothing writes `.base` any more.

## Map — corrected against the code

The dispatch map in the brief was right about the files. One correction and one
addition:

| Brief said | Code says |
|---|---|
| Make `spawn_managed_cloned` reuse the in-project provisioning | Not directly callable. `try_inproject_spawn` needs a LOCAL checkout to detect a remote from; a URL spawn has no local path. The reuse that IS available is the topology, so the provisioner now establishes the same base clone at the same place. Calling `ensure_base_clone` itself would have dropped the per-project `GitIdentity` (#2184) that `RealGitBackend` applies to every git subprocess — private-repo clone auth. |
| `harness_root.rs:52-60` doc is stale | Confirmed, corrected. `BASE_CLONE_DIRNAME` is now `pub(crate)` — one literal for `.base`, shared by the reader that maps a `.base` common-dir back to its project and the two writers that refuse to clobber one. |
| — | `inproject::migrate_old_layout_aside` needed a guard. A project directory holding a `.base` store matches the pre-#1803 old-layout signature exactly (non-empty, no top-level `.git`), so it renamed the directory to `<repo>.old-layout-backup-<ts>` — taking `.base` and every live worktree with it. This is requirement 3's actual mechanism; see the fail-before output below. |

`lifecycle.rs:626-636` needed only a comment correction: the dispatch is unchanged,
`provision_in` does the work.

## Out of scope, as instructed

No migration, no compatibility layer, no `.base` sweep, no config toggle. Existing
`.base` worktrees are git-native — `git worktree list` finds them, registry entries
hold absolute paths, and `harness_root`'s `.base`→project mapping is untouched — so
they keep working with no code supporting them. Cleanup stays manual and the
owner's.

Two guards exist only because the new topology makes both spawn paths meet a
project directory that holds a `.base`. Both refuse; neither moves anything.
The refusal message deliberately carries no `mv` hint: the generic quarantine hint
aimed at a project directory would rename every worktree under the store out from
under its session — the #3605 shape, one level up.

## Tests — fail before, pass after

Requirements 1 and 2 (`provisioner/workspace/tests.rs`), requirement 3 twice —
once per code path that could disturb a live store.

Fail-before, with the path computation temporarily reverted:

```
test provisioner::workspace::tests::provision_in_creates_no_dot_base_directory ... FAILED
test provisioner::workspace::tests::provision_in_leaves_an_existing_dot_base_store_untouched ... FAILED
test provisioner::workspace::tests::provision_in_puts_worktrees_under_the_project_root_not_dot_base ... FAILED
test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 4610 filtered out
```

Fail-before for the migrator guard, with the guard removed — this is the data-loss
shape stated outright:

```
test daemon::managed_routes::inproject::tests::migrate_old_layout_aside_refuses_a_dir_holding_a_dot_base_store ... FAILED
assertion failed: a dir holding a .base store must be refused, not migrated:
  Some("/tmp/tm-test-dy17uI/owner/repo.old-layout-backup-1785941128950735000")
```

`tests/local_spawn.rs::spawn_managed_local_redirects_to_managed_clone` pinned the
old `.base` path and was updated to the new one — it now produces exactly the
target shape.

## Gates — Rust Test Ladder rung 5 (process lifecycle / session provisioning)

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — 4606 passed, 0 failed, 7 ignored (lib) + every integration target green |
| `bash scripts/check_line_cap.sh` | EXIT=0 — 3688 files, 0 violations |
| `bash scripts/check_changelog_fragment.sh` | EXIT=0 — scanned 9 paths |
| `cargo check --workspace` | EXIT=101 — known baseline: Tauri `frontendDist` proc-macro panic in `trusty-code-gui` and `trusty-mpm-gui`. Neither crate is touched by this diff. |

`--test-threads=1` per the documented `$HOME`-race baseline
(`docs/reference/test-ladder-baseline.md`).

`workspace.rs` hit 502 SLOC against the 500 cap. Fixed by folding the legacy-store
guard into `stale_base_dir_error` — the single "is this path safe to clone into?"
entry point — which removed the duplicated call at both backends rather than
allowlisting the file.

## Not done, deliberately

Issue #4270's table also lists routing `tm launch` and the guided default through
`spawn_managed_inproject`. Both already call `ensure_base_clone` and so already
produce `<project-root>/.worktrees/<name>`; routing them differently changes no
topology. Left out under the scope constraint.

A URL-provisioned project's `.worktrees/` is not added to the base clone's
`.git/info/exclude`, so it shows as untracked in `git status` run inside the base
clone. The in-project path does add it, and any project that has ever used that
path is already covered. Cosmetic, and the fix means moving
`ensure_worktrees_gitignored` out of the daemon-gated module.

Closes #4270

### DIFF
--- a/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
+++ b/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
@@ -3,3 +3,4 @@ Changed
 - new sessions provisioned from a repo URL now get their worktree at `<project-root>/.worktrees/<name>`, following the git convention, instead of nesting it inside a `.base` bare clone (closes [#4270](https://github.com/bobmatnyc/trusty-tools/issues/4270))
   - the base checkout is now the project directory's own clone, the same one the in-project spawn path establishes, so both paths share one base clone per project
   - existing `.base` stores are left exactly as they are — still discoverable via `git worktree list`, still working — and provisioning refuses loudly rather than moving or deleting one; retiring `.base` stays a manual operator step
+  - a project directory holding git or trusty-mpm state (`.git`, `.base`, or `.worktrees`) is never offered for `mv`-aside recovery and never renamed by the old-layout migrator; the presence probe treats an unreadable entry as present, so a permissions blip cannot read as "no git dir"
--- a/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
+++ b/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
@@ -1,0 +1,5 @@
+Changed
+
+- new sessions provisioned from a repo URL now get their worktree at `<project-root>/.worktrees/<name>`, following the git convention, instead of nesting it inside a `.base` bare clone (closes [#4270](https://github.com/bobmatnyc/trusty-tools/issues/4270))
+  - the base checkout is now the project directory's own clone, the same one the in-project spawn path establishes, so both paths share one base clone per project
+  - existing `.base` stores are left exactly as they are — still discoverable via `git worktree list`, still working — and provisioning refuses loudly rather than moving or deleting one; retiring `.base` stays a manual operator step
--- a/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
+++ b/crates/trusty-mpm/changelog.d/4270-provision-worktrees.md
@@ -4,3 +4,4 @@ Changed
   - the base checkout is now the project directory's own clone, the same one the in-project spawn path establishes, so both paths share one base clone per project
   - existing `.base` stores are left exactly as they are — still discoverable via `git worktree list`, still working — and provisioning refuses loudly rather than moving or deleting one; retiring `.base` stays a manual operator step
   - a project directory holding git or trusty-mpm state (`.git`, `.base`, or `.worktrees`) is never offered for `mv`-aside recovery and never renamed by the old-layout migrator; the presence probe treats an unreadable entry as present, so a permissions blip cannot read as "no git dir"
+  - the base clone's `.worktrees/` is excluded in `.git/info/exclude`, as the in-project path already does — without it `git clean -ffd` in the base deletes every session worktree including uncommitted work (single-force `clean` skips them; `-ff` does not)
--- a/crates/trusty-mpm/src/core/harness_root.rs
+++ b/crates/trusty-mpm/src/core/harness_root.rs
@@ -65,6 +65,44 @@ pub const SESSIONS_DIR: &str = "sessions";
 /// `provision_in_leaves_an_existing_dot_base_store_untouched`.
 pub(crate) const BASE_CLONE_DIRNAME: &str = ".base";
 
+/// Name the protected state `dir` holds, or `None` when it holds none.
+///
+/// Why (#4270): two call sites decide whether a directory may be renamed aside
+/// — `provisioner::workspace`'s stale-base recovery hint and
+/// `daemon::managed_routes::inproject::migrate_old_layout_aside`. Both used to
+/// ask a narrower question (is there a `.base`? is there a `.git`?) and both
+/// were wrong in the same way: the thing that must never be moved is any
+/// directory git or trusty-mpm keeps real state in. Renaming one orphans every
+/// worktree beneath it, which is the #3605 failure with the project directory
+/// as its target. One predicate means the two guards cannot drift apart.
+///
+/// The probe is deliberately fail-SAFE rather than accurate. `Path::exists`
+/// answers `false` for a directory it lacks permission to stat, so an EACCES
+/// blip on `.git` reads exactly like a project with no git directory at all —
+/// and the caller then renames live work. Anything that is not a definite
+/// `NotFound` therefore counts as present: the cost of a false positive is a
+/// refusal the operator can act on, the cost of a false negative is lost work.
+/// What: returns the first of `.git`, `.base`, or `.worktrees` found under
+/// `dir`, so the caller's message can name what it found; `None` only when all
+/// three are definitely absent.
+/// Test: `protected_state_in_names_each_protected_entry`,
+/// `protected_state_in_is_none_for_a_foreign_directory`,
+/// `protected_state_in_treats_an_unreadable_entry_as_present`.
+pub(crate) fn protected_state_in(dir: &Path) -> Option<&'static str> {
+    [
+        ".git",
+        BASE_CLONE_DIRNAME,
+        crate::session_manager::decommission::WORKTREES_DIRNAME,
+    ]
+    .into_iter()
+    .find(|name| {
+        !matches!(
+            std::fs::symlink_metadata(dir.join(name)),
+            Err(ref e) if e.kind() == std::io::ErrorKind::NotFound
+        )
+    })
+}
+
 /// Environment variable carrying the managed session id inside a tm pane.
 const MANAGED_SESSION_ID_ENV: &str = "TM_MANAGED_SESSION_ID";
 
--- a/crates/trusty-mpm/src/core/harness_root.rs
+++ b/crates/trusty-mpm/src/core/harness_root.rs
@@ -44,20 +44,26 @@ pub const FRAMEWORK_DIR: &str = "framework";
 /// Test: `session_dir_is_per_session_under_the_harness_root`.
 pub const SESSIONS_DIR: &str = "sessions";
 
-/// The shared bare clone trusty-mpm provisions inside a managed project.
+/// The shared bare clone trusty-mpm provisioned inside a managed project
+/// before #4270.
 ///
-/// Why: `provisioner::workspace` clones the shared base into `<project>/.base`
-/// and adds every managed session's worktree from THERE, so those worktrees'
-/// git common dir is the bare clone — a directory INSIDE the project, not the
-/// project. This is not a path guess: it is the same directory name
-/// `provisioner::workspace::BASE_CHECKOUT_DIRNAME` writes and
-/// `session_manager::worktree_registry` interrogates.
+/// Why: `provisioner::workspace` used to clone the shared base into
+/// `<project>/.base` and add every managed session's worktree from THERE, so
+/// those worktrees' git common dir is the bare clone — a directory INSIDE the
+/// project, not the project. #4270 retired that store: provisioning now clones
+/// the base into `<project>` itself and puts worktrees at
+/// `<project>/.worktrees/<id>`, matching the in-project path. Existing `.base`
+/// stores were deliberately NOT migrated, so this mapping stays load-bearing
+/// for every worktree already under one, and the name is also what
+/// `provisioner::workspace`'s own guard checks before refusing to clone over a
+/// live legacy store.
 /// What: `.base`. The name alone is never sufficient — see
 /// [`map_base_clone_to_project`], which also requires the directory to be a
 /// BARE repository before rewriting it.
 /// Test: `harness_root_maps_a_base_clone_back_to_the_project`,
-/// `harness_root_for_a_non_bare_repo_named_base_is_itself`.
-const BASE_CLONE_DIRNAME: &str = ".base";
+/// `harness_root_for_a_non_bare_repo_named_base_is_itself`,
+/// `provision_in_leaves_an_existing_dot_base_store_untouched`.
+pub(crate) const BASE_CLONE_DIRNAME: &str = ".base";
 
 /// Environment variable carrying the managed session id inside a tm pane.
 const MANAGED_SESSION_ID_ENV: &str = "TM_MANAGED_SESSION_ID";
--- a/crates/trusty-mpm/src/core/harness_root_tests.rs
+++ b/crates/trusty-mpm/src/core/harness_root_tests.rs
@@ -333,3 +333,88 @@ fn session_scope_rejects_a_path_traversal_segment() {
         );
     }
 }
+
+/// #4270: the predicate names each protected entry it finds.
+///
+/// Why: two guards refuse a rename based on this answer, and each wants to say
+/// what it found. A predicate that recognised only one of the three names is
+/// exactly the hole the first round of #4270 shipped — `.base` was guarded and
+/// `.worktrees` was not.
+/// What: asserts each of `.git`, `.base`, `.worktrees` is detected on its own.
+/// Test: this function IS the test.
+#[test]
+fn protected_state_in_names_each_protected_entry() {
+    for name in [".git", ".base", ".worktrees"] {
+        let tmp = crate::test_support::hermetic_temp_dir();
+        let dir = tmp.path().join("owner").join("repo");
+        std::fs::create_dir_all(dir.join(name)).expect("create protected entry");
+        assert_eq!(
+            protected_state_in(&dir),
+            Some(name),
+            "{name} must be recognised as protected state"
+        );
+    }
+}
+
+/// #4270: a directory carrying none of the protected entries is not protected.
+///
+/// Why: the guards must still let genuine foreign debris through to the
+/// quarantine path, or the stale-clone recovery #1937 added stops working.
+/// What: a non-empty directory holding only a stray `HEAD` — the crashed-clone
+/// shape — reads as unprotected.
+/// Test: this function IS the test.
+#[test]
+fn protected_state_in_is_none_for_a_foreign_directory() {
+    let tmp = crate::test_support::hermetic_temp_dir();
+    let dir = tmp.path().join("owner").join("repo");
+    std::fs::create_dir_all(&dir).expect("create dir");
+    std::fs::write(dir.join("HEAD"), "ref: refs/heads/main\n").expect("write HEAD");
+    assert_eq!(protected_state_in(&dir), None);
+}
+
+/// #4270: an entry that cannot be stat'd counts as PRESENT, not absent.
+///
+/// Why: this is the whole reason the probe is not `Path::exists`, which maps
+/// EACCES to `false` — indistinguishable from "no git dir". Under that
+/// spelling a permissions blip on `.git` lets `migrate_old_layout_aside` rename
+/// a live project directory. The failure is silent and the data is gone, so the
+/// fail-safe direction has to be proven, not assumed.
+/// What: chmods the containing directory to 000 so stat'ing a child returns
+/// EACCES, then asserts the probe still reports protected. Skips when running
+/// as root (root bypasses the permission check) or when the platform lets the
+/// stat through anyway.
+/// Test: this function IS the test.
+#[cfg(unix)]
+#[test]
+fn protected_state_in_treats_an_unreadable_entry_as_present() {
+    use std::os::unix::fs::PermissionsExt as _;
+
+    let tmp = crate::test_support::hermetic_temp_dir();
+    let dir = tmp.path().join("owner").join("repo");
+    std::fs::create_dir_all(dir.join(".git")).expect("create .git");
+
+    let restore = std::fs::metadata(&dir).expect("stat dir").permissions();
+    std::fs::set_permissions(&dir, std::fs::Permissions::from_mode(0o000)).expect("chmod 000");
+
+    let unreadable = std::fs::symlink_metadata(dir.join(".git"))
+        .err()
+        .is_some_and(|e| e.kind() != std::io::ErrorKind::NotFound);
+    let observed = protected_state_in(&dir);
+
+    // Restore before asserting so a failure cannot leave an undeletable tempdir.
+    std::fs::set_permissions(&dir, restore).expect("restore permissions");
+
+    if !unreadable {
+        eprintln!(
+            "protected_state_in_treats_an_unreadable_entry_as_present: stat still \
+             succeeds (running as root?), skipping"
+        );
+        return;
+    }
+    assert_eq!(
+        observed,
+        Some(".git"),
+        "an entry that cannot be stat'd must count as present — Path::exists would \
+         report false here and the caller would rename live work"
+    );
+}
--- a/crates/trusty-mpm/src/core/worktree_naming.rs
+++ b/crates/trusty-mpm/src/core/worktree_naming.rs
@@ -1,20 +1,25 @@
-//! Shared per-session git-worktree branch naming convention (issue #2032).
+//! Shared per-session git-worktree conventions (issue #2032, extended #4270).
 //!
-//! Why: both `daemon::managed_routes::inproject` (feature = `"daemon"`;
-//! creates per-session worktrees/branches) and `session_manager::decommission`
-//! (unconditional — compiled with or without the `daemon` feature; removes
-//! them) must agree on EXACTLY the same `session/<name>` branch convention.
-//! Placing the convention here, in `core` (always compiled, no feature gate),
-//! lets the unconditional `session_manager` module depend on it WITHOUT
-//! pulling in the feature-gated `daemon` module — the reverse dependency
-//! direction is what caused the bug this module fixes: `decommission.rs`
-//! previously hardcoded `git branch -D <leaf>` (missing the `session/` prefix
-//! `create_session_worktree` actually uses), so the branch-delete step always
-//! targeted a nonexistent ref and silently no-opped — every session's branch
-//! leaked forever.
+//! Why: modules on BOTH sides of the `daemon` feature gate create and destroy
+//! per-session worktrees, and every one of them must agree on the same
+//! conventions. `daemon::managed_routes::inproject` (feature = `"daemon"`)
+//! creates them, `session_manager::decommission` (unconditional) removes them,
+//! and since #4270 `provisioner::workspace` (also unconditional) creates the
+//! identical topology from the clone-from-URL path. Placing the conventions
+//! here, in `core` (always compiled, no feature gate), lets the unconditional
+//! modules depend on them WITHOUT pulling in the feature-gated `daemon` module.
+//! The reverse dependency direction is what caused the bug this module was
+//! created to fix: `decommission.rs` hardcoded `git branch -D <leaf>`, missing
+//! the `session/` prefix `create_session_worktree` actually uses, so the
+//! branch-delete step targeted a nonexistent ref and silently no-opped — every
+//! session's branch leaked forever.
 //! What: [`worktree_branch_for`] builds the `session/<name>` branch name for a
-//! given worktree leaf/session name.
-//! Test: `worktree_branch_for_adds_session_prefix`.
+//! given worktree leaf/session name; [`ensure_worktrees_gitignored`] keeps the
+//! `.worktrees/` directory out of the base clone's `git status` and out of
+//! reach of `git clean -ffd`.
+//! Test: `worktree_branch_for_adds_session_prefix`,
+//! `ensure_worktrees_gitignored_idempotent`,
+//! `worktrees_exclude_entry_protects_against_double_force_clean`.
 
 /// Compute the per-session branch name for a worktree leaf/session name.
 ///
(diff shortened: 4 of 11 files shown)


bobmatnyc:
## code-critic — Verdict: WARN

Reviewed `73eb57cb6dcaa8d40a9780e164c4aa15130a994d`. Zero CRITICAL, two HIGH.

### The data-loss question first

**No code path in this diff can rename, move, or delete an existing `.base` store or a worktree inside one.** Every filesystem-mutating call reachable from the changed code:

| Site | Call | Target |
|---|---|---|
| `inproject.rs:240` | `fs::rename` | the only relocation in the diff — guarded (below) |
| `inproject.rs:304`, `workspace.rs:362`, `:464` | `create_dir_all` | parent dirs only |
| `workspace.rs:649-650`, `:823`, `:838`, `base_lock.rs:274` | `fs::write` | new files inside a new worktree / fake fixtures |
| `base_lock.rs:350`, `:413` | `remove_file` | the advisory lock marker |

No `remove_dir_all`, no delete of anything pre-existing. `git clone` runs only into an absent or empty destination; `git worktree add` only creates.

Conditions under which `migrate_old_layout_aside` still fires, enumerated:

| Project-dir shape | Outcome |
|---|---|
| `.base`, no `.git` | **refused** (`inproject.rs:205-217`), moves nothing — covered by `migrate_old_layout_aside_refuses_a_dir_holding_a_dot_base_store` |
| `.base` **and** `.git` | `Ok(None)` at the `.git` check; `is_established_checkout` → true → base reused, `.base` never read |
| `.base` present but empty (partial) | `legacy.exists()` is true → refused. Conservative, correct |
| `.base` worktrees live vs stale | not distinguished; both refused. Correct per the ruling |
| `.worktrees/<id>` populated, **no** `.git`, no `.base` | **RENAMED** — see HIGH-2 |

The operator's machine is safe: `/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.git` exists alongside `.base/`, `.worktrees/` and `.claude/worktrees/`, so both guards short-circuit at the `.git` check before any rename or clone is considered.

The `.base` refusal message carries no `mv` and no `rm`, and states outright that trusty-mpm will not move, rename, or delete the store. It cannot induce the destructive act manually. (One dead-end sentence, LOW-4.)

### Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/provisioner/workspace/base_lock.rs` | 216-233 | The generic quarantine hint now names the **project directory**, which by design holds `.worktrees/<id>` for every live session. `legacy_base_store_error` diverts only the `.base` case. Pre-#4270 this was unreachable for an in-project-shaped project: `base_dir` was `<project>/.base`, absent → `stale_base_dir_error` returned `None` and the clone proceeded. Post-#4270 the same project yields `mv <project_dir> <project_dir>.stale-<ts>` — and per #3605 that string is executed verbatim by an agent, relocating every live worktree plus uncommitted work. Trigger is `is_established_checkout(project_dir) == false` while `.worktrees/` is populated: corrupt or unreadable `.git`, `git` missing from the daemon's PATH, dubious-ownership rejection, or a transient `Command::output()` spawn failure — `base_lock.rs:78-79` maps every one of those to `false` identically to "not a checkout". | In `stale_base_dir_error`, before emitting the `mv` hint, return the same moves-nothing refusal when `<base_dir>/.worktrees` exists and is non-empty. Same two-branch shape already written for `.base`. | Fix here |
| HIGH | `crates/trusty-mpm/src/daemon/managed_routes/inproject.rs` | 205-217 | The new guard closes `.base` and leaves the neighbouring shape open. A project dir holding `<project>/.worktrees/<id>` with `.git` absent or unreadable (`Path::exists()` returns `false` on EACCES as well as on absence, line 190) is still non-empty with no top-level `.git`, so line 240 renames the whole directory — live worktrees, uncommitted work and all — to `<repo>.old-layout-backup-<ts>`. This is the identical argument the doc comment at lines 160-166 makes for `.base`, applied to the container the PR is moving everything into. The behaviour is pre-existing, but the surrounding doc now reads as though the data-loss shape is handled, which is the "reads as safe" failure the guard was meant to prevent. | Extend the same `Err` branch: refuse when `base_path.join(".worktrees")` exists and is non-empty, with the same wording. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/provisioner/workspace/tests.rs` | 954-970 | The headline topology has no real-git coverage. All three new #4270 tests use `FakeGitBackend`, whose `worktree_add` is a `create_dir_all`. The one `RealGitBackend::worktree_add` test uses a **sibling** worktree (`scratch/wt`) under a bare `base.git` — not `<project_dir>/.worktrees/<id>` under a non-bare base. That a real `git worktree add` accepts a path nested inside its own base's working tree is asserted nowhere in the crate. (I verified by hand that it does; see Notes.) | Point that test's `base_dir` at `<scratch>/owner/repo` and its `worktree` at `base_dir.join(".worktrees").join(<id>)`. Two lines, and it converts the whole change from fake-verified to git-verified. | Fix here |
| LOW | `crates/trusty-mpm/src/provisioner/workspace/base_lock.rs` | 130-131 | The refusal's closing advice — "spawn this project's sessions from a local checkout (the in-project path), which uses the same `<project>` base clone this path wants" — is a dead end in part of the state that produces it. With `.base` present and `<project>/.git` absent, the in-project path reaches `migrate_old_layout_aside`, which now refuses too. An autonomous agent following the hint hits a second loud failure. Not destructive. | Drop the sentence, or say the in-project path works only once `<project>/.git` exists. | Fix here |

1. Close the `.worktrees` half of the `mv`-hint guard (`base_lock.rs`).
2. Close the `.worktrees` half of the migrator guard (`inproject.rs`).

### Verification I ran (critic worktree at `73eb57cb`)

```
bash scripts/check_line_cap.sh && cargo fmt --check && bash scripts/check_changelog_fragment.sh   → EXIT=0
  line-cap: measured 3600 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

cargo test -p trusty-mpm --lib -- --test-threads=1 provisioner::workspace daemon::managed_routes::inproject   → EXIT=0
  test result: ok. 68 passed; 0 failed; 0 ignored; 0 measured; 4545 filtered out; finished in 3.70s

cargo clippy -p trusty-mpm --all-targets -- -D warnings && cargo test -p trusty-mpm --test local_spawn -- --test-threads=1   → EXIT=0
  test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

No allowlist entry was added (`git diff --name-only origin/main...73eb57cb` matches nothing containing "allowlist"). No `#[ignore]`, no `#[cfg]` gate, no test deleted — `#[test]` counts go 21→24 (`workspace/tests.rs`), 13→14 (`inproject/tests.rs`), 13→13 (`local_spawn.rs`).

### Notes

**The `stale_base_dir_error` fold (question 7).** Behaviour-identical at both backends. `legacy_base_store_error` now sits behind the `non_empty` gate, and a `.base` directory's existence makes its parent non-empty by definition, so the gate can never suppress it. Both backends call the same `stale_base_dir_error`, so `FakeGitBackend` and `RealGitBackend` produce the same refusal — which is why the fake-backed requirement-3 test is meaningful evidence for the real path.

**Real-git topology, verified by hand** in a throwaway directory (never against `~/trusty-mpm-projects/`): `git clone --no-local file://origin.git owner/repo`, `git -C owner/repo fetch origin +HEAD:refs/heads/sess-1`, `git -C owner/repo worktree add owner/repo/.worktrees/sess-1 sess-1` all succeed. `git status` in the base then shows `?? .worktrees/`, and `git clean -xfdn` reports `Would skip repository .worktrees/sess-1` — git skips nested repositories, so the missing `.git/info/exclude` entry costs status noise and nothing else. Disclosed gap, acceptable.

**`is_established_checkout` (question 1, second guard).** The toplevel comparison holds. Probed directly: a directory nested inside an unrelated enclosing repo prints the enclosing root (≠ `dir`) → `false`; a bare repo exits 128 → `false`; a repo root prints itself → `true`. The one weakness is the fail-open at lines 78-79, where a spawn failure is indistinguishable from "not a checkout" — that is what makes HIGH-1 reachable on a healthy repo. Pre-existing conflation; its consequence changed when `base_dir` became the project directory.

**Both engineer corrections check out (question 5).** `try_inproject_spawn` gates on `path.join(".git").exists()` (`inproject.rs:790`), so a URL spawn with no local checkout genuinely cannot use it. And `ensure_base_clone` spawns a bare `std::process::Command::new("git")` (`inproject.rs:316`) while `RealGitBackend::command()` applies `GitIdentity`'s `-c user.*` args and env overrides (`workspace.rs:216-223`) — routing through it would have dropped the #2184 identity. Neither correction hides a simpler design.

**Convergence is real (question 6).** `base_clone_path` → `repos_root()` → delegates to `workspace_root(config)`; `workspace_subpath` → `workspace_root(config)`. Same `<root>/<owner>/<repo>`. Two pre-existing divergences survive and are worth knowing about, neither introduced here: `TRUSTY_MPM_REPOS_ROOT` set to something other than the workspace root splits the two paths onto different roots (`inproject.rs:82-90`), and `spawn_managed_cloned` falls back to `provisioner.provision()` with a single-segment `repo_slug` when the URL is not a parseable GitHub path (`lifecycle.rs:651`).

**Disclosed gaps, judged.** Leaving `tm launch` and the guided default routed as-is is correct — both already call `ensure_base_clone` against `<repos_root>/<owner>/<repo>`, so re-routing them changes no topology and only adds risk. The `.git/info/exclude` gap is cosmetic per the `git clean` evidence above, and self-heals the first time the in-project path touches that project (`ensure_base_clone` calls `ensure_worktrees_gitignored` on its reuse branch, `inproject.rs:275`).

**Why this is WARN and not BLOCK.** Neither HIGH is a live-worktree relocation performed by this diff. HIGH-1 is a message, and #3605 is the reason a message counts — but it needs an already-unreadable base to fire. HIGH-2 is a real `fs::rename`, but its trigger condition is untouched by this diff and it preserves data under a named backup rather than deleting it. If the PM reads the brief's "any path that moves a live worktree is a BLOCK" as covering the unguarded neighbour rather than only newly-introduced paths, HIGH-2 is the row to escalate on.

bobmatnyc:
## Round 2 — both HIGHs fixed, MEDIUM and LOW fixed

Rebased onto `b98fb77c`. Head is now `dced44ca`.

Both HIGHs were the same shape as the one I fixed in round 1, one name over: the guard said `.base` when the invariant is **this directory holds work that must never be moved**. Fixing them separately would have left a third hole, so there is now one predicate — `harness_root::protected_state_in` — and both call sites ask it. It names `.git`, `.base`, or `.worktrees`.

**HIGH-1 — `stale_base_dir_error` no longer offers `mv` for a directory holding protected state.** Your reachability analysis was right and it is the reason this mattered: the loose trigger means a transient git spawn failure on a healthy project lands in that message. Only foreign debris — none of the three names — still reaches the quarantine hint, so the #1937 crashed-clone recovery is intact and `stale_base_dir_error_suggests_quarantine_not_deletion` still passes unchanged.

**HIGH-2 — `migrate_old_layout_aside` covers the class, not one member.** Its probe is fail-safe, which is the part your EACCES note demanded: anything that is not a definite `NotFound` counts as present. `Path::exists` maps EACCES to `false`, so under that spelling a permissions blip on `.git` reads as "no git dir" and the rename proceeds over live work. A false positive is a refusal the operator can act on; a false negative is lost work.

**MEDIUM — the real-git test now uses the shipping topology.** `provisioner_worktree_add_writes_no_foreign_branch_merge` points `base_dir` at `<scratch>/owner/repo` (non-bare) and its worktree at `base_dir.join(".worktrees").join(<id>)`, plus an assertion that the nested worktree has a working `.git`. Your hand-verification is now crate coverage.

**LOW — the dead-end sentence is gone.** It pointed at the in-project path, which refuses the same directory. It now names the manual `.base` step, and for the `.git` case the `git rev-parse --show-toplevel` check that would explain the refusal.

### Fail-before, all three new guards

Predicate narrowed back to `.base`-only + `Path::exists`:

```
test core::harness_root::tests::protected_state_in_treats_an_unreadable_entry_as_present ... FAILED
test daemon::managed_routes::inproject::tests::migrate_old_layout_aside_refuses_a_dir_holding_live_worktrees ... FAILED
test provisioner::workspace::tests::stale_base_dir_error_never_offers_to_move_a_dir_holding_live_worktrees ... FAILED
test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 4617 filtered out
```

HIGH-1's failure prints the regression verbatim — the #3605 message aimed at the project directory:

```
#4270: the quarantine hint must never target a directory holding .worktrees —
moving it orphans every session under it, got: git error: base checkout path
/tmp/tm-test-TnvuNk/live-worktrees/repo exists but is not a valid git checkout ...
WARNING: this path is the SHARED git base for EVERY trusty-mpm worktree of this
project, and other sessions may be using it right now.
```

The EACCES test did **not** skip on this machine — it chmods 000, confirms the stat genuinely fails, restores permissions before asserting, and skips only under root.

### Gates — rung 5, re-run at `dced44ca`

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — `test result: ok. 4613 passed; 0 failed; 7 ignored` plus every integration target green |
| `bash scripts/check_line_cap.sh` | EXIT=0 — 3691 files, 7 allowlisted (unchanged), 0 violations |

Test count 4606 → 4613. No allowlist entry, no `#[ignore]`, no test removed.

Every gate log was written to a uniquely-named scratchpad file and grepped for this worktree's path before its exit code was trusted — 7 matches on the clippy and check logs, 3 on the test log.

### Not changed

The two pre-existing convergence divergences you flagged (`TRUSTY_MPM_REPOS_ROOT` splitting the roots; the non-GitHub-URL fallback to single-segment `repo_slug`) are untouched. Neither is introduced here and both need their own decision.

bobmatnyc:
## Verdict: APPROVE — `dced44cade6e5bf27dd16ba9680bc2619c900cdf` (round 2, rebased on `b98fb77c`)

Both round-1 HIGHs are closed. Zero CRITICAL, zero HIGH at >80% confidence. Four
lower-severity findings below; none block.

### The invariant: can any path in this diff move a directory holding live work?

Re-derived at the new head rather than carried over from round 1. Every
filesystem-mutating call reachable from the changed surface:

| Site | Call | Guarded by |
|---|---|---|
| `inproject.rs:248` | `fs::rename(base_path, backup)` | `protected_state_in` at `:213` |
| `base_lock.rs:357`, `:420` | `remove_file(lock_path)` | n/a — a `<owner>/<repo>.lock` marker file, never a directory |
| `workspace.rs:362` | `create_dir_all(parent)` | creates only |
| `workspace.rs:404` | `git clone … <base_dir>` | git refuses a non-empty destination; path pre-checked |
| `workspace.rs:823`, `:838` | sentinel / `TASK.md` writes | inside the freshly created worktree |

No `remove_dir_all`, no second `rename`. The only other move channel is the
agent-executed `mv` inside `stale_base_dir_error`'s message, and that is now
gated by the same predicate at `base_lock.rs:220`.

Constructed false negatives for `protected_state_in`
(`harness_root.rs:91-104`), which probes with `symlink_metadata` and counts
anything that is not a definite `NotFound` as present:

- **`.git` as a FILE** (worktree/submodule form) — `symlink_metadata` returns
  `Ok`, counts as present.
- **symlinked `.git`, including a DANGLING symlink** — `symlink_metadata` stats
  the link itself, returns `Ok`, counts as present. `metadata` would have
  followed and returned `NotFound`; the spelling here is the correct one.
- **EACCES** — proven present, see the fail-before below.
- **`.worktrees` present but empty** — present, refused. Conservative in the
  right direction.
- **case-insensitive FS (`.GIT`, `.Base`)** — `dir.join(".git")` matches `.GIT`
  on APFS; on a case-sensitive FS a `.GIT` directory is not a git dir. No hole.
- **a project dir holding only `.claude/worktrees/`** — the predicate returns
  `None` and `migrate_old_layout_aside` would rename it. Reaching that requires
  a hand-placed git worktree under `<repos_root>/<owner>/<repo>/.claude/worktrees/`
  in a directory that has no `.git` of its own. Below the 80% bar; recorded in
  Notes, not filed.
- **a worktree registered in git whose directory was moved** — the predicate is
  a filesystem probe of the parent, not a registry query, so a moved-away
  worktree leaves nothing at the old path to protect. Not a false negative.

### Predicate equivalence — does the merge narrow either guard?

No; it is a strict superset in both directions. At `73eb57cb` both guards asked
about `.base` alone (`inproject.rs:206`, `base_lock.rs:117` `legacy_base_store_error`),
using `Path::exists`-family probes. At `dced44ca` both ask `protected_state_in`
for `{.git, .base, .worktrees}` with the fail-safe probe. The call position is
unchanged in both: after the non-empty check and before the quarantine message
in `stale_base_dir_error`, after the empty-dir early return in
`migrate_old_layout_aside`.

### #1937's quarantine hint

`stale_base_dir_error_suggests_quarantine_not_deletion` is genuinely unchanged —
the only edit inside the body is an `.expect()` string. Its fixture (a directory
holding only a stray `HEAD`) still reaches the `mv` hint, because
`protected_state_in` looks for `.git`/`.base`/`.worktrees` *inside* the probed
directory, not at its own name. Passes. The production-shaped equivalent
(`<owner>/<repo>` rather than a legacy `.base` path) is covered by
`fake_ensure_base_checkout_rejects_stale_directory`. See MEDIUM-2 for the part
of the #1937 claim that does not hold.

### EACCES probe

All four claims verified at `harness_root_tests.rs:62-93`: it chmods the
containing directory to `0o000` (`:70`), computes `unreadable` from a real
`symlink_metadata` call and requires a non-`NotFound` error (`:72-74`), restores
permissions **before** asserting (`:78`), and skips with a printed reason when
the stat still succeeds — which is the root case and any platform that lets it
through (`:80-86`).

### Fail-before, reproduced

Narrowed `protected_state_in` to `.base`-only with `Path::exists`, ran the three
guard families, then reverted (tree clean, `HEAD` still `dced44ca`):

```
test core::harness_root::tests::protected_state_in_names_each_protected_entry ... FAILED
test core::harness_root::tests::protected_state_in_treats_an_unreadable_entry_as_present ... FAILED
test daemon::managed_routes::inproject::tests::migrate_old_layout_aside_refuses_a_dir_holding_live_worktrees ... FAILED
test provisioner::workspace::tests::stale_base_dir_error_never_offers_to_move_a_dir_holding_live_worktrees ... FAILED
test result: FAILED. 4 passed; 4 failed; 0 ignored; 0 measured; 4612 filtered out
```

HIGH 1's failure prints the #3605 string verbatim, aimed at the directory
holding `.worktrees/sess-1`:

```
#4270: the quarantine hint must never target a directory holding .worktrees — moving it
orphans every session under it, got: git error: base checkout path
/tmp/tm-test-NjdQ54/live-worktrees/repo exists but is not a valid git checkout …
    mv /tmp/tm-test-NjdQ54/live-worktrees/repo /tmp/tm-test-NjdQ54/live-worktrees/repo.stale-20260805-160125
```

`migrate_old_layout_aside_refuses_a_dir_holding_a_dot_base_store` still passes
under that mutation, as expected — it seeds `.base`, which the mutation kept.

### Gates run at `dced44ca`

```
test result: ok. 4613 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 182.36s
(plus 43 further binaries; workspace-wide for this crate: 7772 passed, 0 failed, 17 ignored)

( cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings && bash scripts/check_line_cap.sh )
EXIT=0
line-cap: measured 3691 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

All 14 new/renamed tests observed running and passing by name. Spot-check clean:
4613 lib tests (from 4606), no `.line-cap-allowlist.tsv` change, no `#[ignore]`
added, zero `#[test]` functions removed (9 added), changelog fragment present at
`crates/trusty-mpm/changelog.d/4270-provision-worktrees.md` with a single
`Changed` category, and 48 `#4270` attribution comments in the diff.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/provisioner/workspace.rs` | 409-422 | The provisioner path never adds `.worktrees/` to the base clone's `.git/info/exclude`. It now produces the identical `<project>/.worktrees/<id>` shape inside a **non-bare** base, which is exactly what `ensure_worktrees_gitignored` exists for on the sibling path (`inproject.rs:283`, `:339`). Reproduced against real git: `git status` in the base prints `?? .worktrees/`, and `.git/info/exclude` carries only git's default template. Beyond the status noise, a `git clean -fd` in the base now removes every session worktree. | Call `crate::daemon::managed_routes::inproject::ensure_worktrees_gitignored(base_dir)` best-effort after the successful-clone branch, warning on failure the way the sentinel write does. Same helper, no second implementation. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/provisioner/workspace/base_lock.rs` | 121 | The PR states foreign debris still reaches the quarantine hint so "#1937's crashed-clone recovery survives". That does not hold for the dominant shape in the shipping topology: a non-bare `git clone` writes `<dir>/.git` before the object database, so a crash mid-clone leaves `.git` present and `protected_dir_error` diverts it. The operator's two concrete suggestions then are "retiring a `.base` store is a manual step" and "check `git rev-parse --show-toplevel`", neither of which addresses a half-clone. On `main` that same crash reached the `mv` hint. The divert itself is right — a `mv` at a dir with `.git` is the #3605 shape. | Either correct the claim, or add the one narrow branch that is provably safe: `.git` present, `.base` and `.worktrees` both absent, `rev-parse` failing → nothing can be a live worktree beneath it, so the quarantine hint applies. | Parent |
| MEDIUM | `crates/trusty-mpm/src/provisioner/workspace/base_lock.rs` | 121-140 | Every project already provisioned through the URL path has `<project>/.base` and no `<project>/.git`, so `is_established_checkout` returns false and `provision_in` returns `Err` on **every** subsequent spawn for that project — pinned by `provision_in_leaves_an_existing_dot_base_store_untouched` asserting `result.is_err()`. The message deliberately names no command, so the operator gets a hard stop with no stated way forward. This is the owner-ruled trade, but "cleanup is manual" and "every existing project stops spawning on upgrade" are different-sized consequences and the second one should be a knowing choice. | Owner call. Either keep serving a project whose `<project>/.base` is still an established bare checkout on the legacy layout, or name a `tm repair`-style step in the refusal so the message is actionable. | Parent |
| LOW | `crates/trusty-mpm/src/provisioner/workspace/tests.rs` | 162 | `provision_in_leaves_an_existing_dot_base_store_untouched` uses an ad-hoc `["rm -rf", "rm -fr", "    mv "]` list instead of `assert_no_recursive_delete` (introduced at `:363` for exactly this), so it misses `rm -r `, `remove_dir_all`, and `rmdir`. | Replace the loop with `assert_no_recursive_delete(&msg);` plus the existing `mv` assertion. | Fix here |
| LOW | `crates/trusty-mpm/src/provisioner/workspace/base_lock.rs`, `workspace.rs` | 25, 752 | Two doc references to `legacy_base_store_error`, renamed to `protected_dir_error` in this same PR. `base_lock.rs:25` is an intra-doc link (`[`legacy_base_store_error`]`) that no longer resolves; `workspace.rs:752` is prose. Clippy does not run rustdoc lints, so both passed the gate. | Rename both to `protected_dir_error`. | Fix here |

## Notes

- **Out of scope, confirmed pre-existing.** Neither `TRUSTY_MPM_REPOS_ROOT`
  splitting the two provisioning paths onto different roots nor
  `spawn_managed_cloned`'s single-segment `repo_slug` fallback for a non-GitHub
  URL is touched by this diff — `git diff b98fb77c...dced44ca` matches
  `repo_slug`/`repos_root`/`workspace_root`/`TRUSTY_MPM_REPOS_ROOT` only inside
  doc comments. Not developed further.
- **Known baseline, not attributed.** `cargo check --workspace` fails on the
  Tauri `frontendDist` proc-macro panic in the GUI crates; not exercised.
  `cargo test -p trusty-mpm` was run with `--test-threads=1` per
  `docs/reference/test-ladder-baseline.md`.
- The nested-worktree question the MEDIUM fix in round 1 was about is genuinely
  settled: `provisioner_worktree_add_writes_no_foreign_branch_merge` now runs
  real git against the shipping topology and asserts `worktree.join(".git")`
  exists. I reproduced the same shape by hand independently.
- Review ran from a dedicated detached worktree at
  `.claude/worktrees/critic-4926b`; no PR file was modified except the
  fail-before mutation, which was reverted (`git status --porcelain` empty,
  `HEAD` = `dced44ca`).

bobmatnyc:
## Round 3 — the `git clean` contradiction, settled empirically

Head is `db980b42`. Already on current main (`b98fb77c`); no rebase needed.

### The measurement

Throwaway repo, shipping topology: non-bare base at the project dir, `git worktree add` into `<base>/.worktrees/sess-1`, one uncommitted `WIP.txt` inside it. Real git 2.54.0. Every row below is a **real** run, not `-n` — the dry runs agreed, but a dry run is a prediction.

| command | without exclude entry | with exclude entry |
|---|---|---|
| `git clean -fd` | `Skipping repository .worktrees/sess-1` — WIP survives | survives |
| `git clean -xfd` | `Skipping repository .worktrees/sess-1` — WIP survives | survives |
| **`git clean -ffd`** | **`Removing .worktrees/` — WIP DELETED, dir gone** | **nothing removed — survives** |
| `git clean -xffd` | `Removing .worktrees/` — DELETED | `Removing .worktrees/` — DELETED |

After the `-ffd` run, `git worktree list` showed the entry as `prunable`.

**Both rounds were half right.** Round 1's `Would skip repository` is real — single-force `clean` refuses to descend into a nested repository, and a linked worktree's `.git` **file** counts as one. That is exactly what makes this look harmless. Round 2's deletion is also real, but it needs `-ff`, not `-fd`.

So the entry is **safety, not parity**: it converts `git clean -ffd` from data loss into a no-op. `-xffd` defeats it because `-x` discards ignore rules; nothing short of not running that command helps there.

**Fixed.** The provisioner now writes the entry on both the fresh-clone and reuse paths, exactly as `inproject::ensure_base_clone` always has.

### Where the helper lives

Not a one-line call — `provisioner::workspace` is not gated behind the `daemon` feature and cannot reach the `inproject` copy. `ensure_worktrees_gitignored` moved to `core::worktree_naming`, which exists for precisely this dependency direction and already hosts `worktree_branch_for`. `inproject` re-exports it, so its callers are untouched. The pattern is now built from `WORKTREES_DIRNAME` instead of a second `.worktrees/` literal.

New test `worktrees_exclude_entry_protects_against_double_force_clean` builds the topology with the production backend and runs a real `git clean -ffd`. Fail-before, with only the provisioner's call removed:

```
ensure_base_checkout must exclude .worktrees/, got: # git ls-files --others --exclude-from=.git/info/exclude
```

### The #1937 claim

You are right and I was wrong. A non-bare clone writes `.git` early, so a half-clone now diverts to `protected_dir_error` and gets no command; on main it reached the `mv` hint, because a partial **bare** clone writes no `.git`. Recovery for that case is genuinely worse.

I corrected it in place rather than deleting it — `protected_dir_error` now carries a `KNOWN REGRESSION, accepted` paragraph saying so, and `ensure_base_checkout`'s doc no longer implies the `.base` case is the only divert. The trade is deliberate: a half-clone and a healthy checkout whose git probe just failed are indistinguishable from there, and one of them holds live worktrees. Guessing wrong toward recovery destroys work; guessing wrong this way costs one manual decision.

I did **not** add a command hint to any branch — that would reintroduce HIGH-1.

### Two LOWs

Ad-hoc forbidden-string list now calls `assert_no_recursive_delete`. Both stale `legacy_base_store_error` references fixed, including the broken intra-doc link.

### Held, as instructed

No command hint added to the `.base` branch. Awaiting the owner's answer on whether existing `.base` projects hard-stopping on upgrade is what "manual cleanup" meant.

### Gates — rung 5 at `db980b42`

| Gate | Result |
|---|---|
| `cargo test -p trusty-mpm -- --test-threads=1` | EXIT=0 — `test result: ok. 4614 passed; 0 failed; 7 ignored` |
| `bash scripts/check_line_cap.sh` | EXIT=0 — 3691 files, 7 allowlisted (unchanged), 0 violations |

Clippy first failed on `doc_lazy_continuation` — a line beginning `+ \`worktree_add\`` parsed as a markdown list. Reworded, not allowed.

One housekeeping note: a `wip(fix-4270): checkpoint … before session wind-down` commit I did not author appeared mid-run and captured the working tree. I squashed it into the round-3 commit; the resulting tree is byte-identical (`git diff` against the pre-squash SHA is empty).

bobmatnyc:
## Correction: this PR does not hard-stop existing `.base` projects

A claim has been circulating that after #4926 lands and the binary is rebuilt, a project whose worktrees live under `.base/.worktrees/` hard-stops on its next spawn. **That is false.** Posting the measurement because the premise reached at least one set of session notes as fact.

### Why it is false

`base_dir` is the project directory itself ([workspace.rs:784](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace.rs#L784)). The first thing `ensure_base_checkout` does is ask git whether that directory is already a checkout ([workspace.rs:358-363](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace.rs#L358-L363)):

```rust
if is_established_checkout(base_dir) {
    debug!(path = %base_dir.display(), "base checkout already present, reusing");
    return exclude_worktrees(base_dir);   // #4270 self-heal on reuse
}
```

`is_established_checkout` ([base_lock.rs:71-86](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace/base_lock.rs#L71-L86)) runs `git -C <dir> rev-parse --show-toplevel` and compares the canonicalized result to the directory. Measured against the real path of a `.base`-shaped project:

```
$ git -C /Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools rev-parse --show-toplevel
/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools
$ git -C … rev-parse --is-bare-repository
false
```

Toplevel equals the directory, so the predicate is true and the function returns at line 363. `protected_dir_error` ([base_lock.rs:130](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace/base_lock.rs#L130)) is only reachable via `stale_base_dir_error` ([base_lock.rs:229](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace/base_lock.rs#L229)), which sits **after** that early return. It cannot be reached for any project directory that is itself a git checkout.

What the user sees: nothing. No warning, no refusal. New sessions provision at `<project>/.worktrees/<id>`; sessions already under `.base/.worktrees/` keep running, because their `.git` file points at `.base/worktrees/<id>` (git's admin dir) and nothing here reads, moves, or deletes either path.

The reuse-path `exclude_worktrees` call is a no-op on an already-guarded repo — `ensure_worktrees_gitignored` skips when the trimmed entry is already present ([worktree_naming.rs:85](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/core/worktree_naming.rs#L85)).

### Blast radius

44 registered project directories scanned. Two hold a `.base`; both are established checkouts, so both take the reuse branch. **Zero affected.**

| project | `.base/.worktrees` entries | path taken |
|---|---|---|
| `bobmatnyc/trusty-tools` | 3 | reuse — no stop |
| `duettoresearch/duetto` | 1 | reuse — no stop |

### What the guard actually catches

Only a project dir that is **not** an established checkout **and** holds `.git`, `.base`, or `.worktrees` — a dir tm only ever cloned into `.base`, or a transient git failure (missing binary, dubious-ownership rejection) with `.git` present. Neither shape exists on this machine. That case is a per-spawn `ProvisionError::Git`, not a refusal to run `tm`.

Pinned by `provision_in_leaves_an_existing_dot_base_store_untouched` ([tests.rs:228](https://github.com/bobmatnyc/trusty-tools/blob/db980b42/crates/trusty-mpm/src/provisioner/workspace/tests.rs#L228)), which pre-seeds a live `.base/.worktrees/live/` worktree holding a marker file and asserts the marker survives provisioning untouched.

No behavior was changed to produce this comment — it is a read of `db980b42` as it stands.

Issue #4847, opened 2026-08-04
tctl stack doctor and tctl stack health treat unknown health as passing


bobmatnyc:
## Defect

`tctl stack doctor` and `tctl stack health` return success and exit 0 when a component's health is `unknown`. Both commands collapse two distinct states — healthy and undetermined — into a single passing verdict.

## Evidence — observed output, 2026-08-04

```
tctl stack doctor
  trusty-search      0.42.0       path:yes plist:NO  port:yes health:healthy
  trusty-memory      0.21.2       path:yes plist:yes port:yes health:healthy
  trusty-analyze     0.7.4        path:yes plist:NO  port:yes health:healthy
  trusty-review      0.11.0       path:yes plist:NO  port:yes health:healthy
  trusty-console     0.4.0        path:yes plist:NO  port:yes health:healthy
  trusty-mpm         1.3.4        path:yes plist:n/a port:yes health:unknown
verdict: ok (exit 0)

tctl stack health
  ...
  trusty-mpm         unknown
verdict: ready (exit 0)
```

Both commands return exit 0 while trusty-mpm reports `health:unknown`. This is a fail-open pattern: the branch that cannot determine a result advances state as though it succeeded.

## Impact

`unknown` means "I could not tell," not "it is fine." Collapsing the two makes a green verdict unfalsifiable for that component. This matters because vmtest-harness under development bases PASS criteria on these health checks. A harness that verifies an installation can currently go green on a stack that is not up.

## Resolution direction

`unknown` must be distinguishable from `healthy` in the exit code and the verdict. Whether that means a third exit state, a stricter default with an opt-out flag, or `unknown` simply counting as failure is a design call — but the two states must not collapse.

## Additional observation

Four of six components report `plist:NO` — only trusty-memory has a plist, and trusty-mpm shows `n/a`. If those daemons genuinely have no launchd plist on that machine, they will not survive a reboot. Note that "installed correctly" and "will survive a restart" are different acceptance criteria. This may reflect how that machine was installed, so frame it as a question worth answering rather than an assertion.

See issue [#920](https://github.com/bobmatnyc/trusty-tools/issues/920) for the broader trusty-controller initiative and related health-reporting standardization work.

bobmatnyc:
Fix in progress — wave-1 batched PR, 2026-08-30 zero-bug drive.

bobmatnyc:
Claimed by session trusty-tools-77 (tm-trusty-tools-01), 2026-08-31.

[...]

PR #4935, opened 2026-08-05
fix(trusty-installer): probe trusty-mpm /health over HTTP instead of reporting unknown

### LINKED ISSUE #4925: Probe trusty-mpm's /health over HTTP instead of reporting it unknown
## Summary

`probe_member_health` returns `ProbeOutcome::Unprobeable` for any member whose
`ManageStrategy` is `OwnVerb`, which is trusty-mpm alone. The daemon answers
`GET http://127.0.0.1:7880/health` with a well-formed envelope, its `http_addr`
discovery file is present, and `probe_http::fixed_port_for` already maps
`"trusty-mpm" => Some(7880)`. Every ingredient for a real probe is in place; the
match arm declines to use them. This was a deliberate carve-out in #4246/#4381 —
this issue proposes reversing it, and states the user-visible consequence that
reversal carries.

## Observed behaviour

```
$ tctl status
tctl status — stack summary
  trusty-search      0.42.0       healthy
  trusty-memory      0.21.2       healthy
  trusty-analyze     0.7.4        healthy
  trusty-review      0.11.0       healthy
  tga                2.11.0       n/a
  trusty-console     0.4.0        healthy
  trusty-mpm         1.3.4        unknown
verdict: ready (exit 0)
```

`tctl stack doctor` on the same host is more pointed — it resolves mpm's port
and then reports its health as undetermined in the same row:

```
$ tctl stack doctor
  trusty-search      0.42.0       path:yes plist:NO  port:yes health:healthy
  trusty-memory      0.21.2       path:yes plist:yes port:yes health:healthy
  trusty-analyze     0.7.4        path:yes plist:NO  port:yes health:healthy
  trusty-review      0.11.0       path:yes plist:NO  port:yes health:healthy
  trusty-console     0.4.0        path:yes plist:NO  port:yes health:healthy
  trusty-mpm         1.3.4        path:yes plist:n/a port:yes health:unknown
verdict: ok (exit 0)
```

`tctl --version` → `trusty-installer 0.5.0`.

## Expected behaviour

trusty-mpm is probed over HTTP like every other daemon member, and reports
`healthy` when it is serving, `down` when it is not.

## Root cause

`crates/trusty-installer/src/commands/probe.rs:141-158`:

```rust
pub fn probe_member_health(binary: &str, manage: ManageStrategy) -> ProbeOutcome {
    if resolve_binary_path(binary).is_none() {
        return ProbeOutcome::NotInstalled;
    }
    match manage {
        ManageStrategy::Launchd => probe_member_http_blocking(binary, binary),
        // #4246: trusty-mpm (`OwnVerb`) is DELIBERATELY left unprobed and
        // reported `unknown`, even though it does answer `/health` on 7880.
        // It is `required: true` (`stable_set`), and `unknown` is the only
        // verdict `VerifyTailReport::build` and `status`'s exit code tolerate —
        // so probing it would flip `tctl status` to exit 2 and `tctl install` to
        // NOT VERIFIED for every user who simply has not started mpm. Enabling
        // it is a separate, user-visible policy change, tracked separately.
        // `None` (non-daemon) never reaches here in production (callers filter on
        // `m.daemon`) but shares the safe default.
        ManageStrategy::OwnVerb | ManageStrategy::None => ProbeOutcome::Unprobeable,
    }
}
```

`ManageStrategy` is doing two jobs that used to coincide and no longer do. It
answers "how do I start and stop this member?" — `launchctl bootstrap` versus
`trusty-mpm start`. Before #4246 the probe was `<binary> health --json`, a
subprocess contract, so lifecycle strategy and probe strategy were plausibly the
same axis. #4246 moved the probe to HTTP `/health`, which is a property of the
daemon's transport and has nothing to do with how the process is supervised. The
two axes separated; the match arm still keys probeability off the lifecycle
enum.

The current host demonstrates the separation directly: trusty-search,
trusty-analyze, trusty-review and trusty-console all show `plist:NO` — no
launchd artifact present — and all four probe `healthy`. Being launchd-managed
is already not a precondition for being probeable.

## Evidence

The daemon serves:

```
$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:7880/health
200
$ curl -s http://127.0.0.1:7880/health
{"status":"ok","catalog_stale":true,"catalog_unknown":false,
 "catalog_changes":["agent dart-engineer: new", ...],"supervised":false}
```

`status: "ok"` maps to `MemberHealth::HealthyVersionOk` via
`up/system_runner.rs:76-82`, so the existing classifier handles this payload with
no changes.

The discovery file is present and correct:

```
$ cat "$HOME/Library/Application Support/trusty-mpm/http_addr"
127.0.0.1:7880
```

The fixed-port fallback is already in the table —
`crates/trusty-installer/src/commands/probe_http.rs:102`:

```rust
"trusty-mpm" => Some(7880),
```

Both resolution paths (`http_addr` first, fixed port as fallback) are wired at
`probe_http.rs:533`. `tctl stack doctor`'s `port:yes` for mpm is produced by
`read_daemon_addr` at `stack/doctor.rs:120-123` — the binary reads this file on
every doctor run today.

One caveat for the implementer: mpm's envelope carries no `version` field
(noted at `probe_http.rs:164-165`), so it will report `healthy` rather than
`stale`. Version-floor detection for mpm is out of scope here.

## Impact

**(a) Misleading `unknown` in every rollup.** Four production call sites render
it: `status.rs:112`, `stack/health.rs:101`, `stack/doctor.rs:131`, and the
install verify tail at `verify_tail.rs:617`. Each prints "I could not tell" about
a daemon that answered 200 on the first try.

**(b) `tctl install` prints VERIFIED with a dead required member.**
`run_verify_tail` (`verify_tail.rs:716-723`) filters on `m.daemon`, and mpm is
`daemon: true` (`stable_set.rs:181`), so mpm *is* probed. But
`VerifyTailReport::build` (`verify_tail.rs:198-209`) derives:

```rust
let health_ok = members
    .iter()
    .filter(|m| m.required)
    .all(|m| m.health != health_str::DOWN && m.health != health_str::NOT_INSTALLED);
```

`unknown` is neither, so mpm — `required: true` — can never fail verification.
An install that leaves mpm dead prints VERIFIED and exits 0.

**(c) `tctl up` issues a redundant start every run.**
`ProbeOutcome::member_health` (`probe_http.rs:233-239`) maps `Unprobeable`
through the catch-all to `MemberHealth::Down`; `system_runner.rs:103-106` feeds
that to `ensure_member`, which falls through to `<binary> start`. The start is
idempotent, so this is cost rather than corruption — but it is a process spawn on
every `tctl up` against a daemon already known to be serving, and it is
indistinguishable in the logs from a real recovery.

## Why this is NOT a `stable_set.rs` misclassification

The tempting fix — reclassify mpm as `Launchd` so it takes the probing arm — is
wrong and dangerous.

The classification is correct. There is no `com.trusty.mpm` launchd job:

```
$ launchctl print gui/501/com.trusty.mpm
Could not find service "com.trusty.mpm" in domain for user gui: 501

$ ls ~/Library/LaunchAgents | grep trusty
com.trusty.memory.plist
com.trusty.mpm.supervisor.plist
```

`com.trusty.mpm.supervisor` exists and is running, but it is a distinct job
(`program = /Users/mac/.local/bin/tm`, `arguments = { tm, supervisor }`) that
supervises sessions — not a launchd job for the mpm daemon. The daemon's own
health envelope agrees: `"supervised": false`. mpm is process-managed via its own
`start`/`stop` verbs, exactly as `stable_set.rs:133-141` and #1332 decision 3
describe.

The danger is `needs_kickstart` (`verify_tail.rs:237-239`):

```rust
pub fn needs_kickstart(outcome: &ProbeOutcome, manage: ManageStrategy) -> bool {
    outcome.is_confirmed_down() && manage == ManageStrategy::Launchd
}
```

Flipping mpm to `Launchd` would make a genuinely-down mpm eligible for
`launchctl kickstart -k` against a label that does not exist — and would also
break `tctl start|stop|restart`, which dispatches on the same enum. The
`ManageStrategy` axis must stay as-is. Only the probe's use of it changes.

## Proposed fix

One arm in `probe.rs:145-157`. Route `OwnVerb` to the same HTTP transport as
`Launchd`, and leave `None` on the safe default:

```rust
match manage {
    ManageStrategy::Launchd | ManageStrategy::OwnVerb => {
        probe_member_http_blocking(binary, binary)
    }
    ManageStrategy::None => ProbeOutcome::Unprobeable,
}
```

`crate_name == binary == app name` holds for trusty-mpm's stable-set entry
(`stable_set.rs:181`), so the `http_addr` lookup key is correct. The module doc
at `probe.rs:132-137` currently asserts the opposite as its justification for the
carve-out and must be updated with it.

`needs_kickstart` needs no change: it independently requires
`manage == Launchd`, so a confirmed-down mpm still cannot be kickstarted. That
is the property that makes this a one-arm change rather than a cross-cutting one.

Rejected alternatives:

- **Reclassify mpm as `Launchd`** — see the section above. Breaks lifecycle
  dispatch and arms `kickstart -k` against a nonexistent label.
- **Add a separate `probeable: bool` to `StableMember`** — a second field to keep
  in sync with `manage` for one member, and `tctl up` works from a `BootMember`
  which has no such field (the divergence `manage_strategy_for` was extracted at
  `stable_set.rs:120-141` to prevent). The `None` arm already expresses
  "not probeable" adequately.
- **Probe mpm via the console gateway (`:7788/api/mpm`)** — mpm's documented
  resolution path, but it makes mpm's health contingent on trusty-console being
  up, and console is `required: false`. Direct loopback is the honest signal.
- **Leave it and fix the verdict policy instead (#4847)** — treats the symptom.
  The rollups should stop *producing* an `unknown` they have no reason to
  produce. Both changes are worth making; they are independent.

## User-visible consequence that must be intentional

This is the reason #4246/#4381 deferred it, and it must be accepted explicitly
rather than discovered in a release.

After this change, a user who has simply not started mpm gets:

- `tctl status` → `trusty-mpm … down`, `verdict: degraded`, **exit 2**
  (`status.rs:65, 88-93`)
- `tctl stack health` / `tctl stack doctor` → `degraded`, **exit 2**
- `tctl install` → **NOT VERIFIED**, non-zero exit, because mpm is
  `required: true` and `down` fails `VerifyTailReport::build`

This is precisely the behaviour trusty-search, trusty-memory and trusty-review
already have — they are `required: true`, and a stopped one degrades the stack
today. The change makes mpm consistent with its declared `required` flag rather
than exempt from it. It is still a behaviour change for anyone whose workflow
tolerates a stopped mpm, and CI that gates on `tctl status` exit codes will start
failing where it previously passed.

If that consequence is not wanted, the alternative is to demote mpm to
`required: false` in `stable_set.rs:181` — but that should be a separate,
argued decision, not an accident of the probe being disabled.

## Tests required

- Rewrite `probe::tests::own_verb_member_is_unknown` (`probe.rs:299-311`). It
  currently asserts `Unprobeable` / `unknown` / `!is_confirmed_down()` and its
  doc comment reads "Deliberately unchanged by this fix; pin it so a
  well-meaning follow-up cannot flip it silently." That pin is exactly what this
  issue proposes removing, so the replacement must assert the new intent: an
  `OwnVerb` member with a planted `http_addr` reports `Serving` / `healthy`.
  Model it on `probe_member_health_serves_from_http_addr` (`probe.rs:302`).
- Add an `OwnVerb` counterpart to
  `probe_member_health_refused_when_nothing_listens` (`probe.rs:328`) asserting
  `Refused` → `down`.
- Assert `needs_kickstart(&ProbeOutcome::Refused, ManageStrategy::OwnVerb)` is
  still `false`. `verify_tail_tests.rs:81` covers this; keep it green and make
  the intent explicit in its doc comment, since it is now load-bearing for a
  live code path rather than a hypothetical one.
- Add a `VerifyTailReport::build` case: a `required` member reporting `down`
  fails verification, using trusty-mpm as the member.
- `stable_set.rs::mpm_uses_own_verb` (`:373-381`) and
  `manage_strategy_for_matches_the_stable_set` (`:350-368`) must remain
  unchanged and green — proof the lifecycle axis was not touched.
- Update `vmtest-harness/lib/verify.sh` (`:564, 575, 747, 980, 983`) and
  `docs/research/tart-vm-testing-harness/02-design/02-harness-contracts.md`
  (`:171, 185, 202, 615, 618`), both of which encode "mpm is deliberately left
  unprobed" as a harness invariant.

## Relationship to #4246 and #4847

**#4246** (open) is the parent. It replaced the `<binary> health --json`
subprocess probe with HTTP `/health`, merged as `6078b21c` (PR #4381) with
follow-ups in `bd280021` (PR #4388). Its body already lists this work as item 4:

> **Change the `ManageStrategy::OwnVerb` arm too** (`probe.rs:271-272`) or the
> trusty-mpm row stays `unknown` — it returns `unknown` without probing
> anything. mpm has an `http_addr` (`127.0.0.1:7880`) and answers `/health`.

and a #4246 comment escalates it:

> **Item 4 is not a bullet point.** trusty-mpm is `required: true`. Today
> `unknown` is why it never trips `any_down` or the required-member gate. Making
> it probeable means a user who simply hasn't started mpm gets `tctl status`
> **exit 2** and `tctl install` → **NOT VERIFIED**.

PR #4381 deferred it under "Deliberately out of scope":

> trusty-mpm's `OwnVerb` arm still returns `unknown`. Probing it would flip
> `tctl status` to exit 2 and `tctl install` to NOT VERIFIED for anyone not
> running mpm, since mpm is `required: true`

That deferral was correct — it kept a behaviour change out of a bug fix. But no
dedicated follow-up was ever filed; `gh issue list --state all` finds nothing
for "OwnVerb", "Unprobeable", or "trusty-mpm probe unknown" beyond #4246 itself.
This issue is that follow-up, scoped to item 4 alone.

**#4847** (open) reports the same symptom from the other end: `tctl stack
doctor` and `tctl stack health` treat `unknown` as passing, which is a fail-open
verdict policy. The two are complementary and independent. #4847 asks "should an
undetermined health advance a green verdict?" — a policy question that stays open
and untouched here, and that matters for every future member whose health
genuinely cannot be determined. This issue asks a narrower question: "why is
mpm's health undetermined at all, when it answered 200?" Fixing this one removes
the only member that currently produces `unknown` in practice, but it does not
resolve #4847's policy gap, and #4847 should not be closed by it.

## Policy decision

The user-visible consequence above has been reviewed and **accepted**: a stopped
trusty-mpm rendering `down` / `degraded` / exit 2, and `tctl install` returning
NOT VERIFIED, is the intended behaviour. mpm should be consistent with its
declared `required: true` flag rather than exempt from it. Implementation may
proceed on that basis; `required: false` demotion is explicitly NOT the chosen
path.

mac-duetto:
Closes #4925.

## What changed

One match arm in `crates/trusty-installer/src/commands/probe.rs` — `probe_member_health`
now routes `ManageStrategy::OwnVerb` through the same HTTP transport as `Launchd`:

```rust
-        ManageStrategy::Launchd => probe_member_http_blocking(binary, binary),
-        ManageStrategy::OwnVerb | ManageStrategy::None => ProbeOutcome::Unprobeable,
+        ManageStrategy::Launchd | ManageStrategy::OwnVerb => {
+            probe_member_http_blocking(binary, binary)
+        }
+        ManageStrategy::None => ProbeOutcome::Unprobeable,
```

`ManageStrategy` was overloaded to mean two things that stopped being the same thing
in #4246: how a member is started/stopped, and whether its health is probeable. Those
coincided while the probe was `<binary> health --json`, a subprocess contract. Once
#4246 moved the transport to HTTP `/health`, probeability became a property of the
daemon's transport rather than its supervision model — but the match arm still keyed it
off the lifecycle enum, leaving trusty-mpm (the only `OwnVerb` member) reported as
`unknown` despite answering `GET http://127.0.0.1:7880/health` with 200.

Everything else in the diff is doc comments in the eight files that cited mpm as the
live "unprobeable" example, tests, a `changelog.d` fragment, and lockstep updates to
`vmtest-harness/lib/verify.sh` and the harness-contracts doc. `stable_set.rs` is not in
the diff at all — mpm keeps `OwnVerb` for lifecycle, which is what keeps `needs_kickstart`
unable to fire for it.

## Accepted policy change

This is user-visible and was reviewed and accepted before implementation (see #4925).

A stopped trusty-mpm now yields `tctl status` → `down` / `degraded` / **exit 2**, and
`tctl install` → **NOT VERIFIED**, because mpm is `required: true` (`stable_set.rs:181`).
This makes mpm consistent with its declared `required` flag rather than exempt from it —
the same behaviour trusty-search, trusty-memory and trusty-review already have. CI that
gates on `tctl status` exit codes may start failing where it previously passed.

`required: false` demotion was explicitly **not** the chosen path. No env-var opt-out,
no `probeable` field, no second probe surface, no `--json` added to `trusty-mpm health`,
no mpm special-case in any caller.

## Verification

Built the patched binary and ran it against the live daemon, with the pre-fix binary
installed at `~/.cargo/bin/tctl` as a same-host, same-moment control:

```
$ ./target/debug/tctl status          # patched
  trusty-mpm         1.3.4        healthy
$ tctl status                         # pre-fix
  trusty-mpm         1.3.4        unknown
```

`tga` correctly stays `n/a` — the non-daemon arm is intact.

`make check` fails identically on this branch **and** on `origin/main` — the two Tauri
crates `trusty-code-gui` and `trusty-mpm-gui` both fail with `The frontendDist
configuration is set to "ui/dist" but this path doesn't exist`, because `crates/*/ui/dist`
is gitignored and needs a pnpm build. Neither crate appears in this diff. Reproduced on
`origin/main` in a throwaway worktree, so the CI-equivalent gate (per
`.github/workflows/ci.yml:23-28`) was run instead:

| Check | Result |
|---|---|
| `cargo clippy --workspace --all-targets -- -D warnings` (3 GUI excludes) | exit 0 |
| `cargo test --workspace` (same excludes) | 21972 passed, 0 failed, 139 ignored, 207 suites |
| `trusty-installer` package suite | 554 passed, 0 failed, 4 ignored |
| line-cap / sld-lint / doc-numbers / test-pointers / generation-artifacts / changelog-fragment | 0 violations |

`probe.rs` is 252 SLOC against the 500 cap (was 217).

## Tests

- Rewrote `probe::tests::own_verb_member_is_unknown` → `own_verb_member_is_probed_over_http`.
  Inverted rather than deleted, so the reversal is recorded where the carve-out was.
- Added `probe::tests::own_verb_member_refused_when_nothing_listens` — drives a real
  probe to `Refused` with `OwnVerb` and asserts `needs_kickstart(&outcome, OwnVerb) == false`.
  This is the safety-critical pin: a confirmed-down mpm must never be eligible for
  `launchctl kickstart -k com.trusty.mpm`, a label that does not exist.
- Added `probe::tests::non_daemon_member_is_unprobeable` — pins that `None` keeps
  `Unprobeable`, so the widened arm cannot creep to cover every strategy.
- Added `verify_tail::tests::required_mpm_reporting_down_fails_verification`.
- `stable_set::tests::mpm_uses_own_verb` and `manage_strategy_for_matches_the_stable_set`
  are unmodified and green — proof the lifecycle axis was untouched.

## The one behavioural non-Rust edit

`vmtest-harness/lib/verify.sh` widens the `H_P` accepted-health set for
`plist_installed == null` members by one word:

```
-                    accepted="${accepted} unknown"
+                    accepted="${accepted} unknown down"
```

Without it the harness would go red on every run: the oracle snapshots `tctl stack doctor`
before anything starts daemons, so post-fix mpm reads `down` there, and `down` was
previously accepted only for `plist_installed == false`.

Net assertive power at that step is unchanged. `null` ⟺ daemon + `OwnVerb` ⟺ trusty-mpm
alone (`doctor.rs:117-123`, `stable_set.rs:133`); a `Launchd` member with a missing plist
gets `Some(false)` and already accepted `down`. Pre-change mpm read `unknown`
unconditionally at that site, so it asserted nothing about mpm either way. `not_installed`
is still rejected. mpm liveness was, and remains, asserted exclusively by
`verify_daemon_liveness`, which runs after `tctl start --json`, probes `/health` directly,
is unconditional in all three scenarios, and fails closed (`die 60`).

The scoping is clean — the `false` and `true` branches are untouched. The formal set
notation in `02-harness-contracts.md` was updated in lockstep.

## Follow-ups, not in scope

- #4847 stays open. It asks whether an undetermined health should advance a green verdict —
  a policy question this change does not resolve, even though it removes the only member
  that produced `unknown` in practice.
- mpm's `/health` envelope carries no `version` field (`probe_http.rs:164-165`), so mpm
  reports `healthy` and can never report `stale`. Version-floor detection does not apply
  to it.
- Post-widening, `verify_stack_doctor` cannot distinguish a serving mpm from a dead one,
  leaving `verify_daemon_liveness` as the sole assertion of mpm liveness. Not a regression —
  it already was — but worth recording.


bobmatnyc:
The diagnosis here was right. #4925 is a real defect — `tctl` reported trusty-mpm as `unknown` instead of probing it — and this PR correctly identified the `ManageStrategy::OwnVerb` match arm in `probe_member_health` as the cause.

What changed is the architecture around it, and it changed after this PR was opened. This PR opened 2026-08-05. ADR-0032 was accepted 2026-08-07 (owner ruling, supersedes ADR-0018, ratifies ADR-0031): no trusty-* service runs its own HTTP daemon, so CLI health commands doing a loopback HTTP GET must move off that transport. ADR-0035 (PR #5233, still open) then settles the replacement as a console-side UDS aggregator rather than a per-CLI HTTP or UDS dial.

So this is closing as superseded-by-architecture, not as wrong. Nothing in it was a mistake at the time it was written.

#4925 remains open and is not resolved by this closure. #5232 is where the fix now lives, tracking the console-aggregator approach.

Thanks for finding and diagnosing this.

Issue #4904, opened 2026-08-05
hook recall precision is poor — 0 of 17 curated facts matched across 1,114 firings


bobmatnyc:
## Defect

The `UserPromptSubmit` hook (`trusty-memory prompt-context`) injects ~1,211 tokens of palace recall on EVERY prompt, but the recalled drawers are frequently not the relevant ones. Across 1,114 real firings over five days, 0 of 17 curated facts were cleanly matched.

🔴 **THIS IS NOT [#4836](https://github.com/bobmatnyc/trusty-tools/issues/4836).** State that explicitly. [#4836](https://github.com/bobmatnyc/trusty-tools/issues/4836) is a query-DROP on the MCP tool path — a readiness gate routing into a degraded branch that ignores the query entirely, fixed in [PR #4903](https://github.com/bobmatnyc/trusty-tools/pull/4903). This measurement came from the HTTP hook path, which was NEVER broken: it produced 148 distinct injection variants across those 1,114 firings, proving it discriminated by query throughout. Fixing the query-drop will not move this number. This is a RANKING and RELEVANCE problem.

## Evidence

Measured 2026-08-05 from the trusty-memory enriched-prompts JSONL log (one line per firing, carrying the injection text): 116 firings in a single morning, 1,114 over five days, average injection 1,211 tokens (median 1,252, range 693-1,438, capped near 4,096 chars). Cost context: a 20-turn session pays ~24,000 tokens in recall injections — more than half a session's entire one-time startup budget, recurring, for results that miss curated facts.

## Candidate areas

From the [#4836](https://github.com/bobmatnyc/trusty-tools/issues/4836) engineer's reading — leads, not conclusions:

- `expand_query` ([crates/trusty-memory/src/.../layers.rs:403](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-memory/src/layers.rs#L403)) hardcodes FOUR trigger groups — speed, vector search, memory safety, concurrency — and is a no-op for anything else. That shape matches the observed symptom of standing behavioral preferences never surfacing.
- L1 importance weighting, `L1_NO_SIMILARITY_PENALTY`, and the closet tag boost are the other ranking inputs worth examining.

## Resolution

[...]

Issue #4898, opened 2026-08-05
memory secret-scanner rejects ordinary prose and branch names as credentials


bobmatnyc:
## Defect

`PalaceHandle::remember_with_options`'s secret/credential heuristic rejects plain text that is not a credential, and the rejection is unbypassable by `force=true` (by design, per #2520). Legitimate memory writes fail with no way through except editing the content.

## Evidence — three observed false positives on 2026-08-05, one session

1. `tcode-goal-harness-around-mpm-model.md` — rejected on `PM+i…(25 chars)`. The flagged string is `PM+instructions+subagents`, from the phrase "harness around the mpm PM+instructions+subagents working model".

2. `tcode-streaming-epic-3696.md` — rejected on `fix-…(25 chars)`. The flagged string is `fix-3696-slice1-gapA-emit` — a git BRANCH NAME.

3. A PM-authored drawer rejected on a `…(4 chars)` token. A four-character string cannot be a credential; the minimum-length floor is either absent or wrong.

The heuristic appears to fire on `+`-joined or `-`-joined mixed-case alphanumeric runs, which describes ordinary technical prose, branch names, and issue slugs — exactly the content a project memory palace is FOR.

## Impact

These two files are the last un-landed items of an owner-authorized static-memory migration (26 of 28 imported successfully; palace now at 1239 drawers). The migration cannot complete while ordinary prose is rejected. Working around it by redacting source content is not acceptable — the owner's standing rule for this migration is that body text is preserved VERBATIM.

## Resolution — options, for the implementer to weigh

- Raise the minimum token length (a 4-char match is never a credential) and require a higher entropy threshold.
- Exclude strings that match known-benign shapes: git branch/ref names, issue slugs, `+`/`-`-joined dictionary words.
- Provide a scoped, auditable override for the IMPORT path specifically — note that #2520 deliberately made `force=true` NOT bypass secret detection, so any override must be a considered change to that decision, not an accidental reversal of it. Say so explicitly in the body.

Note in the body that the scanner working correctly matters — it caught a real leaked token in the same session — so the fix is precision, not removal.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4884, opened 2026-08-05
feat(trusty-common): add Drawer.fact_key and the DRAWERS_BY_FACT_KEY secondary index


bobmatnyc:
Add `pub fact_key: Option<String>` to `Drawer` (crates/trusty-common/src/memory_core/palace.rs:178-209) and to `DrawerRecord` (crates/trusty-common/src/memory_core/store/kg_store.rs:153-180) plus the serde-default migration fallback chain (crates/trusty-common/src/memory_core/store/kg_redb/types.rs:52-146).

Add a `DRAWERS_BY_FACT_KEY` redb secondary table alongside `DRAWERS` (kg_store.rs:57), following the existing `TRIPLES_BY_OBJECT`/`TRIPLES_BY_PREDICATE` precedent (kg_store.rs:27-37).

Purely additive storage groundwork — no consumers yet. No dependencies. Runs parallel with T2/T5.

Part of epic: implement ADR-0028 — three-tier memory recall (Standing/Current/Episodic)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #4975, opened 2026-08-05
fix(trusty-common): stop deferred embedding from dropping failures silently (#4906)

### LINKED ISSUE #4906: deferred embedding drops silently — 39 drawers are durable but permanently unfindable
## P1 — data-loss trap

**DEFECT.** `spawn_deferred_embed` is fire-and-forget. Every failure path is `warn!` + `return` — no retry, no queue, no marker, no record that it happened. The drawer remains durable in KG/redb and permanently invisible to vector recall. Nothing ever re-attempts a dropped embed, so a restart does not repair it. `defer_embedding` (`crates/trusty-memory/src/tools/memory_ops.rs:42` and `:185`, `crates/trusty-memory/src/tools/task_ops.rs:41`) reads the raw readiness latch, so a daemon stranded in `Warming` (the #4836 bug, fixed by PR #4903) routes EVERY MCP write through this fire-and-forget path.

This is the repo's recurring fail-open shape: a failure branch that leaves state looking successful.

## Evidence

Measured live 2026-08-05 against the `trusty-tools` palace:
- `drawer_count: 1241`, `vector_count: 1202` → **39 drawers (3.1%) have no vector**. This gap is the exact, cheap detector.
- A 71-drawer sample, self-retrieving each drawer with its own verbatim text at `top_k=150`: **9 were not retrievable at all**.
- One of those, drawer `1509bc35`, was written `2026-08-05T12:27:00.278Z` — INSIDE an owner-authorized memory-migration batch. 13 of the 14 checked migration drawers are embedded; that one is not, and there is no way to tell which without checking.

**Caveat:** 9/71 is a sample and `top_k=150` self-retrieval is a proxy (a drawer could be embedded yet rank below 150). The hard number is the 1241 − 1202 = 39 gap; the sample is evidence the gap spans history and reaches recent writes, not a precise rate.

## Critical context — why P1

[#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834) plans to delete 149 static memory files after migrating them into the palace. Had deletion proceeded on "the import reported success", any dropped-embed drawer's content would be durable but permanently unfindable, with the only always-loaded copy deleted. Written-but-unembedded is indistinguishable from written-and-findable without an explicit check.

## Resolution — three parts

1. **Durable work-tracking for embeds.** A dropped embed must be recorded and retried (a queue, a pending marker, or startup reconciliation) rather than lost to a `warn!`.
2. **A data-repair backfill** over existing vector-less drawers. Required BEFORE [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834) deletes anything — [PR #4903](https://github.com/bobmatnyc/trusty-tools/pull/4903) fixes the latch but repairs nothing already written.
3. **A `doctor` check** on `vector_count` vs `drawer_count`, so the gap is visible without a manual probe.

## Distinction from related issues

This is distinct from [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904) (recall ranking/precision). A drawer that is embedded but ranks poorly is [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904); a drawer with no vector at all is this. One sampled drawer (`5dfc4721`) initially looked missing at `top_k=8` and is in fact embedded at rank 13 — that one is [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904), not this.

**Gates:** [#4834](https://github.com/bobmatnyc/trusty-tools/issues/4834) (migration delete-phase)

bobmatnyc:
## The defect

`spawn_deferred_embed` was fire-and-forget. Every failure path was a `warn!`
followed by a bare `return` — no retry, no marker, no surfaced error. One
embedder blip and the drawer stayed durable in redb and permanently invisible to
vector recall. `memory_recall` cannot detect this class of failure, because it
only searches what got embedded.

**Measured live against the running daemon while writing this PR**
(`GET /api/v1/palaces`, the same surface the issue used):

| palace | drawers | vectors | gap |
|---|---|---|---|
| code-intelligence | 224 | 179 | **45** |
| trusty-tools | 1275 | 1236 | **39** |
| cto-reports | 133 | 98 | **35** |
| apex | 111 | 93 | **18** |

137 across the 4 palaces the daemon currently holds open, out of 93 total. The
issue's 39 reproduces exactly; it is not the whole estate.

## What changed

**1. Failures stop being silent.** `deferred_embed.rs` retries transient
failures with bounded exponential backoff (3 attempts, 250 ms base) instead of
taking the first `Err` as final, and retries the vector upsert on the same
policy. A final failure writes a durable row to the palace's
`embed_failures.json` ledger — drawer id, timestamp, attempt count, and the last
error text — so the loss outlives the log line. It is still logged, at `error!`
rather than `warn!`.

**2. "No embedder" is not "this drawer is broken".** `EmbedLoss` splits
`EmbedderUnavailable` from `Embed`/`Upsert`. Only the latter marks a drawer, so a
host with no model downloaded, or a cold start, does not get every drawer
flagged. The health surface reports `embedder_ready` alongside the shortfall so
an unexplained gap and an expected one read differently.

**3. The condition is queryable.** `PalaceHandle::embed_health()` set-differences
the drawer table against the vector index. That is exact — a drawer id absent
from the index has no vector by definition.

On PR #4903's suggested `vector_count == drawer_count` check: it is the right
**summary** and it is how the numbers above were measured, but it is not
sufficient on its own, because the vector index can also hold **orphans**
(vectors whose drawer was forgotten), so equal counts do not prove full
coverage. The count comparison is the alarm; the set difference is what the
repair acts on. Both are now available — the counts already ship on
`/api/v1/palaces`, and the exact set comes back from `palace_reembed`.

**4. A repair path for what is already broken.**
`PalaceHandle::backfill_missing_vectors()` re-embeds drawers with no vector,
through the same `embed_and_store` primitive the write path uses. Idempotent
(a repaired drawer is no longer in the missing set), and a genuine no-op on a
healthy palace — it short-circuits before resolving an embedder, so it is safe
on a host with no model. Exposed as the `palace_reembed` MCP tool, defaulting to
`dry_run: true`. It lives in the daemon on purpose: the daemon holds the
palace's writer lock, so a CLI would only ever get a read-only snapshot it
cannot write vectors into.

## Constraints honoured

- **The write path is still asynchronous.** Storing a drawer must not fail or
  hang because the embedder is slow or down (#1970). The fix is to stop LOSING
  the failure, not to make writes synchronous.
- **The `force` / `allow_secret_like` gate at `handle.rs:562-576` is untouched.**
- `filter.rs` and `retrieval/layers.rs` are untouched (owned by #4898 / #4904).
  No crate version bumped.

## Design note — why the marker is a sidecar, not a `Drawer` field

The obvious shape is a field on `Drawer`. Two things argued against it. The
drawer row is postcard-encoded and **positional**, so a new field means a fourth
migration-fallback shape in `decode_drawer_record`; get that chain wrong and
every row silently loses `fact_key` — a worse data bug than the one being fixed,
in a chain #4884/#4886 churned days ago. And `Drawer` is built as a struct
literal in `retrieval/layers.rs`, which #4904 owns this session.

`<data_dir>/embed_failures.json` is durable, atomic (tmp + rename, the `L1Cache`
#154 pattern), needs no schema migration, and cannot drift in the harmful
direction: the authoritative answer is always the vector-index set difference,
and the ledger is annotation on top of it. Rows for repaired drawers are cleared;
rows for drawers that no longer exist are never written.

## Test ladder — RUNG 5

Persistence and durability, shared library, silent-data-loss class.

```
cargo fmt --check                                                                  EXIT=0
cargo clippy -p trusty-common --features memory-core,embedder-test-support \
             --all-targets -- -D warnings                                          EXIT=0
cargo test -p trusty-common --features memory-core,embedder-test-support \
             -- --test-threads=1
    test result: ok. 878 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out
    test result: ok. 6 passed; 0 failed; ...
    test result: ok. 11 passed; 0 failed; ...
    test result: ok. 4 passed; 0 failed; ...
    test result: ok. 3 passed; 0 failed; ...
cargo test -p trusty-memory -- --test-threads=1
    test result: FAILED. 561 passed; 2 failed; 4 ignored
cargo test -p trusty-mpm
    test result: FAILED. 4635 passed; 1 failed; 7 ignored
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui         EXIT=0
    line-cap: measured 3720 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

### The 3 red tests are pre-existing and environmental

Not "re-run until green" — each is provably outside this diff.

**`commands::doctor::tests::check_daemon_health_*` (2)** —
`git diff --name-only origin/main...HEAD -- crates/trusty-memory/src/commands/`
returns **0 files**. Both probe `127.0.0.1:7070/health`; the daemon on this
machine answers `"daemon_state":"warming"` (the #4836 latch), which
`interpret_health_body` maps to `Warn`, and both tests assert
`Fail | Pass | Unknown`. Any checkout of any commit fails these two while that
daemon is warming.

**`daemon::managed_routes::tests::stale_assets_for_many_...` (1)** —
`git diff --name-only origin/main...HEAD -- crates/trusty-mpm/` returns
**0 files**. This PR does not touch `trusty-mpm` at all.

### New coverage — fail-before / pass-after

11 tests in `retrieval/embed_repair_tests.rs`, all asserting on state that
survives the process (the ledger file, the vector index), never on log output:

- `retry_succeeds_after_transient_failures` — runs the SAME flaky embedder under
  a 1-attempt policy (reproducing the pre-fix behaviour: fails, no vector) and
  then a 3-attempt policy (succeeds, vector lands).
- `permanent_failure_writes_a_ledger_row` — a dead embedder; asserts the ledger
  re-read from disk carries the row, the attempt count, and the failing stage.
- `missing_embedder_does_not_mark_the_drawer` — an `EmbedderUnavailable` loss
  writes nothing; a drawer-specific loss for the same drawer does.
- `loss_for_a_forgotten_drawer_is_not_recorded`
- `backfill_reembeds_a_marked_drawer` — asserts the drawer is NOT retrievable via
  `retrieve_l2` before, and IS after, and that the ledger row is cleared; then
  re-runs and asserts the second pass does nothing.
- `backfill_is_a_noop_on_a_healthy_palace`, `backfill_dry_run_repairs_nothing`
- `health_reports_the_drawer_with_no_vector`
- 3 ledger tests (upsert-by-id, surgical clear, missing file reads empty)
- `dispatch_palace_reembed_dry_run_reports_counts` in `trusty-memory`

## Follow-up not in this PR

A `trusty-memory doctor` check on the count gap. The detection logic now exists
(`embed_health()`) and the counts already ship on `/api/v1/palaces`, so the check
is a display change in the doctor orchestrator, not new logic. Left out to keep
this PR to the data-integrity fix; `doctor/mod.rs` is also at 475/500 SLOC, so
wiring it wants a small split first.

Closes #4906. Unblocks #4834 — the migration can now verify retrievability
before deleting a source file.

### DIFF
--- a/.test-pointer-allowlist.tsv
+++ b/.test-pointer-allowlist.tsv
@@ -210,7 +210,6 @@ crates/trusty-common/src/memory_core/dream/recall_benchmark.rs	dream_recall_benc
 crates/trusty-common/src/memory_core/registry.rs	evict_idle_skips_recently_accessed_handle
 crates/trusty-common/src/memory_core/registry.rs	open_hydrates_persisted_palaces
 crates/trusty-common/src/memory_core/retrieval/handle.rs	decay_applied_in_l2_score
-crates/trusty-common/src/memory_core/retrieval/handle.rs	deferred_embed_backfills_vector_once_embedder_ready
 crates/trusty-common/src/memory_core/retrieval/handle.rs	palace_handle_read_only_when_locked_by_another_process
 crates/trusty-common/src/memory_core/retrieval/handle.rs	remember_classifies_session_events
 crates/trusty-common/src/memory_core/semantic_consolidation/inference.rs	ollama_inference_new_stores_fields
--- a/crates/trusty-common/changelog.d/4906-silent-embed-failures.md
+++ b/crates/trusty-common/changelog.d/4906-silent-embed-failures.md
@@ -2,7 +2,7 @@ Fixed
 
 - deferred embedding no longer drops failures silently — a drawer can no longer be stored, durable, and permanently unfindable (closes [#4906](https://github.com/bobmatnyc/trusty-tools/issues/4906))
   - the background embed lane retries transient failures with bounded exponential backoff instead of giving up on the first error
-  - a final failure writes a durable row to the palace's `embed_failures.json` ledger, so the loss outlives the `warn!` that used to be its only trace
+  - a final failure writes a durable row to the palace's `embed_failures.json` ledger, so the loss outlives the `warn!` that used to be its only trace — serialised through `json_rmw`, so a burst of concurrent failures keeps every row instead of only the last writer's
   - "no embedder on this host" is separated from "the embedder is here and this drawer failed"; only the second marks a drawer, so a machine with no model downloaded is not reported as having thousands of broken drawers
   - new `PalaceHandle::embed_health()` answers "which drawers have no vector" by set-differencing the drawer table against the vector index, replacing a self-retrieval guess
   - new `PalaceHandle::backfill_missing_vectors()` re-embeds drawers that already lack a vector — idempotent, safe to re-run, and a no-op on a healthy palace (it does not even resolve an embedder when there is nothing to repair)
--- a/crates/trusty-common/changelog.d/4906-silent-embed-failures.md
+++ b/crates/trusty-common/changelog.d/4906-silent-embed-failures.md
@@ -1,0 +1,8 @@
+Fixed
+
+- deferred embedding no longer drops failures silently — a drawer can no longer be stored, durable, and permanently unfindable (closes [#4906](https://github.com/bobmatnyc/trusty-tools/issues/4906))
+  - the background embed lane retries transient failures with bounded exponential backoff instead of giving up on the first error
+  - a final failure writes a durable row to the palace's `embed_failures.json` ledger, so the loss outlives the `warn!` that used to be its only trace
+  - "no embedder on this host" is separated from "the embedder is here and this drawer failed"; only the second marks a drawer, so a machine with no model downloaded is not reported as having thousands of broken drawers
+  - new `PalaceHandle::embed_health()` answers "which drawers have no vector" by set-differencing the drawer table against the vector index, replacing a self-retrieval guess
+  - new `PalaceHandle::backfill_missing_vectors()` re-embeds drawers that already lack a vector — idempotent, safe to re-run, and a no-op on a healthy palace (it does not even resolve an embedder when there is nothing to repair)
--- a/crates/trusty-common/src/json_rmw.rs
+++ b/crates/trusty-common/src/json_rmw.rs
@@ -30,9 +30,14 @@
 //!    serialises separate processes AND separate threads that each call
 //!    [`update`], on Unix and Windows alike.
 //! 2. **All-or-nothing publish.** Readers of `path` see either the complete
-//!    previous document or the complete new one, never a partial write. The
-//!    temp file carries the writer's pid and a nanosecond stamp, so concurrent
-//!    writers never share a scratch path even if one of them bypasses the lock.
+//!    previous document or the complete new one, never a partial write: the
+//!    document is built in a temp file and moved into place with `rename`.
+//!    That is the whole of this guarantee — the temp name (pid plus a
+//!    nanosecond stamp) is scratch-path hygiene, NOT a second line of defence
+//!    for a writer that bypasses the lock. An earlier version of this comment
+//!    claimed it was; #4906's review falsified that experimentally, with 16
+//!    threads landing on a single nanosecond value and colliding. Only
+//!    guarantee 1 keeps concurrent writers apart.
 //! 3. **Never fail open.** Every failure — lock acquisition, read, parse,
 //!    serialise, write, rename — returns `Err` and leaves `path` byte-for-byte
 //!    unchanged. There is no path on which a failed update advances state, and
--- a/crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs
+++ b/crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs
@@ -254,7 +254,7 @@ pub(super) async fn run_deferred_embed(
     let embedder = match shared_embedder().await {
         Ok(e) => e,
         Err(e) => {
-            record_loss(&ctx, id, &EmbedLoss::EmbedderUnavailable(format!("{e:#}")));
+            record_loss(&ctx, id, &EmbedLoss::EmbedderUnavailable(format!("{e:#}"))).await;
             return;
         }
     };
@@ -286,9 +286,9 @@ pub(super) async fn embed_store_or_record(
             );
             // A drawer that has just been embedded is no longer missing; drop
             // any stale row so the ledger cannot outlive the condition.
-            clear_ledger(ctx, id);
+            clear_ledger(ctx, id).await;
         }
-        Err(loss) => record_loss(ctx, id, &loss),
+        Err(loss) => record_loss(ctx, id, &loss).await,
     }
 }
 
@@ -311,10 +311,12 @@ pub(super) fn spawn(ctx: DeferredEmbedCtx, id: Uuid, content: String) {
 /// and a log line nobody keeps are complementary failures, not alternatives.
 /// What: skips the ledger entirely for a host-level embedder outage; otherwise
 /// writes one row keyed by drawer id, but only if the drawer still exists (a
-/// drawer forgotten mid-flight has nothing to annotate).
+/// drawer forgotten mid-flight has nothing to annotate). The write runs on
+/// `spawn_blocking` because `json_rmw::update` blocks the calling thread on an
+/// advisory `flock`.
 /// Test: `permanent_failure_writes_a_ledger_row`,
 /// `missing_embedder_does_not_mark_the_drawer`.
-pub(super) fn record_loss(ctx: &DeferredEmbedCtx, id: Uuid, loss: &EmbedLoss) {
+pub(super) async fn record_loss(ctx: &DeferredEmbedCtx, id: Uuid, loss: &EmbedLoss) {
     let reason = loss.reason();
     if !loss.is_drawer_specific() {
         // Expected on a host with no model downloaded, and on a cold start
@@ -353,21 +355,40 @@ pub(super) fn record_loss(ctx: &DeferredEmbedCtx, id: Uuid, loss: &EmbedLoss) {
         attempts: loss.attempts(),
         reason,
     };
-    if let Err(e) = embed_ledger::record(data_dir, entry) {
-        tracing::error!(
+    let dir = data_dir.clone();
+    let written = tokio::task::spawn_blocking(move || embed_ledger::record(&dir, entry)).await;
+    match written {
+        Ok(Ok(())) => {}
+        Ok(Err(e)) => tracing::error!(
             palace = %ctx.palace_id, drawer = %id,
             "#4906: could not write the embed-failure ledger: {e:#}"
-        );
+        ),
+        Err(e) => tracing::error!(
+            palace = %ctx.palace_id, drawer = %id,
+            "#4906: embed-failure ledger write task failed: {e}"
+        ),
     }
 }
 
 /// Drop one drawer's ledger row after a successful embed.
-fn clear_ledger(ctx: &DeferredEmbedCtx, id: Uuid) {
+///
+/// Why: a repaired drawer must stop being reported as broken, or the ledger
+/// outlives the condition it describes.
+/// What: a single-id [`embed_ledger::clear`] on `spawn_blocking` (same `flock`
+/// blocking contract as [`record_loss`]). A palace with no ledger short-circuits
+/// inside `clear` without creating one.
+/// Test: `backfill_reembeds_a_marked_drawer`.
+async fn clear_ledger(ctx: &DeferredEmbedCtx, id: Uuid) {
     let Some(data_dir) = ctx.data_dir.as_ref() else {
         return;
     };
-    let ids: std::collections::HashSet<Uuid> = std::iter::once(id).collect();
-    if let Err(e) = embed_ledger::clear(data_dir, &ids) {
+    let dir = data_dir.clone();
+    let cleared = tokio::task::spawn_blocking(move || {
+        let ids: std::collections::HashSet<Uuid> = std::iter::once(id).collect();
+        embed_ledger::clear(&dir, &ids)
+    })
+    .await;
+    if let Ok(Err(e)) = cleared {
         tracing::warn!(palace = %ctx.palace_id, drawer = %id, "#4906: ledger clear failed: {e:#}");
     }
 }
--- a/crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs
+++ b/crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs
@@ -1,0 +1,373 @@
+//! The deferred (background) embed lane: retry, then record what was lost.
+//!
+//! Why (#4906): `PalaceHandle::spawn_deferred_embed` was fire-and-forget. Every
+//! failure path was a `warn!` followed by a bare `return` — one embedder blip
+//! and the drawer stayed durable in redb and permanently invisible to vector
+//! recall, with nothing anywhere that could later say so. Measured on the live
+//! `trusty-tools` palace: 39 of 1,241 drawers had no vector, one of them written
+//! during an in-flight migration. `memory_recall` cannot detect this class of
+//! failure, because it only searches what got embedded.
+//!
+//! What: the same background lane, with three changes. A transient failure is
+//! retried with exponential backoff instead of being dropped on the first
+//! error. A final failure writes a durable row to the palace's embed-failure
+//! ledger. And "no embedder on this host" is separated from "the embedder is
+//! here and this drawer failed" — the first is an expected state on a machine
+//! with no model downloaded and must not mark every drawer broken.
+//!
+//! The write path stays asynchronous on purpose. Storing a drawer must not fail
+//! or hang because the embedder is slow or down (issue #1970); the fix is to
+//! stop LOSING the failure, not to make writes synchronous.
+//!
+//! Test: `retry_succeeds_after_transient_failures`,
+//! `permanent_failure_writes_a_ledger_row`,
+//! `missing_embedder_does_not_mark_the_drawer`.
+
+use super::embedder::shared_embedder;
+use crate::memory_core::embed::Embedder;
+use crate::memory_core::palace::{Drawer, PalaceId};
+use crate::memory_core::store::embed_ledger::{self, EmbedFailure};
+use crate::memory_core::store::vector::{UsearchStore, VectorStore};
+use crate::memory_core::timeouts;
+use parking_lot::RwLock;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use uuid::Uuid;
+
+/// Attempts made before a deferred embed is declared lost.
+///
+/// Why: an ONNX session under memory pressure, or a vector-store write racing a
+/// redb commit, fails once and succeeds immediately after. Three attempts cover
+/// that without turning a genuinely dead embedder into a long stall on a
+/// background task.
+const DEFAULT_EMBED_ATTEMPTS: u32 = 3;
+
+/// First backoff step; doubles per attempt (250 ms, 500 ms).
+const DEFAULT_EMBED_RETRY_BASE_MS: u64 = 250;
+
+/// Bounded exponential-backoff policy for one embed job.
+///
+/// Why: the retry has to be a value rather than a constant so the backfill can
+/// run a longer policy than the write-path lane, and so tests can collapse the
+/// delays to zero instead of sleeping for real.
+/// What: `attempts` total tries (not retries — `attempts: 1` means no retry) and
+/// a base delay doubled on each successive failure.
+/// Test: `retry_succeeds_after_transient_failures`.
+#[derive(Debug, Clone, Copy)]
+pub struct RetryPolicy {
+    pub attempts: u32,
+    pub base_delay: Duration,
+}
+
+impl Default for RetryPolicy {
+    fn default() -> Self {
+        Self {
+            attempts: DEFAULT_EMBED_ATTEMPTS,
+            base_delay: Duration::from_millis(DEFAULT_EMBED_RETRY_BASE_MS),
+        }
+    }
+}
+
+impl RetryPolicy {
+    /// Delay to wait after the `attempt`-th failure (1-based).
+    fn delay_for(&self, attempt: u32) -> Duration {
+        let shift = attempt.saturating_sub(1).min(6);
+        self.base_delay.saturating_mul(1u32 << shift)
+    }
+
+    /// Policy with no sleeping, for tests that only care about attempt counts.
+    #[cfg(any(test, feature = "embedder-test-support"))]
+    pub fn instant(attempts: u32) -> Self {
+        Self {
+            attempts,
+            base_delay: Duration::ZERO,
+        }
+    }
+}
+
+/// Why a drawer ended up without a vector.
+///
+/// Why: the two cases demand opposite responses. `EmbedderUnavailable` is an
+/// expected state on a host with no model downloaded — marking every drawer
+/// broken there would be a false alarm at estate scale. Everything else means
+/// the embedder was present and this particular drawer failed, which is
+/// precisely what the ledger exists to record.
+/// What: a three-way split carrying the last error text.
+/// Test: `missing_embedder_does_not_mark_the_drawer`.
+#[derive(Debug, Clone)]
+pub enum EmbedLoss {
+    /// No embedder could be initialised on this host.
+    EmbedderUnavailable(String),
+    /// The embedder was present; producing the vector failed.
+    Embed { reason: String, attempts: u32 },
+    /// The vector was produced but could not be stored.
+    Upsert { reason: String, attempts: u32 },
+}
+
+impl EmbedLoss {
+    /// Whether this loss is the drawer's fault rather than the host's.
+    ///
+    /// Why: only a host-independent failure justifies a durable per-drawer
+    /// marker. See the type docs.
+    /// What: `false` for [`EmbedLoss::EmbedderUnavailable`], `true` otherwise.
+    /// Test: `missing_embedder_does_not_mark_the_drawer`.
+    pub fn is_drawer_specific(&self) -> bool {
+        !matches!(self, EmbedLoss::EmbedderUnavailable(_))
+    }
+
+    /// How many attempts were made before giving up (0 when we never got an
+    /// embedder to try with).
+    pub(super) fn attempts(&self) -> u32 {
+        match self {
+            EmbedLoss::EmbedderUnavailable(_) => 0,
+            EmbedLoss::Embed { attempts, .. } | EmbedLoss::Upsert { attempts, .. } => *attempts,
+        }
+    }
+
+    /// Human-readable failure text, naming the stage that failed.
+    pub(super) fn reason(&self) -> String {
+        match self {
+            EmbedLoss::EmbedderUnavailable(r) => format!("embedder unavailable: {r}"),
+            EmbedLoss::Embed { reason, .. } => format!("embed failed: {reason}"),
+            EmbedLoss::Upsert { reason, .. } => format!("vector upsert failed: {reason}"),
+        }
+    }
+}
+
+/// Run `op` up to `policy.attempts` times, sleeping between failures.
+///
+/// Why: embed and upsert both need the same bounded-retry shape, and writing it
+/// twice is how the two drift into disagreeing about what "transient" means.
+/// What: returns the first `Ok`, or the LAST error text once attempts are
+/// exhausted. Also returns how many attempts were actually made, which is what
+/// the ledger records.
+/// Test: `retry_succeeds_after_transient_failures`.
+pub(super) async fn with_retry<T, F, Fut>(
+    policy: &RetryPolicy,
+    mut op: F,
+) -> Result<(T, u32), (String, u32)>
+where
+    F: FnMut(u32) -> Fut,
+    Fut: std::future::Future<Output = Result<T, String>>,
+{
+    let total = policy.attempts.max(1);
+    let mut last = String::from("no attempt was made");
+    for attempt in 1..=total {
+        match op(attempt).await {
+            Ok(v) => return Ok((v, attempt)),
+            Err(e) => last = e,
+        }
+        if attempt < total {
+            tokio::time::sleep(policy.delay_for(attempt)).await;
+        }
+    }
+    Err((last, total))
+}
+
+/// Embed `content` and store the vector under `id`, retrying transient failures.
+///
+/// Why: this is the single embed+store primitive shared by the write-path
+/// deferred lane and the repair backfill, so a drawer recovered by the backfill
+/// is embedded by exactly the code that would have embedded it at write time.
+/// What: retries `embed_batch` (each attempt under the existing
+/// `TRUSTY_EMBED_BATCH_TIMEOUT_SECS` bound), then retries the vector upsert.
+/// Returns the failing stage in [`EmbedLoss`] so the caller can decide whether
+/// it warrants a durable marker.
+/// Test: `retry_succeeds_after_transient_failures`,
+/// `permanent_failure_writes_a_ledger_row`.
+pub(super) async fn embed_and_store(
+    embedder: &Arc<dyn Embedder + Send + Sync>,
+    vector_store: &Arc<UsearchStore>,
+    id: Uuid,
+    content: &str,
+    policy: &RetryPolicy,
+) -> Result<u32, EmbedLoss> {
+    let batch = [content.to_string()];
+    let embed_timeout = timeouts::embed_batch_timeout();
+    let (vector, embed_attempts) = with_retry(policy, |_| {
+        let embedder = embedder.clone();
+        let batch = &batch;
+        async move {
+            match tokio::time::timeout(embed_timeout, embedder.embed_batch(batch)).await {
+                Ok(Ok(v)) => v
+                    .into_iter()
+                    .next()
+                    .ok_or_else(|| "embedder returned an empty batch".to_string()),
+                Ok(Err(e)) => Err(format!("{e:#}")),
+                Err(_) => Err(format!("embed_batch timed out after {embed_timeout:?}")),
+            }
+        }
+    })
+    .await
+    .map_err(|(reason, attempts)| EmbedLoss::Embed { reason, attempts })?;
+
+    let (_, upsert_attempts) = with_retry(policy, |_| {
+        let vector = vector.clone();
+        let vector_store = vector_store.clone();
+        async move {
+            vector_store
+                .upsert(id, vector)
+                .await
+                .map_err(|e| format!("{e:#}"))
+        }
+    })
+    .await
+    .map_err(|(reason, attempts)| EmbedLoss::Upsert { reason, attempts })?;
+
+    Ok(embed_attempts.max(upsert_attempts))
+}
+
+/// The slice of `PalaceHandle` the background lane needs.
+///
+/// Why: the spawned task outlives the `&self` borrow it was created from, and
+/// cloning the individual `Arc`s is cheaper (and clearer about what the lane
+/// touches) than requiring callers to hold an `Arc<PalaceHandle>`.
+/// What: palace id, vector store, the in-memory drawer table, and the data dir
+/// the ledger lives in. `data_dir` is `None` for in-memory test handles, in
+/// which case there is nowhere durable to record a failure and the lane says so
+/// in the log instead.
+/// Test: exercised by every test in `embed_repair_tests`.
+#[derive(Clone)]
+pub(super) struct DeferredEmbedCtx {
+    pub palace_id: PalaceId,
+    pub vector_store: Arc<UsearchStore>,
+    pub drawers: Arc<RwLock<Vec<Drawer>>>,
+    pub data_dir: Option<PathBuf>,
+}
+
+/// Body of the background embed task, as an awaitable future.
+///
+/// Why: kept separate from the `tokio::spawn` wrapper so tests can await the
+/// whole job deterministically instead of racing a detached task.
+/// What: resolves the shared embedder, embeds and stores with retry, and on a
+/// drawer-specific failure records a ledger row. An unavailable embedder is
+/// logged and left unmarked — see [`EmbedLoss::is_drawer_specific`].
+/// Test: `permanent_failure_writes_a_ledger_row`,
+/// `missing_embedder_does_not_mark_the_drawer`.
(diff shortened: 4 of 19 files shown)


bobmatnyc:

Adversarial rung-5 review of `419bafcd` against #4906. Reviewed the code and the spec only.

Gates re-run independently in a clean worktree at the PR head:

```
cargo test -p trusty-common --features memory-core,embedder-test-support embed_repair -- --test-threads=1
    test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 880 filtered out
cargo test -p trusty-common --features memory-core,embedder-test-support --lib -- --test-threads=1
    test result: ok. 878 passed; 0 failed; 13 ignored; 0 measured
cargo clippy -p trusty-common --features memory-core,embedder-test-support --all-targets -- -D warnings   EXIT=0
bash scripts/check_line_cap.sh    3724 files, 0 violations — OK
```

No CRITICAL. Two HIGH, both races in the new durability path. Neither can make a drawer undetectable — the authoritative surface is `embed_health()`'s set difference against the vector index, which needs no ledger at all — so nothing here reintroduces the silent-loss class. Both lose the ledger annotation that this PR exists to add, in exactly the mass-failure scenario it was built for.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/memory_core/store/embed_ledger.rs` | 93-98, 109-123 | `record` and `clear` are unsynchronised load-modify-write. tmp+rename makes each *write* atomic; it does not serialise the read-modify-write, so concurrent writers lose each other's rows | Serialise the whole load-modify-save under a process-wide `parking_lot::Mutex` in this module (or a per-handle lock threaded through `DeferredEmbedCtx`) | Fix here |
| HIGH | `crates/trusty-common/src/memory_core/retrieval/handle.rs` / `deferred_embed.rs` | 649 vs 694 / 339-342 | `record_loss`'s `still_present` guard reads `ctx.drawers`, but `spawn_deferred_embed` fires at 649 and the drawer is not pushed until 694. Nothing orders them | Move `spawn_deferred_embed(id, content)` below the `drawers.push(drawer)` block at 695 | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/store/embed_ledger.rs` | 74-84, 93-98 | A malformed ledger reads as `[]`, and the next `record` overwrites the file — every prior row is destroyed without trace. Branch is untested | Rename the bad file to `embed_failures.json.corrupt.<ts>` before the first overwrite; add a test for the malformed branch | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs` | 317-331 | A permanently misconfigured embedder returns `EmbedderUnavailable` forever: drawers accumulate with no vector and no marker, and nothing alarms. Queryable, but only if someone asks | Ship the `trusty-memory doctor` count-gap check named in the PR body as follow-up; if it stays out, say so in the issue | Promote |
| MEDIUM | `crates/trusty-memory/src/tools/palace_ops.rs` | 278-297 | `handle_palace_reembed` calls `embed_health()` twice per invocation (once inside `backfill_missing_vectors`, once after) — two full `VECTOR_KEYS` scans plus two whole-table walks. And `limit: None` means one MCP call can run an unbounded repair inside the live daemon | Return the health snapshot from `backfill_missing_vectors`; default `limit` to a bounded chunk when `dry_run: false` | Fix here |
| LOW | `crates/trusty-common/src/memory_core/retrieval/deferred_embed.rs` | 356, 370 | Blocking `std::fs` read+write inside a tokio task | `spawn_blocking`, or accept it and note why | Parent |
| LOW | `crates/trusty-common/src/memory_core/retrieval/embed_repair.rs` | 224-233 | The comment says re-reading content under the lock stops a concurrent `forget` creating an orphan. The re-read happens before the embed; the upsert lands seconds later, so the window is still open | Reword the comment to say it narrows the window, not closes it | Fix here |
| LOW | `crates/trusty-common/src/memory_core/retrieval/embed_repair.rs` | 231-233, 256 | A drawer forgotten mid-run `continue`s but stays in `still_missing_ids`, so the report names ids that no longer exist | Drop skipped ids from `still_missing_ids` | Parent |

## The two HIGH findings in detail

**Ledger lost update.** `record` is `load()` → `retain` → `push` → `save()`, with nothing between the load and the save. Two tasks that load the same file both write their own single-row result; the later rename wins and the earlier row is gone.

This is not a corner case, it is the design load. `spawn_deferred_embed` spawns one detached task per drawer. Under the default policy every task sleeps 250 ms + 500 ms before giving up, so a burst of writes against a broken embedder produces N tasks all reaching `record_loss` inside the same few milliseconds. A 20-drawer burst keeps a handful of the 20 rows.

The module doc states the opposite at lines 11-13: "written atomically (tmp + rename) with the same per-invocation tmp naming `L1Cache::save_l1_cache` uses so concurrent writers cannot trample each other." The per-invocation tmp name stops two writers deleting each other's temp file. It does nothing about lost updates. Whatever the fix, that sentence needs correcting.

Not CRITICAL because `embed_health()` still returns every affected drawer and `backfill_missing_vectors` still repairs it. What is lost is the attempt count and the error text — the "why" — for most of a mass failure.

**Spawn-before-push.** `record_loss` at `deferred_embed.rs:339` refuses to write a row when the drawer is absent from `ctx.drawers`. Correct intent: a forgotten drawer has nothing to annotate. But `spawn_deferred_embed` fires at `handle.rs:649`, and the drawer reaches `self.drawers` at `handle.rs:694`, after `tier_c::persist_with_retirement`. A deferred task that fails before that push sees an empty slot and writes nothing — for a drawer that is about to become durable and permanently vector-less.

The window needs the redb commit to exceed the ~750 ms of backoff sleep, so this fires rarely, not routinely. I am flagging it as HIGH on the structure rather than the frequency: a durability guard that depends on an ordering the code does not establish is the "fail-open / cursor-advance" shape this repo keeps hitting. The fix is to move the spawn below line 695 — no reason it has to be where it is, and `content` is already cloned.

## Attacks that did not land

Recording these so they are not re-run.

**"3 attempts at 250 ms is ~1.75 s, too short for a cold start."** Wrong. The cold start never reaches the retry loop. `shared_embedder()` is a `OnceCell::get_or_try_init` bounded at 180 s (`TRUSTY_EMBEDDER_INIT_TIMEOUT_SECS`), and it resolves before `embed_and_store` runs; concurrent first-users collapse onto that one init. Once resolved, each of the 3 attempts is separately bounded by `embed_batch_timeout()` (default 30 s), so a slow embedder gets up to ~90 s, not 1.75 s. The 1.75 s figure only describes an embedder that returns errors instantly, which is the case that *should* exhaust. No transient condition gets converted into a permanent ledger entry by the timing.

**`backfill_missing_vectors` short-circuits before resolving an embedder.** Holds, in every path. `embed_repair.rs:198` returns on `dry_run || missing.is_empty()`; the read-only bail is 202, `shared_embedder()` is not reached until 210. Both claims — no-op on a healthy palace, safe with no model — are true as written, and `backfill_is_a_noop_on_a_healthy_palace` proves the first.

**Orphans in the index.** Handled correctly. `all_ids()` reads `VECTOR_KEYS`, and `HnswStore::delete` removes the key row before tombstoning, so a deleted vector cannot make a drawer look covered. Orphans inflate `vector_count` but cannot shrink `missing_vector_ids` — the set difference is one-directional and the PR's reasoning about equal counts not proving coverage is right.

**`all_ids()` failing open.** `vector.rs:400-403` returns an empty `Vec` on a redb scan error. That makes `embed_health` report every drawer as missing — fails toward over-reporting and a redundant re-embed, never toward "everything is fine". `upsert` is idempotent, so the wasted run cannot corrupt anything. Safe direction.

**`embed_health()` at scale / lock contention.** The `drawers.read()` guard covers a filter+collect with no `.await` inside, and `all_ids()` runs before it is taken. `backfill_missing_vectors` re-reads per drawer in a scoped block that drops before the await. No guard is held across an await anywhere in the new code, and the write path is never blocked for longer than one table walk. Cost at 100k is one `VECTOR_KEYS` scan plus 100k UUID parses — real, but the reason to fix it is the double call in `palace_reembed` (MEDIUM above), not lock contention.

**Cross-process ledger writes.** Bounded in practice. Only the lock-holding process can write vectors (`backfill_missing_vectors` bails read-only at `embed_repair.rs:202-208`), and each palace has its own `data_dir`, so there is no cross-palace collision. Reads during a rename are atomic on POSIX. The lost-update problem is intra-process.

**Deleted ledger file.** `load` returns `[]`, health still reports the gap from the index, backfill still repairs. Degrades to "no annotation", not to "healthy".

## Tests

They prove the four claims rather than exercising the happy path. `retry_succeeds_after_transient_failures` runs the *same* flaky embedder under a 1-attempt policy first and asserts no vector — a genuine fail-before, not a retrofit. Every assertion is on state that survives the process (the ledger re-read from disk, `vector_store.all_ids()`), and `backfill_reembeds_a_marked_drawer` asserts through `retrieve_l2` on both sides, which is the behaviour a user would notice.

Two gaps: no test writes the ledger from concurrent tasks (which is why HIGH-1 survived), and `embed_ledger::load`'s malformed-JSON branch is untested — `embed_ledger_load_is_empty_when_absent` only covers the missing file.

**`dry_run: true` as the default** is adequate protection against accidental invocation, and `dispatch_palace_reembed_dry_run_reports_counts` proves the default holds through `dispatch_tool`. It is not adequate protection against a deliberate `dry_run: false` on a large palace, because there is no default `limit` — see the MEDIUM above.

1. Serialise `embed_ledger::record` / `clear` so concurrent tasks cannot lose rows, and correct the "concurrent writers cannot trample each other" claim at `embed_ledger.rs:11-13`. Add a test that drives N concurrent `record` calls and asserts N rows.
2. Move `spawn_deferred_embed` below the `drawers.push(drawer)` block at `handle.rs:695`.

## Notes

- The three red tests named in the PR body are outside the diff; I did not treat them as blocking.
- Reviewed at `419bafcd`. `filter.rs` and `layers.rs` are untouched, as claimed.

bobmatnyc:
Both HIGH findings fixed on the same branch. Head is now `1cb79b384d32284d397f1337d626853ff863563d`.

## HIGH-1 — ledger lost update

Confirmed, and worse than estimated. The new test records 16 distinct drawers from 16 barrier-released threads; with the unlocked read-modify-write in place it keeps **1 of 16**, not a handful.

The fix is not the per-directory mutex I first wrote. While proving the fail-before state I hit `json_rmw::tests::update_serialises_concurrent_threads` in this same crate — `trusty_common::json_rmw` is already this workspace's single entry point for a locked load → mutate → publish-atomically cycle, built for exactly this bug shape in `projects.json` and `tokens.json`. Writing a second implementation next to it would have been the duplicate-capability defect the repo's common-entry-point rule exists to stop.

So `record` and `clear` now route through `json_rmw::update`. Strictly stronger than the mutex: the advisory `flock` serialises separate **processes** as well as threads, it re-reads the document under the lock rather than trusting a caller's stale copy, and it fsyncs both file and parent directory. It also **deletes** the bespoke `save`/`TMP_COUNTER` atomic-write helper this module carried — net less code. Both call sites now run on `spawn_blocking`, per `json_rmw`'s documented blocking contract.

One deliberate semantics change falls out, and I want it on the record rather than discovered later: `json_rmw` never fails open, so a write over a ledger that exists but cannot be parsed now returns `Err` and leaves the bytes untouched, instead of replacing them with a fresh single-row document. `load` keeps the opposite position and degrades to empty. That asymmetry is intended — losing an annotation costs less than refusing access to the memories, and destroying one costs more than declining to write. `embed_ledger_load_degrades_to_empty_on_malformed_json` now pins both sides.

The module doc that claimed the per-invocation tmp name made concurrent writers safe is corrected. It stopped a torn FILE; it never touched a lost UPDATE. The two hazards are now named separately, with the reason the confusion is easy to make.

## HIGH-2 — spawn before push

Confirmed and fixed. The spawn moved below the redb commit and the in-memory push, so any failure arriving from that point on finds the drawer present and gets recorded. The synchronous branch now borrows `content` via `slice::from_ref` instead of moving it, since the deferred branch below still needs it. Inline comments at both sites explain the ordering constraint so it does not get moved back.

## Test gaps

- `concurrent_ledger_records_keep_every_row` — 16 threads, `std::sync::Barrier` so the load→save windows overlap. **Fail-before / pass-after verified by temporarily reverting the serialisation**: `got 1 of 16` red, all 16 green.
- `concurrent_clear_and_record_do_not_lose_each_other` — 8 clears racing 8 records; a repair backfill runs both against one palace, so the clear side needed its own coverage.
- `embed_ledger_load_degrades_to_empty_on_malformed_json` — the previously-untested branch, now asserting the read degrades, the write refuses, and the unparseable bytes survive.

No dedicated test for HIGH-2: its only observable effect is under a race a test cannot force without instrumentation hooks. The behaviour it protects — a failure for an absent drawer writes nothing — is already pinned by `loss_for_a_forgotten_drawer_is_not_recorded`; the fix makes the drawer present by construction.

## Rung-5 gate, re-run in full

```
cargo fmt --check                                                              EXIT=0
cargo clippy -p trusty-common --features memory-core,embedder-test-support
             --all-targets -- -D warnings                                      EXIT=0
cargo test -p trusty-common --features memory-core,embedder-test-support -- --test-threads=1
  test result: ok. 881 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out
  test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
cargo test -p trusty-memory -- --test-threads=1
  test result: FAILED. 561 passed; 2 failed; 4 ignored; 0 measured; 0 filtered out
cargo test -p trusty-mpm
  test result: FAILED. 4634 passed; 2 failed; 7 ignored; 0 measured; 0 filtered out
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui     EXIT=0
  line-cap: measured 3724 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
bash scripts/check_sld.sh                  → 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh   → both crates recorded
bash scripts/check_generation_artifacts.sh → 0 leaked artifacts
```

### The 4 red tests are outside this diff

`git diff --name-only origin/main...HEAD -- crates/trusty-mpm/` returns **0 files**; the same query for `crates/trusty-memory/src/commands/` returns **0 files**.

- `commands::doctor::tests::check_daemon_health_*` (2) — both probe `127.0.0.1:7070/health`, which on this machine answers `"daemon_state":"warming"` (the #4836 latch). `interpret_health_body` maps that to `Warn`; both tests assert `Fail | Pass | Unknown`. Deterministically red on any commit while that daemon is warming.
- `daemon::managed_routes::tests::stale_assets_for_many_...` and `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — in `trusty-mpm`, which this PR does not touch at all. The second was green on the previous run of the identical tree and red on this one, so it is flaky on top of being unrelated.

bobmatnyc:
## Code Critic — Re-review of `1cb79b38`: WARN

Both round-1 HIGH findings are genuinely fixed. I verified each independently rather than accepting the summary. One new HIGH: this PR turns a required CI check red, and it is pre-authorized to merge on approval, so that has to be said before anything lands.

## Findings

|---|---|---|---|---|---|
| HIGH | `.test-pointer-allowlist.tsv` | 213 | The doc rewrite retired the dangling `deferred_embed_backfills_vector_once_embedder_ready` citation, leaving a stale allowlist row. `scripts/check_test_pointers.sh` fails on it — the "Test: doc-comment pointer lint" check is **red on the PR right now**, and `origin/main` is green | Delete line 213, or run `scripts/check_test_pointers.sh --update` | Fix here |
| LOW | `crates/trusty-common/src/memory_core/retrieval/embed_repair_tests.rs` | 588 | `without [`embed_ledger::dir_lock`] this keeps a handful of the 16 rows` names the discarded per-directory mutex, which does not exist in this tree | Say `without `json_rmw`'s advisory lock` | Fix here |
| LOW | `crates/trusty-common/src/memory_core/store/embed_ledger.rs` | 151-161 | `clear` lost the old `if kept.len() == rows.len() { return Ok(()); }` early return. `json_rmw::update` publishes unconditionally, so once a palace holds any ledger row, **every successful deferred embed** pays a locked write + two `fsync`s to remove nothing | Pre-check with an unlocked `load`: `if !load(data_dir).iter().any(|r| ids.contains(&r.drawer_id)) { return Ok(()); }`. Racing it is harmless either way | Fix here |
| LOW | `crates/trusty-common/src/memory_core/retrieval/embed_repair.rs` | 47-61 | `load` degrades to empty and `record` refuses, so a corrupt ledger surfaces as `recorded_failures: 0` — indistinguishable from "these drawers predate the fix". The `warn!` in `load` is the only signal | Add `ledger_readable: bool` to `EmbedHealth` and surface it on `palace_reembed` | Parent |
| LOW | `deferred_embed.rs` / `embed_repair.rs` | 383-390 / 262-266, 292-295 | Three of the four ledger call sites match `if let Ok(Err(e))`, silently dropping a `JoinError`; `record_loss` logs it | Match the `record_loss` shape at all four | Parent |

## HIGH-1 — verified fixed, and the routing decision is right

Routing through `json_rmw::update` instead of a bespoke mutex is correct on the merits, not just on the common-entry-point rule. An in-process `Mutex` would have fixed threads and left processes racing; `json_rmw` fixes both, and the bespoke `save`/`TMP_COUNTER` helper is gone.

I checked the serialisation claim at the layer that decides it rather than taking the module doc's word. `fd-lock` 4.0.4 resolves to `rustix::fs::flock` on every non-Solaris Unix (`sys/unix/mod.rs:15-16`). `flock(2)` locks are held by the **open file description**, so two `open()` calls from two threads in one process get independent descriptions and do block each other. That is what makes the doc's "processes AND threads" claim true — it would have been false with `fcntl` record locks, which are per-process. Solaris takes the `fcntl_lock` branch and would not serialise threads; not a target here.

Reentrancy: neither `record` nor `clear` calls `update` from inside an `update` closure, so the documented self-deadlock does not apply.

Lock ordering against the palace write lock: no cycle. `remember_with_options` holds the palace write guard but only *spawns*; it never takes the flock. The deferred task takes the flock but never the palace guard — `record_loss` binds a `bool` from `ctx.drawers.read().iter().any(...)` at `deferred_embed.rs:341`, so the guard drops at the end of that statement, well before the `spawn_blocking` await. `backfill_missing_vectors` reads the drawer table only inside scoped blocks. No path holds one lock while waiting on the other.

Blocking contract: all **four** ledger mutation sites are on `spawn_blocking` — `record_loss`, `clear_ledger`, `record_backfill_failure`, and the post-backfill `clear`. `load` correctly stays a plain unlocked read; it needs no lock, because publish is a `rename` and a reader sees the whole old or whole new document.

Missing directory: `record` does `create_dir_all(data_dir)` and `update` does `create_dir_all(lock.parent())`. `clear` returns before either when the ledger does not exist, so a healthy palace never creates the file or its `.lock` sidecar. I checked that the new sidecar does not trip the two leaked-temp-file assertions (`retrieval/tests.rs:312`, `l1_cache.rs:296`) — both filter on an `l1_cache.json` prefix.

**Fail-before independently reproduced.** I patched my throwaway worktree to skip only the `flock` acquisition, keeping the atomic publish, and both new tests went red:

```
test concurrent_clear_and_record_do_not_lose_each_other ... FAILED
test concurrent_ledger_records_keep_every_row ... FAILED
```

Restored, they pass, and I ran both 20 consecutive times looking for flake — 20/20 green, no variance.

## HIGH-2 — verified fixed on every path

`spawn_deferred_embed` now fires at `handle.rs:708-711`, after the redb commit (`persist_with_retirement`, 684) and after the in-memory push (689-695). Any failure arriving from that point on finds the drawer present.

The branch inversion is sound: `if !opts.defer_embedding { …sync… }` at 648 and `if opts.defer_embedding { spawn }` at 708 are two plain `if`s on the same boolean with no `else`, so they stay mutually exclusive and cannot double-embed. Every early `?` between them (`persist_with_retirement` at 686) returns before the spawn, which is correct — no drawer, nothing to embed. The `?` sites *after* the spawn (`save_l1_cache` at 715) leave the drawer genuinely durable in redb and in the table, so a marker written then is accurate.

`std::slice::from_ref(&content)` is behaviourally identical to `&[content]` from the callee's side — same length, same `String` value, only ownership differs. It also removes the `content.clone()` the deferred branch used to pay.

## The deliberate semantics change — I agree with it

Read degrades to empty, write refuses over an unparseable ledger. That asymmetry is right, and it is a stronger fix than the rename-then-continue I suggested in round 1: rename-then-continue leaves the system reporting success while a human has to notice a `.corrupt` file, whereas refusing produces a loud error on every subsequent attempt and keeps the bytes for forensics.

The property that had to hold is that corruption cannot block the **repair**, and it does. `backfill_missing_vectors` reads health through `load` (degrades to empty), repairs, and only the trailing `clear` errors — logged at `warn!`, with the backfill still returning `Ok` and `repaired: N`. A corrupt ledger cannot stop drawers being made findable again.

No caller mishandles the new `Err`. All four swallow-and-log, none propagates into the embed's or the backfill's `Result` — correct, since a ledger write failure must never fail the operation it annotates. The only residuals are the two LOWs above.

## The no-test-for-HIGH-2 argument — accepted

The guard's semantics are already pinned from both sides: `loss_for_a_forgotten_drawer_is_not_recorded` (absent → no row) and `permanent_failure_writes_a_ledger_row` (present → row). What is unpinned is that the production call site satisfies the precondition, and pinning *that* means intercepting `spawn` to observe the drawer table at spawn time — an instrumentation hook for a positional constraint. Disproportionate. The inline comments at both sites state the constraint where a future editor will see it, which is the right-sized mitigation. I would not hold the PR on this.

Worth knowing: if someone moves the spawn back above the push, no existing test fails. The consequence is a return to the pre-PR behaviour, which the set-difference detector still covers, so the blast radius is bounded.

## Gates re-run at `1cb79b38`

```
cargo fmt --check                                                              EXIT=0
cargo clippy -p trusty-common --features memory-core,embedder-test-support \
             --all-targets -- -D warnings                                      EXIT=0
cargo test -p trusty-common --features memory-core,embedder-test-support --lib -- --test-threads=1
    test result: ok. 881 passed; 0 failed; 13 ignored; 0 measured
cargo test -p trusty-common ... json_rmw
    test result: ok. 8 passed; 0 failed
bash scripts/check_sld.sh                                                      OK
bash scripts/check_line_cap.sh                                                 0 violations
bash scripts/check_changelog_fragment.sh                                       OK
bash scripts/check_test_pointers.sh                                            FAILED  ← the HIGH
```

881 lib tests, up from 878 — the three new tests are real additions, not renames.

## Required changes

1. Delete `.test-pointer-allowlist.tsv:213` (or run `scripts/check_test_pointers.sh --update`) and confirm the "Test: doc-comment pointer lint" check goes green. **Do not merge while it is red** — it is red on the PR now and green on `origin/main`, so this PR is the cause.
2. Fix the stale `dir_lock` doc reference at `embed_repair_tests.rs:588`.
3. Restore the no-op early return in `clear`.

With those three done — none of which touches the durability logic I verified — this is an APPROVE. The fixes to both round-1 HIGHs are correct and independently confirmed.

## Notes

- **Unchanged code, so not a finding, but this PR makes it load-bearing.** `json_rmw.rs:33-35` claims the temp path's pid+nanosecond stamp means "concurrent writers never share a scratch path even if one of them bypasses the lock." My lock-bypass experiment falsified exactly that: 16 barrier-released threads landed on the same nanosecond and collided on one temp path, producing `rename` → `ENOENT`. Under the lock this is unreachable, and this PR relies on the lock, not on temp uniqueness — so nothing here is at risk. Flagging it only because the sentence reads as a second line of defence that is not actually there.
- Round-1 attacks recorded as settled were not re-run: cold-start retry budget, orphan coverage, backfill short-circuit ordering, lock-across-await, `all_ids()` fail-open direction.
- The environmental reds (two `trusty-memory` doctor probes, two `trusty-mpm` tests) were not treated as blocking. The pointer-lint failure is a separate thing — I confirmed it against `origin/main` before flagging it.

bobmatnyc:
All three done, plus the `json_rmw` comment. Head is now `e300943690724bf2c224d30d9ddf903526cb248a`.

## HIGH — the red required check

Confirmed and fixed. `.test-pointer-allowlist.tsv` is a **ratchet**: it fails on an entry that is no longer needed, precisely so a fixed pointer cannot leave dead weight behind. My doc rewrite retired the `deferred_embed_backfills_vector_once_embedder_ready` citation — a pointer that had never resolved, since that test does not exist on `main` either — which made its grandfather row stale. One line deleted.

Fail-before verified rather than assumed: I restored the row, ran the gate, and got

```
FAIL: crates/trusty-common/src/memory_core/retrieval/handle.rs:
  `deferred_embed_backfills_vector_once_embedder_ready` is allowlisted in
  .test-pointer-allowlist.tsv but is no longer dangling anywhere in this file
  — remove its entry (run --update).
test-pointers: 0 dangling pointer(s), 1 stale allowlist entry — FAILED.
```

then removed it again and got green. Worth recording why round 1 shipped it red: `scripts/check_test_pointers.sh` was not in the rung-5 command set I was working from, so I ran line-cap, SLD, changelog-fragment and generation-artifacts but not this one. It is in the list now.

## MEDIUM — a real regression, correctly called

The old `clear` returned without writing when nothing matched. `json_rmw::update` has no "nothing changed" branch, so once a palace held any ledger row, every subsequent **successful** deferred embed paid a locked double-fsync to republish an unchanged document — on the hot path, for no reason.

The membership check is restored ahead of the lock, not inside it: paying an flock to discover there is nothing to do is most of the cost being avoided. The unlocked check-then-act is benign, and the doc says why — a row appearing in that window belongs to a drawer that failed *after* this caller's drawer succeeded, so clearing it would have been wrong anyway. The authoritative surface stays `embed_health()`, which never reads this file.

Pinned by `clear_without_a_matching_row_never_writes`: asserts the mtime and the bytes are both untouched after a non-matching clear, and that a palace with no ledger gets neither the file nor its `.lock` sidecar created.

## LOW — stale doc

The `embed_ledger::dir_lock` citation pointed at the per-directory mutex I discarded in favour of `json_rmw`. Rewritten to describe the flock that is actually there, including the open-file-description scoping that makes it serialise threads — the property your round-2 verification established at the `rustix::fs::flock` layer.

## json_rmw's atomicity contract

Corrected. The claim that the pid + nanosecond temp name keeps writers apart "even if one of them bypasses the lock" is now stated as what it is: scratch-path hygiene that prevents a torn file, and nothing more. The comment names the falsifying experiment so the next reader does not re-derive the same false confidence. Comment only — no behaviour change to `json_rmw`.

## Rung-5 gate + the lint that was missing

```
cargo fmt --check                                                              EXIT=0
cargo clippy -p trusty-common --features memory-core,embedder-test-support
             --all-targets -- -D warnings                                      EXIT=0
cargo test -p trusty-common --features memory-core,embedder-test-support -- --test-threads=1
  test result: ok. 882 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out
  test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
cargo test -p trusty-memory -- --test-threads=1
  test result: FAILED. 561 passed; 2 failed; 4 ignored; 0 measured; 0 filtered out
cargo test -p trusty-mpm
  test result: FAILED. 4635 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui     EXIT=0

  test-pointers: resolved 21870 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
  line-cap: measured 3724 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
bash scripts/check_sld.sh
  sld-lint: scanned 56 spec doc(s) + 3108 code file(s); 0 error(s), 0 warning(s)
  changelog-fragment gate: scanned 17 changed path(s); all 2 crate(s) with source changes are recorded.
bash scripts/check_generation_artifacts.sh
  generation-artifacts: scanned 16 stylesheet(s) + 1221 prose file(s) — 0 leaked artifacts — OK.
```

### The 3 red tests remain outside this diff

`git diff --name-only origin/main...HEAD` returns 0 files for both `crates/trusty-mpm/` and `crates/trusty-memory/src/commands/`.

- `commands::doctor::tests::check_daemon_health_*` (2) — probe `127.0.0.1:7070/health`, which answers `"daemon_state":"warming"` on this machine (#4836 latch). `interpret_health_body` maps that to `Warn`; the tests assert `Fail | Pass | Unknown`.
- `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — in `trusty-mpm`, untouched here. It is also demonstrably flaky: across three runs of effectively identical trees it has been green, red, and red, while `daemon::managed_routes::tests::stale_assets_for_many_...` went red, red, green. Different test each time, same crate, no source change between them.

bobmatnyc:
## Code Critic — Third pass on `e3009436`: APPROVE

Verification only; settled ground not reopened. All four items check out. One LOW correction to a doc comment's stated reasoning — the code it describes is right, the justification for it is not.

## Findings

|---|---|---|---|---|---|
| LOW | `crates/trusty-common/src/memory_core/store/embed_ledger.rs` | 157-161 | The benign-race justification is backwards in one clause and inaccurate in another. A row appearing in the window is, by construction, for a drawer this caller just embedded **successfully** — so it is stale and clearing it would have been *right*, not wrong. And `embed_health` does read this file (into `recorded_failures`); what never reads it is `missing_vector_ids` | Reword: a row appearing in the window leaves one stale annotation for a drawer that does have a vector; detection (`missing_vector_ids`) never consults the ledger, so neither health nor repair is affected | Parent |

## 1. The red required check — cleared

The deletion is exactly one row: 945 → 944, and the diff removes only `crates/trusty-common/src/memory_core/retrieval/handle.rs<TAB>deferred_embed_backfills_vector_once_embedder_ready`. No other allowlist entry moved.

```
    test-pointers: resolved 21870 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```

Fail-before reproduced independently: I reinserted the row at its original position and re-ran.

```
FAIL: … `deferred_embed_backfills_vector_once_embedder_ready` is allowlisted … but is no longer dangling
test-pointers: 0 dangling pointer(s), 1 stale allowlist entry — FAILED.
```

Reverted, green again. The ratchet explanation holds: the gate fails on an entry no longer *needed*, so replacing the dangling citation with real test names is what made the grandfather row stale. Live CI agrees — **"Test: doc-comment pointer lint" now passes** on the PR, where it was red at `1cb79b38`.

## 2. The `clear` no-op restoration — right call, wrong reason written down

Placing the check ahead of the lock is correct. Paying an `flock` plus two `fsync`s to discover there is nothing to do is most of the cost being avoided, and `json_rmw::update` has no "nothing changed" branch to exploit from inside. On a palace holding any ledger row, every successful deferred embed now pays one unlocked read-and-parse of a small JSON file instead. Good trade.

The race is benign, which is what I was asked to confirm rather than assume — but not for the reason given.

For the skip to matter at all, the row appearing in the window must have `drawer_id ∈ ids`; a non-matching row was never going to be cleared. Both callers pass ids they just finished with: `clear_ledger(ctx, id)` passes the drawer that just embedded successfully, and the post-backfill `clear(repaired)` passes the drawers just repaired. So the appearing row describes a drawer that **now has a vector** — `embed_and_store` only upserts on success and never deletes on failure, so the successful attempt's vector survives the concurrent failing one. The row is therefore stale, and clearing it would have been correct. The comment says the opposite.

The reachable interleaving is narrow: a deferred embed for drawer X still in flight while `palace_reembed` runs a backfill that also targets X, with one attempt succeeding and the other failing. Consequence is one stale row in a diagnostic file that nothing will later clean up.

It stays benign because the second clause is nearly right even though it is imprecise. `embed_health` *does* read the ledger — `embed_repair.rs:147-151` loads it into `recorded_failures`. What never reads it is `missing_vector_ids`, and that is the surface `backfill_missing_vectors` acts on. So a stale row cannot cause a drawer to be missed, cannot cause one to be needlessly re-embedded, and cannot affect the repair. It is noise in `recorded_failures`, nothing more. LOW, and not worth another round.

`clear_without_a_matching_row_never_writes` pins the check properly: mtime plus a byte comparison after a non-matching clear, and separately that a healthy palace gets neither `embed_failures.json` nor its `.lock` sidecar. mtime as a no-write proxy is sound here — APFS carries nanosecond granularity, so a write landing in the same nanosecond is not a realistic false pass, and the byte assertion backs it up regardless.

## 3. The `dir_lock` doc — accurate now

The dead reference is gone. The replacement states that `record` routes through `json_rmw::update`, whose advisory `flock` is held by the open file description and so serialises threads as well as processes. That matches what I established at the `rustix::fs::flock` layer in round 2, including why it would have been false with `fcntl` record locks.

## 4. The `json_rmw` atomicity comment — comment-only, and correct

The diff touches the `//!` block and nothing else: no code, no signature, no behaviour change in `json_rmw`. Guarantee 2 now says the temp name is scratch-path hygiene rather than a second line of defence, names the falsification, and points at guarantee 1 as the only thing keeping concurrent writers apart. That is exactly the claim my experiment supported.

## Gates at `e3009436`

```
bash scripts/check_test_pointers.sh     0 dangling, 0 stale — OK
bash scripts/check_line_cap.sh          3724 files, 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh               OK
bash scripts/check_changelog_fragment.sh  both trusty-common and trusty-memory recorded
cargo fmt --check                       EXIT=0
cargo clippy -p trusty-common --features memory-core,embedder-test-support \
             --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-common --features memory-core,embedder-test-support --lib -- --test-threads=1
    test result: ok. 882 passed; 0 failed; 13 ignored; 0 measured
```

882, up from 881 — `clear_without_a_matching_row_never_writes` is a real addition. I ran it alongside the two concurrency tests 20 consecutive times: 20/20 at 3 passed, 0 failed, no variance.

## Notes

- **Live CI at this head: 26 pass, 2 skipping, 1 pending.** The pending one is the 17-minute `Test` job. Nothing is red. Merging still needs that check to land green — I am not treating pending as passing.
- `mergeStateStatus` is `BEHIND`. Per the standing guidance that BEHIND is not a correctness gate, that is not a reason to hold, but it is worth knowing before the merge.
- The doc now says the lock-bypass patch keeps "1 of 16 rows". My own bypass, patched at a slightly different point, never reached the row count — the unlocked writers collided on a shared temp path and failed the `rename` with `ENOENT` first. Both results show the test red without the lock; the difference is only where the unserialised version breaks. Not worth changing.
- The three environmental reds were not treated as blocking. The `trusty-mpm` evidence is convincing on its own terms: a different test failing on each of three runs over effectively identical trees is flakiness, not signal.

PR #4978, opened 2026-08-05
feat(trusty-mpm): add `n [name]` picker command for launching a named session

### LINKED ISSUE #4965: feat(trusty-mpm): add `n [name]` picker command for launching a named session
> **Revised 2026-08-05.** The first version of this issue cited a tree 678 commits stale. All line numbers and the command-word inventory below have been re-verified against `origin/main` @ c679c3c3. The proposal itself is unchanged — `n` is confirmed still free.

### Problem

In the interactive session picker (bare `tm` and `tm ls`), creating a new session is only possible by typing an integer — the next free launch slot. Two issues:

1. The integer is unstable. It shifts with the session list, so there is nothing to memorize.
2. There is no way to name the session. It always gets a daemon-derived name from the repo/cwd, so a project with several concurrent workstreams produces `tm-trusty-tools-01…05` with nothing distinguishing them.

The naming capability already exists everywhere else — `tm session new --name-hint <name>`, the MCP `tm_new_session` tool, and the daemon's `SpawnRequest.name_hint` all accept an explicit name. The picker is the only entry point that doesn't expose it.

### Proposal

Add an `n` command to the picker, following the same prefix grammar as the existing `d<N>` (delete) and `r<N> <name>` (rename) commands:

- `n` → launch a new session, auto-named (same as the integer choice today)
- `n some-name` → launch a new session named from the argument

The existing `[N] launch new session` entry stays. `n` is purely additive; no current behaviour changes.

### Current picker grammar (verified @ c679c3c3)

`parse_picker_choice` — `crates/trusty-mpm/src/bin/tm/commands/session_picker.rs:659-740` — accepts exactly:

| Input | Meaning | Added by |
|---|---|---|
| `q` / `Q` | Quit | — |
| `ls` / `list` | Re-print session list | #3863 |
| `r<N> <name>` / `r <N> <name>` (also `R`) | Rename session N | #3724 |
| `d<N>` / `d <N>` (also `D`) | Delete session N | #2304 |
| *(empty)* | Bare-Enter default (resume / confirm-restart / launch) | — |
| numeric matching a slot | Resume / confirm / blocked | — |
| numeric == next launch slot | Launch new | — |
| anything else | `Unrecognised` | — |

**`n` is free.** It matches no prefix (`strip_prefix(['r','R'])` and `strip_prefix(['d','D'])` don't match it), is not `q`/`ls`/`list`, and does not parse as `u32` — so it currently falls straight through to `Unrecognised`.

**No collision with the `tm ls` filter grammar.** The `tm ls recent|alpha <filter>` sort/substring grammar (#3485) is a CLI positional-argument construct consumed by `run_ls_connector` *before* the picker loop starts and re-applied only around re-fetches (`session_picker.rs:1017-1019`, `1111-1112`). `parse_picker_choice` has no filter branch, so stdin `n` cannot be swallowed as a filter term.

### Decisions taken

| Decision | Choice |
|---|---|
| Bare `n` | Launches auto-named. Stable regardless of session count. |
| Integer entry | Kept. Additive change, no muscle-memory break. |
| Name sanitization | Silent. The raw string goes to `name_hint`; the daemon sanitizes. |
| Name collisions | Not possible — the `NN` serial is daemon-allocated, so `n foo` twice yields `tm-foo-01` and `tm-foo-02`. |

### Implementation

Three mechanical changes. No new naming or validation logic — `crates/trusty-common/src/session_naming.rs` already lowercases, collapses non-alphanumeric runs to `-`, truncates, and falls back to `local` on empty, and is tested.

1. **`session_picker.rs:67-114`** — `PickerDecision::LaunchNew` carries `Option<String>`. The enum already has 10 variants including `Rename(usize, String)`, so a payload-carrying `LaunchNew` is consistent with existing shape.
2. **`session_picker.rs:659-740`** — add an `n`/`N` branch to `parse_picker_choice`. Place it alongside the existing `r`/`d` prefix branches: `strip_prefix(['n','N'])`, trim the remainder; empty → `LaunchNew(None)`, non-empty → `LaunchNew(Some(name))`. Note the rename branch already establishes the `<letter><slot> <name>` precedent for carrying a name through this parser. Menu text in `run_tty_picker` (`session_picker.rs:795-1022`) gains an `[n <name>]` line.
3. **`crates/trusty-mpm/src/bin/tm/commands/guided_launch.rs:142-223`** — `launch_new_session_and_attach` gains `name_hint: Option<&str>`, forwarded into the POST body to `/api/v1/sessions/managed` (body constructed at `:154-168`) as `"name_hint"` when present. This is still the only picker-reachable call site that hardcodes a body with no `name_hint` key; everything downstream (`ManagedSpawnRequest` → `SpawnRequest` → `SessionManager::create_with_id` → `resolve_session_name`) already handles it.

No daemon, HTTP-type, `TrustyCommand`, or naming-module changes required.

### Tests

The picker's tests now span **two** files — new cases go in the first, matching where the other `parse_picker_choice` cases live:

- **`crates/trusty-mpm/src/bin/tm/tests_behavior_c_tests.rs`** — 21 `guided_picker_*` tests. Model the new ones on `guided_picker_delete_parses_d_prefix` (`:196`) and `guided_picker_ls_does_not_shadow_launch_new_or_delete_prefix` (`:430`).
- `crates/trusty-mpm/src/bin/tm/commands/session_picker_tests.rs` — owns the `Rename`/render/color/slot-staleness tests. Worth reading for the `parse_picker_choice_rename_*` naming style.

Cases:

- `n` → `LaunchNew(None)`
- `N` → `LaunchNew(None)` (case-insensitive, matching `d`/`D` and `r`/`R`)
- `n auth-refactor` → `LaunchNew(Some("auth-refactor"))`
- `nauth` → `LaunchNew(Some("auth"))` (prefix-strip, matching `d2` ≡ `d 2`)
- `n   ` (whitespace-only remainder) → `LaunchNew(None)`
- `n my auth fix` → `LaunchNew(Some("my auth fix"))` — multi-word survives to `name_hint`, daemon sanitizes to `my-auth-fix`
- Shadowing guard: `n` must not shadow, and must not be shadowed by, `ls` / `list` / `d` / `r` / numeric / bare-Enter. Existing `q`, `d<N>`, `r<N> <name>`, `ls`, numeric, and empty-line cases unchanged.

### Test ladder rung

**Rung 3** (localized behavior inside one crate). PR gate:

```
cargo fmt --check && cargo check -p trusty-mpm && cargo clippy -p trusty-mpm -- -D warnings && cargo test -p trusty-mpm
```

### Notes

- This targets the numbered picker in `session_picker.rs`, **not** the separate full-screen ratatui coordinator TUI (`tm sessions tui`, `crates/trusty-mpm/src/tui/coordinator/`), which has its own `/new <repo> <ref> <task>` slash-command language.
- The picker has been split across siblings since the original research: `session_picker_rename.rs`, `session_picker_render.rs`, `session_picker_prune.rs`, `session_picker_tests.rs`, `picker_delete.rs`. Only `session_picker.rs` and `guided_launch.rs` need edits.

mac-duetto:
Closes #4965

Adds an `n` command to the interactive session picker (bare `tm` / `tm ls`) so a new session can be launched with a meaningful name instead of the daemon-derived `tm-<repo>-NN`.

## Grammar

| Input | Result |
|---|---|
| `n` / `N` (exact) | Launch new, auto-named |
| `n <name>` / `N <name>` (separator required) | Launch new, named |
| `new`, `no`, `nn`, `n1`, `nauth` | `Unrecognised` — unchanged |

The integer "launch new session" entry is unchanged. This is purely additive.

## Two commits

**`c8ccb317`** — the feature as originally specified in #4965.

**`7fdbf4d7`** — fixes from an adversarial review that returned WARN. The first commit used `strip_prefix(['n','N'])`, which consumed **any** n-prefixed input and returned a decision that unconditionally clones a repo, creates a worktree, starts tmux, and attaches — with no confirm step. The menu line literally reads *launch new session*, so an operator typing `new` would get a real session named `tm-ew-01`. The delete confirm is spelled `[y/N]`, so `n` already means "no" inside this tool.

The `d<N>` / `r<N> <name>` no-space precedent is only safe because their remainder must resolve to a live slot — `d9` on a 3-session list is `Unrecognised`. `n` had no such gate. It is now an exact-match token plus a required separator.

## Sanitization moved CLI-side

Previously the raw string was forwarded and the daemon sanitized it. Two bad outcomes:

- `n feature/auth` and `n hotfix/auth` both produced `tm-auth-NN` — the daemon's `resolve_session_name` routes the hint through `leaf_slug_from_dir`, whose `Path::file_name()` discards everything before the last `/`. That defeats the point of the feature.
- `n !!!` sanitized to empty and fell back to `tm-local-NN`, a name in a leaf namespace shared with every other project on the daemon and *less* identifiable than the default.

The name is now kebab-cased in the picker before it is sent. Rather than exposing the private `sanitize_slug` (which would also mean exposing `MAX_FOLDER_LEN`, inviting drift), this adds one public item — `leaf_slug_from_hint` in `trusty-common` — and refactors `leaf_slug_from_dir` to delegate to it. No second kebab-caser. `leaf_slug_from_hint_is_a_fixed_point_for_its_own_output` proves the daemon's own pass is a no-op rather than asserting it.

A name that sanitizes to empty is now rejected with a message instead of silently spawning. When the sanitized name differs from what was typed, one line is printed before launch: `tm: naming it 'my-auth-fix' (from 'My Auth Fix!').`

## Other changes

- The command legend moved out of `session_picker.rs` into `session_picker_render.rs` as `command_legend(Option<u32>) -> Vec<String>`. Inline, the file hit 515 SLOC against the repo's 500 cap and `check_line_cap.sh` refused it. This also made the column alignment and the `tm-<name>-NN` wording assertable — 4 tests.
- The `n` legend line is self-documenting (`launch new session as tm-<name>-NN`), since the adjacent `r<N> <name>` takes a *verbatim* full session name while `n <name>` takes a leaf.
- `launch_new_session_and_attach`'s `name_hint` plumbing had zero coverage. Two wire tests now assert the key is present on `Some` and **absent** (not null) on `None`.

## Test ladder — rung 4

`trusty-common` gained public API, so this is a cross-crate change.

```
cargo fmt --check                                        → 0
cargo clippy -p trusty-mpm --all-targets -- -D warnings   → 0
cargo clippy -p trusty-common --features session-naming   → 0
scripts/check_line_cap.sh                                 → 0
cargo test -p trusty-common  → 340 passed; 0 failed; 6 ignored
cargo test (lib)             → 4636 passed; 0 failed; 7 ignored
cargo test (bin tm)          → 1491 passed; 0 failed; 1 ignored
cargo test (bin trusty-mpm)  → 1491 passed; 0 failed; 1 ignored
cargo check --workspace (excl. GUI crates) → 0
```

## Pre-existing failures on this base — not from this change

Each was verified against the stashed base before being set aside. Nothing was `#[ignore]`d or narrowed to make a gate green.

- `ensure_managed_config_dir_emits_the_frozen_skill_warning` — fails ~1 run in 4, passes in isolation. `#[serial]` only serializes against other `#[serial]` tests, and installing a subscriber perturbs the process-global interest cache. Failed 1-of-4 on base.
- `compose_session_instructions_display_matches_live_prompt[_with_override]` — `tm` and `trusty-mpm` are duplicate bin targets over the same `main.rs`, so `cargo test -p trusty-mpm` runs the test twice concurrently against the same on-disk project state. Failed 1-of-3 on base; green when the targets run separately.
- `cargo check --workspace` — `trusty-code-gui` and `trusty-mpm-gui` panic in `tauri::generate_context!` because `ui/dist` is not built in this worktree. Identical exit 101 on base.

CI may be flaky on this PR for those reasons, unrelated to the change.

### DIFF
--- a/crates/trusty-common/src/session_naming/mod.rs
+++ b/crates/trusty-common/src/session_naming/mod.rs
@@ -261,7 +261,33 @@ pub fn leaf_slug_from_dir(path: &Path) -> String {
         .file_name()
         .map(|s| s.to_string_lossy().into_owned())
         .unwrap_or_default();
-    sanitize_slug(&basename, MAX_FOLDER_LEN)
+    leaf_slug_from_hint(&basename)
+}
+
+/// Sanitize an operator-typed name hint to the SAME leaf slug the daemon's
+/// `name_hint` path would produce — the shared CLI/daemon contract (#4965).
+///
+/// Why: `SessionManager::resolve_session_name` sanitizes a `name_hint` by
+/// routing it through [`leaf_slug_from_dir`], which first takes
+/// `Path::file_name()` — so a hint of `feature/auth` loses everything before
+/// the last `/` and lands as `tm-auth-NN`, indistinguishable from a
+/// `hotfix/auth` hint. The CLI must therefore slugify BEFORE sending, and it
+/// must do so with the exact same algorithm and the exact same cap, or the two
+/// sides drift. This is that entry point: the private [`sanitize_slug`] under
+/// its `name_hint`-path cap ([`MAX_FOLDER_LEN`]), with no `file_name()` step.
+/// Because its output is always a single component of `[a-z0-9]`-bounded
+/// kebab-case within the cap, feeding it back through [`leaf_slug_from_dir`]
+/// daemon-side is a provable no-op — the CLI's chosen leaf is the leaf the
+/// session gets.
+/// What: delegates to [`sanitize_slug`] with [`MAX_FOLDER_LEN`]. Returns an
+/// empty string when nothing alphanumeric survives — the caller decides
+/// whether that is a fallback (`build_managed_session_name`'s `"local"`) or a
+/// refusal (the picker's `n <name>`, which will not spawn an unusable name).
+/// Test: `leaf_slug_from_hint_keeps_path_like_segments`,
+/// `leaf_slug_from_hint_matches_leaf_slug_from_dir_for_plain_names`,
+/// `leaf_slug_from_hint_is_a_fixed_point_for_its_own_output`.
+pub fn leaf_slug_from_hint(hint: &str) -> String {
+    sanitize_slug(hint, MAX_FOLDER_LEN)
 }
 
 /// Derive a session name from a project directory's basename (no serial).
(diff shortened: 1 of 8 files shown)


bobmatnyc:
Superseded by #5352 (merged, squash `67eb6b346`), which rebases this PR onto current `main` and cherry-picks its two reviewed commits (`c8ccb317`, `7fdbf4d7`) unmodified. Closing #4965 already closed by #5352. All feature logic — `PickerDecision::LaunchNew`/`UnusableName`, `leaf_slug_from_hint`, the naming message, the `n <name>` legend — is present unchanged in main.

PR #4979, opened 2026-08-06
fix(trusty-mpm): guard `tm catalog apply --prune` against deleting user content (#391)

### LINKED ISSUE #391: feat(trusty-mpm): safely prune stale manifest-owned agent and skill deployments
## Current state

Agent and skill deployers now maintain ownership manifests and checksums. They can distinguish tool-managed files from untracked or user-modified files, but deployment still leaves a managed copy behind when its source is removed, renamed, or deselected.

The remaining work is a conservative prune operation, not a new manifest system.

## Scope

- Compare the resolved deployment set with manifest-owned target files.
- Classify stale entries caused by source removal, rename, or roster deselection.
- Preview the exact prune set before mutation.
- Remove a stale file only when its current checksum still matches the manifest's last deployed checksum.
- Preserve all untracked and user-modified files, reporting why each was skipped.
- Update the ownership manifest atomically after successful removals.
- Apply the same safety rules to agents and skills.
- Emit concise audit output for removed and preserved files.

## Acceptance criteria

- Dry-run/preview is available.
- Removed and renamed managed assets are detected.
- Unmodified manifest-owned stale files can be pruned.
- Modified or untracked files are never deleted.
- Partial failure leaves the manifest consistent with files actually removed.
- Agent and skill paths have equivalent tests.

bobmatnyc:
## Defect

`prune_skills_locked` (`crates/trusty-mpm/src/core/update_check/apply.rs`) iterated every stem in `manifest.managed` and called `std::fs::remove_dir_all` on any the current `HarnessPlan::skill_selected()` rejected. No checksum comparison, no frozen check, no backup — the only skill-mutating path with none of the guards its neighbours have. `skills::deployer::deploy_one_file` refuses to overwrite a checksum-mismatched file; `core::skill_repair` backs up before every overwrite and deletes nothing at all.

**A checksum gate alone is not enough.** It protects a hand-edited skill, whose bytes no longer match the ledger. It does not protect a *pristine* user-custom one: `deploy_all_skill_tiers` deploys `~/.trusty-mpm/skills/` in full and explicitly exempt from the bundled include/exclude, but the resulting ledger entry is indistinguishable from a bundled entry — matching checksum and all. An operator writing a `[skills] include` rule to curate the *bundled* set could delete their own skill with it.

## Resolution

Both guards, plus the same checksum gate and backup on the agent path:

| Guard | Applies to |
|---|---|
| Checksum + frozen gate before removal | agents and skills |
| A skill directory holding any file no ledger key claims is left alone (`remove_dir_all` would take it too) | skills |
| A stem the live `~/.trusty-mpm/skills/` source supplies is never pruned | skills |
| Everything deleted is copied to `<framework-root>/backup-catalog-prune-<ts>/` first | agents and skills |
| A kept asset keeps its ledger entry — dropping it would reclassify it as user-owned and freeze it | agents and skills |

### Why the tier is derived, not recorded

Recording an origin tier in the skill manifest is the obvious route and is a schema change. It would have to decide what every *existing* entry means, and the only available default — "bundled" — is precisely this bug: a user-custom skill deployed before the field existed would be pruned exactly as before. Reading the live `~/.trusty-mpm/skills/` source through `skills::tiers::list_source_stems` — the same function `deploy_all_skill_tiers` uses to build its user set — makes the two structurally unable to disagree, needs no migration, and has no wrong default. Documented consequence: a stem the user has since *deleted* from that source is no longer user-tier and prunes normally, which is the intent.

### Also fixed here

Prune judged raw ledger keys, so an `include` rule matching a stem but not its carried `<stem>/references/*.md` key dropped that file's ledger entry while leaving the file on disk — after which the deployer reads it as user-owned and never updates it again ([#4881](https://github.com/bobmatnyc/trusty-tools/issues/4881)'s freeze shape, reached from this path). Candidates are now stems; a carried file is pruned with its skill or not at all.

### Doc correction

`apply_catalog`'s doc said "user-owned files are never removed". True of *unmanaged* files, silent on the frozen and user-tier ones — the code implemented the sentence's literal words, not the safety a reader infers. It now enumerates exactly what is protected, and what is not (a stem removed from the user source; nothing here restores a backup).

## Tests — fail-before / pass-after

Seven new tests in `apply/tests.rs`, plus one unit test on the backup-root path. **6 of 7 fail on the pre-fix code**; the seventh (`apply_prune_removes_deselected_skill`) passes by design — it is the "the feature still works" case.

<details><summary>FAIL BEFORE — production files reverted to <code>73599f364</code>, tests kept</summary>

```
running 12 tests
test core::update_check::apply::tests::apply_prune_skips_hand_edited_agent ... FAILED
test core::update_check::apply::tests::apply_prune_spares_user_owned ... ok
test core::update_check::apply::tests::apply_is_stale_before_apply_then_fresh_after ... ok
test core::update_check::apply::tests::apply_redeploys_and_clears_staleness ... ok
test core::update_check::apply::tests::apply_prune_spares_untracked_file_in_a_managed_skill ... FAILED
test core::update_check::apply::tests::apply_preserves_user_tier_override_deployed_via_launch_path ... ok
test core::update_check::apply::tests::apply_prune_removes_deselected_skill ... ok
test core::update_check::apply::tests::apply_prune_keeps_the_ledger_entry_for_a_selected_skills_carried_file ... FAILED
test core::update_check::apply::tests::apply_prune_skips_hand_edited_skill ... FAILED
test core::update_check::apply::tests::apply_prune_backs_up_before_removing_a_skill ... FAILED
test core::update_check::apply::tests::apply_prune_removes_deselected ... ok
test core::update_check::apply::tests::apply_prune_spares_user_tier_skill ... FAILED

---- apply_prune_skips_hand_edited_skill stdout ----
panicked at tests.rs:405:39:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, ... }

---- apply_prune_spares_untracked_file_in_a_managed_skill stdout ----
panicked at tests.rs:495:5:
the user's own file must survive --prune

---- apply_prune_keeps_the_ledger_entry_for_a_selected_skills_carried_file stdout ----
panicked at tests.rs:569:5:
nothing is deselected here: ApplyReport { ..., skills_pruned: ["carrier/references/notes.md"] }

---- apply_prune_backs_up_before_removing_a_skill stdout ----
panicked at tests.rs:529:45: a prune that deleted must leave a backup

---- apply_prune_spares_user_tier_skill stdout ----
panicked at tests.rs:461:39:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, ... }

test result: FAILED. 6 passed; 6 failed; 0 ignored; 0 measured; 4640 filtered out
```

(The `NotFound` unwraps are the data loss itself — the test reads back a file prune had just deleted. One test was withheld from this run: `apply_prune_reports_what_it_kept_and_where_backups_went` asserts on report fields that do not exist before the fix, so it cannot compile against the old code.)
</details>

<details><summary>PASS AFTER</summary>

```
running 13 tests
test core::update_check::apply::tests::apply_prune_skips_hand_edited_agent ... ok
test core::update_check::apply::tests::apply_redeploys_and_clears_staleness ... ok
test core::update_check::apply::tests::apply_prune_spares_user_owned ... ok
test core::update_check::apply::tests::apply_prune_skips_hand_edited_skill ... ok
test core::update_check::apply::tests::apply_is_stale_before_apply_then_fresh_after ... ok
test core::update_check::apply::tests::apply_prune_spares_untracked_file_in_a_managed_skill ... ok
test core::update_check::apply::tests::apply_prune_spares_user_tier_skill ... ok
test core::update_check::apply::tests::apply_prune_keeps_the_ledger_entry_for_a_selected_skills_carried_file ... ok
test core::update_check::apply::tests::apply_preserves_user_tier_override_deployed_via_launch_path ... ok
test core::update_check::apply::tests::apply_prune_removes_deselected ... ok
test core::update_check::apply::tests::apply_prune_reports_what_it_kept_and_where_backups_went ... ok
test core::update_check::apply::tests::apply_prune_removes_deselected_skill ... ok
test core::update_check::apply::tests::apply_prune_backs_up_before_removing_a_skill ... ok

test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 4640 filtered out
```
</details>

## Gates — Rust Test Ladder rung 5 (destructive path, cross-crate reads, user-data loss)

```
cargo fmt --check                                              # EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings        # EXIT=0
cargo clippy -p trusty-agents-common --all-targets -- -D warnings  # EXIT=0
cargo test -p trusty-mpm                                       # 4646 passed; 1 failed (known flake, below)
cargo test -p trusty-agents-common                             # 482 passed; 0 failed
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui  # see note
bash scripts/check_line_cap.sh          # 3725 files, 0 violations — OK
bash scripts/check_sld.sh               # 0 errors, 0 warnings
bash scripts/check_test_pointers.sh     # 21913 citations, 0 dangling
bash scripts/check_changelog_fragment.sh # OK trusty-mpm
bash scripts/check_capabilities.sh      # up to date (7 generated files)
```

Two reds, both pre-existing and neither caused by this branch:

- `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` — the `$HOME`-mutation parallel-run race listed in `docs/reference/test-ladder-baseline.md` line 38. Passes in isolation on this branch:
  ```
  running 1 test
  test daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet ... ok
  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4653 filtered out
  ```
  This diff touches no file in `daemon/`.
- `cargo check --workspace` fails on `trusty-agents-ui`: `The frontendDist configuration is set to "../dist" but this path doesn't exist` — a Tauri `generate_context!` proc-macro panic on a fresh worktree with no built Svelte bundle. Adding `--exclude trusty-agents-ui` gives `EXIT=0` across every other member.

Full raw output for each command is in the run logs; nothing above is a summary of results in my own words beyond the counts cargo printed.

## Scope note

The agent prune path got the checksum gate and backup (it had the same hand-edit hazard, and the issue's own acceptance criteria ask for equivalent agent and skill coverage). It did **not** get tier awareness: no user-level agent tier deploys into `agent_deploy_dir()`, so there is no second tier for a bundled rule to reach across.

Closes [#391](https://github.com/bobmatnyc/trusty-tools/issues/391)

### DIFF
--- a/crates/trusty-mpm/changelog.d/391-guard-catalog-apply-prune.md
+++ b/crates/trusty-mpm/changelog.d/391-guard-catalog-apply-prune.md
@@ -1,0 +1,8 @@
+Fixed
+
+- `tm catalog apply --prune` no longer deletes user content (closes [#391](https://github.com/bobmatnyc/trusty-tools/issues/391))
+  - it removed any managed skill directory the current include/exclude rejected, with no checksum comparison, no frozen check and no backup — the only skill-mutating path with none of the guards `deploy_one_file` and `skill_repair` already have
+  - a hand-edited (frozen) agent or skill is now left in place, as is a skill directory holding any file trusty-mpm did not deploy — `remove_dir_all` would have taken that file too
+  - a skill the user-custom tier supplies is never pruned, whatever the bundled include/exclude says: `deploy_all_skill_tiers` deploys that tier in full and exempt from those rules, but its ledger entry is indistinguishable from a bundled one, so a checksum gate alone waved a pristine user skill through. The tier is derived from the live `~/.trusty-mpm/skills/` source rather than a new manifest field, so no existing ledger entry has to be guessed at
+  - everything actually deleted is copied to `~/.trusty-mpm/backup-catalog-prune-<timestamp>/` first, and the command now prints what it kept and why
+  - prune judges skill STEMS, not raw ledger keys: an `include` rule matching a stem but not its carried `<stem>/references/*.md` key used to drop that file's ledger entry while leaving the file on disk, after which the deployer read it as user-owned and never updated it again
--- a/crates/trusty-mpm/src/bin/tm/commands/managed.rs
+++ b/crates/trusty-mpm/src/bin/tm/commands/managed.rs
@@ -904,6 +904,19 @@ pub(crate) async fn catalog(action: CatalogAction) -> anyhow::Result<()> {
                     report.skills_pruned.len()
                 );
             }
+            // #391: a guard that declines to delete is invisible unless it says
+            // so — the operator asked for a prune and must be told what survived
+            // it, and where the deleted content went.
+            if let Some(backup) = &report.prune_backup_dir {
+                println!("  backed up to {}", backup.display());
+            }
+            for line in report
+                .agents_prune_kept
+                .iter()
+                .chain(report.skills_prune_kept.iter())
+            {
+                println!("  kept: {line}");
+            }
         }
     }
     Ok(())
--- a/crates/trusty-mpm/src/core/update_check/apply.rs
+++ b/crates/trusty-mpm/src/core/update_check/apply.rs
@@ -14,7 +14,8 @@
 //! Test: this module's `tests` cover redeploy-clears-staleness and prune-removes-
 //! deselected against a `FakeGitBackend`-seeded catalog.
 
-use std::path::Path;
+use std::collections::BTreeSet;
+use std::path::{Path, PathBuf};
 
 use crate::core::agent_deployer::deploy_agents_filtered;
 use crate::core::agent_manifest::{
@@ -22,11 +23,16 @@ use crate::core::agent_manifest::{
 };
 use crate::core::manifest::HarnessPlan;
 use crate::core::paths::FrameworkPaths;
+use crate::core::skill_drift::key_stem;
 use crate::core::skill_manifest::{
     SKILL_MANIFEST_FILE, SkillManifest, SkillManifestSave, with_skill_manifest_lock,
 };
-use crate::core::skill_tiers::deploy_all_skill_tiers;
+use crate::core::skill_tiers::{deploy_all_skill_tiers, list_source_stems};
 use crate::provisioner::GitBackend;
+
+mod prune_guard;
+
+use prune_guard::Verdict;
 
 /// A failure raised while applying a catalog update.
 ///
@@ -90,6 +96,31 @@ pub struct ApplyReport {
     pub agents_pruned: Vec<String>,
     /// Managed skill stems removed because the manifest no longer selects them.
     pub skills_pruned: Vec<String>,
+    /// Deselected agents a guard left in place, as `<filename>: <reason>` (#391).
+    pub agents_prune_kept: Vec<String>,
+    /// Deselected skills a guard left in place, as `<stem>: <reason>` (#391).
+    pub skills_prune_kept: Vec<String>,
+    /// Where the pruned content was copied before deletion, when anything was
+    /// deleted (#391). `None` means nothing was removed, so nothing was backed up.
+    pub prune_backup_dir: Option<PathBuf>,
+}
+
+/// One prune stage's result: what went, what stayed, and whether a backup exists.
+///
+/// Why (#391): a prune that silently declines to delete something is a guard the
+/// operator cannot see working. Carrying the kept set alongside the removed set
+/// is what lets `tm catalog apply --prune` say "left 2 alone, here is why".
+/// What: removed names, kept names each with their reason, and whether this
+/// stage wrote anything under the run's backup root.
+/// Test: `apply_prune_skips_hand_edited_skill`, `apply_prune_removes_deselected`.
+#[derive(Debug, Default)]
+struct PruneOutcome {
+    /// Names actually deleted this run.
+    removed: Vec<String>,
+    /// Names a guard declined to delete, formatted `<name>: <reason>`.
+    kept: Vec<String>,
+    /// True once this stage has copied something under the backup root.
+    used_backup: bool,
 }
 
 /// Apply a catalog update: sync, redeploy the selected set, optionally prune.
@@ -104,10 +135,36 @@ pub struct ApplyReport {
 /// [`HarnessPlan`], redeploys the selected agents/skills via the filtered
 /// deployers, and — when `prune` is set — removes managed agents/skills the plan
 /// no longer selects (and drops their manifest entries). Returns an
-/// [`ApplyReport`]. Pruning is opt-in and only ever touches MANAGED files (those
-/// in the checksum manifest); user-owned files are never removed.
+/// [`ApplyReport`].
+///
+/// Pruning is opt-in, and #391 replaced the old summary ("only ever touches
+/// MANAGED files; user-owned files are never removed") because it was true of
+/// UNMANAGED files and silent on the two kinds of user content that were in fact
+/// deletable. Precisely what prune will and will not remove:
+///
+/// - Only files the checksum manifest records as trusty-mpm-managed are eligible.
+///   A file the user dropped in was never, and is still not, a candidate.
+/// - An eligible asset is removed only when its bytes still match the checksum
+///   the ledger recorded, so a hand-edited (frozen) agent or skill stays. A
+///   skill directory is judged whole: any file under it that no ledger key
+///   claims disqualifies the directory, because `remove_dir_all` would take that
+///   file too.
+/// - A skill stem the user-custom tier supplies is never pruned, whatever the
+///   bundled include/exclude says — [`deploy_all_skill_tiers`] deploys that tier
+///   in full and exempt from those rules, but its ledger entries look identical
+///   to bundled ones.
+/// - Everything actually deleted is copied under
+///   `<framework-root>/backup-catalog-prune-<timestamp>/` first;
+///   [`ApplyReport::prune_backup_dir`] names it.
+///
+/// What this does NOT promise: a stem the user has since deleted from
+/// `~/.trusty-mpm/skills/` is no longer user-tier and prunes normally, and
+/// nothing here restores a backup — that stays a human decision.
 /// Test: `apply_redeploys_and_clears_staleness`, `apply_prune_removes_deselected`,
-/// `apply_prune_spares_user_owned`.
+/// `apply_prune_spares_user_owned`, `apply_prune_removes_deselected_skill`,
+/// `apply_prune_skips_hand_edited_skill`, `apply_prune_spares_user_tier_skill`,
+/// `apply_prune_spares_untracked_file_in_a_managed_skill`,
+/// `apply_prune_backs_up_before_removing_a_skill`.
 pub fn apply_catalog<G: GitBackend>(
     git: G,
     fw: &FrameworkPaths,
@@ -167,8 +224,24 @@ pub fn apply_catalog<G: GitBackend>(
 
     // 4. Optionally prune managed artifacts the manifest no longer selects.
     if prune {
-        report.agents_pruned = prune_agents(&agent_target, &plan)?;
-        report.skills_pruned = prune_skills(&skill_target, &plan)?;
+        // #391: ONE backup root per run, so an operator undoing a prune finds
+        // everything it took in one place. Named here but created lazily — a
+        // prune that removes nothing must not litter the framework root.
+        let backup_root = prune_guard::prune_backup_root(&fw.root, chrono::Utc::now());
+        let agents = prune_agents(&agent_target, &plan, &backup_root)?;
+        let skills = prune_skills(
+            &skill_target,
+            &fw.user_skill_source_dir(),
+            &plan,
+            &backup_root,
+        )?;
+        if agents.used_backup || skills.used_backup {
+            report.prune_backup_dir = Some(backup_root);
+        }
+        report.agents_pruned = agents.removed;
+        report.agents_prune_kept = agents.kept;
+        report.skills_pruned = skills.removed;
+        report.skills_prune_kept = skills.kept;
     }
 
     Ok(report)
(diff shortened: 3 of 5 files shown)


bobmatnyc:

Independent adversarial review of `3aa022dce` (rung 5, destructive path). Reviewed against the spec and the code. Everything below was reproduced in a fresh worktree off `origin/main`, not read off the PR body.

### Verified independently

**Fail-before is real, and stronger than the PR claims: 7 of 12, not 6.** I reverted `apply.rs` + `managed.rs` to `73599f364`, deleted `prune_guard.rs`, kept the tests:

```
test result: FAILED. 5 passed; 7 failed; 0 ignored; 0 measured; 4640 filtered out

apply_prune_skips_hand_edited_skill    tests.rs:414:39  NotFound  (the deleted file)
apply_prune_spares_user_tier_skill     tests.rs:470:39  NotFound  (the deleted file)
apply_prune_skips_hand_edited_agent    tests.rs:616:39  NotFound  (the deleted file)
apply_prune_spares_untracked_file...   tests.rs:504:5   "the user's own file must survive --prune"
apply_prune_keeps_the_ledger_entry...  tests.rs:578:5   skills_pruned: ["carrier/references/notes.md"]
apply_prune_backs_up_before_removing   tests.rs:538:45  "a prune that deleted must leave a backup"
apply_prune_removes_deselected         tests.rs:242:45  "a prune that deleted must leave a backup"
```

The three `NotFound` unwraps are not coverage theatre. Each test asserts `deployed.is_file()` (or writes to it) *before* the prune, so a deploy failure would surface at a different line; the `NotFound` is specifically the read-back of a file prune had just deleted. That is the data loss, expressed as a panic.

The out-of-brief bug is real too, and the pre-fix output names it exactly: `skills_pruned: ["carrier/references/notes.md"]` on a skill the `include` rule fully selected. Base `prune_skills_locked` guarded the delete with `skill_dir.is_dir()`, so the file survived — but `manifest.managed.remove(&stem)` with `stem` = the full key dropped the ledger entry anyway, leaving a managed file on disk with no entry. That is #4881's freeze shape. The stem-based fix is complete: the new candidate set is `key_stem(key)` over all keys, strictly narrower than the old raw-key set for stems that have both forms, and broader only where a stem exists solely via a carried key — and there `skill_verdict` finds the unclaimed `SKILL.md` and keeps it.

**Pass-after:** `test result: ok. 14 passed; 0 failed` (targeted), `test result: ok. 4647 passed; 0 failed; 7 ignored` (full `cargo test -p trusty-mpm`). `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` passed here. `cargo fmt --check`, `cargo clippy -p trusty-mpm --all-targets -- -D warnings`, `check_line_cap.sh` (3725 files, 0 violations), `check_sld.sh` (0/0), `check_changelog_fragment.sh` all EXIT=0.

### Attacks run against the guard

Six throwaway probes written into my own worktree, then discarded:

| Probe | Setup | Result |
|---|---|---|
| **Backup failure** | Planted a regular file at every backup-root name the run could pick | `Err("prune io error: could not create .../backup-catalog-prune-.../skills/doomed: Not a directory")`; **skill dir survives, ledger entry survives**. Backup cannot fail silently — `back_up_tree(...)?` at `apply.rs:449` precedes `remove_dir_all` at `:451`; same shape for agents at `:320`/`:323` |
| **User source gone** | Deployed a user-tier skill, deleted `~/.trusty-mpm/skills/`, pruned | Deleted — but `skills_pruned: ["mine"]`, backup at `backup-catalog-prune-<ts>/skills/mine/SKILL.md` holding the original bytes. Documented at `apply.rs:160-162` |
| **Hidden stray** | `.DS_Store` inside a deployed skill dir | Kept: `"hosted: it holds .../.DS_Store, which trusty-mpm never deployed"` — see LOW-1 |
| **Nested user file** | User file inside the deployed `references/` subdir | Kept. `collect_files` recursion works |
| **Symlinked `SKILL.md`** | Claimed file replaced by a symlink to a file outside the tree, identical content | Pruned; `remove_dir_all` removes the link, **the user's real file survives**, and `copy_tree` followed the link so the content is in the backup |
| **Symlinked skill dir** | Whole `<skills>/<stem>` is a symlink into the user's dotfiles, contents pristine | Pruned; `remove_dir_all` deletes only the symlink. **The user's real directory and its `SKILL.md` survive.** Rust does not follow a top-level symlink |

**The tier derivation holds where it matters.** `user_skill_source_dir()` and `claude_skills_dir()` are both fields of one `FrameworkPaths` (`paths.rs:156`/`:161`), and `apply_catalog`'s only production caller (`managed.rs:876`) passes `FrameworkPaths::default()`. A relocated `$HOME` moves the ledger and the user source together, so there is no "different machine, ledger says bundled" variant — the ledger at the new target is empty and there is nothing to prune. `for_managed_project` rewrites `claude_skills` but not `user_skills`; `apply_catalog` never uses it, so that asymmetry is unreachable from this path. The read at `apply.rs:390` happens *after* step 3's deploy, so the prune's stem set is a superset of the deploy's in every ordering except the user deleting the source in between — the documented case above. A permissions error on the source propagates as `ApplyError` and aborts the whole prune (fail-closed); only a *missing* directory yields an empty set.

**`skill_verdict` enumeration.** It returns `Prunable` in exactly two situations: the directory is absent, or every file the recursive walk finds is claimed by a `key_stem == stem` ledger key *and* checksums to that key's entry. `checksum_matches` is `.get(key).is_some_and(...)`, so an unknown key is `false` → `Kept`. A symlink is `lstat`-ed by `DirEntry::file_type`, so it is never recursed into and lands as an unclaimed path unless it sits exactly at a claimed path. A read error is `Kept`. A walk error is `Kept`. I could not construct a case where a user-authored or hand-edited file is deleted without first being copied out.

**`cargo check --workspace` does need a third exclusion.** Reproduced independently in my own fresh worktree:

```
error: proc macro panicked
   --> crates/trusty-agents/ui/src-tauri/src/main.rs:144:16
    | .build(tauri::generate_context!())
    = help: message: The `frontendDist` configuration is set to `"../dist"` but this path doesn't exist
```

This is a property of any worktree with no built Svelte `dist/`, not of this PR. Every worktree-based PR's workspace check line needs `--exclude trusty-agents-ui` alongside `trusty-code-gui` and `trusty-mpm-gui`, or a `pnpm -C crates/trusty-agents/ui build` first. Worth fixing in the docs once rather than re-deriving per PR.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/update_check/apply/prune_guard.rs` | 81-97 | `agent_verdict` gates on checksum alone and ignores `ManifestEntry::origin`. The agent ledger already carries the tier field the skill ledger lacks, and `agents::deployer.rs:355-368` branches on `origin.is_framework_owned()` to preserve `Origin::User` / `Origin::Registry` entries on mismatch. Prune does not, so a *pristine* user-origin agent is `Prunable` — the same hole this PR fixes for skills, on the path where the field already exists. Latent today: a workspace-wide grep finds `Origin::User`/`Origin::Registry` only in tests and doc comments, so nothing writes a non-`Bundled` origin. It goes live the day anything does (`tm agent import`, a registry pull, a hand-edited ledger). | Before the read: return `Verdict::Kept` when `manifest.managed.get(filename).is_some_and(\|e\| !e.origin.is_framework_owned())`, plus one test seeding `Origin::User` | Fix here |
| LOW | `crates/trusty-mpm/src/core/update_check/apply/prune_guard.rs` | 139-144, 180-191 | Any unclaimed file disqualifies the whole skill, including dotfiles. On macOS a Finder visit drops `.DS_Store` into `~/.claude/skills/<stem>/`, after which that skill is permanently un-prunable (probe reproduced). Safe direction, and the kept-reason names the offending file so it is discoverable, but `--prune` will become a partial no-op on a normal macOS box. | Either skip `.`-prefixed entries in `collect_files` (matching `source_scan::is_hidden`), or leave as-is and say so in the `apply_catalog` doc's "what this does NOT promise" list. Skipping trades a `.DS_Store` false-negative for deleting a user's `.env`, so leaving it is defensible — just make it a stated choice | Parent |
| LOW | `crates/trusty-mpm/src/core/update_check/apply.rs` | 320-334, 449-472 | An abort mid-loop (backup I/O failure on iteration N) leaves iterations 1..N-1's deletions on disk with their ledger entries still present, because the `save`/`save_merging` below the loop is never reached. Self-heals — the deployer treats an absent-but-managed target as a fresh write — so no freeze and no loss, and the backups for what was deleted are intact. One sentence in the concurrency doc-comment would save the next reader re-deriving it. | Note it in the `prune_agents` / `prune_skills` doc comments | Parent |

### Notes

- The central claim survives the attack. Checksum-gating alone is not sufficient, tier awareness is a genuinely independent second requirement, and deriving the tier from `list_source_stems` on the live source is correct at the one call site that exists. `list_source_stems` and `deploy_all_skill_tiers` both route through `scan_skill_sources`, so the stem derivation cannot drift, and `key_stem(key)` is the first path segment of the deployer's own key — same name space, no translation layer to get wrong.
- The residual edge is the one the PR documents: a stem the user has removed from `~/.trusty-mpm/skills/` stops being user-tier. I reproduced it and it does delete a pristine user skill. Not a blocker: the content is copied out first, the stem is named in `skills_pruned`, the backup root is printed by `managed.rs:910`, and the precondition is the user having removed their own source. The alternative — a manifest `origin` field — would have to default every pre-existing entry to `bundled`, which reproduces the original bug for exactly the skills at risk. The tradeoff is argued correctly.
- The doc correction on `apply_catalog` is the right call. "User-owned files are never removed" was true of unmanaged files and silent on the two kinds that were deletable; the replacement enumerates both what is protected and what is not.
- Nothing outside `update_check/apply*` and one CLI print block changed. `apply_catalog` has one production caller, and `skills_pruned` switching from ledger keys to stems is only ever read for `.len()`.

Issue #3291, opened 2026-07-19
fix(trusty-mpm): tm ls reports running/attached sessions as (stopped)


bobmatnyc:
## Bug

`tm ls` / the session picker lists sessions the operator is actively attached to as `(stopped)` and offers `restart` — because the list handler serves each session's PERSISTED `state` field with no reconciliation against live tmux. After a daemon restart (before the reconcile pass runs) every record can read `stopped` even though its tmux session is alive.

## Root cause

`daemon::managed_routes::list_managed_sessions` builds summaries straight from `record.state` (`summary::record_to_summary`) and never probes real tmux. The picker's `(stopped)` label and its restart-vs-connect decision (`guided_resume::needs_restart`) both key off that stale string.

## Fix

The list handler now reconciles each displayed state against live tmux (`ManagedTmuxDriver::list_sessions` + a new `attached_session_names`): a session whose tmux exists reads `active` (`attached` when a client is attached), one with no live tmux reads `stopped`. Terminal states (`decommissioned`/`deleted`), `provisioning`, and `errored` are left as persisted. The picker then offers connect (resume/attach) for a live session, restart only for a truly-stopped one.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4931, opened 2026-08-05
: two order-dependent tests fail under parallel execution, pass in isolation


bobmatnyc:
**Affected tests:**
- `ensure_managed_config_dir_emits_the_frozen_skill_warning` — captures through `tracing::subscriber::with_default`, a THREAD-LOCAL subscriber that loses to any globally-installed subscriber another test in the same binary sets.
- `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call` — writes into `FrameworkPaths::default()` under a `fake_home()` env guard and is marked `#[serial]`, which serialises it only against other `#[serial]` tests, NOT against parallel tests reading `HOME`.

**Symptom:** both failed twice mid-session, both pass in isolation, and a full re-run on an identical tree passed 4602/0. Order-dependent, not a regression.

**Observed during:** PR #4919 work, which touches neither path.

**DIAGNOSIS STATUS: DIAGNOSED — do not re-derive this diagnosis.**

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
**Additional flake family recorded:** `daemon::managed_routes::tests::stale_assets_for_many_*` (broader than the single variant noted above)

**Defect.** The full `stale_assets_for_many_*` test family fails under load — measured 2 of 3 full-suite runs on a pristine `origin/main` worktree with zero local changes (2026-08-06).

**Root cause, newly diagnosed.** `fake_home()` mutates `/Users/masa` process-wide, while the fixture seeds its skill source at `FrameworkPaths::default()`'s path, which the code under test resolves by name. A concurrent non-serial test that self-heals that source materializes the whole skill roster into it. Each workspace then gets roughly 14 skills instead of one, so the read count measures 14 against an expected 1. Confirmed by dumping the read log.

**Resolution.** Not fixed here. An exploratory fix was written and deliberately reverted to keep [PR #4991](https://github.com/bobmatnyc/trusty-tools/pull/4991) to a single outcome. The shape of the fix is to stop the fixture depending on a process-global `/Users/masa` mutation, or to make the affected tests `#[serial]`.

[...]

PR #4983, opened 2026-08-06
fix(trusty-mpm): stop `tm ls` labelling a multi-client session `(active)`

bobmatnyc:
## Defect

`tm ls` labelled `tm-dogfood-relaunch-01` — the session the operator was sitting in — `(active)` instead of `(attached)`, while nine other sessions read `(attached)` correctly.

`#{session_attached}` is the NUMBER of clients attached to a tmux session, not a boolean. `SessionInfo::parse` compared the field to the literal `"1"`, so any session with two or more clients parsed as detached.

**Root cause:** `crates/trusty-mpm/src/daemon/tmux.rs:153` (on `origin/main` 73599f36)

```rust
let attached = parts.next().map(|s| s == "1").unwrap_or(false);
```

This is the single site where the attachment field is parsed. The label is derived from a LIVE `tmux list-sessions` on every `tm ls` — no cache, no persisted marker, no daemon-held value — through `real_tmux.rs:89` → `summary.rs:212` → `session_picker_render.rs:158`. The render layer and the reconcile layer were both correct.

## Evidence

Raw `tmux list-sessions -F "$SESSION_LIST_FORMAT"` at the time of the report:

```
tm-apex-01:1785104361:1               -> (attached)  correct
tm-cto-02:1785888572:1                -> (attached)  correct
tm-xflux-01:1785851153:1              -> (attached)  correct
...six more single-client sessions, all correct...
tm-dogfood-relaunch-01:1785443731:2   -> (active)    WRONG
tm-compset-engine-notes-01:...:0      -> (active)    correct
```

Nine clients on nine sessions across nine ttys is real — each is a separate terminal tab. Exactly one row was wrong, and it was the one with the most clients (`/dev/ttys000,/dev/ttys002`). A second terminal on a session is ordinary, so the label failed precisely for the session most likely to be the operator's.

## `(active)` vs `(attached)`

They are distinct states, not a fallback pair. `state_color` (`session_picker_render.rs:100`) gives `attached` precedence over every `state` value. `(active)` legitimately means the tmux session is live but has zero clients — correct for the six detached sessions in that listing. Only sessions with two or more clients were mislabelled.

## Resolution

Parse the field as the count it is; any value `> 0` is attached. A missing or unparseable field still degrades to `false` rather than failing the row.

No cache was removed because none existed, and no tmux invocation was added — this is a string→integer parse on a line already fetched. `tm ls` cost is unchanged.

## Tests

`crates/trusty-mpm/src/daemon/tmux_tests.rs`:

- `parses_multi_client_session_row_as_attached` — counts 1/2/3/17 all read attached, `0` does not, garbage degrades to detached.
- `attachment_flips_to_detached_when_client_count_drops_to_zero` — the same session flips when its last client goes away.

Fail-before, on the unfixed parser:

```
thread 'daemon::tmux::tests::parses_multi_client_session_row_as_attached' panicked at
crates/trusty-mpm/src/daemon/tmux_tests.rs:183:9:
session_attached=2 is a client COUNT — any non-zero value means a client is
connected, so the row must read attached

test result: FAILED. 11 passed; 1 failed; 0 ignored; 0 measured; 4635 filtered out
```

Pass-after:

```
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 4634 filtered out
```

## Gates — test ladder rung 3 (localized inside `trusty-mpm`)

```
cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0
  test result: ok. 4640 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out
bash scripts/check_line_cap.sh           3724 files, 0 violations — OK
bash scripts/check_test_pointers.sh      21893 citations, 0 dangling
bash scripts/check_changelog_fragment.sh OK trusty-mpm
```

Two default-threaded runs each produced one failure, in a different test each time, both passing in isolation and both in files this diff does not touch (`git diff --name-only origin/main -- crates/trusty-mpm/src/core/` is empty):

- `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` — documented in `docs/reference/test-ladder-baseline.md:38` as a parallel `$HOME`-mutation race, remedy `--test-threads=1`.
- `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — same family: a `tracing::subscriber::with_default` capture test whose `#[serial]` guard only serializes against other `#[serial]` tests. Not yet in the baseline doc; worth adding there, not fixed here.

The documented single-threaded run is clean.

## Scope

Attached/active labelling only. The auto-prune gap (#4384), the `tm-deadrt*` DEAD detection, and default-resume-target behaviour are untouched. This fix does not make any of them easier.

No open issue covers this defect — the closest, #3291, is a different bug (`(stopped)` for running sessions) and is closed.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4983-tm-ls-multi-client-attached.md
+++ b/crates/trusty-mpm/changelog.d/4983-tm-ls-multi-client-attached.md
@@ -1,0 +1,7 @@
+Fixed
+
+- `tm ls` no longer labels a session with two or more attached tmux clients
+  `(active)` instead of `(attached)`. `#{session_attached}` is a client count,
+  not a boolean, and the row parser compared it to the literal `"1"` — so the
+  session the operator was actually sitting in (two terminals on it) read as
+  detached while every single-client session read correctly.
--- a/crates/trusty-mpm/src/daemon/tmux.rs
+++ b/crates/trusty-mpm/src/daemon/tmux.rs
@@ -127,7 +127,7 @@ pub struct SessionInfo {
     pub name: String,
     /// Unix epoch seconds the session was created.
     pub created: i64,
-    /// Whether a client is currently attached.
+    /// Whether at least one client is currently attached.
     pub attached: bool,
 }
 
@@ -135,10 +135,16 @@ impl SessionInfo {
     /// Parse one `name:created:attached` row from `list-sessions`.
     ///
     /// Why: a single parser keeps the format in sync with
-    /// `core::tmux::SESSION_LIST_FORMAT`.
-    /// What: splits on `:`; tolerates a malformed `attached` flag by defaulting
-    /// it to `false`.
-    /// Test: `parses_session_row`.
+    /// `core::tmux::SESSION_LIST_FORMAT`. The third field is
+    /// `#{session_attached}`, which tmux documents as the NUMBER of clients the
+    /// session is attached to — not a boolean. Comparing it to the literal
+    /// `"1"` therefore read every session with two or more clients as detached,
+    /// and `tm ls` labelled it `(active)` instead of `(attached)`.
+    /// What: splits on `:`; the attachment field is parsed as a count and any
+    /// value `> 0` means attached. A missing or unparseable field still
+    /// degrades to `false` rather than failing the row.
+    /// Test: `parses_session_row`, `parses_multi_client_session_row_as_attached`,
+    /// `attachment_flips_to_detached_when_client_count_drops_to_zero`.
     pub fn parse(line: &str) -> Result<Self> {
         let mut parts = line.splitn(3, ':');
         let name = parts
@@ -150,7 +156,13 @@ impl SessionInfo {
             .next()
             .and_then(|s| s.parse::<i64>().ok())
             .ok_or_else(|| Error::Protocol(format!("bad tmux created field: {line:?}")))?;
-        let attached = parts.next().map(|s| s == "1").unwrap_or(false);
+        // PR #4983: `#{session_attached}` is a client COUNT, not a flag — a
+        // session with a second terminal on it reports `2`, and the old
+        // `== "1"` test read that as detached.
+        let attached = parts
+            .next()
+            .and_then(|s| s.trim().parse::<u32>().ok())
+            .is_some_and(|clients| clients > 0);
         Ok(Self {
             name,
             created,
--- a/crates/trusty-mpm/src/daemon/tmux_tests.rs
+++ b/crates/trusty-mpm/src/daemon/tmux_tests.rs
@@ -163,6 +163,68 @@ fn parses_session_row() {
     assert!(!detached.attached);
 }
 
+/// `#{session_attached}` is a CLIENT COUNT, not a boolean.
+///
+/// Why: `tm ls` labelled `tm-dogfood-relaunch-01` — the session the operator
+/// was sitting in, with two tmux clients on it — `(active)` rather than
+/// `(attached)`, while nine single-client sessions read `(attached)`. The
+/// parser compared the field to the literal `"1"`, so every count above one
+/// fell through to `false`. Two clients on one session is ordinary (a second
+/// terminal tab, an `attach -r` observer, a detached-and-reattached shell), so
+/// the label failed exactly for the session most likely to be the operator's.
+/// What: any count `> 0` is attached; `0` is not; a malformed field still
+/// degrades to `false` rather than erroring the whole row.
+/// Test: this function IS the test.
+#[test]
+fn parses_multi_client_session_row_as_attached() {
+    for clients in ["1", "2", "3", "17"] {
+        let info = SessionInfo::parse(&format!("tm-dogfood-relaunch-01:1785443731:{clients}"))
+            .unwrap_or_else(|e| panic!("row with {clients} client(s) must parse: {e}"));
+        assert!(
+            info.attached,
+            "session_attached={clients} is a client COUNT — any non-zero value \
+             means a client is connected, so the row must read attached"
+        );
+    }
+
+    assert!(
+        !SessionInfo::parse("tm-detached-01:1785443731:0")
+            .unwrap()
+            .attached,
+        "zero clients is the only value that means detached"
+    );
+
+    // A field tmux would never emit must not be read as attachment.
+    assert!(
+        !SessionInfo::parse("tm-garbage-01:1785443731:banana")
+            .unwrap()
+            .attached,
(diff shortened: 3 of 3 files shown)


bobmatnyc:
## Verdict: APPROVE

Narrow review, one question: who else consumes the `attached` parse, and does correcting it change any decision beyond the display label.

## Blast radius — full consumer enumeration

`SessionInfo::parse` (`daemon/tmux.rs:156`) is the only site that reads `#{session_attached}`. It feeds `TmuxDriver::list_sessions()`, which fans out two ways.

**A. `list_all_sessions()` → `ExternalSession::new(name, attached, created)`** (`tmux.rs:572`) → `services/tmux_service.rs:126` → the HTTP `{sessions: [ExternalSession]}` response (`api.rs:1566`). `ExternalSession.attached` is never read anywhere in the workspace — it is a serialized wire field with no in-tree consumer. No decision.

**B. `RealTmuxDriver::attached_session_names()`** (`real_tmux.rs:89`) — the only `filter(|s| s.attached)` in the tree → `summary.rs:250` → `reconcile_live_state`, `summary.attached = is_live && attached.contains(name)` (`summary.rs:212`).

The load-bearing detail is what line 212 does **not** touch. `is_live` comes from the name set plus `pane_exists_checked`; `summary.state` is derived from `is_live` alone. Neither reads the attached field. **This parse never participated in liveness determination**, so the PM's rung-5 escalation condition does not trigger — rung 3 is correct.

`SessionSummary.attached` reaches the wire and has exactly four readers:

| Site | What it does | Effect of the fix |
|---|---|---|
| `session_picker_render.rs:158,163` | label + color | multi-client now reads `(attached)` — the fix |
| `managed.rs:310` | `tm ls` STATE column | same |
| `session_picker.rs:361` `group_rank` | sort group 0 vs 1 | multi-client sessions now sort with attached, not active. Ordering only |
| `session_picker_prune.rs:508` `is_clearable_state` | the only decision | see below |

`ManagedTmuxDriver::list_sessions` (`real_tmux.rs:196`) maps rows to names only, so the `live` set never saw this field.

## Destructive paths — was anything eligible for a kill because a multi-client session read detached?

One destructive consumer: auto-prune, via `is_clearable_state` → `is_dead_record` → record removal. Direction is `s.attached` ⇒ **not** clearable. So before the fix that guard was silently inert for every session with 2+ clients.

It was never actually exposed, because three independent conditions also gate it:

1. `s.state` must be `stopped`/`errored`. `reconcile_live_state` rewrites Active|Stopped records to `active` whenever the tmux name is live, and a session with clients on it is live by construction. Only a record persisted as `Errored` can reach the prune with a live tmux session.
2. `!live_tmux_names.contains(&s.name)`. `live_tmux_session_names()` (`session_picker_prune.rs:521`) runs its **own** tmux enumeration and takes field 0 only (`line.split(':').next()`) — it never reads the attached field, so a multi-client session's name was always in that set. It also fails closed: `None` prunes nothing.
3. `workspace_verified_gone(s)`.

So the buggy read cost a redundant guard, not a real one — the independent name-only liveness check already covered the same sessions. Correcting it restores defense-in-depth. Strictly safety-improving, no behavior change in practice.

Nothing else kills, prunes, decommissions, reaps, restarts, or reassigns on the basis of *not* attached. `session_decommission`, `session_stop`, restart, `orphan_gc`, `adopt`, `dedup`, and `reconcile` all read `list_sessions()` names and never the attached field.

## The inverse — anything now permanently deferred while a second terminal stays open?

`is_clearable_state` is the only skip-because-attached path, and no. The skip already fired via `live_tmux_names`, and it releases the moment the tmux session dies. No upgrade, migration, or asset-sync path gates on attachment — `staleness_meaningful_for` and `probe_staleness_in_list` key on `state`. The daemon-side `attached` and the CLI-side `live_tmux_names` are two enumerations at slightly different instants, so a race can defer a prune by one `tm ls` cycle at most.

## Garbage → detached is the right direction

Correct as written, and not a fail-open. A false `attached=false` can only mislabel a row, misorder a row, or drop the redundant prune guard — and that guard is already covered by the independent name-only check. A false `attached=true` would suppress a legitimate prune. Attachment never *authorizes* a destructive action; it only ever *blocks* one, so neither direction is dangerous and the tiebreaker is consistency: `false` preserves the pre-existing degradation contract and matches `ManagedTmuxDriver::attached_session_names`'s trait default (empty = nothing attached).

The real format can't produce garbage anyway: tmux expands `#{session_attached}` to an integer on every version that has the variable, and a version old enough to lack it expands the whole `#{...}` to empty, which parses to detached. The added `.trim()` also tolerates a stray `\r`, which the old `== "1"` did not.

## Tests

Both pin behaviour, not implementation — they call the public `SessionInfo::parse` and assert on the public `attached` field, with no reference to `u32`, `is_some_and`, or any internal. The 1/2/3/17 table covers the boundary (1 is the value the old code special-cased, 2 is the minimum regression, 17 rules out a single-digit-only parse); `0` pins the negative and `banana` pins the degradation. `attachment_flips_to_detached_when_client_count_drops_to_zero` pins both directions, which stops a future "true if the field exists" shortcut from passing.

Fail-before independently reproduced: reverting only the parse expression on head `418fc742` fails **both** new tests.

```
test daemon::tmux::tests::parses_multi_client_session_row_as_attached ... FAILED
test daemon::tmux::tests::attachment_flips_to_detached_when_client_count_drops_to_zero ... FAILED
panicked at crates/trusty-mpm/src/daemon/tmux_tests.rs:183:9:
session_attached=2 is a client COUNT — any non-zero value means a client is
connected, so the row must read attached

test result: FAILED. 13 passed; 2 failed; 0 ignored; 0 measured; 4632 filtered out
```

Pass-after at head:

```
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 4632 filtered out
```

## Notes

- The PR body's fail-before snapshot reads `11 passed; 1 failed`; the actual head commit fails both new tests (`13 passed; 2 failed`). The snapshot was presumably taken before the second test was added. The evidence understates itself, which is the harmless direction — the stronger result is verified above.
- Scope held as dispatched: auto-prune (#4384), `tm-deadrt*` DEAD detection, and default-resume-target behaviour are untouched by this diff and were not reviewed.
- Not blocking on the two documented flakes (`stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet`, `ensure_managed_config_dir_emits_the_frozen_skill_warning` / #4931). The second one belongs in `docs/reference/test-ladder-baseline.md`; that is a separate docs change, not this PR's debt.

Issue #4957, opened 2026-08-05
Session worktrees are cut from stale local main, not origin/main


bobmatnyc:
Defect
------

`inproject::create_session_worktree` runs `git -C <base> worktree add -b session/<name> <path>` with **no start-point and no preceding fetch**, so every new session branch is cut from the base checkout's `HEAD` — the local `main` ref. When local `main` is stale, every session starts stale.

`crates/trusty-mpm/src/daemon/managed_routes/inproject.rs:447` (`create_session_worktree`).

This repo's own `CLAUDE.md` states the rule the code breaks: "SOURCE OF TRUTH = origin/main:HEAD. Local main may be stale. Always `git fetch origin main` and branch worktrees off origin/main (not local main)."

The clone-based sibling path already does it right — `provisioner::workspace::worktree_add` fetches an explicit refspec from `origin` before `git worktree add`. Only the in-project path is affected.

Evidence
--------

Reported on tm 1.3.4 / macOS. A new session worktree landed 667 commits behind `origin/main`, and every session branch ever created on that machine pointed at the same July commit as local `main`:

```
dce4e9a98a18e65e2e95bb9ab5e46aadc1a97e30  Sat Jul 18 2026
  HEAD -> session/tm-trusty-tools-15, session/tm-trusty-tools-17, session/tm-trusty-tools-16,
  session/tm-trusty-tools-14, session/tm-trusty-tools-13, session/tm-trusty-tools-11,
  session/tm-trusty-tools-06, main
```

`origin/main` was at `194b5101` (2026-08-05). `trusty-search` was 0.34.0 in the worktree and 0.42.3 on main.

Reproduction: open a new tm session, exit it, `git log` in the worktree — the brand-new `session/tm-trusty-tools-18` branch points at the same July commit.

Resolution
----------

Resolve the repo's actual default branch (do not hardcode `main`), fetch it from `origin`, and cut the session branch from `origin/<default>`. Degrade visibly, never silently: on fetch failure use the last-known `origin/<default>` and warn with the stale tip's sha and date; with no remote at all, `HEAD` stays correct and is not a warning.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4961, opened 2026-08-05
Startup hygiene destroys gitignored working-tree content in the managed checkout


bobmatnyc:
Startup hygiene destroys gitignored working-tree content in the tm-managed checkout.

## Defect

`crates/trusty-mpm/src/daemon/managed_routes/inproject_hygiene.rs` runs `git reset --hard origin/<default_branch>` against `~/trusty-mpm-projects/<owner>/<repo>` on every daemon start. The guard `decide_reset(ahead, dirty)` requires `dirty == Some(false)`, and `is_dirty` runs `git status --porcelain` with no `--ignored`.

`git status --porcelain` does not report gitignored paths at all. Gitignored working-tree content is therefore invisible to the gate: `is_dirty` returns `Some(false)` and the reset proceeds. When a gitignored path collides with a path tracked by the target commit, `reset --hard` overwrites it silently. `write_recovery_ref` only records a commit SHA — it cannot restore never-committed content.

Live, not theoretical: hygiene is on by default, fires once per daemon start, restarts happen several times a day, and there is no per-repo opt-out. Users point Obsidian or Cursor at this directory; `.obsidian/` and note files are routinely gitignored.

Second finding, same code path: the ahead-count is measured as `rev-list --count origin/<checked-out>..<checked-out>` while the reset targets `origin/<default_branch>`. Off the default branch the safety check validates one ref while the operation moves the local branch to a different one.

## Reproduction

```
origin: .gitignore contains notes.md            (committed)
base:   notes.md holds "MY PRECIOUS NOTES"      (gitignored, never committed)
        git status --porcelain -> empty, branch not ahead
origin: commits notes.md as a tracked file
        git reset --hard origin/main
base:   notes.md == "origin version"            <- user content destroyed
```

`git merge --ff-only origin/main` clobbers it identically — git's `--overwrite-ignore` is the default and `merge` offers no way to disable it, so swapping the verb alone is not sufficient.

## Resolution

Make the update non-destructive and gate it on a collision check:

- Replace `reset --hard` with `merge --ff-only`, which refuses rather than overwrites when it cannot fast-forward.
- Before updating, refuse when any path the update would write already exists untracked on disk (`git diff --name-only HEAD <target>` minus `git ls-files`). This is what catches the gitignored case; adding `--ignored` to `is_dirty` would instead make hygiene a permanent no-op, since `target/` is gitignored and present in any built checkout.
[...]

PR #4988, opened 2026-08-06
feat(trusty-mpm): `tm run <owner>/<repo>` cold-starts a daemon-managed session

bobmatnyc:
## What

`tm run` accepted only a DOC-24 registry alias. Starting work on a repo that was neither registered nor already cloned had no entry point: the daemon-managed system can only be entered from a checkout that already exists, because `try_inproject_spawn` (`daemon/managed_routes/inproject.rs:781-800`) gates on `path.join(".git").exists()` and derives the GitHub identity from `remote.origin.url`.

`tm run <owner>/<repo>` is that cold start:

1. classify the positional (shared with `tm register`) → clone URL + `owner`/`repo`;
2. `inproject_cold_start::ensure_managed_checkout` → the managed checkout at `~/trusty-mpm-projects/<owner>/<repo>`, cloned or verified;
3. hand off to the existing `tm launch` path.

Step 3 is a handoff, not a reimplementation. `launch()` registers the project alias, provisions the per-session worktree, prepares the session, registers it with the daemon, creates the tmux host, and attaches — so the outcome is a real `SessionRecord` with a tmux pane, visible in `tm ls` and `tm sessions`, not a blocking foreground `claude`. A session started cold is indistinguishable from one started in-project.

`tm run <alias>` is unchanged.

## Reusing an existing checkout: one refusal, one warning

**Wrong remote → FAIL LOUD.** No check exists anywhere today — the standalone driver pulls from whatever `origin` happens to be, so re-pointing an alias at a new URL keeps serving the old remote forever under the new name. Refusing is the only non-destructive answer: re-pointing `origin` under a directory that may hold branches and unpushed commits from the *other* repository is not a repair. Checked before anything writes. Remotes compare through a host-aware canonicalization, so `git@github.com:o/r.git` and `https://github.com/o/r` match while two different hosts do not.

**Dirty tree → WARN AND PROCEED.** This reverses the PR's first revision, on the finding below. Since #4957 the session branch is cut from a freshly-fetched `origin/<default>` (`inproject_start_point::resolve`) and never inherits the base checkout's local `HEAD`, so a skipped fast-forward cannot leak stale content into the session. Meanwhile the tm checkout is shared with the operator and their editors, making uncommitted content its expected steady state (ADR-0030 §4/§5) — refusing blocks the common case, including this repo's own checkout, which is dirty right now.

What must not happen is silence, which is the half of the `pull_ff_only` (`core/standalone/load.rs:186-199`) shape that really is a defect: it warns where nobody looks and returns `Ok(())`. So the skip goes two places — a `warn!` for the log path, and `ManagedCheckout::refresh_skipped`, which `tm run` prints to stderr naming the path and what is dirty.

The fetch runs either way. Only the fast-forward is gated — ADR-0030 §4 exactly.

## Incidental fix

`tm run ./some/dir` routed a relative path into a registry lookup and reported `alias './some/dir' not found`. Caught by a test; it now reports that a relative path is not a repository, the same remedy `tm register` gives.

## 🔴 ADR-0030 is NOT flipped to Accepted — three findings

The original brief asked for the status flip here. It is not in this PR: ADR-0030's Consequences section is wrong about this codebase, and the ADR's decision is not what this PR implements.

**1. The `reset --hard` bullet describes code that is not there.** ADR-0030's Negative consequence says `run_hygiene_for_base` "runs a safety-gated `reset --hard` against the tm checkout at daemon startup (`inproject_hygiene.rs:291-303`)". On `origin/main` at a08e7bd8 there is no `reset --hard` in that file — it became `git merge --ff-only` in #2181 (`62fdc1c11`), before the ADR was written. Lines 291-303 are `write_recovery_ref`'s doc comment.

**2. The fix it calls "in flight on another branch" has merged.** The gitignored-clobber hole (`is_dirty` omitting `--ignored`) landed as #4962 (`51c08788e`), which added `colliding_untracked_paths` and `HYGIENE_OPT_OUT_MARKER`. Decision point 4's "never `reset --hard`" is already satisfied.

**3. Of its eight decision points only #5 is implemented, and this PR implements none of them.** A cold-start entry point is orthogonal to session↔workstream cardinality; the ADR never mentions `tm run`, cloning, or registration.

Flipping to Accepted is defensible as recording the owner's 2026-08-05 ruling — this repo uses Accepted for decision-acceptance, not implementation — but it should carry those three corrections, so it is its own change.

Finding 3's §4/§5 reading is what reversed the dirty-tree decision above.

## Gate — rung 5

```
cargo fmt --check && cargo check -p trusty-mpm --all-targets \
  && cargo clippy -p trusty-mpm --all-targets -- -D warnings \
  && cargo test -p trusty-mpm -- --test-threads=1
RUNG5_GATE_EXIT=0
TOTAL: 7919 passed, 0 failed, 17 ignored across 46 binaries
```

All nine repo lints:

```
check_line_cap.sh            EXIT=0  | 3731 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
check_sld.sh                 EXIT=0  | 56 spec doc(s) + 3108 code file(s); 0 error(s), 0 warning(s)
check_doc_numbers.sh         EXIT=0  | 102 doc(s) / 96 claim(s); 4 grandfathered, 0 violations — OK.
check_test_pointers.sh       EXIT=0  | resolved 21947 Test: citation(s) — 0 dangling pointers — OK.
check_capabilities.sh        EXIT=0  | tm-capabilities: up to date (7 generated files).
check_changelog_fragment.sh  EXIT=0  | all 1 crate(s) with source changes are recorded.
check_generation_artifacts.sh EXIT=0 | 16 stylesheet(s) + 1232 prose file(s) — 0 leaked artifacts — OK.
check_agent_assets.sh        EXIT=0  | 30 byte-parity file(s) all match; 4 pinned deviation(s) — OK.
check_buildrs_sync.sh        EXIT=0  | build.rs canonical blocks in sync across all 4 crates.
```

`tm generate capabilities` was re-run and its one-line `tm run` description change committed.

```
cargo test -p trusty-mpm -- --include-ignored --test-threads=1
TOTAL: 7752 passed, 1 failed
  meta_run_demo_writes_and_verifies_artifact
```

### Baseline failures, not change-specific

Three tests fail on a parallel run; all three are already rows in `docs/reference/test-ladder-baseline.md`, and `git diff --name-only origin/main...HEAD` returns empty for every path they live in.

| Test | Baseline row |
|---|---|
| `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` | "Parallel-run race on `$HOME` mutation between tests. Fails 3 of 5 runs on a clean baseline worktree with no branch changes. Passes under `-- --test-threads=1`" |
| `ensure_managed_config_dir_emits_the_frozen_skill_warning` | #4931 thread-local tracing subscriber losing to a global one; signature `Captured lines: []` |
| `meta_run_demo_writes_and_verifies_artifact` | "Times out at ~187s… launches a real Claude Code session… Reproduced identically on a clean baseline worktree" |

The doc prescribes `-- --test-threads=1` as the remedy, which is what the green run above uses. That runs every test — no `#[ignore]`, no `cfg`-gating, no `--exclude`, no `--lib` narrowing.

## Coverage

| Case | Test |
|---|---|
| Invalid `owner/repo` rejected before any network call | `classify_rejects_browser_pastes_and_paths`, `classify_rejects_empty_target` (both pure) |
| Dirty checkout warns and proceeds | `dirty_existing_checkout_warns_and_proceeds` — asserts the call succeeds, the skip is reported to the caller naming the dirty file, the uncommitted edit survives, the local branch did NOT move, and `origin/main` WAS fetched |
| Wrong-remote checkout fails loud, HEAD unmoved | `existing_checkout_on_a_different_remote_fails_loud`, `wrong_remote_reuse_is_refused_without_touching_the_checkout` |
| No-origin checkout fails loud | `existing_checkout_without_an_origin_fails_loud` |
| Fresh clone feeds the real daemon-managed spawn gate | `cold_start_produces_a_checkout_the_managed_spawn_path_accepts` — asserts `try_inproject_spawn` accepts the cold-started checkout and resolves to the same path |
| Clean reuse fast-forwards and reports no skip | `clean_matching_checkout_is_reused_and_refreshed` |
| `tm run` and `tm register` agree on a string | `shorthand_identity_agrees_with_url_identity` |

The dirty-tree test moves origin forward first, so "was not fast-forwarded" cannot pass vacuously.

The `SessionRecord` claim is proven at the gate that produces one (`try_inproject_spawn` returning `Ok(Some(..))` is what makes `spawn_managed_routed` register a record), not through a live daemon — that would need a running daemon, tmux server, and `claude` binary.

Integration tests pin `TRUSTY_MPM_REPOS_ROOT` to a tempdir. Without that they would have written into the operator's real `~/trusty-mpm-projects/bobmatnyc/trusty-tools`, or cloned it over the network on a machine that lacks it.

## Incidental split

`managed_routes/mod.rs` sat at exactly the 500-SLOC cap, so one `pub mod` line broke the gate. The two record-deletion handlers move to a sibling `delete.rs` and are re-exported; route wiring and callers are unchanged. Not allowlisted.

### DIFF
--- a/crates/trusty-mpm/changelog.d/4991-tm-run-owner-repo.md
+++ b/crates/trusty-mpm/changelog.d/4991-tm-run-owner-repo.md
@@ -1,0 +1,7 @@
+Added
+
+- `tm run <owner>/<repo>` cold-starts a daemon-managed session for a repo that is not on disk yet
+  - clones to the managed checkout at `~/trusty-mpm-projects/<owner>/<repo>`, then hands off to the existing `tm launch` path — so the result is a real `SessionRecord` with a tmux pane, visible in `tm ls` and `tm sessions`, not a blocking foreground `claude`
+  - the positional is classified by the same predicate `tm register` uses, so `tm run bobmatnyc/trusty-tools` and `tm register bobmatnyc/trusty-tools` cannot disagree about what a string means; a registered alias keeps the unchanged DOC-24 standalone behaviour
+  - reusing an existing managed checkout FAILS LOUD in two cases rather than proceeding silently: its `origin` names a different repository, or its tree is dirty and so cannot be refreshed. Neither is auto-fixed, and both refuse before anything writes to the directory
+  - `tm run ./some/dir` now reports that a relative path is not a repository instead of "alias not found"
--- a/crates/trusty-mpm/changelog.d/4991-tm-run-owner-repo.md
+++ b/crates/trusty-mpm/changelog.d/4991-tm-run-owner-repo.md
@@ -3,5 +3,6 @@ Added
 - `tm run <owner>/<repo>` cold-starts a daemon-managed session for a repo that is not on disk yet
   - clones to the managed checkout at `~/trusty-mpm-projects/<owner>/<repo>`, then hands off to the existing `tm launch` path — so the result is a real `SessionRecord` with a tmux pane, visible in `tm ls` and `tm sessions`, not a blocking foreground `claude`
   - the positional is classified by the same predicate `tm register` uses, so `tm run bobmatnyc/trusty-tools` and `tm register bobmatnyc/trusty-tools` cannot disagree about what a string means; a registered alias keeps the unchanged DOC-24 standalone behaviour
-  - reusing an existing managed checkout FAILS LOUD in two cases rather than proceeding silently: its `origin` names a different repository, or its tree is dirty and so cannot be refreshed. Neither is auto-fixed, and both refuse before anything writes to the directory
+  - reusing an existing managed checkout FAILS LOUD when its `origin` names a different repository — no auto-repoint, and the refusal lands before anything writes to the directory
+  - a dirty existing checkout is not an error: the fast-forward is skipped and said so in normal output, while the fetch still runs. The session's branch is cut from a freshly-fetched `origin/<default-branch>`, so it never inherits the checkout's local `HEAD`
   - `tm run ./some/dir` now reports that a relative path is not a repository instead of "alias not found"
--- a/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/cli.md
+++ b/crates/trusty-mpm/src/assets/skills/tm-capabilities/references/cli.md
@@ -101,7 +101,7 @@ Generated from `Cli::command()` (clap's command-tree introspection) — every `t
   - `push-guard` — Retrofit the #2867 cross-branch `pre-push` guard onto an existing clone
 - `restart` — Stop the running daemon and start a fresh one
 - `rm` — Remove a managed alias: deregister and delete its project dir (DOC-24)
-- `run` — Launch an interactive `claude` session for a managed alias (DOC-24)
+- `run` — Start a session for a GitHub repo (`<owner>/<repo>`) or a managed alias
 - `serve` — Alias for `start` — start the daemon if not running, no-op if it is
 - `services` — Inspect and probe workspace service daemons
   - `health` — Probe the health endpoint and print OK or FAIL
--- a/crates/trusty-mpm/src/bin/tm/cli/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/cli/mod.rs
@@ -923,17 +923,31 @@ pub(crate) enum Command {
         root: Option<String>,
     },
 
-    /// Launch an interactive `claude` session for a managed alias (DOC-24).
+    /// Start a session for a GitHub repo (`<owner>/<repo>`) or a managed alias.
     ///
-    /// Why: `tm run` is the claude-mpm replacement — it launches Claude Code
-    /// with `CLAUDE_CONFIG_DIR=<root>/claude-config` so the global
-    /// hooks/MCPs are supplied and the real `~/.claude` is excluded.
-    /// What: loads the alias if needed, checks credentials, and spawns `claude`
-    /// with inherited stdio.
-    /// Test: `cli_parses_run_standalone`.
+    /// Why (#4990): `tm run` accepted only a DOC-24 registry alias, so starting
+    /// work on a repo that was neither registered nor already cloned had no
+    /// entry point. `tm run <owner>/<repo>` is that cold start, and it lands on
+    /// the DAEMON-MANAGED system (ADR-0030): the checkout goes to
+    /// `~/trusty-mpm-projects/<owner>/<repo>`, and the session is a real
+    /// `SessionRecord` with a tmux pane that shows up in `tm ls` and
+    /// `tm sessions` — not a blocking foreground `claude`.
+    /// What: a positional naming a repo (`<owner>/<repo>` with GitHub assumed,
+    /// or a full URL for any host — classified by the same predicate
+    /// `tm register` uses) clones or verifies the managed checkout, then hands
+    /// off to the `tm launch` path. A positional naming a registered alias
+    /// keeps the unchanged standalone behaviour: load if needed, check
+    /// credentials, spawn `claude` with inherited stdio and
+    /// `CLAUDE_CONFIG_DIR=<root>/claude-config`.
+    ///
+    /// Reusing an existing managed checkout FAILS LOUD when its `origin` names
+    /// a different repository, or when its tree is dirty and so cannot be
+    /// refreshed. Neither case is auto-fixed.
+    /// Test: `cli_parses_run_standalone`, `run_target_tests.rs`.
     Run {
-        /// Registered alias to run.
-        alias: String,
+        /// `<owner>/<repo>` (GitHub assumed), a full repository URL, or a
+        /// registered alias (see `tm ls --projects`).
+        target: String,
         /// Optional initial task to pre-seed (currently unused by MVP).
         #[arg(long)]
         task: Option<String>,
--- a/crates/trusty-mpm/src/bin/tm/cli/mod.rs
+++ b/crates/trusty-mpm/src/bin/tm/cli/mod.rs
@@ -941,8 +941,14 @@ pub(crate) enum Command {
     /// `CLAUDE_CONFIG_DIR=<root>/claude-config`.
     ///
     /// Reusing an existing managed checkout FAILS LOUD when its `origin` names
-    /// a different repository, or when its tree is dirty and so cannot be
-    /// refreshed. Neither case is auto-fixed.
+    /// a different repository — trusty-mpm never re-points a remote. A DIRTY
+    /// checkout is not an error: the fast-forward is skipped, a warning naming
+    /// the path and what is uncommitted is printed, and the session starts
+    /// anyway. Its branch is cut from a freshly-fetched
+    /// `origin/<default-branch>`, so it does not inherit the checkout's state.
+    ///
+    /// `--root` applies to the alias form only. The managed checkout's location
+    /// is set by `TRUSTY_MPM_REPOS_ROOT` / `TRUSTY_MPM_WORKSPACE_ROOT`.
     /// Test: `cli_parses_run_standalone`, `run_target_tests.rs`.
     Run {
         /// `<owner>/<repo>` (GitHub assumed), a full repository URL, or a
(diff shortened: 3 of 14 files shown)


bobmatnyc:
## Verdict: APPROVE

Adversarial review at rung 5, against head `1af64ea8f`, from a detached throwaway worktree. Zero CRITICAL, zero HIGH. Five findings, all MEDIUM/LOW, none of which risks destruction.

### The destructive-operation surface holds

I traced the real code, not ADR-0030.

- **No `reset --hard` anywhere in the reachable tree.** `run_hygiene_for_base` (`inproject_hygiene.rs:512-574`) is fetch → `update_to_origin` → `worktree prune`. The update is `git merge --ff-only origin/<default>` (`:468-472`) behind six gates: detached HEAD, non-default branch, unknown ahead-count, unknown dirty, `ahead > 0`, `dirty` (`decide_update`, `:98-138`), plus the #4961 `colliding_untracked_paths` refusal (`:446-462`) and a recovery ref written first (`:464-466`). The ADR's `reset --hard` bullet is stale, as the PR body says.
- **`migrate_old_layout_aside` is the only path that moves operator files**, and it is unchanged by this PR: it renames (never deletes) a non-empty directory that has no top-level `.git`, refuses outright when `protected_state_in` finds `.base`/`.worktrees`, and `warn!`s loudly. Pre-existing on main via `tm launch` and the daemon; the cold start is a new caller of an unchanged guard.
- **`git clone --no-local <src> <dest>`** is read-only on the source.

**Wrong-remote is genuinely fail-loud before any write.** `verify_remote_matches` is the first statement in the `reused` branch (`inproject_cold_start.rs:848`); everything before it is `Path::exists`, everything after it is either read-only (`porcelain_status`) or the hygiene sweep. No path reaches a clone or a pull with a mismatched remote. `wrong_remote_reuse_is_refused_without_touching_the_checkout` pins HEAD unmoved.

### The dirty-tree reversal checks out

I attacked all four claims:

1. **Does the session branch avoid the local HEAD?** Yes. `create_session_worktree` (`inproject.rs:473-499`) calls `inproject_start_point::resolve`, which fetches `+refs/heads/<default>:refs/remotes/origin/<default>` and returns `origin/<default>` as the `git worktree add` start-point. `tm run <owner>/<repo>` reaches it: `run_managed` → `launch()` → `provision_for_launch` → `provision` (`managed_workspace.rs:115-121`) → `create_session_worktree`. Not taken on trust — the call chain is intact for this new entry point.
2. **Does the fetch still happen?** Yes. `run_hygiene_for_base` is called unconditionally in the `reused` branch (`:871`); only `update_to_origin`'s merge is gated.
3. **Is the warning visible?** Yes, twice over. `run_managed` `eprintln!`s it (`run_target.rs:340-350`), and `tm`'s stderr `EnvFilter` defaults to `info`, so the `warn!` prints too.
4. **Is `dirty_existing_checkout_warns_and_proceeds` vacuous?** No. It commits to origin *before* the call, so a fast-forward is genuinely available; the "HEAD did not move" assertion cannot pass trivially. Verified by reading the fixture and by running it.

### Silent-failure sweep — nothing found

No failed clone treated as usable (`ensure_base_clone` returns `Err` on non-zero exit, `?`-propagated). No failed provision that still launches (`run_managed` bails before `launch`). No error swallowed into an `Ok`. No cursor/state advance on a failure branch.

### Cold-start correctness

`classify_run_target` is pure — no git, no socket — so every rejection lands before the network. It routes through `register_args::classify` / `looks_like_repo` / `resolved_url` and `trusty_common::github_path::parse_owner_repo`, not a second parser, so `tm run` and `tm register` cannot disagree. I looked for siblings of the `tm run ./some/dir` hole and found one untested arm (LOW, below), not a second divergence.

Path traversal is closed: `slugify_component` maps every non-`[a-z0-9]` byte to `-`, so no `owner/repo` string can escape `<repos_root>`.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/cli/mod.rs` | 943-945 | `tm run --help` tells the user a dirty checkout **fails loud**. It warns and proceeds. The reversal commit `1af64ea8f` touched 6 files and this was not one of them. Verified by running the built binary, not by reading — paste below. | Replace with the warn-and-proceed wording already used in `inproject_cold_start.rs`'s module doc. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/run_target.rs` | 153-154 | Same stale claim in `run_managed`'s own doc — "fails loud on a mismatched remote **or a dirty tree**" — contradicting the warn-and-proceed block 20 lines below it in the same function. | Drop "or a dirty tree"; say the dirty case is reported through `refresh_skipped`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/daemon/managed_routes/inproject_cold_start.rs` | 793-798, 937-946 | `refresh_skipped: None` does **not** mean the fast-forward happened, but its doc says "`None` means it was not [skipped]". `fast_forward_skip_reason` predicts only the *dirty* gate; `decide_update` skips on five more conditions, and `run_hygiene_for_base` returns early with **no fetch at all** when `HYGIENE_OPT_OUT_MARKER` is present — which also falsifies the module doc's "the fetch runs either way" (`:702`). | Narrow the contract: rename to `dirty_skip_reason` and say it reports the dirty gate only. Or plumb it properly — have `run_hygiene_for_base` return the real `UpdateDecision`, which would also give `:871`'s currently-unreachable `.map_err(ColdStartError::Provision)?` something to catch (`run_hygiene_for_base` returns `Ok(())` on every path today). | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/run_target.rs` | 282-292 | `--root` is honoured on the `Alias` arm and silently dropped on the `Repo` arm. Ignoring it is semantically right — it names the DOC-24 standalone root, which daemon-managed sessions don't use — but dropping it silently is inconsistent with the `--task` warning three lines above. | Same treatment as `--task`: one `eprintln!` when `root.is_some()` on the repo arm. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/run_target_tests.rs` | 432-452 | The absolute-path arm is accepted but untested. `classify("/Users/me/code/app")` → `Url`, and `parse_owner_repo`'s last-two-segments rule derives `owner = "code"` from the **parent directory name**, so `tm run /Users/me/code/app` clones into `<repos_root>/code/app`. Non-destructive (`--no-local` reads only), but surprising and unpinned. | One row in `classify_resolves_full_urls`. | Fix here |

The two MEDIUM doc rows are one defect — an incomplete reversal at the doc layer — in two files.

### Evidence

Stale help text, from the binary built at `1af64ea8f`:

```
$ cargo run -q -p trusty-mpm --bin tm -- run --help
...
Reusing an existing managed checkout FAILS LOUD when its `origin` names a different
repository, or when its tree is dirty and so cannot be refreshed. Neither case is
auto-fixed.
```

The `refresh_skipped` gap, reproduced with real git — a clean tree on a non-default branch:

```
--- porcelain status (what fast_forward_skip_reason reads) ---
[empty == CLEAN == refresh_skipped is None]
--- what decide_update reads ---
checked_out=my-feature
default=main
[differ => decide_update returns Skip => NO fast-forward, refresh_skipped still None]
```

Same shape for `ahead > 0` (unpushed commits), detached HEAD, and the #4961 gitignored-collision refusal.

New tests, run from the PR head:

```
cargo test -p trusty-mpm --lib inproject_cold_start -- --test-threads=1
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 4657 filtered out

cargo test -p trusty-mpm --test inproject_cold_start -- --test-threads=1

cargo test -p trusty-mpm --bin tm classify -- --test-threads=1
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 1468 filtered out

EXIT=0  | 3731 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
```

No coverage was deleted: zero `#[ignore]` added, zero `#[test]` removed, `.line-cap-allowlist.tsv` untouched. The `managed_routes/mod.rs` → `delete.rs` split moves two handlers verbatim and re-exports them; route wiring is unchanged.

### Notes

- The `refresh_skipped` gap is bounded, which is why it is MEDIUM and not HIGH. `update_to_origin`'s `warn!` reaches stderr at the default filter, and `inproject_start_point::resolve` fetches independently at worktree-creation time — so no stale content reaches the session even when the base checkout is not fast-forwarded. What is wrong is the field's stated contract, not the outcome.
- Two GitHub repos whose names slugify identically (`o/cool_app` and `o/cool-app`) share one directory under `base_clone_path`. The remote check catches the collision and fails loud, though "Move or remove that directory" is unhelpful advice for that specific case. Pre-existing property of `base_clone_path`, not introduced here — on the record only.
- `tm run --help` now emits the full `Why:`/`What:`/`Test:` doc block, including the internal `Test: cli_parses_run_standalone, run_target_tests.rs` line, into user-facing help. That is a repo-wide clap-derive pattern (`--url` does the same), not this PR's defect, but this PR grew the block from 6 lines to 18.
- ADR-0030 not being flipped to Accepted is intentional and out of scope; not reported.
- Baseline failures from `docs/reference/test-ladder-baseline.md` were excluded as instructed. The author's `--test-threads=1` run is the doc's own prescribed remedy, with nothing excluded — legitimate.

PR #4991, opened 2026-08-06
feat(trusty-mpm): `tm reinstall` — redeploy assets to every destination

### LINKED ISSUE #4849: standalone driver (tm run/load/login) never refreshes agent assets from the compiled bundle
## Defect

The standalone driver does not self-heal stale agent assets. `tm run` / `tm load` / `tm login` route through `global_config::ensure_global_config_dir`, which deploys from `~/.trusty-mpm/framework/agents/` without refreshing that directory from the compiled bundle.

PR [#4848](https://github.com/bobmatnyc/trusty-tools/pull/4848) added auto-deployment of the first hop (bundle → `~/.trusty-mpm/framework/agents/`) to the managed-session path only, via `managed_config::ensure_managed_config_dir_with_root`. The standalone driver was left out.

## Impact

A machine that never starts a daemon-managed session never refreshes its agent assets, so it keeps the exact defect [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840) describes. Machines that do start managed sessions are fixed transitively, because the managed path writes the same source directory.

## Design considerations

Wiring the refresh directly into `global_config` would require resolving two existing test contracts:
- One asserts the agents directory is NOT created when the source is missing
- One hand-writes a non-bundled agent into the source directory that pruning would delete

Resolving those is real design work, not a one-line extension, so it belongs in a separate issue with its own verdict and design review.

## Related

Closes [#4840](https://github.com/bobmatnyc/trusty-tools/issues/4840) (after managed-session fix) and [#4848](https://github.com/bobmatnyc/trusty-tools/pull/4848) (which implemented the managed-path half).

bobmatnyc:
## What

`tm reinstall` — redeploys the bundled agents and skills to **every** deploy destination, and optionally reinstalls the binary.

Asset deployment is two hops, and no command covered both across all the destinations. Hop 1 materializes the compiled-in bundle into `~/.trusty-mpm/framework/{agents,skills}`. Hop 2 copies it into a deploy target — and the targets are owned by different commands:

| Written by | Destination |
|---|---|
| `tm install` | `~/.trusty-tools/trusty-mpm/claude-config/agents`, `~/.claude/skills` |
| daemon-managed session | the same, plus `<project>/.claude/skills` |
| `tm run` / `load` / `login` | `~/.trusty-mpm/claude-config/{agents,skills}` — and never runs hop 1 at all |

So redeploying through one command still left stale assets under another, with nothing saying so. `Closes #4849`.

`tm reinstall` runs hop 1 once and hop 2 into every destination, reporting deployed / repaired / preserved / unchanged / failed **per destination**. A destination that cannot be written is reported and the rest are still served; the command exits non-zero rather than printing failures under a success banner.

**Ownership** is the deployers' existing model, unchanged. A framework-owned file that drifted or was corrupted is repaired (#4408's carve-out — invalid frontmatter is corruption, not customization, and is fixed without `--force`); a file the operator owns or edited is preserved, and the note names `tm reinstall --force` as the escape hatch. `--force` clobbers, backing every file up first: agents through `agent_reset::reset_agents` (in-place `.bak-<nanos>`), skills through the new `skills::reconcile::force_adopt_bundled_skills` (copies under `~/.trusty-mpm/backup-reinstall-<ts>/` with the existing `moved-paths.log` ledger). The bundled roster stays the only admission test, so a skill matching nothing bundled is never touched however `--force` is spelled.

**The binary is not touched by default.** `--binary` routes on install provenance (#4033): a crates.io install goes through `trusty_common::update::{check_crates_io, upgrade_and_restart}`, a `cargo install --path` install rebuilds from its recorded source directory, and everything else **refuses and says why** — no ledger, no ledger record, two installs providing the same binary, a running executable that is not the file cargo placed, a version disagreement, a reaped path source, a git or unrecognised source.

## The manifest race (#4881)

`tm reinstall` can plausibly run while the daemon deploys for a session spawn, so it takes **no lock of its own** and instead routes every write through the deployers, which already hold the target directory's exclusive cross-process ledger lock (#4409 for agents, #4881 for skills). The new `force_adopt_bundled_skills` follows the same shape as its `adopt_unmanaged_bundled_skills` sibling: cheap read-only precheck outside the lock, whole load-modify-save inside it, published with `save_merging` so a lock-bypassing writer's entries are folded in rather than dropped. `reinstall_serialises_against_a_held_skill_ledger_lock` proves it blocks: with another writer holding the ledger lock, the reinstall neither completes nor writes a file, and completes once released.

**What remains racy, stated plainly:** the *sequence* is not atomic. Each destination's agent deploy, skill deploy, and force pass take and release the lock separately, so a concurrent writer can interleave between them. That is a visibility window (a destination can be observed mid-run), not a lost update — every individual load-modify-save cycle is protected. Making the whole multi-destination sweep one critical section would mean holding a machine-global lock across four directories for the duration of a full redeploy, which would block session spawns; that trade was not taken. `deploy_all_skill_tiers` already had the same per-call (not per-sequence) lock granularity before this PR.

## Gate — rung 5

Justified: process lifecycle (`--binary` can replace the running executable and exit under launchd), three concurrent manifest writers with a real regression history, cross-tier persistence.

```
cargo fmt --check && cargo check -p trusty-mpm
  && cargo clippy -p trusty-agents-common --all-targets -- -D warnings
  && cargo clippy -p trusty-mpm --all-targets -- -D warnings
  && cargo test -p trusty-agents-common && cargo test -p trusty-mpm
```

- fmt, check, both clippy runs: clean, zero warnings.
- `cargo test -p trusty-agents-common` — `test result: ok. 486 passed; 0 failed; 0 ignored`
- `cargo test -p trusty-mpm` — `test result: FAILED. 4675 passed; 1 failed; 7 ignored`
- `cargo test -p trusty-mpm -- --include-ignored` — `test result: FAILED. 4682 passed; 1 failed; 0 ignored` (all 7 gated tests ran and passed)

The one failure in both runs is `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — the documented #4931 flake, signature `Captured lines: []` (thread-local tracing subscriber losing to a global one). Pre-existence proven: `git status --porcelain crates/trusty-mpm/src/core/managed_config.rs crates/trusty-mpm/src/core/managed_config_tests.rs` is empty, and the test passes in isolation (`test result: ok. 1 passed`). This is **not** "all tests pass" — change-specific gates pass; that test is blocked by #4931.

All nine `scripts/check_*.sh` merge blockers, exit 0 each: `check_line_cap`, `check_sld`, `check_doc_numbers`, `check_test_pointers`, `check_capabilities`, `check_changelog_fragment`, `check_generation_artifacts`, `check_agent_assets`, `check_buildrs_sync`. `tm generate capabilities` was re-run and the regenerated `tm-capabilities` assets are committed. Two files went over the 500-SLOC production cap and were fixed by relocation, not allowlisting: the dispatch arm in `main.rs` was collapsed to one line, and the two new CLI parse tests moved from the production-classified `tests_behavior_a.rs` into the test-classified `tests_behavior_b_tests.rs`.

Entry point smoke-tested: `tm reinstall --help` renders.

## Tests

Failure paths and concurrency, not just happy paths:

- ownership matrix — a customized agent and a customized skill each preserved without `--force` and replaced with it; a corrupt-frontmatter agent repaired either way; an operator-authored skill untouched by `--force`
- `reinstall_creates_a_missing_destination` — a machine that never ran the standalone driver
- `reinstall_reports_a_failed_destination_and_serves_the_others` — a regular file where a skills directory belongs; the run reports the failure and still serves the remaining destinations
- `reinstall_serialises_against_a_held_skill_ledger_lock` — the #4881 race
- `binary_reinstall_tests` — one test per refusal branch, each asserting the reason names the actual condition; a route that silently degraded to "reinstall from crates.io" would fail
- `force_adopt_*` in `trusty-agents-common` — the frozen skill re-stamped and then refreshed by the next deploy, a current skill left completely alone (no backup churn), an operator skill left alone

## Note for review

`daemon::managed_routes::tests::stale_assets_for_many_*` (the #4619 load-sensitive flake) failed during development and I traced the mechanism, which may be useful on that issue: `fake_home()` mutates `$HOME` process-wide, and the fixture seeds its skill SOURCE at `FrameworkPaths::default()`'s path — which the code under test resolves by name, so it cannot be moved somewhere private. A concurrent non-serial test that resolves `default()` and self-heals that source materializes the whole bundled roster into it; each workspace then receives ~14 skills instead of the one the fixture seeded, and the read-count assertion measures 14 where it expects 1. Confirmed by dumping the read log. I reverted my exploratory fix — `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/daemon/` is empty — to keep this PR to one outcome. It passed in the final gate runs.


bobmatnyc:
## code-critic — adversarial review (rung 5)

**Head reviewed:** `2617aa189dd97b534d7807814cbefce75b53b584`

## Verdict: WARN

Zero CRITICAL. Three HIGH. No data loss, no lost-update race, no `--binary` route that can brick an install — I tried to build all three and could not. The HIGHs are a red required check, a silent no-op in exactly the class this command exists to end, and an unrequested write into `$PWD`.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/reinstall.rs` | 421–440 | `deploy_skills_into` reads `deploy.stats` and throws away `TierDeploy::shadowed`. An untracked bundled-named skill is put in `shadowed` (winner Project, loser Bundled) by `plan_skill_tiers` and removed from `bundled_deploy`, so it never reaches `deploy_skills_filtered` and lands in NONE of deployed/preserved/unchanged/failed. A bare `tm reinstall` prints zeros for it. That is the #4605 defect, and the module doc calls a silent per-tier no-op "the exact failure this command exists to end". `tm install` prints a warning line for it (`install.rs:215`); `tm reinstall` prints nothing. `shadowed` never appears anywhere in `reinstall.rs`. | Emit a note per `deploy.shadowed` entry whose `loser` is `SkillTier::Bundled`, or call `unmanaged_bundled_skills(dest, bundled_stems)` and reuse `install_skills::unmanaged_report_lines`' wording, pointing at `tm reinstall --force`. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/reinstall.rs` | 137, 297 | Required check **`Test: doc-comment pointer lint`** is RED. Both lines cite `reinstall_reports_a_failed_destination`; the test is `reinstall_reports_a_failed_destination_and_serves_the_others`. `scripts/check_test_pointers.sh`: `2 dangling pointer(s) — FAILED`. | Correct both pointers to the full test name. | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/reinstall.rs` | 40 | `let project_dir = std::env::current_dir().ok();` is unconditional, so `<cwd>/.claude/skills` becomes a WRITE destination for every invocation. `tm install` deliberately passes `None` (`install.rs:213`, `:215`) — this is the first non-session writer to take the project tier, and it takes it from whatever directory the operator happens to be in, with no project check. A bare flagless `tm reinstall` from `~/Downloads` or an unrelated repo materializes the full bundled roster (~52 skill dirs, plus `.trusty-mpm-skills-manifest.json` and a `.lock` sidecar) there. The PR's own test proves the write: `reinstall_tests.rs:157–161` asserts `SKILL.md` lands under a bare `TempDir`. An unwritable cwd additionally sets `tier.errored`, so `has_failures()` bails and the command exits 1 even when all four real destinations succeeded. | Resolve the project through the registry (or take an explicit `--project <dir>`) and pass `None` when the cwd is not a registered project. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/reinstall_tests.rs` | 226–255, 321–336 | `reinstall_preserves_a_customized_agent_without_force` hand-seeds `Origin::User`. No production path writes a non-`Bundled` origin — every insert in `agents/deployer.rs`, `agent_reset.rs` and the update-apply path is `Origin::Bundled`, and `prune_guard.rs:86` says it outright: "Nothing writes a non-`Bundled` origin today." So the test proves agent preservation through a branch (`deployer.rs:382`) production never reaches. The reachable preserve path is untracked-and-different (`deployer.rs:339`). | Re-point the test at the untracked branch (deploy, edit the file, drop its manifest entry), or add a second test that covers it. | Fix here |
| MEDIUM | `crates/trusty-agents-common/src/skills/reconcile.rs` | 240–256 | `force_adopt_locked` stamps every file `bundled_skill_dirs` reports, and `unmanaged.rs:163` fills that from `reference_files()` — every `.md` under `<stem>/references/`, including files the operator added that the bundle never shipped. Claiming a stray removes the `prune_guard::skill_verdict` guard at `prune_guard.rs:175` ("it holds X, which trusty-mpm never deployed") that currently spares the whole directory from `prune_skills`. Both paths back up, so not data loss, but `--force` silently widens what a later prune may delete. | Intersect `skill.manifest_keys()` with the files the bundled source actually supplies for that stem before stamping. | Fix here |
| LOW | `crates/trusty-mpm/src/core/reinstall.rs` | 430, 433 | Skills' `repaired` and `failed` are hardcoded 0, and `DeployStats::skipped` is rendered by `preserved_note` as "you own or edited them". `skills/deployer.rs:203` (source dir with no `SKILL.md`) and `:289` (an `mcp`-named stem) also push into `skipped`; neither is something the operator owns or edited. | Separate the rejection reasons from the ownership skips, or route them through `failed`. | Fix here |
| LOW | `crates/trusty-agents-common/src/agents/deployer.rs` | 115–126 | `DeployResult` is `pub` with `pub` fields and no `#[non_exhaustive]`, so adding `repaired` is a semver break for any external crate using a struct literal. In-tree it IS additive — verified every construction outside the defining crate uses `..DeployResult::default()` (`agent_source_tests.rs` ×5, `session_launch/mod.rs:729`, `tests_roster.rs:57`, `tests_behavior_reset_agents_tests.rs:63`). `trusty-agents-common` is on crates.io at 0.5.0. | Add `#[non_exhaustive]`, or make sure the next release is a minor bump. | Parent |

## Required Changes

1. Fix the two doc pointers at `reinstall.rs:137` and `:297` — the required check is red.
2. Surface `TierDeploy::shadowed` (or the unmanaged scan) in the per-destination notes.
3. Stop treating an arbitrary cwd as a deploy destination.

## The claims I was asked to attack

**The lock test does prove what it claims.** `with_ledger_lock` (`agents/manifest.rs:164–182`) takes a blocking `fd_lock` write lock, and `fd_lock` uses `flock(2)` on unix, which conflicts across descriptors within a single process — so the in-process test is a real cross-process proof. The test holds the lock on the standalone tier's skills dir, which `reinstall_targets` orders LAST, so the timing assertion alone would also be satisfied by "hasn't got there yet". What makes it sound is the final assertion that the marker appears once the lock releases, plus the fixture pinning both bundle stamps so hop 1 no-ops and the whole run is 2 agents × 2 dirs and 1 skill × 3 dirs — milliseconds. A 2-second non-completion is the lock, not the workload. It does not exercise the AGENT ledger lock; that is a different lock file, and the standalone agent deploy runs before the skill deploy the test blocks.

**The residual race is characterized correctly, with one addition.** I could not construct a lost update or a corrupt manifest. Every load-modify-save sits inside one lock span and publishes through `save_merging`, which replays this run's delta over the current on-disk document instead of overwriting it, and `force_adopt_locked` is never nested inside another `with_skill_manifest_lock`, so the not-reentrant self-deadlock is avoided. What the gap CAN produce is more than a torn view: if the skill file is rewritten between force-adopt's re-stamp and the deploy that follows, the ledger ends up holding a checksum for content no longer on disk and the deploy skips the file — the #4881 freeze shape, reached through the acknowledged window. It needs a writer inside a millisecond gap, the file content survives, and re-running `--force` clears it. Not filed.

**`--force` clobbers only what it should.** No path overwrites a user-customized file without `--force` beyond what #4408 already mandates for framework-owned files. The corruption carve-out still repairs without `--force` — `reinstall_repairs_a_corrupt_agent_without_force` passes, and the only change to that branch is the `repaired.push` bookkeeping, placed ahead of a write that was already unconditional there, so `repaired ⊆ deployed` holds on every `Ok` return. `force_adopt_locked` propagates a failed backup before `save_merging` runs, so a partial re-stamp cannot be published.

**`--binary` refusal is real and no route can cross.** All seven refusal branches verified (`binary_reinstall.rs:81–161`): no ledger, no record, duplicate providers, running exe ≠ cargo's file, version disagreement, git source, unrecognised source. A recorded path that no longer exists refuses at `:145–151` with `source_exists` wired to `dir.exists()` at `commands/reinstall.rs:183`. Every `Refuse` bails at `commands/reinstall.rs:187–189`. Registry and Path cannot cross — the match is on `record.source` after the same-file and version gates. Registry health-gates inside `upgrade_and_restart`; Path health-gates via `verify_installed_binary` at the call site. `upgrade_and_restart`'s `exit(1)` needs `launchd_supervision()` to find this PID in `launchctl list`, which an interactive `tm` is not.

**Four destinations, none skipped, partial failure reported.** managed config (agents + skills), operator home, project, standalone (agents + skills) — `reinstall_reports_every_target` asserts 4. A failed destination sets `tier.errored`, the loop continues, and `has_failures()` bails; `reinstall_reports_a_failed_destination_and_serves_the_others` proves the other tiers still get served. `~/.trusty-tools/trusty-mpm/claude-config` and `~/.trusty-mpm/claude-config` are genuinely different roots, so the standalone tier is not silently deduped away in the default layout.

## Gates run locally at the PR head

```
test result: ok. 486 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.89s

SKIP_UI_BUILD=1 cargo test -p trusty-mpm
test result: FAILED. 4675 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 49.78s
failures:
    daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet
```

That failure is the documented pre-existing `$HOME`-mutation flake at `docs/reference/test-ladder-baseline.md:38`. In isolation:

```
cargo test -p trusty-mpm --lib -- --exact daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4682 filtered out; finished in 0.03s
```

Cross-crate (rung 5):

```
SKIP_UI_BUILD=1 cargo check --workspace --all-targets --exclude trusty-code-gui --exclude trusty-mpm-gui \
  && cargo clippy -p trusty-agents-common -p trusty-mpm --all-targets -- -D warnings
EXIT=0
```

The two Tauri GUI crates are excluded because `tauri::generate_context!` panics on a missing `ui/dist` under `SKIP_UI_BUILD=1` — a local environment limitation, unrelated to this PR; CI builds those UIs and both pass there.

bobmatnyc:
## Review round 1 — all five findings addressed (head `66c0bf35`)

**HIGH 1 — shadowed skills were invisible.** Confirmed as described. `list_project_custom_stems` classifies every untracked skill directory at a destination as project-custom, `plan_skill_tiers` then drops the bundled stem from `bundled_deploy`, and the deployer never sees it — so it reached none of deployed/preserved/unchanged/failed. `AssetCounts` now carries `shadowed`, and `shadowed_note` names the skills in one bounded count-plus-preview line. It separates the two kinds, because they mean opposite things: a stem the operator deliberately overrode from `~/.trusty-mpm/skills` (winner `user-custom`) is the tier system working and gets no remedy pointer; a BUNDLED-named stem the destination classifies project-custom is #4605 and names `tm reinstall --force`, which adopts exactly that set. An operator-authored skill matching no bundled name is never named — the same admission rule `force_adopt_bundled_skills` applies. `reinstall_surfaces_a_shadowed_bundled_skill` pins zeros in the other four counts, a 1 in `shadowed`, the skill named in the notes, and `--force` actually fixing it.

**HIGH 2 — dangling `Test:` pointers.** Both fixed. `check_test_pointers.sh` now resolves 21972 citations with 0 dangling.

**HIGH 3 — `tm reinstall` wrote into `$PWD`.** Fixed by gating, not by dropping the tier. Rationale for that choice: passing `None` like `tm install` would abandon a destination the daemon-managed session path genuinely writes, and a project-detection predicate is a new notion of "is this a project" that can be wrong. The tier is now included only when `<project_dir>/.claude/skills` is **already a directory**, so a reinstall REFRESHES a project tier and never creates one — contamination is impossible by construction rather than by a predicate. A project never launched under tm still gets its tier from `project_skill_tier::ensure_project_skill_tier` at launch, which knows it is a genuine managed project; this command does not. The tier is also marked `optional`, so `has_failures()` ignores it: an unwritable project directory is still reported, in the same words, but does not fail a command whose three machine-global destinations succeeded. Three tests: `targets_omit_the_project_tier_until_it_exists`, `reinstall_never_writes_into_a_bare_working_directory` (asserts the bare cwd has **zero** entries afterward), and `reinstall_does_not_fail_the_command_for_an_optional_destination`.

**MEDIUM — the `Origin::User` test.** You are right that it proved nothing: no production path writes that origin, and no shipped artifact uses the `SeedOnce` tier it belongs to. Both agent-ownership tests now drive the **untracked-and-differing** state, which is what a real machine reaches — an agent file hand-placed or edited before per-file manifest tracking existed. The `Origin::User` branch stays pinned where it belongs, at the unit level, by `deploy_preserves_modified_user_owned_entry` in `trusty-agents-common`. The fixture's doc comment records why it is not seeded here.

**MEDIUM — `force_adopt_locked` stamping reference strays.** Narrowed. Inside a skill the ledger **already tracks**, only files the ledger also tracks are re-stamped. An operator's `references/*.md` dropped next to a managed skill is left untracked, so it stays in `prune_guard::skill_verdict`'s spared set instead of becoming tm-managed and prunable. When the skill's own entry point is untracked the whole directory is still adopted, references included — that is the #4605 state, and it is what `adopt_unmanaged_bundled_skills` already does. `force_adopt_leaves_an_operator_reference_stray_untracked` asserts `adopted_keys == ["tm-workflow"]` only, the stray untracked, and its bytes unchanged. Your read that it was not data loss was correct either way — both paths back up first.

## Gate — rung 5, re-run

```
cargo fmt --check && cargo check -p trusty-mpm
  && cargo clippy -p trusty-agents-common --all-targets -- -D warnings
  && cargo clippy -p trusty-mpm --all-targets -- -D warnings
  && cargo test -p trusty-agents-common && cargo test -p trusty-mpm
```

fmt, check and both clippy runs: exit 0, zero warnings.

```
  test result: ok. 487 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm            (default parallelism)
  test result: FAILED. 4680 passed; 2 failed; 7 ignored; 0 measured; 0 filtered out
    ensure_managed_config_dir_emits_the_frozen_skill_warning   (#4931)
    stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet  (baseline row 38)

cargo test -p trusty-mpm --lib -- --test-threads=1
  test result: ok. 4682 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm -- --include-ignored --test-threads=1
  test result: ok. 4689 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
  test result: FAILED. 0 passed; 1 failed  — meta_run_demo_writes_and_verifies_artifact (baseline row 40)
```

Both parallel failures are the documented isolation flakes and both disappear under `--test-threads=1`, which is the documented discriminator; all 7 previously-ignored lib tests ran and passed. The integration failure is baseline row 40 — it spawns a real Claude Code session and reports `launch_outcome: "timed-out"` in this environment; `git status --porcelain crates/trusty-mpm/tests/` is empty. This is not "all tests pass": change-specific gates pass, and those three are blocked by their canonical issues.

All nine merge-blocking scripts, exit 0 each:

```
check_line_cap        3730 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
check_sld             56 spec doc(s) + 3108 code file(s); 0 error(s), 0 warning(s)
check_doc_numbers     102 doc(s) / 96 claim(s); 4 grandfathered, 0 violations — OK.
check_test_pointers   21972 Test: citation(s) — 0 dangling pointers — OK.
check_capabilities    up to date (7 generated files).
check_changelog_fragment   OK trusty-mpm; OK trusty-agents-common
check_generation_artifacts 16 stylesheet(s) + 1233 prose file(s) — 0 leaked artifacts — OK.
check_agent_assets    30 byte-parity file(s), all match; 4 pinned deviations match upstream — OK.
check_buildrs_sync    canonical blocks in sync across all 4 crates.
```

`tm generate capabilities` re-run; the regenerated assets are committed.

bobmatnyc:
## Rebased onto `main` — the line-cap check was red, now fixed (head `fc15614f`)

`main.rs` sits against the 500-SLOC production cap. `tm shell-init` (#4986) landed one dispatch line while this branch was in review, and this branch added two — an alias import plus the arm — so the merged file measured 501. The check passed on my pre-rebase tree and failed in CI, which measures the merge result; after rebasing I reproduced it locally.

Fixed by shrinking my own footprint to one line rather than allowlisting: the three flags now ride a `clap::Args` struct (`ReinstallArgs`), so the arm fits on one line with a fully-qualified path and no import — the same shape `tm shell-init` uses. The parsed CLI surface is unchanged; `--force`, `--binary`, and `--yes` still parse identically and `--yes` still requires `--binary`, pinned by the two existing parse tests. `tm reinstall --help` verified.

The rebase itself was clean — no conflict with #4986 despite both touching `cli/mod.rs` and `main.rs`. `tm-capabilities` was regenerated afterward, since it is produced from the live clap surface and had to pick up both new subcommands.

## Gate — rung 5, on the rebased tree

fmt, `cargo check -p trusty-mpm`, and both `clippy --all-targets -- -D warnings` runs: exit 0, zero warnings.

```
  test result: ok. 487 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm --lib -- --test-threads=1
  test result: ok. 4685 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm -- --include-ignored --test-threads=1
  test result: ok. 4692 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
  … 12 further integration targets ok …
  test result: FAILED. 0 passed; 1 failed — meta_run_demo_writes_and_verifies_artifact
```

The single failure is baseline row 40: it spawns a real Claude Code session and reports `launch_outcome: "timed-out"` in this environment. `git status --porcelain crates/trusty-mpm/tests/` is empty. The two parallel-run flakes from the previous comment (#4931, baseline row 38) are absent under `--test-threads=1`. Change-specific gates pass; that one test is blocked by its canonical issue.

All nine merge-blocking scripts exit 0, including `check_line_cap` and `check_capabilities`, which were the two red ones.

bobmatnyc:
## code-critic re-review (round 2) — Verdict: APPROVE

Head `fc15614f694162b42763dcf703a648a48bb9ac30`, reviewed against `origin/main` = `3df37076`. Scoped to the fix round (`fbb815c1`), the clap-`Args` refactor (`92c0173b`), the catalog regen (`fc15614f`), and the rebase. Prior verdict: https://github.com/bobmatnyc/trusty-tools/pull/4991#issuecomment-5199778342

### Prior findings — all closed

**HIGH 1 — shadowed skills were invisible. RESOLVED.** `AssetCounts::shadowed` (`core/reinstall.rs:84`) is fed from `TierDeploy::shadowed` at `core/reinstall.rs:502`, and `shadowed_note` (`core/reinstall.rs:543`) names them. The two-way split is correct: the actionable filter is `s.winner == SkillTier::Project && bundled_stems.contains(&s.stem)` (`:558`), so a deliberate `~/.trusty-mpm/skills` override arrives as `winner == User` and gets no remedy pointer, a bundled-named stem the destination classifies project-custom gets `tm reinstall --force`, and an operator-authored skill matching no bundled name is never named — which matches `force_adopt_bundled_skills`'s own admission rule exactly, so the pointer never promises a remedy that would not fire. Bounded: both previews go through `agent_source::preview(_, 5)`, which truncates to five and appends `…`, so a 110-skill roster still prints one line.

`reinstall_surfaces_a_shadowed_bundled_skill` (`core/reinstall_tests.rs:512`) proves what it claims. It drives the production `reinstall_assets`, asserts `deployed == preserved == unchanged == 0` and `shadowed == 1` — the exact "prints zeros" state — asserts the note names `demo-skill` and carries `tm reinstall --force`, then runs the force pass and asserts the file on disk now contains the bundled text. That last assertion is what makes it a proof rather than a count check.

**HIGH 2 — dangling `Test:` pointers. RESOLVED.**
```
test-pointers: resolved 21985 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```
(21985 vs the 21972 in the PR body — main moved under the branch.) CI agrees: `Test: doc-comment pointer lint` pass.

**HIGH 3 — `$PWD` contamination. RESOLVED.** The gate is `core/reinstall.rs:230`, `project_dir.filter(|dir| dir.join(".claude").join("skills").is_dir())`, so the tier is refreshed and never created. I could not find a remaining path that writes into a bare cwd: `backup_root` resolves under `paths.root`, not cwd; `force_adopt_bundled_skills` returns early on an empty `bundled_skill_dirs` scan before taking a lock, so no sidecar lands either; and `list_project_custom_stems` early-returns on a non-directory.

`optional` does not mask a required failure. It is set only for `tier.label == "project"` (`:235`), and `skill_deploy_tiers` already dedups, so the managed-workspace case where `claude_skills_dir` *is* the project dir keeps the `operator home` label and stays non-optional. `has_failures` filters `!t.optional` (`:182`). `reinstall_reports_a_failed_destination_and_serves_the_others` pins the other direction: a failing **managed config** tier does return `has_failures() == true`.

`reinstall_never_writes_into_a_bare_working_directory` (`core/reinstall_tests.rs:629`) exercises the production path — it calls `reinstall_assets` with a bare tempdir as the project and asserts both `!bare/.claude.exists()` and `read_dir(bare).count() == 0`.

**MEDIUM — the dead-branch test. RESOLVED.** Both agent-ownership tests now use `seed_untracked_agent` (`core/reinstall_tests.rs:334`), which writes the file and no manifest — the untracked-and-differing state a real machine reaches. The `Origin::User` branch stays pinned at unit level by `deploy_preserves_modified_user_owned_entry` (`agents/deployer_tests.rs:123`), which is pre-existing, deploys twice to prove the preservation is not one-shot, and asserts the origin survives. Preservation is proved by content equality, not just by the count.

**MEDIUM — `force_adopt_locked` narrowed. RESOLVED, and the narrowing matters.** `skills/reconcile.rs` guards with `if entry_managed && !manifest.is_managed(&key) { continue; }`. Since the entry-point manifest key *is* the stem, `entry_managed` is exactly "the ledger tracks this skill", so inside a tracked skill only tracked keys are re-stamped. I confirmed the stated consequence at `core/update_check/apply/prune_guard.rs:172` — `skill_verdict` returns `Kept` the moment it finds one file no ledger key claims, so leaving the stray untracked is what keeps the whole directory out of the prunable set. `force_adopt_leaves_an_operator_reference_stray_untracked` asserts the entry point *was* re-stamped while `tm-workflow/references/my-notes.md` stayed untracked with its bytes intact.

The remaining adopt-whole-directory case cannot reach unrelated files. `bundled_skill_dirs` only descends into `<dest>/<stem>/` where `stem ∈ bundled` and `SKILL.md` exists, and `reference_files` takes only `references/*` passing `is_skill_file`. Nothing outside that shape is visible, content is never rewritten, and every touched file is copied to `backup_root` first. It is the same scope `adopt_unmanaged_bundled_skills` already had.

### The rebase — nothing dropped or duplicated

Against `origin/main` = `3df37076` (which contains #4986 `tm shell-init` and #4987), the PR is 5 commits and the two contested files are pure additions: `cli/mod.rs` +50/-0, `main.rs` +1/-0. `shell-init` survives in both `cli/mod.rs` and `cli/actions/shell.rs`.

The catalog needed two passes and landed right. `92c0173b` was built on the pre-rebase tree and its `cli.md` dropped the `shell-init` line; `fc15614f` regenerated and restored it. Net vs main: `56 → 57`, one added `reinstall` row, `shell-init` present at `cli.md:188`, and `grep -c '^- \`'` counts exactly 57. CI's `tm-capabilities generated-skill drift check` passes, which is the mechanical proof that the catalog matches the live `Cli::command()` surface.

### The SLOC refactor did not change the CLI surface

`Reinstall { force, binary, yes }` → `Reinstall(ReinstallArgs)`. Field names, types, and every `#[arg]` attribute are carried over byte-for-byte, including `#[arg(long, requires = "binary")]` on `yes`. clap flattens a single-field tuple variant whose type derives `Args`, so `--force` / `--binary` / `--yes` parse identically. `cli_parses_reinstall` still asserts all three default to `false`, and `cli_parses_reinstall_binary` still asserts `reinstall --yes` without `--binary` is an error. `500-line file-size cap` passes; `main.rs` is not allowlisted.

### Gates

```
cargo test -p trusty-agents-common && cargo test -p trusty-mpm
  → 8453 passed; 0 failed; 17 ignored   (exit 0)
cargo fmt --check && cargo clippy -p trusty-agents-common --all-targets -- -D warnings
                  && cargo clippy -p trusty-mpm --all-targets -- -D warnings
  → LINT_EXIT=0
bash scripts/check_test_pointers.sh → 21985 citations, 0 dangling — OK
bash scripts/check_line_cap.sh      → 3733 files, 7 allowlisted, 0 violations — OK
```

Nothing was removed, `#[ignore]`d, `cfg`-gated, `--exclude`d, or `--lib`-narrowed: the diff adds 39 `#[test]` functions and removes zero, and contains no added `#[ignore]` or `#[cfg(not(…))]`. The three known baseline rows are intact — `ensure_managed_config_dir_emits_the_frozen_skill_warning` and all four `stale_assets_for_many_*` ran and passed this time, and `meta_run_demo_writes_and_verifies_artifact` keeps its pre-existing `#[ignore]` with the same `#1053` reason string.

### Findings

|---|---|---|---|---|---|
| LOW | `crates/trusty-mpm/src/core/reinstall.rs` | 502, 552–560 | A stem present in project + user + bundled produces TWO `Shadow` records (`skills/tiers.rs:161` and `:173`), so `shadowed` counts it twice, the preview names it twice, and the actionable count says "2 of them" for one skill. Over-reports, never under-reports. | Dedup by stem: collect `actionable` into a `BTreeSet<String>`, and build `names` from one record per stem (highest-precedence winner). | Fix here |
| LOW | `crates/trusty-mpm/src/core/reinstall.rs` | 218 | The doc still says `reinstall_targets` is "Pure path arithmetic — nothing is read or created", but line 230 now stats the filesystem. The purity claim is the stated reason the target set is assertable without a real home dir. | Say it reads one `is_dir` and creates nothing. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/reinstall.rs` | 12 | Module doc says "``[`reinstall`]`` resolves the real framework and standalone layouts", but `92c0173b` renamed that function to `run`. | ``[`run`]`` | Fix here |

### Notes

One thing I looked at and am not flagging: there is a TOCTOU window between the `is_dir` check at `reinstall.rs:230` and the project tier's own deploy, spanning the hop-1 refresh and up to three earlier tier deploys. Deleting `<cwd>/.claude` inside that window makes the deploy recreate it. That re-creates a directory that existed when the operator started the command, in the operator's own project — not contamination of an unrelated directory, and not worth a guard.

No CRITICAL and no HIGH findings. APPROVED for the next pipeline stage. `Test` was still pending in CI at the time of writing — this verdict covers the code, not the CI rollup.

bobmatnyc:
## Review round 2 — three LOW findings fixed (head `2e2c1a8c`)

**LOW 1 — double-counted shadow.** Confirmed: `plan_skill_tiers` emits one `Shadow` per LOSING tier, so a stem the project, user, and bundled tiers all carry produced two records for one skill, and the report counted records. `distinct_shadowed` now collapses to one entry per stem. Every record for a given stem carries the same winner — the planner picks the nearest higher tier once per stem, and only the loser differs — so nothing is lost by collapsing. `shadowed_note_counts_one_skill_once` builds the exact two-record case and asserts "1 skill(s)", "1 of them", and the name appearing exactly twice in the line (once in the name list, once in the actionable list).

You were right that it is the same class as the HIGH: a count that misstates what happened, in the one feature whose purpose is an accurate count.

**LOW 2 — doc claimed purity.** `reinstall_targets` said "Pure path arithmetic — nothing is read or created" after gaining the `is_dir` probe. It now says: path arithmetic plus one filesystem probe, naming which probe and pointing at the PROJECT TIER paragraph that explains it. Nothing is created and nothing but that probe is read.

**LOW 3 — stale doc link.** `[reinstall]` → `[run]`.

**Not fixed, per your instruction:** the `is_dir`/deploy TOCTOU. Deleting `<cwd>/.claude` mid-run makes the deploy recreate a directory that existed at command start, inside the operator's own project — not contamination. No guard added.

## Gate — rung 5

fmt, `cargo check -p trusty-mpm`, and both `clippy --all-targets -- -D warnings` runs: exit 0, zero warnings.

```
  test result: ok. 487 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-mpm -- --include-ignored --test-threads=1
  test result: ok. 4693 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
  … 14 further integration targets, all ok …
  test result: FAILED. 0 passed; 1 failed — meta_run_demo_writes_and_verifies_artifact
```

The single failure is baseline row 40: it spawns a real Claude Code session and reports `launch_outcome: "timed-out"` here. `git status --porcelain crates/trusty-mpm/tests/` is empty. Change-specific gates pass; that test is blocked by its canonical issue.

All nine merge-blocking scripts exit 0:

```
check_line_cap        3733 tracked .rs file(s); 7 allowlisted, 0 violations — OK.
check_sld             56 spec doc(s) + 3110 code file(s); 0 error(s), 0 warning(s)
check_doc_numbers     102 doc(s) / 96 claim(s); 4 grandfathered, 0 violations — OK.
check_test_pointers   21986 Test: citation(s) — 0 dangling pointers — OK.
check_capabilities    up to date (7 generated files).
check_changelog_fragment   22 changed path(s); both crates with source changes carry a fragment
check_generation_artifacts 16 stylesheet(s) + 1236 prose file(s) — 0 leaked artifacts — OK.
check_agent_assets    30 byte-parity file(s), all match; 4 pinned deviations match upstream — OK.
check_buildrs_sync    canonical blocks in sync across all 4 crates.
```

Issue #4728, opened 2026-08-04
SHIPPED (1.3.3/1.3.4): tm ls auto-prune can SIGTERM and kill a live session — record_only never gated the runtime teardown


bobmatnyc:
Found during the [PR #4725](https://github.com/bobmatnyc/trusty-tools/pull/4725) review gate (credit: code-critic). Filing separately because it is **independent of and older than that PR** — it ships in 1.3.3 and 1.3.4, which is what is installed on the dogfood machine right now.

## Defect

`SessionManager::decommission_record_only` is documented and named as a non-destructive path. It is not. In `decommission_with_root`, the runtime teardown ran **above** the `record_only` guard:

```rust
// crates/trusty-mpm/src/session_manager/decommission.rs (pre-fix)
self.graceful_terminate_runtime(&record.tmux_name).await;   // line 435 — unconditional

let mut workspace_removed = false;
if !record_only && let Some(ref ws) = record.workspace_path {  // line 444 — the guard
```

`graceful_terminate_runtime` SIGTERMs the claude process and then calls `kill_session`. Its only self-guard is:

```rust
// session_manager/restart_ops.rs:143
if !self.tmux.session_exists(tmux_name) { return; }
```

and `session_exists` is plain live-tmux **name membership**:

```rust
// session_manager/driver.rs:363
fn session_exists(&self, name: &str) -> bool {
    self.list_sessions().map(|names| names.iter().any(|n| n == name)).unwrap_or(false)
}
```

Never the record's captured `pane_id`. That routes around the pane-scoped containment #3714 added for display.

## Why it ships and is reachable

The `tm ls` / picker auto-prune from #4384 (`3dfcb657`, in 1.3.3+) calls this path for every record the daemon flags `unresumable`, on every interactive listing.

**No name collision is required.** `is_unresumable` gates on persisted state (`Stopped`/`Errored`) plus filesystem probes — **it never consults tmux liveness**. And `mark_errored` is a pure record mutation that never touches the pane:

```rust
// manager.rs:1106
pub async fn mark_errored(&self, id: &ManagedSessionId, error_msg: &str) -> Result<(), ManagedError> {
    let mut record = self.get(id).await?;
    record.state = ManagedSessionState::Errored;
    ...
}
```

So any `Stopped`/`Errored` record whose workspace was deleted while its pane is still alive is a live kill target on the next `tm ls`.

[...]

PR #4995, opened 2026-08-06
fix(trusty-mpm): hide dead (unresumable) records from the default `tm ls` view

### LINKED ISSUE #4994: tm ls default view shows dead (unresumable) session records
`is_live_session_state` (`crates/trusty-mpm/src/bin/tm/commands/managed.rs:106`) decides what the default `tm ls` / picker view shows. It excludes only `decommissioned` and non-tombstone `deleted`; every other state takes the catch-all `_ => true` arm:

```rust
pub(crate) fn is_live_session_state(state: &str, is_slot_tombstone: bool) -> bool {
    match state {
        "decommissioned" => false,
        "deleted" => is_slot_tombstone,
        _ => true,
    }
}
```

It never consults the daemon's `unresumable` flag. So a record with `state: "stopped"` and `unresumable: true` — no tmux pane, no workspace directory anywhere on disk, genuinely unrecoverable — renders in the default view.

Evidence, owner's machine 2026-08-06: `GET /api/v1/sessions/managed` returns 116 rows, 6 with `unresumable: true` (slots 108–113). All 6 render `[dead]`, matching the flag one-for-one. `tm ls` additionally prints `tm: 6 more dead records pending confirmation`.

The detection is correct and already reaches the CLI — `ManagedSessionSummary::unresumable` (`crates/trusty-mpm/src/client/http_client/types.rs:512`) deserializes it and `render_session_table` prints `[dead]` off it. The filter simply does not act on it.

Resolution: hide `unresumable` records from the DEFAULT view, keep them under `tm ls --all` exactly as `decommissioned` records already are. Visibility only — no record is deleted, decommissioned, or mutated. The filter must run AFTER the listing-time auto-prune (#4702), or the sweep loses the records it exists to reap.

Note on the `pending confirmation` banner: not a second defect. `SIGHTING_MIN_AGE_SECS` is 10 minutes and `AUTO_PRUNE_CAP` is 5 per call, so a newly-dead record is legitimately visible for at least one window. The 6 current rows were first-sighted 8 minutes before this report (`~/.trusty-mpm/auto-prune-seen.json`, all six stamped `2026-08-06T04:21:54Z`). Their names (`tm-deadrt*`, `tm-tmprbhxrt-01`) and `cwd: /unknown` show they are leaked artifacts of test runs against the live daemon, so fresh ones keep arriving and the banner looks permanent while individual records do age out.

bobmatnyc:
Closes [#4994](https://github.com/bobmatnyc/trusty-tools/issues/4994).

## Defect

`is_live_session_state` (`crates/trusty-mpm/src/bin/tm/commands/managed.rs`) decided default-view visibility from lifecycle state alone — `decommissioned` hidden, non-tombstone `deleted` hidden, everything else through the catch-all `_ => true`. A record with `state: "stopped"` and `unresumable: true` (no tmux pane, no workspace directory left on disk) therefore rendered in the default `tm ls`, tagged `[dead]`.

The detection was already correct and already reaching the CLI: `ManagedSessionSummary::unresumable` (`crates/trusty-mpm/src/client/http_client/types.rs:512`) deserializes the daemon's verdict, and `render_session_table` prints `[dead]` off it. Only the filter ignored it.

## Resolution

`unresumable` becomes a third input to the same predicate, placed BELOW the `"deleted"` arm so a #3034 slot tombstone still renders at its slot regardless of the flag. Visibility only — no record is deleted, decommissioned, or mutated, and `tm ls --all` lists every one.

**The filter had to move.** `parse_scoped_sessions` applied it upstream of the listing-time auto-prune (#4702). Hiding dead rows there would have starved the sweep of the exact records it reaps, so they would accumulate in the store forever, now invisibly. Split into `parse_managed_sessions` (deserialize) and `scope_for_display` (the single visibility gate), and run parse → prune → scope on both listing paths.

The `N more dead records pending confirmation` line now names where those rows went, since they are no longer on screen to reconcile the count against.

## Row counts, owner's live daemon

| | wire rows | default view | hidden by this PR | `--all` |
|---|---|---|---|---|
| before | 116 | 40 | — | 116 |
| after | 116 | 34 | 6 | 116 |

18 resumable `stopped` rows survive unchanged — hiding those would be the regression, and `is_live_session_state_keeps_resumable_stopped_record` pins it.

## Not a second defect: the "pending confirmation" banner

`SIGHTING_MIN_AGE_SECS` is 10 minutes and `AUTO_PRUNE_CAP` is 5 per call, so a newly-dead record is legitimately visible for at least one window. The six rows in the report were first-sighted 8 minutes earlier (`~/.trusty-mpm/auto-prune-seen.json`, all six stamped `2026-08-06T04:21:54Z`, marker mtime matching). Their names (`tm-deadrt*`, `tm-tmprbhxrt-01`) and `cwd: /unknown` show they are leaked artifacts of test runs against the live daemon — a re-fetch 40 minutes later showed 8, not 6. Fresh ones keep arriving, so the banner looks permanent while individual records do age out. Left alone; the test leakage is a separate matter.

## Mutation check

| mutation | tests that went red |
|---|---|
| revert to `_ => true` | `is_live_session_state_hides_unresumable_record`, `picker_filter_hides_dead_records_keeps_resumable_stopped`, `scope_for_display_all_keeps_dead_record_visible` |
| over-fire: hide on `stopped`/`errored` state instead of the flag | `is_live_session_state_keeps_resumable_stopped_record` + 5 pre-existing |
| reorder: `unresumable` arm above the `"deleted"` arm | `is_live_session_state_keeps_slot_tombstone_even_when_unresumable` only |
| scope BEFORE prune | `session_ls_prunes_dead_records_on_piped_invocation` |

Each row is absent for the reason the test names, not by accident.

## Gate — rung 3, one crate

```
cargo check -p trusty-mpm                              EXIT=0
cargo test -p trusty-mpm --no-fail-fast — 4660 passed, 1 failed (lib); 1507 passed, 0 failed (bin tm); every integration target ok
```

The single failure is `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` — baseline row 38, an isolation-shaped flake. It passes under `--test-threads=1`:

```
test daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet ... ok
test result: ok. 5 passed; 0 failed; 1 ignored; 0 measured; 4662 filtered out
```

This diff touches zero files in that test's blast radius (`daemon/managed_routes/summary.rs` is a doc-comment-only edit).

All nine repo gates: `check_line_cap` `check_sld` `check_doc_numbers` `check_test_pointers` `check_capabilities` `check_changelog_fragment` `check_generation_artifacts` `check_agent_assets` `check_buildrs_sync` — EXIT=0 each.

## Seeing and acting on hidden records

`tm ls --all` lists them at their slots, `[dead]` intact. The remedy verb is unchanged: `tm sessions delete <id> --force`, the one `unresumable_remedy_line` cites. The auto-prune also still reaps them on its own, which is the whole reason the filter runs after it.


bobmatnyc:
## Verdict: BLOCK

Reviewed at `23da737b4`. The restructuring is right — parse → prune → scope is the only order that works, and the prune's candidate set is provably unchanged (everything `filter_live_sessions` used to drop upstream is rejected by `is_clearable_state` anyway). The `--all` sort, the slot-tombstone ordering, and the `#[serde(default)]` failure direction (missing field → `false` → visible) all hold.

What does not hold is the predicate's input. `unresumable` is a daemon verdict about a record's PERSISTED state, computed before the display state is reconciled and without any liveness signal. The auto-prune spent three review rounds (#4725, #4728, #4872) building `live_tmux_names` + `workspace_verified_gone` + `absence_corroborated` specifically so it would never act on that flag alone. This filter acts on it alone — so the set it hides is strictly larger than the set the prune will ever reap.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-mpm/src/bin/tm/commands/managed.rs` | 134 | `_ => !unresumable` hides a session with a LIVE tmux pane, which the auto-prune then refuses to reap — hidden AND unreapable | key visibility off the prune's own confirmed-dead verdict, not the raw wire flag | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs` | 1852 | the `default_view` rewrite makes three post-prune assertions vacuous — proven by mutation | assert the record went `decommissioned` against the raw list | Fix here |
| HIGH | `crates/trusty-mpm/src/bin/tm/commands/managed.rs` | 134 | an unmounted volume silently empties `tm ls` — `absence_corroborated` keeps those rows, the filter hides them, no banner fires | same fix | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 305 | the banner says "hidden" on `--json` and `--all`, where the rows are printed | make the suffix conditional on the scope | Fix here |

---

### CRITICAL — a live session can be hidden, and cannot be reaped

`checked_summaries_with` computes `unresumable` from the PERSISTED state (`summary.rs:606-616`). `reconcile_live_state` then overwrites the DISPLAY `state` (`summary.rs:213-218`) and never recomputes the flag. So the wire can ship `state: "active"` alongside `unresumable: true`.

Proven, not argued — two tests added to `daemon/managed_routes/tests.rs` at this head:

```
running 2 tests
test daemon::managed_routes::tests::critic_proof_4995_live_pane_can_ship_active_plus_unresumable ... ok
test daemon::managed_routes::tests::critic_proof_4995_errored_with_live_detached_pane_ships_unresumable ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 4668 filtered out
```

The first builds a record persisted `Stopped` with every workdir candidate gone and a live pane, and asserts the summary comes out `state == "active"` **and** `unresumable == true`. The second does the same for `Errored`, which `reconcile_live_state` deliberately leaves as persisted.

Both wire shapes hit `is_live_session_state`'s new `_ => !unresumable` arm and are hidden. Neither is reapable:

- `state: "active"` — `is_clearable_state` requires display state `stopped|errored` (`session_picker_prune.rs:511`), so the sweep never sees it.
- `state: "errored"` + live pane — `live_tmux_names.contains(&s.name)` rejects it (`session_picker_prune.rs:514`).

That second guard exists for exactly this outcome. Its own doc, `session_picker_prune.rs:487-490`:

> Tombstoning such a record is not cosmetic — `Decommissioned` is TERMINAL, so a running agent silently becomes unreachable through `tm`: no resume, no reattach, hidden from the picker.

The filter produces that end state without the tombstone.

Reachability is not exotic. `reconcile_live_state` is read-only (`summary.rs:147-148`), so after a daemon restart every record reads persisted `Stopped` until something writes it. Any session whose worktree was removed per this repo's own post-merge cleanup (`git worktree remove --force`, CLAUDE.md) while its pane kept running lands here — running agent, attachable, absent from `tm ls`.

**Fix.** `is_dead_record` already computes the authoritative answer, in the same pipeline, immediately before `scope_for_display` runs. Hide the rows the prune classified `dead` and held back for confirmation or the cap — not the rows carrying the raw flag. That buys three things at once: no live-pane row is ever hidden; hidden ⟺ prunable, so nothing hidden is unreapable; and the hidden set becomes exactly the set the new "pending confirmation" banner is counting, which is what the banner already claims.

A narrower patch — `_ => !(unresumable && matches!(state, "stopped" | "errored"))` — closes the `active` instance only. It leaves the errored-with-live-pane instance and the unmounted-volume case below.

### HIGH — the `default_view` rewrite made three assertions vacuous

`seed_dead_session` fixtures carry `unresumable: true`, so `default_view` drops the record on the flag whether or not the prune did anything. Under a mutation where `decommission_dead_record` returns `Pruned` without ever contacting the daemon:

```
running 3 tests
test tests_behavior_d::auto_prune_dead_records_removes_confirmed_unresumable_records ... ok
test tests_behavior_d::session_ls_json_passthrough_prunes_dead_records ... ok
test tests_behavior_d::session_ls_prunes_dead_records_on_piped_invocation ... ok

test result: ok. 3 passed; 0 failed
```

Same mutation, with `default_view` reverted to pre-PR scoping (`is_live_session_state(state, deleted, false)`):

```
```

```
thread 'tests_behavior_d::session_ls_prunes_dead_records_on_piped_invocation' panicked at
crates/trusty-mpm/src/bin/tm/tests_behavior_d_tests.rs:2654:5:
a non-TTY `tm ls` must clear a confirmed dead record from the registry
```

The registry check at `:1969`, `:2649`, `:2689` is the only thing proving the prune reaches the daemon at all. It no longer does.

**Fix.** Keep those three on the raw list and assert what the prune actually produces — `s.persisted_state == "decommissioned"`, or the record's absence from a filter that does not consult `unresumable`. Do not route a registry-teardown assertion through a filter keyed on the flag the fixture sets.

### HIGH — unplugging a volume silently empties `tm ls`

`absence_corroborated` (`session_picker_prune.rs:449`) refuses to call a record dead when its whole checkout is unreachable. Its doc, line 355-358: *"This errs toward keeping — costing one stale row and risking nothing."* The daemon's `is_unresumable` has no such guard — a bare `try_exists` returns `Ok(false)` for every path on an unmounted volume — so those records ship `unresumable: true`.

The cost is no longer one stale row. The rows disappear, and they never enter `dead`, so `outcome.pending` stays 0 and no banner fires. An operator who unplugs a drive sees their sessions vanish with no message. It self-heals on remount and `--all` recovers them, which is why this is HIGH and not CRITICAL. The CRITICAL's fix closes it too — the prune already declines to call these dead.

### MEDIUM — the banner claims rows are hidden when they are printed

`prune_and_report_at:305` appends the "hidden" suffix unconditionally, but it runs on both `session_ls_at` branches. On `--json` the raw body is echoed verbatim; on `--all` `scope_for_display` keeps everything. Captured from a real run above:

```
tm: 1 more dead record pending confirmation (hidden; `tm ls --all` to see them)
{"sessions":[{"id":"7b391178-…","state":"errored",…,"unresumable":true,…}]}
```

**Fix.** Pass the effective scope into `prune_and_report_at` and append the suffix only when the rows were actually dropped.

## Required Changes

1. Stop deriving default-view visibility from the raw `unresumable` wire flag. Hide only what the prune classified dead — a live pane must never be hidden, and nothing hidden may be unreapable.
2. Restore the registry-teardown assertions at `tests_behavior_d_tests.rs:1969`, `:2649`, `:2689` so they fail when the prune does not reach the daemon.
3. Make the "hidden" suffix conditional on the scope that actually hid the rows.

## Notes

Verified and clean, not findings:

- **Prune candidate set** — strictly wider than before, never narrower. `filter_live_sessions` only ever dropped `decommissioned` and non-tombstone `deleted`; `is_clearable_state` rejects both. `AUTO_PRUNE_CAP` and the sighting window are untouched, and the `&&` short-circuit means no extra filesystem probes.
- **Ordering on every path** — `session_ls_at` table (`managed.rs:255-272`), `session_ls_at --json` (`:257-267`, prune runs, no scope), `fetch_live_sessions` (`session_picker.rs:478-481`, feeding the TTY picker at `:1034`/`:1113` and `guided.rs:372`), and `--all`. All four parse → prune → scope.
- **`#[serde(default)]` direction** — a missing field deserializes `false`, which means visible. An older daemon or a partial response cannot hide a live session this way.
- **Slot tombstones** — `"deleted" => is_slot_tombstone` shadows the `_` arm, so the new input cannot reach a tombstone. Your ordering mutation reproduces exactly as claimed: hoisting an `if unresumable { return false }` above the match turns `is_live_session_state_keeps_slot_tombstone_even_when_unresumable` red and nothing else (12 passed, 1 failed).
- **Resumable stopped rows** — `is_live_session_state("stopped", false, false)` is `true`, pinned. The 18 rows survive.
- **Test-function list vs `origin/main`** — additive plus one rename. Nothing dropped: `+default_view`, `+is_live_session_state_hides_unresumable_record`, `+is_live_session_state_keeps_resumable_stopped_record`, `+is_live_session_state_keeps_slot_tombstone_even_when_unresumable`, `+picker_filter_hides_dead_records_keeps_resumable_stopped`, `+scope_for_display_all_keeps_dead_record_visible`, `parse_scoped_sessions_all_keeps_tombstone_in_slot_order` → `scope_for_display_all_keeps_tombstone_in_slot_order`. The weakening at HIGH is behavioral, not a deletion.

The `tm-deadrt*` leak and the pending-confirmation banner's apparent permanence were out of scope and are not reported.

Reviewed read-only in a throwaway worktree at `23da737b4`, since removed. No PR files were modified.

bobmatnyc:
Re-pushed at `7b3842745` addressing all four findings.

**CRITICAL + HIGH (unmounted volume) — visibility now keys off the sweep's verdict.** `AutoPruneOutcome` carries `dead_ids`: the ids `is_dead_record` classified dead and the sweep could not clear this call. `prune_and_report_at` returns a `PrunedListing { sessions, dead_ids }`, and `scope_for_display` filters against that set. `is_live_session_state`'s third input is now `classified_dead`, never the wire flag. `hidden ⟺ prunable` holds by construction, so a row carrying `unresumable: true` with a live pane stays visible, and an unmounted volume's rows stay visible too — `absence_corroborated` already declines to call them dead.

**MEDIUM — banner suffix.** `prune_and_report{,_at}` take `hides_dead_rows`. `session_ls_at` passes `!all` on the table branch and `false` on `--json`; `fetch_live_sessions` passes `!all`.

**HIGH — vacuous teardown assertions.** `default_view` is gone. `default_view_pre_4994` scopes with an EMPTY dead set, so only a real `decommissioned` flip hides the record.

## Mutation check

| mutation | result |
|---|---|
| `decommission_dead_record` returns `Pruned` without contacting the daemon | `test result: FAILED. 0 passed; 3 failed` — `auto_prune_dead_records_removes_confirmed_unresumable_records`, `session_ls_json_passthrough_prunes_dead_records`, `session_ls_prunes_dead_records_on_piped_invocation` |
| revert to the CRITICAL (`s.unresumable` instead of `dead_ids.contains`) | `11 passed; 3 failed` — including `picker_filter_keeps_a_live_row_flagged_unresumable`, the guard for the exact wire shape your two proofs produce |
| remove the fix (`_ => true`) | `11 passed; 3 failed` |

## Gate

```
cargo fmt --check && cargo check -p trusty-mpm && cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-mpm --no-fail-fast                                                                     EXIT=0
```

46 test targets, zero failures — both known baseline flakes passed this run.

Nine repo gates, EXIT=0 each: `check_line_cap` `check_sld` `check_doc_numbers` `check_test_pointers` (21965 citations, 0 dangling) `check_capabilities` `check_changelog_fragment` `check_generation_artifacts` `check_agent_assets` `check_buildrs_sync`.

bobmatnyc:
## Verdict: APPROVE

Re-review of head `7b3842745`, independent worktree, read-only. The prior BLOCK ([comment](https://github.com/bobmatnyc/trusty-tools/pull/4995#issuecomment-5200656322)) is **resolved** — not papered over. Two findings remain (MEDIUM + LOW); neither can hide a live session.

### The BLOCK is gone

The CRITICAL was: the filter keyed on the wire's `unresumable` flag, which `reconcile_live_state` lets ship alongside `state: "active"`, so a live-but-unreapable session could be hidden. That flag is now out of the visibility path entirely.

`filter_live_sessions` (`crates/trusty-mpm/src/bin/tm/commands/managed.rs:175`) consults `dead_ids.contains(&s.id)`. `dead_ids` is built at `session_picker_prune.rs:707` from the `is_dead_record` partition the sweep itself acts on — `is_clearable_state && workspace_verified_gone`. Both proved cases stay visible by construction:

- `state: "active"` + `unresumable: true` → `is_clearable_state` rejects on the `stopped|errored` gate (`session_picker_prune.rs:558`) → never enters `dead` → never in `dead_ids` → **visible**.
- `state: "errored"` with a live detached pane → rejected by the `live_tmux_names` guard (`session_picker_prune.rs:561`, PR #4725 round 2) → **visible**.

Verified by mutation, not by reading. Reverting line 175 to `s.unresumable`:

```
failures:
    tests_behavior_e::picker_filter_hides_only_what_the_prune_classified_dead
    tests_behavior_e::picker_filter_keeps_a_live_row_flagged_unresumable
    tests_behavior_e::scope_for_display_all_keeps_dead_record_visible
test result: FAILED. 1505 passed; 3 failed; 1 ignored
```

### The invariant, checked in both directions

- **Hidden but unreapable?** Every id in `dead_ids` passed `is_dead_record`, so it is exactly what the sweep is willing to reap. No path adds an id from any other source.
- **Reaped but visible?** A reaped row leaves `kept` and is removed from `dead_ids` in the same arm (`session_picker_prune.rs:783-787`). `dead_ids ⊆ kept` holds, so there is no kept-and-reapable row outside `dead_ids`.
- **Slot tombstones**: doubly protected. `is_clearable_state` rejects `s.deleted` before classification, and the `"deleted" => is_slot_tombstone` arm sits above the `classified_dead` arm in `is_live_session_state`. Pinned by `is_live_session_state_keeps_slot_tombstone_even_when_classified_dead`.
- **Resumable stopped rows**: `is_live_session_state("stopped", false, false) == true`, and `dead_ids` only ever contains rows whose workspace `absence_corroborated` confirmed gone. The 18 resumable rows are untouched.

### Fail-closed contract

The early return exists at `session_picker_prune.rs:681-691` and hands back an empty `dead_ids`, so nothing is hidden. It is reached on every failure mode `live_tmux_session_names` produces: `run_tmux` error (`.ok()?`, line 588) and non-zero exit (lines 590-592). tmux present with zero sessions returns `Some(empty)`, which is correct — no panes means no live panes.

An unmounted volume does not empty `tm ls`. `try_exists` answers `Ok(false)` there, but `absence_corroborated` (`session_picker_prune.rs:496-518`) then declines: a path under a worktree container whose container is also gone returns `false`, and a path with no container ancestor falls back to the immediate parent, which is equally gone on an unmounted volume. Classification never happens, so nothing is hidden.

### The vacuity fix is genuine

`default_view_pre_4994` (`tests_behavior_d_tests.rs:1852`) scopes with an empty dead set, so the only thing that can hide the record is a real `decommissioned` flip. Reproduced the author's mutation — `decommission_dead_record` returning `Pruned` without contacting the daemon:

```
failures:
    tests_behavior_d::auto_prune_always_requests_record_only_never_full_teardown
    tests_behavior_d::auto_prune_clears_record_whose_worktree_root_survived_the_removal
    tests_behavior_d::auto_prune_dead_records_removes_confirmed_unresumable_records
    tests_behavior_d::auto_prune_dead_records_stale_daemon_sentinel_expires_after_ttl
    tests_behavior_d::auto_prune_dead_records_stops_sweep_when_daemon_reports_workspace_removed
    tests_behavior_d::session_ls_json_never_refetches_after_pruning
    tests_behavior_d::session_ls_json_passthrough_prunes_dead_records
    tests_behavior_d::session_ls_prunes_dead_records_on_piped_invocation
test result: FAILED. 1500 passed; 8 failed; 1 ignored
```

All three teardown assertions are in that list. They now require the daemon round-trip.

### Coverage was not weakened

`git diff origin/main...HEAD` over the three touched test files shows additions plus one rename (`parse_scoped_sessions_all_keeps_tombstone_in_slot_order` → `scope_for_display_all_keeps_tombstone_in_slot_order`). Nothing deleted. No `#[ignore]`, no `cfg`-gate, no `--exclude`, no `--lib` narrowing anywhere in the diff.

### Banner scoping (the prior MEDIUM)

`session_ls_at` passes `false` on the `--json` branch (`managed.rs:276-280`) and `!all` on the table branch (`managed.rs:285`); `fetch_live_sessions` passes `!all` (`session_picker.rs:488`). The "(hidden; `tm ls --all` to see them)" suffix now prints only where rows were actually dropped, and `--json` output is unfiltered — it echoes `raw` verbatim at `managed.rs:282`. Resolved.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker_prune.rs` | 790, 817 | A record whose decommission returns `Failed` stays in `kept` AND in `dead_ids` — so it is hidden — but is counted in neither `pruned` nor `pending`. `pending = confirmed.len() + first_sighting.len() + stale_daemon_held` omits it. If every confirmed record's decommission fails, `pending == 0`, the `if outcome.pending > 0` guard skips the banner, and rows vanish with no "(hidden; `tm ls --all` to see them)" line. The `dead_ids` doc at lines 195-196 — "makes the hidden set exactly what `pending` counts" — is false whenever a decommission fails. | Track `let mut failed = 0usize;`, increment it in the `DecommissionOutcome::Failed` arm, and fold it into `pending` at line 817 exactly as `stale_daemon_held` already is. The doc claim then holds. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/tests_behavior_e_tests.rs` | 533 | The comment cites `critic_proof_4995_live_pane_can_ship_active_plus_unresumable` as "proven by". That test does not exist anywhere in the repo — it was a throwaway probe from the prior review round. It escapes `scripts/check_test_pointers.sh` only because it sits in a `//` comment and is line-wrapped mid-identifier. | Cite the mechanism instead (`daemon::managed_routes::summary::reconcile_live_state` overwrites display state after `checked_summaries` computes `unresumable`), or drop the clause. | Fix here |

## Notes

**Hiding fires on the first sighting; reaping still needs the 10-minute window.** `dead_ids` is captured at line 707, before `classify_sighting` runs, so a newly-classified row disappears from the default view on the very first `tm ls` rather than after the two-sightings confirmation. Not flagged: the confirmation window exists to gate a *destructive* tombstone, and hiding is reversible via `--all`. A row that reaches `dead_ids` has already cleared `stopped|errored`, not-attached, not-a-tombstone, no live tmux pane, non-terminal persisted state, and corroborated workspace absence. The strict reading is `hidden ⟺ classified dead`, a superset of `reapable this call` — but every member of that superset is a row the sweep reaps on a later call.

**Flicker is possible and points the safe way.** If tmux enumeration fails on one invocation, the fail-closed return empties `dead_ids` and every dead row reappears; they vanish again next call. Rows flicker toward *visible*, never toward hidden.

**The `unresumable` flag is not entirely gone from decision code, and should not be.** Remaining reads outside rendering are resume-blocking guards, all pre-existing and all in the keep-safe direction: `session_picker.rs:620` (`Unresumable(idx)`), `session_picker.rs:900` (bare-Enter warning), `tui/project_ctl/events/mod.rs:318` (`r` blocked). Only the visibility filter needed to stop reading it, and it has.

## Gates run (independent worktree at `7b3842745`)

```
cargo fmt --check && cargo clippy -p trusty-mpm --all-targets -- -D warnings    EXIT=0
cargo test -p trusty-mpm     test result: ok. 4661 passed; 0 failed; 7 ignored
                             test result: ok. 1508 passed; 0 failed; 1 ignored   (x2)
                             + 43 further targets, all ok; 0 failed overall      EXIT=0
bash scripts/check_line_cap.sh                                                  EXIT=0
bash scripts/check_test_pointers.sh    21965 citations, 0 dangling               EXIT=0
```

Both known flakes passed: `stale_assets_for_many_*` (4/4 ok) and `ensure_managed_config_dir_emits_the_frozen_skill_warning` (ok). `meta_run_demo_writes_and_verifies_artifact` stayed `ignored`, as on main.

Zero CRITICAL, zero HIGH → **APPROVE**. The two findings are worth landing but do not gate the merge.

bobmatnyc:
Obligation carried over from the [#5048](https://github.com/bobmatnyc/trusty-tools/pull/5048) review, to be discharged by the piece that flips the default.

The startup sweep landed in #5048 but has **never enumerated a non-zero count against a real data root** — `spawn_startup_backfill` returns immediately unless `TRUSTY_BM25_DAEMON=1`, which no shipped path sets. Every proof behind it is a test fixture. That is declared rather than hidden, but it is not evidence the sweep works.

Before anyone calls the lexical lane working, this piece owes a real log line showing a non-zero `enumerated` against the ~99-palace root, with the `swept` / `incomplete` / `unopenable` breakdown. Not a test count.

One cost question already settled, so it does not need re-litigating: the sweep opens palaces serially, but `load_palaces_from_disk()` hydrates every palace immediately before the sweep spawns, so those opens hit a warm registry.

Issue #4595, opened 2026-08-01
PM instructions: restructure around risk tiers and intent routing, target 2-3k words


bobmatnyc:
Blocked by #4594 — the direct-work policy decision there changes the shape of this restructure. Do not start until that's ruled.

## Context

Always-loaded instruction set is ~6,160 words / 42KB. Startup context is ~90k tokens before the user types (#4513). Two independent reviews — an internal audit (#4574) and an external one (#4594) — converged on the same conclusion: the dominant problem is internal contradiction, not wording.

**The owner has authorized changing the instruction package schema** to accommodate this. The restructure is not limited to prose edits within the existing section shape — `instruction-package.schema.json` and `pm-instruction-package.json` may both change.

## Proposed changes

- [ ] Replace phase-by-phase mandatory/skip exceptions with RISK TIERS (routine / standard / high-risk) selecting the required path. Aligns with the existing sprint-then-harden doctrine, which is already a risk-tier model.
- [ ] Replace KEYWORD routing ("mentions test -> QA") with INTENT routing (classify the requested outcome). Keyword routing over-delegates: "explain this test" is not a QA task. Current keyword rules: `sections/agent-delegation.md:51-63`.
- [ ] Move rare procedures out of the always-loaded prompt into on-demand skills, leaving a one-line trigger: parked-subagent recovery (`sections/core.md:146-183`), the GitHub label shell example (`:313-340`), skill deployment internals (`:389-427`), Rust-specific CI waiting guidance (`:172-178`), release implementation detail (`sections/workflow.md:157-163`).
- [ ] Apply the four-section response template only to task-completion reports, not to every response (`sections/core.md:449-455`).
- [ ] Replace the "90% success probability" escalation threshold with observable conditions (`sections/core.md:225-229`) — an uncalibrated probability estimate is not actionable.
- [ ] Replace the lexical forbidden-phrases list with ONE semantic rule: do not claim completion, deployment, or a fix without naming the verification performed and its result; label unverified hypotheses as hypotheses.
- [ ] Target: 2,000-3,000 words always-loaded.

## Non-goal (explicit — do not "optimize" this away later)

The external review recommended dropping the default requirement for RAW command output in favor of summarized pass/fail counts. **Rejected.**

[...]

Issue #2647, opened 2026-07-14
session provisioning: new PM session starts at ~100k tokens — de-duplicate CLAUDE.md and PM-instruction injection (regression of #2299/#2300)


bobmatnyc:
## Summary

A brand-new tm-managed PM session on `bobmatnyc/trusty-tools` starts with
**~100k tokens (~10% of a 1M context) already consumed** before any work
happens, driven by trusty-mpm's own session-provisioning: duplicated
`CLAUDE.md` content, double-injected PM/delegation instructions, and large
static rosters (MCP tools, agents, skills) loaded unconditionally.

Critically, **part of this is a regression of already-shipped fixes**
(#2299/#2300, and separately #2170) that never reached this session because
trusty-mpm never re-syncs a long-lived session worktree against
`origin/main`. See root-cause section below.

## Measurement table (this session, `.base/.worktrees/2eb72dca-de08-481b-8dfa-22ab7f81b1f9`)

| Component | Bytes | Est. tokens (÷4) | % of ~100k |
|---|---|---|---|
| `/Users/masa/CLAUDE.md` (user-level) | 378 | ~95 | 0.1% |
| Project root `CLAUDE.md` (**stale — deleted on `origin/main` by #2300**) | 27,066 | ~6,767 | 6.8% |
| Worktree `CLAUDE.md` (root copy + legacy #2125 delegation banner) | 27,452 | ~6,863 | 6.9% |
| — of which pure duplication of the root copy | ~27,066 | ~6,767 | ~6.8% |
| Output-style `trusty-mpm.md` (loaded via Claude Code `outputStyle`) | 10,677 | ~2,669 | 2.7% |
| `.trusty-mpm/last-instructions.md` (`--append-system-prompt-file`, assembled from `PM_INSTRUCTIONS.md`+`WORKFLOW.md`+`AGENT_DELEGATION.md`+`BASE_PM.md`) | 28,281 | ~7,070 | 7.1% |
| Deferred MCP tool name list (~106 tools, names only) | 3,361 | ~840 | 0.8% |
| Memory index `MEMORY.md` | 2,548 | ~637 | 0.6% |
| **Measured subtotal** | **99,763** | **~24,941** | **~25%** |
| Agent roster (~40 agents w/ descriptions) — order-of-magnitude, visible in this session's tool-list reminder | not separately captured to disk | ~900–1,200 | ~1% |
[...]

Issue #2214, opened 2026-07-07
feat(trusty-mpm): seed outputStyle/statusLine into tm-owned CLAUDE_CONFIG_DIR settings.json (defense-in-depth)


bobmatnyc:
## Context

**DOC-34 / #1996** made managed Claude Code sessions launch under a tm-owned `CLAUDE_CONFIG_DIR` at `~/.trusty-tools/trusty-mpm/claude-config/`.

## Current State (Verified, NOT a Bug)

The effective `outputStyle`/`statusLine` for managed sessions come from the **PROJECT tier** (`<workspace>/.claude/settings.json`, written by `write_output_style`/`write_status_line` in `crates/trusty-mpm/src/core/session_launch/settings.rs:200,759`). 

The daemon spawns `claude` with `--setting-sources project,local`, which excludes the user tier that CLAUDE_CONFIG_DIR relocates. So the global `~/.claude/settings.json` (managed by the unrelated Python claude-mpm tool) is never consulted — isolation is already complete.

## The Gap (Defense-in-Depth Only)

The tm-owned `claude-config/settings.json` carries **NO** `outputStyle`/`statusLine` keys at all, so the protected dir is not a self-sufficient settings file. 

If `--setting-sources` is ever dropped or a standalone `tm run` path is added where the config-dir tier becomes load-bearing (see the WHICH-LAYER caveat at `crates/trusty-mpm/src/core/managed_config.rs:25-38`), the tm-owned dir should already carry these keys.

## Proposed Fix

Seed `outputStyle` (resolved style id, default `trusty-mpm`) and `statusLine` (absolute `<tm-binary> statusline` command) into the initial settings.json write in `ensure_global_config_dir` (`crates/trusty-mpm/src/core/standalone/global_config.rs:54-74`), parallel to how `ensure_managed_hooks` merges hooks. 

Keep it **non-destructive**: only seed when absent / merge, don't clobber the client's self-persisted keys like theme/tui.

## Acceptance Criteria

- `make check` green
- A freshly-provisioned config dir contains both keys
- Existing project-tier behavior unchanged

PR #5002, opened 2026-08-06
refactor(trusty-mpm): relocate and deduplicate the always-loaded PM prompt (#4574)

### LINKED ISSUE #4574: PM instructions: ~12k tokens misplaced, two live contradictions, unenforceable prohibitions
## Measured shape

Composed PM prompt: 91,337 bytes / 1,779 lines / ~22,830 tokens. 43% of it is `.trusty-mpm/INSTRUCTIONS.md` (39,321 bytes / ~9,830 tokens) — Rust coding conventions given to an agent forbidden from writing Rust.

## Two contradictions hit in normal operation

**Model selection.** `core.md:88`: "EVERY Agent tool call MUST include `model: \"sonnet\"` or `model: \"haiku\"`. No exceptions." Nine lines later the routing table mandates `model: \"opus\"` for coding/engineering tasks. 400 bytes apart; whichever the agent follows violates the other. Hit on every engineering delegation.

**`git push`.** PM allowlist lists `git push/branch` (`:171`); prohibition P7 (`non-overridable-rules.md`) lists `push/branch` as a circuit-breaker violation. Hit whenever the PM finishes a commit.

## Prohibitions stated but not enforced

`evaluate_bash_command` (`bin/tm/commands/pm_guard_bash/mod.rs:189-220`) denies: `patch`, non-readonly `sed`/`awk`, `curl`, `wget`, `make`, `pytest`, `npm test`, file-write redirection. Consequences:

- `cargo` is unguarded — the build command of a Rust-only workspace is listed in no prohibition table.
- P9 (`rm`/`rmdir`) is unenforced — the prohibition with irreversible blast radius.
- P2 ("read >3 files") is undetectable by the model; nothing counts reads.
- The 3-strike circuit-breaker state machine (~370 tokens) has no state anywhere in the prompt or returned by any tool. The model cannot know it is on strike 2.

## Dead rules and misplacement

Placement: "never turn red green by deleting coverage" sits at 32% depth, mid-document, in an unmarked blockquote. "The main checkout is inspection-only" sits at 83% depth inside the addendum.

Duplication: attribution footer appears verbatim 3× in the composed prompt (golden `testdata/pm-prompt-bundled-fallback.md` lines 307, 674, 895), plus a 4th in `CLAUDE_MD_STUB`. The architecture makes single-sourcing impossible (one section per block, no `$ref`/`include`).

Dead rules: `/mcp-configure` and `/mcp-init` (no such commands; live skill is `tm-init`), `.claude-mpm/` references (legacy dir; pipeline reads `.trusty-mpm/`), "deployed to `~/.claude/agents/`" recovery (false per `core/managed_config.rs:29`), a pytest example in a Rust repo, closed-ticket archaeology (#170/#171/#172), "Domain Consolidation Audit" status board from 2026-07-11.

## Two unaddressed failure modes

Nothing prohibits merging red or pending CI. Zero occurrences of "admin", "red CI", or "branch protection". That rule exists in the SUBAGENT base prompt but not the PM's — and the PM authorizes the merges.

Nothing tells the PM to verify an agent's completion claim against state the agent did not author (commit exists, PR exists, checks green, file on disk). Observed failure: merge agent reported "VERIFIED WORKING" having checked `port.rs` when asked to verify `port_guard.rs`.

## Estimated saving ~12,100 tokens (22,830 → ~10,700, a 53% cut)

- Relocate Rust-authoring conventions to rust-engineer/local-ops (~7,300 tokens)
- Workflow.md duplicate half (~1,200)
- Agent-delegation Make/Mise + duplicate routing table (~800)
- Roster filtered to detected stack (~700)
- Core: Skill Deployment / Auto-Config / Response Format / Doc Routing / gh-label block / duplicate Identity (~1,400)
- Dead rules (~500)
- Footer 4× → 1× (~220)

Reference #4513 (startup token measurement) and #4286 (override mechanism).

bobmatnyc:
Closes [#4574](https://github.com/bobmatnyc/trusty-tools/issues/4574).

## What this is

The composed PM prompt drops **6,611 bytes — 56,783 → 50,172, −11.6%** (8,462 → 7,502 words, roughly 1,650 tokens). Every removed block either went to a destination that is not session-resident, or was a second copy of text already resident through another channel.

**Test ladder rung 3–4.** Localized to `trusty-mpm`'s bundled assets, but the composed prompt is a cross-cutting surface and `BASE-AGENT.md` is shared with `trusty-code`, so `cargo test -p trusty-code` ran too.

## Before / after

| Artifact | Before | After | Δ |
|---|---|---|---|
| `pm-prompt-bundled-fallback.md` (composed prompt) | 56,783 B / 8,462 w | 50,172 B / 7,502 w | **−6,611 B (−11.6%)** |
| `pm-prompt-roster-absent.md` | 56,589 B | 49,978 B | −6,611 B |
| `pm-prompt-claude-md-override.md` | 42,560 B | 36,695 B | −5,865 B |
| `output-styles/trusty-mpm.md` | 14,751 B | 15,530 B | +779 B (two new rules + note) |
| **Session-resident total** (prompt + style) | **71,534 B** | **65,702 B** | **−5,832 B** |

Goldens regenerated with `UPDATE_GOLDEN=1`; the diff is the acceptance artifact. Exactly four headings left the prompt and one arrived — nothing vanished silently:

```
-## Skills System            -## Skill Deployment
-## Agent Deployment         -## Auto-Configuration
+## Skills and Agents — What You Need at Dispatch Time
```

## What moved where, and why the destination is not per-session-resident

| Block | From | To | Why that destination is not resident |
|---|---|---|---|
| `Skill Deployment` + `Agent Deployment` tier tables | `sections/core.md` | `tm-capabilities` → `references/framework.md` (already there) | A skill's body loads only on `Skill(...)`; the harness injects name + one-line description. The tables were **already** generated from `core::paths::FrameworkPaths` / `deployed_agent_dirs_from` and drift-gated by `scripts/check_capabilities.sh`, and the prompt already named the file — this deletes a stale prose copy, it does not create one |
| Deployment lifecycle (redeploy freeze, orphaned copies, tier-collision logging, `mcp`-slug exclusion, managed-tier reach) | `sections/core.md` | `tm-capabilities` → `references/workflows.md` §5 (new) | Same skill. `workflows.md` is the hand-authored member of that skill and is explicitly excluded from the drift gate, so it is a legal write target |
| `gh label create` / `gh issue create` / `gh pr create` shell block | `sections/core.md` | `tm-ticketing` L73–81 and `tm-pr-workflow` L178–192 (already there) | Skill bodies, loaded by `ticketing` / `version-control` on dispatch. P6/P7 forbid the PM running these; the paragraph directly below the block said so |
| Fail-Open Check — the five review steps | `sections/workflow.md` | `code-review-standards` (new §) | Skill body loaded by `code-critic` (already declared) and `code-analyzer` (declared in this PR). The audience is a reviewer, not the PM |
| `Prose Style — Write Plainly` body | `sections/core.md` | `output-styles/trusty-mpm{,-teacher,-research}.md` (already there) | **Deduplication, not relocation** — see below |
| `Security Check Template` fence | `sections/workflow.md` | folded into the 3-step rule above it | It restated the three steps verbatim |
| Documentation Routing one-row table | `sections/core.md` | one sentence | Table scaffolding for a single row |

Deleted outright as dead rules: `/mpm-configure --preview` and `/mpm-init` name no command that exists (`tm-init` is the live skill). Corrected: `Delegation Mechanics` claimed bundled agents deploy to `~/.claude/agents/` — they deploy to `$CLAUDE_CONFIG_DIR/agents/`, and tm never writes framework agents into the operator's own install (`core/managed_config.rs`, `references/framework.md`).

## The prose-rule consolidation

`sections/core.md` and the three bundled output styles carried the same voice rules. The output style said so out loud: *"Both copies are live — change one, change the other."* Both channels are session-resident, so every session paid for the rules twice.

**The output style wins.** It is the only channel that survives a manual `claude` launch with no tm-appended system prompt ([#2647](https://github.com/bobmatnyc/trusty-tools/issues/2647)) — the same reason the PRIMARY DIRECTIVE lives there. `core.md` keeps the heading and a pointer, so the PM still knows the rules bind.

**Drift found, and how it was reconciled.** Diffing the two copies:

| Only in the output style | Ruling |
|---|---|
| `**Tone**: professional, neutral` / `**No mocks**` / `**No placeholders** — never todo!() or stubs` | **Kept.** The last two are engineering rules, not voice rules; they belong in the style's self-contained floor for a manual launch |
| Extra praise example: `"Excellent!", "Perfect!", "Amazing!", "You're absolutely right!"` | **Kept.** Strictly more coverage of the same rule |

Nothing was in `core.md` that the output style lacked, so no wording was lost. Remaining differences were line wrapping only.

**Not moved, deliberately:** `Clickable References` and `Banned Word — "honest"` stay as package `text` blocks in the composed prompt. They sit in section `core`, whose `customization_tier` is `"fixed"` and whose `permits()` returns false — so they are *not* project-overridable, and an earlier draft of this body said otherwise. They stay because they are prompt-authoring rules with no mirror in the output style, not because of any overridability property.

## PM-only vs agent-facing

| Rule set | Home | Reason |
|---|---|---|
| Full voice rules — hedges, openers, aphorisms, praise, framing openers, embellishment, ticket/PR body shape | Output styles (all three) | PM-facing. Governs what the user reads |
| `## Agent-Authored Prose` — the same standard, condensed, plus verbosity-scales-with-findings and the raw-output carve-out | `BASE-AGENT.md` (both crates) | Agent-facing. A subagent receives neither the PM prompt nor its output style, so this is not a duplicate — it is the only copy that reaches an agent. Untouched except for the two new rules and a repointed reference |
| Fail-Open Check five steps | `code-review-standards` | Reviewer-facing |
| Deployment tiers / lifecycle | `tm-capabilities` | Operator-facing lookup |

## Two new rules (owner, 2026-08-06)

Added to all three output styles and to `BASE-AGENT.md` in both crates, next to the closing-aphorism rule they share a shape with:

- **Don't justify the restraint.** *"I don't know yet"* is the whole answer — the trailing *"I'm not going to guess at a number this specific"* explains why you are declining, which is process narration wearing a caveat's costume. Same for *"rather than guess"*, *"I won't speculate"*.
- **No trailing emphatic negation.** *"The effect is real once the binary is installed — not before"* restates the sentence by negating its opposite. Same shape as *"…, not the other way around"* appended to a sentence that already said it.

## InstallPolicy / auto-set: confirmed non-defect, nothing changed

Both halves of the owner's ruling already hold. Evidence:

**1. Auto-installed — framework-owned, overwrite.** `crates/trusty-mpm/src/core/output_style_deployer.rs:29-30`: *"There is no 'skipped' category — output styles are always framework-owned, so the deployer unconditionally overwrites stale copies."* `:74-99` reads the on-disk bytes, compares to the bundled content, and atomically rewrites on any difference. Called from `core/standalone/global_config.rs:198` on **every** managed-driver bootstrap. Output styles are not in the `bundle_all.rs` `ALL` table at all, so the `SeedOnce` variant never applies to them.

**Consequence: an edit to the bundled asset reaches an existing deployed copy on the next `tm` session launch.** No `tm reinstall` required. This is not the [#4676](https://github.com/bobmatnyc/trusty-tools/issues/4676) failure mode.

**2. Auto-set — two independent writers, neither human.** Project tier: `core/session_launch/settings.rs` `write_output_style` sets `"outputStyle": "<active>"` in the project `settings.json`. Config-dir tier (defense-in-depth, [#2214](https://github.com/bobmatnyc/trusty-tools/issues/2214)): `core/standalone/settings_defaults.rs:78-80` seeds the same key into the tm-owned `CLAUDE_CONFIG_DIR/settings.json`, called from `ensure_global_config_dir`. The seed is `or_insert_with`, so a value a human deliberately set is preserved — correct for a setting, and the reason the live config's `"outputStyle": "trusty-mpm"` is what tm would have written anyway.

Nothing to fix, so nothing was changed. Per the brief's third branch, this did not become a lifecycle change bolted onto an instruction-slimming PR.

**One more reason the dedup is safe either way:** when Claude Code lacks native output-style support, `core/instruction_pipeline.rs:402` injects the style into the composed prompt instead. On that path the two copies were adjacent in one document; deleting one is still exactly one copy, never zero. All three bundled styles carry the identical block, so switching to `teacher` or `research` does not lose the rules.

## Gates

> **Correction.** An earlier revision of this body claimed `cargo fmt --check EXIT=0`. That run was real but **stale**: it happened before the final edit to `claude_md_sections_tests.rs`, which added a 101-char assertion, and fmt was not re-run after it. CI was right and the body was wrong. Fixed in `bc92fbb`; the block below is the current head.

```
$ cargo fmt --check
EXIT=0

$ cargo check -p trusty-mpm --all-targets && cargo clippy -p trusty-mpm --all-targets -- -D warnings
EXIT=0

$ cargo test -p trusty-mpm
test result: ok. 4695 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 96.28s
test result: ok. 1504 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 9.67s
test result: ok. 1504 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 10.12s
(+ 37 further binaries, all ok, 0 failed)
EXIT=0

$ cargo test -p trusty-code
(all suites ok, 0 failed)

line-cap: measured 3739 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

$ bash scripts/check_sld.sh
sld-lint: scanned 56 spec doc(s) + 3113 code file(s); 0 error(s), 0 warning(s)

tm-capabilities: up to date (7 generated files).

OK   trusty-code: changelog.d fragment present and valid
changelog-fragment gate: scanned 17 changed path(s); all 2 crate(s) with source changes are recorded.
```

### Three tests failed on the first run; all three are accounted for

1. `core::claude_md_sections::tests::the_prose_rules_ban_categories_not_phrase_lists` — asserted the prose rules live in `core`. **Retargeted, not deleted:** it now asserts every bundled output style carries them (including the two new rules), and a new sibling `the_prose_rules_live_in_the_output_style_not_core` asserts `core` keeps the heading, names the destination, and restates none of the rules. That second test is what stops the duplicate coming back.
2. `core::instruction_pipeline::tests::primary_directive_mandate_not_duplicated_across_channels` — my rewritten mirror note put `PRIMARY DIRECTIVE` on one line where the old wording had a line break inside it, making the sentinel count 2. **Real catch.** Reworded to "the mandate banner above"; count back to 1.
3. `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — failed with `Captured lines: []`, a stderr-capture race under parallelism. Passes in isolation, and passed on the full re-run. `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/core/managed_config.rs crates/trusty-mpm/src/core/managed_config_tests.rs` is **empty** — this branch touches nothing in that path. Pre-existing isolation flake, not caused here and not suppressed.

## tcode embedded agent-asset sync

`code-analyzer.md` is one of four files that deliberately deviate between the two crates and are therefore pinned by trusty-mpm **source** hash in `scripts/agent-asset-pins.tsv` rather than byte-compared. Adding `skills: [code-review-standards]` upstream moved the source behind its trusty-code copy, and `scripts/check_agent_assets.sh` exited 1 — the gate working as designed.

**Direction synced: `crates/trusty-mpm/` → `crates/trusty-code/`.** trusty-mpm is the source; the trusty-code copy is derived and adds a `tools:` restriction. `code-critic.md` sets the precedent — its trusty-code copy already carries `tools:` *and* the upstream `skills:` line, so the deviation is additive. The trusty-code `code-analyzer` copy keeps its `tools:` restriction and gains the same `skills:` line; that list already includes `use_skill` and `crates/trusty-code/src/assets/skills/code-review-standards` exists, so it resolves. Then `scripts/check_agent_assets.sh --update` re-pinned the one changed hash — reconciled first, never a blind `--update`.

Noted and deliberately left: before this PR the trusty-code `code-analyzer` copy declared no skills at all while its trusty-mpm source did, an asymmetry that predates this change. The owner has ruled trusty-code out of scope, so this PR does the minimum that makes the gate green and nothing more. Recording it here so it is not lost. `BASE-AGENT.md` is in the 30-file byte-parity set and was edited identically in both crates, which the passing gate confirms.

## Relation to [#4595](https://github.com/bobmatnyc/trusty-tools/issues/4595)

Deliberately out of scope: the risk-tier rewrite, intent routing, and the phase-exception collapse. This PR touches no phase table, no routing table, and no schema, so #4595 rebases cleanly on top. It does partly satisfy #4595's "move rare procedures into on-demand skills" checkbox, which makes that issue smaller rather than harder.

**Worth flagging:** #4574's headline estimate was ~12,100 tokens, ~7,300 of it from relocating Rust-authoring conventions out of `.trusty-mpm/INSTRUCTIONS.md`. That file was retired by [#4286](https://github.com/bobmatnyc/trusty-tools/issues/4286) and the saving is already banked — the 56,783-byte baseline here is post-retirement. Several other items on the issue's list (the footer 4× → 1×, the duplicate Identity block, the workflow QA/commit duplication) had also already landed. What remained is what this PR takes.

### DIFF
--- a/crates/trusty-code/changelog.d/4574-base-agent-prose-rules.md
+++ b/crates/trusty-code/changelog.d/4574-base-agent-prose-rules.md
@@ -1,0 +1,6 @@
+Changed
+
+- `BASE-AGENT.md` gains two prose rules and repoints its voice standard at the output style rather than the retired `sections/core.md` block ([#4574](https://github.com/bobmatnyc/trusty-tools/issues/4574))
+  - don't justify the restraint — "I don't know yet" is the whole answer; the trailing "I'm not going to guess" is process narration in a caveat's costume
+  - no trailing emphatic negation — "— not before" restates the sentence by negating its opposite
+  - kept byte-identical to `crates/trusty-mpm/src/assets/agents/BASE-AGENT.md`, as it has always been
--- a/crates/trusty-code/src/assets/agents/BASE-AGENT.md
+++ b/crates/trusty-code/src/assets/agents/BASE-AGENT.md
@@ -313,9 +313,11 @@ invocation — `$PIPESTATUS` is a bashism and this harness runs zsh.
 
 ## Agent-Authored Prose
 
-The PM's prose standard ("Prose Style — Write Plainly" in `sections/core.md`)
-governs everything you write back: the report to your dispatcher, a review
-verdict, ticket and PR body text, and any generated documentation.
+The PM's prose standard ("Communication — Write Plainly", in the active output
+style) governs everything you write back: the report to your dispatcher, a
+review verdict, ticket and PR body text, and any generated documentation. This
+is that standard restated for an agent, which receives neither the PM's prompt
+nor its output style.
 
 - Lead with the concrete referent, not its category — name the file, the
   function, the finding. Let the reader infer the category.
@@ -327,6 +329,15 @@ verdict, ticket and PR body text, and any generated documentation.
 - Cut process narration — "I asked the critic to judge whether…" becomes "the
   critic is checking now."
 - End options as a bare enumeration: "Two options: A, or B."
+- Don't justify the restraint. "I don't know yet" is the whole answer — the
+  trailing "I'm not going to guess at a number this specific" explains why you
+  are declining, which is process narration wearing a caveat's costume. Same
+  for "rather than guess", "I won't speculate". Delete the tail.
+- No trailing emphatic negation. "The effect is real once the binary is
+  installed — not before" restates the sentence by negating its opposite. It
+  adds no fact and underlines a point that already landed. Same shape as
+  "…, not the other way around" or "…, never X" appended to a sentence that
+  already said it.
 - Never announce the register you're writing in. No heading or preamble labelling
   the writing plain, honest, direct, candid, blunt, or unvarnished — the label
   implies the alternative was on the table. Wrong: "What remains unknown, stated
--- a/crates/trusty-code/src/assets/agents/code-analyzer.md
+++ b/crates/trusty-code/src/assets/agents/code-analyzer.md
@@ -5,6 +5,7 @@ description: Code analysis specialist. Reviews code for correctness, quality, se
 model: sonnet
 extends: base-research
 tools: [read_file, grep, glob, list_dir, search_code, use_skill, finish_task]
+skills: [code-review-standards]
 ---
 
 # Code Analyzer Agent
--- a/crates/trusty-mpm/changelog.d/4574-slim-pm-instructions.md
+++ b/crates/trusty-mpm/changelog.d/4574-slim-pm-instructions.md
@@ -1,0 +1,10 @@
+Changed
+
+- the composed PM prompt drops 6,611 bytes (56,783 → 50,172, −11.6%) by relocating content the PM cannot act on and deduplicating the voice rules ([#4574](https://github.com/bobmatnyc/trusty-tools/issues/4574))
+  - `Skill Deployment` / `Agent Deployment` / `Skills System` collapse to one dispatch-time block; the tier tables were already generated and drift-gated in `tm-capabilities` (`references/framework.md`), and the deployment lifecycle they also carried moves to that skill's hand-authored `references/workflows.md`
+  - the `gh label create` / `gh issue create` shell block leaves the prompt — `tm-ticketing` and `tm-pr-workflow` already carry it verbatim, and P6/P7 forbid the PM running it
+  - the Fail-Open Check's five review steps move to the `code-review-standards` skill, which `code-critic` already loaded and `code-analyzer` now declares; the BLOCKING rule and its error-arm-test requirement stay in the prompt
+  - `Prose Style — Write Plainly` becomes a pointer: the voice rules are stated once, in the output style, which was carrying a live mirror of the same text. Both copies were session-resident, so the project paid for them twice
+  - dead rules dropped: `/mpm-configure --preview` and `/mpm-init` name no command that exists
+  - `Delegation Mechanics` said bundled agents deploy to `~/.claude/agents/`; they deploy to `$CLAUDE_CONFIG_DIR/agents/`, which tm never writes into the operator's own install
+- the PM voice rules gain two entries, in the output styles and in `BASE-AGENT.md`: don't justify the restraint ("I don't know yet" is the whole answer), and no trailing emphatic negation ("— not before" restates the sentence by negating its opposite)
--- a/crates/trusty-mpm/src/assets/agents/BASE-AGENT.md
+++ b/crates/trusty-mpm/src/assets/agents/BASE-AGENT.md
@@ -313,9 +313,11 @@ invocation — `$PIPESTATUS` is a bashism and this harness runs zsh.
 
 ## Agent-Authored Prose
 
-The PM's prose standard ("Prose Style — Write Plainly" in `sections/core.md`)
-governs everything you write back: the report to your dispatcher, a review
-verdict, ticket and PR body text, and any generated documentation.
+The PM's prose standard ("Communication — Write Plainly", in the active output
+style) governs everything you write back: the report to your dispatcher, a
+review verdict, ticket and PR body text, and any generated documentation. This
+is that standard restated for an agent, which receives neither the PM's prompt
+nor its output style.
 
 - Lead with the concrete referent, not its category — name the file, the
   function, the finding. Let the reader infer the category.
@@ -327,6 +329,15 @@ verdict, ticket and PR body text, and any generated documentation.
 - Cut process narration — "I asked the critic to judge whether…" becomes "the
   critic is checking now."
 - End options as a bare enumeration: "Two options: A, or B."
+- Don't justify the restraint. "I don't know yet" is the whole answer — the
+  trailing "I'm not going to guess at a number this specific" explains why you
+  are declining, which is process narration wearing a caveat's costume. Same
+  for "rather than guess", "I won't speculate". Delete the tail.
+- No trailing emphatic negation. "The effect is real once the binary is
+  installed — not before" restates the sentence by negating its opposite. It
+  adds no fact and underlines a point that already landed. Same shape as
+  "…, not the other way around" or "…, never X" appended to a sentence that
+  already said it.
 - Never announce the register you're writing in. No heading or preamble labelling
   the writing plain, honest, direct, candid, blunt, or unvarnished — the label
   implies the alternative was on the table. Wrong: "What remains unknown, stated
--- a/crates/trusty-mpm/src/assets/agents/code-analyzer.md
+++ b/crates/trusty-mpm/src/assets/agents/code-analyzer.md
@@ -4,6 +4,7 @@ role: code-analyzer
 description: Code analysis specialist. Reviews code for correctness, quality, security, and architectural health using static analysis.
 model: sonnet
 extends: base-research
+skills: [code-review-standards]
 ---
 
 # Code Analyzer Agent
--- a/crates/trusty-mpm/src/assets/instructions/sections/core.md
+++ b/crates/trusty-mpm/src/assets/instructions/sections/core.md
@@ -48,9 +48,9 @@ agents this project actually received.
 **Execution path = the native Agent/Task tool.** Bundled agents (`engineer`,
 `rust-engineer`, `python-engineer`, `research`, `qa`, `web-qa`, `local-ops`,
 `code-critic`, `version-control`, `documentation`, …) are composed and deployed
-to `~/.claude/agents/`. Run one by calling the **Agent tool** with the deployed
-name, e.g. `Agent(subagent_type="rust-engineer", model="opus", prompt=...)`.
-This is the ONLY way a subagent actually runs.
+to `$CLAUDE_CONFIG_DIR/agents/`. Run one by calling the **Agent tool** with the
+deployed name, e.g. `Agent(subagent_type="rust-engineer", model="opus",
+prompt=...)`. This is the ONLY way a subagent actually runs.
 
 **`mcp__trusty-mpm__agent_delegate` does NOT execute an agent.** It is an
 optional tracking + circuit-breaker gate: it records the delegation in the
@@ -59,8 +59,8 @@ the agent. Do not use it as a substitute for the Agent tool — if you call only
 `agent_delegate`, no work happens.
 
 **Recovery — "Agent type 'X' not found".** This means the composed agents are
-not deployed to `~/.claude/agents/` (a deployment gap, NOT a reason to switch to
-`agent_delegate`). Do NOT silently fall back to `general-purpose` — that loses
+not deployed where this session reads them (a deployment gap, NOT a reason to
+switch to `agent_delegate`). Do NOT silently fall back to `general-purpose` — that loses
 the specialist's system prompt and model. Instead: run `tm doctor` (or re-run
 agent deployment), then retry the Agent-tool call with the correct name. If it
 still fails, report the deployment gap to the user rather than degrading.
@@ -239,25 +239,16 @@ session's own tmux session name — resolve it with `tmux display-message -p
 no workstream name and applies `trusty-mpm` alone). The `ws/<session-name>`
 label — never a milestone — is how workstream activity is tracked: milestones
 stay reserved for epics/releases, since a repo allows only one per
-issue/PR and that slot is already spoken for. `tm` itself ensures the label
-exists at session launch (issue #3726); create it defensively anyway in case
-this session predates that launch step:
+issue/PR and that slot is already spoken for. `tm` itself ensures both labels
+exist at session launch (issue #3726); the delegate creates them defensively
+anyway in case this session predates that launch step.
 
-```bash
-gh label create trusty-mpm \
-  --description "Created/managed by a trusty-mpm session" --color 8250df \
-  2>/dev/null || true
-WS_NAME="$(tmux display-message -p '#{session_name}' 2>/dev/null || true)"
-[ -n "$WS_NAME" ] && gh label create "ws/$WS_NAME" \
-  --description "trusty-mpm workstream $WS_NAME" --color 5319E7 \
-  2>/dev/null || true
-gh issue create --assignee @me --label trusty-mpm ${WS_NAME:+--label "ws/$WS_NAME"} --title "…" --body "…"
-gh pr    create --assignee @me --label trusty-mpm ${WS_NAME:+--label "ws/$WS_NAME"} --title "…" --body "…"
-```
-
-The mechanical `gh` calls are delegated to `ticketing` (issues) or `version-control` (PRs) per P6/P7 and CB#6; the
-`--label trusty-mpm --label ws/<session-name> --assignee @me` default and the
-footer are part of that delegation prompt.
+The mechanical `gh` calls are delegated to `ticketing` (issues) or
+`version-control` (PRs) per P6/P7 and CB#6; the `--label trusty-mpm --label
+ws/<session-name> --assignee @me` default and the footer are part of that
+delegation prompt. The exact `gh label create` / `gh issue create` / `gh pr
+create` invocations live in the `tm-ticketing` and `tm-pr-workflow` skills that
+those two agents load — the PM never runs them.
 
 ## PR Workflow
 
@@ -281,11 +272,9 @@ is version control. No direct ticket tool access either way.
 
 ## Documentation Routing
 
-| Context | Route | Path |
-|---------|-------|------|
-| No ticket | Local file | `{docs_path}/{topic}-{date}.md` |
-
-Default `docs_path`: `docs/research/`. Configurable via `.trusty-mpm/config.toml` key `documentation.docs_path`.
+A report with no ticket goes to `{docs_path}/{topic}-{date}.md`. Default
+`docs_path` is `docs/research/`, configurable via `.trusty-mpm/config.toml` key
+`documentation.docs_path`.
 
 ## Worktree Isolation
 
@@ -314,7 +303,7 @@ Each artifact type has exactly one place it is customized:
 - **Prompt/instruction sections** — named-section marker blocks in the
   project's root `CLAUDE.md`. Nothing else.
 - **Skills** — the skill tier system: project `.claude/skills/` > user
-  `~/.trusty-mpm/skills/` > bundled (**Skill Deployment**, below). A
+  `~/.trusty-mpm/skills/` > bundled (**Skills and Agents**, below). A
   hand-edited deployed skill freezes against redeploy on purpose.
 
 Ad-hoc override channels are BANNED: the retired `.trusty-mpm/` files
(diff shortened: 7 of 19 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review at `5d7c9dcf`, from a dedicated worktree off the PR head. Two merge-blocking gate failures, both introduced by this diff and both reproduced locally. The substance of the relocation — which is what actually carries risk here — holds up under every check I ran.

## Findings

| HIGH | `crates/trusty-mpm/src/assets/agents/code-analyzer.md` | 7 | `skills: [code-review-standards]` is added without reconciling the deliberately-deviated `crates/trusty-code/src/assets/agents/code-analyzer.md` or re-pinning `scripts/agent-asset-pins.tsv`. `check_agent_assets.sh` is red. | Decide whether tcode's `code-analyzer` should declare the skill too — tcode already ships `code-review-standards` — then `scripts/check_agent_assets.sh --update` | Fix here |
| HIGH | `crates/trusty-mpm/src/core/claude_md_sections_tests.rs` | 1738 | The line this PR adds is 101 chars and violates rustfmt. CI **Format check** is red. The PR body reports `cargo fmt --check … EXIT=0`, which does not hold at this head. | `cargo fmt` | Fix here |

Both reproduce in a clean worktree:

```
$ bash scripts/check_agent_assets.sh
FAIL: UPSTREAM CHANGED BEHIND A DEVIATED COPY — crates/trusty-mpm/src/assets/agents/code-analyzer.md no longer
      matches its pin in scripts/agent-asset-pins.tsv
      (pinned=55d810bc… current=4396b3dd…).
agent-assets: FAILED — see FAIL lines above.
EXIT=1

$ cargo fmt --check
Diff in …/crates/trusty-mpm/src/core/claude_md_sections_tests.rs:1738:
-        core.contains("Communication —\nWrite Plainly") || core.contains("Communication — Write Plainly"),
+        core.contains("Communication —\nWrite Plainly")
+            || core.contains("Communication — Write Plainly"),
EXIT=1
```

The asset-pin failure is the one worth a second of thought rather than a reflex `--update`. The Fail-Open Check just left the always-loaded PM prompt and now lives only in `code-review-standards`; `workflow.md` line 126 says "`code-analyzer` and `code-critic` already load that skill". For trusty-mpm that becomes true with this diff. tcode's `code-analyzer` has no `skills:` key at all, and tcode ships that skill. tcode never consumed the mpm PM prompt, so nothing regressed there — but the gate is asking for a decision, and re-pinning without making it would bank an unexamined asymmetry.

## Required Changes

1. Run `cargo fmt`.
2. Reconcile `crates/trusty-code/src/assets/agents/code-analyzer.md` against the new upstream content on purpose, then `scripts/check_agent_assets.sh --update`.

## What I verified, and how

**Byte counts and the heading claim — confirmed exactly.** `pm-prompt-bundled-fallback.md` 56,783 → 50,172 B, −6,611. Four headings out (`Skills System`, `Skill Deployment`, `Agent Deployment`, `Auto-Configuration`), one in (`Skills and Agents — What You Need at Dispatch Time`). Nothing else moved.

**Did a rule die?** I diffed the composed golden line-by-line — 205 removed lines, not four headings — and traced every removed line to a destination. Nothing was lost. The ones worth naming:

- Skill/agent tier precedence, the hand-edit freeze, orphaned copies, the `mcp`-slug exclusion, the managed-tier reach → `references/framework.md:36–46` and the new `references/workflows.md` §5.
- `There is no ~/.trusty-mpm/agents/ tier … (#4946)` → `references/framework.md:34`.
- The `gh label create` / `gh issue create` / `gh pr create` block → `tm-ticketing.md:74–88` and `tm-pr-workflow.md:178–192`, both carrying the `--assignee @me --label trusty-mpm --label ws/<session-name>` default and the footer.
- Fail-Open Check five steps → `code-review-standards.md`, byte-identical.
- Documentation Routing: the table collapsed to a sentence that keeps both the `docs/research/` default and the `documentation.docs_path` config key.
- `**CRITICAL**: PM MUST DELEGATE all version bumps` survives, rewrapped and with `Cargo.toml` added.
- `/mpm-configure --preview` genuinely resolves to nothing — zero hits across `crates/trusty-mpm/src/`. Deleting it is correct.

**Prohibitions, Circuit Breakers, the direct-action budget, Framework-Guaranteed Conventions, Non-Overridable Rules — byte-identical.** Extracted from both goldens and compared: 2845/2271/383/1118/440 bytes, unchanged in every case.

**The two guard tests are not vacuous.** I broke what each one names and confirmed red.

- Rules deleted from the output style → `the_prose_rules_ban_categories_not_phrase_lists` FAILED at `claude_md_sections_tests.rs:1687`, "trusty-mpm: must carry the sycophancy rule".
- Two new rules removed from `trusty-mpm-teacher.md` only → FAILED at `:1715`, "trusty-mpm-teacher: must carry the don't-justify-the-restraint rule". The loop over `OUTPUT_STYLES` means a single-style drift cannot slip through.
- Duplicate reintroduced into `core.md` → `the_prose_rules_live_in_the_output_style_not_core` FAILED at `:1751`. All mutations reverted; tree clean.

**PM/subagent split — no drift.** `crates/trusty-mpm/src/assets/agents/BASE-AGENT.md` and the trusty-code copy are identical (`diff` EXIT=0). The two new rules are byte-identical between `BASE-AGENT.md` and the output style. No PM-facing rule moved to a channel a subagent needed, because a subagent never received the composed prompt in the first place.

**Multi-style — no drift.** The full `## Communication — Write Plainly` section is identical across all three styles except the deliberate per-style `Tone` line and one documented carve-out each (teacher's teaching-narration exemption, research's confidence-label exemption). Three copies, but the guard test now pins all three together, which is a tighter constraint than before this PR.

**The pre-existing-failure claim holds, on better evidence than the PR body gives.** `ensure_managed_config_dir_emits_the_frozen_skill_warning` is already listed in `docs/reference/test-ladder-baseline.md:42` as a known baseline flake, tracked at #4931, with exactly the described mechanism — `#[serial]` only serializes against other `#[serial]` tests, so a parallel test's globally-installed subscriber beats the capturing one. I also checked the ring-buffer hypothesis (a `LogBuffer::new(64)` overflow pushing the warning out): this PR adds zero new asset files, all 11 asset changes are `M`, so the deploy emits the same number of lines. Not caused here.

**Gates I ran:**

```
$ bash scripts/check_capabilities.sh   → tm-capabilities: up to date (7 generated files). EXIT=0
$ cargo test -p trusty-mpm --lib -- core::claude_md_sections core::instruction_pipeline \
      core::instruction_package core::instruction_overrides core::output_style
test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 4544 filtered out
EXIT=0
```

## Notes

Fix the two gates and this is an APPROVE. The relocation itself is the careful part of the PR and it is done carefully — every destination is a channel that genuinely loads, the deduplication is real rather than a deletion wearing a pointer, and the new negative guard (`the_prose_rules_live_in_the_output_style_not_core`) is what stops the duplicate creeping back, which is the failure mode this kind of change usually ships with.

One thing the PR body gets wrong without consequence: it justifies leaving `Clickable References` and the banned-word rule in place on the grounds that they are project-overridable. They sit in section `core`, which is `customization_tier: "fixed"` — the only fixed section, and `permits()` returns false for it. So they are not project-overridable today. Leaving them where they are preserves the status quo either way, so there is nothing to change in this diff; the reasoning just does not describe the mechanism.

bobmatnyc:
## Verdict: APPROVE

Re-review at `e05fb40f` from a fresh worktree. Both HIGH findings from [my earlier review](https://github.com/bobmatnyc/trusty-tools/pull/5002#issuecomment-5205402739) are cleared, and nothing that was already verified regressed across the rebase. Scoped to those three questions — I did not re-run the full pass.

The fix commit is 3 files, 4 insertions, 2 deletions. Nothing else moved.

## Finding 1 — asset staleness: cleared

```
$ bash scripts/check_agent_assets.sh
agent-assets: scanned 30 byte-parity file(s) (floor 20), all match; 4 pinned deviation(s) match upstream — OK.
EXIT=0
```

The re-pin is exactly one line, and it is the right one:

```
-crates/trusty-mpm/src/assets/agents/code-analyzer.md	55d810bc…7727bad0
+crates/trusty-mpm/src/assets/agents/code-analyzer.md	4396b3dd…ef0a277a
```

`4396b3dd` is the exact `current=` hash the gate reported when it failed. The other three pins — `code-critic.md`, `qa.md`, `web-qa.md` — are untouched, so this was a targeted re-pin and not a blanket `--update` that quietly re-baselined the other deviations. That was the specific risk I flagged, and it is cleanly avoided.

The tcode copy gained `skills: [code-review-standards]` and kept `tools: [read_file, grep, glob, list_dir, search_code, use_skill, finish_task]`. The `code-critic.md` precedent holds exactly as described — tcode's copy carries `tools:` on line 7 and `skills: [code-review-standards, contract-driven-testing]` on line 8, so an additive `skills:` line alongside a `tools:` restriction is the established shape, not a new one. `crates/trusty-code/src/assets/skills/code-review-standards/SKILL.md` exists, so the skill is present in the crate.

One correction to the reasoning as it reached me, with no bearing on the verdict. The claim was that the line "resolves" via `use_skill`. The frontmatter key itself does not: `crates/trusty-code/src/agents/md_loader.rs:222` documents `skills:` as **intentionally DROPPED**, and `:258` sets `append_skills: Vec::new()` unconditionally, with `:255` naming the gap — tcode has no consumer for agent-declared skills yet. So the declaration is inert in tcode today. What is true is the weaker, sufficient version: the agent holds `use_skill` and the crate ships the skill, so it remains reachable, and the line costs nothing while matching upstream. Given trusty-code is out of scope and the bar is minimum-to-green, adding an inert line that mirrors the code-critic precedent is the right call. Not a finding.

## Finding 2 — format: cleared

```
$ cargo fmt --check
EXIT=0
```

`claude_md_sections_tests.rs:1738` is now wrapped across two lines. Confirmed the fix is the reformat and nothing else.

The PR body correction is real and visible, not a silent swap. Line 90 carries it explicitly:

> **Correction.** An earlier revision of this body claimed `cargo fmt --check EXIT=0`. That run was real but **stale**: it happened before the final edit to `claude_md_sections_tests.rs`, which added a 101-char assertion, and fmt was not re-run after it. CI was right and the body was wrong. Fixed in `bc92fbb`; the block below is the current head.

That is the right way to correct a gate claim — it names what was wrong, why, and which commit fixed it, instead of quietly overwriting the number.

## No regression across the rebase

The fix commit touches no golden and no asset (`git diff --stat 5d7c9dcf..e05fb40f -- …/testdata/ …/assets/` is empty), and I re-derived everything against a freshly fetched `origin/main` rather than trusting the earlier run.

```
before=56783  after=50172  delta=-6611
```

Non-overridable blocks, origin/main vs head:

| Block | Bytes | Result |
|---|---|---|
| Prohibitions | 2845 → 2845 | IDENTICAL |
| Circuit Breakers | 2271 → 2271 | IDENTICAL |
| Direct-action budget | 383 → 383 | IDENTICAL |
| Framework-Guaranteed Conventions | 1118 → 1118 | IDENTICAL |
| Non-Overridable Rules | 440 → 440 | IDENTICAL |

Headings: four out (`Skills System`, `Skill Deployment`, `Agent Deployment`, `Auto-Configuration`), one in (`Skills and Agents — What You Need at Dispatch Time`). Unchanged from the pre-rebase state.

**Guard mutations still go red**, re-run at this head. I ran each in isolation this time after catching a truncation bug in my own first mutation script, so each failure is attributable to exactly one change:

| Mutation | Result |
|---|---|
| Rules deleted from `trusty-mpm.md` | FAILED `:1687` — "trusty-mpm: must carry the sycophancy rule" |
| One rule removed from `trusty-mpm-research.md` only | FAILED `:1719` — "trusty-mpm-research: must carry the trailing-emphatic-negation rule" |
| Duplicate reintroduced into `core.md` | FAILED `:1752` — "core restates \"**No praise for the user.**\" — the rules live in the output style only (#4574)" |

Single-style drift is still caught, which is the property that matters most for the three-copy arrangement. All mutations reverted; `git status --porcelain` empty; gates re-run green on the restored tree.

```
$ cargo test -p trusty-mpm --lib -- core::claude_md_sections core::instruction_pipeline \
      core::instruction_package core::instruction_overrides core::output_style
test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 4544 filtered out
tm-capabilities: up to date (7 generated files).
```

## The remaining pre-existing failure

`execute_doctor_against_test_daemon` is not caused by this branch, and I checked it rather than taking the claim:

```
$ cargo test -p trusty-mpm --lib -- client::executor::tests::execute_doctor_against_test_daemon --exact
test result: ok. 1 passed; 0 failed. finished in 7.68s
```

It passes in isolation at 7.68 s against the 10 s client timeout, which is exactly the mechanism `docs/reference/test-ladder-baseline.md:37` records — "takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load". Unloaded it fits; loaded it does not. Independently, `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client/` is empty, and the branch's full 19-file diff contains zero production Rust — only markdown assets, one test file, the pin table, and two changelog fragments. A daemon-timing integration test cannot be reached from that set.

## Notes

Verdict flips WARN → APPROVE. Both gates I blocked on are green, and the fix stayed inside its own blast radius instead of growing to fit the review.

Per the scope ruling, I did not treat trusty-code's `code-analyzer` declaring skills that tcode's loader drops as a finding — it is a recorded pre-existing gap in an out-of-scope crate, and this PR's job there was to reach green, which it did.

Issue #4005, opened 2026-07-26
trusty-memory /health false negative in tm doctor while MCP surface is verifiably serving


bobmatnyc:
## Context

`tm doctor` reports:

```
❌ memory — trusty-memory unreachable at 127.0.0.1:7070 (process running, port bound, but /health fails)
```

But the MCP surface on that same daemon is verifiably serving traffic at the same time: `memory_recall` and `memory_remember` round-trips succeeded within the same minutes — live evidence from session `f443c12d` on 2026-07-26 ~20:00Z. The daemon is not actually down; only the `/health` probe (or doctor's interpretation of it) is reporting failure.

This is a diagnostics-accuracy bug, not (necessarily) a daemon outage: either the `/health` endpoint has a bug/timeout under load/activity that the MCP request path doesn't hit, or `tm doctor`'s probe (timeout, retry, or interpretation) is too strict relative to what the endpoint actually needs to respond.

## Ask

Investigate and fix the false negative:

- Reproduce `/health` failing against `127.0.0.1:7070` while MCP `memory_recall`/`memory_remember` succeed against the same daemon in the same window.
- Determine whether the bug is in the `/health` endpoint itself (e.g. it does more/different work than a live MCP call and times out or errors under load) or in `tm doctor`'s probe (timeout too short, wrong retry behavior, misinterpreting a slow-but-valid response as failure).
- Note: this may be transient right after a `tm` restart — repro should account for that, and if it only reproduces in that window, the fix should specifically handle the post-restart warm-up case rather than reporting a hard failure.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5025, opened 2026-08-06
fix(trusty-mpm): split binary_provenance version skew by direction, amend ADR-0021 (#4964)

bobmatnyc:
Reconciles `binary_provenance` and [ADR-0021](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0021-cargo-bin-policy.md) with epic [#4964](https://github.com/bobmatnyc/trusty-tools/issues/4964)'s install-location consolidation.

🔴 **This must land BEFORE Phase 3.** Phase 3 flips the install destination to `$CARGO_HOME/bin`. Without this change, that flip makes `tm doctor` report `Fail` on a clean, successful, cargo-free upgrade — a new false-positive class manufactured by the fix itself. Landing this first is what unblocks Phase 3.

Own PR, own crate: this is `crates/trusty-mpm/`, and it touches nothing in `crates/trusty-installer/` or `trusty-common`. No overlap with [#5011](https://github.com/bobmatnyc/trusty-tools/pull/5011).

## The defect

[`provenance_report`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/core/binary_provenance.rs#L185) matches the running binary to a cargo ledger record by PATH (`same_file` against the cargo bin dir), then compares the record's version to the running binary's compiled-in version, and returns `Fail` on any mismatch — "the binary on disk is not the one cargo tracks."

Today a downloader-placed binary in `~/.local/bin` fails the path match and correctly reports `Unknown`. After Phase 1 (daemon upgrades stop invoking cargo) plus Phase 3 (destination flip), the identical binary lands at the same path a stale ledger entry describes: the path match succeeds, the versions disagree, `Fail`.

## The provenance model

A version disagreement at the SAME path means **cargo's record does not describe the binary on disk**. It does not identify who wrote that binary. Cargo normally rewrites `.crates2.json` whenever it writes, so a non-cargo writer is the common cause — but two cargo paths produce the same state: `cargo install --no-track --force` writes the binary and by design neither reads nor updates the ledger, and cargo renames a binary into place *before* saving the tracker, so an interrupted or failed `tracker.save()` leaves the record stale.

Every one of those paths lands on the same side of the split — the running binary being the newer file — so `Unknown` is honest for all of them. What the check can determine is the ordering, not the writer's identity, and the code and messages now say only that. (An earlier revision of this PR asserted the writer's identity as a certainty; corrected in `8c8d9933` after review.)

That state was a defect when the downloader wrote elsewhere. Post-#4964 it is also a sanctioned install. One `Fail` cannot say which, so the branch splits by which side is newer.

### Which inputs produce which outcome, before and after

| Input (running binary vs. cargo's ledger record) | Before | After | Changed? |
|---|---|---|---|
| no ledger at all | `Unknown` | `Unknown` | no |
| ledger has no record providing this binary name | `Unknown` | `Unknown` | no |
| two records provide the same binary name | `Fail` | `Fail` | no |
| running exe is NOT the file at `<cargo bin>/<name>` (shadowing, `~/.local/bin`, `target/debug`) | `Unknown` | `Unknown` | no |
| same path, running version **NEWER** than the record | **`Fail`** | **`Unknown`** | **YES — the whole change** |
| same path, running version **OLDER** than the record | `Fail` | `Fail` | no |
| same path, versions not orderable as semver | `Fail` | `Fail` | no |
| same path, versions agree, registry source | `Ok` | `Ok` | no |
| same path, versions agree, path source, source dir gone | `Fail` | `Fail` | no |
| same path, versions agree, path source, source dir survives | `Warn` | `Warn` | no |
| same path, versions agree, git / unrecognised source | `Warn` | `Warn` | no |

Exactly one cell moved. Every other verdict is byte-for-byte what it was.

### Why `Unknown` and not `Warn`

Widening the mismatch branch to `Warn` was explicitly rejected — it would have softened the real defect (an older binary sitting on a newer install) to accommodate the new legitimate case. That is why the split is by direction rather than by severity.

`Unknown` rather than `Warn` for the new class, for three reasons a skeptic can check independently:

1. **It is the module's own doctrine.** `binary_provenance`'s founding rule ([#4033](https://github.com/bobmatnyc/trusty-tools/issues/4033)) is that a probe which learned nothing reports `Unknown`, never `Ok`. For a downloader-placed file the ledger record describes a file that no longer exists — the check genuinely learned nothing about provenance.
2. **It makes the verdict invariant under the destination flip.** The same downloader-placed binary reports `Unknown` today from `~/.local/bin` (it fails the same-file guard). After Phase 3 it sits in `$CARGO_HOME/bin`. Nothing about the binary changed, only its directory, so the severity must not change either. Pinned by `downloader_placed_binary_is_unknown_in_either_directory`, which asserts both paths produce the same status.
3. **It does not soften the aggregate.** `CheckStatus::Unknown` ranks ABOVE `Warn` (2 vs 1) and never renders as healthy. This is not a quieting change; the owner's daily-driver doctor output is unchanged before and after Phase 3.

### Why the "running is newer" case cannot be proven benign, and why `Unknown` is still right

The check cannot distinguish a `tctl upgrade` placement from a hand-copied dev build — both produce a newer file at the cargo path with no metadata. That is precisely the case `Unknown` exists for. The message names both possibilities rather than asserting one. Claiming `Ok` would be false; claiming `Fail` would be a claim the inputs do not support.

### Acknowledged limits, stated rather than papered over

- **Same version.** When a non-cargo writer places the SAME version cargo last recorded, the check still reports `Ok`. It cannot tell the two writers apart there and does not need to — its claim is that the running version agrees with cargo's record, and that claim holds.
- **Deliberate downgrade.** `tctl install --force` documents a rollback to an "older-or-equal version" (`crates/trusty-installer/src/cli.rs:236`). Today that binary lands in `~/.local/bin` and reports `Unknown`; after Phase 3 it lands in `$CARGO_HOME/bin` and reports `Fail` — the same manufactured-failure shape this PR removes, surviving in the opposite direction. Accepted rather than fixed: an older binary on top of a newer record is genuinely worth flagging, a rollback is rare and explicit, and softening it would retire the real check. Whether Phase 3 should clear or refresh the stale record on placement is a call for the epic.

Both are documented in the module header and in the ADR amendment.

### `binary_reinstall` deliberately unchanged

[`reinstall_route`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-mpm/src/core/binary_reinstall.rs#L128) runs the same version-mismatch guard, but its outcome is `Refuse`, which is correct in BOTH directions: reinstalling from cargo when the ledger is stale would replace a newer binary with whatever the registry holds. Only doctor's severity was wrong, so only doctor changed.

## ADR-0021: amended, not superseded

Amended. The decision itself — `~/.cargo/bin` holds registry installs only, never path installs — is unchanged and unreversed by #4964. What the epic adds is a class the original text could not have enumerated, because at the time only cargo wrote to that directory.

Superseding would mint a new ADR number for a decision nobody is reversing, and would orphan the `ADR-0021` citations already embedded in `binary_provenance.rs` and `binary_reinstall.rs` operator messages.

The amendment records:

- the three sanctioned writers and their ledger presence, with the doctor verdict each maps to;
- **`Warn` remains the CORRECT and terminal verdict for a path install** — unchanged, not a defect to chase to `Ok`;
- prebuilt placement is permitted and is NOT a path install;
- the direction split and why each direction keeps its severity;
- Homebrew stays out of scope per the [owner's ruling](https://github.com/bobmatnyc/trusty-tools/issues/4964#issuecomment-5205828780); "single canonical location" means excluding Homebrew-installed copies.

**Status stays `Proposed`.** The Blocking Prerequisites (three unpublished/yanked crates) are untouched by this PR, so nothing about the adoption gate changed.

⚠️ **Independent of this work, noted so it is not lost:** two files are both numbered ADR-0021 — [`0021-cargo-bin-policy.md`](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0021-cargo-bin-policy.md) and [`0021-slack-inbound-hybrid-gateway-eventstream.md`](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0021-slack-inbound-hybrid-gateway-eventstream.md). Deliberately NOT fixed here — renumbering would rewrite citations across unrelated code.

## Tests — every one broken and confirmed red

New tests, and the two mutation rounds that prove each one is load-bearing rather than passing for the wrong reason.

**Mutation 1** — swap only the two `CheckStatus` values in `version_skew_verdict` (`LedgerStale`→`Fail`, `RunningStale`→`Unknown`), messages untouched so only the status assertions can fail:

```
test core::binary_provenance::tests::fails_on_version_disagreement ... FAILED
test daemon::doctor::doctor_binary_provenance::tests::provenance_fails_when_the_running_binary_is_older_than_the_ledger ... FAILED
test daemon::doctor::doctor_binary_provenance::tests::provenance_is_unknown_for_a_downloader_replaced_cargo_install ... FAILED
test core::binary_provenance::tests::unknown_when_downloader_replaced_the_cargo_install ... FAILED
test core::binary_provenance::tests::downloader_placed_binary_is_unknown_in_either_directory ... FAILED
test result: FAILED. 28 passed; 5 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.01s
```

**Mutation 2** — replace the `semver` parse in `classify_version_skew` with a raw string comparison:

```
test core::binary_provenance::tests::fails_when_versions_cannot_be_ordered ... FAILED
test core::binary_provenance::tests::skew_orders_prereleases_by_semver_not_string ... FAILED
test core::binary_provenance::tests::skew_is_unordered_for_unparseable_versions ... FAILED
test result: FAILED. 30 passed; 3 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.04s
```

Every new test is killed by a mutation of the exact behaviour it names. `skew_orders_prereleases_by_semver_not_string` exists because `1.4.0-rc.1` sorts AFTER `1.4.0` as a string and BEFORE it as a version — string comparison would have excused a real downgrade as a stale ledger.

Restored, all green:

```
test result: ok. 33 passed; 0 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.01s
```

## Gates — rung 5

```
cargo check -p trusty-mpm --all-targets               EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0

sld-lint: scanned 56 spec doc(s) + 3115 code file(s); 0 error(s), 0 warning(s)
line-cap: measured 3742 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
test-pointers: resolved 22085 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
doc-numbers: scanned 102 doc(s) / 96 claim(s) (floors 20/20); 4 grandfathered, 0 violations — OK.
changelog-fragment gate: scanned 7 changed path(s); all 1 crate(s) with source changes are recorded.
```

`cargo test -p trusty-mpm`:

```
test result: FAILED. 4732 passed; 2 failed; 7 ignored; 0 measured; 0 filtered out; finished in 154.99s

failures:
```

Both are the documented baseline flakes ([#4931](https://github.com/bobmatnyc/trusty-tools/issues/4931)), proven pre-existing rather than silenced:

- `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/core/managed_config* crates/trusty-mpm/src/client` → **0 paths**. This diff touches 7 files, none in either module.
- Re-run in isolation on this branch, both pass:

```
test core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 4739 filtered out; finished in 8.84s
```

Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d, or narrowed.

## Dependency

`semver = { workspace = true }` added to `crates/trusty-mpm/Cargo.toml`. Already in `[workspace.dependencies]`; no new external crate enters the workspace. `trusty-common`'s `is_newer` could not be reused — it is private, returns `bool`, and collapses "older" and "unparseable" into the same `false`, which is exactly the distinction this change depends on.

Refs [#4964](https://github.com/bobmatnyc/trusty-tools/issues/4964). The epic stays open.


bobmatnyc:
## Verdict: APPROVE

Adversarial review, rung 5. Reviewed against the spec and the code only. Ran every claim rather than reading the PR's table: both mutation rounds reproduced, a third mutation of my own, the full crate suite, and an empirical probe of the semver edge cases.

Zero CRITICAL, zero HIGH. Four findings, all documentation: three MEDIUM, one LOW. None of them changes a verdict the code produces.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/core/binary_provenance.rs` | 39-41, 348-350, 443-450 | "cargo rewrites its ledger whenever it writes, so a disagreement at the same path **can only mean** a non-cargo writer replaced the file" is false — `cargo install --no-track --force` is a documented cargo path that writes the binary and leaves the record stale | Soften "can only mean" to "almost always means" at all three sites; add `cargo install --no-track` and the interrupted-install window to the writers the operator message enumerates | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/generate/doctor.rs` | 148 | The operator-facing description of the check still says it "Fails ... when the ledger's recorded version disagrees with what is running" — the exact cell this PR moved. Its checked-in render at `crates/trusty-mpm/src/assets/skills/tm-capabilities/references/doctor.md:37` carries the same stale sentence, and that file is a bundled asset deployed into every user's `.claude/skills/` | Reword to "Fails when the running binary is OLDER than the ledger's record, or the two cannot be ordered as semver; reports UNKNOWN when it is newer (a prebuilt-installer placement)", then `tm generate capabilities` | Fix here |
| MEDIUM | `docs/adr/0021-cargo-bin-policy.md` | Amendment 1, three-writer table | The table maps "Prebuilt placement" flatly to `Unknown`, but the direction split three paragraphs below maps a placement OLDER than the ledger record to `Fail`. The table contradicts its own section, and the downgrade-placement case is a residual Phase-3 false positive the module's known-limits paragraph does not name | Qualify the row: "`Unknown` when the placed version is newer; `Fail` on a downgrade placement — see the direction split below". Add the downgrade case to the module header's known-limit paragraph beside the same-version one | Fix here |
| LOW | `docs/adr/0021-cargo-bin-policy.md` | Consequences | Two bullets are now false for the class the amendment introduces: "Every binary in `~/.cargo/bin` now has an auditable source: `cargo install --list` shows the version" and "Orphaned binaries ... cannot happen — every install is tied to a crates.io version". "What this amendment does NOT settle" covers the enforcement gate but not these | One sentence in the amendment noting the two Consequences bullets are scoped to cargo-written files | Fix here |

## The central claim — attacked, and it does not hold as written

The fix rests on: cargo rewrites `.crates2.json` whenever it writes a binary, therefore a version mismatch at the cargo bin path can only mean a non-cargo writer. I found two paths where cargo itself leaves the ledger and the binary disagreeing at the same path.

**`cargo install --no-track --force`.** Verified present in the installed toolchain (`cargo install --help` → `--no-track  Do not save tracking information`). Cargo's own documentation for it: "This flag tells Cargo not to use or create that file. With this flag, Cargo will refuse to overwrite any existing files unless the `--force` flag is used." So `--no-track --force` over a previously tracked install writes a new binary into `$CARGO_HOME/bin` and leaves the prior `.crates2.json` record untouched and stale. Cargo did it; no downloader was involved.

**The write-then-record window.** Cargo renames binaries from its staging dir into `$CARGO_HOME/bin` before it calls `tracker.save()` — it has to, since it cannot record what it has not yet written. A SIGKILL, a power loss, or a `tracker.save()` failure (read-only `.crates2.json`, full disk) in that window leaves exactly this state. Lower confidence than the `--no-track` case, which I verified directly.

**Why this is MEDIUM and not HIGH.** I looked for a counterexample in the harmful direction and did not find one. Every path I can construct lands on the same side: the running binary is the newer, legitimate file. `Unknown` — "the binary is not stale; what cannot be verified is where it came from" — is honest for all of them. No verdict the code produces is wrong. What is wrong is the premise stated three times as an absolute, once in text the operator reads: *"this binary was placed by a writer that keeps no cargo metadata — the prebuilt installer, a package manager, or a manual copy."* Under `--no-track` that sends someone hunting for a downloader that never ran. The message's own tail already hedges ("what cannot be verified from here is where it came from"), so the fix is to make the head agree with the tail.

The inference is not load-bearing for safety, only for the explanation. That is what keeps this out of HIGH.

## The verdict table — verified mechanically, not read

I did not trust the table. I filtered the diff of `binary_provenance.rs` down to non-comment lines. Inside `provenance_report` the entire behavioural change is one statement:

```
-        return (
-            CheckStatus::Fail,
-            format!("the running `{bin_name}` reports version {running_version}, but ...
-        );
+        return version_skew_verdict(bin_name, running_version, record, &installed, &where_from);
```

Every other verdict-producing arm — no ledger, no record, duplicate installs, the `same_file` shadowing guard, path-source-gone, live path, git, other, registry-match — is byte-identical. No second cell can have moved, because no other outcome-producing code was edited. `Warn` remains terminal for a path install. Confirmed, independent of the PR body.

## The `Unknown` vs `Warn` choice — all three justifications check out

**Ranking (justification 3).** `core/doctor.rs:55-66`: `Ok = 0`, `Warn = 1`, `Unknown = 2`, `Fail = 3`. `Unknown` ranks above `Warn`, and it is pinned by a test, not just a comment — `daemon/doctor_tests.rs:578` `unknown_never_aggregates_to_ok` asserts a `Warn` + `Unknown` report aggregates to `Unknown`. I also walked every render surface the owner sees, since a ranking can be correct while a formatter undoes it:

- `bin/tm/commands/misc.rs:199` — "one or more checks could not be determined — health is UNKNOWN", against `Warn`'s "passed with warnings"
- `bin/tm/commands/misc.rs:364` — `❔`, against `Warn`'s `⚠️`
- `telegram/formatter/mod.rs:560` and `slack/formatter/mod.rs:458` — "undetermined", against "warnings"

`tm doctor` sets no exit code from `overall`, so there is no gate to soften. Nothing treats `Unknown` as more benign than `Warn` anywhere. The justification stands.

**Destination invariance (justification 2).** Traced rather than assumed. `same_file` (`binary_provenance.rs:480`) canonicalises both paths; a binary at `~/.local/bin/tm` never canonicalises to `$CARGO_HOME/bin/tm`, so it exits at the shadowing guard with `Unknown` — today, unchanged by this diff. The claim about current behaviour is correct. See MEDIUM-3 for the one direction where the invariance does not hold.

**Module doctrine (justification 1).** Consistent with the `UNVERIFIABLE REPORTS UNKNOWN` rule in the module header and with the #4005 rationale on `CheckStatus::Unknown`.

## `binary_reinstall` — the reasoning holds

`binary_reinstall.rs:128` runs the same `record.version != running_version` guard and returns `Refuse`. I checked whether the unsound inference is embedded there too: it is not, because `reinstall_route` never asks *which* writer placed the file. It asks whether reinstalling is safe, and the answer is no in both directions — reinstalling over a newer downloader-placed binary replaces it with whatever the registry holds, and reinstalling over an older one is the #4033 case. `Refuse` is fail-closed and direction-independent. Leaving it alone is right, not a gap.

## Semver — probed, not assumed

I added a temporary probe and ran it. Actual output:

```
1.3.1      vs 1.3.1+build.5 => LedgerStale
1.3.1+b5   vs 1.3.1         => RunningStale
1.3.5      vs 1.3.5-dev     => RunningStale
1.3.5-dev  vs 1.3.5         => LedgerStale
1.4.0-rc.1 vs 1.4.0-rc.2    => LedgerStale
```

Every exotic-but-valid version is ordered correctly, including build metadata, which I expected to collapse to `Equal` and it does not. `Unordered` therefore catches only genuinely unparseable input (`"1.3"`, `"wat"`), which is what it claims to do. `Ordering::Equal => Unordered` is unreachable from the caller — no two distinct strings parse to an equal `Version` — but it is defensive, documented, and pinned by `skew_is_unordered_for_unparseable_versions`. Correct as written. (Probe removed; tree left clean.)

## Test honesty — both mutations reproduced, plus a third

**Mutation 1**, swapping the two `CheckStatus` values in `version_skew_verdict`, messages untouched:

```
test core::binary_provenance::tests::unknown_when_downloader_replaced_the_cargo_install ... FAILED
test core::binary_provenance::tests::downloader_placed_binary_is_unknown_in_either_directory ... FAILED
test core::binary_provenance::tests::fails_on_version_disagreement ... FAILED
test daemon::doctor::doctor_binary_provenance::tests::provenance_fails_when_the_running_binary_is_older_than_the_ledger ... FAILED
test daemon::doctor::doctor_binary_provenance::tests::provenance_is_unknown_for_a_downloader_replaced_cargo_install ... FAILED
test result: FAILED. 28 passed; 5 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.01s
```

5 failures, the same 5 tests. Count and names match.

**Mutation 2**, replacing the semver parse with a raw string comparison:

```
test core::binary_provenance::tests::skew_orders_prereleases_by_semver_not_string ... FAILED
test core::binary_provenance::tests::skew_is_unordered_for_unparseable_versions ... FAILED
test core::binary_provenance::tests::fails_when_versions_cannot_be_ordered ... FAILED
test result: FAILED. 30 passed; 3 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.01s
```

3 failures, the same 3 tests. Count and names match.

**Mutation 3, mine** — the harder question: is there a mutation that should kill a test and does not? I made `Unknown` and `Warn` interchangeable in the new arm (`LedgerStale => Warn`, everything else untouched), which is precisely the substitution justification 3 argues against:

```
test core::binary_provenance::tests::unknown_when_downloader_replaced_the_cargo_install ... FAILED
test core::binary_provenance::tests::downloader_placed_binary_is_unknown_in_either_directory ... FAILED
test daemon::doctor::doctor_binary_provenance::tests::provenance_is_unknown_for_a_downloader_replaced_cargo_install ... FAILED
test result: FAILED. 30 passed; 3 failed; 0 ignored; 0 measured; 4708 filtered out; finished in 0.01s
```

Caught by 3 tests. The `Unknown` choice is pinned specifically, not incidentally — `assert_ne!(status, Fail)` alone would have let `Warn` through, and the tests assert the exact status as well. Baseline restored: `33 passed; 0 failed`.

## Gates — run independently

```
cargo fmt --check                                         EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings    EXIT=0
  line-cap: measured 3742 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.

cargo test -p trusty-mpm
  test result: FAILED. 4733 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 235.78s
  failures:
```

One failure on my run, not two — `ensure_managed_config_dir_emits_the_frozen_skill_warning` passed here and failed on the author's. That difference is itself evidence of flakiness rather than of a regression. The pre-existing claim checks out both ways stated:

```
git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/client crates/trusty-mpm/src/core/managed_config*
  → 0 paths   (the diff is 7 files, none in either module)

cargo test -p trusty-mpm --lib execute_doctor_against_test_daemon
  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4740 filtered out; finished in 7.94s
```

Nothing was `#[ignore]`d, `cfg`-gated, `--exclude`d, or narrowed. Judged on local gates; GitHub Actions is in a major outage.

## Notes

The residual case behind MEDIUM-3 is the one worth a decision rather than a doc edit, so stating it plainly: `tctl install` places BOM-pinned versions, and `crates/trusty-installer/src/cli.rs:236` documents a `--force` override specifically for replacing a registered supervisor with an "older-or-equal version". A sanctioned rollback that places an older binary than a newer cargo record reports `Unknown` today from `~/.local/bin` and will report `Fail` after Phase 3 — same binary, same operation, different directory. That is the same shape of manufactured failure this PR removes, surviving in the opposite direction. The code's choice is defensible: an older binary over a newer record is worth flagging, and softening it was explicitly rejected. But the ADR table currently asserts the opposite, and the limit is unstated. Whether Phase 3 needs to handle it is a call for the epic, not a blocker here.

The duplicate ADR-0021 numbering was excluded from review per the dispatch.

Read-only review from `.claude/worktrees/critic-5025`. Mutations applied and reverted in that worktree only; `git status --porcelain` clean afterwards.

Issue #1219, opened 2026-06-15
fix(trusty-mpm): supervisor and daemon hold independent SessionManager views → stale managed-session state


bobmatnyc:
Found by live e2e validation of main @ f5c55e69.

The `tm supervisor` process and the `trusty-mpm` daemon each instantiate their own `SessionManager` over the same on-disk `~/.trusty-mpm/session-manager/sessions.json`, with no cross-process notification. The daemon loads via a `OnceCell` at first access; the supervisor loads at startup.

Observed: after the supervisor auto-resumed a stopped session (disk sessions.json correctly showed `active`), the daemon's `GET /api/v1/sessions/managed/{id}` continued returning `state: stopped` indefinitely. Operators querying the daemon while a supervisor runs see stale state.

Options: reload-on-read / mtime check on the store, a file watcher, or have the supervisor drive the daemon via its HTTP API instead of owning a second SessionManager. Acceptance: daemon API reflects supervisor-driven transitions within one poll interval.

PR #5027, opened 2026-08-06
fix(trusty-mpm): detect, disclose, and repair a corrupt session store

### LINKED ISSUE #5007: session-manager store corruption: undetected, unrecoverable, silently masked by tm ls fallback
`~/.trusty-mpm/session-manager/sessions.json` was found corrupt on the owner's live machine on 2026-08-06: a complete, valid JSON document ending at byte 145090, followed by 1111 stale bytes belonging to a longer earlier version of the same document. `SessionManager::load` (`crates/trusty-mpm/src/session_manager/store.rs`) fails with `trailing characters at line 3755 column 2`.

Every mutation path failed, not just rename — `upsert` reloads the store before saving, so the daemon could not write because it could not read. It was silent: `tm ls` kept working because `session list: reload failed: …; returning last-known in-memory set` (`crates/trusty-mpm/src/session_manager/manager.rs:357`) falls back to the in-memory set, so the daemon looked healthy. It surfaced only because the owner attempted a rename and the CLI had just stopped swallowing the error body ([#5004](https://github.com/bobmatnyc/trusty-tools/pull/5004)). The store cannot self-heal: read precedes write on every path, so a corrupt file is permanently wedged until a human truncates it.

Repaired manually by truncating to the end of the valid document, after verifying the discarded bytes held only a session id already present in the valid prefix. Backup at `~/.trusty-mpm/session-manager/sessions.json.corrupt-backup-20260806-090117`.

Cause is unknown. `Store::save` (`crates/trusty-mpm/src/session_manager/store.rs:227`) writes atomically (temp file plus rename), so the daemon's own write path cannot produce this shape. Untested hypothesis only: a shorter document written over a longer one with the old tail left behind is the exact signature of a writer that opens the file for writing without truncating — some process other than `Store::save` touching the path directly. Not asserted as the cause.

Resolution, independent of the unknown cause:
1. Detect: validate the store on load and fail with a diagnostic naming the file and byte offset, plus a `tm doctor` check so corruption is visible before someone stumbles on it.
2. Recover: a corrupt store should be repairable without hand-editing JSON — back up and truncate to the last valid document, or equivalent.
3. Stop the silent mask: the `manager.rs:357` fallback made a fully wedged store look healthy. A read failure should surface, not be absorbed.
4. Identify the writer: enumerate everything that opens `sessions.json` for writing besides `Store::save`.

Searched open and closed issues for "sessions.json", "session store", "corrupt", "trailing characters", "serialization error", "out-of-band", "store integrity", "session-manager". Closest results: [#3715](https://github.com/bobmatnyc/trusty-tools/issues/3715) and [#3764](https://github.com/bobmatnyc/trusty-tools/issues/3764) are a different failure family (out-of-band worktree deletion, not store corruption). [#5001](https://github.com/bobmatnyc/trusty-tools/issues/5001) and [#5004](https://github.com/bobmatnyc/trusty-tools/pull/5004) are this same corruption event's rename symptom — [#5004](https://github.com/bobmatnyc/trusty-tools/pull/5004) fixes only the CLI swallowing the daemon's error message. It does not detect, prevent, or recover from the corruption, and should not be treated as closing this. No open or closed ticket covers detection, recovery, or the silent-fallback masking, so filing new.

bobmatnyc:
## Defect

`~/.trusty-mpm/session-manager/sessions.json` was corrupt on the owner's machine on 2026-08-06: a complete JSON document ending at byte 145090, followed by 1111 stale bytes from a longer earlier version of the same document.

`SessionStore::upsert` reloads the store before saving, so a file the daemon could not read was a file it could never rewrite — **every** mutation path failed, not just the rename that surfaced it. It was silent: `SessionManager::list` falls back to the last-known in-memory set, so `tm ls` kept printing a healthy fleet. The store cannot self-heal, so it stayed wedged until a human hand-truncated the JSON.

Closes [#5007](https://github.com/bobmatnyc/trusty-tools/issues/5007). [#5004](https://github.com/bobmatnyc/trusty-tools/pull/5004) fixed only the CLI swallowing the daemon's message; it did not detect, prevent, or recover from this.

## Resolution

**1. Detect.** A parse failure is now `StoreError::Corrupt`, carrying a `StoreIntegrity` diagnosis whose `Display` names the file, the byte offset where the valid document ends, how many trailing bytes follow, and the repair command — replacing the bare `trailing characters at line 3755 column 2`. The offset is *measured* (a streaming deserializer's `byte_offset()` after one complete value), not inferred from serde's line/column. A new read-only `session_store` check in `tm doctor` reports the same before anyone stumbles into it.

**2. Recover.** `tm repair session-store` automates exactly the manual repair: verify the discarded bytes name no session absent from the recovered document, copy the file to `sessions.json.corrupt-backup-<YYYYMMDD-HHMMSS>` (the same name shape as the 2026-08-06 manual backup), then truncate to the longest complete document at the head. It **refuses** when the tail names an unknown session id, naming the ids; `--force` overrides, `--dry-run` shows the cut and writes nothing (including no backup). A store that parses is a no-op success, so re-running after a repair is safe.

**3. Stop it being silent — and why the fallback stays.** The fallback exists for [#1219](https://github.com/bobmatnyc/trusty-tools/issues/1219)'s follow-up: a transient read error (an NFS hiccup, a momentarily unreadable file) must not make the daemon report an *empty* fleet, which would mislead the supervisor into re-provisioning sessions that are still running. That reasoning is sound and the fallback is kept. What was wrong is that it left no trace. Now:

- `SessionStore` records a `StoreDegradation` on any failed reload and clears it on the next success.
- `log_reload_fallback` splits severity: corruption logs at ERROR saying every write is failing; a possibly-transient I/O error stays WARN.
- `GET /api/v1/sessions/managed` carries `store_health` — **omitted entirely** when healthy, so a healthy response is byte-identical to before and no existing consumer changes shape.
- `fetch_managed_raw` — the single path `tm ls`, `tm ls --json`, and the interactive picker all go through — prints a stderr banner when it is present, so stdout stays machine-readable.

**4. Every writer of that path.** Enumerated below. The corruption's cause was unknown; the enumeration produced a mechanism that fits the observed bytes exactly, and it is fixed here.

## Everything that writes `sessions.json`

| Writer | Where | Mechanism |
|---|---|---|
| `SessionStore::save` | [`session_manager/store.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5007-session-store-resilience/crates/trusty-mpm/src/session_manager/store.rs) | The **only** code in the repo that writes the file. Staging write + `rename`. Reached from `upsert`, `upsert_many`, `remove`, `remove_many`. |
| the daemon | `DaemonState::session_manager()` → `SessionManager::new(<fw_root>/session-manager, …)` | via `SessionStore::save` |
| `tm supervisor` | [`bin/tm/commands/supervisor.rs:86-95`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5007-session-store-resilience/crates/trusty-mpm/src/bin/tm/commands/supervisor.rs) | **A second process** with its own `SessionManager` over the same data dir — this is by design, and `store.rs`'s own #1219 doc says so |
| `tm install --reset-agents-workspaces` | [`core/agent_reset_workspace.rs:167`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5007-session-store-resilience/crates/trusty-mpm/src/core/agent_reset_workspace.rs) | Builds its **own** `SessionStore` in the `tm` CLI process. Reads only (`store.all()`) in production; the `upsert` at line 400 is in its test module |

Readers only, confirmed non-writing: `trusty-agents`' `session_state` tools ([`tools/session_state/store.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5007-session-store-resilience/crates/trusty-agents/src/tools/session_state/store.rs)) reach the path through a plain `std::fs::read_to_string`. No shell script, no non-Rust code, and nothing outside `crates/` touches it.

### Hypothesis for the cause — evidenced, not asserted

`save()` staged through `self.path.with_extension("json.tmp")` — one fixed name, `sessions.json.tmp`, shared by **every** writer in **every** process. The `rename` is atomic; the staging write is not exclusive. Two writers each open that one name with truncate (both truncate early, while it is short), then each streams its own serialization in at its own file offset. The resulting file's **length is the longer document's** and its **head is whichever writer finished last**. If the shorter one finishes last you get a complete shorter document followed by the longer document's tail — the observed shape, with a tail length equal to the difference between the two serializations.

What supports it: the table above shows at least two processes (the daemon and `tm supervisor`) that legitimately write this file concurrently, plus a third `SessionStore` built by the CLI. What is **not** proven: that this is what actually happened on 2026-08-06. Nobody was watching, the daemon logs for the window are gone, and I did not attempt to reproduce a race against the live store. Treat it as the mechanism that fits, not as the established cause.

Fixed regardless: staging is now `sessions.json.tmp.<pid>.<uuid>`, private to each `SessionStore` instance, and a failed rename removes it rather than leaving it behind.

### One more data-loss path, found on the way

`read_file` answered **every** read failure with an empty store — `Err(_) => Ok((StoredData::default(), None))`. A permissions failure, an EIO, or a non-UTF-8 file therefore produced a store with zero sessions, and the very next `save()` would have written that emptiness over the operator's entire fleet. Only `ErrorKind::NotFound` now means "fresh store"; `InvalidData` is reported as corruption; everything else propagates.

## Tests — each one proved red by breaking what it names

Two mutation rounds, reverting the production mechanism each test is named for, then restoring. Raw output:

**Round 1** — revert `read_file`'s corruption typing, revert `read_file`'s error swallowing, revert the staging path to the shared `sessions.json.tmp`:

```
test session_manager::store_integrity::store_integrity_tests::two_stores_over_one_path_do_not_share_a_staging_file ... FAILED
test session_manager::store_integrity::store_integrity_tests::load_propagates_a_read_error_instead_of_reporting_an_empty_store ... FAILED
test session_manager::store_integrity::store_integrity_tests::store_records_and_clears_a_reload_failure ... FAILED
test session_manager::store_integrity::store_integrity_tests::load_reports_a_trailing_tail_as_corruption_with_a_byte_offset ... FAILED
test result: FAILED. 18 passed; 4 failed; 0 ignored; 0 measured; 4739 filtered out
```

**Round 2** — drop the orphan refusal, skip the backup write, mis-measure the valid prefix, downgrade the doctor `Fail`/`Unknown` verdicts to `Ok`, and neuter the `tm ls` banner:

```
test daemon::doctor::doctor_session_store::doctor_session_store_tests::session_store_check_fails_on_a_trailing_tail_and_names_the_offset ... FAILED
test daemon::doctor::doctor_session_store::doctor_session_store_tests::session_store_check_is_unknown_when_the_file_cannot_be_read ... FAILED
test session_manager::store_integrity::store_integrity_tests::analyze_locates_the_valid_prefix_of_a_trailing_tail ... FAILED
test session_manager::store_integrity::store_integrity_tests::repair_refuses_an_orphaned_tail_without_force ... FAILED
test session_manager::store_integrity::store_integrity_tests::repair_backs_up_then_truncates_to_the_valid_document ... FAILED
test session_manager::store_integrity::store_integrity_tests::valid_prefix_bytes_measures_a_whole_document ... FAILED
test commands::repair_session_store::repair_session_store_tests::repair_session_store_truncates_and_backs_up ... FAILED
test commands::session_picker::tests::store_banner_marks_a_transient_read_failure_as_stale_not_corrupt ... FAILED
test commands::session_picker::tests::store_banner_names_corruption_and_the_repair_command ... FAILED
```

The 18 tests that stayed green in round 1 are the check that these are not coupled — each mutation reddens only the tests naming it.

**The fixture is the real corruption shape, not generic garbage.** `corrupt_with_stale_tail` serializes the *same two sessions* twice through the production `SessionStore::save` — once with a long task, once short — then lays the short document over the long one. That yields a complete document followed by the longer document's tail, with an identical session set, which is why the tail names no orphan and the repair proceeds exactly as the manual one did. A `{ not json ]` fixture (which the pre-existing `store_cached_accessors_ignore_disk` uses) would have passed the detection assertions for the wrong reason.

## Gates — rung 5

```
cargo fmt --check                                    → clean
cargo check -p trusty-mpm --all-targets              → clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings → clean
bash scripts/check_line_cap.sh   → 3749 files, 7 allowlisted, 0 violations — OK
bash scripts/check_sld.sh        → 56 specs + 3119 code files, 0 errors, 0 warnings
bash scripts/check_test_pointers.sh → 22127 citations resolved, 0 dangling — OK
```

```
cargo test -p trusty-mpm
test result: FAILED. 4753 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 456.08s

failures:
```

That one failure is the documented baseline flake ([`docs/reference/test-ladder-baseline.md`](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/reference/test-ladder-baseline.md) line 37: *"takes 9–13 s against a 10 s client timeout, so it loses on timing alone under any load"*). Its message is the documented shape — `daemon unreachable: error sending request` — and it passes in isolation:

```
cargo test -p trusty-mpm --lib execute_doctor_against_test_daemon
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4760 filtered out; finished in 8.46s
```

8.46 s against the 10 s timeout — the margin the baseline doc describes. The new `session_store` probe is one synchronous read of one small file and does not move that number meaningfully.

## Notes for review

- **Behaviour change worth a second look:** `read_file` no longer treats a non-`NotFound` read error as an empty store. `manager_create_store_failure_reaps_orphan` depended on the old behaviour — its sabotage (replace `sessions.json` with a non-empty directory) now fails the reload *before* `create_session` makes anything to orphan. The test's assertions are unchanged; only its sabotage moved to a read-only parent directory, which fails the write and not the read. That is documented inline at the change site.
- **`tm doctor` grew a 30th check**, so both expected-check lists and `run_doctor_produces_twenty_nine_checks` → `…_thirty_checks` were updated.
- **Wire compatibility:** `store_health` is `skip_serializing_if = "Option::is_none"` on the daemon and `#[serde(default)]` on the client, so an old client against a new daemon and a new client against an old daemon both work. `list_response_omits_store_health_when_healthy` pins the omission.
- **The owner's live store was never touched.** Every test uses a `TempDir`; no daemon was bounced.
- **In-flight collisions:** none. This PR touches only `crates/trusty-mpm/`, and nothing in `trusty-installer`, `trusty-common`'s `bin_resolve`, or `memory_core` — the areas [#5011](https://github.com/bobmatnyc/trusty-tools/pull/5011) and [#5013](https://github.com/bobmatnyc/trusty-tools/pull/5013) own.

## What this does not do

It does not prove the cause of the 2026-08-06 corruption, and it does not prevent an out-of-band writer that opens the path directly — no such writer exists in this repo, but nothing here would stop one. What it guarantees is that the next occurrence is detected on the first read, visible in `tm ls` and `tm doctor`, and repairable with one command.


bobmatnyc:
## Verdict: WARN

Adversarial review, rung 5. Reviewed against the spec and the diff only. Built and ran everything below in a throwaway worktree off `19772543`; the owner's live store was never read or written.

The three headline claims hold. The fallback is intact and the change is strictly in the safer direction. The staging fix closes the window it names. The `read_file` data-loss path was real on `origin/main` and is now closed — and it was worse than the PR says. What blocks a clean approve is that the two new surfaces built to make a wedged store *visible* and *fixable* validate a different schema than the store itself does, so both can report health on a store the daemon cannot load.

---

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/daemon/doctor_session_store.rs` | 66 | The probe parses `serde_json::Value`; the store parses `StoredData`. A store the daemon cannot load reports `Ok` with a record count. | Parse `StoredData` (or call `SessionStore::load`) and count from that. | Fix here |
| HIGH | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 235 | `plan()` parses `Value`, so a `StoredData`-only failure returns `NotCorrupt` — the store says "run `tm repair session-store`" and the command says "nothing to repair". | Add a `NotAStore` variant for valid-JSON-invalid-schema; suppress the repair advice in `diagnostic()` when `trailing_bytes() == 0`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 246-251 | The orphan refusal misses a session id that straddles the cut, because `uuid_tokens` only sees the bytes after `valid`. | Scan `raw[valid.saturating_sub(UUID_LEN - 1)..]` — 35 extra bytes, and over-reporting is already the stated design. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 275, 283, 294-295 | `repair()` reads the file twice and slices the second read at an offset measured from the first: panic if it shrank, silent mis-truncation if it changed. | Carry the bytes on `RepairPlan` and truncate those; or re-check `kept <= raw.len() && raw.is_char_boundary(kept)` and re-plan. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/session_picker.rs` | 186 | The stderr routing — the one thing keeping `tm ls --json` machine-readable — is asserted by no test. `eprintln!` → `println!` kills nothing. | Assert stdout/stderr separation, or write through an injectable sink and assert the sink. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/session_manager/store_health.rs` | 51 | The severity split the function exists for is asserted by no test. Swapping `error!`/`warn!` kills nothing. | Capture the event level, or return a `Level` from a pure helper and assert it. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/daemon/state/core.rs` | 748-763 | A store corrupt at daemon boot silently becomes an empty `$TMPDIR` store: `store_health` is `None`, no banner, `tm ls` shows a clean empty fleet. Pre-existing, but it is the sequence the disclosure was built for. | Record the boot failure on the fallback manager and serve it through the same `store_health` field. | Parent |
| LOW | `crates/trusty-mpm/src/session_manager/store.rs` | 151-159, 419-424 | Per-instance staging names accumulate without a reaper. `save` cleans up on a failed rename, but a crash between write and rename orphans ~145 KB forever, once per store instance. Previously at most one file could exist. | On `load`, unlink `sessions.json.tmp.*` siblings older than an hour. | Fix here |
| LOW | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 295 | The repair truncates in place while `save()` stages-and-renames precisely so a cross-process reader never sees a torn file (#1219). Small exposure — `plan` only proceeds on an already-wedged store — but it is the one place a reader could see a zero-length store. | Stage to `<path>.repair.<pid>` and rename. | Fix here |

---

## The two HIGH findings, measured

`SessionRecord` (`record.rs:190-205`) has six fields with no `#[serde(default)]`, so plenty of valid JSON is not a valid store. The probe below uses this PR's **own** healthy-case fixture from `doctor_session_store_tests.rs:33`:

```
PROBE-E empty-records doctor = Ok / …/sessions.json parses cleanly — 2 session record(s), 28 byte(s)
PROBE-E empty-records store  = Err("…/sessions.json is corrupt: missing field `id` at line 1 column 19
                                   (line 1, column 19); a complete JSON document ends at byte 28 of 28,
                                   followed by 0 trailing byte(s). Repair with `tm repair session-store`")
```

`tm doctor` positively asserts health, with a record count, for a file that blocks every write. That is the condition the probe was added to catch.

And the repair dead end, on `{"sessions": 5}`:

```
PROBE-A store error = …/sessions.json is corrupt: invalid type: integer `5`, expected a map
                      (line 1, column 14); a complete JSON document ends at byte 15 of 15,
                      followed by 0 trailing byte(s). Repair with `tm repair session-store`
PROBE-A plan        = Err(session store …/sessions.json parses cleanly; nothing to repair)
```

Three surfaces, three answers. The diagnostic also describes trailing junk ("0 trailing byte(s)") that does not exist, and advertises a command that refuses. The trigger is not hypothetical: a hand-edit — the documented 2026-08-06 remedy — or a daemon rollback past a new `ManagedSessionState` variant both produce exactly this.

`plan()` also cannot touch a non-UTF-8 store (`read_to_string` → `RepairError::Io`), while `read_file` classifies non-UTF-8 as `Corrupt` and points at the same command. Same family.

## The orphan hole

The cut lands at an arbitrary offset inside the longer document, so a session key can span it, and `uuid_tokens(tail)` then sees only its second half:

```
PROBE-B orphan_ids = []
PROBE-B tail       = "555555555555\",\"state\":\"Active\"}}}"
PROBE-B repair     = Ok(33)
```

The discarded tail belonged to `aaaaaaaa-2222-3333-4444-555555555555`, absent from the recovered document, and the refusal did not fire. The backup is written first and holds the original bytes verbatim, so this is recoverable rather than lost — which is why it is MEDIUM and not HIGH.

---

## What I verified as claimed

**1. The fallback is intact, and the change is safer than before.** `SessionManager::list` still serves `cached_all()` on every store error (`manager.rs:353-366`). No input reports a new empty or partial fleet through that path — the opposite: on `origin/main` a non-`NotFound` read error made `read_file` return `Ok(empty)`, which both returned an empty fleet *and* replaced the in-memory map with an empty one, so the cached fallback had nothing left to serve. The one remaining clean-empty listing is finding 7, and it predates this PR.

`tm ls --json` is safe. The body goes to stdout (`managed.rs:281`, `println!("{raw}")`), the banner to stderr (`session_picker.rs:186`). Correct today — untested, which is finding 5.

**2. The repair's destructiveness.** `--dry-run` returns at `repair_session_store.rs:81-84` before `repair()` is called, so no backup either; `repair_session_store_dry_run_writes_nothing` pins the directory to exactly `sessions.json`. Degenerate inputs all route to the safe refusal:

```
PROBE-C empty      -> holds no complete JSON document in its 0 byte(s) … restore from a backup
PROBE-C whitespace -> … 6 byte(s) …
PROBE-C garbage    -> … 12 byte(s) …
PROBE-C two-docs   -> plan Ok prefix=Some(15) sessions=0 orphans=[]
```

A tail repeating an id that is in the head with different content is correctly not an orphan — the head's copy survives, and discarding the tail's is what truncation means. `valid_prefix_bytes` can never be `Some(0)`, so the file can never be truncated to nothing.

**3. The staging fix, judged independently of the hypothesis.** `staging_path` (`store.rs:151`) is `<name>.tmp.<pid>.<uuid-v4>`: distinct across processes (pid), across instances in one process (uuid), and across restarts (uuid, not a counter). Correct whether or not the shared `sessions.json.tmp` caused the 2026-08-06 file. `two_stores_over_one_path_do_not_share_a_staging_file` reddens when I strip the uuid. Cleanup is finding 8.

**4. The second data-loss path was real, and worse than stated.** On `origin/main`:

```rust
Err(_) => {
    debug!(path = %path.display(), "no session store file found; starting fresh");
    Ok((StoredData::default(), None))
}
```

Beyond the `save()`-writes-emptiness path the PR describes, this also poisoned `known_tmux_names` (`manager.rs:957-962`) — the orphan-GC's fail-closed tracked-names set. A permission error made that set **empty**, so every live tmux session looked untracked and reapable. The fix closes that too, and it deserves the changelog line it did not get.

**5. The diagnostic is truthful on the real shape.** `byte_offset()` does not absorb whitespace between the document and the junk, so the reported offset is the exact end of the recoverable document:

```
PROBE-D "{\"sessions\":{}}garbage"           -> prefix=Some(15)
PROBE-D "{\"sessions\":{}}\n\n\ngarbage"     -> prefix=Some(15)
PROBE-D "{\"sessions\":{}}   {\"sessions\":{}}" -> prefix=Some(15)
```

It misleads only for the class in finding 2.

**6. Mutations — four rounds of my own.**

| Mutation | Result |
|---|---|
| `repair()`: drop the `!force` orphan refusal | `repair_refuses_an_orphaned_tail_without_force` FAILED |
| `valid_prefix_bytes`: return `Some(raw.len())` | 7 tests FAILED, incl. `load_reports_a_trailing_tail_…`, `repair_backs_up_then_truncates_…`, `session_store_check_fails_on_a_trailing_tail_…` |
| `read_file`: restore `Err(_) => Ok((default, None))` | `load_propagates_a_read_error_instead_of_reporting_an_empty_store` FAILED |
| `repair()`: write the backup after the truncation | `repair_backs_up_then_truncates_to_the_valid_document` + `repair_session_store_truncates_and_backs_up` FAILED |
| `log_reload_fallback`: swap `error!`/`warn!` | **survived** — `…_separates_corruption_from_a_transient_error … ok` |
| `fetch_managed_raw`: `eprintln!` → `println!` | **survived** — `10 passed; 0 failed` |

The four that should kill do kill, and the fixture is genuinely the incident's shape rather than generic garbage. The two survivors are findings 5 and 6: both tests name a mechanism they do not assert. The severity test asserts `is_corruption()`, which `corruption_is_distinguished_from_transient_io` already covers, then calls the function for coverage.

**7. The pre-existing test updates.** `manager_create_store_failure_reaps_orphan`'s assertions are genuinely unchanged — the diff touches the sabotage and adds a permission restore; `expect_err`, `matches!(err, ManagedError::Store(_))` and the `kill_session` assertion are context lines. The new sabotage exercises the intended sequence: `sessions.json` is absent at reload time so `read_file` takes the `NotFound` branch and succeeds, and `save`'s staging write fails EACCES under mode `0o500`. One caveat worth a comment: as root, `0o500` is not enforced and the test fails on `expect_err` — fail-closed, so acceptable. The doctor rename and the two expected-name lists are additive.

## Gates

Reproduced locally on `19772543`:

```
cargo test -p trusty-mpm --lib --bin tm
test result: FAILED. 4752 passed; 2 failed; 7 ignored; 0 measured; 0 filtered out; finished in 113.89s

failures:
```

Both pass in isolation, neither touches `session_manager`:

```
cargo test -p trusty-mpm --lib -- --exact core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning client::executor::tests::execute_doctor_against_test_daemon
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 4759 filtered out; finished in 8.69s
```

`execute_doctor_against_test_daemon` is the documented baseline flake at 8.69 s against the 10 s timeout, matching the PR's 8.46 s. `ensure_managed_config_dir_emits_the_frozen_skill_warning` is a second isolation-shaped flake on my machine, unrelated to this diff.

```
line-cap: measured 3749 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
```

CI reports no checks on the branch — the GitHub Actions outage, judged on local gates as instructed.

## Required changes

1. Make `check_session_store` (`doctor_session_store.rs:66`) validate `StoredData`, not `serde_json::Value`, so the probe cannot report `Ok` for a store the daemon rejects. Update the `session_store_check_is_ok_for_a_healthy_store` fixture — `{"sessions":{"a":{},"b":{}}}` is not a loadable store.
2. Make `plan()` (`store_integrity.rs:235`) agree with the store about what "corrupt" means: a distinct error for valid JSON that is not a valid store, and no `tm repair session-store` advice in `diagnostic()` when `trailing_bytes() == 0`.
3. Widen the orphan scan by `UUID_LEN - 1` bytes (`store_integrity.rs:246`) so an id straddling the cut is seen.
4. Truncate the bytes `plan` actually measured (`store_integrity.rs:283-295`), or re-validate `kept` against the re-read string before slicing.
5. Assert the banner's stderr routing and `log_reload_fallback`'s severity split, or drop the two test names that promise them.

## Notes

- Finding 7 is pre-existing — `origin/main` already failed `SessionStore::load` on a parse error, so a daemon restarted against a corrupt store already fell back to an empty `$TMPDIR` store. It logs at ERROR and `tm doctor` catches it, which is why it is MEDIUM and `Parent` rather than blocking. It is listed because the PR's abstract claims `tm ls` visibility that does not hold for that sequence.
- `run_doctor` now reads the real `~/.trusty-mpm/session-manager/sessions.json` during `run_doctor_produces_thirty_checks` and `doctor_endpoint_returns_report`. Read-only, and consistent with the sibling probes, so not flagged — noting it so nobody is surprised later.
- Not flagged: the store still has no cross-process lock, so two savers can lose an update even with private staging files. Out of scope; the PR does not claim to fix it.

bobmatnyc:
## Verdict: WARN

Re-review of `d0c1419e` against the spec and the diff. Throwaway worktree off the PR head plus a second one off the merge base `8ab2c147`; the owner's live store was never read or written.

All five required changes from the previous round landed and hold under my own mutation testing. Both HIGH findings are closed. What keeps this off APPROVE is a gate the diff turns red: the PR adds a doctor check and a CLI subcommand, both of which feed the generated `tm-capabilities` catalog, and the catalog was not regenerated. `scripts/check_capabilities.sh` passes on the merge base and fails on this head. One command fixes it.

---

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/assets/skills/tm-capabilities/references/doctor.md` | 5 | `capabilities-drift.yml` goes red: the file still says "29 checks" and has no `session_store` row. Siblings `references/cli.md:102` (no `session-store` under `repair`) and `tm-capabilities.md:44` ("29 checks") drift with it. The shipped skill therefore omits both the new check and `tm repair session-store` — the recovery command this PR exists to add is invisible to every agent that consults the catalog. | `cargo run -p trusty-mpm --bin tm -- generate capabilities`, commit the 3-file diff. | Fix here |
| LOW | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 321-326 | `read()` reports a non-UTF-8 store as `RepairError::Io` — "session store repair I/O error … stream did not contain valid UTF-8" — which reads transient for a permanent condition. `read_file` (`store.rs:329-342`) calls the same file `Corrupt`; `check_session_store` (`doctor_session_store.rs:59-65`) calls it `Unknown`. Three classifications for one file, in the PR that unified the other three. | Hoist the synthesized `StoreIntegrity` already in `store.rs:329-342` into `store_integrity`, and route `read()` and the doctor's non-`NotFound` arm through it. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/daemon/state/core.rs` | 748-763 | Pre-existing, untouched by this PR (`git diff 8ab2c147..d0c1419e -- …/state/core.rs` is empty). A store corrupt at daemon boot substitutes an empty `$TMPDIR` store; `store_health` is then `None` and the list endpoint discloses nothing. | Record the boot failure on the fallback manager and serve it through the same `store_health` field. | Parent |
| LOW | `crates/trusty-mpm/src/session_manager/store.rs` | 157-165, 428-435 | Per-instance staging names accumulate without a reaper; a crash between `fs::write(tmp)` and `fs::rename` orphans one store-sized file per instance, forever. | Unlink `sessions.json.tmp.*` siblings older than an hour on `load`. | Promote |
| LOW | `crates/trusty-mpm/src/session_manager/store_integrity.rs` | 401 | `repair` truncates in place (`fs::write(path, &raw[..kept])`) while `save()` was just changed to stage-and-rename so a cross-process reader never sees a torn file. | Stage to `<path>.repair.<pid>` and rename. | Promote |

---

## The four deferrals, judged

**`sessions.json.tmp.*` reaper (store.rs) — defensible.** The consequence is disk litter, bounded to one file per store instance per crash, in a directory nothing else globs. `check_session_store` reads only `sessions.json`. No correctness or disclosure path touches it. Separable janitorial work.

**In-place truncation in `repair` (store_integrity.rs:401) — defensible, on one specific ground.** The asymmetry against `save()`'s new stage-and-rename is exactly the sibling-branch tell I would normally escalate on. What defuses it is ordering: the backup is written first and unconditionally (`store_integrity.rs:392-396`), so the worst crash outcome is a partial `sessions.json` next to a complete `.corrupt-backup-*`, and the command prints the backup path. The file is also already unwritable by the daemon at that point, so the concurrent-reader exposure `save()` protects against is not live here. Recoverable, not lost. Fix it when convenient.

**`daemon/state/core.rs:748-763` — defensible, and more so than at the last review.** It is genuinely pre-existing and this PR does not touch the file. More to the point, the new `session_store` doctor check reads `sessions.json` directly rather than asking the daemon, so it now catches precisely this state — the operator who was previously blind here has a probe. The remaining gap is that `tm ls` alone stays quiet on this one path.

**Non-UTF-8 classification — defensible as a merge blocker, not as a final state.** No surface reports health: the doctor says `Unknown`, never `Ok`. The one line that reads wrong is `plan`'s, and it is a message defect, not a correctness one. I have kept it as a LOW `Fix here` rather than dropping it because it is the last residual of this PR's own thesis, and the synthesized `StoreIntegrity` it needs already exists ten lines away.

---

## Verifying the six claims

**1. One validation entry point — holds, and no fourth surface was missed.** `store_integrity::validate` is called from `store.rs:349-350`, `doctor_session_store.rs:72`, and `plan_bytes` (`store_integrity.rs:338`, and again at 351 for the recovered head). Every production reader of the store in `trusty-mpm` reaches it: `manager.rs:227`, `agent_reset_workspace.rs:168`, `supervisor.rs:95`, `meta/launch.rs:148`, `state/core.rs:748` all go through `SessionStore::load` → `read_file` → `validate`, and the two CLI paths propagate the error rather than substituting a default.

I walked the input classes for divergence:

| Input | store | doctor | plan |
|---|---|---|---|
| loadable store | Ok | Ok + count | `NotCorrupt` |
| valid JSON, wrong schema | Corrupt | Fail | `NotAStore` |
| document + stale tail | Corrupt | Fail | plan with the cut |
| head not a document | Corrupt | Fail | `NoValidPrefix` |
| empty / whitespace | Corrupt | Fail | `NoValidPrefix` |
| **non-UTF-8** | **Corrupt** | **Unknown** | **`Io`** |

Only the last row diverges, and no arm of it claims health. That is the LOW above.

One reader outside this crate: `trusty-agents/src/tools/session_state/store.rs:139` deserializes the same file through its own maximally-permissive `SessionView`. Not flagged — `trusty-agents` deliberately declines the `trusty-mpm` dependency (documented at that module's head), it renders a list rather than a verdict, and it is outside this PR's crate.

**2. `repairable_by_truncation` — holds; the phantom-tail report is gone.** `diagnostic_does_not_advertise_a_repair_that_would_refuse` asserts both halves. The field cannot go stale relative to what repair would do, because both compute the same expression against the same bytes: `analyze` at `store_integrity.rs:172-173` runs `serde_json::from_str::<StoredData>(&raw[..v]).is_ok()`, and `plan_bytes` at 351 runs `validate(path, &raw[..valid])`. Hardcoding the field to `true` kills the test:

```
thread '…diagnostic_does_not_advertise_a_repair_that_would_refuse' panicked
test result: FAILED. 25 passed; 1 failed; 0 ignored; 4740 filtered out
```

**3. `plan` refuses a head that is not a store — holds.** Replacing the `NotAStore` gate with `unwrap_or_default()`:

```
    session_manager::store_integrity::store_integrity_tests::plan_and_the_store_agree_that_valid_json_is_not_a_store
    session_manager::store_integrity::store_integrity_tests::plan_refuses_to_cut_down_to_a_head_that_is_not_a_store
test result: FAILED. 24 passed; 2 failed; 0 ignored; 4740 filtered out
```

**4. `repair` reads once — holds; both defects are gone.** `repair` (`store_integrity.rs:382-404`) calls `read(path)` once, passes those bytes to `plan_bytes`, writes the backup from the same `raw`, and slices `&raw[..kept]` where `kept` was measured from that same `raw`. There is no second read to disagree with, so neither the out-of-range panic nor the misplaced cut has an input that produces it. `unwrap_or(raw.len())` at line 400 is unreachable — `plan_bytes` returns `NoValidPrefix` when the option is `None` — and safe if it were reached.

Residual, below the bar to flag: `repair_session_store` (`repair_session_store.rs:53, 85`) calls `plan` and then `repair`, so the CLI reads twice. Harmless — `repair` re-plans from its own read and cannot cut bytes it did not measure, and the target file is by construction one the daemon cannot rewrite.

**5. The straddle case is genuinely unreachable through `plan` — verified, not accepted.** I probed `byte_offset()` directly rather than trusting the reasoning:

```
no-ws   -> Some(15) (doc.len=15)
ws+junk -> Some(15)
ws only -> Some(15)
nl+doc  -> Some(15)
store cut byte = "}"
```

`byte_offset()` does not absorb trailing whitespace, so `raw[..valid]` ends at the closing `}` of the first value; the schema gate at line 351 then forces that value to be a JSON object. `raw[valid-1]` is therefore always `}`, which is neither a hex digit nor `-`, so no canonical UUID can span the cut. The claim is correct and `orphan_scan_sees_an_id_that_straddles_the_cut` is defense-in-depth at the helper, not load-bearing coverage of a live path.

I also chased the inverse risk — whether the 35-byte backscan can now invent a false orphan out of the kept document — and it cannot. For `uuid_tokens` to match, a whole 36-byte window must fit inside `raw[valid-35..]`, which requires it to end at `valid+1` or later. An id lying entirely in the head is always clipped. Confirmed empirically with a compact store whose last field before the closing braces is a non-key UUID and a tail containing no UUID at all: `orphan_ids = []`. No finding.

**6. Mutants — four of my own, applied one at a time, including both previous survivors.**

| Mutation | Result |
|---|---|
| `log_reload_fallback`: swap `error!` / `warn!` | **killed** — `…_separates_corruption_from_a_transient_error` FAILED at `store_integrity_tests.rs:275`, quoting the captured `[WARN …] session list: store is CORRUPT …` line |
| `warn_if_store_degraded`: `eprintln!` → `println!` | **killed** — `store_banner_goes_to_stderr_so_json_stays_machine_readable` FAILED at `session_picker_tests.rs:469` (the "must NEVER reach stdout" assertion); the same test passes unmutated |
| `analyze`: hardcode `repairable_by_truncation: true` | **killed** (above) |
| `plan_bytes`: drop the `NotAStore` gate | **killed** (above) |

Neither unusual mechanism passes vacuously. The stream test re-runs itself as a child and guards on `stdout.contains("1 passed")`, so a name-resolution failure fails the parent rather than silently asserting over an empty stdout — I confirmed the guard by running it clean first, then under the mutation. The severity test reads the level off the rendered `LogBufferLayer` line, not off `is_corruption()`.

---

## `session_store` in `DOCTOR_CHECKS`, and what it dragged in

The engineer's account is correct and I reproduced it. Deleting the `session_store` entry from `DOCTOR_CHECKS` and running the drift guard:

```
assertion `left == right` failed: generate::doctor::DOCTOR_CHECKS has drifted from run_doctor's actual checks
  left:  [… "binary_provenance", "session_store"]
  right: [… "binary_provenance"]
test result: FAILED. 0 passed; 1 failed; 0 ignored; 1529 filtered out
```

The test lives in the `tm` **bin** target, which is why `--lib` never surfaced it. Fixed, and the appended position matches `run_doctor`'s real execution order.

What it dragged in is the HIGH. `DOCTOR_CHECKS` and the clap tree both feed `tm generate capabilities`, and the regenerated output was not committed:

```
$ bash scripts/check_capabilities.sh          # on d0c1419e
tm-capabilities: drift detected in 3 file(s):
  - tm-capabilities.md (content differs)
  - tm-capabilities/references/cli.md (content differs)
  - tm-capabilities/references/doctor.md (content differs)
Error: tm-capabilities drift check failed

$ bash scripts/check_capabilities.sh          # on the merge base 8ab2c147
tm-capabilities: up to date (7 generated files).
```

The base is clean, so the drift is entirely this PR's. Regenerating produces exactly three lines of substance — `29 checks` → `30 checks` in two places, the `session_store` row, and `- session-store — Recover a corrupt managed-session store (sessions.json)` under `repair` in `cli.md`. I reverted my regeneration; the worktree is clean.

---

## Other things I looked at

`repair_session_store_tests.rs`'s `corrupt_store` now builds a genuinely loadable head (`RECORD` carries all six no-default fields) and the assertion moved from `serde_json::from_str::<Value>` to `SessionStore::load`. That is the right direction: with the new `NotAStore` gate the old fixture would have made the repair refuse, so the test would have failed loudly rather than silently proving less — but the stronger assertion is what the test's own name claims.

`loadable_store_json` (`store_integrity.rs:478`) as the single source of "healthy store" fixtures is the correct structural answer to how the doctor test came to assert health for an unloadable file. `session_store_check_is_ok_for_a_healthy_store` now asserts the fixture invariant with a real `SessionStore::load` before exercising the probe.

SLOC: no pressure. `store_integrity.rs` is 527 raw lines but heavily doc-commented; `check_line_cap.sh` reports 0 violations with none of the touched files newly allowlisted.

No test in the diff asserts implementation shape over behavior that I could find. The closest is `two_stores_over_one_path_do_not_share_a_staging_file`, which asserts path inequality — but the invariant genuinely is about the paths, and it was already there.

---

## Gates

```
cargo check -p trusty-mpm --all-targets                  EXIT=0
bash scripts/check_line_cap.sh                           EXIT=0
  line-cap: measured 3749 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
  sld-lint: scanned 56 spec doc(s) + 3119 code file(s); 0 error(s), 0 warning(s)
bash scripts/check_changelog_fragment.sh                 EXIT=0
bash scripts/check_capabilities.sh                       EXIT=1   ← finding 1
```

Full crate surface, not `--lib`:

```
cargo test -p trusty-mpm --bins --tests
test result: ok. 4759 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 60.01s
test result: ok. 1529 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 4.05s
… 43 further targets, all `ok`, 0 failed
EXIT=0
```

Three earlier `cargo test -p trusty-mpm` runs were not clean, and all three failures classify as pre-existing:

```
run 1: 4757 passed; 2 failed
  daemon::managed_routes::tests::stale_assets_for_many_sees_an_agent_change_on_the_very_next_call
run 2: 4758 passed; 1 failed
  daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet
run 3: 4758 passed; 1 failed   (same test as run 2)
```

Both are process-global-state isolation flakes, and I measured rather than asserted that:

```
merge base 8ab2c147, cargo test -p trusty-mpm --lib
test result: FAILED. 4724 passed; 1 failed; 7 ignored
  daemon::managed_routes::tests::stale_assets_for_many_sees_an_agent_change_on_the_very_next_call
```

The `stale_assets_for_many_*` family fails on the merge base with none of this PR's code, and the diff's only touch to that module is the additive `store_health` field on `ListSessionsResponse`. For the tracing-capture test — the one whose mechanism the new severity test copies — I ran 20 iterations of the filtered suite at 8 threads on both trees:

```
head d0c1419e:  failures=7/20
base 8ab2c147:  failures=6/20
```

Statistically identical, so the new `#[serial]` subscriber test is not the perturbation. Both failures pass in isolation:

```
cargo test -p trusty-mpm --lib -- --exact core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning
test result: ok. 1 passed; 0 failed; 0 ignored; 4765 filtered out
cargo test -p trusty-mpm --lib -- --exact daemon::managed_routes::tests::stale_assets_for_many_sees_an_agent_change_on_the_very_next_call
test result: ok. 1 passed; 0 failed; 0 ignored; 4765 filtered out
```

Every `session_manager`, `store_integrity`, `doctor_session_store`, `repair_session_store` and `session_picker` test passed in every run. `execute_doctor_against_test_daemon` did not fail this time. Nothing failed on a refused local connection, so the stopped Ollama did not come into it. CI reports no checks — the GitHub Actions outage, as stated.

---

## Required changes

1. `cargo run -p trusty-mpm --bin tm -- generate capabilities`, then commit the three-file diff under `crates/trusty-mpm/src/assets/skills/`. Re-run `bash scripts/check_capabilities.sh` and confirm `up to date (7 generated files)`.

Optional in the same PR: route the non-UTF-8 case through the synthesized `StoreIntegrity` so `plan` stops calling a permanent condition an I/O error.

---

## Notes

- The two mutation-killed tests are the only ones I checked for vacuity; the rest of the suite I took at face value, as before.
- I did not re-audit the parts of the diff that were unchanged since `19772543` beyond spot checks — the fallback semantics, the staging-name fix, and the `read_file` data-loss closure were verified in the previous round and are untouched here.
- Not flagged, unchanged from last time: the store still has no cross-process lock, so two savers can lose an update even with private staging files. Out of scope.

Issue #2846, opened 2026-07-16
[trusty-search] Memory limit (rss_limit_mb) not enforced — unbounded RSS growth leads to OOM-kill


bob-duetto:
**Component:** trusty-search

**Problem**
The daemon reports a self-declared `rss_limit_mb` (observed default 12000 / 12GB) but does NOT appear to enforce it against actual process RSS. In production on 2026-07-16, a serving instance's trusty-search RSS grew to **~26.6GB — 2.2x its own 12GB stated limit** — over ~20 days of uptime with no periodic reclaim. The growth is almost entirely anonymous heap (`RssAnon` ~99.97%, not reclaimable page cache) and monotonic with uptime, consistent with a slow leak or unbounded cache growth rather than normal working set.

**Impact**
On a 32GB host with no swap, this eventually exhausted host RAM (down to ~830MB free), then during a large PR-review context fan-out the process tripped the concurrency limiter (503 `server_busy` storm) and was **kernel-OOM-killed**, taking down `main` search for hours. A sibling instance that restarts trusty-search daily (~08:05 UTC) never accumulates this growth, which further points to a leak masked only by periodic restarts.

**Requested**
1. Investigate the unbounded RSS / heap growth (allocator/heap profiling — e.g. jemalloc stats — to characterize whether it's a leak vs. an unbounded cache).
2. Actually ENFORCE the configured memory limit (`rss_limit_mb`) — e.g. trigger reclaim / cache eviction / backpressure as RSS approaches the limit, so the daemon self-caps instead of growing until the OS OOM-killer intervenes.
3. Consider surfacing the current RSS-vs-limit in `/health` telemetry so operators can see approaching-limit conditions before an OOM.

**Workaround currently in place (ops-side, not a fix)**
Operators added a systemd `MemoryHigh`/`MemoryMax` cgroup cap (24G/28G) + swap to the serving hosts, and restart to reclaim — but the daemon should enforce its own limit rather than relying on the OS.

**Context**
Reported from the duettoresearch/code-intelligence deployment.

Issue #2845, opened 2026-07-16
[trusty-search] Add a default max fan-out count and a --serial flag for multi-index search fan-out


bob-duetto:
**Component:** trusty-search

**Problem / motivation**
A single search request that fans out across many per-repo indexes issues all of the per-index queries near-simultaneously. When the fan-out is large (observed: a PR-review context search fanning out across ~150+ repo indexes at once), it overruns trusty-search's internal admission/concurrency limiter, which returns `concurrency limiter: queue full, returning 503` in a sub-millisecond burst. Downstream, the caller gets zero results for the whole request.

This was hit in production on 2026-07-16: thousands of `server_busy`/503s per burst (peak ~8,000/min), causing multiple PR reviews to skip with "no context." There is currently **no exposed configuration** to bound the fan-out concurrency or to force serial execution (checked `trusty-search --help` / `start --help` / config file — no concurrency/queue-depth/max-inflight setting exists).

**Requested changes**
1. **Default max fan-out count** — a bounded, configurable cap on the number of per-index queries executed concurrently within a single fan-out search (e.g. a semaphore / bounded worker pool). Ship a sane default so a large fan-out degrades gracefully (bounded concurrency, results still returned) instead of tripping the admission limiter and 503-ing the whole request. Make the cap configurable (flag/env/config).
2. **A `--serial` flag** — an option to process a fan-out's per-index queries serially (one at a time / strictly bounded) rather than concurrently. A safety valve that trades latency for guaranteed non-overload — useful on memory/CPU-constrained hosts or when the concurrency limiter is being tripped.

**Impact / severity**
Directly caused user-visible search-context failures (PR reviews returning "no context") during a production incident. A bounded fan-out and/or `--serial` mode would have prevented it.

**Context**
Reported from the duettoresearch/code-intelligence deployment (serving instances query trusty-search over HTTP; the PR-review pipeline is the large-fan-out caller).

Issue #4637, opened 2026-08-03
trusty-memory: HTTP handlers force-open every palace (O(n) cold disk I/O) — /api/v1/status and /api/v1/palaces unusable at 5,794 palaces


bobmatnyc:
## Problem

The live `trusty-memory` daemon has **5,794 palaces on disk**. Four HTTP handler paths loop over *every* palace calling `PalaceRegistry::open_palace()` — a synchronous function that does blocking disk I/O per palace (usearch vector index load, KG redb open, full drawer-table load + prune, recall-log redb open) — inline on the async axum executor, with no `spawn_blocking`, no timeout, and no pagination.

The LRU open-handle cache caps at `DEFAULT_MAX_OPEN_PALACES = 64` (`crates/trusty-common/src/memory_core/registry.rs:70`), so **~98.9% of palaces are cold opens** at ~0.9–1.1s each. That is **~87–106 minutes of blocking disk I/O per request**.

### Measured live (daemon PID 29856, launchd `com.trusty.memory`)

| Route | Observed |
|---|---|
| `GET /api/v1/status` | did **not** respond within 90s |
| `GET /api/v1/palaces` | same shape (full-registry open) |
| `GET /health` | 36 ms (does not touch the registry) |
| `GET /api/v1/palaces/{id}` | ~0.9 s (single cold open — correct cost) |

## Affected sites

- `crates/trusty-memory/src/service/core.rs:151-167` — `list_palaces()`, backs `GET /api/v1/palaces`. `open_palace` at ~line 162.
- `crates/trusty-memory/src/service/core.rs:86-103` → `crates/trusty-memory/src/service/helpers.rs:164-182` `collect_palace_stats`, backs `GET /api/v1/status`. `open_palace` at ~`helpers.rs:171`.
- `crates/trusty-memory/src/service/core.rs:648-665` — `recall_all()`, backs `GET /api/v1/recall`.
- `crates/trusty-memory/src/service/core_kg.rs:184-199` — `dream_run()`, backs `POST /api/v1/dream/run`.

Same-shape twins on the chat/MCP surfaces:
- `crates/trusty-memory/src/chat/tools.rs:280` `execute_list_palaces`
- `crates/trusty-memory/src/chat/tools.rs:423` `execute_get_status`
- `crates/trusty-memory/src/chat/tools.rs:344` `execute_recall_all`
- `crates/trusty-memory/src/tools/memory_ops.rs:455` `handle_memory_recall_all`

[...]

PR #5048, opened 2026-08-07
feat(bm25): lossless backfill + bounded daemon process model + stats op

bobmatnyc:
## Defect

The BM25 lexical lane has never run. `bm25_search_optional` returns on its first line unless `TRUSTY_BM25_DAEMON=1` was set at daemon startup, and no plist, installer, or shipped path sets it. There are zero BM25 index directories across ~99 palaces, so the fusion code in `handle_memory_recall` has never executed.

Turning it on, as-is, would not have produced a working lexical lane. Three things stood in the way.

**1. The write path drops silently under load.** `tools/bm25.rs:82` and `:189-198` — a 256-slot mpsc written with `try_send`, dropping on full with a `warn!`. trusty-tools holds 1311 drawers, five times the queue. A backfill reusing `bm25_index_enqueue` would have dropped roughly 80% of the corpus and left the palace answering lexical queries from a fifth of its content — indistinguishable from working fusion.

**2. `Bm25Supervisor` never evicted.** Only `shutdown` removed a map entry. One cross-palace `recall_all` over ~99 palaces left 99 child processes resident for the trusty-memory daemon's whole life, each one's memory scaling with its palace's drawer text (`PalaceBm25Index` retains every document's full text in a `BTreeMap`).

**3. Nothing could ask whether a palace was fully indexed.** The protocol had `index` and `search` only, so an empty hit list was ambiguous between "no match" and "nothing indexed" — the fail-open shape this whole change exists to remove.

## Resolution

**The write path is left exactly as it is.** Drop-on-full is defensible there: the drawer is durable in redb, `memory_remember` must not wait on daemon RTT, and a dropped index op now self-heals because backfill detects the coverage shortfall via `stats` and repairs it. Widening the queue would only move the cliff.

**Backfill gets its own feeder** (`bm25_backfill.rs`) that awaits each document's ack, so nothing is ever offered to a full queue. Idempotent (`upsert_document` is keyed by `doc_id`), bounded by a 10 s per-op and 120 s per-palace deadline, and it reads coverage back from the daemon rather than trusting its own submission count — `fully_indexed()` returns `false` for a `Completed` run whose read-back failed. Runs as a serial startup sweep over palaces that have drawers; `TRUSTY_BM25_NO_BACKFILL=1` defers it.

**The supervisor bounds its population** (#2845, #2846). A cap with LRU reaping (`TRUSTY_BM25_MAX_DAEMONS`, default 3) and a per-daemon RSS ceiling compared against a real measurement rather than merely declared (`TRUSTY_BM25_RSS_LIMIT_MB`, default 512, `0` disables). Spawns are serialised so a burst fan-out cannot satisfy the cap per-caller and violate it in aggregate. An unmeasurable RSS never reaps — `None` means "no reading", not "zero", and reaping on it would turn a memory guardrail into an outage. `process_rss_mb` lands in `trusty-common::sys_metrics` so the next supervisor that needs it does not grow its own `/proc` parser.

**A `stats` op** reports `doc_count` and `total_text_bytes`, answered inline by the batch worker so it reflects un-flushed writes.

**Plus a durability defect the cap surfaced.** The daemon exited on SIGTERM without flushing: the worker's final flush is gated on the op channel closing, and the accept loop holds a sender for the process lifetime, so up to one write window was lost on every exit. Latent while daemons only died at shutdown; routine once the supervisor reaps them. `a_reaped_palace_respawns_on_next_use` caught it and now pins the fix.

## Fail-open evidence

Each arm degrades to a reported status with a bounded wait — never an error into a caller's path, never a hang.

| Arm | Test |
|---|---|
| Daemon absent | `backfill_reports_daemon_unavailable_when_socket_is_dead` (also asserts it fails fast, not by waiting out the deadline) |
| Daemon spawn fails / lane off | `backfill_state_palace_is_disabled_without_a_client` |
| Socket accepts but never answers | `backfill_gives_up_on_a_socket_that_never_answers` |
| Index empty | `empty_palace_is_already_indexed`, `batch_queue_stats_go_to_zero_after_rebuild` |
| Index partial | `partial_index_is_detected_and_repaired`, `fully_indexed_requires_a_read_back_count` |
| Queue saturated | `backfill_indexes_every_drawer_without_drops` — 1024 docs, 4× the write path's queue; a drop-on-full feeder lands ~256 |
| Unbounded fan-out | `daemon_population_stays_within_the_cap`, `exceeding_the_cap_reaps_the_least_recently_used` |
| Unenforced RSS limit | `over_rss_limit_children_are_reaped`, `under_rss_limit_children_are_left_alone`, `over_rss_limit_is_false_without_a_measurement` |

## Gates — rung 5

```
cargo fmt --check                                          — clean
cargo test -p trusty-bm25-daemon                           — 41 passed, 0 failed (unit) + 3 passed (integration)
cargo test -p trusty-common --features memory-core         — 892 passed, 0 failed, 13 ignored (+ 21 across 3 integration targets)
cargo test -p trusty-memory                                — 581 passed, 0 failed, 4 ignored (lib) + all integration targets green
cargo test -p trusty-memory --test bm25_backfill_e2e -- --include-ignored
                                                           — 4 passed, 0 failed
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui — clean
cargo clippy -p trusty-bm25-daemon --all-targets -- -D warnings — clean
scripts/check_line_cap.sh                                  — 3746 files, 0 violations
scripts/check_changelog_fragment.sh                        — 3/3 crates recorded
scripts/check_sld.sh                                       — 0 errors, 0 warnings
```

Two gates were scoped, both with proof rather than by narrowing coverage:

- **`cargo check --workspace`** fails on `trusty-code-gui` for an unbuilt `ui/dist` in a fresh worktree. `git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/ crates/trusty-mpm-gui/` lists **0 paths**.
- **`cargo clippy -p trusty-memory -- -D warnings`** is blocked by a pre-existing `nonminimal_bool` lint in `trusty-common/src/memory_core/retrieval/embed_repair.rs:139`, which builds as a dependency. `git diff origin/main -- <that file>` is **empty** — the file is byte-identical to main. With that one lint allowed, `cargo clippy -p trusty-memory --all-targets -- -D warnings -A clippy::nonminimal_bool` is clean, so nothing in this diff introduces a lint.

## Measured

- **Backfill throughput:** 1024 documents in **84 ms** against a real daemon. The whole ~2300-drawer corpus is a few seconds.
- **Steady-state daemon count: 3** (the default cap), regardless of how many palaces are touched. `daemon_population_stays_within_the_cap` drives 5 palaces through a cap of 2 and asserts the population never exceeds it, ends at exactly 2, and reaped exactly 3.

## Scope

Piece 1 of 3. **The lane is still off** — this PR does not flip `TRUSTY_BM25_DAEMON`'s default, and `backfill_state_palace_is_disabled_without_a_client` asserts that. Not incomplete; the flip is piece 2, along with hook routing to `/rpc` and `handle_memory_recall_deep`'s missing fusion.

Keeping `TRUSTY_BM25_DAEMON` as the switch rather than adding a config key: agreed after reading the code. It is a trusty-memory daemon-lifetime setting and does not belong in trusty-mpm's `config.toml`. The two new knobs follow the same convention.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11320,7 +11320,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-bm25-daemon"
-version = "0.1.3"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,7 @@ trusty-embedderd = { path = "crates/trusty-embedderd", version = "0.3.10" }
 # trusty-bm25-daemon: per-palace BM25 lexical-index daemon — referenced by
 # trusty-memory so `cargo install trusty-memory` produces both binaries from
 # one command (mirrors the trusty-embedderd / trusty-search pattern, PR #190).
-trusty-bm25-daemon = { path = "crates/trusty-bm25-daemon", version = "0.1.3" }
+trusty-bm25-daemon = { path = "crates/trusty-bm25-daemon", version = "0.2.0" }
 # tga: trusty-git-analytics — developer productivity analytics.
 # Referenced by trusty-review for contributor-profile data pipeline (#558).
 tga = { path = "crates/trusty-git-analytics", version = "2.10" }
--- a/crates/trusty-bm25-daemon/Cargo.toml
+++ b/crates/trusty-bm25-daemon/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trusty-bm25-daemon"
-version = "0.1.3"
+version = "0.2.0"
 publish = true
 edition = "2021"
 rust-version = "1.94"
--- a/crates/trusty-bm25-daemon/changelog.d/4995-shutdown-flush-fixed.md
+++ b/crates/trusty-bm25-daemon/changelog.d/4995-shutdown-flush-fixed.md
@@ -1,0 +1,9 @@
+Fixed
+
+- The daemon now flushes its snapshot before exiting on SIGTERM/SIGINT.
+  Previously it removed the socket and returned with the batch worker still
+  holding up to one write window of applied-but-unwritten documents, which were
+  lost: the worker's final flush is gated on the op channel closing, and the
+  accept loop holds a sender for the process lifetime. Latent while a daemon
+  only ever died at shutdown; routine now that trusty-memory reaps daemons to
+  hold its process cap.
--- a/crates/trusty-bm25-daemon/changelog.d/4995-stats-op-and-shutdown-flush.md
+++ b/crates/trusty-bm25-daemon/changelog.d/4995-stats-op-and-shutdown-flush.md
@@ -1,0 +1,6 @@
+Added
+
+- `stats` JSON-RPC method, reporting `doc_count` and `total_text_bytes`. The
+  protocol could say "here are your hits" but not "is there anything to hit",
+  so a caller could not tell an indexed palace with no match from an unindexed
+  one.
--- a/crates/trusty-bm25-daemon/changelog.d/5048-missing-docs-and-testable-shutdown.md
+++ b/crates/trusty-bm25-daemon/changelog.d/5048-missing-docs-and-testable-shutdown.md
@@ -1,0 +1,12 @@
+Added
+
+- `missing_docs` JSON-RPC method: given a list of doc ids, returns the subset the
+  daemon does not hold. Callers can now establish index coverage as a set
+  statement instead of inferring it from `stats.doc_count`, which is satisfied by
+  documents the caller never asked about.
+- `run_until(config, shutdown)` runs the daemon against an explicit shutdown
+  trigger. `run` supplies the real SIGTERM/SIGINT trigger and is otherwise
+  unchanged; the split gives the shutdown snapshot flush a deterministic,
+  CI-runnable regression test instead of a timing-dependent `#[ignore]`d one.
+  Signal handlers are now installed before snapshot load and socket bind, so a
+  SIGTERM during startup reaches the shutdown path.
--- a/crates/trusty-bm25-daemon/src/batch_queue.rs
+++ b/crates/trusty-bm25-daemon/src/batch_queue.rs
@@ -25,7 +25,7 @@ use tokio::sync::{mpsc, oneshot};
 use tokio::time::sleep;
 
 use crate::index::PalaceBm25Index;
-use crate::protocol::SearchHit;
+use crate::protocol::{SearchHit, StatsResult};
 
 /// Default write-coalescing window — how long the worker waits to coalesce
 /// more write ops after seeing the first one.
@@ -71,6 +71,12 @@ enum Op {
         query: String,
         top_k: usize,
         reply: oneshot::Sender<Result<Vec<SearchHit>>>,
+    },
+    Stats {
+        reply: oneshot::Sender<Result<StatsResult>>,
+    },
+    Flush {
+        reply: oneshot::Sender<Result<usize>>,
     },
 }
 
@@ -188,6 +194,52 @@ impl BatchQueue {
             .map_err(|_| anyhow!("bm25 batch worker dropped reply channel"))?
     }
 
+    /// Force the snapshot to disk and return the document count written.
+    ///
+    /// Why: the worker flushes at the end of each write batch, so a process
+    /// that exits mid-window loses that window's writes. That was survivable
+    /// while SIGTERM only arrived at daemon shutdown; it is not survivable now
+    /// that the supervisor reaps daemons routinely to stay within its cap. A
+    /// reap that silently discarded the last batch would make backfill
+    /// coverage depend on how recently a palace was evicted — the exact class
+    /// of invisible partial state this work exists to remove.
+    /// What: sends `Op::Flush`; the worker runs `index.flush()` inline (like
+    /// `Search` and `Stats`, so it never queues behind the write window) and
+    /// replies with the live document count. A no-op when nothing is dirty.
+    /// Test: `batch_queue_flush_persists_within_the_write_window`.
+    pub async fn flush_now(&self) -> Result<usize> {
+        let (reply_tx, reply_rx) = oneshot::channel();
+        self.tx
+            .send(Op::Flush { reply: reply_tx })
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker channel closed"))?;
+        reply_rx
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker dropped reply channel"))?
+    }
+
+    /// Report the live corpus size.
+    ///
+    /// Why: a backfiller needs a cheap, authoritative answer to "how much of
+    /// this palace does the daemon already hold" before deciding to skip,
+    /// repair, or run in full — and after the run, to confirm what landed.
+    /// Reading it through the queue (rather than a side channel) means the
+    /// answer reflects every write applied so far, including ones still
+    /// inside an unflushed batch.
+    /// What: sends `Op::Stats`; the worker answers it inline, like `Search`,
+    /// so it never waits out the write-coalescing window.
+    /// Test: `batch_queue_stats_reflect_pending_writes`.
+    pub async fn stats(&self) -> Result<StatsResult> {
+        let (reply_tx, reply_rx) = oneshot::channel();
+        self.tx
+            .send(Op::Stats { reply: reply_tx })
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker channel closed"))?;
+        reply_rx
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker dropped reply channel"))?
+    }
+
     /// Run a BM25 search against the live index.
     ///
     /// Why: search is read-only but still flows through the worker so the
@@ -257,6 +309,14 @@ async fn batch_worker(mut rx: mpsc::Receiver<Op>, mut index: PalaceBm25Index, co
                 let _ = reply.send(Ok(hits));
                 continue;
             }
+            Op::Stats { reply } => {
+                let _ = reply.send(Ok(stats_of(&index)));
+                continue;
+            }
+            Op::Flush { reply } => {
+                let _ = reply.send(flush_now(&mut index));
+                continue;
+            }
             Op::Rebuild { reply } => {
                 let new_count = index.rebuild();
                 let flush_result = index
@@ -291,6 +351,21 @@ async fn batch_worker(mut rx: mpsc::Receiver<Op>, mut index: PalaceBm25Index, co
                         // race-free.
                         let hits = index.search(&query, top_k);
                         let _ = reply.send(Ok(hits));
+                    }
+                    Some(Op::Stats { reply }) => {
+                        // Same race-freedom argument as `Search`: every write
+                        // dispatched before this op is already applied, so the
+                        // figures a backfiller reads mid-run never lag behind
+                        // the acks it has already collected.
+                        let _ = reply.send(Ok(stats_of(&index)));
+                    }
+                    Some(Op::Flush { reply }) => {
+                        // Flushing mid-batch is safe and is the whole point:
+                        // the caller is usually a shutdown or reap path that
+                        // must not lose the window's writes. The outer loop
+                        // still flushes at the window's end; `flush` is a
+                        // no-op once the dirty bit is clear.
+                        let _ = reply.send(flush_now(&mut index));
                     }
                     Some(Op::Rebuild { reply }) => {
                         // Honour the rebuild atomically. Flush the pending
@@ -344,6 +419,33 @@ async fn batch_worker(mut rx: mpsc::Receiver<Op>, mut index: PalaceBm25Index, co
     }
 }
 
+/// Snapshot the two corpus figures the `stats` op reports.
+///
+/// Why: the worker answers `Op::Stats` from two places (cold path and
+/// mid-batch), and the two must not drift into different definitions of
+/// "indexed".
+/// What: reads `doc_count` + `total_text_bytes` off the live index. Pure.
+/// Test: `batch_queue_stats_reflect_pending_writes`.
+fn stats_of(index: &PalaceBm25Index) -> StatsResult {
+    StatsResult {
+        doc_count: index.doc_count(),
+        total_text_bytes: index.total_text_bytes(),
+    }
+}
+
+/// Flush the snapshot and report the resulting document count.
+///
+/// Why: the worker answers `Op::Flush` from two places and both must report
+/// the same thing — the count that is now ON DISK, so a caller can treat a
+/// successful flush as durability and not merely as "the call returned".
+/// What: `index.flush()` then `doc_count()`. A clean index flushes nothing and
+/// still reports its count.
+/// Test: `batch_queue_flush_persists_within_the_write_window`.
+fn flush_now(index: &mut PalaceBm25Index) -> Result<usize> {
+    index.flush()?;
+    Ok(index.doc_count())
+}
+
 /// Apply one write op to the index, forwarding the typed reply.
 ///
 /// Why: factored out so the worker loop reads top-to-bottom and the per-op
@@ -368,7 +470,7 @@ fn apply_write_op(index: &mut PalaceBm25Index, op: Op) {
         }
         // Caller guarantees only write ops reach here; treat anything else
         // as a programmer error in this module.
-        Op::Search { .. } | Op::Rebuild { .. } => {
+        Op::Search { .. } | Op::Rebuild { .. } | Op::Stats { .. } | Op::Flush { .. } => {
             tracing::error!("apply_write_op called with non-write op — this is a bug");
         }
     }
@@ -444,6 +546,79 @@ mod tests {
         }
         let hits = q.search("shared".into(), 2).await.unwrap();
         assert!(hits.len() <= 2);
+    }
+
+    /// Why: `stats` is only useful to a backfiller if it reflects writes the
+    /// daemon has already acked but not yet flushed. If it read from the
+    /// snapshot on disk instead of the live index, a backfiller would see
+    /// stale counts inside the 50 ms write window and re-send work it had
+    /// already landed — or worse, declare a palace unindexed that isn't.
+    /// What: indexes three docs, immediately (inside the write window) asks
+    /// for stats, and asserts both figures already include all three.
+    /// Test: this test itself.
+    #[tokio::test]
+    async fn batch_queue_stats_reflect_pending_writes() {
+        let (q, _dir) = fresh_queue();
+        let empty = q.stats().await.unwrap();
+        assert_eq!(empty.doc_count, 0);
+        assert_eq!(empty.total_text_bytes, 0);
+
+        q.index_doc("d1".into(), "alpha".into()).await.unwrap();
+        q.index_doc("d2".into(), "beta".into()).await.unwrap();
+        q.index_doc("d3".into(), "gamma".into()).await.unwrap();
+
+        // No sleep — this deliberately lands inside the coalescing window,
+        // before any snapshot flush.
+        let s = q.stats().await.unwrap();
+        assert_eq!(s.doc_count, 3, "stats must see un-flushed writes");
+        assert_eq!(s.total_text_bytes, 14);
+    }
+
+    /// Why: a backfiller distinguishes "not indexed" from "indexed, no hits"
+    /// purely by `doc_count`, so a `rebuild` that left the count stale would
+    /// make a wiped palace read as fully indexed and never get repaired.
+    /// Test: this test itself.
+    #[tokio::test]
+    async fn batch_queue_stats_go_to_zero_after_rebuild() {
+        let (q, _dir) = fresh_queue();
+        q.index_doc("d1".into(), "alpha".into()).await.unwrap();
+        assert_eq!(q.stats().await.unwrap().doc_count, 1);
+        q.rebuild().await.unwrap();
+        let s = q.stats().await.unwrap();
+        assert_eq!(s.doc_count, 0);
+        assert_eq!(s.total_text_bytes, 0);
+    }
+
+    /// Why: the worker only writes the snapshot at the end of a write batch,
+    /// so anything indexed inside the still-open window exists in memory
+    /// only. That was survivable while the only thing that ended a daemon was
+    /// a deliberate shutdown; it stopped being survivable once the supervisor
+    /// began reaping daemons routinely to hold its process cap. Without an
+    /// explicit flush, a reaped palace silently loses its last window — this
+    /// asserts it does not.
+    /// What: indexes a doc and, WITHOUT sleeping past the 50 ms window, calls
+    /// `flush_now`, then reads the snapshot straight off disk.
+    /// Test: this test itself.
+    #[tokio::test]
+    async fn batch_queue_flush_persists_within_the_write_window() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let index = PalaceBm25Index::load_or_create(dir.path()).expect("load palace index");
+        let snapshot = index.snapshot_path().to_path_buf();
+        let q = BatchQueue::new(index, BatchConfig::default());
+
+        q.index_doc("d1".into(), "phoenix rising".into())
+            .await
+            .unwrap();
+        assert!(
+            !snapshot.exists(),
+            "precondition: nothing is on disk until a flush happens"
+        );
+
+        let doc_count = q.flush_now().await.expect("explicit flush must succeed");
+        assert_eq!(doc_count, 1);
+
+        let raw = std::fs::read_to_string(&snapshot).expect("snapshot must exist after flush");
+        assert!(raw.contains("phoenix rising"), "got: {raw}");
     }
 
     #[tokio::test]
--- a/crates/trusty-bm25-daemon/src/batch_queue.rs
+++ b/crates/trusty-bm25-daemon/src/batch_queue.rs
@@ -78,6 +78,10 @@ enum Op {
     Flush {
         reply: oneshot::Sender<Result<usize>>,
     },
+    MissingDocs {
+        doc_ids: Vec<String>,
+        reply: oneshot::Sender<Result<Vec<String>>>,
+    },
 }
 
 /// Configuration for the write-coalescing window and size.
@@ -240,6 +244,31 @@ impl BatchQueue {
             .map_err(|_| anyhow!("bm25 batch worker dropped reply channel"))?
     }
 
+    /// Which of `doc_ids` the live index does not hold.
+    ///
+    /// Why: this is the coverage question `stats` cannot answer. A backfiller
+    /// comparing counts declares a palace covered as soon as the daemon holds
+    /// enough documents, regardless of WHICH documents those are — so a corpus
+    /// carrying stale entries reads as complete over drawers it has never
+    /// seen. Asking by id removes the inference entirely.
+    /// What: sends `Op::MissingDocs`; the worker answers it inline (like
+    /// `Search` and `Stats`) so it never waits out the write-coalescing
+    /// window, and so the answer includes writes acked but not yet flushed.
+    /// Test: `batch_queue_missing_docs_sees_pending_writes`.
+    pub async fn missing_docs(&self, doc_ids: Vec<String>) -> Result<Vec<String>> {
+        let (reply_tx, reply_rx) = oneshot::channel();
+        self.tx
+            .send(Op::MissingDocs {
+                doc_ids,
+                reply: reply_tx,
+            })
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker channel closed"))?;
+        reply_rx
+            .await
+            .map_err(|_| anyhow!("bm25 batch worker dropped reply channel"))?
+    }
+
     /// Run a BM25 search against the live index.
     ///
     /// Why: search is read-only but still flows through the worker so the
@@ -317,6 +346,10 @@ async fn batch_worker(mut rx: mpsc::Receiver<Op>, mut index: PalaceBm25Index, co
                 let _ = reply.send(flush_now(&mut index));
                 continue;
             }
+            Op::MissingDocs { doc_ids, reply } => {
+                let _ = reply.send(Ok(index.missing_docs(&doc_ids)));
+                continue;
+            }
             Op::Rebuild { reply } => {
                 let new_count = index.rebuild();
                 let flush_result = index
@@ -358,6 +391,13 @@ async fn batch_worker(mut rx: mpsc::Receiver<Op>, mut index: PalaceBm25Index, co
                         // figures a backfiller reads mid-run never lag behind
                         // the acks it has already collected.
                         let _ = reply.send(Ok(stats_of(&index)));
+                    }
+                    Some(Op::MissingDocs { doc_ids, reply }) => {
+                        // Same race-freedom argument as `Search` and `Stats`:
+                        // every write dispatched before this op is already
+                        // applied, so a coverage check taken mid-run never
+                        // reports a document missing that was already acked.
+                        let _ = reply.send(Ok(index.missing_docs(&doc_ids)));
                     }
                     Some(Op::Flush { reply }) => {
                         // Flushing mid-batch is safe and is the whole point:
@@ -470,7 +510,11 @@ fn apply_write_op(index: &mut PalaceBm25Index, op: Op) {
         }
         // Caller guarantees only write ops reach here; treat anything else
         // as a programmer error in this module.
-        Op::Search { .. } | Op::Rebuild { .. } | Op::Stats { .. } | Op::Flush { .. } => {
+        Op::Search { .. }
+        | Op::Rebuild { .. }
+        | Op::Stats { .. }
+        | Op::Flush { .. }
+        | Op::MissingDocs { .. } => {
             tracing::error!("apply_write_op called with non-write op — this is a bug");
         }
     }
@@ -572,6 +616,31 @@ mod tests {
         let s = q.stats().await.unwrap();
         assert_eq!(s.doc_count, 3, "stats must see un-flushed writes");
         assert_eq!(s.total_text_bytes, 14);
+    }
+
+    /// Why: a coverage check taken while a backfill is still running must see
+    /// every document the daemon has already acked, otherwise the backfiller
+    /// re-sends work it landed — or, on the post-run check, reports a document
+    /// missing that is sitting un-flushed in the open write window and marks a
+    /// complete palace incomplete.
+    /// What: indexes two docs and asks about three WITHOUT sleeping past the
+    /// coalescing window; only the never-sent id comes back.
+    /// Test: this test itself.
+    #[tokio::test]
+    async fn batch_queue_missing_docs_sees_pending_writes() {
+        let (q, _dir) = fresh_queue();
+        let asked = vec!["d1".to_string(), "d2".to_string(), "d3".to_string()];
+        assert_eq!(q.missing_docs(asked.clone()).await.unwrap(), asked);
+
+        q.index_doc("d1".into(), "alpha".into()).await.unwrap();
+        q.index_doc("d2".into(), "beta".into()).await.unwrap();
+
+        // No sleep — deliberately inside the coalescing window.
+        assert_eq!(
+            q.missing_docs(asked).await.unwrap(),
+            vec!["d3".to_string()],
+            "missing_docs must see un-flushed writes"
+        );
     }
 
     /// Why: a backfiller distinguishes "not indexed" from "indexed, no hits"
--- a/crates/trusty-bm25-daemon/src/index.rs
+++ b/crates/trusty-bm25-daemon/src/index.rs
@@ -206,6 +206,20 @@ impl PalaceBm25Index {
         self.inner.len()
     }
 
+    /// Corpus size in bytes of retained document text.
+    ///
+    /// Why (#2846): this struct keeps every document's full text in `docs` so
+    /// it can re-serialise the snapshot, which means per-daemon RSS grows
+    /// linearly with the corpus. A supervisor enforcing a memory cap should be
+    /// able to see that figure rather than infer it from doc count alone —
+    /// 1311 one-line drawers and 1311 page-long ones cost very different RAM.
+    /// What: sums `str::len()` over the retained text. O(n) over documents, not
+    /// bytes, and only called on the `stats` path — never in the hot loop.
+    /// Test: `palace_index_stats_track_docs_and_bytes`.
+    pub fn total_text_bytes(&self) -> u64 {
+        self.docs.values().map(|t| t.len() as u64).sum()
+    }
+
     /// Snapshot path the daemon writes to. Exposed for diagnostics / tests.
     #[allow(dead_code)] // public for diagnostics; consumed by tests
     pub fn snapshot_path(&self) -> &Path {
@@ -353,6 +367,33 @@ mod tests {
         let raw = std::fs::read_to_string(idx.snapshot_path()).unwrap();
         assert!(raw.contains("\"doc_id\":\"x\""));
         assert!(raw.contains("\"text\":\"one two three\""));
+    }
+
+    /// Why: `stats` is the operation callers use to tell "indexed, no hits"
+    /// from "not indexed", so both figures must track mutation exactly —
+    /// including the delete path, where a stale byte total would overstate the
+    /// corpus a supervisor is budgeting RAM for.
+    /// What: indexes two docs, checks both figures, deletes one, checks again.
+    /// Test: this test itself.
+    #[test]
+    fn palace_index_stats_track_docs_and_bytes() {
+        let dir = tempdir();
+        let mut idx = PalaceBm25Index::load_or_create(dir.path()).unwrap();
+        assert_eq!(idx.doc_count(), 0);
+        assert_eq!(idx.total_text_bytes(), 0);
+
+        idx.index_doc("a", "hello");
+        idx.index_doc("b", "world!");
+        assert_eq!(idx.doc_count(), 2);
+        assert_eq!(idx.total_text_bytes(), 11);
+
+        assert!(idx.delete_doc("a"));
+        assert_eq!(idx.doc_count(), 1);
+        assert_eq!(idx.total_text_bytes(), 6);
+
+        idx.rebuild();
+        assert_eq!(idx.doc_count(), 0);
+        assert_eq!(idx.total_text_bytes(), 0);
     }
 
     #[test]
--- a/crates/trusty-bm25-daemon/src/index.rs
+++ b/crates/trusty-bm25-daemon/src/index.rs
@@ -220,6 +220,25 @@ impl PalaceBm25Index {
         self.docs.values().map(|t| t.len() as u64).sum()
     }
 
+    /// Which of `doc_ids` this index does not hold.
+    ///
+    /// Why: coverage is a SET question, and [`Self::doc_count`] answers a
+    /// different one. The two agree only while the index holds exactly the
+    /// caller's documents and nothing else; one stale document left behind by
+    /// a delete that never happened is enough to make the count report
+    /// coverage the set does not have. Answering by id removes the inference.
+    /// What: retains the requested ids absent from `docs`, in request order.
+    /// Duplicated request ids are reported once per occurrence — the caller
+    /// owns de-duplication. O(n log m) over the requested ids.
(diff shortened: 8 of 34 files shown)


bobmatnyc:
## Verdict: BLOCK

`code-critic` adversarial review, rung 5. Reviewed against the spec and the code only.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-memory/src/bm25_backfill.rs` | 224, 157 | Pre-flight skip compares a **count** where it needs a **set**: `stats.doc_count >= total`. trusty-memory never issues BM25 `delete`, so deleted drawers leave permanent stale docs and `doc_count` drifts above live coverage. `AlreadyIndexed` then returns `fully_indexed() == true` for a palace with unindexed drawers. | Delete the skip — the feeder is idempotent and measured at 1024 docs / 84 ms, so it buys nothing. `AlreadyIndexed` should survive only for the genuinely-empty case at :216. If the skip stays it needs a set comparison (a `has_document` / doc-id digest op), and `fully_indexed()` must not return `true` off a bare count. | Fix here |
| HIGH | `crates/trusty-memory/src/main.rs` | 1033 | `spawn_startup_backfill` is the **only** production caller of backfill. Nothing runs it after a runtime drop, so a `try_send` drop stays invisible until the next trusty-memory restart. The "drop-on-full is now recoverable" argument holds only across a restart — and with the CRITICAL above, not even then. | Either trigger a repair when a drop is observed (`bm25_index_enqueue`'s `Full` arm can set a per-palace dirty flag a periodic sweep consumes), or state in the module doc that repair is restart-scoped and drop the recoverability claim. | Fix here |
| HIGH | `crates/trusty-memory/src/bm25_supervisor.rs` | 201, 327 | `spawn_gate` — the double-spawn fix and the aggregate-cap fix — has **zero concurrent coverage**. Every cap test is a serial `for … .await` loop on `flavor = "current_thread"` (`bm25_backfill_e2e.rs:228`, `bm25_supervisor_tests.rs`). Deleting `spawn_gate` entirely leaves every test green. | Add a test that drives ≥2 `ensure_running` calls concurrently (`JoinSet` on a `multi_thread` runtime): one for the same palace asserting a single spawn, one across N>cap palaces asserting the population never exceeds the cap. | Fix here |
| HIGH | `.github/workflows/test-pointers.yml` | — | **CI gate is red.** `scripts/check_test_pointers.sh` exits 1 on two dangling pointers this diff introduces. Absent from the PR body's gate list. | `bm25_supervisor.rs:424` cites `dead_child_is_evicted_and_respawned` (does not exist anywhere); `:478` cites `unmeasurable_rss_does_not_reap` (actual name: `over_rss_limit_is_false_without_a_measurement`). Correct both pointers. | Fix here |
| HIGH | `crates/trusty-bm25-daemon/src/lib.rs` | 251 | The SIGTERM flush fix has **no CI-runnable regression test**. `a_reaped_palace_respawns_on_next_use` is `#[ignore]`d and CI runs without `--include-ignored` (`ci.yml:334`). It is also timing-dependent: it only goes red pre-fix if the reap lands inside the 50 ms write window, so a scheduling hiccup makes it pass against broken code. Reverting this block leaves every CI-run test green. | Add a non-ignored test that drives `run()` in-process, indexes inside the write window, sends SIGTERM, and asserts the snapshot on disk holds the doc. The fail-open rule says review the fix harder than the bug; right now the bug had better coverage. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_supervisor.rs` | 830 | Supervisor SIGTERM patience (2000 ms) exactly equals the daemon's own `SHUTDOWN_FLUSH_TIMEOUT` (`lib.rs:149`, 2 s). The daemon additionally needs signal delivery, socket cleanup and exit, so it always overruns when the flush is slow — SIGKILL can land inside the shutdown flush the fix just added. No corruption (flush is tmp+rename) but the window's writes are lost. | Give the supervisor margin: 5 s wait, or drop the daemon budget to 1 s. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_backfill_tests.rs` | 167, 186, 21 | Three tests re-implement the logic they name instead of calling it. `palace_docs_skips_blank_drawers` inlines the filter and never calls `palace_docs`. `startup_backfill_respects_the_opt_out` inlines the env guard and never calls `spawn_startup_backfill`. `backfill_state_palace_is_disabled_without_a_client` never calls `backfill_state_palace` — it builds a `short_circuit` report by hand. All three pass if the production logic is deleted; the PR body cites the third as the fail-open evidence for "daemon spawn fails / lane off". | Call the functions under test. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_backfill.rs` | 121, 138, 316 | `skipped_empty` is never assigned non-zero. `palace_docs` drops blank drawers before `total` is computed, so `drawers_total` is already the non-empty count and the field is dead — which makes the `saturating_sub` at :163 a no-op over dead state. `drawers_total`'s doc ("Drawers the palace holds") is also wrong. | Either populate `skipped_empty` from `handle.drawers.len() - docs.len()` and set `drawers_total` to the true total, or remove the field and fix the doc. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_backfill.rs` | 245 | A newer client against a 0.1.x daemon gets `ERR_METHOD_NOT_FOUND` for `stats`, which this arm classifies as `DaemonUnavailable`. Fails closed, but misreports a healthy daemon and silently disables backfill for it. Reachable via socket adoption (`bm25_supervisor.rs:342`) or `TRUSTY_BM25_EXTERNAL=1`. | Distinguish method-not-found (`-32601`) from a transport error and log "daemon too old for `stats` — upgrade to ≥0.2.0". | Fix here |
| LOW | `crates/trusty-common/src/bm25_client.rs` | ~85 | `Bm25Stats` carries `#[serde(default)]` on both fields, so a daemon-side rename decodes as `doc_count: 0` instead of erroring — the exact silent drift `stats_response_decodes`' doc comment claims to prevent. That test decodes a hand-written literal and never involves the daemon's `StatsResult`, so it cannot detect drift. Fails in the safe direction here (0 → full backfill runs). | Drop `#[serde(default)]`, or correct the comment to say e2e is what catches drift. | Fix here |
| LOW | `crates/trusty-memory/src/tools/memory_ops.rs` | 554 | `memory_forget` never deletes the drawer's BM25 doc, though the daemon implements `delete` and the client exposes it. Stale docs accumulate forever, keep matching lexical queries, and are what makes the CRITICAL above reachable. | Wire `client.delete(drawer_id)` into the forget path. Genuinely separable from this PR's scope — but the CRITICAL fix must not assume it lands first. | Promote |

## Required Changes

1. Fix the coverage predicate (CRITICAL). This is the alarm the whole design rests on, and it can currently report healthy over a palace it has never indexed.
2. Resolve or scope the runtime-repair gap — a drop must have something that repairs it, or the recoverability claim comes out of the module doc and the PR body.
3. Add concurrent coverage for `spawn_gate`, or the headline concurrency fix ships unverified.
4. Turn the `test-pointers` CI gate green.
5. Give the SIGTERM durability fix a test that runs in CI.

## Notes

Verified sound, for the record:

- **Lock ordering is correct.** Exhaustive check of every path including error and early-return arms: `spawn_gate` → `children` only, never the reverse. `shutdown` and `supervised_count` take `children` alone; `lookup_live`, `enforce_limits` and the insert take `children` only while the gate is already held; the fast-path `lookup_live` releases its guard before the gate is acquired. No deadlock.
- **`fully_indexed()` on the `Completed` arm fails closed** — a failed post-run read-back leaves `final_doc_count: None` and reports `false`. Correct. The defect is confined to the `AlreadyIndexed` arm.
- **RSS `None` never reaps** (`over_rss_limit:648`). Correct, and `under_rss_limit_children_are_left_alone` stops the reaps-unconditionally mutation from passing.
- **Reap during an in-flight backfill fails closed** — index calls error, status goes `Partial`, the post-run `stats` fails, `fully_indexed()` is `false`.
- **The SIGTERM flush is not narrowed to the reap path.** All three `select!` arms in `lib.rs:227` — sigterm, sigint, and the accept loop returning or panicking — fall through to the flush at :251.
- **`Op::Flush` mid-batch is correct.** It clears the dirty bit, the end-of-window `if write_count > 0` flush is then a no-op, and later writes in the same window re-dirty and re-flush. The `Op::Rebuild` arm's `write_count = 0; break` is also correct. No bug found in target 3.
- **Snapshot flush is atomic** (tmp + rename), so a SIGKILL mid-flush loses the window but cannot corrupt the snapshot.
- **Semver**: 0.1.3 → 0.2.0 is additive-only — new method, new result type, new queue methods, nothing removed or changed. 0.1.4 would have been accurate; 0.2.0 is conservative, not wrong.

Not flagged, per scope: the lane staying off by default, `TRUSTY_BM25_DAEMON` as an env var, and the two scoped gates.

One downgraded observation: `enforce_limits:488` holds the `children` mutex across `process_rss_mb`, which on Linux is a blocking `/proc` read. Bounded to `max_live` reads of a few hundred bytes, so it is not worth a row.

bobmatnyc:
## Review fixes — all CRITICAL and HIGH addressed

Rebased onto `b8746f59`. Head is `b17e5c163`.

### CRITICAL — coverage established by identity, not by counting

The skip and the predicate both went through `stats.doc_count`. Replaced with a new protocol op.

- `missing_docs` (`trusty-bm25-daemon`): takes a list of doc ids, returns the subset the daemon does not hold. Answered inline by the batch worker in both the cold path and mid-batch, so it sees writes acked but not yet flushed.
- `BackfillReport::fully_indexed()` is now `missing_after == Some(0)`. There is no arithmetic left in it and **no status can satisfy it** — `AlreadyIndexed` no longer implies coverage, it just reports that the run was skipped. The pre-flight skip fires only on a verified empty missing set.
- A probe that could not run reports `None` and logs at `error!`. A daemon predating the op is distinguished from an unreachable one via a typed `Bm25RpcError` carrying the JSON-RPC code (`is_method_not_found`), which also closes the MEDIUM about `-32601` misreporting a healthy daemon.
- `stats` is kept for observability only. `stale_doc_estimate()` reports `doc_count - indexable_drawers` at `warn!`, so the drift that broke the old predicate is now visible while it accumulates instead of after it lies.

**Mutation proof** — restored the old count-based predicate:

```
test bm25_backfill::tests::fully_indexed_requires_a_verified_empty_missing_set ... FAILED
panicked at crates/trusty-memory/src/bm25_backfill_tests.rs:203:5:
400 documents cannot cover 10 drawers when one of the 10 is absent
test result: FAILED. 12 passed; 1 failed
```

The e2e test also deletes a live doc and seeds two stale ones, asserts `final_doc_count > drawers_total` as a precondition (so every count comparison reports coverage), then asserts the run actually happened.

### HIGH — recovery is now real, not theoretical

`spawn_startup_backfill` was the only production caller. New `bm25_repair` module: a dropped enqueue, a failed index call, and an unverified startup sweep all `mark_dirty(palace)`; a periodic sweep (`TRUSTY_BM25_REPAIR_INTERVAL_SECS`, default 300s, `0` disables) re-runs the lossless backfill and re-queues anything still unverified, so a daemon that is down during one pass does not convert a recoverable drop into a permanent gap.

**Mutation proof** — removed the `mark_dirty` call and the requeue:

```
test tools::tests::bm25_index_queue_drops_when_full ... FAILED
  assertion failed: a dropped index op must queue its palace for coverage repair
test bm25_repair::tests::an_unrepairable_palace_stays_queued ... FAILED
  assertion failed: an unverified palace must stay queued for the next pass
```

### HIGH — `check_test_pointers.sh` is green

```
test-pointers: resolved 22192 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
```

Both pointers fixed. `dead_child_is_evicted_and_respawned` did not exist because the restart path had no test at all — it now has one. Script added to the gate chain below.

### HIGH — concurrency fixes are verified

New `tests/bm25_supervisor_concurrency.rs`, **not** `#[ignore]`d — `env!("CARGO_BIN_EXE_trusty-bm25-daemon")` makes cargo build the daemon before the test binary runs, so the "binary might not be on disk" reason the older e2e files skip themselves does not apply. `multi_thread` runtime, `Barrier` + `JoinSet` so the callers are simultaneous rather than interleaved.

Added `spawned_count()` because a double spawn is otherwise invisible: the loser's daemon dies on EADDRINUSE, so `supervised_count()` reads 1 either way.

**Mutation proof** — deleted `let _spawn = self.spawn_gate.lock().await;`:

```
test concurrent_callers_for_one_palace_spawn_exactly_one_daemon ... FAILED
  left: 8   right: 1
test a_concurrent_fanout_never_exceeds_the_cap ... FAILED
  a 6-way simultaneous fan-out left 6 daemons resident against a cap of 2
test result: FAILED. 1 passed; 2 failed
```

### HIGH — the durability fix has CI-visible, deterministic coverage

`run_until(config, shutdown)` takes the shutdown trigger as a parameter; `run` supplies the real SIGTERM/SIGINT select and is otherwise unchanged (handlers now install before snapshot load and bind, so a SIGTERM during startup reaches the shutdown path rather than the kernel default). New `tests/shutdown_flush.rs` — not ignored, no signals — sets a 10-minute write window so the indexed document is guaranteed un-flushed, asserts that as a precondition, then fires shutdown.

**Mutation proof** — removed the flush block:

```
test shutdown_flushes_the_open_write_window ... FAILED
panicked at tests/shutdown_flush.rs:117:9:
shutdown must flush the open write window to .../bm25_index.json: No such file or directory
test a_flushed_snapshot_is_loaded_by_the_next_run ... FAILED
```

### MEDIUM/LOW

Fixed: the three self-implementing tests now call `docs_from_drawers`, `startup_backfill_opted_out`, and `backfill_state_palace` (with a real `PalaceHandle`); supervisor SIGTERM patience 2s → 5s with a test pinning it above the daemon's own budget; `skipped_empty` / `drawers_total` are honest via `PalaceDocs`, which also removed the `saturating_sub`; `-32601` classified separately; `#[serde(default)]` dropped from `Bm25Stats` with a test asserting a renamed field now fails the decode.

**Promoted, not fixed here:** the missing BM25 `delete` on `memory_forget` → [#5053](https://github.com/bobmatnyc/trusty-tools/issues/5053). The CRITICAL fix does not depend on it — coverage is by drawer id, so stale documents cannot manufacture coverage — but stale lexical hits and unreclaimed daemon RSS are real and separable. The repair sweep is additive and cannot remove a stale document, which is noted in the issue.

### Gates — rung 5

`git diff --name-only origin/main...HEAD` touches `trusty-bm25-daemon`, `trusty-common`, `trusty-memory` only; zero matches for `trusty-code-gui`/`trusty-mpm-gui`, so the workspace check excludes only those.

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` (minus the 3 GUI crates) | clean |
| `cargo clippy -p trusty-bm25-daemon --all-targets -- -D warnings` | clean |
| `cargo clippy -p trusty-common --all-targets -- -D warnings` | clean |
| `cargo clippy -p trusty-memory --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-bm25-daemon` | 46 + 3 + 2 passed, 0 failed |
| `cargo test -p trusty-common --features memory-core` | 892 + 6 + 11 + 4 passed, 0 failed, 13 ignored |
| `cargo test -p trusty-memory` | 593 lib + 78 integration passed, 0 failed |
| `--include-ignored` (bm25 e2e + supervisor e2e + concurrency) | 4 + 3 + 1 passed, 0 failed |
| `bash scripts/check_test_pointers.sh` | 22192 resolved, 0 dangling |
| `bash scripts/check_line_cap.sh` | 3753 files, 0 violations |
| `bash scripts/check_changelog_fragment.sh` | all 3 crates recorded |

`-A clippy::nonminimal_bool` confirmed unnecessary after #5039 and dropped — clippy is clean without it.

The lane is still off by default.

bobmatnyc:
## Verdict: WARN

`code-critic` re-review of `b17e5c16`, rung 5. Reviewed against the spec and the code only.

## Prior round — all five Required Changes verified fixed, by mutation

Every proof was reproduced in the worktree, not accepted from the PR body.

| Prior finding | Mutation applied | Result |
|---|---|---|
| CRITICAL — count-vs-set predicate | restored `final_doc_count >= drawers_total - skipped_empty` in `fully_indexed()` | `FAILED. 39 passed; 3 failed` — `fully_indexed_requires_a_verified_empty_missing_set`, `empty_palace_is_already_indexed`, `a_palace_of_only_blank_drawers_is_covered_and_counted` |
| HIGH — `spawn_gate` unverified | deleted `let _spawn = self.spawn_gate.lock().await;` | `left: 8 / right: 1` on spawns, and `a 6-way simultaneous fan-out left 6 daemons resident against a cap of 2`. Exactly as claimed |
| HIGH — durability test weaker than the bug | deleted the `timeout(SHUTDOWN_FLUSH_TIMEOUT, shutdown_queue.flush_now())` block | `FAILED. 0 passed; 2 failed` on **5 of 5** runs. Deterministic — the 10-minute window does what the author says it does |
| HIGH — recovery was theoretical | deleted `mark_dirty` from `bm25_index_enqueue`'s `Full` arm **and** the re-insert in `run_repair_pass`'s failure arm | `bm25_index_queue_drops_when_full` and `an_unrepairable_palace_stays_queued` both FAILED |
| HIGH — dangling test pointers | — | `test-pointers: resolved 22192 Test: citation(s) — 0 dangling pointers — OK` |

The four MEDIUMs and two LOWs are also addressed: `skipped_empty` is now populated by `docs_from_drawers` and asserted; the three self-referential unit tests call their production functions (mutation-visible — deleting the count predicate broke three of them, including two that previously restated logic); `-32601` is classified separately via `Bm25RpcError` / `is_method_not_found`; `#[serde(default)]` is gone and `stats_response_rejects_a_renamed_field` pins it; `SIGTERM_PATIENCE` is 5 s.

Supporting gates, all green on the unmutated tree: `cargo fmt --check` + `cargo clippy -p trusty-memory -p trusty-bm25-daemon -p trusty-common --all-targets -- -D warnings` → EXIT=0. `cargo test -p trusty-memory -p trusty-bm25-daemon -p trusty-common` → `1352 passed; 0 failed`, `593 passed; 0 failed`, and 28 smaller targets, 0 failures anywhere. The four `#[ignore]`d e2e tests pass with `--include-ignored`, including the new stale-drift assertion. `check_line_cap.sh` → 0 violations.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-memory/src/bm25_backfill.rs` | 630, 633 | The startup sweep enumerates `state.registry.list()`, which is the **LRU key set of currently-open handles**, not the corpus (`registry.rs:325` — "this only returns currently-open handles; to list all persisted palaces use `list_palaces`"). The cache is capped at `DEFAULT_MAX_OPEN_PALACES = 64`; the module doc's own host has ~99 palaces. At least 35 are never enumerated, never probed, never marked dirty — and the sweep then logs `startup sweep complete, all coverage verified`. Line 633 compounds it: the idle-to-disk eviction ticker is armed by `spawn_background_maintenance` *before* the sweep is spawned, so a palace evicted between the `list()` snapshot and the `get()` is silently `continue`d with no `mark_dirty`. This is the fail-open the PR exists to remove, relocated from the predicate to the enumeration. | Enumerate `PalaceRegistry::list_palaces(&state.data_root)` and hydrate each with `registry.open_palace(&state.data_root, &id)` on the blocking pool — the repo already has the idiom and the precedent in `service/helpers.rs:244` `open_palaces_blocking`, whose doc says using the cached view "would silently drop ~98.9% of the corpus, which is a correctness regression, not an optimisation" (#4637). On an open failure, `mark_dirty` instead of `continue`. | Fix here |
| HIGH | `crates/trusty-memory/src/bm25_repair.rs` | 137–142 | `run_repair_pass` drains the dirty set first, then resolves each palace with `registry.get`. `get` is a bare LRU lookup — it does not hydrate. A palace marked dirty by a dropped enqueue that then goes idle and is LRU-evicted resolves to `None`, and the arm drops it from the queue permanently ("The palace was evicted or deleted"). Eviction is routine and temporary; deletion is permanent; the arm treats them identically. The coverage gap then survives until the next trusty-memory restart — which is the exact condition the module's own doc says it exists to remove. The sibling arm (coverage unverified) re-queues; this one does not. `a_vanished_palace_is_dropped_from_the_queue` uses `"no-such-palace"`, so it asserts the behaviour rather than distinguishing the two cases. | Resolve with `state.registry.open_palace(&state.data_root, &id)`, which handles resident and evicted alike. Drop only when the palace is genuinely absent from `PalaceRegistry::list_palaces(&state.data_root)`; re-queue on a transient open failure (`open_palace` also errs on a bounded redb open-queue timeout). Add a test that registers a palace, marks it dirty, evicts it from the registry, and asserts it stays queued. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_supervisor_tests.rs` | 478 | `sigterm_patience_exceeds_the_daemon_flush_budget` hardcodes `let daemon_flush_budget = Duration::from_secs(2);`. Raising the daemon's `SHUTDOWN_FLUSH_TIMEOUT` to 6 s reintroduces the exact bug the test is named for and the test stays green — it pins one number against a copy of the other, not against the relationship. | `trusty-memory` already declares `trusty-bm25-daemon = { workspace = true }` (`Cargo.toml:93`), so make `SHUTDOWN_FLUSH_TIMEOUT` `pub` in `trusty-bm25-daemon/src/lib.rs:150` and import it. The comment's premise — "not importable" — is wrong; only "private" is true, and that is the thing to change. | Fix here |
| MEDIUM | `crates/trusty-memory/src/tools/bm25.rs` · `bm25_backfill.rs` | 155, 164 · 649 | Three of the four places that mark a palace dirty have no test. Only the `Full` arm of `bm25_index_enqueue` is mutation-covered. The spawn-failure arm (155), the index-failure arm (164), and the startup sweep's own `mark_dirty` (649) can all be deleted with every test still green — and finding 1 above makes the sweep's marking path the one that matters most. | Drive the worker with a supervisor pointed at a non-existent binary and assert the palace lands in `dirty_palaces(&state)`; same for a client pointed at a dead socket. `bm25_supervisor_concurrency.rs` proves `env!("CARGO_BIN_EXE_trusty-bm25-daemon")` makes a real-daemon integration test CI-runnable, so the sweep's arm is now testable too. | Fix here |
| LOW | `crates/trusty-memory/src/bm25_backfill.rs` | 322 | `probe_coverage` sums `cov.missing.len()` and never reads `cov.checked`, while `protocol.rs`'s `MissingDocsResult` doc says `checked` is echoed back "so a caller can tell a genuinely-empty answer from a request that was silently truncated". Nothing performs that check, so the guard is documented and absent. | Either `if cov.checked != chunk.len() { return Coverage::Unreachable }`, or drop the claim from the protocol doc. | Fix here |
| LOW | `crates/trusty-memory/src/bm25_supervisor.rs` | 581 | `shutdown`'s doc still says the SIGTERM wait is "up to ~2 s" after `SIGTERM_PATIENCE` moved to 5 s — the stale half of this round's own fix. | Say 5 s, or point at `SIGTERM_PATIENCE`. | Fix here |

## Required Changes

1. Make the startup sweep enumerate the corpus, not the open-handle cache (finding 1). Right now a host with more palaces than the LRU cap gets a sweep that reports full verification over a subset it silently chose.
2. Stop `run_repair_pass` from dropping an evicted palace (finding 2). Eviction is not deletion, and the module's whole purpose is that a gap does not wait for a restart.

## New surface — what was checked and found sound

Answering the five hunt targets directly.

1. **`missing_docs` answered inline by the batch worker.** No deadlock: the worker is the sole consumer, never sends to its own channel, and holds no lock across the reply. No starvation: `index.missing_docs` is a `BTreeMap` lookup per id, chunked at 256, and the `biased;` select still breaks on the write-window deadline first. Not stale: the worker applies writes in arrival order, so any op reaching the `MissingDocs` arm sees every previously-acked write — `batch_queue_missing_docs_sees_pending_writes` pins that inside the window. During a flush the op simply queues, because `index.flush()` runs synchronously on the worker. No finding.
2. **`bm25_repair`'s sweep.** Cannot overlap itself — one task, `run_repair_pass` awaited serially inside the tick loop. Against a palace mid-backfill it is idempotent, because `upsert_document` is keyed by `doc_id` and both probes fail closed. It respects both limits, since it reaches the daemon through `socket_for_palace` → `ensure_running` → `enforce_limits`. The drain-then-process order does not lose a concurrent mark: the pass removes the id once at the start, a later `mark_dirty` re-inserts, and the success arm never removes again. The one soft spot is `tokio::time::interval`'s default `MissedTickBehavior::Burst` — a pass that outruns the interval yields the next tick immediately — but an unreachable daemon short-circuits in the pre-flight probe, so the pass cannot realistically outrun 300 s. Below the bar.
3. **Signal handlers installed before snapshot load and bind.** `signal()` registers and buffers; the future is not polled until the `select!` at step 5, after `queue` and `listener` exist. So a SIGTERM during startup is deferred, not dropped, and shutdown never runs against partially-initialised state — the flush target `shutdown_queue` is always a live clone. The behaviour change is real but benign: a SIGTERM during startup no longer takes the kernel's default disposition, it binds and then exits cleanly. The supervisor's 5 s SIGTERM→SIGKILL still bounds it. No finding.
4. **Real daemon subprocesses in CI.** The `CARGO_BIN_EXE_` trick is sound and cargo guarantees the binary exists. No process leak: I ran the deleted-`spawn_gate` mutation, which panics with six daemons resident, and `ps` showed zero `trusty-bm25-daemon` afterwards — `kill_on_drop(true)` fires through the unwind. Socket *files* do leak on a panicking run (nine in my `$TMPDIR` from the mutation runs), but each test clears its own name at start, so it is residue, not breakage. The genuine risk is timing: libtest runs the three tests in parallel threads, so up to 15 daemon spawns can be in flight at once, each needing to bind inside `SPAWN_PROBE_TIMEOUT` (3 s) on a 2-core `ubuntu-latest` runner under a debug build. Worth watching on the first few CI runs; not a finding at >80% confidence.
5. **`fully_indexed()` returning `None`.** Three production callers — `backfill_palace`'s log branch, `spawn_startup_backfill`, `run_repair_pass` — and none coerces. `missing_after == Some(0)` makes `None` false structurally. The `error!` is not the only signal: the startup sweep marks the palace dirty and the repair pass re-queues it. Correct. (Finding 1 is about palaces that never reach any of those callers, which is a different failure.)

## Notes

**On deferring the `memory_forget` BM25 `delete` to #5053** — the reasoning holds. Coverage is now a set statement over drawer ids, so a stale document cannot manufacture it; `missing_docs_answers_by_identity_not_count` is the direct proof (a two-document index satisfying `doc_count == asked.len()` while missing one of the two ids), and the e2e `backfill_indexes_every_drawer_without_drops` now deletes a live doc, seeds two stale ones, and asserts the run happens anyway. The CRITICAL fix genuinely does not depend on the delete landing.

Two things stay reachable, neither newly so:

- **Stale lexical hits.** A forgotten drawer's document keeps matching queries and `memory_recall` will fuse a doc_id for a drawer that no longer exists. Pre-existing — the lane has always indexed on write — and the mitigation this PR ships (`stale_doc_estimate` → one `warn!`) is observability, not repair.
- **Stale text is persisted, and the new RSS cap is not.** `PalaceBm25Index` retains every document's full text and re-serialises it into the snapshot, so a corpus that grows past `DEFAULT_RSS_LIMIT_MB` (512) cannot come back under it: reap → respawn → reload the same snapshot → over the limit again → reaped on the next `ensure_running`, indefinitely. That thrash is newly *possible* because this PR is what makes the RSS limit enforced. At the stated corpus size — single-digit MB across ~99 palaces — it is three orders of magnitude away, so it is a note, not a row. It is the concrete cost of leaving #5053 open, and worth putting in that issue.

**Explicitly not flagged**, per scope: the lane staying off by default. Also considered and dropped below the bar: post-run coverage is verified against the daemon's in-memory index, so a SIGKILL inside the ~50 ms tail window after the last batch could lose acked documents while the report said covered — but the graceful reap path SIGTERMs and flushes, and the window is bounded by one batch.

The four e2e tests in `bm25_backfill_e2e.rs` are still `#[ignore]`d. That is now optional rather than necessary: `bm25_supervisor_concurrency.rs` demonstrates `env!("CARGO_BIN_EXE_trusty-bm25-daemon")` puts the binary on disk unconditionally, which is the reason those tests gate themselves. Not a finding — the CI-runnable unit and dispatch tests carry the CRITICAL fix on their own, and I verified the e2e set passes with `--include-ignored`.

bobmatnyc:
## Re-review fixes — both HIGHs, plus one more instance of the pattern

Rebased onto `e97e40e6`. Head is `4850e2874`.

### HIGH 1 — the startup sweep enumerated only open palaces

`registry.list()` snapshots the LRU key set of currently-**open** handles, capped at 64; this host holds ~99 palaces. The registry's own doc says it: *"this only returns currently-open handles; to list all persisted palaces use `PalaceRegistry::list_palaces`."*

Followed the #4637 precedent in `service/helpers.rs:244` — `PalaceRegistry::list_palaces` to enumerate, `open_palace` on the blocking pool to resolve, serial so cold opens don't thrash the LRU. Holding the opened `Arc` for each palace's whole backfill also closes the mid-sweep eviction hole, since the idle ticker is armed before the sweep runs.

The sweep body is now an awaitable `run_startup_sweep` returning a `SweepOutcome`, so its enumeration is testable without racing a spawned task.

**Mutation proof** — reverted the enumeration to `registry.list()`:

```
test startup_sweep_enumerates_every_palace_on_disk ... FAILED
  left: 0   right: 3
```

The test drives the sweep from a **cold** `AppState` over a data root that already holds three palaces — the worst case of the eviction the cap causes.

### HIGH 2 — the repair pass dropped evicted palaces permanently

`registry.get` is a bare LRU lookup, and a dirty palace is exactly the kind that goes idle — its writes are failing. Now `open_palace`, with the on-disk palace list used to tell *evicted* (stays queued) from *deleted* (dropped); without that list the two are indistinguishable and one of the two answers is always wrong.

Entries also now leave the queue only on verified coverage rather than being drained up front, so a panic or cancellation mid-pass cannot discard the gap the set exists to remember.

**Mutation proof** — reverted to `registry.get` + drop-on-miss:

```
test an_evicted_palace_is_rehydrated_not_dropped ... FAILED
  left: []   right: ["evicted"]
```

The old test could not catch this — it used a never-existing palace, so it could not distinguish eviction from deletion. It is now split into that pair, and the fixture builds the palace through `create_palace` + `upsert_drawer_sync` so it genuinely exists on disk and genuinely is absent from the LRU.

### The pattern walk — one more instance, in code from this round

`SweepOutcome::default()` is what a failed enumeration returns. It has `incomplete == 0`, so `all_verified()` read **true** over a sweep that examined nothing — the same fail-open, in the accessor I had just written.

`enumerated` is now `Option<usize>` and `all_verified()` requires `Some`, which is the same discipline as `BackfillReport::missing_after`: encoding "did not ask" in the type makes the fail-open unrepresentable rather than merely avoided.

Two adjacent paths closed with it:

- A palace that cannot be **opened** is counted (`unopenable`), queued, and keeps the sweep from reporting clean — previously a bare `continue` into the "all verified" tally.
- The repair sweep catches a panicking pass, so one bad tick cannot silently end all future repairs while the queue grows.

**Mutation proofs:**

```
# all_verified ignoring whether enumeration happened
test a_sweep_that_cannot_enumerate_verifies_nothing ... FAILED

# unopenable palaces silently skipped
test startup_sweep_marks_unopenable_palaces_instead_of_skipping_them ... FAILED
  left: 0   right: 1
```

After these, I do not find a remaining path where a palace can be absent from consideration and the outcome still reads as complete. The four ways a palace can go unexamined — not enumerated, enumeration failed, not openable, not verifiable — each now produce a queued entry or an outcome that cannot report clean. A palace with zero drawers is still skipped, which is correct: there is no id that could be missing.

### MEDIUM/LOW

- `SHUTDOWN_FLUSH_TIMEOUT` is now `pub` and the SIGTERM-margin test imports it. A hardcoded copy stays equal to itself while the real budget drifts, so it could never detect the drift it exists to name.
- All four `mark_dirty` call sites are tested and mutation-proven; three were untested. The two worker paths — a daemon that will not spawn, and an index call that fails — each lose a write as completely as a queue drop. Deleting either `dirty.insert` turns both new tests red.

**Skipped:** nothing from this round. The two LOW items in the verdict are folded into the above (the `Option` change covers the outcome-type LOW; the import covers the constant LOW).

### For #5053, not this PR

Added the critic's note to the issue: stale text is persisted in the snapshot, so a corpus past the RSS cap can never come back under it — reap, reload, over limit, reap. Three orders of magnitude away at current corpus size, so it is recorded as a consequence of the missing delete rather than as a live finding.

### Gates — rung 5

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` (minus the 3 GUI crates) | clean |
| `cargo clippy -p trusty-bm25-daemon\|trusty-common\|trusty-memory --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-bm25-daemon` + `-p trusty-common --features memory-core` | 964 passed, 0 failed |
| `cargo test -p trusty-memory` | 682 passed, 0 failed |
| `--include-ignored` (bm25 e2e + supervisor e2e + concurrency) | 8 passed, 0 failed |
| `bash scripts/check_test_pointers.sh` | 22204 resolved, 0 dangling |
| `bash scripts/check_line_cap.sh` | 3755 files, 0 violations |
| `bash scripts/check_changelog_fragment.sh` | all 3 crates recorded |

The lane is still off by default.

bobmatnyc (commented):
## Verdict: APPROVE

Round 3 re-review at `4850e2874`. Both prior-round HIGHs are closed, and I re-ran every claimed mutation myself rather than taking the counts on trust. Three MEDIUM/LOW findings remain, none of which hold the PR.

### Mutations re-run independently

Each was applied to the source in a separate detached worktree at the PR head, then the named test was run. All six behaved exactly as the engineer described.

| # | Mutation | Result |
|---|---|---|
| 1 | Enumeration reverted to `state.registry.list()` | `startup_sweep_enumerates_every_palace_on_disk` FAILED — `enumerated` read `Some(0)`, expected `Some(3)`. `startup_sweep_marks_unopenable_…` also FAILED (`Some(0)` vs `Some(1)`) |
| 2 | `all_verified()` drops `self.enumerated.is_some() &&` | `a_sweep_that_cannot_enumerate_verifies_nothing` FAILED; 15 passed, 1 failed |
| 3 | Unopenable arm → bare `continue` | `startup_sweep_marks_unopenable_palaces_instead_of_skipping_them` FAILED — `unopenable` 0 vs 1 |
| 4 | All three write-path `dirty` marks deleted (`tools/bm25.rs:155`, `:164`, `:206`) | `bm25_index_queue_drops_when_full`, `a_failed_index_call_queues_the_palace_for_repair`, `a_daemon_that_will_not_spawn_queues_the_palace_for_repair` all FAILED; 77 passed, 3 failed |
| 5 | `open_palace` reverted to `registry.get` in `run_repair_pass` | `an_evicted_palace_is_rehydrated_not_dropped` FAILED — queue `[]` vs `["evicted"]` |
| 6 | Sweep's `incomplete` arm `mark_dirty` deleted | `startup_sweep_enumerates_every_palace_on_disk` FAILED — queue `[]` vs `["sweep-0","sweep-1","sweep-2"]` |

No test was silently filtered. All eleven bm25 coverage tests are named in the baseline run output, and the run reports `602 filtered out` only when a filter was deliberately applied.

### Gates re-run

```
( cargo fmt --check && cargo test -p trusty-memory && cargo test -p trusty-bm25-daemon \
  && cargo clippy -p trusty-memory --all-targets -- -D warnings -A clippy::nonminimal_bool )
EXIT=0
```
733 passed, 0 failed across both crates' full target sets (lib 599 + 5 bins + 17 integration targets + bm25-daemon 46/3/2). `-A clippy::nonminimal_bool` matches the PR body's documented pre-existing `trusty-common` lint; nothing in this diff lints.

### `Option<usize>` — verified complete

`SweepOutcome` has exactly three construction sites, all in `bm25_backfill.rs`: `:678` and `:685` (`::default()` → `enumerated: None`) and `:689` (the literal, `Some(palaces.len())`). No site was missed. `all_verified()` at `:637` cannot return `true` on `None` — mutation 2 proves the `is_some()` guard is load-bearing, not decorative.

### Is the sweep reached? — No, and the PR says so

`spawn_startup_backfill` (main.rs:1033) returns immediately unless `bm25_client` is `Some`, which requires `TRUSTY_BM25_DAEMON=1`, which no shipped path sets. `run_startup_sweep` and `backfill_state_palace` have no other production caller. **This sweep has never enumerated a non-zero count against a real data root — only in tests.** It differs from the earlier fully-inert-feature case in one respect that matters: the inertness is declared, not hidden. The PR body states the lane stays off and names the flip as piece 2. The obligation this creates is on piece 2, which must produce a real log line showing a non-zero `enumerated` against the ~99-palace root before the lane is called working.

Cost of the serial `open_palace` is a non-issue: `load_palaces_from_disk()` already hydrates every persisted palace immediately before the sweep is spawned (main.rs:1000), so the sweep's opens are LRU hits for the resident 64 and at most ~35 cold re-opens, serially, on the blocking pool, in a background task.

### The panicking-pass catch — nothing is swallowed

There is no `catch_unwind` anywhere in `trusty-memory` outside `worker_liveness_tests.rs`. The claim resolves to the snapshot-not-drain design at `bm25_repair.rs:132-138`, which is correct: entries leave `bm25_dirty` only at `:167` (verified absent from disk) or `:192` (verified covered), so a panic mid-pass discards no queued gap. Nothing degrades a real defect into a warning. See LOW-2 for the residual.

## Findings

| MEDIUM | crates/trusty-memory/src/bm25_backfill.rs | 663-687, 605-630 | A fifth unexamined-palace path exists, so the completeness claim is overstated as written. `PalaceStore::list_palaces` returns `Ok` while silently skipping any palace directory whose `palace.json` fails to decode (`palace_store.rs:192-195`) and any `read_dir` entry that errors (`:181-184`). The sweep treats `Ok(vec)` as a complete enumeration. Proven empirically: one palace on disk holding a persisted drawer, `palace.json` overwritten with `{ not json` → `SweepOutcome { enumerated: Some(0), swept: 0, incomplete: 0, unopenable: 0 }`, `all_verified() == true`, repair queue empty. The sweep logs "startup sweep complete, all coverage verified." | Stop claiming exhaustiveness the enumerator does not provide. Reword the `SweepOutcome::enumerated` doc (`:612-615`) and the sweep's "Why the disk" doc (`:642-662`) from "palaces found ON DISK" to "palaces the registry could decode", and name the residual gap. | Fix here |
| MEDIUM | crates/trusty-memory/src/tools/bm25.rs | 214-221 | Sibling-branch asymmetry in the same `match`. `TrySendError::Full` marks the palace dirty (`:206`); `TrySendError::Closed`, three lines later, logs at `debug!` and returns. Both lose the write identically. A dead indexer worker therefore drops every subsequent drawer from the BM25 corpus with nothing queued to repair it — and repair would work, since `run_repair_pass` calls `backfill_state_palace` directly rather than through the worker. | Add `crate::bm25_repair::mark_dirty(state, &req.palace);` to the `Closed` arm, matching `:206`. One line. | Fix here |
| LOW | crates/trusty-memory/src/main.rs | 1026-1028 | Stale justification. The comment reads "Must run AFTER hydration — before it, the registry is empty and the sweep would find nothing to index." That was true of the `registry.list()` version this commit replaced; the sweep now enumerates from disk and no longer depends on hydration ordering. Leaving it documents a dependency that does not exist. | Reword to note the ordering is now incidental, not required. | Fix here |
| LOW | crates/trusty-memory/src/bm25_repair.rs | 233-244 | A panic inside `run_repair_pass` aborts the spawned task, so the `loop` never ticks again and the repair sweep is silently disarmed for the process lifetime. The queue is preserved (correctly) but nothing drains it until restart — the exact pre-fix state the module exists to remove. No panic source identified in the current body, so this is a hardening note, not a live defect. | Wrap the body in `AssertUnwindSafe(...).catch_unwind()` and log at `error!` on a caught panic, or accept and note it. | Parent |

## Notes

- Two paths I checked and am *not* flagging, because the code is right. `PalaceHandle::open_with_intent` falls back to an empty drawer table when `load_drawers` fails (`handle.rs:319-325`), and `load_drawers` itself skips undecodable rows (`kg_redb/read_ops.rs:253-273`). Either would make the sweep's zero-drawer skip (`bm25_backfill.rs:720`) pass over content. But `handle.drawers` is what *every* lane serves from, vector included — a drawer absent there is absent from recall entirely, so BM25 coverage over it is not a meaningful claim. The zero-drawer skip is correct: coverage is defined relative to what the palace serves, and no id can be missing from a set the palace does not serve.
- MEDIUM-1's practical blast radius is small for the same reason. An undecodable `palace.json` also fails `open_palace` (`registry.rs:479` calls the same `load_palace`), so that palace is dark to the whole memory system, not just to BM25. The residual case where a *healthy* palace goes unenumerated is the transient `read_dir` per-entry error, which is rare and self-heals on the next sweep. This is why it is MEDIUM and not HIGH: unlike the round-2 HIGHs, which fired deterministically on a healthy 99-palace host, neither of these fires without upstream I/O damage.
- The snapshot-not-drain change in `run_repair_pass` is a strict improvement over the round-2 drain-up-front version, but nothing tests it — the observable difference only appears under a panic or cancellation mid-pass, which no test induces. Not worth a test; worth knowing.

bobmatnyc:
## Re-review round 3 — the fifth path, plus three fix-heres

Head is `fb0a31b8c`. All four addressed, none deferred.

### The fifth path — you were right, and my completeness claim was wrong

`PalaceStore::list_palaces` returns `Ok` while silently dropping undecodable palaces. I checked that the sweep enumerated from disk and stopped there; I never checked whether "from disk" meant *all of it*.

The fix removes `list_palaces` from the path rather than compensating for it. Enumeration now yields **ids off the data root** — every immediate subdirectory holding a `palace.json`. A palace with unreadable metadata is therefore SEEN, then fails `open_palace`, and lands in the `unopenable` count and the repair queue. An errored `read_dir` entry propagates via `?` and fails the whole enumeration into `enumerated: None`, because a partial list is indistinguishable from a short one.

That is simpler than what it replaces — no `Palace` structs, no metadata decode on the enumeration path — and it closes the gap by construction instead of by a second check that could drift.

**Mutation proof** — restored `list_palaces` as the enumeration source:

```
test startup_sweep_counts_an_undecodable_palace_instead_of_skipping_it ... FAILED
  left: Some(0)   right: Some(1)
```

Exactly your reproduction. I kept your MEDIUM framing in the commit message: an undecodable `palace.json` makes the palace dark to the whole memory system, so BM25's job is to report it, not repair it, and the only *healthy*-palace case is a transient `read_dir` error.

### The zero-drawer justification

Corrected, and the correction matters more than the original. `open_with_intent` falls back to an empty table when `load_drawers` fails, and `load_drawers` skips undecodable rows — so drawers can be absent from `handle.drawers` while sitting in redb. The comment now grounds the skip in `handle.drawers` being what every lane serves from, vector included, and says explicitly that changing where the lanes read from breaks the skip.

### `tools/bm25.rs` sibling asymmetry

Fixed. `Closed` marks dirty and warns, same as `Full`. The doc comment no longer describes it as the lesser case.

**Mutation proof** — removed the new `mark_dirty`:

```
test a_closed_index_queue_queues_the_palace_for_repair ... FAILED
  left: []   right: ["default"]
```

The test swaps in a sender with a dropped receiver rather than waiting for the real worker to exit, so it closes the channel deterministically instead of racing the spawn.

### The two LOWs

- `main.rs` ordering comment deleted. It justified a dependency this change removed.
- `bm25_repair.rs` now says what the design actually guarantees — an aborted pass loses no work, because entries leave the queue only on verified coverage — and records the disarm-for-process-lifetime risk as unguarded hardening with no identified panic source.

**And your correction is right:** there is no `catch_unwind` anywhere in this crate. My "catches a panicking pass" was the snapshot-not-drain design described wrongly. The claim is retracted; nothing swallows a panic into a degraded sweep.

### On the inertness obligation

Agreed, and recorded on the piece-2 issue: before anyone calls the lane working, a real log line with non-zero `enumerated` against the ~99-palace root. Noted there that `load_palaces_from_disk()` already hydrates every palace immediately before the sweep spawns, so serial `open_palace` is warm-cache and the cost objection does not apply.

### Gates — rung 5

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` (minus 3 GUI crates) | clean |
| `cargo clippy -p trusty-memory\|trusty-bm25-daemon --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-memory` | 684 passed, 0 failed |
| `cargo test -p trusty-bm25-daemon` + `-p trusty-common --features memory-core` | 964 passed, 0 failed |
| `--include-ignored` (bm25 e2e + supervisor e2e + concurrency) | 8 passed, 0 failed |
| `check_test_pointers.sh` | 22209 resolved, 0 dangling |
| `check_line_cap.sh` | 3755 files, 0 violations |
| `check_changelog_fragment.sh` | all 3 crates recorded |

Eight mutations reproduced across the three rounds. Lane still off by default.

Issue #3331, opened 2026-07-19
fix(trusty-console): add trusty-agents to the proxy allowlist


bobmatnyc:
Part of #3328.

**Evidence**: `proxy/routes.rs:53` — `full_id()` covers search/memory/analyze/review/mpm but NOT agents. The agents surface is unreachable through the console.

**Fix**: Add trusty-agents to the proxy allowlist. (Small change.)

PR #5051, opened 2026-08-07
fix(trusty-agents): pin the loopback bind default with a regression test; correct stale 0.0.0.0 docs

### LINKED ISSUE #3329: fix(trusty-agents): api server binds 0.0.0.0 unauthenticated by default — default to loopback + require token for non-loopback
Part of #3328.

**Evidence**: `api/server/routes.rs:271` binds `0.0.0.0` unconditionally; `auth.rs:3` doc admits LAN reachability; default `serve()` is `ApiConfig::unauthenticated` (routes.rs:213-215).

**Fix**:
1. Default bind to loopback (`127.0.0.1`).
2. Add explicit `--bind` opt-in requiring `--api-token` for non-loopback binds.
3. Adopt `trusty_common::server::with_guarded_middleware` (same shape as #3317 PR B).

**Subsumes** architecture-review T1 item 8.

bobmatnyc:
## The reported exposure is already fixed; the docs saying otherwise are not

trusty-agents was reported as binding `0.0.0.0:7654` unauthenticated. It does not. [#3341](https://github.com/bobmatnyc/trusty-tools/pull/3341) landed the loopback default on 2026-07-19 and closed [#3329](https://github.com/bobmatnyc/trusty-tools/issues/3329). Current `origin/main`:

- `ApiConfig::unauthenticated` binds `127.0.0.1`; both CLI paths default to loopback.
- `serve_with_config` **refuses to start** on a non-loopback bind without `--api-token`, before opening a socket (`api/server/routes.rs`).
- The router-wide `SelfOrigins` write guard is applied via the shared `trusty_common::server::with_guarded_middleware`.
- `agents` is in the console proxy allowlist (`full_id("agents") => Some("trusty-agents")`, #3331).

Verified on this machine: nothing listens on 7654, and every running daemon binds loopback.

```
trusty-co  3804 TCP 127.0.0.1:7788 (LISTEN)
trusty-mp 28532 TCP 127.0.0.1:7880 (LISTEN)
trusty-se 30736 TCP 127.0.0.1:7878 (LISTEN)
trusty-an 36061 TCP 127.0.0.1:7879 (LISTEN)
trusty-me 98860 TCP 127.0.0.1:7070 (LISTEN)
```

Two things were genuinely wrong, and this PR fixes both.

### 1. The default was untested

`non_loopback_without_token_refuses_start` covers the refusal branch — which only fires once something has already chosen a non-loopback address. Nothing pinned the default itself. Proof: reintroducing `0.0.0.0` as the default leaves all six pre-existing guard tests green.

```
test result: FAILED. 6 passed; 2 failed
  api::server::tests::guard::unauthenticated_config_binds_loopback
    panicked: the unauthenticated default must bind loopback, got 0.0.0.0
  api::server::tests::guard::unspecified_bind_defaults_to_loopback
    panicked: omitting --bind must resolve to loopback
```

Both new tests fail against the pre-fix state and pass against this branch. `--bind` now parses into `Option<IpAddr>` and resolves through `ApiConfig::with_bind`, so the raw-argv path in `runtime::startup` (the Tauri sidecar's PM-state bypass) and the clap path in `runtime::mode_dispatch` cannot drift apart on a security-relevant default.

### 2. The docs still described the vulnerable state

`auth.rs:3` and `auth.rs:93` asserted as fact that "the server binds `0.0.0.0`, so any process on the LAN can reach the REST API + UI", and that bearer auth is what gates it. Both false since #3341 — the bind is the control, and on the loopback default the auth middleware is **not installed at all**. That stale text is the evidence quoted in #3329 and is what caused this to be re-reported as a live exposure. `docs/reference/threat-model.md` carried the same drift: its trusty-agents row said `0.0.0.0:7654`, origin guard "No", allowlist "No" — all three wrong — plus a paragraph calling the fix "still in progress".

## What an unauthenticated caller can actually reach

Asked explicitly, so stated in full.

**On the default config (loopback, no token), a local caller reaches all 45 routes.** `auth_middleware` is only layered in when a token is set, so with `token: None` there is no authentication anywhere. That includes:

| Route | What it grants |
|---|---|
| `POST /api/task` | Spawns arbitrary subprocesses — code execution as the user. The repo's own guard docs call it "DESTRUCTIVE" |
| `POST /rpc` | The full JSON-RPC/MCP tool surface |
| `POST /api/tm/sessions/{name}/send`, `/api/tm/tell` | Injects text into live trusty-mpm sessions — code execution by proxy |
| `POST /api/ctrl/sessions`, `.../attach` | Create/attach CTRL sessions |
| `DELETE /api/task/{id}`, `DELETE /api/tasks`, `POST /api/clear-context` | Destroy in-flight and historical work |
| `PATCH /api/agents/{name}/permissions`, `/persona`, `/skills`, `/stores` | Mutate agent permission ceilings |
| `GET /api/tm/sessions/{name}/pane` | Reads live terminal pane content |
| `GET /api/agents/{name}/knowledge`, `/kg/*` | Agent knowledge-graph contents |

Two controls stand between that surface and an attacker, and only one of them is load-bearing:

1. **The loopback bind.** Load-bearing. An attacker must already have local code execution, at which point they run as the user anyway. This is the standard same-host daemon posture, identical to trusty-search/memory/analyze/review/mpm.
2. **The write-origin guard.** Not an authn control, and should not be read as one. It is method-gated (writes only) and **fails open on a missing `Origin` header** by design, so it stops a malicious *web page* the operator visits from driving writes cross-origin (CSRF), but does nothing against `curl` or a malicious npm postinstall script running locally.

**Severity: Low as it stands.** Pre-#3341 this was Critical — unauthenticated RCE from any host on the LAN. Today, exploitation presupposes the local code execution that would make it redundant.

**One residual gap, unchanged by this PR and reported rather than fixed:** `GET /api/health`, `/api/config`, and `/api/events` stay exempt from auth even when a token *is* set (`auth.rs`). `/api/events` is the SSE telemetry stream, exempt because browser `EventSource` cannot attach headers. On the loopback default that is local-only and uninteresting. On the token-guarded `--bind` opt-in it means the telemetry stream is readable by any host that can route to the port. Deliberate and documented — the in-code guidance is to front that deployment with a reverse proxy — but worth naming since it is the one thing a token does not cover.

## Remote access: option (a), through the console proxy

Asked to choose between (a) loopback-only behind trusty-console's proxy and (b) a per-daemon Tailscale path. **(a)**, and it is already built:

- [ADR-0018](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0018-loopback-only-doctrine.md) makes trusty-console the sole off-loopback surface and the only component that widens over Tailscale. A Tailscale binder inside trusty-agents contradicts it directly.
- [ADR-0031](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0031-uds-for-inter-crate-transport-http-for-external.md) routes all external comms through the one shared HTTP server. Same conclusion.
- The plumbing exists: trusty-agents writes its bound address to the standard `http_addr` discovery file, `agents` is in the proxy allowlist, and the startup refusal message already points operators at `/api/agents/*`. Option (b) would build a second remote path duplicating a working one.
- Nothing needs (b). trusty-agents is unreleased (0.38.6, no tags, not on crates.io), is not running here, and its only direct consumer — `agents-ui` — is a Tauri app whose writes travel over Tauri IPC, never HTTP.

The `--bind` escape hatch stays: explicit opt-in, never the default, refused without a token, and now documented in the threat model as required.

## Gates — rung 5

```
cargo check -p trusty-agents --all-targets             EXIT=0
cargo clippy -p trusty-agents --all-targets -D warnings EXIT=0
cargo test -p trusty-agents      test result: ok. 3377 passed; 0 failed; 11 ignored
                                 (+ 10 further suites, all ok, 0 failed)
cargo test -p trusty-agents-local EXIT=0 (0 tests — crate has none; compiles against the change)
scripts/check_line_cap.sh        3745 files, 0 violations
scripts/check_sld.sh             56 specs + 4093 files, 0 errors, 0 warnings
scripts/check_changelog_fragment.sh  all crates with source changes recorded
```

`cargo check --workspace` is not run: it fails on `trusty-code-gui`/`trusty-mpm-gui` for unbuilt `ui/dist`. This diff touches neither — it is confined to `crates/trusty-agents/` and `docs/reference/threat-model.md`. `ApiConfig` is used nowhere outside trusty-agents, and the only crate depending on the package (`trusty-agents-local`) builds and tests clean; the change to `ApiConfig` is additive.

Refs #3329

### DIFF
--- a/crates/trusty-agents/changelog.d/3329-loopback-bind-regression-guard.md
+++ b/crates/trusty-agents/changelog.d/3329-loopback-bind-regression-guard.md
@@ -1,0 +1,21 @@
+Fixed
+
+- Pinned the API server's loopback bind default with a regression test and
+  collapsed the duplicated default into one place
+  (follow-up to [#3329](https://github.com/bobmatnyc/trusty-tools/issues/3329)).
+  The `0.0.0.0` bind itself was fixed in
+  [#3341](https://github.com/bobmatnyc/trusty-tools/pull/3341); the suite
+  covered only the *refusal* of a non-loopback bind without a token, so nothing
+  would have caught the default itself being flipped back — verified by
+  reintroducing `0.0.0.0` and watching the two new tests fail while all six
+  pre-existing guard tests stayed green. `--bind` is now parsed into an
+  `Option<IpAddr>` and resolved through the new `ApiConfig::with_bind`, so the
+  raw-argv path in `runtime::startup` and the clap path in
+  `runtime::mode_dispatch` can no longer drift apart on a security-relevant
+  default.
+- Corrected the `auth.rs` module and `auth_middleware` doc comments, which
+  still stated as fact that the server binds `0.0.0.0` and that bearer auth is
+  what keeps it off the LAN. Both have been false since #3341 — the bind is the
+  control, and on the loopback default the auth middleware is not installed at
+  all. This stale text was the evidence cited in the original report and caused
+  the exposure to be re-investigated as if it were still live.
--- a/crates/trusty-agents/src/api/server/auth.rs
+++ b/crates/trusty-agents/src/api/server/auth.rs
@@ -1,11 +1,14 @@
 //! Server configuration + bearer-token auth middleware (#181).
 //!
-//! Why: The server binds `0.0.0.0`, so any process on the LAN can reach the
-//! REST API + UI. Bearer-token auth, configurable per launch, gates the
-//! sensitive `/api/*` routes while exempting health/config/events probes and
-//! the embedded UI assets.
-//! What: `ApiConfig` carries the port + optional token; `auth_middleware`
-//! enforces the token; `ApiClientConfig` tells the UI whether auth is needed.
+//! Why: The server binds loopback (`127.0.0.1`) by default, so auth is not the
+//! control that keeps it off the LAN — the bind is (#3329). Bearer-token auth
+//! covers the one case the bind cannot: an operator who explicitly opts into a
+//! non-loopback `--bind`, which `serve_with_config` refuses to start without a
+//! token. It gates the sensitive `/api/*` routes while exempting
+//! health/config/events probes and the embedded UI assets.
+//! What: `ApiConfig` carries the bind + port + optional token;
+//! `auth_middleware` enforces the token; `ApiClientConfig` tells the UI whether
+//! auth is needed.
 //! Test: `auth_middleware_*` and `config_endpoint_*` in `super::tests`.
 
 use axum::{
@@ -57,6 +60,26 @@ impl ApiConfig {
             token: None,
         }
     }
+
+    /// Build a config from an optional operator-supplied bind. (#3329)
+    ///
+    /// Why: `--bind` is parsed in two independent places — the `--api` early
+    /// dispatch in `runtime::startup` (raw argv, so the Tauri sidecar can skip
+    /// PM state init) and the clap path in `runtime::mode_dispatch`. Both must
+    /// apply the same loopback default; an `unwrap_or(LOCALHOST)` open-coded at
+    /// each site leaves a security-relevant default free to drift between them.
+    /// `serve_with_config` still enforces the doctrine whatever this returns,
+    /// so this is defence in depth rather than the primary control.
+    /// What: `None` → `127.0.0.1`. `Some(ip)` is taken verbatim; a non-loopback
+    /// `ip` is then refused by `serve_with_config` unless `token` is `Some`.
+    /// Test: `unspecified_bind_defaults_to_loopback`.
+    pub fn with_bind(bind: Option<std::net::IpAddr>, port: u16, token: Option<String>) -> Self {
+        Self {
+            bind: bind.unwrap_or_else(|| std::net::Ipv4Addr::LOCALHOST.into()),
+            port,
+            token,
+        }
+    }
 }
 
 /// Wrapper used by the auth middleware so axum can extract the optional
@@ -90,10 +113,13 @@ pub(super) fn bearer_token_matches(expected: &str, provided: &str) -> bool {
 
 /// Bearer-token authentication middleware. (#181)
 ///
-/// Why: The server binds `0.0.0.0`, so any process on the LAN can reach the
-/// REST API + UI. When the operator sets a token, we reject requests that
-/// don't present it. We deliberately exempt `GET /api/health` so probes from
-/// load balancers or healthchecks don't need credentials, and exempt the
+/// Why: On the loopback default this middleware is not installed at all — a
+/// local caller reaches every route, which is the accepted posture for a
+/// same-host daemon (#3329). It exists for the non-loopback `--bind` opt-in,
+/// where `serve_with_config` requires a token: there, any host that can route
+/// to the port must present it. We deliberately exempt `GET /api/health` so
+/// probes from load balancers or healthchecks don't need credentials, and
+/// exempt the
 /// embedded UI's static assets so a browser can load `index.html` and obtain
 /// the token via `/api/config` before issuing authenticated requests.
 /// What: For requests under `/api/*` (other than `/api/health`), checks
--- a/crates/trusty-agents/src/api/server/tests/guard.rs
+++ b/crates/trusty-agents/src/api/server/tests/guard.rs
@@ -153,3 +153,46 @@ async fn non_loopback_without_token_refuses_start() {
         "refusal should point at the console proxy as the intended remote path, got: {msg}"
     );
 }
+
+/// Why: #3329's control is the DEFAULT bind, not the refusal — the refusal only
+/// fires once something has already chosen a non-loopback address.
+/// `non_loopback_without_token_refuses_start` covers that branch; nothing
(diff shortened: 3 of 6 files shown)


bobmatnyc:
## Verdict: APPROVE

Reviewed against `origin/main` (`b8746f59d`), not the PR description. Ran the crate's full test suite, `fmt --check`, and `clippy -D warnings`, and independently mutation-tested the two new tests by reverting the default to `0.0.0.0` in a scratch copy of `auth.rs`.

### Documentation correctness — verified against code, not claims

- `serve_with_config` (`crates/trusty-agents/src/api/server/routes.rs:356-369`): the `!cfg.bind.is_loopback() && cfg.token.is_none()` check is the *first* statement in the function, before the docs-index spawn, before `TcpListener::bind`, before anything else — refuses before opening a socket, as the PR claims.
- Default is genuinely `127.0.0.1`: `ApiConfig::unauthenticated` (auth.rs:58) and `ApiConfig::with_bind(None, ...)` (auth.rs:78) both resolve to `Ipv4Addr::LOCALHOST`.
- `agents` is in the console allowlist: `crates/trusty-console/src/proxy/routes.rs:67` — `"agents" => Some("trusty-agents")`.
- Router-wide guard confirmed: `build_router_with_origins` → `with_guarded_middleware(router, self_origins)` (routes.rs:318), using `trusty_common::server::{SelfOrigins, with_guarded_middleware}`.
- Threat-model row (`docs/reference/threat-model.md`): port 8080 matches `cli.port.unwrap_or(8080)` (mode_dispatch.rs:207); the "7654 passed explicitly" claim matches the Tauri sidecar comment at mode_dispatch.rs:476 (`--api --port 7654`); guard/allowlist columns match the code above. ADR-0018 and ADR-0031 both exist at the linked paths. No new false statement introduced.

### Test coverage — mutation-verified independently

Reverted `auth.rs`'s two default sites to `Ipv4Addr::UNSPECIFIED` in a throwaway copy and ran `cargo test -p trusty-agents --lib api::server::tests::guard`:

```
test api::server::tests::guard::unauthenticated_config_binds_loopback ... FAILED
test api::server::tests::guard::unspecified_bind_defaults_to_loopback ... FAILED
test api::server::tests::guard::non_loopback_without_token_refuses_start ... ok
test api::server::tests::guard::cross_origin_write_rejected_403 ... ok
test api::server::tests::guard::cross_origin_get_read_unaffected ... ok
test api::server::tests::guard::missing_origin_write_allowed ... ok
test api::server::tests::guard::loopback_write_allowed ... ok
test api::server::tests::guard::tauri_localhost_origin_allowed ... ok

test result: FAILED. 6 passed; 2 failed
```

Exactly the claimed shape — the new tests exercise the actual default, not a reimplementation of it (unlike the #5048 pattern flagged in the task). Restored the file; `git diff` clean afterward.

Full crate suite on the unmutated branch: `cargo test -p trusty-agents --lib` → `3377 passed; 0 failed; 11 ignored`. `cargo fmt --check -p trusty-agents` and `cargo clippy -p trusty-agents --all-targets -- -D warnings` both clean.

### `ApiConfig::with_bind` consolidation

`grep -rn "ApiConfig::" crates/trusty-agents/src` shows exactly three call sites: `startup.rs:393` (raw-argv path) and `mode_dispatch.rs:218` (clap path) both go through `with_bind`; `routes.rs:337`'s `serve` uses the separate `unauthenticated(port)` convenience constructor, which never parses `--bind` at all so it isn't a fourth drift point. `cli_def.rs:175`'s `pub(super) bind: Option<std::net::IpAddr>` matches `with_bind`'s signature. No bare `ApiConfig { bind, port, token }` struct literal remains outside the struct/impl definitions and tests. No third path found anywhere in the workspace.

### Residual `/api/events` auth gap — assessed, not fixed here

Confirmed in `auth.rs:148-154`: `/api/health`, `/api/config`, `/api/events` are exempt from the bearer check unconditionally, pre-existing since #192 Phase B, untouched by this PR.

Partial disagreement with the "telemetry" framing: `Event` (`crates/trusty-agents/src/events.rs:56+`) carries `PmThinking.text`, `AgentMessage.text`, `PmDelegating.task_preview` — actual conversation/task content, not health-style metrics. On a `--bind`+token deployment an operator could reasonably assume the token covers this stream and it doesn't. That pushes the *content sensitivity* above "boring telemetry," but the reachability gate is unchanged from today: it requires the same non-default `--bind` opt-in this PR's own guard already restricts to token-holders for every other route. Net: Low-to-Medium, not Low-and-done — worth a sharper code comment at minimum, but the fix (header-less SSE auth, e.g., a signed query-param token or reverse-proxy requirement) is real design work, not a two-line change, and unrelated to the loopback-default regression this PR closes. Recommend **Promote** as a separate issue rather than folding into this PR — consistent with the repo's fix-here-or-drop rule for genuinely separable work.

### Other checks

- `closingIssuesReferences` on #5051 lists only #3329 (already closed, branch-name inference, harmless) — confirmed no other issue gets auto-closed.
- `crates/trusty-memory/ui/package-lock.json` is absent from `git diff origin/main...HEAD --name-only`; not in the final diff.

No CRITICAL or HIGH findings. No findings table — nothing cleared the 80% confidence bar as a defect in the diff itself.

APPROVED for next pipeline stage.

Issue #4088, opened 2026-07-27
trusty-common: prevent patch-level SemVer breaks from bricking dependents' cargo install


bobmatnyc:
Incident: #4079 (`cargo install trusty-analyze` fails with E0063) / PR #4082 (releases 0.7.4).

**The root cause is a SemVer break by `trusty-common`, not a `trusty-analyze` bug.** PR #4082 fixes
the symptom for one dependent and leaves the trap armed for the next one. This issue tracks
disarming it.

---

## 1. Root-cause chain

`trusty-common` 0.22.5 added a **required public field**:

- `pub no_spawn_hint: Option<String>` — `crates/trusty-common/src/mcp/daemon_bridge.rs:117`
- commit `e993c18a` (*fix(trusty-mpm): launchd-aware bridge no-spawn + /health supervised flag (closes #2486) (#2491)*)
- published as a **PATCH** bump: 0.22.4 (2026-07-09) → 0.22.5 (2026-07-13)

`DaemonBridgeConfig` (`daemon_bridge.rs:62`) has **neither `#[non_exhaustive]` nor a `Default`
impl**, so any cross-crate struct literal that omits the new field is a hard compile error (E0063).

The collision:

| | |
|---|---|
| `trusty-analyze` 0.7.3 published | **2026-07-07** |
| its declared dep | `trusty-common = "0.22.0"` → caret range `[0.22.0, 0.23.0)` |
| `trusty-common` 0.22.5 published | **2026-07-13** — six days later |

A fresh `cargo install` has **no lockfile**, so it re-resolves to the newest in-range `0.22.x` and
pairs **pre-field source with post-field dependency**. Users could not install the crate at all.

Verified against crates.io at filing time: `trusty-analyze` 0.7.3 does declare `^0.22.0`.

---

## 2. The unmitigated recurrence risk — the actual point of this issue

`trusty-common` has now **demonstrated** patch-level public-field additions. The following
dependents sit on wide caret floors that would resolve exactly the same way. Floors verified from
published crates.io metadata, not from workspace paths:

| Dependent | Published version | `trusty-common` req |
|---|---|---|
| `trusty-console` | 0.4.0 | `^0.22.0` |
| `trusty-code` | 0.2.0 | `^0.23.2` |
| `trusty-bm25-daemon` | 0.1.3 | `^0.22.0` |

**Any one of these can be bricked by the next such change, with no code change of its own.** The
[...]

Issue #702, opened 2026-06-03
Upgrade redb 2.6 → 4.x workspace-wide (+ graceful old-format handling)


bobmatnyc:
## Summary
Upgrade the `redb` dependency from 2.6 to the latest 4.x (4.1.0) across the trusty-tools workspace, fix all API breakage, and handle the on-disk format incompatibility gracefully.

## Why now
redb 3.0 dropped support for the v2 on-disk format; redb 4.x cannot open a redb-2.x file (`Database::open` returns an upgrade/format error). This is the load-bearing risk and connects to:
- #697 — cross-version index unreadable
- #694 / #601 — false-healthy on corpus-open failure (daemon reports `status=ready`/healthy-empty when the corpus failed to open)

## Pre-flight (cleared)
- Latest redb: **4.1.0**, MSRV **1.89** (≤ our workspace floor of 1.91). Not a blocker.
- Breaking deltas (3.0 + 4.0): `RedbKey`→`Key`, `RedbValue`→`Value`; `begin_read`/`cache_stats` moved to `ReadableDatabase` trait; lifetimes removed from txns/read tables; `drain()`/`drain_filter()` removed (→ `retain`/`extract_if`); `check_integrity()` returns `DatabaseError`; `Durability::Paranoid` removed (→ `set_two_phase_commit(true)`); `AccessGuardMut`/`AccessGuardMutInPlace` now `Drop`; `Legacy` type removed. File format v2 unreadable.

## Scope
**Consuming crates:** trusty-search, trusty-common (memory-core), trusty-analyze, trusty-review, open-mpm, trusty-memory.

**PART 3 — graceful old-format handling (the point):**
- Detect format/version-mismatch errors at `Database::open`.
- Never panic; never report `status=ready`/healthy-empty (the #601/#694 false-healthy bug).
- trusty-search corpus (`index.redb`): on unreadable old-format → WARN/ERROR, back up old file (`.redb.v2-incompatible`), create fresh v4 corpus, surface `degraded`/`rebuilding` (not `ready`).
- Other stores (memory-core palaces, trusty-analyze `facts.redb`, trusty-review dedup): detect → clear error/log → recreate-empty or documented recovery, never silent-healthy.

The redb file-format version is distinct from the app schema-version (`_meta` `schema_version`); the format check happens at `Database::open` before app migrations run.

Issue #67, opened 2026-05-23
trusty-analyze: list_facts p99 latency spike (~900ms) suggests redb transaction stall


bobmatnyc:
## Bug

`list_facts` has a p50 latency of 0.25ms but a p99 of ~897ms — a 3500× spike. The tool is a simple read from the facts store (redb). Under stdio harness load (N=50), one in ~50 calls stalls for nearly a second.

## Observed in stdio_harness.rs
```
list_facts  p50: 0.25ms  p95: 1.04ms  p99: 896.50ms  mean: 18.25ms
```

## Suspected Root Cause
A write transaction elsewhere (e.g. from a concurrent `upsert_fact` or background compaction) holds the redb write lock, blocking the read. redb's MVCC should allow concurrent reads — if a read is blocking on a write lock, it suggests the facts store is opened with a write transaction when a read-only transaction would suffice.

## Fix
Audit `list_facts` implementation: ensure it opens a **read-only** transaction (`ReadTransaction`) rather than a write transaction. If the stall is from lock contention with a write path, add a timeout or use `begin_read()` exclusively for read operations.

bobmatnyc:
Closed by commit 93a192b57be58595ce99a97159bf6378d86df602

Issue #5050, opened 2026-08-07
The workspace has no public-API/SemVer gate


bobmatnyc:
Defect: no `cargo-semver-checks` or equivalent anywhere — confirmed by empty grep across `.github/workflows/`, `scripts/`, and `docs/`. Breaking changes reach crates.io undetected.

Evidence: #4088 — trusty-analyze 0.7.3 was yanked after a SemVer break shipped that CI did not catch.

Resolution: add a public-API diff gate to CI, running per-crate against the last published version. Required before the trusty-review 1.0.0 and trusty-analyze version bumps now planned.

PR #5054, opened 2026-08-07
fix(trusty-analyze): persist SCIP graph overlays and make absent-vs-empty explicit

### LINKED ISSUE #5049: trusty-analyze loses SCIP graph overlays on restart
Defect: `AnalyzerAppState::scip_overlays` in `service/events.rs` is an in-memory `HashMap<String, KgGraph>` with no disk backing. `POST /indexes/{id}/scip` populates it; `GET /indexes/{id}/graph` reads it. A daemon restart between the two silently discards the ingest.

Evidence: no redb or other persistence for that field anywhere in the crate; contrast `FactStore`, which is redb-backed.

Resolution: persist the overlay, or reject graph queries against an un-ingested index with a distinguishable error instead of returning an empty graph. Silent emptiness is the defect.

bobmatnyc:
Closes #5049

## The defect

`POST /indexes/{id}/scip` wrote into `AnalyzerAppState::scip_overlays`, an
`Arc<RwLock<HashMap<String, KgGraph>>>` with no disk backing, and answered
HTTP 200. A daemon restart discarded the ingest, and `GET /indexes/{id}/graph`
then served a tree-sitter-only graph that no caller could tell apart from one
where the overlay had been applied.

## Why persistence, not fail-loud

The issue offered both. I took persistence, for two reasons.

**Fail-loud as stated is not implementable without breaking the primary use
case.** "Reject a graph query against an un-ingested index" would reject every
`/graph` request for an index that was never SCIP-ingested — which is the
normal case. Tree-sitter-only graphs are the product's main path (the whole
`AnalyzerRegistry` + 15 language adapters exist to produce them); SCIP is an
optional precision upgrade. There is no signal at query time that says an
overlay was *expected*, so the endpoint cannot distinguish "should have had
SCIP" from "correctly has none".

**Fail-loud does not fix the data loss.** The user's uploaded SCIP index is
still gone after a restart; they'd get an error instead of silence. A SCIP index
is uploaded by the operator and cannot be re-derived from the corpus — unlike
`FactStore`'s facts, which recover by being re-derived from source. Losing one
loses real work, and only durability recovers it.

The crate already carries the pattern: `redb` is a direct dependency and
`FactStore` is the local precedent, so this is ~140 lines in the shape the crate
already uses.

## Persistence alone is not enough

Storing the overlay stops the loss but not the silence. An index nobody ingested
and an index whose SCIP payload carried zero symbols both contribute zero nodes
to `/graph`. Two surfaces now separate them, both reading the same
`ScipOverlayStore::get` → `Option`:

| Surface | No SCIP data | Ingested, zero symbols |
|---|---|---|
| `GET /indexes/{id}/scip` (new) | `404 {"error": "no SCIP overlay has been ingested for index …"}` | `200 {index_id, nodes: 0, edges, ingested_at}` |
| `GET /indexes/{id}/graph` | `x-scip-overlay: absent` | `x-scip-overlay: present` |

`/graph`'s JSON body is still a bare `KgGraph`, so every existing consumer is
untouched — the answer rides in a header instead. A failure to *read* the
overlay store is now a 500, not a silent fall-through to the tree-sitter-only
graph.

## Unreadable database: quarantine, shared with `FactStore`

The first revision refused daemon startup on an unreadable
`scip_overlays.redb`, justified by a claim that `FactStore` "recreates empty".
Review checked, and that is wrong: `FactStore::open` renames the file to
`facts.redb.v2-incompatible`, logs an ERROR, and boots. It quarantines; it does
not discard. The trade that claim supported was wrong too — redb's on-disk
format has already gone obsolete once here (#702), so the next bump would have
taken facts, smells, review, clusters, NER and the whole MCP surface down over
one optional feature's sidecar file, with `rm scip_overlays.redb` as the only
remedy.

Both stores now share one policy in `core::redb_open`, which classifies the
redb error before acting:

| Error class | Behavior | Why |
|---|---|---|
| Obsolete on-disk format (`UpgradeRequired`, `RepairAborted`, `Corrupted`, `Io(InvalidData)`) | Rename aside, log ERROR with the recovery step, start empty | The running redb cannot read it at all; moving it aside is the only option, and nothing is deleted |
| Anything else — permissions, disk, a held lock | Propagate; startup fails | The file may be perfectly good and merely unavailable; recreating on top of it would destroy data |

A shared implementation rather than a second copy is the point: two classifiers
would drift into giving opposite answers to the same byte-level cause, which is
exactly the divergence review caught.

Quarantining is safe here specifically because of the other half of this PR.
After one, `GET /indexes/{id}/scip` answers 404 — an honest report that no
overlay is available, rather than a graph that silently lost one.

Verified live, reproducing review's own scenario against the built binary and a
real `trusty-search`:

```
=== corrupt scip_overlays.redb, healthy facts.redb ===
DAEMON: BOOTED (still running)
health:              HTTP 200
GET /indexes/x/scip: {"error":"no SCIP overlay has been ingested for index x"} [HTTP 404]
files: facts.redb scip_overlays.redb scip_overlays.redb.quarantined

=== same corruption applied to facts.redb instead ===
DAEMON: BOOTED (still running)
health:              HTTP 200
files: facts.redb facts.redb.v2-incompatible scip_overlays.redb
```

The ERROR an operator sees now names what to redo:

```
ERROR SCIP overlay redb is in an unreadable/obsolete format; quarantined it and
      created a fresh empty store
      path=…/scip_overlays.redb backup=…/scip_overlays.redb.quarantined
      error=I/O error: invalid data
      recovery=re-ingest SCIP (POST /indexes/{id}/scip) for every affected index;
               until then GET /indexes/{id}/scip reports 404 for them
```

Durability is unaffected by the rewrite — same ingest / `kill -9` / re-read run,
with the identical `ingested_at` across two processes proving the second read
came off disk:

```
--- BEFORE ingest ---                {"error":"no SCIP overlay …realidx"} [404]
--- POST /indexes/realidx/scip ---   {"index_id":"realidx",…,"kg_nodes":0} [200]
--- AFTER ingest, same process ---   {"index_id":"realidx","nodes":0,"ingested_at":1786075547} [200]
--- kill -9, brand new process ---   {"index_id":"realidx","nodes":0,"ingested_at":1786075547} [200]
--- never ingested ---               {"error":"no SCIP overlay …never-touched"} [404]
```

## Other review fixes

- The redb calls in the graph handlers now run on `spawn_blocking`, the
  convention `handlers/facts.rs` established after #67. The `put` is the
  expensive one — it serializes a whole-repository `KgGraph` and fsyncs it.
- `ScipOverlayStore::put` returns `()`; its doc claimed the caller used the
  returned record, and no caller did.
- Documented that `x-scip-overlay` describes the overlay's existence, so a
  `?language=` filter that removes every SCIP node still reports `present`.

## Breaking library API — flagging for `local-ops`

`trusty-analyze` is published (0.8.0, with 0.7.3 yanked for a SemVer accident,
#4088). This PR breaks the library surface:

- `AnalyzerAppState::scip_overlays` changed type from
  `Arc<RwLock<HashMap<String, KgGraph>>>` to `core::ScipOverlayStore`.
- `AnalyzerAppState::new` and `::with_registry` take the store as a **required**
  argument rather than a `with_*` override — precisely so no caller can end up
  with a non-durable overlay store by omission, which is the bug.

That is a minor-version bump (0.8.0 → 0.9.0) under 0.x SemVer. **I have not
touched `version` in `Cargo.toml`** — the bump goes to `local-ops`.

The HTTP surface is additive only: one new `GET` route and one new response
header. No crate in the workspace depends on `trusty-analyze` (the five
`Cargo.toml` hits are comment mentions), so this is a leaf change.

## Test ladder — rung 3, plus the public-API note

Localized behavior inside one crate. Gates, all run with `SKIP_UI_BUILD=1` so
the build script cannot churn tracked `ui/dist` artifacts:

```
cargo fmt --check                                       — clean
cargo check -p trusty-analyze --all-targets             — clean
cargo clippy -p trusty-analyze --all-targets -D warnings — zero warnings
cargo test -p trusty-analyze                            — 454 + 25 + 2 + 2 passed, 0 failed
bash scripts/check_line_cap.sh                          — 3749 files, 0 violations
bash scripts/check_test_pointers.sh                     — 22148 citations, 0 dangling
bash scripts/check_sld.sh                               — 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh                — OK
```

`cargo check --workspace` is red on `main` for the two GUI crates' unbuilt
`ui/dist`. This diff touches neither:

```
$ git diff --name-only origin/main...HEAD | grep -Ei 'gui|ui/'
NONE
```

## Mutation proof

**The persistence fix.** Reverted `ScipOverlayStore`'s internals to the pre-fix
mechanism — a per-store `HashMap`, no disk — keeping the public API identical.
Four tests went red:

```
test core::scip_overlay::tests::overlay_round_trips_across_reopen ... FAILED
test core::scip_overlay::tests::corrupt_db_fails_open_loudly ... FAILED
test service::tests::graph_marks_scip_overlay_present_after_ingest ... FAILED
test service::tests::scip_overlay_survives_state_rebuild ... FAILED

test result: FAILED. 449 passed; 4 failed; 0 ignored
```

The load-bearing one, ingesting through one router and querying through a second
state built over the same data directory:

```
---- service::tests::scip_overlay_survives_state_rebuild stdout ----
panicked at crates/trusty-analyze/src/service/tests.rs:504:5:
assertion `left == right` failed: SCIP overlay must survive a daemon restart;
got {"error":"no SCIP overlay has been ingested for index myidx"}
  left: 404
 right: 200
```

**The two tests that replaced `corrupt_db_fails_open_loudly`**, each killed by
the mutation it exists to catch.

Mutation A — restore the old refuse-always `open`:

```
test core::scip_overlay::tests::incompatible_overlay_db_is_quarantined ... FAILED
test core::scip_overlay::tests::unopenable_overlay_path_propagates_error ... ok
```

Mutation B — the over-correction, quarantine every error class instead of the
obsolete-format one (`is_format_obsolete` returns `true` unconditionally):

```
test core::scip_overlay::tests::incompatible_overlay_db_is_quarantined ... ok
test core::scip_overlay::tests::unopenable_overlay_path_propagates_error ... FAILED
```

Mutation B is why `unopenable_overlay_path_propagates_error` puts a *directory*
where the database belongs rather than using a missing parent: `rename` succeeds
on a directory, so the wrong branch opens cleanly and the test catches it. With a
missing path `rename` fails either way and the mutation would survive.

Both files restored from checksummed copies (`shasum` matched); the full chain
above is green afterwards.

## New tests

- `scip_overlay_survives_state_rebuild` — ingest, rebuild `AnalyzerAppState`
  over the same data dir, overlay still reported.
- `graph_marks_scip_overlay_present_after_ingest` / `..._absent_without_ingest` —
  the header, end-to-end against a local stub trusty-search that serves an empty
  chunk corpus (`/graph` fetches chunks before it reads the overlay, so the merge
  path was previously untestable with the unreachable default client).
- `empty_scip_ingest_is_distinguishable_from_no_ingest` — 200/`nodes: 0` versus
  404.
- `incompatible_overlay_db_is_quarantined` / `unopenable_overlay_path_propagates_error` —
  the two error classes, replacing the first revision's `corrupt_db_fails_open_loudly`.
- `overlay_round_trips_across_reopen`,
  `put_overwrites_prior_overlay`, `absent_overlay_reads_as_none`,
  `empty_overlay_is_distinguishable_from_absent`,
  `overlay_path_is_a_sibling_of_the_facts_store` — store unit coverage.

Also removed a now-satisfied `.test-pointer-allowlist.tsv` entry: `ingest_scip`
cited a `scip_ingest_round_trip` test that never existed, and it now cites the
real ones.

## Storage layout

`scip_overlays.redb`, a sibling of the facts store, derived via
`overlay_path_beside_facts`. No new CLI flag — `--facts-path` still governs the
whole data directory.


bobmatnyc:
## code-critic verdict: WARN

Zero CRITICAL, one HIGH. The fix is real and both halves are independently verified — the HIGH is a new availability failure the diff introduces, not a defect in the fix.

### What I verified myself, not from the PR body

**Durability across a genuine process restart** (not just a state rebuild). Built the binary, ran the real daemon, `kill -9` between ingest and query:

```
$ trusty-analyze --facts-path $D/facts.redb serve --foreground --port 38771 &
--- files after boot ---
facts.redb   scip_overlays.redb          # sibling, as designed
--- BEFORE ingest ---
{"error":"no SCIP overlay has been ingested for index realidx"} [HTTP 404]
--- POST /indexes/realidx/scip ---
{"index_id":"realidx","documents":0,...,"kg_nodes":0} [HTTP 200]
--- AFTER ingest, same process ---
{"index_id":"realidx","nodes":0,"edges":0,"ingested_at":1786074699} [HTTP 200]
--- kill -9, brand new process, same data dir ---
{"index_id":"realidx","nodes":0,"edges":0,"ingested_at":1786074699} [HTTP 200]
{"error":"no SCIP overlay has been ingested for index never-touched"} [HTTP 404]
```

The identical `ingested_at` across the two processes is the proof the second daemon read it off disk. And this run is the *hard* case: an ingest that produced zero nodes still reports 200 after a SIGKILL restart, while a never-ingested index reports 404. Absent and empty are distinguishable end-to-end, across a real restart.

**Mutation proof reproduced independently.** Reverted `ScipOverlayStore` to a per-store `HashMap<String, ScipOverlayRecord>` behind a byte-identical public API, ran the suite, restored from a checksummed copy:

```
test core::scip_overlay::tests::overlay_round_trips_across_reopen ... FAILED
test core::scip_overlay::tests::corrupt_db_fails_open_loudly ... FAILED
test service::tests::graph_marks_scip_overlay_present_after_ingest ... FAILED
test service::tests::scip_overlay_survives_state_rebuild ... FAILED
test result: FAILED. 449 passed; 4 failed; 0 ignored
```

Exactly the four claimed. `scip_overlay_survives_state_rebuild` builds two `AnalyzerAppState`s over one data directory, so it proves durability across the *state* but not across the *process*; the daemon run above closes that gap, and `overlay_round_trips_across_reopen` closes it at the store level by dropping the `Database` and reopening the file.

**Fail-open hunt — clean.** `absent` is emitted only when `get` returned `Ok(None)`. A read failure is a 500 (`graph.rs:73`, `graph.rs:395`), a write failure is a 500 before the `ScipIngested` event is emitted (`graph.rs:357`), a `fetch_chunks` failure is a 502 with no header at all — I confirmed that last one live. No branch turns "could not tell" into `absent`. `put` is one redb transaction, so there is no partial apply; a write that commits and then fails to respond leaves durable data and an idempotent overwrite on retry.

**Housekeeping.** No `ui/dist` in the diff (`git diff --name-only origin/main...HEAD | grep -Ei 'ui/'` → empty). The removed `.test-pointer-allowlist.tsv` line is replaced by citations to tests that exist; the doc-pointer lint passes in CI. Local gates: `fmt --check` + `clippy -D warnings` + `test -p trusty-analyze` → EXIT=0, 453 + 25 + 2 + 2 passed, 0 failed.

---

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-analyze/src/core/scip_overlay.rs` | 90–101 | A corrupt or format-obsolete `scip_overlays.redb` now refuses daemon startup entirely. Every unrelated function goes down with it, the error names no recovery step, and there is no escape hatch. | Classify the redb error the way `facts.rs:88-101` already does; quarantine the file aside on the format-incompatible class only, log ERROR, boot. Keep propagating transient errors. | Fix here |
| MEDIUM | `crates/trusty-analyze/src/core/scip_overlay.rs` | 17–20 | Module doc says `FactStore` "recreated empty" and calls that "silently replacing it with an empty store". It does neither. | Correct the doc to what `FactStore` does, and restate the real justification. | Fix here |
| MEDIUM | `crates/trusty-analyze/src/mcp/mod.rs` | 443 | `extract_graph` returns only the JSON body, so `x-scip-overlay` is dropped, and no MCP tool wraps `GET /indexes/{id}/scip`. The distinguishability half of the fix does not reach the crate's own MCP consumer. | Add a `scip_overlay_status` MCP tool, or fold the header into `extract_graph`'s result envelope. | Promote |
| MEDIUM | `crates/trusty-analyze/src/service/handlers/graph.rs` | 73, 357, 395 | Blocking redb calls run directly on the tokio worker. `handlers/facts.rs:41-47,92-96` wraps every `FactStore` call in `spawn_blocking` with a doc comment saying why. | Wrap at minimum the `put` at line 357 — it serializes a whole-repo `KgGraph` to JSON and fsyncs. | Fix here |
| LOW | `crates/trusty-analyze/src/core/scip_overlay.rs` | 109–110 | `put`'s doc says it "Returns the record so the handler can report it without a second read". The only caller discards it. | Drop the sentence, or have `ingest_scip` report `record.ingested_at`. | Fix here |

---

## Required changes

### 1. HIGH — the daemon now has a new hard startup dependency

I ran both stores against the same corruption (`StorageError::Io(InvalidData)`, which `facts_db_is_incompatible` classifies as recoverable):

```
=== corrupt scip_overlays.redb, healthy facts.redb ===
DAEMON_EXIT=1
Error: open SCIP overlay store at /tmp/…/scip_overlays.redb
Caused by:
    0: open SCIP overlay redb at /tmp/…/scip_overlays.redb
    1: I/O error: invalid data

=== same corruption applied to facts.redb instead ===
DAEMON_EXIT=124   (still running — it booted)
ERROR facts redb is in an incompatible/old format (redb 2.x); moved it aside
      and creating a fresh empty facts store … backup=…/facts.redb.v2-incompatible
INFO  trusty-analyze listening on http://127.0.0.1:38773
```

Three things follow.

**The stated reason for the divergence is not what `FactStore` does.** `FactStore` does not replace an unreadable database with nothing. It renames the file to `facts.redb.v2-incompatible`, logs a loud ERROR, and creates a fresh one — the bytes survive and the reset is visible. Applying that same quarantine here would preserve the uploaded SCIP index *and* keep the daemon up. The argument that actually holds is narrower: after a quarantine, `/scip` would answer 404 for the quarantined overlays, and 404 is the indistinguishable silence this PR exists to remove. That is a good argument, and it is not the one in the doc.

**The `UpgradeRequired` case is foreseeable, because it already happened here.** `facts_db_is_incompatible` exists because redb 4.x could not open a redb 2.x file (#702). When redb next bumps its on-disk format, every deployed daemon that has ever ingested SCIP refuses to start — facts, complexity, smells, review, clusters, NER and the whole MCP surface, all unavailable because one optional feature's sidecar file is a version behind. Under launchd `KeepAlive` that is a crash loop whose only symptom is a log line nobody reads.

**Recovery is the loss you were avoiding, plus an outage.** The remedy is `rm scip_overlays.redb`, which the error message does not mention and which destroys the overlays anyway.

The choice is right for a *transient* error — permissions, disk, a held lock. Refusing to start there is correct, and `FactStore` agrees (non-incompatible errors propagate). What is wrong is treating one error class as the other. `open` currently cannot tell "this file is in a format I no longer read" from "I cannot read this file right now".

Smallest change that keeps the intent:

```rust
// scip_overlay.rs — mirror facts.rs:88-101
match Database::create(path) {
    Ok(db) => …,
    Err(e) if overlay_db_is_incompatible(&e) => {
        // quarantine, don't delete: the bytes are the operator's upload
        let backup = path.with_extension("redb.quarantined");
        std::fs::rename(path, &backup)?;
        tracing::error!(path=%path.display(), backup=%backup.display(),
            "SCIP overlay redb unreadable; quarantined and starting empty — \
             re-ingest SCIP for every affected index");
        Ok(Database::create(path)?)
    }
    Err(e) => Err(e).with_context(…),   // transient: still fatal
}
```

If you'd rather keep the refusal, that is defensible — but then the error must carry the recovery step, because right now the operator gets a path and nothing else:

```rust
.with_context(|| format!(
    "open SCIP overlay redb at {}. If this file is corrupt or from an older \
     redb, move it aside and restart; every index will then need re-ingesting.",
    path.display()))?
```

I'd take the quarantine. It preserves the upload, keeps the daemon up, and leaves only the 404-vs-quarantined ambiguity — which you can close later with a `quarantined` flag on the `/scip` response, and which is a much smaller hole than a dead daemon.

Either way, `corrupt_db_fails_open_loudly` needs a sibling asserting the chosen behavior on the format-incompatible class specifically.

### 2. MEDIUM — the doc's account of `FactStore`

`scip_overlay.rs:17-20` and the mirrored text in `events.rs` will teach the next maintainer that `FactStore` destroys data on a bad open. It quarantines and logs. Fix the claim; the design decision it supports needs a different justification (see above), not a stronger adjective.

---

## Notes

**The `x-scip-overlay` header is computed before the `?language=` filter.** So a filtered response that contains no SCIP nodes still says `present`. That reads correct to me — the header describes the overlay's existence, which is what the doc says — and I am not flagging it. Worth a sentence in the header doc if you touch it anyway.

**Version bump.** Leaving `version` alone and routing 0.8.0 → 0.9.0 to `local-ops` is what CLAUDE.md requires, and 0.9.0 is the right call: under 0.x, minor is the breaking position, and `AnalyzerAppState::scip_overlays` changing type plus two constructors gaining a required argument is unambiguously breaking. I confirmed the leaf claim — every `trusty-analyze` hit in a sibling `Cargo.toml` is a comment, and the only in-workspace constructor call sites are `commands/run.rs:117` and the crate's own tests. Given #4088 and the absence of a `cargo-semver-checks` gate (#5050), the risk is not this PR but the publish that follows it: nothing mechanical will stop a 0.8.1. Worth naming the bump in the merge handoff rather than trusting the changelog fragment to be read.

**Making `scip_overlays` a required constructor argument rather than a `with_*` override is the right call** and is the part of this diff I'd least want changed. It makes the original bug — a state built without durable overlay storage — unrepresentable.

**Test quality.** The new tests fail for the right reason against the reverted mechanism, which I confirmed rather than took on faith. `spawn_empty_chunk_search` is the piece that makes the `/graph` merge path testable at all, and an always-empty corpus is the correct choice: it keeps the tree-sitter half at zero nodes, which is precisely the ambiguous response the fix is about. `graph_marks_scip_overlay_present_after_ingest` asserting both the header and the merged body across a state rebuild is doing real work.

bobmatnyc:
## code-critic verdict: APPROVE

Re-review of `3c975cfd` against my [WARN](https://github.com/bobmatnyc/trusty-tools/pull/5054#issuecomment-5212163790). The HIGH is resolved, and resolved better than what I asked for. Zero CRITICAL, zero HIGH. One MEDIUM, which is a recommendation for separate work and does not gate this PR.

---

## The HIGH — extraction verified, not taken on faith

I asked for the classification to be mirrored. The author extracted both stores onto one policy in `core/redb_open.rs` instead. That is the better call, and the risk it carries is that the store which was already correct gets altered on the way out. It did not.

**`FactStore`'s behavior is byte-for-byte preserved.** I diffed the extracted policy against the deleted `facts_db_is_incompatible` / `open_facts_db_or_recreate`: the classification arms are identical, rename-then-create is identical, and `FACTS_QUARANTINE_SUFFIX = ".v2-incompatible"` is kept deliberately so files quarantined by earlier builds stay recognisable (`facts.rs:83-86`). Only the context strings and the log message changed. Confirmed live — same corruption, same outcome as before the rewrite:

```
=== CASE B: corrupt facts.redb ===
EXIT=124 (booted and kept running)
facts.redb   facts.redb.v2-incompatible
ERROR facts redb is in an unreadable/obsolete format; quarantined it and created
      a fresh empty store  path=… backup=…/facts.redb.v2-incompatible
      error=I/O error: invalid data
      recovery=facts are re-derived automatically on the next analyzer run
```

**The classification is correct and complete for redb 4.1.** I read `redb-4.1.0/src/error.rs` rather than trusting the list. Every variant `is_format_obsolete` returns `false` for is genuinely transient — `DatabaseAlreadyOpen`, `TransactionInProgress`, `Storage(ValueTooLarge | PreviousIo | DatabaseClosed | LockPoisoned)`, and any non-`InvalidData` `Io`. Nothing format-related is missing. `Io(InvalidData)` is produced at `page_manager.rs:167,172`, which is the header check, so it is genuinely a format signal.

One variant deserved a second look. `RepairAborted`'s rustdoc says "repair was aborted for another reason (such as the database being read-only)", which would make it transient. I traced every construction site: `db.rs:456` is inside `ReadOnlyDatabase::new`, and `db.rs:822/840/859/929` plus `page_manager.rs:236` all require a user `repair_callback` that calls `abort()`. `Database::create` uses neither path, so `RepairAborted` is unreachable here. Classifying it as obsolete is harmless, and it is inherited unchanged from the code that already shipped.

**The destructive branch does not fire on an unclean shutdown.** This was the regression I most wanted to rule out — adding a quarantine arm means a hard kill could start moving the operator's SCIP data aside on every restart. It does not:

```
--- ingest, kill -9, restart new process ---
{"index_id":"realidx","nodes":0,"edges":0,"ingested_at":1786076109} [200]
{"index_id":"realidx","nodes":0,"edges":0,"ingested_at":1786076109} [200]
--- quarantine files created? ---
none — clean restart did not quarantine
```

Identical `ingested_at` across two processes, so durability survived the rewrite.

**Both recovery arms behave correctly in the built binary.**

```
=== CASE A: corrupt scip_overlays.redb, healthy facts.redb ===
daemon booted: yes
{"status":"ok","version":"0.8.0","search_reachable":true} [200]
{"error":"no SCIP overlay has been ingested for index realidx"} [404]
scip_overlays.redb   scip_overlays.redb.quarantined
quarantined bytes preserved: 4096 bytes
ERROR SCIP overlay redb is in an unreadable/obsolete format; quarantined it and
      created a fresh empty store … recovery=re-ingest SCIP (POST /indexes/{id}/scip)
      for every affected index; until then GET /indexes/{id}/scip reports 404 for them

=== CASE C: directory in place of the db — transient must stay fatal ===
EXIT=1
Error: open SCIP overlay store at /tmp/…/scip_overlays.redb
Caused by:
    0: … The file is present but could not be opened (permissions, disk, or a lock
       held by another process); this is not a format problem, so it was left untouched
    1: I/O error: Is a directory (os error 21)
path untouched: yes-still-a-directory
quarantine files created (must be 0): 0
```

The quarantine preserves all 4096 bytes rather than truncating, the `recovery` field names the redo step, and the transient error message now says why it was *not* recovered. That last sentence is the one an operator actually needs.

---

## The two replacement tests — both mutations reproduced

**Mutation 1** — restored refuse-always `open` on the overlay store:

```
test core::scip_overlay::tests::incompatible_overlay_db_is_quarantined ... FAILED
test core::scip_overlay::tests::unopenable_overlay_path_propagates_error ... ok
test result: FAILED. 453 passed; 1 failed
```

**Mutation 2** — `is_format_obsolete` returns `true` unconditionally:

```
test core::scip_overlay::tests::unopenable_overlay_path_propagates_error ... FAILED
test core::scip_overlay::tests::incompatible_overlay_db_is_quarantined ... ok
test core::facts::tests::incompatible_facts_db_is_recreated ... ok
test result: FAILED. 453 passed; 1 failed
```

Exactly as reported, in both directions.

**The directory-not-missing-path reasoning holds, and mutation 2 is what proves it.** `rename` succeeds on a directory, so a store that quarantined every error class opens cleanly at that path — and `unopenable_overlay_path_propagates_error` is the single test in the crate that goes red when the classification collapses. A missing parent directory would have failed the rename on both branches and the mutation would have survived.

Worth stating precisely, because the certification claim is narrower than it sounds: `incompatible_facts_db_is_recreated` **passes** under mutation 2. It pins the recovery arm, the `.v2-incompatible` suffix and the empty-store outcome — which is the part of `FactStore` that could have regressed in the extraction, so it does certify what it is being used to certify. It does not pin the classification. That arm is pinned only by the new overlay test. Since both stores now call one function, one test per arm is sufficient, and this is exactly the coverage the extraction buys.

---

## The other three

| Prior finding | Status |
|---|---|
| Module doc mischaracterized `FactStore` | Fixed. `scip_overlay.rs:17-24` now carries the argument that actually holds — quarantining is safe *here specifically* because `get` then reports `None`, which the HTTP layer serves as an honest 404 rather than a graph that silently lost an overlay. That is the real justification, and it is the one on the page. |
| Blocking redb on the tokio worker | Fixed, 3/3. `read_overlay` (`graph.rs:64-77`) covers both `graph_for_index` and `scip_overlay_status`; the `put` has its own wrapper at `graph.rs:387-389`. `grep` confirms no direct `state.scip_overlays.` call remains outside a `spawn_blocking`. The `put` wrapper is at the right level — it wraps the fsync and not the protobuf parse, so it does not quietly expand scope. Join errors map to 500 on both, so the wrapping added no fail-open. |
| `put` returning an unused record | Fixed — returns `()` (`scip_overlay.rs:129`). |

The header doc aside was taken (`graph.rs:44-46`): it now states the header describes the overlay's existence, so a `?language=` filter that strips every SCIP node still reports `present`.

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-analyze/src/core/redb_open.rs` | 1–99 | This is the seventh implementation of open-with-quarantine in the workspace, and `trusty-common::memory_core::store::redb_open::open_or_recreate` is already a public, exported entry point for the same capability. CLAUDE.md's common-entry-point rule names trusty-common as the home. | Consolidate onto the trusty-common entry point, extended with the caller-supplied suffix and `recovery` string this module added. | Promote |

The other copies are `trusty-agents/src/memory/redb_recovery.rs:65`, `trusty-review/src/store/dedup.rs:135`, `trusty-common/src/memory_core/store/concurrent_open.rs:295`, `trusty-common/src/memory_core/store/redb_open.rs:160`, and `trusty-search/src/core/corpus_recovery.rs:59`.

Two things keep this off the critical path. This PR does not increase the count — trusty-analyze had one copy in `facts.rs` before and has one in `redb_open.rs` now, with a second caller. And trusty-common's version is not a drop-in: it returns `Result<Database, DatabaseError>`, hardcodes its backup suffix, and returns the *original* error when the rename fails rather than the rename error. Adopting it as-is would change `FactStore`'s `.v2-incompatible` suffix and orphan files quarantined by shipped builds. That is a cross-crate change across six crates — rung 4–5 — and grafting it onto a scoped persistence fix would be the wrong trade. Recommending it as separate work; the PM decides whether it gets filed.

---

## Notes

**One residual I am deliberately not flagging.** After a quarantine, `/scip` answers 404 for an index that did have an overlay, so "never ingested" and "was ingested, then quarantined" are not distinguishable at the endpoint — the same shape as the original defect, at much lower probability. I am leaving it because the alarm actually fires and is not an aggregate: a loud ERROR at boot naming the path, the backup, and the redo step, plus the preserved bytes on disk. Closing it properly needs a persisted quarantine marker so `/scip` can report the third state, which is more design than this PR should carry. It was the acknowledged trade in round one and the module doc states it at `scip_overlay.rs:21-24`.

**The facts ERROR log string changed** from "incompatible/old format (redb 2.x); moved it aside … facts must be re-derived" to "unreadable/obsolete format; quarantined it …" with the redo step moved into a structured `recovery` field. I checked for consumers — nothing in the repo greps it, and the information is preserved. Flagging only in case a log alert exists outside the tree.

**Verification.** Local gates on `3c975cfd`: `fmt --check` + `clippy -D warnings` + `test -p trusty-analyze` → `EXIT=0`, 454 + 25 + 2 + 2 passed, 0 failed. `check_line_cap.sh` → 3749 files, 0 violations. `corrupt_db_fails_open_loudly` has no dangling references anywhere. No `ui/dist` in the diff. `version = "0.8.0"` untouched, confirmed both in `Cargo.toml` and in the live `/health` response — the 0.9.0 bump stays with `local-ops`. All three changelog fragments match what the code now does: the persistence fragment's startup-refusal sentence is gone, and the new `5049-redb-open-policy.md` describes the shared policy accurately, one category each. CI at last one-shot read: everything pass or skipping, `Test` still pending.

The MCP gap is [#5056](https://github.com/bobmatnyc/trusty-tools/issues/5056) and out of scope here as agreed.

PR #5055, opened 2026-08-07
ci: add a public-API/SemVer gate against crates.io (#5050)

### LINKED ISSUE #5050: The workspace has no public-API/SemVer gate

bobmatnyc:
Closes #5050.

## Defect

No `cargo-semver-checks` or equivalent existed anywhere in `.github/workflows/`,
`scripts/`, or `docs/`. Breaking changes reached crates.io undetected.

## Evidence

#4088. `trusty-common` 0.22.5 added a required public field —
`DaemonBridgeConfig.no_spawn_hint`, `daemon_bridge.rs:117` — on a PATCH bump.
Dependents on a `^0.22` floor re-resolved to it on a lockfile-free
`cargo install` and failed with E0063; `trusty-analyze` 0.7.3 was yanked.

## Resolution

`scripts/check_semver.sh` + `.github/workflows/semver-checks.yml`. For each crate
whose `crates/<crate>/src/**` changed, it resolves the latest non-yanked
crates.io release and runs `cargo semver-checks` against it.

**Scope: published crates only, and every skip prints its reason.**

| Condition | Behaviour |
|---|---|
| `publish = false` | skip |
| no library target | skip |
| crates.io 404 / only-yanked | skip — no baseline exists |
| version already carries a major bump | skip — the tool would run 0 lints |
| **probe fails any other way** | **hard failure, `TOOL ERROR`** |

Five workspace crates are genuinely unpublished (`trusty-kb`, `trusty-channels`,
`trusty-sld-lint`, `trusty-tui`, `trusty-agents`), so the absent-baseline path is
load-bearing, not hypothetical.

**Fail-loud, not fail-open.** A network error, a 5xx, or a malformed index entry
exits non-zero. `scripts/check_semver_selftest.sh` runs first in CI and proves
it, including the 404 case that keeps the other three honest:

```
  ok  scan floor rejects an empty diff (exit 1)
  ok  probe rejects an unreachable index (exit 1)
  ok  probe rejects HTTP 503 (exit 1)
  ok  probe treats HTTP 404 as 'never published' (exit 0, baseline=NONE)
check_semver_selftest: 4 passed, 0 failed.
```

**Feature policy.** `--default-features` would be theatre: `trusty-common`
declares `default = []` and `DaemonBridgeConfig` sits behind `mcp`, so the real
#4088 break passes clean under it (verified below). The default all-features
heuristic instead builds CUDA/CoreML backends no runner can build. The gate
enumerates every declared feature and subtracts only
`scripts/semver-checks-feature-exclusions.tsv`, two rows, each with a reason.

### 1. #4088's actual shape, registry versus registry

`trusty-common` 0.22.5 source from crates.io, baseline 0.22.4 — no workspace
involved:

```
$ cargo semver-checks --manifest-path trusty-common-0.22.5/Cargo.toml \
      --baseline-version 0.22.4 --only-explicit-features --features mcp
    Checking trusty-common v0.22.4 -> v0.22.5 (minor change)
     Checked [0.007s] 196 checks: 195 pass, 1 fail, 0 warn, 58 skip

--- failure constructible_struct_adds_field: struct exhaustively constructible through public API adds field ---

Failed in:
  field DaemonBridgeConfig.no_spawn_hint in .../src/mcp/daemon_bridge.rs:117

     Summary semver requires new major version: 1 major and 0 minor checks failed
EXIT=100
```

The same command with `--default-features` exits **0** — "196 checks: 196 pass" —
which is why the feature policy above is not a detail.

### 2. The gate script, end to end, through its real diff path

A required new public field on `PerRepoFetch` in `tga` (2.11.0, equal to the
published version) — the identical `constructible_struct_adds_field` shape:

```
$ bash scripts/check_semver.sh --base origin/main
CHECK tga: 2.11.0 -> 2.11.0 (none release), 1 feature(s)
    Checking tga v2.11.0 -> v2.11.0 (no change; assume patch)
     Checked [0.045s] 223 checks: 222 pass, 1 fail, 0 warn, 31 skip

--- failure constructible_struct_adds_field: struct exhaustively constructible through public API adds field ---

Failed in:
  field PerRepoFetch.remote_name in .../src/collect/collector.rs:65

     Summary semver requires new major version: 1 major and 0 minor checks failed
FAIL tga: cargo semver-checks exited 100 against baseline 2.11.0
EXIT=1
```

Same tree with the break kept and `tga` bumped 2.11.0 → 3.0.0:

```
SKIP tga: 2.11.0 -> 3.0.0 is already a major release — every lint is inapplicable
semver gate: scanned 9 changed path(s); 0 crate(s) checked, 1 skipped — OK.
EXIT=0
```

Unmodified `tga`, gate run in full:

```
CHECK tga: 2.11.0 -> 2.11.0 (none release), 1 feature(s)
     Checked [0.040s] 223 checks: 223 pass, 31 skip
semver gate: scanned (explicit); 1 crate(s) checked, 0 skipped — OK.
EXIT=0
```

Both demo commits were reverted; the branch is `bdddff170` plus the docs commit.

## Cost

Measured on this workspace (10-core arm64, warm cargo registry):

| Case | Wall clock |
|---|---|
| Docs-only / CI-only PR (this PR's own run) | ~1 s — git + `cargo metadata`, no build |
| Crate at an already-major version (`trusty-common` 0.28.1 → 0.29.0) | ~1 s — skipped before any build |
| `tga`, cold — two rustdoc builds | 1 m 04 s |
| `tga`, warm baseline | 7 s |
| `trusty-common`, cold, all 37 buildable features | 3 m 50 s |

The last row is the worst case and only lands on a PR that changes
`trusty-common/src/**` while its version sits at or one patch above the published
release. The already-major skip is what keeps it off the common path — and it
costs no coverage, because `cargo-semver-checks` itself runs zero lints there:

```
Checking trusty-common v0.28.1 -> v0.29.0 (major change)
 Checked 0 checks: 0 pass, 254 skip
```

Tool install is a pinned prebuilt binary via `taiki-e/install-action`, not
`cargo install` — that alone compiles for ~2 minutes before checking anything.
Nothing in the job touches `Cargo.lock` (`cargo metadata --no-deps`, and
`cargo-semver-checks` builds in its own scratch workspace; `git status` was clean
after every run above), so the MSRV 1.94 job is unaffected.

## Gates

CI-only + docs-only change; no crate `src/**` touched, so no changelog fragment
(confirmed by the gate, not assumed).

```
cargo fmt --check                       — clean
scripts/check_line_cap.sh               — 3745 files, 0 violations
scripts/check_sld.sh                    — 0 errors, 0 warnings
scripts/check_doc_numbers.sh            — 0 violations
scripts/check_changelog_fragment.sh     — "no crate source changed (docs-only / CI-only / test-only) — OK"
scripts/check_semver_selftest.sh        — 4 passed, 0 failed
scripts/check_semver.sh                 — "scanned 4 changed path(s); no crate source changed — OK"
```

## Follow-up, not in this PR

`SemVer checks` needs adding to the branch-protection required-check set, which
needs repo-admin rights. #4088's other two acceptance items — the registry-only
install smoke test, and the `#[non_exhaustive]`/builder decision on
`DaemonBridgeConfig` — stay open there.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -1,0 +1,88 @@
+# Public-API / SemVer gate for the trusty-tools workspace (issue #5050).
+#
+# The workspace publishes 20+ crates to crates.io and had no check that a
+# breaking public-API change carried a matching version bump. #4088 is what that
+# costs: `trusty-common` 0.22.5 added a required public field on a patch bump,
+# every dependent on a `^0.22` floor stopped compiling on a lockfile-free
+# `cargo install`, and `trusty-analyze` 0.7.3 was yanked. A workspace
+# `cargo check` cannot see that class of break — the root Cargo.toml's path
+# override always pairs local source with local dependency — so this job
+# compares against the REGISTRY instead.
+#
+# Cost control: the gate is diff-scoped, the same way check_changelog_fragment.sh
+# is. It builds rustdoc for a crate only when that crate's `src/**` changed AND
+# it is published AND its declared version does not already carry a breaking
+# bump. On a docs-only, CI-only, or unpublished-crate PR the job is a few
+# seconds of git and metadata; the two rustdoc builds are paid only by the PRs
+# that can actually cause the defect.
+#
+# Deliberately NO `paths:` filter (issue #4468): a required check skipped by a
+# path filter never reports on the PRs it skips and leaves them pending forever.
+# This job always runs and reports SUCCESS on an exempt PR — the exemption lives
+# in the script, not the trigger.
+#
+# `pull_request` only: crate selection is a diff against the base branch, which
+# is empty (and would trip the gate's own scan floor) on a push to main.
+name: SemVer checks
+
+on:
+  pull_request:
+    # Explicit activity types (#4179). `edited` is NOT in GitHub's default set
+    # and is the event a base-branch RETARGET fires, so without it a stacked PR
+    # retargeted onto main never re-triggers. Matches ci.yml.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
+    branches: [main]
+
+concurrency:
+  group: semver-checks-${{ github.ref }}
+  cancel-in-progress: ${{ github.event.action != 'edited' }}
+
+jobs:
+  semver-checks:
+    name: Public API / SemVer
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    steps:
+      # fetch-depth: 0 — the gate needs `git merge-base` against the base
+      # branch, which a shallow checkout cannot resolve.
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      # A prebuilt binary, PINNED. `cargo install cargo-semver-checks` compiles
+      # ~2 minutes of tool before a single crate is checked, which is exactly the
+      # kind of overhead that gets a gate disabled. install-action downloads a
+      # release artifact instead, and pinning the version keeps a tool release
+      # from silently changing the lint set under a PR. It also never touches
+      # this workspace's Cargo.lock — refreshing that turns the MSRV 1.94 job
+      # red (see CLAUDE.md, "Version bumps must not refresh Cargo.lock").
+      - name: Install cargo-semver-checks (pinned prebuilt)
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-semver-checks@0.50.0
+
+      - name: Cache cargo artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: semver-checks
+          # cargo-semver-checks builds into target/semver-checks/, which
+          # rust-cache prunes as "unrecognised" by default. Keeping it is the
+          # difference between a warm 1-minute check and a cold 5-minute one.
+          cache-directories: target/semver-checks
+
+      # Runs BEFORE the real gate, matching line-cap.yml's selftest-then-gate
+      # ordering: a gate that cannot fail makes the real run's green meaningless.
+      # These four cases are the gate's fail-open surface — an unscanned diff and
+      # an unreachable crates.io — plus the 404 case that proves the other three
+      # fail for the right reason.
+      - name: SemVer gate selftest (must refuse a blind run)
+        run: bash scripts/check_semver_selftest.sh
+
+      - name: Enforce public-API SemVer against crates.io
+        env:
+          SEMVER_GATE_BASE: ${{ github.event.pull_request.base.sha }}
+          SKIP_UI_BUILD: '1'
+        run: bash scripts/check_semver.sh
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -211,6 +211,18 @@ waits, the `tga` tag aliases (#1128), and the connection-safe daemon restart.
 > **Full release workflow, `scripts/bump-version.sh`, and Developer-ID signing
 > setup:** see [docs/reference/release-workflow.md](docs/reference/release-workflow.md).
 
+🔴 **A breaking public-API change needs a matching version bump, and CI now
+enforces it (#5050).** `scripts/check_semver.sh` runs `cargo-semver-checks`
+against the crate's latest crates.io release for every crate whose `src/**` the
+PR changed. Cargo's 0.x rule applies: for a `0.y.z` crate the breaking bump is
+the MINOR position. A workspace `cargo check` can never catch this class of
+break — the root `Cargo.toml` path override pairs local source with local
+dependency — which is how #4088 shipped `trusty-common` 0.22.5's required new
+public field on a patch bump and cost `trusty-analyze` 0.7.3 a yank. Prefer
+`#[non_exhaustive]` on public structs and enums so field and variant additions
+stay non-breaking by construction. See
+[docs/reference/semver-gate.md](docs/reference/semver-gate.md).
+
 🔴 **CRITICAL macOS note:** never use `cp` to install a release binary on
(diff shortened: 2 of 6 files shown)


bobmatnyc:
## Verdict: APPROVE

Independent verification performed against a live `cargo-semver-checks 0.50.0` install (matches the pinned tool version) — not just a read of the script.

### Core claim (feature-gating theatre) — CONFIRMED

Downloaded the real `trusty-common-0.22.5` and `0.22.4` crate tarballs from crates.io and ran both invocations myself:

```
--default-features:      196 checks: 196 pass, 58 skip   (misses the break)
--features mcp:           195 pass, 1 fail — constructible_struct_adds_field,
                           DaemonBridgeConfig.no_spawn_hint, daemon_bridge.rs:117
```

`default = []` at `crates/trusty-common/Cargo.toml:266`, and `daemon_bridge` sits behind `#[cfg(feature = "mcp")]` in `crates/trusty-common/src/lib.rs:166`. The PR's central justification for enumerating features instead of `--default-features` is real, not asserted.

Checked whether the enumeration generalizes beyond `trusty-common`: `trusty-git-analytics` (tga) also declares `default = []`, and the PR's own second demo (`PerRepoFetch.remote_name`) exercises that crate through the real diff path. No other published crate combines `default = []` with a feature-gated struct the way `trusty-common` does, so the two rows the PR already demonstrates are the two that matter today.

Ran the full 36-feature union locally (`--only-explicit-features` + every declared feature minus the two TSV exclusions) against baseline 0.22.4: builds clean in 286s (~4m46s on this machine vs. the PR's claimed 3m50s — cache/hardware variance, not a discrepancy in kind). `git status` on `Cargo.lock` stayed empty through every run.

### Fail-open hunt — mutation-tested, not just read

Ran `check_semver_selftest.sh` clean (4/4), then mutated `check_semver.sh` three separate ways to reintroduce the fail-open shapes the PR claims to guard against, re-ran the self-test against each, and restored the original after each mutation (`git status` clean throughout):

1. HTTP-503 branch reverted to `echo NONE; return 0` → self-test caught it (`probe/503: a 503 from the index exited 0`).
2. curl-failure branch reverted to `echo NONE; return 0` → self-test caught it (`probe/refused: an unreachable index exited 0`).
3. Scan-floor check reverted to `exit 0` on an empty diff → self-test caught it (`scan floor: the gate exited 0 over a diff of 0 paths`).

All three mutations were correctly flagged, each for the specific reason claimed. This is a real self-test, not the "passes against a broken script" shape flagged elsewhere tonight.

Beyond the self-test's own scope: read the per-crate loop (`check_semver.sh:485-495`) — any non-zero exit from `cargo semver-checks` itself (rustdoc build failure, OOM, timeout, or a real break) sets `fail=1` and the gate hard-fails at the end. There is no branch that treats a tool crash as a pass. The "already-major" skip is driven by `release_type()`, which correctly implements Cargo's 0.x-minor-is-breaking rule (checked by hand for the 0.x→0.x, 0.x→1.x, and 1.x→1.x cases) — it can only fire when the version was actually bumped past the break, which is the correct outcome, not a hole.

### Exclusions file, lockfile, tool pinning — all as claimed

- `scripts/semver-checks-feature-exclusions.tsv` — verified both rows are real: `embedder-cuda` pulls `ort/cuda` (needs nvcc), `embedder-coreml` links the macOS CoreML framework. Neither builds on `ubuntu-latest`. Nothing mechanically stops a future row from silencing a real feature (no build-must-actually-fail check on the TSV itself), but the file is two lines with mandatory inline reasons, and any addition is a visible diff line in normal PR review — same trust model this repo already uses for `.line-cap-allowlist.tsv`. Not a defect in this PR.
- `Cargo.lock` — confirmed untouched (`git status` empty) after every `cargo semver-checks` invocation I ran, including the full-feature run and the registry-tarball runs. `cargo metadata --no-deps` genuinely does not resolve the graph.
- Tool pin — `taiki-e/install-action@v2` with `cargo-semver-checks@0.50.0` in `.github/workflows/semver-checks.yml:65`; my local install is already exactly that version.
- Diff scoping (`check_semver.sh:399-424`) — the `crates/*/src/*` regex correctly handles nested paths, directory-vs-crate-name mismatches (tga), and crate removal. Ran `bash scripts/check_semver.sh --base origin/main` against this PR's own diff: correctly self-classified as docs/CI-only, 0 crates checked. One narrow gap: a public-API change introduced purely via a `Cargo.toml` dependency bump (a re-exported type shifting shape) without touching `src/**` would not trigger a scan — outside #4088's actual shape, not blocking.
- CI gates — reran `check_semver_selftest.sh`, `check_changelog_fragment.sh`, `check_line_cap.sh`, `check_sld.sh`, `check_doc_numbers.sh`, `cargo fmt --check` locally; all clean, matching the PR body's reported output.

## Findings

No findings at >80% confidence. All candidate concerns (exclusions-file governance, dependency-bump scoping gap) stayed below the confidence/severity bar for a table row — see Notes.

## Notes

- **Advisory-only until branch protection is updated.** `SemVer checks` is not yet in the required-check set; that needs repo-admin rights and is explicitly out of scope here. A PR can merge red against this gate until that follow-up lands. Worth tracking, not blocking this PR.
- Two soft spots, both below the bar for a finding: exclusion-file rows aren't mechanically re-validated against an actual build failure, and dependency-only (Cargo.toml) API surface changes aren't diff-scoped in. Neither is the shape #4088 was, and both are visible/fixable in normal review of whatever PR would exploit them.

APPROVED for merge (rung 5, release tooling — critic round complete).

Issue #5045, opened 2026-08-07
tm doctor is blind to missing trusty-search index coverage — ~94% of live worktrees have no registered index


bobmatnyc:
## Defect

Of 70+ live git worktrees for `bobmatnyc/trusty-tools`, only 4 have a registered trusty-search index — about 6% coverage. A bare `search` call 404s in the large majority of live worktrees, not as an edge case. `tm doctor` has no check that resolves a session's pinned index and confirms it exists, so the one tool meant to catch this is blind to it.

## Confidence

Observed — direct cross-reference, this session.

## Evidence

`git worktree list` across both live topologies for this repo (46 top-level plus 29 under `.base`) cross-referenced against `list_indexes`: 28 indexes registered, 4 matching a trusty-tools worktree path (`feat-4765-manifest-authority`, `fix-4594-direct-work`, `feat-critic-standard`, `tm-trusty-tools-05`). `crates/trusty-mpm/src/core/doctor.rs` has no check that resolves a session's pinned trusty-search index. Observed live this session as `POST /indexes/2eb72dca-…/search → 404 unknown index`.

## Resolution

Add a `tm doctor` check that resolves the running session's pinned trusty-search index and reports failure when `GET /indexes/:id/status` 404s.

## Relationship to #4715 — not a duplicate

#4715 covers the root cause: the daemon advertises a default index for a fresh worktree that does not exist, and frames the fix as an owner design decision (auto-provision vs. document the 404). This issue is narrower and additive: it quantifies the blast radius (94% of live worktrees) and closes the `tm doctor` blind spot independently of whichever fix #4715 lands. Both should stay open; cross-referenced here rather than merged.

## Design question — explicitly out of scope here

Whether indexes should be keyed per worktree at all, given worktrees are ephemeral by this repo's own doctrine while branches are durable, is a design question already tracked by #2611 (trusty-search managed-repo awareness / DOC-37 spec — main, `.base`, and worktree identity). Not part of this defect; not re-filed.

## Test level expected for closure

A `tm doctor` integration test against a session with no registered index, asserting the new check fires.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #138, opened 2026-05-25
feat(trusty-search): split search into per-lane MCP tools (search_lexical / search_semantic / search_kg / search_all) — push intent classification to the LLM


bobmatnyc:
## Proposal

Expose four distinct MCP tools — one per search-lane combination — instead of the current single `search` tool with a `stage` query parameter. Tool descriptions become the LLM's intent-classification prompt. The LLM picks the right tool for its query; the server stops trying to do regex-based intent classification (#119) and instead acts on the LLM's already-formed intent.

| New tool | Lanes consulted | When the LLM should use it | Cost |
|----------|----------------|---------------------------|------|
| `search_lexical` | BM25 + grep-fallback (Stage 1 only) | Exact symbol name, regex, literal string, file-glob | Cheap; ripgrep-equivalent latency |
| `search_semantic` | BM25 + HNSW vector via RRF | Conceptual queries where literal-text match isn't expected ("how does X work", "find code that does Y") | Medium; needs embedder warm |
| `search_kg` | Symbol-graph traversal from a seed (chunk_id or symbol name) | "Find callers of X", "what calls Y", refactoring impact, neighborhood exploration | Cheap once you have a seed |
| `search_all` (alias for current `search`) | Full hybrid (BM25 + HNSW + KG expansion + community_cohesion) | When the LLM is unsure, or when the query mixes lexical + conceptual signals | Most expensive; full pipeline |

Backward-compat: keep the existing `search` tool as an alias for `search_all`. No breaking change.

## Why this is better than the current `?stage=` parameter

1. **The LLM is the best intent classifier we have access to.** It already does this kind of classification implicitly when picking which tool to call. Making the choice explicit at the tool-selection layer is more reliable than a server-side regex (per the #119 benchmark, the regex classifier under-fired on 12/14 queries — and that was AFTER we tried to fix it).

2. **Tool descriptions become first-class prompts.** MCP tool descriptions are seen by every LLM that connects. We can write them carefully — with examples, anti-patterns, cost hints — and the LLM uses them to make better decisions than any server-side rule.

3. **Composability.** The LLM can chain tools naturally:
   - Call `search_lexical` to find a definition file
   - Call `search_kg` from that file's primary symbol to explore the neighborhood
   - Fall back to `search_semantic` if neither produced useful results
   This composition is invisible behind a single `search` tool with a `stage` param.

[...]

PR #5057, opened 2026-08-07
fix(trusty-search): answer INDEX_NOT_READY, not 404, for a never-indexed worktree

### LINKED ISSUE #4715: Bare search call 404s on the daemon's advertised default index when worktree was never indexed
## Defect

Per issue #1373, omitting `index_id` on a bare search call should resolve to the session's pinned project index. On a worktree that has never been indexed, the daemon advertises the session id as the default index in the tool schema, but that index does not exist, so the bare call fails:

```
POST /indexes/2eb72dca-de08-481b-8dfa-22ab7f81b1f9/search → 404 "unknown index"
```

`index_status` on the same advertised default also 404s. This is not a caller guessing an id wrong — it is the daemon's own default not existing.

`search_health` is otherwise fine: 14 indexes loaded, embedder ready. An explicit `index_id` works correctly and returns current results.

## Why It Matters

The contract is broken: advertising a default that does not exist is a usability defect and a silent failure mode for any tooling that relies on the bare call contract.

## Design Decision Needed

Should session start auto-provision an index for a fresh worktree, or should the bare-call contract explicitly document "404 if the project was never indexed"? Either is defensible; the current state (advertised but nonexistent) is not.

## Evidence

Observed during trusty-mpm 1.3.4 validation sweep on 2026-08-03; reproducible on any worktree with zero search indexes.

bobmatnyc:
## Defect

A `search` against a worktree the daemon has never indexed returned the daemon's `404 unknown index` verbatim. "Unknown" reads as permanent, so a caller gives up or silently falls back to `git grep` — which is what happened in the session that produced this ticket, and the gap stayed invisible for some time. The state is transient: the index will exist, it just does not yet.

This is a state misreported as a different state. [#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) measures the blast radius — roughly 4 of 70+ live worktrees for this repo have a registered index.

## Resolution

Per the owner's ruling on [#4715](https://github.com/bobmatnyc/trusty-tools/issues/4715), the caller-visible fix is a JSON-RPC error contract on the MCP tool surface, not an HTTP status — consistent with [ADR-0031](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0031-uds-for-inter-crate-transport-http-for-external.md).

🔴 **The HTTP side did change, in `e5c246033`.** An earlier revision of this section said it had not; that is no longer true, and a release engineer reading the old wording would cut a patch. Three index-scoped endpoints now return **503 where they previously returned 404** for an index that is registered but not resident:

| Endpoint | Was | Now, for a cold-parked or restore-failed index |
|---|---|---|
| `GET /indexes/:id/status` | 404 | **503** |
| `GET /indexes/:id/chunks` | 404 | **503** |
| `POST /indexes/:id/grep` | 404 | **503** |

404 from these endpoints now means what it has always meant on `POST /indexes/:id/search`: absent from the hot registry, the cold store, and the failed set. This was a precondition, not a nicety — the MCP classifier reads a 404 as "never built", and without it three routed tools reported a cold-parked index as one that never existed.

**Ruling: ship as a minor, 0.43.0.** It is an observable wire change and must not go out as a patch. It is not a Cargo-SemVer break: the two in-tree callers that branch on 404 from `/indexes/:id/status` (`trusty-agents/src/stores/status.rs:277`, `index_feed.rs:174`) land on adjacent error arms, so only the human-facing string differs, and nothing consumes the `index_not_resident` / `index_restore_failed` strings. Version file untouched — the bump is `local-ops`'.

**The discriminator.** The daemon cannot tell "never indexed" from "unknown id" — from its side both are "no such id". The MCP layer can, because it knows which id it *advertised* to the caller as this session's default (`tools/list` annotates the pinned index, [#1373](https://github.com/bobmatnyc/trusty-tools/issues/1373)). A 404 on the id the daemon told the caller to use means "not built yet". A 404 on any other id is a caller-supplied id that really does not exist, and stays exactly as it was.

**The contract**, mirroring the existing `STAGE_NOT_READY` shape ([#138](https://github.com/bobmatnyc/trusty-tools/issues/138)) rather than inventing a second convention:

| Form | Carrier |
|---|---|
| `tools/call` | `isError: true` + `_meta` payload |
| bare method | JSON-RPC code `-32011` + `error.data` payload |

```json
{
  "error_code": "INDEX_NOT_READY",
  "state": "not_indexed",
  "index_id": "2eb72dca-…",
  "retryable": true,
  "reason": "the daemon advertised this index as the session default, but no index has been built for it yet",
  "suggested_fallback": ["grep", "find"]
}
```

Covered tools: `search`, `search_lexical`, `search_semantic`, `search_kg`, `search_all`, `grep`, `index_status`. The pinned `grep` tool is index-backed too, so it reports the same state — otherwise the fallback advice would loop back into a tool that cannot answer either. The `index_id` schema description now names the contract, so a model learns it exists before hitting one.

**One state, not three.** "Indexing in progress" already has its own signals — the daemon's `503 index_loading` and the per-lane `STAGE_NOT_READY` — and an in-progress index still answers, partially. A stale index answers too, and the caller's action does not change. Only "never indexed" produces no answer at all, so it is the only state a caller can act on differently. Rationale is inline at `not_ready.rs:39-49`.

## Not in scope

**This does not close #4715.** Whether session launch should synchronously create the index, auto-provision on first miss, or stay lazy is the separate design decision on that ticket, and the owner has not made it. This PR makes the *not-ready* answer honest and actionable; it does not decide when indexes get created. The index-creation path is untouched.

## Fail-open check

The bug is a state reported as another state, so the fix must not introduce a new instance.

- **Not-ready vs. zero hits.** `not_ready_is_distinguishable_from_an_empty_result_set` runs the same call against a daemon returning `{results: []}` and one returning 404, then asserts they differ in both fields a caller branches on. An empty result set is `isError: false` with no `_meta`; not-ready is `isError: true` with `_meta.error_code`.
- **A failing readiness check does not degrade.** The classification is a pure comparison over data already held (`pinned_index` vs. the requested id) — it has no fallible step. It can only turn one error into a more specific error; it never produces `Ok`. `readiness_check_failure_does_not_degrade_to_ready_or_empty` pins this for a 500 and for an unreachable daemon.
- **Scope containment.** `classify_index_miss_only_fires_for_the_pinned_index` asserts a foreign id and an unpinned session both keep the old behavior.

## Mutation proof

Disabled the decision point (`classify_index_miss` returns `None` unconditionally = pre-fix behavior), reran:

```
test result: FAILED. 1 passed; 8 failed; 0 ignored; 0 measured; 1496 filtered out

---- not_ready_message_says_retryable_and_names_the_fallback stdout ----
must say it is retryable: POST http://127.0.0.1:54430/indexes/wt-1/search
returned 404 Not Found: {"error":"unknown index: wt-1"}

---- bare_method_search_on_unindexed_pin_returns_index_not_ready_code stdout ----
  left: -32603
 right: -32011

---- tools_call_search_on_unindexed_pin_returns_structured_not_ready stdout ----
_meta must carry the machine-readable contract
```

The one that passes under mutation is `readiness_check_failure_does_not_degrade_to_ready_or_empty` — it is the invariant guard, correct both before and after.

Restored:

```
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 1496 filtered out
```

## Gates — Rust test ladder rung 3 (localized behavior inside one crate)

```
cargo check -p trusty-search --all-targets             EXIT=0
cargo clippy -p trusty-search --all-targets -D warnings EXIT=0
cargo test -p trusty-search                            EXIT=0
```

`cargo test -p trusty-search` — 26 test binaries, all `ok`; lib 1504 passed / 0 failed / 1 ignored, bin 372 passed / 0 failed / 3 ignored. No watcher flake surfaced on this run.

```
  line-cap: measured 3747 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
  test-pointers: resolved 22130 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
bash scripts/check_sld.sh
  sld-lint: scanned 56 spec doc(s) + 3120 code file(s); 0 error(s), 0 warning(s)
```

## Published-crate note

`trusty-search` is published, and this touches its public contract three ways:

1. Two new `pub` items in `trusty_search::mcp::tools` (`INDEX_NOT_READY_CODE`, `INDEX_NOT_READY`) — additive.
2. A new error shape on the MCP tool surface — additive; no existing tool response changes for an index that exists, and a 404 on any non-advertised id is byte-identical to before.
3. **The 404 → 503 change on three HTTP management endpoints** (table in Resolution) — *not* additive. Additive for the 404-means-absent contract, but an observable response-code change on the management surface.

Item 3 is what forces **0.43.0** rather than 0.42.4. Flagging only; the bump belongs to `local-ops`, not this PR.

## LOC

Added: 421 / Removed: 16 / Net: +405 (of which 302 are the new test file and the contract module's docs).

## Not observed end to end

No 503 has been seen over a real socket. All six `tests_4715` tests call the handler functions directly and bypass the axum router; the MCP-side tests drive a stub daemon. The status codes, the cold-store branches, and the arm ordering are proven at the handler boundary and under mutation — the router wiring and a live cold-parked daemon are not. Recording it as unobserved rather than leaving it implicit.

Refs #4715


bobmatnyc:
## Verdict: WARN

Verified directly: mutation proof reproduced exactly (forcing `classify_index_miss` to
return `None` gives `1 passed; 8 failed`, matching the PR body byte-for-byte, with
`readiness_check_failure_does_not_degrade_to_ready_or_empty` as the sole survivor).
Traced every call into `classify_index_miss` (`get_scoped`/`post_scoped` in
`http.rs`) — the 404 branch is checked and returned before either helper attempts
to decode the response body, so a non-JSON 404 can never fall through to `Ok`.
Fail-open guards hold. `pinned_index` is set once at `McpServer` construction
(`mod.rs:88`, `123-127`) and never mutated afterward — one process per session in
both stdio and HTTP/SSE mode — so it cannot drift within a session or across a
reconnect. `-32011` doesn't collide with any other code in this crate (`-32010` is
the only sibling) or with the JSON-RPC reserved range. `tools/call` degrades to an
opaque `isError: true` for an old client that ignores `_meta`; bare-method degrades
to a generic JSON-RPC error object — never to apparent success. SemVer: only two
new `pub const`s added, no existing signature touched — additive, 0.43.0 is right.
Gates: `cargo test -p trusty-search --lib` — 1504 passed, 0 failed, 1 ignored;
`cargo clippy -p trusty-search --all-targets -- -D warnings` — clean;
`scripts/check_line_cap.sh` — 0 violations.

One HIGH finding on the "fourth case" question, gated behind an opt-in daemon
setting that's off by default.

## Findings

| HIGH | `crates/trusty-search/src/service/server/status.rs:91` | 91 | `index_status_handler` looks up the pinned id with a raw `state.registry.get()` — the hot in-memory registry only, no cold-store fallback. `search_handler` (`search.rs:170-176`) checks `cold_store` and lazily reloads via `get_or_load_index` before ever returning 404; `index_status` has no such path. When an operator sets `TRUSTY_MAX_RESIDENT_INDEXES` (opt-in runtime eviction, issue #2161, off by default), a cold-parked index that genuinely exists 404s on `/status`. `classify_index_miss` then reports it as `INDEX_NOT_READY`/`state: "not_indexed"` with the message "has never been built yet" — false for an index that was built and merely needs a reload. `run_lane_search`'s Stage 2/3 pre-flight (`search.rs:257-259`) calls this same endpoint before ever reaching the daemon's cold-aware search path, so `search_semantic`/`search_kg` inherit the same misclassification. This is the "new lie, in the opposite direction" the design brief asks to rule out — narrowed by the opt-in gate, but not eliminated by it. | Give `index_status_handler` (and the chunks/status family) the same `cold_store` check `search_handler` already has, or have `classify_index_miss` consult something that reflects "ever registered" (`indexes.toml`/`cold_store` membership) rather than only "currently hot", before labeling the state `not_indexed`. Daemon-side change, out of this PR's stated MCP-layer scope. | Promote |
| MEDIUM | `crates/trusty-search/src/mcp/tools/index.rs:181` | 181 | `list_chunks` calls the unscoped `get_query`, so a never-indexed pinned index still gets the old raw `GET ... returned 404 Not Found: {"error":"unknown index: ..."}` instead of the new contract. Its daemon handler (`get_index_chunks_handler`, `files.rs:130`) has the exact same `state.registry.get(&index_id).ok_or(StatusCode::NOT_FOUND)` shape as `index_status_handler` — same defect, same fix shape, just not wired through `get_scoped`. Inconsistent with its neighbor `index_status` two arms up in the same file. | Add a scoped variant of `get_query` (or thread `index_id` through the existing one) and pass `Some(&index_id)`, mirroring `index_status`'s change. | Parent |

## Notes

- Confirmed the seven paths named in the commit message (`search`, the four
  per-lane tools, `grep`, `index_status`) are the only ones routed through
  `get_scoped`/`post_scoped`, and confirmed none of the excluded paths
  (`index_file`, `remove_file`, `reindex`, `delete_index`, `search_similar`) was
  wrongly routed — all are either mutating (out of the stated "read path" scope)
  or don't consult the pinned index at all (`search_similar` uses its own `index`
  arg, defaulting to the literal `"default"`, never `resolve_index_id`).
- The three-state reduction (never-indexed / in-progress / stale) holds for every
  case the PR's own reasoning covers. The HIGH finding above is a fourth case the
  reduction didn't consider — "registered, built, currently evicted to cold
  storage" — which is real but requires an opt-in, default-off daemon setting to
  reach. Given that gate, it doesn't block merge; it should be tracked before
  `TRUSTY_MAX_RESIDENT_INDEXES` sees real production use with pinned MCP sessions.
- Backward-compat and SemVer both check out as claimed in the PR body.

bobmatnyc:
## Review follow-up — the MEDIUM, and what assessing it turned up

**Chose option 3, but the scope was not what the finding described.** Assessing `list_chunks` showed I had already shipped the misclassification in three routed places, not zero. The critic caught the one endpoint I had *not* routed; the three I *had* routed have the same daemon-side shape.

Only `search_handler` has the cold-store fallback. Every other index-scoped handler is a bare `state.registry.get(&id).ok_or(NOT_FOUND)`:

| Routed tool | Daemon endpoint | Cold-store aware before this commit |
|---|---|---|
| `search`, `search_all`, `search_lexical` | `POST /indexes/:id/search` | yes |
| `search_semantic`, `search_kg` (pre-flight) | `GET /indexes/:id/status` | **no** |
| `index_status` | `GET /indexes/:id/status` | **no** |
| `grep` | `POST /indexes/:id/grep` | **no** |
| `list_chunks` (the MEDIUM) | `GET /indexes/:id/chunks` | **no** |

So three of the seven already reported a cold-parked index as `not_indexed` — "never built" for an index that was built. Leaving `list_chunks` unrouted with a comment would have documented the lie while leaving three live instances in place.

**Reachability is also higher than the review assumed.** `TRUSTY_MAX_RESIDENT_INDEXES` (#2161) is not the only route to cold-parked. `restore_eager_entry` (`commands/start/restore.rs:440`) calls `park_if_parkable` on **every** eager restore, and `ColdIndexStore::park_if_parkable` parks any entry whose restore timed out (`lazy_loader/store.rs:288`, #4087/#4250). No env var. A slow boot on the default configuration puts a real index in the cold store and out of the hot registry.

## The fix

Three handlers now check the cold store and the failed set before answering — 503 for "registered but not resident", 404 reserved for "absent from every store", matching `search_handler`. Roughly four lines each, no signature changes:

```rust
let handle = match state.registry.get(&index_id) {
    Some(h) => h,
    None if state.cold_store.contains(&index_id) || state.cold_store.is_failed(&index_id) => {
        return Err(StatusCode::SERVICE_UNAVAILABLE)
    }
    None => return Err(StatusCode::NOT_FOUND),
};
```

With that precondition in place, `list_chunks` is routed — the MEDIUM as originally scoped.

**This is not the HIGH.** The HIGH is that these endpoints cannot *reload* a cold index the way `search_handler` does. That is still open and still separable. What this commit does is narrower: stop the endpoints from saying "does not exist" about something that does. It removes the precondition that let my classifier lie; it does not add lazy restore.

The invariant is now written down at `not_ready.rs`, so a future handler change has something explicit to violate: only route an endpoint whose 404 rules out the hot registry, the cold store, and the failed set.

## Mutation proof — daemon side

Removed the three cold-store guards:

```
test result: FAILED. 1 passed; 4 failed; 0 ignored; 1507 filtered out

---- cold_parked_index_status_is_503_not_404 stdout ----
  left: 404
 right: 503
---- cold_parked_index_grep_is_503_not_404 stdout ----
  left: 404
 right: 503
---- cold_parked_index_chunks_is_503_not_404 stdout ----
  left: 404
 right: 503
---- restore_failed_index_status_is_503_not_404 stdout ----
  left: 404
 right: 503
```

The one that passes under mutation is `status_404_only_when_absent_from_every_store` — the invariant that must hold in both states. Restored: `ok. 5 passed; 0 failed`.

New MCP-side test `non_resident_503_is_not_classified_as_never_indexed` asserts the classifier leaves a 503 alone, so the misclassification cannot come back through the other door.

## Gates — rebased on `origin/main` (124d4ed33 → e5c246033)

```
cargo fmt --check && cargo check -p trusty-search --all-targets
  && cargo clippy -p trusty-search --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-search                                       EXIT=0
```

26 test binaries `ok`; lib 1511 passed / 0 failed / 1 ignored, bin 372 passed / 0 failed / 3 ignored. No watcher flake this run.

```
scripts/check_line_cap.sh          rc=0  3750 files, 0 violations
scripts/check_test_pointers.sh     rc=0  22135 citations, 0 dangling
scripts/check_sld.sh               rc=0  0 errors, 0 warnings
scripts/check_changelog_fragment.sh rc=0 14 changed paths, 1 crate recorded
```

Captured with `rc=$?`, not `echo "EXIT=$?"`.

## Unchanged

Version stays 0.42.3 — the 0.43.0 bump is `local-ops`'. The public-surface note still holds and now also covers a behavior change: `GET /indexes/:id/status`, `GET /indexes/:id/chunks`, and `POST /indexes/:id/grep` return 503 where they previously returned 404 for a non-resident index. Additive for the 404-means-absent contract, but it is an HTTP response-code change on the management surface and belongs in the bump note.

bobmatnyc (commented):
## code-critic — round 3 @ `e5c246033`

## Verdict: APPROVE

Zero CRITICAL, zero HIGH. The widened scope is sound: the daemon-side precondition is real, the mutation reproduces exactly as claimed, and every routed endpoint now satisfies the invariant the PR writes down. The round-2 HIGH is reclassified to MEDIUM (reasoning below) — it is not being silently dropped.

### 1. Mutation re-run — reproduced independently

Reverted all three guards to the pre-commit `registry.get(&id).ok_or(NOT_FOUND)`, in `status.rs:103-110`, `files.rs:132-138`, and `files.rs:274-294`:

```
test service::server::tests_4715::cold_parked_index_status_is_503_not_404 ... FAILED
test service::server::tests_4715::cold_parked_index_chunks_is_503_not_404 ... FAILED
test service::server::tests_4715::restore_failed_index_status_is_503_not_404 ... FAILED
test service::server::tests_4715::cold_parked_index_grep_is_503_not_404 ... FAILED
test service::server::tests_4715::status_404_only_when_absent_from_every_store ... ok
  left: 404
 right: 503        (x4, at tests_4715.rs:71, 92, 121, 145)
test result: FAILED. 1 passed; 4 failed; 0 ignored; 0 measured; 1507 filtered out
```

Restored:

```
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1507 filtered out
```

The table in the PR discussion is accurate — including which test survives and why.

### 2. Reachability — claim verified, no env var required

- `commands/start/restore.rs:426-441` — `restore_eager_entry` calls `state.cold_store.park_if_parkable(parked, outcome)` unconditionally on every eager restore.
- `lazy_loader/store.rs:288-307` — parks on `TimedOut` only, stamped `ColdReason::TimedOut`.
- `warm_boot/mod.rs:104-111` — `warmboot_index_timeout()` falls back to `ROOT_SCAN_TIMEOUT` when `TRUSTY_WARMBOOT_INDEX_TIMEOUT_SECS` is unset, so the eager restore is deadline-bounded on stock config.
- `lazy_restore.rs:93,159` and `timeout_recovery.rs:182` reach `mark_failed` on default config too.

`TRUSTY_MAX_RESIDENT_INDEXES` (#2161) is a second route, not the only one. The round-2 review understated this; the correction stands.

### 3. Completeness of the sweep — confirmed by enumeration

Six `*_scoped` call sites cover four distinct daemon endpoints:

| MCP call site | Endpoint | Handler rules out cold store + failed set |
|---|---|---|
| `search.rs:154`, `search.rs:349` | `POST /indexes/:id/search` | yes (`search.rs:178-197`, pre-existing) |
| `search.rs:258`, `index.rs:155` | `GET /indexes/:id/status` | yes (`status.rs:103-110`, new) |
| `misc.rs:167` | `POST /indexes/:id/grep` | yes (`files.rs:274-294`, new) |
| `index.rs:183` | `GET /indexes/:id/chunks` | yes (`files.rs:132-138`, new) |

Three index-scoped MCP tools remain on the unscoped helpers — `typeahead` (`typeahead.rs:63`), `search_similar` (`search.rs:207`), `reindex` (`index.rs:142`). That is correct, not an omission: `search_similar_handler` (`service/server/routing.rs:206`) still 404s on a bare hot-registry miss, so routing it would violate the invariant. Nothing routed is missing its guard, and nothing guarded is unrouted.

### 4. 404/503 boundary in the other direction

`contains()` reads `entries` only, `is_failed()` reads `failed_entries` only, and `mark_failed` moves an id between them (`store.rs:540-544`), so the two checks together are exactly "known to the store in either population". No overlap gap, no way to claim residency for an id that was never registered — except the delete case in finding 2 below.

### 5. Fail-open

None. All three new arms are early returns on a pure O(1) read. Nothing advances a cursor, marks anything loaded, or downgrades a failure to a default. Every new error arm has a test that fails against the pre-fix code (proven above); the fifth test, `status_404_only_when_absent_from_every_store`, is the invariant guard and is correct in both states.

### 6. Gates re-run in a clean worktree at `e5c246033`

```
cargo fmt --check && cargo clippy -p trusty-search --all-targets -- -D warnings   EXIT=0
cargo test -p trusty-search                                                       EXIT=0
  lib 1511 passed / 0 failed / 1 ignored; bin 372 passed / 0 failed / 3 ignored
  26 test binaries, all ok
bash scripts/check_line_cap.sh          3750 files, 0 violations
bash scripts/check_test_pointers.sh     22135 citations, 0 dangling
bash scripts/check_sld.sh               0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh  OK — trusty-search fragment present
```

---

## Breaking-change judgment (called explicitly)

**It is an observable wire-behavior change and must not ship as a patch release. It is not a Cargo-SemVer break, and no in-tree caller breaks.**

- **Rust API**: strictly additive — two new `pub` consts in `trusty_search::mcp::tools`. Minor bump under Cargo SemVer.
- **HTTP surface**: `GET /indexes/:id/status`, `GET /indexes/:id/chunks`, and `POST /indexes/:id/grep` now answer 503 where they answered 404, for an index in the cold store or the failed set. External daemon clients can observe that. Under 0.x, `0.42.3 → 0.43.0` is the correct slot — the same bump the PR already recommends for the MCP additions. `0.42.4` would be wrong.
- **Version file untouched at 0.42.3 is correct** — bumps are `local-ops`' job in this repo, not the implementation PR's.

**In-tree callers, grepped rather than reasoned about.** Two branch on 404 from `/indexes/:id/status`:

- `crates/trusty-agents/src/stores/status.rs:277` — 404 arm and the generic non-2xx arm both return `StoreStatus::disconnected`. Only the human-facing string changes: "index is not registered" → "trusty-search returned HTTP 503".
- `crates/trusty-agents/src/stores/index_feed.rs:174` — same shape, both arms `anyhow::bail!`.

Everything else is status-agnostic: `crates/trusty-mpm/src/session_manager/search_gc.rs:225` maps every non-2xx to `0`; `crates/trusty-analyze/src/core/client.rs:171` and `crates/trusty-common/src/monitor/search_client.rs` use `error_for_status()`; `crates/trusty-common/src/search_readiness.rs:184` and `search_index.rs` test `is_success()` only. The Svelte UI's only 404-adjacent branch is `Search.svelte:145` on `e.status === 503` from the search endpoint, unaffected. No caller loses correctness; nothing consumes the `index_not_resident` / `index_restore_failed` discriminator strings.

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | PR description | Resolution § | "Nothing on the HTTP/management side changed — consistent with ADR-0031" is false as of `e5c246033`. The Published-crate note also covers only the MCP additions. A release engineer reading the body concludes no wire change and cuts a patch — this repo has been bitten by exactly that (trusty-common 0.26.3, trusty-analyze 0.7.3 yank). | Rewrite the Resolution sentence and extend the Published-crate note to name the three endpoints' 404→503 change as the reason 0.43.0 is required. The changelog fragment already says it correctly. | Fix here |
| MEDIUM | crates/trusty-search/src/service/server/search.rs | 106-152 | `unregister_index` never purges `cold_store.entries` or `failed_entries`. `DELETE /indexes/:id` on an index that is cold-parked or restore-failed leaves the stale entry, so the three new guards answer 503 forever for an index that no longer exists — violating the invariant this PR just wrote at `not_ready.rs:78-85` ("404 only when absent from every store"). The MCP layer then cannot report `INDEX_NOT_READY` for the delete-then-reindex case, which is the #4715 scenario. Pre-existing on `search_handler`; this PR newly depends on it. | Purge both maps in `unregister_index` alongside the registry removal. | Promote — the fix lives in a handler this PR does not touch, and it is pre-existing on the search path; it is separable, schedulable work rather than in-scope repair |
| MEDIUM | crates/trusty-search/src/service/server/status.rs, files.rs | 106-108, 134-136 | Round-2's HIGH, **reclassified**. These three endpoints report the cold state truthfully but cannot reload it, while `search_handler:195-240` does. `index_status_handler` and `get_index_chunks_handler` return a bare `StatusCode` with no body, so the MCP layer surfaces `GET … returned 503 Service Unavailable: ` with empty text — an agent gets a dead end even though a plain `search` on the same index would self-heal and succeed. Not HIGH: the state is now reported accurately, it is not a regression (both endpoints previously answered 404, which was worse), and a retry path exists. | Either route these three through `get_or_load_index` as search does, or give the 503 a body naming `search` as the trigger that restores the index. | Promote |
| LOW | crates/trusty-search/src/service/server/files.rs | 276-287 | The grep 503 body says "registered but not loaded — retry after it restores" for both the cold-parked case and the permanently-failed case. `mark_failed` is documented as permanent for the daemon's lifetime (`store.rs:521-528`), so that advice is wrong for the failed arm — `search_handler:183-193` gets it right with `index_restore_failed` / "restart the daemon or re-register". Two endpoints of the same daemon give different answers to the same question. | Split the arm: `is_failed` → `index_restore_failed` with search's wording; `contains` → keep `index_not_resident`. | Fix here |
| LOW | crates/trusty-search/src/mcp/tools/http.rs | 88 | `#[cfg_attr(not(test), allow(dead_code))]` on `get_query` is unnecessary — `typeahead.rs:63` still calls it in non-test builds. A suppression that is currently a no-op will silently hide a real dead-code signal the day typeahead moves to the scoped helper. | Delete the attribute. | Fix here |

## Notes

- **Has a 503 ever been observed end to end? No.** All five `tests_4715` tests call the handler functions directly, bypassing the axum router and its middleware; the MCP-side tests drive a stub axum server returning a canned status. There is no test, and no manual run, in which a real daemon parked a real index and answered 503 over the wire. The live daemon on this machine (0.42.3) reports `indexes_lazy: 0, indexes_failed: 0, indexes_skipped_timeout: 0`, so the state is not currently present to probe. The handlers return `StatusCode` / `(StatusCode, Json)` which axum maps verbatim, so the router-level divergence risk is low — but it is unproven, and the verified evidence stops at the handler function boundary.
- `non_resident_503_is_not_classified_as_never_indexed` (`tests_not_ready.rs:277`) passes both pre- and post-fix — `classify_index_miss` only fires on 404, so a 503 was never at risk. It is a correct invariant guard, not a regression test; do not read it as mutation-proven coverage.
- CI at the time of review: all checks SUCCESS except `Test`, still IN_PROGRESS. Not blocking on it.
- Reviewed from a detached worktree at `.claude/worktrees/qa-5057-round3`; the mutation was applied and reverted there only, tree left clean.

bobmatnyc:
## All three fixed at `6917e4cb8`. One of them corrected me.

### 1. MEDIUM — PR body was false. Rewritten.

The Resolution section now carries the wire change as a table (404 → 503 on the three endpoints), the reason it was a precondition rather than a nicety, and your breaking-change ruling verbatim: **0.43.0, not a patch, not a Cargo-SemVer break**, with the two in-tree 404 callers named. The Published-crate note is now three items instead of two, and item 3 — the one that is *not* additive — is called out as the thing forcing the minor. Version file untouched.

### 2. LOW — grep's remedy. Fixed, and the fix taught me something.

`grep_handler` now returns `index_restore_failed` with `search_handler`'s wording. I put the failed arm ahead of the cold-parked arm and wrote a test asserting the overlap state — **and the test failed on its own precondition.**

`mark_failed` removes the entry (`store.rs:541`), so `mark_failed` alone leaves the sets disjoint. My comment claiming "a failed entry can be present in `entries` too" was asserted, not checked. It is reachable, but not the way I wrote it: nothing ever removes from `failed_entries`, and `register_with_reason` (`store.rs:203`) never consults it, so a **re-registration after a failure** puts an id in both. The test now builds that state through the public store API, and the comment says which call produces it.

Mutation on the new arm order — swapped the guards so `contains` matches first:

```
---- restore_failed_index_grep_says_restart_not_retry stdout ----
  left: String("index_not_resident")
 right: "index_restore_failed"
```

Restored: `ok. 6 passed; 0 failed`. So the ordering is load-bearing, just for a different reason than I first wrote down.

### 3. LOW — dead suppression. Removed.

`#[cfg_attr(not(test), allow(dead_code))]` gone from `get_query`. Clippy stays green at `-D warnings`, which is the confirmation it was a no-op.

## Gates — `6917e4cb8`

```
cargo fmt --check                                      rc=0
cargo clippy -p trusty-search --all-targets -- -D warnings   rc=0
cargo test -p trusty-search                            rc=0
```

26 test binaries, 26 `ok` — 1931 passed, 0 failed, 41 ignored. No watcher flake.

```
scripts/check_line_cap.sh           rc=0  3750 files, 0 violations
scripts/check_test_pointers.sh      rc=0  22135 citations, 0 dangling
scripts/check_sld.sh                rc=0  0 errors, 0 warnings
scripts/check_changelog_fragment.sh rc=0  1 crate recorded
```

Changelog fragment gained a line for grep's corrected remedy string.

## Not observed end to end — now stated in the body

No 503 has been seen over a real socket. All six `tests_4715` tests call handlers directly and bypass the router; the MCP tests drive a stub daemon. Status codes, cold-store branches, and arm ordering are proven at the handler boundary and under mutation. Router wiring and a live cold-parked daemon are not.

Not touched, per your instruction: `unregister_index` never purging `cold_store.entries` / `failed_entries`, and the carried-over HIGH on reload-vs-report.

Issue #5040, opened 2026-08-07
stale_assets_for_many_* flake: full cargo test -p trusty-mpm is red on main, passes in isolation


bobmatnyc:
## Defect

The `stale_assets_for_many_*` pair in `daemon::managed_routes::tests` fails under a full `cargo test -p trusty-mpm` run but passes in isolation. Which of the two fails varies between runs.

- `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call`
- `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet`

Both die on the same precondition, `crates/trusty-mpm/src/daemon/managed_routes/tests.rs:1220`:

```
a freshly deployed fleet must start out fresh, else the assertion below is vacuous
```

## Evidence

Reproduced on `origin/main` at `8ab2c147a`, no local changes:

```
$ cargo test -p trusty-mpm
test daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet ... FAILED
test result: FAILED. 4724 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 48.96s
```

On a branch off that same commit, touching zero files under `crates/trusty-mpm/src/daemon/`, the *other* member of the pair failed instead:

```
$ cargo test -p trusty-mpm
test daemon::managed_routes::tests::stale_assets_for_many_sees_an_agent_change_on_the_very_next_call ... FAILED
test result: FAILED. 4732 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 49.98s
```

Each passes alone:

```
$ cargo test -p trusty-mpm --lib stale_assets_for_many_sees_an_agent_change_on_the_very_next_call
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4739 filtered out; finished in 0.03s
```

## Suspected cause

Stated as a suspect, not a conclusion. Both tests take a `fake_home()` guard and then resolve paths through `FrameworkPaths::default()`, which reads the process-global `HOME`. `#[serial_test::serial]` serialises the pair against each other but not against the rest of the suite, so any concurrently-running test that reads or rewrites `HOME` can make the "freshly deployed" precondition observe someone else's tree. That would explain both the isolation/full-suite split and the failure moving between the two tests.

## Impact

[...]

Issue #4255, opened 2026-07-28
trusty-search: test suite registers into the operator's live indexes.toml and never cleans up


bobmatnyc:
The trusty-search test suite writes index registrations into the live `indexes.toml` and does not remove them at teardown. 18 dead `.trusty-search-test-roots/*` entries were present on 2026-07-28.

Evidence — five of these stale entries caused a real outage the same day. Each dead path blocked the warm-boot loop for the full 60s `TRUSTY_WARMBOOT_INDEX_TIMEOUT_SECS`; five consumed the entire boot window, so zero of the seven real indexes warm-booted:
```
warm-boot: index 'skip-kg-reregister' restore TIMED OUT (>60s) — path /Users/…/.trusty-search-test-roots/ts-2984-skip-kg-hCDgvq...
warm-boot DEGRADED: 5 index(es) TIMED OUT during restore ... loaded=0 skipped_timeout=5
```
That took trusty-review (which requires trusty-search) offline for ~13.5 hours.

Resolution: tests must not register into the operator's live `indexes.toml` — use an isolated registry path — or must tear down everything they register. Also prune the 18 existing dead entries.

Related: #4250, which tracks the separate no-self-heal/latched-flag defect this exposed.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
#4864 (squash `da1e8dcb`, trusty-search 0.42.3) closed the in-process case: it made the live index registry unreachable from tests running in the same process.

#4870 (squash `b4d2375b`) completes it for spawned child processes. `trusty_common::test_harness::running_under_test_harness()` decides from the running executable's path — cargo test binaries live under `target/<profile>/deps/<name>-<hash>`, but a spawned `tcode` binary resolves to `target/<profile>/tcode`, outside `deps/`, so the child reported "not a test" and warmed its tempfile fixture project into whatever trusty-search daemon it found. `crates/trusty-common/src/test_harness.rs:26-29` documented this exact boundary and said such a child must be told explicitly; nothing on main closed it until now.

Measured on a clean `origin/main` worktree: `cargo test -p trusty-code --tests` took the live index registry from 80 to 88 entries, all eight new roots `$TMPDIR/.tmp*` fixtures — no real project paths. With #4870 applied, the same run: 88 to 88, zero new roots.

[...]

Issue #2984, opened 2026-07-18
design(trusty-search): per-index, runtime-manageable component flags for KG and vector/semantic indexing


bobmatnyc:
# design(trusty-search): per-index, runtime-manageable component flags for KG and vector/semantic indexing

## Motivation

Bob (2026-07-18): KG construction and vector/semantic indexing cause the most operational problems in trusty-search — embedder wedges (#1448/#1450), EMFILE/FD pressure (#2914, #2947), warm-boot memory cost, KG build cost — and must be **individually manageable**. **Requirement: per-index granularity, runtime manageable** — enable/disable each component on a live index without daemon restart and without index recreation.

## Current state (verified against origin/main)

**More exists than expected — but it's create-time only:**
- `lexical_only: bool` (issue #109, v0.9.0) — suppresses Stage 2 (embed) AND Stage 3 (KG) together. No independent "vector off, KG on" mode.
- `skip_kg: bool` (issue #313, v0.17.0) — suppresses Stage 3 only, orthogonal to `lexical_only` (`core/registry.rs:397`).
- Creation-time surfaces: CLI `--no-kg` (`commands/index.rs:47,138`), per-index YAML `skip_kg:` (`core/repo_config.rs:112`), HTTP `POST /indexes` body (`service/server/router.rs:148`), machine default `TRUSTY_NO_KG=1` applied at creation only (`service/server/indexes.rs:420`).
- **MCP gap:** `create_index` MCP tool exposes NEITHER flag (`mcp/tools/index.rs`, `types.rs` — zero references). Only HTTP/CLI/YAML can set them.
- **Runtime immutability is deliberate:** `PATCH /indexes/:id/config` (#1372) explicitly preserves `existing.skip_kg`/`existing.lexical_only` (`index_config.rs:228`); reindex preserves `skip_kg: handle.skip_kg` (`reindex_handlers.rs:173`). Changing either flag today = delete + recreate.
- **Query-time degradation already works and is the reuse point:** `search_capabilities` grows per stage (`core/registry.rs:137`) and the search handler consults it, not static flags (`service/server/search.rs:278`). A runtime toggle only needs to flip a stage to `Skipped` — the query path already degrades gracefully.
- KG-absent contract: `call_chain` returns structured 503 `kg_unavailable/skipped_by_config` (README.md:322-328).

**Component asymmetry (matters for phasing):**
[...]

PR #5065, opened 2026-08-07
feat(trusty-mpm): index a worktree when it is created, BM25+KG only

### LINKED ISSUE #5060: tm: index a worktree when it is created, BM25+KG only
## Defect

A worktree only becomes searchable when a managed session is *launched* in it — `core::session_launch::register_project_index` runs at launch, never at creation. Most worktrees never reach that point in a state anyone queries, so a bare `search` from a worktree answers `404 unknown index`, which reads as permanent and pushes agents into silent grep fallbacks.

## Confidence

Observed — measured this session against the live daemon.

## Evidence

[#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) measured 4 of 70+ live worktrees for this repo carrying a registered index. Both worktree-minting paths in trusty-mpm (`daemon::managed_routes::inproject::create_session_worktree`, `provisioner::workspace`'s `RealGitBackend::worktree_add`) return without touching trusty-search.

Cost measurement, cold BM25+KG-only build of this workspace (5,865 tracked files → 75,282 chunks) against the live daemon: walk started `03:47:27.259`, `lexical: ready` at `03:48:00.709` (33.5 s), `graph: ready` at `03:48:02.270` (35.0 s total). `semantic: skipped`, `search_capabilities: ["bm25","literal","exact_match","kg"]`.

## Resolution

Index at creation, on a detached thread, BM25 + KG only (`skip_vector: true`). Exact text and the symbol graph are branch-specific and must be worktree-accurate; conceptual similarity is not, so the embedding lane stays on the base checkout. 35 s is nowhere near unnoticeable, so there is no sync-under-threshold path — creation and launch never wait on indexing at any repo size.

Complements [#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) (doctor visibility) and the not-ready error contract on `fix/4715-index-not-ready`: this reduces how often the not-ready answer is the answer; neither replaces the other.

bobmatnyc:
## Defect

A worktree only became searchable when a managed session was *launched* in it — `core::session_launch::register_project_index` runs at launch, never at creation. Most worktrees never reach that point in a state anyone queries, so a bare `search` from a worktree answers `404 unknown index`, which reads as permanent and pushes agents into silent grep fallbacks.

## Evidence

[#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) measured 4 of 70+ live worktrees for this repo carrying a registered index. Both worktree-minting paths in trusty-mpm returned without touching trusty-search.

Cost measurement, cold BM25+KG-only build of this workspace (5,865 tracked files → 75,282 chunks) against the live daemon:

```
walk started    2026-08-07T03:47:27.259Z
lexical ready   2026-08-07T03:48:00.709Z   (33.5 s)
graph   ready   2026-08-07T03:48:02.270Z   (35.0 s total)
semantic        skipped
search_capabilities  ["bm25","literal","exact_match","kg"]
```

## Resolution

Register the index at worktree creation, on a detached thread, BM25 + KG only.

**Paths covered.** `daemon::managed_routes::inproject::create_session_worktree` (in-project managed sessions, `<project>/.worktrees/<name>`) and `provisioner::workspace`'s `RealGitBackend::worktree_add` (clone-provisioned sessions). Deliberately not covered: an ad-hoc `git worktree add` an agent runs itself — no trusty-mpm code path observes it, the same blind spot #2867's push-guard hook documents.

**BM25 + KG only.** `skip_vector: true`. Exact text and the symbol graph are branch-specific and must be worktree-accurate; conceptual similarity is not, so the embedding lane stays on the base checkout. trusty-search already supports this per-index (#2984 Phase 1) — **no trusty-search change was needed**, and none is made here.

**Never blocks.** 35.0 s is nowhere near unnoticeable, so there is no sync-under-threshold path. The one `spawn_detached` helper owns the guarantee and is the only place it can be tested (see the mutation evidence below).

**Idempotent.** `POST /indexes` is find-or-create and the reindex trigger is freshness-gated, so a repeat call is two short HTTP round trips and no walk.

**Opt-in.** No new bypass and no new grant: `allow_sensitive_path: false` keeps trusty-search's full hard denylist armed, exactly as `register_project_index` already does for the same directories. What this narrows is the input — only a path trusty-mpm itself just created as a worktree reaches the daemon, and a plain clone is refused, so worktree creation never registers a checkout on the operator's behalf.

**Deletion.** Unchanged and not re-implemented: `session_manager::search_gc` derives the same id from the same `resolve_project_root` + `derive_index_id` pair and deletes it at decommission, with a periodic orphan sweep behind it (#2033).

`trusty_common::search_index` gains `IndexOptions` + `ensure_project_indexed_with`; `ensure_project_indexed` is now a one-line wrapper and `IndexOptions::default()` reproduces its previous behaviour exactly.

## Gates

Rung 4 (cross-crate — `trusty-common` is a shared library).

| Gate | Result |
|---|---|
| `bash scripts/check_line_cap.sh` | 3747 files, 0 violations |
| `bash scripts/check_test_pointers.sh` | 22127 citations, 0 dangling |
| `bash scripts/check_sld.sh` | 0 errors, 0 warnings |
| `cargo clippy -p trusty-common --features search-index --all-targets -- -D warnings` | clean |
| `cargo check --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui` | clean |
| `cargo test -p trusty-common --features search-index` | 318 passed, 0 failed, 6 ignored |
| `cargo test -p trusty-mpm` | 4739–4740 passed, 0 change-related failures |
| `cargo test -p trusty-code` (consumer) | 1668 passed, 0 change-related failures |

`--exclude trusty-mpm-gui --exclude trusty-code-gui`: both fail `tauri::generate_context!` because `ui/dist` is not built in this environment. Pre-existing — `git diff --name-only origin/main...HEAD -- crates/trusty-mpm-gui/ crates/trusty-code-gui/` returns 0 files.

Three load-induced flakes appeared across full-suite runs, each with **0 diff files in its crate/module** and each passing 3/3 in isolation. The set varied between runs, which is the flake signature rather than a regression:

- `daemon::managed_routes::tests::stale_assets_for_many_*` — the known [#5040](https://github.com/bobmatnyc/trusty-tools/issues/5040) baseline red. My only edit under `managed_routes/` is one call in `create_session_worktree`; the failing tests reference it zero times. 4/4 in isolation.
- `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` — 3/3 in isolation.
- `trusty-code`'s `task_run_rejects_a_second_overlapping_run` (daemon-exit timeout) — 3/3 in isolation.

## Mutation evidence

Each regression test was proven to fail against the reverted behaviour, then restored.

**1. Relax `is_git_worktree` to `.git` merely existing** (the obvious simplification — a plain clone is still "a git repo"):

```
test ...skips_path_that_is_not_a_worktree ... FAILED
test ...is_git_worktree_matches_git_on_disk_shapes ... FAILED
  assertion `left == right` failed: a primary clone must not be auto-indexed
test result: FAILED. 5 passed; 2 failed
```
restored → `test result: ok. 7 passed; 0 failed`

**2. Drop `skip_vector` from the `POST /indexes` body** (every worktree silently embeds again):

```
test search_index::tests::create_index_request_body_sets_skip_vector ... FAILED
test result: FAILED. 20 passed; 1 failed
```
restored → `test result: ok. 21 passed; 0 failed`

**3. Join the spawned thread** (worktree creation waits on the walk):

```
test ...spawn_detached_returns_before_work_finishes ... FAILED
  spawn must not wait on its work; returned after 3.058219s
test result: FAILED. 7 passed; 1 failed
```
restored → `test result: ok. 8 passed; 0 failed`

Mutation 3 is why the non-blocking property is pinned on `spawn_detached` rather than on the public entry point. The first attempt asserted timing on `index_new_worktree_in_background` directly and **did not catch the join** — under the test harness the daemon write is refused instantly (#4255), so the real work finishes in microseconds whether joined or not. Routing through a helper whose work the test supplies makes the join fail loudly.

Closes #5060


bobmatnyc:
## code-critic: WARN

Adversarial review of `ca9e488f` against the spec in [#5060](https://github.com/bobmatnyc/trusty-tools/issues/5060). One HIGH, no CRITICAL. Code proceeds; the finding table travels with it.

CI was not re-run. `daemon::managed_routes::tests::stale_assets_for_many_*` is the [#5040](https://github.com/bobmatnyc/trusty-tools/issues/5040) baseline red and is correctly attributed in the PR body. `bash scripts/check_test_pointers.sh` re-run here: `resolved 22140 Test: citation(s) — 0 dangling pointers — OK`, and all 13 new `Test:` names resolve to exactly one definition each.

### Ruling on the new fail-open call site

**Acceptable to ship.** The Fail-Open Check's first question is whether anything advances past the failure. Here nothing does. `index_new_worktree_in_background` (`crates/trusty-mpm/src/core/worktree_index.rs:91`) discards the returned id — it is used for a log line and nothing else. No `.mcp.json` pin, no watermark, no "done" marker. The damaging half of the [#5091](https://github.com/bobmatnyc/trusty-tools/issues/5091) contract — a pin advancing for an index that was never created — is not reachable from this call site. The state a failed call leaves behind is "worktree has no index", which is exactly the status quo it is trying to improve on, so there is nothing to lose. With [#5093](https://github.com/bobmatnyc/trusty-tools/issues/5093) detecting the symptom and [#5091](https://github.com/bobmatnyc/trusty-tools/issues/5091) tracking the contract, the deferral is the right call.

Two qualifiers, both in the findings below:

1. The code logs success it never observed (finding 2). That is the one respect in which this is worse than before: the PR adds an observability layer whose whole stated purpose is to make skip reasons distinguishable, and it cannot distinguish the failure [#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) measured at ~94%.
2. `worktree_index.rs:80-86` claims a failure "IS discoverable" through `tm doctor`'s existing `search` check. That check probes the index id derived from the *current* project directory only, so it finds nothing for the other 70 worktrees — which is why [#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045) exists. Point the sentence at [#5093](https://github.com/bobmatnyc/trusty-tools/issues/5093) instead of at a check that does not sweep.

### Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-mpm/src/core/worktree_index.rs` | 22-33 | Cost model covers wall-clock only; the dominant cost is disk. Every registered index writes a colocated `<root>/.trusty-search/index.redb`, so this multiplies on-disk index storage by the number of live worktrees. `skip_vector` drops `hnsw.usearch` but not the corpus, which is the bigger file. | Measure it and state a disk budget beside the 35 s; then either gate creation-time indexing on repo size or route worktree indexes to non-colocated storage. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/worktree_index.rs` | 188-191, 95-101 | `ensure_project_indexed_with` always returns `Some(id)`, so `Registered` is returned — and `info!("worktree index registered")` logged — when the daemon was unreachable or answered non-2xx. The enum's own doc (`:46-56`) says it exists so each outcome is distinguishable; it cannot distinguish the one that actually happens. | Have `ensure_project_indexed_with` report whether the daemon call succeeded and add a `DaemonUnreachable` variant. Minimum: log "requested", not "registered". | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/session_launch/search_index.rs` | 113-114 | BM25+KG-only is not an invariant. Session launch calls `ensure_project_indexed(project_dir, false)` against the SAME worktree path (`session_launch/mod.rs:1071`), with `skip_vector` defaulting to `false`. `POST /indexes` short-circuits on an existing id (`trusty-search/src/service/server/indexes.rs:235`), so whichever call lands first decides. Creation-time fails, is skipped, or loses the race → the worktree gets a vector-bearing index, persisted. | `register_project_index` passes `skip_vector: crate::core::worktree_index::is_git_worktree(project_root)`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/session_manager/search_gc.rs` | 184-199 | The orphan sweep deletes any `.worktrees`-rooted index that no session record claims and that reports `chunk_count == 0`. This PR creates the index *before* the record carries `workspace_path`, and the index reads 0 chunks for the whole ~33.5 s lexical walk; the sweep runs every 60 s (`daemon/mod.rs:452`). Pre-PR the index was minted at launch, when the record already claimed the path. Consequence folds into the row above: launch re-creates it with the vector lane on. | Give `is_orphan_index` a grace window (skip indexes younger than one sweep interval), or register only once the record claims the path. | Fix here |
| MEDIUM | `crates/trusty-common/src/search_index.rs` | 46-58 | `IndexOptions` is public in a published crate with no `#[non_exhaustive]`. The daemon already carries a third orthogonal flag this bag omits — `skip_kg` (`trusty-search/src/service/server/router.rs:43`) — so a third field is near-certain, and adding it later breaks any external struct-literal construction. This crate has been bitten by a SemVer break before. | Add `#[non_exhaustive]`; construct as `IndexOptions { skip_vector: true, ..Default::default() }` at `worktree_index.rs:184`. | Fix here |
| LOW | `crates/trusty-mpm/src/core/worktree_index.rs` | 170-177 | `index_id` is derived only to test it for emptiness. `ensure_project_indexed_with` runs the identical check and returns `None`, which the next `match` maps to `EmptyIndexId`. Eight lines, no behaviour. | Delete them. | Fix here |

### Evidence for the HIGH

Measured on this machine, existing colocated stores for this repo:

```
3.0G    <project>/.trusty-search                              (base checkout)
        └─ index.redb        2.35G     ← corpus + KG; skip_vector does NOT shrink it
        └─ hnsw.usearch       815M     ← the only part skip_vector removes
        └─ hnsw.keys.json      60M

1.2G    .worktrees/index-on-create/.trusty-search
821M    .base/.worktrees/2eb72dca-.../.trusty-search
666M    .worktrees/tm-trusty-tools-05/.trusty-search

git worktree list | wc -l  →  57
```

`create_index_handler` forces `colocated: true` (`indexes.rs:335`), so the store lands inside each worktree. There is no embeddings table in the redb schema (`core/corpus/tables.rs`) — vectors live in `hnsw.usearch` — so a `skip_vector` worktree still writes a corpus in the same order of magnitude as the three measured above. At 57 live worktrees that is tens of GB, and the PR's cost section measures only the 35 s.

The cost is reclaimed when the worktree directory is removed, and `search_gc` plus the [#2033](https://github.com/bobmatnyc/trusty-tools/issues/2033) orphan sweep handle the registration side, so this is a peak-usage problem rather than a leak. It still deserves a stated number before merge.

### What was checked and found clean

- **`IndexOptions` shape and additive compatibility.** `Default` is `(false, false)`, `ensure_project_indexed` is a faithful one-line wrapper, and `index_options_default_matches_legacy_ensure_call` pins that as a contract. The only external caller of the two-arg form is `trusty-code/src/run_task/mod.rs:127`, unchanged. Version bumps belong to release time, so no bump is owed in this PR.
- **`skip_vector` end to end.** `create_index_request_body` → `CreateIndexRequest.skip_vector: Option<bool>` (no `deny_unknown_fields`, so an older daemon ignores it) → `indexes.rs:331` `unwrap_or(false)` → persisted in `indexes.toml` → preserved by `reindex_handlers.rs:215` and restored by `start_restore.rs:386`. The create path holds. The gap is the launch-time re-register above, not the wire.
- **Concurrency at the new call site.** No shared mutable state; one detached thread per creation, each doing two joined HTTP calls under 1 s / 2 s caps. Worktree basenames are `tm-<project>-NN` (in-project) or the session UUID (provisioner), so no cross-repo index-id collision. Reindex fan-out is bounded by trusty-search's 2-permit interactive / 1-permit background semaphores — no thundering herd.
- **`is_git_worktree`.** `.git`-is-a-file correctly separates a worktree from a primary clone, and `registers_index_for_real_worktree` pins the id to the worktree's own basename rather than the base repo's. The submodule overlap is real and correctly called out as harmless.
- **Mutation evidence.** All three mutations are the right ones, and the note about why the non-blocking assertion had to move onto `spawn_detached` rather than the public entry point is the sharpest thing in the PR.
- **Deletion.** `search_gc::disposable_workspace_index_id` derives the same id from the same pair, so teardown genuinely is already owned.

Issue #5073, opened 2026-08-07
test(trusty-code): spawns_projectless_when_the_tui_is_projectless fails under parallel execution, passes in isolation


bobmatnyc:
Sibling of [#4931](https://github.com/bobmatnyc/trusty-tools/issues/4931) in `trusty-code`.

```
cargo test -p trusty-code
test result: FAILED. 20 passed; 1 failed; 0 ignored
---- cli::daemon_autospawn::daemon_autospawn_tests::spawns_projectless_when_the_tui_is_projectless stdout ----
panicked at crates/trusty-code/src/cli/daemon_autospawn_tests.rs:385:51:
stub must have recorded its argv: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

In isolation, 3/3 pass:

```
run1 EXIT=0 test result: ok. 1 passed; 0 failed; 20 filtered out
run2 EXIT=0 test result: ok. 1 passed; 0 failed; 20 filtered out
run3 EXIT=0 test result: ok. 1 passed; 0 failed; 20 filtered out
```

The test spawns a stub binary and then reads the argv file the stub writes. `NotFound` at line 385 means the read raced the stub's write, or a parallel sibling reused/cleared the same stub path. Order-dependent, not a regression.

Observed while running `trusty-code` as a direct consumer gate for [#5072](https://github.com/bobmatnyc/trusty-tools/issues/5072), whose diff touches only `crates/trusty-common/src/catchup/`.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bob-duetto:
Candidate implementation is present on `codex/tcode-r1-spec-reconcile` at `f082b384e`: the test now waits for the stub PID marker before reading argv and uses panic-safe process cleanup. The full Trusty Code suite passed after the change. Keep open until the candidate branch lands.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Fix in progress — 2026-08-31 drive, trusty-code batch (parking lifted). #5439 leads as the only P1.

bobmatnyc:
[...]

PR #5074, opened 2026-08-07
fix(trusty-common): stop catch-up returning only the session it cannot date (#5072)

### LINKED ISSUE #5072: fix(trusty-mpm): session_context_catchup returns only the ONE snapshot it cannot date, dropping every well-formed one
## Defect

`session_context_catchup` returns an inverted paused-session digest: the only session that survives the watermark filter is one whose pause timestamp could not be derived, while every well-formed snapshot is dropped.

Two independent bugs compose:

1. `session_finder.rs:292` derives `paused_at` **only** from a `session-YYYYMMDD-HHMMSS.md` filename. A hand-written snapshot (`session-20260730-bounce.md`) yields `None`. There is no mtime fallback — and the field's own doc comment advertised a `## Paused At:` header fallback that was never implemented.
2. The watermark filter, duplicated verbatim in `catchup/mod.rs:168` and `catchup/json.rs:196`, is `filter(|s| s.sort_key().is_none_or(|ts| ts > wm))`. `is_none_or` returns `true` for an unknown key, so an undatable session is admitted by **every** watermark, forever.

Fail-open shape: the timestamp derivation fails, the failure is downgraded to `None`, and the filter treats `None` as "newer than the watermark". The digest reports healthy while carrying only the one record it could not date.

## Evidence

Live call on this repo, 2026-08-07:

```json
{
  "recent_commits": [],
  "recent_memory": [],
  "resolved_snapshot": ".../.trusty-mpm/sessions/session-20260806-211305.md",
  "sessions": [{
    "format": "trusty-mpm",
    "source_file": ".../.trusty-mpm/sessions/session-20260730-bounce.md",
    "summary": "", "paused_at": null, "in_progress": null,
    "next_steps": null, "git_context": null, "tmux_window": null
  }],
  "watermark_advanced": false
}
```

- Watermark: `~/.trusty-mpm/projects/bobmatnyc-trusty-tools/catchup-state.json` → `2026-08-07T04:25:08Z`.
- 100 snapshots in that directory; 99 have parseable filenames and all predate the watermark, so all 99 are dropped. `session-20260730-bounce.md` has no parseable stamp and is kept.
- Its fields are empty because it is hand-written and uses none of the parsed section headers (`## What landed this leg`, `## The bounce being performed now`, …). The parser is correct on it; the bug is that it should never have been the only survivor.
- `resolved_snapshot` disagrees with `sessions[0].source_file` because it uses a different resolver (`sessions-log.jsonl`, which is correct) — the disagreement is a symptom, not a second defect.
- `recent_commits: []` and `recent_memory: []` are **correct**: `git log --since=2026-08-07T04:25:08Z` is empty (newest commit is `8ab2c147`, 2026-08-06T15:24Z). Not a bug; no change made there.

Reproduction, as a test: `generate_catchup_json_excludes_undatable_session_behind_watermark` — one dated snapshot plus one undated file behind a 2099 watermark. Pre-fix the digest contains the undated file; post-fix it is empty.

## Resolution

- `parse_trusty_mpm_session` falls back to the file's mtime when the filename carries no stamp.
- Extract the duplicated predicate into `session_finder::filter_sessions_since` and make it fail-closed: an undatable session is dropped from a watermark-filtered digest, with a stderr warning naming the drop. `full=true` (no watermark) still returns everything.

bobmatnyc:
## Defect

`session_context_catchup` returned an inverted paused-session digest: the only session surviving the watermark filter was the one whose pause instant could not be derived, while every well-formed snapshot was dropped.

Two bugs compose, and the second is the fail-open shape:

1. `catchup/session_finder.rs:292` derived `paused_at` **only** from a `session-YYYYMMDD-HHMMSS.md` filename. A hand-written snapshot (`session-20260730-bounce.md`) yields `None`. No mtime fallback — and the field's own doc comment advertised a `## Paused At:` header fallback that was never implemented.
2. The watermark filter, duplicated verbatim in `catchup/mod.rs:168` and `catchup/json.rs:196`, was `filter(|s| s.sort_key().is_none_or(|ts| ts > wm))`. `is_none_or` returns `true` for an unknown key, so an undatable session was admitted by **every** watermark, forever, while every datable one was correctly filtered out.

Timestamp derivation fails → downgraded to `None` → the filter reads `None` as "newer than the watermark" → state advances and the digest reports healthy while carrying only the record it could not date.

## Evidence

Live call on this repo, 2026-08-07:

```json
{
  "recent_commits": [], "recent_memory": [],
  "resolved_snapshot": ".../.trusty-mpm/sessions/session-20260806-211305.md",
  "sessions": [{
    "format": "trusty-mpm",
    "source_file": ".../.trusty-mpm/sessions/session-20260730-bounce.md",
    "summary": "", "paused_at": null, "in_progress": null,
    "next_steps": null, "git_context": null, "tmux_window": null
  }],
  "watermark_advanced": false
}
```

Watermark `2026-08-07T04:25:08Z`; 100 snapshots in that directory, 99 with parseable filenames, all predating it. The undatable one is the sole survivor. Its fields are empty because it's hand-written and carries none of the parsed section headers — the parser is correct on it; the bug is that it was the only survivor.

`resolved_snapshot` and `sessions[]` **legitimately disagree** under a recent watermark: they answer different questions — "what should I resume from" versus "what paused since your last catch-up". What this PR removes is the *inversion*, not the disagreement. A caller that treats a future disagreement as a defect will be chasing correct behavior.

`recent_commits: []` and `recent_memory: []` are **correct, not a bug**: `git log --since=2026-08-07T04:25:08Z` is empty (newest commit `8ab2c147`, 2026-08-06T15:24Z). A non-`full` call reports only what postdates the watermark. Unchanged.

**Red before, green after.** Against the pre-fix code:

```
test catchup::json::tests::generate_catchup_json_excludes_undatable_session_behind_watermark ... FAILED
test catchup::session_finder::tests::filter_sessions_since_drops_undatable_session ... FAILED
test catchup::session_finder::tests::find_orders_undated_snapshot_by_mtime ... FAILED
test catchup::session_finder::tests::parse_falls_back_to_mtime_when_filename_lacks_timestamp ... FAILED

a snapshot with no derivable timestamp must not survive a watermark that every
datable snapshot fails: [Some(".../.trusty-mpm/sessions/session-20260730-bounce.md")]

  left: None    (paused_at — an undated filename must fall back to mtime)
  left: 2       (filter — only the datable, newer-than-watermark session survives)
  left: "session-20260101-000000.md"  (ordering — undated-but-newer must sort first)

test result: FAILED. 73 passed; 4 failed; 1 ignored
```

After: `test result: ok. 81 passed; 0 failed; 1 ignored`.

## Review round — two HIGHs, both in the new failure branch

Review found the first commit reintroduced the shape it fixes, one layer up. Fixed in `4c70450b8`:

**The withheld session had no receipt.** The drop was reported only by `eprintln!` to process stderr; the markdown digest rendered the byte-identical `"No paused sessions since last catch-up."`, and `run_catchup` (via `session_launch/mod.rs:1280`) advanced the watermark anyway. Every later watermark is newer, so a withheld session left every future window permanently — a failure branch advancing state and reporting success, with nothing telling the operator to re-run with `full`. `filter_sessions_since` now returns a `FilteredSessions` receipt; `CatchupJson` and the MCP payload expose `undatable_sessions_dropped`; the digest names the count and the recovery.

**The mtime rescue was asymmetric.** Fail-closed filtering applies to both `PausedSession` arms, but only `TrustyMpm` got the fallback. A claude-mpm JSON carrying only `session_id` deserialises with `paused_at: None` (`roundtrip_partial_json_uses_defaults` pins that), so it went from appearing in every digest to being silently withheld from all of them — a regression on the arm the fix did not rescue. Worse, the one test pinning the new fail-closed direction *used* such a session and asserted the drop was correct. `parse_session_file` now stamps `source_mtime`, `sort_key` falls back to it, and that test uses a session with no file behind it — the only genuinely undatable shape left.

Also noted: mtime tracks the last **write**, not the pause, so editing an old hand-written snapshot re-dates it and promotes it to the head of the digest. Documented on `mtime_utc`. It does not fire on a fresh clone — `.trusty-mpm/` and `.claude-mpm/` are gitignored, so no snapshot carries a checkout mtime.

**Isolation matrix** — each new guard independently pinned; one mutation, exactly one red:

```
revert ClaudeMpm mtime rescue (.or(session.source_mtime) → .or(None))
  → claude_mpm_session_with_no_paused_at_is_dated_by_mtime ... FAILED   (80 passed; 1 failed)
stop counting withheld sessions (dropped_undatable += 1 → += 0)
  → filter_sessions_since_reports_dropped_count ... FAILED              (80 passed; 1 failed)
collapse the notice back to one string (if dropped == 0 → if true)
  → no_sessions_notice_distinguishes_empty_from_withheld ... FAILED     (80 passed; 1 failed)
restored → test result: ok. 81 passed; 0 failed; 1 ignored
```

## Resolution

- Both `PausedSession` arms fall back to the file's mtime when the recorded timestamp is missing or unparseable.
- The duplicated predicate becomes one `session_finder::filter_sessions_since` returning a receipt, fail-closed: an undatable session is withheld and *counted*, never silently admitted. `full=true` still returns everything.
- `undatable_sessions_dropped` on `CatchupJson` and the MCP response distinguishes "nothing paused" from "sessions existed and could not be dated".

## Second review round — receipt wiring, and the mixed case

**Finding 1 — the wire was unpinned.** The producer (`filter_sessions_since`) and the renderers were each pinned; nothing pinned the path between them. Substituting `0` at `json.rs`, `mod.rs`, or the `mcp_context` payload key left the suite green. No end-to-end test can cover it: once both `PausedSession` arms fall back to mtime, an undatable session is unreachable through the filesystem. Four pure seams, one test each — `sessions_payload`, `CatchupJson::absorb`, `render_sessions_section`, `catchup_payload`.

**Finding 2 — the markdown digest emitted the receipt only when nothing survived.** Markdown is the surface that advances the watermark, so the mixed case (one kept, one withheld) advanced silently — hiding the loss in exactly the run where a populated digest reads as complete. The notice is now unconditional on a non-zero count, matching what JSON already did.

`CatchupJson::absorb` also replaces the hand-rolled per-project merge in `mcp_context`, where the count sums while every other field concatenates.

**Finding 3 (LOW)** — `mcp/tools/session.rs` tool description and `tm-session-resume.md`'s sample payload now carry `undatable_sessions_dropped` and state that `sessions` and `resolved_snapshot` answer different questions.

Wire mutation matrix — one red each, the named test every time:

```
sessions_payload returns 0 instead of the count
  → sessions_payload_propagates_dropped_count ... FAILED                    (83 passed; 1 failed)
absorb overwrites instead of summing
  → absorb_sums_dropped_counts_and_concatenates_the_rest ... FAILED         (83 passed; 1 failed)
markdown receipt back to else-on-empty (&& filtered.kept.is_empty())
  → sessions_section_reports_withheld_alongside_kept ... FAILED             (83 passed; 1 failed)
mcp_context payload key hardcoded to 0
  → catchup_payload_carries_the_undatable_drop_count ... FAILED             (5 passed; 1 failed)
restored → test result: ok. 84 passed; 0 failed; 1 ignored
```

## Release note — this is a MINOR bump, not a patch

`CatchupJson` gains a public field and `ClaudeMpmSession` gains `source_mtime`. Both are `pub` with all-`pub` fields and no `#[non_exhaustive]`, so any external struct-literal construction or exhaustive destructuring breaks: **`trusty-common` 0.29.0 → 0.30.0**, the same shape as the ChatEvent trap that forced 0.27.0. Additive-safe over the JSON wire; nothing in-tree breaks. Version file untouched here — that is `local-ops`.

## Gates at `4bb3499e5`

```
bash scripts/check_line_cap.sh   → 3747 files, 7 allowlisted, 0 violations — OK
cargo fmt --check                → clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings  → EXIT=0
cargo test -p trusty-common --features catchup → 402 passed; 0 failed; 8 ignored
cargo test -p trusty-mpm                       → 4732 passed; 2 failed (pre-existing, #4931)
```

The two failures are `ensure_managed_config_dir_emits_the_frozen_skill_warning` and `stale_assets_for_many_sees_an_agent_change_on_the_very_next_call` — both named verbatim in [#4931](https://github.com/bobmatnyc/trusty-tools/issues/4931).

## Gates — rung 4 (shared library `trusty-common` + both direct consumers)

```
bash scripts/check_line_cap.sh   → 3747 files, 7 allowlisted, 0 violations — OK
cargo fmt --check                → clean
cargo check --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui  → EXIT=0
cargo clippy -p trusty-common --features catchup --all-targets -- -D warnings  → EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings                        → EXIT=0
cargo test -p trusty-common --features catchup → 399 passed; 0 failed; 8 ignored
cargo test -p trusty-mpm                       → 4731 passed; 2 failed (pre-existing, #4931)
cargo test -p trusty-code                      → 1600 + 21 + 3 passed; 0 failed
```

Pre-existing failures, none caused by this diff (the diff is five files under `crates/trusty-common/src/catchup/`, one `crates/trusty-mpm/src/daemon/mcp_context.rs`, and two changelog fragments):

- **`trusty-mpm-gui` / `trusty-code-gui`** fail `cargo check` with `frontendDist "ui/dist" doesn't exist`. `git ls-files crates/*-gui/ui/dist` returns 0 — never committed, so any fresh worktree fails identically until `pnpm build` runs. Environmental.
- **`ensure_managed_config_dir_emits_the_frozen_skill_warning`** and **`stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet`** — both named verbatim in [#4931](https://github.com/bobmatnyc/trusty-tools/issues/4931), both pass in isolation (`EXIT=0`, 1 passed each), occurrence appended there.
- **`spawns_projectless_when_the_tui_is_projectless`** ([#5073](https://github.com/bobmatnyc/trusty-tools/issues/5073)) failed on the earlier run and passed on this one — consistent with the order-dependence in that ticket.

Closes #5072


bobmatnyc:
## Verdict: WARN

Independent adversarial review, own detached worktree at `f2a5c286`. Zero CRITICAL, two HIGH.

The core fix is right and the red-then-green is real — I reproduced it rather than accepting the transcript. Both HIGH findings are about the *new* failure branch, which is what this repo reviews harder than the original bug.

### Red-before, reproduced independently

Restored both pre-fix behaviors in place (mtime fallback removed; predicate back to `is_none_or`), keeping the refactor so it compiles:

```
test result: FAILED. 73 passed; 4 failed; 1 ignored
```

and the `json` failure reproduces the live evidence verbatim:

```
a snapshot with no derivable timestamp must not survive a watermark that every
datable snapshot fails: [Some(".../.trusty-mpm/sessions/session-20260730-bounce.md")]
```

Head is green: `cargo test -p trusty-common --features catchup catchup::` → `77 passed; 0 failed; 1 ignored`.

**Isolation matrix** — each guard mutated alone:

| Mutation | Result |
|---|---|
| Both reverted | `73 passed; 4 failed` |
| Only the fail-closed predicate reverted | `76 passed; 1 failed` — `filter_sessions_since_drops_undatable_session` |
| Only the mtime fallback reverted | `75 passed; 2 failed` — the two mtime tests |

Both guards are independently pinned. Note the middle row: exactly one test pins the new fail-closed direction, and its input is a `ClaudeMpm` session — see HIGH-2.

---

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/catchup/mod.rs` | 250–263 | The watermark advances past a session `filter_sessions_since` just dropped, and the digest reports healthy while it happens | Return the drop count from `filter_sessions_since`; render it in the digest and add `undatable_sessions_dropped` to `CatchupJson` | Fix here |
| HIGH | `crates/trusty-common/src/catchup/session_finder.rs` | 67–75, 355–363 | Fail-closed applies to both `PausedSession` arms; the mtime rescue applies only to `TrustyMpm`. A `ClaudeMpm` session with no `paused_at` used to be shown in every digest and is now silently dropped | Give `ClaudeMpm` the same fallback — `mtime_utc` on its JSON path, or the filename stamp `mpm_session.rs:136-140` already computes | Fix here |
| MEDIUM | `crates/trusty-common/src/catchup/session_finder.rs` | 355–363 | mtime is last-modified, not pause time — touching an old hand-written snapshot promotes it to the head of the digest and through every watermark | Prefer the `sessions-log.jsonl` entry when one exists; at minimum say in the `paused_at` doc that mtime is a lower-confidence key | Fix here |
| LOW | `crates/trusty-common/src/catchup/session_finder.rs` | 157–164 | `eprintln!` rather than `tracing::warn!`, and it prints one identical line per dropped session | Use `tracing::warn!`, or aggregate to one line with a count | Parent |

---

### HIGH-1 — the cursor advances past the dropped item

`session_launch/mod.rs:1280` is the auto-inject path:

```rust
let ctx = crate::core::catchup::run_catchup_blocking(opts, true);   // advance_watermark = true
```

The sequence inside is:

1. `generate_catchup_context` calls `filter_sessions_since`, which drops the undatable session and `eprintln!`s to process stderr.
2. `mod.rs:166-168` then renders **`"No paused sessions since last catch-up."`** — byte-identical to the output when there genuinely were none.
3. `run_catchup` (`mod.rs:252-263`) unconditionally writes `last_catchup_at = Utc::now()`.

Every subsequent auto-catchup carries a newer watermark, so the dropped session is outside every future window. Permanently.

That is Fail-Open Check item 1 (a cursor moves forward past a failed item) and item 2 (name the alarm, then break it — the alarm here is the digest itself, and it reports healthy). It is the same class as the defect being fixed, one layer up, which is exactly the recurrence this repo reviews for.

Not CRITICAL: the snapshot file is untouched on disk and `full=true` recovers it. But the operator has no signal telling them to pass `full=true` — the MCP payload at `daemon/mcp_context.rs:110-116` carries `sessions`, `recent_commits`, `recent_memory`, `resolved_snapshot`, `watermark_advanced`, and no drop count. The `eprintln!` lands in the daemon's stderr, never in the MCP response the PM reads.

The stderr warning is not a sufficient receipt. The count belongs in the returned payload.

```rust
// session_finder.rs
pub fn filter_sessions_since(
    sessions: Vec<PausedSession>,
    watermark: Option<DateTime<Utc>>,
) -> (Vec<PausedSession>, usize)   // (kept, dropped_undatable)
```

then `mod.rs:166` renders `"N paused session(s) had no derivable pause timestamp and were omitted — re-run with full=true"`, and `CatchupJson` gains `undatable_sessions_dropped: usize` surfaced in the tool result.

### HIGH-2 — the two arms are asymmetric

`sort_key()` (`session_finder.rs:67-75`):

```rust
PausedSession::TrustyMpm { paused_at, .. } => *paused_at,          // now has an mtime fallback
PausedSession::ClaudeMpm { session } => session
    .paused_at.as_deref().and_then(|s| s.parse::<DateTime<Utc>>().ok()),   // no fallback
```

`paused_at: None` is a real, loadable state on the `ClaudeMpm` arm, not a hypothetical — `mpm_session.rs:218-227` (`roundtrip_partial_json_uses_defaults`) asserts a JSON carrying only `session_id` deserializes with `paused_at.is_none()`. And `mpm_session.rs:136-140` already computes a filename fallback for its own sort, so the rescue is sitting right there unused.

Before this PR such a session was admitted by every watermark — noisy, but never lost. After it, the session is dropped, silently (HIGH-1), and the watermark walks past it. That is a regression the fix introduces on the arm it did not rescue.

Fail-Open Check item 3: compare sibling branches; asymmetry between arms of one state machine is the tell, and the arm that was left alone is the bug.

### MEDIUM-1 — what mtime actually means

The fallback is sound on untouched data. I checked the live directory rather than assuming:

```
2026-07-30T10:55:02 EDT  session-20260730-bounce.md     (= 2026-07-30T14:55:02Z)
2026-08-06T17:13:05 EDT  session-20260806-211305.md     (= 2026-08-06T21:13:05Z — matches its filename to the second)
```

All 99 dated files' mtimes agree with their filenames. The bounce file recovers an instant matching its own intended date. So the fallback does fire, and to a meaningful value.

**The worktree-checkout concern does not fire here, but the premise is correct.** `git checkout` does stamp every file it writes — every file in my fresh worktree reads `2026-08-07T00:58:19`, identical. It doesn't matter because `.trusty-mpm/*` is gitignored (`.gitignore:69`) and `git ls-files | grep '.trusty-mpm/sessions/'` returns 0, so git never writes a snapshot.

What remains is mutation, not checkout. `find_orders_undated_snapshot_by_mtime` pins that an undated snapshot with a newer mtime sorts **first**. So an editor save, a `cp` without `-p`, or an archive restore on a hand-written 2024 snapshot puts it ahead of today's real one in `render_resume_context` and admits it through every watermark. mtime answers "when was this file last written", and the code treats it as "when did this session pause".

**Is fail-closed the right direction at all?** Yes, on the `TrustyMpm` arm — with the mtime fallback in place, `paused_at: None` requires a filesystem that cannot report a modification time, so fail-closed costs nothing and the "sessions since T" contract stays honest. On the `ClaudeMpm` arm it is currently wrong, because there is no fallback to make the antecedent rare (HIGH-2). Fix HIGH-2 and fail-closed becomes right in both arms.

---

## Verified, not flagged

- **Duplication genuinely collapsed.** `grep -rn 'is_none_or\|is_some_and'` across `trusty-common/src` and `trusty-mpm/src`: no remaining `sort_key()`-based predicate. The only `catchup` hits are `palace.rs:144,184`, which use `is_some_and` — already the fail-closed direction — plus two doc/comment references. No third copy.
- **`trusty-mpm` shim is pure re-export.** `crates/trusty-mpm/src/core/catchup/mod.rs` is two `pub use trusty_common::catchup::*` lines and nothing else.
- **`full=true` still returns everything.** `json.rs:186-190` sets `watermark = None` on `opts.full`; `session_finder.rs:151-153` returns untouched on `None`; pinned by `filter_sessions_since_keeps_everything_without_watermark`. `full` is plumbed from the MCP tool at `mcp_context.rs:52,96`.
- **Symptom 3 — the engineer's judgment is correct.** Newest commit `8ab2c147` is `2026-08-06T11:24:10-04:00` = `15:24:10Z`; watermark `2026-08-07T04:25:08Z`. `git log --since=2026-08-07T04:25:08Z` on that worktree returns nothing. `recent_commits: []` is correct watermark behavior. The digest is not still broken on this axis.
- **`resolved_snapshot` vs `sessions[]`** — one correction to the framing. The two are computed by different resolvers with different contracts: `resolved_snapshot` is unconditional, `sessions[]` is watermark-filtered. They disagree by construction whenever the watermark is recent, and no test pins agreement — nor can a general one. What the fix removes is the *inversion*: `sessions[]` no longer holds a file `resolved_snapshot` would never pick. Post-fix on the live directory, `sessions[]` is empty and `resolved_snapshot` is `session-20260806-211305.md` — consistent, not agreeing. Worth saying that way in the PR body rather than "the two now agree."

## Gates run

```
cargo test -p trusty-common --features catchup catchup::  → 77 passed; 0 failed; 1 ignored
cargo check --workspace --all-targets --exclude trusty-mpm-gui --exclude trusty-code-gui
  && cargo clippy -p trusty-common --features catchup --all-targets -- -D warnings
  && cargo fmt --check                                    → EXIT=0
cargo test -p trusty-mpm                                  → 4732 passed; 1 failed; 7 ignored
cargo test -p trusty-code                                 → 1600 + 21 + … all green, 0 failed
bash scripts/check_line_cap.sh                            → 3747 files, 7 allowlisted, 0 violations
```

Both direct consumers of the changed `trusty-common/catchup` module are covered — `trusty-mpm` and `trusty-code` are the only two crates enabling the `catchup` feature.

**Pre-existing failures — checked, not accepted on assertion.** `git diff --name-only origin/main...HEAD` is six files, all under `crates/trusty-common/src/catchup/` plus two changelog fragments; the diff touches **zero** files in `crates/trusty-mpm/src/`, `crates/trusty-code/`, or either `*-gui` crate.

- `*-gui` `frontendDist "ui/dist" doesn't exist` — reproduced; `git ls-files 'crates/*-gui/ui/dist'` returns 0. Environmental.
- `daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` — failed in the full suite, `4 passed; 0 failed` in isolation. #4931.
- `spawns_projectless_when_the_tui_is_projectless` — did not reproduce; `trusty-code` was fully green here. Consistent with the order-dependent flake filed as #5073.

## Required Changes

1. Surface the drop as a receipt the caller can see: `filter_sessions_since` returns `(Vec<PausedSession>, usize)`; `generate_catchup_context` renders the count instead of `"No paused sessions since last catch-up."`; `CatchupJson` gains `undatable_sessions_dropped` and `mcp_context.rs` includes it in the tool result. A test asserting the count is non-zero when a session is dropped.
2. Give `PausedSession::ClaudeMpm` the same timestamp fallback `TrustyMpm` got, so fail-closed never fires on a legacy session that merely lacks `paused_at`. A test covering a `ClaudeMpm` with `paused_at: None` surviving a watermark it postdates.
3. Correct the `paused_at` doc at `session_finder.rs:31-35` to say mtime is last-modified, not pause time, and note that editing a snapshot re-dates it.

Not merged.

bobmatnyc (commented):
## Verdict: APPROVE

Focused re-review of the two HIGH fixes at `4c70450b8`. Both are real fixes, and both prior HIGHs are cleared. Three MEDIUM/LOW findings remain, none blocking.

### Mutation matrix — re-run independently, one mutation at a time

`cargo test -p trusty-common --features catchup catchup`

| Mutation | Result |
|---|---|
| baseline | `test result: ok. 81 passed; 0 failed; 1 ignored` |
| drop `.or(session.source_mtime)` in `sort_key` | `FAILED. 80 passed; 1 failed` — `claude_mpm_session_with_no_paused_at_is_dated_by_mtime` |
| drop `out.dropped_undatable += 1` | `FAILED. 80 passed; 1 failed` — `filter_sessions_since_reports_dropped_count` |
| collapse `render_no_sessions_notice` to one string | `FAILED. 80 passed; 1 failed` — `no_sessions_notice_distinguishes_empty_from_withheld` |

Claimed matrix confirmed: one red each, and the named test is the one that fails.

Also clean: `cargo fmt --check -p trusty-common -p trusty-mpm && cargo clippy -p trusty-common --features catchup --all-targets -- -D warnings` → `EXIT=0`; `scripts/check_line_cap.sh` → `0 violations`; `cargo test -p trusty-mpm --lib mcp_context` → `5 passed; 0 failed`.

### On the deliberate non-blocking of the watermark

The argument holds. `CatchupState.last_catchup_at` is one instant shared by the session, git, and palace windows, so holding it holds all three. An undatable record is undatable permanently — its mtime does not become readable later — so a fail-closed hold would never clear itself, and every subsequent catch-up would re-report every commit since the last successful advance, growing without bound. Meanwhile the withheld session is omitted, not destroyed: the file is still on disk and `full=true` returns it in full.

Fail-closed is warranted when the advance makes the item **unrecoverable**. Here it does not. "Do not advance *silently*" is the correct reading, and a receipt satisfies it — provided the receipt reaches the operator on the path that actually advances. That proviso is where findings 1 and 2 land.

### Verified clean

- **Every `ClaudeMpmSession` construction path stamps `source_mtime`.** Both loaders (`load_latest_claude_mpm_session` `mpm_session.rs:103`, `load_all_claude_mpm_sessions` `mpm_session.rs:136`) route through `parse_session_file`, which stamps at `mpm_session.rs:177`. Every other construction site in the workspace is a test fixture. `crates/trusty-mpm/src/core/claude_mpm_session.rs` re-exports only.
- **No third arm.** `PausedSession` has exactly two variants — `sort_key` matches both with no wildcard, so exhaustiveness is compiler-enforced. Both are rescued: `TrustyMpm` at `session_finder.rs:412`, `ClaudeMpm` at `session_finder.rs:85`.
- **The pure-function seam is justified.** After this PR `sort_key() == None` requires `fs::metadata()` to fail on a file whose contents `fs::read`/`read_to_string` just returned successfully — a TOCTOU delete or a filesystem with no mtime. Genuinely unreachable through normal filesystem state, so extracting `render_no_sessions_notice` as a pure function is the right call and an integration test is not owed.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-common/src/catchup/mod.rs` | 195-200 | The markdown digest emits the withheld receipt **only when `filtered.kept` is empty**. This is the one surface that advances the watermark (`run_catchup` → `generate_catchup_context` → `save_catchup_state` at `mod.rs:289`). If one session survives the filter while another is withheld, the operator gets an ordinary digest, no signal, and the watermark advances past the withheld record. The JSON surface reports unconditionally — the asymmetry between the two is the tell, and the fail-open arm is the one that advances state. | Emit the withheld notice regardless of whether `kept` is empty — append it after `render_resume_context` rather than only in the `else`. | Fix here |
| MEDIUM | `crates/trusty-common/src/catchup/json.rs` / `mod.rs` | 207 / 199 | The receipt **wiring** is unpinned at every consumer surface. Replacing `filtered.dropped_undatable` with `0` at *both* sites simultaneously leaves the suite fully green: `test result: ok. 81 passed; 0 failed; 1 ignored`. The three claimed mutations pin the producer (`filter_sessions_since`) and the renderer (`render_no_sessions_notice`) but nothing pins the wire between them, so the count can be severed from both surfaces without a single test noticing. Same at `crates/trusty-mpm/src/daemon/mcp_context.rs:111` — the shape test at `mcp_context.rs:275-279` asserts five keys, none of them the new one. | Apply the same pure-seam trick already used for the notice: extract `fn sessions_section(found: Vec<PausedSession>, wm: Option<DateTime<Utc>>) -> (String, usize)` in `mod.rs` and the equivalent `(Vec<PausedSessionJson>, usize)` assembly in `json.rs`, then drive each with the existing in-memory `undatable_session()` fixture. That pins the wiring and covers finding 1 in the same test. Add `undatable_sessions_dropped` to the `mcp_context` shape assertions. | Fix here |
| LOW | `crates/trusty-mpm/src/mcp/tools/session.rs` | 267-275 | The `session_context_catchup` tool description enumerates what the response carries and no longer does. The consumer is an LLM PM, for which the description **is** the contract; the sample payload at `crates/trusty-mpm/src/assets/skills/tm-session-resume.md:103-120` is likewise missing the field. The value does reach the JSON body and is self-describing, so this is incompleteness, not invisibility. | One clause in the tool description naming `undatable_sessions_dropped` and what non-zero means; one line in the skill's sample payload. | Fix here |

### Notes

- **Why finding 1 is MEDIUM and not HIGH.** It is textbook fail-open shape — the alarm can report healthy while the loss occurs — and the Fail-Open Check calls that HIGH by construction. I am downgrading on reachability only: before this PR the undatable branch was reachable (a hand-written `session-20260730-bounce.md`), and after it, it is not. I can assert the shape with certainty and cannot assert the impact, so the 80% filter says downgrade. If you judge the TOCTOU/exotic-FS route more reachable than I have, HIGH is defensible and I would not argue.
- **`CatchupJson` and `ClaudeMpmSession` are both `pub` with all-`pub` fields, no `#[non_exhaustive]`.** Adding a field is additive-safe over the JSON wire (unknown keys are ignored, and the MCP client is JSON), but it breaks any external crate that builds either type by struct literal. `trusty-common` is published, so this is a minor-version break for 0.29.0 → 0.30.0 rather than a patch — the same shape as the ChatEvent 0.26.3/0.27.0 trap. Not a finding against the PR, since version bumps belong to `local-ops` at release time; flagged so it reaches whoever cuts it. In-workspace the only consumer (`mcp_context.rs:108`) reads and never constructs, so nothing in-tree breaks.
- Changelog fragments present and correctly formed for both crates.
- Out of scope per dispatch and not re-examined: the original red-then-green reproduction, symptom 3, the mtime-in-a-worktree question, and the three pre-existing failures.

Issue #5038, opened 2026-08-07
fix(trusty-memory): injected drawer preview renders provenance tags and full tag lists uncapped


bobmatnyc:
Confidence: Reproduced.

Defect, two symptoms of the same unbudgeted render path in `crates/trusty-memory/src/commands/prompt_context/format.rs` (drawer-preview rendering):

1. `creator:client`, `creator:version`, `creator:source`, `creator:cwd` render into every drawer preview. `creator:cwd` alone is a ~90-character absolute path, repeated per drawer, ~5 drawers per firing.
2. Per-drawer tag lists are rendered in full with no cap. Drawers with 15+ tags were observed injected whole; the tag list frequently exceeds the drawer's own content preview length.

Resolution:
1. Filter the `creator:*` predicates out of the injected preview — rendering change only, they remain in storage.
2. Cap the rendered tag count, or drop tags from the injected preview entirely. Tags help a human browsing `memory_list`; they add little for a model reading a terse recall snippet.

Do not quote a token-savings estimate in this issue or its fix: an earlier estimate of 450-750 tokens/firing for symptom 1 alone fails its own sanity check against a measured mean total firing size of ~680 tokens. State the mechanism; let the implementer measure the actual saving.

Independent of #5036 (BM25 fusion) and #5037 (score floor / adaptive K) — this is a rendering-budget fix that stands on its own regardless of match quality.

Related to #4889 (T6, restructures `compose_injection` into four independently-capped sections with fixed byte allocations): that issue caps injection at the *section* level; this issue caps content at the *drawer* level. Both apply — #4889's section caps would still waste budget on provenance noise and an oversized tag list without this fix. Not a duplicate.

Related, not duplicated: #4836, #4904 (closed), #4971/#4972/#4973 (open, other ranking-input defects), #4892 (epic).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Governed by the injection-contract design recorded on #5037 — see that issue for the full four-part ruling.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #5036, opened 2026-08-07
trusty-memory: HTTP recall path is vector-only — existing RRF BM25 fusion never wired into the UserPromptSubmit hook


bobmatnyc:
Confidence: Reproduced.

Defect: `handle_memory_recall` (`crates/trusty-memory/src/tools/memory_ops.rs::handle_memory_recall`) runs vector + BM25 in parallel via `tokio::join!` and fuses through `fuse_bm25_into_recall` (`crates/trusty-memory/src/tools/bm25.rs`, RRF k=60). The path the `UserPromptSubmit` hook actually uses does not: `commands/prompt_context/fetch.rs` → `GET /api/v1/palaces/{slug}/recall` → `web/recall_routes.rs` → `service/core.rs` → `trusty-common/src/memory_core/retrieval/layers.rs`. No BM25 anywhere on that chain. `handle_memory_recall_deep` (`memory_ops.rs`) also lacks fusion, even though its sibling `handle_memory_recall` has it.

Evidence: a one-word prompt returns 5 unrelated drawers at full confidence; a prompt containing "memory" retrieves every memory-flavored drawer because there is no lexical counterweight to the vector centroid; short keyword-bearing curated facts retrieve poorly, which is dense retrieval's known weak spot and BM25's strength. #4970 raised vector-only self-retrieval recall@1 from 25.5% to 72.8% — that ceiling is for a vector-only system and does not add a lexical anchor.

Open question, unresolved here: whether the BM25 index is populated for existing palaces or built lazily per-palace on first supervisor use (`ensure_bm25_running_for_palace`, `bm25_supervisor.rs`). If lazy/unpopulated, fusion contributes nothing silently — a fail-open shape.

Resolution: wire the existing `fuse_bm25_into_recall` into `MemoryService::recall`, and add the same fusion to `handle_memory_recall_deep`. Reuse the existing implementation; do not write a second one.

An engineer is already implementing this on branch `fix/recall-http-bm25-fusion`.

Related: #4836, #4904, #4971, #4972, #4973, #4889, #4892 (none of these cover the missing HTTP-path fusion). This is the root cause behind two of four sibling issues filed alongside it — those note the relationship in their own bodies.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Sibling issues filed alongside this root cause:

[...]

Issue #4972, opened 2026-08-05
over half of hook queries are silently truncated at the embedder's 512-token window


bobmatnyc:
Defect: trusty-memory's `prompt-context` hook sends the raw user prompt as the query. 51.5% of real queries exceed the embedder's 512-token input window and are truncated with no warning, no metric, and no signal to the caller. The embedding therefore represents a prefix of the query, not the query.

Evidence: measured over the logged corpus at `~/Library/Application Support/trusty-memory/logs/enriched-prompts.*.jsonl` (1,114+ firings) during [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904).

Compounding factor: 63% of those queries are `task-notification` blobs whose first ~647 characters are envelope boilerplate, so the surviving prefix is disproportionately non-content. Together these mean the hook frequently embeds machine framing rather than user intent.

Resolution options: truncate deliberately from the semantically dense end rather than implicitly from the tail; strip known envelope boilerplate before embedding; chunk-and-max-pool; or at minimum emit a metric so the truncation rate is visible.

bobmatnyc:
Governed by the injection-contract design recorded on #5037 — see that issue for the full four-part ruling.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Owner ruled this IS a [tm 1.3.5](https://github.com/bobmatnyc/trusty-tools/milestone/26) blocker: his [#5037](https://github.com/bobmatnyc/trusty-tools/issues/5037) comment made all four injection-contract parts required and non-subsumable.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5100, opened 2026-08-07
fix(trusty-memory): give prompt-context recall a relevance floor, a bigger budget, and a withheld signal (#5037)

### LINKED ISSUE #5037: fix(trusty-memory): prompt-context recall has no confidence floor — unconditional top-K regardless of match quality or query signal
Confidence: Reproduced.

Defect: `retrieve_l2` (`crates/trusty-common/src/memory_core/retrieval/layers.rs::retrieve_l2`) sorts candidates by score then calls `results.truncate(top_k)` with no minimum-score gate; `recall()`'s final merge does the same. Separately, `DEFAULT_TOP_K` (`crates/trusty-memory/src/commands/prompt_context/mod.rs::DEFAULT_TOP_K` = 5, overridable only by `TRUSTY_MEMORY_PROMPT_TOP_K`, clamped [1,20]) is fixed regardless of query length or retrieval confidence. A query with no meaningful match still returns a full K results, formatted identically to high-confidence ones, and a low-signal turn ("go", "yes", "ok") pays the same ~700-token injection as a substantive question.

Evidence: a one-word prompt "go" returned 5 unrelated session-pause/checkpoint drawers. Confirmed on origin/main.

Related root cause: #5036 (HTTP recall path is vector-only, no BM25 fusion) is the primary fix for this symptom shape — once lexical fusion is wired in, match quality should improve and this becomes tuning rather than the primary fix. Still worth doing independently: even a well-fused recall can legitimately have no good match for a given prompt, and K should reflect that.

Resolution: add a configurable minimum-score threshold below which candidates are dropped, even if that returns fewer than K or zero; and skip injection entirely below a signal threshold, or scale K to confidence. Standing context is L0/L1 and always-on, so skipping per-turn L2 recall does not lose it. Tune against the self-retrieval ground-truth corpus PR #4970 already built (400 drawers queried by their own text) plus the enriched-prompts log — do not guess a constant.

Related, not duplicated: #4836 (closed, MCP-path query-drop, different bug), #4904 (closed, general ranking/relevance parent), #4971/#4972/#4973 (open, other ranking-input defects), #4889/#4892 (injection-budget restructuring, different layer).

bobmatnyc:
Closes #5037

## Defect

`prompt-context` asked for `top_k` drawers and rendered whatever came back. `truncate(top_k)` is the only length control in the whole retrieval chain (`layers.rs::recall_scoped` ~667, `retrieve_l2_scoped` ~397) and no layer applies a minimum score. `rescore_l1_by_similarity` gives an essential drawer the vector search never returned `importance * L1_NO_SIMILARITY_PENALTY` — at most `0.15` — which is nonzero, so it competes for a slot.

Probe against the live 1,332-drawer palace: "what is the capital of France" returned five drawers all at exactly 0.15, formatted identically to a genuine 0.56 hit.

## The threshold, from measurement

Two probes against the live daemon, script in the session scratchpad.

| population | n | min | p10 | median | max |
|---|---|---|---|---|---|
| 15 off-topic prompts × top-10 | 150 | 0.1500 | 0.1500 | 0.1510 | **0.3439** |
| 60 self-retrieval queries, score of the *correct* drawer | 57 | **0.4844** | 0.6117 | 0.6950 | 0.9743 |
| 120 real logged hook prompts × top-10 | 1200 | 0.4042 | 0.4671 | 0.5422 | 0.7527 |
| 12 paraphrased on-topic questions, best hit per query | 12 | 0.3293 | — | 0.4317 | 0.5359 |

75 of the 150 off-topic candidates scored exactly 0.1500 — the L1 penalty itself. Floor sweep:

```
floor=0.25  noise kept  18/150 (12.0%)  noise queries still firing  3/15  correct-drawer retained 57/57
floor=0.30  noise kept   5/150 ( 3.3%)  noise queries still firing  1/15  correct-drawer retained 57/57
floor=0.35  noise kept   0/150 ( 0.0%)  noise queries still firing  0/15  correct-drawer retained 57/57
floor=0.45  noise kept   0/150 ( 0.0%)  noise queries still firing  0/15  correct-drawer retained 57/57
```

**0.35** is the smallest swept value at which no off-topic candidate survives, rounded up from the highest observed noise score (0.3439). At it the real logged corpus loses 0 of 1200 candidates and 0 of 120 queries.

The cost, stated plainly: 2 of 12 paraphrased on-topic questions stop injecting. Both are short keyword-bearing queries ("what is the MSRV for this workspace", 0.3293) — dense retrieval's weak spot and BM25's strength, which #5036 is the named fix for. The floor is configurable (`TRUSTY_MEMORY_PROMPT_MIN_SCORE`, `0` disables) and the loss is announced, not silent.

## The four requirements

1. **Relevance floor.** New `trusty_common::memory_core::retrieval::relevance` owns the one implementation. `RecalledDrawer` now parses the `score` the recall endpoint has always sent (`service::recall_entry_json`) and drops what falls below. Runs *after* the deny-tag filter so a tag-excluded drawer is never also counted as withheld. An unknown score is kept, never dropped.
2. **Larger max size.** `DEFAULT_TOP_K` 5 → 12 (override ceiling 20 → 30, compile-time asserted above the default), `INJECTION_BYTE_CAP` 4 KB → 8 KB. Measured over 80 real logged prompts, K=12 with the floor renders a drawer section of at most ~2.7 KB.
3. **Whole-input-only — already satisfied, now pinned.** `parse_user_prompt` returns the entire prompt and `fetch_palace_recall` sends it whole; nothing on that path truncates. Verified against the live daemon at query lengths 500 B → 64 KB (no URI failure). `recall_query_is_the_whole_prompt` pins it against the truncating `hook_prompt_excerpt` helper sitting a few lines away. The residual fragmentation is the embedder's 512-token window — scores plateau identically from ~5 KB up, visible in that same probe — which is #4972, a different layer.
4. **"More" signal.** When the floor drops drawers the injection says so and points at `memory_recall`, with distinct wording for a partial drop and for a recall that kept nothing. Zero candidates still renders nothing: an empty palace has nothing to announce.

`recall`/`recall_scoped` are deliberately unchanged. Gating inside `truncate(top_k)` would change every MCP and CLI recall caller's contract, which the ruling does not ask for.

## Scope

Included: **#5037** only. Left alone and stated rather than silently absorbed:
- **#5038** (uncapped provenance/tag rendering) — a rendering-budget fix in the same `format.rs`, independent of match quality. Its symptom is visible in the red-check output below (`creator:client=…`, `creator:version=…` on every bullet).
- **#5036** (BM25 fusion) — an engineer is already on `fix/recall-http-bm25-fusion` per the ticket.
- **#4972** (512-token truncation) — embedder layer; see requirement 3.

## Red before green

Isolation: on a scratch branch, set `DEFAULT_RELEVANCE_FLOOR = 0.0` (the defect: "no confidence floor") and revert `DEFAULT_TOP_K`/`INJECTION_BYTE_CAP`. Every test file unchanged.

```
---- relevance_floor_drops_all_noise_drawers stdout ----
0.15 L1-penalty drawers must not reach the injection; got 5

---- relevance_floor_keeps_high_scoring_drawer stdout ----
assertion `left == right` failed: the genuine hit must survive
  left: 5
 right: 1

---- prompt_context_off_topic_prompt_withholds_and_says_so stdout ----
off-topic prompt must not inject `tokio`; got:
## Relevant memories from palace `prompt-ctx-floor-e2e`
- Knowledge graph stores triples in redb with valid_from intervals per edge  _(tags: `kg`, `redb`, `creator:client=trusty-memory-mcp`, …)_
- Python bindings ship via PyO3 with custom ABI shims for the runtime  _(tags: `python`, `pyo3`, …)_
- Rust integration uses tokio for async tasks and serde for JSON encoding  _(tags: `rust`, `tokio`, …)_

test result: FAILED. 28 passed; 4 failed; 0 ignored; 0 measured; 585 filtered out
```

```
---- memory_core::retrieval::relevance::tests::relevance_floor_drops_l1_penalty_noise ----
0.15 noise must not survive the floor
test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 923 filtered out
```

The e2e failure reproduces the reported defect exactly: the France probe injecting three unrelated drawers at full formatting.

Green after restoring the fix: `test result: ok. 32 passed; 0 failed` (prompt_context filter).

## Gate — Rust Test Ladder rung 4

| gate | result |
|---|---|
| `cargo check --workspace --all-targets` * | EXIT=0 |
| `cargo clippy -p trusty-common --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-memory --all-targets -- -D warnings` | EXIT=0 |
| `cargo test -p trusty-common --features memory-core` | `ok. 914 passed; 0 failed; 13 ignored` |
| `cargo test -p trusty-memory` | `ok. 613 passed; 0 failed; 4 ignored` |
| `bash scripts/check_line_cap.sh` | `3763 files; 7 allowlisted, 0 violations — OK` |
| `bash scripts/check_changelog_fragment.sh` | `all 2 crate(s) with source changes are recorded` |
| `bash scripts/check_sld.sh` | `0 error(s), 0 warning(s)` |

\* with `SKIP_UI_BUILD=1 --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui`, exactly as `.github/workflows/ci.yml` runs every workspace job. Without the excludes the three Tauri crates fail on a missing `ui/dist`, unrelated to this diff (which touches zero files in them).

`-p trusty-common` alone runs with default features, which exclude `memory-core` — the new tests would not have run. `--features memory-core` is what actually covers the change.

No coverage was deleted, `#[ignore]`d, `cfg`-gated, or excluded to make anything green.

### DIFF
--- a/crates/trusty-common/changelog.d/5037-relevance-floor.md
+++ b/crates/trusty-common/changelog.d/5037-relevance-floor.md
@@ -1,0 +1,24 @@
+Added
+
+- **A relevance floor for recall results ([#5037](https://github.com/bobmatnyc/trusty-tools/issues/5037)).** `DEFAULT_RELEVANCE_FLOOR`,
+  `FloorOutcome`, and `apply_relevance_floor` in a new
+  `memory_core::retrieval::relevance` module. Every retrieval path in `layers.rs`
+  ended in `truncate(top_k)` and nothing else, so a query with no good answer
+  still returned a full `top_k` of whatever ranked highest — including L1
+  drawers scoring `importance * L1_NO_SIMILARITY_PENALTY`, at most `0.15`.
+  `apply_relevance_floor` is the one implementation of "below the floor, it is
+  not shown", and it returns the count of what it dropped so a caller can say so
+  rather than going silent. An item whose score is unknown is kept, never
+  dropped.
+
+  The default is `0.35`, picked from measured distributions against the live
+  1,332-drawer `trusty-tools` palace rather than guessed: 150 candidates from 15
+  off-topic prompts span 0.1500–0.3439 (75 of them at exactly 0.1500, the L1
+  penalty), while 57 self-retrieval correct-drawer hits span 0.4844–0.9743 and
+  1,200 candidates from real logged hook prompts span 0.4042–0.7527. `0.35` is
+  the smallest swept value at which no off-topic candidate survives.
+
+  `recall`/`recall_scoped` are deliberately unchanged: `truncate(top_k)` stays a
+  length cap, and gating inside it would change every MCP and CLI recall
+  caller's contract. Callers that must not show a weak match apply the floor to
+  what comes back.
--- a/crates/trusty-common/src/memory_core/retrieval/layers.rs
+++ b/crates/trusty-common/src/memory_core/retrieval/layers.rs
@@ -233,6 +233,9 @@ pub fn rescore_l1_by_similarity(
                 Some(&sim) => sim,
                 // Drawer was not in the HNSW results — likely off-topic.
                 // Apply penalty so importance alone can't dominate ranking.
+                // #5037: demoted is not excluded — this branch still yields a
+                // score that competes for a `top_k` slot. See
+                // `super::relevance::DEFAULT_RELEVANCE_FLOOR`.
                 None => r.drawer.importance * L1_NO_SIMILARITY_PENALTY,
             };
         }
@@ -664,6 +667,10 @@ pub async fn recall_scoped(
     // L1_CAP+1 entries before any L2 hits are merged; without this truncation
     // the caller receives more than top_k results whenever the palace has a
     // non-empty identity string plus several high-importance drawers.
+    // #5037: this stays a length cap, not a quality gate — callers that must
+    // not show a weak match apply `super::relevance::apply_relevance_floor` to
+    // what comes back. Gating here would silently change every MCP and CLI
+    // recall caller's contract, which #5037's ruling does not ask for.
     combined.truncate(top_k);
 
     handle.log_recall(query, &combined);
--- a/crates/trusty-common/src/memory_core/retrieval/mod.rs
+++ b/crates/trusty-common/src/memory_core/retrieval/mod.rs
@@ -18,6 +18,9 @@ mod embed_repair;
 mod embedder;
 mod handle;
 mod layers;
+// #5037: the minimum-score gate `layers` never had — `truncate(top_k)` was its
+// only length control.
+mod relevance;
 // ADR-0027 T9: room/wing recall scoping, shared by L2 and the list paths.
 mod scope;
 // ADR-0028 D3/D4/D5 (#4886): Tier C admission and atomic retire-on-write.
@@ -59,6 +62,10 @@ pub use embed_repair::{
 
 // Recall scoping (ADR-0027 T9)
 pub use scope::{RecallScope, list_drawers_in_wing, scope_admits};
+
+// #5037: relevance floor. Public because the consumer that must apply it —
+// trusty-memory's prompt-context injection — lives in another crate.
+pub use relevance::{DEFAULT_RELEVANCE_FLOOR, FloorOutcome, apply_relevance_floor};
 
 // ADR-0028 Tier C admission (#4886). The write half (`persist_with_retirement`)
 // stays crate-internal — every writer reaches it through
--- a/crates/trusty-common/src/memory_core/retrieval/relevance.rs
+++ b/crates/trusty-common/src/memory_core/retrieval/relevance.rs
@@ -1,0 +1,179 @@
+//! Relevance floor for recall results (#5037).
+//!
+//! Why: every retrieval path in [`super::layers`] ends in `truncate(top_k)` and
+//! nothing else — `top_k` is the only length control, so a query with no good
+//! answer still returns a full `top_k` of whatever ranked highest. This module
+//! is the missing half: a minimum score a candidate must clear to be shown at
+//! all, plus a count of what was dropped so the caller can say so out loud.
+//! What: [`DEFAULT_RELEVANCE_FLOOR`], [`FloorOutcome`], and
+//! [`apply_relevance_floor`]. Pure — no I/O, no allocation beyond the returned
+//! vector.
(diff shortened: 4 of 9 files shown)


bobmatnyc:
## Requirement 4 evidence — the withheld signal

Implemented at `crates/trusty-memory/src/commands/prompt_context/format.rs:109` (`withheld_notice`), called from `format.rs:63-66`, fed by `mod.rs:489` → `mod.rs:500`. Four rendered states, captured from a real `compose_injection` run:

```
STATE A — 1 kept, 3 withheld
## Relevant memories from palace `trusty-tools`
- worktrees branch off origin/main, never local main
_(3 further memories withheld below the relevance floor — call `memory_recall` for the full ranked set.)_

STATE B — 0 kept, 5 withheld
## Relevant memories from palace `trusty-tools`
_(No stored memories cleared the relevance floor for this prompt — 5 withheld as too weak a match. Nothing is missing from the palace; call `memory_recall` to search it directly.)_

STATE C — 0 kept, 0 withheld
""

STATE D — 0 kept, 1 withheld (singular)
_(No stored memory cleared the relevance floor for this prompt — 1 withheld as too weak a match. …)_
```

B and C are distinct states with distinct output. "5 were withheld" is not collapsed into "nothing matched": B names the count and says *nothing is missing from the palace*, C renders empty because an empty palace has nothing to announce.

### Isolation — the tests fail without it

Suppressed the notice only (floor still drops, reader never told), every test file unchanged:

```
---- compose_injection_announces_total_silence ----
an all-withheld recall must not render as an empty injection

---- compose_injection_announces_withheld_drawers ----
the withheld count must be visible; got:
## Relevant memories from palace `alpha`
- kept drawer 0
- kept drawer 1

---- prompt_context_off_topic_prompt_withholds_and_says_so ----
a withheld recall must announce itself, not go silent; got:
No prompt facts stored yet.

test result: FAILED. 30 passed; 3 failed; 0 ignored; 0 measured; 585 filtered out
```

The e2e line is the point. Without the notice, "what is the capital of France" against a populated palace renders `No prompt facts stored yet.` — byte-identical to an empty palace. That is the failure mode the floor introduces, and it is what requirement 4 closes.

No code changed for this comment; HEAD stays `9b5c89139ee270f58c293aaa91821a7a99cae894`.

Issue #5089, opened 2026-08-07
Implement ADR-0034: console webhook ingress relayed over UDS to a supervised on-demand process (1.3.5 criterion c)


bobmatnyc:
**Defect:** ADR-0032 removed every sibling daemon's HTTP listener but explicitly left open how an external GitHub webhook reaches a service that no longer listens and is not resident. That gap is what blocks milestone `tm 1.3.5` criterion (c) (undaemonize trusty-analyze/trusty-review). ADR-0034 (owner ruling 2026-08-07, "Console relays over UDS.") closes the design question; this issue is the implementation.

**Evidence:**

- Two webhook handlers assume an always-listening HTTP server: `crates/trusty-review/src/service/webhook.rs:141` and `crates/trusty-analyze/src/service/handlers/review.rs:128`.
- The two disagree on the unset-secret policy. `trusty-review` returns 401 and rejects every delivery (`webhook.rs:147-161`). `trusty-analyze` logs `"no webhook secret configured — skipping webhook signature verification"` and processes the payload anyway (`review.rs:153-157`) — a live fail-open.
- Both return `202` before the work runs, then `tokio::spawn` and downgrade failure to a log line (`review.rs:210-214`; `webhook.rs:305-309`). GitHub does not retry an acknowledged delivery, so a post-ack failure loses the event while health reports fine. Same shape as `ensure_project_indexed` (`crates/trusty-code/src/run_task/mod.rs:127`: `let _ = trusty_common::search_index::ensure_project_indexed(&project, true);`) and #5020's `in_flight` counter, measured pinned at 9 for ten days.
- The `0600` socket permission that ADR-0031 and ADR-0032 both cite as UDS's access-control advantage is not implemented: no production file in the workspace calls `set_permissions`/`PermissionsExt` on a socket (every hit is a test fixture). Sockets are created at the default umask, and `embedder_client/uds.rs` + `bm25_client.rs` place them in `$TMPDIR` with a `/tmp` fallback.
- The supervision mechanism ADR-0032 said "does not yet exist" does exist: `crates/trusty-memory/src/bm25_supervisor.rs` (`ensure_running`) already implements spawn-gate serialisation, socket adoption, LRU cap, RSS ceiling, backoff probing, and SIGTERM→SIGKILL, hardened by #2845/#2846.
- Cost is measured, not assumed: #5028 reports 191 ms cold start against a 36.7 s median review (0.5%), and zero webhook deliveries in 14 days.

[...]

PR #5112, opened 2026-08-07
fix(trusty-review): scope the dedup redb lock to one operation, not the process

### LINKED ISSUE #5064: fix(trusty-review): serve --stdio opens the daemon's redb dedup store, colliding with serve (HTTP)
**Defect:** `trusty-review serve --stdio` and `trusty-review serve` (HTTP) both call the same `build_app_state()` in `commands/serve.rs`, which unconditionally opens `dedup.redb`. redb takes an exclusive cross-process flock, so when both run against the same `--log-dir` the second opener's `DedupStore::open` fails with `DatabaseAlreadyOpen` — caught, swallowed at `warn!`, execution continues with `dedup: None`. Same shape #1078 fixed for trusty-search and trusty-memory by converting their stdio paths into pure proxies; trusty-review is the one first-party server never converted.

**Evidence:** `mcp/tools.rs` hardcodes `allow_posting: false` on every MCP tool call, so the dedup store's write path is unreachable today — blast radius is contained. #5028's production measurement corroborates independently: of `AppState`'s fields only `llm`, `verifier`, `search`, `analyze` are used on that path, `dedup` explicitly `None`. #3992 recorded 10s/1800s contention stalls from this shape on trusty-memory before #1078. Risk is structural, not observed: flip `allow_posting` for any future MCP tool, or add a read of existing dedup state, and it degrades silently.

**Resolution:** convert `serve --stdio` into a pure proxy to the running HTTP daemon, mirroring trusty-memory's `serve_stdio_bridge.rs`. Short term, since `allow_posting: false` makes the open provably unnecessary on the MCP path, skip it rather than attempt-and-swallow. Regression test: stand up `serve` and `serve --stdio` against the same `--log-dir` concurrently, assert the stdio process makes zero direct redb calls.

trusty-review's daemon is slated for removal (see the trusty-mcp epic), so this may end up resolved by that work — but it shouldn't wait on it, since the removal is a later epic and this is a live structural crack today.

bobmatnyc:
Closes #5064. Hard prerequisite for milestone `tm 1.3.5` criterion (c).

## Outcome

`DedupStore` held redb's exclusive file lock for its whole lifetime, so the second process to open `{LOG_DIR}/dedup.redb` got `DatabaseAlreadyOpen`. `build_app_state` downgraded that to a `warn!` and continued with `dedup: None` — a server running with no claim gate and every health signal green.

The lock's scope was the *process*; the thing it protects is a *transaction*. Every symptom follows from that mismatch.

## Why it blocks criterion (c), not only d2

#5064 was filed as a `serve --stdio` vs `serve` collision, resolvable by skipping the open on the MCP path (`allow_posting: false` there makes the store provably unreachable). Under the just-ratified ADR-0034 (#5090, implementation #5089) trusty-console spawns a webhook worker on demand; that worker and a concurrent stdio session both call `build_app_state` on one `--log-dir`, and the webhook path runs `allow_posting: true` and genuinely needs the store. Two posting-capable openers cannot be fixed by skipping one open.

## Fix shape: serialise

`DedupStore` owns a path, not an open `Database`. Each `claim` / `complete` / `release` opens redb, runs one write transaction, and drops it, so the exclusive lock is held for microseconds instead of for the process. A concurrent holder is waited out (2s budget, 20ms poll); a lock that never frees returns the new typed `DedupError::Contended`. An in-process `Mutex` serialises this crate's own callers, since redb's lock is per open-file-description and two handles inside one process contend exactly as two processes do.

On top of that, `build_app_state` now takes an explicit `DedupNeed`:

- `serve --stdio` → `NotNeeded`. Never touches the file.
- HTTP / webhook → `Required`. An open failure propagates as `Err`; it is never downgraded to `dedup: None`.

**Rejected: per-caller scoped stores.** Dedup's contract *is* shared state keyed on `(owner, repo, pr, head_sha)`. Private stores would each see an empty history and both post the same comment — that trades a lock collision for a duplicate-post defect. The key that would make them non-colliding is exactly the key they must share.

**Not the fail-open shape.** Every path either succeeds authoritatively or returns `Err`. There is no state in which a `DedupStore` exists and has silently stopped deduplicating, and `Required` no longer has a "continuing without it" branch.

## Out of scope

- Converting `serve --stdio` into a pure HTTP proxy (the issue's long-term note). Under ADR-0034 criterion (c) there is no resident daemon to proxy *to*, so that shape does not resolve the topology this PR has to serve.
- `cmd_run` passing `dedup: None` with `allow_posting: true` (`commands/run.rs:321`). Pre-existing, unrelated to the lock, and a behaviour change rather than a fix.
- `OutcomeStore`'s warn-and-continue in `service/webhook.rs:435`. Already operation-scoped; outcomes are telemetry, not a posting gate.

## Blast radius

`crates/trusty-review` only: `store/dedup.rs` and `commands/serve.rs`. `DedupStore`'s public method signatures are unchanged; `DedupError` gains one variant. `cargo check --workspace --all-targets` confirms no consumer breaks.

## Test evidence — Rust Test Ladder rung 5 (persistence + process lifecycle)

Nine new tests, seven of which fail against the pre-fix code. The defect is a concurrency defect, so the coverage is genuinely concurrent: 8 threads with independent handles racing one `claim`, plus a **re-exec'd second process** holding the real file lock.

**Red before green.** Pre-fix production sources restored from `origin/main` with the new tests attached (the new `open_dedup_for` / `DedupError::Contended` API shimmed in with the *pre-fix bodies*, verbatim, so the tests measure old behaviour rather than a missing symbol):

```
test result: FAILED. 8 passed; 6 failed; 0 ignored   (store::dedup::tests)
test result: FAILED. 1 passed; 3 failed; 0 ignored   (commands::serve::tests)

a second holder of the same dedup.redb must not be locked out (#5064): Open("Database already open. Cannot acquire lock.")
DedupStore must not hold redb's exclusive lock at rest (#5064): DatabaseAlreadyOpen
concurrent claim must not error under contention: dedup store open failed: Database already open. Cannot acquire lock.
lock-holder child exited early with exit status: 101
holder takes the exclusive lock: DatabaseAlreadyOpen
second handle must be able to open the file: DatabaseAlreadyOpen
stdio mode must not hold a dedup store
an unopenable required store must be an error, not `None`: None
assertion failed: first.is_some() && second.is_some()
```

**Green after.**

```
$ cargo test -p trusty-review store::dedup::tests::
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 1563 filtered out; finished in 2.37s

$ cargo test -p trusty-review commands::serve::tests::
test commands::serve::tests::stdio_mode_never_creates_the_dedup_file ... ok
test commands::serve::tests::http_mode_propagates_a_dedup_failure ... ok
test commands::serve::tests::http_mode_opens_the_dedup_store ... ok
test commands::serve::tests::two_posting_servers_can_share_one_log_dir ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out; finished in 0.13s
```

**Repeat run** (the concurrency tests are timing-sensitive): `store::dedup::tests::` run 10×, `pass=10 fail=0`.

**Gates.**

```
cargo check --workspace --all-targets                  EXIT=0   (14m 08s)
cargo clippy -p trusty-review --all-targets -D warnings EXIT=0
cargo test -p trusty-review                            EXIT=0   1572 passed, 0 failed, 5 ignored (+ 8 integration binaries, all ok)
cargo test -p trusty-review -- --include-ignored       EXIT=0   1577 passed, 0 failed, 0 ignored
bash scripts/check_line_cap.sh                         EXIT=0   3764 files, 0 violations
bash scripts/check_changelog_fragment.sh               EXIT=0
bash scripts/check_sld.sh                              EXIT=0   0 errors, 0 warnings
```

The workspace check needs `crates/trusty-code-gui/ui/dist` and `crates/trusty-mpm-gui/ui/dist` to exist (`pnpm install --frozen-lockfile && pnpm build` in each); a fresh worktree fails there with a Tauri `frontendDist` proc-macro panic that is environmental, not a code failure.

No baseline failures encountered. The known flakes #5040 and #4931 are in other crates and were not in scope.

## Docs / changelog

`crates/trusty-review/changelog.d/5064-dedup-flock-collision.md`. `dedup.rs`'s module doc gains a Locking section; `DedupStore`, `with_db`, `open_dedup_db_waiting`, `DedupNeed`, and `open_dedup_for` carry the full Why/What/Test pattern, since a locking and lifecycle contract is exactly the surprising-code case. Change sites are marked `// #5064:`.

The inline `mod tests` in `dedup.rs` moved to `store/dedup_tests.rs` so the new coverage has the 3000-SLOC test cap instead of pushing the production file toward 500. `dedup.rs` is 263 SLOC.

## Review-finding disposition

None yet — opening for the trusty-review gate.


bobmatnyc:
## Code Critic — [#5112](https://github.com/bobmatnyc/trusty-tools/pull/5112) (`Closes` [#5064](https://github.com/bobmatnyc/trusty-tools/issues/5064))

## Verdict: WARN

Rung 5 review, independent worktree off `origin/fix/5064-redb-flock-collision` (`4c48228e`). The root-cause diagnosis is right and the fix shape is right — process-lifetime lock replaced by operation-scoped lock, `DedupNeed` replacing the `warn! + dedup: None` branch. One HIGH: the diff's own new error variant lands on a caller that fail-opens, so the PR body's claim "there is no state where a `DedupStore` exists and has silently stopped deduplicating" is not true end to end.

## Findings

| HIGH | `crates/trusty-review/src/pipeline/runner.rs` | 268–270 | `Err(e) => warn!("dedup claim failed (proceeding without dedup): {e}")` — the only caller of `claim()` advances past the failure and posts. The new `DedupError::Contended` reaches this arm, so a stuck lock holder makes every webhook review post an ungated comment. `complete()` then also fails, so the claim is never recorded and the *next* redelivery posts again. | Fail closed on a posting-capable GitHub review: `result.infra_unavailable = true; result.verdict = Verdict::Unknown; result.error = Some(format!("dedup claim failed: {e}")); return abort_dry(result, config, &input, &deps);` — the existing early-abort exit at `runner.rs:386` already releases the claim and writes an inspectable dry-run log. | Fix here |
| HIGH | `crates/trusty-review/src/store/dedup.rs` | 22–24 vs 85–86 | Two contradictory contracts twelve lines apart in the same diff. Module doc: "the caller (the runner) is expected to *log and proceed* on error". `DedupError::Contended` doc: "The caller must never treat this as 'dedup is off'". Whichever one a future change follows, it can cite the file. | Delete the blanket log-and-proceed sentence and state the split: transaction/serde errors are best-effort, `Contended` on a posting-capable path is fatal to that review. | Fix here |
| MEDIUM | `crates/trusty-review/src/store/dedup.rs` | 173–193 | The #702 incompatible-format recovery (`rename` aside, then `Database::create`) used to run once per process at `open()`. Routing every `claim`/`complete`/`release` through `with_db` makes it run per operation, and it is not concurrency-safe: two processes that both read the stale v2 file can both rename — the second renames away the *fresh* database the first already committed claims into. The window is small but the trigger (warm boot after a binary upgrade with several openers) is exactly the ADR-0034 topology this PR serves. | Confine recovery to the initial probe: give `try_open_dedup_db` a `recover: bool` (or an `open`-only wrapper) so `with_db` never renames, and let a mid-life incompatible file surface as `DedupError::Open`. | Fix here |
| MEDIUM | `crates/trusty-review/src/store/dedup.rs` | 210–227, 316–325 | `with_db` holds a `std::sync::Mutex` and `std::thread::sleep`s up to `LOCK_WAIT_BUDGET` (2s) on whatever thread calls it. `claim` runs inside the `tokio::spawn`'d webhook task (`service/webhook.rs:239`, `:291`), so a stuck holder blocks a runtime worker for 2s per operation, three operations per review, serialised across in-process reviews by `gate`. Enough concurrent reviews and the HTTP surface itself stalls. | `tokio::task::spawn_blocking` around the three store calls in `runner.rs` / `post.rs` / `runner_helpers.rs`, or an async-aware wait. Not required to land this PR. | Parent |
| MEDIUM | `crates/trusty-review/src/pipeline/runner.rs` | 246–262 | A process that dies between `claim` and `complete`/`release` strands an `InProgress` record for `DEDUP_STALE_SECS` (7200s, `config/constants.rs:100`). Pre-fix this only blocked the one process that held the store; now every sibling process sharing the `--log-dir` is blocked, and each sees `Skipped` → `Verdict::Approve` + `error: "skipped: duplicate of a completed review"`. A crash therefore silently drops re-reviews for two hours and labels them approved. This is the intended cost of making dedup actually work cross-process, but it is a new blast radius and it is undocumented. | No code change required. Distinguish a stale-window skip from a genuine completed-duplicate skip in the log line, and record the 2h recovery window in the module doc. | Parent |
| MEDIUM | `crates/trusty-review/src/mcp/mod.rs` | 232, 267 | Third `AppState` construction site (`build_review_state`, the embedded MCP host) does not go through `build_app_state` and so never sees `DedupNeed`; it passes `dedup: None` with a prose justification. The justification checks out — `mcp/tools.rs:223` and `:280` are both `allow_posting: false` — but the whole point of `DedupNeed` is to make that declaration a type rather than a comment. | Lift `DedupNeed` out of `commands/serve.rs` into `store::dedup` and have `build_review_state` declare `NotNeeded` too. Also unblocks reusing it for [#5113](https://github.com/bobmatnyc/trusty-tools/issues/5113). | Parent |
| LOW | `crates/trusty-review/src/store/dedup.rs` | 148–149 | `OpenAttempt`'s doc names `two_stores_on_one_path_both_work` as the test for the `Locked` → `Ready` transition. That test never reaches `Locked` — neither store holds a lock at rest, so both opens succeed on the first attempt. The transition is actually covered by `concurrent_threads_claim_exactly_once` and `cross_process_holder_serialises_rather_than_locking_out`. | Swap the two test names in the `Test:` line. | Fix here |
| LOW | `crates/trusty-review/src/store/dedup.rs` | 190–192 | Sibling-branch asymmetry: the retry loop treats `DatabaseAlreadyOpen` as retryable `Locked`, but the recovery branch's `Database::create` maps every error including `DatabaseAlreadyOpen` to a hard `DedupError::Open`. Contention inside recovery bypasses the wait budget. | `Err(redb::DatabaseError::DatabaseAlreadyOpen) => Ok(OpenAttempt::Locked)` on that create too. Moot if the MEDIUM above is fixed by confining recovery to `open()`. | Fix here |

## Required Changes

1. `runner.rs:268` must not post a review whose dedup claim failed. Fail closed via `abort_dry`.
2. Reconcile the two contracts in `dedup.rs` (22–24 vs 85–86).
3. Confine the #702 recovery path to `DedupStore::open` so it stops running — and racing — on every operation.
4. The two LOW doc/branch fixes are one-liners; take them in the same pass.

## The five questions

**1. Open-per-operation and the 2s budget.** The budget is defensible for its stated purpose: each holder now keeps the lock for one open + write + fsync, single-digit milliseconds, so 2s absorbs on the order of a hundred queued operations. Two or three posting-capable processes will not exhaust it. What the budget does *not* survive is a holder that never releases — a pre-fix binary still running `serve --stdio` after the new daemon started, which is a live rolling-upgrade shape for this very fix. Then every operation costs 2s and returns `Contended`.

What happens to the in-flight review is the HIGH above: `runner.rs:269` logs a warning and proceeds, the review posts with `allow_posting: true` (`post.rs:164`), and `complete()` fails too so nothing is recorded. That is fail-open — a duplicate PR comment, then another on the next redelivery. The right choice is fail-closed: a dropped review is recoverable (redelivery, or the operator sees `infra_unavailable`), a duplicate comment is not. Note the daemon *startup* path already gets this right — `DedupNeed::Required` propagates `Err` and the server refuses to start. Only the per-operation path fail-opens.

**2. Atomicity across claim→complete→release.** A stranded claim is possible, and it was possible before this PR too — the durable `InProgress` record outlives the process, and the old process-lifetime handle never protected against a crash. What clears it is only the `DEDUP_STALE_SECS` window (7200s) or an explicit `release`. There is no cross-operation transaction and no lease renewal for a long-running review.

What this PR changes is who is blocked. Before, a stranded claim blocked only the single process that could open the store; every sibling had `dedup: None` and reviewed anyway. Now it blocks every process on the `--log-dir`, and the block is reported as `Verdict::Approve`. That is the MEDIUM above. It is the correct trade — that is what shared dedup means — but it is a real new failure mode and it is not in the changelog fragment.

**3. Fail-open check on `DedupNeed`, end to end.** The `warn! + dedup: None` branch is genuinely gone from `build_app_state`, and the mode declarations are correct: `mcp/tools.rs:223`/`:280` are both `allow_posting: false`, so `NotNeeded` on stdio is provable; `service/webhook.rs:291` is `allow_posting: true` and reads `state.dedup`, so `Required` is right. All construction sites: `commands/serve.rs:207` (both modes, now typed), `mcp/mod.rs:267` (third site, bypasses the type — MEDIUM above, but provably non-posting), `service/handlers.rs:637` (POST /review, `allow_posting: false` at `:649`), `commands/compare.rs:173` and `commands/calibrate.rs:382` (both false), `commands/run.rs:321` ([#5113](https://github.com/bobmatnyc/trusty-tools/issues/5113), not re-reported).

So the PR body's claim holds at *construction* and fails at *operation*. A `DedupStore` that exists and returns `Contended` on every call is a store that has silently stopped deduplicating, because the caller swallows it.

[#5113](https://github.com/bobmatnyc/trusty-tools/issues/5113): behaviourally unchanged — `commands/run.rs` is not in the diff. Strictly better in one way that matters, though: pre-fix, giving `cmd_run` a real store would have locked out any running daemon, so #5113 was not fixable without this PR. It is now. Slightly worse in one small way: `DedupNeed` is private to `commands/serve.rs`, so the #5113 fix cannot reuse the vocabulary without moving it first.

**4. The rejected alternative.** Sound. Dedup's contract is shared state keyed on `(owner, repo, pr, head_sha)` and its whole job is idempotency across processes and restarts. Per-caller stores would each start empty, both `claim` would return `Claimed`, and both would post — trading a lock collision for the exact defect the store exists to prevent. The key that would make them non-colliding is the key they must share. The only alternative that preserves the contract is a single-writer arbiter with the others going through RPC, and the PR is right that ADR-0034 ([#5090](https://github.com/bobmatnyc/trusty-tools/issues/5090)) leaves no resident daemon to be that arbiter.

**5. Concurrency test quality.** Real, not simulated. `cross_process_holder_serialises_rather_than_locking_out` (`dedup_tests.rs:311`) re-execs the test binary via `current_exe()` with `--exact store::dedup::tests::dedup_lock_holder_child`, and the child takes redb's actual advisory file lock and signals with a `.held` sentinel before the parent claims. I ran it in my own worktree and the nested child harness is visible in the output:

```
running 14 tests
...
running 1 test
test store::dedup::tests::dedup_lock_holder_child ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1576 filtered out; finished in 0.66s
test store::dedup::tests::cross_process_holder_serialises_rather_than_locking_out ... ok
test store::dedup::tests::concurrent_threads_claim_exactly_once ... ok
test store::dedup::tests::held_lock_that_never_releases_reports_contention ... ok
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 1563 filtered out; finished in 2.34s
```

Against pre-fix code it fails for the right reason: the parent store holds the lock for its lifetime, so the child's `Database::create` panics, the child exits 101, and the parent's bounded `try_wait` poll turns that into `lock-holder child exited early with exit status: 101` rather than a hang. The PR body's red run shows exactly that string. `concurrent_threads_claim_exactly_once` also asserts the right invariant — exactly one `Claimed`, never zero — so it catches silent degradation, not just errors.

The gap is on the other side of the boundary: `held_lock_that_never_releases_reports_contention` proves the store *produces* `Contended`, and nothing tests what the runner *does* with it. That untested error arm is where the HIGH lives.

## Notes

- CI not re-run per the review brief. My one scoped run above was to confirm the concurrency tests actually execute rather than no-op; the child-holder test is a `#[test]` that returns early without its env var, so it appears in the pass count twice and contributes nothing in a normal run — worth knowing when reading the 14/1572 counts, not a defect.
- `held_lock_that_never_releases_reports_contention` burns the full 2s budget on every suite run. Acceptable at this size; worth watching if more contention tests land.
- `DedupStore`'s `gate` mutex is per-instance, so the doc at `dedup.rs:257` ("serialises this process's own operations") is true only per handle — two `DedupStore` values in one process contend at the redb layer. `concurrent_threads_claim_exactly_once` proves that works correctly via the retry loop, so this is a wording imprecision, not a bug, and I have not filed it as a finding.
- Not flagged: `OutcomeStore`'s warn-and-continue at `service/webhook.rs:435`. Correctly scoped out — outcomes are telemetry, not a posting gate.
- Changelog fragment, SLOC (`dedup.rs` 263), and Why/What/Test coverage all check out; `bash scripts/check_line_cap.sh` reports 0 violations in my worktree.

bobmatnyc:
All six findings fixed in `500f4a42`.

**HIGH 1 — operation path now fails closed.** The inline match arm became `classify_claim`, and every `DedupError` maps to `ClaimGate::Abort`, not just `Contended`. Carving out one variant would leave the same defect under three other names — an `Open` or `Transaction` error means the same thing operationally: the caller does not know whether this SHA was reviewed, and could not record that it is reviewing it. The runner routes `Abort` through `abort_dry`, so nothing posts and GitHub can redeliver.

`complete()` is the one place fail-closed buys nothing — the comment is already out by then. Upgraded to `error!` and noted that the `InProgress` record still gates redelivery until it ages out. The stale-window blast radius stays out of scope as ruled.

**HIGH 2 — contradiction resolved** in favour of fail-closed, in the module doc and in `DedupError`'s own doc, which also said "log + proceed".

**MEDIUM 3 — recovery guarded.** `recover_incompatible_db` takes an exclusive lock on a `dedup.redb.recovery-lock` sidecar and then **re-checks**: a loser finds the healthy database and adopts it, or finds it locked and retries. Only a process that still sees an unreadable file under the lock rebuilds.

**MEDIUM 4 — off the worker.** `claim`/`complete`/`release` are now async wrappers over `*_blocking`, running on `spawn_blocking`. A join failure surfaces as `DedupError::Transaction`, so it reaches the same fail-closed path. `abort_dry` became async to await `release`.

**MEDIUM 5 — third site declares it.** `DedupNeed` and `open_for` moved from `commands/serve.rs` into `store/dedup_open.rs`. All three `AppState` builders — HTTP serve, `serve --stdio`, and `mcp::build_review_state` — now call `open_dedup_for(&config.log_dir, …)`. No site passes a bare `None`.

**LOW 6** — `OpenAttempt`'s doc now names the two tests that actually reach `Locked`; the recovery branch returns `Locked` instead of a hard `Open`.

## Red-then-green

**HIGH 1** — pre-fix arm restored (`Err(_) => ClaimGate::Proceed`):

```
test pipeline::runner::tests::classify_claim_contended_aborts ... FAILED
test pipeline::runner::tests::classify_claim_open_error_aborts ... FAILED
test result: FAILED. 2 passed; 2 failed

a contended claim must NOT proceed — that posts an ungated comment (#5064)
every store error must abort — the gate did not engage
```

**MEDIUM 3** — recovery lock and re-check removed:

```
test store::dedup::tests::concurrent_recovery_waits_and_adopts_the_winners_database ... FAILED

recovery is not serialised — the second recoverer rebuilt immediately
instead of waiting for the recovery lock (waited 82.218083ms)
```

One thing I could not do honestly: produce a deterministic red for the **loss ordering itself** — a recoverer renaming away a database committed inside its own microsecond-wide window. Reaching it needs a test hook inside the production open path, which this repo bans. The test asserts the guard that closes it (the wait) plus the property it protects (the winner's committed claim survives), and the wait assertion is what goes red. The test's doc comment says exactly this.

**Green, full chain:**

```
cargo fmt --check                                       EXIT=0
cargo clippy -p trusty-review --all-targets -D warnings EXIT=0
cargo test -p trusty-review -- --include-ignored        EXIT=0   1588 passed, 0 failed, 0 ignored
SKIP_UI_BUILD=1 cargo check --workspace --all-targets \
                                                        EXIT=0   (1m 00s)
bash scripts/check_line_cap.sh                          EXIT=0   3763 files, 0 violations
bash scripts/check_changelog_fragment.sh                EXIT=0
```

`store::dedup::tests::` run 10×: pass=10 fail=0. Nothing in this PR is feature-gated beyond `trusty-review`'s own defaults, and no `trusty-common` code is touched.

New tests this round: `classify_claim_contended_aborts`, `classify_claim_open_error_aborts`, `classify_claim_claimed_proceeds`, `classify_claim_skipped_is_duplicate`, `concurrent_recovery_waits_and_adopts_the_winners_database`, `async_claim_runs_off_the_runtime_worker`, `async_claim_complete_release_round_trip`, plus the `open_for_*` set that replaced the deleted `serve_tests.rs` now that the decision lives in the library.

`async_claim_runs_off_the_runtime_worker` is the MEDIUM 4 proof: on a `current_thread` runtime, a 50ms timer must fire while a 300ms lock wait is in progress. If the wait ran on the worker, the timer could not be polled.

Out of scope as ruled: the stranded-`InProgress` blast radius at `runner.rs:246-262`, and #5113.

bobmatnyc:
## Code Critic — round 2, [#5112](https://github.com/bobmatnyc/trusty-tools/pull/5112) (`Closes` [#5064](https://github.com/bobmatnyc/trusty-tools/issues/5064))

## Verdict: WARN

Scoped to the four mechanisms the round-1 fixes introduced. Boundary confirmed: round 1 reviewed `4c48228e`; the fix round is `500f4a42`, `9ce98417`, `d10ed26d`, `514857d3`, and the merge `8577b232` touches zero files under `crates/trusty-review/` (`git diff --stat 514857d3 8577b232 -- crates/trusty-review/` is empty). Settled findings not re-opened.

The recovery sidecar, the `spawn_blocking` wrappers, the `#[non_exhaustive]` change and the `dedup_open.rs` extraction all hold up. One HIGH, in the wiring of the fail-closed abort: round 1's own suggested fix — `return abort_dry(...)` — reaches a helper that unconditionally deletes the dedup record for that SHA, including a `Completed` one this process never owned.

## Findings

| HIGH | `crates/trusty-review/src/pipeline/runner.rs` | 278 (with `runner_helpers.rs` 240–253) | The `ClaimGate::Abort` arm returns `abort_dry(...)`, and `abort_dry` calls `store.release(owner, repo, pr, head_sha)` whenever `head_sha` is non-empty and a store exists. `release_blocking` is `table.remove(key)` (`dedup.rs:363`) — unconditional, with no check that this process ever held the claim. So a *failed* claim deletes whatever record is on that key. If the record was `Completed` (SHA already reviewed and posted) and the claim failed transiently — `Contended` after the 2s budget is the flagship case, and by the time `release` starts a fresh 2s budget the sibling has usually let go — the dedup gate for that SHA is erased and the next trigger re-reviews and posts a second comment. Fail-closed on the claim, fail-open one line later. Before this round `abort_dry` was only ever reached with an owned claim, so the invariant it relies on is new-broken, and no test covers the arm (`classify_claim_*` test the mapping function only). | Do not release a claim you did not acquire. Give `abort_dry` an ownership parameter — `abort_dry(result, config, &input, &deps, ClaimHeld::No)` from the `Abort` arm, `ClaimHeld::Yes` from the other seven call sites — and skip the `release` block when `No`. Add the error-arm test: seed a `Completed` record, drive the runner through a claim failure, assert the record still reads `Skipped` afterwards. | Fix here |
| MEDIUM | `crates/trusty-review/src/store/dedup_tests.rs` | 481–518 | `async_claim_runs_off_the_runtime_worker` has no barrier between spawning the holder thread and calling `claim()`. On a loaded machine the holder can lose the race: either `claim` acquires first and returns in ~0 ms, so the 50 ms timer has not fired and the assertion at :509 fails, or the holder's `Database::create(...).expect("holder takes the lock")` panics and `holder.join().unwrap()` fails. Both are load-induced red on a correct implementation. The two sibling concurrency tests in this same file already synchronise — `cross_process_holder_serialises_rather_than_locking_out` waits on a `.held` sentinel (:362), `concurrent_recovery_waits_and_adopts_the_winners_database` waits on `holding_rx` (:446). The asymmetry is the tell. | Have the holder send on a channel immediately after `Database::create` succeeds, and `recv_timeout` on it before spawning the timer and calling `claim` — the same shape as :424–448. | Fix here |
| MEDIUM | `crates/trusty-review/src/store/dedup.rs` | 26–27 (also `runner.rs:833`, `changelog.d/5064-dedup-flock-collision.md`) | "A dropped review is recoverable (GitHub redelivers)" is the stated justification for aborting, and it is false on the only path where `dedup` is `Some`. `webhook.rs:240` spawns the review and `webhook.rs:323` returns `202 ACCEPTED` before it runs, so GitHub's delivery already succeeded and nothing redelivers. An aborted review is dropped until a human re-requests it; the operator's only signal is the `error!` at `runner.rs:270` plus `last_error`. The abort decision is still right — a duplicate comment is still worse — but the recovery story a maintainer will act on does not exist. | State the actual recovery: the review is dropped and needs a re-request; `error!` + `last_error` + the dry-run log are the signals. Same correction in the changelog fragment, which currently tells operators GitHub will redeliver. | Fix here |
| LOW | `crates/trusty-review/src/pipeline/runner.rs` | 240–242 | `// Store errors are fail-safe: we log and proceed (never block a review).` sits three lines above the code that now does the opposite. Same shape as round 1's contradictory-contracts finding, at a site that one wasn't scoped to. | Replace with the fail-closed rule, or delete — `classify_claim`'s doc already carries it. | Fix here |
| LOW | `crates/trusty-review/src/store/dedup.rs` | 197 | `DedupStore::open` reaches `open_dedup_db_waiting`, which `std::thread::sleep`s up to `LOCK_WAIT_BUDGET`, and it is called from an async fn (`serve.rs:123` → `open_dedup_for` → `open`). Every other blocking method in this file carries a `**Blocking**` marker; `open` does not, so it reads as safe to call from a task. Harmless where it stands — HTTP startup, once, on a path `serve.rs:119` explicitly documents as deadline-free — but the contract is stated and this is the one method that quietly breaks it. | Add the `**Blocking**` line to `open`'s doc, noting it is intended for a startup path only. | Fix here |

## Required Changes

1. `runner.rs:278` must not release a claim it never acquired. Thread ownership into `abort_dry` and add the error-arm test.
2. Fix the barrier in `async_claim_runs_off_the_runtime_worker`.
3. Correct the "GitHub redelivers" premise in `dedup.rs`, `runner.rs`, and the changelog fragment.
4. The two LOW doc fixes are one-liners; take them in the same pass.

## The four mechanisms

**1. `spawn_blocking` wrappers.** Correct. The `Arc<Self>` receiver is right, not merely acceptable: `owned_args` (`dedup.rs:468`) clones the `Arc` and the three `String`s into the closure, which is what satisfies `spawn_blocking`'s `'static` bound, and the whole operation including the `gate: Mutex<()>` acquisition happens inside the closure — no `std::sync::MutexGuard` is ever held across an `.await`. No aliasing issue; `DedupStore` is `Sync` and the `Mutex` serialises this process's callers as before.

One blocking call does remain on an async path — `DedupStore::open`, the LOW above. Everything else is covered: `claim_blocking` / `complete_blocking` / `release_blocking` have no non-test callers outside the async wrappers.

`abort_dry` becoming `async` is correct at all eight call sites (`runner.rs` 278/293/304/365/395/494/540, `runner_mapreduce.rs:124`) — the compiler enforces the `.await`, and none of them was in a `Drop` or other non-async context.

Runtime shutdown mid-`spawn_blocking`: the join error maps to `DedupError::Transaction` (`dedup.rs:495`), which for `claim` reaches `classify_claim` → `Abort` → no post. Fail-closed, correct. For `complete` it reaches the `error!` at `post.rs:270` with the comment already posted, which is the same residual as any other `complete` failure. No partial-write risk either way — redb transactions are atomic and the blocking pool is drained on graceful shutdown.

**2. The recovery-lock sidecar.** Holds up under all four probes.

- *Lock before any destructive step:* yes. `File::create` + `lock()` at `dedup_open.rs:108–120`; the `rename` is at :132, and nothing between them touches the database.
- *Process death holding it:* `File::lock` is `flock`/`LockFileEx` on the descriptor, released by the kernel when the process exits. No stranding. The sidecar file itself is never unlinked, which is the right choice — it means `File::create` and `lock()` always land on the same inode, so there is no unlink/recreate race between two contenders.
- *Two adopters versus a third mid-rename:* not reachable. The rename runs only under the sidecar lock, and the re-check at :124 runs under that same lock. A process that already holds the fresh database open makes that re-check return `DatabaseAlreadyOpen` → `Locked` → retry, so the rename can never fire against a healthy-but-open file. The retry at :149 mapping `DatabaseAlreadyOpen` to `Locked` rather than a hard error closes the last gap here (round 1's LOW).
- *Genuine re-read:* yes. `Database::create(path)` at :124 goes to the filesystem; the pre-lock `cause` is used only for the log message at :141, never as a decision input.

Considered and not flagged: `lock()` at :115 is unbounded while the surrounding wait loop has a 2s budget, so a wedged sidecar holder blocks siblings indefinitely. It runs on a blocking thread now, recovery fires once per redb-2.x→4.x upgrade, and the flock dies with the process. Not worth a timeout.

**3. `classify_claim` mapping every variant to `Abort`.** Right, and the generalisation is the correct call. No variant means "definitely not previously reviewed" — `Open`, `Transaction`, `Serde` and `Contended` all mean the gate did not engage and the prior state is unknown. `Serde` comes closest (a record exists and is corrupt) but corrupt is unknown, not known-absent. With `#[non_exhaustive]` and the `Err(e)` catch-all, future variants default to `Abort`, which is the right default for a gate.

The cost of aborting is higher than the doc claims, though — that is the MEDIUM above. And it compounds the HIGH: the abort deletes the `Completed` record, and the review that eventually repairs the gap is the one that posts the duplicate.

`complete()` is confirmed as the one place fail-closed buys nothing. At `post.rs:250–270` the comment is already on the PR; there is no state to hold and nothing to un-post. `error!` over `warn!` is the right escalation. The residual — the surviving `InProgress` record suppresses re-runs only until `DEDUP_STALE_SECS` (7200s, `constants.rs:100`), after which a later trigger could double-post — is inherent to the design and pre-dates this PR.

**4. `#[non_exhaustive]` and the extraction.** No swallowing risk. Nothing in-tree enumerates `DedupError`'s variants: the only consumers are `classify_claim`'s `Err(e)` catch-all (correct — one policy for all variants) and two `matches!` assertions in tests. In-crate matches still get exhaustiveness checking, so a new variant that *did* need distinct handling would surface at the one site that matters.

The extraction widens the public surface by exactly two items — `DedupNeed` and `open_for` (re-exported as `open_dedup_for`, `store/mod.rs:21`). Everything else stays contained: `OpenAttempt`, `try_open_dedup_db`, `recover_incompatible_db` and `INCOMPATIBLE_SUFFIX` are private; `LOCK_WAIT_BUDGET`, `RECOVERY_LOCK_SUFFIX`, `sibling` and `open_dedup_db_waiting` are `pub(super)`. The widening is deliberate — `mcp/mod.rs:269` is the third construction site that needed it — and matches the 0.12.0 bump. No unintended exposure.

**Test pointers.** `concurrent_recovery_waits_and_adopts_the_winners_database` exists at `dedup_tests.rs:405` and does what both pointers (`dedup_open.rs:20`, `:105`) claim: it asserts the loser waited ≥120 ms for the sidecar (:456) and that the winner's completed claim survives as `Skipped` (:461). The winner takes the sidecar by hand rather than going through `recover_incompatible_db`, so only the loser drives production code — which is the behaviour the pointers name.

Coverage moved rather than shrank: the four deleted `serve_tests.rs` tests map one-to-one onto `open_for_not_needed_touches_nothing`, `open_for_required_opens`, `open_for_required_propagates_failure`, `two_posting_servers_share_a_log_dir`. Nothing was dropped.

## The untestable loss ordering

Adequate, and I would not have asked for more.

The rename-away interleaving — a recoverer renaming a database committed inside its own window — is unreachable once the lock is held, so a test for it would be testing a state the guard makes impossible. The right target is the guard, and the test pins both halves of it: the wait goes red without the sidecar (an unguarded recovery returns in single-digit milliseconds, far under the 120 ms floor) and the adoption assertion goes red if a loser renames the winner's database away. Between them, a regression in either direction fails.

The residual hole is smaller than the one described: no test drives *both* sides through `recover_incompatible_db` itself. That is constructible with the two-child pattern `cross_process_holder_serialises_rather_than_locking_out` already uses, but it would be probabilistic where the current pair is deterministic. Not worth trading. Nothing here needs a production test hook.

bobmatnyc:
All five fixed in `363d258b`.

**HIGH — an abort no longer releases a claim it never held.** `abort_dry` takes a `DedupClaim`; only `Held` releases. The `ClaimGate::Abort` arm passes `NotHeld`, every other abort site passes `Held` — each of those is downstream of a successful claim, or reaches a release that was already a no-op (`dedup: None`, or empty `head_sha` on a local diff).

Worth naming why this got through: round 1's advice was correct when it was written. Every abort reachable at that point held its claim, so reusing `abort_dry` was safe. Routing a not-held path into it is what broke the invariant — the reuse, not the advice.

**MEDIUM 2** — `async_claim_runs_off_the_runtime_worker` now waits on a channel for the holder to signal before starting the timer and the claim, matching both sibling concurrency tests.

**MEDIUM 3** — "GitHub redelivers" corrected in the module doc, `classify_claim`, and the fragment. `webhook.rs` acks 202 before the review runs, so an abort loses the review until a human re-requests it. The decision is unchanged; the reasoning now matches reality.

**LOW 4** — the `// Store errors are fail-safe: we log and proceed` comment above the claim block.
**LOW 5** — `DedupStore::open` now carries the `**Blocking**` marker, with where it is reachable: the HTTP daemon's synchronous startup only, since the stdio path declares `NotNeeded` and never opens the file. Marked rather than moved off the worker — a future hot-path caller has to wrap it, and the marker is what tells them.

## Red-then-green

`abort_dry`'s ownership check removed:

```
test pipeline::runner::tests::held_claim_abort_still_releases ... ok
test pipeline::runner::tests::failed_claim_abort_does_not_delete_another_processes_record ... FAILED

assertion `left == right` failed: aborting on a FAILED claim deleted another
process's completed record — the next trigger for this SHA would post a
duplicate comment (#5064)
  left: Claimed
 right: Skipped
```

`held_claim_abort_still_releases` passes in both directions on purpose: it pins the behaviour the fix must *not* change, so a future "simplification" back to an unconditional release fails the first test while this one stays green.

## Gates

```
cargo fmt --check                                       EXIT=0
cargo clippy -p trusty-review --all-targets -D warnings EXIT=0
cargo test -p trusty-review -- --include-ignored        EXIT=0   1590 passed, 0 failed, 0 ignored
SKIP_UI_BUILD=1 cargo check --workspace --all-targets --offline \
                                                        EXIT=0
scripts/check_test_pointers.sh                          EXIT=0   22384 citations, 0 dangling
scripts/check_semver.sh (local)                         EXIT=0   1 skipped, already-major
scripts/check_line_cap.sh                               EXIT=0   3767 files, 0 violations
scripts/check_changelog_fragment.sh                     EXIT=0
scripts/check_sld.sh                                    EXIT=0   0 errors, 0 warnings
```

`store::dedup::tests::` 10×: pass=10 fail=0. Branch is level with main; `Cargo.lock` is still the single `trusty-review` version line.

Issue #3631, opened 2026-07-22
EmbedderSupervisor crash-storm counter never escalates when respawns succeed — swap-back-to-ort can stall forever under a real crash-loop


bobmatnyc:
## Summary

`EmbedderSupervisor`'s crash-storm give-up counter (`consecutive_failures`,
`crates/trusty-common/src/embedder_client/supervisor.rs:582`) is **structurally
incapable of ever exceeding 1**, for any `max_restarts`, whenever the sidecar
is killed/crashes and each individual respawn attempt's startup probe itself
succeeds — which is the normal, expected shape of "kill it, it comes back
healthy" (including a plain `SIGKILL`). This means `should_give_up`
(supervisor.rs:517), `terminated_signal` (supervisor.rs:778,
`is_confirmed_terminated()` at supervisor.rs:202), and therefore the
python→ort **swap-back watchdog's only trigger predicate**
(`trusty-search/src/commands/start/swap_back_watchdog.rs`, "The predicate is
now simply: `active.kind == Python && python_teardown.is_confirmed_terminated()`")
can **never fire** for this failure shape — not "slowly," but **never**,
regardless of how long you wait or how many times the sidecar is killed.

Discovered while investigating PR #3626 (epic #3524 slice 7)'s two soak
tests timing out at their 120s deadline. The identical symptom already
exists in an **already-merged, `#[ignore]`d-by-default** real-hardware test —
`swap_back_real_hardware_kills_sidecar_past_max_restarts` in
`crates/trusty-search/src/commands/start/tests.rs:1980` (90s deadline, same
repeated-SIGKILL mechanism) — which is why it has never been caught: it never
runs in CI.

## Root cause, with evidence

`supervision_loop` (supervisor.rs:571-878) tracks two independent escalation
counters:

- `consecutive_failures` — incremented on every restart trigger
  (supervisor.rs:749), **unconditionally reset to 0 on every successful
  respawn** (supervisor.rs:867, inside the `Ok((new_child, new_client)) =>`
  arm of the respawn `match`).
- `consecutive_wedge_restarts` — incremented only for `RestartTrigger::Unhealthy`
  (a live-but-stuck sidecar detected via `unhealthy_signal`), and **deliberately
  NOT reset by an ordinary respawn success** — only by `wedge_reset_secs`
  (default 300s) of sustained health (`wedge_counter_should_reset`,
  supervisor.rs:496). This "don't reset on mere respawn success" design was
[...]

PR #5119, opened 2026-08-07
fix(trusty-memory): stop the bm25 supervisor reusing a daemon that stopped serving (#5085)

### LINKED ISSUE #5085: bm25_supervisor: dead child silently reused instead of respawned (fast-path/eviction race)
## Defect

`crates/trusty-memory/tests/bm25_supervisor_concurrency.rs:229`, test `a_dead_child_is_evicted_and_respawned`, failed in CI run [31179690320](https://github.com/bobmatnyc/trusty-tools/actions/runs/31179690320) (job `92888380574`):

```
assertion `left == right` failed: a dead child must be replaced by a NEW process, not silently reused
  left: 1
 right: 2
```

`spawned_count()` was 1 where 2 was expected — a SIGKILLed child was silently reused instead of respawned. Fail-open shape: the supervisor believes it has a live child, the child is dead, nothing reports it.

Suspected (not established) mechanism: `ensure_running`'s fast path wins a race against eviction. `lookup_live` → `try_wait()` may report a killed child alive because the kernel hasn't reaped it yet, so eviction/respawn never runs. Investigate `bm25_supervisor.rs`'s `ensure_running` / `lookup_live` / `kill_daemon_for` interaction — do not assert this as confirmed.

## Evidence

- Not caused by [#5074](https://github.com/bobmatnyc/trusty-tools/pull/5074), proven at byte level. CI tested `refs/pull/5074/merge` (`b3d2b4064620a2202b5b93c8ceeddef3be7085bb`), not the raw head. `git diff cbe0b083d b3d2b4064 -- crates/trusty-common/src/bm25_client.rs crates/trusty-common/src/sys_metrics.rs crates/trusty-memory/src/bm25_supervisor.rs` exits 0 — byte-identical to main. The code that failed is exactly what's on main today.
- Not reproducible locally: 62 runs, 0 failures. 15x single test on `origin/main` (`e62cc7dcf`), 15x at the merge SHA, 8x whole-file serial, 10x whole-file default-parallel, 12x whole-file parallel under 28 `yes` processes saturating 16 cores, plus 2 full `trusty-memory` suite runs (601+ tests each). `pgrep -fl bm25` empty before, during, and after every batch — no leaked daemons.
- Likely why local repro failed: GitHub runners are typically 2-4 cores vs. the 16-core test machine. CI's `cargo test --workspace` also runs `bm25_supervisor_e2e.rs` and `bm25_backfill_e2e.rs` as concurrent daemon-spawning processes — that contention wasn't replicated locally.
- First failure since #5048 introduced the test. No prior issue references this test name.

## Resolution

Two separable things:
1. The race itself — if `try_wait()` on a killed-but-unreaped child can report it alive, the fast path needs to distinguish "alive" from "not yet reaped."
2. Reproducibility — the test can't currently be exercised under the failing conditions. Injecting the unreaped-child state deterministically (rather than relying on timing) would let a fix be proven.

Do not `#[ignore]`, cfg-gate, retry, or weaken the assertion — it's the only thing that caught this. Fix the supervisor so the assertion passes on its own.

bobmatnyc:
## Primary outcome

`Closes #5085` — the BM25 supervisor no longer hands a caller a daemon that has stopped serving.

## Root cause

`Bm25Supervisor::ensure_running`'s fast path decided liveness from `tokio::process::Child::try_wait()` alone. That call answers **"has this child been reaped"**, which is not **"is it serving"**. A SIGKILLed daemon closes its listening socket during `do_exit` and only becomes reapable once the kernel finishes tearing it down; for that window `try_wait()` returns `Ok(None)` for a process that is already refusing connections. `lookup_live` then returned its socket path and `ensure_running` skipped the respawn — the fail-open the issue describes, with nothing reporting a problem.

Measured directly (40 SIGKILL rounds against a real daemon, idle 16-core host):

```
ROUNDS=40 socket_dead_first=35 reap_first=5 same=0 max_gap=32.625µs
```

In 35 of 40 rounds the socket stopped accepting **before** `try_wait()` reported the child exited. The window is microseconds when the host is idle — which is why 62 local runs found nothing — and long enough to lose on a 2-core runner executing `cargo test --workspace`.

## What changed

`crates/trusty-memory/src/bm25_supervisor.rs`:

- `lookup_live` now requires **both** that the child is un-reaped **and** that its socket is not definitively unserved. The probe runs with the map lock released so a slow connect never blocks another palace's lookup; eviction re-acquires and fires only if the map still holds the same child, so a concurrently-spawned replacement is never evicted on a stale verdict.
- New three-state `SocketVerdict` replaces the bare bool for the eviction decision. Only ENOENT and ECONNREFUSED — where the kernel proves no listener is bound — count as `NotServing`. A timeout or a saturated accept backlog is `Inconclusive` and leaves the child alone, so a load spike cannot become a respawn storm (relevant to #3631). `probe_socket` keeps its exact previous semantics for the spawn and adoption paths.

**Out of scope / deliberately unchanged:** the spawn gate, socket adoption, the LRU cap, RSS enforcement, `kill_on_drop`, and the SIGTERM→SIGKILL patience window are all untouched. No eviction-time `unlink` was added — `trusty-bm25-daemon` already calls `cleanup_stale_socket` before `bind`, so a respawn over a stale file cannot EADDRINUSE.

## Risk / blast radius

One crate, one function's decision logic. The added cost is one UDS connect per `ensure_running` on the already-supervised fast path (~30–80µs). A daemon that is genuinely serving is never evicted; the `Inconclusive` bucket exists specifically so that a busy one is not either.

## Test evidence — Rung 5 (process lifecycle + concurrency)

**Red before green — deterministic seam.** `a_live_child_that_stopped_serving_is_evicted_and_respawned` pins the *same* `lookup_live` state the race reaches transiently — child un-reaped, socket unserved — by unlinking the socket while the daemon keeps running. The child is then alive by `try_wait()` forever, so there is no timing dependence at all. That is also a real production state: a `$TMPDIR` reaper leaves a daemon alive and permanently unreachable.

Against the pre-fix supervisor (`origin/main`'s `bm25_supervisor.rs` swapped back in, new test retained):

```
thread 'a_live_child_that_stopped_serving_is_evicted_and_respawned' panicked at
crates/trusty-memory/tests/bm25_supervisor_concurrency.rs:286:5:
assertion `left == right` failed: a child that is no longer serving its socket must be
replaced — try_wait() reporting it un-reaped is not evidence that it is serving
  left: 1
 right: 2

PRE-FIX new test: 20 runs -> passed=0 FAILED=20
```

20/20 red pre-fix; the failure signature is identical to the CI one (`left: 1 / right: 2`). Post-fix it passes.

**The original flaky test, reproduced and fixed.** Running `a_dead_child_is_evicted_and_respawned` 16-way concurrently, each instance with a private `TMPDIR` so socket paths are disjoint and cross-instance adoption cannot confound the result:

```
PRE-FIX,  160 runs -> 13 failed  (8.1%)   [13x "a dead child must be replaced by a NEW process"]
POST-FIX, 160 runs ->  0 failed  (identical experiment)
```

**Gates:**

```
cargo fmt --check                                          -> EXIT=0
cargo check --workspace --all-targets                       -> EXIT=0
  (SKIP_UI_BUILD=1, --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui, per ci.yml:23)
cargo clippy -p trusty-memory --all-targets -- -D warnings  -> EXIT=0
cargo test -p trusty-memory                                 -> EXIT=0; 606 passed, 0 failed, 4 ignored (lib)
                                                               + all integration targets green, 0 failures
cargo test -p trusty-memory --test bm25_supervisor_concurrency, 50x
                                                            -> passed=50 failed=0 (200 test executions)
bash scripts/check_line_cap.sh          -> 3762 files, 0 violations — OK
bash scripts/check_changelog_fragment.sh -> OK trusty-memory: fragment present and valid
bash scripts/check_sld.sh                -> 0 errors, 0 warnings
```

New coverage: 3 unit tests pinning the `SocketVerdict` classification (missing socket, stale socket file left by a crash, serving socket) + the deterministic integration test above.

## Baseline failures

None. Every gate above is green on this branch.

## Documentation / changelog

`crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md` (Fixed). `lookup_live` and `SocketVerdict` carry full Why/What/Test docs; the change site is marked `// #5085:`.

## Review-finding disposition

Nothing deferred.

### DIFF
--- a/crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md
+++ b/crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md
@@ -1,0 +1,10 @@
+Fixed
+
+- BM25 supervisor: a daemon that had stopped serving its socket could be handed
+  back to callers instead of respawned. `ensure_running`'s liveness check
+  trusted `try_wait()`, which reports whether a child has been reaped rather
+  than whether it is serving, so a killed daemon read as alive for the window
+  between closing its listener and becoming reapable. Liveness is now backed by
+  the socket the caller is about to use, and eviction fires only when the kernel
+  proves nothing is listening (ENOENT/ECONNREFUSED) so a busy daemon is never
+  mistaken for a dead one.
--- a/crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md
+++ b/crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md
@@ -7,4 +7,6 @@ Fixed
   between closing its listener and becoming reapable. Liveness is now backed by
   the socket the caller is about to use, and eviction fires only when the kernel
   proves nothing is listening (ENOENT/ECONNREFUSED) so a busy daemon is never
-  mistaken for a dead one.
+  mistaken for a dead one. A daemon evicted this way is still running, so it is
+  now given a SIGTERM and time to flush its acked writes before the replacement
+  takes over its socket, rather than being SIGKILLed on the spot.
--- a/crates/trusty-memory/src/bm25_supervisor.rs
+++ b/crates/trusty-memory/src/bm25_supervisor.rs
@@ -456,46 +456,88 @@ impl Bm25Supervisor {
     /// and once under the spawn gate — and the two must agree exactly,
     /// including the dead-child eviction. Two copies would be two chances to
     /// drift.
-    /// What: returns `Some(socket_path)` when the stored child is still
-    /// running, and stamps it as most-recently-used so the LRU victim choice
-    /// reflects real traffic. A child that has exited (or whose status cannot
-    /// be read) is removed from the map and `None` is returned, so the caller
-    /// falls through to a fresh spawn. Removing a corpse does NOT count as a
+    ///
+    /// Liveness is decided by TWO questions, not one (#5085). `try_wait()`
+    /// answers "has this child been reaped", which is not the same as "is it
+    /// serving": a SIGKILLed daemon closes its listening socket during
+    /// `do_exit` and only becomes reapable once the kernel finishes tearing it
+    /// down, so for that window `try_wait()` reports `Ok(None)` for a process
+    /// that is already refusing connections. Trusting it alone made this the
+    /// fast path's fail-open — the caller was handed a socket path nothing was
+    /// listening on and nothing reported a problem. The socket the caller is
+    /// about to use is the authority, so we ask it too.
+    ///
+    /// What: returns `Some(socket_path)` when the stored child is both
+    /// un-reaped AND its socket is not definitively unserved, and stamps it as
+    /// most-recently-used so the LRU victim choice reflects real traffic. A
+    /// child that has exited, whose status cannot be read, or whose socket
+    /// answers [`SocketVerdict::NotServing`] is removed from the map and `None`
+    /// is returned, so the caller falls through to a fresh spawn. An
+    /// [`SocketVerdict::Inconclusive`] probe keeps the child: a saturated
+    /// accept backlog is a busy daemon, not a dead one, and reaping on it would
+    /// turn load into a respawn storm. Removing a corpse does NOT count as a
     /// reap — nothing was reclaimed.
-    /// Test: `a_dead_child_is_evicted_and_respawned` (e2e).
+    ///
+    /// The probe runs with the map lock released, so a slow connect never
+    /// blocks another palace's lookup. Eviction then re-acquires and fires only
+    /// if the map still holds the same child, so a replacement spawned
+    /// concurrently is never evicted by this call's stale verdict.
+    /// Test: `a_dead_child_is_evicted_and_respawned` and
+    /// `a_live_child_that_stopped_serving_is_evicted_and_respawned`.
     async fn lookup_live(&self, palace: &str) -> Option<PathBuf> {
-        let stamp = self.tick();
-        let mut guard = self.children.lock().await;
-        let entry = guard.get_mut(palace)?;
-        match entry.child.try_wait() {
-            Ok(None) => {
-                entry.last_used = stamp;
-                let path = entry.socket_path.clone();
-                tracing::trace!(
-                    palace = %palace,
-                    socket = %path.display(),
-                    "bm25 supervisor: child already running"
-                );
-                Some(path)
+        // Phase 1 — under the map lock: reap check, LRU stamp, snapshot.
+        let (pid, path) = {
+            let stamp = self.tick();
+            let mut guard = self.children.lock().await;
+            let entry = guard.get_mut(palace)?;
+            match entry.child.try_wait() {
+                Ok(None) => {
+                    entry.last_used = stamp;
+                    (entry.child.id(), entry.socket_path.clone())
+                }
+                Ok(Some(status)) => {
+                    tracing::warn!(
+                        palace = %palace,
+                        ?status,
+                        "bm25 daemon exited unexpectedly — attempting one restart"
+                    );
+                    guard.remove(palace);
+                    return None;
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        palace = %palace,
+                        "bm25 supervisor: try_wait failed: {e:#} — evicting and retrying"
+                    );
+                    guard.remove(palace);
+                    return None;
+                }
             }
-            Ok(Some(status)) => {
-                tracing::warn!(
-                    palace = %palace,
-                    ?status,
-                    "bm25 daemon exited unexpectedly — attempting one restart"
-                );
-                guard.remove(palace);
-                None
-            }
-            Err(e) => {
-                tracing::warn!(
-                    palace = %palace,
-                    "bm25 supervisor: try_wait failed: {e:#} — evicting and retrying"
-                );
-                guard.remove(palace);
-                None
-            }
+        };
+
+        // Phase 2 — lock released. #5085: an un-reaped child is not the same
+        // claim as a serving one; ask the socket the caller will actually use.
+        if probe_socket_verdict(&path).await != SocketVerdict::NotServing {
+            tracing::trace!(
+                palace = %palace,
+                socket = %path.display(),
+                "bm25 supervisor: child already running"
+            );
+            return Some(path);
         }
+
+        // Phase 3 — evict, but only if the map still holds THIS child.
+        tracing::warn!(
+            palace = %palace,
+            socket = %path.display(),
+            pid = ?pid,
+            "bm25 daemon is not serving its socket — evicting and respawning"
+        );
+        let mut guard = self.children.lock().await;
+        if guard.get(palace).map(|h| h.child.id()) == Some(pid) {
+            guard.remove(palace);
+        }
+        None
     }
 
     /// Bring the live-daemon population within both limits, leaving room for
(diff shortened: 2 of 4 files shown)


bobmatnyc:
## `code-critic` verdict: WARN

Adversarial review of [#5119](https://github.com/bobmatnyc/trusty-tools/pull/5119) (`Closes` [#5085](https://github.com/bobmatnyc/trusty-tools/issues/5085)), rung 5. Reviewed against the spec at head `cf6777cb` in an independent worktree.

**The root-cause claim holds and the fix is correct.** `try_wait()` answers "reaped", not "serving", and the socket is the right authority. The three-phase lock discipline is sound — I could not construct an ABA or a double-evict against it (details under Q1). What earns the WARN is one thing the fix introduces: the new eviction arm is the first path in this supervisor that removes a **live** child, and it drops it on the floor instead of routing it through the SIGTERM→flush contract every other termination path honours.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-memory/src/bm25_supervisor.rs` | 536–539 | Phase-3 eviction removes a **live** child and lets `kill_on_drop` SIGKILL it — no SIGTERM, no flush, no socket unlink. Acked-but-unflushed writes are lost, silently. | Move the handle out of the map and hand it to a detached `terminate_child` + `remove_socket_file` task. | Fix here |
| MEDIUM | `crates/trusty-memory/src/bm25_supervisor.rs` | 798–801, 808 | The `SocketVerdict` contract states ECONNREFUSED "proves no listener is bound" and that a saturated backlog surfaces as `EAGAIN`/`Inconclusive`. False on macOS/BSD — measured below. | State the platform split in the doc; do not carry the false invariant into `trusty-common`. | Fix here |
| MEDIUM | `crates/trusty-memory/tests/bm25_supervisor_concurrency.rs` | 318 | No test drives the **ECONNREFUSED** variant through `lookup_live` — both integration tests unlink the socket (ENOENT), but the real #5085 race leaves the file in place and refusing. | `remove_file`, then bind-and-drop a `UnixListener` at the same path: the file exists and refuses while the child stays alive. ~3 lines, still deterministic. | Fix here |
| LOW | `crates/trusty-memory/src/bm25_supervisor.rs` | 537 | The same-child key is `Option<u32>`. A `None` pid would match any entry whose child had been reaped. | Bind `Some(pid)` in phase 1; compare `u32` to `u32`. | Fix here |

### HIGH — the evicted child is SIGKILLed without a flush

```rust
let mut guard = self.children.lock().await;
if guard.get(palace).map(|h| h.child.id()) == Some(pid) {
    guard.remove(palace);      // ChildHandle dropped here → kill_on_drop(true) → SIGKILL
}
```

`enforce_limits` (`bm25_supervisor.rs:594-609`) and `shutdown` (`bm25_supervisor.rs:635-648`) both call `terminate_child` + `remove_socket_file`. The pre-existing `Ok(Some(status))` arm skips both, correctly — that child is already a corpse. This new arm is the first one that evicts a child that is **alive by construction**, and it inherited the corpse arm's handling.

The write path acks before it flushes: `apply_write_op` sends the oneshot reply on the in-memory apply, and `index.flush()` runs once at the end of the 50 ms window (`batch_queue.rs:449-458`). SIGKILL inside that window drops acked documents. `trusty-bm25-daemon/src/lib.rs:269-279` exists specifically for this and names this supervisor as the reason:

> Harmless-looking while SIGTERM only arrived at daemon shutdown; not harmless now that the trusty-memory supervisor reaps daemons routinely to stay within its process cap, which turns a shutdown-only edge case into an ordinary one.

Observed directly in this PR's own new test — two daemons started, one SIGTERM, one flush:

```
$ cargo test -p trusty-memory --test bm25_supervisor_concurrency \
    a_live_child_that_stopped_serving -- --exact ...
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.58s

$ grep -c "trusty-bm25-daemon ready"        -> 2
$ grep -c "received SIGTERM"                -> 1
$ grep -c "flushed BM25 snapshot before exit" -> 1
```

Nothing recovers it. The supervisor cannot reach `mark_dirty`, so the repair sweep never sees this palace; startup backfill is the only backstop, and `bm25_repair.rs:6-10` already rules on that standard: *"'Recoverable across a restart' is not recoverability; it is a promise the process makes and does not keep."*

Pre-fix, the alive-but-unreachable child was never evicted, so it was never SIGKILLed either. This is a loss the fix introduces. The magnitude is small (≤ one 50 ms window per eviction) and the trigger is narrow, but #5085's whole finding is that this trigger fires more often than anyone thought.

Note the fix must be **detached** — `terminate_child` waits up to `SIGTERM_PATIENCE` (5 s), and `lookup_live` is on `ensure_running`'s fast path.

### MEDIUM — `Inconclusive` does not cover a saturated backlog on macOS

The doc asserts:

> `NotServing` is reserved for errors that prove no listener is bound to the path — ENOENT and ECONNREFUSED. A timeout or any other errno (EAGAIN from a saturated backlog, EPERM) is `Inconclusive`.

Measured on this host (darwin 25.5.0) — `listen(1)`, never accept, six non-blocking connects:

```
0 OK
1 ERR 61 ECONNREFUSED
2 ERR 61 ECONNREFUSED
3 ERR 61 ECONNREFUSED
4 ERR 61 ECONNREFUSED
5 ERR 61 ECONNREFUSED
```

BSD `unp_connect` returns ECONNREFUSED when `so_qlen >= so_qlimit`. Linux returns EAGAIN and behaves as documented. So on macOS a saturated backlog reads as `NotServing`, and the healthy busy daemon is evicted and SIGKILLed — the exact respawn storm the third state exists to prevent, on the platform this project is developed and run on.

I am **not** claiming this is reachable today: tokio's `UnixListener::bind` uses a backlog of 1024 and `run_accept_loop` spawns a task per connection, so 1024 unaccepted connects is not a state this daemon gets into. Flagged because the sentence is the design contract, and #5089 promotes that contract into `trusty-common` where the next consumer may have a much smaller backlog or a slower accept loop.

### MEDIUM — the ECONNREFUSED path is never composed end to end

`a_live_child_that_stopped_serving_is_evicted_and_respawned` unlinks the socket → ENOENT. `a_dead_child_is_evicted_and_respawned` also removes the socket file (`kill_daemon_for`, per its own doc) → ENOENT. The classifier's ECONNREFUSED arm is unit-tested in isolation (`bm25_supervisor_tests.rs:252`), but nothing proves the classifier and `lookup_live` compose on the shape the actual race produces: a daemon closing its listener with the socket file still on disk.

Not a "fails for the wrong reason" defect — see Q4 — but the union of the two tests leaves the production shape uncovered, and it is three deterministic lines to close.

---

## Answers to the five judgement calls

**Q1 — is the same-child identity check sufficient? Any ABA?** Yes, sufficient. Three cases:

- *A replacement was spawned at the same key.* It carries a different OS pid, so `Some(new_pid) != Some(pid)` and the replacement survives. A false match needs the old pid reaped **and** recycled **and** re-spawned inside the probe window (≤200 ms, typically ~50 µs) — thousands of process creations in that span. Below the confidence bar for a finding; the blast radius if it ever happened is one extra respawn on the next call, not a lost daemon.
- *Two concurrent probes both decide to evict.* The first removes; the second's `guard.get(palace)` returns `None`, `map` gives `None != Some(pid)`, no removal. Idempotent. Both then fall through to the spawn gate, which serialises them — the second's re-check finds the new child and probes it. No double spawn.
- *The entry was removed by `enforce_limits` in between.* Same `None` path. Correct.

The one soft spot is the `Option<u32>` key (LOW above), which is unreachable only because phase 1's `Ok(None)` guarantees `id()` is `Some`. That is a non-local invariant holding up a correctness guard.

**Q2 — can `Inconclusive` become fail-open? Can a dead daemon be permanently `Inconclusive`?** No. A genuinely dead process becomes reapable within milliseconds, and phase 1's `try_wait()` is still the backstop — it evicts on `Ok(Some(status))` before phase 2 ever runs. Phase 2 only adds an *earlier* trigger; it never removes the reap-based one. Permanent `Inconclusive` can therefore only hold a daemon that is **alive and unreachable**: EPERM on the socket, the path replaced by a non-socket, or a bound-but-wedged listener that times out every probe (Linux only — on macOS that reads ECONNREFUSED, see above).

For that residual case, nothing breaks the tie on purpose. The LRU cap does it incidentally once three other palaces are touched (default `max_live = 3`), and the RSS ceiling only if it leaks. Otherwise the caller gets a path its own connect fails on — an error it surfaces, not the silent success #5085 was about. Pre-fix behaviour in that state was identical (no socket-based eviction at all), so this is a limitation of the fix, not a defect it introduces. Not flagged.

**Q3 — does the probe add a failure surface?** Bounded and cheap. One `UnixStream::connect` per `ensure_running` under a hard `timeout(200ms)`, run with the map lock **released**, so a slow connect never blocks another palace. Worst case is +200 ms latency per call while a daemon is bound-but-wedged, which is strictly better than the silent wrong answer it replaces. The probe connection is accepted then immediately EOFs; `server.rs:93-96` returns `Ok(())` on `n == 0`, so no log spam and no error counter — I checked, because a warn-per-probe would have been a real cost on a per-query path. The one structural note: a probe under the spawn gate (`ensure_running`'s re-check) holds that gate for up to 200 ms, but `wait_for_socket`'s 3 s already dominates there.

**Q4 — does the deterministic test prove what it claims?** Yes, for the right reason. Pre-fix, `lookup_live`'s `Ok(None)` arm returns `Some(path)` unconditionally, `ensure_running` returns at its fast path, `spawned_count()` stays 1, and the assertion reads `left: 1 / right: 2` — mechanically forced by the diff, not a coincidence of a different failure. Post-fix I ran it and confirmed 2 spawns with the evict path firing.

The nuance is which *variant* of the state it pins. The test reaches `lookup_live` phase 2 with `try_wait() == Ok(None)` and the socket unserved — the same branch — but via ENOENT, where the real race produces ECONNREFUSED. Same branch, different errno arm of the classifier. That is the MEDIUM above, not a "wrong reason" failure: the test is a valid regression test, its coverage just stops one composition short.

**Q5 — [#3631](https://github.com/bobmatnyc/trusty-tools/issues/3631) interaction.** Structurally unchanged; marginally worse in rate. This PR adds a new respawn trigger with no attempt counter, no backoff and no give-up — but the pre-existing `try_wait()` trigger had none of those either, so a crash-looping daemon was already respawned unboundedly. What changes is that the trigger now fires earlier and on a wider set of states, including states where the daemon is not crashing at all (the ENOENT reaper case, and the macOS backlog case above). So a crash-loop respawns faster than it used to, and the give-up that #3631 is about still does not exist here. Out of scope for this PR; noting it because #5089 will inherit it.

---

## Promotion readiness for [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) step 2 (`UdsServiceSupervisor` in `trusty-common`)

**The liveness contract itself promotes cleanly.** `SocketVerdict`, `probe_socket_verdict`, and the phase-1/2/3 shape of `lookup_live` contain nothing BM25-specific — they are pure UDS mechanics over a `tokio::process::Child`. The author is right that the promotion must carry the socket-backed contract rather than `try_wait()`, and the code is in a shape that allows it. Both are private today (`enum SocketVerdict`, `async fn probe_socket_verdict`); making them `pub` is the whole mechanical change.

**Three things to settle before the promotion, not after:**

1. **The two findings above get inherited by every consumer.** The SIGKILL-without-flush arm is survivable here because the BM25 snapshot is a derived index. For a promoted service whose on-disk state is authoritative, the same arm is unqualified data loss. Fix it in this PR and the promoted version starts correct.
2. **The ECONNREFUSED invariant will be load-bearing at that layer.** "The kernel proved no listener is bound" is the sentence a future consumer will design against, and it is false on darwin. A consumer with a small backlog or a slow accept loop turns it into the respawn storm the enum was written to prevent.
3. **`Inconclusive` needs an escalation policy at the generic layer.** Trusty-memory gets away without one because the LRU cap incidentally clears a stuck entry at `max_live = 3`. A generic supervisor supervising one service has no such backstop, and an alive-but-permanently-unreachable child stays in the map forever. Consecutive-`Inconclusive` counting, or an explicit "this is the caller's problem" clause in the contract.

**What is trusty-memory-specific and must become parameters, not constants:** `socket_path_for_palace` / `locate_bm25_daemon_binary` (path + binary resolution), the `--palace <name> --data-dir <dir>` argv in `spawn_child`, the `TRUSTY_BM25_*` env-var names, and — the one with a real trap — the two timeouts. `SIGTERM_PATIENCE = 5s` is justified as *strictly greater than the daemon's own 2 s `SHUTDOWN_FLUSH_TIMEOUT`*, and `SPAWN_PROBE_TIMEOUT = 3s` is justified by *"BM25 has no model-loading step"* while the same comment cites 30 s for the embedder. Promoted as constants, both invariants break silently for the first service with a longer flush or a model to load. They need to be per-service inputs with the relationship (`patience > flush budget`) asserted at construction.

---

## Gates

I re-ran only what the review needed; I did not re-run the 50× concurrency loop.

```
cargo test -p trusty-memory --test bm25_supervisor_concurrency \
  a_live_child_that_stopped_serving -- --exact ...
  -> EXIT=0; test result: ok. 1 passed; 0 failed; 0 ignored; 3 filtered out; 0.58s

  -> EXIT=0; measured 3762 tracked .rs file(s) (floor 500); 7 allowlisted, 0 violations — OK.
  -> .line-cap-allowlist.tsv untouched by this PR; bm25_supervisor.rs not allowlisted.
```

Changelog fragment present (`crates/trusty-memory/changelog.d/5085-supervisor-eviction-race.md`, category `Fixed`). Docs carry the Why/What/Test pattern; the change site is marked `// #5085:`. No `unwrap()` added to library code.

**WARN, not BLOCK** — nothing here is a correctness hazard in the fix's main claim, and the code is better than what it replaces. The four findings are all `Fix here` and all small; the HIGH is worth landing before merge because #5089 will copy it.

PR #5120, opened 2026-08-07
fix(trusty-analyze): remove the unused neural embedder that stalled boot (#5067)

### LINKED ISSUE #5067: trusty-analyze: remove unused neural embedder (fastembed/ort) — untimed hf-hub call stalls daemon boot 30+ min
**Defect:** `trusty-analyze` links `fastembed`/`ort` directly for `method=neural` concept clustering (`GET /indexes/:id/clusters`, MCP `cluster_concepts`), but nothing requests it. The only cross-crate consumer, `trusty-console`'s `server/mod.rs`, calls `cluster_concepts` without a `method` param and gets the `bow` default. `bow` is the default in the trait derive, the HTTP handler, the MCP schema, and the embedded UI's JS. The only route to neural is a human toggling a dropdown in the embedded UI.

Despite that, `NeuralEmbedder::new()` runs unconditionally at every daemon boot in `commands/run.rs` whenever an ORT feature is compiled in — true for every shipped release asset per `.github/workflows/release.yml` and `trusty-installer`'s `ORT_CRATES` routing. It makes an untimed hf-hub network call even when the model is already cached locally.

**Evidence — this is a live hang, not a theoretical cost.** From `~/.trusty-analyze/logs/stderr.log` on this machine:

```
2026-08-04T12:24:57.794694Z  INFO Using token file found ".../huggingface/token"
2026-08-04T12:56:43.533041Z  INFO neural embedder loaded from .../.fastembed_cache
```

31m46s between those lines. Four other observed boots (07-12, 07-13 ×2, 07-19, 07-21) completed the same step in under a second, and the gap from "loaded" to "listening on :7879" is ~1ms — so the ONNX construction itself is negligible. The stall is the network call, and it gates the whole daemon's startup for a feature no caller uses.

This corrects a recorded assumption that trusty-analyze's ~32-minute boot was normal ONNX/fastembed load time. It is neither normal nor the ONNX load.

Also: trusty-analyze has zero references to `trusty_common::embedder_client` / `trusty-embedderd`, which #110 created to solve exactly this warm-model problem and which `trusty-search` already consumes via `UdsEmbedderClient`. Under the common-entry-point rule this is a second independent implementation — and it is the copy carrying the defect the shared one was built to avoid.

**Resolution:** remove the neural path. Delete `embedder/neural.rs`; the optional `fastembed`/`ort`/`tokenizers` deps and the `bundled-ort`/`load-dynamic`/`cuda` features from `crates/trusty-analyze/Cargo.toml`; the `method=neural` branches in `service/handlers/graph.rs` and `mcp/descriptors.rs`; the neural option in `ui/src/lib/views/Clusters.svelte`; and the unconditional `NeuralEmbedder::new()` in `commands/run.rs`. `bow` becomes the sole embedder. This also drops `trusty-analyze` from `trusty-installer`'s `ORT_CRATES` glibc-asset special-casing, leaving only `trusty-search`.

Removal is preferred over routing through `trusty-embedderd` because there is no live consumer to preserve.

bobmatnyc:
## Defect

`run_serve` constructed `NeuralEmbedder::new()` at every daemon boot whenever an ORT backend feature was compiled in — true for every shipped release asset, because `bundled-ort` was in the default feature set. That construction made an **untimed** hf-hub request. The daemon's bind waited on it.

Nothing selected the backend. `trusty-console`'s `server/mod.rs` calls `cluster_concepts` with no `method`; `bow` is the default in the trait derive, the HTTP handler, the MCP schema, and the UI's JS. The only route to neural was a dropdown a human had to touch.

**Now a hard blocker.** Under ADR-0034 (#5090) `trusty-console` spawns the target service on demand *per webhook delivery*. A boot stall is bad once per restart; it is fatal per delivery. #5089 step 4 was blocked on this.

## Measurement

Both legs use the same harness (`serve` against the live `trusty-search` on 7878; `HF_ENDPOINT` pointed at a stub that counts requests and delays each one by `HF_DELAY`; `TRUSTY_FASTEMBED_CACHE`/`HF_HOME`/`FASTEMBED_CACHE_*` pointed at an empty dir; wall clock from exec until 127.0.0.1:PORT accepts).

| Leg | HF requests at boot | Time to listening |
|---|---|---|
| before, `HF_DELAY=60` | 1 (`GET /Qdrant/all-MiniLM-L6-v2-onnx/resolve/main/model.onnx` at +0.083s) | **60.170s** |
| before, `HF_DELAY=120` | 1 (same request at +0.108s) | **120.131s** |
| after, `HF_DELAY=60` | **0** | **0.065s / 0.134s / 0.230s** (3 runs) |

Boot time tracked the injected latency 1:1, which is the point: the wait is unbounded, so 31m46s is not an outlier. The pre-fix log shape reproduced exactly —

```
2026-08-07T16:04:26.110454Z  INFO Using token file found "/Users/masa/.cache/huggingface/token"
2026-08-07T16:05:26.112311Z  WARN neural embedder failed to load from /tmp/ta-coldcache (...503...); using BOW
2026-08-07T16:05:26.113165Z  INFO trusty-analyze listening on http://127.0.0.1:62015
```

— matching the production log's "token file found" → 31m46s gap → "loaded" → listening ~1ms later.

With a fully warm cache the pre-fix boot was 0.307s and made no request, which is why four other observed boots looked fine. The stall is a cache miss on an unbounded call, not ONNX load time.

## Fix

Removed: `embedder/neural.rs`, the `fastembed` dependency, the `bundled-ort` / `load-dynamic` / `cuda` features (`default` is now `["http-server"]`), `EmbedderKind::Neural`, the neural branch and its `spawn_blocking` wrapper in `clusters_for_index`, and the UI's method dropdown.

`bow` is the sole embedder. `BowEmbedder::embed_batch` is pure hashing — no model, no cache, no network, and it cannot fail.

### How a late init failure surfaces

It cannot arise: there is no deferred initialization. The fix is removal, not lazy-init. `AnalyzerAppState::new` installs `BowEmbedder` by construction, and `run_serve` no longer calls `with_embedder`. The only fallible step left on the boot path is opening the two redb stores, which already propagates with `?`.

The fail-open that *did* exist is gone in both places:

- **Startup:** a model that failed to load logged `warn` and installed BOW while the API kept advertising neural. No load, no warn, no substitution.
- **Request:** `clusters_for_index` caught embedder errors and silently fell back to BOW. Removed — the remaining embedder is infallible, so a catch there could only hide something new. `?method=neural` returns 400 naming the value; pre-fix it returned **200 with `"method":"neural"` and `dim: 256`** (BOW vectors labelled as semantic ones — captured in the red run below).

### Blast radius beyond the crate

- `trusty-installer`: `ORT_CRATES` drops `trusty-analyze`. With no ONNX Runtime linked, its Linux asset has no elevated glibc floor and the workflow no longer publishes an AL2023 variant — routing a low-glibc host there would 404 on an asset that is never built.
- `.github/workflows/release.yml`: `AL2023_VARIANT="false"` for `trusty-analyze`. It now takes the ordinary glibc-2.17 zigbuild legs.
- `--fastembed-cache` / `TRUSTY_FASTEMBED_CACHE` kept as a hidden no-op so existing launchd plists and wrappers keep starting rather than failing on an unknown flag.

`ort` / `tokenizers` stay, behind the off-by-default `ner` feature. That is a separate capability, not on the boot path; removing it is a different outcome.

## Tests

Timing assertions would be flaky, so the guards assert the initialization *cannot* happen rather than that it was fast.

| Test | Asserts |
|---|---|
| `analyze_declares_no_in_process_model_deps` | the manifest declares no `fastembed` — no boot path can pay the cost |
| `analyze_declares_no_ort_backend_features` | no `bundled-ort` / `load-dynamic` / `cuda` to turn one back on |
| `embedder_kind_has_only_bow` | `"neural"` no longer deserializes into `EmbedderKind` |
| `clusters_reject_removed_neural_method` | `?method=neural` → 400 with a JSON error naming both values |
| `clusters_return_bow_vectors_for_a_live_corpus` | **used path still works** — 6-chunk corpus, 200, `method: "bow"`, `dim: 256`, every chunk clustered |
| `select_analyze_is_no_longer_an_ort_crate` | a below-floor x86_64 host gets the native asset, not AL2023 |

### Red before green

Run on a throwaway worktree detached at `origin/main` with only the new tests applied (`--no-default-features --features http-server` to skip the ORT build):

```
---- embedder::tests::analyze_declares_no_in_process_model_deps stdout ----
Found: ["fastembed  = { workspace = true, features = [\"hf-hub-rustls-tls\"], optional = true }"]

---- embedder::tests::analyze_declares_no_ort_backend_features stdout ----
Found: ["bundled-ort  = [\"dep:fastembed\", \"fastembed/ort-download-binaries\"]", "load-dynamic = [\"dep:fastembed\", \"fastembed/ort-load-dynamic\"]", "cuda         = [\"dep:fastembed\", \"fastembed/cuda\", \"fastembed/ort-load-dynamic\"]"]

---- embedder::tests::embedder_kind_has_only_bow stdout ----
`neural` must no longer deserialize — a caller asking for the removed backend has to be told, not silently given BOW vectors

---- service::tests::clusters_reject_removed_neural_method stdout ----
assertion `left == right` failed: body: {"chunk_count":6,...,"dim":256,"iterations":2,"k":3,"method":"neural"}
  left: 200
 right: 400

test result: FAILED. 2 passed; 4 failed; 0 ignored; 0 measured; 451 filtered out
```

```
---- download::glibc::tests::proof_5067::select_analyze_is_no_longer_an_ort_crate stdout ----
assertion failed: !ORT_CRATES.contains(&"trusty-analyze")

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 558 filtered out
```

Green on this branch:

```
running 9 tests
test embedder::bow::tests::bow_embedder_reports_correct_kind_and_dim ... ok
test embedder::tests::embedder_kind_has_only_bow ... ok
test embedder::bow::tests::bow_embedder_produces_normalized_256d_vectors ... ok
test embedder::bow::tests::bow_embedder_different_texts_differ ... ok
test core::concept_cluster::tests::stable_clusters_on_synthetic_data ... ok
test embedder::tests::analyze_declares_no_in_process_model_deps ... ok
test embedder::tests::analyze_declares_no_ort_backend_features ... ok
test service::tests::clusters_reject_removed_neural_method ... ok
test service::tests::clusters_return_bow_vectors_for_a_live_corpus ... ok

test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 448 filtered out; finished in 0.20s
```

## Gate — Rust Test Ladder rung 5 (process lifecycle + release tooling + performance claim), plus rung 6 for the UI/MCP surface

```
cargo fmt --check                                             EXIT=0
cargo check --workspace --all-targets
    --exclude trusty-code-gui --exclude trusty-mpm-gui        EXIT=0
cargo clippy -p trusty-analyze --all-targets -- -D warnings   EXIT=0
cargo clippy -p trusty-installer --all-targets -- -D warnings EXIT=0
cargo test -p trusty-analyze                  457 + 25 + 2 + 2 passed, 0 failed, 1 ignored
cargo test -p trusty-analyze -- --include-ignored   3 passed, 0 failed, 0 ignored (stdio_harness)
cargo test -p trusty-installer                555 + 3 + 3 passed, 0 failed, 4 ignored
bash scripts/check_line_cap.sh          3762 files, 0 violations — OK
bash scripts/check_changelog_fragment.sh 22 paths, 2 crates recorded — OK
```

Two notes, neither hidden:

- **`cargo check --workspace` excludes `trusty-code-gui` / `trusty-mpm-gui`.** A fresh worktree has no built Tauri UI dist, so `tauri::generate_context!()` panics with ``The `frontendDist` configuration is set to `"ui/dist"` but this path doesn't exist``. Environment gap, not a code failure; the diff touches neither crate.
- **`--include-ignored` first failed** with `release binary not found at target/release/trusty-analyze — run cargo build --release -p trusty-analyze first`. That is the test's own stated prerequisite. After `cargo build --release -p trusty-analyze` it passes; `full_stdio_suite` exercises `cluster_concepts` end-to-end over MCP stdio against a real index, which is the binary smoke run rung 6 asks for.

The embedded UI (`ui/dist`) was rebuilt with `pnpm install --frozen-lockfile && pnpm build` — the tracked bundle no longer carries the `neural` option, which matters because a stale bundle would now send a request the API rejects.

Closes #5067


bobmatnyc:
## Verdict: BLOCK

Reviewed at `2dbfef07` against the spec in [#5067](https://github.com/bobmatnyc/trusty-tools/issues/5067). The diagnosis, the removal-over-lazy-init choice, and the two fail-open deletions are right. Two gates are deterministically red on this head, one of them a build break that survives the merge.

## Findings

|---|---|---|---|---|---|
| CRITICAL | `.github/workflows/al2023-build.yml` | 253–258 | The `al2023-load-dynamic` job builds `cargo build -p trusty-analyze --no-default-features --features http-server,load-dynamic`. This PR deletes that feature. Job fails with `error: the package 'trusty-analyze' does not contain this feature: load-dynamic` ([job log](https://github.com/bobmatnyc/trusty-tools/actions/runs/31200054881/job/92937630103)). It is not scoped to this PR: the workflow's `paths:` filter (lines 63–67, 79–83) also fires on `crates/trusty-search/**`, `crates/trusty-embedderd/**` and `crates/trusty-common/**`, so after merge every PR touching any of those four crates gets the same red. | Delete the `al2023-load-dynamic` job (lines 106–258) and drop `crates/trusty-analyze/**` from both `paths:` filters; `al2023-load-dynamic-search` already covers the only remaining ORT crate. Update the file header prose at lines 11–13 and 26–27, which names trusty-analyze as the proof case. | Fix here |
| HIGH | `crates/trusty-analyze/src/main.rs` | 92 | `/// Test: \`serve_accepts_deprecated_fastembed_cache_flag\`.` names a test that does not exist. `scripts/check_test_pointers.sh` fails on it locally and in CI ([job log](https://github.com/bobmatnyc/trusty-tools/actions/runs/31200054956/job/92937630602)): `1 dangling pointer(s)`. The pointer is not merely stale — the compatibility promise it names (an existing launchd plist passing `--fastembed-cache` still starts) has zero coverage anywhere in the diff. | Add the test rather than delete the pointer. A `#[cfg(test)]` block in `main.rs` asserting `Cli::try_parse_from(["trusty-analyze", "serve", "--fastembed-cache", "/x"]).is_ok()` closes the lint and the coverage gap in one move. | Fix here |
| MEDIUM | `docs/distribution/INSTALL-CONVENTION.md` | 33, 40, 229, 327 | Still lists `trusty-analyze` as an ONNX-runtime crate with an AL2023 variant, and gives two literal `cargo install trusty-analyze --no-default-features --features http-server,load-dynamic --locked` recipes. Both now abort with `does not contain this feature`. Same instruction survives at `docs/trusty-analyze/spec/PRD.md:273`, `docs/trusty-analyze/spec/ARCHITECTURE.md:265`, `docs/reference/common-pitfalls.md:77-78`, and `docs/reference/environment-variables.md:19`. | Remove `trusty-analyze` from the ORT/AL2023 lists and delete the recipes. The two spec docs are SLD-governed surfaces, so they belong in this PR, not a follow-up. | Fix here |
| MEDIUM | `crates/trusty-analyze/Cargo.toml` | 3 | `version = "0.8.0"` unchanged, correct per the local-ops release policy — but nothing in the PR records that the next release must be **0.9.0**, not 0.8.1. This removes public library API: `EmbedderKind::Neural`, `NeuralEmbedder`, `ClusterQueryParams::method`'s type (`Option<EmbedderKind>` → `Option<String>`), and three Cargo features. Under 0.x that is a breaking bump, and CI is [known blind to SemVer breaks](https://github.com/bobmatnyc/trusty-tools/issues/4088) — the same gap that yanked `trusty-analyze` 0.7.3. | One line in `changelog.d/5067-remove-neural-embedder.md` or the PR body stating the required 0.9.0 minor bump. | Fix here |
| LOW | `crates/trusty-analyze/help.yaml` | 23–26 | Still declares `fastembed-cache` with `default: .fastembed_cache` and `description: Path to the fastembed model cache`. The clap arg is now `hide = true` with no default. The file's own header says it "Mirrors the clap definitions in `src/main.rs`". Impact is contained — this YAML feeds `print_suggestion_hint`, not `--help` rendering. | Delete the four lines. | Fix here |
| LOW | `crates/trusty-analyze/CLAUDE.md` | 466 | `\| Embeddings \| fastembed 5.x (optional; uses cached model from trusty-search) \|` | Change to hashed bag-of-words. | Fix here |

## Required Changes

1. Delete the `al2023-load-dynamic` job and its path filter entry in `.github/workflows/al2023-build.yml`.
2. Add `serve_accepts_deprecated_fastembed_cache_flag` (or rename the pointer to a test that exists).
3. Correct the six documentation sites that still hand out a `load-dynamic` build recipe for `trusty-analyze`.
4. Record the 0.9.0 bump requirement.
5. Drop the stale `help.yaml` flag entry and the `CLAUDE.md` row.

Re-run `bash scripts/check_test_pointers.sh` (currently `EXIT=1`) before re-review.

## Answers to the five review questions

**1. Blast radius on existing data — nothing persists. Question closed.**

I traced every consumer of the mislabelled response. `clusters_for_index` (`crates/trusty-analyze/src/service/handlers/graph.rs:256-297`) fetches chunks live from trusty-search, embeds, k-means, returns JSON. It writes nothing: no `FactStore` call, no SCIP overlay write, no cache. `grep` for `Vec<f32>` in `core/facts.rs`, `core/scip_overlay.rs` and `core/registry.rs` returns nothing — trusty-analyze's two redb stores hold no vectors and no cluster assignments. The three consumers are all pass-through: `trusty-console`'s `analyze_visualize_handler` (`crates/trusty-console/src/server/mod.rs:682-780`) folds the result into a live JSON body for the browser and returns it; the embedded UI renders it; the MCP `cluster_concepts` tool returns it to the caller. `crates/trusty-agents/src/context/cluster.rs`'s `clusters.jsonl` is unrelated — memory consolidation with its own embedder.

So `?method=neural` returning `"method":"neural"` with `dim: 256` was wrong on the wire and nowhere else. Nothing to detect, nothing to clear. Worth noting for anyone auditing old logs: the mislabelling was self-inconsistent, since a real neural response carried `dim: 384` and the substituted one carried 256.

**2. The 400 is the right contract, but the version story is incomplete.**

Returning BOW under a `"method":"bow"` label would be honest, and it is what the endpoint does when `method` is omitted. Accepting `neural` and answering `bow` is not the same thing — that is precisely the substitution the pre-fix daemon performed silently, and keeping it would make it permanent. The 400 is correct.

Callers checked: `trusty-console/src/server/mod.rs:714` sends `{index_id, k}` with no `method` — unaffected. `crates/trusty-analyze/src/mcp/mod.rs:452-460` only appends `method` when the caller supplies one, and the schema at `descriptors.rs:148` now pins `"enum": ["bow"]`. `ui/src/lib/api.js:57` defaults to `bow`; `Clusters.svelte` hardcodes it and the tracked `ui/dist` bundle was rebuilt. No caller in-tree hits the 400 path.

The gap is SemVer, finding 4 above — the removed public items make this a 0.9.0, and nothing in the PR says so.

**3. The coupling is not airtight. It missed a third build site.**

`release.yml` and `ORT_CRATES` land together, so the routing itself has no window. But `al2023-build.yml` also builds the removed feature and is now red (finding 1).

On the installed-client window: an old `tctl` carries `trusty-analyze` in its compiled-in `ORT_CRATES`. On a below-floor x86_64 Linux host it will request `trusty-analyze-<v>-x86_64-linux-al2023.tar.gz`, which the new release does not build. That 404 is handled — `try_install_prebuilt` (`crates/trusty-installer/src/download/mod.rs:75-155`) returns `Outcome::Fallback` on any failure including 404, and the caller proceeds to `cargo install`. The client experiences a slow source build, not a failure. There is one narrow residual: between the GitHub release of the new trusty-analyze and its crates.io publish, that fallback compiles the *old* ORT-linked version and produces a binary that dies on `GLIBC_2.39`. Population is below-glibc-2.38 x86_64 Linux hosts running a stale tctl; I am not filing it, but sequencing the crates.io publish before or with the GitHub release removes it entirely.

**4. Both non-removals are sound. The stall is not relocated.**

`ner` does not touch hf-hub. `NerExtractor::try_load` (`crates/trusty-analyze/src/core/ner.rs:45-75`) reads a local file at `~/.trusty-analyzer/models/ner.onnx`, returns a disabled extractor when it is absent, and never downloads. `ort`/`tokenizers` are `optional = true`, reachable only through `ner`, which the release workflow does not enable (`CARGO_FEATURES=""`). No cold-cache path, and nothing on the boot path — `try_load` is reached from entity extraction, not startup.

`--fastembed-cache` as a hidden no-op is the right call for an installed launchd plist. It is also the one behavior in this PR with no test at all (finding 2).

**5. Test quality — mixed, and the split matters.**

Two of the nine are genuine runtime behavior tests through the real router: `clusters_return_bow_vectors_for_a_live_corpus` and `clusters_reject_removed_neural_method` (`crates/trusty-analyze/src/service/tests.rs:718-776`) drive `/clusters` end to end against a stub corpus and assert status, `method`, `dim`, and full chunk coverage. `embedder_kind_has_only_bow` pins the deserialization contract, which is the actual mechanism behind the 400. Those three are not build-config assertions.

`analyze_declares_no_in_process_model_deps` and `analyze_declares_no_ort_backend_features` (`crates/trusty-analyze/src/embedder/tests.rs:42-79`) are exactly what the question suspects: they read `Cargo.toml` as a string. They would stay green if someone reintroduced a boot-time download through a different dependency — `hf-hub` directly, or a fetch under the `ner` feature. As named guards against *these three feature names and this one crate name* they are honest and they did fail red on `origin/main`; as a statement that boot performs no network I/O they are weak. That weakness is acceptable here only because the removal is structural: `run_serve` no longer has an embedder parameter, so the type system carries most of the guarantee the manifest test is gesturing at.

The real coverage hole is not in the four guards. It is the deprecated flag, which is the only behavior in this PR that is a promise rather than a deletion, and it has no test — while its doc comment claims one.

## Notes

Both red checks are deterministic, not flaky. The AL2023 failure is a cargo feature-resolution error that cannot vary by run, and I reproduced the test-pointer failure locally on `2dbfef07`:

```
FAIL: crates/trusty-analyze/src/main.rs:92: Test: pointer cites `serve_accepts_deprecated_fastembed_cache_flag` — no `fn serve_accepts_deprecated_fastembed_cache_flag(` or `mod serve_accepts_deprecated_fastembed_cache_flag` found in crate `crates/trusty-analyze`.
test-pointers: 1 dangling pointer(s), 0 stale allowlist entries — FAILED.
```

Not flagged: `resolve_method` accepts `BOW`/`Bow` case-insensitively where the old serde `rename_all = "lowercase"` derive would have rejected them. That is a loosening, not a defect.

I did not re-run the author's gates or the boot harness, and I did not re-run CI.

bobmatnyc:
## Review round — all six findings addressed on `6f7f8f03`

**CRITICAL — third build site.** `.github/workflows/al2023-build.yml` built `-p trusty-analyze --features http-server,load-dynamic`. Job and its two `paths:` entries deleted (417 → 262 lines); the `trusty-search` job stays with its cross-references repaired. Verified the deleted job's name is **not** in `required_status_checks.contexts` on `main`, so nothing is left permanently pending.

Re-grepped the whole `.github/` tree as instructed. Remaining hits are all inert: `ci.yml` (analyzer UI vite build, fastembed cache for *other* crates' tests), `token-drift.yml` (UI tokens), `e2e-docker.yml` (a version input, no feature flags). No fourth site.

**HIGH — missing test.** `serve_accepts_deprecated_fastembed_cache_flag` added in a new `src/main_tests.rs` (main.rs sits near the 500-SLOC cap; `_tests.rs` is classified as a test file). Pointer kept, not deleted. Red before:

```
FAIL: crates/trusty-analyze/src/main.rs:92: Test: pointer cites `serve_accepts_deprecated_fastembed_cache_flag` — no `fn ...` or `mod ...` found in crate `crates/trusty-analyze`.
test-pointers: 1 dangling pointer(s), 0 stale allowlist entries — FAILED.
```

Green after — and the flag now has real coverage: it parses, carries its value, is hidden from `--help`, and still binds `TRUSTY_FASTEMBED_CACHE` (a plist may set the env var instead of the flag).

```
test main_tests::serve_accepts_deprecated_fastembed_cache_flag ... ok
test main_tests::serve_without_fastembed_cache_parses_to_none ... ok
```

**MEDIUM 3 — stale install recipes.** Eight surfaces, not six: the five named plus `docs/architecture/harnesses.md` (claimed `trusty_common::embedder` is used by trusty-analyze — it never was), the crate `README.md`, and a second `INSTALL-CONVENTION.md` block at :320 that led with trusty-analyze. Historical records (`docs/research/**`, `docs/releases/**`, `regression-testing/**`) left alone — they are dated evidence.

**MEDIUM 4 — SemVer.** See the version section below.

**LOW 5 / LOW 6** — `help.yaml` `fastembed-cache` default removed, description now "Deprecated no-op, accepted and ignored (#5067)"; crate `CLAUDE.md` embedding-stack row and four other neural references corrected.

Per the critic's note, no new manifest-string assertions were added — the hole was finding 2, and that is what got the test.

## Versions — the gate is green without a bump

I ran the `Public API / SemVer` gate itself (`scripts/check_semver.sh` from `origin/main`) against this branch rather than guessing:

```
SKIP trusty-analyze: 0.7.4 -> 0.8.0 is already a major release — every lint is inapplicable
CHECK trusty-installer: 0.5.0 -> 0.5.1 (minor release), 0 feature(s)
    Checking trusty-installer v0.5.0 -> v0.5.1 (minor change)
     Checked [   0.018s] 196 checks: 196 pass, 58 skip
semver gate: scanned 22 changed path(s); 1 crate(s) checked, 1 skipped — OK.
```

**`trusty-analyze` needs no bump.** In-tree `0.8.0` is unpublished (crates.io latest non-yanked is `0.7.4`; `0.7.3` is the yanked one from #4088). `0.7.4 → 0.8.0` is already a major bump under Cargo's 0.x rules, so the removed API belongs in the pending `0.8.0` and the gate skips it. Going to `0.9.0` would burn a version for nothing. I left the version line untouched and did not run `bump-version.sh`.

**`trusty-installer` passes, but by a coverage gap.** Narrowing `pub const download::glibc::ORT_CRATES` from `[&str; 2]` to `[&str; 1]` **is** breaking — the length is part of the type — yet `cargo-semver-checks` has no lint for a `const`'s type changing, so it reports 196 pass. That is exactly the #4088 blind spot. Recorded in the installer's changelog fragment so a releaser decides deliberately; not silently bumped.

`Cargo.lock` carries exactly one delta across the whole PR — the removed `fastembed` edge — with zero transitive version churn, so MSRV is untouched.

If the owner still wants the bumps made here, say so and I will do them as a version-line-only edit — but the gate does not currently require either.

## Gates — full rung-5 chain re-run on `6f7f8f03`

```
cargo fmt --check                                             EXIT=0
cargo check --workspace --all-targets
    --exclude trusty-code-gui --exclude trusty-mpm-gui        EXIT=0
cargo clippy -p trusty-analyze --all-targets -- -D warnings   EXIT=0
cargo clippy -p trusty-installer --all-targets -- -D warnings EXIT=0

cargo test -p trusty-analyze                     EXIT=0
  test result: ok. 457 passed; 0 failed; 0 ignored
  test result: ok.  27 passed; 0 failed; 0 ignored
  test result: ok.   2 passed; 0 failed; 0 ignored
  test result: ok.   2 passed; 0 failed; 1 ignored

cargo test -p trusty-analyze -- --include-ignored EXIT=0
  test result: ok. 457 passed; 0 failed; 0 ignored
  test result: ok.  27 passed; 0 failed; 0 ignored
  test result: ok.   2 passed; 0 failed; 0 ignored
  test result: ok.   3 passed; 0 failed; 0 ignored   (stdio_harness, 67.93s)

cargo test -p trusty-installer                    EXIT=0
  test result: ok. 555 passed; 0 failed; 4 ignored
  test result: ok.   3 passed; 0 failed; 0 ignored
  test result: ok.   3 passed; 0 failed; 0 ignored

bash scripts/check_test_pointers.sh   22320 citations, 0 dangling — OK      (was FAILED)
bash scripts/check_sld.sh             56 specs + 3123 code files, 0 errors, 0 warnings
bash scripts/check_line_cap.sh        3763 files, 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh  32 paths, both crates recorded — OK
bash scripts/check_semver.sh          1 checked, 1 skipped — OK
```

`cargo check --workspace` still excludes the two Tauri GUI crates: a fresh worktree has no built `ui/dist`, so `tauri::generate_context!()` panics. Environment gap; the diff touches neither crate.

## Residual noted for the releaser

Between the GitHub release and the crates.io publish there is a narrow window where an old `tctl` could request an AL2023 asset that is no longer built. A 404 returns `Outcome::Fallback` → `cargo install`, so the worst case is a slow source build, not a failed install.

bobmatnyc:
## `trusty-installer` bumped 0.5.1 → 0.6.0 (`49e9ea15`)

Verified the numbers rather than taking them on relay: declared version was `0.5.1` (`crates/trusty-installer/Cargo.toml:3`), crates.io latest non-yanked is `0.5.0`. Under Cargo's 0.x rules a break takes the MINOR position, so 0.6.0.

**Version line only.** `Cargo.lock` synced with `cargo metadata --offline` — not `bump-version.sh` — and moved by exactly one line:

```
 [[package]]
 name = "trusty-installer"
-version = "0.5.1"
+version = "0.6.0"
```

Zero external re-resolution, so MSRV is untouched. `trusty-mpm` stays 1.3.5, `trusty-analyze` stays 0.8.0. No `[workspace.dependencies]` entry exists for `trusty-installer` — nothing in the workspace depends on it — so no floor change was needed.

Post-bump the SemVer gate skips both crates outright:

```
SKIP trusty-analyze:   0.7.4 -> 0.8.0 is already a major release — every lint is inapplicable
SKIP trusty-installer: 0.5.0 -> 0.6.0 is already a major release — every lint is inapplicable
semver gate: scanned 31 changed path(s); 0 crate(s) checked, 2 skipped — OK.
```

The installer changelog fragment now states plainly that `cargo semver-checks` **does not catch this shape** — it has no lint for a `const`'s type changing, and its pre-bump `196 checks: 196 pass` was absence of coverage, not evidence of compatibility. A future reader will not mistake that green for a compatibility finding.

## Gates re-run on `49e9ea15`

```
cargo fmt --check                                             EXIT=0
cargo clippy -p trusty-analyze --all-targets -- -D warnings   EXIT=0
cargo clippy -p trusty-installer --all-targets -- -D warnings EXIT=0
cargo check --workspace --all-targets
    --exclude trusty-code-gui --exclude trusty-mpm-gui        EXIT=0

cargo test -p trusty-analyze     457 / 27 / 2 / 2 passed, 0 failed, 1 ignored
cargo test -p trusty-installer   555 / 3 / 3 passed, 0 failed, 4 ignored

bash scripts/check_test_pointers.sh       22363 citations, 0 dangling — OK
bash scripts/check_sld.sh                 56 specs + 3123 code files, 0 errors, 0 warnings
bash scripts/check_line_cap.sh            3766 files, 7 allowlisted, 0 violations — OK
bash scripts/check_changelog_fragment.sh  both crates recorded — OK
bash scripts/check_semver.sh              0 checked, 2 skipped — OK
```

Not merged.

PR #5124, opened 2026-08-07
fix(security): enforce UDS socket permissions the ADRs already claim exist

### LINKED ISSUE #5099: UDS sockets have no permissions enforcement — ADR-0031/0032 cite a 0600 socket that no production code creates
## Defect

No production code in this workspace sets permissions on any Unix-domain socket. Sockets are created at the process umask (commonly `0755`), and two of the three path conventions place them under `$TMPDIR` with a `/tmp` fallback — world-writable on a Linux host with `TMPDIR` unset.

[ADR-0031](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0031-uds-for-inter-crate-transport-http-for-external.md) and [ADR-0032](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0032-no-service-owns-http-console-is-the-only-http-surface.md) both rest their access-control argument on the socket being `0600`. That property does not exist.

## Evidence

`grep -rn "set_permissions\|PermissionsExt" crates/` returns 60+ hits; every one is a test fixture, a credential file (`inference/credentials/file_store.rs`), a trust file (`core/project_trust.rs`), or an installed binary. None is a socket.

Four production bind sites, none of which chmod:

| Site | Socket path |
|---|---|
| `crates/trusty-embedderd/src/uds_server.rs:39` | `$TMPDIR/trusty-embedderd.sock`, `/tmp` fallback |
| `crates/trusty-bm25-daemon/src/server.rs:39` | `$TMPDIR/trusty-bm25-<palace>.sock`, `/tmp` fallback |
| `crates/trusty-agents/src/ctrl/socket.rs:193,225` | `~/.trusty-agents/sockets/<project>.ctrl.sock` |
| `crates/trusty-agents/src/bus/mod.rs:82` | `~/.trusty-agents/sockets/<project>.sock` |

The parent directory is created with `create_dir_all`, which applies the umask — so `~/.trusty-agents/sockets/` is typically `0755` too. No accept path checks peer credentials.

[ADR-0034](https://github.com/bobmatnyc/trusty-tools/pull/5090) states the requirement: `0700` on the containing directory, `0600` on the socket, and a peer-uid check on accept.

## Resolution

One shared implementation in `trusty-common` that all four bind sites route through:

- Parent directory created atomically at `0700` via `DirBuilder::mode()` — not `create_dir_all` followed by a chmod, which has the same ordering window as the socket.
- Socket `chmod 0600` immediately after bind, before the first accept.
- Peer-uid check on accept: `SO_PEERCRED` on Linux, `getpeereid` on macOS. Refuse any connection whose uid is not the process's own.
- `$TMPDIR`-with-`/tmp`-fallback sockets move into a per-uid subdirectory so the `0700` directory mode is enforceable. `/tmp` itself cannot be `0700`.

Shaped so [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) step 1's shared UDS module adopts it rather than reimplementing it.

bobmatnyc:
## Defect

No production code in this workspace set permissions on any Unix-domain socket. Every `set_permissions` / `PermissionsExt` hit was a test fixture, a credential file, a trust file, or an installed binary — none was a socket. Sockets were created at the process umask (commonly `0755`), and two of the three path conventions placed them in `$TMPDIR` falling back to `/tmp` — mode `1777`, owned by root, on a Linux host with `TMPDIR` unset.

[ADR-0031](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0031-uds-for-inter-crate-transport-http-for-external.md) and [ADR-0032](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0032-no-service-owns-http-console-is-the-only-http-surface.md) both rest their case for UDS over loopback TCP on "a `0600` socket". [ADR-0034](https://github.com/bobmatnyc/trusty-tools/pull/5090) states the requirement it must actually meet: `0700` on the containing directory, `0600` on the socket, peer-uid check on accept.

## Resolution

New `trusty_common::uds` — the single entry point all four bind sites route through.

**Four bind sites, not three.** ADR-0034's table names three *clients*; the binds live on the server side, and `trusty-agents`' message bus is a fourth site the ADR does not list:

| Site | Was |
|---|---|
| `trusty-embedderd/src/uds_server.rs` | bare `UnixListener::bind` |
| `trusty-bm25-daemon/src/server.rs` | bare `UnixListener::bind` |
| `trusty-agents/src/ctrl/socket.rs` (`bind` + `bind_singleton`) | `create_dir_all` + bare bind |
| `trusty-agents/src/bus/mod.rs` | `create_dir_all` + bare bind |

### The ordering trap is closed by the directory, not the chmod

Binding creates the file, so a `chmod` after `bind` leaves a window in which the socket exists at the umask-derived mode. `prepare_socket_dir` creates the parent at `0700` through `DirBuilder::mode()`, which passes the mode to `mkdir(2)` **atomically** — unlike `create_dir_all` followed by `set_permissions`, which reproduces the same ordering bug one level up. Unix path resolution requires search permission on every component, so no other uid can *name* the socket at any point, including during the window. The `0600` chmod after bind is defence in depth.

Setting the process umask around the bind was rejected: umask is process-global, so under a multi-threaded tokio runtime a sibling thread creating an unrelated file would silently inherit it.

### The `$TMPDIR`-with-`/tmp` fallback

Addressed, not waived. `/tmp` can be neither narrowed to `0700` nor trusted, so `uds::scratch_socket_dir()` interposes a uid-keyed `<$TMPDIR or /tmp>/trusty-<uid>/`. All four path resolvers (two clients in `trusty-common`, the bm25 daemon, the search embedder supervisor) moved in lockstep so client and daemon still agree.

Two consequences, both deliberate: **a running daemon must be restarted** to be found at the new path, and the extra segment costs 11 bytes of the kernel's `sun_path` budget (104 on macOS, where `$TMPDIR` alone is ~50), narrowing usable palace-name length from ~37 to ~26 characters.

### Peer-uid check

`SO_PEERCRED` on Linux, `getpeereid` on macOS/BSD, via `libc` — already in `[workspace.dependencies]`, so no new dependency; its `trusty-common` target gate widens from macOS-only to `cfg(unix)`. A unix target with neither syscall fails closed rather than compiling to an unchecked accept.

### Coordination with #5089

Deliberately not a transport — no framing, no dialing, no JSON-RPC. [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) step 1 builds the shared UDS transport module; this is the security layer that module mounts, so it is adopted rather than becoming a fourth copy.

## Red-before-green

The fix was reverted in place (`create_dir_all`, bare bind, no chmod) and the regression tests re-run. `git stash` was avoided — the stash stack is repo-level and shared across this repo's concurrent worktrees.

```
test uds::tests::bind_hardened_sets_socket_0600_and_dir_0700 ... FAILED
panicked at crates/trusty-common/src/uds/tests.rs:122:5:
assertion `left == right` failed: socket must be 0600
  left: 493
 right: 384
test result: FAILED. 0 passed; 1 failed; 0 ignored

test uds::tests::prepare_socket_dir_creates_at_0700 ... FAILED
test uds::tests::prepare_socket_dir_narrows_a_wide_existing_dir ... FAILED
test uds::tests::prepare_socket_dir_creates_missing_ancestors ... FAILED
test uds::tests::prepare_socket_dir_is_idempotent ... FAILED
  left: 493     (0o755, umask-derived)
 right: 448     (0o700)
test result: FAILED. 0 passed; 4 failed; 1 ignored
```

Restored: `test result: ok. 12 passed; 0 failed; 2 ignored`.

Each migrated bind site also carries its own mode assertion — `bind_uds_listener_produces_a_0600_socket`, `bind_listener_produces_a_0600_socket`, `bind_produces_a_0600_socket_in_a_0700_dir`, `bind_singleton_hardens_the_socket_it_takes_over` — so a future site that stops routing through `bind_hardened` fails in its own crate.

## What is NOT covered

The two refusal paths — foreign peer uid and foreign directory owner — are `#[ignore]`d because they need a second uid (or root). They now **skip loudly rather than pass silently**, printing `SKIPPED: … This run proves NOTHING about the …refusal path.` **Do not read their `ok` under `--include-ignored` as coverage.** They are manual harnesses gated on `TRUSTY_UDS_FOREIGN_PEER_SOCK` / `TRUSTY_UDS_FOREIGN_DIR`.

## Gate — Rust Test Ladder rung 5 (security + cross-crate contract)

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` | EXIT=0 |
| `cargo clippy -p trusty-common --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-agents --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-memory --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-bm25-daemon --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-embedderd --all-targets -- -D warnings` | EXIT=0 |
| `cargo clippy -p trusty-search --all-targets -- -D warnings` | EXIT=0 |
| `cargo test -p trusty-common` (+`--features uds`) | 290 passed, 0 failed / 12 uds passed, 2 ignored |
| `cargo test -p trusty-agents` | 3379 passed, 0 failed, 11 ignored |
| `cargo test -p trusty-memory` | 603 passed, 0 failed, 4 ignored |
| `cargo test -p trusty-bm25-daemon` | 47 passed, 0 failed |
| `cargo test -p trusty-embedderd` | 27 passed, 0 failed |
| `cargo test -p trusty-search` | 1512 passed, 0 failed, 1 ignored |
| `--include-ignored` (uds, 3 crates) | 19 passed, 0 failed |
| `bash scripts/check_line_cap.sh` | 3765 files, 0 violations |
| `bash scripts/check_changelog_fragment.sh` | all 5 crates recorded |

**`cargo test -p trusty-common` alone does not run the uds tests** — `uds` is not a default feature. Feature unification enables it under any multi-package invocation, which is verified: `cargo test -p trusty-common -p trusty-embedderd uds::` → 17 passed. CI runs `cargo test --workspace`, so it is covered there.

No baseline failures encountered. The one test that broke — `trusty-memory`'s `already_running_skips_spawn` — was caused by this diff (it bound bare at the canonical path, now ENOENT inside the uid-keyed directory) and is fixed by binding the way the real daemon does.

`cargo check --workspace` needs `crates/{trusty-code-gui,trusty-mpm-gui}/ui/dist` to exist; a placeholder `index.html` was created locally and is gitignored, not committed.

Closes #5099

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11517,7 +11517,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-embedderd"
-version = "0.3.10"
+version = "0.3.11"
 dependencies = [
  "anyhow",
  "axum",
--- a/crates/trusty-agents/Cargo.toml
+++ b/crates/trusty-agents/Cargo.toml
@@ -58,7 +58,7 @@ trusty-code = { workspace = true }
 #      stub. Zero added dependency weight: `aws-config`/`aws-sdk-bedrockruntime`/
 #      `aws-types` below are already unconditional direct dependencies of this
 #      crate for the existing `llm::bedrock` (Converse, non-streaming) client.
-trusty-common = { workspace = true, features = ["mcp", "symgraph-parser", "memory-core", "cli-help", "stdio-mcp-client", "session-naming", "config-cli", "axum-server", "inference-client", "bedrock"] }
+trusty-common = { workspace = true, features = ["mcp", "symgraph-parser", "memory-core", "cli-help", "stdio-mcp-client", "session-naming", "config-cli", "axum-server", "inference-client", "bedrock", "uds"] }
 # Why (issue #226): trusty-agents only links `trusty_memory::MemoryMcpService` for
 #      in-process `rpc.discover` merging; it does not need the axum HTTP/SSE
 #      surface (it owns its own axum stack for the orchestrator's UI). Setting
--- a/crates/trusty-agents/changelog.d/5099-uds-socket-permissions.md
+++ b/crates/trusty-agents/changelog.d/5099-uds-socket-permissions.md
@@ -1,0 +1,12 @@
+Fixed
+
+- **The controller and message-bus sockets were world-readable**
+  ([#5099](https://github.com/bobmatnyc/trusty-tools/issues/5099)).
+  `CtrlSocket::bind`, `CtrlSocket::bind_singleton`, and `MessageBus::start` each
+  created `~/.trusty-agents/sockets/` with `create_dir_all` and then bound bare,
+  so both the directory and the socket landed at the process umask (`0755`).
+  All three now route through `trusty_common::uds::bind_hardened` (`0700`
+  directory, `0600` socket before the first accept); `bind_singleton` hardens the
+  directory *before* probing, so a takeover of a stale socket is narrowed too.
+  The ctrl and bus accept loops drop any connection whose peer uid is not this
+  process's own.
--- a/crates/trusty-agents/src/bus/mod.rs
+++ b/crates/trusty-agents/src/bus/mod.rs
@@ -71,15 +71,14 @@ impl MessageBus {
     /// and assert the envelope arrives.
     pub async fn start(project_id: &str) -> Result<Arc<Self>> {
         let socket_path = Self::socket_path_for(project_id)?;
-        if let Some(parent) = socket_path.parent() {
-            fs::create_dir_all(parent).await?;
-        }
         // Remove stale socket from a previous run.
         if socket_path.exists() {
             fs::remove_file(&socket_path).await.ok();
         }
 
-        let listener = UnixListener::bind(&socket_path)?;
+        // #5099: was `create_dir_all` + a bare bind, so both the sockets
+        // directory and the socket landed at the process umask.
+        let listener = trusty_common::uds::bind_hardened(&socket_path)?;
         let (tx, _) = broadcast::channel(CHANNEL_CAPACITY);
 
         let bus = Arc::new(Self {
@@ -255,6 +254,11 @@ async fn accept_loop(listener: UnixListener, tx: broadcast::Sender<BusEnvelope>)
     loop {
         match listener.accept().await {
             Ok((stream, _addr)) => {
+                // #5099: a foreign-uid peer is dropped without being served.
+                if let Err(e) = trusty_common::uds::ensure_peer_is_self(&stream) {
+                    tracing::warn!(error = %e, "bus accept_loop: rejected connection");
+                    continue;
+                }
                 let tx_clone = tx.clone();
                 tokio::spawn(handle_connection(stream, tx_clone));
             }
--- a/crates/trusty-agents/src/ctrl/socket.rs
+++ b/crates/trusty-agents/src/ctrl/socket.rs
@@ -180,17 +180,19 @@ impl CtrlSocket {
     /// Why: Encapsulates parent-dir creation + stale-file cleanup + the
     /// actual bind, which would otherwise be duplicated between the
     /// controller startup path and any future "restart the listener" code.
-    /// What: Creates `~/.trusty-agents/sockets/`, removes any stale file at
-    /// `path`, then binds. Caller is responsible for `tokio::spawn`-ing the
-    /// accept loop and removing the socket file on shutdown.
-    /// Test: `bind_creates_listener_and_replaces_stale_file`.
+    /// What: removes any stale file at `path`, then binds through
+    /// [`trusty_common::uds::bind_hardened`], which creates
+    /// `~/.trusty-agents/sockets/` at `0700` and narrows the socket to `0600`
+    /// before the first accept (#5099 — this used to be `create_dir_all` plus a
+    /// bare bind, so both landed at the process umask). Caller is responsible
+    /// for `tokio::spawn`-ing the accept loop and removing the socket file on
+    /// shutdown.
+    /// Test: `bind_creates_listener_and_replaces_stale_file`,
+    /// `bind_produces_a_0600_socket_in_a_0700_dir`.
     pub async fn bind(path: &Path) -> io::Result<UnixListener> {
-        if let Some(parent) = path.parent() {
-            tokio::fs::create_dir_all(parent).await?;
-        }
         // Remove stale socket before binding (otherwise bind fails with EADDRINUSE).
         Self::cleanup(path);
-        UnixListener::bind(path)
+        trusty_common::uds::bind_hardened(path).map_err(io::Error::other)
     }
 
     /// Singleton-safe bind: probe first, only clobber a confirmed-dead socket.
@@ -213,8 +215,11 @@ impl CtrlSocket {
     /// `bind_singleton_binds_when_socket_absent`, and
     /// `bind_singleton_binds_over_stale_socket`.
     pub async fn bind_singleton(path: &Path, timeout: Duration) -> io::Result<BindOutcome> {
+        // #5099: harden the directory BEFORE probing. The probe can succeed
+        // against a socket sitting in a wide directory, and the takeover branch
+        // below must not be the only path that narrows it.
         if let Some(parent) = path.parent() {
-            tokio::fs::create_dir_all(parent).await?;
+            trusty_common::uds::prepare_socket_dir(parent).map_err(io::Error::other)?;
         }
         match Self::probe(path, timeout).await {
             // A controller answered — refuse to clobber; hand the stream back.
@@ -222,7 +227,8 @@ impl CtrlSocket {
             // Stale socket (or no socket at all): safe to take over.
             Err(e) if is_connection_refused(&e) || e.kind() == io::ErrorKind::TimedOut => {
                 Self::cleanup(path);
-                Ok(BindOutcome::Bound(UnixListener::bind(path)?))
+                let listener = trusty_common::uds::bind_hardened(path).map_err(io::Error::other)?;
+                Ok(BindOutcome::Bound(listener))
             }
             // Any other error (permission denied, etc.): do NOT clobber.
             Err(e) => Err(e),
@@ -252,6 +258,47 @@ impl CtrlSocket {
 mod tests {
     use super::*;
     use std::path::PathBuf;
+
+    /// #5099 regression: `CtrlSocket::bind` must produce a 0600 socket in a
+    /// 0700 directory. Against the pre-fix commit it used `create_dir_all`
+    /// plus a bare bind, so both landed at the process umask (0755/0755).
+    #[tokio::test]
+    async fn bind_produces_a_0600_socket_in_a_0700_dir() {
+        use std::os::unix::fs::PermissionsExt as _;
+
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let dir = tmp.path().join("sockets");
+        let sock = dir.join("p.ctrl.sock");
+
+        let _listener = CtrlSocket::bind(&sock).await.expect("bind");
+
+        let mode = |p: &Path| std::fs::metadata(p).expect("stat").permissions().mode() & 0o777;
+        assert_eq!(mode(&sock), 0o600, "ctrl socket must be 0600");
+        assert_eq!(mode(&dir), 0o700, "ctrl socket dir must be 0700");
+    }
+
+    /// #5099: the singleton path hardens the directory before probing, so a
+    /// takeover of a stale socket also lands 0600/0700.
+    #[tokio::test]
+    async fn bind_singleton_hardens_the_socket_it_takes_over() {
+        use std::os::unix::fs::PermissionsExt as _;
+
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let dir = tmp.path().join("sockets");
+        let sock = dir.join("p.ctrl.sock");
+
+        let outcome = CtrlSocket::bind_singleton_default(&sock)
+            .await
+            .expect("bind singleton");
+        assert!(
+            matches!(outcome, BindOutcome::Bound(_)),
+            "must win the race"
+        );
+
+        let mode = |p: &Path| std::fs::metadata(p).expect("stat").permissions().mode() & 0o777;
+        assert_eq!(mode(&sock), 0o600, "ctrl socket must be 0600");
+        assert_eq!(mode(&dir), 0o700, "ctrl socket dir must be 0700");
+    }
 
     #[test]
     fn sanitize_project_id_replaces_unsafe_chars() {
--- a/crates/trusty-agents/src/ctrl/socket_listener.rs
+++ b/crates/trusty-agents/src/ctrl/socket_listener.rs
@@ -41,6 +41,11 @@ pub async fn spawn_socket_listener(listener: tokio::net::UnixListener) {
     loop {
         match listener.accept().await {
             Ok((stream, _addr)) => {
+                // #5099: a foreign-uid peer is dropped without being served.
+                if let Err(e) = trusty_common::uds::ensure_peer_is_self(&stream) {
+                    tracing::warn!(error = %e, "ctrl: rejected connection");
+                    continue;
+                }
                 tokio::spawn(handle_socket_connection(stream));
             }
             Err(e) => {
--- a/crates/trusty-bm25-daemon/Cargo.toml
+++ b/crates/trusty-bm25-daemon/Cargo.toml
@@ -39,7 +39,7 @@ crate-type = ["rlib"]
 # Uses the in-tree, zero-dep BM25 implementation from trusty-common
 # (`bm25` feature). Keeps the tokenizer's camelCase / alpha↔digit splits
 # in lock-step with every other consumer (trusty-search, trusty-memory).
-trusty-common = { workspace = true, features = ["bm25"] }
+trusty-common = { workspace = true, features = ["bm25", "uds"] }
 tokio = { workspace = true, features = ["full"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
--- a/crates/trusty-bm25-daemon/changelog.d/5099-uds-socket-permissions.md
+++ b/crates/trusty-bm25-daemon/changelog.d/5099-uds-socket-permissions.md
@@ -1,0 +1,11 @@
+Fixed
+
+- **The daemon socket was world-readable and lived in a shared directory**
+  ([#5099](https://github.com/bobmatnyc/trusty-tools/issues/5099)).
+  `bind_listener` called `UnixListener::bind` bare, leaving the socket at the
+  process umask, and `default_socket_path` resolved to `$TMPDIR` falling back to
+  `/tmp` — world-writable on a Linux host with `TMPDIR` unset. The bind now goes
+  through `trusty_common::uds::bind_hardened` (`0700` directory, `0600` socket),
+  the path resolves under `<$TMPDIR or /tmp>/trusty-<uid>/`, and `run_accept_loop`
+  drops any connection whose peer uid is not this process's own. **A running
+  daemon must be restarted** to be found at the new path.
--- a/crates/trusty-bm25-daemon/src/server.rs
+++ b/crates/trusty-bm25-daemon/src/server.rs
@@ -33,11 +33,14 @@ use crate::protocol::{
 /// Why: extracted so the binary's startup sequence (cleanup → bind) is
 /// readable and so tests can exercise the same bind path against a tempfile
 /// socket.
-/// What: returns the `UnixListener` ready to accept connections. The caller
-/// must have already removed any stale socket file at `path`.
-/// Test: covered by the integration test.
+/// What: returns the `UnixListener` ready to accept connections, bound through
+/// [`trusty_common::uds::bind_hardened`] so the containing directory is `0700`
+/// and the socket `0600` before the first accept (#5099 — this bind used to be
+/// bare, leaving the socket at the process umask). The caller must have already
+/// removed any stale socket file at `path`.
+/// Test: `bind_listener_produces_a_0600_socket`, plus the integration test.
 pub fn bind_listener(path: &Path) -> Result<UnixListener> {
-    UnixListener::bind(path)
+    trusty_common::uds::bind_hardened(path)
         .with_context(|| format!("bind unix domain socket at {}", path.display()))
 }
 
@@ -46,13 +49,19 @@ pub fn bind_listener(path: &Path) -> Result<UnixListener> {
 /// Why: a single-threaded accept loop with detached per-connection tasks
 /// scales well for the daemon's expected load (a handful of in-host
 /// clients, short-lived requests) without the complexity of a thread pool.
-/// What: loops `listener.accept().await`; on each accept, clones the
-/// `BatchQueue` handle and spawns [`handle_connection`].
+/// What: loops `listener.accept().await`; on each accept, verifies the peer's
+/// uid matches this process's own (#5099), then clones the `BatchQueue` handle
+/// and spawns [`handle_connection`]. A foreign-uid peer is dropped without being
+/// served.
 /// Test: covered by the integration test.
 pub async fn run_accept_loop(listener: UnixListener, queue: Arc<BatchQueue>) {
     loop {
         match listener.accept().await {
             Ok((stream, _addr)) => {
+                if let Err(e) = trusty_common::uds::ensure_peer_is_self(&stream) {
+                    tracing::warn!("rejected UDS connection: {e}");
+                    continue;
+                }
                 let queue = Arc::clone(&queue);
                 tokio::spawn(async move {
                     if let Err(e) = handle_connection(stream, queue).await {
@@ -337,6 +346,25 @@ mod tests {
         // Keep the TempDir alive — `rebuild` and write batches flush the
         // on-disk snapshot, so the data dir must outlive the queue.
         (BatchQueue::new(index, BatchConfig::default()), dir)
+    }
+
+    /// #5099 regression: the daemon's own bind path must produce a 0600 socket
+    /// in a 0700 directory. Against the pre-fix commit this bind was bare and
+    /// the socket came back at the process umask (0755 under 022).
+    #[tokio::test]
+    async fn bind_listener_produces_a_0600_socket() {
+        use std::os::unix::fs::PermissionsExt as _;
+
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let dir = tmp.path().join("sockets");
+        let sock = dir.join("bm25.sock");
+
+        let _listener = bind_listener(&sock).expect("bind");
+
+        let mode =
+            |p: &std::path::Path| std::fs::metadata(p).expect("stat").permissions().mode() & 0o777;
+        assert_eq!(mode(&sock), 0o600, "socket must be 0600");
+        assert_eq!(mode(&dir), 0o700, "socket dir must be 0700");
     }
 
     #[tokio::test]
--- a/crates/trusty-bm25-daemon/src/socket.rs
+++ b/crates/trusty-bm25-daemon/src/socket.rs
@@ -6,7 +6,7 @@
 //! logic keeps the two contracts identical.
 //!
 //! What: `default_socket_path(palace)` returns
-//! `$TMPDIR/trusty-bm25-<palace>.sock` (falling back to `/tmp` if `TMPDIR`
+//! `$TMPDIR/trusty-<uid>/trusty-bm25-<palace>.sock` (falling back to `/tmp`
 //! is unset or empty); `cleanup_stale_socket` best-effort removes the file
 //! at the given path, ignoring "not found".
 //!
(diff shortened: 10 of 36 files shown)


bobmatnyc:

Reviewed at `d9a654e6` (branch `fix/uds-socket-permissions`, 7 commits behind `origin/main`). Independent worktree, no CI re-run.

Three HIGH, no CRITICAL. The shape of the fix is right — one shared module, `0700` directory via `DirBuilder::mode()` so `mkdir(2)` applies the mode atomically, `0600` socket, peer-uid check on all four accept loops, every failure branch propagating. The directory-first argument for closing the bind→chmod window is correct **as far as it goes**. What it misses is that path resolution is only trustworthy when the components are what the code believes they are, and that the guarantee is one-sided.

One of the three is a mechanical CI gate that is red on this branch right now.

---

## Findings

|---|---|---|---|---|---|
| HIGH | [`crates/trusty-common/src/uds/mod.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/mod.rs#L233-L253) | 233–253 | `prepare_socket_dir` resolves the pre-existing directory through symlinks. `std::fs::metadata` and `std::fs::set_permissions` both follow; `DirBuilder::create` returns `AlreadyExists` on a symlink-to-dir. An attacker who pre-creates `/tmp/trusty-<victim-uid>` as a symlink to any victim-owned directory passes the owner check, gets that directory chmod'd to `0700`, and gets the daemon's socket bound inside it. | Stat and chmod the leaf without following: `std::fs::symlink_metadata(dir)` and reject when `file_type().is_symlink() \|\| !is_dir()` (new `UdsSecurityError` variant), or better, open with `O_NOFOLLOW\|O_DIRECTORY\|O_CLOEXEC` and use `fstat`/`fchmod` on the fd. | Fix here |
| HIGH | [`crates/trusty-common/src/uds/peer.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/peer.rs#L17) | 17, 128 | `scripts/check_test_pointers.sh` is red on this branch. Both `Test:` pointers cite `ensure_peer_is_self_accepts_same_uid`, which does not exist in the crate. This gate blocks merge and was not in the reported gate list. | Point both at the tests that actually exercise it: `bind_hardened_socket_is_connectable_after_hardening` and `peer_uid_of_self_connection_is_self`. | Fix here |
| HIGH | [`crates/trusty-common/src/bm25_client.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/bm25_client.rs#L107) | 107 | Server-side only. No client verifies the socket's owner or its directory's mode before `UnixStream::connect`. On the `/tmp` fallback an attacker who wins the race to create `/tmp/trusty-<uid>` (their own dir, `0777`) and plants a listener there gets every client's traffic: `bm25_supervisor.rs:379` probes the socket first and, seeing it accept, **never spawns the daemon** — so `ForeignDirOwner` never fires. Same shape for `embedder_client/uds.rs:133` and `trusty-search`'s `default_socket_path`. | Add the mirror primitive — `uds::connect_hardened(path)`: `symlink_metadata` the socket and its parent, refuse a foreign owner or a parent wider than `0700`, then connect. Route `socket_path_for_palace` callers, `UdsEmbedderClient`, and `probe_socket` through it. | Fix here |
| MEDIUM | [`crates/trusty-common/src/uds/mod.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/mod.rs#L279-L305) | 279–305 | No `sun_path` budget check. The extra `trusty-<uid>/` segment cuts the usable palace name from ~37 to ~26 chars on macOS; a name in between now dies at bind with a bare `invalid argument`. | Reject before `bind`: `path.as_os_str().as_bytes().len() >= size_of sun_path` → a named `UdsSecurityError::PathTooLong { path, len, limit }` that states the budget. | Fix here |
| MEDIUM | [`crates/trusty-common/src/uds/peer.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/peer.rs#L131-L138) | 131–138, and `mod.rs:237–244` | The two refusal branches are untestable only because the comparison is fused to the syscall. A pure `fn refuse_foreign_uid(peer: u32, expected: u32) -> Result<(), UdsSecurityError>` (and the same for the directory owner) covers the *decision* with no second uid; `peer_uid_of_self_connection_is_self` already covers the syscall wiring. | Extract both comparisons and unit-test the mismatch arm directly. | Fix here |
| MEDIUM | [`crates/trusty-common/src/lib.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/lib.rs#L741) | 741 | `Test: cargo test -p trusty-common -- uds::` runs zero tests and exits 0 — `default = []` and the module is `#[cfg(all(unix, feature = "uds"))]`. A maintainer verifying the security module gets a green no-op. | `cargo test -p trusty-common --features uds -- uds::` | Fix here |
| MEDIUM | [`crates/trusty-common/src/uds/peer.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/peer.rs#L9) | 9, 118–121 | ADR-0034 is already taken. [`0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md`](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md) landed on `main` 16 minutes after this branch's base. Line 118 quotes "ADR-0034" verbatim for a sentence that document does not contain, and this PR adds no ADR file. | Renumber to the next free slot and add the ADR, or drop the citation and point at [#5099](https://github.com/bobmatnyc/trusty-tools/issues/5099). | Fix here |
| LOW | [`crates/trusty-common/src/uds/tests.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/tests.rs#L207-L213) | 207–213, 231–237 | The loud skip is not loud. libtest captures `eprintln!` on a passing test, so under `--include-ignored` both report `ok` and contribute to the "19 passed" count while proving nothing. The doc comments say so honestly; the runner output does not. | Once the pure-function seam above lands, these two become pure syscall-integration probes — gate them on the env var with `panic!` instead of `return`, so `--include-ignored` cannot count them as passes. | Fix here |

---

1. `prepare_socket_dir` must not follow a symlink at the leaf (mod.rs:233–253).
2. Fix the two dangling `Test:` pointers — `scripts/check_test_pointers.sh` exits 1 on this branch:
   ```
   FAIL: crates/trusty-common/src/uds/peer.rs:17: Test: pointer cites `ensure_peer_is_self_accepts_same_uid` — no `fn ensure_peer_is_self_accepts_same_uid(` or `mod ensure_peer_is_self_accepts_same_uid` found in crate `crates/trusty-common`.
   FAIL: crates/trusty-common/src/uds/peer.rs:128: Test: pointer cites `ensure_peer_is_self_accepts_same_uid` — no `fn ensure_peer_is_self_accepts_same_uid(` or `mod ensure_peer_is_self_accepts_same_uid` found in crate `crates/trusty-common`.
   EXIT=1
   ```
3. Add the client-side half, or state in the module docs that the guarantee is server-side only and the connect side is unverified.
4. Resolve the ADR-0034 number collision.

---

## The six questions

**1. Does the directory argument hold?** Partly. `DirBuilder::mode()` → `mkdir(2)` is genuinely atomic, and the pre-existing-directory case *is* checked for owner and mode rather than accepted blindly — a real attacker-owned directory produces `ForeignDirOwner` and the bind fails closed. That is the right design. It is defeated by a symlink, verified on this host:

```
mkdir errno: EEXIST                          # DirBuilder falls into the "exists" branch
stat mode via link: 0o755 uid: 502           # metadata() reports the TARGET
target mode after chmod via link: 0o700      # set_permissions() chmods the TARGET
socket created in target: True               # bind() lands in the TARGET
```

So `mod.rs:195-197` — "no other uid can name the socket at any point, including during the window" — is false whenever the leaf is a symlink the attacker placed. Reachable on the `/tmp` fallback, which is the exact path the module exists to fix. `fs.protected_symlinks=1` mitigates on mainstream Linux, but that is an external control this code does not verify, which is the same posture as the `0600` claim this PR is correcting.

**2. Symlink and path substitution.** Answered above. `metadata` follows, so a symlink to a victim-owned directory is indistinguishable from the real thing. A symlink to an *attacker*-owned directory is caught. A path that exists and is not a directory: `DirBuilder::create` returns `AlreadyExists` on a regular file too, then `metadata` succeeds, the uid matches (the attacker can't create a victim-owned file under sticky `/tmp`, so realistically this is victim-owned), `set_permissions` chmods the file, and `bind` fails `ENOTDIR`. Fails closed, but for the wrong reason and after chmod'ing a file. Rejecting non-directories explicitly covers this in the same edit as finding 1. The post-check TOCTOU swap is not a practical residual once symlinks are rejected: `/tmp`'s sticky bit stops an attacker unlinking a victim-owned directory.

**3. `sun_path` truncation.** No silent truncation, and no collision — the failure is loud at both ends. Verified against the macOS boundary:

```
len=103 bound OK, file at exact path exists=True
len=104 FAILED: AF_UNIX path too long
```

Rust's `SocketAddr::from_pathname` rejects before the syscall with the same result. Client side gets `ENOENT`/`ECONNREFUSED`, never a wrong socket. What remains is the loudness *quality*: a 30-char palace name that worked yesterday now produces `bind unix domain socket at …: invalid argument`, which names neither the budget nor the cause. The PR handles the regression by shortening one test's palace name and adding a doc comment — that documents it for the next maintainer, not for the operator who hits it. Hence the MEDIUM.

**4. Peer-check completeness.** Complete on the accept side. Grepped every `UnixListener` and `.accept()` in `crates/`; four production accept loops exist and all four call `ensure_peer_is_self` and `continue` on error:

- `trusty-agents/src/ctrl/socket_listener.rs:45`
- `trusty-agents/src/bus/mod.rs:258`
- `trusty-bm25-daemon/src/server.rs:61`
- `trusty-embedderd/src/uds_server.rs:68`

`CtrlSocket::bind`'s only production consumer is `ctrl/repl/mod.rs:360` → `spawn_socket_listener`, covered. The remaining bare `UnixListener::bind` at `bus/mod.rs:218` is `#[cfg(test)]`. `PeerCred`, `ForeignPeer` and `UnsupportedPlatform` all reach the same `continue`, so there is no fail-open arm. The gap is not on this axis at all — it is the connect side (finding 3).

**5. The uncovered refusal paths.** The honesty is correct and I'm not penalising it. But a seam was available and not taken: fuse-free comparison functions (finding 5) cover both refusal *decisions* with no second uid, leaving only the syscall wiring privileged — and that half is already covered by `peer_uid_of_self_connection_is_self`. On reliability: the loud-skip does degrade to a silent pass. libtest captures `eprintln!` on a passing test, so under `--include-ignored` both return early, print nothing the runner shows, and count as `ok` inside the reported "19 passed".

**6. Migration.** No silent split-brain, and this is handled better than the framing implies. `bm25_supervisor.rs:413` spawns, polls the new path, and on timeout drops the `Child` (SIGKILL via `kill_on_drop`) and returns an error naming the expected path — a new client against an old on-disk binary fails closed and diagnostically, with no process leak. The ctrl and bus sockets keep their path under `~/.trusty-agents/sockets/`, so an old controller still answers `bind_singleton`'s probe and is not clobbered. `trusty-search`'s embedderd UDS path only applies under an explicit `TRUSTY_EMBEDDER=unix:/path`. Old client against a new daemon degrades to "unavailable". The one real window is finding 3: between the old daemon dying and the new one binding, the new path is an empty attacker-creatable directory under `/tmp`, and the supervisor's probe will trust whatever is listening there. Stale sockets at the old paths are orphaned rather than cleaned up — cosmetic, not flagged.

---

## Adoption readiness for [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) step 1

The bind side is ready. `bind_hardened` returns a bare `tokio::net::UnixListener` rather than an owned abstraction, so a transport wraps it instead of fighting it; the mode constants are public; the error type is a `thiserror` enum a transport can wrap; and declining to unlink stale sockets correctly leaves that policy with the caller. No framing, no dialing, no JSON-RPC — the separation claimed in the module docs is real.

The connect side is where the fourth copy will appear. Step 1 needs a dialer, this module gives it nothing for that, and there are already three independent `UnixStream::connect` call sites (`bm25_client`, `UdsEmbedderClient`, `probe_socket`) with no shared verification. Landing `connect_hardened` here — finding 3 — is what makes step 1 an adoption rather than a re-implementation. One consumer-side note: `uds` is not a default feature, so each adopting crate must add it or depend on a feature that implies it (`bm25-client` and `embedder-client` now do).

**ADR amendment.** `trusty-agents/src/bus/mod.rs` is a fifth UDS bind site the ADR does not list — and the ADR it does not list it in doesn't exist yet, since ADR-0034 belongs to webhook ingress. Whichever number the socket-permission ADR takes, it should enumerate all five (`trusty-embedderd/src/uds_server.rs`, `trusty-bm25-daemon/src/server.rs`, `trusty-agents/src/ctrl/socket.rs` ×2, `trusty-agents/src/bus/mod.rs`) and state whether the requirement is server-side only. ADR-0031 and ADR-0032 should also be amended to record that the `0600` premise they rested on was not true until this PR.

---

## Notes

- Changelog fragments present for all five source-touched crates. `trusty-memory`'s only change is in `bm25_supervisor_tests.rs`, which is test-only and exempt.
- SLOC: `uds/mod.rs` 305, `peer.rs` 138, `tests.rs` 243 — all under cap.
- Not flagged: the `& 0o777` mask in the mode comparison ignores setgid/sticky on the socket dir (harmless at `0700`); ancestor directories are not owner-checked (correct — only the leaf holds sockets); orphaned sockets at the pre-migration paths.
- Verdict is WARN, not BLOCK: the two security findings need the `/tmp` fallback plus a hostile local uid, and finding 3 is a pre-existing gap this PR narrows the framing of rather than introduces. Findings 1 and 2 should not merge as-is regardless — 2 cannot, mechanically.

bobmatnyc:
## Code Critic — Round 2 — Verdict: APPROVE

Scope: only what round 1 did not see. Boundary established from the round-1 comment's stated SHA `d9a654e6`; the branch was rebased onto [`9196c667`](https://github.com/bobmatnyc/trusty-tools/commit/9196c66755f6cc12f8068d80a5a3eb8172182b5d), so the round-1 content survives as `aba7fcda` + `6a6bef0f` and the new work is exactly [`3d7db159`](https://github.com/bobmatnyc/trusty-tools/commit/3d7db159dab013a39f52c0d0752dcc5f28a6a788) + [`eb99afd5`](https://github.com/bobmatnyc/trusty-tools/commit/eb99afd52e01e8448f4819ca1e0b8c333c695c1f). Reviewed `git diff 6a6bef0f..eb99afd5`. Independent worktree, no CI re-run.

Zero CRITICAL, zero HIGH. The client half is sound and fails closed on every branch I could construct. Three sub-HIGH findings below.

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | [`crates/trusty-common/src/uds/dir.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/dir.rs#L63-L87) | 63–87 | `classify_existing_dir` rejects symlink and foreign owner but never asserts the path is a directory. A regular file owned by this uid at the socket-dir path takes `mkdir` → `EEXIST` → `is_symlink()` false → owner matches → `0o644 != 0o700` → `Narrow` → `prepare_socket_dir` **chmods that file to `0700`** and returns `Ok`, and only then does `bind` fail `ENOTDIR`. Fails closed, but after mangling the mode of a file the daemon did not create, and with a diagnostic that names neither cause. Round 1 called for this in the same edit as finding 1 (question 2); it did not land. | Add a fourth arm, after the symlink check and before the owner check: take `is_dir: bool` from the same `symlink_metadata` and return a new `UdsSecurityError::NotADirectory { path }` when false. One extra argument, one extra `classify_existing_dir_rejects_a_non_directory` test. | Fix here |
| LOW | [`crates/trusty-common/src/uds/mod.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/mod.rs#L382-L386) | 382–386 | `verify_socket_for_connect` maps a failed `symlink_metadata(parent)` to `UdsSecurityError::CreateDir`, which renders as `create socket directory /…: No such file or directory` — from a dialer that creates nothing. An operator reading that log looks for a permissions problem in the daemon's startup path. | Map it to `UntrustedSocket { reason: format!("cannot stat containing directory {}: {source}", parent.display()) }`, or add a `StatDir` variant. | Fix here |
| LOW | [`crates/trusty-memory/ui/package-lock.json`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-memory/ui/package-lock.json) | — | 39 deletions stripping the `libc: ["glibc"]` / `["musl"]` fields from rollup's optional platform packages. Unrelated to UDS permissions — an incidental rewrite from a local npm run under an older npm during `build.rs`. It weakens optional-dep filtering on musl hosts and adds unreviewed churn to a security PR. | `git checkout origin/main -- crates/trusty-memory/ui/package-lock.json`. | Fix here |

---

## The five points

**1. `connect_hardened` — what it verifies, and can it be defeated?**

`verify_socket_for_connect` ([mod.rs:374–431](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/mod.rs#L374-L431)) checks six things in order: parent is not a symlink → parent owned by self → parent mode exactly `0700` → socket is not a symlink → socket is a socket → socket owned by self → socket mode exactly `0600`. Every stat is `symlink_metadata`, so nothing follows a link. `check_sun_path_budget` runs first.

Not defeatable by the attacks the PR is about. To get a client onto a hostile listener the attacker needs a socket whose parent is a non-symlink `0700` directory owned by the victim — a directory only the victim and root can write into. Attacker-owned parent → `ForeignDirOwner`; symlinked parent → `SymlinkDir`; wide parent → `UntrustedSocket`; planted regular file → `not a socket`. A socket whose parent passes but whose own mode does not is exactly the `connect_hardened_refuses_a_world_readable_socket` case and is refused.

Fail-closed on every error path, including stat errors: `symlink_metadata` on either path maps to an `Err` (`CreateDir` / `Connect`), never to a default or a skip. `socket_parent` returns `NoParent` for a bare filename. Both call sites propagate — [`bm25_client.rs:356`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/bm25_client.rs#L356) via `?`, [`embedder_client/uds.rs:185`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/embedder_client/uds.rs#L185) via `map_err(...)?`.

TOCTOU: real but not the practical attack. Between the two `lstat`s and `UnixStream::connect` the socket could be swapped — by someone with write access to the `0700` parent, i.e. the victim or root. Reaching it from outside needs the ability to rename or unlink the victim's directory entry in `/tmp`, which the sticky bit forbids. mod.rs:25–33 states this correctly.

**Other `UnixStream::connect` sites still bypassing it** — I checked all of them:

| Site | Verdict |
|---|---|
| [`bm25_supervisor.rs:761`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-memory/src/bm25_supervisor.rs#L761) `probe_socket` | Round-1 finding 3's fix text named it; still bare. Harmless in the end — a hostile listener wins the probe, the supervisor declines to spawn, and then `Bm25Client` refuses at `connect_hardened`. The failure mode is a loud unavailable, not a leak. |
| [`bus/mod.rs:120`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-agents/src/bus/mod.rs#L120) `send_to`, `:179` `list_running` | Bare connect that then **writes payload**. Path is `~/.trusty-agents/sockets/`, not attacker-writable under the threat model, and both lines are unchanged by this PR. |
| [`ctrl/socket.rs:164`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-agents/src/ctrl/socket.rs#L164) `CtrlSocket::probe` | Bare, but `bind_singleton` now calls `prepare_socket_dir` on the parent *before* probing (line 218–222), so the directory is verified on the path that matters. |
| Remaining hits | Test fixtures and one `#[cfg(test)]` helper. |

None of these is a HIGH: each is either a probe whose result is re-verified downstream, or a home-directory path this PR did not touch. Worth folding into #5089 step 1 when the shared transport lands.

**2. Symlink rejection in `classify_existing_dir` — ordering.**

Correct. `prepare_socket_dir` [dir.rs:145–155](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/dir.rs#L145-L155) takes exactly one `symlink_metadata` and feeds `is_symlink`, `uid`, and `mode` from that single `lstat` into the pure function — no re-stat between decide and use, and no `metadata()` call anywhere in the module. `classify_existing_dir` returns `SymlinkDir` before touching owner or mode, and `classify_existing_dir_checks_symlink_before_owner` pins that with an attacker-owned symlink asserting `SymlinkDir` rather than `ForeignDirOwner`. `set_permissions` (which follows) runs only on the `Narrow` arm, which a symlink can no longer reach. `prepare_socket_dir_rejects_a_symlink` asserts the target is still `0777` afterwards — the exact bit round 1 demonstrated on this host.

Residual characterization is accurate, not understated. dir.rs:109–114 names the swap window (`lstat` → `chmod`), names why it is impractical (`/tmp` sticky bit), and names what would close it (`openat`/`fchmod` on a directory fd). mod.rs:25–33 makes the same claim conditional on the caller's path rather than absolute, and says root is not defended against. That is the honest scope.

**3. `check_sun_path_budget`.**

Fires before any bind: [`bind_hardened`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/mod.rs#L316-L318) calls it as the first statement, ahead of `prepare_socket_dir` — so an over-long path does not even create a directory. `connect_hardened` does the same.

Arithmetic matches both platforms. `sun_path_capacity()` = `size_of::<sockaddr_un>() - offset_of!(sockaddr_un, sun_path)` → macOS `106 - 2 = 104`, Linux `110 - 2 = 108`. The reject condition is `len >= capacity`, leaving the NUL byte, which is byte-identical to what std's own `sockaddr_un` conversion enforces — and matches round 1's measured macOS boundary (103 binds, 104 fails).

Cannot panic on an unusual path: `as_os_str().as_encoded_bytes()` is infallible on any `OsStr`, valid UTF-8 or not. No `to_str().unwrap()`, no `to_string_lossy` length. `sun_path_capacity_is_platform_plausible` guards the derivation against a layout surprise.

**4. Pure functions vs. the real call sites.**

No drift; both are on the live path.

- `peer_uid_verdict` — [`ensure_peer_is_self` is `peer_uid_verdict(peer_uid(stream)?, self_uid())`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-common/src/uds/peer.rs#L157-L159), one line, no second comparison anywhere. All four accept loops call `ensure_peer_is_self`, including the newly-added [`ctrl/socket_listener.rs:44-48`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-agents/src/ctrl/socket_listener.rs#L44-L48), which `continue`s on refusal without spawning a handler.
- `classify_existing_dir` — sole caller is `prepare_socket_dir`, and it is the only place a pre-existing directory is judged. The doc comment states the `lstat`-not-`stat` precondition, and the caller honors it.

The two deleted silent-skip tests are replaced by assertions that actually run: `peer_uid_verdict_refuses_a_foreign_uid`, `peer_uid_verdict_refuses_root_when_we_are_not_root`, `classify_existing_dir_rejects_a_foreign_owner`. The syscall wiring keeps its own live coverage (`peer_uid_of_self_connection_is_self`). Net: refusal policy went from `#[ignore]`d-and-never-run to asserted unprivileged, and nothing is `#[ignore]`d now.

**5. The five harnesses.**

All five bind through `bind_hardened`; none was weakened. [`bm25_daemon.rs:140/163/184`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-bm25-daemon/tests/bm25_daemon.rs#L140), [`concurrent_embed.rs:305/364`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-embedderd/tests/concurrent_embed.rs#L305), [`bm25_backfill_tests.rs:119/377`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-memory/src/bm25_backfill_tests.rs#L119), [`bm25_supervisor_tests.rs:118`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-memory/src/bm25_supervisor_tests.rs#L118). Grepping the round-2 diff for added `#[ignore]`, `#[allow(`, `#[cfg(`, `--exclude`, or a loosened `set_permissions` on a socket returns nothing outside the tests that *deliberately* widen a mode to prove the refusal (`0o755` dir, `0o777` socket, `0o600` planted file). The one remaining bare `UnixListener::bind` in `bm25_supervisor_tests.rs:176` is correct — it is the fixture for `probe_returns_true_for_bound_socket`, which tests the bare-connect probe in its own tempdir.

Local re-run:
```
running 30 tests
test result: ok. 30 passed; 0 failed; 0 ignored; 0 measured; 296 filtered out; finished in 0.01s
EXIT=0
```
```
test-pointers: resolved 22385 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
PTR_EXIT=0
```
(the gate that was red at round 1 is green — and `--features uds` is now in the `lib.rs` `Test:` pointer, so the 0-tests-exit-0 trap is gone.)

---

## The semver exclusions

Justified and narrow. `trusty-embedderd`'s [`cuda = ["trusty-common/embedder-cuda"]`](https://github.com/bobmatnyc/trusty-tools/blob/fix/uds-socket-permissions/crates/trusty-embedderd/Cargo.toml#L119) and `coreml = ["trusty-common/embedder-coreml"]` are pure pass-throughs to the two features already excluded on the line above them, for the identical reason, and the gate runs on `ubuntu-latest`. The file's header demands each row "say why the feature cannot be built on a stock CI runner" and rejects "it is slow" — both rows meet that. Nothing that can build was excluded: these fail at `build.rs`, not at rustdoc.

## ADR-0034 (round-1 finding 7)

Resolved by the rebase rather than by an edit, which is the right outcome. [ADR-0034 §3 "The trust boundary"](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md) is now on `main` at line 189, and line 210 carries the sentence `peer.rs:146-149` quotes, verbatim. The citation is accurate.

## The 29-minute `trusty-memory` hang

Nothing in this diff plausibly causes it.

- No lock is taken anywhere in the new code. `verify_socket_for_connect` does two `lstat`s and holds nothing.
- No new unbounded wait. `connect_hardened` has no timeout, but `UnixStream::connect` had none before it either — the change is a strict addition of two stats in front of an unchanged call.
- Every `trusty-memory` test the PR touched has a bounded deadline: `backfill_gives_up_on_a_socket_that_never_answers` is governed by `OP_TIMEOUT`, `kill_daemon_for` polls 200×10ms, `wait_for_socket` by `SPAWN_PROBE_TIMEOUT`.
- The direction of the change is toward failing *faster*: a client that would previously have connected to a wrong-mode socket and waited on a read now returns an error from an `lstat`.

The only theoretical stall the new code adds is a blocking `lstat` on a runtime thread against a wedged filesystem, and these paths are local tmpfs. A 32.75s clean re-run against the same tree is the stronger evidence. Unexplained, and I would not attribute it here.

---

**APPROVE.** No CRITICAL, no HIGH. Three sub-HIGH findings, all `Fix here`; the MEDIUM is a four-line change plus one test.

PR #5157, opened 2026-08-07
fix(trusty-channels): create channel canvases via conversations.canvases.create

### LINKED ISSUE #5155: slack-mcp: canvases created for a channel are bot-owned and land view-only for the recipient
Every canvas delivered through `slack_canvas_create` arrives view-only for the human recipient. They cannot edit in place and have to duplicate it. Observed at least four times in a week: `F0BLZQV32UT`, `F0BMCAH10R0`, `F0BME8W4V09`, `F0BPNQQN3L0`.

### Root cause

Slack exposes two canvas-creation endpoints with different access semantics:

- `canvases.create` creates a **standalone** canvas "owned by the acting user". The acting identity is whoever's token calls it, so a bot-token call makes the bot the owner. Non-owners get no edit rights without an explicit, separate share step.
- `conversations.canvases.create` creates a **channel** canvas. Per Slack's docs, "Access is tied to channel access... there are no access implications nor is it necessary to share a channel canvas to grant access."

`trusty-channels` only ever calls the first one.

- `crates/trusty-channels/src/slack/handlers/canvas.rs:107-109` — `post_create_canvas` calls `client.call_method(CANVASES_CREATE, &body)`. Both `create_canvas` (:122) and `canvas_create` (:156) route through it.
- `crates/trusty-channels/src/slack/client.rs:166-172` — `call_method` always uses the bot token. A user-token path `call_method_user` exists at :187-196 but canvas creation never uses it.
- `crates/trusty-channels/src/slack/handlers/mod.rs:103` — only `CANVASES_CREATE = "canvases.create"` is defined. `conversations.canvases.create` appears nowhere in the crate.

### Compounding detail

Callers pass `channel_id` to `slack_canvas_create` expecting channel-bound behaviour. That parameter on `canvases.create` only tabs the canvas into the channel cosmetically. It does not confer channel-membership-based edit rights. The result looks channel-integrated and grants nothing, which is why this went misdiagnosed for weeks.

### Misdiagnosis worth recording

This was previously attributed to a `canvases.access.set` limitation. `canvases.access.set` is not implemented in this crate at all (repo-wide grep, zero hits), so no call through this server ever exercised it. That explanation should be considered void.

### Proposed fix

1. Add `CONVERSATIONS_CANVASES_CREATE = "conversations.canvases.create"` to `handlers/mod.rs`.
2. In `canvas.rs`, branch on `channel_id`: when present, call the channel endpoint; when absent, keep `canvases.create` for the standalone case.
3. Same `canvases:write` scope, already granted. No new OAuth grant, no user token, no action required from the workspace owner.

### Secondary, optional

Implement `canvases.access.set` as a tool so the four already-broken canvases can be retrofitted rather than recreated. Note Slack's prerequisite that a canvas must already be shared with a user before that user can be promoted via `user_ids`, so this may need a `channel_ids` share first.

### Acceptance

A canvas created via `slack_canvas_create` with a `channel_id` is editable in place by a member of that channel, verified by that member.

bob-duetto:
## The defect

`slack_canvas_create` and `slack_create_canvas` both route through
`post_create_canvas`, which only ever called `canvases.create`. That endpoint
creates a **standalone** canvas owned by the acting identity — with a bot token,
the bot owns it, and non-owners get no edit rights without a separate explicit
share. The human who asked for the document received it view-only and had to
duplicate it to edit. Four occurrences in a week.

Callers were already passing `channel_id`, expecting channel-bound behaviour.
That is why this looked like it should already work: on `canvases.create`,
`channel_id` only tabs the canvas into the channel cosmetically. It confers no
edit rights.

`conversations.canvases.create` creates a **channel** canvas. Per Slack's docs,
"access is tied to channel access... there are no access implications nor is it
necessary to share a channel canvas to grant access."

## The fix

`post_create_canvas` now picks the endpoint by whether `channel_id` was supplied:

- **present** → `conversations.canvases.create` with `channel_id`, `title`, and
  `document_content` (Slack's documented request shape for that method)
- **absent** → `canvases.create`, unchanged

Same bot token, same `canvases:write` scope. No new OAuth grant, and
`call_method_user` is untouched.

Both endpoints return the created id under the same top-level `canvas_id`, and
`post_create_canvas` is the single place that reads it, so the MCP response is
byte-identical either way — `both_create_paths_return_the_same_structure` pins
that. A caller never has to know which endpoint ran.

Canvas-create failures now carry the attempted method: `<slug> (from <method>)`,
slug still leading. The two endpoints fail for different reasons
(`free_teams_cannot_create_non_tabbed_canvases` only from `canvases.create`;
`channel_canvas_already_exists` and `team_tier_cannot_create_channel_canvases`
only from the channel one), so a bare slug meant reading the source to work out
which call ran.

`channel_id` stays optional in both schemas; tool signatures are unchanged. The
free-tier failure mode is no worse — `free_teams_cannot_create_non_tabbed_canvases`
is a `canvases.create` rejection, and the channel path avoids it entirely.

## Behaviour change worth knowing at merge time

A channel has at most **one** channel canvas. A second create for the same
channel returns `channel_canvas_already_exists` rather than minting another
canvas, which is what the old code did. Both tool descriptions now say so and
point the caller at editing the existing canvas instead. This is inherent to
channel canvases, not something the implementation chose.

## Tests

Five new/updated cases in `crates/trusty-channels/tests/tools_http.rs`, each
mounting the endpoint it expects and an `.expect(0)` mock on the other:

- `create_canvas_with_channel_and_markdown` (updated) — `channel_id` routes to
  the channel endpoint
- `create_canvas_without_channel_uses_standalone_endpoint` (new)
- `canvas_create_posts_document_content_and_channel` (updated) — same for the
  connector-parity clone
- `canvas_create_without_channel_uses_standalone_endpoint` (new)
- `both_create_paths_return_the_same_structure` (new) — identical MCP response
  from both paths
- `canvas_create_channel_endpoint_error_names_the_endpoint` (new)
- `canvas_create_surfaces_slack_api_error` (updated) — asserts the new message

Verified red before green: with the routing reverted to `CANVASES_CREATE`, the
four channel-path tests fail (`test result: FAILED. 0 passed; 4 failed`).

## Gates

```
cargo fmt -p trusty-channels --check    → clean
cargo clippy -p trusty-channels --all-targets -- -D warnings → clean
cargo check -p trusty-channels --all-targets → EXIT=0
cargo test -p trusty-channels           → 110 + 9 + 15 + 12 + 54 passed; 0 failed; 0 ignored
bash scripts/check_line_cap.sh          → 0 violations
```

## Rebuild needed for the local user

The installed `slack-mcp` binary on the reporting machine is pinned to commit
`b45f2adc` (2026-07-26), well behind main. This fix does not take effect there
until `slack-mcp` is reinstalled.

Closes #5155


bob-duetto:
## Repeat delivery to a conversation that already has a canvas

Worth settling before merge. Nearly every document this server delivers goes to
the same DM with the same person. Post-fix, the first delivery creates that
conversation's channel canvas and every delivery after it fails with
`channel_canvas_already_exists`. That surfaces on the second document, not in
some edge case, and it is a genuine change from the old behaviour of minting a
new canvas per call.

**1. Should a repeat delivery replace the existing canvas content in place?**

No — not silently, and not from inside a tool named `create`. The canvas is now
shared and editable, which is the entire point of this fix, so its current
content may not be what this server last wrote. Overwriting it on a create call
would destroy a human's edits. `slack_canvas_push` already owns the update path
with an explicit `mode` (`append` / `replace_all`), and `slack_update_canvas`
owns the whole-document replace. The create tool should route the caller to
those, not decide for them.

**2. Should the caller get an error naming the existing `canvas_id`?**

Yes. This is the piece I would build. `conversations.info` returns the existing
channel canvas at `properties.canvas.file_id`
([conversation type](https://api.slack.com/types/conversations)), so on
`channel_canvas_already_exists` the handler can make one extra call and return
something actionable:

```
channel_canvas_already_exists (from conversations.canvases.create):
channel C123 already has canvas F456 — update it with slack_canvas_push
(mode "replace_all") or slack_update_canvas
```

One extra API call, only on the failure path, and it turns a dead end into a
next step without anyone reading source. `conversations.info` needs its own read
scope for the conversation type involved (`channels:read` / `groups:read` /
`im:read` / `mpim:read`) — if that lookup fails the error must degrade to the
bare slug rather than masking the original failure.

**3. Is there a case for a deliberate standalone canvas?**

Yes, and it already exists: omit `channel_id`. This PR left that path untouched
and both tool descriptions now say what it gives you. What must not exist is a
*fallback* — catching `channel_canvas_already_exists` and quietly retrying
against `canvases.create` would restore the bot-owned view-only canvas #5155 is
about, and would do it on the second and every later delivery, which is most of
them. Standalone stays something the caller asks for by leaving `channel_id`
off.

**What I would build:** option 2 only. Error enrichment on the collision path,
naming the existing canvas id and pointing at the update tools. Create keeps
meaning create.

**What I could not verify:** whether `conversations.canvases.create` accepts an
IM/DM channel id at all. Its method docs say only "Channel ID where the canvas
will be created", and the canvases surface page does not address DMs either way.
Since nearly all traffic here is DM, this needs checking against the live
workspace. If DMs are not supported, this fix helps channels and changes nothing
for the dominant delivery path, and the routing needs a DM carve-out.

This deserves its own issue covering both the error enrichment and the DM
verification. I have not filed one.

Issue #5111, opened 2026-08-07
flake: execute_doctor_against_test_daemon fails 'daemon unreachable' reproducibly on this machine


bobmatnyc:
Defect

`client::executor::tests::execute_doctor_against_test_daemon` fails against a
locally spawned test daemon. `CLAUDE.md` already names this test as a known
environmental timing flake, but no issue tracked it.

Evidence

Same commit, same worktree, 40 minutes apart:

```
# earlier full run
test result: ok. 4740 passed; 0 failed; 7 ignored

# later, isolated, 3 consecutive runs
test result: FAILED. 0 passed; 1 failed; 0 ignored; 4746 filtered out
test result: FAILED. 0 passed; 1 failed; 0 ignored; 4746 filtered out
test result: FAILED. 0 passed; 1 failed; 0 ignored; 4746 filtered out
```

```
panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18:
expected Doctor, got Error("daemon unreachable: error sending request for url
(http://127.0.0.1:56384/api/v1/doctor?project=...)")
```

Not caused by the branch that surfaced it (feat/instruction-trim-to-target
touches markdown assets plus one unrelated test module). The failure is an HTTP
connect to a spawned test daemon, so it is a startup-race or port-bind timing
issue, not a doctor-logic one.

Resolution

Make the test wait on a readiness condition instead of racing the daemon
startup, or mark the dependency explicitly. Not fixed here — filed so the next
red run appends to this rather than opening a second issue.

bobmatnyc:
Occurrence during #5623 / PR #5632 baseline verification on this machine, confirming it is still reproducible independent of any branch.

Reproduced on `origin/main` (`9bd0fbc5f`) with all branch changes stashed:

```
cargo test -p trusty-mpm --lib -- --exact client::executor::tests::execute_doctor_against_test_daemon
thread 'client::executor::tests::execute_doctor_against_test_daemon' panicked at crates/trusty-mpm/src/client/executor/tests.rs:215:18:
expected Doctor, got Error("daemon unreachable: error sending request for url (http://127.0.0.1:63643/api/v1/doctor?project=...)")
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4968 filtered out; finished in 14.11s
```

[...]

PR #5169, opened 2026-08-07
fix(trusty-agents): close the remaining bare UnixStream::connect sites; correct ADR-0034's site inventory

bobmatnyc:
Step 1a of four for [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) (tm 1.3.5 criterion c). Does **not** close it.

## Defect

[PR #5124](https://github.com/bobmatnyc/trusty-tools/pull/5124) landed `trusty_common::uds` and wired the two clients ADR-0034 listed. Three dial sites survived it because ADR-0034's transport-clients table never listed them:

- `crates/trusty-agents/src/bus/mod.rs` `send_to` — **writes an NDJSON payload** to whatever answers the target path
- `crates/trusty-agents/src/bus/mod.rs` `list_running` — probes each `*.sock`
- `crates/trusty-agents/src/ctrl/socket.rs` `CtrlSocket::probe` — the CTRL singleton probe

## Resolution

All three route through `connect_hardened`, so a `0600`-and-self-owned socket in a `0700` directory is the precondition for talking to a peer rather than a documented intention.

**`probe`'s semantics change.** An existing socket that fails verification is now an error, not an open stream. Refusing alone would brick the project — nothing would replace the bad socket — so a new `is_stale_socket` classifies an unverifiable socket takeover-safe and `bind_singleton` unlinks it and binds a hardened replacement. The check is not relaxed: the untrusted socket is never written to. `is_connection_refused` keeps its exact meaning; `mode_dispatch`'s argv-forward path moves to `is_stale_socket`.

One user-visible consequence: a live pre-#5099 controller whose socket is still at the process umask is replaced rather than adopted, once, on upgrade.

`probe_error_from_uds` unwraps `Connect`/`StatForConnect` to the original `io::Error` — wrapping every variant in `io::Error::other` would erase `NotFound` for an absent socket and leave the controller unable to ever start.

No caller became async: `send_to`, `list_running`, and `probe` were already `async`.

### Refusals `connect_hardened` caused, and how each was resolved

| Site | Refusal | Resolution |
|---|---|---|
| `bus::subscribe_and_drain` (test helper) | `create_dir_all` + bare bind left the stub at the process umask; `send_to` would now refuse to dial it | Binds through `bind_hardened` |
| `ctrl::socket::probe_times_out_when_no_listener` | `tempfile::tempdir()` creates at the umask (0755), not 0700 like `mkdtemp`, so the dial refused the *directory* before looking for the socket | Test lays the dir out via `prepare_socket_dir`, mirroring `~/.trusty-agents/sockets/` |

Both are harness fixes. No bypass flag, no `#[allow]`, no weakened check.

## ADR-0034 inventory correction

Its Context table lists three UDS clients and omits `trusty-agents`' `MessageBus` entirely — on both halves (`send_to`/`list_running` dial, `MessageBus::start` binds) — and lists the ctrl socket at `<project>.sock`, which is actually the bus's path (ctrl is `<project>.ctrl.sock`). The real counts are four clients and four bind sites, not three and two. §4's "three current implementations" inherits the same error.

**Convention choice: in-place `Correction` notes, original text unedited.** This follows the precedent #5099 set in the same PR that created the module — ADR-0031 and ADR-0032 each carry a `> 🔴 **Correction (#5099, 2026-08-07).**` blockquote appended after the passage it corrects, with the original left intact and an explicit "left unedited per DOC-46 §4". Two notes here, because two passages are wrong (Context inventory, §4 count). No amending ADR: DOC-46 §4's `Amended by NNNN` status is for a later ADR that *refines a decision*. Nothing about ADR-0034's decision changes — every implementation still migrates onto the shared module; the migration is one crate wider than the ADR said. Minting an ADR to record a miscount would misuse the status lifecycle, and editing the text away would erase that the omission had a consequence: it is why these three sites survived #5124.

## Gate — Rust Test Ladder rung 4

```
cargo fmt --check                                                        EXIT=0
cargo check --workspace --all-targets (SKIP_UI_BUILD=1, 3 GUI excludes)   EXIT=0
cargo clippy -p trusty-common --all-targets --features uds -- -D warnings EXIT=0
cargo clippy -p trusty-agents --all-targets -- -D warnings                EXIT=0
cargo test --no-fail-fast -p trusty-common --features uds  — 329 passed; 0 failed; 6 ignored
cargo test --no-fail-fast -p trusty-agents — 3385 passed; 0 failed; 11 ignored (lib)
                                             + 27 passed across 9 integration targets
scripts/check_line_cap.sh            EXIT=0  (3784 files, 0 violations)
scripts/check_changelog_fragment.sh  EXIT=0
scripts/check_test_pointers.sh       EXIT=0  (22492 citations, 0 dangling)
scripts/check_sld.sh                 EXIT=0
```

No baseline reds encountered — none of #5040, #4931, #5111, or the `l2_ranks_*` sighting is in these two crates.

### Red-before-green

Four refusal tests, run against the unmigrated code:

```
test bus::tests::send_envelope_to_refuses_a_world_readable_socket ... FAILED
test ctrl::socket::tests::bind_singleton_takes_over_a_socket_that_fails_verification ... FAILED
test ctrl::socket::tests::probe_refuses_a_socket_that_fails_verification ... FAILED
test bus::tests::list_running_in_skips_a_world_readable_socket ... FAILED
test bus::tests::list_running_in_reports_a_hardened_socket ... ok
test bus::tests::send_envelope_to_delivers_over_a_hardened_socket ... ok

thread '...send_envelope_to_refuses_a_world_readable_socket' panicked at crates/trusty-agents/src/bus/mod.rs:353:14:
must refuse a socket that is not 0600: ()
thread '...probe_refuses_a_socket_that_fails_verification' panicked at crates/trusty-agents/src/ctrl/socket.rs:355:14:
probe must refuse a socket that is not 0600: UnixStream { fd: FileDesc(OwnedFd { fd: 35 }), ... }
thread '...list_running_in_skips_a_world_readable_socket' panicked at crates/trusty-agents/src/bus/mod.rs:382:9:
a 0666 socket must not be reported running: ["wide"]

test result: FAILED. 2 passed; 4 failed; 0 ignored; 0 measured; 3390 filtered out
```

After the migration:

```
test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 3376 filtered out
```

`send_to` and `list_running` were split into path-taking `send_envelope_to` / `list_running_in` so the dial halves are testable — the public entry points hardcode `~/.trusty-agents/sockets/`, so a test driving them would write into the developer's real home and be perturbed by their live buses.

## Left alone, deliberately

- `crates/trusty-memory/src/bm25_supervisor.rs` `probe_socket` (~line 761) — owned by step 2 of the same issue. Untouched.
- `crates/trusty-agents/src/debugger/socket_monitor.rs:65` — a sync `std::os::unix::net::UnixStream::connect` liveness probe for the debug TUI. Writes nothing, and `connect_hardened` is async. Out of this step's scope; it is the only bare connect left in the workspace outside step 2's file.


bobmatnyc:
## Verdict: WARN

Reviewed at `5a435fd8` in a detached worktree. One HIGH, one LOW. The security direction is right and the four refusal tests are real refusals, not weakened checks — but the takeover predicate is applied at one call site that does not own the socket it deletes.

## Findings

| HIGH | `crates/trusty-agents/src/runtime/mode_dispatch.rs` | 384-387 | The argv-forward client path unlinks a live controller's socket. `is_stale_socket` includes "containing directory is not `0700`", which says nothing about the socket itself | Revert this arm to `ctrl::is_connection_refused` | Fix here |
| LOW | `crates/trusty-agents/src/debugger/socket_monitor.rs` | 65 | Left unhardened on the stated grounds that `connect_hardened` is async — but `trusty_common::uds::verify_socket_for_connect` is `pub` and **synchronous** | Guard `check_alive` with `verify_socket_for_connect(&self.socket_path).is_ok()` | Fix here |

## Required Changes

1. `crates/trusty-agents/src/runtime/mode_dispatch.rs:384` — change `ctrl::is_stale_socket(&e)` back to `ctrl::is_connection_refused(&e)`.
2. `crates/trusty-agents/src/ctrl/socket.rs:167-177` — `is_stale_socket`'s doc enumerates "wrong mode, wrong owner, a symlink, or not a socket at all" and omits the directory-mode arm, which is the broadest one. Add it, and state the precondition the predicate carries: it is safe only for a caller that owns the path it is about to bind.

---

## 1. Is "unverifiable" a safe predicate for "unlink and take over"?

Not on its own. It is safe in `bind_singleton` and unsafe in `mode_dispatch`, and the difference is not the predicate — it is two properties `bind_singleton` has that the other call site does not.

Every way [`connect_hardened`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-common/src/uds/mod.rs#L395) can fail, and whether unlinking is correct:

| Failure | Variant | `is_stale_socket`? | Unlink correct? |
|---|---|---|---|
| socket absent | `StatForConnect` → `NotFound` | yes | yes — nothing to destroy |
| no listener | `Connect` → `ConnectionRefused` | yes | yes — confirmed dead |
| regular file / fifo / dir at the path | `UntrustedSocket` "path is not a socket" | yes | yes — cannot be a live peer |
| path is a symlink | `UntrustedSocket` "socket path is a symlink" | yes | yes — `bind_hardened` never makes one |
| socket mode ≠ `0600` | `UntrustedSocket` "socket is mode …" | yes | **only if we own the path** — a live listener can be behind it |
| socket owned by another uid | `UntrustedSocket` "socket is owned by uid …" | yes | **only if we own the path** — unlink is granted by directory write permission, not file ownership |
| **containing dir mode ≠ `0700`** | `UntrustedSocket` "containing directory … is mode …" | **yes** | **no** — condemns a healthy `0600` live socket for its neighbour directory's mode |
| containing dir owned by another uid | `ForeignDirOwner` | no | fails closed ✅ |
| containing dir is a symlink | `SymlinkDir` | no | fails closed ✅ |
| containing dir is not a directory | `NotADirectory` | no | fails closed ✅ |
| dir stat fails (`EACCES`) | `StatForConnect` → `PermissionDenied` | no | fails closed ✅ |
| path over `sun_path` budget | `PathTooLong` | no | fails closed ✅ |
| bare filename | `NoParent` | no | fails closed ✅ |

The six fail-closed rows are right, and they are the ones that matter for a hostile filesystem: a foreign-owned or symlinked directory propagates as an error and clobbers nothing.

**The row that widens the predicate is the directory mode.** It is the only reason a socket can be "unverifiable" while being a perfectly good `0600` socket with a live process behind it.

`bind_singleton` neutralises that row, and does so deliberately — [`socket.rs:285-287`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/ctrl/socket.rs#L285) calls `prepare_socket_dir(parent)?` **before** probing. By the time `probe` runs, the directory is `0700` and owned by this uid or the function has already returned an error. So inside `bind_singleton` the only reachable `UntrustedSocket` reasons are properties of the socket file itself — exactly the ones a self-healing owner should replace. That, plus the fact that `bind_singleton` is about to bind the path it unlinks, is what makes the self-heal correct there.

The attacker question: an unprivileged other-uid attacker cannot reach any of the four "only if we own the path" rows through the `0700` directory, and if they could widen the directory they would trip `ForeignDirOwner`. `chmod` on the socket requires ownership, so triggering the mode row requires same-uid — already inside the trust boundary, where unlinking the socket directly is simpler. **This is not an other-uid DoS primitive.** It is a self-inflicted one, which is finding 1.

## 2. Does the ownership distinction hold in this PR's code?

For `bind_singleton`, yes, for the two reasons above.

For [`mode_dispatch.rs:384-387`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/runtime/mode_dispatch.rs#L384), no. It is the only production `CtrlSocket::cleanup` call outside `CtrlSocket` itself, and neither property holds:

- **No `prepare_socket_dir` runs before this probe**, so the directory-mode row is live here and only here.
- **This process is a client.** It is a one-shot `tagent "do X"` forwarding argv to an incumbent controller. It does not own the socket, and it binds nothing after unlinking.

So a `~/.trusty-agents/sockets/` at any mode other than `0700` makes every argv-carrying invocation delete the incumbent controller's socket file. The incumbent keeps its listener fd on an unlinked inode and is silently orphaned: still running, still holding the user's REPL session, unreachable forever. The invocation then falls through to `bind_singleton` at [`repl/mod.rs:360`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/ctrl/repl/mod.rs#L360) and becomes a second controller.

That outcome is what the singleton design exists to prevent, in its own words at [`repl/mod.rs:351-357`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/ctrl/repl/mod.rs#L351):

> `bind_singleton` returns `AlreadyRunning` when a controller already owns the socket — in that case we do NOT start our own accept loop (which would have stolen the socket file and broken the first controller's CLI forwarding).

`mode_dispatch` now performs that steal, ahead of `bind_singleton` and outside its protection.

Before this PR the predicate here was `is_connection_refused` over a bare `UnixStream::connect`: a live controller returned `Ok(stream)` and the cleanup arm was unreachable. This is a behavioural regression introduced by the diff, and the reachable trigger — a sockets directory at the process umask — is precisely the pre-[#5124](https://github.com/bobmatnyc/trusty-tools/pull/5124) on-disk state every upgrading user has.

The fix is one identifier. The unlink at 386 is redundant anyway: `bind_singleton` re-probes moments later and does its own cleanup from the owner side. Narrowing this arm to `is_connection_refused` loses nothing and restores "the kernel said nobody is home" as the only licence to delete another process's socket.

This is consistent with the [#5170](https://github.com/bobmatnyc/trusty-tools/pull/5170) ruling rather than in tension with it — that review drew the line at ownership, and this is the one place in this PR that crosses it.

## 3. `is_connection_refused` and `probe_error_from_uds`

Both claims verified.

`is_connection_refused` is byte-for-byte unchanged at [`socket.rs:121-126`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/ctrl/socket.rs#L121), and `is_connection_refused_classifies_enotsock` is unchanged and passing. Worth naming, though: after this diff the narrow predicate has **no production caller left** — `mode_dispatch` was its only one. Keeping the function exact while moving its sole user to the wider predicate is the mechanism of finding 1, and the required change gives it its caller back.

`probe_error_from_uds` is correct and load-bearing. Traced the absent-socket path: `verify_socket_for_connect` reaches `symlink_metadata(path)`, gets `ENOENT`, and returns `StatForConnect { source: NotFound }`. Unwrapped, `kind()` is `NotFound`, `is_connection_refused` is true, `bind_singleton` takes over. Wrapped in `io::Error::other` it would be `Other` — neither connection-refused nor `UntrustedSocket` — and would fall to the conservative `Err(e) => Err(e)` arm, so the controller could never start on a clean machine. The claim holds exactly.

The errnos that still round-trip after this change: `NotFound` (absent socket or absent directory), `ConnectionRefused` (dead listener), `PermissionDenied` (dir stat denied → fails closed, correct). No errno stops round-tripping. `ENOTSOCK` no longer arrives from the kernel because verification catches non-sockets first, but it lands in the same takeover branch via `UntrustedSocket`, so control flow is preserved. The wildcard arm keeps the match valid against `#[non_exhaustive]`.

## 4. The two harness refusals

No bypass. Grepped the full diff for `#[allow]`, `#[ignore]`, `cfg(not(`, mode literals and skip markers. The only permission literals introduced are `0o666` in three test setups (widening a live socket to *prove* refusal) and `0o600`/`0o700` in assertions.

- `bus::subscribe_and_drain` ([`bus/mod.rs:158`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/bus/mod.rs#L158)) swaps `create_dir_all` + bare `UnixListener::bind` for `bind_hardened`. The stub now binds the way the real bus does.
- `probe_times_out_when_no_listener` lays the directory out with `prepare_socket_dir` instead of relying on `tempfile`'s umask. Its assertion widened from `is_connection_refused` to `is_stale_socket`, which is correct for a test whose socket genuinely does not exist.

Both are fixes at the harness. Neither weakens the check under test. Test count went up by six.

## 5. The ADR correction

The convention call is right. Nothing in ADR-0034's Decision changes — the shared module still exists and every implementation still migrates onto it — so an in-place `> 🔴 Correction` blockquote with the original text unedited is what DOC-46 §4 asks for, and it matches the precedent [#5099](https://github.com/bobmatnyc/trusty-tools/issues/5099) set. An amending ADR would imply a decision reversal that did not happen.

Every factual claim in the correction checks out:

- **Four UDS clients, not three.** `embedder_client/uds.rs`, `bm25_client.rs`, `ctrl/socket.rs`, `bus/mod.rs`.
- **Four bind sites, not two.** The sentence being corrected names only `trusty-embedderd/src/uds_server.rs` and `trusty-bm25-daemon/src/socket.rs`; the bus and `CtrlSocket::bind` are the other two. This also agrees with `uds/mod.rs`'s own "four bind sites".
- **The ctrl-row path is wrong, and the author's reading of it is right.** The table row for `CtrlSocket` lists `~/.trusty-agents/sockets/<project>.sock`, but [`ctrl_socket_path`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-agents/src/ctrl/socket.rs#L94) builds `{project_id}.ctrl.sock`. `<project>.sock` is the bus's path. The row that omitted the bus was describing it — which is a neat demonstration that the omission was a real gap and not a bookkeeping slip.

## 6. What was left: `socket_monitor.rs:65`

Leaving it does not create an exposure. `check_alive` writes nothing, reads nothing, and returns a bool to the debug TUI. A planted socket can only make the TUI say "alive". No payload crosses the boundary, so it is genuinely out of scope for a PR about dial-site hardening.

But **the stated reason for leaving it is wrong**, which is the LOW finding. `connect_hardened` being async is not the blocker: [`verify_socket_for_connect`](https://github.com/bobmatnyc/trusty-tools/blob/5a435fd8/crates/trusty-common/src/uds/mod.rs#L411) is `pub` and synchronous — it is the filesystem half split out precisely so it can be called without standing up a listener. The one-line guard is available today:

```rust
pub fn check_alive(&self) -> bool {
    if trusty_common::uds::verify_socket_for_connect(&self.socket_path).is_err() {
        return false;
    }
    StdUnixStream::connect(&self.socket_path).is_ok()
}
```

Without it the TUI reports "alive" for exactly the sockets `probe` refuses, which is the confusing state during the upgrade window that finding 1 also lands in.

## Gates I re-ran myself

```
test-pointers: resolved 22492 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
EXIT=0
```

Clean. The PR body's claim holds — this is not a repeat of #5170's false green.

```
$ cargo test -p trusty-agents --lib ctrl::
test result: FAILED. 231 passed; 1 failed; 0 ignored; 0 measured; 3164 filtered out

$ cargo test -p trusty-agents --lib bus::
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 3392 filtered out
```

All sixteen `ctrl::socket::tests` pass, including the four new refusal tests:

```
test ctrl::socket::tests::probe_refuses_a_socket_that_fails_verification ... ok
test ctrl::socket::tests::bind_singleton_takes_over_a_socket_that_fails_verification ... ok
test ctrl::socket::tests::probe_succeeds_against_live_listener ... ok
test ctrl::socket::tests::probe_times_out_when_no_listener ... ok
```

**The one failure is pre-existing and environmental, not caused by this PR:**

```
---- ctrl::tests::config_tests::resolve_agent_config_falls_back_to_project_ctrl_toml stdout ----
panicked at crates/trusty-agents/src/ctrl/tests/config_tests.rs:253:5:
assertion failed: matches!(cfg.agent.role.as_str(), "controller" | "ctrl")
```

It is an agent-config resolution test that reads the real `$HOME` without taking `crate::test_env::HOME_LOCK` (the very next test in the file takes it and says why). `git diff --name-only 2a2b6cfb..HEAD` touches six files, none in the config-resolution path, and this PR's CI is green.

---

bobmatnyc:
## Review round 1 — both findings addressed at `ee85953b`

**HIGH, `mode_dispatch.rs:384`** — narrowed to `ctrl::is_connection_refused`, as required. I traced the mechanism independently and it holds: `prepare_socket_dir` runs only in `bind_singleton`, so the directory-mode arm of `UntrustedSocket` is reachable at the client site and only there, and it is the one arm that says nothing about the socket. A `~/.trusty-agents/sockets/` at the process umask made every argv-carrying invocation unlink a live incumbent.

Two things beyond the literal identifier change, both inside the finding's own logic:

1. **`is_stale_socket` → `is_stale_socket_for_owner`.** Required change 2 asked me to state the precondition in the doc. I did that — the doc now enumerates the directory-mode arm and both properties `bind_singleton` has — and also put the precondition in the name. The symbol is new in this unreleased PR, so the rename costs nothing, and it converts the misuse from a silent runtime hazard into a call site that reads wrong. Say the word if you would rather have the plain name back.
2. **A `// #5089 review:` pointer at the client site** recording why the owner predicate is wrong there and that `bind_singleton` re-probes moments later.

`is_connection_refused` has its production caller back, which is the resolution of the zero-callers tell you flagged.

**Regression test** — `a_live_controller_in_a_wide_directory_is_not_client_clobberable`. Live controller bound hardened, directory regressed to 0755, socket untouched. Asserts the probe fails, `is_connection_refused` is **false** (a client may not unlink), `is_stale_socket_for_owner` is **true** (which is precisely why a client must not use it), the socket file still exists, and — repairing the directory — the same listener still answers, proving the incumbent was healthy all along.

Stated plainly: this test pins the predicate contract the call site now depends on; it does not go red against the identifier itself, because `mode_dispatch` is not reachable from a unit test. The rename is what guards the call site — the old name no longer compiles.

**LOW, `socket_monitor.rs:65`** — guard added, your exact one-liner. You are right that my stated reason was wrong: `verify_socket_for_connect` is `pub` and synchronous. New test `socket_monitor_dead_when_socket_fails_verification` asserts alive-then-dead across a `0666` chmod on a live listener. I did not capture a red for this one before adding the guard — the second assertion would have failed, since a bare connect to a live `0666` socket succeeds, but I am not claiming output I did not observe.

## Gates re-run at `ee85953b`

```
cargo fmt --check                                                         EXIT=0
cargo check --workspace --all-targets (SKIP_UI_BUILD=1, 3 GUI excludes)   EXIT=0
cargo clippy -p trusty-common --all-targets --features uds -- -D warnings EXIT=0
cargo clippy -p trusty-agents --all-targets -- -D warnings                EXIT=0
cargo test --no-fail-fast -p trusty-common --features uds
  test result: ok. 329 passed; 0 failed; 6 ignored
cargo test --no-fail-fast -p trusty-agents
  test result: ok. 3387 passed; 0 failed; 11 ignored (lib)
                  + 27 passed across 9 integration targets
scripts/check_line_cap.sh            EXIT=0  (3784 files, 0 violations)
scripts/check_changelog_fragment.sh  EXIT=0
scripts/check_test_pointers.sh       EXIT=0  (22494 citations, 0 dangling)
scripts/check_sld.sh                 EXIT=0
```

Lib count 3385 → 3387: the two new tests. No baseline reds.

PR #5170, opened 2026-08-07
feat(trusty-common): promote the BM25 supervisor to a generic UdsServiceSupervisor (#5089 step 2)

### LINKED ISSUE #5089: Implement ADR-0034: console webhook ingress relayed over UDS to a supervised on-demand process (1.3.5 criterion c)

bobmatnyc:
Step 2 of four for #5089 (ADR-0034 §1, milestone `tm 1.3.5` criterion (c)). Does **not** close #5089.

## What changed

`trusty-memory`'s `Bm25Supervisor` is promoted into `trusty-common` as
`uds::supervisor::UdsServiceSupervisor`, behind a new `uds-supervisor` feature,
so `trusty-console` can spawn `trusty-review` / `trusty-analyze` on demand for
webhook delivery. `Bm25Supervisor` becomes a thin face over it — same
constructors, same `ensure_running(palace, data_dir)`, same counters, no call
site changed.

New public surface in `trusty-common`: `UdsServiceSupervisor`, `SupervisorConfig`,
`ServiceTimeouts`, `SpawnSpec`, `SupervisorError`, `SocketVerdict`,
`probe_socket_verdict`, `socket_is_serving`, `over_rss_limit`.
Every new enum and struct is `#[non_exhaustive]`.

Built on #5124's `uds/` primitives rather than reimplementing: adoption verifies
through `verify_socket_for_connect`, and the spawn path pre-checks
`check_sun_path_budget`.

## The two invariants that had to survive

**1. Socket-backed liveness, not `try_wait()`.** Carried verbatim. `try_wait()`
answers "has this child been reaped", not "is it serving" — a SIGKILLed daemon
closes its listener during `do_exit` and only becomes reapable once the kernel
finishes tearing it down. `SocketVerdict` stays three-state and only ENOENT /
ECONNREFUSED mean `NotServing`; a timeout or any other errno is `Inconclusive`
and leaves the child alone. The `doomed` queue is still drained **under the spawn
gate**, because the doomed child unlinks the shared socket path as the last step
of its own shutdown and a concurrent termination could land that unlink after the
replacement bound the same path.

The macOS/BSD saturated-backlog fact (a full accept queue answers ECONNREFUSED,
so it is indistinguishable from a dead listener) is now written as an explicit
**requirement on anything supervised**: accept promptly and off the request path,
or eviction under load is a live risk. It is stated on `SocketVerdict` where the
next service author will read it, not left as a BM25 footnote.

**2. The timeouts are per-service.** `SPAWN_PROBE_TIMEOUT` (3 s, justified by
BM25 having no model to load, against 30 s for the embedder) and
`SIGTERM_PATIENCE` (5 s, bound to the daemon's own `SHUTDOWN_FLUSH_TIMEOUT`) are
now fields of a `ServiceTimeouts` value the service supplies.

**What replaced the compile-time assert.** `const _: () = assert!(SIGTERM_PATIENCE_SECS
> trusty_bm25_daemon::SHUTDOWN_FLUSH_TIMEOUT.as_secs(), …)` could not cross the
abstraction — `trusty-common` cannot name any particular daemon's flush budget
without depending on it. So the check moved into `ServiceTimeouts::new`, which is
a **`const fn`** whose `assert!` requires `sigterm_patience > shutdown_flush`,
and the service now declares its child's flush budget as an argument.
`trusty-memory` writes `const BM25_TIMEOUTS: ServiceTimeouts = ServiceTimeouts::new(…)`,
so the assert fires during const evaluation exactly as before. It is strictly
stronger than what it replaces: it is bound to the value actually handed to the
supervisor rather than to a free-standing constant a refactor could leave behind.
Lower `SIGTERM_PATIENCE` to 2 s and the build fails. A service that builds
timeouts at runtime instead gets a panic at construction — startup, before any
child exists.

Coverage: `service_timeouts_reject_patience_equal_to_the_flush`,
`service_timeouts_reject_patience_below_the_flush`,
`service_timeouts_accept_a_subsecond_margin` (the const comparison is
hand-written, since `Duration`'s operators are not `const`), plus
`sigterm_patience_exceeds_the_daemon_flush_budget` which now asserts
`BM25_TIMEOUTS.shutdown_flush == trusty_bm25_daemon::SHUTDOWN_FLUSH_TIMEOUT` —
so a supervisor configured with the *wrong* budget cannot pass a guard that
checks the right relationship between wrong numbers.

## One behaviour change

Adoption is verified rather than assumed. A socket that answers but fails
`verify_socket_for_connect` now produces `SupervisorError::UntrustedSocket`
instead of being adopted silently or falling through to a spawn that dies on
EADDRINUSE. This is the one path where the target's own `bind_hardened` never
runs, which #5099's dialer docs already named as the reason a dialer must
verify, and ADR-0034 §3 makes the socket's permissions the credential the relay
hop rests on. Post-#5099 daemons bind through `bind_hardened`, so this is inert
for them.

## Everything else preserved, verified

| Behaviour | Where it is proved |
|---|---|
| Spawn-gate serialisation vs. double-spawn | `concurrent_callers_for_one_palace_spawn_exactly_one_daemon` (asserts launches, not map size) |
| Cap not violated in aggregate under fan-out | `a_concurrent_fanout_never_exceeds_the_cap` |
| Adoption of a socket another process bound | `already_running_skips_spawn`, `adoption_accepts_a_hardened_socket_without_spawning` |
| LRU cap | `exceeding_the_cap_reaps_the_least_recently_used`, `cap_of_one_keeps_a_single_child_live` |
| RSS ceiling vs. a real measurement | `over_rss_limit_children_are_reaped`, `under_rss_limit_children_are_left_alone`, `over_rss_limit_is_false_without_a_measurement` |
| `kill_on_drop` + SIGTERM→SIGKILL patience | `terminate_child_reaps_a_live_child`, `a_child_that_never_binds_fails_with_the_service_budget` |
| `doomed` + `reap_doomed` instead of a bare SIGKILL | `an_evicted_live_child_is_given_a_chance_to_flush` (asserts the acked doc reaches the evictee's own snapshot), `an_unserved_live_child_is_queued_for_graceful_termination` |
| Dead-child eviction, corpse is not a reap | `a_dead_child_is_evicted_and_respawned`, `a_dead_child_is_evicted_without_counting_as_a_reap` |
| Unserved-socket eviction, both errnos | `a_live_child_that_stopped_serving_is_evicted_and_respawned`, `a_child_behind_a_stale_socket_file_is_evicted_and_respawned` |
| External-mode opt-out | `external_mode_skips_spawn` (its spec closure panics if called, so the path provably never resolves a binary), `external_env_only_honours_exactly_one` |

Unit coverage for the shared machinery moved to `trusty-common`'s
`uds/supervisor/tests.rs` **with the code it covers**; every assertion has a
counterpart there. `trusty-memory` keeps what is still BM25's: the env knobs,
adoption at BM25's canonical path, and the patience-vs-flush margin.

## The flake, now with its mechanism

`probe_verdict_reports_not_serving_for_a_stale_socket_file` passed 3/3 alone and
3/3 alongside the other probe tests, but failed 4 of 5 runs of the whole module
and 2 of 3 of the wider `uds::` filter — reading `Serving` for a listener the
test had already dropped.

Mechanism, identified in review: **fd inheritance across a concurrent fork.**
macOS has no atomic `SOCK_CLOEXEC`, so `UnixListener::bind` creates the socket
and sets `FD_CLOEXEC` in a second syscall. A `Command::spawn` on another test
thread that forks inside that window hands the listening fd to the child — here
a `sleep 60` stub — which holds it open. The parent's `drop(listener)` then
closes only its own descriptor; the socket stays bound and the connect completes.

It is a property of the test harness, not of the supervisor: a supervised service
binds its own socket in its own process, and this supervisor never holds one to
leak. Remedy: every test in that file is `#[serial]`, which closes the window.
10/10 clean after. The mechanism is written into the test file's module doc.

## Gates — rung 5

🔴 **Corrected 2026-08-07.** The first version of this table reported
`check_test_pointers.sh … 0 dangling — OK`. It was **red**: five dangling
pointers in `crates/trusty-common/src/uds/supervisor/`. See "How that entry came
to say OK" below. Every row here was re-run after `git add`, at `060eda02`.

```
cargo fmt --check                                                    EXIT=0
cargo clippy -p trusty-common --features uds,uds-supervisor \
  --all-targets -- -D warnings                                       EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings           EXIT=0
cargo check --workspace --all-targets \
  --exclude trusty-mpm-gui --exclude trusty-code-gui \
  --exclude trusty-agents-ui                                         EXIT=0  (SKIP_UI_BUILD=1)
cargo test --no-fail-fast -p trusty-common --features uds,uds-supervisor
   test result: ok. 365 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
cargo test --no-fail-fast -p trusty-memory
   test result: ok. 606 passed; 0 failed; 4 ignored   (+23 integration binaries, 24 suites ok, 0 failed)
cargo test -p trusty-memory --test bm25_supervisor_concurrency  ×50
   CONCURRENCY 50x: pass=50 fail=0     (6 tests per run)
cargo test -p trusty-common --features uds,uds-supervisor --lib uds::  ×10
   uds:: 10x: pass=10 fail=0           (the fd-inheritance flake, post-#[serial])
check_test_pointers.sh   resolved 22496 Test: citation(s) — 0 dangling pointers — OK
check_line_cap.sh        measured 3790 tracked .rs file(s); 4 allowlisted, 0 violations — OK
check_changelog_fragment.sh   scanned 13 changed path(s); both crates recorded — OK
check_sld.sh             56 spec doc(s) + 3135 code file(s); 0 errors, 0 warnings — OK
```

The gate command differs from the issue's in one place: `--features uds` alone
does not compile the new module. `--features uds,uds-supervisor` is what covers
this change.

### How that entry came to say OK

Not a stale run, not a wrong directory, not a pre-rename invocation. The run was
real, in this worktree, at the right moment — over a file set that excluded
everything under review.

`check_test_pointers.sh` enumerates with `git ls-files '*.rs'` (line 502). I ran
it before `git add`, while all six files under
`crates/trusty-common/src/uds/supervisor/` were untracked. `git ls-files` does
not list untracked files, so the scan covered the rest of the repo, found it
clean, and exited 0. Nothing about the output distinguishes "your files passed"
from "your files were not looked at" — the counts are repo-wide.

`check_line_cap.sh` has the identical blind spot, visible in its own output:
**3784** tracked `.rs` files before `git add`, **3790** after. It passed either
way, so it cost nothing here, but it was measuring the same incomplete set.

The changelog gate is the counter-example: it failed loudly with *"Nothing was
examined, so this gate could not have failed"* (#4618), which is exactly the
guard the other two lack.

Fix, and what stops it recurring: **stage first, then gate.** Every row above was
re-run after `git add -A`. A `git ls-files`-driven gate run against an untracked
change is not a run of that gate, and its green is not evidence.

### Baseline reds under `--include-ignored`

`cargo test --no-fail-fast -p trusty-memory -- --include-ignored` is red with 8
failures in two `#[ignore]`d suites, neither of which this diff touches
(`git diff --name-only origin/main...HEAD` lists no file in either) and neither
of which references the supervisor at all (`grep -c supervisor` → 0 in both):

- `tests/concurrent_perf.rs` — 4 × `palace_create failed: no project root found
  at or above '/'`. Environmental: the spawned daemon's cwd is `/`.
- `tests/mcp_stdio_tools.rs` — 4 perf-threshold failures on a debug build, e.g.
  `kg_assert took 28.84825ms (budget: 10ms)`.

Not matched to a canonical issue by test name or panic text; reported rather than
suppressed. **Change-specific gates pass**; `--include-ignored` for
`trusty-memory` is blocked by these two pre-existing suites.

## Review round (critic WARN, two HIGH) — addressed at `060eda02`

- **HIGH 1** — five dangling `Test:` pointers, all citing names renamed during
  authoring. Corrected to `adoption_refuses_a_world_writable_socket` /
  `adoption_accepts_a_hardened_socket_without_spawning`,
  `a_serving_child_is_reused_without_a_spawn`,
  `a_dead_child_is_evicted_without_counting_as_a_reap`,
  `an_unserved_live_child_is_queued_for_graceful_termination`, and
  `probe_verdict_reports_not_serving_for_a_missing_socket`.
- **HIGH 2** — the gate table reported that red gate as green. Corrected above,
  with the cause.
- **MEDIUM 3** — nothing bound `shutdown_flush` to a service's real budget.
  `ServiceTimeouts` now carries the sourcing rule where the next implementer
  reads it: `shutdown_flush` MUST be the supervised binary's own flush constant,
  imported, never a literal; the assert enforces the relation and has no way to
  know the child's real budget, so a literal that understates it compiles, passes,
  and SIGKILLs mid-flush.
- **LOW 4** — `ServiceTimeouts::try_new` added for runtime-derived values, with
  `SupervisorError::InvalidTimeouts`. Covered by
  `try_new_rejects_an_inverted_pair_without_panicking` and
  `try_new_matches_new_for_a_valid_pair`.
- Consequence of adding the first fallible non-async constructor:
  `SupervisorError::UntrustedSocket` and `SocketPath` now box their
  `UdsSecurityError` source, which unboxed pushed the enum past clippy's
  `result_large_err` threshold.

## Versioning

`trusty-common` is 0.30.0 in-tree against a published max of 0.28.1 (checked
against the crates.io versions API, not assumed). Under Cargo's 0.x rule the
breaking position is MINOR, and 0.28 → 0.30 is already past it, so this additive
API needs no further bump.

Refs #5089. Owes a `code-critic` round after CI.

### DIFF
--- a/crates/trusty-common/Cargo.toml
+++ b/crates/trusty-common/Cargo.toml
@@ -329,6 +329,15 @@ bm25-client = ["uds"]
 # Pulls in `thiserror` (error enum) and uses the `libc` and `tokio` workspace
 # deps already required — adds no new dependencies to the lockfile.
 uds = ["dep:thiserror"]
+# Enables the `uds::supervisor` module: the generic on-demand spawn supervisor
+# (`UdsServiceSupervisor`) promoted from trusty-memory's `Bm25Supervisor`
+# (issue #5089 step 2, ADR-0034 §1). Implies `uds` — it verifies an adopted
+# socket through `verify_socket_for_connect` and checks the `sun_path` budget
+# before spawning. Adds no new dependencies: `tokio/process`, `libc` on unix and
+# the unconditional `sys_metrics` module are already required by this crate.
+# Off by default so a consumer that only binds a socket does not compile the
+# child-process lifecycle.
+uds-supervisor = ["uds"]
 # Bundle the ONNX Runtime static libs via fastembed's
 # `ort-download-binaries-native-tls`. Right choice for macOS and modern
 # Linux (glibc ≥ 2.38). Mutually exclusive with `embedder-load-dynamic`
--- a/crates/trusty-common/changelog.d/5089-uds-service-supervisor.md
+++ b/crates/trusty-common/changelog.d/5089-uds-service-supervisor.md
@@ -34,3 +34,10 @@ Added
   #5089)
 - Every new public enum and struct is `#[non_exhaustive]` while the crate is
   unpublished at 0.30.0 against a released 0.28.1 (#5089)
+- `ServiceTimeouts::try_new` is the fallible sibling of the `const fn`
+  constructor, for a service deriving its timeouts from config rather than
+  declaring them as a `const`. `#[non_exhaustive]` left no other way to build
+  the value, so without it the runtime case could only panic. The type also now
+  documents the sourcing rule the assert cannot check: `shutdown_flush` must be
+  the supervised binary's own flush constant, imported, not a literal that
+  happens to match it today (#5089)
--- a/crates/trusty-common/changelog.d/5089-uds-service-supervisor.md
+++ b/crates/trusty-common/changelog.d/5089-uds-service-supervisor.md
@@ -1,0 +1,36 @@
+Added
+
+- `uds::supervisor::UdsServiceSupervisor`, behind the new `uds-supervisor`
+  feature: the on-demand spawn supervisor generalised out of `trusty-memory`'s
+  `Bm25Supervisor`, so `trusty-console` can start `trusty-review` /
+  `trusty-analyze` at webhook-delivery time without either being resident
+  (ADR-0034 §1, milestone `tm 1.3.5` criterion (c)). It carries the whole state
+  machine that crate had already hardened: spawn-gate serialisation against
+  double-spawn and against a fan-out that satisfies the cap per-caller,
+  adoption of a socket another process bound, an LRU cap on live children, an
+  RSS ceiling compared against a real measurement, exponential-backoff socket
+  probing, `kill_on_drop`, and SIGTERM→SIGKILL with a patience window (#5089)
+- Liveness is decided by the socket, never by `try_wait()`. `SocketVerdict` is
+  three-state and only ENOENT / ECONNREFUSED mean `NotServing`; a timeout or any
+  other errno is `Inconclusive` and leaves the child alone, so load cannot turn
+  into a respawn storm. A child evicted while still alive goes onto a `doomed`
+  queue drained under the spawn gate — SIGTERM plus socket unlink, in that order
+  — rather than being dropped onto `kill_on_drop`'s SIGKILL, because the doomed
+  child unlinks the shared socket path as the last step of its own shutdown and
+  a concurrent termination could land that unlink after the replacement bound
+  the same path (#5085, #5119, #5089)
+- `ServiceTimeouts` makes the spawn-probe budget, the child's own shutdown-flush
+  budget and the SIGTERM patience per-service values rather than constants. Its
+  constructor is a `const fn` whose assert enforces `sigterm_patience >
+  shutdown_flush`, so a service that declares its timeouts as a `const` gets the
+  same compile-time failure `bm25_supervisor.rs`'s `const _: () = assert!(…)`
+  gave — now bound to the value actually passed to the supervisor rather than to
+  a free-standing constant (#5089)
+- Socket adoption is verified, not assumed: a socket that answers but fails
+  `verify_socket_for_connect` is refused with `SupervisorError::UntrustedSocket`
+  instead of being adopted silently or falling through to a spawn that dies on
+  EADDRINUSE. This is the one path where the target's own `bind_hardened` never
+  runs, which ADR-0034 §3 makes the permission bits load-bearing for (#5099,
+  #5089)
+- Every new public enum and struct is `#[non_exhaustive]` while the crate is
+  unpublished at 0.30.0 against a released 0.28.1 (#5089)
--- a/crates/trusty-common/src/uds/mod.rs
+++ b/crates/trusty-common/src/uds/mod.rs
@@ -49,8 +49,29 @@ mod tests;
 pub mod dir;
 mod peer;
 
+/// On-demand supervision of a UDS-serving child process (#5089 step 2).
+///
+/// Why: ADR-0034 §1 needs `trusty-console` to start `trusty-review` /
+/// `trusty-analyze` at delivery time so neither has to be resident. The
+/// mechanism ADR-0032 said "does not yet exist" did exist, as `trusty-memory`'s
+/// `Bm25Supervisor`; this is that supervisor with its BM25-specific parts lifted
+/// into per-service configuration.
+/// What: [`supervisor::UdsServiceSupervisor`] plus its config, error and probe
+/// types. Gated behind the `uds-supervisor` feature because it pulls the whole
+/// child-process lifecycle in, which a crate that only binds a socket does not
+/// need.
+/// Test: `supervisor/tests.rs`, and `trusty-memory`'s
+/// `tests/bm25_supervisor_concurrency.rs` against real children.
+#[cfg(feature = "uds-supervisor")]
+pub mod supervisor;
+
 pub use dir::prepare_socket_dir;
 pub use peer::{ensure_peer_is_self, peer_uid, self_uid};
+#[cfg(feature = "uds-supervisor")]
+pub use supervisor::{
+    ServiceTimeouts, SocketVerdict, SpawnSpec, SupervisorConfig, SupervisorError,
+    UdsServiceSupervisor,
+};
 
 use std::fs::Permissions;
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
--- a/crates/trusty-common/src/uds/supervisor/child.rs
+++ b/crates/trusty-common/src/uds/supervisor/child.rs
@@ -1,0 +1,163 @@
+//! Child-process lifecycle for [`super::UdsServiceSupervisor`] (#5089).
+//!
+//! Why: spawn, graceful termination, socket-file cleanup and the RSS
+//! measurement are the four mechanical operations the supervisor's state
+//! machine sits on top of. Keeping them here leaves the state machine readable
+//! and keeps each one testable without standing up a supervisor.
+//! What: [`ChildHandle`] (the per-instance bookkeeping), [`spawn_child`],
+//! [`terminate_child`], [`remove_socket_file`], and [`over_rss_limit`].
+//! Test: `tests.rs` — `over_rss_limit_is_false_without_a_measurement`,
+//! `over_rss_limit_is_false_when_disabled`, `terminate_child_reaps_a_live_child`.
+
+use std::path::{Path, PathBuf};
+use std::process::Stdio;
+use std::time::Duration;
+
+use tokio::process::{Child, Command};
+
+use super::{SpawnSpec, SupervisorError};
+
+/// Per-instance child bookkeeping stored in the supervisor's map.
+///
+/// Why: keeping the `Child` alongside the socket path lets every reaping path
+/// terminate the process and clean up its socket in one pass without
+/// re-resolving anything.
+/// What: the `tokio::process::Child`, the socket it was told to bind, and a
+/// monotonic LRU stamp. The stamp is a counter rather than an `Instant` because
+/// two `ensure_running` calls inside the same clock tick would compare equal and
+/// make the victim choice arbitrary — which is exactly what a burst fan-out
+/// produces.
+/// Test: covered through every supervisor test that reaps.
+#[derive(Debug)]
+pub(super) struct ChildHandle {
+    pub(super) child: Child,
+    pub(super) socket_path: PathBuf,
+    pub(super) last_used: u64,
+}
+
+/// Spawn one child from `spec`.
+///
+/// Why: isolates the `Command` builder so the supervisor's state machine can be
+/// read without it, and so the stdio and `kill_on_drop` decisions live in one
+/// place rather than at each service's call site.
+/// What: creates `spec.create_dirs` first (failing here gives a cleaner error
+/// than a child that immediately exits), closes stdin and stdout — a supervised
+/// UDS service speaks its socket, not its pipes — inherits stderr so the child's
+/// tracing output reaches the parent's log stream, and sets `kill_on_drop` so an
+/// unsupervised drop reaps the child rather than leaking it.
+/// Test: `spawn_child_creates_requested_directories`,
+/// `spawn_child_reports_a_missing_binary`.
+pub(super) async fn spawn_child(
+    service: &str,
+    key: &str,
+    spec: &SpawnSpec,
+) -> Result<Child, SupervisorError> {
+    for dir in &spec.create_dirs {
+        if !dir.exists() {
+            tokio::fs::create_dir_all(dir)
+                .await
+                .map_err(|source| SupervisorError::CreateDir {
+                    service: service.to_string(),
+                    key: key.to_string(),
+                    path: dir.clone(),
+                    source,
+                })?;
+        }
+    }
+
+    Command::new(&spec.program)
+        .args(&spec.args)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::inherit())
+        .kill_on_drop(true)
+        .spawn()
+        .map_err(|source| SupervisorError::Spawn {
+            service: service.to_string(),
+            key: key.to_string(),
+            program: spec.program.clone(),
+            source,
+        })
+}
+
+/// Send SIGTERM, wait out the service's patience window, then SIGKILL.
+///
+/// Why: a clean SIGTERM is the only signal a child's own shutdown handler can
+/// act on, and for a service that acks writes before flushing them, that
+/// handler is the difference between durable and lost. The wait must exceed the
+/// child's OWN budget with margin, not merely match it — signal delivery, the
+/// flush, socket cleanup and exit all have to fit. [`super::ServiceTimeouts`]
+/// enforces that relationship at construction.
+/// What: `libc::kill(SIGTERM)` on unix (tokio's `Child` exposes no SIGTERM),
+/// then `wait()` under `patience`, then tokio's `kill()` — which sends SIGKILL
+/// and waits, so the process is definitely gone when it returns.
+/// Test: `terminate_child_reaps_a_live_child`.
+pub(super) async fn terminate_child(child: &mut Child, patience: Duration) -> std::io::Result<()> {
+    #[cfg(unix)]
+    if let Some(pid) = child.id() {
+        // SAFETY: `libc::kill` is safe to call with any pid; the kernel returns
+        // -1/EINVAL/ESRCH rather than misbehaving on bad input. The return value
+        // is intentionally ignored — either the signal landed (the child exits)
+        // or it did not (the SIGKILL below covers it).
+        unsafe {
+            let _ = libc::kill(pid as libc::pid_t, libc::SIGTERM);
+        }
+    }
+
+    match tokio::time::timeout(patience, child.wait()).await {
+        Ok(Ok(status)) => {
+            tracing::debug!(?status, "supervised child exited after SIGTERM");
+            Ok(())
+        }
+        Ok(Err(e)) => Err(e),
+        Err(_elapsed) => {
+            tracing::warn!("supervised child ignored SIGTERM after {patience:?} — sending SIGKILL");
+            child.kill().await
+        }
+    }
+}
+
+/// Best-effort unlink of a child's socket file.
+///
+/// Why: a child that exits cleanly unlinks its own socket, but a SIGKILLed one
+/// leaves the file behind and the next spawn for that instance then fails to
+/// bind with EADDRINUSE — and a reaped instance is exactly the one most likely
+/// to be spawned again shortly.
+/// What: `remove_file`; `NotFound` is the expected clean-exit case and is not
+/// logged. Any other error is logged at `debug!` and ignored — a stale socket
+/// costs one failed spawn, not correctness.
+/// Test: covered through the supervisor's shutdown and reap paths.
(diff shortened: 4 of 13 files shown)


bobmatnyc:

Reviewed at `853ed1af` in a detached read-only worktree. Every invariant the promotion claims to carry was checked against `origin/main`'s `bm25_supervisor.rs` line by line, not against the PR body's table. The state machine survived intact. One CI gate is red and the PR body reports it green.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/uds/supervisor/mod.rs` | 188, 336 | Four `Test:` pointers cite test names that do not exist. [`Test: doc-comment pointer lint`](https://github.com/bobmatnyc/trusty-tools/actions/runs/31216240254/job/92990313881) is **failing** on this head; the PR body reports `scripts/check_test_pointers.sh EXIT=0`. | Rename to the real tests (below) | Fix here |
| HIGH | `crates/trusty-common/src/uds/supervisor/probe.rs` | 109 | Same gate: `probe_reports_false_for_a_missing_socket` does not exist | → `probe_verdict_reports_not_serving_for_a_missing_socket` | Fix here |
| MEDIUM | `crates/trusty-common/src/uds/supervisor/config.rs` | 110–132 | `ServiceTimeouts::new` checks the *relation* between two declared numbers, but nothing binds `shutdown_flush` to the service's real budget. A second service passing a literal compiles, passes the assert, and still SIGKILLs mid-flush. BM25 escapes only because it imports `trusty_bm25_daemon::SHUTDOWN_FLUSH_TIMEOUT` — and that discipline is documented in `bm25_supervisor.rs`, which a second service's author never opens. | Add one line to `new`'s doc: `shutdown_flush` MUST be the supervised service's own exported constant, never a restated literal, because a copy stays equal to itself while the real value drifts | Fix here |
| LOW | `crates/trusty-common/src/uds/supervisor/config.rs` | 122 | `#[non_exhaustive]` + a panicking-only constructor leaves no way to build `ServiceTimeouts` from a runtime value without risking a daemon abort. This crate already ships env knobs of exactly this shape (`TRUSTY_BM25_MAX_DAEMONS`, `TRUSTY_BM25_RSS_LIMIT_MB`). | Add `try_new -> Result<Self, _>` beside the `const fn`. Non-breaking to add later, cheap to add now | Fix here |

Exact renames for the two HIGH rows:

| Cited | Actual |
|---|---|
| `adoption_requires_a_hardened_socket` | `adoption_refuses_a_world_writable_socket` / `adoption_accepts_a_hardened_socket_without_spawning` |
| `a_serving_child_is_reused` | `a_serving_child_is_reused_without_a_spawn` |
| `a_dead_child_is_evicted` | `a_dead_child_is_evicted_without_counting_as_a_reap` |
| `an_unserved_child_is_queued_as_doomed` | `an_unserved_live_child_is_queued_for_graceful_termination` |
| `probe_reports_false_for_a_missing_socket` | `probe_verdict_reports_not_serving_for_a_missing_socket` |

## Required Changes

1. Correct the five test pointers so [`Test: doc-comment pointer lint`](https://github.com/bobmatnyc/trusty-tools/actions/runs/31216240254/job/92990313881) goes green, and correct the gate table in the PR body — it currently claims `EXIT=0` for a check that is red.
2. Document the `shutdown_flush` sourcing contract in `ServiceTimeouts::new`.
3. Optional: `ServiceTimeouts::try_new`.

## Invariant-by-invariant

**1. Socket-backed liveness ([#5085](https://github.com/bobmatnyc/trusty-tools/issues/5085) / [#5119](https://github.com/bobmatnyc/trusty-tools/pull/5119)) — SURVIVED.** `lookup_live` (`mod.rs:337`) still requires both answers: `try_wait() == Ok(None)` at 342, then `probe_socket_verdict(...) != NotServing` at 371 with the map lock released. The pid-identity re-check before eviction (390–393) is byte-for-byte the original, including the "two unreadable pids must not alias" `matches!` guard. `SocketVerdict` is three-state; `probe.rs:93–100` maps only `NotFound` and `ConnectionRefused` to `NotServing`, and both the elapsed arm and every other errno land on `Inconclusive`. No respawn-storm path.

**2. The doomed queue ([#5119](https://github.com/bobmatnyc/trusty-tools/pull/5119)) — SURVIVED, ordering included.** The evicted-LIVE child goes to `doomed` (396–398), not to `kill_on_drop`. `reap_doomed` (415) runs `terminate_child` + `remove_socket_file`, does not touch `reaped`, and is called at `mod.rs:230` — after the under-gate re-check, **before** the adopt probe at 232, before `enforce_limits` at 253, before the spawn at 268. That is the ordering the doc claims and it is the ordering the code has. `shutdown` drains the queue first (516), same as before.

**3. The `const fn` guard — equivalent-or-stronger for BM25; one residual for a second service.** `BM25_TIMEOUTS` (`bm25_supervisor.rs:127`) is a `const` item evaluating `ServiceTimeouts::new(SPAWN_PROBE_TIMEOUT, trusty_bm25_daemon::SHUTDOWN_FLUSH_TIMEOUT, SIGTERM_PATIENCE)`. The assert fires during const evaluation against the daemon's imported real constant, so the compile-time failure is preserved and is now bound to the value actually handed to the supervisor rather than to a free-standing constant a refactor could orphan. `sigterm_patience_exceeds_the_daemon_flush_budget` additionally pins `shutdown_flush == the daemon's real budget` and a 2 s margin. Stronger than what it replaced.

The residual is the MEDIUM above: for a *second* service, the assert constrains only the relation. `ServiceTimeouts::new(Duration::from_millis(100), Duration::from_secs(1), Duration::from_secs(2))` compiles, passes, and is wrong if that service's real flush budget is 10 s. The mechanism has no type-level fix; the doc is the fix.

**4. Everything else — spot-checked against `origin/main`, all preserved.** Spawn gate serialising re-check → drain → adopt → limits → spawn (`mod.rs:217–299`, same order). Aggregate cap with `headroom` evaluated under the gate (253, 455–483) with the two-pass RSS-then-LRU victim selection unchanged. `over_rss_limit` (`child.rs:158`) still returns `false` for both "no pid" and "no reading" — `None` never means zero. `kill_on_drop(true)` at `child.rs:73`. SIGTERM → `wait` under patience → `kill()` at `child.rs:95–118`. Dead-child eviction returns `None` without incrementing `reaped` and without entering `doomed` (a corpse owes no flush) — asserted at `tests.rs:526–530`. External-mode opt-out reads `== Ok("1")` only. New, not a loss: `check_sun_path_budget` now runs before the spawn (255).

Every unit test deleted from `bm25_supervisor_tests.rs` has a counterpart in `uds/supervisor/tests.rs`. No coverage was dropped, and I confirmed the 34 supervisor tests are actually reachable under CI's `cargo test --workspace` (see point 6).

**5. Adoption now verifies — inert for BM25, and #5169 does not conflict.** The "inert for post-#5099 daemons" claim is true, and true more strongly than stated: [#5124](https://github.com/bobmatnyc/trusty-tools/pull/5124) moved `socket_path_for_palace` from `$TMPDIR/trusty-bm25-<palace>.sock` to `scratch_socket_dir()/trusty-bm25-<palace>.sock`. A live pre-#5099 daemon is at a *different path*, so the new supervisor never probes it and the `UntrustedSocket` arm is unreachable on the upgrade boundary. The orphaned old-path daemon is a #5124 consequence, not this PR's.

**Ruling on #5169 vs #5170: consistent, because the two sites differ in ownership, not in policy.** [#5169](https://github.com/bobmatnyc/trusty-tools/pull/5169) self-heals in `bind_singleton` — the ctrl socket is a singleton the process *owns*, so unlinking and rebinding replaces its own bad artifact, and refusing there would genuinely brick the project because nothing else would ever write that path. Here the supervisor is a *client* deciding whether to adopt a socket some other live process bound. Unlinking that would kill a service the supervisor did not spawn and cannot identify; refusing with a named error and leaving it alone is the correct asymmetry. Neither PR bricks what the other heals. Both hold the same line on the underlying check: the untrusted socket is never written to.

**6. Feature gating — correct and actually covered.** `uds-supervisor = ["uds"]`, and `pub mod uds` is `#[cfg(all(unix, feature = "uds"))]`, so nothing is reachable without it and the Windows path never sees `libc::kill`. Verified rather than assumed:

```
cargo check -p trusty-common --no-default-features --features uds-supervisor --all-targets
EXIT=0
```

CI coverage: CI runs `cargo test --workspace`, `trusty-memory` enables `uds-supervisor`, and resolver-v2 unification turns it on for `trusty-common`'s own lib test binary. Confirmed empirically —

```
cargo test -p trusty-common -p trusty-memory --lib -- --list   →   34 uds::supervisor::tests::* listed
```

The 823-line test file is not dead in CI.

**7. `#[non_exhaustive]` — verified on all five, and all five stay usable cross-crate.** `SocketVerdict` (`probe.rs:68`, enum, wildcard-arm cost), `SupervisorError` (`error.rs:22`, enum), `ServiceTimeouts` (`config.rs:76`), `SupervisorConfig` (`config.rs:180`), `SpawnSpec` (`config.rs:251`). The E0639 trap the attribute sets on a struct is avoided in every case: each has a constructor plus `with_*`/builder methods, the `..self` functional updates at `config.rs:147` and `156` are intra-crate and therefore legal, and `SpawnSpec`'s derived `Default` is a method call rather than construction syntax. `UdsServiceSupervisor` carries no attribute and needs none — it has no public fields.

## Notes

**On the `#[serial]` remedy for `probe_verdict_reports_not_serving_for_a_stale_socket_file`.** Adequate, and not masking a production bug. The likely mechanism, which the file declines to name: the failing assertion read `Serving` for a listener it had already dropped, and the interference tracked the fork-heavy tests. A concurrent `Command::spawn` forks between `socket()` and the `FD_CLOEXEC` `fcntl` on macOS, the `sleep` child inherits the listening fd, and the listener stays open in that child after the parent drops it — so the connect completes and the verdict is `Serving`. Serialising closes the window because no fork happens while the listener is alive.

That mechanism has no analogue in the supervisor's own process: the parent never holds a supervised service's listening socket, so its forks cannot leak one. Worth writing into the module doc when someone next touches it, since an uncharacterised "do not parallelise" tends to get removed by the next reader who cannot reproduce it.

**Not flagged.** `shutdown()` does not take the spawn gate, so a concurrent `ensure_running` can insert a child after the map is drained. Behaviour-identical to what shipped in `origin/main`, bounded by `kill_on_drop`, and out of scope for a promotion.

Issue #5084, opened 2026-08-07
disabled_salvage_budget_costs_a_dead_entry_nothing_but_a_stat has no wall-clock headroom under --workspace load


bobmatnyc:
## Defect

`crates/trusty-search/src/commands/start/tests_4846.rs`, test `disabled_salvage_budget_costs_a_dead_entry_nothing_but_a_stat` (added by #4846), asserts a hard wall-clock ceiling: `boot <= max(2 * one_walk, 250ms)`.

It has no headroom. Measured alone under `--test-threads=1` on a 16-core machine at load average 3.8-6.75, whole-test wall clock sits at 230-300ms against the 250ms floor — already at or over its own ceiling with nothing competing against it.

In CI it runs inside `cargo test --workspace` at full parallelism on one shared GitHub Actions runner, which credibly consumes that remaining margin. It failed exactly that way on 2026-08-07: run `31179690320`, PR #5074, head `ea92679cd`, `boot took 457.328292ms against a single-walk cost of 14.764325ms (ceiling 250ms)`.

Second defect, and it's why the first is hard to see coming: the test prints `boot`, `one_walk`, and `ceiling` only on panic. A passing run gives zero visibility into how close it came — no way to notice the margin eroding before it fails.

## Evidence this is headroom, not a regression

20 runs on `origin/main` (`trusty-common` 0.29) and 20 runs at PR #5074's SHA (0.30): 40/40 passed, means 0.2405s vs 0.242s — 1.5ms apart, inside each side's ~40ms run-to-run spread. The trusty-common 0.29->0.30 diff touches nothing reachable from trusty-search's boot path; trusty-search doesn't enable the `memory-core` feature gating most of it. `Cargo.lock` shows no new transitive dependency. The CI failure was a load-correlated excursion.

## Resolution — two parts, don't conflate them

1. Success-path visibility: print `boot`, `one_walk`, and the computed ceiling on pass as well as failure, so margin erosion is observable before it becomes a red gate.
[...]

Issue #4741, opened 2026-08-04
flaky: fastembed init failure poisons ENV_LOCK and cascades into 6 unrelated trusty-common embedder tests


bobmatnyc:
`cargo test --workspace` went red in CI with **7** failures in `trusty-common`, but there is only ONE real failure — the other six are collateral from a poisoned mutex.

## Root failure

```
thread 'embedder::tests::reference_accuracy_tests::default_model_matches_sentence_transformers_reference'
panicked at crates/trusty-common/src/embedder/tests/reference_accuracy_tests.rs:143:43:
called `Result::unwrap()` on an `Err` value: failed to initialise fastembed
(tried AllMiniLML6V2 and AllMiniLML6V2Q)
```

The model could not be initialised — a network/cache-dependent condition, not a logic failure.

## The cascade

That test takes `ENV_LOCK` at line 135 and holds it across the whole construct-and-embed window (deliberately, per #3711's fix for a model-selection race). When the `unwrap()` at 143 panics, the lock is **poisoned**, and every subsequent test that locks it with a bare `.unwrap()` dies:

```
thread 'embedder::tests::resolve_default_embedding_model_defaults_to_fp32'
panicked at crates/trusty-common/src/embedder/mod.rs:448:34:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

Six victims, all `PoisonError`: `resolve_default_embedding_model_defaults_to_fp32`, `resolve_default_embedding_model_ignores_unknown`, `resolve_default_embedding_model_int8_opt_in`, `resolve_expected_provider_default_matches_platform`, `resolve_expected_provider_forces_cpu`, `resolve_fastembed_cache_dir_prefers_env_vars`.

The codebase is already split on this idiom:

```
27 × ENV_LOCK.lock().unwrap();                              ← poison-intolerant
25 × ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());    ← poison-tolerant
```

The six victims are in the first group. Nothing about them depends on the model.

## Two separable defects

1. **The cascade is avoidable.** These are env-var tests whose correctness does not depend on a previous holder finishing cleanly. Bringing them onto the poison-tolerant idiom already used by 25 other sites turns a 7-failure run into a 1-failure run that names the real cause.
[...]

PR #5178, opened 2026-08-08
fix(trusty-common): scan exactly below 256 points instead of trusting HNSW recall

### LINKED ISSUE #5171: trusty-common: HnswStore::search returns only the descent pivot's connected component, silently dropping true nearest neighbours
## Defect

`HnswStore::search` can return a strict subset of an index's points and omit a true nearest neighbour. Found while root-causing [#5162](https://github.com/bobmatnyc/trusty-tools/issues/5162); that issue's fix works around it in one test, it does not address this.

`hnsw_rs` seeds its level-assignment RNG from OS entropy inside `Hnsw::new` (`rng: Arc<Mutex<StdRng>>` = `StdRng::from_os_rng()`), so every index build is a different random graph. `HnswStore::open` rebuilds the in-memory graph from redb on every open, so the miss set changes on every daemon restart.

With `ef_search` (64) larger than the point count, `search_layer`'s early-return can never fire — `c` is always a visited point, so `-(c.dist) > f.dist` is false. The loop therefore drains its candidate heap and returns exactly the set reachable from the descent pivot along layer-0 neighbour lists. Those lists are pruned by Navarro's heuristic, and the reachable set is not the whole index.

## Evidence

A 26-vector index built from `crates/trusty-common/src/memory_core/retrieval/tests.rs`'s `vec_at_cosine` fixture — one vector at cosine 0.80, one at 0.75, 24 padding vectors from 0.60 down to 0.14 — asked for `search(qv, 30, 64)`:

```
trials=20000  2nd-nearest absent from the returned pool = 85  (0.42%)
              nearest absent = 0
              returned pool size on a miss = 15..20 of 26
```

Rate by index size (4,000 trials each), showing this is connectivity and not density:

```
6 points  25/4000   10 points  9/4000   14 points 14/4000   16 points 12/4000
18 points 15/4000   22 points 12/4000   26 points 20/4000   42 points 19/4000
```

Neither tuning knob removes it, at 26 points over 20,000 trials each:

```
baseline                  85
set_keeping_pruned(true)  58
set_extend_candidates     90
both                      81
```

## Impact

Memory recall drops a genuinely relevant drawer from the candidate pool at a low rate, and which drawer is invisible changes on every palace open. This is the failure mode [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904) set out to measure — its self-retrieval probe scored recall@1 291/400 for similarity alone, and this sits underneath that number. Note the fixture's vectors are synthetic and share sign-pattern structure, which is unusually hard on the pruning heuristic; the rate on real embeddings is unmeasured and should be measured before sizing the fix.

## Resolution

Not diagnosed beyond the above. Candidate directions, none evaluated: raise `HNSW_MAX_NB_CONNECTION` from 16; add a reverse-edge pass at insert so layer-0 reachability is symmetric; fall back to a brute-force scan below some point count.

bobmatnyc:
## Outcome

Closes [#5171](https://github.com/bobmatnyc/trusty-tools/issues/5171). `HnswStore::search` now ranks by scanning every point when the index holds 256 or fewer, so a small palace can no longer drop a genuinely relevant drawer from the candidate pool.

## Step 1 — the measurement, before choosing a fix

The issue's 0.42% came from `vec_at_cosine`, whose vectors share sign-pattern structure. The real-embedding rate was unknown, so I measured it first.

**Methodology.** I copied two live palaces' `index.usearch.redb` files out of `~/Library/Application Support/trusty-memory/palaces/` (read-only copy; the live files were never opened for write) and decoded their `VECTORS` rows — 1354 live 384-dim vectors written by the configured embedder against this repo's own memory. These are production embeddings, not a model of them. For each collection size I sampled a pool from that corpus, built a graph, and queried with corpus vectors held out of the pool, which is the realistic case: a recall query is a new sentence, not a row already in the index. Truth is a brute-force cosine ranking computed in f64 independently of `hnsw_rs`. A query counts as a miss when any member of the exact top-5 is absent from the pool `search` returned. 40 independently seeded graphs × 100 queries = 4000 queries per cell, so graph-to-graph variation is sampled rather than assumed away.

**Result — miss rate is 3–10× the fixture's, and `ef_search` is not the lever.**

```
n=    6 ef=  64  queries=  4000  recall@5<1.0 on    85 (2.125%)  nearest_lost=   0 (0.000%)  neighbours_lost=   85/ 20000 (0.425%)
n=   10 ef=  64  queries=  4000  recall@5<1.0 on   174 (4.350%)  nearest_lost=   0 (0.000%)  neighbours_lost=  174/ 20000 (0.870%)
n=   16 ef=  64  queries=  4000  recall@5<1.0 on   117 (2.925%)  nearest_lost=   0 (0.000%)  neighbours_lost=  119/ 20000 (0.595%)
n=   26 ef=  64  queries=  4000  recall@5<1.0 on    13 (0.325%)  nearest_lost=   0 (0.000%)  neighbours_lost=   13/ 20000 (0.065%)
n=   42 ef=  64  queries=  4000  recall@5<1.0 on    37 (0.925%)  nearest_lost=   0 (0.000%)  neighbours_lost=   39/ 20000 (0.195%)
n=   64 ef=  64  queries=  4000  recall@5<1.0 on    38 (0.950%)  nearest_lost=   5 (0.125%)  neighbours_lost=   38/ 20000 (0.190%)
n=  128 ef=  64  queries=  4000  recall@5<1.0 on    44 (1.100%)  nearest_lost=   0 (0.000%)  neighbours_lost=   45/ 20000 (0.225%)
n=  128 ef= 256  queries=  4000  recall@5<1.0 on    35 (0.875%)  nearest_lost=   0 (0.000%)  neighbours_lost=   36/ 20000 (0.180%)
n=  256 ef=  64  queries=  4000  recall@5<1.0 on   129 (3.225%)  nearest_lost=  13 (0.325%)  neighbours_lost=  132/ 20000 (0.660%)
n=  256 ef= 256  queries=  4000  recall@5<1.0 on   112 (2.800%)  nearest_lost=  13 (0.325%)  neighbours_lost=  114/ 20000 (0.570%)
n=  512 ef=  64  queries=  4000  recall@5<1.0 on   295 (7.375%)  nearest_lost=  63 (1.575%)  neighbours_lost=  304/ 20000 (1.520%)
n=  512 ef= 256  queries=  4000  recall@5<1.0 on   253 (6.325%)  nearest_lost=  58 (1.450%)  neighbours_lost=  259/ 20000 (1.295%)
n= 1024 ef=  64  queries=  4000  recall@5<1.0 on   335 (8.375%)  nearest_lost=  73 (1.825%)  neighbours_lost=  356/ 20000 (1.780%)
n= 1024 ef= 128  queries=  4000  recall@5<1.0 on   303 (7.575%)  nearest_lost=  69 (1.725%)  neighbours_lost=  278/ 20000 (1.605%)
n= 1024 ef= 256  queries=  4000  recall@5<1.0 on   278 (6.950%)  nearest_lost=  65 (1.625%)  neighbours_lost=  292/ 20000 (1.460%)
```

Two caveats on how to read it. The held-out queries are document embeddings, not question embeddings, and questions occupy a different region of the space — so the rate is an estimate whose bias direction is unknown. And the multiplier against the fixture's 0.42% is not a measured constant: with 40 graphs per cell and no error bars, and an n=26 cell at 0.325% sitting below n=10 at 4.35%, graph-to-graph variance dominates the cell-to-cell differences. What the data supports is that the loss is percent-level rather than sub-percent, and that the `ef=64/128/256` identity below the boundary is structural.

Two things fall out of that table.

At 6–16 points, 2.1–4.4% of queries lose a true top-5 neighbour, and the `ef=64` / `ef=128` / `ef=256` rows are *byte-identical* — the candidate budget already exceeds the point count, so raising it changes nothing. That is the issue's mechanism reproduced on production data: the answer is fixed by which points are reachable from the descent pivot, not by how many candidates the search is allowed to hold.

Above the boundary `ef_search` starts to bite, but weakly: 8.375% → 6.950% going from 64 to 256 at 1024 points, at 4× the candidate work. That is the fourth knob the investigation had not tried, and it is not the fix either.

**Which sizes real palaces actually reach.** Counting live vectors in every palace on this machine:

```
CENSUS-SUMMARY palaces=94  <=64: 87 (93%)  <=128: 89 (95%)  <=256: 93 (99%)  <=1024: 93 (99%)
```

`trusty-tools` at 1354 is the only palace above 256; the median is under 10. So the sizes where the defect is worst are the sizes essentially every palace occupies.

## Step 2 — the fix, and why the rate justifies it

`HnswStore::search` scans every point when `Hnsw::get_nb_point()` is at most 256, and takes the graph path above that. The scan lives in [`exhaustive.rs`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5171-hnsw-recall/crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs) and evaluates the index's own `DistCosine`, so distances are identical to what the traversal would have reported for the same points.

The threshold is not a guess about where the bug stops — the recall loss continues past it. It is where the scan stops being free. `search` already builds the whole `VECTOR_KEYS` reverse map and the tombstone set on every call, so it is already O(n) in redb reads and `String` allocations; at 256 points a 384-dim float scan costs about the same as work the function was already doing. Past that, HNSW's sublinear cost is worth its approximation, which is the trade the algorithm exists to make.

Deterministic seeding was considered and rejected as a fix. It makes the failure reproducible, not absent — the same drawer would go missing on every open instead of a different one. It is worth having for testability, but the regression test below gets that property from the scan being exact, without pinning the graph.

**Performance.** End-to-end `HnswStore::search`, k=10, same seeds and machine, before and after:

| points | before | after | delta |
|---|---|---|---|
| 16 | 12.56µs | 11.08µs | −12% |
| 64 | 34.14µs | 27.33µs | −20% |
| 128 | 55.57µs | 45.33µs | −18% |
| 256 | 93.06µs | 98.17µs | +5% |
| 512 | 145.79µs | 139.87µs | graph path, unchanged |
| 1024 | 199.86µs | 201.71µs | graph path, unchanged |

The scan is *faster* than the traversal below 128 points. With `ef=64` over 64–128 points the graph search evaluates most of the collection anyway, through a visited set and two heaps; one flat pass with sequential access beats it. Exact results at lower cost, up to a +5% ceiling at the threshold itself.

**Post-fix recall, measured the same way:**

```
POSTFIX n=    6  queries= 1000  recall@5<1.0 on    0 (0.000%)  mean_search=6.563µs
POSTFIX n=   16  queries= 1000  recall@5<1.0 on    0 (0.000%)  mean_search=13.389µs
POSTFIX n=   64  queries= 1000  recall@5<1.0 on    0 (0.000%)  mean_search=40.77µs
POSTFIX n=  128  queries= 1000  recall@5<1.0 on    0 (0.000%)  mean_search=60.081µs
POSTFIX n=  256  queries= 1000  recall@5<1.0 on    0 (0.000%)  mean_search=107.028µs
POSTFIX n=  257  queries= 1000  recall@5<1.0 on   34 (3.400%)  mean_search=111.538µs
POSTFIX n=  512  queries= 1000  recall@5<1.0 on   56 (5.600%)  mean_search=166.823µs
POSTFIX n= 1024  queries= 1000  recall@5<1.0 on   87 (8.700%)  mean_search=240.568µs
```

Zero misses at every size through 256, 3.4% at 257. The seam is exactly where the constant puts it.

Three smaller fixes ride along because the scan reaches states the traversal did not:

- **Shadowed drawers are scored by their current vector.** `upsert` on an already-mapped uuid leaves the old embedding in the graph under the same `vector_id`; `hnsw_rs` cannot remove a point. Ranking that drawer by whichever copy is nearer means a query against text it NO LONGER HOLDS scores it at distance 0.0 — similarity 1.0 to `VectorStore::search`, enough to clear a 0.99 near-duplicate threshold against superseded content, and `palace_reembed` creates that state in bulk. `upsert` now records which ids have a shadow copy, and `search` re-scores exactly those against their `VECTORS` row, which is the only authority on what a drawer currently holds. One point read per shadowed candidate, and nothing at all in the steady state where no re-upsert has happened since the palace opened. The drawer also stops occupying two result slots.
- **Tombstones are excluded inside the scan, not after it.** `want` is a small over-fetch and `delete` never removes a point from the graph, so a palace that had deleted most of its drawers filled every returned slot with dead ids and handed the caller nothing.
- **Empty index panic.** `hnsw_rs`'s point iterator unwraps an entry point that does not exist until the first insert (`hnsw.rs:662`), so routing a fresh palace down the scan panicked inside the library. Caught by `backfill_reembeds_a_marked_drawer`, whose store searches before it embeds; guarded and covered by its own test.

The path is selected on the **live drawer count** (`reverse.len()`), not `Hnsw::get_nb_point()`. The graph count includes tombstones and shadows, which accumulate within a session, so gating on it would let a 200-drawer palace churn past 256 points and silently revert to the path this change exists to replace. Scan cost does follow the graph count, but the overhang is bounded by churn since the last open — one extra point per delete or re-upsert — so a bulk `palace_reembed` at most doubles it, and a restart resets it.

## Out of scope

Recall above 256 points. At 1024 the loss is 8.4% of queries and 1.8% of true nearest neighbours, which is within the range an approximate index is expected to produce and which no `hnsw_rs` knob measured here removes — the four tried across this issue are `set_keeping_pruned`, `set_extend_candidates`, both together, and now `ef_search`. Improving it means a different index structure or a rerank pass over a wider candidate pool, not a parameter change. This PR does not touch it and does not claim to; `trusty-tools` is the only palace on this machine in that regime. It also sits underneath [#4904](https://github.com/bobmatnyc/trusty-tools/issues/4904)'s recall@1 of 291/400, which this PR improves for 99% of palaces but does not resolve.

## Regression test, with failing-before evidence

`search_returns_the_exact_top_k_below_the_exhaustive_threshold` builds the same 12-vector collection under 12 independently seeded graphs and asserts `search` returns exactly the brute-force top-5 for 25 queries against each. Rebuilding is what samples the OS-seeded RNG a single-build test cannot see. Post-fix the assertion holds with probability 1, because the scan is exact — it cannot flake. Pre-fix it failed on ~3% of the 300 (graph, query) pairs, so a passing run was vanishingly unlikely.

Demonstrated by setting `EXHAUSTIVE_SCAN_MAX_POINTS` to 0, which routes `search` down the pre-fix graph path with nothing else changed, and running the test 10 times:

```
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.86s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.41s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.10s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.87s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.55s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.80s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.27s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 1.71s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 1.45s
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 939 filtered out; finished in 0.73s
```

10 of 10. A representative failure — rank 4 of the true top-5 is missing and a worse drawer has taken its place:

```
assertion `left == right` failed: build 1, query 14: search must return the exact top-5 for a 12-point index (#5171)
  left: [..."894b25e5-5cbb-4327-a2ce-ca1814913d3f", "ad748527-ec61-4dff-9d23-395baecc2d62", "b59b4ae6-44b9-4bee-9bdd-2b4fbdac8ca9"]
 right: [..."894b25e5-5cbb-4327-a2ce-ca1814913d3f", "10d81b67-5a82-45b3-a46a-3385b7eb85d5", "ad748527-ec61-4dff-9d23-395baecc2d62"]
```

Review round 1 added two more, each demonstrated failing against the behaviour it replaces (`CRITIC_DEMO_*` switches, applied temporarily and removed before commit):

- `search_scores_a_re_upserted_drawer_by_its_current_vector` — queries with the **stale** vector, the only input that separates "scored by the current vector" from "scored by the nearer copy"; querying `current` makes the two branches agree, which is why the first version of this test did not catch it. With shadow resolution disabled, 3 of 3 runs FAILED:

```
a superseded vector must not score its drawer as an exact match (distance 0, hits [("27cd10c8-…", 0.0), ("3dcb10f0-…", 0.89625216), …])
```

- `deleting_drawers_does_not_push_a_small_palace_off_the_exhaustive_path` — fills past the threshold, deletes down to 6 live drawers leaving 300 graph points, asserts search is still exact. Gated on `get_nb_point()`, 3 of 3 runs FAILED with an empty result set.

Four more tests, none probabilistic:

- `exhaustive_scan_returns_every_point_the_graph_holds` — asserts the invariant directly and deterministically: over a 40-point graph the scan returns all 40 ids in exactly the order of an independently computed f64 brute-force ranking, regardless of which graph `hnsw_rs` built.
- `search_uses_the_graph_above_the_exhaustive_threshold` — pins the constant and checks an index one point past it still answers correctly through the graph, so nobody silently turns every recall into a linear scan.
- `search_drops_a_shadowed_candidate_whose_vector_row_is_gone` — `compact_orphans` can remove the row between the search and the re-read; there is no honest distance to report, and the stale copy is exactly what must not be reported, so the candidate is dropped.
- `search_on_an_empty_index_returns_nothing` — the `hnsw_rs` iterator panic.

The measurement harness itself was a temporary `#[ignore]`d module and is not in the diff; it needs a live palace, so it cannot be a repo test.

## Risk

`trusty-common`'s `memory-core` feature; behaviour change to recall semantics on every palace with ≤256 vectors, which is 99% of them. Results below the threshold become exact, so any caller depending on the old approximate ordering changes — no caller does; `PalaceHandle` and the retrieval layer treat the output as a ranked candidate list. Above the threshold the only change is duplicate suppression.

## Test evidence — rung 5

Every gate run in this worktree. `cargo test --workspace` deliberately not run; `--workspace` invocations use the three GUI excludes documented in the root `Cargo.toml` and `docs/reference/common-pitfalls.md` (#4982).

```
cargo fmt --check                                                        — clean
cargo check -p trusty-common --features memory-core                      — EXIT=0
cargo clippy -p trusty-common --all-targets --features memory-core -D warnings — EXIT=0
cargo test  -p trusty-common --features memory-core
    test result: ok. 925 passed; 0 failed; 13 ignored; 0 measured; 0 filtered out; finished in 19.53s
    (plus 6 integration binaries, all ok: 6, 0, 11, 4, 0 passed)
cargo check --workspace --exclude trusty-mpm-gui --exclude trusty-code-gui --exclude trusty-agents-ui
    Finished `dev` profile in 1m 10s                                     — EXIT=0
```

Direct dependents of `memory-core` (`trusty-agents`, `trusty-cto-db`, `tga`, `trusty-memory`, `trusty-mpm`):

```
cargo test -p trusty-memory   test result: ok. 606 passed; 0 failed; 4 ignored  (+5, 6, 11, 2, 3, 5, 2, 14, 1, 3, 3, 4, 2, 4, 1, 5, 15 across 22 integration binaries, all ok)
cargo test -p trusty-cto-db   test result: ok. 12 passed; 0 failed; 0 ignored
cargo test -p trusty-agents --lib  test result: ok. 3387 passed; 0 failed; 11 ignored
cargo test -p tga --lib            test result: ok. 774 passed; 0 failed; 1 ignored
cargo test -p trusty-mpm --lib     test result: FAILED. 4783 passed; 1 failed; 7 ignored  (see Baseline failures)
```

Repo gates:

```
line-cap: measured 3791 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK.
sld-lint: scanned 56 spec doc(s) + 5765 code file(s); 0 error(s), 0 warning(s)
```

The new logic went in a sibling module because `hnsw_store.rs` was at 462 of its 500 SLOC.

## Baseline failures

A bare `cargo check --workspace` fails in `trusty-code-gui` on a missing `ui/dist` — the pnpm/Svelte toolchain gap that the three documented excludes exist for, not a branch failure. The diff touches nothing in that crate:

```
$ git diff --name-only origin/main...HEAD -- crates/trusty-code-gui/
(no output)
```

`ensure_managed_config_dir_emits_the_frozen_skill_warning` failed once in the round-2 `cargo test -p trusty-mpm --lib` run. It is the documented parallel-isolation race in `docs/reference/test-ladder-baseline.md`, tracked at [#4931](https://github.com/bobmatnyc/trusty-tools/issues/4931): a non-serial test's globally-installed tracing subscriber beats this one's capturing subscriber. It passed in the round-1 run of the same command, and passes in isolation:

```
$ cargo test -p trusty-mpm --lib -- --exact core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning
test core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4790 filtered out; finished in 0.82s

$ git diff --name-only origin/main...HEAD -- crates/trusty-mpm/
(no output)
```

Change-specific gates pass; `cargo test -p trusty-mpm --lib` intermittently blocked by canonical issue #4931.

`a_dead_child_is_evicted_and_respawned` ([#5085](https://github.com/bobmatnyc/trusty-tools/issues/5085)) and `disabled_salvage_budget…` ([#5084](https://github.com/bobmatnyc/trusty-tools/issues/5084)) both passed in both runs.

## Documentation / changelog

`crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md`, category `Fixed`. `exhaustive.rs` carries the Why/What/Test module and item docs, including the measurement behind the threshold constant. No spec governs this module, so no `# Spec References` block.

## Review-finding disposition

`code-critic` round 1 returned WARN with two findings, both fixed here in commit `d11a9ad`:

- **HIGH — shadowed drawers resolved by minimum distance.** Fixed at the seam the critic named: resolve against the `VECTORS` row, not by distance, on both paths. The test now queries the stale vector, the discriminating input the first version missed.
- **MEDIUM — threshold gated on `get_nb_point()`.** Fixed: gated on the live drawer count, with a test that deletes a palace down to 6 live drawers behind 300 graph points.

Also addressed: the changelog fragment now states plainly that recall loss continues above 256 and is larger there, and the measurement section above carries the two caveats on query distribution and on the "3–10×" multiplier not being a measured constant.

The above-256 residual is tracked separately and is not in this PR's scope.

### DIFF
--- a/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
+++ b/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
@@ -10,6 +10,19 @@ Fixed
   true top-5 neighbour at 6–16 points; recall is now exact below the threshold.
   The scan is also faster than the traversal it replaces at these sizes
   (27µs vs 34µs at 64 points, 45µs vs 56µs at 128); at 256 it costs 5% more.
-  Larger indexes are unchanged (#5171)
-- `HnswStore::search` no longer returns the same drawer twice when a re-upsert
-  has left a shadow copy of its vector in the in-memory graph (#5171)
+  Palaces above 256 drawers keep the graph path and its recall loss, which is
+  larger there, not smaller — 8.4% of queries lost a true top-5 neighbour at
+  1024 points, and 1.8% lost the nearest one. That residual is unchanged by
+  this fix (#5171)
+- `HnswStore::search` now scores a re-embedded drawer against the vector
+  `VECTORS` currently holds for it. Because `hnsw_rs` cannot remove a point, a
+  re-upsert leaves the old embedding in the graph until the next palace open,
+  and the drawer was ranked by whichever copy was closer — so a query matching
+  text the drawer no longer holds came back at distance 0.0, which
+  `VectorStore::search` reports as similarity 1.0. `palace_reembed` creates
+  that state in bulk. The drawer also no longer occupies two result slots
+  (#5171)
+- `HnswStore::search` selects the exact path on the live drawer count rather
+  than the graph's point count, so deletes and re-embeds accumulating within a
+  session can no longer push a small palace back onto the approximate path
+  (#5171)
--- a/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
+++ b/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
@@ -1,0 +1,15 @@
+Fixed
+
+- `HnswStore::search` now ranks by scanning every point when the index holds
+  256 or fewer, instead of traversing the HNSW graph. The traversal returned
+  only what was reachable from its descent pivot along pruned layer-0 neighbour
+  lists, so a genuinely relevant drawer could be absent from the candidate
+  pool — and because `hnsw_rs` seeds its level RNG from OS entropy, *which*
+  drawer went missing changed on every palace open. Measured against
+  brute-force truth on real 384-dim palace embeddings, 2–4% of queries lost a
+  true top-5 neighbour at 6–16 points; recall is now exact below the threshold.
+  The scan is also faster than the traversal it replaces at these sizes
+  (27µs vs 34µs at 64 points, 45µs vs 56µs at 128); at 256 it costs 5% more.
+  Larger indexes are unchanged (#5171)
+- `HnswStore::search` no longer returns the same drawer twice when a re-upsert
+  has left a shadow copy of its vector in the in-memory graph (#5171)
--- a/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
+++ b/crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md
@@ -26,3 +26,11 @@ Fixed
   than the graph's point count, so deletes and re-embeds accumulating within a
   session can no longer push a small palace back onto the approximate path
   (#5171)
+- Below the threshold, `HnswStore::search` no longer trims its candidate list
+  before re-scoring re-embedded drawers. A drawer that has been re-embedded is
+  ranked provisionally by whichever of its two embeddings is nearer, and
+  trimming on that optimistic score let a drawer whose SUPERSEDED vector sat
+  near the query push a genuinely nearer drawer out of the results entirely —
+  the same lost-neighbour symptom this fix exists to remove, reached through
+  re-embedding rather than through the graph. `palace_reembed` puts every
+  drawer in that state (#5171)
--- a/crates/trusty-common/src/memory_core/store/hnsw_store.rs
+++ b/crates/trusty-common/src/memory_core/store/hnsw_store.rs
@@ -566,6 +566,10 @@ impl HnswStore {
         // copy is nearer, which is the defect. A search that sees the flag
         // without the shadow re-reads the already-committed `VECTORS` row and
         // gets the right answer, so over-marking costs one point read.
+        //
+        // #5171: safe only because `exhaustive_nearest` (exhaustive.rs) never
+        // truncates before `resolve_shadowed` corrects every candidate — a
+        // pre-correction cap there would reopen this race.
         if shadows_previous {
             self.shadowed.write().insert(vector_id);
         }
--- a/crates/trusty-common/src/memory_core/store/hnsw_store.rs
+++ b/crates/trusty-common/src/memory_core/store/hnsw_store.rs
@@ -332,8 +332,10 @@ pub struct HnswStore {
     /// candidate — the steady state is empty, and this costs nothing there.
     /// What: written by `upsert` when the uuid already had a mapping. Starts
     /// empty at every `open`, which is correct: `open` rebuilds the graph with
-    /// one point per live vector.
-    /// Test: `search_scores_a_re_upserted_drawer_by_its_current_vector`.
+    /// one point per live vector. Marked before the shadow point enters the
+    /// graph, so no search can observe the point without the flag.
+    /// Test: `search_scores_a_re_upserted_drawer_by_its_current_vector`,
+    /// `upsert_marks_a_shadow_before_the_graph_can_serve_it`.
     shadowed: RwLock<std::collections::HashSet<u64>>,
 }
 
@@ -558,10 +560,16 @@ impl HnswStore {
         }
         wtx.commit()?;
 
-        self.index.read().insert((vector, vector_id as usize));
+        // #5171: mark BEFORE the graph can serve the shadow point. The two
+        // steps cannot be atomic, and the asymmetry runs one way: a search
+        // that sees the shadow without the flag scores the drawer by whichever
+        // copy is nearer, which is the defect. A search that sees the flag
+        // without the shadow re-reads the already-committed `VECTORS` row and
+        // gets the right answer, so over-marking costs one point read.
         if shadows_previous {
             self.shadowed.write().insert(vector_id);
         }
+        self.index.read().insert((vector, vector_id as usize));
 
         Ok(vector_id)
     }
@@ -617,19 +625,23 @@ impl HnswStore {
             }
         }
 
-        // Over-fetch so tombstoning doesn't starve callers asking for k.
-        // 2x is sufficient in the common case; for pathological cases the
-        // caller can re-issue.
+        // Over-fetch so tombstoning doesn't starve callers asking for k. 2x is
+        // sufficient in the common case; for pathological cases the caller can
+        // re-issue. Bounds the graph arm only — the scan returns everything.
         let want = k.saturating_mul(2).max(k);
         let ef = HNSW_DEFAULT_EF_SEARCH.max(k * 2);
-        // #5171: `reverse.len()` is the LIVE drawer count. `get_nb_point()`
-        // would also count tombstones and re-upsert shadows, so a small palace
-        // could churn past the threshold mid-session and silently revert to the
-        // approximate path.
+        // #5171: `reverse.len()` is the LIVE drawer count — `get_nb_point()`
+        // also counts tombstones and shadows, so a small palace could churn
+        // past the threshold mid-session and silently revert to the approximate
+        // path. The scan returns every live candidate rather than a
+        // `want`-sized prefix: a shadowed drawer's provisional distance is
+        // `min(stale, current)`, so cutting before `resolve_shadowed` lets a
+        // superseded vector evict a true neighbour. The `out.len() >= k` break
+        // below does the bounding instead, on the corrected ranking.
         let mut raw: Vec<(u64, f32)> = {
             let index = self.index.read();
             if reverse.len() <= EXHAUSTIVE_SCAN_MAX_POINTS {
-                exhaustive_nearest(&index, query, want, &tombstones)
+                exhaustive_nearest(&index, query, &tombstones)
             } else {
                 index
                     .search(query, want, ef)
--- a/crates/trusty-common/src/memory_core/store/hnsw_store.rs
+++ b/crates/trusty-common/src/memory_core/store/hnsw_store.rs
@@ -31,7 +31,7 @@ use crate::memory_core::store::kg_store::{
 };
 
 mod exhaustive;
-use exhaustive::{EXHAUSTIVE_SCAN_MAX_POINTS, exhaustive_nearest};
+use exhaustive::{EXHAUSTIVE_SCAN_MAX_POINTS, exhaustive_nearest, resolve_shadowed};
 
 /// Default HNSW connectivity. Maps to `max_nb_connection` in `hnsw_rs`.
 ///
@@ -322,6 +322,19 @@ pub struct HnswStore {
     /// `delete`, and `compact_orphans`.
     /// Test: `vector_writes_rejected_on_snapshot` (in `vector.rs`).
     read_only: bool,
+    /// `vector_id`s that have more than one point in the in-memory graph.
+    ///
+    /// Why (#5171): `hnsw_rs` cannot remove a point, so re-upserting a mapped
+    /// uuid leaves its previous embedding in the graph under the same id. A
+    /// search that scores the drawer by that stale copy reports content the
+    /// drawer no longer holds. Knowing exactly which ids are ambiguous lets
+    /// `search` re-read only those from `VECTORS` instead of re-reading every
+    /// candidate — the steady state is empty, and this costs nothing there.
+    /// What: written by `upsert` when the uuid already had a mapping. Starts
+    /// empty at every `open`, which is correct: `open` rebuilds the graph with
+    /// one point per live vector.
+    /// Test: `search_scores_a_re_upserted_drawer_by_its_current_vector`.
+    shadowed: RwLock<std::collections::HashSet<u64>>,
 }
 
 impl HnswStore {
@@ -460,6 +473,7 @@ impl HnswStore {
             index: Arc::new(RwLock::new(index)),
             dim,
             read_only,
+            shadowed: RwLock::new(std::collections::HashSet::new()),
         })
     }
 
@@ -515,6 +529,9 @@ impl HnswStore {
         let encoded: Vec<u8> = postcard::to_allocvec(&vector.to_vec())?;
         let wtx = self.db.begin_write()?;
         let vector_id;
+        // #5171: a re-upsert leaves the old embedding in the graph under this
+        // same id, so `search` must re-read the authoritative vector for it.
+        let shadows_previous;
         {
             let mut vectors = wtx.open_table(VECTORS)?;
             let mut keys = wtx.open_table(VECTOR_KEYS)?;
@@ -525,6 +542,7 @@ impl HnswStore {
             // (immutable borrow of `keys`) is dropped before we re-borrow
             // `keys` mutably for `insert`.
             let existing: Option<u64> = keys.get(uuid)?.map(|g| g.value());
+            shadows_previous = existing.is_some();
             vector_id = match existing {
                 Some(id) => id,
                 None => {
@@ -541,6 +559,9 @@ impl HnswStore {
         wtx.commit()?;
 
         self.index.read().insert((vector, vector_id as usize));
+        if shadows_previous {
+            self.shadowed.write().insert(vector_id);
+        }
 
         Ok(vector_id)
     }
@@ -554,11 +575,12 @@ impl HnswStore {
     /// hit. Tombstoned ids are filtered out before the lookup so callers
     /// never see deleted points.
     /// What: Ranks candidates — exactly, by scanning every point, when the
-    /// index holds at most [`EXHAUSTIVE_SCAN_MAX_POINTS`]; otherwise via
-    /// `Hnsw::search(query, k, ef_search)`. Filters the result against
-    /// `DELETED_VECTORS`, then maps each surviving `vector_id` back to its UUID
-    /// via the `VECTOR_KEYS` reverse map. Returns hits sorted by ascending
-    /// distance (best first), each UUID at most once.
+    /// palace holds at most [`EXHAUSTIVE_SCAN_MAX_POINTS`] live drawers;
+    /// otherwise via `Hnsw::search(query, k, ef_search)`. Re-scores any drawer
+    /// that has a shadow copy in the graph against its `VECTORS` row, filters
+    /// the result against `DELETED_VECTORS`, then maps each surviving
+    /// `vector_id` back to its UUID via the `VECTOR_KEYS` reverse map. Returns
+    /// hits sorted by ascending distance (best first), each UUID at most once.
     ///
     /// #5171: below the threshold the graph traversal returns only the points
     /// reachable from its descent pivot along pruned layer-0 neighbour lists,
@@ -567,7 +589,8 @@ impl HnswStore {
     /// *which* drawer went missing changed on every palace open. See
     /// [`exhaustive`] for the measurement and the threshold's rationale.
     /// Test: `upsert_and_search_round_trips`, `delete_filters_results`,
-    /// `exhaustive_scan_finds_every_true_neighbour_across_rebuilds`.
+    /// `search_returns_the_exact_top_k_below_the_exhaustive_threshold`,
+    /// `search_scores_a_re_upserted_drawer_by_its_current_vector`.
     pub fn search(&self, query: &[f32], k: usize) -> Result<Vec<(String, f32)>> {
         if query.len() != self.dim {
             return Err(HnswStoreError::DimensionMismatch {
@@ -599,12 +622,14 @@ impl HnswStore {
         // caller can re-issue.
         let want = k.saturating_mul(2).max(k);
         let ef = HNSW_DEFAULT_EF_SEARCH.max(k * 2);
-        // One guard for both the size test and the search, so the branch and
-        // the work it selects observe the same graph.
-        let raw: Vec<(u64, f32)> = {
+        // #5171: `reverse.len()` is the LIVE drawer count. `get_nb_point()`
+        // would also count tombstones and re-upsert shadows, so a small palace
+        // could churn past the threshold mid-session and silently revert to the
+        // approximate path.
+        let mut raw: Vec<(u64, f32)> = {
             let index = self.index.read();
-            if index.get_nb_point() <= EXHAUSTIVE_SCAN_MAX_POINTS {
-                exhaustive_nearest(&index, query, want)
+            if reverse.len() <= EXHAUSTIVE_SCAN_MAX_POINTS {
+                exhaustive_nearest(&index, query, want, &tombstones)
             } else {
                 index
                     .search(query, want, ef)
@@ -613,6 +638,16 @@ impl HnswStore {
                     .collect()
             }
         };
+
+        // #5171: a drawer with two embeddings in the graph is scored against
+        // the one `VECTORS` holds, never against the copy it superseded.
+        let shadowed = self.shadowed.read();
+        if !shadowed.is_empty() {
+            let rtx = self.db.begin_read()?;
+            let vectors = rtx.open_table(VECTORS)?;
+            resolve_shadowed(&mut raw, &shadowed, query, &vectors)?;
+        }
+        drop(shadowed);
 
         let mut out: Vec<(String, f32)> = Vec::with_capacity(k);
         let mut emitted: std::collections::HashSet<u64> = std::collections::HashSet::new();
--- a/crates/trusty-common/src/memory_core/store/hnsw_store.rs
+++ b/crates/trusty-common/src/memory_core/store/hnsw_store.rs
@@ -30,6 +30,9 @@ use crate::memory_core::store::kg_store::{
     DELETED_VECTORS, NEXT_VECTOR_ID, VECTOR_ID_SEQ, VECTOR_KEYS, VECTORS,
 };
 
+mod exhaustive;
+use exhaustive::{EXHAUSTIVE_SCAN_MAX_POINTS, exhaustive_nearest};
+
 /// Default HNSW connectivity. Maps to `max_nb_connection` in `hnsw_rs`.
 ///
 /// Why: 16 is the recommended value from the original HNSW paper for
@@ -550,11 +553,21 @@ impl HnswStore {
     /// path is a single in-memory graph traversal plus a hash lookup per
     /// hit. Tombstoned ids are filtered out before the lookup so callers
     /// never see deleted points.
-    /// What: Calls `Hnsw::search(query, k, ef_search)`, filters results
-    /// against `DELETED_VECTORS`, then maps each surviving `d_id` back to
-    /// its UUID via the `VECTOR_KEYS` reverse map. Returns hits sorted by
-    /// ascending distance (best first).
-    /// Test: `upsert_and_search_round_trips`, `delete_filters_results`.
+    /// What: Ranks candidates — exactly, by scanning every point, when the
+    /// index holds at most [`EXHAUSTIVE_SCAN_MAX_POINTS`]; otherwise via
+    /// `Hnsw::search(query, k, ef_search)`. Filters the result against
+    /// `DELETED_VECTORS`, then maps each surviving `vector_id` back to its UUID
+    /// via the `VECTOR_KEYS` reverse map. Returns hits sorted by ascending
+    /// distance (best first), each UUID at most once.
+    ///
+    /// #5171: below the threshold the graph traversal returns only the points
+    /// reachable from its descent pivot along pruned layer-0 neighbour lists,
+    /// which on real embeddings dropped a true top-5 neighbour on 2–4% of
+    /// queries — and because `hnsw_rs` seeds its level RNG from OS entropy,
+    /// *which* drawer went missing changed on every palace open. See
+    /// [`exhaustive`] for the measurement and the threshold's rationale.
+    /// Test: `upsert_and_search_round_trips`, `delete_filters_results`,
+    /// `exhaustive_scan_finds_every_true_neighbour_across_rebuilds`.
     pub fn search(&self, query: &[f32], k: usize) -> Result<Vec<(String, f32)>> {
         if query.len() != self.dim {
             return Err(HnswStoreError::DimensionMismatch {
@@ -584,20 +597,31 @@ impl HnswStore {
         // Over-fetch so tombstoning doesn't starve callers asking for k.
         // 2x is sufficient in the common case; for pathological cases the
         // caller can re-issue.
+        let want = k.saturating_mul(2).max(k);
         let ef = HNSW_DEFAULT_EF_SEARCH.max(k * 2);
-        let raw = self
-            .index
-            .read()
-            .search(query, k.saturating_mul(2).max(k), ef);
+        // One guard for both the size test and the search, so the branch and
+        // the work it selects observe the same graph.
+        let raw: Vec<(u64, f32)> = {
+            let index = self.index.read();
+            if index.get_nb_point() <= EXHAUSTIVE_SCAN_MAX_POINTS {
+                exhaustive_nearest(&index, query, want)
+            } else {
+                index
+                    .search(query, want, ef)
+                    .into_iter()
+                    .map(|hit| (hit.d_id as u64, hit.distance))
+                    .collect()
+            }
+        };
 
         let mut out: Vec<(String, f32)> = Vec::with_capacity(k);
-        for hit in raw {
-            let id = hit.d_id as u64;
-            if tombstones.contains(&id) {
+        let mut emitted: std::collections::HashSet<u64> = std::collections::HashSet::new();
+        for (id, distance) in raw {
+            if tombstones.contains(&id) || !emitted.insert(id) {
                 continue;
             }
             if let Some(uuid) = reverse.get(&id) {
-                out.push((uuid.clone(), hit.distance));
+                out.push((uuid.clone(), distance));
                 if out.len() >= k {
                     break;
                 }
--- a/crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs
+++ b/crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs
@@ -13,10 +13,13 @@
 //! at these sizes, no more expensive. 93% of this machine's 94 real palaces
 //! hold 64 vectors or fewer.
 //! What: [`exhaustive_nearest`] evaluates the index's own `DistCosine` against
-//! every point the graph holds and keeps the closest `want` by `vector_id`, so
-//! its distances are identical to the ones the traversal would have reported
-//! for the same points. [`resolve_shadowed`] then re-scores any drawer that has
-//! more than one embedding in the graph against the one `VECTORS` holds.
+//! every point the graph holds and keeps one entry per `vector_id`, so its
+//! distances are identical to the ones the traversal would have reported for
+//! the same points. [`resolve_shadowed`] then re-scores any drawer that has
+//! more than one embedding in the graph against the one `VECTORS` holds. The
+//! scan deliberately hands back its FULL ranking: a shadowed drawer's
+//! provisional distance is optimistic, so any cut taken before re-scoring can
+//! drop a true neighbour — see [`exhaustive_nearest`].
 //! Test: `exhaustive_scan_returns_every_point_the_graph_holds`,
 //! `search_returns_the_exact_top_k_below_the_exhaustive_threshold`,
 //! `search_scores_a_re_upserted_drawer_by_its_current_vector`.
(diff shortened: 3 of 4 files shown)


bobmatnyc:
## code-critic verdict: WARN

Independent review against #5171, rung 5. Gates re-run in a detached review worktree at `8be29a49c`: `cargo test -p trusty-common --features memory-core hnsw` → `test result: ok. 17 passed; 0 failed`; `scripts/check_line_cap.sh` → `0 violations`. `hnsw_store.rs` 473 SLOC, `exhaustive.rs` 29 — both under cap, and the split is a real module boundary, not a cap dodge.

The core fix is sound and the pre-fix-path demonstration is genuine: setting `EXHAUSTIVE_SCAN_MAX_POINTS` to 0 routes into `index.search(query, want, ef)` with byte-identical arguments to the deleted pre-fix line, so the 10/10 failure is the real defect, not a different one. Consumers were checked — `store/vector.rs:618`, `retrieval/layers.rs:310,472`, `dream/cycle.rs:209` all re-sort and re-filter the output as a ranked candidate list; the "no caller depends on the old ordering" claim holds.

One HIGH, below.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs` | 70-77 | Duplicate `vector_id`s are collapsed by **minimum** distance, so a re-upserted drawer is ranked by whichever of its old and new embeddings is closer to the query — never worse than its true relevance, and reported as an exact match against superseded content. The docstring on `search_reports_a_re_upserted_drawer_once_at_its_current_vector` claims "at the distance of the CURRENT vector, not the stale copy's"; the test only passes because it queries with `current` itself, where `min(stale, current) = current` trivially. Probed on this branch: upsert `stale`, upsert `current`, then query `stale` → the drawer is returned at distance **0.0**. `vector.rs:635` maps that to score 1.0, so a re-embedded drawer clears a 0.99 near-duplicate threshold against text it no longer holds, for the whole process lifetime (the shadow only clears on the next `HnswStore::open`, which rebuilds one point per id). `palace_reembed` and the backfill path create these in bulk. | Resolve duplicates against the authoritative vector rather than by distance: when `exhaustive_nearest` sees a `vector_id` more than once, have `search` recompute that id's distance from its `VECTORS` row (it already holds a read txn). A cheaper variant is to keep the last-inserted copy, but that depends on `get_point_indexation()` iteration order tracking insert order — verify before relying on it. The graph branch needs the same treatment: `hnsw_store.rs:620`'s `emitted.insert(id)` likewise keeps the best-ranked copy. Then extend the test to query the stale vector, which is the case that discriminates. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/store/hnsw_store.rs` | 606 | The threshold is compared against `Hnsw::get_nb_point()`, which counts tombstoned points (`delete` never touches the graph) and re-upsert shadows (`upsert` inserts unconditionally at line 543). A palace with well under 256 live drawers can drift past 256 graph points mid-session and silently revert to the defective path, then heal on the next open — the fix disengages with nothing surfaced. The constant's doc acknowledges shadows but frames it as harmless; it is the recall fix silently switching off. | Compare against live count instead — `reverse.len()` is already built two statements earlier and is exactly the live-drawer count. That also makes the seam test's meaning stable. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs` | 84 | `ranked.truncate(want)` runs **before** tombstone filtering in `search`. Below the threshold the scan has every point in hand, then throws all but `2k` away; if `k` or more of the top `2k` are tombstoned, `search` returns fewer than `k` hits although exact live neighbours exist. Equal to prior behaviour, so not a regression — but it is a hole in this PR's "exact below the threshold" postcondition, and it costs nothing to close on this path. | Pass the tombstone set into `exhaustive_nearest` and filter before truncation, or skip the truncate below the threshold and let `search`'s `out.len() >= k` break do the bounding. | Fix here |
| LOW | `crates/trusty-common/changelog.d/5171-hnsw-exact-small-index-search.md` | 11 | "Larger indexes are unchanged" is where an operator reads about this fix, and it does not say that indexes above 256 still silently lose true neighbours (3.4% of queries at 257 by the PR's own post-fix table). The residual is stated honestly to reviewers in the PR body and in the constant's doc, and nowhere an operator would look. | Add one clause: recall above 256 points remains approximate and is not addressed here. | Fix here |
| MEDIUM | — | — | Recall above 256 points: 5.6% of queries at 512, 8.7% at 1024 lose a true top-5 neighbour, and the PR closes #5171 with no artifact tracking it. The author's own analysis says no `hnsw_rs` knob fixes it (four tried) — it needs a rerank pass over a wider pool or a different index. That is separable, schedulable work. | Recommend the PM decide whether this earns its own issue; the critic does not file one. | Promote |

## Measurement methodology

Holding out corpus vectors as queries is the right call and better than the synthetic fixture, with one caveat the PR does not state: a real recall query is a *question* embedded by the same model, and question embeddings sit in a somewhat different region of the manifold than document embeddings. The measured rate is an estimate for the real query distribution, not a measurement of it, and the direction of that bias is unknown. It does not touch the load-bearing conclusion, because the `ef=64`/`128`/`256` rows being byte-identical at 6–16 points is a structural fact about reachability that no query distribution changes.

The truth ranking is genuinely independent of `hnsw_rs` in the repo test — `reference_ranking` (tests.rs) is hand-written f64 cosine over the pool. The harness that produced the headline table is not in the diff and needs a live palace, so those numbers are unreproducible from the repo; that is stated in the PR, and it is the right trade, but it means the table is testimony rather than evidence.

40 graphs × 100 queries is enough to establish the effect and not enough to support the precision the "3–10×" band implies. Queries within one graph are correlated, so the effective sample is nearer 40 clusters than 4000 draws, and no error bars are given. The n=26 cell at 0.325% sitting well below n=10 at 4.35% and n=16 at 2.925% is non-monotone in a way that only graph-to-graph variance explains — read any single cell loosely. The qualitative claim (percent-level, orders above the fixture's 0.42%) is well supported; the multiplier is not a measured constant.

The "faster and more correct" combination is not suspicious here. `search` already iterates all of `VECTOR_KEYS` and `DELETED_VECTORS` from redb and allocates a `String` per drawer on every call (hnsw_store.rs:583-595), and an `ef=64` traversal over 64–128 points evaluates most of the collection anyway through two heaps and a visited map. A flat sequential pass beating that is the expected result, not a surprising one.

## Required changes

1. Fix the duplicate-`vector_id` resolution so a superseded embedding cannot set a drawer's distance, and extend `search_reports_a_re_upserted_drawer_once_at_its_current_vector` to query the stale vector — the case that tells the two rules apart.
2. Compare the threshold against live drawers rather than `get_nb_point()`.
3. Filter tombstones before truncation on the exhaustive path.
4. Note the above-256 residual in the changelog fragment.

## Notes

Not flagged: the empty-index guard is a correct early return, not a swallowed failure — an empty index has no neighbours, and `open` propagates load errors before the store exists, so `get_nb_point() == 0` cannot mean "failed to load". No new `unwrap`/`expect` and no new swallowed `Result` on any branch the diff adds or touches.

Confirmed and left alone: the author's worktree carries uncommitted regenerated UI artifacts under `crates/trusty-analyze/ui/` and `crates/trusty-memory/ui/package-lock.json`. None are in `8be29a49c`, which touches four files, all under `crates/trusty-common/`.

bobmatnyc:
## code-critic re-review of `d11a9ad` — verdict: WARN

Scoped to the new commit. Gates re-run in a detached worktree at `d11a9ad`: `cargo test -p trusty-common --features memory-core hnsw` → `test result: ok. 19 passed; 0 failed`; `scripts/check_line_cap.sh` → `0 violations` (`hnsw_store.rs` 487, `exhaustive.rs` 61).

Both fixes are real, and the shadowed-id design is better than what I asked for — re-reading the authoritative row covers the graph-returns-only-the-stale-copy case that distance comparison structurally could not. The three confirmations hold: #5179 is open and tracks the above-256 residual, the changelog states it with the 1024-point numbers, and the PR body carries both measurement caveats accurately.

Two HIGHs below. The first is the shape this commit just fixed for tombstones, surviving one step further down the same function.

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-common/src/memory_core/store/hnsw_store/exhaustive.rs` | 106 | **Truncate-before-correct, again — one step past where it was just fixed.** `exhaustive_nearest` ranks on the provisional `min(stale, current)` distance and `ranked.truncate(want)` cuts to `2k`; `resolve_shadowed` then re-scores only what survived that cut. `min` is optimistic, so a re-embedded drawer whose *superseded* vector sits near the query is promoted above a drawer whose *current* vector is a true neighbour, and evicts it before anything corrects the score. `palace_reembed` makes every drawer shadowed, which is the maximal case. This puts a genuine top-k neighbour back outside the candidate pool below the threshold — the exact defect #5171 exists to close, through a different door. Probed on this branch with 6 re-upserted decoys, k=2: the true nearest live drawer is absent and the returned pair sits at cosine distance 0.75 / 0.83. <br>`CRITIC-PROBE hits=[("85b23b5d…", 0.74971366), ("253d48b1…", 0.8267484)]` — truth `b12957ba…` not present. | Truncate after resolution, not before. Below the threshold the ranked list is at most the live count, so drop the `truncate(want)` on the scan path (or pass `usize::MAX`) and let `search`'s existing `out.len() >= k` break bound the output — `resolve_shadowed` then sees every candidate. Add the probe above as a test; it fails on this commit. | Fix here |
| HIGH | `crates/trusty-analyze/ui/dist/`, `crates/trusty-analyze/ui/package-lock.json`, `crates/trusty-memory/ui/package-lock.json` | — | **The uncommitted UI artifacts I confirmed clean last round were swept into this commit.** `d11a9ad` rewrites `crates/trusty-analyze/ui/dist/` — which `build.rs` produces with `pnpm build` and `rust-embed` compiles into the shipped binary — adds a new 1874-line **npm** `package-lock.json` beside the `pnpm-lock.yaml` the build actually uses, and drops the `libc` fields from `crates/trusty-memory/ui/package-lock.json`. None of it is reviewable (minified bundle), none of it is in the PR's declared risk surface, and the two disagreeing lockfiles in one package are a standing trap for whoever runs `npm ci` there next. | `git restore --source=8be29a49c -- crates/trusty-analyze/ui crates/trusty-memory/ui` and amend; keep the commit to the four `trusty-common` paths. | Fix here |
| MEDIUM | `crates/trusty-common/src/memory_core/store/hnsw_store.rs` | 634-639 | The graph arm still truncates before tombstone filtering: `index.search(query, want, ef)` returns `want` candidates and dead ids are dropped afterwards in the emit loop. That is the same defect this commit fixed inside the scan, left standing on the sibling arm — a palace above 256 live drawers that has deleted heavily within a session can return fewer than `k` hits, or none, while live neighbours exist. Pre-existing and acknowledged by the original over-fetch comment ("for pathological cases the caller can re-issue"), so not introduced here. | Not `Fix here`: the repair changes above-threshold behaviour this PR deliberately leaves untouched, and #5179 already owns that regime. Recommend it be folded into #5179's scope — the PM decides. | Promote |
| LOW | `crates/trusty-common/src/memory_core/store/hnsw_store.rs` | 561-563 | `upsert` inserts the shadow point into the graph, then marks the id in `shadowed`. A `search` landing between the two statements sees the shadow with no flag and scores that drawer by whichever copy is nearer — the pre-fix behaviour, for one query. The asymmetry runs the wrong way: over-marking is free (`resolve_shadowed` re-reads `VECTORS` and gets the right answer regardless), under-marking is the bug. | Swap them — `self.shadowed.write().insert(vector_id)` before `self.index.read().insert(...)`. The redb commit has already happened, so a search in the new window re-reads the committed current vector and is more correct, not less. | Fix here |

## Can the shadowed set go stale — in either direction?

No, in both directions, apart from the LOW window above.

**Under-reporting** (a shadow exists, the set doesn't know): `upsert` computes `shadows_previous` from the `VECTOR_KEYS` lookup inside the write txn, and nothing fallible sits between `wtx.commit()` and the pair of inserts — `Hnsw::insert` returns `()`. A failed commit returns `Err` before either insert, so no shadow and no entry. A `delete` followed by re-`upsert` of the same uuid is also handled correctly for the opposite reason: `delete` removes the `VECTOR_KEYS` row, so `existing` is `None`, `allocate_vector_id` hands out a *fresh* id, and the old point stays in the graph under its own tombstoned id — `shadows_previous = false` is the right answer, not a miss.

**Over-reporting** (marked shadowed but no longer is): `self.index` is written in exactly one place — the `insert` at line 561 — and is never rebuilt in place; `compact_orphans` touches redb only. Points never leave the graph, so an entry that was ever true stays true for the store's lifetime. Over-marking would be harmless anyway.

**Reset at open** is correct as claimed: `open` iterates `VECTORS`, which is keyed by `vector_id`, skips tombstones, and inserts once per row — one point per live id, zero shadows. Growth is bounded by distinct live ids (it is a set), so no unbounded accumulation.

## The tombstone filter, judged fresh

Correct, and in the right place. Skipping inside the scan — before `dist.eval`, not after — is what makes it work: it fixes the empty-result case *and* makes a dead point cost a hash lookup instead of a 384-dim distance evaluation, which is what keeps the `reverse.len()` gate honest when the graph carries a large dead overhang. Placing it in the caller could not have done either.

The one thing it does not cover is the sibling arm (MEDIUM above). Within the scan, the only other cut is `truncate(want)` — which is the HIGH.

## Required changes

1. Truncate after `resolve_shadowed`, not before, on the scan path; add the probe as a regression test.
2. Drop the `crates/trusty-analyze/ui` and `crates/trusty-memory/ui` paths from the commit.
3. Mark the id in `shadowed` before inserting the shadow point.

## Notes

Nearly flagged and cleared: the `f32::NAN` sentinel in `resolve_shadowed`. The claim that `DistCosine::eval` cannot return NaN is verified — `anndists-0.1.5` `dist/distances.rs:215` guards both accumulated norms with `if res.1 > zero && res.2 > zero`, returns `0.` otherwise, and clamps with `.max(0.)`. A NaN component also lands in the else arm, since `NaN > 0.` is false. The sentinel cannot collide with a real distance.

`search_drops_a_shadowed_candidate_whose_vector_row_is_gone` proves what it claims — removing the `VECTORS` row directly is the only reachable way to construct the state, and the drop can only come from the `None` arm. Dropping is the right call: an id with no `VECTORS` row has no authoritative embedding, and the final emit loop already drops ids missing from `reverse` for the same reason. The apparent asymmetry — a *non*-shadowed id with a missing row is still reported from its graph copy — is unreachable in production, because `compact_orphans` only removes rows whose `VECTOR_KEYS` entry is already gone, and those ids are not in `reverse`.

`search_scores_a_re_upserted_drawer_by_its_current_vector` reproduces my earlier probe exactly: same construction, queries the stale vector, and adds the stronger assertion that the reported distance equals `dist(stale, current)` to within 1e-4. Not an adjacent failure.

`resolve_shadowed` is applied to both arms, which is right. On the graph arm it corrects distances but cannot correct *selection* — a drawer whose current vector is near but whose stale copy is far may never be retrieved. That is inherent to re-scoring after approximate retrieval and belongs to #5179's regime, not here.

The cost claim holds. `if !shadowed.is_empty()` gates the whole block, so a store with no re-upserts since open pays one `RwLock` read. After a bulk `palace_reembed` the work is bounded by candidates, not by set size — at most `want` point reads per search, not one per re-embedded drawer.

Not flagged: with the gate on `reverse.len()`, scan cost is no longer bounded by the threshold — a long session that deletes thousands of drawers from a small palace scans the whole dead overhang. The tombstone skip lands before the distance evaluation, so those points cost a hash lookup each, which is why this stays a note rather than a finding.

bobmatnyc:
## Round 3 fixes pushed — `30a9b956`

Branch was force-pushed with lease; `d11a9ad` was amended to drop the UI artifacts, so the previous head no longer exists.

**HIGH (`exhaustive.rs`) — truncate-before-correct, one step past where round 2 fixed it.** `exhaustive_nearest` ranked on the provisional `min(stale, current)` distance and truncated to `2k` before `resolve_shadowed` re-scored. `min` is optimistic, so a re-embedded drawer whose *superseded* vector sat near the query was promoted above a true neighbour and evicted it before any correction ran. Fixed by dropping the `want` parameter entirely rather than widening it — the scan now returns its full ranking and `search`'s existing `out.len() >= k` break bounds the output on corrected scores. This also makes the function's own summary line ("Every point in `index`, ranked") true, which the truncate had contradicted.

Regression test `search_keeps_a_true_neighbour_a_shadowed_decoy_would_evict`: one truth drawer at 40°, six decoys that used to hold a vector at 10–15° and now hold one at 80–85°. Six exceeds `want = 2k`, so on the provisional key the decoys fill every slot. Fails against the reintroduced defect with the two decoys returned at 0.826/0.844 and truth absent; passes after.

**HIGH — UI artifacts swept into the commit.** Reverse-applied and amended. The branch diff against `origin/main` is now four `trusty-common` paths; `grep -c "/ui/"` over it returns 0.

**LOW (`hnsw_store.rs`) — shadow marked after the graph insert.** Swapped as recommended. Regression test `upsert_marks_a_shadow_before_the_graph_can_serve_it` holds the index write lock so the re-upsert parks at `index.insert`, then polls `shadowed` while parked — exactly the state a search in that window observes. Deterministic both directions.

## Two things the review did not flag

**A cost property changed.** Round 2 documented the shadowed work as "bounded by candidates, not by set size — at most `want` point reads per search." Removing the truncate makes it one point read per *live* shadowed candidate, up to the live count rather than `2k`. After a bulk `palace_reembed` that is ~256 reads instead of ~10. Judged the right trade — it is bounded by the gate selecting this path, on a path that already evaluates a distance against every point — and documented on `resolve_shadowed`. Flagging it as a real property change, not a free fix.

**Fix order mattered.** The LOW fix is only safe because of the HIGH one. Marking `shadowed` before the graph insert means a search in that window sees a flagged id whose only graph copy is the old vector at the old distance. Under truncate-first, that id could be cut before `resolve_shadowed` reached it — so applying the LOW alone would have introduced a selection regression.

**The promoted MEDIUM is untouched.** The graph-arm truncate-before-tombstone-filter finding sits in #5179's above-256 regime, which is out of scope here.

## Gates — rung 5

```
cargo fmt --check && cargo check -p trusty-common --features memory-core
  && cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings
  && scripts/check_line_cap.sh                                    → EXIT=0
line-cap: 3791 tracked .rs file(s); 4 allowlisted, 0 violations — OK.

scripts/check_changelog_fragment.sh                               → EXIT=0

  -- --skip default_model_matches_sentence_transformers_reference  → EXIT=0
test result: ok. 926 passed; 0 failed; 13 ignored; 1 filtered out; finished in 19.03s

cargo test -p trusty-memory                                        → EXIT=0
518 passed; 0 failed; 4 ignored (lib) + 24 integration binaries, 0 failed
```

926 is non-zero and 21 of them are `hnsw_store::tests`, including both new regression tests. The feature flag is load-bearing: the bare `cargo test -p trusty-common` runs 294 tests and exits 0 while matching **zero** `memory_core::` and **zero** `hnsw_store::tests` — it would have gone green regardless of what broke here.

## One skipped test, proven pre-existing

`default_model_matches_sentence_transformers_reference` **hangs** on this machine — 0.0% CPU, parked in `kevent`, no sockets, no cache writes. It holds `ENV_LOCK` across the window, so every other `embedder` test queues behind it and the suite never terminates. Alone with `--test-threads=1` it still hangs (`EXIT=124`). Canonical issue is #4741, which documents the panic-plus-poison signature; this is a new hang signature on the same test and lock. It blocks the full `trusty-common` and `trusty-memory` suites on this machine for any branch.

Two `#[ignore]`d `memory_core` tests also fail under `--include-ignored`. Not assumed pre-existing — `hnsw_store.rs` and its `tests.rs` were reverted to `origin/main` and both still failed identically. `git diff --name-only origin/main...HEAD -- crates/trusty-common/src/embedder/` returns zero files.

Nothing was `#[ignore]`d, `cfg`-gated, or `--exclude`d to reach green.

bobmatnyc:
## code-critic round 4 re-review of `30a9b956` — verdict: APPROVE

Independent review in a dedicated worktree at `30a9b956` (not `.claude/worktrees/fix-5171`). Scoped to what changed since `d84f679` (round 2's head): the truncate-before-correct fix in `exhaustive.rs`, the mark-order swap in `hnsw_store.rs::upsert`, and the UI-artifact removal.

## Round 3's own fixes, verified independently, not just re-read

I didn't take the "fixed" claims on faith — I reverted each one in isolation on top of `30a9b956` and re-ran the named regression test to confirm it fails for the defect it claims to catch, then restored the tree (`git diff --quiet` confirmed clean before continuing):

- Reintroduced `ranked.truncate(want)` before `resolve_shadowed` in `exhaustive_nearest` → `search_keeps_a_true_neighbour_a_shadowed_decoy_would_evict` fails exactly as described: the six decoys evict the truth drawer, hits become `[0.826, 0.844]` with the true neighbour absent. Restored → passes.
- Reverted the mark-order swap in `upsert` (graph insert before `shadowed.write().insert`) → `upsert_marks_a_shadow_before_the_graph_can_serve_it` fails on "the shadow flag must be set before the shadow point can be reached through the graph." Restored → passes.
- `git diff --name-only origin/main...HEAD` is exactly the four `trusty-common` paths the PR claims — no UI artifacts.

Both tests are sound: `search_keeps_a_true_neighbour_a_shadowed_decoy_would_evict` cannot pass by luck, since without the fix the six decoys structurally fill both of `want=2k`'s slots on the provisional key. `upsert_marks_a_shadow_before_the_graph_can_serve_it`'s lock-parking construction is real, not simulated — the write lock on `index` genuinely blocks the spawned thread's `self.index.read()`, and in the pre-fix ordering the thread would block there *before* ever reaching the mark step, so `marked` would correctly stay `false` for the full 2s poll window rather than timing out by chance.

## The five scrutiny items

**1. Cost-change claim.** Verified as stated. `resolve_shadowed` runs over `candidates`, which on the scan path is `exhaustive_nearest`'s full live ranking — capped at `EXHAUSTIVE_SCAN_MAX_POINTS` (256) by the branch gate itself, since `exhaustive_nearest` already excludes tombstones and dedupes by `vector_id`. Worst case (bulk `palace_reembed`, every live drawer shadowed) is one point read per live drawer, up to 256, versus round 2's `want`-bounded ~10. Correctly identified as a real property change, not a free fix, and it's an acceptable trade — the branch is gated at ≤256 live drawers specifically, and the scan already evaluates a distance against every point in that regime.

**2. Fix-ordering dependency.** Verified independently, and it's correct. During the mark-then-insert race window (`shadowed` flagged, graph insert not yet visible), a concurrent search's candidate list for that id can only come from the graph's *old* point (the new one isn't there yet) — but `resolve_shadowed` reads `VECTORS`, which already has the new value committed, and overwrites that candidate's distance with the correct one. That correction only reaches the candidate if it survived to `resolve_shadowed` uncut, which is exactly what round 3's HIGH fix guarantees. Applying the LOW fix without the HIGH fix would have left a window where a query could lose a true neighbour rather than just mis-score one.

However: **this coupling is not documented at either site.** `upsert`'s mark-before-insert comment (`hnsw_store.rs`) explains the search-observes-shadow-without-flag asymmetry but never states it depends on `exhaustive_nearest` returning an untruncated ranking. `exhaustive_nearest`'s and `resolve_shadowed`'s docs (`exhaustive.rs`) explain why truncation must happen after correction but never mention that `upsert`'s concurrency safety also depends on it. `git grep -n "only safe\|depends on"` over both files returns nothing relevant. Given this file's history — three straight rounds where a fix at one layer had the identical defect shape resurface one layer away — a future edit to either function in isolation (e.g., re-adding a `want` cap to the scan for performance) has no comment steering it away from reopening this exact race.

**3. Regression tests.** Both verified above to fail for the right reason, not to pass by construction.

**4. Concurrency.** `upsert`'s two mutable operations (`shadowed.write().insert`, `index.read().insert`) are on independent locks, never held simultaneously, so no new deadlock surface. redb's single-writer guarantee serializes concurrent `upsert`s on the same uuid, so `shadows_previous`'s read-then-decide is not itself racy. `search` reads `VECTOR_KEYS`/`DELETED_VECTORS` in one txn, the live graph next, and `VECTORS` in a second txn for shadow correction — three different points in time under concurrent writes, but this cross-table non-atomicity predates this PR (unchanged from round 2, itself twice-reviewed) and isn't something round 3 touched.

**5. Above-256 / tombstoned / shadowed / empty regressions.** None found on the ≤256 path. `exhaustive_nearest` filters tombstones before evaluating distance (unchanged from round 2). The empty-index guard (`get_nb_point() == 0`) is deliberately on the raw graph-point count, not the live-drawer count used for threshold selection — different purpose (avoiding `hnsw_rs`'s panic on an empty iterator), correctly so. The graph arm's own truncate-before-shadow-correct (which has the same shape, one layer up, in the >256 regime) is pre-existing since round 2 and belongs to #5179, per the task scope — not re-flagged.

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-common/src/memory_core/store/hnsw_store.rs` / `hnsw_store/exhaustive.rs` | `HnswStore::upsert`, `exhaustive_nearest`, `resolve_shadowed` | The LOW fix's correctness depends on the HIGH fix (verified above), but neither function's doc comment says so. A future change to either side in isolation — most plausibly re-adding a `want` cap to `exhaustive_nearest` for cost reasons now that resolve_shadowed can read up to 256 points per search — would silently reopen the mark-before-insert race's failure mode: a true neighbour dropped during a re-upsert, not just mis-scored. | One cross-reference line in each doc: on `upsert`'s mark-before-insert comment, note it depends on `exhaustive_nearest` never truncating before `resolve_shadowed`; on `exhaustive_nearest`'s "why return everything" paragraph, note `upsert`'s concurrency safety also depends on it. | Fix here |

## Gates, reproduced independently

```
cargo fmt --check && cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings \
  && scripts/check_line_cap.sh                                                          → EXIT=0
line-cap: 3791 tracked .rs file(s); 0 violations. hnsw_store.rs measured at 487 SLOC by
hand (944 raw lines − 406 comment lines − 51 blank lines), under cap on its own merits —
not on the ratchet allowlist (`grep -i hnsw .line-cap-allowlist.tsv` → empty).

cargo test -p trusty-common --features memory-core -- --skip default_model_matches_sentence_transformers_reference
  test result: ok. 926 passed; 0 failed; 13 ignored; 0 measured; 1 filtered out
  21 hnsw_store::tests, including both new regression tests — matches the PR's claim exactly.

cargo test -p trusty-common   (bare, no feature — confirms the feature flag is load-bearing)
  test result: ok. 294 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
  0 hits for "hnsw_store::tests" — the bare run would go green regardless of what broke here.
```

`git diff origin/main...HEAD | grep -E '#\[ignore|#\[cfg|--exclude'` → no output. Nothing was ignored, cfg-gated, or excluded to reach green. `default_model_matches_sentence_transformers_reference`'s exclusion is scoped correctly — the PR's diff (`git diff --name-only origin/main...HEAD`) is exactly the four `hnsw_store*`/changelog paths, none in `embedder/`.

## Verdict

Zero CRITICAL, zero HIGH. One MEDIUM (documentation gap on a real, verified invariant — not a live bug). **APPROVE.**

PR #5183, opened 2026-08-08
fix(trusty-memory): shape the recall query to the embedder window (#4972)

### LINKED ISSUE #4972: over half of hook queries are silently truncated at the embedder's 512-token window

bobmatnyc:
Closes https://github.com/bobmatnyc/trusty-tools/issues/4972

**Split (2026-08-08):** the [#5038](https://github.com/bobmatnyc/trusty-tools/issues/5038) tag-rendering half moved to **[#5201](https://github.com/bobmatnyc/trusty-tools/pull/5201)** and is no longer in this diff. That half drew zero findings across two critic rounds; this half carries the outstanding ones, and [#5187](https://github.com/bobmatnyc/trusty-tools/issues/5187) is sequenced to reuse the whole-unit truncation from *this* PR. The clean work should not sit on the hard work's dependency chain. `format.rs` here is back to its `origin/main` state and the two diffs do not overlap.

## Defect B — [#4972](https://github.com/bobmatnyc/trusty-tools/issues/4972): the query was cut inside the embedder, unannounced

`mod.rs:473` fed `user_prompt` straight into `fetch_palace_recall` with no length awareness. The embedder cut it at its window and told nobody.

### The window is real and it is 512

Not read off a config — probed. Recalling a fixed distinctive suffix behind a growing filler prefix against the live daemon:

```
prefix_tok  total_tok  top_score
         0         23     0.4675   <- target drawer retrieved
       456        479     0.2995
       532        555     0.2533
       608        631     0.2533
       760        783     0.2533
       912        935     0.2533
      1216       1239     0.2533
      1824       1847     0.2533
      2432       2455     0.2533
```

Scores vary through 456 tokens and then plateau at *exactly* 0.2533 across a 4.5× range of input length. The embedding stops changing: everything past the window is invisible.

### What the window is spent on

Running the model's own `tokenizer.json` vocabulary (WordPiece greedy-longest-match) over the corpus:

- **52.0%** of prompts exceed 512 tokens (the issue said 51.5%).
- **65.3%** open with `<task-notification>` (the issue said 63%).
- The envelope head — `<task-id>`, `<tool-use-id>`, `<output-file>` (an absolute path), `<status>`, `<note>` — costs a **median 253 tokens, 49% of the whole window**, before any payload begins.

### Strategy: strip the envelope, then whole units, then say so

Against the ruling's part 2 ("the ceiling should rise, not truncation be tuned"): we cannot raise the model's 512, but stripping 253 tokens of framing raises the *effective* ceiling for free — the same window now carries payload. Measured: prompts that fit whole go from **46.1% → 55.3%** corpus-wide, and **18.3% → 32.4%** for task-notifications, with no tuning constant involved.

Against part 3 ("whole units or nothing"): when the stripped prompt is still over budget, the hook keeps whole leading lines, falling back to whole words when one line busts the budget alone. Never a partial unit. Head-first because after the strip the payload's opening is its thesis.

Against part 4 ("withholding is allowed; withholding silently is not"): every reduction lands on the enriched-prompt log line as a `recall_query` object (`original_tokens`, `sent_tokens`, `budget_tokens`, `envelope_stripped`, `units_dropped`) plus a `tracing::warn!`. The truncation rate is now one `jq` filter over the same corpus that produced the 52% figure.

**Chunk-and-max-pool rejected**, stated rather than ignored: it needs either N embedder round trips inside the 1.5 s `BODY_DEADLINE` or a new server-side recall surface, and the ruling steers at raising the ceiling, not at more retrieval machinery.

### The token estimator is calibrated, not guessed

An *under*-estimate is the failure that matters — it hands the embedder a query it then cuts silently, which is the defect. Against the real tokenizer over 1,500 sampled prompts:

| estimator | est/true p01 | p50 | underestimates |
|---|---|---|---|
| `pieces + ceil(len/3)` | 0.99 | 1.25 | **1.1%** |
| `pieces + ceil(len/4)` | 0.84 | 1.08 | 15.6% |
| `chars/4` (the repo's Bedrock-cache heuristic) | 0.17 | 0.73 | 94.5% |

`/3` is the only divisor that essentially never underestimates, at the price of spending ~410 of the 512 tokens rather than risking an overrun. `chars/4` answers a different question (is a prefix big enough to cache, where over-estimating is the harmful direction) and is wrong for this one.

The budget is overridable with `TRUSTY_MEMORY_PROMPT_QUERY_TOKENS`; setting it well above the real window restores the previous behaviour exactly.

The KG triple selector still sees the whole prompt — it matches literal words, has no token window, and narrowing it would be scope the fix does not need.

### The HIGH: the estimator under-reported its own cut

The critic built a WordPiece reference against the vocabulary the daemon loads and ran `piece_tokens` against it. It charged every alphanumeric run `ceil(len/3)` regardless of script, undercounting the real tokenizer 2–5× on CJK (0.34), Korean (0.18), Cyrillic (0.39), Greek (0.41), hex digests (0.42), base64 (0.50) and UUIDs (0.61).

That was worse than a wrong number. A 1520-character Chinese prompt estimated 509 tokens, passed through untouched, lost 66% of itself inside the embedder — and the `recall_query` metric this PR introduces then reported `units_dropped: 0, envelope_stripped: false`. The `jq` filter `config.rs` documents as the way to measure the post-fix truncation rate would have answered wrong in the reassuring direction. The loss was not new; the alarm asserting there was none was.

Charging is now per run class, every branch erring high:

| run | charge | why |
|---|---|---|
| one CJK codepoint | 1 token | `handle_chinese_chars: true` space-pads every Han character before pre-tokenization, so it *is* one token — arithmetic, not estimate |
| any run containing non-ASCII | 3 per char | worst measured is Korean at ~2.4 tokens/char; rounded up from the worst observed |
| ASCII run containing a digit | 1 per char | hex, base64, UUIDs, JWTs measured 0.42–0.61 against `ceil(len/3)` |
| ASCII letters only | `ceil(len/3)` | the calibrated Latin case |
| `_` and other punctuation | 1 | `BertPreTokenizer` splits on `_`; treating it as a word character undercounted `snake_case` identifiers (0.79) |

Re-validated against the same reference implementation:

```
case                         old     new    true   old/t   new/t
plain english                 74      77      44    1.68    1.75
CJK zh                        99     292     292    0.34    1.00
Japanese                      94     566     278    0.34    2.04
Korean                        98     722     638    0.15    1.13
Russian                      194    1538     458    0.42    3.36
Greek                        170    1274     410    0.41    3.11
Arabic                       146    1106     370    0.39    2.99
Hebrew                       138     962     322    0.43    2.99
hex digests                  338     962     695    0.49    1.38
base64                       332     962     644    0.52    1.49
uuids                        512    1078     834    0.61    1.29
JWT                           48     134      94    0.51    1.43
mixed zh+en                  222     282     182    1.22    1.55
long identifiers             602     802     762    0.79    1.05
rust code                   1022    1082     942    1.08    1.15
emoji run                    152     152     152    1.00    1.00
version numbers              250     266     210    1.19    1.27
file paths                   442     442     412    1.07    1.07

UNDERESTIMATES with new estimator: NONE

real Latin corpus n=600
  old  p01=0.99 p50=1.23 p95=1.57  underestimates 1.3%
  new  p01=1.00 p50=1.38 p95=1.60  underestimates 0.0%
```

The cost is real and stated: the median Latin over-charge rises 1.23 → 1.38, so we spend ~370 of the 512 tokens instead of ~410, and Cyrillic/Greek are over-charged ~3×. That is the direction the critic asked for — an over-estimate trims a little more and says so; an under-estimate re-creates the silent cut.

`token_estimate_never_underestimates` is the deliverable, not the tweak: nine inputs, each carrying its true WordPiece count measured against the model's own `tokenizer.json`, asserting `estimate >= true` for all of them. `cjk_is_one_token_per_char` pins the arithmetic case and the 1520-character boundary directly.

The doc at `query.rs:85–88` claimed a property of the corpus as a property of the estimator. It now states exactly what `CHARS_PER_SUBTOKEN` guarantees — runs of ASCII letters, on a sample containing no CJK, no Cyrillic and no hex digests — and points at the branch that handles the rest.

### The four MEDIUMs

1. **Text after `</task-notification>` was discarded uncounted.** A user appending an instruction after the notification block wrote the most intent-bearing text in the prompt. It is now appended to the kept payload. `envelope_strip_keeps_text_after_the_envelope`.
2. **An unbreakable oversized first line produced an empty query.** A base64 blob, `data:` URI or JWT packs zero whole words, and an empty query makes `fetch_palace_recall` return zero drawers — a regression against pre-fix prefix recall. `pack_chars` now fills the budget with a character prefix as a last resort, still counted. Written failing first: `unbreakable_oversized_token_still_yields_a_query`. `pack_chars_matches_the_estimator` guards the incremental run accounting against drifting from `piece_tokens`.
3. **The all-envelope / empty-payload branch was untested.** `envelope_with_no_payload_passes_through_whole` pins it, and the deliberate choice it encodes: pass through rather than emit an empty query, because noisy recall beats none, and the budget packer still bounds what goes out.
4. **`MAX_RENDERED_TAGS` argued from bytes, enforced a count.** Four 55-character tags render 246 characters while satisfying the cap; the old test's 7-character `topic-N` fixtures could not fail however wrong the cap was. New `MAX_RENDERED_TAG_CHARS` bounds the suffix directly, the count test uses real tag shapes from the live palace, and `injection_caps_rendered_tag_bytes` pins the budget with 55-character fixtures.

Untouched, as the critic asked: `is_creator_tag`, the `prompt_log/` change, `pack_units`' arithmetic, the `Closes` targets, and the two changelog fragments.

### Original round — 6 of 11 new tests fail with the fixes reverted

Correcting the earlier "6 of 7": there were **eleven** new tests in the original round, not seven, and **five** pass under revert, of which only one was disclosed. The five are `envelope_strip_recovers_the_payload`, `envelope_strip_tolerates_a_cut_envelope`, `short_query_passes_through_untouched`, `token_estimate_never_splits_a_unit` and `configured_query_budget_clamps_to_bounds`. All five target helpers with no pre-fix counterpart — reverting `shape_recall_query` to return the prompt verbatim leaves those functions defined and still correct in isolation, so only the six tests exercising the shaped *outcome* can go red. Expected for new-helper unit tests; the omission was in the reporting.

The critic's independent run, which is the number to trust: baseline **43 passed / 0 failed** with `--all-features`, revert **37 passed / 6 failed**. Two of those six were the tag tests now living on [#5201](https://github.com/bobmatnyc/trusty-tools/pull/5201).

```
---- prompt_context_logs_recall_query_shape stdout ----
fixture must exceed the budget; got RecallQueryShape { original_tokens: 0,
sent_tokens: 0, budget_tokens: 0, envelope_stripped: false, units_dropped: 0 }
```

### Review round — 5 of 7 new tests fail against the pre-review shape

Reverted `piece_tokens` to the single-divisor form, removed the trailing-text capture and the `pack_chars` last resort.

```
running 7 tests
test cjk_is_one_token_per_char ... FAILED
test envelope_strip_keeps_text_after_the_envelope ... FAILED
test envelope_with_no_payload_passes_through_whole ... ok
test pack_chars_matches_the_estimator ... ok
test token_estimate_never_underestimates ... FAILED
test injection_caps_rendered_tag_bytes ... FAILED
test unbreakable_oversized_token_still_yields_a_query ... FAILED

---- cjk_is_one_token_per_char ----
assertion `left == right` failed: each Han codepoint is its own token
  left: 3
 right: 7

---- token_estimate_never_underestimates ----
`chinese han` underestimates: estimate 99 < true 292. An underestimate hands the
embedder a query it cuts silently while RecallQueryShape reports no loss — always err high.

---- envelope_strip_keeps_text_after_the_envelope ----
user text after the envelope must survive; got:
the agent finished the sweep

---- unbreakable_oversized_token_still_yields_a_query ----
an unbreakable oversized token must not produce an empty query — that returns zero
drawers, worse than the truncation it replaced

test result: FAILED. 2 passed; 5 failed; 0 ignored; 0 measured; 621 filtered out
```

`injection_caps_rendered_tag_bytes` in that run is the tag test, now on [#5201](https://github.com/bobmatnyc/trusty-tools/pull/5201) — four of the five failures above belong to this PR.

Being precise about the two that pass, rather than counting them as a clean baseline: `envelope_with_no_payload_passes_through_whole` pins a branch whose behaviour that round did not change — it is a coverage addition (the finding asked for the test, not a fix). `pack_chars_matches_the_estimator` passes only because the partial revert removed the *call* to `pack_chars` while leaving the function defined; against a true pre-review tree the function does not exist. Neither is evidence, and neither is claimed as such.

## Outstanding

Two HIGHs from the latest critic round are **not** addressed here. This push is the split only, by instruction. The round-3 fix lands separately.

## Gates

Scoped per the build-concurrency limits in force — `CARGO_BUILD_JOBS=2`, no workspace build, no Tauri frontend build.

```
( CARGO_BUILD_JOBS=2 cargo fmt --check \
  && CARGO_BUILD_JOBS=2 cargo clippy -p trusty-memory --all-targets -- -D warnings \
  && CARGO_BUILD_JOBS=2 cargo test -p trusty-memory \
  && CARGO_BUILD_JOBS=2 cargo test -p trusty-common ) ; EXIT=0

cargo test -p trusty-memory  — 620 passed; 0 failed; 4 ignored
cargo test -p trusty-common  — 294 passed; 0 failed; 6 ignored
```

620 against 624 before the split: exactly the four tag tests, gone to [#5201](https://github.com/bobmatnyc/trusty-tools/pull/5201). The 13 query-shaping tests are intact.

Repo gates, run after `git add`:

```
line-cap:          measured 3800 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK.
changelog-fragment gate:     scanned 6 changed path(s); all 1 crate(s) with source changes are recorded.
```


bobmatnyc:
## Verdict: WARN

Adversarial review from an independent worktree (`.claude/worktrees/critic-5183` off this branch). I re-ran the red proof myself and attacked the estimator against the model's own `tokenizer.json` rather than taking the calibration table on trust.

### What I verified myself vs. took on your word

| Claim | Status |
|---|---|
| Red proof — tests fail with the fixes reverted | **Verified, with a correction.** I reverted `render_tags` to render every tag and `shape_recall_query` to return the prompt verbatim, test files untouched, and ran `CARGO_BUILD_JOBS=2 cargo test -p trusty-memory --all-features prompt_context`. **6 failed, 37 passed.** The 6 are exactly the ones you list, and they cover both defects including the two e2e tests. The proof holds. |
| "6 of 7" | **Wrong denominator.** There are 11 new tests, not 7. Five pass under revert, not one — see finding 5. |
| Estimator "underestimates 1.1% of prompts" | **False outside Latin-script prose.** See finding 1; measured, not argued. |
| Corpus statistics (53.0% tag bytes, 52.0% over window, 253-token envelope head, 0.2533 score plateau) | Taken on your word. I have no access to your `enriched-prompts.*.jsonl`. |
| `is_creator_tag` is the right filter | Verified. `attribution.rs:400` is `starts_with("creator:")`, the namespace is reserved, and `is_creator_tag_detects_namespace` already pins `!is_creator_tag("creatorx")`. Correct predicate, correctly reused. No finding. |
| `src/prompt_log/` changes are safe | Verified. `RecallQueryShape` is a plain data struct; `with_recall_query` is a pure setter; `skip_serializing_if` matches the two `Option` fields already on `PromptLogEntry`. No panic, no I/O, no blocking on the logging path. No finding. |
| Repo gates | Verified independently: `check_line_cap.sh` → `3800 tracked .rs file(s); 4 allowlisted, 0 violations — OK` (`query.rs` is inside the 500 cap). No `unwrap()`/`expect()` outside `#[cfg(test)]` in the diff. Two changelog fragments, one `Fixed` category each. `Closes` targets #5038 and #4972, both labelled `bug`; #5037 (the ruling) is correctly not closed. |

## Findings

|---|---|---|---|---|---|
| HIGH | `crates/trusty-memory/src/commands/prompt_context/query.rs` | 95, 321–336 | `piece_tokens` underestimates the real tokenizer by 2–5× on CJK, Cyrillic, Greek, hex digests, base64 and UUIDs. The new `recall_query` metric then reports `units_dropped: 0` while the embedder silently drops most of the query — the alarm this PR introduces reports healthy while the loss occurs. | Detect the scripts the estimator cannot model and charge them correctly: 1 token per CJK codepoint (the tokenizer's `handle_chinese_chars: true` guarantees this), and ~1 token per 1.5 chars for non-Latin alphabetic runs. Failing that, clamp: if `text.chars().any(non_ascii)`, fall back to `chars().count()` as the estimate. | Fix here |
| MEDIUM | `crates/trusty-memory/src/commands/prompt_context/query.rs` | 215–230 | The envelope strip keeps only `<summary>` and `<result>` inner text. Anything the user wrote *after* `</task-notification>` is discarded from the recall query, and `units_dropped` stays 0 so the count does not reflect it. | Append the text following `</task-notification>` to `kept` before joining, or count what was discarded into `units_dropped`. | Fix here |
| MEDIUM | `crates/trusty-memory/src/commands/prompt_context/query.rs` | 273–278 | The word fallback only ever looks at `lines[0]`. When the first line is a single unbroken token over the budget (base64 blob, `data:` URI, JWT, minified bundle), `pack_units` takes zero words, the query is empty, and `fetch_palace_recall` (`fetch.rs:59`) returns zero drawers. Pre-fix that prompt got prefix-based recall; now it gets none. No test covers this outcome. | If the word fallback yields nothing, keep a hard character prefix of `budget_tokens * CHARS_PER_SUBTOKEN` rather than sending an empty query — the whole-unit rule cannot be honoured here and an empty query is strictly worse than the truncation it replaces. Add the regression test. | Fix here |
| MEDIUM | `crates/trusty-memory/src/commands/prompt_context/query.rs` | 226–228 | No test for the all-envelope / empty-payload case: an envelope whose `<summary>` and `<result>` are absent or blank returns `None`, so the *entire* framing head — task ids and absolute paths — becomes the recall query. That is the exact input the fix exists for, and it is the untested branch. | Add a test pinning the `None` fallback and decide deliberately whether the framing head should be sent at all. | Fix here |
| MEDIUM | `crates/trusty-memory/src/commands/prompt_context/format.rs`, `tests.rs` | 40, 1278 | `MAX_RENDERED_TAGS` is justified by a byte rule ("a label must never outweigh its payload", 220) but enforces a *count*. Four 55-char tags exceed 220 bytes. `injection_caps_rendered_tag_count` asserts `rendered.len() < DRAWER_PREVIEW_CHARS` using 7-char `topic-N` fixtures, so it can never catch that. | Either cap bytes as well as count, or change the fixture to realistic long tags so the assertion has teeth. `DRAWER_PREVIEW_CHARS` is also a char count (`format.rs:232`) compared against `.len()` bytes. | Fix here |
| LOW | PR body | — | "6 of 7 new tests fail under revert" understates it. There are 11 new tests; 5 pass under revert (`envelope_strip_recovers_the_payload`, `envelope_strip_tolerates_a_cut_envelope`, `short_query_passes_through_untouched`, `token_estimate_never_splits_a_unit`, `configured_query_budget_clamps_to_bounds`), only one of which is disclosed. All five legitimately target new helpers with no pre-fix counterpart — the omission is in the reporting, not the testing. | State 6 of 11 and list all five. | Parent |

## Finding 1 in detail — the estimator

I built a WordPiece reference against the vocabulary this daemon actually loads (`~/.cache/fastembed/models--Xenova--all-MiniLM-L6-v2/.../tokenizer.json`, 30522 entries, `normalizer: BertNormalizer{clean_text, handle_chinese_chars: true, lowercase: true}`, `BertPreTokenizer`, `max_input_chars_per_word: 100`) and ran `piece_tokens` from `query.rs:321` against it verbatim.

```
case                         est    true   ratio  verdict
plain english                 24      16    1.50  ok
CJK zh (600 chars)           196     582    0.34  UNDERESTIMATE
Japanese kanji+kana          172     512    0.34  UNDERESTIMATE
Korean hangul                182    1037    0.18  UNDERESTIMATE
Russian cyrillic             392    1007    0.39  UNDERESTIMATE
Greek                        362     887    0.41  UNDERESTIMATE
hex digests                  482    1142    0.42  UNDERESTIMATE
base64 words                 482     962    0.50  UNDERESTIMATE
uuids                        682    1122    0.61  UNDERESTIMATE
mixed zh+en                  422     542    0.78  UNDERESTIMATE
long identifiers             602     602    1.00  ok
minified js                  722     602    1.20  ok
rust code block             9053    7194    1.26  ok
emoji run                    302       3  100.67  ok
```

The CJK case is not a sampling artefact, it is arithmetic. `handle_chinese_chars: true` pads every CJK codepoint with spaces before pre-tokenization, so *n* Han characters cost *n* tokens. `piece_tokens` treats them as one alphanumeric run and charges `ceil(n/3)`. A guaranteed 3× undercount, every time.

The consequence, at the boundary:

```
Chinese   largest input the shaper passes UNTOUCHED: 1520 chars, est= 509 (<=512) but TRUE=1522 -> embedder silently drops 66%
Japanese  largest input the shaper passes UNTOUCHED: 1525 chars, est= 511 (<=512) but TRUE=1527 -> embedder silently drops 66%
Russian   largest input the shaper passes UNTOUCHED: 1534 chars, est= 496 (<=512) but TRUE=1328 -> embedder silently drops 61%
hex       largest input the shaper passes UNTOUCHED: 1575 chars, est= 506 (<=512) but TRUE=1199 -> embedder silently drops 57%
```

Two things follow.

The doc comment at `query.rs:85–88` — "`3` is the only divisor that essentially never does that (p01 = 0.99, so even the worst case is within 1%)" — is a property of your corpus, not of the estimator. Stating it as a property of the estimator is what makes the rest of the design safe-looking. It needs to say the calibration is Latin-script prose and code only.

More seriously, this is the fail-open shape. Pre-fix, a 1520-character Chinese prompt was cut at the embedder and nothing recorded it. Post-fix it is cut at the embedder, nothing recorded it, **and** the log line now affirmatively says `units_dropped: 0, envelope_stripped: false`, i.e. `reshaped() == false`, no `tracing::warn!`. Anyone running the `jq 'select(.recall_query.units_dropped > 0)'` filter that `config.rs` documents as the way to measure the post-fix truncation rate gets an answer that is wrong in the reassuring direction. The loss is not new; the alarm that says there is no loss is.

I am holding this at HIGH rather than CRITICAL because no data is destroyed and recall for these prompts is no worse than it is on `main` today. It is the new metric that is wrong, not the stored state.

## Required Changes

1. Fix `piece_tokens` for non-Latin scripts, or scope the estimator's claim and make `shape_recall_query` fail closed when it cannot estimate (treat any prompt containing non-ASCII alphabetic runs as `chars().count()` tokens). Add a test with a CJK prompt asserting `sent_tokens` reflects one token per Han character.
2. Correct the `query.rs:85–88` doc comment to say the calibration corpus is Latin-script prose and code.
3. Stop discarding text that follows `</task-notification>` (finding 2), or account for it in `units_dropped`.
4. Do not send an empty recall query when the word fallback packs nothing (finding 3); add the regression test.
5. Add the all-envelope / empty-payload test (finding 4).
6. Give `injection_caps_rendered_tag_count` a fixture with realistic long tags, or cap bytes as well as count (finding 5).

## Notes

- The parts I attacked hardest and could not break: `pack_units` arithmetic is genuinely exact under `estimate_tokens` (whitespace separators cost zero, so the per-unit sum is additive across the join — I confirmed this holds for both the `"\n"` and `" "` separators); the `clamp_query_budget` bounds are fully covered; `INJECTION_BYTE_CAP` interacts correctly with the tag cap (fewer tag bytes means more drawers survive the 8 KiB truncation, strictly better); the `<summary>`-before-`<result>` ordering in `ENVELOPE_KEEP` is stable regardless of their order in the source envelope, which is what you want.
- Deliberately not flagged: the envelope strip firing even when the prompt already fits the budget. It drops `<note>`/`<status>` boilerplate that carries no user intent, and that is a defensible design call rather than a defect.
- Verdict is WARN, not BLOCK: one HIGH, no CRITICAL. The two defects this PR set out to fix are genuinely fixed for the corpus it was measured against, and the red proof is real. What is missing is that the fix's own success metric can report success on inputs it did not fix.

bobmatnyc:
## Verdict: WARN

Independent second-round review at `c5056047`. I built my own WordPiece reference against `tokenizer.json` (30,522 entries, `handle_chinese_chars: true`, NFD + lowercase, `max_input_chars_per_word: 100`) and compiled the PR's real `piece_tokens` / `pack_chars` / `pack_whole_units` / `strip_notification_envelope` standalone with `rustc`, so the numbers below come from this PR's code, not from a mirror. My reference reproduces the engineer's own `chinese han` figure (292) and `korean hangul` figure (638) exactly, which is why I trust it.

**Verified myself:** the red proof's two quoted failure strings; the estimator's behaviour on 1,734 real corpus prompts; the `pack_chars` invariant (300k-case fuzz); the empty-query hole; the tag-render caps; all gates. **Taken on the engineer's word:** nothing material — where I could not re-run something (the 7-test revert suite) I verified the pre-review source directly instead and say so below.

## Findings

|---|---|---|---|---|---|
| HIGH | `query.rs` | 108, 471–488 | The estimator still underestimates on ordinary Latin-script input, and `RecallQueryShape` still reports no loss when it does | Bound the ASCII-letters branch, or apply headroom to the budget; correct the two "by construction" doc claims | Fix here |
| HIGH | `query.rs` | 330 | One leading newline restores the empty-query regression finding 3 closed | `lines.iter().find(\|l\| !l.trim().is_empty())` instead of `lines.first()` | Fix here |
| MEDIUM | `query.rs` | 269 | The trailing user instruction the envelope fix rescues is the first thing the packer drops | `kept.insert(0, tail)` instead of `kept.push(tail)` | Fix here |
| MEDIUM | `query.rs` | 423–435 | Packing abandons the rest of the budget at the first oversized line | Fall back to words *for that line* and keep filling | Parent |
| MEDIUM | `query.rs` | 123 | The 3/char non-ASCII charge cuts two thirds off Cyrillic/Greek/Arabic/Hebrew prompts that fit whole | Split the branch: 3/char for Hangul, 2/char for other non-ASCII | Parent |
| LOW | `query.rs` | 336 | `units_dropped` carries a character count on the last-resort path | `text.chars().count() - kept.chars().count()` | Fix here |
| LOW | `format.rs` | 205–219 | `MAX_RENDERED_TAG_CHARS` is not a ceiling for a single long tag (900-char tag → 914-char suffix) | Truncate the oversized single tag, or soften "the guarantee" at line 43 | Parent |
| LOW | `query.rs` | 728–737 | Two of nine "true WordPiece count" figures are not the tokenizer's (JWT 94 vs 78, snake_case 762 vs 442) | Re-measure, or drop the "measured against the model's own `tokenizer.json`" claim for those rows | Parent |

---

### HIGH-1 — the estimator still underestimates, and the metric still says it didn't

`query.rs:22-23` claims the estimator "errs high by construction on every input class it cannot model exactly", and `query.rs:468` repeats it as "Every branch errs high by construction". That is true of four of the five branches and false of the fifth.

- CJK = 1/char — bound (one padded pre-token per char, ≤1 token).
- Non-ASCII run = 3/char — bound. I measured 16 scripts; worst is dense Hangul at **2.46** tok/char.
- ASCII run with a digit = 1/char — bound (no NFD expansion, ≤1 token/char).
- Punctuation = 1, whitespace = 0 — bound.
- **ASCII letters = `ceil(len/3)` — not a bound.** It is a divisor calibrated on an English corpus. True cost reaches 1 token/char: `piece_tokens("qqqqqqqqq")` = 3, true = 9. Even `"YX"` is charged 1 and costs 2.

A 60k-case fuzz over mixed scripts found **zero** underestimates in any string without an ASCII-letter run, and all 26 underestimates inside one. The defect is that one branch, precisely.

What it does in the shape this PR exists to prevent — all figures from the compiled `estimate_tokens`:

```
input (chars)              RecallQueryShape                       true   embedder
german prose      1540     {original:510, sent:510, dropped:0}     543    6% cut, unreported
hungarian         1500     {original:512, sent:512, dropped:0}     609   16% cut, unreported
finnish           1520     {original:511, sent:511, dropped:0}     565    9% cut, unreported
camelCase idents  1560     {original:511, sent:511, dropped:0}     533    4% cut, unreported
```

The worse form is *after* shaping runs:

```
german x60   {original_tokens:4622, sent_tokens:504, units_dropped:748}
             true tokens in the text actually sent = 536   -> STILL OVER THE WINDOW
```

The code drops 748 units, asserts the remainder fits in 512, sends 536, and the embedder cuts it again. `jq 'select(.recall_query.units_dropped > 0)'` — the filter `config.rs` advertises — reports a reduction that did not achieve what it claims.

**Why this is HIGH and not CRITICAL.** I ran my reference over 314 real prompts sampled from all 31 `enriched-prompts.*.jsonl` files and got `p01=1.00 p50=1.38 p95=1.60, underestimates 0.00%` — the engineer's 600-prompt Latin figure reproduces exactly. This does not fire on the current corpus. And where it does fire, delivered behaviour is not worse than pre-PR: shaping is head-first, so the first 512 true tokens the embedder sees are the same ones it saw before. What is new and wrong is the asserted bound and a metric that reads clean while a cut happens.

`token_estimate_never_underestimates` is a nine-case table, not the property its name states. Either the name and the two doc paragraphs come down to what is actually guaranteed, or the branch becomes a bound.

### HIGH-2 — a leading newline empties the query

`query.rs:330`:

```rust
let first = lines.first().copied().unwrap_or(text);
```

When the prompt starts with `\n`, `lines()[0]` is `""`. Zero words pack from `""`, `pack_chars("")` returns `""`, and `fetch.rs:59` (`if prompt.is_empty() { return Vec::new(); }`) returns zero drawers. Against the compiled code:

```
plain 6000-char blob             -> kept_bytes= 1530  EMPTY=false   (the fix works)
LEADING NEWLINE + blob           -> kept_bytes=    0  EMPTY=true
leading blank lines + blob       -> kept_bytes=    0  EMPTY=true
leading space-only line + blob   -> kept_bytes=    3  EMPTY=false   (query is "   ")
```

`unbreakable_oversized_token_still_yields_a_query` misses it because its fixture has no leading newline. This is the identical failure the previous round raised, and the guard added for it has a one-character bypass. Add the leading-newline case to that test.

### MEDIUM-1 — the rescued trailing text is dropped first

`query.rs:264` says the post-`</task-notification>` text "outranks everything above it and must never be dropped", and `query.rs:269` appends it **last**. `pack_whole_units` packs head-first, so on any over-budget envelope it goes first. Compiled repro — envelope with a long `<result>` plus `now open the PR and set the milestone`:

```
envelope_stripped=true units_dropped=3
trailing user instruction survived? false
sent bytes = 11
```

The fix keeps the user's instruction only in the case where nothing needed keeping. `kept.insert(0, tail)` makes the implementation match the stated ranking.

### MEDIUM-2 — the budget is abandoned at the first oversized line

Same repro: a 512-token budget spends 11 bytes. `pack_units` breaks at the first overflowing unit, and `pack_whole_units` only reaches the word fallback when *no* line fits. Corpus impact, measured on 1,734 real firings (864 reshaped): p50 sent = 468/512, but **1.2% send under 128 tokens**, worst 57. Real, uncommon — hence MEDIUM and `Parent`.

Two corpus figures from the PR body reproduce independently here: envelope rate **65.5%** (claimed 65.3%) and reshape rate **49.8%** (claimed 52.0% over-window).

### MEDIUM-3 — is the over-charge harmful? Yes, quantifiably

Measured against the compiled shaper, prompts grown until they just fill the real 512-token window:

```
script     true (fits)   est    sent true    kept
russian        466      1546        147      32%
greek          482      1522        160      33%
arabic         497      1469        174      35%
hebrew         494      1454        168      34%
japanese       482       962        256      53%
korean         452       686        350      77%
chinese        482       482        482     100%
```

A Russian prompt that fit the window whole now loses two thirds of itself. Cause: `NON_ASCII_TOKENS_PER_CHAR = 3` is calibrated to Hangul (2.46 tok/char); Cyrillic measures 0.89, Greek 0.94, Arabic 1.00, Hebrew 1.00. Splitting the branch — 3/char for Hangul, 2/char for the rest — keeps the bound and halves the over-charge.

On English, the cost is smaller but not zero: of 800 sampled real prompts, 360 fit the real window whole and **29 of those (8.1%) are now trimmed anyway**, retaining a median 77% of what fit (min 56%).

Is there a threshold where over-charging is worse than the truncation it prevents? For Latin, no — 77% of a prompt head-first is close to what the embedder would have kept. For Cyrillic/Greek/Arabic/Hebrew, yes: 32–35% retention is a worse query than the un-shaped one the embedder would have accepted whole. It is disclosed and env-reversible, so MEDIUM.

---

## Items I was asked to confirm or refute

**Red proof (item 3).** I did not mutate the repo, so I did not re-run the 7-test revert. Instead I read the pre-review source at `7b5aa0cf` and compiled the old estimator standalone. Both quoted strings reproduce **exactly**:

```
`chinese han` underestimates: estimate 99 < true 292
cjk_is_one_token_per_char: piece_tokens("检索相关性下限") = 3 (test asserts 7)
```

The characterisation of the other five is correct on the source: `strip_notification_envelope` at `7b5aa0cf` has no trailing-text clause; `pack_whole_units` there returns `(kept, …)` with no last resort, so `"a".repeat(6000)` yields `""`; `render_tags` there is count-only and renders 246 chars for four 55-char tags. The two disclosed passes check out — `pack_chars` genuinely **does not exist** at `7b5aa0cf` (`grep -c "fn pack_chars"` = 0), so `pack_chars_matches_the_estimator` passing under a partial revert is evidence of nothing, exactly as stated; and `envelope_with_no_payload_passes_through_whole` pins a branch that already returned `None`. Count I observe: **5 of 7**, and the disclosure of the two is accurate.

**`pack_chars` drift (item 4).** The concern is unfounded — but not because the test is strong. 300,000 random cases over a mixed-script alphabet at random budgets: **0 overshoots, 0 non-maximal results**. `pack_chars` and `piece_tokens` agree because run cost is monotone in prefix length, so the greedy break is exact. `pack_chars_matches_the_estimator` is weaker than that (one fixture, four budgets, one direction, no maximality check) — it would not catch an undershoot regression that packed one character and stopped. Not worth a finding given the property holds; worth knowing the fuzz is what established it, not the test.

**The four MEDIUM fixes (item 5).** (a) present but ranked wrong — see MEDIUM-1. (b) present but holed — see HIGH-2. (c) present and honest; it pins an unchanged branch and says so. (d) genuinely fixed for the multi-tag case: four 55-char tags render 196 chars against the count-only cap's 246, and the test fails against the old code. It is not a ceiling for one oversized tag (900-char tag → 914-char suffix), which `format.rs:58-60` discloses.

**Fail-open check.** No new failure branch swallows an error. `shape_recall_query` is total, `warn_if_reshaped` is side-effect-only, `query_shape` is correctly `None` on the two paths that never reach a recall. HIGH-2's empty query is a functional hole, not a fail-open — `units_dropped` still records it. HIGH-1 *is* the fail-open shape (alarm reads healthy while loss occurs), which is why it is the lead finding.

**Repo gates.** All reproduced by me at `c5056047`:

```
CARGO_BUILD_JOBS=2 cargo fmt --check && cargo clippy -p trusty-memory --all-targets -- -D warnings
  && cargo test -p trusty-memory                                        EXIT=0
cargo test -p trusty-memory  — test result: ok. 624 passed; 0 failed; 4 ignored
scripts/check_line_cap.sh    — 3800 tracked .rs file(s); 4 allowlisted, 0 violations — OK
```

No `unwrap()`/`expect()` outside `#[cfg(test)]` in any of the four changed production files. Two changelog fragments, one category each.

**Dropping `cargo check --workspace`.** The file scope is honest — `git diff --name-only origin/main...HEAD` is entirely under `crates/trusty-memory/`. The justification is thinner than the conclusion, though: this PR *does* change public API (`pub use ENV_QUERY_TOKEN_BUDGET`, `pub struct RecallQueryShape`, a new `pub` field on `PromptLogEntry`), which is the ladder's rung-4 trigger regardless of file scope, and `trusty-agents` is a direct dependent. It happens to be safe — `commands` is gated behind `axum-server` which `trusty-agents` disables, `prompt_log` is not gated but the additions are purely additive, and `PromptLogEntry` appears nowhere outside `trusty-memory`. I did not build `trusty-agents` (build limits in force), so that last step rests on grep, not a compile.

## Required Changes

1. **HIGH-1** — at minimum, correct `query.rs:22-23` and `query.rs:468`: the estimator does not err high by construction on every branch. Then either bound the ASCII-letters branch or give the budget headroom so a modest underestimate cannot cross the window, and rename or re-scope `token_estimate_never_underestimates` to what it actually pins.
2. **HIGH-2** — `query.rs:330`: take the first non-blank line for the last-resort fallback, and add `"\n" + blob` to `unbreakable_oversized_token_still_yields_a_query`.
3. **MEDIUM-1** — `query.rs:269`: `insert(0, tail)`, with a test that has an over-budget `<result>` and asserts the trailing instruction survives.
4. **LOW-1** — `query.rs:336`: report units actually dropped, not the input's character count.

## Notes

- Findings 2–5 of the previous round were re-checked only where the new commits touched them, as directed. `is_creator_tag` reuse, the `prompt_log/` change, `pack_units` additivity, and the `Closes` targets are unchanged and not re-litigated.
- The 512-token window itself I took on the engineer's empirical probe; I did not re-run it against the live daemon. Note `tokenizer.json` in the fastembed cache carries `truncation.max_length: 128`, which fastembed overrides — the plateau data (scores varying through 456 tokens, flat from 532) is consistent with 512 and inconsistent with 128, so the probe stands.
- MEDIUM-2 and MEDIUM-3 are `Parent` because each changes a packing or charging contract the owner ruled on in #5037; they are the PM's call to schedule, not mine to file.

bobmatnyc:
## Round-3 fixes — both HIGHs, MEDIUM-1, LOW-1

Pushed `7ab55d7c6` on top of a merge of `origin/main` (`41c6e4d48`).

Every token figure below is measured against the model's own
`tokenizer.json` through the **`tokenizers` crate itself**, not a
reimplementation — `Tokenizer.from_file(...)` with `no_truncation()` and
`no_padding()`. (`padding.strategy: Fixed(128)` is set in that file, so a naive
read returns 128 for every short input; that trap cost me one wrong table
before I caught it.) My WordPiece reference agrees with the real tokenizer on
all 14 cases, including the previous critic's independently-derived `292` and
`638`.

### HIGH-1 — the estimator underestimates, and the shape reports no loss

Two defects, fixed separately, because fixing only the first leaves the second.

**The count.** `ceil(len/3)` was calibrated on English. Measured at the point
each class just fills a 512-token budget:

| input class | true | est ÷3 | est ÷2 |
|---|---|---|---|
| Hungarian agglutinative | 372 | **342** | 492 |
| Finnish compounds | 392 | **362** | 522 |
| Dutch compounds | 362 | **332** | 492 |
| German compounds | 442 | 442 *(zero margin)* | 632 |
| English prose | 44 | 77 | 113 |

`CHARS_PER_SUBTOKEN` is now `2`, which clears every compound-word language with
27–40% margin. **The cost is real and paid by English:** a prompt delivers a
measured **189** true tokens of the 512-token window rather than 291.
`TRUSTY_MEMORY_PROMPT_QUERY_TOKENS` raises it back.

**The reporting.** Correcting the divisor narrows the gap; it cannot close it.
No divisor above 1 token/char bounds a run of ASCII letters — `qzjvxwkfy` ×80 is
truly 562 tokens and `÷2` charges 402. Charging 1/char instead would deliver
**104** tokens of the window, so that is not a viable packing charge.

So the module now carries **two** numbers. `estimate_tokens` is what the packer
spends. `max_tokens` charges ASCII-letter runs 1/char and *is* a bound (every
ASCII letter is in vocab as both `x` and `##x`; a word over
`max_input_chars_per_word` collapses to one `[UNK]`). `RecallQueryShape` gains
`sent_tokens_max` and `may_exceed_window()`.

`sent_tokens <= budget_tokens` is an estimate clearing a budget, not a proof.
The log now distinguishes the two rather than asserting a clean pass on a query
the embedder may have cut. `config.rs` documents both filters.

It is a log field, not a second warn — `max_tokens` runs ~2.5× true on English,
so most prompts near the budget qualify and warning on all of them is a warn
nobody reads.

### HIGH-2 — one leading newline emptied the query

`lines.first()` on `"\n" + blob` is `""`: zero words pack, `pack_chars("")`
returns `""`, `fetch.rs:59` returns zero drawers. Now takes the first non-blank
line.

**The existing regression test passed against the bug** — its fixture had no
leading newline, exactly as suspected. Strengthened rather than duplicated: it
now runs bare, one leading newline, several blank lines, a space-only first
line, and CRLF.

### MEDIUM-1 / LOW-1

`kept.insert(0, tail)` — the code documented the post-`</task-notification>`
instruction as outranking everything above it and then appended it last, so the
head-first packer dropped it first. `units_dropped` on the last-resort path now
counts characters actually dropped, not the size of the input.

Also: the two true-token figures the review flagged as unmeasured are
re-measured (**JWT 78 not 94**, **snake_case 442 not 762**; hex is 890 not 695),
and `token_estimate_never_underestimates` → `token_estimate_covers_the_measured_scripts`,
which is what a nine-case table can pin.

### Not fixed, deliberately

**MEDIUM-2** (packing abandons the budget at the first oversized line) and
**MEDIUM-3** (split the non-ASCII charge by script) are left alone — routed to
`Parent`, and each changes a packing or charging contract the owner ruled on in
 #5037. One caveat the PM should weigh: the divisor change makes MEDIUM-2 fire
somewhat more often, since more lines now score as oversized.

## Red proof

Five production behaviours reverted, **test files untouched**, then
`cargo test -p trusty-memory --lib prompt_context::query`:

```
running 17 tests
test commands::prompt_context::query::tests::token_estimate_never_splits_a_unit ... FAILED
test commands::prompt_context::query::tests::shape_flags_a_send_it_cannot_prove_fits ... FAILED
test commands::prompt_context::query::tests::max_tokens_bounds_the_real_tokenizer ... FAILED
test commands::prompt_context::query::tests::latin_compounds_are_not_underestimated ... FAILED
test commands::prompt_context::query::tests::unbreakable_oversized_token_still_yields_a_query ... FAILED
test commands::prompt_context::query::tests::envelope_trailing_instruction_outranks_the_payload ... FAILED

---- latin_compounds_are_not_underestimated stdout ----
`hungarian agglutinative` underestimates: estimate 342 < true 372. Ordinary Latin
prose, not an edge case — the embedder cuts this query and RecallQueryShape
reports units_dropped: 0.

---- shape_flags_a_send_it_cannot_prove_fits stdout ----
the true cost is 562 against a 512 window: the shape must not report a clean pass
it cannot prove; got RecallQueryShape { original_tokens: 242, sent_tokens: 242,
sent_tokens_max: 242, budget_tokens: 512, envelope_stripped: false, units_dropped: 0 }

---- max_tokens_bounds_the_real_tokenizer stdout ----
`nine random letters` breaks the bound: max_tokens 242 < true 562

---- unbreakable_oversized_token_still_yields_a_query stdout ----
`bare`: units_dropped 6000 must count what was dropped, not the whole input (6000 chars)

---- envelope_trailing_instruction_outranks_the_payload stdout ----
the user's instruction outranks the agent report and must survive the packer, not
just the strip; got:
The agent swept the retrieval floor and reported back. The agent swept the ...

test result: FAILED. 11 passed; 6 failed; 0 ignored; 0 measured; 615 filtered out
```

`unbreakable_oversized_token_still_yields_a_query` fails there on the LOW-1
assertion, which masks HIGH-2. Isolated — **only** `lines.first()` reverted, the
other four fixes in place:

```
running 1 test
test commands::prompt_context::query::tests::unbreakable_oversized_token_still_yields_a_query ... FAILED

---- unbreakable_oversized_token_still_yields_a_query stdout ----
`one leading newline`: an unbreakable oversized token must not produce an empty
query — that returns zero drawers, worse than the truncation it replaced

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 631 filtered out
```

The `bare` case runs first and **passes** in that run — that is the proof the
old fixture could not catch this.

Being precise about one of the six: `token_estimate_never_splits_a_unit` fails
under revert only because it pins `piece_tokens("abcdefghijkl")`, which moved
4 → 6 with the divisor. It proves the divisor changed, not that a defect was
fixed. The other five are behavioural.

Green, same command:

```
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 615 filtered out; finished in 0.03s
```

## Gates

```
( CARGO_BUILD_JOBS=2 cargo fmt --check
  && CARGO_BUILD_JOBS=2 cargo clippy -p trusty-memory --all-targets -- -D warnings
  && CARGO_BUILD_JOBS=2 cargo test -p trusty-memory
  && CARGO_BUILD_JOBS=2 cargo test -p trusty-common ) ; EXIT=0

cargo test -p trusty-memory  — test result: ok. 628 passed; 0 failed; 4 ignored
cargo test -p trusty-common  — test result: ok. 294 passed; 0 failed; 6 ignored
```

628 = 624 on main after #5201 merged its four tag tests back in, plus the four
added here. Both `axum-server` e2e tests ran and passed, which is what confirms
the merge resolution kept both sides:

```
test commands::prompt_context::tests::prompt_context_injection_has_no_provenance_tags ... ok
test commands::prompt_context::tests::prompt_context_logs_recall_query_shape ... ok
```

```
line-cap: measured 3803 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK.
changelog-fragment gate: scanned 6 changed path(s); all 1 crate(s) with source changes are recorded.
```

**`cargo check --workspace` — reported honestly.** It fails, on
`trusty-code-gui` and `trusty-mpm-gui`, both with:

```
= help: message: The `frontendDist` configuration is set to "ui/dist" but this path doesn't exist
error: proc macro panicked
```

That is the Tauri build macro requiring a built frontend under `SKIP_UI_BUILD=1`,
not this diff. `git diff --name-only origin/main -- crates/trusty-code-gui/
crates/trusty-mpm-gui/` is empty. Excluding those two:

```
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui   EXIT=0
```

On the rung-4 dependent question the previous critic left resting on grep: I
re-ran it and it holds — `RecallQueryShape`, `PromptLogEntry`,
`ENV_QUERY_TOKEN_BUDGET`, `sent_tokens_max` and `may_exceed_window` appear
nowhere outside `crates/trusty-memory/`. `trusty-agents` compiles in the
workspace check above.

## Changelog

`crates/trusty-memory/changelog.d/4972-recall-query-shaping.md` was accurate for
the split (no #5038 content) but understated the work; it now covers the divisor
change, its cost to English, and `sent_tokens_max`.

## One thing worth a decision

`max_tokens` exists because the estimator cannot be exact without the
vocabulary. The design that removes the whole class of finding is to stop
estimating — have the embedder daemon report the true token count, or expose a
count endpoint. That is a bigger change than a review round and I did not take
it. If a fourth round lands on this same estimator, that is the signal to do it
rather than tune again.

Issue #3761, opened 2026-07-23
relay token: constant-time compare + validate SlackReplySent.identity server-side


bobmatnyc:
#3754 re-review LOWs: relay_authorized uses ordinary string equality (loopback, low practical risk — cheap hardening via subtle::ConstantTimeEq); identity remains free-form for valid-token holders — server knows the one honest value, validate/replace as defense-in-depth.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #3269, opened 2026-07-19
fix(trusty-console): Tailscale bind mode 403s the console's own write UI (origin guard is loopback-only)


bobmatnyc:
## Root Cause

`BindMode::Tailscale` (in `crates/trusty-console/src/bind.rs`) binds to both loopback and the detected CGNAT 100.64.0.0/10 address. However, `origin_is_loopback` (in `crates/trusty-console/src/routes/origin_guard.rs`, ~lines 35-56) accepts only localhost/127.x/::1. The crate's own test `origin_is_loopback_rejects_remote_hosts` asserts that `http://100.64.1.2:7788` is rejected.

Browsers send Origin headers on same-origin state-changing fetches, so an SPA loaded from `http://100.x.x.x:7788` gets 403 on every session/config write while reads work (GET/HEAD bypass the guard).

## Fix Approach (Recommended)

Make the guard bind-aware by passing the server's actual resolved listening addresses (already computed in `resolve_bind_addrs`) into the guard as the trusted "self-origin" allowlist, rather than using a static loopback list. **Do NOT blanket-trust all of 100.64.0.0/10.**

Add a test covering the full Tailscale SPA write flow.

## Cross-Links

- Relates to Issue #3268 (same guard, fixes will likely land together)
- Feature ticket #1092 (introduced Tailscale bind mode)

## Severity

P1 (High) — breaks write UI for Tailscale deployments.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5185, opened 2026-08-08
fix(trusty-agents): authenticate the /api/events SSE stream and stop cross-origin reads (#5052)

### LINKED ISSUE #5052: security: /api/events SSE stream unauthenticated even when token configured
**Defect:** In trusty-agents, `/api/health`, `/api/config`, and `/api/events` remain unauthenticated even when an auth token is configured. `/api/events` is exempt because `EventSource` cannot set request headers (auth.rs:118-122). A token-guarded non-loopback deployment therefore leaks that stream to any host that can route to the port.

**Evidence:** The SSE stream carries `PmThinking.text`, `AgentMessage.text`, and task previews — conversation content, not health-metric telemetry. Verified during PR #5051 review; `auth_middleware` in trusty-agents' `api::server::auth::auth_middleware` is only installed when a token is set, and these three routes sit outside it regardless (routes.rs:279, auth.rs:123-126).

**Resolution:** Authenticate the SSE stream by a means available to `EventSource` — a query-parameter token, a signed short-lived ticket issued by an authenticated endpoint, or a cookie — and bring `/api/config` under the middleware. Decide `/api/health` deliberately; an unauthenticated liveness probe may be intended.

~~Note: Reachability is currently gated behind a non-default `--bind`, since the default bind is `127.0.0.1` (api::server::auth::ApiConfig) and a non-loopback bind is refused without a token. Severity is bounded by that, not by the content sensitivity alone.~~

**Correction (see comment below for full evidence):** the loopback default does NOT gate reachability. Any web page open in the user's browser can hit `http://127.0.0.1:7654/api/events` cross-origin via `EventSource` — permissive CORS (`allow_origin(Any)`) applies, and the auth exemption fires regardless of whether a token is configured. The shipped Tauri desktop GUI launches this API by default (`trusty-agents --api --port 7654`). Severity is not bounded by `--bind`.

bobmatnyc:
Closes https://github.com/bobmatnyc/trusty-tools/issues/5052

## The defect

`GET /api/events` was exempt from bearer auth before the token was compared
([`auth.rs:149-151`](https://github.com/bobmatnyc/trusty-tools/blob/352fe5d6/crates/trusty-agents/src/api/server/auth.rs#L149-L151)),
and the router carried `CorsLayer::new().allow_origin(Any)`
([`trusty-common/src/server/mod.rs:61-64`](https://github.com/bobmatnyc/trusty-tools/blob/352fe5d6/crates/trusty-common/src/server/mod.rs#L61-L64)).
Those two compose into a cross-origin disclosure of live conversation content.

A loopback bind does not contain it. The attacker's JavaScript runs inside the
operator's own browser, which can reach `127.0.0.1`: any page could call
`new EventSource("http://127.0.0.1:7654/api/events")` and read
`Event::PmThinking.text`, `Event::AgentMessage.text`, and
`Event::PmDelegating.task_preview` — no shell access, no token, no non-loopback
opt-in, and no log entry. The shipped Tauri sidecar starts that port by default
([`mode_dispatch.rs:483`](https://github.com/bobmatnyc/trusty-tools/blob/352fe5d6/crates/trusty-agents/src/runtime/mode_dispatch.rs#L483)).

Remote network access was never broken and is unchanged: `serve_with_config`
still refuses a non-loopback bind without a token.

## Mechanism: a short-lived ticket, not a query-param token

`EventSource` cannot set an `Authorization` header, so removing the exemption
alone breaks the UI. Of the three options in the issue:

- **Query-param bearer token — rejected.** The URL is the part of the request
  that leaks: access logs, `Referer`, browser history. A leaked bearer token
  grants the whole `/api/*` write surface, which can spawn arbitrary
  subprocesses, and never expires.
- **Cookie — rejected.** Ambient authority, and the Tauri webview reaches the
  sidecar cross-origin, which would force `allow-credentials` and a
  reflected-origin CORS policy anyway.
- **Ticket — chosen.** `POST /api/events/ticket` returns a 256-bit opaque
  value with a 5-minute TTL that slides on redemption. Redeemable more than
  once, because `EventSource` reconnects to the *same* URL and a single-use
  ticket would make any transient blip a permanently dead stream. A leaked
  ticket buys five minutes of read-only stream access.

**Why minting is a POST.** That places it behind the existing router-wide
same-origin write guard, which fires even when no operator token is configured.
That is what closes the *shipped default* (sidecar on `127.0.0.1:7654`, no
token): `https://evil.example` gets 403 at the mint endpoint and so never
obtains a ticket. When a token *is* configured, `auth_middleware` gates the mint
route on top, since it is not an exempt path.

`GET /api/events` accepts a matching `Authorization: Bearer` **or** a live
ticket — the header path keeps `curl`, the reverse proxy, and the desktop
shell's Rust SSE bridge working. It is gated whether or not a token is
configured; the token-less default is precisely the configuration that was
attacked.

Constant-time comparison reuses `auth::bearer_token_matches` (#3761) and every
live ticket is compared even after a hit, so timing reveals neither a prefix
nor which ticket matched.

## Exemption decision, per route

| Route | Decision | Why |
|---|---|---|
| `/api/health` | **stays exempt** | Liveness probe; the Tauri sidecar polls it *before* it holds any credential. Body is `{status, version, pid, ppid}` — no conversation content. |
| `/api/config` | **stays exempt** | Its entire purpose is pre-auth bootstrap. Body is the single boolean `auth_required`, which any caller already infers from whether an `/api/*` request 401s — gating it removes no information from an attacker while removing the browser's only way to learn it needs a token. |
| `/api/events` | **exemption removed** | Streams conversation content. |

This diverges from the issue's suggested "bring `/api/config` under the
middleware"; the reasoning is above rather than silent.

## CORS decision: tightened, and it is load-bearing

The auth fix alone is **not** sufficient, and the issue's framing understates
the surface. `/api/tasks`, `/api/task/{id}`, `/api/sessions/{id}/recap` and
`/api/listener-events` are all GET, all carry task narratives and conversation
content, and all were cross-origin readable under `allow_origin(Any)` in the
default config. The same-origin write guard never covered them — it is
method-gated, and reads are GET.

So trusty-agents now takes a new
`trusty_common::server::with_guarded_middleware_same_origin_cors`, which
reflects only same-machine origins: loopback (`origin_is_loopback`, which
already rejects `127.0.0.1.evil.com`-style lookalikes), a local webview shell
(new `origin_is_local_webview` — the three exact Tauri v2 origins), and the
daemon's own resolved non-loopback bind (#3269). `pnpm dev` on
`localhost:5173` keeps working; server-side callers send no `Origin` and are
unaffected.

CORS is browser-enforced, never a server-side access control — it is defence in
depth *behind* the ticket gate, not a replacement for it. The sibling daemons
keep `with_guarded_middleware`; their GET surfaces deserve the same review, but
on their own tickets, not smuggled into a security fix.

## Client side (both transports, in this PR)

- `ui/src/lib/transport.ts` — `mintEventTicket()`, and `connectEventSource`
  now mints before connecting and **owns the reconnect loop**: on transport
  error it closes the socket and re-mints with exponential backoff (1s→15s).
  The browser's native retry re-uses the original URL, so once a ticket expired
  it could only ever 401 forever.
- `ui/src/App.svelte` — the start path is async now, with a `wantEventStream`
  guard so a stop landing during the mint round-trip wins.
- `ui/src-tauri/src/sse_bridge.rs` — the desktop shell's Rust bridge mints a
  fresh ticket per connection attempt.
- Both Playwright specs stub `/api/events/ticket`.

## Regression proof — fails before, passes after

Rung 5 requires a test that fails against the pre-fix code. Both pre-fix states
were restored in place and the suite re-run.

**BEFORE — handler gate disabled + `with_guarded_middleware` (permissive CORS) restored:**

```
running 16 tests
test ...::cross_origin_request_gets_no_cors_reflection ... FAILED
test ...::loopback_origin_is_cors_reflected ... FAILED
test ...::event_stream_without_credential_is_rejected ... FAILED
test ...::event_stream_with_unknown_ticket_is_rejected ... FAILED

---- ...::event_stream_without_credential_is_rejected stdout ----
  left: 200
 right: 401

---- ...::cross_origin_request_gets_no_cors_reflection stdout ----
a remote origin must not be reflected

---- ...::loopback_origin_is_cors_reflected stdout ----
assertion `left == right` failed: http://localhost:5173 must be reflected
  left: Some("*")
 right: Some("http://localhost:5173")

test result: FAILED. 12 passed; 4 failed; 0 ignored; 0 measured; 3398 filtered out
```

**BEFORE (2) — `|| path == "/api/events"` also restored in `auth_middleware`,** to
prove the issue's own claim (unauthenticated *even when a token is configured*):

```
running 2 tests
test ...::event_stream_without_credential_is_rejected_with_token_configured ... FAILED
test ...::event_stream_without_credential_is_rejected ... FAILED

---- ...::event_stream_without_credential_is_rejected_with_token_configured stdout ----
  left: 200
 right: 401

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 3412 filtered out
```

**AFTER:**

```
running 23 tests
test api::server::tests::event_tickets::event_stream_without_credential_is_rejected ... ok
test api::server::tests::event_tickets::event_stream_without_credential_is_rejected_with_token_configured ... ok
test api::server::tests::event_tickets::event_stream_with_unknown_ticket_is_rejected ... ok
test api::server::tests::event_tickets::minted_ticket_opens_event_stream ... ok
test api::server::tests::event_tickets::minted_ticket_is_redeemable_more_than_once ... ok
test api::server::tests::event_tickets::bearer_token_opens_event_stream ... ok
test api::server::tests::event_tickets::ticket_does_not_unlock_the_rest_of_the_api ... ok
test api::server::tests::event_tickets::mint_route_requires_bearer_token_when_configured ... ok
test api::server::tests::event_tickets::mint_route_rejects_cross_origin_caller ... ok
test api::server::tests::event_tickets::mint_route_allows_loopback_origin ... ok
test api::server::tests::event_tickets::cross_origin_request_gets_no_cors_reflection ... ok
test api::server::tests::event_tickets::loopback_origin_is_cors_reflected ... ok
test api::server::tests::event_tickets::mint_returns_distinct_unguessable_tickets ... ok
test api::server::tests::event_tickets::redeem_rejects_expired_ticket ... ok
test api::server::tests::event_tickets::redeem_rejects_unknown_ticket ... ok
test api::server::tests::event_tickets::store_is_capacity_bounded ... ok
test api::server::tests::events_sse::events_query_rejects_unknown_parameter ... ok
test api::server::tests::events_sse::events_query_accepts_known_parameters ... ok
...
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 3391 filtered out; finished in 0.02s
```

The pre-existing `events_sse` query tests were **kept, not narrowed** — their
helper now mints a ticket, because a bare 401 would have satisfied their
`is_client_error()` assertions for the wrong reason.

## Live binary smoke — `tagent --api --port 7699`, no token (the shipped default)

```
=== health (exempt):
200
=== UNAUTH /api/events:
{"error":"unauthorized"}
HTTP 401
=== cross-origin EventSource simulation (Origin: https://evil.example):
HTTP/1.1 401 Unauthorized
=== cross-origin ticket mint:
{"error":"cross-origin write requests are not allowed"}
HTTP 403
=== same-origin mint (as the UI does):
{"ticket":"fUROku4_vY7szvDhDaYWkavC-1092UaRN_eiI4hpXb4","expires_in_secs":300}
=== AUTHENTICATED stream with that ticket:
HTTP/1.1 200 OK
content-type: text/event-stream
cache-control: no-cache
vary: origin, access-control-request-method, access-control-request-headers
access-control-allow-origin: http://127.0.0.1:7699
transfer-encoding: chunked
=== stream with a forged ticket:
{"error":"unauthorized"}
HTTP 401
```

## Gates — test ladder rung 5 (security) + rung 6 (UI/API surface)

```
cargo fmt --check                                                → FMT=0
cargo check -p trusty-agents -p trusty-common                    → CHECK=0
cargo clippy -p trusty-common --all-targets --features axum-server -- -D warnings
                                                                 → CLIPPY_COMMON=0
cargo clippy -p trusty-agents --all-targets -- -D warnings       → CLIPPY_AGENTS=0
```

```
cargo test -p trusty-common --features axum-server
test result: ok. 312 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 0.98s
TCOMMON=0
```

```
cargo test -p trusty-agents
test result: ok. 3403 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 34.15s
test result: ok. 6 passed; 0 failed; 3 ignored
test result: ok. 2 passed; 0 failed; 0 ignored
test result: ok. 1 passed; 0 failed; 0 ignored
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.82s   # plain_cli_repl
test result: ok. 2 passed; 0 failed; 0 ignored
test result: ok. 2 passed; 0 failed; 0 ignored
```

The `plain_cli_repl` integration binary timed out (7× `tagent did not exit
within 120s`) on the first pass, while the `tagent` debug binary was still
linking inside the spawn window under a constrained-parallelism build. It
passes in 3.82s on a warm re-run, shown above. It spawns the CLI REPL and
touches no `/api/*`, CORS, or SSE code path in this diff.

**UI (rung 6):**

```
pnpm test:unit    → 442 passed / 443 (1 pre-existing failure, below)
npx vitest run src/lib/transport.test.ts
                  → Test Files 1 passed (1);  Tests 8 passed (8)
pnpm check        → COMPLETED 1776 FILES 0 ERRORS 3 WARNINGS   (all 3 pre-existing a11y warnings)
pnpm build        → ✓ built in 6.03s
```

`ProjectFilesPanel.test.ts > loads markdown into the editor` fails with
"timed out waiting for the editor to mount", standalone as well as in the full
run. It imports only `ProjectFilesPanel.svelte` → svelte / lucide-svelte /
`MarkdownEditor.svelte` / `lib/projectFiles.ts`; not one file in this diff is in
that graph. A CodeMirror mount timeout, pre-existing.

**Repo gates:**

```
scripts/check_line_cap.sh          → 3801 files, 4 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh     → 22626 citations, 0 dangling — OK
scripts/check_sld.sh               → 56 specs + 5771 code files, 0 errors, 0 warnings
scripts/check_changelog_fragment.sh→ trusty-agents OK, trusty-common OK
```

`cargo check --workspace` is **not** run: it fails on `trusty-code-gui` and
`trusty-mpm-gui` with `frontendDist "ui/dist" doesn't exist` — the reason CI
excludes both. Neither is in this diff, and building unrelated Tauri frontends
to turn a gate green is not a fix.

## Follow-up worth its own ticket

The sibling daemons (trusty-search, trusty-memory, trusty-analyze,
trusty-review, trusty-code) still serve `allow_origin(Any)`. `trusty-memory` in
particular exposes palace content over GET. `with_guarded_middleware_same_origin_cors`
is now a one-line adoption for each, but each needs its own review of what its
GET surface actually discloses.

### DIFF
--- a/crates/trusty-agents/changelog.d/5052-sse-auth.md
+++ b/crates/trusty-agents/changelog.d/5052-sse-auth.md
@@ -14,9 +14,10 @@ Security
   short-lived ticket minted at the new `POST /api/events/ticket`, which a
   browser can put in the `EventSource` URL. Minting is a write method, so it is
   behind the router-wide same-origin guard even when no operator token is
-  configured — which is what closes the shipped default. A ticket that leaks
-  into an access log or `Referer` grants five minutes of read-only stream
-  access, not the whole `/api/*` write surface.
+  configured — which is what closes the shipped default. A ticket expires 5
+  minutes after its last use and 1 hour after issue whatever happens in
+  between, and grants only the read-only event stream — never the `/api/*`
+  write surface, which can spawn arbitrary subprocesses.
 
   `/api/health` and `/api/config` stay exempt deliberately: the sidecar polls
   health before it holds any credential, and `/api/config` publishes only the
--- a/crates/trusty-agents/changelog.d/5052-sse-auth.md
+++ b/crates/trusty-agents/changelog.d/5052-sse-auth.md
@@ -1,0 +1,34 @@
+Security
+
+- **`GET /api/events` no longer streams conversation content to unauthenticated
+  callers** ([#5052](https://github.com/bobmatnyc/trusty-tools/issues/5052)).
+  The SSE stream was exempt from bearer auth outright, on the grounds that a
+  browser `EventSource` cannot send an `Authorization` header. A loopback bind
+  did not contain that: any web page open in the operator's browser could call
+  `new EventSource("http://127.0.0.1:7654/api/events")` and read
+  `PmThinking.text`, `AgentMessage.text`, and `PmDelegating.task_preview`
+  cross-origin — no shell access, no token, no non-loopback opt-in, and no log
+  entry. The shipped Tauri sidecar starts that port by default.
+
+  The stream now requires either a matching `Authorization: Bearer <token>` or a
+  short-lived ticket minted at the new `POST /api/events/ticket`, which a
+  browser can put in the `EventSource` URL. Minting is a write method, so it is
+  behind the router-wide same-origin guard even when no operator token is
+  configured — which is what closes the shipped default. A ticket that leaks
+  into an access log or `Referer` grants five minutes of read-only stream
+  access, not the whole `/api/*` write surface.
+
+  `/api/health` and `/api/config` stay exempt deliberately: the sidecar polls
+  health before it holds any credential, and `/api/config` publishes only the
+  `auth_required` boolean that any 401 already reveals.
+
+- **The trusty-agents API no longer serves permissive CORS**
+  ([#5052](https://github.com/bobmatnyc/trusty-tools/issues/5052)).
+  `allow_origin(Any)` let a page on any origin read every GET response —
+  `/api/tasks`, `/api/task/{id}`, `/api/sessions/{id}/recap` all carry
+  conversation content, and the same-origin write guard did not cover them
+  because it is method-gated. The daemon now reflects only same-machine origins:
+  loopback (including a `pnpm dev` server on `localhost:5173`), the Tauri
+  webview, and its own resolved non-loopback bind. Server-side callers — the
+  trusty-console reverse proxy, `curl`, MCP stdio bridges — send no `Origin` and
+  are unaffected.
--- a/crates/trusty-agents/src/api/server/auth.rs
+++ b/crates/trusty-agents/src/api/server/auth.rs
@@ -4,8 +4,10 @@
 //! control that keeps it off the LAN — the bind is (#3329). Bearer-token auth
 //! covers the one case the bind cannot: an operator who explicitly opts into a
 //! non-loopback `--bind`, which `serve_with_config` refuses to start without a
-//! token. It gates the sensitive `/api/*` routes while exempting
-//! health/config/events probes and the embedded UI assets.
+//! token. It gates the sensitive `/api/*` routes while exempting the
+//! `/api/health` and `/api/config` bootstrap probes and the embedded UI assets.
+//! #5052: `/api/events` is NO LONGER exempt — see `auth_middleware` and
+//! `super::event_tickets`.
 //! What: `ApiConfig` carries the bind + port + optional token;
 //! `auth_middleware` enforces the token; `ApiClientConfig` tells the UI whether
 //! auth is needed.
@@ -122,14 +124,35 @@ pub(super) fn bearer_token_matches(expected: &str, provided: &str) -> bool {
 /// exempt the
 /// embedded UI's static assets so a browser can load `index.html` and obtain
 /// the token via `/api/config` before issuing authenticated requests.
-/// What: For requests under `/api/*` (other than `/api/health`), checks
-/// `Authorization: Bearer <token>` and returns 401 JSON
+///
+/// Two routes remain exempt, and #5052 settled each deliberately rather than by
+/// inheritance:
+///   - `/api/health` — a liveness probe, and the Tauri sidecar polls it BEFORE
+///     it has any credential to present. Its body is
+///     `{status, version, pid, ppid}`; nothing there is conversation content.
+///   - `/api/config` — its entire purpose is pre-auth bootstrap: it tells the UI
+///     whether to prompt for a token. Its body is the single boolean
+///     `auth_required`, which any caller can already infer from whether an
+///     `/api/*` request 401s, so putting it behind the middleware would remove
+///     no information from an attacker while breaking the browser's only way to
+///     learn it needs a token.
+///
+/// `/api/events` USED to be a third exemption, on the grounds that a browser
+/// `EventSource` cannot send an `Authorization` header. That was a
+/// cross-origin disclosure of live conversation content, not a telemetry
+/// trade-off (#5052): the stream carries `PmThinking.text`, `AgentMessage.text`,
+/// and `PmDelegating.task_preview`, and any page in the operator's browser could
+/// read it from `127.0.0.1` with no credential at all. It is now authenticated
+/// by ticket — see `super::event_tickets`.
+/// What: For requests under `/api/*` (other than `/api/health` and
+/// `/api/config`), checks `Authorization: Bearer <token>` and returns 401 JSON
 /// `{"error":"unauthorized"}` on mismatch. #3761: the comparison is
 /// constant-time (`bearer_token_matches`) — this is the higher-value secret of
 /// the two the server holds, so it gets the same treatment as the relay token.
 /// Test: `auth_middleware_rejects_missing_token`,
 /// `auth_middleware_accepts_valid_token`, `auth_middleware_allows_health`,
-/// `bearer_token_matches_is_exact`.
+/// `bearer_token_matches_is_exact`,
+/// `event_tickets::mint_route_requires_bearer_token_when_configured`.
 pub(super) async fn auth_middleware(
     State(auth): State<AuthState>,
     req: Request<axum::body::Body>,
@@ -138,19 +161,14 @@ pub(super) async fn auth_middleware(
     let path = req.uri().path();
 
     // Public endpoints — never require auth:
-    //   - GET /api/health  (health probes)
+    //   - GET /api/health  (health probes; the sidecar polls it pre-credential)
     //   - GET /api/config  (UI bootstrap: tells client whether auth is needed)
     //   - any non-/api path (UI static assets at "/" and "/*path")
-    // #192 Phase B: `/api/events` is exempt from Bearer auth because the
-    // browser EventSource API cannot attach custom Authorization headers.
-    // Auth-sensitive deployments should front the server with a reverse proxy
-    // and gate `/api/events` there (e.g. mTLS or cookie-based auth) since the
-    // event stream itself contains only telemetry, not actionable controls.
-    if path == "/api/health"
-        || path == "/api/config"
-        || path == "/api/events"
-        || !path.starts_with("/api/")
-    {
+    // #5052: `/api/events` was a third exemption and is not one any more — it
+    // streams conversation content and is now gated by a ticket that a browser
+    // EventSource CAN carry (see `super::event_tickets`). Both remaining
+    // exemptions are justified in this function's doc comment.
+    if path == "/api/health" || path == "/api/config" || !path.starts_with("/api/") {
         return next.run(req).await;
     }
 
--- a/crates/trusty-agents/src/api/server/auth.rs
+++ b/crates/trusty-agents/src/api/server/auth.rs
@@ -170,10 +170,6 @@ pub(super) async fn auth_middleware(
     //   - GET /api/health  (health probes; the sidecar polls it pre-credential)
     //   - GET /api/config  (UI bootstrap: tells client whether auth is needed)
     //   - any non-/api path (UI static assets at "/" and "/*path")
-    // #5052: `/api/events` was a third exemption and is not one any more — it
-    // streams conversation content and is now gated by a ticket that a browser
-    // EventSource CAN carry (see `super::event_tickets`). Both remaining
-    // exemptions are justified in this function's doc comment.
     // #5052 (critic HIGH): `/api/events` is exempt HERE and gated in the handler
     // instead. This middleware is layered OUTSIDE the routes, so a request
     // carrying only a ticket never reached `EventStreamAuth` — the server minted
--- a/crates/trusty-agents/src/api/server/auth.rs
+++ b/crates/trusty-agents/src/api/server/auth.rs
@@ -6,7 +6,9 @@
 //! non-loopback `--bind`, which `serve_with_config` refuses to start without a
 //! token. It gates the sensitive `/api/*` routes while exempting the
 //! `/api/health` and `/api/config` bootstrap probes and the embedded UI assets.
-//! #5052: `/api/events` is NO LONGER exempt — see `auth_middleware` and
+//! #5052: `/api/events` is exempt HERE but not unauthenticated — it is gated in
+//! `events_sse::events_handler`, which takes this bearer token OR a ticket a
+//! browser `EventSource` can carry. See `auth_middleware` and
 //! `super::event_tickets`.
 //! What: `ApiConfig` carries the bind + port + optional token;
 //! `auth_middleware` enforces the token; `ApiClientConfig` tells the UI whether
@@ -137,15 +139,19 @@ pub(super) fn bearer_token_matches(expected: &str, provided: &str) -> bool {
 ///     no information from an attacker while breaking the browser's only way to
 ///     learn it needs a token.
 ///
-/// `/api/events` USED to be a third exemption, on the grounds that a browser
-/// `EventSource` cannot send an `Authorization` header. That was a
-/// cross-origin disclosure of live conversation content, not a telemetry
-/// trade-off (#5052): the stream carries `PmThinking.text`, `AgentMessage.text`,
-/// and `PmDelegating.task_preview`, and any page in the operator's browser could
-/// read it from `127.0.0.1` with no credential at all. It is now authenticated
-/// by ticket — see `super::event_tickets`.
-/// What: For requests under `/api/*` (other than `/api/health` and
-/// `/api/config`), checks `Authorization: Bearer <token>` and returns 401 JSON
+/// `/api/events` is a third exempt path, but for the opposite reason it used to
+/// be: it is not unauthenticated, it is authenticated SOMEWHERE ELSE. Pre-#5052
+/// it streamed conversation content — `PmThinking.text`, `AgentMessage.text`,
+/// `PmDelegating.task_preview` — to anyone, and any page in the operator's
+/// browser could read it from `127.0.0.1`. It is now gated inside
+/// `events_sse::events_handler`, which accepts this same bearer token OR a
+/// ticket a browser `EventSource` can carry in the URL, and 401s with neither.
+/// The gate had to move into the handler because this middleware wraps the
+/// routes from the outside: a ticket-only request was rejected here before
+/// `EventStreamAuth` ever saw it.
+/// What: For requests under `/api/*` (other than `/api/health`, `/api/config`,
+/// and the separately-gated `/api/events`), checks
+/// `Authorization: Bearer <token>` and returns 401 JSON
 /// `{"error":"unauthorized"}` on mismatch. #3761: the comparison is
 /// constant-time (`bearer_token_matches`) — this is the higher-value secret of
 /// the two the server holds, so it gets the same treatment as the relay token.
@@ -168,7 +174,20 @@ pub(super) async fn auth_middleware(
     // streams conversation content and is now gated by a ticket that a browser
     // EventSource CAN carry (see `super::event_tickets`). Both remaining
     // exemptions are justified in this function's doc comment.
-    if path == "/api/health" || path == "/api/config" || !path.starts_with("/api/") {
+    // #5052 (critic HIGH): `/api/events` is exempt HERE and gated in the handler
+    // instead. This middleware is layered OUTSIDE the routes, so a request
+    // carrying only a ticket never reached `EventStreamAuth` — the server minted
+    // tickets it would then always 401, and since a non-loopback bind REQUIRES a
+    // token, the browser stream was dead in exactly the deployment #5052 was
+    // filed about. The exemption is safe only because `events_handler` accepts
+    // the same bearer token this middleware would have checked, AND a ticket;
+    // with neither it returns 401. It is an EXACT path match, so the sibling
+    // `/api/events/ticket` mint route stays fully covered here.
+    if path == "/api/health"
+        || path == "/api/config"
+        || path == "/api/events"
+        || !path.starts_with("/api/")
+    {
         return next.run(req).await;
     }
 
--- a/crates/trusty-agents/src/api/server/event_tickets.rs
+++ b/crates/trusty-agents/src/api/server/event_tickets.rs
@@ -1,0 +1,253 @@
+//! Short-lived tickets that authenticate the `/api/events` SSE stream (#5052).
+//!
+//! Why: `/api/events` streams conversation content — `PmThinking.text`,
+//! `AgentMessage.text`, `PmDelegating.task_preview` — and used to be exempt from
+//! bearer auth outright, because the browser `EventSource` API cannot attach an
+//! `Authorization` header. A loopback bind does not contain that: any page in
+//! the operator's browser could open
+//! `new EventSource("http://127.0.0.1:7654/api/events")` and read the stream,
+//! with no shell access, no token, and no non-loopback opt-in. Removing the
+//! exemption alone would break the shipped UI, so the stream needs a credential
+//! a `EventSource` URL can carry.
+//!
+//! A ticket rather than the bearer token itself in `?token=`: the URL is the one
+//! part of the request that leaks — into access logs, the `Referer` header, and
+//! browser history. A ticket that leaks grants read-only access to the event
+//! stream for a few minutes; the bearer token that leaks grants the whole
+//! `/api/*` write surface, which can spawn arbitrary subprocesses, forever.
+//!
+//! What: [`EventTicketStore`] mints unguessable ticket strings and redeems them
+//! for a sliding TTL window. [`EventStreamAuth`] is the request-time decision:
+//! a matching bearer token, or a live ticket, or 401. Minting happens at
+//! `POST /api/events/ticket`, which — being a write method — is already behind
+//! the router-wide same-origin guard even when no bearer token is configured, so
+//! a cross-origin page cannot mint one for itself.
+//! Test: `super::tests::event_tickets`.
+
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use axum::http::{HeaderMap, header};
+use base64::Engine as _;
+use rand::RngCore as _;
+use serde::Serialize;
+
+use super::auth::bearer_token_matches;
+
+/// How long a ticket stays valid, measured from mint OR from its last
+/// successful redemption (a sliding window).
+///
+/// Long enough that the browser's own `EventSource` reconnect after a transient
+/// blip re-uses the same URL successfully; short enough that a ticket recovered
+/// from an access log after the fact is dead. The UI mints a fresh ticket when a
+/// reconnect does fail (`connectEventSource` in `ui/src/lib/transport.ts`), so
+/// this value is not an upper bound on stream lifetime.
+pub(super) const TICKET_TTL: Duration = Duration::from_secs(300);
+
+/// Cap on simultaneously-live tickets, so a client that mints in a loop cannot
+/// grow the store without bound. The oldest ticket is evicted past this.
+const MAX_LIVE_TICKETS: usize = 32;
+
+/// One minted, not-yet-expired ticket.
+struct LiveTicket {
+    value: String,
+    expires_at: Instant,
+}
+
+/// Body returned by `POST /api/events/ticket`. (#5052)
+///
+/// `expires_in_secs` is advisory — it lets a client schedule a refresh instead
+/// of discovering expiry as a failed reconnect.
+#[derive(Debug, Clone, Serialize)]
+pub(super) struct EventTicketResponse {
+    pub(super) ticket: String,
+    pub(super) expires_in_secs: u64,
+}
+
+/// In-memory store of live event-stream tickets. (#5052)
+///
+/// Why: process-local and non-persistent on purpose — a ticket outliving the
+/// server that issued it has no legitimate use, and persisting it would put a
+/// credential on disk. A `Vec` rather than a `HashMap` because redemption is a
+/// constant-time comparison against every entry (see [`Self::redeem`]) and the
+/// store is capped at [`MAX_LIVE_TICKETS`].
+/// What: [`Self::mint`] issues a 256-bit random ticket; [`Self::redeem`]
+/// consumes one for a refreshed TTL window; both prune expired entries first.
+/// Test: `event_tickets::*` in `super::tests`.
+#[derive(Default)]
+pub(super) struct EventTicketStore {
+    live: Mutex<Vec<LiveTicket>>,
+}
+
+impl EventTicketStore {
+    /// Issue a new ticket, valid for [`TICKET_TTL`].
+    ///
+    /// Why: the caller has already proven it is allowed to read the stream (it
+    /// got past the same-origin write guard and, when one is configured, the
+    /// bearer-token middleware). This converts that one-shot proof into a
+    /// credential the browser can put in an `EventSource` URL.
+    /// What: draws 32 bytes from the OS CSPRNG, encodes them URL-safe (so the
+    /// value survives a query string unescaped), prunes expired entries, evicts
+    /// the oldest if the store is at capacity, and returns the ticket.
+    /// Test: `mint_returns_distinct_unguessable_tickets`,
+    /// `store_is_capacity_bounded`.
+    pub(super) fn mint(&self) -> EventTicketResponse {
+        let mut bytes = [0u8; 32];
+        rand::rngs::OsRng.fill_bytes(&mut bytes);
+        let value = base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(bytes);
+
+        let now = Instant::now();
+        let mut live = self.lock();
+        live.retain(|t| t.expires_at > now);
+        if live.len() >= MAX_LIVE_TICKETS {
+            // Evict the entry closest to expiry — the one whose loss costs the
+            // least. `min_by_key` over a <= 32-entry Vec is not worth a heap.
+            if let Some(idx) = live
+                .iter()
+                .enumerate()
+                .min_by_key(|(_, t)| t.expires_at)
+                .map(|(i, _)| i)
+            {
+                live.remove(idx);
+            }
+        }
+        live.push(LiveTicket {
+            value: value.clone(),
+            expires_at: now + TICKET_TTL,
+        });
+
+        EventTicketResponse {
+            ticket: value,
+            expires_in_secs: TICKET_TTL.as_secs(),
+        }
+    }
+
+    /// Whether `presented` is a live ticket; refreshes its TTL if so.
+    ///
+    /// Why: a ticket is not single-use because `EventSource` reconnects to the
+    /// SAME url after a dropped connection — invalidating on first use would
+    /// turn every transient network blip into a permanently dead stream.
+    /// Refreshing on redemption instead means an ACTIVELY reconnecting client
+    /// keeps working indefinitely while an abandoned ticket still ages out.
+    /// What: prunes expired entries, then compares `presented` against every
+    /// remaining ticket with the same constant-time comparison the bearer token
+    /// uses (#3761) — a short-circuiting `==` or a `HashMap` lookup would leak
+    /// the matching prefix to an attacker who can time their own requests.
+    /// Every entry is compared even after a hit, so the time taken does not
+    /// reveal WHICH ticket matched.
+    /// Test: `redeem_rejects_unknown_ticket`, `redeem_rejects_expired_ticket`,
+    /// `minted_ticket_is_redeemable_more_than_once`.
+    pub(super) fn redeem(&self, presented: &str) -> bool {
+        let now = Instant::now();
+        let mut live = self.lock();
+        live.retain(|t| t.expires_at > now);
+
+        let mut hit = None;
+        for (idx, ticket) in live.iter().enumerate() {
+            if bearer_token_matches(&ticket.value, presented) {
+                hit = Some(idx);
+            }
+        }
+        match hit {
+            Some(idx) => {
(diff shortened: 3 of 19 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review at `4dba7d44`, run from a throwaway worktree off this branch. Nothing edited, committed, or pushed.

**The headline: I could not defeat the ticket gate in the shipped default config.** I attacked it directly against a live binary and every vector closed. But the mechanism is inert in the *other* config — with a bearer token set, a valid ticket returns 401, which breaks the browser UI in exactly the deployment #5052 was filed about.

---

## What I verified myself, against running binaries

**The pre-fix exposure is real.** I reproduced it against the *installed* `tagent v0.38.6` (2026-08-04, pre-fix) rather than restoring code in place — independent of the PR's own before/after:

```
--- unauthenticated GET /api/events:
HTTP/1.1 200 OK
content-type: text/event-stream
access-control-allow-origin: *
--- cross-origin (Origin: https://evil.example):
HTTP/1.1 200 OK
content-type: text/event-stream
access-control-allow-origin: *
--- cross-origin GET /api/tasks:
HTTP/1.1 200 OK
access-control-allow-origin: *
```

`200` + `access-control-allow-origin: *` on a conversation stream. The issue's corrected severity was right.

**Attack matrix against this branch, `tagent --api --port 7732`, no token — the shipped sidecar default:**

| Vector | Result |
|---|---|
| unauth `GET /api/events` | `HTTP 401` |
| mint, `Origin: https://evil.example` | `403 cross-origin write requests are not allowed` |
| mint, `Origin: null` (sandboxed iframe, `file://`) | `403` |
| mint, DNS-rebind shape `Origin: http://evil.com:7732` | `403` |
| mint, `Content-Type: text/plain` (simple request, no preflight) | `403` |
| mint, `application/x-www-form-urlencoded` | `403` |
| mint, `multipart/form-data` | `403` |
| mint, `Origin: http://127.0.0.1.evil.com` | `403` |
| mint, `Origin: https://tauri.localhost.evil.com` | `403` |
| mint, `Origin: http://localhost.evil.com` | `403` |
| cross-origin `GET /api/tasks` | no `access-control-allow-origin` header |
| same-origin mint → stream (UI happy path) | `200 text/event-stream` |

The simple-request bypass I was looking for does not exist: browsers attach `Origin` to every POST regardless of content-type, and `guard_write_origin` is method-gated on POST, so the form-POST path 403s before CORS is even consulted. A `<form>` POST would also be unreadable by the attacker page (top-level cross-origin navigation). DNS rebinding fails too — the rebound request is same-origin from the browser's view, so `Origin: http://evil.com:7732` is sent and matches neither loopback nor `SelfOrigins`.

**`/api/config` — the retained exemption. Argument holds.** Verified against a token-configured server, not taken on faith:

```
$ curl http://127.0.0.1:7731/api/config
{"auth_required":true}
```

That is the whole body. `ApiClientConfig` (`auth.rs:201-204`) has exactly one field. The reasoning for keeping it exempt is sound and the divergence from the issue's suggested resolution is justified.

`/api/health` returns `{"status","version","pid","ppid"}`. `pid`/`ppid` is a small unauthenticated info leak, but it is unchanged code outside this diff — noted, not flagged.

**Test suite, this branch:**

```
cargo test -p trusty-agents --lib event_tickets
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 3398 filtered out
```

**Disclosed pre-existing failures — both confirmed outside this diff's import graph.** `ProjectFilesPanel.test.ts` imports only `./ProjectFilesPanel.svelte`; neither it, `MarkdownEditor.svelte`, nor `lib/projectFiles.ts` references `transport`, `connectEventSource`, or `mintEventTicket`. `tests/plain_cli_repl.rs` touches no `/api/*` path.

**Cross-crate CORS change is genuinely additive.** Every other consumer — `trusty-memory/src/web/mod.rs:234`, `trusty-analyze/src/service/routes.rs:133`, `trusty-review/src/service/mod.rs:135`, `trusty-search/src/service/server/mod.rs:313` — still calls `with_guarded_middleware`, which still builds the same permissive `CorsLayer` in the same layer order via the extracted `with_middleware_stack`. `with_guarded_middleware_same_origin_cors` has exactly one caller. No silent behaviour change.

**Taken on the engineer's word, not re-run:** the `pnpm test:unit` / `pnpm check` / `pnpm build` numbers, and the full `cargo test -p trusty-agents` / `-p trusty-common` counts.

---

## Findings

| HIGH | `crates/trusty-agents/src/api/server/routes.rs` | 320-325 | With a token configured, `auth_middleware` 401s `/api/events?ticket=<valid>` before the handler runs — the ticket is inert and the browser UI's `EventSource` can never open the stream | Exempt `/api/events` in `auth_middleware`; `EventStreamAuth::authorize` already checks the bearer token itself, so the route stays token-gated | Fix here |
| MEDIUM | `crates/trusty-agents/src/api/server/event_tickets.rs` | 151-155 | Sliding TTL has no absolute cap, so a holder who redeems every <5 min keeps the credential alive forever — the "five minutes" claim in the changelog and module doc is false as written | Add an absolute `minted_at`/deadline to `LiveTicket` and refuse redemption past it; correct the changelog wording | Fix here |
| MEDIUM | `crates/trusty-agents/ui/src-tauri/src/sse_bridge.rs` | 78-98 | Neither the mint POST nor the stream GET sends `Authorization`, and the resulting permanent failure is logged only at `tracing::debug!` — against a token-guarded sidecar the desktop shell shows no live events, with no visible error, retrying every 5s forever | Plumb the sidecar token into the bridge and send it on both calls; raise the mint failure to `warn!` | Fix here |
| MEDIUM | `crates/trusty-common/src/server/mod.rs` | 119, 202-203 | Two doc pointers name coverage that does not exist: `cross_origin_event_stream_is_not_cors_readable` appears nowhere in the tree, and the tests live in `api::server::tests::event_tickets`, not `::guard` | Point both at `api::server::tests::event_tickets` | Fix here |
| LOW | `crates/trusty-agents/src/api/server/events_sse.rs` | 59-62 | The ticket rides in the query string and `TraceLayer::new_for_http()`'s default span records `uri`, so at `RUST_LOG=debug` the full ticket lands in the daemon log — and per the row above it is renewable indefinitely once recovered | Redact the `ticket` parameter in the span, or note the `RUST_LOG` interaction on `TICKET_TTL` | Fix here |

---

## The HIGH, in detail

`build_router_with_origins` layers `Extension(EventStreamAuth)` first, then `auth_middleware`, then the guard/CORS stack. Since axum's `Router::layer` makes each successive layer *outer*, the request order is CORS → trace → compress → guard → `auth_middleware` → Extension → handler. `auth_middleware` (`auth.rs:169`) exempts only `/api/health` and `/api/config`, so `/api/events` needs a bearer header to reach `events_handler` at all — and `EventSource` cannot send one.

Live proof, `tagent --api --api-token secret123 --port 7731`:

```
=== C. mint WITH bearer (as the UI does) ===
{"ticket":"qTBscicYpXql8Ry0VLn80NUQCCFDRBm4gXRevJHTXYc","expires_in_secs":300}
=== D. EventSource-style GET with that ticket, NO auth header ===
HTTP 401
{"error":"unauthorized"}
=== E. control: same GET WITH bearer header ===
HTTP/1.1 200 OK
content-type: text/event-stream
```

The mint succeeds and hands out a ticket the server will never accept. The client then loops: `connectEventSource` re-mints and retries on every `onerror`, backing off to a 15s ceiling, forever.

This matters because `serve_with_config` refuses a non-loopback bind without a token (`routes.rs:381`). A token-configured server is therefore the *only* way to run trusty-agents off-loopback — the deployment #5052 was filed about — and the UI's live event stream is dead there. It fails closed, which is the safe direction, so this is HIGH and not CRITICAL. But the PR states the contract as "`GET /api/events` accepts a matching `Authorization: Bearer` **or** a live ticket," and the `or` does not hold.

No test covers it. `minted_ticket_opens_event_stream` and `minted_ticket_is_redeemable_more_than_once` both use `default_router()` (no token); `bearer_token_opens_event_stream` passes because `auth_middleware` let the header through, not because the ticket path works. The missing case is the whole mechanism under the one config where it is load-bearing.

## Required changes

1. Make the ticket work when a token is configured — exempt `/api/events` in `auth_middleware` and let `EventStreamAuth::authorize` be the sole gate for that path. Add `minted_ticket_opens_event_stream_with_token_configured`: `token_router("secret")`, mint with the bearer, then `GET /api/events?ticket=…` with **no** `Authorization` header, expect 200. It must fail against this commit.
2. Give a ticket an absolute lifetime alongside the sliding window, or amend the changelog and module doc to stop claiming a leaked ticket expires in five minutes.
3. Send the token from `sse_bridge.rs` on both the mint and the stream, and log a permanent mint failure above `debug`.
4. Fix the two doc pointers in `trusty-common/src/server/mod.rs`.

## Notes

- Repo gates I ran: `scripts/check_line_cap.sh` → `3801 files, 4 allowlisted, 0 violations — OK`. No `unwrap()`/`expect()` outside `#[cfg(test)]` in the new code (`EventTicketStore::lock`'s `unwrap_or_else(|e| e.into_inner())` is the correct poison-recovery form). Changelog fragments present for both crates. `Closes` targets only #5052, no epic.
- `bearer_token_matches` (`auth.rs:109-114`) is `subtle::ct_eq` behind a length guard, and `redeem` compares every live ticket even after a hit — the timing story checks out. Ticket entropy is `OsRng` × 32 bytes; capacity eviction drops the entry closest to expiry, so an actively-redeemed ticket is never the victim.
- **Not a bug, but worth stating:** any page served from *any* loopback origin — any local dev server, any locally-installed app on 127.0.0.1 — can mint a ticket and read the stream. That is strictly better than the `allow_origin(Any)` it replaces and consistent with the #3329 loopback-trust doctrine, so I am not flagging it. It is the boundary this fix draws, and it should be a conscious one.
- An established SSE connection is authenticated once at connect and never re-checked, so the 5-minute TTL does not kill a live stream. The client's re-mint-on-reconnect handles expiry correctly.
- The follow-up the PR names — sibling daemons still on `allow_origin(Any)`, `trusty-memory` in particular exposing palace content over GET — is real and I agree it belongs on its own ticket, not here.

bobmatnyc:
## Fix round — `b4f1d650`

All four findings addressed. The critic's live probes are re-run below on the
configuration that failed.

### HIGH — ticket inert whenever a token is configured

Correct, and the diagnosis was exact: `auth_middleware` layers outside the
routes, so `/api/events?ticket=<valid>` was rejected there before
`EventStreamAuth` ran. The server minted tickets it would always 401, and
`EventSource` cannot send the header that would have worked. Since
`serve_with_config` refuses a non-loopback bind without a token, that is the
only way to run off-loopback — the browser stream was dead exactly where #5052
was filed.

**Fix:** `/api/events` is exempt from `auth_middleware` again — but not
unauthenticated. `events_handler` accepts the same bearer token the middleware
would have checked, **or** a ticket, and 401s with neither. The exemption is an
exact path match, so `/api/events/ticket` stays fully covered by the middleware.

Live, `tagent --api --port 7701 --api-token s3cr3t-token`:

```
=== mint WITH bearer:
{"ticket":"z-D4gr5FW91jNbxbBJusVF8Hg2YRcMlhvH0x94mqinI","expires_in_secs":300}
=== stream with that ticket, NO auth header (what EventSource does) — was 401:
HTTP/1.1 200 OK
content-type: text/event-stream
=== stream with bearer header (unchanged):
HTTP 200
=== no credential at all (must stay 401):
{"error":"unauthorized"}
HTTP 401
=== forged ticket:                          HTTP 401
=== mint WITHOUT bearer (must stay 401):    HTTP 401
=== cross-origin mint (must stay 403):      HTTP 403
=== other /api routes still bearer-gated:
/api/tasks no-auth:            HTTP 401
/api/tasks with ticket only:   HTTP 401
```

### The test gap — the real finding

Every ticket success case used `default_router()`, which has no token, so none
of them exercised `auth_middleware`. `bearer_token_opens_event_stream` passed
because the **header** cleared the middleware, not because the ticket path
worked. The suite was green over a dead path.

Three cases added against a token-configured router:

- `ticket_opens_event_stream_on_a_token_configured_server`
- `bogus_ticket_is_rejected_on_a_token_configured_server`
- `expired_ticket_is_rejected_by_stream_auth_with_a_token_configured`

**Fails against the previous head** (exemption removed again, MEDIUM-1 cap reverted):

```
---- ...::ticket_opens_event_stream_on_a_token_configured_server stdout ----
  left: 401
 right: 200

---- ...::sliding_ttl_cannot_outlive_the_absolute_cap stdout ----
a refreshed idle window must not outlive MAX_TICKET_LIFETIME

test result: FAILED. 18 passed; 2 failed; 0 ignored; 0 measured; 3398 filtered out
```

After:

```
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 3393 filtered out
```

### MEDIUM-1 — unbounded sliding TTL

`MAX_TICKET_LIFETIME` (1 hour) now sits alongside the 5-minute sliding window.
`is_live` requires both deadlines, and a redemption clamps the refreshed idle
deadline with `.min(expires_hard_at)` so it can never push past the cap. An hour
self-heals: a live client's next reconnect 401s and both shipped clients
re-mint. Pinned by `sliding_ttl_cannot_outlive_the_absolute_cap`.

Changelog wording corrected — "five minutes of read-only stream access" was
false as written; it now states 5 minutes idle **and** 1 hour absolute.

### MEDIUM-2 — desktop bridge sent no credential, failed silently

`sse_bridge.rs` now reads `TAGENT_API_TOKEN` (with the `OPEN_MPM_API_TOKEN`
compat alias) — the sidecar inherits the same environment, so it is the same
token the server checks — and attaches it via a `with_bearer` helper to **both**
the mint and the stream request. The permanent-failure log moved `debug!` →
`warn!`; a token-guarded sidecar used to kill the desktop stream with nothing
visible in default logs. `cargo check -p trusty-agents-ui` → `UICHECK=0`.

### MEDIUM-3 — dangling doc pointers

`same_origin_cors` and `with_guarded_middleware_same_origin_cors` cited
`cross_origin_event_stream_is_not_cors_readable` in `::guard`; neither existed.
Both now point at `api::server::tests::event_tickets` —
`cross_origin_request_gets_no_cors_reflection` and
`loopback_origin_is_cors_reflected`.

### Gates

```
cargo fmt --check                                              FMT=0
cargo check -p trusty-agents -p trusty-common                  CHECK=0
cargo check -p trusty-agents-ui                                UICHECK=0
cargo clippy -p trusty-common --all-targets --features axum-server -- -D warnings
                                                               CL_COMMON=0
cargo clippy -p trusty-agents --all-targets -- -D warnings     CL_AGENTS=0

cargo test -p trusty-common --features axum-server
test result: ok. 312 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out

cargo test -p trusty-agents
test result: ok. 3407 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out
test result: ok. 7 passed; 0 failed; 0 ignored   # plain_cli_repl, green this run
(+ 8 further integration binaries, all ok)

check_line_cap.sh       0 violations
check_test_pointers.sh  22629 citations, 0 dangling
check_sld.sh            0 errors, 0 warnings
```

Nothing the review cleared was touched: the `/api/config` exemption, the
same-origin write guard, CORS additivity, and mint-is-a-POST are unchanged.

bobmatnyc:
## Code Critic — round 2, independent reviewer, head `b4f1d650`

## Verdict: APPROVE

Zero CRITICAL, zero HIGH. Three findings, all MEDIUM/LOW, none of them a gate.

I reviewed the code against issue #5052 and the diff only. I did not read the round-1 verdict text before forming my own view of the exemption, the handler, and the TTL cap.

---

## What I verified myself

**1. The re-exemption does not open a hole.** `auth.rs:188` is an exact `path == "/api/events"` equality, and `req.uri().path()` is the raw, undecoded path. I built a scratch worktree at this head and probed 21 path variants against `token_router("secret")` with no credential. Nothing reached the mint route or the stream:

```
MINTPROBE POST /api/events/ticket            -> 401 Unauthorized
MINTPROBE POST /api/events/ticket/           -> 401 Unauthorized
MINTPROBE POST //api/events/ticket           -> 405 Method Not Allowed
MINTPROBE POST /api//events/ticket           -> 401 Unauthorized
MINTPROBE POST /api/events%2fticket          -> 401 Unauthorized
MINTPROBE POST /api/events/./ticket          -> 401 Unauthorized
MINTPROBE POST /api/events/../events/ticket  -> 401 Unauthorized
MINTPROBE POST /API/EVENTS/TICKET            -> 405 Method Not Allowed
MINTPROBE POST /api/Events/ticket            -> 401 Unauthorized
MINTPROBE POST /api/events/ticket?x=1        -> 401 Unauthorized
MINTPROBE POST /api/events/ticket#frag       -> 401 Unauthorized
MINTPROBE POST /./api/events/ticket          -> 405 Method Not Allowed
STREAMPROBE GET /api/events                  -> 401 Unauthorized
STREAMPROBE GET /api/events/                 -> 401 Unauthorized
```

One probe looked alarming and is not: `GET //api/events` returns 200, because a double leading slash makes `!path.starts_with("/api/")` true, so the middleware exempts it — and matchit then routes it to the SPA catch-all, not the stream. I dumped the bodies to be sure:

```
DS [token] GET //api/events           -> 200 OK ct=text/html len=3745 head="<!DOCTYPE html>\n<html lang=\"en\">…"
DS [token] GET //api/events?ticket=x  -> 200 OK ct=text/html len=3745 head="<!DOCTYPE html>\n<html lang=\"en\">…"
DS [token] GET //api/events/ticket    -> 200 OK ct=text/html len=3745 head="<!DOCTYPE html>\n<html lang=\"en\">…"
DS [token] GET //api/tasks            -> 200 OK ct=text/html len=3745 head="<!DOCTYPE html>\n<html lang=\"en\">…"
```

It is `index.html` for every unknown GET path, `//api/tasks` included — pre-existing SPA-fallback behaviour, not introduced here and not a disclosure.

**2. `events_handler` fails closed on every branch I could construct.** No credential, empty `?ticket=`, forged ticket, one-char-truncated ticket, idle-expired ticket, hard-expired ticket, wrong bearer — all 401. A malformed query (`deny_unknown_fields`) rejects at the `Query` extractor with 400 *before* the auth check, which is also closed. The 401 returns before `events::subscribe()`, so an unauthorized caller never holds a receiver. `EventStreamAuth` is delivered by an `Extension` layer registered unconditionally (`routes.rs:317`), and `build_router_with_origins` is the only router construction path in the crate that serves `events_handler` — there is no second router that could miss it.

**3. The absolute TTL cap holds.** `is_live` (`event_tickets.rs:69`) requires both deadlines independently, and `redeem` is the only writer of `expires_at` outside `mint`, so `.min(hard)` at line ~180 is the only refresh path and is in fact belt-and-braces — even with the clamp deleted, `is_live`'s hard check alone would enforce the cap. `expires_hard_at` never moves after mint. Deadlines are `Instant`, so wall-clock manipulation and NTP steps cannot extend a ticket; `now + 300s` / `now + 3600s` cannot saturate. A fresh `mint()` grants a fresh hour, correctly — minting requires clearing the same-origin guard and, when configured, the bearer.

**4. `TAGENT_API_TOKEN` is the right variable.** `sidecar.rs:158-167` spawns `tagent --api --port N --parent-pid X` with no `--api-token` and inherits the shell's environment, and the server resolves its token from `env_compat::env_var("TAGENT_API_TOKEN", "OPEN_MPM_API_TOKEN")` (`mode_dispatch.rs:214`, `startup.rs:342`). `sse_bridge.rs:64-67` reads the same pair in the same order. Correct — with one degenerate-config divergence, LOW-3 below.

**5. Red proof reproduced.** I restored round-1's `auth.rs` onto this head (`git checkout 4dba7d44 -- crates/trusty-agents/src/api/server/auth.rs`), keeping the new tests, and ran `cargo test -p trusty-agents --lib api::server::tests::event`:

```
test api::server::tests::event_tickets::ticket_opens_event_stream_on_a_token_configured_server ... FAILED
  left: 401
 right: 200
test result: FAILED. 26 passed; 1 failed; 0 ignored; 0 measured; 3391 filtered out; finished in 0.02s
```

Same command at `b4f1d650`:

```
test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 3391 filtered out; finished in 0.02s
```

My counts differ from the 18/2 → 25/0 in the dispatch note because my filter spans `event_tickets` + `events_sse`; the module alone is `20 passed; 0 failed`. `sliding_ttl_cannot_outlive_the_absolute_cap` passes in my BEFORE run because I reverted only `auth.rs`; its red proof is against `origin/main`, where `MAX_TICKET_LIFETIME`, `expires_hard_at` and `is_live` do not exist at all — mechanically certain from the diff, and I took that one on the engineer's word rather than rebuilding a third state.

**6. The three new tests are genuinely token-configured, and nothing was weakened.** `ticket_opens_event_stream_on_a_token_configured_server`, `bogus_ticket_is_rejected_on_a_token_configured_server` and `expired_ticket_is_rejected_by_stream_auth_with_a_token_configured` all use `token_router("secret")` / `EventStreamAuth::new(Some(...))`, and the first provably goes red when the exemption regresses (above). The pre-existing `events_sse` helper was tightened, not narrowed: `get_events` now mints a real ticket, with a comment explaining that a bare 401 would otherwise have satisfied `is_client_error()` for the wrong reason.

**7. Gates.**

```
scripts/check_line_cap.sh       → 3801 files, 4 allowlisted, 0 violations — OK
scripts/check_test_pointers.sh  → 22629 citations, 0 dangling — OK
scripts/check_changelog_fragment.sh → trusty-agents OK, trusty-common OK
```

No `unwrap()` / `expect()` outside `#[cfg(test)]` in the new Rust (`event_tickets.rs`, `events_sse.rs`, `auth.rs` all clean; `lock()` recovers from poisoning with `unwrap_or_else(|e| e.into_inner())`). `Closes` targets #5052 and nothing else. The changelog fragment wording round 1 flagged is corrected — it now reads "expires 5 minutes after its last use and 1 hour after issue whatever happens in between". Both doc pointers in `trusty-common/src/server/mod.rs` resolve: `with_standard_middleware_composes` is at `mod.rs:271`, and the trusty-agents end-to-end pointers name `cross_origin_request_gets_no_cors_reflection` / `loopback_origin_is_cors_reflected`, both real.

**Taken on the engineer's word:** the live-binary smoke transcript, the UI gates (`pnpm test:unit`, `pnpm check`, `pnpm build`), the full `cargo test -p trusty-agents` / `-p trusty-common` runs, and the round-1-cleared items I was told not to re-litigate (`/api/config` body, the mint write guard, CORS additivity across siblings).

---

## Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-agents/ui/src-tauri/src/sse_bridge.rs` | 74, 130-139 | A live ticket is written to the desktop shell's default logs on every stream drop | Strip the URL from the reqwest error before formatting | Fix here |
| LOW | `crates/trusty-agents/src/api/server/auth.rs` | 173-176 | Comment directly above the security branch contradicts the branch | Delete the stale sentence | Fix here |
| LOW | `docs/trusty-agents/user/cli-reference.md` | 41, 75 | Docs still describe `/api/events` as reachable without a credential | Note the ticket requirement | Fix here |
| LOW | `crates/trusty-agents/ui/src-tauri/src/sse_bridge.rs` | 64-67 | Bridge's token resolution diverges from the server's on an empty value | Reject an empty configured token server-side | Parent |

### MEDIUM-1 — the ticket lands in the log

`sse_bridge.rs:130` builds the stream URL with the credential in the query:

```rust
let url = format!("{base}/api/events?ticket={ticket}");
let mut resp = with_bearer(client.get(&url), token)
    .send().await.map_err(|e| format!("connect: {e}"))?     // 134
    .error_for_status().map_err(|e| format!("status: {e}"))?; // 136
while let Some(chunk) = resp.chunk().await.map_err(|e| format!("read: {e}"))? { // 139
```

`reqwest::Error`'s `Display` appends the full URL, query included — `error.rs:279-281`, and the type's own doc says so: *"Errors may include the full URL used to make the Request. If the URL contains sensitive information (e.g. an API key as a query parameter), be sure to remove it (`without_url`)."* All three strings propagate to line 74:

```rust
tracing::warn!(?e, "event bridge stream failed; retrying");
```

This same commit raised that from `debug!` to `warn!`, so a still-live ticket now reaches the default log on every routine stream end — sidecar restart, idle reap, machine wake. This is the exact leak channel the PR's own design rationale cites when rejecting `?token=`: *"the URL is the one part of the request that leaks — into access logs, the `Referer` header, and browser history."* The blast radius is genuinely small (local log, same user, read-only, ≤5 min idle / ≤1 h absolute), which is why it is MEDIUM and not HIGH.

```rust
.map_err(|e| format!("connect: {}", e.without_url()))?
.error_for_status().map_err(|e| format!("status: {}", e.without_url()))?;
… .map_err(|e| format!("read: {}", e.without_url()))?
```

Do the same for `mint status:` at line 101 for symmetry (that URL is clean today, but it is one refactor away from carrying a query).

### LOW-2 — contradictory comment on the security branch

`auth.rs:173-176` says the exemption is gone; `auth.rs:177-186` says it is back and why. Both sit in one comment run immediately above the `if`. The next reader of a security-critical branch should not have to work out which half is current. Delete lines 173-176; the block below already says everything.

### LOW-3 — `cli-reference.md` still documents the old contract

Line 41 says `--api-token` requires a bearer *"on every `/api/*` request (except `/api/health` and `/api/config`)"* and line 75 documents the env var, neither mentioning that `/api/events` needs a ticket and is gated even with no token configured. An operator following the documented `curl -N http://localhost:7654/api/events` now gets an unexplained 401. One line in the table plus the two-command recipe already written in `events_sse.rs`'s doc comment.

### LOW-4 — empty-token divergence

`sse_bridge.rs:64-67` applies `.filter(|t| !t.is_empty())`; the server's `env_compat::env_var` does not, so `TAGENT_API_TOKEN=""` gives a server with `auth_middleware` installed over an empty token while the bridge presents no bearer at all — the desktop stream then 401s permanently, which is the same failure class round 1 found. (`.or_else(|_| …)` also falls back on `NotUnicode` where `env_compat` falls back only on `NotPresent`.)

I confirmed the underlying hazard directly: on `token_router("")`, a request carrying `Authorization: Bearer ` (empty) gets 200 on both the mint route and the stream. That is pre-existing — `bearer_token_matches`'s own doc notes an empty configured token is not special-cased — and out of scope here. The right fix is server-side: treat `Some("")` as no token, or refuse to start. Not this PR's job; flagging so it is not lost.

---

## Notes

- I did not re-litigate the round-1-cleared items (`/api/config` exemption body, cross-origin mint / `Origin: null` / DNS-rebind / three content-types, CORS additivity, mint-is-a-POST) — this head does not touch that code.
- `Instant` rather than `SystemTime` for both deadlines is the right call and worth keeping if this store is ever refactored: it makes clock-skew and NTP-step attacks on the TTL structurally impossible.
- I created and removed my own scratch worktree (`critic-5185-r2-redproof`) for the red proof; the review worktree `critic-5185-r2` is untouched, and I did not go near `critic-5191`.

bobmatnyc:
## `17fb0b6b` — MEDIUM + all three LOWs

### MEDIUM — the ticket reached the default log

Correct, and this branch is what made it true. `reqwest::Error`'s Display
appends the request URL including the query string, the stream URL carries
`?ticket=<live credential>`, and the previous commit raised that log from
`debug!` to `warn!`. A still-valid ticket was written on every routine stream
drop — the exact leak channel this PR's rationale cites when rejecting a
`?token=<bearer>` design.

All five propagated errors now go through one helper:

```rust
fn redacted(context: &str, e: reqwest::Error) -> String {
    format!("{context}: {}", e.without_url())
}
```

The cause survives; only the URL goes.

**A test was cheap.** `redacted_error_does_not_leak_the_ticket` connects to
`127.0.0.1:1` — unbound and privileged, so it fails instantly with no server
and no timing dependence — and asserts **both** directions:

- the raw `format!("{err}")` **contains** the ticket, so the test cannot pass
  vacuously if reqwest ever stops embedding the URL;
- `redacted(...)` does not.

Red proof, `without_url()` reverted:

```
---- sse_bridge::tests::redacted_error_does_not_leak_the_ticket stdout ----
panicked at crates/trusty-agents/ui/src-tauri/src/sse_bridge.rs:304:9:
a live ticket must never reach the log: connect: error sending request for url (http://127.0.0.1:1/api/events?ticket=s3cret-ticket-value)
```

Green after: `test result: ok. 25 passed; 0 failed` (`cargo test -p trusty-agents-ui`).

### The three LOWs — all taken, all trivial

- **`cli-reference.md:41`** documented `--api-token` as exempting only
  `/api/health` and `/api/config`, which left `/api/events` reading as
  credential-free — user-facing and false on merge. It now states the stream
  takes the bearer token **or** a `POST /api/events/ticket` ticket, and needs
  one whether or not a token is configured.
- **`auth.rs`** — a stale comment above the exemption branch still said
  `/api/events` was "not one any more", contradicting both the branch under it
  and the comment beside it. Deleted.
- **Empty-token divergence** — dropped `.filter(|t| !t.is_empty())` so the
  bridge mirrors `runtime::mode_dispatch` exactly. An empty `TAGENT_API_TOKEN`
  configured a token server-side that the bridge would then never present.

### Gates

```
cargo fmt --check                                            FMT=0
cargo clippy -p trusty-agents --all-targets -- -D warnings   CL_AGENTS=0
cargo clippy -p trusty-agents-ui --all-targets -- -D warnings  CL_UI=0

cargo test -p trusty-agents-ui
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-agents
test result: ok. 3407 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out
(+ 8 integration binaries, all ok, incl. plain_cli_repl 7 passed)

cargo test -p trusty-common --features axum-server
test result: ok. 312 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out

check_line_cap.sh       0 violations
check_test_pointers.sh  22630 citations, 0 dangling
check_sld.sh            0 errors, 0 warnings
```

No behaviour change to the auth or CORS paths this review verified — the diff
is the bridge's error formatting, its token resolution, one deleted comment,
and one doc row.

Issue #5181, opened 2026-08-08
#5089 step 6: retire the two legacy HTTP webhook routes — blocked on the UDS receivers and the inbox drain


bobmatnyc:
Blocked by [#5182](https://github.com/bobmatnyc/trusty-tools/issues/5182) and [#5192](https://github.com/bobmatnyc/trusty-tools/issues/5192) — the UDS receivers and the inbox drain this route retirement depends on do not exist yet.

Defect: both legacy HTTP webhook routes that [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089) (reopened — it auto-closed before these steps landed) exists to eliminate are still registered and unchanged on `origin/main` at `352fe5d6`, so GitHub can still reach the old ack-then-drop path that bypasses the console's durable spool entirely.

- trusty-analyze: `POST /webhooks/github` -> `handlers::review::github_webhook_handler`, registered at `crates/trusty-analyze/src/service/routes.rs:118-121`. The handler still spawns work off the request path and only logs on failure — `crates/trusty-analyze/src/service/handlers/review.rs:213-217` (`tracing::warn!` inside `tokio::spawn`).
- trusty-review: `POST /pr/github/webhook` -> `handle_github_webhook`, registered at `crates/trusty-review/src/service/mod.rs:133`.

Neither is removed, deprecated, or flag-gated. `git log origin/main` on both route files shows no commits since #5089 opened. `crates/trusty-console/src/webhook/mod.rs:1-11` names these two as "the two handlers this path replaces" and cites their fail-open lines. GitHub does not retry an acknowledged delivery, so every delivery landing on the old path can be lost silently.

BLOCKING OPEN QUESTION, must be resolved before implementation: ADR-0034's Open Questions (`docs/adr/0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md:316-323`) still asks whether trusty-analyze's webhook route survives at all — "that is a product question, not a transport one." Scope is therefore undecided: retire both, or only trusty-review's. Do not let an implementer pick silently.

Resolution: resolve the open question above first, then retire the route(s) in scope.

Rung: 5. Requires a code-critic round before merge.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Correction: this was mis-scoped when filed, as PM. Retitled from "step 4: retire..." to "step 5: retire..., blocked on the UDS receivers".

[...]

PR #5191, opened 2026-08-08
fix(trusty-analyze,trusty-review): bind the webhook UDS listeners console relays to (#5089 step 4)

### LINKED ISSUE #5182: #5089 step 4: bind the webhook UDS listeners in trusty-analyze and trusty-review — console's relay has no receive side
Defect (verified on origin/main at [352fe5d6](https://github.com/bobmatnyc/trusty-tools/commit/352fe5d675dc48c739420c53794ae3779b06a700)):

- `WebhookIngress::from_env` targets two sockets that nothing binds — `UdsRelay::new(sockets.join("trusty-review-webhook.sock"))` and the analyze equivalent, `crates/trusty-console/src/webhook/mod.rs`. A repo-wide grep for both socket filenames returns only those two lines: no binder, no launchd plist, no config sample.
- Neither `trusty-analyze` nor `trusty-review` has a UDS listener: zero `UnixListener` bind sites and zero `trusty_common::uds::bind_hardened` calls anywhere in either crate. Workspace grep finds `bind_hardened` only in trusty-agents, trusty-bm25-daemon, trusty-common, trusty-embedderd, trusty-memory, and trusty-console's own tests.
- `trusty_common::uds::send_framed_request` has exactly one call site outside its own tests — `crates/trusty-console/src/webhook/relay.rs`, the sender (`UdsRelay::deliver`). There is no receive side.
- `trusty_common::webhook_relay::RELAY_METHOD` has zero consumers outside `trusty-common` and `trusty-console`. The `RelayFrame`/`RelayResponse` contract is defined and sent, never received.
- `trusty-review`'s own binary still documents itself as a long-lived HTTP webhook server on port 7891 (`crates/trusty-review/src/main.rs`, the `Serve` subcommand doc), including `POST /pr/github/webhook`.
- The relay module's own doc names this gap and the step: `crates/trusty-console/src/webhook/relay.rs` — "spawn-on-demand is #5089 step 4's job"; `crates/trusty-console/src/webhook/mod.rs` — "Until then no target binds a listener and every relay lands in `RelayOutcome::Unreachable`, which is a durable pending state, not a dropped delivery."

Resolution — in both `trusty-analyze` and `trusty-review`:
1. Bind a `UnixListener` via `trusty_common::uds::bind_hardened` at the socket paths console already targets (`WebhookIngress::from_env` in `crates/trusty-console/src/webhook/mod.rs`).
2. Serve the `RELAY_METHOD` = `"webhook.deliver"` frame contract from `crates/trusty-common/src/webhook_relay.rs`. Reject any other method. Answer `RelayResponse` with `ack: true` only once the work is durably owned — `RelayOutcome::Acked` is the sole state that lets console delete a spool entry, so a premature ack reintroduces the silent-loss defect [ADR-0034](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md) exists to eliminate.
3. Add the supervised on-demand spawn ADR-0034 specifies — these sockets must exist without the services running resident.

Rung 5 (cross-crate contract, process lifecycle). Owes a code-critic round.

🔴 This BLOCKS [#5181](https://github.com/bobmatnyc/trusty-tools/issues/5181). Retiring the HTTP routes before this lands turns a lossy-but-working path into a completely non-functional one — deliveries would spool forever, unacked, with no review or analysis ever running.

Part of [#5089](https://github.com/bobmatnyc/trusty-tools/issues/5089).

bobmatnyc:
Closes https://github.com/bobmatnyc/trusty-tools/issues/5182
Part of https://github.com/bobmatnyc/trusty-tools/issues/5089

## The defect

Console's webhook relay sent `webhook.deliver` frames to two Unix sockets that nothing bound.

- `WebhookIngress::from_env` built `UdsRelay::new(sockets.join("trusty-review-webhook.sock"))` and the analyze equivalent in [`crates/trusty-console/src/webhook/mod.rs`](https://github.com/bobmatnyc/trusty-tools/blob/352fe5d675dc48c739420c53794ae3779b06a700/crates/trusty-console/src/webhook/mod.rs#L234-L239). A repo-wide grep for either filename returned only those two lines.
- Zero `UnixListener` bind sites and zero `bind_hardened` calls in `crates/trusty-analyze/` or `crates/trusty-review/`.
- [`trusty_common::webhook_relay::RELAY_METHOD`](https://github.com/bobmatnyc/trusty-tools/blob/352fe5d675dc48c739420c53794ae3779b06a700/crates/trusty-common/src/webhook_relay.rs#L29) had no consumer outside `trusty-common` and `trusty-console`. The contract was defined and sent, never received.

Every delivery landed in `RelayOutcome::Unreachable` and stayed pending forever. Durable, but never reviewed and never analysed.

## What this does

The receive half lives once, in [`trusty_common::webhook_relay`](https://github.com/bobmatnyc/trusty-tools/blob/fix/5182-webhook-uds-listeners/crates/trusty-common/src/webhook_relay/serve.rs) — the module both the sender and the two receivers already share — because the property that matters is a single ordering rule, and two copies of an ordering rule is one copy plus a future regression.

- `Inbox` writes one JSON file per delivery: temp file with `create_new` → `sync_all` → `hard_link` into place → `sync_all` the directory. `hard_link` rather than `rename` so a redelivery collides with `EEXIST` instead of silently overwriting.
- `serve::dispatch_frame` parses, rejects a non-`2.0` envelope, rejects any method other than `webhook.deliver` **by name**, rejects a frame claiming `provenance.verified = false`, and only then takes ownership.
- `WebhookListener` is the whole body of a supervised child: bind through `bind_singleton_hardened`, check the peer uid on accept, serve, exit on SIGTERM, unlink.
- `trusty-review webhook-listen` and `trusty-analyze webhook-listen` mount it. Socket paths come from `webhook_relay::{review_socket_path, analyze_socket_path}` rather than a literal per crate, so the sender and receivers cannot disagree about a path — the failure that disagreement produces is a delivery that never arrives and never errors.
- Console's `webhook::spawn::TargetSupervisor` wraps the shared `UdsServiceSupervisor` promoted in #5170 and runs `ensure_running` before each relay, so neither target is resident. `TRUSTY_WEBHOOK_TARGET_EXTERNAL=1` hands lifecycle back to `tctl` (ADR-0011). A target that will not start is `Unreachable` — the spool entry stays.

The legacy HTTP webhook routes are untouched. Retiring them is https://github.com/bobmatnyc/trusty-tools/issues/5181.

## The ack-ordering guarantee

`RelayOutcome::Acked` is the sole state that lets console delete a spool entry, and that entry is the only remaining copy — GitHub does not re-send a delivery it has seen acknowledged. So an ack sent before the work is durably owned does not make the failure less likely, it makes it unrecoverable.

The guarantee is structural: `dispatch_frame`'s ack arm is reachable only from `Ok` out of `DeliverySink::take_ownership`, and `Inbox::take_ownership` has fsync'd both the file and the directory entry before it returns. A failure at any point returns a JSON-RPC error, which console's own `RelayResponse::is_ack` reads as not-an-acknowledgement.

Redelivery of an id already held is still an ack. Relay is at-least-once by construction, so refusing a delivery we already own would wedge the sender's spool forever.

### How the required tests prove it

| Test | What it pins | If the ack were sent before durable ownership |
|---|---|---|
| `delivery_reaches_the_real_target_and_is_spooled_there_before_the_ack` (`trusty-console`) | Console ingests a signed delivery, the **production** receiver acks, the receiver's inbox holds the body byte-exact, and console's spool is cleared | The inbox assertion runs after the ack has already been received. Ack-first makes the write concurrent with these assertions, so `inbox.list()` can be empty — while console has already deleted its own copy, which is the unrecoverable state |
| `dispatch_acks_only_after_the_sink_has_taken_ownership` (`trusty-common`) | The sink is consulted exactly once, and the ack is downstream of its `Ok` | Fails directly: an ack hoisted above the call leaves `sink.calls() == 0` |
| `a_receiver_that_cannot_own_the_work_never_produces_an_ack` (`trusty-console`) | Against a sink that always refuses: the outcome is not `Acked`, the reason carries the receiver's own words, the spool entry survives, and `attempts` is 1 on disk | Fails at the first assertion. Ack-first cannot distinguish a receiver that owned the work from one that could not, so console deletes the only copy and the spool assertion goes to zero |
| `dispatch_refuses_when_the_sink_cannot_take_ownership` (`trusty-common`) | Same at the decision level, with the `-32000` code | Same |
| `the_real_listener_rejects_a_method_that_is_not_webhook_deliver` (`trusty-console`) | `webhook.execute` gets `-32601` naming both what arrived and what is served, and nothing is stored | An ack that precedes the method check would take responsibility for a payload no sender in this workspace produces |
| `dispatch_rejects_a_method_that_is_not_webhook_deliver` (`trusty-common`) | The sink is never reached at all (`calls() == 0`) | Same |

**Correction to an earlier version of this body.** It claimed the two console E2E cases "fail at their first assertion" against [352fe5d6](https://github.com/bobmatnyc/trusty-tools/commit/352fe5d675dc48c739420c53794ae3779b06a700). They do not fail — they **do not compile**. That commit has no `serve_until`, no `Inbox` and no `DeliverySink`, and the tests drive the production receiver this PR introduces. So **all four** console cases have no pre-fix counterpart, not two. What I did verify against the pre-fix `from_env` is the substance: nothing bound either socket, so every relay landed in `Unreachable`. The `trusty-common` rows above are the ones that fail as described, because they exercise the decision function directly.

## Fix round — critic verdict WARN, one HIGH and eight others

Full verdict: https://github.com/bobmatnyc/trusty-tools/pull/5191#issuecomment-5226824049

**HIGH — the health signal inverted.** An acknowledged delivery is deleted from console's spool, so `pending` dropped to zero and `ServiceHealth` went `Degraded` → `Ok` while the delivery sat unprocessed in a directory nothing read. Before the listeners existed an undrained backlog was at least visible as a pending entry; this PR would have made it invisible and reported healthy. `GET /api/console/metrics/webhooks` now carries `undrained` (per source) and `undrained_total`: a held delivery is `Degraded`, an inbox that cannot be counted is `Error`, and `held_count` never creates or chmods the directory it meters. Pinned by `health_is_degraded_while_a_delivery_sits_undrained`, which asserts `Ok` on an empty spool and empty inbox, `Degraded` after the ack has cleared the spool, and `Ok` again once the entry is drained; plus `health_is_error_when_a_targets_inbox_cannot_be_counted` and `from_env_meters_every_targets_inbox`. The drain step itself is https://github.com/bobmatnyc/trusty-tools/issues/5192.

| Finding | Fix |
|---|---|
| Inline accept violates `UdsServiceSupervisor`'s stated accept-promptly REQUIREMENT — a saturated queue answers ECONNREFUSED on macOS and the child is evicted | `serve_until` spawns each handler; `serve_handles_concurrent_connections_without_serialising` deadlocks against the inline version |
| The supervisor's liveness probe connects and closes on every `ensure_running`, firing the WARN an operator greps for on every successful relay | Zero-byte reads return `Served::LivenessProbe` and log at `debug`. `handle_connection_reports_a_liveness_probe_rather_than_a_failure`; the smoke below shows one probe line and zero WARNs |
| `already_owned` acked on a bare `EEXIST` — a filename match, not a delivery match | `take_ownership` reads the held copy back and refuses a mismatch (`InboxError::KeyCollision`). The key is a truncated SHA-256 rather than FNV-1a, but the ack rests on the read-back. `inbox_refuses_a_different_delivery_on_the_same_path`, `inbox_refuses_a_reused_id_carrying_a_different_body`, `inbox_refuses_when_the_held_copy_is_unreadable` |
| Socket unlinked after the listener was dropped — the #5085 window in which a successor's fresh socket is deleted by its predecessor | `serve_until` borrows the listener; `run` unlinks while it is still bound |
| `bind_singleton_hardened` treated every connect error as "corpse" | Takes over only on a proved `SocketVerdict::NotServing`; `Inconclusive` refuses. The classification moved from `uds::supervisor::probe` to `uds::probe` (re-exported) so both callers share one copy |
| `RelayRequest.id` never compared to `params.delivery_id` | Rejected with `-32602` before `take_ownership`. `dispatch_rejects_a_frame_whose_id_disagrees_with_its_delivery_id` |
| A 1874-line npm lockfile and a rebuilt UI bundle rode in on the listener commit | Removed; `git diff origin/main...HEAD -- crates/trusty-analyze/ui` is now empty |

**One defect the spawn fix exposed.** With handlers serialised, two writers could never share a nanosecond. Spawned, they can: the temp filename was `pid.nanos.tmp`, `create_new` failed with `EEXIST`, and a perfectly good delivery was **refused as undurable** because a sibling was concurrent. `serve_concurrent_deliveries_of_one_id_produce_one_stored_copy` caught it immediately. The temp name now carries a process-wide counter.

**Coverage gap, stated rather than papered over.** Console's spawn-a-target path — the thing `TargetSupervisor` exists for — has no automated end-to-end coverage. `spawn_adopts_a_socket_that_is_already_served` covers adoption and deliberately never resolves a binary, and the `#[ignore]`d integration test sets `TRUSTY_WEBHOOK_TARGET_EXTERNAL=1` so it does not launch a real `trusty-review` from the developer's PATH. The spawn itself rests on the manual smoke below. Covering it properly needs a fixture binary rather than a PATH-resolved one.

## Gates — rung 5 (cross-crate contract, process lifecycle)

```
cargo fmt --check                                                                  EXIT=0
cargo check -p trusty-analyze -p trusty-review -p trusty-console -p trusty-common   EXIT=0
cargo clippy -p trusty-analyze -p trusty-review -p trusty-console --all-targets -- -D warnings   EXIT=0
cargo clippy -p trusty-common --features webhook-relay,uds-supervisor --all-targets -- -D warnings   EXIT=0

cargo test -p trusty-common --features webhook-relay,uds-supervisor -- --include-ignored
  test result: ok. 430 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-console -- --include-ignored
  test result: ok. 228 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo test -p trusty-review
  test result: ok. 1588 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out

cargo test -p trusty-analyze
  test result: ok. 461 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

bash scripts/check_line_cap.sh            0 violations — OK
bash scripts/check_test_pointers.sh       22664 citations, 0 dangling — OK
bash scripts/check_changelog_fragment.sh  all 4 crates with source changes recorded
bash scripts/check_sld.sh                 0 errors, 0 warnings
```

Concurrency and failure paths: `serve_concurrent_deliveries_of_one_id_produce_one_stored_copy` (8 racing relays, one stored copy, every attempt acked), `serve_rejects_an_oversized_frame`, `serve_keeps_answering_after_a_refusal`, `inbox_persist_fails_when_the_root_is_not_a_directory`, `bind_singleton_refuses_a_socket_someone_is_serving`.

### Binary smoke against the real `trusty-review` binary

A liveness probe (connect, close) first, then five frames:

```
--- socket ---
srw-------  1 masa  staff  0 Aug  8 12:01 /var/folders/.../T/trusty-502/trusty-review-webhook.sock
ACK:           {"result":{"ack":true}}
REDELIVERY:    {"result":{"ack":true}}
REUSED ID:     {"error":{"code":-32000,"message":"delivery smoke2-1 collides with the entry already at .../webhook-inbox/smoke2-1-3089790caf326238.json: same delivery id, different body — the sender re-used an id"}}
WRONG METHOD:  {"error":{"code":-32601,"message":"unknown method \"webhook.evaluate\"; this listener serves only webhook.deliver"}}
ID MISMATCH:   {"error":{"code":-32602,"message":"request id \"other\" does not match params.delivery_id \"smoke2-3\""}}
--- inbox ---
  smoke2-1-3089790caf326238.json
exit=0
socket unlinked on exit
```

One stored copy across the ack and the redelivery; the reused id refused rather than silently discarded. The probe produced this and nothing else:

```
DEBUG trusty_common::webhook_relay::serve: liveness probe connected and closed without a frame
```

`grep -c WARN` over the whole run: `0`. Before the fix that probe logged `webhook delivery connection failed; not acknowledged`.

## Note for the reviewer

One existing `#[ignore]`d console test, `integration_from_env_delivery_survives_a_console_restart`, started failing once `from_env` attached the supervisor: it launched a real `trusty-review` from PATH, which acked and cleared the spool the test asserts on. It now sets `TRUSTY_WEBHOOK_TARGET_EXTERNAL=1` and asserts the one remaining precondition — an operator's own listener already serving the socket — rather than flaking on it. No assertion was weakened.

Draining each target's inbox into its review/analysis pipeline is not in this PR. What is in scope is that nothing is lost while a delivery waits there.


bobmatnyc:
## Verdict: WARN

Adversarial review at rung 5. Zero CRITICAL, one HIGH. The ack-ordering guarantee — the property this PR turns on — holds, and I verified it by reading every path rather than taking the claim. The HIGH is about what happens *after* the ack, which the PR discloses as out of scope but does not follow through to its consequence for console's health signal.

## What I verified myself

| Claim | How | Result |
|---|---|---|
| Ack is reachable only from `Ok` out of `take_ownership` | `grep -rn "RelayResponse::ack()"` across the workspace | Exactly one production call site, `webhook_relay/serve.rs:227`, inside the `Ok(())` arm. No error arm, retry, timeout, default, or panic-recovery path reaches it |
| `sync_all` failures propagate rather than falling through to an ack | read `inbox.rs:219-248`, `sync_dir` at `:320-327` | Every failure returns `Err`. `InboxError` → `SinkRejection::not_durable` (`serve.rs:127-134`) → `RelayResponse::refuse` → `is_ack() == false`. No log-and-continue arm |
| Directory fsync is on the directory | `inbox.rs:242` calls `sync_dir(&self.root)`, which opens `root` | Correct — it is the directory, not the file again |
| Temp file cleaned up on `EEXIST` | `inbox.rs:223-237` | Yes. The `remove_file` at `:237` is outside the match and runs for both the `Ok` and `AlreadyExists` arms. No per-redelivery temp leak |
| Ack write failing after ownership → console retries into `EEXIST` | `inbox.rs:225` maps `AlreadyExists` → `already_owned: true` → still `Ok` → ack | Handled. Fails closed in the safe direction: no response ⇒ `Unreachable` ⇒ spool retained ⇒ retry acked on `already_owned` |
| Timeout cancels the response but not the blocking write | `serve.rs:278` `spawn_blocking`, wrapped by `timeout` at `:324` | Tokio does not cancel a blocking task. The write completes, the ack is not sent, the sender keeps its copy. Fail-closed |
| Rejections precede every side effect | `serve.rs:184-229`, in order: parse → `jsonrpc` → method → `provenance.verified` → `take_ownership` | Correct. Codes are `-32700 / -32600 / -32601 / -32602 / -32000` and each is `error`-shaped, so `RelayResponse::is_ack()` (`mod.rs:327`) is false and the entry survives |
| Only `is_acked()` deletes a spool entry | `webhook/mod.rs:458`, `relay.rs:76` | Single guarded deletion path |
| Peer-uid check cannot be bypassed | `serve.rs:255` is the first statement of `handle_connection`; `peer.rs` fails closed on unsupported platforms and refuses root | Sound |
| Socket mode is `0600` | `uds/mod.rs:94` `SOCKET_MODE = 0o600`, applied in `bind_hardened`; inbox dir `0700`, files `0600` (`inbox.rs:35,41`) | Matches the smoke output |
| Gates | ran them here | `cargo fmt --check` + `check_line_cap.sh` + `check_changelog_fragment.sh` EXIT=0; `check_sld.sh` 0 errors 0 warnings; `check_test_pointers.sh` 22648 citations 0 dangling; `cargo test -p trusty-common --features webhook-relay -- --include-ignored webhook_relay` → `ok. 32 passed; 0 failed`; `cargo test -p trusty-console -- --include-ignored` → `ok. 225 passed; 0 failed; 0 ignored`; `cargo clippy -p trusty-common --features webhook-relay --all-targets` and `-p trusty-console --all-targets` with `-D warnings` EXIT=0; `cargo check -p trusty-review -p trusty-analyze` EXIT=0 |
| Changelog fragments, SLOC, `unwrap`/`expect` | fragments present for all four crates; `check_line_cap.sh` clean; grep over the seven new production files | Zero `unwrap()` / `expect(` outside `#[cfg(test)]` |

**The `#[ignore]`d test (`integration_from_env_delivery_survives_a_console_restart`).** Diffed line by line against `origin/main`. **No assertion was weakened, removed, or loosened.** The diff is two additions and zero deletions: `TRUSTY_WEBHOOK_TARGET_EXTERNAL=1` plus its restore, and a new precondition assertion that nothing is already serving `review_socket_path()`. Every original assertion survives verbatim, including the closing `report.acked == 1`, `still_pending == 0`, `final_health.status == Ok`. This is the legitimate case — a premise the feature invalidated, restored — not "the test went red so I changed the test." What the PR note leaves out is in finding 6 below.

## What I took on the engineer's word

- **"Against `352fe5d6` the two console E2E cases fail at their first assertion."** Not verifiable as stated, and I checked: `git show 352fe5d6:crates/trusty-common/src/webhook_relay.rs` has no `serve_until`, no `Inbox`, no `DeliverySink`. The tests use the production receiver this PR introduces, so against the pre-fix commit they do not fail — they do not compile. The substance of the claim (nothing bound the socket, so the outcome was `Unreachable`) I did verify from the pre-fix `from_env`. This is inherent to a fix that adds a component that did not exist; it is not a defect, but the PR body should say "no pre-fix counterpart" for all four console cases, not only two.
- The `trusty-review` / `trusty-analyze` suites (1588 / 461) and the binary smoke run against the real `trusty-review`. I ran `cargo check` on both crates but not their full suites.
- That the legacy HTTP routes still drive review and analysis today.

## Findings

| HIGH | crates/trusty-common/src/webhook_relay/inbox.rs | 261 | Nothing reads the inbox, and console's health inverts because of it. `grep` finds zero callers of `Inbox::list` outside tests. After this merges a delivery is acked, console deletes its spool entry (`webhook/mod.rs:458`), `health.pending` goes 1 → 0 and `ServiceHealth` goes `Degraded` → `Ok` — the exact transition the ignored test asserts at `webhook/tests.rs:1240` — while the delivery sits unprocessed in a directory nothing reads and nothing meters. The operator signal for an unprocessed delivery moves from visible to invisible, and the inbox has no drain, no retention, and no depth metric | Surface inbox depth before this reaches an operator: either report the receiver's held count in console's `/health` and `console_metrics`, or hold the spool entry in a distinct `Delivered`-not-`Acked` state until a drain step exists. At minimum, `WebhookListener` should log held-count at bind and the PR body should state that console now reports `Ok` for an unprocessed delivery | Parent |
| MEDIUM | crates/trusty-common/src/webhook_relay/serve.rs | 294 | `serve_until` handles each connection inline, which is what `supervisor/probe.rs:50-55` names as a REQUIREMENT violation for anything put behind `UdsServiceSupervisor`: "accept promptly and off the request path… A service that accepts inline, or sets a small backlog, breaks the assumption and will be evicted under load. Check that before putting a new service behind this supervisor." On macOS a saturated backlog answers ECONNREFUSED, which `SocketVerdict` classifies as `NotServing` — so a busy listener is evicted and killed rather than waited for. The justification given (zero deliveries in 14 days) is a workload argument, not a contract one, and nothing in the new code acknowledges the requirement exists | `tokio::spawn` the per-connection handler and let the shutdown future drop the loop, or state explicitly in `serve_until`'s docs why the 128-deep default backlog plus the 1 s per-connection budget satisfies `probe.rs`'s requirement at this volume, citing it by name | Fix here |
| MEDIUM | crates/trusty-common/src/webhook_relay/serve.rs | 332 | Every liveness probe logs a failed delivery. `lookup_live` (`supervisor/mod.rs:373`) probes on **every** `ensure_running`, and the probe connects and immediately closes. `handle_connection` reads zero bytes, returns `UnexpectedEof` (`serve.rs:261-266`), and `serve_until` logs `warn!("webhook delivery connection failed; not acknowledged")`. So the one log line an operator would grep for to find a real dropped delivery fires on every successful relay, and again on every `wait_for_socket` poll during spawn. Reasoned from code, not observed — the tests install no subscriber, so I could not capture it without writing code | Distinguish the zero-byte case: return a dedicated error (or `Ok`) for `read == 0 && frame.is_empty()` and log it at `debug` with a message that names it as a liveness probe. Keep `warn` for a peer that sent bytes and then failed | Fix here |
| MEDIUM | crates/trusty-common/src/webhook_relay/inbox.rs | 223 | `already_owned` acks without checking the held copy is the same delivery. The dedup key is `sanitise(id)-fnv1a64(raw id)`; on `EEXIST` the incoming delivery is discarded and acked, and console then deletes its only copy. Two ways two distinct deliveries share a key: (a) FNV-1a-64 is not collision resistant, so `entry_path`'s doc claim at `:190` that "two hostile ids cannot alias onto one file" overstates what the hash provides — reachable only by a signature-capable sender, which is why this is MEDIUM and not higher; (b) console's own fallback id `format!("no-delivery-header-{received_at_unix_ms}")` (`webhook/mod.rs:343`) is not unique across two deliveries in the same millisecond, and that path needs no attacker at all | On the `AlreadyExists` arm, read the held file back and compare `delivery_id` (and ideally the `body_b64` digest); on mismatch return `Err` so the sender keeps its copy. Separately, swap FNV for a truncated SHA-256 — `sha2` is already a dependency, pulled in by `webhook-hmac`, which `webhook-relay` implies | Fix here |
| MEDIUM | crates/trusty-analyze/ui/package-lock.json | 1 | A 1874-line npm lockfile is newly tracked next to the existing `pnpm-lock.yaml`, and `ui/dist/assets/index-CV0cyWoU.js` is replaced by a rebuilt `index-Cy9V9pGj.js` (139680 → 139832 bytes) with `dist/index.html` repointed. Both landed in `6e943603`, a commit whose subject is "bind the webhook UDS listeners". Two lockfiles for one package resolve different trees, and the committed bundle is now whatever the author's local npm produced — an unreviewed dependency-resolution change riding in a webhook PR | `git checkout 352fe5d6 -- crates/trusty-analyze/ui/dist` and drop `crates/trusty-analyze/ui/package-lock.json` from the branch. If the UI genuinely needs rebuilding, that is its own PR | Fix here |
| MEDIUM | crates/trusty-common/src/webhook_relay/listener.rs | 132 | The socket is unlinked after the listener is already closed — the #5169 shape. `serve_until` takes `listener` by value and drops it on return, so between that drop and `remove_file` nothing answers the path. A successor that binds in that window (probe fails → `bind_singleton_hardened` unlinks and rebinds) has its fresh socket deleted by the predecessor. Inside `UdsServiceSupervisor` the #5085 reap ordering closes this — the doomed child is waited out before the socket is reprobed — so the window is only open for a listener the supervisor does not own: an operator's Ctrl-C'd `webhook-listen`, or external mode. Consequence is a wasted respawn cycle and an orphan on an unlinked inode, not data loss, which is why this is MEDIUM | Unlink before dropping the listener — take `&UnixListener` in `serve_until` or `remove_file` inside `run` while the listener is still bound. A successor's probe then succeeds and it refuses with `AlreadyServing` instead of clobbering | Fix here |
| LOW | crates/trusty-common/src/uds/singleton.rs | 49 | `UnixStream::connect(path).await.is_ok()` treats every connect error as "corpse". `probe.rs:41-46` documents that on macOS a live listener with a full accept queue answers ECONNREFUSED, and the module doc here reasons only about wrong-mode misreads. Combined with finding 2 (inline accept), a busy live owner can be misread as stale and unlinked out from under itself. Low because a bind happens once per child start, when the predecessor is not serving | Reuse `probe_socket_verdict`'s three-state classification and take over only on `NotServing`; treat `Inconclusive` as `AlreadyServing` | Fix here |
| LOW | crates/trusty-common/src/webhook_relay/serve.rs | 183 | `RelayRequest.id` is never compared to `params.delivery_id` and never echoed in the response — `RelayResponse` has no `id` field. A frame with `id: "A"` and `delivery_id: "B"` is accepted and stored under B, so a receiver log correlated by JSON-RPC id points at a different delivery than the one on disk. Not exploitable (one frame per connection), but it is a frame that passes every check and should not | Reject when `request.id != request.params.delivery_id` with `CODE_INVALID_PARAMS`, before `take_ownership` | Fix here |
| LOW | crates/trusty-console/src/webhook/tests.rs | 1175 | Setting `TRUSTY_WEBHOOK_TARGET_EXTERNAL=1` is the right call, and it leaves console's spawn-a-target path — the thing this PR exists to add — without automated end-to-end coverage anywhere. `spawn_adopts_a_socket_that_is_already_served` covers adoption and explicitly avoids resolving a binary; the spawn itself rests on the manual smoke run in the PR body. "No assertion was weakened" is true but incomplete | State the gap in the PR body. If a covered spawn is wanted, point `SpawnSpec` at a test fixture binary rather than a PATH-resolved `trusty-review` | Parent |

## Required Changes

1. Finding 1 — decide, before this merges, how an operator sees a delivery sitting undrained in an inbox. This is the blocker for #5181 (see Notes).
2. Findings 2, 3, 4, 5, 6, 7, 8 — fix in this PR; each is small and local.
3. Amend the PR body: "no pre-fix counterpart" applies to all four console E2E cases, not two, because the tests do not compile against `352fe5d6`.

## Notes

**#5181 cannot follow this PR directly.** The scope boundary is stated honestly in the PR body and in both `webhook_listener.rs` module docs, and I confirmed it in code: nothing reads the inbox. So today a delivery relayed over UDS reaches `~/…/trusty-{review,analyze}/webhook-inbox/<id>.json` and stops. Reviews and analyses still happen, but only because the legacy HTTP routes are untouched. Retiring those routes (#5181) while the inbox has no drain would make every webhook ack into a directory nothing reads — reviews would stop silently, with console reporting `Ok`. **A drain step has to land between this PR and #5181.** That is the single most important thing for the epic to carry forward.

**Fail-Open Check.** Ran over every failure branch the diff adds. The write path fails closed everywhere: each `InboxError` propagates to a refusal, no branch advances state past a failure, and the sibling arms of `dispatch_frame` are symmetric — every non-ack arm returns an `error`-shaped response and none of them touches the sink. Error-arm tests exist and are real: `dispatch_refuses_when_the_sink_cannot_take_ownership`, `inbox_persist_fails_when_the_root_is_not_a_directory`, `inbox_as_a_sink_refuses_when_it_cannot_write`, and the console-side `a_receiver_that_cannot_own_the_work_never_produces_an_ack`, which asserts the spool entry survives and `attempts == 1` on disk. The one place the check bites is step 2 — "name the alarm, then break it" — and that is findings 1 and 3: the aggregate health signal goes green while the work is undone, and the per-delivery WARN line fires on every liveness probe.

**Not flagged, deliberately.** `Inbox::open` chmods an existing directory without an ownership or symlink check, unlike `prepare_socket_dir`. The inbox lives under the user's own data directory, so exploiting it needs prior compromise of that directory — below the 80% bar for a real finding, worth a line of doc if the module is ever pointed somewhere less private.

Test quality is high. The three load-bearing cases named in `tests.rs:3-6` do pin what they claim, `serve_concurrent_deliveries_of_one_id_produce_one_stored_copy` genuinely races 8 relays through a real socket, and the console E2E cases drive the production receiver rather than a stub that agrees with whichever half was edited last.

bobmatnyc:
## Code Critic — Round 2 Verdict: APPROVE

Reviewed at `64822cfa342fe89f3727cfe1d5790a39cb787c83` (branch `fix/5182-webhook-uds-listeners`, rebased onto `c32dd545`). Independent pass — did not read round 1's findings before reviewing.

### The three flagged targets, verified against the actual code

**1. `classify_read_failure` reclassification (`crates/trusty-common/src/uds/rpc.rs`) — no success-over-failure risk found.**
Traced every consumer of `UdsRpcError` workspace-wide (`grep -rn UdsRpcError`): the only non-test call site is `UdsRelay::deliver` in `crates/trusty-console/src/webhook/relay.rs:199-204`, which matches `Err(e) => RelayOutcome::Unreachable { reason: format!("{e}") }` — every variant, uniformly, with no per-variant branch. No metric, health field, log level, or status anywhere in the workspace discriminates `NoResponse` from `Read` (checked every `reason`/`SpoolHealth` field that touches this path). The claim in the PR description holds.

**2. `TEMP_SEQ` coverage (`crates/trusty-common/src/webhook_relay/inbox.rs:36,394`) — mechanism confirmed load-bearing; Linux-specific reliability unresolved.**
I reverted the `seq` component locally and stress-ran `serve_concurrent_deliveries_of_one_id_produce_one_stored_copy` 30× on macOS: **28/30 runs failed** without it. The counter is real, not decorative. But the module's own doc comment concedes the collision this test relies on is "stronger on macOS (coarser SystemTime) than on Linux," and Linux's finer-grained `clock_gettime` makes the same-nanosecond collision across 8 concurrent tasks considerably less likely to fire — meaning the regression protection may be materially weaker on `ubuntu-latest`, the platform that actually gates the merge. I don't have Linux access in this session and could not settle this directly; flagging as a coverage-quality gap rather than a proven live miss (see Finding 2).

**3. CI feature-gate coverage — the "294 vs 427 tests" note does not describe the actual required check; verified via `cargo metadata` and a real CI log.**
`cargo metadata`'s default (workspace) resolution shows `trusty-common` activates both `webhook-relay` and `uds-supervisor` under a plain workspace build — `trusty-console`'s `Cargo.toml` requests both, `trusty-review`/`trusty-analyze` request `webhook-relay`, and Cargo unifies features across one build invocation. CI's actual "Test" job runs `cargo test --workspace …`, not `-p trusty-common` alone. Confirmed against a real CI log on this branch (run 93124910938): `webhook_relay::receive_tests::*` visibly executed (46+ matches, 1944 tests in that shard). The author's own 294-vs-427 note describes only the narrower `-p trusty-common` local command — not a gate-integrity failure in the required check.

### Findings

| Severity | File | Issue | Fix | Disposition |
|----------|------|-------|-----|--------------|
| MEDIUM | `crates/trusty-review/src/webhook_listener.rs:31-52`, `crates/trusty-analyze/src/webhook_listener.rs:31-52` | `inbox_root()` in both crates hand-rolls `resolve_data_dir(app).join(INBOX_DIR_NAME)` with a locally redeclared `INBOX_DIR_NAME` const, instead of calling `trusty_common::webhook_relay::inbox_root_for(source)` — the shared entry point this very PR introduces for exactly this path (`webhook_relay/mod.rs:117-120`, whose own doc says "a second spelling of the path would make console meter a directory nobody writes to and report `Ok` forever"). `socket_path()` gets an equality test against the shared contract (`socket_path_matches_the_shared_contract`); `inbox_root()` only gets a substring check (`inbox_root_lives_under_the_review_data_dir` asserts `ends_with("webhook-inbox")` / parent `ends_with("trusty-review")`), not equality against `inbox_root_for`. Today the three independent computations agree, so nothing is broken — but a future rename of `webhook_relay::INBOX_DIR_NAME` or `inbox_app_name` would silently desync console's meter from the receiver's real inbox, reintroducing the exact health-inversion bug round 1 fixed (spool empties, `undrained` reads zero, status goes `Ok`), invisibly, since nothing tests parity. | Call `trusty_common::webhook_relay::inbox_root_for(source)` from both `inbox_root()` functions; add a parity test mirroring `socket_path_matches_the_shared_contract`. | Fix here |
| MEDIUM | `crates/trusty-common/src/webhook_relay/inbox.rs` (`write_temp`) + its only regression coverage (`webhook_relay/tests.rs:593`) | The concurrent-8-relay test is the sole guard against the `TEMP_SEQ` regression, and per the module's own comment its trigger rate is platform-dependent (coarser clock ⇒ more likely to collide). Confirmed on macOS the test reliably catches a reverted `TEMP_SEQ` (28/30 fails); could not confirm the same on Linux, the platform CI actually runs on, and general clock-resolution behavior suggests it may fire far less often there. | Add a unit test for `write_temp`'s temp-path uniqueness that does not depend on real clock collisions — e.g. exercise it directly with an injected/frozen timestamp, or assert uniqueness across a tight sequential loop irrespective of `SystemTime` granularity — so the regression is provably caught on every CI platform, not just the one where the OS clock happens to be coarse enough. | Fix here |

### Also checked, no issues found
- Singleton bind (`bind_singleton_hardened`), listener bind/unlink ordering, and `serve_until`'s accept loop — sequencing matches the documented invariants (ack only after durable fsync; unlink before listener drop).
- `probe.rs` move to `crate::uds::probe` is a pure relocation (diffed byte-for-byte against the moved block); `supervisor/tests.rs` — including the pre-existing macOS flake `probe_verdict_reports_not_serving_for_a_stale_socket_file` the author says was left alone — has an empty `git diff origin/main...HEAD`, confirmed.
- No new `#[ignore]` except `from_env_meters_every_targets_inbox`, which mutates `TRUSTY_DATA_DIR_OVERRIDE` and follows the existing `integration_from_env_*` convention already used elsewhere in this crate. No `cfg`-gating or `--exclude` added anywhere in the diff.
- `Inbox::confirm_same_delivery` correctly handles console's non-unique `no-delivery-header-<ms>` fallback id: same id + different body is a `KeyCollision` refusal (sender keeps its copy), not a silent overwrite.
- Cross-crate Cargo.toml feature wiring (`trusty-console`, `trusty-review`, `trusty-analyze` → `trusty-common`) matches what `cargo metadata` resolves.

### Not independently confirmed
The live CI run for this exact HEAD (`31288057354`) was still `in_progress` (Test job unresolved) when this review was written — I could not fold its result in. Given the two prior CI failures I found on this branch (`webhook_relay::receive_tests::serve_rejects_an_oversized_frame`, expecting `NoResponse`, getting `Read{ConnectionReset}`) trace to commit `907055246`, an ancestor that **predates** `classify_read_failure` (`6c99884bd`) entirely — pre-fix code unconditionally maps every read error to `Read`, so that failure was expected and is not evidence of a live problem — I don't consider it a finding. I stress-ran the current version of that test 30/30 locally (macOS) with no failures. Worth a final green check on the actual current-HEAD CI run before merge.

---

bobmatnyc:
Follow-up: CI run `31288057354` for this exact HEAD (`64822cfa3`) finished after I posted — **Test job: success**. That resolves the "not independently confirmed" caveat above: `serve_rejects_an_oversized_frame` and the rest of `webhook_relay::receive_tests` pass on Linux for the current head, so the two prior failures were indeed the pre-`classify_read_failure` ancestor commit, not a live flake.

Doesn't change Finding 2 (`TEMP_SEQ` coverage) — a green run proves the fix works today, not that the test would catch a *regression* on Linux if `TEMP_SEQ` were reverted, which is the question a same-nanosecond-collision-dependent test can't answer without a platform-independent assertion. Verdict stands: **APPROVE**, two MEDIUM findings, neither blocking.

---

Issue #5204, opened 2026-08-08
feat(trusty-mpm): make the worktree base configurable, unitary across all four crates


bobmatnyc:
**Confidence: Reproduced** (direct source read via `git grep -n '"\.worktrees"'`
across the workspace; behavior follows deterministically).

The worktree base directory name is a hardcoded `.worktrees` literal, with no
config knob at all — unlike the managed workspace root, which is already
configurable in `trusty-mpm` (env > config template > default; see companion
issue #5203). Per the owner ruling (2026-08-08): "it should be unitary — if
it's configurable, all systems use it." This is also `CLAUDE.md`, "Common
entry point, clean domain demarcation": a shared capability must have exactly
one implementation every consumer routes through. `.worktrees` is hardcoded
independently in four crates:

- `crates/trusty-mpm/src/daemon/managed_routes/inproject.rs:321` — creation:
  `base_path.join(".worktrees").join(worktree_name)`
- `crates/trusty-mpm/src/session_manager/decommission.rs:54` —
  `pub(crate) const WORKTREES_DIRNAME: &str = ".worktrees"` — a crate-private
  partial consolidation that already exists and is the shape to build on
- `crates/trusty-mpm/src/session_manager/prune.rs:406,411` — scans both
  `<repo>/.worktrees` and `<repo>/.base/.worktrees`
- `crates/trusty-mpm/src/session_manager/search_gc.rs:366`,
  `crates/trusty-mpm/src/core/agent_reset_workspace.rs:364`
- `crates/trusty-search/src/service/constants.rs:35` —
  `EPHEMERAL_DIR_NAMES: &[&str] = &[".worktrees"]`, the auto-discovery
  exclusion
- `crates/trusty-memory/src/attribution.rs:425` — `resolve_workstream_name`:
  `if part == ".worktrees"`, how `creator:workstream=` / `ws:` tags are
  derived from cwd

**Ask**: make the worktree base name configurable, and make every creating
AND detecting site resolve the same configured value — not just `trusty-mpm`'s
own call sites.

**Same architectural gap as #5203 for two of the four crates**: `trusty-search`
and `trusty-memory` have no in-process dependency on `trusty-mpm`'s resolver
(same class of problem as `trusty-code` in #5203). Whatever design #5203
settles on (an exposed daemon route vs. a duplicated env-only precedence
chain) governs these two as well — linking rather than re-deciding here.

**Two consequences that make this more than a rename, and make silent
breakage likely**:
[...]

Issue #5203, opened 2026-08-08
fix(trusty-code): hardcoded ~/trusty-mpm-projects ignores the configured workspace root


bobmatnyc:
**Confidence: Reproduced** (direct source read; behavior follows deterministically from the code).

`trusty-mpm`'s managed workspace root is configurable —
`workspace_root(config)` in `crates/trusty-mpm/src/core/trusty_tools_config.rs:584-616`
resolves `TRUSTY_MPM_WORKSPACE_ROOT` env > `workspace_root_template` in
`~/.trusty-tools/trusty-mpm/config.yaml` > default `~/trusty-mpm-projects`
(`DEFAULT_WORKSPACE_DIR` at `trusty_tools_config.rs:68`). `TRUSTY_MPM_REPOS_ROOT`
delegates into the same resolver (`daemon/managed_routes/inproject.rs:94-95`) —
that delegation was the fix for #1803.

`trusty-code` never reads any of that. It hardcodes the literal instead:
- `crates/trusty-code/src/fs_browse/mpm_registry.rs:254-257` —
  `resolve_local_path`: `home.join("trusty-mpm-projects").join(owner).join(repo)`
- `crates/trusty-code/src/fs_browse/roster.rs:195-196` — `list_projects_in`:
  `home.join("trusty-mpm-projects")`, then the two-level owner/repo scan at
  `roster.rs:221-260`

`grep -rn "TRUSTY_MPM_WORKSPACE_ROOT\|TRUSTY_MPM_REPOS_ROOT\|workspace_root_template" crates/trusty-code/src/`
returns nothing. An operator who retargets the managed root via config or env
gets a tcode GUI project picker that keeps scanning `~/trusty-mpm-projects` and
reports real checkouts under the retargeted root as missing/unregistered.
Silent — no error, no warning.

Per the owner ruling (2026-08-08): "it should be unitary — if it's
configurable, all systems use it." This is also `CLAUDE.md`, "Common entry
point, clean domain demarcation": a capability shared across crates must have
exactly one implementation every consumer routes through; a second independent
implementation is a defect. `trusty-code` is a second, drifted implementation
of workspace-root resolution.

**Why it survived #3435**: that issue moved the roster to registry-primary but
left the local-existence check and the fs-scan fallback hardcoded.

**Design question for the implementer** (not resolved here — `trusty-code` has
no Cargo dependency on `trusty-mpm`; `grep trusty-mpm crates/trusty-code/Cargo.toml`
[...]

PR #5214, opened 2026-08-08
feat(trusty-mpm): read a committed project-level config.toml; consolidate the config surfaces

### LINKED ISSUE #5207: feat(trusty-mpm): read a committed project-level config.toml; consolidate the config surfaces
Project-scoped settings (like `worktree`) live only in machine-global stores, so a project moved or re-cloned loses them. `worktree` resolves from `~/.trusty-mpm/project-registry/projects.json` via `Project::worktree_enabled()` (`crates/trusty-mpm/src/project/record.rs`, `self.worktree.unwrap_or(true)`). `~/.trusty-tools/trusty-mpm/config.yaml`'s `projects:` array only seeds that registry at daemon startup (`crates/trusty-mpm/src/project/registry.rs::seed_from_config`). `tm projects config <name> set worktree ...` writes the JSON registry directly (`crates/trusty-mpm/src/bin/tm/commands/projects/registry.rs` → `PATCH /api/v1/projects/{name}` → `crates/trusty-mpm/src/daemon/managed_routes/project_registry_routes.rs` → `ProjectStore`). No code path today writes a project-level config file from that command.

The machinery to fix this already exists twice, unused. `tm` already writes `<project>/.trusty-mpm/config.toml` at `crates/trusty-mpm/src/bin/tm/commands/project.rs::scaffold_project_dir`, and nothing reads it back — inert legacy from the deprecated `tm project init/list/info` registry, superseded per closed #2123 and kept only for backward-compat output. `<project>/.trusty-mpm/manifest.toml` is a real, read, project-level TOML with working layered precedence — project > user > catalog > compiled-in default — see `crates/trusty-mpm/src/core/manifest/resolve.rs::MANIFEST_FILE`. That is the pattern to reuse; do not invent a fourth surface.

`worktree` is not bootstrap-blocked (verified): in `spawn_managed_routed` (`crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs`) the in-project branch resolves `local_path` first, then consults `worktree_enabled_for_origin` (`crates/trusty-mpm/src/daemon/managed_routes/launch_on_main.rs`) — the project directory exists on disk before the setting is read.

`workspace_root_template` stays machine-level by owner decision — `workspace_root()` (`crates/trusty-mpm/src/core/trusty_tools_config.rs`) runs before any project directory exists, so it cannot be read from the project it would create. Do not fold this into the project-level file.

Owner ruling: the project-level `config.toml` is committed to the repo — tracked in git, travels with clones, reviewable in PR diffs. Not gitignored.

Machine-global surfaces, for the record: `~/.trusty-mpm/config.toml` via `FrameworkPaths::config_toml()` (`crates/trusty-mpm/src/core/paths.rs`; documented as "exactly one global path per machine/user regardless of which project workspace the caller happens to be running inside"); `MpmConfig::load` / `load_default()` (`crates/trusty-mpm/src/core/config.rs`); XDG `~/.config/trusty-mpm/config.toml` (`crates/trusty-mpm/src/bin/tm/commands/managed_root.rs::xdg_config_path`).

In scope, two adjacent incidental defects the owner approved fixing here:
- Add `#[serde(deny_unknown_fields)]` to the config structs — zero of nine config files have it today (`core/config.rs`, `core/trusty_tools_config.rs`, `project/record.rs`, trusty-agents `runtime/indexer.rs` / `repl/statusline.rs` / `mcp/config/mod.rs`, `trusty-memory/src/service/user_config.rs`, `trusty-search/src/service/config.rs`, `trusty-review/src/config/mod.rs`), while the attribute is used 21 times elsewhere in the workspace. A typo'd key silently no-ops.
- Resolve the orphaned `TrustyToolsConfig.default_model` field — only `MpmConfig.models.default` is consumed, by `resolve_agent_model()` (`crates/trusty-mpm/src/core/config.rs`). Either wire it or delete it.

Confidence: Observed — every claim above verified directly against source (`git grep` / file reads) the same day this issue was filed.

Done when: the project-level `config.toml` is actually read (reusing the `manifest.toml` layering pattern), `tm projects config ... set worktree` writes to it instead of only the JSON registry, it is committed/tracked, `deny_unknown_fields` is added to all nine listed structs, and `default_model` is wired or removed.

Siblings under the same owner ruling: #5203, #5204.

bobmatnyc:
## What changed

Every trusty-mpm config surface was per-HOST — `~/.trusty-mpm/config.toml`, `~/.trusty-tools/trusty-mpm/config.yaml`, the machine-global `projects.json`, `$XDG_CONFIG_HOME`. A project's own conventions had to be re-declared by every operator on every machine, and could never be reviewed, versioned, or diffed.

This adds `<project>/.trusty-mpm.toml` — **tracked in git** — as the highest-precedence layer, and makes `worktree` the first setting to use it.

| Setting | Precedence, highest first |
|---|---|
| `worktree` | `.trusty-mpm.toml` → `projects.json` registry → built-in `true` |
| `default_model` | `.trusty-mpm.toml` → `config.yaml` `default_model` → `config.toml` `[models] default` → built-in |

The project layer wins in **both** directions — a repo can force isolation back ON over a local opt-out, not only off. `default_model` is a *default*: explicit `--model`, per-agent overrides, and agent frontmatter are more specific and still win.

The clone-based spawn branch, which asks before any project directory exists, is unchanged and still resolves from the registry alone. `workspace_root` and `auto_resume` stay host-level, per the ruling.

## Decision 1 — why `.trusty-mpm.toml` and not `.trusty-mpm/config.toml`

I did **not** adopt the path `scaffold_project_dir` already writes inertly. Owner ruling 4 requires the file to be committed, and `<project>/.trusty-mpm/` is where trusty-mpm writes machine-local session state (`sessions/`, `logs/`), so projects gitignore it wholesale. This repository is the proof:

```
$ git check-ignore -v .trusty-mpm/config.toml
.gitignore:69:.trusty-mpm/*	.trusty-mpm/config.toml
```

That rule already makes the #4832 `framework/manifest.toml` project layer untrackable here. Putting a file that *must* be committed inside a directory that exists to hold uncommittable state would require every consumer project to carve a `!` re-include out of its own ignore rules — and trusty-mpm cannot reach into a consumer's `.gitignore`. The `.gitignore` comment at that line also records that re-includes under `.trusty-mpm/` were deliberately retired (#4286/#4676) because they caused a `tm doctor` failure.

A root dotfile is trackable by default everywhere and sits beside the other committed-by-convention project files. `scaffold_project_dir` is left alone — it stays as the inert deprecated-registry-A leftover it was.

`core::scaffold_gitignore`'s existing test now also asserts the new file is never scaffolded into a project's ignore block, so ruling 4 is mechanically enforced rather than merely intended.

## Decision 2 — `deny_unknown_fields` is a breaking parse change, applied to one struct

**Applied:** `ProjectLevelConfig` (new). `worktre = false` is now an error, not a silent no-op.

**Not applied — and this is a deliberate reversal of the brief, because applying it would be a regression:**

- `MpmConfig` + children, `TrustyToolsConfig` + children. Both loaders answer *any* parse error by returning `Default`. Adding `deny_unknown_fields` there upgrades "one key is ignored" into "**the entire file is ignored**" — every model override, every section, gone, behind one `warn` line. The failure it introduces is strictly worse than the one it fixes.
- `Project` (`projects.json`). Machine-written by `ProjectStore`, not hand-edited. Denying unknown fields on a persisted record breaks version skew in both directions: an older `tm` would refuse to read a registry a newer `tm` wrote — and that registry is what the worktree opt-out reads.

**What those structs get instead**, so the owner's actual complaint is still fixed: `core::config_keys` reports every unrecognised key at `warn`, naming the file and the dotted path (`models.defualt`). It is derived from a serde round-trip rather than a declared key list, so it needs no maintenance and cannot drift as sections are added, and it descends into nested tables. The typo stops being silent; the other keys keep working. No new dependency.

Strictness on the new struct is safe *because* there is no legacy corpus and the file is reviewed in a PR before it reaches anyone else. At spawn time a rejected file contributes nothing and is logged at `error`, and resolution falls through to the registry — a committed file is shared, so one bad push must not brick every operator's launches.

## Decision 3 — `TrustyToolsConfig.default_model`: wired, not deleted

Wired. Deleting it would break three live operator-facing surfaces: the trusty-console Config tab (`ConfigTab.svelte`), the `config_write` MCP tool, and the documented capability in `tm-capabilities`. The field's own UI placeholder reads *"(unset — uses ~/.trusty-mpm/config.toml)"* — an explicit promise that setting it changes the launch model, which it did not keep.

`MpmConfig::load_effective` folds it into `[models] default` at load time rather than adding a parameter to `resolve_agent_model`. That is what makes it reach every launch path — PM launch, agent delegation, daemon — without touching their signatures, and it is what owner ruling 1 means by unitary. The YAML sits above the TOML because that is the contract the placeholder already advertises.

## Test ladder — Rung 4

```
bash scripts/check_line_cap.sh                        — 3806 files, 4 allowlisted, 0 violations
cargo check -p trusty-mpm --all-targets               — clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings — clean
cargo test -p trusty-mpm                              — 4814 passed, 1 failed (pre-existing, see below)
cargo check --workspace --exclude trusty-code-gui --exclude trusty-mpm-gui — clean
```

`lifecycle.rs` is frozen on `.line-cap-allowlist.tsv` at 818 SLOC. Rather than grow it, the daemon-side adapter went into `launch_on_main.rs` — which exists for exactly that purpose per its own module doc — leaving `lifecycle.rs` a net line shorter.

### Red proof

Neutering the three behavioural changes (drop `deny_unknown_fields`; make the project layer always return `None`; make the default-model fold a no-op) turns **8 tests red**:

```
core::project_config::tests::project_config_rejects_unknown_key ... FAILED
core::project_config::tests::project_config_rejects_unknown_key_alongside_valid_ones ... FAILED
core::project_config::tests::project_config_load_surfaces_a_bad_file ... FAILED
core::project_config::tests::load_or_report_returns_none_for_unknown_key ... FAILED
project::worktree_policy::tests::worktree_project_config_overrides_registry ... FAILED
project::worktree_policy::tests::worktree_project_config_can_force_isolation_on ... FAILED
core::config::tests::project_default_model_tops_the_chain ... FAILED
core::config::tests::yaml_default_model_beats_toml_default ... FAILED

test result: FAILED. 45 passed; 8 failed
```

Restored:

```
test result: ok. 62 passed; 0 failed; 0 ignored; 0 measured; 4760 filtered out
```

Of the five required regression tests, two are **no-regression guards** and pass both before and after by design — `worktree_project_config_absent_falls_back_to_registry` and `worktree_defaults_true_with_neither_layer`, along with `worktree_clone_branch_resolves_without_a_project_dir`. Their job is to prove the new layer did not disturb the existing chain, so a red for them would mean I had broken something.

### Pre-existing failures, not caused by this branch

`core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning` fails under the full parallel suite. It fails identically on `origin/main`:

```
# origin/main @ 5e6054f6a, cargo test -p trusty-mpm
test core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning ... FAILED
test result: FAILED. 4783 passed; 1 failed; 7 ignored
```

The failing set moves between runs (`stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet`, `create_managed_session_confirms_server_before_applying_options`), and all three pass in isolation. This is the canonical open issue **#4931** — "two order-dependent tests fail under parallel execution, pass in isolation". None are in files this branch touches.

`cargo check --workspace` fails on `trusty-code-gui` and `trusty-mpm-gui`: `The frontendDist configuration is set to "ui/dist" but this path doesn't exist`. Both are Tauri crates needing a pnpm frontend build that a fresh worktree does not have; this branch touches neither crate. Everything else compiles clean.

## Scope

trusty-mpm only. `workspace_root` behaviour unchanged, `auto_resume` not rescoped. Siblings #5203 (trusty-code hardcodes `~/trusty-mpm-projects`) and #5204 (`.worktrees` literal across four crates) stay separate and are not touched.

Closes #5207


bobmatnyc:
Both CI failures were real and branch-caused. Neither was the `#4931` flake.

## 1. `Test` — the #4203 guard, not a flake

`tests_behavior_b::launch_paths_prepare_through_the_isolated_seam` failed at `tests_behavior_b_tests.rs:1712`:

```
commands/launch.rs spawns `claude` with `--setting-sources project,local`, which
excludes the `user` tier `FrameworkPaths::default()` deploys into — deploy via
`session_launch::prepare_isolated_session` instead (issue #4203)
```

The guard asserts `commands/launch.rs` never names the framework paths' default constructor, because a self-resolved framework root there is how the deploy-tier bug got reintroduced twice. To read the effective config, this PR called it directly to obtain a root — tripping the guard.

Fixed by moving the root resolution inside `config.rs`. `MpmConfig::load_effective_default` delegates to `load_default` (same home handling and stripped-CI fallback), and a shared `with_outer_default_model_layers` keeps it from drifting in precedence from the root-taking `load_effective`. Behavior is unchanged — `FrameworkPaths::default().root` and `load_default` both resolve `~/.trusty-mpm`.

The guard matches raw source text, so the first fix still failed: the replacement *comment* spelled the symbol out. The comment is now worded around that and says why.

The known flake `ensure_managed_config_dir_emits_the_frozen_skill_warning` did not fail in CI or locally on this round.

## 2. `Per-PR changelog fragment` — one fragment, two categories

`5207-project-level-config.md` had `Added` on line 1 and a second bare `Fixed` on line 10. The assembler rejects that, since everything after line 1 is copied verbatim and the second heading would render as body text under `### Added`.

No local/CI divergence: the same rejection landed in `f35029a57`, before this branch's merge base. I re-ran `scripts/check_changelog_fragment.sh` locally against the exact rejected file and it reproduced CI byte-for-byte, so the earlier "passed locally" claim was stale, not a gate mismatch.

Split into `5207-project-level-config.md` (Added) and `5207-project-level-config-fixed.md` (Fixed).

## Coverage added

`load_effective` had no direct test — the existing layering test operates on an in-memory struct and would have stayed green had the refactor dropped the project layer entirely. `load_effective_applies_the_project_layer` now reads a real `config.toml` and a real `.trusty-mpm.toml` from disk. Nothing was ignored, cfg-gated, or excluded to make a gate green.

## Gates (rung 3, `-p trusty-mpm`), after merging `origin/main` @ `c217a1ad`

```
cargo fmt --check                                   EXIT=0
cargo check -p trusty-mpm                           EXIT=0
cargo clippy -p trusty-mpm --all-targets -D warnings EXIT=0
cargo test -p trusty-mpm                            EXIT=0
  test result: ok. 4816 passed; 0 failed; 7 ignored
  test result: ok. 1546 passed; 0 failed; 1 ignored
  test core::config::tests::load_effective_applies_the_project_layer ... ok
  test tests_behavior_b::launch_paths_prepare_through_the_isolated_seam ... ok
scripts/check_line_cap.sh    3822 files, 0 violations — OK
scripts/check_changelog_fragment.sh   OK trusty-mpm
scripts/check_sld.sh         0 errors, 0 warnings
```

Head: `39dd745838195077d7a4c3d51440eeb6723bc8ae`

PR #5221, opened 2026-08-08
refactor(trusty-mpm): consolidate workflow, ticketing, and version-control policy (#5202)

bobmatnyc:
## Ownership model

| Surface | Owns |
|---|---|
| Core instructions (`core.md`) | Detect the work shape, load the two skills, route. No workflow policy of its own |
| `tm-workflow` skill | The whole delivery chain: phases/gates, worktree/branch, changelog, PR body, review, merge, cleanup, and the handoff. Performs no operation |
| `tm-ticketing` skill | Whether an issue exists; dedup disposition, title/body, labels, milestone, lifecycle, issue attribution |
| `ticketing` agent | Issue API mechanics. No git, no PR mutation |
| `version-control` agent | Git plus the whole PR lifecycle, including title and body |
| Output styles / `BASE-AGENT` / `SM_WORKFLOW` / `tm-bug-reporting` | Voice and specialization only — pointers, not policy |

## Contradictions found, and which side won

1. **PR body ownership.** Core, `tm-ticketing`, and the `ticketing` agent called PR-body authorship "ticketing bookkeeping"; P7, `tm-pr-workflow`, and the `version-control` agent assigned the PR to version control. **Version control wins.** Routing is now by artifact, so one `gh pr edit` has one owner. `bookkeeping vs. mechanics` is gone as a concept.
2. **Line-number citations.** All three output styles said "link the file and line"; the `ticketing` agent forbids line numbers because a 500-SLOC cap splits modules constantly. **The agent wins.** The styles now govern voice and defer the schema.
3. **Ticketing model.** `tm-delegation-patterns` said `haiku`; the embedded agent frontmatter and the generated roster say `sonnet`. **`sonnet` wins.**
4. **Labels that do not exist.** `tm-bug-reporting` required `agent-error`, `skill-error`, and `high-priority`. `gh label list` has none of them. Removed; it now defers to `tm-ticketing`'s four label families.
5. **Unconditional reopening.** The agent said "never file a fresh ticket for a recurrence". A verified fix followed by a different failure mode is new work. Replaced with four dispositions and a criterion each.
6. **My own regression, caught by an existing test.** My first P6 draft named `mcp__mcp-ticketer__*` literally, which broke `assemble_system_prompt_contains_all_sections`' standing rule that the resident prompt never names a ticketing MCP family. Reworded rather than relaxing the assertion.

## Policy DELETED, and where it now lives

| Deleted from | What | Single home now |
|---|---|---|
[...]

Issue #4605, opened 2026-08-01
Untracked deployed skills are classified project-custom and permanently excluded from redeploy — no tm command can fix them


bobmatnyc:
**Defect**: a bundled skill deployed to a location where it is ABSENT from that location's `.trusty-mpm-skills-manifest.json` is classified by the skill-tier planner as PROJECT-CUSTOM — the highest-precedence tier — and excluded from `plan.bundled_deploy` entirely, before `deploy_skills_filtered` ever runs. It is therefore permanently stale, and no documented `tm` command reaches it. This is not the checksum-freeze protecting a hand edit; the file was never hand-edited, it simply lost (or predates) manifest tracking.

**Evidence**, measured 2026-08-01 against installed `trusty-mpm 1.3.1` (main @ `11ef5c27`), after #4583 merged as `2b5d4c9b`. Ran `tm install` and `tm sessions sync-assets <id>`:

| Path | Outcome | Evidence |
|---|---|---|
| `<worktree>/.claude/skills/tm-workflow/SKILL.md` | FIXED | tracked in its manifest with a matching checksum, deployer refreshed it. mtime 07-29 → 08-01 10:38; `PM_INSTRUCTIONS.md` count 1 → 0 |
| `~/.claude/skills/tm-workflow/SKILL.md` | UNREACHABLE | absent from that manifest (only 1 managed entry: `tm-slack-canvas-delivery`). Byte-for-byte unchanged, mtime still 07-17, still contains `PM_INSTRUCTIONS.md` |
| `$CLAUDE_CONFIG_DIR/skills/tm-workflow/SKILL.md` | UNREACHABLE | same. mtime still 07-03 |

`tm install`'s deploy step printed only `= tm-slack-canvas-delivery (unchanged)`. `tm-workflow` never appeared as deployed, skipped, OR unchanged — excluded upstream by the tier planner, not skipped by the ownership check. That silence is part of the defect.

[...]

Issue #1905, opened 2026-07-02
tm: /tm-* skills not appearing — mpm-* → tm-* rename never actually done


bobmatnyc:
Problem: The `/mpm-*` skill family was supposed to be renamed to `/tm-*`, but only one new skill (`tm-doctor`) was added; the rest of the `mpm-*` skills were never renamed, and even `tm-doctor` is deployed in a non-discoverable shape.

Evidence:
- Bundled skill source files still use the old `mpm-` prefix: `crates/trusty-mpm/src/assets/skills/mpm-delegation-patterns.md`, `mpm-circuit-breaker-enforcement.md`, `mpm-bug-reporting.md`, `mpm-git-file-tracking.md`, `mpm-session-pause.md`, `mpm-session-management.md`, `mpm-session-resume.md`, `mpm-pr-workflow.md`, `mpm-verification-protocols.md`, `mpm-tool-usage-guide.md`, `mpm-ticketing-integration.md`. Frontmatter also still declares old names (e.g. `crates/trusty-mpm/src/assets/skills/mpm-session-management.md:2` → `name: mpm-session-management`).
- Compiled in via `include_str!` in `crates/trusty-mpm/src/core/bundle_skills.rs` (lines 20, 29, 37, 45, 55, 65, 74, 82, 90, 98, 106), re-exported in `crates/trusty-mpm/src/core/bundle.rs:168-169`, and referenced in `crates/trusty-mpm/src/core/bundle_all.rs` (e.g. lines 285, 320) and test assertions in `crates/trusty-mpm/src/core/bundle_tests.rs:443,453`.
- Deployer (`crates/trusty-mpm/src/core/skill_deployer.rs`) has NO hardcoded prefix filter — `deploy_skills_filtered` copies whatever `.md` stems exist in the source dir (`is_skill_file` line 46, `skill_stem` line 57) into `~/.claude/skills/<name>/SKILL.md`, gated only by an ownership/checksum manifest. It would deploy `tm-*.md` files fine — they just don't exist except `tm-doctor.md`.
- `tm-doctor` itself is deployed as a bare flat file `~/.claude/skills/tm-doctor.md`, NOT a `tm-doctor/SKILL.md` directory — per the deployer's own doc comment (`skill_deployer.rs:3-4`), Claude Code discovers skills from `~/.claude/skills/<name>/SKILL.md`, so this flat file is not discoverable at all.
[...]

PR #5229, opened 2026-08-08
fix(trusty-mpm): remove deployed skills the binary no longer ships (#5224)

### LINKED ISSUE #5224: fix(trusty-mpm): deployed skill directories are never removed when the binary stops shipping the skill
`trusty_agents_common::skills::deployer::deploy_skills_filtered` never deletes a previously-deployed skill directory once the binary stops shipping that skill — its own doc comment says so: "Deselecting a skill does not remove a previously deployed copy." Once a skill is retired, the deployed copy becomes an orphan whose deploy-manifest entry has no matching bundled asset.

Consequence, traced on `origin/main`: `trusty-mpm/src/core/skill_drift.rs::classify` returns `SkillDrift::Unverifiable` for exactly this case (`reference.assets.get(key)` is `None`). `trusty-mpm/src/daemon/doctor_skill_drift.rs::report` folds the whole `skill_staleness` check to `CheckStatus::Unknown` whenever any deploy tier has an unverifiable entry and nothing conventions-bearing drifted. A check that reports `Unknown` instead of `Ok`/`Warn`/`Fail` has stopped protecting anything, silently, until a human deletes the directory and its ledger entry by hand.

Live trigger: PR #5221 (closes #5202) deletes `crates/trusty-mpm/src/assets/skills/tm-pr-workflow.md` from the bundled asset set. Every machine with a previously-deployed `tm-pr-workflow/` will orphan it on the next `tm install` from that binary, folding `skill_staleness` there to `Unknown`. PR #5221's own body documents this exact mechanism and explicitly defers the fix: "wiring an equivalent for `tm-pr-workflow` is a separate outcome, not something to half-land here."

Precedent to extend: `trusty-mpm/src/core/stale_skills.rs` already solved this shape once, narrowly. `find_stale_mpm_skills` / `remove_stale_mpm_skills` / `run_stale_mpm_skills_migration_once` is a one-time, allowlist-gated (`FORMER_TRUSTY_MPM_SKILLS`) sweep for the #1905 `mpm-*`→`tm-*` rename — its own doc comment says it is "never wired into `crate::core::doctor`" because it's scoped to that one rename, not general retired-skill cleanup.

Confidence: Reproduced — verified directly against `origin/main` source (`skill_drift.rs`, `doctor_skill_drift.rs`, `stale_skills.rs`, `deployer.rs`), cross-checked with PR #5221's own body, which independently describes the same mechanism.

Closure conditions:
- [ ] A retired skill's orphaned deployed directory and ledger entry get removed (automatically, or via an explicit `tm doctor --fix-skills`-style command), generalizing `stale_skills.rs`'s pattern beyond its frozen `mpm-*` allowlist
- [ ] `skill_staleness` no longer folds to `Unknown` solely because a skill was intentionally retired from the bundle
- [ ] Verified against the `tm-pr-workflow` retirement in #5221 as the concrete case

Relates to #5202; triggered by PR #5221.

bobmatnyc:
Retiring a bundled skill left its deployed copy behind forever. Claude Code kept loading text the binary deliberately removed, and the orphaned ledger key made `tm doctor`'s `skill_staleness` check report `Unknown` — which stops it protecting anything. See #5224 for the traced mechanism.

## Derivation, not an allowlist

`core::stale_skills` solved this shape once behind a frozen `mpm-*` list, scoped to the #1905 rename. This generalises it: a skill is RETIRED at a deploy target when its stem is tracked by that target's ownership ledger and appears in NO live source.

I chose derivation over a second frozen allowlist because an allowlist needs a manual edit per retirement — exactly the step a release forgets, and the failure is silent. The stated risk of derivation is deleting a legitimately hand-placed or user-tier skill, and that is answered by building the live set from every source that could feed the target rather than from the one a given deploy is reading:

- the compiled-in `bundle::ALL` (cannot lag the binary running it)
- the resolved bundled source — the `agents/skills` submodule when checked out, else the extraction cache
- the user-custom tier, `~/.trusty-mpm/skills/`
- the synced catalog's skills
- the project-custom stems already sitting in the target

Over-including a source is the safe direction: the failure mode is leaving an orphan behind, never deleting something live. An unreadable source aborts the sweep entirely rather than narrowing the set.

## Tier safety

Never deleting a user-tier or project-tier skill is held by three independent gates, any one of which suffices:

1. **Candidates come only from the ledger.** A skill hand-placed in `.claude/skills/` has no ledger entry, so it is never a candidate. `retire_spares_a_project_tier_skill` proves this against a deliberately hostile empty live set.
2. **The live set unions the user tier and the project-custom stems.** This is the tier oracle `update_check::apply::prune_skills` already uses for #391 — the live user source, not a recorded origin field, because the ledger records no tier. `retire_spares_a_user_tier_skill` proves it.
3. **The removal verdict refuses anything unproven.** A directory holding a file the ledger does not claim, or a claimed file whose bytes changed after deployment, is never deleted.

**Deselection is not retirement.** The live set is built from what sources CONTAIN, never from what a harness manifest selects, so `deploy_skills_filtered`'s documented HR-3 behaviour is unchanged and deselection remains `tm catalog apply --prune`'s business.

## The hand-edited case — deliberate

A retired skill the operator edited keeps every file exactly where it is. Only the ledger claim is released, and the operator is told at `warn` that the files are now theirs. Same for a directory holding an untracked file.

Releasing the ledger entry in both cases is the point: a ledger entry claims tm deployed the file and can refresh it, and once nothing ships the skill that claim is false. The stale claim is precisely what pins `skill_staleness` to `Unknown`, so keeping it would fix the deletion but not the check. If the skill ever returns to the bundle, #4605's unmanaged-skill detector already finds and adopts a file in exactly this state.

## Where removal happens

`tm install` and session launch — not doctor. The drift check is documented READ-ONLY and its `--fix-skills` repair re-deploys drifted content rather than deleting; adding a third surface would give three places that can delete files. Retirement is a consequence of the binary changing, so the install that changes the binary is the natural moment, and session launch covers machines that never run `tm install` explicitly. Both sweep all three deploy tiers, because the orphan occurs at each independently.

## Consolidation

`prune_guard::skill_verdict` now delegates to the shared `skill_removal_verdict` instead of carrying its own copy. Two independent deletion gates is how they come to disagree about what is safe.

`session_launch/mod.rs` sat at exactly 500 SLOC on main, so the sweep could not be added without a split. Its skill block moved to `session_launch/skills.rs` — the natural seam, since everything the launch does about skills is one sequence.

## Gates — Rung 5

Destructive path, cross-crate blast radius.

- `cargo fmt --check` — clean
- `cargo check -p trusty-mpm -p trusty-agents-common --all-targets` — clean
- `cargo clippy -p trusty-mpm -p trusty-agents-common --all-targets -- -D warnings` — zero warnings
- `cargo test -p trusty-agents-common` — 487 passed, 0 failed
- `cargo test -p trusty-mpm` — 4802 passed, 0 failed, 7 ignored (lib), all integration suites green
- `bash scripts/check_line_cap.sh` — 3820 files, 0 violations
- `bash scripts/check_sld.sh` — 0 errors
- `bash scripts/check_changelog_fragment.sh` — fragment present and valid

`cargo check --workspace` fails in the Tauri GUI crates (`trusty-mpm-gui`, `trusty-code-gui`) with `frontendDist "ui/dist" doesn't exist` — a fresh-worktree condition where the workspace run does not order the UI build ahead of proc-macro expansion. Each crate checks clean on its own (`cargo check -p trusty-mpm-gui` and `-p trusty-code-gui` both exit 0), and `git diff --name-only origin/main...HEAD -- crates/trusty-mpm-gui/ crates/trusty-code-gui/` is empty. Change-specific gates pass; this one is environmental.

## Red proof

Every behavioural test was proven to fail before the fix, by stubbing the mechanism under test and re-running:

- **Sweep disabled** — `retire_removes_a_pristine_orphan`, `retire_spares_a_user_tier_skill`, `retire_spares_a_project_tier_skill`, `retire_keeps_a_hand_edited_orphan_but_releases_the_ledger`, `retire_keeps_an_orphan_holding_an_untracked_file`, `retire_orphaned_skills_sweeps_every_tier` all FAILED (6 failed; 10 passed).
- **Tier gates disabled** (candidates taken from disk, live set ignored) — `retire_spares_a_user_tier_skill` failed with `RetiredSkill { stem: "my-own-skill", removed: true }`, i.e. a real deletion of a user-tier skill; `retire_spares_a_project_tier_skill`, `retire_does_not_touch_a_deselected_but_still_shipped_skill` and `retire_is_a_noop_when_nothing_is_orphaned` also failed. This proves the tier-safety assertions bite rather than passing vacuously.
- **Removal and ledger release disabled** — `a_retired_skill_no_longer_pins_staleness_to_unknown` failed with the defect verbatim: `the check is still degraded after the orphan was retired: … 1 could NOT be verified (operator home/tm-pr-workflow (\`tm-pr-workflow\` is not among this binary's embedded bundled assets)) … left: Unknown, right: Ok`.

## Unblocks #5221

That PR deletes `tm-pr-workflow` from the bundle, which is the live trigger. `skill_source::materialize_skill_artifacts` already prunes the extraction cache of files `bundle::ALL` no longer lists, so once #5221 lands the stem is absent from every live source and the first `tm install` or session launch on each machine removes the orphan. `a_retired_skill_no_longer_pins_staleness_to_unknown` drives that exact case by name.

The two PRs do not overlap: this one touches no file in #5221's diff.

Closes #5224


bobmatnyc:
## Pointer lint fix — pushed `2540b306f`

### What the lint actually flagged

```
FAIL: crates/trusty-mpm/src/core/skill_retire.rs:87: Test: pointer cites `retire_keeps_a_hand_edited_orphan` — no `fn retire_keeps_a_hand_edited_orphan(` or `mod retire_keeps_a_hand_edited_orphan` found in crate `crates/trusty-mpm`.
FAIL: crates/trusty-mpm/src/core/skill_retire.rs:289: Test: pointer cites `retire_keeps_a_hand_edited_orphan` — no `fn retire_keeps_a_hand_edited_orphan(` or `mod retire_keeps_a_hand_edited_orphan` found in crate `crates/trusty-mpm`.
test-pointers: 2 dangling pointer(s), 0 stale allowlist entries — FAILED.
```

Two pointers, one cited name, both in `skill_retire.rs` — the `RetiredSkill` struct doc and the `retire_orphans_in` doc.

### Root cause

The test is named `retire_keeps_a_hand_edited_orphan_but_releases_the_ledger`. Both citations stopped at `..._orphan` and dropped the `_but_releases_the_ledger` half. The name they wrote was never a test — it reads like one because it is a prefix of the real one, which is exactly why it survived review.

That truncation also lost the more interesting half of the claim. "Keeps a hand-edited orphan" is only half the contract; the other half is that the ledger claim is released anyway, which is the part that stops `skill_staleness` folding to `Unknown`. Both pointers now name the full test.

Not main drift: merging `c217a1ad` (task_runner, credentials/redact, inference/config/ops, vmtest-harness) was clean and touched no `trusty-mpm` file.

### Also in this push

`website/tests/build-smoke.test.ts` — `a0a0b2d` deleted its "renders each flagship page from its own tool record, not a shell" assertion. That has nothing to do with #5224; it was a stray working-tree change swept into the commit. Restored from `origin/main`.

### Gates

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo check -p trusty-mpm -p trusty-agents-common` | pass |
| `cargo clippy … --all-targets -- -D warnings` | pass |
| `cargo test -p trusty-agents-common` | 487 passed, 0 failed |
| `cargo test -p trusty-mpm` | 4801 passed, 1 failed — pre-existing flake, below |
| `check_line_cap.sh` | 3821 files, 0 violations |
| `check_sld.sh` | 0 errors, 0 warnings |
| `check_changelog_fragment.sh` | fragment present |
| `check_test_pointers.sh` | `resolved 22768 Test: citation(s) (floor 200) — 0 dangling pointers — OK.` |

The `-p trusty-mpm` red is the known parallel-execution flake set, not a regression. Two full runs failed a *different* pair each time and all pass in isolation:

- run 1: `stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet` (canonical #5040)
- run 2: `ensure_managed_config_dir_emits_the_frozen_skill_warning` (canonical #4931), `create_managed_session_confirms_server_before_applying_options`
- all three together, filtered: `test result: ok. 3 passed; 0 failed`

The tmux one has no issue of its own but is the same `$HOME`-mutation class #4931 describes: it computes its expected history-limit from `TrustyToolsConfig::load()`, which resolves `~/.trusty-mpm`, so any concurrent test swapping `$HOME` behind a `fake_home()` guard makes the expectation and the call disagree. My diff touches none of `managed_config`, `tmux`, or `managed_routes` — `git diff --name-only origin/main...HEAD` over those paths is empty.

### Still owed by whoever lands #5221

Drop `"tm-pr-workflow"` from `CONVENTIONS_BEARING_SKILLS` in `doctor_skill_drift.rs`. Inert once the ledger stops tracking the skill, but dead configuration.

Issue #5200, opened 2026-08-08
no workflow runs the website/ test suite — vitest coverage gap, not a skip


bobmatnyc:
No CI workflow runs the `website/` test suite. Not skipped — absent.

## Evidence (verifying PR #5189, head `9495dd0b7e1a019ec2cc35a75c541993d664b51b`)

- Five workflows ran and reported success (`Changelog fragment`, `Generation artifact lint`, `Line cap`, `CI`, `Capabilities drift`); none is scoped to `website/`.
- `scripts/detect-docs-only.sh` buckets `website/**` as inert, same as `docs/**`. Its own comment claims `website/` "has its own build/lint runs in its own workflow, never Cargo's." That workflow does not exist.
- `CI`'s `Test` job (including `cargo test --workspace`) reports `"conclusion":"skipped"` under `docs_only=true` — correct behavior for a website-only diff, not the bug.
- The only Node/pnpm surface in `.github/workflows/ci.yml` is the `ui-checks` matrix; its inline comment says it covers the seven `crates/*/ui` directories and that no other workflow runs pnpm/vitest/playwright/eslint/tsc. `website/` isn't in that matrix.
- `website/package.json`'s `test` script is `vitest run`. No workflow invokes it; `build` is plain `vite build` with no `prebuild`/`postinstall` test chain, and there's no `website/vercel.json`.
- The one green check touching website output, `Vercel — Deployment has completed`, is a build/deploy, not a test run.

## Impact

`website/`'s suite — 167 tests as of PR #5190, including the `'loads no subresource from a third-party origin'` gate — has never run in CI. Any website PR merges on a green that tested nothing. Same vacuous-green class as #4618 and #4097.

PRs #5189 and #5190 were merged on local `pnpm test` evidence because no CI signal existed — a stopgap, not a fix.

## Resolution direction

Either add `website/` to the `ui-checks` matrix in `.github/workflows/ci.yml`, or give it its own workflow — design left to whoever picks this up. Whichever is chosen, the false claim in `scripts/detect-docs-only.sh`'s comment (`detect_docs_only`) must be corrected — that's what made the gap invisible.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

PR #5248, opened 2026-08-09
feat(website): What's New, generated from the crates' CHANGELOG.md files

bobmatnyc:
The six flagship crates' `CHANGELOG.md` files are the source of truth. The site
is a read-only derived view, built at prerender time. Nothing writes back to a
changelog, and the published pages read no file at runtime.

## What lands

- `website/src/lib/changelog/` — reads the six files and fails the build on a
  missing file, zero releases, a `## [` heading that never closes, a newest
  release with no items, or a link it cannot resolve. No placeholder and no
  empty array anywhere, so a green build is itself the evidence that all six
  sections are populated.
- A "What's new" strip below the pitch bullets on each flagship card: newest
  version and date, up to three items (fewer if the release has fewer, never
  padded), linking to `/whats-new#<crate-name>`.
- `/whats-new` — recent releases in full per crate, an index of every earlier
  one, a crate jump-list, and the living changelog linked in each header.

Reuses `$lib/docs`: `findRepoRoot`, the remark/rehype pipeline, the failure
record and its `FAIL CODE file:line: problem — remedy` rendering. No new
dependency; two shared helpers gained an export rather than a second copy.

Deliberately unpinned links: the doc reader pins in-repo links to a commit SHA
so cited lines cannot drift. These point at living documents and use
`blob/main`. The reasoning sits at both link-building sites.

## Three things the corpus forced

Full detail in `website/src/lib/changelog/README.md`.

- **The link-reference trap.** trusty-search's 58 definitions point at the
  pre-monorepo repo and their labels match its version headings, so remark
  turned every heading into a link to a dead repository. Definitions are
  blanked (not deleted, so line numbers hold) before parsing.
- **The grammar is not Keep a Changelog.** 49 release headings carry a title
  where the date should be, three carry no semver, ~40 category headings carry
  a qualifier. Bucketing is by leading word; the date is optional.
- **`<path>` outside backticks renders as nothing** — eight of these are live in
  the corpus. The doc reader fails the build on this hazard; here it recovers
  the literal text, because a changelog is append-only history this site must
  not edit and a gate would be unfixable.

## Gates

Website tests do not run in CI (#5200); all five were run locally from inside
`website/` and pass. `pnpm check` 481 files 0 errors; `pnpm lint` clean;
`pnpm test` 215 passed (13 files, including the smoke suite's real `vite
build`); `pnpm build` clean. Each of the six required tests was verified to
FAIL under a deliberate mutation of the code it covers.

No changelog fragment: this PR touches only `website/`, nothing under `crates/`.


bobmatnyc:
## Verdict: BLOCK

Reviewed at head `d19be5b78` from a detached worktree at `.claude/worktrees/critic-5248` (fetched from `origin/main`, merge-base `c32dd545d`). Ran `pnpm install`, `pnpm check`, `pnpm lint`, `pnpm test`, `pnpm build` locally inside `website/` — all green (481 files / 0 errors; lint clean; 215/215 tests; production build succeeds). Re-derived all seven claimed mutations by hand-editing the source and re-running the targeted test file; all seven are genuinely caught (see Verification below). Also ran one live `pnpm build` against a deliberately corrupted real changelog to confirm the build gate fires through the actual SvelteKit prerender pipeline, not just in a unit test.

One of the four required fail-closed gates has a real gap, constructed with a concrete failing input below.

## Findings

|----------|------|--------|-------|-----|--------------|
| CRITICAL | `website/src/lib/changelog/parse.ts` | `harden()`, the `target.startsWith('..')` recovery block (~L386–393) | An out-of-repo relative link with the wrong number of `../` is silently re-read as repo-root-relative and shipped as a real link **whenever anything happens to exist at the stripped path** — regardless of whether that's the author's actual target. This is not scoped to "one `../` too many" as the comment and PR body claim: `target.replace(/^(?:\.\.\/)+/, '')` strips *any* number of leading `../`, so a typo with 2, 3, or 5 extra levels collapses to the exact same "repo-root-relative" guess. I constructed and ran a concrete failing case: a changelog item in `crates/trusty-search/CHANGELOG.md` (dir depth 2) with `[the README](../../../README.md)` — one extra `../`, same shape as the one real occurrence this code was built for — parses with **zero failures** and emits `href="https://github.com/bobmatnyc/trusty-tools/blob/main/README.md"`, i.e. the repo-root README, not the crate's own. `parseChangelog` returns `failures: []` and a real, wrong, plausible-looking `<a>` tag. The repo has 47 `README.md` files at depth ≤3, plus `CONTRIBUTING.md`/`LICENSE`/etc. at the root, so this is not a contrived coincidence — it's a live landmine for any future hand-typed link with an off-by-one. This directly falsifies the PR's own central claim ("fails the build on ... a link it cannot resolve. No placeholder ... anywhere, so a green build is itself the evidence"): a green build is *not* evidence here, because the recovery heuristic launders an unresolvable-as-typed link into a resolvable-but-wrong one. It's also unnecessary: I grepped all six changelogs and found exactly **one** relative link anywhere in the corpus that needs this fallback at all — `crates/trusty-mpm/CHANGELOG.md` line 1982, `[...](../../../docs/specs/agent-bundled-skills.md)`. The PR's own "changelog is append-only history, can't be edited" rationale (correctly used to justify the metavariable recovery elsewhere) doesn't apply to fixing a relative-link typo — correcting `../../../` to `../../` on that one line is a factual correction, not a rewrite of history, and the doc reader (`$lib/docs/render.ts`) already fails hard on any unresolvable link with no analogous recovery. | Delete the recovery block (L387–393); fix the one real corpus line (`crates/trusty-mpm/CHANGELOG.md:1982`) to use the correct `../../` depth; let the existing `CHANGELOG-BAD-LINK` gate do its job on anything else. Replace the now-misleading "recovers a relative link written with one `../` too many" test in `parse.test.ts` with a negative test proving a coincidental-match case (e.g. the README example above) is rejected. | Fix here |
| MEDIUM | `website/tests/build-smoke.test.ts` | `describe('production build', …)` | This file's own stated purpose is to catch the class of bug its header names explicitly: "a page whose component threw during prerender... still writes an HTML shell and still passes it." It exercises exactly that check for `/docs/**` (`'renders real documentation, with rewritten links and no blob/main'`) and for the six `/tools/<slug>` pages (`'renders each flagship page from its own tool record, not a shell'`), but has no equivalent assertion for `/whats-new` or for the new "What's new" strip on the landing page — `'renders real landing-page content, not a shell'` (L239–246) checks flagship names/slugs/install commands but never the strip's version/date/item text. A `+page.svelte` template bug (e.g. a wrong field access, a `{#if}` that silently skips a card) that doesn't throw would render without failing prerender and nothing in this suite would notice — the same risk class this file was written to catch for every other surface. The underlying data (`buildChangelogSite`) is well covered at the unit level, so this is a coverage gap in the one layer that proves the actual shipped artifact isn't a shell, not a live bug I could reproduce. | Add to `build-smoke.test.ts`: assert `.vercel/output/static/whats-new.html` exists and is non-trivial in size, contains each flagship's newest version string, and (mirroring the docs check) contains no unrewritten `.md` relative links; extend `'renders real landing-page content, not a shell'` to assert the strip's version/date text appears for each flagship. | Fix here |

## Required Changes

1. Remove or replace the out-of-repo relative-link "recovery" heuristic in `harden()` (`website/src/lib/changelog/parse.ts` ~L386–393) so a link that doesn't resolve at its literal path always fails the build — no coincidental-match fallback. Fix the one real corpus line that currently needs it (`crates/trusty-mpm/CHANGELOG.md:1982`) directly. Update/replace the corresponding test in `parse.test.ts` to prove a coincidental wrong-match is rejected, not silently accepted.
2. Extend `website/tests/build-smoke.test.ts` to assert `/whats-new` and the landing-page strip render real, crate-specific content in the actual production build output — not just in the `site.ts` unit tests.

## Verification

**Gates run locally, from inside `website/`:**
- `pnpm check` → `COMPLETED 481 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`
- `pnpm lint` → prettier clean, eslint clean
- `pnpm test` → `Test Files 13 passed (13)` / `Tests 215 passed (215)`
- `pnpm build` → succeeds; `.svelte-kit/output/prerendered/pages/whats-new.html` = 401,944 bytes (93,525 bytes gzipped), matching the PR's claimed 402 KB/94 KB

**Mutations re-derived by hand (not trusted from the PR body), each run against the real test file, each confirmed to fail exactly the expected test(s):**
1. `parsePage(stripLinkDefinitions(md))` → `parsePage(md)` in `parseChangelog` — breaks 2 tests in "the link-reference trap" (proves the guard is actually bound to the parse path, not just to `stripLinkDefinitions` in isolation).
2. `categoryKey` returning the whole label — breaks 2 tests in "category headings".
3. `parseReleaseHeading` returning a release object instead of `undefined` on an unclosed `[` bracket — breaks "fails the build on a `## [` line that never closes its bracket".
4. `RELEASE_HEADING` regex made unmatchable — breaks 34 tests across `parse.test.ts`/`site.test.ts` (every real trusty-git-analytics release fails to parse, cascading to `CHANGELOG-NO-RELEASES`).
5. `CHANGELOG_CODES.MISSING` swapped to `'CHANGELOG-NO-RELEASES'` — breaks 2 tests asserting the specific `MISSING` code.
6. `if (!latest)` gate disabled (`if (false)`) — breaks 3 tests (zero-release and bad-heading fixtures no longer produce the expected `NO_RELEASES`/`BAD_RELEASE` failure sequence).
7. `for (const flagship of FLAGSHIPS)` → `FLAGSHIPS.slice(0, 4)` — breaks 6 tests, including "covers exactly the six flagships" and "keeps the other 21 crates out of both surfaces".

**Live end-to-end gate proof (not just unit-level):** temporarily stripped all bullets from `crates/trusty-review/CHANGELOG.md`'s newest release (leaving only the heading + preamble) and ran the real `pnpm build`. It died with exit 1 and:
```
ChangelogBuildError: changelog build gate: 1 finding(s)
FAIL CHANGELOG-EMPTY-LATEST crates/trusty-review/CHANGELOG.md:9: the newest release `0.11.0` has no items under any category, so the card strip would be blank — ...
```
No `.vercel/output` artifact was left in a servable state. Restored the file afterward; worktree is clean.

**Item 1 (link-reference trap) verified clean in the real prerendered output:** `grep -c 'bobmatnyc/trusty-search' .svelte-kit/output/prerendered/pages/whats-new.html` → 0; no `<h3>` version heading in the built HTML is wrapped in an `<a>`.

**Item 5/6 verified:** `git diff --name-only origin/main...HEAD` touches only the 12 files under `website/`; nothing under `crates/` changed; no `crates/*/CHANGELOG.md` is written anywhere outside the test fixtures in `site.test.ts` (which write to `mkdtempSync` temp dirs, not the repo).

## Notes

- Item 3 (unbackticked-metavariable recovery in `harden()`) — I traced through this carefully and did not find a real content-loss or injection risk >80% confidence: the recovery only fires for genuinely unrecognized tag names (the `ALLOWED_ELEMENTS` allowlist already covers legitimate markup), the replacement text is a plain text node so it's HTML-escaped on serialization (no injection even for something like a stray `<script>`), and the real-corpus test (`'keeps every metavariable in the real corpus, dropping none'`) pins all four known sites. Scoped correctly as far as I can tell.
- Item 4 (truncation to `DETAILED_RELEASES = 3`) — the "N earlier releases" `<details>` is clearly labeled and links back to the source file each time; I don't think this overstates completeness.
- The CRITICAL finding above is about the *general mechanism*, not the one real corpus link (which currently resolves correctly). Nothing on the live site is broken today; the risk is the next hand-typed changelog entry with a link-depth typo shipping silently wrong.

bobmatnyc:
## Round 2 verdict: APPROVE — supersedes the round-1 BLOCK

Re-reviewed at head `91877501c` from a fresh detached worktree (`.claude/worktrees/critic-5248-r2`, fetched from `origin/main`). Did not take the author's report as proof — reproduced the original attack myself, tried to defeat the new gate on my own terms, re-ran every gate from inside `website/`, and live-mutated both the parser and the Svelte templates to confirm the new tests actually catch what they claim to.

Both round-1 findings are genuinely discharged. No new CRITICAL or HIGH findings.

## What changed (`d19be5b78..91877501c`)

`crates/trusty-mpm/CHANGELOG.md` (1 line, typo fix) · `website/src/lib/changelog/{README.md,parse.ts,parse.test.ts,site.test.ts}` · `website/tests/build-smoke.test.ts`. `website/src/lib/docs/render.ts` is untouched (empty diff against round 1).

## Round-1 CRITICAL — `parse.ts` `harden()` escape-the-repo recovery

**Fixed. Verified independently, not from the report.**

- The `recovered`/`probe(recovered)` fallback is gone. `target` is now `const`; `target.startsWith('..')` is an unconditional `CHANGELOG-BAD-LINK`, no second chance.
- Reproduced my exact round-1 attack — `[the README](../../../README.md)` from `crates/trusty-search/CHANGELOG.md` — against the new code: now produces 1 failure, `CHANGELOG-BAD-LINK`, `problem: "... resolves to \`../README.md\`, outside the repository"`. No href is rewritten.
- Tried to defeat the closed gate myself, beyond what the diff's own tests cover:
  - **Depth-agnostic**: 2, 3, 4 extra `../` — all fail (`CHANGELOG-BAD-LINK`), not just the one-level case.
  - **`foo/../../../docs/specs/agent-bundled-skills.md`** (re-entry via `a/../..`) — normalizes to a real in-repo path *directly* (not via the removed recovery branch) and resolves correctly to `blob/main/docs/specs/agent-bundled-skills.md`. Not a vulnerability: this is ordinary `path.posix.normalize` resolution to a target that genuinely exists, unrelated to the removed heuristic.
  - **URL-encoded traversal** (`%2e%2e/...`, `..%2f...`) — `path.posix` does not decode `%2e`/`%2f`, so these are treated as literal (non-existent) filenames and fail `CHANGELOG-BAD-LINK`, never resolving to the real README.
  - **Windows-style backslashes** (`..\..\..\README.md`) — markdown's link normalizer percent-encodes the backslashes before this code ever sees them; the result doesn't match any real path and fails closed.
  - **Absolute filesystem path** (`/etc/passwd`) — already caught earlier in `harden()` by the `rawTarget.startsWith('/')` check, unrelated to this fix, unchanged, still fires.
  - **`file://...` / `javascript:...`** — these match the pre-existing `ABSOLUTE` regex (`/^(?:[a-z][a-z0-9+.-]*:|\/\/)/i`) and are classified as external, same as `https://`, and pass through untouched with no scheme allowlist. This is **not new** — the identical `ABSOLUTE` pattern and same early-return already exist in `website/src/lib/docs/links.ts:75,104-105` for the doc reader, so it's shared, pre-existing, out-of-scope-for-this-fix behavior, not a regression introduced here. The real six-changelog corpus contains no such scheme links (confirmed by the new `site.test.ts` assertion below). I flag this only as a Note, not a finding — see below.
- All eight of the above were run as actual `parseChangelog(...)` calls against the built code, not asserted from documentation.

**Corpus claim verified independently.** `grep`'d all six `CHANGELOG.md` files myself for any `](...)` target that isn't `https?://` or `#...`: exactly 3 hits, all three in `crates/trusty-mpm/CHANGELOG.md` (lines 1767, 1771, 1982), all now `../../docs/...` (correct depth for a file two levels under `crates/`). The line-1982 diff is a one-character-class fix (`../../../` → `../../`) to an existing link's target path — a factual correction to a URL, not a rewrite of the surrounding prose/history. No other line in that file, or any of the other five changelogs, changed. `git diff --name-only origin/main...HEAD -- crates/` shows exactly this one file, one line.

**New regression tests are properly targeted**, not just descriptive:
- `parse.test.ts`: the old (now-wrong) "recovers a relative link written with one `../` too many" test was replaced — not left stale — with `fails the build on a link that escapes the repository, even when the stripped path exists`, which deliberately probes `'README.md'` as existing (the exact condition that made the old code fail open) and asserts `CHANGELOG-BAD-LINK` + no `blob/main` href reaches the item. Plus a depth-sweep test (`../../../x.md` / 6-deep / `a/../` re-entry).
- `site.test.ts`: new test asserts every href emitted across the real corpus (500+ links) matches `^https?://` — this transitively proves no relative/scheme-oddity href leaks through anywhere in the real six files, not just at the unit level.

## Round-1 MEDIUM — `build-smoke.test.ts` missing `/whats-new` coverage

**Fixed, and verified to actually catch the failure class it claims to, not just to exist.**

- Two new assertions: `/whats-new` renders real content per crate (version, opening 5 words of the newest item's first bullet, section anchor, changelog link), and the landing-page strip does the same.
- The expected strings come from `newestRelease()`, which re-reads each `CHANGELOG.md` directly with its own regex — not through `$lib/changelog` — so the test cannot pass merely because the parser agrees with itself. This directly satisfies this repo's "a test must fail in the configuration the defect occurs in" standard, and is stronger than what I asked for in round 1.
- **I did not trust this from the diff — I manufactured the exact defect shape the task described** ("every heading, anchor, version, release count, changelog link present, zero items") by setting `{#if block.kind === 'items'}` → `{#if false}` in `whats-new/+page.svelte`, ran the real smoke test: it failed with `AssertionError: trusty-search is missing its newest item: expected '...' to contain 'Warm boot paid a full'` against the full rendered page text — confirming the assertion checks real bullet content, not a string that happens to appear elsewhere on the page (headings, versions, and release counts were all still present in the failing output, exactly as described).
- Did the same for the landing-page strip (`{#if strips.get(flagship.name)}` → `{#if false}` in `+page.svelte`): smoke test failed with `trusty-search card has no All-changes link`, confirming that assertion is live too.
- Restored both files after each test; worktree left clean (`git status --short` empty).

## Gates re-run myself, from inside `website/`

- `pnpm install` → clean
- `pnpm check` → `COMPLETED 481 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`
- `pnpm lint` → prettier clean, eslint clean
- `pnpm test` → `Test Files 13 passed (13)` / `Tests 219 passed (219)` — matches the author's reported count exactly
- `pnpm build` → succeeds; `whats-new.html` = 401,945 bytes (93,527 gzipped), consistent with round 1
- `src/lib/docs` suite (unaffected by this diff) → `Test Files 6 passed (6)` / `Tests 80 passed (80)` — confirms no regression to `/docs/**` through the shared `render.ts` exports (which have an empty diff between the two heads)

## Findings

## Notes

- The `file://`/`javascript:` absolute-scheme passthrough in `harden()` (shared `ABSOLUTE` pattern, also present in `website/src/lib/docs/links.ts`) is real but pre-existing, unrelated to this fix, not present in the current six-file corpus, and only reachable via a first-party PR-reviewed `CHANGELOG.md` edit — not user/request-controlled input. Below the bar for a finding on this diff; noting it for visibility only, per the framework's "don't flag unchanged code" guard rail.
- Round-1's CRITICAL is fully retired: I could not reconstruct any variant of the original attack, or any of the additional vectors I tried, that reintroduces a silently-wrong link.

PR #5253, opened 2026-08-09
fix(trusty-common,trusty-analyze,trusty-review,trusty-console): drain the webhook inbox into the review and analysis pipelines (#5089 step 5)

### LINKED ISSUE #5192: #5089 step 5: nothing drains the webhook inbox into the review or analysis pipeline
Defect: the UDS receive path takes durable ownership of a webhook delivery and acks it, but nothing reads it back out. `Inbox::list` (`crates/trusty-common/src/webhook_relay/inbox.rs`), the one read method — its own doc says "Why: what a drain step reads" — has zero production call sites. All ten call sites are tests: `crates/trusty-common/src/webhook_relay/tests.rs` (7) and `crates/trusty-console/src/webhook/tests.rs` (3). `Inbox::take_ownership` is the only write path, reached from `dispatch_frame` (`crates/trusty-common/src/webhook_relay/serve.rs`) via `impl DeliverySink for Inbox`. `dispatch_frame` parses, validates, takes ownership, and acks — it never reads the inbox back or hands anything onward. Both receivers say so in their own module docs, `crates/trusty-analyze/src/webhook_listener.rs` and `crates/trusty-review/src/webhook_listener.rs`: "Draining the inbox into the analysis/review pipeline is a separate step; what matters here is that nothing is lost while it waits."

What the legacy HTTP handlers do that the UDS path does not:
- trusty-analyze: `github_webhook_handler` (`crates/trusty-analyze/src/service/handlers/review.rs`) spawns `process_pr_webhook`, which fetches the PR diff from GitHub, runs `analyze_diff_with_client` against trusty-search, formats a markdown report, and posts it as a PR comment.
- trusty-review: `handle_github_webhook` (`crates/trusty-review/src/service/webhook.rs`) does HMAC verification, filters on the `review_requested` action, classifies live-vs-dry-run, takes a per-PR in-flight guard, and spawns the review pipeline.

🔴 This blocks [#5181](https://github.com/bobmatnyc/trusty-tools/issues/5181). Retiring the legacy routes before this lands means a webhook reaches an inbox file and stops — no review, no analysis, no comment.

🔴 [ADR-0034](https://github.com/bobmatnyc/trusty-tools/blob/main/docs/adr/0034-webhook-ingress-console-relays-over-uds-to-a-supervised-on-demand-process.md) has no section describing this step. Its four numbered decisions are residency/supervision, the spool, the trust boundary, and the transport/framing. Amending the ADR to name this step is part of this work.

Resolution: consume inbox entries in each crate and drive the same pipeline the legacy handler drove. Entries are already durably owned, so the drain must be restart-safe and must not lose or double-process an entry.

Part of https://github.com/bobmatnyc/trusty-tools/issues/5089.

Rung 5. Requires a code-critic round before merge.

bobmatnyc:
## Defect

`Inbox::list` — the inbox's only read method, whose own doc said "what a drain step reads" — had zero production call sites. #5182 made the receiver take durable ownership of a relayed delivery and acknowledge it, and the ack is what licenses `trusty-console` to delete its only copy. From that moment the delivery existed in exactly one place and nothing read it back: a webhook reached a JSON file under `webhook-inbox/` and stopped. No review, no analysis, no PR comment — and console's spool reported empty, because the hand-off had succeeded.

See #5192 for the full inventory.

## Resolution

`trusty_common::webhook_relay::drain` reads the inbox back out and drives each target's pipeline, in the same short-lived console-supervised process that binds the socket. The ack is unchanged and still rests on durability alone, so a slow or failing review can never turn into a refused delivery.

Four rules, and the mechanism that holds each:

| Rule | Mechanism |
|---|---|
| Nothing is destroyed before the pipeline accepted it | `retry::remove_processed` is the only call that DESTROYS a delivery, reachable from one arm of `drain_once` — the arm behind `Ok(_)` from `DeliveryProcessor::process`. `retry::quarantine` also unlinks, but only after `hard_link` has put the same inode under `<inbox>/quarantine/`, so it moves a delivery and never destroys one |
| One delivery is processed at most once | `retry::mark_processed` records it under `<inbox>/processed/` **before** the unlink, and `retry::is_processed` is consulted before any processor runs |
| A processing failure is visible | Every non-success lands in `DrainReport::failures` with the delivery id and the processor's own words; `DrainReport::log_summary` raises it at `error!`. Bounded retry, then quarantine — never a swallowed `Err` and never a delete |
| Counts are true under failure | `DrainReport::processed` counts entries accepted **and** removed. An acceptance whose unlink failed is `FailureOutcome::Stuck`, counted nowhere else, so the number under-reports success rather than inventing it. `DrainReport::accounted() == scanned` is asserted |
| Health reflects a stuck inbox | Nothing deletes work to make a number go down, so `held_count` keeps reporting the backlog. Quarantining takes an entry *out* of that count, so `quarantined_count` is metered separately and is `Error` in `SpoolHealth` |

That crash-safety design is itself what makes a duplicate possible, so the drain — not each receiver — closes it. A drainer can return `Ok` from the pipeline and die before the entry is unlinked; the entry is then claimable by design, and `trusty-analyze` (which posts a PR comment and has no dedup store of its own) would post a second identical comment, unbounded on repeated crashes. The ledger under `<inbox>/processed/` gives every receiver the `delivery_id` deduplication the relay contract requires, from one implementation. Markers are pruned after 30 days.

Crash safety is an `flock` on the entry file rather than a claim marker on disk: the kernel releases it when the holding fd closes, including on SIGKILL, so "crashed mid-review" and "finished" are the same state to the next drainer — and that state is claimable. A marker would instead strand the delivery forever, which is #5182's silent loss arriving one hop later. The `st_nlink == 0` check after the lock is what stops the loser of a contended lock processing an inode the winner already unlinked; it is extracted as `entry_is_still_linked` and proven directly, because the race it guards cannot be scheduled from outside `try_acquire`.

Both targets supply their pipeline through an injectable trait (`PrPipeline` / `ReviewPipeline`), and their legacy HTTP routes now call the drain's event classifier, so the UDS and HTTP transports cannot drift on what a webhook means. #5181 then deletes a route, not a pipeline.

ADR-0034 gains §5 describing this step, per #5192.

**Known gap, stated rather than papered over:** `trusty-review`'s outcome poll on `closed` + `merged` (opt-in, default off) schedules a task that sleeps an hour. A process that exits on SIGTERM cannot honour that, so the drain records those deliveries as deliberately ignored and the legacy route remains their only path. Called out in ADR-0034 §5; blocks nothing in criterion (c).

## Evidence

Rung 5 (cross-crate contract + process lifecycle). Every failure-path test drives a pipeline that actually fails, in the configuration the defect occurs in — an injected failing pipeline rather than an unreachable host, so the branch is exercised regardless of whether `GITHUB_TOKEN` happens to be set on the runner.

Coverage for each required case:

- **accepted → acked and removed** — `drain_removes_an_entry_the_processor_accepted`
- **pipeline FAILS → not lost, not falsely counted** — `drain_keeps_an_entry_whose_processor_failed` (asserts `processed == 0`, entry still held, reason reaches the report), plus `processor_reports_a_failed_pipeline_as_retryable` in both target crates
- **interrupted mid-processing → claimable** — `drain_leaves_an_interrupted_entry_claimable` (deterministic: hold the claim, prove lockout, drop, prove the next pass completes it) and `claim_is_released_when_the_holder_is_dropped`
- **concurrent drainers → no double-process** — `drain_does_not_double_process_a_claimed_entry`, interleaving pinned by a barrier rather than a sleep, plus the deterministic `claim_is_exclusive_between_two_holders` for the underlying property
- **crash between acceptance and unlink → no duplicate side effect** — `drain_posts_exactly_one_comment_when_a_drainer_dies_before_the_unlink` (trusty-analyze, over a real inbox through the real drain: **2 posts before the fix, 1 after**), plus `drain_does_not_reprocess_a_delivery_marked_done`, `drain_does_not_reprocess_a_console_redelivery_of_a_drained_id`, `drain_marks_a_processed_delivery_before_removing_it`, and `drain_keeps_an_entry_whose_processed_marker_could_not_be_written`
- **counts and health under partial failure** — `drain_report_accounts_for_every_scanned_entry` (one pass with every outcome; `accounted() == scanned`), `drain_processed_count_excludes_an_acceptance_whose_entry_could_not_be_removed`, and `health_is_error_while_a_target_holds_a_quarantined_delivery` (green against this change with `quarantined_total` removed)

```
cargo fmt --check
cargo clippy -p trusty-common --features webhook-relay,uds-supervisor --all-targets -- -D warnings
cargo clippy -p trusty-analyze -p trusty-review --all-targets -- -D warnings
cargo clippy -p trusty-console --all-targets -- -D warnings
cargo test -p trusty-analyze -p trusty-review
cargo test -p trusty-common --features webhook-relay,uds-supervisor --lib
cargo test -p trusty-console
bash scripts/check_sld.sh
```

All pass. trusty-analyze 474 + 27 + 2 + 5 + 2; trusty-review 1599 + 68 + integration suites; trusty-common lib 464 (36 `drain_tests`); trusty-console 225. `line-cap: 0 violations`, `sld-lint: 0 errors, 0 warnings`.

## API surface — breaking, correcting an earlier claim in this PR

`webhook_listener::run` went from zero-arg to requiring a parameter in **both** `trusty-analyze` (`TrustySearchClient`) and `trusty-review` (`ReviewConfig`). An earlier revision of this body called the PR's API surface "additive, non-breaking"; that was wrong. Under Cargo's 0.x rule this needs the **MINOR** position bumped at release for both crates. `ReviewDeps` additionally derives `Clone`, which is additive. `trusty-common` gains a module (additive) and `trusty-console`'s `SpoolHealth` gains fields (additive, schema version 3).

No version is bumped in this PR — releases are `local-ops`' job. This section exists so the release step is not working from a wrong characterization; a workspace `cargo check` structurally cannot catch this class of break (#4088).

Closes #5192
Part of #5089


bobmatnyc:
## Verdict: BLOCK

Reviewed at head `f5c0c9964e727425dea925bf7996658e62a12d21`, detached worktree, feature-gated build (`--features webhook-relay,uds-supervisor` for trusty-common). All four crates' relevant test scopes pass (trusty-common webhook_relay: 71/71, trusty-review webhook_drain: 10/10, trusty-analyze webhook_drain: 10/10, trusty-console webhook incl. `--include-ignored`: 79/79), clippy `-D warnings` clean on all four, SLOC cap clean, no new `#[ignore]`/`--exclude`/scope-narrowing found.

The four claimed mechanisms hold up **except one**, and it's the same failure shape this epic has shipped twice already (#5183, step 4): a safety property is documented and implemented for one sibling crate but silently absent in the other.

## Findings

| Severity | File | Symbol/Line | Issue | Fix | Disposition |
|---|---|---|---|---|---|
| CRITICAL | `crates/trusty-analyze/src/webhook_drain.rs:242` (`AnalyzeProcessor::process`) + `crates/trusty-analyze/src/core/github.rs:118` (`post_pr_comment`) | — | No deduplication gate. ADR-0034 §5 rule 3 states the at-least-once contract's safety net as a blanket rule: "receivers deduplicate on `delivery_id`." `trusty-review`'s sibling processor implements exactly this (`webhook_drain.rs:234`, `DedupNeed::Required`, changelog fragment: *"open the dedup store as `Required`, so a redelivery cannot post a second review comment"*) — `trusty-analyze` has **zero** dedup infrastructure anywhere in the crate (`grep -rl DedupNeed\|dedup.redb crates/trusty-analyze/src/` is empty). `post_pr_comment` is an unconditional `POST` with no idempotency key or existing-comment check. Concrete failing input: claim an entry, let `AnalyzeProcessor::process` return `Ok(Disposition::Processed)` after `run_pr_analysis` has already posted the comment, then kill the process (SIGKILL, or SIGTERM past console's flush window — the listener's own doc at `listener.rs:229-237` says an aborted drain "is safe for the same reason a crash is") before `retry::remove_processed` runs. The entry survives (by design — that's the whole crash-safety point), the next drain claims it again, and `run_pr_analysis` posts a second, identical PR comment. Nothing bounds this to once: it isn't a retry-count failure, so `DEFAULT_MAX_ATTEMPTS` never engages. | Gate `AnalyzeProcessor`'s post behind the same class of protection `ReviewProcessor` uses — reuse the dedup-store pattern (keyed on `delivery_id` or owner/repo/pr+head_sha) rather than building a second implementation. If genuinely out of scope for this PR, at minimum this needs the same explicit ADR-0034 §5 "known gap" treatment the outcome-poll item got, plus a test asserting the pipeline is NOT re-invoked on a second claim of an already-processed-but-not-yet-removed entry. | Fix here |
| MEDIUM | `crates/trusty-common/src/webhook_relay/retry.rs:20-22` (module doc) and `crates/trusty-common/src/webhook_relay/drain.rs:19-21` (rule 1) | `retry::quarantine` at `retry.rs:212` | The invariant "`remove_processed` is the only path... that unlinks [a live delivery]" is false as written: `retry::quarantine` also calls `std::fs::remove_file(entry)` (`retry.rs:212`), reachable from `drain_once` while a `Claim` is still held (the out-of-retries arm at `drain.rs:397-409`) — not only from the claim-free undecodable path the nearby comment describes. Functionally safe (hard-link into `quarantine/` precedes the unlink, and it's idempotent on `EEXIST`/`NotFound`), so no data is lost — but the stated invariant is what a future change would rely on, and it doesn't hold as written. ADR-0034 §5 rule 2 makes the identical overstatement. | Reword to something like "the only unlink that discards a delivery outright — `quarantine` also unlinks, but only after the entry is durably linked into `quarantine/` first." | Fix here |
| MEDIUM | `crates/trusty-common/src/webhook_relay/drain_tests.rs:236-256` (`claim_refuses_an_entry_unlinked_while_the_lock_was_contended`) | vs. `crates/trusty-common/src/webhook_relay/claim.rs:185` (`nlink() == 0`) | This test is the one the module doc singles out ("🔴 The `nlink` check ... is not defensive padding") but it doesn't reach that branch. It calls `std::fs::remove_file(&path)` *before* `Claim::try_acquire(&path)`, so `try_acquire`'s `File::open` hits `NotFound` and returns `Vanished` at `claim.rs:152-156` — the early bail-out, never reaching the flock-then-`nlink()` check at `claim.rs:181-187`. The test's own comment admits the substitution: "Re-create the *name* pointing at nothing is not possible; instead assert... Vanished, which is the observable the drain acts on." That's a different, easier claim than the one the docstring advertises. I traced the actual `nlink()` logic by hand and it looks correct, so this is a coverage gap on correct code, not a live bug — but the specific safety property this review was asked to scrutinize hardest has no test that exercises it. | Reproduce the real race with two real threads/fds and a sync barrier (the `BlockingProcessor` pattern already used for `drain_does_not_double_process_a_claimed_entry` is the template): open two fds on the same path, lock+unlink+drop on one, then flock+stat on the second — that's the only way to hit `nlink()==0` after a successful open. | Fix here |
| MEDIUM | `crates/trusty-review/src/webhook_listener.rs:97`, `crates/trusty-analyze/src/webhook_listener.rs:109` (`pub async fn run`) | `pub mod webhook_listener` at `trusty-review/src/lib.rs:76`, `trusty-analyze/src/lib.rs:82` | Both `webhook_listener::run` went from a zero-argument public async fn to one requiring a new parameter (`ReviewConfig` / `TrustySearchClient`). Both modules are `pub mod`, so this is a breaking public-API change on a published crate under Cargo's 0.x rule (trusty-review 0.12.0, trusty-analyze 0.8.0 — both need a MINOR bump, not patch). Everything else in the "additive" claim checks out: `ReviewDeps: Clone` is a pure trait-impl addition, and `trusty-common`'s new `claim`/`drain`/`retry` modules plus `run_until_signal_with_processor` are additive alongside the untouched `run_until_signal`. This doesn't block the PR itself (version bumps are release-time per this repo's workflow, and `preflight-publish.sh`'s `cargo-semver-checks` should catch it), but it should be called out explicitly rather than characterized as non-breaking, so whoever cuts the next release doesn't reach for a patch bump. | No code change needed; note in the PR body that trusty-review and trusty-analyze both need a MINOR version bump at next release. | Parent |

## Verified, not flagged

- **Claim 1 (never consumed before accepted):** True except for the `retry::quarantine` unlink noted above (which preserves data, so the *substance* of the claim holds even though the *wording* doesn't).
- **Claim 2 (failure visibility):** `DrainReport::log_summary` raises at `error!` for anything not clean; no swallowed `Err` found in `drain_once`'s match arms.
- **Claim 3 (counts true under failure):** `DrainReport::accept` correctly reports a processor-accepted-but-unlink-failed entry as `FailureOutcome::Stuck`, not `processed` — verified with a deterministic readonly-directory test (`drain_processed_count_excludes_an_acceptance_whose_entry_could_not_be_removed`), not a timing-dependent one. `accounted() == scanned` is test-only (not a runtime `debug_assert`), by design — it's exercised across a mixed-outcome pass in `drain_report_accounts_for_every_scanned_entry`, which is reasonable given the match in `drain_once` is exhaustive over `ClaimOutcome`.
- **Claim 4 (health reflects a stuck inbox):** `quarantined_count` → `scan_undrained` → `quarantined_total > 0` → `ServiceHealth::Error` (`health.rs:245`), wired end to end and covered by `health_is_error_while_a_target_holds_a_quarantined_delivery`. `held_count`'s `.json`-extension filter correctly steps over both the `quarantine/` subdirectory and `.attempt` sidecars (non-recursive `read_dir`).
- **Author-flagged item 1 (outcome-poll gap):** Matches ADR-0034 §5's "What this amendment does not cover" text — `classify_review_event` has no `closed`/`merged` special case, so those deliveries fall into the generic non-`review_requested` → `Ignored` branch and are reported as such. Honestly reported, not silently lossy.
- **Crash safety / TOCTOU:** The lock→stat(nlink)→read sequence in `Claim::try_acquire` looks correct by manual trace (see the MEDIUM finding above for why I can't say "tested correct"). The window between `process()` returning `Ok` and `remove_processed()` completing has no `.await` point, so `JoinHandle::abort()` on graceful shutdown can't land inside it — the real exposure is OS-level SIGKILL, which the `flock`-release design handles by construction for `trusty-review` (dedup-gated) but not for `trusty-analyze` (see CRITICAL finding).

## Confidence notes

All findings above are held at >80% confidence; the CRITICAL finding I'd put at ~95% given I confirmed by grep there is no dedup mechanism anywhere in `trusty-analyze`, confirmed `post_pr_comment` has no idempotency check, and confirmed the sibling crate's changelog fragment explicitly credits its own dedup gate with preventing the exact same failure mode.

bobmatnyc:
## Verdict: WARN (round-1 CRITICAL resolved; one new HIGH — a required CI check is currently red and blocks merge regardless of this verdict's label)

Round 2, reviewed at head `fe05c964e79101d97745e4c3f02c00704cd780dc`, fresh detached worktree, same feature-gated build. This supersedes my round-1 BLOCK.

## Round-1 CRITICAL: resolved

The missing-dedup finding (`AnalyzeProcessor` posting PR comments with no protection against reprocess-after-crash) is fixed, and fixed at the right layer. Rather than copying `trusty-review`'s `DedupNeed::Required` into `trusty-analyze` (which would have been the second parallel implementation the common-entry-point rule exists to stop), the drain itself now owns deduplication: `retry::mark_processed` (`crates/trusty-common/src/webhook_relay/retry.rs:227`) records the delivery under `<inbox>/processed/` *before* `remove_processed` unlinks it, and `drain_once` (`crates/trusty-common/src/webhook_relay/drain.rs:444`) consults `retry::is_processed` before ever calling a processor. This is the correct fix location — the reprocess window is created by the drain's own crash-recovery design (flock released on death), not by an omission in each receiver.

**Verified, not just read:**
- **Ordering is right.** `DrainReport::accept` (`drain.rs:315`) calls `mark_processed` only after `processor.process().await` returned `Ok`, and only calls `remove_processed` after `mark_processed` succeeded. A `mark_processed` failure keeps the entry rather than removing it unrecorded (`drain.rs:316-331`, covered by `drain_keeps_an_entry_whose_processed_marker_could_not_be_written`).
- **The crash-between-mark-and-unlink window is genuinely closed, not just narrowed on paper.** I disabled the `is_processed` gate (`if false && retry::is_processed(...)`) and reran `drain_does_not_reprocess_a_delivery_marked_done` and `drain_does_not_reprocess_a_console_redelivery_of_a_drained_id` — both fail without the fix (`2 failed; 0 passed`), confirming they exercise the actual defect rather than passing vacuously. Restored the file afterward; diff against the pre-edit copy is clean.
- **The marked-but-not-yet-removed entry doesn't linger.** On the next drain pass, `is_processed` is true, so the entry is unlinked via the `deduplicated` branch (`drain.rs:450-459`) without re-invoking the processor — it does not become a permanent skip that inflates the held count or the health signal.
- **No new concurrent-write hazard.** `mark_processed`/`is_processed` are only ever reached while a `Claim`'s flock is held on that entry, so two drainers can't race the same marker; `write_replace`'s tmp-then-rename with a PID-qualified tmp name means even a same-path write from two processes can't tear.
- **`trusty-review` still works and isn't double-guarded.** Its SHA-keyed `DedupStore` (pre-reservation `InProgress`/`Completed`, `crates/trusty-review/src/store/dedup.rs:109`) and the drain's new delivery-id-keyed ledger key on different axes and don't conflict — a legitimate new delivery for an already-reviewed SHA still reaches the pipeline (passes `is_processed`, since it's a new delivery id) and is correctly short-circuited by the *inner* SHA dedup instead. I did not find a case where the new outer gate drops a delivery the inner gate would have accepted.
- **30-day pruning is an honestly disclosed, bounded residual risk**, not a hidden gap — the module doc states plainly that an early-pruned marker plus a redelivery reproduces one duplicate, and scopes that to deliveries nobody has touched in a month.

One caveat I want on record rather than silently accepting: the ledger protects the window from `process()` returning `Ok` through the unlink. It does **not** protect a crash *during* `process()` itself — e.g., mid-network-call, after GitHub has accepted the POST server-side but before the client sees the response and returns control to the drain. In that case `process()` returns `Err` (transport failure), the marker is never written, and a retry can re-post. This is a pre-existing, fundamental limitation of at-least-once delivery against a non-idempotent external side effect with no server-side idempotency key — it equally applies to `trusty-review`'s own SHA-store when an `InProgress` claim goes stale and is reclaimed — and it is not what this fix claims to close (the module doc's language is correctly scoped to "returned `Ok`, not yet unlinked"). Not a finding; noting it so it isn't mistaken for solved.

## MEDIUM findings from round 1: resolved

- **Unlink-invariant wording** — `retry.rs:20-27`, `drain.rs:19-26`, and ADR-0034 §5 rule 2 now correctly distinguish `remove_processed` (destroys) from `quarantine` (moves, hard-link first). Matches the code.
- **`nlink()` test coverage** — the old mislabeled test was renamed to what it actually proves (`claim_of_a_removed_entry_reports_vanished`, `claim.rs`/`drain_tests.rs:236`), and the TOCTOU predicate was extracted to `claim::entry_is_still_linked` (`claim.rs:245`) with two direct unit tests. I agree with the author's reasoning that the true race (an unlink landing inside `try_acquire`'s own open-to-stat window) cannot be scheduled from a black-box test without adding a test-only synchronization hook to production code, which this project's own testing standards reject. Extracting and directly testing the predicate is the right call given that constraint — the surrounding `try_acquire` composition is now three trivial lines (`if !predicate { return Vanished }`), structurally identical to the already-tested `NotFound` bail-out immediately above it, so the residual untested-composition risk is low. Treating this as resolved, not still open.
- **Breaking-signature disclosure** — `crates/trusty-analyze/changelog.d/5192-webhook-listen-signature.md` and the `trusty-review` equivalent now correctly file under `Breaking`, and `scripts/assemble-changelog.sh trusty-analyze --stdout` / `... trusty-review --stdout` render them under a `### Breaking` heading. Confirmed by running the assembler myself, not by reading the fragment text alone.

## Changelog gate: confirmed green, independently

Ran `scripts/assemble-changelog.sh <crate> --stdout` for all four touched crates (`trusty-analyze`, `trusty-common`, `trusty-console`, `trusty-review`) — all four assemble cleanly, fragments render as proper `- ` bullets under correct category headers, exit 0. Also ran `scripts/check_changelog_fragment.sh` directly: `OK` for all four crates, "all 4 crate(s) with source changes are recorded." The author's fix claim is verified, not just trusted.

## New finding: a required CI check is currently red on `fe05c964`, unremarked

|---|---|---|---|---|---|
| HIGH | `crates/trusty-analyze/src/webhook_listener.rs:89`, `crates/trusty-common/src/webhook_relay/drain.rs:89`, `crates/trusty-common/src/webhook_relay/drain.rs:376`, `crates/trusty-common/src/webhook_relay/retry.rs:94` | doc-comment `Test:` pointers | `gh pr view 5253` on head `fe05c964e79101d97745e4c3f02c00704cd780dc` shows **"Test: doc-comment pointer lint" = FAILURE** (job [93198384238](https://github.com/bobmatnyc/trusty-tools/actions/runs/31294927855/job/93198384238)), separate from and not one of the stale rows for `9dc61bf8`/`f5c0c996` the coordinator flagged as noise. The gate reports 4 dangling `Test:` citations: (1) `webhook_listener.rs:89` cites `listener_with_a_processor_opens_against_a_temp_inbox` — no such fn exists anywhere in the workspace; (2) `drain.rs:89` cites `drain_retries_a_retryable_failure` — the actual test is named `drain_retries_a_retryable_failure_until_the_bound_then_quarantines`; (3) `drain.rs:376` cites `drain_quarantines_an_entry_that_is_out_of_retries` — actual name is `drain_quarantines_an_entry_whose_attempt_budget_was_already_spent`; (4) `retry.rs:94` cites `attempt_record_is_removed_with_its_entry` — no such fn exists. I traced all four against the round-2 diff: none of the four citation lines were touched between `f5c0c9964` (round 1) and `fe05c964` (round 2) — these are pre-existing from round 1, which I failed to catch in my round-1 review because I ran `cargo test`/`clippy`/the SLOC-cap script but not this lint. Regardless of origin, this is objectively red on the current head and a required check, so the PR cannot merge as-is. | Rename each `Test:` citation to the actual current function name (items 2–3), and either write the two aspirational tests (1, 4) or point their citations at whichever existing test actually covers that claim, or drop the dangling name if no such coverage exists. Run `bash scripts/check_test_pointers.sh` locally before pushing. | Fix here |

## Confidence notes

Everything above is held at >80% confidence. The one thing I did not independently re-verify against upstream sources: GitHub's exact manual-redelivery retention window (cited as "30 days" in `retry.rs:71-75` and the ADR amendment) — plausible and consistent with my recollection of GitHub's documented behavior, but I did not fetch GitHub's docs to confirm the exact figure. Immaterial to the verdict either way, since the design explicitly discloses and bounds the failure mode if that number is off (at most one duplicate for a delivery untouched in a month).

bobmatnyc:
## Verdict: APPROVE — supersedes the round-2 WARN

Round 3, head `65ba7b074854b36af56699919d2e0920b9fb5cbe`, fresh detached worktree. Targeted verification of the round-2 HIGH (dangling `Test:` doc-comment pointers) per the coordinator's brief; round-2 already cleared the CRITICAL and all three MEDIUMs and nothing in those areas changed in this diff (`git diff` round-2→round-3 touches exactly 3 files: `webhook_listener.rs`, `drain.rs`, `drain_tests.rs`).

No issues found at >80% confidence. APPROVED for merge.

## Verification performed

**The engineer's claim that 3 of 4 pointers cited tests that were never written, not just renamed — confirmed true, and each was fixed for real:**

1. **`attempt_record_is_removed_with_its_entry` — genuinely written, not a lint-satisfying stub.** Reverted `remove_processed`'s sidecar cleanup (`crates/trusty-common/src/webhook_relay/retry.rs`, removed the `let _ = std::fs::remove_file(attempt_path(entry));` line) and reran the test: it fails (`1 failed; 0 passed`, assertion at `drain_tests.rs:325`). Restored the file; diff against the pre-edit copy is clean. The test exercises exactly what its comment claims — a stale `.attempt` sidecar would charge a later arrival at the same path with failures it never had — and it fails without the behavior it guards.

2. **The two re-pointed citations for the never-existed `drain_quarantines_an_entry_that_is_out_of_retries`** (`crates/trusty-common/src/webhook_relay/drain.rs:89,377`) now cite `drain_retries_a_retryable_failure_until_the_bound_then_quarantines` and `drain_quarantines_an_entry_whose_attempt_budget_was_already_spent`. Read both bodies directly: the first drives three full `drain_once` passes and asserts `quarantined` flips only on the third, matching "bound reached during a pass"; the second pre-seeds `DEFAULT_MAX_ATTEMPTS` failure records *before any drain runs* and asserts `processor.calls() == 0` — the pipeline is never invoked because the budget was already spent, matching "budget already spent by an earlier process." Both citations point at tests that genuinely cover the claim, not a real-but-unrelated test.

3. **`draining_listener` deletion** (`crates/trusty-analyze/src/webhook_listener.rs`) — `grep -rn draining_listener crates/` across the whole workspace returns nothing: it had zero callers anywhere, including in tests. `run()` builds its `Arc<dyn DeliveryProcessor>` inline and independently (unchanged in this diff), so no code path or test coverage was reachable only through the deleted helper. Confirmed dead-code removal, not a coverage cut.

4. **The lint is green for the right reason, re-run myself, not trusted from the author's paste.** `bash scripts/check_scan_floor_selftest.sh test_pointers` → `scan-floor self-test: 1 gate(s) correctly refuse a vacuous scan — OK` (exit 0). `bash scripts/check_test_pointers.sh` → `test-pointers: resolved 23084 Test: citation(s) (floor 200) — 0 dangling pointers — OK` (exit 0), matching the author's reported count exactly.

5. **No gate was greened by deletion.** `git diff fe05c964..65ba7b074` shows the only removed `Test:` lines are the two stale citations being corrected in place (one replaced with the real name on an adjacent `+` line, one removed along with the dead function it documented) — not a citation dropped to silence the lint with no replacement. No new `#[ignore]`, `cfg`-gate, or `--exclude` anywhere in the 3-file diff.

**Own gate re-run on this exact head:** `trusty-common` webhook_relay 80/80 (the +1 over round 2 is the new sidecar test), `trusty-review` webhook_drain 10/10, `trusty-analyze` webhook_drain 11/11, `trusty-common`/`trusty-analyze` clippy `-D warnings` clean. All `EXIT=0`.

**CI rollup, read fresh via the type-aware query** (`.status // "-", .conclusion // .state`) rather than trusting the author's queued-state paste: `Test: doc-comment pointer lint` → `COMPLETED / SUCCESS`. `Per-PR changelog fragment` → `COMPLETED / SUCCESS`. Both confirmed green on `65ba7b074`, not stale from an earlier head. One row, `Test`, is still `IN_PROGRESS` at time of this read — the workspace-wide CI test job, distinct from the scoped tests I ran myself above, which are the substantive coverage for the changed code and are green. Per this review's own no-CI-blocking discipline, I'm not waiting on it; the PM should take one more one-shot read before merge to confirm it lands green.

## Confidence notes

Everything above is held at >80% confidence — each claim was independently re-derived (revert-and-fail for the new test, direct reading of the two re-pointed test bodies, a workspace-wide grep for dead-code callers, and re-running rather than trusting both the lint and the CI rollup) rather than accepted from the author's report.

PR #5254, opened 2026-08-09
fix(website): stop /whats-new scrolling horizontally on mobile

bobmatnyc:
## Defect

`/whats-new` at a 375px viewport scrolled sideways: `document.documentElement.scrollWidth` measured 471px against a 375px `clientWidth`. Cause: inline `<code>` spans in changelog prose (e.g. `pipeline::citation_check::downgrade_uncitable_findings`, `TRUSTY_REVIEW_SOLO_MEDIUM_ESCALATION_CONFIDENCE`) have no space or hyphen to wrap at. The page shipped in #5248 (squash `b52b5fc9`).

## Fix

- `.doc-prose :not(pre) > code` (`website/src/app.css`) is now `display: inline-block; max-width: 100%; overflow-x: auto`. A code span that contains spaces still wraps normally inside its own box (no scrollbar). A space-free identifier that has nowhere to wrap instead scrolls within its own box — the same "wide content scrolls in its own container" contract `.doc-table` and `.doc-prose pre` already use, so the identifier stays fully legible instead of breaking mid-character.
- That alone still left the page 1px over at 320px — not any single element's doing. `main` carries `overflow-x: hidden` as the backstop. Scoped to `main`, not `html`/`body`: those sit in `<SiteHeader>`'s `position: sticky` ancestor chain, and giving either a non-`visible` overflow silently demotes the sticky header to plain in-flow positioning. `main` is `SiteHeader`'s sibling, not its ancestor — verified the header still sticks after this change.

## The backstop needed its own proof (review round 2)

`main{overflow-x:hidden}` clips at `main`'s own boundary, so `document.documentElement.scrollWidth` alone cannot see anything that overflows `main` and gets clipped there — it would read clean while real content rendered cut off. Proved this both ways:

1. **Injected a long unbroken run of plain prose** (no `<code>` wrapper, so none of the inline-code fix's own containment applies) into a changelog item. With the ORIGINAL test (document-level only): `document.documentElement.scrollWidth` stayed at `clientWidth` — the test passed — while a screenshot showed the injected text visibly sliced off mid-run, and `main.scrollWidth` read 1697 against a 375 `clientWidth`. The document-level assertion alone is decorative against this failure mode.
2. **Fixed the measurement**, not the CSS: the test now also asserts `main.scrollWidth <= main.clientWidth`, which measures inside the clipping boundary. Re-ran the same injection — now fails hard: `<main> scrollWidth 1612 > clientWidth 375 (tolerance 1px)`. Reverted the injection (`git checkout -- crates/trusty-review/CHANGELOG.md`, confirmed clean via `git status`); the suite is back to green.

That stricter measurement surfaced two things the document-level check had been hiding:

- **`/whats-new` still sits 1px over `main` at 320px on the clean build — cosmetic.** An exhaustive per-element and per-text-node audit (every `<code>` span, every text node, `getBoundingClientRect`/`getClientRects` by hand) found nothing rendered past the viewport edge; it took removing whole changelog sections to move the number, the signature of sub-pixel layout rounding accumulated across ~225 code spans on that page, not a real spatial overflow. Kept as a named, 1px-only `MAIN_TOLERANCE_PX` test constant — not a wider `overflow-x: hidden`, and two orders of magnitude below anything a real clipped-content regression produces (1612–1697px in the injection proof above).
- **`/` (homepage) at 320px overflows `main` by ~37px — real, and out of scope.** A tool card in the grid renders with its right edge past the viewport; unrelated to changelog prose, and confirmed present on `origin/main` before this PR (re-ran the measurement against the unmodified `app.css`). Filed as #5255 and excluded from this test's route×width matrix by name (`isKnownPreexistingGap`), not silently tolerated — every other route/width combination, including `/` at 375px, is still asserted.

## Evidence

Measured against the production build (`pnpm build` + a static server), before and after:

| Route | Width | Before | After |
|---|---|---|---|
| `/whats-new` | 375px | scrollWidth 471 / clientWidth 375 | scrollWidth 375 / clientWidth 375 |
| `/whats-new` | 320px | scrollWidth 471 / clientWidth 320 | scrollWidth 320 / clientWidth 320 |
| `/` (homepage) | 375px | no overflow | no overflow |
| `/docs/getting-started/install` | 375px | no overflow | no overflow; `.doc-prose pre` and `.doc-table` still scroll internally (contained) |
| `/tools/trusty-search` | 320/375px | no overflow | no overflow |

Sticky header: `getBoundingClientRect().top` stays `0` after `scrollTo(0, 2000)` on `/whats-new`, before and after.

Playwright/CI footprint: `playwright` (no browsers bundled) has no `postinstall` script — confirmed by inspecting its `package.json` and by a fresh `npm install playwright@1.62.1` into an empty temp project (759ms, no browser download, no cache activity). A `pnpm install` on a fresh clone pulls no browser; only actually *running* `tests/mobile-overflow.test.ts` needs one, and it resolves to whatever's already cached locally (this box already has one from `crates/trusty-agents/ui`'s `@playwright/test`) or fails with Playwright's own actionable "run `npx playwright install`" error — it does not download silently. Website tests don't run in CI (#5200), so this never touches CI at all.

`fileParallelism: false` wall-clock: the whole `smoke` project (`build-smoke.test.ts` + `mobile-overflow.test.ts`, now serialized so they don't race on the same `.vercel/output`) — `Duration 14.14s (... tests 13.81s)`. Comfortably inside the 300s timeout.

## Test plan

- [x] `pnpm install`
- [x] `pnpm check` — 0 errors, 0 warnings
- [x] `pnpm lint` — pass
- [x] `pnpm test` — 229 passed, 1 skipped (documented, #5255) / 230, across 14 files
- [x] `pnpm build` — clean

Website tests don't run in CI (#5200); the counts above are this session's local run. No changelog fragment — `website/` only, nothing under `crates/` changed (the injection used to prove the test's teeth was reverted via `git checkout`, confirmed clean).

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -30,6 +30,7 @@
 		"globals": "^15.14.0",
 		"hast-util-to-string": "^3.0.1",
 		"jsdom": "^25.0.1",
+		"playwright": "^1.62.1",
 		"postcss": "^8.4.47",
 		"prettier": "^3.4.2",
 		"prettier-plugin-svelte": "^3.3.3",
--- a/website/pnpm-lock.yaml
+++ b/website/pnpm-lock.yaml
@@ -54,6 +54,9 @@ importers:
       jsdom:
         specifier: ^25.0.1
         version: 25.0.1
+      playwright:
+        specifier: ^1.62.1
+        version: 1.62.1
       postcss:
         specifier: ^8.4.47
         version: 8.5.26
@@ -1241,6 +1244,11 @@ packages:
   fraction.js@5.3.4:
     resolution: {integrity: sha512-1X1NTtiJphryn/uLQz3whtY6jK3fTqoE3ohKs0tT+Ujr1W59oopxmoEh7Lu5p6vBaPbgoM0bzveAW4Qi5RyWDQ==}
 
+  fsevents@2.3.2:
+    resolution: {integrity: sha512-xiqMQR4xAeHTuB9uWm+fFRcIOgKBMiOBP+eXiyT7jsgVCq1bkVygt00oASowB7EdtpOHaaPgKt812P9ab+DDKA==}
+    engines: {node: ^8.16.0 || ^10.6.0 || >=11.0.0}
+    os: [darwin]
+
   fsevents@2.3.3:
     resolution: {integrity: sha512-5xoDfX+fL7faATnagmWPpbFtwh/R77WmMMqqHGS65C3vvB0YHrgF+B1YmZ3441tMj5n63k0212XNoJwzlhffQw==}
     engines: {node: ^8.16.0 || ^10.6.0 || >=11.0.0}
@@ -1759,6 +1767,16 @@ packages:
   pirates@4.0.7:
     resolution: {integrity: sha512-TfySrs/5nm8fQJDcBDuUng3VOUKsd7S+zqvbOTiGXHfxX4wK31ard+hoNuvkicM/2YFzlpDgABOevKSsB4G/FA==}
     engines: {node: '>= 6'}
+
+  playwright-core@1.62.1:
+    resolution: {integrity: sha512-wPYSwEBJY9GHraISXqyqtx0na0LpO3XEX7jNDhntbex7tzUS7kLnZsOlFruFJB4Hi/rhDMjXGqHewDZ68nYZVw==}
+    engines: {node: '>=20'}
+    hasBin: true
+
+  playwright@1.62.1:
+    resolution: {integrity: sha512-0M+L3LAD8/nm554LOla9Ayx0j0tmFZ0FBcoQ7F1VuVHpM/XpiC8RcDzBQB8W5+hA8L22THxELzeF+2WcUzvcLg==}
+    engines: {node: '>=20'}
+    hasBin: true
 
   postcss-import@15.1.0:
     resolution: {integrity: sha512-hpr+J05B2FVYUAXHeK1YyI267J/dDDhMU6B6civm8hSY1jYJnBXxzKDKDswzJmtLHryrjhnDjqqp/49t8FALew==}
@@ -3384,6 +3402,9 @@ snapshots:
 
   fraction.js@5.3.4: {}
 
+  fsevents@2.3.2:
+    optional: true
+
   fsevents@2.3.3:
     optional: true
 
@@ -4092,6 +4113,14 @@ snapshots:
   pify@2.3.0: {}
 
   pirates@4.0.7: {}
+
+  playwright-core@1.62.1: {}
+
+  playwright@1.62.1:
+    dependencies:
+      playwright-core: 1.62.1
+    optionalDependencies:
+      fsevents: 2.3.2
 
   postcss-import@15.1.0(postcss@8.5.26):
     dependencies:
--- a/website/src/app.css
+++ b/website/src/app.css
@@ -153,6 +153,27 @@
 		@apply bg-foundry-bg font-sans text-foundry-text antialiased;
 	}
 
+	/* Why: every wide-content case below is capped to its own scroll
+	   container (`.doc-table`, `.doc-prose pre`, `.doc-prose :not(pre) >
+	   code`), but a changelog section this size (225 inline `<code>` spans on
+	   trusty-review alone) can still land the PAGE a sub-pixel past its
+	   viewport on some routes — not any one element's doing (none of them
+	   measures past the viewport edge in DevTools; it takes removing whole
+	   sections of content to clear it, which is the signature of accumulated
+	   layout rounding across that many boxes, not a visible overflow).
+	   `overflow-x: hidden` on `#main` (`+layout.svelte`) is the backstop for
+	   exactly that residue (whats-new mobile overflow) — never a substitute
+	   for containing wide content locally, which is what actually keeps this
+	   page's content legible instead of clipped. Scoped to `main` rather than
+	   `html`/`body`: those sit in `<SiteHeader>`'s sticky-positioning ancestor
+	   chain, and giving either a non-`visible` overflow demotes `position:
+	   sticky` on the header to plain in-flow positioning (its containing
+	   block becomes that non-scrolling box instead of the viewport). `main`
+	   is `SiteHeader`'s sibling, not its ancestor, so it carries no such risk. */
+	main {
+		overflow-x: hidden;
+	}
+
 	/* Foundry: "machines talk mono" — ids, paths, counts, versions. */
 	code,
 	pre,
@@ -282,9 +303,21 @@
 
 	/* Inline code sets NO colour, deliberately: inside a link it must stay
 	   link-coloured (5.47:1 on this ground), and in body copy it inherits
-	   text-primary (16.18:1). Setting a colour here would break the first. */
+	   text-primary (16.18:1). Setting a colour here would break the first.
+	   Why: an identifier like `pipeline::citation_check::downgrade_uncitable_findings`
+	   has no space or hyphen for the browser to wrap at, so on a narrow
+	   viewport it pushed the PAGE itself into horizontal scroll (whats-new
+	   mobile overflow). `display: inline-block` + `max-width: 100%` lets a
+	   code span that DOES contain spaces still wrap normally onto several
+	   lines inside its own box — no scrollbar, no overflow, same as before.
+	   A span with no break opportunity at all (the identifiers above) instead
+	   overflows that now-constrained box, so `overflow-x: auto` scrolls the
+	   SPAN, never the page — one more wide-content-scrolls-in-its-own-container
+	   contract, alongside `.doc-table` and `.doc-prose pre` below. Breaking the
+	   token mid-character to avoid that scrollbar would be worse: it stops
+	   reading as one identifier. */
 	.doc-prose :not(pre) > code {
-		@apply rounded-sm border border-foundry-border bg-foundry-card px-1 py-px text-[0.9em];
+		@apply inline-block max-w-full overflow-x-auto rounded-sm border border-foundry-border bg-foundry-card px-1 py-px align-bottom text-[0.9em];
 	}
 
 	.doc-prose pre {
--- a/website/src/app.css
+++ b/website/src/app.css
@@ -160,18 +160,36 @@
 	   viewport on some routes — not any one element's doing (none of them
 	   measures past the viewport edge in DevTools; it takes removing whole
 	   sections of content to clear it, which is the signature of accumulated
-	   layout rounding across that many boxes, not a visible overflow).
-	   `overflow-x: hidden` on `#main` (`+layout.svelte`) is the backstop for
-	   exactly that residue (whats-new mobile overflow) — never a substitute
-	   for containing wide content locally, which is what actually keeps this
-	   page's content legible instead of clipped. Scoped to `main` rather than
-	   `html`/`body`: those sit in `<SiteHeader>`'s sticky-positioning ancestor
-	   chain, and giving either a non-`visible` overflow demotes `position:
-	   sticky` on the header to plain in-flow positioning (its containing
-	   block becomes that non-scrolling box instead of the viewport). `main`
-	   is `SiteHeader`'s sibling, not its ancestor, so it carries no such risk. */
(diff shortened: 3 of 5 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial re-review at `6de64ec8`, worktree `.claude/worktrees/critic-5254`. All commands run locally from inside `website/` (website tests do not run in CI, #5200).

### Independent verification

- `pnpm install`: ~1.3s, no browser download — matches the PR's claim (browser already cached locally, `playwright`'s `package.json` has no `scripts`/`postinstall` key).
- `pnpm check`: `0 ERRORS 0 WARNINGS`.
- `pnpm lint`: `prettier --check .` clean, `eslint .` clean.
- `pnpm test`: **229 passed | 1 skipped (230)** across 14 files — matches the PR body exactly. The skip is `no page scrolls horizontally on mobile > / at 320px`, matching `isKnownPreexistingGap` verbatim.
- `pnpm build`: clean.
- Reproduced the author's injection proof independently: added a long unbroken run of plain prose (no `<code>` wrapper) to `crates/trusty-review/CHANGELOG.md`, rebuilt, reran `mobile-overflow.test.ts` — it fails hard (`<main> scrollWidth 2832 > clientWidth 375/320`). Reverted with `git checkout --`, confirmed `git status` clean. The test has teeth.
- Reproduced the `/` @ 320px pre-existing gap against `origin/main`'s unmodified `app.css` (before this PR touched anything): `document.documentElement.scrollWidth` reads 357 vs a 320 `clientWidth` — same ~37px, same root cause, present before this diff. The exclusion is real and correctly scoped to that one route/width.

### Findings

| HIGH | `website/src/app.css` | 173-175 | `main { overflow-x: hidden; }` forces the browser to compute `overflow-y: auto` on `main` (CSS Overflow Module's "one axis visible, the other not → the visible one becomes auto" rule — confirmed via `getComputedStyle`: `overflow-y` is `visible` on `origin/main`, `auto` after this diff). That makes `main` a scroll container for `position: sticky` purposes, and both docs sidebars — the left nav (`website/src/routes/docs/+layout.svelte:35`) and the in-page TOC (`website/src/routes/docs/[...slug]/+page.svelte:61`), both `sticky top-20`, both descendants of `main` — stop sticking. Measured at 1280×800 on `/docs/getting-started/install`: pre-PR, `rectTop` stays `80` after scrolling 800px (correctly pinned); post-PR, `rectTop` goes to `-667` — the sidebar now scrolls away with the page like static content. These sidebars are `hidden … lg:block`/`xl:block`, so this is desktop-only and the PR's own test suite (320/375px only) cannot see it — nor can it be caught anywhere else, since website tests don't run in CI. The PR's own safety claim ("main is SiteHeader's sibling, not its ancestor, so it carries no such risk") is true for the header and false for these two sidebars, which are inside main. | Change `overflow-x: hidden` to `overflow-x: clip` on `main`. Verified: `overflow-y` then computes to `visible` (not forced), both sidebars stick correctly again (`rectTop` back to 80 after the same 800px scroll), and the original `/whats-new` fix is unaffected (`main.scrollWidth` 321 vs `clientWidth` 320, `document.scrollWidth` 320/320 — identical to the `hidden` numbers). | Fix here |
| MEDIUM | `website/tests/mobile-overflow.test.ts` | 340-349 | The two-tier assertion (`document` + `main`) can be defeated by a third clipping ancestor between `main` and content. Constructed: `main{overflow-x:hidden} > div{overflow:hidden;width:100px} > span{white-space:nowrap}` with a wide span — `document.scrollWidth`, `main.scrollWidth`, and `main.clientWidth` all read exactly `320` (pass), while the span's real `getBoundingClientRect().right` is `1918` in a 320px viewport: content genuinely clipped, invisible to both assertions. No such intermediate `overflow:hidden` exists anywhere in `src/` today (grepped — only `main` itself sets non-visible overflow), so this isn't a live bug, but it's the same "decorative test" shape this PR's own second review round already found and fixed once, one level shallower, and nothing in CI would catch a recurrence one level deeper (website tests don't run in CI). | Not required for this fix. A follow-up that walks every element under `main` (or does per-route screenshot diffing) instead of trusting two fixed measurement points would close this generally. | Promote |

### Notes

- Issue #5255 is real, correctly scoped to `/` @ 320px only, and I independently confirmed it's pre-existing and unrelated to this diff (see verification above). One inaccuracy in the issue's own text, unrelated to this PR: it states `main.scrollWidth reads ~1530-1560px`; my measurement, on both `origin/main` and this PR's head, reads exactly `357`. The `~37px` figure elsewhere in the same issue body is correct. Doesn't affect this PR — the issue text itself could use a correction.
- `MAIN_TOLERANCE_PX = 1` is genuine sub-pixel rounding, not a masked bug: `/whats-new` @ 320px measures `main.scrollWidth` 321 vs `clientWidth` 320 (exactly 1px), and the injection proof above shows a real regression reads ~2800, three orders of magnitude past the tolerance.
- `SiteHeader`'s own sticky-ness (a sibling of `main`, not affected by the HIGH finding above) is fine — confirmed `getBoundingClientRect().top` stays `0` after scrolling 2000px, as claimed.
- Fenced code blocks confirmed contained in their own `overflow-x: auto` box (pre-existing, unchanged by this diff) — a wide `<pre><code>` block's content extends past the viewport in its own bounding rect without moving `main.scrollWidth`.
- `/docs/**` and `/tools/*` screenshots at 375px show no visual regression from the `.doc-prose :not(pre) > code` change — long identifiers stay intact (not broken mid-character), scroll within their own box, `align-bottom` keeps the surrounding line's baseline correct.
- `fileParallelism: false` is correctly scoped to the `smoke` Vitest project only; the `unit` project still runs in parallel.
- Working tree left clean; confirmed nothing under `crates/` is part of this diff.

### Required changes (before merge)

1. `website/src/app.css:174` — `overflow-x: hidden` → `overflow-x: clip` on `main`, to stop breaking `position: sticky` for the docs sidebars while keeping the `/whats-new` fix intact (verified above).
2. Add a regression assertion (or extend `mobile-overflow.test.ts`) that the docs `nav`/`aside` (`sticky top-20`) stays pinned near `top: 80px` after scrolling, at a desktop viewport (these are `hidden` below `lg`/`xl`, so the existing 320/375px matrix can't cover them) — the same pattern the PR already applies to the header, just for main's own sticky descendants.

bobmatnyc:
## Addressed code-critic WARN (HIGH + MEDIUM)

**HIGH — sticky sidebars, fixed.** \`overflow-x: hidden\` on \`main\` left \`overflow-y\` at its default \`visible\`; the CSS Overflow spec forces a mismatched visible axis to \`auto\`, promoting \`main\` into a scroll container and breaking every \`position: sticky\` element inside it (the \`/docs\` sidebar rail, the "On this page" aside). Confirmed the critic's measurement exactly: sidebar \`top\` went from a stable 80px to -667px after scrolling 800px.

Switched to \`overflow: clip\` (both axes explicitly, not just \`overflow-x\`) — \`clip\` provides no scrolling mechanism, so a UA must not turn the box into a scroll container for it. Verified:
- \`getComputedStyle(main).overflowY\` is now \`clip\`, not \`auto\`
- both docs sidebars hold at \`top: 80px\` through an 800px scroll at desktop width
- the site header (\`main\`'s sibling, not descendant) is unaffected — still sticks
- \`main.scrollWidth\` still reports true content width under \`clip\` exactly as it did under \`hidden\`, so the \`/whats-new\` backstop measurement is unaffected
- \`/whats-new\` and \`/tools/*\` stay clean at 320/375px

Added a desktop-width regression test (\`keeps the /docs sidebar sticky at desktop width\`) and proved it has teeth the same way the \`main\`-vs-\`document\` measurement was proved in the prior round: reverted \`app.css\` to \`overflow-x: hidden\` and reran — fails (\`top drifted 133 -> -667\`) — then restored \`clip\` and reran clean.

**MEDIUM — documented, not solved speculatively.** No third clipping ancestor between \`main\` and content exists in this codebase today, so no synthetic one is tested against. Left a plain comment in \`tests/mobile-overflow.test.ts\` naming the limitation for whoever adds one next.

## Gates (inside \`website/\`)

\`\`\`
pnpm install  -> Already up to date. Done in 353ms.
pnpm check    -> COMPLETED 487 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS
pnpm lint     -> All matched files use Prettier code style! eslint clean.
pnpm test     -> Test Files 14 passed (14); Tests 230 passed | 1 skipped (231)
pnpm build    -> built in 3.09s
\`\`\`

The 1 skip is the pre-existing \`/\` @ 320px gap (#5255), unchanged from the prior round.

bobmatnyc:
## Round 2 verdict: APPROVE

Re-review at `d815d177`, worktree `.claude/worktrees/critic-5254-r2`. Scope: the delta from `6de64ec8` is exactly `website/src/app.css` and `website/tests/mobile-overflow.test.ts` (`git diff --stat 6de64ec8..d815d177`) — nothing else moved.

### Independent verification

- `pnpm install`: clean, no browser download.
- `pnpm check`: `0 ERRORS 0 WARNINGS`.
- `pnpm lint`: clean.
- `pnpm test`: **230 passed | 1 skipped (231)** — matches the PR's report exactly. The skip is still `no page scrolls horizontally on mobile > / at 320px`, i.e. `isKnownPreexistingGap`, tied to #5255. No new skip.
- `pnpm build`: clean.

### Round-1 HIGH — verified fixed

`main { overflow: clip }` (both axes explicit) replaces `overflow-x: hidden`.

- `getComputedStyle(main)`: `{ overflowX: 'clip', overflowY: 'clip' }` — neither axis computes to `visible`, so the CSS Overflow spec's other-axis-forced-to-`auto` promotion that caused the original break does not fire. Confirmed directly, not inferred.
- `main.scrollWidth` still reports true, unclipped content width under `clip`: injected a synthetic 2000-character span into `main` on `/whats-new` at 320px — `scrollWidth` went from 321 to 19616. The backstop measurement my round-1 MEDIUM was about is intact; `clip` did not quietly change what `scrollWidth` reports.
- Sticky restored at every width where the sidebars actually render, not just the PR's one test point (1280px): checked 1024, 1100, 1280, 1440. At 1024/1100 (`lg`, nav visible, aside still `hidden` below `xl`) the nav holds at `top: 80` before and after scrolling 800px. At 1280/1440 (`xl`, both visible) both nav and aside settle to `top: 80` after scroll. No drift at any of the four widths.
- Reran their revert-and-fail proof myself: swapped `overflow: clip` back to `overflow-x: hidden`, ran `keeps the /docs sidebar sticky at desktop width` alone — fails, `sidebar top drifted from 133 to -667 after scrolling — main became a scroll container: expected -667 to be 133`. Restored the fix, confirmed `git status` clean.
- Page-scroll mechanics: `main.scrollTop` stays `0` through a 1000px wheel-scroll while `window.scrollY` advances — `main`'s height has always equaled its content height (flex-1 / min-h-screen, no independent height cap), so it was never an internal scroll container before or after; nothing depended on `main` scrolling.
- Anchor navigation: `/whats-new#trusty-search` scrolls the target into view correctly (`getBoundingClientRect().top` in-viewport after navigation).
- `SiteHeader`'s own sticky-ness (sibling of `main`) unaffected — not re-broken by the two-axis `clip`.

### Round-1 MEDIUM — documented, accepted

No machinery was added for a third clipping ancestor between `main` and content, because none exists in the codebase today (still true — reconfirmed no other non-`visible` overflow in `src/`). Instead the limitation is stated directly in `tests/mobile-overflow.test.ts`'s file-level doc comment, right next to the explanation of the two-tier measurement it qualifies: it names exactly what the two assertions do and don't cover, says why no synthetic ancestor was added, and tells a future author what pattern to follow if one is ever introduced. That's accurate and it's in the place a reader auditing this measurement would already be looking — building speculative test machinery for a case that doesn't exist would have been the wrong call. Treating this as resolved, not reopening it.

### Findings

No issues found at >80% confidence in this round. APPROVED for next pipeline stage.

### Notes

- All round-1 findings (HIGH: sticky-breaking `overflow-x: hidden`; MEDIUM: third-clipping-ancestor coverage gap) are closed as above.
- Everything else from round 1 (MAIN_TOLERANCE_PX genuineness, `/` @ 320px pre-existing-gap scoping, `.doc-prose :not(pre) > code` fix, fenced-block containment, playwright footprint, clean working tree) is unchanged by this delta and was already verified.

Issue #5255, opened 2026-08-09
website: homepage tool-card grid overflows horizontally at 320px


bobmatnyc:
## Defect

The homepage (\`/\`) tool-card grid overflows its container at a 320px viewport. Measured against the production build: \`main.scrollWidth\` reads ~1530-1560px against a 320px \`main.clientWidth\` — roughly 37px past the card grid's own right edge, not sub-pixel rounding.

A single \`article.card\` (e.g. the \`trusty-search\` "Hybrid code search" tile) renders with \`getBoundingClientRect().right\` around 357px against a 320px viewport, so part of the card is pushed off-screen and clipped by \`main\`'s \`overflow-x: hidden\` (added in #5254) rather than wrapping into the grid.

## Scope

Pre-existing and unrelated to #5254 (the \`/whats-new\` inline-code mobile overflow fix) — reproduced against \`origin/main\` before that PR's changes. \`/whats-new\` and other routes are clean at 320px; only the homepage's card grid is affected. Likely a \`grid-template-columns\`/\`minmax()\` floor that doesn't shrink enough at 320px — not yet root-caused.

## Repro

\`\`\`
cd website && pnpm build && pnpm preview --port 4173
# then, in a real browser or Playwright, at a 320px viewport:
document.querySelector('main').scrollWidth   // ~1530-1560
document.querySelector('main').clientWidth   // 320
\`\`\`

Found while proving out the regression test in #5254; \`website/tests/mobile-overflow.test.ts\` deliberately skips \`/\` at 320px and documents why, pointing at this issue.

bobmatnyc:
New data point at a wider viewport: `/ at 375px` now fails too (not just 320px):

AssertionError: / at 375px: <main> scrollWidth 433 > clientWidth 375 (tolerance 1px) — content overflows main and renders clipped: expected 433 to be less than or equal to 376
 ❯ tests/mobile-overflow.test.ts:220:9

Reproduce: `cd website && pnpm install && pnpm test -- mobile-overflow`

PR #5254's evidence table recorded `/` at 375px as clean, and `isKnownPreexistingGap` in mobile-overflow.test.ts only skips `/` at 320px — 375px is asserted and currently red. `website/src/lib/tools.ts` and `app.css`'s homepage/grid rules are byte-identical since #5254 (`git diff 80c47a884..origin/main` is empty for both), so the cause isn't a content or CSS change on this branch — root cause still open.

[...]

PR #5257, opened 2026-08-09
fix(website): scheme allowlist for docs/changelog links

### LINKED ISSUE #5256: fix(website): scheme allowlist for docs/changelog links — javascript:/file: URLs pass through unrewritten
## Defect

`website/src/lib/docs/links.ts`'s `ABSOLUTE` regex (`/^(?:[a-z][a-z0-9+.-]*:|\/\/)/i`) matches any URL scheme and passes it through unrewritten. `website/src/lib/changelog/parse.ts` carries an identical copy for the same purpose. Both the docs reader (`/docs/**`) and the changelog reader (`/whats-new`) rely on it.

A `[click me](javascript:alert(1))` or a `file:///etc/passwd` link committed to any published `docs/` page or flagship `CHANGELOG.md` renders as a live `href` on `https://trusty-tools.vercel.app`.

**Threat model, stated honestly:** exploiting this requires commit access to a published doc or changelog — it is not a path from an untrusted visitor to the site. This is a guard against a mistake or a compromised commit, not a live vulnerability.

**Why it's still worth closing:** the build already fails on an unresolvable *relative* link. A mistyped relative path stops the build; a `javascript:` URL sailed through — that asymmetry was the actual defect. Found by code-critic while reviewing #5248; correctly left out of that PR as pre-existing and unrelated.

## Evidence

Red (defect present) — 6 of the added tests fail:

```
❯ |unit| src/lib/docs/links.test.ts (24 tests | 4 failed)
    × fails the build on a javascript: link
    × fails the build on a file:// link
    × fails the build on mailto:, which this site does not allowlist
    × fails the build on data: and vbscript: as well as any unrecognised scheme
❯ |unit| src/lib/changelog/parse.test.ts (32 tests | 2 failed)
    × fails the build on a javascript: link, the same as any other bad link
    × fails the build on a file:// link
 Test Files  2 failed (2)
      Tests  6 failed | 50 passed (56)
```

Green (fix applied) — same two files, all passing:

```
 Test Files  2 passed (2)
      Tests  56 passed (56)
```

The real corpus (27 published `docs/` pages, all 6 flagship `CHANGELOG.md` files) carries no non-`http(s)` scheme link today — confirmed by grep before the fix and by a clean `pnpm build` after it.

## Resolution

Replaced the scheme-matching regex with an explicit allowlist (`http:`, `https:`) in one shared function, `classifyAbsoluteHref` in `website/src/lib/docs/links.ts`. The changelog reader (`website/src/lib/changelog/parse.ts`) now imports and calls the same function instead of carrying its own copy of the scheme check.

- Docs reader: a disallowed scheme fails the build with `UNSAFE-SCHEME`.
- Changelog reader: a disallowed scheme fails the build with `CHANGELOG-BAD-LINK` (its existing catch-all for link problems).
- `https://`, `http://`, and protocol-relative (`//host/…`) links are unchanged.
- Anchor-only and site-relative link handling is unchanged (verified before and after).

bobmatnyc:
Closes #5256

## Defect

`website/src/lib/docs/links.ts`'s `ABSOLUTE` regex matched any URL scheme and passed it through unrewritten, so `[click me](javascript:alert(1))` or `file:///etc/passwd` committed to a published doc or flagship CHANGELOG would render as a live href on the site. `website/src/lib/changelog/parse.ts` carried an identical copy of the same regex.

Threat model, stated honestly: this needs commit access to a published doc or changelog — not a path from an untrusted visitor. It's a guard against a mistake or compromised commit, not a live vulnerability. Worth closing because the build already fails on an unresolvable relative link; a `javascript:` URL sailing through unchecked was the actual asymmetry.

## Evidence

Red (defect present, fix reverted, new tests only):

```
❯ |unit| src/lib/docs/links.test.ts (24 tests | 4 failed)
    × fails the build on a javascript: link
    × fails the build on a file:// link
    × fails the build on mailto:, which this site does not allowlist
    × fails the build on data: and vbscript: as well as any unrecognised scheme
❯ |unit| src/lib/changelog/parse.test.ts (32 tests | 2 failed)
    × fails the build on a javascript: link, the same as any other bad link
    × fails the build on a file:// link
 Test Files  2 failed (2)
      Tests  6 failed | 50 passed (56)
```

Green (fix restored):

```
 Test Files  2 passed (2)
      Tests  56 passed (56)
```

Full gate chain, from inside `website/`:

```
pnpm check → 487 FILES 0 ERRORS 0 WARNINGS
pnpm lint  → prettier + eslint clean
pnpm test  → Test Files 14 passed (14), Tests 239 passed | 1 skipped (240)
pnpm build → succeeds (real corpus has no non-http(s) scheme link today)
```

Website tests don't run in CI (#5200) — the above is local output, not a CI artifact.

## Resolution

- Added `classifyAbsoluteHref` to `website/src/lib/docs/links.ts`: allowlists `http:`/`https:` (and protocol-relative `//host/…`, which carries no scheme of its own); everything else is `unsafe-scheme`.
- `resolveLink` now fails the build with `UNSAFE-SCHEME` on a disallowed scheme.
- `website/src/lib/changelog/parse.ts` dropped its duplicate `ABSOLUTE` regex and imports the same `classifyAbsoluteHref`, failing with `CHANGELOG-BAD-LINK` (its existing catch-all) on a disallowed scheme — one implementation, both readers.
- Anchor-only, site-relative, and protocol-relative link handling is unchanged in both readers (confirmed before/after).
- Updated `website/src/lib/docs/README.md`'s link-rule and failure-code tables.
- Nothing under `crates/` changed — no changelog fragment needed.

### DIFF
--- a/website/src/lib/changelog/parse.test.ts
+++ b/website/src/lib/changelog/parse.test.ts
@@ -291,6 +291,26 @@ describe('links and metavariables inside an item', () => {
 		);
 	});
 
+	it('leaves an http:// URL alone', () => {
+		const { releases, failures } = withProbe(item('see [x](http://example.com/a)'), []);
+		expect(failures).toEqual([]);
+		expect(releases[0].categories[0].items[0].html).toContain('href="http://example.com/a"');
+	});
+
+	it('fails the build on a javascript: link, the same as any other bad link', () => {
+		const { failures } = withProbe(item('[click me](javascript:alert(1))'), []);
+		expect(failures).toHaveLength(1);
+		expect(failures[0].code).toBe('CHANGELOG-BAD-LINK');
+		expect(failures[0].problem).toContain('javascript:');
+	});
+
+	it('fails the build on a file:// link', () => {
+		const { failures } = withProbe(item('[passwd](file:///etc/passwd)'), []);
+		expect(failures).toHaveLength(1);
+		expect(failures[0].code).toBe('CHANGELOG-BAD-LINK');
+		expect(failures[0].problem).toContain('file:');
+	});
+
 	// `<path>` outside backticks parses as an unknown element and renders as
 	// NOTHING. Eight of these sit in the corpus, and a changelog is history
 	// this site must not edit — so the text is recovered, never dropped.
--- a/website/src/lib/changelog/parse.ts
+++ b/website/src/lib/changelog/parse.ts
@@ -46,6 +46,7 @@ import { toString as hastToString } from 'hast-util-to-string';
 import { visit } from 'unist-util-visit';
 import type { Element, Root, RootContent } from 'hast';
 
+import { classifyAbsoluteHref } from '../docs/links';
 import { ALLOWED_ELEMENTS, parsePage, stringifyHast } from '../docs/render';
 import { GITHUB_REPO } from '../docs/repo';
 import { CHANGELOG_CODES, type ChangelogFailure } from './errors';
@@ -127,8 +128,6 @@ const RELEASE_HEADING = /^\[([^\]]+)\](.*)$/s;
 /** A separator between the version and the rest: em dash, en dash, or hyphen. */
 const LEADING_SEPARATOR = /^[\s—–·-]+/;
 const TRAILING_SEPARATOR = /[\s—–·-]+$/;
-/** Anything with a scheme, or a protocol-relative URL, already leaves the repo. */
-const ABSOLUTE = /^(?:[a-z][a-z0-9+.-]*:|\/\/)/i;
 
 /**
  * Why: see THE LINK-REFERENCE TRAP in the module header.
@@ -301,7 +300,7 @@ export function parseChangelog(
 }
 
 /**
- * Why: three problems in changelog prose would each ship a defect quietly.
+ * Why: four problems in changelog prose would each ship a defect quietly.
  *
  *   1. A RELATIVE link (`[spec](../../docs/specs/foo.md)`) is written to be
  *      followed on GitHub. Rendered verbatim on `/whats-new` it resolves
@@ -313,6 +312,10 @@ export function parseChangelog(
  *      renders as NOTHING — `run trusty-search index <path>` would publish as
  *      `run trusty-search index`. Eight of these are in the corpus right now.
  *   3. An image would be a third-party request from a site that makes none.
+ *   4. An absolute link using a scheme this site refuses to publish
+ *      (`javascript:`, `file:`, …) would render as a live href. Checked via
+ *      `classifyAbsoluteHref` — the same allowlist the doc reader applies, so
+ *      the two never drift (`../docs/links.ts`).
  *
  * What: rewrites every relative link into a `blob/main` link on the crate's
  * own path, and fails the build on anything it cannot resolve. Mutates the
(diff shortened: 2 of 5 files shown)


bobmatnyc:
## Verdict: APPROVE

Reviewed at `ad125cf9` in a detached worktree. Went after the scheme allowlist directly — case variation, whitespace, embedded control characters, percent-encoding, unicode lookalikes, zero-width space, nested angle brackets, reference-style links, autolinks — all driven through the real entry points (`resolveLink` via `parsePage`/`renderPage`, and `parseChangelog`), not just `classifyAbsoluteHref` in isolation. Every attempt either got caught as `UNSAFE-SCHEME`/`CHANGELOG-BAD-LINK` or fell through to relative-path resolution and died as `BROKEN-LINK`/`CHANGELOG-BAD-LINK` (file doesn't exist). Nothing renders a raw disallowed-scheme href. Full transcript of the adversarial run available on request; none of it is in this PR's diff so I'm not pasting 30 cases inline.

Confirmed the loud-failure claim by injecting real bad links and running the actual `pnpm build` (not a unit test):
- `docs/intro.md` + `[click me](javascript:alert(1))` → `pnpm build` exit 1, `Error [DocBuildError]: documentation build gate: 1 finding(s) / FAIL UNSAFE-SCHEME docs/intro.md:38`.
- `crates/trusty-search/CHANGELOG.md` + the same link → exit 1, `ChangelogBuildError: changelog build gate: 1 finding(s) / FAIL CHANGELOG-BAD-LINK crates/trusty-search/CHANGELOG.md:13`.

Both files restored afterward; worktree is clean.

Confirmed de-duplication: `parse.ts`'s own `ABSOLUTE` regex is gone (diff removes it at the old L130-131), replaced by `import { classifyAbsoluteHref } from '../docs/links'` (parse.ts:49). `grep` across `website/src/lib` turns up no second scheme regex.

Confirmed the corpus claim: no non-`http(s)` scheme anywhere in the 27 manifest `docs/` pages or the 6 flagship `CHANGELOG.md` files (checked inline links, reference-style definitions, and `mailto:` specifically — `mailto:` is newly disallowed by this PR and the corpus has zero uses, so nothing breaks on merge).

Reran the test claim myself rather than trusting the pasted output: replaced `links.ts`/`parse.ts` with `origin/main`'s pre-fix versions while keeping the new tests → exactly 6 failures (`class 6 — disallowed scheme` ×4, the two `javascript:`/`file://` cases in `parse.test.ts`), everything else in both files still green. Restored the fix, ran the full gate chain from a clean tree: `pnpm check` → 487 FILES 0 ERRORS 0 WARNINGS; `pnpm lint` clean; `pnpm test` → 14 files, 239 passed, 1 skipped (240); `pnpm build` → succeeds 3×. The 1 skip is `tests/mobile-overflow.test.ts`'s `/` at 320px case, gated by `route === '/' && width === 320` and tied to #5255 — pre-existing, unrelated to this change.

## Findings

| LOW | website/src/lib/docs/links.ts | 95-96, 110 | Protocol-relative hrefs are allowed with no host check. `//evil.com/x` and even `///etc/passwd` (verified against Node's WHATWG `URL` parser: resolves to `https://etc/passwd`, host `etc`) render as a live external link to an attacker-chosen host. Unchanged from the pre-PR `ABSOLUTE` regex, which already treated any `//`-prefixed href as external — and under this PR's own stated threat model (attacker already has commit access to a published doc/changelog) a plain `https://evil.com/x` was always available too, so this doesn't add capability. Worth a comment either way since the PR adds a test explicitly asserting this stays allowed (`links.test.ts` "passes a protocol-relative URL through untouched"). | If tighter behavior is wanted: drop the `href.startsWith('//')` early-return and require a resolved `https:`/`http:` scheme instead, i.e. treat `//` as inheriting `https:` and check that against `ALLOWED_SCHEMES` rather than auto-allowing. | Parent |
| LOW | website/src/lib/docs/links.ts | 84, 111-116 | `classifyAbsoluteHref` itself isn't whitespace/control-character tolerant — called directly with `"  javascript:alert(1)"` or `"java\nscript:alert(1)"` it returns `not-absolute`, not `unsafe-scheme`. Safety here is emergent: `resolveLink` trims (`links.ts:135`) and CommonMark's own link-destination grammar normalizes what markdown authoring can even produce, so every markdown-reachable variant I constructed still fails closed via `BROKEN-LINK`/`CHANGELOG-BAD-LINK`. A future third caller of `classifyAbsoluteHref` that skips those two guards would silently reopen this. | Trim (and reject embedded control characters) inside `classifyAbsoluteHref` itself so the guarantee doesn't depend on callers. | Parent |

## Notes

Both findings are informational — neither is reachable as a live bypass through either real entry point, confirmed empirically, not just by reading the code. `{@html}` sinks in `whats-new/+page.svelte` and `docs/[...slug]/+page.svelte` are the only two places this HTML gets rendered, and both consume output exclusively from the pipelines this PR patches, so there's no third unguarded sink.

Issue #987, opened 2026-06-09
chore: deduplicate the three SKIP_UI_BUILD build.rs scripts


bobmatnyc:
Filed from an independent simplification analysis of the workspace (read-only audit, 2026-06).

## Problem
trusty-search/build.rs (~120 lines), trusty-memory/build.rs (~152), and trusty-analyze/build.rs (~122) each reimplement identical SKIP_UI_BUILD logic (rerun-if-env-changed, skip_ui_build(), build_ui() via pnpm, embed check, pnpm-not-found warning). Cargo cannot share build scripts as a library, so full elimination isn't possible.

## Proposed change
Extract the shared logic into a single `scripts/build_ui.sh` (or a tiny build-support helper) invoked by each build.rs, OR accept the duplication but add a `diff`-based sync test so the three cannot drift. Note: each crate genuinely bundles a separate ui/ dashboard, so the UI builds themselves stay separate.

## Effort
S-M

## Risk
Low.

PR #5258, opened 2026-08-09
fix(gui): build the Svelte bundle in build.rs so a clean clone compiles

### LINKED ISSUE #4699: cargo clippy --workspace fails on a fresh worktree until trusty-code-gui's UI is built
Defect: on a fresh worktree `cargo clippy --workspace --all-targets` fails before linting anything:

```
error: proc macro panicked
   |
help: message: The 'frontendDist' configuration is set to "ui/dist" but this path doesn't exist
```

Cause: `crates/trusty-code-gui/build.rs` is a bare `tauri_build::build()` and never invokes pnpm, so `ui/dist` must be built by hand first.

Impact: every agent dispatched into a fresh worktree hits this and has to discover the workaround independently. Observed twice today.

Resolution options (not deciding here):
- have `build.rs` invoke pnpm to build the UI
- document the precondition in CLAUDE.md's build section
- set `SKIP_UI_BUILD` semantics for clippy

bobmatnyc:
## Defect

The three Tauri crates declare `frontendDist` pointing at a gitignored `dist/`
directory, and their `build.rs` was a bare `tauri_build::build()` that never
built it. On a clean clone `tauri::generate_context!()` panics with "this path
doesn't exist", stopping `cargo test --workspace` and
`cargo clippy --workspace --all-targets` before a single test runs.

Two things the issue did not have:

- **Three crates, not one.** `trusty-agents-ui`
  (`crates/trusty-agents/ui/src-tauri`) has the same defect. It is also the crate
  that turns the check ON for everyone: it enables `tauri/custom-protocol` by
  default, and resolver-v2 feature unification propagates that to every Tauri
  crate in a workspace-wide build. Fixing only `trusty-code-gui` and
  `trusty-mpm-gui` would leave the release gate unrunnable.
- **The panic is feature-gated.** `tauri-macros` sets `dev: cfg!(not(feature =
  "custom-protocol"))`, and `tauri-codegen` skips the `frontendDist` existence
  check whenever `dev` is true. So `cargo check -p trusty-code-gui` **passes** on
  a tree with no bundle — it falls back to `devUrl`. Only a build that pulls in
  `custom-protocol` fails. That is why this looked workspace-only, and why the
  reproducer below needs `--features tauri/custom-protocol` for the two GUI
  crates (`trusty-agents-ui` fails on its own, `custom-protocol` being its
  default).

## Evidence

All from a fresh worktree off `origin/main` with no `ui/dist` present.

**Before** — `CARGO_BUILD_JOBS=2 cargo check -p trusty-code-gui --features tauri/custom-protocol`, exit 101:

```
   Compiling trusty-code-gui v0.1.0 (.../crates/trusty-code-gui)
error: proc macro panicked
   |
35 |         .run(tauri::generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |

```

`trusty-mpm-gui` fails identically at `src/lib.rs:42`; `trusty-agents-ui` fails
at `src/main.rs:144` with `frontendDist ... "../dist"`, under a plain
`cargo check -p trusty-agents-ui`.

**After** — `CARGO_BUILD_JOBS=2 cargo check -p trusty-code-gui -p trusty-mpm-gui -p trusty-agents-ui`, all three dists deleted first, exit 0:

```
dists removed:
  crates/trusty-code-gui/ui/dist: ABSENT
  crates/trusty-mpm-gui/ui/dist: ABSENT
  crates/trusty-agents/ui/dist: ABSENT
   Compiling trusty-mpm-gui v0.2.12 (.../crates/trusty-mpm-gui)
   Compiling trusty-code-gui v0.1.0 (.../crates/trusty-code-gui)
   Compiling trusty-agents-ui v0.16.1 (.../crates/trusty-agents/ui/src-tauri)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 17.43s
--- bundles:
  crates/trusty-code-gui/ui/dist: assets fonts index.html
  crates/trusty-mpm-gui/ui/dist: assets index.html
  crates/trusty-agents/ui/dist: assets favicon.svg fonts index.html
```

**pnpm missing** (same command with pnpm off `PATH`), exit 101 — the point of the
change is that this is no longer a bare path panic:

```
thread 'main' panicked at crates/trusty-code-gui/build.rs:242:5:

trusty-code-gui: could not build the frontend bundle in `ui/`.
`pnpm --version` did not succeed in the `ui/` directory, so pnpm is unavailable or unusable there.
`tauri.conf.json` points `frontendDist` at that bundle, so without it `tauri::generate_context!()` fails with a bare "this path doesn't exist" panic (#4699).
Fix: install pnpm (https://pnpm.io/installation) and rebuild, or set SKIP_UI_BUILD=1 to compile against a placeholder UI.
```

**`SKIP_UI_BUILD=1`** (pnpm still off `PATH`), exit 0 — placeholder plus a loud
warning, matching `trusty-search` / `trusty-memory`:

```
warning: trusty-code-gui@0.1.0: trusty-code-gui: SKIP_UI_BUILD=1 and .../ui/dist is empty — embedding a PLACEHOLDER UI. The resulting binary will not show the real interface. Run `pnpm --dir ui install && pnpm --dir ui build` and rebuild without SKIP_UI_BUILD to get a working app.
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.70s
```

**No needless rebuild** — three consecutive checks, `ui/dist/index.html` mtime
unchanged after the first:

```
run1:    Compiling trusty-code-gui v0.1.0 (...)
run1:     Finished `dev` profile in 2.68s
run1 dist mtime: 1786285441
run2:     Finished `dev` profile in 0.19s
run2 dist mtime: 1786285441
run3:     Finished `dev` profile in 0.19s
run3 dist mtime: 1786285441
```

That took a second fix. `trusty-code-gui` had no `capabilities/` directory, but
`tauri_build` unconditionally emits `cargo:rerun-if-changed=capabilities`, and
cargo treats a declared-but-absent path as changed
(`dirty: FsStatusOutdated(StaleItem(MissingFile { path: ".../capabilities" }))`).
The build script re-ran on every invocation — pre-existing, but with this change
it would have re-run the pnpm build every time. The added
`capabilities/default.json` has `"permissions": []`; the regenerated
`gen/schemas/capabilities.json` confirms no permission is granted, so runtime
behaviour is unchanged.

## Resolution

- Each `build.rs` runs `pnpm install` (`--frozen-lockfile` when a lockfile
  exists) + `pnpm run build` in its UI directory, then verifies `index.html`
  actually exists. No `Command` result is discarded; a non-zero exit, a spawn
  error, or a zero exit that produced nothing all abort the crate build.
- `pnpm --version` is probed **from the UI directory**, because corepack resolves
  the `packageManager` pin relative to cwd — probing from the workspace root
  selects a different pnpm and fails on this machine.
- `cargo:rerun-if-changed` covers only paths that exist, and only inputs to
  `pnpm build` (`vitest.config.ts`, `vite.web.config.ts`, `web.html`,
  `playwright.config.ts`, `tests/` are excluded). `dist/` is deliberately not
  declared: cargo's staleness reference is `invoked.timestamp`, stamped before
  the script runs, so declaring the script's own output makes it dirty its own
  fingerprint — measured at a 2.5s pnpm rebuild on every `cargo check`. The
  residual is that deleting `dist/` by hand with a warm `target/` does not
  re-trigger the script; `cargo clean` or touching any UI source recovers.
- The shared logic is a CANONICAL BLOCK kept byte-identical across the three
  crates and gated by `scripts/check_buildrs_sync.sh`, which now checks two
  families. This follows the precedent #987 set for the four UI-embedding daemon
  crates rather than inventing a build-dependency crate mid-release. It is a
  **separate** family from theirs because the failure semantics differ: those
  degrade to a placeholder, which is right for an optional web surface and wrong
  for a desktop app whose entire window is the bundle. `trusty-agents-ui` is
  edition 2021, so the block uses no let-chains.

A shared `[build-dependencies]` path crate would remove the duplication outright
— all three crates are `publish = false`, so nothing blocks it. Left as a
follow-up rather than adding a workspace member during a release cut.

## Test ladder

Rung 3 (localized behavior, per crate) — build scripts only, no library API
changed. Commands run, all exit 0:

```
CARGO_BUILD_JOBS=2 cargo fmt --check
CARGO_BUILD_JOBS=2 cargo clippy -p trusty-code-gui -p trusty-mpm-gui -p trusty-agents-ui --all-targets -- -D warnings
CARGO_BUILD_JOBS=2 cargo test -p trusty-code-gui -p trusty-mpm-gui -p trusty-agents-ui
bash scripts/check_sld.sh
bash scripts/check_buildrs_sync.sh
```

```
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
line-cap: measured 3844 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK.
sld-lint: scanned 57 spec doc(s) + 3181 code file(s); 0 error(s), 0 warning(s)
daemon: build.rs canonical blocks are in sync across 4 crates.
tauri-ui: build.rs canonical blocks are in sync across 3 crates.
```

`cargo test --workspace` was not run here — it is the 37-minute gate this PR
exists to unblock, and a prior attempt on this machine was SIGKILLed. The
three-crate compile-from-clean above is the evidence that the blocker is gone;
the workspace re-gate belongs to the 1.3.5 cut.

Closes #4699

### DIFF
--- a/crates/trusty-agents/changelog.d/4699-agents-ui-build-dist.md
+++ b/crates/trusty-agents/changelog.d/4699-agents-ui-build-dist.md
@@ -1,0 +1,11 @@
+Fixed
+
+- **`trusty-agents-ui`'s `build.rs` now builds the Svelte bundle**, so a clean
+  clone compiles. Its `frontendDist` (`../dist`) is gitignored and nothing built
+  it, so `tauri::generate_context!()` panicked with a bare "this path doesn't
+  exist". This crate enables `tauri/custom-protocol` by default, which is what
+  turns that check on for every Tauri crate in a workspace-wide build — it
+  failed on its own `-p` build too. `build.rs` now runs `pnpm install` +
+  `pnpm run build`, honours `SKIP_UI_BUILD=1`, and aborts loudly if pnpm is
+  missing or the build produces no `index.html`
+  ([#4699](https://github.com/bobmatnyc/trusty-tools/issues/4699))
--- a/crates/trusty-agents/ui/src-tauri/build.rs
+++ b/crates/trusty-agents/ui/src-tauri/build.rs
@@ -1,3 +1,257 @@
+//! build.rs — builds the Svelte frontend bundle, then runs `tauri_build::build()`.
+//!
+//! Why: `src/main.rs`'s `tauri::generate_context!()` embeds `../dist` (this
+//! crate lives in `src-tauri/`, so that is `crates/trusty-agents/ui/dist`), the
+//! `frontendDist` declared in `tauri.conf.json`. That directory is gitignored,
+//! so on a fresh clone or worktree it does not exist and the proc macro panics
+//! with a bare "this path doesn't exist" — which took down `cargo test
+//! --workspace` and `cargo clippy --workspace --all-targets` for the ENTIRE
+//! workspace, since neither reaches a single test before the Tauri crates fail
+//! to compile (#4699). Nothing built the bundle; it existed only where someone
+//! had run pnpm by hand.
+//!
+//! This crate is the one that turns the check on for everybody: it enables
+//! `tauri/custom-protocol` by default, and resolver-v2 feature unification
+//! propagates that to every Tauri crate in a workspace-wide build. Its own
+//! `-p` build therefore fails on its own, with no extra feature flag.
+//!
+//! What: emits `cargo:rerun-if-changed` for the UI sources, then runs
+//! `pnpm install` + `pnpm run build` in the parent `ui/` directory unless
+//! `SKIP_UI_BUILD=1` is set. Every failure path — no usable pnpm, a failed
+//! install, a failed build, or a build that exits 0 without producing
+//! `../dist/index.html` — aborts the crate build with a message naming this
+//! crate and the `SKIP_UI_BUILD=1` escape hatch. A stale or empty `dist/` is
+//! never embedded silently.
+//!
+//! NOTE: the block between the CANONICAL BLOCK markers is kept byte-identical
+//! across all three Tauri crates — this one, `crates/trusty-code-gui/build.rs`,
+//! and `crates/trusty-mpm-gui/build.rs`; `scripts/check_buildrs_sync.sh`
+//! asserts it. THIS crate is edition 2021, which is why the block uses no
+//! let-chains. It is deliberately NOT the block the four UI-embedding daemon
+//! crates share (`trusty-memory`, `trusty-analyze`, `trusty-console`,
+//! `trusty-search`): those degrade to a placeholder on any failure, which is
+//! right for an optional web surface and wrong for a desktop app whose entire
+//! window is the bundle.
+//!
+//! Test: `cargo check -p trusty-agents-ui` on a tree with no
+//! `crates/trusty-agents/ui/dist` — the #4699 reproducer. It panics before this
+//! change and builds the real bundle after it.
+
+use std::path::Path;
+use std::process::Command;
+
+/// Names this crate in every diagnostic the shared block emits.
+const CRATE_NAME: &str = "trusty-agents-ui";
+
 fn main() {
-    tauri_build::build()
+    let crate_root = std::path::PathBuf::from(
+        std::env::var("CARGO_MANIFEST_DIR").expect("cargo always sets CARGO_MANIFEST_DIR"),
+    );
+    // Unlike the other two Tauri crates, this manifest sits in `src-tauri/`
+    // inside the frontend package, so the UI root is the PARENT directory.
+    let ui_dir = crate_root
+        .parent()
+        .expect("crate root always has a parent")
+        .to_path_buf();
+    let dist_dir = ui_dir.join("dist");
+
+    println!("cargo:rerun-if-env-changed=SKIP_UI_BUILD");
+    // Only paths that EXIST may be declared: cargo treats a declared-but-absent
+    // path as changed, so one stale entry re-runs the whole pnpm build on every
+    // `cargo check`. `vitest.config.ts`, `playwright.config.ts` and `tests/` are
+    // omitted for the same reason in reverse — they exist, but feed `pnpm test`,
+    // not the `pnpm build` this script runs.
+    for rel in [
+        "../package.json",
+        "../pnpm-lock.yaml",
+        "../pnpm-workspace.yaml",
+        "../index.html",
+        "../vite.config.ts",
+        "../svelte.config.js",
+        "../tailwind.config.js",
+        "../postcss.config.js",
+        "../tsconfig.json",
+        "../src",
+        "../public",
+    ] {
+        println!("cargo:rerun-if-changed={rel}");
+    }
+    // `../dist` is deliberately NOT declared. It is this script's own output,
+    // and cargo's staleness reference is `invoked.timestamp`, stamped BEFORE
+    // the script runs — so declaring it makes the script dirty its own
+    // fingerprint and re-run the pnpm build on every `cargo check` (measured:
+    // 2.5s each). The residual is that deleting `dist/` by hand while `target/`
+    // stays warm does not re-trigger this script; `cargo clean` or a touch of
+    // any UI source recovers.
+
+    // #4699: build the bundle before tauri_build, so a broken UI fails here
+    // rather than as an opaque proc-macro panic in src/main.rs.
+    build_tauri_ui(&ui_dir, &dist_dir, CRATE_NAME);
+
+    tauri_build::build();
 }
+
+// ── TAURI UI CANONICAL BLOCK BEGIN (kept in sync by scripts/check_buildrs_sync.sh) ──
+
+/// Build the Tauri frontend bundle into `dist_dir`, or abort the crate build.
+///
+/// Why: `frontendDist` must exist and be current before `generate_context!()`
+/// runs. Producing it here is what makes a clean clone compile (#4699); failing
+/// loudly is what keeps a half-built bundle from being embedded and shipped.
+/// What: honours `SKIP_UI_BUILD=1`, then requires pnpm and runs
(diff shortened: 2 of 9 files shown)


bobmatnyc:
## Verdict: APPROVE

Reviewed at head `1734454e` in a detached worktree (`.claude/worktrees/critic-5258`), independent of `.claude/worktrees/gui-ui-build`. No CRITICAL or HIGH findings at >80% confidence.

### Corrections verified against source, not by reasoning

**1. The panic is feature-gated — confirmed.** `tauri-macros-2.6.2/src/context.rs:155` sets `dev: cfg!(not(feature = "custom-protocol"))`; `tauri-codegen-2.6.2/src/context.rs:178` skips the `frontendDist` existence panic whenever `dev && config.build.dev_url.is_some()`, and all three crates' `tauri.conf.json` declare a `devUrl`. Reproduced empirically: `cargo check -p trusty-code-gui` alone, no dist, old bare `tauri_build::build()` → exit 0. Same command + `--features tauri/custom-protocol` → panics at the exact line/message the issue cites.

**2. Three crates, not two — confirmed, and confirmed it isn't `--workspace`-specific.** `trusty-agents-ui/Cargo.toml` has `default = ["custom-protocol"]`. With the pre-fix build.rs restored and dist deleted, `cargo check -p trusty-code-gui -p trusty-agents-ui` (no `--workspace`, no explicit feature flag) panics on `trusty-code-gui` — feature unification across a multi-crate invocation turns the check on, exactly as claimed. `trusty-agents-ui` alone (`-p trusty-agents-ui`, no flag) also panics on its own, confirmed separately.

**3. `capabilities` rerun-if-changed — confirmed, and confirmed the grant is empty.** `tauri-build-2.6.2/src/acl.rs:427` emits `cargo:rerun-if-changed=capabilities` unconditionally unless `capabilities_path_pattern` is overridden; none of the three crates override it. `git diff` confirms `crates/trusty-code-gui/capabilities/` didn't exist on `origin/main` at all (only the crate's own `gen/schemas/capabilities.json`, tracked, containing `{}`). The new `capabilities/default.json` has `"permissions": []`, and the regenerated `gen/schemas/capabilities.json` reflects the same empty array — it grants nothing. Format matches the pre-existing `default.json` files in the other two crates (`"windows": ["main"]`, same schema).

### Failure-mode hunt (swallowed success over failure)

No `let _ =`, `.ok()`, or discarded `ExitStatus` in any of the three build.rs files. `run()` matches every `Command::status()` outcome (`Ok(success)` / `Ok(failure)` / `Err`) and routes non-success into `fail()` (`panic!`, `-> !`). The "trust the artefact, not the exit code" step (`index.exists()` after `pnpm run build`) means a build that exits 0 but writes nothing still fails the crate build.

Reproduced the pnpm-missing path with a `pnpm`-free `PATH`: clean panic naming the crate and `SKIP_UI_BUILD=1`. Reproduced `SKIP_UI_BUILD=1` under the same pnpm-free `PATH`: exit 0, a `cargo:warning=` visible in plain (non-`-v`) `cargo check` output, and a placeholder `index.html` whose body text says outright "the UI bundle was not built" — not mistakable for a real bundle downstream.

One incidental finding while reproducing pnpm-missing: `pnpm --version` run from the workspace root on this machine resolves an incompatible mise-shimmed pnpm/Node combination and errors out, while the same probe from `ui/` (as the build.rs does, per its own `packageManager` pin) resolves the correct pnpm and succeeds. That's exactly the failure mode the code comment at `build.rs:128-130` says probing from `ui_dir` (not the workspace root) avoids — direct evidence the design choice is load-bearing, not defensive-coding theater.

### Design decision — canonical block + sync script

`scripts/check_buildrs_sync.sh` confirmed to pass on the current tree (both families) and confirmed to go red when a single character is changed in one crate's block (tested by mutating `trusty-code-gui/build.rs`'s block and re-running — `FAIL: tauri-ui canonical block in crates/trusty-mpm-gui/build.rs differs from crates/trusty-code-gui/build.rs`, exit 1). The two-family split (`daemon` vs `tauri-ui`, different marker infix) doesn't cross-contaminate — the marker text itself distinguishes them. Byte-identical-plus-a-verified-red-on-drift is an acceptable stand-in for the common-entry-point rule here: Cargo genuinely cannot share a build script as a library across workspace members without adding a new crate, and deferring that consolidation as follow-up during a release cut (all three are `publish = false`, so no urgency) is reasonable.

### Evidence re-verified independently

- Deleted all three `ui/dist` dirs, ran `CARGO_BUILD_JOBS=2 cargo check -p trusty-code-gui -p trusty-mpm-gui -p trusty-agents-ui` → exit 0, real bundles (assets/fonts/index.html per crate, not placeholders).
- Three consecutive `cargo check` runs on the fixed tree → exit 0 each time, `ui/dist/index.html` mtime unchanged across all three (no rebuild triggered).
- `fmt --check`, `clippy --all-targets -- -D warnings`, `test`, `check_line_cap.sh`, `check_sld.sh`, `check_changelog_fragment.sh` all re-run and pass (`test`: 25+4 passed, 0 failed across the two crates with tests; the third has none).
- Changelog fragments present in all three crates, accurately scoped and worded.
- No unintended diff outside the three build.rs files, the one new capabilities file, the one regenerated schema, the sync script, and changelog fragments — the four daemon crates are untouched.

### On the un-run `cargo test --workspace`

Correctly not run, per instructions and per the PR's own scope (this PR exists to unblock that gate, not to pass it standing on its own evidence). The scoped evidence — mechanism confirmed at the pinned tauri source, reproduced independently in both the single-crate and multi-crate-without-`--workspace` shapes, all three crates producing real bundles together — is sufficient to believe the workspace build's Tauri-crate compile step is fixed. It cannot and does not prove no other, unrelated crate has an independent compile break; that was never in scope for this diff.

### Findings

### Confidence notes

All claims above were verified directly (source citation + reproduction), not inferred — no findings held below 80% confidence in either direction.

PR #5278, opened 2026-08-09
fix(trusty-memory): fuse the BM25 lexical lane into the HTTP recall path (#5036)

### LINKED ISSUE #5036: trusty-memory: HTTP recall path is vector-only — existing RRF BM25 fusion never wired into the UserPromptSubmit hook

bobmatnyc:
Closes #5036.

## The defect

`MemoryService::recall` (`crates/trusty-memory/src/service/core.rs:648` on `origin/main`) called `recall_with_default_embedder` / `recall_deep_with_default_embedder` and nothing else. Neither that file nor `trusty-common/src/memory_core/retrieval/layers.rs` mentions bm25 or fusion. That is the whole chain the `UserPromptSubmit` hook uses: `commands/prompt_context/fetch.rs` → `GET /api/v1/palaces/{slug}/recall` → `web/recall_routes.rs` → `service/core.rs`. The file header on `recall_routes.rs` already claimed "vector + BM25 hybrid search"; it wasn't. `handle_memory_recall_deep` had the same gap on the MCP side.

## Eager or lazy? — the question the issue left open

**Eager, and it covers this path — but two other things had to be fixed or the wiring would have been inert anyway.**

`spawn_startup_backfill` (`bm25_backfill.rs:807`) → `run_startup_sweep` enumerates every palace under `state.data_root` via `palace_ids_on_disk`, opens each through the same registry the HTTP route serves, and backfills it at daemon startup; `bm25_repair::spawn_repair_sweep` repairs drops on an interval. Same `AppState`, same data root — so the palaces this route serves are exactly the palaces the sweep covers. Nothing here is lazy.

Two caveats that matter more than the answer:

1. **The whole lane is gated on `TRUSTY_BM25_DAEMON=1`, and ADR-0031 records it as set nowhere, with zero BM25 index dirs across 102 palaces.** The backfill is a no-op while the flag is off, and so is this fix. The issue's severity claim — that this degrades "every prompt-context injection every user gets every turn" — does not hold: with the flag off the MCP path it cites as the working reference is *also* vector-only. This is a latent gap that becomes real the moment the flag is flipped, not a live regression.

2. **The lane was per-palace everywhere except the client.** `AppState::bm25_client` is built exactly once, `Bm25Client::for_palace(default_palace)` (`lib.rs:767`), and `Bm25Client` holds one fixed socket path. Every search and every live index op therefore went to the *default* palace's socket regardless of which palace was being queried, while `bm25_backfill` correctly used the socket the supervisor returned per palace. Wiring fusion in without fixing that would have searched, for any non-default palace, a corpus the backfill never wrote to — a silent no-op for exactly the project palaces the hook serves. The write side was worse than a no-op: palace X's drawers were being filed into the default palace's corpus, and because the write usually *succeeded* it never even marked the palace dirty for repair.

## What changed

- `MemoryService::recall` runs the lexical lane concurrently with the vector lane (`tokio::join!`) and fuses. Covers both `deep` branches.
- `handle_memory_recall_deep` gets the same lane, so a query no longer answers differently depending only on `deep`.
- Fusion can now **promote** a BM25-only hit. The old `fuse_bm25_into_recall` only boosted drawers the vector lane had already returned, so a drawer dense retrieval missed stayed missed — the case a lexical lane exists to cover. Promoted hits hydrate from the in-memory drawer table (no disk walk) and are filtered through the caller's scope, which is why the MCP path now passes `scope_admits` where it previously relied on the vector list already being scoped.
- BM25 search and live index writes route to the socket of the palace being queried.

### Why promoted hits are not scored by textbook RRF

`prompt_context` drops every drawer under `DEFAULT_RELEVANCE_FLOOR` (0.35, #5037). A rank-only RRF score tops out at `2/61 ≈ 0.033`. Re-scoring both lanes the textbook way would have put *every* result under that floor and emptied the injection entirely — a worse regression than the bug. So vector scores are left on their own scale, a co-occurring drawer gets the usual `1/(60+rank+1)` bonus, and a promoted lexical hit is normalised onto the same 0..1 band as its share of the lexical lane's best hit. The test asserts the promoted hit clears the floor, so this cannot silently regress.

Known limitation, documented at the function: normalising against the lane's own best hit means the best lexical match scores 1.0 even when it is only the best of a weak field.

With `TRUSTY_BM25_DAEMON` unset, `bm25_search_optional` returns `None` and the response is byte-identical to before.

## Test ladder

**Rung 5** — cross-crate contract and process lifecycle (the index worker's daemon routing changed), so rung 4 plus `--include-ignored` integration coverage.

```
cargo fmt --check                                          EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings  EXIT=0
cargo check --workspace --all-targets                       EXIT=0
cargo test -p trusty-memory        629 passed; 0 failed; 4 ignored (+25 further binaries)
cargo test -p trusty-common        294 passed; 0 failed; 6 ignored
cargo test -p trusty-agents        3484 passed; 0 failed; 13 ignored  (only direct dependent)
cargo test -p trusty-memory --test bm25_backfill_e2e --test bm25_supervisor_e2e \
    --test bm25_supervisor_concurrency -- --include-ignored     11 passed; 0 failed
bash scripts/check_line_cap.sh / check_changelog_fragment.sh / check_sld.sh   all OK
```

### The regression test fails against the parent commit

`tests/recall_http_bm25_fusion.rs` drives the **real ONNX embedder** and a **real `trusty-bm25-daemon`** against a **non-default palace**, and states its preconditions as assertions: the backfill reports verified coverage, BM25 ranks the target for the query, and dense retrieval does *not* return it. A superseded attempt on `fix/5036-http-recall-bm25-fusion` used `seed_shared_embedder_with_mock()` and asserted the score rose by exactly `1/61` — that proves the call happened, not that anything became retrievable.

Parent commit (`git checkout origin/main -- crates/trusty-memory/src`, test file kept). Every precondition passed, so the failure is the fusion itself:

```
running 2 tests
FAILED
FAILED

---- http_recall_returns_a_lexical_match_the_vector_lane_misses stdout ----
panicked at crates/trusty-memory/tests/recall_http_bm25_fusion.rs:299:5:
#5036: the HTTP recall path must return the drawer only the lexical lane can find. Fused top-3 was:
   0.528  L2  Decision: adopt the staged rollout plan rather than a single cut
   0.513  L2  The rollout decision was recorded and the launch checklist updat
   0.480  L2  Rollout planning meeting concluded that we need a rollback path

---- deep_http_recall_also_fuses_the_lexical_lane stdout ----
panicked at crates/trusty-memory/tests/recall_http_bm25_fusion.rs:394:5:
#5036: deep recall must fuse the lexical lane too. Fused top-3 was:
   0.528  L3  Decision: adopt the staged rollout plan rather than a single cut
   0.513  L3  The rollout decision was recorded and the launch checklist updat
   0.401  L3  We decided against a big-bang deployment because the rollback st

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

With the fix, the target arrives at rank 1 carrying `L4` and a score of `1.000`:

```
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.26s
```

## Note on branches

`fix/5036-http-recall-bm25-fusion` holds an earlier, unmerged attempt from a decommissioned session (no PR). It is based on a stale main, carries ~1900 lines of unrelated `package-lock.json` churn, keeps the boost-only fusion, and does not touch the socket pinning. This PR is on a separate branch rather than a force-push over it.

---

## Review round (WARN → addressed)

All seven items. Two changed behaviour; the rest are tests and lint.

**HIGH 1 — dangling `Test:` pointers, CI red.** The two cited tests did not exist. They do now, in a new `crates/trusty-memory/src/tools/bm25_tests.rs` (the `bm25_backfill_tests.rs` convention), along with three more. `bash scripts/check_test_pointers.sh` → `resolved 23069 Test: citation(s) (floor 200) — 0 dangling pointers — OK.`

**HIGH 2 — the lane keyed off the requested slug, the vector lane off the resolved handle.** Correct, and worse than a mismatch: `palace_aliases.json` maps `bobmatnyc-trusty-tools → trusty-tools`, the alias directory has no `palace.json`, so `palace_ids_on_disk` never enumerates it and the backfill never writes the corpus the search would read — while the sweep reports full coverage. Fixed at all four sites with `handle.id.as_str()`: `core.rs`, both `memory_ops.rs` recall handlers, and the `helpers.rs` write path.

**HIGH 3 — the scoring inverted the premise of hybrid retrieval.** Accepted in full; the ordering was not intended and is not defended. `hit.score / best` put the top lexical hit at exactly `1.0`, above every possible cosine, so a drawer both lanes returned scored at most `cosine + 0.0164` and lost to a drawer only BM25 returned.

Taking the suggested option — cap at the best vector score — with one addition, because the cap alone does not fully remove the inversion: a drawer found weakly by both lanes would still score `0.1 + bonus` against `ceiling` for a lexical-only hit. So a drawer present in both lanes now takes `max(cosine, lexical)` before the RRF bonus. That makes a fused score **monotone in both lanes** — an additional lane finding a drawer can never lower its rank, which is the property that was actually violated.

Proved by mutation. Reverting just the scoring, with the new tests in place:

```
test tools::bm25::tests::fuse_bm25_lane_never_scores_agreement_below_a_lexical_only_hit ... FAILED
test tools::bm25::tests::fuse_bm25_lane_promotes_a_bm25_only_drawer ... FAILED

a drawer both lanes returned must outrank a weaker lexical-only hit;
  agreed=0.11639345 lexical_only=0.8888889

test result: FAILED. 3 passed; 2 failed
```

That is the reviewer's arithmetic reproduced exactly. Restored: `5 passed; 0 failed`.

The remaining tradeoff, stated rather than left implicit: a promoted hit can tie the best vector hit but never outrank it (the layer tie-break gives the vector hit the slot), so a palace whose vector lane is uniformly weak suppresses its lexical hits in proportion. That is the price of a score the 0.35 relevance floor can judge on the same scale. Documented at the function.

**MEDIUM 4 — no error-arm coverage.** Five unit tests added: the empty-hits no-op (asserting scores are not rewritten, not merely that length holds), promotion, the agreement ordering, the ADR-0027 T7 `admits` filter, and both `None` arms via a daemon binary pointed at a nonexistent path.

**MEDIUM 5 — `arm_lane()` returning `false` was a green.** Now panics with the build command. Both tests are `#[ignore]`d, so failing loudly costs CI nothing.

**MEDIUM 6 — the deep test dropped four of five preconditions.** Both tests now construct one `Fixture`, which asserts all five: lane enabled, index verifiably complete, target stored, BM25 ranks it, dense retrieval does not.

**LOW 7 — `best` was an explicit max while `rank` came from `enumerate()`.** One descending copy now feeds both.

### Rung 5 re-run

```
bash scripts/check_test_pointers.sh    0 dangling pointers — OK
cargo fmt --check                       EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings   EXIT=0
cargo check --workspace --all-targets   EXIT=0
cargo test -p trusty-memory             634 passed; 0 failed; 4 ignored  (was 629 — 5 new)
cargo test -p trusty-common             294 passed; 0 failed; 6 ignored
cargo test -p trusty-agents             EXIT=0, 11 suites ok, 0 failed
cargo test -p trusty-memory --test recall_http_bm25_fusion -- --include-ignored
                                        2 passed; 0 failed
cargo test -p trusty-memory --test bm25_backfill_e2e --test bm25_supervisor_e2e \
    --test bm25_supervisor_concurrency -- --include-ignored    11 passed; 0 failed
check_line_cap.sh / check_changelog_fragment.sh / check_sld.sh   all OK
```

---

## Review round 3 (WARN → addressed)

Both HIGH findings were real, and both fixes were small once in the code.

**HIGH A — the round-2 scaling was inert in production.** Correct, and the mechanism is exactly as described: `retrieve_l0_l1` (`layers.rs:180-184`) unconditionally prepends a synthetic identity row at a hardcoded `score: 1.0` on any palace with a non-empty identity. `ceiling` took the max over every row, so it was 1.0 in production and `ceiling * (hit.score / best)` reduced to the round-1 normalise-to-1.0 formula verbatim.

The ceiling is now the best **L2/L3** score. Those are the layers carrying a similarity; L0 is synthetic, and an L1 drawer the vector search never matched carries an importance-derived penalty, not a cosine.

Every existing fixture omitted the identity row, which is precisely why the class went unnoticed. `fuse_bm25_lane_ignores_the_synthetic_l0_row_when_scaling` adds it and reproduces the probe. Reverting only the layer filter:

```
test tools::bm25::tests::fuse_bm25_lane_ignores_the_synthetic_l0_row_when_scaling ... FAILED
test tools::bm25::tests::fuse_bm25_lane_promotes_a_bm25_only_drawer ... ok
test tools::bm25::tests::fuse_bm25_lane_never_scores_agreement_below_a_lexical_only_hit ... ok
test tools::bm25::tests::fuse_bm25_lane_is_a_no_op_without_hits ... ok
test tools::bm25::tests::fuse_bm25_lane_drops_a_drawer_the_scope_excludes ... ok
test tools::bm25::tests::bm25_lane_degrades_to_none_when_the_daemon_cannot_start ... ok

the identity row's synthetic 1.0 must not become the scaling ceiling; got 1

test result: FAILED. 5 passed; 1 failed
```

`got 1` is `PROBE promoted=1`. The new test is the only one of six that catches it — the five that pass are the fixture gap, measured. Restored: `6 passed; 0 failed`.

**HIGH B — a fifth call site.** `recall_without_embedder` no longer takes a `palace` parameter; it reads `handle.id.as_str()` internally, so the two lanes cannot disagree by construction. Three call sites follow (`:472`, `:517`, `:617`). This was the embedder-warming path, where the lexical lane is the only lane, so an aliased palace answered L0/L1 alone.

**Third item — the doc.** Both statements are now true rather than aspirational, and the doc names the L0 exclusion explicitly so the next reader does not have to rediscover why the filter is there.

**On the clippy claim.** Correct, and the correction is in your favour: `!(best > 0.0)` and `!best.is_finite() || best <= 0.0` differ at `+∞`, where the old form let `inf` through and computed `inf/inf = NaN`. The claim is narrowed to *identical NaN handling* — the new form is strictly better at the infinity boundary.

### Rung 5 re-run

```
bash scripts/check_test_pointers.sh   resolved 23071 citations — 0 dangling pointers — OK
cargo clippy -p trusty-memory --all-targets -- -D warnings   EXIT=0
cargo check --workspace --all-targets                        EXIT=0
cargo test -p trusty-memory          635 passed; 0 failed; 4 ignored   (0 FAILED suites)
cargo test -p trusty-common          294 passed; 0 failed; 6 ignored
cargo test -p trusty-agents          EXIT=0; 11 suites ok; 0 FAILED
cargo test -p trusty-memory --test recall_http_bm25_fusion -- --include-ignored
                                     2 passed; 0 failed
cargo test -p trusty-memory --test bm25_backfill_e2e --test bm25_supervisor_e2e \
    --test bm25_supervisor_concurrency -- --include-ignored   11 passed; 0 failed
check_line_cap.sh / check_changelog_fragment.sh / check_sld.sh   all OK
```


bobmatnyc:
Closed by owner decision, not abandoned. The scoring work is deferred to tm 1.3.6 and tracked on #5036, which carries the redesign brief.

Three review rounds each found the same defect shape: the fusion produced a plausible-looking score that no test constrained.

The socket/alias fix from this branch is independently correct and is being re-landed separately for 1.3.5.

The branch `fix/5036-recall-lexical-lane` is deliberately KEPT as the record of the three rounds.

Issue #4402, opened 2026-07-30
CI: ~4min of the 17min critical path is waste (vestigial docker prune + redundant recompile)


bobmatnyc:
**Defect:** the CI critical path is 17m15s median and is entirely the `Test` job. Two segments inside it do no useful work.

**Waste 1 — `docker image prune --all --force`, 3m01s.** `df -h` in the same job reports **88GB available, 40% used** on a 145GB disk immediately before the prune runs. No space pressure exists. The step dates from #3668 (test binaries + un-swept `.rcgu.o` exhausting disk); the actual remedy for that, the `.rcgu.o` sweep step, landed separately and is still present. The prune is vestigial. It also runs in `Clippy`, `MSRV`, and the daemon smoke test, burning runner-minutes off the critical path.

**Waste 2 — 1m07s redundant recompile.** `.github/workflows/ci.yml` (~lines 541-571) runs `cargo test --workspace --no-run`, sweeps `.rcgu.o`, then runs `cargo test --workspace` again with `env: CI: ""`. That override exists so `trusty-common`'s `cache_fresh_returns_some_when_newer` test takes its non-CI branch, but it is applied to the entire workspace step, flipping cargo's fingerprint and forcing `trusty-agents` and `trusty-agents-local` to recompile. Confirmed in job logs.

**Resolution:** drop or conditionalize the docker prune (verify disk headroom empirically first, since #3668 was a real incident); scope the `CI: ""` override to just the one cache-freshness test rather than the whole workspace invocation.

**Expected:** ~17-18min → ~13-14min critical path. No test coverage changes.

**Also worth a look, not part of this fix:** the `Test pointers` lint job took 5m31s in one observed run versus 10-90s for every sibling lint job. Off the critical path, so no wall-clock cost, but disproportionate.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

Issue #3668, opened 2026-07-22
ci: Test job exhausts runner disk — 179 static test binaries × full debuginfo + unswept .rcgu.o


bobmatnyc:
## Root cause

The `Test` job (`cargo test --workspace ...` on `ubuntu-latest`) exhausts the
runner's disk mid-link (`ld` SIGBUS + `ENOSPC`). This happened **2x today on
PR #3665**.

Quantified breakdown:

- 27 workspace crates, 1221 total dependencies.
- 126 integration-test files across the workspace (trusty-mpm 32,
  trusty-search 16, trusty-code 14, trusty-memory 13, plus the remainder
  spread across the other crates) — each integration-test file compiles to
  its own statically-linked test binary.
- ~179 statically-linked test binaries are produced by
  `cargo test --workspace`, each carrying **full default debuginfo**
  (`debug = true`, DWARF level 2).
- Those binaries alone total **~9 GB** on disk.
- The intermediate per-codegen-unit object files (`*.rcgu.o`) produced while
  compiling and linking those binaries total **~18 GB**, and are never swept
  after the final link — cargo has no built-in post-link cleanup for them.
- Linux ELF binaries embed DWARF debuginfo directly in the executable
  (unlike macOS, where debug info lives in a separate `.dSYM` bundle) — so on
  the `ubuntu-latest` runner, full debuginfo directly inflates every one of
  the ~179 linked test binaries. (A macOS-based sanity check is not a valid
  proxy for this disk cost for exactly this reason.)

The existing "Free disk space" step (ci.yml, the comment currently at
`test:` job around line 214) already acknowledges this is a known problem
("...the baseline removal set below still left the runner exhausting its
disk mid-build on `cargo test --workspace`...") but the mitigation shipped
at the time (purging unused preinstalled SDKs: dotnet, android, vcpkg, JVMs,
miniconda, az) was insufficient on its own — it addresses fixed OS-image
bloat, not the growth from this workspace's own test-binary output.

## This PR's mitigations

1. **`CARGO_PROFILE_TEST_DEBUG: line-tables-only`** on the `test` job only.
   Verified locally that `cargo test` picks up `CARGO_PROFILE_TEST_DEBUG`
   directly (the workspace `Cargo.toml` has no `[profile.test]` override, so
   `test` normally inherits `dev`'s implicit default debug level — the env
   var overrides the `test` profile explicitly and cargo honors it without
[...]

PR #5326, opened 2026-08-10
perf(ci): reclaim runner disk only when df says it is needed

### LINKED ISSUE #5325: perf(ci): the Test job takes ~17 minutes, which serializes merges to one per CI cycle
**Problem:** the `CI` workflow's `Test` job runs ~17-18 minutes. Measured 2026-08-10 on [#5314](https://github.com/bobmatnyc/trusty-tools/pull/5314): run [31410011416](https://github.com/bobmatnyc/trusty-tools/actions/runs/31410011416), `Test` started 16:38:59, SUCCESS at 16:57:00. The other five required checks finished in 3-6 minutes (Clippy 16:43:56, smoke test 16:44:41, MSRV 16:42:13). `required_status_checks.strict = true` means a branch must be up to date with `main` to merge, so a ~17-minute job caps merges at roughly one per CI cycle.

**Evidence:** while [#5314](https://github.com/bobmatnyc/trusty-tools/pull/5314) waited for its cycle, five unrelated PRs landed on main ([#5322](https://github.com/bobmatnyc/trusty-tools/pull/5322), [#5315](https://github.com/bobmatnyc/trusty-tools/pull/5315), [#5308](https://github.com/bobmatnyc/trusty-tools/pull/5308), [#5305](https://github.com/bobmatnyc/trusty-tools/pull/5305), [#5303](https://github.com/bobmatnyc/trusty-tools/pull/5303)), putting it `BEHIND` and costing a full additional cycle after `gh pr update-branch`. The 1.3.5 release train hit the same wall — six PRs cost roughly an hour of pure waiting. A duplicate `CI` run also triggers per PR (two `pull_request` events ~19s apart on the same SHA, serialized by a concurrency group), so the wall-clock cost can double — worth checking as part of this, since it may be cheaper to fix than the job itself.

Related but not overlapping: [#4402](https://github.com/bobmatnyc/trusty-tools/issues/4402) (closed) cut ~4 minutes of waste from this same job in the 1.3.1 cycle, yet the job is still ~17-18 minutes today. [#4389](https://github.com/bobmatnyc/trusty-tools/issues/4389) (open) addresses the merge-serialization symptom via a GitHub merge queue, not the job's own runtime. Neither covers reducing how long the `Test` job itself takes.

🔴 **Constraint: the job must get faster without reducing what is tested.** No `--exclude`, no `#[ignore]`, no `cfg`-gating, no narrowing to `--lib`, no dropping test targets. Any accepted fix must demonstrate identical test coverage before and after.

**Confidence:** Observed/Reproduced — timings taken directly from the [#5314](https://github.com/bobmatnyc/trusty-tools/pull/5314) run and cross-checked against the 1.3.5 release train.

**Closure conditions:**
- Investigation identifies where the `Test` job's ~17-18 minutes actually goes (build time vs. test execution vs. duplicate-run overhead).
- A concrete recommendation for cutting wall-clock time with no coverage reduction is recorded (here or in a follow-up issue).
- The duplicate `pull_request`-trigger overlap is confirmed wasteful and resolved, or confirmed non-wasteful and closed out.

bobmatnyc:
Closes #5325.

## Result

**What this diff is mechanically responsible for: the `Free disk space` step, 1m26s -> 0s.** That is the number to hold me to. The step is variable — 56s, 76s, 86s and 147s observed across sampled jobs — so the saving is "the whole step", roughly 1-2.5 min depending on the runner.

Full paired comparison against `main` at the same tip (`96123699`), started one minute apart on the same runner pool:

| | main (push 31415777119) | branch (PR 31415844017) | delta | attributable? |
|---|---:|---:|---:|---|
| Free disk space | 1m26s | 0s | **-1m26s** | **yes — this diff** |
| build phase (`--no-run`) | 8m16s | 8m00s | -16s | no — identical command, noise |
| test execution | 7m14s | 7m09s | -5s | no — identical command, noise |
| checkout + toolchain + apt + cache | 57s | 35s | -22s | no — this PR touches none of these steps |
| **job total** | **18m12s** | **16m08s** | **-2m04s** | only ~86s of it is this diff |

The -2m04s / -11.4% job-total figure is a single-pair observation, and three of its four rows are variance this PR did not cause. I previously led with it; that overstated the evidence. Expect roughly **1-1.5 min** in steady state.

## Where the 18 minutes actually goes

Step-level timings from the `Test` job of run [31410011416](https://github.com/bobmatnyc/trusty-tools/actions/runs/31410011416) (PR #5314), the run #5325 was filed against. Job total 17m58s; the other five required checks finished in 3-6 min.

| Step | Duration | Share |
|---|---:|---:|
| Set up job + checkout | 8s | 0.7% |
| **Free disk space** | **2m27s** | **13.6%** |
| Create local main branch | 1s | |
| Install stable toolchain | 2s | |
| Install system dependencies | 17s | 1.6% |
| Cache cargo artifacts (restore) | 18s | 1.7% |
| fastembed cache key + restore | 1s | |
| **`cargo test --workspace --no-run` (build)** | **7m42s** | **42.9%** |
| Sweep `.rcgu.o` | 0s | |
| **`cargo test --workspace` (execute)** | **6m57s** | **38.7%** |
| `cargo test -p trusty-common` (cache-freshness) | 5s | |

Three segments are 96% of the job. This PR takes the first two.

**Free disk space (56-147s, highly variable).** `df` at the top of the step: `145G 58G 87G 40%`. After the purge: `115G` available. The job's peak need is ~27G ([#4402](https://github.com/bobmatnyc/trusty-tools/issues/4402)'s measurement). So it spends 147s deleting ~29GB to move from 3.2x headroom to 4.3x. #4402 gated the `docker image prune` sitting directly under it on measured headroom — the log line `Skipping docker image prune: 115G free (>= 25G threshold, #4402)` confirms that gate is present and firing — but left the `sudo rm -rf` above it unconditional. The `rm -rf` is the more expensive half.

**Build phase (462s).** The cargo cache is a full hit (`Restored from cache key ... full match: true`, 1228 MB), so this step compiles **25 workspace crates** and links **194 test executables**, and zero dependencies. Linking, not compiling, is most of it.

**Execution phase (417s).** Not touched here — see "Not in this PR" below.

## What changed

**1. `scripts/ci-free-disk-space.sh`** — #4402's own gate, one level up, applied to the purge as well as the prune. Three tiers on one measured number:

- `avail >= 45G` -> skip both (the observed case: 87G)
- `avail < 45G` -> purge unused SDKs, re-measure
- re-measured `< 25G` -> also prune docker images (#4402's tier, unchanged)

Reclamation is **gated, not deleted**: #3668 was a real disk-exhaustion incident, so the purge still fires on its own if a runner image or a heavier suite erodes the margin. Four jobs (`test`, `clippy`, `msrv`, `search-daemon-smoke`) had byte-identical copies of the inlined shell; they now share this one decision.

**2. Nothing else.** A linker swap was tried for the build phase and reverted — see "Measured and rejected" below. The build step's command is byte-identical to main.

## Coverage is unchanged, by construction and by count

Nothing in this PR touches how tests are invoked, which targets are built, or which tests run. Every `cargo` line is byte-identical to main.

Counted from job logs, base commit versus branch. The base is `8ca0b813`, this branch's actual merge-base, measured by its own `push: main` run — **not** another PR's run. That distinction matters: comparing against run 31410011416 (PR #5314, base `d51fd67a`) shows a 48-test gap that is entirely main moving between the two commits, and would have looked like a coverage regression.

| | passed | failed | ignored | filtered out | result blocks | doctest targets |
|---|---:|---:|---:|---:|---:|---:|
| base `8ca0b813` (run 31409039543) | 23,613 | 0 | 151 | 1,905 | 224 | 24 |
| this branch (run 31414074446) | 23,613 | 0 | 151 | 1,905 | 224 | 24 |
| **delta** | **0** | **0** | **0** | **0** | **0** | **0** |

Identical in every field.

## Test ladder

**Rung 5** by risk class — CI/build tooling, where the failure mode is a gate you cannot trust rather than a compile error. The diff contains zero Rust, so rung 5's Rust-side commands (`--include-ignored` integration coverage, `cargo test -p <consumer>`) have no target. What discharges it instead:

```
bash scripts/check-ci-helpers-selftest.sh   -> all 64 cases passed
```

Seven of those 64 are new and cover every branch of the new gate, including both thresholds at their exact boundary:

```
ci-free-disk-space:
  ok   87G observed on the measured run -> no work                -> skip-all
  ok   at the purge floor exactly                                 -> skip-all
  ok   just below the purge floor                                 -> purge
  ok   purge frees enough to skip the prune                        -> purge
  ok   purge leaves under the prune floor                          -> purge+prune
  ok   at the prune floor exactly                                  -> purge
  ok   empty disk takes both tiers                                 -> purge+prune
  ok   ci.yml has no inlined SDK purge left                        -> 0
  ok   all four disk-reclaim jobs call the helper                  -> 4
```

The last two are wiring assertions: the script being correct proves nothing if a job still inlines the old unconditional `rm -rf`. They fail closed if a fifth job is added with a copy-pasted purge, which is how this duplication reached four copies in the first place.

Other gates, all exit 0:

```
scripts/check_changelog_fragment.sh   scripts/check_line_cap.sh
scripts/check_sld.sh                  scripts/check_doc_numbers.sh
scripts/check_test_pointers.sh        scripts/check_public_docs.sh
scripts/check_public_docs_selftest.sh scripts/check_generation_artifacts.sh
scripts/assemble_changelog_selftest.sh scripts/check_line_cap_selftest.sh
cargo fmt --check
```

No changelog fragment: CI-only PR, no crate `src/**` touched.

## Measured and rejected: mold

194 test executables linked on a 4-core runner looks exactly like a linker bottleneck, so mold 2.30 was tried, invoked as `mold -run` so that cargo's fingerprint and `Swatinem/rust-cache`'s key (which hashes every `RUST*`-prefixed variable) stayed untouched and the comparison stayed clean.

It changed nothing measurable. Build-phase duration across five runs, four of them the identical unmodified command:

| run | mold | build phase | cache | crates | executables |
|---|---|---:|---|---:|---:|
| 31410011416 | no | 7m42s | full match | 25 | 194 |
| 31409039543 | no | 7m49s | full match | 25 | 194 |
| **31414074446** | **yes** | **8m01s** | full match, same key | 25 | 194 |
| 31415844017 | no | 8m00s | full match | 25 | 194 |
| 31415777119 | no | 8m16s | full match | 25 | 194 |

mold's 8m01s sits inside the 7m42s-8m16s spread of runs without it. Reverted rather than carry an apt package and an LD_PRELOAD interposer for an effect smaller than the step's own variance.

The useful part is the negative result, and it is recorded at the step so the next person does not repeat it: **this step is compile-bound, not link-bound.** The cost is codegen for 25 large crates down a partly-serial dependency chain — everything waits on `trusty-common` — which no linker can touch.

## Not in this PR

**The 417s execution phase.** `cargo test` runs the 194 test binaries **one at a time**. Summed per-binary wall time is 315s, of which the three slowest are 139.6s back-to-back (49.5s / 46.4s / 43.6s) while 208 binaries account for only 74.7s between them. `cargo nextest` would interleave all of them in one global pool.

It is not proposed here because it is **prior art that was already reverted**: `ci.yml` lines 826-851 record that #4402 tried exactly this and backed it out. Nextest's process-per-test isolation surfaced a real pre-existing trusty-memory bug — `add_alias_round_trip_through_prompt_cache` and `dispatch_discover_aliases_inserts_new_and_dedupes` implicitly depend on `TRUSTY_SKIP_PALACE_ENFORCEMENT` being set as a side effect by some other test in the shared process (`crates/trusty-memory/src/tools/palace_ops.rs:88-101`). Adopting nextest means fixing that first, in trusty-memory, as its own change. That is a real ~200s lever with a named prerequisite, not a dead end.

**Matrix sharding.** Would parallelise both phases, but a matrix renames the check context (`Test` -> `Test (shard-1)`), and `Test` is a required status check. That needs a branch-protection change, which is the owner's call.


bobmatnyc:
### Coverage proof — measured twice, each against its own exact merge base

`actions/checkout` on a `pull_request` builds `refs/pull/N/merge`, so the base is main's tip **at run time**, not the branch point. Comparing against the wrong base manufactures a 48-test gap that looks exactly like a coverage regression. Both pairs below use the true base, verified from the merge commit's first parent, and each base number comes from that commit's own `push: main` run.

**Pair 1** — base `8ca0b813` (run 31409039543) vs branch run 31414074446:

| | passed | failed | ignored | filtered out | result blocks | doctest targets |
|---|---:|---:|---:|---:|---:|---:|
| base | 23,613 | 0 | 151 | 1,905 | 224 | 24 |
| branch | 23,613 | 0 | 151 | 1,905 | 224 | 24 |

**Pair 2** — base `96123699` (run 31415777119) vs branch run 31415844017:

| | passed | failed | ignored | filtered out | result blocks | doctest targets |
|---|---:|---:|---:|---:|---:|---:|
| base | 23,661 | 0 | 151 | 1,905 | 224 | 24 |
| branch | 23,661 | 0 | 151 | 1,905 | 224 | 24 |

Identical in every field, both times. The +48 between the two pairs is main moving; it appears on both sides of pair 2.

### Did #4402 regress, or was it never sufficient?

**Neither — it is intact and still working, and it fixed a different segment.** Both of its changes are present on main and firing:

- The docker-prune gate produced `Skipping docker image prune: 115G free (>= 25G threshold, #4402)` in run 31410011416.
- The `CI: ""` scoping is present as its own `cargo test -p trusty-common` step, costing 5s instead of the workspace-wide recompile it replaced.

What #4402 did not do is gate the `sudo rm -rf` sitting three lines above the prune it gated. That `rm -rf` is the larger half — 1m26s in the paired control, 2m27s at its worst — and it is what this PR takes. So #4402's ~4 min was real and is still banked; the remaining time is a different cost.

One thing #4402's own estimate did miss: it projected 13-14 min and the job stayed at 17-18. Its two fixes together are worth ~4 min against a job that was ~21-22 min at the time, not against today's 17-18.

Issue #5075, opened 2026-08-07
trusty-search: unregister_index leaves stale cold-store entries, so DELETE gives a permanent 503


bobmatnyc:
## Defect

`unregister_index` (`crates/trusty-search/src/service/server/search.rs:106-152`) removes the index from the hot registry but never purges `cold_store.entries` or `cold_store.failed_entries`. Deleting a cold-parked or restore-failed index leaves the stale cold-store entry behind.

That directly violates the invariant #5057 writes at `crates/trusty-search/src/mcp/tools/not_ready.rs:78-85`: 404 is reserved for "absent from every store", 503 means "registered but not resident". After `DELETE /indexes/:id`, the index is absent from every store — but the three cold-store guards #5057 added (`status.rs:103-110`, `files.rs:132-138`, `files.rs:274-294`) answer 503 forever for an id that no longer exists, since they check `cold_store.contains()` / `cold_store.is_failed()` without checking the registry was ever cleared of it too.

Pre-existing on the search path (`search_handler` has the same gap) — #5057 did not cause it, but #5057's guards are what make it observable and permanent, and the MCP layer can no longer report `INDEX_NOT_READY` for the delete-then-reindex case (the #4715 scenario).

## Evidence

Promoted from the code-critic round-3 review of PR #5057 at head `e5c246033`: https://github.com/bobmatnyc/trusty-tools/pull/5057#pullrequestreview-4880053803

## Resolution

Purge both `cold_store.entries` and `cold_store.failed_entries` in `unregister_index` alongside the registry removal.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4708, opened 2026-08-03
`.tmp1rIzAZ` index walker stalled indefinitely


bobmatnyc:
## Summary

An index is stuck in `lexical: in_progress` with zero chunks and never advances. The walk has nothing meaningful to do, yet never completes or times out.

## Measurements (2026-08-03)

- **Measured:** Index `.tmp1rIzAZ` reports stage `lexical: in_progress` with `chunk_count: 0`, unchanged across two checks more than 5 minutes apart.
- **Root path:** `/private/var/folders/7s/.../T/.tmp1rIzAZ` still exists but contains essentially nothing — only a one-line `.gitignore` stub.

## Concerns

- A walk over a near-empty directory should finish or fail fast; an indefinitely `in_progress` stage is indistinguishable from a slow legitimate build
- Nothing surfaces it as stuck — there is no visible timeout or stall detection
- Note this is low-value data (a scratch temp dir), so the bug is the STALL BEHAVIOR, not the lost content

## Suggested Fixes

- Add a stage timeout
- Add a stalled-walk health signal

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Moved to ts 0.45.0 — #4680 already covers the never-walked case. Closing the started-but-stalled gap needs a stage timeout that would abort legitimate multi-minute walks on large indexes, and there is no timestamp to key a stall on. Split off as #5336.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Superseded. #4680 (closed) covers the never-walked case; the remaining started-then-stalled scope was split off into #5336 (open). No residual scope remains here.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5335, opened 2026-08-10
fix(trusty-search): index registry write integrity, root lockstep, and cold-store lifecycle (#4951, #4871, #4317, #4253, #5075)

### LINKED ISSUE #4871: trusty-search: indexes.toml is clobbered by same-process write races and a fail-open parse
**Symptom.** The on-disk registry `~/Library/Application Support/trusty-search/indexes.toml` spontaneously reverted to a byte-identical earlier state during a measurement, discarding 8 entries that had been written minutes before. Counts went 80 → 88 (after a test run that leaked 8 `$TMPDIR/.tmp*` roots) → back to 80, with those 8 gone and the file matching a pre-measurement snapshot exactly.

**Cause.** **Five `trusty-search serve` processes were running concurrently.** Each holds its own in-memory view of the registry and rewrites the whole file from that view. Whichever writes last wins, and every registration made by the others since their own view was loaded is silently discarded. There is no merge, no conflict detection, and no error — a write that is thrown away looks identical to a write that succeeded.

**Why this is worse than it sounds.**
1. **It silently reverts legitimate registrations.** An index registered by one daemon can vanish when another flushes an older view. The caller sees success.
2. **It makes registry cleanup unreliable.** A prune performed against the file (or via one daemon) can be undone wholesale by another daemon's stale view. This is a plausible explanation for an event earlier the same day, where a manual prune took the registry from 396 to 41 entries and it read 43 an hour later — previously attributed entirely to test-suite repollution (#4255). Both mechanisms may have contributed; the relative weight is **not established** and should be measured rather than assumed.
3. **It corrupts measurement.** Any before/after registry comparison is only valid if no other daemon flushes during the window.

**Reproduction sketch.** Start two `trusty-search serve` instances against the same data dir. Register index A via one and index B via the other. Have the first flush its view. Observe B's entry disappear from `indexes.toml` with no error surfaced to either caller.

**Open questions for whoever picks this up.**
- Why are five daemons running at once? Is that a supported configuration, an MCP-bridge auto-spawn artifact, or an orphan-accumulation bug in its own right? The answer changes whether the fix is "make writes safe" or "prevent the second daemon".
- Should registry writes be read-modify-write under a file lock, an append/merge rather than a whole-file rewrite, or should a single owner process be enforced?

**Related, deliberately not merged into this ticket:**
- **#4255** (closed) — the test suite registering into the live registry. A *different* mechanism for registry corruption: pollution rather than reversion. Fixed by #4864 (`da1e8dcb`) and #4870.
- **#4846** — warm boot walks every dead registry entry at ~10s each. Aggravated by this issue if a stale view resurrects entries a prune removed.
- **#4317** (open) — the registry-loss incident. This is a plausible contributing mechanism worth checking against that timeline.
- **#4250** — warm-boot timeout skips indexes with no self-heal.

**Note for anyone measuring the registry:** a 404 from `index_status` does NOT mean an index is empty (an unloaded index returns 404 whether it holds 0 chunks or 70,000), and `list_indexes.size_bytes` reads 0 on populated indexes (#4706). Root-path existence is the only reliable discriminator, and now, per this issue, even the file contents are not stable while multiple daemons run.

### LINKED ISSUE #4951: trusty-search: stale-root file/path mismatch discards all search results after per-index reindex (was: 'snapshot never swapped')
> **UPDATE 2026-08-05:** The root cause originally reported below (snapshot never swapped) was wrong and has been corrected — see the comment thread. Actual cause: the search-side `file_is_within_root` post-filter (added for #64/#541) discards every candidate because the walker built the stored `file` field against a stale parent root, while `path` was built against the current root. The "spontaneous auto-reindex" observation reported below was also retracted — it was our own deliberate second reindex, not daemon behavior. The evidence below (status/search API responses, control comparison) is unchanged and still accurate; only the two conclusions above were corrected.

## Defect Summary

A per-index reindex can complete and report `status: "ready"` while search against that index still returns zero results. The build completes and populates chunks correctly, but every search response comes back empty with `stale_index_root: true`.

**Root cause (corrected):** the search handler in `crates/trusty-search/src/service/server/search.rs` post-filters every candidate through `file_is_within_root(&chunk.file, &handle.root_path)`, and sets `stale_index_root: filtered_out > 0`. The walker stores `path` (relative) and `file` (absolute) built against different root references — when an index's root is set one level below the daemon's original registration point, the absolute `file` is joined against the *old* root and ends up pointing outside the *current* root. `file_is_within_root` then fails for every chunk, so search discards 100% of matches even though they were found. Full trace and reproduction in the comments below.

**Observed on:** `atlassian` index on green (i-0e79d9652eb7fd756, us-east-1) on 2026-08-05.

## Evidence

State reported by `/indexes/atlassian/status` — every field indicating success:

```json
{"chunk_count":85642,
 "components":{"kg":{"status":"ready"},"vector":{"status":"ready"}},
 "last_walk_files_seen":31654,
 "root_path":"/mnt/data/knowledge/Jira",
 "search_capabilities":["bm25","literal","exact_match","vector","kg"],
 "stages":{"graph":{"status":"ready"},"lexical":{"status":"ready"},
           "semantic":{"completed_at":"2026-08-05T19:00:34.229664230+00:00",
                       "embedded":85642,"status":"ready","total":85642}},
 "status":"ready"}
```

**Actual search behavior** at the same time, with terms guaranteed to match:

```
POST /indexes/atlassian/search {"text":"onboarding"}
  -> {"results":[], "meta":{"stale_index_root":true, "search_capabilities":[], "last_indexed":"2026-06-26T22:50:24Z"}}

POST /indexes/atlassian/search {"text":"Mongo pool env"}
  -> same, empty, stale_index_root:true

POST /indexes/atlassian/search {"text":"the","mode":"all"}
  -> same, empty, stale_index_root:true
```

**Control** — two healthy indexes, identical query text, same instant:

```
POST /indexes/apex/search   {"text":"onboarding"} -> 3 results (score ~0.0016), stale_index_root:false, capabilities populated
POST /indexes/github/search {"text":"onboarding"} -> 3 results incl. "Milestone: ONB-M4 — Frontend Core", stale_index_root:false
```

Note `search_capabilities` is populated on `/status` but **EMPTY in the search response `meta`** for the same index — an internal inconsistency.

**Chunk-level evidence for the root cause** — `GET /indexes/atlassian/chunks?limit=5`:

```
"file": "/mnt/data/knowledge/ACP/ACP-1.md"      <- absolute, WRONG, dangling (no Jira/ segment)
"path": "ACP/ACP-1.md"                          <- relative, CORRECT
```

`/mnt/data/knowledge/ACP` does not exist; the real file is `/mnt/data/knowledge/Jira/ACP/ACP-1.md`.

## Secondary Issue: `meta.last_indexed` is Stale/Incorrect

`meta.last_indexed` in the search response read `2026-06-26T22:50:24Z` on **both** the broken `atlassian` index and the healthy `github` index simultaneously. This field is not diagnostic for index freshness — it reflects an earlier build, not the current serving state. Anyone using it to judge index freshness gets the wrong answer.

## Impact

JIRA semantic context has been absent from every PR review for 40+ days. The `atlassian` index held `chunk_count=0` from 2026-06-26; a scoped reindex on 2026-08-05 rebuilt it to 85,642 chunks, but search still returned nothing until the `file`/`path` root mismatch is fixed. Verified across 5 PR reviews and 3 direct API probes — all `raw_count=0`.

## Recommendation for Operators

**Status fields cannot be used as reindex acceptance criteria.** A reindex must be validated by:
1. Issuing a real query with a broad term
2. Confirming non-empty results
3. Confirming `stale_index_root: false` in response `meta`

The daemon should either:
- Refuse to report `status: ready` while `stale_index_root` would be true for the current root, OR
- Surface the stale root prominently in `/status` rather than only in search-response `meta`, OR
- Log/return a sample of the offending `file` paths when the post-filter discards candidates — this single detail would have cut hours off root-causing this issue.

bobmatnyc:
Closes [#4951](https://github.com/bobmatnyc/trusty-tools/issues/4951), [#4871](https://github.com/bobmatnyc/trusty-tools/issues/4871), [#4317](https://github.com/bobmatnyc/trusty-tools/issues/4317), [#4253](https://github.com/bobmatnyc/trusty-tools/issues/4253), [#5075](https://github.com/bobmatnyc/trusty-tools/issues/5075).

References [#4227](https://github.com/bobmatnyc/trusty-tools/issues/4227) without closing it — that issue got documentation only, and the actual recreate-to-empty guard belongs to [#4087](https://github.com/bobmatnyc/trusty-tools/issues/4087).

Not addressed here: [#4708](https://github.com/bobmatnyc/trusty-tools/issues/4708) was triaged out of this release.

## Root cause

[#4871](https://github.com/bobmatnyc/trusty-tools/issues/4871) and [#4317](https://github.com/bobmatnyc/trusty-tools/issues/4317) are one root cause. `load_index_registry_at` mapped a TOML parse error to an empty list, so the next write did load-empty -> push-one -> save and published a single-entry file over the whole registry while reporting success.

Fixes: fail-closed parse, a process-wide lock held across load and save, per-write staging files replacing the shared `indexes.toml.tmp`, and an orphan reaper that removes by id instead of republishing a pre-boot snapshot.

## Caveat the reviewer must see

A prior session (2026-08-05) concluded [#4871](https://github.com/bobmatnyc/trusty-tools/issues/4871)'s stated premise — concurrent daemons reverting `indexes.toml` — was refuted, on the grounds that only one real writer daemon exists. This PR fixes [#4871](https://github.com/bobmatnyc/trusty-tools/issues/4871) anyway because two different defects were reproduced live and independently of that premise: a `rename indexes.toml -> No such file or directory` collision from the shared tmp file, and the fail-closed parse gap. Ask the reviewer to confirm the fix on the evidence actually reproduced, not on the issue's original framing.

## Known limitation

The fail-closed parse guarantees the write path only. Read-only callers (`reindex/runner.rs`, `warm_boot/mod.rs`, `server/tickers.rs`) still treat a corrupt registry as empty, exactly as before — no regression, but the guarantee does not extend to them.

## Test ladder

Rung 5 (persistence, process lifecycle, cross-crate contract):

- `cargo fmt --check`
- `cargo clippy -p trusty-search --all-targets -- -D warnings`
- `cargo test -p trusty-search` -> 1524 passed, 0 failed, 1 ignored, plus 372 integration
- `cargo check --workspace` -> EXIT=0
- `check_line_cap.sh` -> 3860 files, 4 allowlisted, 0 violations
- `check_changelog_fragment.sh` -> all crates recorded

The branch was rebased onto current `main` before opening this PR; the gate output above was produced at `ee45a1926` pre-rebase. Main's two intervening commits (`c2970687f` docs, `96123699e` trusty-mpm) touch only `docs/` and `trusty-mpm`, so the trusty-search gate result carries over unchanged — CI on the pushed head is the post-rebase confirmation.

## Pre-existing baseline failure, not caused by this branch

`commands::start::tests::swap_back_real_hardware_kills_sidecar_past_max_restarts` fails under `--include-ignored`. Reproduced at `origin/main` in a throwaway worktree; this diff touches zero files under `commands/`. Related closed issue [#3631](https://github.com/bobmatnyc/trusty-tools/issues/3631).

### DIFF
--- a/crates/trusty-search/changelog.d/4227-quarantine-doc-correction.md
+++ b/crates/trusty-search/changelog.d/4227-quarantine-doc-correction.md
@@ -1,0 +1,8 @@
+Changed
+
+- The write-quarantine module docs no longer list genuine corruption among the
+  conditions that trigger it. Corruption is absorbed before `corpus_open_failed`
+  can be set: `open_corpus_db_or_recreate` classifies it as recoverable, moves
+  the file aside, and returns a fresh empty corpus. Only lock contention and
+  transient I/O reach the quarantine, so its population is transient-dominated —
+  which the docs now say, along with the recreate-to-empty gap it does not cover.
--- a/crates/trusty-search/changelog.d/4253-transient-corpus-open-orphan.md
+++ b/crates/trusty-search/changelog.d/4253-transient-corpus-open-orphan.md
@@ -1,0 +1,9 @@
+Fixed
+
+- A cold index whose corpus open lost a race (`DatabaseAlreadyOpen`) or ran past
+  its deadline is no longer deregistered for the daemon's lifetime. The restore
+  ran to completion without registering a handle, so the loader consumed the
+  cold entry anyway and the index became a 404 with no way back — one client
+  retry against a slow ~20s cold-start open was enough. Registration is now the
+  only evidence a restore succeeded; a transient failure keeps the entry and
+  returns a retryable 503 so the next query recovers it.
--- a/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
+++ b/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
@@ -15,7 +15,10 @@ Fixed
   pre-boot snapshot of the survivors, so an index registered while the sweep
   was deciding is no longer erased by the cleanup.
 
-Known limitation: the fail-closed parse guarantees the WRITE path only. Read-only
-callers that swallow the load error (`reindex/runner.rs`, `warm_boot/mod.rs`,
-`server/tickers.rs`) still treat a corrupt registry as empty, exactly as before —
-no regression, but the guarantee does not extend to them. See #4871.
+Known limitation: the fail-closed parse guarantees the WRITE path only. Every
+read-only caller still swallows the load error and treats a corrupt registry as
+empty — `reindex/runner.rs`, `warm_boot/mod.rs`, `server/tickers.rs`,
+`reconcile.rs`, `server/indexes.rs`, `server/indexes_relocate.rs`,
+`server/index_config.rs`, `persistence_timestamps.rs`, and
+`commands/start/restore.rs`. That is unchanged behaviour, not a regression, but
+the guarantee does not extend to them. See #4871.
--- a/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
+++ b/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
@@ -14,3 +14,8 @@ Fixed
 - The boot orphan-reaper removes orphans by id instead of republishing a
   pre-boot snapshot of the survivors, so an index registered while the sweep
   was deciding is no longer erased by the cleanup.
+
+Known limitation: the fail-closed parse guarantees the WRITE path only. Read-only
+callers that swallow the load error (`reindex/runner.rs`, `warm_boot/mod.rs`,
+`server/tickers.rs`) still treat a corrupt registry as empty, exactly as before —
+no regression, but the guarantee does not extend to them. See #4871.
--- a/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
+++ b/crates/trusty-search/changelog.d/4871-4317-registry-write-integrity.md
@@ -1,0 +1,16 @@
+Fixed
+
+- An unparseable `indexes.toml` is now an error rather than an empty registry.
+  It previously read back as "no index was ever registered", and the next write
+  published that view — overwriting a whole registry with a single entry, which
+  is the mass-deregistration both #4317 (73 → 31, then 42 → 5) and #4871
+  recorded. The corrupt file is left intact for recovery.
+- Registry mutations are serialized process-wide and stage through a
+  per-write temporary file instead of one shared `indexes.toml.tmp`. Concurrent
+  writers previously interleaved: a registration landing between another task's
+  load and its save was silently discarded (the observed 80 → 88 → 80 revert),
+  and two writers racing on the shared temp file could publish a spliced,
+  unparseable registry.
+- The boot orphan-reaper removes orphans by id instead of republishing a
+  pre-boot snapshot of the survivors, so an index registered while the sweep
+  was deciding is no longer erased by the cleanup.
--- a/crates/trusty-search/changelog.d/4951-stale-indexer-root.md
+++ b/crates/trusty-search/changelog.d/4951-stale-indexer-root.md
@@ -1,0 +1,9 @@
+Fixed
+
+- `POST /indexes/:id/reindex` with a `root_path` override no longer makes every
+  search return nothing. The override rebuilt the index handle around the same
+  indexer, leaving the indexer's own `root_path` on the old value — so the
+  absolute `file` on each result was built against the old root while the search
+  post-filter (#64/#541) checked it against the new one, dropping 100% of
+  matches. Callers saw `results: []` with `stale_index_root: true` on an index
+  whose `/status` read `ready` with a full chunk count.
--- a/crates/trusty-search/changelog.d/5075-unregister-purges-cold-store.md
+++ b/crates/trusty-search/changelog.d/5075-unregister-purges-cold-store.md
@@ -1,0 +1,8 @@
+Fixed
+
+- `DELETE /indexes/:id` now clears the index's cold-store records. Deleting a
+  cold-parked or restore-failed index left them behind, so `/status`,
+  `/chunks`, and `grep` answered 503 forever for an id that no longer existed —
+  inverting #5057's rule that 404 means "absent from every store". A
+  cold-parked-only index is also removed from `indexes.toml` now, instead of
+  being resurrected by the next warm boot.
--- a/crates/trusty-search/src/core/indexer/mod.rs
+++ b/crates/trusty-search/src/core/indexer/mod.rs
@@ -474,6 +474,26 @@ impl CodeIndexer {
     /// incrementally before attaching all components.
     /// What: initialises all fields to their zero/empty states.
     /// Test: every test that constructs a `CodeIndexer` exercises this.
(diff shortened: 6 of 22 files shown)


bobmatnyc:
Addresses the review round: 2 HIGH, 5 MEDIUM, 3 LOW.

**HIGH-1 ([#4951](https://github.com/bobmatnyc/trusty-tools/issues/4951)):** reproduced, with a correction to the finding — the [#2178](https://github.com/bobmatnyc/trusty-tools/issues/2178) guard fires only when a persisted `indexes.toml` entry exists AND disagrees; `root_move_is_trusted(None, _)` returns true, so with no entry the walk runs and the failure mode is different (walk runs against the empty new root, prunes the corpus). Fix: `reindex_handler` now mirrors the runner's own gate and returns 409 naming `relocate`, rather than persisting the new root — persisting would make the override durable and defeat [#2178](https://github.com/bobmatnyc/trusty-tools/issues/2178) outright. The new test asserts the surviving result's `file` exists on disk.

**HIGH-2:** confirmed as a regression introduced by the earlier [#4253](https://github.com/bobmatnyc/trusty-tools/issues/4253) change. Both permanent restore arms now call `mark_failed`; the loader gained a `RestoreFailed` arm. Keep-the-entry now applies only to the transient path.

**A race disclosed rather than closed:** the HIGH-1 fix makes the handler's decision depend on `indexes.toml` at request time, so a registry write landing between the handler's check and the runner's re-check could still let an untrusted move through. The runner's gate is the backstop and the corpus is never walked or pruned, but the indexer root would have already moved, reproducing the divergence until the next successful reindex. Closing it properly means holding the registry write lock across both, which is wider than this round. Being tracked separately.

Gates after the round: `cargo fmt --check` + `clippy -p trusty-search --all-targets -D warnings` EXIT=0; `cargo test -p trusty-search` 1528 passed, 0 failed, 1 ignored, plus 372 integration; `cargo check --workspace` EXIT=0; line-cap 3865 files, 0 violations.

bobmatnyc:
**code-critic verdict: APPROVE.** Zero CRITICAL, zero HIGH above the confidence bar.

Both prior HIGH findings CLOSED: the #4951 root-override gate now mirrors the runner's own two predicates in the same order and refuses with 409 rather than persisting the override, and its test asserts the returned `file` exists on disk — an assertion that fails at the parent commit. All four terminal restore arms now `mark_failed`, with the transient corpus-open arm correctly left retryable.

The disclosed handler/runner registry race was judged **not a ship blocker**: reaching it needs a concurrent write of this index's `root_path` between two reads, it fails loudly when it fires (`mark_reindex_failed`, fatal SSE error, quarantine), and no data is lost because the runner aborts before the walk and the prune. The branch strictly narrows a window that was previously certain for that request shape.

Follow-up work identified and not blocking: move the `set_root_path` sync into the runner so one registry read drives both decision and mutation; the new trust gate degrades to "trusted" on a registry read error; and `commands/prune.rs`, `prune_orphans.rs` and `migrate_storage` still publish the whole registry from a pre-decision snapshot in a separate process, where the in-process lock gives no protection.

Issue #5337, opened 2026-08-10
trusty-mpm: BASE_PM index_id guidance contradicts trusty-search's pinned-index behavior


bobmatnyc:
#5213 made trusty-search's own MCP tool descriptions state the pinned-first resolution order and point an unresolvable `index_id` at `list_indexes`. The matching guidance in trusty-mpm's BASE_PM instructions still describes the old behavior, so the contradiction persists on the agent-facing side. This is #5213's second closure condition and lives outside the trusty-search crate.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #5213, opened 2026-08-08
fix(trusty-search): index_id session-pin default masks search failure instead of erroring loudly


bobmatnyc:
**Defect.** `search`, `search_all`, `index_status`, and `grep` on `mcp__trusty-search__*` all default an omitted `index_id` to the session-pinned index. When that pinned index has not been built yet, the call fails with `INDEX_NOT_READY` and tells the caller to "fall back to grep/find" — which is itself index-backed and reports the same failure. The loss of search capability is masked as a graceful fallback instead of surfacing as an error; `tm doctor` does not catch it either.

**Contract self-contradiction (Reproduced, this session).** `search_all`'s TOOL description says omitting `index_id` "falls back to legacy cross-project fan-out behaviour (issue #10)"; its `index_id` PARAMETER description says omission instead defaults to the session pin and fails with `INDEX_NOT_READY` if unbuilt. Both cannot hold — empirically the pin wins and the documented fan-out path is unreachable. (Note: the docstring's "issue #10" reference appears stale — #10 is an unrelated closed ticket about a `tga` engineer-config migration, not fan-out.)

**BASE_PM steers into the broken path.** `BASE_PM.md` "Trusty Tool Priority" tells every session to *omit* `index_id` because a guessed id 404s (#1373, confirmed: that issue is about resolving to the wrong index, e.g. `claude-mpm`, and its fix was register-and-pin). That guidance was correct for the guessed-id-404 failure mode but now drives every caller into the one path that cannot work when the pin is unbuilt. Whichever fix is chosen below, the tool docstrings and the BASE_PM instruction must change together — fixing one alone leaves them contradicting.

**Owner ruling (Bob, 2026-08-08, verbatim):** "We need to make index ID required for each" / "(or fan out)". Two shapes, not mutually exclusive:
- (a) Require `index_id` — an absent/unknown id is a loud, immediate error, never a silent wrong default. Callers use `list_indexes` to discover valid ids.
- (b) Fan out when omitted — restore the documented issue-#10-style behaviour, searching every index matching repo identity and merging. `search_all` already carries `max_fanout_concurrency`/`serial` knobs from #2845 for this.
- A hybrid (required by default, explicit opt-in fan-out flag) is a third shape.

**Related, not duplicate.**
[...]

Issue #5061, opened 2026-08-07
index_status_handler lacks cold-store fallback, returning 404 for parked-but-existing indexes


bobmatnyc:
**Defect:** `index_status_handler` in trusty-search's `service/server/status.rs` looks up an index via a raw hot-registry lookup with no cold-store fallback, unlike `search_handler`, which transparently reloads through `get_or_load_index`. When `TRUSTY_MAX_RESIDENT_INDEXES` is enabled (#2161, off by default), an index that exists on disk but has been cold-parked returns 404 from `/status`. `search_semantic` and `search_kg` inherit this through their Stage 2/3 pre-flight check against the same endpoint. `list_chunks` shares the same shape via `files.rs`.

**Evidence:** Found during the code-critic review of PR #5057. This matters more after #5057 lands: that PR classifies a 404 on the session's pinned index as `INDEX_NOT_READY` with `state: "not_indexed"`, so a cold-parked index would be reported as never built — false for an index that was built and merely needs a reload. The misclassification is a lie in the opposite direction from the one #5057 fixes.

**Resolution:** Give `index_status_handler` (and the `list_chunks` path) the same cold-store fallback `search_handler` uses, so a parked-but-existing index reloads rather than 404s.

Severity bounded: `TRUSTY_MAX_RESIDENT_INDEXES` is off by default, and #5057 is not blocked on this fix. Related: #2161 (resident-index cap), #4715 (bare search 404s on unindexed default).

bobmatnyc:
## Update from #5057 round-3 review — symptom changed, scope widened, reclassified to MEDIUM

PR #5057's round-2 follow-up (landed on top of this issue) gave `index_status_handler` (`status.rs:103-110`), `get_index_chunks_handler` (`files.rs:132-138`), and the grep handler (`files.rs:274-294`) the cold-store/failed-set check this issue asked for. So the original symptom — a bare 404 for a parked-but-existing index — is fixed; the title is now stale on that point.

What's left, per the round-3 critic (PR #5057 at head `e5c246033`): https://github.com/bobmatnyc/trusty-tools/pull/5057#pullrequestreview-4880053803

- `index_status`, `grep`, and `chunks` now report cold state truthfully (503, not 404) but — unlike `search_handler:195-240` — cannot *reload* a cold-parked index. They stay 503 until something else restores it, while a plain `search` on the same index self-heals.
[...]

Issue #4839, opened 2026-08-04
/health warmboot_summary.indexes_loaded counts registered indexes, not populated ones — a fully-empty deployment reports healthy


bob-duetto:
### Summary

`/health`'s `warmboot_summary.indexes_loaded` counts index *registrations*, not indexes that actually contain data. A deployment where nearly every index is empty reports `status: ok` with a full loaded count, giving operators no signal that search is degraded.

### Observed on a live production deployment

trusty-search **0.39.0**, green serving box, 2026-08-01 through 2026-08-04:

```
status: ok
warm_boot_degraded: false
indexes_loaded: 222/222
indexes_skipped_timeout: 0
indexes_failed: 0
indexes_corpus_failed: 0
```

Every signal nominal. In reality **only 2 of those 222 indexes held any data** at the time (`main` at 315,423 chunks and `apex` at 1,584 — the latter only after a manual repair). The other 220 sat at `chunk_count=0`, `last_walk_files_seen=0`, with stages frozen at `lexical: in_progress` / `semantic: pending` / `graph: pending`.

A per-index `GET /indexes/<id>/status` also reports top-level `status: "ready"` for these — "ready" reflects registration, not population, so the per-index surface has the same gap.

### Impact

This is how the condition in #4680 (indexes stuck and never re-driven) stayed invisible. Downstream, an application consuming these indexes returned empty context on **913 consecutive operations across 44 days** with no health signal, no error, and no alert — because by every metric the daemon exposed, it was healthy. The failure was only discovered by directly comparing `chunk_count` values across indexes.

An operator watching `/health` has no way to distinguish a healthy deployment from one serving nothing.

### Suggested direction (not prescriptive)

Distinguish registered from populated in the health surface. Some options:
- Split the counter — e.g. `indexes_registered` vs `indexes_populated` (`chunk_count > 0`).
- Report an aggregate `total_chunks` so a near-zero corpus is visible at a glance.
- Consider whether a deployment with a large registered/populated gap should report `degraded` rather than `ok`, or at least surface the gap as a distinct field.
- Per-index `status` might likewise distinguish "registered" from "ready to serve results."

### Related

- #4680 — non-main indexes stuck at `chunk_count=0`, never re-driven across restarts. This issue is why that one went unnoticed for so long.
- #4087 — corpus-failed indexes return silently-empty 200s. Same family of problem: a failure state that is indistinguishable from a legitimately empty result.

### Additional context from the same deployment

[...]

PR #5345, opened 2026-08-10
fix(trusty-search): make a degraded index say so on every surface (#5068, #5061, #5213, #4787, #4839, #4706)

### LINKED ISSUE #5068: trusty-search: semantic search against a skip_vector index returns degraded BM25 results with HTTP 200, not a distinguishable error
## Defect

Forcing `stage: semantic` against a live `skip_vector` index returns **HTTP 200 with BM25 results** — not empty, not an error. The caller cannot tell it asked for semantic search and got lexical instead. The only signal is `meta.search_capabilities` omitting `vector`; there is no `vector_unavailable` flag beside the existing `bm25_lane_degraded`.

## Evidence

Probed against a live index during [#5065](https://github.com/bobmatnyc/trusty-tools/pull/5065). This matters more now that worktree indexes are built with `skip_vector` by owner ruling ([#5060](https://github.com/bobmatnyc/trusty-tools/issues/5060)) — worktrees carry BM25+KG only, embeddings live on the tm checkout.

## Resolution

Mirror the precedent already in this crate — `search_kg` returns `503 kg_unavailable` for `skip_kg` indexes. Semantic against a vector-less index should return the analogous distinguishable error rather than degraded results.

bobmatnyc:
Closes [#5068](https://github.com/bobmatnyc/trusty-tools/issues/5068), [#5061](https://github.com/bobmatnyc/trusty-tools/issues/5061), [#5213](https://github.com/bobmatnyc/trusty-tools/issues/5213), [#4787](https://github.com/bobmatnyc/trusty-tools/issues/4787), [#4839](https://github.com/bobmatnyc/trusty-tools/issues/4839), [#4706](https://github.com/bobmatnyc/trusty-tools/issues/4706).

**Test ladder rung 6** — MCP tool schemas and HTTP route responses. Gate run: `cargo fmt --check -p trusty-search` + `cargo clippy -p trusty-search --all-targets -- -D warnings` + `cargo test -p trusty-search` → 1532 passed, 0 failed, 1 ignored, plus 372 integration across 27 targets; `check_line_cap.sh` 3860 files, 4 allowlisted, 0 violations; `check_changelog_fragment.sh` 26 paths, all crates recorded; `check_sld.sh` 0 errors.

**The through-line:** these six are one contract problem — the daemon reported success, or a plausible-looking number, when the real answer was "degraded", "parked", or "not populated". The fix centralizes the verdict in one module so the same daemon state renders identically on every endpoint: corpus-unavailable, not-resident (with `restore_via`), restore-failed, absent, vector-off-by-config, and vector-not-yet-built, each with its own error code and a retryable flag.

🔴 **Two deliberate breaking changes**, both the fix rather than a side effect of it:

1. `POST /indexes/:id/search` with `"stage": "semantic"` against an index whose vector lane is unavailable now returns `503 vector_unavailable` where it returned `200` with BM25 rows ([#5068](https://github.com/bobmatnyc/trusty-tools/issues/5068)). A client that treated the degraded `200` as success will now see an error.
2. `POST /indexes/:id/index-file` and `POST /indexes/:id/remove-file` now return `503 index_not_resident` where they returned `404` for a cold-parked index ([#5061](https://github.com/bobmatnyc/trusty-tools/issues/5061)). A client that treated the `404` as "index gone" and stopped pushing will now correctly retry.

Both stay within the `0.44.0` MINOR bump; a second break does not change it.

`GET /indexes/:id/status` and `GET /indexes/:id/chunks` now return a JSON body on `404`/`503` where they returned none — additive at the wire level, since a client parsing an empty body as JSON was already failing.

Everything else in the PR is additive: new `meta` fields, a new `semantic_coverage` block, three new `/health` counters, and JSON bodies on 404/503/500 responses that previously carried none.

**[#4706](https://github.com/bobmatnyc/trusty-tools/issues/4706) covered more than the issue asked.** The same helper's `last_indexed` half had the identical defect — it probed only the legacy global dir, so a colocated index reported `null` freshness. Fixed and disclosed rather than left as half a defect. Field names and types are unchanged; only the values are corrected.

**Live API evidence was captured** on isolated daemons with residency forced to zero, showing the pre-fix and post-fix behavior for the residency 503s and for `size_bytes` going from 0 to a byte count matching `du`. Say it exists and was verified; do not paste the captures.

**For whoever sequences the merge:** this branch touched only `search_all` descriptor strings in `mcp/tools/descriptors.rs` and changed nothing in `mcp/tools/misc.rs`. The sibling `fix/ts-quick-fixes` branch touches `grep`'s descriptor and `misc.rs`. The overlap is file-level, not string-level.

Note that [#5213](https://github.com/bobmatnyc/trusty-tools/issues/5213)'s second closure condition lives outside this crate and is tracked separately as [#5337](https://github.com/bobmatnyc/trusty-tools/issues/5337).


bobmatnyc:
code-critic returned **WARN**; its one HIGH is now fixed.

**The HIGH:** `search_handler` still hand-rolled its residency bodies while `degraded.rs`'s docstring claimed no endpoint could drift — so the one endpoint every other body points at via `restore_via` was the one not honouring the contract. A caller following that hint got `index_restore_failed` with `retryable` **absent**, which is undefined under the new contract. All three arms now route through the shared builder, and `retryable`/`index_id` were added to `index_loading` and `embedder_initializing` for the same reason.

**A second fail-open found and closed while fixing the first:** `health.rs` swallowed a corpus read error and fell back to the in-memory chunk map, which reads 0 after idle eviction — so an unreadable corpus on a populated index would have been counted as **empty**, manufacturing the exact false signal [#4839](https://github.com/bobmatnyc/trusty-tools/issues/4839) exists to remove.

**Three tests did not test what their names claimed** and were rewritten: the `/health` counters exercised only the fallback path; `last_indexed_takes_the_newer_of_the_two_layouts` created only one layout; and `vectors_present: null` conflated "no vector store" with "read failed" — now split by an explicit `vectors_unavailable_reason`.

**A per-query cost regression was found and removed:** [#4706](https://github.com/bobmatnyc/trusty-tools/issues/4706) had doubled a blocking directory walk on the search path and discarded the result. Split into stat-only and full helpers, with a test pinning that the two cannot drift.

`crates/trusty-search/CLAUDE.md` gained the index-scoped error contract — a table of `error` / status / `retryable`, the `restore_via` mechanism, and the new `/health` counters.

Live API evidence was re-captured end to end on an isolated daemon: the residency 503 with its hint, the hint proving actionable, and the [#5068](https://github.com/bobmatnyc/trusty-tools/issues/5068) guard staying narrow on a healthy vector lane.

Gates: `cargo fmt --check` + `clippy --all-targets -D warnings` + `cargo test -p trusty-search` EXIT=0, 1538 lib passed / 0 failed, 372 integration; line-cap 3869 files 0 violations; SLD 0 errors; changelog fragment gate 29 paths, all crates recorded.

bobmatnyc:
**code-critic verdict: WARN** — one HIGH, now fixed.

The HIGH: `search_handler` still hand-rolled its residency bodies while `degraded.rs`'s docstring claimed no endpoint could drift — so the endpoint every other body points at via `restore_via` was the one not honouring the contract. A caller following that hint got `index_restore_failed` with `retryable` **absent**, undefined under the new contract. All three arms now route through the shared builder.

A second fail-open found and closed while fixing the first: `health.rs` swallowed a corpus read error and fell back to the in-memory chunk map, which reads 0 after idle eviction — an unreadable corpus on a populated index would have counted as **empty**, manufacturing the exact false signal #4839 exists to remove.

Three tests did not test what their names claimed and were rewritten; a per-query blocking directory walk introduced by #4706 was found doubled and removed.

Two follow-ups promoted to their own issues rather than fixed here: #5349 and #5350.

PR #5360, opened 2026-08-10
fix(trusty-common,trusty-memory): key KG triples on (subject, predicate, object) so membership stops overwriting (#4810)

### LINKED ISSUE #4810: bug(trusty-common): KG membership triples silently overwrite — room:/tag:/topic: subjects hold 1 triple regardless of member count
## Defect

The redb triple key is `(subject, predicate)` only
(`crates/trusty-common/src/memory_core/store/kg_store.rs:189-197`), so each
new assert of the same `(subject, predicate)` pair **overwrites the
previous one**. `kg_extract.rs:229-232` asserts `room:<label> contains
drawer:<id>` with a comment claiming *"multiple drawers per room must
coexist"* — they do not.

**Verified live** (2026-08-04, `trusty-tools` palace, trusty-memory v0.22.0,
daemon `127.0.0.1:7070`): `kg_query(subject="room:General")` returns
**exactly 1 triple** against **779 drawers** in that room. `tag:mvp`
likewise returns 1 triple regardless of how many drawers actually carry that
tag.

This silently collapses `tag:*` and `topic:*` membership too — any subject
of the form `<prefix>:<label> contains drawer:<id>` is affected, not just
`room:`. Meaning: **the KG can prove a room/tag/topic *existed* (the subject
is present), but cannot enumerate its members** — every assert after the
first is a silent data loss, not an error, not a warning.

## Severity

High. This is a silent correctness defect with no error signal anywhere in
the write path — a caller has no way to discover, short of manually
comparing triple count to expected membership, that only the most recent
member survived.

## Why this matters right now

ADR-0027 (`docs/adr/0027-rooms-are-real-wings-are-scopes-closets-are-an-index.md`,
branch `design-wings-rooms`, commit `c5c3c20a`) is why this defect was
found and is the reason it must be tracked: this is exactly why the ADR
requires a real `ROOMS` table (redb, keyed by room id) rather than reading
room membership from the KG's `room:` subjects — the KG's `room:*` triples
are a lossy shadow, not an index, and were never a safe source of truth for
"which drawers are in this room." ADR-0027 §C3.3 / §D-1 records this
finding but explicitly defers the fix to this ticket, filed separately.

## Cross-reference

Same audit that found #4775 (the `kg_query` hint defect) found this. Both
surfaced from KG interaction correctness work on 2026-08-04.

## Reference

ADR-0027 §C3.3, §D-1 (`docs/adr/0027-rooms-are-real-wings-are-scopes-closets-are-an-index.md`,
branch `design-wings-rooms`, commit `c5c3c20a`).

bobmatnyc:
## What

The redb `TRIPLES` key gains the object component, so `room:General --contains--> drawer:X` stops collapsing to one row. `FUNCTIONAL_PREDICATES` is an inverted list: 15 named single-valued predicates keep supersession, everything else defaults to multi-valued. Closes #4810.

## Why inverted

Forgetting a multi-valued predicate costs silent data loss (this bug); forgetting a functional one costs an extra queryable row. `has_language` is deliberately excluded because `scan_cargo_toml` and `scan_package_json` derive subjects independently.

## Also in scope

`kg/adjacency.rs` `upsert_edge` was collapsing parallel edges in memory regardless of the storage fix; it now filters on target as well as predicate. Without it `/kg/graph` and `kg_gaps` would keep under-reporting.

## Migration

At-open, per-palace, gated by a new `KG_SCHEMA` marker, one redb transaction, backup written to `kg.redb.pre-4810.bak` (note: `.redb`, not `.db` — `KnowledgeGraph::open` translates the path via `redb_path_for`).

## 🔴 Reviewer must weigh: fail-open leaves triples INVISIBLE, not un-migrated

Readers decode the three-component key, so a palace whose migration failed opens cleanly but returns nothing from `query_active`/`list_active` until a later open succeeds. Rows are intact on disk and the next open retries. Making an un-migrated palace readable would need a legacy-decode fallback in every reader — a different design, deliberately not improvised here. This is the main thing to review.

## 🔴 Two existing tests were changed — read this before approving

`knowledge_gaps_on_sparse_graph` and `suggested_exploration_is_non_empty` built a "hub with 6 leaves" using the multi-valued predicate `rel`, so pre-fix they were really testing a hub and ONE edge. On a real 6-leaf star internal density is 6/21 ≈ 0.29, above the 0.2 gap threshold, so no gap is emitted. Both moved to 12 leaves (12/78 ≈ 0.15) with the arithmetic written into the test. The classifier is unchanged. Flagging explicitly because "changed a test to make it pass" is normally a red flag.

## Downgrade hazard

Do not run a `trusty-memory` older than this against a palace this version has opened. The on-disk key format changed and an old binary computes the old key, matching nothing.

## Gate — rung 5

```
cargo fmt --check                                                  → clean
cargo check -p trusty-common --features memory-core --all-targets  → clean
cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings → clean
cargo test -p trusty-common --features memory-core → ok. 968 passed; 0 failed; 13 ignored
  --test kg_triple_key_migration_tests -- --include-ignored → ok. 1 passed; 0 failed
cargo check --workspace   → clean
cargo test -p trusty-memory → ok. 629 passed; 0 failed; 4 ignored
cargo check -p trusty-agents → clean
bash scripts/check_line_cap.sh → measured 3871 tracked .rs file(s); 4 allowlisted, 0 violations — OK
bash scripts/check_sld.sh → scanned 57 spec doc(s) + 5821 code file(s); 0 error(s), 0 warning(s)
```

## Feature-flag trap worth knowing

Bare `cargo test -p trusty-common` does NOT compile the KG. `memory-core` is not a default feature, so the bare command runs 304 unrelated tests and proves nothing about this change. Every gate above used `--features memory-core`. CI is unaffected because `cargo test --workspace` unifies the feature via trusty-memory.

## Pre-existing failure, not caused here

`cargo test -p trusty-mpm` fails `create_managed_session_confirms_server_before_applying_options` (contends over a live tmux server under parallel execution). `git diff --name-only e2ca949a3...HEAD -- crates/trusty-mpm/` is empty, and the test passes in isolation: `cargo test -p trusty-mpm --lib -- --test-threads=1 create_managed_session_confirms_server_before_applying_options` → ok. 2 passed; 0 failed.

## Changelog fragments

Both `crates/trusty-common/changelog.d/` and `crates/trusty-memory/changelog.d/`.

### DIFF
--- a/crates/trusty-common/changelog.d/4810-kg-triple-key.md
+++ b/crates/trusty-common/changelog.d/4810-kg-triple-key.md
@@ -1,18 +1,9 @@
 Fixed
 
-The knowledge-graph TRIPLES key now includes the object, so a subject can hold
-several objects under one predicate. It previously keyed on
-`(subject, predicate)` alone, which meant `room:General --contains--> drawer:X`
-held ONE row no matter how many drawers the room had — every new member
-silently closed the last one. Predicates listed in
-`kg_store::FUNCTIONAL_PREDICATES` (`is_alias_for`, `has_version`, …) keep the
-one-active-object rule; every other predicate is multi-valued. `upsert_edge`
-in the in-memory adjacency was collapsing the same edges back on every palace
-open and now matches on target as well as predicate.
-
-**Breaking on-disk format.** Palaces are migrated in place at open, once, in a
-single redb transaction, after a size-verified copy of the database is written
-to `<kg>.redb.pre-4810.bak`. A migrated palace cannot be read by an earlier
-version of this crate — downgrading below this version is unsafe. A migration
-that fails is logged at `warn!` and retried on the next open; the palace still
-opens, but its triples do not appear in queries until it succeeds.
+- The knowledge-graph TRIPLES key now includes the object, so a subject can hold several objects under one predicate ([#4810](https://github.com/bobmatnyc/trusty-tools/issues/4810))
+  - It previously keyed on `(subject, predicate)` alone, which meant `room:General --contains--> drawer:X` held ONE row no matter how many drawers the room had — every new member silently closed the last one.
+  - Predicates listed in `kg_store::FUNCTIONAL_PREDICATES` (`is_alias_for`, `has_version`, …) keep the one-active-object rule; every other predicate is multi-valued.
+  - `upsert_edge` in the in-memory adjacency was collapsing the same edges back on every palace open and now matches on target as well as predicate.
+- **Breaking on-disk format.** Palaces are migrated in place at open, once, in a single redb transaction, after a size-verified copy of the database is written to `<kg>.redb.pre-4810.bak`.
+  - A migrated palace cannot be read by an earlier version of this crate — downgrading below this version is unsafe.
+  - A migration that fails is logged at `warn!` and retried on the next open; the palace still opens, but its triples do not appear in queries until it succeeds.
--- a/crates/trusty-common/changelog.d/4810-kg-triple-key.md
+++ b/crates/trusty-common/changelog.d/4810-kg-triple-key.md
@@ -1,0 +1,18 @@
+Fixed
+
+The knowledge-graph TRIPLES key now includes the object, so a subject can hold
+several objects under one predicate. It previously keyed on
+`(subject, predicate)` alone, which meant `room:General --contains--> drawer:X`
+held ONE row no matter how many drawers the room had — every new member
+silently closed the last one. Predicates listed in
+`kg_store::FUNCTIONAL_PREDICATES` (`is_alias_for`, `has_version`, …) keep the
+one-active-object rule; every other predicate is multi-valued. `upsert_edge`
+in the in-memory adjacency was collapsing the same edges back on every palace
+open and now matches on target as well as predicate.
+
+**Breaking on-disk format.** Palaces are migrated in place at open, once, in a
+single redb transaction, after a size-verified copy of the database is written
+to `<kg>.redb.pre-4810.bak`. A migrated palace cannot be read by an earlier
+version of this crate — downgrading below this version is unsafe. A migration
+that fails is logged at `warn!` and retried on the next open; the palace still
+opens, but its triples do not appear in queries until it succeeds.
--- a/crates/trusty-common/src/memory_core/store/kg/adjacency.rs
+++ b/crates/trusty-common/src/memory_core/store/kg/adjacency.rs
@@ -16,6 +16,7 @@ use petgraph::visit::EdgeRef;
 use std::collections::HashMap;
 
 use crate::memory_core::store::kg_redb::KgStoreRedb;
+use crate::memory_core::store::kg_store::is_functional_predicate;
 
 /// In-memory adjacency cache backing the public graph API.
 ///
@@ -61,21 +62,28 @@ impl Adjacency {
     }
 
     /// Why: `assert` must keep the graph in sync with the store; doing it
-    /// here keeps the lock-management in one place.
-    /// What: Removes any prior edge for `(subject, predicate)` between the
-    /// existing subject and object nodes, then inserts the new edge using
-    /// the provided triple's metadata. Nodes are created if absent.
-    /// Test: `assert_adds_edge`, `retract_removes_edge`.
+    /// here keeps the lock-management in one place. #4810: the removal used to
+    /// drop EVERY edge from the subject carrying this predicate, regardless of
+    /// target. Since `hydrate_adjacency` replays each active triple through
+    /// here, a room's parallel `contains` edges collapsed back to one on every
+    /// palace open even once storage kept them — which is what left
+    /// `/kg/graph`, `kg_gaps`, and `expand_neighbors` under-reporting.
+    /// What: mirrors the storage split. For a functional predicate the prior
+    /// edge to any target is removed (one active object per subject); for a
+    /// multi-valued one only the edge to THIS target is, so parallel edges to
+    /// other targets survive. Nodes are created if absent.
+    /// Test: `assert_adds_edge`, `retract_removes_edge`,
+    /// `parallel_edges_survive_hydration`.
     pub(super) fn upsert_edge(&mut self, triple: &Triple) {
         let s_idx = self.ensure_node(&triple.subject);
         let o_idx = self.ensure_node(&triple.object);
-        // Remove any existing edge with the same predicate between the
-        // existing subject and any object (matches the temporal invariant
-        // "at most one active edge per (subject, predicate)").
+        let functional = is_functional_predicate(&triple.predicate);
         let to_remove: Vec<_> = self
             .graph
             .edges(s_idx)
-            .filter(|e| e.weight().predicate == triple.predicate)
+            .filter(|e| {
+                e.weight().predicate == triple.predicate && (functional || e.target() == o_idx)
+            })
             .map(|e| e.id())
             .collect();
         for eid in to_remove {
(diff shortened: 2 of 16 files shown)


bobmatnyc:
## Verdict: WARN

Rung 5 review of #4810's KG triple-key widening (commits a420f0df2, 7a629c768; base e2ca949a3). Verified independently: `cargo test -p trusty-common --features memory-core` → 968 passed/0 failed/13 ignored (matches PR claim); the four `migration_*` tests including the fail-open regression test pass in isolation; `cargo test -p trusty-common --features memory-core --test kg_triple_key_migration_tests -- --include-ignored` → 1 passed; `community_tests` → 6 passed; `cargo check --workspace` + `cargo test -p trusty-memory` (629+ passed) + `cargo check -p trusty-agents` → all clean; `cargo clippy -p trusty-common --features memory-core --all-targets -- -D warnings` → clean; `bash scripts/check_line_cap.sh` → 0 violations. `bash scripts/check_semver.sh --crate trusty-common` timed out in my sandbox after 180s mid-rustdoc-build (EXIT=124) — inconclusive, see finding #2 for what a manual read of the diff found instead.

## Findings

| HIGH | crates/trusty-memory/src/kg_extract.rs | 313-317 | `PATTERN_TABLE`'s `is-a`/`works-at`/`uses`/`depends-on` predicates are absent from `trusty-common`'s `FUNCTIONAL_PREDICATES`, so they default to multi-valued. Every `memory_remember` call that matches one of these markers now permanently accumulates a new active triple — pre-#4810 the same `(subject, predicate)` key silently *overwrote* the prior row on every repeat, which accidentally capped growth; #4810 correctly removes that overwrite-as-corruption behavior but puts nothing in its place for these four predicates. There is no TTL/pruning path for `auto:remember`-provenance KG triples anywhere in the crate (checked `store/kg`, `kg_extract.rs`, and grepped for `prune`/`ttl`/`expire`/`dream` near the KG). Confirmed the mechanism: `subject_tok`/`object_tok` are extracted per-drawer from free text (`extract_patterns`, kg_extract.rs:329-348), so common subject tokens recur across many drawers with different objects and never get retracted. This is a real, unbounded-growth regression exposed (not caused) by the correctness fix, on a write path that runs on every `memory_remember` across the ~94 live palaces this migration will touch. | Either fold `is-a`/`works-at` into `FUNCTIONAL_PREDICATES` (both are naturally close to single-valued in practice) and/or give `auto:remember`-provenance triples a pruning or TTL policy so heuristic extraction can't grow a palace unbounded. This is a design call beyond a one-line fix, and orthogonal to the storage-correctness bug this PR fixes. | Promote |
| MEDIUM | crates/trusty-common/src/memory_core/store/kg_store.rs | 391, 409, 479 | PR body's SemVer framing ("`is_functional_predicate`/`FUNCTIONAL_PREDICATES` are new public API... Additive, so no SemVer break claimed") only covers the additions. `encode_triple_key` (391, `pub fn`, 2 args → 3 args), `decode_triple_key` (409, `pub fn`, returns `(String,String)` → `(String,String,String)`), and `encode_predicate_index_key` (479, `pub fn`, 2 args → 3 args) are pre-existing public functions with **changed signatures**, reachable from the crate root (`lib.rs:299 pub mod memory_core` → `store/mod.rs:16 pub mod kg_store`). That's a breaking change, not additive, under Cargo's 0.x rule. Per this repo's policy the SemVer gate runs at release time (`preflight-publish.sh` CHECK 5), not per-PR, so this doesn't block merge — my own `check_semver.sh --crate trusty-common` run timed out before producing a verdict either way, so I can't independently confirm cargo-semver-checks catches it, only that the source-level signature change is real. | No fix needed in this PR. Worth a note so whoever cuts the next `trusty-common` release version isn't surprised by a required MINOR bump. | Parent |

## Notes

- **Fail-open (attack #1), judged acceptable.** Ran `migration_failure_leaves_the_palace_openable_and_retries` directly — it forces the failure branch (blocks the `.bak` path with a directory so `fs::copy` fails), and confirms: the palace still opens and is fully writable (not degraded), the legacy rows are provably untouched on disk, `query_active` returns empty for the un-migrated subject exactly as documented (not silently wrong data), and a second open with the blocker removed retries and migrates successfully — retry is real, not just claimed. The degraded window requires a rare trigger (full disk / unwritable directory) and self-heals on the next successful open, which for a daemon-owned palace is every restart. The one gap is observability — nothing aggregates "how many palaces are stuck un-migrated" beyond a per-palace `warn!` — but that's a monitoring gap, not silent data loss, and not something I'd block on.
- **Community-density arithmetic (attack #2), verified correct.** 6 leaves: 6/21 ≈ 0.286 (above the 0.2 threshold in `community.rs:115`, unchanged by this PR); 12 leaves: 12/78 ≈ 0.154 (below). `community.rs` itself is not in the diff — confirms the classifier is untouched and the two tests were adjusted to restore the sparse-star topology the multi-valued fix now allows, not to paper over a regression.
- **assert/retract → batch_assert/batch_retract (attack #3), byte-level diff-checked.** The single-op methods previously carried their own copy of the close-prior/insert-new logic; `apply_batch` already had a separate `batch_assert`/`batch_retract`. This PR deletes the duplicate and routes both through the batch functions. Confirmed via `git diff e2ca949a3..HEAD` that the unified logic preserves the same three-table (`TRIPLES`/`TRIPLES_BY_OBJECT`/`TRIPLES_BY_PREDICATE`) + `ACTIVE_SUBJECT_COUNTS` mutation shape the old single-op code had, now correctly widened to a range scan (`active_rows_for_pair`) instead of a point read.
- **Migration atomicity/backup (attack #4), read and test-verified.** Marker write + row rewrite ride one `wtx`; a mid-loop error drops `wtx` without `commit()`, which redb rolls back structurally (not application logic that could have a bug). `ensure_verified_backup` reuses a size-matching `.bak` and replaces a truncated one — both paths have dedicated tests that pass.
- **Key encoding (attack #5), no collision found.** `encode_triple_key`'s explicit `u16` length prefixes on subject and predicate make the subject/predicate boundary and predicate/object boundary unambiguous; object is the tail with no further internal structure to misparse. The `s.len() as u16` truncation risk for a >64KB subject is pre-existing (the old legacy key had the same cast) and out of this PR's scope.
- **delete_by_subject / BTreeSet dedup (attack #7), correct.** Pre-fix, `(subject, predicate)` pairs were already unique per key so a `Vec` and a `BTreeSet` behaved identically; post-fix, multiple rows can share a pair, and the `BTreeSet` dedup is what keeps `delete_by_subject` from calling `retract` N times over the same pair after the first call already closed it. Covered by `cascade_delete_closes_every_object_of_a_multivalued_pair`, which I ran and confirmed passing.
- No `unwrap()` found in any production code path touched by this diff (only in `#[cfg(test)]` bodies and one pre-existing, unmodified `.expect("db_cache poisoned")` mutex-poison invariant in `store.rs`).
- Two uncommitted files in this worktree (`.line-cap-allowlist.tsv`, `crates/trusty-memory/ui/package-lock.json`) are outside the `a420f0df2..7a629c768` commit range and not part of this review.

bobmatnyc:
Rebased onto fad89dc58 to pick up trusty-common 0.30.1 from main. Clean rebase, no conflicts.

The PR version bump vs crates.io failure is resolved by the rebase, not by a bump:
```
Merge base: fad89dc586f9d1a4ef592c36791048ffefdaaa11
[ OK ] trusty-common 0.30.1 — src changed, version not yet published

check-pr-version-bump: clear (0 warning(s)).
```

The gate compares the declared version against crates.io, and 0.30.1 is not published (latest published trusty-common is 0.30.0), so there is nothing to compare against. No version bump and no dependency-pin changes are in this PR.

The two earlier gate failures are also fixed: changelog fragments now carry bullets under their category line, and the Test: doc pointers in migrate.rs now cite the real test name `migration_failure_leaves_the_palace_openable_and_retries`.

Issue #4940, opened 2026-08-05
A poisoned ENV_LOCK turns one embedder failure into seven, and the CI model cache never survives a failing job


bobmatnyc:
Two defects, one root cause chain, fixed together.

## Defect 1 — `ENV_LOCK` poison cascade

Every env-touching test in `crates/trusty-common/src/embedder/` acquires the shared lock with `ENV_LOCK.lock().unwrap()`. `ENV_LOCK` is a `Mutex<()>`. One test panicking while holding the guard poisons it, so every later acquisition panics with `PoisonError`.

When `default_model_matches_sentence_transformers_reference` fails, six `resolve_*` tests fail as casualties. Those six test model/provider/cache-dir resolution — pure logic, no fastembed — and cannot fail on their own code. They bury the real cause.

## Defect 2 — the CI model-cache ratchet

Three compounding problems in `.github/workflows/ci.yml`'s `test` job:

- **Key hashes the whole lockfile.** `key: fastembed-v${{ hashFiles('Cargo.lock') }}-…`, while its comment claims "fastembed crate version (from Cargo.lock)". Any dependency bump anywhere in the workspace rotates the key, so the cache is almost always cold.
- **`actions/cache@v4` declares `post-if: "success()"`.** The save step never runs on a failing job.
- **75-second retry budget** (5 attempts at 5/10/20/40s) against HuggingFace 429s, with `HF_TOKEN` empty so requests are anonymous at the lowest rate limit.

Together they ratchet: HF rate-limits the pre-seed → no model → the accuracy gate fails → the job fails → the cache is never saved → the next run is cold and rate-limited again.

The pre-seed is fail-soft, so it reports step `success` having downloaded nothing, and its warning claims "the tests that need the real model are `#[ignore]`-gated and are not run in this job" — false, and contradicted by the block comment 30 lines above it in the same file.

## Evidence

Run [31026363957](https://github.com/bobmatnyc/trusty-tools/actions/runs/31026363957), Test job step list:

```
 8. Restore fastembed model cache -> success
 9. Pre-seed fastembed models (cache miss only) -> success
12. cargo test --workspace -> failure
24. Post Restore fastembed model cache -> skipped
```

Same job, pre-seed output:

```
Cache not found for input keys: fastembed-v07b0e45deb71cb…
Resolving Xenova/all-MiniLM-L6-v2 HEAD revision …
curl: (22) The requested URL returned error: 429
[...]

PR #5366, opened 2026-08-10
perf(ci): shard the Test job four ways with nextest, and stop PRs evicting the cargo cache

bobmatnyc:
`Test` is the critical path of every CI run. Step-level timings from seven
recent runs (`gh api .../jobs`):

| | warm cargo cache | cold cargo cache |
|---|---|---|
| `cargo test --no-run` (build) | 466–472s | 704–827s |
| `cargo test --workspace` (exec) | 421–422s | 420–454s |
| **`Test` total** | **989–1055s** | **1370–1432s** |

Two causes, two changes.

### 1. Execution: nextest, partitioned four ways

`cargo test` runs the ~220 test binaries one after another and parallelises
only *within* a binary — it drains a tail of idle cores ~220 times. nextest
schedules from one global queue; `--partition hash:N/4` splits that queue
across four runners.

**Partition, not crate group.** Splitting by crate (`-p a -p b` per shard)
would cut the build too, and was the first version of this PR. It silently
loses coverage. Cargo resolves features across the packages in *one*
invocation, so a shard building a subset resolves a subset of features.
Measured: a shard holding `trusty-common` plus its six nearest neighbours saw
**24 of the 37** `trusty-common` features `--workspace` resolves — losing
`embedder`, `memory-core`, `bm25`, `migrations` — and listed **1340**
`trusty-common` tests where the workspace run executes **2099**. Four green
shards, 759 tests not run, nothing anywhere saying so.

`--partition` splits the test list of an unchanged `--workspace` build, so
feature resolution is byte-identical to today and every test lands in exactly
one shard by construction.

Doctests are not part of what nextest runs (it has no doctest harness), so
`cargo test --doc` runs them in shard 1 rather than letting them silently stop.

### 2. Cold builds: PRs were evicting the shared cargo cache

The repo sits at **11.03 GB against GitHub's 10 GB quota**, with four
byte-identical 1.29 GB copies of this job's cache on PR refs — none readable by
any other PR — which had evicted the `refs/heads/main` copy every PR restores
from. That eviction is the cold column above. `save-if` now restricts saving to
main; all four shards share one cache key, since they run the same build.

### Coverage

Nothing is `#[ignore]`d, `cfg`-gated, `--exclude`d or narrowed. The full
workspace was re-verified under nextest's process-per-test isolation before the
switch: **23856 tests run, 23853 passed, 3 failed** — and all 3 reproduce
identically under plain `cargo test`, so none is caused by nextest:

- `trusty-agents ctrl::tests::config_tests::resolve_agent_config_falls_back_to_project_ctrl_toml`
  reads the developer's real `~/.trusty-agents/agents/ctrl.toml`; it fails the
  same way under `cargo test … -- --exact` on a machine that has one, and
  passes on CI, which does not. It is the only test in its file that does not
  sandbox `$HOME`.
- `trusty-common memory_core::dream::tests::{dream_cycle_merges_duplicates,
  dream_cycle_compression_ratio_nonzero_after_dedup}` are the documented
  pre-existing `hnsw_rs` recall flake (see the #4904/#5162 comment in
  `memory_core/retrieval/tests.rs`).

### Branch protection

**No edit needed.** The required check is the literal context `Test`; the new
aggregate job keeps that name and fails unless every shard reports `success`.

### Stale comment

The `ci.yml` block claiming nextest was blocked by a `trusty-memory` test-order
bug was stale — #4413 fixed that coupling at its root in 2026-07.
`.github/scripts/attribute_test_failures.py` (#4402) is deleted: nextest names
the owning crate for every test, which is what that script reconstructed from
panic source paths.

Rung 5 (CI/release tooling).

> `main` is red from `364aba420`, independent of this PR — see the review thread.


bobmatnyc:
**`main` is red, independent of this PR — the `search` shard here will fail on it.**

`364aba420` ("fix(trusty-search): make a degraded index say so on every
surface", #5345) merged with a semantic conflict against `ae9bf142e` (#5335,
merged ~20 min earlier). #5345 changed the index handlers to return
`Err((StatusCode, Json<Value>))`; two assertions #5335 added still compare
against a bare `StatusCode`:

```
error[E0369]: binary operation `==` cannot be applied to type `(StatusCode, axum::Json<serde_json::Value>)`
   --> crates/trusty-search/src/service/server/tests_4715.rs:249:5
   --> crates/trusty-search/src/service/server/tests_4715.rs:263:5
```

`Test` and `Clippy` have failed on every push to main since (runs 31429060686,
31429735081, 31429742338, 31430155923). The third assertion in the same
function, `grep_err.0` at line 277, was updated — these two were missed.

Fix is `status_err.0` and `chunks_err.0`. Not carried here: different outcome,
different crate, and it should land on its own so it can be merged immediately
rather than waiting on this PR's review.

bobmatnyc:
### Measured, cold cache both sides

`main` is red, so the green run below was taken on this branch with the 2-line
`trusty-search` fix from the previous comment stacked on top **temporarily**;
that commit has been removed and the branch is back to CI-only changes. Run
[31434846476](https://github.com/bobmatnyc/trusty-tools/actions/runs/31434846476),
all four shards green.

| | before (cold) | after (cold) |
|---|---|---|
| build (`cargo test --workspace --no-run`) | 704–827s | 825–846s |
| test execution | 420–454s | **148–165s** |
| doctests + trusty-common re-run (shard 1) | — | 88s |
| **`Test` wall clock** | **1370–1432s** | **1099s** |

Execution is 2.7× faster. The build is unchanged *by design* — every shard runs
the same `--workspace` build, which is what makes the feature resolution
identical — so it is now 78% of the job.

Warm is the case that actually matters day to day, and it is where the cache
fix lands: build drops to ~470s, giving ~680s against a 989–1055s warm
baseline. The `save-if`/`shared-key` change is what makes warm the common case
again rather than a coin flip.

### Coverage

```
shard 1: Summary [ 75.277s] 5900 tests run: 5900 passed, 18050 skipped
shard 2: Summary [ 78.403s] 5963 tests run: 5963 passed, 17987 skipped
shard 3: Summary [ 62.935s] 5959 tests run: 5959 passed, 17991 skipped
shard 4: Summary [ 72.944s] 5983 tests run: 5983 passed, 17967 skipped
```

5900 + 5963 + 5959 + 5983 = **23,805** run, against 23,711 reported by the
unsharded `cargo test` it replaces, plus the 7 doctests in shard 1. Each
shard's `run + skipped` is the same 23,950, i.e. all four partitioned the same
list; the 145 never-run are the pre-existing `#[ignore]`s.

### Remaining lever, not taken here

The build is now the whole cost. Nothing in this PR touches it, and the ~825s
cold figure is four runners doing the same work. Linking ~220 test binaries
with GNU `ld` is the obvious next target (`lld`/`mold`), worth its own PR with
its own measurement.

bobmatnyc:
## Verdict: BLOCK

Rung 5 (CI/release tooling). Reviewed against `origin/main` @ `15c328b73`, PR branch `perf/5325-ci-test-shard` @ `64ed7b783`, and live CI run 31436371546 (all 4 `Test shard` legs + the `Test` aggregator: `success`).

## Findings

| CRITICAL | `.github/workflows/ci.yml` | 544 | `bash scripts/ci-preseed-fastembed.sh` — this file does not exist anywhere in the repo. Verified: absent from `origin/main`, absent from this PR's tip, `.github/scripts/` no longer exists at all (only tenant was the deleted `attribute_test_failures.py`), and no other branch carries it. The inline preseed logic (`seed_models`, `XENOVA_REPO`, `QDRANT_REPO`, retry/backoff, HF auth) that this line was extracted from is gone with no replacement committed. On a cache miss this step runs `bash: scripts/ci-preseed-fastembed.sh: No such file or directory` (exit 127); because the step has no `-e` and its non-zero branch only sets `seeded=false` + emits a `::warning::`, the *job* does not fail — the pre-seed mechanism just silently becomes a permanent no-op. That mechanism (#865, #2554, #4940) exists specifically to stop concurrent HuggingFace downloads from 429-ing tests; once it's gone, `default_model_matches_sentence_transformers_reference` (the one non-`#[ignore]`d real-model test) is exposed to that flake again, which fails its shard and the aggregate "Test" required check, looking exactly like an unrelated flake to whoever hits it. This PR's own CI run (31436371546) passed only because the fastembed cache still holds an entry populated by the OLD inline script from before this change — `gh run view --job 93611344659` shows "Pre-seed fastembed models (cache miss only)" as **skipped** (cache hit), so the broken path never executed. The extraction also drops `env: HF_TOKEN: ${{ secrets.HF_TOKEN }}` (present on the old step, absent here), so even once the script is restored, authenticated (higher-rate-limit) downloads regress unless that's added back too. | Add the missing `scripts/ci-preseed-fastembed.sh` (the deleted inline logic, parameterized on `$FASTEMBED_CACHE_DIR`/`$HF_TOKEN`) to this PR, `chmod +x` it, and restore `env: HF_TOKEN: ${{ secrets.HF_TOKEN }}` on the step. | Fix here |
| MEDIUM | `.github/workflows/ci.yml` | 389, 638 | Shard count is a magic number duplicated in two places with nothing tying them together: the matrix list `shard: [1, 2, 3, 4]` (line 389) and the partition literal `--partition hash:${{ matrix.shard }}/4` (line 638). nextest only errors when the numerator exceeds the denominator (`matrix.shard` > divisor); it does *not* error when the divisor itself doesn't match the matrix size. Extend the matrix to 5 shards without also bumping the `/4`, or shrink it without lowering the `/4`, and one partition's tests silently never run — with every shard reporting green, since a mismatched-low matrix count relative to the divisor just drops the missing partition entirely. Nothing in the job output states the divisor, so this drift is invisible short of an audit like this one — exactly the "silent cap" this rung's review is meant to catch. | Replace the literal `4` with GitHub's built-in matrix-size context: `--partition hash:${{ matrix.shard }}/${{ strategy.job-total }}`. Ties the divisor to the actual matrix definition so it can't drift. | Fix here |

## Notes

- **Coverage claim verified, independently, against live CI**: summed `cargo nextest run` "tests run" counts across the 4 shards in run 31436371546 — 5900 + 5963 + 5959 + 5983 = **23805**, matching this task's stated figure exactly. Each shard's run+skipped total is a consistent 23950, confirming all four invocations enumerate the identical (unpartitioned) test list before partitioning — i.e. the "same `--workspace` build, feature resolution byte-identical" claim holds structurally, and the crate-sharding coverage loss the author already caught (24/37 features, 759 dropped `trusty-common` tests) genuinely does not recur under hash partitioning. Doctest (`cargo test --doc`, shard 1 only) and the `cache_fresh_returns_some_when_newer` re-run (shard 1 only, `CI=""`) both executed and passed in the live run, confirming the two carve-outs are live, not just documented.
- The PR body's own stated coverage number ("23856 tests run, 23853 passed, 3 failed") doesn't match the 23805 measured just now on this PR's own CI run. Most likely explanation is ordinary test-count drift between the author's manual verification and this run, on a repo merging this fast — not a defect, but worth a sentence in the PR body distinguishing "measured at time of writing" from "current."
- Considered and downgraded out of the findings table (confidence on real-world impact too low to assert): the shared `shared-key: test` cache across the 4-shard matrix, combined with `save-if: main`, means on a push to main all four shards race to save under the identical key — only the first to reach its POST step wins, GitHub's cache API soft-rejects the rest. Shard 1's tree differs slightly at save time from shards 2–4 (it alone runs `cargo test --doc` and a `CI=""`-fingerprinted recompile of `trusty-common` after the shared build phase), so if shard 1 wins the race, the persisted cache is not byte-identical to what shards 2–4 would have saved. Likely just a minor incremental-recompile cost on the next run, not a correctness issue — flagging for awareness, not blocking.
- The aggregate gate itself is sound: `needs.test-shard.result` is checked with `!= success` (not `!= failure`), so `cancelled`, `skipped`, and any other non-success conclusion all fail the check closed, matching #4179's existing pattern elsewhere in this file. The `test` job's own `if: ${{ !cancelled() }}` is copied verbatim from the pre-existing single `test` job's `if:` (confirmed via `git show origin/main:.github/workflows/ci.yml`) — the one scenario where that condition itself could fail open (an entire-workflow cancellation skipping the aggregator, which GitHub reports as a passing required check) is pre-existing behavior unchanged by this PR, not a new hole it introduces.
- Branch protection verified directly (`gh api .../branches/main/protection`): required contexts are exactly `["Format check","500-line file-size cap","Clippy","Test","trusty-search daemon smoke test","MSRV check (1.94)"]` — no per-shard "Test shard N" context is required, matching the PR's "no edit needed" claim.
- Cache-quota narrative independently checked against `gh api repos/.../actions/caches`: two other currently-open PRs (#5365, #5370) each still carry their own 1.29 GB `v0-rust-test-test-Linux-x64-...` copy on `refs/pull/*`, alongside the `refs/heads/main` copy — confirms the eviction mechanism described in the PR body is real and current, not hypothetical.

## Required Changes

1. Add `scripts/ci-preseed-fastembed.sh` to this PR with the extracted preseed logic, restore `HF_TOKEN` env wiring on the "Pre-seed fastembed models" step, and verify with a manual cache-miss run (e.g. bump the fastembed cache key locally in a throwaway branch) that the step actually succeeds end-to-end rather than merely not-erroring.
2. Derive the nextest partition divisor from `${{ strategy.job-total }}` instead of the hardcoded `4` at line 638.

bobmatnyc:
## Both critic findings fixed — and CI then surfaced a third that blocks this PR

### CRITICAL — `scripts/ci-preseed-fastembed.sh`: fixed by restoring the original

The critic is right, and the framing "the logic was extracted and the script never committed" is generous. There was no extraction. `origin/main` has **209 lines of inline** pre-seed logic; I replaced them with a call to a script I invented. I had read that region from the *main checkout* at a different revision, through truncated output, and reconstructed the part I never saw.

Fix: restored the step verbatim rather than re-extracting it.

```
pre-seed step identical to origin/main: True
209 lines (origin/main had 209)
```

`HF_TOKEN` is back with it. Every script path the workflow invokes now resolves:

```
EXISTS  scripts/check-ci-helpers-selftest.sh
EXISTS  scripts/ci-free-disk-space.sh
EXISTS  scripts/classify-ci-results.sh
EXISTS  scripts/detect-docs-only.sh
```

**Fail vs warn: neither — restored.** The original already draws the distinction the critic asked for, and my version destroyed it. It runs `set -euo pipefail` for the whole step and drops to `set +e` around *only* the download (ci.yml:718-721). So "HuggingFace is angry" warns, and "this step is broken" goes red. Both arms, driven against the step body extracted from the live workflow:

```
########## ARM A: HuggingFace unreachable (transient/external) ##########
exit=0
GITHUB_OUTPUT: seeded=false
::warning::fastembed model pre-seed failed (rc=1; HuggingFace rate-limit/403 or network issue)
cache dir: removed, so no partial tree is ever saved as a false hit

########## ARM B: step cannot run at all (a real bug) ##########
exit=1
GITHUB_OUTPUT: ''  (empty = never claimed success)
mkdir: .../armB-root/denied: Permission denied
```

And the regression as I had written it, confirming what it would have done:

```
########## COUNTERFACTUAL: the step exactly as I had broken it ##########
exit=0   <-- what GitHub reads; 0 means the step reports SUCCESS
bash: scripts/ci-preseed-fastembed.sh: No such file or directory
::warning::fastembed model pre-seed failed (rc=127; ...)
GITHUB_OUTPUT: seeded=false
```

### MEDIUM — shard count: fixed

`--partition hash:${{ matrix.shard }}/${{ strategy.job-total }}`. `shard: [1, 2, 3, 4]` is now the only place the count exists.

### NEW — nextest exposes a second hidden-coupling class, and my "verified" claim was wrong

Run [31438217228](https://github.com/bobmatnyc/trusty-tools/actions/runs/31438217228) (main is green again, so the build is clean): shards 2/3/4 pass, **shard 1 fails 17 `trusty-memory` tests** — none on an assertion:

```
thread 'tools::tests::dedup_allows_different_content' panicked at crates/trusty-memory/src/tools/tests.rs:1363:6:
memory_remember seed: PalaceHandle::remember_with_options
Caused by:
    1: init shared FastEmbedder
    2: failed to initialise fastembed (tried AllMiniLML6V2 and AllMiniLML6V2Q)
    4: request error: http status: 429
```

Same shape as #4413, different shared cell. `seed_shared_embedder_with_mock` says so itself:

> Idempotent — if the cell was already set **(by an earlier test in the same process)**, the call is a silent no-op

Under `cargo test` one process runs a whole binary, so one sibling seeds `SHARED_EMBEDDER` for everyone. Under nextest's process-per-test each of these 17 gets a virgin cell, falls through to the real FastEmbedder, and hits HuggingFace. Every caller of the helper is in `tests/`; none of the failing `src/` unit-test modules calls it.

**I claimed the opposite and was wrong.** The commit message and PR body say the workspace was "re-verified under nextest — no test depends on another test's side effects." My local run did report 23853/23856 passing, but it passed because *my machine downloads the model successfully*, so the fallthrough silently worked. The local run could not have detected this; I should not have generalised from it.

Affected: `tools::tests` (9), `commands::kg_rebuild::tests` (2), `commands::prompt_context::tests` (2), `web::tests::palace_*` (3), `mcp_stdio_tools` (1).

Fix is the #4413 shape — each test calls `seed_shared_embedder_with_mock()` to establish its own precondition, not a global env var or a CI-side workaround. That is a `trusty-memory` source change with a changelog fragment, and it must land before nextest can be enabled. I have not attempted it here; I would rather hand it over than rush 17 tests I cannot verify against a rate-limiting HuggingFace.

PR #5371, opened 2026-08-10
feat(trusty-mpm): detect and quarantine stray .mcp.json files above a workspace

bobmatnyc:
## What prompted this

`/private/tmp/.mcp.json` exists on the development machine: mode `0644`, dated
2026-08-08, declaring the four `trusty-*` MCP servers plus four operator HTTP
servers (`apex`, three `duetto-*`), and pinning `trusty-search serve --index tmp`
and `TRUSTY_MEMORY_PALACE=tmp`. Those are observed properties of the file, not a
claim about impact beyond the discovery mechanism below.

Claude Code discovers `.mcp.json` by walking UP from a session's cwd. A file at
`/private/tmp/.mcp.json` is therefore read by every session whose cwd is anywhere
under `/tmp` — agent scratchpad directories included. Nothing reported it: no
error, no log line, nothing in any project to point at. It was found by looking.

Stopping future writes is a separate, undecided change (ADR-0042 / #5362, still
Proposed). **This work does not depend on that decision, and could not.** Removing
an injector prevents new files; it does not clean up one already on disk, and a
file already on disk keeps being discovered regardless of what the write path
later does.

## The problem this had to solve first

There was no way to tell a `.mcp.json` tm wrote from one the operator wrote.

That is not a detail — it is the whole design constraint. "It contains `trusty-*`
servers" is not provenance: an operator can register exactly those entries by
hand, and the observed file holds framework servers and operator servers in the
same `mcpServers` map. Deleting on that evidence is data loss.

So the missing evidence is built here, and until it exists the repair refuses.

## What landed

**Provenance ledger** — `crates/trusty-mpm/src/core/mcp_provenance.rs`.
A machine-global `~/.trusty-mpm/mcp-json-provenance.json` recording every
project-scope `.mcp.json` tm writes, with a sha256 of the bytes written. Recorded
from `session_launch::settings::inject_mcp_server`, the single write path.

Deliberately a side ledger, not a marker in the file: a marker key would put
tm-invented bytes into a file Claude Code parses and some repos track in git; a
sidecar would litter a second stray file into the very directories this cleans
up. The checksum additionally separates "tm wrote it" from "tm wrote it and
nobody has edited it since" — the ownership rule `skill_repair` already applies
to a FROZEN skill.

**Detection** — `crates/trusty-mpm/src/daemon/doctor_stray_mcp.rs`, check
`stray_mcp_json`. Read-only, `Warn`, names each path, the servers it declares,
and its provenance verdict.

**Repair** — `crates/trusty-mpm/src/core/stray_mcp.rs`, wired into
`tm doctor --fix` plus a new `tm doctor --quarantine-mcp <PATH>`. Dry run unless
`--yes`. `--yes` widened from `requires = "fix"` to a `writes` arg group.

## Design decisions worth reviewing

**Nothing deletes.** A quarantine RENAMES `.mcp.json` to
`.mcp.json.quarantined-<epoch>` beside itself. Claude Code matches the exact
basename, so the rename is what actually stops the discovery, and every byte
survives. This is why the sweep could join `--fix` at all — `doctor_repair`'s
"nothing is deleted, nothing operator-authored is rewritten" rule is preserved
rather than excepted, and no new hidden flag was needed.

**What the sweep refuses**, each with the reason printed and the path named:
unattributed (no ledger record — which is every file on disk today, the observed
one included); tm-written-then-edited (current bytes are somebody's edit);
symlink (the bytes are elsewhere); unreadable ledger or unreadable file
(attribution unavailable — never guessed in either direction); and the
workspace's own `.mcp.json`, which is live managed config.

**The escape hatch.** `--quarantine-mcp <path>` clears a pre-ledger file because
naming an exact path is attribution the operator supplies and tm cannot. It still
refuses a non-`.mcp.json` path, the workspace's own file, and a symlink. The
alternative — loosening the sweep's evidence rule — would remove the only thing
standing between the repair and hand-written config.

**Search bound.** Strict ancestors of the workspace, walking up and stopping at
the home directory (inclusive), the filesystem root (exclusive), or 8 levels —
whichever comes first — plus `$TMPDIR` and `/tmp`. Not a walk to `/` (above home
is system config and none of tm's business) and not a recursive descent (that
would report other tools' legitimate config as strays). The temp roots are listed
explicitly because they are not ancestors of any workspace — they are ancestors
of the agent scratchpad cwds, which is exactly why a file there reaches so many
sessions. Fixed set, a handful of `stat` calls.

Scan dirs dedupe by canonical path and ledger keys canonicalize the parent
directory: on macOS `/tmp` is a symlink to `/private/tmp`, which without this
reported one file twice and would have split its attribution across two keys.

## Test evidence — rung 5

```
cargo check -p trusty-mpm --all-targets                EXIT=0
bash scripts/check_line_cap.sh                          0 violations
bash scripts/check_sld.sh                               0 errors, 0 warnings
```

`cargo test -p trusty-mpm` → `4955 passed; 2 failed; 7 ignored`.

Both failures are **pre-existing on `origin/main`**, not from this branch. Same
commands in a detached `origin/main` worktree at `15c328b73`:

```
mine (33ab321cb):  4955 passed; 2 failed
base (15c328b73):  4905 passed; 2 failed
```

Identical failing set both times:
`core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning`
and `daemon::managed_routes::tests::stale_assets_for_many_reads_shared_agent_dir_once_for_the_whole_fleet`.
Both pass in isolation on this branch and are parallel/shared-state flakes
(the first captures zero log lines; the second asserts a fleet starts fresh).
This branch touches neither file.

~50 new tests. Failure paths carry the weight, since a happy-path suite would
prove nothing about the property that matters: `sweep_refuses_an_unattributed_stray`,
`sweep_refuses_a_stray_edited_after_tm_wrote_it`, `sweep_refuses_when_the_ledger_is_unreadable`,
`sweep_refuses_a_symlink`, `sweep_never_touches_the_workspaces_own_file`,
`explicit_refuses_a_non_mcp_path`, `explicit_refuses_the_workspaces_own_file`,
`record_refuses_to_clobber_an_unreadable_ledger`, `record_serialises_concurrent_writers`,
`quarantine_destination_does_not_collide`.

`git diff --name-only origin/main...HEAD -- crates/trusty-search/` is empty.

### Real-binary smoke run

Against a sandbox fixture — dry run, then apply, then bytes verified byte-identical
at the new name. The `--fix` sweep run from a cwd under `/tmp` reported the live
file and refused it:

```
- [stray_mcp_json] /private/tmp/.mcp.json: quarantine this stray MCP config — REFUSED:
  tm has no record of writing this file, and its contents cannot prove authorship —
  a `.mcp.json` full of trusty-* servers may equally be one the operator wrote.
```

`/private/tmp/.mcp.json` was not modified: still 1179 bytes, mtime Aug 8 13:44.
No live file on this machine was touched at any point; every write went to a
tempdir fixture.

The smoke run caught two defects that the unit tests did not, both fixed here:
the shared step printer told a `--quarantine-mcp` operator to re-run
`tm doctor --fix --yes` (a command that would not touch their file), and the
`/tmp` vs `/private/tmp` duplicate above.

## For the operator

The observed file is unattributable and `--fix` will refuse it. To clear it:

```
tm doctor --quarantine-mcp /private/tmp/.mcp.json        # preview
tm doctor --quarantine-mcp /private/tmp/.mcp.json --yes  # rename aside
```

Worth reading it first — it declares operator servers alongside the framework
ones, and peer sessions are currently running with cwd under `/tmp`.


bobmatnyc:
## Verdict: BLOCK

Reviewed `33ab321c`/`58f09f6b` against `origin/main` (`15c328b7`). Confirmed independently: `cargo fmt --check`, `cargo clippy -p trusty-mpm --all-targets -- -D warnings` (clean), `bash scripts/check_line_cap.sh` (0 violations), and the crate's own `stray_mcp`/`mcp_provenance`/`doctor_stray_mcp` tests (31 + 19 passed, matching the PR body). All verification ran against tempdir fixtures in a disposable worktree; no live `.mcp.json`, no operator ledger, nothing under `/tmp` was touched.

The refusal set holds up for every case its own tests exercise: unattributed, edited-after-write, unreadable ledger, symlink, and the workspace's own file via the *sweep* path (`quarantine_strays`) are all genuinely refused, and `scan_dirs` structurally excludes the workspace itself regardless of path spelling (it's never added to the ancestor list, so no comparison can go wrong there). The `/tmp`-vs-`/private/tmp` fix is real and tested (`absolute_key_collapses_symlinked_parents`, `scan_dirs_has_no_duplicates`).

The escape hatch does not hold up to the same standard — see CRITICAL below, reproduced with a real test.

## Findings

| CRITICAL | `crates/trusty-mpm/src/core/stray_mcp.rs` | 278–286 | `quarantine_explicit`'s "is this the workspace's own file" guard is a literal `target.parent() == Some(workspace)` comparison, with no canonicalization. `workspace` comes from `std::env::current_dir()` and `target` from the operator's CLI argument (`doctor_repair.rs:79-84`) — two independent, un-normalized paths. When they name the same directory via two spellings (the exact macOS `/tmp` vs `/private/tmp` aliasing this PR fixes elsewhere in `absolute_key`), the guard silently fails to fire and `tm doctor --quarantine-mcp <path> --yes` renames the CURRENT project's live, in-use `.mcp.json` — the one case every other refusal in this PR is designed to protect. I reproduced this: added a temporary unit test in `stray_mcp_tests.rs` (tempdir fixture only, run once, then `git checkout --` to revert — no trace left in the tree) that calls `quarantine_explicit(&root, Some(&link_ws), &real_ws.join(".mcp.json"), RepairMode::Apply)` where `link_ws` is a symlink to `real_ws`. Result: `StepStatus::Applied`, and the workspace's own file no longer existed. | Canonicalize both sides before comparing — reuse `mcp_provenance::absolute_key` (already solves exactly this) for both `target` and `workspace.join(MCP_JSON)` and compare the keys, instead of raw `PathBuf` equality. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/stray_mcp.rs` | 215–242, 318–353 | TOCTOU in the sweep: `quarantine_strays` classifies provenance once inside `scan()` (which reads and checksums the file), then hands `TmWritten` findings to `apply_or_plan`, which renames without re-verifying the checksum immediately beforehand. If the file's content changes between the scan's read and the rename — plausible given the PR's own stated context: peer sessions can be running concurrently under the exact ancestor/temp-root directories being swept — a file that was provably tm's at scan time but has since become someone's live edit gets quarantined on the strength of a now-stale classification. No test drives a mutation into that window; every provenance test checks-then-acts with nothing in between. Recoverable (rename, not delete) only if the operator knows to look for the `.quarantined-<epoch>` file. | Re-run `mcp_provenance::classify` (or re-checksum) immediately before the `fs::rename` inside `apply_or_plan`, and turn the outcome into `Refused`/`Failed` if it no longer reads `TmWritten`. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/doctor_repair.rs` | 69–91 | The PR body cites this as one of two defects the real-binary smoke test caught and fixed (the wrong `apply_hint` string for `--quarantine-mcp`). By inspection the fix (line 89) is correct, but nothing exercises it: `tm/tests.rs`'s only `quarantine_mcp`-related tests are CLI-parsing tests (`cli_parses_doctor_quarantine_mcp`, `cli_rejects_doctor_yes_without_a_write_action`) — neither calls `quarantine_one_mcp` or `print_steps`. A future edit to `print_steps` or its call sites can silently reintroduce the same wrong-command message with nothing failing. | Add a test that captures `quarantine_one_mcp`'s output (or extracts the hint-building into a small pure function and asserts on its return value) and pins the `--quarantine-mcp <path> --yes` string. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/core/mcp_provenance.rs` | 244–275, 289–306 | The provenance ledger has no pruning or size bound. `record_write` runs on every managed session launch and adds an entry keyed by the launch's absolute `.mcp.json` path; `forget` only removes an entry when that exact path is later quarantined. On a machine that creates many ephemeral workspaces (this repo's own `.claude/worktrees/<name>` convention, or agent scratchpad tempdirs), the ledger grows by one entry per distinct path, indefinitely, with no expiry. Not a correctness defect on its own — the checksum still gates every classification, so a stale entry at worst produces a wrong `TmWrittenThenEdited` refusal if some unrelated future file happens to collide on both path and checksum, not a wrong deletion — but it's unbounded state with no visibility into its size. | Add a periodic prune (drop entries whose path no longer exists) run from `tm doctor`, or cap/rotate the ledger. | Promote |

## Notes

- Hardlinks: considered and not flagged. A quarantine only detaches the specific directory entry being renamed; a hardlinked file's other name and inode are untouched, and the checksum still describes the actual shared bytes at classification time. No distinct exposure beyond the TOCTOU already flagged above.
- `quarantine_explicit` accepting an operator-named path outside the `scan_dirs` bound (anywhere on the filesystem, as long as the basename is `.mcp.json`) is by design, not a bug — the operator naming the exact path IS the attribution this command exists to supply, the same trust model as any destructive CLI flag that takes an explicit path argument.
- Fail-open check: every failure branch I traced (missing ledger, corrupt ledger, unreadable file, permission denied on read, rename failure, `$TMPDIR` unset, canonicalization failure in `absolute_key`) refuses or falls back to a stable-but-safe key — none of them advance to an unattributed "safe to act" state. The one place this project's stated safety property does NOT hold is the CRITICAL above, which is not a failure-handling gap but a missing normalization step in a direct comparison.
- Write-then-record ordering in `session_launch/settings.rs` (write the `.mcp.json`, then best-effort record it) is the safe order: a crash between the two leaves an `Unattributed` file, which the repair correctly refuses, rather than a phantom ledger entry for a file that was never written.

## Required Changes

1. Fix the `quarantine_explicit` own-file check to compare canonicalized paths (or reuse `absolute_key`), and add a regression test reproducing the symlink/path-spelling bypass — the one I used to confirm this should be adapted into the suite, asserting the workspace's own file survives.
2. Re-verify provenance immediately before the rename in `apply_or_plan`, with a test that mutates the file between scan and apply and asserts a refusal.
3. (Recommended, not blocking re-review) Add the missing regression test for the `--quarantine-mcp` step-printer hint text.

bobmatnyc:
## Critic remediation — all four addressed

### CRITICAL — lexical workspace guard (fixed, and the class audited)

`quarantine_explicit` compared `target.parent() == Some(workspace)` lexically.
Fixed by routing every path comparison in the module through one `same_dir`
helper built on `canonical_dir` — the mirror of what `absolute_key` already did
for the ledger key.

Auditing the class rather than the instance found a **second occurrence**: the
scan's home ceiling (`ancestor == home`) was also lexical, so a symlinked home
spelling never matched and the walk ran past home to the depth cap.

**RED, pre-fix** (`33ab321c`):

```
---- explicit_refuses_the_workspaces_own_file_via_a_symlinked_workspace_spelling stdout ----
the workspace's own .mcp.json must be refused however the workspace is spelled, got
RepairStep { check: "stray_mcp_json", path: ".../home/real-ws/.mcp.json",
  what: "rename aside to `.mcp.json.quarantined-1786402761` ...",
  status: Applied { backup: Some(".../home/real-ws/.mcp.json.quarantined-1786402761") } }

---- scan_dirs_stops_at_home_through_a_symlinked_spelling stdout ----
the walk must stop at home however home is spelled — it reached above it:
["/var/folders/7s/.../.tmp1kn0VE/real-home/a", ".../real-home", ".../.tmp1kn0VE",
 "/var/folders/7s/g9twvy8j0wl58ffgzsmrccv40000gp/T", "/var/folders/7s/g9twvy8j0wl58ffgzsmrccv40000gp",
 "/var/folders/7s", "/var/folders", "/var", "/tmp"]

test result: FAILED. 31 passed; 2 failed
```

`Applied` on the live file, and a scope escape into `/var/folders` — both confirmed.

**GREEN, post-fix:**

```
test core::stray_mcp::tests::scan_dirs_stops_at_home_through_a_symlinked_spelling ... ok
test core::stray_mcp::tests::explicit_refuses_the_workspaces_own_file_via_a_symlinked_workspace_spelling ... ok
test result: ok. 35 passed; 0 failed
```

### HIGH — TOCTOU between scan and rename (fixed)

New `apply_verified` re-runs `classify` immediately before the rename; only a
still-`TmWritten` verdict proceeds. A file that changed is **refused, not
retried** — retrying races the same way, and the operator's edit is exactly what
must survive. The window cannot be closed atomically, so it is made small and
the stale verdict is never trusted.

`sweep_refuses_a_file_edited_between_scan_and_rename` drives a mutation into the
window and asserts the edited bytes survive;
`sweep_still_quarantines_when_nothing_changed_in_the_window` proves the guard
discriminates rather than always-refusing. Both pass.

### MEDIUM — untested hint string (fixed)

Extracted to `quarantine_apply_hint()` and `FIX_APPLY_HINT`, with
`quarantine_hint_names_the_quarantine_command_and_path` asserting the quarantine
hint never contains `--fix`. Verified against the real binary too.

### MEDIUM — ledger bound: **it needed one, and now has one**

My call, with the reasoning: size alone was not the problem — ~200 bytes per
entry, one per *distinct path*, so re-writes replace rather than append. But the
entry count tracks **worktrees ever created**, not workspaces that exist, and
this repo creates a worktree per PR. That only ever climbs, so "it's fine" was
not defensible.

`prune_vanished` drops entries whose file no longer exists, but **only above 512
entries** — the sweep costs one `stat` per entry and runs on the session-launch
write path, so an unconditional sweep would put thousands of `stat` calls on
every launch to reclaim bytes nobody is short of. Steady state stays near 100 KB
plus live workspaces. Correctness never depended on this: a stale entry already
classified `Unknown` and was refused.

## Gates

```
cargo fmt --check && cargo check -p trusty-mpm --all-targets && cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0
scripts/check_line_cap.sh    0 violations
scripts/check_sld.sh         0 errors, 0 warnings
```

`cargo test -p trusty-mpm` → `4962 passed; 1 failed; 7 ignored`. The one failure
is `core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning`,
which also fails on the `origin/main` baseline and passes in isolation.

Binary re-smoke confirmed the corrected hint, the workspace-guard refusal, and
`/private/tmp/.mcp.json` still 1179 bytes / mtime Aug 8 13:44. No live file
touched at any point.

bobmatnyc:
## Verdict: APPROVE

Re-review of `26655e71` (rebased onto `7a2cf36b`+, remediation commit on top of the original `5dcad35c` feature commit). All three prior findings are genuinely closed; no new CRITICAL or HIGH found.

## Prior findings — status

**CRITICAL (lexical workspace guard) — CLOSED.** `stray_mcp.rs:299-339` (`quarantine_explicit`) and `stray_mcp.rs:95-130` (`scan_dirs`' home ceiling) both now route through one `same_dir`/`canonical_dir` pair (`stray_mcp.rs:152-162`). Grepped every `==`/`.parent()` comparison in `stray_mcp.rs`, `mcp_provenance.rs`, `doctor_stray_mcp.rs`, and `daemon/doctor.rs` — the only path-identity comparison left in the module is `same_dir` itself; `mcp_provenance::absolute_key` already canonicalized the parent for the ledger key (pre-existing, not lexical). No third occurrence.

Checked out the pre-fix commit (`5dcad35c`, `33ab321c` pre-rebase) in a separate worktree, transplanted the two new regression tests onto it, and ran them: both fail, and the failure output is the bug itself —
```
explicit_refuses_the_workspaces_own_file_via_a_symlinked_workspace_spelling: 
"got RepairStep { ... status: Applied { backup: Some(".../.mcp.json.quarantined-1786403998") } }"
scan_dirs_stops_at_home_through_a_symlinked_spelling:
"the walk must stop at home ... it reached above it: [... "/var/folders", "/tmp"]"
```
Confirms the workspace's live `.mcp.json` really was renamed away pre-fix, and the scan really did reach `/var/folders`/`/tmp` — matching the PR description's own repro. Ran `cargo test -p trusty-mpm --lib -- stray_mcp mcp_provenance`: 56 passed, 0 failed.

**HIGH (TOCTOU) — CLOSED.** `apply_verified` (`stray_mcp.rs:360-373`) re-classifies immediately before rename and only proceeds on a still-`TmWritten` verdict. `sweep_refuses_a_file_edited_between_scan_and_rename` proves the refusal path (edit survives, file not renamed); `sweep_still_quarantines_when_nothing_changed_in_the_window` proves the guard isn't an always-refuse — an unchanged file still gets `Applied`. Both pass.

**MEDIUM x2 — CLOSED.** `quarantine_apply_hint`/`FIX_APPLY_HINT` (`doctor_repair.rs`) are named + tested, including the `!hint.contains("--fix")` assertion. `prune_vanished` (`mcp_provenance.rs:298-305`) only sweeps above `PRUNE_THRESHOLD = 512`; `record_keeps_every_entry_below_the_threshold` and `record_prunes_vanished_entries_above_the_threshold` both pass. Confirmed correctness is independent of pruning: `classify` (`mcp_provenance.rs:363-379`) reads the file directly and returns `Unattributed`/`Unknown` regardless of whether a stale ledger entry was pruned — pruning is bookkeeping, not a correctness dependency.

## Fail-Open Check on `canonical_dir`

`canonical_dir` (`stray_mcp.rs:152-154`) falls back to the lexical path when `canonicalize` fails. Traced whether this reopens the CRITICAL: the failure mode needs `canonicalize` to fail on one side of a comparison while the other side resolves to a real path with different spelling. For `quarantine_explicit`'s workspace guard, that requires `target`'s parent (or `workspace`) to be unresolvable while `target` itself still exists as a stat-able regular file — but `symlink_metadata(target)` (called moments later, `stray_mcp.rs:325`) would already fail with the same unresolvability if the path genuinely doesn't resolve, refusing the step before any rename. For `scan_dirs`' home ceiling, a `home` that fails to canonicalize is typically failing identically for the ancestor path derived from the same string, so the lexical fallback still matches. No exploitable path found at >80% confidence; not filed as a finding.

## Rebase

`session_store` (main, PR #5027) and `stray_mcp_json` (this PR) both cleanly registered in `generate/doctor.rs`'s check table and `daemon/doctor.rs`'s check list — no merge markers anywhere in the tree. `run_doctor_produces_thirty_two_checks` lists all 32 names including both, and derives its count from `expected.len()` rather than a literal, so it is a working drift guard, not a stale pinned number.

## Verification performed

- `cargo fmt --check -p trusty-mpm`: clean
- `cargo clippy -p trusty-mpm --all-targets -- -D warnings`: clean
- `cargo test -p trusty-mpm`: 4996 passed, 2 failed, 7 ignored. Both failures (`ensure_managed_config_dir_emits_the_frozen_skill_warning`, `stale_assets_for_many_keeps_per_session_verdicts_correct_under_concurrency`) are outside this PR's diff (`managed_config`/`managed_routes`, untouched by `git diff 7a2cf36b...26655e71`) and pass under `--test-threads=1` in isolation — the documented parallel-execution flake, plus one more of the same class, neither caused by this change.
- `bash scripts/check_line_cap.sh`: 0 violations (`stray_mcp.rs` 462/500 SLOC).
- `cargo test -p trusty-mpm --bin tm -- cli_parses_doctor_quarantine_mcp cli_rejects_doctor_yes_without_a_write_action`: 2 passed.
- Pre-fix regression run at `5dcad35c` (transplanted tests): both new tests fail, confirming they test the real bug.

No temporary test files were added to this checkout; the pre-fix comparison ran in a disposable worktree that has been removed. Nothing was run against live files.

## Notes

No findings at >80% confidence. APPROVED for next pipeline stage.

Issue #5272, opened 2026-08-09
bug(trusty-mpm): session resume can leak another session's pause snapshot — release blocker for 1.3.5


bobmatnyc:
## Release blocker for tm 1.3.5

Owner-designated blocker for the in-tree 1.3.5 bump (not yet tagged). Do not tag/publish 1.3.5 with this open.

## Defect

`session_context_catchup` can resolve and hand a session **another session's** pause snapshot, with nothing in the response flagging the mismatch.

Live on this repo, 2026-08-09: session `7bd5c27a-475b-41df-9e9f-a6f630801717` called the tool and got back `resolved_snapshot: .trusty-mpm/sessions/session-20260809-010155.md`. `sessions-log.jsonl` shows that file was written by a different session:

```
{"session_id":"2eb72dca-de08-481b-8dfa-22ab7f81b1f9","event":"pause","snapshot":"session-20260809-010155.md","timestamp":"2026-08-09T01:01:55.796934+00:00"}
```

Caught only because the snapshot's prose named an agent and a release the resuming session (`7bd5c27a-475b-41df-9e9f-a6f630801717`) had never run.

## Root cause

The resolver's fallback chain — current-session pause → **last pause line overall** → legacy `LATEST-SESSION.txt` → mtime scan of `session-*.md` — is session-agnostic past the first step, so "no snapshot for this session" silently becomes "the most recently paused session's snapshot." That chain was the intentional design of #2731, safe under one-session-per-checkout. ADR-0036's shared main-checkout model (merged `b98950a3`) puts multiple sessions on one checkout with one shared `.trusty-mpm/sessions/` store, so the same fallback now discloses one session's private snapshot to another.

`sessions-log.jsonl` already records `session_id` per line, and `.trusty-mpm/sessions/` already has per-session subdirectories (e.g. `2eb72dca-…/`) — the snapshots themselves are the only thing still flat at the top level.

## Resolution

1. Write snapshots under `sessions/<session-id>/` using the existing per-session directories.
2. Resume never falls back across sessions: no snapshot for the current session id returns empty. Cross-session reads require the caller to pass the explicit `session_id` parameter `session_context_catchup` already accepts.

## Closure conditions

- [ ] No fallback step in the resume path can return a snapshot whose `session_id` differs from the caller's, without an explicit `session_id` argument.
- [ ] Snapshots are written under `sessions/<session-id>/`.
[...]

Issue #3396, opened 2026-07-20
trusty-mpm daemon: duplicate session records for one worktree + stale/crossed tmux pane mappings break session_send


bobmatnyc:
## Summary

The trusty-mpm daemon holds TWO managed-session records that both point at a single worktree, with tmux pane_id mappings that are stale/crossed, causing `session_send` (the MCP inter-session comms tool) to refuse delivery to a session whose pane is actually alive.

## Environment

- Worktree: `/Users/masa/trusty-mpm-projects/bobmatnyc/trusty-tools/.base/.worktrees/f443c12d-2fb6-4ce1-9f70-2e7695306e47`
- tm version: 0.19.26

## The Two Records (from `session_list`)

1. **64d388b5-c3f2-4012-944e-7e9fa1e3fdbf**
   - Name: `tm-tcode-01`
   - State: active
   - Pane ID: `%63`
   - Workspace path: f443c12d worktree

2. **f443c12d-2fb6-4ce1-9f70-2e7695306e47**
   - Name: `tm-tm-tcode-01`
   - State: stopped
   - Pane ID: `%1`
   - Workspace path: SAME f443c12d worktree

## Ground Truth from tmux (Contradicts the Daemon)

- `TMUX_PANE=%63` for the live Opus session; `tmux list-panes -a` shows pane `%63` belongs to tmux session `tm-tcode-01` (matches record 64d388b5).
- Pane `%1` belongs to a DIFFERENT tmux session named `tm-tcode` and has a LIVE `claude` child process (pid 92162, `claude --setting-sources project,local --dangerously-skip-permissions`) running Fable 5 — i.e., a second live Claude in the same worktree. The daemon's `f443c12d` record calls this pane its own but marks the record `stopped`.

## The Failure

`session_send` / `session_proxy_message` to `f443c12d-...` was REFUSED with:

> "recorded pane %1 for session f443c12d-... no longer exists, but its tmux session is still alive (likely a sibling window) — refusing to respawn into an unrelated active pane."

The pane `%1` demonstrably EXISTS and hosts a live claude; the daemon's pane-liveness check is wrong here. Direct `tmux send-keys -t %1` reached the session fine, confirming the pane is live and addressable — so the daemon's refusal is a false negative.

## Impact

1. **Broken inter-session coordination:** Inter-session coordination via the sanctioned MCP channel is impossible for the mis-mapped session, forcing a raw `tmux send-keys` workaround.
[...]

Issue #2268, opened 2026-07-09
feat(trusty-mpm): capture tmux window id at session pause, reattach on resume


bobmatnyc:
## Goal

Paused PM sessions should reattach to the originating tmux window on resume, instead of dropping the user into whatever window happens to be active at resume time.

## Approach

- **Capture at pause**: `tm-session-pause` records the tmux window identity — `#{session_name}:#{window_index}:#{window_id}` — into a new `## Tmux Window` snapshot section alongside the existing todos/git-state snapshot.
- **Parse + render**: the snapshot field is parsed and rendered via `trusty-common`'s `session_finder` module, which gains a new optional field for the window identity (additive, back-compat safe — old snapshots without the section still parse fine).
- **Reattach at resume**: `tm-session-resume` reads the captured window identity and runs `tmux select-window` to reattach the operator to the correct window when resuming, falling back gracefully when no window info is present (e.g. snapshots taken before this change, or non-tmux sessions).

## Scope

This is primarily a skill-docs change (`tm-session-pause.md`, `tm-session-management.md`, `tm-session-resume.md`) plus one additive `Option<String>` field and its parse/render logic in `crates/trusty-common/src/catchup/session_finder.rs` (+3 new tests). No breaking changes to existing snapshot formats.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5386, opened 2026-08-11
fix(trusty-mpm): resume resolves a snapshot by tmux window id when the session id misses

bobmatnyc:
## The problem

Two defects, one resume path.

### 1. A relaunch in the same tmux window resolves nothing

Four identifiers describe one running PM session and never converge:

| Layer | Observed value | Source |
|---|---|---|
| Claude Code harness session id | `69895d04-149d-4c31-a640-29048831f9a5` | harness config/scratchpad path |
| tmux window | `tm-dogfood:0:@230`, pane `%230`, pane pid 2200 | `tmux display-message` |
| trusty-mpm daemon record for that window | `e597b372-bbc7-419f-8576-79aeba268800` | `session_list`: `tmux_name: "tm-dogfood"`, `origin: "tmux"`, `pid: null`, `workdir: ""` |
| Claude session that WROTE the snapshot being resumed | `e262f4c5-d309-4203-ad3b-e0c29084d87e` | `.trusty-mpm/sessions/e262f4c5-.../session-20260811-004501.md` |

Reproduced live:

1. `session_status(session_id: "69895d04-...")` → `no such session`. The daemon has never seen the harness id.
2. `session_context_catchup(project_dir, session_id: "69895d04-...")` → `resolved_snapshot: null`, while `sessions[]` correctly lists two paused snapshots.
3. The daemon's record for the tmux window has `pid: null`, `workdir: ""` — it identifies the WINDOW, not the Claude process inside it.

**Effect:** `/tm-session-resume` cannot resolve a snapshot automatically across a relaunch in the same tmux window. It degrades to a human reading prose summaries to guess which of N snapshots is theirs — which is what happened, with two snapshots paused 19 seconds apart. Each relaunch also orphans another session directory.

### 2. The digest handed out the means to adopt any session (found in review)

`crates/trusty-common/src/catchup/json.rs:61-64` returned every paused session's `tmux_window` AND its exact snapshot `source_file` in `sessions[]`, to any caller, regardless of `session_id`. That is pre-existing code this PR did not introduce, and it is what made the new window fallback exploitable: a caller reads another session's `tmux_window` out of the response, hands it straight back as its own, and resolves that session's snapshot deterministically — no window-id reuse, no guessing. `/tm-session-resume` then has the PM adopt that snapshot as its own continuation. `source_file` is the same disclosure without the tool: the absolute path, readable directly.

**This means #5272's fix was incomplete.** #5272 removed the "latest overall" fallback so an unidentified caller could not inherit an arbitrary session's state. It left in place the data that reconstructs the same result by hand. Removing one code path while still publishing the material to re-create it is not the invariant #5272 was defending.

### Ownership gating on the digest (#5272's invariant, enforced on the response)

`redact_sessions_not_owned_by` runs over `sessions[]` before the response is built. A caller **owns** a session when it is attributed to the caller's `session_id` in `sessions-log.jsonl` (or sits in that id's directory), OR when the caller is in the tmux window that paused it — the same claim `resolve_snapshot_for_caller` accepts, so the digest and `resolved_snapshot` cannot disagree about the same row.

**What a non-owning caller can now see:** `format`, `paused_at`, `summary`, and `owned: false`.

**What it can no longer do:**

- Read `source_file` — so it cannot open another session's snapshot off disk.
- Read `tmux_window` — so it cannot pass another session's window back as its own and resolve that snapshot. This is the exploit above, closed at its source: the value is no longer in the response to copy.
- Read `in_progress`, `next_steps`, `git_context` — the state `/tm-session-resume` restores as its own todo list.

The line is drawn at what a resuming PM would ACT on. `source_file`/`tmux_window` are handles that load a snapshot; `in_progress`/`next_steps`/`git_context` are state a resume restores. `summary` stays because "something else paused here, and it was about X" is what the digest is for and nothing loads from it. `sessions[]` is not deleted — `/tm-session-resume` still renders it, and a caller can still see that other sessions exist.

The window fallback is unchanged in what it accepts: a caller that genuinely knows its window still resolves. What changed is that the tool no longer SUPPLIES the value for a claim the caller could not otherwise make. To read another session on purpose, pass its `session_id` — the explicit opt-in #5272 kept.

A legacy claude-mpm session carries no id attribution and no window, so it is never owned. It also carries no `source_file`, so nothing actionable is withheld — only its `in_progress`/`next_steps` fold.

`tm session catchup` (the CLI markdown digest) is unchanged. It renders one operator's own terminal, not a response to a remote caller, and `generate_catchup_json` still populates every field; redaction is a separate explicit step the MCP tool applies.

### Window-id fallback

`session_context_catchup` takes an optional `tmux_window` and resolves in this order:

1. exact `session_id` match — unchanged, always wins;
2. **new** — the newest paused snapshot under `project_dir` whose recorded `tmux_window` carries the caller's **window id**;
3. otherwise null.

The response gains `resolved_via`: `"session_id"`, `"tmux_window"`, or null, so a caller can tell an exact match from a fallback and say so when reporting what it resumed from.

Matching is on the `@230` window id, not `tm-dogfood:0`. Session names get renamed and window indexes renumber when a window is closed; the `@N` id is stable for the window's lifetime.

**Explicit parameter, not daemon-side derivation.** The daemon's record for a tmux-origin session carries `pid: null` and `workdir: ""`, so it cannot infer which window an MCP caller occupies — row 3 of the table above is exactly that failure. The caller reads its own window from `tmux display-message` and passes it.

**Limitation:** tmux reuses window ids. Kill window `@230` and a later window can be handed `@230` again, and it would then match the dead window's snapshot. That is an accepted, bounded risk. The project-path scope is the second gate — the scan only ever reads `project_dir`'s own store (`window_fallback_is_scoped_to_the_project_dir`). There is deliberately no liveness verification of the recorded window.

### `window_id_of` and colons in session names

tmux permits a literal `:` in a session name, which pushed the field past three components and made the caller's OWN window unmatchable. It failed closed (no match, never a wrong match), so it was a functionality gap, not a leak. `window_id_of` now reads the last `:`-delimited component and requires it to look like a window id (`@N`): `my:proj:0:@7` matches, and `a:b:c:d` still parses to nothing because `d` is not an `@id`. The existing malformed-input table passes unmodified.

### Directions not taken

- **(a) Key daemon sessions by the harness session id, with the window as an attribute.** Larger blast radius: it changes the daemon's session model, every record's identity, and the tmux-origin registration path — to fix a resolution miss in one MCP tool.
- **(c) A stable per-window workstream id with session ids as children.** Resembles the branch-is-the-workstream model already used for worktrees and is probably where this ends up, but it is a design program, not a fix for the reported defect.

### Prior art

- **#5272** (closed) — removed the "latest overall" fallback. This PR completes it: see above.
- **#2268** (closed) — captures the tmux window id at pause and reattaches via `tmux select-window`. This fix reads that already-captured field.
- **#3396** (closed) — duplicate session records and crossed pane mappings breaking `session_send`. Adjacent symptom, different mechanism.
- **#4337** (closed) — daemon record captured a subagent's id instead of the session's. Same identity-confusion family.

## Contract change

`session_context_catchup` gains an optional input (`tmux_window`) and two additive output fields (`resolved_via`, and `owned` per session entry). The documented "`resolved_snapshot` is null without `session_id`" statement stays true.

**Withholding fields from `sessions[]` is a behavior change for a non-owning caller**, not merely additive: a caller that previously read another session's `source_file`/`tmux_window`/`in_progress`/`next_steps`/`git_context` now gets null. That is the fix, and `owned` tells the caller which case it is in.

SemVer, corrected — the earlier draft of this body said "minor bump per the 0.x rule", which was wrong:

- **trusty-mpm is `1.3.6`**, a published 1.x crate. `OrchestratorBackend::session_context_catchup` gains a parameter, which is a breaking public-trait signature change, so it needs the **MAJOR** position (`2.0.0`). The 0.x rule does not apply. Per owner ruling this break is ACCEPTED — there are four such breaks already on main; no redesign was made to avoid it.
- **trusty-common is `0.30.1`**, a 0.x crate. `PausedSessionJson` gains a public field (`owned`) and `#[non_exhaustive]`, both breaking, so its breaking bump is the **MINOR** position (`0.31.0`). `CallerIdentity`, `redact_sessions_not_owned_by`, `resolve_snapshot_for_caller`, `ResolutionPath`, `ResolvedSnapshot`, `window_id_of` and `snapshots_attributed_to` are purely additive.

Both bumps are release-time work, not part of this PR.

## Tests — rung 5

Session lifecycle plus persisted state plus a public trait signature.

```
cargo fmt --all --check                                                        EXIT=0
cargo check -p trusty-mpm                                                      EXIT=0
cargo check --workspace                                                        EXIT=0
cargo clippy -p trusty-common --features catchup --all-targets -- -D warnings  EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings                        EXIT=0
cargo test -p trusty-common --features catchup   EXIT=0   449 passed; 0 failed; 8 ignored
cargo test -p trusty-mpm                         EXIT=0   5002 passed; 0 failed; 7 ignored (lib)
                                                          + 1601 passed x2 (bins), 0 failed anywhere
bash scripts/check_line_cap.sh              3890 files, 0 violations
bash scripts/check_sld.sh                   57 specs + 3205 code files, 0 errors, 0 warnings
bash scripts/check_capabilities.sh          up to date (7 generated files)
bash scripts/check_changelog_fragment.sh    OK trusty-common, OK trusty-mpm
```

`trusty-common`'s `catchup` module is feature-gated, so its gates run with `--features catchup` — a bare `cargo test -p trusty-common` never compiles this code.

### Changelog fragments

One per affected crate, one category per fragment:

- `crates/trusty-common/changelog.d/5386-catchup-ownership-gating.md` (Added)
- `crates/trusty-common/changelog.d/5386-window-id-colon-fixed.md` (Fixed)
- `crates/trusty-mpm/changelog.d/5386-session-resume-window-fallback.md` (Added)
- `crates/trusty-mpm/changelog.d/5386-catchup-digest-ownership.md` (Fixed)

### Red before green — the leak fix

`daemon::mcp_context::tests::session_context_catchup_withholds_a_non_owners_handles`: pause as session A from window `tm-dogfood:0:@230`, then catch up as session B, which owns nothing.

With the `redact_sessions_not_owned_by` call neutralised (head `9f40bc9a`'s behavior), everything else identical — the panic output IS the leak:

```
running 1 test
test daemon::mcp_context::tests::session_context_catchup_withholds_a_non_owners_handles ... FAILED

---- daemon::mcp_context::tests::session_context_catchup_withholds_a_non_owners_handles stdout ----
thread '...' panicked at crates/trusty-mpm/src/daemon/mcp_context.rs:524:9:
B must not receive A's snapshot path: {"format":"trusty-mpm","git_context":null,
"in_progress":"- halfway through X","next_steps":"- finish X","owned":true,
"paused_at":"2026-08-11T02:43:58Z",
"source_file":"/var/folders/.../.trusty-mpm/sessions/2eb72dca-de08-481b-8dfa-22ab7f81b1f9/session-20260811-024358.md",
"summary":"Session A's work.","tmux_window":"tm-dogfood:0:@230"}

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 5008 filtered out; finished in 0.09s
```

With the redaction restored:

```
running 1 test
test daemon::mcp_context::tests::session_context_catchup_withholds_a_non_owners_handles ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5008 filtered out; finished in 0.08s
```

The test also asserts B's whole response body contains neither `@230` nor A's snapshot path anywhere — there is nothing left to hand back — while A's own entry keeps `source_file`, `tmux_window`, `next_steps` and `owned: true`.

### Red before green — the window fallback

`daemon::mcp_context::tests::session_context_catchup_resolves_by_tmux_window_after_a_relaunch`: pause as `e262f4c5-…` from window `tm-dogfood:0:@230`, then catch up as `69895d04-149d-4c31-a640-29048831f9a5` — an id that never paused — passing the same window.

With the fallback body replaced by the pre-change `None`:

```
running 1 test
test daemon::mcp_context::tests::session_context_catchup_resolves_by_tmux_window_after_a_relaunch ... FAILED

---- ... stdout ----
thread '...' panicked at crates/trusty-mpm/src/daemon/mcp_context.rs:440:9:
  left: Null
 right: "/var/folders/.../.trusty-mpm/sessions/e262f4c5-d309-4203-ad3b-e0c29084d87e/session-20260811-010246.md"

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 5006 filtered out; finished in 0.05s
```

With the fallback restored:

```
running 1 test
test daemon::mcp_context::tests::session_context_catchup_resolves_by_tmux_window_after_a_relaunch ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5006 filtered out; finished in 0.12s
```

### Failure paths covered

Ownership gating:

- non-owner loses both handles and all restorable state, keeps `summary` — `redaction_withholds_handles_and_restorable_state`
- the redacted entry, serialized, contains neither the window id nor the path — `redaction_leaves_nothing_to_reconstruct_a_window_claim_from`
- the owning session id keeps every field — `owner_sees_every_field`
- the owning WINDOW keeps every field, so the digest agrees with `resolved_snapshot` — `window_owner_sees_every_field`, and end-to-end through the tool as `session_context_catchup_digest_agrees_with_the_window_fallback`
- an unattributable session (every legacy claude-mpm entry) is owned by nobody — `an_unattributable_session_is_owned_by_nobody`
- a session's attribution set is all of its snapshots, not just the newest — `snapshots_attributed_to_unions_log_and_directory`
- one session's set never contains another's; an unattributed flat file belongs to neither — `snapshots_attributed_to_excludes_other_sessions`

Window fallback (unchanged from the previous round, all still passing unmodified):

- exact-id match still wins when both an id match and a window match exist — `exact_session_id_match_wins_over_window_match`
- window match with a mismatched `project_dir` → null — `window_fallback_is_scoped_to_the_project_dir`
- malformed / 2-component / 4-component / empty `tmux_window` → no panic, no match — `malformed_window_fields_never_match`
- snapshot with a null `tmux_window` (older snapshots) skipped — `snapshot_without_a_recorded_window_is_skipped`
- window renamed and renumbered, same `@id` → still matches — `window_match_ignores_session_name_and_index`
- neither id nor window → null — `no_session_id_and_no_window_resolves_nothing`
- `tmux_window` reaches the backend through dispatch — `dispatch_session_context_catchup_forwards_tmux_window`
- a `:` in the session name no longer breaks the caller's own match — `window_id_of_tolerates_colons_in_the_session_name`


bobmatnyc:
## Verdict: BLOCK

Adversarial review of head `9f40bc9a5`, branch `fix/session-resume-window-id`. Rung 5 (session lifecycle + persisted state + public trait signature) per the repo's test ladder.

## Findings

| CRITICAL | `crates/trusty-common/changelog.d/` | — | New source `crates/trusty-common/src/catchup/resolve.rs` (299 lines) plus a one-line addition to `crates/trusty-common/src/catchup/mod.rs` ship with no `crates/trusty-common/changelog.d/*.md` fragment — only `crates/trusty-mpm/changelog.d/5386-session-resume-window-fallback.md` exists. `bash scripts/check_changelog_fragment.sh` fails locally: `FAIL trusty-common: crates/trusty-common/src/** changed with no changelog record`. The PR's own "Per-PR changelog fragment" CI check is red (`gh pr checks 5386`). CLAUDE.md states this tier equals a failing `cargo test`/`clippy` gate. | Add `crates/trusty-common/changelog.d/5386-session-resume-window-fallback.md` (an `Added` fragment for the new `resolve_snapshot_for_caller`/`ResolutionPath`/`window_id_of` public items). | Fix here |
| CRITICAL | `crates/trusty-common/src/catchup/resolve.rs:100`, `crates/trusty-mpm/src/mcp/tools/session.rs:302-304` | 100 / 302-304 | `tmux_window` is a plain caller-supplied string over MCP (tool schema: "The CALLER's own `session_name:window_index:window_id` from `tmux display-message`") with **zero server-side verification** that the calling process is actually in that window — `resolve_snapshot_for_caller` (`resolve.rs:87-103`) trusts it outright. This is compounded by pre-existing, unchanged code: `crates/trusty-common/src/catchup/json.rs:61-64` already returns every paused session's `tmux_window` *and* `source_file` (the exact snapshot path) unconditionally in `sessions[]`, for **any** caller regardless of `session_id` (`mcp_context.rs:58`, `"sessions": merged.sessions`). So a caller can read another session's exact `tmux_window` value straight out of the same tool call's own response, hand it back as its own `tmux_window`, and get `resolved_snapshot`/`resolved_via: "tmux_window"` pointed at that other session's file — deterministically, not by guessing a low-entropy `@N`. This is exactly the cross-session handoff #5272 closed, reopened through a channel the PR frames as an "ownership claim." The `/tm-session-resume` skill instructs the PM to resume from `resolved_snapshot` when `resolved_via` is set, so downstream this isn't a passive read — the PM adopts another session's state as its own continuation based on nothing but the caller's unverified self-report. The PR body itself concedes the design is honor-system ("The caller reads its own window from `tmux display-message` and passes it," "deliberately no liveness check"), but underestimates the exploit: it doesn't require window-id reuse or guessing, just reading the same response. | Bind `tmux_window` to something the daemon can verify server-side (resolve the connecting client's PID/pane via `tmux list-panes`, or an equivalent local-only proof), or drop the auto-trust in `resolved_via: "tmux_window"` and have the skill require explicit human confirmation before treating a window match as an exact one. | Fix here |
| HIGH | PR description, "## Contract change" | — | States `OrchestratorBackend::session_context_catchup` "gains a parameter — a breaking signature change that needs a minor bump at release, per the 0.x rule." `crates/trusty-mpm/Cargo.toml:3` shows `version = "1.3.6"`, and the crate is not `publish = false` — it is a published 1.x crate, not a `0.y.z` one. CLAUDE.md's 0.x rule ("for a `0.y.z` crate the breaking bump is the MINOR position") does not apply; ordinary SemVer requires a MAJOR bump (2.0.0) for a breaking public-trait method signature change on a >=1.0 crate. `trusty-common`'s actual additions (`ResolutionPath`, `ResolvedSnapshot`, `resolve_snapshot_for_caller`, `window_id_of`) are purely additive at 0.30.1 and need no breaking bump at all — the PR body conflates the two crates. | Correct the PR body to state trusty-mpm needs a MAJOR bump (2.0.0), not a minor one. | Fix here |
| LOW | `crates/trusty-common/src/catchup/resolve.rs:140-147` | 140-147 | `window_id_of` requires exactly 3 `:`-delimited components. A tmux session name containing a literal `:` pushes the field to 4+ components and the caller's own snapshot never matches. Fails CLOSED (no match, not a wrong match) — a functionality gap, not a security bug. | Note the limitation in the doc comment, or split on the last two `:` occurrences instead of all of them. | Parent |

## Required Changes

1. Add the missing `crates/trusty-common/changelog.d/` fragment.
2. Add server-side verification (or remove blind auto-trust) for the `tmux_window` fallback before this can resolve another session's snapshot on nothing but a caller's say-so.
3. Correct the PR body's SemVer bump claim (major, not minor, per trusty-mpm's actual 1.3.6 version).

## Notes

- Items verified clean: the fail-open coverage on malformed `tmux_window` input is real (`window_id_of` is new code; the malformed-input test table would catch a plausible wrong implementation, e.g. "take the text after the last `:`", which would wrongly return `"d"` for `"a:b:c:d"` instead of the expected `None`). The `session_id`-wins-over-window-match ordering is correctly tested (`exact_session_id_match_wins_over_window_match`). Project-path scoping is real and tested (`window_fallback_is_scoped_to_the_project_dir`) — it just isn't the second gate the PR body claims, per the CRITICAL finding above. The two real trait implementors (`mcp_backend.rs`, `mcp/tests.rs`) are both updated; `mcp_proxy.rs`/`mcp_bugreport.rs` are doc-comments pointing at `mcp_backend.rs`'s single impl, not separate implementors — the trait-consumer sweep is complete. The window-id-reuse limitation IS disclosed where a caller will see it (`tm-session-resume.md`: "Window ids are reused after a window is killed and recreated, so a `tmux_window` match is an ownership claim, not a guarantee").
- The `sessions[].tmux_window` / `.source_file` full disclosure that amplifies the CRITICAL authorization finding is pre-existing, unchanged code (`crates/trusty-common/src/catchup/json.rs` has zero diff in this PR) — flagged here because it directly undercuts this PR's own "ownership claim, not a guess" security argument, not as something this PR introduced. Worth a separate look at whether `session_context_catchup` should be gating full session content behind `session_id` ownership at all, independent of this PR.

bobmatnyc:
## Verdict: APPROVE

Re-review of head `234cc06d`, round 2. Rung 5 (session lifecycle + persisted state + public trait signature).

## Round-1 CRITICALs — both closed

**1. Changelog fragment (was CRITICAL).** `bash scripts/check_changelog_fragment.sh` run against this head: `OK trusty-common`, `OK trusty-mpm`, `changelog-fragment gate: scanned 17 changed path(s); all 2 crate(s) with source changes are recorded.` Exit 0. Four fragments exist as claimed: `crates/trusty-mpm/changelog.d/{5386-catchup-digest-ownership,5386-session-resume-window-fallback}.md`, `crates/trusty-common/changelog.d/{5386-catchup-ownership-gating,5386-window-id-colon-fixed}.md`. Closed.

**2. The disclosure leak (was CRITICAL).** `crates/trusty-common/src/catchup/json.rs` now carries `owned: bool` on `PausedSessionJson` (`#[non_exhaustive]`), and `crates/trusty-common/src/catchup/resolve.rs:215-282` (`redact_sessions_not_owned_by` / `is_owned_by` / `withhold`) strips `source_file`, `tmux_window`, `in_progress`, `next_steps`, `git_context` from any digest entry the caller doesn't own, leaving only `format`/`paused_at`/`summary`/`owned:false`.

Verified rather than trusted:
- **Applied on every path.** `mcp_context.rs:151-159` calls `redact_sessions_not_owned_by(dir, &caller, &mut digest.sessions)` *inside* the `for dir in &project_dirs` loop, before `merged.absorb(digest)` — so `all_projects: true` gets the same per-project redaction, not a bypass. There is no second serializer and no `full: true` branch that skips it (`full` only affects the watermark window, not the redaction call).
- **Fail-closed, not fail-open.** `is_owned_by` (resolve.rs:247) defaults to `false` on every branch it doesn't explicitly match; `redact_sessions_not_owned_by` defaults `owned_paths` to empty when `session_id` is absent. No error path defaults to `owned: true`. Confirmed by `an_unattributable_session_is_owned_by_nobody`.
- **`resolved_snapshot` / digest consistency.** `session_context_catchup_digest_agrees_with_the_window_fallback` asserts the row `resolved_snapshot` points at is never the row that got redacted — ran it, passes.
- **Ran the new/changed test surface directly** (not just trusting the PR body): `cargo test -p trusty-common catchup::` → 115 passed, 0 failed. `cargo test -p trusty-mpm mcp_context::` → 10 passed, 0 failed (all six new #5386 tests included). `cargo test -p trusty-mpm mcp::tests::` → 93 passed, 0 failed. `cargo clippy -p trusty-common --features catchup --all-targets -- -D warnings` → clean.

**Did the fix move the trust boundary rather than close it?** The window-ownership route is still a self-reported, server-unverified `@N` claim — see Findings below (MEDIUM, not CRITICAL). What round 1 actually found CRITICAL was that the response *itself* handed any caller the exact `tmux_window`/`source_file` values needed to replay another session deterministically, with zero guessing. That channel is gone: `redaction_leaves_nothing_to_reconstruct_a_window_claim_from` asserts the redacted entry's serialized JSON contains neither the window id nor the snapshot path. What's left is a caller having to already know or guess a low-entropy `@N` — a materially smaller and already-disclosed risk (see MEDIUM finding below), not a reopening of the deterministic hole.

## Findings

| MEDIUM | `crates/trusty-common/src/catchup/resolve.rs` | 247-263 (`is_owned_by`) | Window-based ownership is a self-reported, server-unverified `@N` claim. A local caller can enumerate small integers as `tmux_window` and, on a hit, receive full disclosure (source_file, in_progress, next_steps, git_context) for a session it doesn't actually own. Not a regression from round 1 — this route was already reviewed and its risk already disclosed in `tm-session-resume.md` ("a `tmux_window` match is an ownership claim, not a guarantee") — but it is the residual surface once the deterministic replay hole is closed. | Track as follow-up: bind `tmux_window` to something server-verifiable (e.g. cross-check the connecting PID's pane via `tmux list-panes`), or rate-limit `session_context_catchup` calls that vary only `tmux_window`. | Promote |
| MEDIUM | `crates/trusty-common/src/catchup/resolve.rs` | 274-282 (`withhold`) | Author's open question: `summary` stays visible for non-owned sessions and can contain branch names, issue numbers, file paths, incident detail. **Ruling: WARN, not BLOCK.** The line the author drew (withhold only what a resume would *act on*) is defensible: `summary` doesn't hand back an exploitable handle, and under this daemon's loopback-only trust model (ADR-0018) the same content is already readable by any local process directly off the `.md` snapshot file — the MCP redaction was never a confidentiality boundary against a co-resident process, only a guard against the PM's own automatic cross-session bleed. Worth tightening later, not worth blocking this PR on. | If tightened: `session.summary = None;` added to `withhold()`, one line. | Parent |

## Round-1 HIGH and LOW — both closed correctly

- **HIGH (SemVer).** PR body now states trusty-mpm (`1.3.6`, confirmed in `Cargo.toml:3`) needs a MAJOR bump (`2.0.0`) for the breaking `OrchestratorBackend::session_context_catchup` signature change — correct, the crate is >=1.0 so the 0.x rule doesn't apply. For trusty-common (`0.30.1`, confirmed): body says `PausedSessionJson` gaining a public field *and* `#[non_exhaustive]` are "both breaking," bump target MINOR (`0.31.0`) — correct, and it correctly flags the double-break. I confirmed `PausedSessionJson` predates this PR: `git log --diff-filter=A` shows it introduced in `9d965569` (2026-07-20), and `git merge-base --is-ancestor 9d965569 trusty-common-v0.30.0` succeeds — so it shipped in a published release before this PR, and adding `#[non_exhaustive]` to it now is genuinely a second break on top of the new field, exactly as the body says. Closed.
- **LOW (colons).** `window_id_of` (resolve.rs) now reads the last `:`-delimited component and requires an `@`-prefixed, length>1 tail. Ran `malformed_window_fields_never_match` — the full table (`""`, `"   "`, `"main"`, `"main:0"`, `"a:b:c:d"`, `":"`, `"::"`) still resolves to `None` unmodified, and the new `window_id_of_tolerates_colons_in_the_session_name` test confirms `"my:proj:0:@7"` now resolves. Closed.

## Also checked

- **Build artifacts.** `git diff main...pr-5386-review --stat -- crates/trusty-analyze/ui/dist crates/trusty-memory/ui/package-lock.json` is empty — neither path appears in the diff. Confirmed clean.
- **Round-1 surviving strengths, re-verified live:** malformed-input table (above), `exact_session_id_match_wins_over_window_match` passes, both trait implementors (`daemon/mcp_backend.rs`, `mcp/tests.rs`) carry the new `tmux_window` parameter and both compile/test clean.

## Notes

- No CRITICAL, no HIGH findings remain. Per the repo's round-3 doctrine and the verdict protocol (APPROVE = zero CRITICAL, zero HIGH), this is APPROVE — the two MEDIUM items are tracked, not blocking, and reflect pre-existing, already-disclosed design tradeoffs rather than defects introduced by this round's fix.
- Version bumps (trusty-mpm → 2.0.0, trusty-common → 0.31.0) are release-time work per the PR body and CLAUDE.md — not required in this PR.

Issue #5239, opened 2026-08-08
feat(trusty-review/tga): AUDIT gap reporting


bobmatnyc:
Orchestrator-side check turning `analyze_adapter.rs`'s silent `None` (fail-open by design, unchanged) into a named line in the report's Gaps & Caveats section. Apply the one-shot continue-on-failure default instead of `tga analyze`'s abort-by-default: a repo that failed to collect is excluded from its own scorecard, named explicitly in Gaps & Caveats, and the run continues.

See `docs/specs/DOC-67-tga-audit-mode.md` §9 (Determinism and Failure Behavior) and §13 (row 6).

Depends on: [#5238](https://github.com/bobmatnyc/trusty-tools/issues/5238).

**Blocked:** implementation held until tm 1.3.5 ships (owner directive, 2026-08-08). Do not start work on this before that release lands.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5390, opened 2026-08-11
fix(trusty-review): scrub credentials from DD-report findings (#5323)

### LINKED ISSUE #5323: fix(trusty-review): analyze-adapter findings reach the DD report unscrubbed
trusty-review's analyze-adapter path reaches the generated due-diligence report with no credential scrubbing, in parallel to tga's path which has it. Found during security review of #5322.

- tga scrubs every string it emits: [`dd_manifest.rs:199`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-git-analytics/src/report/dd_manifest.rs#L199) runs `configured_secrets(cfg)` inside `build_dd_manifest`.
- `MetricFinding.description` / `.remediation` ([`metrics.rs:145-164`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-review/src/report/metrics.rs#L145-L164)) never touch it. Populated verbatim at [`analyze_adapter.rs:206-207`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-review/src/report/analyze_adapter.rs#L206-L207) and [`:236-237`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-review/src/report/analyze_adapter.rs#L236-L237), copied into template fields by `FindingRow::from_metric` at [`reporter.rs:395-396`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-review/src/report/reporter.rs#L395-L396), and the investigation-derived twin does the same at [`investigate/mod.rs:339-340`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-review/src/report/investigate/mod.rs#L339-L340).
- Grepping `crates/trusty-review/src/` for `redact`/`scrub`/`configured_secrets` finds only unrelated hits: `pipeline/finding_hygiene.rs` filters LLM self-negation text, and `report/benchmark.rs`'s "redact" truncates a path to its basename.

Not the same defect as #5239 (closed) — that one was a scrub applied in the wrong order (truncate-before-scrub) on tga's path. This is a path never wired to a scrub at all.

Not an active leak today: `rationale` ([`refactor.rs:174-185`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-analyze/src/core/refactor.rs#L174-L185)) is built only from `cyclomatic`/`grade`/fixed smell labels, `suggested_action` ([`refactor.rs:132-171`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-analyze/src/core/refactor.rs#L132-L171)) is a template interpolating a parsed symbol name and line numbers, and `message` ([`tools.rs:49`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-analyze/src/core/tools.rs#L49)) is parsed linter output — none carry raw source text. The gap is that the safety property rests on the upstream generator happening not to emit raw text, not on an enforced boundary, and the artifact is handed to acquirers. A richer linter message format or a rationale that quotes source would silently turn this into a leak with no gate to catch it.

Fix: route `MetricFinding.description`/`.remediation` and the investigation-derived twin through the same `configured_secrets` redaction tga uses. Scrub before any truncation, never after — `scrub_secrets` is an exact full-value substring match and a credential split across a truncation boundary survives as an unmatched fragment (the #5239 lesson).

Open design question, not a prescription: trusty-review doesn't currently receive the tga config that `configured_secrets` derives its needle set from, so this may need a way to pass the secret set across the subprocess boundary — doing that via argv (visible to `ps`) would be worse than the problem.

bobmatnyc:
Closes #5323

## What changed

trusty-review's analyze path reached the generated DD report with no credential
scrubbing, in parallel to tga's path which has had it since #5239. Three
producers put findings on the model, and all three copied producer-authored text
verbatim into the rendered finding bands, the executive summary, the synthesis
digest sent to an LLM provider, and the JSON twin:

| Producer | Site |
|---|---|
| Live analyze daemon | `analyze_adapter.rs::enrich_with_analyze_gaps` |
| Manifest-declared metrics JSON | `model.rs::build_repository` |
| Investigation twin | `investigate/mod.rs::apply_investigation` |

New `report::redact` module. `report_secrets` derives the needle set from
`trusty_common::credentials::resolved_secret_values` — the same provider-registry
walk and env > `.env.local` > store precedence every other consumer uses. That
answers the ticket's open design question without a subprocess boundary: the
registry already covers the forge/tracker/inference credentials tga's config
holds, so nothing needs to be passed by argv. `scrub_metrics` runs where findings
ENTER the model, ahead of every downstream truncation (`dedupe_field`, the
benchmark's basename trim) — the #5239 ordering lesson.

## Instance vs. shape

The ticket named `description` and `remediation` on two producers. This closes
the shape for the `AnalyzeMetrics` type: all five producer-supplied strings on a
finding (`title` is a linter's rule code, `category` its tool name, `component` a
path), plus the daemon-authored bucket labels, language names, and repository
label — across all three producers, including the declared-metrics-file path the
ticket did not name.

What it does not do: make a *fourth* future producer structurally impossible to
miss. The scrub is one function called from three sites, not a type invariant.
Producer-side was chosen over consumer-side because there are three producers and
four consumers, and the earliest consumer is the LLM digest — a scrub at render
time would miss it.

## Fail-open finding

**This change adds no failure branch.** `scrub_metrics` and `scrub_finding` are
infallible, and `report_secrets` returns `Vec<String>` with no `Result`. There is
therefore no error arm to regression-test.

One pre-existing fail-open sits underneath it and is worth naming: `default_store`
(trusty-common) falls back to `MemoryKeyStore` when the keychain/file store is
unavailable, so a machine that cannot open its store yields a smaller needle set
and a weaker scrub, silently. That is trusty-common's behaviour, unchanged here.
The honest limit is documented on the module: scrubbed text is lower-risk, not
proven secret-free.

## Test ladder — rung 5 (security + cross-crate contract)

Regression proof. The three wiring tests fail with the call sites removed and
pass with them in place:

```
running 7 tests
test report::redact::tests::scrub_finding_covers_every_producer_supplied_field ... ok
test report::redact::tests::scrub_metrics_reaches_every_string_field ... ok
test report::redact::tests::scrub_is_a_noop_with_no_needles ... ok
test report::redact::tests::report_secrets_yields_a_usable_needle_set ... ok
test report::redact::tests::enrich_scrubs_configured_credentials_from_findings ... FAILED
test report::redact::tests::declared_metrics_file_findings_are_scrubbed ... FAILED
test report::redact::tests::apply_investigation_scrubs_configured_credentials ... FAILED

thread '...enrich_scrubs_configured_credentials_from_findings' panicked at redact_tests.rs:109:9:
enrich_with_analyze_gaps: credential survived into `repo-ghp_5323FAKETOKENvalue0001`
thread '...declared_metrics_file_findings_are_scrubbed' panicked at redact_tests.rs:109:9:
ReportModel::build: credential survived into `repo-ghp_5323FAKETOKENvalue0001`
thread '...apply_investigation_scrubs_configured_credentials' panicked at redact_tests.rs:109:9:
apply_investigation: credential survived into `hardcoded ghp_5323FAKETOKENvalue0001`

test result: FAILED. 4 passed; 3 failed; 0 ignored; 0 measured; 1609 filtered out
```

With the fix: `test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 1609 filtered out`

Gates run:

| Command | Result |
|---|---|
| `cargo clippy -p trusty-review --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-review` | `1611 passed; 0 failed; 5 ignored` (+ 8 integration targets, all ok) |
| `cargo check --workspace --all-targets` | clean |
| `cargo test -p trusty-analyze --features review` | `474 passed; 0 failed` (+ 5 targets, all ok) |
| `cargo test -p trusty-mpm` | `4997 passed; 1 failed` — see below |

`cargo test -p trusty-mpm -- --include-ignored` was not run for this path.
trusty-review's 5 ignored tests are `live_fireworks_structured_output_enforces_enum`,
`stage_c_live_bedrock`, `run_review_live_post_and_dedup_skip_integration`,
`deferred_dispatch_tool_call_times_out_warming_up`, and
`stdio_serve_tools_list_returns_three_tools` — all live-network or binary-stdio;
none touches the report pipeline.

Every `scripts/check_*.sh` in the repo, dispositioned:

| Script | Result |
|---|---|
| `check_changelog_fragment.sh` | ran — OK (fragment present and valid) |
| `check_line_cap.sh` | ran — OK (3891 files, 0 violations) |
| `check_test_pointers.sh` | ran — OK (23584 citations, 0 dangling) |
| `check_sld.sh` | ran — OK (0 errors, 0 warnings) |
| `check_doc_numbers.sh` | ran — OK |
| `check_generated_regions.sh` | ran — OK |
| `check_public_docs.sh` | ran — OK |
| `check_buildrs_sync.sh` | ran — OK |
| `check_capabilities.sh` | ran — OK |
| `check_generation_artifacts.sh` | ran — OK |
| `check_agent_assets.sh` | ran — OK |
| `check_semver.sh` | not run — release-time gate, not per-PR; this PR adds public items only (`report::redact`), no breaking change |
| `check_line_cap_selftest.sh`, `check_public_docs_selftest.sh`, `check_semver_selftest.sh`, `check_scan_floor_selftest.sh` | not run — self-tests of the gates themselves, untouched by this PR |

## Pre-existing failure

`trusty-mpm::core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning`
fails under parallel execution and passes in isolation (`1 passed; 0 failed`).
This is the tracked #4931. `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/`
is empty — this branch touches zero files in that crate.


bobmatnyc:
## Round 2 — both findings fixed

### CRITICAL — the investigation scrub was discarded before render

Confirmed exactly as described. `merge_investigation_prose` built `FindingProse`
from the raw `VerifiedFinding`, and `FindingRow::merge_prose` overwrote the
scrubbed metrics prose with it unconditionally. Both run in the same
`cli_report.rs:448-452` branch, so round 1's scrub never survived to the page on
any run producing a RED/AMBER finding.

I took the second option you offered — scrub the record where it enters each
sink, not at `run_investigation` — because a test cannot produce an
`Investigation` through `run_investigation` without a live LLM. Scrubbing only at
the producer would have left the bypass at `merge_investigation_prose` structurally
untestable, which is how it got missed the first time. Both entry points are
`pub`, both are directly drivable, and each covers a sink the other cannot:

- `apply_investigation` scrubs a local copy of the whole `Investigation` and
  stores THAT on the model — so the derived `MetricFinding`s, the JSON twin, and
  `investigate::render`'s coverage/dependency sections are all clean. This also
  closes a leak neither of us named: `model.investigation = Some(inv.clone())`
  was serializing the raw record, including the `Unavailable(reason)` string that
  can quote a provider's error body.
- `merge_investigation_prose` scrubs each `FindingProse` before the push. This is
  the only place `evidence_quote` is ever scrubbed — it has no metrics route.

On scrubbing a quote documented as byte-for-byte verbatim: that is the intended
trade, stated at `redact.rs`. A quote is verbatim so a reader can match it against
source; a credential is the one substring that must not be matchable.

### MEDIUM — `schema_version`

Fixed the field, not the claim. `scrub_metrics` now covers it.

### Test — this is the part that mattered

You were right that all three round-1 tests pass with the bypass fully in place.
New `investigation_credentials_never_reach_the_rendered_report` drives the real
`cli_report::run_synthesis` order — `apply_investigation` → `Synthesis` →
`merge_investigation_prose` → `Reporter::render` — and asserts on the rendered
markdown and `serde_json::to_string(&model)`, not on an intermediate.

Pre-fix, with ONLY the `scrub_prose` call removed and everything else in place:

```
running 10 tests
test report::redact::tests::scrub_prose_covers_evidence_and_every_narrative_field ... ok
test report::redact::tests::scrub_finding_covers_every_producer_supplied_field ... ok
test report::redact::tests::scrub_metrics_reaches_every_string_field ... ok
test report::redact::tests::scrub_is_a_noop_with_no_needles ... ok
test report::redact::tests::scrub_investigation_reaches_the_whole_tree ... ok
test report::redact::tests::enrich_scrubs_configured_credentials_from_findings ... ok
test report::redact::tests::report_secrets_yields_a_usable_needle_set ... ok
test report::redact::tests::declared_metrics_file_findings_are_scrubbed ... ok
test report::redact::tests::apply_investigation_scrubs_configured_credentials ... ok
test report::redact::tests::investigation_credentials_never_reach_the_rendered_report ... FAILED

thread '...investigation_credentials_never_reach_the_rendered_report' panicked at
crates/trusty-review/src/report/redact_tests.rs:350:5:
credential rendered into the report markdown

test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 1609 filtered out
```

Nine green while the credential renders — the round-1 failure mode, reproduced.
With the fix: `test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 1609 filtered out`

### The other two wiring tests

`enrich_scrubs_...` asserts on `model.repositories[0].metrics`. That IS the
rendered surface for its producer — the analyze daemon path has no synthesis
counterpart to overwrite it, so `from_metric` is what reaches the page. It would
catch a regression at the adapter. It would not catch a `merge_prose`-style
overwrite, which is why the new test exists rather than being folded into it.

`declared_metrics_file_findings_are_scrubbed` has the same shape and the same
limit. I strengthened `apply_investigation_scrubs_configured_credentials` to also
assert on the serialized `model.investigation` rather than metrics alone.

The honest general limit, unchanged: this is N functions called at N sinks, not a
type invariant. A future fifth sink is still forgettable. What round 2 buys is
that both sinks that exist are covered and both are asserted at the artifact.

### Version

Rebased onto `origin/main` (`90ce9a2d`..). `crates/trusty-review/Cargo.toml` still
read `0.14.0`, so #5387 has not landed — set to **0.14.1** per your instruction.

### Gates

| Command | Result |
|---|---|
| `cargo clippy -p trusty-review --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-review` | `1614 passed; 0 failed; 5 ignored` + 8 integration targets, all ok |
| `cargo check --workspace --all-targets` | clean |
| `cargo test -p trusty-analyze --features review` | `474 passed; 0 failed` + 5 targets, all ok |
| all 11 applicable `scripts/check_*.sh` | OK |

Fail-open finding unchanged: this change adds no failure branch. `scrub_*` are
infallible and `report_secrets` returns `Vec<String>`, not `Result`.

bobmatnyc:
## Round 3 — rebased onto #5387

Head: **`4e17e119`** (rebased onto `11ba79f0`).

### Version — resolved to 0.14.1, and my diff no longer carries it

The rebase produced **no conflict**. Both sides had moved `0.14.0 → 0.14.1`
identically, so git dropped my hunk as already-applied. `crates/trusty-review/Cargo.toml`
now reads `0.14.1`, inherited from #5387, and this branch contributes no
`Cargo.toml` or `Cargo.lock` change at all. That is the correct end state for the
rule as you stated it — 0.14.1 is unpublished, `check-pr-version-bump.sh` compares
against crates.io (0.14.0), and a second `src/**` change accumulates under the same
unpublished number. I did not invent 0.14.2. If the check disagrees on this head I
will report its output rather than bumping.

I amended the commit message, which still claimed a version bump this diff no
longer makes.

### #5387's changes verified intact

`VerifyOutcome::reader_caveat` is present at `models/mod.rs:233`, with 3 references
in `integrations/github/inline.rs` (`:43`, `:356`, `:373`) and 6 in `models/tests.rs`.

One correction to your brief: **my diff does not touch `models/mod.rs`.** It touches
`report/mod.rs` — the report submodule's own `mod.rs`, a different file. There was
never an overlap to clobber, which is also why the rebase was clean.

### Final diff — 7 files, all trusty-review

```
crates/trusty-review/changelog.d/5323-scrub-dd-report-findings.md |  19 +
crates/trusty-review/src/report/analyze_adapter.rs                |  15 +-
crates/trusty-review/src/report/investigate/mod.rs                |  72 +++-
crates/trusty-review/src/report/mod.rs                            |   2 +
crates/trusty-review/src/report/model.rs                          |  31 +-
crates/trusty-review/src/report/redact.rs                         | 243 ++++++
crates/trusty-review/src/report/redact_tests.rs                   | 446 ++++++++++
7 files changed, 802 insertions(+), 26 deletions(-)
```

Staged explicitly by path this round. `git status` was clean after
`cargo check --workspace --all-targets` — no UI-dist or `package-lock.json` dirt
to restore.

### Gates on `4e17e119`

| Command | Result |
|---|---|
| `cargo clippy -p trusty-review --all-targets -- -D warnings` | clean |
| `cargo test -p trusty-review` | `1621 passed; 0 failed; 5 ignored` + 8 integration targets, all ok |
| `cargo check --workspace --all-targets` | clean |
| `cargo test -p trusty-analyze --features review` | `474 passed; 0 failed` + 5 targets, all ok |
| all 11 applicable `scripts/check_*.sh` | OK |

The redaction suite on top of #5387's changes:

```
running 10 tests
test report::redact::tests::scrub_finding_covers_every_producer_supplied_field ... ok
test report::redact::tests::scrub_prose_covers_evidence_and_every_narrative_field ... ok
test report::redact::tests::scrub_is_a_noop_with_no_needles ... ok
test report::redact::tests::scrub_metrics_reaches_every_string_field ... ok
test report::redact::tests::scrub_investigation_reaches_the_whole_tree ... ok
test report::redact::tests::declared_metrics_file_findings_are_scrubbed ... ok
test report::redact::tests::report_secrets_yields_a_usable_needle_set ... ok
test report::redact::tests::apply_investigation_scrubs_configured_credentials ... ok
test report::redact::tests::enrich_scrubs_configured_credentials_from_findings ... ok
test report::redact::tests::investigation_credentials_never_reach_the_rendered_report ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 1616 filtered out
```

bobmatnyc:
**code-critic re-review — APPROVE** (head `4e17e119`)

Both round-1 findings are closed.

**CRITICAL closed.** `scrubbed_prose` (`report/investigate/mod.rs:370-386`) now scrubs each `FindingProse` — including `evidence_quote` — before it reaches `synthesis.findings`, so `FindingRow::merge_prose`'s unconditional overwrite (`reporter.rs:447-457`) now overwrites with already-scrubbed values. Verified load-bearing by disabling the `scrub_prose` call in a scratch worktree and re-running `cargo test -p trusty-review redact::`: 9 passed, 1 failed — `investigation_credentials_never_reach_the_rendered_report` panics with "credential rendered into the report markdown". Restored: 10/10 pass.

**MEDIUM closed.** `AnalyzeMetrics.schema_version` is scrubbed in `scrub_metrics` (`redact.rs:118`), covered end to end by `declared_metrics_file_findings_are_scrubbed`.

**Fifth-sink sweep — none found.** `investigate/render.rs::report_sections`, `coverage_prompt_summary`, `synthesize_prompt.rs:272-277` and `synthesize_digest.rs::gather_compact_findings` all read surfaces that are now scrubbed. `apply_investigation` stores a scrubbed clone rather than the raw `inv` parameter — a third leak fixed in this round, confirmed by diff.

**Verbatim-quote semantics.** The scrub replaces only the matched substring with a visible `[REDACTED]` marker, so a reader sees an explicit marker rather than a silently altered quote.

**Rebase integrity.** Version reads 0.14.1 (bumped by #5387), no `Cargo.lock` diff. `models/mod.rs` and `integrations/github/inline.rs` are byte-identical between `11ba79f0` and `4e17e119` — nothing clobbered.

No new findings above 80% confidence.

Issue #3950, opened 2026-07-25
fix(trusty-mpm): MCP trust-set membership must require actual per-run pin success, not just a toggle (fifth instance of the name-approval/content-pinning class)


bobmatnyc:
## Summary

Fifth instance of the MCP name-approval/content-pinning vulnerability class in this codebase. Found by code-critic while approving PR #3946 (merged `df517417`, closed #3934).

**Lineage:** #3918 (trust derived from `.mcp.json` keys) → #3924 (two builtins had no force-overwrite injector) → #3926 (same derivation bug on `tm launch`) → #3934 (manifest toggle disabled an injector while its name stayed approved) → **this issue (fifth)**.

**The invariant these all violate in different ways:** a name may be in `enabledMcpjsonServers` ONLY IF something pinned the content behind that name in the same run. `enabledMcpjsonServers` approval is name-based and content-blind, so any decoupling of approval from pinning is an RCE path.

## The fifth instance

`crates/trusty-mpm/src/core/session_launch/mod.rs:656,676` (pre-fix) — `inject_trusty_memory_mcp` / `inject_trusty_search_mcp` failures are NON-FATAL (`tracing::warn!` and continue, by design — the session must still launch). But `plan.inject_trusty_memory` / `inject_trusty_search` — the boolean that gated trust-set membership — reflected only the MANIFEST TOGGLE, not whether the injector's write actually SUCCEEDED this run.

So if the toggle was on but `std::fs::write` failed (disk full, permission error, transient I/O fault) while a spoofed or stale `.mcp.json` entry was already present, the name still entered `enabledMcpjsonServers` even though content was never re-pinned. That is the invariant violated again, by a fifth route.

The same audit found the identical gap in the two **unconditional** injectors added by #3924 (`inject_trusty_mpm_mcp` / `inject_trusty_review_mcp`, `mod.rs:698-703`): they have no manifest toggle at all, so before this fix their names were unconditionally added to the trust set regardless of whether the force-overwrite write actually succeeded this run.

**Severity: MEDIUM, not CRITICAL** — a git-tracked repo cannot reliably force a local write failure on an arbitrary machine, so it is not attacker-reachable through committed repo content alone. But it is a genuine gap in the stated guarantee, and this vulnerability class has now bitten five times, so it is closed structurally rather than patched narrowly.

## Fix

[...]

Issue #3926, opened 2026-07-25
security(trusty-mpm): tm launch auto-trusts every MCP server declared in a repo's committed .mcp.json


bobmatnyc:
## Summary

`tm launch` silently pre-approves **every** MCP server declared in a project's
git-tracked `.mcp.json`, except the narrow set of project-scope custom MCP names.
An MCP server declaration committed to a repository is therefore auto-trusted and
connected the first time an operator runs `tm launch` in a clone of that repo —
no consent dialog, no `tm project trust` check.

This is **pre-existing and live** (present in the published trusty-mpm 1.0.2). It
is *not* introduced by PR #3924; it was found while reviewing that PR's
structurally identical defect on the daemon-managed path.

## Evidence

`crates/trusty-mpm/src/core/session_launch/settings.rs:651-663`
(`preseed_workspace_trust`):

```rust
let enabled_mcp = Value::Array(
    mcp_server_names(workspace)          // every key in <workspace>/.mcp.json
        .into_iter()
        .filter(|name| !exclude_mcp_names.contains(name))
        .map(Value::String)
        .collect(),
);
```

and then, unconditionally:

```rust
entry.insert("hasTrustDialogAccepted".to_string(), Value::Bool(true));
entry.insert("enabledMcpjsonServers".to_string(), enabled_mcp);
```

The only caller is `session_launch/mod.rs:710`:

```rust
preseed_workspace_trust_home(project_dir, &project_scope_mcp_names)
```

so `exclude_mcp_names` contains **only** project-scope custom MCP names (the
issue #2739 follow-up). Every other name present in `.mcp.json` is pre-approved.

`mcp_server_names()` reads the raw key set of `<workspace>/.mcp.json`.
`inject_mcp_server` (`settings.rs:377-419`) deliberately *preserves unrelated
servers*, and `.mcp.json` is git-tracked in real projects — so the key set
routinely contains entries that trusty-mpm's own injectors never wrote and that
passed no trust gate. `is_project_trusted` gates only the trusty-mpm manifest's
`[mcp.custom]` table (`custom_mcp.rs:259-289`) — a *different* entry point into
the same file.

Origin: the pre-approval behavior dates to issue #1296 ("so the 'new MCP servers
found' dialog never blocks the spawned session"). Issue #2739's follow-up added
`exclude_mcp_names` for exactly this class of risk, but scoped it only to
project-scope custom bridge targets. Its own comment states the principle:

[...]

PR #5398, opened 2026-08-11
fix(trusty-mpm): relocate CLAUDE_CONFIG_DIR for tm launch and tm connect (#4181, PR B of 3)

bobmatnyc:
PR **B of three** toward [ADR-0042](../blob/main/docs/adr/0042-mcp-configuration-is-static-and-persistent.md) / #4181. It relocates `CLAUDE_CONFIG_DIR` for `tm launch` and `tm connect` and closes the #2246 Keychain risk that relocation creates. **It deletes no MCP injector and no `enabledMcpjsonServers` write** — that pairing is PR C, and splitting it would reintroduce #3926 at maximum blast radius.

## The decision, and the alternative that was rejected

ADR-0042 decision 6 makes this a precondition of the deletion, not a follow-up: under a user-scope-only MCP declaration, `tm launch` and `tm connect` — which spawn `claude --setting-sources project,local` and do not relocate — are left with no MCP servers at all.

Two ways to give them the `user` tier:

- Add `user` to `--setting-sources` while the session still reads the operator's real `~/.claude`. **Rejected** — that is exactly what #1269's exclusion exists to prevent; claude-mpm's global hooks come back.
- Relocate `CLAUDE_CONFIG_DIR` and leave the flag selection to `setting_sources_flag(config_dir)`. **Adopted.** The `user` tier then resolves to the tm-owned config home, so the operator's `~/.claude` stays out — by relocation instead of by exclusion. This is what the daemon path has done since #4451.

## How #1269's isolation is preserved, and the test that proves it

`claude_command_relocated_isolates_by_relocation_not_exclusion` (`core/model_inject.rs`) asserts both halves together, because either alone is the rejected alternative:

- the line carries `--setting-sources user,project,local`, and
- it carries `CLAUDE_CONFIG_DIR='/tm/claude-config'` **before the `claude` program word**, so `env` exports it rather than passing it to `claude` as an argument.

`interactive_config_dir_never_writes_the_home_claude_json` (`tests/relocated_interactive_config_4181.rs`) is the same guarantee in file terms: a decoy `.claude.json` in an isolated fake home comes back byte-identical while the trust seed lands in the managed dir.

`claude_command_includes_setting_sources` is unchanged and still asserts the non-relocated line excludes `user` — that shape is still production, reached when the home cannot be resolved.

## The #2246 Keychain risk, and how the OAuth mirror closes it

On macOS the Keychain entry Claude Code reads is keyed by a hash of `CLAUDE_CONFIG_DIR`. Relocating without compensating gives "login succeeds, then immediately not logged in" — worse than #4181 itself.

`runtime/claude_code.rs`'s `env_bin_prefix` already solves this for the daemon. `build_claude_command` now does the same thing rather than a second mechanism: `oauth_token::resolve_oauth_token()` emitted as `CLAUDE_CODE_OAUTH_TOKEN=…`, single-quoted, placed **after** the `-u` scrub flags — POSIX `env` stops parsing options at the first `NAME=VALUE`, and an assignment ahead of a `-u` makes `env` try to exec `-u` as a command.

When no token resolves the session still launches (the token is a Keychain bypass, not a precondition) and the command prints one line naming `claude setup-token | tm auth set-token`, so the operator is not left diagnosing the loop.

## Smoke test — verbatim

`tm auth status` on this machine: `stored token: absent`, so this exercises the no-token arm — the one that has to fall back to the relocated Keychain/credentials entry.

Ran `./target/debug/tm connect <scratch repo>`; exit 0. stderr:

```
warning: no CLAUDE_CODE_OAUTH_TOKEN resolved; this session reads the keychain entry keyed to the tm-managed config dir. If it starts unauthenticated, run `claude setup-token | tm auth set-token` (see issue #2246).
```

The live pane command (`ps -ax -o pid,ppid,command`, pid 37131):

```
/…/target/debug/tm internal-spawn-disclaimed env -u CLAUDE_CODE_CHILD_SESSION -u CLAUDE_CODE_SESSION_ID -u CLAUDECODE -u CLAUDE_PID -u CLAUDE_EFFORT -u CLAUDE_CODE_EXECPATH CLAUDE_CONFIG_DIR=/Users/masa/.trusty-tools/trusty-mpm/claude-config claude --append-system-prompt-file /var/folders/…/trusty-mpm-system-prompt-85e5991d-….txt --setting-sources user,project,local --dangerously-skip-permissions
```

Every `-u` precedes the assignment; the config dir is the tm-owned home; the flag is the relocated one. `tmux capture-pane -p -t tm-smoke`:

```
 ▐▛███▜▌   Claude Code v2.1.227
▝▜█████▛▘  Opus 5 (1M context) · Claude Max
  ▘▘ ▝▝    /…/scratchpad/smoke
                                                               ● high · /effort
────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────
  TM 1.3.6 | @bob-duetto⚠
  ⏵⏵ bypass permissions on (shift+tab to cycle) · ← 7 agents
```

`Opus 5 (1M context) · Claude Max` is a logged-in account, not a login prompt. The relocated dir carries its own `.credentials.json` (`~/.trusty-tools/trusty-mpm/claude-config/.credentials.json`, 3632 bytes, mode 0600), which is what the session read. Session killed afterwards.

## `build_claude_command`'s four callers

An earlier survey said two. There are four, and the last two would not have compiled:

| Caller | Passes |
|---|---|
| `bin/tm/commands/launch.rs` (`tm launch`) | the resolved config dir |
| `bin/tm/commands/launch.rs` `connect_claude_cmd` (`tm connect`) | the resolved config dir |
| `core/spawn_disclaim/pane.rs` (test) | `None` — pins the non-relocated wrapping |
| `daemon/doctor_transcript_saving.rs` | now `build_claude_command_with(None, None, Some(&config_dir), Some(PROBE_TOKEN))` |

The doctor probe moved to the relocated shape deliberately: it exists to read the REAL spawn's scrub set, and a `None` probe would stop covering what `tm launch` emits. `_with` keeps it hermetic — no ambient token resolution.

`launch.rs` did **not** already resolve the config dir. `grep -n "managed_claude_config_dir\|ensure_managed_config_dir" crates/trusty-mpm/src/bin/tm/commands/launch.rs` returned nothing on `origin/main`; both were added here, routed through a new `managed_config::prepare_interactive_config_dir` so the two commands share one call instead of two copies.

## Trust seeding

`preseed_home_trust(&managed_path)` wrote `~/.claude.json`, which a relocated session never opens. It is now `preseed_managed_trust(<config_dir>, workspace, …)`, keeping `hasTrustDialogAccepted`. The `~/.claude.json` fallback survives for the unresolvable-home case.

#3950's contract carries across unchanged: the four framework builtins enter `enabledMcpjsonServers` only when that run's `PrepReport` says their injector actually wrote. A prep failure passes `(false, false, false, false)` — nothing pinned, nothing approved.

One derivation difference, stated so a reviewer need not find it: `prepare_session_inner` subtracts `project_scope_mcp_names` (#2739's `[mcp.custom]` defense) and `preseed_managed_trust` applies no such subtraction, so a name in BOTH the operator's `tm mcp add` registry and a project's `[mcp.custom]` block is pre-approved here where it was not before. That is the daemon path's existing posture, not a new one, and PR C removes the mechanism entirely. It is documented on `prepare_interactive_config_dir`.

## Per-test disposition

| Test | Disposition |
|---|---|
| `claude_command_bare` / `_with_model` / `_with_prompt` / `_with_both` | Signature update only (`None` third arg). Behavior unchanged. |
| `claude_command_scrubs_inherited_session_markers` | **Not inverted.** Its `!cmd.contains("CLAUDE_CONFIG_DIR")` is asserted on the NON-relocated call, where it is still true and still the #4455 over-scrub guard. The relocated shape gets its own `claude_command_relocated_never_scrubs_what_it_assigns`, which asserts each `DELIBERATE_SPAWN_ENV` name is ASSIGNED and never `-u`-scrubbed. Both shapes are production shapes, so both are pinned. |
| `claude_command_includes_setting_sources` | Kept, scope narrowed in the comment from "every spawned session" to the non-relocated shape. Rewritten to match new behavior — the relocated shape's isolation is proved separately, by relocation rather than exclusion. |
| `setting_source_tiers_parses_the_flag` | **Unchanged.** It tests the constant, which did not change, and its comment ("on a spawn that does NOT relocate…") is still accurate. |
| `tests_behavior_b_tests.rs:1717` (`launch_paths_prepare_through_the_isolated_seam`) | The task brief said it asserts `SETTING_SOURCES_FLAG` appears in the launch line. It does not — that constant is interpolated into a `panic!` MESSAGE. Its real assertions (`!contains("FrameworkPaths::default()")`, `prepare_isolated_session(` count) are untouched and still pass. Only the message text was updated to name both flags. |
| `assert_connect_claude_cmd_carries_persona_flags` | Rewritten to match new behavior: asserts BOTH shapes rather than narrowing to whichever the test machine produces. |
| `every_claude_launch_site_in_the_crate_is_allowlisted` | Allowlist count for `core/model_inject.rs` 4 → 3, with the reason recorded inline. **No launch line was removed** — `format!("env{} claude")` became a `push_str` sequence so the assignments can sit between the scrub flags and the program word, and the literal it now emits (` claude`) matches none of the scanner's patterns. The builder is still registered in `launch_lines()`. |
| `ensure_managed_config_dir_emits_the_frozen_skill_warning` | Stabilized, see below. |

No test was changed to make a failure go away.

## The pre-existing flake this PR had to fix

`ensure_managed_config_dir_emits_the_frozen_skill_warning` captures through a thread-local subscriber, but `tracing` short-circuits every macro on a process-global MAX_LEVEL that starts at OFF and is raised only when some test installs a GLOBAL default. Whether it captured anything depended on which other test in the lib binary ran first; `#[serial]` cannot help, since the level outlives the lock.

It is red on `origin/main` today under a narrow filter, with no changes applied:

```
$ git stash && cargo test -p trusty-mpm --lib core::managed_config
test core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning ... FAILED
test result: FAILED. 16 passed; 1 failed; 1 ignored; 0 measured; 4987 filtered out
```

`origin/main`'s FULL suite passed (`4998 passed; 0 failed`) only because the schedule happened to favour it. Adding any test to that binary flips it, so it had to be made deterministic rather than left to luck: a one-time permissive global default now raises MAX_LEVEL, and `with_default` still overrides it for the capturing thread. The `#4181` new cases live in a separate integration binary (`tests/relocated_interactive_config_4181.rs`) so they cannot perturb that schedule at all.

## Fail-open arms, red then green

Every arm degrades to a warning while the session still starts, so each has a test.

Against `origin/main` all five fail to compile — `build_claude_command` takes two arguments there and `prepare_interactive_config_dir_in` does not exist. That is a real red, but a shallow one, and worth saying plainly rather than dressing up: the signature change is what makes them red, not a behavioral difference the old code would have exhibited. Raw:

```
$ git stash && cargo check -p trusty-mpm --all-targets
error[E0061]: this function takes 2 arguments but 3 arguments were supplied
error[E0433]: failed to resolve: could not find `prepare_interactive_config_dir_in` in `managed_config`
```

Green on this branch:

| Arm | Test | Result |
|---|---|---|
| Relocated dir absent | `interactive_config_dir_creates_an_absent_dir` | provisioned, not refused |
| Managed `.claude.json` malformed | `interactive_config_dir_survives_a_malformed_managed_claude_json` | quarantined (bytes preserved), seeding proceeds |
| `resolve_oauth_token()` returns nothing | `claude_command_omits_the_oauth_token_when_absent` | no assignment, dir still relocated, tier list unchanged |
| Prep failed → nothing pinned | `interactive_config_dir_withholds_builtins_when_a_pin_failed` | no builtin pre-approved |
| Config dir with a space | `claude_command_quotes_a_config_dir_with_a_space` | single-quoted; unquoted would word-split and kill the pane |

```
$ cargo test -p trusty-mpm --test relocated_interactive_config_4181
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Gates — rung 5 (security-sensitive: settings trust + OAuth)

```
$ cargo fmt --check                                        EXIT=0
$ cargo check -p trusty-mpm --all-targets                   EXIT=0
$ cargo clippy -p trusty-mpm --all-targets -- -D warnings   EXIT=0
$ cargo test -p trusty-mpm                                  EXIT=0
$ cargo check --workspace                                   EXIT=0
$ bash scripts/check_line_cap.sh                            EXIT=0
$ bash scripts/check_sld.sh                                 EXIT=0
$ bash scripts/check_capabilities.sh                        EXIT=0
$ bash scripts/check_changelog_fragment.sh                  EXIT=0
```

`cargo test -p trusty-mpm` lib target: `test result: ok. 5002 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 52s`, plus the integration binaries. `--include-ignored` was not run: the 7 ignored tests in this crate are ONNX/static-verification cases unrelated to launch or connect, and there is no ignored launch/connect integration coverage to unlock.

The smoke test above is the direct launch/connect evidence unit tests cannot supply.

## Notes

`shell_single_quote` is reused from `core/spawn_disclaim/pane.rs` (visibility widened to `pub(crate)`) rather than adding a fourth copy to the crate.

Refs #4181, #2246, #1269, #3950, ADR-0042.


bobmatnyc:
## Verdict: WARN

Adversarial review of PR #5398 (`237c9c72f293e92f7c9f2c923bd9460a27ba7418`), rung 5.

### Central question — settled

**This is the accepted design, not the rejected one.** `setting_sources_flag(config_dir)` (`crates/trusty-mpm/src/core/model_inject.rs:132-136`) returns `SETTING_SOURCES_FLAG_RELOCATED` (`user,project,local`) if and only if `config_dir: Some`, and `build_claude_command_with` (`model_inject.rs:316-`) only emits a `CLAUDE_CONFIG_DIR=` assignment — placed before the `claude` program word so `env` exports it rather than passing it as an argument — under that same `Some` branch. `config_dir` traces to exactly one producer, `managed_config::prepare_interactive_config_dir` (`managed_config.rs:68-96`), which returns `Some(dir)` only when `trusty_tools_config::managed_claude_config_dir()` resolves — `<home>/.trusty-tools/trusty-mpm/claude-config`, a tm-owned directory, never the operator's real `~/.claude` — and `None` otherwise (home unresolvable). I traced every caller of `build_claude_command`/`build_claude_command_with` in the crate (`launch.rs:328,640`, `spawn_disclaim/pane.rs:189` test, `daemon/doctor_transcript_saving.rs:116`, `model_inject.rs:452` for delegations) — none passes `user`-tier scope with `config_dir: None`. `claude_command_relocated_isolates_by_relocation_not_exclusion` (`model_inject.rs:606-627`, ran green) pins exactly this pairing. Confirmed empirically, not just by flag-reading: `managed_claude_config_dir()`'s definition (`trusty_tools_config.rs:625-628`) resolves under `crate_config_dir`, not `dirs::home_dir().join(".claude")`.

### The OAuth mirror — token-present arm

The live smoke test only exercised the no-token arm, as the author disclosed. But `claude_command_relocates_the_config_dir` (`model_inject.rs:249-270`, ran green under `cargo test -p trusty-mpm --lib` — 5003 passed, 0 failed) asserts the exact string with a token present (`CLAUDE_CODE_OAUTH_TOKEN='tok-abc'`) AND asserts ordering (`last_unset < first_assignment`) directly off the emitted line, not restated. The construction matches the daemon-path template `env_bin_prefix` (`runtime/claude_code.rs:197-217`) token-for-token: scrub flags first, then `CLAUDE_CONFIG_DIR=`, then `CLAUDE_CODE_OAUTH_TOKEN=`, then the program word. `daemon/doctor_transcript_saving.rs:116-119` also probes the token-present shape (`build_claude_command_with(..., Some(&config_dir), Some(PROBE_TOKEN))`) for over-scrub. The construction is unit-tested with a token present; only the live/manual smoke run was token-absent. Acceptable — the mechanism itself (`oauth_token::resolve_oauth_token`) is unmodified, reused verbatim from the already-shipped daemon path.

### The `[mcp.custom]` name-collision widening — WARN, HIGH

Confirmed as the author describes, and confirmed to be a **real behavior change for `tm launch`/`tm connect`**, not just a restated existing risk:

- `prepare_isolated_session` → `prepare_session_with_repo_url` → `prepare_session_inner` (`session_launch/mod.rs:1103,1153`) computes `trusted_mcp_names` and explicitly **subtracts** `project_scope_mcp_names` (#2739's defense: a project's own `[mcp.custom]` entry must still hit the "new MCP servers found" dialog even after `tm project trust`), then writes that restrictive set via `preseed_workspace_trust_home` into the **operator's real `~/.claude.json`** (`session_launch/settings.rs:823-830`, `dirs::home_dir()`).
- The new `preseed_managed_trust` (`managed_config.rs:132-144` → `standalone/trust_seed.rs:229+`) computes its own `enabledMcpjsonServers` via `mcp_config::managed_mcp_server_names` (`mcp_config.rs:417-440`), which unions the operator's own `mcpServers` registry **unconditionally, with no subtraction**, and writes it into `<config_dir>/.claude.json`.
- Before this PR, `tm launch`/`tm connect` never loaded the `user` tier at all, so `~/.claude.json`'s `enabledMcpjsonServers` was irrelevant to these two commands regardless of which value it held.
- After this PR, the relocated session reads `<config_dir>/.claude.json` instead — the one written by the non-subtracting path. The restrictive write to `~/.claude.json` still happens but is now moot for these commands.
- `inject_custom_trusty_mcps` lets a project's `[mcp.custom]` entry **override** the operator's own registry entry on a name collision (`session_launch/mod.rs` comment, "project overriding user on a name collision"). Combined with the point above: a malicious repo that names a `[mcp.custom]` server identically to one already in the operator's `tm mcp add` registry gets its own command silently pre-approved into `enabledMcpjsonServers` for `tm launch`/`tm connect` — the exact class #2739 exists to stop — where before this PR that class could not reach these two commands at all (no `user`-tier load, no approval mechanism in play).

This is not a fresh invention: `mcp_config::managed_mcp_server_names`/`preseed_managed_trust` already carry this same non-subtracting posture in production for the daemon-managed path, and the author disclosed it inline (`managed_config.rs:46-54`) and in the PR body, asking for an explicit ruling. Ruling: **acceptable to ship WARN, not BLOCK** — narrow precondition (name collision with an operator-registered custom server), inherits an already-shipped posture rather than inventing a new exploit class, and PR C (same sequence, imminent) deletes the entire `enabledMcpjsonServers` mechanism, which resolves this by removal rather than by patching a mechanism about to be deleted. Track it as a `PR C` checklist item, not new work here.

### Pre-existing flake — verified true

Ran `cargo test -p trusty-mpm --lib core::managed_config` against `origin/main` in an isolated worktree, zero PR changes applied:

```
test core::managed_config::tests::ensure_managed_config_dir_emits_the_frozen_skill_warning ... FAILED
test result: FAILED. 16 passed; 1 failed; 1 ignored; 0 measured; 4991 filtered out
```

Confirms the claim — this test is red on `origin/main` today, for the reason stated (thread-local capture vs. process-global `tracing::MAX_LEVEL`). The fix (`managed_config_tests.rs`, a `std::sync::Once`-guarded `set_global_default(tracing_subscriber::registry())`) is confined to `#[cfg(test)]` code — it does not touch production logging. It does change the ambient state for the REST of that lib test binary's run once it fires: `MAX_LEVEL` goes from OFF to unrestricted process-wide, so every subsequent test's `tracing::*!` call bodies actually execute (argument formatting cost) even though the bare `Registry` sink drops them with no output. Worth knowing as a minor forward-looking risk (a test elsewhere that relied on tracing macros being no-ops for a side effect in their arguments would now see that side effect run), but not a finding — the tradeoff is explicit in the doc comment and confined to tests.

### The four corrected claims — all verified

- `tests_behavior_b_tests.rs:1717` (`launch_paths_prepare_through_the_isolated_seam`): confirmed — `SETTING_SOURCES_FLAG`/`SETTING_SOURCES_FLAG_RELOCATED` are interpolated only into the `assert!` panic message, not into an assertion condition; the real assertions (`!contains("FrameworkPaths::default()")`, `prepare_isolated_session(` count) are untouched.
- `setting_source_tiers_parses_the_flag`: confirmed unchanged — no diff hunk touches it; still asserts `vec!["project", "local"]` and the absence of `"user"`.
- `claude_command_scrubs_inherited_session_markers`: confirmed not inverted — its `!cmd.contains("CLAUDE_CONFIG_DIR")` assertion still runs against the non-relocated call (`build_claude_command(None, None, None)`), unchanged in meaning; the relocated shape gets its own, separate `claude_command_relocated_never_scrubs_what_it_assigns`.
- Allowlist count `core/model_inject.rs` 4→3: confirmed mechanical, not a coverage loss. The pre-PR `format!("env{} claude", …)` literal contained the substring `} claude"`, matching the scanner's pattern #5; the post-PR `cmd.push_str(" claude")` does not match any of the scanner's five literal patterns (`Command::new("claude")`, `"claude "`, `"claude {`, `"claude --`, `} claude"`). `launch_lines_covers_every_builder` and `daemon/doctor_transcript_saving.rs:116-119` independently confirm `build_claude_command` is still registered and functionally probed — this is a blind spot in a literal-text heuristic, not a removed launch site.

### Fail-open arms — all covered, all fail red against `origin/main`

`crates/trusty-mpm/tests/relocated_interactive_config_4181.rs` covers: absent config dir (`interactive_config_dir_creates_an_absent_dir`), malformed managed `.claude.json` (`interactive_config_dir_survives_a_malformed_managed_claude_json`, quarantines rather than discards), a failed pin withholding approval (`interactive_config_dir_withholds_builtins_when_a_pin_failed`); `model_inject.rs` covers no-token (`claude_command_omits_the_oauth_token_when_absent`) and a config dir with a space (`claude_command_quotes_a_config_dir_with_a_space`). All five are new-file/new-signature tests that cannot compile against `origin/main` (`build_claude_command` takes 2 args there, `prepare_interactive_config_dir_in` doesn't exist) — a real if shallow red, as the author states plainly rather than dressing it up.

### `--include-ignored` claim

The 7 ignored tests in `cargo test -p trusty-mpm --lib` (verified by running it) are: `stream_json_spawn_fails_gracefully_without_claude` (needs `claude` absent from PATH), `tmux_backend_constructs_without_spawning` (needs tmux absent from PATH), `manual_static_verify_full_roster` (static verification), `probe_stdio_against_trusty_memory` (needs a reachable daemon), `claude_pid_resolves_through_disclaim_wrapper`, `bench_stale_assets_for_many` (benchmark), `spawn_claude_rollback_kills_session_on_send_failure` (needs a real tmux host). None are ONNX tests (the PR body's "ONNX/static-verification" label is imprecise — none of the 7 are ONNX-backed), but the substantive claim holds: none touch launch/connect config-dir relocation, so skipping `--include-ignored` cost no relevant coverage.

### Changelog fragments

Both present and accurate: `crates/trusty-mpm/changelog.d/4181-relocate-claude-config-dir.md` (Changed) and `4181-relocate-claude-config-dir-fixed.md` (Fixed). Only `crates/trusty-mpm/src/**` moved in this PR, both fragments describe the actual behavior change.

### Independent verification run

```
$ cargo fmt --check                    EXIT=0
$ cargo test -p trusty-mpm --lib       test result: ok. 5003 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 56.70s
```

## Findings

| HIGH | crates/trusty-mpm/src/core/mcp_config.rs | 417-440 | `managed_mcp_server_names` (now reachable from `tm launch`/`tm connect` via `preseed_managed_trust`) unions the operator's `tm mcp add` registry unconditionally, with none of `prepare_session_inner`'s `project_scope_mcp_names` subtraction (#2739) — a project `[mcp.custom]` entry that collides with an operator-registered name is silently pre-approved for these two commands, where before this PR it could not reach them at all (no `user`-tier load). | Either thread `project_scope_mcp_names` into `prepare_interactive_config_dir`/`preseed_managed_trust` so it subtracts the same set, or land PR C (deletes the whole `enabledMcpjsonServers` mechanism) promptly. | Parent |
| MEDIUM | crates/trusty-analyze/ui/dist/assets/index-*.js, crates/trusty-analyze/ui/dist/index.html, crates/trusty-memory/ui/package-lock.json | n/a | Unrelated build-artifact diffs (UI bundle rebuild, package-lock churn removing `libc` fields) bundled into a security-sensitive PR with no connection to the stated change. | Drop these files from the branch (`git checkout origin/main -- <paths>` then re-commit, or rebase them out) before merge. | Fix here |
| MEDIUM | crates/trusty-mpm/src/core/model_inject.rs | 316-345 (build_claude_command_with) | `CLAUDE_CODE_OAUTH_TOKEN` is embedded as a literal `NAME=VALUE` in a command string that `disclaim_pane_command` sends via tmux `send-keys` into an interactive pane shell — it will land in that shell's history file and in `tmux capture-pane` output for anyone with access to the pane. This exact pattern is already shipped for the daemon path (`runtime/claude_code.rs`'s `env_bin_prefix`, also piped to `tmux send-keys`), so it is not a new exploit class, but this PR doubles the number of entry points carrying it (now also the interactive CLI paths, likely used with tmux history enabled more often). | Worth an explicit sign-off note (not new engineering) confirming the tradeoff already accepted for #2246's daemon path is knowingly extended here. | Parent |

## Notes

- Central isolation question (relocation vs. exclusion), the OAuth-mirror construction, the four corrected test claims, the fail-open arm coverage, and the flaky-test fix are all verified correct by direct code trace and/or an independent test run — no findings there.
- No CRITICAL findings. Verdict is WARN on the strength of the one HIGH (MCP name-collision widening) — narrow, inherited from an already-shipped posture, and slated for removal by PR C, but real and worth tracking rather than silently accepting.

Issue #1334, opened 2026-06-16
Agentic detection undercounts commits with Co-Authored-By: Claude trailers


bob-duetto:
## Summary
`tga`'s agentic-commit detection under-counts AI-assisted commits. Commits that carry a `Co-Authored-By: Claude <...>` trailer (and similar AI co-author trailers) are frequently NOT flagged as AI-assisted: the emitted `agentic_mode` stays `'none'` and `is_ai_assisted` stays false/unset for these commits.

## Observed impact (downstream consumer: Duetto cto-reports)
In a real dataset, one engineer had `is_ai_assisted = TRUE` on only ~108 of ~461 commits across 3 weeks (W21–W23) despite the majority carrying `Co-Authored-By: Claude` trailers. This causes a large "agentic gap" overcount in downstream reporting (apparent ~77% non-agentic when actual agentic usage is much higher). `tga` only began writing `agentic_mode='full_agentic'` around 2026-06-11, so older history is especially affected.

## Root-cause hypothesis
The commit-message parser that determines `agentic_mode` / `is_ai_assisted` does not treat `Co-Authored-By:` trailers naming an AI assistant (e.g. `Claude`, `claude`, `noreply@anthropic.com`, Cursor, Copilot) as an AI-assistance signal. `agentic_mode` appears to default to the literal string `'none'` rather than being derived from trailer inspection.

## Suggested fix
When parsing commit trailers, detect AI co-author signatures (`Co-Authored-By:` with an AI assistant identity — at minimum Anthropic/Claude `noreply@anthropic.com`, plus common Cursor/Copilot markers) and set `is_ai_assisted = true` (and an appropriate `agentic_mode`) for those commits. Consider a backfill path so historical commits are re-classified.

## Notes
- Detected via `tga` v2.8.0.
- Tracked downstream as cto-reports issue #140.

PR #5403, opened 2026-08-11
fix(tga): widen the agentic marker set (#5249)

### LINKED ISSUE #5249: fix(tga): agentic detection undercounts by 41.6 points — hardcoded, non-configurable pattern set
tga's agentic-commit detection uses a single hardcoded, non-configurable pattern set and materially undercounts. Measured over trusty-tools' 2352 commits: `Generated with trusty-mpm` footer matches 983, `Co-Authored-By: Claude` matches 1162, overlap 5, union 2140 = 90.99% true agentic share. The detector (`ai_patterns()`, `crates/trusty-git-analytics/src/collect/ai_attribution.rs:126-268`, called from `extractor.rs:557,579`) catches only 1162 = 49.4% — a 41.6-point undercount. It misses `Generated with trusty-mpm` because the regex requires the literal "Claude Code"; it also has no entries for Devin, Aider, or OpenHands, and never checks author/committer email (`extractor.rs:547` extracts `author_email` but never passes it to detection at line 555-557). No config surface exists anywhere (`grep -rl "ai_tool\|agentic" src/core/config/` is empty).

[#1334](https://github.com/bobmatnyc/trusty-tools/issues/1334) fixed literal `Co-Authored-By: Claude` detection (closed, verified — this repo's 1162 catches confirm it holds). This issue is the next layer: the pattern set that fix landed is still hardcoded and still incomplete.

**Owner requirement, verbatim:** "The footer detector should be flexible, allow multiple types." Must support multiple marker families (trailers, body footers, author/committer emails) and be configurable so a target org's house footer can be added without a code change.

**Validation target:** `All-Hands-AI/OpenHands`, 7990 commits — 118 Claude, 6 Copilot, 33 Cursor, 92 Devin, 3 Aider, and 3873 `Co-authored-by: openhands <openhands@all-hands.dev>` (48%, matched by nothing in tga today). Chosen over `block/goose` and `cline/cline` because its dominant marker is genuinely agentic and unmatched, so it tests generalization rather than pattern luck. tga cannot clone ([#5215](https://github.com/bobmatnyc/trusty-tools/issues/5215)) — validation needs a manual clone.

**Acceptance:**
- Detection reads a configurable marker list (trailer patterns, body footers, author/committer emails) instead of the hardcoded `OnceLock<AiPatterns>` in `ai_attribution.rs:126-151`.
- A new marker (e.g. a house footer) can be added without a code change.
- Against trusty-tools' own history, detected agentic share rises from 49.4% toward the measured 90.99% union.
- Against a manual OpenHands clone, the `openhands@all-hands.dev` co-author marker is caught by a configured pattern, verified with a real-commit-shape test (not only synthetic fixtures).

DOC-67 §8 (`docs/specs/DOC-67-tga-audit-mode.md`) promises acquirers an `agentic_pct` figure in the Engineering Velocity section — see [#5241](https://github.com/bobmatnyc/trusty-tools/issues/5241) and [#5242](https://github.com/bobmatnyc/trusty-tools/issues/5242). A wrong provenance number in a due-diligence report is worse than an absent one; this should land before AUDIT ships that section.

bobmatnyc:
Agentic detection was seven hardcoded regexes, every one keyed on the literal "Claude". A commit carrying only this repo's house footer scored `agentic_mode = 'none'` and counted as human work. `agentic_pct` feeds the figure DOC-67 §8 hands an acquirer, so the number was confidently wrong rather than absent.

Detection is now an ordered marker list: a tool label, a mode, a scope naming the haystack (`Co-Authored-By:` trailer value / whole message / author-or-committer email), and a regex. The set adds the trusty-mpm footer, Devin, OpenHands and Aider; the walk passes the author and committer email it used to extract at `extractor.rs:547` and discard. One scan yields both `ai_tool` and `agentic_mode`, so the two columns cannot disagree.

The marker set stays a fixed in-crate list. The operator-configurable `ai_detection:` block from the first revision of this PR is gone at the owner's direction, along with the `Config` field that made it a major break — tga stays on 2.x.

## Catch rate, before and after

`tests/agentic_detection_corpus.rs` scores this repo's full history twice: once through a frozen transcription of the pre-#5249 patterns, once through the shipped `detect`.

```
cargo test -p tga --test agentic_detection_corpus -- --include-ignored --nocapture
agentic detection: 12 marker(s) active for [claude, trusty-mpm, devin, openhands, aider, copilot, cursor]; …
commits=2434 known_marker_baseline=2216 (91.04%) before=1161 (47.70%) after=2216 (91.04%) beyond_baseline=0
detections by tool: claude=1161 trusty-mpm=1055
```

Unchanged by the descope, as expected — nothing about the shipped markers moved.

**What this corpus does not prove.** trusty-tools' history contains exactly two agentic markers. The `detections by tool` line names two of the seven tools in the line above it: Devin, OpenHands, Aider, Copilot and Cursor have zero occurrences here, so no assertion below exercises them, and they are proven only by the synthetic unit tests in `collect::ai_markers::tests`. The first revision called its comparison function a "ground truth" and asserted `after >= 85%` against it; since both sides were blind to the same ten markers, those assertions could not fail regardless of how those markers behaved. It is renamed to a known-marker baseline, the run prints the active tool list beside the observed counts so the gap is visible rather than implied, and the assertions are unchanged in strength.

## Fail-open finding: there is no failure branch

With no operator-supplied patterns, nothing outside `ai_markers.rs` can hand detection a regex, so there is no malformed-input branch to fall open on. The only compile site is `BUILTIN`, whose failure is a programmer error caught by `tests::every_builtin_pattern_compiles` before any collection runs. The three error-arm tests from the first revision (`load_rejects_bad_ai_marker`, `validate_rejects_bad_pattern`, `from_config_rejects_bad_pattern`) lost their subject entirely and are gone, not weakened — the code they guarded no longer exists.

What replaces the guard is a disclosure, because a fixed list has its own silent-undercount mode: a target with an unrecognised house footer reports a low share and looks like a repository with no AI in it. `detection_disclosure()` states the active tools and the caveat; `tga collect` and `tga backfill ai-detection-commits` log it once per run, and DOC-67 §8 now records both that the section must render it and that the list is fixed at release time.

## Marker anchoring (critic HIGH / MEDIUM)

- `\bdevin\b` classified `Co-authored-by: Devin Booker <devin@personal.example>` as FullAgentic, contradicting the anchoring rule the marker table itself states. The trailer marker now keys on `devin-ai-integration` or `@devin.ai`; regression test `human_named_devin_is_not_agentic` also asserts both real bot forms still classify.
- The email-scope Copilot marker matched `copilot@` at any domain, so a human team mailbox classified IdeAssisted. It is **removed** rather than re-anchored: I could not confirm GitHub's real Copilot committer address against an actual Copilot-authored commit, and guessing an anchor is how the marker got wrong in the first place. The trailer-scope Copilot marker is untouched, so Copilot detection via `Co-Authored-By:` is unchanged.
- `\bclaude\b` on a trailer value has the same theoretical shape (a co-author named Claude Dupont). It predates this issue (#1334) and narrowing it would drop real detections, so it is left alone and the marker table now says so explicitly instead of claiming every pattern is anchored.

## Version: 2.15.0

`cargo-semver-checks` reports `2.14.0 -> 2.15.0 (minor change) … 196 pass, 58 skip … no semver update required` — no break, so a patch would clear the gate. MINOR is still the correct position because the diff **adds** public API: the `collect::ai_markers` module with `detect`, `detection_disclosure`, `CommitSignals` and `Detection`. The binary crate calls those, which is what forces them public; `MarkerScope`, `AiMarker` and the marker list stay private.

## The collector.rs split is dropped

It happened because the config plumbing pushed `collector.rs` past its frozen 766-SLOC budget. The descoped diff does not touch that file at all, so the cap passes untouched — `line-cap: measured 3895 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK`. This repo forbids a standalone SLOC-cap fix, and with nothing tripping the cap the split would have been exactly that, riding as a passenger. The allowlist is back to its original contents.

## Rung 5

```
cargo fmt --check                                              EXIT=0
cargo check -p tga --all-targets                               EXIT=0
cargo clippy -p tga --all-targets -- -D warnings               EXIT=0
cargo test -p tga                                              1070 passed; 0 failed; 1 ignored
cargo test -p tga --test agentic_detection_corpus -- --include-ignored   1 passed
SKIP_UI_BUILD=1 cargo check --workspace                        EXIT=0
SKIP_UI_BUILD=1 cargo test -p trusty-review --features profile 1621 passed; 0 failed; 5 ignored
bash scripts/check_semver.sh --crate tga                       EXIT=0
```

Check scripts, all run: `check_sld`, `check_line_cap`, `check_changelog_fragment`, `check_test_pointers` (23631 citations, 0 dangling), `check_doc_numbers`, `check_semver`, `check_adr`, `check_agent_assets`, `check_buildrs_sync`, `check_capabilities`, `check_generated_regions`, `check_generation_artifacts`, `check_public_docs` — all PASS. The four `*_selftest.sh` are N/A: they test the gates, not the diff.

Two `cargo test -p tga --doc -- --include-ignored` failures are pre-existing and untouched: the ```ignore doctests at `core/config/mod.rs:13` and `core/config/validator.rs:19` do not compile when forced to run. Neither file is in this branch's diff.

Closes #5249


bobmatnyc:
**code-critic re-review — APPROVE** (head `564ef03b`)

All three prior WARN findings closed.

**Corpus metric** — `is_agentic_by_ground_truth` renamed `matches_known_marker` and documented as bounding what the corpus can explain. The run now prints `detection_disclosure()` and per-tool counts (`claude=1161 trusty-mpm=1055`), making visible that 2 of 7 tool labels carry the measurement. `beyond_baseline` and `pct(after)` are independent, failable checks for the markers that do occur; the other five are covered by direct unit tests in `ai_markers.rs`, not by this corpus. Disclosed, not overstated.

**Devin marker** — now `\bdevin-ai-integration\b|@devin\.ai\b`, with `human_named_devin_is_not_agentic` asserting `Devin Booker <devin@personal.example>` → `None`.

**Copilot email marker** — removed rather than re-anchored, since GitHub's real Copilot committer address could not be confirmed against an actual Copilot-authored commit.

**`\bclaude\b` residual** — pre-existing since #1334, unchanged here. Exposure in this repo is 0 of 1161: every real `Co-Authored-By:` match is `@anthropic.com` or the claude-mpm URL. Disclosed in the marker table and at runtime.

**Descope integrity** — `core/config/mod.rs` byte-identical to the branch point, no `ai_detection` remnant, root `Cargo.toml` clean. `check_semver.sh --crate tga` EXIT=0; new public surface (`CommitSignals`, `Detection`, `detect`, `detection_disclosure`) is purely additive, `MarkerScope`/`AiMarker`/`BUILTIN` stay private. 2.15.0 holds.

Verified independently: `cargo test -p tga` 1070 passed / 0 failed, corpus run reproduced at `before=1161 (47.70%) after=2216 (91.04%)`.

Issue #4206, opened 2026-07-28
Test suite writes trust-seed entries into the operator's real managed .claude.json — 2,443 leaked entries, one config destroyed


bobmatnyc:
DEFECT: `managed_claude_config_dir()` (`crates/trusty-mpm/src/core/trusty_tools_config.rs:541` → `crates/trusty-common/src/crate_config.rs:103-105`) resolves from `dirs::home_dir()` ONLY. It never reads `CLAUDE_CONFIG_DIR` and accepts no injectable base. So `preseed_managed_trust` (`core/standalone/trust_seed.rs:57`), called from `prepare_managed_config` (`runtime/claude_code.rs:593`), writes into the REAL production config dir even when driven from a hermetic test. `DaemonState::with_root_isolated_managed(tempdir)` isolates daemon state but does not reach this resolver.

EVIDENCE (measured 2026-07-27):
- `$CLAUDE_CONFIG_DIR/.claude.json` is 985,739 bytes with **2,538 `projects` entries**; `~/.claude.json` has 33.
- **2,443 (96.3%) are ephemeral `tempfile::TempDir` paths** under `/var/folders/…`, carrying test-fixture suffixes that appear verbatim in the suite: `zombie-ws` (324, `bin/tm/commands/managed_tests.rs:460`), `incomplete-ws` (279, `tests/session_manager_mvp.rs:976`), `stale-heal-ws` (247, `:1080`), `selfheal-ws` (244, `:893`), `frontgate-ws` (228, `:1158`).
- **2,517 of 2,538 (99.2%) carry ONLY the four fields the seeder writes** and none of Claude Code's runtime fields — so ~99% was written by tm for workspaces no session ever ran in.
- Traced path: `tests/session_manager_mvp.rs:893` → `resume_managed` → `spawn_resume` → `prepare_managed_config` → `preseed_managed_trust(managed_claude_config_dir(), tempdir)`.
- **85.6% (2,172) point at directories that no longer exist** — the write is purely additive; there is no removal anywhere, including on decommission.
- Growth is continuous since the seeder landed 2026-06-22 (`66ee38a5`, #1555): 901 entries Jul 10 → 2,720 Jul 15 (+1,819, 0 removed).

[...]

PR #5406, opened 2026-08-11
fix(trusty-mpm): seed builtin MCP servers into user scope (#4181, PR C1 of 3)

bobmatnyc:
PR C1 of the ADR-0042 sequence for #4181. It adds the user-scope declaration. It deletes no injector.

## What seeding does

`mcp_config::seed_builtin_servers(config_dir)` writes the four names in `BUILTIN_MANAGED_MCP_SERVERS` into the top-level `mcpServers` map of `<config_dir>/.claude.json`, using the existing `builtin_server_entry`, under `claude_json_guard::lock()` across the whole read-mutate-write cycle. `ensure_global_config_dir` calls it at the point that used to call `ensure_mcp_config`, which reaches `ensure_managed_config_dir` (every daemon spawn, resume, and in-place relaunch) and the standalone `tm run` driver.

**Insert-if-absent is the contract, not an optimisation.** An entry already sitting under a builtin name is the operator's — `tm mcp add` is the supported edit path under ADR-0042 decision item 3 — so seeding never overwrites, reorders, or removes it. It reports the names it inserted; an empty result means no write happened at all, so a fully-seeded config is left byte-identical (`test_seeding_no_spurious_write_when_unchanged`, `test_seeding_trailing_newline_does_not_trigger_rewrite`).

## `ensure_mcp_config` is gone, and why that is not an injector deletion

It wrote the same four servers into `<CLAUDE_CONFIG_DIR>/.mcp.json`. ADR-0042 measured that Claude Code discovers a `.mcp.json` only by walking up from the session's cwd, and cwd is always the repo — `standalone::run.rs:65` calls `cmd.current_dir(repo_path)`, and the daemon spawns in the workspace. That file was on no session's search path, so nothing read what it wrote. Its six tests were retargeted onto the new home rather than dropped: server set, review entry shape, `$HOME` isolation, no-spurious-write, trailing-newline tolerance.

It is not one of the injectors this PR is forbidden to touch. `native_mcp.rs`, `custom_mcp.rs`, every `inject_*`, `exclude_mcp_json_from_git`, `launch_trusted_mcp_names`, `preseed_workspace_trust`, and every `enabledMcpjsonServers` write are all still here and still called. The #3926 coupling does not reach `ensure_mcp_config` either: an `enabledMcpjsonServers` approval resolves against the `.mcp.json` discovered from cwd, never against the config dir's copy.

## The double-provisioning interval

Until C2, both mechanisms provision the builtins — seeding into user scope, the injectors into the workspace `.mcp.json`. They are different files, so no duplicate and no clobber is possible. `seeding_and_workspace_injection_coexist_without_clobbering` runs seed → `inject_trusty_mpm_mcp` → seed and asserts the second seed inserts nothing, the user-scope entry is the unduplicated canonical builtin, and the workspace file still declares exactly what was injected.

One collision was possible and is closed by construction: `inject_native_trusty_mcps_from` iterates a fixed allowlist (`slack-mcp`, `telegram-mcp`, `gworkspace-mcp`, `trusty-analyze`) that contains no builtin, and `custom_mcp`'s registry loop skips `is_reserved_name`. So seeding the builtins into the registry cannot make either loop overwrite `inject_trusty_search_mcp`'s `--index` pin. #1373 is unaffected.

## The regression seeding caused, found by an existing test

Seeding writes the builtins into the same `mcpServers` map `managed_mcp_server_names` unions into `enabledMcpjsonServers`. That union was wholesale, so a seeded builtin entered the approved set with no `_pinned` evidence — approving a name whose workspace content was never verified this run. That is #3950, reopened by the seeding rather than by any change to the trust derivation.

Red, on this branch before the fix:

```
test runtime::claude_code::tests::prepare_managed_config_excludes_builtins_when_mcp_json_write_fails ... FAILED

thread '...' panicked at crates/trusty-mpm/src/runtime/claude_code_tests.rs:2159:9:
trusty-mpm must not be approved when its pin write failed this run (headless resume
path, no human to decline a consent dialog): ["trusty-memory", "trusty-mpm", "trusty-review", "trusty-search"]

test result: FAILED. 5018 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 67.87s
```

The fix subtracts `BUILTIN_MANAGED_MCP_SERVERS` from the registry half of the union, so a builtin's membership comes from its `_pinned` flag and nothing else. `a_seeded_builtin_is_not_approved_when_its_pin_failed` pins it: a freshly seeded config dir with all four flags false approves nothing, a non-builtin registration still enters, and a successful pin still approves its builtin.

**This reverses a documented assertion.** `managed_mcp_server_names_registry_collision_is_a_separate_trusted_source` said a registry entry named `trusty-memory` is trusted regardless of the conditional toggle, on the grounds that the registry is operator-controlled. Narrowly, that rule cannot survive seeding — it would make all four `_pinned` parameters dead. Separately it was already wrong: the approval is content-blind and resolves against the workspace `.mcp.json`, and `custom_mcp` skips reserved names, so a registry entry under a builtin name is never what pins the workspace entry — only the toggle-gated injector is. The test is renamed and inverted with that reasoning in it.

## Fail-open analysis

Seeding runs on every launch now, so each failure arm was traced to what the operator actually experiences.

| Arm | Seeding | Session |
|---|---|---|
| Config dir absent | Creates the dir and the file with all four builtins | starts |
| `.claude.json` unreadable | `Err`, absorbed by the wrapper into a `warn!` | starts |
| `.claude.json` malformed | `warn!`, returns having written and renamed nothing | starts |
| `mcpServers` present but not an object | `warn!`, returns having written nothing | starts |

Two design decisions carry that table.

**Seeding does not go through `read_config`.** That reader quarantines by renaming, and `.claude.json` holds OAuth state. It is right for `tm mcp add`, where the operator is present and is explicitly asking to write the file; it is wrong for something that runs unattended before every launch. `read_config_for_seeding` declines instead — no rename, no write, file byte-identical.

**The call site does not use `?`.** `seed_builtin_mcp_servers` swallows the outcome into a log line, because it sits between `seed_credentials` and `deploy_agents_and_skills` in `ensure_global_config_dir`. A `?` there would let an unwritable `.claude.json` skip the agent and skill deploy and hand the session an empty roster. A seeding failure costs MCP servers for one launch and nothing else — which is what the `output-styles` assertion in both fail-open tests actually checks.

### What a malformed-config user experiences the first time they launch after this ships

No change from today. `preseed_managed_trust` already quarantines that file on every launch — it renames it to a timestamped `.corrupt` sibling, logs at `ERROR`, and writes a fresh config, and it has done so since #4206. It runs a few milliseconds after seeding in `prepare_managed_config`. So the OAuth state is lost either way; declining in the seeder does not save it. What declining does buy is that this PR adds no second writer to that failure, and the next launch — reading the valid file trust-seeding just wrote — seeds normally.

### Red-then-green per arm

Two of the four new `ensure_global_config_dir` tests fail against `origin/main`, run there verbatim against main's production code:

```
test core::standalone::global_config::tests::red_demo_seeds_builtin_mcp_servers ... FAILED
test core::standalone::global_config::tests::red_demo_never_overwrites_an_operator_mcp_entry ... FAILED
test core::standalone::global_config::tests::red_demo_survives_a_malformed_claude_json ... ok
test core::standalone::global_config::tests::red_demo_survives_an_unreadable_claude_json ... ok

thread '...red_demo_seeds_builtin_mcp_servers' panicked at global_config.rs:383:70
thread '...red_demo_never_overwrites_an_operator_mcp_entry' panicked at global_config.rs:413:9:
assertion failed: val["mcpServers"]["trusty-mpm"].is_object()

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 5014 filtered out; finished in 0.01s
```

The two fail-open tests pass on main, and would on any additive change: main's `ensure_global_config_dir` never opens `.claude.json`, so there is nothing there to break. Reporting them as red-on-main would be false.

They are not vacuous, though. Against the naive design — seeding through the quarantining `read_config`, one line different — both go red:

```
test core::mcp_config::tests::seed_builtin_servers_declines_a_malformed_config_without_quarantining ... FAILED
test core::standalone::global_config::tests::test_global_config_dir_survives_a_malformed_claude_json ... FAILED

thread '...declines_a_malformed_config_without_quarantining' panicked at mcp_config_tests.rs:635:5:
a malformed config must seed nothing

test result: FAILED. 30 passed; 2 failed; 0 ignored; 0 measured; 4995 filtered out; finished in 0.03s
```

Green on this branch, all four:

```
test core::standalone::global_config::tests::test_global_config_dir_seeds_builtin_mcp_servers ... ok
test core::standalone::global_config::tests::test_global_config_dir_never_overwrites_an_operator_mcp_entry ... ok
test core::standalone::global_config::tests::test_global_config_dir_survives_a_malformed_claude_json ... ok
test core::standalone::global_config::tests::test_global_config_dir_survives_an_unreadable_claude_json ... ok
```

## Gates — rung 5 (persisted state, process lifecycle, security)

```
cargo test -p trusty-mpm                               EXIT=0
cargo check --workspace                                EXIT=0
bash scripts/check_line_cap.sh                         EXIT=0
bash scripts/check_sld.sh                              EXIT=0
bash scripts/check_capabilities.sh                     EXIT=0
bash scripts/check_changelog_fragment.sh               EXIT=0
```

```
test result: ok. 5020 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out; finished in 51.10s
line-cap: measured 3898 tracked .rs file(s) (floor 500); 4 allowlisted, 0 violations — OK.
changelog-fragment gate: scanned 5 changed path(s); all 1 crate(s) with source changes are recorded.
```

Two gate failures were fixed rather than worked around, both worth naming because they are non-obvious:

`check_line_cap.sh` failed at 534 SLOC for `global_config.rs`. The cause was not size — it was the `br#"{"oauthAccount": ..."#` literal in the new test, whose unbalanced brace skews the script's matcher and makes it count the whole inline `#[cfg(test)]` module against the 500-SLOC production cap, exactly the fail-closed behavior CLAUDE.md documents. Building the malformed JSON instead of writing it as a literal restores the exclusion. Nothing was ignored, excluded, or allowlisted.

`cargo clippy` then failed on that replacement (`as_bytes` after slicing a string). Fixed in place.

`cargo check --workspace` rebuilt the embedded UIs and dirtied `crates/trusty-analyze/ui/dist/` and `crates/trusty-memory/ui/package-lock.json`. Both restored; `git diff --name-only origin/main...HEAD` lists five paths, all under `crates/trusty-mpm/`.

Refs #4181, [ADR-0042](docs/adr/0042-mcp-configuration-is-static-and-persistent.md).

### DIFF
--- a/crates/trusty-mpm/changelog.d/4181-seed-builtin-mcp-servers.md
+++ b/crates/trusty-mpm/changelog.d/4181-seed-builtin-mcp-servers.md
@@ -1,0 +1,17 @@
+Changed
+
+- The four framework MCP servers (`trusty-memory`, `trusty-mpm`,
+  `trusty-review`, `trusty-search`) are now declared once in the user-scope
+  `mcpServers` map of the tm-owned `.claude.json`, seeded on every launch by
+  `mcp_config::seed_builtin_servers`. Seeding is insert-if-absent: an entry the
+  operator registered under one of those names via `tm mcp add` is never
+  overwritten, reordered, or removed. This replaces the write to
+  `<CLAUDE_CONFIG_DIR>/.mcp.json`, which no session ever read — Claude Code
+  discovers a `.mcp.json` by walking up from the session's cwd, and cwd is
+  always the repo (#4181, ADR-0042).
+- Seeding never quarantines. A malformed `.claude.json`, or one whose
+  `mcpServers` is not an object, makes it warn and return without renaming or
+  writing anything; a failure of any kind is absorbed so the agent and skill
+  deploy that follows still runs. The per-launch injectors and the
+  `enabledMcpjsonServers` approval are untouched here and are deleted together
+  in the next PR of the ADR-0042 sequence.
--- a/crates/trusty-mpm/src/core/managed_config_tests.rs
+++ b/crates/trusty-mpm/src/core/managed_config_tests.rs
@@ -89,7 +89,9 @@ fn ensure_managed_config_dir_deploys_full_roster() {
 
     // Scaffolding landed.
     assert!(config_dir.join("settings.json").exists());
-    assert!(config_dir.join(".mcp.json").exists());
+    // #4181: the MCP declaration is seeded into `.claude.json`'s user-scope
+    // `mcpServers` map now, not into a `.mcp.json` no session discovers.
+    assert!(config_dir.join(".claude.json").exists());
 
     // Every seeded specialist must be present AND spawnable (has a
     // `description` in its deployed frontmatter).
--- a/crates/trusty-mpm/src/core/mcp_config.rs
+++ b/crates/trusty-mpm/src/core/mcp_config.rs
@@ -296,6 +296,102 @@ pub fn add_server(config_dir: &Path, name: &str, entry: Value) -> Result<bool> {
     Ok(true)
 }
 
+/// Seed the framework builtin MCP servers into the user-scope `mcpServers`
+/// map of `<config_dir>/.claude.json`, **inserting only the names that are
+/// absent**.
+///
+/// Why (#4181, [ADR-0042]): a server declared in this map connects with no
+/// approval prompt, while the same server declared in a workspace `.mcp.json`
+/// comes up `⏸ Pending approval`. ADR-0042 moves the framework builtins' one
+/// declaration here so it is written once and persists, instead of being
+/// re-injected into every ephemeral worktree. That declaration must be seeded
+/// on a path that runs before every spawn, which is
+/// [`crate::core::standalone::global_config::ensure_global_config_dir`].
+///
+/// What: reads `<config_dir>/.claude.json`, and for each name in
+/// [`BUILTIN_MANAGED_MCP_SERVERS`] that the top-level `mcpServers` map does
+/// NOT already contain, inserts [`builtin_server_entry`]'s canonical entry.
+/// Returns the names actually inserted; an empty vector means nothing was
+/// written. Holds [`crate::core::claude_json_guard::lock`] across the whole
+/// read-mutate-write cycle, because the daemon runs other `.claude.json`
+/// seeders concurrently (#4072).
+///
+/// **Insert-if-absent is the contract, not an optimisation.** An entry already
+/// under a builtin name is the operator's — `tm mcp add` is the supported edit
+/// path (ADR-0042 decision item 3) and must win. This function never
+/// overwrites, never reorders, and never removes.
+///
+/// **It also never quarantines.** Unlike [`read_config`], a `.claude.json`
+/// that is malformed, or whose `mcpServers` is not an object, makes this
+/// function warn and return `Ok(vec![])` with the file untouched — no rename,
+/// no write. Seeding runs on every launch now, so a quarantine here would turn
+/// a convenience into a per-launch data-loss event against a file that also
+/// holds OAuth state. `standalone::trust_seed::preseed_managed_trust` still
+/// owns that decision (and still logs it at `ERROR` with a timestamped
+/// quarantine path, #4206); once it has replaced the unparseable file, the
+/// next launch seeds normally.
+///
+/// [ADR-0042]: ../../../../docs/adr/0042-mcp-configuration-is-static-and-persistent.md
+///
+/// Test: `seed_builtin_servers_inserts_every_builtin_when_absent`,
+/// `seed_builtin_servers_never_overwrites_an_existing_entry`,
+/// `seed_builtin_servers_is_idempotent`,
+/// `seed_builtin_servers_preserves_unrelated_keys`,
+/// `seed_builtin_servers_declines_a_malformed_config_without_quarantining`,
+/// `seed_builtin_servers_declines_a_non_object_mcp_servers_map`,
+/// `seed_builtin_servers_creates_the_config_when_the_dir_is_absent`,
+/// `seed_builtin_servers_errors_when_claude_json_is_unreadable`,
+/// `seeding_and_workspace_injection_coexist_without_clobbering`.
+pub fn seed_builtin_servers(config_dir: &Path) -> Result<Vec<String>> {
+    // #4181: the guard spans read → mutate → write, not just the write.
+    let _guard = crate::core::claude_json_guard::lock();
+
+    let path = config_dir.join(CLAUDE_JSON);
+    let Some(mut config) = read_config_for_seeding(&path)? else {
+        return Ok(Vec::new());
+    };
+
+    let root = config
+        .as_object_mut()
+        .expect("read_config_for_seeding only yields objects");
+
+    // #4181: a non-object `mcpServers` is operator state this function is not
+    // entitled to destroy — `mcp_servers_mut` would replace it with `{}`.
+    if let Some(existing) = root.get("mcpServers")
+        && !existing.is_object()
+    {
+        tracing::warn!(
+            "skipping builtin MCP seed: {}'s mcpServers is not an object — leaving it untouched",
+            path.display()
+        );
+        return Ok(Vec::new());
+    }
+
+    let servers = root
+        .entry("mcpServers")
+        .or_insert_with(|| Value::Object(Map::new()))
+        .as_object_mut()
+        .expect("mcpServers is an object — checked immediately above");
+
+    let mut seeded: Vec<String> = Vec::new();
+    for name in BUILTIN_MANAGED_MCP_SERVERS {
+        if servers.contains_key(*name) {
+            continue;
+        }
+        let Some(entry) = builtin_server_entry(name) else {
+            continue;
+        };
+        servers.insert((*name).to_string(), entry);
+        seeded.push((*name).to_string());
+    }
+
+    if seeded.is_empty() {
+        return Ok(Vec::new());
+    }
+    write(&path, &config)?;
+    Ok(seeded)
+}
+
 /// Remove a user-scope MCP server.
 ///
 /// Why: `tm mcp remove` drops a server the operator no longer wants injected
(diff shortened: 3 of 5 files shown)


bobmatnyc:
## Verdict: APPROVE

No issues found at >80% confidence. APPROVED for next pipeline stage (rung 5, security-sensitive).

### Priority 1 — the inverted assertion (`managed_mcp_server_names`, `crates/trusty-mpm/src/core/mcp_config.rs:540-556`)

The reversal is a genuine correction, not a red-to-green cheat. `managed_mcp_server_names`'s output feeds only `enabledMcpjsonServers` (verified via call sites in `trust_seed.rs:314`, `session_launch/mod.rs:1147`, `session_launch/settings.rs:674`) — which auto-approves a **workspace** `.mcp.json` entry by name, content-blind. The operator's own top-level `.claude.json` `mcpServers` entry needs no such approval and runs regardless (per the module's own empirically-confirmed doc: `Scope: User config`, never `⏸ Pending approval`). So excluding `BUILTIN_MANAGED_MCP_SERVERS` from the registry-union does not stop an operator's collision-named entry from running — it only stops that registry membership from also silently pre-approving a same-named, unpinned entry in a cloned repo's `.mcp.json`. That is exactly #3950's invariant ("a name may be in `enabledMcpjsonServers` only if something pinned the content behind that name in the same run"), and the old test asserted a violation of it that pre-dated seeding (an operator's registry key already let a same-named repo-controlled `.mcp.json` entry ride to auto-approval with zero content verification — seeding just made it unconditional for all four builtins on every launch). The new test's second half (`slack-mcp` still trusted) confirms the union's actual purpose — non-builtin registry names — is untouched.

### Priority 2 — never-overwrite contract

Verified in `mcp_config.rs:296-397`. `contains_key(*name)` skips a name regardless of what the existing entry's args/command are — never inspects content, so different-args and different-command cases are both covered by the same check. `mcpServers` present-but-non-object: explicit `is_object()` check before any mutation, warns and returns without touching `root` (`mcp_config.rs:322-329`). `mcpServers` absent: `.entry(...).or_insert_with(Object::new)` creates it. `.claude.json` absent: `read_config_for_seeding` returns `Some({})` on `NotFound`, and `write_json_atomic` (`trusty-common/src/claude_config.rs:163-186`) creates parent dirs. All four paths hold; `seed_builtin_servers_never_overwrites_an_existing_entry` and `test_global_config_dir_never_overwrites_an_operator_mcp_entry` both exercise the operator-collision case end to end (through the real `ensure_global_config_dir` entry point for the latter).

### Priority 3 — malformed-config claim

Both halves verified by reading the code, not just the tests. `read_config_for_seeding` (`mcp_config.rs:826-850`) never renames and never writes on malformed or non-object JSON — confirmed structurally, matching `seed_builtin_servers_declines_a_malformed_config_without_quarantining`'s stray-file assertion. `preseed_managed_trust` (`standalone/trust_seed.rs:242-284`) does quarantine on the same path, via a timestamped rename (`quarantine_path`, #4206), and runs immediately after seeding in both call chains that matter: `ensure_managed_config_dir_with_root` → `preseed_managed_trust` (`managed_config.rs:195` then `:434`), and `prepare_managed_config`'s `ensure_managed_config_dir` → `preseed_managed_trust` (`runtime/claude_code.rs:729` then `:823`). The blast-radius claim holds: a malformed-config user's OAuth state is lost either way, and this PR adds no second writer to that failure.

### Priority 4 — red-then-green honesty

Verified by diffing against `origin/main`: `ensure_global_config_dir` on main never opens `.claude.json` at all (only `.mcp.json`, via the removed `ensure_mcp_config`). The two "fail-open" tests (`test_global_config_dir_survives_a_malformed_claude_json`, `..._unreadable_claude_json`) therefore cannot go red against main by construction — there is nothing there to break — which is a correct, not evasive, characterization. The naive-design substitute (one line different: routing through the quarantining `read_config` instead of `read_config_for_seeding`) is a real demonstration, not a rationalization; the committed tests pass against this branch's actual code and are not vacuous.

### Also checked

- SLOC cap: `global_config.rs` is 921 raw lines / verified under the 500-SLOC production cap by CI (`500-line file-size cap` check: pass, "0 violations"); the malformed-JSON test literal was rebuilt via `serde_json::to_string(&json!({...}))` truncated by one byte rather than a raw `br#"..."#` literal, and still exercises the same truncated-mid-object scenario — no coverage lost.
- Clippy: CI's `Clippy` check passed (`-D warnings`), confirming the `as_bytes` fix compiles clean.
- No injector removed: `inject_trusty_mpm_mcp`, `inject_trusty_review_mcp`, `inject_trusty_memory_mcp`, `inject_trusty_search_mcp`, `custom_mcp`'s injector, and both `enabledMcpjsonServers` write sites (`trust_seed.rs:390`, `session_launch/settings.rs:799`) are all untouched by this diff.
- Diff scope: only `crates/trusty-mpm/{changelog.d,src/core/{mcp_config.rs,mcp_config_tests.rs,managed_config_tests.rs,standalone/global_config.rs}}` — no `ui/dist` or `package-lock.json` paths.
- Changelog fragment present and matches the change.

### Local verification

`cargo fmt --check -p trusty-mpm`, `cargo clippy -p trusty-mpm --all-targets -- -D warnings`, and `cargo test -p trusty-mpm` run in the worktree:
```
running 5027 tests (unittests src/lib.rs) ... test result: ok. 5020 passed; 0 failed; 7 ignored
running 1602 tests (unittests src/bin/tm/main.rs, bin target `tm`) ... test result: ok. 1601 passed; 0 failed; 1 ignored
```
Both match the PR body's reported counts exactly. A third, duplicate build of the same `src/bin/tm/main.rs` under the crate's second `[[bin]]` target (`trusty-mpm`, pre-existing duplicate-target setup unrelated to this PR — `crates/trusty-mpm/Cargo.toml`) failed two tests (`tests_behavior_b::compose_session_instructions_display_matches_live_prompt[_with_override]`) on an unrelated agent-roster snapshot mismatch (`copyeditor`/`pangram-editor`/`proofreader`/`writer`/`writing-critic` present in this machine's live-installed agent set but not in a frozen comparison string). This diff touches zero files under `crates/trusty-mpm/src/bin/tm/`, `src/assets/agents/`, `core/agent_source.rs`, or `core/agent_deployer.rs` (confirmed via `git diff --name-only origin/main...HEAD`), so this is pre-existing environmental drift, not caused by this PR.

CI on the PR: all gates green except `Test`, still pending as of this review (Clippy, format, MSRV, line-cap, changelog-fragment, SLD, and every UI-checks job already pass).

Issue #5091, opened 2026-08-07
ensure_project_indexed fail-open contract: silently-failed index create advances the pin anyway


bobmatnyc:
**Defect.** `ensure_project_indexed` in [`crates/trusty-common/src/search_index.rs:106`](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-common/src/search_index.rs#L106) is fail-open by design — its own doc comment ([L98-105](https://github.com/bobmatnyc/trusty-tools/blob/main/crates/trusty-common/src/search_index.rs#L98-L105)) says it "ALWAYS returns the derived id … so the caller can still pin the id even if the daemon is unreachable." `best_effort_create_index` (`search_index.rs:507`) and `best_effort_trigger_reindex` run the `POST /indexes` call on a spawned thread with a 1s/750ms timeout; a transport error or non-2xx is `tracing::warn!`-logged and dropped, never propagated. The `.mcp.json` index pin advances regardless of whether the index was ever created, and `search_health` reports ok for a pin pointing at nothing.

**Evidence.**
- Measured: 4 of 75 live worktrees have an index (~94% miss) — same measurement underlying [#5045](https://github.com/bobmatnyc/trusty-tools/issues/5045).
- Reproduced: `mcp__trusty-search__search` against a pinned index for a live worktree returns `404 Not Found: {"error":"unknown index"}` while health reports fine.
- Canonical fail-open shape for this repo: operation fails → downgraded to warn → state advances anyway → every alarm still reports healthy.

**Why now.** [#5065](https://github.com/bobmatnyc/trusty-tools/pull/5065) / [#5060](https://github.com/bobmatnyc/trusty-tools/issues/5060) add a new `ensure_project_indexed` call site at worktree creation. Worktree creation happens in bulk during session provisioning, so a bouncing/not-yet-up daemon silently produces un-indexed worktrees with no second chance to retry (unlike session launch or task start, a worktree is short-lived and nothing re-triggers indexing later).

[...]

Issue #3918, opened 2026-07-25
fix(trusty-mpm): daemon-managed session spawn seeds MCP trust into the wrong file, leaving trusty-mpm genuinely unapproved (not just deferred)


bobmatnyc:
## Summary

Companion to #3901. Bob reported a SECOND, distinct pause-failure wording:

> "The trusty-mpm MCP server isn't connected in this session, so
> session_context_pause is unavailable. Writing the snapshot directly in the
> same format."

#3901's root cause (a deferred-but-registered tool that the pause skill never
`ToolSearch`-loads) explains a PM *misdiagnosing* "unavailable" as "not
connected." That is real and stays the primary explanation for most
occurrences. But this variant's specific, stronger claim — the whole
trusty-mpm MCP **server** is absent, not just its schema unloaded — sent me
looking for a genuine connectivity gap. **I found one.** It is a real,
code-level defect, independent of the skill-wording issue, that specifically
affects daemon-managed sessions (`tm session new`, agent delegations, and any
`tm session new`-provisioned worktree PM session — which is what this
project's own `.trusty-mpm/sessions/` history and this very investigation's
worktree are).

## Root cause: two different trust-seed writers target two different files, and only the wrong one is actually read

There are **two independent code paths** that pre-approve a project's MCP
servers into Claude Code's per-workspace `enabledMcpjsonServers` list (the
mechanism that prevents the blocking "New MCP servers found" dialog — see the
already-closed #1296, which is the original bug this whole subsystem exists to
fix: *"tm spawn: sessions hang on 'new MCP servers found' dialog —
enabledMcpjsonServers not pre-seeded."*).

**Path 1 — correct, dynamic, but targets the wrong file for a managed session:**
`crates/trusty-mpm/src/core/session_launch/settings.rs:598-689`
(`preseed_workspace_trust`, called via `preseed_workspace_trust_home` at
`session_launch/mod.rs:710`, itself invoked by `prepare_session_inner` on
every `session_launch::prepare_session*` call — confirmed called from the
daemon's own session-provisioning code at
`crates/trusty-mpm/src/daemon/managed_routes/lifecycle.rs:1106`
(`prepare_session_with_repo_url(fw, worktree, Some(repo_url))`), so it DOES
run for `tm session new`). This function derives the server list **dynamically
from the project's actual `.mcp.json`** via `mcp_server_names(workspace)`
(`settings.rs:658-664`) — so it correctly includes `"trusty-mpm"` (and
[...]

Issue #2739, opened 2026-07-15
tm mcp add doesn't reach fleet sessions — inject native trusty MCP servers in session_launch


bobmatnyc:
## Problem
`tm mcp add` writes to the managed `CLAUDE_CONFIG_DIR/.claude.json` `mcpServers` map, and its help text (and the `tm-cli-operations` skill doc) claim the servers are "available in all your projects." This is **false for daemon-managed fleet sessions** (`tm session new`) — it only holds for the standalone `tm run` driver.

## Root cause (two stacking causes, confirmed against origin/main)
1. Fleet sessions are launched with `--setting-sources project,local` (`core/model_inject.rs:70` via `runtime/claude_code.rs:230-250`), which **excludes the `user` tier** where `mcpServers` live. `CLAUDE_CONFIG_DIR` on that path is for auth/keychain isolation only (DOC-34), so the managed `.claude.json` `mcpServers` map is never read.
2. The workspace `.mcp.json` is regenerated on every spawn by `core::session_launch::prepare_session` (`session_launch/mod.rs:505-533`) via exactly two hardcoded injectors — `inject_trusty_memory_mcp` + `inject_trusty_search_mcp`. It never calls `managed_mcp_server_names()`, so `tm mcp`-added servers never reach the project layer either.

Result: native MCP servers added via `tm mcp` (e.g. slack-mcp, telegram-mcp, gworkspace-mcp) are invisible to fleet sessions — existing or freshly spawned (not a restart/timing issue).

## Proposed fix
- Add a new injector in `core::session_launch` (called from `prepare_session`, alongside the memory/search injectors) that reads the **native trusty** MCP servers from the managed `.claude.json` `mcpServers` and merges each idempotently into the workspace `.mcp.json`. Scope: **native trusty servers only** (an explicit allowlist of trusty-family MCP binaries — e.g. slack-mcp, telegram-mcp, gworkspace-mcp, trusty-memory, trusty-analyze), NOT third-party/HTTP managed servers.
- Extend trust-seeding (`session_launch/settings.rs::mcp_server_names` / `preseed_workspace_trust`) so injected natives skip the "new MCP servers" approval dialog.
- Correct the misleading docs: `tm mcp` help (`bin/tm/commands/mcp.rs:226`) and `.claude/skills/tm-cli-operations/SKILL.md:44-47` — clarify that native trusty servers are injected into all managed sessions while non-native managed servers apply to the standalone driver only.

Diagnosed 2026-07-15.

Issue #1939, opened 2026-07-03
fix(memory): palace-level alias resolution for claude-mpm parity (owner-repo -> bare palace)


bobmatnyc:
## Problem

When `tm` (trusty-mpm) launches a session it pins the trusty-memory MCP server to a palace named via `derive_palace_id` (`crates/trusty-common/src/palace_id.rs`), which produces an `owner-repo` slug — e.g. `bobmatnyc-trusty-tools`. But the pre-existing claude-mpm-era palace for this repo is the BARE repo name `trusty-tools`.

The owner-repo palace does NOT exist (no `palace.json` under `~/Library/Application Support/trusty-memory/palaces/bobmatnyc-trusty-tools/`), so every MCP memory tool call fails with `palace metadata missing`, while the real history lives in the `trusty-tools` palace. Memory is split-brained.

## Fix (decision: ALIAS both names)

Implement palace-level alias resolution so `bobmatnyc-trusty-tools` resolves to the same store as `trusty-tools`.

1. **trusty-memory / trusty-common**: add a persisted PALACE-level alias map (distinct from the existing term/KG `add_alias`/`discover_aliases` entity aliases). When a palace is resolved by name and does not exist on disk, the resolver consults the alias map and redirects to the alias target.
2. **trusty-mpm**: on session launch, when the derived owner-repo palace does not exist but a bare-repo-name palace DOES exist, register the alias `owner-repo -> bare-repo-name`. Keep owner-repo canonical when there is no existing bare palace.

## Verify goal

After the fix, recalling palace `bobmatnyc-trusty-tools` returns the data stored under `trusty-tools`.

PR #5409, opened 2026-08-11
fix(trusty-mpm): delete the MCP injectors and the enabledMcpjsonServers approval (#4181, PR C2 of 3)

### LINKED ISSUE #4181: MCP config is written into ephemeral worktree .mcp.json, but these servers are user-scoped — per-worktree drift + provisioning race

bobmatnyc:
The final piece of ADR-0042. tm stops writing MCP configuration into a session
workspace, and stops pre-approving MCP server names. Both halves land here
because ADR-0042 states that shipping either alone "reintroduces #3926 in its
worst form."

Builds on #5394 (A — `trusty-search serve` honours `TRUSTY_INDEX`), #5398 (B —
`tm launch`/`tm connect` relocate `CLAUDE_CONFIG_DIR`), and #5406 (C1 —
`seed_builtin_servers` seeds user scope).

`Closes #4181`

## The five injectors, and both call sites

| Deleted | Was in |
|---|---|
| `inject_trusty_mpm_mcp`, `inject_trusty_review_mcp`, `inject_trusty_memory_mcp`, `trusty_memory_mcp_value`, `inject_mcp_server` | `session_launch/settings.rs` |
| `inject_trusty_search_mcp`, `trusty_search_mcp_value` | `session_launch/search_index.rs` |
| the whole module | `session_launch/native_mcp.rs` + its tests |
| the whole module | `session_launch/custom_mcp.rs` + its tests |
| `launch_trusted_mcp_names`, `launch_trusted_mcp_names_from` | `core/mcp_config.rs` |
| the four `trusty_*_injected` fields, `InjectedMcpPins` | `PrepReport`, `standalone/load.rs` |

Two call sites ran them, not one: `session_launch/mod.rs`'s
`prepare_session_inner`, and `runtime/claude_code.rs`'s `prepare_managed_config`
— which `spawn_resume` and `build_inplace_resume_command` reach with no
`prepare_session*` anywhere in their chain. Gutting only the first would have
left every resume still injecting. Both are gutted here.

`exclude_mcp_json_from_git` went with `native_mcp.rs`, along with its re-export
and both call sites. With no writer there is nothing to exclude.

`ensure_mcp_config` was already gone — C1 replaced it. This PR only clears the
stale doc references to it.

## Why the stale approval is stripped, not merely left unwritten

`enabledMcpjsonServers` is name-based and content-blind, and ADR-0042's
measurement shows what that means: with the name approved, the repo's `.mcp.json`
entry **wins** over the operator's user-scope declaration and its binary is the
one that spawns; unapproved, the colliding entry is inert. The approval is not
merely what enables name-squatting — it is the displacement mechanism itself.

So ceasing to write is not enough. Every machine an earlier tm launched carries
approvals it already wrote, on exactly the workspaces most likely to be reused.
Both seeders therefore `entry.remove("enabledMcpjsonServers")`, and both drop the
key from their idempotency check so an already-trusted entry does not return
early and keep it forever. `hasTrustDialogAccepted` and the onboarding keys stay
— that is #1269's other guarantee and unrelated to MCP.

**This also closes the trust widening #5398 shipped**, which the owner accepted
on the explicit basis that C2 removes the mechanism. `preseed_managed_trust`
unioned the operator registry with no `project_scope_mcp_names` subtraction
(#2739), so a malicious repo's `[mcp.custom]` name collision was pre-approved on
`tm launch`/`tm connect`. With no approval written, there is nothing to collide
with.

## What replaces the per-project pins

A single shared user-scope declaration cannot carry a per-project argument, so
the two pins the injectors wrote move to the environment. New module
`core::mcp_session_env`:

- `TRUSTY_MEMORY_PALACE` from `resolve_palace_slug` (#1605's pin)
- `TRUSTY_INDEX` from `register_project_index`'s **confirmed** id (#1373's pin;
  `None` on an unconfirmed create, per #5091, so a session never pins an index
  the daemon has never heard of)

Both stay gated on the same `[mcp] trusty_memory` / `[mcp] trusty_search`
manifest toggles that used to gate the injectors — which is why
`resolve_conditional_mcp_toggles` survives the deletion of the trust derivation
it was written for. Neither `claude_code.rs` nor `launch.rs` holds a
`HarnessPlan`, so each re-reads the manifest. The function moved to
`mcp_session_env` with its body unchanged.

Every spawn path exports them: `env_bin_prefix` (tmux panes),
`build_claude_command_with` (`tm launch` / `tm connect`), and
`InPlaceResumeCommand.mcp_env` (the bare-`tm` in-pane relaunch, which execs
directly and so sets them on the `Command`).

`managed_mcp_server_names` loses its four `_pinned` flags and C1's
builtin-subtraction with them — that subtraction was scaffolding for the C1→C2
interval, not design. It is now a plain builtins ∪ registry enumeration, and its
one surviving caller (`mcp_test`) is a diagnostic sweep.

`mcp_provenance` the module stays: `stray_mcp.rs` and `daemon/doctor_stray_mcp.rs`
consume the READ side, which ADR-0042's open question 3 needs. Only
`record_write_best_effort`'s call site went with `inject_mcp_server`.

## Alias-healing rehomed

`maybe_register_palace_alias` (#1939) had exactly one call site — inside the
memory injector. Deleting the injector as written would have stopped the
claude-mpm split-brain healing with no test failing. It now runs from
`ensure_managed_config_dir_with_root`, the choke point every spawn, resume and
in-place relaunch reaches, plus `prepare_session_inner` for the standalone path.
`resolve_palace_slug` survives as `pub(crate)` because it now derives the
`TRUSTY_MEMORY_PALACE` export.

## Tests: what was deleted, and why

**Deleted because the subject was deleted** (never because it failed):

- `native_mcp_tests.rs`, `custom_mcp_tests.rs` — their modules are gone.
- `tests_mcp_trust_seed_e2e.rs` — every test asserts an injector's write.
- `tests_manifest_toggle_trust_3934.rs` — asserts the toggle gating a trust
  derivation that no longer exists. Its surviving subject (the toggle gating the
  two pins) is the new `mcp_session_env_tests.rs`.
- In `mcp_config_tests.rs`: the `launch_trusted_mcp_names_from_*` tests and the
  `managed_mcp_server_names_*_pin_failed` / `_disabled_conditional_*` tests — the
  functions and flags are gone.
- In `trust_seed_tests.rs`: the eight tests asserting WHICH names land in
  `enabledMcpjsonServers`.
- In `session_launch/tests.rs`: the `inject_*` and `*_mcp_value_*` tests.
- In `standalone/load.rs` and `claude_code_tests.rs`: the tests asserting the
  four per-run pin bools and the injected `.mcp.json`.

**Rewritten to match new intended behaviour:**

- `tests_launch_trust_3926.rs` — rewritten, not deleted, as ADR-0042 requires.
  Its direction-A successor now uses a repo squatting a **builtin** name (the
  sharpest case the injectors existed for) and asserts no `enabledMcpjsonServers`
  entry is produced. Its direction-B test **inverts**: an operator's `tm mcp add`
  server used to be bridged into the workspace and approved there; it now reaches
  the session through user scope, so "bridged entry present" and "name approved"
  became the failure conditions. Two new tests cover the deletion itself and the
  stale-approval strip.
- `tests_claude_json_concurrency_4072.rs` — the concurrency defect it exists for
  is unchanged (two writers load-mutate-storing one file); the payload it checks
  for a lost update moved from the approval list to the onboarding pair.
- `standalone_isolation.rs` — the config-dir-versus-workspace boundary is
  unchanged; the assertions moved from `enabledMcpjsonServers` to the user-scope
  `mcpServers` map that actually carries the servers now.
- `trust_seed_tests.rs` gains `test_preseed_managed_trust_writes_no_mcp_approval`
  and `test_preseed_managed_trust_strips_a_stale_mcp_approval`.
- `claude_code_tests.rs` gains
  `prepare_managed_config_writes_no_mcp_json_and_no_approval`, covering the
  second call site.
- `session_launch/tests.rs` gains three `resolve_palace_slug_*` tests. That
  function's coverage used to ride on the injector tests; with those gone it
  needed its own, since it now derives an exported variable.

**Rewritten to hide a failure:** none.

Net: 1,124 insertions, 7,187 deletions across 39 files.

## Gates — rung 5

```
cargo fmt --all --check                                → EXIT=0
cargo check -p trusty-mpm --all-targets                → EXIT=0
cargo clippy -p trusty-mpm --all-targets -- -D warnings → EXIT=0
cargo test -p trusty-mpm                               → test result: ok. 4909 passed; 0 failed; 7 ignored
SKIP_UI_BUILD=1 cargo check --workspace                → EXIT=0
scripts/check_line_cap.sh                              → EXIT=0
scripts/check_sld.sh                                   → EXIT=0
scripts/check_capabilities.sh                          → EXIT=0
scripts/check_changelog_fragment.sh                    → EXIT=0
scripts/check_agent_assets.sh                          → EXIT=0
```

`git diff --name-only origin/main...HEAD` carries no `ui/dist` path and no
`package-lock.json`.

### `--include-ignored`: one pre-existing baseline failure

```
test result: FAILED. 4915 passed; 1 failed; 0 ignored
daemon::managed_routes::staleness_bench_tests::bench_stale_assets_for_many
  panicked at staleness_bench_tests.rs:115:14:
  called `Result::unwrap()` on an `Err` value: Io(Os { code: 2, kind: NotFound })
```

Not caused by this branch. It reproduces with `crates/trusty-mpm` checked out at
`origin/main` (`5026 passed; 1 failed`, same test, same panic), it passes when run
alone (`1 passed`), and `git diff --name-only origin/main...HEAD -- crates/trusty-mpm/src/daemon/`
returns only `doctor_transcript_saving.rs` and its tests — zero files under
`daemon/managed_routes/`. It is a parallel-execution/environment flake in a
benchmark, unrelated to MCP.

Separately, `compose_session_instructions_display_matches_live_prompt_with_override`
failed once in a full run and passed on two subsequent full runs and in isolation.
It compares the composed PM prompt against the live one; this branch touches no
file in the instruction pipeline.

## Fail-open arms

Every arm this PR touches degrades to a warning and still launches, and each is
covered red-then-green by a test that fails against the pre-change code:

- **Trust seeding.** `preseed_workspace_trust` / `preseed_managed_trust` remain
  non-fatal. `test_preseed_managed_trust_strips_a_stale_mcp_approval` fails on
  `main` (the key is rewritten, not removed) and passes here.
- **Stale approval on an already-trusted entry.** The idempotency check is the
  arm that would silently skip the strip.
  `prepare_session_strips_a_stale_enabled_mcp_approval` seeds a fully-trusted
  entry — on `main` the early return keeps the approval; here the key is gone and
  `lastCost` / `projectOnboardingSeenCount` survive.
- **A squatted builtin name.** `main` force-overwrites the entry and approves the
  name; `prepare_session_writes_no_approval_for_a_builtin_name_in_workspace_mcp_json`
  asserts the entry is left verbatim and unapproved.
- **Toggle-off pins.** `session_mcp_env_omits_*` assert a disabled toggle exports
  nothing rather than an empty variable — the shape that would send
  `trusty-search` at an index named `""`.

## Not in scope

The owner accepted trusty-mpm's existing SemVer violation, so the several
`pub fn` signature changes here (`env_bin_prefix`, `build_claude_command*`,
`preseed_managed_trust`, `prepare_interactive_config_dir*`,
`managed_mcp_server_names`) are not worked around.

ADR-0042 is updated in this PR: Status Proposed → Accepted, Q4 recorded as
resolved (relocate `CLAUDE_CONFIG_DIR`; the rejected alternative was adding user
scope to `--setting-sources`, which accepts the #1269 isolation loss), and the
`trusty-search` per-project-pin precondition corrected — it was false, since
#5394 had already landed.

---

## Critic round 1 — all four findings addressed (head `3d4af43e`)

Verdict WARN, https://github.com/bobmatnyc/trusty-tools/pull/5409#issuecomment-5249488218. No trust logic, injector deletion, or approval-strip behavior changed.

### HIGH — the rehomed #1939 alias healing had no test

`ensure_managed_config_dir_heals_a_bare_repo_palace_alias` in `crates/trusty-mpm/src/core/managed_config_tests.rs` seeds the bare `trusty-tools` palace with no owner-repo counterpart, provisions a workspace whose `origin` is `git@github.com:bobmatnyc/trusty-tools.git`, runs `ensure_managed_config_dir_with_root`, and asserts the slug `resolve_palace_slug` derives — the same derivation `session_mcp_env` exports as `TRUSTY_MEMORY_PALACE` — resolves through the alias store to the bare palace. Asserting the derived slug rather than a literal ties the healing to the name the session is actually pinned to.

Red with the `maybe_register_palace_alias` call commented out:

```
running 1 test
test core::managed_config::tests::ensure_managed_config_dir_heals_a_bare_repo_palace_alias ... FAILED

thread '...' panicked at crates/trusty-mpm/src/core/managed_config_tests.rs:866:5:
assertion `left == right` failed: provisioning must register bobmatnyc-trusty-tools -> trusty-tools so the pinned palace resolves to the claude-mpm-era store instead of failing with missing metadata
  left: None
 right: Some("trusty-tools")

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4916 filtered out
```

Green with it restored:

```
running 1 test
test core::managed_config::tests::ensure_managed_config_dir_heals_a_bare_repo_palace_alias ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4916 filtered out; finished in 0.38s
```

### HIGH — 14 dangling `Test:` citations

Each citation either points at the successor test or is deleted along with the subject it named:

| File | Disposition |
|---|---|
| `core/mcp_config.rs` | `seeding_and_workspace_injection_coexist_without_clobbering` deleted — its subject was the injector. The other 8 citations on `seed_builtin_servers` stand. |
| `core/session_launch/mod.rs` | `compiled_write_failure_does_not_skip_the_mcp_injectors` deleted, same reason. `prepare_session_with_repo_url` / `prepare_session_inner` now cite `palace_alias`'s tests, since the threaded remote feeds the alias healing rather than a pin. |
| `core/session_launch/settings.rs` | The two `preseed_trust_*_mcp_approval` names repointed at `prepare_session_writes_no_approval_for_a_builtin_name_in_workspace_mcp_json` and `prepare_session_strips_a_stale_enabled_mcp_approval`, which drive this function through `prepare_session` and read back the same `.claude.json` entry. Also drops the reference to `tests_mcp_trust_seed_e2e.rs`, a file this PR deleted. |
| `core/standalone/load.rs` | Four names deleted; `run_prepare_session_never_writes_real_home_claude_dirs` is the one that still exists. |
| `core/standalone/trust_seed.rs` | Module doc rewritten (below); citations now name the seven `test_preseed_managed_trust_*` tests that exist. |

`trust_seed.rs`'s module doc described the deleted four-`_pinned`-parameter `enabledMcpjsonServers` derivation and the #3918/#3934/#3950 hardening as current behavior. It now describes what the module does — trust seeding plus the #4206 stale-entry prune — with one paragraph recording that the derivation and its parameters went with the injectors. `SEEDER_WRITTEN_KEYS`'s doc said "the four keys written below"; three are written now, and the fourth stays in the list because droppings from an older tm still carry it.

Same class of stale prose fixed where it described `repo_url` feeding a deleted injector: `session_launch/mod.rs` (two blocks, including a broken intra-doc link to `inject_trusty_memory_mcp`) and `standalone/load.rs`.

```
test-pointers: resolved 23559 Test: citation(s) (floor 200) — 0 dangling pointers — OK.
EXIT=0
```

Removing the citation in `session_launch/mod.rs` to `prepare_session_repo_url_pins_palace` made that file's grandfathered allowlist row stale; `--update` pruned it, so `.test-pointer-allowlist.tsv` shrinks by one.

### HIGH — ADR-0042 consistency

`Date:` carried `2026-08-10 (accepted 2026-08-11)`, which fails the `YYYY-MM-DD` lint. Split into `**Date:** 2026-08-10` plus `**Accepted:** 2026-08-11`, the field seven other ADRs already use. `docs/adr/INDEX.md`'s row now reads `Accepted`, matching the file.

```
$ bash scripts/check_adr.sh
adr-check: 44 workspace ADRs and crate indexes are consistent
EXIT=0
```

### MEDIUM — the `mcp_env` carrier

One test per composition site, each driving a non-empty map. The palace value contains a space, so an unquoted assignment would split:

- `env_bin_prefix_carries_a_non_empty_mcp_env` — asserts `TRUSTY_MEMORY_PALACE='owner repo slug'` and `TRUSTY_INDEX='idx-42'` appear, and that they sit between the last `-u` flag and the binary (POSIX `env` grammar).
- `claude_command_carries_a_non_empty_mcp_env` — same assignments, and both precede ` claude`.
- `inplace_exec_command_carries_a_non_empty_mcp_env` — reads `Command::get_envs()` and asserts the values arrive verbatim with no quoting added (no shell parses this path), alongside an intact `CLAUDE_CONFIG_DIR`.

```
test core::model_inject::tests::claude_command_carries_a_non_empty_mcp_env ... ok
test runtime::claude_code::tests::env_bin_prefix_carries_a_non_empty_mcp_env ... ok
test commands::guided_inplace::tests::inplace_exec_command_carries_a_non_empty_mcp_env ... ok
```

### Gates — rung 5

| Gate | Result |
|---|---|
| `cargo check -p trusty-mpm --all-targets` | EXIT=0 |
| `cargo test -p trusty-mpm` | EXIT=0 — `test result: ok. 4912 passed; 0 failed; 7 ignored` (lib) plus 41 further binaries, 0 failed |
| `SKIP_UI_BUILD=1 cargo check --workspace` | EXIT=0 |
| `scripts/check_line_cap.sh` | 3896 files, 4 allowlisted, 0 violations |
| `scripts/check_sld.sh` | 57 specs + 3207 code files, 0 errors, 0 warnings |
| `scripts/check_capabilities.sh` | up to date (7 generated files) |
| `scripts/check_changelog_fragment.sh` | 39 changed paths, all crates recorded |
| `scripts/check_adr.sh` | 44 ADRs consistent |
| `scripts/check_test_pointers.sh` | 0 dangling pointers |

`git diff --name-only origin/main...HEAD` contains no `ui/dist` and no `package-lock.json`.

`bench_stale_assets_for_many` is untouched — pre-existing under `--include-ignored`, diff touches zero files in `daemon/managed_routes/`.


bobmatnyc:
## Verdict: WARN

Core security change verified sound: injector deletion and `enabledMcpjsonServers` removal land together (ADR-0042's central requirement); `hasTrustDialogAccepted` survives in both `preseed_workspace_trust` (settings.rs) and `preseed_managed_trust` (trust_seed.rs); both injector call sites are gutted (`session_launch/mod.rs`'s `prepare_session_inner` and `runtime/claude_code.rs`'s `prepare_managed_config`, the latter reached by `spawn_resume`/`build_inplace_resume_command` with no `prepare_session*` in their chain — confirmed via `prepare_managed_config_writes_no_mcp_json_and_no_approval`); the stale-approval strip is real (idempotency check now keys off `!contains_key("enabledMcpjsonServers")`, not a value match, and `test_preseed_managed_trust_strips_a_stale_mcp_approval` / `prepare_session_strips_a_stale_enabled_mcp_approval` both prove an already-trusted entry with a leftover approval gets it removed). `TRUSTY_MEMORY_PALACE`/`TRUSTY_INDEX` export correctly from `core::mcp_session_env`, gated on the right manifest toggles, and reach all three spawn paths (`env_bin_prefix` for tmux, `build_claude_command_with` for `tm launch`/`connect`, `cmd.env()` in `build_inplace_exec_command`). The `bench_stale_assets_for_many` pre-existing-failure claim checks out: diff touches zero files under `daemon/managed_routes/`, and I reproduced the failure only under the full `--include-ignored` suite (4915 passed, 1 failed) while it passes standalone.

Three HIGH findings, all documentation/test-citation drift rather than a live security hole — but two are currently failing this repo's own mechanically-enforced CI gates, and the third is the untested case ADR-0042 itself flagged as the highest-risk item to verify.

## Findings

| HIGH | `crates/trusty-mpm/src/core/standalone/managed_config.rs` | 197-204 | `ensure_managed_config_dir_with_root` now calls `maybe_register_palace_alias(project_dir, None)` to rehome the #1939 alias-healing that used to live inside the deleted `inject_trusty_memory_mcp`. No test exercises this call site — `managed_config_tests.rs` has zero references to `palace`/`alias`. The doc comment on this function cites a test, `ensure_managed_config_dir_heals_a_bare_repo_palace_alias`, that does not exist anywhere in the crate (confirmed via CI's own doc-pointer lint, see next finding). This is exactly the "silently stops running with no test failing" scenario ADR-0042 called out by name for this rehome. | Add a test in `managed_config_tests.rs` that seeds a bare-repo palace with no matching owner-repo palace, calls `ensure_managed_config_dir_with_root`, and asserts the alias was registered — mirroring `palace_alias.rs`'s own `creates_alias_for_split_brain` but driven through the real call site. | Fix here |
| HIGH | `crates/trusty-mpm/src/core/mcp_config.rs:336`, `session_launch/mod.rs:369`, `session_launch/settings.rs:445`, `standalone/load.rs:194`, `standalone/trust_seed.rs:57` | multiple | The repo's own CI gate (`Test: doc-comment pointer lint`, `scripts/check_test_pointers.sh`) is currently RED on this PR: 14 dangling `Test:` citations to test functions this PR deleted or renamed (`seeding_and_workspace_injection_coexist_without_clobbering`, `preseed_trust_writes_no_mcp_approval`, `preseed_trust_strips_a_stale_mcp_approval`, `run_prepare_session_pins_palace_from_clone_url`, `test_preseed_managed_trust_enables_mcp_servers`, and 9 more). `standalone/trust_seed.rs` lines 1-70 (the module-level `//!` doc) is the worst instance — it was never touched by this PR and still describes the deleted four-`_pinned`-parameter `enabledMcpjsonServers` derivation as current behavior, in a security-sensitive module, citing 8 tests that no longer exist. Verified independently: `gh pr checks 5409` shows this check `fail`, and I reproduced it locally. | Rewrite `standalone/trust_seed.rs`'s module doc (lines 1-70) to match the function-level doc already updated further down; fix every dangling `Test:` citation to name the real successor test (e.g. `preseed_trust_writes_no_mcp_approval`/`preseed_trust_strips_a_stale_mcp_approval` → `prepare_session_writes_no_approval_for_a_builtin_name_in_workspace_mcp_json`/`prepare_session_strips_a_stale_enabled_mcp_approval` in `tests_launch_trust_3926.rs`). | Fix here |
| HIGH | `docs/adr/0042-mcp-configuration-is-static-and-persistent.md:2-4`, `docs/adr/INDEX.md:55` | 2-4, 55 | CI's `DOC-N allocation and ADR consistency lint` (`scripts/check_adr.sh`) is currently RED: (1) the ADR's own `Status:` was flipped to `Accepted` but `docs/adr/INDEX.md` line 55 still lists it `Proposed`; (2) the `Date:` field `2026-08-10 (accepted 2026-08-11)` fails the mandatory `YYYY-MM-DD` format. Verified via `gh pr checks 5409` (`fail`) and reproduced locally: `bash scripts/check_adr.sh` exits 1 with exactly these two violations. | Update `docs/adr/INDEX.md`'s ADR-0042 row to `Accepted`; split the `Date:` field into `**Date:** 2026-08-10` plus a separate note (or an `**Accepted:**` field, if the template supports one) rather than embedding prose in the date field. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/runtime/claude_code.rs` (`env_bin_prefix`), `core/model_inject.rs` (`build_claude_command_with`), `bin/tm/commands/guided_inplace.rs` (`build_inplace_exec_command`) | 197-227, 319-350, 653-672 | No test drives a non-empty `mcp_env` through any of the three command-composition sites to confirm the `(name, value)` pairs actually land in the composed string / `Command` env, correctly quoted. `mcp_session_env_tests.rs` thoroughly tests the value-producing side (`session_mcp_env_with`'s toggle gating and slug/index derivation) but nothing tests the carrier once it leaves that function — every call site in production code passes `&[]` in its own tests. Lower risk since the quoting helper (`shell_single_quote`) is already proven elsewhere, but this is the one mechanism the whole ADR depends on to prevent #1373/#1605 regressing. | Add one test per carrier asserting a non-empty `mcp_env` produces the expected `NAME='value'` fragment / `cmd.env(NAME, value)` call. | Fix here |

## Notes

- Deleted-test audit (Priority One) came back clean: `native_mcp_tests.rs`, `custom_mcp_tests.rs`, `tests_mcp_trust_seed_e2e.rs`, `tests_manifest_toggle_trust_3934.rs`, and the trust-derivation tests removed from `mcp_config_tests.rs`/`trust_seed_tests.rs` all tested subjects (injectors, pin-tracking, toggle-gated approval) that this PR deletes outright — no case of a still-applicable security invariant losing coverage. `tests_launch_trust_3926.rs` was correctly rewritten rather than deleted, per the ADR's explicit instruction, and its new `prepare_session_strips_a_stale_enabled_mcp_approval` / `prepare_session_writes_no_approval_for_a_builtin_name_in_workspace_mcp_json` are strong end-to-end replacements for the old #3926 suite.
- `mcp_provenance`, `resolve_palace_slug` (now `pub(crate)`), `mcp_server_names`, `BUILTIN_MANAGED_MCP_SERVERS`/`builtin_server_entry`, and `managed_mcp_server_names` (repurposed to a pure diagnostic, `_pinned` params dropped) all survive exactly as ADR-0042 specifies.
- Changelog fragments correct: one file per category (`Removed`, `Added`) in `crates/trusty-mpm/changelog.d/`. No `ui/dist` or `package-lock.json` in the diff.
- Did not independently verify Q4/#5398 or C1/#5406 beyond confirming they're already on `main` and this PR builds on them correctly (`build_claude_command`'s `config_dir` param, `seed_builtin_servers`) — out of scope per the task framing (final PR of a 3-PR chain).

Issue #5407, opened 2026-08-11
a same-SHA PR-edit run can report Test SUCCESS without running it, masking the real run's result


bobmatnyc:
A PR title/body edit within seconds of a push starts a second CI run for the same commit SHA that reports `Test` passing without running it. GitHub shows only the latest-completing check-run per context name, so the spurious pass supersedes the real one everywhere, including `gh pr checks`.

Mechanism (verified on `origin/main` @ `1fec14a4`):
- `.github/workflows/ci.yml:211-217` — the `changes` job short-circuits to `docs_only=true` whenever `event_name == pull_request && event.action == "edited" && event.changes.base == null`, gating on event type only, never the diff.
- `ci.yml:128` — `cancel-in-progress` deliberately excludes `edited`, so the spurious run and the real run coexist.
- Every downstream job, `test` included, is gated on `docs_only`.

Had the real run failed, the later spurious pass would mask the failure identically, and a merge check reading `gh pr checks` would see green.

Evidence — PR #5403: force-push at `04:41:46Z` fired `synchronize` (run `31459286344`, real, `Test` SUCCESS in 14m55s, `docs_only=false`). A retitle 3s later fired `edited` (run `31459288334`, every step skipped, `Test` "SUCCESS" in 14s), which completed 32s after and now masks the real run in the UI/API.

Introduced by `0cf410b3` (#5395, "ci: scope checks to relevant changes"). Not a regression from #5366 (Test-job nextest sharding) — that PR doesn't touch the `changes` job.

Also stale: `ci.yml:6`'s header claims `Test` is one of five contexts required by branch protection. Live `required_status_checks.contexts` is `Format check`, `500-line file-size cap`, `Clippy`, `trusty-search daemon smoke test`, `MSRV check (1.94)` — no `Test`.

Blast radius measured clean: 469 merged PRs / 14 days, 59 with a sub-60s final `Test`, all verified genuinely docs-only. Latent, not an active incident.

Fix direction (one line, not prescribing): gate the `edited` short-circuit on the actual diff rather than the event type, or make it emit a neutral/pending status instead of success.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

Issue #5311, opened 2026-08-10
design: make doc-pointer-lint and SemVer checks required without deadlocking non-triggering PRs


bobmatnyc:
`semver-checks.yml` (tag-push/`workflow_dispatch` only) and `test-pointers.yml` (`paths-ignore: docs/**`) both report but structurally cannot block: on most PRs neither check runs at all, so making either required under branch protection deadlocks any PR that doesn't trigger them. PR [#5300](https://github.com/bobmatnyc/trusty-tools/pull/5300) would have deadlocked under a naive "add to required checks" rule. Dropped from `tm 1.3.6`'s exit criterion d as unbuildable-as-stated — see the [tm 1.3.6 milestone](https://github.com/bobmatnyc/trusty-tools/milestone/36) description.

The underlying gap is real: a red pointer lint nearly merged during the 1.3.5 train and was only caught because the merge agent was told to treat it as blocking regardless of what GitHub showed.

Needs a design pass before it's actionable, not just a required-checks toggle — either give both workflows a no-op success path on skip (so "did not run" reports green rather than absent), or add a branch-protection rule conditional on path/trigger. Reproduced/Observed, not speculative: the 1.3.5 near-miss and the #5300 deadlock scenario are both recorded incidents.

Closure conditions:
- A design decision recorded (workflow no-op-on-skip vs. conditional branch-protection rule)
- `Test: doc-comment pointer lint` and `Public API / SemVer` are both required checks on `main`
- A PR that doesn't trigger either workflow still merges without manual override

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

Issue #5149, opened 2026-08-07
chore(ci): move the SemVer gate from per-PR to release-time


bobmatnyc:
**Defect.** `.github/workflows/semver-checks.yml` (added by #5055, closing #5050) runs on every `pull_request`. Installing the pinned `cargo-semver-checks` and warming a cold `target/semver-checks` cache costs 20+ minutes under a 40-minute timeout, and that cost is paid before crate selection has decided whether the PR is even exempt — docs-only PRs included.

**Evidence.** A SemVer break only becomes a defect when something is published. The gate never had to run on a PR to prevent #4088; it had to run before `cargo publish`.

**Resolution.** Move it to the release path:

- `scripts/preflight-publish.sh` gains CHECK 5, running `scripts/check_semver.sh --crate <pkg>` immediately before `cargo publish`. This is the blocking barrier — `cargo publish` runs locally, so no CI job can stop an upload, and a crates.io publish is irreversible except by yank.
- `.github/workflows/semver-checks.yml` retriggers on `push: tags: ['*-v*']` plus `workflow_dispatch` for on-demand runs. The `pull_request` trigger is removed.

Accepted trade: between releases the repo is unprotected against this class of break. It surfaces at the release that would ship it.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5459, opened 2026-08-11
ci: report on every PR so the pointer lint and SemVer gate can be required (#5311)

bobmatnyc:
Refs #5311. Step 1 of 3 — see "Not in this PR" below.

## The problem

`Doc-comment pointer lint (Why/What/Test)` and `Public API / SemVer` both report
today but structurally cannot block. `test-pointers.yml` filtered its
`pull_request` trigger on `paths:`; `semver-checks.yml` had no `pull_request`
trigger at all. On most PRs neither check run exists, and GitHub waits forever on
a required context that never reports — so adding either to branch protection as
it stood would have made every open PR unmergeable. #5300 is the recorded
deadlock scenario; the 1.3.5 near-miss (a red pointer lint that nearly merged, and
was caught only because the merge agent had been told to treat it as blocking
regardless of what GitHub showed) is the recorded cost of it not being required.

## What changed, and which constraint each piece satisfies

The design is the one #5311 settled on: **no-op-on-skip**. The job always runs and
always reports; the skip moved off the trigger and into the steps.

**Report SUCCESS, never SKIPPED.** Job-level `if:` is gone from both jobs — a
skipped conclusion does not satisfy a required context. Each job ends in a no-op
step that *executes* and exits 0 (`No pointer-lint input changed — nothing to
lint`, `No release under test — nothing to compare`). The condition removed from
`test-pointers.yml` was `github.event.action != 'edited' || github.event.changes.base != null`.

**Key off the diff, not the event type** (#5407, PR #5442's pattern). Two new
classifiers, both fail-closed, both consulted by an always-running job:

- `scripts/detect-pointer-lint-inputs.sh` — is any changed path one the lint
  reads (`**/*.rs`, the allowlist, the gate scripts, the workflow)? These are the
  same globs that were on the trigger. It deliberately does **not** reuse
  `detect-docs-only.sh`, which classifies `.test-pointer-allowlist.tsv`,
  `check_test_pointers.sh` and `test-pointers.yml` as Cargo-inert — true for a
  build, and exactly wrong here, since a PR that only removes a fixed allowlist
  entry is the PR this gate most needs to run on. The selftest asserts that
  difference directly.
- `scripts/detect-version-bumps.sh` — does any crate declare a version this
  branch did not start with? Empty diff is a SCAN FLOOR error, not "nothing to
  check".

**#5149 is not reversed.** The 20+ minutes (pinned `cargo-semver-checks` + cold
`target/semver-checks`) sit behind `have_work == 'true'` at *step* level, so an
ordinary PR pays one `git diff`. Where it now does run is a release PR: this
workspace bumps the version in a PR and cuts the `<crate>-v<version>` tag from
merged main afterwards, so the verdict lands one step earlier than the tag push,
with the release still reworkable. `timeout-minutes` 40 → 60 because a release PR
can carry several crates serially where the tag path carries one.

`preflight-publish.sh` CHECK 5 remains the blocking gate for a publish; nothing
here weakens it.

## Test ladder

Rung 1 by class (CI/docs-only, no crate `src/**`), but the change is executable
CI configuration, so it ships with the fixture coverage that class of change owes.
Changelog fragment deliberately omitted — CI-only, no crate source.

```
$ bash scripts/check-ci-helpers-selftest.sh
check-ci-helpers-selftest: all 135 cases passed          # 106 before

$ actionlint .github/workflows/test-pointers.yml .github/workflows/semver-checks.yml
ACTIONLINT_EXIT=0

$ shellcheck scripts/detect-pointer-lint-inputs.sh scripts/detect-version-bumps.sh
NEW_SCRIPTS_SHELLCHECK_EXIT=0
```

29 new cases: the two classifiers' verdicts (Rust, allowlist, gate scripts,
unrelated, mixed, empty), `detect-version-bumps` against throwaway repos covering
bumped / unbumped / source-only / added / removed / multi-crate /
`version.workspace = true` / empty-diff-fails-closed, and structural wiring guards
that fail if a `paths:` filter or a job-level `if:` is ever reintroduced on either
workflow, or if either job starts branching on `github.event.action` again.

## Evidence the no-op path reports success

Both classifiers, run against a real branch off `origin/main` carrying a single
docs change:

```
$ git diff --name-only origin/main HEAD
docs/reference/semver-gate.md

$ POINTER_LINT_BASE=origin/main bash scripts/detect-pointer-lint-inputs.sh
detect-pointer-lint-inputs: 1 changed path(s) -> pointer_inputs_changed=false
pointer_inputs_changed=false
EXIT=0

$ VERSION_BUMP_BASE=origin/main bash scripts/detect-version-bumps.sh
detect-version-bumps: 1 changed path(s) -> bumped=false (none)
bumped=false
bumped_crates=
EXIT=0
```

`false` routes to the no-op step, which runs and exits 0 — success, not skipped.

The positive direction, against the real historical bump in #5435
(`trusty-common` 0.31.0 → 0.31.1):

```
$ VERSION_BUMP_BASE=2b0786a01^ bash scripts/detect-version-bumps.sh
  bumped: trusty-common 0.31.0 -> 0.31.1
detect-version-bumps: 2 changed path(s) -> bumped=true (trusty-common)
bumped_crates=trusty-common
```

### On this PR's own checks

**`Public API / SemVer`, [run 31515859033](https://github.com/bobmatnyc/trusty-tools/actions/runs/31515859033)** — the first time this
context has reported on a pull request at all. Job conclusion `success`, reached
through the no-op step, with every expensive step skipped at *step* level:

```
job "Public API / SemVer"                      -> success
  Refresh the base branch ref                  -> success
  Select crates to check                       -> success
  Install stable toolchain                     -> skipped
  Install cargo-semver-checks (pinned prebuilt)-> skipped
  Cache cargo artifacts                        -> skipped
  SemVer gate selftest                         -> skipped
  Enforce public-API SemVer against crates.io  -> skipped
  No release under test — nothing to compare   -> success
```

**`Doc-comment pointer lint (Why/What/Test)`, [run 31515859049](https://github.com/bobmatnyc/trusty-tools/actions/runs/31515859049)** — takes the *real*
path here, because this PR edits `test-pointers.yml` and the classifier, both lint
inputs. The gate still does its work:

```
job "Doc-comment pointer lint (Why/What/Test)" -> success
  Classify pointer-lint inputs from the diff   -> success
  Scan-floor selftest                          -> success
  Lint Test: doc-comment pointers              -> success
  No pointer-lint input changed                -> skipped
```

Its no-op path is proven above against a real docs-only branch and in the
fixtures; the next docs-only PR is its live case, and that is step 2 below.

## Not in this PR

Adding the required status contexts. The order matters and reversing it deadlocks
every open PR:

1. **This PR** — land the workflows.
2. Confirm on a merged PR that does not trigger either gate that both contexts
   report success rather than going pending.
3. Only then add `Doc-comment pointer lint (Why/What/Test)` and
   `Public API / SemVer` to `main`'s required contexts.

Current required contexts, read live before writing any of this (PR #5416 renamed
job names, so older lists are stale): `Format check`, `500-line file-size cap`,
`Clippy`, `trusty-search daemon smoke test`, `MSRV check`,
`PR version bump vs crates.io (issue #4421)`.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -43,6 +43,23 @@
 # therefore produces a real verdict wherever in the sequence it lands, and no
 # ordering between tag and publish needs to hold for it to mean anything.
 #
+# AND IT ALSO RUNS ON PULL REQUESTS, WITHOUT RESTORING WHAT #5149 REMOVED
+# (#5311). `Public API / SemVer` is meant to be a required branch-protection
+# context. Tag-push-only means the context does not exist on a pull request at
+# all, and GitHub waits forever on a required context that never reports — so
+# requiring it today makes every PR in the repo unmergeable. The workflow
+# therefore runs on `pull_request` too, unfiltered, and decides INSIDE the job
+# whether there is anything to compare: `scripts/detect-version-bumps.sh` reads
+# the diff and answers "does this branch declare a version it did not start
+# with". No, on the overwhelming majority of PRs — one git diff, then a success
+# that says so. #5149's cost is untouched: the pinned tool and the cold rustdoc
+# cache are installed only once the answer is yes.
+#
+# Yes is a release PR, and running there is the point. This workspace bumps a
+# crate's version in a PR and cuts the `<crate>-v<version>` tag from merged main
+# afterwards, so the bump PR is strictly EARLIER than the trigger above — a
+# break surfaces while the release can still be reworked, with no tag in flight.
+#
 # ON-DEMAND: workflow_dispatch takes a crate name so a break can be checked
 # without cutting a release. `bash scripts/check_semver.sh --crate <crate>`
 # does the same thing locally.
@@ -52,6 +69,12 @@ on:
   push:
     tags:
       - "*-v*"
+  pull_request:
+    # Explicit activity types (#4179), matching ci.yml and version-parity.yml:
+    # `edited` is not in GitHub's default set and is the event a base-branch
+    # RETARGET fires. No `paths` / `paths-ignore` — see #5311 above.
+    types: [opened, synchronize, reopened, ready_for_review, edited]
+    branches: [main]
   workflow_dispatch:
     inputs:
       crate:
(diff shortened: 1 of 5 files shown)


bobmatnyc:
## code-critic: APPROVE

Reviewed against #5311's accepted no-op-on-skip design. The three constraints hold, and the fail-open check comes back clean — I ran the error arms rather than reading them.

### Constraint 1 — reports SUCCESS, never SKIPPED

Verified structurally, not by inspection alone:

- No job-level `if:` on either job. No `paths:` / `paths-ignore:` on either `pull_request` trigger.
- **No `needs:` on either job**, so there is no upstream job whose skip could cascade. Both are single, standalone jobs.
- Each job terminates in a no-op step gated `!= 'true'` that executes and exits 0.
- The no-op step carries GitHub's implicit `success()`, so a *failed* classifier makes the job red rather than routing to the no-op. That is the correct direction and it is worth stating: the green path is only reachable through a classifier that succeeded.

### Constraint 2 — keys off the diff, and off the right base

Both classifiers compute `git merge-base <base> HEAD`, then `git diff --name-only --no-renames <merge-base> HEAD`. Neither reads `github.event.action`; the selftest asserts zero occurrences of it below `^jobs:` in both files.

The base is right, which is the part that matters most given #5407:

- Not `HEAD~1`, not the frozen `base.sha` snapshot. `Refresh the base branch ref` re-fetches `origin/<base.ref>` to the live tip first.
- On a `pull_request` event HEAD is the merge commit, whose first parent is the base. Whether `origin/main` has advanced past that parent or equals it, `merge-base` resolves to the merge ref's own first parent — so the diff is exactly this PR's changes in both cases.
- `--no-renames` means a renamed `.rs` file surfaces as delete + add and still matches. A rename would otherwise be a silent miss.

### Constraint 3 — branch protection untouched

Five files changed, none of them a protection call: no `branches/main/protection`, no `required_status_checks`, anywhere in the branch diff. `version-parity.yml` has zero diff, so `PR version bump vs crates.io (issue #4421)` keeps its name and is not orphaned.

### Fail-open check — ran, not assumed

| Probe | Result |
|---|---|
| `POINTER_LINT_BASE=origin/does-not-exist` | exit 2 |
| `VERSION_BUMP_BASE=origin/does-not-exist` | exit 2 |
| The semver step's `NAMES="$( … \| sed … )"` verbatim, with the script exiting 2 | `STEP_EXIT=2`; the line after the assignment never ran |
| The test-pointers classify step verbatim under GHA's default `bash -e` | `STEP_EXIT=2`; the line after never ran |

The semver step is the one that could have gone wrong: it ends a chain in a pipe, and a pipe swallows the left-hand exit status. `set -euo pipefail` is present in that step, so pipefail carries the 2 through `sed`, the bare assignment inherits it, and errexit aborts. The test-pointers classify step has no explicit `set`, but GHA's default `bash -e {0}` supplies errexit and there is no pipe to lose a status in.

**Failure is distinguishable from "nothing to do."** A failed lookup exits 2 and reddens the job; "nothing to do" exits 0 with `pointer_inputs_changed=false` / `have_work=false` and reaches the no-op. Two different exit paths, which is the property the required context depends on.

The two classifiers treat an empty diff differently — the pointer one calls it relevant and runs the lint, the version one calls it a SCAN FLOOR and exits 2. Both fail closed; the asymmetry is justified, since running the lint anyway is cheap and running 20 minutes of `cargo-semver-checks` anyway is not.

Error-arm coverage among the 29 new cases: `empty diff (fail closed) -> true` and `empty diff fails closed -> ERR2` both enter the error path, so these are not happy-path-only fixtures. The unresolvable-base arm has no fixture; I verified it by hand above.

### Claims I checked rather than accepted

- **`check_semver.sh --crate` accumulates.** The workflow builds `--crate X --crate Y` for a multi-crate release PR. `check_semver.sh:200` appends to `EXPLICIT_CRATES` and `:712-719` iterates the whole set. Had it been last-wins, every crate but one would have gone unchecked on a release PR. It is not.
- **Directory names resolve.** `detect-version-bumps.sh` emits `crates/` directory names; `resolve_crate` falls through to `dir_to_crate`, so `trusty-git-analytics` works alongside `tga`.
- **The awk version parser sees every real crate.** Ran `package_version` against all 27 `crates/*/Cargo.toml` — zero empty results. No crate can silently evade bump detection through a manifest the parser cannot read, which would have been a confident-green-forever bug.
- **The declined `detect-docs-only.sh` reuse is correct.** `check_test_pointers.sh` reads exactly `git ls-files '*.rs'` plus `.test-pointer-allowlist.tsv`; the classifier covers both, plus the gate scripts and the workflow. The selftest asserts the divergence directly (`detect-docs-only` calls the allowlist inert, this one does not) and both assertions pass.
- **The new cases actually run.** `check-ci-helpers-selftest.sh` runs in `ci.yml`'s `changes` job, which has no `paths:`, no `if:`, and no `needs:` — so the wiring guards against a reintroduced `paths:` filter or job-level `if:` fire on every PR rather than being decoration.
- **#5149's cost is preserved.** Five steps gated on `have_work == 'true'` at step level, asserted by count. The SLOC-cap rule is not touched by this PR.

Selftest run on the PR head: `check-ci-helpers-selftest: all 135 cases passed`, worktree clean afterward.

### Findings

|---|---|---|---|---|---|
| LOW | scripts/check-ci-helpers-selftest.sh | 552 | Comment says an inherited version's "workspace-level bump is its own diff entry", implying something acts on it. The loop's `case "$path" in crates/*/Cargo.toml)` skips the root manifest, so a root-level bump selects nothing. Unreachable today — `[workspace.package]` has no `version` field (#343) and all 27 crates declare a literal one — so this is an inaccurate comment, not a live gap. | Drop the second clause, or say plainly that a workspace-inherited version would not be detected and that #343 is why it cannot occur. | Fix here |
| LOW | .github/workflows/test-pointers.yml | 99 | Comment justifies the `else` branch by citing `workflow_dispatch` as "a deliberate manual run", but the workflow has no `workflow_dispatch` trigger — only `push` and `pull_request`. | Drop the clause, or add the trigger if a manual run was intended. | Fix here |

### Notes

- `detect-version-bumps.sh` writes `bumped=` / `bumped_crates=` into `$GITHUB_OUTPUT` from inside the `Select crates to check` step's command substitution, so `steps.crate.outputs.bumped*` exist alongside the `crates=` / `have_work=` the step sets itself. Unused and harmless — both values are single-line, so neither can corrupt the file. Flagging only so a later reader does not take the duplication for a bug.
- A PR whose diff against the refreshed base is genuinely empty (content already on main) turns `Public API / SemVer` red rather than green. Rare and fail-closed, so correct by this gate's own doctrine, but it is a false red on a context about to become required — worth knowing before step 3.
- Scope assumption: I reviewed step 1 only. Steps 2 and 3 of the PR body's sequence are where the context actually becomes required, and this approval says nothing about them.

No CRITICAL and no HIGH at >80% confidence. Approved.

Issue #5396, opened 2026-08-11
feat(trusty-common): retract_triple(subject, predicate, object) — remove one bad object without collateral loss of good siblings


bobmatnyc:
No primitive removes one object from a multi-valued `(subject, predicate)` KG pair without collateral loss of the good siblings. A legitimate subject can hold one bad object alongside correct ones under the same predicate — e.g. a real subject with `--is-a--> hard` sitting next to correct `is-a` rows — and there is no way to remove just the bad one.

`retract(subject, predicate)` (`crates/trusty-common/src/memory_core/store/kg_redb/write_ops.rs:90-116`) closes every object at that pair, taking the good siblings down too. `delete_by_subject(subject)` (same file, 179-221) is broader still. There is no 3-arg `retract_triple(subject, predicate, object)` — the doc comment at `write_ops.rs:84-85` records this as a deliberate non-goal as of #4810.

The obvious workaround — re-run `kg-rebuild` and let `assert` supersede the bad row — doesn't work either: `is-a` / `works-at` / `uses` / `depends-on` are absent from `FUNCTIONAL_PREDICATES` (`crates/trusty-common/src/memory_core/store/kg_store.rs:277-293`), so `assert` only supersedes an identical object, never displaces a different bad one.

#4678's `--purge-stale-subjects` flag on `kg-rebuild` cleans SUBJECT-side noise via `delete_by_subject`; object-side noise is explicitly out of that issue's scope, and nothing else schedules it. This issue is that schedule.

Closing this means adding `retract_triple(subject, predicate, object)` to `trusty-common`, reopening a shape #4810 declined — the API decision is the first question here, not an implementation detail. It's rung 4+ / cross-crate (trusty-test-ladder): `cargo check --workspace` plus `cargo test -p <consumer>` for each direct dependent of `trusty-common`'s KG write path.

Closure conditions:
- [ ] `retract_triple(subject, predicate, object)` API shape is decided (or explicitly re-declined with a reason) in `trusty-common`
- [ ] If accepted: implemented, with `cargo check --workspace` clean and dependent-crate tests passing
- [ ] A regression test proves a good sibling under the same `(subject, predicate)` SURVIVES the retraction of the bad object — a test that only proves the bad row went away would pass while the defect persists

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #4678, opened 2026-08-03
trusty-memory: KG extractor emits stopwords as first-class entities, capping graph-view usefulness


bobmatnyc:
## Summary

A live investigation of the trusty-tools palace (8,266 triples / 9,311 nodes), done while validating #4671's progressive KG graph loading, found that the extractor is emitting function words as first-class graph entities, and that this dominates the graph's structure badly enough to cap the analytical value of the graph view independently of how well it renders.

## Measured facts

- The top non-drawer entity nodes in the KG by degree are **stopwords**: `the` (degree 23), `a` (12), `it` (7), `that` (5). The extractor treats function words as first-class graph entities on par with real entities.
- Structural consequence: only **0.48% of all 8,266 edges** join two nodes that both have degree >= 2. The graph is effectively a star forest of `drawer:<uuid>` hubs with leaf attachments — the highest-degree hubs are pairwise unconnected. A top-K-by-degree induced subgraph legitimately returns **zero edges**.
- Impact: this caps the usefulness of the palace graph view independently of rendering quality. #4671 made the view scale (seed + expand loading with honest truncation signalling instead of silently mis-reporting a truncated graph as complete), but a graph whose most-connected "entities" are `the` and `a` has limited analytical value no matter how well it's paginated or laid out. Extraction quality is now the binding constraint on the graph view's usefulness, not layout or loading strategy.

## Suggested directions (not decisions)

- Stopword / function-word filtering at extraction time.
- A minimum-salience threshold before a candidate becomes a graph entity.
- Entity-type gating (e.g. require a recognized entity type, not bare token spans).
- Backfill cleanup of already-extracted noise triples in existing palaces.

Note: any backfill option affects existing palaces and needs its own care — it's a separate, riskier piece of work from fixing extraction going forward.

## References

- #4671 — progressive KG graph loading (surfaced this while validating the graph view against the live palace)
- #4670 — the issue #4671 closed (KG graph view silently mis-reporting truncation)

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Precondition 3 of ADR-0038 (KG recall lane) is NOT closed by this issue. This filter is lexical only — it reaches nothing beyond function-word tokens. #5399 tracks the remaining gap (content-word triples like `squash --is-a--> ancestor`) that actually closes it.

[...]

PR #5488, opened 2026-08-11
fix(trusty-memory): merge punctuated KG twins onto their cleaned nodes

### LINKED ISSUE #5401: trusty-memory: dedupe punctuated vs. cleaned entity nodes (#4678)
#4678's token-edge trim fix (`85fb0209` + `ca81166b`, `crates/trusty-memory/src/kg_extract.rs::clean_token`/`TOKEN_EDGE_PUNCT`) closes the extraction bug but opens a data-migration one: a palace with pre-fix content now extracts `redb` where it used to extract `` `redb` ``, and both nodes coexist. Every `kg-rebuild` over old drawer content grows the duplicate count — the same real-world entity stays permanently split across two nodes.

`is-a`/`works-at`/`uses`/`depends-on` are absent from `FUNCTIONAL_PREDICATES` (`crates/trusty-common/src/memory_core/store/kg_store.rs::FUNCTIONAL_PREDICATES`), so `assert` only supersedes an identical object and adds the cleaned spelling alongside the punctuated one instead of replacing it. `--purge-stale-subjects` selects on `is_stop_token`, which a punctuated non-stopword entity never trips.

This is a MERGE, not a delete. The punctuated node's triples — both where it is subject and where it is object — must be re-pointed onto the cleaned node's identity. A fix that deletes the punctuated node without re-pointing loses every triple it participated in.

Re-pointing an object means removing one object from a multi-valued `(subject, predicate)` pair without collateral loss of the other objects — exactly the primitive #5396 says doesn't exist (`retract_triple`). #5396 may be a prerequisite here, not merely related.

Closure conditions:
- [ ] Punctuated/cleaned entity twins (e.g. `` `redb` `` vs `redb`) are identified and merged into one node, never left as two
- [ ] Merge re-points all subject- and object-position triples from the punctuated node onto the cleaned node's identity — nothing is dropped
- [ ] A regression test proves the merged node retains BOTH the cleaned node's pre-existing triples AND the re-pointed ones (a test only asserting the punctuated node is gone must fail this bar)
- [ ] Dependency on #5396's `retract_triple` primitive is confirmed, or an alternative re-pointing mechanism is decided

bobmatnyc:
## Outcome

Closes #5401

Blocked until PR #5412 landed the `retract_triple` primitive it depends on.

## What changed / out of scope

Punctuated entity nodes (e.g. `` `redb` ``) accumulate alongside their cleaned twins (`redb`). This adds a `kg-rebuild --merge-punctuated-twins` pass that identifies twins via a new `canonical_entity(term)`, re-points triples in BOTH subject and object position onto the cleaned node, and retracts the twin's originals.

The pass is opt-in, mirroring #4678's `--purge-stale-subjects`; it does not fire during a plain rebuild. A punctuated node with no existing cleaned twin is still merged — that is a rename, and restricting to "both nodes already exist" would leave singletons that become twins on the next rebuild, which is the growth the issue describes.

Two properties that shape the implementation, worth stating: it asserts before retracting, so a crash between the two leaves the twin standing and the next run merges it again, where the reverse order would lose the fact. And `is_stop_token` partitions this pass from the purge — `canonical_entity` returns `None` for a punctuated stopword, so `("the` belongs to the purge and `` `redb` `` to the merge, and no term is ever both.

New file `commands/kg_twin_merge.rs`; also touches `kg_extract.rs`, `commands/kg_rebuild.rs`, `main.rs`, `tests/kg_rebuild_cli.rs`. `--dry-run` now takes a clap `ArgGroup` so it previews either pass.

## Risk

Mutates persisted KG state, opt-in only. `KgRebuildOptions` gains a public field — semver-breaking, covered by trusty-memory's unpublished 0.22.0 (crates.io tops out at 0.21.2).

## Test evidence — rung 5

Red-to-green proven by three separate neuterings of the implementation, each against the same command that is green on the committed tree:

- no merge pass (the pre-fix state): `two twins must have moved — left: 0, right: 2`
- delete-only "merge", assert dropped: `the merged node must keep its OWN pre-existing triple and gain the re-pointed one — left: ["database"], right: ["database", "mmap"]`
- pair-wide `retract` instead of #5396's `retract_triple`: `left: [], right: ["redb", "tokio"]`

The third is the direct evidence for the #5396 dependency — a pair-wide retract silently drops `trusty-memory --uses--> tokio` while re-pointing the punctuated object beside it.

Gates, all exit 0: `fmt --check`; `clippy -p trusty-memory --all-targets -- -D warnings`; `test -p trusty-memory` (716 lib passed, 27 binaries, 0 failures); `test -p trusty-common --features memory-core` (994 passed); `check --workspace`; line-cap (0 violations), SLD (0 errors), changelog fragment valid, test-pointers (23748 citations, 0 dangling), and `check-pr-version-bump.sh` reporting `[ OK ] trusty-memory 0.22.0 — src changed, version not yet published`.

## Baseline failures

None.

## Docs/changelog

Fragment at `crates/trusty-memory/changelog.d/5401-punctuated-twin-merge.md`. The fragment is in trusty-memory, not trusty-common — the change is trusty-memory-only; the issue's `trusty-common` label reflected a pre-#5396 suspicion that the storage primitive was missing.

## Review disposition

No critic round yet; one is being dispatched. The test-pointer lint caught a real defect pre-push (a doc citing three test names after they were folded into one), fixed and amended.

### DIFF
--- a/crates/trusty-memory/changelog.d/5401-punctuated-twin-merge.md
+++ b/crates/trusty-memory/changelog.d/5401-punctuated-twin-merge.md
@@ -19,6 +19,14 @@ Added
   either maintenance flag rather than on `--purge-stale-subjects` alone) to
   report the list while writing nothing. Selection is as narrow as the purge's: a
   term under the `drawer:`/`tag:`/`topic:`/`room:` namespaces never moves, only
-  triples stamped `auto:remember` are rewritten, and a punctuated stopword is
-  left to the purge — `is_stop_token` partitions the two passes so no term is
-  both deleted and re-pointed. Re-running is a no-op.
+  triples stamped `auto:remember` are rewritten, and a triple carrying a
+  punctuated stopword at EITHER end is left alone — so `is_stop_token` partitions
+  the two passes in the object position too, and the merge cannot hang a `("the`
+  off the cleaned node where the subject-selecting purge could never reach it.
+  Re-running is a no-op.
+
+  A re-point that cannot write is reported as failed rather than merged and
+  exits non-zero, the cleaned node is written before the punctuated row is
+  closed so a failure mid-way leaves the fact readable at one node or the other,
+  and a data root that cannot be listed fails the run instead of reporting an
+  empty one.
--- a/crates/trusty-memory/changelog.d/5401-punctuated-twin-merge.md
+++ b/crates/trusty-memory/changelog.d/5401-punctuated-twin-merge.md
@@ -1,0 +1,24 @@
+Added
+- **`kg-rebuild --merge-punctuated-twins`** folds a pre-#4678 punctuated entity
+  node onto its cleaned twin, so `` `redb` `` and `redb` stop being two nodes for
+  one thing. #4678's edge trim fixed extraction going forward and split every
+  entity already in the graph; nothing removed the old node, because the four
+  pattern predicates are absent from `FUNCTIONAL_PREDICATES` (an `assert` adds
+  the cleaned spelling beside the punctuated one) and `--purge-stale-subjects`
+  only selects subjects `is_stop_token` rejects, which a real entity never trips.
+  Each rebuild over the same drawer content widened the split.
+
+  This is a merge, not a delete: every auto-extracted triple is re-asserted under
+  the cleaned identity and only then retracted at the punctuated one, in BOTH the
+  subject and the object position, so the merged node keeps its own pre-existing
+  triples and gains the re-pointed ones. Object-position re-pointing is what
+  needed #5396's `retract_triple` — closing the whole `(subject, predicate)` pair
+  would have taken the punctuated object's correct siblings with it.
+
+  Off by default, prints every move, and takes `--dry-run` (which now gates on
+  either maintenance flag rather than on `--purge-stale-subjects` alone) to
+  report the list while writing nothing. Selection is as narrow as the purge's: a
+  term under the `drawer:`/`tag:`/`topic:`/`room:` namespaces never moves, only
+  triples stamped `auto:remember` are rewritten, and a punctuated stopword is
+  left to the purge — `is_stop_token` partitions the two passes so no term is
+  both deleted and re-pointed. Re-running is a no-op.
--- a/crates/trusty-memory/src/commands/kg_rebuild.rs
+++ b/crates/trusty-memory/src/commands/kg_rebuild.rs
@@ -19,6 +19,7 @@ use trusty_common::memory_core::palace::PalaceId;
 use trusty_common::memory_core::store::kg::{KnowledgeGraph, Triple};
 use trusty_common::memory_core::store::OpenIntent;
 
+use super::kg_twin_merge::report_merge;
 use crate::kg_extract::{
     extract_triples, is_stop_token, ExtractInput, AUTO_PROVENANCE, DRAWER_SUBJECT_PREFIX,
     ROOM_SUBJECT_PREFIX, TAG_SUBJECT_PREFIX, TOPIC_SUBJECT_PREFIX,
@@ -33,7 +34,7 @@ use crate::{resolve_palace_registry_dir, AppState};
 /// What: the four prefixes `kg_extract` emits. A subject carrying any of them
 /// is skipped before the stopword test runs.
 /// Test: `purge_skips_namespaced_subjects`.
-const STRUCTURAL_PREFIXES: &[&str] = &[
+pub(crate) const STRUCTURAL_PREFIXES: &[&str] = &[
     DRAWER_SUBJECT_PREFIX,
     TAG_SUBJECT_PREFIX,
     TOPIC_SUBJECT_PREFIX,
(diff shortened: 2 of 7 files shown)


bobmatnyc:
## Verdict: APPROVE

Re-review of `76ac21642` (`fix/5401-punctuated-twin-dedup`) against the prior WARN. Verified against pre-fix commit `2f2db2d11` by mechanically reverting the three production hunks (stopword check, `unwrap_or_default`, `ensure!(closed>0)`) onto the post-fix test file and running `cargo test -p trusty-memory --lib -- commands::kg_twin_merge kg_extract::tests::canonical_entity_names_the_cleaned_twin`: **8 passed / 2 failed** with the reverts in place, **10 passed / 0 failed** with the fix restored — matches the reported red→green claim. Full crate suite `cargo test -p trusty-memory --lib`: **718 passed, 0 failed, 4 ignored**. `cargo clippy -p trusty-memory --all-targets -- -D warnings` and `cargo fmt --check -p trusty-memory` clean locally. `merge_reports_a_failed_repoint_and_leaves_the_fact_readable` re-run 5x, stable.

## Prior findings — disposition

| Finding | Status | Evidence |
|---|---|---|
| HIGH — `list_palaces(...).unwrap_or_default()` fail-open | **CLOSED** | `kg_twin_merge.rs:231-232` now `.with_context(...)?`, propagates. New test `merge_palaces_propagates_an_unreadable_data_root` (`kg_twin_merge.rs:642-658`) fails against the reverted line (`expect_err` panics with `[]` — an empty summary vec, exactly the silent-success shape the finding described) and passes with the fix. |
| MEDIUM — stopword partition holds only in subject position | **CLOSED** | `kg_twin_merge.rs:104-109` adds `(subject.is_none() && is_stop_token(&t.subject)) \|\| (object.is_none() && is_stop_token(&t.object))`. Traced the namespace-safety claim directly: `is_stop_token` calls `clean_token` (`TOKEN_EDGE_PUNCT` trims edges only, `:` is never stripped from `tag:the`'s interior) then matches the *whole* normalised token against `STOPWORDS`, which holds only bare words (`"the"`, `"a"`, …) — so `tag:the`/`room:x`/`drawer:y`/`topic:z` never match and the namespace exemption in `canonical_term` is unaffected. Extended test `twin_repoints_leaves_the_purge_its_stopwords` (`kg_twin_merge.rs:179-200`) adds both the mirrored and stopword-as-object cases and fails against the revert (`redb --is-a--> ("the` was selected) before the fix. |
| MEDIUM — no test entered the failure paths (assert-before-retract unverified) | **CLOSED for the path it covers** | `merge_reports_a_failed_repoint_and_leaves_the_fact_readable` (`kg_twin_merge.rs:659-734`) genuinely exercises `repoint`'s `Err` → `summary.failed` → `PalaceMergeSummary::failure_count()` → `report_merge`'s returned count → `kg_rebuild.rs:189`'s `anyhow::bail!` chain end-to-end, and independently confirms the fact is still readable at the punctuated node afterward. Traced `retract_triple` (`crates/trusty-common/.../kg/ops.rs:330-356`): it returns `Ok(0)` on a no-op retract, not an `Err`, so `closed > 0` is a real, distinct invariant from the propagated-`Err` path this test hits — this test does not regress-test the `ensure!` line itself (see next row), which the engineer's own framing ("new coverage of already-sound behaviour, not a regression test") correctly reflects rather than overstates. |
| LOW — discarded rows-closed count | **CLOSED (code), untested by construction** | `repoint` (`kg_twin_merge.rs:337-345`) now captures `retract_triple`'s return and `anyhow::ensure!(closed > 0, ...)`s it into a per-re-point failure. Confirmed no test flips red when *only* this line is reverted (isolated it during the revert experiment) — consistent with the doc comment's own claim that the exclusive `OpenIntent::Writer` lock makes the arm unreachable today. Acceptable for a LOW item; the doc is honest about the gap rather than claiming coverage it doesn't have. |
| MEDIUM — `.env`→`env` needed naming in a test, not a behavior change | **CLOSED** | Diff to `kg_extract.rs` is doc comments + 5 new assertions only; `canonical_entity`'s body is byte-identical to pre-fix. Confirmed via `git diff` (no hunk touches the function body) and the full-suite pass above (no behavior regression anywhere else in the crate). |

## New findings

| LOW | crates/trusty-memory/src/commands/kg_twin_merge.rs | 122 | `out.sort_by(\|a, b\| key(&a.old).cmp(&key(&b.old)))` trips `clippy::unnecessary_sort_by` under the CI toolchain (rust 1.97), and the **Clippy required check is currently red** on this PR (run `31527175303`, `could not compile trusty-memory (lib) due to 1 previous error`). `kg_twin_merge.rs` doesn't exist on `main`, so this is in-scope for the PR to fix even though it predates the remediation commit (introduced in `2f2db2d11`, untouched by `76ac21642`). My local clippy (rustc 1.94.1) didn't flag it, which is presumably a version-gated lint difference — worth confirming before assuming it's fixed elsewhere. | `out.sort_by_key(\|a\| key(&a.old));` | Fix here |

## Notes

- **CI is currently red on two required checks**, neither blocking the code-quality verdict above but both blocking merge:
  - **Clippy** — the `unnecessary_sort_by` finding above; one-line fix.
  - **Doc-comment pointer lint (Why/What/Test)** — fails on `crates/trusty-memory/src/service/helpers.rs:331`, a dangling `Test:` pointer to `count_active_triples_surfaces_read_failure`. Confirmed this exists byte-identical on `origin/main` already — pre-existing baseline failure, unrelated to this diff, not scored against this PR.
- **Sibling fail-open, out of scope for this PR but worth a follow-up**: `merge_palaces`'s own doc comment says it has "the same per-palace error containment as `purge_palaces`" — but `purge_palaces` (`kg_rebuild.rs:368`) and `rebuild_palaces` (`kg_rebuild.rs:535`) still call `PalaceRegistry::list_palaces(...).unwrap_or_default()`, and `purge_palaces` is itself the other DESTRUCTIVE pass (`--purge-stale-subjects`) with the identical TCC/EPERM reachability the HIGH finding here was about. `audit_palaces` (`rooms.rs:128`) has the same pattern but is non-destructive. None of these three are touched by this PR's diff, so they're not scored here, but the fail-open class this PR closed in `merge_palaces` is still open in its two closest siblings — recommending a follow-up issue for `purge_palaces` at minimum, given it's destructive. This is a recommendation, not a filed issue.

No CRITICAL or HIGH findings remain open. APPROVED for the next pipeline stage once the Clippy line above is fixed.

Issue #5424, opened 2026-08-11
fix(trusty-common): poisoned adj lock masks committed KG retraction


bobmatnyc:
In `crates/trusty-common/src/memory_core/store/kg/ops.rs`, both `cascade_delete_by_drawer` and `retract_triple` (the latter added by PR #5412, not yet merged) commit the redb write first — the row closes and becomes durable — and only afterward take `self.adj.write()` to sync the in-memory adjacency graph. If that lock is poisoned, the call returns `Err` even though the persisted change already succeeded: a caller that retries on error sees `closed == 0` and reads it as "nothing to retract" rather than "already retracted, and the adjacency view may now be stale relative to storage."

Not a fail-open defect — it surfaces `Err` rather than silently defaulting. The mismatch is between what actually happened to storage and what the return value claims happened.

Surfaced as a MEDIUM finding (disposition: Parent) reviewing PR #5412 (closes #5396) — `retract_triple` replicates `cascade_delete_by_drawer`'s existing shape into a second call site rather than introducing it. Does not block #5412 or #5396.

Inferred from code reading: the commit-before-adjacency-sync ordering is directly visible in both functions; the poisoned-lock trigger itself has not been reproduced.

Closure:
- Both call sites decide whether a poisoned adjacency lock after a successful commit should surface as a distinct signal (e.g. a warning-carrying `Ok`, or an error variant that says "committed, adjacency stale") instead of a bare `Err`
- A regression test covers the poisoned-lock-after-commit path for at least one call site

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5489, opened 2026-08-11
fix(trusty-common): surface KG count-read failures instead of reporting zero

### LINKED ISSUE #5384: trusty-common: count_active_triples fails open on redb read errors, feeding kg_query's graph_empty misclassification
`count_active_triples()` (`crates/trusty-common/src/memory_core/store/kg_redb/read_ops.rs::count_active_triples`, lines 183-205 on `origin/main` — note this lives in `trusty-common`, not `trusty-memory`) fails open: `begin_read`, `open_table`, and `iter` errors each log a `tracing::warn!` and return `0` rather than surfacing the error.

Callers cannot distinguish a genuinely empty graph from a failed storage read. #4775 made `kg_query` classify a subject miss as `graph_state: "graph_empty"` vs `"subject_not_found"` using this count — so a redb read failure during that classification now yields `graph_state: "graph_empty"`, the exact false claim #4775 exists to prevent, reached via a rarer path (#4775 covers the common case; this is the residual it doesn't reach).

Not fixed in #4775: closing this requires a fallible signature on `count_active_triples`, which changes a `trusty-common` surface consumed across crates (`trusty-memory`, tests) — rung 4+ / cross-crate per this repo's test ladder, not rung 3.

Closure conditions:
- [ ] `count_active_triples` (or its call sites) can distinguish "zero triples" from "read failed" — fallible signature or equivalent
- [ ] `kg_query`'s `graph_state` classification no longer reports `graph_empty` on a storage-read failure
- [ ] A regression test exercises the error arm (forced `begin_read`/`open_table`/`iter` failure) and fails against the pre-fix commit — a fail-open branch is not reviewed without one

bobmatnyc:
## Outcome

Closes #5384

## What changed / out of scope

`KgStoreRedb::count_active_triples` returned `u64`, logging a warning and returning `0` from four separate arms, so a failed read was indistinguishable from an empty graph. It now returns `Result<u64>`; `KnowledgeGraph::count_active_triples` returns `Result<usize>`.

The reasoning that makes this a real fix rather than a signature change: `open_table` failing cannot mean "table not written yet", because `open_with_intent` creates every table on open — so every one of those arms was a genuine failure being swallowed. The per-row `continue` that silently undercounted is gone too.

Callers now decide what a failed read means for their own payload. `handle_kg_query`, `kg_graph_with_cap`'s `truncated` computation, and `GET /kg/count` propagate. Status, console-metrics and palace-info still degrade to 0 — they have no wire field for "unknown", and #4637 already established 0 as "unknown, never empty" there — but through one named helper, `service::helpers::kg_triple_count_or_zero`, which logs palace id and error, so the degrade is visible at the call site instead of hidden in the store. The chat prompt now prints `unknown (read failed)` rather than a `0` the model would repeat as fact.

Out of scope, deliberately: issue #5424 (redb commits before the in-memory adjacency syncs, so a poisoned lock returns `Err` on an already-committed retraction) is separately tracked and untouched.

## Risk

Cross-crate public API change in `trusty-common`, consumed by `trusty-memory`. `cargo check --workspace` passing is itself the proof no other crate consumed the old signature. No version bump owed — trusty-common is at 0.31.1, unpublished (crates.io max 0.31.0).

## Test evidence — rung 5

`count_active_triples_surfaces_read_failure` seeds three live triples then drops `ACTIVE_SUBJECT_COUNTS` in a write txn so `open_table` fails for real. Pre-fix, with the assertion in its pre-fix-compatible form, three live triples are reported as `0`:

`assertion left != right failed: a failed count read must not be reported as a count — left: 0, right: 0`

Post-fix, three tests pass and the error names `active_subject_counts`.

The committed assertion cannot be byte-identical to the red one because the signature changed — the red run proves the fail-open, the green run proves the error surfaces.

Gates, all exit 0: `fmt --check`; `clippy` on both `-p trusty-common` and `-p trusty-memory` `--all-targets -- -D warnings`; `test -p trusty-common --features memory-core` (995 passed); `test -p trusty-memory` (708 passed, 18 integration binaries); `check --workspace`; line-cap, SLD, changelog fragments all clean.

## Baseline failures

None.

## Docs/changelog

Fragments present for both trusty-common and trusty-memory.

## Review disposition

No critic round yet; one is being dispatched. One open question for the reviewer: the roll-up sites keep a 0-on-failure degrade — now logged and centralized, but still a degrade. Making them carry an explicit "unavailable" on the wire is a payload change and belongs in its own ticket.

### DIFF
--- a/crates/trusty-common/changelog.d/5384-count-active-triples-fallible.md
+++ b/crates/trusty-common/changelog.d/5384-count-active-triples-fallible.md
@@ -1,0 +1,6 @@
+Changed
+
+- `KgStoreRedb::count_active_triples` and `KnowledgeGraph::count_active_triples` return `Result` instead of failing open to `0` ([#5384](https://github.com/bobmatnyc/trusty-tools/issues/5384))
+  - `begin_read`, `open_table`, `iter`, and a per-row read each logged a warning and returned `0`, so a caller could not tell an empty graph from a storage read that never completed. `open_table` cannot mean "not written yet" here — `open_with_intent` creates every table on open — so the failure is always real.
+  - A dropped-row read no longer `continue`s past the error and undercounts.
+  - Breaking for callers that consumed the bare `u64` / `usize`; `trusty-memory`'s call sites are updated in the same change.
--- a/crates/trusty-common/src/memory_core/store/kg/ops.rs
+++ b/crates/trusty-common/src/memory_core/store/kg/ops.rs
@@ -68,14 +68,16 @@ impl KnowledgeGraph {
 
     /// Count currently active triples.
     ///
-    /// Why: Dashboard tally of live facts. Returns 0 on internal error so it
-    /// stays diagnostic-grade (matches prior behavior).
+    /// Why: Dashboard tally of live facts. Fallible since #5384 — swallowing
+    /// the store's read error here would hand every caller a `0` that reads
+    /// as "the graph is empty".
     /// What: Delegates to `KgStoreRedb::count_active_triples` and clamps the
-    /// u64 to `usize` for backward compatibility with existing callers.
-    /// Test: `count_active_triples_returns_live_only`.
-    pub fn count_active_triples(&self) -> usize {
-        let n = self.store.count_active_triples();
-        usize::try_from(n).unwrap_or(usize::MAX)
+    /// u64 to `usize`, which is what the existing callers consume.
+    /// Test: `count_active_triples_returns_live_only`,
+    /// `count_active_triples_surfaces_read_failure`.
+    pub fn count_active_triples(&self) -> Result<usize> {
+        let n = self.store.count_active_triples()?;
+        Ok(usize::try_from(n).unwrap_or(usize::MAX))
     }
 
     /// Number of distinct entities (nodes) in the in-memory adjacency.
--- a/crates/trusty-common/src/memory_core/store/kg/tests.rs
+++ b/crates/trusty-common/src/memory_core/store/kg/tests.rs
@@ -291,7 +291,7 @@ async fn upsert_drawer_replaces_existing_row() {
 async fn count_active_triples_returns_live_only() {
     let dir = tempdir().unwrap();
     let kg = KnowledgeGraph::open(&dir.path().join("kg.db")).unwrap();
-    assert_eq!(kg.count_active_triples(), 0);
+    assert_eq!(kg.count_active_triples().unwrap(), 0);
 
     kg.assert(Triple {
         subject: "alice".into(),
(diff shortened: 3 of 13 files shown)


bobmatnyc:
**code-critic verdict: APPROVE** — rung 5 satisfied, zero CRITICAL, zero HIGH.

**Fail-Open Check: PASS, with the residual named.** The two sites where a wrong count drove a decision — `KgMiss::classify` and the `truncated` flag — now propagate. The three that still degrade to `0` are pure read-only reporting where nothing advances past the failure, they log palace and error, and they are centralized in one named helper. Acceptable here, not the same defect relocated.

**The `open_with_intent` argument holds.** Verified at `kg_redb/store.rs:143-181`: a single init write txn touches `ACTIVE_SUBJECT_COUNTS` along with every other table on open. Init is skipped for `OpenMode::Snapshot`, but a snapshot is a byte copy of a file a writer already initialized, and the table has existed since the table set was introduced — so no supported on-disk file reaches this code without it. Every one of the four arms was a genuine failure being swallowed.

**The tell that confirms the diagnosis:** `list_subjects` and `list_subjects_with_counts` open the *same* `ACTIVE_SUBJECT_COUNTS` table and have always propagated `open_table` failure. `count_active_triples` was the one arm out of four that did not.

**The removed per-row `continue` loses nothing.** It skipped a redb row-read error — a storage/corruption condition, not a recoverable per-item one. The old behavior was a silently undercounted total no caller could detect.

**Correction to the PR body's cross-crate claim.** `cargo check --workspace` is *not* sufficient proof that no other crate consumed the old signature: six `trusty-common` features are enabled by no workspace member and never compiled by a workspace check, including `memory-core-kuzu`, which is adjacent to this code. The gap was closed by a repo-wide grep instead — the only `count_active_triples` call sites are in trusty-memory (all four updated) and trusty-common's own tests; `kuzu.rs` does not reference it. Conclusion unchanged, evidence corrected.

**One MEDIUM deferred to its existing ticket.** The monitor TUI renders the degraded `0` as a *known* count: `PalaceRow::kg_triples()` gates on `counts_unknown`, which `monitor/memory_client/parsers.rs:183` sets from `cached == Some(false)` alone, so a loaded palace whose count read failed prints `0` where `—` is correct. The KG Explorer already handles this (`KG.svelte:133-137` sets `count = null` on a failed fetch); the TUI has the equivalent state and cannot reach it. Fixing it means carrying explicit unknown through `PalaceInfo` — a payload change, tracked separately.

**One LOW fixed in `207cc4fa`:** the three propagating sites named the failing table but not the palace. All three now include the palace id; no new parameter was threaded through.

PR #5205, opened 2026-08-08
docs: fix stale claims in README and CLAUDE.md files across six crates

bobmatnyc:
Fixes documentation across six files that asserted things which do not exist. Every claim was verified against source before and after.

Closes #5193
Closes #5194
Closes #5195
Closes #5196

## The four filed issues

| File | Was | Is |
|---|---|---|
| `crates/trusty-search/README.md` | "18 tools" citing `src/mcp/tools.rs` | 21, citing `src/mcp/tools/descriptors.rs` (that path is a directory) |
| " | allowlist at `indexes.toml` | `allowlist.toml` — `allowlist/mod.rs`'s own doc says so explicitly |
| `crates/trusty-memory/README.md` | "25 tools" in prose, "30" in its own diagram | **45**, per `tool_definitions_lists_all_tools` |
| " | diagram drew a Unix domain socket bridge | corrected to `serve --stdio` → `POST /rpc`, matching the file's own prose ("No Unix domain socket is involved") |
| `crates/trusty-analyze/README.md` | `trusty-search daemon` | `trusty-search start` — no `Daemon` variant exists in `Commands` |
| " | "17 tools" | feature-dependent: 18 base + `console_metrics` + 3 `tr_review_*` under `--features review` |
| `crates/trusty-git-analytics/README.md` | pinned `tga-v2.7.0` | version-agnostic release link (crate is at 2.11.0) |

## Found during the sweep, fixed in the same work

Per this repo's opportunistic-fix rule, not filed separately:

- `memory_dream` → the registered tool is `palace_dream` (`chat_definitions.rs`); args corrected too.
- `memory_status` was listed as an MCP tool. It is **not** one — but it names a real capability, `GET /api/v1/status` (`StatusPayload`, `service/types.rs`), which is HTTP-only with no MCP counterpart. Reworded rather than deleted.
- The "full roster" list enumerated 32 of 45 tools. Replaced with a pointer to `tool_definitions()` rather than completing it — a hand-maintained list of 45 names re-rots at the next addition.
- 🔴 `crates/trusty-search/CLAUDE.md` and `crates/trusty-analyze/CLAUDE.md` carried the **same** stale claims as their READMEs. Fixing only the READMEs would have left the identical falsehood one file over, with no way for a reader to tell which was current. Both corrected.

[...]

PR #5503, opened 2026-08-11
feat(trusty-memory): expose kg_retract_triple on the MCP tool surface

bobmatnyc:
**Outcome.** No `Closes` line — there is no issue for this, by owner instruction ("don't file, just fix"). The gap was found empirically: a daemon verification created a KG test triple and could not retract it through any exposed MCP tool.

**What changed / out of scope.** Adds `kg_retract_triple`, wired to `KnowledgeGraph::retract_triple` (the trusty-common primitive that landed in #5412). Response is `{palace, subject, predicate, object, closed, retracted}`, plus `reason` when `closed == 0`. `object` is required — omitting it errors rather than silently widening to the whole pair. A hot-predicate retraction rebuilds the prompt cache, mirroring what `handle_kg_assert` does on the write side; without it a retracted Tier S fact keeps being injected.

Name choice: `_triple` names the key it takes. In this codebase `retract` means the whole `(subject, predicate)` pair, so a tool called `kg_retract` beside `kg_assert` would read as the pair-level operation and invite the over-deletion #5396 was about.

**`MemoryService::kg_retract_triple` (`service/core_kg.rs:108`) does not do what its name says.** It calls `handle.kg.retract(subject, predicate)` — the pair-level close — takes no `object`, and returns `bool`. It backs the HTTP route `kg_delete_triple` (`web/kg_routes.rs:284`), so an HTTP caller asking to delete one triple closes every object at that pair. That is the over-deletion #5396 exists to prevent, on a shipped endpoint. This PR does not fix it — the handler deliberately bypasses that method and calls `handle.kg.retract_triple` directly, which is also how `handle_kg_assert`, `handle_kg_query` and `handle_kg_list_subjects` already work (`MemoryService` is the HTTP adapter, not a tool-layer dependency). The HTTP path is being fixed separately.

A second, deliberate consequence: the diff does not touch `service/core_kg.rs` at all, so it cannot conflict with `fix/5384-count-active-triples-fail-open`. The only shared file is `tools/kg_ops.rs`, where this change is one self-contained inserted function.

**Also in this PR — a standing owner-ruled doc fix.** The `kg_list_subjects` description at `tools/definitions.rs` carried two stale claims, both required to be corrected by the next PR touching that file:
- "a subject that does not exist returns the same empty result as an empty graph" — false since #5385 (`41ca5c647`); `handle_kg_query` now returns `graph_state: "subject_not_found"` or `"graph_empty"` with a hint.
- "`truncated: true` means the page filled to `limit`" — false since #4810; the handler over-fetches `limit + 1` and sets `truncated` only when the extra row comes back, so a page that exactly fills `limit` reports `false`. `dispatch_kg_list_subjects_exact_limit_is_not_truncated` already proves it.

Also changed "alphabetically" to "ordered by subject" — the store iterates redb key order over raw subject bytes, which is not alphabetical for mixed case.

**Risk / blast radius.** MCP tool-schema addition, a public API surface. Tool count 46 → 47. No behaviour change to any existing tool.

**Test evidence — rung 6.** Rung 6 is owed because this is an API-surface change, and it was met with direct evidence rather than crate tests alone: an isolated throwaway daemon on `127.0.0.1:53852` under `TRUSTY_DATA_DIR_OVERRIDE`, real JSON-RPC `initialize` → `tools/list` handshake showing the served schema, then end-to-end — assert two objects at one pair, retract one (`closed: 1`), retract again (`closed: 0` with `reason`), and `kg_query` confirming the sibling `Globex` still active. The live `com.trusty.memory` daemon was neither contacted nor restarted.

Gates all exit 0: `fmt --check`; `clippy -p trusty-memory --all-targets -- -D warnings`; `check --workspace`; `test -p trusty-memory` → 711 passed, 0 failed, 4 ignored, 27 binaries; line-cap 0 violations; SLD 0 errors; changelog fragments valid.

**Documentation / changelog.** Two fragments (the changelog gate rejects a single fragment carrying two categories). `README.md` regenerated via `UPDATE_DOCS=1 cargo test -p trusty-memory --test generated_docs`.

Note the SLOC-cap handling: the new tests pushed `tools/tests.rs` to 3005 SLOC, over the 3000 test cap, so the split happened in this PR per the no-standalone-cap-fix rule — the tests moved to `tools/tests/kg_retract_tests.rs` as a child module with `use super::*`. No coverage moved or disappeared.

**Review disposition.** No critic round yet.

### DIFF
--- a/crates/trusty-memory/README.md
+++ b/crates/trusty-memory/README.md
@@ -246,7 +246,7 @@ per-palace conversation store: turns are stored verbatim and bypass the
 `memory_remember` signal/noise and dedup gates.
 
 <!-- BEGIN GENERATED: mcp-tools -->
-The MCP server registers **46 tools**. Authoritative source: `trusty_memory::tools::tool_definitions` —
+The MCP server registers **47 tools**. Authoritative source: `trusty_memory::tools::tool_definitions` —
 this table is generated from it, not maintained by hand.
 
 | Tool | Arguments | Summary |
@@ -266,8 +266,9 @@ this table is generated from it, not maintained by hand.
 | `kg_assert` | `palace`, `subject`, `predicate`, `object`, `confidence?`, `provenance?` | Assert a fact in the temporal knowledge graph. |
 | `kg_bootstrap` | `palace?`, `project_path?` | Seed the knowledge graph from well-known project files (Cargo.toml, package.json, pyproject.toml, go.mod, CLAUDE.md, .git/config). |
 | `kg_gaps` | `palace?` | List knowledge gaps detected in the memory palace graph. |
-| `kg_list_subjects` | `palace`, `limit?`, `with_counts?` | List the subjects this palace's knowledge graph actually holds, alphabetically. |
+| `kg_list_subjects` | `palace`, `limit?`, `with_counts?` | List the subjects this palace's knowledge graph actually holds, ordered by subject. |
 | `kg_query` | `palace`, `subject` | Query active knowledge-graph triples for a subject. |
+| `kg_retract_triple` | `palace`, `subject`, `predicate`, `object` | Retract one fact from the temporal knowledge graph — the inverse of kg_assert. |
 | `list_prompt_facts` | — | List every active prompt-fact triple (aliases, conventions, facts, shorthands) across all palaces. |
 | `memory_forget` | `palace`, `drawer_id` | Delete a drawer from a palace by its UUID. |
 | `memory_list` | `palace`, `limit?`, `room?`, `tag?`, `wing?` | List drawers in a palace, optionally filtered by wing, room type, or tag. |
--- a/crates/trusty-memory/changelog.d/kg-list-subjects-description-drift.md
+++ b/crates/trusty-memory/changelog.d/kg-list-subjects-description-drift.md
@@ -1,0 +1,12 @@
+Fixed
+
+- **`kg_list_subjects`' tool description told callers two things that stopped
+  being true.** It claimed a `kg_query` miss "returns the same empty result as
+  an empty graph, so a guess tells you nothing" — but since
+  [#5385](https://github.com/bobmatnyc/trusty-tools/issues/5385) a miss carries
+  `graph_state: subject_not_found` or `graph_empty` plus a hint. It also
+  claimed `truncated: true` means "the page filled to `limit` and more subjects
+  may exist" — but since
+  [#4810](https://github.com/bobmatnyc/trusty-tools/issues/4810) the handler
+  over-fetches one row, so `truncated` is set only when a further subject was
+  actually seen, and a page that exactly fills `limit` reports `false`.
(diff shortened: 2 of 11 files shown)


bobmatnyc:
**code-critic verdict: APPROVE** — rung 6 satisfied, zero CRITICAL, zero HIGH.

**Over-deletion probes clean.** `object` is required at all three layers — schema (`kg_retract_triple_required` always contains `object` in both `has_default` arms), handler (`.and_then(|v| v.as_str()).ok_or_else(...)` rejects missing, `null`, and non-string), and primitive (`retract_rows` narrows on exact equality). An empty-string object does not widen to the pair — it matches only a triple whose object is literally `""`, and `handle_kg_assert` stores objects untrimmed too, so assert and retract stay symmetric. No path reaches the two-argument `retract`. Adjacency does not over-delete either: `remove_edge_to` filters on `predicate == p && target == o_idx`.

**Scoping correct.** `memory.write`, same as `kg_assert`; `every_tool_has_scopes` would fail an omission. Enforcement is orchestrator-side by design — there is no in-dispatcher gate for any tool — so this adds no new class of exposure alongside the existing `palace_delete` / `memory_forget`.

**Schema/description mismatch on `palace` is not a contradiction.** `required` is built conditionally on `has_default`, and `resolve_palace` falls back to `state.default_palace` or errors naming the tool. The README shows it required because it renders `tool_definitions()` with `has_default = false`; a live server emits the schema matching its own state.

**Fail-Open Check: clean.** The diff adds exactly one downgraded failure — `rebuild_prompt_cache` error to `tracing::warn!` while the retraction still reports `retracted: true`. Nothing advances past a failure: `retracted` is `closed > 0` read straight off the store return value, the redb write commits before the rebuild is attempted, and the prompt cache is a pure denormalisation rebuilt from the KG. All five existing call sites downgrade identically, so there is no asymmetry to exploit. Residual: if the rebuild ever did fail, the cache holds the retracted fact until the next hot write, recorded only in the warn log — bounded staleness on a pre-existing path, not loss this diff opens.

**Retraction is a soft close.** `close_active_row` copies the row to a `hist:`-prefixed key with `valid_to_ms` set before removing the active row, so a wrong retraction is recoverable via `dump_all_triples`.

**One MEDIUM, fixed in `c7b74c79b`:** the prompt-cache rebuild branch had zero coverage — all three tests used `works_at`, which `is_hot_predicate` returns false for, so every test took the false arm. A new test now asserts under the hot predicate `is_fact` and was verified by neutering the rebuild call, which produced `retracted hot fact must not still be injected into the prompt`.

**One LOW, fixed in the same commit:** two stale hardcoded tool counts (`mcp_service.rs`, `openrpc.rs`). Dropped rather than updated, per the #5205 precedent — the real count is asserted by `tools_returns_expected_count`.

**Confirmed separately, outside this diff:** `MemoryService::kg_retract_triple` (`service/core_kg.rs:108`) takes no `object`, returns `bool`, and calls the pair-level `KnowledgeGraph::retract`. It backs `DELETE /api/v1/palaces/{id}/kg/triples/{triple_id}`, whose triple id is `base64url(subject\0predicate)` — the object is not in the key, so that endpoint is structurally incapable of naming one triple and closes every object at the pair. Being fixed separately.

PR #5510, opened 2026-08-11
fix(trusty-search): quarantine a corrupt corpus instead of recreating it empty (#4227)

### LINKED ISSUE #4227: trusty-search: genuine corpus corruption never quarantines — #4122's population is transient-dominated

bobmatnyc:
## What was broken

A genuinely corrupt `index.redb` was recreated empty and served as healthy, so a live watcher rebuilt a partial corpus over it.

`open_corpus_db_or_recreate` (`core/corpus_recovery.rs`) treated `Storage(Corrupted(_))`, `RepairAborted` and `Storage(Io(InvalidData))` the same as a stale redb-2.x file: move it aside, return `Ok` with a fresh EMPTY corpus. Because that was an `Ok`, `build_indexer_from_entry` (`service/persistence_loader/mod.rs:151`) took its success arm and wired the empty store with `corpus_open_failed` left **false** — and the #4122 write quarantine, the #4087 query-surface `503`s, and `/health`'s degraded signal all key on that flag. So the index came up reporting healthy with 0 chunks and an active watcher, and ordinary file saves persisted a fresh partial corpus over the recreated one. That is the #4122 data-loss shape with every guard against it switched off.

A side effect of the same bug: `CorpusOpenFailure::FormatIncompatible` — the one classification that names corruption and prescribes a rebuild — was unreachable from the real open path, because corruption never propagated as `Err`.

## Which half of #4227 this is, and the #4087 boundary

#4227 asked for two things: correct the quarantine module's condition list, and decide whether the recreate-to-empty path warrants its own guard.

- The **doc half already shipped** on main (`core/indexer/quarantine.rs:31-55`, the 0.44.0 index-integrity work). This PR rewrites that section, because the fix makes it false.
- This PR is the **guard half**. #4227's own comment points at #4087 for it, but #4087 closed on 2026-08-04 via PR #4718 and shipped only the **query-surface** half — making an already-`corpus_open_failed` index fail loudly instead of returning `200 []`. Nothing in it reaches the producer side, and every guard it added keys on the flag corruption never set. This PR fixes the producer so #4087's shipped machinery actually fires for corruption. No #4087 scope is absorbed; there is none left to absorb.

## The fix

Corruption backs the file aside and returns a typed `CorpusCorrupted`, which reaches the loader's failure arm and quarantines the index.

`UpgradeRequired` is deliberately **excluded** and still recreates silently. It names a known old on-disk format with a data-preserving recovery tool (`trusty-search migrate-redb`, which links redb 2.6 under the `redb2` alias precisely to read these files), so quarantining it would turn every redb major upgrade into an outage. The split is the new `is_genuine_corpus_corruption`, beside the existing `is_incompatible_corpus_format` that decides only whether the file is moved aside.

- The damaged bytes are still renamed to the `.v2-incompatible` sibling **before** the error returns, so failing closed never destroys the operator's recovery source.
- No replacement file is created, so the next boot opens a clean corpus and boot reconcile reindexes from source.
- `CorpusOpenFailure::classify` recognises the marker as `FormatIncompatible`, so status surfaces say "rebuild" rather than the `Unclassified` "diagnose first" wording.
- The `corpus_open_failed == true` implies `corpus == None` invariant that `quarantine.rs` calls load-bearing is preserved by construction: the corrupt path wires no store rather than wiring one alongside the flag.

## Fail-Open Check

Every failure branch touched has an error-arm regression test that fails against the pre-fix commit. All three integration tests were written first and confirmed red on `8e6ca079d` for the right reason — not an unrelated panic:

```
test corruption_quarantine_preserves_the_original_corpus_bytes ... FAILED
test genuine_corruption_quarantines_the_index ... FAILED
test corrupt_corpus_refuses_watcher_writes_and_chunk_count_stays_zero ... FAILED

genuine_corruption_quarantines_the_index panicked at ...:168:
  a GENUINELY CORRUPT corpus must set corpus_open_failed — pre-fix it was
  recreated empty and wired, leaving the index reporting healthy (#4227)

corrupt_corpus_refuses_watcher_writes_and_chunk_count_stays_zero panicked at ...:246:
  precondition: a corrupt corpus must quarantine the index (#4227)

corruption_quarantine_preserves_the_original_corpus_bytes panicked at ...:339:
  precondition: corruption must quarantine (#4227)

```

Post-fix, same binary:

```
test genuine_corruption_quarantines_the_index ... ok
test corruption_quarantine_preserves_the_original_corpus_bytes ... ok
test corrupt_corpus_refuses_watcher_writes_and_chunk_count_stays_zero ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.65s
```

`corrupt_corpus_refuses_watcher_writes_and_chunk_count_stays_zero` drives the real `spawn_watch_loop` and OS watcher with a healthy sibling index as a positive control, so a flat `chunk_count` is a proven refusal and not a slow watcher. All three use `#[tokio::test(flavor = "multi_thread")]` — the corpus open parks on `spawn_blocking` via `open_serialized`, and the watcher bridges an OS notification thread into the async loop.

## Test fixture repair, folded in

`incompatible_corpus_is_backed_up_and_recreated` claimed to simulate a redb-2.x file but wrote 0xAB garbage, which produces `Storage(Io(InvalidData))` — the **corruption** bucket — so the `UpgradeRequired` path it was named for was never exercised. Replaced with `old_format_is_backed_up_and_recreated_without_error`, which builds a real redb 2.x database through the `redb2` dependency and asserts the fixture really produces `UpgradeRequired` before asserting the #702 contract. A future change that quarantines old-format files now fails here instead of in production on the next redb major bump.

## Gates — rung 5 (persistence / data integrity)

All run at `-p trusty-search` on this branch:

| Gate | Result |
|---|---|
| `cargo check -p trusty-search --all-targets` | `EXIT=0` |
| `cargo clippy -p trusty-search --all-targets -- -D warnings` | `EXIT=0` |
| `cargo test -p trusty-search` | `1585 passed; 0 failed; 1 ignored` (lib) + `387 passed; 0 failed; 3 ignored` (integration) + 28 further binaries, all ok |
| `cargo check --workspace` | `EXIT=0` |
| `scripts/check_line_cap.sh` | `3914 files, 0 violations` |
| `scripts/check_changelog_fragment.sh` | `scanned 6 changed path(s); all 1 crate(s) with source changes are recorded` |

`--include-ignored`, scoped to this change's blast radius:

```
test core::redb_migrate::tests::real_v2_incompatible_smoke ... ok

test migration_e2e_already_current_is_noop ... ok
test migration_e2e_empty_corpus_advances_to_current ... ok
test result: ok. 2 passed; 0 failed
```

No baseline flake fired on this run — the `trusty-search` watcher tests and `execute_doctor_against_test_daemon` all passed. Nothing was ignored, cfg-gated, excluded, or narrowed to make a gate green.

## Version

**No public API shape changed.** `CorpusCorrupted`, `is_genuine_corpus_corruption` and `open_corpus_db_or_recreate` are all `pub(crate)`. The two public surfaces touched — `CorpusStore::open` and `CorpusOpenFailure::classify` — keep their signatures; `open` returns `Err` in a new case, which is a behavior change inside an existing `Result`, not an API break. A patch bump is sufficient.

This branch reads `0.45.0`, which is published, so the `PR version bump vs crates.io` check fails until PR #5508 (`patch bump to 0.45.1`) lands on main; rebasing then clears it with no change here.

Closes #4227


bobmatnyc:
**code-critic verdict: APPROVE** (reviewed `b9a194bc2` against `8e6ca079d`)

Verified by direct execution, not inspection: checked out both commits and ran `cargo test -p trusty-search --test corpus_corruption_quarantine_4227` — 3 passed at the PR head, 0 passed at the base with all three failing on their named assertions (`corpus_open_failed`, `is_write_quarantined()`, preconditions) rather than an unrelated panic. All three use `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]`, avoiding the documented single-thread `block_in_place` trap. Unit suites `corpus_recovery` 7/7 and `open_failure` 13/13; sibling integration suites `corpus_open_quarantine_4122` 3/3 and `quarantine_durable_writes_4226` 4/4 — no regressions.

Independently confirmed against redb's own source that `is_incompatible_corpus_format` / `is_genuine_corpus_corruption` bucket all 5 `DatabaseError` and 6 `StorageError` variants correctly, with everything outside both predicates falling through to the raw error arm — file left alone, never silently recreated. `InvalidData` is precisely targeted: only redb's magic-number mismatch on a non-empty file is reachable via the `create()` path this code uses. Rename-then-error fails closed. `CorpusCorrupted` traced end-to-end to `is_write_quarantined()`. SemVer: all new items `pub(crate)`, `CorpusStore::open` and `CorpusOpenFailure::classify` signatures unchanged, `redb2` was already a dependency — patch bump correct.

One MEDIUM finding, documentation-precision only, being fixed in this PR: redb reports `UpgradeRequired` from a version byte read at two fixed offsets *before* that slot's checksum is computed, so a single-byte corruption landing there and yielding 1 or 2 is indistinguishable from a genuine old-format file and is silently recreated rather than quarantined. Inherited from redb's classification; not reachable from trusty-search without re-parsing the header independently.

Issue #278, opened 2026-05-27
fix(trusty-memory): KG auto-extract — tag-deny-list, cascade-delete, per-triple delete endpoint


bobmatnyc:
## Background

Follow-up split from #148. Three quality gaps in KG auto-extract remain after the provenance/confidence/drawer-ID work landed. Provenance (`auto:remember`), confidence (`AUTO_CONFIDENCE = 0.6`), drawer-ID tracking, and `KnowledgeGraph::retract(subject, predicate)` are all DONE per audit. This ticket covers the three remaining gaps.

## Gap 1 — tag-deny-list at extract time

`extract_triples()` in `crates/trusty-memory/src/commands/kg_extract.rs` runs unconditionally. Drawers tagged `cross-project-qa`, `test`, `fixture`, etc. should be excluded from auto-extraction (their content is by definition not factual project knowledge).

Acceptance: a configurable deny-list (default: `["cross-project-qa", "test", "fixture"]`) is consulted before extract; tagged drawers produce zero triples. Unit test covering the deny path.

## Gap 2 — cascade-delete of derived triples when a drawer is deleted

When a drawer is deleted, triples that were auto-extracted from it (identified via the `drawer:<uuid>` subject prefix) should be removed too. Currently no `source_drawer_id` cross-reference exists in `kg_redb.rs`.

Acceptance: `memory_forget(drawer_id)` cascades to delete all triples whose subject is `drawer:<that-uuid>`. Unit test verifying both directions (extract → forget → triples gone).

## Gap 3 — per-triple DELETE endpoint

Today the HTTP API exposes `(subject, predicate)` retract via the `/kg/prompt-facts` endpoint, but no `DELETE /kg/triples/<id>` for surgical removal of one triple at a time.

Acceptance: `DELETE /kg/triples/<id>` removes one triple by its primary key; returns 404 if not found; returns 204 on success. Integration test.

## Related

- Parent: #148 (closed when this is filed)

bobmatnyc:
Shipped in **trusty-memory 0.6.0** — released to crates.io. See [release tag](https://github.com/bobmatnyc/trusty-tools/releases/tag/trusty-memory-v0.6.0). All three sub-features delivered:

1. **Tag deny-list at extract time** — `DEFAULT_DENY_TAGS` (`cross-project-qa`, `test`, `fixture`) prevents auto-extraction from drawers tagged as test/fixture content. Configurable via `KgExtractConfig.deny_tags`.
[...]

PR #5512, opened 2026-08-11
fix(trusty-memory): put the object in the KG triple id so DELETE names one edge

bobmatnyc:
**The defect.** `DELETE /api/v1/palaces/{id}/kg/triples/{triple_id}` was structurally incapable of naming one triple. The id was `base64url(subject\0predicate)` — the object was not in the key — so the handler called the pair-level retract and closed EVERY object at that pair. Confirmed independently by code-critic before the fix.

**The fix.** The id is now `base64url(subject\0predicate\0object)` and the route calls `KnowledgeGraph::retract_triple`.

**Why the id gained the object rather than the route gaining a query param or being renamed.** Nothing in the repo produces these ids: `encode_triple_id` is `#[allow(dead_code)]` and called only from tests; no endpoint returns a triple id (`/kg/all`, `/kg/graph`, `/kg/graph/seed`, `/kg/graph/neighbors` all return raw `Triple` rows with no id field); `crates/trusty-memory/ui/` never calls the route; ids are never persisted, a caller builds one from fields it already has. The alternatives were worse — a query param splits one resource's identity across a path segment and a query string while still calling the segment `triple_id`, and keeping pair-wide deletion defeats the route's whole purpose (#278), which pair-wide retraction already serves via `/kg/prompt-facts`.

**🔴 BREAKING, state it prominently.** Two breaks:
- An external caller holding a two-field id now gets **400** with the new format — deliberately NOT folded into the existing malformed-id 404, because on a delete path "not found" reads as "already deleted", which is the wrong thing to tell someone whose request was never understood. Undecodable garbage still 404s, unchanged.
- `MemoryService::kg_retract_triple` changed signature: takes the object, returns rows-closed `usize` instead of `bool`. Its only caller is the HTTP handler.

Note that the release-time SemVer gate (`preflight-publish.sh` CHECK 5) is what enforces the matching version bump, and that it does not run on PRs — so this break is recorded here deliberately rather than left for the release to discover. Do not bump any version in this PR.

**One fix beyond the literal bug.** The HTTP path never rebuilt the prompt cache after a retraction, so a retracted Tier S fact kept being injected until the next write. The service method now mirrors `tools/kg_ops.rs`: rebuild when `closed > 0 && is_hot_predicate(predicate)`, failure logged not fatal.

**Test ladder: rung 5.** Include the red→green proof. The red, against `origin/main`'s code, seeding two objects at one pair and deleting one through the router:

```
thread 'web::tests::kg_tests::red_proof_delete_over_deletes_siblings' panicked at
crates/trusty-memory/src/web/tests/kg_tests.rs:808:5:
assertion `left == right` failed: deleting one triple must leave the sibling object at the same pair active
  left: 0
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 712 filtered out
```

`left: 0` means the endpoint returned 204 and closed both objects. Three permanent tests replace it: the sibling survives and can then itself be deleted so the pair still empties one row at a time; a legacy pair-id is rejected 400; a miss returns 404, covers idempotency (204 then 404 on repeat), and asserts a miss closes nothing.

**Gates, all on the rebased head** — summarise as command + result:
```
cargo fmt --check                                           EXIT=0
cargo clippy -p trusty-memory --all-targets -- -D warnings  EXIT=0
cargo test -p trusty-memory                                 EXIT=0  (716 passed, 0 failed, 4 ignored, + 26 further binaries ok)
cargo check --workspace                                     EXIT=0
scripts/check_line_cap.sh          0 violations — OK
scripts/check_changelog_fragment.sh  OK trusty-memory
scripts/check_sld.sh               0 errors, 0 warnings
```

**One gate NOT green, and say so plainly rather than burying it.** Rung 5's `--include-ignored` run fails 4 tests in `crates/trusty-memory/tests/concurrent_perf.rs`. That file posts to `const HTTP_BASE: &str = "http://127.0.0.1:7070"` — the launchd-managed daemon installed on this machine, whose cwd is `/`, so `palace_create` fails with "no project root found at or above '/'". The branch's code never runs in that path. Evidence it is not this PR: `git diff --name-only origin/main -- crates/trusty-memory/tests/` returns zero paths. Environmental, pre-existing.


bobmatnyc:
## Verdict: WARN

Adversarial review of `7d06ca76f` against base `025d71772` (rung 5, trusty-memory). Zero CRITICAL findings; one HIGH.

### Blast-radius claims (crux of the breaking-change decision) — all verified true

- `encode_triple_id` is `#[allow(dead_code)]` and called only from tests — confirmed: every non-test call site is `crates/trusty-memory/src/web/tests/{kg_tests.rs,recall_tests.rs}`; nothing in production code calls it.
- No endpoint returns a triple id — confirmed. `kg_query`, `kg_list_all`/`kg/all`, `kg_graph`, `kg_graph_seed`, `kg_graph_neighbors` (`crates/trusty-memory/src/web/kg_routes.rs`) all return `Json<Vec<Triple>>` or a struct built from `Triple`, whose type (`crates/trusty-common/src/memory_core/store/kg/types.rs:46-56`) has no `id` field.
- The UI never calls the route — confirmed. `grep -rn "kg/triples\|DELETE" crates/trusty-memory/ui/src` is empty; `ui/src/lib/api.js` has no client for this endpoint.
- Ids are never persisted — consistent with the above; nothing in the diff or the surrounding code stores an encoded id anywhere.

Given all four hold, keying the id on the object (rather than a query param or a route rename) is the right call — nothing decodes an old-format id except a caller constructing one from fields it already has.

### 400 vs 404 split — correctly implemented

`decode_triple_id` (`crates/trusty-memory/src/web/kg_routes.rs:290-303`) collects UTF-8 fields first (any decode/UTF-8 failure → `Malformed` → 404), then checks arity: exactly 3 → `Ok`, exactly 2 → `LegacyPair` → 400, anything else → `Malformed` → 404. No input can land in both branches: the split is on `\0`, so field count is deterministic, and a 3-field encoding of an all-empty-tail triple (`subject\0predicate\0`) still splits to 3 parts (verified against the existing `("subject", "", "")` case in `decode_triple_id_round_trips`, `recall_tests.rs:29-32`), so it can't be confused with the 2-field legacy form.

### Signature change — caller audit confirmed

`grep -rn "\.kg_retract_triple(" crates/` (workspace-wide) returns exactly one call site: `crates/trusty-memory/src/web/kg_routes.rs:350`. The MCP tool of the same name (`crates/trusty-memory/src/tools/kg_ops.rs:103-150`, `handle_kg_retract_triple`) is a separate, pre-existing implementation added in a prior PR (025d71772) that calls the store layer (`handle.kg.retract_triple`) directly — it does not go through `MemoryService::kg_retract_triple` and is untouched by this diff. The signature-change claim ("its only caller is the HTTP handler") is accurate.

### Scope creep (prompt-cache rebuild) — in scope, faithfully mirrored

`crates/trusty-memory/src/service/core_kg.rs:132-137` mirrors `crates/trusty-memory/src/tools/kg_ops.rs:128-134` exactly: same condition (`closed > 0 && is_hot_predicate(predicate)`), same log-and-continue on rebuild failure. Belongs in this PR — it's the same retract path the PR is already touching, and leaving the HTTP surface stale relative to the MCP surface would be a second, related bug.

### Fail-Open Check — HIGH finding

| HIGH | crates/trusty-memory/src/service/core_kg.rs | 132-137 | The new cache-rebuild-on-retract branch is exercised by zero tests. The three new HTTP-path tests (`kg_tests.rs:834`, `887`, `942`) all assert on `(alpha, is, …)`, and `"is"` is not in `HOT_PREDICATES` (`crates/trusty-memory/src/prompt_facts.rs:66+`), so `is_hot_predicate(predicate)` is `false` on every call `cargo test -p trusty-memory` makes into this function — neither the success arm (cache actually rebuilt) nor the log-not-fatal failure arm ever runs. The doc comment's own `Test:` line (`core_kg.rs:117-118`) cites those same two tests as coverage for this method, which is misleading: they cover the retract-and-return-count behavior, not the rebuild branch the doc paragraph right above spends four sentences justifying. Contrast with the MCP-side sibling test `dispatch_kg_retract_triple_rebuilds_prompt_cache_for_hot_predicate` (`crates/trusty-memory/src/tools/tests/kg_retract_tests.rs:169`), whose own docstring (lines 155-162) explicitly names this exact hazard — "nothing stops a later edit from inverting the condition or dropping the call while every existing test stays green" — and was written specifically to guard the twin implementation. That same reasoning wasn't applied to the newly-added HTTP-side copy. | Add one test analogous to `dispatch_kg_retract_triple_rebuilds_prompt_cache_for_hot_predicate`, but through the HTTP router: assert a hot-predicate triple, confirm it's in `get_prompt_context`, `DELETE` it through `/kg/triples/{id}`, confirm the prompt cache no longer contains it. | Fix here |

This is not a data-loss finding — the underlying `retract_triple` delete is precisely scoped to the named triple and independently covered at the store layer (`crates/trusty-common/src/memory_core/store/kg_redb/tests.rs:978-1075`, unchanged by this PR). The exposure is a stale prompt-context cache continuing to surface a retracted Tier-S fact, which is what the doc comment itself says is the failure mode — it just isn't proven.

### Red→green proof — reproduced independently

Checked out `025d71772` in a separate worktree, hand-wrote the scenario the PR body describes (seed two objects at one pair via the HTTP router, `encode_triple_id(subject, predicate)` two-arg form, `DELETE`, assert the sibling survives) and ran it:

```
thread 'web::tests::kg_tests::red_proof_delete_over_deletes_siblings' panicked at crates/trusty-memory/src/web/tests/kg_tests.rs:809:5:
assertion `left == right` failed: deleting one triple must leave the sibling object at the same pair active
  left: 0
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 716 filtered out
```

Matches the PR body's claimed red exactly. The three permanent tests are not vacuous:
- `kg_delete_triple_closes_one_object_and_keeps_siblings` (`kg_tests.rs:834`) directly re-proves the fixed invariant (sibling survives), then empties the pair one row at a time — this is the test that would fail against the pre-fix code, confirmed above.
- `kg_delete_triple_returns_404_for_missing` (`kg_tests.rs:887`) covers the miss path (404, names the object, doesn't touch the sibling) and idempotency (204 then 404 on repeat).
- `kg_delete_triple_rejects_a_legacy_pair_id` (`kg_tests.rs:942`) covers the 400 path and asserts zero rows closed.

Ran the PR's own gate chain on the head commit: `cargo fmt --check -p trusty-memory` clean, `cargo clippy -p trusty-memory --all-targets -- -D warnings` clean (zero warnings), `cargo test -p trusty-memory` → `716 passed; 0 failed; 4 ignored`, matching the PR body's reported counts exactly.

### concurrent_perf.rs claim — confirmed in one line

`git diff --name-only origin/main -- crates/trusty-memory/tests/` and `git diff --name-only 025d71772 7d06ca76f -- crates/trusty-memory/tests/` both return zero paths — the diff touches nothing in that directory, so the 4 `--include-ignored` failures being pre-existing/environmental is credible and not attributable to this PR.

### Notes

- `crates/trusty-memory/src/tools/kg_ops.rs::handle_kg_retract_triple` duplicates the retract + hot-predicate-cache-rebuild logic that now also lives in `core_kg.rs::kg_retract_triple`, rather than the HTTP handler routing through the MCP tool's implementation or vice versa. Not flagged as a finding — `kg_ops.rs` is unchanged by this diff and CLAUDE.md's common-entry-point rule explicitly scopes to cross-crate duplication, not within-crate. Worth a `Promote` recommendation for whoever next touches either surface, not a blocker here.

bobmatnyc:
## Re-review after remediation — head `2d678ea4`

## Verdict: APPROVE

**The prior HIGH is CLOSED.**

## Delta reviewed

`git diff 7d06ca76...2d678ea4` touches exactly 3 files: `.test-pointer-allowlist.tsv` (2 lines removed), `crates/trusty-memory/src/service/core_kg.rs` (doc comment + 3-line inline comment), `crates/trusty-memory/src/web/tests/kg_tests.rs` (+104, one new test). No production logic changed — only the new test and documentation.

## Findings

| — | — | — | No issues found at >80% confidence | — | — |

Zero-finding review. No CRITICAL, HIGH, MEDIUM, or LOW findings in the delta.

## Verification performed

**1. Mutation test — real coverage, independently reproduced.** Built a worktree at `2d678ea4`, ran `cargo test -p trusty-memory --lib web::tests::kg_tests`: baseline 16 passed / 0 failed, including `kg_delete_triple_rebuilds_prompt_cache_for_hot_predicate`. Mutated `service/core_kg.rs:135` (`if closed > 0 && ...` → `if false && closed > 0 && ...`) and reran: `test result: FAILED. 15 passed; 1 failed` — only `kg_delete_triple_rebuilds_prompt_cache_for_hot_predicate` failed; the other three delete tests (`_closes_one_object_and_keeps_siblings`, `_returns_404_for_missing`, `_rejects_a_legacy_pair_id`) stayed green. Reverted the mutation. The new test genuinely enters the branch and nothing else does.

**2. Unreachability of the `if let Err(e)` arm — each link checked, all hold.**
- `rebuild_prompt_cache` (`prompt_facts.rs:575-582`) has exactly one fallible step: `gather_hot_triples(state).await?`.
- `gather_hot_triples` (`:255-261`) is `Ok(gather_hot_facts(state).await?...)` — its only fallibility is `gather_hot_facts`.
- `gather_hot_facts` (`:310-348`) never returns `Err` as written: the per-palace `Err(e)` arm inside its loop only does `tracing::warn!` and falls through to the next palace; the function unconditionally returns `Ok(out)` at `:347`.
- `build_prompt_context` (`:191`) returns a bare `String`, not a `Result` — infallible by signature.
- The cache write is `tokio::sync::RwLock` (confirmed at `lib.rs:25,412,689`, not `std::sync`), which carries no poisoning; `.write().await` (`:579`) returns the guard directly, never a `Result`.

So `rebuild_prompt_cache` cannot currently return `Err` — the comment at `core_kg.rs:137-138` ("it cannot currently return `Err`") is accurate, hedged correctly (claims "currently," not "never"), and matches what the code does today. This is a genuine dead branch, not a disguised fail-open.

**3. Allowlist edit — verified via a full workspace scan, not just inspection.** Ran `scripts/check_test_pointers.sh` (check mode) against the worktree at `2d678ea4` (~4 min). Result: exactly one violation, `helpers.rs:331` citing `count_active_triples_surfaces_read_failure` — the known, out-of-scope #5489 pointer the task explicitly excludes (fixed by #5521). Zero violations and zero stale-allowlist entries touching `kg_routes.rs`. Cross-checked by hand: `kg_delete_triple_returns_204_on_success` is cited nowhere in `kg_routes.rs` at `2d678ea4` (superseded by `kg_delete_triple_closes_one_object_and_keeps_siblings`, already true at the WARN commit `7d06ca76`); `kg_delete_triple_returns_404_for_missing` is still cited at `kg_routes.rs:329` and the test of that name exists at `web/tests/kg_tests.rs:887` — both already resolved before this PR's delta. Removing the two allowlist rows is a correct staleness prune, not suppression of a real dangling pointer.

**4. Doc-pointer honesty — accurate.** `core_kg.rs:117-121` lists all three tests and states "Only the third reaches the cache rebuild — the other two retract under the predicate `is`, which is not hot." Verified: `kg_delete_triple_closes_one_object_and_keeps_siblings` and `kg_delete_triple_returns_404_for_missing` both seed with predicate `"is"` (`kg_tests.rs:838,889`); `HOT_PREDICATES` (`prompt_facts.rs:66-71`) is `is_alias_for`, `has_convention`, `is_fact`, `is_shorthand_for` — `"is"` is not among them. The new test uses `is_alias_for` (hot). `core_kg.rs:136-138`'s claim that the error arm has no test and why is accurate per finding 2 above. Cross-reference to `dispatch_kg_retract_triple_rebuilds_prompt_cache_for_hot_predicate` (new test's doc comment) resolves to a real test at `tools/tests/kg_retract_tests.rs:169`.

## Notes

`helpers.rs:331` is red in `check_test_pointers.sh` at this head, exactly as flagged in the task — pre-existing drift from #5489, outside this diff, already fixed by open PR #5521. Not scored here.

Issue #5357, opened 2026-08-10
trusty-search: the reindex root-override trust gate reads indexes.toml at request time, so a concurrent registry write can slip an untrusted move through


bobmatnyc:
[PR #5335](https://github.com/bobmatnyc/trusty-tools/pull/5335) made `reindex_handler` mirror the runner's own gate (`needs_path_relativization` then `root_move_is_trusted` against the persisted root) and refuse a disagreeing override with 409. That closed the case where the handler accepted an override the runner would then abort.

The residue: the handler decides on `indexes.toml` at request time and mutates both roots; the runner re-reads at spawn time, possibly minutes later behind the global and per-index reindex semaphores. A write flipping this index's `root_path` in between lets an override through that the runner then aborts — leaving the indexer on the new root with the corpus still relativized against the old one, producing dangling paths that pass the lexical `file_is_within_root` filter with `stale_index_root: false`.

Judged not a ship blocker for 0.44.0 and disclosed in the PR: reaching it needs a concurrent write of that index's `root_path` between the two reads, it fails loudly when it fires (`mark_reindex_failed`, fatal SSE `error` event, `quarantine.record_failure`), and no data is lost because the runner aborts before the walk and the prune.

Suggested direction, not a design: move the `set_root_path` sync into `service/reindex/runner.rs` immediately after its own gate passes, so one read of `indexes.toml` drives both the decision and the mutation. The handler would keep registering the new `handle.root_path`, which alone fails closed — empty results with `stale_index_root: true`.

Second, related defect from the same review, include it here rather than filing separately: both inputs to the new trust gate degrade to "trusted" on a read failure. `handle.read_indexed_root().await.unwrap_or(None)` turns a redb read error into "no prior root" and skips the gate; `load_index_registry(...).ok().and_then(...)` turns an unparseable `indexes.toml` into "no persisted entry", and `root_move_is_trusted(None, _)` returns `true`. Refusing costs a retry; proceeding costs the corpus.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Reopening — closed prematurely.

[...]

PR #5514, opened 2026-08-11
fix(trusty-search): stop reporting ingest success for an unqueryable contribution (#5505)

### LINKED ISSUE #5505: bug(trusty-search): contrib ingest reports success while a failed graph merge leaves the contribution unqueryable
\`save_then_merge_contrib\` (\`crates/trusty-search/src/core/symbol_graph/contrib.rs:190-196\`) returns the derived-only graph when \`corpus.load_contrib_graphs()\` errors, logging only a warning. \`POST /indexes/{id}/graph\` (\`crates/trusty-search/src/service/server/contrib_graph.rs:144-160\`) installs that graph as the serving graph and still answers \`200\` with \`replaced: true\` (from the separately-checked persist step, which already 500s on its own failure at contrib_graph.rs:126-141) and graph totals that exclude the contribution just ingested. The contribution is durable in redb, so the next successful rebuild restores it — but queries in the interim silently return incomplete results while the caller was told the ingest succeeded.

Two lesser instances of the same fail-open shape in the same file:
- \`contrib.rs:187\` downgrades a KG-persist failure (\`graph.save_to_corpus\`) to a warn.
- \`contrib.rs:221-226\` installs an empty \`SymbolGraph\` if the blocking merge task panics, discarding the derived graph too.

Distinct from #1133 (closed, fixed) — that was an Arc-sharing race in the same function; this is the corpus-load/persist/panic error paths, unrelated to Arc ownership.

Confidence: Inferred — code evidence supports the risk; not exercised by a reproduction or failing test.

Closure conditions:
- [ ] \`POST /indexes/{id}/graph\` distinguishes "merged and queryable" from "persisted but not yet merged" in its response (status or field), rather than always answering plain 200.
- [ ] The KG-persist-failure and panic-installs-empty-graph paths surface the same distinguishable signal instead of a bare warn/error log.
- [ ] A test exercises \`load_contrib_graphs\` failing during \`save_then_merge_contrib\` and asserts the handler does not report an unqualified success.

bobmatnyc:
Fixes the fail-open shape in `POST /indexes/{id}/graph`: the endpoint reported an
ingest as successful while the contribution was not in the serving graph.

## 🔴 Needs a MINOR version bump (0.45.1 → 0.46.0)

This is a **breaking library-API change**. Two public signatures change:

- `save_then_merge_contrib` — `Arc<SymbolGraph>` → `(Option<Arc<SymbolGraph>>, ContribMergeOutcome)`
- `CodeIndexer::rebuild_symbol_graph_now` — `()` → `ContribMergeOutcome`

Both are `pub` and reachable from `lib.rs` (`core::symbol_graph` is re-published).
The version file is owned elsewhere; I did not touch it.
`scripts/check_semver.sh` could not produce a verdict on this machine — it builds
rustdoc with all features, including `cuda`, and `nvcc` is absent (`NO VERDICT`,
exit 3). That is an environment limitation, not evidence either way; the reading
above is from the signatures.

The **HTTP** contract change is narrower: successful ingests are byte-identical
apart from one optional new field. Only the previously-mis-reported failure path
changes status.

## What was wrong

`save_then_merge_contrib` (`crates/trusty-search/src/core/symbol_graph/contrib.rs`)
ended each of its three failure paths by handing back a graph anyway:

| Line (pre-fix) | Failure | Old behaviour |
|---|---|---|
| 190-196 | `load_contrib_graphs` errors | returns the derived-only graph; it is installed as the serving graph |
| 187 | `save_to_corpus` errors | `warn!` only |
| 221-226 | blocking task lost | installs an **empty** `SymbolGraph` |

The handler (`contrib_graph.rs:144-160`) could not see any of it, so it answered
`200` with `replaced: true` and graph totals that excluded the contribution just
ingested. Confirmed by running the new tests against the pre-fix tree:

```
#5505: ingest reported success (replaced=false, graph_nodes=0, graph_edges=0)
while the contribution is absent from the serving graph
```

Severity is a visibility gap, not data loss: the contribution is already durable
in `kg_contrib` via the separately-checked persist step, which 500s on its own
failure.

## The response-contract decision, and why

**`503 contrib_not_merged`, not a `200` with a degraded field.**

A `200` carrying `degraded: true` does not fix a fail-open: every existing client
branches on the status, so the failure stays invisible to exactly the callers the
endpoint misled. The endpoint's promise is "queryable", and it was not kept.

`503` rather than `500` because the state clears itself: the contribution is
stored, so the next successful rebuild merges it. The body carries
`retryable: true` and `persisted: true` — the load-bearing field, since a caller
must not respond by re-deriving data that is not lost. Re-sending the same
document is a safe retry because ingest is replace-per-producer. This matches the
`503` family already documented in `crates/trusty-search/CLAUDE.md`
(`index_corpus_unavailable`, `index_not_resident`, `vector_unavailable`), and is
rendered by the same `degraded.rs` builder so the four cannot drift apart.

**Do not install a graph you know to be wrong.** On a failed contrib load the
pass now installs *nothing* and the previously-served graph keeps serving. The
derived-only graph is known to be missing every contributed edge, so installing
it would un-answer queries the daemon could already answer — advancing state on a
failure path, which is the defect class itself. A lost worker likewise installs
nothing instead of an empty graph.

## The two lesser arms

- **`contrib.rs:187`, KG-persist failure — the warn was *half* right, so it is
  changed, not removed.** Continuing the merge is correct: the in-memory graph is
  complete and only durability is at stake. But it is not nothing — warm boot
  loads the derived graph from `kg_*`, so a swallowed persist failure means the
  next restart serves stale derived edges with no signal anywhere. It now still
  merges and installs, and reports `derived_graph_persist_degraded` on the `200`.
- **`contrib.rs:221-226`, empty graph on panic — genuinely wrong, fixed.** A join
  error (panic *or* runtime cancellation, which is the likelier one during
  shutdown) says nothing about the graph, so it is no basis for discarding a good
  one.

## Tests

Every touched failure branch has a regression test. The fault injectors are real
on-disk faults planted through the store's own handle (the existing #5357
`test_support` pattern), not hand-written `Err` values, so the wiring is covered
and not just the decision:

| Test | Branch |
|---|---|
| `contrib_load_failure_installs_nothing` | contrib load errors → `None` + `merge_error` |
| `contrib_persist_failure_still_merges` | persist errors → merged anyway + `persist_error` |
| `lost_merge_task_installs_nothing_rather_than_an_empty_graph` | real `JoinError` from a panicking `spawn_blocking` |
| `ingest_reports_503_when_the_contributed_overlay_cannot_be_merged` | handler verdict + a query proving the contribution really is unqueryable |
| `ingest_succeeds_but_flags_a_derived_kg_persist_failure` | 200 + `derived_graph_persist_degraded` |
| `ingest_route_answers_503_over_http_when_the_merge_fails` | real router on a loopback socket, real HTTP client, healthy index 200 / broken index 503 |

The two endpoint tests were written first and run against the pre-fix tree: both
failed on the assertion, with the message quoted above. All use
`flavor = "multi_thread"`.

## Test ladder

**Rung 6** — the change alters the HTTP response contract of
`POST /indexes/{id}/graph`. Rung-3 chain on the Rust side plus over-the-wire API
evidence (`ingest_route_answers_503_over_http_when_the_merge_fails` serves the
real `build_router` on `127.0.0.1:0` and drives it with `reqwest`, asserting the
status line and the deserialized body for both the healthy and the degraded
index). No daemon was started: this machine runs a live 82-index daemon and a
second one risks its port lock and data dir.

```
( cargo fmt --check -p trusty-search
  && cargo check -p trusty-search --all-targets
  && cargo clippy -p trusty-search --all-targets -- -D warnings
  && cargo test -p trusty-search ) ; EXIT=0
```

```
test result: ok. 1588 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 387 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
```

Also green: `scripts/check_line_cap.sh` (0 violations), `scripts/check_sld.sh`
(0 errors), `scripts/check_changelog_fragment.sh` (exit 0). No known-flaky test
fired.

Closes #5505


bobmatnyc:
## Critic follow-ups — both fixed in `5f2c7c924`

**HIGH 1 — reindex reported `Ready` off a stale graph.** Confirmed at `completion.rs:75`: the outcome was discarded. `KgRebuildOutcome` now carries `contrib_merge_error`; `rebuild_kg` (`finish.rs`) branches on it and calls the new `mark_graph_failed` instead of `mark_graph_ready`, and both `kg_complete` and `complete` carry `kg_contrib_merge_error`. `mark_graph_failed` fails ONLY the graph stage — unlike `mark_reindex_failed`, which also fails semantic — because the lexical and vector lanes were genuinely rebuilt; failing them would overstate the damage.

**HIGH 2 — `retryable: true` was a promise the daemon could not keep.** `load_contrib_graphs` now keeps the row's producer key and reports it via `ContribRowError`, carried inside the `anyhow::Error` and recovered with `downcast_ref` — so no signature change and no second breaking API surface. `ContribMergeOutcome.blocking_producer` surfaces it and the response sets `retryable` from it:

| blocker | `retryable` | why |
|---|---|---|
| another producer's row | `false` | this ingest never touches that row; retrying fails identically forever. `blocking_producer` names who must re-send. |
| this producer's own row | `true` | a re-send replaces exactly that row |
| no row implicated (table fault, lost worker) | `true` | nothing durable is poisoned; the next attempt can succeed |

The `true` case is proven end-to-end, not asserted: `ingest_503_is_retryable_when_the_blocking_row_is_this_producers_own` plants the corrupt row under `navigatsql`, ingests as `navigatsql`, and gets `200` with `graph_nodes: 3` — the retry the response advertises actually clears the fault.

### Test scope, stated exactly

`reindex_does_not_report_ready_when_the_contrib_merge_fails` drives the SEAM (`rebuild_symbol_graph_for_reindex` + `mark_graph_failed`), not a full `spawn_reindex_awaitable` run. I wrote the end-to-end version first and it failed: with the fault planted in the corpus handed to `set_corpus_store`, the reindex's rebuild does not read it, while that same handle still errors after the run. The reindex's corpus swap appears to hand the rebuild a different store. Pre-fix the seam test does not compile — `KgRebuildOutcome` had no field to carry the failure, which is exactly why `rebuild_kg` could not tell and marked `Ready` unconditionally.

```
test service::reindex::tests::reindex_does_not_report_ready_when_the_contrib_merge_fails ... ok
test service::server::tests_contrib_graph::ingest_503_is_retryable_when_the_blocking_row_is_this_producers_own ... ok
test service::server::tests_contrib_graph::ingest_reports_503_when_the_contributed_overlay_cannot_be_merged ... ok
test service::server::tests_contrib_graph::ingest_route_answers_503_over_http_when_the_merge_fails ... ok
test service::server::tests_contrib_graph::ingest_succeeds_but_flags_a_derived_kg_persist_failure ... ok
test core::symbol_graph::contrib_tests::contrib_load_failure_installs_nothing ... ok
test core::symbol_graph::contrib_tests::contrib_persist_failure_still_merges ... ok
test core::symbol_graph::contrib_tests::lost_merge_task_installs_nothing_rather_than_an_empty_graph ... ok
```

Gates (rung 6, unchanged):

```
( cargo fmt --check -p trusty-search
  && cargo clippy -p trusty-search --all-targets -- -D warnings
  && cargo test -p trusty-search ) ; EXIT=0

test result: ok. 1590 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 387 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
```

Also green: `check_line_cap.sh` (0 violations), `check_sld.sh` (0 errors), `check_changelog_fragment.sh` (exit 0). The `retryable` row in `crates/trusty-search/CLAUDE.md` and the changelog fragment both match the new rule. No new breaking API surface beyond the two signatures already named for 0.46.0.

PR #5515, opened 2026-08-11
fix(trusty-search): DELETE quiesces in-flight work; reap orphaned HNSW staging files (#3049, #2936)

### LINKED ISSUE #2936: trusty-search: clean up orphaned .usearch.tmp on load + add abort-race test against real save()
Two MEDIUM findings from the PR #2930 code-critic review (which shipped the P0 #2922 HNSW-truncation fix):

**(a) Orphaned `.usearch.tmp` cleanup on load/warm-boot**

A shutdown-flush abort landing between `spawn_blocking` completion and the rename poll orphans a fully-written `<hnsw_path>.usearch.tmp` file with no cleanup path. Fix: remove stale `*.tmp` files beside live snapshots during `load_from`/warm-boot. This is safe post-restart because `save_lock` guarantees no live writer can be holding that tmp file at load time.

**(b) No test exercises a real abort against a real `UsearchStore::save()`**

The "aborted flush never reaches its rename" guarantee is currently asserted only by inspection/logic, not exercised by a test. Add an `AbortHandle` mid-flight abort test that aborts a real in-progress `save()` and asserts the live file stays loadable afterward (i.e. the abort never landed a partial/corrupt rename over the live snapshot).

Reference: PR #2930, issue #2922.

### LINKED ISSUE #3049: trusty-search: DELETE /indexes/:id does not await/cancel in-flight reindex or catch-up work (epoch race on recreated ids)
Surfaced during the final review of PR #3024: `unregister_index` removes the index (and now its per-index semaphore, per the #3024 fix) without awaiting or cancelling in-flight reindex/deferred-embed/catch-up tasks. A delete-then-immediate-recreate under the same id lets a new task run concurrently with an old-epoch task, and on-disk paths are keyed purely by sanitized id (persistence.rs::index_data_dir) so both can write the same directory. Delete-during-in-flight was already an unguarded hazard (remove_dir_all racing open writers) — #3024's semaphore eviction only dropped one accidental partial mitigation; documented as an accepted trade-off in remove_index_semaphore's doc comment.

Fix options: have delete await/cancel in-flight work for the id before removing state, or key on-disk paths by an instance epoch.

Refs #2984, PR #3024.

bobmatnyc:
Both defects were re-verified against current source before any code was written; both were live.

## #3049 — DELETE neither awaited nor cancelled in-flight work

`unregister_index` (`service/server/search.rs:114-152` pre-fix) removed the registration and `remove_dir_all`'d the data directory with no regard for in-flight writers. The comment at lines 137-143 said evicting the semaphore was "safe even if a task is still mid-flight on it", which is true of the eviction and says nothing about the `remove_dir_all` two lines below. `grep -c 'CancellationToken|cancel_flag'` over the crate returned 0.

The per-index semaphore is the existing quiescence point — `runner::run_reindex:85`, `defer_embed_queue:359`, and `index_config.rs:225` all hold it for the full duration of their work. The delete now signals a per-index cancel flag, waits up to 30s on that permit, and holds it across teardown. `run_reindex` polls the flag at its producer and consumer batch boundaries, so the wait costs one batch instead of one corpus. Neither caller of `unregister_index` holds the permit, so there is no deadlock.

### Round 2 behavior (now superseded)
On timeout the registration was torn down while data removal was refused; the response's `quiesced: false` told the caller data removal did not happen. This half-delete caused the orphaned-directory leak: retries hit `removed=false` and never reached the data-removal branch again.

### Round 3 behavior (now superseded)
On timeout with `?delete_data=true`, the delete **abandons entirely** — it reports `removed: false`, `data_deleted: false`, `quiesced: false`, and changes nothing. No registration teardown, no data removal, no leakage of a half-deleted state. The cancel flag is cleared via `clear_index_cancel` on the abandoned path, so writers already holding the `Arc` observe the reset.

When `?delete_data=false` is requested and a timeout occurs, the registration IS torn down — no orphan is possible because no data removal was planned anyway, and this is the mode the orphan-reaper ticker runs in.

🔴 **Behavior change: wedged writers.** An index with a permanently wedged writer (one that never releases the permit) can no longer be deleted with `?delete_data=true` at all — it returns `removed: false` every time. Previously it half-succeeded and leaked. The trade-off: escape hatches are `?delete_data=false` (deregisters, preserves data for manual cleanup) or stopping the daemon and editing `indexes.toml`, which issue #4123 documents.

**Residual, not papered over:** `embed_deferred_chunks` is a single long call with no interior checkpoint. A delete landing during one waits for that call to finish rather than for a batch boundary, and hits the timeout if it outlasts 30s — at which point the abandonment above applies.

### Round 4 behavior (current)

🔴 **New mechanical gate: `scripts/check_teardown_guard.sh`** enumerates every call site of 11 durable-write methods and requires a guard acquired earlier in the same scope, where a spawned task body counts as its own scope. It exists because manual enumeration missed a path in three consecutive rounds — round 4 found a fourth: `watch_loop::handle_modified` at line 292 acquired the guard while its stale-chunk loop at line 280 already called `remove_chunk` → `delete_chunks_from_redb`. The guard was real and sat after the write; `remove_chunk` was never on the hand-written seed list. The gate is wired in `.github/workflows/teardown-guard.yml`, runs BEFORE the main test gate, and carries 8 mutation cases including `late_guard` and `spawn_escape` which reproduce the two real misses from prior rounds. Seed list is hand-maintained — the gate closes "existing method, new call site" drift but not "someone adds a brand-new durable-write method". Documented in the script header.

On `?delete_data=false` with a quiesce timeout, eviction of `INDEX_TEARDOWN_LOCKS` and `INDEX_LOCKS` is now **conditional on `quiesced`**. Round 3 evicted them unconditionally, which handed the next operation on that id a fresh, uncontended lock disconnected from a writer still running — so a later `?delete_data=true` reported `quiesced: true` and could destroy the directory that writer was using. The cancel flag is still evicted on both paths, deliberately — reaching that point means deregistration went through either way, so the flag's `true` must keep reaching the in-flight writer, which holds its own `Arc` that eviction does not touch. Eviction of the cancel flag adds that a recreate under the same id starts uncancelled.

The startup schema migration (`commands/start/daemon.rs` spawning `core::migration::run_migrations` per index at every daemon boot) **now holds the teardown guard for its full run**. It previously held no guard, so a `DELETE ?delete_data=true` during a migration saw an uncontended lock, reported `quiesced: true`, and removed the data directory while `commit_parsed_batch` was writing into it. A delete racing a boot-time migration now waits or refuses.

### Fail-open fix (the confirmed one)
`data_deleted` was computed as `removed && params.delete_data` — from the request — while a `remove_index_data_dir` failure was downgraded to `tracing::warn!`. A delete whose data removal failed reported `data_deleted: true`, so a caller recorded the corpus as reclaimed while every byte remained. It is now assigned only when the removal returns `Ok`. `unregister_index` returns `UnregisterOutcome { removed, data_deleted, quiesced }` instead of a bare `bool`.

## #2936 — orphaned staging files, and the untested abort race

**(a)** Confirmed worse than when filed. `staging_path`'s own doc (`usearch_store.rs:43-47`) states the residual: #4395 made staging names pid-scoped, so a process SIGKILLed between the write and the rename leaves a file no later process overwrites. The existing `let _ = remove_file(&tmp_hnsw)` cleanups cover only `Err` paths. No `read_dir` existed anywhere in `core/store/`, so nothing could reap them.

`load_from` now reaps first: dead-pid staging files, plus pre-#4395 bare `.tmp` names unconditionally. A **live**-pid file is left alone — colocated snapshots live outside every data directory, so two daemons can legitimately stage beside the same snapshot, and deleting one mid-write is the corruption #4395 removed. Such a file is consumed by its own process's next successful save, since `staging_path` is deterministic per process; `a_completed_save_consumes_this_processs_own_abandoned_staging_file` pins that so the live-pid exemption does not read as an unhandled leak.

`service/daemon.rs`'s `pid_alive` is reused, not reimplemented. It had to become `pub` rather than `pub(crate)`: `commands::stop` lives in the binary crate and reaches it through the library's public `service` module. That is also an adjacent fix — `commands/stop.rs` carried a byte-identical private copy, which the new caller would have made a third. Additive only; no existing signature changed.

**(b)** A real `AbortHandle` abort against a real in-progress `save()`, swept across seven delays. The sweep genuinely straddles the rename — instrumented, delays 0/50/500µs landed before it (old snapshot preserved, 600 vectors) and 200µs/1/3/8ms after (900) — and any torn state fails. Both outcomes are valid, so the assertion is timing-independent even though which branch runs is not.

## Test ladder — rung 5

Index delete, cancellation, process lifecycle, and teardown-guard enforcement. Every gate below was run on this branch.

```
bash scripts/check_teardown_guard.sh                  EXIT=0  (284 files, 23 sites, 16 guards)
bash scripts/check_teardown_guard_selftest.sh         EXIT=0  (8 mutations detected)
cargo check -p trusty-search --all-targets            EXIT=0
cargo clippy -p trusty-search --all-targets -D warnings  EXIT=0
cargo check --workspace                               EXIT=0
bash scripts/check_line_cap.sh                        EXIT=0  (3914 files, 0 violations)
bash scripts/check_sld.sh                             EXIT=0  (0 errors, 0 warnings)
bash scripts/check_test_pointers.sh                   EXIT=0
cargo test -p trusty-search                           EXIT=0
```

```
test result: ok. 1591 passed; 0 failed; 1 ignored
test result: ok. 387 passed; 0 failed; 3 ignored
(+ 15 integration binaries, all ok, 0 failed)
```

No baseline reds fired — neither the `trusty-search` filesystem-watcher tests nor `execute_doctor_against_test_daemon`.

### Pre-fix proof

Each new test was run against the pre-fix behaviour (production logic reverted, type shape kept so the file still compiles) and failed as a real assertion failure, not an unrelated panic:

```
unregister_index_waits_for_an_in_flight_writer_to_release_the_permit ... FAILED
  unregister_index returned while a writer still held the per-index permit —
  the delete did not await in-flight work (issue #3049)

delete_reports_data_deleted_false_when_the_removal_fails ... FAILED
  a FAILED data removal must report data_deleted=false, not echo the request flag

delete_with_delete_data_refuses_removal_when_a_writer_never_quiesces ... FAILED
  a permit held for the whole call must expire the quiesce wait

index_cancel_flag_is_evicted_so_a_recreated_index_starts_uncancelled ... FAILED
  unregister_index must SIGNAL the cancel flag so in-flight writers stop

reap_removes_dead_pid_and_bare_staging_files ... FAILED
  load_from must reap the abandoned staging file …/hnsw.usearch.2000000000.tmp
```

The error-arm test for a failed removal plants a regular **file** where `remove_index_data_dir` expects the directory, so `remove_dir_all` returns `NotADirectory` — deterministic, with no permission games and no root/non-root distinction.

## Writer-guard table corrections

Round 3 reclassified the writer-guard table after finding round 2's table claimed `PATCH /indexes/:id/config` took the read side when it never did. Round 4 discovered `watch_loop::handle_modified` did not hold a guard where it called `remove_chunk`, and added it to the enforcement gate. The corrected table now has **12 rows** after adding `watch_loop::handle_modified`, and carries 14 previously-unclassified sites now individually classified — including three `set_root_path` calls marked EXEMPT with the reason stated rather than sitting silently under a row implying they were guarded. `create_index_handler` remains in the guarded list (it acquired the write permit but round 2 had it under "deliberately NOT guarded"). No functional change to round 3 code — the guards were already there; the table is now accurate and complete.

## Notes

- Two pre-existing red gates were fixed: the `2936` changelog fragment carried two categories and failed the assembler (split into `2936-pid-alive-consolidation-changed.md`), and two round-3 doc pointers cited tests that were never written, now repointed at the tests that cover the behavior.
- `PR version bump vs crates.io (#4421)` will be red until the `0.45.0` → `0.45.1` bump on #5506 reaches main. Rebasing this branch afterwards clears it; no bump was made here, to avoid colliding with that one.
- No collision with the in-flight flake work: `usearch_store.rs` gained only a 5-line call at the top of `load_from`, and the reaper lives in its own new file.

Closes #3049
Closes #2936


bobmatnyc:
## Round 2 — CRITICAL fixed, two HIGHs deferred with reasons

Head: `fae927c14`

### Exhaustive writer enumeration

Derived by grepping every mutating method (`index_file`, `remove_file`, `commit_parsed_batch`, `save_contrib_graph`, `rebuild_symbol_graph_now`, `embed_deferred_chunks`, `set_root_path`, `clear_*`) across `src/service`. The critic named 2 ungated paths; there are **6**. That is the convergence failure the instruction warned about, confirmed empirically.

| Path | Entry point | Held permit before | Now |
|---|---|---|---|
| `POST /indexes/:id/index-file` | `server::files::index_file_handler` | none | read side |
| `POST /indexes/:id/remove-file` | `server::files::remove_file_handler` | none | read side |
| `POST /indexes/:id/graph` | `server::contrib_graph::ingest_graph_handler` | none | read side |
| `PATCH /indexes/:id` (relocate) | `server::indexes_relocate::relocate_index_handler` | none | read side |
| filesystem watcher | `service::watch_loop` | none | read side |
| boot reconcile | `service::reconcile` | none | read side |
| full reindex | `reindex::runner::run_reindex` | `index_semaphore` | read side + semaphore |
| deferred embed | `reindex::defer_embed_queue` | `index_semaphore` | read side + semaphore |
| `PATCH /indexes/:id/config` | `server::index_config` + `components` | `index_semaphore` | read side + semaphore |

Deliberately unguarded, with reasons: read-only handlers (search/grep/typeahead/status/chunks/call_chain/graph GETs) never mutate; `tickers` idle-eviction and `shutdown_flush` touch only in-memory caches rebuilt from the durable corpus; `create_index_handler` registers a new handle rather than writing an existing one (that is the recreate-window HIGH, below).

The table lives on `INDEX_TEARDOWN_LOCKS` in `service/reindex/semaphore.rs`, not only here, so the next reviewer checks a list in the source and a new write path has an obvious place to be added.

### CRITICAL — fixed, by replacing the primitive

`index_semaphore` is a 1-permit **mutual-exclusion** lock. Gating the six ungated paths on it would have made one `index-file` call block for the entire duration of a reindex — a serialisation regression on the crate's documented incremental-indexing path for network mounts. So the quiescence point is now a separate per-index **teardown RwLock**: writers take the shared side (as concurrent with each other as before), `unregister_index` takes the exclusive side. Tokio's RwLock is write-preferring, so a pending delete stops admitting new readers rather than starving behind a stream of short writes. `index_semaphore` keeps its unchanged mutual-exclusion job.

### MEDIUM — accepted and documented

`is_reapable_staging` keeps bare `kill(pid, 0)`. A recycled pid leaks one staging file, which is the pre-#2936 status quo for that file rather than a new failure mode; the alternative is per-platform `/proc/<pid>/stat` / `proc_pidinfo` parsing to protect a best-effort tidy-up, and the same unguarded call already decides stale-lockfile removal where the consequence is larger. The residual is now written at the call site.

### Two HIGHs NOT fixed — deferred deliberately

Landing the CRITICAL completely rather than half-landing three concurrency fixes, per your instruction. Both are real; neither is mitigated by this round.

**Orphan directory leak on timeout.** The recovery instruction in the log is still false — a re-issued DELETE finds `removed=false` and never reaches `remove_index_data_dir`. Fixing it properly means either a persisted "delete pending" record or an orphan-reaper pass that detects data directories with no `indexes.toml` entry, plus a second-delete-attempt test. Note this round makes the timeout *less* likely to be reached (six more paths now release promptly), but the deferred-embed pass still has no interior checkpoint, so the leak path remains live.

**Recreate-during-teardown window.** `remove_and_get` still runs at the start while eviction runs at the end. Unchanged by this round.

### Gates — rung 5

```
cargo clippy -p trusty-search --all-targets -D warnings  EXIT=0
cargo check --workspace                                  EXIT=0
bash scripts/check_line_cap.sh                           EXIT=0  (3917 files, 0 violations)
cargo test -p trusty-search                              EXIT=0
```

```
test result: ok. 1592 passed; 0 failed; 1 ignored
test result: ok. 387 passed; 0 failed; 3 ignored
(+ integration binaries, all ok, 0 failed)
```

No baseline reds fired — notably the watcher tests pass despite `spawn_watch_loop` gaining an `IndexId` parameter.

### Pre-fix proof for the new test

`delete_waits_for_an_ungated_index_file_write`, run against round-1 behaviour (quiescence reverted to `index_semaphore`, type shape kept):

```
thread '...::delete_waits_for_an_ungated_index_file_write' panicked at
crates/trusty-search/src/service/server/tests_3049.rs:290:5:
DELETE returned while an index-file write was still in flight — the quiescence
point does not cover writers that hold no reindex permit (issue #3049)
```

The four round-1 tests still pass; two were repointed from the semaphore to the teardown lock, since the semaphore is no longer the quiescence point.

bobmatnyc:
**code-critic verdict: BLOCK** (round 3, head `804034aa7`). Three required changes; a fix round is in flight.

**CRITICAL — the startup schema migration is unguarded.** `commands/start/daemon.rs:385-394` spawns `core::migration::run_migrations` per registered index at every daemon boot when `TRUSTY_DISABLE_MIGRATIONS` is unset, and never acquires `acquire_index_teardown_read`. `M001PerPubConstRust::apply` (`core/migration/m001.rs:172-198`) calls `commit_parsed_batch` in a loop over 64-file batches — the exact method this PR's own doc comment (`semaphore.rs:213-216`) names as a grep term used to build the writer table. A `DELETE /indexes/:id?delete_data=true` during a still-unmigrated index's migration sees the teardown lock uncontended, reports `quiesced: true`, and `remove_dir_all`s the directory while `commit_parsed_batch` is writing into it. Deterministically reproducible: restart with any pre-v0.11.1 index present, then delete it with `delete_data=true`.

**HIGH — eviction on the unquiesced `delete_data=false` path.** `service/server/search.rs:333,336,340` evict `INDEX_LOCKS` / `INDEX_CANCEL_FLAGS` / `INDEX_TEARDOWN_LOCKS` unconditionally, including on the `!quiesced && !delete_data` branch at line 250 where the exclusive guard was never acquired. `delete_data` defaults to false, so this is the default endpoint behavior. A writer that has not quiesced within 30s keeps its own valid `Arc`, but the next operation on that id gets a fresh, disconnected lock — and a later `delete_data=true` then reports `quiesced: true` and destroys the directory that writer is still using. The existing parked-writer test misses it: its gen-1 writer releases before the wait expires, so `first.quiesced` is true and the branch is never entered.

**LOW — `Arc::ptr_eq` retry loop has no cap or backoff** (`semaphore.rs:279-308`). Traced as terminating in the designed case, since eviction happens only within the exclusive guard's scope; unbounded spinning would require a full teardown cycle with several real `.await` points to complete inside a few non-yielding instructions. Not reachable under real I/O, but nothing in the code proves a bound. Being documented rather than changed.

Round 4 additionally owes a mechanical check for the writer-guard table. Manual enumeration has now missed a path in three consecutive rounds — 2 in round 1, 1 in round 2, 1 in round 3 while claiming exhaustive — so the table needs a gate that fails when a corpus-mutating call lacks the guard, with a floor assertion so it cannot pass by inspecting nothing.

Scope note: the reviewer read control flow rather than running the suite, and said so. Both actionable findings are read directly off `unregister_index`'s and `run_migrations`'s control flow and do not depend on test behavior.

Issue #5440, opened 2026-08-11
bug(trusty-common): rustdoc build fails under rustc 1.97.1, root cause unidentified


bobmatnyc:
Split off from [#5297](https://github.com/bobmatnyc/trusty-tools/issues/5297), which bundled two unrelated problems. This is part (a): the rustdoc build failure itself. Part (b) — whether an already-major-bump crate should still get a diff for the record — stays on [#5297](https://github.com/bobmatnyc/trusty-tools/issues/5297).

`trusty-common` fails to build rustdoc under rustc 1.97.1 (the version the CI/publish runner selects, ignoring `Cargo.lock`'s pinned 1.94.1), which is why `preflight-publish.sh` CHECK 5 exits 3 with no SemVer verdict — observed twice, both cited in [#5297](https://github.com/bobmatnyc/trusty-tools/issues/5297) (trusty-common 0.30.0, then trusty-search 0.44.0 release prep).

Root cause is unidentified. This is not the same defect as [#5289](https://github.com/bobmatnyc/trusty-tools/issues/5289) (closed by [#5290](https://github.com/bobmatnyc/trusty-tools/pull/5290)) — that was a `takecell` dependency needing rustc 1.96 against the pinned 1.94.1 MSRV toolchain; 1.97.1 clears that floor, so something else is breaking rustdoc for this crate specifically.

Ruled out so far: a research pass over `crates/trusty-common/src/lib.rs` found no `doc(cfg`, no `docsrs`, and no other doc-gating attribute that would explain a version-specific rustdoc failure. The actual rustc 1.97.1 error text has not yet been captured — reproducing the failure and capturing that output is the first step for whoever picks this up.

Closure conditions:
- The rustc 1.97.1 rustdoc failure is reproduced locally and its error output captured
- Root cause identified and either fixed or tracked with a linked follow-up

Confidence: reproduced (the CHECK 5 exit-3/no-verdict behavior, twice), but the underlying rustdoc error itself is not yet captured — root cause is Speculative pending that capture.

bobmatnyc:
**Toolchain probe: refuted.** Pinning `rustc 1.94.1` via `SEMVER_GATE_TOOLCHAIN_BIN` does NOT reproduce a clean rustdoc build with a real verdict — it exits 0, but the checked-count line shows why that's not a pass:

```
toolchain: rustc 1.94.1 (pinned by SEMVER_GATE_TOOLCHAIN_BIN)
[...]

Issue #5296, opened 2026-08-10
trusty-mpm: tag-triggered SemVer workflow's baseline can already be the version under test


bobmatnyc:
\`.github/workflows/semver-checks.yml\` fires on tag push; \`scripts/check_semver.sh\`'s \`registry_latest\` pulls the crate's latest non-yanked crates.io release as baseline. The workflow's own comments and \`cargo-publish\` SKILL.md (steps 4-6) assume the tag push precedes \`cargo publish\` — but by the time the CI job actually executes and queries the registry, the publish has already landed. Baseline becomes the version under test.

Observed on three crates during the 1.3.5 cut: trusty-agents-common 0.5.0, trusty-progress 0.3.0, tga 2.12.0. Each self-compared and reported "no change; assume minor" — no real verdict.

Distinct from #5289/#5290: that fix made the gate refuse to report a build failure as a verdict. This is different — \`cargo-semver-checks\` genuinely runs and prints its "Checked N checks" line here, so #5290's positive-evidence guard passes it through. The comparison just runs against the wrong baseline.

ASK: either select the previous release as baseline instead of latest, or move the trigger to fire before the publish actually happens (not just before it in the documented step sequence). \`scripts/preflight-publish.sh\` CHECK 5 is what actually gates an upload, so this workflow is advisory — but an advisory check that structurally cannot produce a verdict reads as coverage while providing none.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

PR #5522, opened 2026-08-11
fix(ci): SemVer gate must not read zero executed checks as a pass (#5440)

bobmatnyc:
Fixes the false-pass path in the SemVer gate reported on #5440, and answers two
of that issue's three open questions. Part (a) — the false pass — is fixed here.
The rustdoc crash under rustc 1.97.1 is **not** fixed and stays on #5440.

## What was wrong

`cargo-semver-checks` skips its entire lint set when the baseline → current
delta already permits breakage, and reports that at exit 0:

```
Checking trusty-common v0.31.0 -> v0.30.1 (major change)
 Checked [   0.000s] 0 checks: 0 pass, 254 skip
```

`scripts/check_semver.sh`'s `verdict_computed` keyed on the summary line
*existing*, so it read "254 lints declined" as "254 lints satisfied" and exited
0. `scripts/preflight-publish.sh` CHECK 5 then passed having verified nothing.

The gate failed closed on the reporting machine only by accident: under the
newest installed toolchain the same crate crashed rustdoc, printed no summary,
and was correctly classified NO VERDICT. `check_semver.sh` auto-selects the
highest installed rustc, so removing that toolchain would have turned the gate
green for every publish.

## The change

`scripts/check_semver.sh:771-799` — `verdict_computed` now parses the leading
`N checks:` count (the number of lints that RAN; the trailing `N skip` is what
was declined) and requires it to be at least 1. Zero, or a count that does not
parse, is NO VERDICT → exit 3, the same status a rustdoc failure already gets.
A new `no_verdict_because` gives the two causes different messages, because
"it never ran" and "it ran and skipped every lint" have different remedies.

## Regression tests

`scripts/check_semver_selftest.sh` (runs in CI via `semver-checks.yml`) gains
cases 20, 20b and 21. The fixture is the self-test's own former `clean.out`,
renamed to `all-skipped.out`: the case that was supposed to prove the gate can
pass a crate was itself being satisfied by a run that checked nothing.
`clean.out` is now a real 196-pass capture, so "a genuine pass still exits 0"
stays covered.

Against the **pre-fix** gate (`git show HEAD~1:scripts/check_semver.sh` in
place, new fixtures and cases):

```
SELF-TEST FAIL: verdict/zero checks executed (case 20): expected exit 3, got 0
       CHECK trusty-agents: 0.38.5 -> 0.38.6 (minor release), 0 feature(s)
            Checked [   0.000s] 0 checks: 0 pass, 254 skip
       semver gate: scanned (explicit); 1 crate(s) checked, 0 skipped — OK.
SELF-TEST FAIL: zero-checks/diagnosis: the refusal did not say the run executed no checks
SELF-TEST FAIL: inventory/zero-checks: a run that executed 0 of 254 checks was reported as an empty inventory (#5440)
       INVENTORY trusty-agents: no breaking changes found against v0.37.0.

check_semver_selftest: 19 passed, 3 FAILED.
```

Post-fix: `check_semver_selftest: 22 passed, 0 failed.`

## Part B findings

Measured with `cargo semver-checks -p trusty-progress --baseline-version 0.2.0
--only-explicit-features`, same crate, three flag settings:

| `--release-type` | tool's own reading | checks executed |
|---|---|---|
| *(omitted)* | `(major change)` | **0** — `0 checks: 0 pass, 254 skip` |
| `minor` | `(assume minor change)` | 196 — `196 checks: 196 pass, 58 skip` |
| `patch` | `(assume patch change)` | 223 — `223 checks: 223 pass, 31 skip` |

So the mass skip is not a rustdoc problem and not toolchain-specific: it is the
tool declining to lint a delta that already permits breakage. The INVENTORY arm
already passes `--release-type minor` and is unaffected — that flag works as
its comment claims.

The PASS/FAIL arm passes no `--release-type`, so the tool re-derives it from the
version pair. The gate reaches that arm only when its own `release_type()` says
`patch`, `minor`, or `none` — and `none` (current ≤ baseline) is the one value
that can disagree with the tool, which reads the same pair by position and calls
it major. That is exactly the recorded run: `1 crate(s) checked, 0 skipped — OK`
is the PASS/FAIL arm's summary, and only `none` routes a major-shaped delta
there.

On the baseline question: selection on current `main` is correct, measured —

```
$ bash scripts/check_semver.sh --probe trusty-common
probe trusty-common: baseline=0.32.0 (declared=0.33.0, latest on crates.io=0.33.0, pre-release below=NONE)
$ bash scripts/check_semver.sh --probe trusty-common --probe-version 0.30.1
probe trusty-common: baseline=0.30.0 (declared=0.30.1, latest on crates.io=0.33.0, pre-release below=NONE)
```

The `#5296` ceiling makes baseline < declared by construction, so the recorded
inversion (baseline 0.31.0 above current 0.30.1) cannot be reproduced through
today's selection code. It is the shape a pre-#5296 gate produced (baseline =
crates.io *latest*), or a tree whose declared version sat at or below its
registry baseline. No `none` branch is added here: it is unreachable through
`registry_probe`, and this PR's fix already converts every such run into a
correct exit-3 block rather than a green one. Left as a note on #5440.

## Gates (rung 5 — release tooling; scripts and docs only, no crate `src/**`)

```
bash scripts/check_semver_selftest.sh     → check_semver_selftest: 22 passed, 0 failed.   (exit 0)
bash scripts/check-ci-helpers-selftest.sh → check-ci-helpers-selftest: all 135 cases passed (exit 0)
shellcheck scripts/check_semver.sh scripts/check_semver_selftest.sh → exit 0, no findings
bash scripts/check_semver.sh              → scanned 7 changed path(s); no crate source changed — OK. (exit 0)
bash scripts/check_changelog_fragment.sh  → no crate source changed — OK. (exit 0)
bash scripts/check_sld.sh                 → exit 0
bash scripts/check_doc_numbers.sh         → exit 0
bash scripts/check_public_docs.sh         → exit 0
bash scripts/check_line_cap.sh            → exit 0
```

No changelog fragment: no crate `src/**` is touched. No workflow files changed,
so no `actionlint` run. `shellcheck scripts/preflight-publish.sh` reports 8
pre-existing SC2001/SC2181 style findings on lines this PR does not touch (the
only edit there is a comment).

### DIFF
--- a/docs/reference/semver-gate.md
+++ b/docs/reference/semver-gate.md
@@ -159,12 +159,36 @@ checks: …`). No summary line means no comparison happened — whatever the exi
 status was, *including exit 0* — and that is reported as `NO VERDICT` on its own
 exit status.
 
+### Zero checks executed is not a pass ([#5440](https://github.com/bobmatnyc/trusty-tools/issues/5440))
+
+The summary line alone was not enough. `cargo-semver-checks` skips its whole lint
+set when the baseline → current delta already permits breakage, and prints a
+summary saying so at exit 0:
+
+```
+Checking trusty-common v0.31.0 -> v0.30.1 (major change)
+ Checked [   0.000s] 0 checks: 0 pass, 254 skip
+ Summary no semver update required
+```
+
+The leading `0 checks:` is the number of lints that **ran**; the trailing
+`254 skip` is what the tool declined to run. Keying on the line's presence read
+that as a pass, so `preflight-publish.sh` CHECK 5 approved the publish having
+verified nothing. Measured on `trusty-common` under rustc 1.94.1; the gate failed
+closed on that machine only because a newer installed toolchain crashed rustdoc
+instead, which produced no summary at all.
+
+The gate now parses the count and requires it to be at least 1. A summary whose
+count is zero, or does not parse, is `NO VERDICT` (exit 3) and says which of the
+two happened — "executed no checks" and "never completed a check run" have
+different remedies.
+
 | Exit | Meaning |
 |---|---|
 | 0 | every checked crate is clean, or is a recorded skip |
 | 1 | a verdict was computed **and it says break** — the only status that means "the API changed" |
 | 2 | usage error |
-| 3 | **no verdict** — rustdoc build failure, unreachable registry, missing tool, or a diff that scanned nothing. Nothing was compared, so nothing may be concluded. |
+| 3 | **no verdict** — rustdoc build failure, a run that executed zero checks, unreachable registry, missing tool, or a diff that scanned nothing. Nothing was compared, so nothing may be concluded. |
 
 `scripts/preflight-publish.sh` CHECK 5 and `.github/workflows/semver-checks.yml`
 both report exit 3 separately from exit 1. Both still stop the publish: a
@@ -313,6 +337,14 @@ never reach the version-bump remediation; an exit-0 run that compared nothing
 must still fail; and a genuinely clean run must exit 0, which is what proves the
 other three fail on classification rather than because the gate is broken
 outright.
+
+Cases 20-21 (#5440) pin the other way a completed run says nothing: a summary
+reporting `0 checks: 0 pass, 254 skip` at exit 0 must be `NO VERDICT` in the
+pass/fail arm and a blind inventory in the advisory one, never "no breaking
+changes found". Both fail against the pre-fix gate, which exits 0 on the first
+and reports an empty inventory on the second. Their fixture, `all-skipped.out`,
+is the former `clean.out` — the case that was supposed to prove the gate can pass
+a crate was itself being satisfied by a run that checked nothing.
 
 Those four replace only the `cargo semver-checks` subprocess, via a stub `cargo`
 on `PATH` that forwards everything else to the real one — so crate resolution,
--- a/scripts/check_semver.sh
+++ b/scripts/check_semver.sh
@@ -95,7 +95,10 @@
 #   requires a matching version bump" remediation below, so a rustdoc error was
 #   presented as a SemVer verdict and the two were indistinguishable from the
 #   output. The gate now classifies on POSITIVE EVIDENCE instead: a verdict
-#   exists only when the tool printed its own per-crate `N checks:` summary line.
+#   exists only when the tool printed its own per-crate `N checks:` summary line
+#   AND that line says at least one check RAN (issue #5440 — a summary reporting
+#   `0 checks: 0 pass, 254 skip` is the tool declining to compare, and the gate
+#   used to read it as a pass).
 #   No summary line means no comparison happened, whatever the exit status was —
 #   including exit 0 — and that is reported as NO VERDICT on its own exit code.
 #   Absence of evidence is never a pass; same rule as the scan floor below.
@@ -164,7 +167,9 @@
 #   declared 1.0.1, which selected 1.0.0-rc1 before the rank element existed.
 #   Cases 16-18 (#5500) replay PR #5458's own CI bytes, where forced colour hid the
 #   summary line and both a clean run and a four-failure run were reported as
-#   never having happened.
+#   never having happened. Cases 20-21 (#5440) pin the zero-check false pass:
+#   a summary that says `0 checks: 0 pass, 254 skip` at exit 0 is NO VERDICT in
+#   the PASS/FAIL arm and a BLIND inventory in the advisory one.
 #   The catch itself is demonstrated in PR #5051 against #4088's real shape.
 #
 # Portability: bash 3.2 (macOS system bash) and bash 5 (Linux CI). POSIX tools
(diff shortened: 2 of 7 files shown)


bobmatnyc:
## code-critic: APPROVE

Reviewed against the spec (issue #5440) and the code only. Every claim below was
re-derived by running the gate; no pasted output from the PR body was accepted.

### The 0.x-minor-blocks-every-publish concern: refuted, measured

A routine `0.32.0 -> 0.33.0` bump does **not** reach the arm this PR hardened.
`release_type()` (`scripts/check_semver.sh:402-423`) calls a 0.x MINOR-position
change `major`, and `:993` routes `major` to the INVENTORY arm, which passes
`--release-type minor` (`:1000`) so the tool actually lints. That arm never
touches `fail` or `noverdict`, so it cannot change the exit code.

Instrumented through the self-test's stub registry (`trusty-agents`, the same
0.x shape, baseline `0.37.0` -> current `0.38.6`):

```
=== EXP A: 0.x MINOR bump (0.37.0 -> 0.38.6), real 196-pass capture ===
EXP A exit=0
INVENTORY trusty-agents: 0.37.0 -> 0.38.6 is already a breaking release — ... advisory
     Checked [   0.011s] 196 checks: 196 pass, 58 skip
INVENTORY trusty-agents: no breaking changes found against v0.37.0.
semver gate: scanned (explicit); 0 crate(s) checked, 0 skipped, 1 inventoried (advisory) — OK.

=== EXP B: same bump, degenerate all-skipped capture ===
EXP B exit=0
NO INVENTORY trusty-agents: cargo semver-checks exited 0 and completed a run that
EXECUTED NO CHECKS — 'Checked [   0.000s] 0 checks: 0 pass, 254 skip'.
semver gate: scanned (explicit); 0 crate(s) checked, 0 skipped, 1 inventory NOT computed — OK.

=== EXP C: 0.x patch-position bump (0.38.5 -> 0.38.6), PASS/FAIL arm ===
EXP C exit=0
CHECK trusty-agents: 0.38.5 -> 0.38.6 (minor release), 0 feature(s)
     Checked [   0.011s] 196 checks: 196 pass, 58 skip
semver gate: scanned (explicit); 1 crate(s) checked, 0 skipped — OK.
```

**Is "0 checks because the delta permits breakage" conflated with "0 checks
because the tool failed"?** In the PASS/FAIL arm, only where the gate and the
tool disagree about the delta — and for every version shape in this workspace
they agree. `patch`/`minor` reach that arm only when the tool also reads the pair
as non-major, so it lints (223 and 196 checks respectively). The one disagreeing
value is `none` (current <= baseline), and `registry_probe` selects a baseline
strictly below the declared version (`key(v) >= ceiling: continue`), so it is
unreachable — the author's stated reason holds, and I confirmed no env override
supplies a baseline directly (`SEMVER_GATE_BASE` is the diff base,
`SEMVER_GATE_INDEX_BASE` the registry URL). The recorded `0.31.0 -> 0.30.1`
inversion on #5440 is exactly that `none` shape from the pre-#5296 gate, and
exit 3 is the right answer for it. No `--release-type` needs adding to the
PASS/FAIL arm.

### Regression tests genuinely fail pre-fix

Re-run in an independent clone, substituting `git show HEAD~1:scripts/check_semver.sh`:

```
SELF-TEST FAIL: verdict/zero checks executed (case 20): expected exit 3, got 0
SELF-TEST FAIL: zero-checks/diagnosis: the refusal did not say the run executed no checks
SELF-TEST FAIL: inventory/zero-checks: a run that executed 0 of 254 checks was reported as an empty inventory (#5440)
check_semver_selftest: 19 passed, 3 FAILED.     (exit 1)
```

Post-fix, unmodified: `check_semver_selftest: 22 passed, 0 failed.` (exit 0).
`shellcheck scripts/check_semver.sh scripts/check_semver_selftest.sh` → exit 0,
no findings.

### Count parsing fails closed

Exercised `verdict_computed` / `no_verdict_because` directly:

| input | result |
|---|---|
| `0 checks: 0 pass, 254 skip` | NO — EXECUTED NO CHECKS |
| ANSI-coloured zero-check line | NO — EXECUTED NO CHECKS (#5496 mode covered: the strip runs before the count is read) |
| ANSI split mid-digit (`19`/`6`) | YES, 196 |
| format changed (`checks=196`) | NO — never completed a check run |
| no summary / empty file | NO — never completed a check run |
| `1,234 checks:` | YES (parses 234; harmless — never a false pass) |

`Closes #5440` is **not** present: `closingIssuesReferences` is empty, so the
rustdoc-crash half stays tracked. The renamed fixtures check out — the new
`clean.out` is a real `196 checks: 196 pass, 58 skip` capture, and
`all-skipped.out` is byte-identical to the old degenerate `clean.out`.

## Findings

|---|---|---|---|---|---|
| MEDIUM | scripts/check_semver.sh | 415-416 | A `0.0.z` crate is the one shape where the gate and the tool disagree in the direction this PR now blocks. `release_type("0.0.5","0.0.6")` returns `minor`, so it lands in the PASS/FAIL arm; Cargo makes every `0.0.z` bump breaking, so the tool reads major, runs 0 of 254 checks, and the gate exits 3 — a publish blocked with a NO VERDICT that nothing can clear. Latent only: the lowest crate in the workspace is `0.1.0`, so no crate hits it today. | In `release_type`, return `major` when `b[0] == 0 and b[1] == 0`, which routes those crates to the advisory INVENTORY arm exactly like every other permitted break. | Fix here |
| MEDIUM | scripts/check_semver.sh | 1085-1091 | The leading NO VERDICT remediation bullet names a cause that cannot reach that arm — "ANY change to the MINOR position of a 0.y.z crate" is routed to INVENTORY and never produces this message. Whoever hits exit 3 reads it, concludes a routine minor bump tripped the gate, and the obvious next move is to weaken the gate rather than fix the baseline. | Say what is actually reachable: the baseline is at or above the declared version (a version inversion — check the `CHECK` line), or the crate is `0.0.z`. Add one clause noting a normal 0.x minor bump is handled by the advisory INVENTORY arm and never lands here. | Fix here |
| LOW | scripts/check_semver.sh | 769 | The comment claims "the LAST summary is the conservative one to judge". It is not: a log whose first summary is `0 checks` and whose last is `196 checks` returns a verdict (verified). Unreachable today — every invocation passes a single `--package` and emits exactly one summary, including the 11-feature `break.out`. | Either judge the minimum across all summaries, or drop the "conservative" claim and say the invocation is single-crate by construction. | Parent |

## Notes

- No CRITICAL, no HIGH. The fix moves a false PASS to a fail-closed exit 3 on a
  path a real release cannot currently take, and leaves the advisory arm exit 0.
- `clean.out` was captured with `--release-type minor`, so it carries
  `(assume minor change)` while the PASS/FAIL arm that consumes it never passes
  that flag. Provenance is documented honestly in the fixture README and the
  classifier reads only the count, so this changes nothing — noted, not flagged.
- The INVENTORY blind branch still exits 0 (`inventory_blind` is reported, not
  counted against the status). That is pre-existing and correct for an advisory
  arm; not a finding against this PR.
- CI at review time: everything green except the four `Rust tests (pre-publish
  gate)` shards, still IN_PROGRESS. The `Public API / SemVer` job — which runs
  `check_semver_selftest.sh` on `pull_request` — is SUCCESS.

Issue #3672, opened 2026-07-22
ci: post-#3669 cleanup — no-op .rcgu.o sweep, trusty-agents build.rs self-touch double-compile, unguarded find


bobmatnyc:
Post-merge code-critic WARN on PR #3669 (merged de6e5485, closes #3668). The operative fix (CARGO_PROFILE_TEST_DEBUG=line-tables-only) is verified working — Test job passed at 33G peak vs 118G available. Three cleanups owed:

1. **Sweep step is a measured no-op**: the `find target/debug/deps -name '*.rcgu.o' -delete` step logged 33G before AND after (job 88978057864 lines ~3298-3301) — with CARGO_INCREMENTAL=0 job-wide and line-tables-only, there's little/no .rcgu.o to reclaim. Action: add a `| wc -l` count to before/after logging to prove match count; if 0, remove the step and correct the ci.yml comment narrative (fix = debuginfo profile alone).
2. **trusty-agents/build.rs self-touch**: it declares `cargo:rerun-if-changed=ui/dist/index.html` but `ensure_ui_dist()` unconditionally rewrites that file every run (SKIP_UI_BUILD=1 path), invalidating cargo's fingerprint — so the second `cargo test` invocation re-Compiles trusty-agents v0.38.6 + trusty-agents-local v0.1.3 every CI run (log lines 3322,3324 vs 3055,3118). Fix: only write the placeholder when content would change. Benefits every double-invocation workflow, not just this one.
3. **Unguarded find**: add `2>/dev/null || true` to the find line for symmetry with the du guards (latent-only today).

Also record the confirmed-clean facts so nobody re-litigates: doctest coverage preserved (185 test-result blocks incl. doctests in the execution invocation; --no-run builds none), flag identity verified, no workspace test depends on debuginfo beyond file:line. Refs #3668, PR #3669.

bobmatnyc:
Closing as part of a 1.3.8 backlog reset. Current shipped behavior is being taken as the baseline, and open non-feature issues are being cleared rather than individually re-verified. This is not a claim that the issue was fixed — if it is still causing problems, please reopen with a current reproduction.

PR #5525, opened 2026-08-11
perf(ci): cut Rust test build overhead

bob-duetto:
## Summary

Reduce the Rust test gate from four execution shards to two and remove build work that every runner was repeating.

The test surface is unchanged: every shard still resolves the full workspace feature set, nextest still partitions the complete test list, shard 1 still runs doctests and the `CI=""` carve-out, and the aggregate `Rust tests (pre-publish gate)` job still fails unless every shard succeeds.

## Measured root cause

Two recent successful `main` runs showed the same shape:

- every shard spent 423–501 seconds compiling the same full workspace;
- each shard's nextest step took 143–187 seconds, including an 82-second rebuild of `trusty-agents` and `trusty-agents-local` caused by switching from Cargo's native build fingerprint to nextest's;
- actual test execution was about 75 seconds per shard;
- the four shard counts were balanced within roughly 1.7%; and
- the slowest individual test was 31 seconds, so neither imbalance nor slow-test tails dominate.

On run 31526290386 alone, the four duplicated build phases consumed about 31 runner-minutes before test execution.

## Changes

- Reduce the matrix from four shards to two. This roughly halves duplicated workspace compilation and per-runner setup while keeping the critical path close to the four-shard result.
- Build with `cargo nextest list --list-type binaries-only`, so `cargo nextest run` consumes the same fingerprint instead of recompiling two large workspace crates.
- Remove the `.rcgu.o` sweep tracked by #3672; live logs showed 42G both before and after it, so it was a measured no-op.
- Derive displayed/partition shard totals from `strategy.job-total` where possible so matrix changes cannot silently drop a partition.

## CI measurement

The first two-shard run (31535913005) was intentionally cold because an experimental profile change rotated the Cargo cache key:

- full workspace builds: 12:03 and 11:58, both within the repository's previous cold-build range;
- nextest execution: 3:32 and 3:33, with no separate 82-second rebuild after the nextest-native build phase;
- shard jobs: 17:35 and 16:25, including shard 1's once-only doctest and feature carve-outs; and
- both shards and the aggregate gate passed.

That run did not show a cold-build benefit from disabling test debug info and incremental compilation, so those experimental profile changes were reverted. The follow-up run retains the established `line-tables-only` profile and its compatible cache key.

## Expected impact

On warm-cache runs, Rust-test compute should fall from roughly 47 runner-minutes to about 25 runner-minutes by paying for two full workspace builds instead of four. The main-test portion now scales cleanly to two shards, while critical-path time should stay near the existing gate.

## Validation

- `actionlint .github/workflows/ci.yml`
- `bash scripts/check-ci-helpers-selftest.sh` — all 135 cases passed
- `git diff --check`
- cold two-shard CI run 31535913005 — both shards and aggregate gate passed

Refs #3672.


bobmatnyc:
Superseded by #5704 (`fa6befa3`), which stops the `Rust tests (pre-publish gate)` shards from running on pull requests at all — they now run on push to `main` and via `workflow_dispatch`. This PR cuts the shard matrix from four to two to reduce per-PR cost, and that cost no longer exists on a PR. It also edits the same job block `fa6befa3` changed, so it would conflict.

Closing unmerged; branch preserved. If the two-shard split is still wanted for the push-to-`main` run, that is a fresh change against the current job definition rather than a rebase of this one.

PR #5526, opened 2026-08-11
fix(trusty-memory): fail a registry read in kg_rebuild instead of reporting an empty run

### LINKED ISSUE #5511: fix(trusty-memory): purge_palaces fail-open on list_palaces read error
`purge_palaces` in `crates/trusty-memory/src/commands/kg_twin_merge.rs` calls `PalaceRegistry::list_palaces(...).unwrap_or_default()` — the same fail-open PR #5488 just fixed in the sibling `merge_palaces` (see the critic finding at https://github.com/bobmatnyc/trusty-tools/pull/5488#issuecomment-5257918017).

`purge_palaces` is itself a destructive pass. A failed registry read silently becomes an empty palace list, so the pass prints a zero-count success summary and exits 0 while having purged nothing — the operator believes the purge ran. Same reachable trigger as the `merge_palaces` twin: a macOS TCC denial surfacing as EPERM from `read_dir`.

Pre-existing on main, not introduced by #5488; kept out of that PR to preserve its single outcome (the twin merge).

Fix shape: propagate the `list_palaces` error instead of defaulting, matching what #5488 now does in `merge_palaces`.

Closure conditions:
- `purge_palaces` returns/propagates the `list_palaces` error instead of `unwrap_or_default()`
- A new regression test in `kg_twin_merge.rs` (reusing the harness #5488 established for this function family) fails against the pre-fix commit and passes after

Confidence: Reproduced — independently confirmed by code-critic during the #5488 re-review, which scored the sibling instance HIGH before it was fixed.

bobmatnyc:
Closes #5511

**The defect.** Two passes in `crates/trusty-memory/src/commands/kg_rebuild.rs` swallowed a failed `PalaceRegistry::list_palaces` read into an empty palace list via `unwrap_or_default()`, then reported a clean zero-count run over data they never read. Reachable trigger: a macOS TCC denial producing EPERM from `read_dir`.

**Both are fixed, and their costs differ — say so, do not flatten them:**
- `purge_palaces` (:362) is DESTRUCTIVE. It exited 0 while the operator believed subjects had been deleted.
- `rebuild_palaces` (:544) only asserts. It destroys nothing, but reports a real-looking `0 drawers, 0 triples` to someone watching the triple count for a back-fill that never opened a palace.

Both now propagate with `with_context`. Both callers (`report_purge`, and `kg_rebuild.rs:149`) already propagated with `?`, so no plumbing changed.

**Scope note worth stating:** `rebuild_palaces` was folded in rather than filed separately, per this repo's rule that an easy fix found while working on a file ships in the same work. It is the identical one-line shape in the file this PR already edits.

**This completes the family.** #5488 fixed the same fail-open in `merge_palaces` (`kg_twin_merge.rs`). All three now have error-arm regression tests, and they pass together:
```
test commands::kg_twin_merge::tests::merge_palaces_propagates_an_unreadable_data_root ... ok
test commands::kg_rebuild::tests::purge_palaces_propagates_an_unreadable_data_root ... ok
test commands::kg_rebuild::tests::rebuild_palaces_propagates_an_unreadable_data_root ... ok
```

**Test ladder: rung 5.** Red→green for each, with only that arm's production fix reverted. The purge:
```
an unlistable data root must fail the purge, not report an empty one: []
```
The back-fill:
```
an unlistable data root must fail the back-fill, not report an empty one: []
```
The trailing `: []` is the defect verbatim — `Ok(vec![])`, an empty summary over data never read. The harness reuses #5488's approach: a regular file at `data_root` reaches the same `read_dir` error arm as an EPERM denial without depending on the suite's uid. Coverage includes `apply=false`, `apply=true`, and the `report_purge` wrapper.

**Gates:**
```
cargo fmt --check                                             EXIT=0
cargo test -p trusty-memory                                   EXIT=0  (lib 724 passed, 0 failed, 4 ignored; 21 integration binaries, 0 failed)
cargo check --workspace                                       EXIT=0
clippy 0.1.97, CI's workspace invocation                      EXIT=0
scripts/check_line_cap.sh / check_changelog_fragment.sh / check_sld.sh   EXIT=0
scripts/check_test_pointers.sh                                EXIT=1  — pre-existing, see below
```

**The one red gate is not this PR's.** `check_test_pointers.sh` reports exactly one dangling pointer, byte-identical before and after: `helpers.rs:331` citing `count_active_triples_surfaces_read_failure`, which lives in `trusty-common` while the linter searches only the citing crate. This PR's two new `Test:` pointers both resolve — the count stayed at 1, it did not go to 2. PR #5521 already fixed the `helpers.rs` citation on main; this branch predates that merge.

**Expect `PR version bump vs crates.io (issue #4421)` to FAIL on this PR, and here's why:** this branch changes `crates/trusty-memory/src/**` while the crate's in-tree version is `0.22.0`, published to crates.io at 2026-08-11T20:15:27Z. The bump to `0.23.0` is landing separately on PR #5512. Once it is on main, this branch rebases and the gate passes with no bump of its own — a second competing bump would only collide. Do NOT add a version bump to this PR.


bobmatnyc:
## Verdict: APPROVE

Adversarial review of `6be5d9855` (fix/5511-purge-palaces-fail-open) against merge-base `1e327311d` (#5488). Reviewed with my own worktree at the exact head SHA (`.claude/worktrees/agent-a2d3d614a8bbb1f02`), not by reading the diff alone.

### 1. Mirror fidelity — faithful, and "no plumbing change" is true

`git show 1e327311d -- crates/trusty-memory/src/commands/kg_twin_merge.rs` (`merge_palaces`, line 322-323) and this PR's `purge_palaces` (kg_rebuild.rs:377) / `rebuild_palaces` (kg_rebuild.rs:553) use the identical pattern: `.with_context(|| format!("list palaces under {}", state.data_root.display()))?`.

Both callers genuinely propagate, traced end to end:
- `purge_palaces` → `report_purge` (kg_rebuild.rs:210, `?`) → `handle_kg_rebuild_with` (kg_rebuild.rs:177, `?`) → `main()`'s `Result<()>` (main.rs:613), which on `Err` prints Debug and exits 1.
- `rebuild_palaces` → `handle_kg_rebuild_with` directly (kg_rebuild.rs:149, `?`) → same exit path.

No dead-end catch anywhere in the chain. The claim holds.

### 2. Fail-Open Check — whole file swept

Both fixed sites are correct and complete. One additional instance found, **outside this diff's scope**:

| MEDIUM | `crates/trusty-memory/src/commands/kg_rebuild.rs` | 585-622 (`rebuild_one`) | Per-triple `handle.kg.assert()` failures are logged via `tracing::warn!` (line 607) and dropped; `PalaceRebuildSummary.error` is hardcoded `None` at line 621 regardless of how many asserts failed. The CLI loop at line 153-165 reads that field to decide `[ok]`/`[error]`, so an operator sees a clean `[ok] palace=… drawers=N triples=M` line even when some triples silently failed to assert. Unlike its siblings — `purge_one` tracks `outcome.failed` and sets `error = Some(...)` (kg_rebuild.rs:443-448), `merge_one` does the same (kg_twin_merge.rs:391-397) — `rebuild_one` has the identical `error: Option<String>` field on its summary struct but never populates it. That asymmetry between three structurally-parallel per-palace summary builders is the tell. | Accumulate failed asserts (subject, predicate, error) the same way `apply_deletions`/`repoint` do, and set `error = Some(...)` when non-empty, mirroring `purge_one`. | Promote |

Bounded impact, not a reason to gate this PR: `handle_kg_rebuild_with` never `bail!`s on `rebuild_palaces`' per-palace errors the way it does for purge/merge failure counts — the rebuild pass was already documented as "non-fatal" by design, so this bug degrades an `[ok]` label and an error tally, not an exit code. Severity is MEDIUM rather than HIGH for that reason. It is unchanged code this branch does not touch, so it does not affect the verdict on the actual diff — flagged only because the review was asked to sweep the whole file, not just the two fixed call sites.

Nothing else in `kg_rebuild.rs`, `purge_one`, `scan_active_triples`, `stale_subject_candidates`, or `apply_deletions` fits the shape (each either propagates or already tracks failures correctly in its return type).

### 3. Reproduced the claimed red myself

Reverted `purge_palaces` (kg_rebuild.rs:377 → `.unwrap_or_default();`) in the worktree and ran `purge_palaces_propagates_an_unreadable_data_root` alone:

```
thread '...purge_palaces_propagates_an_unreadable_data_root' panicked at crates/trusty-memory/src/commands/kg_rebuild.rs:1108:18:
an unlistable data root must fail the purge, not report an empty one: []
test result: FAILED. 0 passed; 1 failed; ...
```

Restored, then did the same for `rebuild_palaces` (kg_rebuild.rs:553):

```
thread '...rebuild_palaces_propagates_an_unreadable_data_root' panicked at crates/trusty-memory/src/commands/kg_rebuild.rs:1319:14:
an unlistable data root must fail the back-fill, not report an empty one: []
test result: FAILED. 0 passed; 1 failed; ...
```

Both match the claimed red exactly — panic ending `: []`, the empty `Vec` the pre-fix code silently returned. Restored the file (`git diff` clean) and re-ran green: `test result: ok. 3 passed` for the three `propagates_an_unreadable_data_root` tests together (purge, rebuild, and the sibling `merge_palaces` one from #5488).

On the harness fidelity question: `std::fs::read_dir()` on a regular file returns `Err(ENOTDIR)`, not the `EPERM` a TCC denial produces — different errno, but the same code path (`PalaceStore::list_palaces`'s single `.map_err(...)?` at palace_store.rs:177, unconditional on the specific `io::Error`). The one open question the harness doesn't settle is whether `Path::exists()` (palace_store.rs:173, checked before `read_dir`) itself survives a TCC denial — if TCC blocked `stat()` too, `exists()` would return `false` and the function would take the *early* `Ok(Vec::new())` return at line 174 instead of ever reaching `read_dir`, which is a different fail-open this test cannot see. That's plausible-but-unverified on this machine, not something I can settle without an actual TCC-denied directory. Not a blocker: this is the identical harness shape #5488 already used and shipped for `merge_palaces`, so it isn't a new gap introduced by this PR.

### 4. Coverage completeness

`purge_palaces_propagates_an_unreadable_data_root` (kg_rebuild.rs:1099) loops `for apply in [false, true]`, asserting `Err` on both, then separately calls `report_purge(&state, None, true)` and asserts `Err` there too. `apply=true` — the arm that matters for a destructive pass — is genuinely covered, not just the dry-run path. Confirmed by direct execution above.

### 5. Was folding `rebuild_palaces` in correct?

Defensible, not a process finding. Both changes are the same one-line pattern (`.unwrap_or_default()` → `.with_context(...)?`) on the same shared call (`PalaceRegistry::list_palaces`) inside the same file, and the changelog fragment explicitly separates their severity (destructive vs. assert-only) rather than blurring them into one claim. This reads as one outcome — "the registry-list fail-open, wherever it still exists in this file" — not two.

### Known, not scored

`check_test_pointers.sh`: `test-pointers: 1 dangling pointer(s), 0 stale allowlist entries` — the only hit is `helpers.rs:331` (`count_active_triples_surfaces_read_failure`, #5489, predates this branch). This branch's own new `Test:` pointers all resolve.

### Gates run

- `cargo fmt --check -p trusty-memory && cargo check -p trusty-memory && cargo clippy -p trusty-memory --all-targets -- -D warnings` — clean, exit 0.
- `cargo test -p trusty-memory --lib` — `test result: ok. 724 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out`.

### Notes

The one MEDIUM finding above (`rebuild_one`'s hardcoded `error: None`) is a genuine, separable defect discovered by the requested whole-file sweep. It sits in code this PR does not touch and does not gate this PR's verdict; recommend it as a follow-up ticket (Promote) rather than folding it in here.

---

Issue #839, opened 2026-06-06
reindex: warm incremental reindex drops hash-skipped files' chunks from durable redb — data loss on next daemon restart


bobmatnyc:
## Symptom
On a WARM daemon, an incremental reindex (where unchanged files are hash-skipped) writes ONLY the re-embedded (changed) files to the durable corpus. Skipped files' chunks survive only in the in-process `chunks` HashMap / HNSW — so **after the next daemon restart, the durable `index.redb` contains only the files that were re-indexed in the last run, and the skipped files' chunks are lost** (the index effectively shrinks to the last delta).

## Root cause (file:line)
- Every non-force reindex stages into a FRESH EMPTY redb: `begin_force_corpus_swap` → `CorpusStore::open_fresh` deletes + creates an empty `index.redb.tmp` (`crates/trusty-search/src/core/corpus.rs:276-286`; staging is enabled for ALL durable reindexes via `staging::should_stage`, `staging.rs:34-36`).
- `swap_corpus_store` replaces only the redb handle, NOT the in-memory `chunks` map (`reindex/mod.rs` ~1452). Hash-skipped files are `continue`d and never written to the staging store (`reindex/mod.rs:1091-1099`).
- On promote, `commit_force_corpus_swap` atomically renames staging over live (`reindex/mod.rs` ~1475). The promoted redb now lacks the skipped files' chunks. Next warm-boot loads only what's in redb → data loss.

## Fix (proposed)
Carry unchanged chunks into the staging corpus before/around the batch loop: after `open_fresh`, bulk-copy the existing chunk rows (+ hash entries) for files that will be skipped from the live corpus into the staging store (add a `copy_chunks_for_files`/`copy_all_except` bulk method on `CorpusStore`), OR determine the skip set first and copy those chunks, then re-embed only changed files. Add a regression test: reindex with a subset changed → restart → assert all chunks (changed + unchanged) are present.

Severity: **HIGH** — silent durable data loss on incremental reindex + restart. Found via code diagnosis (not yet reproduced live).

PR #5527, opened 2026-08-11
fix(trusty-search): carry contributed graphs across the reindex corpus swap

bobmatnyc:
This PR addresses a silent data-loss defect discovered during review of PR #5514. No issue exists for it; the owner authorized fixing it in-wave rather than filing one.

## The defect

A reindex destroys every contributed graph and reports success.

`begin_staged_corpus_swap` (`service/reindex/corpus_swap.rs:104`) opens a fresh `index.redb.tmp` via `CorpusStore::open_fresh`, seeds it from the live corpus with `copy_all_from` (`core/corpus/meta_ops.rs:347` — chunks, entities, file hashes, `_meta`), and `commit_staged_corpus_swap` renames it over the live file. `kg_contrib` was in neither the fresh file nor the copy, so the rename deleted it. A force reindex skipped the seeding call entirely and lost it the same way.

Nothing rebuilt it. `kg_nodes` / `kg_edges` are derived from the chunk corpus and genuinely regenerated by `rebuild_symbol_graph_for_reindex`; `kg_contrib` is written only by `POST /indexes/{id}/graph`, which a reindex never calls. By the time `rebuild_kg` ran (`finish.rs:337`, after `resolve_corpus_swap` at `finish.rs:267`), `self.corpus` was the promoted contrib-empty store.

Nothing detected it either: `load_contrib_graphs()` on an empty table returns `Ok(vec![])`, which is indistinguishable from "nothing was ever contributed". It is not an error path, so none of #5514's new machinery (`ContribMergeOutcome`, `mark_graph_failed`, the 503) could see it.

Observed against the real entry point before the fix — reindex `Complete`, graph stage `Ready`, contribution gone:

```
assertion `left == right` failed: the promoted corpus must still hold the contribution (force=true)
  left: []
 right: ["navigatsql"]
```

[ADR-0009](docs/adr/0009-external-extractor-kg-ingest-contract.md) §Decision 1 states the contract this restores: "Reindex regenerates derived tables only; contributed data survives restart and reindex by construction."

Pre-existing, not a regression from #5514 or #5505: `corpus_swap.rs` and `copy_all_from` are untouched by that work, and `copy_all_from`'s claim dates to #839.

## The fix

**Seam: the copy, at swap-begin — not the swap and not a re-merge after the rebuild.**

`CorpusStore::copy_contrib_from` (`core/corpus/contrib.rs`) carries `kg_contrib` from live into staging, and `begin_staged_corpus_swap` calls it unconditionally, before any batch writes. Three consequences follow from that one placement, which is why it beats the alternatives:

- The contributions are in the staging file when the rename happens, so the promoted corpus holds them — no second write after the swap that could itself fail.
- The KG rebuild that follows the swap reads them from the promoted store through the path it already uses, so the serving graph is correct without touching `rebuild_kg`. That also means **no collision with #5514**, which owns `finish.rs`'s `rebuild_kg` and `completion.rs`. This PR touches neither.
- A rollback or abort leaves the live corpus untouched, contributions included.

**Yes, force copies it too.** Force starting empty is right for chunks, which the run rebuilds. It is wrong for data the daemon cannot regenerate: a force reindex has no more licence to discard externally-supplied rows than an incremental one does. That is the ADR's "by construction", and force was the worse of the two paths — it carried nothing at all.

Failure handling reuses the existing rule and is safe on both sides: on the incremental path a copy failure aborts the run with the live corpus intact (as `copy_all_from` failures already do); on the force path it falls through to direct-write-to-live, which performs no rename, so the live `kg_contrib` rows stay exactly where they are.

**Doc correction.** `copy_all_from` asserted that "KG tables ... are rebuilt from scratch at the end of every reindex", which is true of the derived pair and false of the contributed one. That sentence is what made this defect invisible to a reader; it now names which half is regenerable and which is externally supplied and unrecoverable if dropped.

## Testing — rung 5 (persistence / data integrity)

The regression tests drive `spawn_reindex_awaitable`, not the rebuild seam. The seam already had passing coverage (`ingest_survives_derived_rebuild`) throughout the defect's lifetime, which is exactly how it survived.

```
cargo fmt --check                                                    # clean
cargo clippy -p trusty-search --all-targets -- -D warnings           # 0 warnings
cargo test -p trusty-search                                          # 1588 + 387 passed, 0 failed (lib + bin) + 15 integration targets, 0 failed
cargo test -p trusty-search -- --include-ignored                     # 1589 passed lib; see below
cargo check --workspace                                              # clean
cargo test -p trusty-agents                                          # clean (only path-dependent crate)
```

New coverage, all three failing pre-fix on the assertion quoted above (`left: []`) after the run reported `Complete`:

- `contributed_graph_survives_an_incremental_reindex`
- `contributed_graph_survives_a_force_reindex`
- `contributed_graph_survives_repeated_reindexes` — force → incremental → force, because each run reads from the corpus the previous run promoted
- `copy_contrib_from_carries_every_producer`, `copy_contrib_from_missing_source_table_is_a_no_op` (a never-contributed index has no such table and must not abort), `copy_contrib_from_preserves_an_unreadable_row` (the copy must not become a second place that decides which contributions are valid)

Each end-to-end test asserts both halves: the row is in the promoted corpus, and the contributed node is back in the serving graph.

## Race window

`copy_contrib_from` opens its own read transaction, after `copy_all_from` commits. An ingest landing between that snapshot and `swap_corpus_store` writes to the LIVE corpus and is missed by the promotion — a window bounded by this copy's own duration, so **narrower** than the #839 chunk window it sits inside. It also **extends** that pre-existing chunk window by however long the copy takes, which is small: `kg_contrib` holds one blob per producer against a whole-corpus copy.

## Behavior change

A corrupt `kg_contrib` row now survives reindexes. `load_contrib_graphs` fails all-or-nothing, so a row whose bytes will not deserialize poisons the whole load. Before this fix, every reindex destroyed that row along with the good ones, which incidentally healed the table into an empty, readable state. It now persists, so the load keeps failing every boot and `save_then_merge_contrib` keeps degrading to the derived-only graph until the producer re-ingests or the row is deleted. #5505's `blocking_producer` field names which row. Pinned by `copy_contrib_from_preserves_an_unreadable_row`. This trades "reindex silently destroys corrupt data" for "reindex preserves it inertly and warns" — deliberate, and the right way round.

Refs #5514


bobmatnyc:
**code-critic verdict: APPROVE** — no required changes. (Reviewed `e230092ec` against baseline `f0181f5f0`.)

Verified by execution, not inspection. Built a worktree at the PR head and a second at the pre-fix baseline with only `contrib_survival_tests.rs` copied over and registered: 3 passed at HEAD, 3 failed at baseline on exactly the quoted assertion (`left: []`, `right: ["navigatsql"]`) — each after `progress.status.load() == Complete` had already been asserted true. The defect is real and silent, and the regression tests are not vacuous. Full crate at HEAD: 1588 passed, 0 failed, 1 ignored. `fmt --check` and `clippy -D warnings` clean.

Traced every `kg_contrib` reader — `symbol_graph/graph.rs` (warm boot) and `symbol_graph/contrib.rs` (`save_then_merge_contrib`) — both reach it through `indexer.corpus_store()` under the same `RwLock` the swap uses, so there is no split-brain path. Both failure arms of `begin_staged_corpus_swap` (`corpus_swap.rs:182-217`) match the PR's claims: incremental aborts with the live corpus untouched and best-effort deletes the orphaned staging tmp; force falls through to direct-write-to-live with no rename. `commit_staged_corpus_swap` / `abort_staged_corpus_swap` / the resume-adoption path are byte-for-byte untouched, so crash-between-copy-and-rename is unaffected — adoption performs no re-copy and the staged file already carries the copied rows. No overlap with #5514's `rebuild_kg` machinery. "No public API change" independently confirmed: the diff adds one `pub(crate)` method and doc changes.

One LOW finding, wording only, being corrected in the PR: the residual-limitation paragraph calls the post-copy race window "identical" to the pre-existing #839 chunk window. It is narrower for contrib (its transaction opens a later snapshot) and marginally extends the chunk window (the swap moves later by `copy_contrib_from`'s duration). Bounded and negligible against `copy_all_from` on a real corpus.

Behavior change worth recording: a corrupt `kg_contrib` row now survives reindexes rather than being destructively wiped. `load_contrib_graphs` fails all-or-nothing, so such a row keeps erroring and the graph degrades to derived-only via the existing warn path — a deliberate, tested trade (`copy_contrib_from_preserves_an_unreadable_row`) of "reindex silently destroys corrupt data" for "reindex preserves it inertly and warns."

Issue #5350, opened 2026-08-10
trusty-search: MCP transport flattens a structured 503 into a prose error string


bobmatnyc:
\`mcp/tools/http.rs:119-127, 222-234\` collapses a structured 503 body into a prose \`Transport\` string.

[PR #5345](https://github.com/bobmatnyc/trusty-tools/pull/5345) gave the daemon's 503s a machine-readable shape — \`error\` code, \`retryable\` flag, \`restore_via\` — specifically so a caller can branch on them. An MCP client gets prose instead and cannot.

\`classify_index_miss\` already does the right thing for 404: map a known \`error\` code to a structured error carrying the body as \`data\`. The same treatment applied to 503 would preserve the contract end to end.

Without this, the residency and vector-availability contract added in #5345 stops at the HTTP boundary and does not reach MCP consumers.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

Issue #5264, opened 2026-08-09
Codex MCP startup must make trusty-search usable and project-scoped


bobmatnyc:
## Outcome

A fresh Codex task receives a self-contained, project-scoped trusty-search MCP connection whose health and search calls work without manually starting a separate daemon.

## Why this is reopened

The original missing-`serve` registration was repaired, but its end-to-end acceptance is still failing. On 2026-08-11, Codex starts `trusty-search serve` and exposes the tool catalog, yet `search_health` fails while connecting to `http://127.0.0.1:7878/health`. The project instruction's `trusty-search search` path likewise waits for daemon readiness and yields no result.

The current registration is also argless beyond `serve`, even though the CLI supports `serve --project .` or an explicit index. Tool discovery alone is not a successful integration.

## Done when

- [ ] The Codex stdio entrypoint is self-contained: it starts or attaches to the required search service, or performs search directly, without an undocumented manual daemon step.
- [ ] Setup binds the MCP session to the active project or resolves the project/index deterministically and reports that choice.
- [ ] Health failures return structured, actionable diagnostics rather than a raw localhost connection error.
- [ ] Re-running setup repairs stale registrations without overwriting unrelated user configuration.
- [ ] A clean-session regression test completes MCP initialize, `tools/list`, health, index discovery, and one project-scoped search.
- [ ] Doctor reports executable version, effective arguments, selected project/index, daemon ownership, and remediation.

Sibling: #5265.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bob-duetto:
Sibling failure for the other Codex MCP registration: #5265.

bob-duetto:
Follow-up reproduction after manual editing: `codex mcp get trusty-search --json` now reports `args: ["[\\"serve\\"]"]`. This is still one literal JSON-looking argument, so the process receives the subcommand text `["serve"]` rather than `serve` and cannot initialize MCP. The required semantic argument vector remains `args = ["serve"]`.

bobmatnyc:
Moving to 0.45.0: PR #5347 (which carries this issue) has returned code-critic BLOCK twice — the second round fixed the original CRITICAL but reintroduced the same defect one layer up (a read error on `daemon.env` downgrades to "empty file", then the merge overwrites the operator's file with that). Rather than hold trusty-search 0.44.0 for it, this moves to 0.45.0 so the PR gets a third review round without deadline pressure.

PR #5534, opened 2026-08-11
feat(trusty-search): structured MCP health, doctor registration check, stdio e2e (#5264)

bobmatnyc:
Refs #5264 — lands three of the four acceptance criteria. Project scoping (criterion 1) carries a design question with the owner and is dispatched separately; this PR deliberately does not touch it and adds no `env` parameter to `codex_config::patch_mcp_server`.

## What this satisfies

**Criterion 3 — structured health diagnostics.** `mcp/tools/misc.rs` forwarded `GET /health` verbatim, so a refused connection and an HTTP 500 both arrived as one `DispatchError::Transport` prose string with no remedy in it, and a healthy 200 named neither the daemon that answered nor whether the caller's project was indexed on it. `search_health` now returns a report carrying `status`, `healthy`, `message`, `remediation`, `daemon`, and `index`. Four verdicts: `daemon_unreachable`, `daemon_error`, `index_not_registered`, `index_empty`, plus `ok`.

The daemon block forwards the responder's own `/health` fields — base URL, version, index count, total chunks, uptime — so "healthy" and "healthy, but not yours" are distinguishable. That is the concrete failure this half was filed against: a probe from an isolated project silently attached to the machine's production daemon (42 indexes, 430k chunks) and got a green 200 back that said nothing about which daemon replied.

Two shape changes worth reviewing directly:
- The tool now returns `Ok` even when the daemon is down. A health probe that cannot reach the daemon has determined health; that verdict belongs in the body, not in an error envelope a caller must parse prose out of. `healthy` is the field to branch on — a successful call no longer implies a healthy daemon. The tool description says so.

  This diverges from `not_ready.rs` (#4715) and `unavailable.rs` (#5350), which enrich the body but still return `Err(DispatchError::…)`. The divergence is deliberate and rests on what the tool is for. Those two report that a REQUESTED OPERATION could not be performed — a search that could not run is a failed search, and `isError: true` is the correct answer. `search_health` performs no operation on the caller's behalf; its entire job is to return a verdict about the system, and "the daemon is down" is that verdict successfully produced, not a failure to produce one. Modelling it as `isError` would mean the tool can only succeed when it has nothing interesting to say. The cost is that a caller reading `isError` alone learns nothing, which is why `healthy` is stated first in the tool description and in the report.
- `index_id` is a new optional argument. Otherwise the session pin decides, then the id the CLI would derive from the working directory. The report says which source decided via `index.resolved_from`, so a `cwd`-derived answer cannot be mistaken for one about a pinned session's index.

**Criterion 6 — doctor coverage.** `doctor` had ten checks, none of which read a client registration, which is why a `trusty-search` running perfectly alongside a Codex registration that could never reach it produced an all-green report. `McpRegistrationCheck` reports executable and version, effective arguments, selected project/index, daemon ownership, and remediation. A registration missing `serve` is an `Error`; an absent one is a `Warn`.

Claude Code is covered as well as Codex — it was a cheap generalization, the same predicate over a different serialization. Scope limit stated in the module doc: only Claude's two GLOBAL settings files. `setup` also patches per-project `.claude/settings.json` files it finds by an eight-deep `$HOME` walk; repeating that walk in a diagnostic would make `doctor` take tens of seconds on a large `$HOME`, and a diagnostic nobody runs reports nothing. Cursor is untouched — `integrate.rs` registers it by a different path.

**Criterion 5 — end-to-end regression.** `setup.rs`'s existing tests assert file contents: they prove the vector on disk says `serve`, not that launching it yields a process that speaks MCP. That is the gap that let this reopen. `tests/mcp_stdio_e2e_5264.rs` spawns the built binary and drives initialize → tools/list → search_health → list_indexes → search over a real stdio pipe against a real `build_router` daemon.

Hermeticity is asserted, not just configured. `TRUSTY_DATA_DIR` and `HOME` are tempdirs, and both discovery files (`http_addr`, `daemon.port`) are seeded with the test's ephemeral port so even `daemon_base_url()`'s fallback cannot reach the compiled-in `127.0.0.1:7878`. Then `assert_answering_daemon_is_ours` checks the health report's `base_url` equals the test's own address and that the daemon serves exactly one index. Without that assertion, a discovery regression that routed the child to the machine-wide daemon would leave every other assertion passing.

## Critic round (WARN — both HIGH findings fixed)

Both findings were the same defect this PR fixes, one layer over: a green verdict on a condition nothing checked.

**HIGH 1 — `search_health` reported ok on an index-level check that never ran.** When no `index_id` was given, no session pin existed, and no id could be derived from the working directory, the report was `status: ok, healthy: true` and byte-identical to one where the index was probed and found fine. That branch now returns `index_unknown` with `healthy: false` and states in prose that no project-level check ran. `handle_search_health` split into a thin wrapper plus `report_health`, which takes the scope as a parameter — `resolve_scope`'s last fallback reads `current_dir()`, so the branch is otherwise unreachable in-process. That is the same split `doctor_data_dir_from` uses in `commands::doctor_checks`, for the same reason.

**HIGH 2 — doctor said nothing when both global Claude settings files were absent.** The loop skipped missing files and emitted zero Claude Code lines. That is not exotic: `setup` patches project-local `.claude/settings.json` files via an eight-deep `$HOME` walk, so zero global files is a normal state, and a broken project-local `args=[]` registration was then reported by nothing on the machine — #5264's own failure, reproduced for the client this PR added coverage for. `check_claude_registrations` now returns a warning naming the project-local files doctor does not scan. The module doc claimed the report already said this; it did not, and now it does.

Pre-fix evidence for both, each failing on the assertion the fix introduces:

```
# HIGH 1 — branch reverted to HEALTH_OK
test mcp::tools::tests_health::search_health_does_not_report_ok_when_no_index_could_be_resolved ... FAILED
    left: String("ok")   right: "index_unknown"

# HIGH 2 — branch reverted to returning an empty Vec
test commands::doctor_checks::mcp_registration::tests::absent_global_claude_files_still_emit_a_scope_warning ... FAILED
  assertion `left == right` failed: silence is the defect; expected one verdict, got []
    left: 0   right: 1
```

Gates re-run after the fix, on `origin/main` as of the rebase:

```
cargo fmt                                              # clean
cargo clippy -p trusty-search --all-targets -- -D warnings   # zero warnings
cargo test -p trusty-search
    test result: ok. 1595 passed; 0 failed; 1 ignored   # lib
    test result: ok. 399 passed; 0 failed; 3 ignored    # bin
    test result: ok. 1 passed; 0 failed                 # mcp_stdio_e2e_5264
bash scripts/check_line_cap.sh    # 3921 files, 0 violations — OK
bash scripts/check_sld.sh         # 0 errors, 0 warnings
bash scripts/check_changelog_fragment.sh   # OK, fragment valid
```

`lsof` after the run shows only the production daemon on 7878 — the e2e leaked nothing.

## Criteria still open

- **Criterion 1 — project scoping.** Not in this PR, by instruction.

## Found while implementing, not fixed here

A first draft of the e2e added a second test for a `serve` session with no reachable daemon. `handle_serve` builds its `DaemonBridgeConfig` with `no_spawn: false` and reads no environment override, so that session spawned a real `start --foreground` daemon which outlived the killed `serve` child — it left one listening on port 7882, two ports from production. The test was removed rather than worked around, and the reason is recorded in the file's module doc. There is no hermetic way to drive a daemonless `serve` until it gains a no-spawn switch. The unreachable-daemon verdict itself is covered by `search_health_reports_daemon_unreachable_with_remediation`. The leaked daemon was killed; production on 7878 was untouched.

## Gates

**Rung 6** — part 2 changes an MCP tool's response shape, so this owes direct API evidence beyond crate tests. The `mcp_stdio_e2e_5264` run below IS that evidence: a real client process, over a real pipe, against the real router.

```
cargo fmt                                              # clean
cargo clippy -p trusty-search --all-targets -- -D warnings
    Finished `dev` profile in 38.48s                   # zero warnings

cargo test -p trusty-search
    test result: ok. 1589 passed; 0 failed; 1 ignored   # lib
    test result: ok. 397 passed; 0 failed; 3 ignored    # bin
    (28 further integration binaries, all ok, 0 failed)

cargo test -p trusty-search --test mcp_stdio_e2e_5264
    test mcp_stdio_session_initializes_and_serves_tools_against_an_isolated_daemon ... ok

    line-cap: measured 3920 tracked .rs file(s); 4 allowlisted, 0 violations — OK.
bash scripts/check_sld.sh
    sld-lint: scanned 57 spec doc(s) + 3222 code file(s); 0 error(s), 0 warning(s)
```

`execute_doctor_against_test_daemon` — the documented timing flake, and this PR touches doctor — passed in the full run above. No baseline reds were hit.

### Pre-fix failure evidence

The seven `search_health` tests were run against the pre-fix dispatch arm (`"search_health" => Some(server.get("/health").await)`) with everything else in place. All seven fail, each on the field the fix introduces, and the raw pre-fix strings are visible:

```
test result: FAILED. 0 passed; 7 failed

search_health_reports_daemon_unreachable_with_remediation
  report is JSON: expected value at line 1 column 1
  Error: GET http://127.0.0.1:1/health: error sending request for url (http://127.0.0.1:1/health)

search_health_reports_a_non_2xx_daemon
  report is JSON: expected value at line 1 column 1
  Error: GET http://127.0.0.1:50579/health returned 500 Internal Server Error: {"error":"boom"}

search_health_reports_an_unregistered_project_index
    left: String("ok")   right: "index_not_registered"

search_health_names_the_answering_daemon
    left: Null           right: Bool(true)
```

The first two are the defect verbatim — the raw reqwest string a caller used to receive, with no remedy and no way to tell the two failures apart.

Parts 3 and 4 add surfaces that did not previously exist, so there is no pre-fix behavior for their tests to fail against. Their error arms are covered forward: `registration_with_empty_args_is_an_error` and `registration_with_nested_json_string_args_is_an_error` pin the two broken shapes #5264 reported, and `absent_registration_warns` pins the third.

## Coordination

Verified disjoint from the six live `trusty-search` PRs, which all sit in the indexing/corpus/reindex subsystem. The one shared surface is the generated MCP-tools table in `CLAUDE.md` and `README.md` — this PR changes exactly one row (`search_health`), regenerated with `UPDATE_DOCS=1 cargo test -p trusty-search --test generated_docs`. Any concurrent PR that adds or edits a tool descriptor regenerates the same table and will conflict on that line.

No version bump; `Cargo.toml` stays at 0.46.0. No breaking public-API change — the additions are new `pub` constants and a new `pub` module under `commands`.


bobmatnyc:
**code-critic verdict: WARN** — two HIGH findings, both since fixed in this PR. Cleared to merge at head `99a8dd616`.

Verified by execution, not inspection. The reviewer ran `mcp_stdio_e2e_5264` itself and confirmed by `ps` and `lsof` immediately afterward that production trusty-search on port 7878 was untouched and no daemon was left listening anywhere else — the test cleans up after itself. It traced the hermeticity claim through `daemon_base_url()` (`commands/daemon_utils.rs:53`) and `ensure_daemon_up()`'s fast-path probe (`trusty-common/src/mcp/daemon_bridge.rs:173`): both discovery files are seeded with the isolated daemon's live ephemeral address, so the fast path succeeds and the spawn branch is never reached. It also confirmed by direct read that `handle_serve` has no no-spawn path and nothing in the shipped suite calls it, matching this PR's account of the removed test and the daemon it leaked on port 7882.

`index_id`'s fallback chain (explicit → session pin → cwd) is correct and `resolved_from` cannot misreport its source — the three return points at `health.rs:110-118` are distinct and non-overlapping.

No breaking Rust-API change, verified independently: `commands` and everything under it, including the new `doctor_checks::mcp_registration`, is declared only in `main.rs`, never in `lib.rs`, so none of it is published library surface regardless of its own visibility. The only lib-visible additions are the `HEALTH_*` constants and a `pub(crate) mod health` re-exported through `mcp::tools` — purely additive. Separate point, not a defect: the `search_health` MCP tool's WIRE contract did change substantially in both shape and failure signalling, which `cargo-semver-checks` cannot see and which the "no breaking public-API change" line is not asserting.

**HIGH 1 — a green verdict on a check that never ran.** When `resolve_scope` returned `None` (no explicit `index_id`, no session pin, `derive_index_id` empty or `current_dir()` failing), `health.rs:120-132` returned `HEALTH_OK` / `healthy: true`, byte-identical to a report where the index was actually probed. The same shape as the bug this PR fixes, moved from the daemon layer to the index layer. FIXED: a new `HEALTH_INDEX_UNKNOWN` status with `healthy: false`, saying in prose that no project-level check ran. Pinned by `search_health_does_not_report_ok_when_no_index_could_be_resolved`, which fails pre-fix on `left: "ok"` vs `right: "index_unknown"`.

**HIGH 2 — doctor reported nothing when both global Claude files were absent.** `doctor_pipeline.rs:296-306` skipped silently, so zero Claude Code lines appeared. Not exotic: `setup` patches project-local `.claude/settings.json` via `discover_claude_settings`'s 8-deep walk (`trusty-common/src/claude_config.rs:78`), so a user can legitimately have no global files — and a broken `args=[]` project-local registration then produced no line at all, reproducing #5264's own "nothing on the machine reports it" failure for the client this PR added coverage for. FIXED: `check_claude_registrations` never returns an empty list; with neither global file present it emits one `Warn` naming both paths and the project-local files doctor does not scan. The module doc, which had claimed the report already said this, was corrected. Pinned by `absent_global_claude_files_still_emit_a_scope_warning`, failing pre-fix on `left: 0` vs `right: 1`.

**One finding promoted out of this PR, not a defect in it.** `trusty-mpm`'s injected agent docs (`tm-tool-usage-guide.md`, `tm-circuit-breaker.md`, `non-overridable-rules.md`) describe `search_health` as a bare liveness check and never mention the `healthy` field, so an agent that sees no error gets a false all-clear for a dead daemon. Cross-crate; addressed in PR #5536.

Gates at the fixed head: `fmt` clean, clippy zero warnings, `cargo test -p trusty-search` 1595 lib / 399 bin / 1 e2e passed, 0 failed. Line-cap 3921 files 0 violations, SLD 0 errors, changelog fragment valid.

PR #5539, opened 2026-08-11
feat(trusty-mpm): `d <glob>` bulk-deletes sessions by tmux name in the `tm ls` picker

bobmatnyc:
The `tm ls` picker's `d<N>` deletes one slot per confirmation round, so clearing
a batch of throwaway sessions costs one prompt each. `d tm-test-*` now deletes
every match in one action; `d <glob> --dry-run` previews without deleting.

Requested by Bob: "`tm ls` should allow wildcards — we have a bunch of test
sessions to delete", clarified to the delete function. Matching on the tmux name
is the owner's ruling.

## What the pattern matches

The tmux session name (`ManagedSessionSummary::name`). Real names off
`tm sessions ls` are `tm-test-01`, `tm-tmpsywch8`, `tm-apex-02` — prefix/suffix
structured and the only field an operator reads off the menu. The id is a UUID
nobody writes a pattern against; the slot number is already served by `d<N>`.

Glob, not regex, via `globset` — already a `[workspace.dependencies]` entry and
already in the lockfile, so this resolves no new dependency. Compiled with
`literal_separator(false)` so `*` spans the `-` in a name, and
`case_insensitive(true)`.

## Guards

This is a bulk destructive operation, so every guard narrows rather than widens.

- **Only stopped/errored sessions are ever deleted in bulk.** A running match is
  listed as kept. `force` is hard-wired `false`, so a session needing force can
  only be deleted individually through `d<N>`'s own force-confirm. This protects
  every live session on the machine, not just the caller's.
- **The caller's own session is excluded by tmux name too**, so it survives even
  if its record reads stopped (stale or display-reconciled). Two independent
  guards, since this is the one deletion an operator cannot undo from inside the
  session. Outside tmux the probe returns `None`, which disables only this check.
- **A session with an empty tmux name matches no pattern at all, `*` included.**
  Treating an absent name as an empty string is what would let `*` sweep up
  every nameless record.
- **Confirmation is the match count typed back, not `y`** — after the full match
  set is printed, each row leading with the tmux name plus a short id and the
  project, so two same-named records never render identically.
- **Zero matches says so and exits without deleting.** So does a match set that
  is entirely protected; neither prints a success line.
- The `d` prefix enters the bulk path only when the remainder carries a glob
  metacharacter, so a mistyped `delete` stays `Unrecognised` rather than
  becoming a bulk action against a session named `elete`.

Deletions route through the existing `picker_delete::delete_managed_then_local`
— the same path `d<N>` and `tm session delete` use. A daemon 409 is reported and
the row skipped, never escalated to a force.
`statusline::branch::tmux_session_name` is promoted to `pub(crate)` and reused
for the self-check rather than copied.

## Test ladder

Rung 3 — localized behavior inside `trusty-mpm`.

```
cargo fmt --check                                       # clean
cargo clippy -p trusty-mpm --all-targets -- -D warnings # clean
cargo test -p trusty-mpm                                # 4913 + 1633 passed, 0 failed
scripts/check_line_cap.sh          # 3932 files, 0 violations
scripts/check_changelog_fragment.sh # OK, fragment present
scripts/check_capabilities.sh       # up to date
scripts/check_test_pointers.sh      # 23859 citations, 0 dangling
```

30 new tests. The destructive-path ones assert the outcome is anything other
than `Confirm` — the single entrance to the deletion loop — which is a complete
proof no deletion follows, without mocking a daemon:
`glob_decide_no_match_is_not_destructive`,
`glob_decide_all_running_is_not_destructive`,
`glob_decide_star_never_deletes_nameless_sessions`,
`glob_decide_self_only_match_is_not_destructive`,
`glob_decide_dry_run_never_confirms`, `glob_decide_invalid_pattern`,
`glob_confirm_rejects_yes`, `glob_plan_nameless_session_never_matches`,
`glob_plan_excludes_current_session`,
`parse_picker_choice_glob_delete_typo_stays_unrecognised`.

No real sessions were deleted during development — all coverage runs against
in-memory fixtures.


bobmatnyc:
## code-critic: Verdict — APPROVE

Adversarial review, independent of the implementer. Centre question first.

### Can this delete a session the operator did not see? No.

The match set is computed **once** and never recomputed:
`confirm_and_delete_glob` receives the picker's own `&[ManagedSessionSummary]`
(`session_picker.rs:1062` renders every row of that same slice, with no
truncation or pagination), calls `decide_glob_delete` once, prints
`plan_lines(sessions, &plan)` for **every** matched row — deletable and
protected — and then indexes back into that same immutable slice inside the
deletion loop. There is no re-fetch and no second match pass between preview and
confirm, so the set shown and the set deleted are the same object. A session
created or resumed after the menu was fetched is not in the slice and cannot be
deleted; one that has since gone live is refused server-side (below) and the
refusal is printed.

Verified against the other probes:

- **Nameless guard** — `build_plan` skips `s.name.trim().is_empty()`, so both
  `""` and whitespace are excluded. `name` is a non-optional `String` on the
  wire type, so `None` is not reachable. I confirmed against globset built with
  this PR's exact flags that `*` **does** match `""` and `"   "`, so the guard
  is doing real work and its removal makes
  `glob_plan_nameless_session_never_matches` (expects `deletable == [2]`) and
  `glob_decide_star_never_deletes_nameless_sessions` (expects `NoMatch`) fail.
- **Stopped/errored filter** — applied to `plan.deletable`, which *is* what the
  deletion loop iterates, not merely to the display. `delete_needs_force` is
  `!needs_restart(state)` and `needs_restart` is a two-token allowlist
  (`"stopped" | "errored"`), so an unknown or new state word is treated as
  running and skipped. Fail-closed.
- **Second, authoritative guard** — `force` is hard-wired `false`, and the
  daemon's `delete_record` (`session_manager/delete.rs:71`) gates on
  `is_running`, which probes **live tmux** (`prune.rs:380`), not the persisted
  state. Any session whose tmux is alive gets a 409 regardless of what the
  client-side display state said, and the 409 is reported and the row skipped.
  The managed delete is also a **soft record delete** that never touches the
  workspace and never stops a runtime, which bounds the worst case.
- **`GlobDeleteOutcome::Confirm` is the single entrance to deletion** — every
  other arm of the driver's `match` ends in `return Ok(0)` before any HTTP
  client is touched. That makes the `assert_ne!(outcome, Confirm)` tests
  meaningful. `delete_managed_then_local` is genuinely the same entry point
  `d<N>` and `tm session delete` use, with the same 409/404 routing.
- **Tests** — re-run, not accepted from the PR body, in an isolated
  `CARGO_TARGET_DIR` so the author's worktree was untouched:
  `cargo test -p trusty-mpm --bin tm` → `test result: ok. 1633 passed; 0 failed;
  1 ignored`. 31 `glob_*` test lines executed, including all four cited
  fail-safe tests. `check_line_cap.sh` → 0 violations;
  `check_changelog_fragment.sh` → fragment present. No stale `#5533` reference
  survives anywhere in the tree.
- **`tmux_session_name` promotion** — widening `statusline::branch` to
  `pub(crate)` exposes exactly two items (`project_segment`,
  `tmux_session_name`), both read-only probes. Nothing that should have stayed
  private leaks.

### Findings

|---|---|---|---|---|---|
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/picker_delete_glob.rs` | 420 | The prompt prints the number the operator must type (`type 2 to confirm`), so the count-confirm can be satisfied without reading a single row — it is a higher-friction `y`, not the proof the module doc claims ("the only input that proves the number on screen was actually read is that number, typed back"). With 200 matches the rows scroll off the terminal while the token to type sits in the last line. | Drop the value from the prompt: `eprint!("tm: type the number of sessions to be deleted to confirm > ")`. The count stays in the header line and in the printed plan, so it is read from the match set rather than from the prompt. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/bin/tm/commands/picker_delete_glob.rs` | 441 | `delete_managed_then_local(...).await?` propagates on any non-200/404/409 status or transport error. Row 3 of 20 hitting a daemon 500 or a dropped connection aborts the whole bulk run **and** the picker with an anyhow error — partially applied, and the `N of M deleted` summary never prints. The sibling failure modes (409, 404) are correctly reported-and-continued; this arm is the asymmetric one. | Bind the result: `match delete_managed_then_local(...).await { Ok(r) => …, Err(e) => { eprintln!("tm: '{}' failed: {e}", s.name); failed += 1; } }`, and include the failure count in the summary line. | Fix here |
| MEDIUM | `crates/trusty-mpm/src/assets/skills/tm-cli-operations.md` | 275 | The bundled skill's picker section documents only `d<N>`, and its "Bulk cleanup" section lists `prune`/`decommission-ephemeral`. `d <glob>` and its `--dry-run` preview appear nowhere, so every agent that loads this skill is unaware both of the new destructive command and of its preview. | Add one bullet under the picker section (`d <glob>` — bulk delete by tmux-name pattern, stopped/errored only, count-confirmed) and one line to Bulk cleanup naming `d <glob> --dry-run`. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/picker_delete_glob.rs` | 368 | The self-exclusion silently disappears when the probe returns `None` for a reason other than "not in tmux" — `tmux_session_name` gives up after a 100 ms wall-clock timeout and on any non-zero exit. The preview shows no sign that one of the two guards the module doc calls "independent" is off. In practice the loss is covered by the daemon's live-tmux 409, so this is a doc-accuracy gap rather than an exploitable one. | Distinguish the cases: when `$TMUX` is set but the probe returned `None`, print one line before the prompt (`tm: could not resolve this tmux session's name — the self-exclusion guard is off for this run`). Alternatively soften the "two independent guards" claim to name the daemon probe as the second one. | Fix here |
| LOW | `crates/trusty-mpm/src/bin/tm/commands/picker_delete_glob.rs` | 161 | `--dry-run` is recognised only as a trailing token. `d --dry-run tm-test-*` — flag-first is a natural typing order — parses as the literal glob `--dry-run tm-test-*`, matches nothing, and reports "no sessions match", so the operator's attempt to preview looks like an empty result set. Fails safe, but it hides the feature at exactly the moment it is wanted. | Strip a leading `--dry-run ` as well as a trailing one before the metacharacter check. | Fix here |

### Notes

- Not flagged, by explicit scope: matching on the tmux name (owner ruling) and
  making it configurable.
- `build_plan` reads `s.state`, the display-reconciled field, where
  `types.rs:372` steers resume/restart decisions to `persisted_state`. This
  matches what `d<N>` already does, and the daemon's live-tmux probe is the
  authoritative gate on the actual delete, so it is not a finding. The
  consequence worth knowing: after a `tmux kill-server` or a reboot every
  session displays `stopped`, and `d *` — a blessed pattern, asserted by
  `glob_decide_match_all_never_reaches_running_sessions` — would then propose
  the entire fleet. The preview does print every row, which is why this is a
  note and not a finding; MEDIUM #1 above is the reason it still deserves a
  stronger confirmation.
- Name-based self-comparison at line 288 is exact while the emptiness check
  trims. Two records sharing a reused tmux name both resolve to `SelfSession`,
  which is the safe direction, so the asymmetry has no reachable bad outcome.

APPROVE — zero CRITICAL, zero HIGH. The three MEDIUMs are worth landing before
merge; none of them blocks.

Issue #5502, opened 2026-08-11
feat(trusty-audit): scaffold new workspace crate — headless library + thin CLI


bobmatnyc:
New workspace crate `trusty-audit` — owner: "the auditor client is really an
installer runner" (context: #5477/#5473's epic tree). This issue is scaffold
only: crate skeleton + conventions, no download/selection logic.

🔴 **Owner-settled, permanent: headless first, CLI-testable always.**
"ok on trusty-audit headless" — and stronger: "in fact you should be able to
test all features via cli." `trusty-audit` ships as a LIBRARY plus a thin CLI
for this milestone; the Tauri shell is a later milestone and must call the
same API the CLI does. No feature may exist only behind the GUI — for any
capability, there must be a CLI invocation that exercises it end to end with
no window launched. This is a design constraint on the crate, not a phase
that lapses once the GUI ships.

🔴 **Owns a working directory; layout is designed now, not retrofitted.**
Owner: "the package should save state" (amends the earlier stay-open/restart
choice recorded on #5494). The directory holds downloaded pinned tools,
cloned repos, the tga extract database, and run-state — see #5494 for what
already persists naturally vs. what is genuinely new (run progress).
Open questions to record, not resolve here: where it lives (alongside an
unzipped package? under the user's home?), whether it's safe to delete
outright, and what happens if two runs share one. It holds the recipient's
cloned SOURCE and their extract database, so it is a data-handling surface —
the eventual README must state what is written where and how to delete it
cleanly.

🔴 **Reads the companion `manifest.toml` rather than duplicating it.**
`tga audit` already writes `manifest.toml` (DOC-67 §6) into its output
directory before rendering the report
(`crates/trusty-git-analytics/src/commands/audit.rs:158-176`) — engagement
metadata plus one `[[repositories]]` entry per configured repo
(`crates/trusty-git-analytics/src/report/dd_manifest.rs:112-136`). Confirmed
by inspection: it is written ONCE, after the whole sweep completes, and
carries the configured repo set, not per-repo/per-phase completion status
(`AuditSweepStats.outcomes` is stage-level, not persisted structurally —
`crates/trusty-git-analytics/src/audit/stage.rs:140-147`). `trusty-audit`
reads it for engagement metadata and the configured repo set; the crate's own
run-state record covers only what the manifest doesn't (full detail on
#5494).

🔴 **Must not duplicate download logic.** Consumes `trusty-installer`'s
[...]

Issue #5499, opened 2026-08-11
feat(tga): auditor client — assemble & return the unencrypted deliverable zip


bobmatnyc:
Final client responsibility: once the sweep completes, assemble the return package from #5479 (extract database), #5481 (ed25519 manifest signing), and #5478 (engagement config/metadata), and offer it to the recipient to send back.

🔴 The return zip is unencrypted, same posture as the non-encrypted engagement config — the recipient can inspect exactly what they send back before doing so, same as they can inspect what runs. Deliberate, not an omission.

Boundary: #5483 is the OUTBOUND zip's structure and contents (what arrives at the recipient); this issue is the client logic that builds the RETURN package (what leaves their network). Keep the boundary clear so the two do not overlap.

Part of #5477.

Confidence: Observed (owner directive: non-encrypted config for transparency, quoted in #5473) — accepted feature work.

Closure conditions:
- [ ] Client assembles a return zip from #5479/#5481/#5478's outputs once the run completes
- [ ] The return zip is unencrypted and human-inspectable
- [ ] The client surfaces the finished zip's location (or a save dialog) for the recipient to send back

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
The return channel is email/manual per [#5642](https://github.com/bobmatnyc/trusty-tools/issues/5642), so the assembled package must be a single self-contained file with no dependency on any service we run. Its size is a design constraint tracked by [#5480](https://github.com/bobmatnyc/trusty-tools/issues/5480).

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
Correction: the size constraint referenced here (via [#5480](https://github.com/bobmatnyc/trusty-tools/issues/5480)) is removed — the owner lifted the 25 MB ceiling that had been recorded on [#5642](https://github.com/bobmatnyc/trusty-tools/issues/5642). A large package can go out as a Drive attachment.

The self-contained-file requirement **stands** — it follows from manual transfer (no service dependency), not from size. So: single self-contained file, still required; a specific size ceiling to design against, no longer required.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
[...]

Issue #5491, opened 2026-08-11
feat(trusty-installer): pinned-version, fail-closed prebuilt install entry point


bobmatnyc:
The auditor client (issue tracking this: filed alongside) must download `tga`, `trusty-analyze`, `trusty-review`, and `gh` at **pinned exact versions** and fail closed on a checksum/version mismatch. `trusty-installer`'s existing prebuilt-download machinery cannot serve that today.

Public building blocks already exist and are reusable as a library: `platform::current_target` (`crates/trusty-installer/src/download/platform.rs:32-52`), `release::asset_url`/`sha256_url`/`asset_filename` (`crates/trusty-installer/src/download/release.rs:89-117`), `fetch::download_and_verify` (`crates/trusty-installer/src/download/fetch.rs:75-110`, SHA-256 checked, `Err` on mismatch). `trusty-installer` ships a `[lib]` target (`crates/trusty-installer/Cargo.toml:18-20`) and `pub mod download;` (`crates/trusty-installer/src/lib.rs:17`), so these are already externally callable.

🔴 The one orchestrating entry point, `try_install_prebuilt` (`crates/trusty-installer/src/download/mod.rs:80-136`), is the wrong shape for a pinned consumer on two counts: it resolves `latest` only, via the `pub(crate)`-gated `release::resolve_latest_tag_from_url` path (`release.rs:193-208`, no pinned-version resolver exists), and its own doc comment (`mod.rs:75`) says it returns `Outcome::Fallback` — falling back to `cargo install` — "on any failure (network, 404, SHA mismatch, I/O)". That is fail-OPEN; a pinned consumer needs fail-CLOSED on a version or checksum mismatch, never a silent `cargo install` of whatever `latest` happens to be.

A new public entry point (e.g. `try_install_pinned_version(crate_name, version, install_dir)`) is additive — `try_install_prebuilt`'s existing signature and behavior are untouched, so this is not a SemVer break for `trusty-installer`.

Consumers: the auditor client (tga/trusty-analyze/trusty-review, and — per #5476 — `gh`).

Confidence: Reproduced (file:line citations above) — accepted feature work.

Closure conditions:
- [ ] A public `trusty-installer::download` entry point installs a caller-specified exact version and never falls back to `latest`
- [ ] On checksum or version mismatch it returns an error/fail-closed outcome, not `Outcome::Fallback` to `cargo install`
[...]

Issue #5477, opened 2026-08-11
epic(tga,trusty-installer): auditor client — standalone Tauri app that downloads its tools and drives the audit workflow


bobmatnyc:
Owner spec, verbatim: "the local app should be a lightweight http server that allows us to install dependencies, track progress on the cloning, give simple guidance via web forms, and track progress on the auditing, and build the deliverable unencrypted zip for returning to us" — later revised twice by the owner (below). Five responsibilities stand; the delivery shape does not.

🔴 **Shape, final: a standalone Tauri app, not an HTTP server.** Owner confirmed "should we just do as a self contained tauri app?" Binaries are opaque either way — compiled either way — so Tauri costs nothing on transparency. Transparency was always about the readable engagement config (what the tool will do) and the unencrypted return zip (what leaves the recipient's network); neither changes. Tauri gains one signable/notarizable artifact instead of a binary-plus-HTML-plus-localhost-server the recipient must trust is listening only locally, no port juggling, and double-click-to-native-window. macOS arm64 only removes most of Tauri's cross-platform cost. Prior art in this workspace: `crates/trusty-agents/ui/src-tauri`, `trusty-mpm-gui`, `trusty-code-gui`.

🔴 **Owner: "the auditor client is really an installer runner."** This decides where code lives:
- The INSTALL leg belongs to `trusty-installer`, consumed as a library (#5491, #5495, #5476) — platform detection, pinned-version fetch, SHA-256 verification are `trusty-installer`'s, not reimplemented here.
- The WORKFLOW leg (orchestrating an hour-long multi-repo run, streaming progress, assembling and signing a return package) does not migrate into `trusty-installer`. A separate client crate depends on `trusty-installer` for downloads and owns the rest.
- 🟡 This crate-boundary call is the PM's working decision, reversible by the owner. `tctl audit` (a `trusty-installer` subcommand) is the defensible alternative — it trades independent release cadence for fewer moving parts. Reasoning for the split as filed: `trusty-installer` is published on its own release cadence and an hour-long audit GUI shouldn't couple to it; the client is macOS-arm64-only Tauri where `tctl` is a cross-platform CLI; installing binaries and auditing codebases are different domains (the same boundary the owner applied to contributor profiling this session — profiling belongs to `tga`, not `trusty-review`).

[...]

PR #5540, opened 2026-08-11
feat(trusty-audit): download and verify the pinned tool triple (#5495)

### LINKED ISSUE #5495: feat(tga): auditor client downloads + verifies pinned tga/trusty-analyze/trusty-review
Owner: "the auditor client is really an installer runner." The install leg belongs to `trusty-installer`, consumed as a library — this issue is the client's consumption of it, not a second download implementation. CLAUDE.md's "Common entry point, clean domain demarcation" rule applies directly: a second implementation of downloading-and-verifying-a-binary in this workspace is a defect.

Depends on #5491 (`trusty-installer` gaining a pinned-version, fail-closed public entry point — `try_install_prebuilt` today is `latest`-only and fail-opens to `cargo install` on any mismatch, per that issue's citations). This issue is client-side: call that entry point for `tga`, `trusty-analyze`, and `trusty-review`. `gh` follows the same mechanism but is tracked separately in #5476 (it has its own on-PATH skip-check and auth-flow nuances).

🔴 Version-skew requirement: this session closed a version-skew hole (#5454 / PR #5458) where a new `tga` paired with an old `trusty-review` silently produced a deterministic report and exited 0. A downloader that fetches "latest" reintroduces that exact class of bug from the install side. The exact version triple (`tga`, `trusty-analyze`, `trusty-review`) must be pinned in the engagement config (schema owned by #5478), readable alongside `key_env`, the model, and the audit window — consistent with the transparency premise. The report package records which versions produced it, strengthening the recomputation story against the shipped extract database. On a mismatch between pinned and downloaded version, the client fails closed — never a silent degrade.

Deferred, not v1: a company with an egress proxy blocking arbitrary binary downloads fails at install rather than at audit, in the first thirty seconds of using the client. Documented as a known risk; whether a fully-bundled fallback variant should exist for locked-down environments is a separate, deferred question.

Part of the auditor client epic (#5477).

Confidence: Observed (owner directive, quoted above) — accepted feature work.

Closure conditions:
- [ ] Client downloads and verifies `tga`, `trusty-analyze`, and `trusty-review` via #5491's pinned entry point — no bespoke download/verify code in the client
- [ ] All three versions are pinned in the engagement config (#5478) and the client fails closed on any mismatch between pinned and downloaded version
- [ ] The report package records the exact versions that produced it
- [ ] Egress-blocked/locked-down-network failure mode is documented as a known v1 risk, not solved

bobmatnyc:
Closes #5495. Part of #5477 / #5473, milestone 62.

`tools::install` was a seam returning `NotImplemented`, so the client could report which tools were missing and do nothing about it. It now calls `trusty-installer`'s pinned, fail-closed entry point (#5491 / PR #5517) for `tga`, `trusty-analyze` and `trusty-review`, and installs all three or none.

## Closure conditions

- **Installs via #5491's entry point, no bespoke download code.** `tools::install` builds `PinnedTool` values and calls `install_pinned_set`. There is no download, checksum, extraction or archive code in `trusty-audit`, and `reqwest` is deliberately absent from its manifest — see the installer note below.
- **All three pinned in the engagement config, fail closed on mismatch.** `[tools]` carries the triple as a bare version or with the artifact digest pinned as well. All three keys are required, so a partly-pinned config does not parse. `tools::verify` re-checks each installed version against its pin after the installer has already proved it.
- **The package records the versions that produced it.** `state/tool-versions.toml` holds the triple after a successful install; `taudit tools` reports it. The package assembly that carries the file is #5499 — this issue captures the fact, that one ships it.
- **Egress-blocked network documented as a known v1 risk.** README and the `tools` module docs state it: the recipient fails at install in the first thirty seconds rather than an hour into an audit, nothing retries through a proxy, and there is no bundled-binary fallback. `a_blocked_download_host_refuses_rather_than_proceeding` pins the behaviour against a non-routable TEST-NET-3 proxy.

## The blocker the smoke run found

Running the binary against the real published triple failed:

```
Error: installing trusty-analyze 0.9.0 failed; nothing was installed:
two tools in this set both install `LICENSE` to .../trusty-audit-work/tools/LICENSE
```

`fetch::extract_binaries` returns every regular file in the archive, and every release tarball in this workspace ships a `LICENSE`, so PR #5517's set-collision check refused every multi-tool set. The three-tool set could never install. Fixed in `trusty-installer` by placing executables only (`executables_among`, unix mode bits); a genuine two-binaries-one-name conflict stays an error. The fixture tarballs now ship a `LICENSE` too, so the existing suite has the production shape — without that it was green while the feature was unusable.

## Departures from the brief

- **One additive change to `trusty-installer`, not pure consumption.** `install_pinned_set` takes a `&reqwest::Client` with no public constructor, so a consumer had to build a second download client — which CLAUDE.md's common-entry-point rule names explicitly. `download::http_client()` is now public. Bumped 0.7.0 → 0.8.0 for the additive API, matching #5517's precedent for the same seam.
- **`Session::execute` is now `async`.** Blocking on a runtime inside a synchronous `execute` works for the CLI and panics inside the Tauri shell, which calls from an async context. `Command::InstallTools` has a CLI path (`taudit install`), so #5502's compile-enforced constraint holds — `every_command_variant_has_a_cli_invocation` covers six variants now.
- **One guard beyond the issue.** `workdir.rs` carried a `// #5495:` note that a pre-planted symlink at an area was inert until something installed through it. Installing is what makes it live, so `install` refuses a `tools/` area that is not a real directory.
- **No `code-critic` round.** Rung 5 asks for one; agents do not dispatch subagents. Owed before merge.

## Gates — rung 5

Verification is a security path, so this is rung 5: rung-4 gates plus `--include-ignored` coverage and explicit failure-path tests.

```
cargo fmt --check                                            clean
cargo check --workspace --all-targets                        clean
cargo clippy -p trusty-audit    --all-targets -- -D warnings  clean
cargo clippy -p trusty-installer --all-targets -- -D warnings clean

cargo test -p trusty-audit                    59 passed; 0 failed; 1 ignored
                                               3 passed (binary_names)
                                               4 passed; 2 ignored (install_fails_closed)
cargo test -p trusty-audit -- --include-ignored
                                              60 passed; 0 failed; 0 ignored
                                               6 passed; 0 failed; 0 ignored
cargo test -p trusty-installer               592 passed; 0 failed; 4 ignored
```

`cargo check --workspace` plus `cargo test -p trusty-audit` is rung 4's dependent leg: `trusty-audit` is `trusty-installer`'s only library consumer.

### Fail-closed branches proved by mutation

Each guard was removed and the suite re-run, so these are not tests that pass whatever the code does:

```
MUTATION: ensure_real_directory removed, version equality bypassed, record parse errors swallowed

test tools::tool_tests::install_refuses_a_symlinked_tools_area_before_any_download ... FAILED
test tools::tool_tests::a_version_that_drifts_from_the_pin_is_refused ... FAILED
test tools::tool_tests::a_malformed_record_is_an_error_not_an_empty_list ... FAILED
test result: FAILED. 57 passed; 3 failed
```

Restored: `test result: ok. 59 passed; 0 failed; 1 ignored`.

### Binary smoke run

Refusal, against the real release host:

```
$ taudit install
Error: tga 0.0.0-never-published is not a published stable release, so it could not be
installed; nothing was installed. Published stable versions: 2.11.0, ..., 2.16.0
exit=1
tools/  (empty)     state/  (empty)
```

Success, against real published versions:

```
$ taudit install
Installed and verified 3 tools:
  tga              2.16.0       .../tools/tga
  trusty-analyze   0.9.0        .../tools/trusty-analyze
  trusty-review    0.14.1       .../tools/trusty-review

$ tools/tga --version            → tga 2.16.0
$ tools/trusty-analyze --version → trusty-analyze 0.9.0
$ tools/trusty-review --version  → trusty-review 0.14.1
```

`state/tool-versions.toml` records all three; `tools/` holds the binaries and no `LICENSE`.

## Test boundary

This crate does not re-prove downloading. `trusty-installer`'s `download::pinned::tests` already drives every fail-closed arm against a loopback fixture server through the same entry point, and its `Endpoints` seam is `pub(crate)`. What `trusty-audit` adds on top — pin construction, the version re-check, the record, the symlink guard — is covered offline; the `#[ignore]`d tests cover the real release host and the blocked-egress case.


bobmatnyc:
## Verdict: APPROVE

Reviewed against `docs/reference/test-ladder-baseline.md` rung 5 and the fail-open check. All claims in the PR body were independently reproduced rather than taken on trust.

### Verified independently

**Gates** (all run fresh, not copied from the PR body):
- `cargo fmt --check` — clean
- `cargo check --workspace --all-targets` — clean
- `cargo clippy -p trusty-audit --all-targets -- -D warnings` — clean
- `cargo clippy -p trusty-installer --all-targets -- -D warnings` — clean
- `cargo test -p trusty-audit` — 59 passed, 1 ignored (+ 3 in `binary_names`, 4 passed/2 ignored in `install_fails_closed`)
- `cargo test -p trusty-audit -- --include-ignored` — 60 passed, 0 failed (6/6 in `install_fails_closed`, including the real-network `an_unpublished_pin_refuses_against_the_real_release_host` and the blocked-egress `a_blocked_download_host_refuses_rather_than_proceeding`)
- `cargo test -p trusty-installer` — 592 passed, 4 ignored
- `bash scripts/check_line_cap.sh` — 0 violations (`tools.rs` at 718 raw lines stays under the 500-SLOC cap once comments/blanks are excluded)
- `bash scripts/check_sld.sh` — 0 errors, 0 warnings
- `bash scripts/check_changelog_fragment.sh` — fragments present and valid for both `trusty-audit` and `trusty-installer`

**Mutation claims, reproduced by hand** (guard removed, named test re-run, guard restored, diff confirmed clean afterward):
- `ensure_real_directory` call removed from `tools::install` → `install_refuses_a_symlinked_tools_area_before_any_download` fails on the `matches!(err, AuditError::UnsafeToolsDir { .. })` assertion. Confirmed.
- `verify`'s version-equality arm bypassed (`Some(install) if install.version == pinned` → `Some(install)`) → `a_version_that_drifts_from_the_pin_is_refused` fails, install proceeds with the drifted `trusty-review` version. Confirmed.
- `read_record`'s TOML parse error swallowed (`?` → `.unwrap_or_default()`) → `a_malformed_record_is_an_error_not_an_empty_list` fails, a tampered `tool-versions.toml` reads back as an empty (i.e. "nothing installed") list instead of an error. Confirmed.

**A fourth mutation not claimed in the PR body, run to close the loop on finding #1:** removed the `executables_among` filter in `trusty-installer/src/download/pinned.rs` (`let placeable = executables_among(...)` → `let placeable = binary_names.clone()`) and re-ran `documentation_files_are_not_installed`. It reproduces the exact production failure verbatim: `two tools in this set both install \`LICENSE\` to .../LICENSE`. This is the bug the PR body describes finding in the smoke run, and the fix removes it.

### Finding #1 — `executables_among`'s unix-mode-bit filter

Sound for the archives this workspace actually produces. `.github/workflows/release.yml`'s packaging step `strip`s and `cp`s each named binary (inheriting the build's executable bit) and separately `cp`s `LICENSE` / `README.md` from the git checkout (inheriting their non-executable git-tracked mode, normally 644) — so the filter's binary/non-binary split lines up with how these specific tarballs are actually built, not by coincidence.

Two edge cases worth naming, neither risky enough to block:
- `extract_binaries` (unchanged by this PR) already drops directory entries and symlinks before they reach `executables_among` — filtering happens over `entry_type.is_file()` results only, so "a directory entry with the bit set" can't occur.
- A future archive that ships an executable helper script (some `+x` file that isn't the pinned binary) would pass the filter and land in `tools/`. It's not a security issue — it can never collide with a genuine binary name without tripping the existing two-tools-one-name collision check, and everything placed already passed the checksum/pin gate — but it is a gap in "is this a binary" if the archive shape ever changes. The `#[cfg(not(unix))]` fallback (keep everything) is honestly documented as inert today since the pinned path is macOS/Linux-only, and a Windows collision still fails closed per the code comment.

Disposition: **Fix here** was not warranted (no reproducible defect at >80% confidence) — noted for the PM as a **Promote**-eligible watch item only if a future release ever bundles an executable non-binary asset, not before.

### Finding #2 — `Session::execute` async

`cli.rs`'s `to_command` (`Verb` → `Command`) and `render` (`Outcome` → `String`) stay exhaustive matches over both 6-variant enums; `every_command_variant_has_a_cli_invocation` still asserts all six route correctly. `main.rs` now owns the runtime via `#[tokio::main]`, so the async change doesn't leak an executor requirement onto callers. #5502's CLI-testability invariant holds.

### Finding #3 — `download::http_client()` SemVer

Ran `bash scripts/check_semver.sh --crate trusty-installer` directly rather than trusting the characterization. Registry baseline is 0.5.0 (the last published release); current is 0.8.0. The gate reports this range as "already a breaking release" (0.x MINOR already moved), so it runs advisory rather than blocking, and the three failures it lists (`SignTargetArg` enum variants, `service_bootstrap` parameter-count changes) are all in files this PR never touches — pre-existing from earlier merged work, not introduced here. This PR's own delta (`http_client()`) is a genuinely new public item with no removed or altered public signature; the private `build_client` it replaces was never part of the public API. The MINOR bump (0.7.0 → 0.8.0) matches this workspace's convention that PATCH is reserved for pure bug fixes and MINOR covers any public-surface growth in a 0.x crate — correctly classified, not a hazard.

### Finding #5 — `SecretKey` leak surface

`openrouter_key` is never `.expose()`d anywhere outside `config.rs`'s own tests — this PR adds no consumer of the key at all. `EngagementConfig` derives `Debug`, which routes through `SecretKey`'s hand-written redacting `Debug`/`Display`, and `SecretKey` has no `Serialize` impl, so it cannot reach a serialized artifact. Deferring the `openrouter_key` vs. `key_env` rename to #5478 is reasonable: nothing in this PR reads, logs, or serializes the key.

### Mechanical gates

- Changelog fragments: `crates/trusty-audit/changelog.d/5495-{pinned-tool-install,symlinked-tools-area,tools-reports-versions}.md` and `crates/trusty-installer/changelog.d/5495-{no-doc-files-in-install-dir,public-download-client}.md` — both crates covered, confirmed by `check_changelog_fragment.sh`.
- SLOC caps: confirmed clean via `check_line_cap.sh`.
- `unwrap()`/`expect()` audit: every occurrence in the diff is inside `#[cfg(test)]` modules or is the single pre-existing, already-documented programmer-error invariant in `install_pinned` (`crates/trusty-installer/src/download/pinned.rs:361`, unrelated to this diff). None in the new production paths.
- `thiserror`/`anyhow` split: `trusty-audit`'s `[lib]` target uses `thiserror` (`error.rs`); its `main.rs` binary uses `anyhow`. Correct per convention.
- Why/What/Test docs: present and proportional on every new public item touching the verification surface (`install`, `verify`, `ensure_real_directory`, `executables_among`, `http_client`).

No CRITICAL or HIGH findings at >80% confidence. **APPROVED for next pipeline stage.**

PR #5541, opened 2026-08-12
fix(trusty-common): propagate a stat failure in list_palaces instead of reporting no palaces

### LINKED ISSUE #5532: investigate(trusty-common): does a TCC denial short-circuit list_palaces at Path::exists() before read_dir?
UNVERIFIED HYPOTHESIS, not a confirmed defect — nobody has reproduced a real TCC denial against these code paths. Filing to track the question, not the fix.

`PalaceRegistry::list_palaces` in `crates/trusty-common/src/memory_core/store/palace_store.rs` (`pub fn list_palaces(registry_dir: &Path) -> Result<Vec<Palace>>` at line 172) guards on `Path::exists()` before ever calling `read_dir`: `if !registry_dir.exists() { return Ok(Vec::new()); }` at lines 173-175, then `std::fs::read_dir(registry_dir).map_err(...)` at lines 176-177.

Three fixes — #5511/#5488's `merge_palaces` (merged, `kg_twin_merge.rs`) and PR #5526's `purge_palaces` / `rebuild_palaces` (open, `kg_rebuild.rs`) — all changed `list_palaces(...).unwrap_or_default()` to propagate the error instead of silently returning an empty list. Every one is justified by a macOS TCC denial as the reachable trigger, and every one is tested by placing a REGULAR FILE at `data_root`, which makes `read_dir` return `ENOTDIR`.

`ENOTDIR` and a TCC-driven `EPERM` do take the identical path once inside `read_dir` — one `map_err` handles both — so the existing tests are not worthless. The edge they cannot reach: if a TCC denial also blocks `stat()`, `Path::exists()` at line 173 could itself return `false`, taking the early `Ok(Vec::new())` at line 174 and never reaching `read_dir` at all. In that case the propagation all three fixes added is never exercised, and the operator gets exactly the empty-run behavior the fixes exist to prevent.

Raised by code-critic during PR #5526's review (https://github.com/bobmatnyc/trusty-tools/pull/5526) as a plausible-but-unverified gap, explicitly not a blocker for that PR.

Ask: determine, against actual macOS TCC behavior, whether `Path::exists()` short-circuits (returns `false`) under a TCC-denied directory, or whether it errors/returns `true` and proceeds to `read_dir` as normal. The entire justification for three separate fixes rests on a trigger nobody has exercised end to end — verify the premise before more code is built on top of it.

Closure condition: either (a) evidence — from a real TCC-denied directory on macOS, or authoritative documentation of `Path::exists()`/`stat()` under TCC denial — that `read_dir` is genuinely reached and the existing fixes are sufficient (close as verified), or (b) a fix restructuring the `exists()`-then-`read_dir` sequence plus a test covering the short-circuit arm.

bobmatnyc:
Closes #5532

**Lead with the measurement, because it partly refutes the issue that prompted the fix.**

#5532 asked whether a macOS TCC denial short-circuits `PalaceRegistry::list_palaces` at `Path::exists()` before reaching `read_dir` — which, if true, would mean the fail-open fixes in #5488 and #5526 never fire for their stated trigger. Measured on macOS 26.5.2 as uid 502 against real TCC-protected directories:

| Path | `exists()` | `metadata()` | `read_dir()` |
|---|---|---|---|
| `~/Library/Safari` | true | Ok | Err(PermissionDenied, raw=1) |
| `~/Library/Mail` | true | Ok | Err(PermissionDenied, raw=1) |
| `/Library/Application Support/com.apple.TCC` | true | Ok | Err(PermissionDenied, raw=1) |
| parent `chmod 000` (POSIX) | **false** | Err(PermissionDenied, raw=13) | Err(raw=13) |

Genuinely TCC, not POSIX: those directories are `drwxr-xr-x` and world-readable, yet enumeration fails. **TCC denies with EPERM(1) at the operation; POSIX denies with EACCES(13) at `stat`.**

So the hypothesis is refuted for its named trigger — under TCC, `read_dir` is always reached and #5488/#5526 fire as designed. State that plainly: **those three merged fixes are verified sound.**

**What survives, and is what this PR fixes.** `Path::exists()` is `fs::metadata(self).is_ok()` (std source, `library/std/src/path.rs:3429-3431`, toolchain 1.94.1); its docs say it coerces permission errors to `false` and recommend `try_exists()`. So a POSIX `EACCES` from an unsearchable parent collapses to `false` and `list_palaces` returns `Ok(vec![])` — a fail-open feeding destructive passes.

**The fix:** the guard uses `try_exists()`. Boundary verified case by case, not assumed:

| Case | `try_exists()` | Behaviour |
|---|---|---|
| absent, any depth of missing parent | `Ok(false)` | empty list — first run unchanged |
| broken symlink | `Ok(false)` | empty list |
| regular file at the path | `Ok(true)` | falls to `read_dir` → ENOTDIR — existing #5488/#5526 harnesses unaffected |
| TCC-denied dir | `Ok(true)` | falls to `read_dir` → EPERM — those fixes still fire |
| `EACCES` from locked parent | `Err` | **propagates** |

Say explicitly that first-run behaviour is untouched — `Ok(false)` covers every genuinely-absent shape including missing parents.

**Test ladder: rung 5.** Red, pre-fix guard with the new test:
```
panicked at crates/trusty-common/src/memory_core/store/palace_store.rs:393:14:
a registry dir that cannot be stat'd must not report as empty: []
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 1006 filtered out
```
Green after: `3 passed; 0 failed`.

The test strips a **parent** to mode 000 so the failure lands on `stat`. Two hazards handled — it asserts the denial actually took hold (`metadata()` must return `PermissionDenied`) so it cannot pass vacuously under root, which bypasses permission checks; and it restores the mode from a `Drop` guard so a panicking assertion cannot strand an undeletable temp dir.

**Downstream composition deliberately not re-tested:** the change makes `EACCES` produce the same `PalaceStoreError::Io` variant `ENOTDIR` already produced, and the three existing tests (`purge_palaces_propagates_an_unreadable_data_root` and siblings) already prove purge/rebuild/merge propagate that error. A fourth test would re-prove the second half with a more fragile cross-crate harness.

**Version:** `trusty-common` 0.33.0 → **0.33.1**. Patch — `list_palaces`'s signature is unchanged; only its behaviour on a permission error changes. crates.io `max_version` was 0.33.0, so the `#4421` gate required it. Eight `trusty-common` requirement sites enumerated; only the crate's own `version` field needed changing, as every consumer carries a caret `"0.33"` that already admits `0.33.1`.

**Gates:**
```
cargo test -p trusty-common --features memory-core           EXIT=0  (1017 passed, 0 failed)
cargo check --workspace                                      EXIT=0
clippy 0.1.97, CI's workspace invocation                     EXIT=0
cargo test -p trusty-memory  (rung-4 dependent)              EXIT=0  (729 + 27 targets, 0 failed)
check_line_cap / check_changelog_fragment / check_sld        EXIT=0
check_test_pointers.sh                                       EXIT=0  (23835 citations, 0 dangling)
```

**One gate red and proven pre-existing:** `cargo test -p trusty-common --features memory-core -- --include-ignored` fails `cold_restart_recalls_beyond_l1_snapshot` and `timeout_fires_on_embedder_init_with_tiny_limit`, both ONNX/embedder-backed. Proven not-ours by reverting `palace_store.rs` to base `c898ba67a` and re-running those two — identical `0 passed; 2 failed`. The diff touches two files; the failures live in `retrieval/tests.rs` and `retrieval/timeout_tests.rs`, and `retrieval/handle.rs` never calls `list_palaces`. Closest prior art is #852, same class.

Also note: `Cargo.lock` carries one unrelated resolver-slack flip (`proc-macro-crate` 1.3.1 vs 2.0.0 for `num_enum_derive`, both semver-compatible), an automatic `cargo check` side effect, left as-is rather than hand-edited.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6535,7 +6535,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "680998035259dcfcafe653688bf2aa6d3e2dc05e98be6ab46afb089dc84f1df8"
 dependencies = [
- "proc-macro-crate 2.0.0",
+ "proc-macro-crate 1.3.1",
  "proc-macro2",
  "quote",
  "syn 2.0.117",
@@ -11410,7 +11410,7 @@ dependencies = [
 
 [[package]]
 name = "trusty-common"
-version = "0.33.0"
+version = "0.33.1"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/crates/trusty-common/Cargo.toml
+++ b/crates/trusty-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trusty-common"
-version = "0.33.0"
+version = "0.33.1"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
--- a/crates/trusty-common/changelog.d/5532-list-palaces-stat-guard.md
+++ b/crates/trusty-common/changelog.d/5532-list-palaces-stat-guard.md
@@ -1,0 +1,6 @@
+Fixed
+
+- `PalaceStore::list_palaces` no longer reports a registry it cannot stat as an empty one ([#5532](https://github.com/bobmatnyc/trusty-tools/issues/5532))
+  - The absence guard used `Path::exists`, which is `fs::metadata(..).is_ok()` and coerces every stat failure — `EACCES` from an unsearchable parent directory, `EIO`, `ELOOP` — to `false`. The function then returned `Ok(vec![])` without reaching `read_dir`, so the error propagation added in #5488 and #5526 never ran and the destructive callers (`purge_palaces`, `rebuild_palaces`, `merge_palaces`) reported a clean zero-palace run over data they never read.
+  - It now uses `try_exists`, which keeps "absent" (`Ok(false)`) distinct from "cannot determine" (`Err`). A data root that does not exist yet — including one whose parents are missing, and a broken symlink — still returns an empty list, so first run is unchanged.
+  - A macOS TCC denial is NOT this trigger: measured against real TCC-protected directories, TCC permits `stat` and denies only enumeration, so `read_dir` was always reached and the #5488/#5526 fixes fire as intended.
--- a/crates/trusty-common/src/memory_core/store/palace_store.rs
+++ b/crates/trusty-common/src/memory_core/store/palace_store.rs
@@ -168,9 +168,26 @@ impl PalaceStore {
     /// loads and pushes the `Palace`. Subdirs without metadata are silently
     /// skipped.
     /// Test: `list_palaces_finds_saved_palaces` (registry tests cover the full
-    /// stack).
+    /// stack); `list_palaces_missing_dir_is_empty` pins the absent-root arm and
+    /// `list_palaces_propagates_an_unstattable_registry_dir` the denied one.
+    ///
+    /// The absence guard uses `try_exists` rather than `exists`, because
+    /// `Path::exists` is `fs::metadata(..).is_ok()` and so coerces *every* stat
+    /// failure to `false` — a registry we are forbidden to stat would read as a
+    /// registry that is not there, and callers on the destructive paths
+    /// (`purge_palaces`, `rebuild_palaces`, `merge_palaces`) would report a
+    /// clean zero-palace run over data they never read. `try_exists` keeps
+    /// "absent" (`Ok(false)`, including a broken symlink or any missing parent)
+    /// distinct from "cannot determine" (`Err`), so only the former still
+    /// yields an empty list. A first run against a data root that does not
+    /// exist yet is normal and must not error.
     pub fn list_palaces(registry_dir: &Path) -> Result<Vec<Palace>> {
-        if !registry_dir.exists() {
+        // #5532: `exists()` reports a stat we are denied as "absent", silently
+        // turning an unreadable registry into an empty one.
+        let present = registry_dir
+            .try_exists()
+            .map_err(|e| PalaceStoreError::io(registry_dir, e))?;
+        if !present {
             return Ok(Vec::new());
         }
         let read_dir =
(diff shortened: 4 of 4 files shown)


bobmatnyc:
## Verdict: WARN

Adversarial review of `28ad922f` (base `c898ba67`), crate `trusty-common`.

### 1. First-run regression — verified negative

Measured `Path::try_exists()` directly against real filesystem shapes (non-root, `rustc 1.94.1`, this repo's MSRV toolchain):

| Shape | Result |
|---|---|
| absent path, existing parent | `Ok(false)` |
| absent path, missing intermediate parent | `Ok(false)` |
| dangling symlink | `Ok(false)` |
| parent exists but is mode 000 (no traverse), child absent | `Err(PermissionDenied)` |

The doc comment at `crates/trusty-common/src/memory_core/store/palace_store.rs:180` claims "absent (`Ok(false)`, including a broken symlink or any missing parent)" — that claim is true for both cases actually tested (missing parent, dangling symlink) and does not over-claim about the fourth row. The fourth row (parent present but access-denied) does return `Err`, but this is fail-closed by design intent, not the first-run path — a fresh install's immediate parent (home dir, or an app-data dir the installer created) is not normally access-denied. `list_palaces_missing_dir_is_empty` (pre-existing, `palace_store.rs:324`) already pins the real first-run shape and still passes. No regression.

### 2. New test — verified non-vacuous

Ran the suite in a non-root shell (`whoami` → `masa`) so the root-bypass guard inside the test didn't trip:

```
cargo test -p trusty-common --features memory-core --lib memory_core::store::palace_store::
running 6 tests
test memory_core::store::palace_store::tests::list_palaces_propagates_an_unstattable_registry_dir ... ok
...
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 1003 filtered out
```

Then reverted only the guard (`try_exists` → `exists`, keeping the test) and reran that one test:

```
test memory_core::store::palace_store::tests::list_palaces_propagates_an_unstattable_registry_dir ... FAILED
panicked at .../palace_store.rs:393:14:
a registry dir that cannot be stat'd must not report as empty: []
```

Confirmed red without the fix, green with it. Restored the file afterward — worktree is clean, no leaked mode-000 directories (the `Drop`-based `RestoreMode` guard did its job).

One thing worth checking before merge: this repo's `ci.yml` test job runs on bare `ubuntu-latest` (no `container:`), so it executes as the non-root `runner` user — the test's root-bypass panic (`palace_store.rs:384`) will not fire in CI. The only containerized jobs (`al2023-build.yml`, `release.yml`) don't run `cargo test`. So this is safe, but it's worth being aware the test would hard-fail (not skip) if it ever runs inside a root container.

### 3. `try_exists()` vs `metadata()` + `kind()` match

`try_exists()`'s std implementation is exactly `metadata().map(|_| true).or_else(|e| if e.kind() == NotFound { Ok(false) } else { Err(e) })` — the built-in equivalent of the hand-rolled alternative, not a different-behaving shortcut. Right instrument; doc comment is accurate.

### 4. Fail-open sweep — found a live instance, out of this diff's scope

`purge_palaces` / `rebuild_palaces` / `merge_palaces` (in `trusty-memory`'s `kg_rebuild.rs` and `kg_twin_merge.rs`) all route through `PalaceRegistry::list_palaces` → `PalaceStore::list_palaces`, and all three already have their own `propagates_an_unreadable_data_root`-style regression tests from #5488/#5526/#5511. The chain to this PR's fixed guard is a straight passthrough (`registry.rs:587-590`, `.with_context` only, no swallowing). That part is clean.

But the loop this PR's guard feeds directly into is unchanged and carries the same shape three more times:

|---|---|---|---|---|---|
| HIGH | crates/trusty-common/src/memory_core/store/palace_store.rs | 211 | `match Self::load_palace(&path) { ... Err(_) => continue }` discards every error from `load_palace` — not just "no palace.json," but a genuine permission-denied or I/O failure reading an existing `palace.json`. A single palace whose metadata file becomes unstattable silently vanishes from every `list_palaces` result, exactly the class of bug #5532 fixes one level up. | Distinguish `PalaceStoreError::NotFound` (skip, correct) from `Io`/`Json` (propagate, or at minimum collect and surface as a partial-failure report) instead of blanket `Err(_) => continue`. | Promote |
| HIGH | crates/trusty-common/src/memory_core/store/palace_store.rs | 206 | `if !path.join(PALACE_JSON).exists() { continue; }` — `Path::exists()` coerces a stat failure (e.g. permission-denied on this one subdir) to `false`, so a single inaccessible palace directory is silently treated as "not a palace" and skipped, same coercion this PR just removed at the registry root. | Swap for `try_exists()` and propagate `Err` (or collect it) rather than treating it as absent. | Promote |
| MEDIUM | crates/trusty-common/src/memory_core/store/palace_store.rs | 203 | `if !path.is_dir() { continue; }` has the same `exists()`-style coercion (`is_dir()` returns `false` on stat failure, not just "is a file"). Lower confidence this fires independently of line 206 in practice, since both trip on the same stat call, but it's the same shape. | Same treatment as line 206, or fold in if 206 gets a `try_exists`-based rewrite. | Promote |

These three are unchanged code — this PR doesn't touch the loop body — so none of them are a defect *introduced* by `28ad922f`, and none block this specific fix (the top-level registry guard is what #5532 targeted, and it is now correctly fixed and tested). Flagging because the loop these findings sit in is the direct caller of the line this PR just fixed, and it's the same fail-open family (#5488/#5526/#5511/#5532) recurring a level deeper. Recommend a follow-up issue; not a blocker for this PR.

### 5. Version bump

`list_palaces`'s signature is byte-identical before/after (`pub fn list_palaces(registry_dir: &Path) -> Result<Vec<Palace>>`) — only its internal error path changed. Diffed the full changeset; no other `pub` item's signature moved. `0.33.0` → `0.33.1` is the correct patch bump for a 0.y.z crate under Cargo's semver rules (ADR-0029 / #4088 precedent about the MINOR-position breaking rule doesn't apply — nothing here is breaking).

### Known, not scored

- `--include-ignored` ONNX/embedder failures: pre-existing per the stated base-commit revert; accepted as-is.
- `Cargo.lock`'s `proc-macro-crate 2.0.0` → `1.3.1` flip is the only unrelated line in that file (plus the version bump itself) — confirmed by diffing `Cargo.lock` directly; consistent with an unattended `cargo check` resolver-slack flip, not hand-edited.

### Notes

- Both `clippy -p trusty-common --features memory-core --lib -- -D warnings` and the full `palace_store` test module (6/6) are green on the PR head as posted.

bobmatnyc:
## Scoped verdict: APPROVE

Follow-up to the WARN above, per PM request, to give this merge a clean auditable gate.

The WARN rested entirely on three findings at `palace_store.rs:203/:206/:211` — the same `exists()`/blanket-`continue` coercion this PR fixes at the registry root, recurring one level deeper in the `list_palaces` loop body. That loop is **unchanged by this diff** (the PR touches only the guard at lines 184-194 plus the new test module) and those three are being filed as their own tracked issue, closure condition an error-arm test per site.

With those excluded from this PR's scope: items 1 (first-run safety), 2 (test validity), 3 (`try_exists` vs `metadata`+`kind`), and 5 (version bump) all closed with zero findings at any severity, verified directly (standalone `try_exists()` probe, revert-and-rerun on the new test, clippy clean). Nothing remains above the 80% confidence floor for the code this PR actually changes.

**Verdict on this PR's own changes: APPROVE.** The WARN stands as a systemic observation on the surrounding loop, tracked separately, not a gate on this fix.

PR #5545, opened 2026-08-12
fix(trusty-common): propagate filesystem and decode failures in the list_palaces walk

### LINKED ISSUE #5543: fix(trusty-common): list_palaces loop coerces per-entry stat/load failures into silent skips
`PalaceRegistry::list_palaces` (`crates/trusty-common/src/memory_core/store/palace_store.rs`, `pub fn list_palaces`) can return a shorter list than reality while reporting success. Its callers include destructive passes — `purge_palaces`, `rebuild_palaces`, `merge_palaces` — so a palace that silently drops out of the listing is one those passes skip while reporting a clean run.

Three coercions in the per-entry loop, all confirmed against `origin/main`:

- Line 194 — `Err(_) => continue` on a `load_palace` failure. Any error (permission-denied, truncated JSON, I/O) becomes "not a palace," not just a missing-metadata case.
- Line 189 — `!path.join(PALACE_JSON).exists()` coerces a stat failure to "no metadata file" and skips.
- Line 186 — `!path.is_dir()` has the same `exists()`-style coercion.

This is the same class as [#5488](https://github.com/bobmatnyc/trusty-tools/pull/5488), [#5511](https://github.com/bobmatnyc/trusty-tools/issues/5511), [#5526](https://github.com/bobmatnyc/trusty-tools/pull/5526), and [#5532](https://github.com/bobmatnyc/trusty-tools/issues/5532)/[#5541](https://github.com/bobmatnyc/trusty-tools/pull/5541) — one level deeper, and it survives all of those fixes. Raised by code-critic during #5541's review: https://github.com/bobmatnyc/trusty-tools/pull/5541#issuecomment-5260363342

Each prior instance was found only because someone looked at the site next door. Nothing mechanical catches a failure downgraded to a benign default on a path feeding destructive work — that's why this is its own issue rather than a fix-in-place.

Open design question for whoever picks this up: should one corrupt palace fail the whole listing, or be reported-and-skipped? Skipping may be the right call among many palaces — but not silently. Write the reasoning down with the fix.

Closure conditions:
- An error-arm regression test per site (three total) that fails against the pre-fix commit and passes after — this repo's fail-open check. Reuse the harness: #5541 already added `list_palaces_propagates_an_unstattable_registry_dir` in this file, and #5526 built the `Drop`-guard mode-000 pattern.
- The design decision above stated in the PR, not assumed.

Confidence: Reproduced — code-critic exercised the red/green harness pattern this same file already uses; these three sites are the same shape, unverified against a live TCC/permission denial but structurally identical to the fixed sibling at line 173.

bobmatnyc:
Closes #5543

**The defect.** Four sites in `PalaceRegistry::list_palaces`'s per-entry loop coerced failures into "not a palace" and continued, so the function returned a SHORTER list than reality and reported success. Callers include the destructive `purge_palaces`, `rebuild_palaces`, `merge_palaces` — a palace that silently vanishes from the listing is one those passes skip while reporting a clean run.

**The design call, and why it isn't "fail hard vs skip".** That framing dissolves once failures are separated by kind. Present it as a table:

| Failure | Kind | Disposition |
|---|---|---|
| entry `Err` mid-enumeration | filesystem-level | propagate |
| `is_dir()` stat denied | filesystem-level (the parent was just `read_dir`'d) | propagate |
| `palace.json` stat denied | filesystem-level | propagate |
| `load_palace` Io/Json | palace-local — the corrupt-palace case | propagate |
| no `palace.json`, non-dir, dangling symlink, unlinked mid-walk | genuinely not a palace | skip, unchanged |

The denial-of-service worry — one corrupt file making the whole registry unusable — applies only to the corrupt-palace row, and it is weaker than it looks: the current silent skip makes a palace **permanently invisible with no diagnostic**, while the error carries `PalaceStoreError::Json { path, .. }` naming the exact broken file. "You see 9 of 10 and never learn about the 10th" is the worse failure.

**Precedent that settled it, and state this prominently:** `crates/trusty-memory/src/bm25_backfill.rs:644` already contains a private reimplementation of this walk, written because `list_palaces` fails open, whose doc says *"An errored directory entry fails the whole enumeration rather than shrinking it, because a partial list is indistinguishable from a short one."* This fix removes the reason that duplicate exists.

**No public API change, deliberately.** An additive `list_palaces_partial` + `PalaceListing`/`PalaceSkip` was considered and rejected — no in-tree consumer wants it, and every public item on a published crate is a permanent SemVer commitment. A new error variant was also rejected: `PalaceStoreError` is not `#[non_exhaustive]`, so adding one would itself be breaking. `Io`/`Json` already carry the actionable path. **No version bump owed** — behaviour change, not API break.

**A fourth site, not in the issue.** `Err(_) => continue` on the `read_dir` iterator itself, same defect, same loop. Fixed here rather than left to become a sixth instance. State plainly that it has NO test: it needs a directory to fail mid-iteration after a successful open, which could not be driven deterministically.

**Test ladder: rung 5.** Three red→green pairs, raw:
```
an entry that cannot be stat'd must not drop out of the listing: []
a palace.json that cannot be stat'd must not read as a non-palace dir: []
an undecodable palace must not silently shorten the listing: [Palace { id: PalaceId("alpha"), ... }]
```
Pre-fix `FAILED. 1 passed; 3 failed`; post-fix `ok. 9 passed; 0 failed`.

Call out the third message specifically — `Ok([alpha])` is a **partial** listing reported as success, which is the issue's exact hazard caught live rather than an empty-list stand-in.

Both permission tests assert the denial took hold and panic with a run-as-non-root message otherwise (#5541's pattern), so neither can pass vacuously under root. Site 1 uses mode `0o400` on the registry — readable so `read_dir` works, non-traversable so the child stat is denied — which isolates it from the registry-level stat #5541 pins. Site 3 uses no permission bits, so it exercises under any uid.

**Also updated:** `crates/trusty-common/src/memory_core/registry.rs` — its `open()` doc claimed hydration "matches the resiliency choice in `PalaceStore::list_palaces`", no longer true; it now states enumeration is stricter than hydration.

**Gates:**
```
cargo test -p trusty-common --features memory-core       EXIT=0  (999 passed, 0 failed, 13 ignored, + 6 suites)
cargo test -p trusty-memory  (rung-4 dependent)          EXIT=0  (729 passed, 0 failed, + 26 suites)
cargo check --workspace                                  EXIT=0
clippy 0.1.97, CI's workspace invocation                 EXIT=0
check_line_cap / check_changelog_fragment / check_sld    EXIT=0
check_test_pointers.sh                                   EXIT=1  — pre-existing, not this diff
```

**On the pointer lint:** 14 dangling pointers, ALL in `crates/trusty-search`. `git diff --name-only origin/main...HEAD -- crates/trusty-search/` returns zero paths; this diff is three files, all under `crates/trusty-common`. One cited pointer is byte-identical at `origin/main`. Zero failures in `trusty-common`, which also confirms this PR's four new `Test:` pointers resolve. A sibling branch is already on it.

### DIFF
--- a/crates/trusty-common/changelog.d/5543-list-palaces-loop-coercions.md
+++ b/crates/trusty-common/changelog.d/5543-list-palaces-loop-coercions.md
@@ -1,0 +1,3 @@
+Fixed
+
+- `PalaceStore::list_palaces` no longer returns a short list and reports success when it cannot classify an entry. A denied stat on a registry child, a `palace.json` it cannot stat, an undecodable `palace.json`, and a directory entry that fails mid-enumeration now propagate an error naming the offending path instead of dropping that palace from the listing. Genuine absence — a subdirectory holding no `palace.json`, a dangling symlink, an entry unlinked mid-walk — still skips silently. The destructive callers (`purge_palaces`, `rebuild_palaces`, `merge_palaces`) already propagate, so they no longer report a clean pass over a palace they never read (#5543).
--- a/crates/trusty-common/src/memory_core/registry.rs
+++ b/crates/trusty-common/src/memory_core/registry.rs
@@ -605,7 +605,10 @@ impl PalaceRegistry {
     /// and for each persisted palace builds a `PalaceHandle` via
     /// `PalaceHandle::open` and registers it. Errors hydrating a single palace
     /// are logged and skipped so one corrupt palace doesn't take the whole
-    /// registry down — matches the resiliency choice in `PalaceStore::list_palaces`.
+    /// registry down. Enumeration is stricter than hydration: since #5543
+    /// `PalaceStore::list_palaces` fails rather than return a short list, so a
+    /// palace missing from this warmup was skipped by `PalaceHandle::open`, not
+    /// lost before it was ever seen.
     /// Test: `open_hydrates_persisted_palaces` exercises restart by writing,
     /// dropping, and reopening.
     pub fn open(data_root: &Path) -> Result<Self> {
--- a/crates/trusty-common/src/memory_core/store/palace_store.rs
+++ b/crates/trusty-common/src/memory_core/store/palace_store.rs
@@ -162,25 +162,37 @@ impl PalaceStore {
     /// palaces.
     ///
     /// Why: `palace list` and the daemon startup path both need to enumerate
-    /// every palace on disk. Skipping unreadable subdirs keeps a single broken
-    /// palace from taking the whole registry down.
-    /// What: For each immediate child directory containing a `palace.json`,
-    /// loads and pushes the `Palace`. Subdirs without metadata are silently
-    /// skipped.
-    /// Test: `list_palaces_finds_saved_palaces` (registry tests cover the full
-    /// stack); `list_palaces_missing_dir_is_empty` pins the absent-root arm and
-    /// `list_palaces_propagates_an_unstattable_registry_dir` the denied one.
+    /// every palace on disk, and the destructive passes (`purge_palaces`,
+    /// `rebuild_palaces`, `merge_palaces`) act on whatever this returns. A
+    /// short listing reads exactly like a complete one, so any palace dropped
+    /// here is one those passes skip while reporting a clean run.
+    /// What: For each immediate child directory holding a `palace.json`, loads
+    /// and pushes the `Palace`. A child that is genuinely not a palace is
+    /// skipped; a child that cannot be classified aborts the walk with the
+    /// underlying error, which names the offending path.
+    /// Test: `list_palaces_finds_saved_palaces` and
+    /// `list_palaces_missing_dir_is_empty` pin the benign arms;
+    /// `list_palaces_propagates_an_unstattable_registry_dir`,
+    /// `list_palaces_propagates_an_unstattable_entry`,
+    /// `list_palaces_propagates_an_unstattable_palace_json` and
+    /// `list_palaces_propagates_an_undecodable_palace` pin the error arms.
     ///
-    /// The absence guard uses `try_exists` rather than `exists`, because
-    /// `Path::exists` is `fs::metadata(..).is_ok()` and so coerces *every* stat
-    /// failure to `false` — a registry we are forbidden to stat would read as a
-    /// registry that is not there, and callers on the destructive paths
-    /// (`purge_palaces`, `rebuild_palaces`, `merge_palaces`) would report a
-    /// clean zero-palace run over data they never read. `try_exists` keeps
-    /// "absent" (`Ok(false)`, including a broken symlink or any missing parent)
-    /// distinct from "cannot determine" (`Err`), so only the former still
-    /// yields an empty list. A first run against a data root that does not
-    /// exist yet is normal and must not error.
+    /// Absence and failure-to-determine stay distinct at every step, which is
+    /// why the guards are `try_exists` and `metadata` rather than `exists` and
+    /// `is_dir`: the latter are `fs::metadata(..).is_ok()`-shaped and coerce
+    /// *every* stat failure to `false`, so a path we are forbidden to stat
+    /// reads as a path that is not there. Only genuine absence still skips
+    /// silently — `Ok(false)`, a dangling symlink, an entry unlinked mid-walk,
+    /// or a subdirectory that simply holds no `palace.json`. A first run
+    /// against a data root that does not exist yet is normal and must not
+    /// error.
+    ///
+    /// Failing the whole walk on one corrupt palace and skipping it are both
+    /// bad, and this picks failing (#5543): the error names the broken path, so
+    /// an operator can repair or remove it in one step, whereas a silent skip
+    /// makes that palace permanently invisible with no diagnostic while the
+    /// destructive callers quietly pass over it. `bm25_backfill`'s private
+    /// enumeration already made the same call for the same reason.
     pub fn list_palaces(registry_dir: &Path) -> Result<Vec<Palace>> {
         // #5532: `exists()` reports a stat we are denied as "absent", silently
         // turning an unreadable registry into an empty one.
@@ -195,20 +207,38 @@ impl PalaceStore {
 
         let mut palaces = Vec::new();
         for entry in read_dir {
-            let entry = match entry {
-                Ok(e) => e,
-                Err(_) => continue,
-            };
+            // #5543: an entry failing mid-enumeration truncated the walk.
+            let entry = entry.map_err(|e| PalaceStoreError::io(registry_dir, e))?;
             let path = entry.path();
-            if !path.is_dir() {
+
+            // #5543: `is_dir()` coerced a denied stat to "not a directory".
+            let meta = match std::fs::metadata(&path) {
+                Ok(m) => m,
+                // Unlinked mid-walk, or a dangling symlink: genuinely absent,
+                // the one arm `try_exists` also reports as `Ok(false)`.
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => continue,
+                Err(e) => return Err(PalaceStoreError::io(&path, e)),
+            };
+            if !meta.is_dir() {
                 continue;
             }
-            if !path.join(PALACE_JSON).exists() {
+
+            // #5543: `exists()` coerced a denied stat to "holds no metadata".
+            let metadata_path = path.join(PALACE_JSON);
+            let has_metadata = metadata_path
+                .try_exists()
+                .map_err(|e| PalaceStoreError::io(&metadata_path, e))?;
+            if !has_metadata {
                 continue;
             }
+
             match Self::load_palace(&path) {
                 Ok(p) => palaces.push(p),
-                Err(_) => continue,
+                // Raced away between the guard above and the read.
+                Err(PalaceStoreError::NotFound(_)) => continue,
+                // #5543: a denied read or an undecodable `palace.json` used to
+                // drop the palace and still report success.
+                Err(e) => return Err(e),
             }
         }
         Ok(palaces)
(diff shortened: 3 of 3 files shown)


bobmatnyc:
## Adversarial review — fix/5543-list-palaces-loop-coercions @ 6e228b710

## Verdict: APPROVE

Reviewed against commit `6e228b7106217dd32d2e735676adbdde7c4d6fe4` in an isolated worktree (main checkout untouched). Scope: the three changed files under `crates/trusty-common`.

## Partition check (the core question)

Walked the loop in `list_palaces` (`crates/trusty-common/src/memory_core/store/palace_store.rs:209-244`) against cases not in the test suite:

- **Dangling symlink** — `entry.path()` on it: `std::fs::metadata` follows the link, target missing → `ErrorKind::NotFound` → hits the `continue` arm at line 219. Skips, as claimed.
- **Path unlinked mid-walk** — same `metadata()` call, same `NotFound` → same skip arm. Correct.
- **Regular file / FIFO / socket entry** — `metadata()` succeeds (stat never blocks on a FIFO/socket), `meta.is_dir()` is `false` → skip at line 222-224. Correct; these can never be palaces.
- **Non-UTF8 filename** — `PathBuf` carries it fine through every call in this function (`metadata`, `join`, the error's `PathBuf` field); nothing here requires UTF-8. No panic path.
- **Empty directory, no `palace.json`** — `try_exists()` on the missing `palace.json` returns `Ok(false)` → skip at line 231-233. Correct.

All five match the claimed partition. One gap, not in that list:

**MEDIUM — `load_palace`'s own existence guard still coerces, one call past where this PR hardened the check.** `crates/trusty-common/src/memory_core/store/palace_store.rs:152` (unchanged by this PR):
```rust
pub fn load_palace(data_dir: &Path) -> Result<Palace> {
    let target = data_dir.join(PALACE_JSON);
    if !target.exists() {
        return Err(PalaceStoreError::NotFound(target));
    }
```
`list_palaces` already confirmed presence via `try_exists()` a few lines earlier (line 228-230, correctly propagates on denial) before calling `Self::load_palace(&path)` at line 235. But `load_palace` re-checks with the coercive `Path::exists()` — the exact `is_ok()`-shaped pattern this PR eliminates everywhere else. If a permission denial lands specifically in the window between the two calls (not a delete — a delete correctly lands on the `NotFound` arm and is legitimate absence), `load_palace` reports `NotFound`, and the caller's `Err(PalaceStoreError::NotFound(_)) => continue` at line 237-238 silently skips it — reintroducing this PR's own bug class one layer deeper. The window is narrow (a concurrent permission change between two syscalls) and `load_palace` isn't part of the diff, which is why this doesn't move the verdict, but the new doc comment's claim ("Absence and failure-to-determine stay distinct at every step") isn't fully true while this call exists. Fix: `load_palace`'s guard to `try_exists()`, matching the sibling site. **Disposition: Promote** — separable, narrow, not blocking.

Everything else in the loop (entry `Err`, `is_dir` denial, `palace.json` stat denial, `load_palace`'s `Io`/`Json` arms) matches the stated design and is exercised by a real test.

## Empty / fresh registry

`PalaceRegistry::open` (`crates/trusty-common/src/memory_core/registry.rs:614-619`) calls `create_dir_all` then `list_palaces`. `list_palaces`'s top `try_exists()` on a just-created, empty dir returns `true`, `read_dir` is empty, loop body never runs, returns `Ok(vec![])`. `list_palaces_missing_dir_is_empty` and the fresh-install path both confirmed green in the run below. No regression.

## Tests — verified, not taken on claim

Ran `cargo test -p trusty-common --features memory-core --lib palace_store` in the worktree (non-root, `uid=502`):
```
running 9 tests
test memory_core::store::palace_store::tests::identity_txt_roundtrip ... ok
test memory_core::store::palace_store::tests::list_palaces_finds_saved_palaces ... ok
test memory_core::store::palace_store::tests::list_palaces_missing_dir_is_empty ... ok
test memory_core::store::palace_store::tests::list_palaces_propagates_an_undecodable_palace ... ok
test memory_core::store::palace_store::tests::list_palaces_propagates_an_unstattable_entry ... ok
test memory_core::store::palace_store::tests::list_palaces_propagates_an_unstattable_palace_json ... ok
test memory_core::store::palace_store::tests::list_palaces_propagates_an_unstattable_registry_dir ... ok
test memory_core::store::palace_store::tests::load_identity_missing_returns_none ... ok
test memory_core::store::palace_store::tests::palace_store_roundtrip ... ok
test result: ok. 9 passed; 0 failed
```
Then reverted only the loop body (lines 209-244) to the pre-fix `Err(_) => continue` / `is_dir()` / `.exists()` shape and reran — all 3 new tests go red for the stated reason, the pre-existing #5541 test unaffected:
```
test list_palaces_propagates_an_undecodable_palace ... FAILED
  panicked: an undecodable palace must not silently shorten the listing: [Palace { id: PalaceId("alpha"), ... }]
test list_palaces_propagates_an_unstattable_entry ... FAILED
  panicked: an entry that cannot be stat'd must not drop out of the listing: []
test list_palaces_propagates_an_unstattable_palace_json ... FAILED
  panicked: a palace.json that cannot be stat'd must not read as a non-palace dir: []
test list_palaces_propagates_an_unstattable_registry_dir ... ok   (unaffected — different code path, as claimed)
test result: FAILED. 6 passed; 3 failed
```
Restored the fixed file afterward (`git status` confirms clean, matches HEAD).

The permission tests are not vacuous under root: `uid=502` here, and both tests assert the denial actually took hold (panicking with an instructive message otherwise) before asserting the propagation — I observed the real assertion failures/panics, not the vacuous-pass message. Site 1's `0o400` on the registry (line 470) genuinely isolates from #5541's parent-mode-000 test: confirmed empirically above — reverting only the per-entry loop left `list_palaces_propagates_an_unstattable_registry_dir` (which locks the *parent* of the registry, exercising the top-level `try_exists()` call untouched by the revert) passing throughout.

## MEDIUM — the four new citations in the `Test:` doc comment are never mechanically checked, contrary to the PR's claim

`crates/trusty-common/src/memory_core/store/palace_store.rs:173-178`:
```
/// Test: `list_palaces_finds_saved_palaces` and
/// `list_palaces_missing_dir_is_empty` pin the benign arms;
/// `list_palaces_propagates_an_unstattable_registry_dir`,
/// `list_palaces_propagates_an_unstattable_entry`,
/// `list_palaces_propagates_an_unstattable_palace_json` and
/// `list_palaces_propagates_an_undecodable_palace` pin the error arms.
```
`scripts/check_test_pointers.sh`'s `candidate_names()` only scans the *leading run* of backtick citations and stops at the first token that isn't a citation/separator — by its own doc comment, this is a documented, accepted false-negative direction. I ran the real function from the script against this exact blob:
```
$ candidate_names "$blob"
NAME	list_palaces_finds_saved_palaces
NAME	list_palaces_missing_dir_is_empty
```
"pin the benign arms;" is prose, not a separator, so the scan stops there — the four citations after it (including all three of this PR's new tests) are never extracted as candidates, never checked, not even flagged dangling. Confirmed against the actual pushed commit (worktree at `6e228b710`, `git status` clean, file tracked via `git ls-tree`/`git ls-files`) — this is not a case of the file being skipped by `git ls-files` (the file is fully tracked and scanned; CI's own "Doc-comment pointer lint" run at https://github.com/bobmatnyc/trusty-tools/actions/runs/31551165741 confirms via its 14 reported violations, all in `crates/trusty-search`, none in this file). It's this specific citation list's prose that breaks the leading-run parse.

So the PR's claim — "zero failures in `trusty-common`, which also confirms my four new `Test:` pointers resolve" — overstates what green CI proves. The tests themselves are real and genuinely exercise the code (shown above), so there's no coverage gap in substance, but the mechanical cross-reference this codebase's Why/What/Test convention exists to enforce isn't actually wired up for these four names: if one of these three test functions gets renamed later, this gate will not catch the resulting dangling pointer. Fix is a one-line reword to keep all six citations in one unbroken leading run before any prose, e.g. "Test: `list_palaces_finds_saved_palaces`, `list_palaces_missing_dir_is_empty`, `list_palaces_propagates_an_unstattable_registry_dir`, `list_palaces_propagates_an_unstattable_entry`, `list_palaces_propagates_an_unstattable_palace_json` and `list_palaces_propagates_an_undecodable_palace` cover this: the first two pin the benign arms, the rest the error arms." **Disposition: Fix here** — same file, one line, no new risk.

## Untested fourth site (`read_dir` iterator `Err`)

`crates/trusty-common/src/memory_core/store/palace_store.rs:211`: `entry.map_err(|e| PalaceStoreError::io(registry_dir, e))?`. Forcing a *mid-iteration* `readdir` failure (as opposed to the already-covered open-time failures) needs either OS-level fault injection or a race whose timing isn't controllable — permission changes generally don't affect an already-open directory handle on POSIX, so the chmod tricks the other three tests use don't reach this site. I did not find a practical deterministic way to drive it either; the author's claim holds up. The code fails closed (propagates) rather than open, so the Fail-Open Check's actual runtime concern doesn't apply — only regression coverage is missing. **Disposition: Parent**, not blocking.

## Rejected alternatives

`PalaceStoreError` (`crates/trusty-common/src/memory_core/store/palace_store.rs:32`) carries no `#[non_exhaustive]` — confirmed by reading the enum definition. So the PR's stated reason for reusing `Io`/`Json` instead of adding a variant (adding one would be a semver break) is accurate. Both reused variants carry `path` + the underlying error — enough for a caller to name and act on the failure.

## Version claim

No `Cargo.toml` change in this PR's diff (only `changelog.d/`, `registry.rs`, `palace_store.rs` touched — confirmed via `gh pr diff 5545`'s file list). No public signature, field, or variant changed. No bump owed, as claimed.

## Known items, confirmed out of scope

`check_test_pointers.sh` reports 14 dangling pointers — ran the full scan locally (also matches CI job `31551165741`): all 14 are in `crates/trusty-search` (`doctor_checks/mcp_registration.rs`, `core/corpus/contrib.rs`, `core/corpus_recovery.rs`, `service/boot_markers.rs` x4, `service/persistence.rs` x2, `service/reindex/mod.rs`, `service/server/degraded.rs`, `service/server/health.rs`), none in this diff's three files. CI's "Public API / SemVer" check reports SUCCESS for this PR (not the libdbus-sys infra failure).

## Findings

| MEDIUM | crates/trusty-common/src/memory_core/store/palace_store.rs | 173-178 | The four newly-cited `Test:` names after "pin the benign arms;" are never extracted by `check_test_pointers.sh`'s leading-run parser (prose breaks the scan) — CI green here doesn't verify these citations, contrary to the PR's claim | Reorder the `Test:` line so all six citations form one unbroken leading backtick run before any prose | Fix here |
| MEDIUM | crates/trusty-common/src/memory_core/store/palace_store.rs | 152 | `load_palace`'s own `Path::exists()` guard still coerces a permission denial to `NotFound`; a denial landing in the TOCTOU window between `list_palaces`'s `try_exists()` check (line 228) and the `load_palace` call (line 235) is silently skipped, one layer under where this PR hardened the check | Change `load_palace`'s guard to `try_exists()` | Promote |
| MEDIUM | crates/trusty-common/src/memory_core/store/palace_store.rs | 211 | `read_dir` iterator `Err` arm has no regression test | Acceptable — genuinely hard to drive deterministically; code already fails closed | Parent |

## Notes

- A circular symlink (`ELOOP`) in the registry now hard-propagates via the `Err(e) => return Err(...)` arm at line 220, where before the fix it silently skipped (coerced through `is_dir()`). This is a behavior change in a case not enumerated by the PR's partition table. I'm under 80% confidence on which side of the partition it belongs — a circular symlink is arguably "cannot classify" (propagate is defensible) or arguably "definitely not a palace regardless of cause" (skip is defensible) — so this is a note, not a finding.
- `purge_palaces`/`rebuild_palaces`/`merge_palaces` (`trusty-memory`) already propagate `list_palaces`'s result via `?` (confirmed by reading `kg_rebuild.rs:369-377` and the corresponding `*_propagates_an_unreadable_data_root` tests) — the PR's premise holds.

---

Issue #4937, opened 2026-08-05
compose_session_instructions_display_matches_live_prompt asserts on live ~/.claude/agents, flakes under concurrent deploys


bobmatnyc:
**Defect.** `compose_session_instructions_display_matches_live_prompt` (crate `trusty-mpm`, `crates/trusty-mpm/src/bin/tm/tests_behavior_b_tests.rs:315`) asserts by comparing two enumerations of the developer's live `~/.claude/agents` directory rather than a fixture. Any process that deploys or removes an agent between the two calls fails the test.

**Evidence.** Observed 2026-08-05: five agents (`copyeditor`, `pangram-editor`, `proofreader`, `writer`, `writing-critic`) were deployed by another process mid-run, and the assertion diff was exactly those five.

Same defect class as [#4766](https://github.com/bobmatnyc/trusty-tools/issues/4766) (`resolve_pm_prompt_wrapper_matches_the_source_reporting_form` scanning the same directory twice) and the TMPDIR cross-test pollution fixed in [PR #4936](https://github.com/bobmatnyc/trusty-tools/pull/4936) — a different test, same shape: coupled to ambient environment rather than a fixture.

**Resolution.** Pin the test to a controlled fixture directory instead of the live roster, so concurrent agent deployment cannot affect it.

🤖🤖🤖 Generated with trusty-mpm — https://github.com/bobmatnyc/trusty-tools

bobmatnyc:
**New evidence 2026-08-06**, on branch `fix/5007-session-store-resilience` and its merge base — via PR #5027:
https://github.com/bobmatnyc/trusty-tools/pull/5027#issuecomment-5208687532
https://github.com/bobmatnyc/trusty-tools/pull/5027#issuecomment-5208976590

Both `tests_behavior_b::compose_session_instructions_display_matches_live_prompt` and its `_with_override` sibling failed together in a concurrent `cargo test -p trusty-mpm --bins --tests` run (1527 passed; 2 failed), and both pass alone (2 passed; 0 failed, 0.16s). Confirms the described mechanism: the lib suite's `managed_config` tests deploy skills/agents into the same live tier these tests read.

**Impact beyond flakiness:** because reds here are routine, agents started running `cargo test -p trusty-mpm --lib` only to get a clean read, which masked a genuinely red test in the `tm` bin target (`doctor_checks_match_run_doctor_names`) on the same PR. A flaky gate trained a narrower gate, which then hid a real defect.

[...]

PR #5565, opened 2026-08-12
fix(trusty-mpm): make a truncated delegation roster loud, and stop the compose test racing $HOME

### LINKED ISSUE #5544: flaky(trusty-mpm): $HOME-repointing writers in the tm bin test target race every non-serial reader
**Defect.** Six functions across four files in the `tm` bin test target repoint `$HOME` (and in some cases `$CLAUDE_CONFIG_DIR`) for the duration of a test, and are `#[serial_test::serial]` in `serial_test`'s default group. That serialises them against each other only — not against any non-serial test in the same target that reads the live roster tiers (`deployed_roster_section`, `scan_agents`, and anything downstream of them). A non-serial reader can straddle a live repoint mid-scan.

**Not this issue's original hypothesis.** The size-budget theory this issue originally shipped with is disproven — see [PR #5565](https://github.com/bobmatnyc/trusty-tools/pull/5565): `compose_session_instructions` is `build_system_prompt_for(project_dir)` verbatim, no second composition path, no truncation budget anywhere in the composer.

**Evidence.** #5565 measured the actual mechanism directly: `$CLAUDE_CONFIG_DIR` is absolute and survives a `$HOME` repoint, `~/.claude/agents` does not — `REAL=43 FAKED=38`, missing exactly `copyeditor`, `pangram-editor`, `proofreader`, `writer`, `writing-critic`. Pre-fix, three consecutive `--bin tm` runs: ok / ok / FAILED.

**What #5565 fixed vs. what's left.** #5565 pinned the two named victims (`compose_session_instructions_display_matches_live_prompt` and its `_with_override` sibling) to a seeded fixture and gave them `#[serial]`. It did not touch the six writers themselves — any other current or future non-serial test in this target that reads a `$HOME`-relative tier is still exposed to the same race.

**Relation to [#4937](https://github.com/bobmatnyc/trusty-tools/issues/4937).** #4937 names only the same two victims #5565 already fixed, with the same per-victim pin-to-fixture resolution — it does not cover the generator-side hazard this issue tracks. Also related: [#4766](https://github.com/bobmatnyc/trusty-tools/issues/4766), another victim of the same class in a different test.

**Durable lesson (from a [code-critic](https://github.com/bobmatnyc/trusty-tools/pull/4482) finding).** The audit unit for this class is per TEST TARGET, never per crate — `serial_test` groups and process-global env mutation cannot span test binaries, since each Cargo test target is its own process.

**Confidence: Reproduced** — mechanism and the exact five-agent diff measured directly in #5565; flake confirmed across repeated runs.

**Closure conditions.**
- Every `$HOME` / `$CLAUDE_CONFIG_DIR`-repointing test in the `tm` bin target is either joined to the same serial group as every reader of the tiers it mutates, or scoped so no reader can observe the repoint (e.g. injected root rather than a real env var).
- A regression test or shared guard demonstrates a newly added non-serial reader cannot silently race one of the six writers.
- `cargo test -p trusty-mpm --bin tm` runs green across multiple consecutive full-parallel runs with no recurrence of this flake class.

bobmatnyc:
Refs #5544, #4937. Does not close #5544 on its own — see "What this leaves open".

## Which composition path is correct: neither, they are the same call

`compose_session_instructions` does not filter the catalog. It calls
`build_system_prompt_for(project_dir)` verbatim
(`crates/trusty-mpm/src/bin/tm/commands/session.rs:650`) — the exact function the
test calls on the other side. There is no second composition path and no size
budget anywhere in the composer. #5544's hypothesis that bundled-asset content
shares a budget with the agent-catalog section is wrong.

What differs between the two sides is the roster INPUT. Each composition calls
`deployed_roster_section`, which rescans three tiers — `<project>/.claude/agents`,
`$CLAUDE_CONFIG_DIR/agents`, `$HOME/.claude/agents` — on every call.

## Order-dependence root cause: `#[serial]` protects writers, not readers

Sibling tests in the same `tm` bin target repoint `$HOME` at a tempdir
(`tests_project_trust_tests.rs:51`, `formatters/banner/source.rs`, four more).
They are `#[serial_test::serial]`, which serialises them against each other and
against nothing else. The compose tests were not serial, so they ran in parallel
with a live `$HOME` repoint and their two scans straddled it.

Why exactly five names: on this workstation `$CLAUDE_CONFIG_DIR` is an absolute
path, so that tier survives a `$HOME` repoint; `~/.claude/agents` does not, and
it is the only tier carrying `copyeditor`, `pangram-editor`, `proofreader`,
`writer`, `writing-critic`. Measured directly with a throwaway probe:

```
REAL=43 FAKED=38 MISSING=["copyeditor", "pangram-editor", "proofreader", "writer", "writing-critic"]
```

That is #5544's and #4937's diff, reproduced on demand. The failure is a flake,
not deterministic — three consecutive pre-fix `--bin tm` runs on the branch point:

```
run 1: test result: ok. 1645 passed; 0 failed; 1 ignored
run 2: test result: ok. 1645 passed; 0 failed; 1 ignored
run 3: test result: FAILED. 1643 passed; 2 failed; 1 ignored
```

#4937's diagnosis is the correct one; its "concurrent deploy" is in-process
`$HOME` faking rather than an external process. #5544's deterministic
content-size reproduction was timing, not a budget.

Fix: both tests pin `$HOME` and `$CLAUDE_CONFIG_DIR` to a seeded fixture and take
`#[serial]` themselves, joining the same group as every `$HOME` mutator in this
target (#4407). Not order-pinning, not `--test-threads=1`.

## Fail-open: two silent branches, now loud

The catalog was not truncated by a budget, but the audit found two reads that
absorb a failure into an empty result indistinguishable from "nothing was there":

- `delegation_authority.rs` — an agent file that exists but cannot be read was
  dropped by a bare `else { continue }`. No log, no error, roster one short.
- `scan_agents` — a tier failing to enumerate for any reason other than
  `NotFound` returned `Vec::new()`, byte-identical to an absent tier.

Both now land in a `RosterScan { agents, unreadable }`, log at `error!`, and
render a `ROSTER INCOMPLETE` banner naming every lost path at the top of the
composed `## Delegation Authority` section. A roster that is empty *because*
reads failed no longer degrades to `None`. Still fail-open — a bad directory
must not block a launch — but it can no longer claim a roster it does not have.

Error-arm tests, run against the PRE-FIX tree expressed in the pre-fix API:

```
running 2 tests
test prefix_probe_unreadable_tier_still_renders_a_section ... FAILED
test prefix_probe_unreadable_agent_file_is_reported ... FAILED

---- prefix_probe_unreadable_agent_file_is_reported stdout ----
a truncated roster must declare itself incomplete; got:
## Delegation Authority

The following agents are available for delegation. Route work to the
appropriate agent based on task type.

### engineer

---- prefix_probe_unreadable_tier_still_renders_a_section stdout ----
an unreadable tier must not be indistinguishable from an absent one

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 4920 filtered out
```

Post-fix the same assertions live as
`an_unreadable_agent_file_is_reported_not_silently_dropped`,
`an_unreadable_tier_directory_is_reported_not_silently_dropped`,
`a_roster_that_is_empty_because_reads_failed_still_renders_a_section`, plus
`roster_section_from_dirs_is_none_when_nothing_is_deployed` and
`generate_authority_is_unchanged_when_nothing_was_lost` for the healthy paths.

## Test ladder: rung 3

Localized behavior inside one crate. Every new item is additive (`scan_agents`,
`roster_from_dirs`, `generate_authority`, `deployed_roster_section` keep their
signatures and behaviour) and no crate depends on the trusty-mpm lib — checked
across all eight crates whose manifests mention it; none carries a `trusty-mpm =`
dependency line. So rung 4's workspace step does not apply.

```
( cargo fmt --check && cargo check -p trusty-mpm \
  && cargo clippy -p trusty-mpm --all-targets -- -D warnings \
  && cargo test -p trusty-mpm ) > /tmp/gates.txt 2>&1; echo "EXIT=$?"
EXIT=0
```

lib `4918 passed; 0 failed; 7 ignored`, bin `tm` `1645 passed; 0 failed; 1 ignored`,
all remaining targets green. Flake re-run, `cargo test -p trusty-mpm --bin tm`
×10 post-fix: `1645 passed; 0 failed` every run.

`bash scripts/check_line_cap.sh` → `0 violations — OK`.
`bash scripts/check_changelog_fragment.sh` → `OK trusty-mpm: changelog.d fragment present and valid`.
`bash scripts/check_sld.sh` → `0 error(s), 0 warning(s)`.

## What this leaves open

#5544's closure conditions ask for the truncation mechanism to be located and
documented, and for overflow to error or be confirmed not to be the cause. This
answers both — there is no size budget; the mechanism was a racing test — and
makes the genuine silent-truncation branches loud. It does not audit the other
`#[serial]`-writer/parallel-reader pairs in this target, of which there are more.


bobmatnyc:
## Verdict: WARN

Adversarial review at head `b630ab54a`. Independently recompiled the branch in an isolated worktree and reran the touched tests rather than trusting the pasted output — both confirm the PR's claims (see Verification below).

## Attack #1 — does the new fail-open path itself fail open?

Inside the four changed files: no. `roster_section_from_dirs` renders the `ROSTER INCOMPLETE` banner whenever `scan.unreadable` is non-empty, regardless of whether `agents` is also empty (`delegation_authority.rs:687-ish`, the `if agents.is_empty() && scan.unreadable.is_empty()` guard). `generate_authority_reporting` prepends the banner unconditionally when `unreadable` is non-empty. I could not construct a path where `unreadable` is non-empty and the banner is skipped, or where the composed prompt (`compose_session_instructions` / `build_system_prompt_for`) reads as complete while a read actually failed.

Outside the four changed files: yes, in the sense that matters — the PR's own claim ("it can no longer claim a roster it does not have") is false for two other consumers of the identical scan:

- `resolve_roster` (`delegation_authority.rs:625-627`, untouched by this diff) still calls the OLD non-reporting `roster_from_dirs`, discarding `.unreadable` entirely.
- `instruction_pipeline.rs:735` — `let agent_count = resolve_roster(&input.project_dir).len();` — feeds `PipelineOutput::agent_count`, which `commands/session/start.rs:180-183` prints on every `tm session start` as `"Instructions: {N} agents in delegation authority"`. A file/tier that fails to read now silently lowers this number with zero indication, on literally the command a user runs to start a session.
- `doctor_fs_checks.rs:99` — `check_agents`'s `Ok`/`Warn` roster count also comes from `resolve_roster(project).len()`. `tm doctor` — the tool the new banner text explicitly tells the operator to run ("run `tm doctor` before concluding it does not exist") — has no visibility into the loss it's being pointed at. The recommended remediation doesn't work.

So the fix closes the composed-prompt path (the actual subject of #5544) correctly, but "the fix can no longer claim a roster it does not have" oversells scope: two sibling consumers of the same shared resolver are exactly as fail-open-and-silent as before, one of them being the diagnostic tool the PR's own banner recommends.

## Attack #2 — is the diagnosis convenient or correct?

Verified independently, not taken on faith. `compose_session_instructions` (`session.rs`) calls `build_system_prompt_for(project_dir)` verbatim for `resolved_prompt`; `build_system_prompt_for` → `build_system_prompt_for_with_style` → `resolve_pm_prompt` + output-style injection, no truncation/size-budget code anywhere in that chain (grepped for `truncat|MAX_LEN|byte_limit|char_limit` across `instruction_overrides.rs`, `session_launch/mod.rs`, `instruction_pipeline.rs` — the only hit is an unrelated string-slice in delegation-block stripping). The diagnosis holds.

## Attack #3 — does pinning `$HOME`/`$CLAUDE_CONFIG_DIR` shrink test coverage?

No. Before the fix, on CI (no ambient `~/.claude/agents`) these two tests only ever exercised the roster-EMPTY branch (bundled-asset fallback) — the roster-present branch was untested there. `with_pinned_agent_tiers` seeds one real agent and the test asserts `display.contains("### engineer")`, so it now deterministically exercises the roster-present composer branch on every run, everywhere. Net improvement, not a coverage loss in the guise of isolation.

## Attack #4 — root cause left unfixed elsewhere

Confirmed: `grep -rn "serial_test::serial\]"` in `crates/trusty-mpm/src/bin/tm/` turns up writers in `tests_behavior_b_tests.rs`, `tests_behavior_c_tests.rs`, `tests_project_trust_tests.rs`, and `formatters/banner/source.rs`, all in the same default (unnamed) `serial_test` group the two fixed tests now join — consistent with the PR's own account. The PR discloses this gap in its body ("does not audit the other `#[serial]`-writer/parallel-reader pairs ... there are more") — correctly scoped as future work, not a hidden gap. No action needed; not blocking.

## Attack #5 — panics / stdout

No `unwrap()`/`expect()` added to library code (`delegation_authority.rs`); the `.expect(...)` calls are all in test fixtures. `tracing::error!` (upgraded from `warn!`) goes through the existing subscriber, which per project convention writes to stderr — this PR adds no new stdout write. `env::set_var`/`remove_var` are correctly wrapped in `unsafe` blocks (edition 2024 requirement) and gated to `#[serial_test::serial]` callers, with restoration running via `catch_unwind`/`resume_unwind` even on panic.

## Attack #6 — mechanism vs. outcome

The new tests deny read via real `chmod 0o000` (`deny_read`, `delegation_authority_tests.rs`) and assert on the actual `RosterScan.unreadable` / banner content, not a mocked failure — they target the mechanism, and the module-level comment confirms each fails against the pre-fix commit (independently plausible: `scan_agents_reporting`, `RosterScan`, and `ROSTER_INCOMPLETE_MARKER` don't exist pre-fix, so these are new API + new tests together, not retrofitted assertions). No vacuous-test concern.

## Verification performed

```
cd <isolated worktree at b630ab54a>
cargo test -p trusty-mpm --lib delegation_authority
  → test result: ok. 35 passed; 0 failed; 0 ignored; 4890 filtered out
cargo test -p trusty-mpm --bin tm compose_session_instructions_display_matches_live_prompt -- --test-threads=1
  → test result: ok. 2 passed; 0 failed; 0 ignored; 1644 filtered out
```

## Findings

| HIGH | `crates/trusty-mpm/src/core/delegation_authority.rs` | 625-627 | `resolve_roster` (the "one shared resolver," per its own doc comment) still routes through the old non-reporting `roster_from_dirs`, so `agent_count` (`instruction_pipeline.rs:735`, printed by `tm session start` at `commands/session/start.rs:183`) and `tm doctor`'s roster count (`doctor_fs_checks.rs:99`) can silently undercount on exactly the failure this PR set out to make loud — with no signal, on the two places a human is most likely to be looking for a signal. | Route `resolve_roster` through `roster_from_dirs_reporting` (or add a `resolve_roster_reporting`) and surface `unreadable` in `PipelineOutput` and `DoctorCheck`, or at minimum note the gap explicitly in the PR body's "What this leaves open" section, which currently only names the test-flake root cause. | Promote |
| LOW | `crates/trusty-mpm/src/core/delegation_authority.rs` | ~687 (the `roster_section_from_dirs` no-agents branch) | The `info!` log for "no deployed agents found" lost the `project = %project_dir.display()` field it had before the `deployed_roster_section` → `roster_section_from_dirs` split (it now only logs `tiers = dirs.len()`) — a minor observability regression, not a correctness issue. | Have `deployed_roster_section` log the project-qualified message itself, or accept the loss. | Fix here |

## Notes

- The two doctor-adjacent HIGH-finding consumers are pre-existing behavior, not newly introduced by this diff — the PR doesn't make them worse, it just doesn't deliver on its stated "can no longer claim a roster it does not have" for them. That's why this is WARN, not BLOCK.
- Inherited red `Durable writes hold the teardown guard` (trusty-search/src/core/corpus/contrib.rs) is another session's failure; this PR's diff touches zero files there and is excluded from this review per the dispatch instructions.
- Line-cap, changelog-fragment, and SLD gates all reported green in CI; not independently re-verified since they're mechanical and unrelated to the logic under review.

bobmatnyc:
Both critic findings fixed here, at `99489721e`.

## HIGH — `resolve_roster`'s two count consumers

`resolve_roster_reporting` added; `resolve_roster` is now `.agents` off it. Both surfaces that publish the count route through it.

**`tm session start`** — `PipelineOutput` gains `unreadable_agent_paths` plus `roster_is_complete()`, filled from the reporting scan in `build_instructions`. The print moved into a testable `delegation_roster_line` (`formatters/session.rs`): unchanged when nothing was lost, otherwise `Instructions: at least N agents in delegation authority — ROSTER INCOMPLETE: 1 path(s) could not be read, so the real count is higher (…)`, with each lost path also on stderr.

**`tm doctor`** — `check_agents` reports `at least N delegatable — ROSTER INCOMPLETE: 1 path(s) unreadable (…)` and **downgrades `Ok` to `Warn`**. This was the sharp edge: the banner says "run `tm doctor`", and doctor answered `Ok … 43 delegatable` with an agent it could not read.

## LOW — logging-field regression

`deployed_roster_section` now emits the "no deployed agents found" `info!` itself with `project = …` restored; the tier-only seam no longer logs, since only the project-aware entry point knows the project.

## Error-arm evidence

Pre-fix (round-2 changes stashed, probes written against the pre-fix API):

```
running 2 tests
test daemon::doctor::doctor_fs_checks::tests::prefix_probe_agents_check_is_not_ok_when_a_roster_read_failed ... FAILED
test core::instruction_pipeline::tests::prefix_probe_session_start_line_declares_incompleteness ... FAILED

---- prefix_probe_agents_check_is_not_ok_when_a_roster_read_failed stdout ----
assertion `left != right` failed: an incomplete roster must not report as healthy: 1 agent file(s) deployed in /var/folders/…/claude-config/agents with manifest; 43 delegatable
  left: Ok
 right: Ok

---- prefix_probe_session_start_line_declares_incompleteness stdout ----
the session-start line must declare a truncated roster; got: Instructions: 43 agents in delegation authority

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 4925 filtered out
```

Post-fix, as committed tests:

```
running 3 tests
test daemon::doctor::doctor_fs_checks::tests::agents_check_is_not_ok_when_a_roster_read_failed ... ok
test core::instruction_pipeline::tests::build_instructions_reports_an_unreadable_agent_file ... ok
test core::instruction_pipeline::tests::build_instructions_reports_a_complete_roster_when_nothing_failed ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 4925 filtered out

running 2 tests
test tests::session_start_roster_line_is_plain_when_the_roster_is_whole ... ok
test tests::session_start_roster_line_declares_an_incomplete_roster ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1646 filtered out
```

## Gates, re-run on the merged head

```
( cargo fmt --check && cargo check -p trusty-mpm \
  && cargo clippy -p trusty-mpm --all-targets -- -D warnings \
  && cargo test -p trusty-mpm ) ; EXIT=0
```
lib `4921 passed; 0 failed; 7 ignored`, bin `tm` `1647 passed; 0 failed; 1 ignored`, all 47 targets ok.
`check_line_cap.sh` + `check_changelog_fragment.sh` → EXIT=0 (`0 violations`; `all 1 crate(s) with source changes are recorded`). Changelog fragment extended with the two count surfaces.

Head also carries the merge of `origin/main` through #5562, which clears the inherited `Durable writes hold the teardown guard` red.

